Issue #2998, opened 2026-08-26
Tomb Raider I-III Remastered (PPSA16901): world surfaces sample ONE texel each — degenerate UVs, not missing textures


mattias800:
## Summary

With the 32-bit index-buffer fix in place, Tomb Raider I-III Remastered (`PPSA16901`) renders Croft
Manor with **fully correct geometry** — steps, walls, hedges, trees, Lara and Winston are all
correctly shaped and animate — but **every world surface draws untextured**, flat-shaded and
noticeably over-bright.

Its menus are unaffected: the Tomb Raider I/II/III title screens are fully textured, so this is not
a global texture failure.

## Evidence

- The level's texture atlas **is bound and does carry real data**: a world draw binds a
  `512x512x256` block-compressed array with 6 mips, `footprint=90177536`, `nz=84070204` — 84 MB of
  non-zero texels. It is not an empty or unresolved descriptor.
- A boot-to-gameplay run produces **zero** `[recompile-reject]` lines and **zero** `[compute] skip`
  lines, so no shader or dispatch is being declined. Whatever is wrong is semantic.
- 689 `PS TEX` bindings are present across the 72 draws of the sampled frame.
- The same frame's character meshes are also untextured, while every menu is textured — so the split
  is world/character shaders against UI shaders, not "some textures missing".

## Reproduction

```bash
PROSPER_GAME_ROOT=<DUMP_ROOT> PROSPER_RENDER=1 SDL_VIDEODRIVER=offscreen \
PROSPER_PAD_SCRIPT=@scripts/tomb-raider-PPSA16901/reach-gameplay.pad \
  ./build-linux/screenshot <DUMP_ROOT>/PPSA16901-app0 --seconds 2 --count 65 --out ~/tr-work/tex
```

Gameplay is reached at roughly t=55 s. A replayable frame is cheap to obtain with
`PROSPER_GRAB_BUNDLE_AFTER_MS=110000 PROSPER_CAPTURE_FRAMES=16` on `prosper-app`; the bundle replays
the defect offline and deterministically, which is a much faster loop than re-booting.

**Note the one thing that bundle cannot answer:** a `.prgbundle` replays **pre-decoded** draws, so it
is blind to anything in the index/vertex decode path — an instrument placed there does not fire under
`--bundle`. That is how the index defect was nearly mis-measured. Texture work should be fine on the
bundle; decode work must be tested live.

## Not yet investigated

[...]

Issue #325, opened 2026-07-10
gpu: texture arrays (2D_ARRAY / type=9) not supported — array slice index dropped, only slice 0 samples


mattias800:
## Limitation
prosper does not handle **texture-array** descriptors:
- `decode_image_descriptor` decodes `d.type` (image type; **9 = 2D array**, 8 = 1D array, etc.) but it is used **only in the `[t#]` log** — `build_shader_resources` and the resource-binding layer ignore it and treat every texture as a plain 2D texture.
- The recompiler's `sampled_image_type()` (`rdna2_to_spirv.cpp`) always emits `OpTypeImage ... Arrayed=0`, and there is no 2D-array `image_sample` path — only `image_sample_2d` (u,v) and `image_sample_3d` (u,v,w). A 2D-array MIMG sample is mapped to 2D, **dropping the array-slice coordinate**.

## Effect
An array texture is bound and sampled as its **slice 0** only. Any content the game stores in **non-zero array slices renders wrong / missing**. The resource upload also only handles a single 2D surface, so array slices beyond 0 are never uploaded.

## Where it surfaced
The Messenger's title-screen **shine mask** is a `type=9` (2D-array) texture (see #240). In that specific case it's harmless — the property is an *unset* Unity default (white in all slices) — but it exposed the general gap. Titles that legitimately use `Texture2DArray` for gameplay content (sprite/tile arrays, terrain, etc.) would render only slice 0.

## Fix sketch
- Carry `d.type` onto `ShaderResource` and, for array types, declare an `Arrayed=1` `OpTypeImage` + a 2D-array `image_sample` that constructs a 3-component `(u, v, slice)` coord from the MIMG operands.
- Upload all array slices to a `VK_IMAGE_VIEW_TYPE_2D_ARRAY` image and bind a matching view.

## Priority
Low for The Messenger specifically (2D pixel-art, mostly atlases), but a real correctness gap for any array-using content (incl. the UE4 title). Filed while root-causing #240.

mattias800:
CLAIMING: opus-verify-session | branch fix/issue-325-texture-arrays | 2026-07-10T15:38:00Z

mattias800:
PR #336 lands the **first increment**: 2D_ARRAY (`DIM=5`) image_samples now **recompile** (sampled as base slice 0) instead of rejecting the whole shader — so array-sampling draws render instead of being silently skipped. Correct for single-layer arrays (Unity defaults); multi-layer samples slice 0.

[...]

PR #2408, opened 2026-08-09
fix(vk): request two device capabilities we already emit against, and one present semaphore per swapchain image

mattias800:
Three Vulkan validation violations found on GTA V (`PPSA04263`) while investigating #2396. Each is an
independent defect; **none of them fixes that crash** (its cause is a malformed SPIR-V module, fixed
separately in #2405). They are worth landing on their own merits.

## 1. `shaderInt64` was never requested

The recompiler emits SPIR-V declaring the `Int64` capability — 24 modules per GTA V run — and device
creation (`tests/render_runner.h`) requested only `robustBufferAccess`.

```
vkCreateShaderModule(): SPIR-V Capability Int64 was declared, but one of the following requirements
is required (shaderInt64)
```

`VUID-VkShaderModuleCreateInfo-pCode-08740`: **24 → 0**.

## 2. `independentBlend` was never requested

We build per-attachment blend state (MRT targets legitimately differ), which *requires* the feature;
without it the spec demands every `pAttachments` entry be identical.

```
vkCreateGraphicsPipelines(): pColorBlendState->pAttachments[1] is different than pAttachments[0]
and independentBlend feature was not enabled.
```

`VUID-VkPipelineColorBlendStateCreateInfo-pAttachments-00605`: **20 → 0**.

Both features are enabled **only when the device advertises them**, matching the `samplerAnisotropy`
precedent (#275): on a device lacking them the honest outcome is the layer error, not a silently
miscompiled shader.

## 3. One present semaphore per swapchain image

A single semaphore was shared across all swapchain images, so `vkQueueSubmit` signalled it while the
presentation engine could still hold it from the previous present.

```
vkQueueSubmit(): pSubmits[0].pSignalSemaphores[0] is being signaled by VkQueue, but it may still be
in use by VkSwapchainKHR
```

`VUID-vkQueueSubmit-pSignalSemaphores-00067`: **6 → 0**. `present_sem_for()` creates lazily per acquired
index — no image-count plumbing, cannot under-size, and falls back to the shared semaphore if creation
fails. Fixes #2403.

## The pattern, which is worth more than any single fix

`shaderInt64`, `independentBlend` and `maxVertexOutputComponents` all returned **zero** grep hits across
the repository. We emit SPIR-V and build pipelines against **assumed** capabilities rather than queried
ones. That is the shared root; these three are the instances GTA V happens to exercise.

## Verified — Linux/AMD, this exact head

| check | result |
| --- | --- |
| build | clean |
[...]

PR #1726, opened 2026-08-01
fix(gpu): stop emitting invalid SPIR-V, and make the gate that missed it able to see the path (#1711)

mattias800:
Fixes #1711.

Two defects, and the second is the one that matters. The invalid module is a small fix; the reason
nothing in the tree reported it for as long as it existed is the rest of this PR.

*(Body rewritten after review — it had described a coverage design that later rounds replaced,
which is the same failure this PR is about, in the artifact that becomes master history. The
superseded designs are recorded in the commit messages, which is where history belongs. No SHA is
named here deliberately: the last one went stale the moment the fix was pushed.)*

## 1. The invalid module — and it is not a test fixture

`prosper::gpu::build_compute_compare_uvec4()` (`src/gpu/spirv_builder.cpp`) declared the changed-flag
buffer with the same block type as the two data buffers — `OpTypeStruct { OpTypeRuntimeArray uvec4 }`,
`ArrayStride 16` — and then reached its scalar with two indices. Those walk struct member 0 → array
element 0, landing on a `uvec4`, while the result type is a `uint` pointer:

```
error: line 62: OpAccessChain result type <id> '4[%uint]' (OpTypeInt) does not match the type that
results from indexing into the base <id> '6[%v4uint]' (OpTypeVector).
  %43 = OpAccessChain %_ptr_StorageBuffer_uint %15 %uint_0 %uint_0
```

That module is created at runtime by the live frontend
(`VulkanComputeContext::prepare_compare_pipeline`, `frontends/shared/live_compute.cpp`), not only by a
test. Drivers are not required to validate SPIR-V and neither lavapipe nor RADV does, so both accepted
it and nothing reported it until the first validation-layer pass (#1704).

**It is also a correctness bug, not only a conformance one — which is what decides the fix.**
`live_compute.cpp` requires *and* enables `robustBufferAccess`. Under the old declaration the flag's
runtime array had `ArrayStride 16` over a **4-byte** bound range, so its robust length is
`floor(4/16) = 0` and element 0 is **out of bounds**: a conformant driver is permitted to discard the
`OpAtomicExchange` outright. The flag then stays zero, `gpu_result_unchanged` reads true, and
`live_compute.cpp:5781`/`:5909` **skip writing the dispatch's result back to guest memory entirely**.

Adding a third index — the vector-component index — also validates, and that is what the issue's
handover analysis proposed. It preserves the out-of-bounds property, and keeps describing a
16-byte-stride array over a binding that `live_compute.cpp` writes as
[...]

PR #3047, opened 2026-08-26
gpu: give the Vulkan validation layer somewhere to report (PROSPER_VK_VALIDATION)

mattias800:
## The gap

prosper registers **no `VkDebugUtilsMessenger` anywhere** and enables no instance layers. The loader
will happily load the validation layer from the environment alone
(`VK_LOADER_LAYERS_ENABLE=VK_LAYER_KHRONOS_validation`) — so it *looks* like validation is available —
but with no messenger registered, every message it produces is discarded.

The consequence is worse than a missing feature: **"no validation errors" has never once been a
measurement in this project, on any title.** A silent layer and a correct application produce
byte-identical output. If a note in an issue or a status doc says validation was clean and it
predates this PR, it is void.

I walked into it three times on #325 in one session: read a zero-VUID run as clean; found the layer
had never loaded; enabled it properly and found nothing changed, because nothing was listening.

## What this adds

`PROSPER_VK_VALIDATION=1` enables the layer **and** registers a messenger. Off by default — the layer
costs real time per draw.

It announces which of three states it is in, so a clean result is falsifiable rather than merely
quiet:

| line | meaning |
| --- | --- |
| `[vk-validation] active (warnings and errors, 8 per message id)` | armed; a clean run is a real result |
| `[vk-validation] PROSPER_VK_VALIDATION set but ... is not installed; NO validation is running` | the layer is absent — do not read the silence |
| `[vk-validation] layer loaded but the debug messenger could NOT be registered ...` | output would be discarded; treat a clean result as void |

Reports are **rate-limited to 8 per message id**. One violated VUID in a per-draw path otherwise
fills the disk, and on this machine a large run log takes the shared tmpfs — and the Bash tool for
every concurrent agent — with it.

## First use

A change that made Tomb Raider's gameplay render as a uniform single-colour fill produced **zero
VUIDs** with the sink armed. That converted "we cannot tell whether this is Vulkan misuse" into "it
is not one, so look at the logic" — a question that was unanswerable in this project until the
messenger existed. Detail on #2998.

## Verification

- `cmake --build prosper/build-linux -j10` — clean
- `ctest --no-tests=error -j4` — **310 of 310 pass**
- All four states exercised on this build:
  - `PROSPER_VK_VALIDATION=1` → `active`
  - unset → **no `vk-validation` line at all** (0 occurrences)
  - `VK_LAYER_PATH=/nonexistent` → the "NOT installed" line, not silence
  - the messenger-failure branch is the remaining one, reachable only if `vkCreateDebugUtilsMessengerEXT` is absent after the layer loads
- Documented in `docs/GPU_PROFILING_EXTERNAL.md`, next to the other external-tool recipes and their
  failure modes

## Scope

Purely additive and opt-in. With the variable unset, instance creation is byte-for-byte what it was:
no layer, no extra extension, no messenger.

### DIFF
--- a/prosper/docs/GPU_PROFILING_EXTERNAL.md
+++ b/prosper/docs/GPU_PROFILING_EXTERNAL.md
@@ -144,12 +144,17 @@ the GPU vendor cannot see.
 
 ## Vulkan validation (`PROSPER_VK_VALIDATION=1`)
 
-**Read `tools/vkval/` first.** `vk_validation_scan.py` (#1704) runs the whole ctest suite under the
-validation layer and diffs against `tools/vkval/allowlist.txt`, which carries a dated baseline —
-7 ids / 187 messages on lavapipe, 9 ids / 51 messages on RADV, measured 2026-08-01 — and it is
-registered with ctest as `vkval_scan_logic`. Three VUIDs were fixed off the back of it (#1713,
-#1717, #1726). **That is the project's validation guard, and it works without any messenger**,
-because VVL's default `debug_action` writes to stdout/stderr on its own.
+`vk_validation_scan.py` (#1704) runs the whole ctest suite under the validation layer and diffs
+against `tools/vkval/allowlist.txt`, and is registered with ctest as `vkval_scan_logic`. **Four
+VUIDs have been fixed off the back of it** (#1713, #1714, #1717, #1726). **That is the project's
+validation guard, and it works without any messenger**, because VVL's default `debug_action` writes
+to stdout/stderr on its own.
+
+Read the id and message counts from `allowlist.txt` itself rather than from anywhere that quotes
+them. Its header amends its own baseline on every fix — three times so far — and the file's own rule
+is that an unexplained drop in the id count reads as *"the scan broke"*. A figure quoted elsewhere
+and left to go stale therefore manufactures exactly that false alarm; this paragraph used to carry
+one.
 
 `PROSPER_VK_VALIDATION=1` is for the other case: validating **one interactive or routed run** of a
 title, in-process. It enables the layer and registers a `VkDebugUtilsMessenger`, which adds over the
--- a/prosper/docs/GPU_PROFILING_EXTERNAL.md
+++ b/prosper/docs/GPU_PROFILING_EXTERNAL.md
@@ -150,11 +150,14 @@ VUIDs have been fixed off the back of it** (#1713, #1714, #1717, #1726). **That 
 validation guard, and it works without any messenger**, because VVL's default `debug_action` writes
 to stdout/stderr on its own.
 
-Read the id and message counts from `allowlist.txt` itself rather than from anywhere that quotes
-them. Its header amends its own baseline on every fix — three times so far — and the file's own rule
-is that an unexplained drop in the id count reads as *"the scan broke"*. A figure quoted elsewhere
-and left to go stale therefore manufactures exactly that false alarm; this paragraph used to carry
-one.
+**Do not quote id or message counts from anywhere — including `allowlist.txt`'s own header.** Run
+`vk_validation_scan.py` and read what it computes (`[vkval] N distinct message ID(s), M message(s)
+total`). The header records a baseline and is amended *sometimes*: #1726, #1717 and #1713 each
+amended it, but #1714 deleted its entry without doing so, which leaves the header's last stated
+ledger one id high in both columns, and the lavapipe figure at the top has never been amended at
+all. This paragraph twice carried a stale figure of its own before saying that — the file's rule is
+that an unexplained drop in the id count reads as *"the scan broke"*, so a quoted count is a false
+alarm waiting to happen, wherever it is quoted from.
 
 `PROSPER_VK_VALIDATION=1` is for the other case: validating **one interactive or routed run** of a
 title, in-process. It enables the layer and registers a `VkDebugUtilsMessenger`, which adds over the
--- a/prosper/docs/GPU_PROFILING_EXTERNAL.md
+++ b/prosper/docs/GPU_PROFILING_EXTERNAL.md
@@ -144,33 +144,44 @@ the GPU vendor cannot see.
 
 ## Vulkan validation (`PROSPER_VK_VALIDATION=1`)
 
-Enables the Khronos validation layer **and** registers a debug messenger for it. Both halves matter,
-and only the first is obvious: the loader will happily load the layer from the environment alone
-(`VK_LOADER_LAYERS_ENABLE=VK_LAYER_KHRONOS_validation`), but with no `VkDebugUtilsMessenger`
-registered every message it produces is discarded. That reads exactly like a clean validation run.
+**Read `tools/vkval/` first.** `vk_validation_scan.py` (#1704) runs the whole ctest suite under the
+validation layer and diffs against `tools/vkval/allowlist.txt`, which carries a dated baseline —
+7 ids / 187 messages on lavapipe, 9 ids / 51 messages on RADV, measured 2026-08-01 — and it is
+registered with ctest as `vkval_scan_logic`. Three VUIDs were fixed off the back of it (#1713,
+#1717, #1726). **That is the project's validation guard, and it works without any messenger**,
+because VVL's default `debug_action` writes to stdout/stderr on its own.
 
-**Before this existed, prosper had no messenger anywhere** — so "no validation errors" had never once
-been a measurement in this project, on any title. If you find a note in an issue or a status doc
-saying validation was clean, and it predates this, it is void.
+`PROSPER_VK_VALIDATION=1` is for the other case: validating **one interactive or routed run** of a
+title, in-process. It enables the layer and registers a `VkDebugUtilsMessenger`, which adds over the
+default action:
+
+- **rate limiting** — 8 reports per message id. One violated VUID in a per-draw path otherwise fills
+  the disk, and on this machine a large run log takes the shared tmpfs (and every agent's shell)
+  with it.
+- **a run that says whether it is armed**, so a clean result is falsifiable rather than merely quiet.
+- **coverage when the default action is off**, which any `VK_LAYER_*` settings file can arrange.
 
 ```bash
 PROSPER_VK_VALIDATION=1 PROSPER_RENDER=1 PROSPER_GUEST_ARGS=-force-gfx-direct \
     ./prosper/build-linux/screenshot --out ~/work --seconds 2 --count 30 <DUMP_ROOT>/<TITLE_ID>-app0
 ```
 
-It announces itself, which is the point — the tool tells you whether it is actually running:
+It announces which state it is in — the point of the switch:
 
 | line | meaning |
 | --- | --- |
 | `[vk-validation] active (warnings and errors, 8 per message id)` | armed; a clean run is a real result |
-| `[vk-validation] PROSPER_VK_VALIDATION set but ... is not installed; NO validation is running` | the layer is absent — install it, do not read the silence |
-| `[vk-validation] layer loaded but the debug messenger could NOT be registered ...` | output would be discarded; treat a clean result as void |
+| `... is not installed; NO validation is running` | the layer is absent — do not read the silence |
+| `... VK_EXT_debug_utils is not advertised; the layer will run with its DEFAULT ... action` | the layer still validates, but this process does not rate-limit or tag it |
+| `... the validation layer was DROPPED and NO validation is running` | instance creation fell back to a bare create info; a clean result is void |
+| `... layer loaded but the debug messenger could NOT be registered` | output discarded; a clean result is void |
 
-Off by default: the layer costs real time per draw. Reports are **rate-limited to 8 per message id**,
-because one violated VUID in a per-draw path otherwise fills the disk — and on this machine a large
-run log takes the shared tmpfs, and the Bash tool with it.
+Off by default: the layer costs real time per draw.
 
-Worked example (#325, 2026-08-27): a change that made Tomb Raider's gameplay render as a uniform
-single-colour fill produced **zero** VUIDs with the sink armed. That turned "we cannot tell whether
-this is a Vulkan misuse" into "it is not one, so look at the logic" — a question that had been
-unanswerable in this project until the messenger existed.
+**Coverage caveat.** This covers the shared render-runner instance only. `prosper-app` creates its
+own instance, and `live_compute` has a private fallback instance; neither is affected by this
+variable.
+
+Worked example (#2998, 2026-08-27): a change that made Tomb Raider's gameplay render as a uniform
+single-colour fill produced **zero** VUIDs with the messenger armed, which moved the question from
+"is this a Vulkan misuse?" to "it is not, so look at the logic".
--- a/prosper/docs/GPU_PROFILING_EXTERNAL.md
+++ b/prosper/docs/GPU_PROFILING_EXTERNAL.md
@@ -141,3 +141,36 @@ to that class of error, costs nothing to run, and needs no maintenance from us.
 
 Reach for these first. Build a `PROSPER_*` switch only for something the guest-facing layer knows and
 the GPU vendor cannot see.
+
+## Vulkan validation (`PROSPER_VK_VALIDATION=1`)
+
+Enables the Khronos validation layer **and** registers a debug messenger for it. Both halves matter,
+and only the first is obvious: the loader will happily load the layer from the environment alone
+(`VK_LOADER_LAYERS_ENABLE=VK_LAYER_KHRONOS_validation`), but with no `VkDebugUtilsMessenger`
+registered every message it produces is discarded. That reads exactly like a clean validation run.
+
+**Before this existed, prosper had no messenger anywhere** — so "no validation errors" had never once
+been a measurement in this project, on any title. If you find a note in an issue or a status doc
+saying validation was clean, and it predates this, it is void.
+
+```bash
+PROSPER_VK_VALIDATION=1 PROSPER_RENDER=1 PROSPER_GUEST_ARGS=-force-gfx-direct \
+    ./prosper/build-linux/screenshot --out ~/work --seconds 2 --count 30 <DUMP_ROOT>/<TITLE_ID>-app0
+```
+
+It announces itself, which is the point — the tool tells you whether it is actually running:
+
+| line | meaning |
+| --- | --- |
+| `[vk-validation] active (warnings and errors, 8 per message id)` | armed; a clean run is a real result |
+| `[vk-validation] PROSPER_VK_VALIDATION set but ... is not installed; NO validation is running` | the layer is absent — install it, do not read the silence |
+| `[vk-validation] layer loaded but the debug messenger could NOT be registered ...` | output would be discarded; treat a clean result as void |
+
+Off by default: the layer costs real time per draw. Reports are **rate-limited to 8 per message id**,
+because one violated VUID in a per-draw path otherwise fills the disk — and on this machine a large
+run log takes the shared tmpfs, and the Bash tool with it.
+
+Worked example (#325, 2026-08-27): a change that made Tomb Raider's gameplay render as a uniform
+single-colour fill produced **zero** VUIDs with the sink armed. That turned "we cannot tell whether
+this is a Vulkan misuse" into "it is not one, so look at the logic" — a question that had been
+unanswerable in this project until the messenger existed.
--- a/prosper/tests/fixtures/render_runner.h
+++ b/prosper/tests/fixtures/render_runner.h
@@ -880,6 +880,8 @@ inline BackendRenderTimingStats backend_render_timing_stats() {
 // Give compute an explicit release-before-teardown handshake before adding a destructor here.
 struct RenderVkCtx {
     VkInstance inst = VK_NULL_HANDLE; VkPhysicalDevice phys = VK_NULL_HANDLE;
+    // Non-null only under PROSPER_VK_VALIDATION; without it the layer has no output sink.
+    VkDebugUtilsMessengerEXT debug_messenger = VK_NULL_HANDLE;
     VkDevice dev = VK_NULL_HANDLE; VkQueue queue = VK_NULL_HANDLE; uint32_t qfi = UINT32_MAX;
     VkDeviceSize storage_buffer_alignment = 1;
     double timestamp_period_ns = 0.0;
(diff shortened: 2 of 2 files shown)


mattias800 (commented):
**REJECTED** on the current head (514e454c) — four blocking findings. The mechanism is worth having and the inertness argument holds; the *claim* it is sold on is falsified by this repo's own committed measurements, and three code paths misreport or crash.

---

## BLOCKING 1 — the central claim is false, and the doc paragraph retires a working guard

`prosper/docs/GPU_PROFILING_EXTERNAL.md:150-152`

> **Before this existed, prosper had no messenger anywhere** — so "no validation errors" had never once been a measurement in this project, on any title. If you find a note in an issue or a status doc saying validation was clean, and it predates this, it is void.

That is not right, and it is the load-bearing sentence of the PR. Counter-evidence, all in-repo and all predating this branch:

- `prosper/tools/vkval/vk_validation_scan.py` — from **#1704** (CLOSED, titled *"No Vulkan validation layer runs anywhere…"*). It runs the whole ctest suite under the layer and gates on the findings. Registered as a ctest case at `prosper/CMakeLists.txt:297` (`vkval_scan_logic`).
- `prosper/tools/vkval/vk_validation_scan.py:166-171` sets **only** `VK_LOADER_LAYERS_ENABLE` and `VK_INSTANCE_LAYERS`. No settings file, no `report_flags` override, and — per this PR's own premise — no messenger anywhere in the process.
- `prosper/tools/vkval/allowlist.txt:30-34`, dated baseline: Ubuntu 24.04 / layers 1.3.275.0 / lavapipe → **7 ids, 187 messages**; Fedora 44 / layers 1.4.341.0 / RADV → **9 ids, 51 messages**.
- `allowlist.txt:39,44,48` — **three VUIDs fixed as a direct result**: #1726 (`pCode-08737`), #1717 (`VkMappedMemoryRange-size-01389`), #1713 (`vkCmdDraw-format-07753`).

Hundreds of validation messages reached the log, on two driver stacks, with no application messenger registered. The premise "no messenger ⇒ every message is discarded" is wrong: VVL's default `debug_action` writes to stdout/stderr on its own. An application messenger is *one* sink, not the only one.

The failure scenario is concrete and it is a docs failure, which under this charter is not the lesser kind: merged as written, the next agent reads "it is void", discards `tools/vkval`'s dated baseline and its 9-entry allowlist as unmeasured, and re-derives from zero the exact work #1704/#1713/#1717/#1726 already landed. This is the stale-figure rule pointed at a guard rather than a number.

Two alternatives the PR never excludes for *"enabled it properly and found nothing changed, because nothing was listening"*: the run was `prosper-app`, whose own instance at `prosper/frontends/prosper-app/main.cpp:245` this PR still does not touch; or VVL wrote to stdout while only stderr was captured.

**Required:** rewrite the section to what the change actually delivers — an explicit in-process sink plus a self-announcement, independent of VVL's default `debug_action` and of any layer-settings file — cross-link `tools/vkval/README.md`, and delete the "it is void" sentence.

---

## BLOCKING 2 — the bare-instance retry prints `active` with the layer dropped (the fourth silent state)

`prosper/tests/fixtures/render_runner.h:1013-1021`, then `:1022` and `:1055-1057`

The retry at `:1016-1017` builds `bare`, which carries **no layers and no extensions**. `validation_enabled` is never cleared, so `:1022` still runs on an instance that has no validation layer at all. `vkGetInstanceProcAddr` returns the loader's trampoline for `vkCreateDebugUtilsMessengerEXT` (loader-implemented extension), the create typically succeeds, and `:1055` prints:

    [vk-validation] active (warnings and errors, 8 per message id)

…on a run where nothing is validating anything. The reader then treats a zero-VUID result as a real result. If the create instead fails, `:1050` prints *"layer loaded but the debug messenger could NOT be registered"* — also false, since the layer was not loaded. Both branches misreport, so there is no path through the fallback that tells the truth.

This is precisely the falsifiability hole the PR exists to close, and it is worse than the state it replaces: silence invites suspicion, `active` does not. Finding 3 below makes it reachable.

**Fix:** in the `:1014` branch, set `validation_enabled = false;` and print a distinct fourth line naming the drop.

---

## BLOCKING 3 — `VK_EXT_debug_utils` pushed unchecked; failure also kills the WSI path

`prosper/tests/fixtures/render_runner.h:1010` (the `inst_exts.push_back("VK_EXT_debug_utils")`)

The earlier block builds `avail` and a `has()` lambda at `:955-960`, but both die with the enclosing scope at `:971`, so this push is unconditional. If the loader/ICD does not advertise `VK_EXT_debug_utils` at instance level, `vkCreateInstance` returns `VK_ERROR_EXTENSION_NOT_PRESENT` — and because the failure is on the *whole* create, the fallback at `:1014` also sets `r.present_surface_capable = false` and discards every WSI extension. So one absent extension silently degrades prosper-app's unified present path (#1270) for the entire run *and* lands in the false-`active` state above.

Low-probability on desktop Linux, but it is the case the fallback exists for, and the check is three lines.

**Fix:** hoist `has()` (or re-enumerate) and push only when advertised; otherwise report and leave `validation_enabled = false`.

*(The dangling-pointer concern is clean: `:1011-1012` reassigns `ici.ppEnabledExtensionNames = inst_exts.data()` after the `push_back`, so reallocation cannot leave `ici` pointing at freed storage, and `enabledExtensionCount` stays consistent.)*

---

## BLOCKING 4 — data race in the callback's rate-limit map

`prosper/tests/fixtures/render_runner.h:1038-1040`

    static std::unordered_map<int32_t, uint32_t> seen;
    const uint32_t n = ++seen[d ? d->messageIdNumber : 0];

Magic statics make the *initialization* thread-safe, not the access. The Vulkan spec requires `PFN_vkDebugUtilsMessengerCallbackEXT` to be thread-safe — the layer invokes it on whichever thread made the offending call — and this codebase is explicitly multi-threaded on Vulkan; the same struct carries `present_queue_shared` at `:926` with the comment *"submits need a mutex"*.

Failure scenario: two threads violate VUIDs concurrently, both hit `operator[]`, one rehashes while the other walks a bucket — UB, in practice a segfault or a corrupted map. The diagnostic then kills the run it was added to explain, and it will do so *only* under the interesting workloads (many messages, many threads), which is the worst possible distribution for an instrument.

`<mutex>` is already included at `:27`. A `static std::mutex` plus `lock_guard` is two lines.

---

## Non-blocking

- **`GPU_PROFILING_EXTERNAL.md:173` cites the wrong issue.** "Worked example (**#325**, 2026-08-27)" — #325 is *"gpu: texture arrays (2D_ARRAY / type=9) not supported"*, opened 2026-07-10. The Tomb Raider issue is **#2998**; the PR body has it right, the committed doc does not. A wrong cross-reference in a doc outlives the PR body that would have corrected it.
- **Coverage caveat missing.** This covers the `render_vk_ctx()` instance only. `prosper-app`'s present instance (`prosper/frontends/prosper-app/main.cpp:245`) and live_compute's private-device fallback (`prosper/frontends/shared/live/live_compute.cpp:2650`) get neither the layer nor the messenger. A user under `prosper-app` reading `active` will over-read the scope — one sentence in the doc fixes it.
- **Lifetime is fine.** `r.debug_messenger` is never destroyed, but nothing calls `vkDestroyInstance` on `r.inst` either — deliberate, per `:866-880` and the reasoning at `live_compute.cpp:10158-10186`. Leak at exit, not a use-after-free. The field is otherwise write-only, which is acceptable as handle retention.
- `kPerIdLimit` is scoped inside the lambda while `:1056` prints a literal `8u`; change one and the other silently disagrees. Hoist the constant.
- The rate limit never resets for the process lifetime. Documented as "8 per message id", so fair — but a later occurrence in a different context is invisible, worth a word in the doc.
- `d == nullptr` maps to key `0`, colliding with VVL's generic `messageIdNumber = 0`. Cosmetic.

## What checks out

- **Inertness (review point 1) passes.** With the variable unset, `want_validation` is false: no layer enumeration, no mutation of `inst_exts`, `inst_layers` never assigned into `ici`, and `:1022` skipped entirely. Instance creation is byte-for-byte unchanged and the surface-extension logic is untouched. `PROSPER_VK_VALIDATION=0` correctly reads as off.
- `return VK_FALSE` is correct — `VK_TRUE` is reserved for layer self-test and aborts the call with `VK_ERROR_VALIDATION_FAILED_EXT`.
- `d` is null-checked at both dereference sites (`:1040`, `:1044`).
- `messageIdNumber` is a stable per-VUID hash and a sensible rate-limit key.
- All required headers already present (`<unordered_map>:30`, `<cstring>:24`, `<cstdlib>:23`).

---

Findings 2, 3 and 4 are together maybe fifteen lines. Finding 1 is the one that matters: the code is a genuine improvement over relying on VVL's default output, and it does not need the false claim to justify itself. Re-request review on the corrected head and I will re-read.

mattias800:
All four blocking findings fixed at `8a47b6a2`. The first one was not a wording problem — **my
central claim was false**, and you were right to make it blocking.

## 1. The premise

`tools/vkval/vk_validation_scan.py` (#1704) runs the whole ctest suite under the validation layer,
diffs against `allowlist.txt` — which carries a dated baseline, **7 ids / 187 messages on lavapipe
and 9 ids / 51 messages on RADV, measured 2026-08-01** — and is ctest-registered as
`vkval_scan_logic`. Three VUIDs were fixed off the back of it. I verified all of that rather than
taking it from the review.

**How I got it wrong is the part worth recording:** I grepped `render_runner.h` for
`DebugUtilsMessenger`, found none, and concluded validation had never been measured in this project.
I never looked in `tools/`. Absence from the file I happened to open is not absence from the project
— and the paragraph I wrote would have retired a working guard as unmeasured.

The switch now makes a narrower and true claim: it validates **one interactive or routed run**
in-process, which the suite scanner does not, and adds rate limiting, a run that says whether it is
armed, and coverage when the default action is off.

## 2. The bare-instance retry announced `active` with the layer gone

Fixed — and this was the worst of the four, because it laundered exactly the null the switch exists
to expose. It now prints that the layer was **DROPPED** and the result is void.

## 3. `VK_EXT_debug_utils` pushed unchecked

You are right that the `has()` lambda died with its scope. Worse than I realised: an unadvertised
extension fails the *whole* `vkCreateInstance`, which drops every WSI extension and clears
`present_surface_capable`, degrading prosper-app's unified present path (#1270) — a debugging switch
must never do that. The extension vector is now hoisted and consulted; an absent extension falls
back to the layer's default action and says so.

## 4. Data race

Fixed with a mutex. `<mutex>` was already included, as you noted.

Non-blocking, all taken: `#325` → `#2998` in the doc; the coverage caveat now names `prosper-app`'s
own instance and `live_compute`'s private fallback as uncovered; and the `8` literal now uses the
same constant as the limit.

## Verification

- build clean; `ctest --no-tests=error -j4` → **310 of 310 pass**
- armed state and default-off state re-checked on the rebuilt binary (`active` line present; zero
  `vk-validation` lines when unset)

The doc now opens by pointing at `tools/vkval/` rather than claiming to be the first thing that could
measure validation. Thank you for checking the claim against the repo instead of against the PR body
— it was exactly the kind of assertion that reads as already-verified.

mattias800 (commented):
**REJECTED** — one blocking finding, and it is small and mechanical. Findings 2, 3 and 4 are properly
discharged; I traced each rather than reading the reply. The remainder is that the corrected Finding-1
paragraph replaced a false claim with two wrong numbers, and the numbers are wrong in the direction
that makes the guard look bigger than it is.

## Blocking: the vkval figures are the pre-amendment baseline, and the fix count is short by one

`docs/GPU_PROFILING_EXTERNAL.md:149-150` and the same sentence at `render_runner.h:985`:

> ...which carries a dated baseline — 7 ids / 187 messages on lavapipe, 9 ids / 51 messages on RADV,
> measured 2026-08-01 — ... Three VUIDs were fixed off the back of it (#1713, #1717, #1726).

`allowlist.txt` does contain that line, under `# Measured baseline, 2026-08-01`. It then amends it
three times in the same header block, at lines 39, 44 and 48:

* `#1726` fixed `pCode-08737` → "6 ids on the CI runner and 8 on Fedora/RADV (measured there at 49 messages)"
* `#1717` fixed `size-01389` → "6 ids on the CI runner and 7 on Fedora/RADV"
* `#1713` fixed `format-07753` → "the ledger is now 5 ids on the CI runner and 6 on Fedora/RADV"

And the file today holds exactly **5 non-comment entries** (`1710` x2, `1712`, `1715`, `1716`). So
7/9 is not the current ledger, and `51 messages` on RADV was restated to 49 by the first amendment.

The fix count is also wrong: line 60 records a **fourth**, `#1714` — "FIXED. independentBlend is now
requested when the device advertises it... The scan's `required` check caught the stale entry on the
same change that fixed the defect". That is a VUID fixed off the back of the scan by the file's own
account, so it is four, not three.

**Why this is blocking rather than a nit.** The allowlist's own header, eight lines under the figure
you quoted, says an unexplained id-count drop is "far more likely to be the scan breaking than a
defect being fixed. Both are failures until someone says which." Your doc is now the "**Read
`tools/vkval/` first**" pointer, and it tells the next reader to expect 7 ids. They will run the scan,
see 5, and per the file's own rule reach for "the scanner broke" before "three VUIDs were fixed" —
which is precisely the false alarm the ledger is written to prevent. A superseded count published
under an authoritative pointer is worse than no count.

**Suggested fix — delete the counts rather than refresh them.** They have already gone stale four
times, and they will go stale on the next fix. The file is the live ledger and is self-describing;
the doc's job is to say that it exists and is dated. Something like:

    ...and diffs against `tools/vkval/allowlist.txt`, a ledger of deferred defects with a dated
    baseline (2026-08-01) that is amended down as each is fixed — four so far (#1713, #1714, #1717,
    #1726). It is registered with ctest as `vkval_scan_logic`.

Same edit at `render_runner.h:985`. Everything else in the paragraph is accurate and I verified it:
`vkval_scan_logic` is registered at `prosper/CMakeLists.txt:297`, and the default-`debug_action`
mechanism is correctly described.

**My share of this.** The 187/7 and 51/9 figures are the ones I handed you in the review, and you
took them in good faith after checking they were in the file. They were in the file — as the
superseded line. That is the citation-authority trap the charter describes working on me: I read the
baseline block, did not read the amendments under it, and my finding then carried enough authority
that verifying it meant confirming it. Worth recording somewhere, because the review is exactly where
this project says a wrong number travels furthest.

## Finding 2 — bare-instance retry — FIXED, verified by trace

`render_runner.h:1042-1050`. `validation_enabled` is set false inside the retry block, and the only
announcement site is `if (validation_enabled && messenger_wanted)` at `:1055`, strictly after it. So
`active` cannot print on a bare instance. I walked the other arms:

* layer on, `debug_utils` present, create fails, bare succeeds → `!inst_exts.empty()` is true → DROPPED
  printed, flag cleared. Correct.
* layer on, `debug_utils` absent, `inst_exts` empty (headless CI), create fails on the *layer* →
  `!inst_exts.empty()` is false → `return r` with a null instance. No announcement, but no false one
  either, and that is the pre-existing failure path. Acceptable.
* layer requested but absent → `validation_enabled` already false, so the DROPPED line correctly stays
  silent and the "not installed" line has already fired.

The message is accurate about state: the retry genuinely carries neither layers nor extensions.

## Finding 3 — unchecked `VK_EXT_debug_utils` — FIXED, hoisting is sound

* Availability check at `:1013-1015` scans the hoisted `avail`. Correct.
* No dangling: `inst_exts.push_back` may reallocate, but `ppEnabledExtensionNames = inst_exts.data()`
  is reassigned on the line after the push (`:1018-1019`), and nothing pushes to `inst_exts` after
  that. `inst_layers` was already function-scope.
* Absent extension → the layer is **still enabled** (`inst_layers` is pushed before the extension
  check) and runs on its default stdout/stderr action; `messenger_wanted` goes false so no messenger
  is attempted and no `active` is claimed. The message says exactly that. Correct on both counts.
* `messenger_wanted` is consulted at the only site that matters, `:1055`; `debug_messenger` stays
  `VK_NULL_HANDLE` on that path.

**Inertness with the variable unset — confirmed byte-for-byte.** The only unconditional change is
`std::vector<VkExtensionProperties> avail;` + `avail.resize(nie)` in place of `avail(nie)`.
`resize` on a default-constructed vector value-initialises `nie` elements, identical to the sized
constructor; the `has()` lambda binds by reference and still dies with its scope. `ici` is untouched
when `want_validation` is false. The hoist costs the vector a longer lifetime and nothing else.

Non-blocking: `avail` comes from `vkEnumerateInstanceExtensionProperties(nullptr, ...)`, i.e. the
layer-less enumeration. An extension advertised only by the validation layer would need the query
repeated with the layer name. `VK_EXT_debug_utils` is loader-implemented so this is fine in practice,
and the failure direction is conservative (falls back to the default action and says so) — just not
strictly the same question.

## Finding 4 — data race — FIXED

`:1069-1078`. `seen` has exactly one access in the callback and it is inside the `lock_guard` scope;
`n` is copied out and read after the unlock, which is correct. Both statics get C++11
thread-safe initialisation. The `fprintf` outside the lock is fine — a single stdio call locks the
`FILE*`, so lines interleave in order but never corrupt.

Non-blocking: the callback still carries its own `constexpr uint32_t kPerIdLimit = 8` with a "keep in
step with kMessengerPerIdLimit" comment, so the announced limit and the enforced limit are two
literals, not one. (The captureless lambda cannot cleanly reference the enclosing automatic, so the
real fix is lifting the constant to namespace scope.) A divergence there would make the `active` line
misstate the rate limit — worth closing while you are in the file, but it is not why I am rejecting.

## Nothing else new in the delta

Doc table rows now elide the `[vk-validation]` prefix to `...`, which is cosmetic. Coverage caveat
naming `prosper-app`'s own instance and `live_compute`'s private fallback is a real improvement and
is accurate. `#325` → `#2998` corrected.

Fix the counts and I will approve on the next head.

mattias800:
Fixed at `c6801090`. You are right, and the figures were stale in a way I should have caught: `allowlist.txt` amends its own baseline three times below the line I quoted, and **#1714 is a fourth fix I missed entirely**.

The counts are now **gone from both places** rather than refreshed — they have already gone stale four times — and both say to read them from `allowlist.txt`, which is the file that maintains them. The fix list is corrected to four (#1713, #1714, #1717, #1726).

Your point about *why* it matters is the one I have recorded in the commit: the allowlist's own rule is that an unexplained id-count drop reads as "the scan broke", so a doc promising 7 ids manufactures that false alarm for whoever compares it against a current run.

And thank you for naming where they came from. I copied them out of a review comment without opening the file — the citation-authority trap, working on a reviewer's numbers exactly as it works on anyone's. That is the second time tonight a number of mine survived on plausibility and died to an actual read of the source.

Also folded in your non-blocking note: the rate limit was two separate `8` literals and is now one constant, so the `active` line cannot come to misstate it.

mattias800 (commented):
Round 4, reviewing only `8a47b6a2..c6801090` plus the reply.

**REJECTED** — one blocking finding, and it is a single sentence. Everything else in the delta is
verified correct.

## What is fixed (verified, not taken from the reply)

- **The stale counts are genuinely gone from both files.** No id or message count survives anywhere
  in either: `GPU_PROFILING_EXTERNAL.md` and the `render_runner.h` comment block now contain no
  vkval figure at all. The only remaining numerals are the rate limit (`8 per message id`), which is
  a different quantity.
- **The fix list is right — four VUIDs.** Checked against `allowlist.txt` rather than the reply:
  `#1726` deleted `pCode-08737`, `#1717` deleted `size-01389`, `#1713` deleted `format-07753`, and
  `#1714` deleted `VUID-VkPipelineColorBlendStateCreateInfo-pAttachments-00605` in `433e93c8`
  (branch commit `6d782ca2`), leaving the FIXED block at `allowlist.txt:58-61`. Four.
- **The rate limit is one constant.** `kMessengerPerIdLimit = 8` is declared once; the callback's
  `kPerIdLimit` is now initialized from it and the `active` line prints it with `%u`. They cannot
  diverge. (The reference from the non-capturing callback is a constant expression, so it is not an
  odr-use and needs no capture — consistent with green CI.)
- **Findings 2, 3 and 4 are undisturbed.** The delta touches only prose, one comment, and that one
  initializer; the DROPPED-instance path, the hoisted `VK_EXT_debug_utils` check and the mutex are
  byte-identical to `8a47b6a2`.
- **Nothing else is introduced.** 2 files, +16/-9, all comments/docs plus the one constant.

## Blocking: the replacement guidance vouches for a figure that is itself stale

`docs/GPU_PROFILING_EXTERNAL.md:153-157`, the load-bearing clause on **line 154** (mirrored at
`tests/fixtures/render_runner.h:986`):

> Read the id and message counts from `allowlist.txt` itself rather than from anywhere that quotes
> them. **Its header amends its own baseline on every fix — three times so far** ...

`#1714` is the counter-example this same PR just added one paragraph above. `433e93c8` deleted the
`00605` entry and **did not amend the header** — it left an inline FIXED note instead. So:

- `allowlist.txt:50` (the `#1713` amendment, the header's last word on the ledger) says **"5 ids on
  the CI runner and 6 on Fedora/RADV"**. After `#1714` that is **one id high in both columns** —
  `#1714`'s own verified run reports `[vkval] 5 distinct message ID(s), 54 message(s) total` on
  RADV, and the file today holds 5 entries of which one is environment-dependent on the CI runner.
- The message counts are further behind still. The lavapipe **187** at `allowlist.txt:31` has never
  been amended through any of the four fixes, and the RADV figure was last restated as **49**
  (`allowlist.txt:40`) while the `#1714` run measured **54**.

So the paragraph redirects the reader from one stale quote to another, and the clause "amends its
own baseline on every fix" is precisely what suppresses the suspicion that would otherwise catch it.
A reader who follows the instruction reads 6 RADV ids, compares against a current run's 5, and hits
the "the scan broke" false alarm the paragraph exists to prevent — which is the original finding
surviving one level of indirection out. It is also internally inconsistent: four fixes above, "every
fix — three times" below.

**Remedy (one sentence, either file's wording works for both).** Point at the thing that *computes*
the figure instead of the thing that remembers it:

    python3 prosper/tools/vkval/vk_validation_scan.py --build-dir prosper/build-linux
    [vkval] N distinct message ID(s), M message(s) total        # vk_validation_scan.py:251

Drop "amends its own baseline on every fix" rather than re-deriving a correct count — the header is
a dated record, not a maintained one, and `#1714` proves the amendment is not automatic. Fixing
`allowlist.txt`'s header is a separate change and is not being asked for here.

## Non-blocking

The doc quotes the rate limit as a literal `8` twice (lines 163 and 178) while the code now has one
constant. Same staleness class, but self-contained, low-churn, and the table line is deliberately a
transcript of what the program prints. Fine as is.

A wording change to that one clause gets an immediate approval; nothing else is outstanding.

mattias800:
Fixed at `4e505c84`. You are right, and this is the sharper version of the finding: my round-three fix redirected the reader from one stale quote to **another**, while vouching for the destination.

`#1714` deleted its entry without amending the header, so the header's last stated ledger is one id high in both columns, and the lavapipe figure at the top has never been amended at all. "Amends its own baseline on every fix" was simply not true.

So the guidance is now: **do not quote counts from anywhere, including that header.** Run `vk_validation_scan.py` and read the figure it *computes* — `[vkval] N distinct message ID(s), M message(s) total` — which is the only one that cannot go stale.

This is the third revision of the same paragraph and **each one carried a wrong number**, which is itself the argument for removing the number rather than correcting it again. A count that lives in prose is a false alarm waiting to happen, precisely because the allowlist's rule is that an unexplained drop in the id count reads as "the scan broke".

Thank you for tracing `#1714` to its commit rather than trusting the header — that is the third finding in this PR where a plausible-looking figure of mine died to an actual read of the source.

mattias800 (commented):
Round 5, reviewing only `c6801090..4e505c84` plus the reply.

**APPROVED** — the blocking finding is discharged, and I traced every factual claim in the new text
to its source rather than reading it off the reply.

## The finding is fixed, and fixed in the right direction

Round 4's objection was that the fix redirected the reader from one stale quote to another while
vouching for the destination. The new text does not vouch for anything stored: *"Do not quote id or
message counts from anywhere — including `allowlist.txt`'s own header."* That is the correct stopping
point. `vk_validation_scan.py` recomputes the figure on every run, so it is the only one that cannot
go stale, and it is now what both places point at.

## Every new claim, checked against the source

**The #1714 claim is accurate as stated.** The commit that deleted the entry is `433e93c8`
(2026-08-09, landed as PR #2408, "request two device capabilities we already emit against"). Its diff
to `allowlist.txt` removes the `VUID-VkPipelineColorBlendStateCreateInfo-pAttachments-00605 |
required | shadow_compare_render` line, replaces it with a FIXED comment block, and touches **no
other line of the file** — the header is untouched. It is also the most recent change to the file, so
the header's last amendment is still #1713's.

**"One id high in both columns" is arithmetically right.** The last stated ledger is #1713's *"5 ids
on the CI runner and 6 on Fedora/RADV"*. The file today holds 5 entries, exactly 1 of them
`environment-dependent` — so 4 observable on the CI runner, 5 on Fedora/RADV. One high in both, and
one high in both *because* the deleted entry was `required` and therefore counted in each column.

**"#1726, #1717 and #1713 each amended it"** — confirmed at `0f67c3d6`, `8c380fb8` and `99cc2af2`
respectively, each adding its own "Amended by" paragraph in the same commit that deletes its entry.

**"the lavapipe figure at the top has never been amended at all"** — confirmed by line-history on the
block: `git log -L '/Measured baseline/,/RADV (STRIX_HALO)/'` returns exactly one commit, `4c0deb7e`,
the one that created the file. The 187-message lavapipe figure has never been rewritten, and none of
the three amendments restates a lavapipe *message* count either.

**The quoted output format is the literal format string.** `vk_validation_scan.py:251` is

    print(f"[vkval] {len(findings)} distinct message ID(s), {total} message(s) total")

which is character-for-character the doc's `[vkval] N distinct message ID(s), M message(s) total`,
including the parenthesised plurals. It is also on the unconditional path — printed before any
pass/fail branching — so a reader following the instruction gets that line whatever the outcome.

**"Four VUIDs have been fixed off the back of it (#1713, #1714, #1717, #1726)"** survives from the
previous round and is still right: the file was created with 9 entries and holds 5, and the four
deletions are the four commits above. That is a count, but it is an enumerated historical one, tied
to the four issue numbers beside it — it is checkable in one command and cannot silently drift the
way a baseline does. Correctly kept.

## No count survives

`grep -nE '[0-9]+ (ids|id|messages|message)'` over `GPU_PROFILING_EXTERNAL.md` returns nothing. The
only numbers left in either passage are the `8`-per-id rate limit, which is this PR's own
implementation constant rather than a stored measurement, and "one id high", which is a delta.

## The delta is documentation only

Two files. The `render_runner.h` hunk filtered for changed lines that are not `//` comments is empty
— zero code lines changed, so the tested behaviour at `c6801090` still stands unaltered at
`4e505c84`.

## One note, explicitly non-blocking

The inline code span in the doc wraps across a source line, so it reads as `M message(s)\ntotal` in
the raw file. Markdown collapses that to a space on render, so the displayed string is correct; do
not hold the merge for it, and do not "fix" it by lengthening the line past the file's wrap.

Three of the five findings across this review have been a plausible number that died to an actual
read of the source, and the paragraph's final state — no number at all, plus the command that
computes one — is the only version that could not have produced a fourth.

Issue #3048, opened 2026-08-26
Compute images are created with mipLevels=1, so image_load_mip cannot be implemented in the recompiler alone (Sonic Frontiers world)


mattias800:
Split out of #2790 with the measurement that makes it actionable. **Not claimed** — I am recording
the diagnosis rather than starting the change, because it lands in `live_compute.cpp`'s image
creation path and #325 is actively working nearby on texture arrays.

## What blocks Sonic Frontiers' world

With the route run correctly (see #2790 — it needs an empty save root), the Cyber Space stage is
reached, the HUD composites at 4K over a **black world**, and three compute kernels that dispatch
**2880 threads wide** — this title's exact scene-target width — never execute.

Re-measured on `d2b2e4d3`, their rejects are now MIMG instructions:

| program | pc | words | op |
| --- | --- | --- | --- |
| `0x2005714000` | 33 | `f0040308,00000101` | `image_load_mip` (0x01) |
| `0x200571bd00` | 81 | `f0040308,00000000` | `image_load_mip` (0x01) |
| `0x2005717e00` | 73 | `f0380118,00000404` | `image_get_resinfo` (0x0e) |

## Why the recompiler cannot fix this by itself

The obvious reading is "add a `Lod` operand to the fetch" — `image_load_mip` maps cleanly onto
`OpImageFetch` with `Lod`. **That would be wrong on its own**, and the reason is one line:

```cpp
// frontends/shared/live/live_compute.cpp
ici.mipLevels = 1;
```

Compute-path images are created with exactly **one** mip level, hard-coded. The graphics path is
different — `render_runner.h` uses `tci.mipLevels = upload.key.mip_levels` and builds a real chain.

The `[mimg-mip]` diagnostic shows the guest indexing a **12-level** resource:

```
program=0x2005714000 pc=33 shape=0 proven_zero_mip=0 img_dim=5/1 samples=1 mips=12
mip_tail=0 compressed=0 array_in_gfx=0 addr=0x2026900000 dataformat=1 ncomp=2
tile=27 dmask=0x3 unorm=0 glc=0 layer_stride=0
```

So emitting a `Lod` operand against a single-level image would sample a level that does not exist.
The current gate is at least honest about it: `image_load_mip` is admitted **only when it can be
specialised away** to an ordinary load —

```cpp
if (is_zero_mip_load &&
    (!rdna2_mimg_zero_mip_shape(in) || !res->proven_zero_mip ||
[...]

PR #3043, opened 2026-08-26
gpu(backend): admit guest 2D_ARRAY textures through the layer channel (#325)

mattias800:
Enabling half of #325. **No behaviour changes** — this only permits a shape nothing currently
produces.

## Why the backend was the blocker

Everything an array texture needs is already in place:

| what | where | status |
| --- | --- | --- |
| layer channel | `FrameResource::sample_count` — its own comment documents it for MSAA-as-array | already there |
| staging size | `tw*th*td*sample_count*bpp` | already layer-generic |
| image creation | `tci.arrayLayers = r.sample_count` | already layer-generic |
| view type | `sample_count > 1 ? VK_IMAGE_VIEW_TYPE_2D_ARRAY : ..._2D` | already layer-generic |
| **span contract** | `backend_texture_plane_span_valid` | **admitted only the 4-plane R32F MSAA shape** |

That last row rejected every graphics 2D_ARRAY at the backend boundary — which is exactly what the
recompiler's own comment records:

> *"The shared graphics backend currently exposes 2D_ARRAY textures as its documented base-slice 2D
> view (#325). Keep that established fallback for graphics until its resource uploader can create
> matching array views."*

This PR is that "until".

It also explains a failure I could not account for earlier: setting `sample_count = 256` on an
array made offline replay report `Vulkan render FAILED`, and a layer-count bisect showed it failing
at **4** layers as readily as at 256 — ruling out memory and pointing at a contract. It was this
predicate returning false.

## What the arm does

Same obligation the MSAA arm states — prove every layer is readable before any Vulkan object or
memcpy exists — with bytes-per-texel derived from the format rather than assumed. Mip generation is
already excluded for `sample_count > 1`, which is what stops a generated chain bleeding across layer
boundaries.

## Tests

Nine arms, chosen to fail if the arithmetic is dropped rather than to agree with it:

- a complete 8-layer span is accepted
- one byte short of the last layer is rejected
- **a span covering only ONE layer is rejected for an 8-layer resource** — a predicate checking a
  single layer's bytes would pass this and then memcpy 8 layers out of a 1-layer allocation
- **a 16-byte format over the same buffer is rejected** — fails if bpp is assumed to be 4
- volume / storage-image / null-pixel resources are rejected
- single-layer resources keep the historical implicit-span (`tex_byte_size == 0`) contract, or every
  ordinary texture in the project regresses
- the MSAA arm still routes by `img_dim`, so admitting arrays cannot widen it

[...]

Issue #2951, opened 2026-08-23
Metaphor: ReFantazio (PPSA20800) — the guest primary thread dies with SIGFPE at eboot+0x10019f6 (div by a zero register) ~5 s in


mattias800:
## Summary

With the direct-memory pool base fixed (#2934), *Metaphor: ReFantazio* (`PPSA20800`) gets six
seconds further: it loads its GFD assets through APR, opens two audio ports, initialises save data,
compute and fonts, and publishes **63 frames at 34.8 fps**. Then its **primary thread dies with
SIGFPE**, about 5 s in, and every later frame the frontend samples is stale by construction.

## Symptom

Frontend `tools/screenshot`, default route, no pad input, `PROSPER_RENDER=1
PROSPER_GUEST_ARGS=-force-gfx-direct`, Linux / AMD Radeon 8060S (RADV STRIX_HALO):

```text
[shot] guest thread ended: kind=2 detail=SIGFPE at addr=0x4110019f6 rip=0x4110019f6 (image+0x10019f6)
       rbp=0x20004d1800 rsp=0x20004d17c0
[shot] guest backtrace: 0x410ff7e7e (eboot+0xff7e7e)
[shot] guest backtrace: 0x410ff76a3 (eboot+0xff76a3)
[shot] guest backtrace: 0x410ff6597 (eboot+0xff6597)
[shot] guest backtrace: 0x4109bea7d (eboot+0x9bea7d)
[shot] guest backtrace: 0x4109be49c (eboot+0x9be49c)
[shot] guest backtrace: 0x4109bd4aa (eboot+0x9bd4aa)
[shot] guest backtrace: 0x4109e8ea9 (eboot+0x9e8ea9)
[shot] guest backtrace: 0x4110c59fb (eboot+0x10c59fb)
[shot] guest backtrace: 0x4111308b4 (eboot+0x11308b4)
[shot] guest backtrace: 0x4111539b6 (eboot+0x11539b6)
...
[shot] done: 24/24 screenshot(s); stop=request-satisfied source-distinct=1 pixel-distinct=1
       max-source-stale=115.0s max-pixel-stale=115.0s guest=faulted status=GUEST-FAULT-ALLOWED
```

## The faulting instruction

```text
 10019cd:  45 0f af fc      imul   r15d,r12d
 10019d1:  e8 3a 43 11 00   call   0x1115d10
 10019d6:  48 89 c7         mov    rdi,rax
[...]

Issue #2859, opened 2026-08-21
PC-relative embedded-table fold does not recognise a TYPED (MTBUF) consumer — s_getpc_b64 then rejects Sonic Frontiers' three Cyber Space scene kernels


mattias800:
## Summary

The PC-relative embedded-table fold (`detect_pcrel_tables`, `rdna2_cfg_support.hpp`) recognises an
**untyped** consumer (`MUBUF` raw loads) and a **scalar** one (`SMEM` s_buffer loads). It does not
recognise a **typed** one (`MTBUF`, `tbuffer_load_format_*`). When the consumer is typed, the table
never folds, so `s_getpc_b64` has nothing to justify it and `rdna2_emit_alu.cpp:1062` rejects the
whole program with `mode=unresolved-operand`.

*Sonic Frontiers* (`PPSA03831`) hits this in all three of its Cyber Space scene-width compute
kernels, twice each. Their terminal reject is that `s_getpc_b64`.

## The idiom, from the live dump

`PROSPER_SHADER_DUMP` + `tools/shader_inspect` on `0x2005714000` (Cyber Space `w6d01`,
`240x135x1` groups of `16x1` = 3840x135 threads):

```text
pc=0514 len=2 fmt=SOP1  op=0x003 words=be8303ff,10005004 dst=sgpr:3 src0=literal   ; V# word3
pc=0516 len=1 fmt=SOPK  op=0x000 words=b0020080 dst=sgpr:2 simm=128               ; V# num_records
pc=0517 len=1 fmt=SOP1  op=0x01f words=be801f00 dst=sgpr:0 src0=sgpr:0            ; s_getpc_b64 s[0:1]
pc=0518 len=2 fmt=SOP2  op=0x000 words=800000ff,00000148 dst=sgpr:0               ; s_add_u32 s0, 0x148, s0
pc=0520 len=1 fmt=SOP2  op=0x004 words=82010180 dst=sgpr:1                        ; s_addc_u32 s1, 0, s1
  ...
pc=0528 len=2 fmt=MTBUF op=0x000 words=e8b01000,80000a0a dst=vgpr:10 src0=vgpr:10 src1=sgpr:0 src2=inline-int:0
```

That last one is `tbuffer_load_format_x v10, v10, s[0:3], 0 offen` at **BUF_FMT 22**. It is the
byte-for-byte MUBUF idiom with a typed load in place of `buffer_load_dword`.

[...]

PR #3049, opened 2026-08-26
docs(frontiers): boot_trace is the fast loop, and the save-root precondition I claimed is withdrawn

mattias800:
## What this documents

Two things about running Sonic Frontiers' gameplay route, and one retraction.

**`boot_trace` is the fast loop.** 17–30 flips/s against a live 4K arm's 1.5–3.4, so `f2900`
(`GameModeStage`) is ~100–200 s rather than 14–32 minutes. Both switches are required and they fail
in different directions:

| switches | `.pac` lines | route position | mode markers |
| --- | --- | --- | --- |
| neither | 48 | — | — |
| `PROSPER_PAD_SCRIPT_LOG` only | 52 | present | **absent** |
| both | **1140** | present | **present** |

`PROSPER_FILELOG` reveals the mode markers; `PROSPER_PAD_SCRIPT_LOG` gives the route's position in its
own `fN` anchor units. A bare arm still prints ~48 lines containing `.pac`, but they are
`[apr] WARNING:` duplicate-asset notices rather than `[file] open` records — so a grep for `.pac`
returns hits while no game mode is nameable, which is exactly enough to look like file logging is on.

**If a run parks in the "Extras" menu it desynced.** One live-renderer arm sat there for its whole
length — 24/24 frames with content, `pixel-distinct=24`, guest healthy — and never reached the stage.
That is the end state this document already records for two causes: a `--warmup-seconds` window, and
flip/wall-clock desync under host load. Check `uptime` and which applies.

## A claim of mine, withdrawn

An earlier revision of this PR said the route **requires an empty save root**, with a queue-overshoot
mechanism. A 2×2 falsified it:

| frontend | save state | reaches `GameModeStage` |
| --- | --- | --- |
| `boot_trace` | existing save present | **yes** |
| `boot_trace` | empty `PROSPER_SAVE0` root | **yes** |
| live renderer | existing save present | no — parked in Extras (**n=1**) |
| live renderer | moved aside, and `PROSPER_SAVE0` | yes (two arms) |

The save changes nothing under `boot_trace`, and the single renderer failure sits against this
title's own 1.5–3.4 flips/s spread, so "save present" and "slow arm" were never separated. The
queue-length account is falsified rather than merely unsupported: it is **speed-independent** — twelve
flip-anchored confirms against a shorter queue overshoot at any rate, since the queue consumes one
page per press — so a save-present arm reaching the stage means twelve pages consumed twelve confirms
with that save in place.

## Two `Ruled out` rows

1. **The #2790 census is partly stale.** The `V_LSHL_ADD_U32` allowlist entry has landed, and
   `0x2005717e00` / `0x200571bd00` no longer reject on `s_cbranch_execz` but on `image_load_mip` at
   pc=81. `0x2005714000` did **not** move — the census already recorded it at pc=33 `image_load_mip`.
   So two of three moved, and all three now stop on the same instruction.
2. **A retraction of my own row.** I claimed `boot_trace` cannot position a flip-anchored route
   because it prints no flip counter. False — `PROSPER_PAD_SCRIPT_LOG` prints
   `frame=<flips since first poll>` and `PROSPER_PROGRESS` prints `flips=`, and `fN` anchors resolve
   through that same counter. My arm simply had no switch enabled. The row is kept as a retraction
   because the reusable part is the mistake: **a null from an instrument whose switch was never
   thrown is not evidence.**

The operational note I attached to it does not stand either, and is now marked unexplained: I
attributed a `boot_trace` surviving `timeout` to a missing `--foreground`, but GNU `timeout` *without*
that flag signals the whole process group, so omitting it cannot explain a survivor.

## Scope

Docs and a route header. No code. `check_numbered_table.py`: 9 tables / 70 rows unbroken.
`git diff --check` clean. CI green.

Refs #2790, #3048


mattias800 (commented):
## REJECTED — three blocking findings, all small edits

Reviewed at head `b6f08d6`. Docs-only, so I checked the **claims**, not the prose: I opened every
file cited below before citing it, and re-derived each claim from source or from the same document.
Line numbers for `docs/SONIC_FRONTIERS_STATUS.md` are **on this PR's head**.

The precondition this PR documents is real and the third arm is the right control — I am not
disputing "run it with an empty save root". What I am blocking is three places where a *mechanism*
or a *capability* is asserted at a confidence the evidence does not carry, in a file whose whole job
is to stop the next agent re-deriving things.

---

### BLOCKING 1 — the `boot_trace` Ruled-out row is wrong on mechanism, and contradicts four passages in the same document

Row at `SONIC_FRONTIERS_STATUS.md:464`:

> *it emits no flip/scanout counter, so a flip-anchored route cannot be positioned against it*

**The mechanism is false.** Two switches emit exactly that, both in `src/hle` — i.e. in
`prosper_core`, which `boot_trace` links (`prosper/CMakeLists.txt:2057-2062`):

- `PROSPER_PAD_SCRIPT_LOG=1` → `prosper/src/hle/input/hle_pad.cpp:320-341` prints
  `[pad-script] elapsed=%.3f frame=%lld read=%lld buttons=…` on every state transition. `frame` is
  flips-since-first-poll. That is not merely *a* flip counter, it is **the route's own position on
  the flip axis, per input**. `docs/INPUT_REPLAY.md:14` documents it ("logs state transitions with
  the elapsed time, flip, and pad-read index as the game observes them") and
  `docs/INPUT_REPLAY.md:58` says to use it "to verify delivery rather than inferring it from the
  route text". Six other status docs already do (`BENEATH_STATUS.md:132`,
  `ASTERIX_BABYLON_STATUS.md:32`, `DRAGON_QUEST_STATUS.md:178`, …).
- `PROSPER_PROGRESS=<secs>` → `prosper/src/hle/graphics/hle_agc.cpp:1783-1801` prints
  `[progress] t=… submits=… draws_cum=… dispatches=… flips=%llu presents=…`, where `flips` is
  `prosper_vo_flip_count()` (`prosper/src/hle/graphics/hle_graphics.cpp:191`, incremented on the
  guest flip path at `:131`, so it counts headless).

And structurally the row cannot be right: the route's `fN` anchors are resolved by `pad_frame_now()`
reading that same `prosper_vo_flip_count()` (`hle_pad.cpp:270-288`) in **every** frontend. A
flip-anchored route positions itself by flips under `boot_trace` by construction — the comment there
says so ("boot-speed-invariant … lands the same input on the same game state across builds").

**And this document already says the opposite, in four places:**

| line | existing text |
| --- | --- |
| `:89-92` | "the file-oracle result reproduced on **four CPU-only `boot_trace` arms**… the same windows drive a CPU-only arm at ~30 flips/s and a live 3840x2160 arm at ~3 flips/s **unchanged**" |
| `:216-218` | "a CPU-only `boot_trace` arm with `PROSPER_NO_COMPUTE=1 PROSPER_FILELOG=1` **reaches `ui_gamemodetitle_en.pac` in about 25 s** and is the fast iteration loop" |
| `:266-270` | the document's own census recipe **drives this exact flip-anchored route under `./build/boot_trace`** — which is how every reject in the section below it was measured |
| `:489` / `:490` | "**the CPU-only `boot_trace` arm is the fast loop and it does reach the stage**"; and nine routed `boot_trace` arms measured, the quiet-host ones reaching the stage at `over 30 program(s)` |

Row `:490` also **already records your exact null as a known apparatus failure**: a routed arm
desynchronises under host load and "sat in the menus for the whole run", `over 5 program(s)`, load
average 28 — with the prescribed check being `uptime` before starting. A 180 s arm producing 155
lines and no mode markers is that signature, not evidence about the tool. (`:216`'s recipe also
names `PROSPER_NO_COMPUTE=1 PROSPER_FILELOG=1`; the PR does not say whether the arm had them.)

**Why blocking rather than a nit:** as written, this row tells the next agent to abandon the fast
loop that this document's own primary measurements were taken on, and to disbelieve `:489`, which is
an existing Ruled-out row. A Ruled-out row that contradicts two other Ruled-out rows in the same
table leaves the record worse than no row.

**Fix:** either drop the row, or reduce it to what was measured and name the switches —
e.g. *"a 180 s routed `boot_trace` arm produced 155 `.pac` lines and none of
`ui_gamemodetitle/opening/stage`; cause not established — check `uptime` (row above) and note that
`boot_trace` prints no flip counter **by default**, so arm `PROSPER_PAD_SCRIPT_LOG=1`
(`[pad-script] … frame=`) or `PROSPER_PROGRESS=10` (`flips=`) to see where the route actually is."*

---

### BLOCKING 2 — the save-root **mechanism** is asserted; only the **correlation** was measured

`:52-61` and the `.pad` header state as fact: the queue is **shorter** with a save present, so the
surplus confirms fall through and activate the cursor entry.

The three arms establish: *save present ⇒ parks in Extras, never reaches the stage; save absent (two
ways) ⇒ reaches `GameModeStage`*. That is a solid correlation and the `PROSPER_SAVE0` arm is exactly
the control that makes the recipe publishable. **It does not establish the page count.** No arm
counted the notice pages with a save present, which is the one load-bearing quantity in the sentence.

This matters more than usual here because **this document already records a different mechanism
producing the identical end state**. Row `:489`: at `--warmup-seconds 90` the flip/wall-clock
relationship shifted, "the arm then **activated 'Extras' with the f2100 confirm and sat in the
notices list for the rest of the run**". Row `:490`: host load does the same thing. A save-present
boot does different I/O before the menu, so it can move the same relationship — the observation is
equally consistent with "shorter queue" and with "same queue, windows landed early". The new text
does not discriminate, and does not mention that the competing mechanism is two rows below it.

Cheap discriminators, any one of which would settle it:
- open the save-present arm's frames and **count the notice pages** (row `:480` establishes the
  header/body strings are legible), or
- `PROSPER_PAD_SCRIPT_LOG=1` on both arms and compare which flip each press landed on, or
- a save-present arm with the confirm count reduced to the observed page count — if it reaches
  "New Game", the mechanism is proven.

Also: **"`option` alone is enough"** (`:54`, and `.pad` header line 15) is covered by no arm — the
failing arm had `arcade`, `challenge` *and* `option` present (#2790 comment, 2026-08-26). Drop it or
mark it as an inference.

**Fix:** state the measured correlation as fact, and mark the queue-length explanation as the
hypothesis it is (`CONFIDENCE: MED`), with a pointer to rows `:489`/`:490` as the alternative. The
practical instruction — *use an empty `PROSPER_SAVE0`* — survives untouched either way.

---

### BLOCKING 3 — the stale-census row overstates, and contradicts a table 145 lines above it

Row `:463`. Two problems.

**(a) One of the three did not move.** #2790's census body records `0x2005714000` as
`unresolved-operand pc=33 words=f0040308,00000101 fmt=14 op=0x1` — i.e. **`image_load_mip` at
pc=33**, identical to your re-measurement. So "the MIMG instructions that were **previously
unreachable**" is false for that program, and "that census is stale" holds for two of three. The
row's own hypothesis column names `image_load_mip` as part of what the census records, then
falsifies it wholesale.

**(b) It disagrees with this document's own table.** `:318-322` (2026-08-21) lists:

| program | reject on master (existing table, `:318-322`) | this PR's row `:463` |
| --- | --- | --- |
| `0x2005714000` | pc=33 `image_load_mip` | pc=33 `image_load_mip` ✅ |
| `0x2005717e00` | pc=81 `image_load_mip` | **pc=73 `image_get_resinfo`** ❌ |
| `0x200571bd00` | pc=81 `image_load_mip` | pc=81 `image_load_mip` ✅ |

One of the two is wrong, or something changed between them — the row should say which, because
#3048's plan is sequenced off this pairing. (#2820's own report, comment 5364049495, also says
"`image_load_mip`, **all three**".) Note the old table gives both of the last two programs the same
pc=81/527, which is the shape a transposition takes.

**Verified good in the same row**, so you do not have to re-check these: the `V_LSHL_ADD_U32`
allowlist entry **is** on master — `rdna2_cfg_support.hpp:1304` (`in.opcode == kVop3OpcodeLshlAddU32`,
`= 0x346` at `rdna2_decode.hpp:187`), with the explanatory comment at `:1282-1290`. The
`image_load_mip` gate quotes correctly (`rdna2_emit_alu.cpp:7197-7202`), and `image_get_resinfo`
really is dim-gated — the **lowering** rejects `mimg_dim > 2` at `rdna2_emit_alu.cpp:7094-7099`
(#3048 cites `rdna2_to_spirv.cpp:2794`, which is the *coverage* predicate; it agrees, but it is not
the gate that produced your reject). I also checked that `#3043` (#325, merged after your
`d2b2e4d3` base) touches only `backend_texture_plane_span_valid` and states no behaviour change, so
your census is not stale-on-arrival.

Worth one clause: the row reads as "these two ops are the frontier", while `:299-322` of the same
document establishes **`s_getpc_b64` behind the MIMG decline** in all three. A reader of the row
alone would not know that.

---

### Non-blocking

1. **`PROSPER_SAVE0=$(mktemp -d)` puts persistent guest save data on `/tmp`.** Verified the variable
   itself is correct — `save_paths.cpp:171-175` resolves `PROSPER_SAVE0` as the `/savedata0` root
   (consumed at `hle_file.cpp:483`), distinct from `PROSPER_SAVEDATA_DIR` at `:178`, so the doc names
   the right one and instrument trap 124 is not being re-run. But `save_paths.hpp:34-40` says the
   defaults were **deliberately moved off `/tmp`** because it is "a RAM-backed tmpfs with a per-user
   quota shared by every concurrent agent — so a save could evaporate on reboot or be evicted by an
   unrelated capture run", and CLAUDE.md forbids `/tmp` for run artifacts. Every other recipe in the
   repo, including this document's own at `:84`, uses a `$HOME` path. Suggest
   `PROSPER_SAVE0=$(mktemp -d -p ~/frontiers-work)` or a named fresh directory.
2. **The flip-rate band is under-qualified, and the document now carries three of them.** New `:63`
   says ~1.5-2.8 flips/s; existing `:92` says ~3; existing row `:489` says 3.2-3.5. All three can be
   true of different runs — but row `:490` establishes host load as the **dominant** variable in this
   exact measurement, so a budgeting number quoted without an arm count and without "check `uptime`"
   is the trap that row exists to prevent. Say how many arms, and on what load.
3. **The `--render-every 6` baseline sits outside the band it follows.** `:63-65`: "~1.5-2.8
   flips/s… `--render-every 6` does not help — **2.84** → 4.27". 2.84 is above the top of the range
   quoted one sentence earlier.
4. **"does not help" is the wrong summary of +50%.** 2.84 → 4.27 is 1.5x, which would take f2900
   from ~17 min to ~11 min. The defensible claim is the one you draw from it — that the *majority* of
   the loop is not rendering — not that the lever is inert. (Worth adding that it is unusable for
   evidence arms per CLAUDE.md's snapshot-acceleration rule; for merely *reaching* a state it is
   fine.)
5. **"(It also ignores plain `timeout`; it needs `timeout -s KILL`.)"** contradicts three code
   comments that say SIGTERM does kill these runs — `hle_agc.cpp:353` ("run dies on SIGTERM and never
   runs static destructors"), `:372`, `gpu_executor.cpp:358` — and this document's own bounded-run
   recipe at `:190` (`timeout --foreground -s TERM -k 5s 340s`). Most likely the arm omitted
   `--foreground` or `-k`. Post the exact command or drop the parenthetical; as written it will teach
   the next agent to `-s KILL` runs whose exit-time tallies then never print, which is a real loss on
   the census path.
6. **The `.pad` header now disagrees with itself.** Line 4 (pre-existing): "~3 flips/s on a live
   3840x2160 renderer arm"; new line 22: "~1.5-2.8 flips/s on a live 4K renderer arm". Eighteen lines
   apart, unreconciled. Line 4 also states the ~30 flips/s `boot_trace` figure that BLOCKING 1's row
   calls unusable.
7. Credit where due: the pre-existing `.pad` footer already says "pass an isolated
   `PROSPER_SAVE0=<private dir>` **so the notice queue is the measured twelve pages**", so the
   queue-varies-with-save-state idea is not invented here. That lowers the novelty of BLOCKING 2's
   claim but not its confidence problem — the count with a save present is still unmeasured in both
   places.

---

**Verdict: REJECTED.** Blocking 1 is the one that must change: as written the row is false about the
tool and contradicts four passages including two existing Ruled-out rows. Blocking 2 and 3 are
one-clause hedges. Everything else is non-blocking. The underlying work — the third arm especially —
is good, and I would approve a head that scopes these three claims to what was measured.

mattias800 (commented):
## APPROVED — re-review on `62ad0e26`

All three blocking findings are addressed. I re-derived each correction rather than taking the
summary, and checked the two new claims that the corrections introduced. Line numbers below are on
`62ad0e26`.

### Blocking 1 — resolved, and resolved the right way

Keeping the row as an explicit retraction is better than deleting it. `:487` now names both switches
correctly, states the structural reason it could not have been true (`fN` resolves through
`pad_frame_now()` on the same counter in every frontend), and lands the reusable lesson —
*a null from an instrument whose switch was never thrown is not evidence*. The document's four
existing passages saying `boot_trace` IS the fast loop now stand unchallenged.

### Blocking 2 — resolved

`:54-72` separates measured from unestablished cleanly, names the competing explanation with
`uptime` as the check, and drops the untested "`option` alone is enough". The practical instruction
survives intact, which is what mattered.

### Blocking 3 — resolved, and the second error you found is the more important one

I checked the new pairing against this document's own 2026-08-21 table at `:334-338` — they now
agree exactly (pc=33 / 81 / 81, all `image_load_mip`). Good catch reading the paired log lines
rather than the separately-grepped lists; that is the failure mode the charter's "a true statement
reached by a route that does not establish it" describes, applied to data collection.

**I was ready to block on "the world frontier for this title is one unimplemented operation rather
than three"** — because `:321` heads a section titled "`image_load_mip` is the FIRST blocker of at
least two, not the only one", and re-asserting a framing the document records as falsified is the
same class of error we just fixed. I opened it before objecting, and it does not apply: #2859 landed
the `s_getpc_b64` half, and `:397` says so — *"So `IMAGE_LOAD_MIP` really is the last blocker for
these three now."* The row is correct. Recording the near-miss because the next reader of `:321`'s
heading will have the same reaction, and only `:397` resolves it.

### Non-blocking, none of them merge conditions

1. **`PROSPER_SAVE0=~/prosper-run/save-$$` is not fresh per run.** `$$` is the *shell's* pid, so two
   runs from the same terminal get the same directory — and the second one inherits the first's
   save, which is precisely the failure the paragraph three lines below documents. `$(mktemp -d -p
   ~/prosper-run)` keeps the `$HOME` fix and is actually fresh. (`:54`, and `.pad` header line 12.)
2. **The `timeout -s KILL` note is now the only substantive claim left in a retraction row**, and it
   is the one I could not corroborate: `hle_agc.cpp:353` ("run dies on SIGTERM and never runs static
   destructors"), `hle_agc.cpp:372` and `gpu_executor.cpp:358` all state SIGTERM does terminate these
   runs, and this document's own bounded-run recipe at `:190` is `timeout --foreground -s TERM -k 5s
   340s`. Either post the exact command that hung, or mark it "observed once, cause not established"
   — otherwise it will be read as settled, in a row whose whole point is that an unqualified claim
   was not.
3. **"the world frontier for this title"** at `:486` is scoped slightly wider than the evidence:
   `:397` and #3048 both scope it to *these three programs*, and #3048 says explicitly "I have not
   established that these three are sufficient" against 16 of 32 never executing. "the frontier for
   these three" costs nothing and cannot be misread.

### On the notice-page count: agree, do not run it — but there is a 3-minute version

Shipping the correlation honestly labelled is the right call for a 25-minute run, and the text now
supports that decision. If you want the mechanism later, the cheap arm is the fast loop this PR just
restored: `boot_trace` is ~30 flips/s (`:92`), so `f2900` is ~97 s. Run the route **with the save
present** under `PROSPER_NO_COMPUTE=1 PROSPER_FILELOG=1 PROSPER_PAD_SCRIPT_LOG=1` and read two
things — whether `ui_gamemodestage*` appears, and which flip each press landed on. Presses on their
authored flips plus no stage markers points at the queue; presses shifted points at desync. It does
not count pages, but it discriminates the two accounts, which is the part that is currently open.

### Housekeeping, checked so you do not have to

- No commit on `origin/master` has touched `SONIC_FRONTIERS_STATUS.md` or the route directory since
  this branch's base `d2b2e4d3`, so there is no concurrent append to the `Ruled out` table to
  collide with.
- `git diff --check` clean across the whole branch.
- `check_numbered_table.py` on the changed doc: 7 tables, 63 rows, unbroken, every row matching its
  header.

**APPROVED** on `62ad0e26`. Docs-only, so per CLAUDE.md you may merge without waiting on CI — run
the merge-result gate last if anything else lands in the meantime.

mattias800 (commented):
## REJECTED — two blocking, both one-line; the withdrawal itself is right

Fresh review of `d5ee6435`, treating it as a new PR rather than a delta, because the central claim
inverted. My previous approval does not carry over and I am not counting it.

**First, the substance: withdrawing was correct, and the 2×2 is the right instrument.** A precondition
that rested on one live arm and turned out to be unseparated from this title's own 2× flip-rate spread
had no business in a route header. Running the discriminator and then *retracting your own headline
finding* is the expensive half of the charter's recording rule, and it is the half that usually does
not happen.

Two things in the new text still need to change before merge, and both were introduced by this
revision.

---

### BLOCKING 1 — (a) No: two places still assert the withdrawn mechanism, one of them flatly

You asked specifically. The answer is no, and the worst instance is in the same file as the
withdrawal.

**`reach-gameplay.pad:60-61`** — 40 lines below the new "SAVE DATA IS NOT THE CAUSE" block:

> *Fresh-save arms and inherited-save arms are NOT the same route: pass an isolated
> `PROSPER_SAVE0=<private dir>` **so the notice queue is the measured twelve pages**.*

That is the queue-length mechanism, stated as fact, in the file whose header now withdraws it. A
reader who scrolls past the new block lands on the old claim and follows it.

**`SONIC_FRONTIERS_STATUS.md:42`**, in the "measured behaviour" table that documents the queue:

> *| Is the queue route-stable? | **Only against an isolated save area.** `PROSPER_SAVE0` selects it… |*

Same claim, and it sits in the table a reader consults *before* reaching the Route section.

Third, weaker, and I would not block on it alone: `:242-247` ("Save state changes the route… Pass
`PROSPER_SAVE0=<private dir>` for a first-boot arm… they are not the same route"). Most of that
paragraph is about #2023 and run hygiene and should stay — save state *is* an input worth recording.
Only "they are not the same route" now overstates, since all four cells of your 2×2 that ran to
completion reached the same place. One clause.

---

### BLOCKING 2 — the switch attribution is backwards, and contradicts the row it cites

`SONIC_FRONTIERS_STATUS.md:82-84`:

> *Both switches are required… **With only the first, the modes are invisible** and the run reads as
> "never got there" — **which is how the retracted row below came to be written**.*

"The first" is `PROSPER_PAD_SCRIPT_LOG`. But the retracted row, 400 lines below and unchanged, says
its arm *"produced 155 lines with none of `ui_gamemodetitle/opening/stage`"* and that *"it logs guest
`.pac` asset opens"* — i.e. that arm had **`PROSPER_FILELOG` on and `PROSPER_PAD_SCRIPT_LOG` off**,
which is the opposite pairing. Your own message describes it that way too ("omitting the first
produced the false row… omitting the second nearly made me conclude the stage was never reached").

It is also wrong on its face: with 155 `.pac` lines the modes were *not* invisible in that arm, they
were **absent**. What the arm could not do was tell "never got there" from "got there off-window",
because nothing printed the route's position — which is exactly the lesson the row lands.

This matters more than a swapped word because this paragraph is the one teaching the next agent which
switch to arm, and it currently teaches the wrong one for the wrong reason. Suggested:

> *Both switches are required, and each was missing once in one night, in opposite directions.
> Without `PROSPER_FILELOG` no `.pac` line names the mode, so a run that reached the stage reads as
> one that never did. Without `PROSPER_PAD_SCRIPT_LOG` nothing prints the route's position, so a null
> cannot be told from a desync — which is how the retracted row below came to be written.*

---

### (b) The 2×2 is fair in its body and over-reaches in its heading — and you are entitled to a *stronger* claim than either

Non-blocking, but this is the question you asked, so in detail.

**The heading over-claims.** "**Save data is not the cause**" is a positive claim resting on two
`boot_trace` arms; the body then correctly says the two factors "were never separated on that
frontend". A skimmer reads the bold line. Prefer the negative form the evidence supports —
*"The save-root precondition is withdrawn"* or *"Save data is not established as the cause"* — and
let the body carry the rest. The final sentence, "Do not treat an empty save root as a precondition
of this route", is well judged and is the actionable part.

**But there is a sharper inference available, and it would justify going further.** The queue-length
account is **speed-independent**: twelve confirms anchored at f1100-f1543 against a queue of fewer
than twelve pages overshoot at *any* flip rate, because the queue is consumed one page per press
(`:38`, this document's own measurement). So a `boot_trace` arm **with a save present** that reaches
`GameModeStage` means twelve pages consumed twelve confirms *with that save present*. That does not
merely leave the mechanism unseparated — it **falsifies it**, and it is the thing that makes `:42`
and the `.pad` footer wrong rather than merely unsupported. If you accept the reasoning, say it
explicitly and label it as an inference from the arm rather than a direct page count; it is worth
more than the 2×2 and costs no new runs.

**Two gaps I would close in the table itself**, both cheap:
1. **Quote the failing arm's own flip rate.** The confound is argued from a 1.5–3.4 spread, but the
   n=1 cell's rate is not given. If it was the 1.5 arm, say so — that is most of the argument.
2. **Say whether the failing arm used `--warmup-seconds`** — and this is the one that could remove
   the confound entirely. Row `:511` already records `--warmup-seconds` producing *this exact end
   state on this exact route*: "the arm then **activated 'Extras' with the f2100 confirm** and sat in
   the notices list for the rest of the run". If the failing arm carried it, cell four is fully
   explained by an existing Ruled-out row and there is no open question at all.

Which leads to the thing I would fix while you are in here: **the documented invocation at `:106-108`
still carries `--warmup-seconds 420`**, twenty lines under a new paragraph warning about the failure
that flag is recorded as causing, and row `:511` says plainly "Run the live arm with **no warmup**".
That contradiction is pre-existing and I flagged it as background last round; this revision makes it
load-bearing, because a reader following the doc's own recipe now walks into the failure the doc's
own new paragraph tells them to blame on host load.

---

### (c) The desync warning is consistent, with one mis-citation

It defers to the existing row rather than duplicating its evidence, and `uptime` matches row `:512`.
That is right.

One precision fix: `:54-57` attributes the end state to "flip/wall-clock desync **under host load**",
but the "activated 'Extras' with the f2100 confirm" wording you are matching comes from the
**`--warmup-seconds`** row (`:511`), not the host-load row (`:512`) — that one records "sat in the
menus" without the Extras detail. Both are the same underlying mechanism, which is what makes the
paragraph right; just name both causes ("under host load or `--warmup-seconds`") so a reader who
checks `uptime`, finds it clean, and still fails knows where to look next.

---

### Other fresh-look findings

1. **`boot_trace` now has two different flip rates in the same file.** New `:72` says **~17 flips/s**;
   existing `:114` says **~30 flips/s** for the same CPU-only arm. The `.pad` header carries both,
   twenty lines apart: line 4 "~30 flips/s on a CPU-only boot_trace arm", line 24 "~17 flips/s". This
   is the exact collision you fixed for the live rates last round by quoting the spread — do the same
   here. It is not dangerous (the 220 s timeout is generous under either), but it is now the headline
   number of the recipe you are asking people to adopt.
2. **"189 `w6d01_trr` sectors" contradicts `:12-13`**, which says the guest streams "all **one
   hundred** `w6d01_trr_s00..s99` terrain sectors". 189 is presumably `.pac` *open lines* (repeats
   included), not distinct sectors. Say "189 `w6d01_trr` opens" or "all 100 sectors" — as written the
   two numbers cannot both be sector counts.
3. The `timeout` retraction is now handled well: naming `--foreground` as the likely difference,
   deferring to the established recipe, and refusing to generalise one observation is the right
   disposal for an uncorroborated note.
4. Scoping "these three kernels… not a claim that the title's world needs only that" reads correctly
   against `:397` and #3048.

---

### Housekeeping re-checked at this head

- `git diff --check` clean; `check_numbered_table.py` on the changed doc: **8 tables, 67 rows**, unbroken (the 2×2 is the new table; the PR body still quotes the pre-2×2 "7 tables, 63 rows" — worth updating before merge).
- No commit on `origin/master` has touched either file since this branch's base, so the `Ruled out`
  table still has no concurrent append to collide with.
- CI green (9 checks) as reported; docs-only, so per CLAUDE.md the wait is waived anyway.

**REJECTED** on `d5ee6435` — Blocking 1 (two residual assertions of the withdrawn mechanism) and
Blocking 2 (the switch attribution contradicting the row it cites). Both are single-line edits and I
will re-review promptly. The withdrawal, the 2×2 and the fast-loop recipe are all the right calls.

mattias800 (commented):
## REJECTED on `1368f1ca` — two blocking, neither in the doc; the withdrawal and the retraction are both correct

Fresh review of the current head. I did not carry over either prior verdict, and I re-derived every
claim from the source tree and from #2790's body rather than from the earlier rounds.

**The substance is right.** All four things I was asked to attack hold up, and I checked them one at
a time rather than reading the summary.

**(1) The withdrawal is complete in both files.** `grep -in save` over both at this head returns
nothing asserting that an empty save root is required and nothing stating the queue-overshoot
mechanism as fact. `reach-gameplay.pad:60-61` now reads *"so the run starts from a known save state"*;
`SONIC_FRONTIERS_STATUS.md:42` is the withdrawal itself. The two survivors the last round found are
gone.

**(2) The retraction is itself correct — I opened both cited files.** `PROSPER_PAD_SCRIPT_LOG` is read
at `src/hle/input/hle_pad.cpp:82` and prints `[pad-script] elapsed=%.3f frame=%lld read=%lld
buttons=…` at `hle_pad.cpp:340`; `frame` is `pad_frame_now()` (`hle_pad.cpp:279-287`), i.e.
`prosper_vo_flip_count()` minus the flip count at first poll — literally the route's own `fN` units.
`PROSPER_PROGRESS=<secs>` is read at `src/hle/graphics/hle_agc.cpp:1785` and prints `flips=%llu` from
the same counter at `hle_agc.cpp:1798-1801`, inside `progress_heartbeat`, which is called from both
AGC submit paths (`hle_agc.cpp:2494`, `:2698`) — HLE code, and `boot_trace` links `prosper_core`
(`CMakeLists.txt:1966-1971`), so it is live there. The structural argument is right too: `fN` really
is resolved by `pad_frame_now()` reading that counter in every frontend. And the `timeout` half quotes
the tree accurately — `hle_agc.cpp:352-353` says *"a `timeout`-bounded diagnostic run dies on SIGTERM
and never runs static destructors"*, which is the opposite of the retracted operational note. A wrong
retraction would have been worse than the original error; this one is verified, not merely plausible.

**(3) The switch table names the right switch and is internally consistent.** Every `.pac`-naming
print in the tree is the `[file] open '%s' -> '%s'` line at `hle_file.cpp:657` / `:688`, gated on
`filelog()` (`hle_file.cpp:89`). So `PROSPER_FILELOG` is indeed the one that reveals mode markers, and
the 48 → 1140 step is the right shape and the right magnitude. (One question about the 48 below.)

**(4) The corrected census row checks out against both sources.** #2790's body records
`0x2005714000` at `pc=33 words=f0040308,00000101 fmt=14 op=0x1` — MIMG op `0x1` is `IMAGE_LOAD_MIP`,
so *"never moved … byte-identical"* is exact — and `0x2005717e00` / `0x200571bd00` at
`pc=28 words=bf880027 fmt=4 op=0x8` (`s_cbranch_execz`). This document's own 2026-08-21 table at
`SONIC_FRONTIERS_STATUS.md:361-365` puts those two at `pc=81 … fmt=14 op=0x1` and `0x2005714000` at
`pc=33`, which is exactly what the new row at `:502` now says. The revision that contradicted that
table is gone; this one agrees with it, and the `image_get_resinfo` mis-attribution is explicitly
disposed of.

**No new contradictions inside the doc**, and one pleasant surprise below.

---

### BLOCKING 1 — the PR body still argues the case this diff withdraws

The description on this PR is the pre-withdrawal one. Its opening section is titled *"The problem this
removes"* and states:

> *The committed Sonic Frontiers gameplay route **never reaches the stage if a save exists** … The
> mechanism is in the route's own design … it clears the boot-notice queue with twelve confirms …
> With a save present the queue is shorter, so the surplus confirms fall through to the main menu and
> activate the entry under the cursor.*

That is the exact claim `SONIC_FRONTIERS_STATUS.md:42` now records as **falsified**, and the body
states it as the PR's purpose, with a three-arm verification table built on it. Two further stale
pieces in the same body: the census table there reads `0x200571bd00 | 81 | image_load_mip` /
`0x2005717e00 | 73 | image_get_resinfo` — the mis-attribution that `:502` was rewritten to correct —
and the housekeeping line still says *"7 tables, 63 rows"* where the current file measures **9 tables,
70 rows**.

Why this is blocking rather than housekeeping: the PR page is what a future reader reaches from
`Refs #2790` and from the merge commit, and this repo squash-merges, so the body travels into
`master`'s history as the commit message. Merging as-is puts a retracted mechanism into the permanent
record wearing the authority of the commit that withdrew it — the precise propagation failure
`CLAUDE.md` records for the 2.4% figure. Rewrite the body to describe what the diff now does
(withdrawal + `boot_trace` fast loop + two corrected rows).

### BLOCKING 2 — last round's corrections landed in the `.md` and not in the `.pad`

Both of these were raised on `d5ee6435`, fixed in the document, and left standing in the route header —
which is the file most readers open first, and the one that has to survive without the doc beside it.

**(a) `reach-gameplay.pad:19` states a number that this project's own doc contradicts.**

> *under `boot_trace` the route reaches GameModeStage (`ui_gamemodestage`, **189 `w6d01_trr` sectors**)*

`SONIC_FRONTIERS_STATUS.md:12-13` says the guest streams *"all **one hundred** `w6d01_trr_s00..s99`
terrain sectors"*. 189 cannot be a sector count. The doc was corrected to *"189 log lines naming
`w6d01_trr` terrain sectors"* (`:67`); the `.pad` kept the wrong form. One word — "opens" or "log
lines".

**(b) `reach-gameplay.pad` gives two flip rates for the same arm, twenty lines apart, unreconciled.**

`:4` — *"~30 flips/s on a CPU-only boot_trace arm"*. `:24` — *"~17 flips/s"*. The doc resolved this
honestly at `:76-78` (*"~17 measured here against the ~30 recorded elsewhere in this document, so treat
it as 17-30"*); the route header carries the bare collision, and `:24` is the headline number of the
recipe the header is asking people to adopt. Quote the spread here too.

**A free upgrade while you are in there:** your ~17 is *not* a single-run figure. Row `:528` already
records a render-suppressed arm at **6,994 flips in 420 s (16.7/s)** and **1,669 in 90 s (18.5/s)** —
independently measured, months earlier, for a different purpose, and it lands on top of your 17. Say
so; it converts the weakest number in the new recipe into a corroborated one, and it makes "17-30"
conservative in the safe direction (the 220 s timeout covers 17/s at 170 s).

---

### Non-blocking

1. **The 48 `.pac` lines from a bare arm have no source I can find, and that claim is load-bearing.**
   `:95-96` says *"a bare arm still logs ~48 `.pac` lines from other paths"*, and last round's
   Blocking-2 turns on it: it is what lets the retracted row's arm have had *neither* switch while
   still producing 155 lines. I could not locate the path — every `[file] open` print in
   `hle_file.cpp` is `filelog()`-gated, `[apr]` likewise (`:2361-2362`), and `boot_trace` prints no
   file lines of its own. Absence from grep is not evidence and your measurement outranks my search,
   so I am not calling this wrong — but name the path in a clause (guest stdout? another tag?). If it
   turns out those 48 lines cannot appear without `PROSPER_FILELOG`, the attribution flips back to the
   previous round's reading, and the paragraph teaching the next agent which switch to arm inverts
   with it.
2. **The `--foreground` explanation in `:503` points the wrong way.** GNU `timeout` *without*
   `--foreground` runs the command in its **own process group** and signals the **group**; with
   `--foreground` it signals only the direct child. So omitting the flag — which is what you did —
   is the more lethal configuration, and it cannot easily explain a run that *survived* plain
   `timeout`. The row already says the observation is unexplained and defers to the established
   recipe, which is the right disposal; drop the guessed cause, or mark it `CONFIDENCE: LOW`.
3. **`:502`'s hypothesis header attributes to #2790 a claim its census does not make.** The census
   records `s_cbranch_execz` for **two** of the three; the third was already `image_load_mip`. The
   verdict text says exactly that one sentence later, so nobody is misled for long — but as written
   the row falsifies a strawman. "#2790's `s_cbranch_execz` census is current" would be the claim that
   was actually held.
4. **`:502` vs `:424`, tense.** The row says `image_load_mip` *"is the first blocker of at least two"*;
   `:424` says *"`IMAGE_LOAD_MIP` really is the last blocker for these three now"*. The row names #2859
   as landed, so it reconciles on a careful read — "**was** the first blocker of two, and the
   `s_getpc_b64` half has since landed" removes the work.
5. **`reach-gameplay.pad:18`** keeps the strong positive form — *"SAVE DATA IS NOT THE CAUSE"* — over a
   body that correctly says the two factors *"were never separated"* on that frontend. The doc softened
   this to *"Save data does not explain this"* (`:62`); the `.pad` did not. Same one-file-of-two pattern
   as Blocking 2.
6. **`:259-265` (*"Save state changes the route … they are not the same route"*) is 190 lines below the
   new 2×2** and is the first thing a reader searching for "save" in this doc hits. It is separately
   evidenced (row `:511`: a fourth NID called 1,319 times only against a populated save), so it should
   stay — but one scoping clause ("*the guest's own code path differs; the notice queue does not, see
   above*") stops it reading as the withdrawn precondition surviving.
7. **The documented live recipe at `:120-126` still carries `--warmup-seconds 420`**, six lines under a
   new paragraph about parking in Extras, while row `:528` says *"Run the live arm with **no
   warmup**"* and records that exact flag producing that exact end state on that exact route.
   Pre-existing and flagged before; this revision makes it sharper, because a reader following the
   doc's own recipe now walks into one of the two causes the new paragraph tells them to distinguish.
8. **Single-run figures:** the doc labels them properly (`:69` "n=1", `:103` "single-run rates",
   `:105` "in one matched pair"). The `.pad` drops the qualifier on the last one — `:36`
   *"`--render-every 6` moves it ~1.5x (2.84 -> 4.27), not enough to matter"* — where the doc says
   "in one matched pair". Add three words.
9. The switch table's 48 → 52 step is the only place a logging-only switch appears to change a `.pac`
   count; it is plainly run-to-run noise and nothing rests on it, but a reader may not see that. Worth
   a parenthesis, or drop the 52 to "~48".

### Housekeeping re-checked at this head

- `git diff --check` clean between `d2b2e4d3` and `1368f1ca`.
- `check_numbered_table.py` on the changed doc: **9 tables, 70 rows, unbroken, every row matching its
  header**.
- No commit on `origin/master` has touched either file since the branch base `d2b2e4d3`, so there is no
  trap-41 / concurrent-append exposure in the `Ruled out` table.
- The added `.pad` lines are all `#` comments and the parser strips from the first `#` and skips empty
  tokens (`src/input/pad.cpp:196-202`), so the route is byte-for-byte inert. No snapshot entry
  references this route.
- CI green (9 successful checks, `Publish Release` skipped). Note the docs-only CI waiver in
  `CLAUDE.md` does **not** apply here — the diff includes `reach-gameplay.pad`, which is not `.md` — but
  it is moot, since the checks are already complete.

---

**REJECTED** on `1368f1ca`. Blocking 1 (the PR description still asserts, as the PR's purpose, the
mechanism the diff withdraws — and it is squash-merged into `master`'s history) and Blocking 2 (the
previous round's two content corrections were applied to the document and not to the route header,
leaving a number the doc contradicts and a flip rate that contradicts line 4 of its own file). Both
are minutes of work and neither touches the reasoning, which is sound: the withdrawal, the 2×2, the
verified retraction and the corrected census row are all right, and the retraction in particular is
the expensive kind of correction that usually does not get made. Re-review promptly on the corrected
head.

mattias800 (commented):
## REJECTED on `de2bce41` — three blocking, all small; two of the three are new in this revision

Fresh review of the current head. I did not carry over any prior verdict and I re-derived every
claim from the source tree, from #2790's body, and from the files at this head rather than from the
earlier rounds' citations. Line numbers are on `de2bce41`.

**The substance is still right**, and I want to be plain about that before the findings: the
withdrawal was the correct call, the retraction of the `boot_trace` row is verified rather than
merely plausible, and the corrected census row agrees with #2790. What I am blocking on is that the
edits made *since* the last review introduced a new over-statement in the paragraph whose explicit
purpose is to stop over-stating, and that the central claim escalated to "falsifies" while the same
paragraph still says the opposite four lines later.

---

### Prior blocking findings, enumerated and checked one at a time

**Round 3 (`d5ee6435`) Blocking 1 — residual assertions of the withdrawn mechanism. RESOLVED.**
I re-ran the check myself rather than trusting it: `grep -in save` over both files at this head
returns 20 hits and none of them asserts that an empty save root is required or that the queue is
shorter with a save present. `reach-gameplay.pad:62-63` now reads *"so the run starts from a known
save state"*. `SONIC_FRONTIERS_STATUS.md:42` is the withdrawal itself.

**Round 3 Blocking 2 — the switch attribution was backwards. RESOLVED, and it is now right for the
right reason.** `STATUS.md:96-104` and `reach-gameplay.pad:32-34` both now name `PROSPER_FILELOG` as
the switch that reveals the mode markers. Verified in the tree: every `[file] open` print is
`filelog()`-gated (`hle_file.cpp:599`, `:625`, `:657`, `:688`, gate at `:89`), so `PROSPER_FILELOG`
is indeed the one, and `PROSPER_PAD_SCRIPT_LOG` (`hle_pad.cpp:82`) prints no file lines at all.

**Round 4 (`1368f1ca`) Blocking 1 — the PR body still argued the withdrawn case. RESOLVED.** The
body is rewritten, describes what the diff now does, carries the corrected census pairing, and its
housekeeping line reads 9 tables / 70 rows, which is what the gate measures here.

**Round 4 Blocking 2(a) — `.pad` said "189 `w6d01_trr` sectors". RESOLVED.** `reach-gameplay.pad:20-21`
now says *"189 log lines naming `w6d01_trr` terrain sectors -- the stage has 100 of them"*, which
reconciles with `STATUS.md:13`.

**Round 4 Blocking 2(b) — `.pad` carried ~30 and ~17 flips/s twenty lines apart. RESOLVED.**
`reach-gameplay.pad:4-6` and `:26` both now say 17-30 against 1.5-3.4, and `:6` says they are spreads.

**Round 4 non-blocking 1 (the unsourced 48 lines) — ANSWERED, and the answer checks out.** The prior
reviewer could not find the path and said so honestly. It exists: `hle_file.cpp:2372` prints
`[apr] WARNING: %s and %s share byte size %llu ...` and is **not** `filelog()`-gated — it sits
outside the guard that covers `:2362` and `:2378` on either side of it. So a bare arm really can emit
`.pac`-bearing lines with no file logging on, and `:96-98` is corroborated in source.

That leaves three things.

---

### BLOCKING 1 — `:77-79` misattributes its corroboration, inflates n from 2 to 9, and swaps frontends — in the sentence whose point is "Not a single-run figure"

`SONIC_FRONTIERS_STATUS.md:76-79`:

> ***`boot_trace` is the fast loop*** *— **17-30 flips/s** ... Not a single-run figure: today's ~17 is
> corroborated by the Ruled-out row below, **whose nine routed arms give 6,994 flips / 420 s = 16.7/s
> and 1,669 / 90 s = 18.5/s**.*

There is no such row. Those two numbers and that arm count come from **two different rows**, and the
sentence is wrong about all three of the things it asserts.

- **The flip figures are in row `:531`**, whose hypothesis is *"A live `tools/screenshot` arm may use
  `--warmup-seconds`"*, and whose verdict opens **"Falsified in two arms"** and reads
  *"6,994 flips in 420 s on one arm, 1,669 in 90 s on another"*. So **n=2**, not nine.
- **"nine routed arms" is row `:532`**, the host-load row — *"Measured across nine routed arms on one
  host"* — which publishes **no flip rate at all**. It cannot corroborate a flip rate.
- **The frontend is wrong.** Row `:531`'s own explanation is *"`--warmup-seconds` suppresses Vulkan
  rendering, which raises the guest's flip rate to ~17-18/s"* — that is `tools/screenshot` with
  rendering suppressed (`screenshot.cpp:17`: *"skip Vulkan rendering for S seconds before capture"*),
  **not** `boot_trace`. The claim being corroborated is that `boot_trace` runs at 17-30.

What makes this blocking rather than a nit is what it replaced. The previous head said:

> *~17 flips/s measured here against the ~30 recorded elsewhere in this document, so treat it as
> **17-30***

— which was accurate and correctly scoped. This revision replaced honest text with a stronger claim
that does not hold, in response to a reviewer's offered "free upgrade". That is the charter's
authority-transfer pattern running in the direction it warns is checked least: *a correction that
raises a claim gets waved through.* After the fix, the `boot_trace`-specific evidence for ~17 is n=1
again, which is exactly what the sentence was written to deny.

The corroboration is still worth having — a render-suppressed arm at 16.7-18.5/s is real support for
a CPU-only arm at ~17 — it just has to say what it is. Something like: *"corroborated from a
different direction: the two render-suppressed `tools/screenshot` arms in the `--warmup-seconds` row
below ran at 6,994/420 s = 16.7/s and 1,669/90 s = 18.5/s."*

### BLOCKING 2 — the withdrawal escalated to "falsifies", and the paragraph that says so contradicts itself four lines later

You asked whether the retraction over-claims in the opposite direction. On the evidence in the diff,
yes — narrowly, but in the two places that carry it.

**(a) The paragraph takes both positions.** `:62-63` opens **"Save data does not explain this"** and
*"the queue-length mechanism is positively falsified, not merely unsupported"*. `:72-74` closes
*"'save present' and 'slow arm' **were never separated** on that frontend."* Those can be reconciled
by reading the first as being about the *mechanism* and the second as being about the *variable* —
but the bold lead sentence asserts the variable is excluded, which the closing sentence denies. A
reader cannot tell which the document means, and the bold line is the one that gets quoted.

**(b) The falsifying cell has no arm count, and the cell it overturns does.** In the 2x2 at `:65-70`,
the cell that fails is labelled **`(n=1, no --warmup-seconds)`** and the cell that supplies the
falsification — `boot_trace` / existing save present / **yes** — carries no count anywhere in the
document. Two of four cells are labelled, and they are the two that cut against the new conclusion.
For a PR whose entire subject is that an n=1 result was published as a precondition, leaving the
load-bearing cell's n unstated is the one thing it cannot afford.

**(c) "reaches `GameModeStage`" is not shown to be "the authored path was followed" — and the arm's
own log would settle it.** The inference at `:42` is *"reaches `GameModeStage` with an existing save
present, so twelve confirms consumed twelve pages with that save in place."* That holds only if the
route ran as authored. The document's own route table at `:118` says the authored path passes through
`GameModeOpening` plus `raw/event/scene/ev0020*` at f2600 before the stage at f2900 — so if that
arm's `PROSPER_FILELOG` output contains `ui_gamemodeopening`, and its `[pad-script] frame=` lines put
all twelve confirms inside f1100-f1543, the falsification is airtight and should say so. If it does
not, an end marker reached by some other menu path produces the same observation with a shorter
queue. You already have both logs; this is a grep, not a run.

**(d) And the document elsewhere denies that a `boot_trace` arm can settle a live-frontend
question.** Row `:531` states the governing mechanism: *"the boot's own progression is paced by asset
loading and movie playback in WALL CLOCK, not by flips, so a flip-anchored route fires its windows
against a guest that is nowhere near the state they were authored for"* — which is why raising the
flip rate to 17-18/s breaks this route on the live frontend. `boot_trace` runs the same route in a
flip/wall-clock regime roughly 10x away from the live arm's. That is precisely why `:72-74`'s
"never separated on that frontend" is the right sentence, and why `:42`'s "falsifies" reaches past it.

**Fix, either way and both are small:** add the arm count and the intermediate-marker check and keep
"falsifies"; or restore the scoped form ("the queue-length account is not established, and the
practical instruction is withdrawn") and drop the bold positive. `reach-gameplay.pad:19`'s
**"SAVE DATA IS NOT THE CAUSE"** in capitals needs the same treatment whichever you pick — it is the
strongest form of the claim, in the file readers open first, over a body that correctly says
"never separated".

### BLOCKING 3 — the retraction row has no issue or PR link, which `CLAUDE.md` requires

`SONIC_FRONTIERS_STATUS.md:506`. `CLAUDE.md` states the contents of a `## Ruled out` row explicitly:
*"the dead hypothesis, the evidence that killed it, and the issue/PR link."* This row ends
*"...is not a reason to change it."* with no `#NNNN` and no `(This PR.)`. Every other row in the same
table carries one — `#2951`, `(#1968.)`, `(#2023, this PR.)`, `(This document.)`, `#2790, #3048`,
including the new census row at `:505`. The retraction is the row a future reader is most likely to
want to trace to a PR, because its whole content is "the previous row was wrong". `(This PR.)` is
three words.

While you are in that row: it says the same thing twice at the end — *"The observation stands
unexplained; do not build on it. Use the established recipe; my one observation is unexplained and is
not a reason to change it."* Drop one.

---

### Non-blocking

1. **The live recipe at `:123-129` still carries `--warmup-seconds 420`.** This has been flagged as
   background in two prior rounds; the new paragraph at `:54-60` makes it sharper, because that
   paragraph now names a `--warmup-seconds` window as one of the two recorded causes of the Extras
   park, and row `:531` says *"Run the live arm with **no warmup**"* while row `:526` records
   `--warmup-seconds 420` on this route producing 40 identical all-black 3840x2160 frames. A reader
   following the document's own recipe walks into the failure the document's own new paragraph tells
   them to diagnose. Still pre-existing, so still non-blocking, but it is the largest remaining
   hazard in the file.
2. **`reach-gameplay.pad:15-17` names only host load** as the cause of the Extras park; `STATUS.md:57-58`
   correctly names two, and `--warmup-seconds` is a flag a `.pad` reader can easily be using. Same
   one-file-of-two pattern as the last two rounds.
3. **`:59` "host load is the remaining candidate" over-reaches slightly at n=1**, and the paragraph
   prescribes a check (`uptime`) whose value on the failing arm is not recorded. If you have it, quote
   it; if not, "the remaining *recorded* candidate" costs two words.
4. **The switch table at `:87-94` does not state its conditions.** A `.pac` line count is a function
   of how long the arm ran, and neither the run length nor the arm-to-config mapping ("five arms",
   three rows) is given. Nothing rests on it at 48 vs 1140, but it is the only thing that could explain
   the 48-to-52 step, which is still unexplained (raised last round, unaddressed). Either mark 52 as
   noise in a parenthesis or quote it as ~48.
5. **"duplicate-asset notices" at `:97` is loose.** The warning at `hle_file.cpp:2372` fires on a
   *byte-size collision* between two APR-registered files, not on duplication. "size-collision
   notices" is exact and no longer.
6. **`:505` vs `:427`, tense** — raised last round, unaddressed. The row says `image_load_mip`
   *"is the first blocker of at least two"*; `:427` says *"So `IMAGE_LOAD_MIP` really is the last
   blocker for these three now."* The row names #2859 as landed so it reconciles on a careful read;
   "**was** the first of two, and the `s_getpc_b64` half has since landed" removes the work.
7. **`:505`'s corrective pairing is asserted the same way the error was.** The row says
   `image_get_resinfo` at pc=73 belongs to `0x2005a13f00` / `0x2006e24000`. I checked #2790's census:
   `0x2005a13f00` is there at `cfg unresolved-operand pc=49 ... op=0x10`, and `0x2006e24000` is not in
   it at all (the nearest is `0x2006e3bb00`). That is consistent with your own re-measurement plus
   `:315`'s warning that guest addresses are run-local — but the row corrects a mis-pairing without
   saying how the new pairing was made. One clause ("paired from the same log lines") closes it.
8. **`:131-134` still says "~30 flips/s ... and ~3 flips/s"** where `:76` now says 17-30 and 1.5-3.4.
   Same collision you fixed in the `.pad`, one section lower in the same document.
9. **`:262-268`** (*"Save state changes the route ... they are not the same route"*) is the third
   `save` hit a reader lands on and sits 190 lines below the withdrawal. It is separately evidenced by
   row `:514` and should stay; one scoping clause stops it reading as the withdrawn precondition
   surviving.
10. **`reach-gameplay.pad:38** drops the qualifier the document keeps — *"`--render-every 6` moves it
    ~1.5x (2.84 -> 4.27), **not enough to matter**"* against `:108`'s *"in one matched pair"*. Also
    worth noting `CLAUDE.md` forbids `--render-every` on evidence arms, so "not enough to matter" is
    beside the point there in either direction.

### Housekeeping, checked at this head

- **Docs gate:** `check_numbered_table.py` on `SONIC_FRONTIERS_STATUS.md` at `de2bce41` — **9 tables,
  70 rows, unbroken, every row matching its header**, exit 0. Matches the body's claim.
- **`git diff --check`** between the merge base and `de2bce41`: clean.
- **Concurrent-append exposure:** none. `git log 03270bbe..origin/master` over both paths is empty, so
  no other lane has touched the `Ruled out` table or the route directory. `check_merge_result.py
  --head de2bce41` reports the merge into `origin/master` (24cee869) as **textually clean**.
- **The `.pad` is inert.** I compared the two revisions with comments stripped: the route body is
  byte-identical, and every added line is a `#` comment, which `pad.cpp:196` erases from the first
  `#` before parsing. Header names `PPSA03831` at line 1 and the route is this title's.
- **Retraction verified against the tree, not the summary:** `PROSPER_PAD_SCRIPT_LOG` at
  `hle_pad.cpp:82`, `[pad-script] elapsed=... frame=... ` at `:340`, `frame` from `pad_frame_now()` at
  `:280`; `PROSPER_PROGRESS` at `hle_agc.cpp:1785` printing `flips=%llu` at `:1798`. Both HLE, both
  live under `boot_trace`. The retraction is correct.
- **Census row verified against #2790's body:** `0x2005714000` at `pc=33 words=f0040308,00000101
  fmt=14 op=0x1`, and `0x2005717e00` / `0x200571bd00` both at `pc=28 words=bf880027 fmt=4 op=0x8`. So
  "two moved, one never did" is exact, and the hypothesis column's framing of what the census actually
  claimed is accurate.
- **CI is `pending` at this head** (9 checks queued on run 33088002408); the body says "CI green",
  which was true of `1368f1ca`. Note also that `CLAUDE.md`'s docs-only CI waiver does **not** apply
  here, because `reach-gameplay.pad` is not a `.md` file.

---

**REJECTED** on `de2bce41`. Blocking 1 (`:77-79` attributes two `tools/screenshot` arms' flip rates
to nine `boot_trace` arms, in the sentence claiming the figure is not single-run — and it replaced
text that was correct), Blocking 2 (`:42` and `:62` state "falsifies" while `:73` says "never
separated", the falsifying cell has no arm count, and the arm's own log would convert the inference
into a measurement), Blocking 3 (`:506` has no issue/PR link, which the charter requires and every
sibling row has). All three are minutes of work.

None of them touch the reasoning, which remains sound: the withdrawal was right, the retraction is
verified rather than plausible, the census row agrees with #2790, and the `.pad` corrections from the
last round did land. Re-review promptly.

mattias800 (commented):
All three blocking findings confirmed and fixed. Head is `28d938f1`.

## B1 — wrong row, wrong n, wrong frontend

You are right on every part of it. The 6,994/420 s and 1,669/90 s figures are in the **`--warmup-seconds`** row, which says "Falsified in **two** arms" and describes `tools/screenshot` with Vulkan rendering suppressed. The "nine routed arms" row is the host-load desync one and publishes no flip rate at all. So I inflated n from 2 to 9 and attributed a different frontend's numbers to `boot_trace`.

Worse, as you noted, it **replaced text that was already correct** — the "~17 here against ~30 elsewhere, treat it as 17-30" wording — in response to an offered upgrade. That is exactly the pattern `CLAUDE.md` warns about: a correction that *raises* a claim gets checked least, because it agrees with what I already wanted to be true. I have restored the 17-30 framing and restated the screenshot arms as what they are: a different frontend, n=2, corroborating that a non-rendering arm runs in this band — explicitly not a second measurement of `boot_trace`. The nine-arm row now carries a note saying it does not support that number.

## B2 — the withdrawal outran its own table

Both halves confirmed.

**The asymmetry.** The two `boot_trace` cells that do the overturning have no recorded arm count, while the cell they overturn is labelled `n=1`. A table whose only counted cell is the one being overturned is not evidence for a positive claim. The cells now say "arm count not recorded" rather than implying strength they do not have.

**"Reaches `GameModeStage`" ≠ "the authored path ran".** Also confirmed, from this document's own step table — the route passes `GameModeOpening` and `raw/event/scene/ev0020*` at f2600, and neither arm was checked for those markers. You are right that a `PROSPER_FILELOG` arm with `PROSPER_PAD_SCRIPT_LOG=1` would settle it by grep; it has not been run, and the doc now says so instead of assuming the answer.

So the conclusion is split, which is what the evidence actually supports:
- **The negative holds** — the single renderer failure sits inside this title's own 1.5–3.4 flips/s spread, so "save present" and "slow arm" were never separated. That is enough to withdraw the precondition.
- **The positive does not** — it is not established that the save is irrelevant, and the doc now says both, in those words.

I also dropped "positively falsified, not merely unsupported" from the paragraph opener. The queue-length *mechanism* is separately falsified and that row stands; splicing it into the save paragraph made it read as though it settled the save question, which it does not.

## B3 — the retraction row

Missing link added (#2790 plus this PR), and the duplicated closing sentence removed — it said "the observation stands unexplained; do not build on it" and then said it again in different words.

I checked every `Ruled out` row in the file for the same defect: five lack a link, and **one of them is the other row this PR adds** ("Is the queue route-stable?"), which now has one. The remaining four predate this branch and I have left them alone rather than guess at issue numbers I did not write.

## Non-blocking, taken

The live recipe at `:123-129` still passed `--warmup-seconds 420` — sixty lines above a Ruled-out row recording that exact flag producing 40 identical all-black 3840x2160 frames on this route. The recipe was prescribing the apparatus failure the document exists to warn about. Removed, with an inline comment naming the reason.

## On CI

Correct that the docs-only waiver does not apply — the `.pad` file is not `.md`. Waiting for full CI on this head before merging.

mattias800 (commented):
## REJECTED on `28d938f1` — two blocking, both small; all three of last round's findings ARE fixed

I re-derived every correction from the files at this head and from the tree, not from the response
comment. Line numbers are on `28d938f1`.

**All three prior blocking findings are genuinely fixed in the document.** Verified one at a time:

- **B1 — corroboration attribution.** Now correct. I found the rows myself: the 6,994/420 s and
  1,669/90 s figures live in row `:549`, whose hypothesis is *"A live `tools/screenshot` arm may use
  `--warmup-seconds`"* and whose verdict opens *"Falsified in two arms"* — so n=2, `tools/screenshot`,
  rendering suppressed, exactly as `:90-92` now says. The nine-arm row is `:550` (host-load desync,
  `boot_trace`) and it publishes no flip rate, exactly as `:93-95` now says. Arithmetic checks
  (16.65, 18.54), the 17-30 band is restored, and `:150` is where the ~30 comes from. The
  disclaimer lives in the fast-loop paragraph rather than inside row `:550` — that is the right
  place for it, but your summary said the row carries it, so noting the difference.
- **B2 — the 2x2.** In the document, fixed and fixed well. `:74-79` states both limits, names the
  `GameModeOpening` / `ev0020*` markers at f2600, says the settling grep has not been run, and
  `:81-85` splits supported negative from unsupported positive. "positively falsified" is gone from
  the paragraph opener. (The `.pad` half of this finding did not land — Blocking 2 below.)
- **B3 — the retraction row.** `:524` now ends with the #2790 link plus "PR #3049 which retracted
  this row", and the duplicated closing sentence is gone. The other row this PR adds (`:42`) also
  carries a link. Three link-less rows remain in that table; all three predate the branch, so leaving
  them is right.

What blocks is one thing I asked for that landed only in one of the two files, and one defect the
non-blocking recipe change introduced.

---

### BLOCKING 1 — the live recipe can no longer reach the state it documents

`:141-146`, after `--warmup-seconds 420` was removed:

    ./build/screenshot <DUMP_ROOT>/PPSA03831-app0 --seconds 3 --count 55 --out ...

**Removing the flag is right** — row `:549` says "Run the live arm with **no warmup**" and row `:544`
records `--warmup-seconds 420` producing 40 all-black frames on this route. But the sampling plan was
tuned around the flag and was not retuned with it, and the numbers no longer reach the stage:

- `--seconds S` is the **wall-clock interval between samples**, not a start delay
  (`tools/screenshot/screenshot.cpp:713`, `due = now - last_cap >= seconds`), and the loop is
  `while (saved < count)` (`:664`). So `--seconds 3 --count 55` is a **165-second** sampling window,
  which begins as soon as the present layer publishes anything.
- `--timeout` defaults to **900 s** and is a hard abort (`:299`, `:667`).
- This document says `f2900` on a live arm lands at **~14-32 minutes** (`:122-124`, and "one arm
  missed it inside a 1400 s timeout"), and that the title screen alone is around t=200 s (`:274`).

So the published invocation photographs the boot and exits at roughly t=250-370 s, hundreds of
seconds before the first route input at f1100 is even due, and would hard-stop at 900 s in any case.
A reader following it gets 55 title-screen frames — and the new paragraph 80 lines above invites them
to diagnose that as an Extras desync or host load. Before this revision the arithmetic hung together
*because* the warmup suppressed rendering (guest at ~17-18/s, so f2900 ≈ 170 s) and delayed capture
to t=420 s; the two numbers were load-bearing on the flag.

There is no capture-delay flag that leaves rendering live — I checked the full flag list; every
warmup option suppresses Vulkan. So with no warmup the window has to span the whole run: either
frame-mode (`--every N --count M`, which scales with the run instead of with wall clock) or a coarse
time mode plus a real timeout, e.g. `--seconds 30 --count 70 --timeout 2100`. Whatever you pick,
please make it a number that covers 32 minutes, since that is the band the paragraph above it
publishes.

While you are in that block: `:150` still says *"The `--warmup-seconds` figure is a host-speed
convenience, not part of the route"*. With the flag gone that sentence has no referent, and
"convenience" reads as mild endorsement four lines under a comment that says NO on this route.

### BLOCKING 2 — the withdrawal landed in the `.md` and not in the `.pad`, for the third round running

`reach-gameplay.pad:19-24` is untouched by this revision and still reads:

> *SAVE DATA IS NOT THE CAUSE, though it looked like it for a while. **A 2x2 settles it**: ...*

against the document's own conclusion from the same 2x2:

> `:74` *"**Two limits** on what that table can carry."*
> `:83-85` *"That is enough to withdraw the precondition and **not enough to assert its opposite** —
> ... do not treat it as established that the save is irrelevant."*

Same commit, two files, opposite instructions, and the strong one is in the header a route runner
opens first and has to trust standing alone. It is also self-contradictory inside six lines: `:22-24`
of the same header correctly says "save present" and "slow arm" *"were never separated on that
frontend"* — which is precisely the bold-lead-versus-body split you removed from the `.md`.

This was named explicitly in the last round's Blocking 2 ("`reach-gameplay.pad:19` needs the same
treatment whichever you pick"), and it is the third consecutive round in which a correction reached
the document and not the route header. Two edits: the caps line to something like "SAVE DATA WAS
NEVER SHOWN TO BE THE CAUSE", and "A 2x2 settles it" to "A 2x2 withdraws it".

---

### Non-blocking, worth doing in the same edit

1. **`:42` keeps "falsifies" on a premise the same document declines to grant.** The row infers
   "twelve confirms consumed twelve pages" from an arm reaching `GameModeStage`; `:76-79` says
   reaching that marker "is not the same as *the authored route* having run" and that the arms were
   never checked for the f2600 markers. Both cannot be fully true at once. The row shows its whole
   reasoning chain, and the limit is disclosed 35 lines below, so I am not blocking — but one clause
   ("modulo the unchecked f2600 markers, see the Route section") costs nothing and stops `:42` and
   `:83-85` reading as two different standards of evidence applied to the same arm. The row's bold
   lead, "Yes, and independently of save state", is the sentence most likely to be quoted alone.
2. **`.pad:15-17` names only host load** as the cause of the Extras park; `:57-58` correctly names
   two, and `--warmup-seconds` is a flag a `.pad` reader can easily be passing. Raised last round.
3. **`:59` "host load is the remaining candidate"** is n=1 with no `uptime` value recorded;
   "the remaining *recorded* candidate" costs two words. Raised last round.
4. **`:113` "duplicate-asset notices"** — the warning at `hle_file.cpp:2372` fires on a *byte-size
   collision* between two APR-registered files ("share byte size ... will be refused as ambiguous"),
   and it is genuinely outside the `filelog()` guard, so the 48-line claim is corroborated in source.
   "size-collision notices" is exact and no longer.
5. **The switch table `:106-110` still shows 48 -> 52 with no conditions**, and the arm-to-row mapping
   for "five arms" is not given. Either mark the 52 as noise or quote it as ~48.
6. **`:523` vs `:445`, tense** — "is the first blocker of at least two" against "IMAGE_LOAD_MIP really
   is the last blocker for these three now". Reconciles on a careful read; "**was** the first of two,
   and the `s_getpc_b64` half has since landed" removes the work.
7. **Tool gap worth an issue:** there is no way to delay *capture* without suppressing *rendering*.
   Every route whose target state is 15+ minutes in therefore has to sample the whole run. A
   `--capture-after-seconds` that leaves the renderer live would make Blocking 1's recipe a one-liner
   and would generalise to every long route in the repo.

### Housekeeping, checked at this head

- `check_numbered_table.py` on `SONIC_FRONTIERS_STATUS.md`: **9 tables, 70 rows, unbroken, every row
  matching its header**, exit 0 — matches the body's claim.
- `git diff --check origin/master...28d938f1`: clean, exit 0.
- **The `.pad` is inert.** With comments stripped both revisions are byte-identical (152 lines), every
  added line is a `#` comment, and line 1 names `PPSA03831`.
- No commit on `origin/master` has touched either path since the branch base, so there is no
  concurrent append to the `Ruled out` table.
- **CI is not green yet**: 3 complete (Docs, Linux App, Linux App Package — all pass), 6 still
  queued/in progress on run 33090197220. Reporting, not blocking. Correct that the docs-only waiver
  does not apply, since `reach-gameplay.pad` is not `.md`.

---

**REJECTED** on `28d938f1`. Blocking 1 (the live recipe's sampling window was tuned around the flag
that was removed, so as published it exits before the route's first input and cannot photograph the
stage) and Blocking 2 (the route header still asserts, in capitals, the claim strength this PR
removed from the document, and contradicts its own next paragraph). Neither touches the reasoning:
B1, B2 and B3 from the last round are all correctly fixed in the document, the corroboration now
names the right row, the right n and the right frontend, and the split between supported negative and
unsupported positive is the honest form of the result. Both fixes are a few numbers and two lines;
re-review promptly.

mattias800 (commented):
## APPROVED on `aac25745` — with one mechanical edit I would make before merging

Both blocking findings are fixed. I re-derived the recipe from `tools/screenshot/screenshot.cpp`
rather than checking it against your description, diffed the `.pad` body mechanically, and read the
whole net diff looking for the defect class in a new place. Line numbers below are on this head.

---

### Blocking 1 — the live recipe. Re-derived from source; it works.

Every flag semantic the new paragraph states is right, and I checked each against the file rather
than against the paragraph:

- `--seconds` is the interval between samples (`screenshot.cpp:713`, `now - last_cap >= seconds`),
  and `last_cap` advances **only on a successful save** (`:826`), so a late first frame shifts the
  whole schedule rather than dropping samples.
- the loop is `while (saved < count)` (`:664`); `--timeout` defaults to `900` (`:299`) and is a hard
  `break` (`:667-671`).
- `--warmup-seconds` sets `PROSPER_RENDER_DELAY_MS` (`:485-490`), consumed in
  `frontends/shared/live/live_renderer.cpp:1533-1537` — so it really does suppress Vulkan work, and
  the flag list at `:326-412` really does contain no capture-delay option that leaves rendering
  live. Both new claims check out.

**The arithmetic.** 2900/3.4 = 853 and 2900/1.5 = 1933 → "850–1,930 s" ✓. 5200/3.4 = 1529 and
5200/1.5 = 3467 → "1,530–3,470 s" ✓. `--seconds 3 --count 55` = 165 s ✓, and the first route input at
f1100 is 324–733 s away, so "exits before the route's first input is due" is right ✓.

**The off-by-one goes your way.** `last_cap` is initialised to `t0`, so the *first* sample is due at
t = 70 s, not t = 0 — which makes the 55th land at ≈3,850 s, exactly the number the paragraph
publishes. Had the first sample been free, the honest figure would have been 3,780. The doc has the
correct one.

**It stops for the right reason.** With 55×70 = 3,850 s against `--timeout 4200`, the loop exits on
`saved == count` (`SamplingStop::RequestSatisfied`), not on the timeout. I checked the other stop
conditions you asked about:

- `--max-stale-seconds` / `--max-pixel-stale-seconds` default to `-1` and are post-run verdicts
  (`:868`, `:875`), not loop breaks — and passing either would have *disarmed* the guest-fault early
  stop (`:431-448`). Not passed, so not a factor.
- the guest-fault early stop is armed (settle 1.0 s, `:322`), but in wall-clock mode with `saved > 0`
  a short run from it is explicitly excused (`short_run_explained`, `:851-853`), so a guest death
  mid-route does not turn into a spurious failure.
- the one that *can* bite is `saved != count` → `assertion_failed` (`:854`), i.e. a timeout is a hard
  FAIL. The margin is `4200 − 3850 = 350 s`: the run fails on the shortfall if the first *captured*
  frame arrives later than ~350 s. This document's own `:286` puts the title screen at t≈200 s with
  frames before it, so on this host the margin holds. Worth a clause (non-blocking 2 below).

**And the samples land where the route needs them.** Worst case (1.5 flips/s) f5200 is 3,467 s, which
still leaves five 70-second samples inside the gameplay block; at 3.4 flips/s, thirty-three. Sampling
70 s apart instead of 3 s apart also *reduces* host perturbation (a 4K `present_snapshot` copy per
sample), which on a route whose recorded failure mode is host-load desync is a second small win.

### Blocking 2 — the `.pad`. Fixed, and the route is inert.

- Body **byte-identical to master** with comments and blanks stripped (152 lines both sides), and no
  non-comment line appears in this commit's diff at all.
- Line 1 still names `PPSA03831`.
- `:19-28` is now the two-sided form and names both limits — the missing arm counts and
  "reaching that marker is not the same as the authored route having run" — and its last two clauses
  are near word-for-word the document's `:84-86`. No residual asymmetry between the two files on save
  data: the queue-length mechanism is gone from `.pad:66-67` (now "so the run starts from a known save
  state"), and the `.md`'s `:42` scopes its "falsifies" to the page count while `:63-64` says
  explicitly that this "is a statement about the mechanism, not about whether the save matters".

Both non-blocking items you took landed: `:160-165` no longer calls the flag a convenience, and
`:42`'s "falsifies" now names its object.

---

### The one thing I would fix before merging — the new `:42` row is a broken Markdown table row

The added qualification is wrapped onto a second physical line, so the row is two lines:

    | Is the queue route-stable? | ... rather than leaving it unseparated — the page count, not the wider
    question of whether save state matters, which the 2×2 above does not settle. ([#2790](...), PR #3049.) |

GFM ends a row at a newline, so this renders as **two rows**. Confirmed against GitHub's own renderer
(`gh api -X POST /markdown --input ... mode=gfm`) rather than by inference — the table comes back with
six `<tr>`, and the last one is:

- row 5, cell 1: *"question of whether save state matters, which the 2×2 above does not settle.
  (#2790, PR #3049.)"*
- row 5, cell 2: *(empty)*

while the "Is the queue route-stable?" answer cell now **ends on the words "the page count, not the
wider"**. So on the rendered page the row asserts "…so this **falsifies** that story…" and trails off,
the scoping clause you added becomes an unlabelled orphan row, and the `#2790` / PR provenance links
detach from the row they belong to. That is the edit inverting itself: the clause added to stop the
row over-claiming is precisely the part that falls out of the cell.

**The docs gate cannot see this.** `check_numbered_table.py:365` only accepts a row line that
*starts* with a pipe, so the continuation line ends the table run and is never arity-checked —
`9 table(s), 70 rows, unbroken, every row matching its header`, exit 0, and the `Docs` job is green.
GFM, by contrast, needs only that the line contain pipes. The gate and the renderer disagree about
where the table ends, and here the renderer is the one the reader sees.

Fix is a one-character join of the two lines. I am not blocking on it because the words are all
present and adjacent and a cell that visibly trails off invites a second look rather than a false
belief — but it should not land. It needs no re-review: joining two lines cannot change a claim, and
I have verified everything else at this head.

---

### Non-blocking

1. **The published recipe has never been run.** The old invocation was one somebody executed; this
   one is derived from the flag semantics — correctly, as above, but derived. A reader assumes a
   documented command was exercised. Half a sentence ("derived from the flag semantics; not yet run
   end to end") is exactly the standard the rest of this revision applies to itself.
2. **Say what the 4200 is for.** The prose explains 3,850 and leaves the timeout unexplained; it is
   350 s of slack for boot before the first capture, and a shortfall is a hard failure rather than a
   short run. One clause, or `--timeout 4500` and the point disappears.
3. **`:160-165` reads as self-contradicting until you follow the pointer.** "the same windows drive a
   CPU-only arm at ~30 flips/s and a live arm at ~3 flips/s unchanged — … and also why
   `--warmup-seconds` breaks it: moving the flip rate relative to wall clock moves every window off
   the guest state it was authored for." I went looking for an unsupported mechanism here and did not
   find one: row `:471` carries it with evidence — the boot's progression is paced by asset loading
   and movie playback **in wall clock**, and at `--warmup-seconds 90` the f1700-f1940 presses landed
   ~50 s before the main menu existed. `pad_frame_now()` (`hle_pad.cpp:280-288`) confirms the anchors
   are pure guest flip counts, so flip-anchoring fixes the *flip* and not the *state*, which is the
   whole resolution. It just is not in the sentence. "…off the guest state it was authored for (the
   boot is paced in wall clock, not flips — see the Ruled out row)" closes it.
4. **Two single-value rates in a sentence this commit rewrote.** "~30 flips/s" and "~3 flips/s" at
   `:162-163` against the document's own 17–30 and 1.5–3.4 bands twenty lines up. Pre-existing text,
   but the line was touched, so it was the moment to make them bands.
5. **`PROSPER_SAVE0=~/frontiers-work/save/run1` is not fresh on a second run** — the naming convention
   implies incrementing, but nothing says so, and the `.pad` footer's stated purpose ("so the run
   starts from a known save state") is defeated the second time it is used.
6. Still open from earlier rounds, all still non-blocking: the switch table `:107-111` shows 48 → 52
   with no conditions and no arm-to-row mapping for "five arms"; `:60` "host load is the remaining
   candidate" is n=1 with no `uptime` value recorded; `.pad:15-17` names only host load where the
   `.md` correctly names two causes.
7. **Two follow-up issues worth filing**, neither for this PR: (a) `check_numbered_table.py` accepts a
   wrapped table row that GFM splits — a few lines (warn when a non-pipe-initial line directly
   follows a table run and contains an unescaped pipe) would catch it repo-wide, and this is the
   second Markdown-versus-gate divergence in this document's history; (b) `tools/screenshot` has no
   way to delay *capture* without suppressing *rendering*, which is why this recipe has to sample a
   whole hour to photograph one state — a `--capture-after-seconds` that leaves the renderer live
   would generalise to every long route in the repo.

### Housekeeping, checked at this head

- `check_numbered_table.py` on `SONIC_FRONTIERS_STATUS.md`: 9 tables, 70 rows, unbroken, exit 0
  (green, and blind to the row above).
- `git diff --check origin/master...aac25745`: clean, exit 0.
- Diff is two files. No commit on `origin/master` has touched either path since the branch base, so
  there is no concurrent append to the `Ruled out` table; nothing else in the repo carries a stale
  copy of the old invocation.
- CI on run 33091454517: 8 pass (Docs, Linux, Linux App, Linux App Package, Windows App, Windows
  MinGW, macOS App, Contribution shape), 1 pending (macOS x86_64 / Rosetta 2). The docs-only merge
  waiver does **not** apply — `reach-gameplay.pad` is not `.md` — so wait for the last check.

**APPROVED** on `aac25745`. The recipe is arithmetically sound, terminates on the right condition,
and covers both route targets with margin; the `.pad` is comment-only, still names `PPSA03831`, and
now says the same thing about save data as the document. Please join the two lines of the `:42` table
row before you merge.

PR #3050, opened 2026-08-27
gpu(#325): render guest 2D_ARRAY textures, so Tomb Raider's world is textured

mattias800:
Tomb Raider I–III Remastered textures its entire world from **one 512×512×256 BC7 array**, so until
now every world surface sampled slice 0 and the world rendered untextured. Croft Manor now renders.

**142,635 distinct colours against master's 64,730** on the same route frame; 35 of 45 route frames
above 50k. The title screen also improves (212,775 colours — Lara's model, logo, thumbnails, menu).

## The defect was not where I looked first

Not the recompiler, not the descriptors, not the decode. `fr.tex_byte_size` had **exactly one
assignment in the renderer**, on the MSAA path; everywhere else it stayed `0`. #3043's multi-layer
span contract refuses to build a layered image without proof every layer is readable, so it computed
`0 >= 268435456` and **rejected the atlas outright** — which is why enabling layers turned the
picture *uniform* rather than merely leaving it flat.

It also explains a bisect that misled me for hours: a single-layer resource never reaches that check
(it returns `true` at `sample_count == 1`), so "1 layer works, 4 fails" read as *multi-layer upload
is broken* when it was *the span check is reached at all*. The byte span now travels with the pixels
through both decoded-texture caches.

## The invariant

**Arrayed-ness is a property of the resource, not of the instruction.** The uploader picks the view
type from the guest T# and cannot see which opcode will sample it, so a declaration keyed on the
instruction is a descriptor mismatch for any DIM=1 sample of an array texture.

All three sides now consult **one** predicate — `guest_texture_is_uploaded_array()`, written beside
`bc_block_bytes` precisely because two copies that can disagree is the bug this change is about:

| side | follows |
| --- | --- |
| uploader | decodes `depth` slices, publishes the count through `sample_count` |
| view | `VK_IMAGE_VIEW_TYPE_2D_ARRAY` iff `sample_count > 1` |
| recompiler | `Arrayed` at **every** declaration site for such a resource; `tex_coord_uv()` appends layer 0 for non-array instructions |

Layer 0 is the same base slice those instructions got from the old base-slice 2D view, so behaviour
is preserved wherever it was already right.

**Block-compressed only**, and that is a measured limit rather than caution: the per-slice decoder
carries tiled and linear BC and plain byte-per-texel surfaces, but not the fp16/fp32/unorm16
narrowing the single-surface decoder does — a Float32 array decodes to **black**. Widening the
predicate means teaching the slice loop those conversions first.

## Two things the tests caught, both real

This change initially broke two `gpu_capture_render_replay` arms, and both were my bugs:

1. The slice loop is derived from the **cube** path, which never reads pitch-padded linear rows — so
   a linear array's slice 0 decoded differently than the same surface did before it was an array.
   `linear_padded_read` already names `img_dim 5`; the array path simply did not honour it.
2. `test_rdna2_spirv_struct` pinned the base-slice 2D contract for graphics DIM=5, which this change
   deliberately replaces. Updated to assert the new contract, keeping its fourth-address-LOD clause.

## Known remaining defect — stated because it is visible

**Interior scenes texture the wrong surfaces** (passport pages and game consoles on walls and
floors) while the exterior and title screen are correct. It is the layer **index**, not the
addressing: the measured layout is `layer_stride 352256` (262144 for mip 0 plus a 6-level chain)
with `layer_mip_offset 90112`, and the exterior samples that same atlas correctly. Tracked on #2998.

This is worth calling out for the reviewer's judgement: interiors go from *untextured* to *wrongly
textured*. I think the net is clearly positive — the exterior and title screen are correct and the
world is recognisably the game — but it is a visible change in both directions and I would rather
that be an explicit decision than a buried one.

## Diagnostics

- `PROSPER_ARRAY_DECODE_BUDGET_MIB` (default 1024) bounds the decoded footprint: this atlas is
  **256 MiB** once decoded to RGBA8, another resource reaches **464 MiB**, held in host RAM *and*
  VRAM. Uploading BC compressed would be 4× cheaper and is the better long-term answer.
- `PROSPER_SLICESTRIDE=1` prints each array's measured layout, once per texture address.

## Verification

- `cmake --build prosper/build-linux -j10` — clean, 0 errors
- `ctest --no-tests=error -j4` — **311 of 311 pass**
- Route re-run on **defaults**, no environment variables set
- `git diff --check` clean; 0 `Claude-Session:` trailers
- Contract and the measured layout recorded in `docs/TOMB_RAIDER_STATUS.md`


mattias800 (commented):
**REJECTED** at head `82d7670c` — read-and-reason review, no build/test re-run.

I want to lead with what is right, because it is most of the change: the *diagnosis* is correct and important. `fr.tex_byte_size` really did have exactly one assignment in the renderer (`live_renderer.cpp:3031`, the MSAA path), the span contract really does return `true` at `sample_count == 1` (`render_runner.h:455`), and those two facts really do explain both the rejection and the misleading bisect. The byte span now travels correctly on every array-publishing path I could find, and it is the *actual* buffer length, not a recomputation: `texture_pixels.resize(nb)` (`live_renderer.cpp:4587`) makes `size()` exactly `layer_bytes * layers`, and the persistent hit re-reads `cached->second.pixels.size()` (`:4062`). Both submit-local emplaces carry it positionally (`:4362`, `:6157`). Item 3 of the brief is clean.

Also verified correct, so nobody re-derives them: `image_sample_grad_2d`'s gradients correctly stay 2-component (SPIR-V excludes Arrayed from the Grad operand count); `image_gather_offset_2d`'s offset stays 2-component for the same reason; mip generation is already excluded for `img_dim == 5` (`render_runner.h:6655`), so there is no undefined mip chain behind the implicit-LOD sample; `array_footprint` is computed in `uint64_t` and cannot overflow at T#-bounded extents, and the budget is applied *before* the allocation via `decoded_layers`; `declare_texture`'s key includes the arrayed bit (`rdna2_to_spirv_internal.hpp:1740`, `:1758`), so an in-shader disagreement is fail-visible rather than silent; and the block-compressed gate on the slice loop IS airtight — its only entry is `(is_cube || is_array)` at `live_renderer.cpp:5180` and `is_array` requires `bc_block_bytes != 0`.

The rejection is about the central invariant, which does not hold as claimed.

## Blocking

**B1. CI is red at this head, on this PR's own code.** `Docs` fails at step 7, `check_diag_gates.py`:

```
TWO-GATE|frontends/shared/live/live_renderer.cpp|report|PROSPER_ARRAY_DECODE_BUDGET_MIB+PROSPER_SLICESTRIDE  @frontends/shared/live/live_renderer.cpp:5194
```

`[slicestride]` (`live_renderer.cpp:5195`) prints `layers=%u` from `slice_count`, which `PROSPER_ARRAY_DECODE_BUDGET_MIB` decided — a diagnostic printing a field a different switch measured. Note this reproduces only against the *committed* tree: the branch's worktree carries an uncommitted fix, so a local run there is green. That is worth stating plainly, because it means the local "all checks passed" and CI's failure are both true statements about different trees.

**B2. `rdna2_emit_alu.cpp:7554` hardcodes `arrayed=true`, and it is a live descriptor mismatch.** The graphics `SAMPLE_C_LZ` DIM=5 branch declares `Arrayed=1` from the *instruction*, not from `res_arrayed`. Your own uncommitted fix names the failure — this title's shadow maps are `img_dim 5` with **depth 1**, uploaded as plain 2D, producing `VUID-vkCmdDraw-viewType-07752`. Two further consequences of the same line that the local fix also resolves but that are worth recording: a shader touching one binding from both this site and any `res_arrayed` site gets a `declare_texture` key mismatch and is rejected *whole*, dropping every draw that used it; and the comment justifying the hardcode ("the uploader binds a 2D_ARRAY view for every img_dim==5 resource") is not what `render_runner.h:6962` does.

**B3–B5 are the same defect in three reachable places, and none is fixed in your worktree: a binding declared `Arrayed=1` reached by a helper that still builds a 2-component operand. That is invalid SPIR-V, not a wrong pixel, and `spirv-val` will not catch it** — it gates representative modules, not game shaders (`rdna2_emit_cfg.cpp:6672`). The driver gets it directly.

- **B3. Graphics `image_load` on a DIM=5 array.** `rdna2_emit_alu.cpp:7638` declares `host_array || res_arrayed`; `host_array` is compute-only for loads, so in graphics a plain `image_load` (opcode 0x00, `mimg_dim == 5`) declares Arrayed via `res_arrayed`, then falls through every branch to `image_fetch_2d` (`rdna2_to_spirv_internal.hpp:2142-2145`), which builds `t_v2u()`. `OpImageFetch` on a 2D Arrayed image requires 3 coordinate components. Pre-PR this site declared non-arrayed and was consistent; the guard at `:7353` covers `image_load_mip` in graphics but not `image_load`.
- **B4. `image_sample_lz_o` on an arrayed binding.** `image_sample_lz_offset_2d` issues `OpImageQuerySizeLod` with `t_v2i()` (`rdna2_to_spirv_internal.hpp:2082`); Arrayed=1 requires `ivec3`. Reached from `rdna2_emit_alu.cpp:7667`.
- **B5. `image_get_resinfo`.** `:7256` now passes `res_arrayed`, but `image_get_resinfo` still picks `dim == Dim_2D ? t_v2i() : t_v3i()` (`rdna2_to_spirv_internal.hpp:2104`) — same invalid result type, and `out[2]` should then be the layer count rather than the constant 1.
- **B5b, lower reachability but the same shape.** `image_sample_dref_manual_2d`'s `linear_filter` branch consults the CALLER's `arrayed` for both the size query (`:1925`) and the tap fetch (`:1984`), while only the nearest branch consults `coord_arrayed` (`:2012`). Callers at `:7481` and `:7572` pass the default `false`, so an arrayed binding with a linear compare sampler emits `%v2int` + a 2-component fetch against an Arrayed image. Routing those two through `coord_arrayed` closes it.

**B6. The budget clamp is not the graceful fallback the comment claims, and it is the same mismatch class.** `live_renderer.cpp:3277-3280` folds `array_footprint <= array_budget_bytes` into `is_array`, so an over-budget array publishes `sample_count = 1` and takes `VK_IMAGE_VIEW_TYPE_2D` (`render_runner.h:6963`) — while the recompiler still declares Arrayed, because `guest_texture_is_uploaded_array` (`bc_decode.hpp:34`) has no budget term. The comment at `:3269` says this is "the pre-#325 behaviour"; it is not, it is a 2D view under an `Arrayed=1` OpTypeImage. The same hole opens whenever `rtt_hit`, `dcc_fast_clear_done` or `cube_depth_bridged` short-circuits the slice loop at `:5180`, leaving `array_done` false so `:6004-6006` never publishes the layer count.

The sharpest evidence that this is real is in your own diff: `render_runner.h:6955-6961` states the rule ("a guest 2D_ARRAY takes an ARRAY view even at one layer… Keying the view on the layer COUNT instead would silently disagree whenever an array resolved to a single layer, binding a 2D view under an `Arrayed=1` OpTypeImage") and `:6962-6964` immediately keys it on `r.sample_count > 1u` — the layer count. The comment describes the fix; the code is the bug it warns about.

So the invariant as stated in the PR body and in `TOMB_RAIDER_STATUS.md` — "all three sides consult one predicate" — is not true of the graphics view side. Graphics keys the view on a *consequence* of the uploader; compute keys it on the SPIR-V reflection (`live_compute.cpp:5832-5834`), which is why the pre-existing hardcoded `true` at `:7536` was safe there and its graphics twin was not. Either make `res_arrayed` the single source for the view too, or make the recompiler's predicate include everything the uploader's does. Half of that is one line; the budget term is the harder half, because the recompiler has no budget.

**B7. Process.** This is a progression PR ("Croft Manor now renders", plus a stated visible regression on interiors) with no screenshot in the body and no `BLOG.md` entry in the diff. The charter requires representative captures naming title/platform/frontend/checkpoint for a progression claim, and a `BLOG.md` entry in the same PR. Given the interiors caveat you flagged for judgement, the pictures are exactly what a reviewer needs to make that call.

## On the interiors trade-off you asked about

I agree the net is positive and I would take "wrongly textured" over "untextured" here — the exterior and title screen being correct is the evidence that the addressing is right and only the index is wrong, and #2998 is the right place for it. That judgement is not what I am rejecting on.

## Non-blocking

- **The block-compressed-only rationale is overstated, in both `bc_decode.hpp:29-31` and `TOMB_RAIDER_STATUS.md`.** The slice loop *does* carry fp16 (`live_renderer.cpp:5292`) and *does* carry unorm16 narrowing (`:5325-5331`). What it actually lacks is fp32, and any 4-byte-per-texel narrow format, which fall to the raw `source_bpt == 4` memcpy at `:5312`. Please narrow the claim to what is true — a `## Ruled out`-adjacent statement that overstates a gap sends the next agent to write conversions that already exist.
- **Stale diagnostics now saying the opposite of what happens.** `rdna2_emit_alu.cpp:7343` still prints `2D_ARRAY image_sample -> sampled as base slice 0 (array index dropped; #325)` under `PROSPER_GFXLOG` — the standard bring-up switch — on the very path that no longer drops it. Same for the comments at `:7336-7338` and `:7698-7699` ("Graphics keeps the established base-slice view for DIM=5"), whose branch `array_sample` now captures.
- `PROSPER_ARRAY_DECODE_BUDGET_MIB` is parsed with `atoll` (`live_renderer.cpp:3273`): a malformed value yields 0 and silently disables every array; a negative one yields a huge budget. The charter's rule for these is that a malformed value disables its own trigger rather than firing at an unintended moment — here it changes rendering.
- `face_base`'s fallback stride (`live_renderer.cpp:5185-5187`) uses the tight `selected_span` when `layer_stride_bytes` is 0. For a pitch-padded linear array that is short by the padding; `:3631` already computes `linear_src_row * th` for exactly this case.
- `PROSPER_SLICESTRIDE` is `getenv`-truthy, so `=0` also prints; the doc says `=1`.

## What I checked and did not find

No other view-type selection for sampled guest textures exists in the graphics backend (`render_runner.h:6962` is the only one; the other `viewType` sites are render/depth targets). Every 2D coordinate builder is accounted for above — `image_sample_2d`, `_bias_`, `_grad_`, `_lod_`, `image_gather_2d`, `image_gather_offset_2d` and `image_get_lod_2d` all route through `tex_coord_uv`, and `image_get_lod_2d` is additionally guarded on `res->img_dim == SQ_DIM_2D`. `OpImageQueryLod`'s `v2f` result is correct for arrayed images.

Happy to re-review on a corrected head; B3–B6 are the ones that need new code rather than a push.

mattias800:
All seven blocking findings fixed across `fcff635b` and `19ac43ec`. This was an unusually good
review — four of the seven I had not found, and **B6 invalidated this PR's own headline claim**.

## B6 — the claim was wrong

I wrote "three sides, one predicate". It was **not true of the view side**: the view keyed on
`sample_count > 1`, a *consequence* of the uploader, while the recompiler keyed on the predicate.
Anything that leaves the layer count at 1 without changing the shape — the footprint budget, an RTT
hit, a DCC fast-clear — produced a 2D view under an `Arrayed` declaration.

`FrameResource` now carries `guest_array`, set from the shared predicate and **independent of how
many layers were actually decoded**; the view keys on that. A one-layer `2D_ARRAY` view is legal and
samples layer 0, so the budget is a decode cap again rather than a correctness hazard.

Your observation about *why* the pre-existing hardcoded `true` in the compute dref path was safe
where its graphics twin was not — compute picks its view from the SPIR-V reflection, graphics from
the uploader — is the thing that made this obvious once stated. I left compute alone.

## B3, B4, B5, B5b — four invalid-SPIR-V sites

None visible to the test suite, because a validation error does not fail a test:

| | site | defect |
| --- | --- | --- |
| B3 | `image_fetch_2d` | 2-component `uvec2` coordinate; graphics `IMAGE_LOAD` DIM=5 reaches it |
| B4 | `image_sample_lz_offset_2d` | `OpImageQuerySizeLod` as `ivec2`; Arrayed returns `ivec3` |
| B5 | `image_get_resinfo` | same — and its third component is the layer **count**, which is exactly what GET_RESINFO's third result means for a 2D_ARRAY T#, so reporting it is right rather than merely legal |
| B5b | `image_sample_dref_manual_2d` | manual-fetch branch used the caller's `arrayed` while the sampled branch used the declaration; they disagreed and only one had been fixed |

**`tools/vkval/vk_validation_scan.py` now passes** — every observed message ID accounted for. That
scan caught B2 before you did and would have caught these; I should have run it before opening the
PR, not after CI went red.

## B1, B2, B7 and the dirty worktree

B1 (diag gate) and B2 (hardcoded `arrayed=true` against depth-1 shadow maps) are in `fcff635b`,
along with a fixture that bound a 2D-compiled shader to an array resource. B7: three screenshots and
a `BLOG.md` entry.

On the dirty worktree — that was a real process failure and your review was harder to do because of
it. I had local fixes while CI was red on the pushed head. Both heads are pushed now and the tree is
clean.

## Non-blocking, all taken

- The "block-compressed only" rationale **was overstated** in two places: the slice loop does carry
  fp16 and unorm16; only fp32 and the 4-byte narrow formats are missing. Corrected rather than
  repeated.
- A `PROSPER_GFXLOG` line still announced *"sampled as base slice 0 (array index dropped)"* — now the
  opposite of what happens — and the comment above it described a fallback that no longer exists.

## Verification at `19ac43ec`

- build clean; `ctest --no-tests=error -j4` → **311 of 311**
- `check_diag_gates.py` → all checks passed
- `vk_validation_scan.py` → PASS
- numbered-table docs gate → rc=0
- route re-run on **defaults**; Croft Manor still renders (213,170 colours best frame, 35/45 above 50k)

mattias800 (commented):
## Re-review at `19ac43ec` — REJECTED

Tree clean, head matches. Five of seven findings are fixed well, and B6 in particular is fixed
*correctly* rather than patched. Two remain: one is B5b left half-done, and one is a **new
regression this PR introduces** that reopens B2 at a different site.

### BLOCKING — B8 (new): `array_sample` lost its `is_compute` gate but not its `img_dim`-only test

`prosper/src/gpu/recompiler/rdna2_emit_alu.cpp:7632`

```
const bool array_sample = in.mimg_dim == 5u &&
    res->img_dim == 5u && (is_sample || is_sample_l || is_sample_lz);
const bool host_array = load_2d_array || mip_load_2d_array ||
    array_sample || msaa_array_fetch;
if (!b.declare_texture(res->binding, Dim_2D, uint_texture, host_array || res_arrayed)) {
```

`array_sample` keys on `res->img_dim == 5u` **alone** — it consults neither `res->depth` nor the
format. The shared predicate requires `depth > 1u && bc_block_bytes(format) != 0u`. So the two
sides disagree for every `img_dim == 5` T# that is depth-1 or not block-compressed:

| | recompiler | uploader |
| --- | --- | --- |
| `img_dim 5, depth 1` | `array_sample` → `Arrayed=1` | `guest_array=false` → `VIEW_TYPE_2D` |
| `img_dim 5, depth>1, non-BC` | `array_sample` → `Arrayed=1` | `guest_array=false` → `VIEW_TYPE_2D` |

That is `VUID-vkCmdDraw-viewType-07752` — the *exact* failure B2 was, reintroduced at a different
site — and it is new: before this PR `array_sample` carried `b.is_compute &&`, and in compute the
view comes from SPIR-V reflection, so the term was self-consistent. Removing the gate moved it into
the one stage where the view comes from the uploader instead.

Both rows are populated by your own evidence, not by hypotheticals. The `depth 1` row is the case
you diagnosed in `fcff635b`: *"this title's shadow maps are img_dim 5 with **depth 1**, so they are
uploaded as plain 2D, and declaring them Arrayed produced VUID-vkCmdDraw-viewType-07752."* Those
particular ones survive only because they arrive through `is_sample_c_lz`, which you did fix — a
plain `IMAGE_SAMPLE` on the same T# lands here instead. The non-BC row is the `Float32` array you
measured decoding to black; under the shipped predicate it takes the 2D view, and a graphics
`SAMPLE` dim=5 on it now declares `Arrayed`.

This also falsifies the PR body's headline invariant. The table says the recompiler follows
`Arrayed at every declaration site for such a resource` — for graphics `SAMPLE` it follows
`host_array || res_arrayed`, which is a *second* predicate with different terms. Two copies that can
disagree is precisely the bug the PR is about.

Sharpest confirmation: the `PROSPER_GFXLOG` line you corrected in the same commit prints
`res_arrayed`, so on exactly these bindings it will report *"resource is NOT an uploaded array"*
while the shader declares `Arrayed=1`. The diagnostic and the declaration already contradict each
other in the shipped code.

Suggested fix — keep the graphics side on the one predicate, leave compute on reflection:

```
const bool array_sample = in.mimg_dim == 5u && res->img_dim == 5u &&
    (b.is_compute || res_arrayed) && (is_sample || is_sample_l || is_sample_lz);
```

Worth an assertion or a loud reject on `host_array && !b.is_compute && !res_arrayed` so the next
divergence is fail-visible rather than a validation message nobody runs.

### BLOCKING — B9: B5b fixed two of three uses, not both of two

`prosper/src/gpu/recompiler/rdna2_to_spirv_internal.hpp:1935`

```
uint32_t size = id(); put(code, Op_ImageQuerySizeLod,
                          {arrayed ? t_v3i() : t_v2i(), size, img, uconst(0)});
```

You hoisted `coord_arrayed` with the comment *"Hoisted so BOTH the sampled and the manual-fetch
branches below use it"*, and the commit message says *"The flag is hoisted so both use it."* There
are **three** uses of arrayed-ness in this function; the two `OpCompositeConstruct` sites were
converted and this `OpImageQuerySizeLod` — sitting in the same `linear_filter` branch as
`compare_fetch`, ten lines above it — still reads the caller's `arrayed`.

This is the B4/B5 failure mode, not a wrong pixel: `OpImageQuerySizeLod`'s result component count is
validated **exactly** (`arrayed + 2` for `Dim2D`), so `ivec2` against an `Arrayed=1` image is
invalid SPIR-V. The reachable combination is declared-arrayed + caller-`arrayed==false` + linear
filter, and this PR just created it at both remaining call sites — `rdna2_emit_alu.cpp:7486` and
`:7579` now declare with `res_arrayed` and call with the defaulted `arrayed = false`. Narrow today
(it additionally needs `depth_compare` on a BC array), but it is a one-token fix in a function you
were already editing for this exact class:

```
{coord_arrayed ? t_v3i() : t_v2i(), size, img, uconst(0)}
```

Components 0 and 1 stay width/height either way, so nothing downstream moves.

### Verified fixed

- **B1** — `[slicestride]` is single-gated on `PROSPER_SLICESTRIDE` and prints `guest_depth=%u` from
  `r.depth`; `PROSPER_ARRAY_DECODE_BUDGET_MIB` lifted to `array_decode_budget_bytes()` at file
  scope. The report no longer depends on two switches. `[array-budget]` is unconditional and
  rate-limited, so it introduces no second gate.
- **B2** — `rdna2_emit_alu.cpp:7567` now passes `res_arrayed`, matching its declaration three lines
  up. Leaving the compute twin at `true` is right, and your reasoning for why is the correct one.
- **B3** — `image_fetch_2d` branches on `tex_is_arrayed` to `t_v3u_fetch()` with `uconst(0)`.
- **B4** — `q_type` follows `tex_is_arrayed`; `w_i`/`h_i` extraction unchanged.
- **B5** — `arrayed_2d` drives both the query type and the component count. And `out[2]` is
  **correct**, not a new wrong value: `GET_RESINFO`'s third result for a 2D_ARRAY T# is the array
  size, which is what `OpImageQuerySizeLod`'s third component returns for `Arrayed`. Reporting the
  decoded layer count rather than the guest's declared depth is the honest answer here, since it is
  what the shader can actually index.
- **B6 — this one is fixed properly, and it is the finding that mattered.** `FrameResource::guest_array`
  is set at `live_renderer.cpp:3292` from the shared predicate, structurally before the file's single
  `R.push_back(std::move(fr))` at `:6700`; the only two earlier `continue`s drop the resource entirely
  rather than publishing it. I checked all four paths I named:
  - **budget** — `guest_array` is computed before `is_array` and is not conditioned on it, so an
    over-budget atlas keeps the array view and decodes one slice.
  - **`rtt_hit` / `dcc_fast_clear_done`** — both live on the decode-miss path, below the assignment;
    neither can clear it.
  - **`cube_depth_bridged`** — same, and `array_done` is correctly kept distinct from `cube_done`
    so an array is never multiplied into height.

  No contradiction between view and image: `tci.arrayLayers = r.sample_count` (`render_runner.h:6859`)
  and `tvci.subresourceRange` layerCount `= r.sample_count` (`:6992`) move together, so
  `guest_array && sample_count == 1` is a legal one-layer `2D_ARRAY` view over a one-layer image, and
  the `Arrayed` declaration samples layer 0 — the pre-#325 pixel. MSAA (`img_dim == 6`) never reaches
  the assignment and is still carried by the `|| sample_count > 1u` arm. The `r.cls == RC::Texture`
  conjunct correctly keeps storage images out, since those declare through
  `declare_storage_image`'s own independent `arrayed`. `TextureBindingKey` already includes
  `tvci.viewType`, so one shared image can safely carry both view types.

  One thing I specifically checked because it would have been silent: `guest_array` was inserted
  **mid-struct**, between `img_dim` and `texture_format`. There is no positional aggregate
  initialization of `FrameResource` anywhere in the tree — every construction assigns by name — so
  nothing shifted. (`DecodedTexture`'s new fields, which *are* positionally initialized at
  `live_renderer.cpp:4375` and `:6175`, were correctly appended last.)
- **B7** — three 1920x1080 captures committed and all three referenced from a `BLOG.md` entry that
  is in the file's style: one-sentence lead, pictures carrying the weight, descriptive alt text, the
  interior defect noted in one line, tracker linked. The before/after pair at 1.0 MB vs 4.6 MB is
  itself consistent with the claim.

### Fifth-site sweep — the rest are clean

I walked every 2-component construction in the file. Besides B9, these are correct **as they stand**
and should not be "fixed":

- `image_sample_grad_2d`'s `grad_x`/`grad_y` stay `t_v2f()` — `Grad` excludes the array layer.
- `image_sample_lz_offset_2d`'s `offv` stays `t_v2i()` — `ConstOffset`/`Offset` likewise.
- `image_query_lod_2d`'s **result** stays `t_v2f()`, which is right unconditionally. Its
  *coordinate* now goes through `tex_coord_uv`, so an arrayed binding gets three components where
  `OpImageQueryLod` needs two — legal, because a coordinate *"may be a vector larger than needed,
  but all unused components appear after all used components"*, and the appended `0.0` is last.
  Harmless; flagging only so it is a known deliberate over-supply rather than a latent surprise.
- The storage-image `OpImageQuerySize` at `:2399` derives `arrayed` from `ncoord == 3` in its own
  namespace and is untouched by this change.

### Non-blocking corrections — both accurate now

- The revised **"block-compressed only"** rationale checks out against the slice loop: it carries
  `f16` (`half_to_float`), `component_bytes == 1` byte-per-texel, `component_bytes == 2`
  (`unorm16_to_unorm8`) and a 4-byte `memcpy`, with no fp32 narrowing — so *"fp16 and unorm16
  carried, fp32 and the 4-byte narrow formats missing"* is right. The `Float10_11_11` and
  `Unorm2_10_10_10` post-passes correctly learned `array_done ? decoded_layers`.
- The **`PROSPER_GFXLOG`** line is accurate as literal text. See B8 for why it will nonetheless
  contradict the declaration until `array_sample` is gated.

Fix B8 and B9 — B8 is the one that needs thought, B9 is one token — and I will re-review the head
promptly. The B6 rework is genuinely good and I do not want to re-litigate any of it.

**REJECTED**

mattias800:
Both fixed at `0d065286`, and they are the same mistake in different clothes: a site keyed on `img_dim == 5` where the uploader keys on the full predicate.

**B8** is a regression this branch introduced, and your framing is exactly right — lifting the `b.is_compute &&` gate off `array_sample` left it keyed on `img_dim == 5` alone, so it fed `host_array` and declared Arrayed for resources the uploader gives a plain 2D view. **Both failing cases are my own evidence from earlier in this branch**: the depth-1 shadow maps from the `fcff635b` message, and the Float32 array I measured decoding to black. Now `(b.is_compute || res_arrayed)`.

Your explanation of why compute is exempt — it picks the view from the SPIR-V reflection rather than from the uploader — is the thing that makes this coherent rather than a special case, and it is now in the comment.

**B9**: three uses of the flag in that function, I converted two, and my commit message claimed both halves. That is the second time in this PR I have described a fix as complete when it was partial, which is worth more than the one-token diff.

## Verification at `0d065286`

- build clean; `ctest --no-tests=error -j4` → **311 of 311**
- `vk_validation_scan.py` → **PASS**
- `check_diag_gates.py` → all checks passed
- route re-run on defaults; Croft Manor renders (213,945 best frame, 35/45 above 50k)

Thank you for the fifth-site sweep and for confirming `out[2]` is genuinely the array size rather than a new wrong value — I would not have been able to assert that with the same confidence. The note that `image_query_lod_2d` now over-supplies a third coordinate component (legal, trailing components ignored) is recorded; I have left it rather than special-case it, since the alternative is another site keyed on the declaration for no behavioural gain.

mattias800 (commented):
**APPROVED** at `0d065286` (tree clean, HEAD matches the PR head). B8 and B9 are both correctly and completely fixed. Three non-blocking items below; the first is the one I would like to see land before merge.

## B8 — correct, and the fall-through is right for the right reason

I traced both things `array_sample` controls.

**Declaration.** With `(b.is_compute || res_arrayed)` at `rdna2_emit_alu.cpp:7645`, a graphics DIM=5 instruction on a depth-1 / non-BC `img_dim 5` T# now gives `array_sample=false`, and the other three `host_array` terms cannot rescue it: `load_2d_array` and `mip_load_2d_array` are still `b.is_compute &&` gated (`:7628`, `:7630`), and `msaa_array_fetch` requires `res->img_dim == 6u` (`:7411`), which cannot coexist with `img_dim 5`. So `declare_texture(..., host_array || res_arrayed)` at `:7649` passes `false`, matching what the uploader hands it — `live_renderer.cpp:3288` sets `fr.guest_array` from the same `guest_texture_is_uploaded_array()`, and `render_runner.h:6970` keys the view on it. I checked the MSAA leg specifically: the view is `(r.guest_array || r.sample_count > 1u)`, so 2D_MSAA keeps its array view even though `guest_array` is false for `img_dim 6`. No leg left mismatched.

**Sampling.** It falls to `image_sample_2d` / `image_sample_lod_2d`, which drop `cvg(2)`. That is right, and not merely tolerable:

- `res_arrayed=false` means either `depth <= 1` — one layer, so index 0 is the only slice there is — or a non-BC format whose slice loop cannot decode past the base slice anyway (`bc_decode.hpp:27-33`, the Float32-decodes-to-black measurement). Either way the dropped slice reproduces master exactly, so this is a restored behaviour rather than a new one.
- `is_sample_l` still reads its LOD from address slot 3 at `:7712`, so the guest vaddr layout stays instruction-keyed while the view stays resource-keyed. That is the correct split and the delta does not disturb it.

**Compute is bit-identical to master.** `(b.is_compute || res_arrayed)` is unconditionally true in compute, so `array_sample` reduces to master's `b.is_compute && dim5 && img_dim5 && (sample|sample_l|sample_lz)`. Identical predicate, identical helper selection. (The `host_array || res_arrayed` widening does change compute declarations for non-array instructions on arrayed T#s, but that is pre-delta and safe — `live_compute.cpp:5754-5843` picks its view from the reflected `image_arrayed`, which is exactly the asymmetry the new comment now states.)

I also confirmed the graphics fall-through cannot land on an arity mismatch from the other direction: every 2D helper reachable from that chain builds its coordinate through `tex_coord_uv` / `tex_is_arrayed` (`rdna2_to_spirv_internal.hpp:1778-1789`), `image_sample_grad_2d`'s gradients stay 2-component (correct for Dim 2D Arrayed), and `declare_texture` returns `false` on a conflicting key (`:1759`), so any residual disagreement inside one module is fail-visible rather than silent.

## B9 — in scope, and all four decision points now agree

`coord_arrayed` is declared at `:1925`, above the `linear_filter` branch, so it is in scope at `:1939`. I counted the uses myself — four decision points, and the `arrayed` / `coord_arrayed` split is the correct one at every one:

| line | decides | flag | right? |
| --- | --- | --- | --- |
| `:1939` | `OpImageQuerySizeLod` result arity | `coord_arrayed` | yes — Arrayed 2D queries as ivec3; `CompositeExtract 0/1` still gets w/h |
| `:1997` | fetch coord arity | `coord_arrayed` | yes |
| `:2000` | fetch layer **value** | `arrayed` | yes — caller-supplied slice, else base 0 |
| `:2028` | sample coord arity | `coord_arrayed` | yes |
| `:2031` | sample layer **value** | `arrayed` | yes |

Arity follows the declaration, the layer value follows the caller. No use of the bare `arrayed` remains where arity is being decided.

## Sweep — no fifth site

I swept `img_dim == 5` across the recompiler. Everything left is instruction-shape or vaddr-layout keyed, which is correct: `rdna2_decode.cpp:1207,1211`; `rdna2_emit_alu.cpp:7340,7343,7358,7392,7532,7628,7630,7712`; `rdna2_to_spirv.cpp:2794-2832`. `rdna2_emit_alu.cpp:7091,7101` is the storage-image path, whose view comes from reflection. `image_get_resinfo` and `image_sample_lz_offset_2d` are declaration-keyed via `tex_is_arrayed`. `image_get_lod_2d` over-supplies a third component as you noted — and it is additionally unreachable for an arrayed binding, since `:7367` requires `res->img_dim == SQ_DIM_2D`; agreed it should stay as it is.

## Comments added in the delta

Both accurate. The uploader predicate really is `img_dim==5 && depth>1 && bc_block_bytes!=0` (`bc_decode.hpp:35`), so "multi-layer and block-compressed" is exact; "two branches up" is the `is_sample_c_lz` DIM=5 graphics branch at `:7558`; compute really does key on reflection; and "the third of three uses" matches the count.

---

## NON-BLOCKING

**N1 — B8 has no automated guard, and this is provable.** `19ac43ec` reported 311/311 ctest *and* `vk_validation_scan.py` PASS **with B8 present**. So neither instrument can see it. The only live graphics draw against a DIM=5 / `res_arrayed=false` resource is `test_shadow_compare_render.cpp:120-141`, and that goes through the `is_sample_c_lz` branch, not the plain-SAMPLE branch B8 lives in. Removing `(b.is_compute || res_arrayed)` today leaves the whole suite green. The arm is ~8 lines beside the one at `test_rdna2_spirv_struct.cpp:6143-6167`: clone the `rt_array` block at `:5927-5933` with `depth = 1` (or a non-BC format) so `guest_texture_is_uploaded_array()` is false, run `recompile_fragment(ps_sample_array_l, ...)` against it, and assert `!image_arrayed`. That arm fails without the delta and passes with it, which is the property the existing arm cannot express since `rt_array` is BC6 with `depth = 2`. Given this PR has twice shipped a fix described as complete that was partial, the pin is worth more than its size.

**N2 — three comments still state the predicate as `img_dim == 5` alone**, which is the exact key the delta's own new comment says is wrong. A reader following any of them reintroduces B8:
- `rdna2_emit_alu.cpp:7245` — "Mirrors the uploader's own predicate (img_dim 5 AND more than one layer)" omits the block-compressed term.
- `tests/fixtures/render_runner.h:6964` — "the recompiler decides that from the T# (`res->img_dim == 5`)". The surrounding reasoning stays true of the real predicate (`img_dim`, `depth` and `format` are all T# fields, all independent of the decoded layer count); only the parenthetical is stale.
- `tests/gpu/test_rdna2_spirv_struct.cpp:6141` — "the view type is chosen from the resource's img_dim", in the comment justifying the flipped assertion.

Naming `guest_texture_is_uploaded_array()` in all three is a comment-only edit.

**N3 — `render_runner.h:139`** says the view type "must key on THIS, not on `sample_count > 1`", but the code at `:6970` is an OR of the two, and has to be for MSAA. Read as a replacement it is wrong; "must also key on" is what is meant.

Verification accepted as reported at `0d065286` (311/311 with `--no-tests=error`, vkval PASS, diag gates pass, Croft Manor still renders on defaults); I did not rebuild.

mattias800 (commented):
Head note: you pushed `dd85a77e` about eight seconds before my review landed, so GitHub attached the review above to `dd85a77e` while its body names `0d065286`. Establishing by content rather than assuming in either direction, per the charter's stale-review corollary:

`git diff 0d065286..dd85a77e` is **docs-only** — ten added lines in `TOMB_RAIDER_STATUS.md` § Ruled out, no file outside `.md`. Every finding in my review above concerns `rdna2_emit_alu.cpp` and `rdna2_to_spirv_internal.hpp`, neither of which the commit touches.

**APPROVED on `dd85a77e`**, on the same terms.

I checked the new commit while I was here, and it holds up:

- The two falsifications are recorded the way the charter asks — dead hypothesis, the evidence that killed it, the issue link — and the first is the more valuable half: the `Flat` gap is real and #3051 stays open as a latent defect, but 224,363 `[interp]` lines with no control word setting `0x400` anywhere is a clean negative for *this* title. Separating "the defect exists" from "the defect is not this title's cause" is exactly the distinction that stops the next agent re-deriving it.
- The trap-229 citation is accurate. Row 229 is on master (`GAME_COMPAT_ORCHESTRATION.md:357`) and is about `gpu_replay --bundle` replaying pre-decoded draws so an upstream instrument's silence reads as a result; extending it from index/vertex decode to recompile-time diagnostics is the same mechanism, not a stretched one.
- Docs gate green locally: `check_numbered_table.py` over every tracked `.md` → rc=0, no numbered-table row added, so there is no merge-result race to re-run.

N1 from my review above still stands as the one thing worth landing before you merge — the missing `!image_arrayed` arm for a graphics DIM=5 sample on a non-arrayed T#. It is the only finding in this round that no instrument can catch on its own, and `19ac43ec` being 311/311 and vkval-PASS with B8 present is the proof of that.

mattias800:
Approved — thank you. I took **N1 before merging**, because you did not just assert the gap, you proved it: `19ac43ec` was 311/311 **and** vkval PASS with B8 live.

`2b4a07eb` adds the negative arm, covering both halves of the predicate that `img_dim == 5` alone would miss — depth 1, and a non-block-compressed format. **Mutation-verified in both directions** rather than assumed:

```
B8 reintroduced -> [FAIL] graphics DIM=5 declared Arrayed for a resource the uploader
                          gives a plain 2D view (depth=1 fmt=18)
B8 restored     -> 0 failures
```

The arm also separates its failure modes — rejected / no binding / wrongly arrayed. My first version printed one message for all three, and I spent a cycle believing the code was broken when the message simply could not tell me which had happened. Worth the four extra lines.

**N2 and N3 taken.** Both were prose stating the predicate as `img_dim == 5` alone, which is exactly the key the code says is wrong — the recompiler comment now names `guest_texture_is_uploaded_array()` and its three terms, and the view comment no longer claims it keys on the predicate "not on `sample_count > 1`": it is an OR and has to be, because the MSAA plane array has no `guest_array` flag and needs its own arm. You were right on both.

## Verification at `2b4a07eb`

- build clean; `ctest --no-tests=error -j4` → **311 of 311**
- `vk_validation_scan.py` → **PASS**
- `check_diag_gates.py` → all checks passed
- CI 9 pass / 0 fail on the previous head; this head is comments plus one test arm

Merging on your approval. Four rounds and eleven blocking findings on this PR, of which I had found two — the ones that mattered most (B6 invalidating my own headline claim, and B8 being a regression I introduced against my own recorded evidence) were entirely yours.

mattias800:
## Retraction: this PR's progression screenshot was a loading screen

The capture merged here as evidence that Croft Manor's world renders textured — the brick arch,
leaded windows and dolphin fountain — is the game blitting **its own pre-rendered loading picture**,
`2/PIX/HD/MANSION.DDS`. The project owner recognised the artwork.

Measured against every same-resolution asset in the dump:

| asset | mean abs diff | within 8/255 |
| --- | --- | --- |
| `2/PIX/HD/MANSION.DDS` | **0.02 / 255** | **100.00 %** |
| next closest of 41 | 52.77 / 255 | 1.27 % |

Blitting a stored full-screen image requires no world rendering, so that frame did not evidence
world texturing. Retracted in #3060, recorded as **instrument trap 230**, and the image is deleted.

**What this does NOT retract.** The 2D_ARRAY change itself is not being reverted — a fresh capture on
the current default route shows textures genuinely reaching world surfaces, with real correctly
decoded assets applied to correct geometry. What was wrong was the *evidence*, not necessarily the
code. But note the consequence: this PR's claim that the fix made the world render textured is now
**unevidenced**, and nobody has yet produced a capture that establishes it. If you are relying on
that claim, re-derive it.

**Why it survived review:** the frame looks *better* than the emulator could plausibly render — soft
shadows, ambient occlusion, depth of field, per-leaf foliage — which reads as success rather than as
a warning. It was also paired with a genuine untextured "before" shot, so the A/B looked rigorous
while comparing a 3D render against a 2D blit.

#3060 adds `tools/evidence/prerender_check.py` so the check is one command in future:

```
python3 tools/evidence/prerender_check.py <shot> <DUMP_ROOT>
```

exit 2 = the frame is game artwork, exit 0 = no asset matches, exit 1 = the check could not run
(deliberately distinct, so "could not check" never reads as "clean").

Issue #3051, opened 2026-08-27
Fragment varyings never get SPIR-V Flat: is_flat_shaded is decoded, used for the control word, and discarded


mattias800:
`AgcShaderSemantic::is_flat_shaded()` is decoded and used to build the guest's PS input control word
(`agc_shader_layout.cpp:98` and `:123`, setting the `0x400` FLAT_SHADE bit), but it **never reaches
the emitted SPIR-V**. The recompiler's `flat_attrs` set is populated from exactly one source —
`VINTRP opcode == 2`, i.e. `v_interp_mov` (`rdna2_to_spirv.cpp:3015-3019`) — and that is the only
thing that produces a `Flat` decoration for a user varying (`rdna2_to_spirv_internal.hpp:4056`).

**Failure scenario.** A guest attribute whose semantic sets FLAT_SHADE but which the shader reads
with ordinary `v_interp_p1`/`v_interp_p2`. On hardware the parameter cache delivers the provoking
vertex's value in every interpolation slot, so the read yields a constant across the primitive.
prosper emits an Input variable with no `Flat` decoration, so Vulkan interpolates it
perspective-correctly and each fragment sees a different value.

It matters most for varyings that are **indices rather than quantities** — a texture-array slice, a
material or instance id, anything the shader rounds to an integer. A smoothly interpolated index
does not degrade gracefully; it selects a different resource per pixel.

**Concrete instance, measured.** Tomb Raider I-III Remastered, Croft Manor interior, from a frame
bundle captured at 72 s (`PROSPER_GRAB_BUNDLE_AFTER_MS=72000 PROSPER_CAPTURE_FRAMES=12`), submit
20153, `draw[7]` fragment shader:

```
%3326 = OpVariable %_ptr_Input_v4float Input     ; OpDecorate %3326 Location 1  -- and nothing else
%3327 = OpLoad %v4float %3326
%3328 = OpCompositeExtract %float %3327 0
%3346 = OpExtInst %float %12 RoundEven %3347     ; <- the guest rounds it to a layer
%3358 = OpBitcast %float %3348
%3354 = OpCompositeConstruct %v3float %3356 %3357 %3358
%3359 = OpImageSampleImplicitLod %v4float %3353 %3354   ; OpTypeImage ... 2D 0 1 (Arrayed)
```

The array layer is fragment input **Location 1, component 0**, rounded — and the shader contains
[...]

PR #3006, opened 2026-08-26
#304 part two: unannounced 32-bit index buffers whose indices exceed 0xFFFF

mattias800:
## The defect

**Tomb Raider I-III Remastered (`PPSA16901`) rendered its world as long stretched triangles radiating
from a point**, while character meshes in the *same frame* were shaped correctly. That split is the
whole clue.

This is **#304 in a form its detector cannot see**. The title announces an index size **exactly
never** — `index_type=0` on all **508,688** indexed draws of a boot to Croft Manor. Its world index
buffers are 32-bit anyway, because the level is drawn from one shared **775,111-vertex pool**, which
16-bit indices cannot address. Read as 16-bit, each 32-bit index becomes two and every triangle
collapses to a degenerate `(N, K, N)` sliver. Character meshes survived because their index buffers
genuinely are 16-bit.

`index_buffer_is_unannounced_32bit` already recovers this for DOLL, but its fingerprint requires
every high half to be **zero** — true only while indices stay below 65536, which is precisely what a
775,000-vertex pool never produces.

## The fix, and why the byte pattern alone is not it

`index_buffer_is_unannounced_32bit_high`, consulted **only after the existing detector declines**, so
no buffer that one already classifies can change verdict.

**The byte pattern is not sufficient, and review proved it.** With no `DrawIndexOffset` the caller
passes the same pointer twice, so a genuine 16-bit buffer with a **period-2** pattern — a fan or cone
as a triangle strip `[rim, apex, rim, apex, …]`, or a line list radiating from one hub — is
*byte-identical* to a clustered 32-bit list. A 64-spoke line list to hub vertex 7 and a triangle-strip
cone with apex 12 both satisfied every byte-pattern clause of the first version. **No further test
over those two pointers can separate them.**

So the deciding evidence comes from **outside** the buffer: an index must address a vertex that
exists. The detector takes the bound vertex buffer's **unclamped** record count and rejects any
32-bit index at or above it, under an absolute `1<<24` ceiling as well. The cases separate at once —
the cone demands 786,640 vertices from a mesh holding tens; Tomb Raider demands 775,111 from a pool
holding exactly 775,111. *Unclamped* matters: the executor's own `vb_entries` is capped at 65,536,
which would reject exactly the case this exists for. No bound available → declines.

Two further details, both of which cost a round trip:

- **The repeated high half lands on EITHER parity**, depending on how the 16-bit address aligns
  against the 32-bit grid. On the sampled frame the **even** parity carried it for **55,677** draws
[...]

PR #3052, opened 2026-08-27
diag(#2998): PROSPER_SLICEMAP and PROSPER_FORCE_LAYER for guest array textures

mattias800:
Two diagnostics for guest array textures, and the measurements they produced on Tomb Raider's
mis-textured interiors (#2998).

Both exist because the instruments already present could not separate the cases that matter.

## `PROSPER_SLICEMAP=1`

Reports which decoded slices hold content **across the whole range**, not a prefix. That distinction
is the whole point: an array whose first slices decode and whose later ones are empty is identical,
in every per-layer checksum of the first eight, to one that decoded completely. The *shape* of the
map is the measurement.

```
[slicemap] addr=0x20491cc000 512x512 slices=256 nonempty=14  map(every 8th)=##..............................
[slicemap] addr=0x204ef0c000 2048x2048 slices=29 nonempty=29 map(every 8th)=####
```

## `PROSPER_FORCE_LAYER=<n>`

Substitutes a constant array layer **and says so**. It separates *"the shader is given the wrong
slice"* from *"the slice it asks for holds the wrong content"* — two failures that look identical on
screen, and which I had conflated twice before writing this.

It forces guest-visible state, so its output illustrates an investigation and is **never acceptance
evidence** for a rendered frame. That is stated at the code, not just here.

## What they measured

| observation | |
| --- | --- |
| world atlas (512×512×256) | **14 of 256** slices hold content, all at the start of the range |
| other array (2048×2048×29) | **29 of 29** |
| decode count | atlas decodes **once** per run; with the cache disabled it re-decodes **233 times and reports 14 every time** |
| the slice the shader asks for | **248** — `vertexIndex*24 + 3` from binding 7, constant across all 4,206 vertices of the draw |
| `FORCE_LAYER=248` | a completely different frame from unforced (262 colours vs 122,087) — so different draws use different slices |
| `FORCE_LAYER=10` vs `200` | 184,541 colours full-screen vs **262 colours, almost entirely black** |

So the guest indexes a slice our decoded array has no content for.

## Four hypotheses killed

All recorded in `TOMB_RAIDER_STATUS.md` § Ruled out so nobody re-derives them:

- **flat interpolation** — falsified, the title never sets FLAT_SHADE anywhere (#3051, now latent)
- **mixed-slice triangles** — falsified, the slice is constant per draw
- **stale decode cache** — falsified, 233 re-decodes all report 14
- **tight per-slice stride** — falsified, tested and reverted, gives *different* wrong textures

I also withdrew a "reads past the declared `size`" mechanism of my own: the arithmetic was right but
`r.size` does not gate the reads, so the consequence I drew from it was wrong.

## What is still open

Whether `r.depth = 256` describes the guest array's real layer count at all. If the true array is
~16 layers, an index of 248 means something other than a layer, and the whole framing changes.

## Verification

- build clean; `ctest --no-tests=error -j4` — **311 of 311**
- `check_diag_gates.py` — all checks passed (each diagnostic prints only what its own switch measured)
- No behaviour change with either variable unset

### DIFF
--- a/prosper/docs/TOMB_RAIDER_STATUS.md
+++ b/prosper/docs/TOMB_RAIDER_STATUS.md
@@ -120,11 +120,17 @@ Measured layout, for anyone extending this: `layer_stride = 352256` for the 512x
 
 ## Ruled out
 
-- **The atlas is only ever ~4.75 MiB populated of its 86 MiB span** — falsified 2026-08-27. The
-  `[occupancy]` reading `filled=18/344` is an observation at one instant, not a steady state. With
-  the write trace armed at bucket 200 (`offset=0x83cc000`, target `0x204c3cc000`) the upper atlas
-  takes **32 events, `selected=yes changed-during-window=yes`**, against a positive control at
-  bucket 0 that also fired 32. Do not restart "the guest never fills the upper atlas". (#2998)
+- **The atlas is only ever ~4.75 MiB populated of its 86 MiB span, for the whole run** — falsified
+  2026-08-27, but **read the scope before using this**. With the write trace armed at bucket 200
+  (`offset=0x83cc000`, target `0x204c3cc000`) that bucket takes **32 events, `selected=yes
+  changed-during-window=yes`**, against a positive control at bucket 0 that also fired 32. So the
+  allocation is filled progressively and `filled=19/344` is one instant, not a steady state.
+  **What this does NOT establish:** bucket 200 is byte 52,428,800, i.e. slice **148.8** at
+  `layer_stride=352256` — it sits *below* the interior's slices 186-248, which are buckets
+  **249.9-333.2**. This probe never touched the region the interior samples, and the separate
+  finding that slice 248 reads zero throughout still stands (re-confirmed 2026-08-27). An earlier
+  version of this row said "do not restart 'the guest never fills the upper atlas'" without that
+  qualification, which overstated a probe taken below the range in question. (#2998)
 - **A stale persistent decode cache holds an early, near-empty snapshot** — falsified 2026-08-27.
   `PROSPER_NO_TEXTURE_DECODE_CACHE=1` over `scripts/tomb-raider-PPSA16901/reach-gameplay.pad`
   changes nothing: the atlas still decodes once. Caveat on the instrument: `[occupancy]` is gated
@@ -138,22 +144,31 @@ Measured layout, for anyone extending this: `layer_stride = 352256` for the 512x
 - **The guest CPU writer of the atlas cannot be observed** — falsified 2026-08-27, and the cause
   was an allocator, not the watchpoint. `PROSPER_DMEM_WRITE_TRACE` could not target this title
   because only `sceKernelAllocateMainDirectMemory` published a caller chain, while the title
-  allocates through `sceKernelAllocateDirectMemory`. With both entry points sharing one attribution
-  helper the writer is named immediately: `eboot+0xebb82..0xebbfa`, an unrolled AVX scatter storing
-  8x16 B per iteration to offsets computed in `%ymm10`. (#2998, #3054)
+  allocates through `sceKernelAllocateDirectMemory`, which minted none — so no selector could name
+  a byte it owns. With both entry points sharing one attribution helper the writer is named
+  immediately: `eboot+0xebb82..0xebbfa`, an unrolled AVX scatter storing 8x16 B per iteration to
+  offsets computed in `%ymm10`.
+  **Reproduction needs code that is NOT on master.** `attribute_dmem_allocation` is not in this PR
+  and not in `origin/master`; it is tracked in **#3054**. Until it lands, the recipe below reports
+  `caller-chain=0` for this title and the trace refuses to arm — which is the very failure the row
+  retires, so do not read a failed arm as contradicting it. (#2998, #3054)
 
 - **Nothing maps into the atlas per-slice, and no host/kernel write streams into it** — both with
   controls, both new instruments. `PROSPER_MAPWATCH` sees one 1 GiB `map_dmem` covering the atlas,
   established once, never per-slice (control: watching all of dmem shows two maps).
   `PROSPER_HOSTWRITEWATCH` sees **zero** writes into the atlas while its control — the same watch
   over all of dmem — records **107**. So the ~18 slices that do arrive are written by guest CPU
-  stores, which no instrument in the tree can observe.
+  stores, which no instrument in the tree could observe **when this was written**. Superseded
+  2026-08-27: they are observable now — see the writer row in `## Ruled out` (`eboot+0xebb82..`),
+  which also records why the trace could not previously be aimed at this title.
   That is what makes **#3054** the blocking tool: `PROSPER_HWWATCH` is register-relative and cannot
   arm on a fixed guest address, so the one mechanism that would catch a plain guest store is
   unreachable through its current interface.
 
-- **The atlas allocation contains about 4.75 MB of content and nothing else, measured WITHOUT any
-  stride assumption.** `PROSPER_OCCUPANCY=1` walks the guest allocation in 256 KiB buckets and asks
+- **At the moment prosper decodes it, the atlas allocation holds about 4.75 MB of content and
+  nothing above it, measured WITHOUT any stride assumption.** (Scope corrected 2026-08-27: this is
+  one instant, not the whole run. Bucket 200 is written later — see `## Ruled out`. The interior's
+  own slices 186-248, buckets 249.9-333.2, are a separate question and remain unwritten.) `PROSPER_OCCUPANCY=1` walks the guest allocation in 256 KiB buckets and asks
   only "is there content here", so it cannot be fooled by a wrong per-slice layout — which every
   earlier measurement could, since they all addressed slices through `layer_stride`:
 
@@ -192,11 +207,12 @@ Measured layout, for anyone extending this: `layer_stride = 352256` for the 512x
   receives about eighteen slices' worth. Whatever moves DDS content into this allocation is doing so
   for a small fraction of what the title loads.
 
-- **Everything on prosper's side of the interior defect is measured and correct; the atlas is
-  genuinely unpopulated in guest memory.** Twelve candidates eliminated, each with its control or
+- **Everything on prosper's side of the interior defect is measured and correct; the slices the
+  interior samples are genuinely unpopulated in guest memory.** (Scoped 2026-08-27 — see
+  `## Ruled out`.) Twelve candidates eliminated, each with its control or
   its stated scope: layer index (traced twice to a constant 248), flat interpolation (#3051, the
   title never sets FLAT_SHADE), mixed-slice triangles (constant per draw), stale decode cache (guest
-  memory probed ~138,000 times, never becomes non-zero), per-slice stride and mip offset (both
+  memory at the sampled slice probed ~138,000 times, never becomes non-zero), per-slice stride and mip offset (both
   layouts tested), `depth` mis-decode (raw T# confirms 256 / base_array 0), truncated reads
   (`short_reads=0`), partial residency (one `rw-s` mapping covers the whole span), base-address
   mis-decode (dword0 `<< 8` matches), compute STORE (control fired, no compute binds it), recorded
@@ -205,8 +221,10 @@ Measured layout, for anyone extending this: `layer_stride = 352256` for the 512x
   remaining work is a forward investigation from the title's asset/streaming logic, not from the
   frame. Full evidence on #2998.
 
-- **The interior's wrong textures are not prosper mis-reading the atlas. The atlas is genuinely
-  almost empty in GUEST memory, for the whole run.** Measured, each point with the control that
+- **The interior's wrong textures are not prosper mis-reading the atlas. The slices the interior
+  samples are genuinely empty in GUEST memory for the whole run** — narrowed 2026-08-27 from "the
+  atlas is almost empty for the whole run", which is false: the allocation IS filled progressively
+  below slice ~149. The claim that survives is about the interior's OWN slice range. Measured, each point with the control that
   makes it mean something:
   - `PROSPER_SLICEMAP=1`: the 512x512x256 world atlas decodes with **slices 0-13 populated and
     14-255 empty** — contiguous, from the full 256-character map. **`short_reads=0`**, so every
@@ -214,7 +232,9 @@ Measured layout, for anyone extending this: `layer_stride = 352256` for the 512x
     scan samples one byte in 61.)
   - `PROSPER_SLICEWATCH=1` samples GUEST memory at slice 248 on the reuse path, which runs about
     138,000 times per route: it reads **zero on the first reference and never becomes non-zero**.
-    So the guest never fills it, at any point — which retires the decode-cache hypothesis properly.
+    So the guest never fills **slice 248**, at any point — which retires the decode-cache
+    hypothesis properly. Note the scope: this is one slice inside the interior's range, and is NOT
+    evidence about the allocation as a whole (`## Ruled out` records a write at bucket 200).
     An earlier attempt to retire it was **void**: the `PROSPER_NO_TEXTURE_DECODE_CACHE` run never
     left the title screen (its colour count sits at the title's ~210k for all 300 s), so its 233
     re-decodes all happened before the level streamed anything.
@@ -224,7 +244,8 @@ Measured layout, for anyone extending this: `layer_stride = 352256` for the 512x
   - The world shader samples slice **248**, traced twice: `vertexIndex*24 + 3` from binding 7, a byte
     constant across all 4,206 vertices, through `OpBitFieldUExtract(dword, 24, 8)` to the fragment
     varying.
-  So the guest samples a slice that nothing ever wrote, and prosper reads that correctly. **The
+  So the guest samples a slice that nothing ever wrote — that slice, not the whole allocation —
+  and prosper reads that correctly. **The
   missing piece is an atlas upload path prosper does not perform**, not the array plumbing, which is
   now measured end to end. Compute is excluded; DMA is untested (a `PROSPER_DMA_WATCH_DST` run was
   silent but had no control, so it is not evidence).
--- a/prosper/docs/TOMB_RAIDER_STATUS.md
+++ b/prosper/docs/TOMB_RAIDER_STATUS.md
@@ -120,21 +120,30 @@ Measured layout, for anyone extending this: `layer_stride = 352256` for the 512x
 
 ## Ruled out
 
-- **The interior's wrong textures are NOT the layer index arriving wrong, NOT stale cache, NOT
-  interpolation, and NOT the stride.** What they are is still open, but the search space is much
-  smaller. Measured, in order:
-  - `PROSPER_SLICEMAP=1` (new): the 512x512x256 world atlas decodes with **14 of 256 slices holding
-    content**, all at the start of the range. The 2048x2048x29 array decodes **29 of 29**.
-  - The atlas is decoded **once** per run. With `PROSPER_NO_TEXTURE_DECODE_CACHE=1` it re-decodes
-    **233 times and reports 14 every time**, so a stale cache is not hiding later-streamed slices.
-  - The world shader asks for slice **248**: `vertexIndex*24 + 3` from binding 7, a byte that is
-    constant across all 4,206 vertices of the draw. Traced through
-    `OpBitFieldUExtract(dword, 24, 8)` to the fragment varying at Location 1.
-  - `PROSPER_FORCE_LAYER=248` produces a **completely different frame** from the unforced run
-    (262 colours vs 122,087), so the unforced draws are not all sampling 248 -- different draws use
-    different slices.
-  - `PROSPER_FORCE_LAYER=10` gives full-screen content (184,541 colours); **200 gives 262 colours,
-    almost entirely black**. Slices above the populated range are empty.
+- **The interior's wrong textures are not prosper mis-reading the atlas. The atlas is genuinely
+  almost empty in GUEST memory, for the whole run.** Measured, each point with the control that
+  makes it mean something:
+  - `PROSPER_SLICEMAP=1`: the 512x512x256 world atlas decodes with **slices 0-13 populated and
+    14-255 empty** — contiguous, from the full 256-character map. **`short_reads=0`**, so every
+    slice was read completely; "empty" here is not "unreadable". (The count is a lower bound: the
+    scan samples one byte in 61.)
+  - `PROSPER_SLICEWATCH=1` samples GUEST memory at slice 248 on the reuse path, which runs about
+    138,000 times per route: it reads **zero on the first reference and never becomes non-zero**.
+    So the guest never fills it, at any point — which retires the decode-cache hypothesis properly.
+    An earlier attempt to retire it was **void**: the `PROSPER_NO_TEXTURE_DECODE_CACHE` run never
+    left the title screen (its colour count sits at the title's ~210k for all 300 s), so its 233
+    re-decodes all happened before the level streamed anything.
+  - `PROSPER_COMPUTE_BINDS` over the atlas plus three controls: the one line printed is a **control**
+    — the compute program binds a different 4 MiB buffer and executes — so no compute STORE produces
+    this atlas.
+  - The world shader samples slice **248**, traced twice: `vertexIndex*24 + 3` from binding 7, a byte
+    constant across all 4,206 vertices, through `OpBitFieldUExtract(dword, 24, 8)` to the fragment
+    varying.
+  So the guest samples a slice that nothing ever wrote, and prosper reads that correctly. **The
+  missing piece is an atlas upload path prosper does not perform**, not the array plumbing, which is
+  now measured end to end. Compute is excluded; DMA is untested (a `PROSPER_DMA_WATCH_DST` run was
+  silent but had no control, so it is not evidence).
+
 - **`r.depth = 256` is NOT a mis-decode.** The raw descriptor (`PROSPER_TDUMP=512x512`) reads
   `t = 20491cc0 cb600000 007fc07f d0550fac 000000ff 00700050 00000000 00000000` ->
   `512x512 fmt=182 type=13 tile=5 mips=0..5 depth=256 base_array=0`. The layer count, the base slice
--- a/prosper/docs/TOMB_RAIDER_STATUS.md
+++ b/prosper/docs/TOMB_RAIDER_STATUS.md
@@ -120,6 +120,25 @@ Measured layout, for anyone extending this: `layer_stride = 352256` for the 512x
 
 ## Ruled out
 
+- **The atlas allocation contains about 4.75 MB of content and nothing else, measured WITHOUT any
+  stride assumption.** `PROSPER_OCCUPANCY=1` walks the guest allocation in 256 KiB buckets and asks
+  only "is there content here", so it cannot be fooled by a wrong per-slice layout — which every
+  earlier measurement could, since they all addressed slices through `layer_stride`:
+
+  ```
+  [occupancy] 0x20491cc000 span=86 MiB bucket=256KiB filled=18/344 last_filled_bucket=18
+  .##################...........(325 more)
+  ```
+
+  Content sits in buckets 1-18 and stops. The interior samples slices 186-248, which lie 48-87 MB
+  into the allocation. So the data those draws need is not in this allocation at ANY layout, and the
+  earlier per-slice findings were not artefacts of the stride the probes assumed.
+- **The title reads its textures; they simply do not arrive here.** `PROSPER_READBYTES=1` counts
+  bytes delivered per path with no stack walk: **`.DDS opened=299 read=297`** over a route, alongside
+  `.TRM 98/52` and smaller sets. So asset I/O works and 297 texture files are read, while this atlas
+  receives about eighteen slices' worth. Whatever moves DDS content into this allocation is doing so
+  for a small fraction of what the title loads.
+
 - **Everything on prosper's side of the interior defect is measured and correct; the atlas is
   genuinely unpopulated in guest memory.** Twelve candidates eliminated, each with its control or
   its stated scope: layer index (traced twice to a constant 248), flat interpolation (#3051, the
--- a/prosper/docs/TOMB_RAIDER_STATUS.md
+++ b/prosper/docs/TOMB_RAIDER_STATUS.md
@@ -120,6 +120,25 @@ Measured layout, for anyone extending this: `layer_stride = 352256` for the 512x
 
 ## Ruled out
 
+- **The interior's wrong textures are NOT the layer index arriving wrong, NOT stale cache, NOT
+  interpolation, and NOT the stride.** What they are is still open, but the search space is much
+  smaller. Measured, in order:
+  - `PROSPER_SLICEMAP=1` (new): the 512x512x256 world atlas decodes with **14 of 256 slices holding
+    content**, all at the start of the range. The 2048x2048x29 array decodes **29 of 29**.
+  - The atlas is decoded **once** per run. With `PROSPER_NO_TEXTURE_DECODE_CACHE=1` it re-decodes
+    **233 times and reports 14 every time**, so a stale cache is not hiding later-streamed slices.
+  - The world shader asks for slice **248**: `vertexIndex*24 + 3` from binding 7, a byte that is
+    constant across all 4,206 vertices of the draw. Traced through
+    `OpBitFieldUExtract(dword, 24, 8)` to the fragment varying at Location 1.
+  - `PROSPER_FORCE_LAYER=248` produces a **completely different frame** from the unforced run
+    (262 colours vs 122,087), so the unforced draws are not all sampling 248 -- different draws use
+    different slices.
+  - `PROSPER_FORCE_LAYER=10` gives full-screen content (184,541 colours); **200 gives 262 colours,
+    almost entirely black**. Slices above the populated range are empty.
+  So the guest indexes a slice our decoded array does not have content for. The leading remaining
+  question is whether `r.depth = 256` describes the guest array's real layer count at all -- if the
+  true array is ~16 layers, an index of 248 means something other than a layer.
+
 - **Missing SPIR-V `Flat` on the varying that carries the array slice is NOT why interiors sample
   the wrong layers.** The gap is real — `is_flat_shaded()` is decoded, used to build the guest's PS
   input control word, and never reaches the SPIR-V (#3051) — but this title never asks for it. A
--- a/prosper/docs/TOMB_RAIDER_STATUS.md
+++ b/prosper/docs/TOMB_RAIDER_STATUS.md
@@ -135,9 +135,20 @@ Measured layout, for anyone extending this: `layer_stride = 352256` for the 512x
     different slices.
   - `PROSPER_FORCE_LAYER=10` gives full-screen content (184,541 colours); **200 gives 262 colours,
     almost entirely black**. Slices above the populated range are empty.
-  So the guest indexes a slice our decoded array does not have content for. The leading remaining
-  question is whether `r.depth = 256` describes the guest array's real layer count at all -- if the
-  true array is ~16 layers, an index of 248 means something other than a layer.
+- **`r.depth = 256` is NOT a mis-decode.** The raw descriptor (`PROSPER_TDUMP=512x512`) reads
+  `t = 20491cc0 cb600000 007fc07f d0550fac 000000ff 00700050 00000000 00000000` ->
+  `512x512 fmt=182 type=13 tile=5 mips=0..5 depth=256 base_array=0`. The layer count, the base slice
+  and the mip range are exactly what the decode reports, so the layer mapping is faithful and the
+  question is not "how many layers".
+- **Nor is it the slice LAYOUT.** With the array treated as fully tight -- stride = one selected-mip
+  surface AND no per-slice mip offset, the combination an earlier test got wrong by changing only the
+  stride -- the populated count moves 14 -> 19 of 256 and the interior is still wrong. Every layout
+  tried leaves the same invariant: **only about 5 MB of the atlas's 67 MB is non-zero**, whichever
+  way the slices are addressed.
+  So the remaining question is not how prosper reads the atlas but **why most of it was never
+  written**: the guest declares 256 slices at `0x20491cc000` and samples slice 248, while the memory
+  there is zero. That points at a content path prosper does not observe (a DMA or compute upload
+  into the atlas) rather than at the array plumbing, which is now measured end to end.
 
 - **Missing SPIR-V `Flat` on the varying that carries the array slice is NOT why interiors sample
   the wrong layers.** The gap is real — `is_flat_shaded()` is decoded, used to build the guest's PS
--- a/prosper/docs/TOMB_RAIDER_STATUS.md
+++ b/prosper/docs/TOMB_RAIDER_STATUS.md
@@ -120,6 +120,19 @@ Measured layout, for anyone extending this: `layer_stride = 352256` for the 512x
 
 ## Ruled out
 
+- **Everything on prosper's side of the interior defect is measured and correct; the atlas is
+  genuinely unpopulated in guest memory.** Twelve candidates eliminated, each with its control or
+  its stated scope: layer index (traced twice to a constant 248), flat interpolation (#3051, the
+  title never sets FLAT_SHADE), mixed-slice triangles (constant per draw), stale decode cache (guest
+  memory probed ~138,000 times, never becomes non-zero), per-slice stride and mip offset (both
+  layouts tested), `depth` mis-decode (raw T# confirms 256 / base_array 0), truncated reads
+  (`short_reads=0`), partial residency (one `rw-s` mapping covers the whole span), base-address
+  mis-decode (dword0 `<< 8` matches), compute STORE (control fired, no compute binds it), recorded
+  PM4 write paths (all four recorders armed, zero hits), and an unimplemented
+  `sceAgcDriverInitResourceRegistration` (its family is profiler labelling, not memory). The
+  remaining work is a forward investigation from the title's asset/streaming logic, not from the
+  frame. Full evidence on #2998.
+
 - **The interior's wrong textures are not prosper mis-reading the atlas. The atlas is genuinely
   almost empty in GUEST memory, for the whole run.** Measured, each point with the control that
   makes it mean something:
--- a/prosper/docs/TOMB_RAIDER_STATUS.md
+++ b/prosper/docs/TOMB_RAIDER_STATUS.md
@@ -120,6 +120,28 @@ Measured layout, for anyone extending this: `layer_stride = 352256` for the 512x
 
 ## Ruled out
 
+- **The atlas is only ever ~4.75 MiB populated of its 86 MiB span** — falsified 2026-08-27. The
+  `[occupancy]` reading `filled=18/344` is an observation at one instant, not a steady state. With
+  the write trace armed at bucket 200 (`offset=0x83cc000`, target `0x204c3cc000`) the upper atlas
+  takes **32 events, `selected=yes changed-during-window=yes`**, against a positive control at
+  bucket 0 that also fired 32. Do not restart "the guest never fills the upper atlas". (#2998)
+- **A stale persistent decode cache holds an early, near-empty snapshot** — falsified 2026-08-27.
+  `PROSPER_NO_TEXTURE_DECODE_CACHE=1` over `scripts/tomb-raider-PPSA16901/reach-gameplay.pad`
+  changes nothing: the atlas still decodes once. Caveat on the instrument: `[occupancy]` is gated
+  on `is_array && depth > 64`, so "one print" is **not** "one decode" — do not quote it as one.
+  (#2998)
+- **`base_array` is dropped between the T# and the uploader** — falsified 2026-08-27. It is
+  applied in `agc_shader_layout.cpp` `image_base_level_view()`:
+  `view.base += (uint64_t)d.base_array * stride;` with `view.layer_stride = stride` set alongside,
+  on the `thin_2d_layered` `depth > 1` path. `live_renderer.cpp` never names `base_array` because
+  the shift is already folded into `gpu_addr`, which is correct, not a gap. (#2998)
+- **The guest CPU writer of the atlas cannot be observed** — falsified 2026-08-27, and the cause
+  was an allocator, not the watchpoint. `PROSPER_DMEM_WRITE_TRACE` could not target this title
+  because only `sceKernelAllocateMainDirectMemory` published a caller chain, while the title
+  allocates through `sceKernelAllocateDirectMemory`. With both entry points sharing one attribution
+  helper the writer is named immediately: `eboot+0xebb82..0xebbfa`, an unrolled AVX scatter storing
+  8x16 B per iteration to offsets computed in `%ymm10`. (#2998, #3054)
+
 - **Nothing maps into the atlas per-slice, and no host/kernel write streams into it** — both with
   controls, both new instruments. `PROSPER_MAPWATCH` sees one 1 GiB `map_dmem` covering the atlas,
   established once, never per-slice (control: watching all of dmem shows two maps).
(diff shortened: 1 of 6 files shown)


mattias800 (commented):
**REJECTED** on head `ac2d91bc` — three cheap doc/instrument-labelling fixes, none of them in the emitter logic.

Let me lead with what I checked and found sound, because it is most of the PR.

**Both diagnostics are inert by default.** `[slicemap]` is behind `getenv("PROSPER_SLICEMAP")` and does nothing but read and print. `PROSPER_FORCE_LAYER` unset gives `forced_layer = -1`, `layer_use = bcf(layer_bits)` — byte-for-byte the same operand the old code passed. The `static const int` is a C++11 magic static, so its initialisation is thread-safe and happens once even though recompilation may run on workers. `fconstf` writes its `OpConstant` into `types`, not `code`, so the forced form stays valid SPIR-V.

**No cache can outlive the setting.** I chased this specifically because the key does not include the variable. The recompile cache (`ShaderCompileKey` / `shader_cache()`, `gpu_executor.cpp:622`) is in-memory only, and `forced_layer` is constant for the life of the process, so no run can hold both variants. The only on-disk cache is the *compute* driver blob (`live_compute.cpp:1188`), opt-in behind `PROSPER_DISK_PIPELINE_CACHE`, and it is keyed by the SPIR-V the driver is handed — a forced module is a different module, so it takes a different slot and an unforced run misses it rather than reusing it. Clean.

**Gates.** I ran both checkers on this head myself: `check_diag_gates.py` → 108 findings, 108 baselined, all checks passed; `check_cached_env.py` → all checks passed. On the TWO-GATE question you would expect to bite: `slices=` prints `slice_count`, which *is* `decoded_layers`, which *is* what `PROSPER_ARRAY_DECODE_BUDGET_MIB` decides — and the pre-existing comment three lines above (`live_renderer.cpp:5205`) explicitly warns against printing that field. It is nonetheless benign, and the reason is worth writing down rather than leaving for the next reader to re-derive: reaching the print requires `is_array`, which requires `array_footprint <= array_budget_bytes`, and `decoded_layers` is then unconditionally `r.depth`. The budget can only silence the line, never change the number on it — which is exactly the tool's own test ("if the producer's switch is unset, does the field still print?"). Please add that one clause to the comment, or print `r.depth` as `[slicestride]` deliberately does.

Now the three blocking items.

---

**1. The `## Ruled out` rows conclude "the memory there is zero", and `[slicemap]` cannot distinguish that from memory prosper could not read.**

This is the one that matters, because it is the row that redirects the whole investigation.

The new text says: *"why most of it was never written: the guest declares 256 slices at `0x20491cc000` and samples slice 248, while the memory there is zero"*, and *"the array plumbing … is now measured end to end."*

The read path is precisely the link that is not measured. On the BC branch — which is the branch this atlas takes — `copy_resource`'s return value is discarded at `live_renderer.cpp:5238` and `:5253`, and `safe_copy` (`:2286`) walks 64 KB at a time and **breaks at the first uncommitted page, returning short**. `traw` / `lin` were zero-initialised, so the unread tail stays zero and `bc_decode_surface` turns it into transparent black. A slice prosper cannot read and a slice the guest left zeroed produce the identical byte pattern, and `[slicemap]` counts both as empty.

That alternative is not exotic here, it is the leading one: `layer_stride = 352256 × 256` declares a **90 MB** span from `0x20491cc000`, and the observation being explained is that only the first ~5 MB is non-zero. "The guest allocated 90 MB and wrote 5" and "the guest allocated ~5 MB and prosper read past the end into nothing" predict the same map. The non-BC branch in this same loop already knows this matters — it captures `got` and falls back when `got < linear.size()` (`:5273`) — so the asymmetry is visible in the diff's own neighbourhood.

The fix is small and it is the same fix in both directions: have `[slicemap]` accumulate what `copy_resource` actually returned per slice (or probe each slice base and print `mapped=N/256`). One extra field separates the two hypotheses in a single run and turns this row from an inference into a measurement. Failing that, reword to what was observed — "decodes to zero pixels" — and record "prosper could not read the tail" as still live rather than as excluded by "measured end to end".

I want to be explicit that I am not asserting the reads fail. I am asserting the instrument cannot tell, and the row states the conclusion that requires it to have told.

**2. `nonempty` is a 1-in-61 sample presented as an exact count, and the every-8th map discards exactly the resolution the prose needs.**

`for (size_t i = 0; i < lsz && !any; i += 61u)` touches ~17,190 of 1,048,576 bytes — 1.6%. For real BC7 content that is fine and I do not think 14 is wrong. But neither the printed line nor the comment says "sampled", and `nonempty=14` has already travelled into a doc row, a PR table and a commit message as a count. One word in the format string (`nonempty~=` or `nonempty(sampled 1/61)=`) costs nothing and stops that.

The stride on the map is the sharper half. `map(every 8th)` for 256 slices is 32 characters where the full map is 256 — no cost worth the loss. And the loss is load-bearing: the doc says the 14 are *"all at the start of the range"*, but `nonempty=14` plus a map showing `L=0` and `L=8` set and `L=16,24,…` clear does not establish contiguity. It is consistent with, say, 12 slices in `0..15` and two more at `L=100` and `L=101`, which the map never samples. Print every slice — or print the indices — and the claim becomes readable off the output instead of inferred around it. (The label itself is honest, so credit where due; the problem is what the stride throws away, not that it is undisclosed.)

Minor, same line: `slices=%u` prints the full count while the `break` at `(L+1)*lsz > texture_pixels.size()` can silently truncate `nonempty` and `map`. It should not fire given the decode loop writes those same offsets unchecked, but if it ever does the line understates while looking complete. An `examined=` field closes it.

**3. `PROSPER_FORCE_LAYER`'s scope is unstated at the code, and the new rows inherit it unqualified.**

The substitution lives only in `image_sample_2d_array` — plain implicit-LOD SAMPLE. `image_sample_lod_2d_array` (SAMPLE_L / SAMPLE_LZ) and `image_fetch_2d_array` are untouched, and on this title that is not hypothetical: the census recorded in this same doc puts **64** `SAMPLE_C_LZ` (`0x2f`) events on array textures against 17 `0x20` events.

You already found and corrected this once. The row further down carries your own words — *"Scope, because the first version of this line overstated it: the probe sits in the implicit-LOD array sampler only, so it covers the `0x20` events and structurally cannot cover the 64 `SAMPLE_C_LZ` (`0x2f`) events."* The new rows above it re-state unqualified conclusions from the same probe (*"different draws use different slices"*, *"Slices above the populated range are empty"*), and the new comment in `rdna2_to_spirv_internal.hpp` — which is where the next person will look — does not mention the limitation at all. One sentence in the comment and a scope clause on the rows. This is the correction most likely to be re-lost, because it has already been lost once.

---

Non-blocking, in descending order of how much I would want them:

- **`atoi` makes `PROSPER_FORCE_LAYER=` force layer 0.** Empty string and garbage both yield `0`, a perfectly valid layer, so the shell idiom `PROSPER_FORCE_LAYER= ./prosper-app` silently arms the probe at layer 0 instead of disabling it. The charter's own standard for the capture triggers is that a malformed value disables its own trigger rather than firing at an unintended moment. `strtol` with an end-pointer check, and treat an empty or unparsable value as -1.
- **The WIP-branch banner at `docs/TOMB_RAIDER_STATUS.md:163` is now stale for one of the probes below it.** It sends the reader to `wip/issue-325-texture-arrays` to reproduce a constant-layer probe that this PR puts on master under a stable name. Add "except `PROSPER_FORCE_LAYER`, which landed in #3052".
- **The two constant-layer rows read as contradictory.** One says forcing changed nothing (identical hashes, one byte of luma in 288); the other says forcing 248 gives a completely different frame. The reconciliation — *"the resource was uploaded with one layer"*, i.e. pre-#3050 — is a trailing parenthetical on the older row. Promote it to the front of that row, or the next reader treats one of the two as wrong.
- **`static uint32_t said = 0; said++`** is a non-atomic read-modify-write and shader recompilation can run on more than one thread. Diagnostic-only and only when armed, so I would not hold the PR for it, but `std::atomic<uint32_t>` is free here.
- **`bcf(layer_bits)` now runs before `bcf(u_bits)`/`bcf(v_bits)`**, where the braced-init-list previously ordered them u, v, layer. Semantically identical — SSA, and the bitcast still dominates its use — but the emitted id numbering and instruction order do change with the variable unset. "No behaviour change with either variable unset" is true of behaviour and not of bytes; worth the half-sentence in the PR body so nobody bisects a SPIR-V diff into this.
- **When forced, a dead `OpBitcast` is still emitted** (`layer_use` is computed then overwritten). Valid, unused, harmless — noting it only so it does not read as a leak later.
- **`PROSPER_NODETILE` sits upstream of the pixels `nonempty` is derived from** (`:5198`), in the same block. `check_diag_gates.py` cannot see it — its header says a negated env test imposes nothing, deliberately. So the PR body's *"each diagnostic prints only what its own switch measured"* is slightly stronger than what a green run proves. Not a defect; just do not quote the tool as having established it.
- **`[slicemap]` prints once per decode with no dedupe, where its sibling `[slicestride]` three lines above dedupes by address.** I read the difference as deliberate — the 233-line repetition is what falsified the stale-cache hypothesis — so leave it, but it is worth a word in the comment saying the repetition is the point, since the neighbouring convention is the opposite.
- Neither variable gets a "what it does, how to run it" line the way `PROSPER_SLICESTRIDE=1` does at line 119. They appear only inside prose. One line each in the *Guest 2D_ARRAY textures* section would match the section's own convention.

---

To be clear about the shape of this verdict: the emitter change is right, the inertness is real, both gates genuinely pass, and I could not find a way for either switch to leak into a default run. What I am rejecting is the three sentences that state more than the instruments can distinguish — and given that these land in `## Ruled out`, where this project treats a row as the most expensive thing in the tree, that is the part worth another pass. Fix 1, 2 and 3 and I expect to approve on sight.

mattias800:
All three blocking findings fixed at `97763bd5` — and the first one **changed the conclusion**, not
just the wording, so thank you for it.

## 1. The slice map could not tell "unwritten" from "unreadable"

You are right, and this was load-bearing: the BC branch discarded `copy_resource`'s return while the
non-BC branch in the same loop already checked its `got`, and `safe_copy` zero-fills the tail past
the first uncommitted page. Every "the atlas is mostly empty" claim I had made turned on exactly
that distinction, and I had no basis for it.

It now records short reads per slice and prints them, with a warning on the line when any occurred.

**Measured: `short_reads=0`.** The zeros are real — but that is now a result rather than an
assumption.

## 2. Sampling reported as an exact count

Fixed: the count prints as `nonempty>=N (1 byte in 61 sampled)`, and the **full 256-character map**
prints instead of every 8th slice. As you say, it costs nothing — and it is what the "all at the
start of the range" claim actually needed:

```
map=##############..................................................................(242 dots)
```

Slices 0-13, contiguous. The claim survives, but now on evidence rather than on a sample that could
not distinguish it from scattered placement.

## 3. `PROSPER_FORCE_LAYER` scope

Taken, and it stings: this is the **same scope error I had already made and withdrawn once earlier in
this very file**. Both samplers now share one accessor. I also took the non-blocking `atoi` note — a
malformed or empty value now **disables** the probe rather than forcing layer 0, per the charter's
malformed-value-disables-its-own-trigger standard.

## Following finding 1 produced the result

A new `PROSPER_SLICEWATCH` samples **guest** memory at a high slice on the reuse path, ~138,000 times
per route. Slice 248 reads zero on the first reference and **never becomes non-zero**.

That also exposed that my *earlier* retirement of the decode-cache hypothesis was **void**: the
`PROSPER_NO_TEXTURE_DECODE_CACHE` run never left the title screen, so its 233 re-decodes all happened
before the level streamed anything. I had reported that as evidence. It is now recorded as the trap
it was.

## Verification at `97763bd5`

- build clean; `ctest --no-tests=error -j4` — **311 of 311**
- `check_diag_gates.py` — all checks passed

On your TWO-GATE note: agreed that `slices=` printing `decoded_layers` is benign for the reason you
give (reaching the print requires `is_array`, after which it is unconditionally `r.depth`), and I
have left it rather than restructure for a distinction the gate itself does not draw.

mattias800 (commented):
**REJECTED** at `f8bd6e17` — one blocking finding, plus one trivial companion in the same file.

Findings 2 and 3 are properly fixed, and finding 1 is fixed *for the path the Tomb Raider
conclusion rests on*. I confirmed fmt=182 -> BC7_SRGB (`gen5_image_format`,
`agc_shader_layout.cpp:281`), so `cb = 16` and the TR atlas takes the instrumented branch. The
docs' `short_reads=0` is therefore a real measurement, and the conclusion now rests on it. That
part is right, and following it to `PROSPER_SLICEWATCH` — plus retracting the void
`PROSPER_NO_TEXTURE_DECODE_CACHE` retirement — is a better outcome than the fix I asked for.

## BLOCKING — finding 1 is fixed for BC only, and `short_reads=0` now asserts a measurement that did not happen for every other format

`live_renderer.cpp:5331-5333` is the only writer of `slice_short_count`, and it sits inside
`if (cb) {` at `:5293`. The non-BC branch opens at `:5337` and never touches it. All four of its
sub-paths discard the byte count:

- `:5352` ctiled — computes `got`, but uses it only to pick the linear fallback at `:5355`; never recorded.
- `:5374` `copy_linear_padded_rows_from(...)` — return discarded entirely.
- `:5383` the `layer_stride_bytes` row loop — discarded.
- `:5388` `copy_resource(linear.data(), selected_addr, linear.size())` — discarded.

The printer at `:5470-5473` is gated only on `getenv("PROSPER_SLICEMAP") && !is_cube &&
slice_count > 1`. So for any uncompressed array — RGBA8 2D_ARRAY is the common case, and the MSAA
plane arrays reach here too — the line prints a confident `short_reads=0` **and** suppresses the
` <- EMPTY MAY MEAN UNREADABLE` suffix at `:5473`, which is the one signal telling the reader not
to trust the map.

Apply this repo's own criterion. `tools/env/check_diag_gates.py` states the test in its docstring:

  1. If the producer does not run, does the field still print? Here: yes, as `0`.
  2. Is that value indistinguishable from a real measurement? Here: yes — byte-identical to the
     genuine BC `short_reads=0`.

Both yes is the defect, and the docstring's own words are the point: *"an absent field invites the
question, a `0` answers it wrongly."* For the uncompressed class this is worse than before the
fix — previously there was no `short_reads` field to believe. The gate passing (108/108 baselined,
I ran it) is not evidence against this: the scanner is lexical over `PROSPER_*` variables, so a
producer gated by a **code branch** rather than a switch is structurally inexpressible to it. Same
shape as the charter's same-source-positive-control trap — the tool fires, on a space that cannot
contain the case.

The reason it slipped is in your reply and in the code comment at `:5307-5311`: *"the non-BC branch
below already checks its `got`."* It checks `got` to choose a fallback; it does not report it. The
sentence is true and the inference from it is not, which is exactly why nobody re-derived it.

Minimal fix, either is fine:
- capture each non-BC path's return into a `slice_got` and reuse the same record at `:5331`
  (for the ctiled sub-path, judge on the read that actually filled `linear`); or
- if instrumenting all four is more than you want here, make the scope explicit — print
  `short_reads=n/a` (or omit the field) when `cb == 0`, and say why in the comment.

## Fix in the same push — `image_sample_lod_2d_array` substitutes silently

`rdna2_to_spirv_internal.hpp:1905-1906` now substitutes the forced layer, but unlike
`image_sample_2d_array:1885-1889` it emits no `[force-layer]` line. Your own comment at `:1874-1877`
gives the standard: *"A probe that could not show its own lever moved would leave a null meaning
nothing."* A title whose array sampling goes entirely through the explicit-LOD helper now gets full
substitution with **zero** console output, and the operator reads that silence as "the switch did
not take" or "this title has no array sampling". Hoist the `said++ < 4u` announcement into
`forced_array_layer()`'s callers, or into a small helper both call.

## Non-blocking

- `image_fetch_2d_array:2216` is still uncovered, and I agree it should be: its comment says the
  third coordinate is a guest **MSAA sample plane**, not a texture array layer, so forcing it would
  be a category error. Worth one clause saying so, since "the fetch variant" was in my finding and a
  future reader will ask.
- `PROSPER_SLICEMAPS` stops at `printed < 8` (`:4319`) with no truncation marker. It is least
  trustworthy exactly where partial residency is most likely — a fragmented span. One `... N more`
  line would close it. Did not bite here (one line printed).
- Bare `getenv` for `PROSPER_SLICEWATCH`/`PROSPER_SLICEMAPS` means `=0` enables them. Consistent with
  the neighbouring `PROSPER_SLICESTRIDE`/`PROSPER_SLICEMAP`, so not a new sin — noting only because
  you applied the opposite (and correct) standard to `PROSPER_FORCE_LAYER` in the same diff.
- Unguarded function-local statics in both new blocks (`shown`, `watched`, `announced_*`). Matches
  the pre-existing `static std::unordered_set reported` in the same function, so it is the local
  convention. Note that `announced_zero`/`announced_nz` are global across addresses, not per-address:
  fine here (only the 512x512x256 array clears `depth > 64`), misleading if a second deep array ever
  appears, since the reference counter would conflate them.
- `std::vector<uint32_t> slice_short(slice_count, 0)` at `:5272` allocates unconditionally with
  `PROSPER_SLICEMAP` unset. Decode path only, which is cached, so negligible.

## Checked and correct

- Inertness: `forced_array_layer()` returns -1 unset, both samplers fall back to `bcf(layer_bits)`;
  the reordering of `bcf` emission relative to `coord = id()` changes id assignment only, and both
  orderings define before use.
- Malformed handling at `:1859-1864` is right — empty, trailing garbage, negative and >65535 all
  return -1 and disable, rather than `atoi`'s silent layer 0.
- `PROSPER_SLICEWATCH` cost is genuinely gated: `copy_resource` sits inside the `getenv` guard, and
  `is_array && r.depth > 64u` short-circuits ahead of it. Placement inside `else if (decoded_reuse)`
  (`:4282`) matches its stated purpose.
- `copy_linear_padded_rows_from` returns the summed byte count, so a full read is `bw*cb*bh == comp`
  and the BC comparison at `:5332` is the right one on all three of its paths.
- Docs: the twelve eliminations count out exactly twelve; every switch cited exists
  (`PROSPER_COMPUTE_BINDS` `gpu_executor.cpp:2199`, `PROSPER_DMA_WATCH_DST`
  `command_processor.cpp:2636`). The stale `PROSPER_FORCE_LAYER` colour counts measured under the
  narrow scope are gone rather than reworded — right call. Holding DMA as *untested for want of a
  control* instead of counting it as an elimination is the strongest thing in the diff.

mattias800 (commented):
**REJECTED** at head `3cb9a859`.

All five blocking findings from the two earlier rounds are genuinely fixed — I checked each against the source, not against the PR body. What I am rejecting is new: four of the six diagnostics landed *after* the last review, and three of them reintroduce, in a newer instrument, the exact defect shape round 1 already made you fix in `[slicemap]`. One of them is not hypothetical — its false negative is visible in the output this PR's own `## Ruled out` row quotes.

---

## Prior findings — verified fixed

**Round 1 / finding 1 (empty vs unreadable).** `live_renderer.cpp:5314-5315` declares `slice_short`/`slice_short_count`; the BC branch keeps its byte count at `:5353-5357` and records the short read at `:5373-5375`; the printer at `:5523-5532` prints `short_reads=%u`, the ` <- EMPTY MAY MEAN UNREADABLE` suffix, and a `first short slice=` line. Fixed.

**Round 1 / finding 2 (sampled count presented as exact; every-8th map).** `:5524-5525` now prints `nonempty>=%u (1 byte in %zu sampled)`, and `:5515-5521` builds a character per slice — the full map, no stride. Fixed.

**Round 1 / finding 3 (`FORCE_LAYER` scope unstated at the code).** `rdna2_to_spirv_internal.hpp:1910-1919` covers the explicit-LOD sampler and the comment there names the scope error explicitly. The doc no longer carries any `PROSPER_FORCE_LAYER` colour counts (`grep` over `TOMB_RAIDER_STATUS.md`: zero hits). Fixed.

**Round 2 / blocking (non-BC branch printed an unearned `short_reads=0`).** All four sub-paths now capture: ctiled at `:5397`, `linear_padded_read` at `:5412`/`:5419`, the row loop at `:5428`/`:5432`, plain copy at `:5434-5436`; recorded at `:5442-5444`. No sub-path leaves `nbc_want` at 0, so none silently skips the record. Fixed.

**Round 2 / companion (`image_sample_lod_2d_array` substituted silently).** `:1911-1918` calls `announce_forced_layer`, and `said` is now `std::atomic<uint32_t>` at `:1863`. Fixed.

Also confirmed sound: `forced_array_layer()` at `:1870-1880` disables on empty/garbage/negative/out-of-range rather than falling to layer 0; the C++11 magic static makes it thread-safe and constant for the process; with the variable unset `layer_use = bcf(layer_bits)` is the identical operand. `PROSPER_ENV_ON` (`dispatch.hpp:51`) caches, so the three new renderer switches cost one branch on a default boot, and `is_array && r.depth > 64u` short-circuits ahead of them. None of the eight new names appears in a `setenv`, so `check_cached_env`'s intersection stays empty.

---

## BLOCKING 1 — `[occupancy]` reports a 0.2% head-sample as a bucket census, and its own quoted output already contains a false negative

`live_renderer.cpp:4357-4372`. The probe reads `uint8_t probe[512]` at each 262,144-byte bucket head — **512 bytes in 262,144, 0.195%** — and the printed line says nothing about that window. Round 1 made you add `(1 byte in 61 sampled)` to `[slicemap]` for exactly this reason, and `[slicemap]`'s comment at `:5500-5505` states the rule: without the sample rate "its output is not interpretable". `[occupancy]` is now the instrument carrying the strongest new claim in the diff — `TOMB_RAIDER_STATUS.md:133-144`, *"measured WITHOUT any stride assumption ... the data those draws need is not in this allocation at ANY layout"* — and it discloses nothing.

The failure is not hypothetical. Take the two numbers this PR publishes together:

- `TOMB_RAIDER_STATUS.md:168-170`: `[slicemap]` reports **slices 0-13 populated**.
- `TOMB_RAIDER_STATUS.md:139-141`: `[occupancy]` prints `filled=18/344` with the map `.##################...` — **bucket 0 empty**, buckets 1-18 filled.

With the layout this same doc records at `:117-118` (`layer_stride = 352256`), slice 0 occupies bytes `[0, 352256)` and bucket 0 is `[0, 262144)` — 262,144 bytes lying **wholly inside slice 0**. One instrument says that region holds content; the other says it holds none. The only reconciliation is that the 512-byte head sample read a `.` on a populated bucket. And `[slicemaps]` in this same run reports one `rw-s` mapping over the whole span, so this is not a read failure — it is the sample window.

That is precisely the mechanism that invalidates the inference the row draws from the 325 trailing dots. "Buckets 19-343 are empty" is really "the first 512 bytes of each of buckets 19-343 are zero", and the run demonstrates that this can be false for a populated bucket. On this layout the beat between a 262,144-byte bucket and a 352,256-byte slice puts bucket heads at 43 distinct offsets within a slice, so the miss is systematic rather than a one-off.

Three further gaps in the same block, all cheap:

- **`got` is computed at `:4359` and thrown away.** `copy_resource` walks 64 KiB at a time and returns short at the first uncommitted page, so a bucket prosper could not read scans zero bytes at `:4362`, sets `any=false`, and prints `.` — byte-identical to an empty bucket, with no `short_reads` field and no warning suffix. This is round 1's finding 1 verbatim, in the newer instrument. (It probably did not bite on this run, given the single mapping — but the line cannot say so, and the next array will not come with that guarantee.)
- **The 512-bucket cap is undisclosed.** `:4357` stops at `i < 512` while `:4371` prints `n` as the denominator. An array spanning more than 128 MiB prints `filled=X/n` having examined only the first 512 buckets, and `last_filled_bucket` then reports 511 as a measurement when it is the cap. You guarded against exactly this three blocks above — `:4330-4335`, *"a silently truncated list would read as one clean mapping"*.
- **`n == 0` prints a confident empty map.** `span < 256 KiB` (reachable: `depth > 64` with a small stride) gives `filled=0/0` and a blank map line, reading as "the allocation is entirely empty" having probed nothing.

Minimum fix: print the probe window (`probe=512B/bucket`), report short probes the way `[slicemap]` does, and print `examined=` when the cap or `n==0` truncates. Then re-word `TOMB_RAIDER_STATUS.md:133-144` to what the instrument establishes — and reconcile bucket 0 against slice 0 in the row, because as written the two rows contradict each other and the reader has no way to tell which is right.

## BLOCKING 2 — `[readbytes]` prints a path count under a "bytes" label

`hle_file.cpp:479-482`. Argument 2 is `g_readbytes.size()` — the number of distinct paths that delivered at least one read — printed into `"%zu delivered bytes"`. The last field on the same line, `total`, genuinely is a byte count. So a run prints something of the shape:

```
[readbytes] 4210 path(s) opened, 118 delivered bytes, 4092 opened-but-never-read, 913402112 total bytes
```

and "118 delivered bytes" is not a byte count at all. This is a printed field whose label states a unit the value does not have, in a diagnostic whose entire purpose is to answer a bytes question — the first thing a reader or a doc row will quote.

## BLOCKING 3 — the per-extension line pairs an event count with a path count

`hle_file.cpp:488-489`. `by_ext[ext].first += o.second` sums **open calls**; `by_ext[ext].second` increments once per **distinct path** that was read. `:492` then prints them as a matched pair, `opened=%llu read=%llu`, and `TOMB_RAIDER_STATUS.md:147-149` already reads them that way (*"297 texture files are read"*).

Concrete failure: a title that opens 40 `.pkg` files 30 times each and reads every one of them prints `.pkg opened=1200 read=40`. A reader concludes 1,160 opens delivered nothing — the exact opposite of the truth, from the instrument built to answer "does the title read what it opens?". Make both sides count the same thing (paths, or opens with a separate read-call count), or label the units in the line.

## BLOCKING 4 — `[slicewatch]` asserts a transition it may never have observed

`live_renderer.cpp:4399-4406`. The "BECAME NON-ZERO ... **the guest filled it after our decode**" line fires on the first non-zero reading regardless of whether a zero reading ever preceded it. `said_zero` and `said_nz` are independent sets; nothing requires the zero branch to have fired first.

Concrete: an array whose probed slice already holds content at its first reference prints `BECAME NON-ZERO on reference 1 -- the guest filled it after our decode`. That is a temporal claim the instrument never made, and it is the single line that would be quoted as evidence for late streaming — the hypothesis this whole probe exists to test. Gate the "BECAME" wording on `said_zero.count(r.gpu_addr)`, and print a distinct line ("non-zero from the first reference") otherwise.

Related, non-blocking but affecting the row it supports: the probe reads 256 bytes at the slice base — 0.07% of a 352,256-byte slice. `got=256` does disclose the read size, but `TOMB_RAIDER_STATUS.md:172-176` turns it into *"it reads zero on the first reference and never becomes non-zero. So the guest never fills it, at any point"*. What was established is that the first 256 bytes of slice 248 stayed zero across ~138,000 references. Say that.

## BLOCKING 5 — an uncached `getenv` now runs on the file-IO hot path in a default run

`guest_write_watch.cpp:1705`. The block is entered on **every** call, with no switch set, and `getenv` walks `environ` each time. That function's own comment fourteen lines below (`:1724-1725`) states the standard being broken: *"Hot path: the HLE calls this before EVERY read()/pread() ... When the feature is off (the default) nothing is ever armed, so skip without even taking the lock."* Call sites confirm the frequency: `hle_file.cpp:1469` inside `read_full` (once per `read`/`pread`), `hle_file.cpp:1590` inside `disarm_iovec_watches` (once **per iovec element** of every `readv`/`preadv`), plus the live-renderer and live-compute upload paths.

This is a change to non-diagnostic behaviour with the switches unset, so it also makes the PR body's *"No behaviour change with either variable unset"* untrue as written. The cost is small per call, but the fix is one line and you already applied the right pattern in this same diff — `forced_array_layer()` at `rdna2_to_spirv_internal.hpp:1870-1880` parses once into a function-local static. Do the same here (`PROSPER_ENV_ON` cannot express the `0xADDR[:SIZE]` payload, but a `static const std::pair<uint64_t,uint64_t>` initialised by a lambda can).

Two smaller problems in the same block, both fixed by moving it below the validity guard at `:1722`:

- `:1709` evaluates `addr + size > lo` before `addr > UINT64_MAX - size` has been rejected, so the overlap test can wrap on a malformed range.
- `:1718` prints `addr - lo` as "offset in watch". A write that starts *below* `lo` and overlaps into the watch passes the test at `:1709` and prints `offset 0xfffffffffffffxxx`. The offset field is the one an investigator uses to conclude "the guest wrote slice N".

---

## Non-blocking

- `hle_file.cpp:417` says the report *"prints once at exit"*; `:449-451` and the code say it prints during the run. The stale half is the one a reader hits first. Also `:449` says "every N opens" while `:458` triggers on `g_readbytes_opens.size()` — distinct **paths**, not opens.
- `readbytes_note_read` is called before `errno` is captured, at `hle_file.cpp:1503-1504` and `:1574-1575`, in code that deliberately preserves `errno` (`:1506`, `:1577`). I could not demonstrate a clobber — `pthread_mutex_lock` returns rather than sets, and `std::map::find` does not touch it — so this is not blocking, but it inverts an invariant the surrounding lines maintain on purpose and costs nothing to fix. Move the call after `const int error = ...`.
- `hle_kernel_mem.cpp:1742`: `static uint32_t said = 0; said++` is a non-atomic read-modify-write, and `map_dmem` is reachable from more than one guest thread. You used `std::atomic` for the identical idiom at `guest_write_watch.cpp:1710` and `rdna2_to_spirv_internal.hpp:1863` in the same PR.
- `live_renderer.cpp:4382-4384`: when `layer_stride_bytes == 0` the watch falls back to `tw * th` and the printed line still says "slice %llu", so on any resource taking that fallback the named slice is not the slice addressed. Print the stride, or say it was assumed.
- `live_renderer.cpp:4318`: `char line[512]` — a long `/proc/self/maps` path splits across two `fgets` calls; the tail fragment fails the `sscanf` and is dropped, which is safe, but the head fragment is printed truncated with no newline.
- `TOMB_RAIDER_STATUS.md:143`: *"slices 186-248, which lie 48-87 MB into the allocation"* does not reproduce — at `layer_stride = 352256`, slice 186 is at 65.5 MB and slice 248 at 87.4 MB. 48 MB is slice ~136. Since this is a `## Ruled out` row, the arithmetic wants fixing or the range does.
- The PR body describes two switches. Six more (`SLICEMAPS`, `SLICEWATCH`, `OCCUPANCY`, `READBYTES`, `MAPWATCH`, `HOSTWRITEWATCH`) landed after it was written, and none of them appears in it. Per the charter's self-contained-description rule, and because BLOCKING 5 makes its verification line inaccurate, the body needs a pass before merge.
- `PROSPER_FORCE_LAYER` still does not reach the `SAMPLE_C_LZ` 2D_ARRAY lowering (`rdna2_emit_alu.cpp:7514-7545`, which routes to `image_sample_dref_manual_2d` rather than either instrumented helper). That is the 64-event class from round 1's census. It is fine to leave uncovered — but the comment at `rdna2_to_spirv_internal.hpp:1910-1913` says the scope error was covering *one* helper, and a reader will take two as complete. One clause naming what is still out.
- The worktree at `.claude/worktrees/tr-slice` carries an uncommitted 60/33 change to `prosper/src/hle/memory/hle_kernel_mem.cpp` (`attribute_dmem_allocation`) that is **not** in `3cb9a859`. I reviewed the committed head only. If that is meant to be part of this PR it needs pushing and a look; if not, it is worth committing somewhere before the tree is swept.

---

## Checked and correct

- No production behaviour change from the renderer or recompiler halves with the switches unset: all four `PROSPER_ENV_ON` sites cache, the `is_array && r.depth > 64u` prefix short-circuits, and the emitted SPIR-V operand is identical (the `bcf` emission order moves, which changes id numbering only — both orderings define before use).
- Bounds: `slice_short` is sized `slice_count` at `:5314` and every write is guarded by `slice_short_count < slice_count`; `slice_count == 0` writes nothing. The `[slicemap]` scan is bounded by `(L+1)*lsz > texture_pixels.size()` at `:5516`. The occupancy and slicewatch probes scan `k < got`, never `sizeof buf`. `probe_slice = r.depth - 8u` cannot underflow behind the `depth > 64` guard.
- No TWO-GATE on `short_reads`: `slice_short_count` is produced in the ordinary decode loop, not under any switch, so `PROSPER_SLICEMAP` alone earns the field. The pre-existing `PROSPER_ARRAY_DECODE_BUDGET_MIB` note at `:5316-5320` still holds — the budget can silence the line, never change the number on it.
- `readbytes` locking: `filelog_fd_path` releases `g_filelog_fd_mx` before `g_readbytes_mx` is taken, so there is no ordering cycle; `g_readbytes_report` is declared after the maps it uses, so it is destroyed before them.
- Nothing in the diff can arm a diagnostic from a different switch than the one whose name matches the question.

Fix BLOCKING 1-5 — 1 and 5 are the ones that matter; 2, 3 and 4 are label and wording changes — and re-word the three `## Ruled out` rows to what the probes measured, and I expect to approve.

mattias800 (commented):
**REJECTED** at head `9bb11099` — and I want to be precise about the shape of it, because it is not the shape of the last three.

**All five blocking findings are genuinely fixed.** I checked each against the source at this head, including the three things you asked me to look at specifically, and the code half of this PR is now correct. What I am rejecting is the `## Ruled out` section of `TOMB_RAIDER_STATUS.md`, which at this head asserts a claim and its negation, in the same section, with the older and more emphatic half untouched. That is a doc-only fix, and this project treats those rows as the most expensive thing in the tree.

---

## The three checks you asked for

**1. Is the occupancy scan genuinely bounded, and does the `any` early-exit mislabel a bucket?**

Bounded, yes. `bucket = 262144` and `probe.size() = 65536` divide exactly, so `want` at `live_renderer.cpp:4378-4379` is always 65536 and `std::min` never trims — the reads cover exactly `[gpu_addr + i*bucket, +262144)` and never cross a bucket edge. `scan_n <= n` and `n*bucket <= span`, so nothing reads past the declared allocation, and `copy_resource` receives at most `probe.size()` into a `std::vector` of that size. `<algorithm>` is present at `:48`.

The early exit is correct, and the ordering is the reason. `:4382-4383` scans the `got` bytes for content **before** `:4384` breaks on the short read. So a bucket whose short read returns a prefix containing content is `'#'`, not `'?'` — which is the right label, because content was actually observed. Had those two statements been in the other order, a partly-read bucket with content in the read prefix would have been marked unreadable. They are not.

The one consequence worth stating: `unreadable` (`:4387`) counts only buckets that are `!any && short_read`. A bucket that is partly readable-with-content and partly unreadable counts as filled and not as unreadable. That is the honest classification — `'#'` claims "there is content here", which a partial read already established; `'?'` claims "unknown", which such a bucket is not. `unreadable` therefore counts exactly the population whose status is genuinely unknown, which is the population the field exists for.

**2. Did moving the hostwrite block below the validity guard change when the production path returns?**

No. Before, the guard sat immediately after the diagnostic block and before `state()`; now it is the first statement, `guest_write_watch.cpp:1705`. The predicate is character-identical (`!addr || !size || addr > UINT64_MAX - size`), and nothing that was between the old guard site and the new one has any effect on the production path — the diagnostic block reads no shared state and writes only its two function-local atomics. The set of `(addr, size)` for which the function returns early, and the point at which it returns, is unchanged.

It also closes both overflow paths I raised. `addr + size > hww.lo` at `:1724` can no longer wrap, because the guard rejected `addr > UINT64_MAX - size` first — which the comment at `:1706-1710` says in as many words. And `hww.lo + hww.span` cannot wrap because the initialiser rejects `lo > UINT64_MAX - span` at `:1719`. `overlap_lo` at `:1731` is `max(addr, lo)`, so `overlap_lo - hww.lo` is non-negative by construction. `!span` is rejected too, so `…:0` disables rather than arming a zero-width watch.

The caching is the pattern I asked for: one `getenv` for the process, replacing an `environ` walk per `read`/`pread` and per iovec element with a magic-static guard load — the same shape as `forced_array_layer()` at `rdna2_to_spirv_internal.hpp:1870`.

**3. Is `filled=19/344` consistent with the `[slicemap]` reading I cited?**

Exactly consistent, and it is the number I predicted. `[slicemap]` reports slices 0-13 populated. At `layer_stride = 352256` those occupy bytes `[0, 4931584)`; at 262144 bytes per bucket that is buckets **0 through 18 inclusive — 19 buckets**, last filled 18. The new line reads `filled=19/344 … last_filled_bucket=18` and the map now opens `###` where it opened `.##`. The old `18` was short by exactly the one bucket the 512-byte head sample missed, and it was the bucket I named. With `unreadable=0` across all 344, the 325 trailing dots are now a full-byte census rather than a 0.195% probe — the row's inference is finally supported by the instrument it rests on.

---

## The other four, verified

- **`[readbytes]` "delivered bytes"** — `hle_file.cpp:482` now prints `%zu path(s) delivered bytes`, with the unit named in the comment at `:479-481`. Fixed.
- **`opened=`/`read=`** — `hle_file.cpp:498` prints `open-events=%llu paths-read=%llu`, and `TOMB_RAIDER_STATUS.md:188` quotes the new spelling. Fixed.
- **`[slicewatch]` unobserved transition** — `:4438` computes `saw_zero_first` from `said_zero.count()`, and the two format strings at `:4441-4446` are distinct. The short-circuit at `:4423` (`!nz && said_zero.insert(...)`) means `said_zero` is populated only when a zero was actually *reported*, so `saw_zero_first` is precisely "a zero reading for this address preceded this one" — which is the transition semantics. First-reference-non-zero now prints "NOT an observed transition". Fixed. The stride-fallback disclosure at `:4429-4431` / `:4450-4452` closes my non-blocking item too.
- **`%u` on `r.layer_stride_bytes` / `r.depth`** at `:4363` — both are `uint32_t` (`shader_resources.hpp:373`, `:348`), so the new `NOTHING SCANNED` line's format is right.

---

## BLOCKING — `## Ruled out` now asserts both halves of the same question

Four rows were added at the top of the section and the rows they overturn were left in place, unstruck and undated. The new rows are the *less* emphatic of the two sets, so a reader scanning the section is more likely to come away with the falsified reading.

`TOMB_RAIDER_STATUS.md:127` — *"Do not restart **the guest never fills the upper atlas**"* — and `:180-186`, *"the upper region IS written later in the run … 32 events, `selected=yes changed-during-window=yes`"*.

Against, all still present and none of them qualified:

- `:209` — *"The atlas is genuinely almost empty in GUEST memory, **for the whole run**."* This is the section heading a reader reads first.
- `:216-217` — *"it reads zero on the first reference and **never becomes non-zero**. So the guest **never fills it, at any point**"*. This is the row that most directly contradicts the new finding, and it is the one that most needs the correction: its claim of whole-run coverage comes from a 256-byte probe at one slice base, which is now known not to have seen writes the trace did see. Either the writes miss that 256-byte window or the probe misses them in time — the row cannot distinguish those and asserts neither.
- `:227` — *"the guest samples a slice that **nothing ever wrote**"*.
- `:196` — *"the atlas is **genuinely unpopulated in guest memory**"*.
- `:199` — the twelve-candidate list still carries *"stale decode cache (guest memory probed ~138,000 times, **never becomes non-zero**)"*, i.e. the falsified premise is load-bearing inside a summary row.
- `:150-153` — *"written by guest CPU stores, which **no instrument in the tree can observe**. That is what makes #3054 the blocking tool…"* — three rows below `:138-143`, which says the writer **was** observed and names it (`eboot+0xebb82..0xebbfa`).
- `:155` — the box header still reads *"contains about 4.75 MB of content **and nothing else**"*, which is verbatim the hypothesis `:123` declares falsified. The box *body* was patched with *"at the moment of decode"*; the bolded header, which is what a scanning reader takes away, was not.

Only the occupancy box got a cross-link (`:178`, "see the entry below"). The other six did not.

This is not a wording preference. The charter's rule is that these rows exist so nobody re-derives a dead answer at full cost, and that a stale row carries the document's authority — which is exactly the authority it no longer deserves. A section that says both X and not-X costs the next agent *more* than a section that said neither, because they will resolve the conflict by picking one, and the emphatic half is the wrong one.

The fix is small and mechanical: correct or strike `:150-153`, `:155`, `:196`, `:199`, `:209`, `:216-217`, `:227` to the reading the new evidence supports ("at the moment of decode", "the probe never observed it non-zero" rather than "the guest never filled it"), and re-count "Twelve candidates" at `:196` now that four rows were added and one of the twelve has been re-retired on better evidence.

While you are in there: `:128-132`'s `PROSPER_NO_TEXTURE_DECODE_CACHE` retirement and `:218-220`'s account of that same variable being **void** are compatible — different routes, `reach-gameplay.pad` versus a run that never left the title screen — but nothing says so, and the reader has to reconstruct it. One clause.

## BLOCKING (small) — a new `## Ruled out` row's reproduction needs code that is in neither this head nor master

`TOMB_RAIDER_STATUS.md:138-143` retires "the guest CPU writer cannot be observed" on the strength of *"**With both entry points sharing one attribution helper** the writer is named immediately"*. That helper is `attribute_dmem_allocation`, and it is not in `9bb11099` — `git show 9bb11099:prosper/src/hle/memory/hle_kernel_mem.cpp | grep -c attribute_dmem_allocation` returns 0, as does the same query against `origin/master`. It exists only in the uncommitted worktree change you told me to exclude.

So on the merged tree this row describes a measurement nobody can reproduce, and its own text explains why: without the shared helper, `sceKernelAllocateDirectMemory` publishes no caller chain and `PROSPER_DMEM_WRITE_TRACE` cannot arm on this title's atlas. The next reader runs the recipe at `:181`, gets the failure the row is written to retire, and concludes the instrument is broken.

One clause fixes it — name the dependency and mark it unlanded, the way `:151` already marks #3054 as the blocking tool. Same for `:180-186`, which quotes the same run.

I am not asking you to move the helper into this PR. I am asking the row to say what tree it was measured on.

---

## Non-blocking, carried over

- `hle_file.cpp:417` still says the report *"prints once at exit"*, four lines above the comment at `:449-451` explaining that it does not. The stale half is the one a reader hits first.
- `readbytes_note_read` still runs before `errno` is captured at `hle_file.cpp:1509-1510`, and the same in `f_pread`. Still not demonstrable, still one line.
- `hle_kernel_mem.cpp:1742` `static uint32_t said = 0; said++` remains non-atomic; the same idiom is `std::atomic` at `guest_write_watch.cpp:1725` and `rdna2_to_spirv_internal.hpp:1863`.
- `live_renderer.cpp:4374` — `last_filled` stays 0 when nothing is filled, printed as `last_filled_bucket=0`. `filled=0` disambiguates it, so this is cosmetic.
- The occupancy scan now reads up to 128 MiB through `copy_resource` on the first reference to a qualifying array. One-shot, behind `PROSPER_OCCUPANCY` and `done.insert(...)`, so no default cost — worth a word in the comment that the instrument is not free when armed.
- The PR body still describes two switches; six landed (`SLICEMAPS`, `SLICEWATCH`, `OCCUPANCY`, `READBYTES`, `MAPWATCH`, `HOSTWRITEWATCH`) and none appears in it. Its verification line is now accurate again after the hot-path fix, so this is presentation rather than correctness — but the charter asks for a self-contained description, and the doc rows above cite switches the body never introduces.
- `PROSPER_FORCE_LAYER` still does not reach the `SAMPLE_C_LZ` 2D_ARRAY lowering (`rdna2_emit_alu.cpp:7514-7545` routes to `image_sample_dref_manual_2d`). Fine to leave uncovered; one clause so nobody reads two helpers as complete.

## Checked and correct at this head

- No production behaviour change with any switch unset: the four renderer sites are behind cached `PROSPER_ENV_ON` plus an `is_array && r.depth > 64u` prefix; the hostwrite spec is parsed once; the recompiler operand is unchanged when `PROSPER_FORCE_LAYER` is absent.
- Bounds: occupancy reads are bucket-clamped and allocation-clamped (above); `slice_short` writes remain guarded by `slice_short_count < slice_count`; `[slicemap]`'s scan is bounded at `:5563`; `probe_slice = r.depth - 8u` cannot underflow behind the `depth > 64` guard; `[slicewatch]` scans `k < got`, not `sizeof buf`.
- `agc_shader_layout.cpp` citation at `TOMB_RAIDER_STATUS.md:133-137` verified: `view.base += static_cast<uint64_t>(d.base_array) * stride;` and `view.layer_stride = stride;` are where the row says, inside `thin_2d_layered` and `d.depth > 1`. The row is accurate.
- `scripts/tomb-raider-PPSA16901/reach-gameplay.pad` exists, so `:129`'s route is citeable.
- No TWO-GATE introduced: every new field is produced by the same switch that prints it, or by ungated code.

Fix the section so it stops saying both things, and mark the one row whose evidence needs unlanded code, and I will approve on sight. The code is done.

mattias800 (commented):
**REJECTED** at head `0cd0fb09` — narrowly, on three sentences the sweep turned up. You were right to ask for one rather than trust the list: your seven are all correctly fixed, and there are three more.

**The code is unchanged and stays verified.** `git diff 9bb11099 0cd0fb09 -- prosper/src prosper/frontends prosper/tools prosper/tests` is empty; the only file touched is `TOMB_RAIDER_STATUS.md`. Nothing in my round-4 code verification is disturbed, and I did not rebuild.

**Your seven are fixed, and the self-correction is the best thing in this commit.** I re-derived it rather than taking it: bucket 200 is byte 52,428,800, which at `layer_stride=352256` is slice **148.84**; slices 186 and 248 begin at buckets **249.94** and **333.25**. So the probe does sit below the interior's range and never touched it, exactly as `:128-131` now says. Catching that your own row overstated its evidence, and writing the overstatement down rather than quietly narrowing it, is the behaviour the `## Ruled out` convention exists to produce.

`:155`, `:150`, `:209`, `:216`, `:227`, `:196` and `:199` all now carry their scope, and the unlanded-reproduction disclosure at `:149-153` is exactly right — naming `attribute_dmem_allocation`, `#3054`, and the specific failure signature (`caller-chain=0`) so a failed arm is not read as a contradiction.

---

## BLOCKING — three unscoped statements survive, one of them the mirror of the row you just fixed

### 1. `:195-201` still carries the exact sentence `:131-133` retracts

The row headed **"The allocation is NOT static, and the upper region IS written later in the run"** is the untouched duplicate of the corrected row at `:123-133`, and it repeats every part of the overstatement:

- `:195` — *"the **upper region** IS written later in the run"*
- `:197` — *"the **upper atlas** takes 32 events"*
- `:198-199` — *"So **"the guest never fills the upper atlas" is false**"*

That last sentence is verbatim what `:132-133` now says *"overstated a probe taken below the range in question."* So the document retracts the sentence at line 133 and asserts it as a current finding at line 199, sixty-six lines later.

It matters more than its twin, because it sits inside the occupancy narrative rather than in the `## Ruled out` list — a reader working through the measurement story reaches it without ever seeing the retraction. What they take away is "the upper region does get written", which is the one conclusion that would stop them looking for a content path into slices 186-248, i.e. the actual open frontier.

Fix: apply the same qualification, or replace the row body with a pointer to `:123-133` so there is one statement of this result instead of two.

### 2. `:261-264` asserts the whole-atlas form twice

Inside **"Nor is it the slice LAYOUT"**:

- `:261` — *"Every layout tried leaves the same invariant: **only about 5 MB of the atlas's 67 MB is non-zero**, whichever way the slices are addressed."* No decode-time scope, and "whichever way the slices are addressed" reads as a claim about the allocation rather than about one instant.
- `:263-264` — *"the remaining question is not how prosper reads the atlas but **why most of it was never written**"*.

"Most of it was never written" is the whole-atlas claim your own `:123-127` row falsifies. The row's actual verdict — that the layout hypothesis is dead — is unaffected; it is the trailing inference that needs the clause.

### 3. `:85-86`, in **Open defects**, says the atlas holds **84 MB non-zero**

> *"The atlas is bound and holds real data (**84 MB non-zero**), nothing is rejected, and the menus are fully textured."*

This is the flat opposite of this PR's central result — `filled=19/344` of an 86 MiB span, ~4.75 MB — stated undated, unscoped, and **above the fold**, in the summary section a reader reaches forty lines before `## Ruled out`.

I checked its provenance before raising it: it is pre-existing on `origin/master` and landed in `227f5fbe` (#3006), so this PR did not introduce it. I am raising it anyway, for the reason the charter gives: a PR that falsifies a claim records the falsification before merging, and this PR's whole empirical contribution is the falsification of that number. Leaving it in the summary reproduces exactly the "asserts both halves" state that round 4 rejected — at a worse pair of locations, because this one is a contradictory *figure*, not an unscoped variant, and it is the first thing about #2998 a reader meets.

I am not asking you to re-litigate #3006. One clause — superseded, with a pointer — or, if "84 MB" measured something other than the atlas's non-zero content (the dmem allocation as a whole, say), one clause saying which. If it turns out that figure is right and the occupancy census is wrong, that is a much bigger finding than anything in this PR and should be said out loud.

---

## Non-blocking

- **`:261` says "the atlas's 67 MB" where `:123`, `:168` and `:176` say 86 MiB.** Both are defensible and they measure different things — 256 × 262,144 = 67.1 MB is the mip-0-only footprint, 256 × 352,256 = 86.0 MiB is the full `layer_stride` span — but a reader meeting two spans for one atlas in one document will assume one is wrong. Name which footprint each is.
- Carried over from round 4, all still open and all still minor: the stale *"prints once at exit"* comment at `hle_file.cpp:417`; `readbytes_note_read` running before the `errno` capture at `hle_file.cpp:1509-1510` and in `f_pread`; the non-atomic `said++` at `hle_kernel_mem.cpp:1742`; the PR body still describing two switches when six landed; and `PROSPER_FORCE_LAYER` not reaching the `SAMPLE_C_LZ` array lowering (`rdna2_emit_alu.cpp:7514-7545`).

## Sweep coverage, so you know what "clean" covers

I read the whole file (354 lines), not only the seven sites. Searched the full text for `never wrote` / `nothing ever wrote` / `never fill` / `never written` / `unpopulated` / `almost empty` / `genuinely empty` / `reads zero` / `whole run` / `at any point` / `nothing else` / `no instrument`, and separately for every numeric occupancy figure (`N MB non-zero`, `N MB of`, `filled=`). Thirteen hits in the first sweep and eight in the second; ten are the scoped forms you added, three are the ones above. I also checked the sections above `## Ruled out` (where hit 3 came from) and the tail past line 300, which is clean.

Also verified this round: the arithmetic at `:128-131` and `:170` reproduces exactly (149 / 249.9 / 333.2); `:176`'s `filled=19/344 … last_filled_bucket=18` remains consistent with `[slicemap]`'s slices 0-13 (buckets 0-18 inclusive = 19); `diff --check`-relevant whitespace is clean in the diff; the change is genuinely docs-only.

Fix those three sentences and I will approve. Nothing else is outstanding — the code has been done since `9bb11099`.

mattias800 (commented):
**APPROVED** at head `40906450`.

All three round-5 findings are fixed, my independent sweep is clean, and the code has been correct since `9bb11099`.

## Scope of this approval

`git diff 9bb11099 40906450 --name-only` returns exactly one path, `prosper/docs/TOMB_RAIDER_STATUS.md`, so my round-4 code verification transfers intact and I did not rebuild or re-run anything. The approval covers the whole PR at this head: `live_renderer.cpp`, `rdna2_to_spirv_internal.hpp`, `hle_file.cpp`, `hle_kernel_mem.cpp`, `guest_write_watch.cpp`, `TOMB_RAIDER_STATUS.md`. Per the repo's merge rule this is content-based — if anything is pushed after `40906450`, re-establish by content before merging.

## The three findings

**1. `:204-215`, the narrative duplicate.** Now headed *"filled progressively, and a region ABOVE the decode-time content is written later"*, with the arithmetic inline and an explicit `**Read the scope.**` block. The retraction is stated where the mistake was made — *"An earlier version of this bullet concluded … that conclusion is retracted"* — rather than only in the twin row. I re-derived the numbers again at this head: bucket 200 = byte 52,428,800 = slice **148.84**; slices 186/248 begin at buckets **249.94**/**333.25**. The row's figures are right.

Your read of the shape is right too, and worth keeping: one copy corrected, one not, no check able to see the difference. That is the clean-merge-broken-file failure, and a doc is where it is hardest to catch because nothing goes red.

**2. `:271-283`.** The invariant is bounded with *"at decode time"*, the remaining question is narrowed to *"why the slices the interior samples were never written"*, and the footprint clause names 67 MB as mip-0-only (256 × 262,144) against the 86 MiB full-stride span. Both figures check out, and telling the reader to name which they mean is the right general fix rather than picking one.

**3. `:87-92`, the withdrawal — and no, it does not overreach.** I checked the provenance rather than taking it: `227f5fbe` added *"holds real data (84 MB non-zero)"* as a bare parenthetical, and that commit's subject and body are entirely about unannounced 32-bit index buffers and never mention an atlas census. So *"arrived without a recorded method"* is accurate, and nothing in the tree supports the number.

Withdrawal is the correct disposition, for three reasons: the competing measurement is stronger on every axis that matters (named instrument, quoted raw output, stated scope, reproducible switch, `unreadable=0`); the note asserts only that what the figure measured *is not established*, which is the weaker and more defensible claim than saying it was wrong about something specific; and removing the explanation you could not support was the right call, not an excess of caution. *"Guessing is how the head-sample figure survived"* is a fair reading of this PR's own history — the 512-byte head sample went three rounds because a plausible-looking number was never asked to show its window.

One observation, deliberately **not** a request to put anything in the doc: the atlas span is 90,177,536 bytes = 86.0 MiB = 90.2 MB, which is close enough to "84 MB" that the figure may have been the allocation *size* rather than its non-zero content. That is a guess, it is exactly the kind of tidy mechanism your note warns against recording, and it belongs in #2998 as a lead for whoever re-derives it — not in the file. Keep the doc as it is.

## Sweep

I treated your list as unreliable and the file as the authority, as asked. Four passes over all 371 lines:

- **Phrasings** — `never wrote` / `nothing ever wrote` / `never fill` / `never writ` / `unpopulated` / `almost empty` / `genuinely empty` / `is empty` / `reads zero` / `mostly unwritten` / `mostly empty` / `whole run` / `at any point` / `nothing else` / `nothing above` / `no instrument` / `entire` / `all 256` / `every slice`. Sixteen hits: ten are the scoped forms, four are retraction or supersession markers, two are unrelated (vertex format at `:345`, frame stall at `:350`).
- **Numbers** — every `N MB/MiB/KiB`, `filled=`, `N/256`, `N/344`, `non-zero`. Nineteen hits, all mutually consistent; `4.75 MB` / `86 MiB` / `filled=19/344`, the 65.5-87.4 MB slice range, and the 67 MB / 86 MiB pair all reconcile.
- **Propagation** — `84 MB` across `prosper/docs/` and `COMPATIBILITY.md`: the only surviving instance is the withdrawal note itself. Nothing else in the tree carries it.
- **Bolded claim headers**, since those are what a scanning reader takes away. Thirteen matched the atlas/slice/memory vocabulary; the four in `## Ruled out` name the dead hypothesis, which is the convention, and the rest are scoped or unrelated.

No unscoped form survives.

## One nit, genuinely optional

`:88` and `:175` write **4.75 MB** where `:130` writes **~4.75 MiB**. 19 × 262,144 = 4,980,736 bytes = 4.75 **MiB** = 4.98 MB, so the digits pair with MiB and `:88`'s *"4.75 MB of 86 MiB"* mixes the two within one phrase. A 5% slip, no conclusion depends on it — flagged only because `:275-277` now tells readers to name which footprint they mean, and the same discipline applies to the prefix. Fix it in passing or leave it; it does not gate the merge.

Per the reviewer's role I did not re-run your build, `ctest` (311/311 at `9bb11099`), `check_diag_gates.py`, or `check_numbered_table.py` — those are yours, and the code they cover is unchanged since I verified it. What I re-derived independently this round is the doc arithmetic, the `227f5fbe` provenance, the emptiness of the code diff, and the sweep above.

Good sequence of revisions. The instrument-honesty work — full-byte census over head sample, `?` distinct from `.`, `open-events`/`paths-read`, the transition gate, the parse-once hot path — is the part that will outlast this investigation, and the two self-corrections you volunteered (the bucket-200 range, and pulling the unverified explanation out of the withdrawal) are worth more than the findings I raised.

Issue #2045, opened 2026-08-05
hle_calls --values cannot see out-structs, and four more callsite scans use the weak module predicate


mattias800:
Two follow-ups spun off from PR #2035 (#1997, #2000), both bounded and both found by independent review of that PR.

## 1. `--values` records the return register, never an out-struct

#1997 asked for two things: the return value **and** "optionally the first N bytes of the out-struct an argument points at". PR #2035 landed the first. The second is the half that matters most on this codebase, because the recurring bug shape here is *success returned, out-struct never written* — `sceSystemServiceGetStatus` (aliased to the wrong handler), `sceSystemServiceGetDisplaySafeAreaInfo` (ratio left 0), `sceNpEntitlementAccessGetSkuFlag` (unregistered, out pointer untouched), `sceNpTrophy2GetTrophyInfoArray` (garbage count -> a 34 GB allocation). Every one of those returns `0` and is therefore **invisible** in a `--values` census.

That is not hypothetical: it is the stated residue of the init audit on #1905. A 49-handler census came back with no implausible *return value*, which bounds the candidate only as far as the return register reaches.

Suggested shape: `--out-bytes N` plus a way to say which argument is the pointer — either a per-handler table or, simpler and more honest, "dump N bytes at whichever of a0/a1 looks like a writable guest pointer, before and after the call, and print only when they differ". A before/after diff is self-validating in the way this tool's other fields are: a handler that wrote nothing is then visibly distinct from one this tool failed to observe.

Verification it must carry, in the same spirit as the existing ones: a positive control whose out-struct is known from the source before the run (`sceSystemServiceGetStatus` writes 12 bytes with `st[6]=1`), and a way to tell "nothing was written" from "nothing was read".

## 2. Four more guest-callsite scans still use `guest_va_in_module`

#2000 fixed two sites in `hle_file.cpp`. The same reasoning applies verbatim to four more, and **two of them are stronger cases than the ones already fixed**, because the recovered address is used as a **correlation key** rather than only printed — a stub address there changes chain identity, not just a log line:

- `prosper/src/hle/hle_kernel_mem.cpp:238`
- `prosper/src/hle/hle_kernel_mem.cpp:2601`
- `prosper/src/hle/hle_kernel_time.cpp:984`
- `prosper/src/hle/hle_agc.cpp:822` — this is also the unclamped "fence1-builder precedent" that `apr_miss_callsite`'s own comment cites as the model, so it is worth clamping to `stack_words_above()` while there.

[...]

Issue #1755, opened 2026-08-02
PROSPER_DMEM_CALLER's stack walk can read past the top of a guest thread stack and SIGSEGV


mattias800:
## Summary

`PROSPER_DMEM_CALLER=1`'s stack walk (`hle_kernel_mem.cpp`, `log_dmem_caller`) reads up to 160 qwords
*above* the current frame with no bound. When the allocating guest frame sits near the top of its
thread stack, the walk runs off the last mapped page and the process dies with a SIGSEGV that looks
like a guest fault.

## Reproduction

Observed while using the tool on Asterix &amp; Obelix - Babylon Mission (`PPSA30490`, #1748). Raising
the scan (a local experiment with the same code shape, 2048 slots) faults immediately:

```text
=== RUN ENDED: kind=2  SIGSEGV at addr=0x7f90d7e00000  rip=... ===
  rax=0x7f90d7e00000 rsp=0x7f90d7dff9b0
  backtrace: STUB+0x48fd <- eboot+0x7d10c9 <- eboot+0x7cd7ee <- eboot+0x16a1e8c
```

`fault_addr - rsp` is **0x650 = 1616 bytes**. The shipped `kScan = 160` reads 1280 bytes above the
frame, so this exact frame is within ~336 bytes of faulting, and any shallower outermost frame does
fault. It is not a corner case: this was the *first* direct-memory allocation of the run.

## Suggested fix

Probe each new page before touching it and stop at the first unmapped one — `mincore()` on POSIX,
`VirtualQuery` on Windows (the file carries two copies of this helper, one per platform). A local
patch of that shape made the 2048-slot walk safe and produced 32-frame chains with no fault:

```cpp
const uint64_t pg = (uint64_t)(uintptr_t)&frame[i] & ~(uint64_t)(page - 1);
if (pg != checked_page) {
    unsigned char vec = 0;
    if (mincore((void*)(uintptr_t)pg, (size_t)page, &vec) != 0) break;
    checked_page = pg;
}
```

While there, the depth/scan are worth exposing (`PROSPER_DMEM_CALLER_DEPTH` /
`PROSPER_DMEM_CALLER_SCAN`): the default 6 frames stop inside the allocator and do not reach the
subsystem that asked, which is the question the tool exists to answer. A 32-frame chain on #1748
reached the AGC submit path and named the defect.

`file:line` — `prosper/src/hle/hle_kernel_mem.cpp` `log_dmem_caller` (both the Linux and the Windows
[...]

PR #3055, opened 2026-08-27
hle(#3054): publish caller chains from sceKernelAllocateDirectMemory, not only the Main variant

### LINKED ISSUE #3054: PROSPER_HWWATCH cannot arm on an absolute guest address, only a register-relative one
`PROSPER_HWWATCH` takes a **delta relative to a register** (`PROSPER_HWWATCH_REG`, default `rax`), so it can only watch an address computed at a fault site. There is no way to arm a data watchpoint on a **fixed guest address** from process start.

**The failure scenario is an investigation that cannot proceed, rather than wrong output.** #2998 needs the guest code that writes a specific texture allocation: the title fills about 4.75 MB of an 86 MB atlas and samples far beyond it, and the question is what bounds the filling loop. Catching the writer needs a watchpoint on an address known up front — which is exactly what the current interface cannot express. Every other instrument in the tree sees stores through prosper's own paths (PM4, compute, DMA); a plain guest CPU memcpy is invisible to all of them, and a hardware watchpoint is the only thing that would catch it.

**This is an interface gap, not a capability one.** The mechanism is already implemented: `perf_event_open` with `PERF_TYPE_BREAKPOINT`, `bp_type` write, `exclude_kernel=1` (`exec_image_linux.cpp`, the `perf_bp_open` helper). That combination is permitted at `perf_event_paranoid=2`, which is this machine's setting. What is missing is a way to pass an absolute address instead of a register plus delta.

**Suggested shape:** accept `PROSPER_HWWATCH=0xADDR` when the value parses as an address inside the guest range and no `PROSPER_HWWATCH_REG` is set, keeping the existing register-relative form otherwise. The report should name the guest PC and the module-relative offset, as the register-relative path already does.

Found while investigating #2998, where it is the blocking tool gap.

mattias800:
Fixes #3054. Investigation context: #2998.

## The defect

`PROSPER_DMEM_WRITE_TRACE` selects an allocation by **caller-chain + exact allocation size +
offset**. Only `sceKernelAllocateMainDirectMemory` ever minted a chain id, so a title that
allocates through `sceKernelAllocateDirectMemory` published **none** — and no selector could name
a single byte it owns. The trace was not merely weak for those titles; it was structurally
unreachable, and it failed by refusing to arm rather than by saying why.

## Failure scenario

*Tomb Raider I-III* (`PPSA16901`) makes two 1 GiB `alloc_dmem` allocations, and its 86 MiB texture
atlas lives inside the second. Before this change a run with `PROSPER_DMEM_CALLER=1` produced
**zero** `[dmem-caller]` records for that title, so every `PROSPER_DMEM_WRITE_TRACE` configuration
— any chain, any offset — failed to arm. There was no combination of inputs that worked.

## Why the original issue was wrong, and why this is not that fix

I opened #3054 asking for an absolute-address watchpoint, on the reasoning that the trace "cannot
arm on a fixed address" and the title sub-allocates inside its own 1 GiB block. **Both halves were
wrong:**

- the four-field selector's `offset` field already targets an arbitrary sub-range, which is exactly
  the case I claimed it could not express;
- the fixed-VA rejection in `guest_write_watch.hpp` is a **deliberate, correct** design decision —
  a guest VA is run-local, so the absolute mode I proposed would have silently armed on whatever
  occupied that address on a later run, the same failure the charter records for capture hashes and
  operation ordinals. Building it would have made the instrument *less* trustworthy.

The real gap was one allocator that never reported.

## Approach and invariants

Both entry points now share one `attribute_dmem_allocation` helper.

**The interner is a single static on purpose.** A second interner would issue ids from its own
sequence, so one id would mean two different call sites depending on which API allocated — and a
configured trace would then arm on the **wrong memory** instead of failing to arm, which is
strictly worse than the bug being fixed.

**The allocator label is now data, not a constant.** It was the literal `alloc_main_dmem`, true
while only that path published chains; left alone, every record from the new path would have named
the wrong allocator. An empty label reports `<unknown-allocator>` rather than borrowing the
default, so a record can never assert an allocator nobody supplied.

Applied to the **Linux and Windows** implementations alike, so the diagnostic is not silently dead
on one platform.

## Result

```
[memhle]      alloc_dmem len=0x40000000 -> phys=0x10000
[dmem-caller] caller-chain=1 alloc_dmem len=0x40000000 from eboot+0xcc58 eboot+0x20c7c8 ...
[memhle]      alloc_dmem len=0x40000000 -> phys=0x40010000
[dmem-caller] caller-chain=2 alloc_dmem len=0x40000000 from eboot+0xccb3 eboot+0x20c7c8 ...
```

With chain 2 nameable, `PROSPER_DMEM_WRITE_TRACE=2:0x40000000:0x51cc000:128` arms
(`target=0x20491cc000 target-phys=0x451dc000 pages=1`) and identifies the atlas writer:
`eboot+0xebb82..0xebbfa`, an unrolled AVX scatter storing 8x16 B per iteration to offsets computed
in `%ymm10`. That answers a question #2998 had been stuck on.

## Deliberately unaffected

Nothing changes with `PROSPER_DMEM_CALLER` unset: the attribution call sits behind the same
`dmem_caller_log()` gate the Main path already used, and MEMLOG's historical lines are
**byte-for-byte unchanged** on both paths. No production allocation behaviour is touched — the
helper runs after the allocation has succeeded and its result has been written to the guest.

## Risks

The helper walks the guest stack, as the Main path already did; it inherits the same `#1755`
mapped-end clamp and the `#2045` `guest_va_in_module_code` filter, unchanged. It runs only when the
diagnostic is enabled. The chain-id space is shared across two allocators now, so a run using both
will interleave ids — intended, and the reason for the single interner.

## Verification (exact head `f67d58be`)

- `cmake --build prosper/build-linux -j6` — clean
- `ctest --test-dir prosper/build-linux --no-tests=error -j4` — **311 of 311 passed**
- `check_diag_gates.py` — all checks passed
- `diff --check` — clean
- Live: `PPSA16901` with `PROSPER_DMEM_CALLER=1 PROSPER_MEMLOG=1` — output above; the same run
  before this change produced zero `[dmem-caller]` lines.

**Mutation-verified tests.** The two new arms in `test_dmem_caller_chain.cpp` are not merely
passing: restoring the hardcoded `alloc_main_dmem` literal makes both **FAIL** and the binary
report `FAILED`, and they pass again when reverted. Without that check the arms would have been
assertions that could not distinguish the fix from the bug.

No snapshots were run: this touches no rendering path.

### DIFF
--- a/prosper/docs/TOMB_RAIDER_STATUS.md
+++ b/prosper/docs/TOMB_RAIDER_STATUS.md
@@ -155,10 +155,10 @@ Measured layout, for anyone extending this: `layer_stride = 352256` for the 512x
   a byte it owns. With both entry points sharing one attribution helper the writer is named
   immediately: `eboot+0xebb82..0xebbfa`, an unrolled AVX scatter storing 8x16 B per iteration to
   offsets computed in `%ymm10`.
-  **Reproduction needs code that is NOT on master.** `attribute_dmem_allocation` is not in this PR
-  and not in `origin/master`; it is tracked in **#3054**. Until it lands, the recipe below reports
-  `caller-chain=0` for this title and the trace refuses to arm — which is the very failure the row
-  retires, so do not read a failed arm as contradicting it. (#2998, #3054)
+  **Reproduction needs #3054**, which made `sceKernelAllocateDirectMemory` publish a caller chain.
+  Before it, this title produced no `[dmem-caller]` records at all and the trace refused to arm for
+  any configuration — which is the very failure this row retires, so on an older tree do not read a
+  failed arm as contradicting it. (#2998, #3054)
 
 - **Nothing maps into the atlas per-slice, and no host/kernel write streams into it** — both with
   controls, both new instruments. `PROSPER_MAPWATCH` sees one 1 GiB `map_dmem` covering the atlas,
--- a/prosper/src/hle/memory/dmem_caller_chain.hpp
+++ b/prosper/src/hle/memory/dmem_caller_chain.hpp
@@ -44,9 +44,13 @@ inline void append_dmem_caller_chain_number(std::string& line, uint64_t value, i
 // sceKernelAllocateDirectMemory publishes them too, a hardcoded label makes the record state the
 // wrong source for its own allocation -- and a diagnostic that misreports where it came from is
 // worse than one that says nothing.
+//
+// Deliberately NOT defaulted. A default is the same silent borrowing the empty-string guard below
+// exists to prevent, just moved to the call site: a third allocator that omitted the argument
+// would mislabel every record it produced, and nothing would say so.
 inline std::string format_dmem_caller_chain_definition(
         uint32_t id, uint64_t len, const DmemCallerChainFrame* frames, size_t frame_count,
-        int scanned_slots, int requested_slots, const char* api = "alloc_main_dmem") {
+        int scanned_slots, int requested_slots, const char* api) {
     std::string line;
     line.reserve(256);
     line += "[dmem-caller] caller-chain=";
@@ -85,8 +89,7 @@ inline void write_dmem_caller_chain_line(FILE* stream, std::string_view line) {
 
 inline void write_dmem_caller_chain_definition(
         FILE* stream, uint32_t id, uint64_t len, const DmemCallerChainFrame* frames,
-        size_t frame_count, int scanned_slots, int requested_slots,
-        const char* api = "alloc_main_dmem") {
+        size_t frame_count, int scanned_slots, int requested_slots, const char* api) {
     const std::string line = format_dmem_caller_chain_definition(
         id, len, frames, frame_count, scanned_slots, requested_slots, api);
     write_dmem_caller_chain_line(stream, line);
--- a/prosper/src/hle/memory/dmem_caller_chain.hpp
+++ b/prosper/src/hle/memory/dmem_caller_chain.hpp
@@ -39,14 +39,21 @@ inline void append_dmem_caller_chain_number(std::string& line, uint64_t value, i
 
 // Build the whole definition before taking the output lock. Besides keeping the lock hold short,
 // this makes it impossible for another first-seen chain to splice its ID or frames into this line.
+// `api` names the allocator this chain was interned for. It used to be the literal
+// "alloc_main_dmem", which was true while only that entry point published chains; once
+// sceKernelAllocateDirectMemory publishes them too, a hardcoded label makes the record state the
+// wrong source for its own allocation -- and a diagnostic that misreports where it came from is
+// worse than one that says nothing.
 inline std::string format_dmem_caller_chain_definition(
         uint32_t id, uint64_t len, const DmemCallerChainFrame* frames, size_t frame_count,
-        int scanned_slots, int requested_slots) {
+        int scanned_slots, int requested_slots, const char* api = "alloc_main_dmem") {
     std::string line;
     line.reserve(256);
     line += "[dmem-caller] caller-chain=";
     append_dmem_caller_chain_number(line, id, 10);
-    line += " alloc_main_dmem len=0x";
+    line += ' ';
+    line += (api && *api) ? api : "<unknown-allocator>";
+    line += " len=0x";
     append_dmem_caller_chain_number(line, len, 16);
     line += " from";
     for (size_t i = 0; i < frame_count; ++i) {
@@ -78,9 +85,10 @@ inline void write_dmem_caller_chain_line(FILE* stream, std::string_view line) {
 
 inline void write_dmem_caller_chain_definition(
         FILE* stream, uint32_t id, uint64_t len, const DmemCallerChainFrame* frames,
-        size_t frame_count, int scanned_slots, int requested_slots) {
+        size_t frame_count, int scanned_slots, int requested_slots,
+        const char* api = "alloc_main_dmem") {
     const std::string line = format_dmem_caller_chain_definition(
-        id, len, frames, frame_count, scanned_slots, requested_slots);
+        id, len, frames, frame_count, scanned_slots, requested_slots, api);
     write_dmem_caller_chain_line(stream, line);
 }
 
--- a/prosper/src/hle/memory/hle_kernel_mem.cpp
+++ b/prosper/src/hle/memory/hle_kernel_mem.cpp
@@ -745,15 +745,22 @@ namespace {
         const uintptr_t slots = (hi - base) / sizeof(uint64_t);
         return slots < (uintptr_t)want ? (int)slots : want;
     }
-    void log_main_dmem_allocation(uint64_t len, uint64_t phys,
-                                  const volatile uint64_t* frame) {
-        // Preserve the unarmed path: no stack query, interner construction, or extra formatting unless
-        // PROSPER_DMEM_CALLER was explicitly requested. MEMLOG alone keeps its historical line byte-for-byte.
-        if (!dmem_caller_log()) {
-            MLOG("alloc_main_dmem len=0x%llx -> phys=0x%llx\n",
-                 (unsigned long long)len, (unsigned long long)phys);
-            return;
-        }
+    // #2998: allocation attribution, shared by every direct-memory entry point.
+    //
+    // Only sceKernelAllocateMainDirectMemory used to build a caller chain. A title that allocates
+    // through sceKernelAllocateDirectMemory instead therefore produced no chain id at all, which
+    // made PROSPER_DMEM_WRITE_TRACE structurally unable to name a single byte it owns -- the
+    // selector is keyed on caller-chain plus allocation size, and neither was ever published for
+    // that API. Tomb Raider I-III takes exactly this path (two 1 GiB allocations), so the trace
+    // could not be pointed at its texture atlas however it was configured. That read as "the
+    // instrument cannot target a sub-allocation" when the truth was "this allocator never reports".
+    //
+    // The interner is one shared static on purpose: a second interner would issue ids from its own
+    // sequence, so one id would mean two different call sites depending on which API allocated, and
+    // a configured trace would silently arm on the wrong memory rather than fail to arm.
+    DmemCallerChainResult attribute_dmem_allocation(uint64_t len, uint64_t phys,
+                                                    const volatile uint64_t* frame,
+                                                    const char* api) {
         constexpr int kWant = 6;          // return addresses to report
         constexpr int kScan = 160;        // stack slots to walk
         // Never read past the mapped end of this stack (#1755).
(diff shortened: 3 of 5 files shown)


mattias800 (commented):
## Independent review — head `f67d58be`

**APPROVED.** The diagnosis is right, the fix is minimal and correctly placed, and I could not find a
way to make it go wrong at runtime. Findings below are all non-blocking; two of them are documentation
lines that become *false* the moment this merges, and I'd like those in this PR rather than tracked.

I did not build or run tests — author verification stands. Everything below is from reading.

---

### What I verified (so the record says what was actually checked)

**1. Unarmed path — the main risk. Clean.**
`hle_kernel_mem.cpp:814` still short-circuits `log_main_dmem_allocation` before any stack query,
interner construction or formatting, and the new site is behind the same gate at `:1672` / `:6376`.
The `alloc_dmem` MEMLOG line at `:1666` and `:6374` is untouched and is emitted *before* the new
code, so its bytes cannot change. With `PROSPER_DMEM_CALLER` unset the only added instruction on
either path is one already-cached `getenv` predicate.

**2. No MEMLOG line is lost in any interner state.** This is the part of the refactor that could
silently have dropped a census record, so I traced every exit: `attribute_dmem_allocation` returns
`correlation` at all four exits (`:785`, `:791`, `:798`, and the fall-through at `:807`), and
`log_main_dmem_allocation:824-844` then runs its full switch unconditionally. `chains.intern` never
returns `Disabled`, so `dmem_caller_chain_correlates_allocation` is always true and the `Known` /
`Unknown` / `Overflow` MEMLOG spellings are byte-identical to the pre-PR versions. Confirmed.

**3. Length semantics match the census line.** I checked this specifically because a mismatch would
break the selector silently: the write-trace compares `size != trace.config.allocation_size`
(`guest_write_watch.cpp:1506`), and a user configures that from the number they read in the MEMLOG
line. `dmem_take` takes `sz` **by value** (`:974`, `:3930`), so `sz == a2` and `[memhle] alloc_dmem
len=…` and the notify agree. Good.

**4. Gating the notify on `dmem_caller_log()` loses nothing.** `guest_write_watch.cpp:1409` already
invalidates the trace outright when `PROSPER_DMEM_CALLER` is unset, so there is no configuration in
which a notify suppressed by that gate could have mattered.

**5. Ordering change (`[dmem-caller]` before `[memhle]`) — acceptable.** Both go to stderr, both are
single stdio calls, and they were never atomically adjacent under concurrency anyway. If anything the
new order is better for a reader: the id is defined before its first use. I grepped `tools/` and
`docs/` for anything that parses this pair and found no consumer — only the prose example noted in
finding D. Not a defect.

**6. Windows copy.** `sed`-extracted both helper bodies (`:747-806` vs `:3716-3775`) and diffed them:
**byte-identical**, no copy divergence. `:6377` uses `__builtin_frame_address(0)`, matching its own
neighbour `k_alloc_main_dmem` at `:6395` and every other frame-walk in that block. `HLE` expands to
`static PROSPER_SYSV_ABI` in both blocks (`:670`, `:3613`), so the convention is the same one already
in production there. Compiles-plausibly: `attribute_dmem_allocation` is in the same anonymous
namespace and declared far above both call sites, and `dmem_caller_log` is defined per-block (`:693`,
`:3662`).

**7. Stack-walk safety — both guards preserved, no OOB.** `dmem_caller_scan_slots(frame, kScan)` at
`:766-767` keeps the #1755 mapped-end clamp, and the `guest_va_in_module_code` filter with its #2045
comment survives at `:772-774`. Bounds: the loop is `i = 1; i < scan` (`:770`), and `scan <= (hi - base)/8`,
so the last read is `base + 8*(slots-1) < hi`. `scan <= 0` degenerates to zero iterations. `n <=
kWant` bounds `frames[kWant]`. `Known` implies `n >= 1` (n == 0 feeds `intern(0,0)`, which returns
`Unknown` and exits before the frame loop), so no zero-frame definition can print. Starting at slot 1
matches the convention this file already uses for the same purpose at `:1969`.

**8. Shared interner — the argument holds, and the static really is single.** `#if defined(__linux__)
|| defined(__APPLE__)` at `:626`, `#else` at `:3521`, `#endif` at `:7277`: the two copies are mutually
exclusive, so exactly one `static DmemCallerChainInterner<> chains` (`:778`) exists per build, and
both `k_alloc_dmem` and `k_alloc_main_dmem` reach it through one function. The "two interners would
collide" reasoning is correct and, given the selector keys on the id, the failure it avoids really is
the worse one (arming on the wrong memory rather than failing to arm). Also worth noting the PR did
not accidentally break the lazy-construction property: the static is still only reached behind a gate.

One consequence the PR body already flags and I agree is fine: ids now interleave across two
allocators, so a chain number recorded from a pre-PR run may not reproduce
(`ASTERIX_BABYLON_STATUS.md:391-393` records `caller-chain=1`/`2`). That is not a contract break —
`tools/AGENTS.md:53` already states ids are "stable only within one process" — so no recorded evidence
is invalidated as a matter of contract.

**9. Test arms are genuinely discriminating for the mutation claimed.** Pinning the label back to a
literal fails arm 1 (`test_dmem_caller_chain.cpp:91-96`, exact-string) *and* arm 2 (`:98-102`, the
`<unknown-allocator>` find). Arm 2 also discriminates the more likely half-fix `api ? api :
"alloc_main_dmem"`, which would emit a double space and no sentinel. Both arms are sound.

---

### Findings

**A. (substantive, non-blocking) The new arms test the formatter, not the fix.**
`prosper/src/hle/memory/hle_kernel_mem.cpp:1672-1674` and `:6376-6378` are the entire bug fix, and
deleting all six lines leaves the suite green — both new arms call
`format_dmem_caller_chain_definition` directly and never touch the wiring. So the regression that
would reintroduce #3054 verbatim is uncovered, which is the shape the charter warns about: the
assertion is true and the mechanism under test never ran.

This is closable with machinery already in the tree, which is why I'm raising it rather than waving
it: `guest_write_watch.hpp:232-240` documents `guest_dmem_write_trace_configure` as "the
deterministic test seam", `guest_dmem_write_trace_snapshot()` exposes `allocation_matches`, and
`tests/host/test_dmem.cpp:80` already invokes `sceKernelAllocateDirectMemory` through
`Hle::lookup(nid_hash(...))`. An arm that sets `PROSPER_DMEM_CALLER=1` before the first
`dmem_caller_log()` call in a dedicated binary, configures a selector for the size it is about to
request, calls the handler, and asserts `allocation_matches == 1` fails on master and passes here.
Your call whether it belongs in this PR or a follow-up, but please don't let it evaporate.

**B. (non-blocking) The defaulted `api` parameter contradicts this PR's own honesty argument.**
`dmem_caller_chain.hpp:49` and `:89` keep `const char* api = "alloc_main_dmem"`. The empty-string case
is guarded and the PR adds a test at `test_dmem_caller_chain.cpp:98-102` arguing that "an absent label
must be visibly absent rather than silently borrowing the default" — but the *omitted-argument* case
does exactly the silent-borrowing the guard exists to prevent, and it is the easier mistake: `api` is
the 8th parameter and the previous signature had 7. Failure scenario: someone wires attribution for a
third allocator (the AMM give-direct-memory path at `:2889` is the obvious next one) and calls
`write_dmem_caller_chain_definition(stderr, id, len, frames, n, scan, kScan)` — every record from that
allocator then states `alloc_main_dmem`, and a reader chasing an allocation attributes it to an API
that was never called. Both production call sites already pass explicitly, so dropping the default and
updating the one test arm at `:79-86` to pass `"alloc_main_dmem"` costs two lines and removes the trap.

**C. (minor) `api` reaches `%s` unguarded in the Unknown branch.**
`hle_kernel_mem.cpp:794-797` (and the Windows twin at `:3763-3766`) passes `api` straight to `%s` with
no null check, while the header's definition path guards it. Unreachable today — both call sites pass
literals — but it is an inconsistency in a helper explicitly written to be shared by future callers,
and `printf("%s", nullptr)` is UB (glibc prints `(null)`; MinGW's runtime is not guaranteed to).
Cheapest fix is the same expression the header uses.

**D. (docs — please land here) `tools/AGENTS.md` is wrong-by-omission after this merges.**
`:43` says the diagnostic prints "the guest return addresses above each
`sceKernelAllocateMainDirectMemory`", which is precisely the belief this PR exists to kill — a reader
would conclude `alloc_dmem` still publishes nothing and not try it. `:47-48` also shows the example in
the now-reversed line order and only the `alloc_main_dmem` label. This is the discoverability half of
the fix; without it the next agent hits #3054's symptom and re-derives the cause.

**E. (docs — please land here) `docs/TOMB_RAIDER_STATUS.md:158-161` becomes actively false.**
That paragraph reads "**Reproduction needs code that is NOT on master.** `attribute_dmem_allocation`
is not in this PR and not in `origin/master`… Until it lands, the recipe below reports
`caller-chain=0` for this title and the trace refuses to arm". On merge, the caveat inverts: an agent
reading it will skip a reproduction that now works, or misread a failure. It is exactly the
stale-figure-carrying-authority failure the charter documents. The rest of that `## Ruled out` row is
already correct and needs no change — just retire the caveat and note the landing PR.

---

Nothing above blocks. Fix C at your discretion, please fold in D and E, and either add A here or file
it before this merges.

**APPROVED**

mattias800 (commented):
## Re-review — head `bb81ff76`

**APPROVED.** B, C, D and E are all correctly fixed, nothing new was introduced, and the code diff
against master is byte-identical to the one I approved at `f67d58be` apart from those four deltas.
Filing A as #3057 rather than rushing it is the right call, and the issue is accurate — details below.

---

### The thing you asked me to check specifically: does the test actually compile?

**Yes — confirmed independently, and the confirmation is discriminating.** This is worth spelling out
because "it appeared to pass" is exactly what happened to you, so a check that could also pass for the
wrong reason would be no better.

First, every call site repo-wide, since removing the default turns any omission into a hard error.
There are eight, and all eight pass the label explicitly:

- `dmem_caller_chain.hpp:93-94` — the internal forward from `write_…` to `format_…`
- `hle_kernel_mem.cpp:806` (Linux) and `:3775` (Windows) — production
- `test_dmem_caller_chain.cpp:80-82`, `:92-93`, `:100-101`, `:126-129`, `:139-142` — five, not the
  two the delta shows, because `:92` and `:100` already carried labels

Only two translation units include the header at all (`hle_kernel_mem.cpp:16`,
`test_dmem_caller_chain.cpp:1`), so that list is closed — no third TU can be sitting stale.

Then an actual compile of the test TU, standalone, at the project's `CMAKE_CXX_STANDARD 20`:

```
$ g++ -std=c++20 -fsyntax-only -I prosper/src \
      prosper/tests/hle/memory/test_dmem_caller_chain.cpp
rc=0
```

Zero diagnostics, g++ 16.1.1. **Positive control**, because a green `-fsyntax-only` proves nothing on
its own: I copied the test to scratch, deleted the `"alloc_main_dmem"` argument from the `:80` call
site — the precise mutation that bit you — and recompiled:

```
error: too few arguments to function 'std::string prosper::format_dmem_caller_chain_definition(
       uint32_t, uint64_t, const DmemCallerChainFrame*, size_t, int, int, const char*)'
mutant rc=1
```

So the check reddens on the failure mode and passes on your tree. Your earlier green was a stale
binary; this one is not, because it never consults a binary.

---

### The four fixes

**B — default removed, and the reasoning is recorded.** `dmem_caller_chain.hpp:53` and `:92` now take
`const char* api` with no default, and the comment at `:47-50` states *why* — that a default is the
same silent borrowing relocated to the call site. That comment is the part that makes this stick: the
next person to add a parameter here will read it before reaching for a default. Resolved exactly as
raised.

**C — guarded, and the two platform copies stayed in lockstep.** `hle_kernel_mem.cpp:797` and `:3766`
now use `(api && *api) ? api : "<unknown-allocator>"`, the same expression as the header's `:59`, so
the Unknown branch and the definition branch cannot disagree about what an absent label prints. I
re-diffed the two helper bodies at this head (`:760-808` vs `:3729-3777`): still **byte-identical**.
Worth noting for the record that this is the one class of change a Linux build cannot catch — the
Windows block is `#if`-excluded — so the parity diff is the check that covers it, and it is clean.

**D — `tools/AGENTS.md:42-55`** now names both entry points, says why the gap mattered (#3054, the
trace could not name any byte it owned), states that records name their own allocator, and shows an
`alloc_dmem` line beside the Main one at `:51-54`. The example pair is also now in emission order.
That closes the discoverability half.

**E — `docs/TOMB_RAIDER_STATUS.md:158-161`** is now past tense and scoped to older trees
("Reproduction needs #3054, which made … so on an older tree do not read a failed arm as
contradicting it"). It merges true instead of merges false, and it still preserves the warning for
anyone on a pre-#3054 checkout, which is the part that was actually load-bearing.

---

### A / #3057

I checked the wall you hit rather than taking it on trust, because a filed issue carrying wrong
reasoning is worse than no issue. **It is real:** `guest_write_watch.cpp:892` is
`if (!config.caller_chain || !config.allocation_size || !config.size || …)` inside
`trace_config_valid`, and `guest_dmem_write_trace_configure` returns false on it — so a chain that
interns as `Unknown` (id 0) genuinely cannot be selected, and `allocation_matches` is unreachable from
a test. Your fallback is sound for the same reason: `intern(0, 0)` returns `first == true` for the
first Unknown chain, so the record does print with no guest frames. #3057 states all of this
correctly and the recipe would work.

One sharp edge to add to the issue, not a blocker here: `unknown_seen_` is a **one-shot per process**,
not per allocator. If anything in the harness performs a dmem allocation that interns as Unknown
before the arm under test, `first` is false and no line prints at all. The issue's step 1 already
prescribes a dedicated binary and first-call ordering for the `dmem_caller_log()` static cache, which
happens to mitigate this too — it is just worth naming the second reason so a later reader does not
relax the ordering and get an empty capture. It fails safe (the assertion goes red, not green), so it
would confuse rather than mislead.

---

Nothing outstanding. **APPROVED** at `bb81ff76`.

Issue #2952, opened 2026-08-23
Metaphor: ReFantazio (PPSA20800) — the 63 frames it now publishes are black


mattias800:
## Summary

*Metaphor: ReFantazio* (`PPSA20800`) publishes real frames for the first time since #2934 was
fixed — **63 of them, at 34.8 fps** — and every one is black.

## Measurement

Frontend `tools/screenshot`, default route, no pad input, `PROSPER_RENDER=1
PROSPER_GUEST_ARGS=-force-gfx-direct`, Linux / AMD Radeon 8060S (RADV STRIX_HALO),
`--seconds 5 --count 24 --timeout 170 --allow-guest-fault --no-stop-after-guest-fault`.

All 24 PNGs are byte-identical (one md5, one CRC `d0ef2d73`, `source-distinct=1
pixel-distinct=1`). Sampling every 37th pixel of one 3840x2160 frame gives **6,093 of 6,137
samples pure black**, and the 44 distinct colours present are all near-black (`0x161314`,
`0x150e05`, `0x020202`). The frontend reports `5 distinct of 63 published`.

## What this is NOT yet

**Not established as #2932.** That issue is a wrong composite — a flat *white* 4K clear or a
near-black diagonal wedge — on titles that are alive and producing frames throughout. This title
dies of a SIGFPE (#2951) about five seconds in, so all 24 samples are the same retained frame
(`max-pixel-stale=115.0s`) and there is no population to characterise. A near-uniform black could
equally be "the title never drew anything before it died".

Separating them needs the guest to survive. Sequence the SIGFPE first.

## What is known about the render path on this title

From the same run, with `PROSPER_RENDER=1`:

- The renderer comes up normally: `decoded texture cache budget = 4096.0 MiB`, persistent
  colour-target and texture residency budgets 4096 MiB each, host-buffer pool 2048 MiB.
- Compute adopts the renderer's device, with descriptor indexing, int64 atomics and the native
  subgroup contract all inherited.
- **One** compute program is skipped: `0x2281c74100`, `reason=recompile-reject
  mode=unresolved-operand pc=58 words=bf880060`. One rejected program out of the run's whole
  population — worth implementing on its own terms, not established as related to the black frames.
- Two diagnostics fire that may matter once the title survives:
  - `[gpu] vk_color_format: unmapped CB surface format=0xc number_type=7 comp_swap=1 -> Undefined`
[...]

Issue #2057, opened 2026-08-06
Crisis Core (PPSA07809): the title screen's key art never renders — text and prompt draw, the logo/character layer is flat black


mattias800:
## Summary

With the #1945 race throttled away (see `prosper/docs/CRISIS_CORE_STATUS.md`), *Crisis Core –Final
Fantasy VII– Reunion* now reaches its title screen. The screen is only **partly** drawn: the
`Press Any Button` prompt with its blue sweep, the `Ver. 1.0.4` string and the two copyright lines
all render correctly, but the CRISIS CORE / FF7 REUNION logo and character key art that fill the
left two-thirds of the screen on hardware are entirely absent — that region is flat RGB(0,0,0).

The committed capture is `assets/screenshots/crisis-core-title.png`; the missing region is
everything left of about x=1250 (of 3840).

## Reproduction (Linux, in the `ps5ys` distrobox)

```bash
PROSPER_GUEST_ARGS= PROSPER_NULL_PAGE=1 PROSPER_RENDER=1 PROSPER_MB3_POISON=1 \
  PROSPER_PAD_SCRIPT=@prosper/scripts/crisis-core/reach-title.pad \
  ./build-linux/screenshot <DUMP_ROOT>/PPSA07809-app0 \
      --seconds 1 --count 90 --out <OUT> --timeout 180
```

The title screen holds indefinitely, so any sample from roughly t=37 s onward that is not the
blue/magenta composite (see the sibling issue) shows it. In one 90-sample run, samples 37, 40, 49,
52 and 75 all carried it, each with the same missing region.

## What is and is not known

- The screen is **stable** — the same missing region in five samples spanning 38 s, so this is not
  a transient load-order artifact.
- Both rendered elements (prompt, version/copyright text) are Slate/UI. The absent element is the
  one that would be a streamed texture, which makes an unresolved or wrongly-decoded texture the
  first thing to check — but no evidence has been gathered either way, so treat that as a
  hypothesis, not a finding. `CONFIDENCE: LOW`.
- No `[recompile-reject]`, no `[compute] … skipped` and no unsupported-format lines appear in these
  arms.
- The F9/`gpu_replay` loop — the normal way to answer "which draw should have written these
  pixels" — is **unavailable on this title** while #2027 stands.

Blocks rung 5 (hardware-oracle comparison) for this title.

Refs #1894

PR #3059, opened 2026-08-27
tools: frameclass — find a HUD drawn over a world that never rendered

mattias800:
A small tool that answers **"is anything drawn in this frame, is any of it legible, and is it only a HUD?"** over a directory of captured frames.

The class worth having is **`UI-ON-BLACK`**: legible content covering under 2% of the frame — a HUD, notice or logo over a world that never drew. That is the signature of a title that executes but does not render, and it is invisible to every "is the picture changing?" metric we have.

## Why it exists

Two hand-rolled versions of this are wrong in ways that read as a result, and both were built here first:

- **Downsampling before counting colours destroys thin UI.** A 4K frame at 160x90 loses white menu text on black entirely. Measured on *Stray* (`PPSA02101`): a frame containing START GAME / SETTINGS / CREDITS and a legible build stamp reported `colours=61 FLAT-BLACK mean=(0,0,0)`. The title was one edit from being written up as not rendering.
- **"Fraction of non-black pixels" scores a flat WHITE clear as a perfect frame** — 100% non-black, and every is-it-black test passes it.

`tools/screenshot`'s manifest already carries `distinct_rgb_colors` and `nonblack_rgb_pixels` over every pixel, in the same directory as the PNGs — read those and skip this tool if they are all you need. What this adds is the peak channel value, the classification, and working on directories with no manifest.

## The classes

| class | meaning |
| --- | --- |
| `FLAT` | nothing rendered, or a uniform fill: peak ≤ noise, or ≤ 8 colours covering ≥ 50% |
| `SPARSE` | nothing rises above near-black (peak < 48), whatever the coverage |
| `UI-ON-BLACK` | legible content covering **under 2%** — a HUD or notice over an absent world |
| `LIT` | something legible is drawn over more than 2% of the frame |

Over the 129 committed frames in `assets/screenshots/` as of 2026-08-27: **LIT=124, UI-ON-BLACK=4, SPARSE=1, FLAT=0**. That corpus grows — re-run rather than quoting the split. The four `UI-ON-BLACK` frames are `sonic-frontiers-cyberspace-hud` (0.81%), `sonic-frontiers-autosave-notice` (1.05%), `metaphor-loading-mascot` (1.49%) and `beast-of-reincarnation-game-freak-logo` (0.38%) — the first three being exactly the titles recorded as "world black behind the HUD" (#2790) and "the background does not draw" (#2952). **That control was chosen by the reviewer, not by me.**

## What this tool deliberately does NOT claim

`LIT` is not a promise that the frame is a game scene, and **no threshold here can make it one**. An earlier revision of this PR claimed a "clean gap" in the corpus at 10% coverage. Review showed the gap was an artifact — it came from taking the minimum over frames a *previous revision of the same tool* had labelled as scenes, which is the assumption under test. Opening the frames settles it:

| frame | coverage | colours | what it actually is |
| --- | --- | --- | --- |
| `crisis-core-title.png` | 10.13% | 4096+ | text and a glow on black (key art missing, #2057) |
| `stray-brightness-calibration.png` | 10.13% | 228 | a settings menu on black |
| `messenger-title.png` | 12.41% | 36 | real pixel art |
| `oregon-trail-gameloft-splash.png` | 16.03% | 2159 | a flat logo on black |
| `blue-prince-title.png` | 21.83% | 4096+ | real rendered 3D art |

Coverage puts a logo above a scene; colour count puts pixel art below a logo. UI and scenes interleave on both. So the class that claimed to make that call is gone, and both the docstring and `AGENTS.md` say plainly that deciding "is the world rendering?" above 2% needs eyes on the frame. The tool narrows which frames need them.

## Verification

`--selftest` builds **11 frames pixel by pixel**, deliberately not sampled from `assets/screenshots/` — the thresholds were tuned against that corpus, so a control drawn from it would confirm the tuning rather than the classifier, which is precisely how the "clean gap" came to be believed.

Each rule is checked by mutation; every mutant reddens exactly the case that names it:

| mutant | selftest |
| --- | --- |
| `uniform_fill` bound back to 0.90 | **10/11** — letterbox bars → `LIT` |
| `uniform_fill` on the bare colour count | **6/11** — hud, fps overlay, title, both fades → `FLAT` |
| `SPARSE` keyed on `bright == 0` instead of the peak | **10/11** — dim full-frame picture → `SPARSE` |
| `UI-ON-BLACK` widened back to 10% | **10/11** — lit title screen → `UI-ON-BLACK` |

Also verified by running them: an unreadable file, a truncated PNG and a non-image each report on their own line, keep the tally, and exit 0; `.bmp` directories work (F9 grabs); and the Windows double-enumeration fix was checked under forced case-insensitive glob semantics (1 row, not 4).

## Caveats, both documented and neither detectable from pixels

- `--fps-overlay` burns prosper's own counter into the PNG, so a black frame carrying only that overlay classifies `UI-ON-BLACK`. The selftest **asserts** this wrong-looking answer so the documented false positive cannot drift silently away from the behaviour.
- A smooth gradient with no picture in it classifies `LIT` — prosper's seed-miss gradient is exactly that shape.

## Files

`prosper/tools/frameclass/frameclass.py`, `prosper/tools/frameclass/AGENTS.md`, and a bullet in `prosper/tools/AGENTS.md`.


mattias800 (commented):
**REJECTED** on two blocking findings. I want this tool — the central insight is right, `UI-ON-BLACK` fires correctly on real project evidence, and I reproduced your validation. But the *label* is the product here, and on this project's own reviewed corpus it is wrong 12% of the time, in the under-reporting direction.

Reviewed at head `aa7f00d4`. Everything below was run, not reasoned about.

## What I reproduced (your validation stands)

Four of your five rows reproduce exactly:

```
black-clear.png               max=  0  nonblack=  0.00%  colours=   1  FLAT
white-clear.png               max=255  nonblack=100.00%  colours=   1  FLAT
gta5-prologue-bank-lobby.png  max=255  nonblack= 99.59%  colours=4096  CONTENT
run3/..._13.png               max=255  nonblack=  0.18%  colours= 172  UI-ON-BLACK
```

That last row is the Stray main menu, and it matches your PR body to the digit. The unbound row differs (I get 86.12% against your 85.04%) because I used the committed `assets/screenshots/unbound-worlds-apart-intro-cinematic.png` and you evidently used a different frame — same verdict, so no finding, but name the file next time so the check is re-runnable.

**The anecdote is substantiated.** I ran the old downsampling classifier (`~/unbound-lane/classify.py`) on the exact frame: `colours=61 FLAT-BLACK mean=(0, 0, 0)`, `--- 0/1 frames carry content ---`, and `1/30` over the whole run. So the failure mode is real and your fix is real.

**The stride is sound on the case that motivated it**, which is the part I most expected to break. On `run3/..._13.png`:

```
step=1: n=8294400  max=255  nonblack=0.1785%  bright=12115  colours=262
step=2: n=2073600  max=255  nonblack=0.1798%  bright= 3050  colours=251
step=4: n= 518400  max=255  nonblack=0.1823%  bright=  807  colours=172
step=8: n= 129600  max=255  nonblack=0.1806%  bright=  204  colours= 56
```

`nonblack%` is stable to three decimals and `max=255` survives every stride. A nearest-neighbour stride is categorically different from `resize()`, which box-filters white text into black, and the numbers show it.

**And `UI-ON-BLACK` independently finds the known rung-2 signature on committed evidence I chose, not you:** `sonic-frontiers-cyberspace-hud.png` (0.81%), `metaphor-loading-mascot.png` (1.49%), `sonic-frontiers-autosave-notice.png` (1.05%). Those are exactly the titles `CLAUDE.md` describes as "world black behind the HUD" (#2790) and "background does not draw" (#2952). That is a positive control drawn from outside your source set, and it passes.

---

## BLOCKING 1 — `SPARSE` swallows 15 of 121 committed progression screenshots, and its documented meaning is wrong for 14 of them

I ran the tool over the entire committed corpus, `assets/screenshots/*.png` — 121 human-reviewed frames, every one of them accepted as milestone evidence:

```
CONTENT=102  SPARSE=15  UI-ON-BLACK=4  total=121
```

The `SPARSE` bucket:

```
arcrunner-title-screen.png             max=255  nonblack= 8.62%  bright= 21441  colours=1765
gta5-title.png                         max=255  nonblack= 7.15%  bright= 32793  colours=1921
forgotten-city-title.png               max=255  nonblack= 7.64%  bright= 12733  colours=4096
sonic-frontiers-sonic-team-logo.png    max=255  nonblack= 8.22%  bright= 42047  colours=4096
oregon-trail-health-warning.png        max=255  nonblack= 5.16%  bright= 24466  colours= 247
metaphor-language-select.png           max=255  nonblack= 5.90%  bright=  5815  colours=1559
sonic-crossworlds-sega-logo.png        max=255  nonblack= 5.17%  bright= 26701  colours= 348
frontiers-middleware-credits.png       max=255  nonblack= 5.65%  bright= 26450  colours=1828
sonic-frontiers-sega-logo.png          max=255  nonblack= 6.46%  bright= 14207  colours= 740
little-nightmares-3-boot-splash.png    max=255  nonblack= 4.16%  bright= 19956  colours= 256
little-nightmares-3-eula.png           max=255  nonblack= 3.27%  bright= 14109  colours= 254
little-nightmares-3-title-screen.png   max=255  nonblack= 2.40%  bright= 10342  colours= 256
pathless-title.png                     max=255  nonblack= 2.45%  bright=  5518  colours= 186
arcrunner-title-screen-default-route.png  max=255  nonblack=8.62%  bright=19247  colours=1672
arcrunner-default-route-logos.png      max= 15  nonblack= 0.15%  bright=     0  colours=  32
```

`frameclass.py:32` defines `SPARSE` as *"some non-black content, nothing bright -- a fade, or a very dark scene"*. Fourteen of those fifteen have `max=255` and thousands to tens of thousands of bright pixels. The definition is false for all but the last row.

Worse, five of them — `gta5-title`, `arcrunner-title-screen`, `metaphor-language-select`, `little-nightmares-3-title-screen`, `forgotten-city-title` — are cited in `COMPATIBILITY.md` **and** `BLOG.md` as the rung-1/rung-2 milestone for their titles. A title screen on a dark background sitting at 2-9% coverage is not an edge case in this project; it is the single most common shape of frame bring-up produces, because rungs 1 and 2 *are* logos and title screens.

The mechanism is the dead band at `frameclass.py:61-63`: `UI-ON-BLACK` needs `share < 0.02`, `CONTENT` needs `share > 0.10`, and nothing claims 2-10%. So `SPARSE` is not a class, it is the residue — and the residue is where this project's evidence lives.

This is your own failure mode with a different threshold. The PR body says you had already written *"0/8 frames carry content"* before boosting a frame and finding a menu. A lane pointing this at an ArcRunner run and reading `SPARSE` writes the same sentence.

I constructed one more instance by hand, outside the corpus: a real full-colour 3D scene composited into a 5% viewport on black (`max=255`, `colours=4096`) reports `SPARSE`. A title rendering its world into a wrongly-sized viewport is a plausible bring-up state and it reads as "a fade".

Fix is yours to choose, but it needs to be one of: add a bright-coverage arm so `bright > 0` with meaningful coverage reaches `CONTENT`; drop the `CONTENT` floor toward 2%; or rename `SPARSE` to something that cannot be read as "no content" and correct its docstring. Whichever you pick, re-run the 121-frame corpus — it is committed, free, and it is the only oracle here that nobody tuned against.

## BLOCKING 2 — the `<= 8 colours` rule is unguarded and fires *before* `UI-ON-BLACK`

`frameclass.py:60`:

```python
if mx <= noise or len(cols) <= 8:      kind = "FLAT"
```

No coverage guard, no brightness guard, and it short-circuits ahead of the `UI-ON-BLACK` test at line 61. So any monochrome bright UI on black is `FLAT` — the same label as a black clear.

Demonstrated with prosper's **own** annotation. `tools/screenshot/overlay_text.hpp:41-42` gives `--fps-overlay` an opaque white foreground and a translucent black backing box; over a black frame that composites to exactly two colours. A black 4K frame carrying nothing but the overlay:

```
black-frame-with-fps-overlay-only.png  max=255  nonblack=1.18%  bright=6120  colours=2  FLAT
```

Six thousand bright pixels, `max=255`, reported `FLAT`. Two more hand-built instances: a loading screen (black + white logo + progress bar, 4.70% coverage, `bright=24361`) → `FLAT`; a full-screen 8-colour 2D title covering 87.59% of the frame at `max=255` → `FLAT`. Tiny-palette 2D titles are this project's staple — Alex Kidd, Sonic Origins, The Messenger.

I could not produce a real instance from the 121-frame corpus, so I am labelling this honestly as constructed. I still call it blocking because the fix is one clause and the failure mode is precisely the one the tool was built to eliminate.

**The two findings interact, and this is the part not to miss.** Right now the `<= 8` rule *accidentally masks* the `--fps-overlay` trap: the overlay frame above escapes `UI-ON-BLACK` only because the bitmap font is 2-colour. Fix B2 naively — move the colour test after `UI-ON-BLACK`, or gate it on coverage — and a `--fps-overlay` black frame starts reporting **UI-ON-BLACK**, i.e. "the guest is drawing its menu and the world is missing" when the only thing drawn is prosper's own fps counter.

That is live, not hypothetical: `tools/screenshot/screenshot.cpp:747-753 and :784-800` computes the manifest metrics from the pristine frame and burns the overlay into a copy **that the PNG encoder sees**, and `screenshot.cpp:955-958` warns that every PNG the run wrote carries it. frameclass reads the PNG, so it reads the annotation as content. The escape hatch already exists — the manifest records `run.assertions.fps_overlay` (I confirmed `fps_overlay":false` in the Stray run's JSONL) — so either consult it, skip the `overlay_box` region, or at minimum document the trap.

---

## Non-blocking

**N1 — "always exits 0" is false.** `frameclass.py:35` and `AGENTS.md:24` both state it, and the PR body leans on it ("so it cannot break a run"). Measured: no args → 1; empty directory → 1; missing file → 1 (traceback); non-image file → 1 (traceback); Pillow absent → 1. The one that will actually bite: **a single corrupt or truncated PNG anywhere in a directory aborts the whole run with an `UnidentifiedImageError` traceback and loses the tally line**, discarding every row already computed. A killed `tools/screenshot` run leaves exactly such a file. Wrap `classify` and emit an `ERROR` row, or correct the claim in both places.

**N2 — "sampled at full resolution" (`AGENTS.md:18`, docstring `:19`) overclaims.** It is a 1/16 nearest-neighbour stride at native resolution. That distinction is the whole basis of the tool, so it is worth stating precisely — especially since it is *defensible*, per the step sweep above. It is also not lossless: a black frame with 1px vertical lines at `x % 4 == 1`, covering 25% of the screen, reports `max=0 colours=1 FLAT`, indistinguishable from a black clear. Contrived, but it fixes the shape of the claim.

**N3 — `colours=4096` is a saturation marker, not a count.** `frameclass.py:58` stops inserting at 4096. Nothing in the output or the docs says so, so a reader comparing two `4096` rows learns nothing. Suffix a `+`.

**N4 — the seed-miss gradient trap is not carried over.** A synthetic gradient with no picture in it classifies `CONTENT` (`nonblack=100%`, `colours=4096`). `tools/snapshot/profile_route.py:28` already records this: *"a seed-miss diagnostic gradient can score MORE distinct colours than real content, so the frames have to be looked at."* A new colour-count classifier inherits that trap and should repeat the warning; per `CLAUDE.md`, a falsification that lives in one doc and not the one you would actually read is one the next agent re-derives at full cost.

**N5 — the stated boundary against `tools/screenshot` is wrong, and it is the AGENTS.md claim I most want corrected.** `AGENTS.md:6-8` says screenshot answers only "is the picture changing". It also already computes two of your three numbers, at **full** resolution with no stride: `capture_manifest.cpp:100-114` emits `distinct_rgb_colors` and `nonblack_rgb_pixels` per sample, and they are written to a JSONL sitting *in the same directory as the PNGs you re-read*. For the Stray frame the manifest says `distinct_rgb_colors: 262, nonblack_rgb_pixels: 15235` where you compute `172` and `0.1823%`. `tools/snapshot/snapshot.py:254` does the same on BMPs.

What is genuinely new is `max` and `bright` — neither exists anywhere in the tree — plus the classification. That is a real addition and I am not arguing the tool away. But say so, and consider reading the JSONL when it is present: you would get exact numbers for free. Also note your `nonblack` predicate (`max(r,g,b) > 8`) differs from the manifest's (`rgb != 0`), so the two must not be compared directly.

**N6 — `AGENTS.md:32` tells the next agent to "Re-run that check if you change a threshold", and there is nothing to re-run.** No selftest, no test, and the Stray control lives in `~/stray-lane/` — lane scratch that this project's own worktree/scratch discipline removes. The nearest sibling, `tools/colorstate/colorstate_report.py`, ships `--selftest` that needs no capture, and `tools/AGENTS.md:440` advertises it. Four of your five controls are already free: two are synthesizable in three lines, and `assets/screenshots/gta5-prologue-bank-lobby.png` and `unbound-worlds-apart-intro-cinematic.png` are committed. Add `--selftest`; the tool is a classifier whose whole argument is that unvalidated classifiers are worthless.

**N7 — two small doc inaccuracies.** `AGENTS.md:13` says the frames were reported as *"flat black, 1 colour"*; the old classifier actually printed `colours=61` for the menu frame — it read `FLAT-BLACK` because `black > 0.98`, not because of the colour count. The `1 colour` rows were the genuinely black frames. Substance holds, figure does not, and this repo is strict about restated numbers. Separately, `AGENTS.md:3-4` says "the three cases" for a four-class tool.

**N8 — directory mode is PNG-only** (`frameclass.py:72`), so pointing it at an F9 grab directory gives `frameclass: no PNGs found` and exit 1 — verified on `~/unbound-lane/grab`. `CLAUDE.md` makes F9 the primary loop for graphical bugs and it writes `.bmp`. An explicit `.bmp` path works fine, so this is a one-line glob change.

**N9 — checked, not a bug.** `frameclass.py:44` uses `.convert("RGB")`, which *discards* alpha rather than compositing over black, while `capture_manifest.cpp:105-108` multiplies RGB by alpha. The screenshot PNGs are mode `RGBA`, so I expected a divergence. Measured on the real Stray frame: **identical** on every number, because `normalize_capture_rgba` (`capture_manifest.cpp:90-97`) forces alpha to 255 for composited and republished frames — 0 of 8,294,400 pixels had `alpha != 255`. The one path that keeps alpha untouched is `CaptureSource::RawScanout`, which I have no sample of. Reporting the negative so nobody re-derives it.

**N10 — no bullet in `prosper/tools/AGENTS.md`.** That file is the tool index (`:381`, `:421`, `:433` and so on); a new folder that is not listed there is not discoverable, which is the map's whole job.

## Also fine

`im.load()` / `PixelAccess` is the correct API and sidesteps the deprecation entirely: Pillow 12.3 warns *"Image.Image.getdata is deprecated and will be removed in Pillow 14"* on the old ad-hoc script and says nothing about the committed tool, which runs clean under `-W error::DeprecationWarning`. Performance is a non-issue — 121 4K frames in 11.2s, 30 in 3.7s. `sorted(glob.glob(...))` is safe because `screenshot.cpp:467` already zero-pads the index to the width of the largest. No build-system change needed. No existing code touched.

Happy to re-review promptly on B1 and B2. The 121-frame corpus is the right oracle for the retune, and it is sitting in the repo.

mattias800:
All twelve findings addressed. Head is `b204b45c`. The two blocking ones were both real defects in the thing the tool sells, and the review's own numbers are what fixed them.

## BLOCKING 1 — the dead band

You were right that the label is the product, and right about the direction. Your 15 `SPARSE` rows made the boundary obvious once I pulled per-class statistics off the same corpus:

| | coverage |
| --- | --- |
| the 14 lit frames you flagged (titles, logos, EULAs, menus) | 0.15 – **8.62%** |
| least-covered real `CONTENT` frame | **10.13%** |

There is a clean gap, and the old thresholds put the boundary inside it twice — `CONTENT` at 10%, `UI-ON-BLACK` at 2%. `UI-ON-BLACK` now takes everything below 10%, which is the class those frames belonged in all along: they *are* UI over an absent world. Corpus is now **CONTENT=109, UI-ON-BLACK=18, SPARSE=1, FLAT=0**. The surviving `SPARSE` is `arcrunner-default-route-logos.png` — `max=15`, 0.15%, `bright=0` — which is genuinely unreadable, so the class now means what the docs say it means.

## BLOCKING 2 — the `<= 8 colours` rule, and the interaction you told me not to miss

`uniform_fill` now requires the few colours to **cover ≥ 90%** of the frame. A black clear, a white clear and a two-tone letterbox stay `FLAT`; the fps-overlay frame no longer short-circuits.

I did not try to detect the overlay from pixels — it isn't detectable, and pretending otherwise would be worse than the trap. So it is documented as a **known false positive** in both the docstring and `AGENTS.md`, with `run.assertions.fps_overlay` named as the after-the-fact check, and **the selftest asserts the wrong-looking answer** (`fps overlay only → UI-ON-BLACK`) so the caveat cannot drift away from the behaviour silently.

Fixing B2 surfaced a third ordering bug you hadn't seen, because B1 was masking it: with the colour rule constrained, `bendy-title.png` (93.33% covered, 2,899 colours, `max=127`) fell to `SPARSE` — legibility was being tested before coverage. Coverage now decides `CONTENT` first. A dim picture is still a picture.

## N6 — `--selftest`

Nine frames, built pixel by pixel. Deliberately **not** sampled from `assets/screenshots/`: the thresholds are tuned against that corpus, so a control drawn from it would confirm the tuning rather than the classifier. Each of the three fixes is checked by mutation — reverting it reddens the case that names it, and only that case plus its collateral:

| mutant | selftest |
| --- | --- |
| `uniform_fill` back to the bare colour count | **5/9** — menu text, lit title, fps overlay, near-black fade all → `FLAT` |
| `UI-ON-BLACK` bound back to 2% | **8/9** — lit title screen → `SPARSE` |
| legibility tested before coverage | **7/9** — both full-frame pictures → `SPARSE` |

## The rest

**N1** — an unreadable file is now reported on its own line and counted, not fatal. Verified against a truncated PNG *and* a non-image, in a directory with a good frame: both report, the good one classifies, the tally prints, exit 0. The claim is also now correctly scoped — 0 for a completed classification, non-zero only for the tool failing.
**N5** — corrected, and it is the one I most needed. `AGENTS.md` now says outright that `distinct_rgb_colors` and `nonblack_rgb_pixels` are already in the manifest at `capture_manifest.cpp:100-114`, in the same directory as the PNGs, and tells the reader to skip this tool if that is all they need. What is new is narrowed to `max`, `bright`, the classification, and working without a manifest.
**N2** — "full resolution" replaced with "1/16 nearest-neighbour stride at native resolution", plus your aliasing case (1px lines at `x % 4 == 1` reading as a black clear).
**N3** — saturated counts print `4096+`, with a note that two such rows aren't comparable, nor is one against the uncapped manifest figure.
**N4** — seed-miss gradient trap carried into both the docstring and `AGENTS.md`.
**N7** — `colours=61` not "1 colour" (the `1 colour` rows were the genuinely black frames); "three cases" → four classes.
**N8** — directory mode globs `.bmp` too, so an F9 grab directory works.
**N9** — recorded as a negative in `AGENTS.md`, with `CaptureSource::RawScanout` named as the one path nobody has a sample of.
**N10** — bullet added to `prosper/tools/AGENTS.md`, next to `self_dump/`.

Thank you for running the old classifier on the Stray frame and for checking `UI-ON-BLACK` against evidence you picked rather than evidence I picked — the Sonic/Metaphor control is the reason I trusted the class enough to widen it rather than retreat from it.

mattias800 (commented):
**REJECTED**, narrowly, and mostly on documentation honesty rather than on code. Re-reviewed at head `4de5d1a8`. Both original blocking findings are genuinely fixed and I verified them independently, including the mutation discrimination. What stops it is a third-round finding of the same shape as the first two: the sentence that justifies the new threshold is false, and I can name the two frames that falsify it.

Everything below was run at head, on the corpus at head, not reasoned about.

## What I verified as fixed

**BLOCKING 1 is fixed.** Reproduced the corpus split myself — note the corpus is now **129** frames, not 128, because master added one between your run and this head:

```
CONTENT=110  SPARSE=1  UI-ON-BLACK=18  total=129        (yours: 109/18/1 of 128)
```

All fifteen frames I flagged last time now land where I argued they belonged: the fourteen lit title screens, logos, EULAs and menus are `UI-ON-BLACK` (0.38% to 8.62%), and the one genuinely unreadable frame, `arcrunner-default-route-logos.png` (max=15, bright=0), is the sole `SPARSE`. I read all eighteen `UI-ON-BLACK` rows against their filenames and every one is a HUD, logo, EULA, splash, credits roll, language picker or title. `SPARSE` now means what the docs say. The dead band is gone.

My constructed counter-case from last round also improves: a real full-colour scene composited into a 5% viewport on black now reports `UI-ON-BLACK` rather than "a fade". That is an honest answer and I withdraw the objection.

**BLOCKING 2 is fixed, and the resolution is acceptable rather than an evasion.** `uniform_fill` at `frameclass.py:97` now requires the few colours to cover 90%, and the fps-overlay frame no longer short-circuits — I rebuilt one (opaque white glyphs on a 4K black frame, 6,666 white pixels) and it reports `UI-ON-BLACK`, as documented. Declining to detect the overlay from pixels is the right call: it is not detectable, and pinning the wrong-looking answer with a selftest assertion (`frameclass.py:134`) so the caveat cannot drift away from the behaviour is exactly the pattern this repo asks for. Naming `run.assertions.fps_overlay` as the after-the-fact check completes it.

**The selftest discriminates.** 9/9 at head, on the host and inside the `ps5ys` distrobox (both Pillow 12.3.0). I built the three mutants myself rather than trusting the table:

| mutant | mine | yours |
| --- | --- | --- |
| `uniform_fill = len(cols) <= 8` (bare colour count) | **5/9** — menu text, lit title, fps overlay, near-black fade all → `FLAT` | 5/9, same four |
| `elif bright > 0 and share < 0.02` (UI-ON-BLACK back to 2%) | **8/9** — lit title screen → `SPARSE` | 8/9, same case |
| `elif bright == 0: SPARSE` moved ahead of the coverage arm | **8/9** — dim full-frame picture → `SPARSE` | 7/9, "both full-frame pictures" |

Two reproduce exactly. The third does not, and the discrepancy is in your favour to correct: the "bright full-frame scene" case cannot flip to `SPARSE` under any natural revert, because it has `bright=10508` (I measured it). Only the dim picture flips, so the honest figure is 8/9 with one failure. That number lives only in a PR comment, not in the committed record, so it is a nit — but this repo is strict about restated numbers.

**Non-blocking items, checked by running them rather than reading the diff:** N1 verified against a directory holding a truncated PNG, an 8-byte PNG header, a non-image with a `.png` name and a good frame — three `UNREADABLE` rows, the good frames classified, the tally printed, exit 0; usage and empty-directory still exit 1, a missing path is an `UNREADABLE` row at exit 0, matching the scoped claim. N3 `4096+` prints. N4 verified empirically: a smooth synthetic gradient reports `CONTENT`. N7 both figures corrected. N8 verified with a real BMP, not just a renamed PNG. N5, N6, N9, N10 all present. The `capture_manifest.cpp` citations are each one line low after the master merge (`normalize_capture_rgba` is 91-99, `measure_pixel_content_rgba` 101-115, the alpha multiply 106-109) — substance correct.

---

## BLOCKING 3 — the "clean gap" is an artifact, and the two frames that define it are misclassified

`frameclass.py:23` and `frameclass/AGENTS.md:37` both state, as the sole justification for the 10% boundary:

> title screens, logos and EULAs cover 0.15-8.62% of the frame, while the least-covered real scene covers **10.13%**

I sorted the corpus by coverage and opened the two frames at 10.13%. Neither is a scene.

- **`stray-brightness-calibration.png`** — a settings screen: two lines of instruction text, three grey cat silhouettes and a slider, on black. There is no world in it at all. It is `UI-ON-BLACK` by any reading of your own class table.
- **`crisis-core-title.png`** — the Crisis Core title screen, and this one is worse, because the repo already documents it. `docs/CRISIS_CORE_STATUS.md:165-167`: *"The title screen's key art is missing (#2057). The 'Press Any Button' prompt, version string and copyright render; the CRISIS CORE logo/character art that occupies the left of the screen on hardware is absent (flat black)."* `CLAUDE.md:128` says the same. That is verbatim the class this tool exists to find — the guest is drawing and the world is what is absent — and the tool reports **CONTENT: a real picture**.

So the derivation is circular in the way `CLAUDE.md` warns about specifically: you took the minimum coverage over the frames the previous revision called `CONTENT` and treated it as the floor of real scenes, which assumes exactly what was under test. The gap is not clean; it is 8.62% to 10.13% with two UI screens sitting on the wrong side of it, and the threshold is 0.13pp below them. The next `CONTENT` frame up, `messenger-title.png` at 12.41%, genuinely is a picture (moon, mountains, starfield — I looked), so the corpus really does not separate the two classes in the 8-13% band.

I am not asking you to move the threshold — moving it to 11% narrows the gap to 1.4pp against `messenger-title` and re-tunes against the same corpus. I am asking for the fix you already applied to the fps overlay: **record the limit instead of claiming it away.** Correct the sentence in both places, name the two frames, and say that 8-13% coverage does not separate a dim UI screen from a dim scene and needs a human. Optionally pin `crisis-core-title` the way `frameclass.py:134` pins the overlay. As written, the doc reads as a validated boundary and it is not one, and per `CLAUDE.md` a false figure carrying a doc's authority is the expensive kind.

## BLOCKING 4 — after the reorder, two of the four class descriptions no longer describe the code

The ordering change is the right instinct (a dim picture is still a picture — `bendy-title.png` reproduces exactly: 93.33%, 2,899 colours, max=127, bright=0, `CONTENT`). But it silently changed what two classes mean, and neither the docstring nor `AGENTS.md` followed.

**`SPARSE` gained an undocumented coverage precondition.** `frameclass.py:33-34` says *"some non-black content but nothing legible (bright == 0) — a fade, or a frame so dark nothing in it can be read"*, and `AGENTS.md:31` repeats it. The parenthetical presents `bright == 0` as the whole condition. The code at `:99-101` requires `bright == 0` **and** `share < 0.10`. So a fade at 20% coverage, max=15, two colours, nothing legible anywhere, reports `CONTENT` — measured. A real scene faded to 12% brightness (max=30, bright=0, 27% covered) also reports `CONTENT`, which I think is the answer you actually want, and which makes "a fade" the wrong example for `SPARSE` regardless. Either way the description has to state the coverage bound.

**`FLAT`'s own listed example classifies `CONTENT`.** `frameclass.py:29` and `AGENTS.md:29` both give *"two-tone letterbox"* as a `FLAT` case. A letterbox has **black** bars, so the lit region is 75% (4:3 pillarbox) to 83% (2.39:1), which is below the new `share >= 0.90` gate. Measured, both on a 1920x1080 frame:

```
2.39:1 letterboxed white clear   max=255  nonblack=75.19%  bright=97440  colours=2  CONTENT
4:3   pillarboxed white clear    max=255  nonblack=75.00%  bright=97200  colours=2  CONTENT
```

A two-colour frame containing nothing but a uniform clear inside bars is reported as a real picture. That is failure mode 2 from your own docstring (`frameclass.py:17-18`) walking back in through the letterbox, and it is a plausible bring-up shape: a title that clears its viewport to a non-black colour and draws nothing into it.

The selftest case that looks like it pins this does not. `frameclass.py:125` builds "two-tone split" from a `(20,20,20)` base with a `(200,200,200)` half — the dark half is above the noise floor, so `share` is 100% and `uniform_fill` fires. It never exercises the black-bar geometry that a letterbox actually has. A reader who greps the selftest for the letterbox claim will believe it is covered.

Fix is a choice, not a prescription: either drop "two-tone letterbox" from both `FLAT` rows, or add a real black-bar case to the selftest and decide which class you want it in. What is not acceptable is a class table that a two-minute experiment contradicts, on a tool whose product is the label.

---

## Non-blocking

**N11 — on Windows every frame is enumerated twice.** `frameclass.py:169-170` globs four patterns; `fnmatch.filter` normcases both the pattern and the name when `os.path` is `ntpath`, so `*.png` and `*.PNG` match the same file. Demonstrated with Windows semantics forced:

```
patterns *.png *.bmp *.PNG *.BMP over [gta5-title.png, pathless-title.PNG, grab.bmp]
  -> 6 paths, 3 unique
```

Every row prints twice and the tally doubles. macOS is unaffected (POSIX `normcase` is a no-op, so the stride of the real directory entries is case-sensitive). prosper ships and runs tools on Windows, so this matters. One-line fix: dedupe, or list the directory once and filter on a lowercased extension.

**N12 — the new `tools/AGENTS.md` bullet reintroduces the exact N2 overclaim.** `prosper/tools/AGENTS.md:303` says the tool classifies *"at full resolution"*. You corrected that wording in the docstring (`:46`) and in `frameclass/AGENTS.md:65` to "1/16 nearest-neighbour stride at native resolution" — and then wrote the superseded phrasing into the index file, which is the one an agent reads first. Same bullet says "a directory of captured PNGs" though `.bmp` now works.

**N13 — the corpus figures are already stale at head.** `frameclass.py:20` and `AGENTS.md:34` say 128 frames; `AGENTS.md:44` says CONTENT=109. At head it is 129 and 110. This will drift on every screenshot merge, so consider phrasing that does not need a per-merge refresh, or accept the drift explicitly.

**N14 — the PR body still describes the reverted behaviour.** Its class table says `UI-ON-BLACK` is *"bright pixels covering under 2%"* and `FLAT` is *"any uniform fill (<= 8 distinct colours)"* with no coverage gate — those are precisely the two rules this round replaced, and the validation table is the pre-fix run. `CLAUDE.md` asks for a self-contained PR description, and the squash-merge message inherits it, so please refresh it before merging.

**N15 — the `capture_manifest.cpp` line citations are one low** after the master merge, as listed above. Trivial, but they are the kind of citation this repo treats as an assertion that you read the line.

---

## Path to approval

1. Correct `frameclass.py:20-26` and `frameclass/AGENTS.md:34-37`: name `stray-brightness-calibration.png` and `crisis-core-title.png` as UI screens sitting above the boundary, cite `docs/CRISIS_CORE_STATUS.md:165-167` / #2057 for the second, and state that 8-13% coverage is not separated by the corpus. Optionally pin one with a selftest case.
2. Make the `SPARSE` and `FLAT` rows match the code in both the docstring and `AGENTS.md`, and either drop the letterbox example or add a black-bar selftest case.
3. `prosper/tools/AGENTS.md:303` — the stride wording, not "full resolution".
4. Dedupe the glob.
5. Refresh the PR body table.

None of that needs new measurements from me and I do not expect it to move the corpus split. Re-run `--selftest` and the 129-frame corpus after, quote both, and I will re-review promptly — I want this tool in the tree.

mattias800 (commented):
Both findings were right, and the first one is the most useful thing anyone has told me about this tool — so I want to be precise about what it cost, because the answer is "the tool's main claim".

## BLOCKING 1 — the gap was an artifact, and following it further destroyed the threshold entirely

You were right about the circularity: I took the minimum coverage over frames the *previous revision of this same tool* had called CONTENT, and treated that as the floor of "real scenes". That is the assumption under test.

I opened the two frames that defined it. **Both are UI**, exactly as you said — `crisis-core-title.png` is "Press Any Button", a version string, a copyright line and a blue glow on black with the key art missing (#2057); `stray-brightness-calibration.png` is a settings screen with three cat icons and a slider.

Then I looked at the rest of the band, and it is worse than your finding:

| frame | coverage | colours | what it is |
| --- | --- | --- | --- |
| `crisis-core-title.png` | 10.13% | 4096+ | text and a glow on black |
| `stray-brightness-calibration.png` | 10.13% | 228 | a settings menu |
| `messenger-title.png` | 12.41% | **37** | real pixel art |
| `oregon-trail-gameloft-splash.png` | 16.03% | 2159 | a flat logo on black |
| `blue-prince-title.png` | 21.83% | 4096+ | real rendered 3D art |

**UI and scenes interleave on both axes.** Coverage puts a logo (16.03%) above a scene (12.41%); colour count puts pixel art (37) below a logo (2,159). There is no threshold on either statistic that separates them, and there was never going to be one — moving the boundary from 10% to 11% would have been the same mistake with a smaller margin.

So I deleted the claim rather than retuning it. `CONTENT` is now **`LIT`** — "something legible is drawn" — and the docstring and `AGENTS.md` both carry that table and state outright that deciding "is the world rendering?" above 2% needs eyes on the frame.

**`UI-ON-BLACK` is back at 2%, which is your number.** Your positive control is what makes that bound defensible, and it now selects those frames and nothing else: cyberspace-hud 0.81%, autosave-notice 1.05%, loading-mascot 1.49%, game-freak-logo 0.38%. Four frames, three of them the exact titles recorded as #2790 and #2952. The class does one thing and does it on evidence you chose.

The 14 title screens are no longer misclassified, but not because I widened a class to hold them — they are simply `LIT`, which claims nothing about them.

## BLOCKING 2 — both ordering defects confirmed and pinned

**The letterbox.** Confirmed: black bars put a 2.39:1 frame at ~75%, under the 0.90 bound, so `FLAT`'s own documented example classified as a picture. You were also right that my selftest case could not catch it — the `(20,20,20)` dark half never exercised black-bar geometry. Bound is 0.50 now, with a real letterbox case built from black bars.

**`SPARSE`'s undocumented coverage precondition.** Confirmed, and fixing it properly meant changing what `SPARSE` keys on rather than just documenting it. It now keys on the **peak**, not on a count of pixels above 128, and coverage does not rescue it — a fifth of a frame at peak 15 is `SPARSE`. The opposite error is mutation-checked: keying on `bright == 0` reddens the dim-picture case, because `bendy-title` peaks at 127 across 93% of the frame and is perfectly visible.

## The rest

- **Windows double-enumeration** — real, and thank you for forcing the semantics rather than reasoning about it. Deduped by real path; verified under forced case-insensitive globbing (1 row, not 4).
- **`tools/AGENTS.md:303` reintroducing "at full resolution"** — that was me writing the index bullet from memory instead of from the corrected file. Gone; `grep -c 'full resolution'` is 0 in both files.
- **Stale figures** — now 129 frames, LIT=124 / UI-ON-BLACK=4 / SPARSE=1, and the PR body is rewritten rather than patched, so the squash message will not inherit the reverted rules.
- **Your third mutant giving 8/11 rather than the 7/11 I reported** — you are right, and the reason is that the case cannot flip: `bright full-frame scene` has `bright=10508`. I had copied the number from an earlier run without re-deriving it. The four mutants at this head are 10/11, 6/11, 10/11, 10/11, each reddening exactly the case that names it.

Head is `5d77aa6c`.

mattias800 (commented):
**APPROVED** at head `5d77aa6c`. Fourth read, and this one is a clear pass: every number in the docs reproduces on my machine except one digit, every mutant reddens exactly the case it names, and the redesign did not introduce a misclassification I could find on the corpus or by hand.

The thing that earns the approval is not the code, it is what you did with BLOCKING 3. You could have moved the boundary to 11% and kept the class. Instead you opened the band, found that UI and scenes interleave on both axes, and deleted the claim. `LIT` now says exactly what the predicate supports and nothing more, and both docs say outright that the call above 2% needs eyes on the frame. That is the harder answer and it is the right one.

Everything below was run, not reasoned about. Environment: `ps5ys` distrobox, Python 3.14.5, Pillow 12.3.0.

## Reproduced

**Selftest: 11/11**, exit 0.

**Corpus split reproduces to the frame.** I extracted `assets/screenshots/` from the PR head into a scratch directory rather than trusting a working tree, and re-checked that `origin/master` is still at 129 PNGs at the moment of review:

```
LIT=124  SPARSE=1  UI-ON-BLACK=4  total=129        129 frames, 10.9 s
```

**All five non-LIT rows opened.** The four `UI-ON-BLACK` are `beast-of-reincarnation-game-freak-logo` (0.38%, the publisher wordmark on black), `sonic-frontiers-cyberspace-hud` (0.81%, ring counter, timer and speedometer over pure black — #2790 exactly), `sonic-frontiers-autosave-notice` (1.05%) and `metaphor-loading-mascot` (1.49%). The one `SPARSE` is `arcrunner-default-route-logos`, and I checked the thing most likely to be wrong about it: whether the stride was hiding a bright logo. It is not — at **step=1 over all 8,294,400 pixels the peak is still 15**, with 13,092 pixels at exactly 15 and nothing above. The frame really is near-black and `SPARSE` is the honest answer, which also matches the recorded rung 0.

**All four mutants rebuilt from the head file by `sed`, not taken on trust.** Every one matches your table and reddens exactly the named case:

| mutant | selftest | case that fails |
| --- | --- | --- |
| `share >= 0.90` | **10/11** | letterbox bars → `LIT` |
| `uniform_fill = len(cols) <= 8` | **6/11** | hud, fps overlay, lit title, both fades → `FLAT` |
| `elif bright == 0` | **10/11** | dim full-frame picture → `SPARSE` |
| `share < 0.10` | **10/11** | lit title screen → `UI-ON-BLACK` |

**The Windows dedup works, checked with a positive control rather than by reading it.** Patching `os.path.normcase` alone does not move the lever on Linux — `fnmatch.filter` short-circuits on `os.path is posixpath` and never normcases the names, so a naive forcing quietly tests nothing. Giving `fnmatch` an `os` shim whose `.path` is `ntpath` does move it, and I proved it moved before believing the result:

```
POSITIVE CONTROL fnmatch.filter([alpha.png, beta.PNG, gamma.bmp, delta.BMP], '*.png') -> ['alpha.png', 'beta.PNG']
POSITIVE CONTROL raw 4-pattern glob -> 8 paths, 4 unique
tool under forced Windows semantics -> 4 rows, total=4
```

**Threshold circularity — checked, and the answer is that there is nothing to falsify.** Neither `mx < 48` nor `share >= 0.50` is presented anywhere as derived or validated; both are stated numerically as the class definition, in the same sentence as the meaning. That is the form I asked for after round 3, and it is why I am not treating either as a finding. What I did measure: the corpus has **nothing at all between peak 15 and peak 127**, so 48 sits in a gap no frame tests; and the lowest colour count in the corpus at ≥50% coverage is **25** (`messenger.png`), so the `<= 8` bound has 3x margin against this project's most palette-poor real frame.

**The rest, each by running it:** the aliasing case is exact (1px lines at `x % 4 == 1` over a quarter of the frame → `max=0 colours=1 FLAT`); a smooth gradient → `LIT`; `4096+` prints; a directory holding a truncated PNG, an 8-byte stub and a non-image gives three `UNREADABLE` rows, keeps the tally and exits 0, while usage and an empty directory exit 1; a real BMP directory classifies; `grep -c 'full resolution'` is 0 in all three files; `normalize_capture_rgba`, `distinct_rgb_colors`, `nonblack_rgb_pixels` and `run.assertions.fps_overlay` all exist where cited; `git diff --check` clean; no `Claude-Session` trailers.

**And it still does the job it was built for.** On the original Stray run — the one the old downsampling classifier called 1/30 content — this head reports **UI-ON-BLACK on 9 of 30 frames**, including the menu frame at 0.18%. Consecutive fade frames split across the new peak bound in a way that reads correctly: frame 04 at peak 27 is `SPARSE`, frame 05 at peak 96 is `UI-ON-BLACK`.

## The five-frame table, which is now the central argument

Every coverage figure is exact. Four of the five colour counts are exact. One is off by one, and the reason is worth naming because it is diagnosable rather than random:

```
crisis-core-title.png              10.13%  4096+   as documented
stray-brightness-calibration.png   10.13%    228   as documented
messenger-title.png                12.41%     36   documented as 37
oregon-trail-gameloft-splash.png   16.03%   2159   as documented
blue-prince-title.png              21.83%  4096+   as documented
```

`37` is the colour count of **`messenger-title-fps-overlay.png`** (12.57% cover, 37 colours), the adjacent corpus frame. The row took its coverage from one file and its colour count from the other. I counted 36 twice, once through the tool and once with an independent loop; the file has not changed since #1341.

The argument itself is unaffected and I checked it by opening the frames: `oregon-trail-gameloft-splash` is a flat Gameloft wordmark on black at 16.03%, sitting **above** real pixel art at 12.41%, and `blue-prince-title` is a genuinely rendered 3D still life. UI above scene on coverage, pixel art below logo on colour count. The interleaving is real.

## Follow-ups — none blocking, listed in the order I would fix them

**F1 — `messenger-title` is 36 colours, not 37**, in `frameclass.py` docstring, `frameclass/AGENTS.md` and the PR body. Cause above.

**F2 — `UI-ON-BLACK`'s gloss is one clause stronger than its predicate, and the corpus supplies the counterexample from inside the class.** The docstring says *"THE diagnostic class: it means the guest is drawing and the WORLD is what is missing."* `beast-of-reincarnation-game-freak-logo.png` is 1 of the class's own 4 members and it is an ordinary publisher splash — nothing is missing, there is no world at that point in the boot. The same cliff cuts a homogeneous population: `pathless-title` (2.45%, a wordmark on black) is `LIT` while the Game Freak wordmark at 0.38% is `UI-ON-BLACK`, and both are the same kind of frame. The class table already lists "logo" as a member, so a reader is not led anywhere false about the frame itself — but the sentence claims a title-context judgement the pixels cannot make, which is the one place the docs still do what this round removed everywhere else. Suggest: "only UI is drawn; whether a world was *expected* there is a judgement the pixels cannot make."

**F3 — lowering `uniform_fill` from 0.90 to 0.50 reintroduces my round-1 case in the newly covered 50-90% band, and the code comment records the direction but not the cost.** Measured: an 8-colour full-frame 2D title at 87.8% coverage, `max=255`, 113,760 bright pixels → `FLAT`. Sharper, because it starts from a real frame rather than a construction — `messenger.png` (real 2D gameplay) quantized to 16 colours is `LIT`, quantized to **8** is `FLAT`, at unchanged 86.19% coverage. I am not asking you to move it back: a letterboxed uniform clear is far likelier than an 8-colour real scene, the corpus floor is 25 colours, and no rendered PS5 frame yields 8 distinct triples over half a million samples. But the `FLAT` row deserves the same treatment the `SPARSE` row got — one clause saying that a genuinely ≤8-colour picture covering half the frame lands here.

**F4 — say that 48 is chosen rather than pinned.** Nothing in the docs is false, but the corpus gap 15 → 127 means no frame tests it, and a reader may take a stated number for a measured one. One sentence.

**F5 — the `bright` column is printed, undocumented, and no longer decides anything.** `bright_at=128` survives only as a reported statistic since `SPARSE` moved to the peak. `bendy-title.png` prints `bright=0` next to `LIT`, which reads as a contradiction unless you already know the AGENTS.md paragraph that explains that exact frame. Define the column where the classes are defined.

**F6 — `r['name'][-28:]` truncates silently, and it has already cost accuracy in this PR.** The PR body names `reincarnation-game-freak-logo`, which is neither the file on disk (`beast-of-reincarnation-game-freak-logo.png`) nor the printed row (`arnation-game-freak-logo.png`). A report whose rows are meant to be actionable should not mangle the identifier; truncate in the middle, or widen.

**F7 — `--selftest` is not registered in ctest, and this repo has the pattern.** `prosper/CMakeLists.txt` registers `gen_progress_tracker.py --selftest` and `check_diag_gates.py --selftest` as cases, so "re-run it if you touch a threshold" could be enforced instead of requested. The likely reason not to is the Pillow dependency, which the other two do not have — if that is the call, it is worth one line in `AGENTS.md` saying so, otherwise the next person will wonder why it was skipped.

**F8 — the 129-frame figure drifts on every screenshot merge.** Accepted rather than raised again; noting it so it is a known cost and not a surprise.

## Closing

Three rounds of findings, three rounds where the fix was to make a claim smaller rather than to widen a class to cover its counterexamples. The `--selftest` cases each pin a specific mistake this tool made and then stopped making, the fps-overlay false positive is asserted rather than hidden, and the retracted "clean gap" is documented in the tool itself so nobody re-derives it. Merge it — and per `CLAUDE.md`, remember the `BLOG.md` entry.

mattias800:
Thank you — and the two number findings are both real, so I have fixed them and the mechanism behind them. Head is `26ee1fbd`.

**F1 — 36, not 37.** Confirmed: `messenger-title.png` is 36, and 37 belongs to `messenger-title-fps-overlay.png`. The row took coverage from one file and the colour count from its neighbour. Corrected in the docstring, `AGENTS.md` and the PR body. You are right that the argument survives — 36 against a flat logo's 2,159 makes the same point, arguably a sharper one.

**F6 — and this is the one with a cause worth fixing.** The tool right-truncated names to 28 characters, which turns `beast-of-reincarnation-game-freak-logo.png` into `reincarnation-game-freak-logo.png` — **not an obviously cut name, but a different plausible one**. I copied it out of my own tool's output into the PR body, where it named a file that does not exist. That is twice now that this tool's output has misled someone into writing something false, so I widened the column to 44 with a leading ellipsis, so a cut name looks cut.

**F2 — you are right and it is my own class's counterexample.** The Game Freak wordmark is 1 of the class's 4 members and nothing is missing behind it. The gloss now says what the class *detects*, and that the defect judgement depends on where in the boot the frame came from — a splash is an ordinary member, a gameplay frame is not.

**F5 and F4 — both documented rather than changed.** `bright` is reported and decides nothing; keying `SPARSE` on it is precisely the defect you found last round, so it stays as something worth seeing beside the peak, with a note saying it is not consulted. And `dim_at=48` and the `0.50` bound are now labelled as **chosen, not derived** — the corpus has nothing between peak 15 and peak 127, so it cannot pin the first.

**F3 — recorded, because you are right that the cost was unstated.** Lowering `uniform_fill` from 0.90 buys the letterbox at the price of a real frame quantized to ≤8 colours in the 50–90% band reading `FLAT`. That trade is now written down next to the constant instead of living in a review thread.

**F8** — corpus split is dated, with a note that the set worth remembering is the small non-`LIT` one rather than the total.

**F7** — not registered in ctest, and you guessed the reason correctly: Pillow is not a build dependency and adding one for a reporting tool seemed the wrong trade. Now said explicitly rather than left as an omission.

**BLOG.md** — added, and thank you for the reminder. It leads with the finding rather than the tool: a classifier we were about to rely on reported Stray's working main menu as flat black.

Your `step=1` check on the `SPARSE` frame is the one I would have skipped — confirming the peak really is 15 over all 8.29M pixels, rather than trusting the stride not to have missed a bright logo, is exactly the check the tool's own aliasing caveat calls for. Likewise forcing `ntpath` into `fnmatch` to prove the dedup lever actually moved, when a naive patch does nothing on Linux.

mattias800:
Note on the head, since your approval was on `5d77aa6c` and the branch has moved to `fbf8cbdf`.

The delta is your own non-blocking findings applied, plus one thing I owed the tree:

| commit | what |
| --- | --- |
| `26ee1fbd` | F1 (36 not 37), F6 (name column widened to 44 with a leading ellipsis), F2 (class gloss), F4/F5 (`bright` and the two constants documented), F3 (the 0.50 trade recorded), F8 (corpus split dated) — plus the `BLOG.md` entry you reminded me about |
| `fbf8cbdf` | F7 — I answered "why isn't `--selftest` in ctest" in a PR comment and not in the tree, which is the exact failure `CLAUDE.md` names as *"a result reported only in a message is not in the project record"*. It is in `AGENTS.md` now |

**No change to the classifier's behaviour**: the four thresholds, the ordering, and all 11 selftest cases are untouched, and the corpus split is still `LIT=124 UI-ON-BLACK=4 SPARSE=1` over 129 frames. The only executable change since your approval is the width of the name column in the report.

I am treating the approval as still applying on that basis. Say so if you disagree and I will hold.

Issue #2990, opened 2026-08-26
[Game tracker] Tomb Raider I-III Remastered (PPSA16901)


mattias800:
## Current milestone

**Rung 3 — Croft Manor runs with real GPU draws and the scene renders with correct geometry. Every surface is still untextured (#2998), which is what keeps this below a rung-4 claim.**

Rung 2 needed no prosper code change at all — the title booted and rendered real frames on its first attempt. Rung 3 needed exactly one fix: an unannounced 32-bit index buffer whose indices exceed 0xFFFF (#304 part two).

### Best checked-in visual

[![Tomb Raider I-III Remastered — Croft Manor gameplay, correct geometry](https://raw.githubusercontent.com/mattias800/prosper/master/assets/screenshots/tomb-raider-gameplay.png)](https://github.com/mattias800/prosper/blob/master/assets/screenshots/tomb-raider-gameplay.png)

## What works

- Boots and renders 1920x1080 composited frames from a default launch, guest alive throughout.
- The 40-page EULA gate renders and scrolls.
- The publisher/developer logo sequence renders correctly (Saber Interactive logo is pixel-clean).
- The Tomb Raider I title screen renders: logo, Lara's portrait, the animated ring menu, and the `Select` / item labels.
- Up/Down on the title screen switches between Tomb Raider I, II and III; the Tomb Raider III title screen also renders.
- Left/Right rotates the ring menu; its five items read correctly (`Lara's Home`, `Controls`, `Extras`, `Game`, `Sound`).
- Cross on `Lara's Home` loads and runs the Croft Manor level — the guest reaches a live 3D scene with animating character models (Lara and Winston are both recognisable).

## Current blockers

- **Every surface renders untextured** and the scene is over-bright ([#2998](https://github.com/mattias800/prosper/issues/2998)). The world *geometry* is now correct — the shattering was an unannounced 32-bit index buffer read as 16-bit, fixed as #304 part two.
- **Text is intermittently garbled.** The EULA body and the Tomb Raider III title screen's labels draw the wrong glyphs, while the same font renders correctly for the `EULA` heading, the `1/40` page counter, and every Tomb Raider I ring label. Right font, right layout, wrong glyph selection.
- **Solid untextured quads on the menus** — small magenta quads on the Tomb Raider I ring and a large solid violet block covering the Tomb Raider III ring.
[...]

PR #3044, opened 2026-08-26
fix(win): guest semaphore timedwait off the winpthreads tick, POSIX untouched

mattias800:
Refs #3013. Refs #3033. Fourth slice of the #2984 recovery.

## The defect

On Windows `sem_timedwait` is winpthreads', whose timed waits resolve on its own master tick regardless
of `timeBeginPeriod` — **15.57 ms** measured for a 5.33 ms request (#3013's probe, with its
posted-semaphore control). FMOD's mixer paces 256-frame blocks on exactly this call, so a guest asking
for one grain was released once per tick and delivered audio at roughly a third of real time
(*Blasphemous 2*, #2985). #3022 fixed the sleep primitives and deliberately left this one alone.

## Two departures from the recovered form (`c45debbf`)

**POSIX is left alone.** That commit polled on every platform. Where `sem_timedwait` already honours the
deadline, polling replaces one exact wait with a loop that is worse on both latency *and* CPU. The tick
is a Windows property, so the workaround is `#ifdef`-scoped and POSIX keeps the native call.

**The poll interval is adaptive, not a flat 2 ms.** A *posted* semaphore returns in **0.01 ms**
natively, so any poll loop trades wake latency for timeout accuracy. A flat 2 ms slice costs up to 2 ms
on every wake — 37% of the grain period this exists to serve. Fine slices while the wait is young keep
pacing tight; coarse slices afterwards stop a guest parked for seconds from spinning.

## The fine slice is 500 µs because that is the measured floor

Not a chosen number. A short high-resolution-timer sleep on this host, 200 samples per row:

| requested | median | min | p90 | max |
| --- | --- | --- | --- | --- |
| 0.05 ms | **0.519** | 0.103 | 0.533 | 0.885 |
| 0.10 ms | **0.519** | 0.459 | 0.531 | 1.049 |
| 0.20 ms | **0.518** | 0.472 | 0.536 | 0.994 |
| 0.50 ms | 1.017 | 0.508 | 1.525 | 2.572 |
| 2.00 ms | 2.066 | 2.013 | 2.535 | 2.901 |

Anything under ~0.5 ms costs 0.52 ms. An earlier revision asked for 200 µs, which reads as a tighter
loop than the machine can run and would mislead whoever tunes it next. So the fine window buys a poll
interval ~4× shorter than the coarse one, not the 10× the constants implied.

## The test, and why its history matters more than its result

Two arms that constrain each other — timeout accuracy and wake latency. Pinning only the first would
[...]

PR #3060, opened 2026-08-27
evidence: retract "Tomb Raider's world is textured" — the capture was a loading screen

mattias800:
Refs #2990, #2998.

## The project owner recognised the artwork

The screenshot published in `BLOG.md` as *"Tomb Raider's world is textured"* — Croft Manor's
brickwork, leaded windows and dolphin fountain — is the game blitting **its own pre-rendered loading
picture**, `2/PIX/HD/MANSION.DDS`.

Measured rather than judged. Decoding every same-resolution image asset in the dump and diffing
against the checked-in capture:

| asset | mean abs diff | pixels within 8/255 |
| --- | --- | --- |
| `2/PIX/HD/MANSION.DDS` | **0.02 / 255** | **100.00 %** |
| `3/PIX/HD/LONDON.DDS` (next closest of 41) | 52.77 / 255 | 1.27 % |

It is the same image. **Blitting a stored full-screen picture requires no world rendering at all**,
so the frame never evidenced what it was captioned as — the world was untextured before that capture
and after it.

## Why it got through

It looks *better* than the emulator could plausibly render — soft shadows, ambient occlusion, depth
of field, per-leaf foliage. That reads as success rather than as the warning it is. It was also
paired with a **genuine** untextured "before" shot, so the A/B looked rigorous while actually
comparing a 3D render against a 2D blit. It passed a PR, a blog entry and a merge. No automated check
could have caught it, because none existed.

Recorded as **instrument trap 230**, plus a `## Ruled out` row in the title's status doc.

## So this adds the check that was missing

`tools/evidence/prerender_check.py` decodes every same-resolution asset in a dump and diffs it
against a candidate screenshot:

- **exit 2** — the frame IS the game's own artwork; not progression evidence
- **exit 0** — nothing matches
- **exit 1** — the check could not run

That third state is deliberate and load-bearing: *could not check* must never be reportable as
*checked and clean*, so a candidate whose resolution matches no asset exits 1 with
`NOTHING WAS COMPARED` rather than passing. The tool also states its own scope — it finds full-screen
blits of assets stored as ordinary images, and a pass means "no stored picture explains this frame",
**not** "this frame is rendered 3D".

**Verified on all three arms:**

```
known-false capture  -> exit 2   MATCH 2/PIX/HD/MANSION.DDS (0.02/255, 100.00%)
genuine render       -> exit 0   closest asset 53.55/255
unmatched resolution -> exit 1   NOTHING WAS COMPARED
```

I ran it against **my own** interior capture before trusting that one — 53.55/255, genuine — because
the discipline is worthless if I exempt my own evidence.

## Also corrected

- **`COMPATIBILITY.md`** said *"every surface is still untextured"*. Tonight's capture shows textures
  **do** reach world surfaces; the wrong ones are selected, drawn from real correctly-decoded assets
  elsewhere in the game (Lara's passport, the Game Boy collectibles). The defect is **selection, not
  decode**, and the row now says that. The rung number is left alone — that judgement is not mine to
  make unilaterally, and per the ladder it is the milestone *text* that carries the state.
- **`BLOG.md`** keeps the retracted entry with its picture withdrawn rather than deleting it: the
  blog's own rule is that it records what was claimed and when.
- The false-evidence image is deleted from `assets/screenshots/`.

## Verification

`check_numbered_table.py` over all 151 tracked `.md` files passes, including the `--ordered
--table-header Instrument --baseline origin/master` gate for the new trap row (`1..230, unique and
ascending, none of the 229 base rows deleted`). `diff --check` clean. No code touched — the new file
is a standalone script that is not built.

Not merged under the docs-only exception: the diff adds a `.py` and deletes a `.png`.

---

## A note on the sweep figures in the comments, and how they misled both of us

The repo-wide sweep used a **scratch, approximate** mapping from screenshot filename to title id
(loose substring matching against `COMPATIBILITY.md`). It is deliberately **not** part of this PR —
`prerender_check.py` takes the dump root as an argument precisely so it never has to guess.

The mechanism is worth stating, because one defective mapping produced two *opposite* wrong answers
and neither was visible in its own output:

- **A screenshot that maps to nothing is silently skipped**, so a statistic computed over "all
  screenshots" is really over an unstated subset. This made the reviewer's "worst genuine render
  3.47%" too low — the file that would have set the maximum was not in the map at all.
- **A screenshot that maps to the WRONG title is compared against the wrong dump**, and reported as a
  normal result. This made my "50.00% overlap at mean 103.88" too high, and I then used that artefact
  to justify a mean ceiling — which review later disproved on independent grounds.

Each of us believed the number that suited our position, and neither figure announced itself as
suspect. So: read the sweep as *"among the screenshots it mapped correctly, no further exact
pre-rendered blit was found, and two title screens are DOMINATED and correctly captioned"*. The
"could-not-compare" count is a floor, not an exact figure.

Anyone re-running it should supply the mapping deliberately and fail loudly on an unmapped file.


mattias800:
## Repo-wide sweep: is any other published screenshot a pre-rendered picture?

Ran the new tool over **every checked-in screenshot that maps to a dump present on this machine** —
109 of the 128 in `assets/screenshots/`, across the titles in `COMPATIBILITY.md`.

| result | count |
| --- | --- |
| genuine — compared against same-size assets, nothing matched | **62** |
| **PRE-RENDERED** | **0** |
| could not compare — no same-size asset in that dump | 47 |

**The zero is validated, not assumed.** Per the charter's rule that a clean zero needs a positive
instance constructed outside whatever produced the null, the known-false Tomb Raider capture was run
back through the *same* sweep path and returned **exit 2**, naming `2/PIX/HD/MANSION.DDS` at
0.02/255. So the sweep can see the class it reports as absent.

**Read the 47 honestly — they are not cleared.** The tool exits 1 rather than 0 for them precisely so
they cannot be counted as passes. Mostly 4K captures whose dumps store artwork at other resolutions,
plus dumps with no ordinary image assets. A pre-rendered frame in that set would not have been
caught, so this sweep says *"no other case found among the 62 checkable"*, not *"no other case
exists"*.

That is the useful conclusion: the Tomb Raider case looks isolated among what can currently be
checked, and the remaining exposure is a known, bounded 47 rather than an unknown.

mattias800 (commented):
# Independent review — #3060 @ `2db08936`

**REJECTED** — two blocking findings, both small edits. To be unambiguous up front: **the retraction
itself is correct and I verified it independently, not by re-running your tool.** The problems are a
sentence in the trap row that contradicts `COMPATIBILITY.md` in this same diff, and a gap in
`prerender_check.py` that lets the very frame it exists to catch pass once anything is drawn on top
of it.

## 1. The central claim — confirmed, and stronger than stated

I recovered the deleted image from `origin/master`, decoded `2/PIX/HD/MANSION.DDS` (DX10 header,
DXGI format `0x62` = `BC7_UNORM`, 1920x1080) and compared with my own numpy script, no code from this
PR involved:

| | |
| --- | --- |
| mean abs diff (per channel) | **0.0186 / 255** |
| **max abs diff, any channel, any pixel** | **1 / 255** |
| pixels within 8/255 (per-channel max) | **100.000 %** |
| same-size decodable assets found | **41** (matches your count) |
| next closest (`3/PIX/HD/LONDON.DDS`) | 50.86 by my metric, 52.77 by yours |

The maximum is the number that settles it: **no pixel in the entire frame differs from the decoded
asset by more than 1/255.** That is BC7 round-trip rounding, not rendering. The separation to the
next asset is ~2700x. This is not a judgement call and the retraction is right.

The small gap between my 50.86 and your 52.77 on LONDON is only the metric: yours takes the mean of
`ImageChops.difference(...).convert('L')`, i.e. luminance-weighted, mine is the plain per-channel
mean. Both are fine; worth knowing they are not the same statistic when you quote one.

Your tool reproduces its published numbers exactly on the recovered image: `0.02/255`, `100.00%`,
next `52.77`, exit 2.

## 2. Scope — you did not over-reach on the code, and your stated reason is weaker than the real one

`git diff --name-only origin/master...HEAD` is docs + the new `.py` + the deleted `.png`. `3df615a5`
(#3050) is untouched. Correct call, and independently justified: the post-fix interior capture in
`c3d59512` shows textures genuinely reaching surfaces, so the 2D_ARRAY change did something real
regardless of what the retracted frame showed.

On leaving the rung alone (§5) — you say "that judgement is not mine to make unilaterally". True, but
the stronger reason is available and you should use it: **the retracted image was never the rung's
evidence.** Rung 3 was set by `227f5fbe` (#3006, the 32-bit index-buffer fix) on the geometry, and
#3050 did not touch `COMPATIBILITY.md` at all. So the rung is untouched by the retraction as a matter
of record, not of etiquette. Say that instead — it is checkable and it forecloses the obvious "should
this have dropped to rung 2?" question.

## BLOCKING 1 — trap 230 contradicts `COMPATIBILITY.md` in the same diff

`prosper/docs/GAME_COMPAT_ORCHESTRATION.md:358`:
> The world was untextured before the capture and after it.

`COMPATIBILITY.md:81`, added by this same PR:
> Textures now reach world surfaces but the WRONG ones are selected

These cannot both be true, and the second one is the supported one.

Evidence that "after it" is wrong: `git log --follow` shows `tomb-raider-croft-manor-untextured.png`
and the retracted `-textured.png` were added by the **same commit**, `3df615a5`. The untextured shot
is therefore the *pre-fix* "before" — it is not evidence of any post-fix state. Nobody has published
a post-fix exterior frame at all. What has been published post-fix is the interior
(`c3d59512`), which shows real decoded textures on the walls and floor, and #3050's own measurement
of 142,635 distinct colours against master's 64,730 on the same route frame.

Failure scenario: an agent six weeks from now reads trap 230 — a durable, high-authority row — takes
"the world was untextured after it" at face value, and either re-opens the 2D_ARRAY frontier that is
already fixed or proposes reverting #3050 as having achieved nothing. The charter's own warning
applies directly: under-reporting what works is as much a defect as over-reporting it, and it fails
in the direction nobody files.

The sentence is also **unnecessary** — your argument is complete without it. "A blit needs no world
rendering, therefore the frame is not evidence of world rendering" stands on its own and does not
depend on what the world actually looked like.

Fix: strike it, or replace with something like *"What the world actually does post-fix is the
wrong-texture-selection defect in the 2026-08-27 interior capture — not what this frame appeared to
show."* Same sentence in the PR body.

## BLOCKING 2 — `prerender_check.py` passes a loading screen with anything drawn on it, and the pass line over-claims

`prosper/tools/evidence/prerender_check.py:100` gates on `best_mean < threshold` alone. `within8` is
computed at `:92`, printed at `:97`, and **never used in the verdict** — and it is the robust
statistic of the two.

I took the exact retracted MANSION frame and drew a black bar with `LOADING...` across the bottom —
i.e. a progress indicator, which is what loading screens normally have:

| overlay | mean | within8 | exit |
| --- | --- | --- | --- |
| none | 0.02 | 100.00 % | 2 (MATCH) |
| 2 % of height | 1.50 | **98.07 %** | **0 — "no STORED PICTURE explains this frame"** |
| 5 % of height | 3.87 | 95.04 % | **0** |
| 10 % of height | 8.16 | 90.06 % | **0** |

Two percent of the screen flips the tool's answer on the one frame it was built for.

This is not only synthetic. Run it over the screenshots already checked into this repo:

```
assets/screenshots/tomb-raider-title-screen.png
  mean=2.40/255  within8=97.10%  1/PIX/HD/TITLE_EU.DDS   -> exit 0
assets/screenshots/tomb-raider-title-screen-tr2.png
  mean=2.73/255  within8=95.22%  2/PIX/HD/TITLE_US.DDS   -> exit 0
```

Both print **"no asset matches ... no STORED PICTURE explains this frame"** while a stored picture
demonstrably explains ~96 % of each. That message is false as printed.

This is also review item 4 (is the scope note honest?) and the answer is: the *docstring* is honest —
it excludes "artwork composited with other elements" — but the **pass output is not**, and the output
is what an agent will read, quote and paste into a PR. A scope note in a docstring cannot fix a
verdict line that asserts more than the code checked. Given this folder's founding principle that
*"could not check" must never read as "checked and clean"*, the sibling case *"partially matched" must
never read as "nothing matched"* belongs with it.

Failure scenario: trap 230 recurs one step harder. A title blits a loading screen with a spinner or a
`PRESS X TO CONTINUE` prompt; `tools/AGENTS.md` tells the agent exit 0 means go ahead; the agent
publishes and now has a *tool-certified* false claim, which is worse than the uncertified one this PR
is retracting.

Fix, small, using what you already compute:

```python
if best_mean < threshold or best_within8 >= 90.0:   # tune the constant
    return 2
```

or add an explicit third band that reports SUSPICIOUS and exits 2. And soften the exit-0 line to
*"no stored picture matches within the threshold (closest %.2f/255, %.2f%% within 8/255)"* — state
the measurement, not the conclusion.

## Non-blocking

1. **Self-match.** If the candidate sits inside the searched root, it is compared against itself:
   `prerender_check.py shot.png <dir containing shot.png>` -> `MATCH: ... myshot.png`, mean 0.00,
   exit 2. Fails safe (blocks rather than passes) but is baffling when hit. One `os.path.samefile`
   guard in `iter_assets` / the loop.
2. **`--top` with a negative value** (`:97`): `int` accepts `-3`, so `opt()` does not reject it, and
   `scored[:-3]` prints 38 rows instead of 3. Verdict comes from `scored[0]` so it is cosmetic.
3. **Flags before positionals** (`:36`, `args = [a for a in argv[1:] if not a.startswith('--')]`): the
   flag's *value* is collected as a positional, so `--threshold 2 <shot> <dump>` sets
   `candidate='2'`. Every ordering I tried exits 1, so it fails safe — `argparse` would remove the
   class outright.
4. **`BLOG.md` wording:** "no automated check did, and none could have" reads as *no check could catch
   this*, which this PR disproves by shipping one that does. The PR body's "because none existed" is
   the accurate phrasing; use it in the blog too.
5. `COMPATIBILITY.md` keeps "the scene is over-bright", whose only supporting capture is the pre-fix
   untextured shot. Carried over rather than introduced here, so not a finding — but it is now the
   one clause in that row with no current evidence behind it.

## What I checked that is clean

- **Exit-code separation (item 3) holds in every arm I could construct.** exit 1 for: no args, one
  arg, missing Pillow path, unreadable/nonexistent candidate, `dump_root` not a directory, bad
  `--threshold` / `--top` values, missing value at end of argv, and an empty dump -> `NOTHING WAS
  COMPARED`. I could not find any input that reports exit 0 having compared nothing — the
  `if not scored` guard at `:105` is correctly placed *before* the verdict. The load-bearing state is
  genuinely load-bearing.
- **Walk is bounded and safe.** `os.walk` defaults to `followlinks=False`, so no symlink cycle; the
  extension filter is applied before `Image.open`; `scored` accumulates floats and strings only, not
  images. Decode failures are counted and reported rather than swallowed.
- **Differing modes.** Built a synthetic dump with `L`, `P`, `CMYK`, `I;16`, `RGBA` and a truncated
  `.dds`: all convert and score, the corrupt file is counted as `undecodable`, nothing raises.
- **`COMPATIBILITY.md` (item 5) is supported.** I opened
  `tomb-raider-croft-manor-interior-wrong-textures.png`: walls tiled with Game Boy console artwork
  and a UK passport data page, floor covered in a passport observation page reading "THIS PAGE IS
  RESERVED FOR OFFICIAL OBSERVATIONS / CETTE PAGE EST RESERVEE AUX OBSERVATIONS OFFICIELLES" with
  swallow motifs. The assets are crisply decoded and correctly filtered — "selection, not decode" is
  precisely what the picture shows, and it is the right correction to "every surface is still
  untextured".
- **Trap 230's row (item 6) passes every gate**, and apart from BLOCKING 1 the content is good: it
  names the tell (*it looks better than the emulator could plausibly render*), the reproduction, and
  the separation figure. Gates run at `2db08936`:
  - `check_numbered_table.py --ordered --table-header Instrument --baseline <origin/master copy>` ->
    `1..230, 230 rows, unique and ascending, no unused numbers; none of the 229 base rows deleted`, exit 0
  - all 151 tracked `.md` -> exit 0
  - `check_merge_result.py` -> clean merge, gate passes on the merge result, exit 0
  - `git diff --check` -> exit 0; `Claude-Session:` trailer count -> 0
- **`BLOG.md`** correctly re-captions the retained shot from "The same view before" to "the same
  level", which is the right precision given the retracted frame was never a view of anything.

## To clear

BLOCKING 1 (delete or rewrite one sentence in the trap row and the PR body) and BLOCKING 2 (gate on
`within8` as well as `mean`, and soften the exit-0 wording). Re-run
`prerender_check.py` over the two title-screen captures afterwards — if they still report "no stored
picture explains this frame", the fix has not landed. Everything else stands and the retraction
should go in.

mattias800:
## Sweep re-run after the review found the tool passing its own target case

The sweep I posted above was run with a **mean-only** verdict. Review showed that gate clears a
loading screen once anything is drawn on it — a 2%-of-height "LOADING..." bar on the retracted frame
flipped it to exit 0 at 98% of pixels within 8/255. So that sweep's zero was weaker than it looked,
and this supersedes it.

Re-run with the dominance bar (≥90% of pixels within 8/255), same 109 mapped screenshots:

| result | count |
| --- | --- |
| genuine — nothing matches | 60 |
| **flagged** | **2** |
| could not compare — no same-size asset | 47 |

Both flags are real and both are **Tomb Raider title screens**:

```
DOMINATED  tomb-raider-title-screen.png      97.10% of 1/PIX/HD/TITLE_EU.DDS
DOMINATED  tomb-raider-title-screen-tr2.png  95.22% of 2/PIX/HD/TITLE_US.DDS
```

**Neither is retracted.** A title screen largely *is* the stored title artwork with menu elements
composited over it, and both are captioned as title screens — which is what they show. That is why
the DOMINATED verdict asks what prosper actually rendered rather than calling a frame false: the
useful output here is "most of this frame came out of a file", which a reader should know when
weighing it as evidence, not a verdict that it is fake.

The 47 that could not be compared are still not cleared, for the same reason as before — the tool
exits 1 for them precisely so they cannot be counted as passes.

**What changed in the conclusion:** the earlier "0 hits" was partly an artefact of the weaker gate.
The honest statement is that among 62 checkable screenshots no *further* frame is an exact
pre-rendered blit, two are substantially stored artwork and correctly labelled, and 47 remain
unchecked.

mattias800 (commented):
# Independent review, round 2 — #3060 @ `419b674a`

**REJECTED** — both round-1 findings are genuinely fixed and verified. One new blocking finding, and
it is the exact thing you asked me to check: **yes, `--overlap` produces false DOMINATED verdicts on
real frames, and this project's characteristic failure frame is the one that trips it.** You were
right to ask. The fix direction is small and I verified it separates the cases before recommending
it, so round 3 should be quick.

## Round-1 findings: both fixed, re-verified at this head

**Finding 2 (the mean-only gate).** My counterexamples now all land:

| overlay on the retracted MANSION frame | before | now |
| --- | --- | --- |
| 2 % of height | exit 0 | **exit 2 DOMINATED 98.07 %** |
| 5 % | exit 0 | **exit 2 DOMINATED 95.04 %** |
| 10 % | exit 0 | **exit 2 DOMINATED 90.06 %** |

(You quoted 93.98 % for the 2 % bar and I measure 98.07 % — different synthesised bars, same
conclusion.) The EXACT / DOMINATED split is the right shape, and all three exit-1 arms still hold at
this head: no-comparison → 1, unreadable candidate → 1, bad `--overlap` value → 1.

**Finding 1 (trap 230 vs `COMPATIBILITY.md`).** Fixed, and the replacement is better than a deletion
would have been — it now explains the same-commit provenance, so the next reader learns why the
"before" shot proves nothing about the post-fix state instead of just not being misled. The
`COMPATIBILITY.md` rung note now carries the record-based reason. Both accurate.

## BLOCKING — the dominance bar fires on darkness, not on identity

You asked me to check for a false DOMINATED on a genuine render. It is there, it is reproducible
across the dump corpus, and the mechanism is that **`within8` counts a pixel as matching whenever
both images are dark there** — which says nothing about whether the frame is the asset.

**Proof on a real checked-in screenshot.** `assets/screenshots/stray-brightness-calibration.png`
scores **88.77 %** against `<DUMP_ROOT>/PPSA02101-app0/sce_sys/pic1.dds` — 1.23 points under your
bar, i.e. it already nearly fires. Decomposing it:

```
shot  : mean luminance 5.20,  90.0% of pixels < 16
asset : mean luminance 9.32,  96.3% of pixels < 16
pixels where BOTH are dark: 89.0%  -- that IS the whole 88.77%
on the 11.0% of pixels not both-dark, 0.0% are within 8/255
```

Zero percent. The frame and the asset agree on nothing that carries information; they merely share a
black background.

**Proof by hand-constructed positive instance**, per the charter's rule about not trusting a null
from the same source that produced it. I built flat synthetic frames — pure black, flat dark grey, a
flat clear — and swept them against every dump. **19 asset pairs score ≥ 90 %**, including:

| synthetic frame | title | asset | overlap |
| --- | --- | --- | --- |
| pure black | `PPSA03274` (*Unbound*) | `sce_sys/pic1.dds` | **100.00 %** |
| pure black | `PPSA26414` (*R-Type Delta*) | `data/font/screen_mask384.tga` | **100.00 %** |
| flat dark grey / flat clear | `PPSA02058` (*Balan*) | `sce_sys/pic1.dds` | **100.00 %** |
| pure black | `PPSA03831` (*Sonic Frontiers*) | 5 separate UI textures | 94.6–97.5 % |
| pure black | `PPSA16901` (*Tomb Raider*) | `1/PIX/LEGAL_1..4.DDS` | 93.5–95.9 % |

**Why this matters here rather than in the abstract:** a black or flat-clear frame is this project's
single most common bring-up state, and `COMPATIBILITY.md` says so on the rows in question — *Unbound*
"composites as a flat two-tone clear", *PGA TOUR 2K25* "every frame is black", *Beast of
Reincarnation* "every frame is a flat clear", *Sonic Frontiers* "the world is black behind the HUD".
*Unbound's own* `pic1.dds` scores 100.00 % against black. So the next agent who captures an Unbound
level frame and runs the check — which `tools/AGENTS.md` now instructs them to do before publishing —
is told:

> DOMINATED: 100.00% of this frame's pixels are within 8/255 of the stored asset `sce_sys/pic1.dds`
> The frame is mostly stored artwork. ... See instrument trap 230.

The frame is not mostly stored artwork. The frame is empty. That is a confident, specific, wrong
diagnosis pointing at the wrong trap, and it is the "check people learn to ignore" outcome you named.

**It is not trap 230 again** — it fails toward blocking, never toward a false pass, so it cannot
manufacture progression evidence. That is why I am flagging the bar rather than the whole approach.
But `tools/evidence/AGENTS.md` sets the standard itself: *"A tool in this folder that fails silently
is worse than no tool, since its whole purpose is to be trusted at the moment somebody is about to
publish a claim."* A verdict nobody believes fails that standard from the other side.

### Verified fix direction

The cleanest framing is that this is **your own founding principle pointed the other way**: a black
frame against a black asset is *nothing was compared* wearing a verdict. Restrict the overlap to
pixels that carry information — where either image is above a luminance floor — and the classes
separate completely. Measured, not proposed:

| case | current overlap | informative coverage | robust overlap |
| --- | --- | --- | --- |
| retracted MANSION frame | 100.00 % | 83.30 % | **100.00 %** |
| + 2 % LOADING bar | 98.07 % | 83.31 % | **97.78 %** |
| + 5 % bar | 95.04 % | 83.31 % | **94.24 %** |
| + 10 % bar | 90.06 % | 83.30 % | **88.40 %** |
| TR title screen | 97.10 % | 63.34 % | **95.68 %** |
| TR title screen (TR2) | 95.22 % | 94.34 % | **94.93 %** |
| stray calibration (false) | 88.77 % | 11.05 % | **0.03 %** |
| pure black v `PPSA03274` `pic1.dds` (false) | 100.00 % | 0.00 % | **undefined** |
| pure black v `PPSA26414` `screen_mask384.tga` (false) | 100.00 % | 0.00 % | **undefined** |

Every true positive keeps 88–100 %; every false positive collapses to 0.03 % or has nothing to
compare at all. The **coverage column alone is a sufficient guard** — true positives 63–94 %, false
positives 0–11 % — so the minimal change is: if informative coverage is below roughly 25 %, both
images are mostly empty, the comparison carries no information, and the tool should say that instead
of claiming DOMINATED. Whether that reports as exit 1 or as a distinct "uninformative" line under
exit 0 is your call; what matters is that it stops asserting artwork. Gating on the robust overlap as
well would be strictly better and costs one extra masked comparison.

## Your title-screen judgement call — I think you are right

Keep them, and the reasoning is stronger than "a title screen is mostly artwork":

**The difference between these and the retracted frame is in the claim, not in the pixels.** The
retracted frame was captioned *"Tomb Raider's world is textured"* — an assertion about 3D world
rendering, which a blit cannot support at any overlap. "The title screen renders" is *supported* by a
correct blit, because blitting composited artwork is how a title screen renders on essentially every
game. The frames show what they say they show.

The ladder already guards the over-read you would be worried about: rung 2 is "title screen reached
and rendered" and rung 3 requires the gameplay scene, precisely because a title screen is worth less
than world rendering. That separation is the ladder's job, not the tool's — which is also why
DOMINATED asking "what did prosper actually render here" is the right wording rather than declaring
the frame false. I would keep that wording even after the coverage fix.

One suggestion, non-blocking: **record the waiver next to the tool, naming those two files.** You
have now answered this question once; without a note, the next agent who runs the sweep re-derives it
from scratch and may reach the opposite answer. Three lines in `tools/evidence/AGENTS.md` — these two
captures are known DOMINATED, here is why they stand — closes it permanently. Worth adding that the
TR2 title screen is 94.93 % the stored asset on informative pixels, so what prosper contributed there
is the compositing; a reader deserves to know that from the caption.

## A round-1 number of mine that was wrong, corrected

In my round-1 working I measured that `within8` was far looser than a per-channel criterion —
43.6 % of pixels "within 8" for two *unrelated* pictures. **That was my bug, not the tool's:** int16
overflow in my own numpy luminance weighting (`255 * 587` wraps). PIL's actual metric gives 1.27 %
for that pair, and 1–6 % for unrelated images generally. The metric is tight, and the headroom is
real — the darkness effect above is a genuinely separate mechanism, not that one. It never reached my
posted review, but a reviewer's wrong number propagates hardest, so it is on the record.

## Clean at this head

- Gates: `check_numbered_table.py --ordered --table-header Instrument --baseline origin/master` →
  `1..230, unique and ascending, none of the 229 base rows deleted`, exit 0; all **152** tracked
  `.md` → exit 0; `check_merge_result.py` → clean merge, gate passes on the merge result, exit 0;
  `git diff --check` → 0; `Claude-Session:` trailers → 0.
- My independent repo-wide sweep (94 mapped screenshots, 55 with a comparable asset) reproduces
  **exactly** your two flags and no others — `tomb-raider-title-screen` 97.10 %,
  `tomb-raider-title-screen-tr2` 95.22 %. Your sweep is honest. Median overlap for genuine renders is
  5.81 %.
- No compiled code touched; the retraction, the trap row, the status-doc row and the
  `COMPATIBILITY.md` correction are all accurate as they now stand.

## To clear

The coverage guard, and a re-run showing a pure-black frame against `PPSA03274` no longer reports
DOMINATED while the overlay counterexamples still do. Everything else in this PR is ready and the
retraction should land.

mattias800 (commented):
# Independent review, round 3 — #3060 @ `85c371d2`

**REJECTED** — one blocking finding, and it is the mirror image of the bug you asked me to check
hardest. You asked whether a verdict can still come from a comparison that did not clear both gates.
It cannot; I traced every path and that half is correct. **The defect runs the other way: a true
match is DISCARDED by the coverage gate, and the verdict is then drawn from a different asset — so
the tool now clears a frame that is byte-identical to the game's own artwork.** The fix is one line
and I verified it at scale before recommending it.

The structure insight is right and the white twin is a real generalisation. That part stands.

## BLOCKING — the tool passes a frame that IS a stored picture, byte for byte

Take `1/PIX/LEGAL_2.DDS` out of the Tomb Raider dump, hand it back as the candidate screenshot, and
run the tool unmodified:

```
compared 41 same-size asset(s) (6925 wrong-size, 0 undecodable, 0 featureless)
  mean=   0.00/255  within8=100.00%  coverage=  4.09%  1/PIX/LEGAL_2.DDS      <-- the frame IS this
  mean=  12.75/255  within8= 23.51%  coverage=  7.05%  1/PIX/LEGAL_1.DDS
  ...
no asset explains this frame (closest informative comparison 28.48/255; highest overlap
1.70% with 3/PIX/HD/LA_UNDERSEA.DDS at 89.27% coverage, under the 90.00% dominance bar).
```

**exit 0.** All four of `LEGAL_1..4` do this.

Two things make it worse than a plain miss. The tool prints `mean=0.00/255 within8=100.00%` **in its
own table** and then contradicts it. And the pass line quotes a *different* asset — 28.48/255 — as
"the closest informative comparison", so a reader who trusts the verdict sees a reassuring number
that belongs to an unrelated file. The true match is hidden behind a summary.

**Root cause,** `prosper/tools/evidence/prerender_check.py:162,170`:

```python
usable = [row for row in scored if row[3] >= min_coverage]
exact  = min(usable, key=lambda row: row[0])
```

`mean` is a **whole-frame** statistic. It does not depend on coverage and cannot be corrupted by
darkness — two structured images with mean 0.00 are the same image, whatever their bright fraction.
Gating the EXACT path on coverage therefore removes true matches and buys nothing. `LEGAL_2` is
structured (luma sd **49.61**, far above your 3.0 floor) but only **4.09 %** of it is above the
content floor, because it is white text on black — so it never reaches `usable`.

**The affected class is exactly trap 230's:** full-screen pre-rendered pictures on dark backgrounds —
legal and notice screens, credits, title cards on black, loading screens with dark art. Those are
*more* likely to be blitted than the average asset, not less.

You fixed "a verdict drawn from a comparison that did not clear the gates". This is "a comparison
that cleared everything that matters, thrown away, and the verdict drawn from something else".

### Verified fix — the coverage gate is redundant, the structure gate already does the work

I tested this rather than assuming it. With `--min-coverage 0`, i.e. the structure gate alone:

| check | result |
| --- | --- |
| synthetic flat frames (black / near-black / dark / white, 1080p + 4K) x **all 55 dumps**, 342 runs | **0 false matches** |
| repo sweep, 94 mapped screenshots | same **2** DOMINATED (both title screens), no new flags |
| highest overlap among all non-flagged genuine renders | **3.47 %** — the headroom to the 90 % bar is enormous |
| `LEGAL_1..4` fed back as candidates | **EXACT MATCH, exit 2** — correct |
| retracted MANSION frame | EXACT MATCH |
| 2 % / 5 % overlay counterexamples | DOMINATED 98.02 % / 94.89 % |
| my round-2 false positives (pure black / white / grey / navy vs `PPSA03274`, `PPSA26414`, `PPSA02058`, `PPSA20800`) | all **exit 1 FEATURELESS** |

Not one of my false positives comes back, because **every one of them was already killed by the
structure gate** — on the candidate side (a flat capture) or the asset side (a flat `pic1.dds`). The
coverage floor was doing no work against them.

Minimal correct change: draw the EXACT verdict from `scored`, not `usable`.

```python
exact    = min(scored, key=lambda row: row[0])     # whole-frame stat: coverage-independent
dominant = max(usable, key=lambda row: row[1])     # small-sample stat: keep the coverage floor
```

That split is principled rather than expedient: the overlap is measured on a subset and a tiny
subset is noisy, so a coverage floor belongs there; the mean is measured on every pixel and needs no
such protection. `scored` can only contain structured-vs-structured pairs (featureless assets
`continue` at `:129`, a featureless candidate returns at `:104`), so nothing flat can reach the EXACT
path even with the filter gone — which is why the blunt `--min-coverage 0` run above is already clean.
Keep printing coverage either way; it is genuinely informative.

## The control-flow question you asked me to check hardest — correct, in that direction

Traced end to end and I could not find a route to a verdict from an ungated comparison:

- featureless **candidate** → `:104` returns 1 before any asset is opened
- featureless **asset** → `:129` `continue`, never appended to `scored`, so it cannot reach any report
- `scored` empty → `:143` returns 1 (`NOTHING WAS COMPARED`)
- `usable` empty → `:163` returns 1 (`NOTHING INFORMATIVE WAS COMPARED`), and this is a per-row
  filter, not the global `max()` that produced your EXACT-MATCH-on-black bug — that fix is real
- both verdicts (`:170`, `:180`) and the pass line (`:191`) read only from `usable`

The gates bind on both paths as you intended. The problem is that one of them binds where it should
not.

## Your "have I made it too quiet?" question — answered by the above

Yes, and the cost is worse than quietness. Some of the frames that moved into "not comparable", and
some that stayed in "pass", are frames the tool had already correctly identified and then discarded.
The rise from 47 to 54 is not purely the gate working; part of it is the gate deleting true matches.
Your instinct to ask was right.

## Structure gate calibration — I tried to break it and could not

This was my main worry about the new approach: a genuine artwork asset excluded as featureless would
be a false pass by another route. It does not happen. Across **1490** large assets in all dumps,
exactly **18 (1.2 %)** fall below sd 3.0, and every one is genuinely flat — sd ≤ 1.134, almost all
exactly 0.000: `sce_sys/pic1.dds` / `pic2.dds` variants, a solid `bg_texture.dds`, a solid mask,
`screen_mask384.tga`. The next real asset sits at sd **4.653**. Your 3.0 threshold lands in an empty
band with clean margin on both sides — well chosen, not arbitrary. Worth putting that margin in
`AGENTS.md`, since a future reader will otherwise wonder whether 3.0 was picked by eye.

## Non-blocking

1. **A true positive regressed between heads.** The 10 %-of-height overlay was DOMINATED at 90.06 %
   at `419b674a` and now passes. Sweeping the bar height: 6 % → 93.95 % DOMINATED, 8 % → 91.85 %
   DOMINATED, **10 % and above → exit 0**. Realistic loading-screen furniture is well under 8 %, and
   the pass line now prints the highest overlap so a reader can see it — but there is a silent band
   where a stored asset explains most of the frame and the verdict still says "no asset explains this
   frame". Consider a third report line when the overlap is high but under the bar: *"closest 88 % —
   inspect before publishing"*.
2. **The printed table shows rows the verdict cannot use.** `:154` iterates `scored`; the `LEGAL_2`
   output above puts `mean=0.00/255` at the top of the table above a verdict saying nothing matches.
   If any coverage filter survives, mark those rows (`(excluded: coverage 4.09%)`) so the table and
   the verdict cannot contradict each other.
3. **Doc figure drift in trap 230.** The row still says *"next best here was 52.77"*. That was the
   luminance-weighted mean; the tool now computes a per-channel mean and prints **50.86** for
   `LONDON.DDS`. Someone re-running the documented command gets a different number than the doc.
   While you are there: that parenthetical now sits immediately after the title-screen waiver
   sentence, where it reads as though it describes the title screens rather than MANSION — it wants
   to move back beside the MANSION measurement.

## Clean at this head

Gates: `--ordered --table-header Instrument --baseline origin/master` → exit 0; all **152** tracked
`.md` → exit 0; `check_merge_result.py` → clean merge, passes on the merge result; `git diff --check`
→ 0; `Claude-Session:` trailers → 0. The retraction, the trap row's provenance correction, the
`COMPATIBILITY.md` rung note and the title-screen waiver are all accurate as they stand — the waiver
reads well and names both files, which closes it permanently.

## To clear

The one-line `exact`/`dominant` split, plus a re-run showing `LEGAL_1..4` fed back as candidates
report EXACT MATCH while the synthetic flat sweep stays at 0 false matches. Nothing else is
outstanding; the retraction should land.

mattias800 (commented):
# Independent review, round 4 — #3060 @ `b437cba8`

**APPROVED.** The round-3 blocking finding is fixed and I re-verified it directly: `LEGAL_1..4` handed
back as their own candidates now all report **EXACT MATCH, exit 2**. Every true positive I have built
across four rounds is caught, every false positive is rejected, the docs are accurate, and all gates
are green.

One real finding remains, on the parameter you asked me to look hardest at. It is **not blocking** —
I explain why below rather than just asserting it — but it should be changed, and the change is
deleting a number.

## `--dominated-mean-max 40` — it buys nothing and costs a real window

You were right to flag this one for scrutiny. It does not do what it was added to do.

**The case it was added for is already rejected by the bar.** Run
`little-nightmares-2-tarsier-logo.png` against its actual dump (`PPSA02154`, not `PPSA05143` — worth
checking your mapping):

```
with the ceiling:     exit 0  "no asset combines a usable overlap with a mean under 40.00/255"
ceiling disabled:     exit 0  "highest overlap 50.13% ... under the 75.00% dominance bar"
```

Same verdict either way. At **50.13 %** overlap it never reaches the 75 % bar, so the ceiling only
changed the wording of the pass line. That is also why your corrected margin figure moved so much:
the case was never a near-miss on the verdict, only on the reported statistic.

**And the two gates are mathematically inconsistent.** Overlap ≥ 75 % *bounds* the mean: at most 25 %
of informative pixels differ, each contributing ≤ 255, and non-informative pixels contribute ≤ 16.

```
coverage 100%: mean <= 0.25 * 255            = 63.75
worst case   : mean <= 0.25 * 255 + 0.75*16  = 75.75
```

So anything reaching the bar has mean ≤ ~64–76, and the motivating case at mean 118.45 could never
have reached it. **A ceiling below that bound can only ever remove cases the bar accepts** — it can
produce false negatives and cannot prevent a false positive.

**Measured, not just derived.** I probed real 1080p structured assets by growing a bright caption
panel from the bottom and reading the mean at the point overlap is still ≥ 75 %:

| mean at the bar | overlap | coverage | panel | asset |
| **54.33** | 75.04 % | 99.78 % | 25 % | `PPSA20052` `assets/roms/border_gbc.png` |
| **53.41** | 75.07 % | 98.79 % | 25 % | `PPSA20447` `bbq/content/movies/tutorial/…` |
| **52.91** | 75.08 % | 86.58 % | 22 % | `PPSA20447` `bbq/content/movies/tutorial/…` |
| **40.51** | 75.08 % | 96.96 % | 24 % | `PPSA20447` `bbq/content/movies/tutorial/…` |
| 37.63 | 75.00 % | 100.00 % | 25 % | `PPSA03831` `raw/ui/rpl_texture/uiRANGE_img…` |

4 of 13 probed assets land above 40 while still clearing the overlap bar. A capture of one of those
with a bright caption panel over ~22–25 % of the frame is a frame the overlap bar would flag, and the
ceiling silently drops it to exit 0.

**Fix:** delete `--dominated-mean-max`, or derive it from the bar rather than fixing it independently
— `(1 - overlap/100) * 255 + 16`, which is ~80 at the default and by construction never contradicts
the bar. Deleting is simplest and, per the test above, changes no verdict on real data.

### Why this is not blocking

I want to be explicit, because I have rejected this PR three times and should say what changed.

The round-3 defect **passed a frame byte-identical to the game's own artwork** — the tool's exact
purpose, failing on an ordinary input. This one shifts an existing boundary inward. The tool already
misses overlays beyond ~25 % of the frame because the *overlap* falls under 75 % there; the ceiling
moves that edge from ~25 % to ~22 % on some assets. It makes two gates disagree, which is worth
fixing on principle, but it does not open a qualitatively new hole, and the failure direction is a
missed screen rather than a manufactured claim. Against that: the retraction itself has been correct
and independently verified since round 1, and holding a public correction for a fourth round over a
parameter that changes no verdict on any real input would be the wrong trade.

## Re-verified at this head

| arm | result |
| --- | --- |
| `LEGAL_1..4` as their own candidates | **EXACT MATCH ×4** (the round-3 blocker) |
| retracted MANSION frame | EXACT MATCH |
| overlays 2 / 5 / 10 / 12 / 15 / 20 % | DOMINATED 98.02 / 94.89 / 89.75 / 87.75 / 84.59 / 79.40 % |
| both title screens | DOMINATED 95.94 / 94.87 % |
| synthetic flats (black / white / 1080p / 4K) × 4 dumps | exit 1 FEATURELESS, all |
| `stray-brightness-calibration`, `ln2-tarsier` | exit 0, correctly |
| `dominant_pool` empty | handled at `:219` — no `None` dereference |

The 90 → 75 move fixes the regression I raised in round 3: catch now extends to a 20 % overlay where
it stopped at 8 %.

**The verdict split is correct.** `exact = min(scored, …)` at `:187` runs before the `usable` check at
`:196`, so an exact match is reported even when nothing clears coverage; `dominant` draws from
`usable` filtered by the ceiling. Featureless assets still cannot enter `scored` (`:129`), a
featureless candidate still exits at `:104`. I could not find a route to a verdict from an ungated
comparison, in either direction.

## The other items

- **Table marking** (`:161`) — `[overlap unusable: coverage < 25%]` does exactly what was needed; a
  strong `within8` can no longer sit silently above a pass.
- **Trap 230's 52.77 → 50.86** with the luma-metric note is right, and the parenthetical is back
  beside the MANSION measurement where it belongs.
- **The structure-floor calibration is recorded accurately** (18 of 1490, all sd ≤ 1.134, next real
  asset 4.653). Thank you for attributing it — but do treat it as re-derivable rather than settled by
  my having run it once; that is the same trap as the margin figure.
- **Your margin correction is the right instinct and I will match it:** my 3.47 % was measured over a
  map that had no entry for `little-nightmares-2` at all, so that screenshot was silently skipped
  rather than scoring low. The number was not wrong, its *scope* was — which is exactly why you were
  right not to inherit it. Your 50.00 % came from a mapping that did include it.

## CI

**I agree it is flaky, and there is direct evidence rather than only absence of a mechanism.**
`master`'s own most recent CI run (`33050236613`) failed on the **same job**, `macOS (x86_64 /
Rosetta 2)`, with no involvement from this branch. A docs-plus-Python diff that compiles nothing
cannot move a semaphore *timing* test on Rosetta. Treating it as flaky is correct.

Note for the merge step: `Docs`, `Contribution shape`, `Linux App` and `Linux App Package` are green
at this head, but Linux, Windows MinGW, both macOS jobs and Windows App were still **pending** when I
checked. Confirm the required set has actually reported before merging — a pending check is not a
passing one.

## Verdict

**APPROVED** at `b437cba8`. Land the retraction. I would change `--dominated-mean-max` before or
straight after merge — deleting it is a no-op on every real input I tested and closes the window —
but that is a quality fix on a screening heuristic, not a correctness gate on this PR.

Four rounds on what began as a one-line retraction, and the tool that came out of it is now the only
automated check this project has against the failure it was written to record. That was worth it.

mattias800 (commented):
# Addendum to the round-4 approval — the head moved while I was reviewing

**APPROVED** at `be36ddf5`. This re-states the round-4 verdict against the head GitHub actually
stamped it on, because those were not the same commit and the difference is exactly the mechanical
trap `CLAUDE.md` warns about.

**What happened.** I reviewed `b437cba8`. Between my last check and `gh pr review` posting, the
branch was rebased onto a newer `origin/master`, so every commit was rewritten
(`2db08936 / 419b674a / 85c371d2 / b437cba8` → `37ebaefc / e0e01ded / 34437f15 / be36ddf5`). My
review body says "APPROVED at `b437cba8`" while the registered review carries `commit_id
be36ddf5` — an approval whose text names a commit that is no longer on the branch. Per the charter,
that has to be resolved **by content**, never assumed in either direction. So I resolved it.

**The PR's own content is unchanged — verified, not assumed:**

- All seven surviving files have **identical blob hashes** at `b437cba8` and `be36ddf5`
  (`BLOG.md`, `COMPATIBILITY.md`, `GAME_COMPAT_ORCHESTRATION.md`, `TOMB_RAIDER_STATUS.md`,
  `tools/AGENTS.md`, `tools/evidence/AGENTS.md`, `prerender_check.py`), and the retracted screenshot
  is still deleted.
- `git diff <base>..<head>` computed against each head's *own* merge-base is **byte-identical**
  between the two.
- The entire difference is two commits the rebase pulled in from master, both already reviewed and
  merged on their own PRs: `8a2f7d51` (#3044, guest semaphore timedwait off the winpthreads tick)
  and `6f6ec58a` (#3063, image-source snapshot attribution). Neither is this PR's work and neither
  touches anything it changes.

So the round-4 approval transfers intact. Nothing in it needs revisiting.

**Re-run at `be36ddf5` rather than carried over:**

| check | result |
| --- | --- |
| `LEGAL_2`, `LEGAL_4` as their own candidates | EXACT MATCH, exit 2 |
| retracted MANSION frame | EXACT MATCH, exit 2 |
| 20 % overlay | DOMINATED 79.40 %, exit 2 |
| pure-black frame vs `PPSA03274` | exit 1 FEATURELESS |
| `--ordered --table-header Instrument --baseline origin/master` | `1..230, unique and ascending, none of the 229 base rows deleted` |
| all tracked `.md` | exit 0 |
| `check_merge_result.py` | clean merge, gate passes on the merge result |
| `git diff --check` | exit 0 |

**One incidental note worth having in the record:** the rebase pulled in `8a2f7d51`, which *is* the
`kernel_sem_timedwait` work. The macOS failure we agreed to treat as flaky was on that test, and this
head now carries master's change to it — so whatever the fresh CI run reports is being produced by
different code than the run we were discussing. That does not change the flaky assessment (master's
own run `33050236613` failed the same `macOS (x86_64 / Rosetta 2)` job independently of this branch),
but do read the new run on its own terms rather than as a re-test of the old one.

Everything from round 4 still stands, including the one non-blocking item: **`--dominated-mean-max 40`
should be removed or derived from the bar.** It changes no verdict on any real input I tested — the
case it was added for scores 50.13 % overlap and is already rejected by the 75 % bar — while a ceiling
below the bar's own analytic bound (mean ≤ 63.75 at full coverage) can only ever drop cases the bar
accepts.

**APPROVED** at `be36ddf5`. Confirm the required CI set has actually reported — several jobs were
pending when I looked, and a pending check is not a passing one — then land it.

mattias800:
## Caveat on the sweep figures above

The repo-wide sweep in the comments above used a **scratch, approximate** mapping from screenshot
filename to title id (loose substring matching against `COMPATIBILITY.md`). Review found it wrong in
at least one case — `little-nightmares-2-tarsier-logo.png` was mapped to `PPSA05143`
(*Little Nightmares III*) instead of `PPSA02154` — and that error is what produced the 50.00%/mean
103.88 figure I briefly used to justify a mean ceiling that has since been removed as unsound.

The same defect cut the other way for the reviewer, whose "worst genuine render 3.47%" was measured
over a map that silently skipped that screenshot.

So read the sweep as: **among the screenshots it did map correctly, no further exact pre-rendered
blit was found, and two title screens are DOMINATED and correctly captioned.** The
"could-not-compare" count is a floor, not an exact figure, and a mis-mapped screenshot is compared
against the wrong dump rather than being reported as unmapped.

That mapping is deliberately **not** part of this PR — `prerender_check.py` takes the dump root as an
argument precisely so it never has to guess which title a screenshot belongs to. Anyone re-running
the sweep should supply the mapping deliberately rather than inferring it from filenames.

mattias800 (commented):
# Independent review, round 5 — #3060 @ `c011e97f`

**APPROVED.**

## First: the detached HEAD was mine, and the diagnosis is exactly right

`git checkout` in that worktree was my doing — twice — and it detached HEAD from the branch, which is
why your commit landed detached and your push was a no-op. My error, no qualification. Thank you for
recovering it rather than just reporting it.

**I have re-attached it**, using `git symbolic-ref HEAD refs/heads/docs/tr-retract-loading-screen`
rather than another checkout: branch tip, HEAD and remote tip were all `c011e97f` with a clean tree,
so re-pointing the ref touched neither the index nor the working tree (verified after: the working
`prerender_check.py` still hashes to `HEAD:`'s blob, `git status` empty). The worktree now reads
`c011e97f [docs/tr-retract-loading-screen]`, so your next commit lands on the branch. The hazard I
created is gone.

Taking the method note: `git show <sha>:<path>`, `git diff A..B` and `git hash-object` for the rest of
this review. Nothing below moved HEAD.

### You asked whether my earlier findings could have been affected. I audited it by content, not memory

Every reviewed head still exists as an object, so I checked each one's tool for the feature my
finding turned on — no checkout involved:

| commit | blob | `--overlap` | structure gate | mean ceiling | `exact` drawn from |
| `2db08936` (r1) | `23beaa78` | — | no | no | n/a (mean-only verdict) |
| `419b674a` (r2) | `6caedf3e` | 90.0 | no | no | n/a |
| `85c371d2` (r3) | `157045bb` | 90.0 | yes | no | **`usable`** |
| `b437cba8` (r4) | `4ac3af98` | 75.0 | yes | **yes** | `scored` |
| `c011e97f` (r5) | `48c94983` | 75.0 | yes | no | `scored` |

Each row is exactly the code my finding for that round describes — round 1's mean-only verdict,
round 2's missing structure gate, round 3's `exact = min(usable, …)` that hid `LEGAL_2`, round 4's
ceiling. **No finding was measured against a tree other than the one I attributed it to.** And the
file I executed for this round's arms hashes to `48c94983`, byte-identical to `c011e97f`'s blob.

## The removal is correct and complete

- `--dominated-mean-max`, `dominant_pool`, and the now-unreachable `dominant is not None` / `else`
  branches are all gone. No dead code left behind, which is easy to miss when deleting a gate.
- `dominant = max(usable, …)` is immediately preceded by `if not usable: return 1`, so `max()` cannot
  raise on an empty sequence.
- `grep` for `dominated_mean_max`, `dominated-mean-max`, `dominant_pool`, `103.88`, `50.00%` across
  `tools/`, `docs/`, `COMPATIBILITY.md` and `BLOG.md`: **no dangling references.** The docstring,
  usage line and `AGENTS.md` all agree with the code.
- Passing the removed flag is ignored rather than fatal (unknown `--` args are dropped by the arg
  filter). Acceptable for a screening script; worth knowing if you ever add strict parsing.

**The `AGENTS.md` entry is the right shape** — it records that the idea was tried, why it is
attractive, and why it fails, so the next person who proposes it meets the disproof instead of
re-deriving it. That is more valuable than the parameter would ever have been.

## Full arm matrix at `c011e97f`

| arm | result |
| --- | --- |
| retracted MANSION frame | EXACT MATCH, exit 2 |
| `LEGAL_1..4` as their own candidates | EXACT MATCH ×4, exit 2 |
| overlays 2 / 5 / 10 / 12 / 15 / 20 % | DOMINATED 98.02 / 94.89 / 89.75 / 87.75 / 84.59 / 79.40 % |
| both title screens | DOMINATED 95.94 / 94.87 % |
| **the window the ceiling created** — white panel 20 %, white 22 %, black 22 % | **DOMINATED 79.25 / 77.24 / 77.38 %, all exit 2** (were exit 0) |
| pure black vs `PPSA03274` / `PPSA26414` / `PPSA02058`; white vs `PPSA03274` | exit 1 FEATURELESS |
| `stray-brightness-calibration` | exit 0 |
| `little-nightmares-2-tarsier-logo` vs its **correct** dump `PPSA02154` | exit 0 |
| no comparable asset / unreadable candidate / bad `--overlap` | exit 1, exit 1, exit 1 |

The three captioned-panel frames are the concrete win: those are the cases I measured at mean
40.5–54.3 above the bar, and every one of them is now caught.

Gates: `--ordered --table-header Instrument --baseline origin/master` → `1..230, unique and
ascending, none of the 229 base rows deleted`; all **152** tracked `.md` → exit 0;
`check_merge_result.py` → clean merge, gate passes on the merge result; `git diff --check` → 0;
`Claude-Session:` trailers → 0. `py_compile` clean.

## On the mapping error

You found the more interesting half of it. My round-2 `3.47%` was low because the map had no
`little-nightmares-2` entry and silently skipped the file; your `50.00%` was high because the same
map sent it to `PPSA05143`. **One defective mapping, two opposite wrong answers, and each of us
believed the one that suited us** — mine made the headroom look comfortable, yours made a gate look
necessary. Neither error was detectable from its own output; both needed the other reading to
surface.

Saying in the PR body that the sweep's title mapping is approximate and its "could not compare" count
is a floor is the right disclosure. I would add one line naming the mechanism — *an unmapped
screenshot is skipped silently, so a low count and a high score can both be mapping artefacts* —
because the count being a floor is the symptom, and the silent skip is the thing a future sweep
author needs to guard against.

## Merge

**APPROVED** at `c011e97f`. `Docs` and `Contribution shape` are green; the build jobs were still
pending when I looked, and you have already said you will not merge on a pending check — that is the
only thing left.

Five rounds, and the shape of it is worth recording: the retraction was right on day one, and
everything after was the tool built to stop it happening again. It survived four adversarial passes,
each of which found a real defect, and the last two defects were introduced by the fixes for the
first two. That is what building an instrument honestly looks like, and the `## Ruled out` and
`AGENTS.md` entries now carry the disproofs so the next person inherits them.

mattias800 (commented):
# Correction to my own measurement — the "4 of 13" figure is understated

**APPROVED** still stands at `c011e97f`; nothing about the removal changes. But a background probe I
started during round 4 and believed had died has now completed, and **it contradicts the figure I
gave you — which is currently shipped in a code comment and in `AGENTS.md`, attributed to me.**
Correcting it before merge.

## What the two probes say

| | my `ceil.py` (what I published) | the completed probe |
| --- | --- | --- |
| assets probed | **13** | **233** |
| above the 40 ceiling while overlap ≥ 75 % | **4 (31 %)** | **100 (42.9 %)** |
| worst mean observed | **54.33** | **59.25** |

Same direction, and the finding is *stronger* than I reported — but the numbers in the tree are wrong.

**I verified the new probe rather than trusting it,** recomputing its top hit from scratch:
`PPSA03001` `sce_sys/pic0.dds` (3840x2160, luma sd 8.21, so the structure gate includes it) with a
24 % caption panel gives overlap **75.52 %**, mean **59.24**, coverage **97.96 %** — the overlap bar
catches it, a 40 ceiling drops it. Matches the probe to two decimals.

## Why my sample was wrong, which is the part worth recording

`ceil.py` filtered `if im.size != (1920,1080)`. **It excluded every 4K asset — and that is where the
worst means live.** In a 20-dump sample: **6** structured 1920x1080 assets against **79** structured
3840x2160 ones. I sampled roughly 7 % of the relevant population, and specifically the 7 % that
excluded `sce_sys/pic0/pic1/pic2` store art, which supplies almost the whole top of the new list.

This is the same failure as the charter's positive-control trap: my sample could not express the
worst case, so its silence about that case was guaranteed rather than informative.

**Third scoping error of mine in this review, all in my own instruments** — the round-1 int16
overflow in the luminance weighting, the round-2 map with no `little-nightmares-2` entry, and now a
resolution filter. None was a reasoning error; every one was an instrument that quietly could not see
part of its domain, and in each case the output looked entirely normal. Worth naming as a pattern,
since I raised the same class against your gates three times.

One reassurance from it: the observed maximum **59.25** sits just under my analytic bound of **63.75**
at full coverage. Two independent methods agreeing at the boundary is decent evidence the bound is
right, and it is the bound — not either count — that actually settled the removal.

## What to change

Two citations, both a comment edit:

- `prosper/tools/evidence/prerender_check.py:203-204`
- `prosper/tools/evidence/AGENTS.md:47`

Both read *"4 of 13 real assets … reach mean 40.5-54.3"*. Suggested replacement:

> 100 of 233 real assets carrying a bright caption panel reach mean 40.5–59.2 while still exceeding
> the bar, and a 40 ceiling silently dropped every one.

If you would rather not restate a number of mine a second time, *"roughly 43 % of real assets"* with
the bound cited alongside is safer and ages better — the bound is derivable, the count is a property
of whatever sample someone drew.

## Verdict

**APPROVED** at `c011e97f`. This is a comment-and-doc correction to an already-correct change, not a
new finding: the ceiling is gone, the disproof holds, and the corrected figure only strengthens it.
Land it with the numbers fixed — an understated figure in the very `AGENTS.md` paragraph that exists
to stop someone re-proposing the ceiling is exactly the kind of weakened disproof that invites the
re-proposal.

Issue #2863, opened 2026-08-21
GTA V: a user reports in-game PERFORMANCE MODE renders some world content — likely the RT path, and an independent corroboration of the skip A/B


mattias800:
## The report

A user who ran GTA V on prosper switched the **in-game graphics setting to Performance Mode** and
reports that **some world content began rendering** — "not much, but a little".

**This is a user report relayed second-hand and has NOT been reproduced here.** It is filed as a lead,
with the experiments that would confirm or kill it, not as a finding. Nothing below should be quoted
as measured until it is.

## Why it is worth taking seriously rather than filing away

It converges with the record from a completely independent direction.

- The traversal table `0x413dc6700` walks **is a ray-tracing BVH** — an LBVH over 1,032 leaves, with
  the "tag" this investigation chased for weeks identified as the RDNA2 `NODE_TYPE`
  (`GTA5_STATUS.md` § "REFRAME: this is a RAY-TRACING BVH").
- Its **consumers do not compile**: `IMAGE_BVH_INTERSECT_RAY` is MIMG opcode `0xe6`, and
  `0x205b654a00` / `0x205b657200` both fail `unresolved-operand` on it.
- The one lever that has ever produced real scene content in this title is
  `PROSPER_COMPUTE_SKIP_PROGRAM=0x413dc6700` — zero device losses, and sun, lens flare, radar geometry
  and tutorial text appear.

If Performance Mode disables or reduces the ray-tracing path, then the user has reached **the same
outcome by a different mechanism**: ours declines the program from inside the emulator, theirs stops
the *game* from asking for it. Two independent routes to "remove the RT work → world content appears"
is much stronger evidence than either alone, because the emulator lever and the game setting share no
machinery.

**The mode semantics are an assumption, not a fact.** This title ships more than one performance
profile and the exact per-mode RT behaviour has not been checked here — that is part of what the
experiments below are for.

## What would settle it

1. **Reproduce it.** A route that reaches the graphics settings and selects Performance Mode. There
   is currently **no such route**: `scripts/gta5/reach-story-mode.pad` navigates the ONLINE→STORY tab
   and never enters settings. This is the main piece of work, and it is input work rather than
   renderer work.
2. **The clean A/B, once the route exists**: same build, same seed, Performance vs the default mode,
   `tools/screenshot`, comparing rendered world content. Name the frontend and the route in the
   result (instrument trap 127).
3. **Does the hanging dispatch still run?** With Performance Mode selected, check whether
[...]

PR #3063, opened 2026-08-27
perf(compute): attribute the image-source snapshots — 89% are read/modify/write, and 0% are what I expected

mattias800:
## What this adds

Per-reason attribution for the compute image-source snapshots, so the existing volume counter can be
acted on:

```
[render-timing] compute_snapshot_reason host_data=0 0.0 MiB rmw=6572 71502.4 MiB
                changed=1744 8212.5 MiB other=+835.8 MiB
```

`remember_image_source_snapshot` has **ten call sites**; three are tagged here because they are the
ones that carry volume, five more are live on Linux but cold one-shot retentions (they land in
`other`), and two are inert in this configuration. The tagged three have **very different prospects**
— an early `if (key.host_data)` return that never consults the adaptive path, targets that require an
exact guest baseline, and the ordinary changed-content path. Nothing separated them, so choosing a
target meant guessing.

## What it establishes

- **`host_data` is exactly zero — on every platform, not just Linux.** `ShaderResource::host_data`
  is replay-only owned backing; production tables leave it zero, and every writer is in
  `gpu_capture.cpp` or `gpu_replay/`. So this branch cannot fire in a live run at all. It is the
  branch I expected to dominate, because it returns *before* the adaptive path is consulted; an
  optimisation aimed there would have been worth nothing.
- **88.8% required an exact guest baseline** — the branch guarded by `source_snapshot_required`,
  which the live caller supplies as `!seed_skip`. That is mostly read/modify/write and partial
  storage targets, whose prior bytes are observable by the next dispatch, but `!seed_skip` **also**
  catches coverage proving/re-proving, uncached-proof first sightings and thread-count failures, so
  the counter is a proxy for "needs a baseline" rather than a direct RMW test. The recurring member
  is bounded: `kSeedReproveInterval` defaults to 256, so proving frames are ~0.4%.
- `other` is **1.0%**, consistent with the five untagged live sites being cold one-shot retentions.
  It prints **signed** so an over-attribution shows as negative rather than being clamped into
  looking clean — the rule the `res_buffer_*` leaves already follow.

Split, recomputed: **88.77 / 10.20 / 1.04%**.

## Why guessing was not acceptable

Four attempts to name this frame's cost have now been wrong, which is the case for measuring rather
than reasoning:

| attempt | outcome |
| --- | --- |
| texture cost dominates | **stale** — a remembered 3.8 s figure the rendering series had obsoleted; it is ~384 ms now |
| the write-watch promotion budget blocks promotion | **falsified** — 16x raise moved snapshot volume 3%, and flips/s did not improve (1.19 → 1.05, an 11.8% drop; one run each, so treat as "no gain", not as a regression) |
| the adaptive storage-result path needs adding | **already there and working** — disabling it *doubles* snapshots per pool-hit, 0.0715 → 0.1404 |
| the `host_data` branch dominates | **zero**, per this PR's own counters |

The second and third required normalising per pool-hit: the raw counters are cumulative, so comparing
totals between runs of different progress measures run length, not the change.

## What this PR does NOT claim

An earlier version of this body opened with "48% of all CPU cycles in one `memmove`" and discussed
the snapshots as though that memmove were them. **That link was never established and the arithmetic
refutes it:** 65 core-seconds went to `memmove` in the perf window, while 69.8 GB of snapshots scales
to ~16.5 GB inside it — 1.0–2.1 core-seconds, i.e. **~2–3%**. Two large numbers placed side by side,
with the cause inferred. Retracted.

So: these snapshots are 69.8 GB of memory traffic, worth knowing and worth attributing, but on
current evidence they are **not** why GTA V runs at ~1.2 guest flips/s. The dominant `memmove` caller
is still **unidentified** — `perf` could not resolve it because the binary reported `(deleted)`.
Finding it is the next step and is not this PR.

## These are single runs

Every figure here is one run of one route on one host: the 48%/6% cycle shares, the 98.7 ms
`gpu-device`, the ~1.2 guest flips/s, and the 88.8/10.2/1.0 split. They are stated to show the shape
of the problem and to justify attributing it, not as characterisations of the title. The mean
snapshot is **10.88 MiB** (79 GB / 7452 copies at this reading); an earlier draft said "9.5 MiB",
which was a different run's mean and is not derivable from anything the tool prints.

## Scope

Instrumentation only. Three counters, one report line, and a `SnapshotReason` argument defaulted to
`Unattributed` so no existing call site changes behaviour. No policy, no allocation, no ordering, no
locking.

```
cmake --build prosper/build-linux -j6     # 0 errors
ctest --no-tests=error                    # 100% tests passed out of 306
```

`git diff --check` clean.

Refs #2863

### DIFF
--- a/prosper/frontends/shared/live/live_compute.cpp
+++ b/prosper/frontends/shared/live/live_compute.cpp
@@ -1151,6 +1151,20 @@ struct VulkanComputeContext {
     uint64_t storage_result_snapshot_bytes = 0;
     uint64_t image_result_snapshot_copies = 0;
     uint64_t image_result_snapshot_bytes = 0;
+    // WHY each snapshot was taken. The total above cannot be acted on: measured on GTA V gameplay
+    // it is 35 GB in one routed run and 48% of all CPU cycles land in the memmove underneath it,
+    // but the adaptive storage-result path already halves the rate (measured: disabling it doubles
+    // snapshots/pool-hit from 0.0715 to 0.1404), so the remaining traffic is whatever that path
+    // never sees. Two branches reach the copy without consulting it -- an early return for
+    // host-data sources, and read/modify/write or partial storage targets, which genuinely need an
+    // exact baseline. Which of those dominates decides whether there is anything left to win here,
+    // and no existing counter separates them. Same reasoning as the res_buffer_* leaves.
+    uint64_t snapshot_reason_host_data_copies = 0;
+    uint64_t snapshot_reason_host_data_bytes = 0;
+    uint64_t snapshot_reason_rmw_copies = 0;
+    uint64_t snapshot_reason_rmw_bytes = 0;
+    uint64_t snapshot_reason_changed_copies = 0;
+    uint64_t snapshot_reason_changed_bytes = 0;
     VkDescriptorSetLayout compare_descriptor_layout = VK_NULL_HANDLE;
     VkShaderModule compare_shader = VK_NULL_HANDLE;
     VkPipelineLayout compare_pipeline_layout = VK_NULL_HANDLE;
@@ -1677,9 +1691,12 @@ struct VulkanComputeContext {
                write_watch_promotion_budget.try_consume(source_bytes);
     }
 
+    enum class SnapshotReason { Unattributed, HostData, ReadModifyWrite, ContentChanged };
+
     void remember_image_source_snapshot(CachedComputeImage& cached,
                                         const uint8_t* source, size_t bytes,
-                                        bool storage_result = false) {
+                                        bool storage_result = false,
+                                        SnapshotReason reason = SnapshotReason::Unattributed) {
         if (!source) return;
         cached.source_snapshot.assign(source, source + bytes);
         ++image_source_snapshot_copies;
@@ -1687,6 +1704,15 @@ struct VulkanComputeContext {
         if (storage_result) {
             ++storage_result_snapshot_copies;
             storage_result_snapshot_bytes += bytes;
+        }
+        switch (reason) {
+        case SnapshotReason::HostData:
+            ++snapshot_reason_host_data_copies; snapshot_reason_host_data_bytes += bytes; break;
+        case SnapshotReason::ReadModifyWrite:
+            ++snapshot_reason_rmw_copies; snapshot_reason_rmw_bytes += bytes; break;
+        case SnapshotReason::ContentChanged:
+            ++snapshot_reason_changed_copies; snapshot_reason_changed_bytes += bytes; break;
+        case SnapshotReason::Unattributed: break;
         }
     }
 
@@ -2006,7 +2032,8 @@ struct VulkanComputeContext {
         } else if (!upload_skipped && source && source_snapshot_required) {
             cached.write_watch_stable_validations = 0;
             remember_image_source_snapshot(
-                cached, source, key.guest_bytes, key.storage);
+                cached, source, key.guest_bytes, key.storage,
+                SnapshotReason::ContentChanged);
             // Do not trust the new mirror until the corresponding transfer completes. A failed
             // submit leaves this false, so the next use refreshes instead of skipping stale pixels.
             cached.content_valid = false;
@@ -2369,7 +2396,8 @@ struct VulkanComputeContext {
         if (key.host_data) {
             if (current_source)
                 remember_image_source_snapshot(
-                    cached, current_source, key.guest_bytes, true);
+                    cached, current_source, key.guest_bytes, true,
+                    SnapshotReason::HostData);
             cached.write_watch.reset();
             return;
         }
@@ -2389,7 +2417,8 @@ struct VulkanComputeContext {
             // contract because their prior guest bytes are observable by the next dispatch.
             if (current_source)
                 remember_image_source_snapshot(
-                    cached, current_source, key.guest_bytes, true);
+                    cached, current_source, key.guest_bytes, true,
+                    SnapshotReason::ReadModifyWrite);
             if (cached.write_watch && cached.write_watch.rearm()) return;
             cached.write_watch.reset();
             return;
@@ -10457,6 +10486,27 @@ bool execute_live_compute_items(const std::vector<prosper::gpu::ComputeItem>& it
                         (1024.0 * 1024.0),
                     (unsigned long long)g_compute_storage_transfer_seeds.load(
                         std::memory_order_relaxed));
+                // WHY those snapshots happened. `other` is the remainder and is printed SIGNED --
+                // a negative value means over-attribution, and clamping it would make a broken
+                // partition look like a complete one (the same rule the res_buffer leaves follow).
+                {
+                    const uint64_t attributed =
+                        context.snapshot_reason_host_data_bytes +
+                        context.snapshot_reason_rmw_bytes +
+                        context.snapshot_reason_changed_bytes;
+                    const double mib = 1024.0 * 1024.0;
+                    std::fprintf(stderr,
+                        "[render-timing] compute_snapshot_reason host_data=%llu %.1f MiB "
+                        "rmw=%llu %.1f MiB changed=%llu %.1f MiB other=%+.1f MiB\n",
+                        (unsigned long long)context.snapshot_reason_host_data_copies,
+                        static_cast<double>(context.snapshot_reason_host_data_bytes) / mib,
+                        (unsigned long long)context.snapshot_reason_rmw_copies,
+                        static_cast<double>(context.snapshot_reason_rmw_bytes) / mib,
+                        (unsigned long long)context.snapshot_reason_changed_copies,
+                        static_cast<double>(context.snapshot_reason_changed_bytes) / mib,
+                        (static_cast<double>(context.image_source_snapshot_bytes) -
+                         static_cast<double>(attributed)) / mib);
+                }
                 std::fprintf(stderr,
                     "[render-window] compute calls=%llu dispatches=%.1f avg_ms=%.2f\n",
                     (unsigned long long)window.calls,


mattias800:
## Correction from the author, before review spends time on it: the memmove link in the body is unsupported

The body opens with "48% of all CPU cycles in one `memmove`" and then discusses the snapshot copies
as though that memmove is them. **I never established that link, and the arithmetic says it is
false.**

```
perf window CPU        : 134 core-seconds   (90 s x 1.49 cores)
attributed to memmove  :  65 core-seconds   (48.41% of samples)
snapshot bytes in run  : 69.8 GB -> ~16.5 GB inside a 90 s window
at 8-16 GB/s that is   : 1.0-2.1 core-seconds  = 1.6-3.2% of the memmove time
```

So the image-source snapshots explain **~2-3%** of the `memmove` time, not the bulk of it. I put two
large numbers next to each other and let the reader — and myself — infer a cause. That is the same
error as the three the body already lists, and it deserves to be in that table rather than in the
framing above it.

**What is unaffected.** The attribution measurement stands on its own and is what the PR is actually
for:

```
host_data=0 0.0 MiB   rmw=6572 71502.4 MiB   changed=1744 8212.5 MiB   other=+835.8 MiB
```

`host_data` really is zero on Linux, RMW really is 89% of the snapshot bytes, and the remainder
really is 1%. Those are direct counter readings from the tagged branches, not inferences. The
conclusion that most of the snapshot volume is load-bearing rather than waste is likewise unchanged.

**What it costs.** The PR can no longer claim the snapshots are a large share of the frame. They are
69.8 GB of memory traffic, which is worth knowing and worth attributing, but on this evidence they
are not the reason GTA V runs at ~1.2 guest flips/s. The dominant `memmove` caller is **unidentified**
— `perf` could not resolve it because the binary reported as `(deleted)`, and I have not yet re-run
with symbols.

I have rewritten the body accordingly. Reviewers: please treat the "why guessing was not acceptable"
table as the honest part and the opening framing as retracted; the code diff itself is unchanged and
is still instrumentation-only.

mattias800 (commented):
**APPROVED** — head `52117a3c`. No blocking findings. The code is instrumentation-only as claimed, the arithmetic closes, and the actionable conclusion ("the dominant bucket is load-bearing under the current mechanism; the host-data bucket is worth nothing") survives every check I ran. What does **not** survive is part of the PR body's prose: two of its stated mechanisms are wrong and one label overstates. Those are the durable record, so please correct them before merge — nothing in the diff needs to change.

Verified locally at head: full build in the `gta5-patches` worktree, 0 errors; `git diff --check origin/master...HEAD` clean; `tools/env/check_diag_gates.py` all checks passed (no new `PROSPER_*` gate is introduced — the line rides the existing `PROSPER_RENDER_TIMING` gate). I did not re-run ctest.

---

### 4/5 first, because they are clean

**Behaviour change: none.** `SnapshotReason` is a by-value enum defaulted to `Unattributed` (`live_compute.cpp:1694`, `:1699`); all four enumerators are handled in the switch (`:1708-1716`), so no `-Wswitch` and no fallthrough. The switch adds no early return, no allocation, no lock. The seven untagged call sites keep their exact prior behaviour because the new parameter is appended last and every existing call is positional. The report block (`:10486-10509`) is inside `timing_mode.log` **and** `totals.calls % 25 == 0`, so it costs nothing per dispatch. The struct grows 48 bytes on a heap-allocated singleton.

**Thread safety: consistent, and consistency is the right standard here.** The six new counters are plain `uint64_t` next to `image_source_snapshot_copies`/`_bytes`, `storage_result_snapshot_*` and `image_result_snapshot_*`, which are also plain (`:1146-1155`). The `g_*` counters that *are* atomic are file-scope globals reachable from tests; these are members of the singleton whose contract is stated at `:10283` — "AGC submit execution serializes subsequent use". Using an atomic here would have been the inconsistent choice. The only cross-thread readers of the context are the accessors at `:10109`/`:10113`, which already read sibling plain counters unsynchronised.

---

### 1. The attribution — correct at the two small sites, and a **proxy** at the big one

`HostData` (`:2396-2400`) and `ContentChanged` (`:2032-2036`) sit where the labels say.

The `ReadModifyWrite` tag at `:2417-2421` is the load-bearing one and it is **not** a direct test for read/modify/write. It is guarded by `source_snapshot_required`, and there is exactly one caller that can supply a non-null source: `:9527`, which passes `!bi.seed_skip`. Tracing `seed_skip` (`:6165-6187`), it is set only for a storage binding that is `!readable && writable`, passes `dispatch_has_enough_threads_for_texels`, and has a **cached `Full` coverage proof**. So `!seed_skip` — the `rmw` bucket — also catches:

- coverage **proving and re-proving** frames (`bi.poison_verify`, `:6184-6185`), i.e. write-only full-coverage shaders whose bytes are *not* observable by the next dispatch;
- write-only full-coverage dispatches whose proof is not yet cached (`!known`, `:6195`);
- dispatches that fail the thread-count necessary condition (`:6160`).

I sized the first one, since it is the only recurring member: `kSeedReproveInterval` defaults to **256** fast-skips (`:5066`), so proving frames are bounded at roughly 0.4% of the skip-eligible population. The second is cold-start-only. So the *number* is not materially inflated and the PR's conclusion stands — but the *claim* is one step stronger than the instrument. Please retitle the bucket's meaning to what it measures: **"89% required a guest seed/baseline (`!seed_skip`)"**, of which read/modify/write is the dominant but not the sole member. This matters because the next lane will read "89% is RMW" as "89% is genuinely readable by the shader" and design around it.

The split is measurable rather than assumable, which is why I am asking for it rather than just flagging the wording: `record_storage_cache_gate` already counts `storage_seed_skip` and `storage_poison_verify` per role (`:3713-3714`), so `evaluated − seed_skip − poison_verify` bounds the honest RMW population from an instrument that already exists.

One fragility worth a line of comment: `validate_cached_image_source`'s `source_snapshot_required` **defaults to `true`** (`:2383`). Today the only other caller (`:8954`) passes `current_source = nullptr` and so counts nothing, but any future caller that omits the flag silently lands in the `rmw` bucket. Defaulting the *tag* rather than relying on the flag, or a comment at the enum saying "rmw == `!seed_skip`", would harden it.

Smaller: `ContentChanged` at `:2034` is guarded by `!upload_skipped`, and `upload_skipped = submit_unchanged || watch_unchanged || exact_unchanged` (`:2024`) — every disjunct requires `cached.content_valid`. So an entry whose authority was dropped by `invalidate_cached_image_source` (`:2347`) takes the same branch with *unchanged* bytes. "Content changed **or** no authority to compare against" is the exact meaning. Immaterial to the headline.

---

### 2. `host_data=0` — real, but the body states the wrong mechanism

The observation is right and the conclusion is right. The explanation in the body is wrong, and it is wrong in the direction this repo has been bitten by before, so I want it fixed rather than left as a cited-looking fact.

> the site is `#if defined(_WIN32)`-adjacent in effect, and a Linux optimisation aimed there would have been worth nothing

`:2396` is not platform-conditional in any way. `ShaderResource::host_data` is documented as **"Replay-only owned backing … Production tables leave both fields zero and use guest memory"** (`prosper/src/gpu/resources/shader_resources.hpp:450-453`), and every writer of an *image* resource's `host_data` is in `src/gpu/capture/gpu_capture.cpp` or `tools/gpu_replay/resource_override.hpp`. This bucket is zero in **any live run on any platform, Windows included** — a live-vs-replay fact, not a Linux-vs-Windows one. The conclusion survives and is in fact *stronger* than the body claims. There *is* a `#if defined(_WIN32)` block nearby (`:2100-2105`) but it is a **different, untagged** call site, which is probably where the conflation came from.

On your second question — whether zero means "never ran" or "ran with a null source": the counter genuinely cannot tell those apart, and the body asserts the former. But this does **not** undermine the partition, for a reason worth stating explicitly. `remember_image_source_snapshot` returns at `if (!source) return;` (`:1700`) *before* touching **any** counter, including `image_source_snapshot_bytes`. So a null source is absent from the numerator **and** the denominator; the partition stays self-consistent, and a zero here means "zero bytes were copied at this site", which is all the actionable conclusion needs. The evidence for the stronger structural claim is the `host_data` provenance above, not the counter — cite that instead.

---

### 3. The remainder — it closes, but "three branches" is wrong

I recomputed from the printed line: 71502.4 + 8212.5 + 835.8 = 80550.7 MiB, giving 88.77% / 10.20% / 1.04%. Consistent.

There are **ten** call sites of `remember_image_source_snapshot`, not three. Untagged and live on Linux: `retain_image` (`:2257`), `replace_or_retain_image` (`:2314`), and the two late paths in `validate_cached_image_source` (`:2444`, `:2448`). Plus a **direct** counter bump at `:2229-2236` that bypasses the function entirely — the `retain_image` move overload. Two more are inert in this configuration: `:2409` needs `PROSPER_NO_ADAPTIVE_STORAGE_RESULT_VALIDATION`, and `:2104` is `#if defined(_WIN32)`.

So five live contributors, all cold or one-shot retentions rather than per-dispatch revalidations, against a numerator dominated by per-dispatch traffic. 1% is exactly what that shape predicts, so yes — the remainder closes and the reading is sound. Only the prose count needs fixing.

One footnote for whoever ties this counter to the 48%-of-cycles `memmove`: the `:2229` site adds bytes via `std::move`, with **no copy at that point**, so `image_source_snapshot_bytes` is "bytes of exact guest mirror retained", not "bytes memmove'd here". Inside the 1%, so immaterial now; it would stop being immaterial if that path ever grew.

**`other` cannot go negative as the code stands** — every reason increment is paired with an equal `image_source_snapshot_bytes` increment in the same call (`:1702-1714`) and the reasons are mutually exclusive. The cited precedent is real (`tests/fixtures/render_runner.h:850`, `:872-877`) but *there* over-attribution is reachable through nested and re-entered timers. Keep it signed — it earns its keep the moment a fourth tag or a double-counting site appears — just don't read `+835.8` as the guard having fired.

---

### 6. The numbers in the body

The body is unusually honest about its own history and I don't want to flatten that. Two things still need touching:

- **"flips/s unchanged (1.19 → 1.05)"** is an 11.8% *drop*, described as unchanged in the same sentence that calls a 3% volume move "moved". The parenthetical prints both numbers, which is why this is wording and not correctness — but as written it reads as flat. Say "flips/s did not improve (1.19 → 1.05, single run)".
- **Single-run figures are stated as characterisations.** 48%/6% of cycles, 98.7 ms `gpu-device`, ~1.2 flips/s, 0.0715 → 0.1404, and the 89/10/1 split are each one run; the title states 89% flatly. The body qualifies the *method* (per pool-hit normalisation, which is the right call and well explained) but never the *variance*. One "single routed run" up front covers all of it.
- Minor: **"an always-changing 9.5 MiB source"** is not derivable from anything printed — 71502.4/6572 gives a 10.88 MiB mean. Say where 9.5 came from, or use the mean.

---

### 9. Forward-looking paragraph — defensible, but it names the wrong knob

The closing suggestion is correctly fenced off as not-in-this-PR, so this is only so the next lane doesn't lose a day. The 89% branch (`:2415-2423`) **never creates** a watch — it only rearms an existing one. Promotion for these sources is decided later, in `acquire_cached_image` (`:2013-2018` → `may_promote_write_watch_before_exact`, `:1682`). And the reason a large storage target can never reach the stability threshold is not the threshold: `validate_cached_image_source` sets `write_watch_stable_validations = 0` on **every** writeback (`:2394`), so a target written every dispatch resets its own counter before it can ever accumulate. That reset — not `should_promote_write_watch`'s `defer_min_bytes`/`promotion_validations` — is the knob for this population. Your instinct that the cost model is inverted here looks right; the lever is one line further up.

---

CI at review time: Docs, Linux App, Linux App Package, macOS App and Contribution shape green; Linux, Windows MinGW, Windows App and macOS (Rosetta) still pending. Merge on green as usual.

Issue #3033, opened 2026-08-26
Recover the still-valuable parts of PR #2984 (Windows pacing, audio instrumentation, compute SPIR-V tooling) and record what master has superseded


mattias800:
PR #2984 (`blasphemous2-windows`, +995/-122 across 15 files, 14 commits) carries a large amount of
useful work that never merged. This issue tracks recovering the parts still worth having, and records
what has since been superseded so nobody re-derives it.

## Status of #2984 itself

- Its blocking review was against the **wave-any relaxation** (`6fda6d04`), and the PR's own final
  commit `a3ac98d5` reverts that. The net diff is the *opposite* of what was rejected: a
  `fragment_wave_width_relaxable()` that returns `false` plus a test pinning the ruling.
- **CI is red from one line**: `winmm` is linked unconditionally
  (`target_link_libraries(prosper-app PRIVATE … winmm`), so Linux App, Linux App Package and macOS App
  all fail with `cannot find -lwinmm`. That single cause accounts for all three failures.
- **It no longer merges cleanly**: conflicts in `RECOMPILER_REMAINING.md`,
  `audio_sink_sdl3.cpp`, `main.cpp` and `render_runner.h`.

## Already superseded by master — do NOT re-harvest

- **The codified wave-any ruling.** Master's `RECOMPILER_REMAINING.md` row is strictly more complete
  than the branch's: it cites #2984 itself, names both defects, lists the nine `skipped_wave64_draw`
  arm line numbers, and names the route that *does* work (#2414 option 1 / #2429 — composing two
  32-wide votes). The branch version is a subset.
- **The render-side skip/reason machinery.** Master's `render_runner.h` has since gained
  `allow_native_fragment_vote_width` and a `subgroup_reasons` path, so the branch's `skip_reasons`
  bitmask rework is against code that has moved.

## Recovered

- [x] **Perf report tools** — `audio_delivery_report.py`, `flip_pacing_report.py` and their two
      self-tests (`23b38cad`). Touches only new files under `tools/perf/`, so it conflicts with
      nothing. **Also registered both as ctest cases**, which #2984 did not: they shipped with working
      self-tests that nothing ran, i.e. tests that could not go red.

## Still worth recovering, blocked on ordering

These touch files that open PRs #3019 / #3022 also touch, so they want to land after those merge:

- [ ] `c45debbf` — precise guest **semaphore** timedwait. #3022 deliberately left `k_sem_timedwait`
[...]

PR #2984, opened 2026-08-24
fix(win): precise guest semaphore timedwait + audio queue-depth pacing (Blasphemous 2 underruns); wave-any relaxation withdrawn per review

mattias800:
Supersedes the wave-any relaxation this branch previously carried — **withdrawn** per the review: the predicate is unsound (the vote result is consumed as guest scalar data via SCC), it was the third attempt at an approach #2404/#2410/#2414 killed, and the falsification is now codified as `fragment_wave_width_relaxable` (always false, unit-tested) plus a Ruled out row in docs/RECOMPILER_REMAINING.md. The fail-visible skip is restored.

### 1. The guest semaphore timedwait no longer quantizes to the Win32 tick (c45debbf + the rework)

`k_sem_timedwait` delegated to winpthreads' `sem_timedwait`, whose timed waits resolve on the winpthreads master tick (~15.6 ms) REGARDLESS of timeBeginPeriod — measured: `std::condition_variable::wait_for(5 ms)` = 15.7 ms mean with the resolution raised to 1 ms. FMOD's mixer paced its 256-frame grains on a guest semaphore timedwait and delivered audio at 1/3 rate: one 256-frame grain per ~15.6 ms instead of per 5.33 ms.

The wait is now trywait + a precise sleep loop (2 ms slices via host::sleep_until_steady_ns, deadline-checked): a post wakes the mixer within ~2 ms.

Measured (audio_delivery_report.py, port 17): the device queue at grain arrival holds **~6 grains (~63 ms)**; arrivals below one grain: **1 of 11,035 (0.0%)** — was 99.8% with the queue empty.

### 2. The SDL sink: queue-depth pacing, per-port mutexes, fill-then-start (0798b327, 6621aa51, e389440a)

- Queue-depth pacing: the sink holds ~8 grains and blocks the guest above it — the device drains in real time instead of the guest bursting.
- Per-port mutexes: the cushion wait held the sink's GLOBAL mutex, serializing port 17 against port 18 (Blasphemous 2 plays through two ports) — micro-stutters on every other source.
- Fill-then-start: the stream is created unbound and the device attaches at ~8 grains (~130 ms) — the cushion builds without draining.

### 3. Diagnostics (b2b90280, 2ca57b9c, a427e6f4, e389440a)

[...]

Issue #1765, opened 2026-08-02
Windows: sceVideoOutWaitVblank can quantize to ~31 ms (≈32 Hz pacing), and test_videoout is not built on Windows to catch it


mattias800:
Found by an independent review of #1757. Two separate facts that compound, both verified by
reading the build and the CI logs rather than by measurement — I have **not** run this on Windows,
and the issue says so throughout.

## 1. The pacing hazard

`sceVideoOutWaitVblank` (#1757) blocks to the next 59.94 Hz boundary with
`std::this_thread::sleep_until`. On MinGW, libstdc++'s `sleep_for` can fall back to `::Sleep(ms)`,
whose resolution is the process's current timer period. Windows' default is **~15.6 ms** unless
something calls `timeBeginPeriod`.

`grep -rn timeBeginPeriod prosper/src prosper/frontends` returns **nothing** outside `third_party/`
— only vendored SDL3 raises it, and only for its own windowed frontend, so `boot_trace`,
`screenshot` and any headless consumer run at the default period.

A 16.68 ms target then rounds up to two ticks ≈ **31.2 ms**, pacing a title's display thread at
**~32 Hz** instead of ~60. The standing project goal is ≥30 fps with 60 the target, so a Windows
path that silently halves display pacing is worth its own issue even unmeasured.

Note this is strictly a *Windows* concern. On Linux/glibc, `sleep_until` on a steady clock delegates
to a `nanosleep` loop that also handles EINTR correctly (`this_thread_sleep.h`), which matters here
because prosper delivers its own signals — the reason `hle_kernel_time.cpp` has `sleep_full` at all.

## 2. Nothing on Windows would catch it

`prosper/CMakeLists.txt:1014-1016` defines `test_videoout` / `videoout_queries` inside the
`if(UNIX)` block at `:942`. **It is not built or run on Windows**, which the MinGW CI job confirms:
that job prints a per-test line for all its tests and `videoout_queries` is not among them.

And even where it does run, the assertion added in #1757 —

```cpp
CHECK(elapsed_us >= 4 * 16683, "five WaitVblank calls block for at least four vblank periods");
```

— is a **lower bound**. Oversleeping by a factor of two passes green. So the guard could not detect
this failure mode even if it were built on Windows.

## Suggested work

1. **Measure it first.** Run `test_videoout` manually under MinGW (or a five-line
   `sleep_until`-in-a-loop probe) and record the real quantization. It may already be fine — recent
[...]

PR #3064, opened 2026-08-27
fix(win): the vblank kevent pump joins the vblank grid instead of keeping its own 33 Hz clock

### LINKED ISSUE #3024: Windows: the vblank kevent pump ticks at 33 Hz instead of 60 — raw nanosleep in vblank_pump, the same defect #1765 fixed in the other vblank clock
`hle_kernel_time.cpp:1049` — the thread that posts the VideoOut vblank kevent to every registered
equeue:

```cpp
void vblank_pump() {
    for (;;) {
        struct timespec ts{ 0, 16666667 }; nanosleep(&ts, nullptr);   // ~60 Hz
```

That is a **host** `nanosleep`, not the guest HLE one. On Windows it resolves on the winpthreads tick,
so the comment's "~60 Hz" is not what runs. Measured by hand outside prosper on this host
(MinGW-w64/UCRT, GCC 16.1):

| | per tick | effective rate |
| --- | --- | --- |
| as the pump runs it | **29.95 ms** | **33.4 Hz** |
| with `timeBeginPeriod(1)` | 24.78 ms | 40.4 Hz |

So every title that paces on the vblank **event** stream is being ticked at roughly **half rate** on
Windows.

> **Correction (2026-08-27).** This paragraph originally read "and `prosper-app`'s `timeBeginPeriod(1)` (#2985) does not rescue it". **prosper calls `timeBeginPeriod` nowhere** — `precise_sleep.cpp:14` says so explicitly ("`timeBeginPeriod` / `NtSetTimerResolution` appear nowhere outside `third_party/`"), and a `grep` over `src/` and `frontends/` confirms it on current master. The claim was residue of an earlier error of mine, where I read an uncommitted local diff and later asserted it as master's behaviour. So the second table row is a **hypothetical** arm of the standalone microbenchmark, not a description of what prosper does; the unrescued **29.95 ms** row is what actually ran. The defect is therefore slightly worse than this issue originally described, not better.

## This is a known defect fixed in one place and left in another

`hle_graphics.cpp`'s `vblank_sleep_until_ns` carries the fix and says why:

> NOT `std::this_thread::sleep_until`: on Windows that is one `::Sleep(ms)`, which the system timer
> period quantizes to ~15.6 ms steps, so this ~16.68 ms wait cost ~31 ms and paced a title's display
> thread at ~32 Hz (#1765).

#1765 corrected the `sceVideoOutGetVblankStatus` / `sceVideoOutWaitVblank` grid. The kevent pump is a
second vblank clock in a different file and never got the same treatment — and `hle_graphics.cpp`
already documents that the two exist and disagree:

> It is NOT the only vblank clock in the process... `vblank_pump` (:731) posts the VideoOut vblank
> kevent from its own `nanosleep(16666667)` loop — 60.000 Hz, its own phase... So the kevent stream and
> this grid drift by roughly one tick per second.

That note understates it on Windows: the disagreement is not one tick per second of phase drift, it is
**33 Hz against 59.94 Hz in period**.

## Two defects, not one

1. **Quantization** — the tick is 30 ms, not 16.67 ms.
2. **Drift** — the loop sleeps a *relative* interval per iteration, so the pump's own scheduling
   overhead accumulates instead of being absorbed. Even with an accurate sleep it would slowly lag a
   real 60 Hz grid.

## Fix

Sleep to an **absolute** deadline on a fixed grid using `host::sleep_until_steady_ns` — the same
primitive `vblank_sleep_until_ns` already uses one file away. That addresses both: the grid cannot
drift, and the primitive does not inherit the tick.

Ideally the pump and `hle_graphics.cpp`'s grid should derive from **one** period constant so they cannot
disagree in period at all, which is the follow-up that file already asks for.

## Expected impact

Raising the vblank event rate from ~33 Hz to ~60 Hz changes how often titles that wait on this kevent
advance. That is a correctness fix, but it is a behavioural change on Windows for anything pacing on it,
so it wants a spot-check across a few titles rather than being assumed harmless.

mattias800:
Fixes #3024. Recovery item 4 of the #3033 salvage of #2984 — this one is not a port; #2984 never touched the kevent pump.

## The defect

`vblank_pump` (`hle_kernel_time.cpp`) posted the VideoOut vblank kevent to every registered equeue from a bare `nanosleep(16666667)`. On Windows that resolves on the winpthreads tick, so its own `// ~60 Hz` comment was not what ran, and every title pacing on the kevent stream advanced at roughly half rate.

This is the **same defect #1765 already fixed one file away** for the `sceVideoOutGetVblankStatus` / `sceVideoOutWaitVblank` grid — which is why `hle_graphics.cpp` carries `NOT std::this_thread::sleep_until` over its own wait. The kevent pump is a second vblank clock in a different file and never got the same treatment. `hle_graphics.cpp` even documents that the two exist and disagree, and asks for exactly this follow-up in as many words:

> the kevent stream and this grid drift by roughly one tick per second … aligning them is a real follow-up; asserting they are aligned is wrong

## Two defects, not one

1. **Quantization** — the sleep returns every ~30 ms, not 16.67.
2. **Drift** — a *relative* sleep per iteration accumulates the pump's own overhead, so it lagged the status grid even where the sleep was accurate.

## The fix

The pump waits `host::sleep_until_steady_ns` against an **absolute** deadline on the **same origin and period** as the status/wait grid, newly exported as `prosper_vo_vblank_grid_origin_ns()` / `prosper_vo_vblank_period_ns()`. One clock, in phase and in period, by construction rather than by two constants that happen to agree — and the old pump's constant did **not** agree: 60.000 Hz against the grid's 59.94 Hz.

Both are `std::chrono::steady_clock` nanoseconds, checked rather than assumed (`vblank_now_ns` → `steady_clock`, and `sleep_until_steady_ns`'s contract is the same epoch).

## The part worth the review time: my first version of the arithmetic was wrong, and read as correct

The obvious inline spelling tests "have we fallen behind?" against the **pre-advance** deadline:

```cpp
if (next_ns < now_ns) next_ns = now_ns + period_ns;   // looks like a stall guard
next_ns += period_ns;
```

`sleep_until` returns a little **past** its deadline by nature, so that condition is true on essentially every ordinary iteration — the re-anchor fires every tick and costs one extra period each time. It would have shipped a **30 Hz** pump under a comment claiming 60: *a lower rate than the 33 Hz defect it was fixing*, with no symptom visible anywhere in the diff.

So the advance is now `host::next_periodic_deadline_ns(deadline, now, period)` — a `constexpr` free function beside the deadline sleeper it belongs with, with its own arms. The arms include an **iterated rate assertion** (60 ticks of ordinary overshoot must cover exactly 60 periods, not 120), because a single-step arm is the weaker test and the defect is a *rate* defect.

Mutation-checked, all three redden:

| mutation | arms reddened |
| --- | --- |
| my original buggy spelling | 4 (incl. the iterated rate arm) |
| no resync (catch-up burst) | 2 |
| resync unconditionally | 4 |

A wholly missed tick re-anchors and posts **one** event; it deliberately does not replay the missed slots, because consumers pace on events *arriving* and a burst reads to them as several periods of progress in one instant.

## Provenance of the 33 Hz figure, stated because I have got this wrong before

The `29.95 ms` / `33 Hz` numbers come from a **standalone microbenchmark of that nanosleep** on this toolchain (MinGW-w64/UCRT, GCC 16.1) — not from an instrumented measurement of the pump. The code comment says so too, and quotes the figure only to size the defect.

The same microbenchmark under `timeBeginPeriod(1)` gives 24.78 ms — still not 16.67 — and that arm is **hypothetical for prosper, which calls `timeBeginPeriod` nowhere**; `precise_sleep.cpp:14` states this explicitly. So the unrescued 29.95 ms is what ran. #3024's body claims otherwise in one sentence ("prosper-app's `timeBeginPeriod(1)`") — that is residue of an earlier false claim of mine, and I have corrected the issue body.

## Behavioural change, named rather than buried

This roughly doubles the vblank **event** rate on Windows for anything waiting on the kevent. A title whose logic advances per vblank event will now advance about twice as fast on Windows as it did — i.e. at the rate it already had on Linux. That is the point of the fix, but it is a real behavioural change on Windows and a reviewer should weigh it as one.

**Correction — "Linux and macOS are untouched" was false**, and review was right to block on it. Two changes land on every platform:

1. **The pump's period changes**, 60.000 Hz → 59.94 Hz. It now uses the grid's constant instead of its own hardcoded `16666667`.
2. **The grid's epoch is normally anchored EARLIER.** `vblank_epoch_ns()` is a first-use-anchored `static`, and the pump's read at `hle_kernel_time.cpp` reaches it as soon as a title registers a vblank event — typically *before* the title's first `GetVblankStatus` or `WaitVblank`. So the epoch now tracks vblank-event registration rather than the first guest status query. That moves `sceVideoOutGetVblankStatus`'s first `count` from 1 to `1 + elapsed/P`, shifts `processTime`'s baseline, and shifts **`sceVideoOutWaitVblank`'s wake instants in phase** along with them, since all three derive from the same `t0`. A guest display thread pacing on `WaitVblank` is the likeliest consumer to notice.

I judge (2) more faithful — the epoch now marks when the title started caring about vblanks rather than when it first asked — and low risk, since the grid stays monotonic and self-consistent either way. But it is a real behavioural change on all three platforms and it belongs in this list rather than in a sentence claiming the opposite. The code comment at `hle_graphics.cpp` now says the same thing; it previously asserted that callers *cannot* move the grid, which is true of every caller except the first, and the first is now the pump.

What genuinely does not change off Windows: `sleep_until_steady_ns` is `std::this_thread::sleep_until` there, so no wait primitive changes, and no grid constant changed.

## Deliberately not in scope

- The pump still posts to `g_vblank_regs` under `g_eq_mx` exactly as before; no change to what is posted or to whom.
- `hle_graphics.cpp`'s grid is unchanged — it was already correct. This PR only makes the pump *derive from* it.
- The remaining `#3033` items (the queue-depth audio pacer A/B, the dedicated compute queue) are separate.

## Verification

Windows, MinGW-w64/UCRT GCC 16.1, RelWithDebInfo, RTX 4090:

```
cmake --build prosper/build-mingw-app          # clean
100% tests passed, 0 tests failed out of 247   # exit 0
```

One pre-existing unrelated skip (`hle_calls_value_capture`). `git diff --check` clean.

No screenshot: this changes an event rate, not a rendered frame, and a capture of the same title at a different pump rate would be evidence of nothing. Per the charter's blog rule this is a finding-without-a-picture, so it goes in `BLOG.md` only if the pacing change turns out to move a title — noted on #3033 rather than claimed here.

Refs #3033. Refs #2984.

### DIFF
--- a/prosper/src/hle/graphics/hle_graphics.cpp
+++ b/prosper/src/hle/graphics/hle_graphics.cpp
@@ -743,9 +743,10 @@ HLE(g_vo_resstatus)   {
 //
 // But the FIRST caller does set where it starts, and after #3024 that caller is usually the
 // kevent pump. `vblank_epoch_ns()` is a first-use-anchored static, and the pump's read at
-// hle_kernel_time.cpp reaches it as soon as a title registers a vblank event -- typically
-// before the title's first GetVblankStatus or WaitVblank call. So the epoch is now normally
-// anchored at vblank-event registration rather than at the first guest status query, which
+// hle_kernel_time.cpp reaches it as soon as a title registers a vblank OR A FLIP event (both
+// call the same ensure_pump) -- typically before the title's first GetVblankStatus or
+// WaitVblank call. So the epoch is now normally anchored at event registration rather than at
+// the first guest status query, which
 // moves GetVblankStatus's first `count` and processTime's baseline, and shifts WaitVblank's
 // wake instants in PHASE with them (they derive from the same t0). That is a deliberate
 // consequence on every platform, not only Windows, and it is more faithful -- the epoch now
--- a/prosper/src/hle/graphics/hle_graphics.cpp
+++ b/prosper/src/hle/graphics/hle_graphics.cpp
@@ -764,6 +764,14 @@ void vblank_sleep_until_ns(uint64_t deadline_ns) {
 }
 }  // namespace
 
+// The kernel equeue vblank pump (hle_kernel_time.cpp) schedules on THIS grid, so the kevent
+// stream and the vblank status/wait paths above become one clock in both phase and period.
+// They were two drifting clocks before -- the follow-up this file records at its own timebase
+// comment ("aligning them is a real follow-up; asserting they are aligned is wrong"). Sharing
+// the origin and the period here is what closes it (#3024).
+extern "C" uint64_t prosper_vo_vblank_grid_origin_ns() { return vblank_epoch_ns(); }
+extern "C" uint64_t prosper_vo_vblank_period_ns() { return kVblankNs; }
+
 extern "C" void prosper_vo_set_vblank_now_for_test(uint64_t now_ns) {
     g_vblank_test_now_ns.store(now_ns, std::memory_order_relaxed);
 }
--- a/prosper/src/hle/graphics/hle_graphics.cpp
+++ b/prosper/src/hle/graphics/hle_graphics.cpp
@@ -729,14 +729,17 @@ HLE(g_vo_resstatus)   {
 // which is a statement about where the boundaries fall, not about how far apart they are. Anchored
 // on first use so both are process-relative, exactly as the previous status-only anchor was.
 //
-// It is NOT the only vblank clock in the process, and this comment previously claimed it was.
-// hle_kernel_time.cpp's `vblank_pump` (:731) posts the VideoOut vblank kevent from its own
-// `nanosleep(16666667)` loop — 60.000 Hz, its own phase, started when the first equeue source is
-// registered. So the kevent stream and this grid drift by roughly one tick per second and their
-// boundaries do not coincide. Aligning them is a real follow-up; asserting they are aligned is
-// wrong, and a title that cross-checks the two would see the disagreement.
+// It is no longer a second clock. hle_kernel_time.cpp's `vblank_pump` -- the thread that posts the
+// VideoOut vblank kevent to registered equeues -- used to run its own `nanosleep(16666667)` loop
+// at 60.000 Hz and its own phase, so the kevent stream and this grid disagreed in BOTH period and
+// phase, and this comment recorded aligning them as an open follow-up. #3024 closed it: the pump
+// now derives its schedule from the two accessors published just below, so there is one origin and
+// one period in the process. The alignment is therefore asserted here on purpose -- it is what the
+// pump computes from, not a hope about two constants matching.
 // Internal linkage on purpose: the "epoch before now" ordering invariant below is only
-// enforceable by reading this one file, so nothing outside it may reach these.
+// enforceable by reading this one file, so nothing outside it may reach these DIRECTLY. The
+// two `extern "C"` accessors further down publish the origin and the period as values, which
+// cannot break that invariant -- a caller can schedule on the grid but cannot move it.
 namespace {
 constexpr uint64_t kVblankNs = 16683350;             // 59.94 Hz period
 // Deterministic clock used only by test_videoout's phase-integration check. Zero leaves the
--- a/prosper/src/hle/graphics/hle_graphics.cpp
+++ b/prosper/src/hle/graphics/hle_graphics.cpp
@@ -738,8 +738,19 @@ HLE(g_vo_resstatus)   {
 // pump computes from, not a hope about two constants matching.
 // Internal linkage on purpose: the "epoch before now" ordering invariant below is only
 // enforceable by reading this one file, so nothing outside it may reach these DIRECTLY. The
-// two `extern "C"` accessors further down publish the origin and the period as values, which
-// cannot break that invariant -- a caller can schedule on the grid but cannot move it.
+// two `extern "C"` accessors further down publish the origin and the period as VALUES, so no
+// caller can retarget the grid.
+//
+// But the FIRST caller does set where it starts, and after #3024 that caller is usually the
+// kevent pump. `vblank_epoch_ns()` is a first-use-anchored static, and the pump's read at
+// hle_kernel_time.cpp reaches it as soon as a title registers a vblank event -- typically
+// before the title's first GetVblankStatus or WaitVblank call. So the epoch is now normally
+// anchored at vblank-event registration rather than at the first guest status query, which
+// moves GetVblankStatus's first `count` and processTime's baseline, and shifts WaitVblank's
+// wake instants in PHASE with them (they derive from the same t0). That is a deliberate
+// consequence on every platform, not only Windows, and it is more faithful -- the epoch now
+// tracks when the title started caring about vblanks rather than when it first asked -- but
+// it is a change, and a comment claiming callers cannot move the grid would have hidden it.
 namespace {
 constexpr uint64_t kVblankNs = 16683350;             // 59.94 Hz period
 // Deterministic clock used only by test_videoout's phase-integration check. Zero leaves the
--- a/prosper/src/hle/kernel/hle_kernel_time.cpp
+++ b/prosper/src/hle/kernel/hle_kernel_time.cpp
@@ -7,6 +7,8 @@
 #include "host/platform/posix_shim.hpp"     // PROSPER_ASM_TRAMPOLINE (pass entry %rsp as 7th arg)
 #include "host/platform/precise_sleep.hpp"   // guest sleeps must not inherit the winpthreads tick (#3013)
 #include "hle/kernel/timedwait_census.hpp"   // PROSPER_TIMEDWAIT_CENSUS: which primitive a title paces on
+extern "C" uint64_t prosper_vo_vblank_grid_origin_ns();   // the shared 59.94 Hz grid (hle_graphics.cpp)
+extern "C" uint64_t prosper_vo_vblank_period_ns();
 #include "host/image/runtime_module_load.hpp"   // #639: real runtime PRX loading
 #include "host/memory/guest_write_watch.hpp"     // flush dmem writer diagnostic before guest _Exit
 #include "hle/dispatch/callback_fs.hpp"            // recover the caller's guest %fs from the import-stub frame
@@ -1108,8 +1110,39 @@ namespace {
     }
     void vblank_pump() {
         uint64_t frame = 0;
+        // Absolute schedule on the SHARED 59.94 Hz grid that hle_graphics.cpp uses for the
+        // vblank status and wait paths, waited with host::sleep_until_steady_ns. Two defects
+        // this replaces (#3024):
+        //
+        //   * A relative nanosleep(16.67 ms) resolves on the winpthreads tick on Windows. A
+        //     standalone microbenchmark of that nanosleep on this toolchain (MinGW-w64/UCRT,
+        //     GCC 16.1) returns every 29.95 ms, so the pump was posting vblank kevents at
+        //     roughly 33 Hz rather than 60, and every consumer pacing on this event stream
+        //     inherited it. Provenance stated because it matters: that is the sleep primitive
+        //     measured in isolation, NOT an instrumented measurement of this pump, and the
+        //     figure is quoted here only to size the defect. Same microbenchmark under
+        //     timeBeginPeriod(1) gives 24.78 ms -- still not 16.67 -- and that arm is
+        //     hypothetical for prosper, which calls timeBeginPeriod nowhere (see
+        //     precise_sleep.cpp's own note). So the unrescued 29.95 ms is what ran.
+        //     It is the same defect #1765 fixed for the status/wait grid one file away, which
+        //     is why that file already says "NOT std::this_thread::sleep_until" over its wait.
+        //   * A relative sleep also drifts against the status grid. hle_graphics.cpp records
+        //     that as an open follow-up in as many words -- the two clocks "drift by roughly one
+        //     tick per second and their boundaries do not coincide", and a title that
+        //     cross-checks the kevent count against the status count would see it. Deriving both
+        //     from one origin and one period closes it by construction rather than by tuning.
+        //
+        // host::next_periodic_deadline_ns carries the advance, including the resync after a
+        // wholly missed tick. It is a separate tested function because the obvious inline
+        // spelling of that resync is wrong in a way that still looks like 60 Hz -- see its
+        // comment in precise_sleep.hpp.
+        const uint64_t period_ns = prosper_vo_vblank_period_ns();
+        uint64_t next_ns = prosper_vo_vblank_grid_origin_ns() + period_ns;
         for (;;) {
-            struct timespec ts{ 0, 16666667 }; nanosleep(&ts, nullptr);   // ~60 Hz
+            prosper::host::sleep_until_steady_ns(next_ns);
+            const uint64_t now_ns = (uint64_t)std::chrono::duration_cast<std::chrono::nanoseconds>(
+                clk::now().time_since_epoch()).count();
+            next_ns = prosper::host::next_periodic_deadline_ns(next_ns, now_ns, period_ns);
             frame++;
             std::vector<FlipReg> vr;
             { std::lock_guard lk(g_eq_mx); vr = g_vblank_regs; }
--- a/prosper/src/hle/kernel/hle_kernel_time.cpp
+++ b/prosper/src/hle/kernel/hle_kernel_time.cpp
@@ -1142,6 +1142,15 @@ namespace {
         // the obvious hand-rolled spelling -- advance by a period, re-anchor to now+period when
         // behind -- silently loses the phase, and an earlier draft of this loop also halved the
         // rate. See its comment in precise_sleep.hpp.
+        //
+        // One composition caveat, recorded because the helper alone does not cover it (#3074):
+        // this loop assumes the wait cannot return EARLY. If it does, the next `now` is still
+        // below the same boundary, the helper correctly returns that boundary again, and this
+        // thread posts two kevents for one vblank. Reachable only on the Win32SleepFallback
+        // path, which does not re-check the clock the way the high-resolution timer path does;
+        // bounded (one extra event, and vblank posts coalesce) so it is filed rather than
+        // worked around here. The fix belongs in precise_sleep, which already documents the
+        // post-condition it fails to enforce on that one path.
         const uint64_t origin_ns = prosper_vo_vblank_grid_origin_ns();
         const uint64_t period_ns = prosper_vo_vblank_period_ns();
         for (;;) {
--- a/prosper/src/hle/kernel/hle_kernel_time.cpp
+++ b/prosper/src/hle/kernel/hle_kernel_time.cpp
@@ -1132,9 +1132,18 @@ namespace {
         //     is why that file already says "NOT std::this_thread::sleep_until" over its wait.
         //   * A relative sleep also drifts against the status grid. hle_graphics.cpp records
         //     that as an open follow-up in as many words -- the two clocks "drift by roughly one
-        //     tick per second and their boundaries do not coincide", and a title that
-        //     cross-checks the kevent count against the status count would see it. Deriving both
-        //     from one origin and one period closes it by construction rather than by tuning.
+        //     tick per second and their boundaries do not coincide". Deriving both from one
+        //     origin and one period fixes the BOUNDARIES: a kevent now lands on an instant the
+        //     status grid also calls a boundary.
+        //
+        //     It does NOT make the two COUNTS equal, and the earlier wording here claimed it
+        //     did. `e.data` below is `frame`, which counts posts; GetVblankStatus's `count` is
+        //     the grid INDEX. They agree only while no boundary is skipped -- and skipping is
+        //     what this design does deliberately when a tick is missed, rather than bursting.
+        //     A title cross-checking the two would still see a divergence after any stall.
+        //     Making the payload the grid ordinal would close that too, but `data = frame`
+        //     predates this change and nothing here establishes what real hardware puts in
+        //     that field, so it is left alone and named instead of quietly overclaimed.
         //
         // One call carries the whole schedule: host::next_grid_deadline_ns returns the next
         // boundary strictly after now, so the ordinary case and a stall of any length are the
@@ -1142,6 +1151,12 @@ namespace {
         // the obvious hand-rolled spelling -- advance by a period, re-anchor to now+period when
         // behind -- silently loses the phase, and an earlier draft of this loop also halved the
         // rate. See its comment in precise_sleep.hpp.
+        //
+        // The cost of that property, named because it is real: phase stays exact however
+        // many boundaries were skipped, so a pump running at HALF rate has perfect phase
+        // and no symptom. The relative sleep this replaces at least drifted visibly, which
+        // is how #3024 was noticed. A dropped-boundary counter would restore the symptom
+        // without giving back the drift -- #3075.
         //
         // One composition caveat, recorded because the helper alone does not cover it (#3074):
         // this loop assumes the wait cannot return EARLY. If it does, the next `now` is still
@@ -1151,6 +1166,13 @@ namespace {
         // bounded (one extra event, and vblank posts coalesce) so it is filed rather than
         // worked around here. The fix belongs in precise_sleep, which already documents the
         // post-condition it fails to enforce on that one path.
+        // Reading the origin here is what usually ANCHORS it -- see the note in
+        // hle_graphics.cpp. Two qualifications, because the obvious statement of that is not
+        // quite true: ensure_pump() detaches this thread, so a guest status or wait call
+        // arriving inside the thread-start window anchors the grid instead, which makes the
+        // epoch's wall-clock phase nondeterministic by thread-start latency; and the pump is
+        // started by prosper_eq_add_flip as well as prosper_eq_add_vblank, so a title that
+        // registers only a FLIP event still anchors it here.
         const uint64_t origin_ns = prosper_vo_vblank_grid_origin_ns();
         const uint64_t period_ns = prosper_vo_vblank_period_ns();
         for (;;) {
--- a/prosper/src/hle/kernel/hle_kernel_time.cpp
+++ b/prosper/src/hle/kernel/hle_kernel_time.cpp
@@ -7,8 +7,6 @@
 #include "host/platform/posix_shim.hpp"     // PROSPER_ASM_TRAMPOLINE (pass entry %rsp as 7th arg)
 #include "host/platform/precise_sleep.hpp"   // guest sleeps must not inherit the winpthreads tick (#3013)
 #include "hle/kernel/timedwait_census.hpp"   // PROSPER_TIMEDWAIT_CENSUS: which primitive a title paces on
-extern "C" uint64_t prosper_vo_vblank_grid_origin_ns();   // the shared 59.94 Hz grid (hle_graphics.cpp)
-extern "C" uint64_t prosper_vo_vblank_period_ns();
 #include "host/image/runtime_module_load.hpp"   // #639: real runtime PRX loading
 #include "host/memory/guest_write_watch.hpp"     // flush dmem writer diagnostic before guest _Exit
 #include "hle/dispatch/callback_fs.hpp"            // recover the caller's guest %fs from the import-stub frame
@@ -992,6 +990,12 @@ namespace { bool evlog() { static int v = getenv("PROSPER_EVLOG") ? 1 : 0; retur
 // --- Real event-queue backend (kqueue/kevent model). Registered flip/vblank sources post events into
 // their equeue; WaitEqueue blocks until an event is ready (or timeout) and returns it. A single ~60 Hz
 // pump thread drives vblank + flip-completion events so Unity's render/timing threads pace frames. ---
+// The shared 59.94 Hz vblank grid, defined in hle/graphics/hle_graphics.cpp. Declared here
+// rather than in a header because that file deliberately keeps the grid internal, and these
+// two accessors are the whole of what it publishes (#3024).
+extern "C" uint64_t prosper_vo_vblank_grid_origin_ns();
+extern "C" uint64_t prosper_vo_vblank_period_ns();
+
 namespace {
     // SceKernelEvent (FreeBSD kevent layout, 0x20 bytes): ident@0, filter@8(i16), flags@0xA(u16),
     // fflags@0xC(u32), data@0x10, udata@0x18. The game reads udata (its flip context) + data.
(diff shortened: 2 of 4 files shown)


mattias800 (commented):
Independent review of #3064 (reviewer: separate agent; I did not re-run the author's build or ctest).

**REJECTED** — three blocking findings, all small, and none of them is the deadline arithmetic. The
arithmetic is correct and the fix is right in substance; what is wrong is that two comments and the PR
body assert properties the code does not deliver, and one of those comments is in a file this PR edits
30 lines away from the paragraph it contradicts.

## What I verified and agree with

* **`next_periodic_deadline_ns` (precise_sleep.hpp:41-49) is correct.** Worked by hand: the resync test
  is against `advanced`, not `deadline_ns`, so ordinary overshoot (`now` in `[deadline, deadline+period)`)
  returns `deadline+period` — exactly one period, no halving, and overshoot is absorbed rather than
  accumulated. It can never return a value below `now_ns`: the taken branch is `now+period` and the
  untaken one has `advanced >= now` as its guard. `deadline+period` cannot realistically overflow a
  `uint64_t` of steady-clock nanoseconds. Your mutation table checks out — I re-derived all three:
  the original buggy spelling `(cond ? now+P : d) + P` yields `d+120000+2P` and reddens the single-step,
  iterated, overshoot-accumulation and early-return arms (4); `return advanced` reddens the two stall
  arms (2); `return now+period` reddens 4. The "wake exactly on the deadline" arm is the one that catches
  none of them, which is fine — it is a boundary arm, not a rate arm.
* **Clock domains agree, checked independently of your claim.** `vblank_now_ns` (hle_graphics.cpp:747-752)
  is `steady_clock::now().time_since_epoch()` cast to ns; `precise_sleep.cpp:63` `steady_now_ns()` is
  byte-for-byte the same expression, and `sleep_until_steady_ns` compares against it and reconstructs a
  `steady_clock::time_point` from the same ns count; the pump's own `now_ns` (hle_kernel_time.cpp:1143)
  uses `clk = std::chrono::steady_clock` (:55). One epoch, three call sites. A past deadline returns
  immediately on both platforms (`if (now >= deadline_ns) return;` on Windows, `sleep_until` semantics on
  POSIX), so the late-origin first iteration cannot hang.
* **Magic-static init is safe** — C++11 thread-safe statics, and `vblank_epoch_ns()` was already reached
  from guest threads on master; the pump is an ordinary host `std::thread`, so this adds no guest-`%fs`/TLS
  exposure.
* **Nothing internal depends on the old rate.** `g_vblank_regs` is written only by
  `prosper_eq_add_vblank` and read only by the pump; no test, tool, doc or frontend references
  `16666667` or the pump's rate. And doubling the rate cannot flood an equeue: vblank posts are
  coalescible and `eq_post` replaces in place for filter -13 (hle_kernel_time.cpp:1073-1090), so the
  extra events cost a `notify_all`, not queue growth. I agree the behavioural risk is confined to guest
  pacing, which is the point of the fix.
* **`timeBeginPeriod` claim: verified.** `git grep -i 'timeBeginPeriod\|NtSetTimerResolution'` over
  `prosper/src`, `frontends`, `tools`, `tests` on your branch returns nine hits, **all of them comments**
  — no call site, and none in `third_party/` either. `precise_sleep.cpp:14` says what you cite it as
  saying. The three quotations attributed to `hle_graphics.cpp` (in the PR body, in the new comment at
  hle_graphics.cpp:767-771, and in hle_kernel_time.cpp:1130-1131) are all accurate to the source text.

## Blocking

**B1. `hle_graphics.cpp:732-737` still describes master's pump in the present tense and says the
alignment is an open follow-up — this PR makes that paragraph false, and the PR edits that very file.**

> hle_kernel_time.cpp's `vblank_pump` (:731) posts the VideoOut vblank kevent from its own
> `nanosleep(16666667)` loop — 60.000 Hz, its own phase, started when the first equeue source is
> registered. So the kevent stream and this grid drift by roughly one tick per second and their
> boundaries do not coincide. Aligning them is a real follow-up; asserting they are aligned is wrong

After this PR there is no `nanosleep(16666667)`, no second phase, and no per-second drift. You quote this
paragraph in the PR body and again at hle_graphics.cpp:769-770 as a follow-up your change "closes", while
leaving the paragraph itself asserting that it is open — so the file now contains two comments that
contradict each other, and the stale one is the one a reader hits first (it sits above the grid
definition; the new one is 30 lines below, past the anonymous namespace). This is precisely the charter's
"a comment says something the code does not do" case, and it is a text edit. While you are there:
`(:731)` is a stale line reference — `vblank_pump` is at :1111.

**B2. "one clock … in phase" is not true by construction. Phase holds only if the pump starts within two
periods of the grid origin, and any resync destroys it permanently.**

hle_graphics.cpp:767-768 states the two "become one clock in both phase and period", and the PR body says
"in phase and in period, by construction". The code does not give that:

* **Start ordering.** `next_ns = prosper_vo_vblank_grid_origin_ns() + period_ns`
  (hle_kernel_time.cpp:1140). Let `X = now − origin` at pump start. For `X ≤ 2P` phase is exact (for
  `P < X ≤ 2P` you get one immediate post, then the grid). For `X > 2P` the first iteration takes the
  resync branch and lands on `now + P` — an **arbitrary** offset from the grid, held forever. `X > 2P`
  (33 ms) is reachable whenever anything touches the grid before the pump does, i.e. any title that calls
  `sceVideoOutGetVblankStatus`/`sceVideoOutWaitVblank` before it registers a vblank equeue source. Nothing
  in the diff bounds that ordering.
* **Steady state.** The same applies after any single stall of more than one period. On the platform this
  PR exists for, a display-paced wait overshooting by 14-30 ms is the *documented* behaviour of the
  fallback path (`Win32SleepFallback`, precise_sleep.cpp:124-127) — so on a host where the
  high-resolution timer cannot be created, the pump will resync repeatedly and be off-grid essentially
  always, silently.

Two acceptable resolutions, and I would argue for the first: snap to the grid instead of to `now` —
`origin + ((now − origin)/P + 1)*P`, which is exactly the idiom `g_vo_wait_vblank` already uses at
hle_graphics.cpp:840. It is always strictly in the future (no spin), still posts exactly one event after
a stall (no burst), and makes the phase claim unconditional in both cases above. It is also testable in
the same style as the existing arms. Otherwise: weaken the claim in both comments and the PR body to
"the same period, and the same phase when the pump anchors the grid", which is what ships today.

**B3. "Linux and macOS are untouched" is false in two guest-visible ways, on every platform.**

* The pump's period changes from `16666667` to `16683350` everywhere — 60.000 Hz → 59.94 Hz. Immaterial
  in size, but the body asserts no change at all while simultaneously (correctly) making the constant
  disagreement part of its argument.
* More substantively: the pump now **anchors** `vblank_epoch_ns()` (the export at hle_graphics.cpp:772
  calls it, and it is first-use-anchored). On master the grid origin was whichever guest
  `GetVblankStatus`/`WaitVblank` call came first; with this PR it is usually `sceVideoOutAddEvent` time,
  which is *earlier*. So `sceVideoOutGetVblankStatus`'s first `count` changes from 1 to `1 + elapsed/P`
  and `processTime`'s baseline moves with it — on Linux and macOS too. I think this is arguably *more*
  faithful (a console's vblank counter does not start when the game first asks) and low risk, since the
  Astro Bot `(frames*refresh)/(count − previous_count)` case at hle_graphics.cpp:794-798 still divides by
  a positive number. But it is a guest-visible cross-platform change that the PR's scope paragraph
  actively tells a reader not to look for. Name it (body + one clause at the export), and if you decide
  it is fine, say why.

## Non-blocking

1. **No arm pins the exported ORIGIN to anything.** test_videoout.cpp:890-892 asserts only `!= 0` and
   stability across two calls. A mutation returning a fixed constant, or `vblank_epoch_ns() + kVblankNs/2`
   — a **half-period phase error, the exact mutation the phase test at :338-370 exists to catch** — passes
   every new arm. There is a cheap oracle available in the same file: `GetVblankStatus` gives
   `tsc = ns` @0x10 and `processTime = (ns − t0)/1000` @0x08, so `t0 == tsc − processTime*1000` within the
   1 µs ABI quantisation. One `CHECK` closes the phase half of #3024's test coverage. (The stability arm
   does catch `vblank_now_ns()`-instead-of-epoch, which is the likeliest slip.)
2. **No arm exercises the pump.** Reverting hle_kernel_time.cpp:1139-1145 to the old `nanosleep` reddens
   nothing. Your test comment half-says this; the PR body's mutation table should say it outright — the
   three mutations are of the helper, not of the fix site. Not a request to test a detached thread; a
   request that the table not read as "the fix is mutation-covered".
3. **test_videoout.cpp:888** (`period != 16666667`) is implied by :886 (`period == P`). Harmless, but it
   is a tautology dressed as an independent arm.
4. **precise_sleep.hpp:43** — "a zero period would busy-spin on a fixed deadline" is not what the guard
   prevents: returning `now_ns` busy-spins too (`sleep_until(now)` returns immediately). What it prevents
   is a deadline receding into the past. The test name at :882 gets this right; the code comment does not.
5. **hle_graphics.cpp:738-739** — "Internal linkage on purpose … nothing outside it may reach these" now
   sits immediately above two functions that publish both values. Still literally true (the accessors stay
   file-local, and the export can only *advance* the anchoring, which strengthens the "epoch before now"
   invariant rather than breaking it) — but it wants a clause acknowledging the value getters.
6. **hle_kernel_time.cpp:10-11** — the two `extern "C"` declarations are spliced into the middle of the
   include block, between `timedwait_census.hpp` and `runtime_module_load.hpp`. Move them below the
   includes; this file has order-sensitive Windows-header rules elsewhere and an interleaved declaration
   invites someone to break them.
7. **Latent test-seam coupling.** The pump's grid now derives from `vblank_epoch_ns()`, which honours
   `g_vblank_test_now_ns`. No test starts the pump today (nothing under `prosper/tests` calls
   `prosper_eq_add_vblank`), so there is no live hazard — but a future test that sets the fake clock
   *before* a vblank registration hands the pump a fake origin while it sleeps on the real clock, and a
   fake origin in the future means zero vblank kevents for as long as the difference. One sentence at the
   export, or read the epoch through a hook-free path there.
8. **New observable worth one line:** when the origin is already in the past, the pump posts its first
   vblank kevent essentially at `sceVideoOutAddEvent` time instead of one period later. Harmless (kevents
   queue and coalesce), but it is new.
9. `frame` still starts at 1 and is unchanged — noting it only because `e.data` now increments ~1.8x
   faster on Windows for any title reading it as a vblank ordinal. Consistent with the fix's intent.

mattias800 (commented):
Re-review of #3064 at head `bade7abc` (same reviewer as the previous review; I again did not re-run the
author's build or ctest).

**REJECTED** — one blocking finding, a single clause in a comment. B1 and B2 are properly resolved, B2 by
a genuinely better design than the patch I suggested. The blocking item is a *new* comment on this head
that denies the very effect B3 exists to name.

## B2 — resolved, and the new shape is stronger than what I asked for

I re-read `next_grid_deadline_ns` (precise_sleep.hpp:49-59) and the loop (hle_kernel_time.cpp:1145-1151)
rather than skimming the diff. Answering your three questions in order:

**(a) First iteration and late start: the origin's age genuinely cannot matter now.** The result is
`origin + ((now−origin)/P + 1)*P` computed from a *fresh* `now` every iteration, so for any `X = now−origin`
it is the next grid boundary strictly after `now` — phase-exact for `X = 1 ns` and for `X = 3000 P` alike,
which is what the :888-891 arm pins. Two things I flagged last time are gone as side effects, not just
mitigated: the stale-first-deadline resync cannot occur (there is no carried deadline to be stale), and the
immediate first post is gone (the first wait is always in `(0, P]`, so the pump can no longer fire a kevent
at `sceVideoOutAddEvent` time). Dropping the carried deadline also removes the failure mode the previous
helper's whole comment was about, which is the right way to retire a hazard.

**(b) A stall yields exactly one boundary, and never a backlog.** Because the answer is a pure function of
`now`, a stall of 5 periods and a stall of 997 return one boundary each (:868 probes, :884). Nothing
accumulates; there is no state in which a backlog could be stored. And phase survives the stall, which the
old shape could not do — that was the substance of B2.

**(c) `now_ns < origin_ns → origin_ns` is the right answer, not a case that should assert.** Three
reasons, and I checked each against the code rather than reasoning from the name:
* It is *required*, not defensive. Without it, `now − origin` underflows to ≈2^64, `/P + 1` is ≈1.1e12, and
  the multiply overflows to an arbitrary value that may be in the past — i.e. the exact spin the function
  exists to prevent. The guard converts a silent wrap into a defined answer.
* `origin` is the correct answer under this file's own model of the grid. `vblank_epoch_ns()` is a
  first-use anchor and hle_graphics.cpp:794-797 already treats it as the start of time (it anchors *before*
  sampling precisely so `ns − t0` cannot underflow). There are no vblanks before the epoch, so the first
  boundary of a grid anchored in the future is the origin itself, not `origin − mP`.
* Asserting would be wrong for a `constexpr` pure helper on a path a test seam can reach. And for the pump
  the consequence is defined and harmless: it reads the origin once at :1145 and would sleep until that
  instant on the real clock, then run normally. No test starts the pump today (nothing under
  `prosper/tests` calls `prosper_eq_add_vblank`), so this is latent-correctness, not live behaviour.

I also re-derived the new arms and they all hold arithmetically (`P/3`, `P−1`, `P`, `P+1`, `4P+P/2`,
`997P+12345`; the `P` probe is the interesting one — result `T0+2P`, distance exactly `P`, so
`all_within_one`'s `<=` is deliberate and correct). The `T0 = 1e9` origin keeps `T0 − 5P` well clear of
zero for the future-grid arm.

## The oracle arm — good, and the catch you hit is the arm working

`t0 == tsc − processTime*1000` (test_videoout.cpp:925-933) is exactly the oracle I meant, and the
tolerance derivation is right: `processTime` is truncated to whole µs, so `exported ≤ o_epoch < exported +
1000` and `skew < 1000` is tight rather than tuned. It reddens on the half-period mutation (skew 8341675)
and on `vblank_now_ns()`-instead-of-epoch. Two notes, neither blocking:
* It is a **same-source** oracle: both sides derive from `vblank_epoch_ns()`, so it proves the export
  *agrees with* the status path — which is precisely the claim being made, so that is fine — but it cannot
  see both being wrong together (e.g. an epoch anchored to a fake clock). Worth one sentence in the arm so
  nobody later reads it as pinning the epoch to real time.
* Your fix for the retired-handle mistake is the right one, and the added
  `vbl(...) == 0` precondition at :922-923 is the part that makes the arm honest. That the first version
  passed on a zeroed buffer is a clean instance of the charter's "check that the instrument can see the
  thing" rule; it is worth keeping the comment you wrote about it.

## B1 — resolved

hle_graphics.cpp:732-738 now states the pump is no longer a second clock, that #3024 closed the follow-up,
and why the alignment may be asserted; the stale `(:731)` is gone. I checked it against the code: there is
no `nanosleep` in the pump, one origin and one period, and the two accessors are where the comment says
they are (:775-776). The declarations also moved out of the include block (hle_kernel_time.cpp:993-997) and
now carry a note saying why they are not in a header. Accurate.

## Blocking

**B4 (new on this head). hle_graphics.cpp:742 — "a caller can schedule on the grid but cannot move it" is
false for the first caller, and that caller is normally the pump.**

`prosper_vo_vblank_grid_origin_ns()` (:775) calls `vblank_epoch_ns()`, which is *first-use anchored*. When
the pump reaches :1145 before any guest `GetVblankStatus`/`WaitVblank` — the common ordering, since
`sceVideoOutAddEvent` starts the pump — the pump is what sets the origin, and the grid then sits at a
different phase than master's would have. That is not a hypothetical: it is exactly the effect B3 is about,
and it moves guest-visible values on every platform.

So this head has the PR body about to say "the export can now anchor the epoch earlier, which shifts
guest-visible state" while the code comment says a caller "cannot move it". One of the two is wrong, and it
is the comment. The charter's rule that a correction living only in a PR body is not in the project record
is the whole reason this is blocking rather than a nit — the PR body is not what the next reader of
`hle_graphics.cpp` will open. A clause is enough:

> …publish the origin and the period as values, which cannot break that invariant. What a caller *can* do
> is be the FIRST to read the origin, which anchors it — so the pump usually sets the epoch now, earlier
> than the first guest status/wait call did on master (#3024).

## B3 — the wording I would accept, with two additions

Yours is accurate as far as it goes. I would add:

* **Say which thread does the anchoring, and where.** "the export can anchor the epoch earlier" leaves the
  mechanism implicit; "the pump thread's read of `prosper_vo_vblank_grid_origin_ns()`
  (hle_kernel_time.cpp:1145) is now usually the first use, so it anchors the epoch at
  `sceVideoOutAddEvent` time rather than at the first guest status/wait call" is checkable.
* **`sceVideoOutWaitVblank`'s wake instants shift too**, not only `GetVblankStatus`'s `count` and
  `processTime`. It computes its boundary from the same `t0` (hle_graphics.cpp:836-840), so an earlier
  origin moves every wait boundary's phase relative to wall-clock on Linux and macOS as well. Same
  judgement as yours — more faithful, low risk — but a title pacing a display thread on `WaitVblank` is
  the consumer most likely to notice, so it belongs in the sentence.

I agree with your judgement that this is more faithful than master and low risk: `count` stays positive, so
the Astro Bot `(frames*refresh)/(count − previous)` case at :794-797 is unaffected in kind.

## Non-blocking

1. **Your question 1, answered concretely: the untested surface is not nil — it is the composition, and
   there is a reachable defect living in it.** The loop samples `now`, then waits, then posts, and it
   assumes the wait does not return *before* the deadline. If it can, the next iteration's `now` is still
   below the same boundary, `next_grid_deadline_ns` correctly returns that same boundary again — and the
   pump posts **two kevents for one vblank**, a few hundred µs apart. Note the helper is not wrong here;
   returning the same boundary is what it should do. It is the composition that turns an early wake into a
   duplicate event, and the previous shape happened to absorb it (`advanced ≥ now → deadline + P`).

   Reachability, stated with the one link I could not verify marked as such: `sleep_until_steady_ns`'s
   Windows *fallback* path has no post-condition re-check — `if (steady_now_ns() >= deadline_ns) return;
   sleep_until_with_std(deadline_ns);` (precise_sleep.cpp:126-128) — unlike the timer loop above it, which
   re-reads the clock precisely so it cannot come back early. libstdc++'s `sleep_until` on a steady clock
   forwards to one `sleep_for` and does not re-check (I read
   `/usr/include/c++/13/bits/this_thread_sleep.h:87-99`, which confirms precise_sleep.cpp's fact (1)), so
   whether an early return is possible reduces to whether mingw's win32 `__sleep_for` truncates its ns→ms
   conversion. **I could not read mingw's `thread.cc` on this host, so I am flagging that, not asserting
   it.** POSIX is safe (`nanosleep` is a minimum), and the primary Windows path is safe (the re-check loop).
   Exposure is therefore: high-resolution timer creation failed *and* the ns→ms conversion truncates.

   Bounded and coalescible (vblank posts replace in place, hle_kernel_time.cpp:1073-1090), so I am not
   blocking on it. Two candidate resolutions: guard in the pump, or — better, and where the contract is
   actually stated (precise_sleep.hpp:22-23, "Block until `deadline_ns`") — make the fallback honour its own
   documented postcondition. The second is a separate one-line change and wants its own issue rather than
   scope creep here.
2. Other things the arms still cannot see, for completeness: that the pump calls the helper at all
   (acknowledged plainly at :838-843 — good, and the acknowledgement is the right resolution); that it reads
   the origin once rather than per iteration (harmless, the epoch is immutable after anchoring); and the
   same-source limit of the oracle noted above. Nothing else struck me as reachable.
3. precise_sleep.hpp:52-53's zero-period comment is now accurate ("it still spins, and deliberately so") —
   noting only that I checked it, since I raised the old wording.

## Your question 2 — agreed, the macOS red is not attributable to this diff

Verified rather than accepted:
* Run 33049583115 is head `bade7abc`, and its **only** failure is `kernel_sem_timedwait` (47/255),
  `timeout took 46.46 ms via none`. Every other job passes, including Windows MinGW and Linux.
* Master at `6f6ec58a` (run 33050236613) fails the **identical** test with the **identical** shape —
  `timeout took 23.22 ms via none`, 1 of 255 — with none of this diff in it. That is the evidence that
  makes it inherited rather than caused; the differing durations are what a wall-clock flake looks like.
* The run on `8a2f7d51`, this branch's parent, passed — which is not counter-evidence, because an
  intermittent failure passes sometimes; it is why the master-side reproduction above is the load-bearing
  part.
* The diff touches neither `tests/hle/kernel/test_kernel_sem_timedwait.cpp` nor any semaphore path, and its
  only shared file is precise_sleep.hpp, where it adds and removes a `constexpr` free function without
  touching `sleep_until_steady_ns`, `sleep_backend()` or the `"none"` string the failure prints.

Your plan (merge #3066, merge master in, require green) is the right one. One practical note: #3066 also
touches `prosper/CMakeLists.txt`, so after merging master in, **re-quote the ctest count** rather than
carrying 247/247 forward — the charter wants the count next to the exit code, and the count is what makes
it falsifiable.

Also checked, since it is a merge gate and cheap: no `Claude-Session:` trailers on either commit
(`git log origin/master..HEAD --format=%B | grep -c '^Claude-Session:'` → 0).

Fix B4's clause and I will re-approve on the corrected head; nothing else here needs another round.

mattias800 (commented):
Third review of #3064, at head `371a7ac4` (same reviewer throughout; I have not re-run the author's build
or ctest on any head).

**APPROVED.**

B4 is fixed, and fixed in the direction I would have asked for rather than the minimum: the new paragraph
at hle_graphics.cpp:739-753 keeps the true half (the accessors publish values, so nothing can retarget the
grid), then states plainly that the first *reader* sets where the grid starts, names the site, and
enumerates all three guest-visible consequences including `WaitVblank`'s phase. I checked each clause
against the code: `vblank_epoch_ns()` is the first-use anchor behind the export at :786; `count`'s formula
at :812 is `1 + (ns − t0)/kVblankNs`, so an earlier `t0` does move the first value; and the wait boundary
derives from the same `t0`. Accurate, and the last sentence — that a comment claiming callers cannot move
the grid would have hidden it — is the part that will stop this being re-derived.

The #3074 caveat at hle_kernel_time.cpp:1146-1153 is also accurate to what I found, including the part I
could not verify being left out of the assertion rather than stated as fact. #3074 itself is open with
`bug` / `area:infra` and a title that names the asymmetry rather than the symptom. Good.

## Your question, asked a second time — no, it is not empty. Four more, one of them worth acting on

I re-read the whole loop (hle_kernel_time.cpp:1154-1179) rather than the diff, and looked at what the loop
composes with rather than at the loop alone.

**1. `frame` is a count of POSTS, not a grid ordinal — so the two "counts" can still diverge, and the loop
drops ticks by design.** `e.data = (int64_t)frame` (:1170) increments once per iteration (:1161), while
`GetVblankStatus`'s `count` is the grid index `1 + (ns − t0)/P` (hle_graphics.cpp:812). Those agree only
while the pump never skips a boundary — and skipping is exactly what the new design does on purpose: any
iteration whose overshoot plus posting work exceeds a period lands on a later boundary and posts once.
Every dropped tick offsets `data` from `count` by one, permanently and cumulatively.

So what this PR aligns is the *instants*, not the *ordinals*. That is a real and large improvement — under
master the two had no relationship at all — but hle_kernel_time.cpp:1135-1137 names the count cross-check
as the observable and says deriving from one origin and one period "closes it by construction". Under the
natural referent ("the drift of the two clocks") that sentence is fine, which is why this is not a blocking
finding and not the same class as B1/B4; under the reading its own example invites, it overclaims by
"except when a tick is dropped".

Two resolutions, and **I am deliberately not asking you to change behaviour on my word**: either qualify
the sentence, or make the payload the grid ordinal — `(now_ns − origin_ns)/period_ns + 1`, the same
expression as hle_graphics.cpp:812 — which would make the claim true as written. I would take the clause
and, if you think the second is right, a separate issue: `data = frame` predates this PR, and what the
vblank kevent's `data` field carries on hardware is not established anywhere I can see in this file (the
neighbouring comment at :1165-1169 establishes what it must *not* be for FLIP, not what it is for vblank).
Changing a guest-visible payload on a reviewer's hunch is how a CONFIDENCE: MED becomes a silent HIGH.

**2. The strongest one: dropped ticks are silent, and this change removes the symptom that used to reveal
them.** Under master a halved pump rate showed up as drift against the status grid. Now the phase stays
exact however many boundaries are missed, so a pump running at half rate looks *perfect* from the outside
— same grid, same phase, just fewer events. In #3074's regime (fallback path, ~31 ms per wait against a
16.68 ms period) the pump would skip roughly every other boundary and post at ~32 Hz under a comment that
says 60. That is the exact shape of the defect #3024 is about, one layer down and now unobservable.

The loop already computes everything needed to see it: keep the previous boundary index and compare
`(now_ns − origin_ns)/period_ns` against it; a delta above 1 is a skipped tick. Count them and report
rate-limited under an existing gate (`PROSPER_EVLOG` is already read in this file), so the halved-rate
regime is self-reporting rather than needing another standalone microbenchmark to find. This is the
charter's "prefer experiments that detect their own invalidity" applied to a production path, and it is the
one item here I would actually spend code on — though it is a follow-up, not a condition of this PR.

**3. *Which* code anchors the grid is a race, not a fact.** `ensure_pump()` (:1181) detaches a thread; the
anchoring happens when that thread first reaches :1154, which can be well after the registration call
returned, and a guest `GetVblankStatus`/`WaitVblank` inside that window anchors it instead. Both orderings
produce one grid used by everything, so it is benign — but hle_graphics.cpp:748's "anchored at
vblank-event registration" is approximate, and the grid's phase relative to wall-clock is now
nondeterministic across runs by the thread-start latency. Worth a clause, because the alternative is
somebody later chasing "why did `count` start at 3 this run and 1 the last".

**4. A flip-only registration also starts the pump, so it also anchors the grid.**
`prosper_eq_add_flip` calls the same `ensure_pump()` (:1187) as `prosper_eq_add_vblank` (:1203). A title
that registers a flip event and never a vblank event still gets a pump thread, which anchors the epoch and
then wakes 60 times a second to post to an empty `g_vblank_regs`. hle_graphics.cpp:748's "as soon as a
title registers a vblank event" should read "a vblank *or flip* event" — small, but it is the sentence a
reader will use to predict when the epoch is set.

**5. Pre-existing, not attributable to this diff, mentioned only because the rate change touches its
probability.** The pump is detached and never exits, so at process teardown it can be mid-post touching
statics (`g_eq_mx`, `g_vblank_regs`) while destructors run; doubling the tick rate roughly doubles the
per-second exposure. Likewise the `thread_local` waitable-timer handle that `sleep_until_steady_ns` creates
on that thread is never destroyed. Both were true before this PR and are harmless in practice; I raise them
only because you asked what a reader of the loop would not anticipate, and "this thread outlives the
statics it writes" is on that list.

Nothing in 1-5 is a reason to hold the PR. 1, 3 and 4 are one clause each; 2 is a follow-up issue.

## Merge preconditions, restating only what is already agreed

* #3066 merged, master merged in, macOS green — my verification that the red is inherited stands
  (identical `kernel_sem_timedwait` / "via none" failure on master `6f6ec58a` with none of this diff in it).
* Re-quote the ctest count after the merge rather than carrying 247/247 forward, since #3066 and #3070 both
  move it.
* B3's fuller wording in the body, which you have already landed.

No further review round needed from me for 1-5; if you take the payload change in item 1 rather than the
clause, that one is worth a fresh read, since it alters a guest-visible field.

mattias800:
### Merged as `19df6974`, and a note on the approval's staleness

The approval was posted on `371a7ac4`. The merged head was `d0623f56`, which is `371a7ac4` plus one commit of mine (`6615d386`) and a merge of master. So the approval did not literally cover the head I merged, and the charter requires establishing by **content** whether it still applies rather than assuming in either direction.

It does, and here is the content: `git diff 371a7ac4 <head>` over the four touched files is **entirely `//` comment lines**, and every hunk is one of the non-blocking items the third review itself enumerated —

- the `frame`-vs-grid-ordinal qualification (review item 1),
- the thread-start race and the flip-registration path in the anchoring note (items 3 and 4),
- and the recorded cost of the phase property, filed as #3075 (item 2).

No executable line changed. The reviewer had also said of the earlier B2 remainder that it was "asked for in this push, no re-review needed", which is the same disposition.

**Where I fell short of my own process:** the command I ran to check this printed a fixed `(empty above = my four files are byte-identical to the approved commit)` line regardless of what the diff actually said — and the diff was not empty. I read it and the conclusion holds, but the check as written could not have told me otherwise, which is precisely the void-control shape this series has already tripped over twice (#3071's `check_merge_result` probe, and #3066's green-Rosetta-at-reviewed-head). A comparison whose verdict is printed unconditionally is not a comparison.

Recording it because the outcome being fine is not the point: the next person to reach for that one-liner should know it does not gate anything.

Fixes #3024, now closed. Follow-ups from this review: #3074 (the `Win32SleepFallback` early return) and #3075 (dropped ticks are now invisible).

Issue #3013, opened 2026-08-26
Windows: audio underruns in every title — every guest timed wait resolves on the winpthreads ~15.6 ms tick, so a 5.33 ms grain cadence runs 3x too slow


mattias800:
Audio underruns continuously on Windows in **every** title, while Linux is clean. The cause is not in
the audio path at all: on Windows **every guest timed wait resolves on the winpthreads master tick
(~15.6 ms)**, so a guest mixer that paces small grains gets woken roughly 3x too slowly and can never
keep the device queue fed.

## Measured on this host (Windows 11, MinGW-w64/UCRT, GCC 16.1)

A hand-built probe **outside prosper** (so it tests the primitives, not our wrapper), asking for
**5.33 ms** — one 256-frame grain at 48 kHz, the cadence FMOD paces a mixer on:

| primitive | no `timeBeginPeriod` | with `timeBeginPeriod(1)` |
| --- | --- | --- |
| `nanosleep(5.33 ms)` | **15.67 ms** | 15.59 ms |
| `sem_timedwait(5.33 ms)` | **15.57 ms** | 15.66 ms |
| `cv.wait_for(5.33 ms)` | **15.73 ms** | 15.63 ms |
| `sem_timedwait` **with a poster every 0.5 ms** | 0.01 ms | 0.00 ms |

Two things that last row settles, and it is the control that makes this diagnosis rather than a
correlation:

- **The wake path is fine.** A semaphore that is actually posted returns in 0.01 ms. Only the
  **timeout** path is quantized, so this is not general scheduling latency or thread-wake cost.
- **The lever moves**, so the probe is not simply reporting a constant.

`timeBeginPeriod(1)` not helping is the counter-intuitive part: raising the process timer resolution
is the obvious remedy and it does nothing here, so the tick is winpthreads' own rather than the OS
timer's.

**prosper does not call `timeBeginPeriod` anywhere** — six prose-only lines across four files, no call, `winmm`
linked nowhere, and `src/host/platform/precise_sleep.cpp:14` states it outright. An earlier version of
this text claimed it did; that came from an uncommitted local diff which was later discarded, and was
wrong. The probe applied the call in-process itself, which is where the second column above comes
from. Since nothing in the tree raises the period, `precise_sleep.cpp:16`'s own figure — `Sleep(1)` =
15.554 ms by default against 1.930 ms after the call — describes a **live lever**, not one already
pulled.

For contrast, on the same box and with no timer-period change,
[...]

PR #3066, opened 2026-08-27
fix(test): the sem_timedwait wall-clock ceiling could pass the defect it guarded, and flaked on Rosetta

mattias800:
**Master is red on `macOS (x86_64 / Rosetta 2)` because of a test I merged in #3044.** This fixes it. Refs #3013, #3033.

```
47 - kernel_sem_timedwait (Failed)
   (timeout took 18.54 ms via none)
```

## What broke

#3044 asserted the 5 ms guest semaphore timed wait returned in under 12 ms. On POSIX that ceiling asserts nothing about prosper — the native `sem_timedwait` is explicitly *untouched* by that change (`#ifndef _WIN32` at `hle_kernel.cpp:3097`) — so it was purely an assertion about the host scheduler. On an emulated-x86 VM under load it delivered at 18.54 ms.

The charter names this exact failure, and `precise_sleep.hpp` names it again over `sleep_backend()`: assert the mechanism, "instead of asserting a wall-clock duration, which on a loaded host is a future flake". I introduced it in the same PR whose review had just removed a *different* unfalsifiable arm.

## The bound was not fragile — it was unsound

Making the ceiling Windows-only was the obvious fix. I measured it before shipping it, and it fails in the worse direction. With the fix reverted (`#ifndef _WIN32` → `#if 1`, forcing the native winpthreads path on Windows), the defect delivers at:

| run | delivered |
| --- | --- |
| 1 | 14.69 ms |
| 2 | 12.06 ms |
| 3 | 17.75 ms |
| 4 | **10.00 ms** |
| 5 | 15.32 ms |

**The 10.00 ms run passes a 12 ms ceiling.** The bound would have let the defect through, which is strictly worse than flaking — a flake is noisy, this is silent.

And it must be so: a quantized wait returns at the next **tick boundary**, so a 5 ms request lands anywhere in roughly [5, 20.6] ms depending on where the request falls within the ~15.6 ms tick. The defect's timing distribution *overlaps the fix's*. No single-run wall-clock bound separates them on any platform.

The mechanism arm caught all five, because it reads a state variable rather than a clock.

## The change

- **Arm 3 bounds** are now a stub/unit/hang guard on both platforms: `>= 4.0 ms` (an instant-returning stub fails) and `< 500.0 ms` (a wrong unit or a hang fails). Neither pretends to discriminate.
- **Arm 1 (mechanism)** is the sole discriminator, unchanged, and is what the file header already claimed.
- **Arm 4's `< 1.0 ms`** fast-path bound had the same latent problem — fine on a real x86 host, a waiting flake on an emulated one. It is now `< 100.0 ms` **on POSIX only**; Windows keeps `< 1.0 ms`. (This bullet originally said "on both platforms", which the diff does not do — the code was right and the description was wrong. Caught in review.) The arm has also been strengthened to assert the **count**, not the latency: see the review response below.
- **The header's arm list** is updated with the five measurements. The first version of this test was wrong about its own arms *in a confident label*, and that label is what stopped anyone re-deriving it — so a bound that changes changes the list too.

## Verification

Windows, MinGW-w64/UCRT GCC 16.1, RelWithDebInfo:

```
cmake --build prosper/build-mingw-app                        # clean
100% tests passed, 0 tests failed out of 247                 # exit 0
```

Mutation (forcing the native path at `hle_kernel.cpp:3097`): arm 1 reddens 3 of 3 runs; restored, passes 3 of 3 at 5.49 / 5.47 / 5.39 ms.

**The POSIX arms cannot be executed on this host.** Both changes only *widen* a bound, so they can remove a failure and cannot add one — but `Linux` and `macOS (x86_64 / Rosetta 2)` are the checks that matter here, and this should not merge until both are green.

## Note on process

Two things went wrong that are worth stating rather than burying. #3044's own `macOS (x86_64 / Rosetta 2)` job was **green** at the reviewed head, which is what I merged on; the flake surfaced on the next PR to merge master. A green run on a timing assertion is evidence about one sample, not about the assertion — I treated it as the latter. And while verifying this fix, my first mutation attempt silently patched the wrong `#ifndef _WIN32` (line 829, not 3097) and reported `PASSED`, which I nearly took as "the discriminator is intact". A mutation that does not visibly apply is not a mutation.

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -648,7 +648,15 @@ if(TARGET prosper_core)
   # other two arms measured a POSTED semaphore, which unfixed winpthreads serves in 0.01 ms -- better
   # than the poll loop, so they preferred the defect. A mechanism assertion cannot be satisfied by a
   # lucky duration and does not flake under load. Mutation-checked: reverting to the winpthreads
-  # delegation reddens two arms and prints "via none".
+  # delegation reddens the MECHANISM arm and prints "via none".
+  #
+  # It used to say "reddens two arms", the second being a wall-clock ceiling, and that ceiling
+  # is gone: measured across five reverted runs the defect delivered at 14.69, 12.06, 17.75,
+  # 10.00 and 15.32 ms, so a 12 ms ceiling PASSES the defect. A quantized wait returns at the
+  # next tick boundary, so the defect's latency distribution overlaps the fix's and no
+  # single-run bound separates them. The same ceiling also flaked on macOS/Rosetta at 18.54 and
+  # 46.46 ms, where the native call is untouched by the fix anyway. The mechanism arm is the
+  # discriminator; the bounds that remain are a stub/unit guard.
   add_executable(test_kernel_sem_timedwait tests/hle/kernel/test_kernel_sem_timedwait.cpp)
   target_link_libraries(test_kernel_sem_timedwait PRIVATE prosper_core)
   add_test(NAME kernel_sem_timedwait COMMAND test_kernel_sem_timedwait)
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -650,13 +650,14 @@ if(TARGET prosper_core)
   # lucky duration and does not flake under load. Mutation-checked: reverting to the winpthreads
   # delegation reddens the MECHANISM arm and prints "via none".
   #
-  # It used to say "reddens two arms", the second being a wall-clock ceiling, and that ceiling
-  # is gone: measured across five reverted runs the defect delivered at 14.69, 12.06, 17.75,
-  # 10.00 and 15.32 ms, so a 12 ms ceiling PASSES the defect. A quantized wait returns at the
-  # next tick boundary, so the defect's latency distribution overlaps the fix's and no
-  # single-run bound separates them. The same ceiling also flaked on macOS/Rosetta at 18.54 and
-  # 46.46 ms, where the native call is untouched by the fix anyway. The mechanism arm is the
-  # discriminator; the bounds that remain are a stub/unit guard.
+  # It used to say "reddens two arms", the second being a wall-clock ceiling, and that ceiling is
+  # gone: it would have PASSED the defect, and it also flaked on macOS/Rosetta. The mechanism
+  # arm is the discriminator; the bounds that remain are a stub/unit guard.
+  #
+  # The measurements and the argument live in ONE place, the test file's own header. They were
+  # briefly in both, with two different lists -- and a third Rosetta figure existed in neither
+  # (23.22 ms, found in a CI log during review). Two copies of a figure list is one copy too
+  # many, and this is the copy that gets read second.
   add_executable(test_kernel_sem_timedwait tests/hle/kernel/test_kernel_sem_timedwait.cpp)
   target_link_libraries(test_kernel_sem_timedwait PRIVATE prosper_core)
   add_test(NAME kernel_sem_timedwait COMMAND test_kernel_sem_timedwait)
--- a/prosper/tests/hle/kernel/test_kernel_sem_timedwait.cpp
+++ b/prosper/tests/hle/kernel/test_kernel_sem_timedwait.cpp
@@ -16,8 +16,11 @@
 // on a loaded host.
 //
 // So the arms are:
-//   1. MECHANISM  - the wait was served by the high-resolution timer, not ::Sleep's tick. Reddens on
-//                   the pre-fix code, which never touches precise_sleep at all.
+//   1. MECHANISM  - WHICH primitive served the wait. On Windows: the high-resolution timer, not
+//                   ::Sleep's tick -- reddens on the pre-fix code, which never touches
+//                   precise_sleep at all. On POSIX the assertion is the OPPOSITE value,
+//                   "none", because the native sem_timedwait is kept there and must not start
+//                   polling; the two are different assertions, not one with a platform tweak.
 //   2. ENCODING   - the timeout returns exactly 0x8002003c, FreeBSD ETIMEDOUT as the guest reads it.
 //                   Nothing else in the suite covers this path's encoding.
 //   3. BOUNDS     - a stub/unit/hang guard, and NOT the discriminator. With the fix reverted the
@@ -27,7 +30,12 @@
 //                   ~[5, 20.6] ms and the two distributions overlap. The same 12 ms ceiling
 //                   also flaked on macOS/Rosetta at 18.54 ms, where the native call is
 //                   untouched by this change anyway. Arm 1 caught all five.
-//   4. FAST PATH  - an already-posted semaphore never enters the loop.
+//   4. FAST PATH  - an already-posted semaphore is acquired at once, asserted as a COUNT of 0
+//                   afterwards rather than as a latency. The latency bound is kept as well but
+//                   is Windows-tight / POSIX-loose, and it cannot see the fast path at all:
+//                   the pre-loop trywait and the first call inside the loop are identical, so
+//                   deleting the fast path changes no duration. "Never enters the loop" is
+//                   therefore NOT what this arm asserts, and there is no loop on POSIX.
 //
 // The header being right about the arms is part of the test. Its first version was wrong about
 // them in the same breath as the arms were wrong, and the confident label is what stopped anyone
@@ -61,20 +69,34 @@ int main() {
     HleFn sem_init_fn = Hle::lookup(nid_hash("scePthreadSemInit").c_str());
     HleFn timedwait   = Hle::lookup(nid_hash("scePthreadSemTimedwait").c_str());
     HleFn post        = Hle::lookup(nid_hash("scePthreadSemPost").c_str());
+    // Optional: used only to make ARM 4 assert the COUNT rather than a duration. Guarded so a
+    // build where it is unregistered loses one assertion instead of the whole test.
+    HleFn getvalue    = Hle::lookup(nid_hash("scePthreadSemGetvalue").c_str());
     CHECK(sem_init_fn != nullptr, "scePthreadSemInit is registered");
     CHECK(timedwait != nullptr, "scePthreadSemTimedwait is registered");
     CHECK(post != nullptr, "scePthreadSemPost is registered");
     if (!sem_init_fn || !timedwait || !post) { printf("== FAIL (unresolved) ==\n"); return 1; }
 
-    // The guest handle is an opaque slot; ensure_sem backs it with a real sem_t. Initialised through
-    // the HLE and CHECKED - the first version declared a bare `sem_t slot;` and called init with its
-    // result ignored, so an init that silently failed would have left every arm below operating on an
-    // uninitialised object.
-    sem_t slot;
+    // The guest handle is a POINTER CELL, not a semaphore. k_sem_init does
+    //     `*(void**)(uintptr_t)a0 = s;`
+    // i.e. it heap-allocates the sem_t and writes the POINTER through the handle, and every
+    // other member of the family reads it back with an 8-byte load (`ensure_sem`). So the slot
+    // must be pointer-sized.
+    //
+    // This was `sem_t slot;`, which is a latent stack overwrite that happens to be invisible
+    // on two of three platforms: winpthreads' sem_t is a pointer (8 bytes) and glibc's is 32,
+    // but **Darwin's is `int`** -- so on macOS the 8-byte write ran four bytes past a 4-byte
+    // automatic. Found in review of the very PR whose purpose was to green the macOS job.
+    void* slot = nullptr;
     const uint64_t handle = (uint64_t)(uintptr_t)&slot;
     // Asserts that init REPORTS success, not that the count is observably 0 -- the count is then
-    // established by arm 1 timing out rather than being acquired. The earlier message claimed the
-    // stronger thing.
+    // established by arm 1 timing out rather than being acquired.
+    //
+    // Weaker still than that, and named rather than left implied: k_sem_init DISCARDS
+    // sem_init's return and returns 0 unconditionally, so this CHECK cannot fail as the code
+    // stands. It is kept as a tripwire for the day that is fixed, and the fix is #3068 --
+    // which matters on macOS, where unnamed POSIX semaphores are ENOSYS and this family may
+    // therefore be operating on an uninitialised object while reporting success.
     CHECK(sem_init_fn(handle, 0, 0, 0, 0, 0) == 0, "scePthreadSemInit reports success");
 
     // ARM 1 + 2 + 3: one unposted wait, three independent properties.
@@ -147,6 +169,16 @@ int main() {
         const uint64_t rc = timedwait(handle, 1000000, 0, 0, 0, 0);
         const double ms = std::chrono::duration<double, std::milli>(clk::now() - t0).count();
         CHECK(rc == 0, "an already-posted semaphore is acquired");
+        // The count is the observable, and it is what "acquired" MEANS. Review found the arm
+        // could not see its own stated purpose: the pre-loop sem_trywait and the first
+        // statement inside the loop are the identical call, so deleting the fast path is
+        // invisible to any timing bound. A count of 0 after the acquire is not.
+        if (getvalue) {
+            int v = -1;
+            CHECK(getvalue(handle, (uint64_t)(uintptr_t)&v, 0, 0, 0, 0) == 0,
+                  "getvalue succeeds after the acquire");
+            CHECK(v == 0, "...and the count really was consumed, not merely reported as taken");
+        }
         // Same asymmetry as ARM 3, and the same reasoning: 1 ms is a fair bound on a real x86
         // host, and on an emulated-x86 CI VM it is a latent flake that happened not to fire yet.
         // 100 ms still fails a fast path that fell through to waiting out the 1 s timeout, which
--- a/prosper/tests/hle/kernel/test_kernel_sem_timedwait.cpp
+++ b/prosper/tests/hle/kernel/test_kernel_sem_timedwait.cpp
@@ -23,13 +23,29 @@
 //                   polling; the two are different assertions, not one with a platform tweak.
 //   2. ENCODING   - the timeout returns exactly 0x8002003c, FreeBSD ETIMEDOUT as the guest reads it.
 //                   Nothing else in the suite covers this path's encoding.
-//   3. BOUNDS     - a stub/unit/hang guard, and NOT the discriminator. With the fix reverted the
-//                   defect delivered at 14.69, 12.06, 17.75, 10.00 and 15.32 ms -- so a 12 ms
-//                   ceiling would have PASSED the defect, not merely flaked. A quantized wait
-//                   returns at the next tick boundary, so a 5 ms request lands anywhere in
-//                   ~[5, 20.6] ms and the two distributions overlap. The same 12 ms ceiling
-//                   also flaked on macOS/Rosetta at 18.54 ms, where the native call is
-//                   untouched by this change anyway. Arm 1 caught all five.
+//   3. BOUNDS     - a stub/unit guard, and NOT the discriminator. THE canonical figures for this
+//                   file, referenced from CMakeLists.txt rather than restated there:
+//
+//                     defect, Windows, fix reverted:  14.69  12.06  17.75  10.00  15.32 ms
+//                     fix, Windows:                    5.49   5.47   5.39 ms
+//                     macOS/Rosetta, native path:     18.54  23.22  46.46 ms
+//
+//                   A 12 ms ceiling would have PASSED the defect (the 10.00 ms run), not
+//                   merely flaked -- silently, which is worse. And it cannot be repaired by
+//                   moving it, for a reason that needs no measurement at all: a quantized
+//                   wait returns at the next tick BOUNDARY, so its latency is T-p or 2T-p for
+//                   phase p within the tick, and as p approaches T-N from below the defect's
+//                   latency approaches the REQUESTED 5 ms exactly -- below the fix's own
+//                   5.39 ms, for any tick length T. The distributions overlap by
+//                   construction. (Second review of #3066 supplied that argument; the five
+//                   samples are kept because they also discriminate the MODEL -- a
+//                   whole-ticks-from-the-request model predicts clustering at ~15.6/31.2 ms,
+//                   which 10.00 and 12.06 falsify.)
+//
+//                   Arm 1 caught all five reverted runs. On Rosetta the native path is
+//                   untouched by this change, and arm 4's posted acquire measures SUB-
+//                   MILLISECOND there, so the 100 ms bound has over 100x headroom on a
+//                   measured value rather than a guessed one.
 //   4. FAST PATH  - an already-posted semaphore is acquired at once, asserted as a COUNT of 0
 //                   afterwards rather than as a latency. The latency bound is kept as well but
 //                   is Windows-tight / POSIX-loose, and it cannot see the fast path at all:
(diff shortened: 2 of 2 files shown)


mattias800:
### A second Rosetta data point, from an unrelated PR, and it widens the case

#3064 (a vblank change that merely *contains* master) failed the same job while this PR was in review:

```
(timeout took 46.46 ms via none)
FAILED (1)
```

So macOS/Rosetta has now delivered a 5 ms guest semaphore timeout in **18.54 ms** and **46.46 ms** on two independent runs — a 2.5x spread between two samples of the same assertion. That is the variance a wall-clock ceiling on that platform is exposed to, and it is why the ceiling here is 500 ms rather than something tighter-looking: 46.46 ms is the worst observed, so 500 ms keeps roughly an order of magnitude of headroom while still failing a wrong unit (5 s) or a hang.

It also means the flake was not a one-off unlucky runner. Two of the last two PRs to touch master's CI hit it.

Note that arm 1 reported `via none` on both — correctly, since POSIX keeps the native call — so the mechanism assertion was accurate on exactly the runs where the timing assertion was wrong. That is the whole argument of this PR, observed rather than reasoned.

mattias800 (commented):
**REJECTED** — three blocking findings, all small (two are comment-only, one is a one-token code change). The central claim holds and the direction of the change is right; I'd re-review immediately on a corrected head.

---

## 1. Is the central claim right?

**Yes, and by a sound route — with one caveat about which sentence is doing the work.**

The tick model is correct as stated. A `WaitForSingleObject`-family timeout is converted to whole clock ticks and expires at the first clock interrupt at or after the due time, so for a request `N` and tick `T` the delivered latency is the distance from the request to the next boundary after `N`: with the request at phase `p` in the tick, latency is `T-p` when `p+N <= T` (range `[N, T]`) and `2T-p` otherwise (range `(T, N+T]`). Overall `[N, N+T]` = `[5, 20.6]` for `N=5, T=15.6`. **A 10.00 ms observation follows directly** — it is `p = 5.6`. Nothing about it is anomalous.

Your own five samples are independent evidence *for* this model and against the competing one. If winpthreads instead polled or slept in whole ticks from the request, latencies would cluster at ~15.6/31.2 ms; 10.00 and 12.06 falsify that, and the five values spread across the interval rather than clustering. That is a nicer confirmation than the argument in the comment, and it is free.

**Where I'd tighten the reasoning, because as written it leans on the weakest datum.** The comment argues overlap from the *interval* `[5, 20.6]` and then from the *single* 10.00 ms sample. The interval is not needed and neither is the sample. The decisive fact is the infimum: as `p -> (T-N)+`, latency approaches **exactly the requested 5 ms** from above, while the fix delivers 5.39-5.49 ms. The defect's distribution therefore reaches *below* the fix's typical value for **any** tick length, and no ceiling can separate them without also cutting the fix. Stated that way the conclusion survives a challenge to any individual measurement — which matters here, see the last non-blocking item. Worth one sentence in the comment (`test_kernel_sem_timedwait.cpp:106-110`).

Non-blocking, and pre-existing rather than yours: `hle_kernel.cpp:3065-3066` attributes the quantization to "winpthreads', whose timed waits resolve on its own master tick regardless of `timeBeginPeriod`". The quantization your model relies on belongs to the NT wait timer applied to the millisecond timeout winpthreads passes down, not to anything winpthreads owns. The *model* is right and the *conclusion* is right; that older sentence's mechanism is the shaky part. `CONFIDENCE: MED` — I did not open winpthreads' `sem.c` for this review, so I am not asking you to change it on my word, only flagging that this PR's model and that comment's model are not the same claim.

## 2. Is the test now weaker than it needs to be?

**Arm 1 is sufficient on Windows for the defect this file exists to guard**, and I enumerated the alternatives rather than assuming: reverting to the winpthreads delegation reddens it; the HR timer being unavailable reddens it (`Win32SleepFallback` is a distinct name); a wrong unit in either direction is caught by arm 3's 4.0/500.0; a wrong rc is caught by arm 2.

**One regression class is genuinely given up**: an overshooting poll loop in the 12-500 ms band. Note it takes *two* simultaneous edits to get there — the deadline stays exact as long as the `remaining < slice ? remaining : slice` clamp at `hle_kernel.cpp:3133` survives, so you also have to grow the slices past ~7 ms *and* drop the clamp. And the 12 ms bound bought that coverage unsoundly, so this is not a loss of real power, it is the honest price.

**The assertion that recovers it without being a clock assertion** is not another bound: extract the deadline/slice arithmetic (`hle_kernel.cpp:3095`, `:3109-3110`, `:3132-3133`) into a pure, directly-testable helper and unit-test it. That gets you (a) the clamp, i.e. "the loop never sleeps past the deadline", (b) the two saturating computations added in #3044, which have **no test at all** today, before or after this PR, and (c) the fine/coarse crossover — all clock-free. Non-blocking follow-up, but it is the right answer to "what replaces the bound".

**The POSIX arm is a guard, not a tautology** — narrowly. It is not compile-time vacuous: `sleep_backend()` is a real thread-local that a unification onto the poll loop would move, and the `"none"` pre-check at `:84` is what stops it passing on a thread that simply never slept. What it is *not* is any assertion about the POSIX implementation being right; the comment at `:132-134` says exactly that, which is the standard this file is held to. Accept as is.

## 3. Arm 4

**100 ms is meaningful on POSIX** — two orders of magnitude under the 1 s timeout, and "fell through to waiting out the timeout" is the only defect reachable there (there is no poll loop on that branch at all). Don't tighten it; on an emulated host you would just be re-introducing the thing this PR is fixing.

But the arm does not test what it says, and this is the interesting part. `hle_kernel.cpp:3103` is the pre-loop `sem_trywait`, and `:3123` — the **first statement inside the loop** — is the identical `sem_trywait`. So deleting the pre-loop fast path changes the delivered latency by microseconds and **arm 4 cannot see it**. What the arm actually detects is "a posted semaphore is not made to wait", which is a real and worthwhile defect class, just not "never enters the loop". See blocking finding 2.

The clock-free strengthening is cheap and available: `scePthreadSemGetvalue` is registered. Assert the count is **0** after arm 4 — that discriminates "acquired" from "returned 0 without decrementing", which no timing bound can. The same call after arm 3 asserts the timed-out wait consumed nothing. Both read state instead of a clock, which is this file's own thesis.

## 4. Does the header match the arms? — BLOCKING

Checked line by line against the code. Arm 2 (`:21-22`) is accurate (`kSceKernelErrorETIMEDOUT` = `sce_kernel_error(ETimedOut)` = 0x8002003c, `sce_errno.hpp:141`; the "nothing else covers *this path's* encoding" qualifier is still true — the other 0x8002003c assertions in the suite are other entry points). Arm 3 (`:23-29`) is accurate and is the best line in the file. Two are not:

**BLOCKING (1) — `prosper/CMakeLists.txt:650-651`.** *"Mutation-checked: reverting to the winpthreads delegation reddens two arms and prints \"via none\"."* After this PR only **one** arm reddens, and this PR's own table falsifies the sentence even for the old code: the 10.00 ms run reddened one arm, not two. This is the registration site — the first thing a reader hunting the test's power finds — and it is the same "confident label over an assertion that cannot discriminate" pattern the whole file documents. Your own rule at `:32-34` ("a bound that changes changes this list too") applies to it. The PR currently touches only the test file.

**BLOCKING (2) — arm labels that are Windows-only descriptions of arms that now run on three platforms.** This is the precise failure #3044 shipped and this PR is cleaning up, so it should not survive it:
- `test_kernel_sem_timedwait.cpp:19-20` — arm 1 is described as "the wait was served by the high-resolution timer ... Reddens on the pre-fix code". On POSIX the arm asserts the *opposite value* (`"none"`, `:135-136`) and cannot redden on the pre-fix code, because there is no pre-fix code there. The header needs both halves.
- `:30` and `:141` — "an already-posted semaphore never enters the loop" / "is taken by the trywait fast path and never enters the loop". Per section 3 the test cannot distinguish the fast path from the loop's own first `sem_trywait`, and on POSIX there is no loop to enter. Both lines claim more than any arm can fail on.
- Same line: the header gives arm 4 no platform split, but the code now has one (`:154-158`), and the inline comment above it reads as though 100 ms applies everywhere.

## 5. Anything else unable to fail?

**BLOCKING (3) — `test_kernel_sem_timedwait.cpp:73`, `sem_t slot;` is the wrong type and overflows on macOS.** The slot is a pointer cell everywhere it is touched: `k_sem_init` does `*(void**)a0 = s` (`hle_kernel.cpp:3057`) and `ensure_sem` does an 8-byte `__atomic_load_n((void**)slot_addr)` (`hle_kernel.cpp:694-697`). The test does not include `posix_shim.hpp`, so on Darwin `sem_t` is the system `typedef int sem_t;` — 4 bytes — and the HLE writes 8 into it and reads 8 back out. That is an overflow of a stack automatic on **exactly the platform whose job this PR exists to make green**, and a green Rosetta run over it is not the evidence the PR is asking for. Fix is one token: `void* slot = nullptr;` (or `uint64_t`), which also makes the "opaque slot" in the comment at `:69` literally true. Pre-existing from #3044, not introduced here; I am blocking on it because this PR is the natural and cheapest place and the platform is the point.

Non-blocking, unable to fail, all pre-existing:
- **`:78`, the init CHECK.** `k_sem_init` (`hle_kernel.cpp:3057`) **discards `sem_init`'s return value** and returns 0 for any non-zero `a0`. So this CHECK can only fail if the handle is null, which `&slot` never is. The comment at `:70-72` says the point of it is that "an init that silently failed would have left every arm below operating on an uninitialised object" — that hole is still open, and it is closed in the HLE, not the test. It matters concretely: Darwin's native `sem_init` **is** ENOSYS (`posix_shim.hpp:163`), which is the whole reason the shim exists, so a failing init is not hypothetical here.
- **`:143`, "post succeeds".** Same shape; `sem_post` on a live semaphore does not fail. Fine as a precondition, not an arm.
- **`:117`'s "or a hang fails this"** overstates slightly: a real hang never reaches the CHECK and is caught by `TIMEOUT 60` at `CMakeLists.txt:658`. The 500 ms bound catches a *slow return*, which is a different thing, and worth distinguishing since the CMake comment already claims the hang coverage.
- **PR body vs diff.** The body says arm 4's bound was widened "on both platforms"; the diff keeps `< 1.0` on Windows behind `#ifdef _WIN32` (`:154-155`). The code is better than the body — please fix the body, since it is the durable record, and the only reason I went looking was that the description told me to expect something the code does not do.
- **The five-run table should carry its own validity column.** Your process note records the near-miss where a mutation silently did not apply and reported PASSED. This experiment self-detects — arm 1 reddens when the mutation is live — so add "arm 1 red? y/y/y/y/y" to the table. As posted, the five rows and the "3 of 3" mutation runs read as separate sets, and the 10.00 ms row is the single datum the unsoundness argument is built on. Landing the infimum argument from section 1 stops it being load-bearing, which is the better fix.

## Not blocking, and I checked it

Both bound changes only widen, and the arm-4 change is POSIX-only, so neither can introduce a red where there was a green — your reasoning there is right. I did not re-run your builds. Your gate on `Linux` and `macOS (x86_64 / Rosetta 2)` before merging is the right call; both were still pending when I looked.

mattias800:
## Review response — all three blocking findings fixed at `b8138e50`

**B3 — `sem_t slot` (the one that matters).** Fixed, and thank you: this was an 8-byte write into a 4-byte automatic on Darwin specifically, in the file whose entire purpose in this PR is to stop reddening the macOS job. `k_sem_init` does `*(void**)(uintptr_t)a0 = s`, so the handle is a pointer cell; `void* slot = nullptr;` now. Invisible from Windows (winpthreads' `sem_t` is a pointer) and from Linux (glibc's is 32 bytes) — only the platform it corrupts could show it, and that platform was already red for a different reason.

**B1 — `CMakeLists.txt:650`.** Fixed. You are right that my own "a bound that changes changes this list too" rule applies to the registration site, and I had applied it only to the file header. The comment now names the mechanism arm as the single discriminator and carries the five measurements plus both Rosetta observations.

**B2 — header/arm mismatches.** Fixed. Arm 1 is now described as *two different assertions* rather than one with a platform tweak (`win32-high-resolution-timer` on Windows, the opposite value `"none"` on POSIX). Arm 4's entry no longer claims "never enters the loop" — see below — and its platform split is stated.

## Non-blocking, taken

**Arm 4 now asserts the count.** Your point that `hle_kernel.cpp:3103` and `:3123` are the identical call is exactly right, so no latency bound could ever see the fast path. It now checks `scePthreadSemGetvalue == 0` after the acquire, guarded on the lookup so an unregistered build loses one assertion rather than the test.

Mutation-checked, and only the new assertion sees it — replacing the fast path with a peek that reports success *without consuming*:

```cpp
{ int _v=0; sem_getvalue(s,&_v); if (_v > 0) return 0; }   // MUTATION
```

reddens `...and the count really was consumed, not merely reported as taken`. No timing bound moves.

**The `:78` init CHECK.** You are right that it cannot fail, and the Darwin consequence you named is serious enough that I filed it rather than absorbing it: **#3068**. `k_sem_init` discards `sem_init`'s return and returns 0 unconditionally, and on Darwin unnamed POSIX semaphores are ENOSYS — so that family may be operating on an uninitialised object while reporting success, which is precisely the silent-unsync class the function's own header comment says it was written to eliminate. I marked the consequence `CONFIDENCE: MED` there because I cannot run macOS and Rosetta *does* return ETIMEDOUT, so either the wait on an uninitialised `sem_t` happens to time out or Darwin differs from my expectation. Someone with a macOS host should settle which. The CHECK stays as a tripwire with a pointer to the issue.

**PR body.** Corrected — it claimed arm 4 was widened "on both platforms"; the diff keeps `< 1.0` on Windows behind `#ifdef _WIN32`. The code was right and the description was wrong, which is the direction that reads as verified and travels.

## Your strengthening of the argument, adopted

> the argument doesn't need the interval or the 10.00 ms sample at all. The infimum does it — as `p → (T−N)⁺` the defect's latency approaches exactly the requested 5 ms, below the fix's 5.39–5.49 ms, for *any* tick length.

That is a better proof than mine and it does not rest on any single datum. My five samples become corroboration rather than the argument, and they still carry independent weight against the competing "whole ticks from the request" model, which would cluster at ~15.6/31.2 — 10.00 and 12.06 falsify it.

## Validity column, as asked

Per your point about my own mutation-didn't-apply near-miss, here is the table with the arm-1 verdict alongside, so the runs are visibly measuring the mutated build rather than a stale one:

| run | delivered | arm 1 verdict | backend reported |
| --- | --- | --- | --- |
| 1 | 14.69 ms | reddened | `none` |
| 2 | 12.06 ms | reddened | `none` |
| 3 | 17.75 ms | reddened | `none` |
| 4 | **10.00 ms** | reddened | `none` |
| 5 | 15.32 ms | reddened | `none` |
| restored ×3 | 5.49 / 5.47 / 5.39 ms | passed | `win32-high-resolution-timer` |

`via none` on every mutated run is the positive control: it is the state the mutation was supposed to produce, so a run that had silently built the unmutated source would have reported the timer instead. That check is exactly what my first attempt lacked when it patched line 829 and reported PASSED.

## Deliberately not done

**Extracting the deadline/slice arithmetic to a clock-free helper.** I agree with the analysis — it would cover the 12–500 ms overshoot band this PR gives up, and it would give #3044's saturating conversion and addition the test they currently lack. It is also a change to `hle_kernel.cpp`, and this PR is test-only and unblocking a red master. Doing it here would mean re-reviewing a behavioural diff before master goes green. Filed as follow-up work rather than smuggled in; **#3069**.

`Linux`, `Windows MinGW` and `macOS (x86_64 / Rosetta 2)` all went green on the previous head (`9600cdf6`) and are re-running on `b8138e50`. Merging on those three, as you said.

mattias800 (commented):
**APPROVED** on head `b8138e50`.

All three blocking findings are fixed, and one of them is fixed better than I asked. Diff scope confirmed: `git diff --stat origin/master...b8138e50` touches exactly `prosper/tests/hle/kernel/test_kernel_sem_timedwait.cpp` and `prosper/CMakeLists.txt`, 94 insertions / 16 deletions, comment-and-test only.

## Re-verification of the three

**B3 — `void* slot = nullptr;` (`:90`). Confirmed, and the comment is accurate on every claim I could check.** `k_sem_init` does `*(void**)(uintptr_t)a0 = s` (`hle_kernel.cpp:3057`) ✓. `ensure_sem` reads it with an 8-byte `__atomic_load_n((void**)…)` (`hle_kernel.cpp:694-697`) ✓. I also checked the one family member that does *not* go through `ensure_sem`: `k_sem_destroy` uses `pt_claim_slot`, which is an 8-byte `__atomic_exchange_n((void**)…)` (`hle_kernel.cpp:~700`) — same pointer-cell contract, so "every other member … reads it back with an 8-byte load" is right in substance. Darwin `sem_t` is `int` ✓; glibc's is 32 ✓. I could **not** verify "winpthreads' sem_t is a pointer (8 bytes)" — no MinGW headers on this machine — but it is not load-bearing: the claim only explains why the old bug was invisible on Windows, and a green Windows run is consistent with any size ≥ 8.

Nothing else in the file assumes the old type: `sem_t` and the `sem_*` functions now appear **only inside comments**. One consequence — `#include <semaphore.h>` at `:56` is now dead. Worth dropping; leaving it is a standing invitation to re-introduce `sem_t slot`.

An unremarked improvement: initialising to `nullptr` also converts a silently-failing init into a *detected* one. `ensure_sem` on a null slot returns null, `k_sem_timedwait` returns EINVAL, and arm 2 reddens. The old uninitialised `sem_t slot` would instead have handed the HLE a wild pointer. Worth a clause in the comment, since it is now part of why this is the right type.

**B1 — `CMakeLists.txt:651-659`. Does not overclaim.** "reddens the MECHANISM arm" matches the 3-of-3 mutation; the five timings match the test header; "a 12 ms ceiling PASSES the defect" is supported by the 10.00 ms row; "the bounds that remain are a stub/unit guard" is *more* accurate than the test's own "stub/unit/hang", since a real hang is caught by `TIMEOUT 60` at `:666` rather than by the 500 ms CHECK. No objection.

**B2 — header. Arm 1's entry (`:19-23`) is now right on both platforms**, and "the two are different assertions, not one with a platform tweak" is the correct framing. **Arm 4's entry (`:33-38`) is right and unusually honest** — it states its own blind spot, which is what I was asking for. See finding 1 below for the half that did not travel.

## The count assertion — sanity check, as asked

It cannot pass under the defect it was written for: `v` starts at `-1`, `k_sem_getvalue` writes through `a1` (`hle_kernel.cpp:3140`), so a getvalue that never ran fails too. rc-without-consume gives `v == 1` and reddens. Good arm, and your mutation is the right one.

**But it does not answer the question you asked, and the header should not imply that it does.** Deleting the pre-loop `sem_trywait` at `hle_kernel.cpp:3103` leaves the loop's first statement (`:3123`) to consume the count on iteration one — so the count still reads 0 and **the new assertion passes too**. Neither the latency bound nor the count can see that deletion, and no test at the HLE boundary can: the pre-loop trywait is *semantically indistinguishable* from the loop's first iteration. It is a pure optimisation, and its unobservability is a fact to record rather than a gap to close.

The header at `:33-35` currently reads as "asserted as a COUNT of 0 afterwards **rather than** as a latency", which invites the next reader to conclude the count is what recovers the fast-path claim. It isn't. One clause — "neither assertion can see the pre-loop trywait's deletion, because it is identical to the loop's first iteration; the count arm covers a different defect, rc==0 without consuming" — closes the door properly.

## 1. Your Q: should the infimum argument go in the code?

**Supplement, don't replace — your instinct is right.** Keep the five measurements: they are falsifiable by a future reader, they are dated to a real toolchain, and they do work the abstract argument cannot — they *discriminate the model*. Under a whole-tick-from-request model the latencies would cluster at ~15.6/31.2 ms; 10.00 and 12.06 rule that out. So the list is evidence for the boundary model, not just for the overlap.

What the infimum adds is independence from any single row, which matters because the 10.00 ms row is currently the load-bearing one. Concretely, after the measured list at `:127-132`:

> And the overlap is structural, not a property of how these five landed: as the request's phase approaches one tick-length minus the request, the boundary being waited for approaches the request itself, so the defect's latency has an infimum of exactly the 5 ms asked for — *below* the 5.39-5.49 ms the FIX delivers. That holds for any tick length, so no ceiling can cut the defect without also cutting the fix. The five samples show it is not merely theoretical.

## 2. Your Q: is one green Rosetta run enough?

**A single green run is weak evidence for a flake fix in general — your #3044 experience is the proof — but this approval does not rest on it, and neither should your merge note.** Three reasons, in order of strength:

1. **The flaking assertion was deleted, not retuned.** No remaining POSIX assertion's outcome depends on host scheduling within an order of magnitude. That is structural; no number of green runs would add to it, and one green run does not subtract from it. Had the fix been "retune the ceiling to 60 ms", a single green run would be worth nothing and I would be asking for N runs — and even then the tail is unbounded, as 18.54 → 46.46 across two samples shows.
2. **The margins are measured on the platform in question, and you can quote them.** I pulled the red Rosetta job (run `33050236613`, job `98443675501`) and it carries a **third** figure your lists do not have — **23.22 ms** — plus something more useful: `[ok] ...immediately, without entering the poll loop`. That is arm 4 passing the **old `< 1.0 ms`** bound *on Rosetta*. So the posted fast path is sub-millisecond even there, and the new 100 ms POSIX bound has >100x headroom over a measured value rather than an assumed one. For the 500 ms bound to flake, Rosetta would have to deliver a 5 ms wait at ~100x the request — about 11x worse than the worst of the three observations.
3. Everything on POSIX that *can* now fail is not clock-shaped: arm 2's encoding, arm 1's `"none"`, and the new count.

So: merge on that basis. My one ask is that the merge note says **that**, not "9/9 green" — a green `--output-on-failure` run prints no per-test output at all, so the green log by itself does not contain a single one of the numbers above. That is exactly the shape of the #3044 mistake, and quoting margins instead of a colour is what makes it not repeat.

Two small follow-ons from the same observation: **arm 4 never prints its `ms`**, so even a red log can only say pass/fail for it — one `printf` alongside arm 3's would put the margin in every future log. And the Rosetta figures now live in three places with three different lists (test header `:31` says 18.54; `CMakeLists.txt:657` says 18.54 and 46.46; the real set is 18.54, 23.22, 46.46). Keep **one** list — the test header — and have CMake point at it. A measured list duplicated across two files is the charter's stale-figure trap with a two-file blast radius.

## Non-blocking findings, in the order I would fix them

1. **`:163` and `:187` still carry the phrase the header now disavows.** `// ARM 4: an already-posted semaphore is taken by the trywait fast path and never enters the loop.` and the failure message `"...immediately, without entering the poll loop"`. The header at `:37-38` says in terms that *"Never enters the loop" is therefore NOT what this arm asserts*, and `:187`'s string is what a future reader sees **printed on failure**, telling them the loop was entered when the arm cannot show that. This is the one part of B2 that did not travel; I am not blocking on it because the correction now stands prominently 130 lines above and a reader cannot be misled without reading past it, and because it is not worth holding a red master. But it is the same class of defect this file exists to document, so please take it in this push. No re-review needed.
2. **`:74`'s optional lookup is a silent skip.** If `nid_hash("scePthreadSemGetvalue")` ever fails to resolve — a rename, a typo, a registration reshuffle — both new CHECKs vanish and the run stays green with no signal, which is the failure mode the charter's "prefer experiments that detect their own invalidity" rule names. The three lookups above it are each `CHECK`ed; this one should be too, then used unconditionally. Note your mutation does establish it ran on Windows; nothing establishes it ran on Linux or macOS, and CI cannot tell you (see above).
3. **Drop the dead `#include <semaphore.h>` (`:56`).**
4. **Add the infimum sentence** (section 1) and the **pre-loop-trywait clause** (above) to the header.
5. Optional: note in the `:80-89` comment that `nullptr` init also makes a failed init *detectable* via arm 2.

`#3068` and `#3069` are the right calls — neither belongs in a test-only PR unblocking a red master, and both are now in the durable record rather than in a review comment. `CONFIDENCE: MED` on #3068's macOS consequence is the honest label given you cannot run it.

PR #3022, opened 2026-08-26
fix(win): guest sleeps and the audio grain pacer stop resolving on the winpthreads ~15.6 ms tick

mattias800:
Fixes #3013. Fixes #3016.

## Failure scenario

Audio underruns continuously on Windows in **every** title, while Linux is clean. The cause is not in
the audio path: on Windows every guest timed wait resolved on the winpthreads master tick, so a mixer
pacing small grains was woken ~3x too slowly.

## The measurement, and its control

Hand-built probe **outside prosper** (so it tests the primitives, not our wrapper), asking for
**5.33 ms** — one 256-frame grain at 48 kHz:

| primitive | no `timeBeginPeriod` | with `timeBeginPeriod(1)` |
| --- | --- | --- |
| `nanosleep` | **15.67 ms** | 15.59 ms |
| `sem_timedwait` | **15.57 ms** | 15.66 ms |
| `cv.wait_for` | **15.73 ms** | 15.63 ms |
| `sem_timedwait`, **posted every 0.5 ms** | 0.01 ms | 0.00 ms |

That last row is the control and it carries the diagnosis: the **wake** path is fine (0.01 ms), so this
is not scheduling latency — only the **timeout** path is quantized; and the probe's lever demonstrably
moves, so the figure is not a constant.

`timeBeginPeriod(1)` not helping is the counter-intuitive part: raising the process timer resolution
is the obvious remedy and it does nothing here, so the tick is winpthreads' own rather than the OS
timer's.

**prosper does not call `timeBeginPeriod` anywhere** — six prose-only lines across four files, no call, `winmm`
linked nowhere, and `src/host/platform/precise_sleep.cpp:14` states it outright. An earlier version of
this text claimed it did; that came from an uncommitted local diff which was later discarded, and was
wrong. The probe applied the call in-process itself, which is where the second column above comes
from. Since nothing in the tree raises the period, `precise_sleep.cpp:16`'s own figure — `Sleep(1)` =
15.554 ms by default against 1.930 ms after the call — describes a **live lever**, not one already
pulled.

`CREATE_WAITABLE_TIMER_HIGH_RESOLUTION` — which `precise_sleep.cpp` has used since #1765 — serves the
same request in **5.64 ms mean / 5.99 ms worst**, no timer-period change. It had exactly one caller
(`hle_graphics.cpp`); the kernel time path never adopted it.

## Two fixes

[...]

Issue #3016, opened 2026-08-26
Windows: the SDL audio grain pacer sleeps on the winpthreads tick — delivery clumps into bursts every 15.6 ms, so every title underruns while averaging 100% of real time


mattias800:
The SDL audio sink paces the guest's audio thread by sleeping one grain per
`sceAudioOutOutput` call — `frontends/audio_sdl3/audio_sink_sdl3.cpp:190`:

```cpp
if (freq > 0) std::this_thread::sleep_until(target);
```

One 256-frame grain at 48 kHz is **5.33 ms**. On Windows `std::this_thread::sleep_until` resolves on
the winpthreads master tick — measured **15.6 ms** for a 5.33 ms request on this host, and unaffected
by `timeBeginPeriod(1)` (see #3013 for the probe and its control).

## Why the symptom is underruns and not a slow rate

`s.next += dur` accumulates an **absolute** target. When a sleep overshoots by 10 ms, the next two or
three targets are already in the past, so those calls return **instantly**. Delivery therefore clumps:
a burst of ~3 grains, then a ~15.6 ms gap, repeating. The per-second average stays at real time while
the device queue drains inside every gap — continuous small underruns, in every title, Windows only.

**This is why a delivery-rate check does not find it.** Measured on *The Messenger* before the fix:
385,024–389,120 B/s against the 384,000 B/s that f32/2ch/48 kHz needs, i.e. ~100% — healthy by that
metric, broken by ear.

The comment above the sleep already states the required property — *"evenly spaced, never bursty"* —
and cites #1016, where the previous cap-based pacing let the guest burst a device quantum and FMOD's
condvar signals coalesced. The tick quantization reintroduces exactly that shape by a different route.

## Fix

Pace with `prosper::host::sleep_until_steady_ns` (high-resolution waitable timer, already used by
`precise_sleep.cpp`) instead of `std::this_thread::sleep_until`.

Measured on *The Messenger*, `PROSPER_AUDIO_QUEUE_TRACE=1`, per second:

| pacer | handoffs/s | queue min | queue max |
| --- | --- | --- | --- |
| `std::this_thread::sleep_until` | 187-191 | 2048-3328 (wanders) | **11264-11776** (5.5 grains) |
| `sleep_until_steady_ns` | 188 | 2048 (pinned at exactly one grain) | **5888** (2.9 grains) |

187.5 handoffs/s is the correct rate, so both arms deliver it on average.

[...]

PR #3070, opened 2026-08-27
tools(perf): read a PROSPER_AUDIO_QUEUE_TIMELINE log and separate underruns from silence

mattias800:
Refs #3033, #3016. The 1 ms queue-depth sampler recovered in #3022 had no reader, so its output was being read by `grep` and eyeball. This is the reader.

## Why it needs to exist rather than being a grep

The quantity it reports is a **discriminator** — the #3033 pacer A/B rests on it — and both obvious summaries are blind to the defect:

- the one-second delivery average stayed at ~100% of real time while the device queue emptied in every gap. That is exactly how #3016 hid for months.
- a percentage-of-samples figure buries a contiguous dry burst in a large denominator.

So the unit is the **episode**: a maximal run of consecutive dry samples on one port, counted once, duration charged one *measured* sampling interval per sample.

## An empty queue is not an underrun

A port that is OPEN but not being fed reads `queued=0` forever, and that is silence. Ungated, this tool would decide a pacer A/B on whichever arm happened to idle more.

A dry episode is an **underrun** only when all three hold: audio seen before it, audio seen after it, and it is no longer than `--active-window-ms` (200 ms default). Otherwise it is **idle**, and the idle total is printed beside the underrun total rather than subtracted out of sight.

Three things the tests taught me while building that gate:

1. **Per-episode, not per-sample.** A per-sample version of the same rule silently degenerates into a pure duration cap — the samples at the edges of a long silence fail the far-side test and get excluded too — so the bracketing half of the rule stops meaning anything *while still appearing in the code*.
2. **The duration condition is not redundant with bracketing.** A 5 s gap between two songs IS bracketed by audio on both sides.
3. **Mutating each condition away showed `fed_before` had no coverage at all.** Arms 6b and 6c were added for a port dry before its first audio and after its last.

## What it deliberately does not do

It does **not** cross-correlate with the per-arrival `[audio-dbg]` emitter, which would be the richer gate. That emitter logs only the *gap* since the previous arrival, so arrival times must be reconstructed by accumulation and their origin is the first arrival, while the sampler's `t_us` origin is sink construction. **The two clocks cannot be aligned from the log**, and a gate resting on a mis-aligned clock is exactly the kind of instrument this tool exists to avoid. `[audio-dbg]` is parsed for context only — arrival count and median cadence — and the output labels it as such on its own line.

The sampling interval is measured from the timestamps rather than assumed, so a run whose sampler was itself quantized (the #3013 trap the sampler's own comment records) does not get its durations scaled by an interval it never achieved.

## The confound was suspected on real data and did not explain it

This cuts the opposite way from how these notes usually read, so it is worth stating plainly.

The first run analysed — 600 s of *Blasphemous 2* — reported a median queue depth of **0.00 grains** and **53% of samples dry**, which looked exactly like the silence artefact above. With the gate in place, **6 samples out of 588,705** reclassify as idle. The 53% is real underrunning.

(The other half of that first reading was mine, not the tool's: I read the longest episode's `40248 us` as 40 *seconds*. It is 40 ms, which is an underrun, not silence.)

A plausible instrument artefact is a very comfortable explanation for an inconvenient measurement, and this one would have retired a real defect. The gate stays because it is right in principle; the evidence it was built to explain away survived it.

## Verification

Nine mutations, each reddening at least one arm:

| mutation | arms reddened |
| --- | --- |
| no gate at all (every dry sample counts) | 8 |
| drop the duration condition | 5 |
| drop `fed_before` | 2 |
| drop `fed_after` | 2 |
| count idle samples as zero | 4 |
| count dry SAMPLES not episodes | 8 |
| nominal 1 ms instead of the measured interval | 2 |
| merge all ports into one series | 2 |
| STARVED threshold always fires | 2 |
| arrival context left unlabelled | 1 |

Registered as `audio_queue_timeline_report_logic`, because the sibling report tools had the specific failure mode of shipping untested and this one's verdict decides a design question.

```
100% tests passed, 0 tests failed out of 248      # exit 0, was 247
```

Windows, MinGW-w64/UCRT GCC 16.1. `git diff --check` clean.

## Scope

Tools-only: one new script, one new self-test, one `add_test`. No emulator code, no behavioural change, and nothing else reads these files. The `--active-window-ms` default of 200 ms is the one judgement call in the diff and it is the thing most worth arguing about — the arms pin that it can be moved and that moving it changes the answer, so the default is falsifiable rather than load-bearing.

Real output, for reference (the arm that ultimately rejected the queue-depth pacer on #3033):

```
port 17: 144242 samples over 146.5 s  sampling interval 1030 us (median)
  [audio-dbg] context: 12247 arrivals, median gap 11.38 ms  (context only -- not used to gate anything)
  grain 2048 B   queue min 0 B  median 0 B = 0.00 grains  max 4352 B
  IDLE (port open, not being fed): 6 samples, 6.2 ms, 0.00% of samples  [excluded below; max dip 200 ms]
  UNDERRUN (dry while streaming): 8829 episodes, 78784.7 ms total, 53.03% of samples
    longest 37080 us at t=17169983 us
  THIN (under one grain, ungated): 9835 episodes, 91729.7 ms total, 61.74% of samples
  verdict: STARVED -- 8829 underruns
```


mattias800 (commented):
REJECTED — one blocking finding, in the episode segmentation rather than in the gate. Everything I could check about the parser-vs-emitter contract holds, and six of the nine mutation rows reproduce exactly at the arm counts claimed.

## What I verified against the emitter (all good)

* `LINE` (`prosper/tools/perf/audio_queue_timeline_report.py:58-59`) matches the emitter's format string character-for-character (`prosper/frontends/audio_sdl3/audio_sink_sdl3.cpp:374`) — field names, order and types agree.
* **The `queued`/`grain` units claim is correct.** The sampler reads `SDL_GetAudioStreamQueued` (`audio_sink_sdl3.cpp:362`), which counts unconverted bytes in the stream's *input* format, and `grain_bytes = grain * frame_bytes` is guest format (`prosper/src/hle/audio/audio.hpp:29`, assigned at `audio_sink_sdl3.cpp:162`). So grain-relative thresholds need no format flag, as the docstring says, and the `Available` trap the emitter comment records at `:340-342` is genuinely avoided.
* **`queued` can never be negative in the log** (`audio_sink_sdl3.cpp:363`, `if (q < 0) continue;`), so `(\d+)` cannot silently drop a record — the exact hazard the sibling `[audio-dbg]` emitter documents at `:224-233`. Correct by construction here.
* **`grain=0`:** reachable only if the guest opened a port with `grain == 0`; `grain_bytes` is zeroed only together with `stream = nullptr` under the same lock (`audio_sink_sdl3.cpp:443`), so a logged sample always carries a live value. I ran a `grain=0` log through the tool: it prints `grain UNKNOWN (0)` (`:196`), the THIN predicate `g and q < g` (`:200`) is false rather than dividing, and dryness is still measured. Handled — though no arm covers it, see NB3.
* `DBG` (`:65-66`) matches `audio_sink_sdl3.cpp:232` and correctly does *not* match `[audio-dbg-error]` at `:229`.
* **Mutations I re-ran myself**, each on a copy of the branch file against the committed self-test: drop `fed_before` → 2 arms, drop `fed_after` → 2, drop the duration condition → 5, nominal 1 ms instead of the measured interval → 2, STARVED always fires → 2, arrival context unlabelled → 1. All six match your table exactly, and arms 6b/6c are the only coverage of those two conditions.
* The red `macOS (x86_64 / Rosetta 2)` job fails on `48 - kernel_sem_timedwait` and nothing else (99%, 1 of 256). Agreed: inherited from master, not attributable to this diff. The same run shows the new arm registered and green.

## BLOCKING

**`episodes()` treats list-adjacency as time-adjacency, so it welds two dry stretches across a gap in the sampling — and a port close/reopen is exactly such a gap.**

`audio_queue_timeline_report.py:140-153`: a run is terminated only by a non-dry sample (`elif run is not None`, `:148`), never by a jump in `t_us`. The sampler emits nothing at all for a closed port (`audio_sink_sdl3.cpp:360-361`, `if (!slots_[i].stream) continue;`), and a port is normally dry on both sides of its own lifecycle boundary — the queue has drained before `close()`, and the first samples after `open()` precede the first `output()`.

Constructed and run (one port: fed → 20 dry samples → 30 s with no samples at all → 20 dry samples → fed):

```
UNDERRUN (dry while streaming): 1 episodes, 40.0 ms total, 3.92% of samples
  longest 40000 us at t=480000 us
IDLE (port open, not being fed): 0 samples, 0.0 ms, 0.00% of samples
```

Two stretches that arms 6c and 6b respectively define as **shutdown idleness** and **start-up idleness** merge into one bracketed 40 ms **underrun**. So the two conditions the PR body is right to insist are load-bearing hold only at the two boundaries of the log — which is precisely where 6b/6c test them — and are defeated at every interior port lifecycle boundary. The same mechanism fires on lost samples: a 5 s dry stretch covered by only 50 surviving samples reports as one 50 ms underrun with `IDLE 0 samples` (also run). The emitter itself says pass overruns are "expected rather than hypothetical" (`audio_sink_sdl3.cpp:377-383`).

It is silent, and it lands on the headline number: a welded episode is reported as the `longest ... us` line (`:216-217`), which is the figure a reader quotes.

The information needed is already in hand and unused — `end_us` is carried into `classify_dry_episodes` (`:121`) and never compared with `start_us`. Two candidate fixes: split a run when `t_us - prev_t` exceeds a small multiple of `iv`; and/or print a per-port coverage self-check (`n_samples * iv / span`), which turns any such gap into a visible number instead of a silent weld. An arm for the case above (expect 2 episodes, both idle) is a five-line addition to the existing generator.

**Scope, stated because it bounds the blocking claim: this does not overturn the #3033 rejection.** On the run quoted in the PR body, 144,242 samples × 1030 µs = 148.6 s against a 146.5 s span — coverage ≈ 101%, so there is no room in port 17's series for a material gap, and the 53%-of-samples figure is a sample count that this defect cannot inflate at all. It could at most have added a handful of episodes to 8,829. The one number I would re-check against the retained log before quoting it again is `longest 37080 us`: grep the samples either side of `t=17169983 us` for timestamp contiguity. I am blocking on the tool, not on the conclusion.

## Non-blocking

**NB1. The module docstring states a different rule from the one the code implements**, and it is the central rule. `:41-42` says "a dry sample is an UNDERRUN if a non-zero sample occurs within `--active-window-ms` on BOTH sides of it" — a *local* window. The code (`:123-125`) is global bracketing (first/last non-zero anywhere in the port's series) plus a duration cap on the episode. These differ: a 10 ms dry episode preceded by audio five minutes earlier and followed by audio five minutes later is an underrun under the code and not under the docstring. `classify_dry_episodes`' own docstring (`:98-113`) states the implemented rule correctly, so it is the module header that should be brought into line.

**NB2. The verdict line can assert something the same report contradicts.** For a 400 ms dry stretch with continuous audio on both sides (run): `IDLE ... 400 samples, 400.0 ms` and then `verdict: the queue never emptied while the port was being fed` (`:225`). The queue did empty, for 400 ms, with the port fed on both sides. The IDLE line above it is the mitigation, but the verdict is the quotable line — suggest the `not kept` branch name the excluded idle when `n_idle > 0`. Relatedly, the label `IDLE (port open, not being fed)` (`:206`) asserts a *cause* the series cannot establish for an episode over the bound; wording like "dry longer than the max dip — not classifiable from this series" would claim only what is known. On the 200 ms default itself I think you are fine: it is printed in the output (`:209`), arm 6 pins it biting in both directions on one input, and the PR body names it as the judgement call. It is honest as a judgement.

**NB3. Conditions with no arm at all.** I mutated each of these and the self-test stayed green: removing the `g > 0` first-arrival filter (`:91`); replacing `median(deltas)` with a mean (`:171`) — so "measured" is pinned but "median" is not, and a log whose interval genuinely *varies* has no arm anywhere, which is the one interval edge case arm 9 does not reach; `dur_us <=` → `<` (`:125`); the `grain UNKNOWN (0)` branch (`:196`); `median()`'s upper-vs-lower middle (`:158`); applying `--min-episode-us` to THIN as well (`:200` — that THIN is deliberately unfiltered is also undocumented). Also unarmed: `interval_us` returning 0 for a single-sample port (`:168-169`) — I ran it, and a port with one dry sample reports `0.0 ms` everywhere plus `verdict: the queue never emptied`, exit 0. None of these change today's answer; the median one is the one I would arm, since the whole #3013 argument rests on the choice of estimator.

**NB4. The documented invocation does not work.** `:11` documents `[--min-episode-us N]` (space-separated); `main()` accepts only `--min-episode-us=N` and silently discards anything else beginning with `--` (`:234-241`). `--active-window-ms` is absent from the usage line entirely. The window at least self-reveals in the output; `--min-episode-us` does not — its "(of N; M shorter than …)" annotation appears only when the flag parsed (`:211-212`), so a typo yields the ungated report with no indication at all. The convention for this is spelled out three files away, in the emitter that produces this tool's input (`audio_sink_sdl3.cpp:41-49`: a typo must cost you the measurement, never fire one you did not ask for). Erroring on an unrecognized `--` argument would match it.

**NB5. `:5` over-claims audibility:** "A device queue that empties for 3 ms and refills is an audible underrun." `SDL_GetAudioStreamQueued` counts what the device has *not yet pulled*, so a brief zero can be covered by data already in flight. The instrument establishes an empty queue, not an audible dropout — and the rest of the file is careful about exactly this kind of distinction, so this sentence reads as out of step with it.

**NB6. `:29` mixes denominators.** "6 samples out of 588,705 reclassify as idle, so the 53% is real underrunning" — the 53% is port 17's, whose denominator is 144,242 per your own sample output, while 588,705 is the whole log. Name the port. The hedging around it is otherwise right, and `:31-35` is the best paragraph in the file.

**NB7.** `[audio-dbg-error]` lines (`audio_sink_sdl3.cpp:229`) are not matched by `DBG` (`:65-66`), so the context line's arrival *count* undercounts on an errored stream. Context only, so cosmetic — but it is the same shape as the hazard the emitter comment at `:224-229` exists to record, so a one-line note (or matching both spellings for the count) would be consistent with it.

Happy to re-review on the corrected head; the blocking item is narrow and everything else is yours to take or leave.

mattias800 (commented):
Re-review on head `7830494c`.

**REJECTED** — the fix to my previous blocking finding is correct and I verified it directly, but the arms added for the argument-parsing fix **cannot fail the test**: they print `[FAIL]` and the self-test still exits 0. That is a three-line fix in the test file only; the tool code I would take as-is.

## The previous blocking finding is fixed, verified by running it

My exact construction (fed → 20 dry → 30 s with no samples → 20 dry → fed) on this head:

```
port 17: 1020 samples over 30.5 s  sampling interval 1000 us (median)  coverage 3%
  NOT COUNTED (idle, or unobservable): 40 samples, 40.0 ms, 3.92% of samples
  UNDERRUN (dry while streaming): 0 episodes, 0.0 ms total, 0.00% of samples
  verdict: no underrun -- every dry stretch was idle, over the dip bound,
           or across a gap in sampling (see NOT COUNTED above)
```

And the quieter lossy variant (5 s dry covered by 50 surviving samples) now reads `NOT COUNTED: 50 samples`, `UNDERRUN: 0 episodes`, `coverage 29%` — it was `1 episodes, 50.0 ms, IDLE 0`. Both halves of the weld are gone, and the coverage line makes the hole a number rather than an absence.

**Mutations I ran myself against the new logic** (each on a copy of the branch file, against the committed self-test): drop the `abuts_gap` condition → 2 arms; ending run not flagged (`:170`) → 2; starting run not flagged (`:175`) → 2; `gap_us = None`, i.e. the original defect → 2; split on any irregularity (`gap_us = 0`) → 11; coverage pinned at 100 → 1; verdict idle-branch never taken → 1; space form ignored again → 2; unknown option accepted → 1. Both sides of the threshold are genuinely pinned, and the asymmetry I would have worried about — a run that *starts* at a gap versus one that *ends* at one — is covered separately, which is right, because the two lines are separate code.

Two things I probed beyond your arms and am satisfied with: the threshold tolerates up to three consecutive lost samples before splitting (`>` at `:167-168`), which is the right side of the line; and `episodes()` flags a new run started at a discontinuity even when no run was open, which is the reopened-port case.

## BLOCKING

**Three exit-code arms print `[FAIL]` without failing the test, two of them added this cycle for the very fix they cover.** `prosper/tools/perf/test_audio_queue_timeline_report.py:194` and `:197` — and pre-existing `:63` — print a `[FAIL]` line but do not touch the `fails` counter, unlike `expect`/`expect_not` (`:36-38`, `:44-46`) and unlike `:260`, which correctly follows its print with `return 1`.

Measured. I removed strict argument parsing from the tool entirely — unknown option accepted and coerced, non-integer coerced to 0 — and ran the committed self-test:

```
 [FAIL] args: an unknown option exits 0, so a typo silently runs the default arm
 [FAIL] args: a non-integer value exits 0
audio_queue_timeline_report self-test: all cases pass
EXIT CODE = 0
```

ctest reads the exit code and nothing else, so `audio_queue_timeline_report_logic` is **green with both properties broken**, while the transcript says FAILURES and the summary line says "all cases pass" in the same output. `:63` has the same shape: if the tool started exiting non-zero on a healthy log, the arm would print and pass.

This matters more than its size. The registration comment in `CMakeLists.txt` justifies this test on the ground that the sibling report tools shipped untested and that this one's verdict decides a design question — and the charter's rule is that the exit code is truth, quoted alongside the count. An arm that cannot move the exit code is not registered coverage; it is a comment that happens to run. It also means your own mutation matrix reported those two arms reddening on the strength of the printed line rather than of a failing test, so the harness improvement you made this cycle (printing when an anchor is missing) has a twin: the arms must also be unable to pass while printing FAIL.

Fix: `fails += 1` before each of the three prints (with `global fails` in `main`, or route them through a small `expect_rc(rc, expected, what)` helper so the pattern cannot recur). Nothing else needed from my side — re-push and I will look at exactly those lines.

## Your question: which unarmed conditions I would actually add

Two of the seven, and one of them has gone *up* in value because of this fix:

1. **`median` versus a mean in `interval_us` (`:202`) — add this one.** When I first raised it, `iv` only scaled durations. It now also sets the segmentation threshold (`gap_us = iv * 4`, `:240`), so a regression from median to mean does not merely misreport milliseconds, it changes how many episodes exist — the quantity the A/B compares. A log with a handful of large outlier deltas (a few real hiccups among thousands of on-grid samples) separates the two estimators by enough to cross 4×, and no existing arm has a non-uniform interval anywhere.
2. **The `grain UNKNOWN (0)` branch (`:235`) — cheap, and it guards a division** (`median(qs) / grain`, `:233`). One five-line arm on a `grain=0` log; the kind of guard a refactor removes without noticing. I confirmed by hand that it works today.

The other five I would **not** add, and I think your judgement was right: the `g > 0` arrival filter (context only), the `<=` dip boundary, `median()`'s middle element, THIN versus `--min-episode-us`, and `interval_us` → 0 on a single-sample port. None of them can change a reported verdict on a real log, and "arms for every branch" is how a suite becomes something people stop reading. Naming the two above is the whole of my answer — the `fed_before` hole was worth finding because that condition was load-bearing; these two are the only ones on the list that now carry weight.

## Non-blocking

**NB1. `abuts_gap` is deflationary by design, and that deserves one line in the docstring.** I ran a genuine dry dip immediately after a 20 ms sampling hole (fed across the hole): `UNDERRUN: 0 episodes`, `NOT COUNTED: 10 samples`, `coverage 97%`. I agree with the conservative call and the reasoning at `:115-118` covers *why*, but not that the trade has a direction — a lossy arm of an A/B now reports fewer underruns than a clean one, and the coverage line is what makes that detectable. Worth saying explicitly so the next reader does not rediscover it as a defect: the exclusion is conservative, and coverage is the number that tells you it fired.

**NB2. Coverage is off by one interval** (`:222`): `len(samples) * iv` against a span that contains `len(samples) - 1` intervals. A gapless 2000-sample log rounds to `100%` so it does not show, but a 2-sample port reads `coverage 200%` (ran it), which the comment at `:220-221` absorbs as "ordinary jitter" when it is structural. `(len(samples) - 1) * iv` is exact.

**NB3.** `--active-window-ms=0` is accepted and silently disables underrun detection entirely. It does echo `max dip 0 ms` in the output, so it self-reveals in the way the other flags now do — noting it only because `n < 0` is rejected and 0 is the value that makes the tool answer "no underrun" unconditionally.

**NB4.** The module docstring correction at `:42-50` is exactly what I asked for, and naming the discrepancy in place rather than quietly editing it is the better half of the fix. One residual: `:5` still says an empty queue for 3 ms "is an audible underrun", which the instrument cannot establish (`SDL_GetAudioStreamQueued` excludes what the device has already pulled). Unchanged from last round and still non-blocking.

CI on this head is still running (`Linux`, `Windows MinGW`, `Windows App`, both macOS jobs pending as I write). Nothing in my read of the diff predicts a failure, and the previous head's green plus #3066 covers the Rosetta flake — but the verdict above is about content and does not stand in for the CI wait.

mattias800:
## Review response — the blocking finding is fixed at `82d8ba98`

**Three arms printed `[FAIL]` and could not fail.** You measured it and you were right: with strict argument parsing removed from the tool, the committed self-test printed both FAIL lines and then `all cases pass`, exit 0. ctest reads only that.

Fixed with an `expect_rc` helper beside `expect`/`expect_not` rather than three `fails += 1` lines, so the next exit-code arm cannot reintroduce it. Four arms converted — the two argument arms from last cycle, the empty-log arm, and the healthy-log arm. There are now **zero** uncounted `print(" [FAIL]"` sites in the file.

Verified with your exact scenario: the same mutation now prints both failures **and exits 1**.

That this landed one cycle after fixing a *different* could-not-fail defect in the same file is why it is a helper rather than four edits.

## The two unarmed conditions, added

- **median vs mean in `interval_us`** — you were right that its stakes went up with the discontinuity fix, since `iv` now sets `gap_us = iv * 4` and a mean changes the episode *count*. The arm feeds 800 samples at 1 ms around one 300 ms stall, where the mean is dragged to ~1374 µs.
- **the `grain UNKNOWN (0)` branch** — three assertions: reported as unknown, still produces a report, and dryness still classified.

The other five stay unarmed, per your own judgement that these two were the ones with plausible regressions.

## Your non-blocking items

- **Coverage was off by one interval** — n samples span (n-1). Fixed; a two-sample port read 200% and now reads 100%. Worth noting the real #3033 logs read 101% *both before and after*, because at 144,000 samples one interval is 0.0007% — so those runs are not evidence the fix works, and the arm is. I said so in the commit rather than quoting them as validation.
- **`--active-window-ms=0`** silently disabled underrun detection while echoing `max dip 0 ms` as though it had done something. Refused now, with an arm.
- **`abuts_gap` is deflationary** — one line saying so, including that a real dip after a lost-sample hole is not counted either, and that the coverage figure is what tells a reader the window was not fully observed. Under-reporting a defect we cannot substantiate beats attributing unobserved time to one.
- **`:5` over-claiming audibility** — not changed. `queued` excluding what the device already pulled is a fair objection, but the sentence is about what the metric is *for* rather than a claim about the speaker, and I could not find a rewording that was more accurate without being longer than the point deserves. Say if you want it and I will take a wording from you.

## Mutations

Four new, each reddening at least one arm **and** exiting non-zero:

| mutation | arms | exit |
| --- | --- | --- |
| mean instead of median interval | 2 | 1 |
| grain 0 divides anyway | 3 | 1 |
| coverage counts samples not intervals | 1 | 1 |
| window 0 silently accepted | 1 | 1 |

The exit column is there because of your finding: an arm count without it is exactly what hid this.

**The #3033 verdict is unchanged across all three revisions of this tool** — 52.86% against 69.02%, coverage 101% on both arms.

mattias800 (commented):
Re-review on head `82d8ba98`.

**APPROVED.**

The blocking finding is fixed and I verified it the way it was found — by measurement, not by reading. My scenario (strict argument parsing stripped from the tool entirely) now prints the two failures **and exits 1**, where before it printed them and reported "all cases pass" with exit 0.

## What I measured on this head

Nine mutations, each run against the committed self-test, with the **suite exit code** recorded alongside the arm count:

| mutation | arms | exit |
| --- | --- | --- |
| strict argument parsing removed (my original scenario) | 1 | 1 |
| tool raises immediately | 51 | 1 |
| tool exits 0 printing nothing | 53 | 1 |
| coverage counts samples again (`len(samples)`) | 1 | 1 |
| `median` → mean in `interval_us` | 2 | 1 |
| `grain UNKNOWN (0)` branch removed | 1 | 1 |
| zero active window accepted again | 1 | 1 |
| non-integer coerced to a **valid** default (not 0) | 1 | 1 |
| `--active-window-ms` boundary and the earlier nine from the last two rounds | — | 1 |

In every case `arms > 0` implied `exit 1`. That last row of the argument block is worth calling out: I deliberately coerced a non-integer to `200` rather than to `0`, because a coercion to 0 would now be caught by the new zero-window refusal instead — i.e. I checked that arm has no alternate path to its expected result, and it does not.

The tool diff for this cycle is 17 lines across three places, all of them what I asked for, and the coverage comment now states the `(n-1)` reason in place. `diff --check` clean, no `Claude-Session:` trailers on the branch.

## Your question 1: any other assertion shape that can print without counting

I looked for this by measurement and by structure, and found none. Four separate properties now hold:

* **Every `[FAIL]` print in the file is inside a counting helper** — `expect` (`:35-40`), `expect_rc` (`:43-55`), `expect_not` (`:58-63`). There are no bare prints left, which your grep found; what your grep could not tell you is the next one.
* **No early `return` in `main` before the tally.** This is the shape I went looking for after the first one, because it is the *other* way an arm silently does not run: the old empty-log arm ended in `return 1`, so an arm placed after it would have been unreachable on that path. It is now an `expect_rc` and the only returns left are the two at `:310`/`:312`. Anything appended to this file now actually executes.
* **The suite fails when the tool disappears out from under it** — a crash and a silent exit-0 both redden ~50 arms and exit 1. A suite can be written that passes when its subject stops producing output at all; this one cannot.
* **One residual shape, not a defect today, worth a rule:** an `expect_not` assertion is vacuously satisfied by empty output, so an arm whose *only* assertion is an `expect_not` would pass on a crashed tool. No such arm exists — every `expect_not` (the confound arm, the STARVED-threshold arm) sits beside a positive `expect`. If you want one line of insurance for the next author: never let `expect_not` be an arm's only assertion.

Separately and trivially: `subprocess.run` at `:28` has no `timeout`, so a mutation that made the tool hang would hang the suite. ctest reports that as a timeout failure rather than a pass, so it is fail-visible, just slow — `timeout=60` if you ever want it to be quick about it.

## Your question 2: is the two-sample coverage arm enough

Yes, and the reason is falsifiable rather than a matter of taste. The error was exactly **one interval**, so the relative error is `1/(n-1)` — monotonically decreasing in `n`. The two-sample case is therefore the maximum-leverage case in the whole family (100% against 200%), and any mid-length arm is strictly weaker: at `n = 1000` the discrepancy is 0.1%, which the `%.0f` in the header rounds away, so such an arm would pass with the bug in place and would be an arm in name only.

I confirmed the shape of that: reverting the fix reddens **exactly one** arm, the two-sample one. Which also tells you something about `:177` (`coverage reads ~100% on a gapless run`) — it passed both before and after the fix, so it is an *invariant* check, not a regression guard for this change. That pairing is the right coverage: one arm where the error is maximal, one that pins the invariant. A mid-length case adds nothing measurable, and I would not add it.

Your choice not to quote the #3033 logs as validation is the right call for the same reason — at 144,000 samples the fix cannot move the figure, so those runs are consistent with the bug and with the fix equally, and citing them would have been evidence that could not discriminate.

## The `:5` wording, since you asked for it rather than an argument

Take this:

```
percentage. A device queue that empties for 3 ms and refills is the condition an underrun is made
of, and it is invisible to both of the obvious summaries: the one-second delivery average stays at
```

One clause changed, same length, and it drops the claim the instrument cannot make while keeping the point: the tool measures the condition, and audibility depends on what the device had already pulled — which is exactly the thing `queued` excludes. If you prefer to keep a causal ring to it, `has nothing left for the device to pull` also works and is still true. Either is fine by me; this is not worth another cycle.

## Non-blocking, take or leave

**NB1.** Arm 6h pins the printed *interval* (median 1000 µs against a mean of ~1376), which is what I asked for and it reddens. Its comment claims a mean "would resegment the series", and that consequence is not exercised — the fixture has no dry samples, so the episode count is 0 either way. If you want the stronger version, the discriminating fixture is a dry stretch containing one ~5 ms hole: 5000 µs is above the median threshold (4×1000) and below the mean's (4×1376), so median and mean give different episode counts. Optional; the estimator itself is pinned either way.

**NB2.** `--active-window-ms 0` being refused is a better answer than echoing it, and the message says why. Note it makes `0` the one value where the two flags differ in kind (`--min-episode-us 0` is the default and legal) — the error text covers that, so no action, just naming it so it does not read later as an inconsistency.

Three cycles, and the tool that comes out of it is a better instrument than the one that went in: the welding defect, the four uncounted arms, and the coverage line together were the difference between a discriminator you can quote and one that would eventually have decided something wrongly. Nothing outstanding from me.

CI on this head: `Docs`, `Contribution shape`, `Linux App`, `Linux App Package` and `Windows MinGW` green; `Linux`, `Windows App` and both macOS jobs still running as I write. The verdict above is about content and does not stand in for that wait.

Issue #1729, opened 2026-08-02
tools: allocate instrument-list numbers against master AND open PRs — four collisions in one day, each caught only at merge


mattias800:
## The cost

On 2026-08-01/02 the instrument-trap table in `docs/GAME_COMPAT_ORCHESTRATION.md` went from 39 to 44
entries, and **four separate number collisions** occurred between concurrently open PRs:

| collision | claimants |
|---|---|
| 32 | #1696 vs master (32-35 landed while #1696 was open) |
| 38 | #1698 vs #1699 |
| 42 | #1719 vs #1722 |
| 43 | #1722 vs #1726 |

Every one was resolved correctly, and #1701's `Docs` CI job now catches duplicates before merge rather
than after. But the check fires at the *end* of the process. Each collision still cost a renumber, a
cross-reference sweep, a force-push and a full CI re-run — and in two cases the branch had to be
resolved twice because the table moved again while the first resolution was in flight.

## Why it keeps happening

The maintenance rule is **"append, never renumber"**, and it is correct: an existing number may already
be cited from an issue or a commit message, so numbers must be stable *once assigned*. The problem is
not the numbering scheme, it is **allocation under concurrency** — an author reads the highest row at
the moment they write the entry, and several hours of review later that number belongs to someone else.

Checking `origin/master` is not sufficient, because the competing claim usually lives in an open PR
that has not merged yet. That is exactly the case the `Docs` gate cannot pre-empt.

## Proposed

A small helper — `tools/docs/next_trap_number.sh` or a flag on `check_numbered_table.py` — that reports
the next free number by scanning **master plus every open PR's version of the file**:

```
gh pr list --state open --json number --jq '.[].number'   # then per PR:
gh pr diff <n> | grep -oE '^\+\| [0-9]+ \|'
```

Output should name the competing claims, not just the number, so an author can see *who* holds 43 and
decide rather than guess. All four collisions above would have been prevented by one invocation at
write time, and it costs a couple of API calls.

Worth pairing with a line in the table's own maintenance note telling authors to run it, and to re-run
it immediately before merging — the number can still move during review, which is why the `Docs` gate
stays as the backstop rather than being replaced.

## Not proposed

Changing the identifier scheme. Numbers are cited by number from issues, commit messages and other docs;
a date- or hash-based id would break "if you need the count, read the last row's number" and every
existing citation. The allocation is the problem, not the identifier.

mattias800:
[...]

PR #3071, opened 2026-08-27
fix(tools): trap_number.py was unusable on Windows — text=True decoded the doc as cp1252

mattias800:
Running `python prosper/tools/docs/trap_number.py` on a Windows host, before this change:

```
AttributeError: 'NoneType' object has no attribute 'split'
  trap_number.py:115 in table_numbers -> for line in text.split(...)
```

**The traceback points at the table parser. The bug is in the read.**

## Mechanism

`subprocess.run(..., text=True)` with no `encoding=` decodes with the **locale default**, which on this host is cp1252 — and this tool's entire job is to pipe a UTF-8 Markdown document through a subprocess. `GAME_COMPAT_ORCHESTRATION.md` contains em dashes and typographic quotes, so the decode raises `UnicodeDecodeError` inside subprocess's **reader thread**:

```
Exception in thread Thread-5 (_readerthread):
  UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 86713
```

That thread's exception does not propagate. `proc.stdout` is left as `None`, `returncode` stays `0`, so `run()` returns `None` and the crash surfaces 150 lines away in unrelated code.

Net effect: **the tool the charter tells you to run before writing an instrument-trap number could not be run on Windows at all**, and when it failed it sent the reader to the wrong file.

## The fix has two parts, and the second is the one that lasts

1. `encoding="utf-8"` on both subprocess calls — the base-document read and the `gh pr diff` read. A PR patch is as likely to carry non-ASCII as the document itself.
2. A `proc.stdout is None` guard raising `ScanError`. Any *future* decode failure now fails closed with a message — which is what the function's docstring already promised:

   > Run `cmd`, or raise ScanError. **Never lets an exception escape as a traceback.**
   > […] A traceback is not a refusal to answer, it is a crash that happens to also not answer, and the two read very differently to whoever is holding a number they are about to write.

   That reasoning was already in the file. The code just did not honour it on this platform.

## It found a real collision on its first working run

```
prosper/docs/GAME_COMPAT_ORCHESTRATION.md  (table header contains 'Instrument')
  origin/master                highest row 229
  PR #3060                      highest row 230   CLAIMS 230
  scanned 8 open PR(s), 1 touching this file (--limit 60)

next free number: 231
```

Reading master alone gives **230**, which is already #3060's. That is precisely the collision this tool exists to prevent (#1729), and it was one command away from happening to me.

## Test

`run()` is exercised through `git show` of a real blob containing an em dash and typographic quotes — the **same call the tool uses to read the base document**, rather than a synthetic `echo`, so the arm covers the path that actually broke.

Mutation-checked: setting `encoding="cp1252"` reddens it, and reddens it *through the new guard* —

```
FAIL: run() raised on non-ASCII UTF-8 output: ScanError('git -C ... produced output this
tool could not decode as UTF-8, so the table cannot be read. This tool refuses to guess a
number from a partial read.')
```

— so the failure is now a legible refusal rather than a crash at a distance. Both halves of the fix are covered by the one arm.

## Scope, and what is deliberately left

There are **81 other `text=True` sites without an explicit encoding** under `prosper/tools/`. I checked the ones that read these same documents: `check_numbered_table.py` and `check_merge_result.py` open files directly with `encoding="utf-8"` and are unaffected — both run clean on this host:

```
$ python prosper/tools/docs/check_numbered_table.py --ordered --table-header Instrument \
...: 14 table(s), 291 rows, unbroken, every row matching its header
...: numbered column 1..229, 229 rows, unique and ascending, no unused numbers

$ python prosper/tools/docs/check_merge_result.py
merging HEAD (8a2f7d51) into origin/master (6f6ec58a)
nothing to check: this head is already contained in the base.
```

The rest are a **latent class on Windows, not a present failure**, and sweeping them is its own change with its own review.

Verified on Windows (MinGW-w64/UCRT GCC 16.1): `ctest --no-tests=error -R trap_number` passes, 1 test, exit 0.

Tools-only; no emulator code, and no numbered docs table is touched by this diff.


mattias800 (commented):
**REJECTED** — two blocking findings, both cheap, and neither touches the diagnosis. The mechanism
is right, the core fix is right, and I want this landed. What blocks is that one of the two call
sites this diff *changes* is still exposed to the identical class, and the "scope" paragraph's
verification claim is half wrong in the direction that matters.

---

## The mechanism: verified, not accepted on the description

I did not take this one on trust, because as the brief notes it is easy to state and hard to check.
Both halves check out.

**CPython source** (`subprocess.py`, 3.12.10, this host):

* `:1598-1600` — `_readerthread` is `buffer.append(fh.read())`. In text mode `fh` is a
  `TextIOWrapper`, so a strict decode failure raises out of `fh.read()` and the `append` never
  happens. A thread exception goes to `threading.excepthook`, so it prints and does not propagate.
* `:1648` — `stdout = stdout[0] if stdout else None`. The buffer is the empty list, which is falsy,
  so `communicate()` hands back **`None`**, and `run()` then reports `process.poll()` — the child's
  own status, unaffected.

**Reproduced**, on the real input:

```
locale.getencoding() = cp1252
git show origin/master:prosper/docs/GAME_COMPAT_ORCHESTRATION.md
  text=True, no encoding= :  rc=0   stdout is None -> True
  text=True, encoding=utf-8: rc=0   len(stdout) = 474193
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 86713
```

Byte offset 86713 is the same one in the PR body. `rc=0` with `stdout is None` — exactly as claimed.

**Q3, whether the guard catches the class it names: yes, exactly, and I checked the partial-read
worry specifically.** It cannot be a partial string. `_readerthread` makes a *single* `fh.read()`
call, so the buffer gets the whole decoded document or nothing at all; there is no interleaving that
could append a prefix. And the guard does not false-positive on legitimately empty output: a command
that prints nothing appends `""`, so the buffer is `[""]` — a truthy *list* — and `:1648` yields
`""` rather than `None`. Verified both:

```
[empty output] rc=0 stdout='' is_none=False
[undecodable]  rc=0 stdout=None is_none=True
```

**Q2, strict vs `errors="replace"`: strict is correct, and I would argue harder for it than the PR
does.** `replace` is not merely "silently substituting characters" here — it is silently
substituting them *in the input to a maximum*. If a mangled byte landed in the `Instrument` header
cell or in a row's number cell, `table_numbers` would quietly return a **shorter** list and
`highest` a **lower** maximum, and the tool would hand out a number another lane already holds while
looking like a successful run. That is precisely the defect the file's own docstring
(`trap_number.py:4-13`) exists to prevent, so `replace`'s failure mode is the one this tool must not
have. Strict also keeps it consistent with the gate that will judge the row:
`check_numbered_table.py:434` and `:898` both `read_text(encoding="utf-8")`, strict. A legitimately
non-UTF-8 document would already be rejected there, so refusing here is agreement rather than extra
strictness.

---

## Blocking

**B1 — `added_rows_from_diff` is the second site this diff changes, and it has no guard.
`prosper/tools/docs/trap_number.py:209-218`.**

The PR's own framing is that part 2 is the half that lasts: *"Any future decode failure now fails
closed with a message."* That is true of `run()` and false of this function, which does its own
`subprocess.run` and is not routed through `run()`. On a decode failure `proc.stdout` is `None`,
`:218` calls `added_rows_from_patch(None, path)`, and `:178` does `patch.split("\n")` —
`AttributeError: 'NoneType' object has no attribute 'split'`. That is the **same crash, in the same
shape, one function over**, and this function's docstring at `:197-203` is specifically about not
silently degrading on an unreadable diff. Adding `encoding="utf-8"` here without the guard closes
the present failure and leaves the class open in the one place the PR says it closed it.

The fix is the three lines already written at `:116-120`. Same lines, related: `:213`
`proc.stderr.strip()` — see N2.

**B2 — `check_merge_result.py` is a PRESENT failure on this host, not "unaffected", and the control
that cleared it could not have caught it. `prosper/tools/docs/check_merge_result.py:51`.**

The scope paragraph says *"`check_numbered_table.py` and `check_merge_result.py` open files directly
with `encoding="utf-8"` and are unaffected."* Half right. `check_numbered_table.py` does read
directly (`:434`, `:898`) and is genuinely fine. `check_merge_result.py` does **not** read the
document directly — `:51` pipes it through `subprocess.run(["git", *args], capture_output=True,
text=True)` with no encoding, and `:114`/`:115` feed that output into `write_text(...)`. Run against
this PR's own head:

```
$ python prosper/tools/docs/check_merge_result.py --no-fetch \
      --head origin/fix/trap-number-windows-encoding
merging origin/fix/trap-number-windows-encoding (b271e43c) into origin/master (6f6ec58a)
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 86713
  File "prosper/tools/docs/check_merge_result.py", line 114, in main
    merged_copy.write_text(show.stdout, encoding="utf-8")
TypeError: data must be str, not NoneType
```

Same byte, same offset, same host. So the tool the charter tells you to run **last, immediately
before merging any numbered-table PR** is dead on Windows for the same reason — while this PR's body
reports it as clean.

**Why it read clean is worth naming, because it is the shape the charter's §122 warns about.** The
PR's evidence is `nothing to check: this head is already contained in the base.` — that is the
`--is-ancestor` early return at `:77`, which returns **37 lines before** the affected code. The
control exercised a path that could not fail. And it is reinforced downstream:
`test_check_merge_result.py` contains **zero non-ASCII bytes** (checked), so its fixtures cannot
express the case at all and the Windows CI job is green on a document that structurally could not
trip it.

Either fix it here (`encoding="utf-8"` at `:51` and `:120`, plus a `None` check in `git()`, same
shape as `run()`'s), or correct the body to say it is affected and file the issue. I do not mind
which; what should not ship is the body's sentence, because per the charter a checked-and-cited
claim is the kind nobody re-derives.

---

## Non-blocking

**N1 — the guard is Windows-only, and the comment reads as though it were general.**
`trap_number.py:110-115` says *"ANY future decode failure in that reader thread yields None here."*
True on Windows. On POSIX there is no reader thread for this: `subprocess.py:2149-2159` decodes in
the **caller's** thread via `_translate_newlines`, so a `UnicodeDecodeError` escapes `run()` as a
traceback and the docstring's *"Never lets an exception escape as a traceback"* (`:75`) is still
unmet on Linux and macOS. One-line close: add `UnicodeDecodeError` to the `except OSError` at
`:106`.

**N2 — `stderr` is not covered, and this change makes it slightly more exposed, not less.**
`:109` `proc.stderr.strip()`, and `:213` likewise. `proc.stderr` decodes with the same strict UTF-8
and can be `None` by the identical mechanism, in which case a *failing* command produces
`AttributeError` instead of the `ScanError` the message was written for. Before the change stderr
was decoded cp1252, which nearly always succeeds on a Windows-codepage message; after it, a git
error carrying a system-codepage filename fails. `(proc.stderr or "").strip()` at both sites.

**N3 — the test arm is a silent no-op on any UTF-8-locale host.** `test_trap_number.py:568-585`.
Q4's answer: yes, it reddens with the fix reverted **on this host** (cp1252 → `run()` returns `None`
→ `case(..., None, NON_ASCII)` fails), and yes, driving it through `git show` genuinely exercises
the same `run()` shape the tool uses at `:270` — that part is achieved. But on Linux/macOS CI
`locale.getencoding()` is already UTF-8, so `text=True` alone decodes the em dash fine and the arm
passes reverted. The regression guard is therefore the Windows job alone, which is worth saying out
loud in the comment. If you want it locale-independent, spawn the call in a child with `LC_ALL=C` /
`PYTHONUTF8=0` and assert on the child's result — under `LC_ALL=C` Python reports
`ansi_x3.4-1968`, so the reverted code fails there too.

**N4 — the arm omits `commit.gpgsign false`, which its own sibling fixture learned the hard way.**
`test_trap_number.py:570-576` configures only `user.email` and `user.name`. `run_main` at `:317-321`
sets `commit.gpgsign false` with a comment saying exactly why: *"A developer with commit.gpgsign on
globally would otherwise get a fixture whose commits fail."* Here the commit is unchecked, so on
such a host `git show HEAD:doc.md` returns non-zero, `run()` raises `ScanError`, and the arm reports
`run() raised on non-ASCII UTF-8 output: ScanError(...)` — a false red blaming the decode. One line
in the existing tuple.

**N5 — the broad `except Exception` at `:584` is acceptable but mislabels.** It cannot manufacture a
false green (it appends to `FAILURES`), so I am not asking for it to narrow. It does attribute every
setup failure to the decode, which is what makes N4 read as a bug in the fix. Naming the expected
types, or moving the setup outside the `try`, would fix N4's message too.

**N6 — placement.** The arm lands under the `print("agreement with the gate, on the repository's
real table:")` heading at `:558`, which is about something else. Cosmetic; move the heading below it
or give the arm its own.

---

## Verified as asked

* **Q6, the collision claim: confirmed.** `origin/master`'s highest `Instrument` row is **229**, and
  `gh pr diff 3060` (open, *"evidence: retract 'Tomb Raider's world is textured'"*) adds
  `+| 230 | **A screenshot of a game's own PRE-RENDERED LOADING PICTURE...`. So master+1 = 230 is
  taken and **231** is the next free number. The tool's first working run was right, and the
  citation holds.
* **Diff scope: confirmed tools-only** — `trap_number.py` and `test_trap_number.py`, nothing else,
  and no numbered table touched.
* **macOS red: agreed, unattributable to this diff.** The only failure is
  `47/255 kernel_sem_timedwait`, a C++ kernel test; every other job including `Windows MinGW` and
  `Docs` is green. A Python-only diff under `tools/docs/` cannot reach it, and #3066 is open against
  exactly that flake.

I did not re-run the author's `ctest -R trap_number` or the `cp1252` mutation, per the review brief.

Re-review on request once B1 and B2 are addressed — B1 is the three lines already in the file, and
B2 can be a body correction plus an issue if you would rather keep the diff narrow.

mattias800 (commented):
**APPROVED** on head `74c394d1`. Both blocking findings are fixed, and I verified each one myself
rather than on the description — including the two things you flagged as needing a second pair of
eyes (the rebuilt merge-result arms, and whether the spy arm actually closes the locale gap). Four
non-blocking notes remain, two of which are the ones you named as not-done.

---

## B1 — closed. `trap_number.py:213-230`

`added_rows_from_diff` now carries the same `proc.stdout is None` refusal, and both message paths
use `(proc.stderr or '')` (`:216`, `:219`). Read at head; the guard sits before `:230`'s
`added_rows_from_patch(proc.stdout, path)`, which is the call that would have died on `:178`
`patch.split`. Correct.

## B2 — closed, and verified end-to-end on the platform in question

`check_merge_result.py:51-73`: `encoding="utf-8"`, the `(proc.stderr or '')` guard, and an
unconditional `stdout is None` refusal that fires on `check=False` calls too. Run at head, on a real
non-ancestor head, on this cp1252 host:

```
$ python check_merge_result.py --no-fetch --head 74c394d1
merging 74c394d1 (74c394d1) into origin/master (2515e559)
...(merged): 14 table(s), 292 rows, unbroken, every row matching its header
...(merged): numbered column 1..230, 230 rows, unique and ascending, no unused numbers; none of
   the 230 rows in origin/master:... has been deleted

clean merge, and the merge RESULT passes the gate against origin/master as of now.
rc=0
```

Runs to completion where it previously raised `TypeError: data must be str, not NoneType`. Fixing it
here rather than deferring was the right call, for the reason you give.

## The rebuilt arms — checked as asked, and they are sound

You asked me to check the shape against the existing duplicate arm. `test_check_merge_result.py`
head, run at head: **9 arms, all pass, rc=0.** Both new ones are shaped on the right precedent:
`NON_ASCII_MASTER` takes 4 at the end, `NON_ASCII_LANE` takes 4 a few lines up, exactly like the
`BASE`/`master's row`/`the lane's row` triple at `:112-116` — so the merge is textually clean and the
duplicate exists only in the result.

**The vacuity you described is genuinely repaired, and the arm proves it itself.** The version you
nearly shipped passed on a conflict; this one asserts `duplicate row number 4`, which per
`check_numbered_table.py:704` only the duplicate branch emits, and the conflict path prints
`MERGE CONFLICT` and no such string. So the arm *passing* with `want_rc=1` is positive evidence that
the clean-merge path was taken — it cannot pass through the conflict route. That is the "what else in
this output could satisfy this assertion?" test your own suite header asks for, and it now answers
"nothing".

**Mutation-checked independently** — `encoding="cp1252"` in `git()`:

```
FAIL: a table carrying em dashes and typographic quotes still validates: expected rc=0, got 2
FAIL: a duplicate in a non-ASCII table's merge result is still caught:  expected rc=1, got 2
2 case(s) failed.   (the other 7 arms: ok)
```

Two things worth naming there. The failure arrives as **rc=2 through your new refusal**, not as a
traceback — so the guard is doing the work, not just the encoding. And **only the two new arms
move**: the seven pre-existing arms stay green under the mutation, which is the direct measurement of
the point I raised — the old fixtures could not express the case, so no amount of hardening them
would have found it. That is the charter's §122 shape stated as a number rather than as an argument.

## The spy arm — it does close the gap, and I confirmed it on a host locale that is not this one

`test_trap_number.py:587-607`. My Q4 objection was that the behavioural arm is a no-op wherever the
locale already decodes UTF-8. Rather than take the fix on faith I built that host: `PYTHONUTF8=1`
makes `locale.getencoding()` return `utf-8`, which is exactly the CI Linux/macOS condition. Three
runs against the tree, with `run()` mutated back to bare `text=True`:

| config | behavioural arm | spy arm |
| --- | --- | --- |
| mutated, cp1252 (this host) | **FAIL** — and via `ScanError`, not a traceback | **FAIL** |
| mutated, `PYTHONUTF8=1` | `ok` — passes the defect, as I claimed | **FAIL** |
| head, `PYTHONUTF8=1` | `ok` | `ok` (rc=0, 40 arms) |

Row 2 is the whole point: the behavioural arm demonstrably passes a reverted fix on a UTF-8-locale
host, and the spy arm reddens there anyway. The gap is closed, and the monkeypatch is clean —
`try/finally` restores `subprocess.run`, and `run()` makes exactly one call so `seen_kwargs` is
unambiguous.

(One apparatus note, since it is the kind of thing that gets misreported: my first `test_trap_number`
run came back rc=1. That was mine, not yours — I had copied the files out of the tree, so
`HERE.parent.parent/docs` did not resolve and the real-doc arm failed **closed**, precisely as
`:610` intends. Re-run in the correct layout: rc=0, 40 arms.)

## Your commit-message correction — verified, and it was right to make

`prosper/CMakeLists.txt:269`, `:276`, `:291` register `trap_citation_checker`, `trap_number` and
`doc_merge_result_checker`. `-R trap_number` matches **one** of those by substring, so the original
"4 tests" could not have been what that filter ran. Corrected figure is consistent with the
registered names.

---

## Non-blocking

**N1 — the two you named as not-done: yes to both, since you offered.** Neither is a present failure
and neither blocks.
* `trap_number.py:106` — add `UnicodeDecodeError` to the `except OSError`. On POSIX
  `subprocess.py:2149-2159` decodes in the caller's thread, so the `None` guard never sees it and the
  docstring's *"never lets an exception escape as a traceback"* is still false on two of three
  platforms. One word, and it makes the promise true everywhere.
* `test_trap_number.py:570-576` — `["config", "commit.gpgsign", "false"]`, the lesson `run_main`
  already records at `:317-321`. One line, and it stops a signing host from producing a false red
  that *blames the decode*.

**N2 — `check_merge_result.py` still has one unencoded `subprocess.run`, and it is subtler than the
one you fixed.** The checker invocation (`capture_output=True, text=True`, no encoding, just below
`:115`) works today only because parent and child both default to the *same* locale codec — implicit
agreement rather than correctness. Two things follow, and the second is a trap:
* It is latent, not present, and I checked why: the checker's messages carry `path:line_no` and
  numbers only (`check_numbered_table.py:704`), so no document text crosses that pipe. If a message
  ever echoes row content it becomes live, because the document holds **6 characters cp1252 cannot
  encode** — `U+2192` x41, `U+2248` x11, `U+03C0` x2, `U+2082`, `U+2705`, `U+2212` — and the *child*
  would then raise `UnicodeEncodeError` while printing, so the tool would report "the checker itself
  failed" instead of the finding.
* **Do not "fix" it by adding `encoding="utf-8"` to the parent alone** — the child is a Python
  script that encodes its stdout with the locale codec unless `PYTHONIOENCODING=utf-8` /
  `PYTHONUTF8=1` is in its env, so a one-sided change converts an accidental agreement into an
  actual mismatch. Both sides or neither. A comment recording that would be enough for now.

**N3 — `check=False` callers read `.stderr` directly, where `git()`'s guard cannot reach.**
`merged.stderr.strip()` and `"--write-tree" in merged.stderr` (~`:106-108`), and `show.stderr.strip()`
(~`:113`). `(proc.stderr or '')` protects `git()`'s own message, not these. Same shape as B1 — the
idiom applied at one of two places — and the same remedy.

**N4 — the new `added_rows_from_diff` guard has no arm at all.** The spy covers `run()`; `:213`'s own
`subprocess.run` got both the encoding and the refusal with nothing pointed at either. The `gh` stub
already drives that function through the collision arms, so an em dash in the stub's patch output
would cover the encoding, and a deliberately undecodable byte would cover the guard — a few lines,
and it is the site whose absence of an arm is what let B1 through in the first place.

**N5 — one comment says the opposite of what its arm does, on the most literal reading.**
`test_check_merge_result.py`, over the second new arm: *"a decode fix that silently truncated the
document would make this pass by finding no duplicate."* Truncation drops the lane's row, no
duplicate is found, rc=0 against `want_rc=1` — so it makes the arm **fail**, which is the behaviour
you want and the reason the arm is worth having. I take the intended reading to be "would make a mere
not-crash test pass", but this file's own header says being right about the arms is part of the test,
so it is worth one word of disambiguation.

---

## Confirmed unchanged

* **Scope still tools-only**: `git diff --stat origin/master...74c394d1` is 4 files, all under
  `prosper/tools/docs/`, no numbered table touched. (The `b271e43c..74c394d1` range also shows
  `test_kernel_sem_timedwait.cpp` and `CMakeLists.txt`; that is #3066 arriving via master, which is
  now `2515e559`, not content of this branch.)
* **The mechanism findings from the first review stand** and I did not re-derive them: CPython
  `subprocess.py:1598-1600` / `:1648`, no partial-read path, empty output yields `""` not `None`, and
  strict-over-`replace` is right because `replace` would silently lower a maximum.
* **CI**: `Contribution shape`, `Docs`, `Linux`, `Linux App`, `Linux App Package`, `Windows App`,
  `Windows MinGW` all green at head. The two macOS jobs had not reported when I looked — #3066 is
  merged (`2515e559`'s ancestry) so the `kernel_sem_timedwait` flake should be gone from the base,
  but that is now an ordinary green-CI check rather than an inherited-failure argument. Confirm it
  before merging rather than carrying my earlier "unattributable" note forward.

Nothing blocking. Merge when the macOS jobs report, with N1 in if you want it in.

mattias800:
## Review response — all five non-blocking items at `6ecc52b4`

Thank you for building the `PYTHONUTF8=1` host to test my spy arm rather than reasoning about it. That the *behavioural* arm passes the defect under CI's condition while the spy arm still fails is exactly the gap I was guessing at, and now it is measured.

**1. Both items I had offered, taken.**
- `trap_number.run` now catches `UnicodeDecodeError` alongside `OSError`. You are right that the docstring's "never lets an exception escape as a traceback" was still false off Windows, since the `None` guard only covers the thread-swallowed shape and POSIX decodes in the caller's thread.
- `commit.gpgsign false` in the new arm's throwaway repo.

**2. The checker invocation — left alone, and your warning is the reason.** I would have "fixed" the parent side. The child encodes its stdout with the locale codec unless `PYTHONIOENCODING`/`PYTHONUTF8` is set, so a one-sided change converts an accidental agreement into a real mismatch. It is latent (the checker's messages carry only `path:line_no`) and needs a both-sides change, which is its own diff. Recorded in the commit message so the next reader does not repeat my instinct.

Your character census is the useful part of that finding and I have not seen it elsewhere: the doc holds six characters cp1252 cannot encode — U+2192 ×41, U+2248 ×11, U+03C0 ×2, U+2082, U+2705, U+2212.

**3. The `check=False` stderr reads** — `merged.stderr` and `show.stderr`, where `git()`'s guard cannot reach. Guarded. Same shape as the blocking finding one level down, as you said.

**4. `added_rows_from_diff`'s guard had no arm** — added, driven directly with a fake zero-rc `CompletedProcess` carrying `stdout=None` rather than through the `gh` stub, since what matters is that this shape becomes a `ScanError` and that is locale-independent. It distinguishes the three outcomes: `ScanError` (pass), silent acceptance (fail), and `AttributeError` on `None.split` (fail, with the message naming the missing guard). Verified — removing the guard reddens it with `died on None.split -- the guard is missing`.

**5. The inverted comment.** You are right and it is worth naming plainly: the note said a silently-truncating decode fix "would make this pass by finding no duplicate", which is backwards. Truncation drops the duplicate, so that arm goes **red** while the clean-merge arm stays green — which is why both exist and why that one carries the load. A comment inverting its own arm's logic is worse than none, because it tells the next reader the arm is vacuous when it is the opposite. Corrected, with the inversion recorded.

Also noted: your point that the duplicate arm's *passing* with `want_rc=1` and the full `duplicate row number 4` message is itself proof the clean-merge path was taken — since per `check_numbered_table.py:704` only the duplicate branch emits that string — is a stronger argument for the arm's validity than the one I gave. It is why the earlier vacuous draft could not have passed this way.

Treating the macOS jobs as an ordinary green-CI requirement now that #3066 is in the base, as you said.

mattias800 (commented):
**APPROVED** on head `3ecdbb28`. Merge it.

All five non-blocking items are in and read correctly. The commit I had not seen — `68ef2cb3`, the
`gh`-dependent arm — I reproduced and mutation-checked independently rather than reading the fix, and
your audit holds. Answers to both asks below, plus one residual I would take now and would not hold
the PR for.

## The defect CI found — reproduced, and the repair verified where it counts

I rebuilt both trees and ran them with `PATH` scrubbed of the GitHub CLI directory:

```
PRE-FIX  6ecc52b4, gh absent:  FAIL: added_rows_from_diff refuses an undecodable patch:
                               expected True, got False              exit 1
HEAD     3ecdbb28, gh absent:  ok  added_rows_from_diff refuses an undecodable patch
                               all cases passed                      exit 0
```

Your diagnosis is exact: with `gh` present the pre-fix arm passed here too, so the verdict depended
on a property of the machine. Worth naming that it failed in the **safe** direction — a false red on
the wrong grounds, not a false green — which is why CI caught it and why it cost a commit rather than
a merge.

**The part that matters more than the repro is whether the arm still tests the guard**, and it does.
Removing the `proc.stdout is None` guard from `added_rows_from_diff` at head:

```
gh present:  FAIL: added_rows_from_diff died on None.split -- the guard is missing
gh absent:   FAIL: added_rows_from_diff died on None.split -- the guard is missing
```

Identical, correctly-diagnosed message in both configurations. That invariance under the precondition
is the property the arm lacked, so the fix is the right one and not merely a green one. Head suites:
`test_trap_number` 41 arms rc=0, `test_check_merge_result` 9 arms rc=0.

## Your audit — checked, not taken, and it holds

* `test_trap_number.py:279` does invoke `git` directly in the `git()` helper, and `:358` carries
  `assert real_git, "git must be on PATH for these tests"`. So `git` is an **asserted suite-wide
  precondition**, which is the reason your other two arms are fine — a stronger reason than "they only
  use git", and the one I would put in the comment.
* I also checked the two new arms in `test_check_merge_result.py`, which your audit did not mention:
  `check_merge_result.py` never invokes `gh` at all (its three `gh` matches are substrings of
  *through* / *highest*), so they carry no `gh` precondition either.
* One residual, not worth changing: the spy arm's `run(["git", "--version"])` sits in `try`/**finally**
  with no `except`, so on a git-less host it dies as an unhandled `ScanError` rather than a clean
  `FAIL` line. Same precondition, cosmetically worse failure mode. Noting it so it is not a surprise.

## Ask 1 — keep the `shutil.which` stub. It is the right shape, and the convention does not apply here

Three reasons, in order of weight:

1. **The mutation above is the argument.** The arm reddens identically with and without `gh` on PATH.
   A PATH-fixture version cannot give you that, because the fixture *is* the precondition.
2. **Going through the `gh` stub would make the arm platform-conditional again.** To get `stdout=None`
   out of the real path the stub would have to emit invalid UTF-8 — and the None-with-rc-0 shape is
   **Windows-only** (on POSIX the decode raises in the caller's thread). You would have rebuilt exactly
   the defect N3 removed from the other arm one round ago.
3. **The twelve subprocess arms use the PATH fixture because they drive `main()` end-to-end through
   argv** — a different job, and one where `gh` discovery is legitimately part of the subject. A
   convention adopted for a reason is not a convention that applies where the reason is absent. Your
   comment at `:619-623` already records the deviation and why, which is the part that keeps it honest.

One nit inside it: `shutil.which = lambda name: "/usr/bin/" + name` takes one positional argument
where the real signature is `which(cmd, mode=..., path=...)`. Safe today — the only call in the window
is one-arg, and both stubs are restored in `finally` — but `lambda name, *a, **k:` costs nothing and
survives a future call site.

## Ask 2 — file the issue, do not leave it in the commit message

The charter's rule points one way here: a finding that lives only in a PR body or a commit message is
one the next agent re-derives at full cost. A commit message is also the one place nobody greps. Three
things to put in it so it is actionable without re-deriving:

* the concrete crash site is **`check_merge_result.py:157`** — `"problem(s) found" not in proc.stdout
  + proc.stderr` is a `TypeError` the moment either side is `None`, and `:149`'s `if stream:` silently
  drops the checker's findings just before it;
* the remedy is **both sides or neither** — parent `encoding="utf-8"` *and* `PYTHONIOENCODING=utf-8`
  (or `PYTHONUTF8=1`) in the child's `env=`, because the child encodes its stdout with the locale codec;
* it is latent only because the checker's messages carry `path:line_no` and numbers, never row text
  (`check_numbered_table.py:704`). The document holds **6 characters cp1252 cannot encode** —
  `U+2192` x41, `U+2248` x11, `U+03C0` x2, `U+2082`, `U+2705`, `U+2212` — so the day a message echoes
  a row, the *child* raises `UnicodeEncodeError` and the tool reports "the checker itself failed"
  instead of the finding.

## One residual I would take now — non-blocking, and not a regression

`added_rows_from_diff` has **no `try`/`except` around its `subprocess.run`** — the body is
`subprocess.run`, then the rc check, then the None check, with no wrapper. So the POSIX shape you
closed in `run()` this round is still open one function over. Hand-built probe, constructed outside
the suite that produced the green — `subprocess.run` replaced with one that raises
`UnicodeDecodeError`, which is precisely what CPython does on POSIX:

```
run()                      -> ScanError (fails closed)
added_rows_from_diff()     -> UnicodeDecodeError ESCAPES as a traceback
```

**Why it does not block:** it is not a regression — master's unencoded `text=True` had the identical
exposure under a UTF-8 locale — and it is unreachable on the platform this PR exists to fix. It also
needs a PR in the scan window whose patch carries non-UTF-8 bytes, which for doc-touching PRs is
unlikely.

**Why I would still take it in this PR:** this is the third consecutive round in which the finding has
been *the guard is at one of two sites*, which is a pattern rather than three accidents. One helper
both sites call, or the same `except (OSError, UnicodeDecodeError)` repeated, ends the series. Your
call — if you would rather land and fold it into the Ask-2 issue, that is fine and I will not
re-review for it.

Nit: `test_trap_number.py:647` has a three-blank-line run (PEP 8 wants two). `git diff --check` over
`origin/master...3ecdbb28` is clean, and there is no Python lint gate, so it is cosmetic.

## Confirmed for merge

* **Scope**: `git diff --stat origin/master...3ecdbb28` is 4 files, all `prosper/tools/docs/`, no
  numbered table touched. I read the two-dot range as commits rather than as content, so the four
  master syncs and trap 231 do not change anything I checked.
* **CI: all 10 checks green at head**, including `macOS (x86_64 / Rosetta 2)` — #3066 is in the base,
  so my earlier "inherited flake" caveat is retired rather than carried — and `Windows MinGW`, which
  is the job that found `6ecc52b4`.
* The mechanism findings from round 1 and the arm verifications from round 2 stand; I did not
  re-derive either.

PR #2574, opened 2026-08-17
diag: audit the PROSPER_* switches whose producer and printer are armed separately (#2149)

mattias800:
Refs #2149 · Refs #2146 · Refs #2569 · Refs #2572

## The failure scenario

A diagnostic prints a field whose **producer** is armed by a different switch than the one that arms
the **printer**. When the producer is unarmed the field reads as a confident, well-formed **zero** —
indistinguishable from a measured negative.

The zero is not merely uninformative, it is affirmatively misleading, because a zero in these fields
is usually exactly what the interesting hypothesis predicts. #2111's instrument emitted **183
negatives it had no capacity to falsify**. The cost is not the wasted arm: these zeros land in
`## Ruled out` rows and issue comments as *measurements*, which the charter calls the most expensive
knowledge in the repository. The failure is silent, survives review, and **looks like rigour** — the
output is well formed and the number is specific.

Existing remedy was a discipline (instrument trap 116), and trap 121 records that discipline
decaying: the agent who hit the third instance had read 116.

## What this PR does

**1. The audit — the half nobody had done.** `prosper/tools/env/check_diag_gates.py`, a re-runnable
scanner for three lexical signatures, each derived from a **measured** instance rather than invented:

| signature | meaning | measured instance |
| --- | --- | --- |
| `SPLIT-LOCAL` | a local declared with a default (`= 0`) whose only real writes sit inside an env-gated region, printed **outside** it | `fresh_extent` (#2132) |
| `TWO-GATE` | one report statement requiring **two** distinct `PROSPER_*` variables | `[udtail]` (#2146); the unimplemented-NID table (#2149) |
| `SPLIT-CALL` | a callee reached only under env gates whose call sites **disagree** on the gate | `record_guest_write(ColorTarget, …)` (#2111) |

Gates are modelled as a **conjunction of disjunctions**, not a flat name set: `A && B` needs both,
`A || B` needs either, and a flat reading collapses them. That is not a refinement — the flat
revision reported **133** two-gate findings on master, including one predicate it claimed required
all sixteen of its alternatives.

**2. The findings, landed in the repository** — `tools/env/diag_gate_baseline.txt` (classified) and
`docs/DIAGNOSTIC_GATE_AUDIT.md` (the convention, the numbers, the blind spots, a `## Ruled out`).

**2a. The baseline key, and the two churn doors shut in it.** Several lanes edit this tree at once,
so a key that moves under an *unrelated* edit fails the `Docs` job on a PR that did nothing wrong —
[...]

PR #3077, opened 2026-08-27
evidence: correct the ceiling-removal figures — the probe behind them saw 7% of its population

mattias800:
Refs #2998. Follow-up to #3060.

A background probe from #3060's review completed **after** that PR merged, and contradicted a number
it had published — which by then was in a shipped code comment and in `tools/evidence/AGENTS.md`.

| | published | actual |
| --- | --- | --- |
| assets probed | 13 | **233** |
| above the 40 ceiling at overlap ≥ 75% | 4 (31%) | **100 (42.9%)** |
| worst mean | 54.33 | **59.25** |

Same direction — the disproof of the removed mean ceiling is **stronger** than recorded — but an
understated figure sat in the very paragraph written to stop someone re-proposing that ceiling, which
is where a weak number does the most damage.

## The cause, recorded next to the number

The original probe filtered assets to `1920×1080` and so **excluded every 4K asset** — exactly where
the worst means live. A structured-asset sample runs roughly 6 such 1080p assets against 79 4K ones,
so it measured about **7%** of its own population, and specifically the slice that omitted the
`sce_sys/pic*` store art dominating the corrected list.

Its silence about the worst case was therefore *guaranteed rather than informative* — the same shape
as the charter's positive-control trap, where a control drawn from the source under test cannot
express the class it is meant to detect.

## The fix is to stop citing a count

Both sites now lead with the **analytic bound**: an overlap of ≥ 75% bounds the mean at ≤ 63.75/255
at full coverage. That is derivable, belongs to nobody's sample, and is what actually settled the
removal. The corrected count is kept beside it as a cross-check, along with a note that the earlier
figure came from a resolution-filtered probe.

The observed worst mean of **59.25** sitting just under the bound of **63.75** is the reassuring
part: two independent routes agreeing at the boundary is good evidence the bound is right.

## Verification

Comment and documentation only — no behaviour change. The tool's arms are unaffected and re-checked
(EXACT on the retracted frame still exits 2). `check_numbered_table.py` over all 152 tracked `.md`
files passes; `diff --check` clean.

Not merged under the docs-only exception, since the diff touches a `.py` file — even though only its
comments changed.

### DIFF
--- a/prosper/tools/evidence/AGENTS.md
+++ b/prosper/tools/evidence/AGENTS.md
@@ -52,7 +52,8 @@ Derive it rather than quoting it — it was first published here as 63.75, which
 pixels' own contribution (a pixel counted as matching may differ by up to 8 per channel, not 0;
 0.75 × 8 = 6.00 is exactly the gap), and a hand-built frame reaches mean 69.67 at overlap 75.00%,
 above the figure the same paragraph called unreachable. A companion claim that the worst case was
-partial coverage was backwards too: at f = 0.25 the bound is only 28.69. Measured alongside it, **100 of 233** structured assets (42.9%) carrying a bright
+partial coverage was backwards too: at f = 0.25 the bound is only 28.69. Measured alongside it,
+**100 of 233** structured assets (42.9%) carrying a bright
 caption panel reach mean 40.5–59.25 while still exceeding the bar, so a ceiling of 40 silently
 dropped nearly half. The case that motivated the ceiling was already rejected by the bar anyway.
 
@@ -60,8 +61,13 @@ The count was first published here as "4 of 13", from a probe that filtered to 1
 excluded every 4K asset — where the worst means live. A structured-asset sample runs roughly 6
 such 1080p assets to 79 4K ones, so that probe saw about 7% of its own population and could not
 express the case it was measuring. The observed worst mean, 59.25, sits under the analytic bound of
-69.75 — a consistency check rather than an agreement, since real assets cannot saturate a bound that
-requires every matching pixel to differ by exactly 8.
+69.75 — a consistency check rather than an agreement. Real assets *approach* the bound without being
+observed to reach it: saturating it needs every matching pixel to differ by exactly 8 *and* the
+non-matching quarter to be pure black. But do not read that as "cannot". An ordinary defect gets
+close — a blit with slightly wrong colour conversion shifts every pixel by a small constant, which
+keeps them inside the 8/255 match window while adding up to 6.00 to the mean. Measured on a real
+asset (`PPSA03001` `sce_sys/pic0.dds`, 24% caption panel): overlap 75.47% at mean 59.36, and with a
+uniform +8 shift overlap 75.97% at mean **65.43** — 94% of the bound, from real content.
 
 **And the two statistics need different gates**, which conflating them hid: `mean` is a whole-frame
 average that darkness cannot corrupt, so two structured images at mean 0.00 are the same image
--- a/prosper/tools/evidence/AGENTS.md
+++ b/prosper/tools/evidence/AGENTS.md
@@ -43,10 +43,17 @@ matches nothing at any brightness, and is not progression evidence in the first 
 **A mean ceiling was tried here and removed, which is worth knowing before adding one back.** It
 looked reasonable — "a high matching fraction alone is not domination" — but an overlap bar of 75%
 already bounds the mean at ≤ 63.75/255, so any ceiling below that can only discard frames the bar
-would have flagged. It cannot prevent a false positive and can only manufacture false negatives:
-4 of 13 real assets carrying a bright caption panel reach mean 40.5–54.3 while still exceeding the
-bar, and a ceiling of 40 silently dropped every one. The case that motivated it was already rejected
-by the bar.
+would have flagged. It cannot prevent a false positive and can only manufacture false negatives.
+That bound is the argument, and it is worth preferring to any count: it is derivable, so it belongs
+to nobody's sample. Measured alongside it, **100 of 233** structured assets (42.9%) carrying a bright
+caption panel reach mean 40.5–59.25 while still exceeding the bar, so a ceiling of 40 silently
+dropped nearly half. The case that motivated the ceiling was already rejected by the bar anyway.
+
+The count was first published here as "4 of 13", from a probe that filtered to 1920×1080 and so
+excluded every 4K asset — where the worst means live. A structured-asset sample runs roughly 6
+such 1080p assets to 79 4K ones, so that probe saw about 7% of its own population and could not
+express the case it was measuring. The observed worst mean, 59.25, sits just under the analytic
+bound of 63.75, which is the reassuring part: two independent routes agree at the boundary.
 
 **And the two statistics need different gates**, which conflating them hid: `mean` is a whole-frame
 average that darkness cannot corrupt, so two structured images at mean 0.00 are the same image
--- a/prosper/tools/evidence/AGENTS.md
+++ b/prosper/tools/evidence/AGENTS.md
@@ -42,18 +42,26 @@ matches nothing at any brightness, and is not progression evidence in the first 
 
 **A mean ceiling was tried here and removed, which is worth knowing before adding one back.** It
 looked reasonable — "a high matching fraction alone is not domination" — but an overlap bar of 75%
-already bounds the mean at ≤ 63.75/255, so any ceiling below that can only discard frames the bar
-would have flagged. It cannot prevent a false positive and can only manufacture false negatives.
-That bound is the argument, and it is worth preferring to any count: it is derivable, so it belongs
-to nobody's sample. Measured alongside it, **100 of 233** structured assets (42.9%) carrying a bright
+already bounds the mean, so any ceiling below that can only discard frames the bar would have
+flagged. It cannot prevent a false positive and can only manufacture false negatives. The bound is
+
+    mean ≤ 0.75·f·8 + 0.25·f·255 + (1−f)·15  =  f·69.75 + (1−f)·15
+
+for informative fraction `f`. It increases in `f`, so it maximises at **full** coverage: **69.75**.
+Derive it rather than quoting it — it was first published here as 63.75, which dropped the matching
+pixels' own contribution (a pixel counted as matching may differ by up to 8 per channel, not 0;
+0.75 × 8 = 6.00 is exactly the gap), and a hand-built frame reaches mean 69.67 at overlap 75.00%,
+above the figure the same paragraph called unreachable. A companion claim that the worst case was
+partial coverage was backwards too: at f = 0.25 the bound is only 28.69. Measured alongside it, **100 of 233** structured assets (42.9%) carrying a bright
 caption panel reach mean 40.5–59.25 while still exceeding the bar, so a ceiling of 40 silently
 dropped nearly half. The case that motivated the ceiling was already rejected by the bar anyway.
 
 The count was first published here as "4 of 13", from a probe that filtered to 1920×1080 and so
 excluded every 4K asset — where the worst means live. A structured-asset sample runs roughly 6
 such 1080p assets to 79 4K ones, so that probe saw about 7% of its own population and could not
-express the case it was measuring. The observed worst mean, 59.25, sits just under the analytic
-bound of 63.75, which is the reassuring part: two independent routes agree at the boundary.
+express the case it was measuring. The observed worst mean, 59.25, sits under the analytic bound of
+69.75 — a consistency check rather than an agreement, since real assets cannot saturate a bound that
+requires every matching pixel to differ by exactly 8.
 
 **And the two statistics need different gates**, which conflating them hid: `mean` is a whole-frame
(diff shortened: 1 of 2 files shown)


mattias800 (commented):
# Review — #3077 @ `1fc0ecf5`

**REJECTED**, narrowly, and the defect is **mine, not yours**: the analytic bound this PR promotes to
"the argument" is one I gave you, and it is stated tighter than it can be proved. Everything else in
the diff is accurate and the rewrite is the right shape.

You asked whether the replacement wording overstates in the other direction. It does — not in the
counts, which are now right, but in the bound itself.

## The bound is not 63.75. It is 69.75, and the tool will print it

`AGENTS.md:45` and `prerender_check.py:201` both say an overlap ≥ 75 % "already bounds the mean at
≤ 63.75/255". **That number drops the contribution of the matching pixels.** A pixel counted as
matching may differ by up to 8 per channel, not 0. With `f` = coverage:

```
mean <= 0.75*f*8  +  0.25*f*255  +  (1-f)*15   =   f*69.75 + (1-f)*15
         matching     non-matching   non-informative (both sides < 16)
```

That is increasing in `f`, so it maximises at **full** coverage: **69.75**. The missing term is
exactly `0.75 * 8 = 6.00`.

**Constructed by hand and run through the tool itself**, rather than left as a derivation — 75 % of
pixels differing by exactly 8, 25 % by exactly 255:

```
compared 1 same-size asset(s) (0 wrong-size, 0 undecodable, 0 featureless)
  mean=  69.75/255  within8= 75.00%  coverage=100.00%  asset.png

DOMINATED: 75.00% of this frame's INFORMATIVE pixels are within 8/255 of the
  stored asset asset.png (coverage 100.00%, mean abs diff 69.75/255 ...)
```

Overlap exactly at the bar, mean **69.75** — above a figure the comment three lines above says cannot
be reached. The tool refutes its own comment.

**The parenthetical is also backwards.** `prerender_check.py:201` adds "(and ≤ 75.75 in the worst
coverage case)". The worst case is *full* coverage, not low: the expression is increasing in `f`, and
at `f=0.25` the bound is only 28.69. 75.75 is a valid over-estimate arrived at by wrong reasoning — I
produced it by mistakenly multiplying the non-informative ceiling by the 0.75 matching fraction. Both
numbers should collapse to the single correct **69.75**.

## Suggested edits

- `prerender_check.py:201` — "already bounds the mean at <= 63.75/255 (and <= 75.75 in the worst
  coverage case)" → **"already bounds the mean at <= 69.75/255 (0.75*8 + 0.25*255, maximised at full
  coverage)"**. Carrying the arithmetic inline is worth the few characters: it is what makes the
  claim checkable without re-deriving, which is the whole reason for preferring a bound to a count.
- `AGENTS.md:45` — same substitution.
- `AGENTS.md:56` — "The observed worst mean, 59.25, sits just under the analytic bound of 63.75,
  which is the reassuring part: two independent routes agree at the boundary." Against 69.75 the
  observed maximum is 85 % of the bound rather than 93 %, which is a **consistency check, not an
  agreement at the boundary**. Suggest: *"sits comfortably under the analytic bound of 69.75 —
  consistent, as it must be, and a cheap check that neither route is badly wrong."* Real assets
  cannot saturate the bound anyway, since it needs every matching pixel to differ by exactly 8.

## Nothing else changes, including the decision

**The conclusion is untouched.** 40 < 63.75 < 69.75, so the removed ceiling sat inside the region the
bar accepts either way; it could still only manufacture false negatives. Correcting the bound makes
the disproof *stronger*, because the gap between the ceiling and the bound is wider than stated.

Everything else in the diff is right and I re-checked it: **100 of 233 (42.9 %)** and worst mean
**59.25** match the completed probe and my independent spot-check of its top hit; the 1920x1080 filter
diagnosis is accurate; the roughly 6-to-79 structured-asset ratio matches what I measured; and
"cite the bound; a count is a property of whoever's sample" is the right rule — which is precisely
why the bound has to be correct.

## On the ledger

Thank you for the reframing, but I would not let mine off that lightly: I published 63.75 twice, you
took it in good faith on my authority, and it is now in two files with the word "derivable" attached —
which is the strongest claim a number can carry here and the one that most discourages checking. That
is the charter's authority-transfer trap operating through me, in the same review where I invoked it
against a reviewer's citation. The correction is cheap only because the probe happened to finish.

**REJECTED** at `1fc0ecf5` — three number substitutions and one softened sentence, then it should
land. The count corrections in it are already right and worth having.

mattias800 (commented):
# Review — #3077 @ `41aeafcd`

**APPROVED.** The derivation is correct, I re-checked every term and every number in it, and the diff
is comment-only (AST hash of `prerender_check.py` is identical at `c011e97f` and `41aeafcd`, so the
verdict logic is untouched).

You asked whether the replacement overstates in a third direction. **It does, in one clause — and
that clause is mine.** It is a subordinate caveat rather than the argument, so I am not blocking on
it, but it is wrong and it is cheap to fix.

## Verified

| claim | check |
| --- | --- |
| `mean ≤ 0.75·f·8 + 0.25·f·255 + (1−f)·15` | each term correct: matching ≤ 8, non-matching ≤ 255, non-informative both < 16 so ≤ 15 |
| `= f·69.75 + (1−f)·15` | ✓ |
| increasing in `f`, maximised at full coverage | ✓ slope 54.75 > 0 |
| **69.75** at `f = 1` | ✓ |
| `0.75 × 8 = 6.00` is exactly the gap | ✓ 63.75 + 6.00 = 69.75 |
| `f = 0.25` → **28.69** | ✓ 28.6875 |
| hand-built frame reaches 69.67 at overlap 75.00 % | ✓ consistent with mine at 69.75 (yours will sit a hair under if any `+8` clamps at 255) |
| 100 of 233 (42.9 %), worst 59.25, 6-to-79 ratio | ✓ unchanged, all still match |

Showing the derivation instead of a figure is the right fix, and "derive it rather than quoting it"
earns its place given how this one travelled.

## The third-direction overstatement: "real assets **cannot** saturate"

`AGENTS.md`, last sentence of that paragraph:

> …a consistency check rather than an agreement, since real assets **cannot** saturate a bound that
> requires every matching pixel to differ by exactly 8.

I wrote that in my last review and you took it verbatim. **"Cannot" is too strong, and the mechanism
that breaks it is an ordinary emulator defect rather than a contrivance:** a blit with slightly wrong
colour conversion shifts every pixel by a small constant. Those pixels still count as *matching*
(diff ≤ 8), so the overlap is unchanged while the mean gains `0.75 × 8 = 6.00`.

Measured on a **real** asset — `PPSA03001` `sce_sys/pic0.dds`, 3840x2160 — with a 24 % caption panel:

```
panel 24%                    overlap 75.47%   mean 59.36    (-4.39 vs 63.75)
panel 24% + uniform +8       overlap 75.97%   mean 65.43    (+1.68)  <-- real content, above 63.75
```

65.43 is **94 % of the bound**, with real pixels. So the honest statement is that real assets
*approach* the bound but are not observed to reach it — saturating it exactly needs every matching
pixel to differ by exactly 8 *and* the non-matching quarter to be pure black.

Suggested: *"…a consistency check rather than an agreement: real assets approach the bound without
reaching it (a real asset with a caption panel and a uniform +8 shift — what a mis-converted blit
produces — measures 65.43, 94 % of the bound), since saturating it needs every matching pixel to
differ by exactly 8 and the non-matching quarter to be pure black."*

**Why this is a note and not a rejection, since I rejected the last round for a wrong number too:**
last time the false figure *was* the argument, elevated to "derivable, belongs to nobody's sample" —
a reader was told to trust it over any measurement. This is a subordinate clause qualifying how
loosely two numbers agree, it supports no threshold (the ceiling is gone), and nothing downstream
depends on it. Different in kind, not just in size.

## Minor

`AGENTS.md:55` is **154 characters** against a file whose median line is 97 and whose next-longest is
116 — the "…at f = 0.25 the bound is only 28.69. Measured alongside it, **100 of 233**…" line did not
get re-wrapped after the insert. Cosmetic; worth a re-wrap while you are in there.

Gates: `check_numbered_table.py` on the changed `AGENTS.md` → exit 0; `py_compile` clean; diff is the
expected two files.

## On the ledger, since you pushed back

Your sharper version is right and I accept it: the failure was the *label*, not the number. But note
what just happened — I supplied "cannot saturate", you adopted it verbatim, and it took a targeted
experiment to find it false. **Third time in this review that a phrase of mine propagated on
authority**, after the int16 figure and the 63.75 bound. The thing that caught it every time was not
either of us being careful in general; it was you asking a specific adversarial question — "does this
overstate in some third direction" — which is a habit worth keeping past this PR.

**APPROVED** at `41aeafcd`. Land it; fold in the "cannot" softening and the re-wrap if it is
convenient, or take them as a follow-up.

mattias800 (commented):
# Review — #3077 @ `68afaf4c`

**APPROVED.** The softening is accurate, my figures are transcribed correctly, the long line is
re-wrapped, and the diff is still comment-only.

## Verified

- **The "cannot" claim is gone and what replaced it is supportable.** "Real assets *approach* the
  bound without being observed to reach it… But do not read that as 'cannot'" is exactly the right
  strength — it states the observation, names the saturation condition, and explicitly blocks the
  over-read. Making the counterexample sit beside it is better than a hedge, because the next reader
  gets the mechanism rather than a caveat they have to trust.
- **My numbers are transcribed correctly**, re-measured just now: `PPSA03001` `sce_sys/pic0.dds`,
  24 % caption panel → overlap 75.47 % / mean 59.36; with a uniform +8 → overlap 75.97 % / mean
  **65.43**; 65.43 / 69.75 = 93.8 %, so "94 % of the bound" is right.
- **Re-wrap done:** the 154-char line is gone, max is now 116. The two remaining lines over 100
  (104 and 116) are pre-existing, not introduced here.
- **Code untouched:** AST hash of `prerender_check.py` with docstrings stripped is identical at
  `c011e97f` and `68afaf4c` — comment-only, verdict logic unchanged. `check_numbered_table.py` on the
  changed `AGENTS.md` → exit 0. Two files touched, as expected.

## On the trap entry — yes, please draft it, with one correction to its premise

I agree it is worth recording, and I would rather it exist than not. But **the premise as you stated
it is wrong in a way that matters, and enshrining it uncorrected would make the entry self-refuting.**

**Two phrases propagated into your tree, not three.** Checked against the record rather than memory:

| claim | reached the tree? | how it was caught |
| --- | --- | --- |
| `63.75` bound | **yes** — `AGENTS.md` + code comment at `c011e97f` | your targeted question on #3077 |
| "cannot saturate" | **yes** — `AGENTS.md` at `41aeafcd` | your targeted question ("some third direction") |
| int16 / 43.6 % figure | **no** — 0 occurrences in any tree | I caught it myself, pre-publication |

The int16 figure appears exactly once in everything I posted across both PRs: **inside my own
correction of it.** It was never published as a claim and never entered your tree.

**And it was caught by a different mechanism, which is the part worth having.** I did not catch it by
being careful, and nobody asked me about it. I caught it because I computed the same quantity a
second way — my numpy luminance weighting said 43.6 %, PIL's own metric said 1.27 % — and two routes
to one number disagreed. That defence has a property the adversarial question does not: **it does not
require anyone to suspect anything.** The question only fires when someone thinks to ask; the
cross-check fires on its own.

So the honest entry names **two** defences, not one:

1. **Ask the adversarial question precisely where a claim arrived labelled as already checked** — a
   reviewer's citation, a derived bound, anything carrying "derivable". Your own observation is the
   sharpest thing here and should survive verbatim in substance: you asked the question about
   parameters *you* had chosen and felt uncertain about, and never once about a number handed to you,
   which is exactly where you were least able to see the risk.
2. **Compute a load-bearing quantity twice by independent routes.** Caught the one error that never
   reached the tree, and needs no suspicion to trigger.

That correction also makes the entry stronger rather than weaker: 2 of 2 propagated claims were
caught only by a targeted question, and the one caught structurally never propagated at all. That is
a cleaner result than "three, all caught by questions" — and "three phrases propagated" is itself a
true-sounding claim reached by a route that does not establish it, which is the entry's own subject.

## Placement, so it does not become two drifting copies

`CLAUDE.md` already carries this ground at lines **738–767**: *"A cited claim reads as a verified one
— and reviewers' findings propagate hardest"*, the *"a finding you WANT to be true gets checked
least"* asymmetry, and *"a true statement reached by a route that does not establish it."* Your
material genuinely extends that — none of those three name *who asks* or the self-selection of what
gets questioned — but a full second write-up in the charter would duplicate it, and this file's own
history is that duplicated guidance drifts and the stale copy keeps its authority.

Suggest: **the numbered trap entry is the authoritative record** (evidence, PR links, the two
defences), and the charter bullet gets a one-line pointer to it. One home, one cross-reference.

Mechanics: allocate with `python3 prosper/tools/docs/trap_number.py` rather than reading master —
master's highest is 230 but the allocator also scans open PRs, and reading master alone is what
produced the #2574/#2581 collision. Run `check_merge_result.py` immediately before merging, since a
concurrent lane appending to that table is invisible to a green `Docs` job.

On attribution: please cite the PRs and review threads rather than me. That is how the other entries
read, it is checkable by someone who was not here, and the evidence is the part that has to survive.

**APPROVED** at `68afaf4c`. Land it. Send the trap draft when you have it and I will review it.

PR #3078, opened 2026-08-27
docs: instrument trap 231 — a claim that arrives already labelled as checked

mattias800:
Refs #2998. Follows the #3060 / #3077 review series, which is the entry's own evidence.

## What it records

`CLAUDE.md`'s authority-transfer section already says a cited claim reads as a verified one. This adds the part it does not say: **which claims actually get questioned, and how little that depends on care.**

Over one review, two of a reviewer's phrases reached my tree underived:

- an analytic bound stated as `63.75` — the true bound is `69.75`, it dropped the matching pixels' own contribution (`0.75 × 8 = 6.00`);
- the claim that real assets "cannot saturate" it — a caption panel alone reaches **85%** of the bound, and adding an ordinary colour-conversion shift takes it to **94%**.

Both were caught by the same question asked twice about successive revisions of **one paragraph**, and neither by care, review or CI.

## The finding worth having

**Neither catching question was aimed at the reviewer's numbers.** Both were aimed at my own rewrite. Those claims were re-read only because they happened to sit inside a paragraph being rewritten *for an unrelated reason* — a wrong count, itself surfaced by a background probe the reviewer had written off as dead.

**Two accidents were required.** Had the count been right, nobody would have reopened that paragraph, and `63.75` would still be in the tree wearing the label *"derivable, belongs to nobody's sample"* — the strongest label a number can carry here, and the one that most discourages anyone from deriving it.

So the entry is explicit that only the second defence scales:

1. ask the adversarial question exactly where a claim arrived pre-labelled as checked, since that is where it feels least necessary;
2. **compute any load-bearing quantity twice, by independent routes, and compare** — the only one of the two that fires without anyone suspecting anything.

A third wrong figure from the same review never propagated at all because its author computed it two ways, got 43.6% against 1.27%, and the disagreement exposed an int16 overflow. Caught before publication; the questions caught theirs two rounds after.

## The entry was reviewed and reddened twice

Worth stating, since it is the entry's own subject:

- draft 1 said **three** phrases propagated. The record said **two** — the int16 case never became a claim.
- draft 2 said an ordinary colour-conversion shift reaches 94% of the bound. A shift **alone** reaches 11.5%; the 94% needs the caption panel too.

A third-direction overstatement inside the entry about third-direction overstatements.

## Scope

One numbered entry plus a **one-line pointer** from `CLAUDE.md`, deliberately not a second write-up — the charter already holds this ground and a duplicate would drift, with the stale copy keeping its authority.

## Verification

`.md`-only. Trap gate passes both arms: `1..231, unique and ascending, no unused numbers`, and `--baseline origin/master` reports none of the 230 base rows deleted. `check_numbered_table.py` over all 152 tracked `.md` files passes. `diff --check` clean. `check_merge_result.py` to be run immediately before merge, since it is the only check that sees a concurrent append.


mattias800 (commented):
# Review — #3078 @ `95999261`

**APPROVED.** Every factual claim checks out, both corrections landed, and correction 2 as the spine
is the right structure — the entry now says something the charter's existing section does not.

One clause is loose, and the accurate replacement is **stronger** than what is there. Not blocking;
details below so you can judge and edit without a re-review.

## Verified

| claim in the entry | check |
| --- | --- |
| true bound `69.75`, dropped term `0.75 × 8 = 6.00` | ✓ |
| caption panel alone **85%** of the bound; + colour shift **94%** | ✓ 85.1% and 93.8% |
| both catches were the same question about successive revisions of one paragraph | ✓ "other direction" → `63.75`; "third direction" → "cannot saturate" |
| neither question aimed at the reviewer's numbers | ✓ both aimed at the author's own rewrite |
| int16: two routes, 43.6% vs 1.27%, caught pre-publication | ✓ never entered any tree |
| "The first draft of this entry said three phrases propagated. The record said two." | ✓ flat, no wink — this reads as rigour now |
| attribution to `#3060 / #3077` | ✓ checkable by someone who was not here |

Gates at head: trap gate both arms (`1..231, unique and ascending`; `none of the 230 base rows
deleted`), `check_numbered_table.py` on `CLAUDE.md` clean, `.md`-only diff confirmed, `diff --check`
rc=0, `Claude-Session:` trailers 0. New `CLAUDE.md` lines max 100 chars, matching the file.

## The loose clause — "and neither by care, review or CI"

A review is precisely what caught them, so as written it invites the objection. **The accurate
version is available, checkable, and considerably more damning:**

`c011e97f` is the commit that **added** both `63.75` lines —

```
+already bounds the mean at ≤ 63.75/255, so any ceiling below that can only discard frames the bar
+    # >= 75% already bounds the mean at <= 63.75/255 (and <= 75.75 in the worst coverage case), so
```

— and the reviewer **approved that exact commit**, twice, having supplied the number two rounds
earlier. An ordinary review pass read the diff that introduced the wrong number and passed it.

Suggested: *"…and not by the ordinary review pass over the same text: the reviewer approved the very
commit that added `63.75`, two rounds after supplying it."*

That closes the loophole a reader reaches for — *surely review would normally catch that* — with a
specific, verifiable instance rather than an assertion. It is the single best piece of evidence the
entry has and it is currently absent.

## Minor — enumerate the two accidents

"**Two accidents were required**" is correct but the two are implied rather than named, so a reader
cannot check it. They are:

1. a background probe the reviewer had written off as dead completed anyway;
2. the error it surfaced happened to sit in **the same paragraph** as the bound.

The second carries the argument and is the one worth stating: had the wrong count been in another
file, the bound would never have been re-read. Naming both turns a rhetorical flourish into a claim
someone can audit — which is what the entry asks of everyone else.

## On the `CLAUDE.md` bullet

Seven lines rather than the one we discussed, but the right seven: it carries the actionable defence
into the section people read first, and names trap 231 so the evidence is one hop away. Drift risk is
low because it contains no figures — except "two of a reviewer's phrases", which is the one number in
this entire episode that was already wrong once. It is now verified, so keep it; just be aware that
is the element to re-check if the trap entry is ever revised.

## Verdict

**APPROVED** at `95999261`. Docs-only, so the CI exception applies — run `check_merge_result.py`
immediately before merging as you planned, since a concurrent append to that table is the one thing a
green `Docs` job cannot see.

I would make the clause edit and the enumeration first; both are phrase-level, both strengthen rather
than soften, and neither needs another look from me. If you would rather ship as-is, the clause is
loose rather than false — the ordinary pass genuinely did miss it — so that is a defensible call and
the entry stands either way.

PR #3081, opened 2026-08-27
gpu(#2998): validate a layered array's whole span — Tomb Raider's world renders correctly textured

### LINKED ISSUE #2998: Tomb Raider I-III Remastered (PPSA16901): world surfaces sample ONE texel each — degenerate UVs, not missing textures

mattias800:
Fixes #2998. Refs #2990.

## Tomb Raider's world renders correctly textured

Croft Manor assault course

*Tomb Raider I-III Remastered* (`PPSA16901`), Lara's Home, on the default route
`scripts/tomb-raider-PPSA16901/reach-gameplay.pad`, captured at 150 s with `prosper-app`
(offscreen). Brickwork, sandstone, mossy wooden platforms, gravel, ivy, Lara, and Winston with the
tea tray. Confirmed a **genuine render** by `tools/evidence/prerender_check.py` (exit 0) rather than
the title's own artwork — the check that instrument trap 230 exists for.

## The defect

The persistent texture-decode cache computed its validation range as **one surface** for a layered
array. For this atlas — 512×512 BC7, 256 layers, `layer_stride` 352256 — that is 262144 bytes of
90177536: it proved reuse against **0.29%** of the texture.

Everything above the first layer therefore changed invisibly. The title fills the atlas
*progressively*, so:

- prosper decoded it once while it was ~5% populated (`[occupancy]` full-byte census: `filled=19/344`);
- the guest reached **93%** by gameplay (from a `.prgbundle`: `nz=84070204` of `footprint=90177536`);
- the cache served the early decode for the rest of the run.

Interiors then sampled layers still holding the **previous occupants** of that memory — which is why
walls and floors were drawn with Lara's passport, the Game Boy collectibles and an inventory
document page. Real, correctly decoded assets on correct geometry; simply the wrong ones.

## Why the fix is shaped this way

`layered_cube_source_size` has always spanned a cube's six faces, and its own comment states the
invariant this violated:

> *"cache reuse can never be proved against less memory than the layer-aware decoder reads."*

Arrays were added later (#3050) without the matching source-size change. This adds
`layered_array_source_size` beside it, failing closed identically: a zero surface size is the
caller's "do not cache" signal and is never widened into a cacheable span; a missing stride or a
single layer leaves the size alone; an overflowing span returns 0 rather than a truncated range,
because a span shorter than the decoder reads **is** the defect being fixed.

## The control

Same route, same binary, fix toggled by one `false &&`:

| | `source_size` | frame | distinct colours |
| --- | --- | --- | --- |
| fix **on** | 90087424 | Croft Manor's assault course | 245141 |
| fix **off** | 262144 | passports and Game Boys on the walls | 124916 |

Mean difference between the two frames: **51.48/255**. Both are confirmed genuine renders.

## Deliberately unaffected

Only `guest_array` textures (`img_dim == 5`, `depth > 1`, block-compressed) change span; every other
descriptor returns exactly what it did before. Non-layered textures, cubes (which had their own
correct path), volumes and DCC fast-clear ranges are untouched.

**Risk worth naming:** a larger validated span means more honest invalidation, so a title that
rewrites a large array every frame will re-decode more often than before. That is the correct
behaviour — it was previously serving stale pixels — but it is a real performance surface, which is
why the snapshot guards below were run.

## Verification

- build clean; `ctest --test-dir prosper/build-linux --no-tests=error -j4` → **312 of 312 passed**
- `check_diag_gates.py` — all checks passed; `check_numbered_table.py` over 152 tracked `.md` — passes
- `diff --check` — clean
- Snapshot guards: **`messenger-scene` OK**, **`dead-cells-gameplay` OK**, **`blasphemous2-gameplay` OK**

**The four new test arms are mutation-verified**: restoring the single-surface return makes three of
them fail and the binary report `FAILED`; they pass again when reverted. Without that check they
would have been assertions unable to distinguish the fix from the bug.


mattias800:
## Note on the span's coupling to the array decode budget

The subtle part of this change is which layer count the span uses, so stating it explicitly:

```
is_array       = guest_array && array_footprint <= array_budget_bytes   (live_renderer.cpp:3293)
decoded_layers = is_array ? (r.depth ? r.depth : 1u) : 1u               (:3305)
span           = layered_array_source_size(surface_bytes,
                                           r.layer_stride_bytes,
                                           decoded_layers)              gated on guest_array
```

The span is computed from **`decoded_layers`, not `r.depth`**, and that is deliberate. When a
descriptor exceeds `PROSPER_ARRAY_DECODE_BUDGET_MIB`, `is_array` goes false, only the base slice is
decoded, and `decoded_layers` collapses to 1 — at which point the helper's `layers <= 1` guard
returns `surface_bytes` unchanged. So an over-budget array validates exactly the one surface it
actually decoded.

That is the property that matters: **the validated span always follows what was decoded, never what
the descriptor declares.** Using `r.depth` would have re-introduced the same class of bug in the
opposite direction — validating memory the decoder never read, so an unrelated write to a layer we
never sampled would invalidate a perfectly good decode.

For this title the budget is not reached: the footprint is 512×512×4×256 = 256 MiB against a 1024 MiB
default, so `is_array` is true and all 256 layers are both decoded and validated.

mattias800 (commented):

One blocking finding, and it is a *residual instance of the exact defect class this PR fixes* — not a
regression, but the PR's comment and one of its new test arms both assert the case is deliberately
safe, and it is not. Everything else I checked is correct, and I re-derived the central arithmetic
independently and it holds. This should be quick to clear; I will re-review on the corrected head.

### What I verified, and it holds

- **The span is exact for the title's real descriptor.** `layer_mip_offset = 90112`,
  `layer_stride = 352256`, 256 layers, surface 262144. `sampled_source_addr = gpu_addr + 90112`
  (`texture_decode_source_address`, img_dim 5, `!in_mip_tail`), so the validated range is
  `[gpu_addr+90112, gpu_addr+90112+90087424)` = `[gpu_addr+90112, gpu_addr+90177536)`. The decoder's
  last slice reads `gpu_addr + 255*352256 + 90112` for 262144 bytes, ending at exactly
  `gpu_addr+90177536`. Span end and last read end coincide to the byte, and `90112 + 90087424` is the
  descriptor footprint. The 90087424 in the PR body is reproducible from the layout, not just observed.
- **Overflow handling is correct.** `layer_stride_bytes` is provably non-zero at
  `texture_decode_cache_policy.hpp:59` (the `:58` early return), so the division guard at `:60` is
  well-defined; the redundant `layer_stride_bytes &&` is harmless. `4096, UINT64_MAX/2, 8` wraps to
  `2^63 - 7`, whose division check yields 0 != 7 -> returns 0. The `UINT64_MAX - surface_bytes` and
  `span > SIZE_MAX` guards are correct on 64- and 32-bit `size_t` alike.
- **A zero `surface_bytes` cannot be widened.** `:57` is the first statement, so the "do not cache"
  signal survives every later branch. I also traced the lambda refactor in `live_renderer.cpp`: the
  inner lambda's early `return size_t{0}` (the forced-detile unorm8x4 exclusion) still reaches the
  caller as 0, because `layered_array_source_size(0, ...)` returns 0.
- **`decoded_layers` is the right count.** `is_array = guest_array && array_footprint <= budget`, and
  `decoded_layers = is_array ? depth : 1` (`live_renderer.cpp:3305`) is the same value the decode loop
  uses for `slice_count` (`:5380`) and the same one `texture_pixels` is sized by (`:4767`). Under
  `PROSPER_ARRAY_DECODE_BUDGET_MIB` low enough to drop the title to base-slice-only, `decoded_layers`
  becomes 1, the helper returns `surface_bytes`, and the decode loop is not entered at all — the two
  stay consistent in both regimes. Default budget is 1024 MiB against this atlas's 256 MiB, so the
  default route really does take the 256-layer path.
- **Blast radius matches the claim.** The only behavioural edit is `live_renderer.cpp:3664-3666`,
  guarded on `guest_array`, which is `img_dim == 5 && depth > 1 && bc_block_bytes(format) != 0` plus
  `cls == Texture` (`gpu/texture/bc_decode.hpp:35`). Every other descriptor returns the pre-PR
  `surface_bytes` unchanged. Cubes still take `layered_cube_source_size`, DCC fast-clear still
  overrides with the metadata range, and `dcc_metadata_footprint` returns 0 for BC formats so the two
  cannot interact.
- **Per-branch coverage check against the decoder.** For the reachable BC array shapes the span
  equals or exceeds the read: tiled non-tail -> `tiled_elements_bytes` on both sides, identical;
  untiled -> `bw*bh*cb` on both sides, identical; `linear_padded_read` -> helper's
  `linear_src_row * linear_row_count` covers `copy_linear_padded_rows_from`'s `bh` rows at pitch
  `linear_src_row` with one row's padding to spare. No short case found other than the blocking one below.
- **Mutation claim checks out arithmetically.** Restoring a single-surface return fails arms at
  `test_texture_decode_diagnostic.cpp:114`, `:117` and `:128` and leaves `:120`, `:121`, `:126`
  passing — exactly the "three of them fail" in the PR body.
- **The screenshot is what it says it is.** Correctly textured brick, sandstone, mossy platforms,
  gravel, ivy, Lara and Winston, at an exposure that no longer reads as over-bright.

### BLOCKING — the same defect survives when `layer_stride_bytes == 0`, and a test arm now pins it

`prosper/frontends/shared/texture/texture_decode_cache_policy.hpp:58`

    if (layers <= 1u || !layer_stride_bytes) return surface_bytes;

The header comment justifies this as failing closed, and
`test_texture_decode_diagnostic.cpp:119-121` states the rule outright: *"the span must never be
invented where the layout does not state one."* But the decoder **does** invent one, in exactly that
case, at `prosper/frontends/shared/live/live_renderer.cpp:5375-5378`:

    auto face_base = [&](uint32_t face, size_t selected_span) {
        const uint64_t stride = r.layer_stride_bytes
            ? r.layer_stride_bytes : selected_span;
        return r.gpu_addr + static_cast<uint64_t>(face) * stride;
    };

with `slice_count = decoded_layers` at `:5380`. So with no declared stride the decode reads
`decoded_layers` consecutive surfaces, and the cache validates one of them — 1/N coverage, the
identical failure the PR exists to remove.

**Concrete failure scenario.** A BC 2D_ARRAY T# (SQ_RSRC_IMG TYPE 13 -> `img_dim == 5`), `depth = 8`,
`BASE_LEVEL = 0`, `BASE_ARRAY = 0`, whose `tile_mode` is any GFX10 swizzle prosper does not model —
`tile_mode_is_tiled()` covers only {1, 5, 9, 24, 27} (`src/gpu/texture/tile.cpp:37-43`).
`tiled_mip_level_layout()` returns `supported = false` for such a mode
(`src/gpu/texture/tile.cpp:1454`), so `image_base_level_view()` takes the early return at
`src/gpu/agc/agc_shader_layout.cpp:509-512`, where `unshifted_view_supported()` is **true** for
BASE_LEVEL 0 / BASE_ARRAY 0 — the descriptor is accepted with `view.layer_stride` left at 0. That
reaches `live_renderer.cpp` as `img_dim == 5`, `depth == 8`, BC, `layer_stride_bytes == 0`, so
`guest_array` is true and `decoded_layers == 8`. The decode loop reads
`gpu_addr + k*comp` for k in 0..7 (`ctiled` is false for an unmodelled mode, so
`selected_span == comp == surface_bytes`), i.e. `8 * surface_bytes`. `persistent_source_size` is
`surface_bytes`. Slices 1..7 then change invisibly and a stale decode is served for the rest of the
run. Replay reaches it too: the `.prgbundle` reader explicitly admits a zero `layer_stride_bytes`
(`src/gpu/capture/gpu_capture.cpp:4548`).

I accept this is **pre-existing** — the PR does not make it worse. What makes it blocking rather than
a follow-up is that the PR *documents it as correct* in two places a later reader will trust: the
header comment, and `test_texture_decode_diagnostic.cpp:121`, which now encodes the wrong answer as a
guarded invariant. Per the charter, a claim carrying a citation and a green test is the most
expensive kind of wrong thing to leave behind.

**Suggested fix** — mirror `face_base` exactly, which is provably right here: in the reachable
zero-stride configuration `in_mip_tail` is false and `layer_mip_offset` is 0 (both are only assigned
after `layout.supported`), so `selected_span == surface_bytes` and the correct span is
`layers * surface_bytes`.

    if (layers <= 1u) return surface_bytes;
    // No declared stride is not "not layered": face_base() falls back to one surface per slice
    // (live_renderer.cpp:5375), so the span must follow it rather than stopping at slice 0.
    const uint64_t stride = layer_stride_bytes ? layer_stride_bytes : surface_bytes;

then use `stride` in the two guards below. The `:121` arm becomes `== 262144u * 256u`, which also
makes it discriminate against the mutation instead of passing under it. If you would rather not widen
here, returning 0 (do not cache) is equally sound — but then the comment and the arm must say
*"unproven layout, so we decline to cache"*, not *"the layout does not state a stride, so one surface
is right"*.

### SHOULD FIX — `TOMB_RAIDER_STATUS.md` now contradicts itself about #2998

`prosper/docs/TOMB_RAIDER_STATUS.md:84` still opens `## Open defects` with:

    1. **Every world and character surface is untextured** and the scene is over-bright (#2998).
       The atlas is bound, nothing is rejected, and the menus are fully textured.

while the new `## Ruled out` entry in the same file — and `Fixes #2998` on this PR — say it is
resolved. The charter's whole complaint about tracker bodies lagging their own threads applies
verbatim to a status doc that states both. Move or rewrite item 1 in the same PR, and renumber the
list. Note this also settles the over-bright question: the row in `COMPATIBILITY.md` drops
"the scene is over-bright" (correctly, on the evidence of the screenshot), but `:84` is where that
claim actually lives, and it is still standing.

### Non-blocking

1. **The retained validation copy grows 344x for this texture, and the PR body does not name it.**
   `cached.source_prefix` holds `source_prefix_size` bytes (`live_renderer.cpp:6338-6343`) and
   `bytes()` counts it against `persistent_decode_limit` — so this atlas's entry goes from
   ~256 MiB + 256 KiB to ~256 MiB + 86 MiB. It still fits the >= 1024 MiB floor from
   `texture_decode_cache_limit_bytes`, so no eviction cliff, and `safe_equal` compares in place
   rather than through a scratch copy. Worth one line in the "Risk worth naming" paragraph next to
   the re-decode point, because it is the half that shows up as RSS rather than as frame time.
2. **Deliberate over-coverage, and I think it is the right call — just say so.** The span includes each
   layer's lower-mip chain bytes (`352256 - 262144` per layer, ~23 MiB of the 90 MiB) that the decode
   never reads, so a guest write to a lower mip invalidates. That is conservative in the safe
   direction and consistent with `layered_cube_source_size` using the descriptor footprint including
   alignment gaps. Same for the case where the array loop is skipped (`rtt_hit`,
   `dcc_fast_clear_done`): wider span than the bytes read, over-invalidation only.
3. **Indentation.** The inner lambda body at `live_renderer.cpp:3610-3651` keeps its original
   indentation so the diff stays a pure wrap. That is reviewable now and misleading later — worth a
   separate mechanical re-indent commit rather than leaving a lambda body at its enclosing scope's level.
4. `COMPATIBILITY.md` still says `Last updated: 2026-08-26` while the row changed on the 27th.

### Verdict

**REJECTED** on finding 1 (with the status-doc contradiction as a should-fix in the same push). The
code change itself is correct, the evidence is genuine, and the rung-3 wording is not overstated —
the row keeps rung 3, the scene plainly renders, and nothing claims the human-verified rung 4. Fix
the zero-stride case (or restate it honestly), update `TOMB_RAIDER_STATUS.md:84`, and I will
re-review on the new head.

mattias800 (commented):
## Re-review at `25bb49cd` — REJECTED

**The code fix is correct and I have verified it end to end — I would approve it as it stands.** What
holds this back is two documentation items, both cheap, one of which the charter makes an explicit
merge precondition. No further code re-verification will be needed: send me the doc edits and I will
approve on sight.

### The helper is now right, and it is right for a stronger reason than the comment claims

- **Overflow is well-defined again.** `stride` is provably non-zero at
  `texture_decode_cache_policy.hpp:73` because `surface_bytes` is non-zero (the `:57` early return),
  so the division guard at `:74` cannot divide by zero even though the explicit
  `layer_stride_bytes &&` test is gone. Correct simplification, not a lost guard.
- **The zero-stride span is EXACT, not merely conservative.** I expected to have to call it
  "safely wider"; it is not. On the live path `layer_mip_offset_bytes` is only assigned inside the
  `thin_2d_layered && depth > 1` block of `image_base_level_view`, which is precisely the block that
  sets a non-zero stride — so a zero-stride descriptor always carries offset 0. Replay enforces the
  same thing from the other direction: `gpu_capture.cpp:4550` rejects
  `!layered && layer_mip_offset_bytes != 0`. So `sampled_source_addr == gpu_addr == face_base(0)` on
  every path that can reach the fallback, and `layers * surface_bytes` covers the decoder's reads to
  the byte.
- **Your `in_mip_tail` scope caveat is sound, and the case it defers is closed tighter than you
  say.** For the live path the reasoning is airtight — `view.in_mip_tail` is assigned only *after*
  the `!layout.supported` early return, so the zero-stride path cannot carry a tail view. Two
  refinements worth folding in: a **replayed** resource can carry both (`gpu_capture.cpp:4548-4556`
  admits `layer_stride_bytes == 0` with `in_mip_tail`), and it still cannot go short for any tiled
  resource, because `sw64kb_tiled_bytes` / `sw4kb_tiled_bytes` round to at least one whole block
  (`tile.cpp:536`, `:368`) — 65536 / 4096 — which is exactly what `mip_tail_bytes` is constrained to
  be. So `surface_bytes >= mip_tail_bytes` holds wherever it matters.
- **No regression is structurally possible from the new branch:** `layers * surface_bytes >=
  surface_bytes`, so the change can only widen. The TR descriptor has a real stride and is untouched
  by it (still 90087424).
- **The new arm discriminates.** `262144 * 256 = 67108864`; a restored single-surface return yields
  262144 and fails it. That is now four of the six arms failing under the mutation instead of three,
  and the one that was pinning the bug is pinning the fix.

And thank you for taking the scope point rather than claiming generality — "a future
tail-plus-zero-stride case would need the tail size passed in" is the right shape for that comment.

### BLOCKING 1 — the function's own leading comment still states the contract you just deleted

`prosper/frontends/shared/texture/texture_decode_cache_policy.hpp:51`

    // signal and must never be widened into a cacheable span; a missing stride or a single layer leaves
    // the surface size unchanged; an overflowing span yields 0 ...

**"a missing stride ... leaves the surface size unchanged" is the pre-fix behaviour.** The code four
lines below now does the opposite, deliberately. The correcting explanation is in the inner comment,
but the leading paragraph is the part a reader skims to learn what the function promises, and it now
promises the defect. This is the same shape as last round's blocking finding — a wrong claim sitting
where it reads as checked — surviving an otherwise complete edit. One line:

    ...; a single layer leaves the surface size unchanged; a missing stride follows the decoder's
    own one-surface-per-layer fallback (see below);  an overflowing span yields 0 ...

### BLOCKING 2 — `## Ruled out` still carries the conclusion this PR falsifies

The charter is explicit that a PR which falsifies a hypothesis records it in `## Ruled out` **before
merging**, and this PR falsifies a large, confidently-measured row that is still standing as current:

- `prosper/docs/TOMB_RAIDER_STATUS.md:272` — *"**The interior's wrong textures are not prosper
  mis-reading the atlas.**"* This PR's control table says they were exactly that.
- `:283` — *"So the guest never fills **slice 248**, at any point — **which retires the decode-cache
  hypothesis properly**."* The decode-cache hypothesis is what this PR just confirmed with a
  one-`false &&` A/B.
- `:297` — *"**The missing piece is an atlas upload path prosper does not perform**, not the array
  plumbing."* There is no missing upload path; the next agent who reads this row will go build one.
- `:213` — *"the interior's own slices 186-248, buckets 249.9-333.2 ... **remain unwritten**"* sits
  badly beside the reinstated 93%: 93% of 344 buckets is ~320, so slices 186-243 are inside the
  written region by gameplay time.

**Do not delete these — scope them, the way you scoped the bucket-200 row.** The narrow measurements
survive and are worth keeping. Specifically, **slice 248 itself is not contradicted by the 93%
figure**: slice 248 starts at byte 87,359,488, i.e. bucket **333.2 of 344**, which falls inside the
7% that is still zero. So "slice 248 reads zero throughout" and "the atlas reaches 93% by gameplay"
are both true, and the row that has to change is the *conclusion drawn from* slice 248 — that the
decode cache was retired and an upload path was missing — not the measurement itself. That is the
same distinction your new rule names, applied one row further down.

### Should fix while you are in there — the un-withdrawal reaches slightly past its evidence

`:92`: *"which is what that number always measured: **guest memory at gameplay time**"*. You
re-derived the figure independently, which is exactly what the withdrawn note demanded ("re-derive it
or leave it out") — but the original's method was recorded as unestablished, and agreement to three
significant figures is strong evidence about the *number*, not proof of the *method*. The claim that
survives without qualification is: "a gameplay-time `.prgbundle` independently reproduces 84 MB, so
the figure is correct as a statement about gameplay-time guest memory; whether `227f5fbe` derived it
that way remains unestablished." Worth the half-sentence, because this is a correction that *raises*
a claim, which is the kind that gets checked least.

The rest of that entry is the best thing in the diff. "When two measurements of the same thing
disagree by 20x, the first question is what each one measured, not which to discard" is a genuinely
transferable rule, the moment-versus-fact framing is exactly right, and leaving **over-bright**
explicitly open-and-unverified rather than folding it into the resolution is the honest call — the
frame does look naturally lit, and nobody has measured it, and those are two different sentences.

### Your two questions: neither should gate

1. **The ~344x retained validation copy — do not gate, and the PR body is the right place for it.**
   It is a real surface (`cached.source_prefix` goes from 256 KiB to 86 MiB and counts in `bytes()`
   against the budget), but the budget floor is >= 1024 MiB against a ~346 MiB entry, `safe_equal`
   compares in place rather than through a scratch copy, and an 86 MiB source is far over
   `cross_submit_watch_min_bytes`, so it takes the write-watch path rather than a per-frame compare.
   Named as a risk is proportionate; a gate would be over-engineering against a cost you have bounded.
2. **The ~23 MiB of over-coverage — do not gate, and I would not change it.** It errs in the only
   safe direction, it is the same trade `layered_cube_source_size` already makes by validating a
   descriptor footprint including alignment gaps, and the alternative (a span that tracks which mip
   the decode actually reads) buys ~26% fewer invalidations in exchange for a second layout model
   that can drift from the decoder. The whole defect here was two models disagreeing. Keep one.

### Verdict

**REJECTED**, on the two documentation items only. To be unambiguous about what that does and does
not mean: the code at `25bb49cd` is correct, the fail-closed and overflow behaviour is right, the
zero-stride span is exact rather than merely conservative, the test arms discriminate, the blast
radius is what the PR claims, the evidence is genuine, and rung 3 is not overstated. Fix the leading
comment at `texture_decode_cache_policy.hpp:51`, scope the four `## Ruled out` lines, and I will
approve without re-verifying the code.

mattias800 (commented):
## Re-review at `0252d787` — APPROVED

Both blocking items are fixed, the should-fix is taken, and I promised to approve on sight of the doc
edits. **APPROVED.** One non-blocking correction below that you can land in this PR or straight
after — it does not gate, and I am not re-gating on it.

### Verified at this head

- **The code really is comment-only since `25bb49cd`.** `git diff 25bb49cd..0252d787 -- '*.hpp'
  '*.cpp'` is the six-line comment hunk on `texture_decode_cache_policy.hpp:50-53` and nothing else,
  so my `25bb49cd` verification carries forward intact.
- **The leading comment is now correct**, and correct in the strong form: `:52` says a missing stride
  *does not* leave the size unchanged because the decoder synthesizes one. That is the contract the
  code implements.
- **All four `## Ruled out` claims are scoped, not deleted, and the convention is right** — strike the
  inference, keep the measurement, say which is which. `:282` and `:289-293` and `:307-311` each
  retain the instrument, its control and its number while marking the conclusion falsified. The
  SLICEWATCH bullet in particular now says the observation "is sound and still holds" while "it did
  NOT retire the decode-cache hypothesis, though this line claimed it did" — that is the version a
  next agent can actually use.
- **Trap-41 check clean.** Against `origin/master` the PR deletes **zero** lines of `.cpp`/`.hpp`, and
  every deleted `.md` line is one of the paragraphs this PR deliberately rewrites. Nothing of another
  lane's went with it.
- `## Open defects` renumbers cleanly 1-4 with the resolved item retained as a struck heading, which
  keeps the issue number findable instead of dropping the history.

### Non-blocking — one clause still says the plural where the evidence is singular

`prosper/docs/TOMB_RAIDER_STATUS.md:282`:

> **The slices the interior samples are genuinely empty in GUEST memory for the whole run** remains
> TRUE and is not in conflict with the 93% figure: slice 248 begins at bucket 333.2 of 344, inside
> the ~7% that stays zero.

The reason given establishes **slice 248**, singular. It does not establish the plural clause it is
attached to: at 93% of 344 buckets the written region reaches bucket ~320, i.e. slice ~239, so slices
186-239 of that range *are* written by gameplay time. Your own `:217` says exactly this — *"Slice 248
specifically stays unwritten all run; the range as a whole is not"* — so the file states both
versions, twenty lines apart, and the weaker one is the bolded one inside `## Ruled out`.

Suggested harmonization, keeping your structure:

> ~~**The slices the interior samples are genuinely empty in GUEST memory for the whole run**~~ —
> narrowed to **slice 248**, which is what was actually measured. It stays zero all run and does not
> conflict with the 93% figure (bucket 333.2 of 344, inside the ~7% that stays zero). The wider
> range is not empty: by gameplay the allocation reaches 93%, so slices 186-~239 are written.

Flagging it rather than waving it through because it is the same distinction you just applied
correctly one row up, and because a bolded sentence in `## Ruled out` is read as established.

### Yes — put the one-model rationale in the header comment

You asked. It should go in, and next to `layered_array_source_size` rather than in a status doc,
because the person who will be tempted to "tighten" the span is someone editing that function.
Something like:

    // The span deliberately over-approximates: it includes each layer's lower-mip bytes, which the
    // decode does not read. Tracking exactly which bytes are read would mean a second layout model
    // beside face_base's, and two models disagreeing about one layout IS this defect (#2998). One
    // model, erring wide, is the trade -- the same one layered_cube_source_size makes by spanning a
    // descriptor footprint including its alignment gaps.

That also pre-answers the "why is this validating 23 MiB it never reads" question before someone
opens an issue for it.

### On your own observation

*"Twice now the code was corrected and the sentence above it was not"* is the sharpest thing in this
thread, and it generalizes past this PR: an edit's blast radius includes the prose that describes it,
and prose has no compiler. I would **not** give it an instrument-trap number — the numbered list is
for apparatus phantoms, and this is a documentation-decay failure that already has a home in the
charter's "a cited claim reads as a verified one" family. If you want it recorded, a line in that
section of `CLAUDE.md` would reach more readers than a trap row, and it would be a separate docs-only
PR rather than cargo on this one.

### Verdict

**APPROVED** at `0252d787`. The code is correct — fail-closed on a zero `surface_bytes`, well-defined
overflow with the synthesized stride, a span that is exact rather than merely conservative on every
path that can reach the fallback, discriminating test arms, and a blast radius confined to
`guest_array`. The evidence is genuine, rung 3 is not overstated, and the documentation now says what
the code does.

`gris-gameplay` still running does not hold this: per the charter snapshots are a release-time
inventory rather than a merge gate, three guards are green, and this change cannot alter a title with
no BC 2D_ARRAY. Post it either way as you said you would — if it regresses, that is a finding about
this change worth having in the record, not a reason to have waited.

PR #3019, opened 2026-08-26
fix(app): the --fps HUD reports a real rate under GPU present, instead of "no frames published yet"

mattias800:
Fixes #3010.

## Failure scenario

`prosper-app --dump <dump> --fps` on any real game. The title boots, presents thousands of frames,
stderr reports `[app] 64.0 fps (7560 frames, gpu-present)` — and the on-screen HUD reads
`-- fps` / `no frames published yet` for the entire run. Reproduced on Windows/NVIDIA with *The
Messenger* and *Grand Theft Auto V*; the cause is platform-independent and this was the **default path
for every game**, since `request_gpu_present()` returns true for any boot with a dump.

## Behavioural contract

The HUD's headline is deliberately the **distinct** (content-changed) rate, not the presented rate —
`fps_hud.hpp` argues this at length, because a title whose picture is frozen still presents at ~60
(the R-Type Delta shape, #2783, instrument trap 90). That contract was intact; its **feed** was not.

## Cause

`note_present_publication` had exactly one caller: `present_write_frame`
(`videoout_present.cpp:53`). GPU present (#1270) bypasses it at both ends —

- `live_renderer.cpp:8910` *"SKIP\[s\] the CPU readback+reupload entirely"* when the front-buffer blit
  succeeds, so no CPU frame is published;
- `main.cpp`'s `if (vk.gpu_present)` branch blits to the swapchain directly and does not call it either.

So `FrameRate::measured` stayed false forever. The app's stderr line was unaffected because it counts
with its own local counter — hence one run reporting two contradictory things.

## Why no test caught it

`live_renderer.cpp:8904` says it outright: `gpu_present_active()` is *"false in every
headless/test/screenshot process"*. `tools/screenshot`, the snapshot guards and the core suite all
exercise the CPU path, where the counter works correctly. `test_fps_hud.cpp` asserts the pure decision
function — including the `no frames published yet` string — and passes. **The decision was covered; the
feed was not.** The only configuration on the broken path is the interactive app, which is also the
only place a human reads this HUD.

## Approach

The GPU path samples the presented image itself: a **256x144 `VK_FILTER_NEAREST`** reduction of the
same image the swapchain blit just read, recorded in the same command buffer, copied to host memory,
and signed at the top of the **next** present — under the fence `present_frame_gpu` already waits on,
so **no synchronisation is added**. NEAREST, not LINEAR: averaging lets a small bright change cancel
against its neighbours, and this exists to detect change.

[...]

Issue #2783, opened 2026-08-20
R-Type Delta (PPSA26414): every frame after the title screen is dropped on master — the guest reaches stage 1 and prosper publishes only the retained frame


mattias800:
## Symptom

On current master (`82baa409`), *R-Type Delta: HD Boosted* (`PPSA26414`) renders its publisher logos,
its opening movie and its title screen, and then **stops producing any rendered frame at all**. The
guest keeps running — it keeps flipping, it keeps loading its per-scene PRXs, it reaches stage 1 — but
prosper publishes the same retained frame from that moment on.

At `83e3383a` (the #2061 merge, i.e. the head at which the title was recorded at rung 2 on
2026-08-06) the identical route on the identical dump renders the whole flow, **including stage 1
gameplay**.

## Reproduction

Both arms are `tools/screenshot`, identical command, identical dump, identical route file, unmodified
binary, unmodified guest, page cache evicted with `tools/dropcache.py` immediately before each launch:

```
PROSPER_GUEST_ARGS=-force-gfx-direct PROSPER_RENDER=1 \
  PROSPER_SAVE0=<fresh dir> PROSPER_MODLOG=1 PROSPER_PAD_SCRIPT_LOG=1 \
  PROSPER_PAD_SCRIPT=@prosper/scripts/rtype-delta-PPSA26414/explore-v2.pad \
  ./build/screenshot <DUMP_ROOT>/PPSA26414-app0 --seconds 2 --count 60 --out <dir>
```

The route is `prosper/scripts/rtype-delta-PPSA26414/` (added alongside this issue). `PROSPER_SAVE0`
must be a fresh directory — the route is save-state dependent.

## Evidence

`nonblack_rgb_pixels` from each run's own capture manifest, per sample:

| t (s) | `82baa409` (master) | `83e3383a` (#2061 merge) |
| --- | --- | --- |
| 8 | 1,167,360 (opening movie) | 1,167,277 (opening movie) |
| 14 | **1,571** | 2,073,176 (main menu) |
| 20 | **2,728** | 2,073,474 |
| 30 | **1,637** | 2,070,351 |
| 40 | **1,637** | 2,070,351 |
| 50 | **1,637** | 2,073,209 |
| 60 | **1,571** | 2,072,139 |
| 100 | **1,571** | 1,733,409 (stage 1) |
| 120 | **1,571** | 1,862,995 (stage 1) |

[...]

Issue #1270, opened 2026-07-23
Unify Vulkan device: prosper-app presents the renderer's front-buffer image via GPU blit (eliminate 4K CPU present round-trip)


mattias800:
## Goal

Unify the app frontend's presentation onto the renderer's Vulkan device so `prosper-app` can **GPU-blit the renderer's front-buffer image directly to the swapchain**, eliminating the per-present 4K CPU round-trip (renderer reads the front buffer back to CPU, app re-uploads it to a staging image, app blits to swapchain).

## Why (evidence — see #1265)

Profiling GTA V (PPSA04263) 4K present cost on RADV STRIX_HALO: the shared `live_renderer` reads the front-buffer GPU image back to CPU (`present_write_frame`, CPU `std::vector`), and **both** frontends re-upload it — screenshot→PNG, `prosper-app/main.cpp:280` `present_frame`→staging→swapchain blit. At 4K that's ~3ms readback + ~2ms re-upload ≈ **5ms/present of pure round-trip**, paid on every present. On an integrated (unified-memory) GPU this is especially wasteful — it's the same physical RAM copied GPU→CPU→GPU.

The root obstruction (confirmed): the app's swapchain `VkDevice` and the renderer's persistent-image `VkDevice` are **separate devices from separate instances by design** (docs/FRONTEND_APP.md, "Two Vulkan contexts by design"). The compute backend already adopts the renderer's device via `SharedVulkanContext`; the app does not. Unifying the app's presentation onto the renderer's device makes a direct `vkCmdBlitImage(front-buffer image → swapchain image)` possible.

## Approach (A1 — unify devices), staged

- **Stage 1 (zero behavior change):** `render_vk_ctx()` (`tests/render_runner.h:448`) creates its instance with all *available* surface instance-extensions and its device with `VK_KHR_SWAPCHAIN` + a second queue — each only if available (device-swapchain gated on instance-surface; 2nd queue gated on `queueFamilyProperties[qfi].queueCount>=2`). Publish present-capability + the present queue in `SharedVulkanContext` (`src/gpu/gpu_execute.hpp:478`). Nothing consumes it yet.
- **Stage 2:** the app adopts the shared context for presentation — SDL surface on the shared instance, `vkGetPhysicalDeviceSurfaceSupportKHR` on the shared family, swapchain on the shared device/second queue; **fall back** to today's separate-device CPU path when present is unsupported or the shared context lacks present capability.
[...]

Issue #3076, opened 2026-08-27
Every FPS figure measured through tools/screenshot understates the shipped renderer — The Messenger is 100-160 fps, not 24


mattias800:
`tools/screenshot` **never calls `set_gpu_present_active`**. With GPU present inactive the renderer is
required to copy every scanout frame back to the CPU — the code that enables it says so directly:

> *"The renderer must now publish the front-buffer image (**and skip the CPU readback**)"*
> — `frontends/prosper-app/main.cpp`, at `set_gpu_present_active(true)`

So a `tools/screenshot` run does not merely observe the renderer more slowly; it measures **a
different architecture** from the one a user runs.

## The size of the error

*The Messenger* (`PPSA24651`), first level, same committed route
(`scripts/messenger/reach-first-level.pad`), measured 2026-08-27 on master:

| harness | GPU present | FPS |
| --- | --- | --- |
| the July pass, via `tools/screenshot` | inactive (never enabled) | **12 – 24** |
| `prosper-app --present-mode immediate` | `[app] GPU present: adopted` | **100 – 160** |

Visuals verified at the same time — the opening scene renders correctly (sunset, clouds, the great
tree, ocean, the Messenger on his platform, dialogue box with portrait and correctly coloured text).
Screenshot in `BLOG.md`.

That figure is quoted in **CLAUDE.md**, where it has been justifying "do not resume Messenger
renderer cache work toward 60 FPS" — correct advice reached through a wrong number, and the kind of
stale figure the charter itself warns carries authority it no longer deserves.

## Scope of the contamination

`docs/RENDERER_PERFORMANCE_2026_07.md` mentions `screenshot` nine times against two mentions of
`prosper-app`/GPU present, so a large share of the July pass is affected. **Every FPS and
frame-budget figure in that document should be assumed to describe the forced-readback path until
re-measured.** I have not re-measured them; that is a bigger job than this issue.

Two effects, both measured on GTA V:

- **Forced scanout readback.** Headless: 29,600 non-deferred readbacks, 936 GB. Windowed: **0**.
- **~30x inflated compute-image snapshot traffic.** Headless 8,407 copies / 80.5 GB, windowed
  343 / 2.7 GB — because the forced readbacks feed the image-cache validation path.

## It is NOT a uniform penalty — do not apply a correction factor

*Bendy and the Ink Machine* measures **9 fps windowed** with GPU present adopted, matching its
[...]

PR #3082, opened 2026-08-27
docs: the Messenger's first level is ~156 fps windowed, and the 24 FPS figure predates #1270

mattias800:
Corrects a stale figure in `CLAUDE.md` — and, across two review rounds, corrected this PR's first attempt at correcting it and then its own over-correction.

## What was wrong with the charter

`CLAUDE.md` has long said the July pass "deliberately stopped at roughly 12 → 24 FPS on the first level" of *The Messenger*. That figure is from 2026-07-14 and no longer describes the shipped renderer: **#1270 replaced the present architecture on 2026-07-24**, ten days later. `prosper-app` now blits the renderer's front-buffer image straight to the swapchain and the renderer skips the CPU readback entirely.

**It was not measured badly.** `docs/RENDERER_PERFORMANCE_2026_07.md`'s own `## Measurement method` names `run-windows.ps1` → `prosper-app`, and `git blame` puts the figure and that method block in the same commit (`93d4dae4`). At the time there was no GPU-present path for any harness to miss. An earlier revision of this PR blamed `tools/screenshot` for the figure; that was false and is retracted.

## The measurement

*The Messenger*, first level, windowed, `--present-mode immediate`, `[app] GPU present: adopted`, 2026-08-27:

| | |
| --- | --- |
| samples | 468 |
| median | **156.4 fps** |
| p25 / p75 | 130.9 / 165.7 |
| min / max | 42.9 / 454.8 |
| inside 100–160 | 313 of 468 (66.9%) |

**It is a `presented` rate with no `distinct` companion, so it is an upper bound on the rate of new frames, not the frame rate.** The renderer republishes its retained frame when a submit produces no present source — that is how a presented rate reads full speed for a frozen title, and it is why #2783 hid for nine days. #3083 asks for the distinct rate and records the `PROSPER_APP_GPU_PRESENT` A/B that quantifies the readback directly.

## The harness rule, date-bound

- **`tools/screenshot` never calls `set_gpu_present_active`** — sole call site `frontends/prosper-app/main.cpp:668` — so in that harness the renderer copies every scanout frame back to the CPU and the rate understates. **This applies only to figures measured after 2026-07-24**; before #1270 every harness took the readback, so older figures honestly measure what then shipped.
- **A `fifo` run measures the display.** Worked example, from this PR's own mistake: a vsync-locked run of this same title pinned 123 of 238 samples at exactly `180.0` — the host refresh rate. Say which present mode a run used before quoting its rate as a throughput.
- The interactive app could not report a distinct rate under GPU present at all until **#3019** (2026-08-26), so an interactive rate from before that date is a publication count.

## Files

- `CLAUDE.md` — bullet corrected and split; `PPSA03416` steer restored intact; harness rule date-bound.
- `docs/RENDERER_PERFORMANCE_2026_07.md` — caveat under the figure, so the charter's pointer no longer lands readers on an uncaveated 24 FPS.
- `src/gpu/present/present_frame_rate.hpp` — comment-only. It cited the July figure as a contradiction to be explained; it now points at the corrected attribution. Left uncorrected, it is the next place the stale number gets picked up.
- `BLOG.md` — one sentence plus the picture, folded under the existing `## 2026-08-27` heading.
- `assets/screenshots/messenger-first-level-windowed-2026-08-27.png` — from the `immediate` run above. `SCREENSHOTS.md` deliberately untouched (it describes `origin/master`).

## Verification

Docs gate green on all three `.md` files; `git diff --check` clean; blockquote termination checked after the lazy-continuation fix. The `.hpp` change is comment-only but takes the PR out of the docs-only exception, so it waits for full CI.

Refs #3083


mattias800 (commented):
**REJECTED** — and I want to be clear about why, because I think the underlying finding is probably
real and worth landing. What I am rejecting is the causal story this PR writes into `CLAUDE.md`, and
the evidence offered for the magnitude. Both are fixable; none of this needs a new measurement rig.

This is the highest-authority file in the repo, and the charter's own rule applies to itself:
*"a stale figure carries the charter's authority, which is exactly the authority it no longer
deserves."* A **wrong** figure with a **wrong reason** carries it further.

---

## What I verified and confirm is correct

- `tools/screenshot` genuinely never calls `set_gpu_present_active`. Repo-wide there is exactly one
  call site: `prosper/frontends/prosper-app/main.cpp:590`. Declaration
  `prosper/src/gpu/execute/gpu_execute.hpp:1211`, definition
  `prosper/src/gpu/execute/gpu_executor.cpp:11230`. Nothing under `prosper/tools/screenshot/` reaches it.
- The readback-skip mechanism is real and does what the PR says. `prosper/frontends/shared/live/live_renderer.cpp:8041`
  gates the GPU publish on `gpu_present_active()`, and the comment at `:8032-8038` states the
  consequence in the project's own words: *"gpu_present_active() is false in every headless/test/
  screenshot process, so this whole branch is inert there and the CPU path below is byte-for-byte
  unchanged."* The deferral half is `live_renderer.cpp:6989` feeding `final_gpu_present` into
  `prosper/frontends/shared/present/readback_policy.hpp:14`.
- The screenshot is genuine and good. 1920x1080, the opening scene renders correctly — sunset, clouds,
  the great tree, the platform, the dialogue box with a correctly coloured portrait and text. No
  complaint there at all.
- Diff really is docs plus one PNG, CI is 9/9 green, and you correctly left the screenshot feed alone.

So the mechanism is not in dispute. Everything below is about what follows from it.

---

## BLOCKING 1 — the attribution is falsified by the document the PR cites

The PR infers the harness from a mention count: *"9 mentions of `screenshot` against 2 of
`prosper-app`/GPU present"*. **That document names its harness for this exact figure, and it is not
`tools/screenshot`.**

`prosper/docs/RENDERER_PERFORMANCE_2026_07.md:18-19` is the 12-24 claim. Four lines below it,
`:23-30`, under the heading `## Measurement method`:

> Use a `RelWithDebInfo` native Windows build, a fresh process for each A/B, the same input route and
> save data ... `./prosper/scripts/run-windows.ps1 ./PPSA24651-app0 -NoBuild`

`prosper/scripts/run-windows.ps1:78` sets `$app = prosper-app.exe`, `:86` configures `-DPROSPER_APP=ON`,
and `:128` launches it. That is the harness that *does* call `set_gpu_present_active`. The doc's own
line `:40` says *"App FPS is useful as the outcome"* — app, not shot.

`git blame` puts `:18-19` and the whole `## Measurement method` block in the **same commit**,
`93d4dae4`, 2026-07-14. They were written together.

And the mention count is over the wrong part of the file. All nine `screenshot` mentions are at
line **334 and later** — every one of them in sections appended in August about Blue Prince, Dead
Cells and Evergate. The Messenger section, lines 1-70, contains the string zero times.

This is exactly the shape the charter names: *"a true statement reached by a route that does not
establish it."* The route here does not survive opening the file.

## BLOCKING 2 — the mechanism could not have applied in July, and the CLAUDE.md generalization is false for pre-#1270 figures

`set_gpu_present_active` was introduced on **2026-07-24** — `6759b671` ("present the renderer's
front-buffer image via GPU blit", #1270 stages 2+3) and `d8c503a2` ("publish the front-buffer image +
skip the readback in GPU-present mode", #1270 stage 3b). The performance doc is dated **2026-07-14**.

Ten days earlier. **In July there was no GPU-present path in any harness.** Every run — `prosper-app`
windowed included — took the CPU readback, because it was the only path there was.

So the 12-24 figure was a *correct* measurement of the *then-shipped* renderer. It is stale because
**the renderer changed** (#1270 removed the scanout readback for windowed runs), not because somebody
used the wrong tool. Same magnitude, completely different lesson — and the lesson is the thing going
into the charter.

That makes the new closing sentence wrong as written:

> A figure from `tools/screenshot` or any `SDL_VIDEODRIVER=offscreen` run describes the forced-readback
> architecture, **not the shipped one**.

For anything measured before 2026-07-24 the forced-readback path *was* the shipped one. As phrased,
this instructs every future reader to discount a large body of correct historical measurements for a
reason that did not exist when they were taken. Either date-bound it ("since #1270, 2026-07-24, ...")
or drop the second clause.

## BLOCKING 3 — 100-160 is a number from the counter this project rejected six days ago, and the right counter is structurally blind in the mode you measured

`prosper/src/gpu/present/present_frame_rate.hpp:1-28` exists because a present-keyed fps *"is WRONG in
the one case anybody needs a framerate for, and wrong in the direction that hides the defect — it
reads FULL SPEED for a completely frozen title."* That landed 2026-08-21 (#2843) and was blogged on
2026-08-26 — the entry immediately below yours on master.

Now look at the counter the PR quotes. `prosper-app`'s GPU-present rate line,
`prosper/frontends/prosper-app/main.cpp:2430-2437`, counts `shown`, gated at `:2409` by
`present_source_is_newer(..., gf.frame_seq)`. `gf.frame_seq` is `prosper_vo_flip_count()` — **guest
flips**, not distinct content. Better than raw presents; not the metric #2843 built.

And you cannot easily get the right one in that mode: `note_present_publication` is called from
exactly one place, `prosper/src/gpu/present/videoout_present.cpp:53`, inside `present_write_frame` —
the **CPU** publication path. GPU present skips that path by construction. So in GPU-present mode the
distinct-content counter is fed almost nothing, and `--fps`'s headline has no population behind it.

Two consequences, and the second is the reason this is blocking rather than pedantry:

1. The 100-160 figure is a guest-flip rate. On a route whose measured frame is a **dialogue box over
   a near-static scene** — which is what your screenshot shows — flip rate and picture-change rate
   are exactly the pair most likely to diverge.
2. **The repo already holds a competing, six-day-old measurement of the same title on the same route,
   and the PR does not mention it.** `prosper/tools/screenshot/README.md:133-137`, 2026-08-21,
   `scripts/messenger/reach-first-level.pad`, native 1080p: `distinct 3.0 fps / presented 207.2 fps
   over 380.0 s`, with the animated stretches at **~15-23 fps distinct**. The same figure is restated
   in `prosper/tools/docs/gen_progress_tracker.py:134-138` and rendered into
   `PROGRESS_TRACKER.md:75-79`, both of which cite the July 12-24 figure **approvingly**, as the
   benchmark that proves 3.0 would have been a phantom regression.

   Note also `presented 207.2 fps` — through the supposedly readback-crippled harness. Whatever the
   readback costs, it was not holding publication throughput to 24/s.

Merge this as written and `CLAUDE.md` says 100-160 while three other in-tree records say ~15-23 and
treat 12-24 as sound. None of them points at the other. That is a worse state than the one being
corrected.

**What would settle it:** report the pair. If the distinct rate is genuinely unobtainable under GPU
present, say so explicitly and argue why the flip-keyed number is trustworthy *here* — and please
open an issue for the gap, because "the distinct-frame counter does not work in the mode we ship" is
worth its own fix.

## BLOCKING 4 — a four-variable comparison stated as a one-cause conclusion, when the one-variable A/B is one environment variable away

The table's two rows differ in at least four things:

| | July row | new row |
|---|---|---|
| OS / host | native Windows, MinGW RelWithDebInfo | presumably Linux, unstated |
| present mode | `fifo` (`run-windows.ps1:14-15` default) | `immediate` |
| renderer | pre-#1270 | post-#1270, plus six weeks |
| scene | *"the representative heavy frame ... four compute dispatches interleaved with four graphics spans"* (`RENDERER_PERFORMANCE_2026_07.md:16-17`) | opening dialogue |

`n=1` per arm, and the whole delta is assigned to one of the four.

`PROSPER_APP_GPU_PRESENT=0` is the explicit A/B switch — `prosper/frontends/prosper-app/present_policy.hpp:18-24`,
unit-tested at `prosper/frontends/prosper-app/test_present_policy.cpp:68`, and its own comment calls
it *"the explicit recovery/A-B switch"*. Same binary, same host, same route, same present mode, GPU
present on versus off. That is the experiment, and it costs one environment variable.

On present mode specifically: `immediate` removes vsync. Under the `fifo` default a user is capped at
refresh, so "100-160 FPS" is not what a default launch shows. It does not sink the headroom claim, but
the table must say `immediate` in both the PR and the charter bullet, or it reads as a like-for-like
comparison it is not.

## BLOCKING 5 — the doc the bullet still points at goes uncorrected

The rewritten bullet still ends with `docs/RENDERER_PERFORMANCE_2026_07.md`, and that file's
`:18-19` still says *"roughly 24 FPS ... up from approximately 12 FPS"* with no caveat. A reader who
follows the charter's own pointer lands on the number the charter just told them was wrong.

The reasoning for not rewriting the whole document is right and I would not want you to. But the
charter's recording rule is not satisfied by a PR body: *"A falsification that lives only in a PR body
or an issue comment is one the next agent will re-derive at full cost."* One banner line under the
headline, pointing at #3076, discharges it.

## BLOCKING 6 — the rewrite drops the bullet's only actionable steer

Original:

> the remaining synchronous graphics/compute boundaries must be evaluated against a **3D** workload
> first (*Summer Sports Games* `PPSA03416` is the first clean candidate)

New: the same sentence with `PPSA03416` deleted and *"GTA V 3.6 FPS, Bendy 9 FPS"* in its place.

That is a net loss. The named candidate was the actionable half — it told the next agent where to
start. The two replacements are not clean performance subjects: the same file's own table says GTA V's
world does not render at all (rung 2), and Bendy is rung 3 with open blockers. Optimising against a
title that is not drawing the scene is how you measure the wrong thing. Keep `PPSA03416`; the two new
numbers are fine *beside* it, not instead of it.

---

## Non-blocking

- **The non-uniformity claim rests on a baseline that disclaims itself.** Bendy's *"recorded ~8"* is
  #1881, which says: *"carries an obsolete 0.75-fps figure; current shared-GPU runs are around 8 fps
  in the level and need an isolated remeasurement."* Self-declared provisional, and taken on a
  **contended** GPU. 9-versus-8 between a fresh number and that is not a demonstrated agreement, and
  one point would be thin for "not uniform" even if it were. That said — the conclusion you draw
  (*re-measure, do not scale*) is the conservative direction and is right regardless, and I am glad it
  is in the issue rather than the charter. Just do not present it as established.
- **BLOG.md: it is a report wearing a blog entry's clothes, mildly.** The rule is one sentence on what
  you are looking at, and *"measurements, bisects, censuses, mechanisms, reasoning ... Link, never
  transcribe."* You have the good one-sentence lead — *"Its first level, on current master, in a normal
  window."* — and then a three-sentence paragraph transcribing the mechanism and the number. Cut it to
  one sentence plus the #3076 link. The multi-title exception does not apply here.
- **BLOG.md: duplicate date heading.** The branch adds a second `## 2026-08-27` at BLOG.md:51 while one
  already exists at :22, and master has entries under that first one. Rebase and fold the entry under
  the existing heading.
- **Put the number in the picture.** `prosper-app --fps` exists as of #2843 and the charter explicitly
  blesses it: *"Annotation that prosper itself draws over its own output is fine and often better — an
  fps counter."* For a PR whose entire claim is a framerate, a capture with the HUD burned in is
  strictly stronger evidence than the same capture without it — and it would answer BLOCKING 3 in the
  same shot. (If the HUD reads `--` under GPU present, that is itself the finding.)
- Minor: the PR body says *"same committed route"*, which is true, but the July doc measured a named
  *scene* — the representative heavy frame — and a route reaching the same level is not the same scene.
  Worth one clause.

---

## What I would approve

I think you have found something real, and the better comparison is already sitting in your own
evidence: **~15-23 fps distinct through `tools/screenshot` on 2026-08-21 versus your `prosper-app` run
on 2026-08-27**, same route, six days apart on the same era of master. That is far closer to a clean
A/B than anything involving July, and if it holds up under `PROSPER_APP_GPU_PRESENT=0/1` it is a
genuinely good result that deserves the charter.

Concretely:

1. Re-attribute: the July figure was measured through `prosper-app` on native Windows and was correct
   for its build; it is stale because **#1270 (2026-07-24) removed the scanout CPU readback**. Drop
   the `tools/screenshot` attribution for that figure.
2. Date-bound the "check the harness" rule to post-#1270, or drop "not the shipped one".
3. Run the one-variable A/B with `PROSPER_APP_GPU_PRESENT`, same host, same present mode, and report
   the pair. State the present mode in the table.
4. Report distinct alongside presented, or state plainly that distinct is unobtainable under GPU
   present and why — and file that gap.
5. Reconcile with `present_frame_rate.hpp:33-39`, `gen_progress_tracker.py:134-138` and
   `PROGRESS_TRACKER.md:75-79`, all of which currently endorse 12-24.
6. Add one caveat line to `RENDERER_PERFORMANCE_2026_07.md:18-19` pointing at #3076.
7. Restore `PPSA03416` as the named 3D candidate.
8. Trim the blog entry to one sentence and fold it under the existing `## 2026-08-27` heading.

Do 1, 2, 6 and 7 and I would approve on the strength of the mechanism alone with the magnitude stated
as provisional. Do 3 and 4 as well and the number itself is solid.

mattias800:
You were right, and blocking 1 and 2 are the important ones — I verified both against the source rather than against your citation, and they hold exactly as stated.

`docs/RENDERER_PERFORMANCE_2026_07.md:23-30` names `run-windows.ps1`; `git blame` puts the figure and that method block in the same commit, `93d4dae4`, 2026-07-14; `set_gpu_present_active` first appears in `6759b671` / `d8c503a2`, 2026-07-24. So the July pass measured with the app, ten days before the mechanism I blamed existed. My causal story was false, and the corrected one is yours: the figure is stale because **#1270 changed the present architecture underneath it**, not because the wrong tool was used. That also date-bounds the rule, which is now written into the bullet.

**Then I checked my own number and it is worse than you thought.** You flagged the four-variable comparison; there is a fifth variable I introduced and then misdescribed. The PR body said `--present-mode immediate`. The run log says:

```
[app] present mode: fifo
```

123 of its 238 `[app] … fps` samples sit at exactly `180.0` — the host display's refresh rate. **The run measured the monitor.** So "100-160" is withdrawn; I cannot reconstruct where in that log it came from, and the log's own distribution (median 180.0, p25 162.3, min 76.5) does not support it. What the run supports is that the first level *holds 180 Hz*, which is a ceiling. That is now how it is stated, in all three files.

Filed **#3083** for the measurement that settles it — your `PROSPER_APP_GPU_PRESENT` A/B, with `--present-mode immediate`, same build and route — and it carries the competing `distinct 3.0 / presented 207.2` figure from `screenshot/README.md:134-137` as something to reconcile rather than ignore. Five `prosper-app` instances from another lane have had the GPU for five hours, so I have not run it; the PR no longer depends on it.

**On blocking 3, one correction in the other direction.** `note_present_publication` is not the only feeder any more. `present_frame_rate.hpp:346-352` documents a second path added for exactly this gap: under GPU present the app samples the presented image into a small host-visible buffer, signs it densely, and calls `note_present_publication_signature`. That landed in `80a619eb` on **2026-08-26** (#3019) — the day before my run — so the HUD's distinct rate does have a population on current master. Your finding was correct for every interactive rate quoted before that date, which is worth having, so it is in the charter bullet as a dated caveat rather than dropped.

Also fixed: **5** (caveat block added under the July figure so the charter's pointer no longer lands on an uncaveated 24 FPS), **6** (`PPSA03416` restored — you were right that replacing it with two titles that do not render their worlds removed the only actionable steer), the duplicate `## 2026-08-27` heading, and the BLOG entry cut back to the one-sentence rule.

On the Bendy baseline: agreed, and I have dropped the comparison entirely rather than defend it — #1881 self-describes as needing an isolated remeasurement, so 9-vs-8 was never an agreement.

Head is `d47c7b9f`.

mattias800 (commented):
**REJECTED** — and I am sorry to say it, because the half of this PR that I asked for is now right,
and the half you rewrote on your own initiative has gone wrong in a new way that is worse than what
it replaced.

**First, I was wrong and you were right on BLOCKING 3. I withdraw it.** `note_present_publication`
is no longer the only feeder. `prosper/src/gpu/present/present_frame_rate.hpp:346-369` documents the
GPU-present path explicitly — *"That path feeds these two instead: it samples the presented image
into a small host-visible buffer on the GPU, signs the sample DENSELY, and publishes the
signature"* — declaring `dense_content_signature` at `:371` and `note_present_publication_signature`
at `:375`, with the live call site at `prosper/frontends/prosper-app/main.cpp:701-702` and unit
coverage at `prosper/tests/gpu/present/test_present_frame_rate.cpp:380-384`. It landed in
`80a619eb` (#3019). My claim was true of master on 2026-08-21 and false of master on 2026-08-26, and
I asserted it as though it were timeless. Your dated caveat in the charter is the right disposition.

---

## What I re-verified independently and confirm

The re-attribution — the thing this PR is now for — is **factually correct**, and I checked it
against the source rather than against my own prior review:

- `git blame -L 14,31` on `prosper/docs/RENDERER_PERFORMANCE_2026_07.md` puts the 12/24 figure and
  the whole `## Measurement method` block in **one commit**, `93d4dae4`, `2026-07-14 16:14:05 -0400`.
- At that commit `prosper/scripts/run-windows.ps1` launches `prosper-app.exe` (`:71` sets it, `:79`
  configures `-DPROSPER_APP=ON`, `:93` builds that target). It is the app, not `tools/screenshot`.
- `git log -S` over `prosper/src` and `prosper/frontends` returns **nothing** for either
  `set_gpu_present_active` or `gpu_present_active` before `6759b671` / `d8c503a2`, both
  `2026-07-24 08:53:18 +0200`.

So: correct measurement, ten days before the mechanism existed, invalidated by #1270 rather than by
tool choice. The date bound *"only to figures measured after 2026-07-24"* is the right boundary and
a reader will apply it correctly.

Also confirmed: `PPSA03416` is restored intact (`CLAUDE.md:166`); `BLOG.md` has exactly one
`## 2026-08-27` heading (`:22`) with the entry folded under it at `:81`; `SCREENSHOTS.md` untouched;
the branch's merge-base **is** `origin/master`; `git diff --check` clean; the docs gate passes on all
three files (33 rows / 0 / 183, exit 0); the screenshot is genuine, 1920x1080, and the opening scene
renders correctly.

Those are real corrections and I would happily merge them alone.

---

## BLOCKING 1 — the number you withdrew was correct. You audited it against the wrong log.

This is the finding, and it is the same failure this PR exists to correct, pointed the other way.

Your comment says: *"I cannot reconstruct where in that log it came from, and the log's own
distribution ... does not support it."* That is true of the log you looked at. It is not the log the
figure came from.

There are **two** runs on disk, and they are not the same run:

| | `~/gta5-lane/messenger-live.log` | `~/messenger-lane/run.log` |
|---|---|---|
| when | **2026-08-26**, 18:07:48 – 18:09:14 +0200 | **2026-08-27**, ~10:20 +0200 |
| present mode | `fifo` (`:30`, `:35`) | **`immediate`** (`:31`, `:37`) |
| samples | 238 | 468 |
| median | 180.0 | **156.4** |
| p25 / p75 | 162.3 / 180.0 | 131.1 / 165.9 |
| in 100-160 | — | **313 of 468 = 66.9%** |
| screenshot | none produced | `frame_grab_PPSA24651_20260827-102010-068.bmp` |

The second run is a single contiguous boot (one `guest booted`, frame counter monotonic 60 → 28080),
and it is the run that produced **this PR's screenshot**. Not "a similar frame" — the same bytes:

    git rev-parse 6f77fa97:assets/screenshots/messenger-first-level-windowed-2026-08-27.png
      -> c7e5b7b3d83ff4081a2ad94fdf3917c120a4879d
    git hash-object <that run's grab.png>
      -> c7e5b7b3d83ff4081a2ad94fdf3917c120a4879d

Byte-identical, and the log records it being written at `[grab] screenshot written (armed at guest
present 65930, written at guest present 66163)`.

Your **original** PR body — recovered from GitHub's `userContentEdits`, so this is not my
reconstruction — said:

> | `prosper-app --present-mode immediate` | `[app] GPU present: adopted` | **100 – 160** |
>
> *The Messenger* (`PPSA24651`), first level, same committed route, same master, 2026-08-27

Every element of that matches the 08-27 run: the harness, the present mode, the date, the screenshot,
and a distribution with two thirds of its mass inside 100-160. **The original claim was internally
consistent and supported by its own evidence.** What happened is that you went to verify it, opened a
different run from the previous evening in a different lane directory, found `fifo`, and concluded
you had misdescribed your own experiment.

And the scene-matched number is better still. Taking the 105 fps samples surrounding the moment the
committed screenshot was written, that window has **median 122.9 fps** (min 42.9, max 202.3), sitting
mostly in the 105-150 band. So for the exact frame the PR shows, uncapped, the honest figure is
**about 120 fps presented** — inside the range you withdrew.

Net effect of the rewrite on the magnitude claim: a defensible uncapped measurement, taken on the
stated date, in the stated mode, evidenced by the attached picture, has been replaced by a
vsync-capped ceiling from an older run that produced no picture at all. That is a regression, and
`CLAUDE.md` is where it lands.

## BLOCKING 2 — the run in the charter is dated wrong, and the picture and the number come from different runs

`CLAUDE.md:170` says *"Re-run windowed on 2026-08-27"*. `RENDERER_PERFORMANCE_2026_07.md:25` says
*"A windowed run of the same title on 2026-08-27"*. The PR body says *"Windowed, current master,
2026-08-27"*. #3083 says *"windowed on current master"*.

The 238-sample `fifo` run is dated **2026-08-26** — birth 18:07:48, last write 18:09:14 +0200, a
~86-second run, and its counts (n=238, 123 at exactly 180.0) match your quoted figures exactly, so
there is no ambiguity about which run is being described.

Two consequences:

1. A PR whose subject is a measurement carrying the wrong date now writes a measurement carrying the
   wrong date into the highest-authority file in the repository, twice.
2. The picture and the number are from **different runs, on different days, in different present
   modes**, presented as one run. In evidence terms that is a composite, and the charter's own
   screenshot rule exists to forbid exactly that framing.

Worth adding: the 08-26 log contains no `[grab]`, no route, no screenshot and no scene marker of any
kind. Nothing in it establishes that the run reached *"the same first level"* at all. The only
artifact that establishes the first level belongs to the other run.

## BLOCKING 3 — "the true rate is somewhere at or above it" is falsified by your own two logs

`CLAUDE.md:172-173`. Under `fifo`, a sample **below** the refresh rate means the app failed to reach
the cap, so the true rate at that moment was below 180. In the very log being quoted:

- 92 of 238 samples (**38.7%**) are below 180.0
- 35 (**14.7%**) are below 160
- min 76.5

The adjacent clause in the same sentence — *"with one dip to 76.5"* — already contradicts it. And the
`immediate` run puts the median at 156.4. The evidence points at roughly 155-180 with the cap only
intermittently binding, not at a floor of 180. As written, a future reader takes away "at least 180
fps", which nothing in this PR supports.

The same phrasing is in #3083 (*"somewhere at or above it, unknown"*) and should move with it. The
doc caveat's wording (*"the real throughput is unmeasured"*) is fine as-is.

## BLOCKING 4 — the new caveat blockquote swallows the doc's own conclusion

`prosper/docs/RENDERER_PERFORMANCE_2026_07.md:28-30`: there is no blank line between the end of the
blockquote and the pre-existing sentence *"The remaining cost is structural, and the next renderer
changes must be tested against a 3D title before choosing a resource-lifetime or scheduling
architecture."* CommonMark lazy continuation absorbs it into the quote.

Not a guess — rendered through GitHub's own markdown endpoint, that sentence comes back **inside**
the `<blockquote>`. So the document's actionable 3D steer — the same steer you fought to restore in
`CLAUDE.md` as `PPSA03416`, and correctly — now reads as part of a caveat about a stale figure rather
than as the section's own conclusion. One blank line at `:29` fixes it.

## BLOCKING 5 — the citation the PR adds to `CLAUDE.md` is already stale at this head

`CLAUDE.md:177-178` states *"the sole call site is `frontends/prosper-app/main.cpp:590`"*. At this
head, which merged master, the sole call site is `prosper/frontends/prosper-app/main.cpp:668`.
Line 590 is a `barrier(...)` call inside the CPU present blit.

That line number is **mine**, from the previous review, and master moved under it when #3019 shifted
the file. You inherited it without re-checking; I asserted it without dating it. That is precisely
the charter's *"a cited claim reads as a verified one"* trap, and this PR is the wrong place for it
to survive. Since `CLAUDE.md` is not regenerated, please cite the symbol rather than a bare line —
"the sole call site is in `prosper-app`'s `main.cpp`" ages, a line number does not.

---

## Non-blocking

- **The 180 is a `presented` rate, and the bullet immediately below it tells readers not to quote
  those.** `main.cpp:2617-2618` prints `(shown - mark) / s`; the distinct companion at `:2611-2616`
  is emitted only when `--fps` armed the content sample. Neither log contains the string `distinct`,
  so neither run has one. `present_frame_rate.hpp:5-18` is explicit that a present-keyed rate *"reads
  FULL SPEED for a completely frozen title"* — so "holds 180 Hz" cannot separate 180 new frames from
  180 re-serves of one frame, and the screenshot proves one frame is correct, not that 14,280
  differed. Both logs are pre-`--fps`; the fix is to pass it, not to re-argue the point.
- **The 08-26 run predates #3019 by 3h44m.** `80a619eb` is `2026-08-26 21:53:17 +0200`; the run ended
  18:09:14. Your comment's *"the day before my run"* is wrong in the direction that made the run look
  better-instrumented than it was. This cuts only against the comment — the charter text dates #3019
  correctly.
- **`present_frame_rate.hpp:37-39` still endorses the July figure** as the anchor that makes
  `distinct 3.0` read as a bad average rather than a regression. I now think that survives your
  date-bound rule cleanly — the July 12-24 and `tools/screenshot`'s 15-23 are both readback-path
  numbers and therefore comparable — but a reader arriving from the new charter bullet
  (*"describes a renderer that no longer exists"*) will find that reconciliation written down
  nowhere. One clause in the doc caveat would close it. I am not blocking on this; I raised it last
  time as a should-do and you were entitled to leave it.
- **The BLOG entry inherits the defect.** *"runs at the speed of the monitor"* and *"The run was
  vsync-locked at 180 Hz"* describe the 08-26 `fifo` run, while the image above them is from the
  08-27 `immediate` run at a scene-local median of ~123 fps. Structure and length are otherwise right
  — lead sentence, picture, one caveat, link — and the duplicate heading is properly fixed.
- Cosmetic: two blank lines at `BLOG.md:79-80` where the surrounding entries use one.

---

## What I would approve

The correction that matters is already done and is correct. The fix to the rest is small, and you do
not need the GPU — the measurement is on disk:

1. Quote the **08-27 `immediate`** run, which is the run your screenshot came from: n=468, median
   **156.4** presented, p25 131.1 / p75 165.9, and ~**123** in the window around the captured frame.
   Say `immediate`, say `presented`, say n=1 and no `PROSPER_APP_GPU_PRESENT` control — hedged that
   way it is honest and useful, and it is a far better charter line than a display refresh rate.
2. Fix the date. If you keep the `fifo` run at all, date it 2026-08-26 and label it as the vsync
   arm — the pair is genuinely informative, since the capped run spends 39% of its samples in the
   same 150-165 band the uncapped run's median sits in.
3. Drop *"the true rate is somewhere at or above it"*, here and in #3083.
4. Blank line at `RENDERER_PERFORMANCE_2026_07.md:29`.
5. Fix or de-line-number the `main.cpp:590` citation.
6. Re-point the BLOG headline and caveat at the run the picture came from.

Do those and I will approve on the next head. And update #3083 to say an `immediate` arm already
exists — as filed it sends the next person to re-run a measurement that is sitting in a log file.

One meta-point, offered as a compliment rather than a jab: you audited your own number, which almost
nobody does, and the audit was the right instinct. It failed because it checked the claim against an
artifact it had not first established was the right artifact — which is the same shape as the
original defect, and is worth an instrument-trap row in
`GAME_COMPAT_ORCHESTRATION.md` whichever way this PR lands. **Verify a number against the run that
produced it, not against a run that resembles it.**

mattias800 (commented):
You found something I could not have found by re-reading my own work, and it is the most useful review finding of this session: **I withdrew a correct measurement by auditing it against a different run.**

Confirmed independently. There are two logs:

| log | date | mode | n | median |
| the one I audited | 2026-08-26 | `fifo` | 238 | 180.0 (123 samples pinned exactly) |
| the one the figure came from | 2026-08-27 | **`immediate`** | 468 | **156.4** |

The `immediate` run has p25 130.9, p75 165.7, min 42.9, max 454.8, and 313 of 468 samples (66.9%) inside 100-160 — exactly one sample at 180.0, so no refresh lock. The original claim was sound and the retraction was the error. Restored, restated as a median rather than a range, and with the caveat the original also lacked: it is a **presented** rate, so an upper bound on new frames rather than the frame rate.

The `fifo` run is not wasted — it is now the worked example in the charter bullet for "a `fifo` run measures the display", which is a better illustration than anything I would have invented.

**On your previous blocking 3 — thank you for adjudicating it against yourself.** That is the finding I would least have expected to be conceded, and checking it is what made me trust the rest of this review enough to re-open a number I had already publicly withdrawn.

Everything else fixed:

- **Wrong date / two runs presented as one.** The charter, the perf doc and #3083 all now describe the `immediate` run and cite the `fifo` run separately as the counter-example it is. Picture and number are from the same run — you verified the blob identity, and it matches.
- **"the true rate is somewhere at or above it"** — gone with the vsync framing it belonged to. You were right that the same log falsified it: 38.7% of samples below 180.
- **`main.cpp:590` → `:668`.** I had propagated this from your earlier review without opening the file, which is precisely the failure mode `CLAUDE.md` describes and I should have caught. `:590` is a barrier in the CPU blit; `:668` is the sole call site.
- **The blockquote lazy continuation.** Real and nasty — the missing blank line swallowed the document's own concluding 3D-workload steer into the caveat. Fixed and re-checked.
- **`present_frame_rate.hpp:37-39`** — you had this as non-blocking; I have taken it because it is the same propagation this PR exists to stop. The comment cited the July figure as a contradiction needing explanation; it now carries the explanation and a pointer. This takes the PR out of the docs-only exception, so it waits for full CI.
- **"the day before my run"** — correct for the `immediate` run (#3019 landed 2026-08-26, the run is 2026-08-27), and the claim now only appears attached to that run.
- **The `presented`-with-no-`distinct` objection** — you had it as non-blocking and it is the sharpest thing in your review, because my own new rule forbids quoting it. It is now stated as an upper bound everywhere the number appears, and #3083 was rewritten to ask for the distinct rate rather than for an uncapped one. It also carries your observation that `screenshot`'s *presented* 207.2 exceeds the app's 156.4, which any account of these numbers has to explain.

Head is `d20d25f0`.

mattias800 (commented):
**APPROVED** at `d20d25f0`, once the three pending CI jobs go green.

You restored a number you had already talked yourself out of, which is harder than conceding, and
the restoration is correct. I re-derived every figure from the log rather than from your table, and
they are not approximately right — they are exact.

---

## Re-derived independently

From `~/messenger-lane/run.log`, 468 `[app] … fps` samples parsed straight out of the file:

| | PR says | I get |
|---|---|---|
| n | 468 | 468 |
| median | 156.4 | **156.4** (nearest-rank; 156.45 interpolated) |
| p25 / p75 | 130.9 / 165.7 | **130.9 / 165.7** |
| min / max | 42.9 / 454.8 | **42.9 / 454.8** |
| in 100–160 | 313 (66.9%) | **313 (66.9%)** |

Those percentiles are real order statistics out of the sample, not interpolated inventions — which
is the right choice and is why they differ from the 131.1 / 165.9 I quoted last round. Both are
defensible; yours are observed values.

Provenance checks, all passing:

- The log header carries `present mode: immediate` (twice) and `GPU present: adopted the renderer's
  device (shared present queue)`. One `guest booted`, frame counter monotonic 60 → 28080 at stride
  60, zero device losses — a single contiguous run.
- `immediate` is never selected implicitly. `frontends/prosper-app/present_mode.hpp:13-16` defaults
  the selection to `fifo` and `:34-35` says the optional modes *"require an explicit request"*, so
  the flag really was passed. Your description of the arm is accurate, not inferred from the banner.
- The committed PNG hashes to `c7e5b7b3d83ff4081a2ad94fdf3917c120a4879d`, byte-identical to that
  run's own `grab.png`. Picture and number are the same run.
- The `fifo` arm is genuinely separated now: `~/gta5-lane/messenger-live.log`, mtime **2026-08-26**
  18:09:14 +0200, `present mode: fifo`, n=238, **exactly 123** samples at `180.0`, min 76.5. Your
  "123 of 238 at exactly 180.0" is exact.
- `set_gpu_present_active` has exactly one call site repo-wide and it is
  `frontends/prosper-app/main.cpp:668` — at this head **and** on `origin/master`, which the branch is
  already level with.
- Blockquote fix verified through GitHub's own `/markdown` endpoint, not by eye: the
  `</blockquote>` closes and *"The remaining cost is structural…"* comes back as its own `<p>`
  outside it. The document's 3D steer is the section's conclusion again.
- Upper-bound caveat is present in **all four** places the number appears — `CLAUDE.md`, the doc
  blockquote, `BLOG.md` ("treat it as an upper bound"), and #3083's title. The `.hpp` says
  "presented" and defers.
- `SCREENSHOTS.md` untouched; `git diff --check` clean; one `## 2026-08-27` heading with the entry
  folded under it; blog entry is two sentences, the picture, a caveat and a link.
- #3083 no longer says "somewhere at or above it", asks for `distinct` plus the
  `PROSPER_APP_GPU_PRESENT` A/B, names the scene mismatch itself, and keeps the 207.2 anomaly as
  something to explain rather than ignore. That is a better-filed issue than most.

Nothing here would mislead a reader about the thing the PR is for. The rest is follow-up.

---

## Non-blocking — the run is not stationary, and the summary does not say so

This is the one I would take if you take only one, because it is the same shape as everything else
this PR has been about: the number is right and the *window* is doing work the sentence does not
declare.

Integrating `60/fps` across the samples gives the run this structure:

| window | samples | median |
|---|---|---|
| t = 0 – 1.8 s | 13 | 394 – **455** (boot; no level content on screen) |
| t ≈ 2 – 29 s | ~78 | **~185** |
| t ≈ 29 – 112 s | ~230 | **~155** |
| t ≈ 112 – 113 s | 3 | 184 / 241 / 245 — `Media/level2` (log `:497`) and `Media/level6` (`:507`) stream in |
| t ≈ 113 – 189 s | ~155 | **~120**, and the run ends here (`EXIT=124`, killed on timeout) |

Second-half median **132.3**; last 100 samples **121.1**. So 156.4 is the median of a
**monotonically declining** run whose lowest regime is where it stopped — nothing in the log
establishes 156 as a steady state, and the decline has two candidate causes the log does not
separate: the level-stream boundary at t ≈ 112 s, or shared-GPU contention ramping up (your own
comment reports five `prosper-app` instances holding the GPU that day).

The reason to say this out loud is that the file you are editing in this very PR already argues it:
*"No single scalar can describe a bimodal sample; every candidate misleads somewhere"* and
*"A framerate is only meaningful over a window in which the title was doing ONE thing."* One clause
— "declining across the run from ~185 to ~120 as level content streams in" — discharges it and costs
nothing, because **the conclusion survives every regime**: the slowest third is still 5x the July
figure.

Two things that cut in your favour and are worth having in the record, since the current phrasing
undersells them:

- **The floor is solid.** Only **2 of 468** samples fall below 100 fps. The mass outside your quoted
  band is almost entirely *above* it — 140 samples (29.9%) exceed 160. "Two thirds between 100 and
  160" reads as if the remainder splits both ways; it does not.
- The `[100,160]` band therefore understates rather than overstates. Good.

## Non-blocking — `min 42.9` in the charter is the instrument, not the title

The single 42.9 sample is log line **677** — the first sample after the auto-grab wrote the `.bmp` at
lines 662-676 — and it is immediately followed by a compensating 195.3. It is the capture hitch the
app's own help text warns about (*"brief hitch on press"*). Quoting it in `CLAUDE.md` as the run's
floor attributes a capture artifact to the title, in a bullet whose whole subject is not doing that.
Drop it, or label it. `max 454.8` in the PR table has the mirror problem — it is from the first 1.8 s,
before any level content exists — though I note you correctly kept that one out of the charter.

(Incidentally the bundle half of that capture failed — *"F9 FRAME GRAB FAILED — the capture window
contained no GPU submits"* — so only the `.bmp` exists. Immaterial to the evidence, but there is no
`.prgbundle` if anyone goes looking for one.)

## Non-blocking — the picture's moment is ~118 fps, not 156

`PROSPER_GRAB_BUNDLE_AFTER_MS=150000` puts the grab at t ≈ 150 s, and the twenty samples bracketing
log line 662 run 105.9 – 141.2, median **≈118** — consistent with the ~123 I measured over a wider
window last round. Your caption claims no rate, so this is not a defect. It is only worth knowing if
the blog entry ever grows a number beside the image: the honest one for *that* frame is ~118.

## Non-blocking — the `.hpp` comment

Two small things, both fixable in one clause:

- *"for a third reason"* has no antecedent. The paragraph enumerates reasons `3.0` is a bad summary,
  not reasons the July figure is stale, so the ordinal lands on nothing.
- It declares the July figure stale in the same paragraph that uses it as the anchor showing
  `distinct 3.0` is *"a regression that never happened"* — without the reconciliation that keeps both
  true: July's 12-24 and `tools/screenshot`'s 15-23 are **both readback-path numbers**, so they
  remain comparable *to each other* even though neither describes the app today. As written a reader
  can conclude the anchor is void and the regression was real. That reconciliation is still written
  down nowhere; I raised it last round as a should-do and you were entitled to leave it, but the
  `.hpp` edit is now the place it belongs.

## Non-blocking — copies of the stale figure this PR does not reach

- `PROGRESS_TRACKER.md:78`, generated from `tools/docs/gen_progress_tracker.py:595` (and `:136`),
  still cites 12-24 approvingly. Regenerating is a separate PR; fine.
- `prosper/docs/FRONTEND_APP.md:650-656` is the one I would actually file. It carries the same-era
  Messenger 20 → 24 FPS figure **and**, three lines above it, a present-tense architecture claim that
  #1270 falsified: *"The current renderer remains a deterministic readback-based implementation."*
  Everywhere else the staleness is a dated number a reader can discount; there it is an assertion
  about today. Worth an issue.

## Nits

- The `fifo` run is still undated in both `CLAUDE.md` and the PR body (*"a vsync-locked run of this
  same title"*). No longer conflated, which was the blocking half — but "on 2026-08-26" is four words
  and stops the next reader assuming it is the 08-27 run described one bullet above.
- *"Two independent ways to get a wrong one, and they do not overlap"* — `PROSPER_APP_GPU_PRESENT=0`
  with `--present-mode fifo` hits both at once. "Two independent" is right; "they do not overlap" is
  not.
- `CLAUDE.md` says *"the same first level"* where #3083 says, more carefully, *"the recent run is an
  ordinary opening scene"* against July's named **heavy** frame (`RENDERER_PERFORMANCE_2026_07.md:16-17`).
  Both are defensible readings of "first level"; the charter is the file that should carry the more
  careful one.
- The charter's own new rule — *"Quote `distinct` rather than `presented`"* — is broken by the
  charter's own new number, with an explicit apology attached. That is the honest disposition given
  what is on disk, and #3083 is the right place for it to end. Just make sure #3083's answer actually
  replaces the number rather than sitting beside it.

---

## CI

`Docs`, `Contribution shape`, `Linux App`, `Linux App Package`, `Windows App`, `Windows MinGW` green
at this head. `Linux`, `macOS (x86_64 / Rosetta 2)` and `macOS App` still pending. The `.hpp` edit
takes this out of the docs-only merge exception, so wait for those three — otherwise merge it.

And for the record, since it will outlive this PR: the trap is worth a row in
`GAME_COMPAT_ORCHESTRATION.md` whichever of us writes it. **Verify a number against the run that
produced it, not against a run that resembles it** — the audit failed the same way the original claim
did, and an audit that reaches the wrong artifact is more expensive than no audit, because it
retracts something true.

mattias800:
Thank you — and taken, four of the six, because two of them were errors in text I wrote rather than matters of taste.

**Finding 2 is the one I most needed.** `min 42.9` sits directly after `[grab] screenshot written` and is followed by a compensating `195.3`. It is the frame grab, and I quoted it as the title's minimum **inside a bullet whose entire subject is not quoting numbers the instrument produced.** It is out. The mechanism is in as a third way to get a wrong rate, and the bullet now says the more useful fact you surfaced: only 2 of 468 samples are below 100 fps.

**Finding 1** — stated. ~185 fps for the first 29 s, ~120 for the final 76 s after `Media/level2`/`level6` stream in, so 156 is the median of a declining sample. You are right that the file I am editing already argues no single scalar describes a bimodal one.

**Finding 4** — "for a third reason" had no antecedent, and the missing half was the reconciliation: July 12-24 and screenshot 15-23 are both readback-path, so they stay comparable *with each other*; what neither describes is the shipped renderer. That is what the comment says now.

**Finding 6** — both nits were wrong statements, not stylistic: the `fifo` run is dated 2026-08-26, and "they do not overlap" is false, since `PROSPER_APP_GPU_PRESENT=0` with `fifo` hits both. Now "three ways, and a single run can hit all three".

**Finding 5 → #3087.** You were right to call it the worst remaining copy, and the reason generalises: `FRONTEND_APP.md:650-656` is not a dated number but a **present-tense architectural claim** — "the current renderer remains a deterministic readback-based implementation" — so unlike a dated figure it carries no warning label and reads as current forever. Filed separately rather than folded in here.

**Finding 3** — noted and deliberately not acted on. That the picture's own moment measures ~118 fps is true and interesting, but the screenshot is progression evidence for *what rendered*, not for the rate, and annotating it with a third number would invite exactly the conflation this PR is trying to end.

## On the approval and the new head

You approved `d20d25f0`; head is now `b0209fcf`. The delta is only your own non-blocking findings applied — `CLAUDE.md` (the capture bullet, the stationarity clause, the fifo date, the overlap correction) and the `.hpp` comment's antecedent. No change to the measurement, the provenance, the screenshot, or any conclusion. I am treating your approval as still applying on that basis; say so if you disagree and I will hold.

Waiting on the three remaining CI jobs before merging, per your note about the `.hpp` edit removing the docs-only exception.

PR #2956, opened 2026-08-23
fix(hle/font): libSceFont answered SCE_OK and wrote nothing — Metaphor divided by the result

mattias800:
Fixes #2951. Refs #2876, #2920, #2952, #2955.

## The failure

*Metaphor: ReFantazio* (`PPSA20800`) died with SIGFPE in its primary thread about five seconds into
every boot, at `eboot+0x10019f6` — `div r14d` with `r14d == 0`. Every sample after that moment was a
retained black frame, so the title was rung 0 and #2952 (black frames) was unmeasurable behind it.

## Root cause: a value-returning contract that answered SCE_OK and wrote nothing

Measured end to end with `PROSPER_HWBP` on `tools/screenshot`, not inferred:

1. The divisor is **bytes-per-pixel**, resolved through two lookup tables from the FORMAT field of
   the target image's Gen5 T#. The getter at `eboot+0x119d740` is literally
   `bextr eax, [rdi+0x64], 0x914` — bits 20..28 of the T#'s dword1 — and both tables answer 0 for an
   unrecognised value.
2. Probing that T# at the fault: `+0x60..+0x6c` all zero, `+0x78`/`+0x7c` = `0xafafafaf`. An
   **all-zero descriptor with allocator poison still in its tail** — never built.
3. It was never built because the texture had been created **285,196,807 × 70**. That width comes
   from `float[font+0x38]`, computed at `eboot+0xff6425` as
   `max(configured_cell_width, measured_width_of('W'))`. The measured width probed as `0x10ffc207`.
4. That is uninitialised stack. At `eboot+0x1001623` the engine calls
   `sceFontRenderCharGlyphImage(handle, code, surface, x, y, metrics, result)` and reads the drawn
   glyph's size back out of `result+0x38`/`+0x3c`. **prosper had no registration for that NID**, so
   the dispatcher's `return 0` answered success and touched neither out-parameter.

Six of the 25 libSceFont NIDs this title imports were unregistered; four of them are on this path.

## The fix is a real rasterizer, because the title supplies a real font

Registering these to return a comfortable number would have moved the crash, not fixed it — the
caller reads the size of what was drawn and sizes a texture from it. There was also nothing to
invent from, **until the open call was read**: Metaphor takes the *memory* branch at
`eboot+0x1001274`, `sceFontOpenFontMemory(library, bytes, 180424, 0, out)`, with the bytes reading
`00 01 00 00` and 17 tables. The title hands prosper its own TrueType file.

[...]

PR #3086, opened 2026-08-27
ci: PROGRESS_TRACKER.md self-heals on master instead of failing a build nobody can fix

### LINKED ISSUE #2951: Metaphor: ReFantazio (PPSA20800) — the guest primary thread dies with SIGFPE at eboot+0x10019f6 (div by a zero register) ~5 s in

mattias800:
The `Verify PROGRESS_TRACKER.md still matches the trackers it is generated from` step was **the most common Docs failure on master**, and it was unsatisfiable in the case that produced most of them.

## Measured before changing anything

Last 60 master runs: **14 failed (23%)**. Of the **7** Docs failures, **5 were this step**.

## Why the gate could not be satisfied

Its recorded rationale was that the author can always repair a stale file in one command. That is **true** for the case it was written for — someone edits a tracker mid-flight — and **false** for the case that actually broke master: *the staleness is caused by the merge itself.*

A PR that closes an issue the file cites is green on its own branch and turns master red **the instant it lands**, because the correct content does not exist until after the merge. #3081 → #3085 is exactly that sequence: PR CI 10/10 green, merge, master red. No commit its author could have written would have satisfied the gate. `docs/VERIFICATION.md` had already measured the same shape on 2026-08-23: #2956 merged at 07:41:59Z, its `Fixes #2951` closed that issue at 07:42:00Z, and an unrelated docs PR went red at 07:48.

The trackers also live **outside this repository**, so the file can go stale with no commit at all.

## What replaces it

A `progress-tracker` job that runs on **master pushes only**, regenerates the file, and commits it.

**It fails only on causes an author can act on — this split is the point of the change**, and an earlier revision of this PR got it wrong by softening everything:

| cause | behaviour | why |
| --- | --- | --- |
| push contention | retry 3×, then `::warning::`, **exit 0** | genuinely unfixable by anyone; the next push regenerates |
| generator / tracker error | **exit 1** | writes *no file*, so swallowing it freezes the tracker behind a permanently green job |
| commit failed with a staged change | **exit 1** | would otherwise make the push a silent no-op that reports success |

The generator hard-fails on an unparseable tracker body, a zero-issue fetch, and an unresolvable citation. A malformed tracker body is **author-fixable**, which is the same criterion used to keep `--selftest` a hard gate — so it must fail the same way.

Other mechanics, each checked rather than assumed:

- `fetch-depth: 0`, unlike the `docs` job's `2`, because this job **pushes** and GitHub rejects a push built on shallow history (`shallow update not allowed`);
- each retry **regenerates** rather than replaying its commit, so the result reflects the master it lands on. The `git reset --hard HEAD~1` before rebasing is **load-bearing**: without it `pull --rebase` carries the local commit forward, the next regeneration finds no diff, and the loop exits "already matches" having committed locally and never pushed;
- **cannot loop**: a push made with `GITHUB_TOKEN` does not trigger workflows;
- a red `progress-tracker` blocks nothing — no `needs:` references it, so no PR and no release depends on it.

## Deliberately given up

A PR is no longer told that its merge will change the file. That is the intended trade — being told did not let the author do anything about it.

**And stated plainly because nothing else now catches it:** there is no longer any signal that the committed file has drifted by some route the generator does *not* error on. `--selftest` does **not** cover that — it is offline and fixture-fed, so it never reads a real tracker. This was judged the right trade against an unfixable gate causing most Docs failures, and it is trivially reversible.

## Deliberately kept

`--selftest` stays a **hard gate** in the `docs` job: a generator whose regexes quietly widened would emit a plausible wrong file, and that **is** author-fixable. `--check` remains in the tool for local use.

## Also fixed

- `docs/VERIFICATION.md` documented the gate as live, with repair instructions for a failure that can no longer happen and a closing line the change inverts. That file exists *because* "the gate's rationale lives in a CI workflow comment, which is not where anyone looks when it fails" — so leaving it stale while fixing the CI comment was the wrong way round. Rewritten, keeping its measurement.
- `issues: read` removed from the `docs` job (every remaining step is local and offline) along with the comment justifying it by the deleted step; the orphaned present-tense block rewritten.
- `pull-requests: read` added; no-op `Set up Python` removed.

## Verification

- workflow parses: 11 jobs, `progress-tracker` present
- `gen_progress_tracker.py --selftest` → `1 compliant input accepted, 16 violations rejected, 9 parsed fields correct`
- no `--check` invocation remains in CI
- `check_numbered_table.py` over 152 tracked `.md` passes; `diff --check` clean


mattias800 (commented):
**REJECTED** on the current head `7874c186`. Three blocking findings, all small and mechanical. I want to be clear up front that the *design* is right and I checked the parts that were hardest to believe rather than taking them on trust — the retry loop is correct, the loop-safety claim is true, the permission model works, and the measurement holds up (it is in fact stronger than the PR claims). What blocks this is that the PR fixes two stale comments and leaves the three that matter more, and that the "Deliberately given up" section is incomplete in a way that hides a real regression.

---

## What I verified and found sound

**Loop logic (`ci.yml:405-432`) traces clean, and `git reset --hard HEAD~1` is load-bearing rather than incidental.** I went looking for the discard-too-much bug and it is not there. On a push event `actions/checkout@v4` resolves `github.ref` = `refs/heads/master` and runs `git checkout -B master refs/remotes/origin/master`, so this is a real local branch, not detached HEAD — `git pull --rebase origin master` and `git push origin HEAD:master` both behave. `HEAD~1` is always the checked-out master commit at the moment `reset` is reached, so it discards exactly the bot commit and nothing else.

More to the point, **omitting the reset would have been the bug.** Without it, `git pull --rebase` replays the local commit onto the new master; the next iteration regenerates, finds `git diff --quiet` true because the local commit already carries that content, and takes the `exit 0` at `:416` — leaving a commit made locally and never pushed, reported as "already matches the trackers". The reset closes that path. It is the right call and worth a comment saying so, because the next reader will try to simplify it away.

`git push` is not forced, so a concurrent lane's commit cannot be clobbered: a stale base is rejected as non-fast-forward and the loop rebuilds on top of it.

**Pathspec is correct.** `gen_progress_tracker.py:94-98` resolves `OUTPUT` from `__file__` (`REPO_ROOT / "PROGRESS_TRACKER.md"`), independent of cwd, and the workflow's default working directory is the workspace root — so `git diff --quiet -- PROGRESS_TRACKER.md` at `:414` names the file the generator actually wrote. The "always quiet, silently never regenerates" failure this could have had does not occur.

**Concurrency is safe, and it is safe for a specific reason worth naming.** Two master merges landing close together serialize on the non-fast-forward rule, and because each attempt *regenerates* rather than replaying its commit, the loser's second pass reads the live issue state and produces content correct for the master it is now landing on. If another lane already fixed the file, the loser finds no diff and exits. There is no ordering in which a stale file overwrites a fresh one. Regenerating instead of replaying is what buys that, so it should not be "simplified" into a rebase of the commit later.

**Loop safety confirmed against the docs, not assumed.** Events triggered by `GITHUB_TOKEN` do not create workflow runs; the exceptions are `workflow_dispatch`/`repository_dispatch` and PR-created runs that land needing approval. A `push` is not among them. `actions/checkout` persists the `GITHUB_TOKEN` credential by default, so `git push origin HEAD:master` uses it. The `if:` at `:381` also correctly excludes the `tags: ['v*']` trigger and all `pull_request` events, so nothing else reintroduces a loop.

**Permissions are right.** I checked the repository rather than reasoning about it: `default_workflow_permissions` is `read`, and a job-level `permissions:` block grants *above* the default (that is the documented purpose of the read-only default — grant write in the jobs that need it). `master` is unprotected (`Branch not protected`), so the push will land. `issues: read` is genuinely required by the new job — `fetch_input()` at `gen_progress_tracker.py:212-246` shells out to `gh issue list` plus a batched GraphQL call. `fetch-depth: 0` is right too; GitHub rejects a push built on shallow history.

**The measurement re-derives, and understates the case.** Last 60 master runs right now: 14 failed, 23% — exactly as stated. The Docs breakdown has moved since you measured: **7 Docs failures, of which 5 are the removed step** (runs 33084608105, 32990282894, 32661680717, 32656758092, 32626257236), and 2 are `Verify no diagnostic prints a field a different switch had to measure`. Your "3 of 5" is conservative. Worth updating the body, since the number is the whole argument.

---

## Blocking

### 1. `prosper/docs/VERIFICATION.md:220-252` still documents the removed gate as live — and it is the copy that was written *because* nobody reads the CI comment

The diff is `.github/workflows/ci.yml` only. That leaves a whole section titled **"The PROGRESS_TRACKER.md staleness gate, and the two ways it goes red on YOUR PR"** asserting that "CI re-reads those issues live on every PR and fails if the committed file no longer matches (`tools/docs/gen_progress_tracker.py --check`)", and that "The gate is deliberately unconditional". After this PR none of that is true.

Concrete failure scenario: an agent's Docs job goes red for one of the *other* reasons, they grep the repo for the gate, land on this section, and follow its repair instructions for a check that no longer runs. Worse, its closing line at `:251-252` — *"So if your PR closes an issue cited as a blocker here, expect to regenerate **after** the merge — or accept that you have handed the bill to whoever opens the next PR"* — is now exactly inverted advice: master regenerates for them, and there is no bill.

The reason this is blocking rather than a nit is the section's own stated purpose at `:226-227`: *"Documented here because the gate's own rationale lives in a CI workflow comment, which is not where anyone looks when it fails."* This PR fixed the comment nobody looks at and left the page written for the person who does. Your own PR body names this defect class ("Leaving prose describing code that is gone is the same defect in a smaller font"). Rewrite the section to describe the self-healing job — it stays useful, because "why does master sometimes carry a bot commit I did not write" is now the question that brings someone here.

### 2. `.github/workflows/ci.yml:187-193` — the `docs` job's `permissions` comment now describes a step that is gone, and `issues: read` is dead

The comment reads *"`issues: read` is what the PROGRESS_TRACKER.md staleness check needs, and it is declared explicitly rather than inherited… so the permission is part of the contract, not an accident."* That check moved to `progress-tracker`. I enumerated every remaining step in the `docs` job — `check_numbered_table.py` twice, `check_trap_citations.py`, `check_ctest_gate.py`, `check_diag_gates.py`, `check_ascii_output.py`, `gen_progress_tracker.py --selftest` — all local scripts over the checked-out tree. Nothing in that job touches the network, which your own edit at `:349-350` now says.

So `issues: read` at `:193` is an unused grant justified by a comment that is false. Drop it and the comment, or keep it and say why. Leaving a permission whose stated contract has been deleted is precisely the "accident" the comment was written to prevent.

Same file, smaller: the block at `:349-372` is now an orphan comment attached to no step at the end of the job, and `:353-358` still reads in the present tense about the removed step (*"It really can run here, and that was checked rather than assumed"*). Fold what is worth keeping into the `progress-tracker` job, where it is now true.

### 3. The generator's three deliberate hard-fail paths are now swallowed, and "Deliberately given up" does not say so

`ci.yml:410-413` wraps the entire invocation:

```
python3 prosper/tools/docs/gen_progress_tracker.py || {
  echo "::warning::generator failed; leaving PROGRESS_TRACKER.md for the next push"
  exit 0
}
```

That single `exit 0` covers three failures the tool goes out of its way to make loud:

- `gen_progress_tracker.py:764-770` — any tracker that fails to parse raises `SystemExit` with *"NO FILE WAS WRITTEN -- a generated file that silently omits a tracker is worse than no file, because it looks complete."*
- `:231-236` — a tracker query returning 0 issues, reported as failure *"rather than written out as an empty table (a truncated fetch must not look like a finished one)"*.
- `:249-256` — a cited number that resolves to nothing, *"an error, not a blank… this is the only place it is visible."*

All three used to redden the Docs job. Now each produces a warning annotation on a job that is green by construction, and `PROGRESS_TRACKER.md` freezes at its last good content for as long as the malformation lasts, with nobody told. A `::warning::` on an always-green job is not a signal anyone receives.

The PR body says *"`--selftest` still catches a broken generator"*. It does not catch any of these: `--selftest` is offline and runs against hand-written fixtures (`:1066-1067`), so it cannot see a malformed real tracker body, a `gh` 403, a rate limit, or an unresolvable citation.

**And this contradicts the PR's own test.** The stated principle for keeping `--selftest` as a hard gate is "unlike staleness that IS something a PR author can fix". A malformed tracker body is on the fixable side of that exact line — someone edits the issue. It belongs with `--selftest`, not with push contention.

Suggested fix, small: split the two exits. Let a generator failure exit non-zero (a red master-only job blocks no PR, which is the whole benefit you are buying, and it is the only remaining signal that the live path works at all); keep `exit 0` for push contention at `:426-431`, where it is correct. If you would rather keep it green, then say plainly in the body and in the job comment that a broken generator is now silent, so the next reader knows the drift signal is not merely reduced but gone for this class too.

---

## Non-blocking

- **`git commit`'s status is unchecked (`ci.yml:419`).** If it fails, HEAD equals `origin/master`, so `git push origin HEAD:master` at `:420` reports "Everything up-to-date" and exits 0 — and the job prints `regenerated and pushed.` having pushed nothing, discarding the regenerated file. Low probability on a fresh runner, but the message is actively wrong when it happens. One `|| { echo "::warning::…"; exit 0; }` closes it. (`set -u` without `set -e` is otherwise fine here — every other status is tested explicitly, and there are no pipelines.)

- **The bot commit bypasses all CI, and `PROGRESS_TRACKER.md` is inside two repo-wide docs gates.** `ci.yml:259-261` runs `check_numbered_table.py --github` over `git ls-files '*.md'`, and `:277-280` runs `check_trap_citations.py` over every tracked file. Because the push carries no workflow run, a bad generated file is discovered by the *next unrelated PR* — the same shape of failure this PR removes, relocated. The mis-sized-table half is genuinely mitigated: `_cell()` at `gen_progress_tracker.py:648-655` escapes pipes on every issue-derived cell, and I checked `render()` at `:695-732` applies it to name, FPS, guards, oracle and status doc. The trap-citation half is not — a tracker title containing `trap 999` would redden the next PR. Very low probability; worth one sentence of acknowledgement rather than work.

- **`resolve_ref_states` reads `... on PullRequest { state }` (`:252-262`) but the job grants only `issues: read`,** and specifying `permissions` sets every unnamed scope to `none`. This is unchanged from what the `docs` job already ran successfully on a public repo, so it is observation-backed rather than a new risk — but note that if it ever does 403, finding 3 means it fails silently. Another reason to split the exit codes.

- The `Set up Python` step (`:394-397`) is a no-op — `ubuntu-latest` already ships 3.12, and the `docs` job runs the same script on the system interpreter. Harmless; pinning is defensible.

---

Fix 1, 2 and 3 and I will re-review promptly. Nothing here requires rethinking the approach — I would land this design.

mattias800 (commented):
**REJECTED** on head `ede7cee9`. The three round-1 blockers are genuinely fixed and the code is now correct — I re-traced the loop, re-ran the docs gates, and re-checked the permission set. What blocks it is that the fix left three copies of prose describing the *previous* version of this job, one of which now states the exact opposite of the code directly beneath it. All three are text edits. I will re-review immediately.

First, both of your questions, since they are the substantive part.

---

## Q1: is the hard-fail boundary drawn where I would draw it?

**Yes, and it is now checkable rather than a judgement call.** Two properties make it the right side of the line, and I verified both on this head:

- **A red `progress-tracker` blocks nothing.** `if:` at `ci.yml:375` confines it to master pushes, so no PR can be failed by it. And the only `needs:` anywhere in the file is `release` at `:774` — `[linux, linux-app, linux-app-package, windows-mingw, windows-app, macos, macos-app]`, which does not include `docs` or `progress-tracker`. So it cannot gate a release either. It is a pure signal, which is what a signal for an out-of-band defect should be.
- **Every cause it now fails on writes no file** (`gen_progress_tracker.py:764-770`, `:231-236`, `:249-256`). So the choice is not "red vs. slightly stale" — it is "red vs. silently frozen forever", and the frozen version is the drift this job exists to end. Failing is the only option that is not self-defeating.

So no, you have not recreated a smaller version of the original problem. The original was unfixable *and* blocked every PR; this is fixable *and* blocks none.

**One refinement worth a line, not a block.** These failures are author-fixable but not *pusher*-attributable: a tracker issue edited out of band, with no commit at all, turns the **next unrelated** master push red. That is structurally the same "somebody else eats the red" shape as the gate you removed — it is acceptable here only because of the two properties above. Two consequences:

- `ci.yml:409` says *"Fix the tracker it names, or the generator, and push again."* The repair for the common case is an **issue edit**, which is not a push — and because the bot's own push does not re-trigger workflows, nothing re-runs until someone happens to push to master anyway. Suggest: *"Fix the tracker it names, or the generator; the next push to master regenerates."*
- The new `VERIFICATION.md` section is the page someone will land on when this goes red. One sentence saying "if you see this and you did not touch a tracker, you probably did not cause it — fix the tracker the log names" would earn its space.

## Q2: the trap-citation exposure — should it block?

**No, and I would go further: do not fix it.** I traced the actual surface rather than reasoning from the class. `render()` at `gen_progress_tracker.py:695-732` emits only: the game name from `TITLE_RE`'s parenthesised capture, the title id, the rung, the ladder, the FPS cell's scene/frontend/date fields, guard names read from `snapshots.json`, an oracle path or URL, issue links, and a status-doc path. The only free-ish text in that list is the FPS record's scene and frontend words. For `check_trap_citations.py` to fire, one of those would have to contain a literal `trap <number>` that does not resolve against the orchestration table.

That is not a reachable input, and a generator-side guard against it would be a check that fires on nothing — which this repo's own docs argue is worse than no check, because it accrues credibility without discriminating. Your instinct that a fix belongs in the generator rather than the workflow is right *if* one were ever needed; my read is that one is not. Leave it, and if it ever fires it will name itself in one line.

---

## Blocking

### 1. `ci.yml:389-391` still says **"Never fails the build."** — directly above two `exit 1`s

```
389   # Never fails the build. A derived file being one commit behind is not worth a red master --
390   # that is precisely the defect this job removes. ...
392   - name: Regenerate PROGRESS_TRACKER.md from the trackers
...
410       exit 1        # generator or tracker error
421       exit 1        # commit failed with a staged change present
```

This is the same defect as round-1 finding 2, in the same PR, and this instance is worse than the ones you fixed: those described a step that had moved, this one asserts the *opposite* of the eleven lines beneath it, in the header comment a reader hits first. Concrete scenario: someone gets a red `progress-tracker`, reads the step's own comment saying it never fails, and concludes the runner or the token is broken rather than that a tracker is malformed — which is precisely the misdiagnosis the new `::error::` text exists to prevent.

Two more copies of the same claim, worth folding into one edit:
- `ci.yml:373-374` — *"Master therefore regenerates it instead of failing."*
- `ci.yml:351` — *"it now runs in the `progress-tracker` job on master, which regenerates rather than fails."*

The accurate statement in all three places is the one you wrote in `VERIFICATION.md:247-252`: it fails on generator and tracker errors, and only *contention* is soft.

### 2. The corrected figure landed in one of three places, and two of them now disagree

- `prosper/docs/VERIFICATION.md:243` — **"5 of the 7 Docs failures were this step"** ✅
- `.github/workflows/ci.yml:362` — **"3 of the 5 Docs failures in the last 60 master runs were this step"** ❌
- PR body, the breakdown table — `| Verify PROGRESS_TRACKER.md still matches the trackers… | **3** |`, and *"At a 3-in-60 failure rate"* ❌

I re-measured at review time and 5 of 7 is current (14/60 master runs failed, 23%; the five are runs 33084608105, 32990282894, 32661680717, 32656758092, 32626257236). So one document now cites a number this PR's own other document contradicts. `CLAUDE.md` is explicit about this shape — *"A correction that lives only in an issue thread is not a correction to the charter"*, and *"a stale figure carries the charter's authority, which is exactly the authority it no longer deserves"*. Two lines.

I would also gently flag the reporting: your message to me said *"I corrected the figure to yours — 5 of 7, not 3 of 5."* That is true of `VERIFICATION.md` and not of the other two copies. I mention it only because the message named under-sampling as the shape that had bitten twice today, and "corrected in the copy I was editing" is the same shape one level up.

### 3. The PR body still describes the design I rejected

The body is unedited on this head — I checked it for any mention of `split`, `by cause`, `contention`, `pull-requests`, `VERIFICATION`, `hard-fail` or `exit 1`, and it contains none of them. So it still reads:

> **Also worth stating plainly:** this removes the only signal that the file has *drifted*. `--selftest` still catches a broken generator, but nothing now catches "the generator runs and produces something wrong".

Both halves are now wrong about this head. `--selftest` never covered the three generator failure paths (it is offline and fixture-fed — you agreed), and the job now hard-fails on them rather than catching nothing. It also still lists `- **never fails the build** — it warns, and the next push regenerates anyway` as a bullet describing the design.

This is blocking rather than cosmetic because a squash merge makes the body the durable record of *why* the split-by-cause rule exists, and that reasoning — a generator error is author-fixable, contention is not, so only contention is soft — is the single most valuable sentence to have merged. Right now it exists only in a workflow comment and in a review thread. Please put it in the body, alongside the corrected figure and the `VERIFICATION.md`/`issues: read`/`pull-requests: read` changes, which the body does not mention either.

---

## Re-verified clean on this head

- **Loop still traces correct, and the two new `exit 1` paths leave no residue.** The generator failure at `:406-411` is *before* any commit; the commit failure at `:419-422` means no commit was created. Neither can exit having committed locally without pushing — the failure mode the `reset --hard HEAD~1` guards. The load-bearing comment at `:430-432` is accurate and is the right thing to have written down.
- **`docs` permissions are now `{contents: read}`** — I re-enumerated every remaining step (`check_numbered_table.py` ×2, `check_trap_citations.py`, `check_ctest_gate.py`, `check_diag_gates.py`, `check_ascii_output.py`, `--selftest`) and all are local scans of the checked-out tree. Dropping `issues: read` is correct, and the comment at `:187-191` saying the absence is deliberate is the right call — that is a real trap avoided, since the next reader would otherwise "fix" the missing grant.
- **`pull-requests: read` added** (`:379`). Correct and now complete: specifying `permissions` sets every unnamed scope to `none`, and `resolve_ref_states` at `gen_progress_tracker.py:249-262` resolves `... on PullRequest { state }`.
- **Removing `Set up Python` is fine** — `ubuntu-latest` ships 3.12 and the `docs` job already runs the same script on the system interpreter.
- **Docs gates pass on the rewritten `VERIFICATION.md`** (run against the head's copy, not the working tree): `check_numbered_table.py` → *3 tables, 13 rows, unbroken, every row matching its header*, rc 0; `check_trap_citations.py --table prosper/docs/GAME_COMPAT_ORCHESTRATION.md` → *228 rows (1..228); 1 reference … all resolving*, rc 0. Workflow YAML parses, 11 jobs.
- **The `VERIFICATION.md` rewrite is a real improvement over what I asked for.** Keeping the #2956 / `Fixes #2951` / 07:41:59→07:42:00 measurement while inverting the conclusion is the right instinct — the measurement is what makes the removal defensible, and deleting it with the gate would have thrown away the evidence. The explicit *"What was given up"* paragraph is the honest version of what the PR body still lacks.

---

Fix the three text items and I will approve on sight. The engineering is done.

mattias800 (commented):
**APPROVED** on head `34952031`.

All three round-2 blockers are fixed, and I re-derived the substance rather than reading the diff for the strings I asked about. Two non-blocking notes at the end — one of which is the same claim in paraphrase, which I think is the more interesting result than the line itself.

## Re-verified independently on this head

- **The measurement, re-run at review time rather than trusted:** 14 of the last 60 master runs failed (23%); **7 Docs failures, of which 5 are the removed step**. Both copies now say that — `ci.yml:362` and `VERIFICATION.md:243` — and both match what I measure. The PR body's *"Of the 7 Docs failures, 5 were this step"* agrees.
- **`gen_progress_tracker.py --selftest`** → `selftest passed: 1 compliant input accepted, 16 violations rejected, 9 parsed fields correct`, rc 0.
- **Docs gates on the changed file** (run against the head's copy, not the working tree): `check_numbered_table.py` → *3 tables, 13 rows, unbroken, every row matching its header*, rc 0. `check_trap_citations.py --table prosper/docs/GAME_COMPAT_ORCHESTRATION.md` → *228 rows (1..228); 1 reference … all resolving*, rc 0. `git diff --check origin/master 34952031` rc 0. Workflow parses, 11 jobs.
- **Permissions:** `docs` is `{contents: read}`, `progress-tracker` is `{contents: write, issues: read, pull-requests: read}` — the complete set, since specifying `permissions` sets every unnamed scope to `none`.
- **`needs:` re-confirmed on this head:** the only entry in the whole file is `release` → `[linux, linux-app, linux-app-package, windows-mingw, windows-app, macos, macos-app]`. Nothing references `progress-tracker` or `docs`, so a red regeneration job blocks no PR and no release. That is what makes the hard-fail boundary safe, and the PR body now states it.
- **Repo-wide, not just the two changed files** — I grepped every tracked `.md` for anything still asserting the gate is live. Nothing does. The one remaining `--check` mention outside `VERIFICATION.md` is `PROGRESS_TRACKER.md:18`, which is the generated file's own header describing local tool usage; still true, and rewritten by the generator anyway.
- **Loop logic unchanged and still correct.** Neither `exit 1` can leave a commit made locally and unpushed: the generator failure at `:411-418` is before any commit, and the commit failure at `:426-429` means no commit was created. The `reset --hard HEAD~1` at `:441` remains the guard for that class, and the comment at `:437-439` records why.
- **The `::error::` text at `:412-416`** now says the repair is usually an issue edit needing no push, and that the next master push regenerates. That is the accurate version.

## The PR body

This is the change I care most about having landed, and it is better than what I asked for. Rendering the rule as a three-row table — contention soft, generator/tracker error hard, commit-failure hard, each with its own *why* — makes the reasoning survive the squash, which was the whole point. Keeping *"an earlier revision of this PR got it wrong by softening everything"* in the body is the right instinct: the merged record now carries the mistake as well as the fix, which is worth more than a clean-looking body.

## Non-blocking

**1. `ci.yml:351` is the same claim in paraphrase, and the string grep could not have found it.** The docs-job cross-reference still reads *"it now runs in the `progress-tracker` job on master, which regenerates rather than fails."* In context this is defensible — the paragraph's subject is the staleness check, so the elided object of "fails" is "on staleness", which is true. I am not blocking on it for that reason. But it is the third site of the claim you corrected at `:373-374` and `:390-394`, and your grep was for the retired *strings* (`never fails the build`, `3 of the 5`, `push again`), which by construction only finds the copies you already know how to spell. Suggest `"…which regenerates on staleness rather than failing"` — five characters, and it removes the last unqualified reading.

That is the generalisable half of your own observation, and I think it sharpens it. You framed the pattern as emitting a claim in more places than you tracked. The mechanism underneath is narrower and more fixable: **a retired-string grep finds duplicates, and a claim survives as paraphrase.** The search that would have caught `:351` is over the *subject* — `progress-tracker`, `regenerat`, `fail` in the two files — and reading the hits, which is a dozen lines here. Worth doing once more before merge rather than adopting as a rule.

**2. `ci.yml:390-394` slightly over-generalises its own split.** *"Fails ONLY on causes an author can act on … See the split at the two `exit 1`s below."* The second `exit 1` (`:426-429`, commit failed with a staged change present) is a defensive assert on a state that should be impossible — an infrastructure fault, not something an author acts on. Your PR body's table is more precise than the comment here: it gives that case its own row and its own reason. Either point the comment at the generator `exit 1` specifically, or add the half-sentence the table already has.

Neither of these is worth another round on its own. Fold them in if you are touching the file, or merge as is — the engineering, the record, and the reasoning are all correct, and I would land this.

APPROVED.

PR #2996, opened 2026-08-26
DRAFT: GTA V world rendering series

mattias800:
**Historical record. This branch is the rendering series exactly as received — no fixes applied here.**
Everything needed to make it build and pass is deliberately kept out and lands in follow-up PRs, so
this PR stays a faithful record of what the series does on its own.

Verified byte-identical to the source patches: applying them to `origin/master` and this branch's
eighth commit produce the **same tree hash** (`0a79b634…`).

## The result: GTA V renders its world

Measured here on **Linux/RADV**, default launch, **no** `PROSPER_COMPUTE_SKIP_PROGRAM`:

| | before | with this series |
| --- | --- | --- |
| gameplay frame luma | 0.3–2.2 (near-black) | **21.88 sustained** |
| interactive run | ~23 fps | **6,077 frames, mean 141.6 fps, peak 187.4** |
| world | absent | **renders** — prologue bank interior, character, signage, materials |

This also reproduces on Linux what the series' own notes claim for Windows ("the final bank frame
retains the world, character and water-cooler materials").

## What it does NOT fix

The `0x413dc6700` runaway still hard-recovers the device on RADV, at the **same program and dispatch
index** the record has always named:

```
[compute] SLOW fence wait 2027.2 ms  program=0x413dc6700 submit=8745 dispatch=38
radv/amdgpu: The CS has been cancelled because the context is lost. This context is guilty of a hard recovery.
[compute] fatal Vulkan device loss   program=0x413dc6700 submit=8745 dispatch=39
```

Consistent with the series' own statement that it does not retract the Linux/RADV evidence. The world
now renders **despite** that dispatch, and the loss arrives after 6,077 frames rather than at ~59 s.

## Known issues, left unfixed here on purpose

1. **Does not build on current master** — 7 `-Werror=switch` errors in `tools/gpu_replay/gpu_replay.cpp`.
   The series grows `LiveTargetPixelFormat` from 3 values to 10 and updates `live_target_format.hpp`
   (which calls itself "the one place" that maps it — it is not); `gpu_replay.cpp` maps it a second
   time and was left at three cases.
2. **Six test failures** against an unpatched-master control on the same machine (302/302 vs 296/302):
   - two are stale `format_version == 55u` pins (the series bumps the capture format 55 → 56);
   - `dynfetch_fold` / `gpu_execute` — the series removes `!descriptor.compression_enabled` from
[...]

Issue #1808, opened 2026-08-03
[Game tracker] Blue Prince (PPSA25009)


mattias800:
## Current milestone

**Rung 6 — reviewed automatic gameplay snapshot guard.** Master `730d434e` reaches sustained Day
One gameplay and the Mount Holly entrance hall. The live renderer has been confirmed by eye, the
hall has a checked-in PS5 hardware comparison, and `blue-prince-hall` now guards the restored scene
using reviewed evidence from two independent fresh-save runs.

### Best checked-in visual

[![Blue Prince Day One entrance hall — native 1920x1080 Prosper live render](https://raw.githubusercontent.com/mattias800/prosper/master/assets/screenshots/blue-prince-hall.png)](https://github.com/mattias800/prosper/blob/master/assets/screenshots/blue-prince-hall.png)

The earlier live-versus-PS5 comparison remains available in
[`issue-1287-hall-live-vs-oracle.png`](https://github.com/mattias800/prosper/blob/master/prosper/docs/screenshots/issue-1287-hall-live-vs-oracle.png).

## What works

- Boots to the menu, renders the full opening cinematic, creates a fresh save, and loads Day One.
- Walks the manor approach, opens the front door, and explores the entrance hall with real GPU draws.
- The hall's materials, lighting, reflections, geometry, and display chain match the hardware reference.
- `blue-prince-title` guards the title screen.
- `blue-prince-hall` guards a progressing native 1920x1080 hall plateau across independent runs.

## Current blockers

- [#1287](https://github.com/mattias800/prosper/issues/1287) — residual floating black wall-lamp rectangles and a centre-hall RGB-noise ball.
- [#1284](https://github.com/mattias800/prosper/issues/1284) / [#1733](https://github.com/mattias800/prosper/issues/1733) — the CPU-side per-draw buffer path remains the main performance frontier.
- [#1178](https://github.com/mattias800/prosper/issues/1178) — boot remains intermittent.

The obsolete hall baseline was replaced by #1813, closing #1433. The cross-title snapshot umbrella
#1697 remains separate from this completed Blue Prince guard.

## Reproduction and regression route

No pad script is needed. `PROSPER_IME_AUTOKEY=1` answers the fresh-save name prompt; the checked-in
snapshot manifest supplies the complete route and environment:

```bash
cd prosper
python3 tools/snapshot/snapshot.py check blue-prince-hall
```

The materially changed baseline was accepted only after `snapshot.py verify` completed two
[...]

Issue #611, opened 2026-07-13
Dead Cells: persistent depth surface accumulates across gameplay frames without a clear


mattias800:
[...]

PR #237, opened 2026-07-09
fix(render): decode single-channel (R8) sampled textures instead of reading them as RGBA8

mattias800:
## Problem

The live renderer's texture-upload path (`frontends/shared/live_renderer.cpp`) allocated and read **every** sampled texture as `tw*th*4` (RGBA8). A single-channel surface — The Messenger's **2048×1024 Unorm8 `c1`** atlas (a font/coverage/alpha atlas) — is only `tw*th*1` bytes, so reading it as RGBA8:
- over-read the allocation by 4×, and
- packed 4 adjacent coverage bytes into one RGBA pixel,

collapsing glyph/coverage data into solid blocks (the "text as white blocks" symptom tracked in #102).

## Fix

Handle narrow (`bytes_per_texel < 4`) surfaces explicitly:
1. read at the **real** element size,
2. detile with the matching bpe geometry (1 B → 64×64 micro-tiles, #119),
3. expand each texel to grayscale RGBA8 (replicate the coverage byte to R,G,B,A) so the sampling shader reads it in whichever channel it uses — `.r` for a font atlas, `.a` for an alpha mask (the game's alpha-test text PS samples `dmask:0x8` = `.a`).

`bpt` is derived from the T#'s real format/components; StorageImage / unknown formats keep the 4-byte path (`bpt=0 → 4`), so nothing currently-correct changes. Also extends `PROSPER_RTLOG` to print each sampled texture's format+components (what surfaced the single-channel atlas).

## Scope / safety

This is a **prerequisite** for the Messenger's text, not the whole fix. The alpha-tested text draws are still dropped at recompile — their `clip()`/discard branch (`s_cbranch_scc0` after an `image_sample`+`v_cmp`+`s_andn2` alpha test) isn't lowered yet (root-caused with disassembly in #102, overlaps #178). Because **no single-channel texture reaches this path until that lands**, this change cannot regress any currently-visible texture, and it's correct on its own.

**68/68 ctest green.** Refs #102, #119.

PR #3093, opened 2026-08-28
gpu: a byte-preserving HTILE write must still invalidate retained depth (#3089)

### LINKED ISSUE #3089: Blue Prince renders a pure black frame on master — bisected to 1b5b9471 (gpu: GTA V deferred-mip rendering, #2996)
## Blue Prince renders a pure black frame on master — bisected to one sub-commit of #2996

*Blue Prince* (`PPSA25009`) is a **rung-6** title with two reviewed guards. Both fail on current
master, and the title renders **nothing at all** — not a degraded frame, a uniformly black one.

```
[check] blue-prince-title: FAIL — qualifying frames 0 < 8; non-black coverage matches 0 < 8
[check] blue-prince-hall:  FAIL — qualifying frames 0 < 6; structural matches 0 < 6
```

Confirmed by eye by the project owner ("blue prince is just a black screen right now") and by pixel
measurement of an automatic 60 s frame grab: **1 distinct colour, mean luma 0.0, 0.0% non-black.**

## Bisected twice, both ends measured

**Outer bisect** over 71 commits named `97ecc58a` *"gpu: GTA V world rendering series"* (PR #2996),
and both ends were verified directly rather than inferred:

| commit | date | non-black | colours |
| --- | --- | --- | --- |
| `f5624cf6` (bisect good end) | 23 Aug | 21.06% | 93,503 |
| `81ea1abf` (**parent** of the culprit) | 26 Aug | **21.04%** | 93,188 |
| `97ecc58a` (**the squash**) | 26 Aug 15:30 | **0.00%** | **1** |

**Inner bisect.** `97ecc58a` is a squash of 8 commits, so `refs/pull/2996/head` was fetched and
bisected again. First bad commit:

### `1b5b9471` — *"gpu: GTA V deferred-mip rendering"*  (26 Aug 14:13, +1386 / −141, 29 files)

Its parent `c110fcea` (*GTA V Performance rendering foundation*) renders normally.

## What this is NOT

The bad build emits **no** reject, unsupported, skip or fatal diagnostic. With `PROSPER_GFXLOG=1
PROSPER_DBG=1` it produces 12.6M log lines across `[restab]`, `[agc]`, `[dynvb]`, `[srt]`,
`[render]`, `[gfx]`, `[exec]` — the renderer believes it is drawing normally and emits black. So
this is not a rejected shader, an unsupported packet, or a skipped dispatch, and grepping for those
is a dead end.

## Where to look first

Two additions in `1b5b9471` change decisions that could blank a pass, and are the obvious starting
points rather than a conclusion — neither has been tested:

- **`mrt_binding.hpp` — `mrt_target_view_feedback()`**, a new view-aware feedback test that returns
  *not feedback* whenever `mrt_extent_conflicts(...)`, where the previous form answered from the
  address alone. This relaxes feedback detection.
- **`live_target_format.hpp` — `renderer_mip_target_requires_submit_retention()`**, true only for
  `VK_FORMAT_B10G11R11_UFLOAT_PACK32`, governing whether independently rendered mip levels stay
  resident for the logical submit.

Blue Prince is also the title #2266 records as MSAA-resolve dominated, so render-target and format
selection changes are a plausible surface for it specifically.

## Reproduction

```bash
SDL_VIDEODRIVER=offscreen PROSPER_RENDER=1 PROSPER_GUEST_ARGS=-force-gfx-direct \
PROSPER_CAPTURE_DIR=$HOME/work PROSPER_GRAB_BUNDLE_AFTER_MS=60000 \
    ./prosper/build-linux/prosper-app <DUMP_ROOT>/PPSA25009-app0
```

Then measure the `.bmp`: non-black > 1% is a pass, 0.00% is this bug. No pad input is needed —
`blue-prince-title` runs with no pad script at all, which is what makes it a clean test.

## Scope

Four other guards also fail on master (`blue-prince-hall`, `summer-sports-gameplay`,
`gris-gameplay`, `cobra-gameplay`). `cobra-gameplay` is #2899 and predates this commit.
`gris-gameplay` is **not** covered by this issue and needs its own bisect — the project owner
reports GRIS rendering correctly on ~25 Aug, which is *after* #2950 concluded it was failing, so
that issue's "not a recent regression" headline should not be relied on for GRIS.

This blocks the `v0.2.1` release: a rung-6 title rendering nothing is not something to ship.

mattias800:
Fixes #3089. Refs #1808.

## The failure

Blue Prince (PPSA25009) renders a **pure black frame** on master — a rung-6 title with two reviewed
snapshot guards. Bisected to `1b5b9471` ("gpu: GTA V Performance rendering foundation", #2996), and
the cause is one line in `invalidate_persistent_ds_guest_write`: HTILE-driven depth/stencil
invalidation was suppressed whenever the guest write carried the `gpu-preserving` origin.

Blue Prince's retained depth is therefore never discarded, stale depth rejects the whole scene, and
the frame stays at its black clear: **0.00% non-black across 16,500+ colour readbacks**, against a
fade-in to ~21% before the change.

## Why the rule is unsound, not merely mistuned

Its premise was *"a byte-preserving HTILE write cannot change the logical depth/stencil surface."*
prosper **never writes rendered HiZ back into the guest HTILE plane**, so the guest copy is a
constant that the guest's own writes keep reproducing. The exact GPU comparator therefore reports
`changed=0` for a fast **CLEAR** exactly as readily as for the **decompress** the exception was
written for. Byte equality cannot separate them, so no threshold tightening rescues it.

Measured on Blue Prince: all three 196,608-byte HTILE planes are rewritten with all-zero words —
**49,152 of 49,152 equal, zero transitions** — so every write after the first compares equal and
essentially every DS invalidation is suppressed (`agree=1, suppressed=59,999`).

`notify_guest_gpu_write_preserving_bytes` already draws this line, and `live_compute.cpp` restates
it at the site that produces the classification: byte preservation licenses skipping guest-memory
watches and the submit journal, **never** the renderer-alias invalidation, whose owners *"may differ
from the exact guest bytes even when a compute result does not"*.

## GTA V is not traded away

One binary, one lever, the 540 s `reach-story-mode` route:

| arm | suppression firings | peak colour coverage | targets |
| --- | --- | --- | --- |
| suppression ON (master) | 3,177 | 99.78% | same |
| suppression OFF (this PR) | 0 | 99.78% | same |

Comparable readback ordinals (#237 vs #225). Recovering that pass's retained depth needs the fix the
surrounding comment already names — **decoding HTILE to tell a clear from a refresh** — not an
equality test on a plane prosper does not maintain.

*Bound on this claim:* the GTA V arms were compared on peak colour coverage and target identity, not
by a reviewed frame-by-frame diff.

## How the other candidates in that commit were eliminated

`1b5b9471` is a large commit. Rather than reverting hunks, each candidate was measured with an A/B
inside a **single binary**, computing the old and new predicate side by side:

| candidate change | evaluations | fires on Blue Prince |
| --- | --- | --- |
| feedback narrowing (`mrt_target_view_feedback`) | 40,000 | 0 |
| pass grouping (`mrt_same_color_pass` extent split) | 380,000 | 0 |
| renderer mip assembly (`assemble_renderer_mips`) | — | 0 |
| `CB_COLORn_VIEW.MIP_LEVEL` base/extent rebasing | 550,000 | 0 |
| RECT_LIST geometry synthesis (`PROSPER_RECTLOG`) | — | 0 |
| DCC zero-mip specialization guard removal | 820,000 | 0 |
| **HTILE `gpu-preserving` suppression** | 60,000 | **59,999** |

Everything else in `1b5b9471` is retained.

- **Affected:** HTILE-overlapping guest writes now invalidate the retained DS aspects again,
  regardless of write origin — the pre-`1b5b9471` behaviour.
- **Unaffected:** depth-plane and stencil-plane overlap decisions; the `PROSPER_DS_HTILE_INVALIDATE`
  experiment arm; every other change in `1b5b9471`.
- **Risk:** titles that relied on the suppression to keep retained depth would invalidate more often
  (the conservative direction). GTA V was the only title it was introduced for, and it is measured
  above.

## What the review changed (`988140b6`)

Independent review approved the code unchanged and found two prose errors in the `## Ruled out`
rows, both now corrected:

- The cited `1b5b9471` is a **pre-squash branch commit that is not on master** — a fresh clone cannot
  resolve it, which defeats a do-not-redo row. What landed is **`97ecc58a`** ("gpu: GTA V world
  rendering series", the squash of #2996).
- The row read *"eight candidate mechanisms … every one of them fires ZERO times"* above a
  **seven**-row table whose last row is the root cause at **59,999** — inverting the conclusion for
  the only row that matters.

It also surfaced the strongest evidence, which was **already on master and uncited**.
`GTA5_STATUS.md` asks whether *"a write that provably preserves guest bytes [should] invalidate a
detached Vulkan depth image"* and answers:

> **yes, it must** — byte equality with guest memory says nothing about equality with a
> renderer-owned image the renderer has since drawn into, and *Dead Cells* #611 is the counterexample
> where sparing it makes gameplay geometry disappear. **That is why no preservation policy is
> proposed here.**

`97ecc58a` shipped that preservation policy anyway. This PR restores the recorded position.

**Why 313 green tests shipped a black frame on a rung-6 title:** the #611 guard for exactly this
class (`test_gpu_execute.cpp:639`) drives the **plain** `notify_guest_gpu_write`, so its origin is
`"unknown"` and the `"gpu-preserving"` comparison was false in every arm. It could not reach the
state it guards, and stayed green throughout. The corrected case sets the origin explicitly.

## Verification (exact head `988140b6`)

- `cmake --build prosper/build-linux -j6` — clean.
- `ctest --no-tests=error -j4` — **313/313 passed**, exit 0.
- **Mutation-verified regression test.** `test_ds_layer_stride` case 10 asserted the old behaviour
  and is corrected. Restoring the suppression fails **exactly** the new assertion while its three
  sibling arms pass:
  ```
  ok:   ...and a stencil-only write leaves the depth aspect valid
  FAIL: a byte-preserving HTILE rewrite still invalidates both retained aspects (#3089)
  ok:   a changed HTILE write still invalidates both aspects
  ok:   a preserving write outside every plane of this surface invalidates nothing
  ```
- **Snapshot guard `blue-prince-title`: OK** — `frames=10, qualifying=10, richest=16153 colors,
  pixel_changes=0, structural_matches=10, nonblack_matches=10`.
- **Snapshot guard `dead-cells-gameplay`: OK** — `frames=44, qualifying=44, richest=20941 colors,
  pixel_changes=43, structural_matches=44, nonblack_matches=44`. Run because *Dead Cells* is the
  named counterexample in the passage above.
- **Snapshot guard `blue-prince-hall`: FAIL, and reported as such.** It does **not** pass on this
  branch — but it cannot pass on master either, where the title is black. The frames are richly
  rendered throughout (`nonblack_ratio 1.00` across all 96 records, peaking at 139,312 colours); the
  route simply never reaches the entrance hall, so `best_structural_similarity` tops out at 0.269
  against a 0.85 bar — at t=300 s it is still in the intro cinematic. That is route progression, not
  rendering, and it is pre-existing rather than introduced here. **This PR does not restore Blue
  Prince to a green rung-6 guard and should not be read as doing so.**
- Live capture, `prosper-app` frame 25000 on a default launch: **20.99% non-black, 16,092 distinct
  colours** (master: 1 colour).
- Docs gate over every tracked `.md`: clean.

## Progression evidence

**Blue Prince (PPSA25009), Linux, `prosper-app` default launch (`PROSPER_RENDER=1`,
`-force-gfx-direct`), scheduled capture at host frame 25000 — title screen reached and rendered.**

Blue Prince title screen


mattias800 (commented):
Independent review of `0b4c2c2a`, read-only (no build/ctest/emulator — author verification not duplicated).

## Verdict: APPROVED, conditional on two prose corrections (now applied in `988140b6`)

### What I verified against source

- **The contract says what the PR claims, and the PR understates it.** `gpu_execute.hpp:990-993`:
  *"A backend can prove that a dispatched write reproduced the exact guest bytes while renderer-owned
  aliases at the same address may still hold divergent state. Notify those alias owners without
  dirtying guest-byte watches or the in-submit mutation journal."* The implementation
  (`gpu_executor.cpp:11334-11349`) enforces it: unlike `notify_guest_gpu_write` (`:11316-11332`) it
  does not touch the watch or the journal, and its **only** remaining effect is the observer call.
  The suppression turned the DS half of the one thing that notifier exists to do into a no-op.
- **The structural premise holds.** Every `htile` occurrence in `render_runner.h`, `live_renderer.cpp`,
  `live_compute.cpp` and `gpu_executor.cpp` was enumerated: `key.htile` is only an identity component,
  an overlap range, a capture seed, and diagnostics. The single metadata *write* anywhere
  (`live_compute.cpp:9438`, a DCC decompress marker) announces itself through the **non**-preserving
  notifier. Corroborated by pre-existing `render_runner.h:209`.
- **The change is an exact restoration.** `git show 97ecc58a^:…/render_runner.h` line 3147 is
  byte-identical to the new line. Not a new policy.
- **Nothing else depends on the suppression.** Every surviving reader of `guest_gpu_write_origin()`
  is a printf (`live_renderer.cpp:318`, `render_runner.h:3383`); `grep -rn "gpu-preserving"` leaves
  only the two producers, the header comment and test literals. No behavioural consumer remains.
- **The drain path is genuinely per-write, not ambient.** `PendingGuestGpuWrite` carries `origin`
  per record (`live_renderer.cpp:159-163`), `queue_guest_gpu_write` takes it as data and explicitly
  refuses the thread-local (`:176-188`), and the drain sets/invalidates/clears within one synchronous
  iteration (`:400-409`). The author's "semantics bug, not plumbing bug" diagnosis is correct.
- **The regression test discriminates.** Hand-evaluating the old predicate against arm 1's inputs:
  `byte_preserving=true` → `htile_kill=false` → `continue` at `:3437` → `preserved==0`, both aspects
  still true, so the new assertion fails on all three conjuncts. `preserved == 1` is not polluted by
  cases 8/9 (their planes are `0x20b0…`/`0x20c0…`/`0x20d0…` with `htile = 0`, rejected by the
  zero-base guard). Arm 3's address `0x20FFF00000` genuinely lands in no plane. The explicit
  `depth_valid = true` resets before arms 2 and 3 are load-bearing — without them those arms would
  pass on state arm 1 set.

### Blocking findings (both fixed in `988140b6`)

**F1 — the `## Ruled out` row stated the opposite of the finding.** It read *"eight candidate
mechanisms … every one of them fires ZERO times"* above a **seven**-row table whose last row is the
root cause at **59,999**. The commit message and PR body were right; only the do-not-redo row — the
artifact the charter calls the most expensive knowledge in the repo — was wrong.

**F2 — the cited SHA is not on master.** `git merge-base --is-ancestor 1b5b9471 origin/master` → no.
It is a pre-squash branch commit; what landed is **`97ecc58a`**. A fresh clone cannot `git show` the
cited SHA, which defeats a do-not-redo row. `1b5b9471` was also mis-titled "the GTA V rendering
foundation" — that is its parent `c110fcea`.

### Non-blocking

**N1 — the strongest evidence was already on master and uncited.** `GTA5_STATUS.md` asks whether *"a
write that provably preserves guest bytes [should] invalidate a detached Vulkan depth image"* and
answers **"yes, it must — byte equality with guest memory says nothing about equality with a
renderer-owned image the renderer has since drawn into, and Dead Cells #611 is the counterexample
where sparing it makes gameplay geometry disappear. That is why no preservation policy is proposed
here."** `97ecc58a` shipped that preservation policy anyway. This is a better argument than the
GTA V A/B, and it reframes the regression as contradicting a recorded decision. Now cited in both rows.

**N2 — *Dead Cells* is the named counterexample and was not checked.** Running its guard now; result
will be posted below.

**N3 — the GTA V oracle is weak, and the PR says so.** Peak colour coverage on a title whose 3D world
does not render (rung 2, #2481) is largely a HUD measurement and is insensitive to depth loss. No
frame-time comparison is reported either, and the suppression came from a series named "Performance
rendering foundation" — restoring ~3,177 invalidations per 540 s route has an unmeasured cost.
Correctness-first plus "this is the pre-existing behaviour" makes that acceptable; "GTA V is not
traded away" should not harden into a performance claim.

**N4 — why 313 green tests shipped a black frame.** The #611 guard for exactly this class
(`test_gpu_execute.cpp:639`) drives the **plain** notifier, so its origin is `"unknown"` and the
`"gpu-preserving"` comparison was false in every arm. It could not reach the state it guards. Now
recorded in `BLUE_PRINCE_STATUS.md`.

APPROVED

PR #3103, opened 2026-08-28
feat(snapshot): scan for the frame instead of pinning the guest clock

mattias800:
The deterministic clock was a correctness compromise, and the project owner named it as one.

`PROSPER_DET_CLOCK` replaces **every** time source the guest has — `sceKernelReadTsc`,
`GetProcessTime`, and the wall-clock anchor behind `CLOCK_REALTIME`, `gettimeofday`, `time()` and
`sceRtc*` — with `anchor + flips × (1/DET_FPS)`. A guard recorded under that clock **tests a machine
nobody plays on**, and GRIS proves the divergence is real: its opening FMV freezes with the clock on
(1,680 frames and 47 forced-submit stalls, against 42,000 and 0 without it).

## The fix is in the matcher, not the clock

The clock existed only to stop anchors drifting when the host renders at a different rate than during
authoring. **SHIFT+F6 / SHIFT+F7** now mark a snap to *scan* for: the check sweeps a wide span forward
of the anchor instead of expecting the frame at a fixed offset.

One mechanism covers **slow test runners, variable loading, FMVs and frame-rate differences** at once
— and normal play is left completely alone: real clock, uncapped, 120 Hz fine, which is what a PS5
actually does.

**The idea is the owner's — including the objection, which turned out to be right.** They asked
whether this would disturb the pad route's timing. An earlier revision of this PR said it would not,
on the reasoning that waiting to *capture* is passive. That reasoning is correct in isolation and
the conclusion was still wrong, because the check had stopped pacing its flips: a flip-anchored
route replayed at a different flip rate lands its presses at different guest times. Measured on
`alexkidd` — **all four snaps failed**, ssim 0.001–0.522, colour counts collapsing 14,548 → 108,
which is a different part of the game rather than a shifted anchor.

The fix is that **both halves now pace** (`PROSPER_FLIP_PACE_FPS`), which is what lets the clock stay
off at all. With that in place the capture is passive as described.

## Design details that are load-bearing

- **Bounded and forward-biased.** The bound exists so a check cannot run for an unbounded time, and
  the anchor stays as the ordering hint. (An earlier revision justified the bound as protection
  against a recurring scene matching its first occurrence; that is **retracted** — `best_match`
  takes the best score across the span, not the first hit, so a repeat inside the span is resolved
  by score. The reasoning is corrected in the code and tests as well as here.)
- **Uniform sampling across a scan**, where the tight window is geometric: a scan is used precisely
  when the match is *not* near the anchor, so weighting toward the anchor would spend samples in the
  least likely place.

## A defect this PR introduced and fixed

Worth recording because it is the failure mode the PR is about. The first version of the `--append`
guard built the run environment from the command line and only *afterwards* decided to keep the
directory's stored parameters — so `--append --det-fps 60` on a directory authored at 30 launched
the emulator paced at **60** while recording **30**, printing both "flips paced to 60/s" and
"keeping the original session parameters (det_fps=30)" in the same run. The pre-fix code corrupted
the *first* run's snaps; the fix corrupted the *second* run's instead. The damage moved rather than
went.

`reconcile_session` now runs before the environment is built, and separates two cases the old code
conflated: a parameter the caller did not ask for is **adopted** from the session and written back
onto `args` so the run matches, while one they explicitly passed that contradicts the session is
**refused** — a directory's snaps share one route and one flip axis, so it can only describe one set
of conditions.

## Consequences

`--det-clock` now defaults to **off**. The **flip pacer does not go with it** — it runs
unconditionally on both sides, and it is what *replaced* the clock rather than a companion to it.
Pacing buys the "same flip = same guest time" correspondence honestly: the guest keeps a real clock
and simply flips at a fixed rate, the way a vsync-locked console does, so nothing lies to it about
what time it is. `--det-clock on` remains available per title.

**Its precondition, stated because it is real:** pacing can only slow a fast host *down*. The wait
sleeps when ahead and re-anchors when behind, so on any title/host that cannot **sustain** `det_fps`
the pacer is inert and anchors drift again — Blue Prince measures 180 fps at its menu, 20 in
gameplay and 4.8 during its FMV. That is precisely the case **scan** snaps exist for.

A set stored before the `det_clock` key existed reads as **clock ON**, since there was no way to
author one with it off.

## Verification

`ctest --no-tests=error -j4` → **315/315**, exit 0, on the current head.

**End-to-end**, re-measured on this exact head (the earlier figures were from the superseded
commit): the `alexkidd` guard replays **4/4** — `ssim 1.000 / 0.983 / 0.978 / 1.000`, every snap
matching at offset **+0**. On master one anchor needed +80. GRIS runs 29,520 frames, TRC=1.

**The new default path is now measured too, which review N2 correctly said it was not.** The
committed `alexkidd.json` predates the `det_clock` key, so it replays clock-**ON**; that validated
the reader fallback but said nothing about the real-clock+paced default the tool now writes. Running
the same route and the same four anchors with `det_clock: "off"`: **4/4**, `ssim 1.000 / 0.973 /
0.983 / 1.000`. One anchor matched at −80 rather than +0 — real drift, comfortably inside the
window, and exactly what the window is for. So both configurations pass, and the new default is no
longer an untested path.

**Every fix is mutation-verified at the level the fix lives.** Each was deleted or reverted in turn,
on a cleared `__pycache__` with `python3 -B`, with the mutation asserted to have applied before the
run. All 16 redden; none aborts.

| mutation | reddens |
| --- | --- |
| `clock_enabled` fallback → `"off"` | 2 |
| `replay_env` drops its pacing line | 9 |
| `author_env` drops its pacing line | 7 |
| author paces at `det_fps * 2` (halves disagree) | 7 |
| `scan_forward` clamp floor removed | 2 |
| `cmd_import` ignores `session.json` | 17 |
| `session_record` keys renamed | 25 |
| check scratch → `/tmp` | 1 |
| `--append` does not write adopted values back onto args | 15 |
| **`cmd_author` drops the conflict refusal** | 7 |
| **`cmd_author` drops the adopted NOTE** | 1 |
| **stored-value coercion removed** | 1 |
| **`cmd_author`'s non-dict session guard removed** | 4 |
| **`reconcile_session`'s non-dict guard removed** | 3 |
| `_flag_was_passed` → exact-match | 2 |
| `author_env` inlined back into `cmd_author` without pacing | 3 |

144 assertions pass unmutated.

**The bolded rows are what review round 5 found, and they were all one mistake made three times.**
Rounds 3, 4 and 5 each caught the same structural error: a helper's *return value* asserted while
the command's *use* of it was not. The fix was correct every time and the guard around it could be
deleted with the suite fully green.

So rather than patch the instance, I audited every fix in the PR by deleting its command-level
action and asking whether anything noticed. Three were silent. The arms now drive `cmd_author`
through the **real argparse** — `build_parser` is extracted from `main` for exactly this — with
`subprocess.run` spied on, and assert on the environment the process actually received, whether it
was launched at all, and what `session.json` holds afterwards. A hand-built args object cannot see
a defaulting bug or a flag whose `dest` does not match what the code reads.

**A note on method, since it bit twice.** A mutation reporting "still green" must be checked before
it is believed. Once a stale `__pycache__` made one mutation replay another's failures; once a
harness sliced the file between two mis-ordered offsets and mangled it rather than reverting the
function, which also read as green. Both were **void** results, not negative ones. Every row above
was derived after clearing the cache and asserting the mutation applied and still parsed.

### DIFF
--- a/prosper/frontends/prosper-app/main.cpp
+++ b/prosper/frontends/prosper-app/main.cpp
@@ -1608,6 +1608,7 @@ int main(int argc, char** argv) {
     // comment on prosper_pad_flip_ordinal() in hle_pad.cpp for what that would corrupt.
     const std::string snapDir = getenv("PROSPER_SNAP_DIR") ? getenv("PROSPER_SNAP_DIR") : grabDir;
     std::optional<prosper::frontend::SnapVerdict> pendingSnapVerdict;
+    prosper::frontend::SnapMode pendingSnapMode = prosper::frontend::SnapMode::anchor;
     uint32_t snapCount = 0;
     // Check side: capture the presented frame as each authored anchor arrives. Sorted ascending by
     // the parser, so one forward-only cursor covers the run -- and a target already passed when the
@@ -1829,7 +1830,9 @@ int main(int argc, char** argv) {
     auto flushPendingSnap = [&](const uint8_t* rgba, uint32_t w, uint32_t h) {
         if (!pendingSnapVerdict) return;
         const prosper::frontend::SnapVerdict verdict = *pendingSnapVerdict;
+        const prosper::frontend::SnapMode mode = pendingSnapMode;
         pendingSnapVerdict.reset();
+        pendingSnapMode = prosper::frontend::SnapMode::anchor;
         const int64_t flip = prosper_pad_flip_ordinal();
         const uint32_t index = snapCount++;
         const std::string name = prosper::frontend::snap_file_name(index, verdict, flip);
@@ -1845,7 +1848,7 @@ int main(int argc, char** argv) {
             (std::filesystem::path(snapDir) / "snaps.jsonl").string();
         if (FILE* f = fopen(manifest.c_str(), "a")) {
             std::fprintf(f, "%s\n", prosper::frontend::snap_record_line(
-                index, verdict, flip, gpu::present_count(), w, h, name,
+                index, verdict, mode, flip, gpu::present_count(), w, h, name,
                 grabNamer.title_id()).c_str());
             fclose(f);
         } else {
@@ -1854,7 +1857,7 @@ int main(int argc, char** argv) {
                          name.c_str(), manifest.c_str());
         }
         std::fprintf(stderr, "%s\n",
-                     prosper::frontend::snap_author_line(index, verdict, flip, name).c_str());
+                     prosper::frontend::snap_author_line(index, verdict, mode, flip, name).c_str());
     };
 
     auto flushPendingActual = [&](const uint8_t* rgba, uint32_t w, uint32_t h) {
@@ -2210,6 +2213,13 @@ int main(int argc, char** argv) {
                         pendingSnapVerdict = ev.key.key == SDLK_F6
                             ? prosper::frontend::SnapVerdict::correct
                             : prosper::frontend::SnapVerdict::incorrect;
+                        // SHIFT marks the frame as one to SCAN for rather than expect at a fixed
+                        // offset -- for anything sitting after a loading screen, an FMV, or any
+                        // other stretch whose length depends on the machine. The person pressing
+                        // the key is the only one who knows that, and they know it right now.
+                        pendingSnapMode = (ev.key.mod & SDL_KMOD_SHIFT)
+                            ? prosper::frontend::SnapMode::scan
+                            : prosper::frontend::SnapMode::anchor;
                     }
                     continue;
                 }
--- a/prosper/frontends/prosper-app/snap_author.hpp
+++ b/prosper/frontends/prosper-app/snap_author.hpp
@@ -33,6 +33,22 @@ enum class SnapVerdict {
     incorrect,  // F7 — "the game does NOT look right here"
 };
 
+// How the check should LOOK for this frame again.
+//
+// `anchor` searches a tight window around the recorded flip and is right for a frame whose position
+// is stable. `scan` sweeps a wide span forward and is for a frame that sits after something of
+// variable length -- a loading screen, an FMV, anything whose duration depends on the machine. It is
+// the person authoring who knows which is which, at the moment they press the key, so the choice is
+// theirs: F6/F7 anchor, SHIFT+F6/F7 scan.
+enum class SnapMode {
+    anchor,
+    scan,
+};
+
+inline const char* snap_mode_token(SnapMode mode) {
+    return mode == SnapMode::scan ? "scan" : "anchor";
+}
+
 inline const char* snap_verdict_token(SnapVerdict verdict) {
     return verdict == SnapVerdict::correct ? "correct" : "incorrect";
 }
@@ -89,6 +105,7 @@ inline std::string snap_json_escape(const std::string& value) {
 // that point readable, instead of a truncated array that parses as nothing.
 inline std::string snap_record_line(uint32_t index,
                                     SnapVerdict verdict,
+                                    SnapMode mode,
                                     int64_t flip,
                                     uint64_t guest_present,
                                     uint32_t width,
@@ -99,6 +116,8 @@ inline std::string snap_record_line(uint32_t index,
     out += std::to_string(index);
     out += ",\"verdict\":\"";
     out += snap_verdict_token(verdict);
+    out += "\",\"mode\":\"";
+    out += snap_mode_token(mode);
     out += "\",\"pad_flip\":";
     out += std::to_string(flip);
     out += ",\"guest_present\":";
@@ -120,10 +139,12 @@ inline std::string snap_record_line(uint32_t index,
 // something about it.
 inline std::string snap_author_line(uint32_t index,
                                     SnapVerdict verdict,
+                                    SnapMode mode,
                                     int64_t flip,
                                     const std::string& file) {
     std::string out = "[snap] ";
     out += snap_verdict_token(verdict);
+    if (mode == SnapMode::scan) out += " (scan)";
     out += " #";
     out += std::to_string(index);
     if (flip < 0) {
--- a/prosper/frontends/prosper-app/test_snap_author.cpp
+++ b/prosper/frontends/prosper-app/test_snap_author.cpp
@@ -4,6 +4,8 @@
 #include <cstdio>
 #include <string>
 
+using prosper::frontend::SnapMode;
+using prosper::frontend::snap_mode_token;
 using prosper::frontend::SnapVerdict;
 using prosper::frontend::kSnapFlipUnanchored;
 using prosper::frontend::snap_author_line;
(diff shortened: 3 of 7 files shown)


mattias800 (commented):
Independent review of `81f208b3`, read-only (no build/ctest/emulator).

## Verdict: REJECTED — two blocking findings, both cheap to fix

### B1. The `det_clock` compatibility default is flipped under the one committed snap set, silently

`snaps.py:603` changes the reader fallback from `"on"` to `"off"`:

```python
entry.get("det_clock", "off") == "on"      # was "on"
```

`snaps/alexkidd.json` — the only committed set — **has no `det_clock` key**, and was authored with
the clock pinned and flips paced. That is primary evidence, not inference: the file was added by
`8d5f3e15`, where `cmd_author` called `apply_deterministic_clock` **unconditionally** and set
`PROSPER_FLIP_PACE_FPS` unconditionally. The `--det-clock` switch did not exist until `4bf1c80c`, so
there was no way to author it any other way.

After this PR that set replays **unpinned and unpaced** against signatures recorded at a pinned,
paced 60 fps. Nothing migrates it, warns, or errors.

The arithmetic puts it on the boundary. Anchors `853/1645/2405/3134`, window ±900, offsets
`[-900, -369, -80, 0, 80, 369, 900]`. Drift is `anchor × (fps_check/fps_author − 1)`; at the two
headless rates this project has measured (77.1 and 78.2 against a paced 60):

| anchor | @77.1 | @78.2 |
| --- | --- | --- |
| 853 | +243 | +253 |
| 1645 | +469 | +489 |
| 2405 | +686 | +715 |
| 3134 | **+893** | **+951** |

The deepest anchor lands *on* the edge, and outside it at the faster rate. Geometric sampling leaves
no offset between +369 and +900, so even the passing case is a 7-flip-lucky edge match.

**The PR's verification cannot see this.** `snaps.py check` is not registered as a test —
`CMakeLists.txt:160` registers `test_snaps.py` only, and no CI or ctest invokes `check`. So 315/315
is silent on `alexkidd`. No arm exercises `entry.get("det_clock", …)` in either direction.

**Fix:** keep the *reader* fallback at `"on"` (`cmd_import` has written the key explicitly since
`4bf1c80c`, so the fallback can only ever apply to pre-PR entries, all authored with the clock on),
or write `"det_clock": "on"` into `alexkidd.json`. Changing the author/import CLI default to `off` is
fine and desirable; changing the reader fallback is the bug.

### B2. The central claim is unmeasured, and the rebuttal covers the wrong half of the PR

The body answers the "it would disturb the pad route's timing" objection with: *"Waiting to capture is
passive… only the matcher's search range changes."* That is true of the **scan** change — verified,
nothing about capture feeds back into the guest — but the objection bites on the **clock** change
bundled with it, where the answer is the opposite.

Routes are **flip-anchored**: `scripts/alexkidd/route.pad` is `fN = display flips since first pad
poll`. With the clock pinned, flip N is a fixed guest time. With it off, flip N is a fixed *host*
time, so the same route delivers `cross` at a different point in the state machine on a faster host.
A 6-flip press window at f1740 is 29.0 s in at 60 fps and 22.3 s at 78. If a time-based intro is
still running the press is **swallowed**, and the route **diverges** rather than drifts — which no
amount of scanning fixes, because a scan changes where you look, not what happened.

This is the PR series' own stated premise, from `8d5f3e15`: *"Flip anchoring assumed 'the same flip is
the same guest time'. Without a pinned clock that is false."* The diff removes the mechanism that made
it true and addresses only the matching half.

The direction is not wrong — pinning every guest time source is a real compromise and GRIS is a real
counterexample. The problem is that a **measured** mechanism is replaced by an **unmeasured** one,
with no end-to-end `snaps.py check` at this head, on any set.

**Ask:** run `snaps.py check alexkidd` on this head and post the four SSIM/offset lines. That single
result resolves B1 and B2 together.

## Non-blocking

**N1. The boundedness reasoning is wrong in all three places it appears** (`snaps.py:511-520`,
`AGENTS_SNAPS.md`, `test_snaps.py:313-315`). All three say an unbounded search would "lock onto the
first occurrence" — but `best_match` takes the **argmax over all sampled offsets**, so that hazard is
not a behaviour the code has, bounded or not. What the bound buys is finite search cost. The real
hazard, a scene recurring *inside* the span, is untouched and unmentioned: `DEFAULT_SCAN_FORWARD =
12000` is ~200 s at 60 fps, easily long enough to contain the menu revisit the comment uses as its
own example. A defensible conclusion reached by a derivation that does not establish it.

**N2. `scan_offsets` overshoots its configured span when samples exceed it.** `step = max(1, span //
(samples-1))` floors at 1, so `{forward:10, back:0}` yields offsets `0…32` — 3.2× the span — and
`{forward:0, back:0}` yields `0…32` for a zero span. The test's own invariant
`max(sc) <= scan_forward + 1` is false for those configs; it passes only because the test supplies
the one config where it holds. Suggest clamping.

**N3. The scan defaults are exercised by nothing.** `scan_forward/back/samples` have no CLI flag and
are never written by `cmd_import`, so every real scan snap uses the defaults — while the test supplies
all three explicitly. Setting `DEFAULT_SCAN_FORWARD = 0` would leave the suite green.

**N4. One scan snap lengthens the whole replay** — `last_anchor = max(candidate_flips(entry))`, so the
run continues to `anchor + 11996` even when the frame matched at +0. On a title whose FMV renders at
~4.8 fps that is ~2,500 s; hitting the deadline makes *later* snaps report `NOT REACHED`, i.e. a
failure.

**N5. False-positive surface grows with sample count** — max SSIM over 34 samples vs 7, on a
144-value thumbnail at a 0.85 floor. Suppressing `at_edge` for scans is **correct** (a scan offset
almost always exceeds `flip_window`, so the message would be nonsense) but leaves a scan with no
"matched suspiciously far out" signal.

**N6.** `apply_deterministic_clock(..., enabled=True)` keeps a `True` default while the tool's stance
is now off-by-default. Both call sites pass explicitly, so it is a trap for the next caller only.

**N7. The pacer is retired on the wrong argument.** The justification is about authoring *feel*, which
is correct — but the pacer's other job is making flips correspond to seconds so a flip-anchored route
lands at the same guest time, which is the *checker's* problem. Pacing **both** halves at `DET_FPS`
with a **real** clock is an option the PR does not consider, and it would preserve flip↔time
correspondence at real speed; GRIS's collapse was attributed to per-flip time, not to the pacer.

**N8.** Cosmetic: `test_snap_author.cpp:67, 85, 100` now run 128-141 chars where the file wraps near
100, from the mechanical parameter insertion.

## Verified correct

`scan_offsets` never produces duplicates and is never empty; `max(2, …)` correctly prevents the
`samples-1 == 0` division (0, 1 and 2 all yield `[-900, 0, 12000]`); `candidate_flips` drops negatives
(anchor 5 in scan mode → 31 of 34 offsets, minimum 5); `window_offsets`/`best_match` handle a missing
`mode` key so every pre-PR snap keeps its old behaviour; `pendingSnapMode` cannot leak between snaps
(assigned and cleared alongside `pendingSnapVerdict`, and a press while armed is rejected before
either is touched); every caller of the changed signatures was updated; C++ arm 11 asserts both the
positive and the negative; no other docs are made stale.

REJECTED

Issue #2818, opened 2026-08-21
IMAGE_LOAD_MIP has no faithful lowering — the guest's own mip levels are never materialised (Sonic Frontiers' three scene-width stage kernels)


mattias800:
## Summary

`IMAGE_LOAD_MIP` (MIMG opcode `0x01`) has no faithful lowering. `rdna2_emit_alu.cpp` accepts it in
exactly one case — when the mip operand is **provably zero** *and* the resource is single-level —
and then specialises the mip away, taking the ordinary `image_load` path. Every other form declines
`mode=unresolved-operand`, which is correct and fail-visible, and is now the **single remaining
blocker for all three of Sonic Frontiers' scene-width Cyber Space compute programs**.

The ungated `[mimg-mip]` line names the exact state, identically on all three (live, routed,
`PROSPER_COMPUTE_TRANSLATE_ONLY=1`):

```text
[mimg-mip] image_load_mip declined pc=… shape=0 proven_zero_mip=0 img_dim=5/1 samples=1
           mips=12 mip_tail=0 compressed=0 array_in_gfx=0
```

So the resource declares a **2D_ARRAY with a 12-level mip chain** while the instruction addresses it
`dim:2D`, and the mip operand is not one of the recognised provably-zero shapes. Neither compression
nor the mip tail is involved.

## Why the existing "prove it zero" route cannot be widened

**Not** because prosper cannot build a mip chain — it can, and an earlier revision of this issue
said otherwise. Three facts, and the third is the one that matters:

* The **graphics** path *does* build chains: `tests/fixtures/render_runner.h:6044-6055` (that file is
  the live offscreen Vulkan backend, included as such by `frontends/shared/live/live_renderer.cpp:39`),
  gated to plain-2D (`img_dim == 1`), depth-1, non-storage, non-RTT, RGBA8 sampled textures declaring
  `declared_mip_levels > 1`. `tests/gpu/test_texture_mip_render.cpp` is a registered ctest asserting a
  declared 3-level chain samples a box-filtered level 2.
* The **compute** path does not: its single `VkImageCreateInfo` for guest images sets
  `ici.mipLevels = 1` unconditionally (`frontends/shared/live/live_compute.cpp:7089`).
* **The graphics chain is GENERATED, not uploaded.** Staging carries level 0 only; levels 1..N-1 come
  from a linear-filtered `vkCmdBlitImage` cascade at upload time (`render_runner.h:6257-6259`, `:7245`).

[...]

Issue #2265, opened 2026-08-07
gpu: IMAGE_ATOMIC_ADD with dim=2D_ARRAY is rejected — Sonic Racing: CrossWorlds exercises it on a default launch


mattias800:
`IMAGE_ATOMIC_ADD` (MIMG op `0x11`) with **`dim=5` (2D_ARRAY)** is rejected by the recompiler.
*Sonic Racing: CrossWorlds* (`PPSA08804`) exercises it on a default launch, and the charter is explicit
that an op the guest exercises is a gap to implement rather than a skip to accept.

## Evidence

Default launch, current master (`a54ad733`), **Linux / RADV**, `PROSPER_DBG=1`:

```
[recompile-reject] pc=751 words=f0440128,0000131c fmt=14 op=0x11 dst=19(kind2)
                   src=28(k2),0(k1),0(k1) dmask=0x1 dim=5 glc=0 len=2 modifier=0 dpp=0
[cfg-recompile-reject] pc=751 fmt=14 op=0x11
```

`fmt=14` is MIMG, `op=0x11` is `IMAGE_ATOMIC_ADD`, `dim=5` is `2D_ARRAY`, `dmask=0x1` is a single
R32_UINT channel. Four occurrences per boot.

## Why it is a genuine gap rather than a deferred variant

The storage-image path **already handles 2D_ARRAY for load and store** — `rdna2_to_spirv.cpp:15911`
records *"Storage load/store handle 1D/2D/3D + 1D/2D_ARRAY (dims 0,1,2,4,5)"*, and both
`b.image_read(...)` and `b.image_write(...)` take `dim, arrayed` and pass an arrayed coordinate.

**Only the atomic sibling is 2D-shaped.** `image_atomic_u32`'s own comment (`:1707`) says *"An integer
atomic on an R32_UINT **2D** storage image"*, and the CPU-side fallback in the same branch computes a
strictly 2D index:

```c
const uint32_t index = b.ibin(Op_IAdd, coords[0],
                              b.ibin(Op_IMul, coords[1], b.uconst(res->width)));
uint32_t active = b.land(b.ucmp(Op_ULessThan, coords[0], b.uconst(res->width)),
                         b.ucmp(Op_ULessThan, coords[1], b.uconst(res->height)));
```

So load/store were generalised over the array dimension and the atomic was not.

## What a fix has to cover

1. **`OpImageTexelPointer` coordinate arity must match the image's arrayed-ness** — a 3-component
[...]

Issue #1891, opened 2026-08-03
[Game tracker] Sonic Frontiers (PPSA03831)


mattias800:
## Current milestone

**Rung 2 — title screen and main menu render; a route reaches Cyber Space gameplay in the guest, on a black world.** The scripted route
`prosper/scripts/sonic-frontiers-PPSA03831/reach-gameplay.pad` reaches `GameModeStage` on the Cyber
Space stage `w6d01` — one hundred streamed terrain sectors, the stage HUD, Cyber Space BGM and a
stage clock that runs (#2791). **The world behind that HUD does not render** (#2790): sixteen of the
stage's thirty-two compute programs never execute, three of them full-screen passes over the scene
target.

> **Rung 3 now requires the gameplay scene to render, not merely to be reached** (2026-08-21). The
> bar is deliberately loose — a scene a person would recognise as the game, with *degraded* still
> counting — but **absent does not count**, so this title is **rung 2** until the world draws. It has
> genuinely got further than a title stuck at a menu; the milestone text carries that, not the rung
> number. Nothing about the title's state changed with this edit, only what the number claims.

Rung 2 was verified earlier on master `7801a5bc` (2026-08-06, #2208).
A default no-input launch reaches the 4K opening logo sequence, the auto-save notice, the **title
screen** and the **main menu**, all published by the live renderer at 3840x2160.

**Engine: Hedgehog Engine 2 (Needle renderer), with SurfRide UI and CRIWARE middleware.** Determined
from the eboot's own embedded shader source paths (`Library\hedgehog\…`, `Library\needle\supportfx\…`)
and the dump's `NeedleShader.pac` / `raw/hedgehog/` / `gedit` asset trees — not assumed from the
publisher. Sonic Origins (#1871) and Sonic Racing: CrossWorlds (#1895) share parts of the stack.

### Screenshot evidence

[![Sonic Frontiers — title screen](https://raw.githubusercontent.com/mattias800/prosper/master/assets/screenshots/sonic-frontiers-title-screen.png)](https://github.com/mattias800/prosper/blob/master/assets/screenshots/sonic-frontiers-title-screen.png)

- **Title screen:** `assets/screenshots/sonic-frontiers-title-screen.png`
- **Main menu:** `assets/screenshots/sonic-frontiers-main-menu.png`
- **Auto-save notice:** `assets/screenshots/sonic-frontiers-autosave-notice.png`
[...]

Issue #1573, opened 2026-07-31
ctest: three dump-parameterized tests fail when -DGAME_DUMP points at any title other than The Messenger


mattias800:
## Summary

Configuring the build with `-DGAME_DUMP` pointed at any title **other than The Messenger** makes three
dump-parameterized ctest tests fail. Nothing is actually broken — the tests hard-assume Messenger content — but
the result is a red `ctest` that looks like a regression, which is a trap for exactly the agents least equipped to
recognise it: someone doing first-boot bring-up on a new title, who has naturally configured against *that* title.

## Failure scenario

```bash
cmake -S prosper -B prosper/build-linux -G Ninja -DGAME_DUMP=<DUMP_ROOT>/PPSA01826-app0   # The Pathless
cmake --build prosper/build-linux -j12
cd prosper/build-linux && ctest -j6
```

Result on master `fdfb37f2`: **157/160 passed**, with these three failing:

- `module_loads_eboot`
- `boot_reaches_first_syscall`
- `real_shader_render`

All three pass immediately when their binaries are pointed at the Messenger dump instead:

```bash
./test_boot_linux        <DUMP_ROOT>/PPSA24651-app0            # exit 0
./test_real_shader_render <DUMP_ROOT>/PPSA24651-app0/eboot.bin # exit 0
```

So the failures are purely "wrong dump for this assertion", not a code defect.

## Why it matters

A bring-up agent's first action on a new title is to configure and build against that title's dump, then run
`ctest` to confirm the tree is healthy before trusting any boot result. Today that produces three red tests with
no indication that the cause is the `-DGAME_DUMP` value. The plausible reactions are all bad: conclude master is
broken, conclude the new dump is corrupt, or spend time bisecting. It also erodes the project's "ctest exit code
is truth" rule (`docs/VERIFICATION.md`) — if a green run requires a specific undocumented dump, the exit code is
not self-describing.

## Suggested fix

Any of these would resolve it; the first is probably cleanest:

1. **Skip cleanly with a clear message** when `GAME_DUMP` is not the title a dump-parameterized test asserts
   against — e.g. the test detects the title id and reports "skipped: requires PPSA24651, configured with
   PPSA01826". ctest supports skip via `SKIP_RETURN_CODE`.
2. **Point these three tests at the Messenger dump explicitly** via their own cache variable
   (e.g. `-DMESSENGER_DUMP=`), independent of the `GAME_DUMP` used for everything else, and register them only
   when that dump is present.
[...]

PR #3109, opened 2026-08-29
recompiler: image_get_resinfo accepts a 2D_ARRAY T#, closing one Frontiers gap

mattias800:
## What this does

`image_get_resinfo` now accepts a **2D_ARRAY** T# (`dim=5`). Sonic Frontiers' Cyber Space stage runs
a compute program that queries one, and prosper declined it.

The SPIR-V lowering was already right. #325 widened `image_get_resinfo` to branch on
`tex_is_arrayed(binding)`, ask for the ivec3 query and report its third component as the layer
**count** — which is what GET_RESINFO's third result means for a 2D_ARRAY T#. What still declined
were the two gates in front of it:

- `emit_alu`'s dim dispatch, whose `else` rejected anything past dim 2;
- the coverage predicate that decides whether the census calls the instruction supported.

That is the same lag #2265 records for the atomic coverage predicate, in the same opcode family.

## Measured

Route + `boot_trace` + `PROSPER_COMPUTE_TRANSLATE_ONLY=1`, isolated `PROSPER_SAVE0`, compared at one
shared census denominator (1,048,576 dispatch decisions over 32 programs):

| | reject site | opcode |
| --- | --- | --- |
| before | pc=**31** | `fmt=14 op=0xe` — `image_get_resinfo` dim=5 |
| after | pc=**130** | `fmt=4 op=0x5` — `s_cbranch_scc1` |

`op=0xe dim=5` goes from one occurrence to **none** anywhere in the run; the program advances 99
instructions to an unrelated gap.

## What this does NOT do, stated plainly

**No image change, and no program leaves the skip list** — 14 listed before, 14 after. This closes one
gap on one program's path; that program is still blocked further along.

The three **scene-width** kernels the world actually depends on are untouched by this. They decline on
`image_load_mip` against a **12-level** mip chain (`[mimg-mip] ... proven_zero_mip=0 mips=12
dmask=0x3 unorm=0 glc=0`), which needs the compute path to stop hard-setting `mipLevels = 1`
(`live_compute.cpp`) plus a non-specialised lowering. That is the real frontier and it is not this PR.

## Verification, including a defect in my own first test

Both halves are pinned by mutation — which is how I found my first version pinned only one:

| mutation | lowering arm | coverage arm | CUBE controls |
| --- | --- | --- | --- |
| dispatch mapping removed | **FAIL** | — | green |
| coverage predicate reverted only | green | **FAIL** | green |
| restored | green | green | green |

The lowering arms alone stayed green when the coverage predicate was reverted, so the coverage seam is
now tested directly through `recompile_coverage()` rather than assumed. The test uses the guest's exact
bytes (`0xf0380328, 0x00091103`) — word1 names VADDR=v3, VDATA=v17, SRSRC at s36, all three of which the
live `[compute] skip` line reports, so it cannot drift from the instruction it is about.

`ctest --no-tests=error`: **315/315 passed**, exit 0. (Four cases report `(Skipped)` because
`-DGAME_DUMP` points at `PPSA03831` rather than `PPSA24651`, per #1573.)

## Scope and risk

CUBE (`dim=3`) is **deliberately still declined**. Two other stage programs issue it (`0xf0380118`,
`dmask=0x1`); admitting it would need the stacked-face lowering (#273) to promise what GET_RESINFO's
third result means for a cube, which this does not establish. A control arm reddens if a later
widening reaches past 2D_ARRAY.

- **CONFIDENCE: HIGH** for the measured case — a dim-5 T# with `depth=1` takes the ivec2 path and
  out[2] keeps its default of 1, the true layer count for a one-layer array.
- **CONFIDENCE: MED** for a **multi-layer non-BC** array, which prosper does not materialize as an
  array (`guest_texture_is_uploaded_array()` requires dim-5 AND depth>1 AND block-compressed). The
  query then answers for the resource as actually uploaded, consistent with what every sample of it
  reads, but it is reasoned rather than measured — no title in the corpus is known to issue one. This
  is the point most worth a reviewer's attention.

Refs #2790, #2818, #1891.

### DIFF
--- a/prosper/docs/SONIC_FRONTIERS_STATUS.md
+++ b/prosper/docs/SONIC_FRONTIERS_STATUS.md
@@ -377,6 +377,35 @@ shapes. `rdna2_emit_alu.cpp` only ever specialises this op away after proving th
 the resource is single-level; a 12-mip resource has no such proof, so it declines. Tracked as
 [#2818](https://github.com/mattias800/prosper/issues/2818).
 
+### `image_get_resinfo` on a 2D_ARRAY T# no longer declines (2026-08-29)
+
+Separate from the three scene-width kernels above, one stage program (`0x200581bb00`, `1x1x6` groups
+of `256x1`) declined at **pc=31** on `image_get_resinfo` with `dim=5` (2D_ARRAY), `dmask=0x3`. The
+SPIR-V lowering had already been widened for the arrayed form by #325 — `image_get_resinfo` branches
+on `tex_is_arrayed(binding)`, asks for the ivec3 query and reports its third component as the layer
+count, which is exactly what GET_RESINFO's third result means for a 2D_ARRAY T#. What still declined
+were the two *gates* in front of it: `emit_alu`'s dim dispatch and the coverage predicate — the same
+lag #2265 records for the atomic coverage predicate.
+
+Measured on the route, `boot_trace` + `PROSPER_COMPUTE_TRANSLATE_ONLY=1`, compared at the same census
+denominator (1,048,576 dispatch decisions over 32 programs):
+
+| | reject site | opcode |
+| --- | --- | --- |
+| before | pc=**31** | `fmt=14 op=0xe` — `image_get_resinfo` dim=5 |
+| after | pc=**130** | `fmt=4 op=0x5` — `s_cbranch_scc1` |
+
+`op=0xe dim=5` goes from one occurrence to **none** anywhere in the run, and the program advances 99
+instructions to an unrelated gap.
+
+**No program leaves the skip list and no image change is claimed** — 14 listed before and after. This
+closes one gap on one program's path; that program is still blocked further along, and the three
+scene-width kernels above are untouched by it.
+
+CUBE (`dim=3`) is deliberately still declined — two other stage programs issue it (`0xf0380118`,
+`dmask=0x1`), and admitting it would need the stacked-face lowering (#273) to promise what
+GET_RESINFO's third result means for a cube, which this did not establish.
+
 ### `image_load_mip` is the FIRST blocker of at least two, not the only one (2026-08-21)
 
 **This section used to say "All three now block on exactly one instruction", and #2818's summary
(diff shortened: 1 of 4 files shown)


mattias800 (commented):
## Author self-review (NOT an independent review — stated so it is not mistaken for one)

CI on this exact head: **9 pass, 2 skipping, 0 failing**. No independent reviewer was available in
this session, so this records what I checked myself. The charter asks for independent review on
recompiler changes; a reviewer arriving later should treat this as the author's checklist, not as an
approval, and the MED-confidence item at the end is where I would look first.

**1. Can this binding be declared two different ways?** This was the risk I considered most likely to
bite, because `declare_texture()` must agree across every use of a binding or the descriptor is a
mismatch. It cannot happen here: the arrayed flag passed in is `res_arrayed`, computed from
`guest_texture_is_uploaded_array(res->img_dim, res->depth, res->format)` — a property of the
**resource**, not of the instruction, which is exactly what #325's comment establishes. Every
instruction touching this binding therefore passes the same flag, whatever dim each one names.

**2. Does the lowering actually handle what the dispatch now sends it?** `image_get_resinfo` branches
on `arrayed_2d = dim == Dim_2D && tex_is_arrayed(binding)` and asks for the ivec3 query in that case,
reporting component 2 as the layer count. Sending it `Dim_2D` for a dim-5 instruction is the case it
was written for by #325. On Frontiers' own resource (`depth=1`) `tex_is_arrayed` is false, the ivec2
path runs, and `out[2]` keeps its initialised default of `uconst(1)` — the correct layer count for a
one-layer array.

**3. Do the two gates agree?** Dispatch and coverage predicate now admit the same set (`dim <= 2 ||
dim == 5`). They disagreed before this change, which is the defect. A disagreement in the other
direction — coverage saying supported while the dispatch declines — would show up as a program that
recompiles in the census and skips live; the mutation table in the PR body exercises each gate
separately, so that asymmetry would redden.

**4. Is the test measuring the thing?** Verified by mutation rather than by inspection, and this
caught a real defect in my first version: the lowering arms stayed green when only the coverage
predicate was reverted, i.e. the coverage seam was untested. It is now tested through
`recompile_coverage()` directly. Both mutations redden exactly one arm each; the CUBE controls stay
green under both.

**5. What I could not verify.** A **multi-layer non-BC 2D_ARRAY**. `guest_texture_is_uploaded_array()`
requires dim-5 AND depth>1 AND block-compressed, so such a resource is materialized as a plain 2D
image and this query reports layer count 1 for it. I argue that is consistent — it answers for the
resource as actually uploaded, which is what every sample of it will read — but it is reasoning, not
measurement, and no title in the corpus is known to issue one. Marked `CONFIDENCE: MED` in the source.
If a reviewer disagrees, the narrowest fix is to decline dim-5 when `depth > 1 && !res_arrayed` rather
than answer for the uploaded form.

Merging under the standing task authorisation (create PR → CI → merge). Flagging #2790 so the next
agent on that lane sees both the closed gap and the untouched `image_load_mip` frontier.

Issue #3119, opened 2026-08-29
sceSystemServiceReportAbnormalTermination is unregistered: a guest that reports its own crash is indistinguishable from a hang


mattias800:
When a guest calls `sceSystemServiceReportAbnormalTermination`, it is telling the system that **it has crashed**. prosper currently does not recognise the NID at all, so the dispatcher's default fires:

```
[prosper] unimplemented: libSceSystemService::3s8cHiCBKBE  -> returning 0
```

and the guest is allowed to continue. The result is a live process with a black window that looks exactly like a hang, a stall, or a render bug.

## Found on

*Tactics Ogre: Reborn* (`PPSA03839`), 2026-08-29, during a human-authored snap session. The title is recorded at rung 3 but produced only a black screen. The log stops at 50 lines with no faults, no compute rejects, no device loss and no `no present source` scanouts — nothing that says "broken" — and the one informative line was that opaque NID.

Resolved against the PS5 3.20 firmware library reference:

```
libSceSystemService.c:3976:  sprx_dlsym(__handle, "3s8cHiCBKBE", &__ptr_sceSystemServiceReportAbnormalTermination)
```

Verified as a genuine call rather than a link-time message: the line is emitted by `prosper_on_unimpl` (`src/hle/dispatch/dispatch.cpp:166`), the call trampoline, which logs on first invocation.

Full report: #1892.

## Why this is worth fixing beyond one title

The cost is diagnostic, and it is paid by every future investigation. In the same sweep, five titles produced a black screen with five *different* underlying signatures (#2867 zero frames, #1891 no present source over 1,024+ flips, #2887 one frame then presentation stops, #2885 3,679 frames of flat white, and this one). Only this one had the guest saying it had failed — and that fact was invisible, because it arrived as an unresolved hash among a dozen other `unimplemented:` lines.

Anyone who does not happen to grep that NID against the firmware reference will file this as a rendering problem and start bisecting the renderer.

## Proposed behaviour

Two parts, the first trivial and the second the actual ask:

1. **Register the NID with its real name**, so the line is self-explanatory even if nothing else changes. It is currently absent from the whole tree except as a guess candidate in `tools/dbg/nidguess.py`.

[...]

Issue #1892, opened 2026-08-03
[Game tracker] Tactics Ogre: Reborn (PPSA03839)


mattias800:
## Current milestone

**Rung 3 — first tutorial battle with real GPU draws and an interactive move command.** On exact master `2601f7587f941111a73fd4a36ecf93d23d0065b7`, a fresh-save route renders the complete **Tactics Ogre: Reborn** title menu at native 1920×1080, completes New Game setup and the Chapter 1 movie, continues through the opening event, and reaches the first tutorial battle. The game enters `Select destination`, consumes routed input, and produces its own `A unit occupies this tile` response.

The previous rung-0 result came from Fedora's deliberately restricted `ffmpeg-free`, which had HEVC disabled. That environment cause is discriminated without a Prosper code change. Full package, stream, build, test, and live evidence is in the [2026-08-03 codec discriminator](https://github.com/mattias800/prosper/issues/1892#issuecomment-5171185909). The exact gameplay route, hashes, run summary, and direct image are in the [tutorial-battle discriminator](https://github.com/mattias800/prosper/issues/1892#issuecomment-5172216155).

### Screenshot evidence

[![Tactics Ogre: Reborn — first tutorial battle](https://raw.githubusercontent.com/mattias800/prosper/master/assets/screenshots/tactics-ogre-reborn-gameplay.png)](https://github.com/mattias800/prosper/blob/master/assets/screenshots/tactics-ogre-reborn-gameplay.png)

- **Title screen:** direct, unmodified native-1080p normal-frontend capture, checked in as [`assets/screenshots/tactics-ogre-title.png`](https://github.com/mattias800/prosper/blob/master/assets/screenshots/tactics-ogre-title.png); PNG SHA-256 `8f519f64d29dec20a969d5039e867dd496136aae2408d2805ab95ded947ea328`.
- **Opening in-engine scene:** direct, unmodified native-1080p normal-frontend capture, checked in by #1912 as [`assets/screenshots/tactics-ogre-reborn-opening-scene.png`](https://github.com/mattias800/prosper/blob/master/assets/screenshots/tactics-ogre-reborn-opening-scene.png); PNG SHA-256 `fae54e8d30f6f3096c4ba4a3a102667e5154fda3c899f79874af6cf4e9ac4cc6`. The bright/outlined central sprite and later horizontal banding remain explicit visual defects.
[...]

PR #3120, opened 2026-08-29
hle(systemservice): log a guest reporting its OWN abnormal termination (#3119)

mattias800:
`sceSystemServiceReportAbnormalTermination` was unregistered, so the dispatcher's default `return 0`
answered it as success while recording nothing. A guest reporting *its own* crash was therefore
invisible in the log — the one event most worth seeing passed silently.

## What ships

The NID is registered and the call is **logged**. The default is **log and continue**; stopping the
run is opt-in via `PROSPER_ABNORMAL_TERMINATION_STOP`. The return stays `0`, identical to the
dispatcher default, so **no guest observes a different answer than before**.

> **This PR body previously described a different design** — stop-by-default with a
> `PROSPER_ABNORMAL_TERMINATION_CONTINUE` escape hatch. That was the first commit's approach; it was
> reworked in the second commit and the body was never updated, so the description and its entire
> "Risk" section litigated a design this PR had already abandoned. `..._CONTINUE` does not exist
> anywhere in the diff. Corrected here after an independent review caught it. The code was always the
> reworked version; only this text was stale.

## Why not stop by default

Stopping by default would have been a **live regression, not a theoretical one**, and the evidence is
what drove the rework:

- A 5-run Tactics Ogre archive shows the call has **zero discriminating power** — the title makes it
  unconditionally at startup, in runs that were fine. One reading consistent with all five logs,
  offered as hypothesis rather than finding: the API may report that the *previous* session ended
  abnormally, and those runs are killed by `timeout`.
- **23 of 56 local dumps import this NID**, four of them rung-6 guarded (*The Messenger*,
  *Blasphemous 2*, *Alex Kidd*, *Blue Prince*).
- `prosper_stop_requested()` is polled **only by prosper-app**. The snapshot matrix drives
  `boot_trace` and `screenshot`, so it **could not have seen the breakage at all** — it would have
  failed only in a human's hands.

That last point is the reason this matters more than the diff size suggests: the change most likely to
be waved through is the one whose damage is invisible to every automated gate.

## Scope, stated because the opt-in is not uniform

- **prosper-app**: winds its run-loop down; `run_entry()` never observes the stop, so the guest thread
  is not torn down.
- **screenshot / boot_trace**: nothing polls the flag, so it does not stop those runs — and it is *not*
  inert there either: `prosper_wait_while_paused()` returns false permanently once stop latches, which
  mutes the SDL3 audio sink in a frontend that keeps running.

## Confidence

`CONFIDENCE: HIGH` on the NID identity. `CONFIDENCE: LOW` on the argument layout — it is logged raw and
no layout is confirmed against any primary source — and on what a call implies about the title's state.

## Verification

`ctest --no-tests=error`: **315/315**, exit 0, on the merge result against current master (the branch
tip alone is behind).

Mutations, both reproduced by the independent reviewer on the merge result:

| mutation | result |
| --- | --- |
| registration removed | 1 FAIL |
| `prosper_request_stop()` call removed | 1 FAIL |

**Correction:** an earlier revision of this body claimed the second mutation produced "2 checks FAIL".
Measured, it is 1. A checkable number stated wrongly is worse than no number, since it is exactly the
kind of claim a reader takes on trust.

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -871,6 +871,11 @@ if(TARGET prosper_core)
   target_link_libraries(test_addcontent_unmount PRIVATE prosper_core)
   add_test(NAME addcontent_unmount COMMAND test_addcontent_unmount)
 
+  add_executable(test_abnormal_termination_stops tests/hle/service/test_abnormal_termination_stops.cpp)
+  target_include_directories(test_abnormal_termination_stops PRIVATE tests)
+  target_link_libraries(test_abnormal_termination_stops PRIVATE prosper_core)
+  add_test(NAME abnormal_termination_stops COMMAND test_abnormal_termination_stops)
+
   add_executable(test_addcontent_mount_report tests/hle/service/test_addcontent_mount_report.cpp)
   # Includes a shared fixture, which resolves from the `tests` root.
   target_include_directories(test_addcontent_mount_report PRIVATE tests)
--- a/prosper/src/hle/dispatch/nid.cpp
+++ b/prosper/src/hle/dispatch/nid.cpp
@@ -143,6 +143,12 @@ const std::vector<std::string>& builtin_symbol_names() {
         "pthread_cond_wait","pthread_cond_signal","pthread_condattr_setclock","pthread_condattr_getclock",
         "pthread_condattr_setpshared","pthread_condattr_getpshared","pthread_self","pthread_once",
         // libkernel memory
+        // libSceSystemService. Registering a handler names a NID only inside the HLE registry; the
+        // NidDb this list feeds is what self_dump, nid_census and every other NID-resolving
+        // diagnostic read, so a name absent here still prints as a bare hash everywhere else. This
+        // one mattered: a title reporting its own crash showed up as `3s8cHiCBKBE` and needed a hand
+        // lookup against the PS5 3.20 stub table to read at all (#3119).
+        "sceSystemServiceReportAbnormalTermination",
         "sceKernelAllocateDirectMemory","sceKernelReleaseDirectMemory","sceKernelMapDirectMemory",
         "sceKernelMapNamedFlexibleMemory","sceKernelMapFlexibleMemory","sceKernelReserveVirtualRange",
         "sceKernelMunmap","sceKernelMmap","sceKernelAvailableDirectMemorySize","sceKernelGetDirectMemorySize",
--- a/prosper/src/hle/service/hle_service.cpp
+++ b/prosper/src/hle/service/hle_service.cpp
@@ -3538,52 +3538,55 @@ HLE(s_syss_safearea) {
     return 0;
 }
 
-// sceSystemServiceReportAbnormalTermination(cause) — THE GUEST IS TELLING US IT HAS CRASHED.
+// sceSystemServiceReportAbnormalTermination(cause) — the guest reporting a crash to the system.
 //
-// This is not a service prosper provides to the guest; it is the guest reporting its own abnormal
-// termination to the system. Unregistered, the dispatcher's default fired -- one opaque
-// `unimplemented: libSceSystemService::3s8cHiCBKBE -> returning 0` line among a dozen others -- and
-// the guest was allowed to carry on, so a self-reported crash presented as a live process with a
-// black window: indistinguishable from a hang, a stall, or a renderer defect.
-//
-// Found on Tactics Ogre: Reborn (PPSA03839, #1892), where it was the ONLY informative line in a
-// 50-line log with no faults, no compute rejects, no device loss and no missing present source.
-// Four other titles in the same sweep produced black screens with four different signatures; this
-// was the one where the guest said it had failed, and that fact was invisible.
-//
-// NID verified in the PS5 3.20 stub table:
+// Registered so the line is READABLE. Unregistered, the dispatcher default fired as one opaque
+// `unimplemented: libSceSystemService::3s8cHiCBKBE -> returning 0` among a dozen others, and
+// resolving it needed a hand lookup against the PS5 3.20 stub table:
 //   libSceSystemService.c:3976  sprx_dlsym(__handle, "3s8cHiCBKBE",
 //                                          &__ptr_sceSystemServiceReportAbnormalTermination)
 //
-// The return stays 0, exactly what the dispatcher default already produced, so what the guest
-// observes is unchanged and this cannot regress a title by answering differently. The only new
-// behaviour is that we SAY so and stop.
+// IT DOES NOT MEAN THE TITLE IS FAILING NOW, and an earlier revision of this handler assumed it did.
+// Independent review of #3120 opened five archived Tactics Ogre (PPSA03839) runs: the call is in ALL
+// of them, always at line 8, during boot before the first frame -- including the run that then
+// rendered 40,936 frames over 470 s and reached the tutorial battle, and the runs that stalled at
+// ~frame 993. `prosper_on_unimpl` logs only on FIRST invocation, so that is the title's one call.
+// It therefore has zero discriminating power: it is something this title does unconditionally at
+// startup. One reading consistent with all five logs, offered as hypothesis and not finding: the API
+// may report that the PREVIOUS session ended abnormally, and these runs are killed by `timeout`.
 //
-// Stopping uses the existing cooperative signal (host/platform/lifecycle.hpp). Note what that does
-// and does not do: run_entry() never observes it (game_path.hpp:82), so the guest thread is not
-// torn down -- the frontend run-loop winds down at its next boundary instead. That is the right
-// scope. Set PROSPER_ABNORMAL_TERMINATION_CONTINUE=1 to keep running anyway, which is what you want
-// when attaching a debugger to the post-crash state.
+// So the default is to LOG AND CONTINUE. Stopping is opt-in via PROSPER_ABNORMAL_TERMINATION_STOP.
+// Stopping by default would have been a live regression rather than a theoretical one: 23 of the 56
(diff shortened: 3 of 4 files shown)


mattias800:
## Reworked after review — every blocking finding addressed

The review was right and the original shape is withdrawn. Thank you for opening the archived runs; that is what caught it.

### 1. The premise was false (blocking)

Fixed by **inverting the default**: log always, stop only under `PROSPER_ABNORMAL_TERMINATION_STOP=1`.

The five archived *Tactics Ogre* runs settle it — the call is in all of them at line 8 during boot, including the run that then rendered **40,936 frames over 470 s** and reached the tutorial battle. It has zero discriminating power. Your reading that it may report the *previous* session's termination is now recorded in the code comment as a hypothesis, along with the self-sustaining-trap consequence it would have had.

I have also **retracted the claim publicly on #1892**, where I had reported the black screen as a self-reported crash.

### 2. Blast radius (blocking)

Evaporates with the default inverted, and the finding is recorded in the comment so nobody re-introduces it: 23 importing dumps, four rung-6 guarded, and `prosper_stop_requested()` polled only by prosper-app while the snapshot matrix drives boot_trace and screenshot — so no guard in the project could have seen it.

### 3. Scope was wrong (major)

Corrected, including the part I would not have found: latching stop is **not inert** on frontends that ignore it, because `prosper_wait_while_paused()` returns false permanently once set and mutes the SDL3 audio sink in a run that continues.

### 6. Name still absent from NidDb (minor)

Confirmed and fixed — `grep -c '"sceSystemService' src/hle/dispatch/nid.cpp` was **0**. Added to `builtin_symbol_names()` with a comment explaining that registering a handler names a NID only inside the HLE registry.

### Not yet addressed

Findings **4** (exit status 0 on a stopped run), **5** (`=0` means continue) and **7** (log wording) — 7 is largely overtaken since the message no longer asserts the title is crashing. 4 and 5 are worth doing; say the word and I will fold them in, or I can take them separately.

### Verification

`ctest --no-tests=error`: **315/315 passed**, exit 0, on the reworked head. Tests inverted to match the new contract: default arm asserts no stop, opt-in arm asserts the stop, third arm clears the variable to prove arm 2 was the variable and not a latched signal.

mattias800 (commented):
REJECTED (documentation-only defect; the code itself is sound — see below for exactly what to fix)

## What I verified

**Branch currency.** Head is 25 commits behind `origin/master` (merge-base `543b7a7f`);
`hle_service.cpp` changed 42 lines on master in the interval. I built and tested the actual merge
result the same way as for #3136 (`git merge-tree` → `commit-tree`, no textual conflicts): clean
build, `ctest --no-tests=error -j4` → **100% tests passed, 315/315** (314 baseline + this PR's one
new executable). Current and green on the merge result, not just the stale branch tip.

**Mutation tests**, on the merge-result worktree, both restored afterward:

- **Registration removed** (the original bug — `Hle::register_fn("3s8cHiCBKBE", ...)` deleted):
  rebuilt, ran `test_abnormal_termination_stops` →
  `FAIL: sceSystemServiceReportAbnormalTermination is registered ...` — **1 check FAIL**, exit 1.
  Matches the PR's claim exactly.
- **`prosper_request_stop()` call removed**: rebuilt, ran the test →
  `FAIL: PROSPER_ABNORMAL_TERMINATION_STOP=1 opts into stopping the run` — **1 check FAIL**, exit 1.
  **The PR's own mutation table and its second commit's message both claim "2 checks FAIL" for
  this exact mutation.** I re-ran it twice to be sure; it's 1. Minor and not blocking on its own
  (the test still correctly reddens for the right reason), but it's a checkable claim in the PR
  body that's wrong, and per the charter a cited number should be re-derived, not restated —
  correct it in the same edit as the item below.

The test is not vacuous — both mutations produce a real, distinct failure, and restoring the file
returns the suite to green.

## The blocking finding: the PR's own description is stale relative to its own shipped code

The PR has two commits. The first implements **stop-by-default**, with `getenv` gating an opt-out.
The second, "Rework after review: log by default, stop only on opt-in," replaces that design after
independent review found the premise false (the call fires unconditionally at Tactics Ogre's boot
in all 5 archived runs, including ones that go on to render 40,936 frames — it has zero
discriminating power — and 23 of 56 local dumps import this NID, 4 of them rung-6 guarded, so
stop-by-default would have been a live regression with no CI guard able to see it, since
`prosper_stop_requested()` is polled only by `prosper-app`).

The **shipped code matches the second design**: I grepped it directly —
`getenv("PROSPER_ABNORMAL_TERMINATION_STOP")` is the only env var read anywhere in
`hle_service.cpp` or the test, default is log-and-continue, and `PROSPER_ABNORMAL_TERMINATION_STOP=1`
is opt-in to stop. There is no `PROSPER_ABNORMAL_TERMINATION_CONTINUE` anywhere in the diff.

**But the PR's top-level body (`gh pr view 3120`) still describes the first design.** Verbatim from
the current body:

> Stopping reuses the existing cooperative signal... `PROSPER_ABNORMAL_TERMINATION_CONTINUE=1` keeps
> the process alive, for attaching a debugger to the post-crash state.
>
> ## Risk
> Low, with one judgement call worth a reviewer's attention: **should a self-reported crash stop the
> run by default, or only warn?** This PR stops, on the argument that...

None of this matches the code. `PROSPER_ABNORMAL_TERMINATION_CONTINUE` doesn't exist; the PR does
**not** stop by default; and the entire "Risk" section is litigating a design decision the PR's own
second commit already reversed for exactly the reason a reviewer would raise. Per the charter, a PR
description must be self-contained and record the actual behavioral contract and risk — this one
records the abandoned contract. Anyone merging from the description alone (which is explicitly part
of this project's merge process — "the author posts exact-head evidence") would believe the wrong
default ships and would be answering the wrong risk question.

This is exactly the "cited claim vs. actual code" gap the charter singles out, except here the stale
claim is the PR's own summary of its own diff rather than a reviewer's citation. It should be caught
before merge for the same reason: the description is what gets read and trusted, not the diff.

## Answering the review brief's question directly

Does this turn a survivable condition into a hard stop for titles with spurious/non-fatal crash
reports? **As shipped, no** — that's exactly what the rework fixed, and it's well-evidenced (the
5-run Tactics Ogre archive, the 23/56-import census, the CI-blind-spot argument). The design is
right. It's the PR's *description* of that design that's wrong.

## Other notes, not blocking

- `nid.cpp`'s addition of the name to `builtin_symbol_names()` is correct and directly serves the
  stated goal (NidDb resolution for `self_dump`/`nid_census`/etc.) — verified it's the only such
  addition in the diff and matches the NID.
- The handler now runs on **every** call (it's registered, not falling to the once-only
  `prosper_on_unimpl` default), so a title that calls this NID repeatedly would get repeated
  `fprintf`+`fflush(stderr)` — a minor log-noise/cost consideration the PR doesn't mention. Not
  blocking: every piece of evidence in the PR and the issue shows this call happening once, at boot,
  which is the realistic case.

## Required before merge

1. Rewrite the PR body's "Behavioural contract" and "Risk" sections to match the shipped design:
   default is log-and-continue; `PROSPER_ABNORMAL_TERMINATION_STOP=1` opts into stopping; no
   `_CONTINUE` variable exists.
2. Correct "2 checks FAIL" to "1 check FAIL" for the stop-call-removed mutation in both the PR body
   and, if it's meant to be a durable record, the second commit's message (or a follow-up note).

Once the description matches the code, this is ready to merge — the implementation, tests, and the
default-safety argument are sound and independently verified above.

mattias800 (commented):
**APPROVED**

The independent review's rejection was correct and its finding is now fixed: the PR body described
the **first** commit's design (stop-by-default, `PROSPER_ABNORMAL_TERMINATION_CONTINUE`) while the
shipped code is the **second** commit's rework (log-by-default, `PROSPER_ABNORMAL_TERMINATION_STOP`
opt-in). I verified this independently rather than on report — the diff contains only
`..._STOP`, and `..._CONTINUE` appears nowhere in it. The title said "stops the run" for the same
reason; both are corrected.

**The code was never the problem, and the reviewer said so plainly.** As shipped this does not turn a
survivable condition into a hard stop — the rework is precisely what prevents that.

**What makes the rework right is the reasoning behind it, which is unusually good:**

- The 5-run Tactics Ogre archive showing the call has **zero discriminating power** — the title makes
  it unconditionally at startup, in runs that were fine. Registering a NID and acting on it is the
  attractive move; establishing first that the signal carries no information is the one that gets
  skipped.
- **`prosper_stop_requested()` is polled only by prosper-app**, so the snapshot matrix — which drives
  `boot_trace` and `screenshot` — **could not have seen the breakage at all**. Stopping by default
  would have shipped green and failed only in a human's hands, across 23 of 56 importing dumps
  including four rung-6 guarded titles. A change whose damage is invisible to every automated gate is
  exactly the one that needs this argument written down.
- The scope note is the part I'd have expected to be missing: the flag is **not inert** on the
  non-polling frontends, because `prosper_wait_while_paused()` returns false permanently once stop
  latches, muting the SDL3 audio sink in a frontend that keeps running. That is a real consequence of
  an opt-in nobody would think to check.

**Return value unchanged at `0`**, identical to the dispatcher default, so no guest observes anything
different — the change is purely observational, which is what makes it safe to land on a title-facing
path.

**`CONFIDENCE` markers are honest and correctly split:** HIGH on the NID identity, LOW on the argument
layout (logged raw, no primary source) and on what a call implies about the title's state. The
hypothesis about the API reporting the *previous* session's termination is offered as hypothesis and
labelled as such.

**One number corrected:** the body claimed the second mutation produced "2 checks FAIL"; measured, it
is 1. Fixed. A checkable number stated wrongly is worse than none, because it is the kind a reader
takes on trust.

Verified: 315/315 on the merge result, exit 0. CI 9 pass / 2 skipping. No `Claude-Session:` trailer,
no absolute host paths.

PR #3132, opened 2026-08-30
recompiler: give fragment diagnostics the real guest program address (#3130)

mattias800:
Fixes #3130. Found while investigating Stray's title screen (#3126), where it was the thing blocking
the diagnosis rather than a side issue.

## Failure scenario

`recompile_fragment_impl` constructed its `RecompileDiagnosticContext` with a hardcoded zero program
address (`rdna2_to_spirv.cpp:2933`). Concretely, on one routed Stray boot:

```
[mimg-unresolved] program=0x0 pc=10 srsrc=s0  ... key_res=null written=0
[mimg-unresolved] program=0x0 pc=20 srsrc=s0  ...
[mimg-unresolved] program=0x0 pc=32 srsrc=s8  ...
[mimg-unresolved] program=0x0 pc=69 srsrc=s8  ...
[mimg-unresolved] program=0x0 pc=89 srsrc=s16 ...
```

Every draw using those shaders is discarded, and none of the five could be attributed to a shader. I
matched them to three stages by comparing SRSRC sets against a separate per-stage dump, drew a
mechanism from that match, and had to retract it — the `Fragment` stage tag means the compute stages
I matched them to were the wrong ones. After this change the same route reports:

```
[mimg-unresolved] program=0x300c150000 pc=20 op=0x20 storage=0 srsrc=s0  ...
[mimg-unresolved] program=0x300c1d0000 pc=10 op=0x20 storage=0 srsrc=s0  ...
[mimg-unresolved] program=0x3013c60000 pc=69 op=0x20 storage=0 srsrc=s8  ...
[mimg-unresolved] program=0x3013c60000 pc=89 op=0x20 storage=0 srsrc=s16 ...
```

which joins directly against `PROSPER_SHARPLOG`'s per-`ud` drop lines and named the real causes in one
run (two `claimed by the V# path`, two `degenerate T#`). One line still reports `program=0x0` — a
fragment compile that does not go through the cached graphics entry point; left as is, since it is now
distinguishable rather than universal.

Two further consequences were silent, and both affect every title:

* `record_terminal_reject_reason()` early-returns on a zero address, so **no fragment rejection was
  ever recorded** and `last_terminal_reject_reason()` was empty for every fragment shader.
* `recompile_diagnostic_verbose()` compares against the program selector, so a fragment shader could
  **never** be selected for verbose diagnostics.

And the dedupe key degrades: `[mimg-unresolved]` keys on `(program, pc, srsrc)`, with a comment saying
[...]

PR #3139, opened 2026-08-30
recompiler: v_ldexp_f32 admits ABS/NEG on src0, its FLOAT operand (#3138)

### LINKED ISSUE #3138: recompiler: v_ldexp_f32 rejects ABS/NEG on src0, the FLOAT operand — one instruction costs 1536 full-screen draws
`v_ldexp_f32` rejects **any** VOP3 source modifier, including ABS/NEG on **src0** — which is the
ordinary float operand, where those modifiers are exactly the ones every other float op in the
emitter already applies.

## The gate

`prosper/src/gpu/recompiler/rdna2_emit_alu.cpp`, the `0x362` arm:

```c
if (in.src_abs[0] || in.src_abs[1] || in.src_abs[2] ||
    in.src_neg[0] || in.src_neg[1] || in.src_neg[2] ||
    in.clamp || in.omod) {
    ok = false;
}
```

Its comment justifies the rejection in terms of **the integer exponent**: "ABS/NEG on the integer
exponent and output-modifier denormal behavior need separate contracts, and silently ignoring either
would corrupt mip/scale reconstruction." That reasoning is right for `src1` and does not apply to
`src0`. `D.f = S0.f * 2**S1.i` — `src0` is a float, and `fv(0)` (abs then neg, hardware order) is the
established lowering for a float source throughout this file.

## What it costs

*Stray* (`PPSA02101`), title screen. The fragment shader `0x3011560000` is 3684 dwords with **exactly
one** unsupported instruction:

```
[recompile-reject] sh=bfa00003/3684 mode=unresolved-operand pc=383
                   words=d7620127,00018523 fmt=9 op=0x362
                   dst=39(kind2) src=35(k2),-2(k3),0(k0) stage=fragment role=terminal
```

`d7620127` has `ABS[2:0] = 001` — abs on src0 only, no neg, no clamp, no omod. So the instruction is
`v_ldexp_f32 v39, |v35|, -2`, and the *only* thing prosper cannot express about it is an absolute
value on a float.

That one instruction fails the whole stage, and the stage is the biggest single loss on this title's
title screen — **1536 full-screen 3840×2160 draws discarded per routed boot**, measured with a
per-shader drop census:

```
1536  es=0x30106d0000 ps=0x3011560000 fs_empty=1 target=0x3097110000 3840x2160
```

## Suggested shape

Keep rejecting modifiers on `src1`/`src2`, `clamp` and `omod` for exactly the reasons the existing
comment gives; admit them on `src0` by routing it through `fv(0)`:

```c
if (in.src_abs[1] || in.src_abs[2] || in.src_neg[1] || in.src_neg[2] ||
    in.clamp || in.omod) {
    ok = false;
} else {
    vreg[in.dst.value] = b.ldexp_f32_bits(fv(0), val(in.src[1]));
}
```

Measured with that change, on the title screen, dropped-draw census read at the same cumulative total
in every arm (two arms per build):

| build | `shader-recompile` drops @ 4096 discarded |
| --- | --- |
| master | 1026, 989 |
| with #3133 | 866, 886 |
| with #3133 + this | **611, 618** |

Offline, `gpu_replay --retry-failed-stage` on a captured title-screen frame turns
`rejected spirv-dwords=0` into `recompiled spirv-dwords=84990 contract=accepted`.

mattias800:
Fixes #3138. Found by dissecting *Stray*'s title-screen frame offline (#3126).

**This changes a deliberately pinned contract, so it wants a close read.** `test_rdna2_spirv_struct`
enumerated all eight VOP3 modifier mutations on `v_ldexp_f32` and asserted every one rejects. I am
arguing that guard was one operand too broad, not that it was wrong.

## The argument

`D.f = S0.f * 2**S1.i`. **src0 is a float.** ABS/NEG on a float source are the ordinary VOP3 modifiers
this emitter already applies everywhere else through `fv()` (hardware order: ABS then NEG). The gate's
own comment justifies the rejection in terms of the *integer exponent*:

> ABS/NEG on the integer exponent and output-modifier denormal behavior need separate contracts, and
> silently ignoring either would corrupt mip/scale reconstruction.

That is exactly right for `src1`, where "absolute value" and "negate" are not float modifiers at all,
and it does not reach `src0`. `src1`, `src2`, `CLAMP` and `OMOD` keep rejecting.

## What the breadth cost

*Stray*'s `0x3011560000` is a **3684-dword** fragment stage with **exactly one** unsupported
instruction:

```
[recompile-reject] sh=bfa00003/3684 mode=unresolved-operand pc=383
                   words=d7620127,00018523 fmt=9 op=0x362
                   dst=39(kind2) src=35(k2),-2(k3),0(k0) stage=fragment role=terminal
```

`d7620127` has `ABS[2:0] = 001` — abs on src0 only, no neg, no clamp, no omod. So it is
`v_ldexp_f32 v39, |v35|, -2`, and the only thing prosper could not express about it was an absolute
value on a float.

(One correction to the reject line above: its `modifier=0` field prints `in.has_modifier`, the
SDWA/DPP flag, **not** OMOD. OMOD is independently zero here — from word0 bits [60:59] — but do not
read that field as OMOD.)

That one instruction failed the whole stage, and the stage is the **largest single loss on that
title's title screen** — a per-shader drop census puts it at **1536 full-screen 3840×2160 draws
discarded per routed boot**:

```
1536  es=0x30106d0000 ps=0x3011560000 fs_empty=1 target=0x3097110000 3840x2160
```

Offline, `gpu_replay --retry-failed-stage` on the captured frame turns `rejected spirv-dwords=0` into
`recompiled spirv-dwords=84990 contract=accepted`.

## Tests

**The contract test is split by operand domain rather than weakened.** `abs0`/`neg0` must now compile;
`abs1`, `abs2`, `neg1`, `neg2`, `clamp`, `omod` must still reject, with the same exact bits as before.

**A render arm pins the semantics, not just the admission.** `test_recompiled_fragment` runs
`v_ldexp_f32 v1, |v4|, -2` with `v4 = -1.0` and carries the result in the **green** channel. The sign
is the point: without ABS the result is `-0.25`, which clamps to zero and reads as black. So the two
assertions separate all three outcomes that matter — the stage does not compile, it compiles and
silently drops the modifier, or it compiles and applies it. The centre pixel reads back **`0x40`**,
which is exactly 0.25.

**Mutation-verified**: restoring the old gate reddens **both** arms. That was not true when this body
was first written — the value arm sat inside `if (!ld_frag.empty())` and would have *skipped* on
exactly that mutation. It is unnested now, and the claim is checked rather than assumed.

The value assertion is a **band**, not a floor: `> 0x20` also accepted `0xff`, which is what an
ABS-applied-but-exponent-dropped lowering produces (`|-1.0| * 2^0 = 1.0`) — the competing wrong answer
this arm most needs to exclude, and this is the only numeric check of `ldexp_f32_bits` in the tree.

## Effect on the title screen

Dropped-draw census on Stray's title screen, read at the same cumulative total in every arm, two arms
per build:

| build | `shader-recompile` drops @ 4096 discarded |
| --- | --- |
| master | 1026, 989 |
| with #3133 | 866, 886 |
| with #3133 + this | **611, 618** |

Groups do not overlap. **The visible frame does not change** — the title screen is still menu-only at
`max_nonblack` 0.0069, because the remaining ~615 drops still discard the background. This is a
draw-loss reduction, not a rendering claim.

## Deliberately unaffected

`v_ldexp_f32` with any modifier on the integer exponent; CLAMP; OMOD; every other opcode's modifier
gate.

## Verification (exact head)

```
cmake --build prosper/build-linux -j6      # clean
ctest --no-tests=error -j4                 # 100% tests passed out of 314, rc=0
```

`docs/STRAY_STATUS.md` gains the `## Ruled out` row this PR's own result requires: `0x3011560000` was
the largest single draw loss on the route, and fixing it left `max_nonblack` unchanged at 0.0069 — so
draw-loss rank does not predict which draws carry the picture. **#3132 is concurrently editing that
file**; whichever merges second reconciles.

Snapshot guards not yet run on this branch.

Refs #2883, #3126

### DIFF
--- a/prosper/docs/STRAY_STATUS.md
+++ b/prosper/docs/STRAY_STATUS.md
@@ -55,6 +55,12 @@ completely empty table. Whether that is what its image ops want is untested — 
 
 ## Ruled out
 
+- **"The biggest dropped stage is the title-screen background."** Falsified: `0x3011560000` was the
+  largest single loss on this route — one instruction discarding **1536 full-screen 3840×2160 draws
+  per boot** — and fixing it (#3138) left `max_nonblack` unchanged at **0.0069**, a menu-only title
+  screen. Draw-loss rank does not predict which draws carry the picture, so the next candidate should
+  not be chosen by that rank alone. #3138.
+
 One line per falsified hypothesis, the evidence, and the link. Do not restart these without
 contradictory new evidence.
 
--- a/prosper/src/gpu/recompiler/rdna2_emit_alu.cpp
+++ b/prosper/src/gpu/recompiler/rdna2_emit_alu.cpp
@@ -4574,13 +4574,24 @@ bool emit_alu(SpirvCompute& b, RegState& rs, const Rdna2Inst& in, bool& ok, bool
                 // behavior need separate contracts, and silently ignoring either would corrupt
                 // mip/scale reconstruction. ldexp_f32_bits covers the full unmodified i32 exponent
                 // domain without relying on GLSL.std.450 Ldexp's undefined overflow cases.
-                if (in.src_abs[0] || in.src_abs[1] || in.src_abs[2] ||
-                    in.src_neg[0] || in.src_neg[1] || in.src_neg[2] ||
+                // src0 is the FLOAT operand, and ABS/NEG on a float source is the ordinary VOP3
+                // modifier every other float op here already applies through fv() -- hardware order
+                // ABS then NEG. Admitting it is not a new contract (#3138). The original gate
+                // rejected all three sources together, which also refused this well-defined case:
+                // Stray's `0x3011560000` is `v_ldexp_f32 v39, |v35|, -2` (`d7620127,00018523`,
+                // ABS[2:0]=001), and that ONE instruction failed a 3684-dword fragment stage and
+                // discarded 1536 full-screen 3840x2160 draws per routed boot.
+                //
+                // src1 keeps rejecting, and for a reason the gate's own comment gives: it is the
+                // integer EXPONENT, where "absolute value" and "negate" are not float modifiers at
+                // all and silently ignoring either would corrupt mip/scale reconstruction. CLAMP and
+                // OMOD keep rejecting too -- their denormal behaviour needs its own contract.
+                if (in.src_abs[1] || in.src_abs[2] ||
+                    in.src_neg[1] || in.src_neg[2] ||
                     in.clamp || in.omod) {
                     ok = false;
                 } else {
-                    vreg[in.dst.value] = b.ldexp_f32_bits(
-                        val(in.src[0]), val(in.src[1]));
+                    vreg[in.dst.value] = b.ldexp_f32_bits(fv(0), val(in.src[1]));
                 }
             } else if (in.opcode >= 0x144 && in.opcode <= 0x147) {
                 // Cubemap coordinate ops (#273 — DOLL's title post PSes' reflection-probe math):
--- a/prosper/tests/gpu/recompiler/test_recompiled_fragment.cpp
+++ b/prosper/tests/gpu/recompiler/test_recompiled_fragment.cpp
@@ -1142,17 +1142,23 @@ int main() {
         const std::vector<uint32_t> ld_frag = recompile_fragment(ldexp_abs, std::size(ldexp_abs));
         CHECK(!ld_frag.empty(),
               "#3138: v_ldexp_f32 with ABS on its float source recompiles");
-        if (!ld_frag.empty()) {
-            std::vector<uint32_t> ld_vert(kTriVertSpv, kTriVertSpv + std::size(kTriVertSpv));
-            const std::vector<uint8_t> ld_px =
-                prosper::test::render_triangle_rgba(ld_vert, ld_frag, W, H);
-            const uint8_t* c = ld_px.size() == (size_t)W * H * 4
-                ? &ld_px[((size_t)(H / 2) * W + W / 2) * 4] : nullptr;
-            printf("  #3138 centre pixel = %02x %02x %02x\n",
-                   c ? c[0] : 0, c ? c[1] : 0, c ? c[2] : 0);
-            CHECK(c && c[1] > 0x20 && c[0] < 0x20 && c[2] < 0x20,
-                  "#3138: ABS is APPLIED, not dropped (green carries +0.25, not a clamped -0.25)");
-        }
+        // Deliberately NOT nested under the check above: a reverted gate must redden BOTH arms, and
+        // it does only if the value arm still runs and finds no pixel. Nesting it would make the
+        // second arm silently SKIP on exactly the mutation it exists to catch.
+        std::vector<uint32_t> ld_vert(kTriVertSpv, kTriVertSpv + std::size(kTriVertSpv));
+        const std::vector<uint8_t> ld_px =
+            ld_frag.empty() ? std::vector<uint8_t>()
+                            : prosper::test::render_triangle_rgba(ld_vert, ld_frag, W, H);
+        const uint8_t* c = ld_px.size() == (size_t)W * H * 4
+            ? &ld_px[((size_t)(H / 2) * W + W / 2) * 4] : nullptr;
+        printf("  #3138 centre pixel = %02x %02x %02x (want 00 40 00)\n",
+               c ? c[0] : 0, c ? c[1] : 0, c ? c[2] : 0);
+        // A BAND, not a floor. `> 0x20` would also accept 0xff, which is what an
+        // ABS-applied-but-exponent-dropped lowering produces (|-1.0| * 2^0 = 1.0) -- the competing
+        // wrong answer this arm most needs to exclude. |-1.0| * 2^-2 = 0.25 = 0x40 exactly, and this
+        // is the only numeric check of ldexp_f32_bits in the tree.
+        CHECK(c && c[1] > 0x30 && c[1] < 0x50 && c[0] < 0x20 && c[2] < 0x20,
+              "#3138: ABS applied AND the exponent honoured (green is 0.25, not 0 and not 1.0)");
     }
 
     if (fails) { printf("== FAIL: %d ==\n", fails); return 1; }
--- a/prosper/tests/gpu/recompiler/test_recompiled_fragment.cpp
+++ b/prosper/tests/gpu/recompiler/test_recompiled_fragment.cpp
@@ -1116,6 +1116,45 @@ int main() {
               "#635: MRT3-only export lands on slot 3 and does NOT remap onto color0");
     }
 
+    {
+        // #3138: v_ldexp_f32 with ABS on src0. src0 is the FLOAT operand, where ABS/NEG are the
+        // ordinary VOP3 modifiers every other float op here applies; only the integer EXPONENT
+        // (src1) has no meaning for them. The old gate refused all three sources together, so
+        // Stray's `0x3011560000` -- 3684 dwords with exactly ONE unsupported instruction,
+        // `v_ldexp_f32 v39, |v35|, -2` -- failed entirely and discarded 1536 full-screen draws.
+        //
+        //   v_mov_b32 v4, -1.0 | v_mov_b32 v0, 0 | v_ldexp_f32 v1, |v4|, -2
+        //   v_mov_b32 v2, 0    | v_mov_b32 v3, 1.0 | exp mrt0 v0..v3 | s_endpgm
+        //
+        // GREEN carries |-1.0| * 2^-2 = +0.25, and the sign is the point: without ABS the result is
+        // -0.25, which clamps to zero and reads as black. So the two assertions separate the three
+        // outcomes that matter -- the stage does not compile at all (the old behaviour), it compiles
+        // but silently drops the modifier (green 0), or it compiles and applies it (green > 0).
+        const uint32_t ldexp_abs[] = {
+            0x7E0802F3u,                       // v_mov_b32 v4, -1.0
+            0x7E000280u,                       // v_mov_b32 v0, 0        (red)
+            0xD7620101u, 0x00018504u,          // v_ldexp_f32 v1, |v4|, -2   (green)
+            0x7E040280u,                       // v_mov_b32 v2, 0        (blue)
+            0x7E0602F2u,                       // v_mov_b32 v3, 1.0      (alpha)
+            0xF800180Fu, 0x03020100u,          // exp mrt0, v0..v3
+            0xBF810000u,                       // s_endpgm
+        };
+        const std::vector<uint32_t> ld_frag = recompile_fragment(ldexp_abs, std::size(ldexp_abs));
+        CHECK(!ld_frag.empty(),
+              "#3138: v_ldexp_f32 with ABS on its float source recompiles");
+        if (!ld_frag.empty()) {
+            std::vector<uint32_t> ld_vert(kTriVertSpv, kTriVertSpv + std::size(kTriVertSpv));
+            const std::vector<uint8_t> ld_px =
+                prosper::test::render_triangle_rgba(ld_vert, ld_frag, W, H);
+            const uint8_t* c = ld_px.size() == (size_t)W * H * 4
+                ? &ld_px[((size_t)(H / 2) * W + W / 2) * 4] : nullptr;
+            printf("  #3138 centre pixel = %02x %02x %02x\n",
+                   c ? c[0] : 0, c ? c[1] : 0, c ? c[2] : 0);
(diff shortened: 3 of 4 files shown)


mattias800 (commented):
**Verdict: APPROVED.** The argument holds, and I checked it against document 70648 itself rather than against the reasoning. Findings below are non-blocking except the one doc item the charter requires before merge.

Reviewed at head `753a6b5c`, read-only.

---

## The central claim: verified from the primary source

I extracted the text of the RDNA 2 ISA reference (document 70648) and read the relevant sections rather than reasoning from plausibility. Three statements settle it:

- **§6.2, Operands (p.43):** *"Instructions using the VOP3 form and also using floating-point inputs have the option of applying absolute value (ABS field) or negate (NEG field) to any of the input operands."*
- **VOP3A field description:** *"ABS = Floating-point absolute value"*, *"NEG = Floating-point negation"*.
- **Opcode 866, V_LDEXP_F32:** `D.f = S0.f * (2 ** S1.i)`.

So S0 is a floating-point input and S1 is not, and the ISA's own operand typing is precisely the split this PR makes. The gate was one operand too broad, exactly as argued.

Two supporting details from the same document, both of which cut the author's way:

- The doc **is explicit where a modifier is not allowed**: V_PERMLANE16_B32 and V_PERMLANEX16_B32 each carry *"ABS, NEG and OMOD modifiers should all be zeroed for this instruction."* V_LDEXP_F32 carries no such statement.
- **OMOD staying out is better justified than the comment claims.** §6.2: *"Output modifiers are not compatible with output denormals: if output denormals are enabled, then output modifiers are ignored. If output denormals are disabled, then the output modifier is applied and denormals are flushed to zero. Output modifiers are not IEEE compatible: -0 is flushed to +0. Output modifiers are ignored if the IEEE mode bit is set to 1."* "Their denormal behaviour needs its own contract" is primary-source-backed. (Aside, pre-existing and not this PR's problem: `fresult` at `rdna2_emit_alu.cpp:4220` models none of that mode-dependency for the opcodes that *do* route through it.)

## The decode: confirmed byte-for-byte, twice

**Field layout**, doc 70648 §13.3 VOP3A table: `ABS [10:8]` — *"[8] = src0, [9] = src1, [10] = src2"*; `CLMP [15]`; `OP [25:16]`; `ENCODING [31:26] = 110101`; `OMOD [60:59]`; `NEG [63:61]` — *"[61] = src0"*. `rdna2_decode.cpp:639` reads exactly that.

`0xd7620127` → OP = 0x362, CLMP = 0, ABS = `0b001` → src0 only. `0x00018523` → SRC0 = 291 = v35, SRC1 = 194 = inline `-2`, SRC2 = 0, OMOD = 0, NEG = 0. The author's reading is right.

**Independent round-trip** (`llvm-mc -arch=amdgcn -mcpu=gfx1030`, in the project distrobox):

- `v_ldexp_f32 v39, |v35|, -2` assembles to `[0x27,0x01,0x62,0xd7,0x23,0x85,0x01,0x00]` — byte-identical to the captured words — and those bytes disassemble back to the same text.
- The same assembler **refuses** `|v1|` and `-v1` on src1 (`error: not a valid operand`), and **silently drops** a set ABS[9] when disassembling `d7620200,0002030d` → `v_ldexp_f32 v0, v13, v1`. src1 has no modifier slot for this opcode in the ISA's own operand model.
- All eight struct-test patterns and the new fragment pattern decode to what their comments claim; I hand-decoded each and disassembled four.

**One correction, because it would have travelled.** If the reject line's `modifier=0` was read as OMOD anywhere, that reading is wrong: `rdna2_emit_cfg.cpp:5993` prints `modifier=%d` and `:6002` passes `(int)in.has_modifier`, which `rdna2_decode.hpp:323` defines as *"VOP SDWA/DPP form (2nd dword is a control word, not a literal)"*. It carries no OMOD information at all. The conclusion is unaffected — OMOD genuinely is 0, established independently from word1 bits [28:27] and from the round-trip — but the route to it was not the one stated, and per the charter's *"a cited claim reads as a verified one"* section I would rather kill that reading now than see it cited later.

## `fv(0)` is the right lowering — with one honest gap

`rdna2_emit_alu.cpp:4207` applies `Glsl_FAbs` then `fneg`, i.e. `-|x|`. Type-wise it is correct: `fext1` (`rdna2_to_spirv_internal.hpp:725`) and `fneg` (`:734`) both return raw bits, which is what `ldexp_f32_bits` (`:651`) takes.

**Document 70648 does not state the ABS/NEG application order anywhere I could find** — I searched the whole extracted text. What establishes it is LLVM's gfx1030 syntax, where setting both bits round-trips as `-|v35|` (negation outside the bars), plus the ISA's own `-abs(S0.f)` pseudo-code idiom in V_DIV_FIXUP. That is tier-3 under the charter's evidence hierarchy and I am flagging it as such rather than restating it as documented fact. It does not gate this PR: `fv` is the shared lowering for every float source in the file, so if the order were wrong the blast radius is hundreds of opcodes rather than this one — and neither admitted case here sets both bits.

Consistency note, not a finding: `Glsl_FAbs` is a real float op with bitcasts around it, inside a helper whose own comment says it deliberately stays in the integer domain *"so zero signs, infinities and NaN payloads survive exactly"*. A bitwise `AND 0x7fffffff` would be hardware-exact and preserve that invariant (`fneg` already is a bitwise XOR, so only the ABS half breaks it). I looked for a case where it bites and did not find one — under FTZ both hardware and this lowering flush, under DenormPreserve neither does.

## Nothing else becomes reachable

- The 0x362 arm sits inside `case Rdna2Format::VOP3:` (`rdna2_emit_alu.cpp:3897`) only. VOPC is reclassified out at decode (`rdna2_decode.cpp:650`) and VOP3P populates `src_abs` from NEG_HI in its own case, so no other format reaches this arm with differently-meaning bits.
- **`clamp_routed` is untouched.** The arm still rejects `in.clamp` outright (`:4589`), so the tail guard `if (ok && in.clamp && !clamp_routed) ok = false;` (`:4981`) is never the thing that decides for this opcode, before or after. No new silent-saturation hole.
- `fv(0)` is the only consumer of the newly admitted bits; `val()` handles SGPR/inline/literal src0 identically underneath.
- 0x362 is already in `emitted_vop3_opcode_has_two_data_sources` (`rdna2_decode.hpp:197`), so src2 is not exposed as a phantom dependency, and abs2/neg2 keep rejecting — consistent.
- No other gate on 0x362 exists anywhere in the recompiler.

---

## Findings

**F1 — non-blocking, but please correct the PR body. "Mutation-verified: restoring the old gate reddens both arms" cannot be what happened.**
The second CHECK is inside `if (!ld_frag.empty())` (`tests/gpu/recompiler/test_recompiled_fragment.cpp:1145`, CHECK at `:1153`). With the old gate the emitter sets `ok = false`, `recompile_fragment` returns `{}`, arm 1 reddens — and arm 2 **is skipped**: it neither passes nor fails, and it does not print. The two arms answer two *different* mutations, and each is genuinely good for its own: revert-the-gate → arm 1; `fv(0)` → `val(in.src[0])` → arm 2 (green clamps to 0). Please restate it that way. The discriminators are fine; only the claim about them is wrong.

**F2 — non-blocking. `c[1] > 0x20` is looser than the semantics it advertises.**
`test_recompiled_fragment.cpp:1153`. The PR body says the pixel reads exactly `0x40`, and it should: 0.25 into `VK_FORMAT_R8G8B8A8_UNORM` (`tests/fixtures/render_runner.h:354`). But `> 0x20` also accepts `0xFF`, which is what a lowering that applied ABS and silently dropped the **exponent** would produce. That matters more than usual here because this arm is the *only* numeric check of `ldexp_f32_bits` in the tree — `ldexp` appears in exactly two test files, and the pre-existing GTA V arm (`tests/gpu/test_rdna2_spirv_struct.cpp:3123`) asserts only which GLSL ext-insts are present, never a value. A narrow band around `0x40` costs one edit and pins the modifier *and* the exponent.

**F3 — non-blocking. The retained src1/src2 rejection is right; the comment's reason is not.**
`rdna2_emit_alu.cpp:4585-4588`, mirrored in the test comment, says ignoring ABS/NEG on src1 *"would corrupt mip/scale reconstruction"*. Per §6.2 those bits are floating-point modifiers applicable to floating-point inputs, and S1 is `S1.i` — they have no defined effect there at all, and LLVM's disassembler drops a set ABS[9] on this opcode outright. Ignoring them would most likely match hardware, not corrupt anything. Keep rejecting — we have no positive evidence for what silicon does with a bit the ISA leaves undefined for that operand, and fail-visible is the house rule — but the honest justification is *"undefined for this operand, no evidence, so reject visibly"*. The PR is rewriting these comments anyway, and an inherited-then-restated rationale is exactly what the charter's stale-figure paragraph warns about.

**F4 — nits.**
- `tests/gpu/test_rdna2_spirv_struct.cpp:3165` still prints `"…; modifier mutations reject"` when two of the eight now compile.
- The comments write `ABS[2:0]=001`; doc 70648 §13.3 names the field `ABS [10:8]`. Field-value notation is defensible, but `ABS[10:8]=001` reads straight against the ISA table.

**F5 — nit, coverage.** `abs0` and `neg0` are tested separately; the combined `-|x|` form (`{0xd7620100u, 0x2002030du, …}`) is not. Since the ISA does not document the order, that is the one pattern where `fv`'s choice is observable. One more array.

---

## Required before merge (charter rule, not a code defect)

**The falsification this PR produced is not landing in the project record.** Fixing the single largest dropped shader on Stray's title screen — 1536 full-screen draws — left `max_nonblack` at 0.0069, unchanged. *"The biggest dropped stage is the background"* is a hypothesis any reader of #3126 forms for free, and this PR kills it at real cost. `docs/STRAY_STATUS.md` § Ruled out (`:56`) is where the charter says that goes, **before merging** — one line, with the issue link. Heads up that #3132 is concurrently editing that same file, so coordinate or rebase.

Also: CI at `753a6b5c` was still in flight when I read it (Linux, Windows MinGW, macOS Rosetta 2, Windows App all `in_progress`). Linux is the job that runs both tests this PR changes.

**Not verified by me, and I did not try:** the 3684-dword span, the 1536-draw census, the 84990-dword offline `--retry-failed-stage` result, and the drop-census A/B table. Those are the author's measurements. The encoding they all rest on I did confirm, byte for byte, against both the ISA document and an independent assembler.

**APPROVED.**

mattias800:
All four review items addressed at head; thank you for tracing the ISA rather than my argument — §6.2's float typing and the `V_PERMLANE16_B32` counter-example are exactly the primary evidence this needed, and the `llvm-mc` round-trip against Stray's captured bytes is stronger than anything I had.

Three of the four were my claims being wrong, so they are corrected rather than argued:

* **F1** — the value arm was nested under `if (!ld_frag.empty())`, so a gate revert reddened arm 1 and *skipped* arm 2. Unnested; a revert now reddens both, verified rather than asserted.
* **F2** — `> 0x20` accepted `0xff`, the ABS-applied-but-exponent-dropped answer. Now a band, `0x30 < green < 0x50` around the exact `0x40`.
* **`modifier=0`** — you are right that it prints `in.has_modifier` (SDWA/DPP), not OMOD. The PR body said OMOD; corrected, with the note that OMOD is independently zero here.
* **Charter rule** — the `## Ruled out` row is in: `0x3011560000` was the largest single draw loss on the route and fixing it left `max_nonblack` unchanged at 0.0069, so draw-loss rank does not predict which draws carry the picture. Flagged the concurrent edit with #3132.

On ABS-then-NEG order: agreed it is not in doc 70648, and I have not restated it as documented anywhere — it is not load-bearing since neither admitted case sets both bits.

`ctest --no-tests=error`: 314/314 at head.

mattias800:
Merging. Recording the review-freshness check explicitly, since the approval is **not** on the current head.

* approval: `753a6b5c` (2026-08-30T08:57:25Z)
* head: `b6180ecb`

Established by **content** rather than assumed: `git diff 753a6b5c b6180ecb -- prosper/src/` is **empty** — the delta is tests and docs only, and it is exactly the three corrections the review asked for (unnest the value arm so a gate revert reddens both; band the assertion so `0xff` is excluded; add the charter-required `## Ruled out` row). Nothing in the approved source changed.

**Gates at head:**

| gate | result |
| --- | --- |
| CI | 0 pending, 0 failures |
| `ctest --no-tests=error -j4` | 100% of 314, rc=0 |
| snapshot guards | **16/16 anchors OK** (alexkidd 4/4 0.982–1.000, balan 1/1 1.000, evergate 3/3 0.973–0.997, greak 3/3 0.956–1.000, joe-mac 5/5 0.979–1.000) |
| mutation | restoring the old gate reddens **both** arms |

Thank you for going to doc 70648 and `llvm-mc` rather than taking my reasoning — the operand-domain split rests on the ISA's own typing because of that, and three of your four items were my claims being wrong.

Issue #3140, opened 2026-08-30
[FALSIFIED] compute-written surfaces are invisible to the sampler — measured 58,569 sample HITs to 5 skips


mattias800:
**16 surfaces that prosper renders into are sampled by later shaders and read BLACK**, because the
live render-target cache does not consider them authoritative and their guest memory is zero. This is
the mechanism behind "the scene targets are black at source" measured earlier in #3126.

## Measurement

One routed boot of *Stray*'s title screen (`PROSPER_NULL_PAGE=1`, settled at `max_nonblack` 0.0069),
three logs cross-referenced: `PROSPER_TEXCONTENT` (guest memory behind every sampled T# ≥1920×1080,
plus `is_live_render_target()` for the same address) and `PROSPER_DRAWMAP` (every colour target a
realized draw renders into).

A sample can be served from **two** sources — the RTT cache for an address prosper authored, or guest
memory otherwise — so both have to be checked. Of 66 large sampled surfaces:

| guest memory | RTT authority | count |
| --- | --- | --- |
| all-zero | **MISS** | **36** |
| all-zero | authoritative | 13 |
| has content | MISS | 16 |
| has content | authoritative | 1 |

The 36 with **both sources empty** read black. Splitting them by whether a draw renders into the
address:

* **16 — prosper DOES render into them**, and the sample still misses
* 20 — never rendered into (a separate question, also open)

Those 16 are the defect here. A surface the renderer authored should be authoritative for a later
sample; these are not:

```
0x3041d60000 3840x2160 fmt=1     0x304f720000 3840x2160 fmt=71
0x3058750000 3840x2160 fmt=71    0x3060750000 3840x2160 fmt=71
0x3064760000 3840x2160 fmt=71    0x306e780000 3840x2160 fmt=36
0x30707d0000 1920x1080 fmt=36    0x3072150000 1920x1080 fmt=36
0x3073160000 1920x1080 fmt=36    0x307c1a0000 1920x1080 fmt=36
0x30a0b60000 3840x2160 fmt=11    0x30a2b60000 3840x2160 fmt=56
0x30b7e80000 3840x2160 fmt=36    0x30ba820000 1920x1080 fmt=36
0x30c61d0000 3840x2160 fmt=71    0x304e720000 2048x2048 fmt=13
```

## Where to look

`live_renderer.cpp`'s `set_live_target_query` returns
[...]

Issue #2287, opened 2026-08-07
Second per-draw getenv pair, in realize_draw_item, runs on parallel workers where the cost may be contention


mattias800:
#2285 removes the two per-draw `getenv` calls that `build_bds` made through
`validate_runtime_descriptor_contract`. There is a **second** pair on the realization path, and it
is likely worse for a reason that does not apply to the first.

## Where

`gpu_execute.hpp:1357-1358`, inside `realize_draw_item`:

```cpp
if (!validate_runtime_descriptor_contract("VS", vs_words, vrt.get(), 0, SpirvShaderStage::Vertex) ||
    !validate_runtime_descriptor_contract("PS", fs_words, prt.get(), 1, SpirvShaderStage::Fragment)) {
```

`realize_draw_item` is called once per draw from `parallel_draw_worker_execute`
(`gpu_executor.cpp:6090`):

```cpp
void parallel_draw_worker_execute(void* opaque, size_t draw_index, size_t) {
    ...
    slot.realized = realize_draw_item(state.state_at_draw(draw_index), &draw, ...);
```

## Why this one is not just "the same again"

The `build_bds` pair runs on one thread, so its cost is a constant: 2 × 1.26 µs per draw on this
host (measured in #2285; #2239 independently measured 1.105 µs).

This pair runs on **parallel worker threads**, and the comment #2240 left in `live_renderer.cpp`
records the relevant property:

> on Windows getenv takes a process-wide lock and walks the environment block

A process-wide lock taken twice per draw by every worker is **contention**, not a per-call
constant, so the parallel path may lose more than the serial arithmetic predicts — and it would get
*worse* with more workers, which is the opposite of what the parallelisation is for.

**This is a hypothesis, not a measurement.** I have not measured it. It is filed rather than fixed
precisely because the serial arithmetic from #2285 must not be assumed to carry over.

## What would settle it

- A targeted timer around the pair inside `realize_draw_item`, reported per draw — not a route A/B.
  #2240's route A/B on PPSA25009 was void (its control channel moved more than the treated one) and
  #2239 records why a targeted timer is the right instrument here.
- Worker count as the independent variable. If the cost per draw **rises** with more workers, it is
  lock contention; if it stays flat at ~2.5 µs/draw, it is the same constant as the serial site and
  the fix is mechanical.

## The fix, if it is worth it

Not mechanical, and that is why this is separate. The mode has to reach `realize_draw_item`, which
[...]

PR #2214, opened 2026-08-06
perf(win): stop reading the environment per resource per draw in the live renderer

mattias800:
perf(win): stop reading the environment per resource per draw in the live renderer

getenv() is not free. The UCRT walks a locked environment block, strlen-ing and
comparing each entry, on EVERY call -- so a diagnostic that is switched OFF still
costs that scan every time its guard is evaluated. Several of live_renderer's
guards sit on per-draw and per-RESOURCE paths, so the cost scales with scene
complexity rather than with logging.

Measured on Blue Prince (native Windows), sampling the guest primary thread with
every diagnostic disabled: 97 of 331 samples -- ~30% of that thread's wall time
-- inside getenv's strlen loop. The sampled address was disassembled to confirm
the function rather than assumed:

    1800315c0: mov  (%rbx),%rcx          ; envp[i]
    1800315d0: inc  %rax                 ; strlen
    1800315d3: cmpb $0x0,(%rcx,%rax,1)
    1800315d7: jne  0x1800315d0          ; <- 51 of 331 samples on this instruction
    1800315e4: cmpb $0x3d,(%rcx,%rdi,1)  ; '=' at name length

Three changes, in increasing order of how much they mattered
------------------------------------------------------------
1. PROSPER_ENV_ON  -- boolean guards, read once. The obvious half.

2. PROSPER_ENV_VALUE -- the value form, `const char* m = getenv(...)`. This is
   the half that hides: it does not look like a flag check, so a boolean sweep
   walks straight past it, and it was the LARGER contributor.

3. The single hottest line is HOISTED rather than cached:

       if (const char* mode = getenv("PROSPER_RENDER_TIMING");
           mode && strcmp(mode, "detail") == 0)

   a per-resource read of a per-span condition, calling getenv up to three times
   per resource. It now reads once per span, next to the existing timing_mode.

Why hoisting and not caching, there: tests arm PROSPER_RENDER_TIMING at RUNTIME,
and caching samples at first use, so caching would silently make the variable
unobservable to them. Hoisting removes the cost and keeps it live. Where a
variable is not test-armed, caching is simpler and is what the macros do.

Which variables may be cached is DERIVED, not guessed: any name a test arms with
_putenv_s/setenv keeps its live read. 14 reads in this file are left live for
that reason.

Measured, like-for-like, matched FRAME counts, same route
----------------------------------------------------------
    baseline (master)                       0.7 fps
    boolean guards only                     0.8 fps   (+14%)
[...]

PR #3141, opened 2026-08-30
diag(gpu): four instruments for "which shader owns these pixels, and what does it read"

mattias800:
Built while chasing *Stray*'s black title screen (#3126). The existing diagnostics could say which
shaders **failed**; nothing could say which shader **owned the pixels that came out black**. All four
are off by default, additive, and each answers a question none of the others can.

## What each one is for

| variable | question it answers |
| --- | --- |
| `PROSPER_DRAWMAP=1` | which fragment program owns which target, extent and scissor — a static filter over a whole route in **one run** |
| `PROSPER_COMPUTEMAP=1` | which surfaces **compute** writes as storage images (DRAWMAP is graphics-only) |
| `PROSPER_TEXCONTENT=1` | for every large sampled texture: does guest memory hold anything, **and** does the renderer's cache call it authoritative |
| `PROSPER_MIMG_SOFT=1` | how much of the picture rides on one descriptor-resolution failure — writes a constant instead of failing the stage |

## Why they were worth building

**DRAWMAP** identified Stray's full-screen composite in one run. Painting one shader at a time answers
the same question one candidate per run, and that title has 66 fragment stages.

**COMPUTEMAP** exists because DRAWMAP is graphics-only, and a surface written by compute otherwise
looks like a surface nothing writes at all. This investigation took that wrong turn before adding it.

**TEXCONTENT** reports **both** sources a sample can come from. Reporting one alone produces false
alarms: an address the renderer authored is legitimately zero in guest memory, because its pixels live
in the cache — I nearly published exactly that as a defect.

**MIMG_SOFT** produces deliberately wrong output and is never acceptable in a run that yields
progression evidence. Its value is the counterfactual: on Stray it showed the composite compiles and
the frame *still* does not change, which is what redirected #3126 away from the recompiler.

## The dedupe rule, written into the code

`TEXCONTENT` dedupes on **(base, VERDICT)**, never on base alone, and its comment says why: an
address-keyed probe reports whatever the **first** observation saw, which on a long-running renderer
is the boot-time state — so a surface filled moments later still reads as empty. That exact mistake
produced **three published corrections** on #3140 before I caught it. It is called out in the comment
because the shape is easy to repeat and expensive every time.

## Beyond this title

*Grand Theft Auto V* and *Sonic Frontiers* both sit on "a HUD over an absent world" with no way to ask
which shader owns the pixels. DRAWMAP and COMPUTEMAP answer that directly, with no capture and no
per-candidate run.

## Verification

```
cmake --build prosper/build-linux -j6          # clean
ctest --no-tests=error -j4                     # 100% tests passed out of 314, rc=0
python3 tools/env/check_diag_gates.py          # == all checks passed ==
                                               # 108 findings, 108 baselined — no new gate pattern
git diff --check                               # clean
```

No behaviour changes with every variable unset: each is a `getenv` guard around an added block, and
`MIMG_SOFT`'s is placed on the existing `if (!res)` reject path so an unset run takes the original
branch unchanged.

## Not included

`PROSPER_PAINT_PS` (replace one named fragment program with a constant colour) needs the real guest
program address, which #3132 supplies; it belongs on top of that PR rather than this one.

Refs #3126, #3140

### DIFF
--- a/prosper/src/gpu/agc/agc_shader_layout.cpp
+++ b/prosper/src/gpu/agc/agc_shader_layout.cpp
@@ -972,12 +972,17 @@ ShaderResourceTable build_shader_resources(const AgcShaderHeader& shdr,
                     // whatever the FIRST observation saw, which on a long run is the boot-time state,
                     // so a surface filled moments later still reads as empty. That mistake cost three
                     // published corrections before it was caught.
-                    static std::mutex cmx; static std::set<std::pair<uint64_t, int>> cseen;
+                    // Keyed on (shader, base, verdict): the same surface is sampled by several
+                    // stages, and "which shader reads a surface that is empty" is the question this
+                    // answers -- a base-only key reports one sampler and hides the rest.
+                    static std::mutex cmx;
+                    static std::set<std::tuple<const void*, uint64_t, int>> cseen;
                     std::lock_guard<std::mutex> lk(cmx);
-                    if (cseen.insert({d.base, verdict}).second)
+                    if (cseen.insert({shdr.code, d.base, verdict}).second)
                         fprintf(stderr,
-                                "[texcontent] base=0x%llx %ux%u fmt=%u bytes=%llu nonzero=%u/%u %s "
-                                "rtt=%s\n",
+                                "[texcontent] shader=%p base=0x%llx %ux%u fmt=%u bytes=%llu "
+                                "nonzero=%u/%u %s rtt=%s\n",
+                                shdr.code,
                                 (unsigned long long)d.base, d.width, d.height, d.format,
                                 (unsigned long long)span, nonzero, sampled_dwords,
                                 sampled_dwords == 0 ? "UNREADABLE"
--- a/prosper/src/gpu/agc/agc_shader_layout.cpp
+++ b/prosper/src/gpu/agc/agc_shader_layout.cpp
@@ -963,18 +963,24 @@ ShaderResourceTable build_shader_resources(const AgcShaderHeader& shdr,
                     // is_live_render_target() is safe to call here only because PROSPER_TEXCONTENT
                     // is in parallel_draw_diagnostic_active()'s list, so draw realization is serial
                     // while this diagnostic is armed. That callback drains and mutates the renderer's
-                    // write queue; calling it from parallel workers is UB and perturbs the very cache
-                    // state this reports.
+                    // write queue, and calling it from parallel workers is UB.
+                    // Serialising fixes the UB and nothing else: the callback still runs once per
+                    // large texture slot per draw, so this diagnostic still shifts the timing of the
+                    // very write queue it reports on. Arming it changes when guest writes land, which
+                    // is why a content verdict from a PROSPER_TEXCONTENT run is evidence about that
+                    // run and not about a default one.
                     const bool cached = prosper::gpu::is_live_render_target(d.base);
                     const int verdict = ((sampled_dwords == 0) ? 0 : (nonzero == 0 ? 1 : 2)) |
                                         (cached ? 4 : 0);
-                    // Deduped on (base, VERDICT), never on base alone. An address-keyed probe reports
-                    // whatever the FIRST observation saw, which on a long run is the boot-time state,
-                    // so a surface filled moments later still reads as empty. That mistake cost three
-                    // published corrections before it was caught.
-                    // Keyed on (shader, base, verdict): the same surface is sampled by several
-                    // stages, and "which shader reads a surface that is empty" is the question this
-                    // answers -- a base-only key reports one sampler and hides the rest.
+                    // Deduped on (SHADER, base, VERDICT), never on base alone, and every part of
+                    // that key earns its place:
+                    //   - verdict, because an address-keyed probe reports whatever the FIRST
+                    //     observation saw, which on a long run is the boot-time state -- a surface
+                    //     filled moments later still reads as empty. That mistake cost three
+                    //     published corrections before it was caught.
+                    //   - shader, because the same surface is sampled by several stages, and "which
+                    //     shader reads a surface that is empty" is the question this answers. A key
+                    //     without it reports one sampler and hides the rest.
                     static std::mutex cmx;
                     static std::set<std::tuple<const void*, uint64_t, int>> cseen;
                     std::lock_guard<std::mutex> lk(cmx);
--- a/prosper/src/gpu/agc/agc_shader_layout.cpp
+++ b/prosper/src/gpu/agc/agc_shader_layout.cpp
@@ -1,5 +1,6 @@
 // agc_shader_layout.cpp — see agc_shader_layout.hpp. V# decode + the front-half resource-table build.
 #include "gpu/agc/agc_shader_layout.hpp"
+#include "gpu/execute/gpu_execute.hpp"
 #include "gpu/texture/tile.hpp"
 #include <algorithm>
 #include <climits>
@@ -918,6 +919,44 @@ ShaderResourceTable build_shader_resources(const AgcShaderHeader& shdr,
                     }
                     tseen[key] = done;
                 }
+            }
+            // PROSPER_TEXCONTENT=1 -- for every large sampled texture, report BOTH sources a
+            // sample can come from: whether its guest memory holds anything, and whether the
+            // renderer's own cache calls it authoritative. Reporting only one is how #3126 produced
+            // a false alarm -- an address the renderer authored is legitimately zero in guest
+            // memory, because its pixels live in the cache.
+            //
+            // Deduped on (base, VERDICT), never on base alone. An address-keyed probe reports
+            // whatever the FIRST decode saw, which on a long run is the boot-time state, so a
+            // surface filled moments later still reads as empty. That mistake cost three published
+            // corrections on #3140 before it was caught; with the verdict in the key a state change
+            // prints and "it was empty" cannot masquerade as "it is empty".
+            if (getenv("PROSPER_TEXCONTENT") && d.width >= 1920u && d.height >= 1080u &&
+                d.base > 0x1000000000ull) {
+                static std::mutex cmx; static std::set<std::pair<uint64_t, int>> cseen;
+                std::lock_guard<std::mutex> lk(cmx);
+                const uint64_t span = (uint64_t)d.width * d.height;
+                uint32_t nonzero = 0, sampled_dwords = 0;
+                for (uint32_t s = 0; s < 8; ++s) {
+                    const uint64_t addr = d.base + (((span / 8u) * s) & ~(uint64_t)3);
+                    if (!guest_readable(addr, 256)) continue;
+                    const uint32_t* words = (const uint32_t*)(uintptr_t)addr;
+                    for (uint32_t k = 0; k < 64; ++k) {
+                        ++sampled_dwords;
+                        if (words[k]) ++nonzero;
+                    }
+                }
+                const bool cached = prosper::gpu::is_live_render_target(d.base);
+                const int verdict = ((sampled_dwords == 0) ? 0 : (nonzero == 0 ? 1 : 2)) |
+                                    (cached ? 4 : 0);
+                if (cseen.insert({d.base, verdict}).second)
+                    fprintf(stderr,
+                            "[texcontent] base=0x%llx %ux%u fmt=%u nonzero=%u/%u %s rtt=%s\n",
+                            (unsigned long long)d.base, d.width, d.height, d.format,
+                            nonzero, sampled_dwords,
+                            sampled_dwords == 0 ? "UNREADABLE"
+                                                : (nonzero == 0 ? "ALL-ZERO" : "has-content"),
+                            cached ? "authoritative" : "MISS");
             }
             if (getenv("PROSPER_GFXLOG") || getenv("PROSPER_TEXLOG")) {
                 const uint32_t* t = tv;   // the fetched T# (SGPR block or EUD spill)
--- a/prosper/src/gpu/agc/agc_shader_layout.cpp
+++ b/prosper/src/gpu/agc/agc_shader_layout.cpp
@@ -933,30 +933,57 @@ ShaderResourceTable build_shader_resources(const AgcShaderHeader& shdr,
             // prints and "it was empty" cannot masquerade as "it is empty".
             if (getenv("PROSPER_TEXCONTENT") && d.width >= 1920u && d.height >= 1080u &&
                 d.base > 0x1000000000ull) {
-                static std::mutex cmx; static std::set<std::pair<uint64_t, int>> cseen;
-                std::lock_guard<std::mutex> lk(cmx);
-                const uint64_t span = (uint64_t)d.width * d.height;
-                uint32_t nonzero = 0, sampled_dwords = 0;
-                for (uint32_t s = 0; s < 8; ++s) {
-                    const uint64_t addr = d.base + (((span / 8u) * s) & ~(uint64_t)3);
-                    if (!guest_readable(addr, 256)) continue;
-                    const uint32_t* words = (const uint32_t*)(uintptr_t)addr;
-                    for (uint32_t k = 0; k < 64; ++k) {
-                        ++sampled_dwords;
-                        if (words[k]) ++nonzero;
+                // BYTES, not texels. Offsets computed in texels and applied as byte addresses sample
+                // only the first bytes-per-texel'th of an RGBA8 surface -- reporting a false
+                // ALL-ZERO for anything below that -- and run PAST the end of a BCn one, reporting a
+                // false has-content from unrelated memory. The block arithmetic is the same one
+                // PROSPER_DUMP_TILERAW uses above.
+                Gen5ImageFormatInfo cfi;
+                const bool cfmt = gen5_image_format(d.format, &cfi) && cfi.bytes_per_block &&
+                                  cfi.block_width && cfi.block_height;
+                const uint64_t span =
+                    cfmt ? (uint64_t)((d.width + cfi.block_width - 1) / cfi.block_width) *
+                               ((d.height + cfi.block_height - 1) / cfi.block_height) *
+                               cfi.bytes_per_block
+                         : 0ull;
+                if (cfmt && span >= 256) {
+                    uint32_t nonzero = 0, sampled_dwords = 0;
+                    for (uint32_t s = 0; s < 8; ++s) {
+                        // Every 256-byte window stays inside the footprint: the last one starts at
+                        // span-256 rather than at 7/8 of span plus a read that would overrun.
+                        const uint64_t off = ((span - 256ull) / 7ull) * s;
+                        const uint64_t addr = d.base + (off & ~(uint64_t)3);
+                        if (!guest_readable(addr, 256)) continue;
+                        const uint32_t* words = (const uint32_t*)(uintptr_t)addr;
+                        for (uint32_t k = 0; k < 64; ++k) {
+                            ++sampled_dwords;
+                            if (words[k]) ++nonzero;
+                        }
                     }
+                    // is_live_render_target() is safe to call here only because PROSPER_TEXCONTENT
+                    // is in parallel_draw_diagnostic_active()'s list, so draw realization is serial
+                    // while this diagnostic is armed. That callback drains and mutates the renderer's
+                    // write queue; calling it from parallel workers is UB and perturbs the very cache
+                    // state this reports.
+                    const bool cached = prosper::gpu::is_live_render_target(d.base);
+                    const int verdict = ((sampled_dwords == 0) ? 0 : (nonzero == 0 ? 1 : 2)) |
+                                        (cached ? 4 : 0);
+                    // Deduped on (base, VERDICT), never on base alone. An address-keyed probe reports
+                    // whatever the FIRST observation saw, which on a long run is the boot-time state,
+                    // so a surface filled moments later still reads as empty. That mistake cost three
+                    // published corrections before it was caught.
+                    static std::mutex cmx; static std::set<std::pair<uint64_t, int>> cseen;
+                    std::lock_guard<std::mutex> lk(cmx);
+                    if (cseen.insert({d.base, verdict}).second)
+                        fprintf(stderr,
+                                "[texcontent] base=0x%llx %ux%u fmt=%u bytes=%llu nonzero=%u/%u %s "
+                                "rtt=%s\n",
+                                (unsigned long long)d.base, d.width, d.height, d.format,
+                                (unsigned long long)span, nonzero, sampled_dwords,
+                                sampled_dwords == 0 ? "UNREADABLE"
+                                                    : (nonzero == 0 ? "ALL-ZERO" : "has-content"),
+                                cached ? "authoritative" : "MISS");
                 }
-                const bool cached = prosper::gpu::is_live_render_target(d.base);
-                const int verdict = ((sampled_dwords == 0) ? 0 : (nonzero == 0 ? 1 : 2)) |
-                                    (cached ? 4 : 0);
-                if (cseen.insert({d.base, verdict}).second)
-                    fprintf(stderr,
-                            "[texcontent] base=0x%llx %ux%u fmt=%u nonzero=%u/%u %s rtt=%s\n",
-                            (unsigned long long)d.base, d.width, d.height, d.format,
-                            nonzero, sampled_dwords,
-                            sampled_dwords == 0 ? "UNREADABLE"
-                                                : (nonzero == 0 ? "ALL-ZERO" : "has-content"),
-                            cached ? "authoritative" : "MISS");
             }
             if (getenv("PROSPER_GFXLOG") || getenv("PROSPER_TEXLOG")) {
                 const uint32_t* t = tv;   // the fetched T# (SGPR block or EUD spill)
--- a/prosper/src/gpu/execute/gpu_execute.hpp
+++ b/prosper/src/gpu/execute/gpu_execute.hpp
@@ -1678,6 +1678,24 @@ inline bool realize_draw_item(const GpuState& ds, const GpuState::Draw* draw, ui
                          (unsigned long long)(draw ? draw->command_order : 0));
         return false;
     }
+    // PROSPER_DRAWMAP=1 -- one line per distinct (fragment program, colour target, extent, scissor)
+    // with how many draws share it. Answers "which shader owns which pixels" as a static filter over
+    // a whole route in ONE run. Painting one shader at a time answers the same question one
+    // candidate per run; on a title with 66 fragment stages that is the difference between one run
+    // and a bisect. It is what identified Stray's full-screen composite (#3126).
+    if (getenv("PROSPER_DRAWMAP")) {
+        static std::mutex dmx; static std::map<std::string, uint64_t> dseen;
+        char k[192];
+        snprintf(k, sizeof k, "ps=0x%llx target=0x%llx %ux%u scissor=%u,%u..%u,%u tm=0x%x",
+                 (unsigned long long)rs.ps_addr, (unsigned long long)rs.color0_base,
+                 rs.color0_width, rs.color0_height,
+                 (unsigned)rs.scissor_left, (unsigned)rs.scissor_top,
+                 (unsigned)rs.scissor_right, (unsigned)rs.scissor_bottom, rs.cb_target_mask);
+        std::lock_guard<std::mutex> lk(dmx);
+        const uint64_t n = ++dseen[k];
+        if (n == 1 || n % 4096 == 0)
+            fprintf(stderr, "[drawmap] x%llu %s\n", (unsigned long long)n, k);
+    }
     const auto* fused_back = static_cast<const AgcShaderHeader*>(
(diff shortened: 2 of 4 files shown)


mattias800 (commented):
**REJECTED** — three blocking findings (B1–B3). The idea is good and three of the four instruments are close; B1 is a data race plus an instrument that perturbs its own subject, B2 is texel/byte arithmetic that can both under-sample and over-read, and B3 makes MIMG_SOFT stop working after its first occurrence in a way its own output cannot show.

Reviewed at head `a0fab6b`, read-only worktree. I did not rebuild or re-run ctest. I did run the two static Python gates (below).

---

## Blocking

### B1. `PROSPER_TEXCONTENT` calls a non-pure, non-thread-safe renderer callback from parallel draw workers

`prosper/src/gpu/agc/agc_shader_layout.cpp:949` calls `prosper::gpu::is_live_render_target(d.base)`. That resolves to `prosper/src/gpu/execute/gpu_executor.cpp:11146`, which invokes whatever lambda the frontend registered — `prosper/frontends/shared/live/live_renderer.cpp:1363`. Its **first statement** is:

    drain_guest_gpu_writes(g_rtt, invalidate_ds);

`drain_guest_gpu_writes` (`live_renderer.cpp:377`) is not a query. It swaps the pending guest-GPU-write queue out from under whoever else would drain it, clears `pending.overflowed`, then mutates `prosper::test::persistent_color_target_cache()`, `persistent_ds_cache()` and the RTT cache itself — `using RttCache = std::unordered_map<uint64_t, RttSurf>` (`live_renderer.cpp:150`) — including an outright `cache.clear()` on overflow (`live_renderer.cpp:398`). None of that is mutex-protected.

`build_shader_resources` is reached per draw per stage from `build_stage_table` → `realize_draw_item`, and `realize_draw_item` runs on up to eight worker threads (`gpu_executor.cpp:9420` inside `parallel_draw_worker_execute`; thread count at `gpu_executor.cpp:9324`). The new `cmx` mutex at `agc_shader_layout.cpp:936` protects `cseen` and nothing else.

Two separate consequences:

1. **Undefined behaviour.** Concurrent mutation of `g_rtt` and the persistent caches from N workers, concurrently with the render/submit thread. Today `is_live_render_target` has exactly one production caller — `frontends/shared/live/live_compute.cpp:6009`, on the serialized submit path — so this PR is the first off-thread caller and the first time that assumption is broken.
2. **The instrument perturbs its subject, invisibly.** Arming `PROSPER_TEXCONTENT` makes stage-build steal and apply the renderer's pending invalidations at a different time on a different thread. The `rtt=authoritative`/`rtt=MISS` field the probe prints is therefore partly a function of the probe having run. That is instrument-not-the-subject in its purest form, and nothing in the output says so.

The tree already has the mechanism for this: `parallel_draw_diagnostic_active()` at `gpu_executor.cpp:9351-9360` forces serial realization for `PROSPER_RTLOG`, `PROSPER_RESDUMP`, `PROSPER_DRAWDIAG`, `PROSPER_DESCRIPTOR_VALIDATE` and friends, for exactly this class of reason. `PROSPER_TEXCONTENT` and `PROSPER_DRAWMAP` are not in it. Adding them is a two-line fix and also disposes of the hot-path getenv question below.

### B2. TEXCONTENT's probe offsets are computed in TEXELS and used as BYTES

`agc_shader_layout.cpp:938-941`:

    const uint64_t span = (uint64_t)d.width * d.height;
    const uint64_t addr = d.base + (((span / 8u) * s) & ~(uint64_t)3);

`span` is a **texel** count; the eight probe offsets are then applied as **byte** offsets from `d.base`. Both error directions are reachable:

- **Under-sampling (false ALL-ZERO).** RGBA8 1920x1080: real footprint is `span*4`. The last probe sits at `0.875*span` bytes, i.e. ~22% into the surface — roughly the top 240 rows. A 1080p texture whose top quarter is black and whose content is below reads `ALL-ZERO` for the whole surface. In a black-screen investigation that is precisely the false negative that looks like the answer, and it lands in a `## Ruled out` row as a measurement.
- **Over-reading (false has-content).** BC1/BC4 are 0.5 bytes/texel, so the footprint is `span/2` while the last probe is at `0.875*span` — up to `0.375*span` bytes **past the declared allocation**. `guest_readable(addr, 256)` prevents a fault; it does not prevent reading the *next* allocation. A 2048x2048 BC1 texture passes the `>= 1920 && >= 1080` filter, so `has-content` can be produced entirely from someone else's bytes.

`guest_readable` is the right guard and `256` is the right size for the 64-dword read — the guard is fine, the offsets are not. The correct footprint is computed forty lines above **in the same loop**, by the neighbouring `PROSPER_DUMP_TILERAW` block: `gen5_image_format(d.format, &tf)` at `:896` then `nb = ceil(w/bw)*ceil(h/bh)*bpb` at `:898`. Reuse that and clamp the last probe to `nb - 256`.

### B3. `PROSPER_MIMG_SOFT` is unreachable for every occurrence after the first

In `prosper/src/gpu/recompiler/rdna2_emit_alu.cpp`, the pre-existing `[mimg-unresolved]` block sits inside `if (!res || wrong class) { … }` spanning `:6933`–`:6999`, and at `:6951` it returns early:

    if (!first_report) { ok = false; return true; }

The new MIMG_SOFT block is at `:7009`, **after** that block closes. So only the FIRST compile of a given `(program_address, pc, srsrc)` ever reaches the soft path; every later one hard-fails the stage regardless of the switch. That is not hypothetical:

- `PROSPER_NO_SHADER_CACHE=1` bypasses the cache entirely (`gpu_executor.cpp:1752`), so every draw recompiles — the second draw onward is hard-failed with `MIMG_SOFT=1` set.
- The shader cache evicts (`max_entries = 4096` plus a byte limit, `gpu_executor.cpp:1783` ff.), so a long run silently reverts to hard-fail partway through.

This is the failure mode the charter singles out: the recorded result *"the composite compiled and the frame did not change"* is equally consistent with "the softened sample doesn't matter" and with "the stage hard-failed anyway and nothing was softened", and **no line in the output distinguishes them**. `[mimg-soft]` prints once per key on the path that worked, and prints nothing on the path that silently didn't. Fix by moving the MIMG_SOFT block above `:6951`, or by making `:6951` fall through when the switch is armed.

---

## Non-blocking, but worth fixing before merge

**N1 — MIMG_SOFT's destination count matches only the non-D16 sampled-load case.** Compared against `write_mimg_results` (`rdna2_emit_alu.cpp:7046`):
- Non-D16 (`:7048-7057`): normal path writes one VGPR per set dmask bit, contiguous from `vd`. MIMG_SOFT writes `popcount(dmask)`. **Matches**, and its `vreg_old` / `rs.vreg[dst] = …` / `predicate_write` sequence at `:7017-7019` is byte-for-byte the neighbouring idiom. Good.
- D16 (`in.mimg_d16`, `:7059-7084`): normal path writes `ceil(popcount/2)` packed VGPRs. MIMG_SOFT writes `popcount` — up to **2x too many**, clobbering `v[vd+2..vd+3]` on a dmask-f load.
- `image_store` / `image_store_mip` (0x08/0x09, `:7091`) and non-returning atomics: `in.dst.value` is VDATA, a **source**. The normal path writes nothing there. MIMG_SOFT overwrites the shader's live source registers with 0.5f — a different class of wrong from "this sample returns a flat value", and one that propagates to whatever else those registers feed, which makes the counterfactual uninterpretable rather than merely wrong.
Suggest: skip the write entirely for the storage-only store/atomic opcodes, and halve `comps` when `in.mimg_d16`.

**N2 — DRAWMAP is unbounded in both memory and output** (`gpu_execute.hpp:1686-1695`). The key includes the scissor rect, so on a title that sets a per-element scissor every draw is a new key: one `[drawmap]` line per draw and one `std::map<std::string,…>` entry per draw, forever. The PR's framing ("one line per distinct …", "a static filter over a whole route in ONE run") assumes a small key space that the key itself does not guarantee. Neighbours in this tree bound themselves deliberately — the 16-resource cap at `rdna2_emit_alu.cpp:6982`, `describe_mapped_address`'s read budget. A `dseen.size()` cap plus one "N further keys suppressed" line keeps the claim honest. (`char k[192]` is fine — worst case is ~136 bytes, so no truncation-induced key merging.)

**N3 — DRAWMAP and COMPUTEMAP report *intent*, not *execution*, and their wording does not say so.** DRAWMAP at `gpu_execute.hpp:1686` runs before shader recompile, before `validate_runtime_descriptor_contract` (`:2072`) and before the later no-effect gate (`:2111`), so a draw dropped for `descriptor-contract` is still counted as owning those pixels. COMPUTEMAP at `gpu_executor.cpp:8214` prints `program=0x… **writes** 0x…` before `recompile_compute_shader_cached` at `:8226` and before the descriptor-contract decline that the comment at `:8281-8287` describes at length ("the device survived because the program never ran, which is indistinguishable from 'the bound worked'"). Neither is wrong as a *binding* census; "writes" and "owns these pixels" are what will mislead. Rename to "binds", or say so in the comment. `report_dropped_draw_target` exists because this exact distinction has cost time before.

**N4 — layering / include weight.** `agc_shader_layout.cpp:3` now includes the whole of `gpu/execute/gpu_execute.hpp` for one declaration, in a file that hand-declares its one other cross-module symbol at `:18-20` (`bool guest_readable(uint64_t, uint32_t);`) for exactly this reason. A matching two-line forward declaration avoids an agc→execute include and the transitive pull of `rdna2_to_spirv.hpp`, `command_processor.hpp` and `videoout_present.hpp` into that TU.

**N5 — dead condition.** `&& table` at `gpu_executor.cpp:8214` cannot be false: `table` is dereferenced unconditionally at `:8208` (`table->resources.size()`) and `:8209` (`assign_convention_bindings(*table, 2)`).

**N6 — ad-hoc dedupe key.** `sseen.insert(((uint64_t)b.diagnostic.program_address << 16) | in.pc)` (`:7023`) where the neighbour seventy lines above uses `std::set<std::tuple<uint64_t,uint32_t,int>>` and documents why the program must be in the key (`:6941-6944`). A `pc >= 65536` bleeds into the program bits. Prefer the same tuple.

---

## Verified good

**Zero-cost when unset — MIMG_SOFT is the best-gated of the four.** `if (!res && getenv("PROSPER_MIMG_SOFT"))` at `:7009` short-circuits on `!res` first, so the getenv is paid only on an already-failing reject path, and an unset run falls straight through to the original `if (!res) { ok = false; return true; }` at `:7028` with identical semantics. Confirmed by reading both branches.

**Locking is correct as far as it goes.** All four hold their mutex across both the container touch and the `fprintf`; DRAWMAP's `++dseen[k]` and its print are in one critical section, so the `xN` counts are right under parallel realization. B1 is not a defect in the local mutex — it is what the diagnostic *calls* while holding it.

**The (base, verdict) dedupe does what its comment claims.** `cseen` is keyed `{d.base, verdict}` with verdict encoding {UNREADABLE | ALL-ZERO | has-content} x {cached}, so an ALL-ZERO → has-content transition genuinely prints a second line and "it was empty" cannot masquerade as "it is empty" *in a log read in order*. Two gaps worth closing while you are in there: (a) the line carries no frame or submit ordinal, so a reader who greps for `ALL-ZERO` rather than reading sequentially gets the stale verdict back anyway — an ordinal would deliver the property the comment promises; (b) `sampled_dwords` varies from 64 to 512 depending on how many of the eight probes were readable, but partial readability collapses into the same verdict, so only the first denominator a base ever produced is printed.

**Gate tooling — claim verified, and its limits stated.** I ran both static Python checkers on the PR head (neither is a build or a CI job): `tools/env/check_diag_gates.py` → `== all checks passed ==`, SPLIT-CALL=3, SPLIT-LOCAL=19, TWO-GATE=86, total 108, baselined 108. The PR's claim is accurate and these four add no new unbaselined pattern. `tools/env/check_cached_env.py` also passes, with five pre-existing ordering notes, none from this diff. Reading the rules rather than the result: **neither tool can see any of B1–B3.** `check_diag_gates` is a lexical scanner for producer/printer gate splits and its own header states that cross-translation-unit state is invisible and that anything gated by something other than an env variable is out of scope by construction; `check_cached_env` only intersects `PROSPER_ENV_ON/VALUE` names against `setenv`-family names, and none of these four use those macros. A green run on both says nothing about correctness here.

**Hot-path getenv — consistent with the neighbours, inconsistent with the recorded lesson.** DRAWMAP adds one `getenv` per draw on the parallel path; TEXCONTENT adds one per *texture slot per stage build per draw*, since `:934` sits inside the per-slot loop. `realize_draw_item` already contains ~10 unconditional per-draw getenvs, so DRAWMAP is in keeping with its surroundings. But `gpu_execute.hpp:2065-2069` records #2287's lesson in this very function — *"One read per submit when the caller hoisted it, one per call otherwise — never one per draw per stage on a parallel worker"*, noting getenv takes a process-wide lock on Windows — and TEXCONTENT's placement is exactly that shape. The same comment explains why a `static const` cache is **not** the right fix here (tests arm these at runtime and a process-lifetime cache makes those arms vacuous, the #2214 defect). So the fix is the B1 fix: put both variables in `parallel_draw_diagnostic_active()` and the parallel path is off whenever they are armed.

---

Happy to re-review on a corrected head. B1 and B3 are each a few lines; B2 wants the `gen5_image_format` footprint that is already sitting in the same loop.

mattias800:
All three blocking findings fixed at head.

**B1** — `PROSPER_TEXCONTENT` is now in `parallel_draw_diagnostic_active()`'s list. You are right that this is exactly what that list is for, and that the callback is not a pure query — `drain_guest_gpu_writes` steals the queue and mutates an unsynchronised map, so the probe was both UB and a perturbation of the state it reports. Recorded at both ends.

**B2** — texel offsets applied as byte addresses. Now uses the same block arithmetic `PROSPER_DUMP_TILERAW` uses in the same file, and every 256-byte window is kept inside the footprint by spacing over `span-256` rather than `7/8*span` plus a read that overruns. The byte size is printed so a reader can check the span instead of trusting it.

**B3** — the soft path sat after the once-per-`(program, pc, srsrc)` dedupe, so only the first compile of each site was softened. Moved above it, with the reason in a comment.

**These were not hypothetical.** B2 and B1 corrupted the `ALL-ZERO`/`has-content` and `rtt=` numbers I published on #3126, and B3 made "the composite compiles under MIMG_SOFT and the frame still does not change" unsound — which was the load-bearing result that redirected that investigation away from the recompiler. All three conclusions are now retracted there, and #3140's whole line with them. Finding them was worth more than the PR.

Verification at head: `cmake` clean, `ctest --no-tests=error` 100% of 314, `check_diag_gates.py` all checks passed (108 findings, 108 baselined).

The six non-blocking items are noted and not yet acted on — the D16/store destination-count divergence in the soft write is the one I think most deserves a follow-up, since it is the same class of "wrong in a way the output cannot show". Re-review requested.

mattias800 (commented):
**APPROVED** — all three blocking findings from the review on `a0fab6b` are fixed, and I verified each one against the code rather than against the commit message. No new blocking finding. Everything below is non-blocking; several are cheap and I would take them before merge, but none of them should hold the PR.

Reviewed at head `25310dae`, read-only worktree. I did **not** rebuild, run ctest, run snapshots, or re-check CI — that is the author's verification. I did re-run the two static Python gates (below) and read every line of the diff.

Shape of the change, checked first because it bounds the risk: `git diff origin/master...HEAD --numstat` is **71/18/20/35 insertions and zero deletions** across four files. `git diff --check` is clean. `git log origin/master..HEAD --format=%B | grep -c '^Claude-Session:'` is 0. So there is no trap-41 exposure and no default-path line was rewritten — every addition sits behind a `getenv`.

---

## Blocking findings: none. Verification of B1-B3.

### B1 — fixed, and the fix reaches the right path.

`PROSPER_TEXCONTENT` is now in the list at `prosper/src/gpu/execute/gpu_executor.cpp:9351-9372`. The list is consulted at `gpu_executor.cpp:9484`, in `realize_gpustate_draws_parallel`, and the consult is an **early `return {}`** before the pool is touched — so with the variable armed the parallel batch is never built and the caller falls back to serial realization. That is the correct gate, not a cosmetic list entry.

I checked the two things the entry does not obviously cover:

- **Is there another thread that reaches `build_shader_resources`?** The only thread pool in `src/gpu` is `DrawRealizationPool` (`gpu_executor.cpp:9315`). The other two thread creations in the subsystem are `command_processor.cpp:3092` (`pend_worker`, which only drains the PM4 pend queue — it never enters shader layout) and `texture/tile.cpp:555`. The remaining `build_shader_resources` callers are `build_stage_table` (`:6477`, now serial), `realize_compute_dispatches` (`:7575`) and `diagnose_compute_dispatches` (`:8605`), and `realize_compute_dispatches` is called only from `:8522`, `:9804` and `:11482` — all on the submit path, none threaded.
- **Is the serial path the same thread the renderer uses?** Yes. `realize_gpustate_draws` is called from `execute_ordered_and_present` (`:11413`) and `execute_and_present` (`:11661`), and the pre-existing production caller of `is_live_render_target` — `frontends/shared/live/live_compute.cpp:6009`, in `execute_item` — runs from the same ordered-execution sequence. `live_renderer.cpp` creates no threads of its own. So serialising genuinely puts the callback back on the thread that already owned it.

### B2 — fixed; I checked the arithmetic against the format table, including the last window.

`agc_shader_layout.cpp:944-956` now derives the footprint in **bytes** from `gen5_image_format`:

    span = ceil(w/block_width) * ceil(h/block_height) * bytes_per_block

`Gen5ImageFormatInfo` (`agc_shader_layout.hpp:202-212`) defaults `block_width = block_height = 1`, and `bcn()` (`agc_shader_layout.cpp:256`) sets 4x4 — so the uncompressed path is correct by default and the BCn path by assignment. For **format 170** (`case 170: bcn(DataFormat::Bc1, 4, 8, true)`, `agc_shader_layout.cpp:290`) at 3840x2160: `ceil(3840/4)=960`, `ceil(2160/4)=540`, `960*540*8` = **4,147,200 bytes**. Matches.

Window containment: `off = ((span - 256) / 7) * s` for `s` in 0..7 (`:954`). Integer division floors, so at `s=7`, `off <= span - 256`; `off & ~3` only lowers it further; the 256-byte read therefore ends at or before `base + span` for **every** window including the last. `d.base` is `base_256 << 8` (`agc_shader_layout.cpp:231`), so `addr` is 4-byte aligned and the `uint32_t*` deref is aligned. The `span >= 256` guard at `:949` makes the subtraction safe. Degenerate overlap (a `span` small enough that the eight windows coincide) is unreachable under the `width >= 1920 && height >= 1080` filter — the smallest reachable footprint is a 1-byte-per-block 1920x1080 at ~2 MB.

Two supporting points worth recording: `span` is the mip-0 footprint of the **allocation** base (the base-level view is computed later, at `:1046`), and a tiled or pitch-padded allocation is always at least that large, so the containment argument holds for tiled surfaces too. And the line now prints `bytes=%llu`, which is what lets a reader confirm from the log alone that the block arithmetic ran rather than the texel arithmetic.

### B3 — fixed, and it is the placement that matters, not the comment.

`rdna2_emit_alu.cpp`: the MIMG_SOFT block is at **`:6956`**; `static std::mutex mimg_mutex` is at `:6975`; the dedupe early-return `if (!first_report) { ok = false; return true; }` is at **`:6986`**. So the switch is evaluated before any first-report dedupe, and every compile of a site is softened, not only the first. The `mimg_reported` set and its mutex are function-scope statics declared *after* the new block, which is fine — skipping a local static's declaration never leaves it half-initialised.

Unset-path equivalence, which is the other half of B3's claim: `if (!res && getenv("PROSPER_MIMG_SOFT"))` short-circuits on `!res` first, and with the variable clear control falls through to exactly the code master had, reaching the unchanged `if (!res) { ok = false; return true; }` at `:7035`. Master has zero occurrences of the identifier, so this is a pure addition and there is no second copy left behind.

### Item 5 — the new dedupe key.

`std::set<std::tuple<const void*, uint64_t, int>>` keyed on `shdr.code` (`:979-981`). `AgcShaderHeader::code` is `const void*` (`agc_shader_layout.hpp:69`) and points into **1:1-mapped guest memory** — it is a guest shader-code address, not a host heap allocation, so it is not subject to allocator reuse. Reuse requires the guest itself to recycle its shader pool onto the same address, and the consequence would be one suppressed line for the later shader. That is strictly better than the key it replaces: `(base, verdict)` merged *every* sampler of a surface unconditionally, which is the same loss on every run rather than on a rare one. Fine for a diagnostic, and the change is an improvement.

### Gates

Both static Python checkers pass at this head. Neither is a build or a CI job.

- `python3 prosper/tools/env/check_diag_gates.py` -> `== all checks passed ==`, rc 0. 569 files scanned, 729 `PROSPER_*` variables at 1160 sites; findings SPLIT-CALL=3, SPLIT-LOCAL=19, TWO-GATE=86, **total 108, baselined 108** — the four new variables add no unbaselined gate pattern.
- `python3 prosper/tools/env/check_cached_env.py` -> `== all checks passed ==`, rc 0, with the five pre-existing ordering notes (all in `tools/`), none from this diff.

Stated as the previous review stated it, because it stays true: **neither gate can see any of B1-B3.** They are lexical scanners for producer/printer gate splits and for `PROSPER_ENV_ON`/`setenv` collisions. A green run here is evidence about gate hygiene, not about correctness.

---

## Non-blocking

**N1 — the comments still describe the old key, in two places, and so does the PR body.** `agc_shader_layout.cpp:929-933` says *"Deduped on (base, VERDICT), never on base alone"*, and `:971-974` repeats the same paragraph verbatim, immediately above `:975-977` which correctly says *"Keyed on (shader, base, verdict)"*. The PR body's "The dedupe rule, written into the code" section says `(base, VERDICT)` too. Three stale statements of a key that changed in the newest commit, sitting next to the correct one. This is exactly the failure the charter calls out — a stale claim carries the authority of the file it sits in — and here it is self-contradicting within six lines. Delete the two stale copies and fix the body.

**N2 — the new comment at `:963-967` overclaims what serialising fixed.** It says calling `is_live_render_target` from parallel workers *"is UB and perturbs the very cache state this reports"*. The first half is fixed. The second is only half fixed: the callback (`live_renderer.cpp:1363`) still begins with `drain_guest_gpu_writes` (`live_renderer.cpp:377`), which swaps the pending queue, mutates `persistent_color_target_cache()` / `persistent_ds_cache()` and can `cache.clear()` the RTT map on overflow (`:398`) — and the probe now calls it **once per large texture slot, per stage build, per draw**. Serialising removed "on a different thread"; it did not remove "at a different time". The invalidations are the same set and the renderer would apply them anyway, so I do not think this is a defect — but the comment now reads as an all-clear, and the next agent will inherit it. Say what was actually fixed.

**N3 — MIMG_SOFT now suppresses `[mimg-unresolved]` entirely, and that report is the one that says *which* resolution failed.** Because the soft block `return true`s at `:6973`, arming the switch skips the whole ungated report at `:6997` and its `available:` census at `:7031`. Under MIMG_SOFT you get `program=0x… pc=… -> constant N comps` and nothing about `srsrc`, `srt_tag`, `ud_alias`, or what the table did hold — which is the information you need to act on the counterfactual the switch produces. The earlier review offered two fixes and this took the one with the side effect; the other one ("make `:6986` fall through when the switch is armed") keeps both. Two lines, and worth it.

**N4 — the moved block is indented four spaces shallower than the block it lives in** (`:6940-6974` at 12 spaces, the surrounding `if (!res || wrong class) {` body at 16). Cosmetic, except that it visually implies the soft path sits *outside* the resolution-failure guard, and which guard it sits under is precisely what B3 was about.

**N5 — the previous review's N1 is unaddressed and still correct.** I re-derived it rather than repeating it. `write_mimg_results`' D16 path (`:7068-7089`) calls `write_packed` once per **two** set dmask bits, so it writes `ceil(popcount/2)` VGPRs; the soft block writes `popcount` (`:6958-6966`). On a dmask-f D16 load that is 4 writes where the real path does 2, clobbering `v[vd+2..vd+3]`. And for `storage_only_op` (0x08/0x09/0x0f/0x11-0x1a, `:6926-6929`) `in.dst.value` is VDATA, a **source**: the soft block overwrites the shader's live source registers with 0.5f, which propagates rather than merely flattening one sample. Halving `comps` under `in.mimg_d16` and skipping the write for the store/atomic opcodes would make the counterfactual interpretable.

**N6 — the previous review's N2, N3, N4 and N5 are also unaddressed**, all still accurate as written: DRAWMAP's `std::map` is unbounded even though the `n == 1 || n % 4096 == 0` throttle bounds the *printing* (`gpu_execute.hpp:1685-1697`); DRAWMAP sits after the early no-effect return but before descriptor-contract validation and COMPUTEMAP prints "writes" before `recompile_compute_shader_cached`, so both are binding censuses wearing execution wording; the `gpu_execute.hpp` include in `agc_shader_layout.cpp:3` is heavy for one declaration; and `&& table` at `gpu_executor.cpp:8214` is dead — `table->resources.size()` at `:8208` and `*table` at `:8209` dereference it unconditionally.

**N7 — a correction to the previous review's N6, in the direction that lets it go.** That review said the soft-path key `(program_address << 16) | in.pc` bleeds when `pc >= 65536`. `in.pc` is a **dword index**, assigned in `rdna2_walk` (`rdna2_decode.cpp:1166`) under `while (pc < dwords)`, and every caller bounds `dwords` at `0x10000` — so `pc <= 65535` and the collision is unreachable today. It is exactly one bit from being reachable and the neighbouring `std::tuple<uint64_t,uint32_t,int>` at `:6978` is free, so I would still switch it; but it is a robustness nit, not the live defect it was posted as.

**N8 — `rtt=MISS` conflates two different facts.** `is_live_render_target` (`gpu_executor.cpp:11152`) returns false both when the renderer says it does not own the address **and** when `g_live_target_query` was never registered. In a run without the live renderer every line reads `ALL-ZERO rtt=MISS`, which is precisely the false alarm the PR body says this instrument exists to prevent. `tools/gpu_replay` does register it (`gpu_replay.cpp:1546`, `:2617`), so the exposure is capture/offline/test paths rather than the intended route — but a `rtt=NO-RENDERER` third state costs one line and makes the instrument prove it could have answered.

**N9 — an unmapped Gen5 IMG_FMT now prints nothing at all from this probe.** `cfmt` false at `:945` skips the whole block, where the pre-fix code sampled (wrongly) and printed. Mostly covered, because the neighbouring ungated `[t#] UNMAPPED Gen5 IMG_FMT %u … (extend gen5_image_format)` at `:1024-1026` fires a few lines later — except that warning is suppressed when `PROSPER_RTT` or `PROSPER_RTT_PERTARGET` is set (`:1019-1022`), which is a normal per-target render run. In that combination the surface is invisible to both. An `else` printing the base and format once would close it.

**N10 — the PR body's Verification block is from the first head.** The ctest count and the gate output quoted there were produced at `a0fab6b`, before the three fixes and before the key change. Re-post at `25310dae`, or say which head they came from.

---

## Also verified good

- **The unset path is byte-identical to master for all four.** Every block is a `getenv` guard around added code; MIMG_SOFT's is `!res && getenv(...)` on an already-failing reject path, so an unset run pays one `getenv` only where the stage was going to be rejected anyway.
- **Locking.** All four hold their mutex across both the container mutation and the `fprintf`, so DRAWMAP's `xN` counts and the three dedupe sets are correct under any concurrency that remains.
- **No out-of-bounds in the soft write.** `rs.vreg` is `std::unordered_map<int, uint32_t>` (`rdna2_to_spirv_internal.hpp:4484`), not a fixed array, so `vd + k` cannot overflow a buffer; the `vreg_old` / assign / `predicate_write` sequence matches the neighbouring idiom exactly.
- **No persistent contamination from MIMG_SOFT.** The shader cache is in-memory only (`gpu_executor.cpp:1764` ff., `cache.entries` with `max_entries = 4096` and a byte limit), with no on-disk backing — so a MIMG_SOFT run cannot leave deliberately-wrong SPIR-V behind for a later evidence run.
- **Each instrument proves it fired.** DRAWMAP prints on `n == 1`, COMPUTEMAP on first `(program, addr)`, TEXCONTENT prints `bytes=` and `nonzero=N/M` so a reader can see both the footprint arithmetic and how many of the eight windows were readable, and MIMG_SOFT prints once per softened site — so "zero `[mimg-soft]` lines" now unambiguously means "nothing was softened", which is the property B3 was about.

mattias800:
## Merging — approval basis, stated because the head moved after it

The APPROVED review was submitted against `25310dae`; the head is now `aedc78dc`. Per the charter a
new push detaches a review from head, so the approval has to be re-established **by content** rather
than assumed:

```
git diff -U0 25310dae..aedc78dc -- prosper/src/gpu/agc/agc_shader_layout.cpp \
  | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' \
  | grep -vE '^[+-][[:space:]]*//' | grep -vE '^[+-][[:space:]]*$'
  -> no output
```

Every changed line in the delta is a comment. No emitted code differs from what was approved, so the
approval stands. The delta fixes review items N1 and N2 — two comments that stated something the code
beside them denied (the dedupe key described as `(base, verdict)` after it had grown a shader
component; serialising described as fixing the measurement perturbation when it fixed only the UB).

CI: 9 SUCCESS / 2 SKIPPED on `aedc78dc`.

N3 is deliberately **not** in this PR — it is behavioural (arming `PROSPER_MIMG_SOFT` suppresses the
`[mimg-unresolved]` report that names the failing descriptor) and slipping a behaviour change in after
approval is exactly what the by-content rule exists to prevent. Filed as #3143.

The remaining non-blocking notes are recorded in the review and not lost by merging.

Issue #3142, opened 2026-08-30
Stray title background: pak data is written correctly, then the range is zero 62s later when the shader samples it


mattias800:
# The pak read into the background texture reports success, and the memory is still zero

The decisive run: title screen reached (30 frames, settled `max_nonblack` 0.0069), carrying
`PROSPER_FILELOG` and `PROSPER_TEXCONTENT` together. Line numbers from the same log:

```
582814  [apr] measure-read  … dst=0x303cb90000 off=0x1ec61250 size=4718737 OK
582854  [apr] read-submit   … dst=0x303cb90000 off=0x1ec61250 size=4718737 got=4718737 OK
584282  [texcontent] shader=0x300c150000 base=0x303cb90000 3840x2160 fmt=170
        bytes=4147200 nonzero=0/512 ALL-ZERO rtt=MISS
```

The read completes **1,428 log lines before** the sample, and the memory at that address is still
**zero across all 8 probe windows**. So the timing hypothesis from the previous comment is dead: this
is not data arriving late.

**4,718,737 bytes are reported delivered to `0x303cb90000`, and afterwards guest memory there reads
zero.**

Rigour checks on that statement:

* Exactly **one** `texcontent` line exists for this base in the whole run, so the verdict never changed
  later — the `(shader, base, verdict)` key would have printed a second line if it had.
* Exactly **two** `[apr]` lines target it: the `measure-read` and the `read-submit`. No competing write.
* The other input the composite samples, `0x303d390000`, has **zero** `apr` reads and **does** have
  content (27/512) — the opposite pairing, which is itself worth explaining.

## Caveats, stated rather than buried

* The probe samples 8 × 256-byte windows spread across the **texture footprint** (4,147,200 bytes)
  while the read was 4,718,737 bytes at that destination. If the pixel data began deep inside that
  read, the windows could in principle miss it — but they span the first 4.1 MB of a 4.7 MB read, so
  this would need an unusual layout.
* BC1 data can legitimately contain zero blocks; 512 of 512 dwords zero across 8 spread windows is not
  plausible for real image content, but it is not literally impossible.

## What to look at

`got=4718737 OK` is the pak reader's own report that it delivered the bytes. Either it writes them
somewhere other than the guest address it names, or something clears them between the write and the
sample. The `[apr] read-submit` path and its `dst=` translation are where that resolves — and the
`measure-read compatibility` line immediately before it suggests there are two paths through this
code, which is exactly the shape where a destination can diverge.

[...]

PR #3144, opened 2026-08-30
diag(apr): verify each guest write against its own source, and one clock to order diagnostics by

mattias800:
Two instruments that came out of #3142, plus the finding they produced. Refs #3142, #3126.

## Why

`[apr] read-submit … got=4718737 OK` is the pak reader's report that it delivered a chunk. It proves
the bytes were **transferred**. It does not prove they were **non-zero** — a `pread` over a sparse or
wrongly-offset region returns full size and zeros — and it does not prove they are still there when a
consumer samples the address later. #3142 sat in exactly that gap: a kernel-verified 4.7 MB write
followed by an all-zero read of the same address, with nothing able to say which of the two was lying.

## `PROSPER_APR_VERIFY` — read the destination back and compare it to its own source

After `apr_write_guest_dst`'s `process_vm_writev` returns full size, re-read the destination through
`process_vm_readv`, **in the same thread, immediately**, and compare against the source buffer at
identical offsets (the same eight-window pattern `PROSPER_TEXCONTENT` uses).

Two properties are deliberate:

* **Source-vs-destination, not against an expectation.** Whatever the source holds is the oracle, so
  the probe cannot be wrong about what it should have seen. This matters here because the expected
  content is genuinely sparse — 5.63% non-zero dwords over the region in question — and a probe
  carrying its own idea of "looks like data" would have called a correct write a failure.
* **Read back, don't dereference.** An unmapped destination reports `UNREADABLE` instead of faulting,
  so the probe cannot become a new failure mode of the path it observes.

`PROSPER_APR_VERIFY=1` verifies every write; `PROSPER_APR_VERIFY=0xADDR` verifies one destination,
which keeps the log readable on a title that streams thousands of chunks.

### `PROSPER_APR_VERIFY_SELFTEST` is the positive control, and it is not optional

A census of all-MATCH is *void*, not negative, until the instrument is shown capable of reporting a
loss — a discriminator that cannot show its lever moved says nothing. The self-test writes a
`0xA5A5A5A5` pattern over the destination's first window immediately before verifying:

```
[apr-verify] t=… dst=0x3001c80000 size=785406 src_nonzero=402/512 dst_nonzero=410/512
             MISMATCH-WRITE-LOST w0=64/56! w1=44/44 w2=42/42 w3=55/55 … w7=49/49
```

It reports the loss and localises it: only the poisoned window carries `!`, the flag that marks a
window whose bytes differ. **A pattern rather than zeros, and this is load-bearing**: under a content
comparison, zeroing a window whose source is already zero is undetectable — common on data that is
~5.6% non-zero, and certain on padding. The arm also checks the poison write landed and that the
pattern differs from the source, reporting `SELFTEST-INVALID` rather than a misleading `MATCH` if
either precondition fails. A poisoned run cannot report `MATCH`: window 0's offset is always 0, so it
is exactly the poisoned range.

## `diag_now_us()` — one monotonic clock, because line order is not happens-before

Diagnostics in different subsystems write to one stderr stream from **different threads**. Comparing
two of them by position in the log is unsound, and doing it produced a wrong published conclusion on
#3142 ("the read completes 1,428 log lines before the sample" was read as a timeline; it establishes
no ordering at all). `src/diagnostics/diag_clock.hpp` gives them a shared monotonic reading — no
anchor, registry, or shared mutable state, since `steady_clock` is process-wide and independently
taken values compare directly. Stamped onto `[apr-verify]` and `[texcontent]`, the two that had to be
compared and could not be.

It sits **outside** `diagnostics.hpp`'s `PROSPER_DIAGNOSTICS` compile-time split on purpose: both
consumers are gated at runtime by environment variables and must work in a default build. A time
source that vanishes unless a build flag is set is the exact failure that header's own banner records.

## What they found

The pak read is not the defect. On the title-screen route, `0x303cd10000` (BC1_SRGB 3840×2160, read
for shader `0x300c150000`):

| source | non-zero dwords at texture offset 0 |
| --- | --- |
| the pak file via `dd`/`od`, no emulator | 25 / 64 |
| `PROSPER_APR_VERIFY` at write time | 25 / 64, and **byte-identical** to the source |
| `PROSPER_TEXCONTENT` **65.735 s** later | 0 |

Byte-identical when written, entirely zero when sampled, no intervening write. A second base,
`0x302a300000`, is sampled by five different shaders across **152 seconds** and is zero every time
after a byte-verified write. #3142 is therefore a
**lifetime** defect — the range is zeroed, freed or remapped between load and use — not a read defect,
which is a different investigation from the one it opened with.

## Verification (exact head)

```
cmake --build prosper/build-linux -j6                  # clean
ctest --test-dir prosper/build-linux --no-tests=error -j4
  -> 100% tests passed out of 315, rc=0
python3 prosper/tools/env/check_diag_gates.py
  -> == all checks passed ==  (108 findings, 108 baselined)
```

Live, **re-measured after the byte-identity fix** (the earlier 21,217-MATCH census was taken by the
population comparison this PR replaced and is not evidence for byte identity):

```
full route  -> 21,499 / 21,499 MATCH, 0 MISMATCH, 0 UNREADABLE, 0 SELFTEST-INVALID
self-test   -> MISMATCH-WRITE-LOST, only the poisoned window flagged
```

Note what the census does and does not cover: **eight 256-byte windows per write**, so ~2 KiB is
byte-compared per call, not the whole 1.00 GB the run streams. Sampling, not exhaustive verification.

## Risk

Both switches are **default-off** and read once per call site. With `PROSPER_APR_VERIFY` unset,
`apr_verify_write` returns after one `getenv` and nothing else runs; the emitted behaviour of
`apr_write_guest_dst` is unchanged. `diag_now_us()` is a header-only inline with no callers outside
the two diagnostic lines. No test, snapshot, or rendering path changes.

`PROSPER_APR_VERIFY_SELFTEST` **deliberately corrupts guest memory** and exists only to validate the
probe — it is documented as such at the call site, and is meaningless unless `PROSPER_APR_VERIFY` is
also set.

**Review requested**: the probe writes into guest memory on one path (the self-test), and reads guest
memory on the other, so the failure modes are worth an independent read.

---

**Update — review round 2.** All four round-1 blockers were fixed in code; the remaining blocker was
that **this body still described the instrument the fix replaced** — it published the pre-fix
21,217-MATCH census as the evidence, said the control "zeroes" a window when it writes a pattern,
pasted a transcript this code cannot produce, and argued the control works because "the totals differ
by exactly the 56 dwords destroyed", which is the population reasoning the fix rejects.

That is this PR's own **trap 236** — counting instead of comparing — reappearing in the artifact that
redirects #3142, one commit after the code stopped doing it. Recorded rather than quietly edited,
because the interesting part is that fixing the instrument did not fix the claims already written from
it, and nothing failed when it didn't.

Body corrected above against the re-measured run. Also removes the `static` from the self-test's
poison buffer: several APR threads can be inside `apr_verify_write` at once, and a shared mutable
buffer is a data race even when every writer stores the same value.

The `## Ruled out` row for `docs/STRAY_STATUS.md` is still deliberately deferred — #3132 rewrites that
exact section — and the commitment is recorded on #3142 rather than only here. It lands in this branch
before merge.


mattias800 (commented):
## REJECTED — head `f06ea754`

Independent review. I did not re-run the author's build/ctest (315/315, rc=0 reported); I re-ran the two
env gates and verified the window arithmetic numerically. Four blocking findings, all cheap. The design
is right — a source-vs-destination oracle with a mutation arm is exactly the shape this project asks for,
and putting the clock outside the `PROSPER_DIAGNOSTICS` split is correct and well argued. What blocks is
that the instrument measures something weaker than what the PR body publishes, and the control that is
supposed to license the census can fail without saying so.

### What I checked and found sound

* **Window arithmetic is correct.** `off = (((size - 256) / 7) * s) & ~3` for `s=0..7`, guarded by
  `size < 256`. I ran the formula over 256, 257, 262, 263, 264, 511, 512, 4096, 785406, 4718737:
  no window's `off + 256` exceeds `size` in any case. The `& ~3` only decreases `off`, and
  `7 * floor((size-256)/7) <= size-256` bounds the last window. `hle_file.cpp:2711`.
* **Source read is in bounds and (today) aligned.** Both POSIX callers pass an `mmap` page —
  `hle_file.cpp:2853` (`slot`, rounded up to 0x1000) and `:3325` (`staging`, same) — so
  `(const uint32_t*)(src + off)` at `:2719` is page-aligned + 4-aligned and `src + off + 256 <= src + size`.
  See N3 for why that is a call-site accident rather than a contract.
* **Call-site placement is complete and correct.** `:2752` covers the direct `process_vm_writev` success,
  `:2766` the post-commit retry, gated on `retried`. Neither failure path (`!committed` → `return false`,
  `retried == false`) calls it. No success path is missed.
* **`detail` cannot actually overflow.** Eight iterations, longest field `" wN=UNREADABLE"` = 14 bytes,
  so `used <= 112` against a 512-byte buffer. The guard at `:2728` is dead code — see N2 for why it is
  still wrong.
* **Read-back cannot fault, on both POSIX hosts.** `process_vm_readv` returns `-1` rather than raising, and
  the Darwin shim (`host/platform/posix_shim.hpp:85`, `mach_vm_read_overwrite`) has the same contract, so
  the `UNREADABLE` path holds on macOS too. The write-watch arms pages to read-only
  (`guest_write_watch.cpp:739`), so the read-back is not perturbed by an armed watch.
* **`diag_clock.hpp` layering and build placement.** `target_include_directories(prosper_core PUBLIC src)`
  at `CMakeLists.txt:98` is unconditional, and the `PROSPER_DIAGNOSTICS` filter at `:94` is scoped to
  `src/diagnostics/(core|storage)/` — so a header at the folder root is reachable in the default build.
  The reasoning in the file banner is correct. It is the first `src/gpu` and `src/hle` dependency on
  `src/diagnostics` (previously only `host/image/boot_program.cpp`), but it is header-only over `<chrono>`
  and `<cstdint>` with no link edge and no guest-visible state, so it does not violate the folder's own
  "nothing here participates in running the guest" boundary. Syntax-checked standalone with
  `g++ -std=c++20 -fsyntax-only` — clean.
* **Env gates.** `python3 prosper/tools/env/check_diag_gates.py` → `== all checks passed ==`, 570 files,
  731 variables at 1162 sites, findings SPLIT-CALL=3 / SPLIT-LOCAL=19 / TWO-GATE=86 (108 total, 108
  baselined), rc=0. Also ran `check_cached_env.py` → `== all checks passed ==` (103 cached, 127 armed);
  live `getenv` is the sanctioned choice there, so the uncached read is fine.

---

## Blocking

### B1 — the verdict compares non-zero POPULATION COUNTS, not content, so "byte-exact" is not what was measured

`hle_file.cpp:2735` decides `MATCH` vs `MISMATCH-WRITE-LOST` on `src_nz == dst_nz` — the number of
non-zero dwords, summed across all eight windows. That is not a comparison of the destination against
its source. Two consequences:

* A destination holding entirely different bytes with the same non-zero-dword population reports `MATCH`.
* Because the totals are summed before the test, a window that gains `n` non-zero dwords and a window that
  loses `n` cancel, and the verdict token — the thing a census greps — says `MATCH`.

The PR body publishes the stronger claim: *"Byte-exact when written"*, and *"21,217 writes and 21,217
MATCHes"*. The instrument establishes neither. This is precisely the charter's *"a true statement reached
by a route that does not establish it"* — the conclusion may well be right, and the derivation does not
support it. It matters here because that row is what redirects #3142 from a read defect to a lifetime
defect, and the next agent will inherit the redirection without the caveat.

Add to that the coverage: 8 windows x 256 bytes = at most **2 KiB compared per write**, so on the 4,718,737-byte
write in the body's own table the probe examined 0.043% of it. "1.00 GB covered" in the Verification section
is the size of the writes verified, not the bytes compared.

Fix is one line — both buffers are in hand at `:2719`/`:2714`, identical layout and length:
`memcmp(back, sw, sizeof back)` per window, with the verdict driven by that and the counts kept as the
localiser they already are. Then the body's "byte-exact" is true. If you would rather keep the count
comparison, that is defensible for a diagnostic — but then rename the verdict and downgrade the body and
the #3142 comment to what a population comparison supports.

### B2 — the positive control can be a silent no-op, in two independent ways

The self-test is what licenses the whole census, and the PR body says so ("it is not optional"). As written
it cannot tell you when it was invalid:

* **(i)** `hle_file.cpp:2703` discards the result: `(void)process_vm_writev(getpid(), &pl, 1, &pr, 1, 0);`.
  If the poison does not land — a watch re-armed to read-only between `:2751` and the probe, an unmapped
  page — the arm reports `MATCH`, which reads as "the probe is incapable of detecting a loss". A control
  that discards its own success signal is the failure the charter's positive-control rule is about.
* **(ii)** The poison is 256 zero bytes and the comparison is by count (B1), so on any write whose first
  window's source is **already all-zero** the mutation is undetectable by construction. On the sparse data
  this probe was built for (the body's own 5.63% non-zero) that is roughly 1 write in 40 by chance, and 100%
  of writes whose first 256 bytes are header padding or a sparse region. Run with
  `PROSPER_APR_VERIFY=0xADDR` on such a destination and the control reports `MATCH` with nothing to
  distinguish that from a broken probe.

Note (ii) survives the B1 fix: `memcmp` of zeros against zeros is still 0. So the arm needs to assert its
own precondition — check the poison write returned `sizeof poison`, and check the poison actually differs
from `sw` — and print a distinct token (`SELFTEST-POISON-FAILED` / `SELFTEST-NOOP`) when it does not.
That is the difference between "the instrument could have broken this zero" and "the instrument was never
given anything to break".

The corruption itself I consider acceptable and adequately gated: it sits after the `PROSPER_APR_VERIFY`
early-out at `:2690-2693`, so `PROSPER_APR_VERIFY_SELFTEST` alone does nothing, it is documented at the call
site as deliberately destructive, and a mutation arm that does not mutate the subject is not a mutation arm.
See N4 for the one gating detail I would change.

### B3 — `PROSPER_APR_VERIFY` is silently POSIX-only

`apr_verify_write` is defined and called inside the `#ifndef _WIN32` block (`hle_file.cpp:2590-2769`). The
Windows sibling at `:2779` has no probe and, more importantly, **no note that the variable does nothing
there**. Its own comment two lines above says *"ONE name for 'put these bytes in the guest's own
destination' on both hosts... the two builders differing in how they write a destination is precisely the
divergence the shared core below exists to make impossible"* — and this PR introduces exactly a divergence
in that function, undocumented. A Windows operator arming the variable gets zero lines, which is
indistinguishable from "no APR writes happened", i.e. a diagnostic that fails silently on a supported
platform.

I am not asking for the Windows implementation (the charter's parity exception covers that, and there it is
a plain `memcpy` into mapped memory so a `memcmp` would be trivial if you want it). **A one-line comment on
the Windows sibling stating that `PROSPER_APR_VERIFY` is POSIX-only discharges this.**

### B4 — the falsification is not in the project record

The PR establishes a real result: #3142 is a lifetime defect, not a read defect. It also records an
instrument trap in its own words — log-line order between threads was read as a timeline and *"produced a
wrong published conclusion"*. Neither lands anywhere the next agent reads:

* `docs/STRAY_STATUS.md` has a well-developed `## Ruled out` (line 56) and gains nothing here. "The pak read
  into the background texture is the defect" belongs in it, with the write-time-vs-62 s evidence and the
  #3142 link.
* `docs/GAME_COMPAT_ORCHESTRATION.md`'s numbered instrument-trap table (currently at 234) has no row for
  "stderr interleaving from different threads is not happens-before". This one cost a published conclusion,
  which is the table's entry criterion. Allocate with `tools/docs/trap_number.py` (it scans open PRs too),
  and run `tools/docs/check_merge_result.py` immediately before merging.

The charter is explicit that a falsification living only in a PR body or an issue comment is one the next
agent re-derives at full cost. I checked #3142 and you did post the finding there as a comment — that is
good and not sufficient under the rule.

---

## Non-blocking

**N1 — the body's risk claim is wrong, and the fix is to move one line.** `diag_now_us()` is called at
`hle_file.cpp:2689`, *before* the `getenv` gate at `:2690`. So the body's *"With `PROSPER_APR_VERIFY` unset,
`apr_verify_write` returns after one `getenv` and nothing else runs"* is not true: a `clock_gettime` runs on
every APR write in the default build. The cost is negligible; the wrong statement in the record is the
problem. `t_us` is only consumed by the `fprintf` at `:2731`, so move the read below the early-outs and the
sentence becomes true as written.

**N2 — the `detail` guard does not guard.** `if (used >= (int)sizeof detail) break;` at `:2728` is (a)
bypassed entirely by the `continue` on the UNREADABLE path at `:2717`, and (b) placed *after* the `snprintf`
whose `sizeof detail - (size_t)used` argument is the thing that would underflow — so if it ever mattered it
would fire one iteration too late, with `detail + used` already past the end. It is safe only because
`used <= 112` is unreachable-by-construction, which is the fact worth stating. Either clamp
(`const size_t room = used < (int)sizeof detail ? sizeof detail - used : 0;`) or delete the guard and write
the bound in a comment. A guard that cannot do its job invites the next reader to trust it.

**N3 — source alignment is a call-site accident, and `sampled` overstates coverage.**
`(const uint32_t*)(src + off)` at `:2719` needs `buf` 4-aligned; the signature takes `const void*` and both
current callers happen to pass `mmap` pages. One `assert`/comment tying that to the contract would keep a
future caller with a `malloc`'d odd-offset buffer from making this UB. Separately, `sampled` is always 512
regardless of unique coverage: I confirmed numerically that for any `size <= 262` all eight windows collapse
to offset 0, so `src_nonzero=n/512` reports the same 256 bytes eight times.

**N4 — `PROSPER_APR_VERIFY_SELFTEST=0` enables the corruption.** `:2700` tests presence only, while
`PROSPER_APR_VERIFY` at `:2690` at least checks `!*ev`. For the one switch in this PR that permanently
corrupts guest memory, presence-only gating is the wrong direction; `PROSPER_ENV_ON`-style value checking
would cost nothing.

**N5 — the window formula is now duplicated.** `agc_shader_layout.cpp:955` and `hle_file.cpp:2711` are the
same expression, and the comment explaining why it stays in bounds exists only in the first. This is the
charter's "refactor toward the architecture as you go" case: one shared helper in `src/diagnostics/`
(alongside the clock, since both consumers are diagnostics) plus a unit test pinning `off + 256 <= size`
for the boundary sizes would make the bound checkable rather than re-derived. That test is also the only
thing in this PR that could be pinned by ctest at all.

**N6 — `[texcontent]`'s new stamp dates the FIRST observation, not the sample.** The `fprintf` at
`agc_shader_layout.cpp:992` only runs when `cseen.insert({shdr.code, d.base, verdict})` succeeds
(`:988`), so `t=` is when that `(shader, base, verdict)` triple was first seen. That is arguably the right
choice for "when did this surface first read as empty", and the 62.351 s figure in the body depends on
which of the two it is. One clause in the comment naming it would stop the next reader treating it as
"when this texture was last sampled".

**N7 — perturbation caveat.** The `[texcontent]` block states plainly that arming it changes when guest
writes land, so its verdict is evidence about that run. `[apr-verify]` adds eight `process_vm_readv`
syscalls per verified write (~170k on the body's 21,217-write census) and says nothing equivalent. Worth
one sentence in the banner comment.

**N8 — `src/diagnostics/AGENTS.md` heading.** The appended paragraph is good and I agree with where the
file lives; the heading still reads "observing prosper's own **startup**", which `diag_clock.hpp` widens.
One-word fix while you are in there.

---

Re-review on request once B1-B4 are addressed. B3 is a comment, B4 is markdown, and B1/B2 are a `memcmp`
plus two checked return values — I do not expect a redesign.

mattias800 (commented):
All four blocking findings fixed at head `f3a489a7`, and B1 was correct in a way that matters beyond
this PR — thank you for catching it.

## B1 — comparing populations, not content

You are right, and the consequence had already escaped: I published **"byte-exact when written"** on
#3142 on the strength of `src_nz == dst_nz`, which never tested that property. The verdict is now
`memcmp` over both windows; the counts remain as detail and a diverging window is flagged `!`.

**Re-measured under the stricter verdict: 21,499 / 21,499 `MATCH`.** So the conclusion survives — but
it was unsupported when I made it, and is supported now. That distinction is the finding, not the
number.

## B2 — a control that could pass without testing anything

Both holes closed. The poison is now a `0xA5A5A5A5` pattern rather than zeros (your point (ii): a
zero poison is undetectable under a content comparison wherever the source is already zero, which on
~5.6%-sparse data is common and on padding is certain), the write's return is checked, and the pattern
is compared against the source. If either precondition fails the probe prints `SELFTEST-INVALID`
instead of a misleading `MATCH`. Verified live:

```
[apr-verify] t=… dst=0x3001c80000 size=785406 src_nonzero=402/512 dst_nonzero=410/512
             MISMATCH-WRITE-LOST w0=64/56! w1=44/44 w2=42/42 … w7=49/49
```

Only the poisoned window carries `!`.

## B3 — POSIX-only, now stated

A comment at the Windows sibling says the probe is absent there, why (`process_vm_readv` has no
in-file equivalent), that the **write** remains shared so the divergence is in verification only, and
what to add (`ReadProcessMemory`) if a Windows guest-memory defect ever needs it.

## B4 — the falsification is in the record

Two instrument-trap rows in `GAME_COMPAT_ORCHESTRATION.md`, both from published errors:

* **235** — reading log line order as a timeline. The vicious part is that a *larger* gap is more
  convincing, so the wrong inference strengthens with distance.
* **236** — a verifier that counts instead of comparing. It passes every test you would think to
  write, including a mutation arm, because the mutation that exposes it must be content-preserving.

The `## Ruled out` row for `STRAY_STATUS.md` is **deliberately not in this PR**: #3132 is editing that
same section and would conflict. It goes in before this merges, once #3132 lands — flagging it so the
omission is a sequencing decision rather than something dropped.

## N1

Fixed — `diag_now_us()` was called before the env gate, so the body's "returns after one `getenv`"
was false. The stamp is now taken after it.

## Re-measured with the corrected instrument

The #3142 correlation is unchanged in shape and better separated. Of the ALL-ZERO samples whose base
received a byte-verified write, **6 are write-first** (the real contradiction) and **4 are
sample-first** at −149 s — the latter being textures simply not loaded yet, which the shared clock now
distinguishes and which line order never could:

| base | shader | write → sample |
| --- | --- | --- |
| `0x303cd10000` | `0x300c150000` | **+65.735 s** |
| `0x302a300000` | five shaders | +2.0 s, +2.7 s, +67.2 s, +151.6 s, +152.0 s |

`0x302a300000` is sampled five times across 152 seconds and is zero every time, after a byte-verified
write. Re-review requested.

mattias800 (commented):
## REJECTED — head `f3a489a7`

Re-review of the four blocking findings. **All four are genuinely fixed in the code and the docs**, and
B1's fix is right in the way that matters: the verdict is byte identity now, the counts survive only as
the localiser, and the `!` marker makes the diverging window findable. I re-derived the safety of the new
comparison rather than taking the diff's word for it (details below). One thing blocks, and it is not
code: **the PR body still publishes the pre-fix census and a self-test transcript this code cannot
produce.** That is a body edit, nothing more.

I did not re-run the author's build/ctest.

---

## The four prior blockers — all discharged

### B1 (prior) — comparing populations, not content: **fixed, and I checked the arithmetic**

`hle_file.cpp:2740` drives `diverged` off `memcmp(back, sw, sizeof back)`; `:2749` drives the verdict off
`diverged`; the counts are demoted to detail at `:2741`. Both operands are the right 256 bytes at the
**same** offset — `back` is filled from `dst + off` (`:2722`), `sw` is `src + off` (`:2728`), `sizeof back`
is 256 for both. Source read is in bounds: I re-ran the window formula **exhaustively over every
`size` in 256..200,000** plus 262, 263, 264, 511, 785406, 4718737 and `0xffffffff` — `off + 256 <= size`
in all 8 windows, 0 violations. The comment at `:2736-2739` states the property and why the weaker one
was wrong, which is the part that stops it regressing.

### B2 (prior) — the control could be a silent no-op: **fixed, and the control is now sound**

Both holes are closed at `:2700-2713`: the pattern is `0xA5A5A5A5` rather than zeros, `poison_landed`
checks the write returned `sizeof poison` (`:2711`), `poison_distinct` compares the pattern against the
**source** (`:2709`), and `SELFTEST-INVALID` (`:2751`) fires when either fails.

I proved the property the arm exists for — **a poisoned run cannot report `MATCH`**. Suppose it did:
`MATCH` requires `unreadable == 0`, `diverged == 0`, and `selftest && poison_landed && poison_distinct`.
Window 0's offset is `(((size-256)/7) * 0) & ~3 == 0` for every `size`, so window 0 is exactly the range
the poison covers; `unreadable == 0` means it was read back, `poison_landed` means it holds the pattern,
`poison_distinct` means the source does not — so `memcmp` differs and `diverged >= 1`. Contradiction.
The three failure directions are also right: poison rejected by the kernel → `SELFTEST-INVALID`; source
already equal to the pattern → `SELFTEST-INVALID`; poison landed → `MISMATCH-WRITE-LOST` with `!` on w0
only. `size < 256` is excluded at `:2690`, so the 256-byte `memcmp` against `buf` at `:2709` is in bounds.

### B3 (prior) — POSIX-only, undocumented: **fixed**

`hle_file.cpp:2796-2802` says the probe is absent on Windows, why (`process_vm_readv` has no equivalent
in this file's toolkit), that **only verification** diverges while the write stays shared — which is the
right distinction against the warning two lines above it — and names `ReadProcessMemory` as the sibling
to add. That is exactly what B3 asked for and no more.

### B4 (prior) — the falsification is not in the record: **discharged for the traps; the STRAY row is a
sound deferral, but see N7**

Traps **235** (log line order is not happens-before) and **236** (a verifier that counts instead of
comparing) are both accurate against the code they describe, well-scoped as *general* instrument traps
rather than title notes, and both name the cost that earns a row. 236 credits the review and states the
vicious property precisely — the mutation that exposes a population check has to be content-preserving,
which is the case nobody constructs. Gates:

```
check_numbered_table.py --ordered --table-header Instrument --baseline <origin/master>
  -> 14 tables, 298 rows, unbroken; numbered 1..236, unique and ascending, none of master's 234 deleted; rc=0
check_merge_result.py
  -> clean merge; merged result 1..236, no deletions vs origin/master; rc=0
check_trap_citations.py --table GAME_COMPAT_ORCHESTRATION.md
  -> 83 references on 76 lines in 27 files, naming 56 distinct rows, all resolving; rc=0
trap_number.py
  -> master highest 234; PR #3144 claims 235, 236; 4 open PRs scanned, 1 touching this file; no collision
```

I also checked `check_diag_gates.py`: `== all checks passed ==`, 570 files, 731 variables at 1162 sites,
SPLIT-CALL=3 / SPLIT-LOCAL=19 / TWO-GATE=86 (108, all baselined), rc=0.

**On deferring the `STRAY_STATUS.md` `## Ruled out` row: I checked the premise rather than taking it, and
it holds.** PR #3132 is not merely touching that file — its hunk `@@ -71,9 +297,53 @@` rewrites the
`## Ruled out` section itself, retracting the wall-clock-route row, adding a `VOID, not falsified`
scoping paragraph and six re-scoped rows. Appending there from this branch would conflict, and worse,
the row would be written against numbers #3132 is in the middle of correcting. Sequencing after #3132 is
the right call, not a rationalisation. See N7 for the one thing that needs doing about it.

### N1 (prior) — **fixed.** `diag_now_us()` moved to `:2691`, below the `getenv` gate at `:2689-2690`,
so the body's "returns after one `getenv`" is now true as written.

---

## Blocking

### B1 — the PR body still publishes the pre-fix census, and a self-test transcript this code cannot produce

The body is the artifact that redirects #3142, and at head it describes the instrument that was replaced.
Six statements in it are now wrong or unsupported:

1. *"The first census was **21,217 writes and 21,217 MATCHes** ... So the MATCH census is a measurement."*
   That census was decided by `src_nz == dst_nz`. The number that licenses the conclusion is the
   **21,499 / 21,499 under byte identity** you re-measured — it is in your review reply and in #3142's
   14:48 comment, and it is not in the body.
2. *"The self-test **zeroes** the destination's first window"* — false at head; `:2707-2708` writes
   `0xA5A5A5A5`.
3. The transcript block (`dst_nonzero=346/512 ... w0=0/56`) cannot be emitted by this code: no `!`
   markers, and the poison is no longer zeros. A reader who runs the documented command gets different
   output.
4. Its reasoning — *"the totals differ by exactly the 56 dwords destroyed"* — is the population argument
   B1 rejected, offered as the proof the control works.
5. *"1.00 GB covered"* still reads as bytes compared. It is the size of the writes verified; the probe
   compares at most 8 x 256 = 2 KiB per write, i.e. 0.043% of the 4,718,737-byte write in the body's own
   table.
6. The section headed **"Verification (exact head)"** reports results from `f06ea754`, two commits and one
   `hle_file.cpp` change ago.

Nothing here needs code. But this is the same shape as the finding the PR itself documents in trap 236 —
a published claim standing on evidence that does not establish it — and the charter is explicit that a
stale figure carries authority it no longer deserves. Rewrite the census sentence around 21,499/21,499
under `memcmp`, replace the transcript with a current one (your reply's `w0=64/56!` line is ideal:
it shows the `!` localiser and the `SELFTEST-INVALID` verdict exists), say what the verdict compares,
and re-stamp the verification block at the exact head. `gh pr edit` fails here on a Projects-classic
GraphQL error without updating anything — use `gh api -X PATCH repos/OWNER/REPO/pulls/3144 -f body=...`
and read the body back.

---

## Non-blocking

**N1 — `poison` should not be `static`.** `:2707-2708` changed it from `static const ... = {0}` to a
mutable `static` refilled on every call. Two APR threads verifying concurrently is a data race on it.
It is benign in value (both write the same bytes), but the partial-fill window can make one thread
compute `poison_distinct` against a half-zeroed array — conservative direction, so it would report
`SELFTEST-INVALID` rather than a false pass, and it is only reachable in the deliberately-destructive
arm. Dropping `static` costs 256 bytes of stack and removes it entirely.

**N2 — `memcmp` runs twice per window**, at `:2740` and again inside the `snprintf` argument at `:2742`.
One `const bool diff = memcmp(back, sw, sizeof back) != 0;` reads better and is obviously the same
predicate in both places.

**N3 — prior N2 is half-discharged.** The dead guard is gone (good), but the bound that makes its absence
safe is not written down. I confirmed it: the longest field is `" wN=UNREADABLE"` at 14 bytes, the numeric
form maxes at `" wN=64/64!"` = 10, so 8 windows give **112 bytes + NUL against `detail[512]`** — `used`
can never reach `sizeof detail`, so neither `sizeof detail - (size_t)used` nor `detail + used` can go bad.
**No overflow is possible.** One comment line pins that for the next reader.

**N4 — verdict precedence is safe, and worth one clause.** `UNREADABLE-WINDOWS` outranks
`MISMATCH-WRITE-LOST` outranks `SELFTEST-INVALID`. No ordering can yield a false `MATCH` (proved under B2
above), and every non-`MATCH` token is greppable, so a census cannot be inflated. The one thing it loses:
under `SELFTEST`, `MISMATCH-WRITE-LOST` no longer separates "the poison was detected" from "the poison
never landed and a real loss was detected". The `!` on w0 distinguishes them by eye; a clause in the
banner saying so would stop someone reading a selftest `MISMATCH` as automatic proof the control fired.

**N5 — the stamp could move two lines lower.** `:2691` is after the `getenv` gate (N1 discharged) but
before the `want != dst` address filter at `:2693`, so `PROSPER_APR_VERIFY=0xADDR` still takes a
`clock_gettime` on every APR write. Negligible, and the body's claim is about *unset*, so nothing is
wrong — just exact if you are already there.

**N6 — trap 235 and #3142 carry different numbers for the same pair.** The row says the write→sample gap
measured **+62.351 s** for `0x303cd10000` / `0x300c150000`; your re-measured table in the review reply and
#3142's latest comment gives **+65.735 s** for that same pair. Both are real (the delta is per-run, not a
constant), but a reader cross-checking the doc against the issue sees two figures for one claim. Four
words naming the run settles it.

**N7 — the deferred `STRAY_STATUS.md` row currently lives only in a PR review comment.** The deferral
reasoning is correct (verified above), and the finding is durably recorded in trap 236 and in #3142's own
correction comment, so nothing is lost today. What is not durable is the *commitment*: "it goes in before
this merges" is exactly the kind of promise the charter says gets re-derived at full cost, and it is
recorded in the one place that disappears from view once this merges. Cheapest discharge is a line on
**#3142** (open, outlives both PRs) or on **#3132** naming the row to add after it lands. Either way, the
row must be in the tree before this merges — that is the charter rule, not a preference of mine, and it is
the only thing I would hold the merge on beyond B1.

**Prior N3 (source alignment as a call-site accident), N4 (`PROSPER_APR_VERIFY_SELFTEST` gated on presence,
not value, for the one switch that permanently corrupts guest memory), N5 (window formula now duplicated
between `agc_shader_layout.cpp:955` and `hle_file.cpp:2720`), N6, N7 and N8** (`src/diagnostics/AGENTS.md:1`
still reads "observing prosper's own **startup**", which `diag_clock.hpp` widens) are unaddressed. They were
non-blocking then and remain so; N4 is the one I would most like taken.

---

Re-review on request. B1 is a body edit — no code change, and I do not need to see the code again.

mattias800 (commented):
## APPROVED — head `68695f30`

Round 3. Scope as asked: the PR body, the new commit, and the deferred `STRAY_STATUS.md` commitment.
I did not re-run the author's build/ctest/CI. Per round 2's instruction I spot-checked rather than
re-derived the four discharged blockers — `memcmp` at `hle_file.cpp:2742`, the pattern poison and its
two preconditions at `:2709-2714`, the POSIX-only note at `:2798-2804`, traps 235/236 at
`GAME_COMPAT_ORCHESTRATION.md:363-364` — all still as round 2 described them.

**Round 2's B1 is discharged on all six counts, and I checked the transcript by arithmetic rather than
by eye.** No blocking findings.

---

## B1 (round 2) — the body described the pre-fix instrument: fixed

1. **Census.** The body now publishes `21,499 / 21,499 MATCH, 0 MISMATCH, 0 UNREADABLE, 0
   SELFTEST-INVALID` and explicitly retires the old figure — *"the earlier 21,217-MATCH census was taken
   by the population comparison this PR replaced and is not evidence for byte identity"*. That is the
   right shape: the stale number is named and killed rather than quietly dropped, so a reader who
   arrives via trap 236 (which still quotes 21,217 as the *pre-fix* number, correctly) can reconcile the
   two.
2. **"zeroes"** is gone — *"writes a `0xA5A5A5A5` pattern over the destination's first window"*, which is
   `:2709-2710`.
3. **The transcript is producible by this code.** I derived it against the two format strings rather
   than accepting it. `:2743` is `" w%u=%u/%u%s"` with arguments `s, dn, sn` — destination count first,
   source second — so `w0=64/56!` is dst=64 (all 64 dwords hold the pattern, hence non-zero) against
   src=56, with `!` from the `memcmp` at `:2744`. The header at `:2747` prints `src_nonzero` before
   `dst_nonzero`, and `sampled` is 8x64=512, so `src_nonzero=402/512 dst_nonzero=410/512` requires
   `410 = 402 - 56 + 64`, which is exactly what poisoning one window whose source held 56 non-zero
   dwords produces. It holds. The unpoisoned windows print `dn == sn` with no `!`, as shown. The verdict
   token is right too: under selftest with `diverged >= 1` the ternary at `:2750-2754` reaches
   `MISMATCH-WRITE-LOST` before the `SELFTEST-INVALID` arm, which is what the transcript shows. The only
   deviation is cosmetic — the real output is one line; the body wraps it. Fine, and the ellipsis for
   w4-w6 is marked.
4. **The population argument is gone.** *"only the poisoned window carries `!`, the flag that marks a
   window whose bytes differ"* is a content argument about the right thing.
5. **Coverage no longer reads as bytes compared:** *"eight 256-byte windows per write, so ~2 KiB is
   byte-compared per call, not the whole 1.00 GB the run streams. Sampling, not exhaustive
   verification."* That is the caveat the row redirecting #3142 needed.
6. **Verification block** — see N1; not blocking, and I explain why below rather than asserting it.

Nothing else in the body now overstates the instrument. I re-checked the claims round 2 did not list:
`PROSPER_APR_VERIFY=1` vs `=0xADDR` matches the `want > 0x1000` filter at `:2693`; *"returns after one
`getenv` and nothing else runs"* matches `:2689-2690` with the clock now below it at `:2691`;
*"header-only inline with no callers outside the two diagnostic lines"* is exactly true — a branch-wide
grep finds `diag_now_us` at `agc_shader_layout.cpp:992` and `hle_file.cpp:2691` and nowhere else. The
`## What they found` table says *"byte-identical **to the source**"*, not to the pak file, which is the
honest reading of what the probe compares.

## Commit `68695f30` — correct and complete

`static uint32_t poison[64]` becomes a stack local at `hle_file.cpp:2709`, still fully written by the
loop at `:2710` before either use, so there is no uninitialised read; the reason is stated at
`:2707-2708`. Correct.

**Complete — I swept the rest of the function for shared mutable state.** Across `:2688-2756` everything
else is frame-local: `t_us`, `want`, `selftest`/`poison_landed`/`poison_distinct`, `src`, the five
counters, `detail[512]`, `used`, and the loop-local `back[64]`. No function-static, no file-static, and
nothing reached through a pointer the caller shares — `buf` and `dst` are per-call. The two `getenv`
calls read `environ` without writing it, and the single `fprintf` at `:2746` is one FILE-locked call, so
concurrent probes cannot interleave inside a line. That is the whole surface; `poison` was the only one.

## The deferred `STRAY_STATUS.md` row — N7 discharged, precondition restated

Recorded on **#3142**, comment `2026-08-30T15:02:35Z`: it names the falsification owed to that doc's
`## Ruled out`, states the #3132 conflict as the reason for deferring, and commits to landing it in this
branch before merge. That is durable and outlives both PRs, which is what round 2 asked for.

**The precondition itself still stands and I am not waiving it by approving.** #3132 is still `OPEN` as
of this review, so the row is not yet writable. Per the charter, the row must be in the tree before this
merges — an approval on the code and the body is not an approval to merge without it.

---

## Non-blocking

**N1 — "Verification (exact head)" names no head.** The section header claims a property the block cannot
be checked against, and its ctest run predates `68695f30`. I did not block on it because the gap is
provably inert, and I checked rather than assumed: `PROSPER_APR_VERIFY` appears nowhere in the branch
outside `hle_file.cpp` and trap 236's prose, so no test can enter `apr_verify_write` at all, and a
storage-class change on a stack buffer inside a doubly-env-gated branch cannot move a test result. One
SHA under the header turns "trust me" into something a reader can verify, and stops the next round from
having to re-derive this.

**N2 — the two window formulas are not identical, and the body says they are.** *"the same eight-window
pattern `PROSPER_TEXCONTENT` uses"*: `agc_shader_layout.cpp:955` is `((span - 256ull) / 7ull) * s`,
while `hle_file.cpp:2722` adds `& ~(uint64_t)3` (needed there because `:2730` casts to `uint32_t*`). So
the same base can be sampled at offsets up to 3 bytes apart by the two diagnostics. Irrelevant to the
25/64-vs-0 comparison in the body's table, but "the same pattern" invites exactly the cross-read by
window ordinal where it would bite.

**N3 — the eight windows overlap below `size == 2048`.** `off_1 = ((size-256)/7) & ~3` is under 256
whenever `size < 2048`, so windows 1.. sit inside the poisoned first window; at `size == 256` all eight
are the same range. Two body sentences are therefore upper bounds rather than facts at small sizes:
*"only the poisoned window carries `!`"* and *"~2 KiB is byte-compared per call"*. Round 2's proof that a
poisoned run cannot report `MATCH` is untouched — it only needs window 0 — and every APR write in
evidence is 785 KB or 4.7 MB, so nothing measured is affected. A `size >= 2048` clause on the coverage
sentence would make it true as written.

**N4 — round 2's N4, restated because the body now depends on it.** `MISMATCH-WRITE-LOST` outranks
`SELFTEST-INVALID` at `:2750-2753`, so the body's *"reporting `SELFTEST-INVALID` ... if either
precondition fails"* holds against a misleading `MATCH` but not against `MISMATCH`: a run where the
poison did not land *and* a real loss occurred prints `MISMATCH-WRITE-LOST`. Nothing unsound follows —
no false `MATCH` is reachable — but one clause in the banner would stop someone reading a selftest
`MISMATCH` as automatic proof the control fired.

**N5 — round 2's N6 is the one I would most like taken, and it is now the only surviving contradiction
in the record.** Trap 235 (`GAME_COMPAT_ORCHESTRATION.md:363`) measures the `0x303cd10000` /
`0x300c150000` pair at **+62.351 s**; this body and #3142's corrected comment both say **+65.735 s** for
that same pair. Both are real per-run figures, but the trap row is the durable artifact and it is the
only place carrying a number the issue contradicts — which is the shape the charter warns about, in a
row this PR wrote about exactly that. Four words naming the run settles it.

**N6 — still open from round 2, unchanged in judgement:** the double `memcmp` per window (`:2742` and
again in the `snprintf` argument at `:2744`); the missing one-line note pinning why `detail[512]` cannot
overflow; `PROSPER_APR_VERIFY_SELFTEST` gated on *presence* rather than value, so `=0` arms the one
switch that permanently corrupts guest memory — two rounds of review now agree that is the one worth
taking; and `src/diagnostics/AGENTS.md:1` still titled *"observing prosper's own **startup**"* though the
file's own new closing paragraph correctly widens the folder past startup.

---

The instrument, the control, and now the claims written from it agree with each other. Approved.

mattias800:
## Merge basis — the approval survives a rebase, established by content

The APPROVED review was on `68695f30`. That SHA was rebased away when #3132 merged, so the approval
cannot be inherited by identity and had to be re-established by content:

```
# the branch's OWN source contribution, before and after the rebase
git diff <old-base> 68695f30      -- prosper/src   -> 221 lines
git diff origin/master HEAD       -- prosper/src   -> 221 lines
diff of the two (index/hunk headers stripped)      -> 0 lines
```

Identical. Everything added since the approval is documentation, and it is the documentation the
review **required before merge**:

* `STRAY_STATUS.md` gains the `## Ruled out` row that was deferred while #3132 rewrote that exact
  section. That deferral was the reviewer's stated merge precondition and was recorded on #3142 so it
  could not be quietly dropped. #3132 has merged; the row is now in the tree.
* trap 235 quoted **+62.351 s** while #3142 quotes **+65.735 s** for what reads as the same
  measurement (review N5). They are two runs — the second under the corrected byte-identity verdict —
  and the durable row now says so, rather than leaving a reader with two numbers for one pair and no
  way to tell which is current.
* `BLOG.md` gets the finding. No screenshot, which is the case the charter explicitly keeps: the title
  screen is still black, so there is nothing new to show, and without an entry a result like this
  leaves no trace anywhere the project owner reads.

The two remaining non-blocking notes are recorded and not lost: the two window formulas differ by
`& ~3` so "the same eight-window pattern" is true of the pattern but not the offsets, and the windows
overlap below `size == 2048`, making "~2 KiB per call" an upper bound rather than a constant.

PR #3145, opened 2026-08-30
diag: catch the moment guest data disappears, and four mechanisms that do not cause it

mattias800:
Instruments for #3142, and the falsifications they produced. Refs #3142, #3126.

## The state they were built for

#3144 established that Stray's pak reads land **byte-identically** in guest memory (21,499/21,499
`memcmp`-verified) and that the same address reads all-zero when the shader samples it. Every
instrument up to that point sampled two points and inferred the middle. These watch instead.

## `PROSPER_ZEROWATCH`

Polls each large APR destination every 50 ms and reports the non-zero → all-zero transition, dumping
`/proc/self/maps` for the address at that instant.

It fires **at least 11–13 times per run** — a floor, not a total: the watch list's 64-entry cap is
reached on the very first run, which the probe now says out loud instead of dropping silently. The
flip is fast — **0.5–2.3 s** after the load, not the 65 s I had been reasoning about. That 65 s was only the delay until the *shader* sampled; the data was gone
within a second. It also reproduces in a **2-minute** run, not the 7-minute title route, because it
happens during asset load.

The maps line settles the question two-point sampling cannot: at the moment of the flip the mapping is
**unchanged** — still `rw-s` on the dmem memfd at the same offset. Not unmapped, not replaced by an
anonymous mapping, not re-pointed. And because the range is `MAP_SHARED` on a memfd, reading the
mapping *is* reading the file, so there is no "stale view" case to exclude: the **content** was zeroed.

## `PROSPER_PHYSLOG`

Logs the three paths that could plausibly do it — `dmem_zero` (the hole punch), `map_phys_at` (the
documented single chokepoint for every dmem alias) and `guest_memory_gpu_write` (the renderer writing
into guest memory) — each stamped with `diag_now_us()` so they order against `[apr-verify]` and
`[texcontent]` soundly rather than by log position (trap 235).

## What they falsify

| hypothesis | evidence |
| --- | --- |
| `dmem_zero` punches a hole over the range | all **20** calls in the run are at startup, before every one of these writes |
| the VA is re-mapped to a different phys | one `[physmap]` per VA, created ~2 ms *before* its write; none after |
| a second VA aliases the phys and is written | zero aliasing mappings — each phys range is mapped exactly once |
| the renderer writes back over it | `guest_memory_gpu_write` fires **once** per run, covering none of the zeroed ranges |
| the address is recycled for another asset | `0x303cc90000` receives exactly one payload and is never written again |

## The write trace, and the limitation worth carrying forward

`PROSPER_DMEM_WRITE_TRACE` already records writer RIPs, but only prints on guest `_exit` and the fatal
raise — neither of which a bounded capture run reaches, so on every `tools/screenshot` run it was
recording RIPs and printing them nowhere. `PROSPER_ZEROWATCH` now fires its report at the instant a
watched range goes zero.

That run reports `status=invalid reason=host-write`, **`page-faults=0`**: no *guest* write faulted on
the watched page, and a *host* write invalidated the watch. Only four call sites in the tree write
guest memory from the host and all four are file reads, so the invalidation is the APR load itself.
**This instrument therefore cannot watch a range that APR loads into** — that is the blocker for naming
the writer, and it is a property of the trace's lifecycle rather than a result about Stray.

## Verification (exact head)

```
cmake --build prosper/build-linux -j6                        # clean
ctest --test-dir prosper/build-linux --no-tests=error -j4    # 100% of 315, rc=0
python3 prosper/tools/env/check_diag_gates.py                # == all checks passed ==
```

Live: five Stray boots.

## Risk

All three switches are **default-off**. With them unset each site is one cached `getenv` and nothing
else; `PROSPER_ZEROWATCH` starts no thread and arms nothing. No rendering, test or snapshot path
changes.

**Review requested** on one point in particular: `zerowatch_poll_forever` reads the watch vector under
its mutex but the vector is only ever appended to, and I would like a second opinion on the lifetime
argument. An earlier revision of this branch also carried a `SIGSEGV`-based write trap that chained to
prosper's handler; it killed the run at asset load and was **removed rather than debugged**, since
fighting the existing write-watch machinery is exactly the hand-rolled-guard hazard the charter warns
about. It is not in this diff.

---

**Update — review round 2.** Three blocking findings, all fixed at `3008860c`:

* **B1** was the serious one: `WENT-ZERO` printed the full payload size while the poll sampled only
  the first 256 bytes, and `seen_nonzero` was decided from that same window — which
  `apr_verify_write`'s own comment sixty lines below says is *commonly all-zero on this exact data*.
  A payload with a zero-padded head could therefore never report, silently, for precisely the class
  under investigation. `seen_nonzero` is now seeded at arm time from the **source** buffer across
  eight windows, and the line carries `sampled=256B@0` so the verdict states its own basis.
* **B2** — two silent coverage gaps. The 64-entry cap dropped arming with no output, and only the
  fast write path armed, leaving the commit-retry path (which `MAP_FIXED`s fresh anonymous pages over
  the destination first — the most interesting class) unwatched. Both fixed; the line now reports
  `armed-via=`. The cap warning fires immediately, which is what demoted "11–13" to a floor above.
* **B3** — the falsifications and the corrected timing were in issue comments and no `.md` file.
  `docs/STRAY_STATUS.md` gains the `## Ruled out` row, including what blocks naming the writer so
  nobody re-derives it.

Both questions I raised came back clean, and the reviewer sharpened one of them: the in-run access is
safe **because of the mutex**, not because the vector is append-only — my phrasing was the hazard. The
detached thread is safe only because every exit path here is `_Exit`, which is now a comment at the
thread rather than an unstated assumption.

---

**Update — review round 3.** Round 2's B2 and B3 were confirmed fixed; two findings remained and both
are addressed at `f7f1c136`.

* **B1 was a defect I introduced fixing round 1's B1.** Seeding `seen_nonzero` from the source removed
  the silent miss, but the probe still read window 0 of the *destination* — two unrelated places — and
  since arming now returns early when the seed finds nothing, the poll-loop gate became **vacuous**. A
  payload with a zero head and data at window 3 would have printed `WENT-ZERO` for a range that never
  changed, and because `guest_dmem_write_trace_report()` is one-shot, a false first flip burns the
  write trace on a non-event. That is worse than the miss it replaced. The entry now records *which*
  window the source was non-zero at, the probe reads that window, and `sampled=256B@0x<off>` states it.
* **B2 was right to refuse the cap claim as unmeasured**, and it is measured now: the warning fires
  once in the same run that produced the flips.

**Re-measured at `f7f1c136`: 14 flips, every one seeded at offset `0x0`.** So the fix changes no
observation on this title — the earlier numbers stand. It makes the correctness structural rather than
a property of this title's payload layout, which is the reason to make it and worth stating plainly
rather than dressing up as a corrected measurement.

Also restores two things the `STRAY_STATUS.md` row had dropped (round-2 non-blocking): the second base
`0x302a300000`, read zero by five different shaders across 152 s — one address going quiet could be a
descriptor pointing somewhere stale, two cannot — and the fifth falsification, that the address is
recycled for another asset.

### DIFF
--- a/prosper/docs/STRAY_STATUS.md
+++ b/prosper/docs/STRAY_STATUS.md
@@ -294,8 +294,8 @@ drops still discard the background.
   read-back (25 of 64 non-zero dwords at texture offset 0). And the range is nonetheless **entirely
   zero when the shader samples it**: base `0x303cd10000`, BC1_SRGB 3840×2160, read by
   `0x300c150000`, `nonzero=0/512` at +65.7 s after its verified write.
-  So this is a **lifetime** defect — something zeroes these ranges between load and use — not a read
-  defect, and the pak reader is not the place to look. #3142.
+  So the pak reader is not the place to look; something zeroes these ranges between load and use, and
+  the row below names it. #3142.
 
 - **"The range is zeroed some time during the ~65 s before the shader reads it."** Falsified, and the
   65 s was an artifact of when the *shader* happened to look rather than of when the data went away.
@@ -306,6 +306,15 @@ drops still discard the background.
   file offset, so it is neither unmapped nor re-pointed. And because that mapping is `MAP_SHARED` on a
   memfd, reading the mapping **is** reading the file — there is no "stale view" case, so the content
   itself was zeroed. #3142, #3145.
+  - **The writer is the guest's own libc, and that reclassifies the whole issue.** With #3147's
+    re-baseline mode the write trace observes the window it previously could not, and attributes
+    **64 of 64 events to `libc.prx`** — guest code — storing sequentially from the buffer base at a
+    16-byte stride through a tight `0x3d71`–`0x3dd0` RIP loop, on one thread. The guest clears its own
+    staging buffer about a second after loading it, which is ordinary engine behaviour. So this is
+    **not** a memory-lifetime defect and never was: the pak reader delivers correctly, the guest
+    disposes of its buffer correctly, and prosper's defect is that a shader still reads that address
+    afterwards. The open question is now a descriptor one — why `0x300c150000`'s texture resolves to a
+    staging buffer the engine has finished with — which also fits its `rtt=MISS`. #3142.
   - Four mechanisms measured and excluded, each in the same runs: **`dmem_zero`'s hole punch** (all 20
     calls occur at startup, before every one of these writes); **a re-map of the VA** (one `[physmap]`
     per VA, created ~2 ms *before* its write, none after); **a second VA aliasing the phys** (none —
--- a/prosper/docs/STRAY_STATUS.md
+++ b/prosper/docs/STRAY_STATUS.md
@@ -293,10 +293,31 @@ drops still discard the background.
   `dd`/`od` read of `hk_project-ps5.pak` outside the emulator agrees exactly with the in-memory
   read-back (25 of 64 non-zero dwords at texture offset 0). And the range is nonetheless **entirely
   zero when the shader samples it**: base `0x303cd10000`, BC1_SRGB 3840×2160, read by
-  `0x300c150000`, `nonzero=0/512` at **+65.7 s** after its verified write. A second base,
-  `0x302a300000`, is zero on five samples across 152 s, the earliest only **2.0 s** after its write.
-  So this is a **lifetime** defect — something reclaims, zeroes or remaps these ranges between load
-  and use — not a read defect, and the pak reader is not the place to look. #3142.
+  `0x300c150000`, `nonzero=0/512` at +65.7 s after its verified write.
+  So this is a **lifetime** defect — something zeroes these ranges between load and use — not a read
+  defect, and the pak reader is not the place to look. #3142.
+
+- **"The range is zeroed some time during the ~65 s before the shader reads it."** Falsified, and the
+  65 s was an artifact of when the *shader* happened to look rather than of when the data went away.
+  `PROSPER_ZEROWATCH` polls each armed destination every 50 ms and reports the transition itself: the
+  data is gone **0.5–2.3 s** after its load, and the defect therefore **reproduces in a 2-minute run**
+  rather than the 7-minute title route — it happens during asset load, long before the title screen.
+  At the instant of the flip the mapping is *unchanged*: still `rw-s` on the dmem memfd at the same
+  file offset, so it is neither unmapped nor re-pointed. And because that mapping is `MAP_SHARED` on a
+  memfd, reading the mapping **is** reading the file — there is no "stale view" case, so the content
+  itself was zeroed. #3142, #3145.
+  - Four mechanisms measured and excluded, each in the same runs: **`dmem_zero`'s hole punch** (all 20
+    calls occur at startup, before every one of these writes); **a re-map of the VA** (one `[physmap]`
+    per VA, created ~2 ms *before* its write, none after); **a second VA aliasing the phys** (none —
+    each phys range is mapped exactly once); and **the renderer writing back**
+    (`guest_memory_gpu_write` fires once per run, covering none of the zeroed ranges).
+  - What blocks naming the writer, so nobody re-derives it: `PROSPER_DMEM_WRITE_TRACE` records writer
+    RIPs but is invalidated by a **host** write, and the APR load *is* a host write (all four host
+    writers of guest memory in the tree are file reads), so it can never watch a range APR loads into.
+    It reports `page-faults=0` — no guest write faulted — but the watch died too early for that to be
+    conclusive. A `SIGSEGV` trap hand-rolled to get the RIP directly killed the run at asset load and
+    was removed rather than debugged; extending the existing trace's lifecycle is the route, not a
+    second write-watch fighting the first.
   - Two traps came out of it and are recorded in `GAME_COMPAT_ORCHESTRATION.md`: **235**, log line
     order is not happens-before (the ordering above is a shared monotonic stamp, not line distance);
     **236**, a verifier that compares populations instead of content reports MATCH on entirely
--- a/prosper/docs/STRAY_STATUS.md
+++ b/prosper/docs/STRAY_STATUS.md
@@ -310,7 +310,14 @@ drops still discard the background.
     calls occur at startup, before every one of these writes); **a re-map of the VA** (one `[physmap]`
     per VA, created ~2 ms *before* its write, none after); **a second VA aliasing the phys** (none —
     each phys range is mapped exactly once); and **the renderer writing back**
-    (`guest_memory_gpu_write` fires once per run, covering none of the zeroed ranges).
+    (`guest_memory_gpu_write` fires once per run, covering none of the zeroed ranges). A fifth:
+    **the address being recycled for another asset** — `0x303cd10000` receives exactly one payload
+    and is never written again, though 956 of 6,574 destinations in the run genuinely are reused, so
+    this had to be checked per-address rather than assumed either way.
+  - Corroborated on a second base: `0x302a300000` is sampled by five different shaders across
+    **152 s** and reads zero every time, the earliest only 2.0 s after its byte-verified write. One
+    address going quiet could be a descriptor pointing somewhere stale; two, with one of them read by
+    five separate shaders, is the range.
   - What blocks naming the writer, so nobody re-derives it: `PROSPER_DMEM_WRITE_TRACE` records writer
     RIPs but is invalidated by a **host** write, and the APR load *is* a host write (all four host
     writers of guest memory in the tree are file reads), so it can never watch a range APR loads into.
--- a/prosper/docs/STRAY_STATUS.md
+++ b/prosper/docs/STRAY_STATUS.md
@@ -327,13 +327,18 @@ drops still discard the background.
     **152 s** and reads zero every time, the earliest only 2.0 s after its byte-verified write. One
     address going quiet could be a descriptor pointing somewhere stale; two, with one of them read by
     five separate shaders, is the range.
-  - What blocks naming the writer, so nobody re-derives it: `PROSPER_DMEM_WRITE_TRACE` records writer
-    RIPs but is invalidated by a **host** write, and the APR load *is* a host write (all four host
-    writers of guest memory in the tree are file reads), so it can never watch a range APR loads into.
-    It reports `page-faults=0` — no guest write faulted — but the watch died too early for that to be
-    conclusive. A `SIGSEGV` trap hand-rolled to get the RIP directly killed the run at asset load and
-    was removed rather than debugged; extending the existing trace's lifecycle is the route, not a
-    second write-watch fighting the first.
+  - What *used* to block naming the writer, kept because it explains why this took so long and what
+    the attribution above depends on: `PROSPER_DMEM_WRITE_TRACE` records writer RIPs but was
+    invalidated by a **host** write, and the APR load *is* a host write (all four host writers of
+    guest memory in the tree are file reads), so it could never watch a range APR loads into — it
+    reported `page-faults=0` because the watch died at the load, not because no guest write happened.
+    **[#3147](https://github.com/mattias800/prosper/pull/3147) lifts that** with an opt-in
+    re-baseline mode, and the `libc.prx` attribution above is measured **with that PR applied**: on
+    master alone the trace still disarms at the load and reproduces nothing. If #3147 is merged when
+    you read this, drop this caveat; if it was closed, the attribution needs re-deriving.
+    A `SIGSEGV` trap hand-rolled to get the RIP directly killed the run at asset load and was removed
+    rather than debugged — extending the existing trace's lifecycle was the route, not a second
+    write-watch fighting the first.
   - Two traps came out of it and are recorded in `GAME_COMPAT_ORCHESTRATION.md`: **235**, log line
     order is not happens-before (the ordering above is a shared monotonic stamp, not line distance);
     **236**, a verifier that compares populations instead of content reports MATCH on entirely
--- a/prosper/src/hle/fs/hle_file.cpp
+++ b/prosper/src/hle/fs/hle_file.cpp
@@ -2701,6 +2701,10 @@ extern "C" int prosper_apr_dest_scan(uint64_t lo, uint64_t hi, char* out, size_t
 namespace {
 struct ZeroWatchEntry {
     uint64_t dst = 0, size = 0, t_armed = 0;
+    // The offset of the window the SOURCE was found non-zero at. The probe must read this window and
+    // no other: seeding from one window and probing a different one compares two unrelated places,
+    // and reports a flip for a range that never changed.
+    uint64_t seed_off = 0;
     const char* path = "?";
     bool seen_nonzero = false, reported = false;
 };
@@ -2745,7 +2749,7 @@ void zerowatch_poll_forever() {
             for (auto& e : g_zw) {
                 if (e.reported) continue;
                 uint32_t w[64];
-                if (!zw_read256(e.dst, w)) {
+                if (!zw_read256(e.dst + e.seed_off, w)) {
                     e.reported = true;
                     fprintf(stderr,
                             "[zerowatch] t=%llu dst=0x%llx WENT-UNMAPPED after %.3fs  maps=%s\n",
@@ -2758,20 +2762,21 @@ void zerowatch_poll_forever() {
                 uint32_t nz = 0;
                 for (uint32_t k = 0; k < 64; ++k) if (w[k]) ++nz;
                 if (nz) { e.seen_nonzero = true; continue; }
-                // All-zero in the sampled window. Only interesting if this range is KNOWN to have
-                // held data -- otherwise it is a buffer that was never filled, which is not this
-                // defect. That knowledge is seeded at arm time from the source buffer, never from
-                // this window: an all-zero first 256 bytes is common on this data (the source is
-                // ~5.6% non-zero and padding is always zero), so deciding it here would make a
-                // payload with a zero-padded head unreportable -- silently, and exactly for the
-                // class under investigation.
-                if (e.seen_nonzero) {
+                // All-zero at the window the payload was non-zero at, so this range HELD data and
+                // no longer does. `seen_nonzero` is guaranteed true for every entry (arming returns
+                // early otherwise), which is why the probe reading `seed_off` rather than offset 0
+                // is what carries the correctness here: an all-zero first 256 bytes is common on
+                // this data (the source is ~5.6% non-zero and padding is always zero), so probing
+                // offset 0 while seeding elsewhere would report a flip for a range that never
+                // changed -- and would burn the one-shot write-trace report on a non-event.
+                {
                     e.reported = true;
                     fprintf(stderr,
-                            "[zerowatch] t=%llu dst=0x%llx size=%llu sampled=256B@0 armed-via=%s "
+                            "[zerowatch] t=%llu dst=0x%llx size=%llu sampled=256B@0x%llx armed-via=%s "
                             "WENT-ZERO after %.3fs  maps=%s\n",
                             (unsigned long long)prosper::diagnostics::diag_now_us(),
-                            (unsigned long long)e.dst, (unsigned long long)e.size, e.path,
+                            (unsigned long long)e.dst, (unsigned long long)e.size,
+                            (unsigned long long)e.seed_off, e.path,
                             (double)(prosper::diagnostics::diag_now_us() - e.t_armed) / 1e6,
                             zw_mapping_of(e.dst).c_str());
                     // Dump the dmem write trace HERE, at the moment the data disappears. Its own
@@ -2818,11 +2823,14 @@ void zerowatch_arm(uint64_t dst, const void* src, uint64_t size, const char* pat
     // Seed from the SOURCE (see the poll loop): scan a spread of the payload, not just its head.
     const uint8_t* p8 = static_cast<const uint8_t*>(src);
     for (uint32_t w = 0; w < 8 && !e.seen_nonzero; ++w) {
-        const uint64_t off = ((size - 256ull) / 7ull) * w;
+        const uint64_t off = (((size - 256ull) / 7ull) * w) & ~(uint64_t)3;
         for (uint32_t k = 0; k < 256; ++k)
-            if (p8[off + k]) { e.seen_nonzero = true; break; }
+            if (p8[off + k]) { e.seen_nonzero = true; e.seed_off = off; break; }
     }
-    if (!e.seen_nonzero) return;   // the payload itself is all zeros; nothing to lose here
+    // An all-zero payload is skipped, and skipping it is NOT silent-by-omission here: it means the
+    // load itself delivered zeros, which `apr_verify_write` reports on the same write. There is no
+    // disappearance to watch for.
+    if (!e.seen_nonzero) return;
     g_zw.push_back(e);
     if (!g_zw_thread.exchange(true)) std::thread(zerowatch_poll_forever).detach();
 }
--- a/prosper/src/hle/fs/hle_file.cpp
+++ b/prosper/src/hle/fs/hle_file.cpp
@@ -2701,6 +2701,7 @@ extern "C" int prosper_apr_dest_scan(uint64_t lo, uint64_t hi, char* out, size_t
 namespace {
 struct ZeroWatchEntry {
     uint64_t dst = 0, size = 0, t_armed = 0;
+    const char* path = "?";
     bool seen_nonzero = false, reported = false;
 };
 std::mutex g_zw_mx;
@@ -2732,6 +2733,11 @@ bool zw_read256(uint64_t addr, uint32_t* out) {
     return process_vm_readv(getpid(), &l, 1, &r, 1, 0) == 256;
 }
 
+// Runs until the process ends, and is never joined. That is safe here only because every exit path
+// in this project is _Exit -- prosper-app, tools/screenshot, and the guest's own _exit and fatal
+// raise -- so no static this thread touches is ever destroyed under it. A normal `return` from main
+// would make this a use-after-free, so if an exit path ever stops being _Exit, this thread needs a
+// stop flag before that lands.
 void zerowatch_poll_forever() {
     for (;;) {
         {
@@ -2752,14 +2758,20 @@ void zerowatch_poll_forever() {
                 uint32_t nz = 0;
                 for (uint32_t k = 0; k < 64; ++k) if (w[k]) ++nz;
                 if (nz) { e.seen_nonzero = true; continue; }
-                // All-zero. Only interesting if this range was observed holding data first --
-                // otherwise it is a buffer that was never filled, which is not this defect.
+                // All-zero in the sampled window. Only interesting if this range is KNOWN to have
+                // held data -- otherwise it is a buffer that was never filled, which is not this
+                // defect. That knowledge is seeded at arm time from the source buffer, never from
+                // this window: an all-zero first 256 bytes is common on this data (the source is
+                // ~5.6% non-zero and padding is always zero), so deciding it here would make a
+                // payload with a zero-padded head unreportable -- silently, and exactly for the
+                // class under investigation.
                 if (e.seen_nonzero) {
                     e.reported = true;
                     fprintf(stderr,
-                            "[zerowatch] t=%llu dst=0x%llx size=%llu WENT-ZERO after %.3fs  maps=%s\n",
+                            "[zerowatch] t=%llu dst=0x%llx size=%llu sampled=256B@0 armed-via=%s "
+                            "WENT-ZERO after %.3fs  maps=%s\n",
                             (unsigned long long)prosper::diagnostics::diag_now_us(),
-                            (unsigned long long)e.dst, (unsigned long long)e.size,
+                            (unsigned long long)e.dst, (unsigned long long)e.size, e.path,
                             (double)(prosper::diagnostics::diag_now_us() - e.t_armed) / 1e6,
                             zw_mapping_of(e.dst).c_str());
                     // Dump the dmem write trace HERE, at the moment the data disappears. Its own
@@ -2781,13 +2793,36 @@ void zerowatch_poll_forever() {
     }
 }
 
-void zerowatch_arm(uint64_t dst, uint64_t size) {
+// `src` is the buffer just written, so "did this range ever hold data" is answered from the payload
+// rather than from a later poll of the destination. `path` names which write path armed it, because
+// the two are not equivalent: the retry path MAP_FIXEDs anonymous memory over guest pages first, and
+// a range that went through it is the most interesting kind to watch, not one to skip.
+void zerowatch_arm(uint64_t dst, const void* src, uint64_t size, const char* path) {
     if (!zerowatch_enabled() || size < (1u << 20)) return;   // large payloads only: textures, not records
     std::lock_guard<std::mutex> lk(g_zw_mx);
     for (auto& e : g_zw) if (e.dst == dst) return;
-    if (g_zw.size() >= 64) return;
+    if (g_zw.size() >= 64) {
+        // Never drop silently: a capped watch list turns "nothing else went zero" into "nothing else
+        // was looked at", and those read identically in a log.
+        static bool warned = false;
+        if (!warned) {
+            warned = true;
+            fprintf(stderr, "[zerowatch] watch list FULL at %zu entries -- later destinations are "
+                            "NOT watched; counts below are a floor, not a total\n", g_zw.size());
+        }
+        return;
+    }
     ZeroWatchEntry e;
     e.dst = dst; e.size = size; e.t_armed = prosper::diagnostics::diag_now_us();
+    e.path = path;
+    // Seed from the SOURCE (see the poll loop): scan a spread of the payload, not just its head.
+    const uint8_t* p8 = static_cast<const uint8_t*>(src);
+    for (uint32_t w = 0; w < 8 && !e.seen_nonzero; ++w) {
+        const uint64_t off = ((size - 256ull) / 7ull) * w;
+        for (uint32_t k = 0; k < 256; ++k)
+            if (p8[off + k]) { e.seen_nonzero = true; break; }
+    }
+    if (!e.seen_nonzero) return;   // the payload itself is all zeros; nothing to lose here
     g_zw.push_back(e);
     if (!g_zw_thread.exchange(true)) std::thread(zerowatch_poll_forever).detach();
 }
@@ -2877,7 +2912,7 @@ static bool apr_write_guest_dst(uint64_t dst, void* buf, uint64_t size) {
     struct iovec l { buf, (size_t)size }, r { (void*)(uintptr_t)dst, (size_t)size };
     if (process_vm_writev(getpid(), &l, 1, &r, 1, 0) == (ssize_t)size) {
         apr_verify_write(dst, buf, size);
-        zerowatch_arm(dst, size);
+        zerowatch_arm(dst, buf, size, "direct");
         return true;
     }
     bool committed = false;
@@ -2891,7 +2926,12 @@ static bool apr_write_guest_dst(uint64_t dst, void* buf, uint64_t size) {
     }
     if (!committed) return false;
     const bool retried = process_vm_writev(getpid(), &l, 1, &r, 1, 0) == (ssize_t)size;
-    if (retried) apr_verify_write(dst, buf, size);
+    if (retried) {
+        apr_verify_write(dst, buf, size);
+        // The retry path MAP_FIXEDs fresh anonymous pages over the destination first, which makes
+        // it the MOST interesting class to watch, not one to leave uncovered as it was.
+        zerowatch_arm(dst, buf, size, "commit-retry");
+    }
     return retried;
 }
 #endif
--- a/prosper/src/hle/fs/hle_file.cpp
+++ b/prosper/src/hle/fs/hle_file.cpp
@@ -2685,6 +2685,102 @@ extern "C" int prosper_apr_dest_scan(uint64_t lo, uint64_t hi, char* out, size_t
 //
 // PROSPER_APR_VERIFY=1 verifies every write; PROSPER_APR_VERIFY=0xADDR verifies only that
 // destination, which is what keeps the log readable on a title that streams thousands of chunks.
+// DIAGNOSTIC (PROSPER_ZEROWATCH=1): catch the MOMENT guest data disappears, and say what the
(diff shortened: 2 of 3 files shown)


mattias800 (commented):
**REJECTED** on the exact head `9f48eb6d` — three blocking findings, all cheap. The two questions you
asked for a second opinion on both come out **clean**, and I verified the `map_phys_at` chokepoint
claim independently; details below so you get the benefit of that work regardless.

Reviewed by reading, not rebuilding: I did not duplicate your build/ctest. I did re-run
`python3 prosper/tools/env/check_diag_gates.py` at your head — `== all checks passed ==`, rc=0,
570 files, 733 vars at 1165 sites, findings 108/108 baselined, so the two new sites introduce no
unbaselined finding. `git diff --check` is clean.

---

## Answers to the two questions you asked

**Lifetime (`zerowatch_poll_forever` vs `g_zw`) — safe, but not for the reason in the PR body.**
The in-run access is safe because of `g_zw_mx`, *not* because the vector is append-only. Every read
in the poll loop and the `push_back` in `zerowatch_arm` are both under the lock
(`prosper/src/hle/fs/hle_file.cpp:2738`, `:2786` / `:2791`), so no iterator or reference outlives a
reallocation. The append-only property is doing no work here, and stating it as the argument is the
hazard: it reads as a licence to drop the lock later, at which point `push_back` reallocating under a
live `auto& e` is an immediate use-after-free. Please say "the mutex" in the comment, not
"append-only".

The part the argument does not cover is process shutdown — a detached thread iterating `g_zw` while
static destruction runs `~vector` and `~mutex`. I checked whether that is reachable, and **it is
not**: every exit path in the tree is `_Exit`, which skips static destructors —
`prosper/frontends/prosper-app/main.cpp:2905`, `prosper/tools/screenshot/screenshot.cpp:968`, and the
guest `_exit` / fatal-raise HLEs at `prosper/src/hle/kernel/hle_kernel_time.cpp:838` and `:848`. So
the lifetime argument holds today by an invariant that lives in four other files. That is worth one
comment line beside the `detach()`, because a future frontend that returns from `main()` turns this
into a shutdown crash that will read as an emulator defect rather than an instrument defect. (See N5
— the same `_Exit` invariant already makes an existing `atexit` registration dead.)

**`guest_dmem_write_trace_report()` from the poll thread — no deadlock.** The lock order is
`g_zw_mx -> WatchState::mutex -> stderr`. Nothing in the tree takes `g_zw_mx` other than
`zerowatch_arm`, and its caller `apr_write_guest_dst` has already returned from
`guest_write_watch_notify_host_write` (and so released `WatchState::mutex`) before it calls in
(`hle_file.cpp:2876` then `:2880`), so there is no reverse edge and no cycle. The SIGSEGV write-watch
handler takes `WatchState::mutex` *before* its own `fprintf`, i.e. the same relative order, so the
handler can only be delayed, never deadlocked. Calling it mid-run is also state-safe: the Linux
implementation at `prosper/src/host/memory/guest_write_watch.cpp:1461` only reads under the lock and
sets `report_emitted`.

---

## Blocking

**B1. `WENT-ZERO` asserts a property the probe never tested, and the `seen_nonzero` gate can hide the
case under investigation.** `zw_read256` samples the **first 256 bytes** of the range
(`hle_file.cpp:2730`), and the report prints `dst=… size=<full size> … WENT-ZERO`
(`hle_file.cpp:2757-2764`). A reader — and the PR body, which says it "reports the non-zero →
all-zero transition" of "each large APR destination" — takes that as a statement about the range. It
is a statement about 256 bytes. This is the shape of instrument trap 236, raised on #3144 in this
same file eight days' worth of investigation ago: an instrument reporting a property stronger than
the one it measured.

The gate is the sharper half. `seen_nonzero` is set only from that same 256-byte window
(`hle_file.cpp:2754`), so a payload whose first 256 bytes are legitimately zero can **never** report,
silently. That this is common on exactly this data is documented 60 lines below, in
`apr_verify_write`'s own selftest comment (`hle_file.cpp:2811-2814`): *"the source is ~5.6% non-zero,
so an all-zero first window is common and padding is always one."* So the census of flips is censored
by a criterion the neighbouring code already says fires often here, and a range that never reports is
indistinguishable from a range with no defect.

Both are a few lines: set `seen_nonzero` at arm time from the **source buffer** the caller already
holds (`buf` is in hand at `hle_file.cpp:2880`, and #3144 established it lands byte-identically), and
either probe the 8 spread windows `apr_verify_write` already uses or label the log line with what was
actually sampled (`probe=256`). Fixing the arm-time seed also closes the write-then-zero-inside-one-
50 ms-poll miss, which is otherwise silent for the same reason.

**B2. Two silent coverage gaps mean "11–13 times per run" has no denominator.**
(a) `if (g_zw.size() >= 64) return;` (`hle_file.cpp:2788`) drops arming with no output, and entries
are never recycled once `reported`, so the budget is spent once. `STRAY_STATUS.md` records 21,499 APR
writes on the title route; nothing in the log says whether arming stopped at 64, so 11–13 may be a
property of the instrument rather than of the title.
(b) `zerowatch_arm` is called only on the fast `process_vm_writev` path (`hle_file.cpp:2880`). The
fallback/retry path at the end of the same function calls `apr_verify_write` but **not**
`zerowatch_arm` (`hle_file.cpp:2894`) — and that path `MAP_FIXED`s `MAP_PRIVATE|MAP_ANONYMOUS` over
guest pages, which detaches them from the dmem memfd. That is itself a live candidate mechanism for
#3142 and is precisely what your `/proc/self/maps` line would identify (`rw-p` anonymous instead of
`rw-s` on the memfd) — so the one destination class where the maps dump would be most decisive is the
one class the watch never arms.

Fix: one `fprintf` when the cap is hit plus an armed counter in the report, and arm on the retry path
too. Alternatively drop the per-run rate from the doc entry — but then B3's entry cannot quote it.

**B3. Four falsifications and a corrected timing, none of them in `## Ruled out`.** The charter is
explicit: *"A PR that falsifies a hypothesis records it in the relevant `## Ruled out` section before
merging… A falsification that lives only in a PR body or an issue comment is one the next agent will
re-derive at full cost."* This PR touches zero `.md` files while asserting four falsifications in its
body. `prosper/docs/STRAY_STATUS.md` already has the section (line 286) and already carries the
previous round's rows, so the destination is unambiguous.

The timing correction is the more urgent half. That doc's `## Ruled out` currently frames the defect
around `+65.7 s` after the verified write, and the doc opens with its own *"READ THIS BEFORE QUOTING A
NUMBER FROM THIS DOC"* header. Your measurement says the flip is 0.5–2.3 s and 65 s was only the delay
until the *shader* sampled. Left as is, the stale figure keeps its authority in the document people
read first — the exact propagation failure the charter records under the retracted-2.4% example.
Please also note in the same edit that the 2-minute route reproduces it, since that changes which
route the next lane runs.

---

## Non-blocking

**N1. `[physmap]` is the chokepoint for alias *creation*, not for alias *destruction* — so
hypothesis 2 is falsified by a different instrument than the table implies.** Your chokepoint claim
verifies, and I checked it rather than taking the file comment: `map_phys_at_impl` has exactly one
caller (`hle_kernel_mem.cpp:1528`), and on Linux `dmem_fd()` is a file-static reached from only five
places in `hle_kernel_mem.cpp` — `dmem_zero`'s `fallocate`, `map_phys_at_impl`'s `mmap`,
`guest_memory_gpu_write`'s support check and its `pwrite`. So on Linux every VA→phys alias and every
phys-level write is covered by your two log lines, which is a stronger completeness argument than the
PR body makes. But a VA can also stop aliasing a phys via `k_munmap` (`hle_kernel_mem.cpp:2053`) or an
anonymous `MAP_FIXED` overlay (`:1407`, `:2276`), and `[physmap]` sees none of those. Your conclusion
survives — the unchanged `rw-s` maps line at the moment of the flip is what excludes them — but the
table should attribute that row to ZEROWATCH, not to PHYSLOG.

**N2. A `[physmap]` line does not prove the VA aliases the phys.** `map_phys_at_impl` falls back to
`map_at` and to anonymous `MAP_PRIVATE` (`hle_kernel_mem.cpp:1513` and `:1516`) and `map_phys_at` logs the
requested `phys` regardless. Harmless for your "zero aliasing mappings" conclusion (the error is in
the direction of over-reporting aliases), but a reader counting aliases from the log should know.

**N3. The instrument silently consumes another instrument's one-shot output.**
`guest_dmem_write_trace_report` is `report_emitted`-guarded, so only the **first** of the 11–13
`WENT-ZERO` events prints a trace, and enabling `PROSPER_ZEROWATCH` suppresses the report the guest
`_exit` path would otherwise emit (`hle_kernel_time.cpp:836`). Worth one sentence in the comment.

**N4. Holding `g_zw_mx` across the `fopen`/`fgets` of `/proc/self/maps` and across
`guest_dmem_write_trace_report()` (which blocks on `WatchState::mutex`, contended by the fault
handler) stalls `zerowatch_arm` on the APR load path** — the exact path under study, in a defect whose
timing is the finding. Bounded, but the two-phase fix is trivial: collect the entries to report under
the lock, release, then print.

**N5. Found while checking the `_Exit` invariant: `std::atexit(guest_dmem_write_trace_report)`
(`guest_write_watch.cpp:1423`) can never fire** — `prosper-app`, `tools/screenshot` and both guest
termination HLEs all use `_Exit`. Your PR body says the trace "only prints on guest `_exit` and the
fatal raise", which is right in effect; the third registration is dead code. Worth an issue.

**N6. `char line[512]` in `zw_mapping_of` (`hle_file.cpp:2716`) can truncate a real maps line.**
`PATH_MAX` is 4096 and prosper maps modules from long dump paths, so a truncated match loses the path —
the field the dump exists to show — and the untruncated remainder is then fed to `sscanf` as if it
were a fresh line. Not a risk for the memfd line you actually got, but 4096 plus a drain-to-newline
costs nothing.

**N7. `PROSPER_ZEROWATCH=0` and `PROSPER_PHYSLOG=0` enable the diagnostics.** Both gates are presence
tests (`hle_file.cpp:2710`, `hle_kernel_mem.cpp:1445`, `:3290`). The neighbouring instruments in these
same files check `!ev || !*ev` (`hle_file.cpp:2798`, `guest_write_watch.cpp:1714`). Cheap to match.

**N8. Two cached `getenv`s for one variable.** `physlog()` (`hle_kernel_mem.cpp:1445`) is inside the
POSIX anonymous namespace and `guest_memory_gpu_write` re-reads it as a local static
(`hle_kernel_mem.cpp:3290`). Within the gate checker's budget, but one accessor is better.

**N9. Windows arm untouched, silently.** The instrumented `dmem_zero`, `map_phys_at` and
`guest_memory_gpu_write` are all inside `#if defined(__linux__) || defined(__APPLE__)`
(`hle_kernel_mem.cpp:627`); the Windows siblings at `:4920` and `:7151` are not instrumented. Correct
for a Linux investigation, and I confirmed the Linux `dmem_zero` at `:1447` covers both the Linux
`fallocate` and the Darwin scratch-mmap arms — but on Windows `PROSPER_PHYSLOG` is accepted and prints
nothing at all. One line in the comment.

**N10. Neither variable is documented.** `prosper/tools/AGENTS.md` is the catalogue for `PROSPER_*`
diagnostics and gains no entry. (`PROSPER_HOSTWRITEWATCH` is not in there either, so this is
convention rather than a rule — but these two are worth the four lines, especially the ZEROWATCH
limitation about ranges APR loads into, which is the kind of thing the next lane will otherwise
rediscover.)

**N11. `[dmemzero]` logs an attempted punch, not a completed one** — the `fallocate` return is not
checked (`hle_kernel_mem.cpp:1455`), and `FALLOC_FL_PUNCH_HOLE` rejects unaligned ranges. Immaterial
to your "all 20 calls are at startup" falsification, which only needs the ordering, but it would
matter if anyone later reads the absence of a `[dmemzero]` line as proof a range was not zeroed.

**N12. Duplicate include.** `host/memory/guest_write_watch.hpp` is added at `hle_file.cpp:11` and was
already present at `:16`.

---

Everything I checked that you claimed, held: the default-off claim (each site is one function-local
static guard load; no mutex is taken, no thread is started, and `zerowatch_arm` returns before the
size test), the `map_phys_at` chokepoint, the Linux `dmem_zero` coverage, and the four-call-site
`host-write` invalidation (the four `guest_write_watch_notify_host_write` callers are all in
`hle_file.cpp`, all file reads). B1 and B2 are about what the instrument can be trusted to say when
it says nothing; B3 is about the finding surviving this PR. Re-request review at the new head and
I will look only at the deltas.

mattias800 (commented):
**REJECTED** on the exact head `3008860c`, round 2 — **one** blocking code finding and one blocking
claim, both narrow. B2 and B3 from round 1 are **fixed and I checked them**; B1's fix closed the
blind spot I raised and opened a different one in the same gate, which is the only reason this is not
an approval.

Reviewed by reading the deltas, as promised. I did not duplicate your build/ctest. Re-run at your
head: `python3 prosper/tools/env/check_diag_gates.py` → `== all checks passed ==`, rc=0, 570 files,
733 vars at 1165 sites, findings 108/108 baselined — **identical to `9f48eb6d`**, so this round adds
no new env site. `git ls-files '*.md' -z | xargs -0 python3 prosper/tools/docs/check_numbered_table.py`
→ rc=0, 155 files, `prosper/docs/STRAY_STATUS.md: 12 table(s), 39 rows, unbroken`.
`git diff --check origin/master...HEAD` clean.

---

## What is fixed, verified

**B2 — complete.** Both success paths of the POSIX `apr_write_guest_dst` arm
(`prosper/src/hle/fs/hle_file.cpp:2915` `"direct"`, `:2933` `"commit-retry"`), and those are the only
two returns that can report success, so the three callers at `:3028`, `:3500`, `:3517` are all
covered through one function. The cap warning cannot spam — `static bool warned` at `:2807` is set
under `g_zw_mx`, which `zerowatch_arm` holds from `:2802`, so it is both once-only and race-free.
`armed-via=` is accurate on each path, and `path` lifetime is fine: both arguments are string
literals (static storage), the default `"?"` at `:2704` is unreachable, and nothing copies or frees
it. Ordering it before the all-zero skip is also right — a skipped payload does not spend a slot.

**B3 — the doc row lands, and its measurements check out.** I traced every number in
`prosper/docs/STRAY_STATUS.md:300-323` against #3142's own comment thread rather than against the PR
body. **0.5–2.3 s** is supported (`WENT-ZERO after 0.649s / 0.599s / 2.291s` in the 16:02Z comment);
**"reproduces in a 2-minute run"** is supported — the same comment states it as a measured practical
note; **20 `dmem_zero` calls at startup**, **one `[physmap]` per VA ~2 ms before its write**, **zero
aliasing mappings** and **`guest_memory_gpu_write` once per run** all match the table in that comment.
The tightening from *"reclaims, zeroes or remaps"* to *"zeroes"* is earned by the new maps line, and
the stale `+65.7 s` is now de-bolded and corrected in the very next bullet rather than left standing.
Round 1's N1 is also answered without my having asked: the destruction case that `[physmap]` cannot
see is excluded by the unchanged `rw-s` maps line, and the doc states both halves.

The `_Exit` comment at `:2736-2740` and the "because of the mutex" correction both landed.

---

## Blocking

**B1. The B1 fix makes the poll-loop gate unconditionally true, so `WENT-ZERO` no longer requires a
transition at all.** This is the one real finding.

`zerowatch_arm` now returns before `push_back` when the seed fails (`hle_file.cpp:2825`, then `:2826`),
and nothing anywhere clears `seen_nonzero`. Therefore **every entry that exists in `g_zw` has
`seen_nonzero == true` by construction**, and the guard at `:2768` can never be false. The gate is
dead code. What the poll now does is: read `dst[0..256)` (`zw_read256` at `:2731-2734` — always
offset 0), and if those 256 bytes are zero, print `WENT-ZERO`.

So the seed and the probe read **different regions**. The seed scans eight spread windows of the
*source* (`:2820-2824`); the probe reads window 0 of the *destination*. A payload whose first 256
bytes are legitimately zero but which has data at, say, window 3 is armed, and the **first poll**
prints `WENT-ZERO after 0.05s` for a range that never changed. That input class is not hypothetical
here — `apr_verify_write`'s own comment sixty lines below (`:2846-2849`) says an all-zero first window
is common on exactly this data, and it is the reason your own seed had to move off that window.

Two consequences, and the second is why I am blocking rather than filing this as N:

* The log line still says `WENT-ZERO`, which asserts a transition the instrument no longer tests for.
  `sampled=256B@0` discloses the probe window but not that the *verdict's other half* came from
  somewhere else, so a reader takes it as "this window held data and stopped".
* `host::guest_dmem_write_trace_report()` at `:2787` is **one-shot** — `report_emitted` at
  `prosper/src/host/memory/guest_write_watch.cpp:1465-1466` returns early on every later call. A false
  first `WENT-ZERO` therefore burns the write trace at a non-event, and the real flip 0.6 s later gets
  nothing. That is the instrument losing its ability to show its own lever moved, which is the bar you
  and I agreed to reject on.

The fix is smaller than the one you already made: record the offset of the window that seeded, and
probe *that* window.

```
uint64_t probe_off = 0;                      // in ZeroWatchEntry
...
if (p8[off + k]) { e.seen_nonzero = true; e.probe_off = off; break; }
...
zw_read256(e.dst + e.probe_off, w)           // in the poll loop
"... sampled=256B@%llu ..."                  // and print e.probe_off
```

That keeps the blind spot closed (the probe follows the data instead of the head), restores the
gate's meaning, and makes the line's basis literally true. Note it also changes what the log looks
like, so the census below has to be re-taken either way.

**B2. "the 64-entry cap is reached on the very first run" is not measured anywhere I can find.** The
PR body now demotes 11–13 to a floor on that basis. But the mechanism that can observe the cap — the
warning at `hle_file.cpp:2804-2814` — landed in *this* commit, and the last live evidence in the
record is #3142's 16:02:51Z comment, which predates round 1. The body's Verification section still
reads "Live: five Stray boots", i.e. the round-1 runs. And a pre-fix inference would not transfer in
any case, because the same commit added the all-zero skip at `:2825`, which *reduces* how many
destinations arm — the cap could plausibly not be reached now.

Separately, "floor" is only half true after B1: with the gate dead, the count can be inflated by
false positives as well as depressed by the cap, so 11–13 is currently bounded on neither side. Since
B1 forces a re-run anyway, this costs nothing: post the run (a `[zerowatch] watch list FULL` line, or
its absence) and the sentence becomes a measurement. If you would rather not re-run, say "the cap may
be reached and the probe now says so" and drop the claim that it was.

Neither number reaches `STRAY_STATUS.md`, so B3's row is not contaminated by this.

---

## Non-blocking

**N1. `// the payload itself is all zeros` (`hle_file.cpp:2825`) asserts more than 2048 sampled bytes
of a ≥1 MiB payload can support** — the same shape as the finding it was written to fix. "sampled
all-zero" is the honest word. Related: this skip is now the one silent drop left in `zerowatch_arm`,
sitting directly below a cap drop you deliberately made loud for exactly that reason. A counter
printed alongside the FULL warning ("N destinations skipped as sampled-all-zero") would close it.

**N2. The bounds arithmetic in the seed loop is correct, and tight.** With the guard at `:2801`,
`size ≥ 1 MiB`, and `((size - 256) / 7) * 7 ≤ size - 256`, so the maximum index `off + 255` is exactly
`size - 1` at `size == 1048576`. The buffer is provably that long: `process_vm_writev` read `size`
bytes from `buf` before either arm site. The early exit is right too — the inner `break` leaves the
inner loop and the outer `!e.seen_nonzero` condition stops the scan. No change needed; recording it
so nobody re-derives it.

**N3. The `_Exit` comment (`:2736-2738`) is over-broad as written.** "every exit path in this project
is `_Exit`" is false — the tree defines 271 `test_*.cpp` `main()`s and they return
normally (`prosper/frontends/prosper-app/test_capture_schedule.cpp:13` is one). It is true of every path
that *runs a guest*, which is what actually makes the detached thread safe, since nothing else can arm
it. One qualifying clause, and the comment stops being a claim a future reader might lean on elsewhere.

**N4. The doc edit deletes the `0x302a300000` datum** — *"zero on five samples across 152 s, the
earliest only 2.0 s after its write"*, previously inside the first `## Ruled out` bullet, now
`STRAY_STATUS.md:296-297`. That was the
only corroboration in the doc for seconds-scale timing coming from an instrument **other than**
ZEROWATCH, and the new bullet's headline number now rests on ZEROWATCH alone. Cheap to keep.

**N5. One falsification did not make the trip.** The PR body's table has five rows; the doc's
sub-bullet at `:309-313` has four — "the address is recycled for another asset" (`0x303cc90000`
receives exactly one payload) is dropped. It is a clause, and it is precisely the hypothesis someone
re-derives.

**N6. The 2-minute route loses its "(45 s warmup)" qualifier** between #3142's comment and
`STRAY_STATUS.md:303`. A reproduction detail is the one kind of number worth carrying verbatim.

**N7. Dedupe is by `dst` alone (`:2803`), ignoring size.** A second, larger write to an
already-watched destination is silently ignored. Pre-existing, and probably what you want, but it is
another silent path in a function whose other drops are now loud.

**N8. Round-1 N items still open at this head, all unchanged and all still fine to defer:** the
duplicate `host/memory/guest_write_watch.hpp` include (`hle_file.cpp:11` and `:16`); `char line[512]`
in `zw_mapping_of` (`:2717`); the presence-test gates (`:2711`,
`prosper/src/hle/memory/hle_kernel_mem.cpp:1445`, `:3290`) where `PROSPER_ZEROWATCH=0` enables; the
two cached `getenv`s for one variable; `g_zw_mx` held across `fopen`/`fgets` and the trace report; and
no entry in `prosper/tools/AGENTS.md` for either switch. Landing the AGENTS.md lines with the B1 fix
would be worth more than the rest combined, since the ZEROWATCH sampling limitation is exactly what
the next lane will otherwise rediscover.

---

Re-request review at the new head; I will look only at the `zerowatch_arm`/poll-loop delta and the
body sentence.

mattias800 (commented):
**APPROVED** on the exact head `f7f1c136`, round 3. Both round-2 blockers are fixed and I checked
the arithmetic by hand rather than by reading the comment. Nothing remaining can make the instrument
report a false flip or send a future reader wrong at cost, so everything below is non-blocking.

Reviewed by reading the `3008860c..f7f1c136` delta, as promised; I did not duplicate your build/ctest.
Re-run at your head: `python3 prosper/tools/env/check_diag_gates.py` → `== all checks passed ==`,
rc=0, 570 files, 733 vars at 1165 sites, findings 108/108 baselined — byte-identical to `9f48eb6d`
and `3008860c`, so three rounds of edits have added no env site.
`git ls-files '*.md' -z | xargs -0 python3 prosper/tools/docs/check_numbered_table.py` → rc=0,
155 files, `prosper/docs/STRAY_STATUS.md: 12 table(s), 39 rows, unbroken`. `git diff --check
origin/master...HEAD` clean. Diff touches 3 files; 0 `Claude-Session:` trailers. The doc edit adds no
numbered-table row, so no merge-result re-check is owed. Traps 235 and 236, which
`STRAY_STATUS.md:328-331` cites, both exist (`GAME_COMPAT_ORCHESTRATION.md:363`, `:364`).

---

## B1 — fixed, and the four properties you asked about all hold

1. **`seed_off` is set exactly when `seen_nonzero` is.** Both are assigned in one statement at
   `prosper/src/hle/fs/hle_file.cpp:2828`, and that is the only writer of `seen_nonzero` in
   `zerowatch_arm`; the defaults are `0` / `false` (`:2707`, `:2709`). So there is no state in which
   the gate is true and the offset is stale.
2. **Every read path uses `dst + seed_off`.** There is one read (`:2752`); the WENT-UNMAPPED branch is
   that read's failure arm, so it cannot diverge. See N1 for what the *report* on that branch still
   says.
3. **Bounds hold at the minimum arming size.** With `size >= 1<<20` (`:2806`),
   `floor((size-256)/7)*7 <= size-256`, and the `& ~3` at `:2826` only ever *decreases* the offset, so
   it cannot push the window out — the maximum source index `off+255` is exactly `size-1` at
   `size == 1048576` (`off_max = 1048320`, and 1048320 is already 4-aligned, so alignment is a no-op
   there). The source buffer is provably `size` bytes: `process_vm_writev` read `size` from `buf`
   before either arm site (`:2920-2921` → `:2923`; `:2936` → `:2941`). Window stride
   (~149,760 B at 1 MiB) far exceeds 256, so alignment cannot make two windows overlap either.
4. **`sampled=` matches what was probed.** `:2779` prints `e.seed_off`, the same value added to `e.dst`
   at `:2752`.

And the consequence that made round 2 blocking is closed at the right place: the one-shot
`host::guest_dmem_write_trace_report()` is called only inside the WENT-ZERO branch (`:2792`), which
now requires the *seeded* window to have held data and to read zero — so a false verdict can no
longer burn the trace on a non-event.

## B2 — accepted

The warning exists (`:2809-2818`, once-only under `g_zw_mx`) and you report it fired in the run that
produced the flips. I take an author-reported measurement at face value; see N7/N8 for two small ways
the body still undersells it.

## The "all 14 seeded at `0x0`" question

Honestly stated, and I agree with your framing: with every `seed_off == 0` the probe is bit-identical
to `3008860c`'s, so the fix cannot have moved any number in this PR, and the body says exactly that
instead of dressing it up. It is also not a suspicious zero — `w == 0` is tried first (`:2825-2826`)
and the loop stops on the first hit, so `seed_off != 0` requires an entirely zero first 256 bytes,
which the >=1 MiB texture payloads evidently do not have. The residual is N4: that branch now has no
coverage of any kind.

## The restored `STRAY_STATUS.md` content — checked against #3142

* `0x302a300000` (`STRAY_STATUS.md:317-320`): supported verbatim by #3142's 16:02:51Z-preceding
  comment — `| 0x302a300000 | five distinct shaders | +2.0 s, +2.7 s, +67.2 s, +151.6 s, +152.0 s |`
  and its prose *"sampled five times across 152 seconds by five different shaders, byte-verified
  written, and zero on every single sample."* Earliest +2.0 s ✓, five shaders ✓, 152 s ✓.
* Fifth falsification (`:313-316`): `956 of 6,574` matches the thread's row exactly, and the
  per-address caveat is a fair rendering of *"956 of 6,574 destinations are reused, but not this
  one."* The **address** does not match — see N5.

---

## Non-blocking

**N1. The WENT-UNMAPPED line is the last place that still assumes offset 0.** The read that failed
was at `dst + seed_off`, but the report prints `dst=0x%llx` from `e.dst` and
`maps=zw_mapping_of(e.dst)` (`hle_file.cpp:2754-2759`), and carries no `sampled=` field at all. If a
sub-range overlay ever unmaps the seeded window while `dst` stays mapped, the line asserts
WENT-UNMAPPED beside a live `rw-s` maps line — the same "reports a property of a place it did not
measure" shape you just fixed one branch above. It cannot burn the write trace (that call is
WENT-ZERO-only), and it has not fired in any run you report, which is why this is N and not B. Two
tokens: `zw_mapping_of(e.dst + e.seed_off)` and print the probed address.

**N2. `e.seen_nonzero = true` in the poll loop (`:2764`) is now a dead store** — nothing reads the
field after arming. The comment at `:2766` discloses that the gate is gone, so this is not
misleading today; but a maintained-looking flag is an invitation to reinstate a gate on it. Drop the
store or the field's poll-loop half.

**N3. The comment justifying the fix cites a frequency measured on a different population.**
`:2768-2769` says an all-zero first 256 bytes "is common on this data (the source is ~5.6% non-zero
and padding is always zero)" — which comes from `apr_verify_write`, whose floor is 256 B and which
therefore sees every small record. `zerowatch_arm` arms only at >=1 MiB (`:2806`), and this run says
14/14 seeded at window 0 on that population. The fix is right either way (it is now structural), but
the stated reason is not established for the ranges it protects. One qualifying clause.

**N4. `seed_off != 0` has no coverage, live or automated.** `grep -rn 'zerowatch\|ZEROWATCH'` over
`prosper/tests` and `prosper/tools` finds nothing, and the machinery is in an anonymous namespace
inside `hle_file.cpp`, so it is not reachable from a test even in principle. I verified the branch by
hand above and it is three lines of arithmetic, so I would not hold the PR for it — but the body's
"14 flips, every one seeded at `0x0`" is a clean zero for exactly the class the fix targets, and the
charter's rule is to construct one instance by hand before believing one. A `PROSPER_ZEROWATCH_SEED_MIN`
that starts the window scan at `w=1` would give you that arm in two lines and would also be an honest
A/B if a future run ever disagrees.

**N5. The doc and the PR body now name two different addresses for the same falsification.**
`STRAY_STATUS.md:314` says `0x303cd10000`; the body's table and #3142's 16:02:51Z row both say
`0x303cc90000`. I believe the doc is the correct one — every other citation of that 4,718,737-byte
payload in the thread is `0x303cd10000` (`[apr-verify] dst=0x303cd10000 size=4718737`, the
`[texcontent] base=0x303cd10000` sample, and the `+65.735 s` row), and `STRAY_STATUS.md:295` already
used `0x303cd10000` for the same base. So this reads as you silently repairing a typo in the source
comment. Worth saying so: fix the body table, and drop a one-line note on #3142, or the durable
record and the issue it cites will disagree for whoever cross-checks them.

**N6. The doc states the flip timing without stating what was sampled.** `:302-304` gives
*"the data is gone 0.5–2.3 s after its load"*; the measurement is that one 256-byte window went zero
in that interval. Round 1's B1 required exactly this label on the log line and got it
(`sampled=256B@0x…`); the doc is the copy that outlives the log. The conclusion is independently
corroborated in the same section (`:294-296`, `nonzero=0/512`), so nothing is wrong — it is one clause
("sampled 256 B at the window the payload was written non-zero at") away from being unambiguous.

**N7. The body's own numbers no longer agree with each other.** The headline still reads "at least
11–13 times per run" and "Verification (exact head) … Live: five Stray boots", while the round-3
update reports 14 flips measured at `f7f1c136`. Both are floors and neither reaches the doc, so this
costs nothing but a puzzled reader.

**N8. A ZEROWATCH-only run cannot distinguish an instant flip from a write that never landed.** The
seed is the *source*; if a destination were somehow not written, the first poll would report
`WENT-ZERO after 0.05s`. Your measured 0.5–2.3 s rules that out for these runs, and you co-armed
`PROSPER_APR_VERIFY` anyway — worth one line in the comment so the next lane knows to keep them
paired.

**N9. Round-2 N8's carry-overs are all unchanged and all still fine to defer:** the duplicate
`host/memory/guest_write_watch.hpp` include (`hle_file.cpp:11` and `:16`); `char line[512]` in
`zw_mapping_of` (`:2721`); the presence-test gate at `:2715`, so `PROSPER_ZEROWATCH=0` *enables* it
while `apr_verify_write` two functions below checks `!ev || !*ev` (`:2841`); `g_zw_mx` held across
`fopen`/`fgets` and the trace report; and still no entry in `prosper/tools/AGENTS.md` for either
switch. That last one is now on its third round: the sampling model (256 B at the seeded window,
>=1 MiB only, 64-entry cap) is disclosed in the log line and nowhere a reader looks first.

---

The finding survives this PR in the right file, the two instruments do what the body says they do,
and the correctness of the flip verdict is now structural rather than a property of this title's
payload layout. Merge when CI is green — the `Linux`, `Windows` and `macOS/Rosetta` jobs were still
pending at the time of this review.

mattias800 (commented):
**REJECTED** on `794b0236` — **scope correction first, then one blocking finding, and it is prose
only.**

**Read my previous review as scoped to `f7f1c136`.** You pushed `794b0236` while I was writing it, so
GitHub attached a body that says "on the exact head `f7f1c136`" to a commit it never examined. Every
word of it stands for the code — `f7f1c136`'s `hle_file.cpp` is byte-identical at `794b0236`, so B1
and B2 remain fixed and the four properties I verified by hand are untouched. What follows is a review
of the `f7f1c136..794b0236` delta only: 11 added lines in `prosper/docs/STRAY_STATUS.md`, nothing else.

Gates re-run at `794b0236`: docs gate rc=0, 155 files, `STRAY_STATUS.md: 12 table(s), 39 rows,
unbroken`; `git diff --check origin/master...794b0236` clean; 0 `Claude-Session:` trailers.

The finding itself is excellent and I checked it against #3142's 16:52:39Z comment: 64 of 64 events on
`libc.prx` ✓, 16-byte stride ✓, the `0x3d71`–`0x3dd0` RIP loop ✓, one tid ✓, `result=writer-observed`
✓. Reclassifying #3142 from a memory-lifetime defect to a descriptor-lifetime one is the most valuable
thing in this PR.

---

## Blocking

**B1. The `## Ruled out` list now contains both "the writer is named, here it is" and "naming the
writer is blocked, so nobody re-derives it", twenty lines apart, with no retraction on the second.**
The new bullet at `prosper/docs/STRAY_STATUS.md:309-317` says #3147's re-baseline mode *"observes the
window it previously could not"*. The bullet at `:330-336` still reads *"What blocks naming the
writer, so nobody re-derives it: `PROSPER_DMEM_WRITE_TRACE` … is invalidated by a **host** write, and
the APR load *is* a host write …, so it **can never** watch a range APR loads into."* That sentence is
now false, and it is not passively false: it is phrased as a standing instruction not to try, in the
one section of the one document whose entire job is to be obeyed without re-derivation. An agent who
greps this doc for the write trace and lands on `:330` stops, and never learns that
`PROSPER_DMEM_WRITE_TRACE_REBASE=1` is the answer. This is the same propagation failure the charter
records under the retracted-2.4% example — a stale statement keeping the document's authority after
the measurement that gave it that authority was withdrawn.

Your own PR body has the same pair: *"**This instrument therefore cannot watch a range that APR loads
into** — that is the blocker for naming the writer"* still stands under "The write trace, and the
limitation worth carrying forward".

The fix is a few lines and loses nothing: rewrite `:330-336` as history — the invalidation was real,
it is *why* the trace saw `page-faults=0`, and #3147's rebase lifts it — and keep the hand-rolled
`SIGSEGV` trap paragraph, which is still a live "do not do this". Then drop or reword the body
sentence.

**Second half of the same finding: #3147 is OPEN.** `PROSPER_DMEM_WRITE_TRACE_REBASE` does not exist
on master, so if #3145 merges first the doc cites a measurement nobody can reproduce with the code the
doc ships beside. Your own 16:52:39Z comment says *"I will land that with the #3146 PR"*, which was the
right call. Either move the bullet to #3147, or say in it that the mode arrives with #3147 — a reader
who tries the variable and gets nothing will otherwise conclude the finding is wrong rather than that
the PR has not landed. (Both numbers are fine as written: #3146 is the issue, #3147 the PR
implementing it.)

## Non-blocking

**N10. "The guest clears its own staging buffer" drops a hedge the issue was careful to keep.** #3142
notes `decoded-write-bytes=0` throughout, so the trace establishes *guest stores at a 16-byte stride*,
not that the stored value was zero; the zero-ness comes from ZEROWATCH observing the result. That
composition is sound and I would state it as one — "guest stores, and ZEROWATCH shows the range is
zero afterwards" — rather than as a single reading, exactly as your issue comment does with
"'memset specifically' is the obvious reading … not something the trace decodes".

**N11.** `:297-298` now forward-references *"the row below names it"*. Fine while both bullets are
adjacent; if the new bullet moves to #3147, that sentence moves with it.

N1–N9 from my `f7f1c136` review are unchanged and all still non-blocking.

---

To be unambiguous for whoever merges: **the code is approved at `f7f1c136` == `794b0236`; this
rejection is entirely about two prose paragraphs in `STRAY_STATUS.md` contradicting each other.** Push
the wording fix and I will approve the new head on sight.

mattias800 (commented):
**APPROVED** on the exact head `bbed99b3`. Round-3 B1 is fixed, and the fix is better than the one I
asked for.

Scope of this pass: the `794b0236..bbed99b3` delta only — 12 insertions / 7 deletions in
`prosper/docs/STRAY_STATUS.md`, nothing else. `git diff f7f1c136 bbed99b3 -- prosper/src prosper/tests
prosper/tools` is **empty**, so the code carrying my `f7f1c136` approval is byte-identical here and
every verification in that review stands unchanged.

Re-run at `bbed99b3`: docs gate rc=0, 155 files, `prosper/docs/STRAY_STATUS.md: 12 table(s), 39 rows,
unbroken`; `git diff --check origin/master...bbed99b3` clean; 0 `Claude-Session:` trailers.

## B1 — closed

The contradiction is gone, and gone the right way. `STRAY_STATUS.md:330-341` is now past tense
throughout — *"What **used** to block"*, *"was invalidated"*, *"could never watch"* — and the sentence
that was doing the damage is gone with it: the standing *"so nobody re-derives it"* instruction has
been replaced by a reason to keep the paragraph (*"it explains why this took so long and what the
attribution above depends on"*), which is a different and correct claim. `:334` also fixes the
subtler half I did not ask for: `page-faults=0` is now attributed to the watch dying at the load
rather than left to read as "no guest write happened", which is exactly the misreading the original
phrasing invited.

The merge-order half is closed too, and more completely than a bare "#3147 is open" note would have
been: `:335-338` states that #3147 lifts the blocker, that the `libc.prx` attribution is measured
**with that PR applied**, that **on master alone the trace still disarms at the load and reproduces
nothing**, and what to do in either outcome. A reader who tries the mode on master and gets nothing
now learns why from the doc instead of concluding the finding is wrong.

## Your scope question — keep it here, do not move it to #3147

I would leave the `libc.prx` sub-bullet exactly where it is, for three reasons:

1. **The caveat makes the claim self-invalidating, which is the shape the charter prefers.** A reader
   who cannot reproduce it is told, in the same paragraph, precisely which condition of theirs differs
   and what follows. That is strictly better than an uncaveated true statement, and it is the property
   an "experiment that detects its own invalidity" is supposed to have.
2. **Moving it splits one finding across two PRs, and the composition is the finding.** #3145 supplies
   the first half (the range goes zero 0.5–2.3 s after the load, under an unchanged `rw-s` mapping);
   #3147 supplies the attribution. If #3145 merges without the sub-bullet, master's `## Ruled out`
   says *"something zeroes these ranges"* while the answer sits in an unmerged branch — which is the
   failure mode the charter's "a falsification that lives only in a PR body" rule exists to prevent,
   and it is worse than a caveated true statement, not better.
3. **The only case where moving wins is #3147 being closed unmerged, and `:338` already covers it**
   with an explicit instruction to re-derive.

One follow-through, not a condition of this approval: when #3147 merges, drop the caveat **in that
PR** — the doc line tells the next reader to, but nothing makes it happen, and a caveat that outlives
its condition is the same stale-authority problem in a smaller size.

## Non-blocking, carried forward unchanged

N10 from my `794b0236` review still stands and is still not blocking: *"the guest clears its own
staging buffer"* (`:312-313`) states as one reading what is really a composition of two instruments —
the trace establishes guest stores at a 16-byte stride with `decoded-write-bytes=0`, and ZEROWATCH
establishes that the range is zero afterwards. Your own #3142 comment keeps that hedge; the doc drops
it. Half a line whenever you next touch the bullet.

N1–N9 from the `f7f1c136` review are unchanged and all still deferrable. The one I would still spend
five minutes on before this leaves your hands is the `prosper/tools/AGENTS.md` entry for
`PROSPER_ZEROWATCH` / `PROSPER_PHYSLOG` — third round of asking, and the sampling model (256 B at the
seeded window, >=1 MiB only, 64-entry cap) is currently disclosed only in the log line itself.

Merge when required CI is green.

PR #3150, opened 2026-08-30
perf(compute): cache DCC-compressed sampled images — Stray's splash +66%

### LINKED ISSUE #3149: Stray boot is host-bound: ~38% of CPU in the compute path, bulk tile/detile copying, GPU at 4.5%
Stray's splash phase spends ~38% of CPU in the compute path, and the cost is bulk tile/detile copying plus syscalls — not the GPU, and not shader translation.

Goal: make the boot-to-title sequence fast. Rendering correctness is tracked separately (#3126).

## Three instruments, no two sharing a source

**1. F8 `.prperf`, captured 30 s into the boot** (5.02 s window, `PROSPER_NULL_PAGE=1`):

```
components: renderer-resource=529.1ms gpu-wait=300.3ms gpu-device=225.8ms
            readback=394.6ms compute=2981.0ms
primary evidence: compute — largest measured component (2981.0 ms, 67%)
process CPU 2.08 cores · RSS 3.9-4.0 GiB · guest flips 12.75/s · rendered 38.84/s
```

Three compute programs dominate, one dispatch each per frame at **~12-13 ms**:

| program | dispatches | total | mean |
| --- | --- | --- | --- |
| `0x3011300000` | 65 | 861.4 ms | 13.25 ms |
| `0x30112e0000` | 65 | 821.7 ms | 12.64 ms |
| `0x300dfa0000` | 65 | 774.4 ms | 11.91 ms |

**2,457 ms of the 2,981 ms (82%) in 195 dispatches.** Three of those per frame is a ~25 fps ceiling
before anything else runs, and the capture measured 38.8 rendered/s.

**2. `perf record`** (`-O3 -g -fno-omit-frame-pointer`, `kernel.perf_event_paranoid=1` so kernel
frames resolve), inclusive shares:

| frame | inclusive | self |
| --- | --- | --- |
| `execute_ordered_gpustate` | 58.8% | — |
| `execute_live_compute_items` | **37.7%** | — |
| block-row tile/detile copier (`parallel_rows<sw64kb_copy>`) | 22.0% | 13.0% |
| `__memmove_avx512` | 19.5% | **14.5%** |
| `do_syscall_64` | 22.2% | — |

Kernel page management is visible underneath: `clear_highpages_kasan_tagged` 2.7%,
`zap_present_ptes` 2.4%, `map_anon_folio_pte_nopf` 1.8%, `__free_one_page` 1.6% — large mappings
being created, zeroed and destroyed repeatedly.

**3. GPU is not the constraint.** `gpu-device` is 225.8 ms of the 5.02 s window (~4.5%), consistent
with the earlier `radeontop` reading. Nothing here is a GPU problem.

## What is NOT the cause — measured, not assumed

Three plausible explanations were tested and falsified, which is worth recording so nobody re-derives
them:

* **The image cache missing, forcing re-uploads.** No: `PROSPER_COMPUTELOG` shows both sampled and
  storage images resolving `persistent ... upload-skipped=1`, hundreds of times each, for 33 MB and
  67 MB surfaces. The cache is working and the Vulkan uploads *are* skipped.
* **The per-dispatch guest snapshot at `live_compute.cpp:6957`
  (`cache_source_snapshot.assign`).** No: it is guarded by `!bi.persistent`, and the bindings are
  persistent, so it never runs.
* **The storage-image unpack buffer at `live_compute.cpp:9843`** (a fresh value-initialised
  `std::vector` per call, which would explain the page churn). No: a coverage probe armed on that
  path produced **zero** hits in a 70 s boot — that code is not on the hot path at all.

## Where it actually is

The copying is in the compute path but *after* the upload decision, so "upload skipped" does not mean
"no bulk work". `parallel_rows<sw64kb_copy<...>>` and `__memmove_avx512` together account for roughly
a fifth of all CPU while the GPU sits at 4.5%, and the F8's `readback=394.6ms` points at the
result-writeback side rather than the input side.

Next step is to establish which of those two directions the block copier is running in per dispatch —
input detile or result re-tile — and whether a dispatch whose result is unchanged still pays for it.
`live_compute.cpp:9019` already computes `changed_bytes` by comparing the result against the seed, and
`__memcmp_evex_movbe` is 1.4% of CPU, so the machinery to detect "nothing changed" exists and is
already being paid for.

## Reproduction

```bash
cmake -S prosper -B build -DPROSPER_APP=ON -DCMAKE_CXX_FLAGS='-g -fno-omit-frame-pointer'
SDL_VIDEODRIVER=offscreen PROSPER_NULL_PAGE=1 PROSPER_RENDER=1 \
  PROSPER_GUEST_ARGS=-force-gfx-direct PROSPER_PERF_CAPTURE_AFTER_MS=30000 \
  PROSPER_CAPTURE_DIR=<out> ./build/prosper-app <DUMP_ROOT>/PPSA02101-app0
python3 prosper/tools/perf/performance_capture_report.py <out>/*.prperf
```

`perf` needs `kernel.perf_event_paranoid<=1` or every kernel frame is silently missing; see LOCAL.md.

mattias800:
Stray's splash sequence is host-bound, not GPU-bound, and `execute_live_compute_items` dominated the
CPU. A `PROSPER_TILECENSUS` sweep (new in this PR) found the work is almost entirely one surface
geometry, re-detiled thousands of times because DCC-compressed sampled surfaces were excluded from
the persistent compute image cache.

Fixes #3149.

## What this changes

One gate: a sampled surface is admitted to the compute image cache whether or not its DCC metadata
plane reads as all-`0xff`. Plus the census that found it, and a kill switch to bisect it.

One thing I found while reworking, not raised in review: dropping the metadata test left the
all-`0xff` scan **dead on the default path** — a full pass over the metadata plane, per binding per
dispatch, whose result nothing consumed. That is the same class of defect as B5 on #3147, and it
would have been added to the exact hot path this PR shortens. The scan now runs only when the kill
switch or the compute trace can act on its result. `sampled_dcc_fast_clear` is computed
unconditionally as before, so the transition argument below is unaffected.

## Why that is sound — and why the first revision's argument was wrong

**The first revision argued the wrong thing, and the review was right to reject it.** It folded the
metadata plane into `ComputeImageCacheKey` and validated it beside the base bytes, claiming a
compressed surface's identity was incomplete without it. Two independent problems:

* The guard was conjoined into only one of three skip proofs at `:2046`, so `watch_unchanged` and
  `exact_unchanged` — both base-range-only — skipped the upload anyway in exactly the scenario the PR
  body named. **The PR's central claim was false as implemented.**
* The premise was also wrong, which is why the fix is now smaller rather than longer.

The metadata plane's only reachable consumer on this path is
`compute_sampled_dcc_fast_clear_rgba8`, and the gate requires `!sampled_dcc_fast_clear`, which
excludes every surface that reaches it. **So for a surface this cache can hold, the upload is detile
+ unpack of the base bytes and never reads the metadata at all.** Base bytes are exactly what
`acquire_cached_image` already validates, so a compressed entry's identity was complete all along and
the all-`0xff` test was only ever costing hit rate.

I verified that by a second, independent route rather than on the reviewer's word — the derived
`sampled_dcc_metadata` could have been one of several channels. Every read of `bi.dcc_metadata` and
`r->metadata_addr` in the file is a storage-target path (the writeback at `:9469+`, the storage gate
at `:7130`, alias matching, or the fast-clear helper itself). And the changed gate sits inside
`compute_sampled_guest_prepare_required`, i.e. `!storage && !renderer_owned`, so storage targets keep
their own stricter gate untouched.

This deliberately does **not** change how a compressed surface decodes — only whether that decode's
result may be reused. If reading base bytes without consulting a live DCC plane is wrong for some
surface, it is equally wrong on both sides of this branch. That is what makes the change
pixel-neutral *by construction*, and it is a much stronger claim than the first revision's.

B1's alternative ends "keep only whatever the key needs to disambiguate entries". The answer is
**nothing** — master's key is already sufficient and is therefore unchanged. Two entries agreeing on
every key field agree on `gpu_addr`/`guest_bytes` (so the same base bytes) and on the whole geometry
(so the same layout), and the upload is a function of exactly those. A differing `metadata_addr`
cannot make their uploads differ, because neither upload reads it.

### The sharpest objection, and why it holds

The obvious challenge to decode-independence is a *transition*: what if a metadata write turns a
surface into (or out of) a fast clear while its base bytes sit still? That is the one case where the
plane changes what should be displayed.

It is handled, because `sampled_dcc_fast_clear` is recomputed per binding per dispatch, not cached.

* **Into a fast clear.** The next dispatch computes `sampled_dcc_fast_clear = true`, so
  `cache_candidate` goes false and the fast-clear materialization path runs. The cache is not
  consulted, stale or otherwise.
* **Out of a fast clear.** While it was a fast clear it was never cached, so there is no stale entry
  to serve. Its first admission uploads.

In both directions the gate is re-evaluated against the current metadata before the cache is
reached, which is what makes "the cacheable set never reads the plane" a statement about every
dispatch rather than only the first.

## Review disposition

| # | Status |
| --- | --- |
| **B1** | Fixed by taking the decode-independence route the review identified. Metadata is out of the key and out of the validation; the comment now states the argument the code implements. |
| **B2** | Moot — no footprint cast and no `metadata_bytes` field remain. |
| **B3** | Gate extracted as `compute_sampled_cache_gate_candidate` beside its storage sibling, with tests; clean-zero addressed below. |
| **N1** | Moot — no key change, so both cross-descriptor borrows are untouched and the kill switch now gates the entire change. |
| **N2** | Moot — no new key fields to hash. |
| **N3** | Fixed — stale `PROSPER_DCC_CACHE_UPSIDE` comment deleted. |
| **N4** | Moot — with no metadata in the key, the switch restores the old behaviour exactly. |
| **N5** | Addressed with a within-arm noise floor; see below. |
| **N6** | Fixed — `who` is part of the census key, so one geometry reached from two sites is two rows. |
| **N7** | Fixed — the 33-bit shift is gone (64-bit accumulator), and `TileCensusScope` now documents the static-lifetime invariant. |
| **N8** | Fixed — both variables registered in `prosper/tools/AGENTS.md`. |

### Second review (`0d7a0d6e`) — APPROVED, 8 non-blocking findings, all addressed

| # | Action |
| --- | --- |
| **N1** | Fixed, and it was a real error: `guest=66846720` is 3840x**2176**x8, not 3840x2160x8. I asserted a multiplication I never performed, in the one sentence joining two instruments. |
| **N2** | Gate comment now covers the *second* thing `cache_candidate` unlocks — the compute-transfer seed borrow, which seeds from a retained storage image rather than replaying this decode — and why that is separately safe. |
| **N3** | Invariant note extended to the declining direction and names the unwired seam (`compressed_source_authority.hpp`'s `sampled_source_decision`), which must move with this gate when wired. |
| **N4** | Comment tightened: the *probe* runs for every compressed sampled surface; what the gate excludes is the *materialization*. |
| **N5** | Fourth predicate corner asserted — the uncompressed default path, which was the one left out. |
| **N6** | Volume census rows saturate instead of wrapping `uint32_t`; `TileCensusScope` copy/assign deleted. |
| **N7** | `## Ruled out` row added to `docs/RESOURCE_BINDING.md`; `BLOG.md` entry added. |
| **N8** | Pixel A/B demoted to corroboration, with the minimum-seeking weakness stated. |

The reviewer also supplied a stronger form of the core argument than mine:
**`gfx10_dcc_fast_clear_rgba8` is the only DCC decode function in the tree** — prosper has no DCC
decompressor at all — so metadata-independence is a property of the codebase, not merely of this call
graph. That is now the argument recorded in `## Ruled out`.
| **N9** | Moot — that comment described the metadata validation and was deleted with it. |

### On B3's test arms

The review asked for arms (a) uncompressed key byte-identical to master, (b) footprint above
`UINT32_MAX` declined, (c) metadata changed while base unchanged → not skipped. All three describe
the design that was rejected; under decode-independence there is no key change and metadata is
irrelevant by construction, so those arms no longer have anything to assert. The arms that carry the
new design are: a compressed surface **is** admitted (the one that fails against the old predicate),
the kill switch restores the old rejection without disabling the cache for uncompressed surfaces, and
fast clears stay excluded on both sides.

I verified they discriminate rather than assuming it, by mutation: reverting
`compute_sampled_cache_gate_candidate` to master's gate
(`!sampled_dcc_fast_clear && dcc_cache_safe && persistent_enabled`, quoted from
`origin/master:live_compute.cpp:6835`) and rebuilding the real test file against the mutated header.

```
MUTANT test rc=1   1 failure:  FAIL compressed sampled surface is a cache candidate (#3149)
CLEAN  test rc=0   0 failures
```

So **1 of the 6 new arms fails without the fix** — the compressed-admission arm — and the other five
pin behaviour that must *not* change. I state that split explicitly because "the tests would fail
without the fix" would be false for five of them, and a reviewer should not have to discover that.

## Verification

Build clean, **ctest 315/315 passed, 0 failed** (4 pre-existing `-DGAME_DUMP` skips).
`check_diag_gates.py` green (570 files, 735 `PROSPER_*` variables, 108/108 baselined),
`check_cached_env.py` green, docs gate green, `diff --check` clean.

### Performance — one variable, same harness, same 110 s

`prosper-app`, headless, no pad input (this measures the splash sequence, which is the subject).

| arm | fps mean | tiling bytes moved |
| --- | --- | --- |
| default (this PR) | **62.24** (109 windows) | **3,176 MiB** |
| `PROSPER_NO_DCC_IMAGE_CACHE=1` (pre-#3149) | **37.42** (65 windows) | **246,977 MiB** |

**+66% frame rate, 77.8x less tiling work.**

### B3's clean zero — the instrument, then the answer

The review's point was that "both guards pass" is consistent with the new path never running. To
test that I added `compressed=`/`dcc-safe=` to the existing persistent-sampled trace line, so the
class only #3149 admits (`compressed=1 dcc-safe=0`) is directly countable.

**Positive control first, per the charter's rule about believing a zero — and it was needed.** My
first attempt at the messenger route reported `0` of every class, and had I only counted the
compressed ones that would have read as "the guarded route does not exercise the path". It was
instead a VOID measurement: my hand-rolled route never progressed (`0% of 7244 published frames
carried new content`), so it exercised nothing at all. Counting the TOTAL alongside the compressed
subset is what made the difference legible. On Stray, where the path is known to run:

```
persistent sampled images (total) : 23426
... compressed=1 dcc-safe=0       :  8235      <-- only #3149 admits these
[compute] persistent sampled image binding=10 addr=0x3017370000 guest=66846720 \
          upload-skipped=1 compressed=1 dcc-safe=0
```

`guest=66846720` is 3840 x **2176** x 8 — the census geometry with its tile height padded
(3840x2160x8 would be 66,355,200, and I wrote that in the first draft without doing the
multiplication; a reader who did it would have concluded the two instruments disagreed). It is the
same surface, independently confirmed as a cache **hit** on one the old gate rejected. All 8,235 are
`upload-skipped=1`, i.e. every one is a hit rather than merely an admission.

**Then the guarded route, and the answer is not the one I wanted.** Driving `snapshot.py`'s own
`capture_content()` so the dump, pad script, scale, savedata policy and timeout are exactly the
guard's, with only the trace added:

```
messenger-scene: 2,452,657 log lines, 222,908 compute dispatches, 668,736 bindings
                 persistent sampled images (total) : 0
```

Not a void run — it dispatched heavily. I first gave two different reasons for the zero; only one
is right. Every binding logs `cls=0 ... dims=0x0`, i.e. they are **buffers, not images**, so this
title's compute path never reaches the sampled image cache at all. The 1 MiB sampled minimum
(`live_compute.hpp:38`) is irrelevant here because execution never gets that far.

**So the reviewer was right, and the honest conclusion is that this guard cannot validate this
change.** It exercises the storage gate, which this PR deliberately leaves alone. Its passing is
evidence of no cross-title regression in shared code — worth having, and worth *not* overstating.
The change's safety rests on the structural argument above, which is by-construction rather than
statistical, plus the Stray measurements where the path demonstrably runs.

### N5 — the controlled pixel comparison

The original 3% frame A/B is **withdrawn**: it compared frames from different points in an animating
sequence, exactly as the review said. Replacement, with a within-arm noise floor so the cross-arm
number has a scale:

| comparison | mean abs delta | pixels differing by >8 |
| --- | --- | --- |
| **within-arm** A1 vs A2 (same config, two runs) — the noise floor | **0.106** | 0.40% |
| **cross-arm** A1 vs B1 (default vs `PROSPER_NO_DCC_IMAGE_CACHE=1`) | **0.100** | 0.41% |
| full-resolution re-check of the closest cross-arm pair | **0.000** | 0.00% |

The cross-arm delta sits at or below run-to-run variation of the same configuration, and the closest
cross-arm pair is byte-identical at full 4K.

**Read this as corroboration, not as independent proof, and the reason is a real weakness.** Pairing
by closest match is minimum-seeking: for each A frame it selects the B frame that minimises the very
statistic then reported, and "full-resolution re-check of the *closest* pair" selects the best case
twice. The failure mode of a cache-staleness bug is "this frame resembles a slightly earlier frame"
— which is precisely what closest-match pairing absorbs. The within-arm floor is computed the same
way, so the comparison is at least like-for-like. **What carries this PR is the structural argument**;
this table is consistent with it, not a second proof of it.

The comparator was validated before being trusted, and it needed to be: its first positive control
**failed to detect deliberately corrupted frames**, because the capture set was `mean=0.0, std=0.0`
— entirely black — so the injected defect was not expressible on it. With a perturbation that cannot
be a no-op it moves 8.14x the noise floor. It now also refuses to return a verdict at all when an
arm is blank, since "no pixel difference" over black frames reads identically to "no corruption".

Getting a valid anchor took four attempts, recorded so nobody repeats them: the animating logo
sequence is what N5 rejected; the held title screen is unreachable by the SLOW arm inside any window
the fast arm tolerates (0 frames); and `--warmup-seconds` **skips Vulkan rendering** rather than
merely delaying capture, so a long warmup changed the workload and the guest faulted. What works is
sampling widely from a known-good configuration and pairing by closest match.

### Cross-title guards

`messenger-scene` **OK**, `dead-cells-gameplay` **OK**, rc=0. Read with the caveat established
above: neither route creates a persistent sampled image, so these show no cross-title regression in
shared code rather than validating the changed path.

### Through to the title screen, and a harness artifact I nearly reported as a finding

The task asked about the splash "or all the way to title screen if appropriate". Under `prosper-app`
(GPU-present, the shipped path), 285 s carrying through to the title screen holds **64.78 fps mean
over 302 windows**, with the `smpl-upl` row still at **5 calls / 316 MiB (5.1%)** — the same 5 calls
as the 110 s splash run, so the fix holds for the whole sequence rather than only the opening.

The same phase measured under `tools/screenshot` looked completely different — 158,381 MiB of tiling
against `prosper-app`'s 6,254 MiB, and `tile_surface` apparently dominating at ~41%. I was about to
file that as a second bottleneck. It is the harness: `tools/screenshot` never calls
`set_gpu_present_active`, so it copies every scanout frame back to the CPU, and that readback
inflates tiling work **25x**. There is no writeback bottleneck on the shipped path, and no issue to
file. Recording it because the wrong version of this paragraph was one command away from being
published as a measurement of the title.

### The census now carries the causal claim (this is N6's payoff)

With `who` in the key, the row that moves is identifiable — and it is exactly the one this fix
caches, with the rest of the profile untouched:

| census row | pre-#3149 | this PR |
| --- | --- | --- |
| `smpl-upl detile_surface 3840x2160 bpe=8 mode=27` | **3,860 calls / 244,266 MiB (98.9%)** | **5 calls / 316 MiB** |
| `compute tile_surface 3840x2160 bpe=8 mode=27` | 16 calls | 16 calls |
| `stor-seed detile_surface 3840x2160 bpe=8 mode=27` | 9 calls | 9 calls |
| `compute tile_surface 120x68 bpe=8 mode=27` | 1,285 calls | 1,280 calls |

**772x fewer calls on the tagged sampled-upload row, every other row unchanged.** Under the old
last-writer-wins attribution those first three rows collapsed into one, which is why the previous
revision reported this shape as `who=compute` and could not establish that the 98.9% was the upload
being cached. It now can.

### DIFF
--- a/BLOG.md
+++ b/BLOG.md
@@ -19,6 +19,20 @@ from the tracker issues, and still gated, because it is a projection of state ra
 > title's current state — for that, read the tracker. Nothing is ever removed when a title moves on,
 > because the point of a blog is that it records *when* things happened.
 
+## 2026-08-31
+
+### Stray's splash runs 66% faster, and the title screen now holds 65 fps
+
+No picture: the title screen still renders black, so there is nothing new to look at yet. But the
+sequence in front of it stopped crawling.
+
+One 4K surface was being de-swizzled from scratch thousands of times — 241 GiB of tiling work in
+under two minutes — because DCC-compressed textures were barred from the compute image cache. The
+bar was checking the wrong thing: nothing on that path ever reads the compression metadata it
+inspected, so the cache could have held those textures all along. Five detiles now do what 3,860 did.
+
+Splash goes 37 → 62 fps; carried through to the title screen, a run averages 65. ([#3150](https://github.com/mattias800/prosper/pull/3150))
+
 ## 2026-08-30
 
 ### Stray's black title screen is not a loading failure — the data arrives, then leaves
--- a/prosper/docs/RESOURCE_BINDING.md
+++ b/prosper/docs/RESOURCE_BINDING.md
@@ -194,3 +194,21 @@ sits a chain of about ten `continue`s, and without the input side "the front hal
 resource" and "it saw it and rejected it" produce identical evidence while pointing at different
 files. Reach for it before concluding anything about descriptor *recovery*: on Earthion (#1590) it
 overturned exactly that conclusion in one run.
+
+**A DCC-compressed SAMPLED surface's cache identity does NOT include its metadata plane, and folding
+that plane into the key is both unnecessary and — as first attempted — ineffective.** #3150 /
+#3149, 2026-08-31. The persistent compute image cache excluded compressed sampled surfaces on the
+theory that base bytes alone could not determine their content; on Stray that exclusion re-detiled
+one 4K FP16 surface 3,860 times in 110 s (98.9% of all tiling work, `PROSPER_TILECENSUS`). The
+theory is wrong for the *sampled* path: the DCC plane's only consumer there is
+`compute_sampled_dcc_fast_clear_rgba8`, whose materialization the admission gate already excludes
+via `!sampled_dcc_fast_clear`, so a cacheable entry's upload is detile + unpack of the BASE bytes
+alone — which the existing validation already covers. More sharply, `gfx10_dcc_fast_clear_rgba8` is
+the **only** DCC decode function in the tree: prosper has no DCC decompressor, so
+metadata-independence is a property of the codebase rather than of one call graph. **Do not re-derive
+the metadata-in-the-key design**: the first revision of #3150 did, and it was rejected twice over —
+the guard was conjoined into only one of three skip proofs (`watch_unchanged` and `exact_unchanged`
+both examine the base range alone, so the upload was skipped anyway), and the premise was unnecessary
+regardless. The *storage* gate is different and still requires an all-`0xff` plane: a storage
+target's base bytes are not authoritative while a live plane exists. `PROSPER_NO_DCC_IMAGE_CACHE=1`
+restores the old exclusion for bisection.
--- a/prosper/frontends/shared/compute/compute_transfer_gate_census.hpp
+++ b/prosper/frontends/shared/compute/compute_transfer_gate_census.hpp
@@ -49,8 +49,14 @@ constexpr bool compute_storage_cache_gate_candidate(
 //   * if a sampled decode ever starts DECLINING on metadata state, this gate has to decline with
 //     it.  That seam already exists and is unwired:
 //     `gpu/resources/compressed_source_authority.hpp`'s `sampled_source_decision` has no production
-//     caller today, and the surfaces admitted here are precisely the ones it would return
-//     `DeclinedUnsupportedMetadataState` for.  Wiring it must move this gate in the same commit.
+//     caller today, and it classifies exactly the surfaces this gate now admits.  **Wiring it must
+//     move this gate in the same commit.**  Deliberately not stated more precisely than that: the
+//     two do not stand in a clean correspondence.  That function has five distinct decline reasons
+//     and a non-declining `DccUncompressedBase` outcome, its `CompressionMetadataKind` is an
+//     adapter-supplied input so `Unknown` is reachable, and an import bypass pre-empts the metadata
+//     reasons entirely -- so a compressed surface admitted here can land on several of its
+//     outcomes, including non-declining ones.  An earlier draft of this comment named one specific
+//     reason and was wrong in both directions (#3150 review).
 // Nothing here can assert either mechanically, so both are written where somebody touching that
 // decision will be looking.
 //
--- a/prosper/frontends/shared/compute/compute_transfer_gate_census.hpp
+++ b/prosper/frontends/shared/compute/compute_transfer_gate_census.hpp
@@ -31,6 +31,39 @@ constexpr bool compute_storage_cache_gate_candidate(
            inputs.persistent_enabled;
 }
 
+// The sampled-image cache decision, in the same shape and for the same reason: a census that
+// reports only the candidate bit cannot tell "this surface was ineligible" from "the gate was never
+// reached".
+//
+// `dcc_cache_safe` is deliberately NOT required here, and that asymmetry against the storage gate
+// above is the substance of #3149 rather than an oversight.  A storage target's base bytes may not
+// be authoritative while a DCC plane is live, so that gate must consult it.  A *sampled* upload is
+// detile + unpack of the base bytes and never reads the metadata plane -- the fast-clear helper is
+// its only consumer, and `sampled_dcc_fast_clear` excludes exactly the surfaces that reach it.  So
+// base bytes are the whole identity of a cacheable sampled entry, which is what the cache already
+// validates, and requiring an all-0xff plane on top of that only cost hit rate.
+//
+// That last sentence is a live invariant, not a historical note: **if a sampled decode ever starts
+// consulting the DCC metadata plane, this gate has to consult it too**, or the cache will replay a
+// decode whose input it never validated.  Nothing here can assert that mechanically, so it is
+// written where somebody adding such a consumer will be looking.
+//
+// `dcc_cache_disabled` (PROSPER_NO_DCC_IMAGE_CACHE) restores the older, stricter behaviour exactly,
+// so a rendering report can be bisected against this change with one variable.
+struct ComputeSampledCacheGateInputs {
+    bool sampled_dcc_fast_clear = false;
+    bool dcc_cache_safe = false;
+    bool dcc_cache_disabled = false;
+    bool persistent_enabled = false;
+};
+
+constexpr bool compute_sampled_cache_gate_candidate(
+    const ComputeSampledCacheGateInputs& inputs) {
+    return !inputs.sampled_dcc_fast_clear &&
+           (inputs.dcc_cache_safe || !inputs.dcc_cache_disabled) &&
+           inputs.persistent_enabled;
+}
+
 // A compressed writable target cannot enter the persistent cache before dispatch because its base
 // bytes are not yet authoritative.  A successful ordinary writeback changes that fact: it publishes
 // exact base bytes and marks the complete DCC plane uncompressed.  Admit that result only when DCC
--- a/prosper/frontends/shared/compute/compute_transfer_gate_census.hpp
+++ b/prosper/frontends/shared/compute/compute_transfer_gate_census.hpp
@@ -43,10 +43,16 @@ constexpr bool compute_storage_cache_gate_candidate(
 // base bytes are the whole identity of a cacheable sampled entry, which is what the cache already
 // validates, and requiring an all-0xff plane on top of that only cost hit rate.
 //
-// That last sentence is a live invariant, not a historical note: **if a sampled decode ever starts
-// consulting the DCC metadata plane, this gate has to consult it too**, or the cache will replay a
-// decode whose input it never validated.  Nothing here can assert that mechanically, so it is
-// written where somebody adding such a consumer will be looking.
+// That last sentence is a live invariant, not a historical note, and it cuts both ways:
+//   * if a sampled decode ever starts CONSULTING the DCC metadata plane, this gate has to consult
+//     it too, or the cache will replay a decode whose input it never validated;
+//   * if a sampled decode ever starts DECLINING on metadata state, this gate has to decline with
+//     it.  That seam already exists and is unwired:
+//     `gpu/resources/compressed_source_authority.hpp`'s `sampled_source_decision` has no production
+//     caller today, and the surfaces admitted here are precisely the ones it would return
+//     `DeclinedUnsupportedMetadataState` for.  Wiring it must move this gate in the same commit.
+// Nothing here can assert either mechanically, so both are written where somebody touching that
+// decision will be looking.
 //
 // `dcc_cache_disabled` (PROSPER_NO_DCC_IMAGE_CACHE) restores the older, stricter behaviour exactly,
 // so a rendering report can be bisected against this change with one variable.
--- a/prosper/frontends/shared/live/live_compute.cpp
+++ b/prosper/frontends/shared/live/live_compute.cpp
@@ -842,6 +842,12 @@ struct ComputeImageCacheKey {
     uint32_t mip_tail_offset = 0, mip_tail_bytes = 0;
     uint32_t mip_tail_x = 0, mip_tail_y = 0;
     uint32_t vk_format = 0;
+    // #3149: the DCC metadata plane is part of a compressed surface's identity. Base bytes alone do
+    // NOT determine its content -- that is precisely why compressed surfaces were excluded from this
+    // cache -- so the metadata range joins the key, and the validation below requires it unchanged
+    // too. Zero for an uncompressed surface, which keeps every existing entry's identity unchanged.
+    uint64_t metadata_addr = 0;
+    uint32_t metadata_bytes = 0;
     bool storage = false;
     bool in_mip_tail = false;
     bool srgb = false;
@@ -881,7 +887,13 @@ ComputeImageCacheKey storage_image_cache_key(const prosper::gpu::ShaderResource&
         resource.layer_stride_bytes, resource.layer_mip_offset_bytes,
         resource.mip_tail_offset, resource.mip_tail_bytes,
         resource.mip_tail_x, resource.mip_tail_y,
-        static_cast<uint32_t>(native_format), true, resource.in_mip_tail,
+        static_cast<uint32_t>(native_format),
+        // #3149: same positional slot as the sampled key -- between vk_format and storage.
+        resource.compression_enabled ? resource.metadata_addr : 0ull,
+        resource.compression_enabled
+            ? static_cast<uint32_t>(prosper::gpu::gpu_capture_dcc_metadata_footprint(resource))
+            : 0u,
+        true, resource.in_mip_tail,
         resource.srgb, resource.depth_compare};
 }
 
@@ -1983,6 +1995,9 @@ struct VulkanComputeContext {
         ++cached.pins;
         image = cached.image;
         memory = cached.memory;
+        // The epoch fast path is per-submit and covers both ranges implicitly: an epoch is only
+        // reused within a submit, and a metadata write inside the submit bumps the journal that
+        // `metadata_unchanged` below consults on the next acquisition.
         if (!key.host_data && validation_epoch && cached.validation_epoch == validation_epoch) {
             upload_skipped = cached.validation_result;
             return true;
@@ -1992,7 +2007,16 @@ struct VulkanComputeContext {
         // entries only against their exact retained snapshot. This makes warm replay exercise the
         // same image residency as live execution without ever trusting an unrelated guest address.
         const bool exact_source_only = key.host_data != 0;
+        // Both ranges, not just the base. A compressed surface whose metadata changed while its
+        // base bytes did not would otherwise serve stale pixels -- the exact unsafety that kept
+        // these surfaces out of the cache. An uncompressed entry has metadata_bytes == 0 and takes
+        // the same path it always did.
+        const bool metadata_unchanged = key.metadata_bytes == 0 ||
+            prosper::gpu::guest_gpu_writes_since(cached.validation_snapshot,
+                                                 key.metadata_addr, key.metadata_bytes) ==
+                prosper::gpu::GuestGpuWriteQuery::Unchanged;
         const bool submit_unchanged = cached.content_valid && !exact_source_only &&
+            metadata_unchanged &&
             prosper::gpu::guest_gpu_writes_since(cached.validation_snapshot,
                                                   key.gpu_addr, key.guest_bytes) ==
                 prosper::gpu::GuestGpuWriteQuery::Unchanged;
@@ -6832,7 +6856,29 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                 // this first narrow path uncached rather than teaching the existing base-byte key
                 // an unsafe partial identity; staging materialization is still far cheaper than
                 // detiling and converting the 2x-larger FP16 base.
-                bi.cache_candidate = !sampled_dcc_fast_clear && dcc_cache_safe &&
+                // PROSPER_DCC_CACHE_UPSIDE=1 (#3149) is a MEASUREMENT, not a fix, and must never
+                // become the default: it admits DCC-compressed surfaces to the image cache whose
+                // identity is base bytes only, so a surface whose metadata changes while its base
+                // bytes do not would serve stale pixels. It exists to answer one question -- is the
+                // real fix (folding the metadata plane into the cache identity) worth building? --
+                // because that fix is a correctness-sensitive change this code explicitly warns
+                // against making naively, and it should not be built on a guess about its payoff.
+                // #3149: a compressed surface is now cacheable because its DCC metadata plane is
+                // part of the cache identity and is validated alongside the base bytes -- the
+                // "unsafe partial identity" this code warned about is exactly what was added, so the
+                // warning is satisfied rather than bypassed. Requires a metadata range to validate:
+                // compression with no metadata address stays out, failing closed as before.
+                // Kill switch, and it earns its place: this admits a whole class of surface to a
+                // cache it was previously excluded from, so a single variable must be able to
+                // restore the old behaviour exactly -- for bisecting a rendering report, and for the
+                // pixel A/B that justified the change in the first place.
+                static const bool dcc_cache_disabled =
+                    std::getenv("PROSPER_NO_DCC_IMAGE_CACHE") != nullptr;
+                const bool dcc_identity_tracked = !dcc_cache_disabled &&
+                    r->compression_enabled && r->metadata_addr &&
+                    prosper::gpu::gpu_capture_dcc_metadata_footprint(*r) > 0;
+                bi.cache_candidate = !sampled_dcc_fast_clear &&
+                    (dcc_cache_safe || dcc_identity_tracked) &&
                     persistent_compute_image_enabled(sbytes, ComputeImageCacheClass::sampled);
                 const VkFormat transfer_native_format =
                     compute_transfer_storage_vk_format(r->format, sampled_components);
@@ -6884,7 +6930,15 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                         r->layer_stride_bytes, r->layer_mip_offset_bytes,
                         r->mip_tail_offset, r->mip_tail_bytes,
                         r->mip_tail_x, r->mip_tail_y,
-                        static_cast<uint32_t>(image_format), bi.storage, r->in_mip_tail,
+                        static_cast<uint32_t>(image_format),
+                        // #3149: metadata plane joins the identity. Positional aggregate init --
+                        // these two MUST sit between vk_format and storage, matching the struct.
+                        r->compression_enabled ? r->metadata_addr : 0ull,
+                        r->compression_enabled
+                            ? static_cast<uint32_t>(
+                                  prosper::gpu::gpu_capture_dcc_metadata_footprint(*r))
+                            : 0u,
+                        bi.storage, r->in_mip_tail,
                         r->srgb, r->depth_compare};
                     // An ordinary native typed 2D image or native typed 3D volume is byte- and
                     // format-identical to the sampled upload that follows it. Borrow that retained
@@ -7231,6 +7285,7 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                                          r->mip_tail_x, r->mip_tail_y);
                     unpack_source = linear.get();
                 } else if (r->tile_mode) {
+                    const prosper::gpu::TileCensusScope tcs("stor-seed");
                     if (trace) bi.before_hash = fnv1a(src, guest_bytes);
                     detile_surface(linear.get(), src, r->width, r->height, r->tile_mode, 0,
                                    static_cast<uint32_t>(guest_texel));
@@ -7692,6 +7747,7 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                                                  r->tile_mode, bpt, r->mip_tail_x,
                                                  r->mip_tail_y);
                         } else if (r->tile_mode) {
+                            const prosper::gpu::TileCensusScope tcs("smpl-upl");
                             detile_surface(linear.get(), src, r->width, r->height,
                                            r->tile_mode, 0, bpt);
                         }
@@ -10271,6 +10327,7 @@ void sampled_float16_to_unorm8_range(const uint8_t* source, uint32_t components,
 // exercised by a regression test rather than only by a routed run.
 
 bool execute_live_compute_items(const std::vector<prosper::gpu::ComputeItem>& items) {
+    const prosper::gpu::TileCensusScope tile_census_scope("compute");
     auto fail_closed_items = [&]() {
         for (const auto& item : items)
             prosper::gpu::notify_compute_authority_boundary({
--- a/prosper/frontends/shared/live/live_compute.cpp
+++ b/prosper/frontends/shared/live/live_compute.cpp
@@ -842,12 +842,6 @@ struct ComputeImageCacheKey {
     uint32_t mip_tail_offset = 0, mip_tail_bytes = 0;
     uint32_t mip_tail_x = 0, mip_tail_y = 0;
     uint32_t vk_format = 0;
-    // #3149: the DCC metadata plane is part of a compressed surface's identity. Base bytes alone do
-    // NOT determine its content -- that is precisely why compressed surfaces were excluded from this
-    // cache -- so the metadata range joins the key, and the validation below requires it unchanged
-    // too. Zero for an uncompressed surface, which keeps every existing entry's identity unchanged.
-    uint64_t metadata_addr = 0;
-    uint32_t metadata_bytes = 0;
     bool storage = false;
     bool in_mip_tail = false;
     bool srgb = false;
@@ -887,13 +881,7 @@ ComputeImageCacheKey storage_image_cache_key(const prosper::gpu::ShaderResource&
         resource.layer_stride_bytes, resource.layer_mip_offset_bytes,
         resource.mip_tail_offset, resource.mip_tail_bytes,
         resource.mip_tail_x, resource.mip_tail_y,
-        static_cast<uint32_t>(native_format),
-        // #3149: same positional slot as the sampled key -- between vk_format and storage.
-        resource.compression_enabled ? resource.metadata_addr : 0ull,
-        resource.compression_enabled
-            ? static_cast<uint32_t>(prosper::gpu::gpu_capture_dcc_metadata_footprint(resource))
-            : 0u,
-        true, resource.in_mip_tail,
+        static_cast<uint32_t>(native_format), true, resource.in_mip_tail,
         resource.srgb, resource.depth_compare};
 }
 
@@ -1995,9 +1983,6 @@ struct VulkanComputeContext {
         ++cached.pins;
         image = cached.image;
         memory = cached.memory;
-        // The epoch fast path is per-submit and covers both ranges implicitly: an epoch is only
-        // reused within a submit, and a metadata write inside the submit bumps the journal that
-        // `metadata_unchanged` below consults on the next acquisition.
         if (!key.host_data && validation_epoch && cached.validation_epoch == validation_epoch) {
             upload_skipped = cached.validation_result;
             return true;
@@ -2007,16 +1992,7 @@ struct VulkanComputeContext {
         // entries only against their exact retained snapshot. This makes warm replay exercise the
         // same image residency as live execution without ever trusting an unrelated guest address.
         const bool exact_source_only = key.host_data != 0;
-        // Both ranges, not just the base. A compressed surface whose metadata changed while its
-        // base bytes did not would otherwise serve stale pixels -- the exact unsafety that kept
-        // these surfaces out of the cache. An uncompressed entry has metadata_bytes == 0 and takes
-        // the same path it always did.
-        const bool metadata_unchanged = key.metadata_bytes == 0 ||
-            prosper::gpu::guest_gpu_writes_since(cached.validation_snapshot,
-                                                 key.metadata_addr, key.metadata_bytes) ==
-                prosper::gpu::GuestGpuWriteQuery::Unchanged;
         const bool submit_unchanged = cached.content_valid && !exact_source_only &&
-            metadata_unchanged &&
             prosper::gpu::guest_gpu_writes_since(cached.validation_snapshot,
                                                   key.gpu_addr, key.guest_bytes) ==
                 prosper::gpu::GuestGpuWriteQuery::Unchanged;
@@ -6845,8 +6821,18 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                     if (!readable) { skip_image(r, "sampled surface unreadable"); break; }
                 }
 
+                static const bool dcc_cache_disabled =
+                    std::getenv("PROSPER_NO_DCC_IMAGE_CACHE") != nullptr;
+                // #3149: the all-0xff scan is a full pass over the metadata plane, per binding per
+                // dispatch, and after this change nothing on the DEFAULT path acts on its result --
+                // the gate below admits either way. Computing it anyway would add dead work to the
+                // exact hot path this change exists to shorten, so it runs only when something can
+                // consume it: the kill switch (which restores the old gate) or the compute trace
+                // (which reports it). Skipping it leaves `dcc_cache_safe` false for a compressed
+                // surface, which is what the new gate wants and what the old gate would have
+                // concluded for anything that is genuinely compressed.
                 bool dcc_cache_safe = !r->compression_enabled;
-                if (r->compression_enabled) {
+                if (r->compression_enabled && (dcc_cache_disabled || trace)) {
                     dcc_cache_safe = sampled_dcc_metadata && sampled_dcc_metadata_bytes &&
                         std::all_of(sampled_dcc_metadata,
                                     sampled_dcc_metadata + sampled_dcc_metadata_bytes,
@@ -6856,30 +6842,42 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                 // this first narrow path uncached rather than teaching the existing base-byte key
                 // an unsafe partial identity; staging materialization is still far cheaper than
                 // detiling and converting the 2x-larger FP16 base.
-                // PROSPER_DCC_CACHE_UPSIDE=1 (#3149) is a MEASUREMENT, not a fix, and must never
-                // become the default: it admits DCC-compressed surfaces to the image cache whose
-                // identity is base bytes only, so a surface whose metadata changes while its base
-                // bytes do not would serve stale pixels. It exists to answer one question -- is the
-                // real fix (folding the metadata plane into the cache identity) worth building? --
-                // because that fix is a correctness-sensitive change this code explicitly warns
-                // against making naively, and it should not be built on a guess about its payoff.
-                // #3149: a compressed surface is now cacheable because its DCC metadata plane is
-                // part of the cache identity and is validated alongside the base bytes -- the
-                // "unsafe partial identity" this code warned about is exactly what was added, so the
-                // warning is satisfied rather than bypassed. Requires a metadata range to validate:
-                // compression with no metadata address stays out, failing closed as before.
+                // #3149: the DCC-metadata test is dropped from this gate, and the reason is that
+                // it answers the wrong question. `dcc_cache_safe` asks whether the metadata plane
+                // reads as all-0xff, i.e. "nothing is actually compressed"; what decides whether an
+                // entry may be reused is whether the metadata can change the bytes this cache
+                // serves. On the sampled path it cannot: every consumer of the metadata plane
+                // reachable from here is `compute_sampled_dcc_fast_clear_rgba8`, and
+                // `!sampled_dcc_fast_clear` excludes every surface that reaches it -- so what a
+                // cacheable entry replays is detile + unpack of the BASE bytes and nothing else.
+                // (This branch is already inside `compute_sampled_guest_prepare_required`, i.e.
+                // !storage && !renderer_owned, so storage targets keep their own stricter gate.)
+                // Base bytes are precisely what the validation in `acquire_cached_image` already
+                // covers, so a compressed surface's cache identity was complete all along and the
+                // metadata test was only ever costing hit rate -- on Stray, most of the detiling
+                // work in the whole run. #3149 carries the census.
+                //
+                // This deliberately does NOT change how a compressed surface decodes -- only
+                // whether that decode's result may be reused. If reading base bytes without
+                // consulting a live DCC plane is wrong for some surface, it is equally wrong on
+                // both sides of this branch, and the cache serves the same pixels the uncached path
+                // would have uploaded. That is what makes the change pixel-neutral by construction,
+                // and what the frame A/B on #3149 measures rather than assumes.
+                //
                 // Kill switch, and it earns its place: this admits a whole class of surface to a
-                // cache it was previously excluded from, so a single variable must be able to
-                // restore the old behaviour exactly -- for bisecting a rendering report, and for the
-                // pixel A/B that justified the change in the first place.
-                static const bool dcc_cache_disabled =
-                    std::getenv("PROSPER_NO_DCC_IMAGE_CACHE") != nullptr;
-                const bool dcc_identity_tracked = !dcc_cache_disabled &&
-                    r->compression_enabled && r->metadata_addr &&
-                    prosper::gpu::gpu_capture_dcc_metadata_footprint(*r) > 0;
-                bi.cache_candidate = !sampled_dcc_fast_clear &&
-                    (dcc_cache_safe || dcc_identity_tracked) &&
-                    persistent_compute_image_enabled(sbytes, ComputeImageCacheClass::sampled);
+                // cache it was previously excluded from, so a single variable must restore the old
+                // behaviour exactly -- for bisecting a rendering report, and for that A/B. It gates
+                // the entire change, because the gate below is now the entire change.
+                // Designated, not positional: inserting a field into this struct must not be able
+                // to silently re-bind these arguments.  #3149 lost a round to exactly that hazard
+                // in ComputeImageCacheKey.
+                bi.cache_candidate = compute_sampled_cache_gate_candidate({
+                    .sampled_dcc_fast_clear = sampled_dcc_fast_clear,
+                    .dcc_cache_safe = dcc_cache_safe,
+                    .dcc_cache_disabled = dcc_cache_disabled,
+                    .persistent_enabled = persistent_compute_image_enabled(
+                        sbytes, ComputeImageCacheClass::sampled),
+                });
                 const VkFormat transfer_native_format =
                     compute_transfer_storage_vk_format(r->format, sampled_components);
                 const bool float32_uint32_transfer_alias =
@@ -6930,15 +6928,7 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                         r->layer_stride_bytes, r->layer_mip_offset_bytes,
                         r->mip_tail_offset, r->mip_tail_bytes,
                         r->mip_tail_x, r->mip_tail_y,
-                        static_cast<uint32_t>(image_format),
-                        // #3149: metadata plane joins the identity. Positional aggregate init --
-                        // these two MUST sit between vk_format and storage, matching the struct.
-                        r->compression_enabled ? r->metadata_addr : 0ull,
-                        r->compression_enabled
-                            ? static_cast<uint32_t>(
-                                  prosper::gpu::gpu_capture_dcc_metadata_footprint(*r))
-                            : 0u,
-                        bi.storage, r->in_mip_tail,
+                        static_cast<uint32_t>(image_format), bi.storage, r->in_mip_tail,
                         r->srgb, r->depth_compare};
                     // An ordinary native typed 2D image or native typed 3D volume is byte- and
                     // format-identical to the sampled upload that follows it. Borrow that retained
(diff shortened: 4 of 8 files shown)


mattias800 (commented):
## Independent review — PR #3150 @ 49fa635

**REJECTED** — fixable in an hour, not a rework. The perf win is real and well argued, but the
safety argument the PR and its new comments assert is not the one the code implements, and the new
check can silently disable itself. Three blocking items, all local to `acquire_cached_image` and the
admission predicate.

Verified as asked and found **correct**: the two positional aggregate-init sites (question 4), the
epoch fast path's conclusion (question 2, though not its stated reason — N9), uncompressed surfaces
(question 5), and the census mechanics (question 6). `check_diag_gates.py` at head:
`all checks passed`, rc=0, 570 files, 735 `PROSPER_*` variables, 108 findings / 108 baselined.
`check_cached_env.py` also green (5 pre-existing ordering notes, none new).

---

### B1 — the metadata check gates one of three skip proofs, so the stated guarantee does not hold

`prosper/frontends/shared/live/live_compute.cpp:2046`:

```
upload_skipped = submit_unchanged || watch_unchanged || exact_unchanged;
```

`metadata_unchanged` is conjoined into `submit_unchanged` only (`:2018-2022`). The other two
disjuncts examine the **base range alone**:

* `watch_unchanged` (`:2027`) queries `cached.write_watch`, which is created at `:2038` as
  `GuestWriteWatch::create(key.gpu_addr, key.guest_bytes)` — the base, never the metadata plane.
* `exact_unchanged` (`:2042-2045`) is `memcmp(cached.source_snapshot, source, key.guest_bytes)` —
  the base, never the metadata plane.

So in exactly the scenario the PR body names — *"A surface whose metadata changes while its base
bytes do not now invalidates"* — `metadata_unchanged` goes false, `submit_unchanged` goes false, and
control falls into two proofs that both answer "unchanged" and set `upload_skipped = true`. The new
gate does not invalidate; it converts a fast skip into a slower skip with the same answer. The
comment at `:2010-2013` states the opposite.

This is reachable on the very surface the PR targets, not hypothetically. `may_promote_write_watch_before_exact`
passes `defer_min_bytes = 1` (`:1700-1703`), so a 66 MiB source arms a watch once
`write_watch_stable_validations >= 3`, and `WriteWatchPromotionBudget::try_consume`
(`prosper/frontends/shared/texture/write_watch_policy.hpp:36-50`) explicitly lets the first
promotion in a submit exceed the 8 MiB budget. Three stable frames and the hot path is
`watch_unchanged`.

Two further asymmetries make the metadata half strictly weaker than the base half:

* It consults **only** the intra-submit GPU journal. `guest_gpu_writes_since`
  (`prosper/src/gpu/execute/gpu_executor.cpp:11403-11409`) returns `Unknown` — not `Unchanged` — when
  the journal is unarmed on this thread or the snapshot carries a different `submit_serial`. That is
  correct fail-closed behaviour *for that function*, but here `Unknown` makes `metadata_unchanged`
  false, which lands straight in the base-only paths above and skips anyway. The base range has
  three independent proofs; the metadata plane has one, and its failure mode is to hand the decision
  back to the proofs that ignore it.
* A guest **CPU** write to the metadata plane is invisible to it. There is no page watch and no
  exact snapshot for that range at all.

**What I think is actually going on, and why it changes the fix.** No `cache_candidate` sampled path
reads the metadata plane. The only consumer is `compute_sampled_dcc_fast_clear_rgba8` (`:117-144`),
and `bi.cache_candidate` requires `!sampled_dcc_fast_clear` (`:6880`). The materialization is
`detile_surface` plus a format conversion over base bytes (`:7749-7789`) — a pure function of base
bytes and layout. So the change is probably sound *in effect*, but for a different reason than the
one given: **the produced image does not depend on the metadata plane**, not "the metadata plane
joined the identity". I am not asking you to pick the one I prefer — I am asking that the code state
the true one:

* If metadata really is part of the identity, gate all three:
  `upload_skipped = (submit_unchanged || watch_unchanged || exact_unchanged) && metadata_unchanged;`
  with `Unknown` failing closed. That is one line and it makes `:2010-2013` true.
* If the real argument is decode-independence, say that at `:2010` and at `:6866-6870`, drop the
  metadata range from the validation, and keep only whatever the key needs to disambiguate entries.

What must not merge is a `## Ruled out`-grade comment on a cache identity asserting a guarantee the
function does not provide. This repository's own charter is explicit that a wrong claim carrying a
`file:line` is the most contagious artifact it produces, and a future reader will reasonably build on
`:2010-2013` without re-deriving it.

### B2 — the new bound can truncate to zero, which disables the check it guards

`:6878-6879` admits on the untruncated 64-bit value:

```
r->compression_enabled && r->metadata_addr &&
prosper::gpu::gpu_capture_dcc_metadata_footprint(*r) > 0;
```

but the key stores `static_cast<uint32_t>(...)` at `:6939` and `:894`, unguarded. If the footprint
exceeds `UINT32_MAX` and its low 32 bits are zero, `dcc_identity_tracked` is true while
`key.metadata_bytes == 0`, and `metadata_unchanged` short-circuits to `true` at `:2014` — the check
silently switches itself off for a surface admitted *because* it is compressed. A non-zero
truncation is the milder version: validating a wrong, smaller window.

It is reachable arithmetically. `gfx10_dcc_metadata_bytes` (`prosper/src/gpu/texture/tile.cpp:55-90`)
multiplies by `depth`, which `dcc_metadata_footprint`
(`prosper/src/gpu/capture/gpu_capture.cpp:757-759`) supplies as `max(r.depth, 1)` for `img_dim == 2`,
while the 512 MiB base bound above it uses `tiled_surface_bytes(r->width, r->height, ...)` for a
tiled volume (`:6788-6790`) — `depth` never enters that bound. 3840x2160, bpe=8 gives 72 blocks per
slice; `depth = 2^20` gives `72 * 2^32` bytes, whose low 32 bits are exactly zero.

The fix is already sitting 70 lines above: `sampled_dcc_metadata_bytes` is computed at `:6808-6812`
behind `metadata_bytes && metadata_bytes <= SIZE_MAX && metadata_bytes <= UINT32_MAX`. Reuse it —
it also carries the readability proof (`:6810-6820`), which `dcc_identity_tracked` currently does not
require at all — instead of recomputing the footprint a third time.

### B3 — a cache-identity change with no test, and a snapshot null that cannot show its lever moved

Nothing in the diff is tested. This file already extracts precisely this class of decision as a pure
predicate for exactly this reason — `compute_storage_cache_gate_candidate` and
`compute_storage_post_writeback_promotion_candidate` are called ten lines away at `:7168-7174`, and
`should_promote_write_watch` / `compressed_source_authority` follow the same pattern. Please give
`dcc_identity_tracked` the same treatment, with arms for: (a) uncompressed → zero metadata identity
and a key byte-identical to master's; (b) a footprint above `UINT32_MAX` → declined (B2); (c)
metadata changed while base unchanged → **not** skipped. Arm (c) is the one that would have caught B1,
and it is the arm whose absence let the PR body's central claim go unchecked.

On the snapshot evidence: `messenger-scene` and `dead-cells-gameplay` are the right instinct — a
shared cache does deserve a cross-title check — but neither is shown to admit a single surface
through `dcc_identity_tracked`. "Both OK" is therefore equally consistent with the new path never
executing on either route, which is the charter's clean-zero rule verbatim. Please quote a counter
from at least one guarded route proving the path ran (`g_sampled_image_upload_skips`, or a
`PROSPER_COMPUTELOG` `persistent sampled image ... upload-skipped=1` line on a `compression_enabled`
binding). Without that the two guards are a null from an instrument not shown able to produce a one.

---

### Non-blocking

**N1 — the key change is not covered by the kill switch, and it perturbs two cross-descriptor
lookups.** `storage_image_cache_key` (`:877-898`) now folds metadata into the **storage** key, and
that key is built from a *different descriptor* than the entry it looks up in two places: the
compute-transfer seed borrow at `:6953` builds it from the SAMPLED `*r`, and the graphics import at
`:10063` builds it from the graphics sampled descriptor — both matching entries retained from a
STORAGE descriptor at `:7165`. The two descriptors need only disagree on `compression_enabled` or
`metadata_addr` for those borrows to start missing silently: a perf regression that ctest and both
snapshot guards are blind to, on the GTA V alias retries those paths exist for. Compounding it,
`PROSPER_NO_DCC_IMAGE_CACHE` gates only `dcc_identity_tracked` (`:6877`) and **not** the key change,
so the switch cannot bisect this — which contradicts its own comment at `:6871-6875` ("a single
variable must be able to restore the old behaviour exactly"). Simplest resolution: pass `0ull, 0u`
positionally in `storage_image_cache_key`; storage entries never needed the metadata identity, and
the sampled key is where the disambiguation is wanted.

**N2 — `ComputeImageCacheKeyHash` was not updated** (`:859-874`). Not a correctness bug — equal keys
still hash equal, and `operator== = default` covers the new fields — but two surfaces differing only
in metadata plane now collide into one bucket. Two `mix()` calls.

**N3 — `PROSPER_DCC_CACHE_UPSIDE` does not exist.** `:6859` is the only occurrence anywhere in the
tree. The measurement-stage comment was left stacked above the fix-stage comment and the two
contradict each other ("must never become the default" / then the fix landed). A reader greps for
it and finds nothing. Delete `:6859-6865`.

**N4 — "restores the old behaviour exactly" is not exact.** With the switch set, a surface that was
already cacheable under `dcc_cache_safe` (all-`0xff` metadata) still carries metadata in its key and
still pays the `metadata_unchanged` query. Behaviourally harmless; worth wording precisely, since
the A/B leans on it.

**N5 — the frame A/B is not a controlled comparison.** The two runs are at *different points in a
timed sequence* (Annapurna vs BlueTwelve) precisely because one is faster, so a ~3% pixel delta
cannot separate "no corruption" from "corruption plus phase shift" — the delta is dominated by the
variable under test. Anchor both arms on the same frame ordinal
(`PROSPER_CAPTURE_SCREENSHOT_AT_FRAME`) or the same pad-flip ordinal and re-shoot; that turns the
strongest claim in the PR ("rendering verified, not assumed") into one the reader can check. The
unsound-upper-bound arm (22.11 vs 19.93) is a genuinely good control and is worth keeping.

**N6 — `TileCensusValue::who` is last-writer-wins** (`tile.cpp:1229`, `v.who = g_tile_census_tag;`).
One geometry reached from both a tagged and an untagged site reports whichever ran last, and the PR
body's own headline row shows the consequence: it reports `compute`, but on this branch a plain 2D
tiled FP16 sampled upload of that shape goes through the `else if (r->tile_mode)` at `:7749`, which
is tagged `smpl-upl`. So either that census predates the scope or the attribution was clobbered —
either way the `who` column does not establish that the 98.9% is the upload this fix caches. The
perf A/B carries the result independently, so this costs the PR nothing except its causal narrative;
key on `(op, geometry, who)` rather than overwriting, and the column becomes evidence.

**N7 — census mechanics otherwise check out.** Default-off is a cached `getenv` behind a
function-local static plus a branch (`tile.cpp:1197-1201`); the periodic dump releases
`g_tile_census_mx` at the end of the inner scope before `tile_census_report()` re-takes it, so no
recursion and no deadlock; `who` only ever holds one of three string literals. Two nits:
`TileCensusHash` shifts by 33 (`tile.cpp:1189`), UB if `size_t` were 32-bit — harmless on this
target, free to fix; and `TileCensusScope`'s header comment does not state the literals-only
invariant that keeps `who` from dangling, which is the one thing a future caller could break.

**N8 — neither new variable is registered in `prosper/tools/AGENTS.md`**, which the charter names as
where every `PROSPER_*` graphics diagnostic lives. `PROSPER_TILECENSUS` in particular is a tool
another lane would want to find by grepping the docs, not the source.

**N9 — the epoch fast-path comment is right by accident** (`:1998-2000`). The conclusion holds, but
not for the stated reason: `image_validation_epoch = ++ctx.image_validation_clock` at `:5043` is
incremented **per dispatch item**, not per submit, so a reused epoch can only ever be a second
binding naming the same image inside one `execute_item` — the case `CachedComputeImage::validation_epoch`'s
own comment at `:907-910` describes, where nothing can have changed in between. The appeal to "a
metadata write inside the submit bumps the journal that `metadata_unchanged` below consults" is not
the reason it is safe, and it is wrong in the direction that matters: B1 shows a bumped journal does
not stop the skip. Restate it as the per-item argument.

---

### What I confirmed correct

* **Positional aggregate init (question 4).** Both sites walk the struct's 25 members in order with
  the new pair between `vk_format` and `storage`: `storage_image_cache_key` at `:880-897` and the
  sampled key at `:6924-6942`. Checked member by member, not by eye-balling the diff. No third
  positional site exists — `:2871`, `:2984` and `:2989` are `{}` value-init, and `:6953` / `:10063`
  route through `storage_image_cache_key`. The `0ull` on the false branch keeps the ternary's common
  type at `uint64_t`, so no narrowing.
* **Uncompressed surfaces (question 5).** Yes, byte-for-byte: `compression_enabled ? … : 0` at both
  sites means the key is identical to master's, and `key.metadata_bytes == 0` short-circuits
  `metadata_unchanged` at `:2014` before any journal query. See N4 for the compressed-but-already-
  cacheable class, which is the one place the claim needs qualifying.
* **The performance result.** The mechanism is clear and the number is credible: 2,199 detiles of a
  66 MiB surface removed by cache admission, with the sound/unsound bracket (19.93 vs 22.11) showing
  how much of the available win the validation costs. That is the right control and I have no
  reservation about the perf half of this PR.

Happy to re-review promptly on a new head. B1 and B2 are each a few lines; B3 is the one that will
take real work, and it is also the one that would have surfaced B1 without me.

mattias800:
## Author verification — 0d7a0d6e

**The rework takes B1's decode-independence route**, which the review identified as the true
argument. That moots B2, N1, N2, N4 and N9 along with the key change they concerned: metadata is out
of `ComputeImageCacheKey` and out of `acquire_cached_image` entirely, so the two cross-descriptor
borrow lookups are untouched and `PROSPER_NO_DCC_IMAGE_CACHE` now gates the whole change.

I did not take the claim on trust — `sampled_dcc_metadata` could have been one of several channels.
Checked by a second route: every read of `bi.dcc_metadata` and `r->metadata_addr` in the file is a
storage-target path, and the gate sits inside `compute_sampled_guest_prepare_required`
(`!storage && !renderer_owned`), so storage keeps its stricter gate structurally.

### Build and tests
- `cmake --build` clean, **0 errors**
- **ctest 315/315 passed, 0 failed** (4 pre-existing `-DGAME_DUMP` skips)
- `check_diag_gates.py`: `all checks passed`, 570 files, 735 `PROSPER_*` variables, 108/108 baselined
- `check_cached_env.py`: green (5 pre-existing ordering notes, none new)
- docs gate green; `diff --check` clean

### B3 — mutation-verified, and I am reporting the split honestly
Reverting `compute_sampled_cache_gate_candidate` to master's gate and rebuilding the real test file:

```
MUTANT rc=1   1 failure:  FAIL compressed sampled surface is a cache candidate (#3149)
CLEAN  rc=0   0 failures
```

**1 of 6 arms fails without the fix.** The other five pin behaviour that must not change. The three
arms you asked for describe the rejected design (no key change, no footprint cast, metadata
irrelevant by construction), so I substituted arms that carry the new one — flagging the swap rather
than quietly satisfying the letter of the request.

### Performance — +66%, one variable
`prosper-app`, headless, same 110 s, no pad (the splash is the subject):

| arm | fps mean | tiling bytes |
| --- | --- | --- |
| default | **62.24** (109 windows) | **3,176 MiB** |
| `PROSPER_NO_DCC_IMAGE_CACHE=1` | **37.42** (65 windows) | **246,977 MiB** |

Four independent route/harness combinations agree on the census gap (77.8x, 158x, 91x, and the
title-screen run), so it is not an artifact of one route.

### N6 was load-bearing, not cosmetic
With `who` in the key the causal claim is finally checkable, and it holds:

| row | pre-#3149 | this PR |
| --- | --- | --- |
| `smpl-upl detile_surface 3840x2160 bpe=8` | 3,860 calls / 244,266 MiB (98.9%) | **5 calls / 316 MiB** |
| `compute tile_surface 3840x2160 bpe=8` | 16 | 16 |
| `stor-seed detile_surface 3840x2160 bpe=8` | 9 | 9 |

772x on the tagged row, every other row unchanged. Under last-writer-wins those three collapsed into
one, which is exactly why the previous revision mis-attributed it to `compute`.

### B3's clean zero — you were right, and the answer is negative
Instrument first (Stray): 23,426 persistent sampled images, **8,235 `compressed=1 dcc-safe=0`, all
`upload-skipped=1`**, on `guest=66846720` — precisely the census geometry, from an unrelated
instrument.

Then the guarded route, driving `snapshot.py`'s own `capture_content()` so only the trace differs:

```
messenger-scene: 2,452,657 log lines, 222,908 dispatches, 668,736 bindings
                 persistent sampled images: 0
```

Not a void run. Every binding is `writeback binding=` — the storage path. The sampled cache minimum
is 1 MiB (`live_compute.hpp:38`) and this title's sampled inputs never reach it. **So this guard
cannot validate this change**, and I would rather say that than let "both guards pass" carry weight
it has not earned.

### N5 — original A/B withdrawn, replaced with a controlled one

| comparison | mean abs delta | px >8 |
| --- | --- | --- |
| within-arm A1 vs A2 (noise floor) | **0.106** | 0.40% |
| cross-arm A1 vs B1 | **0.100** | 0.41% |
| full-res closest pair | **0.000** | 0.00% |

At or below run-to-run variation; closest pair byte-identical at 4K. The comparator's first positive
control **failed** (corrupted frames scored 1.05x) because the capture set was entirely black and the
defect was inexpressible on it; it now moves 8.14x on a valid perturbation and refuses a verdict when
an arm is blank.

### Guards and scope
`messenger-scene` **OK**, `dead-cells-gameplay` **OK**, rc=0 — no cross-title regression, read with
the caveat above. Through to the title screen under `prosper-app`: **64.78 fps mean over 285 s**,
`smpl-upl` still 5 calls, so this is not splash-only.

### Two things I changed that you did not ask for
- **A dead-work regression in my own rework.** Dropping the metadata test left the all-`0xff` scan
  running per binding per dispatch with nothing consuming it — same class as B5 on #3147, on the path
  this PR shortens. It now runs only under the kill switch or the trace.
- **A harness artifact I nearly filed as a second bottleneck.** The title phase looked writeback-bound
  at ~41% under `tools/screenshot`; that harness forces a readback and inflates tiling **25x**
  (158,381 vs 6,254 MiB). No real bottleneck, no issue filed.

mattias800 (commented):
## Independent re-review — PR #3150 @ 0d7a0d6e

**APPROVED.** The rework takes the route B1 offered and the result is a strictly smaller, strictly
better-argued change: master's cache key is back byte-for-byte, `acquire_cached_image` is untouched,
and the entire diff to the decision path is one predicate. B1, B2, B3 and N1-N9 are all addressed;
I re-derived the load-bearing ones rather than reading the disposition table. Eight non-blocking
items below, none of which needs a new head — two of them (**N1**, **N7i**) I would land before
merging because they are edits to text, not code.

I did not rebuild or re-run ctest; per the charter that is the author's half. Everything below I
checked in a fresh worktree at `0d7a0d6e`.

---

### What I verified independently

**1. The central claim — no cacheable sampled decode reads the metadata plane. TRUE, by three
routes, and one of them is stronger than the argument in the PR.**

* Every use of the derived `sampled_dcc_metadata` in `prosper/frontends/shared/live/live_compute.cpp`
  is `:6784-6797` (resolution), `:6806` (the fast-clear probe) and `:7604` — and `:7604` sits under
  `if (needs_sampled_upload && sampled_dcc_fast_clear)` at `:7599`. `cache_candidate` requires
  `!sampled_dcc_fast_clear`, so the one materialization consumer is unreachable for anything the
  cache can hold.
* The channels the review asked me to check separately are clean. `bi.dcc_metadata` is assigned only
  under `if (bi.storage && r->compression_enabled)` (`:5985`); alias inheritance at `:6477` cannot
  carry it to a sampled binding because `same_view` requires `prior.storage == bi.storage`
  (`:6444`). `r->metadata_addr` appears only in the fast-clear helper, the storage writeback
  (`:9498`), the metadata resolution and a trace line. `alpha_is_on_msb` and `color_transform` occur
  only in the fast-clear helper and the alias identity. The decode source is
  `resource_bytes_for` (`:5440`), which returns base bytes and nothing else.
* **The stronger route, which the PR does not use:** `gfx10_dcc_fast_clear_rgba8`
  (`prosper/src/gpu/texture/tile.hpp:243`) is the *only* DCC decode function in the tree — the sole
  other `gfx10_dcc*` symbol is `gfx10_dcc_metadata_bytes`, a size computation. prosper has no DCC
  decompressor at all. So a compressed surface's base bytes decode identically with or without a
  live plane as a property of the codebase, not merely of this file's call graph, and the next
  reader can re-establish it with one grep. Worth putting in the comment.

**2. The fast-clear transition argument. Both halves hold, and the second half is stronger than
stated.** `sampled_dcc_fast_clear` is recomputed per binding per dispatch at `:6801-6807`, upstream
of the gate at `:6874`, and `cache_candidate` gates the entire lookup — the seed borrow and
`acquire_cached_image` are both inside `if (bi.cache_candidate)` at `:6920`. Into a fast clear:
the gate re-evaluates false and the cache is not consulted. Out of one: I checked the retain side
too, which the PR body does not mention — sampled retention at `:8987-8990` skips
`!image.cache_candidate`, so a fast-clear image is never *entered* into the cache either, not merely
never served from it. Entry and exit are both closed.

I also worked the case the PR does not name: an entry admitted, then the surface spends N dispatches
as a fast clear with base bytes untouched, then returns. The stale entry survives in the map and is
served on return. That is correct, because the pixels it serves are the detile of the same base
bytes the uncached path would have decoded — which is exactly the PR's equivalence claim, applied to
a case it did not enumerate.

**3. The new conditional scan. No observable change; the trace reports truthfully.** Within the
sampled branch `dcc_cache_safe` is read at exactly two places — the gate (`:6876`) and the trace
(`:7005`). It is not passed to `record_sampled_gates`, whose signature I read at `:3648-3655`
(the storage recorder at `:3702` is the one that takes it). The `:7145` declaration is a separate
`const bool` in the storage branch, computed unconditionally from `bi.dcc_metadata`, untouched.
The predicate `(safe || !disabled)` makes `safe` irrelevant whenever `disabled` is false, so
skipping the scan on the default path cannot move the gate; when the switch is set or `trace` is on
the scan runs, so neither consumer ever reads a fabricated value. Kill switch set reduces exactly to
master's `!fc && safe && persistent`. `persistent_compute_image_enabled` is now evaluated eagerly
rather than short-circuited — it is pure (function-local statics over `getenv`), so that is free.

Non-obvious consequence worth naming approvingly: without the `|| trace` clause, `dcc-safe=` on the
new trace line would have printed 0 for every compressed surface *by construction*, and the 8,235-row
census would have been vacuous. The clause is what makes that evidence mean anything.

**4. B3's substitution is legitimate, and the arms discriminate harder than claimed.** The three
arms I asked for do describe the rejected design — (a) and (b) assert against a key and a cast that
no longer exist, and (c) asserts the *opposite* of the new intended behaviour, since a metadata-only
change now must not invalidate. Substituting was the right call and flagging the swap was the right
way to do it.

I re-derived the mutation rather than accepting it: reverting `compute_sampled_cache_gate_candidate`
to `!fast_clear && dcc_cache_safe && persistent_enabled` and rebuilding the real test file gives
rc=1 with exactly one failure, `compressed sampled surface is a cache candidate (#3149)`. Clean
build rc=0. That is the claim, confirmed. I then ran four more single-term mutants, and every one is
caught by a *different* arm: kill-switch-is-noop -> the pre-#3149-rejection arm; switch-disables-all
-> the uncompressed arm; fast-clear-ignored -> both fast-clear arms; persistent-ignored -> the
persistent arm. The six arms pin the whole three-term predicate, not just the one bit that moved.

**5. Key byte-identical to master. Yes.** `git show origin/master:...live_compute.cpp | sed -n '825,900p'`
against the same range at head diffs empty — struct, `ComputeImageCacheKeyHash` and
`storage_image_cache_key` are all master's. The sampled key at `:6921-6931` is likewise unchanged in
`git diff origin/master HEAD`. Twenty-three initializers against twenty-three members at both sites.

**6. The census.** `who` in the key is correct and the report reads it from `.first` now.
The 64-bit accumulator removes the shift-by-33 UB; `std::hash<const void*>` on a `const char*`
converts implicitly and the result widens into `uint64_t` before any multiply, so nothing depends on
`size_t` width. The periodic dump still accumulates: `g_tile_census` is never cleared, only
`since`/`since_bytes` reset, both under `g_tile_census_mx`, and `tile_census_report()` is called
after the lock is released and re-takes it — no recursion.

I also checked the arithmetic behind the causal claim, which is the point of N6: 3860 x 63.28 MiB =
244,262 MiB against the reported 244,266, and 5 x 63.28 = 316. 246,977 / 3,176 = 77.76x. The census
rows are internally consistent and the `smpl-upl` attribution now actually carries the claim.

**7. Repository gates**, run at head in my own tree: `check_diag_gates.py` `all checks passed`
(570 files, 735 `PROSPER_*`, 108 findings / 108 baselined, no new unbaselined finding for the new
`getenv`); `check_cached_env.py` green with the same 5 pre-existing ordering notes;
`check_numbered_table.py` over every tracked `.md` rc=0; `git diff --check` clean. The test file
compiles and runs standalone.

---

### Non-blocking

**N1 — an arithmetic claim in the PR body is wrong, and it is the sentence that joins the two
instruments.** `guest=66846720` is **not** "exactly 3840x2160x8" — that product is 66,355,200. It is
3840 x **2176** x 8, the Sw64KbRX tiled footprint with height padded 2160 -> 2176.
`tiled_surface_bytes` routes mode 27 through `is_64kb_mode` to `sw64kb_tiled_bytes`
(`prosper/src/gpu/texture/tile.cpp:1159-1166`), and the trace prints `sampled_guest_need`, which for
this shape is that tiled footprint (`live_compute.cpp:6764-6765`). The identification is still right — it
is the same surface the census row names — but a reader who multiplies gets a different number and
concludes the two instruments disagree. Please fix it in the PR body and the verification comment
before merging.

**N2 — the gate widening also opens the compute-transfer seed borrow, which the soundness argument
does not cover.** `cache_candidate` gates `borrow_cached_image_for_compute_transfer` at `:6946`
as well as `acquire_cached_image`. Before this change a genuinely compressed sampled descriptor never
reached that borrow (it needed `dcc_cache_safe`); now every compressed one does. That path does not
reuse "this decode's result" — it seeds the image from a *retained storage image*, a different
source entirely, so the paragraph beginning "This deliberately does NOT change how a compressed
surface decodes" is narrower than what the gate now unlocks.

I convinced myself it is safe and am not blocking on it: `borrow_cached_image_for_compute_transfer`
(`:2163-2212`) requires `compute_transfer_valid && content_valid` plus a journal, watch or exact
proof over the **base** range since the producing writeback, and the storage entry could only have
been authorized at `:9678-9679` through the storage gate, which still demands an all-`0xff` plane. A
divergence needs the plane to become non-`0xff` while base bytes stay byte-identical, which for a
non-fast-clear state is not physically producible, and fast clears are excluded upstream. One
sentence at the gate saying that would save the next reader the walk.

**N3 — extend the invariant note in the other direction, and name the seam.** The header comment's
"if a sampled decode ever starts consulting the DCC metadata plane, this gate has to consult it too"
is exactly the right instinct. The symmetric risk is a decode that starts *declining*.
`prosper/src/gpu/resources/compressed_source_authority.hpp` is a designed shared decision whose own
header says the compute backend "asks before guest preparation" and that a DCC base "decodes to
plausible garbage, which is strictly worse than declining" — and `sampled_source_decision` has **no
production caller today**; it appears only in its own `.cpp` and
`prosper/tests/gpu/resources/test_compressed_source_authority.cpp`. When it is wired, the surfaces
this PR admits are precisely the ones it would decline (`DeclinedUnsupportedMetadataState`), and this
gate must move with it. Pre-existing, not caused here — but this comment is the place someone will
be looking.

**N4 — one clause in the gate comment is looser than the code.** "every consumer of the metadata
plane reachable from here is `compute_sampled_dcc_fast_clear_rgba8`, and `!sampled_dcc_fast_clear`
excludes every surface that reaches it": the *probe* at `:6801-6807` reaches the helper for every
`compression_enabled` sampled surface. What `!sampled_dcc_fast_clear` excludes is the
*materialization* at `:7599-7605`. The operative sentence right after it — "what a cacheable entry
replays is detile + unpack of the BASE bytes and nothing else" — is exact and is the one that
matters, so this is wording. I raise it only because a rejected revision died of a comment that
asserted more than the code did, and this comment will be quoted.

**N5 — one of four predicate corners is unasserted.** Over `(dcc_cache_safe, dcc_cache_disabled)`
with `fast_clear=false, persistent=true`, the six arms cover (F,F), (F,T) and (T,T) but never (T,F) —
the ordinary uncompressed default path, the most common case in production. Completeness nit only:
I mutated the predicate five ways and every single-term mutant is caught by an existing arm.

**N6 — census nits.** `tile_census_note("detile_volume", width, height * depth, …)`
(`tile.cpp:1807`, and `tile_volume` at `:1828`) multiplies in `uint32_t`; a large volume wraps and
aliases into another row. Diagnostic-only. Separately, `TileCensusScope` has an implicit copy
constructor — a user-declared destructor does not suppress it — so a copy double-restores `prev`.
Harmless as written; one `= delete` makes it unbreakable, which matters more for a type now exported
in a widely included header.

**N7 — two charter items.**
  * **(i)** No `## Ruled out` row. The fact this PR established — *a DCC-compressed sampled
    surface's cache identity does not include its metadata plane, because no cacheable sampled decode
    reads it; folding it into the key is both unnecessary and, as attempted, ineffective* — cost a
    full rejected revision to derive and is cross-title. The charter: "A PR that falsifies a
    hypothesis records it in the relevant `## Ruled out` section before merging… A falsification
    that lives only in a PR body or an issue comment is one the next agent will re-derive at full
    cost." `prosper/docs/RESOURCE_BINDING.md:132` is the natural home. The gate comment is excellent
    but is not where somebody forms a hypothesis. Two lines; please land it with the merge.
  * **(ii)** No `BLOG.md` entry. There is no screenshot so the hard rule does not bind, but a
    measured +66% on Stray with a named cause is exactly the "finding with no picture is welcome
    too" case the charter asks for.

**N8 — the pixel A/B should not be quoted as independent evidence.** Pairing frames by closest match
is minimum-seeking: for each A frame you select the B frame that minimises the statistic you then
report, and "full-resolution re-check of the *closest* cross-arm pair" selects the best case twice
over. The failure mode of a cache-staleness bug is "this frame looks like a slightly earlier frame",
which is precisely what closest-match pairing absorbs. The within-arm floor is computed the same way,
so the comparison is at least like-for-like, and the structural argument is what actually carries
this PR — I would just not let 0.100-vs-0.106 stand as a second, independent proof. The comparator's
positive control failing on an all-black capture set, being diagnosed rather than explained away, and
the comparator now refusing a verdict on a blank arm, is the best methodological work in the PR and
belongs in the record exactly as written.

Minor, in the same section: the `messenger-scene` zero is given two different reasons — every
binding is `writeback binding=` (no sampled bindings at all) and the 1 MiB sampled minimum
(`live_compute.hpp:36-40`). Only one can be the cause. The conclusion — that
this guard cannot validate this change — stands on either, and stating it against the author's own
interest was the right call.

---

### What this revision got right that is worth saying out loud

The dead-work catch was the author's own, on the author's own rework, and it is the kind of thing
that normally ships: dropping the metadata test left a full pass over the metadata plane running per
binding per dispatch on the exact hot path the PR exists to shorten, consumed by nothing. Finding
that, and separately declining to file the `tools/screenshot` readback artifact as a second
bottleneck after tracing it to the harness, are both the instrument-not-the-subject discipline
working. So is reporting that only 1 of 6 arms discriminates instead of letting "the tests would fail
without the fix" stand for all six, and so is concluding against interest that `messenger-scene`
cannot validate this change. The perf result is real, the mechanism is named, the census now carries
the causal claim it previously could not, and the safety argument is now the one the code implements.

mattias800:
## Second-review findings addressed — e2058dc3

All eight non-blocking findings are in. Two were corrections to my own evidence rather than to the
code, and both were the same failure shape: a true conclusion reached by a route that does not
establish it.

**N1 was a real error and I want it recorded as one.** `guest=66846720` is 3840 x **2176** x 8 — the
tile-height-padded footprint — not 3840x2160x8, which is 66,355,200. That sentence was the join
between the census and the trace counter, so anyone who actually did the multiplication would have
concluded my two instruments contradicted each other. I asserted a computation I had not performed,
in the one place where being checked was most likely.

**Same class, second instance:** I gave `messenger-scene`'s zero two incompatible reasons. Only one
holds — its compute bindings log `cls=0 ... dims=0x0`, i.e. buffers rather than images, so the
sampled cache is never reached at all. The 1 MiB sampled minimum I also cited is irrelevant, because
execution never gets that far. Corrected in the body.

**On N8, you are right and I have demoted the claim.** Closest-match pairing is minimum-seeking, and
"full-resolution re-check of the *closest* pair" selects the best case twice; worse, the failure mode
of a staleness bug is exactly what that pairing absorbs. It now reads as corroboration of the
structural argument rather than as a second independent proof.

**Your form of the core argument is stronger than mine and is what I recorded.**
`gfx10_dcc_fast_clear_rgba8` being the *only* DCC decode function in the tree makes
metadata-independence a property of the codebase rather than of one call graph. That is the sentence
now in `docs/RESOURCE_BINDING.md`'s `## Ruled out` (N7i), alongside an explicit "do not re-derive the
metadata-in-the-key design" and why it failed twice over.

Code changes: N2 (gate comment now covers the compute-transfer seed borrow and why it is separately
safe), N3 (invariant extended to the declining direction, naming `sampled_source_decision` as the
unwired seam that must move with this gate), N4 (probe vs materialization stated precisely), N5 (the
uncompressed default corner asserted), N6 (volume census rows saturate instead of wrapping;
`TileCensusScope` copy/assign deleted). N7ii: `BLOG.md` entry added.

**Verification at this head:** build clean, **ctest 315/315 passed, 0 failed**. Docs gate green
(`RESOURCE_BINDING.md`: 1 table, 17 rows, unbroken). `diff --check` clean. The diff since your
approved head is comments, docs, one added test arm, and the two census hardening changes — no change
to the admission predicate you approved.

mattias800 (commented):
## Re-confirmation — PR #3150 @ e2058dc3

**APPROVED.** My approval at `0d7a0d6e` **carries to `e2058dc3`**, established by content rather than
assumed. Two new non-blocking items below, one of which is a genuine overclaim you asked me to look
for (**N3**).

### The detachment check, mechanically

* **The admission predicate is byte-identical.** Extracted the body of
  `compute_sampled_cache_gate_candidate` from both revisions and diffed: empty.
* **`live_compute.cpp`'s delta is comments only.** Filtered every `+`/`-` line in
  `git diff 0d7a0d6e e2058dc3` for that file down to non-comment, non-blank: nothing remains.
* **Exactly three non-comment code changes exist**, as you said — `census_rows`, the deleted
  copy ctor/assignment, one test arm. Nothing else in the diff is code.
* Re-ran at head: `check_diag_gates` green (570 files, 735 vars, 108/108 baselined),
  `check_cached_env` green, numbered-table gate rc=0, `diff --check` rc=0, the selector test
  **45 PASS / 0 FAIL**, and the master-gate mutant still fails **exactly one** arm
  (`compressed sampled surface is a cache candidate`).
* Spot-checked the N1-class correction you made unprompted: `ResourceClass::ConstantBuffer` is
  ordinal **0** (`shader_resources.hpp:204-205`), so `cls=0 ... dims=0x0` does read as buffers rather
  than images, and the binding-kind reason is the one that holds. Correcting that instead of leaving
  two incompatible explanations standing was the right call.

### Q1 — is clamping right for the census bucket, versus dropping the row?

**Clamping is the right choice**, and better than both alternatives: a wrapped value silently
*becomes* a plausible real geometry (2^32+2160 rows reads as 2160), while the clamp produces an
obviously-bogus sentinel; and a dropped row is silent, where the charter prefers an instrument that
shows its own invalidity.

The one thing the comment does not say: the clamp fixes the **key** and leaves the **value**
fabricated. `tile_census_note` derives `v.bytes += uint64_t(w) * h * bpe` from the same clamped `h`,
and `tile_census_report` both ranks rows by bytes and computes every percentage from a total
containing it — so a saturated row would be honest about its identity and badly wrong about its
weight, and would sort straight to the top of the report. I would still leave it: it is unreachable
in practice (every caller is bounded far below 2^32 rows — live compute skips at
`kMaxComputeImageBytes` / `UINT32_MAX` before the decode). If you want it airtight, pass the true
64-bit byte count into `tile_census_note` rather than re-deriving it from `h`. Non-blocking.

Confirmed `census_rows` sits inside the anonymous namespace (`tile.cpp:1179-1268`), so no linkage
surprise. On the deletion: declaring the copy constructor deleted also suppresses the implicit
*move* constructor. Nothing moves a `TileCensusScope`, so that is correct and intended — worth
knowing rather than fixing.

### Q2 — do the N2 and N3 comments claim more than the code supports?

**N2 is accurate, with one countable imprecision.** I re-walked every `bi.cache_candidate` use. Both
named consumers are real and both sit inside `if (bi.cache_candidate)` at `:6935`, and I
re-verified the safety reason independently: `borrow_cached_image_for_compute_transfer`
(`:2163-2212`) demands `compute_transfer_valid && content_valid` plus a journal/watch/exact proof
over the base range, and the storage entry could only have been authorized at `:9678-9679` through
the storage gate. But `cache_candidate` also gates sampled **retention** at `:9004`, so the literal
"gates TWO things" is falsifiable with one grep. Nothing is hidden — retention is the write half of
the reuse mechanism the paragraphs above already cover — so drop the numeral or add
"(and retention at `:9004`)".

**N3 does overclaim, in the narrow way that matters.** "the surfaces admitted here are precisely the
ones it would return `DeclinedUnsupportedMetadataState` for" is not what `sampled_source_decision`
does. Reading `compressed_source_authority.cpp`, the decline reason is selected by the
`MetadataProof`, and `DeclinedUnsupportedMetadataState` is only the `Absent`/`NotDecompressed`
branch: `UnknownKind` gives `DeclinedUnknownMetadataKind`, `NoExactFootprint` gives
`DeclinedNoExactMetadataFootprint`, and a `BypassedCompatible` import pre-empts all three with
`DeclinedImportBypassed`. `CompressionMetadataKind` is an **input** the adapter must correlate, not
something derived from the descriptor — the header says exactly that — so `Unknown` is a live
outcome, not a corner. And the relation is not "precisely" in the other direction either: a
compressed descriptor whose plane *is* all-`0xff` is admitted by this gate and would get
`GuestBase` / `DccUncompressedBase` there, not a decline at all.

The load-bearing half — **wiring that seam must move this gate in the same commit** — is correct and
is the part worth having. Please just weaken the specific, e.g. "…the compressed surfaces admitted
here are ones it would DECLINE, under a reason that depends on facts the adapter supplies
(`DeclinedUnknownMetadataKind`, `DeclinedNoExactMetadataFootprint`, or
`DeclinedUnsupportedMetadataState`); a plane that proves plain would still return `GuestBase`."
Non-blocking — it is a forward-looking note about unwired code, not a safety claim about this
cache — but it is exactly the "a cited specific reads as already verified" shape, sitting in a header
where it will be quoted.

### The rest

**N4** now matches the code: the probe at `:6801` is unconditional and the materialization at
`:7599-7605` is the gated consumer, which is what the comment says. **N5**'s new arm asserts the
fourth corner correctly (it pins rather than discriminates, which is what completeness needs here).
**N7**'s `## Ruled out` row is accurate on every point I can check, including the three-skip-proof
detail and the storage/sampled asymmetry, and it uses the codebase-level argument rather than the
call-graph one — that is the version that survives refactoring. **N8** is now stated as
corroboration with the weakness named. The `BLOG.md` entry is in the right register, and
"No picture: the title screen still renders black" is the honest framing the file is for.

Recording N1 as an error you made rather than quietly correcting it is the behaviour that keeps the
project record trustworthy, and it cost you nothing here. Good PR. Merge when CI is green.

mattias800:
## N3 corrected — a42468fd

You were right, and it is worth naming what happened: I wrote a comment that claimed more than the
code supports, in a header where it will be quoted, which is the same failure that sank this PR's
first revision. I checked the enum myself rather than fixing it from your description, and it is
wrong in both directions:

* `DeclinedUnsupportedMetadataState` is one of **five** decline reasons; `DeclinedUnknownMetadataKind`
  and `DeclinedNoExactMetadataFootprint` are separate, and `DeclinedImportBypassed` pre-empts the
  metadata reasons entirely.
* `CompressionMetadataKind` is an adapter-supplied input, so `Unknown` is a live outcome.
* The relation fails the other way too: a compressed descriptor with an all-`0xff` plane is admitted
  by this gate and would land on the **non-declining** `DccUncompressedBase` there.

The comment now keeps only the load-bearing half — **wiring that seam must move this gate in the same
commit** — and states explicitly that the two do not stand in a clean correspondence, with the reasons
above, so the next reader is not tempted to reconstruct the precise claim I got wrong.

**N2 corrected too:** `cache_candidate` gates three things, not two. Sampled retention at `:9004` is
the third. I verified that at the line rather than taking it from the review.

**On `census_rows`:** I took your point into the comment rather than leaving it in the thread. The
clamp fixes the *key* and not the *weight* — `v.bytes` derives from the same clamped value and the
report ranks by bytes, so a saturated row would sort high on a fabricated number. Unreachable for any
real surface, and left visible rather than dropped, because a census that silently discards a
geometry cannot show its own invalidity.

**Scope:** `git diff e2058dc3 a42468fd` filtered to non-comment lines is **empty** — the admission
predicate and all executable code are exactly what you approved.

**Verification at `a42468fd`:** build clean, **ctest 315/315 passed, 0 failed**.

One honest note on that number, since it went the other way first. My initial post-correction ctest
reported 6 failures (`BAD_COMMAND` on six render tests). That was self-inflicted and not in the code:
a background task was killed mid-link, I cleared the wedged `ld` processes, and killing a linker
leaves a truncated binary with a *fresh mtime* — so make considered them up to date and never
relinked. Deleting the six and rebuilding gives 315/315. Recording it because "ctest went red after a
comment-only change" is exactly the kind of result that should not be waved away, and the cause was
worth pinning rather than assuming.

mattias800 (commented):
## Re-confirmation — PR #3150 @ a42468fd

**APPROVED.** Carries from `e2058dc3`, and this time the scope claim is trivially checkable: I
stripped comment and blank lines from all three touched files at both revisions and compared
digests — **identical in every one**. The admission predicate, `census_rows` and every executable
line are what I approved. Gates re-run at head: `check_diag_gates` and `check_cached_env` both
`all checks passed`, numbered-table rc=0, `diff --check` rc=0, selector test **45 PASS / 0 FAIL**.

**N3 is now right, and I checked the corrections rather than accepting them.** Counted mechanically
from `SampledSourceReason`: **five** `Declined*` reasons, exactly as the comment says. Re-read
`sampled_source_decision`'s ordering and confirmed `DccUncompressedBase` is a real non-declining
outcome, that `CompressionMetadataKind` defaults to `Unknown` and is an input rather than a
derivation, and that `BypassedCompatible` returns before the proof switch. The new text keeps the
load-bearing half, refuses to state the correspondence, and says why — which is the right shape for
a note about unwired code. That it also records "an earlier draft named one specific reason and was
wrong in both directions" is the part that stops the next person reconstructing it.

One clause is loose and I am flagging it only for completeness, not for another head: "an import
bypass pre-empts the metadata reasons entirely" is true of the three metadata *decline* reasons but
not of `ProvesPlain`, which is resolved first. It sits inside a paragraph that explicitly declines
to be precise, and the looseness runs conservative, so it reads correctly.

**N2** now enumerates the three gated consumers rather than counting two, and naming retention
alongside the other two is if anything more conservative than needed — retention is the write half
of the reuse the argument already covers. **`census_rows`**: the comment now carries the weight
caveat, the unreachability bound, and the reason for leaving the row visible instead of dropping it.
Taking that out of the review thread and into the source is the right instinct — a caveat that lives
only in a PR conversation is one nobody reads at the call site.

On the `BAD_COMMAND` episode: reporting it rather than the clean number was correct, and the failure
mode is worth more than a footnote. **A truncated binary with a fresh mtime is skipped by the build
and surfaces as six test failures rather than as a link error** — the build system reports success
about a file it never rebuilt. That is the same shape as the traps this project already records, and
it is nastier than the `tools/screenshot` one because it points the reader at the tests. If you are
filing an instrument-trap row for the 25x readback inflation anyway, this deserves the row next to
it.

Nothing further from me. Merge when required CI is green.

Issue #3152, opened 2026-08-31
No second Stray in the UE splash screens: three candidates, three different reasons — and an offscreen profiling trap


mattias800:
## Summary

I went looking for a second Stray (#3150), where a whole surface class was excluded from a cache and
re-uploaded thousands of times. **There is no equivalent in the UE splash screens I measured**, and
the three best candidates each dissolve for a different, checkable reason. Recording it so the next
lane does not re-derive it.

Screen: `prosper-app`, 70 s per title, `PROSPER_TILECENSUS=1`. Rates are fifo (see *Instrument* below).

| Title | early fps | late fps | tiled | verdict |
| --- | --- | --- | --- | --- |
| *The Forgotten City* `PPSA03026` | 29.7 | 46.2 | 32,668 MiB | **not a defect** — 30 fps movie |
| *The Plucky Squire* `PPSA15319` | 28.3 | 32.6 | 2,161 MiB | real work; cache already 100% |
| *Nikoderiko* `PPSA23760` | 33.0 | 33.4 | 2,094 MiB | not investigated (Bink) |
| *Little Nightmares III* `PPSA05143` | 42.9 | 50.2 | 2,086 MiB | not slow |
| *The Oregon Trail* `PPSA19244` | 46.6 | 50.2 | none | not slow; no tiling at all |
| *The Pathless* `PPSA01826` | 60.0 | 74.8 | 24,738 MiB | not slow (census #1739's 14.0 fps is stale) |

## The Forgotten City's splash is not slow — it is a 30 fps movie

`[avp-vaapi] hardware VA-API decode selected` for both splash movies, and `ffprobe` gives
`h264 1920x1080 **30/1**` for `1080_30fps.mp4` and `dearvillagers_sting.mp4`. Measured guest flips:
**29.88/s**. The title is playing 30 fps video and flipping once per video frame, which is correct.

Its F8 capture agrees it is not working hard: `gpu-device` **1.3%** of wall, process **0.13 cores**,
graphics **7.4%** of the window. A title that is idle at exactly the content's frame rate is not a
title with a performance defect. **Do not open a perf issue on a movie splash without checking the
movie's frame rate first** — it is one `ffprobe`.

## The Plucky Squire is doing real work

F8, windowed, GPU present: `compute` **2008.3 ms (49%)**, `gpu-device` 804.7 ms,
`renderer-resource` 665.7 ms (`build_resources texture` 446.2 ms). Its sampled image cache is at
[...]

Issue #1284, opened 2026-07-24
perf(render): Blue Prince 3D-workload decomposition — draw_setup.resources 95ms + readback 79ms dominate the 263ms submit


mattias800:
The 3D-workload evaluation `docs/RENDERER_PERFORMANCE_2026_07.md` called for, measured on post-#1283 master with `PROSPER_RENDER_TIMING=1` on the #1264 Blue Prince route (RADV, Day One gameplay window; window averages over 74 backend-submit calls, ~1,516 draws each):

```
backend-submit avg_ms: measured=263.33  draw_setup=119.87  readback=79.16  gpu_wait=29.29  record_upload=22.58  cleanup=11.03
draw_setup avg_ms:     resources=95.42  fixed=16.43  pipeline=1.22  shaders=0.13
synchronization:       command_buffers=74  queue_submits=61  fence_waits=61   (near-fully synchronous)
pipelines:             refs=1515.9  hits=1502.9  misses=13  entries=1024      (pipeline cache healthy)
color_targets:         writes=42  load_hits=42  sample_hits=30  readbacks=28  deferred=14  cached=169 MiB
texture_cache:         hits=63  submit_reuse=24  misses=0                     (texture cache healthy)
```

Ranked terms and directions:
1. **`draw_setup.resources` = 95 ms/submit (~36%)** — the per-draw resource/descriptor building loop in `render_draws_rgba` (~63 µs × 1,516 draws), *after* #1269 removed the content hashing. Needs a finer sub-attribution first (descriptor-pool allocs? vkUpdateDescriptorSets count? per-draw layout vectors?), then likely per-draw descriptor-set reuse across identical contracts (the binding cache exists per-call; the win is cross-draw batching of updates/allocations).
2. **`readback` = 79 ms/submit (~30%)** — the synchronous render→CPU readback (#1177's term). 28 readbacks/window with 14 already deferred; extending deferral/async (readback only on sample/scanout demand) is the architectural fix.
3. **`gpu_wait` = 29 ms** — 61 fence waits per 74 calls: near-synchronous submission; overlaps with (2).

Context: #1177 (family), #1268/#1269 (hashing term, closed), #1216 (BP gate). The historical Bendy detile term does not appear on this title. Frame math: at 1,516 draws/submit and 263 ms, the route runs ~3.8 submits/s — matching the observed pace of the Day One scene.

mattias800:
[...]

PR #3153, opened 2026-08-31
perf report: split the READBACK verdict on whether GPU present was adopted

mattias800:
## The problem

prosper's shipped present path blits the renderer's image straight to the swapchain and performs
**no readback** (#1270). So a performance capture whose largest component is `readback` is nearly
always measuring its own harness rather than the title:

* `tools/screenshot` **always** reads back — it never calls `set_gpu_present_active`.
* `prosper-app` reads back **whenever GPU present fails**, most commonly under
  `SDL_VIDEODRIVER=offscreen`, where the surface needs `VK_EXT_headless_surface`:
  `[app] GPU present: surface on shared instance failed (VK_EXT_headless_surface ...)`.

The report did not say so, and it named readback in `primary evidence` — the one line a reader acts
on.

## Measured, same title, same phase, same trigger (#3152)

| | offscreen | real window |
| --- | --- | --- |
| `readback` | **396.9 ms (51%)** | **0.0 ms** |
| graphics total | 778.9 ms | 371.4 ms |
| `primary evidence` | **readback** | renderer-resource |

The offscreen verdict sends you to optimise the measuring apparatus. I followed it twice in one
session before the second capture caught it, and the same shape had already cost a separate
false finding via `tools/screenshot` (instrument trap 237).

## The change

One caveat, attached only to the `readback` verdict. The verdict itself is unchanged — readback
really was the largest measured component — but it no longer stands unqualified:

```
primary evidence: readback — readback is the largest measured component (396.9 ms, 51%)
WARNING: readback is not part of the shipped present path -- a dominant readback almost always
means the capture came from a frontend without GPU present (tools/screenshot always, or
prosper-app under SDL_VIDEODRIVER=offscreen). Re-measure through a real window before acting
on this verdict
```

This is the charter's "prefer experiments that detect their own invalidity" applied to the
instrument that produced the trap, rather than relying on the reader having read traps 237/238.

## Verification

Both outputs above are real captures from this investigation, re-run through the patched tool: the
offscreen one now warns, the windowed one does not.

`test_performance_capture_report.py`: **14/14 pass**. Two new arms — the warning fires on a readback
verdict, and is absent on any other so it cannot become noise on every report. **Mutation-verified**:
disabling the note fails exactly the first arm and leaves the second passing, which is the intended
split (the second pins behaviour that must not change).

Registered as ctest `performance_capture_report_logic`. Python-only diff; no build impact.

Refs #3152.

## Evidence status of the two branches (stated plainly, per review)

* **Harness branch (GPU present NOT adopted): confirmed end-to-end.** The offscreen capture of The
  Forgotten City from #3152 classifies as readback and receives the harness note; the windowed
  capture of the same title and phase classifies as renderer-resource and receives none.
* **GPU-present branch (readback is real work): synthetic only.** No real capture in hand is both
  readback-dominant *and* GPU-present, so that branch is exercised by tests rather than by a
  measured capture. Chasing one is a hunt rather than a verification step, so it is not blocking —
  but the asymmetry should be on the record.

### DIFF
--- a/prosper/tools/perf/performance_capture_report.py
+++ b/prosper/tools/perf/performance_capture_report.py
@@ -58,6 +58,29 @@ def validate_capture(records):
             raise CaptureError(f"footer has invalid {key}")
 
 
+def _gpu_present_adopted(post):
+    """Was GPU present adopted? True / False / None when the capture cannot say.
+
+    `rendered_frame_counter` returns nullopt when GPU present is adopted rather than substituting the
+    app's host-present count, so the PRESENCE OF THE FIELD is the signal -- exactly the rule
+    `_resource_breakdown` states below ("Absent and zero are the same number and opposite facts").
+
+    Deliberately NOT derived from `rendered_fps`: that rate is also None for a sample population with
+    fewer than two entries, a non-increasing `t_ns`, or a zero window, so an OFFSCREEN capture (whose
+    samples carry a real counter) would read as adopted and get told its readback is real work --
+    inverting the advice in the one direction this whole note exists to prevent. That inversion was
+    live at 92c27046 and is what this function replaces.
+    """
+    if not post:
+        return None
+    present = [s.get("rendered_frames") is not None for s in post]
+    if all(present):
+        return False        # a real counter on every sample => GPU present was NOT adopted
+    if not any(present):
+        return True         # absent/null throughout => adopted
+    return None             # adopted then lost, or never coherent: say so rather than guess
+
+
 def _counter_rate(samples, field, seconds):
     if len(samples) < 2 or seconds <= 0:
         return None
@@ -306,12 +329,13 @@ def summarize(records):
     # different instrument that happens to share the mechanism.
     readback_note = None
     if classification == "readback":
-        # NOTE THE SENSE: a null rendered-frame population means GPU present WAS adopted, because
-        # `rendered_frame_counter` returns nullopt in that case rather than substituting the app's
-        # host-present count. Getting this backwards inverts the advice, which is how a reader ends
-        # up dismissing real readback as a harness artifact -- I had it inverted until the paired
-        # tests below caught it.
-        if rates.get("rendered_fps") is not None:
+        adopted = _gpu_present_adopted(post)
+        if adopted is None:
+            readback_note = (
+                "this capture cannot say whether GPU present was adopted, so it cannot say whether "
+                "the readback is real or the harness copying every scanout frame to the CPU. Check "
+                "the run log for a GPU-present surface failure before acting on this verdict")
+        elif adopted is False:
             readback_note = (
                 "GPU present was NOT adopted for this capture, so the frontend copied every scanout "
                 "frame to the CPU -- most often prosper-app under SDL_VIDEODRIVER=offscreen, which "
--- a/prosper/tools/perf/performance_capture_report.py
+++ b/prosper/tools/perf/performance_capture_report.py
@@ -285,6 +285,28 @@ def summarize(records):
         reason = (f"the process used {cpu_cores:.2f} CPU cores with no retained renderer/compute "
                   "timing records")
 
+    # A capture whose largest component is READBACK is nearly always measuring its own harness.
+    # prosper's shipped present path (#1270) blits the renderer's image straight to the swapchain and
+    # performs NO readback, so a large readback figure means the frontend fell back to copying every
+    # scanout frame to the CPU -- which `tools/screenshot` always does (it never calls
+    # `set_gpu_present_active`), and which `prosper-app` does whenever GPU present fails, most often
+    # under `SDL_VIDEODRIVER=offscreen`, where the surface needs `VK_EXT_headless_surface`:
+    #
+    #     [app] GPU present: surface on shared instance failed (VK_EXT_headless_surface ...)
+    #
+    # Measured on The Forgotten City (#3152): the same title, phase and trigger reported readback as
+    # 396.9 ms / 51% and "primary evidence: readback" under offscreen, against 0.0 ms through a real
+    # window -- where the verdict became renderer-resource and the whole graphics total halved. The
+    # offscreen answer names the harness as the bottleneck, confidently and in the report's own
+    # headline field. Say so here rather than letting the reader act on it (instrument traps 237/238).
+    harness_note = None
+    if classification == "readback":
+        harness_note = (
+            "readback is not part of the shipped present path -- a dominant readback almost always "
+            "means the capture came from a frontend without GPU present (tools/screenshot always, or "
+            "prosper-app under SDL_VIDEODRIVER=offscreen). Re-measure through a real window before "
+            "acting on this verdict")
+
     pacing_note = None
     if rates["guest_fps"] is not None and rates["rendered_fps"] is not None:
         if rates["guest_fps"] > max(1.0, rates["rendered_fps"] * 1.5):
@@ -315,6 +337,7 @@ def summarize(records):
         "classification": classification,
         "reason": reason,
         "pacing_note": pacing_note,
+        "harness_note": harness_note,
         "truncation": truncation,
         "counts": {
             "pre": footer["pre_samples"],
@@ -408,6 +431,8 @@ def print_summary(summary):
                   "  [breakdown UNAVAILABLE — this capture predates the backend sub-buckets;"
                   " do NOT subtract the frontend figures above from it, they are a different layer]")
     print(f"primary evidence: {summary['classification']} — {summary['reason']}")
+    if summary.get("harness_note"):
+        print(f"WARNING: {summary['harness_note']}")
     if summary["pacing_note"]:
         print(f"pacing: {summary['pacing_note']}")
 
--- a/prosper/tools/perf/performance_capture_report.py
+++ b/prosper/tools/perf/performance_capture_report.py
@@ -285,27 +285,44 @@ def summarize(records):
         reason = (f"the process used {cpu_cores:.2f} CPU cores with no retained renderer/compute "
                   "timing records")
 
-    # A capture whose largest component is READBACK is nearly always measuring its own harness.
-    # prosper's shipped present path (#1270) blits the renderer's image straight to the swapchain and
-    # performs NO readback, so a large readback figure means the frontend fell back to copying every
-    # scanout frame to the CPU -- which `tools/screenshot` always does (it never calls
-    # `set_gpu_present_active`), and which `prosper-app` does whenever GPU present fails, most often
-    # under `SDL_VIDEODRIVER=offscreen`, where the surface needs `VK_EXT_headless_surface`:
+    # A READBACK verdict has two very different meanings, and the capture can tell them apart
+    # rather than guess. `rendered_frame_counter` returns nullopt exactly when GPU present was
+    # adopted (frontends/prosper-app/present_policy.hpp), so a null rendered-frame population IS the
+    # GPU-present signal, serialized on every sample.
     #
-    #     [app] GPU present: surface on shared instance failed (VK_EXT_headless_surface ...)
+    #   * GPU present NOT adopted -- prosper-app fell back to copying every scanout frame to the CPU,
+    #     which happens under SDL_VIDEODRIVER=offscreen where the surface needs
+    #     VK_EXT_headless_surface. Then readback is the harness, and optimising it is wasted work.
+    #     Measured on The Forgotten City (#3152): same title, phase and trigger gave readback
+    #     396.9 ms / 51% and "primary evidence: readback" offscreen, against 0.0 ms through a real
+    #     window, where the verdict became renderer-resource and the graphics total halved.
+    #   * GPU present ADOPTED -- the scanout readback is skipped, so a large `readback_ms` is real
+    #     work: non-deferred colour-target readback, storage writeback, or a copy forced by
+    #     `authoritative_readback`, which every ordered DMA copy sets. That is a finding to chase,
+    #     NOT to dismiss, and saying so is the point of splitting the two cases.
     #
-    # Measured on The Forgotten City (#3152): the same title, phase and trigger reported readback as
-    # 396.9 ms / 51% and "primary evidence: readback" under offscreen, against 0.0 ms through a real
-    # window -- where the verdict became renderer-resource and the whole graphics total halved. The
-    # offscreen answer names the harness as the bottleneck, confidently and in the report's own
-    # headline field. Say so here rather than letting the reader act on it (instrument traps 237/238).
-    harness_note = None
+    # Deliberately not blamed on `tools/screenshot`: it cannot produce a capture at all, since only
+    # prosper-app arms one. The `PROSPER_TILECENSUS` readback trap (instrument trap 237) is a
+    # different instrument that happens to share the mechanism.
+    readback_note = None
     if classification == "readback":
-        harness_note = (
-            "readback is not part of the shipped present path -- a dominant readback almost always "
-            "means the capture came from a frontend without GPU present (tools/screenshot always, or "
-            "prosper-app under SDL_VIDEODRIVER=offscreen). Re-measure through a real window before "
-            "acting on this verdict")
+        # NOTE THE SENSE: a null rendered-frame population means GPU present WAS adopted, because
+        # `rendered_frame_counter` returns nullopt in that case rather than substituting the app's
+        # host-present count. Getting this backwards inverts the advice, which is how a reader ends
+        # up dismissing real readback as a harness artifact -- I had it inverted until the paired
+        # tests below caught it.
+        if rates.get("rendered_fps") is not None:
+            readback_note = (
+                "GPU present was NOT adopted for this capture, so the frontend copied every scanout "
+                "frame to the CPU -- most often prosper-app under SDL_VIDEODRIVER=offscreen, which "
+                "needs VK_EXT_headless_surface. This readback is the harness, not the title. "
+                "Re-measure through a real window before acting on this verdict")
+        else:
+            readback_note = (
+                "GPU present WAS adopted, so scanout readback is skipped and this readback is real "
+                "work -- non-deferred colour-target readback, storage writeback, or a copy forced by "
+                "authoritative_readback (every ordered DMA copy sets it). Chase it rather than "
+                "dismissing it as a harness artifact")
 
     pacing_note = None
     if rates["guest_fps"] is not None and rates["rendered_fps"] is not None:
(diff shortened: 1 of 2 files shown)


mattias800 (commented):
## REJECTED — three small corrections, all in the warning's own wording/citations

Independent review at head `c418ff3b`, in my own worktree. I did not rebuild or re-capture; I read
the sources the claim rests on and ran the Python suite plus four mutations.

The change is the right idea and I want it to land: the report currently states `readback` as
`primary evidence` with no qualification, and #3152 shows that costing the author two runs. The
scoping is clean, the tests discriminate exactly as claimed, and there is no parser risk. What I am
rejecting is that a warning whose entire purpose is to stop readers trusting a confident wrong
statement contains two confident wrong statements of its own, and a third that is stronger than the
code supports. All three are text edits.

### What I verified and agree with

* **The premise holds.** `try_adopt_shared_present` calls `gpu::set_gpu_present_active(true)`
  (`frontends/prosper-app/main.cpp:669`), and the renderer's scanout path skips the CPU
  readback+reupload behind `gpu_present_active()`
  (`frontends/shared/live/live_renderer.cpp:9375-9400`). GPU present is **default-on** for a real
  game boot — `request_gpu_present` returns true unless `PROSPER_APP_GPU_PRESENT=0`, a test pattern,
  or no game (`frontends/prosper-app/present_policy.hpp:21-24`) — so the main.cpp:638 comment
  ("called only when PROSPER_APP_GPU_PRESENT is set") is loose but the default is on, and the PR's
  claim about the shipped path is correct. The offscreen failure line is real
  (`frontends/prosper-app/main.cpp:650`) and leaves `vk.gpu_present` false.
* **Scope is tight.** `if classification == "readback"`
  (`prosper/tools/perf/performance_capture_report.py:303`) sits after the whole classification
  chain, and nothing reassigns `classification` afterwards (only assignments are at :268, :275,
  :278, :284). It cannot fire on `inconclusive` or `cpu-outside-renderer`.
* **Mutation verification reproduces exactly as the PR body describes.** Forcing the note off →
  14 tests, 1 failure, and it is `test_readback_verdict_warns_that_it_is_probably_the_harness`.
  Widening the condition two ways (`if True`, `classification != "inconclusive"`) → 1 failure each,
  and it is the *second* arm, so that arm is not vacuous: it genuinely pins the scope rather than
  passing by construction.
* **The tests use the real record shape.** `total_ms` is present in both new arms, and each asserts
  the classification it depends on (`readback` / `renderer-resource`) rather than assuming it.
* **No output-contract risk.** The only in-tree references to this tool are `CMakeLists.txt:1073`
  (test registration) and `tools/AGENTS.md:670-700` (prose). Nothing parses its stdout, and the
  `--json` mode gains one additive string-or-null key under `sort_keys=True`.

### Blocking

**1. `tools/screenshot` cannot produce a `.prperf`, so the warning names an impossible provenance
for its own input.** The only site that arms and writes a capture is `prosper-app`
(`frontends/prosper-app/main.cpp:1626` for `PROSPER_PERF_CAPTURE_AFTER_MS`, `:1653` for the arm);
`.arm(` appears nowhere else outside `frontends/shared/tests/test_performance_capture.cpp`, and
`tools/screenshot/screenshot.cpp` contains no reference to perf capture at all — it links
`prosper_live_renderer` (hence the renderer's observe hooks) but has no trigger. Trap 237 is about
`PROSPER_TILECENSUS` under `tools/screenshot`, which is a different instrument; the frontend does
not carry over to this one. A reader handed "(tools/screenshot always, ...)" is being told to
consider a provenance that no `.prperf` can have. Drop that clause, or make it "a frontend without
GPU present — in practice `prosper-app` under `SDL_VIDEODRIVER=offscreen`".

**2. The code comment miscites instrument trap 238.** `performance_capture_report.py:301` ends
"(instrument traps 237/238)" and the commit message says "traps 237 and 238 record the two sibling
cases". Trap 238 (`prosper/docs/GAME_COMPAT_ORCHESTRATION.md:366`) is *a linker killed mid-write
leaves a truncated binary with a fresh mtime* — nothing to do with readback. 237 (`:365`) is the
readback sibling and is correctly cited. The `prosper-app`-under-offscreen case has **no trap row at
all**: I grepped `offscreen` across the traps table and the only hit is line 89, an unrelated
configure note. Either cite `#3152` (where it is genuinely recorded) or allocate a row with
`tools/docs/trap_number.py` and cite that.

**3. "readback is not part of the shipped present path" is true of the present path and too strong
as a statement about the component being reported.** The `readback` component is
`_total(renderer, "readback_ms")`, and `readback_ms` is
`ms(timing_gpu_done, timing_readback_done)` (`prosper/tests/fixtures/render_runner.h:9444`) — a
window that spans more than scanout readback:

* the non-deferred colour-target readback (`render_runner.h:9149`), whose deferral the frontend
  refuses when a later pass needs CPU pixels, and **unconditionally** when
  `phase.authoritative_readback` is set (`live_renderer.cpp:8263-8266`);
* and `authoritative_readback` is set by **every ordered DMA copy** —
  `flush_span(true)` at `src/gpu/execute/gpu_executor.cpp:9093` and `:11025`. So a title that
  interleaves DMA with draws pays a full synchronous CPU readback *with GPU present active*;
* plus storage-image writeback to guest memory (`render_runner.h:9129`), which is entirely shipped
  behaviour.

The `PROSPER_READBACK_WHY` classifier (`live_renderer.cpp:8285-8305`, #1284) exists precisely
because non-deferred readback was a real measured cost, and its `kAuth` bucket (`:8305`) is exactly
the case above. So a GPU-present, real-window capture *can* legitimately report large readback, and
the warning as written tells that reader it is the harness.

The hedge ("almost always") and the prescribed action ("re-measure through a real window") bound the
damage to one extra run, which is the right design — but the warning then says nothing about the
outcome that matters most: **what it means when readback survives the re-measure.** Two edits fix
both halves: qualify the first clause (it is the *present* path that does no readback, not every
readback), and add one sentence saying that a readback that persists with GPU present is real work —
a non-deferrable readback, most often forced by an ordered DMA — and to classify it with
`PROSPER_READBACK_WHY` rather than dismiss it. Without that sentence the note does train the reader
toward dismissal, which is the failure mode this PR exists to prevent.

### Non-blocking

* **The report can *know* this instead of inferring it, and I think that is the change worth
  making.** `rendered_frames` is serialized as `null` exactly when prosper-app adopted GPU present:
  `rendered_frame_counter(gpu_present, seq)` returns `nullopt` iff `gpu_present`
  (`frontends/prosper-app/present_policy.hpp:30-33`), and the call site passes `vk.gpu_present`
  (`frontends/prosper-app/main.cpp:1649`, `:2125`). Gate on the raw field's presence (not on
  `rates["rendered_fps"] is None`, which is also None for a short window) and the note stops being a
  base-rate guess: *no GPU present* → this is the harness, say so flatly; *GPU present* → readback is
  real, point at `PROSPER_READBACK_WHY`. That is the instrument detecting its own invalidity, which
  is what the PR body claims for it. It also dissolves blockers 1 and 3 as a side effect. I am not
  requiring it — the caveat has value without it — but if you take it, the wording problems mostly go
  away.
* **Mutation gap: the printed line is untested.** Deleting the two lines at
  `performance_capture_report.py:434-435` leaves all 14 tests green (I ran it). The change's only
  user-visible effect can be removed silently. This matches the file's existing scope — no test
  touches `print_summary` — so it is a note, not a demand, but the arms as written prove the
  *summary key*, not the report.
* `tools/AGENTS.md:670-700` documents this report's output ("separates evidence for ... and
  frame-pacing gaps; deliberately reports unavailable/inconclusive fields instead of inventing a
  verdict"). A new always-visible WARNING line is worth one clause there.
* Cosmetic: `:434` uses `summary.get("harness_note")` while the neighbouring `:436` uses
  `summary["pacing_note"]`. `summarize` always sets the key, so the two spellings are equivalent;
  pick one.

Happy to re-review promptly on the corrected head — blockers 1 and 2 are one line each, and 3 is a
sentence.

mattias800 (commented):
## REJECTED — one new blocking defect: the discriminator reads "unavailable" as "GPU present adopted"

Re-review at head `92c27046`, own worktree, no rebuild. All three prior blockers are correctly
fixed, and the two-branch split is the right design — the "chase it" branch is the half that
mattered and it reads well. I am rejecting on a single new defect introduced by the rework, in the
discriminator itself.

### The prior blockers — all resolved

* **B1** — `tools/screenshot` is gone from the note, and `performance_capture_report.py:304-306`
  now records *why* it cannot be the source. Correct, and correctly reasoned.
* **B2** — trap 238 is gone; `:305` cites 237 alone and says it is a different instrument
  (`PROSPER_TILECENSUS`) that shares the mechanism. Accurate.
* **B3** — dissolved by the split. The GPU-present branch (`:321-325`) says non-deferred
  colour-target readback, storage writeback, or a copy forced by `authoritative_readback`, "which
  every ordered DMA copy sets". I re-checked all three against the source and they hold
  (`src/gpu/execute/gpu_executor.cpp:9093` and `:11025` are the two DMA sites that set it).

### Mutation claims — reproduced exactly

Inverting the sense → 2 failures, both readback arms. Disabling the note → the same 2. Widening to
every verdict → the scope arm. Forcing the harness branch → the GPU-present arm. The paired arms do
what you say they do, and the fifth mutation confirms neither is passing by construction.

### Blocking: `rates["rendered_fps"] is None` is three facts wearing one value

`:314` branches on `rates.get("rendered_fps") is not None`. That rate is `None` for three unrelated
reasons, and only one of them is "GPU present was adopted":

* the field is null on the first or last sample — the intended signal (`_counter_rate:64-66`);
* **fewer than two post samples, or a non-increasing `t_ns`** — `_counter_rate:62-63`, and
  `summarize:225` also short-circuits the whole rate to `None` when `seconds` is falsy.

So an **offscreen** capture — `rendered_frames` a real number on every sample, GPU present
demonstrably not adopted — takes the *GPU-present* branch whenever its post population is
degenerate. Measured against `summarize()` at this head:

| post samples (all carrying `rendered_frames`) | `rendered_fps` | branch taken |
| --- | --- | --- |
| two, well-formed (offscreen) | 10.0 | harness ✔ |
| `rendered_frames: null` (real GPU-present shape) | None | real work ✔ |
| **one sample** | None | **real work ✘** |
| **zero samples** | None | **real work ✘** |
| **two samples, equal `t_ns`** | None | **real work ✘** |
| adopted at first sample, lost by the last | None | real work (should be *unknown*) |

The three ✘ rows are the inversion you say the paired tests caught, re-entering along the
*availability* axis instead of the value axis — and it lands on the branch that says "Chase it
rather than dismissing it as a harness artifact", i.e. it sends the reader to optimise the harness.
That is the exact failure this PR exists to prevent, so the wrong-direction default matters more
than the frequency.

Two things make this blocking rather than a nit:

1. **This file already legislates the rule.** `_resource_breakdown`'s docstring: *"Absent and zero
   are the same number and opposite facts... the presence of the field is what decides, not its
   value."* This is that rule mirrored — *unavailable* and *null* are the same `None` and opposite
   facts — and the fix is the same one that docstring prescribes.
2. The state is contemplated by the format: `test_inconclusive_without_post_population` already
   covers a capture with no post population; nothing before now combined it with renderer records.

On reachability I want to be honest about what I did and did not do: the inversion is a
measurement, made against `summarize()` directly. That a *published* capture can reach it is an
argument, not a measurement — post samples are taken in the app loop
(`frontends/prosper-app/main.cpp:2122-2128`) while renderer records come from the render thread, so
a starved or stalled app loop can produce renderer records with almost no samples, which is
precisely the regime F8 exists to capture. I did not produce such a capture. The defect stands on
(1) regardless of how often (2) fires.

**Fix — decide on the raw field, and make the third state a third branch:**

```python
rendered_population = [sample.get("rendered_frames") for sample in post]
gpu_present = bool(rendered_population) and all(v is None for v in rendered_population)
gpu_present_known = bool(rendered_population) and (
    gpu_present or all(v is not None for v in rendered_population))
```

Then: not known → say the GPU-present signal is unavailable in this capture and give the generic
"re-measure through a real window, and check stderr for the `[app] GPU present:` line" advice,
rather than picking a side. A mixed population (the last row) belongs there too. One arm per branch
keeps the pairing property that already served you well here.

### Non-blocking

* **The GPU-present arm does not pin the real serialized shape.** It builds samples by *removing*
  `rendered_frames`; the serializer emits the key with a null value —
  `write_optional` in `frontends/shared/perf/performance_capture.cpp` writes `"null"`. Both reach
  `None` through `.get`, so the arm passes for the right reason today, but the field's *shape* is now
  the discriminator, so I would set `"rendered_frames": None` explicitly and let the test say so.
* **The GPU-present branch has no real-capture evidence yet**, and that is fine — the windowed
  Forgotten City capture classified as renderer-resource, so it never entered that branch. Just
  don't let the PR body read as though both branches were confirmed end-to-end; only the harness one
  was.
* **The print gap** — your call, and I agree with it. `pacing_note` has no coverage either, so
  bringing only this one under test would be inconsistent. Leave it.
* **The PR title still says "warn that a READBACK verdict is probably the harness"**, which is
  precisely what the rework stopped doing. The commit subject (`key the readback note on the
  GPU-present signal, not a guess`) is the accurate one — worth promoting to the title, since the
  title is what the merge commit and the log carry.
* Cosmetic, unchanged from last round: `:451` uses `.get` where `:453` uses `[...]` for the sibling
  note.

The blocking item is a few lines and one extra arm. Send it back and I will re-review promptly.

mattias800 (commented):
## APPROVED at `6982678a`

The blocking defect is fixed, and fixed at the right level: a named helper with a docstring that
cites the rule it obeys and names the inversion it replaces (`performance_capture_report.py:61-82`).
Keying on the field, and giving "cannot say" its own state rather than folding it into a branch, is
the correct shape. Two test-hygiene items below are non-blocking; I would like them in this PR
because they are trivial, but neither should hold it.

### Verified

I re-ran the seven populations from my last review against `summarize()` at this head. Every one
now resolves correctly, including the three that were inverted at `92c27046`:

| post samples | `_gpu_present_adopted` | note |
| --- | --- | --- |
| `rendered_frames: null` — the real wire shape | True | real work ✔ |
| offscreen, two well-formed | False | harness ✔ |
| offscreen, **one sample** | False | harness ✔ *(was: real work)* |
| offscreen, **two samples, frozen `t_ns`** | False | harness ✔ *(was: real work)* |
| offscreen, counter went backwards | False | harness ✔ |
| **zero samples** | None | cannot say ✔ |
| adopted then lost mid-capture | None | cannot say ✔ |

Row 1 also answers one of my earlier notes: I built it with the key **present and null**, which is
what `write_optional` actually emits, and the helper reads it correctly. So the behaviour is right
for the real wire format even though the arm exercises the key-absent spelling.

**Your mutation claims are accurate.** Restoring the derived-rate discriminator fails exactly 2
(`test_degenerate_sample_population_does_not_flip_an_offscreen_capture` and
`test_mixed_gpu_present_population_says_it_cannot_tell`); collapsing the unknown state fails exactly
1, in either direction (True or False). Inverting the `all(present)` branch fails 2. 17/17 green at
baseline.

**A correction to my own instrument, because I nearly reported your claim as false.** My first
mutation batch reported `mixed -> True` as **OK**, i.e. that the mixed arm did not discriminate.
That was my harness, not your test: `return None` → `return True` is a **same-size** edit, so
rewriting the file within the same second left `__pycache__` valid on the `(mtime, size)` check and
Python re-imported the *unmutated* bytecode. The mutation never ran. Clearing `__pycache__` between
arms gives the results above. Worth a trap row — the failure direction is the nasty one: it reports
a live test as dead, which invites deleting or weakening a guard that works. I have not filed it,
since I hold no claim here; say the word and I will, or take it yourself.

### Non-blocking — both one touch

1. **The two new arms guard their assertion with `if summary["classification"] == "readback":`**
   (`test_performance_capture_report.py:145`, `:150`, `:158`). I checked: all three populations do
   classify as `readback` today, so the arms are live — M1 fails precisely them, which is the proof.
   But the guard is the property that makes the assertion meaningful, and it is never asserted. If a
   later change made a one-sample capture `inconclusive` — arguably *more* correct — these arms go
   silently vacuous and the regression they exist for is unguarded, with nothing red anywhere. Every
   other arm in the file asserts the classification instead. Make it
   `self.assertEqual(summary["classification"], "readback")` and the conditional disappears.
2. **`if not post: return None` (`:74-75`) is the one uncovered line in the helper.** Collapsing it
   to True *or* to False leaves 17/17 green — the other two states are pinned, this one is not. One
   arm with an empty post population closes it, and it is the same edit as (1).

### Your two questions

* **The explicit `"rendered_frames": None` arm** — not blocking, and I would still add it in the
  same touch as the above. My probe confirms the *code* handles the real wire shape, so this is
  about the test documenting what it is discriminating on, now that the field's shape is the whole
  discriminator.
* **No real-capture evidence for the GPU-present branch** — not blocking, and I do not think you
  should chase one. It needs a readback-heavy title *with* GPU present (an ordered-DMA-driven
  `authoritative_readback` case), which is a hunt, not a verification step. The only thing I would
  ask is that the PR body say plainly that the harness branch is confirmed end-to-end on the #3152
  captures and the GPU-present branch is synthetic-only, so nobody later reads the PR as evidence
  that both were exercised against real captures.

Title now matches what shipped. Nothing else outstanding from my previous two reviews.

**APPROVED** — merge when CI is green; the two items above are yours to take or leave.

mattias800 (commented):
## APPROVED at `57469d98` — the approval carries

`git diff 6982678a 57469d98 -- prosper/tools/perf/performance_capture_report.py` is empty, so the
report logic is byte-identical to the head I approved. The delta is the test file only
(+26/-4), and it closes all four items I raised.

Re-verified independently, with `__pycache__` cleared between arms (trap 239):

* **19/19 green** at baseline.
* **All three of your mutation counts reproduce exactly.** Restoring the derived-rate discriminator
  → **3** failures; collapsing the unknown state → **1**; inverting `all(present)` → **5**. I first
  measured 2 for the third and went looking for the spelling difference: swapping the terminal
  returns fails 2, while `all(present)` → `not all(present)` fails 5, which is yours. Reproduced, no
  discrepancy.
* **The previously uncovered line is now genuinely covered**, which was the one I cared most about:
  `if not post: return None` fails **in both directions** — collapsing it to True fails 1, to False
  fails 1. Same for the mixed state. Both unknown branches are now pinned, not just one.
* The new wire-shape arm discriminates: inverting the `not any(present)` branch fails it together
  with the key-absent arm, so the two spellings are both live rather than one shadowing the other.
* The three converted arms now assert `classification == "readback"` rather than guarding on it, so
  the latent vacuity is gone.

One observation, not a request: `test_no_post_samples_cannot_determine_gpu_present` keeps an
`if/else` shape, but it asserts on **both** paths and demonstrably discriminates in both mutation
directions, so it is not the vacuity pattern the others had. Fine as written.

PR body now records the evidence asymmetry plainly (harness branch confirmed end-to-end on the
#3152 captures, GPU-present branch synthetic-only). Thank you for filing trap 239 — recording the
failure direction is the part that makes it useful.

Nothing outstanding. **APPROVED** — merge on green CI.

Issue #3156, opened 2026-08-31
test_game_compute cannot host write-watch promotion (no SIGSEGV handler): PROMOTE_HITS=1 faults there, not in production


mattias800:
## Repro

On **unmodified master** (`9d528a3f`), Linux:

```
cd prosper/build-linux
PROSPER_COMPUTE_WRITE_WATCH_PROMOTE_HITS=1 ./test_game_compute   # rc=139 (SIGSEGV)
./test_game_compute                                              # rc=0
```

Bisected across the value — only `1` crashes:

| value | result |
| --- | --- |
| 1 | **rc=139 SIGSEGV** |
| 2 | rc=0 |
| 3 (default) | rc=0 |
| 4 | rc=1 (ordinary test failure, not a crash) |

Also reproduces through ctest as `game_compute_exec` and
`game_compute_exec_no_adaptive_storage_result`.

## What I know

`may_promote_write_watch_before_exact` computes a *prospective* stability:

```
const uint32_t prospective_stability = update_write_watch_stability(
    stable_validations, true, promotion_validations);
```

With `stable_validations = 0` and `promotion_validations = 1`, `update_write_watch_stability`
returns `0 + 1 = 1`, and `should_promote_write_watch` then sees `1 >= 1` and promotes. So a watch is
armed on the **first** acquisition, before any successful full compare has ever happened. The
counter is meant to mean "proven stable N times"; at N=1 the prospective bump satisfies the
threshold before the proof exists.

## What I do NOT know

**The actual faulting access.** The backtrace is unhelpful — `#0 main ()` in an optimised build with
no deeper frames — so the above is the reachable difference in behaviour, not a demonstrated root
cause. Someone should confirm before fixing.

## Why it matters

`PROSPER_COMPUTE_WRITE_WATCH_PROMOTE_HITS` is a user-facing switch with no documented lower bound, so
a value that crashes deterministically is a trap. It also caps the win available in the linked
promotion-threshold issue: `hits=1` removes more `memcmp` traffic than `hits=2`, and is unusable.

Found while measuring the promotion threshold; I stopped at `hits=2` because of this.

mattias800:
## Correction: this is a TEST-HARNESS limitation, not a production defect

`tests/gpu/recompiler/test_game_compute.cpp` **deliberately does not install the emulator's SIGSEGV
write-fault handler**, and says so itself:

* `:4706` — *"This binary does not install the emulator's SIGSEGV write-fault handler. Mark the same
[...]

PR #3158, opened 2026-08-31
perf(compute): arm a write watch after one validation, not three

### LINKED ISSUE #3155: Write-watch promotion appears inert: only the precondition-skipping hits=1 reduces memcmp traffic
## What

Every cached compute source must be proven unchanged before reuse. Three proofs are tried
cheapest-first: the intra-submit GPU write journal (O(1)), an armed page-protection watch (O(1)), and
finally a byte-for-byte `memcmp` of the whole source against a retained snapshot (**O(size)**).

`cached.write_watch_stable_validations` counts *consecutive successful full compares*, and a page
watch is armed only once that reaches `PROSPER_COMPUTE_WRITE_WATCH_PROMOTE_HITS` — **default 3**. So
every cached texture pays **three complete byte-for-byte scans of itself** before the cheap check
becomes available.

## Measured

The Plucky Squire `PPSA15319`, `prosper-app`, matched 30 s windows, counted with an `LD_PRELOAD`
`memcmp` interposer (a direct count, not a sampled profile):

| | compares | bytes |
| --- | --- | --- |
| `hits=3` (current default) | 2,097,152 | **67.0 GB** |
| `hits=2` | 1,048,576 | **30.5 GB** |

**Half the compares, 55% fewer bytes.** A separate baseline run measured 104.4 GB, 87% of it in
512 KiB–4 MiB buffers at ~47 compares per frame — so the absolute figure moves between runs, but the
within-run comparison is stable.

`perf` attributes 5.31% of CPU to `__memcmp_evex_movbe`, the largest single symbol in the profile.

## Why 2 and not 1

`hits=1` **segfaults** — see the linked issue. 2 is the smallest value that arms after a *real*
validation.

## Why this is safe

The `memcmp` still runs on the acquisition where the watch is armed; the watch only serves the
**next** one. This changes when the cheap path becomes available, never whether data is verified.
The promotion **budget** (8 MiB/submit) still bounds how much is armed.

## Not claimed

An fps improvement. It is real but close to run-to-run spread on this host, and I am not going to
quote a number I cannot separate from noise. The memory-traffic reduction is the claim.

mattias800:
Fixes #3155. Refs #3156, #3152.

## The problem

Reusing a cached compute source requires proving it unchanged. Three proofs, cheapest-first: the intra-submit write journal (O(1)), an armed page watch (O(1)), then a byte-for-byte `memcmp` of the whole source (**O(size)**).

A page watch is armed only after `PROSPER_COMPUTE_WRITE_WATCH_PROMOTE_HITS` **consecutive successful full compares — default 3**. So every cached texture pays three complete scans of itself before the cheap check is available. On these UE titles that is multi-MB buffers, ~47 compares per frame.

## Measured

The Plucky Squire `PPSA15319`, matched 30 s windows, counted with an `LD_PRELOAD` `memcmp` interposer — a direct count, not a sampled profile:

| | compares | bytes |
| --- | --- | --- |
| `hits=3` (was) | 2,097,152 | **67.0 GB** |
| `hits=2` (now) | 1,048,576 | **30.5 GB** |

**Half the compares, 55% fewer bytes.** `perf` attributes 5.31% of CPU to `__memcmp_evex_movbe`, the largest single symbol in the profile.

## Why this is safe

The `memcmp` still runs on the acquisition where the watch is armed — the watch only serves the **next** one. This changes *when* the cheap path becomes available, never whether data is verified. The promotion budget (8 MiB/submit) still bounds how much is armed per submit.

## Why 2 and not 1

`hits=1` **segfaults** `test_game_compute` deterministically on unmodified master (#3156). `update_write_watch_stability` returns 1 on the first call, so it arms a watch before any successful compare exists. Bisected: 1 crashes, 2/3/4 do not. **2 is the smallest value that arms after a real validation.**

## What is deliberately NOT claimed

**An fps improvement.** There is one, but it is close to run-to-run spread on this host. I measured +12.7% earlier and it did not survive interleaved re-testing across a machine regime change (~+5%), so I am not quoting a number I cannot separate from noise. The memory-traffic reduction is the claim, and it is a count rather than a sample.

## Verification

- **ctest 315/315**, 0 failed.
- Both `memcmp` figures from matched back-to-back runs on the same binary, one env var apart.
- Regression-checked on a guarded title: The Messenger shows no loss.

### DIFF
--- a/prosper/frontends/shared/live/live_compute.cpp
+++ b/prosper/frontends/shared/live/live_compute.cpp
@@ -1029,10 +1029,31 @@ VkDeviceSize persistent_compute_image_limit(
     return compute_image_cache_default_limit_bytes(largest_local_heap);
 }
 
+// Arm a page watch on the FIRST unchanged validation rather than the third.
+//
+// The deferral exists because "arming a very large cold texture is more expensive than its first few
+// exact validations" (write_watch_policy.hpp). Measured, that trade does not hold: each deferred
+// validation is a byte-for-byte memcmp of the WHOLE source, and on The Plucky Squire that came to
+// 104.4 GB of memcmp in a 30 s run -- 87% of it in 512 KiB-4 MiB buffers, ~47 compares per frame
+// (#3152). An mprotect over those pages is far cheaper than one such compare, let alone three.
+//
+// Measured effect on Nikoderiko, three interleaved pairs with the order alternated:
+// Measured on The Plucky Squire, memcmp traffic over a matched 30 s window:
+//   hits=3 (was): 2,097,152 compares, 67.0 GB
+//   hits=2 (now): 1,048,576 compares, 30.5 GB   -- half the compares, 55% fewer bytes
+// The promotion BUDGET still bounds how much is armed per submit, so this changes WHEN a watch is
+// armed, not how many.
+//
+// TWO, not one, and the difference is a crash rather than a tuning preference.
+// `update_write_watch_stability` returns 1 on the very first call, so hits=1 arms a watch BEFORE any
+// exact validation has run -- `PROSPER_COMPUTE_WRITE_WATCH_PROMOTE_HITS=1` segfaults
+// `test_game_compute` deterministically on unmodified master (bisected: 1 crashes, 2/3/4 do not).
+// That is a latent defect in the knob, filed separately; 2 is the smallest value that arms after a
+// REAL validation.
 uint32_t compute_write_watch_promotion_validations() {
     static const uint32_t value = [] {
         const char* text = std::getenv("PROSPER_COMPUTE_WRITE_WATCH_PROMOTE_HITS");
-        const uint64_t parsed = text ? std::strtoull(text, nullptr, 10) : 3ull;
+        const uint64_t parsed = text ? std::strtoull(text, nullptr, 10) : 2ull;
         return static_cast<uint32_t>(std::min<uint64_t>(parsed, UINT32_MAX));
     }();
     return value;


mattias800:
## Closing — the measurement behind this was broken, and the change is a no-op

My `memcmp` interposer dumps its tally **periodically** (every 2^20 calls) rather than at exit, and I
read whichever dump last fired as though it were a run total. The headline comparison —
`hits=3: 67.0 GB` against `hits=2: 30.5 GB` — was a tally dumped at **1,048,576 calls** read against
one dumped at **2,097,152 calls**. A partial-run snapshot against a fuller one. The 55% was an
artifact of comparing different amounts of run, not of the change.

Re-measured at 2^16 granularity, twice, second pass with the order reversed:

| | pass 1 | pass 2 (reversed) |
| --- | --- | --- |
| `hits=3` | 65.4 GB | 63.2 GB |
| **`hits=2`** | **63.1 GB** | **64.6 GB** |
| `hits=1` | 29.1 GB | 28.0 GB |

**`hits=2` is within noise of `hits=3`.** This PR changes a default for no measured benefit. Closing.

## What is actually true, and is worth someone's time

Only `hits=1` moves the number — a genuine **~56%** reduction — and that is exactly the value that
arms a page watch *before any successful validation has run*. So the promotion mechanism as designed
appears to help almost only in the configuration that skips its own precondition. The likely reason
`hits=2` does nothing is that `write_watch_stable_validations` resets to 0 on any content change, so
under these workloads sources rarely accumulate two *consecutive* successes — the ladder is climbed
and knocked down repeatedly.

That is the finding worth keeping, and it is a stronger one than the change I was proposing. Moved to
#3155.

## Correction to #3156 as well

The `hits=1` segfault is **not** a production defect. `tests/gpu/recompiler/test_game_compute.cpp`
deliberately does not install the emulator's SIGSEGV write-fault handler (its own comments at
`:4706` and `:4816` say so) and compensates by calling `guest_write_watch_notify_host_write()` at the
write sites it knows about. `hits=1` arms watches at sites it has no hook for, so the write faults
fatally. `prosper-app`, which does install the handler, ran `hits=1` a dozen times without incident.
Re-framed there.

Issue #3157, opened 2026-08-31
Compute dispatch is serialized: 27 submits/frame each pay full GPU completion (batching, not faster waits)


mattias800:
## The measurement `RENDERER_PERFORMANCE_2026_07.md` asked for

That pass stopped with: *"The remaining cost is structural, and the next renderer changes must be
tested against a 3D title before choosing a resource-lifetime or scheduling architecture."* Three UE
3D titles are now measured, and the structural cost has a location and a price.

## The mechanism

`execute_item` blocks on the GPU for **every compute dispatch**, on the normal path:

```
const VkResult compute_submit_rc = vkQueueSubmit(...);
...
const VkResult wait_result = vkWaitForFences(
    ctx.device, 1, &ctx.dispatch_fence, VK_TRUE, 30ull * 1000 * 1000 * 1000);
```

Submit, block until complete, repeat. The CPU can never run ahead of the GPU.

## The price

The Plucky Squire `PPSA15319`, F8 capture over a 5.03 s window at ~34 guest flips/s:

| | |
| --- | --- |
| dispatches | 4,700 (**27.3 per frame**) |
| `gpu-device` per dispatch | **171 us** (actual GPU work) |
| `compute` wall per dispatch | **427 us** |
| round-trip overhead | **256 us each** |
| **cost per frame** | **7.0 ms of a 29.2 ms frame — 24%** |

Corroborating, from the same captures: **`gpu-device` is ~16% of the frame and the GPU is idle ~84%
of it**, while `AgcSubmissionTh` holds ~16 ms of each 30.7 ms frame at 51.8% of a core. Nothing is
saturated — per-thread `/proc` sampling shows no thread above 52%. That is the signature of latency,
not throughput.

Three titles sit in the same regime: Nikoderiko 33.8, Plucky Squire 32.6, Little Nightmares III
29.3 guest flips/s, all classified `compute` by the F8 report (#3152).

## Directions, unranked and unvalidated

* **Batch dispatches per submit.** Several dispatches into one command buffer with one fence.
  Bounded by real inter-dispatch dependencies, which need enumerating first.
* **Pipeline the fences.** Prepare dispatch N+1 while N executes; wait only when a result is
  actually consumed. Needs a per-dispatch fence pool and a readback-ordering audit.
* **Wait only where a result is read back.** Many dispatches produce GPU-resident data nothing
  reads on the CPU that frame.

## Recommended shape of the work

[...]

PR #3159, opened 2026-08-31
perf(compute): reuse the GPU comparison descriptor pool instead of recreating it per dispatch

mattias800:
Refs #3157.

## What

The GPU result-comparison path created **and destroyed** a `VkDescriptorPool` for every dispatch that compares its output. The main descriptor pool a few lines away already avoids this — `prepare_descriptor_pool` resets and reuses, growing capacities monotonically. This applies the same contract to the comparison pool.

## Measured

An `LD_PRELOAD` interposer reporting **rates**, not running totals — a periodic dump of a cumulative total is a *prefix* of the run, and comparing prefixes of different lengths is how I produced a wrong figure earlier on #3155:

| title | baseline | reuse on | baseline is |
| --- | --- | --- | --- |
| *The Plucky Squire* `PPSA15319` | 865 `vkCreateDescriptorPool`/s | **721/s (−17%)** | master |
| *Nikoderiko* `PPSA23760` | 1,212/s | **1,038/s (−14%)** | `PROSPER_NO_COMPARE_POOL_REUSE=1` |
| *Nikoderiko*, earlier run | 784/s | **702/s (−10%)** | master |

Two different baselines, so they are labelled rather than sharing a column. The Plucky Squire also has
an opt-out control arm at **854/s** against master's 865/s.

**The two Nikoderiko rows are kept deliberately, because the discrepancy is informative.** The same
condition measured **702/s** in one run and **1,038/s** in another — ~48% apart — with both arms
shifting together each time. That is the strongest evidence in this PR that the paired design is
doing its job, and that absolute rates from this interposer are **not** comparable across runs. Only
the within-run pair means anything.

`PROSPER_NO_COMPARE_POOL_REUSE=1` restores the old path exactly, and that control arm measures **854/s** against master's 865/s — so the reduction is attributable to the change rather than to run variance.

## What is NOT claimed

**A frame-rate improvement. There is no measurable one.** These titles are dominated by dispatch serialization (#3157) — 27 submits per frame, each awaiting full GPU completion — and trimming per-dispatch driver calls does not move that. This removes real work from the hot path and follows an existing pattern in the same file; it is not the fix for the fundamental problem, and I would rather say so than dress it up.

## Scope — corrected in review

The remaining rate is **not** `prepare_descriptor_pool`: its three capacities grow monotonically and the reuse test is `<=` on each, so it recreates only on a new maximum. The reviewer located the real source — the **renderer creates one `VkDescriptorPool` per render pass** (`tests/fixtures/render_runner.h`, reached via `render_draws_rgba`). Recorded in-code on `prepare_descriptor_pool` so the next reader is not sent to the wrong file, and it is the obvious next target.

## Verification

- **ctest 315/315** — but note this is *not* evidence for this change: nothing under `tests/` names the compare path, so no test demonstrates the reuse branch ran. What does demonstrate it is the control arm above — if the branch never executed, both arms would measure the same.
- The check that would catch a descriptor-lifetime VUID is the **Vulkan validation scan** in the Linux CI job, not ctest.
- Churn figures from matched back-to-back runs, plus an opt-out control arm.
- Ownership: the context now owns and resets the pool; `cleanup` no longer destroys it, and context teardown does.

### DIFF
--- a/prosper/frontends/shared/live/live_compute.cpp
+++ b/prosper/frontends/shared/live/live_compute.cpp
@@ -1119,8 +1119,10 @@ struct VulkanComputeContext {
     std::mutex pipeline_cache_mutex;
     VkDescriptorPool descriptor_pool = VK_NULL_HANDLE;
     // #3157: the GPU result-comparison pool, owned by the context so it is reset rather than
-    // recreated per dispatch. Measured 853 vkCreateDescriptorPool + 853 vkDestroy per second on
-    // The Plucky Squire before this; the main descriptor pool above already works this way.
+    // recreated per dispatch. It contributed about 144 vkCreateDescriptorPool + 144 vkDestroy per
+    // second on The Plucky Squire (whole-process rate 865/s before, 721/s after); the main
+    // descriptor pool above already works this way. The bulk of the remaining rate is NOT this path
+    // and not prepare_descriptor_pool -- see the note on that helper.
     VkDescriptorPool compare_pool = VK_NULL_HANDLE;
     uint32_t compare_pool_sets = 0;
     uint32_t compare_pool_buffers = 0;
@@ -1357,7 +1359,6 @@ struct VulkanComputeContext {
         if (command_pool) vkDestroyCommandPool(device, command_pool, nullptr);
         if (descriptor_pool) vkDestroyDescriptorPool(device, descriptor_pool, nullptr);
         if (compare_pool) vkDestroyDescriptorPool(device, compare_pool, nullptr);
-        compare_pool = VK_NULL_HANDLE;
         if (compare_pipeline) vkDestroyPipeline(device, compare_pipeline, nullptr);
         if (compare_pipeline_layout)
             vkDestroyPipelineLayout(device, compare_pipeline_layout, nullptr);
@@ -1378,6 +1379,11 @@ struct VulkanComputeContext {
         if (borrowed) return;                       // renderer owns the device/instance
         if (device) vkDestroyDevice(device, nullptr);
         if (instance) vkDestroyInstance(instance, nullptr);
+    }
+
+    static bool compare_pool_reuse_enabled() {   // opt-out for bisection
+        static const bool enabled = std::getenv("PROSPER_NO_COMPARE_POOL_REUSE") == nullptr;
+        return enabled;
     }
 
     static bool descriptor_pool_reuse_enabled() {
@@ -1434,6 +1440,10 @@ struct VulkanComputeContext {
         return true;
     }
 
+    // NOT the source of the ~700 remaining vkCreateDescriptorPool/s: all three capacities below
+    // grow monotonically and the reuse test is <= on each, so this recreates only on a new maximum.
+    // The renderer creates one pool per RENDER PASS (tests/fixtures/render_runner.h, reached via
+    // render_draws_rgba) -- that is where to look next (#3157 review).
     VkDescriptorPool prepare_descriptor_pool(uint32_t buffers, uint32_t sampled,
                                              uint32_t storage) {
         if (!descriptor_pool_reuse_enabled()) return VK_NULL_HANDLE;
@@ -1471,9 +1481,7 @@ struct VulkanComputeContext {
     // Reset-and-reuse the comparison descriptor pool. Same contract as prepare_descriptor_pool:
     // grow capacities monotonically, reset when the request fits, recreate only when it does not.
     VkDescriptorPool prepare_compare_descriptor_pool(uint32_t sets, uint32_t buffers) {
-        static const bool reuse =
-            std::getenv("PROSPER_NO_COMPARE_POOL_REUSE") == nullptr;   // opt-out for bisection
-        if (!reuse || !sets || !buffers) return VK_NULL_HANDLE;
+        if (!compare_pool_reuse_enabled() || !sets || !buffers) return VK_NULL_HANDLE;
         if (compare_pool && sets <= compare_pool_sets && buffers <= compare_pool_buffers) {
             if (vkResetDescriptorPool(device, compare_pool, 0) == VK_SUCCESS) return compare_pool;
             vkDestroyDescriptorPool(device, compare_pool, nullptr);
--- a/prosper/frontends/shared/live/live_compute.cpp
+++ b/prosper/frontends/shared/live/live_compute.cpp
@@ -1118,6 +1118,12 @@ struct VulkanComputeContext {
     std::filesystem::path pipeline_cache_path;
     std::mutex pipeline_cache_mutex;
     VkDescriptorPool descriptor_pool = VK_NULL_HANDLE;
+    // #3157: the GPU result-comparison pool, owned by the context so it is reset rather than
+    // recreated per dispatch. Measured 853 vkCreateDescriptorPool + 853 vkDestroy per second on
+    // The Plucky Squire before this; the main descriptor pool above already works this way.
+    VkDescriptorPool compare_pool = VK_NULL_HANDLE;
+    uint32_t compare_pool_sets = 0;
+    uint32_t compare_pool_buffers = 0;
     VkCommandPool command_pool = VK_NULL_HANDLE;
     VkCommandBuffer command_buffer = VK_NULL_HANDLE;
     VkFence dispatch_fence = VK_NULL_HANDLE;
@@ -1350,6 +1356,8 @@ struct VulkanComputeContext {
             vkDestroyQueryPool(device, dispatch_timestamp_pool, nullptr);
         if (command_pool) vkDestroyCommandPool(device, command_pool, nullptr);
         if (descriptor_pool) vkDestroyDescriptorPool(device, descriptor_pool, nullptr);
+        if (compare_pool) vkDestroyDescriptorPool(device, compare_pool, nullptr);
+        compare_pool = VK_NULL_HANDLE;
         if (compare_pipeline) vkDestroyPipeline(device, compare_pipeline, nullptr);
         if (compare_pipeline_layout)
             vkDestroyPipelineLayout(device, compare_pipeline_layout, nullptr);
@@ -1458,6 +1466,32 @@ struct VulkanComputeContext {
         if (!count || vkCreateDescriptorPool(device, &info, nullptr, &descriptor_pool) != VK_SUCCESS)
             descriptor_pool = VK_NULL_HANDLE;
         return descriptor_pool;
+    }
+
+    // Reset-and-reuse the comparison descriptor pool. Same contract as prepare_descriptor_pool:
+    // grow capacities monotonically, reset when the request fits, recreate only when it does not.
+    VkDescriptorPool prepare_compare_descriptor_pool(uint32_t sets, uint32_t buffers) {
+        static const bool reuse =
+            std::getenv("PROSPER_NO_COMPARE_POOL_REUSE") == nullptr;   // opt-out for bisection
+        if (!reuse || !sets || !buffers) return VK_NULL_HANDLE;
+        if (compare_pool && sets <= compare_pool_sets && buffers <= compare_pool_buffers) {
+            if (vkResetDescriptorPool(device, compare_pool, 0) == VK_SUCCESS) return compare_pool;
+            vkDestroyDescriptorPool(device, compare_pool, nullptr);
+            compare_pool = VK_NULL_HANDLE;
+        } else if (compare_pool) {
+            vkDestroyDescriptorPool(device, compare_pool, nullptr);
+            compare_pool = VK_NULL_HANDLE;
+        }
+        compare_pool_sets = std::max(sets, compare_pool_sets);
+        compare_pool_buffers = std::max(buffers, compare_pool_buffers);
+        VkDescriptorPoolSize size{VK_DESCRIPTOR_TYPE_STORAGE_BUFFER, compare_pool_buffers};
+        VkDescriptorPoolCreateInfo dpci{VK_STRUCTURE_TYPE_DESCRIPTOR_POOL_CREATE_INFO};
+        dpci.maxSets = compare_pool_sets;
+        dpci.poolSizeCount = 1;
+        dpci.pPoolSizes = &size;
+        if (vkCreateDescriptorPool(device, &dpci, nullptr, &compare_pool) != VK_SUCCESS)
+            compare_pool = VK_NULL_HANDLE;
+        return compare_pool;
     }
 
     static bool persistent_mapping_enabled() {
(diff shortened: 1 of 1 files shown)


mattias800 (commented):
Independent review of #3159 at head `945adc4e`. Read the whole compare path in
`prosper/frontends/shared/live/live_compute.cpp`, both descriptor-pool helpers, the submit/fence
region, the post-loop teardown, and the context lifecycle. No builds or captures re-run.

## Verdict: **REJECTED**

**The design and the resource lifetime are correct** — I could not find a path that resets the pool
while the GPU may still reference its sets, and ownership/capacity are sound. I say that up front so
none of it gets re-derived. The block is on **two factual claims in artifacts that outlive this PR**,
one of which is a committed code comment contradicted by the PR's own headline number. Both are a few
lines to fix; re-review should be quick.

---

## What I verified, and why it holds (no need to redo this)

**1. Reset safety — safe, and by an invariant that already had to hold.**
`prepare_compare_descriptor_pool` is called at `live_compute.cpp:8291`, i.e. *before*
`prepare_dispatch_commands()` at `:8349`, which already does `vkResetCommandPool` (`:1401`) and
`vkResetFences` (`:1408`), and before `prepare_descriptor_pool`'s own `vkResetDescriptorPool`
(`:1442`) on the main pool. So the compare pool joins three pieces of context state that the existing
design *already* resets at the top of dispatch N+1. If it were possible for dispatch N's work to still
be running at that point, the command pool and the fence would already be undefined behaviour. This
change does not create a new class of hazard; it subscribes to an existing invariant.

I walked every exit between the pool call and the fence anyway:

* Every `break` from `:8291` through `vkEndCommandBuffer` at `:8919` happens **before**
  `vkQueueSubmit` (`:8935`) — nothing is in flight, so no set is referenced.
* Submit failure → `break` at `:8937`. Either nothing was submitted, or the result was
  `VK_ERROR_DEVICE_LOST`, which `vk_note_failure` (`:5369`) latches on the context.
* Fence-wait failure (`:8940`) → `vkQueueWaitIdle` drain at `:8979`; a successful drain sets
  `completion_proven`. A `VK_TIMEOUT` followed by a successful drain is therefore proven complete.
* Device loss → `ctx.device_lost` latched, `ctx.completion_unproven` set at `:9834`, `cleanup()`
  deliberately skipped at `:9840`. **And no further `execute_item` can run**: the batch loop breaks on
  `context.device_lost` (`:10518`) and the entry point declines outright (`:10464`). So the one path
  with no completion proof is also the path where nothing ever resets the pool again. That is the
  argument that closes the question, and it is worth stating explicitly in the PR body.

Single-threading is likewise pre-established, not newly assumed: one production entry point
(`set_submit_compute(execute_live_compute_items)`, `:10652`), one `VkCommandBuffer`, one
`VkFence`, one `vkQueueSubmit` in the whole file. The other `g_live_compute_context` readers
(`:10080`, `:10088`, `:10205`, `:10218`, `:10222`) touch no descriptor pool.

One consequence worth recording somewhere: with the pool retained, `cleanup()` destroys
`compare_flags` while the sets in the retained pool still *reference* it. That is spec-legal — the
sets merely become invalid, and an invalid set is only a problem if it is used in a submitted command
buffer, which cannot happen because the next dispatch resets the pool and re-allocates and re-writes
every set (`:8303-8331`). Worth a sentence in the PR body because it is the first thing a reader
worries about.

**2. Ownership — correct on all four combinations.** `compare_pool_owned_by_context` (`:5351`) is set
only in the branch where the context handed back a live pool (`:8293-8295`), and `cleanup()` (`:5418`)
skips the destroy only under that flag — matching the existing `descriptor_pool_reused` precedent at
`:5355`/`:5429`. The fallback (`:8297`) is byte-identical to the old path and still destroys. When
`prepare_compare_descriptor_pool` fails its own `vkCreateDescriptorPool` it nulls the member (`:1493`)
before returning, so the next dispatch re-enters the create branch rather than reusing a stale handle
— no leak, no double free. Teardown at `:1359` sits with `descriptor_pool`, before the
`if (borrowed) return;` at `:1378` and `vkDestroyDevice`, and is reached only when
`completion_unproven` is false (`:10437`). Correct.

**3. Capacity — exact, and robust beyond the current call site.** The compare set layout is three
`STORAGE_BUFFER` bindings of `descriptorCount = 1` (`:2554-2559`), so an allocation of *N* sets needs
*N* `maxSets` and *3N* descriptors. The call site passes exactly `(N, 3N)` (`:8291`), the reuse branch
serves a request only when **both** `sets <= compare_pool_sets` **and**
`buffers <= compare_pool_buffers` (`:1477`), and the recreate branch maxes both (`:1485-1486`). So the
pool is always at least as large as the request — and that holds for an arbitrary sets:buffers ratio,
not merely the constant 3:1 this one call site uses, so a future second caller cannot undersize it.
`!sets || !buffers` is unreachable from this call site (the block is guarded on
`!compare_targets.empty()`), which is fine as defensive code.

---

## Blocking

**B1. `live_compute.cpp:1122` states a number that this PR's own result contradicts, in the permanent
artifact.**

```
// recreated per dispatch. Measured 853 vkCreateDescriptorPool + 853 vkDestroy per second on
// The Plucky Squire before this; the main descriptor pool above already works this way.
```

Attached to `compare_pool`, that sentence says the comparison pool cost 853 creations/s. The PR's own
table says master was **865/s** and this change leaves **721/s** — so the comparison pool's measured
contribution is **~144/s**, and 853 is (within noise) the *whole-process* rate, indistinguishable from
the 854/s opt-out control arm. The comment therefore overstates the compare path's share by roughly
6x, and does it three lines above the code it describes, where the next reader will take it as the
motivation for the change.

Either the comment's number is the total mislabelled as the compare path's, or the comment's run and
the PR's run disagree by 6x about the same quantity and the discrepancy needs explaining. Both
readings require an edit. Replace it with the decomposed figure the PR actually measured, or state
plainly that 853/s is the process-wide rate of which this path is one contributor. `CLAUDE.md` is
explicit that a wrong figure in a committed artifact carries authority it no longer deserves, and this
one is self-contradicting against the PR that introduces it.

**B2. The "Scope" paragraph misattributes the residual ~700/s, and points the next lane at the wrong
file.**

> The remaining ~700 pool creations per second come from `prepare_descriptor_pool` recreating rather
> than resetting.

`prepare_descriptor_pool` cannot be doing that in steady state. Its three capacities are monotone
(`live_compute.cpp:1450-1452`) and the reuse test is `<=` on all three (`:1440-1441`), so it recreates
only when a *new maximum* appears — a handful of times per run, not ~700 times per second. The
`!count` guard cannot fire either once any dispatch has bound a single buffer.

The likely actual source is the renderer, not compute. `vkCreateDescriptorPool` appears at exactly
three non-test sites in the tree, and the third is
`prosper/tests/fixtures/render_runner.h:5805-5819`, inside `render_draw_pass_rgba` — **one fresh
`VkDescriptorPool` per render pass**, on the live path via
`render_draws_rgba` (`frontends/shared/live/live_renderer.cpp:9270`, `:8422`, `:8685`, `:8823`). At
27 submits/frame with multiple passes each, that reaches hundreds per second easily. Since the
interposer counts process-wide, the 721/s residual is consistent with the renderer, not with a
function whose capacities provably saturate.

This is not a correctness defect in the patch, but it is a follow-up direction stated as fact in the
PR body, and it sends the next person to a file where the work does not exist. Please correct or
downgrade it to a hypothesis.

---

## Non-blocking

* **The measurement is sound where it is controlled, and the control is the best part of it.** Same
  binary, one env var, control at 854/s against master's 865/s while the enabled arm sits at 721/s —
  that is a discriminator that can show its own lever moved, which is exactly right. Two caveats worth
  a line in the PR: the *Nikoderiko* arm (784 → 702) has **no** opt-out control, so it rests on
  run-to-run variance being small, evidenced only from the other title's single control pair; and the
  rate's denominator (window length, route, phase) is not stated, which matters because pool-creation
  rate is phase-dependent. Neither changes the conclusion; both make it auditable.
* **The "no fps improvement" framing is right and should stay.** Declining to claim a frame-rate win
  is the correct read of this evidence, and I would not want it softened on re-submission.
* **ctest 315/315 is not evidence for this change.** `grep` over `tests/` finds no reference to
  `gpu_result_unchanged`, `result_baseline` or `compare_flag_index`, so nothing in the suite names the
  compare path and 315/315 does not show the reuse branch executed even once. The churn delta is the
  real evidence and it is adequate. The check that *would* see a lifetime VUID here is the **Vulkan
  validation scan inside the Linux CI job** (`.github/workflows/ci.yml:135-137`), which is still
  `pending` on this head — quote it rather than ctest when you post final results.
* `live_compute.cpp:1360` — `compare_pool = VK_NULL_HANDLE;` in the destructor is dead (no later
  member destructor reads it) and inconsistent with the neighbouring `descriptor_pool` line. Harmless.
* `prepare_compare_descriptor_pool` collapses three distinct outcomes into `VK_NULL_HANDLE` — reuse
  disabled, degenerate arguments, creation failed — and the call site cannot tell them apart. Benign
  in practice, since a real creation failure then falls into the legacy create which names
  `compare-descriptor-pool` in the census. Worth a comment.
* Style: the main pool gates on a named `descriptor_pool_reuse_enabled()` helper (`:1383`) at both the
  call site and inside the helper; the new one hides its static local inside the function. Matching the
  existing shape would make `PROSPER_NO_COMPARE_POOL_REUSE` discoverable the same way. `check_diag_gates.py`
  passes as-is (I ran it: 570 files, 737 variables, 108 findings all baselined), so this is taste only.
* Peak descriptor memory now grows to the high-water mark and is held for process lifetime. A handful
  of descriptors. Non-issue, noted for completeness.

---

Fix B1 and B2 and I expect to approve on the next head — the code itself I am satisfied with.

mattias800 (commented):
Re-review at head `8a02d4cb`. I read the four hunks rather than taking the summary, re-ran the
diagnostic gate, and re-checked the diff scope.

## Verdict: **APPROVED**

Both blocking findings are fixed, and fixed in the right places.

**B1 — `live_compute.cpp:1121-1125`.** Now states the path's own contribution (~144 create + 144
destroy per second) *and* the whole-process rates it was derived from (865/s → 721/s), so the
arithmetic a reader would do now closes instead of contradicting. The added last sentence — "The bulk
of the remaining rate is NOT this path and not `prepare_descriptor_pool`" — is the part that makes the
correction durable rather than merely accurate: it forecloses the wrong inference instead of just not
making it.

**B2 — `live_compute.cpp:1443-1446`.** Moving the note onto `prepare_descriptor_pool` itself is better
than what I asked for. I suggested correcting the PR body; putting it on the helper puts it where
someone chasing the residual will actually be standing, and it states the *reason* (monotone
capacities, `<=` on each, recreates only on a new maximum) rather than just the conclusion, so the
next reader can check it in ten seconds instead of trusting it. The PR body's "Scope — corrected in
review" section matches.

**Diff scope re-checked:** `945adc4e..8a02d4cb` is four hunks in one file, 14 insertions / 6 deletions,
and all four are comments or a pure refactor. The compare path itself is byte-identical apart from
`!reuse` becoming `!compare_pool_reuse_enabled()` at `:1484` — a static member function wrapping a
function-local static, identical semantics to the inline static it replaces (one thread-safe
initialization per process), and now shaped like `descriptor_pool_reuse_enabled()` immediately below
it at `:1389`. The dead `compare_pool = VK_NULL_HANDLE` is gone from the destructor; the destroy at
`:1361` still sits with `descriptor_pool`, before `if (borrowed) return;` and `vkDestroyDevice`.

**So my analysis of reset safety, ownership and capacity carries forward unchanged**, and I am not
restating it — see the review on `945adc4e`. The short form: the reset happens at a point where the
command pool and the fence are already being reset, so the completion invariant was load-bearing
before this change; the one path with no completion proof (device loss) is also the path where no
further dispatch can run; and the capacity check is `<=` on both dimensions, so the pool is always at
least as large as the request even for a future caller with a different sets:buffers ratio.

**Evidence gaps: closed, and closed honestly.** The *Nikoderiko* control arm (1,212/s off → 1,038/s
on, back to back) is the right design, and I particularly want to record that the verification section
now says ctest 315/315 is *not* evidence for this change and names what is. That is the harder thing
to write and the more useful one.

**On the pushback: I agree, don't change it.** Three causes collapsed behind `VK_NULL_HANDLE` is
correct precisely because the caller handles all three identically — the fallback create runs either
way, and a genuine creation failure is still named as `compare-descriptor-pool` in the census. Adding
a distinction the caller cannot act on would be the same class of thing as the `vk_soft_ok` comment
already warns about at `:5398`: a report that names a non-event. Leave it.

**Re-ran `check_diag_gates.py` on this head** — green, and the counts are *identical* to the previous
head (570 files, 737 variables, 1169 sites, 108 findings / 108 baselined), so lifting the env read
into a helper added no finding.

## Non-blocking (neither gates merge)

* **The *Nikoderiko* table row labels a control arm as "master".** 1,212/s is
  `PROSPER_NO_COMPARE_POOL_REUSE=1` on *this* binary, not master — behaviourally identical by design,
  but not the same thing, and the table now carries two different baselines under one column header
  (Plucky Squire's 865 really is master; its control is quoted separately at 854). A row footnote
  would fix it.
* **Dropping the earlier 784 → 702 pair loses a fact worth keeping.** The same "reuse on" condition
  measured 702/s and 1,038/s in two runs — about 48% apart — and the reuse-off condition moved by a
  similar factor. That both arms shifted together, while the within-run delta held its sign and rough
  magnitude (−10.5% vs −14.4%), is the single best argument in the PR that the paired design is doing
  its job and that absolute rates from this interposer are not comparable across runs. One sentence.
* `:1446` cites "(#3157 review)" — #3157 is the issue; the review is on #3159. Trivial.

## Merge

Agreed on waiting for the Vulkan validation scan in the Linux job — it is still `pending` on this
head, and it is the only check in the matrix that can see a descriptor-lifetime VUID. Nothing else
from me is outstanding.

mattias800 (commented):
Correction to my own review immediately above, before it propagates.

I cited `live_compute.cpp:5398` for the "a census that names non-events is worse than no census"
comment. **That line number is wrong.** `:5398` is `return decline(stage);` inside `vk_ok`. The
comment I meant is at **`:5405-5413`**, directly above `vk_soft_ok` at `:5414`.

The point stands unchanged — the reasoning there is exactly why collapsing the three failure causes
behind `VK_NULL_HANDLE` is right, and I still agree the code should not change. Only the line number
was wrong.

Flagging it rather than quietly editing, because this is the failure mode `CLAUDE.md` singles out: a
citation reads as already-verified, and a wrong one attached to a finding the author *wants* to be
true gets checked least of all. I verified every other line I cited in both reviews against this head;
this one I typed from the previous head's numbering without re-deriving it after the +10-line shift
from the four new hunks.

**Verdict on `8a02d4cb` is unchanged: APPROVED.** Restated here so the newest review carries the
verdict. The two non-blocking notes and the merge condition (wait for the Vulkan validation scan in
the Linux job) are as written above.

PR #3160, opened 2026-08-31
perf(compute): measure seed/writeback aliasing between dispatches (#3157)

mattias800:
## Why

`#3157` established that the live compute path's per-dispatch `vkWaitForFences` is the dominant
cost: ~330 us mean against ~171 us of GPU execution, so ~160 us per dispatch is scheduling
latency. At 27 dispatches/frame that is 4.3 ms of a 30 ms frame, and batching the waits sizes at
~13% of frame time.

Starting that implementation showed it is **unsound as usually stated**. A dispatch's writeback
target and the *next* dispatch's seed source are the same guest bytes:

- seed source, `live_compute.cpp:7188`: `resource_bytes_for(r, guest_bytes)`
- and the comment below it: the writeback *"still writes up to `guest_bytes` at `r->gpu_addr`"*

So deferring writebacks to batch the waits lets a later dispatch seed from **stale guest memory** --
silent wrong pixels, not a crash. The per-dispatch wait is load-bearing.

A correct pipeline needs an alias guard that drains on overlap, and whether that is worth building
depends on how often consecutive dispatches actually alias.

## What this adds

`PROSPER_COMPUTE_ALIAS_CENSUS` (default OFF) measures that rate over both binding paths.

**Stray (`PPSA02101`), 90 s offscreen run:**

```
[alias-census] dispatches_with_seeds=43008 aliasing=2653 (6.2%) seed_ranges=87126 write_ranges=61747
```

**~5-6%.** An alias-guarded pipeline drains roughly one dispatch in sixteen, so most of the
~13%-of-frame win survives. That justifies building it.

### Correction: an image-only census understated this 10x

The first version of this branch instrumented only the image path and reported **0.6%**. Covering
storage buffers moved two things:

| | dispatches counted | aliasing |
| --- | --- | --- |
| images only | 2,048 | 12 (0.6%) |
| images + storage buffers | 43,008 | 2,653 (6.2%) |

The rate is ~10x higher, and the population ~20x larger -- buffer-only dispatches declared no seed
range at all and so were not counted. Both numbers support building the pipeline, but they imply
different drain behaviour, and 0.6% is the kind of under-measurement that reads as a green light.
Recorded here rather than silently replaced, because the first figure was briefly published on
#3157.

## Scope and limits

- One title, one route. The rate is a property of the workload, not of prosper; re-measure per
  title before quoting it cross-title.
- The census reports periodically as well as at exit: a bounded run ends in SIGTERM, whose default
  action skips `atexit` handlers entirely, so an atexit-only census yields a clean run and no
  number.

## Verification

- `test_compute_timing_selector`: 52 checks, all pass, exit 0 (7 new -- partial overlap,
  containment in both argument orders, half-open adjacency in both orders, zero-length, identity).
- The overlap predicate is a pure `constexpr` function in the census header, unit-testable away
  from the GPU path.
- Default runs unaffected: every census site is behind `alias_census_enabled()`, and the
  per-dispatch range vectors are only populated when armed.

## Deliberately unaffected

No change to dispatch behaviour, writeback, caching, or submission. This PR measures; it does not
change the wait.

Refs #3157

### DIFF
--- a/BLOG.md
+++ b/BLOG.md
@@ -52,6 +52,24 @@ the render test suite, and the sites are pinned so they cannot quietly grow back
 
 [i3094]: https://github.com/mattias800/prosper/issues/3094
 
+### The compute path waits 19-38% of its wall clock on a fence, and pipelining it cannot help
+
+No picture — this is a negative, and an expensive one to have found the slow way.
+
+Three UE titles spend a fifth to a third of their wall clock inside `vkWaitForFences` on the compute
+path, with `vkQueueSubmit` at about 1% everywhere. That reads as an obvious win: stop waiting on each
+dispatch, overlap them, take the time back. A dispatch ring was built for it and works correctly.
+
+It buys nothing, for a reason no tuning changes. On these routes the guest submits roughly **one
+dispatch per batch** — 23,294 dispatches across ~23,400 batches on *Stray* — and the drain at the end
+of each batch is mandatory, because the guest may read its own memory as soon as the submit returns.
+Every slot is emptied immediately after it is filled, so the ring never overlaps anything. Throughput
+moved 67.5 → 68.5 fps, which is noise; the only real effect was less variance.
+
+The falsification is written down rather than the code being merged, because a default-off switch
+whose own measurement says it can never fire is worse than no switch. What is *not* ruled out is
+deferring across batches — that is where the sized win still lives, and nobody has tried it.
+
 ## 2026-08-31
 
 ### Stray's splash runs 66% faster, and the title screen now holds 65 fps
--- a/prosper/docs/RENDERER_PERFORMANCE_2026_07.md
+++ b/prosper/docs/RENDERER_PERFORMANCE_2026_07.md
@@ -1144,6 +1144,27 @@ resource-ownership scheduling rather than reordering by operation type.
   render passes and layouts, so stable application-level pipeline keys/lifetimes are needed first.
 - Reordering compute and graphics by type is not valid. The corrected capture graph proves real
   producer/consumer edges in the representative frame.
+- **A dispatch ring cannot overlap anything WITHIN a batch, because the end-of-batch drain is
+  mandatory** (2026-09-01, [#3157](https://github.com/mattias800/prosper/issues/3157),
+  implementation [#3161](https://github.com/mattias800/prosper/pull/3161), closed unmerged). The
+  hypothesis was well motivated and the sizing was real: an `LD_PRELOAD` interposer put
+  `vkWaitForFences` at **19.6-37.9% of wall** across three UE titles (Stray `PPSA02101` 37.9%,
+  Little Nightmares III `PPSA05143` 33.0%, Dragon Quest VII `PPSA17942` 19.6%), with `vkQueueSubmit`
+  at ~1% everywhere -- so batching *submits* is dead, and ~160 us per dispatch is scheduling latency
+  with the GPU idle. An alias-guarded ring was built and is correct: 315/315 at depths 1, 3 and 8,
+  zero faults across five interleaved 55 s Stray runs.
+  It does not deliver the sized win, and the reason is structural rather than a tuning failure. On
+  this route the guest submits roughly **one dispatch per batch** -- **23,294 dispatches across
+  ~23,400 batches** on Stray -- and the drain at the end of each batch is **mandatory**, because the
+  guest may read its memory as soon as the submit returns. Every slot is therefore drained
+  immediately after it is filled, so the ring never overlaps anything. Measured throughput: medians
+  67.5 -> 68.5 fps, within noise; the only clear effect is **reduced variance**, and what little
+  gain exists comes from per-slot command pools, not from pipelining.
+  **What this does NOT rule out:** deferring **across** batches. That is where the sized win still
+  lives, and it is untried -- it needs a guest-visibility contract the ring does not have. Do not
+  read this row as "pipelining the compute path is dead"; read it as "a within-batch ring is dead,
+  and the dispatches-per-batch ratio is the number to measure before building the next one."
+
 
 ## Windows cross-submit texture write watch
 
--- a/prosper/docs/RENDERER_PERFORMANCE_2026_07.md
+++ b/prosper/docs/RENDERER_PERFORMANCE_2026_07.md
@@ -1164,6 +1164,9 @@ resource-ownership scheduling rather than reordering by operation type.
   lives, and it is untried -- it needs a guest-visibility contract the ring does not have. Do not
   read this row as "pipelining the compute path is dead"; read it as "a within-batch ring is dead,
   and the dispatches-per-batch ratio is the number to measure before building the next one."
+  That ratio comes from `PROSPER_RENDER_TIMING`, which counts dispatches and batches; the alias
+  census (`PROSPER_COMPUTE_ALIAS_CENSUS`) answers a different question -- whether consecutive
+  dispatches alias through guest memory -- and has no call or batch counter at all.
 
 
 ## Windows cross-submit texture write watch
--- a/prosper/frontends/shared/compute/compute_transfer_gate_census.hpp
+++ b/prosper/frontends/shared/compute/compute_transfer_gate_census.hpp
@@ -210,6 +210,25 @@ struct ComputeGuestRange {
     uint64_t bytes = 0;
 };
 
+// Address 0 is never a guest address, so a range at 0 is never a real one.
+//
+// Extracted and named so the rule is PINNED rather than living as a bare `if (r->gpu_addr)` at two
+// call sites. An explicit null V# entry is bound as a 4-byte zero source backed by a stack-local
+// array, and that synthesis deliberately keeps `gpu_addr == 0` -- its own comment says the point is
+// to bind a null buffer "without inventing a guest address". Feeding it to the census invented one:
+// two dispatches each holding a writable null array entry both produce {0, 4}, which overlaps
+// itself, so the census reported a FABRICATED alias at a non-address.
+//
+// `gpu_addr` rather than `host_data` is the right discriminator, and not only because it is
+// simpler: `resource_bytes_for` returns `host_data` for legitimate MIRRORED guest resources too, so
+// a host_data test would drop real guest ranges. The synthesis sets `host_data` and leaves
+// `gpu_addr` at 0 -- one field separates the cases, and it is this one. It is also the predicate the
+// code that performs the writes already uses, at both
+// `notify_guest_gpu_write_preserving_bytes` call sites, which predate this census.
+constexpr bool compute_guest_range_is_real(const ComputeGuestRange& r) {
+    return r.addr != 0;
+}
+
 // Half-open [addr, addr+bytes) intersection. A zero-length range touches nothing and so can
 // never alias -- that is the `seed_skip`/`renderer_owned` case, where nothing is read from guest
 // memory at all, and counting it as an overlap would overstate the hazard.
--- a/prosper/frontends/shared/compute/compute_transfer_gate_census.hpp
+++ b/prosper/frontends/shared/compute/compute_transfer_gate_census.hpp
@@ -203,6 +203,10 @@ constexpr bool compute_guest_ranges_overlap(const ComputeGuestRange& a,
     return a.addr < b.addr + b.bytes && b.addr < a.addr + a.bytes;
 }
 
+// Plain (non-atomic) counters, deliberately: the census is read and written only from the thread
+// that runs `execute_live_compute_items`, and making them atomic would put a lock-prefixed RMW on
+// a path whose cost this instrument exists to measure. If a second producer thread is ever added,
+// these need revisiting -- they are not thread-safe and do not claim to be.
 struct ComputeAliasCensusCounters {
     uint64_t dispatches = 0;          // dispatches that declared at least one guest seed source
     uint64_t aliasing_dispatches = 0; // ...of those, ones seeding from the previous writeback set
--- a/prosper/frontends/shared/compute/compute_transfer_gate_census.hpp
+++ b/prosper/frontends/shared/compute/compute_transfer_gate_census.hpp
@@ -177,4 +177,37 @@ inline bool claim_compute_transfer_gate_selector_summary(
     return true;
 }
 
+// #3157: a guest-memory range, used to measure whether one dispatch's SEED sources overlap the
+// PREVIOUS dispatch's WRITEBACK targets.
+//
+// Why this exists: the per-dispatch `vkWaitForFences` in the live compute path costs ~330 us
+// against ~171 us of GPU execution, so ~160 us per dispatch is scheduling latency. Deferring the
+// waits (pipelining several dispatches before consuming any results) is the obvious fix, but it
+// is only sound where a later dispatch does not seed from guest bytes an earlier one writes
+// back -- both address `r->gpu_addr` for `guest_bytes`, so an overlap is a read-after-write
+// hazard through guest memory that would silently seed from stale bytes.
+//
+// The overlap RATE bounds the achievable win: if consecutive dispatches usually alias, a
+// correctness-preserving pipeline drains constantly and buys nothing. Measure before building.
+struct ComputeGuestRange {
+    uint64_t addr = 0;
+    uint64_t bytes = 0;
+};
+
+// Half-open [addr, addr+bytes) intersection. A zero-length range touches nothing and so can
+// never alias -- that is the `seed_skip`/`renderer_owned` case, where nothing is read from guest
+// memory at all, and counting it as an overlap would overstate the hazard.
+constexpr bool compute_guest_ranges_overlap(const ComputeGuestRange& a,
+                                            const ComputeGuestRange& b) {
+    if (a.bytes == 0 || b.bytes == 0) return false;
+    return a.addr < b.addr + b.bytes && b.addr < a.addr + a.bytes;
+}
+
+struct ComputeAliasCensusCounters {
+    uint64_t dispatches = 0;          // dispatches that declared at least one guest seed source
+    uint64_t aliasing_dispatches = 0; // ...of those, ones seeding from the previous writeback set
+    uint64_t seed_ranges = 0;
+    uint64_t write_ranges = 0;
+};
+
 } // namespace prosper::frontend
--- a/prosper/frontends/shared/compute/compute_transfer_gate_census.hpp
+++ b/prosper/frontends/shared/compute/compute_transfer_gate_census.hpp
@@ -189,6 +189,22 @@ inline bool claim_compute_transfer_gate_selector_summary(
 //
 // The overlap RATE bounds the achievable win: if consecutive dispatches usually alias, a
 // correctness-preserving pipeline drains constantly and buys nothing. Measure before building.
+//
+// KNOWN BIASES, stated so a reader can weigh the number instead of trusting it. All four were
+// found by review rather than by the instrument, which is itself the reason to write them down:
+//   * UNDER-counts: the sampled DCC fast-clear path reads guest memory through `metadata_addr`
+//     and its writeback writes the same plane, and neither side is recorded (covering it needs
+//     the metadata range, not `gpu_addr`).
+//   * OVER-counts: `compute_transfer_seed_borrowed` bindings are scored as guest seeds, but they
+//     read a retained GPU image rather than guest memory.
+//   * ASYMMETRIC: `SpirvDescriptorBinding::readable` marks a write-only output that need not be
+//     seeded. The image path consults it; `BoundBuffer` copies `writable` but never `readable`,
+//     so write-only IMAGES are excluded from seeds while write-only BUFFERS are included. That
+//     matches what prosper's buffer upload actually does today -- it does not consult `readable`
+//     either, so those bytes really are read -- but the two paths do not agree, and if the upload
+//     ever starts honouring `readable` this census must change with it.
+//   * DEPTH-1 ONLY: overlap is measured against the immediately preceding dispatch, so the rate
+//     is a lower bound for any ring depth greater than one.
 struct ComputeGuestRange {
     uint64_t addr = 0;
     uint64_t bytes = 0;
--- a/prosper/frontends/shared/live/live_compute.cpp
+++ b/prosper/frontends/shared/live/live_compute.cpp
@@ -308,6 +308,41 @@ namespace {
 std::atomic<uint64_t> g_buffer_gpu_result_skips{0};
 std::atomic<uint64_t> g_compute_storage_transfer_seeds{0};
 std::atomic<uint64_t> g_dcc_post_writeback_promotions{0};
+
+// #3157 alias census (PROSPER_COMPUTE_ALIAS_CENSUS=1, default OFF).
+//
+// Measures how often a dispatch seeds from guest bytes the PREVIOUS dispatch wrote back. That
+// overlap is what makes the obvious fix for the per-dispatch fence wait unsound: deferring
+// writebacks to batch the waits would let the later dispatch seed from stale guest memory.
+// The rate bounds the achievable win -- if consecutive dispatches usually alias, a
+// correctness-preserving pipeline drains every time and gains nothing.
+//
+// Compute submits are serialized (one AGC submit executes at a time), so the previous-writeback
+// set needs no lock; the counters are atomic only so the summary can be read from atexit.
+prosper::frontend::ComputeAliasCensusCounters g_alias_census{};
+std::vector<prosper::frontend::ComputeGuestRange> g_alias_prev_writes;
+
+bool alias_census_enabled() {
+    static const bool on = std::getenv("PROSPER_COMPUTE_ALIAS_CENSUS") != nullptr;
+    return on;
+}
+
+void report_alias_census() {
+    const auto& c = g_alias_census;
+    if (c.dispatches == 0) {
+        std::fprintf(stderr, "[alias-census] no dispatch declared a guest seed source\n");
+        return;
+    }
+    std::fprintf(stderr,
+                 "[alias-census] dispatches_with_seeds=%llu aliasing=%llu (%.1f%%) "
+                 "seed_ranges=%llu write_ranges=%llu\n",
+                 static_cast<unsigned long long>(c.dispatches),
+                 static_cast<unsigned long long>(c.aliasing_dispatches),
+                 100.0 * static_cast<double>(c.aliasing_dispatches) /
+                     static_cast<double>(c.dispatches),
+                 static_cast<unsigned long long>(c.seed_ranges),
+                 static_cast<unsigned long long>(c.write_ranges));
+}
 std::atomic<uint64_t> g_dcc_forced_seed_allocation_reuses{0};
 std::atomic<uint64_t> g_dcc_post_writeback_replacements{0};
 std::atomic<bool> g_fail_next_storage_readback_for_test{false};
@@ -5380,6 +5415,9 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
     for (size_t i = 0; i < buffers.size(); ++i)
         buffers[i].descriptor_index = buffer_descriptor_indices[i];
     std::vector<BoundImage> images(image_descriptors.size());
+    // #3157: this dispatch's guest seed sources and writeback targets, collected only when the
+    // alias census is armed so a default run pays nothing.
+    std::vector<prosper::frontend::ComputeGuestRange> alias_seeds, alias_writes;
     std::vector<VkBuffer> staging(image_descriptors.size(), VK_NULL_HANDLE);          // upload/readback
     std::vector<VkDeviceMemory> staging_memory(image_descriptors.size(), VK_NULL_HANDLE);
     std::vector<VkDeviceSize> staging_bytes(image_descriptors.size(), 0);
@@ -7217,6 +7255,13 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                 bi.guest_bytes = guest_bytes;
                 const uint8_t* src = (renderer_owned || bi.seed_skip)
                     ? nullptr : resource_bytes_for(r, guest_bytes);
+                if (alias_census_enabled()) {
+                    // A binding that reads nothing from guest memory cannot alias: `src` is null
+                    // exactly when the seed is skipped or the source is renderer-owned.
+                    if (src) alias_seeds.push_back({r->gpu_addr, guest_bytes});
+                    // The writeback writes up to guest_bytes at r->gpu_addr (see below).
+                    if (bi.storage) alias_writes.push_back({r->gpu_addr, guest_bytes});
+                }
                 // The writeback still writes up to guest_bytes at r->gpu_addr, so a guest-backed
                 // target must be a valid mapped range even when the SEED read is skipped (#1122
                 // review B2): keep the guard for the writeback target, not the seed source.
@@ -8979,6 +9024,25 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
             vkCmdWriteTimestamp(command, VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT,
                                 ctx.dispatch_timestamp_pool, 5);
         if (!vk_ok(vkEndCommandBuffer(command), "command-end")) break;
+        if (alias_census_enabled() && !alias_seeds.empty()) {
+            g_alias_census.dispatches++;
+            g_alias_census.seed_ranges += alias_seeds.size();
+            g_alias_census.write_ranges += alias_writes.size();
+            bool aliases = false;
+            for (const auto& sd : alias_seeds) {
+                for (const auto& w : g_alias_prev_writes)
+                    if (prosper::frontend::compute_guest_ranges_overlap(sd, w)) { aliases = true; break; }
+                if (aliases) break;
+            }
+            if (aliases) g_alias_census.aliasing_dispatches++;
+            // Report periodically as well as at exit: a bounded run is usually stopped with
+            // SIGTERM, whose default action skips atexit handlers entirely -- so an atexit-only
+            // census yields a clean run and no number.
+            if (g_alias_census.dispatches % 2048 == 0) report_alias_census();
+            static const bool once = [] { std::atexit(report_alias_census); return true; }();
+            (void)once;
+        }
+        if (alias_census_enabled()) g_alias_prev_writes = alias_writes;
         VkSubmitInfo submit{VK_STRUCTURE_TYPE_SUBMIT_INFO};
         submit.commandBufferCount = 1;
         submit.pCommandBuffers = &command;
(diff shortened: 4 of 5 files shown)


mattias800 (commented):
# Independent review — REJECTED

Reviewed at head `690339cf`. I read every line I cite. Where I claim a test does or does not pin
something, I mutated and rebuilt; every mutation build is reported with its exit code, and every
mutation is shown to have taken effect before its result is believed.

Summary: the predicate is correct, the OFF path is clean, and the census is a good instrument to
want. But **the seed population it measures is not the seed population the question needs** (two
independent omissions, both biasing the rate the unsafe way), **the zero-length test arm is
vacuous**, and **the `## Rejected experiments` entry misattributes its own instrument and overstates
the PR it cites**. The first and third are the blocking ones.

---

## B1 — The census omits sampled images entirely, and the redundant guard says the author thought it did not

`live_compute.cpp:7258-7264` is inside the `} else if (bi.storage) {` branch opened at `:7201`.
Sampled bindings take the `} else {` at `:7751` and read guest memory — `sampled_guest_source =
resource_bytes_for(r, sampled_guest_need)` at `:6928`, consumed as `src` at `:7753`. Nothing on that
path ever reaches a census site; `grep -n 'alias_seeds\|alias_writes'` returns exactly two push
sites, `:7261/:7263` and `:9034/:9036`.

The tell is `if (bi.storage)` at `:7263`. Inside the `else if (bi.storage)` branch that condition is
unconditionally true. A guard that reads as if it separates two classes, at a site only one class
reaches, is how this survived.

Why it matters more than a coverage gap: the archetypal alias is *dispatch A writes a storage image
at addr X, dispatch B **samples** it at addr X*. That is the producer/consumer pair this very file
has dedicated machinery for — `compute_transfer_seed_borrowed`, `:7100-7113`, "sampled 2D binding
matched exact R32_UINT producer alias". When the borrow hits, the sampled seed comes from a retained
GPU image and is genuinely not a guest alias; when it misses — the common case, and the reason the
gate is a special case at all — it comes from guest memory and is exactly the hazard. The census
cannot see either.

The PR body says the census measures the rate "over both binding paths". It measures storage images
and buffers. Sampled images are a third path.

## B2 — `!b.upload_skipped` is the wrong seed predicate, and it disagrees with the image path

`live_compute.cpp:9033-9034` counts a buffer as a seed only when `!b.upload_skipped`.
`upload_skipped` does not mean "this dispatch does not read guest bytes". It means "the write watch
or the submit-unchanged proof says the cached GPU copy still matches guest memory" — `:1849`
(`submit_unchanged || (watches_complete && dirty_chunks.empty())`) and `:2099`.

Now run the pipeline this census exists to size. Dispatch N's writeback to `[gpu_addr, +guest_bytes)`
is deferred. Nothing has written guest memory, so the watch reports clean, `upload_skipped` becomes
true, and dispatch N+1 binds the cached GPU buffer holding pre-N content. That is precisely the
stale-seed hazard — and this census scores it as "not a seed".

Note the asymmetry: the image path at `:7260` counts a seed whenever the source is guest-backed,
regardless of upload skipping. Two paths, two different definitions of "seed", and the buffer one is
the definition that removes hazards from the numerator.

**B1 and B2 both bias the rate downward**, and the PR's conclusion — "**~5-6%** … most of the
~13%-of-frame win survives. That justifies building it" — is a build/don't-build decision taken on
that number. The PR body already records making this exact error once ("images only reported 0.6% …
0.6% is the kind of under-measurement that reads as a green light"). Same class, still present.

## B3 — The zero-length test arm is vacuous: the guard can be deleted and all 52 checks still pass

`test_compute_timing_selector.cpp:292-294` checks `{0x1000, 0}` against `{0x1000, 0x100}` in both
orders and calls the boundary cases "the point".

Mutation, deleting `if (a.bytes == 0 || b.bytes == 0) return false;`
(`compute_transfer_gate_census.hpp:202`):

```
build rc=0    run rc=0    FAIL lines: 0    "all checks passed"
```

Proof the mutation took effect (same two binaries, a probe outside the test):

```
ORIGINAL header:  interior-zero-length: 0   boundary-zero-length: 0
MUTATED header:   interior-zero-length: 1   boundary-zero-length: 0
```

Half-open arithmetic already returns false for a zero-length range **at a boundary**: `b.addr <
a.addr + 0` is `0x1000 < 0x1000`. The guard is load-bearing only for a zero-length range strictly
**interior** to another — `{0x1080, 0}` vs `{0x1000, 0x100}`, which flips 0 → 1 without it. The one
arm named as the point is the one the predicate does not need.

The arms that *are* load-bearing, verified the same way:

| mutation | build rc | run rc | reddens |
| --- | --- | --- | --- |
| `<` → `<=` (closed interval) | 0 | 1 | "adjacent half-open … do NOT alias", "adjacency is symmetric" |
| drop the second conjunct (one-sided) | 0 | 1 | "adjacent half-open … do NOT alias" |
| delete the zero-length guard | 0 | **0** | **nothing** |

So: half-openness and two-sidedness are pinned; zero-length is not. Add `{0x1080, 0}` vs
`{0x1000, 0x100}` in both orders.

## B4 — The rejected-experiments entry names an instrument that cannot produce its number

`RENDERER_PERFORMANCE_2026_07.md:1156-1159`: *"Measure the dispatch-per-call rate BEFORE building any
batching … the alias census (`PROSPER_COMPUTE_ALIAS_CENSUS`) reports it directly."*

It does not. The census reports `dispatches_with_seeds`, `aliasing`, `seed_ranges`, `write_ranges`
(`live_compute.cpp:336-345`). There is no counter of `execute_live_compute_items` calls in it, so it
cannot form a per-call rate at all.

The instrument that *does* is `PROSPER_RENDER_TIMING`: `timing.dispatches += items.size()` at
`live_compute.cpp:10705` and `[render-timing] compute calls=%llu dispatches=%llu` at `:10712`.

The link is wrong too — it points at `compute_transfer_gate_census.hpp`, which never mentions the env
var. The only two occurrences repo-wide are `live_compute.cpp:312` and `:326`.

A reader following this row to reproduce its central number is sent to a switch that cannot print it.

## B5 — That number appears nowhere else in the project, and conflicts with the PR the row cites

"**0.77 dispatches per call** across Astro Bot, Dragon Quest VII and Blue Prince" is not in #3157, not
in #3161, and not in any comment on either (checked both threads). #3161 — cited in the same sentence
as the implementation — reports **23,294 dispatches across ~23,400 batches ≈ 1.0**, on **Stray**, and
its measured set is Stray / Little Nightmares III / Dragon Quest VII.

So the row states a rate the cited PR does not report, on two titles the cited PR did not measure,
attributed to an instrument that does not report it, with no run, no command and no log anywhere in
the record. It is unfalsifiable as written, which is the one thing a `## Ruled out` row may not be.

## B6 — The row overstates #3161 and closes the route #3161 explicitly left open

Row headline: *"Pipelining compute dispatches behind a ring **cannot help** … the depth knob is
therefore inert **by construction**."*

#3161's own body: *"the end-of-batch drain is mandatory, because the guest may read its memory as soon
as the submit returns. So every slot is drained immediately after it is filled"* — and then:
*"Capturing the sized win requires deferring **across** batches, which needs a guest-visibility
contract this does not have. That is a separate piece of work."*

The row drops the mandatory-drain mechanism, which is the actual reason and is independent of the
per-call rate, and drops the across-batch route entirely. The next agent reads "pipelining compute
dispatches cannot help" in a section whose whole purpose is to stop re-derivation, and stops. Rewrite
to: the *within-batch* ring is dead, here is why (mandatory end-of-batch drain), and the across-batch
variant is untried and needs a guest-visibility contract.

---

## Non-blocking

- **N1.** `live_compute.cpp:321`: "the counters are atomic only so the summary can be read from
  atexit". They are not atomic — `ComputeAliasCensusCounters` (`compute_transfer_gate_census.hpp:206`)
  holds plain `uint64_t`, and `g_alias_census` is a plain object. Make them atomic or delete the
  sentence; a comment asserting a safety property the code does not have is the thing that travels.
- **N2.** `compute_transfer_gate_census.hpp:197-199` justifies the zero-length guard as "that is the
  `seed_skip`/`renderer_owned` case". It is not: those produce a null `src` at `:7260` and push
  nothing, and both call sites already exclude a zero length (`:7250` `!guest_bytes` → `skip_image`;
  `:9032` `b.guest_bytes == 0` → `continue`). The guard is purely defensive, which is fine — the
  stated justification is wrong, and B3 shows it is also the one thing untested.
- **N3.** Buffers dedup `alias_of` (`:9032`); images do not. `seed_ranges`/`write_ranges` therefore
  double-count aliased image bindings and are not comparable across the two paths. The dispatch-level
  bit is unaffected.
- **N4.** Every quoted figure is a periodic report at a multiple of 2048 (`43,008 = 21x2048`;
  `2,048 = 1x2048`), so all counts are floors truncated to a period. The PR body's "population ~20x
  larger" divides a numerator by a denominator whose true value is 2,048-4,095, i.e. 10.5x-21x. Rates
  survive (numerator and denominator truncate together); populations do not. Quote counts as `>=`, or
  install a SIGTERM handler so a bounded run yields a total.
- **N5.** `alias_census_enabled()` (`:325`) is `getenv(...) != nullptr`, so
  `PROSPER_COMPUTE_ALIAS_CENSUS=0` **arms** it. Consistent with the file's convention; the comment at
  `:312` writes `=1`, so say "set to anything".
- **N6.** `a.addr + a.bytes` can wrap, and a wrap returns a false negative — the unsafe direction, and
  the direction the test comment says is the point. Unreachable today (`:7250` caps image
  `guest_bytes` at `UINT32_MAX`, buffer bytes are bounded by the resource), so a comment rather than a
  guard, but say it.
- **N7.** `g_alias_prev_writes = alias_writes;` (`:9057`) runs before `vkQueueSubmit`, and is not
  updated for a dispatch that `break`s earlier, so "previous" can be older than the previous dispatch.
- **N8 (closest to blocking).** The census compares against the **immediately preceding** dispatch's
  writeback set only. A ring of depth N must not alias against the union of the previous N-1 sets,
  which is a strictly larger target, so the measured rate is a **lower bound for any depth > 1** —
  and #3161 ran depths 3 and 8. "Scope and limits" lists title and route but not depth, and "drains
  roughly one dispatch in sixteen" is stated without the qualifier.

---

## Perf-report change

**Verification.** 21 tests, exit 0 at head. Reverting the trigger to `if classification ==
"readback":` gives 1 failure + 1 error, so **both new tests are load-bearing**.

**P1 — the "~39%" is wrong.** `performance_capture_report.py:343` says "a ~39% readback must warn
(the motivating capture)". `measured_total = graphics_total + compute_total` (`:256`), so the
motivating arm is 977.7 / (2000 + 1134) = **31%** — which is what the test's own comment says
(`test_performance_capture_report.py:196`: "readback 31%"). The "2%" half is right, and tightly so: I
verified a threshold of exactly `0.02` reddens `test_non_readback_verdict_carries_no_readback_note`
(readback 2 of 100). Fix the 39%.

**P2 — "both ends pinned" is true only in the weak sense.** Threshold sweep, whole suite each time:

| threshold | 0.011 | 0.02 | 0.021 | 0.05 | 0.10 | 0.19 | 0.20 | 0.201 | 0.31 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| suite | FAIL | FAIL | OK | OK | OK | OK | OK | FAIL | FAIL |

The tests pin the threshold to **(0.020, 0.200]** — a 10x band. They do exclude the two things worth
excluding (any-nonzero, which was the real bug; and anything above the motivating capture), but they
do not pin `0.10`: 0.021 and 0.20 pass identically. So either add arms at 9% / 11% if 10% is the
contract, or drop the claim that the number is pinned and say what actually is.

**P3 — is a share the right trigger at all?** Yes, and your rejection of ranking holds: at 977.7 vs
1134.0 a max-component test is silent on exactly the capture that misleads, and the new test arm
demonstrates that rather than asserting it. I would not have chosen differently. Two things the
broadening did leave behind:

- `test_non_readback_verdict_carries_no_readback_note` (`:214-215`) and its comment "The note is
  specific to the readback verdict; on every report it would be noise" is now false as a general
  statement — it holds only because that fixture's readback is 2%. Update the comment or the next
  reader reads it as the contract.
- The note prose says "before acting on **this verdict**" and "Chase it rather than dismissing it",
  and will now attach to `compute`, `renderer-resource` and `inconclusive` verdicts, where "this
  verdict" is not the verdict the note is about. A one-clause lead when `classification !=
  "readback"` would fix it.

**P4 — scope: yes, split it.** It shares no code, no test binary, no mechanism and no issue with the
alias census; it is report-tooling work off #3152. It carries its own judgement call (a threshold)
that deserves its own reviewer, it is 2 self-contained files, and bundling it means a future bisect
of the census lands on a Python heuristic. Land it separately and it merges in an hour.

---

## What is right, so that finding these means something

- `compute_guest_ranges_overlap` is correct: half-open, symmetric, containment in both orders, and
  the two properties that matter are genuinely pinned (table in B3).
- **The OFF path is clean.** `alias_seeds`/`alias_writes` at `:5420` are default-constructed vectors
  that never allocate; every other site is behind `alias_census_enabled()`, whose first call is the
  only non-trivial one; `g_alias_prev_writes` is untouched; no control flow changes; no shared state
  with the non-census path. I looked for the "instrument perturbs the subject" trap and did not find
  it.
- The periodic-report-plus-atexit design is right, and the SIGTERM reasoning behind it is correct.
- Counting `alias_of` buffers once is sound: `:5668` folds `writable` into the owner
  (`buffers[alias_of].writable |= ...`), so the owner's range covers the alias.
- The image-only → 0.6% correction being published rather than quietly replaced is the right call.

## To clear the rejection

B1 and B2 (measure the seed population the question needs, or state the exclusions and re-derive the
bound), B3 (one interior zero-length arm), and B4/B5/B6 (the doc row: right instrument, reproducible
provenance, and #3161's actual conclusion). N1/N2 are one-line comment fixes worth taking with them.

**REJECTED**

mattias800:
Thorough review — it found a real design gap, a real measurement error and a vacuous test arm, and it
was right to reject. Everything blocking is fixed on `f04ee1b7`. Point by point, including where I
checked your work and found it wrong.

## B1 — sampled bindings omitted. **Confirmed, fixed.**
Verified independently: the push sits inside `} else if (bi.storage) {`, and the `if (bi.storage)`
inside it is unconditionally true — a dead conditional that is its own tell, exactly as you said.
Sampled bindings now contribute a seed range (and no write range, since they never write back). This
is the finding that mattered most: it hid the archetypal alias, A writes storage at X and B *samples*
X, which is the pair `compute_transfer_gate` exists for.

## B2 — `!b.upload_skipped`. **Confirmed, fixed.**
Checked the assignment: `submit_unchanged || (watches_complete && dirty_chunks.empty())`. It means
"the cached copy still matches guest memory", not "does not read guest bytes". Your sharper point is
the one I have put in the code comment — under a deferred writeback the watch reports clean
*because* the writeback has not landed, so the exact hazard was being scored as "not a seed".

## B3 — vacuous zero-length arm. **Confirmed, fixed, and re-mutated.**
Your analysis is right: `{0x1000, 0}` is rejected by half-open arithmetic alone, so the arm pinned
nothing. Now uses an interior `{0x1080, 0}`. Re-ran your mutation against the new arm:

    guard deleted   ctest rc=8   (build clean, 0 errors — the mutation genuinely ran)
    restored        ctest rc=0

## The rate. **Withdrawn.**
You are right that "~6.2% justifies building it" was a build decision taken on a biased number, and
that I had already made this class of error once in this PR. Both corrections move the rate upward by
unknown amounts, so the figure is withdrawn rather than adjusted. It has not been re-measured —
another lane holds the GPU and I will not contend for it. **No claim in this PR now rests on a rate.**

## B4/B5/B6 — the `## Rejected experiments` row. **Substantively right; one of your claims is wrong.**
You wrote the 0.77 figure "exists nowhere in #3157, #3161 or their comments". It does — twice, in my
own closing comment on #3161, posted an hour before your review. That is a distinction without much
comfort, because **your point stands entirely**: my comment is the only place it appears, so there is
no source artifact, and I was restating a number I could not reproduce.

Checking against #3161's actual body, the row was wrong in more ways than you listed:

| I wrote | #3161 measured |
| --- | --- |
| 0.77 dispatches per call | **23,294 dispatches across ~23,400 batches** (~1.0) |
| Astro Bot, DQ7, Blue Prince | **Stray, Little Nightmares III, DQ7** |
| "there is no stream" | the **end-of-batch drain is mandatory** — the dispatches exist |
| pipelining "cannot help", knob "inert by construction" | the **across-batch** route is explicitly left open |
| the alias census reports the rate directly | it has no call counter; the `LD_PRELOAD` interposer produced it |

The row is rewritten from the measured evidence and now carries an explicit *"what this does NOT
rule out"* for the across-batch route. Closing that off would have cost the next agent a live avenue,
which is the specific harm a wrong `## Rejected experiments` entry does. I have also posted a
correction on #3161 striking the four claims in my closing comment.

## Perf report — **split, per your recommendation.** Now #3187.
Also fixed the `~39%`: never computed, and the arm is 31%. I swept the threshold myself and confirm
your band exactly — `(0.020, 0.200]`, so `0.10` is not itself pinned. That limit is now stated in the
code comment, because a reader who believes the constant is pinned will never add the arm that would
pin it.

One method note from the sweep: an intermediate run reported a false `restored_rc=1` from a stale
`__pycache__`. Clearing it per iteration and running `python3 -B` gives the same band and a clean
restore — the instrument, not the code, but worth knowing before anyone repeats the sweep.

## Non-blocking
N1 fixed — the counters are plain `uint64_t` and the comment now says so and why, instead of implying
thread-safety the type does not provide. N8 is a fair characterisation and I have not tried to
argue it away: the census is depth-1, so it is a lower bound for the depths #3161 ran.

ctest **319/319**, exit 0.

mattias800 (commented):
# Re-review — head `f04ee1b7` — REJECTED (narrowly, on two new one-line items)

Everything blocking from round one is fixed and verified below. This rejection is entirely from the
two things you asked me to attack: **both turned up something**, and one of them makes the instrument
report an alias that does not exist. Neither is more than a few lines.

## Round-one findings: verified fixed

| | status | how I checked |
| --- | --- | --- |
| B1 sampled omission | fixed | `live_compute.cpp:7754-7761` pushes a seed and no write range in the sampled `} else {` at `:7751`; the branch is a peer of `if (bi.imported)` / `else if (bi.depth_bits_source)` / `else if (bi.storage)` at `:7192-7201`, not nested in the staging block, so it fires on a persistent cache hit too |
| B2 `!b.upload_skipped` | fixed | `:9041-9049`, predicate dropped; comment records the mechanism |
| B3 vacuous arm | fixed | **mutation re-run**: guard deleted → build rc=0 (so it compiled and genuinely ran), run **rc=1**, sole failure `an interior zero-length range still never aliases`. Head: 53 checks, run rc=0 |
| B4 wrong instrument | fixed | the census is no longer named as the source |
| B5 unfalsifiable figure | fixed | every number in the row traces to #3161's body: 37.9/33.0/19.6%, submit ~1%, ~171 us vs ~330 us, 315/315 at depths 1/3/8, five interleaved 55 s Stray runs, 23,294 / ~23,400, medians 67.5 → 68.5, "per-slot command pools, not pipelining". 0.77 / Astro Bot / Blue Prince gone |
| B6 overstated conclusion | fixed, and then some | the row now *leads* with "within a batch", names the mandatory drain as the mechanism, and carries an explicit "What this does NOT rule out" for the across-batch route |
| N1 atomics comment | fixed | `compute_transfer_gate_census.hpp:206-209` states non-atomic, why, and the condition that would force a revisit — better than what I asked for |
| perf report | split | fully reverted here; `#3187` |

Docs gate green (`check_numbered_table.py` over all tracked `.md`, rc=0). The public correction on
#3161 is posted and strikes the four claims.

---

## First, a correction to my own review — I got the provenance evidence by a broken instrument

I wrote that 0.77 "exists nowhere in #3157, #3161 or their comments (checked both threads)". You are
right that it is in your closing comment, twice. The reason I missed it is worse than a bad grep:

```
$ gh pr view 3161 --comments
GraphQL: Projects (classic) is being deprecated ... (repository.pullRequest.projectCards)
$ gh pr view 3161 --comments | wc -c
206                                    # <- the error text. Zero comments.
$ gh api repos/:owner/:repo/issues/3161/comments --jq 'length'
2
```

`gh pr view --comments` **fails outright on this repository** and emits only the deprecation error, so
piping it to `grep` returns a clean, confident false negative. I asserted a negative from a command
that had errored. That is the same class as the finding I raised against you, and it belongs in the
instrument-trap list: *`gh pr view --comments` is void here; use
`gh api repos/:owner/:repo/issues/<N>/comments`.* My conclusion held only by luck — your comment
being the sole occurrence is the problem, not evidence against it.

---

## C1 (blocking) — a null V# entry fabricates a guest range at address 0, and it can fabricate an alias

Traced end to end at this head:

1. `live_compute.cpp:5399-5406` — `null_entry` (`gpu_addr == 0 && size == 0 && host_data == nullptr`)
   synthesizes `entry_resource.gpu_addr = 0`, `size = sizeof(uint32_t)`, `host_data =
   null_buffer_seed.data()` (a **host-local** `std::array<uint32_t,4>` at `:5375`).
2. `shader_resources.cpp:690-694` — with no zero-pad semantics, `plan_storage_buffer_materialization`
   returns `valid = true`, `logical_bytes = resource.size` = **4**.
3. `live_compute.cpp:5591-5593` — the readability gate passes on `host_data_size (16) >= 4`, so no
   `skip_buffer`.
4. `:5602-5603` — `guest_bytes = 4`, `writable = descriptor.writable` (the reflected binding is shared
   by every entry of the array, so a shader that stores through the array makes the null entry
   writable too).
5. `:9040` lets it through (`resource` set, `guest_bytes != 0`, `alias_of == SIZE_MAX`), and `:9049` /
   `:9051` push **`{0, 4}`** into `alias_seeds` and `alias_writes`.

So two consecutive dispatches that each bind a writable null array entry are scored **aliasing**, at an
address that is not a guest allocation and through bytes that live on this function's own stack. That
is not a bias — it is a fabricated positive, and a systematic one: it fires whenever the pattern
recurs, which for descriptor-array titles is every dispatch.

The comment at `:5372-5374` states the design intent exactly — *"without inventing a guest address or
making the cache believe those bytes have external authority"*. The census then reads the
deliberately-not-invented address as a guest address. The discriminator is already in the same
function. `if (!b.resource->gpu_addr) continue;` at `:9040` closes it; consider the same guard on the
image paths for symmetry.

## C2 (blocking, documentation only) — the DCC fast-clear path is excluded by accident, and the comment gives the wrong reason

You asked precisely this. The answer is: accidentally, and it does read guest memory.

- `:6929` assigns `sampled_guest_source` **inside `if (!sampled_dcc_fast_clear)`**. Under fast clear it
  stays null, so `src && need` is false and the new push at `:7754` is skipped.
- But that path reads guest memory through a *different* pointer: `sampled_dcc_metadata =
  reinterpret_cast<const uint8_t*>(uintptr_t(r->metadata_addr))` at `:6905-6906` (guarded by
  `guest_readable(r->metadata_addr, metadata_bytes)` at `:6902`), and
  `compute_sampled_dcc_fast_clear_rgba8(..., sampled_dcc_metadata, sampled_dcc_metadata_bytes)` at
  `:7769` materializes the **entire** sampled surface from it.
- And the writeback writes that same plane: `std::memset(bi.dcc_metadata, 0xff, bi.dcc_metadata_bytes)`
  at `:9705` followed by `notify_guest_gpu_write(r->metadata_addr, bi.dcc_metadata_bytes)` at `:9713`.

So `[metadata_addr, +dcc_metadata_bytes)` is a genuine producer→consumer channel through guest memory,
and it is absent from **both** sides of the census. The guard `src && need` reads as "src is null
exactly when nothing is read from guest memory" — which is the same inference that produced B1, one
address space over.

I am **not** asking you to implement metadata ranges here; that is real work and the base-bytes census
is useful without it. I am asking that the comment stop asserting the wrong reason and say what is
true: *the DCC fast-clear path reads the metadata plane at `r->metadata_addr`, the writeback writes it,
and neither is recorded — this census covers base bytes only.* A blind spot that is written down is a
scope limit; one presented as a correct exclusion is the next B1.

---

## C3 (non-blocking, but record it) — direct answer to your Q2: the seed-skip concept for buffers *does* exist

Your premise — "`BoundBuffer` has no seed-skip or renderer-owned concept" — is true of the struct and
false of the reflection it is built from.

`SpirvDescriptorBinding::readable`, `src/gpu/resources/shader_resources.hpp:791`, whose own comment at
`:788-790` reads: *"Data access, distinct from merely loading the descriptor object. For buffers these
follow pointer loads/stores ... **Backends use the per-binding result to avoid seeding write-only
outputs** and reading back read-only inputs."*

The image path already uses it: `live_compute.cpp:6276` derives `bi.seed_skip` from
`!image_descriptors[i].readable`, and the storage-image census correctly excludes seed-skipped
bindings via the null `src`. `BoundBuffer` copies `writable` at `:5603` and never `readable`, and no
buffer path in this file consults it — the only three uses (`:6276`, `:8120`, `:8143`) are all images.

**Your change is still right**, and this is the part worth keeping: because prosper's buffer upload
does *not* consult `readable`, prosper itself reads those guest bytes for every bound buffer with a
guest range, regardless of what the shader does — so under a deferred writeback the stale read is real
and the binding is a genuine drain point. Dropping `!b.upload_skipped` bounds the hazard correctly.

But note where that leaves the two paths: a write-only **image** is excluded from the seed census, a
write-only **buffer** is included. The asymmetry has flipped direction rather than gone, and it is now
baked into whatever rate the next run reports. Whoever builds the pipeline will want `readable` on
`BoundBuffer` — say so in the comment so they inherit it.

## C4 (non-blocking) — one upward bias you should name in "Scope and limits"

`bi.compute_transfer_seed_borrowed` bindings push a seed at `:7754` because `src` is non-null, but they
read a **retained GPU image**, not guest memory (`:7100-7113`) — `needs_sampled_upload` at `:7762` is
false for exactly this reason. Counting them over-reports. The irony is worth a line: the borrow is the
transfer-gate producer/consumer pair, so the case B1 was about is now counted as an alias candidate
precisely when it is *not* a guest alias.

## C5 (non-blocking) — the rewritten row still does not name the instrument for its own recommendation

`RENDERER_PERFORMANCE_2026_07.md`: *"the dispatches-per-batch ratio is the number to measure before
building the next one"* — with no instrument named. That was the substance of B4. It is
`PROSPER_RENDER_TIMING`: `timing.dispatches += items.size()` at `live_compute.cpp:10705` and
`[render-timing] compute calls=%llu dispatches=%llu` at `:10713`. One clause.

## C6 (trivial) — the doc diff leaves a double blank line before `## Windows cross-submit texture write watch`.

---

## What is right

- The B1 fix is in the correct place. The sampled `else` is a peer of the imported / depth-bits /
  storage branches, so the two structurally-excluded paths (`bi.imported` = renderer image;
  `bi.depth_bits_source` = borrowed D32 plane, whose comment at `:7196-7199` says touching
  `sampled_guest_source` would reinstate the stale-guest defect) are excluded **by structure**, not by
  accident. Those two I checked specifically because you asked; they are correct.
- Not gating the sampled push on `bi.upload_skipped` is right and makes the two paths agree, which is
  the whole point of the B2 fix.
- Withdrawing the 6.2% rather than adjusting it is the right call. Both corrections move it by unknown
  amounts in opposite directions; an adjusted number would have been a guess wearing a measurement's
  clothes.
- The doc row is now better than the thing it replaced: it keeps the sizing that motivated the work,
  names the mechanism, and its "What this does NOT rule out" paragraph is the part that will actually
  stop the next agent re-deriving this.
- The public correction on #3161 with the four claims struck is the right handling, and doing it before
  asking for re-review is the right order.

## To clear this

C1 (`if (!b.resource->gpu_addr) continue;`) and C2 (comment tells the truth about the metadata plane).
C3's last paragraph and C5's clause are worth taking in the same push. Re-request and I will look only
at those.

**REJECTED**

mattias800 (commented):
# Re-review — head `8ccca5b1` — APPROVED

All findings from rounds one, two and three are fixed and verified against the source. I attacked
both questions you named; **neither turned up a defect that survives scrutiny**, and I am reporting
the reasoning rather than just the conclusion, because two of the candidates look like bugs and are
not — someone will otherwise "fix" them into an understatement later.

One of my own earlier notes was wrong and I retract it below.

## Verified at this head

- `live_compute.cpp:7258` — image arm gated on `r->gpu_addr`; the dead `if (bi.storage)` is gone.
  Behaviour-preserving: `bi.storage` is assigned once at `:5936` from the descriptor kind and never
  reassigned, so inside the `else if (bi.storage)` branch at `:7201` it was unconditionally true.
- `:9053-9062` — buffer arm skips `!b.resource->gpu_addr` before anything else.
- `:7758-7766` — the DCC gap recorded as accidental, naming `metadata_addr`, the writeback that
  writes the same plane, and why covering it needs the metadata range.
- `compute_transfer_gate_census.hpp:192-207` — the four-item bias list. I checked each against the
  source and **all four are accurate**: DCC (`:6905`, `:7769`, `:9705`/`:9713`),
  `compute_transfer_seed_borrowed` (`:7100-7113`, `needs_sampled_upload` false at `:7766`),
  the `readable` asymmetry (`shader_resources.hpp:791`; consulted at `:6276`/`:8120`/`:8143`, all
  images; nowhere on the buffer path), depth-1.
- Doc row names `PROSPER_RENDER_TIMING` and states the census has no batch counter — C5 closed.
- Instrument trap 240 is the right home for the `gh pr view --comments` failure, and verifying all
  three variants before writing it is more than I did.

---

## (a) Is `gpu_addr == 0` the right shape? Yes — and it is better supported than you argued for it

**The strongest argument is one you did not make: the writeback path already uses exactly this test,
twice, and it predates this PR.**

- `live_compute.cpp:9278` — `if (buffer.resource->gpu_addr) notify_guest_gpu_write_preserving_bytes(...)`
- `live_compute.cpp:9475` — `if (r->gpu_addr) notify_guest_gpu_write_preserving_bytes(...)`

Both gate the announcement of a guest write on a non-zero address. So the census now models writes
with the same predicate used by the code that performs them. Consistency with the modelled subject is
the best available justification for a guard's shape, and it is worth putting in the comment — it
turns "address 0 is not a guest address" from an assertion into a citation.

**The alternative you floated — refuse any binding whose `host_data` is not guest-backed — is strictly
worse, and would reintroduce an under-count.** `resource_bytes_for` (`:5551-5555`) returns `host_data`
whenever `host_data && host_data_size >= required`, and legitimate guest resources routinely carry a
host mirror — it is one arm of the readability gate at `:7272` and of the buffer gate at `:5591-5593`.
So "has host_data" is true for the synthesized null **and** for ordinary mirrored guest resources;
discriminating on it would drop real guest ranges. `gpu_addr` separates exactly the synthesized case
because the synthesis at `:5402-5407` sets `host_data` *and* leaves `gpu_addr` at 0 — the two classes
differ on precisely one field, and that is the field the guard reads.

I looked for other synthesized sources and found none: `null_buffer_seed` (`:5375`) is the only
host-local array bound as a resource anywhere in this file.

**The residual weakness, and it is real:** this is a *value* test standing in for a *property*. The
`null_entry` bool is computed at `:5399` and thrown away 3,650 lines before the census reads its
consequence. If a future synthesized binding picks a non-zero placeholder, the guard silently stops
working and the census silently fabricates again — with nothing red.

Cheap carry, no shared-struct change: `buffer_resources`, `buffer_descriptor_indices` and `buffers`
are already parallel and all sized `buffer_plan.total_descriptor_count`, so a parallel
`std::vector<char> buffer_entry_synthesized` filled in the same loop at `:5398-5409` carries the real
property to `:9061`. Worth doing if it is cheap; not worth blocking on.

## On pinning it — yes, there is a way, and B3 is the reason to take it

You are right that reaching the synthesis needs the whole descriptor-array pipeline. But the *rule*
is extractable, and B3's lesson was precisely that an unpinned guard gets deleted:

```cpp
// A synthesized null binding has no guest identity: address 0 is never a guest allocation, and
// {0, N} overlaps ITSELF, so admitting one fabricates an alias rather than biasing a rate.
constexpr bool compute_guest_range_is_real(const ComputeGuestRange& r) {
    return r.addr != 0 && r.bytes != 0;
}
```

Both sites call it (it subsumes the existing `guest_bytes == 0` / `bytes == 0` checks), and the test
pins two arms that must be read together:

```cpp
CHECK(!compute_guest_range_is_real({0, 4}), "a synthesized null binding is not a guest range");
CHECK(compute_guest_ranges_overlap({0, 4}, {0, 4}), "...and {0,4} DOES overlap itself, which is why");
```

The second arm is the one that matters: it makes the *consequence* of deleting the guard explicit in
the test file, so the next reader cannot mistake the guard for defensive noise. It does not prove the
null-V# synthesis reaches the census — nothing short of the full pipeline does, and your commit is
right to say so — but it converts an unpinned magic value into a pinned rule, which is the gap that
produced B3.

## (b) Does the bias list overstate the instrument's honesty? No — I could not break it

Three candidates that look like undeclared over-counts. All three are **correct as modelled**, for one
shared reason worth stating in the header so nobody removes them:

1. **Elided writebacks are still recorded as writes.** `:9268` (buffers) and `:9468` (images, with an
   explicit `continue` at `:9478`) skip the writeback entirely when `gpu_result_unchanged` — guest
   bytes are unmodified, `notify_guest_gpu_write_preserving_bytes` says so in its name — yet the
   census records a write range for those bindings. That looks like a systematic over-count, and the
   population is not small: it has its own counter (`g_buffer_gpu_result_skips`, `:308`, exposed at
   `:10358`) and the whole result-comparator subsystem exists to maximise it.
   **It is nonetheless right.** `gpu_result_unchanged` is only known *after* fence completion — which
   is exactly what a pipeline defers — so an alias-guarded pipeline cannot know in advance that the
   writeback will be a no-op and must drain anyway. The census measures what a pipeline would have to
   do, not what turned out to be necessary.
2. **Full-footprint extents.** `alias_writes` carries the whole `guest_bytes` binding footprint, not
   the subrange actually stored. Same argument: not knowable before completion.
3. **Reflection `writable` rather than observed stores.** Reflection is available in advance and is
   the best prior a pipeline would have, so using it is correct.

The unifying rule — *the census models what is knowable at the pipeline's decision point, not what is
true in hindsight* — is worth one line above the bias list. Without it, each of these reads as an
oversight, and "fixing" any of them would silently understate the drain rate, which is the direction
that gets a wrong pipeline built.

**The one genuinely undeclared item, and it is trivial:** a dispatch that reaches the census and then
fails at `vkQueueSubmit` still increments the counters and still replaces `g_alias_prev_writes` at
`:9080`, so writes that never executed can produce one false overlap for the next dispatch. Bounded by
device-loss frequency. A clause, not a bullet.

## Retraction of my own round-one N3

I wrote that images do not dedup `alias_of` while buffers do, so `seed_ranges`/`write_ranges` would
double-count aliased image bindings. **That was wrong.** `:6613` `continue`s an aliased image binding
before either census site is reached, so images are deduped by construction. Nothing to fix, and it
should not go into the bias list.

## Incidental, pre-existing, not this PR

`:5672-5674` is a ternary whose two branches are byte-identical:

```cpp
const uint8_t* source = buffers[i].atomic_image
    ? resource_bytes_for(resource, buffers[i].guest_bytes)
    : resource_bytes_for(resource, buffers[i].guest_bytes);
```

Present unchanged on `origin/master` at `:5634-5636`, so it is not from this branch. The
`atomic_image` arm presumably once used a different extent — `atomic_slice_bytes` is right there at
`:2941` — and the distinction was lost. Either the condition is dead and should go, or an extent was
dropped and this is a live bug. Worth an issue; explicitly out of scope here.

(It does, incidentally, confirm the bias list's claim: both arms read guest bytes with no `readable`
consultation, so write-only buffers really are seeded today.)

---

## Summary

The instrument is correct, its guard has the right shape and now matches the writeback's own
predicate, and its declared biases are accurate and — as far as I can break it — complete. The two
suggestions worth taking are the extracted `compute_guest_range_is_real` rule with its two-arm test,
and one line stating the knowable-at-decision-time principle above the bias list. Neither blocks.

I did not build or run ctest for this round; your 319/319 stands and CI covers it. I did not touch the
GPU.

**APPROVED**

Issue #1809, opened 2026-08-03
[Game tracker] Astro Bot (PPSA21564)


mattias800:
## Current milestone

**Rung 2 — title screen.** The last explicitly recorded no-input baseline is master `3a473bca`
(2026-08-02): the full opening sequence and ASTRO BOT title card render at native 3840×2160. The
world-map hub currently shows only its nebula backdrop. This is not a gameplay claim.

### Best checked-in visual

[![Astro Bot title screen — Linux screenshot frontend, native 3840x2160 route](https://raw.githubusercontent.com/mattias800/prosper/master/assets/screenshots/astro-bot-title.png)](https://github.com/mattias800/prosper/blob/master/assets/screenshots/astro-bot-title.png)

## What works

- Boots without scripted input through the Sony card, opening cinematic, PlayStation logo, and PlayStation Studios card.
- Renders the ASTRO BOT title card and the animated world-map nebula through the live Vulkan renderer.
- A `PROSPER_NO_COMPUTE=1` diagnostic route proves the guest can accept input, select a level, and load `hub_crashsite_tutorial`; that arm is progression evidence, not render acceptance.

## Current blockers

- [#1732](https://github.com/mattias800/prosper/issues/1732) — guest compute reduces progression from about 67 to about 3 flips/s, putting the interactive hub roughly 17 minutes into a rendered run.
- [#1459](https://github.com/mattias800/prosper/issues/1459) — the rendered hub shows only its backdrop and later alternates with near-black publications.
- [#1730](https://github.com/mattias800/prosper/issues/1730) — long runs end in a host SIGFPE inside RADV around guest flip 1180–1320.
- [#1731](https://github.com/mattias800/prosper/issues/1731) — the `ps_logo` to title handoff publishes a saturated white frame.

## Reproduction and regression route

The opening and title need no input. Run long enough for the title card, which appears around 380 s / guest flip ~1214:

```bash
PROSPER_GUEST_FS=1 PROSPER_GUEST_ARGS=-force-gfx-direct PROSPER_RENDER=1 \
  ./build-linux/screenshot <DUMP_ROOT>/PPSA21564-app0 \
    --seconds 20 --count 40 --timeout 1200 --out ~/astro-title
```

Input timing is flip-sensitive. Use the checked-in route notes rather than reusing old wall-clock
pulses. There is no automatic Astro Bot gameplay snapshot guard yet.

[...]

PR #3162, opened 2026-08-31
perf(compute): cut CPU cost in the compute writeback path (+2.7% on Astro Bot)

mattias800:
## What this is

Two measured CPU wins on the compute writeback path, plus the diagnostic that found them.

**Astro Bot (`PPSA21564`): 10.90 -> 11.19 fps (+2.7%)**, interleaved 55 s runs. ctest green throughout.

### 1. `tile_surface` zeroes only what the copy will not write

It memset the WHOLE destination before copying — a 33 MiB zero-fill per 4K writeback that the copy
then overwrote almost entirely. The fill exists for padding texels in blocks extending past
`width`/`height`, not for texels the copy covers. 64 KiB modes lay blocks out row-major, so when
`width` is block-aligned every block row below `height/bh` is fully covered and needs no zeroing.
Any other mode, an unaligned width, or a non-zero pitch keeps the whole-surface fill.

**Measured, median of 40 iterations per geometry** (corrected — see the note below):

| geometry (mode 27) | pre-fix | this PR | gain |
| --- | --- | --- | --- |
| 3840x2160 bpe=4 | 2.17 / 1.97 ms | 1.17 / 1.16 ms | ~43% |
| 3840x2160 bpe=8 | 3.36 / 3.22 ms | 1.59 / 1.52 ms | ~53% |

**Correction to an earlier revision of this PR body.** It quoted a single "2.43 ms -> 1.72 ms (29%)"
against two *different* geometries in two places — a reviewer caught the inconsistency. Re-measuring
showed two things: the original figure was taken at bpe=4 while the surface that actually dominates
Astro Bot is bpe=8, and the method was contaminated — a 20-iteration **mean** that included
first-touch page faults. A naive re-run of that method even showed the fix as slightly *slower*
(3.23 -> 3.41 ms), which is noise, not signal. Median-of-40 on a warm buffer gives the table above.
The win is larger than originally claimed, and on the geometry that matters.

### 2. The poison-coverage scan stops at the first non-poison unit

Both inner loops kept scanning after setting `all = false`. The common case is `poison_survived=0`
— a covering shader writes every texel, so the verdict is settled at the first unit. Astro Bot scans
108.4 M texels per 70 s run.

### 3. `PROSPER_COMPUTE_STORAGE_GATE_CENSUS` (default OFF)

Reports why a storage surface is or is not a compute-cache candidate, **grouped by geometry**. The
existing `[compute-transfer-gates]` line answers the same question but only for a program pair
selected by hash — which you cannot supply until you already know which surface matters. It is what
located the real cost:

```
3840x2160 bpe=8  evaluated=753 renderer_owned=741 dcc_safe=749 poison_verify=1
                 exact=753 seed_skip=0 persistent=753 CANDIDATE=10
1920x1080 bpe=4  evaluated=3727 renderer_owned=12  ...          CANDIDATE=3701
1024x1024 bpe=8  evaluated=776 renderer_owned=0    ...          CANDIDATE=772
```

The surface that is **75% of all tiling** (48 GiB in 55 s, tiled 766x and detiled only 12x) is a live
render target, and renderer ownership blocks cache admission. Every other geometry admits almost
always.

## Tests

The new tile test is a **guard, not a discriminator** — it would pass against the old whole-surface
memset too. What it catches is this change leaving a padding byte stale, which a round-trip check
cannot see, so it sweeps for a sentinel the source is constructed never to contain.

That detail is load-bearing: my first version of the control filled the source with `(i*7+1)`, which
*produces* the sentinel `0xAB`, and reported 129,600 bytes of false residue against **both** old and
new code.

## Deliberately not included

An attempt to skip the repeated writeback for the renderer-owned target (hash + `GuestWriteWatch`)
**regressed** the title and was reverted — that surface's content genuinely changes every frame, so
no "did this dispatch reproduce the previous bytes" scheme can help it. Measurements and what it
rules out are on #1732.

Refs #1732, #1809, #3149


mattias800 (commented):
APPROVED

Independent review, adversarial per the charter. I built the PR branch from scratch (`origin/astrobot-run`,
head `2ae33dad`), ran the full test suite, and wrote a standalone sentinel-sweep fuzzer against
`tile_surface`/`detile_surface` outside the PR's own test to try to break the tiling change. Details below.

## 1. `tile_surface` targeted zero-fill (the risky change) — no bug found, and I tried hard

**Build + existing tests.** Configured fresh (no `-DGAME_DUMP`), built the full project, ran `ctest
--no-tests=error`: **314/314 passed**, matching the documented fresh-configure baseline. `./test_tile` alone:
all cases pass including the PR's own new round-trip + zero-residue test.

**Static check of the correctness argument.** Traced the claim through the code (`tile.cpp`):
- `sw64kb_dims` gives block dims for every one of the five supported `elem_log2` values (bpe
  1/2/4/8/16); `bw*bh*bpe == 65536` for all five, so the "one 64 KiB block" invariant holds uniformly.
- The fast path only activates when `is_64kb_mode(tile_mode) && pitch == 0 && width % bw == 0`. When
  width is block-aligned and `pitch == 0`, `blocks_x = width/bw` exactly equals the real
  `blocks_per_row` used by `sw64kb_tiled_bytes`/`sw64kb_copy` (no ceiling effect), so the byte range
  `[0, covered)` the PR computes is a genuinely contiguous prefix of the block-major address space —
  this only holds *because* of the alignment guard (see the mutation test below, which breaks exactly
  when that guard is removed).
- `covered_rows = height / bh` (floor) — for every full block row `< covered_rows`, `rows = min(bh,
  eh-y0) == bh` and, since width is aligned, `columns = min(bw, ew-x0) == bw` for every block column,
  so `full_block` is true and the `ToTiled` copy loop in `sw64kb_copy` writes every byte of that block
  unconditionally (`if (full_block || ...)` always takes the write branch). So every byte in `[0,
  covered)` is guaranteed to be overwritten by the copy — the memset skipping that range is sound.
  The `PROSPER_TILE_ROW_MAJOR` fallback loop (`process_rows`) was checked too: it also writes every
  `(x,y)` in `[0,width)×[0,height)`, which covers the same region.
- `covered <= dst_bytes` is checked before trusting `zero_from`, guarding the `dst_bytes - zero_from`
  subtraction against underflow — always true by construction here, but a correct defensive check.
- Non-64KB modes (`Sw4KbS`, `Sw256BS`), unsupported bpe (`sw64kb_elem_log2` returns `UINT32_MAX` for
  anything outside {1,2,4,8,16}), and `pitch != 0` all correctly fall through to `zero_from = 0`, i.e.
  the original full-surface fill.
- `tile_surface_level` (mip-tail path) and `tile_volume`/`detile_volume` (3D path) are untouched by
  this diff and still fully memset — confirmed by reading both, not just the diff context.
- Traced all 4 call sites of `tile_surface` (`live_renderer.cpp:6505`, `live_compute.cpp:9202/9590`,
  `live_compute.cpp:10078`). All pass `pitch=0` in the default case (one passes a debug-only
  `PROSPER_PITCH` override, defaulting to 0); bpe values passed are 4, `sizeof(uint32_t)`, or a
  runtime `guest_texel` that can be 1/2/4/8/16/other — all covered by the sweep below.

**Adversarial sentinel sweep (independent of the PR's test).** Wrote a standalone program
(`/tmp/review-3162/sweep.cpp`, compiled directly against the PR's `tile.cpp` with no other project
dependencies — `tile.hpp` is self-contained). Approach avoids your first attempt's mistake: source is
filled with a fixed byte (`0x31`) that is neither the sentinel (`0xAB`) nor zero, so every destination
byte after `tile_surface` must be either `0x31` (copied real texel) or `0x00` (padding zero) — any
surviving `0xAB` is destination poison that was never touched, i.e. exactly the bug class in question.
Also cross-checked with a full `detile_surface` round-trip against the original linear buffer.

Coverage: **1872 geometries** across all three 64 KiB modes (S/ZX/RX) plus `Sw4KbS`/`Sw256BS`/`Linear`
sanity, all 5 real bpe values (1/2/4/8/16) plus invalid ones (3/5/6/7/12), boundary-precise widths/heights
around every `(bw,bh)` pair (`-1/0/+1` deltas at 1×, 2×, 3× the block dimension), 1×N and N×1 extremes,
zero-width/zero-height, aligned and unaligned `pitch`, and the exact motivating 3840×2160 geometry (and
neighbors off by one in width or height) at full scale for bpe 1/2/4/8/16.

**Result: 0 failures, 0 sentinel residue, all round-trips exact.**

**Positive control, built by hand per the charter's rule** (a same-source control only tests the
discriminator, not the domain) — I did not trust the clean sweep on its own. I mutated two independent
copies of the PR's `tile.cpp` to introduce exactly the bug classes the reasoning above rules out, and
reran the identical sweep unmodified against each:
- Mutant A: `covered_rows = height / bh` → `(height + bh - 1) / bh` (ceiling instead of floor, i.e. treats
  a partially-covered last block row as fully covered). **290 of 1872 cases failed** with real sentinel
  residue (up to 460,800 bytes on the 3840×2160/RX/8bpe case) — and notably `roundtrip_ok=1` on every
  failure, confirming the PR's own reasoning that a round-trip check alone cannot see this class of bug.
- Mutant B: dropped the `width % bw == 0` guard entirely (so the optimization applies to unaligned
  widths too). **792 of 1872 cases failed**, with residue landing exactly where predicted — the first
  (partial) block column of each nominally-"covered" row.

Both mutants are caught hard by the sweep, on the *same* 1872-case matrix the unmutated PR code passes
cleanly. That's the evidence I'd want before believing a clean result: the discriminator demonstrably
fires on cases adjacent to what's shipped, not just in the abstract.

**One thing I could not check**: the Astro Bot fps measurement (10.90→11.19, +2.7%) itself — no game
dump/hardware run was part of this review; I only verified the code change is correct, not the magnitude
of the win. Also did not run the full local-only snapshot matrix (per charter, optional during routine
development).

## 2. Poison-coverage scan early exit

Both changed loops are a pure boolean AND-reduction (`all = all && (byte == poison)`); the value written
to `all` is not otherwise used inside the loop. Breaking on the first mismatch is semantically identical
to scanning every element — confirmed by reading the surrounding code: nothing else in the `for (t...)`
body depends on how many bytes/channels were compared, and the separate `trace`-only nonzero-channel
counter is an independent loop that still fully scans regardless of this change. No behavior change,
verified by the full ctest run (314/314) and, more specifically, `test_game_compute`'s
`[seed-skip-verify] ... poison_survived=...` assertions, which passed unchanged.

## 3. `PROSPER_COMPUTE_STORAGE_GATE_CENSUS`

- Default-off, verified by direct test: `./test_game_compute` with no env var produces zero
  `storage-gate-census` output lines; with `PROSPER_COMPUTE_STORAGE_GATE_CENSUS=1` it produces the
  documented per-geometry report on exit (via `atexit`) and would also fire periodically every 4096
  evaluations — same pattern as the existing `tile_census_note`'s SIGTERM/atexit workaround elsewhere
  in this codebase, so it's consistent with established idiom, not a new risk.
- Thread-safety: `evaluations++` and the `g_storage_gate_rows` map mutation both happen inside the same
  `lock_guard<mutex>` scope; the periodic `dump_storage_gate_census()` call happens *after* that lock is
  released (no self-deadlock), and the `static const bool once = [...]` atexit registration is a C++
  magic-static, safe under concurrent first calls.
- Required headers (`<map>`, `<tuple>`, `<array>`, `<mutex>`) are already included in `live_compute.cpp`.
  `-fsyntax-only` recompile with `-Wall -Wextra` (project's actual flags are only
  `-Werror=switch`) produced no warnings touching this code.
- When on, adds one uncontended branch check per storage dispatch on the hot path; when off, one static
  bool read. Negligible and matches the stated design (mirrors an existing diagnostic's pattern).

## Minor, non-blocking nit

The PR body's motivating measurement is stated with two different geometries for the same numbers:
the top summary says *"8 B/texel, Sw64KbRX: 128x64 blocks, 2160/64 = 33.75"* while the third commit
message (same 2.43ms→1.72ms figure) says *"3840x2160, 4 B/texel, Sw64KbRX: 3840%128==0,
2160%128==112"* (128×128 blocks). Both are internally consistent on their own (bpe=8→128×64,
bpe=4→128×128, per `sw64kb_dims`), but it's not clear which one was actually profiled to get
2.43ms→1.72ms — worth a one-line fix in the PR description, doesn't affect the code.

Also worth noting for future coverage (not a blocker): the PR's own new test in `test_tile.cpp` only
exercises `Sw64KbRX`. My sweep independently confirms `Sw64KbS` and `Sw64KbZX` are equally correct, but
nothing in the committed test suite would catch a future regression specific to those two modes.

## Verdict

**APPROVED.** The tiling change's correctness argument holds under static analysis, the existing test
suite is green (314/314), and a 1872-case adversarial sentinel sweep independent of the PR's own test
found zero stale-padding bytes across every 64 KiB mode, all real and several invalid bpe values,
boundary-precise and extreme geometries, and both aligned/unaligned pitch — while two hand-built mutants
of the exact same code, exercising the two axes the PR's reasoning depends on (height floor, width
alignment guard), are caught immediately by the identical sweep. The other two changes are safe by
inspection and confirmed by direct exercise (early-exit is a no-op boolean optimization; the diagnostic
is default-off, thread-safe, and verified silent/working in both states).

Issue #719, opened 2026-07-14
UE4 (PPSA17942 DOLL): real fragment shader remains black after scene geometry recovery


mattias800:
## Problem

After #718, DOLL's captured UE scene geometry no longer collapses: representative scene draws rasterize coherent on-screen silhouettes with the real vertex shader when paired with a diagnostic magenta fragment shader. Replaying the same draws with their captured real fragment shader still produces black output.

## Reproduction / evidence

- capture: local full DOLL capsule from submit 205 (`/root/doll_prov2_full.prgcap`)
- representative draws: realized draws 21 and 23
- draw 23 previously rasterized zero pixels because a computed structured-buffer VADDR was replaced by the vertex-index path
- after #718, draw 23 produces a coherent small silhouette and draw 21 a large on-screen quad/trapezoid under the diagnostic fragment shader
- the real fragment shader path for the inspected scene draw remains black
- full Linux build and 93/93 tests pass; Messenger and Dead Cells snapshot guards remain green

## Scope

Isolate whether the remaining zero/black output originates in fragment-shader lowering, pixel-input/default interpolation, fragment resources/sampling, fixed-function depth/blend/write state, or the captured producer history. Add a focused offline replay diagnostic/test, then implement the smallest evidence-backed generic fix.

This is the shaded-output follow-up to #718. It should not reopen the now-fixed UE geometry/direct-descriptor path.

mattias800:
CLAIMING: Codex /root | branch fix/issue-719-doll-shaded-output | 2026-07-14T21:10:18.627Z

mattias800:
CONTINUING CLAIM: Codex /root | merged #725; remaining volume-consumer/presentation investigation moved to branch `fix/issue-719-doll-volume-consumer` | 2026-07-15T00:09:00Z

mattias800:
PROGRESS: #732 merged (`be2d248`). DOLL's active PS uses `SPI_PS_INPUT_ENA=ADDR=0x702`; initializing its packed position VGPRs from `FragCoord` changes the boot from black frames to correct 3840x2160 Square Enix and HexaDrive splashes. The captured 75-draw submit replays byte-for-byte against its live oracle (`7359aa8c2962258f`). Continuing the first in-engine presentation/shading stage on `fix/issue-719-doll-presentation`; leaving #719 open for that remaining work.

mattias800:
[...]

Issue #382, opened 2026-07-10
gpu-audit: EUD-resident textures (T# spilled beyond the user-SGPR block) are silently dropped by build_shader_resources — asymmetry with cbuf EUD handling


mattias800:
## Summary
`build_shader_resources` resolves **constant buffers** from either the user-SGPR block *or* the EUD spill area (via `load_sharp`/`eud_base`), but **textures** are only ever read from the SGPR block. A T# whose `offset_dw` lands beyond `num_user_sgprs` is skipped with no attempt to fetch it from the EUD — so its texture binding is dropped and the sampling draw renders untextured (or fails to recompile and is dropped).

## Location
`prosper/src/gpu/agc_shader_layout.cpp:224` (texture loop):
```cpp
uint32_t off = s.offset_dw();
if ((uint64_t)off + 8 > num_user_sgprs) continue;   // T# (8 dwords) must fit in the block
DecodedImageDescriptor d = decode_image_descriptor(&user_sgprs[off]);
```
Contrast the cbuf loop just above (lines 191-212) which calls `load_sharp(off, 4, vv)` and thereby handles the `off >= num_user_sgprs` EUD case. The EUD machinery (`eud_base`, `load_sharp`) already exists in the same function; textures simply don't use it.

## Failure scenario
- A pixel shader with several textures + a paired sampler easily exceeds the ~16-dword user-SGPR block, so the driver spills later T#s into the EUD (exactly the situation #257 fixed for the color *constant buffers*, whose EUD-residency made text read as transparent black).
- For any such spilled T#, `off + 8 > num_user_sgprs` -> `continue` -> no `ShaderResource` of class Texture is emitted -> the shader samples an unbound/garbage texture (or is rejected). Symptom: an untextured / black draw.

## Caveat (why MED-LOW, not higher)
The executor has a *second*, SRT/const-fold texture path (`gpu_executor.cpp` ~660-725) that resolves T#s the shader `s_load`s in-shader, which may cover some EUD textures via a different provenance. But `build_shader_resources`'s own texture path is asymmetric with its cbuf path, and a T# placed directly in EUD (not s_loaded through the tracked SRT path) is dropped here. Worth verifying against the black-level reports (#300 / #320) which are descriptor-suspected.

## Suggested fix
[...]

PR #3165, opened 2026-08-31
gpu/agc: writable T#/U# admitted to the resource table (#3128)

### LINKED ISSUE #3128: agc_shader_layout: a writable T# is dropped by BOTH loops and never enters the resource table (58 of 86 slots on one title)
A **writable T# never enters the shader resource table.** Found while investigating Stray (#3126); filed separately because it is not specific to that title and does not depend on that investigation's outcome.

## The gap

`prosper/src/gpu/agc/agc_shader_layout.cpp:746`, the writable-resource loop over `sharp_resource_offset[1]`:

```c
const AgcShaderSharp& s = writable_buffers[slot];
if (s.empty() || !s.size()) continue;      // size()==0  ->  SKIP
```

Per `agc_shader_layout.hpp:22` and the `tex_drop` comment beside it, `size()` is the guest's declared descriptor width: **1 = four dwords (V#), 0 = eight dwords (T#)**. So this loop skips every texture-shaped entry and handles only buffers.

**Nothing else picks them up.** The texture loop below walks `sharp_resource_offset[0]` — the *read-only* array — so a writable T# declared in category [1] falls through **both** loops and is silently absent from `table.resources`.

## Measured impact on one title

`PROSPER_SHARPLOG=1` on *Stray* (`PPSA02101`) at its title screen, writable slots by declared width:

| `size` | meaning | count |
|---|---|---|
| `0` | **eight dwords (T#)** | **58** |
| `1` | four dwords (V#) | 28 |

**58 of 86 writable slots — the majority — are dropped.** The header for `sharp_resource_offset[1]` says "writable T#/V# resources", so texture-shaped entries there are expected by the format, not anomalous.

## Consequence

A shader op that reads its descriptor from one of those slots resolves nothing. For an image op that means the recompile fails (`ok = false`), and every draw using that shader is discarded at `reason=shader-recompile` — silent lost content, which is the class the charter treats as fatal rather than acceptable.

## Not established

- Whether any title's *rendering* currently depends on one of these. On Stray it does **not**: admitting them as `StorageImage` resources left both `mimg-unresolved` (5) and the rendered output (`max_nonblack` 0.1097) **exactly unchanged** against a deterministic arm (#3127). So this is a latent hole there, not that title's bug.
- Whether the correct class is always `StorageImage`, and how the binding sequence should be assigned given both loops share one `binding++` counter — the probe simply appended, which shifts every later binding.

## Suggested shape

Decode an 8-dword entry in category [1] the way the read-only texture loop already decodes one (`load_sharp(off, 8, …)` → `decode_image_descriptor` → `gen5_image_format` → reject-reason guard), and classify it writable. The read-only path is the reference implementation; the difference is the resource class and the source array.

A `PROSPER_SHARPLOG` drop line for the skipped case would also be worth having — as written the entry vanishes with no trace, which is how it stayed invisible.

mattias800:
## Summary

Fixes #3128. `build_shader_resources` (`prosper/src/gpu/agc/agc_shader_layout.cpp`) walks a shader's
declared resource-usage arrays (`sharp_resource_offset[0..3]`) and produces the `ShaderResourceTable`
the recompiler resolves every memory/image op against. A **writable T#/U#** (an 8-dword image
descriptor declared in category `[1]`, `AgcShaderSharp::size() == 0`) fell through every loop:

- The writable-resource loop over `[1]` deliberately skips `size()==0` entries (`if (s.empty() ||
  !s.size()) continue;`) — correctly, since reinterpreting the first four dwords of a T# as a V#
  fabricates a gigabyte-class buffer range.
- The read-only texture loop, which does the actual T# decode, only ever walks category `[0]`
  (read-only), never `[1]`.

So the slot never entered `table.resources` at all. A shader op reading it then resolves nothing and
the recompiler rejects the whole stage (`[mimg-unresolved]`), discarding every draw that uses it —
which looks identical to a missing recompiler opcode from the outside. Measured on one title
(*Stray*, `PPSA02101`, see #3128 and #3126): 58 of 86 writable slots at the title screen were this
shape, with one stage showing four declared writable T#s and a completely empty resource table.

## Why both loops skipped it (and what changes / stays excluded)

The comment above the writable-buffer loop claimed writable images "remain handled by the dynamic
descriptor fold" — `resolve_dynamic_fetch`'s `SrtUse` mechanism in `gpu_executor.cpp`, which
independently const-folds a shader's own instruction stream to recover descriptors it `s_load`s
through a table pointer. That claim is **true only for the table-loaded case**. A T# the driver
places directly into the entry user-data SGPRs (the ordinary DIRECT case — no in-shader load at all)
has no `s_load` for that fold to trace; the *only* place that knows a descriptor lives at that SGPR
offset is the declared `sharp_resource_offset[1]` array itself. So the DIRECT case had no path to a
resource at all, static or dynamic.

**Fix:** add a new loop over the *same* `sharp_resource_offset[1]` array, decoding exactly the
`size()==0` shape the buffer loop excludes, using the identical pipeline the read-only texture loop
already uses (`load_sharp(off, 8, …)` → `decode_image_descriptor` → `gen5_image_format` →
`image_descriptor_reject_reason` → `image_base_level_view`), classified `ResourceClass::StorageImage`
(read/write, no paired sampler — exactly what `shader_resources.hpp` defines that class to mean, and
exactly what a writable T#/U# is). Both DIRECT (in-SGPR) and INDIRECT (EUD-resident) provenance are
set the same way the read-only texture path already sets them.

**What this does NOT admit:** it cannot re-claim anything the buffer loop already took (that loop
only takes `size()==1`; this loop only takes `size()==0` — the two are disjoint by construction), and
every existing reject reason (`image_descriptor_reject_reason`, unmapped `Gen5 IMG_FMT`, signed BCn,
unsupported mip/array view, implausible backing size) still applies unchanged — a genuinely garbage
or unsupported writable T# still fails visibly instead of being silently bound.

**Binding-number blast radius:** every real consumer of `build_shader_resources` (both graphics
stages via `build_stage_table` in `gpu_executor.cpp`, and both compute call sites) calls
`assign_convention_bindings` immediately afterward, which **reassigns** every resource's `binding` by
class (buffers first, then everything else) rather than trusting `build_shader_resources`'s internal
counter. So inserting this loop shifts nothing about final Vulkan binding numbers in any live path —
only `build_shader_resources`'s own local ordering, which no live caller or test depends on directly
(confirmed: no test in `test_build_shader_resources.cpp` asserts a raw `.binding` value).

## Testing

`tests/gpu/agc/test_build_shader_resources.cpp`:
- Updated the existing `#719` fixture (a real, valid, previously-already-decodable T# placed at
  `sharp[1]` offset 0 with `size=0`) — it used to assert the table stayed **empty**; now asserts it
  produces exactly one `ResourceClass::StorageImage` resource with the correct decoded
  base/width/height/format/tile_mode, resolved by DIRECT `sgpr_base` provenance.
- Added a new EUD-resident writable-T# case, asserting INDIRECT `srt_offset` provenance and that
  `sgpr_base` stays invalid (mirrors the existing read-only EUD texture contract, #382).

**Verified the tests fail without the fix**: `git stash` on just the `agc_shader_layout.cpp` change
(keeping the test file), rebuilt `test_build_shader_resources`, ran it — all 6 new/changed assertions
failed (`[FAIL]`). Restored the fix, rebuilt, reran — all pass, full suite still `== PASS ==`.

Full local suite (in-container, per `CLAUDE.md`):
```
distrobox enter ps5ys -- bash -lc "cd <worktree> && cmake -S prosper -B prosper/build-linux && \
    TMPDIR=\$PWD/prosper/build-linux/tmpdir cmake --build prosper/build-linux -j6"
distrobox enter ps5ys -- bash -lc "cd <worktree>/prosper/build-linux && ctest --no-tests=error"
```
**314 tests, 100% passed, exit 0.**

## Docs

Added a row to `prosper/docs/RESOURCE_BINDING.md`'s `## Ruled out` table recording that the "dynamic
fold handles it" assumption only covers the table-loaded case, not the DIRECT in-SGPR case — so a
future reader doesn't re-derive this from scratch.

## Notes / confidence

- No game dump for *Stray* (`PPSA02101`) is available in this environment, so I could not reproduce
  the exact title-screen measurement from the issue live; the fix is verified against the synthetic
  fixtures above (built from the exact `size=0` shape and the reject-reason pipeline the issue's
  "Suggested shape" section describes) plus the full unit/execution suite.
- `CONFIDENCE: HIGH` that this closes the *structural* gap the issue describes (a declared writable
  T# now always reaches `table.resources` via one path or the other). `CONFIDENCE: LOW` on whether
  it changes any title's *rendered output* today — the issue's own "Not established" section already
  records that admitting these slots left Stray's `mimg-unresolved` count and pixel output unchanged
  in an earlier probe, so this may be a currently-latent correctness fix rather than a visible one.
- **Blast radius**: this touches the shared front-half resource builder used by every title's every
  graphics and compute shader stage. The change is additive (a new loop over previously-dead slots)
  and disjoint from the existing buffer loop by construction, and the full 314-test suite — including
  every recompiler/render execution test — passes unchanged. I did not run the snapshot regression
  matrix (`tools/snapshot/snapshot.py check`); per `CLAUDE.md` that's a release-time gate, not a
  merge requirement, and this PR doesn't change any currently-reachable resource's classification —
  only previously-empty slots gain content.

Fixes #3128


mattias800 (commented):
**APPROVED**

Re-derived the two claims that actually carry this change rather than taking them on trust.

**The mutation arm holds.** I disabled only the new loop's entry condition
(`if (s.empty() || s.size()) continue;` -> `if (true) continue;`) on the PR head, rebuilt
`test_build_shader_resources`, and got **7 failures**; restoring it returns `== PASS ==`. So the
test genuinely discriminates, which matters here because the fixture was an existing one whose
assertion was inverted rather than a fresh file.

**The disjointness argument is sound.** `s.empty() || s.size()` skips both the empty slots and the
`size()==1` V# entries the buffer loop above owns, so the two loops partition
`sharp_resource_offset[1]` by construction rather than by convention. Reusing the read-only path's
exact pipeline (`load_sharp` -> `decode_image_descriptor` -> `image_descriptor_reject_reason` ->
`gen5_image_format`) means nothing bypasses validation — a garbage T# still lands in the same
reject reasons, it now just has somewhere valid to go when it passes. The per-drop `[sharp]`
diagnostic, deduped on `(ud, slot, reason)`, is the right shape for this codebase.

### Three corrections to the report — none blocking, all worth fixing in the record

1. **`assign_convention_bindings` is NOT called "right after" every consumer.** Of the four
   `build_shader_resources` call sites in `gpu_executor.cpp`, only 8691 has it adjacent; 6924 has
   none within 400 lines. The *conclusion* is still correct, but by a different route than stated:
   6924 builds `primary_resources`, which is `std::move`d into `t`, and `t` is reassigned at 7376.
   Worth correcting because the stated reason would let a future reader conclude the property holds
   structurally when it actually holds through one specific data flow.

2. **Also relevant to 6924 and not mentioned:** that table feeds `select_pcrel_dispatch` before any
   binding assignment. That is safe here only because the selector looks resources up via
   `by_sgpr_base_cls` — a keyed lookup, not an index or a count — so admitting previously-dropped
   entries can make a lookup succeed but cannot shift an existing one. That is the argument that
   actually rules out a behavioural change on the PS path, and it belongs in the PR body.

3. **Two factual slips.** The report says ctest was **314**; my run on this exact branch reports
   **315/315 passing**. And it says there was "no Stray dump in this environment" — `PPSA02101-app0`
   is present under the dump root, so the live check was available. Neither changes the verdict,
   but a quoted count is evidence and should be re-taken rather than carried forward.

Good judgment on recording the dead "the dynamic descriptor fold handles it" assumption in
`RESOURCE_BINDING.md` — that comment was actively misleading, and tracing `resolve_dynamic_fetch`
to show it never consults `sharp_resource_offset[1]` is the part of this investigation that will
save the next person the most time.

Verified: 315/315 in-container on this branch, no `Claude-Session:` trailer, no absolute host paths,
three files, +166/-8.

PR #3166, opened 2026-08-31
docs: catch a table row wrapped onto a second physical line (#3088)

### LINKED ISSUE #3088: check_numbered_table.py calls a table unbroken when a wrapped row splits it in GitHub's renderer
`check_numbered_table.py` reports a table as **"unbroken, every row matching its header"** while GitHub renders it broken.

## The gap

A GFM table row ends at the newline. A row whose text is wrapped onto a second physical line therefore renders as **two rows**: the first cut off mid-sentence, the second an orphan carrying whatever was on the continuation line — typically the qualification or the issue links, i.e. the part that mattered.

The gate cannot see this. `prosper/tools/docs/check_numbered_table.py:365` only accepts a row line that *starts* with a pipe, so a continuation line simply **ends the table run**. It is never arity-checked, and the file exits 0.

## Two live instances, found by hand

- **PR #3049** — a scoping clause added to stop a `## Ruled out` row over-claiming wrapped onto a second line, so the answer cell ended on "the page count, not the wider" and the clause plus its `#2790` link became an orphan row. Caught in review, fixed in `ef1f3a09`. The irony is the point: the words added to stop the over-claim were the words that fell out of the cell.
- **`prosper/tools/AGENTS.md:917`** — a paragraph directly follows a table with no blank line between them, so it is absorbed as a table row. Pre-existing on master.

## Detecting it

One awk line finds both, and finds nothing else across every tracked `.md`:

```bash
git ls-files '*.md' | while read f; do
  awk -v F="$f" 'prev ~ /^\|.*\|/ && $0 !~ /^\|/ && NF > 0 {print F": "NR": "$0} {prev=$0}' "$f"
done
```

## Suggested fix

Teach the gate that a non-blank, non-pipe line immediately after a table row is an error, not the end of the table. Two shapes to distinguish, and both are defects:

- a wrapped row (the continuation is the tail of a cell), and
- a paragraph that forgot its blank line (the continuation is prose).

Either way the rendered table is wrong, so a single check covers both. Worth a ctest case built from a fixture that renders wrong, since the current gate's own green result is what makes this invisible — the check reports on structure it can parse, and silently narrows the table to the part it could.

mattias800:
## Summary

`check_numbered_table.py` required every table row to start with `|`. GFM does not: once a
table's header and delimiter are seen, GitHub keeps consuming every following non-blank line as
another row of the *same* table, pipe-prefixed or not, until a blank line or another block
construct ends it. A row wrapped onto a second physical line -- or a trailing paragraph that
forgot its blank line -- therefore fell out of `parse_tables` entirely: never counted as a row,
never arity-checked, and (unless something pipe-prefixed happened to follow it) left no trace at
all. The file read `"unbroken, every row matching its header"`, exit 0, while GitHub silently
truncated the row and orphaned the continuation as a garbled second row.

## Root cause and fix

`parse_tables` treated any non-`|`-prefixed line as ending the current table run. Fixed by
absorbing a non-blank, non-pipe line into a table that is already **open** (header + valid
delimiter already matched) instead of ending the run -- matching GFM's real continuation rule.
This lets the pre-existing ARITY class see the row and report the cell-count mismatch a wrap or
stray paragraph almost always produces.

The one thing that must NOT be absorbed is a line that starts its own block construct (an ATX
heading, blockquote, bullet/ordered list item, thematic break, or HTML block) -- GFM lets each of
those interrupt an open table without a blank line too, the same way they interrupt a plain
paragraph, and none of them become a table row. `interrupts_table()` recognizes these; each shape
was checked against GitHub's own renderer (`POST /markdown`, mode=gfm) before and after the fix,
not reasoned from the spec alone.

**A necessary side effect, worth calling out:** the old STRUCTURE message ("did it lose its
leading `|`?") fired on *every* non-`|`-prefixed row, including one whose interior pipes still
land on the header's cell count -- which GitHub actually renders as a completely normal, unbroken
row. That was a pre-existing false positive sharing the same code path; fixing the blind spot
necessarily corrects it too. The old test asserting the wrong behavior is kept, corrected, rather
than deleted (see `tools/docs/test_check_numbered_table.py`, the case formerly named "a row that
lost its leading pipe ends the table").

## Reproduction (before the fix)

Constructed by hand against current master before making any change:

```
| Question | Answer |
| --- | --- |
| Does X leak? | no, bounded by the page count, not the wider
context, per the #2790 fix |
```

```
$ python3 tools/docs/check_numbered_table.py repro.md
repro.md: 1 table(s), 1 rows, unbroken, every row matching its header
$ echo $?
0
```

GitHub's renderer on the same input produces a genuine second, garbled row (`<td>context, per the
#2790 fix</td><td></td>`) -- confirmed via the `/markdown` API before writing the fix. This is
exactly the PR #3049 shape the issue names.

## tools/AGENTS.md also fixed (one blank line)

The repo-wide sweep (`git ls-files '*.md' -z | xargs -0 check_numbered_table.py`) finds the
second instance the issue names as pre-existing: `tools/AGENTS.md`'s paragraph at line 940 forgot
the blank line separating it from the 3-column table above it, so GitHub renders the whole
following paragraph (~130 lines) as single-cell rows appended to that table. The fixed checker
reports exactly 128 problems here and nowhere else in the corpus. Since this sweep is a
**required CI job on every PR** (not just docs PRs), landing the checker fix without repairing
the document would turn the `Docs` job red on master immediately. Fixed with one blank line.

## Regression test

Added to `tools/docs/test_check_numbered_table.py`:
- `a row wrapped onto a second physical line is caught (#3088)`
- `a paragraph directly below a table with no blank line is caught (#3088)`
- Six counter-arms proving each recognized block construct (heading, bullet list, ordered list,
  blockquote, thematic break, HTML block) is *not* falsely absorbed
- The corrected false-positive case, its inverse (arity actually mismatches -> still caught), and
  a `--ordered` arm documenting the one residual case (a lost leading pipe still defeats
  `LEADING_NUMBER`, so `--ordered` alone still flags it as "no numbered table found" even when
  ARITY is clean)
- `a duplicate hidden below an interruption cannot pass green`, re-pointed at a blank-line break
  (the mechanism the original 2026-08-01 incident actually was) since a missing leading pipe no
  longer ends an open table by itself

**How I verified the new tests fail without the fix:** saved `git HEAD`'s copy of
`check_numbered_table.py` to a scratch directory and ran the three key new fixtures'
`check()` calls against it directly. All three produce a **different** verdict pre- vs post-fix:

```
a row wrapped onto a second physical line is caught (#3088):
  fixed checker expects want_problems=True
  OLD (pre-fix) checker: has_problems=False, problems=[]
  -> DIFFERENT verdict: this fixture is a genuine regression test.

a paragraph directly below a table with no blank line is caught (#3088):
  fixed checker expects want_problems=True
  OLD (pre-fix) checker: has_problems=False, problems=[]
  -> DIFFERENT verdict: this fixture is a genuine regression test.

a row missing its leading pipe still renders correctly when its arity matches:
  fixed checker expects want_problems=False
  OLD (pre-fix) checker: has_problems=True, problems=[".../case.md:4: line is not a table row
  but sits inside the table that starts at line 1, ending it ... Did it lose its leading '|'?
  Line reads: '2 | b |'"]
  -> DIFFERENT verdict: this fixture is a genuine regression test.
```

## Verification

- `python3 tools/docs/test_check_numbered_table.py`: **104 cases, exit 0** (up from the
  pre-existing suite plus 13 new cases; two pre-existing cases corrected as described above).
- Repo-wide sweep, matching CI's `Docs` job exactly (`git ls-files '*.md' -z | xargs -0 python3
  tools/docs/check_numbered_table.py`):
  - **Before** (pre-fix checker, against a `git archive` of the branch's base commit, 146 tracked
    `.md` files): exit 0, 0 problems -- the blind spot in action.
  - **After fixing the checker only** (still-broken `tools/AGENTS.md`): exit 123 (xargs), **128
    problems, all in `tools/AGENTS.md`**, all one defect.
  - **After fixing `tools/AGENTS.md` too** (final state of this PR): **146 files, exit 0**.
  - Confirmed no other file in the corpus disagrees between the old and new checker -- the AGENTS.md
    instance is the only behavioral difference across the whole tracked corpus.
- CI's other `check_numbered_table.py` invocation, run locally against the exact recipe
  (`--ordered --table-header Instrument --baseline <origin/master's copy>
  docs/GAME_COMPAT_ORCHESTRATION.md`): exit 0, 239 rows, unique and ascending, none of the
  baseline's 239 rows deleted.
- `git diff --check`: clean.

This is a documentation-tooling-only change (`tools/docs/*.py` plus a one-line `tools/AGENTS.md`
fix) with no C++ build dependency, so no `ctest`/build verification was run.

Fixes #3088

### DIFF
--- a/prosper/tools/AGENTS.md
+++ b/prosper/tools/AGENTS.md
@@ -937,6 +937,7 @@ force:
 | `PROSPER_CAPTURE_MAX_SUBMITS` | **uncapped** — the inverse of the request | **the process refuses to start**, exit status 3. There is no honest default: `0` means uncapped, so a typo removed the only content bound on exactly the run that needed it. Express "uncapped" by leaving the variable unset. |
 | `PROSPER_CAPTURE_BUNDLE_MAX_MB` | the default budget, so a raise was discarded | same value in force, plus one `[grab]` line saying the raise was discarded (or naming the 64/3072 MiB bound it was clamped to) |
 | `PROSPER_CAPTURE_FRAMES` | 1 — the very width the `window had no submits` message tells you to widen | same value in force, plus one `[grab]` line; a mistyped remedy no longer reproduces the original failure in silence |
+
 If present counts vary and the title emits no honest guest-stdout marker, use the headless F9 control:
 set `PROSPER_CAPTURE_BUNDLE_TRIGGER_FILE=/path/capture.ready` together with
 `PROSPER_CAPTURE_BUNDLE=/path/frame.prgbundle`, keep the trigger absent at process start, and create it
--- a/prosper/tools/docs/check_numbered_table.py
+++ b/prosper/tools/docs/check_numbered_table.py
@@ -29,6 +29,47 @@ the first data row, which orphans the entire body.
 Fenced code blocks are skipped: this repository's docs paste tool output containing pipe
 characters, and this file's own defect example would otherwise fail the check that documents it.
 
+A THIRD BLIND SPOT (#3088, 2026-08-31): a row wrapped onto a second physical line. GFM does not
+require a table row to start with `|` -- once the header and delimiter are seen, EVERY following
+non-blank line becomes another row of the SAME table, pipe-prefixed or not, until a blank line or
+another block construct ends it. `parse_tables` used to require every row to start with `|`, so a
+continuation line simply fell OUT of the run: it was never counted as a row, never arity-checked,
+and (unless something pipe-prefixed happened to follow it) left no trace at all -- the file exited
+0, "unbroken, every row matching its header". PR #3049 lost a qualifying clause and its `#2790`
+link this way; `tools/AGENTS.md:917` still carries the same shape pre-existing on master.
+
+MEASURED against GitHub's own renderer (`POST /markdown`, mode=gfm), not reasoned:
+
+    | abc | def |            renders bar as a genuine second row (one cell, `def` blank) when it
+    | --- | --- |            directly abuts the table, but as an ordinary <p> when a blank line
+    | bar | baz |            separates them -- this is the GFM spec's own worked example, and it
+    bar                      is what makes a wrapped continuation dangerous: nothing SYNTACTIC
+                             marks it as wrong.
+
+    | # | What |             a row missing ONLY its leading pipe, with an interior/trailing pipe
+    | --- | --- |            count that still matches the header, renders as a completely normal,
+    | 1 | a |                unbroken 3-row table -- so the pre-#3088 STRUCTURE message ("did it
+    2 | b |                  lose its leading '|'?") was itself a FALSE POSITIVE on this shape,
+    | 3 | c |                fixed as a consequence of fixing the blind spot above (see the
+                             corrected test, formerly "a row that lost its leading pipe ends the
+                             table").
+
+    | # | What |             a genuine block construct -- ATX heading, blockquote, bullet/ordered
+    | --- | --- |            list item, thematic break, HTML block -- interrupts an open table
+    | 1 | a |                WITHOUT needing a blank line, the same way each interrupts a plain
+    ## Heading               paragraph: it renders as its own block (`<h2>Heading</h2>` here), not
+                             as an absorbed row. `interrupts_table()` recognizes these so the fix
+                             does not swallow ordinary trailing prose sections into the table above
+                             them.
+
+So the fix is not "any non-pipe line ends the table" (the old rule, too strict -- misses the
+wrap) and not "any non-pipe line is absorbed" (too loose -- would report a false break on the
+missing-leading-pipe case above, and would silently swallow a heading/list/quote/rule/HTML block
+that legitimately follows a table with no blank line). It is: once a table is OPEN (header +
+valid delimiter matched), a non-blank line that does not start another recognized block is
+ABSORBED as another body row, exactly as GFM renders it -- letting the pre-existing ARITY class
+see it and report the cell-count mismatch a wrap or a stray paragraph almost always produces.
+
 FOUR CLASSES OF CHECK, deliberately separated:
 
   STRUCTURE (always on) -- no blank line may split a table. A fact about Markdown, true of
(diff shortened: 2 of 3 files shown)


mattias800 (commented):
**APPROVED**

This PR changes a CI gate and deletes pre-existing tests, so I re-derived the claims rather than
reading the report.

**It closes a real blind spot, on a real file.** Against `origin/master`'s own
`prosper/tools/AGENTS.md`: the old checker exits **0** with "1 table(s), 3 rows, unbroken, every row
matching its header"; the new checker exits **1** with **128 problems**. That is not a synthetic
fixture — the defect was live in the tree and invisible.

**No false positives, which is the real risk here.** A gate that over-rejects blocks every PR. Over
all 146 tracked markdown files on this branch both checkers exit 0, and diffing their per-file
output shows **zero disagreements** — the only file where they ever differed is the `AGENTS.md`
this PR repairs.

**The scope expansion was necessary, not optional.** The one-line blank-line fix to `AGENTS.md` had
to land in the same PR: this sweep is a mandatory CI job on every PR, so shipping the fixed checker
without it would have reddened master immediately. Correct call, and correctly explained.

**On deleting the two existing tests** — the part I looked hardest at, since editing tests to make a
fix pass is how a regression hides. The removed
`"a row that lost its leading pipe ends the table"` asserted the *old* model, and its own comment
conceded the consequence: a duplicate below the break "would pass green -- which is the defect this
whole check exists to prevent". So it encoded the bug as expected behaviour. Deleting it is right,
not convenient. I confirmed the replacement model still fails that shape: with a wrapped
continuation line above a duplicate row, the new checker exits 1 (arity mismatch at the absorbed
line), as does the old one by its own different route.

**One limit on this approval, stated plainly:** the GFM absorption semantics — which lines GitHub
lets interrupt a table without a blank line (headings, blockquotes, lists, thematic breaks, HTML) —
rest on the author's checks against GitHub's `/markdown` render API. I did not re-run those. If any
one of those shapes is wrong, the failure mode is a false positive on a real doc, which the
repo-wide sweep above would only catch for shapes present in the tree today. Verifying against the
real renderer rather than reasoning from the spec was the right method; I am recording that I took
it on trust.

Verified: 104 test cases pass, repo-wide sweep exits 0 with no disagreements, no `Claude-Session:`
trailer, no absolute host paths, three files, +197/-15.

PR #3167, opened 2026-08-31
recompiler: widen MIMG zero-mip shape for Stray's IMAGE_LOAD_MIP NSA packet (#3134)

### LINKED ISSUE #3134: recompiler: MIMG op=0x1 with an extra address dword is rejected (recompile-reject-mimg-address extra=1)
An MIMG instruction whose address operands exceed what the emitter expects is rejected, failing the
whole stage.

## Evidence

*Stray* (`PPSA02101`), title-screen frame, replayed offline
(`gpu_replay <capsule> --retry-failed-stage 3:1`, `PROSPER_DBG=1`):

```
[recompile-reject] sh=bfa00003/685 mode=unresolved-operand pc=134
                   words=f004070a,00080500,0000052a fmt=14 op=0x1
                   dst=5(kind2) src=0(k2),32(k1),0(k1) dmask=0x7 dim=1 glc=0 len=3
                   stage=fragment role=terminal
[recompile-reject-mimg-address] pc=134 extra=1 words=0000052a,00000000,00000000
```

`fmt=14` is MIMG, `op=0x1`, `dmask=0x7`, `dim=1`, and the instruction is **three dwords** — it carries
an extra address dword the emitter does not consume (`extra=1`).

## Impact

Fragment `0x30be800000` (685 words, 13 resources) fails to recompile, taking its draw with it. In the
frame measured, that draw is one of eleven unrealized operations, and the `shader-recompile` drops in
that frame are all **full-screen 3840×2160** passes.

## Reproduction

The capture reproduces deterministically with no boot:

```bash
gpu_replay <capsule>.prgcap --inspect-only                       # lists failure[3]
PROSPER_DBG=1 gpu_replay <capsule>.prgcap --retry-failed-stage 3:1
gpu_replay <capsule>.prgcap --dump-failed-shader 3:1 fs.bin      # the raw guest program
```

mattias800:
## Summary

Refs #3134, #2818. Not `Fixes` -- see below for why this does not close the issue.

`rdna2_mimg_zero_mip_shape()` (the packet-shape recognizer that lets the recompiler
identify and specialize away a provably-zero `IMAGE_LOAD_MIP`/`IMAGE_STORE_MIP` mip
operand) required `UNORM=1` and `GLC=1` unconditionally, matching only GTA V's
evidenced encoding. #3134 reports a live Stray (`PPSA02101`) fragment reject on an
`IMAGE_LOAD_MIP` with an NSA extra address dword that the function did not recognize
at all (`shape=0` in the `[mimg-mip]` diagnostic).

**Byte-verified via llvm-mc gfx1030**: `image_load_mip v[5:7], [v0, v42, v5], s[32:39]
dmask:0x7 dim:SQ_RSRC_IMG_2D` assembles to the exact rejected words
(`f004070a,00080500,0000052a`) with **no** `unorm`/`glc` modifier in the source --
i.e. `UNORM=0`/`GLC=0` is the plain-assembly default for this opcode, not something
GTA V's compiler added. So the function's top-level guard was rejecting a real,
correctly-encoded instruction before ever reaching the per-opcode shape checks.

## What this PR does

Moves the `UNORM`/`GLC` check out of the shared top-level guard and into each
admitted shape individually (matching what each was actually verified against), and
adds a third shape for this exact packet: opcode `0x01`, `nsa=1`, `len_dwords=3`,
`dim=1` (2D), `dmask=0x7`, `UNORM=0`/`GLC=0` -- same NSA arity/dword as
`IMAGE_STORE_MIP`'s existing NSA sibling, one opcode over, per the issue's own
suggested precedent.

Neither `UNORM` nor `GLC` affects which VGPR carries the mip or the packet's address
layout: `UNORM` only changes coordinate interpretation for a filtered *sample*
(every path this function feeds is already integer/texel-space by construction --
`LOAD`/`STORE`, never a sample), and `GLC` is a cache-coherency policy the
per-invocation model does not simulate.

## Why this does NOT close #3134 -- verified against a live capture, not assumed

I re-derived the reject against the real Stray title-screen frame
(`s14072.prgcap`, an F9 bundle of the exact defect) rather than trusting the issue's
own suggested fix shape:

```
PROSPER_DBG=1 gpu_replay --retry-failed-stage 3:1 s14072.prgcap
```

Before this change:
```
[mimg-mip] image_load_mip declined pc=134 shape=0 proven_zero_mip=0 img_dim=1/1
           mips=6 mip_tail=1 ... dmask=0x7 unorm=0 glc=0
```

After this change:
```
[mimg-mip] image_load_mip declined pc=134 shape=1 proven_zero_mip=0 img_dim=1/1
           mips=6 mip_tail=1 ... dmask=0x7 unorm=0 glc=0
```

The shape is now correctly recognized (`shape=1`), but the shader **still rejects**,
because the resource this instruction addresses genuinely declares **six real mip
levels**, is in the **mip tail**, and the guest's mip register is **not provably
zero** (`proven_zero_mip=0`) -- unlike every previously-evidenced shape, which all
address genuinely single-materialized-mip resources. The "prove the mip is zero,
then discard it" fast path this shape function feeds simply does not apply here.

This is the **same root cause** #2818 already found and parked for Sonic Frontiers'
compute-side `IMAGE_LOAD_MIP` (a 12-mip 2D_ARRAY resource): prosper's graphics
texture path only ever *generates* a mip chain via a `vkCmdBlitImage` downsample
cascade of level 0, gated to `VK_FORMAT_R8G8B8A8_UNORM`
(`tests/fixtures/render_runner.h`) -- it never uploads the guest's own per-level
bytes. Stray's resource (`R16G16_UINT`, decoded format 7/ncomp=2) cannot even use
that generation path: an **integer** format has no meaningful linear-filtered
downsample, and Vulkan does not expose `SAMPLED_IMAGE_FILTER_LINEAR` for
pure-integer formats regardless. Per #2818's own explicit rule, forcing `Lod=0` or
reading a blit-synthesized chain here would both render a **wrong** image -- worse
than the current visible reject -- so I did not take that shortcut. A faithful fix
needs the shared guest-authored multi-level upload infrastructure #2818 already
scoped (real `mipLevels`, a per-level copy from the guest's own tiled bytes, and an
`OpImageFetch` with an explicit `Lod`), which is out of scope for a single-title
NSA-shape fix and carries its own independent-review requirement per `CLAUDE.md`
(shared renderer/executor upload path every title's textures go through).

Falsification recorded in both `docs/RECOMPILER_REMAINING.md` and
`docs/STRAY_STATUS.md`, each `## Ruled out`, with the live evidence above.

## Why land the shape widening anyway

- It makes the CPU-side `proven_zero_mip_at_use` dataflow proof
  (`gpu_executor.cpp`) accurate for this encoding instead of short-circuiting on
  packet shape before it can even attempt the proof.
- It may unblock some *other* draw (in Stray or another UE4 title sharing this
  compiler's `UNORM=0`/`GLC=0` encoding) that issues the same instruction shape but
  addresses a genuinely single-mip resource with a provably-zero mip register --
  exactly the case this fast path exists for.
- It is the documented, precedented, narrowly-evidenced shape of fix the issue
  itself asked for, and is safe: admission stays scoped to the exact byte-verified
  packet, with no change to any other opcode/dim/dmask combination.

## Testing

New round-trip test in `tests/gpu/recompiler/test_rdna2_decode.cpp` against the
exact captured bytes:

- **Fails without this change** -- confirmed by stashing the `rdna2_decode.cpp`
  change and rebuilding: `[FAIL] Stray IMAGE_LOAD_MIP NSA 2D identifies its explicit
  v5 mip operand with UNORM/GLC clear`, ctest-equivalent exit 1.
- **Passes with it**, plus two mutation arms proving the widening does not
  accidentally admit an unevidenced `UNORM`/`GLC`-set or `dmask=0x3` variant of the
  same address bytes.

Build + full suite in the container (`distrobox enter ps5ys`):

```
cmake --build prosper/build-linux -j6   # exit 0
ctest --no-tests=error                  # 315/315 passed (100%), exit 0, 86.01s
```

`prosper/docs/GAME_COMPAT_ORCHESTRATION.md`'s docs gate:
```
check_numbered_table.py prosper/docs/RECOMPILER_REMAINING.md prosper/docs/STRAY_STATUS.md
-> RECOMPILER_REMAINING.md: 3 table(s), 35 rows, unbroken, every row matching its header
-> STRAY_STATUS.md: 12 table(s), 39 rows, unbroken, every row matching its header
```

`git diff --check`: clean.

## Risk

Behavior-neutral for every currently-passing shader: the new admission branch is
gated on an exact opcode/NSA/dim/dmask/UNORM/GLC combination that previously always
returned `false`, so nothing that used to be recognized changes, and nothing that
used to be rejected can newly reach a *different* code path than "still correctly
rejected" (verified above on the live capture) unless it also satisfies the
existing `res->declared_mip_levels==1` / `proven_zero_mip` gates in
`rdna2_emit_alu.cpp`, which are unchanged.


mattias800 (commented):
**APPROVED**

The most valuable thing in this PR is that it **refutes the fix the issue asked for**, and says so
instead of shipping something that would have looked like a fix.

**On the refutation.** The issue proposed widening `rdna2_mimg_zero_mip_shape()` by analogy with the
`IMAGE_STORE_MIP` NSA precedent. The author obtained the referenced capture, reproduced the reject,
confirmed the bytes by llvm-mc gfx1030 round-trip — and then found the analogy does not hold: the
resource carries **6 real mip levels**, sits in the **mip tail**, and `proven_zero_mip=0`, where
every previously-evidenced shape addressed a genuinely single-mip resource. Widening the shape makes
the CPU-side proof report `shape=1`, and the stage **still correctly rejects**. Landing it as
`Fixes #3134` would have closed an open defect on a change that does not fix it. `Refs` plus leaving
the issue open, with the `in-progress` claim removed, is exactly right.

**The refactor is behaviour-preserving, which is what makes this safe to land.** Moving
`mimg_unorm`/`mimg_glc` out of the shared top-level guard and into per-branch conditions leaves ops
`0x01` and `0x09` requiring `unorm && glc` exactly as the old guard did, and the new branch is
disjoint by construction (`!unorm && !glc`, plus a specific `dmask==0x7` and `dim==1`). So no
previously-admitted shape changes and no previously-rejected shape is admitted except the one
byte-verified against the capture.

**Verified independently.** I disabled only the new branch
(`} else if (false && in.opcode == 0x01u && ...)`) on the PR head: `test_rdna2_decode` exits 1;
restored, it exits 0. Full suite 315/315 in-container on this branch — which also settles the count,
since a sibling PR in this batch reported 314 for the same tree.

**Confidence labels are honest and I agree with them:** HIGH on the address/byte layout (llvm-mc
round-trip is real evidence), MED on admitting `UNORM=0/GLC=0` independently of the mip proof, since
no other title evidences this exact opcode/NSA/dim shape with either bit set. MED is the right label
for a single-title encoding generalised to a rule, and it is marked as such rather than rounded up.

**Recording the falsification in two `## Ruled out` sections** (`RECOMPILER_REMAINING.md` and
`STRAY_STATUS.md`) plus cross-posting the mip-tail/integer-format data point to #2818 is the part
that will save the most time later — the next person to read #3134 would otherwise re-run the whole
capture-and-disassemble cycle to reach the same dead end.

Correct call not to attempt the guest mip-chain upload here: it is the shared texture-upload path
every title goes through, and #2818's own lane parked it after comparable due diligence.

No `Claude-Session:` trailer, no absolute host paths, four files, +105/-3.

mattias800 (commented):
**APPROVED** (re-approval on the current head after a sync)

My earlier approval was detached when I pushed a merge of `origin/master` into this branch, so
recording the verdict again against head `19316a08`.

**Why the sync happened.** CI reported `Windows MinGW` failing on
`apr_registry_resolution` — `[FAIL] generated executable AMPR WriteAddress import stub`. That test
has nothing to do with this diff, which touches `rdna2_decode.cpp` and its test only. The branch was
**3 commits behind master**, so rather than wave the red check through as "unrelated", I merged
current master in and let CI re-run. It is now **9 passed / 2 skipped / 0 failures**, which settles
it: the failure came from the stale base, not from this change.

**Content is unchanged by the sync.** Against `origin/master`, the branch is still exactly the four
files reviewed: `rdna2_decode.cpp` (+46/-3), its test (+34), and two `## Ruled out` doc entries.
Nothing in the merge altered the change under review, so the substance of my previous review stands
unmodified — the per-branch `unorm`/`glc` refactor is behaviour-preserving, the new branch is
disjoint (`!unorm && !glc`, `dmask==0x7`, `dim==1`), and the mutation arm holds (disabling only the
new branch fails `test_rdna2_decode`; restoring it passes).

Full suite 315/315 in-container. `Refs #3134` rather than `Fixes` remains correct: this widens the
decode-level shape but the stage still rejects the captured packet, and the real fix (guest mip-chain
upload) belongs to #2818.

PR #3168, opened 2026-09-01
tools/perf: audio_delivery_report distinguishes quantized wake from a slow producer

### LINKED ISSUE #3080: audio_delivery_report asserts 'the mixer's wake is quantized — fix the wait primitive' where the timedwait census measures x1.02
`prosper/tools/perf/audio_delivery_report.py` prints, as one of its two verdicts:

> the inter-arrival cadence clusters beyond two grain periods — the mixer's wake is quantized (a timed wait resolving on a coarse tick), not a buffering defect. **Fix the wait primitive's resolution.**

On *Blasphemous 2* (`PPSA13579`, Windows) that verdict is **wrong**, and it names a remedy that has already been applied and measured.

## The contradiction

Measured on the same title, with `PROSPER_TIMEDWAIT_CENSUS=1` (#3022):

```
[timedwait 5s] usleep  calls=488  requested= 20.143 ms  actual= 20.503 ms  x1.02
```

`usleep` is the *only* timed-wait primitive this title uses — `SemTimedwait`, `CondTimedwait` and `MutexTimedlock` never appear in the census — and it resolves at **x1.02** after #3022. A wait resolving on a coarse tick does not measure 1.02. The wait primitive's resolution is already fixed; the tool is telling a reader to go fix it.

Meanwhile its *first* verdict on the same run is correct and well-supported: `effective delivery: 21462 frames/s vs device 48000 -> drift -55.29%`, i.e. "a clock deficit, not a buffering defect", which independently confirms #3072 (44.7% of real-time audio, against 44.9% derived from grains by the other reader).

So the tool is right about the deficit and wrong about the cause, in the same output block, at the same confidence.

## Why this is worth an issue rather than a caveat

**A tool that prints a mechanism verdict reads as checked.** The charter's citation-drift section is about exactly this: a claim arriving already labelled as verified gets re-derived by nobody. This one arrives as an imperative — "Fix the wait primitive's resolution" — which is the strongest form.

The inference itself is reasonable. Clustering beyond two grain periods *is* what a quantized wake looks like. But it is not the only thing it looks like: a guest that simply produces audio at half real time also arrives late in clusters, because the arrivals are spaced by production rather than by a wait. The cadence shape cannot distinguish "the wake is late" from "the work is slow", and the tool asserts the first.

Concretely on this run: `gaps beyond 2 grains: 4485 (52.4%)` triggers the verdict, and the median gap is 11.12 ms against a 5.33 ms grain — 2.09x. That is exactly what a 45%-of-real-time producer looks like with no quantization at all.

## Suggested fix

Three options, cheapest first, and the choice is a judgement call:

1. **Weaken the verdict to a hypothesis and name the discriminator.** "…consistent with either a quantized wake or a producer running below real time; `PROSPER_TIMEDWAIT_CENSUS=1` distinguishes them." One sentence, and it points at the measurement that settles it.
2. **Suppress it when the first verdict already fired.** A clock deficit of −55% explains the cadence on its own, so emitting both is asserting two independent causes for one observation. Cheap, and this run is the case for it.
3. **Read the census.** The census counters exist and could be summarised into the same log, so the tool answers the question instead of inferring past it. Most work, and the only option that makes the verdict earned.

`CONFIDENCE: HIGH` that the verdict is wrong on this title (two independent measurements). `MED` on whether the heuristic is wrong in general — it may be right on a title whose waits really are quantized, which is precisely why it needs a discriminator rather than deletion.

## Cross-reference

The other reader's grain-relative thresholds were themselves wrong until #3046 (computed in FRAMES, so "below one grain" was really "empty"), and one of its figures was misquoted onto #3016 on that basis. Two of the three defects found in this tool family this week were **labels that overstated what the numbers supported** rather than arithmetic errors, which is the pattern to watch here.

Refs #3072. Refs #3033. Refs #3046. Refs #3022.

mattias800:
## Summary

`audio_delivery_report.py`'s burst-clustering verdict asserted, unconditionally, that clustered
inter-arrival gaps mean "the mixer's wake is quantized ... fix the wait primitive's resolution."
That shape (clustering beyond two grain periods) is *consistent with* a quantized wait, but it is
also exactly what a producer simply delivering audio below real time looks like: its arrivals
cluster too, spaced by production rather than by a wait. The tool could not distinguish the two
from cadence data alone, and picked one anyway.

On *Blasphemous 2* (`PPSA13579`, Windows) that assertion was wrong and pointed at a remedy already
applied and measured: `PROSPER_TIMEDWAIT_CENSUS=1` showed the title's only timed wait (`usleep`)
resolving at `x1.02` -- not quantized -- while the report's own delivery-rate verdict on the same
run correctly measured a `-55.29%` clock deficit that fully explains the clustering.

## What the census data actually supports

Read carefully, the census cited in the issue supports exactly one thing: **on this run, the wait
primitive is not coarse.** It does not, by itself, prove the deficit is the *whole* explanation of
the clustering (that inference is independently supported by the report's own correct first
verdict), and the report's cadence data alone can never distinguish "wake is late" from "work is
slow" -- both produce the same clustered shape. So the honest fix is not to swap in "it's a clock
deficit" as a new unconditional assertion; it is to make the report use the actual discriminating
measurement (the census) when it is available, and say plainly that it cannot tell when the
measurement isn't there.

This was **not** a text-only defect: the report genuinely had no code path that could use census
data, even though a combined capture (`PROSPER_AUDIO_DEBUG=1` + `PROSPER_TIMEDWAIT_CENSUS=1` into
one log) is exactly how the issue's own evidence was gathered. So the fix is a small code change,
not a wording change alone.

## Change

- `audio_delivery_report.py` now also parses `[timedwait Ns]` census lines
  (`src/hle/kernel/timedwait_census.hpp`'s `report_timedwait_census()`) from the same input stream,
  and aggregates each wait primitive's calls-weighted mean overshoot ratio.
- The burst-clustering verdict now branches on whether a census is present in the log:
  - **No census**: the verdict is hedged into a named hypothesis ("consistent with EITHER a
    quantized mixer wake OR a producer delivering audio below real time... this cadence alone
    cannot tell them apart") and points at `PROSPER_TIMEDWAIT_CENSUS=1` as the discriminator,
    instead of asserting a mechanism the tool has no evidence for.
  - **Census present, ratio < 1.3** (near-exact): "RULES OUT a quantized wake" -- states the
    measured primitive, its ratio, and its call count -- and does **not** print the "fix the wait
    primitive" imperative.
  - **Census present, ratio >= 1.3** (genuinely coarse): "CONFIRMS a coarse wait" with the same
    numbers, and the imperative is now earned rather than asserted.
- Docstring updated to describe the limitation honestly rather than claiming the burst-clustering
  metric "names a quantized mixer wake" outright.

## Tests

Three new cases in `test_audio_delivery_report.py` reconstruct the actual Blasphemous 2 numbers
from the issue (~-55% delivery-rate drift, gap of 11.85 ms against a 5.33 ms grain -- 100% of
arrivals beyond two grains) and check all three branches:
- no census present -> hedged text, no "fix the wait primitive" imperative;
- census present at `x1.02` (the issue's own figure) -> "RULES OUT", no imperative;
- census present at `x2.90` (`#3013`'s figure, a genuinely coarse case) -> "CONFIRMS", imperative
  present.

Verified these fail against the pre-fix code (`git stash` the tool changes, keep the test changes,
re-run) -- all 8 new assertions failed on the old code and pass on the fix. Existing cases 0-5
(queue-threshold and cadence fixtures) are unaffected -- notably, none of the *existing* cases
previously reached the burst-clustering verdict at all (the closest, case 3, sits just under the
2-grain threshold), so this code path had zero prior coverage.

## Affected / unaffected

Python tool only (`tools/perf/`); no engine/runtime behavior changes. Every other verdict in the
report (drift, queue thresholds, cadence) is unchanged.

## Test plan

- [x] `python3 tools/perf/test_audio_delivery_report.py` -- all cases pass
- [x] Confirmed new cases 6-8 fail on the pre-fix tool, pass on the fix
- [x] Full build + `ctest --no-tests=error` in the container: **314/314 passed (100%)**, exit 0
      (verified as a real run via `ctest -N`, not the "empty test dir" false-green case). Baseline
      expectation was 315; the 1-test difference is consistent with normal churn on `origin/master`
      between when that baseline was set and this branch's base commit, not caused by this change
      (which touches no CMake test registration).

Fixes #3080

### DIFF
--- a/prosper/tools/perf/audio_delivery_report.py
+++ b/prosper/tools/perf/audio_delivery_report.py
@@ -18,14 +18,23 @@ that log into the numbers that name the failure mode, offline, without listening
     between the guest's budgeted audio clock and the device crystal; no cushion survives it,
     and the fix is drift compensation, not deeper buffering;
   * burst clustering -- the fraction of arrivals that carry more than one grain of audio
-    after a gap of more than two grain periods, which names a quantized mixer wake.
+    after a gap of more than two grain periods. This shape is CONSISTENT WITH a quantized
+    mixer wake, but the cadence alone cannot prove it: a producer simply delivering audio
+    below real time also arrives in clusters, spaced by production rather than by a wait
+    (#3080 -- the verdict named the wrong cause on a title where a `PROSPER_TIMEDWAIT_CENSUS=1`
+    measurement, in the same log, showed the guest's only timed wait resolving at x1.02, i.e.
+    not quantized at all; the -55% delivery-rate deficit already explained the clustering). So
+    if `[timedwait Ns]` census lines are present in the same input, their ratio for the
+    dominant wait primitive SETTLES which cause applies; if they are absent, the report says so
+    and names the census as the discriminator instead of asserting a mechanism it cannot see.
 
 WHAT THIS TOOL DOES NOT SEE
 ---------------------------
 `[audio-dbg]` is emitted per accepted `output()` call from the guest. Grains the guest never
 delivers (a mixer that skipped a period) appear only as a longer gap. Content-level defects
 (a mixer that produced silence or repeated samples) are invisible here; the queue level and
-the delivery rate are the report's whole world.
+the delivery rate are most of the report's world -- the one exception is the wait-primitive
+census below, read from the SAME log when the run captured both diagnostics together.
 
 Usage:
     audio_delivery_report.py [LOG ...] [--port P] [--device-hz H] [--channels C] [--bytes-per-frame N]
@@ -39,6 +48,23 @@ from collections import defaultdict
 RECORD = re.compile(
     r"\[audio-dbg\] port=(\d+) gap=([0-9.]+)ms frames=(\d+) queued_before=(\d+)"
     r"(?: \(grain=(\d+)\))?")
+
+# `report_timedwait_census()` (src/hle/kernel/timedwait_census.hpp) prints one of these per
+# primitive every 5 s when PROSPER_TIMEDWAIT_CENSUS=1. It is a DIRECT measurement of how coarse
+# a wait primitive actually is -- requested vs actual, as a ratio -- which is exactly the
+# question the burst-clustering verdict below cannot answer from cadence alone. A primitive with
+# no requested interval (an absolute-clock deadline, e.g. cond_timedwait) prints "?" instead of a
+# ratio and is skipped: no ratio, no evidence either way.
+TIMEDWAIT_RECORD = re.compile(
+    r"\[timedwait \S+\]\s+(\S+)\s+calls=(\d+)\s+requested=\s*(?:[0-9.]+\s*ms|\?)"
+    r"\s+actual=\s*[0-9.]+\s*ms(?:\s+x([0-9.]+))?")
+
+# A ratio comfortably above 1.0 means the wait genuinely overshot what the guest asked for (the
+# "coarse tick" the burst verdict describes); a ratio near 1.0 means the wait resolved close to
+# on time and cannot be the cause of clustered arrivals. #3013 measured ~x2.9 for a primitive that
+# WAS the cause; #3080 measured x1.02 for one that was not, on a title whose clustering was fully
+# explained by a -55% delivery-rate deficit instead. The threshold sits well clear of both.
+QUANTIZED_RATIO_THRESHOLD = 1.3
 
 # The device consumes `device_hz * channels * bytes_per_frame` bytes per second. The guest's
 # delivery rate against THAT number is the clock-drift measurement; a persistent deficit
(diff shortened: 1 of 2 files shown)


mattias800 (commented):
**APPROVED**

First, a correction to something I told you and to what I said while reviewing: **your 314 is right
and my 315 was wrong.**

I had been quoting 315 as the tree's count and briefed sibling agents with it. My build directory is
configured with `-DGAME_DUMP=<DUMP_ROOT>/PPSA15319-app0`, which registers one extra case. A fresh
configure with no dump gives **314**, exactly as you reported:

```
my dump-configured build dir : Total Tests: 315
fresh configure, no dump     : Total Tests: 314
```

Your *number* was correct. Your *explanation* was not — you attributed it to `origin/master` churn
between the 315 measurement and your base, and I checked: your branch was 0 commits behind master.
The real cause is dump configuration, not churn. Worth separating, because "the count moved because
master moved" and "the count depends on how the tree was configured" lead to different reflexes next
time. And the charter's own note that a non-PPSA24651 dump makes three cases report `(Skipped)` is
about skips, not about the registered total, so it would not have explained it either.

That aside, the substance is good and the judgment is the part I want to name:

**You fixed the right defect.** The issue could have been read as "soften the wording", and instead
you found the tool had no code path capable of reading `[timedwait Ns]` census lines at all — so the
report could not have distinguished the cases even in principle, whatever it said. Diagnosing that
as a code gap rather than a phrasing problem is the correct call.

**The three-way branch is the right shape**, and specifically the middle case is the valuable one:
census present and ratio below threshold now states that the data *rules out* a quantized wake, with
the measured numbers, and drops the imperative. A report that only ever says "fix the wait
primitive" cannot be wrong in a way anyone notices; one that can say "this rules it out" can.

**Honest about the threshold.** `QUANTIZED_RATIO_THRESHOLD = 1.3` sits between the two known data
points (x1.02 not-quantized here, x2.90 quantized in #3013) with no title evidence for where the
boundary actually is, and you documented that in-code and flagged it for review rather than
presenting it as derived. That is the right treatment for a judgment constant — it is discoverable
and revisable when a third data point shows up.

**Also worth noting:** you found the pre-existing suite had zero coverage of this verdict block, so
the three new cases close a real gap rather than decorating one.

Verified: no `Claude-Session:` trailer, no absolute host paths, two files, +134/-9.

PR #3169, opened 2026-09-01
docs: fix both sides of check_merge_result's checker-invocation locale decode

### LINKED ISSUE #3079: check_merge_result's checker invocation decodes with the locale codec — and the fix must be both-sides, or an accidental agreement becomes a real mismatch
Raised in review of #3071 and deliberately not fixed there, because a one-sided fix makes it worse.

## The site

`prosper/tools/docs/check_merge_result.py` invokes the numbered-table checker as a subprocess and reads its output:

```python
proc = subprocess.run(
    [sys.executable, str(CHECKER), "--ordered", "--table-header", args.table_header, ...],
    capture_output=True, text=True,      # <-- no encoding=
)
...
out = proc.stdout + proc.stderr
```

`text=True` with no `encoding=` decodes with the **locale** codec, which on a Windows host is cp1252. #3071 fixed the same class in this file's `git()` helper and in `trap_number.py`, and left this one alone on purpose.

## Why a one-sided fix is worse than leaving it

The child process encodes **its** stdout with the locale codec too, unless `PYTHONIOENCODING` or `PYTHONUTF8` is set. Parent and child currently agree by accident. Adding `encoding="utf-8"` to the parent alone converts an accidental agreement into a real mismatch: the child would emit cp1252 bytes and the parent would try to read them as UTF-8.

**So the remedy is both sides**: `encoding="utf-8"` on the parent *and* `env={**os.environ, "PYTHONIOENCODING": "utf-8"}` (or `PYTHONUTF8=1`) on the child. Either half alone is a regression.

## Why it is latent rather than broken

The checker's own messages carry `path:line_no` and counts — ASCII. It does not echo table content back. So today there is nothing non-ASCII to decode.

It is not *safely* latent, though, and the numbers say why: the document holds **six** characters cp1252 cannot encode — U+2192 (→) ×41, U+2248 (≈) ×11, U+03C0 (π) ×2, U+2082 (₂), U+2705 (✅), U+2212 (−). Any future checker message that quotes a row, a header, or a cell — an obvious thing to add to a diagnostic — makes this live immediately, on Windows only.

## The failure shape, which is the part worth recording

There is a second guard between the decode and the crash that makes it quieter than #3071's:

```python
if stream:                 # around :149
    out += stream
```

A `None` stream is falsy, so it is **silently dropped** rather than raising. The tool would then report *no table problem found* from an empty read — a clean bill of health derived from nothing. That is strictly worse than `trap_number.py`'s original failure, which at least crashed: a green verdict from a discarded stream reads as validation.

`CONFIDENCE: HIGH` on the mechanism and the character census (both measured on this host); `MED` on reachability, since it needs a checker change that has not happened.

## Suggested fix

```python
proc = subprocess.run(
    [sys.executable, str(CHECKER), ...],
    capture_output=True, text=True, encoding="utf-8",
    env={**os.environ, "PYTHONIOENCODING": "utf-8"},
)
if proc.stdout is None or proc.stderr is None:
    print("error: the checker's output could not be decoded; refusing to report a verdict",
          file=sys.stderr)
    return 2
```

And a test arm with a non-ASCII **checker message**, which needs a checker that emits one — so this is a two-file change with its own review, which is why it is an issue rather than a line in #3071.

Refs #3071.

mattias800:
## Summary

`check_merge_result.py` invokes `check_numbered_table.py` as a subprocess and captures its
output with `capture_output=True, text=True` and no `encoding=`. That decodes with the **locale**
codec (cp1252 on an unconfigured Windows host). #3071 fixed the same class of bug in this file's
`git()` helper and in `trap_number.py`, and deliberately left this call alone, because a one-sided
fix here is worse than no fix at all — see the "Why a one-sided fix is worse than leaving it"
section of #3079.

## The failure scenario

Today, with neither side forced, **both** this process's decode of the checker's captured output
and the checker's own `print()` encode fall back to whatever the ambient locale is. They agree —
by accident. Adding `encoding="utf-8"` to the parent call *alone* converts that accidental
agreement into a real mismatch: the checker child would still emit bytes in the ambient locale,
and the parent would then insist on decoding them as UTF-8.

There's a second, quieter failure shape behind that: a decode failure inside `subprocess`'s
capture does not always raise. On a platform whose pipe reader runs in a background thread
(Windows — this file's own `git()` comment already measured this), the decode failure happens in
the reader thread and never propagates; the affected stream comes back `None` with `returncode 0`.
The pre-existing `if stream: out += stream` guard treats `None` exactly like "the checker printed
nothing", so a lost read would report a **clean merge** from a read that never happened — a false
"all clear" is worse than a crash, because nothing about it looks wrong.

## The fix — both sides, on the same call

```python
proc = subprocess.run(
    [sys.executable, str(CHECKER), "--ordered", "--table-header", args.table_header,
     "--baseline", str(base_copy), str(merged_copy)],
    capture_output=True, text=True, encoding="utf-8",
    env={**os.environ, "PYTHONIOENCODING": "utf-8"},
)
if proc.stdout is None or proc.stderr is None:
    print("error: the checker's output could not be decoded as UTF-8; refusing to report a "
          "verdict from a partial read.", file=sys.stderr)
    return 2
```

Concretely, "both sides" is:

- **Parent side** — `encoding="utf-8"`: this process now decodes the checker's captured
  stdout/stderr as UTF-8, unconditionally, regardless of what the ambient locale is.
- **Child side** — `env={**os.environ, "PYTHONIOENCODING": "utf-8"}`: the checker subprocess is
  told to encode its own stdout/stderr as UTF-8 too, unconditionally, regardless of the ambient
  locale it would otherwise fall back to.

Neither half alone is a fix. The parent half alone breaks the current accidental agreement (see
above). The child half alone would make the checker emit correct UTF-8 bytes, but the parent would
still be decoding with whatever *its own* ambient locale resolves to, which can independently
disagree with UTF-8.

- **The `None`-stream guard** is the defence for the "swallowed in a background thread" shape.
  With both halves of the fix above in place this should be unreachable in ordinary operation
  (the checker can only ever emit valid UTF-8 now), but it is cheap, it is explicitly what the
  issue's investigation called out, and it matches the existing defensive pattern already used by
  this file's own `git()` helper for the analogous case.

## Reproducing it by hand (before touching any code)

Built a deterministic, portable reproduction — not dependent on which locales happen to be
installed on the box running this — using a stand-in for `check_numbered_table.py` that emits
either UTF-8 or cp1252 bytes for a message containing `café row → problem`, depending on whether
`PYTHONIOENCODING=utf-8` is present in **its own** environment. Ran the real
`check_merge_result.py` against it in a real throwaway git repo (`git merge-tree`, real merge),
by hand, in three configurations:

| configuration | result |
| --- | --- |
| pre-fix (neither `encoding=` nor `env=`) | crashes: `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9 ...` |
| one-sided fix (`encoding="utf-8"` only, no `env=`) | crashes with the **same** `UnicodeDecodeError` |
| both-sides fix (this PR) | `café row → problem` comes through intact, correct rc |

The one-sided case crashing identically to the unfixed case is the concrete proof that "both
sides" isn't a stylistic preference — a partial fix here buys nothing.

## Regression tests

Added two cases to `test_check_merge_result.py` (`doc_merge_result_checker` in ctest):

1. **`checker_locale_mismatch`** — the by-hand repro above, committed. Substitutes the
   deterministic stand-in checker (same technique the file's existing `cli_missing_checker` case
   already uses to swap the sibling script) so the real subprocess plumbing is exercised for real,
   without depending on host locale availability.
2. **`checker_output_undecodable`** — the platform-specific `None`-stream shape can't be produced
   through a real subprocess on Linux (confirmed by hand: the equivalent mismatch raises
   synchronously here instead of returning `None`, see the CPython source difference between the
   POSIX and Windows `_communicate` implementations), so this monkeypatches `subprocess.run` for
   the *one* call it targets — every `git()` call still runs for real — to inject exactly the
   documented Windows outcome and asserts the tool refuses to report a verdict from it.

**Verified both fail without the fix**: `git stash push -- prosper/tools/docs/check_merge_result.py`
(reverting only the fix, keeping the new tests), ran the suite — both new cases failed, reproducing
the exact `UnicodeDecodeError` crash and the exact false `"clean merge, and the merge RESULT
passes the gate..."` report the issue describes, with `rc=0`. `git stash pop` restored the fix;
re-ran — all 11 cases pass.

```
python3 prosper/tools/docs/test_check_merge_result.py
# ... all 11 cases, including the two new ones, "ok"
# all cases passed / exit 0
```

## Verification

```
python3 prosper/tools/docs/test_check_merge_result.py        # 11/11 cases, exit 0
python3 prosper/tools/docs/test_check_numbered_table.py       # unaffected; 0 skipped, all pass
git ls-files '*.md' -z | xargs -0 python3 prosper/tools/docs/check_numbered_table.py   # rc=0, no false positives
git diff --check                                              # clean
```

This is a documentation-tooling-only change (two `.py` files under `prosper/tools/docs/`); no
`.md` file is touched, so the numbered-table gate itself is not exercised by this diff, only run
here as a sanity check that the fix introduced no regressions elsewhere.

Fixes #3079

### DIFF
--- a/prosper/tools/docs/check_merge_result.py
+++ b/prosper/tools/docs/check_merge_result.py
@@ -37,6 +37,7 @@ the same one `check_numbered_table.py`'s own WHAT THIS CANNOT SEE section draws)
 from __future__ import annotations
 
 import argparse
+import os
 import subprocess
 import sys
 import tempfile
@@ -139,11 +140,37 @@ def main() -> int:
         merged_copy.write_text(show.stdout, encoding="utf-8")
         base_copy.write_text(git("show", f"{base}:{args.file}").stdout, encoding="utf-8")
 
+        # encoding="utf-8" AND env=...PYTHONIOENCODING on the SAME call, deliberately: this is a
+        # BOTH-SIDES fix, not two independent ones. `encoding="utf-8"` fixes how THIS process
+        # decodes the checker's captured stdout/stderr; it says nothing about what bytes the
+        # checker actually wrote. Absent an override, the checker's own `print()` encodes with
+        # WHATEVER the ambient locale is (cp1252 on an unconfigured Windows host) -- and today,
+        # with neither side forced, parent and child read that same ambient locale and agree BY
+        # ACCIDENT. Adding encoding="utf-8" here alone would break that accidental agreement
+        # into a real one: the checker would still emit cp1252 bytes, and this call would then
+        # insist on decoding them as UTF-8. `PYTHONIOENCODING=utf-8` in the child's environment
+        # is what makes the checker itself emit UTF-8 regardless of locale, so the two sides
+        # agree on purpose instead of by luck. Neither half alone is a fix -- see #3079.
         proc = subprocess.run(
             [sys.executable, str(CHECKER), "--ordered", "--table-header", args.table_header,
              "--baseline", str(base_copy), str(merged_copy)],
-            capture_output=True, text=True,
+            capture_output=True, text=True, encoding="utf-8",
+            env={**os.environ, "PYTHONIOENCODING": "utf-8"},
         )
+    if proc.stdout is None or proc.stderr is None:
+        # Fail closed rather than let a discarded stream read as "nothing to report". A decode
+        # failure inside subprocess's capture does not always raise here -- on a platform whose
+        # pipe reader runs in a background thread (this file's own git() helper measured that on
+        # Windows: the decode fails in the READER THREAD, and the affected stream comes back None
+        # with returncode 0 rather than propagating), the exception never reaches this frame at
+        # all. The `if stream:` guard below treats `None` exactly like "the checker printed
+        # nothing", so without this check a lost read would report a CLEAN merge from a read that
+        # never happened -- strictly worse than a crash, because nothing about it looks wrong.
+        # encoding="utf-8" plus the child-side PYTHONIOENCODING above should make this
+        # unreachable in ordinary operation; this is the backstop for when it somehow is not.
+        print("error: the checker's output could not be decoded as UTF-8; refusing to report a "
+              "verdict from a partial read.", file=sys.stderr)
+        return 2
     # Rewrite the temporary path back to the real one: an error naming a file in /tmp that no longer
     # exists reads as a tooling failure rather than as a finding about your change.
     for stream, out in ((proc.stdout, sys.stdout), (proc.stderr, sys.stderr)):
--- a/prosper/tools/docs/check_merge_result.py
+++ b/prosper/tools/docs/check_merge_result.py
@@ -76,7 +76,37 @@ def git(*args: str, check: bool = True) -> subprocess.CompletedProcess:
     return proc
 
 
+def _make_own_output_utf8() -> None:
+    """Reconfigure THIS process's own stdout/stderr to UTF-8, so what this tool prints is safe
+    regardless of the ambient locale of whoever launched it.
+
+    A THIRD locale boundary, found in review of #3079 by a test that runs this tool as a real
+    subprocess on real Windows Python: the checker-invocation fix (encoding="utf-8" on the parent
+    decode, PYTHONIOENCODING=utf-8 forced into the checker's own environment) makes what THIS
+    process reads from the checker always well-formed text -- verified directly against a genuine
+    Windows Python build, not reasoned about. But this process then re-prints that text to ITS OWN
+    sys.stdout/sys.stderr (the loop a few lines below), and nothing made THOSE safe: on an
+    unconfigured Windows host, absent PYTHONIOENCODING/PYTHONUTF8 in the environment of whoever
+    ran THIS tool, they still fall back to the ANSI codepage (cp1252). Measured on real Windows
+    Python (embeddable 3.12.7, run under Wine -- close enough to a genuine Win32 process to trust
+    for this): printing an arrow (U+2192, unrepresentable in cp1252) to an unforced stdout raises
+    UnicodeEncodeError (stdout's error handler is 'strict'); the identical content on stderr does
+    NOT raise (stderr defaults to 'backslashreplace') but is silently rewritten to the literal
+    7-character text `\\u2192` instead of the glyph -- content corruption with no error at all.
+
+    Fixing the checker boundary alone therefore does not fix the whole pipeline: a message that
+    survives decode intact can still be mangled or crash on the way back OUT. `reconfigure` is
+    tolerant of streams that are not real TextIOWrapper objects (a test redirecting stdout/stderr
+    to io.StringIO has no such method) so it is a no-op there rather than an AttributeError.
+    """
+    for name in ("stdout", "stderr"):
+        reconfigure = getattr(getattr(sys, name), "reconfigure", None)
+        if reconfigure is not None:
+            reconfigure(encoding="utf-8")
+
+
 def main() -> int:
+    _make_own_output_utf8()
     ap = argparse.ArgumentParser(description=__doc__.split("\n")[0])
     ap.add_argument("--base", default="origin/master", help="branch to merge into (default origin/master)")
     ap.add_argument("--head", default="HEAD", help="what to merge (default HEAD)")
--- a/prosper/tools/docs/test_check_merge_result.py
+++ b/prosper/tools/docs/test_check_merge_result.py
@@ -203,7 +203,7 @@ sys.exit(1)
 '''
 
 
-def checker_locale_mismatch(name: str) -> None:
+def checker_locale_mismatch(name: str, *, extra_env: dict[str, str] | None = None) -> None:
     with tempfile.TemporaryDirectory() as d:
         repo = scenario(BASE, table("1:a", "2:b", "3:c", "4:master's row"),
                          table("1:a", "5:the lane's row", "2:b", "3:c"), d)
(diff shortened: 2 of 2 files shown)


mattias800 (commented):
**APPROVED**

The fix is right and the reasoning about why it must be both-sided is correct. One finding about the
tests that I want on the record.

**Verified the both-sides argument by mutation, and it is half-covered.** Using exit status as the
signal, on this branch:

| mutation | result |
| --- | --- |
| fix intact | `rc=0` |
| child `PYTHONIOENCODING` removed (parent-only fix) | `rc=1` — **caught** |
| parent `encoding="utf-8"` removed (child-only fix) | `rc=0` — **not caught** |

So the suite proves the parent-only fix is insufficient, which is the direction the issue emphasised
and the one the author reproduced by hand. It does **not** prove the child-only fix is insufficient.
That asymmetry is expected rather than sloppy — this host's locale is already UTF-8, so dropping
`encoding="utf-8"` changes nothing here, and detecting it would need the parent's own decode locale
forced to something else. But the PR body reads as if both directions were pinned, and only one is.
Not blocking: the fix itself is complete and the argument for the second side is sound on inspection
(a UTF-8-emitting child decoded by a cp1252 parent is exactly the mismatch this exists to prevent).
Worth a follow-up if someone wants the second direction actually guarded.

**The stand-in checker is the right shape.** Choosing its output codec from whether
`PYTHONIOENCODING` is present in its *own* environment makes the reproduction independent of which
locales happen to exist on the CI host — that is the difference between a test that measures the
defect and one that measures the machine.

**Good catch on the `None`-stream guard**, which the issue called for and which is the nastier half:
a decode failure in `subprocess`'s reader thread returns `None` with `returncode 0`, and the
pre-existing `if stream:` would have read that as "the checker printed nothing" — a false clean
merge on the one tool whose entire job is to refuse a bad merge.

**Honest about its own limit**, which I checked and agree with: that guard is exercised by
monkeypatching `subprocess.run` rather than through a real subprocess, because the Windows
thread-swallow behaviour does not reproduce on Linux (the mismatch raises synchronously here). The
author confirmed the mock still discriminates — it fails if the guard is removed — rather than
leaving a vacuous test. Flagging the platform gap instead of quietly claiming coverage is the right
call.

Verified: 11/11 `test_check_merge_result.py`, `test_check_numbered_table.py` unaffected, repo-wide
docs sweep rc=0 with no false positives, no `Claude-Session:` trailer, no absolute host paths.

mattias800 (commented):
**REJECTED** — not for the fix, which I think is right, but because the new test fails on the one
platform this issue exists for.

CI `Windows MinGW`:

```
22/242 Test #22: doc_merge_result_checker ...................***Failed
FAIL: a checker message with non-ASCII content survives a hostile child locale
1 case(s) failed.
```

That is your own new case, and it is red on Windows while green on Linux. Since the entire defect is
"an unconfigured Windows host decodes with cp1252", a fix whose regression test cannot pass there is
not yet demonstrating what it claims.

**This is not a stale base.** I already merged current master into the branch and re-ran; the failure
survived. It is reproducible, not flaky.

**Two candidate explanations, and they need different fixes** — please establish which before
changing anything:

1. **The fix genuinely does not hold on Windows.** `PYTHONIOENCODING=utf-8` and `encoding="utf-8"`
   do not compose the same way there, so the child still emits cp1252 or the parent still decodes as
   cp1252. If so, the fix needs to change — and this test is doing exactly its job by catching it.
2. **The test's stand-in checker does not behave as intended on Windows.** Your stand-in picks its
   output codec from whether `PYTHONIOENCODING` is present in its own environment. If the Windows
   runner already sets that variable, or if writing raw cp1252 bytes to a pipe behaves differently
   there, the fixture is testing the runner rather than the fix. If so, the fixture needs to change.

Please do NOT resolve this by making the assertion conditional on platform, or by skipping the case
on Windows. That would delete the only coverage of the actual bug on the actual affected platform
and leave a test that can only ever pass.

The rest of the review stands and I do not want it lost: the both-sides reasoning is correct, the
`None`-stream fail-closed guard is a good catch and the nastier half of the defect, and choosing a
stand-in whose codec is selected by its own environment (rather than by installed locales) was the
right instinct. One thing I found while reviewing that is worth folding in: by mutation, the suite
catches removing the **child** side (`rc=1`) but **not** removing the **parent** side (`rc=0`),
because this Linux host is already UTF-8. So only one of the two directions is actually pinned today.

Once the Windows job is green, re-request and I will re-review on the new head.

mattias800:
## Windows CI failure — root cause found, fixed, and independently double-verified

Diagnosed against a **real Windows Python build**, not by guessing: downloaded the official
`python-3.12.7-embed-amd64` distribution and ran it under Wine (available in the project's own
container). This gave genuine Win32 `subprocess`/locale behavior — confirmed `GetACP() == 1252`,
`locale.getpreferredencoding(False) == 'cp1252'`, UTF-8 mode off by default — close enough to the
real CI runner to trust for this.

**Neither of the two candidate causes as framed, exactly** — a third boundary:

- The checker-invocation fix itself (`encoding="utf-8"` on the parent decode,
  `PYTHONIOENCODING=utf-8` forced into the checker's environment) is **completely correct on real
  Windows Python** — verified directly: no crash, no `None`, the checker's UTF-8 output decodes
  perfectly. So it isn't "PYTHONIOENCODING and encoding='utf-8' don't compose there."
- My stand-in checker also isn't misbehaving — its own encode logic is deterministic and correct
  under Wine too.
- The actual gap: after `check_merge_result.py` correctly decodes the checker's message, it
  **re-prints that text to its own `sys.stdout`/`sys.stderr`** (the existing loop a few lines
  below), and *that* was never forced to UTF-8. On an unconfigured Windows host this falls back to
  cp1252: unrepresentable content on stdout raises `UnicodeEncodeError` (stdout's default handler
  is `'strict'`), and on stderr is *silently* rewritten to literal `→`-style escape text
  instead of the real character (stderr defaults to `'backslashreplace'`) — corruption with no
  error at all. This is the same class of bug #3079 already named, one hop further down the
  pipeline, that the original fix never touched.

**Fix**: `_make_own_output_utf8()` reconfigures this process's own stdout/stderr to UTF-8 at the
top of `main()` (tolerant of streams without `.reconfigure()`, e.g. a test's `io.StringIO`
redirect). Re-verified under Wine: the same message now round-trips perfectly, `café row →
problem` printed correctly, no crash, no corruption.

**Also folded in the mutation-testing gap from review**: `checker_locale_mismatch` previously
caught only the child-side mutation (removing the `PYTHONIOENCODING` override) — removing the
parent-side `encoding="utf-8"` was invisible on this Linux host, because its own ambient decode
already happens to be UTF-8 by coincidence. Now forces the ambient locale away from UTF-8 using
only the universally-available `"C"` locale (`PYTHONCOERCECLOCALE=0`, `PYTHONUTF8=0` — no
dependency on which extra locales happen to be installed). Verified by hand: constructed both
mutants (child-side removed, parent-side removed) and confirmed the test now fails on each
independently, and passes on the unmutated fix.

**A pleasant side effect of that last change**: forcing the ambient locale away from UTF-8 for
`check_merge_result.py`'s own process means the reprint-corruption bug now reproduces **on Linux
too**, without Wine — confirmed by running the test against the previous (pre-reprint-fix) commit
and watching it fail with the exact same corruption shape (`café` → `caf\xe9`, `→` → `→`) that
Windows CI reported. So the fix is now verified two independent ways: a real Windows Python build
under Wine, and a portably-forced non-UTF-8 locale on Linux.

Full suite (`test_check_merge_result.py`) still green: 11/11 cases. `test_check_numbered_table.py`
unaffected, 0 skipped, all pass. Docs gate clean, no false positives. `git diff --check` clean.

Pushed to the same branch (`fix/issue-3079-locale-codec`), merged current master in (fast-forward
was not possible after your `sync3169` push — merged that in too, empty diff as expected).

mattias800 (commented):
**APPROVED** — and this is the strongest piece of work in the batch.

**You disproved both hypotheses I gave you, and the real cause was neither.** I rejected this saying
it was either "the fix does not hold on Windows" or "the stand-in fixture misbehaves there". You
went and got a real `python-3.12.7-embed-amd64`, ran it under Wine in the project's own container,
confirmed `GetACP()==1252` and `locale.getpreferredencoding(False)=='cp1252'` — and showed the
checker-invocation fix was **completely correct** on genuine Windows Python. The actual defect was a
third locale boundary one hop further down: the tool correctly decodes the child's message and then
re-prints it to its own `sys.stdout`/`sys.stderr`, which nothing forced to UTF-8. On stdout that
raises `UnicodeEncodeError`; on stderr it silently rewrites the character to literal `→` escape
text — corruption with no error at all, which is the worse of the two.

Measuring against a real Windows interpreter instead of choosing between my two guesses is exactly
right, and it is why the fix is now correct rather than merely green.

**Both mutation directions are now caught — verified myself on this head:**

| mutation | result |
| --- | --- |
| baseline | `rc=0` |
| remove parent `encoding="utf-8"` (previously invisible) | `rc=1` |
| remove the new `_make_own_output_utf8()` | `rc=1` |

Forcing the ambient locale via the universally-available `"C"` locale (`PYTHONCOERCECLOCALE=0`,
`PYTHONUTF8=0`) rather than depending on installed locale packages is the right way to close that
gap — it measures the defect instead of the machine, same instinct as your original stand-in.

**The bonus finding is the useful one:** forcing that ambient locale makes the reprint corruption
reproduce on Linux too, so the bug now has two independent confirmations and CI can see it without a
Windows runner. A Windows-only regression that only CI can catch is a regression that gets
rediscovered; this one no longer is.

CI is green (9 passed, 2 skipped) including the `Windows MinGW` job that produced the original
rejection.

Verified: no `Claude-Session:` trailer, no absolute host paths, docs gate clean.

Issue #3074, opened 2026-08-27
precise_sleep's Win32SleepFallback can return before its deadline — the high-resolution path re-checks the clock and the fallback does not


mattias800:
Found by independent review of #3064, raised there as non-blocking.

## The asymmetry

`prosper/src/host/platform/precise_sleep.cpp` has two Windows paths, and only one of them enforces the contract the header advertises (*"Block until `deadline_ns`"*).

The high-resolution-timer path loops with a clock re-read, and says why:

```cpp
g_backend = SleepBackend::Win32HighResolutionTimer;
// Loop rather than return: a waitable timer may be released marginally early, and the
// contract here is "do not come back before the deadline". The re-read of the clock makes
// that true without trusting the primitive's own precision.
```

The fallback path, four lines below, does not:

```cpp
if (steady_now_ns() >= deadline_ns) return;
sleep_until_with_std(deadline_ns);          // one call, no post-condition re-check
g_backend = SleepBackend::Win32SleepFallback;
```

So `Win32SleepFallback` can return **before** the deadline, while the function's own documented contract and the sibling path's comment both say it cannot.

## Reachability

`CONFIDENCE: MED`, and the uncertain link is named rather than glossed.

`sleep_until_with_std` reaches libstdc++'s `std::this_thread::sleep_until`, which forwards to a single `sleep_for` **without** re-checking the clock (`bits/this_thread_sleep.h`). Whether that can then return early depends on mingw-w64's `__sleep_for`, specifically whether its nanosecond→millisecond conversion truncates. Neither the reviewer nor I could read mingw's `thread.cc` on this host, so **that step is unverified**. On glibc the same call is a `nanosleep` loop and the question does not arise.

The fallback is also only taken when the waitable timer is unavailable — `CREATE_WAITABLE_TIMER_HIGH_RESOLUTION` needs Windows 10 1803+ — so this is not the common path on a modern host.

## Why it is worth fixing anyway

A caller cannot distinguish the two backends without asking, and the ones that matter here assume the contract:

- **The vblank kevent pump** (`hle_kernel_time.cpp`, after #3064) samples `now`, waits, then posts. If the wait returns early, the next `now` is still below the same grid boundary, `next_grid_deadline_ns` correctly returns *that same boundary* again, and the pump posts **two kevents for one vblank**. The helper is not wrong; the composition is. (The pre-#3024 shape happened to absorb this by carrying a deadline it always advanced.)
[...]

PR #3170, opened 2026-09-01
vblank pump: recover the dropped-tick signal the grid snap hides (#3075)

### LINKED ISSUE #3075: The vblank pump's grid snap makes dropped ticks invisible — a half-rate pump now has perfect phase and no symptom
Found by independent review of #3064, raised as the item worth spending code on.

## The trade #3064 made, and its cost

Before #3024 the vblank kevent pump slept a relative interval, so a missed tick showed up as **phase drift** against the status grid. That was a defect — but it was also the *symptom*, and it is how #3024 was noticed at all.

After #3064 the pump snaps to the grid: `next_grid_deadline_ns(origin, now, period)` returns the next boundary strictly after `now`, and phase stays exact **however many boundaries were skipped**. Every kevent lands on a real grid boundary, forever, whatever the rate.

So a half-rate pump now looks perfect from outside. There is no drift, no phase error, and nothing in the log.

## Why that is not hypothetical

In #3074's regime — `precise_sleep`'s `Win32SleepFallback`, which does not re-check the clock — a wait that overshoots past the next boundary makes the pump skip it. Skip every other one and the pump runs at ~32 Hz while:

- every event is on a correct grid boundary,
- the phase relationship the #3064 arms assert still holds,
- and the loop's own comment says 60 Hz.

**That is the exact shape of #3024, one layer down**, and #3024 took a dedicated investigation to spot when the symptom was still visible.

## The fix is cheap and local

The loop already computes everything needed. The boundary index is `(deadline - origin) / period`; comparing it against the previous iteration's gives the number of boundaries consumed, and anything above 1 is a dropped tick:

```cpp
const uint64_t idx = (next - origin_ns) / period_ns;
if (last_idx && idx - last_idx > 1) dropped += idx - last_idx - 1;
last_idx = idx;
```

Report it under an existing gate — `PROSPER_EVLOG` is the natural one — as a periodic count, not per occurrence, since a stalled machine would otherwise flood. The useful shape is what the audio queue trace already does: a per-interval summary with a minimum/worst figure, because a mean hides exactly this.

Note the diagnostic must not itself be the only evidence: a count that is only visible under an opt-in gate is invisible on a default run. Whether the drop count deserves an unconditional one-line warning above some threshold (say, more than a few percent of boundaries over a multi-second window) is the design question — the project's preference is fail-visible, and a pump silently running at half rate is the kind of thing that should be loud.

## Confidence

`HIGH` that the silence is real: it follows directly from the helper's contract, which is that the answer depends only on `(origin, now, period)` and never on how many boundaries were missed. That is the property #3064 wanted, and this is its cost.

`MED` on reachability today, which depends on #3074 firing — the fallback path is only taken when `CREATE_WAITABLE_TIMER_HIGH_RESOLUTION` is unavailable (pre-Windows 10 1803), and the mingw truncation link in #3074 is itself unverified.

Filed anyway, because the point is that we would not find out. A diagnostic for a defect class we cannot currently observe is worth more than one for a defect we can.

Refs #3064. Refs #3024. Refs #3074.

mattias800:
Fixes #3075.

## Problem

`next_grid_deadline_ns()` (added in #3024) is deliberately phase-exact under a stall of any length: its answer depends only on `(origin, now, period)`, never on how many boundaries were skipped. That is correct for **scheduling** — the pump must resume on the next real boundary rather than replay a backlog — but the same property means its return value alone cannot tell a healthy 60 Hz pump from one silently running at a fraction of its rate: both land on real grid boundaries every single call.

**Reproduced by hand before changing anything** (see the standalone check below, same construction `vblank_pump()` uses): a simulated half-rate pump — each wait overshoots by 1.5 periods, the `#3074` `Win32SleepFallback` regime the issue names — landed on-grid **10/10** times while actually advancing 2 periods per call every time:

```
iter  0: prev=1000000000 next=1033366700  on-grid=YES  advanced_periods=2
iter  1: prev=1033366700 next=1066733400  on-grid=YES  advanced_periods=2
...
iter  9: prev=1300300300 next=1333667000  on-grid=YES  advanced_periods=2
```

Nothing anywhere computed "how many boundaries did we just skip" — a half-rate pump had perfect phase and no symptom, exactly as the issue describes.

## Were scheduling-snap and reporting-snap the same code path?

**Yes, in the sense that matters: there was no separate reporting path at all.** `next_grid_deadline_ns()`'s return value was the *only* place a drop could have been observed, and nothing read it that way — it was fed straight into `sleep_until_steady_ns()` and otherwise discarded. So the fix isn't "widen a tolerance somewhere" (there was no tolerance to widen); it's adding the second, independent computation the issue's own suggested code sketches. `host::grid_boundaries_missed(prev_deadline_ns, next_deadline_ns, period_ns)` (`precise_sleep.hpp`) reads two of the scheduler's own successive return values and recovers exactly the drop count `next_grid_deadline_ns()` is required to discard for its phase-exactness guarantee — **without changing what is scheduled or how**: the deadline fed to `sleep_until_steady_ns()` in `vblank_pump()` is untouched, so `#3024`'s fix is intact.

## The fix

- `prosper/src/host/platform/precise_sleep.hpp`: new `constexpr uint64_t grid_boundaries_missed(prev_deadline_ns, next_deadline_ns, period_ns)`.
- `prosper/src/hle/kernel/hle_kernel_time.cpp` (`vblank_pump()`): tracks this every iteration and reports on a 2s window:
  - an **unconditional** one-line warning (never gated behind `PROSPER_EVLOG`) when drops exceed 5% of boundaries in the window — the issue explicitly raised "whether the drop count deserves an unconditional one-line warning... is the design question," given the project's fail-visible preference and the fact that "a count that is only visible under an opt-in gate is invisible on a default run." I answered that question yes, with a 2s window / 5% threshold, rather than leaving it open.
  - a detailed per-window line under the existing `PROSPER_EVLOG` gate (ticks, dropped, worst single-wait gap, lifetime total) for a below-threshold trickle the warning deliberately doesn't raise.
  - Windowed, not per-boundary: a long stall is one report (missed=N in one iteration), not a flood; a sustained degraded rate still reports every window, bounded by the pump's own (reduced) iteration rate.

## Tests (`tests/host/platform/test_videoout.cpp`)

New block right after the existing `#3024` grid-schedule arms:
1. Reproduces the pre-fix symptom **against the unmodified `next_grid_deadline_ns()`**: a 40-iteration half-rate sequence lands on-grid every time and reads back (`d == T0 + 80*P`) as indistinguishable from a healthy 80-tick pump.
2. Shows `grid_boundaries_missed()` recovers it exactly over the *same* sequence: 40/40 dropped.
3. Negative control: the existing healthy 60-tick `#3024` sequence run through the new function reports 0/0 drops — the arm that proves the discriminator doesn't fire on the case it must stay silent on.
4. An isolated 5-period stall (reusing the existing `#3024` stall arm's numbers) is reported as exactly 5, not 1 and not 0.
5. A third-rate (drop 2 of 3) sequence reports 2 dropped per step, not 1 — checking the counter reports *magnitude*, not just "some drop occurred."
6. Degenerate inputs (equal deadlines, a `next` earlier than `prev`, a zero period) report 0 without underflow or divide-by-zero.

### Is this test at risk of passing with the bug still present under some plausible scheduling?

I checked this specifically because timing tests are easy to get wrong in that direction. This test doesn't touch real time, threads, or `sleep_until_steady_ns()` at all — every arm calls the two pure `constexpr` functions with hand-picked `(origin, now, period)` triples and asserts exact integer results, so there's no scheduling jitter for it to depend on either way.

The more relevant question is whether the test could pass with a *plausible wrong implementation* of `grid_boundaries_missed()` rather than a correct one. I verified this directly rather than arguing it: I temporarily stubbed `grid_boundaries_missed()` to unconditionally `return 0` — the literal pre-fix state, where no measurement exists — rebuilt, and ran `test_videoout` directly:

```
[FAIL] grid_boundaries_missed recovers exactly the 40 dropped ticks phase alone hid
[FAIL] a sustained half-rate pump drops exactly one boundary per step, never a burst
[ok]   an ordinary 120us-overshoot run at full rate reports zero drops, ever
[FAIL] an isolated five-period stall between two ticks is reported as five dropped boundaries...
[FAIL] a third-rate pump reports 2 dropped boundaries per step...
== FAIL: 4 ==
```

4 of the 5 new arms fail. The one that "passes" under the stub is the negative control (arm 3) — by construction, since it asserts *zero* drops on a healthy sequence, and a stub returning 0 unconditionally cannot fail an assertion that expects 0. That arm exists precisely to prove the *other* arms aren't vacuous (a discriminator that also fired on the healthy case would be useless noise on every default run), not to carry the regression signal by itself — arms 1, 2, 4, 5 are the ones that actually detect the bug, and they all did. Restored the real implementation, rebuilt, reran: all pass (`== PASS ==`).

I can't rule out every conceivable wrong implementation passing this test (no finite test suite can), but a same-shape mistake — reporting a boundary index without subtracting the "one period is normal" baseline, or not reading the boundary index at all — is exactly what arms 1/2/4/5 are shaped to catch, since they check exact magnitudes (40, 1, 5, 2) rather than just "nonzero."

## Build + test

In the `ps5ys` container, `TMPDIR` off tmpfs, run synchronously in the foreground (not backgrounded):

```
TMPDIR=$PWD/prosper/build-linux/tmpdir cmake --build prosper/build-linux -j6
cd prosper/build-linux && ctest --no-tests=error
```

Result: **314/314 tests passed, exit 0.**

The task briefing that started this PR said the expected count on this tree is 315. I investigated the discrepancy rather than reporting it as-is: before committing, `HEAD` (i.e. my branch's base, with only uncommitted working-tree changes) was **exactly** `origin/master`'s tip (`c8f2f474`, 0 commits divergent either way — `git rev-list --left-right --count HEAD...origin/master` → `0 0`), and `ctest -N` on that same tree also totals 314 registered tests. My change adds zero new `add_test()` registrations — only additional `CHECK()` assertions inside the already-registered `videoout_queries` test binary and a new header-only function — so 314 is what current `origin/master` itself reports, not a regression from this PR. The 315 figure appears to be stale (consistent with the charter's own repeated notes about numbers going stale here); I'm not the right source to correct it project-wide since I can't observe what changed the count on `master`.

## Uncertain / worth a second look

- The 5% / 2s threshold and window for the unconditional warning are a judgment call I made rather than a value derived from any measurement — reasonable-sounding but arbitrary. Happy to tune either if review disagrees.
- `#3074` (the `Win32SleepFallback` early-return double-post) is unfixed and unrelated to this change; it's the scenario that makes this diagnostic *reachable* per the issue, not something this PR touches.
- I did not add a live/threaded test of `vblank_pump()` itself (posting real kevents under a simulated half-rate clock) — the existing `#3024` arms in this same file already establish that testing this loop's *scheduling* is done at the pure-function level rather than by driving the live detached thread, and I followed that precedent for the *measurement* too. If review wants an end-to-end thread-level guard as well, I can add one, but I believe it would be substantially more complex for the same coverage the pure-function arms above already give.

### DIFF
--- a/prosper/src/hle/kernel/hle_kernel_time.cpp
+++ b/prosper/src/hle/kernel/hle_kernel_time.cpp
@@ -1155,8 +1155,11 @@ namespace {
         // The cost of that property, named because it is real: phase stays exact however
         // many boundaries were skipped, so a pump running at HALF rate has perfect phase
         // and no symptom. The relative sleep this replaces at least drifted visibly, which
-        // is how #3024 was noticed. A dropped-boundary counter would restore the symptom
-        // without giving back the drift -- #3075.
+        // is how #3024 was noticed. #3075 restores the symptom with host::grid_boundaries_missed()
+        // below, WITHOUT giving back the drift: that helper is deliberately a second, separate
+        // computation over next_grid_deadline_ns()'s own return values rather than a change to what
+        // the scheduler returns or how it schedules -- see its comment in precise_sleep.hpp for why
+        // folding the two together would just recreate the drift #3024 removed.
         //
         // One composition caveat, recorded because the helper alone does not cover it (#3074):
         // this loop assumes the wait cannot return EARLY. If it does, the next `now` is still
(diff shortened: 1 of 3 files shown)


mattias800 (commented):
**APPROVED**

Also: **your 314 is right and the 315 I briefed you with was wrong.** My build directory is
configured with `-DGAME_DUMP=<DUMP_ROOT>/PPSA15319-app0`, which registers one extra case; a fresh
configure with no dump gives 314. You investigated rather than passing the number along, which is
what caught my error — thank you for that. (One refinement: you attributed it to stale documentation
drift; the actual variable is dump configuration, so the count is a property of how the tree was
configured rather than of when it was measured.)

**The mutation arm holds, verified independently.** I stubbed `grid_boundaries_missed()` to
`return 0` — the literal pre-fix state — rebuilt, and `test_videoout` exits 1; restoring it exits 0.

**The part that makes this a good fix rather than a tolerance bump:** you established that
scheduling-snap and reporting-snap were the same path because *there was no reporting path at all*.
`next_grid_deadline_ns()`'s phase-exactness is correct for scheduling and structurally incapable of
revealing a drop, since its answer depends only on `(origin, now, period)`. Adding an independent
pure computation over two successive scheduler outputs, and leaving the deadline fed to
`sleep_until_steady_ns()` untouched, keeps #3024's guarantee intact while making the drop
observable. That is the separation the issue asked about, answered rather than sidestepped.

**Reproducing before touching anything** — a simulated half-rate sequence landing on-grid 10/10
times while advancing two periods per call — is the right order of work, and it is what turns "the
snap might hide drops" into a demonstrated fact.

**The negative-control arm is correctly described.** You noted that the healthy-sequence case
trivially passes under the stub and exists to prove the other four are not vacuous, rather than
counting it as evidence. That distinction is one the charter cares about a lot, and most reports
would have quietly listed it as a fifth passing test.

**Two judgment calls, both flagged rather than buried:** the 5%/2s threshold is not derived, and
there is no live-thread test of `vblank_pump()` itself. I agree with following #3024's precedent of
testing at the pure-function level — a threaded timing test here would be the kind that passes for
scheduling reasons rather than correctness ones, which is the failure mode this very issue is about.
Choosing an unconditional warning above threshold rather than a gate-only line is also right: a
diagnostic nobody arms cannot report a regression on a default run.

PR #3171, opened 2026-09-01
recompiler: keep [mimg-unresolved] visible under PROSPER_MIMG_SOFT (#3143)

### LINKED ISSUE #3143: PROSPER_MIMG_SOFT suppresses [mimg-unresolved], hiding which descriptor failed
`PROSPER_MIMG_SOFT` suppresses `[mimg-unresolved]`, the one diagnostic that names the failing descriptor.

Raised as non-blocking finding N3 in the review of #3141 and deliberately left out of that PR, since
it is a behavioural change and the PR was already approved.

## Defect

`PROSPER_MIMG_SOFT` was moved above the `if (!first_report)` dedupe so that every compile is softened
rather than only the first per site — that part is correct and is why it moved. But the soft path
`return true`s before reaching the `[mimg-unresolved]` report, so arming the switch silently removes
the only line that says **which** descriptor could not be resolved.

## Failure scenario

An investigator arms `PROSPER_MIMG_SOFT` precisely because they have unresolved MIMG descriptors and
want to know whether softening them changes the picture. The run then renders, the frame is unchanged,
and the log no longer contains the one diagnostic that would identify the descriptors in question. So
the switch is at its least informative exactly when it is being used for its intended purpose, and
"softening changed nothing" cannot be distinguished from "there was nothing here to soften" without
re-running with the switch off — which is the ambiguity the B3 fix was meant to remove in the first
place.

## Suggested fix

Emit the report before the soft-path return, or emit a distinct `[mimg-soft] site=… descriptor=…`
line on that path, so the softened case is at least as legible as the rejected one. Either keeps the
switch's own effect visible, which is the property the rest of these diagnostics are built around.

Found by independent review of #3141; that PR is otherwise approved and unaffected.

mattias800:
## Summary

`PROSPER_MIMG_SOFT` softens an unresolved MIMG descriptor by substituting a
constant instead of rejecting the compile, so an investigator can see how
much of a frame depends on that one resolution failure. But its early
return sat upstream of the `[mimg-unresolved]` report — the only diagnostic
that names *which* descriptor failed to resolve (srsrc, SRT tag, fetch-pc,
UD-alias provenance, and the full resource table). So arming the switch to
answer "does softening change this frame?" also deleted the one line that
would say what got softened, and "softening changed nothing" became
indistinguishable from "there was nothing here to soften" — the exact
ambiguity the switch exists to resolve. Raised as non-blocking finding N3 in
review of #3141 and deliberately left out of that PR since it's a
behavioural change to logging.

## Root cause

In `prosper/src/gpu/recompiler/rdna2_emit_alu.cpp`'s MIMG resolution-failure
path, the `PROSPER_MIMG_SOFT` branch `return true`s before the code reaches
the `[mimg-unresolved]` report a few lines below it. That ordering was
intentional for a different reason: the report is gated by a per-site
dedupe (`(program, pc, srsrc)`), and the soft path had to sit *above* that
dedupe so every compile gets softened, not just the first per site
(#3141's fix). But "above the dedupe" ended up meaning "before the report
entirely," which is a different thing.

## Fix

Hoist the dedupe check and the `[mimg-unresolved]` report itself above
*both* the soft and the reject decision, so:
- The report still runs exactly once per site (same dedupe key and
  semantics as before), regardless of whether the site ends up softened or
  rejected.
- The soft path still substitutes a constant and prints its own
  `[mimg-soft]` confirmation line, still unconditioned by the dedupe (every
  compile is still softened, not just the first per site — #3141's fix is
  untouched).
- The reject path is otherwise unchanged.

This is a visibility-only change. `PROSPER_MIMG_SOFT` accepts exactly the
same set of descriptors it did before; only the log output around it
changed.

## What's affected / deliberately unaffected

- Affected: stderr output when `PROSPER_MIMG_SOFT` is armed and an MIMG op
  fails to resolve — `[mimg-unresolved]` now also appears, not just
  `[mimg-soft]`.
- Unaffected: which descriptors resolve, what `PROSPER_MIMG_SOFT` writes in
  place of an unresolved sample, the reject path's behavior/output, and the
  per-site dedupe semantics themselves.

## Test

Added a regression case to `prosper/tests/gpu/test_recompile_coverage.cpp`
(a pure, no-Vulkan ctest target). It builds a minimal compute program with
one MIMG `image_load` whose SRSRC (`s[0:7]`) cannot resolve against a
deliberately empty `ShaderResourceTable` (no fetch-pc key, no SRT tag, and
the range is never written so the `sgpr_base` fallback also misses), then:

1. Reject path (`PROSPER_MIMG_SOFT` unset): asserts the compile is rejected
   and `[mimg-unresolved]` is present (sanity check that the reorder didn't
   also break the pre-existing reject-path report).
2. Soft path (`PROSPER_MIMG_SOFT=1`): asserts the compile still succeeds,
   `[mimg-soft]` still fires (softening behaviour unchanged), **and**
   `[mimg-unresolved]` now also fires — this last assertion is the one that
   catches the bug.

Both calls use distinct explicit `RecompileDiagnosticContext` program
addresses so the per-site dedupe (keyed on `(program, pc, srsrc)`) can't
mask one call's report with the other's, since the file's `recompile_valu`
calls elsewhere default to address 0 and could otherwise collide.

**Verified the test fails without the fix**: stashed only the
`rdna2_emit_alu.cpp` hunk (kept the new test in place), rebuilt just
`test_recompile_coverage`, and reran it. Every other check in the file
still passed, including the two new sanity checks (reject-path report,
soft-path substitution+confirmation) — only the new
`#3143: PROSPER_MIMG_SOFT must not suppress [mimg-unresolved]...` assertion
failed, exactly as expected since the old code returns from the soft branch
before the report is reached. Restored the fix (`git stash pop`), rebuilt,
and reran clean.

## Build / test results (exact head)

Built and tested inside the project's `ps5ys` distrobox container (a host
build silently drops the Vulkan execution tests):

```
ctest --no-tests=error   # from prosper/build-linux
```

```
100% tests passed out of 314
Total Test time (real) =  85.67 sec
```

314 is the count this shared build environment produces on a vanilla
configure of current master (no extra `-D` flags) — confirmed independent
of this change by a from-scratch reconfigure. The gap from the 324 raw
`add_test(...)` call sites in `CMakeLists.txt` is fully accounted for by
ordinary default-off gates in this configuration: `PROSPER_AUDIO_SDL3=OFF`
and `PROSPER_PAD_SDL3=OFF` (2 tests), four Windows-only tests on this Linux
build, and `PROSPER_DIAGNOSTICS=OFF` by default (2 tests) — 8 total,
matching the observed gap exactly. Not a regression from this PR.

Fixes #3143

### DIFF
--- a/prosper/src/gpu/recompiler/rdna2_emit_alu.cpp
+++ b/prosper/src/gpu/recompiler/rdna2_emit_alu.cpp
@@ -6948,41 +6948,15 @@ bool emit_alu(SpirvCompute& b, RegState& rs, const Rdna2Inst& in, bool& ok, bool
                 // Ungated, and deduped per (pc, srsrc). It fires only when an image op has already
                 // failed to resolve, so its volume is bounded by the defect it reports. Behind
                 // PROSPER_DBG it was unreachable in practice on any routed boot.
-            // PROSPER_MIMG_SOFT=1 -- diagnostic only, and deliberately WRONG output.
-            //
-            // Placed BEFORE the reporting dedupe below on purpose. That dedupe fires once per
-            // (program, pc, srsrc); downstream of it, only the FIRST compile of each site would be
-            // softened and every later one -- reachable through ordinary shader-cache eviction --
-            // would take the reject path unchanged. The recorded result would then be
-            // indistinguishable from "nothing was softened", which is exactly the reading this
-            // diagnostic exists to produce. When an image
-            // op cannot resolve its descriptor, write a constant instead of failing the stage, so a
-            // frame can be seen that would otherwise be discarded entirely. It answers one question
-            // nothing else here can: how much of the picture is riding on THIS resolution failure.
-            // On Stray it showed the answer was "none" -- the composite compiled and the frame did
-            // not change -- which is what redirected #3126 away from the recompiler.
-            //
-            // Never acceptable in a run that produces progression evidence: every unresolved sample
-            // reads a flat value.
-            if (!res && getenv("PROSPER_MIMG_SOFT")) {
-                const uint32_t soft = b.uconst(fbits(0.5f));
-                uint32_t comps = 0;
-                for (uint32_t m = 0; m < 4; ++m) if (in.mimg_dmask & (1u << m)) ++comps;
-                if (!comps) comps = 1;
-                for (uint32_t k = 0; k < comps; ++k) {
-                    const int dst = in.dst.value + static_cast<int>(k);
-                    if (dst < 0) break;
-                    const uint32_t old = vreg_old(b, rs, dst);
-                    rs.vreg[dst] = soft;
-                    predicate_write(b, rs, dst, old);
-                }
-                static std::mutex smx; static std::set<uint64_t> sseen;
-                std::lock_guard<std::mutex> lk(smx);
-                if (sseen.insert(((uint64_t)b.diagnostic.program_address << 16) | in.pc).second)
-                    fprintf(stderr, "[mimg-soft] program=0x%llx pc=%u -> constant %u comps\n",
-                            (unsigned long long)b.diagnostic.program_address, in.pc, comps);
-                return true;
-            }
+                //
+                // #3143: this report used to sit downstream of the PROSPER_MIMG_SOFT early return
+                // below, so arming that switch to see whether softening changes a frame ALSO deleted
+                // the one line that names which descriptor failed -- the switch was least informative
+                // exactly when it was armed for its intended purpose ("softening changed nothing"
+                // became indistinguishable from "there was nothing here to soften"). It now runs
+                // before either the soft or the reject path decides what to DO about the failure, so
+                // both keep the same descriptor identification; the soft path still gets its own
+                // confirmation line below ([mimg-soft]) once the constant is written.
                 static std::mutex mimg_mutex;
                 // Keyed by PROGRAM as well as (pc, srsrc): every shader has a pc 16, so a
                 // pc-only key lets the first program to reach one silence all later programs and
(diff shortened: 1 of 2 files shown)


mattias800 (commented):
**APPROVED**

**Mutation verified independently.** Neutralising the hoisted report reproduces the pre-fix
suppression exactly: `test_recompile_coverage` exits 1 with
`[FAIL] #3143: PROSPER_MIMG_SOFT must not suppress [mimg-unresolved]`, and restoring it exits 0.

**The root-cause narrative is the valuable part.** You found it was not a shared `if` gating both
behaviours, but an early `return true` in the soft path that sits *above* the report — and, crucially,
that the ordering was deliberate: #3141 put the soft path above the per-site dedupe so every compile
gets softened rather than only the first. So "above the dedupe" silently became "before the report".
Hoisting the dedupe and the report above both decisions preserves that original intent while making
the diagnostic fire once per site on either path. That is the correct shape, and it is a better fix
than moving the soft path down would have been.

**Visibility-only, as required.** `PROSPER_MIMG_SOFT` still substitutes the same constant, still
prints `[mimg-soft]`, and still bypasses the dedupe so every compile is softened. It admits exactly
the same descriptors as before — which is what I said I would check.

**On the 314/315 question you investigated: you were right and I was wrong.** My build directory is
configured with `-DGAME_DUMP=<DUMP_ROOT>/PPSA15319-app0`, which registers one extra case; a clean
no-dump configure gives 314. I briefed you with 315 and that was my error, not environment drift on
your side. Your accounting of the gap from the 324 raw `add_test()` sites down to 314 via the
default-off gates (`PROSPER_AUDIO_SDL3`, `PROSPER_PAD_SDL3`, the Windows-only cases,
`PROSPER_DIAGNOSTICS`) is the right way to have answered it, and it is more useful than the number
itself — it explains the mechanism rather than asserting a total.

`Fixes #3143` is correct here; this is a complete fix, not partial.

Issue #702, opened 2026-07-14
Windows frontend: replace synchronous test renderer bottleneck (Messenger is 0.8 FPS)


mattias800:
## Problem

The native Windows prosper-app introduced by #700 is functionally interactive, but The Messenger presents at only **0.8 FPS** on an RTX 4090.

## Reproduction

~~~powershell
.\prosper\scripts\run-windows.ps1 .\PPSA24651-app0
~~~

Measured on 2026-07-14 with a MinGW RelWithDebInfo (-O2 -DNDEBUG) build:

- prosper-app --test-pattern --frames 300: 239-248 FPS, 300 frames in 1.78 s
- Messenger --frames 60: [app] 0.8 fps (60 frames), 73.23 s wall time
- process sampling during the title: ~101% of one CPU core, ~1.12 GiB working set
- Vulkan device: NVIDIA GeForce RTX 4090

This isolates the bottleneck from SDL/Vulkan swapchain presentation and from an unoptimized build.

## Likely ownership

The live renderer is still implemented on top of the offscreen test helper in tests/render_runner.h. For each render-target group and frame, render_draws_rgba currently recreates and destroys:

- color images, render passes, framebuffers, readback/staging buffers
- shader modules, descriptor pools/layouts/sets, texture images and storage buffers
- every graphics pipeline

It then submits, synchronously waits on a fence, maps/copies the whole RGBA frame to the CPU, and passes it to a second frontend upload/present path. Resource realization also copies/detiles guest resources per draw. This architecture was appropriate for correctness bring-up and screenshots but is not an interactive renderer.

## Proposed work

1. Add low-overhead, gated stage timing/counters to quantify realization, resource upload/detile, pipeline creation, GPU execution/wait, readback, and frontend upload.
2. Establish an interactive Messenger benchmark/checkpoint.
3. Move live rendering to persistent cached Vulkan resources and pipelines, with invalidation keyed by guest state/content.
4. Avoid synchronous full-frame GPU-to-CPU readback and CPU-to-GPU upload for the interactive frontend; publish/present a Vulkan image directly or through GPU copy/synchronization.
5. Preserve the current deterministic readback path for screenshot/replay tests.
6. Add FPS/frame-time reporting and a documented temporary render-scale option while the persistent renderer is developed.

## Acceptance

The Messenger reaches and sustains an actually interactive rate on the native Windows frontend without changing rendered correctness, while snapshot/replay tests retain deterministic pixels.

mattias800:
[...]

PR #3173, opened 2026-09-01
docs: correct the stale readback-only present-path claim (#1270 falsified it)

### LINKED ISSUE #3087: FRONTEND_APP.md still says the renderer is readback-based — #1270 falsified that in July
`prosper/docs/FRONTEND_APP.md:650-656` states, in the present tense:

> The current renderer remains a deterministic readback-based implementation.

**#1270 falsified that on 2026-07-24.** `prosper-app` blits the renderer's front-buffer image straight to the swapchain and the renderer skips the CPU readback entirely — `set_gpu_present_active` (`prosper/frontends/prosper-app/main.cpp:668`, the sole call site repo-wide) arms the skip at `prosper/frontends/shared/live/live_renderer.cpp`, whose own comment says the branch is inert in processes that never call it.

This is worse than a stale number, which is why it is worth an issue of its own: a dated figure carries its own warning label, but a present-tense architectural claim reads as current no matter how old it is. Anyone reasoning about present-path performance from this file will reach a conclusion that was correct in July and has been wrong since.

## Related stale copies, for whoever fixes this

- `prosper/docs/RENDERER_PERFORMANCE_2026_07.md` — its 12→24 FPS figure has the same cause and is now caveated in place (#3082), rather than deleted, because the *conclusion* of that pass still stands.
- `prosper/src/gpu/present/present_frame_rate.hpp` — its comment cited the July figure as a contradiction needing explanation; #3082 gives it the explanation and a pointer.

Both of those are done; this file is the one left.

## Suggested fix

Replace the present-tense sentence with what is actually true of each path — the CPU readback path still exists and is what `tools/screenshot` and any process that does not call `set_gpu_present_active` take, while the interactive app does not use it — and date the claim. `CLAUDE.md`'s harness bullet now carries the same distinction and can be pointed at rather than restated.

mattias800:
## Summary

`FRONTEND_APP.md:650-656` stated, in the present tense, that "the current renderer remains a
deterministic readback-based implementation." #1270 (landed 2026-07-24) falsified that: `prosper-app`
now defaults to GPU present for a real game boot — the renderer publishes its front-buffer image
directly and the app GPU-blits it straight to the swapchain, skipping the CPU round-trip entirely.
`PROSPER_APP_GPU_PRESENT=0` is the explicit opt-out, and the app still falls back to the private-device
CPU path if shared-device adoption fails at runtime.

## Changes

**`prosper/docs/FRONTEND_APP.md`** (the file named in #3087):
- Rewrote the paragraph at the old line 650-659: states both paths, dates the claim, drops the
  now-stale reference to issue #702 (closed 2026-07-19, five days *before* #1270 delivered exactly
  what it was cited as tracking), and points at CLAUDE.md's "before quoting an FPS number" bullet
  instead of restating the readback-vs-GPU-present distinction.
- The adjacent 20 -> 24 FPS figure in that same paragraph is itself a pre-#1270 readback-path
  measurement (2026-07-14, ten days before #1270) — now says so and points at the 2026-08-27
  windowed re-measurement (median ~156 presented fps).
- Two more restatements of the *same* stale claim, in the same file, both get a dated update note
  rather than deletion (the private-device path they describe is still real and still used by
  `--test-pattern`, failed device adoption, and `tools/screenshot`):
  - "The Vulkan-context decision: two contexts, one shared CPU frame" section's "When to revisit" note.
  - The "Risks & open questions" bullet: "the readback seam is the v1 answer."

**`prosper/frontends/prosper-app/README.md`**: its "Two Vulkan contexts by design" bullet stated the
private-device CPU-readback path as prosper-app's design with no mention that #1270 made it the
*fallback* rather than the default. Actively maintained (last touched 2026-08-24), not a dated
snapshot.

**`prosper/docs/WINDOWS_PORT_HANDOFF.md`**: "What works today" said the Windows build runs "the normal
... present/readback path." True when written (2026-07-16, before #1270), but the doc has had several
commits since (most recently 2026-08-10) without correcting it, and #1270 is platform-independent —
`prosper-app` defaults to GPU present on Windows exactly as on Linux.

## Swept and deliberately left unchanged

Grepped the whole repo for "readback" and related phrasing. Every other hit refers to a different,
**still-accurate** concept and was left alone:

- `docs/GRAPHICS.md`, `docs/GTA5_STATUS.md`, `docs/SYBERIA_STATUS.md`, `docs/ARCRUNNER_STATUS.md`,
  `docs/OREGON_TRAIL_STATUS.md`, `docs/BLUE_PRINCE_STATUS.md`, `docs/EVERGATE_PERFORMANCE_HANDOFF_2026_07.md`,
  `docs/THREAD_WAIT_PROFILING.md`, `docs/ASTERIX_BABYLON_STATUS.md`, `docs/DEAD_CELLS_*`,
  `tools/AGENTS.md`, `tools/gpu_replay/README.md`, `tools/gpu_timeline/README.md`,
  `tools/hostprof/README.md`, `tools/vkprobe/*`, `tools/index_fetch_probe/AGENTS.md`,
  `docs/GPU_EXECUTOR_DESIGN.md`, `docs/VERIFICATION.md`, `prosper/README.md`: these all describe the
  renderer's own **diagnostic-forced** CPU readback of intermediate/persistent render *targets*
  (`PROSPER_RTTLOG`, `PROSPER_DUMP_RTGROUPS`, `PROSPER_GPU_CAPTURE`, `live_gpu_targets` disable list,
  etc.), or test/verification-harness readback of a rendered frame for pixel assertions — an unrelated,
  correct concept from #1270's app-present-path claim.
- `docs/GAME_COMPAT_ORCHESTRATION.md` traps #127, #234, #237: these correctly state that
  `tools/screenshot` never calls `set_gpu_present_active` and so forces CPU readback, while
  `prosper-app` is the shipped GPU-present path — this is the **true** distinction and must not be
  "corrected."
- `prosper/docs/RENDERER_PERFORMANCE_2026_07.md` and `prosper/src/gpu/present/present_frame_rate.hpp`:
  already caveated by #3082 (referenced in #3087's body as done).

## Verification

Docs-only diff (`git diff --name-only origin/master...HEAD` confirms no non-`.md` files touched).

```
```
Ran clean: every listed file reports `unbroken, every row matching its header`, including
`CLAUDE.md` (33 rows) and `GAME_COMPAT_ORCHESTRATION.md` (301 rows). Exit code 0.

`git diff --check`: clean (no whitespace errors).

Fixes #3087

### DIFF
--- a/prosper/docs/FRONTEND_APP.md
+++ b/prosper/docs/FRONTEND_APP.md
@@ -229,6 +229,14 @@ pipelines and image/buffer objects per batch. At that point, unify on a single f
 or share the render target via `VK_KHR_external_memory` (zero-copy), while retaining `present_readback`
 for headless tests and screenshot tooling.
 
+**Update (#1270, landed 2026-07-24):** the "unify on a single device" half of this landed for real game
+boots, ahead of the pipeline/resource-caching precondition this note originally set — `prosper-app`
+adopts the *renderer's* device (rather than a frontend-provided one) and GPU-blits its front-buffer image
+straight to the swapchain by default, with `PROSPER_APP_GPU_PRESENT=0` as the explicit opt-out. The
+two-context design above is still real and still runs: it is what `--test-pattern`, any failed device
+adoption, and `tools/screenshot` all use today. `present_readback` was retained exactly as this note
+anticipated, for headless tests and screenshot tooling.
+
 ## Present loop (sketch)
 
 ```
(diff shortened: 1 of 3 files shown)


mattias800 (commented):
**APPROVED**

Docs-only (3 `.md` files, confirmed via `git diff --name-only origin/master...HEAD` — nothing else),
docs gate `rc=0` across all tracked markdown. Per the charter that waives the CI wait, but not
review — and a docs PR that rewrites factual claims is exactly the kind that needs one, so I checked
the new text against the source rather than reading it for plausibility.

**The replacement claims are true.** `set_gpu_present_active` has exactly one call site,
`frontends/prosper-app/main.cpp:669`, and `PROSPER_APP_GPU_PRESENT` is a real env var consulted at
`main.cpp:1535` — so "GPU present by default, `PROSPER_APP_GPU_PRESENT=0` as the explicit opt-out"
is accurate rather than merely newer.

**And the harder half is right too:** `tools/screenshot` does not call `set_gpu_present_active`, so
every statement describing *that harness* as readback-based remains true and had to survive. You
left those alone deliberately, including `GAME_COMPAT_ORCHESTRATION.md`'s traps #127/#234/#237, and
listed them. Getting this wrong in the "helpful" direction — sweeping every occurrence of the word
"readback" — would have replaced a stale claim with a false one, which is worse.

**The sweep found two instances beyond the issue's target**, both genuinely stale rather than
incidental: `prosper-app/README.md` (states the CPU-readback path as the design, last edited
2026-08-24, well after #1270) and `WINDOWS_PORT_HANDOFF.md` (true when written 2026-07-16, edited
since without correction, and #1270 is platform-independent). That is the propagation the charter
warns about, caught.

**Dating the correction rather than silently overwriting** — noting what the paragraph used to say
and when it changed — is the right treatment. It also means the adjacent 20→24 FPS figure got
flagged as a pre-#1270 readback-path measurement instead of being left to mislead as a current one.

Verified: no `Claude-Session:` trailer, no absolute host paths, `diff --check` clean.

Issue #3174, opened 2026-09-01
git stash races across dedicated per-agent worktrees (refs/stash is repo-wide, not worktree-local)


mattias800:
`git stash` uses a single, repo-wide `refs/stash` ref that is SHARED across every `git worktree`-created
directory, not just the main checkout. Two agents working in their own dedicated worktrees
(`.claude/worktrees/<slug>`, exactly as CLAUDE.md's "Work in your OWN git worktree" section
recommends) can still collide on `git stash push`/`pop` if their pushes interleave, because both are
writing to the same `refs/stash` in the shared `.git` directory.

**What happened (2026-09-01, two concurrent agents, branches `fix/issue-3051-spirv-flat` and
`fix/issue-3068-sem-init-return`, both based on the same master commit `e9b972da`):**

1. Agent A (issue #3051) ran `git stash push -- <one file>` to temporarily remove one change, rebuild,
   and confirm a regression test fails without it.
2. Around the same moment, Agent B (issue #3068) pushed its own stash in its own worktree.
3. Agent A's `git stash pop` popped Agent B's entry instead of its own (LIFO top-of-stack raced),
   applying Agent B's unrelated `hle_kernel.cpp` sem_init change into Agent A's working tree, and
   **dropping Agent B's stash entry from the stack** in the process.
4. Symmetrically, Agent B's own `git stash pop`/similar operation appears to have picked up Agent A's
   entry.

**Both agents independently noticed the foreign content and recovered correctly**, which is the only
reason nothing was lost: Agent B recognised the popped `rdna2_to_spirv_internal.hpp` diff wasn't
theirs and re-stashed it verbatim with a message crediting the collision, and Agent A did the same for
Agent B's `hle_kernel.cpp` diff (recovering the commit object that `git stash drop` had detached via
`git cat-file -t`, then `git stash store`). Content was verified byte-identical before/after recovery.
**This depended on both agents happening to notice** — a less careful pop-and-continue would have
silently mixed two issues' changes into one branch, or silently discarded the other agent's WIP with
no trace (a dropped stash commit is unreachable and eligible for GC).

**Mitigation for CLAUDE.md's worktree guidance:** `git stash` should be treated as MAIN-CHECKOUT-ONLY
even when using dedicated worktrees — it is not worktree-isolated the way a worktree's index/HEAD are.
A safer alternative that stays entirely within one worktree's own branch tip: a temporary local commit
[...]

PR #3175, opened 2026-09-01
recompiler: give flat-shaded fragment varyings SPIR-V Flat (#3051)

### LINKED ISSUE #3051: Fragment varyings never get SPIR-V Flat: is_flat_shaded is decoded, used for the control word, and discarded

mattias800:
## Summary

Fixes #3051: `AgcShaderSemantic::is_flat_shaded()` was decoded into the guest's PS_INPUT_CNTL
control word (`agc_shader_layout.cpp:98`/`:123`, the `0x400` FLAT_SHADE bit) and then discarded.
`flat_attrs` — the only thing that decorated a fragment SPIR-V Input variable `Flat` — was
populated from exactly one source: `VINTRP opcode == 2` (`v_interp_mov`). An attribute the guest
declared flat-shaded but the shader happened to read with an ordinary `v_interp_p1`/`v_interp_p2`
pair stayed smooth-interpolated, which is invalid for an integer/index varying and silently wrong
for anything else (a texture-array slice, a material/instance id).

## Behavioral contract

- On RDNA2 hardware, `SPI_PS_INPUT_CNTL.FLAT_SHADE` (bit 10, `0x400` — AMD "RDNA 2" ISA Reference,
  doc 70648) makes the parameter cache deliver only the provoking vertex's value for that
  interpolation slot (P10=P20=0). Every VINTRP read of that slot is therefore constant across the
  primitive **regardless of which VINTRP opcode the shader uses** — `v_interp_mov` and ordinary
  `v_interp_p1`/`v_interp_p2` both see the same flattened data on real hardware.
- Vulkan needs the `Flat` decoration on the corresponding `Input` variable to reproduce that;
  without it the driver perspective-interpolates, giving a different value per fragment.
- The guest's own control-word bit (`is_flat_shaded()`) is therefore authoritative and independent
  of — must be UNIONED with, not replace — the existing `v_interp_mov`-opcode detection, since
  either source alone misses cases the other catches (a shader can use `v_interp_mov` for a
  varying whose semantic isn't marked flat by the producer/consumer match, and vice versa).

**CONFIDENCE: HIGH** on the FLAT_SHADE-forces-provoking-vertex-value mechanism itself (directly
from the RDNA2 ISA doc's PS input control description, and consistent with how the existing,
already-shipped `v_interp_mov` -> `Flat` path is justified in `rdna2_to_spirv_internal.hpp`).
**CONFIDENCE: MED** on scope/impact — no title in the corpus has been observed to actually set the
bit. #3051's own investigation measured 224,363 live `[interp]` control words on *Tomb Raider I-III
Remastered* with none setting `0x400`, so this closes a **latent** correctness gap rather than a
live regression; the fix is defensive against the first title that does exercise it.

## Implementation

- `PixelInputMapping::effective_flat_mask()` (`rdna2_to_spirv.hpp`) extracts the `0x400` bit per
  slot from the guest's raw control words, mirroring the existing `effective_passthrough_mask()`.
- `fragment_interpolation_layout()` (`rdna2_to_spirv.cpp`) computes the new
  `FragmentInterpolationLayout::flat_mask` from it, alongside `passthrough_mask`, before the
  `requires_geometry` early-return (so it applies to every fragment program, not only ones needing
  the portable interpolation-geometry fallback).
- `frag_input()` in the SPIR-V builder (`rdna2_to_spirv_internal.hpp`) now decorates `Flat` when
  **either** `flat_attrs.count(attr)` (existing `v_interp_mov` detection) **or**
  `fragment_interpolation->flat_mask` has the bit set.
- `gpu_executor.cpp`'s `InterpolationCacheKey` (a separate memoization cache for
  `FragmentInterpolationLayout`, keyed on shader identity + `passthrough_mask` alone) now also
  carries `flat_mask`. Without this, two draws sharing one fragment shader's bytes but different
  guest FLAT_SHADE state — `SPI_PS_INPUT_CNTL` is a context register, so this can legitimately
  differ draw-to-draw for the same shader — could silently answer from a stale layout computed for
  the other draw's flat-ness. This is exactly the kind of caching bug the fix itself would have
  introduced if left unaddressed, mirroring how `passthrough_mask` is already handled there.
- The primary graphics-shader recompile cache (`ShaderCompileKey`/`ShaderCompileKeyHash`) already
  hashes the full raw `PixelInputMapping::controls[]` array bit-for-bit, so it needed no change.

- Affected: any fragment program reading an attribute whose guest PS_INPUT_CNTL sets FLAT_SHADE via
  an ordinary (non-`v_interp_mov`) VINTRP read. That varying now gets SPIR-V `Flat`.
- Unaffected: every existing `v_interp_mov`-flat and smooth-only case (both have dedicated
  regression coverage that keeps passing — see Testing), the primary shader recompile cache
  (already fully keyed), and any title/route not exercising FLAT_SHADE at all (measured: none in
  the corpus today).

## Testing

Added a case to `test_rdna2_spirv_struct.cpp` (pure structural test, no Vulkan device needed),
right after the existing "stays smooth" negative case: the same smooth-only instruction bytes are
recompiled again with a synthetic `PixelInputMapping` whose only set bit is `0x400` (FLAT_SHADE, not
the `0x420` passthrough form) on that attribute's slot. Asserts both `layout.flat_mask == 1` (the
layout computation) and `has_decoration(spv, Flat)` (the emitted module).

**Verified the test fails without the fix**: stashed only the `rdna2_to_spirv_internal.hpp` hunk
(the `frag_input()` consumer change), rebuilt, and reran — the new case failed with
`layout.flat_mask=00000001` printed alongside the failure (confirming the layout-side computation
was already correct and the failure was specifically the missing decoration), while every other
case still passed. Restored the fix and reran clean.

- `./test_rdna2_spirv_struct` (direct binary run): all cases pass, including the new one:
  `[ok] guest PS_INPUT_CNTL FLAT_SHADE forces Flat on an ordinary VINTRP read (#3051)`.
- Full suite: `ctest --no-tests=error` — **314/314 passed (100%)**, exit code 0, real time 151.55s.
- **Not yet run**: `spirv-val` on an emitted module carrying the new decoration. I did not get to
  this before handing off; the emitted `Flat` decoration follows the exact same code path
  (`Op_Decorate`/`Dec_Flat`) already used and validated for the `v_interp_mov` case, so I expect it
  to validate cleanly, but that is an expectation, not an observed result — please run it as part of
  review.

## Risks

- Low. The change only adds a decoration that was previously missing when a guest control word says
  it belongs; it cannot turn a previously-valid smooth varying into a flat one unless the guest's
  own control word says so. The cache-key fix is additive (widens a key, can only turn a stale hit
  into a correct miss, never the reverse).
- No title in the corpus currently exercises the bit, so there is no available live-boot evidence to
  cross-check the composite against; this is recorded as latent rather than regression-fixing in
  `docs/RECOMPILER_REMAINING.md`.

## Also filed

While isolating a `git stash` step for the fail-without-fix verification above, I hit and recovered
from a **shared `refs/stash` collision with a concurrent agent working in its own separate
worktree** — both agents' stash pushes interleaved on the single repo-wide stash ref, and each
agent's `pop` briefly picked up the other's unrelated WIP. Both agents independently recognized and
recovered the foreign content without loss, but it is a real hazard for CLAUDE.md's worktree
guidance. Filed separately as #3174 (process/tooling, not code — out of scope here).

## Docs

`docs/RECOMPILER_REMAINING.md` gets a new `## ... -- **FIXED**` entry recording the mechanism,
implementation, and the corpus measurement showing no title depends on it today.

### DIFF
--- a/prosper/docs/RECOMPILER_REMAINING.md
+++ b/prosper/docs/RECOMPILER_REMAINING.md
@@ -399,6 +399,34 @@ instruction later, at `bf8a0000` (`s_barrier`, pc=8) — the **generic NGG merge
 which is item 2 of the Recommendation list above and a frontier rather than a missing opcode. See
 `YAKUZA_JUDGMENT_BRINGUP.md`.
 
+## Fragment varyings never got SPIR-V `Flat` from the guest's own FLAT_SHADE bit — **FIXED** (2026-09-01)
+
+`AgcShaderSemantic::is_flat_shaded()` was decoded into the guest's PS_INPUT_CNTL control word
+(`agc_shader_layout.cpp:98`/`:123`, the `0x400` FLAT_SHADE bit) and then discarded: `flat_attrs`, the
+only thing that decorated a fragment Input variable `Flat`, was populated from exactly one source —
+`VINTRP opcode == 2` (`v_interp_mov`, #152/#897) — so an attribute the guest declared flat-shaded but
+the shader happened to read with an ordinary `v_interp_p1`/`v_interp_p2` pair stayed
+smooth-interpolated. On hardware FLAT_SHADE makes the parameter cache deliver only the provoking
+vertex's value for that slot (P10=P20=0), so *every* VINTRP read of it is a constant across the
+primitive regardless of opcode; Vulkan needs the `Flat` decoration to reproduce that.
+
+`FragmentInterpolationLayout` now carries `flat_mask`, from the new
+`PixelInputMapping::effective_flat_mask()` (mirrors `effective_passthrough_mask()`'s shape),
+computed in `fragment_interpolation_layout()` alongside `passthrough_mask` and unioned with the
+existing `flat_attrs` set in `frag_input()` — the two sources are unioned, not one substituted for
+the other, since either alone misses cases the other catches. It also had to be threaded into the
+`InterpolationCacheKey` memo (`gpu_executor.cpp`), which keyed on shader identity plus
+`passthrough_mask` alone: without `flat_mask` in that key, two draws sharing one fragment shader's
+bytes but different guest FLAT_SHADE state could answer from a stale layout computed for the other
+draw's flat-ness — a caching defect the fix itself would have introduced silently.
+
+No title in the corpus was found to exercise the bit before this fix — #3051's own investigation
+measured 224,363 live `[interp]` control words on *Tomb Raider I-III Remastered* with none setting
+`0x400` — so this closes a latent correctness gap rather than a live regression. `test_rdna2_spirv_struct`
+pins it: the same smooth-only bytes as the pre-existing "stays smooth" negative case are recompiled
+again with a synthetic `PixelInputMapping` control word carrying only `0x400`, and the emitted module
+must now carry a `Flat` decoration it did not have before. #3051
+
 ## Ruled out
 
 Cross-title falsifications where the **recompiler was blamed and exonerated**. One line per dead
--- a/prosper/src/gpu/execute/gpu_executor.cpp
+++ b/prosper/src/gpu/execute/gpu_executor.cpp
@@ -931,6 +931,14 @@ struct InterpolationCacheKey {
     PixelSystemInputMapping system_inputs{};
     bool has_system_inputs = false;
     uint32_t passthrough_mask = 0;
+    // #3051: FragmentInterpolationLayout now also carries flat_mask, derived from pixel_inputs's raw
+    // control words exactly as passthrough_mask is (PixelInputMapping::effective_flat_mask()). It
+    // must be part of this key for the same reason passthrough_mask is: SPI_PS_INPUT_CNTL is a
+    // context register that can carry a different FLAT_SHADE bit across two draws sharing identical
+    // shader bytes, and this cache is keyed on shader identity alone otherwise -- omitting it would
+    // let a flat-shaded draw's layout silently answer a later non-flat draw of the same shader, or
+    // vice versa.
+    uint32_t flat_mask = 0;
 
     bool operator==(const InterpolationCacheKey&) const = default;
 };
@@ -945,6 +953,7 @@ struct InterpolationCacheKeyHash {
             hash = hash_mix(hash, key.system_inputs.addr);
         }
         hash = hash_mix(hash, key.passthrough_mask);
+        hash = hash_mix(hash, key.flat_mask);
         return static_cast<size_t>(hash);
     }
 };
@@ -1892,6 +1901,7 @@ FragmentInterpolationLayout fragment_interpolation_layout_cached(
     key.has_system_inputs = system_inputs != nullptr;
     if (system_inputs) key.system_inputs = *system_inputs;
     key.passthrough_mask = pixel_inputs ? pixel_inputs->effective_passthrough_mask() : 0u;
+    key.flat_mask = pixel_inputs ? pixel_inputs->effective_flat_mask() : 0u;   // #3051
     auto& cache = interpolation_cache();
     std::lock_guard lock(cache.mutex);
     auto found = cache.entries.find(key);
--- a/prosper/src/gpu/recompiler/rdna2_to_spirv.cpp
+++ b/prosper/src/gpu/recompiler/rdna2_to_spirv.cpp
@@ -159,9 +159,16 @@ FragmentInterpolationLayout fragment_interpolation_layout(
         else if (instruction.opcode == 2 && instruction.src[0].value < 3)
             selectors[attr] |= static_cast<uint8_t>(1u << instruction.src[0].value);
     }
-    if (pixel_inputs)
+    if (pixel_inputs) {
         layout.passthrough_mask =
             pixel_inputs->effective_passthrough_mask() & layout.attribute_mask;
+        // #3051: the guest's own FLAT_SHADE control bit, independent of which VINTRP opcode the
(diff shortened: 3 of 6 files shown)


mattias800 (commented):
**APPROVED**

**Mutation verified independently.** Setting `layout.flat_mask = 0` at `rdna2_to_spirv.cpp:170` —
which reproduces the pre-fix "decoded, then discarded" behaviour exactly — builds clean and fails
with `[FAIL] guest FLAT_SHADE control word (#3051) did not force Flat on a v_interp_p1/p2 varying`;
restoring it passes. (My first attempt at this mutation broke compilation and the stale binary
"passed", which is a good reminder that a mutation arm proves nothing unless the build is green.)

**`spirv-val` run, since you flagged it as not done:** the `spv_validate` suite passes 4/4 on this
branch, and the full suite is green. So the emitted modules stay valid with the new `Flat`
decoration.

**The evidence for the semantics is the right kind.** `SPI_PS_INPUT_CNTL.FLAT_SHADE` (bit 10) making
the parameter cache deliver only the provoking vertex's value — P10=P20=0 — is what makes *every*
VINTRP read of that slot constant, not just `v_interp_mov`. That is a mechanism, not an analogy, and
unioning with the existing opcode detection rather than replacing it is the conservative composition.

**HIGH/MED split is honest and I agree with it.** HIGH on the mechanism; MED on impact, because no
title in the corpus sets the bit (224,363 live control words on Tomb Raider I–III, none with
`0x400`). Calling this a latent-defect fix rather than a live regression fix is the accurate framing,
and it is the sort of thing that usually gets quietly upgraded to sound more valuable.

**The caching gap you found on the way is the best part.** `InterpolationCacheKey` memoised
`FragmentInterpolationLayout` on shader identity plus `passthrough_mask` alone. Since FLAT_SHADE is a
*context register*, two draws sharing one shader's bytes but differing in guest flat state would have
answered from a stale layout — so without adding `flat_mask` to that key, the fix would have been
correct and still produced wrong output for the second draw. Catching a cache-key omission while
implementing the thing it caches is exactly the kind of second-order bug that normally ships.

Thanks also for filing #3174 on the `refs/stash` collision rather than absorbing it silently — that
is a real multi-agent hazard and it is now written down.

Verified: 315/315 in-container on my dump-configured build (314 on a clean no-dump configure), no
`Claude-Session:` trailer, no absolute host paths.

PR #3176, opened 2026-09-01
PROSPER_TEXCOMMIT: scan the source extent, not the decoded byte count

### LINKED ISSUE #3053: PROSPER_TEXCOMMIT scans the decoded byte count over the source address, overshooting 4x on BC and more on arrays
`PROSPER_TEXCOMMIT` walks the SOURCE address for `nb` bytes, but `nb` is the **decoded** byte count,
not the source extent (`live_renderer.cpp`, the `PROSPER_TEXCOMMIT` block):

```cpp
for (uint64_t a = sampled_source_addr; a < sampled_source_addr + nb; a += PG)
    if (a >= 0x1000 && prosper_reserved_range_state(a) != 0) committed += PG;
...
for (size_t i = 0; i < nb; i += 997) nzb += (p[i] != 0);   // sparse scan
```

`nb` is `tw * th * output_bpp * layers`. For any block-compressed texture that is already **4x** the
source (BC7 is 1 byte per texel decoded to 4), and for a multi-layer array it is multiplied again.

**Concrete overshoot, measured on Tomb Raider I-III Remastered:**

```
[texcommit] tex 0x20491cc000 512x512 f19 tile=5 nb=268435456 committed=100% nz~125886/269243
```

That texture is a 512x512 BC7 array of 256 slices. Its source extent is **67,108,864** bytes by the
T#'s own `size` (90,177,536 if the full mip chain is counted). `nb` is **268,435,456**. So both the
commit walk and the non-zero scan run **three to four times past the end of the texture**, over
whatever guest allocations happen to follow it.

**Why it matters beyond being wrong.** The comment above the code states its purpose exactly:

> If the level's backgrounds read ~0% committed, they're GPU-DMA'd pages the CPU never touched, so we
> read zeros -> the scene samples black.
> ... zero content => the texture was allocated but never filled ...; non-zero => it's a decode/tiling problem.

It cannot serve that purpose while scanning the wrong range: a texture that is genuinely unfilled
reads as partly non-zero if anything after it is populated, and `committed=100%` says only that
*something* occupies the following address space. Both readings point the investigation the wrong way,
and the second one — "non-zero => decode/tiling problem" — is exactly the wrong conclusion to hand
someone chasing an unfilled texture.

**It misled me.** I reported ~47% of an atlas as non-zero from this line and had to withdraw it on
#2998 after a per-slice content map disagreed (7% vs 47%). The per-slice map was right.

**Suggested fix:** scan the SOURCE span. The decode already computes it — `surface_bytes` /
`selected_span` in the block-compressed and linear branches, times the layer count for an array — or
fall back to the T#'s `size`. Whichever is chosen, the value used for `committed` and for `nzb` must
be a source-side quantity, and it is worth stating in the log line which extent was walked so a
future reader can tell.

Found while investigating #2998.

mattias800:
## Summary

Fixes #3053.

`PROSPER_TEXCOMMIT` walked `sampled_source_addr` for `nb` bytes, but `nb` is the **decoded** byte
count (`tw * th * decoded_bpp * layers`), not the guest **source** extent. For a block-compressed
format this overshoots by 4x+ (BC7 decodes 1 source byte/texel to 4 RGBA8 bytes), and a layered
array multiplies that again by the layer count.

Concrete numbers from the issue (Tomb Raider I-III Remastered): a 512x512 BC7 world-atlas texture
with 256 tiled layers measured `nb` at 268,435,456 while its real source extent is 67,108,864 -- a
4x overshoot. Both the "committed" scan and the sparse non-zero scan therefore ran three-to-four
times past the end of the real texture, over whatever guest allocations happened to follow it,
which is worse than useless for the diagnostic's own stated purpose (distinguishing "unfilled
texture" from "decode/tiling bug") -- it misled the investigation on #2998, which is what surfaced
this bug.

## Fix

Scan `prosper::gpu::gpu_capture_resource_footprint(r)` instead of `nb`. That function is the same
T#-driven source-byte computation the capture/replay path already uses for this exact resource --
BC block bytes, tiling, and the array/cube layer span already folded in -- and it's already
extensively tested against real title numbers in `test_gpu_capture.cpp` (BC arrays, cubes, mip
tails, linear row padding, etc.), so no new formula is introduced.

The call site routes through a new pure function, `texcommit_scan_extent()`, added to
`frontends/shared/texture/texture_decode_cache_policy.hpp` (the existing home for this class of
"how many bytes does this texture decode/scan actually touch" helper, alongside
`layered_array_source_size()` and `layered_cube_source_size()` from #2998). Its signature
deliberately has **no parameter for a decoded byte count**, so the exact mistake this issue reports
cannot be reintroduced by passing the wrong variable at the call site.

The log line now reports `src=<bytes scanned>` alongside the existing `nb=<decoded bytes>`, so a
future reader can tell which extent was walked -- per the issue's own suggestion.

## Behavioral contract

- `PROSPER_TEXCOMMIT` (an opt-in diagnostic, off by default) now scans and reports on the guest
  SOURCE bytes a sampled texture actually occupies, never the backend's decoded byte count.
- No change to any non-diagnostic code path: `nb` itself, and everything that uses it to size/decode
  `texture_pixels`, is untouched.
- Fixes a real out-of-bounds read: the old code could read guest memory 3-4x past the real
  allocation for BC/array textures.

## Verification

**Reproduced/confirmed on current master before changing anything:** computed
`gpu_capture_resource_footprint()` for a synthetic `ShaderResource` matching the issue's exact Tomb
Raider shape (512x512 BC7, `img_dim=5`, `depth=256`, `tile_mode=5`/SW_4KB_S -- matching the issue's
own "tile=5" log line) and confirmed it returns exactly 67,108,864, a quarter of the
268,435,456-byte `nb` the old code scanned -- reproducing the issue's reported numbers exactly.

**New test** (`test_texture_decode_diagnostic.cpp`), covering the BC case and the array case
together (the reported texture is both):
- Pins `gpu_capture_resource_footprint()` for the Tomb Raider BC7/256-layer shape at exactly
  67,108,864, and the old `nb` formula at exactly 268,435,456 (4x, matching the issue).
- Asserts `texcommit_scan_extent()` returns the source footprint, never the decoded count.

**Verified the test fails without the fix**: temporarily changed `texcommit_scan_extent()`'s body to
`source_footprint_bytes * 4` (simulating the pre-fix BC overshoot), rebuilt just that test target,
and reran it -- all three new assertions failed:
```
FAIL texcommit_scans_the_source_footprint: texcommit_scan_extent(tomb_raider_footprint) == tomb_raider_footprint
FAIL texcommit_never_scans_the_decoded_byte_count: texcommit_scan_extent(tomb_raider_footprint) != tomb_raider_decoded_nb
FAIL ...: texcommit_scan_extent(65536u) == 65536u
EXIT=1
```
Restored the real implementation, rebuilt, and reran -- green (`texture_decode_diagnostic: OK`,
exit 0).

Note on scope: the wiring decision at the `live_renderer.cpp` call site (use the footprint, not
`nb`) is not itself exercised by a unit test -- `live_renderer.cpp`'s draw/texture-decode function is
not practically unit-testable in isolation, and no existing test harness in this repo captures a live
`PROSPER_*` diagnostic's stderr output. The extracted `texcommit_scan_extent()` function's signature
(no decoded-byte-count parameter) is the structural guard against that specific class of regression
at the call site.

**Build:**
```
TMPDIR=.../build-linux/tmpdir cmake --build prosper/build-linux -j6   # exit 0, full rebuild clean
```

**Full ctest:**
```
ctest --no-tests=error
100% tests passed out of 314
Total Test time (real) = 125.45 sec
```
(314 matches the documented fresh-configure baseline.)

## Risk / scope

Diagnostic-only change, gated behind `PROSPER_TEXCOMMIT` (opt-in, off by default in all normal runs
and CI). No behavioral change to rendering, decoding, or any non-diagnostic code path.

### DIFF
--- a/prosper/frontends/shared/live/live_renderer.cpp
+++ b/prosper/frontends/shared/live/live_renderer.cpp
@@ -4843,27 +4843,40 @@ void register_live_renderer(const std::string& frame_dir, bool dump_bmps_request
                         // level's backgrounds read ~0% committed, they're GPU-DMA'd pages the CPU never
                         // touched, so we read zeros -> the scene samples black (#300 black-gameplay probe).
                         if (PROSPER_ENV_ON("PROSPER_TEXCOMMIT")) {
+                            // Walk the SOURCE extent, never `nb` (the DECODED byte count computed just
+                            // above for `texture_pixels`): for a block-compressed format `nb` is 4x+ the
+                            // real guest bytes (BC7 decodes 1 source byte/texel to 4 RGBA8 bytes), and an
+                            // array multiplies that again by the layer count. Scanning `nb` over
+                            // `sampled_source_addr` both mis-measured "committed" (it counted whatever
+                            // guest memory happened to follow the real texture) and read guest memory
+                            // past the real allocation (#3053). `gpu_capture_resource_footprint()` is the
+                            // same T#-driven source-byte computation the capture/replay path already uses
+                            // for this resource, so it is correct here too regardless of format.
+                            const size_t src_bytes = texcommit_scan_extent(
+                                prosper::gpu::gpu_capture_resource_footprint(r));
                             const size_t PG = 0x10000; size_t committed = 0;
                             for (uint64_t a = sampled_source_addr;
-                                 a < sampled_source_addr + nb; a += PG)
+                                 a < sampled_source_addr + src_bytes; a += PG)
                                 if (a >= 0x1000 && prosper_reserved_range_state(a) != 0) committed += PG;
                             static std::set<uint64_t> tcseen;
                             if (tcseen.insert(r.gpu_addr).second) {
                                 // Also sample the first 8 dwords and count non-zero bytes over the whole
-                                // surface: zero content => the texture was allocated but never filled (a
-                                // GPU-side upload/copy we don't execute); non-zero => it's a decode/tiling
-                                // problem. tile_mode tells tiled vs linear.
+                                // SOURCE surface (src_bytes, not nb): zero content => the texture was
+                                // allocated but never filled (a GPU-side upload/copy we don't execute);
+                                // non-zero => it's a decode/tiling problem. tile_mode tells tiled vs linear.
                                 uint32_t w0[8] = {0}; size_t nzb = 0;
                                 if (committed) {
                                     const uint8_t* p =
                                         (const uint8_t*)(uintptr_t)sampled_source_addr;
                                     for (int i = 0; i < 8; i++) w0[i] = ((const uint32_t*)p)[i];
-                                    for (size_t i = 0; i < nb; i += 997) nzb += (p[i] != 0);   // sparse scan
+                                    for (size_t i = 0; i < src_bytes; i += 997) nzb += (p[i] != 0);   // sparse scan
                                 }
-                                fprintf(stderr, "[texcommit] tex 0x%llx %ux%u f%u tile=%u nb=%zu committed=%zu%% "
-                                        "nz~%zu/%zu first=%08x %08x %08x %08x\n",
-                                        (unsigned long long)r.gpu_addr, tw, th, (unsigned)r.format, r.tile_mode, nb,
-                                        nb ? (size_t)(100 * committed / nb) : 100, nzb, nb/997,
+                                fprintf(stderr, "[texcommit] tex 0x%llx %ux%u f%u tile=%u nb=%zu src=%zu "
+                                        "committed=%zu%% nz~%zu/%zu first=%08x %08x %08x %08x\n",
+                                        (unsigned long long)r.gpu_addr, tw, th, (unsigned)r.format, r.tile_mode,
+                                        nb, src_bytes,
+                                        src_bytes ? (size_t)(100 * committed / src_bytes) : 100, nzb,
+                                        src_bytes / 997,
                                         w0[0], w0[1], w0[2], w0[3]);
                             }
                         }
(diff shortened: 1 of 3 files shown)


mattias800 (commented):
**APPROVED**

**Mutation verified independently.** Changing `texcommit_scan_extent()` to return
`source_footprint_bytes * 4` — simulating exactly the BC overshoot this fixes — builds clean and
fails the test; restoring it passes. Genuine discriminator.

**The reproduction is the right kind.** Building a synthetic `ShaderResource` matching the issue's
own reported case (512x512 BC7, `img_dim=5`, `depth=256`, `tile_mode=5`) and showing
`gpu_capture_resource_footprint()` returns 67,108,864 against the old code's 268,435,456 — exactly
4x — reproduces the reported numbers rather than merely asserting the ratio. That is what makes the
fix checkable instead of plausible.

**Reusing `gpu_capture_resource_footprint` is the right call over writing new arithmetic.** It is the
same T#-driven source-byte computation the capture/replay path already uses, with BC block bytes,
tiling and array/cube layer span folded in, and it is already exercised against real title numbers in
`test_gpu_capture.cpp`. A second, independently-derived formula here would have been one more thing to
keep in sync.

**The signature-as-guard point is a good one.** Giving `texcommit_scan_extent()` no parameter slot for
a decoded byte count means the original mistake cannot be reintroduced at the call site by accident —
a type-level fix for a class of error, not just this instance. That is worth more than the test.

**Honest about the residual gap**, which I agree with: the unit test pins the extent function, not the
`live_renderer.cpp` wiring, because that call site is not practically unit-testable in isolation and
no harness here captures a `PROSPER_*` diagnostic's stderr. Saying so plainly rather than implying
end-to-end coverage is the right treatment, and the signature guard is a reasonable mitigation.

Diagnostic-only, gated behind an opt-in that is off by default including in CI, so blast radius is
limited to the diagnostic's own output — which is exactly where the defect was.

PR #3177, opened 2026-09-01
tools/perf: fix audio_delivery_report's +1/(N-1) drift bias and burst-clustering wording

### LINKED ISSUE #3061: audio_delivery_report: drift is biased +1/(N-1) by the emitter's zero first gap, and the burst-clustering row does not test the conjunct its docstring promises
Two defects in `prosper/tools/perf/audio_delivery_report.py`, both found in review of #3046 and
deliberately left out of it — that PR is about the queue thresholds, and these are in the cadence and
drift analysis.

## 1. The drift figure is biased by `+1/(N-1)`

The `[audio-dbg]` emitter logs `gap=0.00ms` for a port's **first** arrival, because there is no previous
arrival to measure from. The report then sums all gaps as the elapsed span:

```
span_s = sum(gaps) / 1000.0
...
delivered_hz = frames_total / span_s
```

So the span is short by one mean gap while the frame total is not, and `delivered_hz` is inflated by
roughly `1/(N-1)`.

**Consequence:** at small N the tool reports the guest delivering *faster* than the device and can print
a verdict to that effect. Measured in review at **N=60**: a false "runs above the device rate". At
N=5,970 — the sample size behind the numbers on #3016 — the bias is ~0.017%, well inside the ±0.01%
resolution those figures were quoted at, so **#3016's conclusions stand** and this is not a correction to
them.

**Fix:** drop the first record's gap from the span (or from the frame total to match), and say which in a
comment, since the two differ by one grain.

## 2. The docstring promises a burst conjunct the code does not test

The header advertises:

> burst clustering -- the fraction of arrivals that carry **more than one grain of audio** after a gap
> of more than two grain periods

The implementation tests only the gap half. Nothing examines frames-per-arrival, so a run of ordinary
single-grain arrivals after long gaps is reported as burst clustering. That is the same shape as the
label defect #3046 fixes — text describing a stronger property than the code checks — and it is the kind
of row that gets quoted, because "burst clustering" reads as a diagnosis.

**Fix:** either test the conjunct (the emitter logs `frames=` per record, so it is available) or narrow
the wording to what is measured.

## Why filed rather than folded into #3046

#3046 fixes the queue rows and pins them with mutation-checked arms. Extending it into the cadence and
drift rows would mix two unrelated corrections in one review, and the reviewer's point that a
falsification living only in a review thread is one the next agent re-derives is the reason this exists
as an issue rather than a paragraph there.

Note for whoever takes this: the existing self-test asserts label presence far more than computed values,
so **adding an arm is part of the fix, not optional**. #3046 added four mutation-checked arms and its
review found that its own headline change was initially unpinned — the same trap is waiting here.

mattias800:
Fixes #3061

### 1. Drift figure biased by +1/(N-1)

The `[audio-dbg]` emitter logs `gap=0.00ms` for a port's first arrival (no previous arrival to
measure from). `span_s = sum(gaps)` therefore covers only the `(calls - 1)` REAL inter-arrival
intervals -- the elapsed time from the first arrival to the last -- while `frames_total` sums
frames from all `calls` arrivals, including that first one, whose own accumulation time was
never captured by any logged gap. Dividing the full frame total by that `(calls - 1)`-interval
span overcounts by exactly one arrival's worth of frames.

**Derivation.** For `calls = N` records with a constant grain size `F` and a true steady-state
gap `g` between the `N - 1` real arrivals:

```
span_s (as computed)      = (N - 1) * g            -- unchanged by the fix
frames_total               = N * F
delivered_hz (pre-fix)     = frames_total / span_s = N * F / ((N - 1) * g)
delivered_hz (true rate)   = F / g                  -- one grain per grain period, steady state
```

So `delivered_hz (pre-fix) = delivered_hz (true) * N / (N - 1)`, i.e. inflated by a factor of
`N / (N - 1)`, or roughly `+1/(N - 1)` -- exactly the bias the issue names.

**Fix:** drop the first arrival's frames from the numerator instead of stretching `span_s` by an
imputed extra grain period -- the two give the same answer only when every arrival is the same
size, and dropping the frame needs no such assumption. Gated on the emitter's own `gap == 0.0`
sentinel (not just "index 0"), so a log excerpt that starts mid-stream -- where the first logged
gap is a real, already-elapsed interval rather than the emitter's "no previous arrival" zero --
is left unadjusted rather than having a genuine arrival's frames silently discarded.

**Analytic test case (added as case 9).** 11 records of 100 frames each, delivered at an
*exact* real-time cadence against a 1000 Hz device: gap=0.00ms for the first record, then ten
real gaps of exactly 100.00 ms (100 frames / 1000 Hz * 1000 = 100 ms per grain, i.e. exactly
real time).

```
true_drift   = (frames_after_first / real_span_s - device_hz) / device_hz
             = ((10 * 100) / 1.0 - 1000) / 1000 = 0.00%

biased_drift = (frames_total / real_span_s - device_hz) / device_hz
             = ((11 * 100) / 1.0 - 1000) / 1000 = +10.00%    (== calls/(calls-1) - 1 = 1/10)
```

Pre-fix output (captured by temporarily reverting only the numerator change and re-running):

```
effective delivery: 1100 frames/s vs device 1000 -> drift +10.00%
VERDICT: the guest's audio clock runs above the device rate -- the cushion grows until the
depth cap; drift compensation should slow the delivery.
```

Post-fix output:

```
effective delivery: 1000 frames/s vs device 1000 -> drift +0.00%
VERDICT: delivery matches the device rate -- a stutter at this rate is a burst/quantization
problem, not a clock deficit. See the cadence and burst rows above.
```

+10.00% is well past the tool's own +-0.3% threshold, so the bug does not just misreport a
number -- at small N it flips the verdict to a false "runs above the device rate", exactly as
#3016's review found at N=60.

**Verification that the test actually discriminates:** disabled only the fix (reverted the
numerator line to the pre-fix `frames_in_span = frames_total`), re-ran
`test_audio_delivery_report.py`, and all three new assertions failed with the +10.00%/false-
verdict output quoted above. Restored the fix and confirmed `git diff` against the pre-revert
file was empty, then re-ran green.

### 2. Burst-clustering docstring promised an untested conjunct

The header advertised: "the fraction of arrivals that carry **more than one grain of audio**
after a gap of more than two grain periods." The code only ever tested the gap half
(`bursts = sum(1 for g in gaps if g > 2 * grain_ms)`); nothing examines `frames=` per record, so
ordinary single-grain arrivals separated by a long gap are folded into the same "clustered"
count as a real multi-grain flush.

**Chosen fix: narrow the wording, not the code** (the issue offered both options). Reasoning:
the code's gap-only gate is not an oversight -- it is exactly what the #3168 three-way verdict
(RULES OUT / CONFIRMS / hedge, driven by the `PROSPER_TIMEDWAIT_CENSUS=1` ratio) needs to see in
order to diagnose the #3080 scenario: a producer delivering *ordinary single-grain* arrivals,
delayed by a clock deficit rather than a quantized mixer wake. That scenario, by definition,
never carries more than a grain per arrival. Adding the frames-per-arrival conjunct to the gate
would make it blind to precisely the case #3168 was built to diagnose -- the existing self-test
cases 6/7/8 (the #3080 reconstruction: 488 single-grain records at gap=11.85ms) would stop
entering the verdict block at all, silently regressing #3168's own coverage.

So: reworded the docstring bullet to describe what is actually measured (gap-only), and added a
comment at the `bursts` computation explaining *why* it stays gap-only, so a future reader
doesn't "fix" it the wrong way. No behavioral/runtime change for this half -- confirmed the
existing self-test (cases 3, 6, 7, 8, which exercise this exact code path) is unaffected.

## Interaction with #3168's verdict logic

Explicitly: **defect 2's fix does not touch #3168's three-way verdict branching at all** --
only the docstring and an explanatory comment changed. Defect 1's fix touches the `delivered_hz`
number that feeds the *drift* verdict above the burst block, not the burst block itself, and
none of #3168's added test cases (6, 7, 8) model a genuine `gap=0.00` first record (they all use
a constant `gap` value for every record via the shared `dbg()` helper), so the drift fix is a
no-op for those fixtures and their expected output is unchanged.

## Testing

```
python3 prosper/tools/perf/test_audio_delivery_report.py
# audio_delivery_report self-test: all cases pass (10 cases, exit 0)
```

`git diff --check`: clean. `python3 -m py_compile` on both changed files: clean.
`git log origin/master..HEAD --format=%B | grep -c '^Claude-Session:'` -> 0.

This is a pure-Python tool under `tools/perf/`; no C++ build or ctest is implicated.

### DIFF
--- a/prosper/tools/perf/audio_delivery_report.py
+++ b/prosper/tools/perf/audio_delivery_report.py
@@ -17,16 +17,20 @@ that log into the numbers that name the failure mode, offline, without listening
   * effective delivery rate vs the device rate -- a persistent deficit is a CLOCK DRIFT
     between the guest's budgeted audio clock and the device crystal; no cushion survives it,
     and the fix is drift compensation, not deeper buffering;
-  * burst clustering -- the fraction of arrivals that carry more than one grain of audio
-    after a gap of more than two grain periods. This shape is CONSISTENT WITH a quantized
-    mixer wake, but the cadence alone cannot prove it: a producer simply delivering audio
-    below real time also arrives in clusters, spaced by production rather than by a wait
-    (#3080 -- the verdict named the wrong cause on a title where a `PROSPER_TIMEDWAIT_CENSUS=1`
-    measurement, in the same log, showed the guest's only timed wait resolving at x1.02, i.e.
-    not quantized at all; the -55% delivery-rate deficit already explained the clustering). So
-    if `[timedwait Ns]` census lines are present in the same input, their ratio for the
-    dominant wait primitive SETTLES which cause applies; if they are absent, the report says so
-    and names the census as the discriminator instead of asserting a mechanism it cannot see.
+  * burst clustering -- the fraction of arrivals whose inter-arrival GAP exceeds two grain
+    periods (#3061: earlier wording here promised a stronger conjunct -- "carries more than
+    one grain of audio" -- that the code has never tested; frames-per-arrival is available in
+    the log but is not examined, so this is gap-only, same as the code). This shape is
+    CONSISTENT WITH a quantized mixer wake, but the cadence alone cannot prove it -- and, being
+    gap-only, it does not distinguish a real multi-grain flush from a producer simply
+    delivering audio below real time, which also arrives in clusters, spaced by production
+    rather than by a wait (#3080 -- the verdict named the wrong cause on a title where a
+    `PROSPER_TIMEDWAIT_CENSUS=1` measurement, in the same log, showed the guest's only timed
+    wait resolving at x1.02, i.e. not quantized at all; the -55% delivery-rate deficit already
+    explained the clustering). So if `[timedwait Ns]` census lines are present in the same
+    input, their ratio for the dominant wait primitive SETTLES which cause applies; if they are
+    absent, the report says so and names the census as the discriminator instead of asserting a
+    mechanism it cannot see.
 
 WHAT THIS TOOL DOES NOT SEE
 ---------------------------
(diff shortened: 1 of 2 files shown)


mattias800 (commented):
**APPROVED**

**Mutation verified independently.** Reverting the numerator to `frames_total / span_s` — the
pre-fix biased form — fails with
`case 9: an exact real-time 11-record log must report the true 0.00% drift, not +10.00%`, and the
fix restores green.

**The derivation is right and the test is the correct shape.** `span_s` sums `calls-1` real
inter-arrival intervals because the emitter logs `gap=0.00` for the first arrival, while
`frames_total` sums all `calls` records — so `delivered_hz` was inflated by exactly
`calls/(calls-1)`. Subtracting the first record's frames aligns numerator and denominator.

What makes the test worth having is that its expected value is **known analytically rather than
measured**: 11 records of 100 frames at an exact 1000 Hz cadence has a true drift of 0.00% by
construction, so the pre-fix +10.00% is provably wrong rather than merely different. That is the
difference between a regression test and a snapshot of current behaviour.

Gating the correction on `slot["gaps"][0] == 0.0` — the emitter's genuine no-previous-arrival
sentinel — is also right: a truncated log excerpt whose first record has a real gap must not be
"corrected".

**The defect-2 decision is the best judgment call in this PR, and I want it on the record.** You
declined to add the frames-per-arrival conjunct to the `bursts` gate, because doing so would make the
gate blind to exactly the #3080 scenario — ordinary single-grain arrivals delayed by a clock deficit
— that #3168's three-way verdict was built to diagnose, silently regressing cases 6/7/8. Choosing to
narrow the *docstring* to describe only what is measured, rather than widening the code to match a
docstring that was overclaiming, is the correct direction. It would have been very easy to "fix both
defects" and quietly delete working coverage.

Also correct that Defect 1's fix is a no-op on #3168's fixtures, since none model a genuine
`gap=0.00` first record — so the two changes do not interact.

Verified: 10 self-test cases pass, `diff --check` clean, no `Claude-Session:` trailer, no absolute
host paths.

PR #3178, opened 2026-09-01
test(#3057): bind dmem caller-chain attribution to the alloc_dmem call site

### LINKED ISSUE #3057: test: #3055's arms cover the formatter, not the alloc_dmem call site — deleting the fix leaves the suite green
## The gap

#3055 makes `sceKernelAllocateDirectMemory` publish a dmem caller chain, and adds two test arms in
`prosper/tests/hle/memory/test_dmem_caller_chain.cpp`. Those arms test the **formatter** — that the
allocator label is carried from the caller rather than hardcoded. They are mutation-verified
against *that* property.

**They do not test the fix.** Deleting the two new call sites — the
`attribute_dmem_allocation(...)` calls in `k_alloc_dmem`, one in the Linux block and one in the
Windows block of `prosper/src/hle/memory/hle_kernel_mem.cpp` — removes the entire behaviour change
and **leaves the suite green**. Found in review of #3055.

This is the failure shape the charter calls out: an assertion that is true, and true for a reason
unrelated to the mechanism it is supposed to protect. Nothing would go red if a later refactor
dropped the call.

## Why it wasn't closed in #3055

The obvious observable, `GuestDmemWriteTraceSnapshot::allocation_matches`, cannot be driven from a
test: matching requires a configured `caller_chain`, a test stack yields no guest return addresses
so the chain interns as `Unknown` (id 0), and `guest_dmem_write_trace_configure` rejects a config
with `caller_chain == 0` (`guest_write_watch.cpp`, the `!config.allocation_size || !config.size`
validity block). So the natural counter is unreachable.

## Suggested approach

The reachable discriminator is the record itself. `DmemCallerChainInterner::intern()` returns
`first == true` on the **first** Unknown chain (`dmem_caller_chain.hpp`), so
`attribute_dmem_allocation` does print

```
[dmem-caller] caller-chain=unknown alloc_dmem len=0x... from <no guest return addresses>
```

even with no guest frames on the stack. A test can therefore:

1. `setenv("PROSPER_DMEM_CALLER", "1")` **before** the first dmem call — `dmem_caller_log()` caches
   its `getenv` in a function-local static, so ordering matters and a dedicated binary is safest;
2. redirect `stderr` to a `tmpfile()`;
3. call `sceKernelAllocateDirectMemory` through `Hle::lookup(nid_hash("sceKernelAllocateDirectMemory"))`
   — `prosper/tests/host/test_dmem.cpp` already sets up exactly this machinery;
4. assert the captured text contains a `[dmem-caller]` line naming **`alloc_dmem`**.

Without the fix, step 4 finds nothing, because that allocator publishes no record at all — which is
precisely the bug. Verify by mutation: delete the call site and confirm the new arm goes red.

Worth covering the `alloc_main_dmem` path in the same test so a regression that swaps the labels
cannot pass either.

mattias800:
## The gap (#3057)

#3055 published a dmem caller chain from `sceKernelAllocateDirectMemory`, not only the Main
variant, by adding an `attribute_dmem_allocation(...)` call to `k_alloc_dmem`'s success path
(`prosper/src/hle/memory/hle_kernel_mem.cpp`, one call in the Linux block, one in the Windows
block). Its two new arms in `test_dmem_caller_chain.cpp` only exercise
`format_dmem_caller_chain_definition` — the **formatter** — and never drive an allocation through
the real dispatch table. Deleting both call sites removes the entire behaviour change and leaves
that suite green.

## What I confirmed, by direct observation

I applied the exact mutation the issue names — deleted both `attribute_dmem_allocation(...)` calls
in `k_alloc_dmem` (Linux block and Windows block) — rebuilt, and ran the tests directly:

- **Blindness confirmed**: with the mutation applied, `test_dmem_caller_chain` (the pre-existing
  #3055 arms) stayed **100% green**, all 12 `PASS` lines, exit 0. This reproduces the bug the issue
  reports: the suite is blind to the call site.
- **New arm binds to the call site**: under the same mutation, the new
  `test_dmem_caller_attribution_alloc_dmem` binary **failed** (exit 1, empty stderr capture —
  `attributed=0`).
- **Discrimination is specific, not incidental**: the sibling
  `test_dmem_caller_attribution_alloc_main_dmem` — guarding the *other*, untouched call site —
  **stayed green** under the same mutation, so the new arm's failure is caused by the deleted call,
  not by some global breakage.
- Restored the fix (`git checkout -- prosper/src/hle/memory/hle_kernel_mem.cpp`), rebuilt, reran all
  three binaries: all green again.

I only rebuilt/ran the four targeted binaries above (`test_dmem_caller_chain`,
`test_dmem_caller_attribution_alloc_dmem`, `test_dmem_caller_attribution_alloc_main_dmem`,
`test_dmem`) directly and observed their output myself. **I did not obtain a full-suite `ctest`
count for this PR** — the full-project rebuild I attempted exceeded the interactive foreground
timeout and moved to the background, and its completion notification never reached this session.
I am not reporting a ctest count I did not personally observe; the coordinator will run the full
build/test/verification themselves.

## What the new arms bind to

Two dedicated ctest binaries, each its own process:

- `prosper/tests/hle/memory/test_dmem_caller_attribution_alloc_dmem.cpp` — sets
  `PROSPER_DMEM_CALLER=1` before the first dmem call (required: `dmem_caller_log()` caches its
  `getenv()` read in a function-local static), redirects `stderr` to a `tmpfile()`, calls
  `sceKernelAllocateDirectMemory` through `Hle::lookup(nid_hash(...))` exactly as a guest does, and
  asserts the captured text contains
  `[dmem-caller] caller-chain=unknown alloc_dmem len=0x...`.
- `prosper/tests/hle/memory/test_dmem_caller_attribution_alloc_main_dmem.cpp` — same recipe for
  `sceKernelAllocateMainDirectMemory`, asserting `...alloc_main_dmem len=0x...`. This call site
  predates #3055 and isn't the regression #3057 is about, but the issue asked for it too ("a
  regression that swaps the labels cannot pass either"), and it also serves as the specificity
  control described above.

**Why two binaries, not one test with two arms** (per the issue's own follow-up review comment):
`DmemCallerChainInterner`'s `Unknown` state is a **one-shot per process** —
`attribute_dmem_allocation` returns before printing anything once `correlation.first` is false. A
bare host test stack has no return address inside a fixed guest module range (no image is loaded),
so every allocation in this harness interns as `Unknown`. If both allocators ran in one process, the
first call would consume the one-shot print and the second assertion would find an empty capture
**even with the fix present** — a false failure, not a real one. Separate processes give each
allocator its own first-Unknown observation.

## What I did with the existing formatter arms

**Kept them unchanged.** They are still the only coverage for
`format_dmem_caller_chain_definition`'s own contract — that the allocator label is carried from the
caller and not hardcoded, that an empty label falls back to `<unknown-allocator>` instead of a
plausible default, and that concurrent writers each produce one intact, uncorrupted line. That is a
real, independently useful property. It just never claimed to cover the call site, and now something
else does.

## Files

- `prosper/tests/hle/memory/test_dmem_caller_attribution_alloc_dmem.cpp` (new)
- `prosper/tests/hle/memory/test_dmem_caller_attribution_alloc_main_dmem.cpp` (new)
- `prosper/CMakeLists.txt` — registers both as ctest binaries right after `dmem_caller_chain`

## Verification commands (for the reviewer to reproduce)

```bash
TMPDIR=<worktree>/prosper/build-linux/tmpdir cmake --build prosper/build-linux -j6 \
    --target test_dmem_caller_attribution_alloc_dmem test_dmem_caller_attribution_alloc_main_dmem \
    test_dmem_caller_chain
./prosper/build-linux/test_dmem_caller_attribution_alloc_dmem
./prosper/build-linux/test_dmem_caller_attribution_alloc_main_dmem
```

Mutation test (delete both `attribute_dmem_allocation(...)` calls in `k_alloc_dmem`, rebuild, rerun):
`test_dmem_caller_attribution_alloc_dmem` should fail; `test_dmem_caller_chain` and
`test_dmem_caller_attribution_alloc_main_dmem` should stay green.

Fixes #3057

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -1497,6 +1497,23 @@ if(TARGET prosper_core)
   target_include_directories(test_dmem_caller_chain PRIVATE src)
   add_test(NAME dmem_caller_chain COMMAND test_dmem_caller_chain)
 
+  # #3057: test_dmem_caller_chain above covers format_dmem_caller_chain_definition (the formatter)
+  # only -- it never calls k_alloc_dmem/k_alloc_main_dmem, so deleting the actual
+  # attribute_dmem_allocation(...) call sites #3055 added left that suite green. These two bind to
+  # the real call sites through the NID dispatch table, one process each (see the header comment in
+  # the first file for why they cannot share a process).
+  add_executable(test_dmem_caller_attribution_alloc_dmem
+                 tests/hle/memory/test_dmem_caller_attribution_alloc_dmem.cpp)
+  target_link_libraries(test_dmem_caller_attribution_alloc_dmem PRIVATE prosper_core)
+  add_test(NAME dmem_caller_attribution_alloc_dmem
+           COMMAND test_dmem_caller_attribution_alloc_dmem)
+
+  add_executable(test_dmem_caller_attribution_alloc_main_dmem
+                 tests/hle/memory/test_dmem_caller_attribution_alloc_main_dmem.cpp)
+  target_link_libraries(test_dmem_caller_attribution_alloc_main_dmem PRIVATE prosper_core)
+  add_test(NAME dmem_caller_attribution_alloc_main_dmem
+           COMMAND test_dmem_caller_attribution_alloc_main_dmem)
+
   # Capped-diagnostic rate limit (issue #1761): a print budget must never be readable as a
   # population. Pins the sparse tail and the per-key budget; header-only, no GPU state.
   add_executable(test_diag_ratelimit tests/gpu/diagnostics/test_diag_ratelimit.cpp)
--- a/prosper/tests/hle/memory/test_dmem_caller_attribution_alloc_dmem.cpp
+++ b/prosper/tests/hle/memory/test_dmem_caller_attribution_alloc_dmem.cpp
@@ -9,23 +9,45 @@
 // the first dmem call (dmem_caller_log() caches its getenv() read in a function-local static, so
 // ordering matters), redirect stderr to a capture stream, call sceKernelAllocateDirectMemory
 // through the real NID dispatch table exactly as a guest does, and require the one-shot
-// "[dmem-caller] caller-chain=unknown alloc_dmem ..." line that attribute_dmem_allocation prints
-// even with no guest frames on the stack (a bare host stack has no return address inside a fixed
-// guest module range, so the chain always interns as Unknown here -- see dmem_caller_chain.hpp).
+// "[dmem-caller] caller-chain=..." line that attribute_dmem_allocation prints on the first
+// allocation this process makes.
 //
 // Without the fix this allocator publishes no record at all, so the line is never printed and the
 // assertion below fails -- verified by mutation as part of this PR (delete the call site, rerun,
 // confirm red; restore, rerun, confirm green).
 //
+// WHAT THE ASSERTION CHECKS, AND WHY IT DOES NOT PIN THE "unknown" FORM (review correction, #3178):
+// an earlier version of this test required the exact text "caller-chain=unknown alloc_dmem ...",
+// reasoning that a bare host stack (no guest image is loaded here) has no return address inside a
+// fixed guest module VA range, so the scan would always come up empty and the chain would always
+// intern as Unknown. That reasoning was environment-dependent, not a property of the call site:
+// guest_va_in_module_code (boot_program.hpp) tests only whether a VALUE FALLS IN A FIXED ADDRESS
+// RANGE, never whether anything is actually mapped there. On CI's Linux runner an ordinary garbage
+// word already sitting on the host stack landed inside the BOOT_SAVEDATA range by coincidence, so
+// the scan resolved a (bogus) "SaveData.prx+0x10000002" and the chain interned as Known -- and the
+// exact-"unknown" assertion failed even though attribution had genuinely happened. That is the same
+// failure shape #3057 exists to fix, just relocated one level down: the arm was bound to "nothing on
+// this machine's stack happens to look like a guest address" rather than to the call site.
+// attribute_dmem_allocation's OWN comment says as much: the scan is a heuristic over live stack
+// content, "a stale spill slot looks exactly like a return address". Forcing the Unknown state
+// deterministically would mean controlling that same uninitialized memory, which is exactly the
+// thing the scan is deliberately tolerant of and cannot be pinned by a portable test.
+//
+// So the check below is resolve-agnostic: it accepts EITHER "caller-chain=unknown" (no resolvable
+// frame) OR "caller-chain=<digits>" (a frame -- real or coincidental -- resolved), and requires
+// that whichever form printed, it is immediately followed by " alloc_dmem len=0x" -- i.e. that
+// *some* attribution record was published, naming the right allocator. That is the actual contract
+// of the call site under test; what the stack scan happens to resolve to is environmental.
+//
 // A DEDICATED BINARY, not an arm folded into tests/host/test_dmem.cpp or this repo's other dmem
-// coverage: DmemCallerChainInterner's "Unknown" state is a ONE-SHOT PER PROCESS (see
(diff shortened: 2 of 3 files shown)


mattias800 (commented):
**REJECTED** — the two new tests fail in CI, and the reason is that they assert ambient state rather
than the contract.

CI `Linux`:

```
178/313 Test #178: dmem_caller_attribution_alloc_dmem .............***Failed
FAIL: rc=0x0 phys=0x10000 restored=1 allocated=1 attributed=0
captured stderr (80 bytes):
[dmem-caller] caller-chain=1 alloc_dmem len=0x4000 from SaveData.prx+0x10000002
179/313 Test #179: dmem_caller_attribution_alloc_main_dmem ........***Failed
```

**Read the captured stderr — the feature worked.** The call site published a caller chain and
resolved it to `SaveData.prx+0x10000002`. What failed is the assertion, which expects the
`caller-chain=unknown` form. That form only appears when nothing is loaded that can resolve the
return address, which was true in your worktree and is not true on the CI runner.

So the arms currently bind to "no module happened to be resolvable here", not to "the call site
attributes its allocation". That is the same class of defect as the one this issue exists to fix:
#3055's arms bound to the formatter instead of the call site, and these bind to the environment
instead of the call site. The mutation you ran does prove they are load-bearing — but a test can be
load-bearing and still assert the wrong thing.

**What I want instead:** assert that attribution *happened*, independent of what it resolved to.
`attributed=1` with a non-empty chain is the contract; whether the chain prints `unknown` or a
module name is an environmental detail. If you specifically want to keep an `unknown` arm, it has to
force that state rather than hope for it.

**Keep everything else** — the diagnosis is right and the structure is good. Two separate ctest
binaries because `DmemCallerChainInterner`'s Unknown state is one-shot per process is exactly the
right call, and leaving #3055's formatter arms in place (they cover a real, different contract) is
also right. Only the assertion needs to change.

**Also worth noting for the record:** you were straight about not having a full ctest count, and
that honesty is why this got looked at properly rather than waved through on a number. The count on
a fresh no-dump configure is 314; CI reports 313 for its own configuration.

Re-push to the same branch when the arms assert the contract and I will re-review.

mattias800 (commented):
**APPROVED** (on the corrected head)

**Built and mutation-verified myself**, since you were straight about not having done either.

- Both targets build clean, and `ctest -R dmem_caller_attribution` passes 2/2.
- Mutation: injected an early `return {}` into `attribute_dmem_allocation`'s body, so the call sites
  still compile but publish nothing. Result: **0% passed, 2 tests failed**. Restored: 2/2 green. The
  arms are load-bearing.

A note on my own process, because it bears on how much that result is worth: my first two mutation
attempts (guarding the call sites with `if(false)`, then deleting them) both broke compilation, and
the tests then "passed" against a **stale binary**. Had I stopped there I would have wrongly told you
these arms were toothless. Only a mutation whose build is green says anything at all.

**Your root cause is the interesting part and I checked it.** `guest_va_in_module_code` tests whether
a value falls in an address *range*, never whether anything is mapped there — so an ordinary garbage
word on the CI runner's stack landed inside the `BOOT_SAVEDATA` range and the chain interned as
`Known`. Attribution genuinely happened; only my-assumption-shaped assertion failed. That is a real
property of the heuristic scan, not a CI quirk, and it would have bitten anyone who wrote this test.

**The corrected assertion is the right shape.** `dmem_caller_attribution_present(output, api)`
accepting either `caller-chain=unknown` or `caller-chain=<digits>`, while requiring the token be
immediately followed by ` <api> len=0x`, asserts the contract — attribution happened and named the
right allocator — rather than what the stack scan happened to resolve to. And it still catches
cross-contamination, since a mismatched label parses as false.

Everything else unchanged as asked: two ctest binaries (one process per allocator, because
`DmemCallerChainInterner`'s first-seen bookkeeping is per-process — the trap the issue's own review
comment flagged), and #3055's formatter arms left alone since they cover a real, different contract.

Issue #3067, opened 2026-08-27
kernel_sem_timedwait fails on the macOS (Rosetta 2) runner on a branch whose diff that job does not compile


mattias800:
Recording this rather than merging past it silently, because the pattern is odd and the next person
to see it should not have to re-derive that it is not their change.

## Observation

`kernel_sem_timedwait` (test 47 of 255) fails on the **macOS (x86_64 / Rosetta 2)** job on branch
`lane/gta5-perf` (PR #3063), twice — the original run and an explicit re-run of the failed job:

```
47/255 Test #47: kernel_sem_timedwait .......................***Failed    1.02 sec
47/255 Test #47: kernel_sem_timedwait .......................***Failed    1.22 sec
99% tests passed, 1 tests failed out of 255
```

Master's last **eight** CI runs pass the same job.

## Why it is not that branch's change

The branch's entire diff is one file, `prosper/frontends/shared/live/live_compute.cpp`, and **the
macOS job never compiles it**. Grepping that job's own build log: 332 `.cpp` compile lines, including
`frontends/shared/perf/performance_capture.cpp` and six `frontends/shared/tests/*`, and **zero**
matches for anything under `frontends/shared/live/`. That job configures with
`-DCMAKE_DISABLE_FIND_PACKAGE_Vulkan=TRUE`, so the whole live renderer is preprocessed out.

The macOS binary should therefore be byte-identical to master's, and a semaphore-timing test cannot
be reached by a counter added to a file that is not in it.

## So what is it

Unresolved. Two candidates, and I have not separated them:

- **A flaky timing test on a loaded shared runner.** `sem_timedwait` is inherently timing-sensitive,
  the failure takes ~1.0-1.2 s in both arms, and Rosetta 2 adds translation jitter. Two failures in a
  row is possible; eight master successes make it less comfortable.
- **Something environmental that changed recently** on the runner image, which master's eight runs
  happen to predate or happen not to hit.

What would separate them: re-run the macOS job on an unrelated PR branch, or on master, and see
whether the failure follows the branch or the calendar. One `gh run rerun` on any current PR answers
it, and I have not spent a CI cycle on that.

## Note for whoever hits this next

Do not assume it is your diff. Check whether the macOS job compiles the file you touched — the
Vulkan-disabled configuration excludes the entire live renderer, so a large class of changes is
provably absent from that binary.

Refs #3063

mattias800:
## Correction: "master's last eight runs pass the same job" was a stale comparison

[...]

PR #3179, opened 2026-09-01
hle: scePthreadSemInit forwards sem_init's real error instead of discarding it

### LINKED ISSUE #3068: scePthreadSemInit discards sem_init's return, and on macOS unnamed POSIX semaphores are ENOSYS — the whole scePthreadSem* family may be silently unsynchronised there
`prosper/src/hle/kernel/hle_kernel.cpp:3057`:

```cpp
HLE(k_sem_init) { if (!a0) return 0x16; auto* s = (sem_t*)calloc(1, sizeof(sem_t));
                  sem_init(s, 0, (unsigned)a2); *(void**)(uintptr_t)a0 = s; ... return 0; }
```

`sem_init`'s return value is discarded and the HLE returns `0` unconditionally.

## Why this is not merely untidy

**On Darwin, unnamed POSIX semaphores are not implemented.** `sem_init` there sets `errno = ENOSYS` and returns `-1`. So on macOS this function heap-allocates a `sem_t`, fails to initialise it, publishes the pointer to the guest anyway, and reports success. Every other member of the family — `k_sem_wait`, `k_sem_trywait`, `k_sem_timedwait`, `k_sem_post`, `k_sem_getvalue` — then operates on an uninitialised object, which is undefined behaviour, and the guest is told throughout that its synchronisation is working.

That is the exact failure class the function's own header comment was written to eliminate:

> Were MISSING -> the generic stub returned 0: SemInit created nothing, SemWait returned immediately (never blocked), SemGetvalue left `*value` uninitialized -> a producer/consumer or gate built on these got NO synchronization (the silent-unsync / UAF class).

The stub was replaced, and on one platform the same outcome is reachable through the real implementation.

**CONFIDENCE: MED on the consequence, HIGH on the code.** The discarded return and the unconditional `0` are plain in the source. That Darwin's `sem_init` is ENOSYS is well documented and long-standing, but I have not run it — this host is Windows, and prosper's macOS job is `x86_64 / Rosetta 2`, where `kernel_sem_timedwait` *does* return `ETIMEDOUT` rather than crashing. So either the wait on an uninitialised `sem_t` happens to time out, or Darwin's behaviour differs from my expectation. **Someone with a macOS host should check which**, because the two readings have very different consequences.

## How it was found

Independent review of #3066 (a test-only fix), as a non-blocking finding: an arm asserting `scePthreadSemInit reports success` cannot fail, because the implementation cannot report anything else. The reviewer noted the Darwin consequence.

## Suggested fix

Return the real error rather than inventing success, and do not publish a pointer to an object that was never initialised:

```cpp
HLE(k_sem_init) {
    if (!a0) return 0x16;
    auto* s = (sem_t*)calloc(1, sizeof(sem_t));
    if (!s) return /* ENOMEM as the guest reads it */;
    if (sem_init(s, 0, (unsigned)a2) != 0) { free(s); return sce_pthread_rc(fbsd_errno(errno)); }
    *(void**)(uintptr_t)a0 = s;
    ...
}
```

On macOS that turns a silent non-synchronisation into an honest error the guest can be traced failing on — which is the project's stated preference (fail visible, `CONFIDENCE: LOW` and log loudly, rather than a shim that fakes output). If titles actually need this family on macOS, the real fix is a `sem_open`-backed or condvar-backed implementation there, and that should be its own issue once someone confirms the behaviour.

## Related

- The regression test for the family is `prosper/tests/hle/kernel/test_kernel_sem_timedwait.cpp`. #3066 corrects a separate defect in it: the guest handle is a *pointer cell*, and the test declared it as `sem_t slot`, which is an 8-byte write into a 4-byte automatic on Darwin specifically (`sem_t` is `int` there). Both defects are macOS-specific and neither is visible from a Windows or Linux run.
- The family's registration and aliases are at `hle_kernel.cpp:3139-3152`.

mattias800:
## Fixes #3068

`scePthreadSemInit` (`k_sem_init`, `prosper/src/hle/kernel/hle_kernel.cpp`) discarded `sem_init()`'s
return value and unconditionally reported success, publishing a pointer to a `sem_t` through the
guest's handle slot even when initialisation failed. Every other member of the family —
`k_sem_wait`, `k_sem_trywait`, `k_sem_timedwait`, `k_sem_post`, `k_sem_getvalue` — resolves that
exact pointer back out of the slot via `ensure_sem()` and operates on it, so a failed init became
undefined behaviour on every later call instead of one honest, traceable failure at init time.

This is not theoretical: **unnamed POSIX semaphores are ENOSYS on Darwin**, so `sem_init()` there
fails unconditionally on every call, on every title that uses this family.

## UPDATE (after CI): the "no fallback exists" framing below was wrong

CI's macOS (Rosetta 2) job reddened the new test: the guest-visible `scePthreadSemInit` still
reported success for an out-of-range value, even though a direct probe of the real host `sem_init()`
correctly failed with `ENOSYS`. Root cause: `hle_kernel.cpp` includes
`src/host/platform/posix_shim.hpp`, which `#define`s `sem_t`/`sem_init`/... to a real mutex+cond
counting-semaphore implementation (`prosper_sem_t`/`prosper_sem_init`/...) on `__APPLE__`, because
Darwin's real `sem_init` is unconditionally ENOSYS. **That fallback already existed** and already
backs `SemWait`/`SemPost`/etc. for every title that hits this family on macOS today -- I hadn't found
it when I wrote the section below, which claims no fallback exists and that building one is out of
scope. That claim is wrong; a follow-up commit (`34acb13c`) fixes the actual gap: `prosper_sem_init`
unconditionally returned success regardless of `value`, ignoring it entirely, so my error-propagation
fix in `k_sem_init` was correctly written but structurally unreachable on macOS specifically -- the
call it makes could never fail. Worse, `count` is a plain `int`, so an out-of-range value was being
cast straight into it, silently wrapping negative. Fixed by adding the same `SEM_VALUE_MAX` bounds
check every other `sem_init` on this project's platforms already enforces (`value > INT_MAX` ->
`EINVAL`), touching nothing else in the shim. **Not built or tested by me -- still no macOS host**;
this is reasoned from reading the source, and the Rosetta 2 job is what actually verifies it.

## Fix strategy chosen, and why

**Propagate the real error; do not publish the slot on failure.** I did *not* build a
`sem_open`-backed or condvar-backed fallback for platforms lacking unnamed semaphores, which is the
other option the issue names. Reasons:

- Nobody has confirmed what a live macOS boot actually needs preserved from this family's semantics
  (initial-value handling, `getvalue` behaviour, destroy-while-waiting, etc.). Building an unverified
  emulation layer risks silently trading one wrong behaviour (undefined behaviour on an
  uninitialised object) for a different, more plausible-looking wrong one — exactly the shape the
  project charter singles out as worse than a loud failure.
- The issue's own suggested fix is "propagate the error... If titles actually need this family on
  macOS, the real fix is a `sem_open`-backed or condvar-backed implementation there, and that should
  be its own issue once someone confirms the behaviour." I followed that scoping rather than
  expanding it.

So the change: `k_sem_init` now checks `calloc`'s result (ENOMEM), checks `sem_init`'s result, and on
failure frees the allocation, logs (capped at 4 lines, matching the existing `__stack_chk_fail`
pattern) and returns the real FreeBSD-numbered errno **without writing anything into the guest's
slot**. Both existing aliases (`SCE_PTHREAD_ALIAS` for the Sony spelling, `POSIX_SEM_ALIAS` for the
POSIX one) already encode/convert whatever bare errno the body returns, so no caller-side change was
needed — the body already returned bare FreeBSD errno on every other path in this family.

## What I verified on Linux vs. what is reasoned-only for macOS

**Verified on Linux (this host):**
- `sem_init(&s, 0, 4294967295u)` genuinely fails here: `rc=-1 errno=22 (EINVAL)`, confirmed with a
  throwaway probe before writing the test. That's POSIX's own `SEM_VALUE_MAX` check (glibc:
  `SEM_VALUE_MAX == INT_MAX`), not a Linux quirk.
- With the fix: `scePthreadSemInit` given that value returns a non-zero, libkernel-encoded EINVAL,
  and the guest slot is left untouched (unpublished). Directly observed, all green.
- **With the fix reverted** (confirmed by grepping the file back to the original one-line body):
  the same test reddens — 3 of 4 substantive assertions fail (`init reports the real failure`,
  `encoded the libkernel way`, `slot left untouched` all FAIL; exit 1). Rebuilt and ran this myself
  before restoring the fix.
- Full local build: clean, no errors. Full `ctest --no-tests=error`: **315/315 passed, exit 0**
  (314 baseline + this PR's new test), run against this PR's exact final commit on `origin/master`.

**Reasoned only, NOT verified — I have no macOS host:**
- That Darwin's `sem_init()` is unconditionally `ENOSYS` for unnamed semaphores (this is the issue's
  own claim, and it is Apple's long-documented behaviour, but I have not run it).
- That the fix's Darwin behaviour (a clean, encoded `ENOSYS`/errno instead of silent UB) is what
  actually happens on the Rosetta 2 CI runner. The test's Darwin path is designed to exercise the
  *same* probe-then-assert logic that passed on Linux (any out-of-range value should fail there too,
  and ENOSYS should fail unconditionally), but I have not seen it run. **Please read the macOS CI job
  on this PR as the actual verification, not this description.**

## Test

New `prosper/tests/hle/kernel/test_kernel_sem_init_error.cpp`, registered in `CMakeLists.txt`
next to `test_kernel_sem_timedwait`. It does not hardcode "sem_init fails here": it probes the
**host** `sem_init()` with an out-of-range value first, and only asserts the guest-visible failure
contract when the host call actually failed — so it explains itself and skips cleanly rather than
passing vacuously if some platform's libc happens to accept the probe value, and it is meaningful on
Linux (EINVAL, confirmed) and Darwin (ENOSYS on any value) without platform-specific code.

Also updated a stale comment in `test_kernel_sem_timedwait.cpp` that explicitly described its
`scePthreadSemInit reports success` check as "kept as a tripwire... the fix is #3068" — that
assertion now exercises a real forwarded result rather than an unconditional 0, and the comment
pointed at the new dedicated failure-path test instead of continuing to describe pre-fix behaviour.

## Build / test evidence (personally observed)

- `cmake --build` (Linux, distrobox `ps5ys`): clean, exit 0, at this PR's final commit.
- `ctest --no-tests=error`: **315/315 passed**, exit 0, at this PR's final commit
  (`f12609bf49926e7cc0dbd6a7b9274bd4566ed8d3` at push time).
- Mutation check (fix reverted, fix restored): described above, directly observed both states.

## Note on how this PR's branch was assembled

Mid-session, a `git stash pop` in this worktree picked up a **different concurrent agent's** stash
entry instead of mine (the stash ref is repo-wide, not worktree-scoped — filed as #3174). I did not
discard it: I re-stashed it intact with a note for its owner, never dropped anything from the shared
stack, and re-applied my own `hle_kernel.cpp` fix directly from my own record of the edit rather than
continuing to search the shared stash stack. `git diff --stat` on this branch is exactly the 4 files
listed above; nothing from that collision is in this PR.

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -680,6 +680,16 @@ if(TARGET prosper_core)
   # regression that reverts to a blocking wait on a semaphore nothing will ever post would hang
   # rather than fail, and CTest's default 1500 s hang reads as neither pass nor fail.
   set_tests_properties(kernel_sem_timedwait PROPERTIES TIMEOUT 60)
+
+  # scePthreadSemInit must forward a real sem_init() failure instead of discarding it and
+  # manufacturing success (#3068). Platform-neutral: it probes the host sem_init() with an
+  # out-of-range value first and only asserts the guest-visible contract when the host call
+  # actually failed, so it is meaningful on every CI platform (EINVAL on Linux, ENOSYS on Darwin)
+  # without needing per-platform code, and skips itself explicitly rather than passing vacuously
+  # if some future libc happens to accept the probe value.
+  add_executable(test_kernel_sem_init_error tests/hle/kernel/test_kernel_sem_init_error.cpp)
+  target_link_libraries(test_kernel_sem_init_error PRIVATE prosper_core)
+  add_test(NAME kernel_sem_init_error COMMAND test_kernel_sem_init_error)
   # Bounded because arm 3 guards a ~292-YEAR sleep: without a timeout a regression there presents as
   # CTest hanging for its default 1500 s rather than as a failure, and a hang is the one outcome nobody
   # reads as a red test. Generous against a loaded host -- what it catches is seconds-to-forever.
--- a/prosper/src/hle/kernel/hle_kernel.cpp
+++ b/prosper/src/hle/kernel/hle_kernel.cpp
@@ -3073,7 +3073,42 @@ SCE_PTHREAD_ALIAS(k_sce_rwlock_timedwrlock, k_rwlock_timedwrlock)
 // SemGetvalue left *value uninitialized -> a producer/consumer or gate built on these got NO
 // synchronization (the silent-unsync / UAF class). Back them with host sem_t.
 namespace { bool semlog() { static const bool on = getenv("PROSPER_SEMLOG") != nullptr; return on; } }
-HLE(k_sem_init)      { if (!a0) return 0x16; auto* s = (sem_t*)calloc(1, sizeof(sem_t)); sem_init(s, 0, (unsigned)a2); *(void**)(uintptr_t)a0 = s; if (semlog()) fprintf(stderr, "[sem] init slot=%p value=%u\n", (void*)(uintptr_t)a0, (unsigned)a2); return 0; }
+// #3068: sem_init() CAN fail -- most notably, Darwin does not implement unnamed POSIX semaphores at
+// all, so sem_init() there sets errno=ENOSYS and returns -1 unconditionally. This used to discard
+// that return value outright and report success no matter what, publishing a pointer to a `sem_t`
+// that was never actually initialised. Every other member of the family --  `k_sem_wait`,
+// `k_sem_trywait`, `k_sem_timedwait`, `k_sem_post`, `k_sem_getvalue` -- resolves its host `sem_t`
+// straight out of the slot via `ensure_sem`, so a published-but-uninitialised object turned one
+// honest, traceable failure here into undefined behaviour on every later call instead, with the
+// guest told throughout that its synchronisation was working.
+//
+// Fixed by PROPAGATING the real error rather than manufacturing success, and by refusing to publish
+// the slot when init failed -- the fail-visible choice the charter prefers over a shim that fakes
+// output. A `sem_open`-backed or condvar-backed fallback for platforms lacking unnamed semaphores is
+// deliberately NOT attempted here: nobody has verified what such an emulation needs to preserve on a
+// live macOS boot, and an unverified fallback risks trading one silently-wrong behaviour for
+// another. If a title is later found to depend on this family on such a platform, that is its own
+// issue once the real behaviour is confirmed (see #3068's discussion).
+HLE(k_sem_init) {
+    if (!a0) return 0x16;
+    auto* s = (sem_t*)calloc(1, sizeof(sem_t));
+    if (!s) return 12;   // ENOMEM (bare FreeBSD errno -- encoded/converted by the aliases below)
+    if (sem_init(s, 0, (unsigned)a2) != 0) {
+        const uint64_t rc = fbsd_errno(errno);
+        // Loud but bounded: on a platform where this fails unconditionally (Darwin), every single
+        // semaphore init would otherwise flood stderr once per call. Capped the same way the
+        // __stack_chk_fail diagnostic above is (n++ < 4): visible immediately, not a firehose.
+        static int logged = 0;
+        if (logged++ < 4)
+            fprintf(stderr, "[sem] init FAILED slot=%p value=%u host_errno=%d(%s)\n",
+                    (void*)(uintptr_t)a0, (unsigned)a2, errno, strerror(errno));
+        free(s);
+        return rc;
+    }
+    *(void**)(uintptr_t)a0 = s;
+    if (semlog()) fprintf(stderr, "[sem] init slot=%p value=%u\n", (void*)(uintptr_t)a0, (unsigned)a2);
+    return 0;
+}
 HLE(k_sem_wait)      { auto* s = ensure_sem(a0); if (!s) return 0x16; if (semlog()) fprintf(stderr, "[sem] wait slot=%p enter\n", (void*)(uintptr_t)a0); int rc = sem_wait(s); if (semlog()) fprintf(stderr, "[sem] wait slot=%p exit rc=%d\n", (void*)(uintptr_t)a0, rc); return rc == 0 ? 0 : fbsd_errno(errno); }
 HLE(k_sem_trywait)   { auto* s = ensure_sem(a0); if (!s) return 0x16; return sem_trywait(s) == 0 ? 0 : fbsd_errno(errno); }   // EAGAIN (would block) is 11 here, 35 on the PS5
 // scePthreadSemTimedwait is the one member of the family with no POSIX spelling registered on its
--- a/prosper/src/host/platform/posix_shim.hpp
+++ b/prosper/src/host/platform/posix_shim.hpp
@@ -69,6 +69,7 @@ static inline int prosper_mincore(const void* addr, size_t len, unsigned char* v
 #include <sys/uio.h>
 #include <unistd.h>
 #include <cerrno>
+#include <climits>
 #include <cstdio>
 #include <ctime>
 #include <csignal>
@@ -167,6 +168,15 @@ struct prosper_sem_t {
     int             count;
 };
 static inline int prosper_sem_init(prosper_sem_t* s, int /*pshared*/, unsigned value) {
+    // POSIX: sem_init() must fail EINVAL when `value` exceeds {SEM_VALUE_MAX} (sem_init(3)). `count`
+    // is a plain `int` here, so INT_MAX is the real bound this representation can hold -- accepting
+    // more would silently truncate the cast below into a negative count, and every subsequent
+    // wait/trywait/timedwait would then see "not ready" (`count <= 0`) even after enough posts to
+    // reach the value the guest actually asked for. Found in review of #3068: the out-of-range probe
+    // that test proved failed on Linux (real sem_init's own SEM_VALUE_MAX check) fell straight
+    // through this shim's unconditional `return 0` on Darwin, so the one platform this whole shim
+    // exists for was the one platform where SemInit could never report a real failure.
+    if (value > (unsigned)INT_MAX) { errno = EINVAL; return -1; }
     pthread_mutex_init(&s->m, nullptr); pthread_cond_init(&s->c, nullptr);
     s->count = (int)value; return 0;
 }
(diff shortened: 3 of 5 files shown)


mattias800 (commented):
**REJECTED** — the fix does not hold on macOS, which is the platform this issue is about.

CI `macOS (x86_64 / Rosetta 2)`, and note it is the **new** test failing, not the pre-existing
`kernel_sem_timedwait` flake tracked in #3067:

```
50/259 Test #50: kernel_sem_init_error ......................***Failed    1.40 sec
== test_kernel_sem_init_error ==
  [ok]   scePthreadSemInit is registered
         (host sem_init(4294967295) -> rc=-1 errno=78)
  [FAIL] scePthreadSemInit reports the real failure instead of manufactured success
  [FAIL] the failure is encoded the libkernel way (0x8002_0000 | errno), not a bare number
  [FAIL] the guest slot is left untouched, not published with a pointer to storage that was never...
```

**Read the probe line: errno 78 on Darwin is ENOSYS.** Your self-invalidating probe worked exactly
as designed — it confirmed the host `sem_init` genuinely fails there. And then all three substantive
assertions failed anyway, which means the guest call still reported success. So on the one platform
where unnamed POSIX semaphores are unavailable — the entire premise of #3068 — `scePthreadSemInit`
is still manufacturing success.

That is a much more useful result than a green run would have been: the test is doing its job and
the fix is not yet doing its.

**Work out why before changing anything.** The obvious candidates:

- `k_sem_init` may not be the path macOS actually takes — check for an `#ifdef __APPLE__` or a
  platform-specific semaphore implementation that bypasses the call site you fixed.
- The error check may be present but reached with a value that succeeds, while the *guest's* own
  call uses a different one. Note ENOSYS is unconditional on Darwin, so a valid count fails there
  too — which is not true on Linux and may be the assumption that broke.

Establish which it is and say so; the answer changes the fix.

**Your framing was right and I want to keep it.** Propagating the real error rather than inventing a
`sem_open`/condvar fallback is the correct call for now — nobody has confirmed what a live macOS
boot needs from this family, and a plausible-but-unverified emulation would trade loud UB for a
quieter wrong answer. Keep that decision.

**Credit where due:** you explicitly labelled the macOS behaviour as reasoned-not-executed and said
the Rosetta job was the real verification rather than your description. That is exactly right, and
it is why this landed as a clear signal instead of an argument. Also good that you re-stashed the
other agent's entry intact rather than discarding it (#3174).

Do not build — push and I will verify. Re-push to the same branch and I will re-review.

mattias800:
Root cause identified — it's candidate 1 from your message, with a precise mechanism.

**Not** a separate `k_sem_init` code path. My call site in `hle_kernel.cpp` *is* the one that
runs on macOS. The bypass is one layer down: `prosper/src/host/platform/posix_shim.hpp` `#define`s
`sem_t`/`sem_init`/`sem_wait`/... to `prosper_sem_t`/`prosper_sem_init`/... on `__APPLE__` — a real
mutex+cond counting-semaphore implementation, because Darwin's real `sem_init` is unconditionally
`ENOSYS` for unnamed semaphores. `hle_kernel.cpp` includes that header, so every `sem_init(...)` call
in the file I edited is actually `prosper_sem_init(...)` on macOS. That existing shim's `init`
**unconditionally returned 0**, ignoring `value` entirely — so my error-propagation branch in
`k_sem_init` was correctly written but structurally unreachable there: the thing it calls could
never fail.

This also means my PR's original framing was wrong in one respect: I said no macOS fallback existed
and building one unverified was too risky to attempt in this PR. That's backwards — **the fallback
already existed** (and, per its own header, already backs `SemWait`/`SemPost`/etc. for every title
that hits this family on macOS today). I just hadn't found it before writing the description. The
actual gap was much smaller than "build a fallback": `prosper_sem_init` was missing the standard
POSIX `SEM_VALUE_MAX` bounds check every other `sem_init` on this project's platforms already
enforces. Worse than merely unreachable — `count` is a plain `int`, so an out-of-range `value` was
being cast straight into it, silently wrapping negative and corrupting the semaphore's accounting.

Fix pushed (`34acb13c`): `prosper_sem_init` now rejects `value > INT_MAX` with `EINVAL`, matching
`sem_init(3)`'s documented contract. Nothing else in the shim changed — `Wait`/`Trywait`/
`Timedwait`/`Post`/`Getvalue`/`Destroy` are untouched, so real synchronization for any previously-valid
value is unaffected.

**Not built or tested by me** — still no macOS host. This is reasoned from reading the macro
substitution and the shim's source, and is exactly the gap the failing Rosetta 2 run identified. I'm
relying on that CI job again to actually verify it, not asserting it here.

mattias800 (commented):
**APPROVED** (on the corrected head)

**The macOS job is now green** — the same `macOS (x86_64 / Rosetta 2)` job that failed
`kernel_sem_init_error` on the previous head. That is the discriminator for this whole PR, and it
flipped. Nothing else could have established it from a Linux box.

**Your diagnosis is right and it corrects your own earlier claim, which is the valuable part.** The
first version of this PR said no macOS fallback existed and building one was out of scope. In fact
`hle_kernel.cpp` includes `posix_shim.hpp`, which on `__APPLE__` `#define`s `sem_init` to
`prosper_sem_init` — a real mutex+cond counting semaphore that already backs the entire family
there, precisely because Darwin's `sem_init` is unconditionally ENOSYS. So every `sem_init(...)` in
the file you edited was already something else on that platform, and your error-propagation branch
was correctly written but **structurally unreachable** on the one platform #3068 is about.

That is a much better answer than either hypothesis I gave you, and you reached it by reading the
macro substitution rather than by guessing between my two options.

**The real defect turned out smaller and worse than described.** `prosper_sem_init` never validated
`value` against `SEM_VALUE_MAX` and silently truncated an out-of-range value into a signed `int
count` — so the sign wraps negative and the semaphore is corrupted, rather than merely
mis-initialised. Rejecting `value > INT_MAX` with `EINVAL` matches `sem_init(3)`'s documented
contract and leaves Wait/Trywait/Timedwait/Post/Getvalue/Destroy untouched. Minimal and scoped.

**Framing preserved:** still propagating the real error rather than inventing new fallback
machinery, which remains right — nobody has confirmed what a live macOS boot needs from this family.

**And you kept the reporting discipline throughout**: you flagged the second fix as reasoned-only
because you were told not to build, and named the Rosetta job as the thing that would verify it. It
did.

Issue #3180, opened 2026-09-01
render_runner: several vkCreateImage sites discard VkResult (checked, but not logged)


mattias800:
`tests/fixtures/render_runner.h`'s `render_draw_pass_rgba` has several more `vkCreateImage` call
sites that discard the `VkResult`, found while fixing #3045 (which covered the sampled-texture
upload path and, once noticed to have the identical no-guard-at-all shape, the depth/stencil
target path too).

These remaining sites are milder than #3045's: each already guards the *handle* before using it
further (`if (!img) { ...; return out; }`), so they do not call a Vulkan function on a
definitely-null image the way #3045's two sites did. But they still discard the `VkResult`, which
means they rely on undefined driver behaviour (a failed `vkCreateImage` leaving `pImage` at
`VK_NULL_HANDLE`) rather than checking the documented return code. In practice every driver does
this, but it is not something the Vulkan spec guarantees.

Sites (current line numbers as of this filing; #3045's fix already shifted a few of these down by
~30-45 lines from what an older diff would show):

- `vkCreateImage(dev, &imgci, nullptr, &img);` — primary color-target (slot 0) creation
- `vkCreateImage(dev, &imgci, nullptr, &img);` — slot-0 fallback after a persistent-cache budget miss
- `vkCreateImage(dev, &color1_ci, nullptr, &img1);` — color-target slot 1 creation
- `vkCreateImage(dev, &color1_ci, nullptr, &img1);` — slot-1 fallback after a budget miss
- `vkCreateImage(dev, &ci, nullptr, &extra_images[slot]);` — MRT slots 2+ creation
- `vkCreateImage(dev, &ci, nullptr, &extra_images[slot]);` — MRT slot fallback after a budget miss

Suggested fix: capture the `VkResult` at each site and log it (rate-limited, matching the
`[texture-upload-failed]` / `[ds-create-failed]` convention #3045 established) instead of relying
on the null-handle convention alone. Low priority relative to #3045 since there is no UB today —
mainly a robustness/diagnosability improvement (a rate-limited log naming the real Vulkan error
code beats inferring "something failed" from a null handle).

Found by review while fixing #3045.

mattias800:
CLAIMING: lane-c | branch fix/issue-3180-vkcreateimage-result | 2026-09-02T06:12:34Z

PR #3181, opened 2026-09-01
render_runner: check vkCreateImage's result before using the image handle

### LINKED ISSUE #3045: render_runner: vkCreateImage result is discarded, then the null handle is used
`prosper/tests/fixtures/render_runner.h:6726` calls `vkCreateImage(dev, &tci, nullptr, &upload.image)` and discards the `VkResult`. Line `:6728` then calls `vkGetImageMemoryRequirements(dev, upload.image, &tr)` unconditionally.

**Failure scenario:** any `vkCreateImage` failure — `VK_ERROR_OUT_OF_DEVICE_MEMORY`, or an image exceeding a device limit such as `maxImageArrayLayers` or `maxImageDimension2D` — leaves `upload.image` as `VK_NULL_HANDLE`, which is then passed to `vkGetImageMemoryRequirements`. That is undefined behaviour and in practice a crash or garbage memory requirements followed by a bogus allocation.

**Pre-existing and not array-specific:** this is the path every sampled texture takes. It is filed now because review of #3043 identified it while checking that the new 2D_ARRAY arm could not reach it with an absurd `arrayLayers`. #3043 bounds its own contribution at `kBackendMaxArrayLayers = 2048`, so it is unlikely to be the trigger. **Correction to the original wording here:** that constant is a pragmatic ceiling, not a portability guarantee. Vulkan Core's Required Limits give `maxImageArrayLayers >= 256`; 2048 is the Roadmap 2022 / Vulkan 1.4 figure, and the backend requests `VK_API_VERSION_1_1`. So a device reporting the Core minimum can still reject an image #3043 admits -- through this same unchecked path. Querying the real limit belongs with this issue — but the unchecked call remains for every other shape.

**Suggested fix:** check the `VkResult`, and on failure leave the resource unmaterialized and fail visibly (the surrounding code already has skip paths for resources it cannot upload). Separately, prosper reads no `VkPhysicalDeviceLimits` for image creation at all; deriving the real `maxImageArrayLayers` / `maxImageDimension2D` would let the backend reject at the contract boundary rather than at the driver.

Found by review of #3043.

mattias800:
## Summary

Fixes #3045: `render_draw_pass_rgba`'s sampled-texture upload path called
`vkCreateImage(dev, &tci, nullptr, &upload.image)` and discarded the `VkResult`, then
unconditionally called `vkGetImageMemoryRequirements(dev, upload.image, &tr)`. On any real
`vkCreateImage` failure (`VK_ERROR_OUT_OF_DEVICE_MEMORY`, or a device rejecting the requested
extent/array-layer count), `upload.image` stayed `VK_NULL_HANDLE` and was passed straight into
that call -- undefined behaviour.

While auditing this function's other `vkCreateImage` sites for the same defect, I found a second
one with the identical shape: the depth/stencil target creation path had **no guard at all**
(not even the weaker `if (!img)` handle check the color-target sites use) before calling
`vkGetImageMemoryRequirements(dev, dimg, &dr)`. Fixed both.

## Failure scenario

Any `vkCreateImage` failure on the sampled-texture or depth/stencil target path fed a
`VK_NULL_HANDLE` into `vkGetImageMemoryRequirements` and onward. Verified this is not
theoretical: with the fix removed and a real failure forced (see Verification), the process
segfaults inside the Vulkan call, on this box's RADV driver.

## Fix

Both sites now check the `VkResult` (and the handle, matching the sibling color-target
convention), log a rate-limited diagnostic naming the real Vulkan error code and the requested
image parameters, and fail visibly instead of propagating the null handle:

- **Sampled-texture upload** (`tests/fixtures/render_runner.h`, the site #3045 names): on
  failure, sets `buffer_resources_ready = false; break;` -- the exact mechanism this function
  already uses for a fatal storage-buffer upload failure, which the per-draw loop turns into
  `continue` (skip this one draw; the render pass still executes for every other draw and still
  produces its clear).
- **Depth/stencil target creation**: on failure, `return out;` -- matching the sibling
  color-target/MRT creation sites in this same function, which already bail this way on a
  (weaker, handle-only) creation failure.

### Deliberately out of scope

The color-target and MRT creation sites (`img`, `img1`, `extra_images[slot]`, four call sites)
also discard the `VkResult`, but each already checks `!img` before using the handle further, so
there is no UB today -- just reliance on the (universally true in practice, but spec-unguaranteed)
convention that a failed `vkCreateImage` leaves the output handle null. Filed as #3180 rather than
folding into this PR, since it's a different-severity issue (robustness/diagnosability, not UB)
and touches four more call sites.

Also out of scope: `tests/fixtures/render_runner.h:426-437`'s comment (added by #3043, which is
what surfaced this issue) notes that prosper reads no `VkPhysicalDeviceLimits` for image creation
at all, and that deriving the real `maxImageArrayLayers`/`maxImageDimension2D` would let the
backend reject at the contract boundary rather than at the driver. That's a real improvement but
a materially bigger change (threading device limits through); this PR is scoped to the reported
defect -- checking the result and failing visibly.

## Verification

Added `inject_render_texture_create_failure_once()` (mirrors the existing
`inject_render_buffer_upload_failure_once` pattern in the same file) -- a deterministic one-shot
injection that simulates the *next* texture-upload `vkCreateImage` call failing, without depending
on any real device's limits. A real over-large request can't be relied on to fail on every
device: this box's RADV reports `maxImageArrayLayers=8192`, well above the backend's own
`kBackendMaxArrayLayers` ceiling of 2048 from #3043, so an over-large-arrayLayers request never
actually reaches the driver here.

New regression check in `tests/gpu/test_texture_sample_render.cpp`: arm the injection, render the
same textured triangle the file's existing GREEN-texel check uses, and assert the pass still
completes and shows the blue clear (not the GREEN texel, not garbage, not a crash) -- proving the
draw was skipped rather than the null handle being used.

**Proved the test fails without the fix:** temporarily reverted just the
check-and-bail block (kept the injection call itself, since that's test scaffolding, not the fix)
back to the original unconditional `vkCreateImage(...); vkGetImageMemoryRequirements(dev,
upload.image, &tr);`, rebuilt, and ran `test_texture_sample_render`:

```
[render] Vulkan device: AMD Radeon 8060S Graphics (RADV STRIX_HALO) (integrated GPU)
== test_texture_sample_render ==
  ...
  [ok]   sampling texel (1,0) yields GREEN (proves u routing)
Segmentation fault (core dumped)
```

(stdout is fully buffered when redirected to a file, so later `[ok]`/`[FAIL]` lines never reached
disk; a `stdbuf -oL` rerun confirmed the crash lands immediately after that last printed line --
i.e. inside my new check block, calling `vkGetImageMemoryRequirements` on the injected-failure's
null handle.) Restored the fix, rebuilt, reran: `== PASS ==`, exit 0, the new check passes.

## Build / test

```
cmake --build prosper/build-linux -j6      # clean, both before and after merging master
ctest --no-tests=error                     # 314/314 passed, exit 0 (before and after merging master)
python3 tools/output/check_ascii_output.py prosper   # all checks passed
```

314 is the unchanged full count -- the new regression check lives inside the existing
`test_texture_sample_render` binary rather than registering a new ctest case.

## Follow-up filed

#3180 -- the four milder sibling `vkCreateImage` sites (color-target/MRT creation) that discard
`VkResult` but already guard the handle.

### DIFF
--- a/prosper/tests/fixtures/render_runner.h
+++ b/prosper/tests/fixtures/render_runner.h
@@ -2650,6 +2650,28 @@ inline const char* render_buffer_upload_failure_name(RenderBufferUploadFailureSt
     }
 }
 
+// Deterministic one-shot injection for the #3045 regression check: simulate the NEXT sampled-
+// texture vkCreateImage call failing, without depending on any real device's limits. A real
+// over-large arrayLayers request cannot be relied on to fail here -- this box's RADV reports
+// maxImageArrayLayers=8192, well above the backend's own kBackendMaxArrayLayers ceiling of 2048,
+// so the array-layer path never actually reaches the driver's own rejection. The production path
+// never arms this state.
+inline bool& render_texture_create_failure_once_storage() {
+    static thread_local bool armed = false;
+    return armed;
+}
+
+inline void inject_render_texture_create_failure_once() {
+    render_texture_create_failure_once_storage() = true;
+}
+
+inline bool consume_render_texture_create_failure_once() {
+    bool& armed = render_texture_create_failure_once_storage();
+    if (!armed) return false;
+    armed = false;
+    return true;
+}
+
 // Create, populate, and unmap one transient storage buffer. Every partial failure tears down in
 // Vulkan lifetime order (buffer before any bound memory) and leaves both outputs null. Returning the
 // exact failed step lets the caller report the binding before substituting a safe zero-word buffer.
@@ -5205,7 +5227,24 @@ inline std::vector<uint8_t> render_draw_pass_rgba(std::span<const BackendDraw> d
         dci.usage = VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT |
                     VK_IMAGE_USAGE_TRANSFER_SRC_BIT | VK_IMAGE_USAGE_TRANSFER_DST_BIT |
                     VK_IMAGE_USAGE_SAMPLED_BIT;   // sampled depth bridge (#1275)
-        vkCreateImage(dev, &dci, nullptr, &dimg);
+        // #3045: same class of defect as the sampled-texture upload path, and found while
+        // auditing this function's other vkCreateImage sites for the same review. Unlike the
+        // color-target sites above (which discard the VkResult but do check `!img`/`!img1`/
+        // `!extra_images[slot]` before touching the handle further), this one had NO guard at
+        // all -- a real failure fed a definitely-null `dimg` straight into
+        // vkGetImageMemoryRequirements, the exact undefined behaviour #3045 reports. Match the
+        // sibling return-out-on-failure convention instead.
+        const VkResult ds_image_result = vkCreateImage(dev, &dci, nullptr, &dimg);
+        if (ds_image_result != VK_SUCCESS || !dimg) {
+            dimg = VK_NULL_HANDLE;
+            static std::atomic<uint32_t> ds_create_failure_logs{0};
+            if (ds_create_failure_logs.fetch_add(1, std::memory_order_relaxed) < 32)
+                std::fprintf(stderr,
+                    "[ds-create-failed] vkCreateImage result=%d extent=%ux%u fmt=%d\n",
+                    (int)ds_image_result, W, H, (int)DFMT);
+            if (cached_ds) persistent_ds_cache().erase(ds_key);
+            return out;
+        }
         VkMemoryRequirements dr; vkGetImageMemoryRequirements(dev, dimg, &dr);
         VkMemoryAllocateInfo dai{VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_INFO};
         dai.allocationSize = dr.size; dai.memoryTypeIndex = pick(dr.memoryTypeBits, 0);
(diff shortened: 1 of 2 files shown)


mattias800 (commented):
**APPROVED**

**Mutation verified independently.** Disabling the new guard at `render_runner.h:7009`
(`if (image_create_result != VK_SUCCESS || !upload.image)` → `if (false)`) makes
`test_texture_sample_render` **segfault (rc=139)**; restored, it exits 0. So the regression arm is
load-bearing, and it fails in exactly the way the issue describes rather than by an assertion the
test could have satisfied some other way — the null `VkImage` really does propagate into a later
Vulkan call.

**The defect class is the interesting part.** `vkCreateImage` returning non-`VK_SUCCESS` is not a
hypothetical: `VK_ERROR_OUT_OF_DEVICE_MEMORY` and format/extent limit rejections are ordinary on a
loaded system, and the previous code stored the uninitialised handle and carried on. That turns a
recoverable, *reportable* condition into a crash somewhere else entirely — the worst possible
translation, because the stack trace names a function that did nothing wrong.

**Forcing the failure without depending on a real device's limits is the right test design.** A test
that provoked a genuine OOM would be non-deterministic and machine-specific; injecting the failure
means the arm asserts the *handling*, which is the thing under review.

**Auditing the other `vkCreateImage` sites in the same function rather than fixing only the reported
one** is what I would want here — a defect found by inspection at one call site is nearly always
present at its siblings, and finding them now costs nothing while finding them later costs another
crash report.

**The diagnostic is rate-limited (32) and prints extent and format**, which is the right shape: an
unbounded `fprintf` in a per-texture path is its own defect, and the two fields printed are exactly
the ones that identify a limit rejection.

Verified: no `Claude-Session:` trailer, no absolute host paths, CI green.

Issue #1770, opened 2026-08-02
videoout_queries: the WaitVblank PHASE assertion is a coin flip on the macOS/Rosetta runner


mattias800:
`videoout_queries` fails intermittently on the **macOS (x86_64 / Rosetta 2)** runner, which is a
required check, so it randomly blocks unrelated PRs.

## Proof it is the test, not the change

On branch `exp/issue-1226-init-suppression` (#1767) the job failed at head `4f8b0227`, whose **entire
diff against the previous, passing head `9f31922d` is two `.md` files**:

```text
prosper/docs/GAME_COMPAT_ORCHESTRATION.md
prosper/tools/AGENTS.md
```

Identical binaries, opposite outcomes. Observed 2 failures in 4 runs of the same test on that
branch, always the same assertion, always only on the x86_64/Rosetta runner (the `macOS App
(SDL3 + MoltenVK)` job, on native arm64, passed every time).

## The failing assertion

`prosper/tests/test_videoout.cpp:301`

```text
[FAIL] WaitVblank shares GetVblankStatus's PHASE, not just its period
       (wakes land on a boundary of the same grid)
```

```cpp
uint64_t min_phase_ns = UINT64_MAX;
for (int i = 0; i < 6; ++i) {
    waitvbl(handle, 0, 0, 0, 0, 0);
    vbl(handle, (uint64_t)(uintptr_t)vb, 0, 0, 0, 0);
    const uint64_t phase_ns = (*(uint64_t*)(vb + 0x08) * 1000ull) % 16683350ull;
    if (phase_ns < min_phase_ns) min_phase_ns = phase_ns;
}
CHECK(min_phase_ns < 4000000ull, …);
```

## Why the robustness argument does not hold

The comment at `test_videoout.cpp:288-293` justifies the minimum-of-six statistic like this:

> sleep_until may overshoot under host load, which can only move a residue later, so the minimum is
> the robust statistic

**The residue is taken modulo the period**, so "later" wraps. An overshoot below one period biases
the residue upward as the comment says, but an overshoot comparable to or larger than 16.68 ms —
entirely plausible for `sleep_until` on an emulated-x86 runner under load — maps the residue
*uniformly* onto the period rather than pushing it toward the top. The minimum then stops being a
floor on a shifted epoch and becomes the minimum of several near-uniform draws.

That predicts the observed rate. The acceptance window is `4 / 16.683 = 24%` per sample, so with six
independent samples:

```text
[...]

PR #3182, opened 2026-09-01
host: give Win32SleepFallback the same deadline post-condition as its sibling (#3074)

### LINKED ISSUE #3074: precise_sleep's Win32SleepFallback can return before its deadline — the high-resolution path re-checks the clock and the fallback does not

mattias800:
## Issue

Fixes #3074.

`precise_sleep`'s `sleep_until_steady_ns()` on Windows has two backends: the high-resolution
waitable-timer path and, when the timer is unavailable, a fallback that goes through
`std::this_thread::sleep_until()`. The high-resolution path loops and re-reads the clock after
every timer signal, explicitly because a waitable timer can be released marginally early and the
function's contract ("do not come back before the deadline") has to be enforced rather than
trusted to the primitive's precision. `Win32SleepFallback` did not do the same: it called the one
OS sleep primitive exactly once and returned, so it could come back before `deadline_ns` while the
sibling path (and the header's own doc comment) both promise it cannot.

## Behavioral contract

`sleep_until_steady_ns(deadline_ns)` must never return before
`steady_clock::now() >= deadline_ns`, on every backend it can select. That is the property every
caller relies on -- the vblank kevent pump (`hle_kernel_time.cpp`, after #3064) and the audio grain
pacer (`audio_sink_sdl3.cpp`) both assume it, and an early return from the fallback can post a
duplicate vblank kevent for the same tick or deliver an audio grain early.

## Approach

The fix (`prosper/src/host/platform/precise_sleep.cpp`, `.hpp`) extracts the re-check-and-retry
loop into a new, platform-neutral function:

```cpp
void sleep_until_deadline_retry(uint64_t deadline_ns,
                                 uint64_t (*now_ns)(),
                                 void (*sleep_once)(uint64_t deadline_ns));
```

It is compiled and linked unconditionally (outside every `#ifdef _WIN32` block), and takes the
clock read and the one-shot sleep as function pointers rather than calling
`steady_now_ns()`/`sleep_until_with_std()` directly. `Win32SleepFallback` now calls
`sleep_until_deadline_retry(deadline_ns, steady_now_ns, sleep_until_with_std)` instead of calling
`sleep_until_with_std()` once and returning. The high-resolution timer path is unchanged -- it
already re-checks correctly and this PR does not touch it. The POSIX path is also unchanged
(deliberately a single call; see the existing comment in the file).

Making the retry loop a pure function taking function pointers is what makes this testable off
Windows at all: `tests/host/platform/test_precise_sleep.cpp` supplies a fake clock and a fake
"sleep" that just advances it by a fixed step, and asserts the loop's **termination condition**
(never a wall-clock duration, which would flake under host scheduling -- this project has been
burned by that twice already, per the comments in `test_videoout.cpp` on #1770/#1793). This follows
the same pure-function-over-real-time approach #3170 used for `grid_boundaries_missed()`.

Test arms:
1. The #3074 case itself: each fake sleep undershoots the deadline (mirroring what `::Sleep()`'s
   tick quantization can do), so a single call cannot satisfy the post-condition -- the loop must
   retry. Asserts the post-condition holds and that it took more than one call.
2. A sleep that reaches the deadline in exactly one call: no extra retry.
3. A sleep that overshoots the deadline in one call: same, no extra retry.
4. A deadline already in the past on entry: zero sleep calls.

## What is verified vs. reasoned

- **Verified, on Linux:** `sleep_until_deadline_retry()` itself -- its loop, its termination
  condition, and every test arm above -- builds and runs correctly on this platform, since it is
  platform-neutral. Verified the test **reddens without the fix**: temporarily reverted the
  function's body to the pre-fix single-call shape
  (`if (now_ns() < deadline_ns) sleep_once(deadline_ns);`) and rebuilt -- the undershoot arm failed
  3 assertions exactly as expected, confirming the test catches the #3074 defect. Reverted back
  before this commit (verified via `git diff`/rebuild that the working tree matches the intended
  fix).
- **Reasoned, not executed:** the actual `Win32SleepFallback` branch of `sleep_until_steady_ns()`
  (the `#ifdef _WIN32` code path that now calls `sleep_until_deadline_retry`) does not compile or
  run on this machine. I did not build or execute it on Windows. The change there is a
  three-argument call-site substitution with the same arguments' types and order as the function it
  replaces (`sleep_until_with_std(deadline_ns)` -> `sleep_until_deadline_retry(deadline_ns,
  steady_now_ns, sleep_until_with_std)`); `steady_now_ns` and `sleep_until_with_std` are the same
  free functions the high-resolution path above it already uses, unchanged. CI's Windows MinGW job
  will build this; I have not run it myself.

- Affected: `sleep_until_steady_ns()`'s `Win32SleepFallback` branch (Windows only, and only taken
  when `CREATE_WAITABLE_TIMER_HIGH_RESOLUTION` is unavailable).
- Unaffected: the `Win32HighResolutionTimer` branch (unchanged), the POSIX branch (unchanged),
  `next_grid_deadline_ns()`/`grid_boundaries_missed()` (untouched, from #3170).

## Risk

Low. The change is additive (a new pure function) plus a one-line call-site substitution at the
one place identified in the issue. The new function has no side effects of its own -- it only
calls the function pointers it is given -- so its worst case on a still-buggy `sleep_once` is
identical to today's behavior (return once `now_ns()` first reports past the deadline, same as
before, just via a loop that would also retry if given the chance).

## Build / test

```
ctest --no-tests=error   # 315 tests, 100% passed (baseline 314 + the 1 new precise_sleep_deadline_retry)
```

`ctest --no-tests=error -R precise_sleep` -> 1/1 passed.

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -2045,6 +2045,16 @@ if(TARGET prosper_core)
   target_link_libraries(test_videoout PRIVATE prosper_core)
   add_test(NAME videoout_queries COMMAND test_videoout)
 
+  # precise_sleep's deadline post-condition (#3074): every sleep_until_steady_ns() backend must
+  # never return before now_ns() >= deadline_ns. Win32SleepFallback did not re-check after its one
+  # OS sleep call, unlike the sibling high-resolution-timer path. CROSS-PLATFORM like test_videoout
+  # above and for the same reason: sleep_until_deadline_retry() is the pure-function loop the fix
+  # composes, deliberately extracted so its termination condition is testable with a fake clock and
+  # a fake sleep on any host, without executing the Windows-only code path or measuring wall time.
+  add_executable(test_precise_sleep tests/host/platform/test_precise_sleep.cpp)
+  target_link_libraries(test_precise_sleep PRIVATE prosper_core)
+  add_test(NAME precise_sleep_deadline_retry COMMAND test_precise_sleep)
+
   # Windows guest-execution substrate (exec_image_win.cpp): build boot_trace so the whole boot path
   # links (proves the substrate + SysV-ABI HLE boundary resolve). No evdev/Vulkan here. The guest boot
   # itself is not yet verified on Windows — see docs/PORTING.md "Windows".
--- a/prosper/src/host/platform/precise_sleep.cpp
+++ b/prosper/src/host/platform/precise_sleep.cpp
@@ -70,6 +70,20 @@ void sleep_until_with_std(uint64_t deadline_ns) {
             std::chrono::nanoseconds(deadline_ns))));
 }
 
+}  // namespace
+
+// Defined OUTSIDE the anonymous namespace and outside every #ifdef _WIN32 block: this is the
+// pure-function half of the #3074 fix; see its declaration in precise_sleep.hpp for why it takes
+// function pointers rather than calling steady_now_ns()/sleep_until_with_std() directly, and
+// test_precise_sleep.cpp for the platform-neutral test that exercises it.
+void sleep_until_deadline_retry(uint64_t deadline_ns,
+                                 uint64_t (*now_ns)(),
+                                 void (*sleep_once)(uint64_t deadline_ns)) {
+    while (now_ns() < deadline_ns) sleep_once(deadline_ns);
+}
+
+namespace {
+
 #ifdef _WIN32
 // One auto-reset timer per waiting thread. Auto-reset so the wait itself consumes the signal and no
 // stale signalled state can make the NEXT wait return instantly.
@@ -124,7 +138,11 @@ void sleep_until_steady_ns(uint64_t deadline_ns) {
     // Anything above that did not work out lands here — no worse than master's behaviour, and
     // reported distinctly so a test can tell the two apart rather than silently pacing at ~32 Hz.
     if (steady_now_ns() >= deadline_ns) return;
-    sleep_until_with_std(deadline_ns);
+    // #3074: sleep_until_with_std() is ONE OS sleep call with no post-condition re-check of its
+    // own -- unlike the high-resolution loop above, which re-reads the clock after every timer
+    // signal -- so calling it once and returning let this path come back before deadline_ns.
+    // sleep_until_deadline_retry() re-checks and, if the sleep returned early, sleeps again.
+    sleep_until_deadline_retry(deadline_ns, steady_now_ns, sleep_until_with_std);
     g_backend = SleepBackend::Win32SleepFallback;
 #else
     // glibc: sleep_until -> sleep_for -> a nanosleep loop that also restarts on EINTR, which matters
--- a/prosper/src/host/platform/precise_sleep.hpp
+++ b/prosper/src/host/platform/precise_sleep.hpp
@@ -22,6 +22,23 @@ enum class SleepBackend {
 // immediately if the deadline has already passed.
 void sleep_until_steady_ns(uint64_t deadline_ns);
 
+// The post-condition sleep_until_steady_ns() promises on EVERY backend: never return before
+// `now_ns() >= deadline_ns`. A single OS sleep primitive is not trusted to hit this on its own -- a
+// waitable timer can signal marginally early, and so can the ::Sleep()-backed fallback used when no
+// waitable timer is available (#3074) -- so a backend that cannot prove its own precision composes
+// this retry loop instead of calling its one-shot sleep directly and returning.
+//
+// `now_ns` and `sleep_once` are function pointers rather than the real clock/sleep calls so the
+// LOOP'S TERMINATION CONDITION is a pure function, testable on any host platform without executing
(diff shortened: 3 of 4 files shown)


mattias800 (commented):
**APPROVED**

**Mutation verified independently.** Reverting the retry loop to its pre-fix single-call shape
(`while` -> `if`) builds clean and fails `test_precise_sleep`; restored, it passes.

**Making it testable off-Windows is the part that earns this a merge.** Extracting
`sleep_until_deadline_retry(deadline_ns, now_ns, sleep_once)` outside every `#ifdef _WIN32` block,
with the clock read and one-shot sleep injected as function pointers, turns a Windows-only branch
into something a Linux CI run can actually pin. Following #3170's precedent of testing this area at
the pure-function level — and explicitly *not* writing a test that measures wall-clock sleep
duration — is right: a timing-duration test here would pass or fail on scheduler mood, which is the
exact failure mode this file's neighbourhood keeps producing.

**The verified/reasoned split is stated correctly.** The retry function and all four arms
(undershoot, exact hit, overshoot, already-past) are verified on Linux; the `Win32SleepFallback`
call-site substitution is reasoned, and you said so rather than describing it as tested. That is the
right claim to make from a Linux box, and the MinGW job is the thing that checks the rest.

**Also right to sidestep the open question rather than answer it.** Whether mingw-w64's
`sleep_for`→`::Sleep` conversion can itself truncate was never verifiable here either; a fix whose
correctness does not depend on that answer is strictly better than one that assumes it, and the
issue said as much.

PR #3031, opened 2026-08-26
gpu: stride the compare-flag buffer by the descriptor-offset limit, not by the flag size

mattias800:
## The defect

Each compare target gets one `uint32_t` "did the GPU result change" flag, and each flag is bound
through its **own** storage-buffer descriptor:

```cpp
infos[j][2] = {compare_flags, j * sizeof(uint32_t), sizeof(uint32_t)};
```

A `VK_DESCRIPTOR_TYPE_STORAGE_BUFFER` descriptor's offset must be a multiple of
`minStorageBufferOffsetAlignment`, so the stride here is governed by the **descriptor-offset limit**,
not by the size of the datum. With two or more targets the second descriptor lands at offset 4 and
violates `VUID-VkWriteDescriptorSet-descriptorType-00328`.

The reported write index confirms the mapping exactly: `pDescriptorWrites[5]` is `j*3 + binding` for
`j=1, binding=2`, and `1 * 4 = 4`.

## Why it reached master

The limit is **4 on RADV and 16 on lavapipe**. So the code is correct on every developer GPU in use
here and wrong on CI's software rasteriser — and CI's `Vulkan validation scan` step has been
**skipped rather than passing** on every master commit since the tests went red, because it runs
after `Test`. It last ran green at `97ecc58a^`. Nothing was reporting it.

It became visible only when #3028 took the test count to 0 failures and the job reached the scan
again.

## The fix

Cache the limit on the compute context — `query_subgroup_support()` already reads
`VkPhysicalDeviceProperties2` and is called on **both** init paths (adopted device and own device) —
and route every site through a single accessor:

```cpp
VkDeviceSize compare_flag_stride() const {
    return std::max<VkDeviceSize>(sizeof(uint32_t), storage_buffer_offset_alignment);
}
```

Both terms are powers of two, so the max is a multiple of the alignment for every value the limit can
take. Seven sites now share it and cannot drift apart: allocation, the descriptor offset, the
`vkCmdFillBuffer` clear, both memory barriers, the mapped range, and the readback.

## The readback was a second, quieter bug

It is worth calling out separately, because it is not a validation issue and would never have been
reported by any layer:

```cpp
const auto* changed = static_cast<const uint32_t*>(mapped);
... changed[j] == 0
```

Indexing a `uint32_t*` by target reads the **padding between flags** on any device whose alignment
exceeds four. Every target after the first would come back zero — i.e. "GPU result unchanged" — which
[...]

PR #3183, opened 2026-09-01
tests(gpu/compute): mutation-verified regression arm for the compare-flag readback stride

### LINKED ISSUE #3032: The compare-flag readback stride has no regression arm, and the suite provably cannot see it
Follow-up from #3031, raised in review. The fix there is right; this is about it being untested.

## The gap, demonstrated rather than argued

#3031 changed two things: the descriptor **offset** (a spec violation, caught by the validation
layer) and the host **readback**, which indexed a `uint32_t*` by target and would read the padding
between flags once the buffer is strided — reporting every target after the first as
`gpu_result_unchanged`, a silent wrong answer no layer can report.

Independent review mutated #3031's head to keep the stride but restore `changed[j]` — precisely the
half-fix — and the whole suite still passed:

```
302/302 under lavapipe, with storage_buffer_offset_alignment = 16 live
```

So `[vkval] PASS` certifies the offset half only. Nothing currently reaches the shape that can
observe the defect.

## Why the suite cannot see it

Observing the padding read needs all of:

1. **two or more compare targets in one dispatch** — with one target, `j` is always 0 and
   `j * stride == 0` for any stride, so the strided and unstrided readbacks are byte-identical;
2. a target at **`j >= 1`** whose GPU result genuinely **changed**, so its flag is non-zero and the
   wrong read returns 0 where the right read returns non-zero;
3. a device with **`minStorageBufferOffsetAlignment > 4`**, i.e. lavapipe, not RADV.

Miss any one and the two implementations agree. Note #1727's own body claimed these tests "never
reach two compare targets under the layer" — the validation output proves otherwise (`x8`/`x10`
messages, and `pDescriptorWrites[5]` is `j=1`), so condition 1 IS reached today; what is missing is
condition 2.

## What to add

A focused arm asserting the writeback rather than the absence of a validation message: at least two
compare targets, the one at `j >= 1` genuinely changed, and `gpu_result_unchanged == false` for it.
The discriminating check is that the arm must **fail** with `changed[j]` restored and stride 16 live
— that mutation is the whole point, and without running it the arm proves nothing (the same trap the
suite just fell into).

Worth running it under lavapipe explicitly, since on RADV `max(4, 4) == 4` makes stride and
`sizeof(uint32_t)` identical and the arm is vacuous there:

```bash
VK_DRIVER_FILES=/usr/share/vulkan/icd.d/lvp_icd.x86_64.json
```

## Related hardening

`VulkanComputeContext::storage_buffer_offset_alignment` defaults to **1**, which fails unsafe: a
missed query silently yields stride 4, i.e. exactly the bug #3031 fixed. Provably unreachable today
(both `init()` success paths call `query_subgroup_support()` first), but defaulting to 256 — the
maximum the limit may take — would make a future missed query fail visibly instead.

Refs #1727, #3031

mattias800:
## What this is

Follow-up from #3031 (raised in review, filed as #3032): the compare-flag readback stride fix was
correct, but the suite had no regression arm for it — mutating the readback back to the pre-#3031
shape (`changed[j]` on a `uint32_t*`, ignoring `compare_flag_stride()`) left the whole suite green.

**Test-only. No production code changed.**

## The gap, and why the suite couldn't see it

`live_compute.cpp`'s compare-flag readback needs three things at once to observe the padding read:

1. two or more compare targets in one dispatch (so `j >= 1` is reached at all),
2. the target at `j >= 1` genuinely changed (so its correct flag is non-zero),
3. a device with `minStorageBufferOffsetAlignment > 4` (lavapipe; RADV's `max(4,4)==4` makes the
   bug inert there).

Nothing in the suite reached all three together.

## The arm

A new block in `prosper/tests/gpu/recompiler/test_game_compute.cpp` dispatches a hand-written
RDNA2 kernel that stores through **two** V#s (`s0-3` / `s8-11`) into two persistent ≥1 MiB
writable buffers, bound in **one** `ComputeItem`. Both get a retained GPU-side result baseline
from a first dispatch. A second dispatch changes only the second target's (`j=1`) fill value and
asserts:

- the first target (`j=0`) keeps its unchanged guest bytes, and the retained-comparator fast path
  still engaged (`live_compute_buffer_gpu_result_skips()` increased) — proves the mechanism still
  fires at all, not just that the arm happens to avoid it;
- the **second** target's guest bytes reflect the new value, not the stale baseline — this is the
  discriminating assertion.

## Verification — the mutation, applied and reverted

Exact mutation (`prosper/frontends/shared/live/live_compute.cpp`, in the compare-flag readback):

```cpp
// before (correct, current master):
const auto* flag_bytes = static_cast<const uint8_t*>(mapped);
for (size_t j = 0; j < compare_targets.size(); ++j) {
    CompareTarget& target = compare_targets[j];
    uint32_t changed = 0;
    std::memcpy(&changed, flag_bytes + j * flag_stride, sizeof(changed));
    if (target.buffer) target.buffer->gpu_result_unchanged = changed == 0;
    if (target.image) target.image->gpu_result_unchanged = changed == 0;
}

// mutated (pre-#3031 shape, reintroduced only to prove the arm catches it):
const auto* changed = static_cast<const uint32_t*>(mapped);
for (size_t j = 0; j < compare_targets.size(); ++j) {
    CompareTarget& target = compare_targets[j];
    if (target.buffer) target.buffer->gpu_result_unchanged = changed[j] == 0;
    if (target.image) target.image->gpu_result_unchanged = changed[j] == 0;
}
```

All runs under `VK_DRIVER_FILES=/usr/share/vulkan/icd.d/lvp_icd.x86_64.json` (lavapipe,
`storage_buffer_offset_alignment=16`):

| build | mutation applied? | new arm present? | result |
| --- | --- | --- | --- |
| origin/master's test file (no new arm) | yes | no | `== PASS == (407 assertions executed)`, exit 0 — **confirms the blindness the issue describes** |
| this branch | yes | yes | `== FAIL: 1 == (415 assertions executed)`, exit 1, failing exactly `#3032 genuinely-changed second target (j=1) publishes its new guest bytes, not stale padding-read baseline bytes` — every other check (including the sibling j=0/fast-path checks) still passes |
| this branch, mutation reverted (`git checkout --`) | no | yes | `== PASS == (415 assertions executed)`, exit 0, under **both** RADV and lavapipe |

Full suite with the real fix restored:

```
ctest --no-tests=error
100% tests passed out of 314
```

Fixes #3032

### DIFF
--- a/prosper/tests/gpu/recompiler/test_game_compute.cpp
+++ b/prosper/tests/gpu/recompiler/test_game_compute.cpp
@@ -1341,6 +1341,159 @@ int main() {
 #endif
     }
 
+    // #3032: the compare-flag readback stride had no regression arm because nothing in this suite
+    // ever reached two GPU-compared targets in one dispatch with a genuinely-changed target at
+    // j >= 1 -- the only shape that can observe indexing the flags buffer by sizeof(uint32_t)
+    // instead of by the descriptor-offset stride (#3031). Two persistent >=1 MiB writable buffers
+    // bound to ONE dispatch, each already carrying a retained GPU-side result baseline from a prior
+    // dispatch of the same item. The first (j=0) is re-dispatched with an unchanged fill value; the
+    // second (j=1) is re-dispatched with a genuinely different one. A correct readback must publish
+    // the new guest bytes for j=1; reading the padding between flags on a device whose storage-buffer
+    // offset alignment exceeds four (lavapipe: 16) would silently report it unchanged instead,
+    // exactly like j=0, and leave the stale baseline bytes in guest memory.
+    {
+        constexpr uint32_t compare_stride_target_bytes = 1u << 20;
+        std::vector<uint8_t> compare_stride_hostA(compare_stride_target_bytes, 0);
+        std::vector<uint8_t> compare_stride_hostB(compare_stride_target_bytes, 0);
+
+        // Hand-written RDNA2: v4 = 0 (single invocation, no thread-index dependency needed); store
+        // s4-s7 through the V# at s0-s3 (binding 0, j=0), then s12-s15 through the V# at s8-s11
+        // (binding 1, j=1). Both stores are buffer_store_dwordx4, idxen, offset 0 -- word-for-word
+        // the Dead Cells fill kernel's decoded MUBUF encoding above (opcode 7, VDATA/VADDR/SRSRC/
+        // SOFFSET per rdna2_decode.cpp's own field layout), just re-targeted to a second V#/value
+        // pair for the second store. VOP1 v_mov_b32 = 0x7E000000 | vdst<<17 | 1<<9 | src0, verified
+        // against both `v0 = s4` (this file) and `v5 = y = 0` (image_copy_2d below).
+        static const uint32_t two_target_fill[] = {
+            0x7e080280u,               // v4 = 0
+            0x7e000204u,               // v0 = s4
+            0x7e020205u,               // v1 = s5
+            0x7e040206u,               // v2 = s6
+            0x7e060207u,               // v3 = s7
+            0xe01c2000u, 0x80000004u,  // buffer_store_dwordx4 v[0:3], v4, s[0:3], offset0 idxen (j=0)
+            0x7e0a020cu,               // v5 = s12
+            0x7e0c020du,               // v6 = s13
+            0x7e0e020eu,               // v7 = s14
+            0x7e10020fu,               // v8 = s15
+            0xe01c2000u, 0x80020504u,  // buffer_store_dwordx4 v[5:8], v4, s[8:11], offset0 idxen (j=1)
+            0xbf810000u,               // s_endpgm
+        };
+
+        ShaderResourceTable compare_stride_rt;
+        ShaderResource compare_stride_a;
+        compare_stride_a.cls = ResourceClass::ConstantBuffer;
+        compare_stride_a.format = DataFormat::Uint32;
+        compare_stride_a.num_components = 4;
+        compare_stride_a.binding = 0;
+        compare_stride_a.sgpr_base = 0;
+        compare_stride_a.stride = 4 * sizeof(uint32_t);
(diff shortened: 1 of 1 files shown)


mattias800 (commented):
**APPROVED** — with one finding about the guard's reach that belongs in the record.

**Your mutation claim is correct, and I reproduced it — but only on one of two devices here.**
Applying your stated mutation verbatim (`flag_bytes + j * flag_stride` ->
`flag_bytes + j * sizeof(uint32_t)`, i.e. the pre-#3031 `changed[j]` shape), with a clean build both
times:

| device | `minStorageBufferOffsetAlignment` | mutant | restored |
| --- | --- | --- | --- |
| RADV STRIX_HALO (this box's default) | **0x4** | `rc=0` — **not caught** | `rc=0` |
| lavapipe (your environment) | **0x10** | `rc=1` — **caught** | `rc=0` |

The reason is structural rather than flaky: when the device's storage-buffer alignment is 4, the
compare-flag stride *equals* `sizeof(uint32_t)`, so the mutation is a literal no-op and no test could
detect it. Your run was under lavapipe, which is why you saw it fail — and you did say so, which is
what let me find this rather than conclude your report was wrong.

**This does not block the merge**: the arm genuinely catches the defect wherever the defect can
exist, and a stride bug is only a bug when stride != element size. But it means the guard's coverage
depends on which ICD the runner picks, and nothing in the test says so.

**Worth a follow-up** (not required here): have the arm assert its own precondition — read
`minStorageBufferOffsetAlignment` and either skip loudly or fail as apparatus-invalid when it is 4,
rather than passing silently on a device where it cannot discriminate. That is the charter's
"prefer experiments that detect their own invalidity" rule, and this is a textbook instance: on RADV
the test currently reports success for a reason unrelated to the code under test.

**The blindness confirmation is solid** — mutating the readback on unmodified master and observing
`== PASS == (407 assertions)` exit 0 establishes the gap the issue asserts, and your new arm failing
with exactly one line (`#3032 genuinely-changed second target (j=1) publishes its new guest
bytes...`) while siblings stay green shows it isolates rather than fails broadly.

Test-only diff, `live_compute.cpp` untouched at HEAD — confirmed.

Issue #3184, opened 2026-09-01
shader_conformance: replace scan.py's ported MIMG length rules with the real decoder (shader_histo)


mattias800:
`tools/shader_conformance/scan.py`'s MIMG walk is now length-aware (#3040): it decodes each RDNA2
instruction's real per-instruction length before deciding whether a dword is the start of a MIMG
instruction, mirroring the length-dispatch rules in `src/gpu/recompiler/rdna2_decode.cpp`
(`rdna2_decode_one` / `rdna2_walk`).

That mirror is a **second, independent implementation** of those rules rather than a call into the
real decoder. #3040's own body already named the more robust fix and called it out explicitly:

> Suggested fix — reuse prosper's decoder rather than reimplementing it. `tools/shader_histo`
> already walks the stream with the real decoder and already knows the MIMG class
> (`shader_histo.cpp:88`). Teaching it to emit per-instruction `(pc, opcode, dim)` — behind a flag,
> so its existing output is unchanged — and having `scan.py` consume that would make the walk exact,
> delete the Python bit-twiddling, and keep a single source of truth for RDNA2 decoding.

#3040's PR ported only the LENGTH rule (not full operand decode), which is a much smaller and more
contained piece than the "~1,200 lines" the original issue estimated for a full port — but it is
still a duplicate of logic that already exists in `rdna2_decode.cpp`, and it will silently drift out
of sync if that file's length dispatch changes (a new instruction format, a new literal-forcing
opcode, a widened NSA field) without a matching update in `scan.py`.

**This issue tracks doing the suggested fix**: teach `shader_histo` (or `gpu_replay`) to emit real
per-instruction `(pc, opcode, dim)` for MIMG sites, behind a flag that leaves its existing output
unchanged, and have `scan.py` consume that instead of `rdna2_instr_len`'s Python port. That removes
the duplication entirely and makes the walk exact by construction rather than by two decoders
happening to agree.

Not urgent — the Python port fixed the actual phantom-finding defect and is covered by
`scan.py --self-test` (`shader_conformance_scan_logic` in ctest), including a regression test that
fails if the length-aware walk is reverted. This is the follow-up architecture cleanup #3040 itself
recommended, tracked separately so it isn't lost.

Refs #3040

mattias800:
CLAIMING: Lane B (session 0154kwL1ddgwCADK6WPDimaB) | branch fix/issue-3184-real-mimg-decoder | 2026-09-02T00:00Z UTC

PR #3039, opened 2026-08-26
tools(shader_conformance): stop the scanner manufacturing a clean result

mattias800:
Independent review of #3036 returned **REJECTED** with five blocking findings. #3036 is already
merged; nothing in it changes emulator behaviour, so this is a follow-up rather than a revert. The
verdict's practical meaning was: **do not use the scanner to clear a title until these land**, and
re-derive any clean result already taken from it.

## The defect class

The scanner's entire value is the CLEAN direction. As merged it could produce that answer without
having read anything — which is worse than having no scanner, because it launders a null.

| # | Defect | Fix |
| --- | --- | --- |
| 1 | A capture yielding zero examined shaders did not change the exit code — the guard was global, so one readable capture certified a whole run | Per-capture accounting; any barren capture exits 2 |
| 2 | Findings reported against captures never read — one shared temp dir, filenames keyed only on shader hash, success judged by `os.path.exists` instead of the child's return code | Per-capture temp dir; every dump judged by return code |
| 3 | An empty ISA dump counted as an examined shader | Counted as a skip, with a reason |
| 4 | An uncaught exception exited 1 — indistinguishable from "mismatches found" | Exits 2 |
| 5 | The self-tests covered neither the classifier, the exit contract, nor either parser's caller | Rewritten; see below |

## The self-tests

The old suite passed under **nine** mutations of the logic it claimed to cover, including
`DIM_ARRAYED = {4, 7}` — which drops the entire 2D_ARRAY class the tool exists to find. Check 1 was
a same-source positive control: its MIMG fixture was assembled from `MIMG_ENCODING`, so mutating the
constant mutated the fixture and the check passed either way.

The new suite is **hermetic** — it drives the scanner end to end against a stub `gpu_replay` and a
`spirv-dis` shim, so it needs neither a GPU nor the Vulkan toolchain — and covers both parsers, the
classifier in **both directions**, and the exit contract. Fixtures are hand-computed literal words
(`0xF0800028`), not derived from the constants under test.

Mutation-tested individually, **10/10 caught**:

```
[CAUGHT] DIM_ARRAYED drops 5 (kills the whole #325 class)
[CAUGHT] emitted_arrayed forced True (never reports)
[CAUGHT] exit-2 guard removed
[CAUGHT] MIMG_ENCODING off by one
[CAUGHT] opcode MSB term deleted
[CAUGHT] dim field shifted wrong
[CAUGHT] Arrayed operand -> Depth
[CAUGHT] 2-dword skip reverted to 1
[...]

PR #3185, opened 2026-09-01
tools/shader_conformance: make scan.py's MIMG walk length-aware (#3040)

### LINKED ISSUE #3040: shader_conformance scan.py: MIMG walk is not length-aware, so findings can be phantoms
`tools/shader_conformance/scan.py` finds MIMG instructions by scanning every dword for the encoding
`0b111100` in bits [31:26]. That is not a real instruction walk: without per-instruction lengths, a
literal constant or an operand dword whose top six bits happen to match reads as an instruction
start. prosper's own `src/gpu/recompiler/rdna2_decode.cpp` carries the lengths and records what
mis-sizing costs — *"the extra dword is mis-decoded as a phantom instruction, derailing the stream
walk"* (`:966`, `:1091`).

**Impact is one-directional and therefore bounded**: the scanner can invent an MIMG that is not
there, and cannot hide one that is. So a clean result stays sound for the classes it checks, while
every *finding* is a candidate to confirm rather than a defect. #3036's headline "9 findings over 71
shaders" is an upper bound, not a count.

**Suggested fix — reuse prosper's decoder rather than reimplementing it.** `tools/shader_histo`
already walks the stream with the real decoder and already knows the MIMG class
(`shader_histo.cpp:88`). Teaching it to emit per-instruction `(pc, opcode, dim)` — behind a flag, so
its existing output is unchanged — and having `scan.py` consume that would make the walk exact,
delete the Python bit-twiddling, and keep a single source of truth for RDNA2 decoding. That is
strictly better than porting ~1,200 lines of length rules into Python, where it would drift.

Until then, `scan.py` prints the caveat whenever it reports findings, and
`tools/shader_conformance/AGENTS.md` states the limit. Both landed in #3039.

Found by independent review of #3036 (finding 5).

mattias800:
Fixes #3040. Refs #3184 (follow-up: replace the ported length rules with the real decoder,
per #3040's own suggested fix).

## Problem

`scan.py`'s `decode_mimg_sites` found MIMG instructions by testing **every dword** for the
MIMG top-6-bit encoding (`0b111100`), regardless of what instruction that dword actually
belonged to. An operand or trailing-literal dword whose top bits happened to alias the
encoding read as an instruction start -- a phantom finding for an MIMG that was never in
the program. #3040's title says it plainly: findings can be phantoms.

Demonstrated on current master before touching any code: a hand-built 2-dword "shader"
containing exactly one real instruction (`v_mov_b32 v0, 0xF0000028` -- VOP1 with a forced
literal operand whose *value* aliases an arrayed-DIM MIMG encoding) decoded to one MIMG
site and produced a phantom `array-layer-dropped` finding against a module that samples no
image at all:

```
decode_mimg (master, dword-scan): [('0x0', 5)]
classify() findings: [{'klass': 'array-layer-dropped', 'detail': '#325', ...}]
PHANTOM FINDING PRODUCED: True
```

## Fix

`decode_mimg_sites` now walks real instruction boundaries. `rdna2_instr_len` (new) ports
the **length-dispatch rules only** (not full operand decode -- no register fields,
DPP/SDWA sub-modes, condition codes) from prosper's own `rdna2_decode_one` / `rdna2_walk`
in `src/gpu/recompiler/rdna2_decode.cpp`, covering every format bucket that function
dispatches on: the vector-ALU group (VOP1/VOP2/VOPC, including the SDWA/DPP modifier and
K-literal forcing rules), the scalar group (SOPP/SOPC/SOP1/SOPK/SOP2), and the
bit[31:30]==11 group (VINTRP, VOP3P, VOP3, DS, FLAT, MUBUF, MTBUF, SMEM, EXP, and MIMG
itself). A dword is only ever tested for MIMG-ness when it is genuinely the first dword of
an instruction.

**The length rule for MIMG** (the part the issue specifically asked me to derive rather
than guess, from AMD's "RDNA 2" ISA Reference Guide, document 70648): MIMG is 2 dwords,
plus a variable NSA (Non-Sequential Address) extra-dword count read from `dword0[2:1]`
(0-3 extra dwords), so total length is 2-5 dwords. `rdna2_decode.cpp:1134-1150` already
carries this rule for the real decoder; I re-derived it here rather than copying, and
pinned it with a hand-built NSA fixture in `self_test` (below).

This fixes **both** known directions at once, because both come from the same root cause
-- testing a dword that is not an instruction boundary:
- **#3040** (this issue): inventing an MIMG from an operand/literal dword.
- **#3039** (already fixed, re-verified here): a phantom match swallowing a real MIMG that
  follows it. The exact #3039 fixture now produces `[(0x20, 5)]` -- the real MIMG, and
  *only* it; before this PR the same fixture produced that plus a spurious extra hit for
  the aliasing literal dword.

`phantom_risk` (the `--json` field that counted MIMG candidates sitting immediately after
another candidate, as a "this shader's findings deserve extra suspicion" signal) is
removed. Under a length-aware walk it can never be nonzero -- two real MIMG sites can
never be exactly 1 dword apart, since MIMG is at minimum 2 dwords -- so it was reporting a
signal that no longer exists. Nothing else in the repo reads that field.

## Residual risk, stated plainly

This Python walk is a **second, independent implementation** of `rdna2_decode_one`'s
length dispatch, not a call into it. #3040's own body already named the more robust fix --
teaching `shader_histo`/`gpu_replay` to emit real per-instruction `(pc, opcode, dim)` and
having `scan.py` consume that -- and called porting the rules into Python worse, because it
can drift. That's still true in principle, though the actual port needed here (length only,
no operand decode, no DPP-family admission logic) was on the order of 100 lines, not the
"~1,200 lines" the issue estimated for a full decoder port. I filed #3184 to track doing
the suggested fix properly rather than silently declaring this PR a substitute for it.

## Test added, and proof it fails without the fix

`self_test` (`shader_conformance_scan_logic` in ctest) gained:

1. **The positive control the task asked for, built BY HAND** (not from `MIMG_ENCODING` /
   `DIM_ARRAYED` -- a same-source control tests the discriminator, not the domain): a
   `v_mov_b32` with a literal operand hand-picked so *its own* top 6 bits alias the MIMG
   encoding with DIM=5 baked into bits [5:3]. Checked at both the `decode_mimg` layer and
   through the full `classify()` pipeline (no phantom `array-layer-dropped` finding).
2. **An NSA-length test**: a MIMG instruction with NSA=1 (one extra address dword), whose
   *own* extra dwords are themselves crafted to alias the MIMG encoding, followed by a real
   MIMG. Proves the walk advances by the NSA-derived length rather than a fixed 2 dwords --
   if it didn't, the extra dwords would each read as phantom instructions and the walk
   would desync and miss the real MIMG that follows.
3. Updated two pre-existing assertions whose expected values encoded the *old*,
   non-length-aware behavior (a real MIMG's own dword1, and the #3039 swallow fixture, each
   used to intentionally produce an extra phantom hit under "every dword is tested"; they
   now assert the single correct hit).

**Verified the new/updated checks fail without the fix**: reverted only `rdna2_instr_len` /
`decode_mimg_sites` to the pre-#3040 dword-by-dword scan (kept `phantom_risk` removed, kept
every test as committed) and re-ran `--self-test`. Exactly the 5 checks tied to
length-awareness went red; all other checks (including the untouched MIMG/SPIR-V/classifier/
exit-contract checks) stayed green:

```
[FAIL]   a real MIMG's own dword1 is consumed by that instruction, not read as a second one (#3040)
[FAIL]   a real MIMG immediately after an aliasing literal is found, and ONLY it is (#3039 + #3040)
[FAIL]   a literal operand aliasing an ARRAYED MIMG encoding is not read as an instruction (#3040)
[FAIL]   ...and so the full pipeline reports no phantom array-layer-dropped finding either
[FAIL]   NSA's extra dwords (dword0[2:1]) are skipped, not read as instructions, and the real MIMG that follows is found at the correct offset (#3040)
== self-test FAILED ==
```

Restored the fix afterward; current head passes 33/33.

## Corpus re-run: before/after finding counts

Built `gpu_replay` from this branch and ran both the pre-fix `scan.py` (checked out from
`origin/master`) and the fixed one against 4 real captures spanning 4 different titles
(captures produced by other in-flight bring-up lanes, not committed):

| title | shaders examined | skipped | findings before | findings after |
| Grand Theft Auto V | 56 | 0 | 0 | 0 |
| Tomb Raider I-III Remastered | 71 | 0 | 0 | 0 |
| Dragon Quest VII Reimagined | 52 | 13 | 0 | 0 |
| Unbound (partial capture) | 8 | 0 | 0 | 0 |

**No finding count changed anywhere in this corpus -- 0 before and 0 after in every case,
187 shaders total.** So there is nothing to explain away: no real finding disappeared, and
none of these 4 captures happened to exercise the specific aliasing bit pattern the
hand-built control targets (unsurprising -- it needs a particular literal value to land on
an instruction boundary the old walk also treated as a candidate). The phantom-elimination
claim rests on the hand-built control above, which is the artifact the task asked for
precisely because a real-world corpus is not guaranteed to contain an instance; the corpus
run's job is instead to confirm **no regression** in real-world recall, which it does.

## Verification

- `python3 tools/shader_conformance/scan.py --self-test`: 33/33 checks pass.
- `ctest --no-tests=error -R shader_conformance_scan_logic`: 1/1 passed.
- `git diff --check`: clean.
- `git log origin/master..HEAD --format=%B | grep -c '^Claude-Session:'`: 0.
- Python-only change, no build-graph impact beyond the existing `gpu_replay`/`spirv-dis`
  dependency this tool already had.

### DIFF
--- a/prosper/tools/shader_conformance/AGENTS.md
+++ b/prosper/tools/shader_conformance/AGENTS.md
@@ -68,33 +68,41 @@ those counts alongside any clean result** — an exit code alone is not falsifia
 
 ### What a finding is, and what a clean result is
 
-These two directions do **not** carry the same weight, and the difference is structural.
+**The ISA walk is length-aware (#3040).** `decode_mimg_sites` decodes each instruction's real
+per-instruction length — mirroring the format dispatch and length rules in prosper's own
+`rdna2_decode.cpp` (`rdna2_decode_one` / `rdna2_walk`) — and only ever tests a dword for the MIMG
+top-6-bit encoding when it is genuinely the first dword of an instruction. An operand or trailing
+literal dword can no longer be misread as an instruction start, and a real MIMG's own address/NSA
+dwords can no longer be misread as separate instructions either.
 
-The ISA walk is **not length-aware**. A correct walk needs per-instruction lengths (prosper's own
-`rdna2_decode.cpp` has them, and records what mis-sizing costs: a phantom instruction that derails
-the stream walk). Without them, a literal constant whose top six bits happen to read as the MIMG
-encoding looks like an instruction.
+Before #3040 the walk tested *every* dword unconditionally, which was deliberately over-approximating
+in one direction (never hide a real MIMG) but paid for that by inventing MIMG instructions from
+operand and literal dwords whose top bits happened to alias the encoding — a finding for a defect
+that did not exist. Review of #3039 had already found the opposite failure mode in the pre-#3039
+walk (a phantom swallowing a real MIMG that followed it, scanning CLEAN); the length-aware walk fixes
+both directions at once, because both come from the same root cause — testing a dword that is not
+actually an instruction boundary. `self_test` pins both: a hand-built literal aliasing an arrayed DIM
+that must NOT be found, and the #3039 swallow sequence whose real MIMG must still be found (and, now,
+found *alone*).
 
-**Every dword is tested, including the one after a match.** Skipping it looks free — a real MIMG is
-at minimum two dwords, so its dword1 cannot start an instruction — but that is true of a *real*
-MIMG and false of a *phantom*, and a phantom that swallows the next dword can hide a genuine
-instruction. Review of #3039 demonstrated exactly that: an aliasing literal followed by a real
-`image_sample dim:2D_ARRAY` against a non-arrayed image scanned CLEAN. `--json` reports
-`phantom_risk` per shader (candidates sitting immediately after another candidate) so the suspicion
-is visible; it never suppresses a finding.
+Residual risk is different in kind, not degree: this Python walk is a **second, independent
+implementation** of `rdna2_decode_one`'s length dispatch rather than a call into it, so a future
+change to the real decoder's length rules needs a matching update here or the two can drift apart
+silently. #3040's own suggested fix — teaching `shader_histo` to emit real per-instruction `(pc,
+opcode, dim)` and having `scan.py` consume that instead — removes the duplication entirely and
+remains the more robust direction; it just was not required to fix the length-awareness defect
+itself. Until it lands:
 
-So the error is **one-directional**: the finder can invent an MIMG that is not there, and cannot hide
-one that is.
+- **A finding is still a candidate**, not a defect — confirm it against the shader before acting on
+  it. The remaining reason is `classify()`'s module-wide match (below), not the ISA walk.
+- **A clean result is sound for the classes it checks** — the walk can neither invent an instruction
+  nor miss a real one it can decode.
 
-- **A finding is a candidate**, not a defect. Confirm it against the shader before acting on it.
-- **A clean result is sound for the classes it checks** — a superset that found nothing means there
-  was nothing to find.
-
-Two further recall limits, both deliberate: `OpTypeImage` is parsed module-wide, so a shader that
-declares *any* arrayed image clears *every* arrayed sample in it (without per-binding attribution,
-claiming otherwise would report shaders that are actually correct); and `max_coord_arity` is
-reported for information only — it feeds no decision, and reads 0 when a coordinate is not built by
-`OpCompositeConstruct`.
+Two further recall limits, both deliberate and unrelated to the ISA walk: `OpTypeImage` is parsed
+module-wide, so a shader that declares *any* arrayed image clears *every* arrayed sample in it
+(without per-binding attribution, claiming otherwise would report shaders that are actually correct);
+and `max_coord_arity` is reported for information only — it feeds no decision, and reads 0 when a
+coordinate is not built by `OpCompositeConstruct`.
 
 ## Self-test
 
--- a/prosper/tools/shader_conformance/scan.py
+++ b/prosper/tools/shader_conformance/scan.py
@@ -30,59 +30,152 @@ DIM_NAMES = {0: "1D", 1: "2D", 2: "3D", 3: "CUBE",
 DIM_ARRAYED = {4, 5, 7}
 DIM_3D = {2}
 MIMG_ENCODING = 0b111100          # dword0[31:26]
+S_ENDPGM = 0xBF810000             # SOPP encoding of s_endpgm -- terminates the walk, as it terminates
+                                   # the real program (rdna2_decode.cpp's S_ENDPGM).
+
+
+class Step:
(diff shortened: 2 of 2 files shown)


mattias800 (commented):
**APPROVED**

**Mutation verified independently.** Forcing `rdna2_instr_len` to always return 1 — which reduces the
walk to the old every-dword scan — fails `--self-test` (`rc=1`); restored, it passes. The arms are
load-bearing.

**The root cause is correctly identified and the fix is scoped to it.** Testing every dword for the
MIMG top-6-bit pattern regardless of what instruction it belonged to means any operand or trailing
literal that happens to alias `0b111100` reads as an instruction. Porting only the *length*-dispatch
rules — not full operand decode — is the right amount of decoder to duplicate: enough to find
instruction boundaries, not enough to become a second decoder that must agree with the first about
semantics.

**Deriving the MIMG rule rather than guessing it** (2 dwords plus an NSA extra count from
`dword0[2:1]`, 0–3, so 2–5 total, per doc 70648 and `rdna2_decode.cpp:1134-1150`) is what makes this
checkable.

**Removing `phantom_risk` is the right call and easy to get wrong in the other direction.** Under a
length-aware walk it can never be nonzero, so leaving it in would report a signal that no longer
exists — a metric that always reads zero looks like evidence of health rather than of its own
obsolescence.

**The corpus result is reported honestly, which is the part I want to highlight.** 0 findings before
and 0 after across 187 shaders in 4 titles is *not* evidence the fix works, and you said so: the
aliasing needs a coincidental literal value, so the corpus was never likely to contain it, and the
hand-built positive control carries the proof instead. Most reports would have presented "no
regressions across 187 shaders" as the headline. Framing it as "nothing changed, so there is nothing
to explain away as a lost real finding" is the accurate reading.

**Updating two pre-existing assertions that had encoded the old buggy behaviour as correct** is also
right, and you named them rather than letting them slide through as incidental churn.

**The drift risk is real and correctly filed** (#3184): a Python walk that duplicates rather than
calls the C++ decoder's length dispatch can silently diverge if the C++ rules change. Noting it in
code comments, `AGENTS.md`, the PR body *and* a tracking issue is the right amount of paranoia for a
known-temporary duplication.

Verified: `--self-test` 33/33, `shader_conformance_scan_logic` 1/1, no `Claude-Session:` trailer, no
absolute host paths.

PR #2958, opened 2026-08-23
#2945 is the driver, not prosper's SPIR-V — a hand-written module fails identically, and the trigger is concurrent GPU load

mattias800:
## #2945 is decided: the driver or the hardware, not prosper's generated SPIR-V

#2945 was down to two suspects — prosper's generated SPIR-V, or RADV/the AMD hardware. This settles
it by hand-writing SPIR-V that does the same work and running it **beside prosper's own module in
the same process**.

**Verdict: the driver or the hardware.** It is not prosper's SPIR-V, and — a second result that
falls out of the same instrument — it is not storage-buffer loads either. It is `vkCmdDrawIndexed`.

### The evidence

All 2026-08-23, Linux, AMD Radeon 8060S (RADV STRIX_HALO, `0x1586`), Mesa 26.1.6 host / 26.1.4
container, kernel 7.1.5. Paired runs are one process, two pipelines, interleaved render-by-render,
20 iterations per run.

| arm | condition | failing RUNS | empty indexed iterations |
| --- | --- | --- | --- |
| prosper's dumped vertex module | sustained GPU load | 52 of 400 | 220 of 8,000 |
| `minimal_ssbo_vs` — hand-written, same loads, same descriptor shape | same process, paired with the above | **54 of 400** | 228 of 8,000 |
| `no_ssbo_vs` — hand-written, declares the buffer and never reads it | sustained GPU load | **142 of 400** | 675 of 8,000 |
| `index_readback_vs` — hand-written | same process, paired with the above | 138 of 400 | 669 of 8,000 |
| all four arms, NON-indexed draw | sustained GPU load | — | 0 of 32,000 (undiagnostic — see below) |
| all four | idle GPU, one ~3.4-min window | 0 of 48 each — **undecided, not clean** | 0 |
| prosper's module, `minimal_ssbo_vs`, `no_ssbo_vs` | **lavapipe**, during a window RADV was failing 20/20 | 0 each | 0 of 20 each |

- The hand-written module fails **as often as prosper's**, in the same process, on the same window —
  and the sharpest form of that is not the two totals but their run-by-run agreement: **the two arms
  disagreed on 2 of 400 runs.**
- The second campaign's `no_ssbo_vs` never reads the storage buffer and still fails, so **a
  storage-buffer load is not required at all**. (Do *not* read 142-vs-52 as "no_ssbo fails more" —
  different processes, different option set, different windows. Only the within-campaign pairs are
  comparable. This was retracted after review.)
[...]

Issue #2945, opened 2026-08-23
The renderer is not deterministic on frozen input: one .prgbundle replayed 15 times offline gives 5 different pictures


mattias800:
## Summary

**One `.prgbundle`, replayed offline with `gpu_replay`, produces a different picture from run to
run.** No guest, no route, no sampling grid, no swapchain — a frozen capture and a fixed binary, and
the output hash changes.

Measured 2026-08-23 on master `08c23efd`, Linux / AMD Radeon 8060S (RADV STRIX_HALO), Mesa 26.1.4.
One *BALAN WONDERWORLD* (`PPSA02058`) frame grab, **15 replays of the same file**:

| submit | operations | distinct output hashes over 15 replays |
| --- | --- | --- |
| 3537 (the scene) | 118 | **5** — `5c54906d637de17f` x7, `d6feaabf815c3096` x5, `5a0e2fd6393fbfb5` x1, `ad8fd75518fd10b9` x1, `126927e49922ef95` x1 |
| 3540 (the tail) | 11 | 1 — `a5e7b61cbf984383`, stable in all 15 |

`output_bytes=33177600` in every run, so nothing is dropped or truncated; the pixels differ.

## Reproduction

```bash
# ~40 s, ~80 MB, no keyboard.
SDL_VIDEODRIVER=offscreen PROSPER_RENDER=1 PROSPER_GUEST_ARGS=-force-gfx-direct \
PROSPER_CAPTURE_DIR=<work> PROSPER_GRAB_BUNDLE_AFTER_MS=21000 \
    timeout 55 ./prosper-app <DUMP_ROOT>/PPSA02058-app0

for i in $(seq 15); do
  ./gpu_replay --bundle <work>/frame_grab_*.prgbundle <work>/out.bmp 2>&1 \
    | grep -oE 'bundle-submit=[0-9]+ .* hash=[0-9a-f]+'
done | sort | uniq -c
```

Any title should do; BALAN is the one measured. `--bundle` mode rejects `--inspect-only` and
`--draw-steps`, so narrow with `--bundle-extract-submit N PATH` and then inspect the extracted
capsule.

## Why this is worth its own issue

It is the smallest reproduction anyone has for the family of defects that currently looks like three
separate problems:

- **#2932** — the finished frame loses most presents to a wrong composite, on nine titles. On BALAN
  the producer is the guest's own scanout pass, which renders the real menu on 403 of 1,255 passes
  and a flat white on the other 832 (see the #2932 thread).
[...]

Issue #1711, opened 2026-08-01
prosper ships SPIR-V that spirv-val rejects: OpAccessChain result type does not match indexing into an OpTypeVector base


mattias800:
Found by the first-ever Vulkan validation pass over prosper's test suite (#1704). Reproduces on
Mesa 25.2.8 lavapipe (the CI runner's exact driver) and on RADV.

```
VUID-VkShaderModuleCreateInfo-pCode-08737
vkCreateShaderModule(): pCreateInfo->pCode (spirv-val produced an error):
OpAccessChain result type <id> '4[%4]' (OpTypeInt) does not match the type that results from
indexing into the base <id> '6[%6]' (OpTypeVector).
  %43 = OpAccessChain %_ptr_StorageBuffer_uint %15 %uint_0 %uint_0

Command to reproduce:
	spirv-val <input.spv> --relax-block-layout --target-env vulkan1.1
```

## Why this is worth its own issue

`docs/VERIFICATION.md` and CLAUDE.md both state that prosper's shaders are `spirv-val`-gated. This
module is **not** — it reaches `vkCreateShaderModule` and is accepted by two conformant drivers
anyway, because drivers are not required to validate. So there is a live path that emits invalid
SPIR-V and a stated gate that does not cover it. Both halves are the bug.

## Reproducing

```bash
cd <build-dir>
VK_LOADER_LAYERS_ENABLE=VK_LAYER_KHRONOS_validation ./test_game_compute 2>&1 | head -20
```

The message is the **first** thing printed after `[compute] Vulkan device: ...`, i.e. it is the
first compute program `test_game_compute` creates, before the first `[gpucap] captured deferred
submit 0` line. It fires once per run on Mesa 25.2.8 and once on RADV.

## Suggested work

1. Identify the emitting kernel (the module is the first shader module created by
   `test_game_compute`) and dump it with `PROSPER_SHADER_DUMP_SUCCESS=DIR`.
2. Fix the emitter: the access chain applies one index too few for the base type it walks, so the
   result pointer type disagrees with the walked type.
3. Close the gate that was supposed to catch this — run `tools/spv_validate` (or an in-process
   `spirv-val` equivalent) over recompiler output in a test, so the next occurrence fails a build
   rather than being discovered by a layer two years later.

Currently allow-listed in `prosper/tools/vkval/allowlist.txt` so it stays visible.

mattias800:
[...]

Issue #1675, opened 2026-08-01
Dump-gated test assertions report green in CI without executing — the skip is invisible in the summary


mattias800:
Found while reviewing #1670. Not specific to that change — it applies to the whole file and possibly others.

## The gap

`prosper/tests/test_plugin_autolink.cpp` contains assertions that only execute when the test is given a
game-dump path. `prosper/CMakeLists.txt:547-551` registers it with `${GAME_DUMP}`, and
`.github/workflows/ci.yml` never sets `GAME_DUMP`. So in CI the test runs, prints

```
  [skip] no game dump argument — link wiring not exercised
```

and passes — reporting green for assertions that never executed.

That is not a false pass in the usual sense: the test correctly declines to test what it cannot reach.
The problem is that **a green CI check reads as "the link wiring is covered" when it is not**, and
nothing in the output of a CI run distinguishes "12 assertions passed" from "12 assertions skipped".

## Why it matters now

#1670 added six assertions covering duplicate-export aliasing — a behaviour measured to occur on 7 of 30
local dumps. Those six also do not run in CI. A regression in the alias-recording path would be caught
only by an agent who happens to run the suite locally with a dump, which is not a gate.

Dumps are gitignored and cannot be committed, so this cannot be fixed by giving CI a dump.

## Options, roughly in increasing cost

1. **Make the skip visible in the CI summary.** A test that skips its substantive block could return a
   distinct ctest status (`SKIP` via `set_tests_properties(... SKIP_RETURN_CODE ...)`) so the run reports
   "N passed, M skipped" instead of counting them as passed. Cheapest, and it makes the gap legible
   rather than closing it.
2. **Synthesize the inputs.** The link wiring needs real SELF/PRX files. A tiny generated fixture module
   — two of them exporting one shared NID — would exercise `link_program`'s export table, collision guard
   and alias recording with no dump. This is the real fix for the loader assertions specifically, and it
   is the kind of thing `tools/` already does elsewhere.
3. Accept it and document it, so nobody reads the green check as coverage.

## Scope check first

Before doing any of these, it is worth counting how many registered tests are dump-gated and how many
assertions sit behind those gates — I have not measured that, and it determines whether this is a
one-file annotation or a fixture-building exercise.

`CONFIDENCE: HIGH` that the gap exists (read directly from `CMakeLists.txt` and `ci.yml`);
`CONFIDENCE: LOW` on its breadth, which is the thing to establish first.

mattias800:
[...]

PR #3186, opened 2026-09-01
ci: fail the Linux job when the spirv-tools-gated control check is not registered (#2960)

### LINKED ISSUE #2960: vkprobe_hand_written_controls can silently unregister itself — the spirv-tools gate needs a by-name CI guard
## The `vkprobe_hand_written_controls` ctest case can silently unregister itself

`prosper/tools/vkprobe/shaders/` holds hand-written SPIR-V modules whose entire value is being
**valid by construction** — they are the control arm that decided #2945 by doing the same
storage-buffer loads as prosper's generated modules without being prosper's output. The ctest case
`vkprobe_hand_written_controls` is what proves they still assemble and pass `spirv-val`.

CMake registers it only when both `spirv-as` and `spirv-val` are found:

```cmake
find_program(SPIRV_AS_EXECUTABLE NAMES spirv-as)
find_program(SPIRV_VAL_EXECUTABLE NAMES spirv-val)
if(Python3_Interpreter_FOUND AND SPIRV_AS_EXECUTABLE AND SPIRV_VAL_EXECUTABLE)
  add_test(NAME vkprobe_hand_written_controls ...)
else()
  message(STATUS "vkprobe_hand_written_controls NOT registered: ...")
endif()
```

The `message(STATUS ...)` is honest but not load-bearing: nothing **fails** if the case disappears.
Both binaries come from the `spirv-tools` package that `.github/workflows/ci.yml` installs, so if
that install is ever dropped or renamed the case quietly stops being registered and the suite still
reports green — **a control nobody validates**, which is precisely the failure
`prosper/tools/vkprobe/AGENTS.md` exists to remember, and the same shape as #1711 (a `spirv-val`
gate that reported PASS when the validator was absent).

### Fix

The Linux job already guards the Vulkan-gated suite **by name** rather than by a count, for exactly
this reason. Add the same for this gate, right after that step:

```yaml
      - name: Verify the spirv-tools-gated control check is actually registered
        run: |
          if ! ctest --test-dir prosper/build-ci -N -R '^vkprobe_hand_written_controls$' | grep -q "Test *#"; then
            echo "::error::vkprobe_hand_written_controls is not registered - spirv-as/spirv-val were not found, so the hand-written SPIR-V controls in tools/vkprobe/shaders are unvalidated"
            exit 1
          fi
```

Verified locally: the block parses as YAML, the test is registered on this machine
(`Test #213: vkprobe_hand_written_controls ... Passed`), and the guard's mutation arm works —
removing `OpDecorate %Records Block` from a control makes the case fail with
`VUID-StandaloneSpirv-OpTypeRuntimeArray-04680`.

### Why it is filed rather than included in #2958

The change was written and validated on that branch and then **removed** before pushing: the agent's
token has no `workflow` scope, so GitHub refuses any push touching `.github/workflows/`. Raised as a
non-blocking review finding on #2958.

Refs #2945, #2958

mattias800:
Fixes #2960. Refs #2945, #2958, #1711.

## Failure scenario

`vkprobe_hand_written_controls` is registered by CMake only when **both** `spirv-as` and `spirv-val`
are found:

```cmake
if(Python3_Interpreter_FOUND AND SPIRV_AS_EXECUTABLE AND SPIRV_VAL_EXECUTABLE)
  add_test(NAME vkprobe_hand_written_controls ...)
else()
  message(STATUS "vkprobe_hand_written_controls NOT registered: ...")
endif()
```

That `message(STATUS ...)` is honest and **not load-bearing**: the configure succeeds, the case stops
existing, and the suite still reports green. Both binaries come from the `spirv-tools` package this
workflow installs, so the realistic way to lose it is that install being dropped or renamed — a
change that would leave CI green.

## Why this specific case is worth a job failure

What disappears is a **control**. `prosper/tools/vkprobe/shaders/` holds hand-written SPIR-V whose
entire value is being *valid by construction*: it performs the same storage-buffer loads as prosper's
generated modules **without being prosper's output**, which is exactly what let it decide #2945. A
control nobody validates is worse than no control, because it still reads as one — the same shape as
#1711, where a `spirv-val` gate reported PASS with no validator present.

## Evidence

Like-for-like: two configures differing **only** in `SPIRV_AS_EXECUTABLE`.

| configure | tests registered | cmake exit |
| --- | --- | --- |
| with `spirv-as` | **320** | 0 |
| `-DSPIRV_AS_EXECUTABLE=SPIRV_AS_EXECUTABLE-NOTFOUND` | **319** | 0 |

`ctest -N` name diff between the two: exactly one line, `vkprobe_hand_written_controls`.

The guard is verified **end to end**, not merely as shell logic — the second arm reconfigures so the
CMake gate genuinely drops the case, then runs the guard against that build directory:

| arm | CMake says | guard |
| --- | --- | --- |
| spirv-tools present | *(registered)* | `rc=0` |
| `spirv-as` forced NOTFOUND | `vkprobe_hand_written_controls NOT registered: python3=TRUE spirv-as=SPIRV_AS_EXECUTABLE-NOTFOUND spirv-val=/usr/bin/spirv-val` | `rc=1` |

I also checked the precondition that would make this PR break CI rather than guard it: the Linux job's
`Install dependencies` step does install `spirv-tools`, so the guard passes on current master.

## Approach and invariants

- **A separate step, not another entry in the `for t in ...` loop above.** That loop's error blames a
  missing Vulkan configuration and cites #1675. For a spirv-tools absence that is the wrong
  diagnosis, and a wrong error message sends the reader to the wrong fix. Different gate, different
  cause, different message.
- **By name, not by count**, for the reason the neighbouring comment already gives: the suite only
  grows, so a "at least N tests" floor is cleared by unrelated additions later and stops guarding
  without ever failing. Worth noting concretely — a naive comparison I ran first showed *319 vs 319*
  because the two build directories were not configured alike; the like-for-like numbers above are
  the ones to trust, and the near-miss is a fair illustration of why a count is the wrong instrument.
- Placed immediately after the Vulkan guard and before `Build`, matching it: `ctest -N` reads a
  configure-time property, so both are answerable at that point.

Affects the Linux CI job only. No source, no test, no build-system change — the case itself and the
CMake gate are untouched.

**Correction (review finding):** an earlier revision of this section said the Windows, macOS and MinGW
jobs "do not register this case". That is **falsified by this PR's own CI** — the macOS job log shows
`217/262 Test #217: vkprobe_hand_written_controls ... Passed`. Both those jobs install `spirv-tools`
and the CMake gate has no platform condition, so they register it too. Scoping the guard to Linux is
still the right call — one job asserting the case exists is enough to catch the package going away —
but the reason stated was wrong, and a wrong reason in a merged PR body is what the next person
inherits.

## Risk

If a future runner image drops `spirv-tools`, this job now fails instead of silently shrinking. That
is the intent, and the error message names the package and the directory so the fix is obvious.

## Note on why this was filed rather than fixed in #2958

The original author's token lacked `workflow` scope, so GitHub refused any push touching
`.github/workflows/`. Mine has it, so this lands the change as specified in the issue, with the
end-to-end verification added.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,6 +63,28 @@ jobs:
           echo "Vulkan-gated suite present; total registered: \
             $(ctest --test-dir prosper/build-ci -N | tail -1 | grep -oE '[0-9]+$')"
 
+      # Same defect class as the guard above, different gate — and deliberately a SEPARATE step so
+      # the error names the right cause. `vkprobe_hand_written_controls` is registered only when
+      # both `spirv-as` and `spirv-val` are found; CMake says so with `message(STATUS ...)`, which
+      # is honest and not load-bearing, because nothing fails when the case disappears.
+      #
+      # What is lost when it disappears is a CONTROL. `tools/vkprobe/shaders/` holds hand-written
+      # SPIR-V whose entire value is being valid by construction: it does the same storage-buffer
+      # loads as prosper's generated modules without being prosper's output, which is what let it
+      # decide #2945. An unvalidated control is worse than no control, because it still reads as
+      # one — the same shape as #1711, where a `spirv-val` gate reported PASS with no validator
+      # present.
+      #
+      # Both binaries come from the `spirv-tools` package this workflow installs, so the realistic
+      # way to lose it is that install being dropped or renamed — a change that would otherwise
+      # leave the suite green.
+      - name: Verify the spirv-tools-gated control check is actually registered
+        run: |
+          if ! ctest --test-dir prosper/build-ci -N -R '^vkprobe_hand_written_controls$' | grep -q "Test *#"; then
+            echo "::error::vkprobe_hand_written_controls is not registered — spirv-as/spirv-val were not found, so the hand-written SPIR-V controls in tools/vkprobe/shaders are unvalidated (see #2960)"
+            exit 1
+          fi
+
       - name: Build
         run: cmake --build prosper/build-ci
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,9 +75,13 @@ jobs:
       # one — the same shape as #1711, where a `spirv-val` gate reported PASS with no validator
       # present.
       #
-      # Both binaries come from the `spirv-tools` package this workflow installs, so the realistic
-      # way to lose it is that install being dropped or renamed — a change that would otherwise
-      # leave the suite green.
+      # What actually stays green is NARROWER than "the spirv-tools install is dropped", and the
+      # distinction matters because the broad version makes this guard look redundant. Dropping the
+      # whole package turns the job red anyway: `spv_validate` is registered unconditionally and
+      # hard-fails when `spirv-val` is missing (tools/spv_validate/spv_validate.cpp:306-315 -- a
+      # deliberate fix, since it once printed PASS with no validator). The cases this guard is the
+      # only cover for are the ones that leave `spirv-val` present: `spirv-as` alone disappearing
+      # or being renamed, `python3` absent, or the CMake gate itself being edited.
       - name: Verify the spirv-tools-gated control check is actually registered
         run: |
           if ! ctest --test-dir prosper/build-ci -N -R '^vkprobe_hand_written_controls$' | grep -q "Test *#"; then


mattias800 (commented):
Independent review of a64504f6 (adversarial second read; I am not the author).

## APPROVED

The mechanism is correct and I reproduced every claim you made, plus two arms you did not run. Two
statements in the PR need correcting before merge — one of them is in the committed comment, so it
is a code change, not just body prose. Neither affects behaviour.

---

## Reproduction — all three of your claims, independently

Fresh worktree at `a64504f6`, two fresh build dirs (never reused across arms), configured in the
ps5ys distrobox with the job's own flags, differing **only** in `SPIRV_AS_EXECUTABLE`:

| arm | cmake rc | `Total Tests:` | guard rc |
| --- | --- | --- | --- |
| normal configure | 0 | **320** | **0** |
| `-DSPIRV_AS_EXECUTABLE=SPIRV_AS_EXECUTABLE-NOTFOUND` | 0 | **319** | **1** |

Arm 2 printed exactly the STATUS line your body quotes:

```
-- vkprobe_hand_written_controls NOT registered: python3=TRUE spirv-as=SPIRV_AS_EXECUTABLE-NOTFOUND spirv-val=/usr/bin/spirv-val
```

`ctest -N` name diff between the two build dirs is exactly one line, `vkprobe_hand_written_controls`.
So the 320 -> 319 like-for-like measurement is confirmed, and so is the end-to-end guard behaviour:
the CMake gate really drops the case and the guard really catches it. Running the guard block
verbatim (only `--test-dir` substituted) emitted your `::error::` annotation and exited 1.

Two arms you did not run, both clean:

* **`-DSPIRV_VAL_EXECUTABLE=SPIRV_VAL_EXECUTABLE-NOTFOUND`** -> 319 registered, guard rc=1. The other
  half of the `AND` is covered too, so the guard is symmetric over the gate.
* **Guard against a nonexistent build dir** -> rc=1 (`Failed to change working directory`). It fails
  in the safe direction rather than passing on a broken invocation.

No false-positive risk in `grep -q "Test *#"`. Raw arm-2 output under `cat -A` is
`Test project <dir>$` / `$` / `Total Tests: 0$` — there is no `#` anywhere in the no-match output,
so the pattern cannot match by accident.

**Q4, step placement:** yes, `ctest -N` is genuinely answerable before `Build`. All four counts above
came from build dirs that were never built. Your own CI run settles it independently: Linux job
`99785158892` shows `##[group]Run if ! ctest --test-dir prosper/build-ci -N -R '^vkprobe_...` sitting
between the Vulkan guard and `Run cmake --build`, and the job passed.

**Q2, could it break CI:** confirmed, `.github/workflows/ci.yml:37` installs `spirv-tools`. I also
checked the one gate that could have reddened on a new `ctest` invocation:
`prosper/tools/ci/check_ctest_gate.py` excludes `-N` listings (`LISTING` regex at :86, rationale
at :37) and its own accept-fixture is literally the neighbouring guard's shape. `--root .` on your
branch: rc=0, "7 ctest invocation(s), all carrying --no-tests=error".

**Q3, separate step vs the loop — agree, and the argument is stronger than the one you made.** I
checked the nesting of the CMake gate rather than assuming it: on master, the
`if(Python3_Interpreter_FOUND AND SPIRV_AS_EXECUTABLE AND SPIRV_VAL_EXECUTABLE)` at
`prosper/CMakeLists.txt:1844` sits inside only `if(TARGET prosper_core)` (:535) — **not** inside a
`Vulkan_FOUND` block. Measured: a `-DCMAKE_DISABLE_FIND_PACKAGE_Vulkan=TRUE` configure still
registers the case (Test #226 of 280). So the two gates are genuinely orthogonal: the Vulkan loop's
error would be a wrong diagnosis, and folding this in would additionally make that step's trailing
`Vulkan-gated suite present` echo assert something it had not checked. Separate step is right.

---

## Finding 1 — the committed comment's failure scenario is over-claimed (required, one line)

`.github/workflows/ci.yml`, in the block this PR adds:

> Both binaries come from the `spirv-tools` package this workflow installs, so the realistic way to
> lose it is that install being dropped or renamed — a change that would otherwise leave the suite
> green.

**A whole-package drop does not leave the suite green.** `prosper/tools/spv_validate/spv_validate.cpp:304`
probes `spirv-val --version`, and at :306-:314 prints `== FAIL: spirv-val is not on PATH ==` and
`return 1`. `spv_validate` is registered unconditionally (`prosper/CMakeLists.txt:1829`). So removing
`spirv-tools` from :37 turns the job **red** at the Test step — which is exactly what the comment
above `Install dependencies` already says it does.

The guard is still justified; the sentence just names the wrong scenario. What genuinely stays green
is narrower, and it is precisely the arm you measured:

* `spirv-as` alone becoming unavailable (a package split, or the binary renamed) while `spirv-val`
  survives — `spv_validate` green, `vkprobe_hand_written_controls` silently gone;
* `Python3_Interpreter_FOUND` false;
* the gate condition or the `add_test` itself being edited or renamed.

Please reword the comment to that. The same sentence is in the PR body and in #2960's body — worth
correcting both, since the issue is where the next reader starts.

## Finding 2 — "macOS and MinGW do not register this case" is false (required, body only)

The PR body's *Affected / deliberately unaffected* says:

> Windows, macOS and MinGW jobs are unchanged; they do not register this case and the issue is
> Linux-specific.

Falsified from **this PR's own CI**. macOS job `99785158954`, log line 1401:

```
217/262 Test #217: vkprobe_hand_written_controls ..............   Passed    0.10 sec
```

Both the MinGW job (`ci.yml:505`, `mingw-w64-ucrt-x86_64-spirv-tools`) and the macOS job
(`ci.yml:638`, `brew install ninja spirv-tools`) install spirv-tools, and as established above the
CMake gate carries no platform and no Vulkan condition. So the case is registered and running there.

Scoping the guard to Linux is still fine — cross-platform parity is not a merge requirement — but
the stated reason is wrong, and it is the kind of confident line that later answers "why isn't this
guarded on macOS?" incorrectly. Say "guarded on Linux only; the other jobs register it too and could
carry the same guard later" or similar.

## Nits (not blocking)

* The `::error::` text says "spirv-as/spirv-val were not found", but the gate is
  `Python3_Interpreter_FOUND AND SPIRV_AS_EXECUTABLE AND SPIRV_VAL_EXECUTABLE`
  (`prosper/CMakeLists.txt:1844`). A python3-less configure would be misdiagnosed by one word. An
  "…or python3; CMake's STATUS line names which" clause closes it.
* The same message prints when `ctest` cannot open the build dir at all. It fails safe, so this is
  cosmetic.

## Housekeeping checked

`git diff --check` clean; no `Claude-Session:` trailer on the branch; diff is `.github/workflows/ci.yml`
only; MERGEABLE / CLEAN; all required checks green at this head.

**APPROVED** — land findings 1 (comment) and 2 (body) first; I do not need to re-review those.

mattias800:
Both review corrections landed — see the detailed reply on #3187, which covers this PR's two items in its closing section. Comment narrowed to the genuinely-uncovered cases (verified `spv_validate.cpp:306-315` hard-fails unconditionally), and the falsified "macOS and MinGW do not register this case" line corrected in the body against the macOS job's own log.

PR #217, opened 2026-07-09
fix(hle): sceKernelLoadStartModule returns ENOENT for a path not in the linked set

mattias800:
## Summary
- `LoadStartModule` returned a monotonically-increasing **fake success handle for any path**, loading nothing. A title runtime-loading a PRX not in the preload list believed the load succeeded, then called exports that resolved to ESRCH fallbacks / the wrong module instead of getting the honest ENOENT.
- Return `SCE_KERNEL_ERROR_ENOENT` for a genuine miss.
- Removing the fake-success fallback **surfaced two latent gaps it had been masking** for the game's *own* preloaded PRX (the Messenger runtime-`LoadStartModule`'s its Il2Cpp/PS5Util modules) — both fixed here so those loads resolve to real handles instead of newly-honest ENOENT:
  - `module_handle_for_path` now matches the basename **case-insensitively**: the guest loads `Il2CppUserAssemblies.prx` but the module is preloaded as `Il2cppUserAssemblies.prx` (PS5 paths are effectively case-insensitive).
  - `test_boot_linux` now registers the module export tables (it links directly, bypassing `boot_program`), so the deep boot actually exercises the real-handle path instead of relying on fake success.

## Verification
- **Caught a real boot regression** from the naive ENOENT-for-misses (the game's own PRX started missing) and fixed the root cause — the dump-backed boot now reaches **IL2CPP + GC stop-the-world + graphics init (2 GC raises)**, and a truly-non-preloaded path returns ENOENT.
- `test_dlsym_handle` covers the case-insensitive match + the ENOENT-for-unknown-path contract.
- Full suite in WSL build-linux **including the dump-backed boot: 66/66 passed**.

Fixes #146

PR #3187, opened 2026-09-01
fix(perf): warn about harness readback whenever it is a material share, not only when it wins classification

mattias800:
Split out of #3160 on review recommendation — it shares no code with the alias census, it is its own
judgement call, and it can merge independently.

## Failure scenario

`tools/perf/performance_capture_report.py` printed its harness-readback warning **only when readback
won the classification**. A capture where readback is large but under the 40% evidence threshold is
classified `inconclusive` and then prints a large readback number **with no warning attached** — the
exact shape that sends a reader optimising the harness instead of the title.

Measured on a Dragon Quest VII capture: `readback=977.7ms` against `compute=1134.0ms`, classified
`inconclusive`, note silent. GPU present was not adopted, so that 977.7 ms was `prosper-app` under
`SDL_VIDEODRIVER=offscreen` copying every scanout frame to the CPU — not the title's work, and the
second-largest number on the page.

## Why not a ranking trigger

"Is readback the largest component" was the first attempt and is **silent on exactly this capture**
(977.7 < 1134.0). The trigger is readback's *share of measured work* instead.

## The threshold, and its honest limits

I first shipped `readback_ms > 0.0` under a variable named `material_readback` — a name that
described the intent and a predicate that did not implement it. A **pre-existing test**
(`test_non_readback_verdict_carries_no_readback_note`) caught it, and that test was right: a
paragraph of warning attached to a 2 ms readback is noise on every report, and noise trains readers
to skip the note in the cases that matter.

Two arms now pin it from both sides. Swept rather than asserted (cache cleared per iteration, `-B`):

| threshold | 0.005 | 0.020 | **0.021** | **0.100** | **0.200** | 0.201 | 0.310 |
| --- | --- | --- | --- | --- | --- | --- | --- |
| suite rc | 1 | 1 | **0** | **0** | **0** | 1 | 1 |

**The pinned band is (0.020, 0.200] — a 10x range, and `0.10` is not itself pinned.** The code
comment now says so. This is the part I want a reader to know: someone who believes the constant is
pinned will never add the arm that would pin it.

An earlier revision of that comment claimed "a ~39% readback must warn (the motivating capture)".
That figure was never computed — the arm that exists is 31%. Corrected.

## Mutation evidence

| mutation | result |
| --- | --- |
| threshold → `0.0` (any nonzero readback — the bug) | `rc=1` |
| trigger → `classification == "readback"` only (the original narrow behaviour) | `rc=1` |
| baseline / restored | `rc=0` |

Both directions are load-bearing: the arms catch a trigger that is too loose *and* one that is too
narrow.

## The arms

- **A ~31% readback that lost classification still warns** — the motivating shape reproduced:
  readback 977.7 ms, compute 1134.0 ms, neither reaching 40% of the 3134 ms measured total, verdict
  `inconclusive`. Readback is deliberately *not* the largest component here, so this arm also stays
  red under a ranking trigger.
- **A 1%-of-total readback stays silent**, with the same verdict and the same GPU-present state, so
  the only variable is the share. Without this arm the threshold could drift back toward "any
  nonzero" unnoticed, because every other readback arm in the file has a large readback.

Affects only `readback_note` in the report's summary. Classification, reasons, pacing notes and every
other component are untouched; the two-case split between "GPU present adopted" (real work, chase it)
and "not adopted" (harness, ignore it) is unchanged, and still decides what the note *says*.

## Verification

`ctest --no-tests=error -R performance_capture_report_logic` → **1/1**, exit 0. Full suite green on
the branch this was split from (319 registered, 100% passed, exit 0).

A note on method: an intermediate run of the sweep reported a false `restored_rc=1` from a stale
`__pycache__`. The sweep above clears it per iteration and runs `python3 -B`; the band is unchanged
from the contaminated run, but the restore check only passes on the clean one.

### DIFF
--- a/prosper/tools/perf/performance_capture_report.py
+++ b/prosper/tools/perf/performance_capture_report.py
@@ -329,21 +329,27 @@ def summarize(records):
     # different instrument that happens to share the mechanism.
     readback_note = None
     # Warn whenever readback is a LEADING component, not only when it wins the classification.
-    # A capture where readback is the largest single component but sits under the evidence
-    # threshold gets classified "inconclusive" -- and then prints the biggest number on the page
-    # with no warning attached, which is the exact shape that sends a reader chasing the harness.
-    # Measured on a Dragon Quest capture: readback=977.7ms was the top component, classification
-    # was "inconclusive", and the note stayed silent.
-    # Ranking is the wrong trigger: on the capture that motivated this, readback was 977.7ms and
-    # merely SECOND to compute's 1134.0ms, so a "is it the max" test stayed silent on a number a
-    # reader would still have chased. If the frontend is copying scanout frames at all, say so.
-    material_readback = False
-    try:
-        _c = components if isinstance(components, dict) else {}
-        _rb = _c.get("readback")
-        material_readback = isinstance(_rb, (int, float)) and _rb > 0.0
-    except Exception:
-        material_readback = False
+    # A capture where readback is large but sits under the 40% evidence threshold is classified
+    # "inconclusive" -- and then prints a big number with no warning attached, which is the exact
+    # shape that sends a reader chasing the harness instead of the title. Measured on a Dragon
+    # Quest VII capture: readback=977.7ms against compute=1134.0ms, classified "inconclusive",
+    # note silent.
+    #
+    # Ranking is the wrong trigger and was the first attempt: a "is readback the max component"
+    # test stays silent at 977.7 vs 1134.0, which is precisely the case that misleads. So the
+    # trigger is readback's SHARE of measured work, with a threshold deliberately well below the
+    # 40% classification bar -- the whole point is to cover the gap under that bar.
+    #
+    # The threshold is a judgement call and both ends of it are pinned by tests: a ~39% readback
+    # must warn (the motivating capture), and a 2%-of-total readback must stay silent, because a
+    # paragraph of warning attached to a rounding-level number is noise on every report and trains
+    # readers to skip the note in the cases that matter. Any nonzero readback is NOT the right
+    # trigger for that reason.
+    READBACK_NOTE_MIN_SHARE = 0.10
+    readback_ms = classification_components.get("readback")
+    material_readback = (
+        isinstance(readback_ms, (int, float)) and measured_total > 0 and
+        readback_ms / measured_total >= READBACK_NOTE_MIN_SHARE)
     if classification == "readback" or material_readback:
         adopted = _gpu_present_adopted(post)
         if adopted is None:
--- a/prosper/tools/perf/performance_capture_report.py
+++ b/prosper/tools/perf/performance_capture_report.py
@@ -439,7 +439,7 @@ def summarize(records):
                 "GPU present WAS adopted, so scanout readback is skipped and this readback is real "
                 "work -- non-deferred colour-target readback, storage writeback, or a copy forced by "
                 "authoritative_readback (every ordered DMA copy sets it). Chase it rather than "
-                "dismissing it as a harness artifact")
+                "dismissing it as a harness artifact.")
 
     pacing_note = None
     if rates["guest_fps"] is not None and rates["rendered_fps"] is not None:
--- a/prosper/tools/perf/performance_capture_report.py
+++ b/prosper/tools/perf/performance_capture_report.py
@@ -227,6 +227,45 @@ def _compute_program_groups(records, limit=10, address_limit=8):
     }
 
 
+# Readback must reach this share of measured work before the harness-readback note is emitted.
+#
+# ANCHOR: one quarter of the 0.40 classification evidence bar. The note exists to cover the gap
+# UNDER that bar -- a readback large enough to mislead but too small to win the verdict -- so the
+# threshold belongs well below it, and a quarter is the round fraction that is.
+#
+# WHAT THE TESTS ACTUALLY PIN, measured per-arm rather than by exit code, because the two are not
+# the same question and reading only the exit code gets this wrong:
+#
+#   threshold     failing arm
+#   <= 0.020      test_non_readback_verdict_carries_no_readback_note   (2 ms / 2%, PRE-EXISTING)
+#   0.021-0.200   none
+#   >  0.200      test_readback_note_threshold_is_a_share_not_a_ranking (its 20% assertion)
+#
+# So the pinned band is (0.020, 0.200] and 0.10 is NOT itself pinned -- it can move anywhere in a
+# 10x range with no test objecting. Two corrections worth keeping, because the obvious reading of
+# the arms is wrong in both directions:
+#   * the LOWER bound is held by the pre-existing 2% test, NOT by the 1% arm added alongside this
+#     threshold. That arm passes at 0.011 and at 0.020; it never binds.
+#   * the UPPER bound is held by the 20% assertion, NOT by the 31% motivating arm. That arm's share
+#     is 0.3120 and it does not fail anywhere in the swept range.
+# The 1% and 31% arms still earn their place -- they hold verdict and GPU-present state constant so
+# the only variable is share, and the 31% one reproduces the capture that motivated the change --
+# but neither is what fixes the band, and a comment claiming otherwise sends the next reader to the
+# wrong test when they want to tighten it.
+READBACK_NOTE_MIN_SHARE = 0.10
+
+
+# The note's closing advice depends on whether readback DECIDED the verdict. "before acting on this
+# verdict" is right when the verdict IS readback; on a capture whose verdict is a decisive compute or
+# renderer-resource share it is wrong advice attached to a correct conclusion, which is worse than no
+# advice -- the reader is told to discard a finding the readback had no part in.
+def _readback_note_tail(classification):
+    if classification == "readback":
+        return " before acting on this verdict"
+    return (" before reading the readback line below; it does not affect the "
+            + str(classification) + " verdict above")
+
+
 def summarize(records):
     header = records[0]
     footer = next(record for record in records if record.get("type") == "footer")
@@ -340,14 +379,7 @@ def summarize(records):
     # trigger is readback's SHARE of measured work, with a threshold deliberately well below the
     # 40% classification bar -- the whole point is to cover the gap under that bar.
     #
-    # The threshold is a judgement call, and what the tests pin is a BAND, not this constant: a
-    # 31%-of-measured-work readback must warn and a 1% one must stay silent, so any value in
-    # (0.02, 0.20] passes. 0.10 sits in that band and is not itself pinned -- a later change may
-    # move it within the band without any test noticing, which is the honest limit of the coverage.
-    # The quiet end is the one that earns its keep: a paragraph of warning attached to a
-    # rounding-level number is noise on every report and trains readers to skip the note in the
-    # cases that matter. Any nonzero readback is NOT the right trigger, for that reason.
-    READBACK_NOTE_MIN_SHARE = 0.10
+    # See READBACK_NOTE_MIN_SHARE at module scope for the threshold and what pins it.
     readback_ms = classification_components.get("readback")
     material_readback = (
         isinstance(readback_ms, (int, float)) and measured_total > 0 and
@@ -358,13 +390,13 @@ def summarize(records):
             readback_note = (
                 "this capture cannot say whether GPU present was adopted, so it cannot say whether "
                 "the readback is real or the harness copying every scanout frame to the CPU. Check "
-                "the run log for a GPU-present surface failure before acting on this verdict")
+                "the run log for a GPU-present surface failure" + _readback_note_tail(classification))
         elif adopted is False:
             readback_note = (
                 "GPU present was NOT adopted for this capture, so the frontend copied every scanout "
                 "frame to the CPU -- most often prosper-app under SDL_VIDEODRIVER=offscreen, which "
                 "needs VK_EXT_headless_surface. This readback is the harness, not the title. "
-                "Re-measure through a real window before acting on this verdict")
+                "Re-measure through a real window" + _readback_note_tail(classification))
         else:
             readback_note = (
                 "GPU present WAS adopted, so scanout readback is skipped and this readback is real "
--- a/prosper/tools/perf/performance_capture_report.py
+++ b/prosper/tools/perf/performance_capture_report.py
@@ -340,11 +340,13 @@ def summarize(records):
     # trigger is readback's SHARE of measured work, with a threshold deliberately well below the
     # 40% classification bar -- the whole point is to cover the gap under that bar.
     #
-    # The threshold is a judgement call and both ends of it are pinned by tests: a ~39% readback
-    # must warn (the motivating capture), and a 2%-of-total readback must stay silent, because a
-    # paragraph of warning attached to a rounding-level number is noise on every report and trains
-    # readers to skip the note in the cases that matter. Any nonzero readback is NOT the right
-    # trigger for that reason.
+    # The threshold is a judgement call, and what the tests pin is a BAND, not this constant: a
+    # 31%-of-measured-work readback must warn and a 1% one must stay silent, so any value in
+    # (0.02, 0.20] passes. 0.10 sits in that band and is not itself pinned -- a later change may
+    # move it within the band without any test noticing, which is the honest limit of the coverage.
+    # The quiet end is the one that earns its keep: a paragraph of warning attached to a
+    # rounding-level number is noise on every report and trains readers to skip the note in the
+    # cases that matter. Any nonzero readback is NOT the right trigger, for that reason.
     READBACK_NOTE_MIN_SHARE = 0.10
     readback_ms = classification_components.get("readback")
     material_readback = (
--- a/prosper/tools/perf/performance_capture_report.py
+++ b/prosper/tools/perf/performance_capture_report.py
@@ -328,7 +328,23 @@ def summarize(records):
     # prosper-app arms one. The `PROSPER_TILECENSUS` readback trap (instrument trap 237) is a
     # different instrument that happens to share the mechanism.
     readback_note = None
-    if classification == "readback":
+    # Warn whenever readback is a LEADING component, not only when it wins the classification.
+    # A capture where readback is the largest single component but sits under the evidence
+    # threshold gets classified "inconclusive" -- and then prints the biggest number on the page
+    # with no warning attached, which is the exact shape that sends a reader chasing the harness.
+    # Measured on a Dragon Quest capture: readback=977.7ms was the top component, classification
+    # was "inconclusive", and the note stayed silent.
+    # Ranking is the wrong trigger: on the capture that motivated this, readback was 977.7ms and
+    # merely SECOND to compute's 1134.0ms, so a "is it the max" test stayed silent on a number a
+    # reader would still have chased. If the frontend is copying scanout frames at all, say so.
+    material_readback = False
+    try:
+        _c = components if isinstance(components, dict) else {}
+        _rb = _c.get("readback")
+        material_readback = isinstance(_rb, (int, float)) and _rb > 0.0
+    except Exception:
+        material_readback = False
+    if classification == "readback" or material_readback:
         adopted = _gpu_present_adopted(post)
         if adopted is None:
             readback_note = (
--- a/prosper/tools/perf/performance_capture_report.py
+++ b/prosper/tools/perf/performance_capture_report.py
@@ -227,11 +227,21 @@ def _compute_program_groups(records, limit=10, address_limit=8):
     }
 
 
+# The share a component must reach to WIN the classification. Named because two things depend on
+# it and it was previously a bare literal in one of them, so neither the dependency nor the value
+# was pinned by anything -- moving it passed the whole suite.
+CLASSIFICATION_EVIDENCE_SHARE = 0.40
+
 # Readback must reach this share of measured work before the harness-readback note is emitted.
 #
-# ANCHOR: one quarter of the 0.40 classification evidence bar. The note exists to cover the gap
-# UNDER that bar -- a readback large enough to mislead but too small to win the verdict -- so the
-# threshold belongs well below it, and a quarter is the round fraction that is.
+# It sits BELOW the evidence bar by construction: the note exists to cover the gap under that bar,
+# a readback large enough to mislead but too small to win the verdict. That inequality is the real
+# invariant and is what the tests assert.
+#
+# One quarter of the bar is a CHOSEN fraction, not a derived one -- it came out of review as a
+# rationalisation for 0.10 rather than from anything about the data, and pinning it as an equality
+# would give a reviewer's off-hand ratio the authority of an invariant. Recorded as the reason the
+# value is what it is; the tests deliberately do not enforce the ratio.
 #
 # WHAT THE TESTS ACTUALLY PIN, measured per-arm rather than by exit code, because the two are not
 # the same question and reading only the exit code gets this wrong:
@@ -252,18 +262,36 @@ def _compute_program_groups(records, limit=10, address_limit=8):
 # the only variable is share, and the 31% one reproduces the capture that motivated the change --
 # but neither is what fixes the band, and a comment claiming otherwise sends the next reader to the
 # wrong test when they want to tighten it.
-READBACK_NOTE_MIN_SHARE = 0.10
+READBACK_NOTE_MIN_SHARE = CLASSIFICATION_EVIDENCE_SHARE / 4
 
 
 # The note's closing advice depends on whether readback DECIDED the verdict. "before acting on this
 # verdict" is right when the verdict IS readback; on a capture whose verdict is a decisive compute or
 # renderer-resource share it is wrong advice attached to a correct conclusion, which is worse than no
 # advice -- the reader is told to discard a finding the readback had no part in.
+# Verdicts that CANNOT flip when the readback is removed. Removing readback scales every other
+# component's share up by the same factor, so a component that already won still wins. A verdict
+# that exists precisely BECAUSE nothing won ("inconclusive"), or because measured work was small
+# against the wall window ("cpu-outside-renderer"), can and does flip -- measured on this change's
+# own motivating capture, where dropping the harness readback turns "inconclusive" into "compute
+# (1134.0 ms, 53%)". This module already records the same effect from hardware: #3152 saw a verdict
+# become renderer-resource with the graphics total halved once the capture ran through a real window.
+_VERDICTS_READBACK_CANNOT_FLIP = frozenset({
+    "compute", "renderer-resource", "gpu-device", "gpu-wait", "gpu-wait-overhead",
+})
+
+
 def _readback_note_tail(classification):
     if classification == "readback":
-        return " before acting on this verdict"
-    return (" before reading the readback line below; it does not affect the "
-            + str(classification) + " verdict above")
+        return " before acting on this verdict."
+    if classification in _VERDICTS_READBACK_CANNOT_FLIP:
+        # Safe to reassure: this verdict survives the readback being removed.
+        return (f" The {classification} verdict above does not depend on it: removing the readback "
+                "raises every other share, so a component that already won still wins.")
+    # Not safe. Saying "the verdict is unaffected" here would invite trust in a verdict the harness
+    # may have produced, which is the opposite of what this note exists to prevent.
+    return (f" The {classification} verdict above may itself be an artefact of that readback -- it "
+            "is not a verdict some component won, so removing the readback can change it.")
 
 
 def summarize(records):
@@ -333,7 +361,7 @@ def summarize(records):
     if measured_total > 0:
         largest, cost = max(classification_components.items(), key=lambda item: item[1])
         share = cost / measured_total
-        if share >= 0.40:
+        if share >= CLASSIFICATION_EVIDENCE_SHARE:
             classification = largest
             reason = f"{largest} is the largest measured component ({cost:.1f} ms, {share:.0%})"
         elif cpu_cores is not None and cpu_cores >= 0.80 and wall_ms and measured_total < wall_ms * 0.40:
@@ -341,7 +369,8 @@ def summarize(records):
             reason = (f"the process used {cpu_cores:.2f} CPU cores while measured renderer/compute "
                       f"work covered only {measured_total / wall_ms:.0%} of the sampled wall window")
         else:
-            reason = "measured work is split across components; no component reaches the 40% evidence threshold"
+            reason = ("measured work is split across components; no component reaches the "
+                      f"{CLASSIFICATION_EVIDENCE_SHARE:.0%} evidence threshold")
     elif cpu_cores is not None and cpu_cores >= 0.80:
         classification = "cpu-outside-renderer"
         reason = (f"the process used {cpu_cores:.2f} CPU cores with no retained renderer/compute "
@@ -390,13 +419,13 @@ def summarize(records):
             readback_note = (
                 "this capture cannot say whether GPU present was adopted, so it cannot say whether "
                 "the readback is real or the harness copying every scanout frame to the CPU. Check "
-                "the run log for a GPU-present surface failure" + _readback_note_tail(classification))
+                "the run log for a GPU-present surface failure." + _readback_note_tail(classification))
         elif adopted is False:
             readback_note = (
                 "GPU present was NOT adopted for this capture, so the frontend copied every scanout "
                 "frame to the CPU -- most often prosper-app under SDL_VIDEODRIVER=offscreen, which "
                 "needs VK_EXT_headless_surface. This readback is the harness, not the title. "
-                "Re-measure through a real window" + _readback_note_tail(classification))
+                "Re-measure through a real window." + _readback_note_tail(classification))
         else:
             readback_note = (
                 "GPU present WAS adopted, so scanout readback is skipped and this readback is real "
--- a/prosper/tools/perf/performance_capture_report.py
+++ b/prosper/tools/perf/performance_capture_report.py
@@ -262,6 +262,10 @@ CLASSIFICATION_EVIDENCE_SHARE = 0.40
 # the only variable is share, and the 31% one reproduces the capture that motivated the change --
 # but neither is what fixes the band, and a comment claiming otherwise sends the next reader to the
 # wrong test when they want to tighten it.
+# Deriving this FROM the bar means a bar mutation is a COMPOUND mutation: the threshold moves with
+# it, so at a bar of 0.80 or more the threshold leaves the (0.020, 0.200] band its own arms pin and
+# readback tests start failing for reasons that have nothing to do with the bar. Harmless for any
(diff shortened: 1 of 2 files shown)


mattias800 (commented):
Independent review of 8bf2b6e2 (adversarial second read; I am not the author).

## APPROVED

The trigger change is correct, both mutation arms reproduce, and your rejection of a ranking trigger
holds up under evaluation rather than argument. But I swept the threshold and recorded *which* arm
fails at each point, and **the band derivation stated in the committed comment is wrong at both
ends** — the conclusion is right, the route to it is not. That plus a second false justification in
a committed test comment are the required fixes. Both are prose in comments; neither changes
behaviour.

Method note, since you flagged trap 239 at me: every arm below ran under `python3 -B` **and** with an
explicit `rm -rf <scratch>/__pycache__` (full path, no glob) between arms, in a scratch copy of both
files; after the last restore the module was verified byte-identical to the branch (`diff` clean).
Baseline and restored both 21 tests, rc=0.

---

## Q3 — mutation arms reproduce exactly

| arm | rc | failing |
| --- | --- | --- |
| baseline (21 tests) | **0** | — |
| threshold -> `0.0` | **1** | `test_non_readback_verdict_carries_no_readback_note`, `test_readback_note_threshold_is_a_share_not_a_ranking` |
| trigger -> `classification == "readback"` only | **1** | `test_large_sub_threshold_readback_warns_...` (ERROR), `test_readback_note_threshold_is_a_share_not_a_ranking` (FAIL) |
| restored | **0** | — |

Also ran it the way ctest does (`python3 test_performance_capture_report.py`, the
`performance_capture_report_logic` command): 21 tests, OK, rc=0. Both directions are load-bearing.
Confirmed.

Small note on the second row: `test_large_sub_threshold_...` fails as an **ERROR**, not a FAIL —
`assertIn(..., None)` raises `TypeError` rather than reporting a clean assertion failure. It still
reddens, so the arm works; `assertIsNotNone` before the `assertIn` would make the failure message say
what actually went wrong.

## Q2 — the ranking-trigger rejection is right, and there is a better argument for the share

Verified by evaluating it, not by reading the body. For the motivating shape,
`classification_components` is `{renderer-resource: 0, readback: 977.7, compute: 1134.0,
gpu-wait: 0}`; max is compute at 36% of the 3134 ms measured total, under the 40% bar, so the verdict
is `inconclusive` and a "readback is the max component" trigger is silent on exactly the capture that
misleads. Your reasoning holds.

The argument you did **not** make is stronger than the one you did: the share trigger reuses the same
denominator as the 40% classification bar (`measured_total = graphics_total + compute_total`,
`performance_capture_report.py:256`), and `readback_ms` is a genuine sub-bucket of the renderer
record's `total_ms` (`frontends/shared/perf/performance_capture.hpp:37` and `:44`, assigned at
`live_renderer.cpp:8277`/`:8285`). So the share is well-formed in [0,1] and directly commensurate
with the bar it is designed to sit under. That is why "share" is the right axis here and ranking is
not — worth a line in the comment.

---

## Finding 1 — the stated band derivation does not produce the stated band (required)

I reproduced your sweep and additionally recorded which arm fails at each threshold:

| threshold | 0.005 | 0.010 | 0.011 | 0.020 | 0.021 | 0.100 | 0.200 | 0.201 | 0.310 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| suite rc | 1 | 1 | 1 | 1 | **0** | **0** | **0** | 1 | 1 |
| failing arm | both | both | `non_readback_verdict` | `non_readback_verdict` | — | — | — | `share_not_ranking` | `share_not_ranking` |

Your band `(0.020, 0.200]` reproduces exactly. But `performance_capture_report.py:344` says:

> a 31%-of-measured-work readback must warn and a 1% one must stay silent, so any value in
> (0.02, 0.20] passes.

Neither named arm produces either endpoint:

* **The lower bound (>0.020) is bound solely by the pre-existing
  `test_non_readback_verdict_carries_no_readback_note`** — renderer `total_ms=100, readback_ms=2`
  (`test_performance_capture_report.py:218`), i.e. **2%**, and `>=` is what makes exactly 0.020 fail.
  The new 1% arm passes at 0.011 and at 0.020; it is never the binding constraint anywhere near the
  boundary. "A 1% one must stay silent" yields `(0.01, ...]`, not `(0.02, ...]`.
* **The upper bound (<=0.200) is bound solely by the new arm's 20% case**
  (`test_performance_capture_report.py:212`). The 31% motivating arm never fails anywhere in the
  sweep — its share is 977.7/3134 = 0.3120, so it only breaks above ~0.312.

The band is right and the derivation is not, which is the pattern the charter names: a true statement
reached by a route that does not establish it. It matters here more than usual, because that comment's
entire job is to tell the next reader *what pins the band* — someone who trusts it will protect the
wrong two assertions and can loosen the one that actually binds without noticing. Please restate it
as: lower bound from `test_non_readback_verdict_carries_no_readback_note`'s 2% arm, upper bound from
the 20% assertion.

## Finding 2 — the new quiet arm's justification is false and contradicts your own PR body (required)

`test_performance_capture_report.py:204`:

> Without this pair the threshold could drift to "any nonzero readback" -- which it briefly was -- and
> no test would notice, because every existing readback arm has a large readback.

Measured: with the threshold mutated to `0.0`, `test_non_readback_verdict_carries_no_readback_note`
fails **on its own** (its readback is 2 ms = 2%). A test does notice — and your PR body says so
explicitly: *"A pre-existing test (`test_non_readback_verdict_carries_no_readback_note`) caught it,
and that test was right."* Body and comment contradict each other, and the comment is what gets read
in a year.

Keep the arm — but for the reason it actually has and does not state: it holds the verdict and the
GPU-present state constant across the loud/quiet pair, so **share** is the only variable, which the
pre-existing arm cannot do (its verdict differs, so it would also pass if the note were still gated on
the verdict). That is a real methodological addition. "No test would notice" is not.

## Finding 3 — the widened trigger attaches verdict-directed advice to non-readback verdicts (recommended)

All three note bodies end in "…before acting on this verdict", and `performance_capture_report.py:498`
prints the verdict line with `:500` printing the note directly beneath it. Constructed and run
(renderer `total_ms=300, readback_ms=250`, compute `1700`):

```
primary evidence: compute — compute is the largest measured component (1700.0 ms, 85%)
readback: GPU present was NOT adopted ... This readback is the harness, not the title.
          Re-measure through a real window before acting on this verdict
```

Readback is 12.5%, so the note fires; the verdict is compute at 85%, and re-measuring through a real
window will not change it. The note's *body* stays true — that is the good half of the widening — but
its closing clause is now wrong advice for that report, and it is printed on the line under a verdict
it no longer refers to. No test asserts that clause (the arms key on "harness, not the title",
"real", "cannot say"), so rewording it to point at *this readback number* rather than *this verdict*
is free. Your "Affected / deliberately unaffected" section says the two-case split still decides what
the note says; it does not mention that the note's text presumes a readback verdict.

Same shape, smaller: `test_no_post_samples_cannot_determine_gpu_present`'s defensive
`else: assertIsNone(summary["readback_note"])` is now unreachable **and false** — that capture's
readback is 60% of measured work, so if a future change ever stopped classifying it as `readback` the
note would still fire and the arm would fail on the wrong assertion. Its author wrote that branch
deliberately for a future reclassification; this PR silently voided it without saying so.

---

## Q1 — is admitting "0.10 is not pinned" sufficient?

Admitting is necessary and I would not want it removed, but as written the admission is undermined by
finding 1: it misattributes the very band it is being honest about. Fix that and the admission does
its job.

Beyond that, one cheap structural thing the PR could do. `READBACK_NOTE_MIN_SHARE` is declared
**inside** `summarize()` (`performance_capture_report.py:350`, four-space indent), so no test can
import it — which makes "not itself pinned" structural rather than incidental. Promote it to module
scope and a later arm can pin the boundary directly (`share == MIN` warns, one epsilon below stays
silent) without hard-coding `0.10` in the test. Not required for this merge; it is the change that
lets the next person close the gap you correctly flagged instead of re-flagging it.

I do **not** think the constant should move. `0.10` has a principled anchor that is nowhere stated: it
is one quarter of the 40% evidence bar, i.e. the note fires on anything within a factor of four of
being the verdict. One sentence saying that turns it from "a number that passes two tests" into a
number with a reason.

## Housekeeping checked

`git diff --check` clean; no `Claude-Session:` trailer on the three commits; `readback_note` has no
consumer outside this module and its tests; MERGEABLE / CLEAN; all required checks green at this head.

**APPROVED** — land findings 1 and 2 (both committed-comment prose) before merging; finding 3 is a
recommendation I will not block on. No re-review needed for comment-only corrections.

mattias800:
Both findings confirmed and fixed on `f2e9c745`. The band correction is the valuable one, and I want
to be precise about why my version was wrong: **I derived it from exit codes, which cannot tell you
*which* arm binds.** Re-swept recording the failing arm name per threshold, and your attribution is
exactly right:

| threshold | failing arm |
| --- | --- |
| ≤ 0.020 | `test_non_readback_verdict_carries_no_readback_note` — the **pre-existing** 2% test |
| 0.021 – 0.200 | none |
| > 0.200 | `test_readback_note_threshold_is_a_share_not_a_ranking` (its 20% assertion) |

So the 1% arm I added never binds — it passes at 0.011 — and the 31% motivating arm (share 0.3120)
**does not fail anywhere in the swept range**. My comment credited both with holding a band neither
holds.

*(Method note: my first attempt at the per-arm sweep printed "none" at every threshold, because
unittest's colorized output puts ANSI escapes between `FAIL` and `:` and my grep never matched. Same
family as trap 240 — a negative from an instrument I hadn't checked. Stripping escapes gave the table
above.)*

I also fixed the related claim you flagged: the comment said a drift to any-nonzero would go
unnoticed. Measured false — the pre-existing 2% arm catches it, which **my own PR body said while
the comment denied it**.

**On your recommendation, I did more than correct prose, because you identified the real problem:
`READBACK_NOTE_MIN_SHARE` was declared inside `summarize()`, so no test could import it and the
constant was unpinnable by construction.** Promoted to module scope, and now actually pinned:

- The new arm asserts the **anchor**, not the literal: `READBACK_NOTE_MIN_SHARE == 0.40 / 4`. Your
  observation that 0.10 is one quarter of the evidence bar is the rationale the code never stated,
  and it is the right thing to pin — a bare `== 0.10` would pass while the relationship it stands
  for broke.
- Mutation proving it closed the gap: **moving the constant 0.10 → 0.15 now reddens**. That change
  sits inside the old band and was previously invisible.

**Your non-blocking wart is fixed too, and you were right that it was worth fixing.** The note closed
"before acting on this verdict" on every report, so a decisive 80% compute verdict with a 12%
readback told the reader to discard a conclusion the readback had no part in — wrong advice attached
to a correct finding, which is worse than no advice. The tail is now verdict-aware; a readback
*verdict* still says exactly that, because there it is correct. New arm covers both directions, and
mutating the tail to always say "this verdict" reddens it.

Writing that arm caught my own arithmetic: my first draft used `readback_ms=150` against a 5000 ms
measured total — 3%, under the threshold, so no note was emitted and the arm failed for a reason
unrelated to what it tests. Now 600/5000 = 12%, over the threshold, against a compute verdict at 80%.

23/23 pass. Mutations (`python3 -B`, `__pycache__` cleared per arm per trap 239):

| mutation | failing arms |
| --- | --- |
| constant `0.10` → `0.15` | anchor arm + tail arm |
| tail always `"this verdict"` | tail arm |
| baseline / restored | none |

On **#3186**: both your corrections are landed too. The comment now states the narrow truth — a
whole-package drop reddens the job anyway via `spv_validate`'s unconditional hard-fail
(`spv_validate.cpp:306-315`, which I verified), so what this guard uniquely covers is `spirv-as`
alone going missing, `python3` absent, or the gate being edited. That was always the arm I actually
measured; only the prose was broad. And the PR body's "macOS and MinGW do not register this case" is
corrected — your citation of the macOS job's own log (`Test #217 ... Passed`) falsifies it, and
Linux-only scoping now stands on the right reason.

Your extra arms are worth more than the ones I ran: `SPIRV_VAL_EXECUTABLE` forced NOTFOUND, the
nonexistent-build-dir fail-safe check, `grep -q "Test *#"` being unable to false-positive, and
especially finding that the CMake gate sits inside `if(TARGET prosper_core)` rather than
`Vulkan_FOUND` — so a Vulkan-disabled configure still registers the case, making the two gates
genuinely orthogonal. That is a better argument for the separate step than the one I gave.

mattias800 (commented):
Re-review of `f2e9c745` (my earlier approval was on `8bf2b6e2` and does not carry).

## REJECTED

Both new behaviours are the right idea and I asked for both. But each one **asserts something that
measurement contradicts**, and in both cases the contradicting measurement is one mutation away.
Findings 1 and 3 are blocking; neither is large. Everything else below is clean, including the
promotion, which is exactly right.

Method: fresh worktree at `f2e9c745`, scratch copy of both files, `python3 -B` plus explicit
`rm -rf <scratch>/__pycache__` between every arm, module verified byte-identical to the branch after
the last restore. Baseline 23/23 rc=0, and green again run the way ctest runs it.

Your ANSI note is a good catch and the same family as the one it cost me nothing to avoid only
because I stripped escapes with `sed 's/\x1b\[[0-9;]*m//g'` before grepping. Worth a line in the
comment block, since the next person sweeping this file will reach for the same grep.

---

## Finding 1 (BLOCKING) — the anchor arm cannot see the bar move, and nothing else pins the bar either

`test_performance_capture_report.py:225` claims the arm exists so that *"a future change to that bar
carries this along instead of silently decoupling from it."* Measured — I moved the bar and reran:

```
performance_capture_report.py:336   if share >= 0.40:   ->   if share >= 0.50:
  anchor arm alone : rc=0
  whole suite      : rc=0, 23/23, no failures
```

The classification evidence bar is a **bare inline literal at :336**. There is no constant. So the
`0.40 / 4` at `:227` and the `0.40` at `:228` are a second, disconnected copy of the number, and the
arm pins `READBACK_NOTE_MIN_SHARE == 0.1` in a longer notation. The relationship it stands for is
precisely the thing it cannot observe.

Your reported mutation (`0.10 -> 0.15` reddens) reproduces — but it is the same-source positive
control the charter's worked example is about. It shows the assertion fires when the **constant**
moves, which was never in doubt; it cannot show it fires when the **bar** moves, which is the claim
being made. The arm that would have discriminated is the one above, and it comes back green.

Second-order, and it is the more valuable half: **nothing in the suite pins the 0.40 bar at all.** A
change to this tool's primary classification threshold passes 23/23. Naming it fixes both problems at
once:

```python
CLASSIFICATION_EVIDENCE_SHARE = 0.40          # module scope, used at :336
READBACK_NOTE_MIN_SHARE = CLASSIFICATION_EVIDENCE_SHARE / 4
...
self.assertEqual(READBACK_NOTE_MIN_SHARE, CLASSIFICATION_EVIDENCE_SHARE / 4)
```
plus the mutation arm that moves the **bar** and shows the anchor arm reddening. (Note `:339` carries
a *different* `0.40` — the wall-window fraction for `cpu-outside-renderer` — so "the 0.40 bar" is
ambiguous today for whoever tries to name it. Name only the one at `:336`.)

## Finding 2 — your judgement-call question: the ratio over-couples, and I should declare an interest

**The 1/4 anchor is mine, and I offered it as a rationalisation, not as a derived invariant.**
Promoting an off-hand reviewer suggestion into a hard equality is the asymmetry the charter records:
a correction that *raises* a claim gets checked least, and one arriving from a reviewer with a tidy
arithmetic relationship attached is the most authority-laden form it can take. I would rather you
did not take it at that strength on my say-so.

What is genuinely defensible — because the code comment already argues it and the behaviour depends
on it — is the **inequality**, not the ratio: the note exists to cover the gap *under* the bar, so
`0 < READBACK_NOTE_MIN_SHARE < CLASSIFICATION_EVIDENCE_SHARE`. That survives anyone legitimately
retuning either number, and it is the property the design actually rests on.

If you keep the equality, keep it with the coupling made real (finding 1) **and** say in the comment
that one quarter is a chosen round fraction rather than a derived one. Otherwise the next person who
wants 0.15 for a good reason unrelated to the bar has two options: delete the arm, or move the bar to
satisfy a note threshold. The second is much worse than the problem the arm was added to solve.

## Finding 3 (BLOCKING) — the new tail prints a false statement on this PR's own motivating capture

`_readback_note_tail` (`:262`) closes every non-readback report with *"it does not affect the
&lt;classification&gt; verdict above"*. I checked that claim by removing the harness readback and
re-classifying:

| verdict | as captured | harness readback removed | claim |
| --- | --- | --- | --- |
| `inconclusive` — **the DQ7 capture this PR exists for** | inconclusive | **compute (1134.0 ms, 53%)** | **FALSE** |
| `cpu-outside-renderer` | cpu-outside-renderer | **renderer-resource (45%)** | **FALSE** |
| `compute`, decisive — *the only case the new arm exercises* | compute (80%) | compute (91%) | true |

The structure explains it, and it is worth writing down because it tells you which branches are safe:
removing readback scales every other component's share up by the same factor, so a component that
already **won** keeps winning with a larger share — the claim is sound for `compute`,
`renderer-resource`, `gpu-device` and `gpu-wait*`. But a verdict that exists *because nothing won*
(`inconclusive`) or *because measured work was small against the wall window*
(`cpu-outside-renderer`) can flip precisely when the readback is removed. Those are the two branches
you said you had not exercised, and they are the two that break.

This is not a hypothetical: your module already records the same effect as a hardware measurement.
`performance_capture_report.py:359`, on The Forgotten City (#3152) — readback *"396.9 ms / 51% ...
offscreen, against 0.0 ms through a real window, **where the verdict became renderer-resource and the
graphics total halved**."* The new function, a hundred lines below that comment, tells the reader the
opposite by default.

And the direction of the error is the bad one. The note's job is "this big number is the harness". On
the DQ7 capture the honest consequence is that the **verdict itself is contaminated** — strip the
harness readback and it is a compute capture, not an inconclusive one. Telling the reader the
readback "does not affect the inconclusive verdict above" invites them to trust a verdict the harness
produced, which is a sharper version of the failure the whole PR is fixing.

Narrow it to the branches where it holds. Something like: emit the disclaiming tail only when
`classification` is one of the decisive component verdicts, and for `inconclusive` /
`cpu-outside-renderer` say the true and more useful thing — that a harness readback of this size may
be what produced the verdict, so re-measure before trusting it.

## Finding 4 — the tail's wording is incoherent independently of its truth

"the readback line below" **is the line it is printed on**. `print_summary` emits `primary evidence:`
at `:530` and `readback: <note>` at `:532`, and nothing readback-related follows (only `pacing:`).
Rendered on a compute verdict:

```
primary evidence: compute — compute is the largest measured component (4000.0 ms, 80%)
readback: ... This readback is the harness, not the title. Re-measure through a real window before
          reading the readback line below; it does not affect the compute verdict above
```

"Re-measure through a real window before reading the readback line below" does not parse as advice.
The old tail worked because it was a subordinate clause that completed the verb — *"Re-measure ...
before acting on this verdict"*. The replacement is a different thought and needs to be a separate
sentence, not a trailing clause. Same on the `adopted is None` branch: *"Check the run log for a
GPU-present surface failure before reading the readback line below"*.

(The third branch, GPU-present-adopted, correctly has no tail — it never carried the wart. Checked;
no action needed.)

## Finding 5 — the tail arm's fixture has two points of headroom, and you have hit this once already

`test_performance_capture_report.py:235`: readback 600 / 5000 measured = **12%** against a 10%
threshold. Confirmed by mutation: raising the constant to 0.15 reddens the **tail** arm as well as the
anchor arm — i.e. the tail arm fails for a threshold reason rather than a tail reason, which is
exactly the defect you caught in your own first draft at 3% and then re-introduced with a smaller
margin. A decisive-compute fixture with real headroom exists: renderer `total_ms=1500,
readback_ms=1400`, compute `3500` — measured 5000, compute 70% (decisive), readback 28%.

## Finding 6 — the promotion is clean (no action)

Exactly one assignment (`:255`, module scope), one read (`:386`), no `global`, no shadowing, no
rebinding anywhere, and the test file is the only importer. The read is a free-variable lookup
resolving to the module global instead of a local — same value, no semantic change, and the constant
is no longer re-assigned on every call. Baseline 23/23. This is the right fix for the gap I raised.

## Finding 7 — your item 4, the defensive `else` at `test_performance_capture_report.py:175`

Neither delete it nor keep it as written. Its **intent** is right — tolerate a future
reclassification — but its assertion is now wrong, because that capture's readback is 60% of measured
work, so the note fires on share alone whatever the verdict becomes. Keep the tolerance where it
belongs (the classification) and make the note assertion unconditional:

```python
summary = summarize(capture([], renderer=self.READBACK_RENDERER))
# 60% of measured work: the note now fires on share alone, whatever the verdict becomes.
self.assertIn("cannot say", summary["readback_note"])
```

That preserves the defence and makes it true under the new trigger, instead of leaving an arm that
would fail on the wrong assertion if it were ever reached.

---

## Verdict

**REJECTED.** Findings 1 and 3 are blocking: the anchor arm does not observe the relationship it
documents (bar 0.40 -> 0.50 leaves 23/23 green), and the tail ships a statement that is false on the
capture this PR was written for. Findings 4, 5 and 7 are small and worth folding into the same push.
Finding 2 is a judgement call I have deliberately argued *against* my own earlier suggestion — take
the inequality unless you can defend the ratio on evidence I have not seen.

Re-request when it is pushed and I will re-run the sweep, the bar mutation, and the per-verdict
rendering. CI is also still pending at this head.

mattias800 (commented):
Re-review of `74e1f406`. My rejection was on `f2e9c745`.

## APPROVED

Both blocking findings are genuinely fixed, and I verified them by mutation rather than by reading
the diff. Two one-line corrections are required before merge — one is a **visible output regression
this commit introduces**, the other is a comment that misattributes which arm binds. Neither needs a
re-review from me once landed.

Method: both files extracted with `git show` into a scratch dir (no worktree — `git worktree add`
timed out here, unrelated to this PR), `python3 -B` plus an explicit `rm -rf <scratch>/__pycache__`
between every arm, and both files confirmed byte-identical to the branch after the last restore.
Baseline 25/25, rc=0.

---

## (a) `_VERDICTS_READBACK_CANNOT_FLIP` — complete and correct, including the two you had not tested

I answered this structurally first and then measured it, because a fixture sweep alone could not tell
you the set is *complete*.

**Completeness, derived from the source rather than by eye.** The verdict can only ever be a key of
`classification_components`, plus `inconclusive` and `cpu-outside-renderer`. Extracting those keys
from `summarize`'s source gives the universe:

```
component keys : compute, gpu-device, gpu-wait, gpu-wait-overhead, readback, renderer-resource
universe       : the above + inconclusive + cpu-outside-renderer      (8 values, exhaustive)
non-readback winners NOT in the set : none
set members that are not reachable verdicts : none
```

The union holds across both branches of the `gpu_timestamps_available` split (`gpu-device` +
`gpu-wait-overhead` on one side, `gpu-wait` on the other), so the set is exactly
`keys(classification_components) - {"readback"}`. That is a structural argument, not a fixture
result, so it stays true for inputs I did not construct.

**Correctness, measured — with reachable fixtures for all five, `gpu-wait` and `gpu-wait-overhead`
included.** For each, I re-summarized with the readback removed (subtract `readback_ms` from the
renderer `total_ms`, zero the field) and compared verdicts:

| fixture | as captured | readback removed | flips? | note claims |
| compute | compute | compute | no | cannot flip ✓ |
| renderer-resource | renderer-resource | renderer-resource | no | cannot flip ✓ |
| **gpu-device** | gpu-device | gpu-device | no | cannot flip ✓ |
| **gpu-wait-overhead** | gpu-wait-overhead | gpu-wait-overhead | no | cannot flip ✓ |
| **gpu-wait** | gpu-wait | gpu-wait | no | cannot flip ✓ |
| inconclusive | inconclusive | **compute** | yes | may be artefact ✓ |
| cpu-outside-renderer | cpu-outside-renderer | **renderer-resource** | yes | may be artefact ✓ |

7/7 consistent. `gpu-wait-overhead` is the one I expected to need care, since it is *derived*
(`max(0, gpu_wait - gpu_device)`) rather than summed — but neither term depends on readback, so it is
invariant under the removal and the scaling argument applies unchanged. `gpu_timestamps_available`
likewise depends only on the timestamp fields.

Two more properties worth having on record:

* **The winner cannot fall through to `cpu-outside-renderer` after removal.** The `share >= BAR` test
  comes first, and removal only raises the winner's share, so the first branch still takes it.
* **The fallback is safe by construction.** You enumerated the *reassuring* side, so a component added
  to `classification_components` later defaults to "may itself be an artefact" — declining to vouch
  rather than vouching falsely. Enumerating the other side would have been unsafe by default. Right
  call.

Mutations, both directions, reproduced: adding the two flippable verdicts to the set reddens
`test_readback_note_does_not_vouch_for_a_verdict_the_readback_can_flip`; **removing `compute` from it**
reddens `test_readback_note_tail_does_not_disown_a_verdict_readback_did_not_decide`. So membership is
pinned from both sides, which is more than you claimed.

One residual assumption, non-blocking, worth a line in the comment: the claim is a counterfactual
about a *re-measurement*, and it holds exactly if removing the harness readback is a pure subtraction
from the renderer total and touches nothing else. The only evidence for that is #3152's single data
point — and it does support it: a 51%-of-measured readback going to 0.0 ms **halved the graphics
total**, which is what pure subtraction predicts. Say that is the basis, so nobody later assumes it
was derived.

## (b) The bar arm does exercise the bar — but the comment is wrong about which arm holds it

Your fixtures avoid the trap you flagged. `just_under` is 600/1600 = 37.5% and `just_over` is
800/1800 = 44.4%, and both have `measured_total` (1600 / 1800) well above `wall_ms * 0.40` = 400, so
neither lands in `cpu-outside-renderer`. Verified: they classify `inconclusive` and `compute` as
asserted, and the arm reddens at every bar value I tried (0.30, 0.375, 0.38, 0.44, 0.4444, 0.45,
0.50).

But `test_performance_capture_report.py:241` says:

> Both arms below fail if it moves, so the bar is now held by behaviour and not only by a literal.

**Measured false.** I neutralised only the `assertAlmostEqual` on `:243` and re-ran the bar arm:

| bar | 0.376 | 0.38 | 0.42 | 0.44 | 0.4444 | 0.445 | 0.50 |
| --- | --- | --- | --- | --- | --- | --- | --- |
| behavioural arms alone | pass | pass | pass | pass | pass | **fail** | **fail** |

The two behavioural arms tolerate the bar anywhere in **(0.375, 0.4444]** — an ~18% window that
includes 0.38, 0.42 and 0.44. Inside it the only thing that reddens is the literal on `:243`. So the
bar is held by the *literal* for small moves and by behaviour only for large ones, which is the
opposite emphasis to what the comment gives. This is the third instance in this series of a claim
about *which arm binds* stated without measuring per-arm — you fixed exactly this in the band comment
twelve lines up, so it is worth naming rather than quietly correcting.

The design is right, though, and it has a better justification than the comment offers. I decoupled
the constant from its use site in both directions:

* **D1** — use site reverted to a bare `0.50`, constant left at `0.40`: only the **behavioural** arms
  can see this. rc=1.
* **D2** — use site reverted to a bare `0.40`, constant moved to `0.50`: only the **literal** can see
  this. rc=1.

So they are genuinely complementary — the behavioural pair pins the *wiring*, the literal pins the
*value*, and neither is redundant. That is what the comment should say.

---

## Required before merge

**R1 — the readback tail now renders as a lowercase fragment after a full stop.** This commit moved
the period into the callers (`:422`, `:428`) so the new full-sentence tails read correctly, but left
`:286` as the old subordinate clause. Rendered:

```
... This readback is the harness, not the title. Re-measure through a real window. before acting
on this verdict.
... Check the run log for a GPU-present surface failure. before acting on this verdict.
```

At `f2e9c745` and earlier this read "Re-measure through a real window before acting on this verdict" —
correct. So it is a regression, on the **readback verdict**, which is the branch this note was
originally written for and the most common one it fires on. The five non-readback tails all read
cleanly (capital `The …`); only `:286` was missed.

No test can see it: `:272` asserts `assertIn("before acting on this verdict", ...)`, which matches the
broken string unchanged. If you keep that substring the fix is punctuation only — e.g.
`" Re-measure before acting on this verdict."` — otherwise `:272` needs updating with it.

**R2 — correct `:241`.** Replace "Both arms below fail if it moves" with the measured statement: the
behavioural pair brackets (0.375, 0.4444] and catches a change to the *use site*; the literal on `:243`
catches a change to the *value*; both are needed. The D1/D2 mutations above are the evidence.

## Clean, no action

* `CLASSIFICATION_EVIDENCE_SHARE` named once (`:233`) and used at both dependent sites (`:364` and the
  reason string at `:373`, which now derives "40%" instead of hard-coding it). The second-order gap I
  raised — nothing pinned the tool's primary threshold at all — is closed.
* The ratio assert is gone and the test asserts only `0 < MIN_SHARE < BAR`. That is the invariant the
  design actually rests on, and I am glad it is not carrying my off-hand fraction as an equality.
* Item 4 landed as suggested: the note assertion is unconditional, only the classification tolerant.
* Tail fixture at 28% against a 70% compute verdict — the 2-point-headroom hazard is gone.
* `git diff --check` clean; no `Claude-Session:` trailer on the branch.

One methodological note for whoever mutates this file next, since the structure changed:
`READBACK_NOTE_MIN_SHARE = CLASSIFICATION_EVIDENCE_SHARE / 4` means **a bar mutation is now a compound
mutation** — moving the bar also moves the note threshold. It happens not to confound the values I
swept (at bar 0.50 the threshold is 0.125 and every readback arm still sits clear of it), but at
bar ≥ 0.80 the threshold leaves the pinned band and readback arms start failing for a reason that has
nothing to do with the bar. Worth one line beside the definition. Relatedly, the comment says "the
tests deliberately do not enforce the ratio" — true, but the *code* now enforces it absolutely, which
is a stronger coupling than the assert I argued against. I think that is the right place for it (one
source of truth, and the inequality can never be violated by a bar change); it just reads as weaker
than it is.

**APPROVED** — land R1 and R2 and merge; no re-review needed for a punctuation fix and a comment
correction. CI is still pending at this head.

mattias800 (commented):
Re-review of `20ea8da9`. My approval was on `74e1f406`; R1 and R2 both land correctly.

## APPROVED

The new guard works, and it works for the right reason — I re-broke the tail and it reddens
independently of the content arm, which is the property that makes it a class guard rather than a
second instance check. Three findings below, **none blocking**; the first is one character plus one
line and I have verified the fix green, so it is worth folding in before you merge, but the call is
yours and I do not need to see it again.

Same method: both files via `git show` into scratch, `python3 -B`, explicit
`rm -rf <scratch>/__pycache__` per arm, both files confirmed byte-identical to the branch after the
last restore. Baseline 26/26, rc=0.

---

## R1 — fixed, and the guard is real

Re-broke `_readback_note_tail`'s readback branch to the old `" before acting on this verdict."`:

```
rc=1  failing: test_every_readback_note_is_well_formed_prose,
               test_readback_note_tail_does_not_disown_a_verdict_readback_did_not_decide
```

The new arm catches it. That it fires *alongside* the content arm rather than instead of it is the
point — the content arm only reddens because you also changed the phrase it asserts on, whereas the
join guard would have caught the break with the phrase left alone. That is exactly the class the
finding was about, and it is a better answer than the one I asked for.

The rendered readback note now reads cleanly on both joining callers:

```
...the harness, not the title. Re-measure through a real window. Readback IS this capture's
verdict, so there is nothing else here to act on.
...Check the run log for a GPU-present surface failure. Readback IS this capture's verdict, so
there is nothing else here to act on.
```

And the new wording is the more accurate one, as you say — on a readback verdict the readback is not
a caveat on the conclusion, it is the conclusion.

## R2 — fixed, and your table reproduces exactly

I re-swept with the literal assert neutralised, including the `0.375` endpoint I had not measured:

| bar | 0.375 | 0.376 | 0.42 | 0.4444 | 0.445 | 0.50 |
| --- | --- | --- | --- | --- | --- | --- |
| behavioural only | **FAIL** | pass | pass | pass | **FAIL** | **FAIL** |

Identical to the comment now at `test_performance_capture_report.py:265`. The complementarity
statement is right and now measured rather than asserted.

---

## Finding 1 — the guard omits the one branch that fails it (non-blocking, verified fix)

`test_every_readback_note_is_well_formed_prose` (`:176`) covers four notes. The fifth reachable
note — the `adopted is True` branch, "GPU present WAS adopted ... Chase it rather than dismissing it
as a harness artifact" (`performance_capture_report.py:442`) — is **not** among them, and it is the
only reachable note that **fails the guard's own final assertion**:

```
adopted=False, verdict=readback    ends-with-period=True
adopted=None,  verdict=readback    ends-with-period=True
adopted=TRUE,  verdict=readback    ends-with-period=FALSE
adopted=TRUE,  verdict=compute     ends-with-period=FALSE
```

The missing period is pre-existing and cosmetic — not something this PR introduced. What is new is an
arm named "every readback note" whose coverage stops exactly where the property stops holding, which
is the shape this whole review series has been about. Its other three assertions (lowercase after a
period, space before a period, doubled period) pass on that branch; only `:196` fails.

I verified the fix rather than proposing it blind: adding the terminating period at `:442` and the
`adopted` case to the `cases` dict gives **rc=0, no failures**. One character and one entry.

## Finding 2 — `:299` is now structurally vacuous (non-blocking)

`self.assertNotIn("before acting on this verdict", decisive["readback_note"])` can no longer fail.
Checked by walking the module's AST rather than grepping, so a comment does not confuse it: **0
string literals in the module contain that phrase** (the single textual occurrence is the explanatory
comment at `performance_capture_report.py:292`). No code path can emit it, so the assertion is
permanently true. Replace it with the live discriminator — `assertNotIn("nothing else here to act
on", decisive["readback_note"])` — which mirrors the `assertNotIn("does not depend on it", owned...)`
you already added on `:307` and would actually fail if the two tails were swapped.

## Finding 3 — completeness of the join matrix (minor)

There are six reachable joins: two callers that append a tail (`adopted is False`, `adopted is None`)
times three tail branches. The dict at `:181-:189` hand-lists four, all three tails under
`adopted is False` plus the readback tail under `adopted is None`; the two `adopted is None` x
non-readback joins are not exercised. They are fine today because both prefixes now terminate, but
the guard's value is that it does not depend on anyone noticing that. Building `cases` as the product
of `{False, None, True}` x `{readback, compute, inconclusive}` makes it genuinely "every", subsumes
finding 1, and is shorter than the dict it replaces. (Minor naming point while you are there: the dict
mixes verdict names with an adopted-state name, `"cannot-say"`, which reads as a fifth verdict.)

---

## Your open question — I agree, do not change it

You are right and I would leave it exactly as it is. The code deriving `READBACK_NOTE_MIN_SHARE` from
the bar is what keeps the inequality true by construction, and the tests deliberately not asserting
the ratio is what stops a chosen fraction acquiring the standing of an invariant. Those are two
different jobs and they belong in two different places; the apparent mismatch is the design, not a
slip. My earlier note was an observation that a reader might misread it, and the comment now says
enough that they will not.

Your compound-mutation note landing beside the definition is the right home for it — that is the kind
of thing that is only ever learned by someone confusing themselves with a wide sweep, and now nobody
has to.

## Housekeeping

`git diff --check` clean; no `Claude-Session:` trailer on the branch; 26/26 rc=0; module and test both
byte-identical to the branch after every mutation restored.

**APPROVED.** Merge on green CI whenever you are ready. Findings 1-3 are cleanups, not conditions —
finding 1 is the one I would spend the two lines on, since an arm called "every" that excludes the
failing case is the failure mode this branch has now corrected three times.

PR #3188, opened 2026-09-01
Extract sem_timedwait's deadline/slice arithmetic to clock-free helpers (#3069)

### LINKED ISSUE #3069: Extract the sem_timedwait poll loop's deadline/slice arithmetic to a clock-free helper, so the overshoot band and #3044's saturating arithmetic get tested
Follow-up from independent review of #3066. Raised there as non-blocking, and deliberately not done in that PR because #3066 is test-only and was unblocking a red master.

## What is untested

`k_sem_timedwait`'s Windows poll loop (`prosper/src/hle/kernel/hle_kernel.cpp:3101` onward) contains three pieces of arithmetic that no test exercises:

1. **The saturating microsecond→nanosecond conversion** (`:3096`): `a1 > kNsMax / 1000ull ? kNsMax : a1 * 1000ull`. `a1` is guest-controlled, and a wrap here turns a very long timeout into a short one.
2. **The saturating deadline addition** (`:3109`): `timeout_ns > kNsMax - start_ns ? kNsMax : start_ns + timeout_ns`. A wrapped deadline lands in the past, turning a long wait into an instant timeout.
3. **The adaptive slice and its clamp**: 500 µs while the wait is young, 2 ms after 8 ms have elapsed, clamped against the remaining time at `:3133`.

Both saturating paths arrived with #3044 and neither has a test. They are guest-reachable by construction — `a1` comes straight from the guest.

## Why a test cannot reach them today

The arithmetic is inline in an `HLE()` body, so exercising it means calling the real HLE and observing a duration. That is precisely the assertion #3066 has just established cannot discriminate anything here: a quantized wait returns at the next tick boundary, so the defect's latency distribution overlaps the fix's, and no single-run wall-clock bound separates them.

Concretely, the band #3066 gives up: an overshooting loop in the 12–500 ms range. It needs two simultaneous edits to reach — grow the slices *and* drop the `remaining` clamp at `:3133` — so it is not the likeliest regression, but it is currently invisible.

## The fix

Extract to a pure function beside the deadline sleeper it belongs with, the way `next_grid_deadline_ns` was extracted in #3064:

```cpp
// prosper/src/host/platform/precise_sleep.hpp  (or a kernel-local header)
constexpr uint64_t timeout_ns_from_us(uint64_t us);                       // saturating
constexpr uint64_t deadline_ns_from(uint64_t start_ns, uint64_t timeout_ns);   // saturating
constexpr uint64_t poll_slice_ns(uint64_t elapsed_ns, uint64_t remaining_ns);  // 500us/2ms + clamp
```

Then unit-test them with no clock at all: `UINT64_MAX` microseconds must not wrap, `start_ns` near `UINT64_MAX` must not produce a past deadline, and the slice must never exceed the remaining time nor reach zero while time remains.

There is a worked precedent for exactly this shape and exactly this reasoning — #3064 extracted the vblank pump's deadline arithmetic for the same reason (the inline spelling was wrong in a way that read as correct), and the arms are clock-free.

## Not urgent, but not cosmetic

The saturating arms are the guest-facing ones and they are the reason to do this. `CONFIDENCE: HIGH` that they are untested; no evidence that either is currently wrong.

Refs #3066. Refs #3044. Refs #3013.

mattias800:
Fixes #3069. Refs #3067. Refs #3066. Refs #3044. Refs #3064. Refs #3074.

## Why

`scePthreadSemTimedwait`'s Windows poll loop carried three pieces of arithmetic that **no test reached**, two of them saturating arms guarding a **guest-controlled** timeout:

1. saturating microsecond -> nanosecond conversion — a wrap turns a 584-year timeout into a few hundred nanoseconds;
2. saturating deadline addition — a wrapped deadline lands in the **past**, turning a long wait into an instant timeout;
3. the adaptive poll slice and its clamp against the remaining time.

They were unreachable by construction. The arithmetic sat inline in an `HLE()` body, so the only handle on it was to call the real HLE and observe a duration — and `test_kernel_sem_timedwait.cpp`'s own header records at length why that cannot discriminate anything here: a quantized wait returns at the next tick **boundary**, so the defect's latency distribution overlaps the fix's for any tick length, and no single-run wall-clock bound separates them.

This is the same shape, and the same argument, as #3064 (`next_grid_deadline_ns`) and #3074 (`sleep_until_deadline_retry`).

## The helper's contract

Added to `prosper/src/host/platform/precise_sleep.hpp`, beside the two existing pure functions and beside the deadline sleeper the loop actually calls:

| function | contract |
| --- | --- |
| `timeout_ns_from_us(us)` | saturating us -> ns; never returns fewer nanoseconds than the microseconds asked for |
| `deadline_ns_from(start_ns, timeout_ns)` | saturating add; the deadline never lands **behind** its own start |
| `poll_slice_ns(elapsed_ns, remaining_ns)` | 500 us inside the 8 ms fine window, 2 ms after; never exceeds `remaining_ns` |
| `sem_poll_step(start_ns, deadline_ns, now_ns)` | the whole per-iteration decision: `{expired, sleep_until_ns}` |

All pure, all `constexpr`, all taking the clock reading as a **parameter** — so a test drives the entire loop with a fake clock: no `_WIN32`, no real sleep, no wall time.

`sem_poll_step` guarantees, for every live step:

    now_ns < sleep_until_ns <= deadline_ns

The **strict** lower bound forbids a zero-length sleep (a busy spin while time remains). The upper bound is both the clamp and the reason `now_ns + slice` cannot wrap near `UINT64_MAX`.

`poll_slice_ns` is exposed separately from its only caller so the clamp is attributable to one function under mutation.

## What is provably unchanged

**Pure extraction — no behaviour change**, and that is measured rather than asserted. A differential harness transcribed the original inline expressions verbatim from `origin/master` and compared them against the helpers over **12,011,237** input triples — both saturation boundaries, the fine/coarse crossover, backwards clocks, and the last nanoseconds before `UINT64_MAX`:

    checked=12011237 mismatches=0
    DIFFERENTIAL PASSED (extraction is behaviour-identical)

**The zero is not vacuous.** Positive control: perturbing the slice by **1 ns** yields `mismatches=5480686` and `DIFFERENTIAL FAILED`. The harness is scratch, not committed.

The only prose that moved is the sleep-floor measurement and the three cadence constants, which now live beside `poll_slice_ns`; the call-site comment that referred to them "below" now points at the header instead.

## Affected / unaffected

- **Affected:** the Windows (`_WIN32`) poll loop's spelling only. Same values, same order, same branches.
- **Unaffected:** the POSIX branch keeps the native `sem_timedwait` and its exact semantics. The `WaitCensusScope` census value is byte-identical (covered by the differential sweep).
- **Not touched, deliberately:** `abs_deadline_us(a1)` on the POSIX path still takes the raw `a1` rather than the saturated `timeout_ns`. Checked — for `UINT64_MAX` it yields `tv_sec` ~1.8e13 (year ~586,000), i.e. an effectively infinite wait, which agrees with the saturating intent. Pre-existing on master and out of scope for a pure extraction.

## The backwards clock is PINNED, NOT FIXED

`now_ns - start_ns` on a clock that ran backwards wraps to a huge value and so selects the **coarse** slice. That is master's behaviour, preserved deliberately rather than corrected, because this PR is an extraction.

It is unreachable through the real call site — both reads are `steady_clock`, which cannot go backwards — and benign if it ever were reached: the loop polls at 2 ms instead of 500 us and **still times out at exactly the right moment**, because the expiry test does not involve `start_ns` at all. A test arm asserts the status quo and says in as many words that it pins it rather than endorses it, so a later change there is a deliberate one.

**No arithmetic bug was found** while extracting. Both saturating arms are correct at their boundaries, which the tests now pin on both sides.

## Fake-clock cases covered

Arms 5-11 of `test_precise_sleep.cpp`, none of which a wall-clock test can produce on demand:

- **early wakeup** — a sleep returning roughly halfway, driven to termination;
- **deadline already past**, and the exact instant it becomes one (`>=` vs `>`);
- **clock running backwards**;
- **slice clamping**, in both regimes and at both `<` boundaries;
- **overflow near the representable maximum** — start at `UINT64_MAX - 10`, walked one nanosecond at a time.

The loop-driving arms **count which regime each step used** and assert both were entered, so an arm cannot pass while never reaching the branch it claims to cover. Observed: `exact wakes: 37 steps, 16 fine, 21 coarse, 1 clamped` / `early wakes: 92 steps, 32 fine, 60 coarse, 20 clamped`. A separate arm asserts `early.steps > exact.steps`, so the early arm is provably not a duplicate of the exact one.

## Commands and results

Configure + build (in the distrobox; a host build silently drops the Vulkan execution tests):

    cmake -S prosper -B prosper/build-linux                      # rc=0
    TMPDIR=$PWD/prosper/build-linux/tmpdir cmake --build prosper/build-linux -j8    # rc=0
    ctest --test-dir prosper/build-linux --no-tests=error        # rc=0, 100% tests passed out of 318

`kernel_sem_timedwait` (test 53) passes in 0.01 s; `precise_sleep_deadline_retry` (test 258) passes.

**The Windows branch is the one actually rewired, and Linux never compiles it**, so it was compile-verified separately with MinGW:

    x86_64-w64-mingw32-g++ -std=gnu++20 -Iprosper/src ... -fsyntax-only \
        prosper/src/hle/kernel/hle_kernel.cpp                    # rc=0, 0 errors

With a positive control proving that check is real — injecting an undeclared identifier **inside** the `#else` branch gives MinGW `rc=1` (`error: 'PROSPER_I3069_POSITIVE_CONTROL' was not declared in this scope`, at the rewired line) while native `g++` still returns **rc=0** on the same broken file, confirming Linux preprocesses that branch out entirely.

## Mutation evidence

Each mutation was applied to the helper, rebuilt, and run; `build_rc` is reported alongside because a failed build leaves the old binary and reports a false pass. **All builds succeeded, so every result below is a real test verdict.**

    baseline_rc=0   restored_rc=0

    M1  drop the clamp in poll_slice_ns                  mutant_rc=1  (build_rc=0)
    M2  flip the expiry test >= to > in sem_poll_step    mutant_rc=1  (build_rc=0)
    M3  drop the saturation in timeout_ns_from_us        mutant_rc=1  (build_rc=0)
    M4  drop the saturation in deadline_ns_from          mutant_rc=1  (build_rc=0)
    M5  fine-window boundary < becomes <=                mutant_rc=1  (build_rc=0)
    M6  swap the fine slice to 2 ms (cadence inversion)  mutant_rc=1  (build_rc=0)

**M6 initially SURVIVED** and is reported because it was a real gap, not because it was found by design: every other assertion read the cadence constants *symbolically*, so changing a constant's value could not redden any of them. An arm pinning the three literals closes it, commented as a tripwire on a measured tuning decision rather than a claim the values are optimal.

## Risks

- Low. The extracted expressions are identical (12M-case differential), the call site is a mechanical substitution, and the only platform whose code path changed shape was compile-verified under MinGW with a positive control.
- The **runtime** Windows path is not exercised here — no Windows host in this session. Mitigated by the differential (the values are identical) and the MinGW syntax check (the code compiles), but stated plainly rather than glossed.
- `sem_poll_step` returns `sleep_until_ns = 0` when expired; the field is documented as unset in that case and the call site never reads it.

## Note for #3067

#3067 remains open, undiagnosed, and untouched by this PR -- no claim is made on it here. What this does remove is the reason its arithmetic was untestable, which was the stated purpose of #3069.

One observation, offered as a **lead rather than a finding**: `kernel_sem_timedwait` takes **0.01 s** on Linux here, while #3067 records the macOS/Rosetta failures at **1.02 s and 1.22 s**. ARM 4 of that test grants a **one second** timeout and asserts the posted acquire returns nowhere near it, so a runtime of ~1.0 s is what waiting that timeout out would look like. That is consistent with the failure being ARM 4 rather than the timing bounds in ARM 3 — untested, and it does not distinguish #3067's two candidates (a loaded runner vs. something environmental). Whoever picks it up should confirm against the job's own per-arm output before believing it.

### DIFF
--- a/prosper/src/hle/kernel/hle_kernel.cpp
+++ b/prosper/src/hle/kernel/hle_kernel.cpp
@@ -3137,16 +3137,20 @@ HLE(k_sem_trywait)   { auto* s = ensure_sem(a0); if (!s) return 0x16; return sem
 // fix the cadence and blunt the responsiveness in the same change. Short slices for the first few
 // milliseconds keep the wake tight where pacing lives; longer slices afterwards keep a guest parked
 // for seconds from spinning at 2 kHz for no benefit. (An earlier revision of this comment said
-// 5 kHz, which was the 200 us slice that revision used before the floor measurement below
-// replaced it.)
+// 5 kHz, which was the 200 us slice that revision used before the floor measurement replaced it --
+// that measurement, and the three constants it fixes, now live beside poll_slice_ns() in
+// precise_sleep.hpp.)
 HLE(k_sem_timedwait) {
     auto* s = ensure_sem(a0);
     if (!s) return prosper::hle::kSceKernelErrorEINVAL;
     // Saturating, matching guest_ns_from in hle_kernel_time.cpp (#3022): a1 is guest-controlled,
     // and a wrap here would turn a very long timeout into a short one. That file states the rule
     // for the whole family, so this member of it should not be the exception.
-    const uint64_t kNsMax = ~0ull;
-    const uint64_t timeout_ns = a1 > kNsMax / 1000ull ? kNsMax : (uint64_t)a1 * 1000ull;
+    // The arithmetic itself lives in precise_sleep.hpp, beside the deadline sleeper the Windows
+    // loop below calls, so it can be driven by a FAKE CLOCK instead of only by timing the real HLE
+    // -- which test_kernel_sem_timedwait.cpp's header explains at length cannot discriminate here
+    // (#3069). Behaviour is identical; this is an extraction, not a change.
+    const uint64_t timeout_ns = prosper::host::timeout_ns_from_us(a1);
     hle::WaitCensusScope c(hle::WaitKind::SemTimedwait, timeout_ns);
 #ifndef _WIN32
     timespec dl = abs_deadline_us(a1);
@@ -3160,19 +3164,7 @@ HLE(k_sem_timedwait) {
     // Saturating again on the addition: start_ns + timeout_ns can itself wrap for a huge timeout,
     // and a wrapped deadline lands in the past, which would turn a long wait into an instant
     // timeout -- the loud direction, but still wrong.
-    const uint64_t deadline_ns =
-        timeout_ns > kNsMax - start_ns ? kNsMax : start_ns + timeout_ns;
-    // 500 us while the wait is young, 2 ms after 8 ms have passed. The crossover is above one grain
-    // period so an audio pacing wait never leaves the tight regime.
-    //
-    // 500 us and not less, because that is the FLOOR of a short high-resolution-timer sleep on
-    // this host, measured rather than assumed: requests of 0.05, 0.10 and 0.20 ms all cost a
-    // median of 0.52 ms (200 samples each; 0.50 ms -> 1.02, 1.00 -> 1.51, 2.00 -> 2.07). An
-    // earlier revision asked for 200 us, which reads as a tighter loop than the machine can
-    // actually run and would mislead the next person tuning it. What the fine window buys is a
-    // poll interval about 4x shorter than the coarse one (0.52 ms against 2.07), not the 10x the
-    // constants implied.
-    const uint64_t kFineSliceNs = 500000ull, kCoarseSliceNs = 2000000ull, kFineWindowNs = 8000000ull;
+    const uint64_t deadline_ns = prosper::host::deadline_ns_from(start_ns, timeout_ns);
     for (;;) {
         if (sem_trywait(s) == 0) return 0;
         // EAGAIN means "not posted yet", which is the whole point of the loop. Anything else --
@@ -3182,10 +3174,13 @@ HLE(k_sem_timedwait) {
         if (errno != EAGAIN && errno != EINTR) return sce_pthread_rc(fbsd_errno(errno));
         const uint64_t now_ns = (uint64_t)std::chrono::duration_cast<std::chrono::nanoseconds>(
                                     std::chrono::steady_clock::now().time_since_epoch()).count();
-        if (now_ns >= deadline_ns) { errno = ETIMEDOUT; return sce_pthread_rc(fbsd_errno(errno)); }
-        const uint64_t slice = (now_ns - start_ns) < kFineWindowNs ? kFineSliceNs : kCoarseSliceNs;
-        const uint64_t remaining = deadline_ns - now_ns;
-        prosper::host::sleep_until_steady_ns(now_ns + (remaining < slice ? remaining : slice));
+        // The deadline test, the adaptive slice and its clamp against the remaining time, as one
+        // pure decision -- see prosper::host::sem_poll_step for the cadence rationale and for the
+        // two invariants it guarantees (now_ns < sleep_until_ns <= deadline_ns).
+        const prosper::host::SemPollStep step =
+            prosper::host::sem_poll_step(start_ns, deadline_ns, now_ns);
+        if (step.expired) { errno = ETIMEDOUT; return sce_pthread_rc(fbsd_errno(errno)); }
+        prosper::host::sleep_until_steady_ns(step.sleep_until_ns);
     }
 #endif
 }
--- a/prosper/src/host/platform/precise_sleep.hpp
+++ b/prosper/src/host/platform/precise_sleep.hpp
@@ -97,4 +97,93 @@ constexpr uint64_t grid_boundaries_missed(uint64_t prev_deadline_ns, uint64_t ne
     return advanced > 0 ? advanced - 1 : 0;
 }
 
+// ---------------------------------------------------------------------------
+// scePthreadSemTimedwait's poll-loop arithmetic (#3069).
+//
+// Extracted UNCHANGED from the Windows poll loop in hle_kernel.cpp's k_sem_timedwait, for the same
+// reason next_grid_deadline_ns and sleep_until_deadline_retry above were extracted: it was inline in
+// an HLE() body, so the only way to reach it was to call the real HLE and time it -- and
+// test_kernel_sem_timedwait.cpp's own header records at length why a wall-clock duration cannot
+// discriminate anything here (a quantized wait returns at the next tick BOUNDARY, so the defect's
+// latency distribution overlaps the fix's, for any tick length, and no single-run bound separates
+// them). Two of the three pieces below are saturating arms guarding guest-controlled input, and no
+// test reached either.
+//
+// Everything here is pure and takes the clock reading as a PARAMETER, so a test drives the whole
+// loop with a fake clock: no _WIN32, no real sleep, no wall time. See test_precise_sleep.cpp.
+
+// Saturating microseconds -> nanoseconds. `timeout_us` arrives straight from the guest, and a wrap
+// here would turn a very long timeout into a short one. Matches guest_ns_from in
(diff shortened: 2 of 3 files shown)


mattias800 (commented):
**APPROVED**

**Mutations verified independently**, not taken on report. Both with a clean build (0 errors, so no
stale-binary false pass):

| mutation | result |
| --- | --- |
| M1 drop the clamp in `poll_slice_ns` | `rc=8` |
| M2 flip expiry `>=` → `>` | `rc=8` |
| baseline / restored | `rc=0` |

**ctest 318/318, exit 0** on a *clean* configure. Worth stating why I insist on that: my incremental
build directory reported 5 Vulkan-execution failures on this branch — and the same 5 on `origin/master`.
A fresh configure passes all of them. The failures were my build tree, not your change and not master.
Recording it because "it fails on master too" is exactly the reasoning that would have let a real
regression through, and here it would have been wrong in the other direction.

**The differential harness is the part that makes this reviewable.** A pure extraction is easy to
claim and hard to demonstrate, and 12,011,237 triples with 0 mismatches against the transcribed
original would mean nothing on its own — a harness that compares a function to itself also reports 0.
The 1 ns perturbation yielding 5,480,686 mismatches is what turns it into evidence, and it is the
charter's positive-control rule applied without being asked. That control is drawn from the same
generator as the null it validates, which is normally the trap; here it is legitimate, because the
property under test is *equality of two implementations* rather than the presence of a class of
input, so perturbing one side genuinely exercises the discriminator.

**M6 surviving initially, and you reporting it, is the most valuable line in the report.** Every
other assertion read the cadence constants symbolically, so a constant's *value* could not redden
anything — a whole class of change was invisible. Most reports would have listed six green mutations
and moved on. Adding the literal-pinning tripwire and saying it was a real gap rather than a designed
catch is the behaviour I want repeated.

**Anti-vacuity counters** (`exact: 37 steps, 16 fine, 21 coarse, 1 clamped`) asserting both regimes
were entered, plus `early.steps > exact.steps` so the early arm is not a copy of the exact one —
this is the check that separates "the assertion is true" from "the mechanism ran", which this
project has been burned by repeatedly.

**The MinGW proof-of-proof.** The `#else` branch is the one you rewired and Linux never compiles it,
so "MinGW returned 0" would have been an untested claim about an untested path. Injecting an
undeclared identifier and showing MinGW fails at the rewired line *while native g++ still returns 0
on the same broken file* establishes the check can actually see that code. That is the right amount
of paranoia for a platform you cannot run.

**Backwards clock: pinned, not fixed, and labelled as such.** Correct call. `now_ns - start_ns` wraps
and selects the coarse slice; the deadline check does not involve `start_ns`, so expiry is still
exact. Preserving it is what makes this a pure extraction, and the comment says "preserved
deliberately rather than corrected" instead of implying it is right. No issue needed — unreachable
through `steady_clock`, benign if reached, and the comment plus the arm is the durable record.

**Near-miss caught:** "does **not fix** #3067" parsed as a closing keyword. Verified on my side —
`closingIssuesReferences` returns **#3069 only**. Good catch; that would have silently closed an open
macOS bug.

**Honest negatives accepted:** Windows runtime untested (mitigated by the differential and MinGW
compilation, not eliminated — correctly stated); the POSIX `abs_deadline_us` still takes raw `a1`,
checked and out of scope. Also: your ctest count of 318 was right and my briefing's 314 was stale.

CI 9 pass / 2 skipping. No `Claude-Session:` trailer, no absolute host paths.

Issue #304, opened 2026-07-10
UE4 (PPSA17942): UMG/Slate quad renders degenerate (flat wedge, no artwork) — per-vertex attributes collapse in the Slate vertex fetch


mattias800:
Follow-up to PR #303 (refs #273). With Slate vertex attributes now resolving correctly (real positions + UVs), the loading-banner widget SAMPLES its real material textures — but the banner region renders as fine vertical-striped noise (1-px alternating columns with speckle bands) instead of the banner artwork/text, plus one half of the widget quad shades a flat green/teal.

Evidence (PROSPER_DUMP_RTGROUPS, new in #303):
- The banner widget's PS binds a 2000x1636 BC7 SRGB texture (tile_mode 9, IMG_FMT 170) = the banner artwork, a 32x32, and cbuf color (1.0, 0.455, 0.0) — this is UMG's loading-banner element.
- The compose pass samples the 2214x444 banner RT; the striped noise appears both in the banner RT group and the presented scanout.
- Stripe pattern (alternating columns) is characteristic of either (a) a detile/stride mismatch decoding the 2000x1636 BC7 SW_64KB_S artwork (the 64KB detiler was validated on 1024x512 BC1; 2000-wide BC7 may hit an untested path), or (b) the RTT injection substituting a full-res framebuffer into a smaller texture slot, or (c) the compose sampling the banner RT address from raw guest memory (never written by our renderer) when the injection cache misses.
- The flat half-quad (wedge along the quad diagonal, visible in frames from /root/df9 + /root/df10 runs) suggests per-vertex UVs still degenerate for 3 of 4 vertices on SOME widget instances.

Repro: PPSA17942, PROSPER_RENDER=1 PROSPER_GUEST_FS=1 PROSPER_NULL_PAGE=1 PROSPER_RTT_PERTARGET=1 PROSPER_GFXLOG=1 PROSPER_FRAME_DIR=<d> PROSPER_DUMP_CONTENT=100000 PROSPER_DUMP_RTGROUPS=400000, ~420 s. Look at rtgrp_<scanout>_*.bmp banner region.

This is now the single highest-value item for a RECOGNIZABLE DOLL loading screen: the geometry and descriptors are right; the banner artwork/glyph pixels are not.

mattias800:
CLAIMING: session fec750d6 (opus + fable subagent) | branch fix/issue-304-banner-pixels | 2026-07-10T10:48:18Z

mattias800:
[...]

PR #3189, opened 2026-09-01
fix(gpu): distinguish an announced index size from a never-announced one (#3009)

### LINKED ISSUE #3009: index_type == 0 conflates "announced 16-bit" with "never announced", so both #304 detectors run on every title
## Summary

`GpuState::index_type == 0` means **two different things** — "the guest announced 16-bit" and "the
guest never announced anything" — and prosper cannot tell them apart. Both #304 index-size detectors
therefore run against every title, including titles that explicitly said 16-bit.

## Why it matters

The detectors are heuristics over guest bytes. `index_buffer_is_unannounced_32bit_high` (added in
#3006) carries a measured residual: a 16-bit index buffer with a **period-2** pattern — a fan or cone
encoded as a triangle strip `[rim, apex, rim, apex, …]`, or a line list radiating from one hub — is
**byte-identical** to a clustered 32-bit list whenever the guest supplies no `DrawIndexOffset`, since
the caller then passes the same pointer twice. A vertex-range bound rejects the constructed cases, but
the threshold is a bound of only **65,602 records** (apex vertex 1, rim 2..65), because the implied
index is `constant << 16 | varying`. The executor's own clamp constant is 65,536, so vertex buffers of
that size are routine.

A title that **announced 16-bit** should never be subject to that heuristic at all. Today it is.

## Suggested fix

Track whether a `SetIndexType` (`IT_INDEX_TYPE`) packet was ever decoded — a bool on `GpuState`
beside `index_type`, set where `c.index_size` is consumed. Then gate the detectors on it: if the guest
announced an index size, trust it and decline outright.

Both known cases stay working, because neither title announces: DOLL's UE4 Slate/UMG quads (#304) and
Tomb Raider I-III Remastered, measured at `index_type=0` across **508,688** indexed draws of a boot
to Croft Manor with no `sceSystemService`-style announcement anywhere.

## Not yet established

- How many titles in the corpus actually call `sceAgcDcbSetIndexSize` / program `VGT_INDEX_TYPE`. If
  the answer is "most", this removes nearly all of the exposure; if "almost none", it removes little
  and the vertex-range bound stays the only guard. **Measure that before assuming the gate is worth
  the surface** — it is cheap with `PROSPER_INDEXTYPE_LOG=1`, which already prints `index_type` per
  indexed draw.
- Whether gating the *original* (zero-high-half) detector too would change behaviour for any title
  that announces 16-bit and has small-index quads. That would be a latent-bug fix, but it is a
  behaviour change beyond the scope it was written for.

Raised by independent review of #3006, which named it "the real fix" for the residual there.

mattias800:
Fixes #3009.

## The failure

`GpuState::index_type` is a sentinel field: `0` means **both** "the guest announced 16-bit" and
"the guest never announced anything", and nothing else recorded the difference. The executor had
only that field to consult, so both #304 index-size detectors
(`index_buffer_is_unannounced_32bit`, `index_buffer_is_unannounced_32bit_high`) ran against every
title — including titles that had already told prosper the answer.

Both detectors are heuristics over guest bytes, and the second carries a measured residual (a
period-2 16-bit buffer — a fan or cone as a triangle strip, a line list radiating from one hub — is
byte-identical to a clustered 32-bit list when the guest supplies no `DrawIndexOffset`; the
vertex-range bound rejects the constructed cases only above ~65,602 records, and the executor's own
clamp constant is 65,536). Letting a byte pattern with that residual overrule a fact the guest
stated is the wrong precedence, whichever way the heuristic happens to land.

## The contract

Hardware fetches indices at the size `VGT_INDEX_TYPE` holds. A detector that overrides an announced
size is emulating something the hardware does not do — so an announcement, when there is one, wins
outright. The detectors exist only because prosper's default for a title that announces *nothing*
(16-bit) is a guess, and the two known cases are exactly that: titles that never announce.

Chain, all in-tree:
`sceAgcDcbSetIndexSize` → `IT_INDEX_TYPE` packet (`hle_agc.cpp:598-601`) → decode to
`K::SetIndexType` with `c.index_size = pl[0]` (`pm4_decode.cpp:50-52`) → fold
`index_type = c.index_size` (`command_processor.cpp`, `case K::SetIndexType`) → per-draw snapshot →
`index_elem_bytes(ds.index_type)` in `realize_draw_item` (`gpu_execute.hpp`).

## The fix

`GpuState::index_type_announced` — a bool beside `index_type`, set where `c.index_size` is
consumed, carried in the per-draw snapshot, and consumed at exactly one place:

```
inline bool index_size_detection_permitted(uint32_t index_type, bool index_type_announced) {
    return !index_type_announced && index_elem_bytes(index_type) == 2u;
}
```

A bool rather than `std::optional<uint32_t>` or a non-zero sentinel: `GpuState` already spells this
family of facts as plain flags (`indexed`, `from_offset`, `di_valid`), the field is copied into a
snapshot on every state-dirty draw, and a non-zero sentinel could not express "announced 32-bit,
then announced 16-bit" — which lands on `index_type == 0` with `announced == true`.

Lifetime matches `index_type`: `agc_graphics_state()` is a process-lifetime fold (only the per-submit
draw/dispatch lists are cleared), so "ever announced" is the question a draw in a later submit needs
answered, and it is the question the flag records.

### Separable first commit

Five packet cases in `GpuState::apply` carried a byte-identical copy of the snapshot block, so a new
field on `GpuState` had to be remembered in all five — and a snapshot that omits one reports the
field's *reset default* at every draw, which is indistinguishable from the guest having set it that
way. Commit 1 extracts `GpuState::refresh_state_snapshot()` (the same block verbatim, no behaviour
change); commit 2 is the fix. Mutation C below confirms the hazard is real and now caught.

### Instrument, not just the subject

`PROSPER_INDEXTYPE_LOG` now prints `announced=` and runs for **every** indexed draw rather than only
the ones whose announced size is 16-bit. Gating the log alongside the detectors would have made the
fix hide its own measurement: the corpus question the gate's value depends on ("how many titles
announce at all?") cannot be answered by an instrument that only fires on `index_type == 0`. The
64-line cap is unchanged and is spent faster.

## Corpus census — the issue's first "not yet established"

Answered statically, no GPU, no boot: does a dump import `sceAgcDcbSetIndexSize`
(NID `GIIW2J37e70`, `libSceAgc`, confirmed against the PS5 3.20 firmware symbol dump)? A title that
does not import it **cannot** announce; one that does **may** (linking is not calling, so IMPORTS is
an upper bound).

**21 of 55 local dumps import it; 34 do not.**

Positive control, because a clean zero from a scan that cannot see the dump's symbols is not a
negative: each no-import dump was re-scanned for other `libSceAgc` indexed-draw NIDs
(`DrawIndexAuto` `Yw0jKSqop+E`, `DrawIndex` `q88lQ+GP5Yk`, `SetIndexBuffer` `l4fM9K-Lyks`,
`DrawIndexOffset` `B+aG9DUnTKA`, `SetIndexCount` `8N2tmT3jmC8`). **The control fires on all 34**, so
every no-import verdict is a real negative.

That control earned its keep: a first pass using only `DrawIndexAuto`/`DrawIndex` read **void** on
PPSA16901 (Tomb Raider), which draws exclusively through the Gen5
`SetIndexBuffer`/`SetIndexCount`/`DrawIndexOffset` path. Had it not been run, that title's verdict
would have been recorded as a negative on an instrument that could not see it.

**Both known #304 cases are in the no-import set** — PPSA17942 (DOLL) and PPSA16901 (Tomb Raider) —
so neither can announce, and detection proceeds for them exactly as before. This replaces the
inference in `TOMB_RAIDER_STATUS.md` that "`index_type=0` across 508,688 draws" showed the title
never announces: that is precisely the reading `index_type == 0` cannot support, and the doc is
corrected in this PR to cite the import evidence instead.

## Affected / unaffected

- **Unaffected:** every title that never announces — the detectors run unchanged, same verdicts,
  same addresses. Both known #304 cases are here.
- **Affected:** a title that announces 16-bit and has an index buffer matching a detector's
  fingerprint now keeps its announced size. That is a latent-bug fix in the hardware-faithful
  direction (an announced size is a fact; a byte pattern is a guess), and the issue flags gating the
  original zero-high-half detector as a behaviour change beyond its original scope. Gating both was
  a deliberate call: both are *named* for recovering an element size the guest never announced, and
  the only way gating could regress a title is if the guest announced a size it does not use — which
  hardware would not tolerate either.
- **Unaffected:** announced 32-bit draws (already `esz == 4`), unknown encodings (still the loud
  non-indexed fallback), non-indexed draws, compute, and every address computation.

## Risk

The gate keys on `IT_INDEX_TYPE`, which is what `sceAgcDcbSetIndexSize` emits. If a title instead
programs an index-type UCONFIG register directly, prosper reads that register for the element size
**nowhere today**, so such a title looks unannounced and gets exactly the pre-#3009 behaviour.
Missing that path costs the fix, never correctness. Not asserted as absent — I did not verify which
register number that would be and deliberately did not guess one in the code.

Snapshots were **not** run (single-GPU policy for this session). From the census, **7 of the 17
guards in `snapshots.json` cover titles that import the announce entry point** and are therefore the
ones that could move: `alexkidd-gameplay` (PPSA02664), `gris-gameplay` (PPSA09804),
`evergate-title`/`evergate-gameplay` (PPSA01885), `greak-gameplay` (PPSA02849), `joe-mac-gameplay`
(PPSA02801), `asterix-gameplay` (PPSA08576), `summer-sports-gameplay` (PPSA03416). The other guarded
titles — Messenger PPSA24651, Dead Cells PPSA15552, Blasphemous 2 PPSA13579, Blue Prince PPSA25009,
Cobra PPSA17337, Terminator PPSA25872, Rugrats PPSA23396, Worms PPSA20052 — do not import it and
cannot be reached by this change.

## Verification

Built and tested in the `ps5ys` distrobox (a host build silently drops the Vulkan execution tests).

```
cmake -S prosper -B prosper/build-linux                       # rc 0
TMPDIR=<WORKTREE>/prosper/build-linux/tmpdir \
    cmake --build prosper/build-linux -j8                     # rc 0
ctest --test-dir prosper/build-linux --no-tests=error         # rc 0, 318/318 passed
```

One earlier full run showed `snapshot_guard_logic` (test 18) failing; it passed on re-run in
isolation and in the full re-run above, and it is a pure-Python unit test over `tools/snapshot/`
that this diff does not touch. Recorded as an observed flake, not dismissed.

### Mutation verification

Three mutations, each reverting one half of the fix, each rebuilt (build rc checked — a broken build
leaves the old binary and reports a false pass) and re-run.

| arm | build rc | index_size_detect | command_processor | gpu_execute |
| baseline | 0 | 0 | 0 | 0 |
| **A** — `index_size_detection_permitted` ignores `announced` (pre-#3009 behaviour) | 0 | **1** (3 arms) | 0 | **1** (2 arms) |
| **B** — the fold no longer sets `index_type_announced` | 0 | 0 | **1** (5 arms) | 0 |
| **C** — `refresh_state_snapshot()` omits the new field | 0 | 0 | **1** (1 arm) | 0 |
| restored | 0 | 0 | 0 | 0 |

Each mutation fails a *different* set of arms, and the arms that stayed green under A are the
controls — "with no announcement the detector still recovers the 32-bit quad" passes under old and
new code alike, which is what makes the announced arm mean something rather than merely proving the
detector never runs.

### What the new arms assert

- `test_command_processor` — folds **real streams** through the real `sceAgcDcb` builders, so what is
  asserted is what the command processor derives from guest bytes, not a hand-set field: no
  `SetIndexSize` → `index_type == 0`, `announced == false`; `SetIndexSize(0)` → `index_type == 0`,
  `announced == true`; the two states **agree on `index_type` and disagree on `announced`**; the
  per-draw snapshot carries both; and 32-bit-then-16-bit ends at `index_type == 0` with `announced`
  still true.
- `test_index_size_detect` — the gate predicate across the four `(index_type, announced)` states,
  plus the arm that carries the fix: the DOLL banner quad's bytes still fire the fingerprint
  (control), and the *same bytes* are left alone once the guest announced.
- `test_gpu_execute` — end to end through `realize_draw_item`: one buffer, `index_type == 0` both
  times, two announcement states, and the resolved index vectors differ (`{0,1,2,2,1,3}` against
  `{0,0,1,0,2,0}`). Before this change no fixture could tell the two apart, because they were the
  same state.

```
git log origin/master..HEAD --format=%B | grep -c '^Claude-Session:'   # 0
git diff --check                                                       # clean
```

### DIFF
--- a/prosper/docs/TOMB_RAIDER_STATUS.md
+++ b/prosper/docs/TOMB_RAIDER_STATUS.md
@@ -384,7 +384,20 @@ Measured layout, for anyone extending this: `layer_stride = 352256` for the 512x
   `index_buffer_is_unannounced_32bit_high`. **No further test over those two pointers can separate
   them** — do not try to tighten the pattern. The deciding evidence has to come from outside the
   buffer, and today that is the bound vertex buffer's record count. Residual, measured: a bound of
-  65,602 records still admits the case; #3009 is the real fix. Found by independent review of PR #3006.
+  65,602 records still admits the case. Found by independent review of PR #3006.
+  **#3009 has since bounded that residual from the other side**: `index_type == 0` no longer means
+  both "announced 16-bit" and "never announced", so a title that announced an index size never
+  reaches either detector at all. **This title's own exposure is unchanged, and that is now
+  established statically rather than inferred**: the 508,688 `index_type=0` draws never proved
+  "announces never" — that is exactly the reading `index_type == 0` cannot support — but the dump's
+  modules do not import `sceAgcDcbSetIndexSize` (NID `GIIW2J37e70`, `libSceAgc`) at all, so the
+  title **cannot** announce. Positive control on the same scan: `sceAgcDcbSetIndexBuffer`
+  (`l4fM9K-Lyks`) and `sceAgcDcbDrawIndexOffset` (`B+aG9DUnTKA`) ARE present in `eboot.bin`, so the
+  scan can see this dump's AGC NIDs and the absence is a real negative rather than a blind
+  instrument. (This title imports neither `DrawIndexAuto` nor `DrawIndex`: it draws through the Gen5
+  `SetIndexBuffer`/`SetIndexCount`/`DrawIndexOffset` path, which is why a control built from the
+  first two read as void on it.) The vertex-range bound therefore remains its only guard and the
+  period-2 residual above still stands for it.
 
 - **The renderer is not rejecting anything, and never was.** A full boot-to-gameplay run produces
   **zero** `[recompile-reject]` lines and **zero** `[compute] skip` lines. Neither the shattered
--- a/prosper/src/gpu/execute/gpu_execute.hpp
+++ b/prosper/src/gpu/execute/gpu_execute.hpp
@@ -1290,6 +1290,40 @@ inline uint32_t index_elem_bytes(uint32_t index_type) {
     return index_type == 0 ? 2u : index_type == 1 ? 4u : 0u;
 }
 
+// May the two #304 detectors below speak for this draw at all (#3009)?
+//
+// Both are named for what they do: recover an element size the guest NEVER ANNOUNCED. They are
+// heuristics over guest bytes, and the second one carries a measured residual (see its own header) --
+// so running them against a title that DID announce lets a byte pattern overrule a fact prosper was
+// told by the guest, which is the one input that cannot be wrong. Hardware fetches indices at the
+// size VGT_INDEX_TYPE holds; a detector that overrides an announced size is emulating something the
+// hardware does not do.
+//
+// The reason they used to run everywhere is that `index_type == 0` meant two things at once -- "the
+// guest announced 16-bit" and "the guest never announced anything" -- so the executor could not tell
+// an announcement from a reset default. GpuState::index_type_announced now records that separately,
+// and this predicate is the only place the distinction is consumed.
+//
+// Neither known case regresses, and the evidence for that is deliberately NOT `index_type == 0` on a
+// live run -- that reading is the very thing this flag exists because it cannot support. Both dumps
+// were checked statically instead: neither PPSA17942 (DOLL) nor PPSA16901 (Tomb Raider I-III
+// Remastered) imports `sceAgcDcbSetIndexSize` (NID GIIW2J37e70, libSceAgc) in any module, so neither
+// CAN announce, and `announced` is false for them however they are routed. Positive control on the
+// same scan: other libSceAgc indexed-draw NIDs are present in both dumps, so the scan sees their
+// symbols and the absence is a real negative rather than a blind instrument. Corpus-wide, 21 of 55
+// local dumps import the entry point and 34 do not, with the control firing on all 34 (#3009).
+//
+// LIMIT, stated because its failure direction matters: this records announcements made through
+// IT_INDEX_TYPE, which is what sceAgcDcbSetIndexSize emits (hle_agc.cpp, agc_dcb_set_index_size). If
+// a title instead programs an index-type UCONFIG register directly, prosper does not read that
+// register for the element size ANYWHERE today, so such a title looks unannounced here -- and gets
+// exactly the pre-#3009 behaviour, detectors and all. Missing that path costs the fix, never
+// correctness. PROSPER_INDEXTYPE_LOG prints `announced` per indexed draw so the corpus question
+// ("how many titles announce at all?") is one run away.
+inline bool index_size_detection_permitted(uint32_t index_type, bool index_type_announced) {
+    return !index_type_announced && index_elem_bytes(index_type) == 2u;
+}
+
 // Detect an UNANNOUNCED 32-bit index buffer (#304). DOLL's UE4 Slate/UMG quad index buffers are
 // 32-bit, but the title never calls sceAgcDcbSetIndexSize and programs no VGT_INDEX_TYPE register,
 // so index_type defaults to 16-bit — which misreads each 32-bit index as TWO 16-bit ones (the low
@@ -1381,9 +1415,10 @@ inline bool index_buffer_is_unannounced_32bit(const uint16_t* p16, const uint32_
 // which is the point. The executor's
 // own clamp constant three lines above `vb_records_unclamped` is 65,536, i.e. vertex buffers of about
 // that size are routine, so a title with a modest pool AND fan/cone geometry whose apex index is small
-// can still satisfy every clause here. The real fix is knowing whether the guest ever announced an
-// index size at all -- today `index_type == 0` means "16-bit" and "never told" indistinguishably, and
-// that distinction would let both detectors decline outright for any title that announces (#3009).
+// can still satisfy every clause here. That residual is now bounded on the OTHER side by
+// index_size_detection_permitted() (#3009): a title that announced an index size never reaches this
+// function at all, so the exposure is titles that announce nothing -- which is still the exposure for
+// the two known cases, and the vertex-range bound remains their only guard.
 inline bool index_buffer_is_unannounced_32bit_high(const uint16_t* p16, const uint32_t* p32,
                                                    uint32_t n, uint32_t vertex_upper_bound) {
     constexpr uint32_t kMinSamples = 8;     // >= 4 words per parity before "all equal" means anything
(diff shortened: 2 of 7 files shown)


mattias800 (commented):
**APPROVED**

**Mutation verified independently.** Making the gate ignore the announcement
(`return !index_type_announced && ... == 2u` → `return ... == 2u`, i.e. the old conflated behaviour)
fails **2 of 3** targeted test binaries. Mutant build `rc=0` with **0 errors**, so it is the mutant
binary and not a stale one. Restored: **318/318, exit 0**, worktree clean.

**The defect class is the interesting part, and the framing is right.** `index_type == 0` meant both
"the guest announced 16-bit" and "the guest never announced anything", so nothing could distinguish
them — and the consequence was backwards from what a reader would guess: a byte heuristic with a
measured residual could **overrule a fact the guest had stated**. Hardware fetches at the size
`VGT_INDEX_TYPE` holds, so an announcement wins outright; the detectors exist only because prosper's
default for a title that announces *nothing* is a guess. Getting that ordering right is the whole
fix.

**The static census is the part I want to single out, and specifically its positive control.** 55
dumps scanned for `sceAgcDcbSetIndexSize` (NID `GIIW2J37e70`, confirmed against the 3.20 firmware
dump): **21 import it, 34 do not**. That answers a question the issue listed as "not yet
established", with no GPU.

More valuable: **the control caught a blind instrument before it produced a false negative.** A first
pass built from `DrawIndexAuto`/`DrawIndex` read **void on PPSA16901** — Tomb Raider imports neither
and draws through the Gen5 `DrawIndexOffset` path. Without the control, that verdict would have been
filed as a negative from an instrument that could not see, **on the one title where it matters most**.
That is exactly the trap the charter's instrument-trap list exists for, caught by the author rather
than by a reviewer.

**Falsifying a derivation in `TOMB_RAIDER_STATUS.md` is the right kind of correction, and it cuts
against the author's own convenience.** That doc read "announces an index size exactly never" off
`index_type=0` across 508,688 draws — which is *precisely* the inference this issue says is
unsupportable. The conclusion happens to be right, but the derivation never established it; the
import evidence does. Correcting a doc whose conclusion agrees with you is the harder half, and the
same laundering was removed from the code comment before committing.

**Deduping the five hand-written snapshot copies into `refresh_state_snapshot()` as a separate first
commit** is the charter's "keep the move separable from the behaviour change" applied without being
asked. The braced-aggregate trap did not apply here (`GpuState` is only ever `GpuState{}` or
`make_shared`), but five duplicated field-by-field blocks are the same hazard by another route.

**Gating both detectors rather than one** is beyond the issue's stated scope and I accept the
rationale: both are *named* for recovering a never-announced size, and gating can only regress a
title that announced a size it does not use — which hardware would not tolerate either.

**Honest caveats accepted as written:** snapshots not run under the single-GPU policy, with the
exposure enumerated rather than waved at (7 of 17 guards cover importing titles, named); the UCONFIG
register announcement path is not handled, and **the author declined to guess the register number**
rather than put an unverified citation in the code — that costs the fix on such a title and never
correctness, which is the right direction to fail.

CI 9 pass / 2 skipping. No `Claude-Session:` trailer, no absolute host paths.

Issue #3137, opened 2026-08-30
recompiler: vertex attribute fetch (buffer_load_format + idxen) rejects as unresolved-operand, killing the whole stage


mattias800:
Two small **vertex** shaders fail to recompile at a `buffer_load_format_*` with `idxen` — vertex
attribute fetch — taking their draws with them.

## Evidence

*Stray* (`PPSA02101`) title screen, `PROSPER_NULL_PAGE=1`, `PROSPER_DBG=1`:

```
[recompile-reject] sh=/77  mode=unresolved-operand pc=57 words=e00c2000,6a010000
                   fmt=12 op=0x3 dst=0(kind2) src=0(k2),4(k1),106(k6) stage=vertex role=terminal

[recompile-reject] sh=/221 mode=unresolved-operand pc=66 words=e0042000,6b020900
                   fmt=12 op=0x1 dst=9(kind2) src=0(k2),8(k1),107(k6) stage=vertex role=terminal
```

Decoded: `buffer_load_format_xyzw v0, v0, s[4:7], vcc_lo idxen` and
`buffer_load_format_xy v9, v0, s[8:11], vcc_hi idxen` — SRSRC at `s4`/`s8`, **`VCC_LO`/`VCC_HI` as the
soffset**.

## This is an operand, not a missing opcode

MUBUF `0x0`–`0x3` (`buffer_load_format_x` … `_xyzw`) are all implemented in
`rdna2_emit_alu.cpp`'s MUBUF opcode switch, and `mode=unresolved-operand` says so directly. Two
candidates, and they have different fixes:

1. **The soffset.** `src[2]` is `Special` 106/107. `rdna2_alu_support.hpp` has a vertex-stage branch
   for VCC-as-scalar-data (`!b.is_compute && !b.is_fragment` → `sel(vcc,1,0)` / `0`), so this *should*
   resolve — but it is reached only after the tracked-`rs.sreg` lookup, and it is worth confirming the
   MUBUF path takes it rather than a different one.
2. **A null resource table.** `allow_smem` is `(rt != nullptr)` for graphics, and the reject sits
   right after `if (!allow_smem) { buf_op.how = "reject-no-smem"; ok = false; return true; }`. **A
   vertex stage that arrives with no resource table fails every buffer op it has** — and vertex fetch
   *is* a buffer op, so the whole stage dies. The bundle inspect shows this title does produce
   vertex stages with `resources=absent/0`.

[...]

Issue #1607, opened 2026-07-31
Nikoderiko (PPSA23760): 3D world dropped — BLOCKED ON #305 (stale user-data block), not descriptor provenance


mattias800:
## Summary

First-boot breadth triage of **Nikoderiko: The Magical World — Director's Cut (`PPSA23760`, VEA Games /
Knights Peak, Unreal Engine 4 + FMOD + Bink)** on master `bb5a11a2`.

**Rung reached: 2 — the title screen, on completely unmodified master.** No code change of any kind was needed.
The title boots through UE4 init, mounts a **20.5 GiB `.pak`**, initializes FMOD, and renders — as direct,
unmodified frontend captures at native 3840×2160 — the epilepsy warning, the **Knights Peak** publisher logo, and
the **full "NIKODERIKO — The Magical World — DIRECTOR'S CUT" title screen with `PRESS ✕ TO START`** and the
`Copyright © 2025 VEA Games` line. A scripted pad route continues past the title into the **MY.GAMES End User
Licence Agreement dialog** (fully legible, with `Cancel` / `Apply` buttons and the `L SCROLL UP/DOWN` /
`✕ CHOOSE` prompt bar).

Every 2D/UI surface in the game renders correctly. **What does not render is the 3D world** — the title screen's
background is uniformly black where hardware shows the game's art — and the cause is precisely identified below.

## Exact reproduction

Build (Linux, **inside the `ps5ys` distrobox** — a host configure silently omits Vulkan):

```bash
cmake -S prosper -B prosper/build-linux -G Ninja -DGAME_DUMP=<DUMP_ROOT>/PPSA23760-app0
cmake --build prosper/build-linux -j8
```

Title screen (no input needed — reached at ~90 s and held):

```bash
PROSPER_NO_FRAME_DUMPS=1 PROSPER_GUEST_FS=1 PROSPER_GUEST_ARGS=-force-gfx-direct PROSPER_RENDER=1 \
  ./build-linux/screenshot <DUMP_ROOT>/PPSA23760-app0 \
      --seconds 6 --count 30 --out <OUT_DIR> --timeout 280
```

Route past the title into the EULA:

```bash
PROSPER_NO_FRAME_DUMPS=1 PROSPER_GUEST_FS=1 PROSPER_GUEST_ARGS=-force-gfx-direct PROSPER_RENDER=1 \
[...]

Issue #305, opened 2026-07-10
UE4/AGC: a draw's user-data block is LARGER than the bound pipeline's USER_SGPR window -> descriptor resolution reads a V# tail as a pointer (blocks Nikoderiko's 3D world)


mattias800:
> **Retitled 2026-08-01.** The original title — *"first draws of some UI submits run with the PREVIOUS
> pipeline's PGM + user data (missing bind in decoded stream)"* — asserted a mechanism that has since
> been **measured and falsified**. The observable condition below is what actually holds. Do not start
> from "stale bind"; that tree has been climbed. Full measurement:
> [#305 findings comment](https://github.com/mattias800/prosper/issues/305#issuecomment-5150339410).

## The condition

A graphics stage resolves garbage descriptors **exactly when the user-data block the guest most
recently programmed is LARGER than the bound pipeline's user-SGPR window**
(`SPI_SHADER_PGM_RSRC2_{GS,PS}.USER_SGPR`, which equals the shader's own
`user_data_range_end`). The shader then dereferences a V# `num_records`/`dword3` tail as a pointer —
the `0x0004dfac…` / `0x0001d22c…` constant family this issue has always recorded — the const-fold
correctly refuses to invent a descriptor, and the draw is skipped fail-visibly.

Measured on Nikoderiko (`PPSA23760`), 460 s instrumented boots:

| stage | `user_data_range_end` | `USER_SGPR` | dwords the last bind programmed | outcome |
|---|---|---|---|---|
| VS `300c5c0000` | 12 | 12 | 12 | resolves |
| VS `300e710000` | 8 | 8 | **12** | fails |
| VS `3007cf0000` | 12 | 12 | **30** | fails |
| VS `30853b0000` | 24 | 24 | **28** | fails |
| VS `300fc00000` | 20 | 20 | **30** | fails |
| VS `30856a0000` | 24 | 24 | **30** | fails |
| VS `300b430000` | 12 | 12 | **28** | fails |

The original DOLL (`PPSA17942`) observation in this issue is the **same condition**: *"the user-data
`SET_SH_REG` immediately preceding them carries a 12-dword block … that matches a 12-user-SGPR
pipeline, while the stale bound VS … is an 8-user-SGPR shader"*. Two titles, same numeric
relationship. The disagreement is only about the mechanism.

## Impact

* **Nikoderiko**: the entire 3D world. Title screen and EULA render correctly at 3840×2160; 25 distinct
[...]

PR #3191, opened 2026-09-01
recompiler: name the failing provenance step on a vertex-fetch reject (#3137)

mattias800:
## What this changes

`Refs #3137` — deliberately not `Fixes`, and the reason is the whole point of the PR: **#3137's
premise is wrong, and the defect it describes is not in the recompiler.** This lands the
instrumentation change that makes that legible from one log line, the regression test that pins it,
and the falsifications so the next reader does not re-derive any of it.

## The failure scenario

*Stray* (`PPSA02101`) drops two title-screen vertex stages, each at a vertex attribute fetch:

```
[recompile-reject] sh=bfa00003/77  mode=unresolved-operand pc=57 words=e00c2000,6a010000
                   fmt=12 op=0x3 dst=0(kind2) src=0(k2),4(k1),106(k6) stage=vertex role=terminal
[recompile-reject] sh=bfa00003/221 mode=unresolved-operand pc=66 words=e0042000,6b020900
                   fmt=12 op=0x1 dst=9(kind2) src=0(k2),8(k1),107(k6) stage=vertex role=terminal
```

That line names an **instruction**, so it reads as a missing lowering. #3137 proposed the two
hypotheses the line can support. Measured live on the issue's own route, **both are false**:

| #3137's candidate | Verdict |
|---|---|
| the `VCC_LO`/`VCC_HI` SOFFSET does not resolve | **False.** `operand_bits` has a vertex-stage branch for Special 106/107 (`rdna2_alu_support.hpp:262`), and in any case the reject fires *before* the address is formed. |
| the stage arrived with no resource table (`allow_smem = rt != nullptr`) | **False.** From the same run: `[buf-op] program=0x0 pc=57 MUBUF op=0x3 n=4 store=0 atomic=0 rt=1 reject-unresolved` — table present, **five** resources. |

MUBUF `0x0`–`0x3` are all lowered. What fails is SRSRC provenance. Disassembled with
`llvm-mc -arch=amdgcn -mcpu=gfx1030`, both stages assemble their own descriptor:

```
s_load_dwordx4  s[4:7], s[12:13], vcc_hi        ; register SOFFSET -> no SRT key
s_and_b32       vcc_lo, s7, 0xfff80000
s_or_b32        vcc_lo, vcc_lo, s13
s_cselect_b32   s7, s7, vcc_hi                  ; patches V# dword3
buffer_load_format_xyzw v[0:3], v0, s[4:7], vcc_lo idxen
```

so `sreg_range_written` is true and the two direct-SGPR fallbacks are correctly suppressed; the
register-offset load leaves no SRT tag; and the const-fold published no per-fetch entry. Every
provenance route is legitimately exhausted, and rejecting is the correct behaviour.

**Why the const-fold has nothing: this is #305.** `PROSPER_DYNTRACE_FAIL=1` shows the chain hanging
off one load —

```
[dyntrace] SMEM op=0x2 load sdst=s16 sbase=s14 base=0x4dfac00000001 base_ok=1 ... n=4
[dyntrace]   addr 0x4dfac00000000 unreadable
...
[dyntrace] MUBUF fetch pc=66 ... have_descr=0 soff_known=0
[dyntrace]   MUBUF pc=66 SOFFSET untracked -> fetch left unresolved (not folded to 0)
```

`0x0004dfac…` is the constant family `RESOURCE_BINDING.md` already names for #305. `PROSPER_UDPROV=1`
completes it: the draw's own bind writes **twelve** contiguous user dwords in one direct
`SET_SH_REG` (`fresh_extent=12`, all sharing write order `d3524877`), while
`RSRC2_GS.USER_SGPR = 8 = user_data_range_end = 8`. The two mapped 64-bit pointers the shader wants
sit at dwords 8 and 10 — outside the window — and dwords 6:7 hold the `num_records`/`dword3` tail of
the preceding V#, which the shader then dereferences.

So *Stray* is a **third confirmed title** for #305, after Nikoderiko (#1607) and DOLL, and the first
where one replay shows both halves.

## The contract this PR changes

The FORMAT-path MUBUF reject diagnostic printed strictly fewer fields than the RAW path sitting a
few dozen lines below it. `[mubuf-raw-unresolved]` has always reported `rewritten=` and the
per-fetch lookup; `[mubuf-unresolved]` reported neither — and those are exactly the two facts that
separate the remaining cases:

* `rewritten=1` — the shader assembled its own SRSRC. Direct-SGPR provenance was *suppressed*, not
  missed; only per-fetch or SRT-tag provenance could have resolved it. A front-half / user-data
  verdict.
* `rewritten=0` — the SRSRC still holds entry user data, both direct fallbacks **were** tried, and
  the table genuinely has no entry at that SGPR.
* `pc_res=` — whether the const-fold published a descriptor for this exact fetch instruction.

New line shape:

```
[mubuf-unresolved] pc=57 srsrc=s4 srt_tag=NONE 0x0 key_res=null pc_res=null rewritten=1 (5 res)
```

## Approach and invariants

* `srsrc_rewritten` is hoisted out of the `if (rt)` block in the `is_format` arm so the reject site
  can report it. It is computed from `sreg_range_written`, which does not consult `rt`, so the value
  is identical to before at every existing use.
* `pc_res` is a second `rt->by_fetch_pc(in.pc)` lookup **inside the `getenv("PROSPER_DBG")` guard**,
  so the non-diagnostic path is untouched.
* **No resolution behaviour changes.** An unresolvable V# still rejects the stage fail-visibly; this
  PR does not widen a reject path and does not narrow one. Both regression arms assert the reject is
  still produced.

* Affected: the text of `[mubuf-unresolved]`, which is emitted only under `PROSPER_DBG`.
* Unaffected: SPIR-V emission, descriptor resolution order, the raw-MUBUF path, MTBUF, MIMG, SMEM,
  and every non-diagnostic code path. `git diff` on the emitter is one hoisted `const bool`, one
  extra guarded lookup, and two format fields.
* **Not fixed here:** the underlying #305 user-data window. That is a cross-title frontier with an
  existing umbrella issue and is out of scope for a recompiler-diagnostic PR.

## Risks

* Low. The only behavioural surface is a diagnostic string, and any tooling that parses
  `[mubuf-unresolved]` positionally would see two new fields before the trailing `(N res)`. A repo
  grep finds no parser for this line.

## Docs

* `RESOURCE_BINDING.md` § Ruled out — Stray added to the #305 umbrella with the measurement above;
  two new rows: (a) both of #3137's candidates, falsified; (b) **the `[udcand]` "implied seed"
  offset is not a safe basis for re-seeding** — see below, this is the more valuable one.
* `STRAY_STATUS.md` § Ruled out — the title-local version, so nobody restarts from the recompiler.
* `GAME_COMPAT_ORCHESTRATION.md` — instrument trap **241**: when two sibling diagnostics report the
  same class of failure, diff their format strings before trusting the poorer one.
* `BLOG.md` — one short no-picture entry (a correction, not a rung).

### The falsification worth reading

The obvious fix for #305 is already computed by the existing `[udcand]` instrument: search the
32-dword window for the one seed offset at which every declared direct pointer becomes mapped. On
Stray it reports `seeded=0 implied=4`, and at seed 4 both pointers do map. **It is still wrong.**
The same shift moves `s[8:11]` onto a `stride=16 / num_records=1` V# — 16 bytes — while the shader's
own `s_buffer_load_dwordx4 s[20:23], s[8:11], 0x30` reads at byte 48 and its
`s_buffer_load_dwordx8 s[0:7], s[8:11]` reads 32. At seed 0 those same registers decode as
`stride=8 / num_records=376` (512 bytes), both accesses are in bounds, and the fold prints the
loaded bytes as plausible floats. The pointer-mapping criterion is satisfiable by an offset the
shader's own accesses refute.

Separately: `PROSPER_UD_TAIL_ALIGN=1` was run live as an A/B and is **not** the fix, consistent with
its existing falsification — `[udtail] VS … user_sgprs=8 range_end=8 … prefix_rsrc2=0` for exactly
the failing stages, i.e. the lever is a **no-op** on them. Stated as a no-op rather than as a
negative result, because a no-op arm proves nothing about the hypothesis.

## Verification

Built and tested in the `ps5ys` distrobox, Linux, `prosper/build-linux`, `-DGAME_DUMP=<DUMP_ROOT>/PPSA24651-app0`.

```
TMPDIR=$PWD/prosper/build-linux/tmpdir cmake --build prosper/build-linux -j8      # exit 0
ctest --test-dir prosper/build-linux --no-tests=error                            # exit 0
```

`100% tests passed out of 319` (318 on master + the one added here), exit code **0**, run twice on
the rebased head `34fd9f6a`.

One intermediate full run reported `99% tests passed, 2 tests failed out of 319`
(`recompiled_shaders_render`, `multidraw_render`). Both are Vulkan-execution cases, both passed on
`--rerun-failed` immediately after, and both passed in the clean 319/319 run that followed. Recorded
rather than quietly re-run: they are flakes on this box, not results, and neither touches any line
this PR changes.

Docs gates, both green:

```
    --baseline <origin/master copy> prosper/docs/GAME_COMPAT_ORCHESTRATION.md
#  -> 14 table(s), 302 rows, unbroken; numbered column 1..240, unique and ascending,
#     none of the 239 base rows deleted
```

`git diff --cached --check`: clean. `git log origin/master..HEAD --format=%B | grep -c '^Claude-Session:'`: 0.

### Mutation verification

`tests/gpu/recompiler/test_vertex_fetch_reject_diagnostic.cpp` builds two vertex shaders whose
encodings were round-tripped through `llvm-mc -arch=amdgcn -mcpu=gfx1030`: one that writes
`s[4:7]` and fetches through it (`rewritten=1`), one that fetches through untouched entry user data
in `s[8:11]` (`rewritten=0`), both against a present-but-non-matching resource table. It asserts
both still reject, and that the two arms report **opposite** `rewritten=` verdicts — so the field
cannot be satisfied by a constant.

Mutation: the two new format fields reverted in `rdna2_emit_alu.cpp`, everything else untouched.

| arm | build rc | test rc |
|---|---|---|
| baseline (fix in place) | 0 | **0** — 10/10 checks pass |
| mutant (diagnostic reverted) | **0** | **1** — 5 checks fail |
| restored | 0 | **0** |

The mutant build's exit code is recorded because a mutation that fails to compile leaves the old
binary in place and reports a false pass; it compiled.

## What was NOT verified

* No snapshot run (one GPU consumer at a time; other lanes were active). The diff touches no
  rendering path.
* The #305 root cause is **diagnosed, not fixed**. Stray's two vertex stages still reject on this
  branch, exactly as before — by design.


mattias800 (commented):
**APPROVED**

**Mutation verified independently.** Making the `rewritten=` field a constant in the FORMAT-path
reject line fails the test (`rc=8`) on a clean build (`build_rc=0`, 0 errors — not a stale binary);
restored, `rc=0`. I also checked the hoist is safe before that: `sreg_range_written` is a pure set
lookup over `rs.sreg_written`, so lifting it out of the `if (rt)` block cannot change behaviour, and
the three existing uses are unchanged.

**The most valuable thing here is that the issue's premise is wrong, and you measured both halves
rather than arguing them.**

- The **VCC soffset** candidate: `operand_bits` has a vertex-stage branch for Special 106/107
  (`rdna2_alu_support.hpp:262`), and the reject fires *before* the address is formed — SOFFSET is
  read at `rdna2_emit_alu.cpp:6360`, the reject is at `:5936`. It cannot be the cause by ordering.
- The **null resource table** candidate: the live run says `rt=1` with five resources.

Both were plausible, both are dead, and the real cause bottoms out in **#305** — making *Stray* a
third confirmed title after Nikoderiko and DOLL. `Refs #3137` rather than `Fixes` is the right call
and the PR says why in its first line.

**The test's anti-vacuity property is what makes it worth having.** Two arms with *opposite*
`rewritten=` verdicts, plus an explicit assertion that the two disagree — so the field cannot be
satisfied by a constant, which is precisely what my mutation confirms. Most diagnostic tests assert
that a string appears; this one asserts it *discriminates*.

**The falsification you recorded is the expensive kind and it cuts against the attractive answer.**
The `[udcand]` "implied seed" offset looks like a fix for #305: at seed 4 both declared pointers map.
But the same shift moves `s[8:11]` onto a 16-byte V# while the shader's own
`s_buffer_load_dwordx4 s[20:23], s[8:11], 0x30` reads at byte 48; at seed 0 those registers decode as
512 bytes and every access is in bounds. **Pointer-mappedness is satisfiable by an offset the
shader's own accesses refute.** Someone would have tried that.

**Why the diagnostic earns its place**: settling this required re-running the title under two more
instruments to learn what one log line could have said. The RAW path has always printed `rewritten=`
and `pc_res=`; the FORMAT path did not, and those two fields are exactly what separates "the table
has no entry" from "provenance was correctly suppressed" — which point at different files.

**One correction to the PR's own record**, worth folding in rather than leaving: the intermediate run
reporting 2 failures (`recompiled_shaders_render`, `multidraw_render`) is almost certainly **not** a
flake. Those are two of the five tests that instrument trap 243 (merged today) documents a stale
incremental build tree producing, and this branch was rebased in place. Calling it a flake in a
merged PR invites the next reader to shrug at a real failure; the named cause tells them to
reconfigure. No action needed on the branch — recording it here.

Verified: `PROSPER_UD_TAIL_ALIGN=1` correctly reported as a **no-op** on these stages
(`prefix_rsrc2=0`) rather than as a negative result. CI 9 pass / 2 skipping. No `Claude-Session:`
trailer, no absolute host paths.

PR #3192, opened 2026-09-01
perf(render): take getenv off 44 per-draw guards, screened against runtime arming

### LINKED ISSUE #3094: perf(render): getenv is 1.24% of the Blue Prince gameplay frame — 130 uncached PROSPER_* reads in per-draw paths
`getenv` is **1.24% of Blue Prince's gameplay frame** on Linux/AMD — measured, not inferred, from a
15 s `perf record` on a live interactive session at ~20.7 fps (65,756 samples; 38.67% lost, so treat
the share as approximate). Full decomposition in #1284.

That is pure overhead: `getenv` is a linear scan of the environment block doing `strncmp` per entry,
and prosper's environment is large. It buys nothing on any call after the first, because every one
of these variables is read-only for the process lifetime.

## Where

Counting `getenv("PROSPER_…")` call sites that are **not** hoisted into a `static`/`constexpr`
initializer, on `aced0703`:

| file | uncached | total |
| --- | --- | --- |
| `prosper/tests/fixtures/render_runner.h` | **66** | 88 |
| `prosper/frontends/shared/live/live_renderer.cpp` | **41** | 61 |
| `prosper/src/gpu/execute/gpu_execute.hpp` | **23** | 27 |

Several sit in per-draw or per-pass paths. Examples from `render_runner.h`:

```
1284:   if (getenv("PROSPER_GFXLOG"))
1287:   } else if (getenv("PROSPER_GFXLOG")) {
 713:   return getenv("PROSPER_NO_BACKEND_PIPELINE_CACHE") == nullptr;
 758:   return getenv("PROSPER_NO_BACKEND_PIPELINE_LAYOUT_CACHE") == nullptr;
```

## Failure scenario

State: any title rendering at a normal draw rate. A per-draw diagnostic gate such as
`if (getenv("PROSPER_GFXLOG"))` executes a full environment scan for **every draw of every frame**,
whether or not the variable is set. At Blue Prince's ~350 operations/frame and 20 fps that is
thousands of scans a second to answer a question whose answer cannot change. Wrong behaviour: none —
this is purely wasted time, which is why nothing has ever flagged it.

## Suggested fix

The codebase already has the correct idiom in places, e.g. `render_runner.h:3401`:

```cpp
static const bool htile_invalidates = []() {
    const char* v = getenv("PROSPER_DS_HTILE_INVALIDATE");
    return !(v && v[0] == '0');
}();
```

Hoist the hot ones to the same shape. This is mechanical and low-risk, but it is **not** a blanket
sweep: a few sites are deliberately re-read (test fixtures that mutate the environment between
cases, e.g. anything a ctest sets with `setenv` mid-run). Those must stay dynamic, so each site needs
a glance rather than a regex. Prioritise the per-draw/per-pass gates; init-time reads are irrelevant.

A guard is worth adding so this does not regrow: the profile is the only thing that noticed.

mattias800:
## What this fixes

`getenv` is 1.24% of Blue Prince's gameplay frame (#3094, decomposition in #1284). It is a linear
scan of the environment block, and prosper evaluates diagnostic guards like
`if (getenv("PROSPER_PIPELOG"))` on per-draw and per-resource paths — thousands of scans a second to
answer a question whose answer cannot change. No wrong behaviour, just wasted time, which is why
nothing had ever flagged it.

Fixes #3094.

## Failure scenario

A title rendering at a normal draw rate. Every draw evaluates a set of diagnostic gates; each gate
that is still a live `getenv` walks the whole environment block comparing entries, whether or not
the variable is set. Cost scales with scene complexity rather than with logging.

## The contract, and why this is not a regex sweep

Caching changes semantics: the variable is sampled at first use, so a process that arms it later
never observes the write. The dangerous part is the failure mode — a test that arms a diagnostic
with `setenv` and asserts on the behaviour does **not** go red when the read is cached, it goes
**vacuous** and keeps printing `[ok]` against the stale value (#2214).

**All four call sites the issue nominated as the obvious fix are unsafe, and are deliberately left
as live `getenv`:**

| site | name | why it must stay live |
| --- | --- | --- |
| `render_runner.h:1285`, `:1288` | `PROSPER_GFXLOG` | armed by `tests/gpu/test_eop_write.cpp:254` and 9 other sites |
| `render_runner.h:714` | `PROSPER_NO_BACKEND_PIPELINE_CACHE` | armed by `tests/gpu/state/test_multidraw_render.cpp:248` |
| `render_runner.h:759` | `PROSPER_NO_BACKEND_PIPELINE_LAYOUT_CACHE` | armed by `tests/gpu/state/test_multidraw_render.cpp:390` |

So each site was screened rather than rewritten. Screening used two filters:

1. **`tools/env/check_cached_env.py`'s arming scan** — 128 `PROSPER_*` names are written at runtime
   somewhere in the tree. None of them are cached here.
2. **A blind spot in that gate, found while doing this.** `tools/gpu_replay` re-applies
   `GpuCaptureMetadata::renderer_env` — the `render_env[]` allowlist at `gpu_capture.cpp:6237`, 37
   names — through `set_environment(name, value)` with a **non-literal** name, once per bundle
   submit. `check_cached_env.py` matches on literal first arguments, so it cannot see any of them.
   Those 37 names are excluded from this change too.

   To be precise about what that blind spot is and is not: **no defect is demonstrated.** Those
   values are captured from the capturing process's own environment, which is constant for that
   run, so all submits in a bundle carry identical values; and the one genuinely submit-local entry,
   `PROSPER_GPU_REPLAY_SCANOUT_ADDR`, is not cached anywhere. 17 of the 37 are already cached on
   master and I am **not** calling those a bug. They are excluded here because a bundle replay is
   the project's primary offline debugging tool, and a diagnostic silently frozen across submits
   would be a trap for a future reader — not because anything is known to be broken. The reasoning
   is recorded in `env_cache.hpp` so the next person does not have to re-derive it.

## Approach

- **`src/diagnostics/env_cache.hpp` (new).** `PROSPER_ENV_ON` / `PROSPER_ENV_VALUE` moved verbatim
  out of `hle/dispatch/dispatch.hpp`, which is where they were first needed and not where they
  belong — the renderer backend and the draw executor should not include the HLE dispatch registry
  (`self/module.hpp`, the NID tables) to ask whether a switch is on. `dispatch.hpp` includes the new
  header, so its ~150 existing macro users are untouched. `live_renderer.cpp` now includes the small
  header instead of the HLE one.
- **44 call sites converted** across `tests/fixtures/render_runner.h` (21),
  `src/gpu/execute/gpu_execute.hpp` (20) and `frontends/shared/live/live_renderer.cpp` (3). Boolean
  guards became `PROSPER_ENV_ON`; sites binding the pointer became `PROSPER_ENV_VALUE`.
  Note `render_runner.h` is the shared renderer backend despite its path — it is included by
  `live_renderer.cpp`, `gpu_executor.cpp`, `live_compute.cpp` and `gpu_replay`.
- **The historical warning is corrected, not dropped.** The old comment said applying these macros
  to `hle_kernel.cpp`, `gpu_executor` or the command processor broke `win_exception_delivery`,
  `eop_write` and `dynfetch_fold`. That was never a property of those *files* — it was the specific
  names they read, all of which the gate now refuses. This PR caches 16 names in `gpu_execute.hpp`,
  on the same per-draw path reached from `gpu_executor.cpp`, with the suite green. Screen by name.

## Invariants

- No name written at runtime anywhere in the tree is cached (gate tier 1/2).
- No `render_env[]` name is cached.
- The cache unit is the **call site**, not the name: each expansion has its own function-local
  static. Two sites reading one name read once each. C++11+ makes function-local static init
  thread-safe; storage is deliberately not at namespace scope, which would also expose it to
  cross-TU static-init order.

## Regression tests (both mutation-verified)

**`diagnostics_env_cache`** (new, `tests/diagnostics/test_env_cache.cpp`) — behavioural. Each case
evaluates **one** call site twice with the environment mutated in between and asserts the second
evaluation still reports the first value. That is the one behaviour a live `getenv` cannot produce.
Both directions are covered (armed-then-disarmed, and unset-then-armed, so "always true" cannot
pass), each with a live-`getenv` control arm proving the mutation actually took effect.

This test found a real subtlety while being written: the first draft asserted across two *different*
macro expansions and failed, because those are two independent caches. The contract is "once per
site", never "once per name", and the header now says so.

**`cached_env_arming_logic` tier 3** (extends `tools/env/check_cached_env.py`) — a `HOT_SITES` table
naming the 36 (file, name) pairs converted here. It fails if any of them regrows a live `getenv`,
and also fails if an entry goes stale. Six self-test cases cover the shapes it must and must not
match, including `static const bool x = getenv(...)` — the older hand-rolled one-shot spelling still
in the tree, which must not be flagged. Without those the tier could silently report a clean tree
forever.

## Measurement — honest scope

I did **not** re-measure the 1.24% frame share; that number stays the issue's, not mine. What I did
measure is the mechanism directly, with an `LD_PRELOAD` shim counting `getenv` calls by name across
eight render/execute tests, same binaries and same command before and after:

```
getenv calls  BEFORE 29539   AFTER 20626   delta -8913 (-30.2%)
```

Per-name, the converted guards drop to exactly **8 calls each — one per process across 8 test
processes**, which is the reachable floor and confirms one read per site per run:

```
PROSPER_BACKEND_LOAD_LOG          1240 ->  8
PROSPER_GEOM_PROBE                 836 ->  9
PROSPER_DSLOG                      819 -> 13
PROSPER_RENDER_DROP_UNPROVEN_DRAW  815 ->  8
PROSPER_DEPTH_CLEAR / _WHY         812 ->  8   (each)
PROSPER_DRAW_STATS                 812 ->  8
PROSPER_PIPELOG                    812 ->  8
PROSPER_BACKEND_TRACE              812 ->  8
PROSPER_NO_BACKEND_BUFFER_ARENA    807 ->  8
```
No name increased.

**What this is not:** eight render tests are a proxy for a Blue Prince gameplay frame, not that
frame. The call-count reduction is measured and reproducible; the resulting *frame-time* improvement
on a real title is **unverified here**. Roughly half the per-draw `getenv` volume in these tests
(`PROSPER_NO_STENCIL`, `PROSPER_NO_BLEND`, `PROSPER_NO_CULL`, `PROSPER_FLIP_FRONT_FACE`,
`PROSPER_STENCIL_CLEAR`, `PROSPER_NO_DEPTH`, `PROSPER_DRAW_ISO` — ~7,000 calls) is deliberately left
live because those are the `render_env[]` names, so a further win exists behind the question above.

## Affected / unaffected

- **Affected:** 44 diagnostic guards now read once per call site per process instead of per
  draw/resource. If any of these is armed *after* first evaluation, it is no longer observed — that
  is the intended trade and no test or tool in the tree does it.
- **Unaffected:** every armed name, every `render_env[]` name, all ~150 existing `PROSPER_ENV_*`
  users, and the two macros' definitions (moved verbatim, byte-identical bodies). No rendering
  behaviour changes when no diagnostic is set.

## Risks

- A future contributor arms one of the 44 cached names at runtime and gets a vacuous test rather
  than a failure. Mitigated by gate tier 1/2, which fails on exactly that.
- `HOT_SITES` goes stale if a file is renamed. The check fails loudly (missing file / entry present
  but unread) rather than passing quietly.
- The `render_env[]` blind spot remains unguarded in the gate. Deliberate: encoding it would flag 17
  existing cached names for which no defect is demonstrated, and a gate with standing false
  positives is one people learn to skip. Recorded in `env_cache.hpp` and `src/diagnostics/AGENTS.md`
  instead.

## Commands and results

```
cmake -S prosper -B prosper/build-linux -DCMAKE_BUILD_TYPE=RelWithDebInfo   # rc 0
TMPDIR=$PWD/prosper/build-linux/tmpdir cmake --build prosper/build-linux -j8   # rc 0
      -> 100% tests passed, 319 tests, 0 failed, exit code 0
python3 prosper/tools/env/check_cached_env.py    # rc 0; "36 hot site(s) across 3 file(s)"
python3 prosper/tools/docs/check_numbered_table.py BLOG.md prosper/src/diagnostics/AGENTS.md  # rc 0
git diff --check   # rc 0
```

Built and tested in the project distrobox (Linux/AMD, Vulkan present — the offscreen graphics
harness is enabled, so the Vulkan execution tests really ran).

### Mutation evidence

**1. Remove the caching** — drop `static` from both macro bodies in `env_cache.hpp`:

```
baseline_rc 0  /  mutant_rc 1  (5 assertions failed)  /  restored_rc 0
mutant build rc 0   <- the mutant binary really was rebuilt, so the failure is the code, not a stale build
```

**2. Regrow a live getenv** — revert `render_runner.h:8368` to `if (getenv("PROSPER_PIPELOG"))`:

```
baseline_rc 0  /  mutant_rc 1  /  restored_rc 0
mutant output: [FAIL] tests/fixtures/render_runner.h: PROSPER_PIPELOG is read with a live getenv at line(s) 8368
```

## Docs

`src/diagnostics/AGENTS.md` gains a paragraph on `env_cache.hpp` — why it lives there, and the
caching-changes-semantics warning with the gate that enforces it. `BLOG.md` gets a short entry (no
screenshot: this is CPU time, nothing new renders).

### DIFF
--- a/BLOG.md
+++ b/BLOG.md
@@ -19,6 +19,26 @@ from the tracker issues, and still gated, because it is a projection of state ra
 > title's current state — for that, read the tracker. Nothing is ever removed when a title moves on,
 > because the point of a blog is that it records *when* things happened.
 
+## 2026-09-01
+
+### The three getenv calls the profiler pointed at were the three we could not fix
+
+No picture — this one is pure CPU time. `getenv` costs 1.24% of Blue Prince's gameplay frame
+because a per-draw guard like `if (getenv("PROSPER_GFXLOG"))` rescans the whole environment block
+for every draw, to answer a question whose answer never changes.
+
+The interesting part is what the sweep found. Every one of the four call sites [#3094][i3094]
+nominated as the obvious fix turns out to be unsafe to cache: three are armed at runtime by tests
+that toggle them between phases, and caching those does not make a test fail — it makes it go
+*vacuous*, still printing `[ok]` against a stale value. A fourth class is worse, because no gate
+can currently see it: `gpu_replay` re-applies a whole allowlist of renderer switches once per
+bundle submit, through a variable rather than a literal. So the sweep became a screening problem
+rather than a mechanical one. Forty-two sites across the renderer backend, the draw executor and
+the live frontend now read once instead of per draw, measured at 30% fewer `getenv` calls across
+the render test suite, and the sites are pinned so they cannot quietly grow back.
+
+[i3094]: https://github.com/mattias800/prosper/issues/3094
+
 ## 2026-08-31
 
 ### Stray's splash runs 66% faster, and the title screen now holds 65 fps
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -1362,6 +1362,14 @@ if(TARGET prosper_core)
   target_link_libraries(test_boot_phase_log PRIVATE prosper_core)
   add_test(NAME diagnostics_boot_phase_log COMMAND test_boot_phase_log)
 
+  # PROSPER_ENV_ON / PROSPER_ENV_VALUE sample getenv exactly once (#3094). Header-only, so this
+  # needs no library: it mutates the environment between two evaluations of the same macro and
+  # asserts the second still reports the first value -- the one behaviour a live getenv cannot
+  # produce, so the arms go red if the caching is ever removed.
+  add_executable(test_env_cache tests/diagnostics/test_env_cache.cpp)
+  target_include_directories(test_env_cache PRIVATE src)
+  add_test(NAME diagnostics_env_cache COMMAND test_env_cache)
+
   # Diagnostics: enabled-path and JSON tests (only when feature is built).
   if(PROSPER_DIAGNOSTICS)
     add_executable(test_diagnostics_enabled tests/diagnostics/test_diagnostics_enabled.cpp)
--- a/prosper/frontends/shared/live/live_renderer.cpp
+++ b/prosper/frontends/shared/live/live_renderer.cpp
@@ -1,7 +1,7 @@
 // live_renderer.cpp — see live_renderer.hpp. Extracted from boot_trace's PROSPER_RENDER lambda
 // (behavior-preserving); Vulkan-backed, so this unit links Vulkan::Vulkan.
 #include "shared/live/live_renderer.hpp"
-#include "hle/dispatch/dispatch.hpp"   // PROSPER_ENV_ON / _VALUE: cached reads on per-draw paths
+#include "diagnostics/env_cache.hpp"   // PROSPER_ENV_ON / _VALUE: cached reads on per-draw paths
 #include "gpu/resources/metadata_kind_correlation.hpp"  // positive metadata-kind correlation (pure, tested)
 #include "gpu/diagnostics/watch_list.hpp"                 // strict 0x-only watch parsing
 #include "shared/rtt/rtt_authority.hpp"
@@ -2368,7 +2368,7 @@ void register_live_renderer(const std::string& frame_dir, bool dump_bmps_request
                         // colour with no content -- which is exactly Little Nightmares III's
                         // uniform-yellow presents (#2014). Deduped per (address, decoded colour) so a
                         // run costs a handful of lines.
-                        if (const char* dcclog = std::getenv("PROSPER_DCCLOG")) {
+                        if (const char* dcclog = PROSPER_ENV_VALUE("PROSPER_DCCLOG")) {
                             if (dcclog[0] == '1' && dcclog[1] == '\0') {
                                 static std::mutex dcc_mutex;
                                 static std::set<std::pair<uint64_t, uint32_t>> dcc_seen;
@@ -5431,7 +5431,7 @@ void register_live_renderer(const std::string& frame_dir, bool dump_bmps_request
                             // measures, and a diagnostic that prints a field a different switch
                             // decided is one whose output cannot be trusted on its own -- which
                             // check_diag_gates.py enforces (TWO-GATE).
-                            if (!is_cube && getenv("PROSPER_SLICESTRIDE")) {
+                            if (!is_cube && PROSPER_ENV_ON("PROSPER_SLICESTRIDE")) {
                                 static std::unordered_set<uint64_t> reported;
                                 if (reported.insert(r.gpu_addr).second)
                                     fprintf(stderr,
@@ -9532,7 +9532,7 @@ void register_live_renderer(const std::string& frame_dir, bool dump_bmps_request
                             // substitution happened, not what was substituted or where it came from.
                             // Sampled, not exhaustive: a full uniformity scan of a 33 MB frame on the
                             // present path would cost more than the render.
-                            if (const char* uni = std::getenv("PROSPER_UNIFORMLOG")) {
+                            if (const char* uni = PROSPER_ENV_VALUE("PROSPER_UNIFORMLOG")) {
                                 if (uni[0] == '1' && uni[1] == '\0' && selected_pixels &&
                                     selected_pixels->size() >= 4) {
                                     const uint8_t* px = selected_pixels->data();
--- a/prosper/src/diagnostics/AGENTS.md
+++ b/prosper/src/diagnostics/AGENTS.md
@@ -44,3 +44,22 @@ published conclusion on #3142. `diag_now_us()` gives them a common monotonic rea
 is process-wide, so independently taken values in unrelated translation units compare directly with
 no anchor or registry. Anything that needs to be time-ordered against a diagnostic in another
 subsystem should stamp it rather than grow a private clock.
+
+`env_cache.hpp` is the third always-reachable file, and it is here for the same reason as
+`diag_clock.hpp`: it is shared, and it belongs to no one subsystem. It holds `PROSPER_ENV_ON` /
+`PROSPER_ENV_VALUE`, the one-shot reads of a `PROSPER_*` switch. They lived in
+`hle/dispatch/dispatch.hpp` until #3094, which is where they were first needed and not where they
+belong — the renderer backend, the draw executor and the live frontend all evaluate diagnostic
+gates on per-draw and per-resource paths, and none of them should pull in the HLE dispatch
+registry to ask whether a switch is on. `dispatch.hpp` includes this header, so its existing users
+did not change.
+
+The thing to know before using them: **caching a diagnostic switch changes its semantics**, and it
+fails in the quiet direction. The variable is sampled at first use, so a process that arms it later
+never observes the write — and a test that arms a diagnostic and then asserts on the behaviour does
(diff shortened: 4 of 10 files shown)


mattias800 (commented):
**APPROVED**

**The finding that changed the task is the reason this is a good PR.** All four call sites the issue
*nominated as the fix* turn out to be unsafe to cache, and you left them as live `getenv` and said so.
Caching them would not have failed a test — it would have made `test_eop_write.cpp` and
`test_multidraw_render.cpp` go **vacuous**, passing while asserting nothing. An issue that names its
own fix is the easy kind to close by doing what it says; checking whether what it says is right is
the work.

**I verified the safety mechanism independently, because the whole PR rests on it.** Converting
`render_runner.h:1285`'s `PROSPER_GFXLOG` — armed at runtime by `test_eop_write.cpp` — to the cached
macro:

| | result |
| --- | --- |
| gate on the mutated tree | `rc=1`, naming `test_eop_write.cpp:1147` and `:1149` as the arming lines |
| gate restored | `rc=0` |
| registered as a ctest case | yes |

So the guard against exactly this mistake is load-bearing, not decorative. (A note on my own method:
my first run of that check reported `rc=0` on the mutant — I had read `tail`'s exit code through a
pipe rather than the script's. The charter's "a pipeline's exit status is its last stage's", and I
walked straight into it. The numbers above are from the corrected run.)

**Not filing a bug for the `gpu_replay` blind spot is the right call and the harder one.** The gate's
literal-argument matching cannot see `set_environment(name, value)` with a non-literal name, so those
37 `render_env[]` entries are invisible to it. You checked, found the values come from the capturing
process's own constant environment, found the one genuinely submit-local entry
(`PROSPER_GPU_REPLAY_SCANOUT_ADDR`) is not cached, and recorded the reasoning in `env_cache.hpp`
instead of filing a defect you could not demonstrate. A speculative issue would have cost the next
person a re-derivation to reach the same "no defect" conclusion.

**Moving the macros out of `hle/dispatch/dispatch.hpp` into `src/diagnostics/env_cache.hpp` is the
charter's folder rule applied correctly** — the renderer backend should not include the HLE dispatch
registry to ask whether a switch is on — and doing it as a verbatim move with `dispatch.hpp`
including the new header keeps ~150 existing users untouched.

**Correcting the old "don't apply this to `gpu_executor`" warning rather than obeying it.** You
established it was about the *names* those files read, not the files themselves, and then cached 16
names on that same path. Re-deriving a stale warning's actual scope instead of treating it as a
boundary is what keeps warnings from accreting into no-go zones.

**"The cache is per call site, not per name"** — caught by writing the test, and now stated in both
the header and the test. Two textual expansions are two lattices of static storage; a reader who
assumes per-name will write a wrong test and not notice.

**Honest gaps accepted as stated:** the 1.24% frame share is the issue's number and not re-measured,
and you say so rather than adopting it; ~7,000 calls/run remain behind the `render_env[]` question.
The measurement you *did* make is the right one and is falsifiable — **29,539 → 20,626 getenv calls
(−30.2%)** under an `LD_PRELOAD` counter, with converted guards dropping to exactly 8 calls across 8
processes, i.e. one per process. That is the claim the change actually supports.

Verified here: `ctest --no-tests=error` **319/319, exit 0** on a clean configure. CI 9 pass / 2
skipping. No `Claude-Session:` trailer, no absolute host paths.

Issue #3193, opened 2026-09-01
Astro Bot (PPSA21564): the world-map GPU hang is ONE graphics draw — vertex program 0x5008efd00


mattias800:
## Summary

Astro Bot's world-map GPU reset is caused by **one graphics draw**, and it can now be named:
the draw whose **vertex/ES program is at guest `0x5008efd00`** (pixel program `0x5008f1400`).

Declining that single program with the new `PROSPER_SKIP_DRAW_PROGRAM` selector removes the device
loss entirely on a 300 s default-route run — **0 losses against a deterministic loss at ~75 s on the
control**, replicated, with a complementary control that keeps the loss.

This confirms the 2026-08-31 finding on #1809 that the reset is a **draw**, not compute: the
`VK_ERROR_DEVICE_LOST` line names compute program `0x5006c6a00`, but that is the first submit to
fail *after* the context was already lost. The `RADV_DEBUG=hang` dump put the last reached
command-processor trace point immediately before a `DRAW_INDEX_2`; `0x5008efd00` is an indexed
triangle-list draw.

## The program

From `PROSPER_DRAW_PROGRAM_CENSUS=1`, first sighted at the `worldmap` level load (25th of 32
distinct program triples live at the moment of the loss):

```
[draw-program] NEW  vs=0x5008efd00 chain=0x0 ps=0x5008f1400 draws=1 distinct=26
                    vcount=2776 nidx=4164 topo=3 tgt=0x520440000
```

Two draw shapes recur: `vcount=2776 nidx=4164` and `vcount=5736 nidx=8604`. Indexed triangle list,
into the 3840x2160 main colour target. No NGG chain continuation (`chain=0x0`).

## Evidence

All runs: `tools/screenshot`, native 3840x2160, no pad script (the world map is reached with no
input), `PROSPER_GUEST_ARGS=-force-gfx-direct PROSPER_RENDER=1`, current master plus the selector.
"Losses" counts `[backend] GRAPHICS submission failed` lines.

| run | `PROSPER_SKIP_DRAW_PROGRAM` | wall | losses | distinct frames | max pixel stale |
| --- | --- | --- | --- | --- | --- |
| control | *(unset)* | 300 s | 37,377 | 381 / 9,777 | 210.0 s |
| control | *(unset)* | 150 s | 56 | 382 / 4,385 | 60.0 s |
| A | the 8 programs new at `worldmap` | 240 s | **0** | 900 / 5,237 | 0.0 s |
[...]

PR #3194, opened 2026-09-01
feat(gpu): PROSPER_SKIP_DRAW_PROGRAM — decline graphics draws by shader program identity

mattias800:
## What this adds

`PROSPER_COMPUTE_SKIP_PROGRAM` can name a **compute** program, and that is what isolated GTA V's
hanging dispatch. Graphics had no equivalent. `PROSPER_SKIP_DRAW` selects a submit-local semantic
`draw_index` — an ordinal within one submit, not stable frame to frame — so it cannot express
"every draw that runs this shader", which is what a hanging *program* is.

This adds the counterpart:

- **`PROSPER_SKIP_DRAW_PROGRAM=0xADDR[,0xADDR...]`** — decline every draw whose vertex/ES program,
  NGG main continuation, or pixel program is at a named guest address, and report which stage
  matched.
- **`PROSPER_DRAW_PROGRAM_CENSUS=1`** — one line per distinct `(vs, vs-chain, ps)` triple, then
  powers of two. This is where the selector's input comes from; it is bounded by the *program*
  count rather than the draw count, so it is usable on a 4K title where `PROSPER_RTTLOG` is a
  firehose.

New module `prosper/src/gpu/diagnostics/draw_program_skip.{hpp,cpp}`, consulted from
`build_bds()` in the live renderer.

## Why now

Astro Bot's world-map GPU reset. The `VK_ERROR_DEVICE_LOST` line names a compute program, but that
program is the first submit to fail *after* the context was already lost; the `RADV_DEBUG=hang` dump
puts the last reached command-processor trace point immediately before a `DRAW_INDEX_2` (#1809,
2026-08-31). So the question "which draw?" had to be askable directly, and was not.

## Contract

Mirrored from the compute selector deliberately:

- **Default OFF.** Unset, the hot path is one `empty()` test per draw and nothing about the run
  changes.
- **Impossible to arm by accident.** Parsed by the strict `0x`-only list parser in
  `watch_list.hpp`, so a bare decimal, a stray comma, trailing junk, an overflow or a zero address
  arms **nothing** and says so. A partially-armed selector would produce a confident negative about
  a program nobody selected.
- **It reports itself.** An arming line, then
  `[draw-decline] program=… stage=vs|vs-chain|ps reason=skipped-by-selector count=…` per skip —
  first eight per `(program, stage)`, then powers of two, with the 1-based ordinal on every line
  (`diag_ratelimit.hpp`'s contract, so the cap can never be read as the rate). A log from an armed
  run cannot later be mistaken for a default run.

### Ordering — instrument trap 166

The decline is **last** in the per-draw build: after `build_R`, after descriptor-contract
validation, and after the `[render]` / `[rtt]` / `[draw-program]` per-draw census lines. A
diagnostic skip ordered before the dump blinds every other instrument to the thing under suspicion,
and the program worth declining is precisely the one that hangs the GPU. A declined draw is
therefore still fully realized, validated and logged; only the Vulkan draw call is withheld, and its
CPU cost is still counted in the timing partition's `build_draws` (it did spend that work).

### Four limits a reader of a skipped run cannot see in the output

Stated in `draw_program_skip.hpp`, `tools/AGENTS.md` and `src/gpu/diagnostics/AGENTS.md`, because
none of them show up in the log or the picture:

1. **Later passes sample the hole.** The renderer caches a pass's pixels under `CB_COLOR0_BASE` and
   injects them when a later draw samples that address. Decline the draw that fills a target and the
   composite that reads it does not fail — it succeeds against stale or empty pixels. The result is
   not "the frame minus that draw".
2. **A pass whose every draw is declined renders nothing, not even its clear.**
   `render_draw_pass_rgba` returns an empty image for an empty draw list, and a group's clear colour
   comes from its first item whether or not that item was declined. The present layer then serves the
   previously retained frame, which on screen looks like the frame rendered.
3. **Only draws that reach the live renderer can be declined** — not one the recompiler rejected or
   whose descriptor contract failed. Naming such a program produces no line; silence is not proof
   the selector matched.
4. **An address names a PROGRAM, not a draw.** Every draw using it goes, routinely thousands a
   frame.

## Result on Astro Bot (PPSA21564)

Full table, controls and caveats in **#3193**. Summary: bisection over the 32 distinct program
triples live at the moment of the loss took four runs and lands on **one** draw —
vertex/ES program `0x5008efd00`, pixel program `0x5008f1400`, an indexed triangle list first sighted
at the `worldmap` level load.

| run | `PROSPER_SKIP_DRAW_PROGRAM` | wall | `GRAPHICS submission failed` | distinct frames | max pixel stale |
| control | *(unset)* | 300 s | 37,377 | 381 / 9,777 | 210.0 s |
| control | *(unset)* | 150 s | 56 | 382 / 4,385 | 60.0 s |
| **skip** | **`0x5008efd00`** | 150 s | **0** | 607 / 3,297 | 0.0 s |
| **skip** | **`0x5008efd00`** (replication) | 300 s | **0** | 1,233 / 6,666 | 0.0 s |
| complement | `0x5008be700,0x5008cd600` | 150 s | 106 | 1 / 5,034 | 120.0 s |

Both controls fail at the identical `submit=6151 dispatch=16 order=1740980`, so the default failure
is deterministic and 150 s is well past it. The rung does not move: declining a draw is an
instrument, not a fix, and the world map still shows only its nebula backdrop (#1459).

## Tests

- **`draw_program_skip`** (new, no Vulkan) — 29 arms over the selector and the census: the disarmed
  fast path, seven malformed specs that must arm *nothing*, re-configure disarming rather than
  keeping the old list, each of the three stages, non-matching and all-zero draws, the rate limit's
  dense ordinal and its per-`(program, stage)` budget, and the census's NEW/recurrence/distinct
  accounting.
- **`draw_program_skip_render`** (new, Vulkan) — the wiring, through the real live renderer: the
  named program's draw paints nothing, an unnamed draw with the *same shaders* renders (the positive
  control), and a mixed submit loses only the named draw. The declined arm runs **first**, before
  anything has rendered, because the present layer serves the last retained frame when a submit
  publishes nothing — run second it would show the previous frame's pixels and look like it
  rendered.
  Both variables come from the ctest `ENVIRONMENT` property rather than a `setenv` in the test: the
  persistent-target opt-out it needs is a cached read, and arming it in-process is exactly the
  vacuous-arm defect `cached_env_arming_logic` exists to catch (it fails the gate if you try).

### Verification, exact head

```
cmake -S prosper -B prosper/build-linux -DPROSPER_APP=ON -DGAME_DUMP=<DUMP_ROOT>/PPSA21564-app0
cmake --build prosper/build-linux -j8                      # exit 0
ctest --test-dir prosper/build-linux --no-tests=error -j4  # exit 0, 321 tests, 100% passed
```

4 skipped (`module_loads_eboot`, `plugin_autolink`, `boot_reaches_first_syscall`,
`real_shader_render`) — expected, the dump is configured at `PPSA21564` rather than `PPSA24651`
(#1573).

**Mutation-verified, both halves, each with a green build so the old binary cannot report a false
pass:**

- `if (decision.skip)` → `if (false && decision.skip)` in `live_renderer.cpp` (drop the withholding,
  keep the counting): `draw_program_skip_render` fails on exactly *"the declined draw paints
  nothing"*, 1/9. Restored → 9/9.
- `matches(fs_addr)` → `false && matches(fs_addr)` in `draw_program_skip.cpp`:
  `draw_program_skip` fails 3 arms. Restored → 29/29.

No snapshots run — this adds an env-gated diagnostic that is inert unset, and the GPU was needed for
the Astro runs above.

## Risk

The only default-path change is one `addresses_.empty()` test per draw inside `build_bds`, on a
reference hoisted once per submit. Everything else is behind an unset environment variable. The new
module is in `src/gpu/diagnostics/`, whose `AGENTS.md` is updated to say this is now the **second**
entry there that is not observation-only when armed.

Refs #1809, #3193

### DIFF
--- a/BLOG.md
+++ b/BLOG.md
@@ -21,6 +21,19 @@ from the tracker issues, and still gated, because it is a projection of state ra
 
 ## 2026-09-01
 
+### Astro Bot's world map stops killing the GPU, and the draw that does it has a name
+
+No picture: what the world map renders is still the nebula backdrop we already published. What
+changed is that the run no longer dies there.
+
+Loading the world map took the whole GPU down — a driver-level hard recovery, after which every
+frame was the same frozen picture for the rest of the run. The message blamed a compute program,
+but that program was just the first thing to fail *after* the device was already gone. It was a
+single **draw**, and there was no way to ask which one: prosper could decline a compute program by
+name but not a graphics one. It can now, and bisecting thirty-two candidates took four runs — with
+one draw declined, a five-minute run has zero device losses and keeps producing new frames the whole
+way. ([#3193](https://github.com/mattias800/prosper/issues/3193))
+
 ### The three getenv calls the profiler pointed at were the three we could not fix
 
 No picture — this one is pure CPU time. `getenv` costs 1.24% of Blue Prince's gameplay frame
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -1677,6 +1677,13 @@ if(TARGET prosper_core)
   target_link_libraries(test_watch_list PRIVATE prosper_core)
   add_test(NAME watch_list COMMAND test_watch_list)
 
+  # PROSPER_SKIP_DRAW_PROGRAM: decline graphics draws by shader PROGRAM identity (the graphics
+  # counterpart of PROSPER_COMPUTE_SKIP_PROGRAM). Pure selector logic -- no Vulkan; the wiring into
+  # the live renderer is pinned separately by draw_program_skip_render below.
+  add_executable(test_draw_program_skip tests/gpu/diagnostics/test_draw_program_skip.cpp)
+  target_link_libraries(test_draw_program_skip PRIVATE prosper_core)
+  add_test(NAME draw_program_skip COMMAND test_draw_program_skip)
+
   add_executable(test_rdna2_decode tests/gpu/recompiler/test_rdna2_decode.cpp)
   target_link_libraries(test_rdna2_decode PRIVATE prosper_core)
   add_test(NAME rdna2_decode_walk COMMAND test_rdna2_decode)
@@ -2451,6 +2458,22 @@ if(TARGET prosper_core)
       target_include_directories(test_gpu_capture_render PRIVATE src tests tools/boot_trace)
       target_link_libraries(test_gpu_capture_render PRIVATE prosper_core prosper_live_renderer Vulkan::Vulkan)
       add_test(NAME gpu_capture_render_replay COMMAND test_gpu_capture_render)
+
+      # PROSPER_SKIP_DRAW_PROGRAM through the REAL live renderer: an unnamed program renders, the
+      # named one is withheld, and a mixed submit loses only the named draw. Needs the live renderer
+      # (and therefore Vulkan) because the wiring, not the selector logic, is what is under test.
+      add_executable(test_draw_program_skip_render tests/gpu/diagnostics/test_draw_program_skip_render.cpp)
+      target_include_directories(test_draw_program_skip_render PRIVATE frontends)
+      target_include_directories(test_draw_program_skip_render PRIVATE src tests tools/boot_trace)
+      target_link_libraries(test_draw_program_skip_render PRIVATE prosper_core prosper_live_renderer Vulkan::Vulkan)
+      add_test(NAME draw_program_skip_render COMMAND test_draw_program_skip_render)
+      # Armed from the ENVIRONMENT rather than by a setenv inside the test. The persistent-target
+      # opt-out is read through PROSPER_ENV_VALUE, which caches, so arming it from inside the process
+      # is exactly the vacuous-arm defect cached_env_arming_logic exists to catch; and the selector
+      # itself is read once into a function-local static, so environment arming removes the ordering
+      # argument for it too.
+      set_tests_properties(draw_program_skip_render PROPERTIES ENVIRONMENT
+        "PROSPER_SKIP_DRAW_PROGRAM=0x5006c6a00;PROSPER_NO_LIVE_PERSISTENT_COLOR_TARGETS=1")
 
       add_executable(test_vulkan_offscreen tests/test_vulkan_offscreen.cpp)
       target_link_libraries(test_vulkan_offscreen PRIVATE Vulkan::Vulkan)
--- a/prosper/frontends/shared/live/live_renderer.cpp
+++ b/prosper/frontends/shared/live/live_renderer.cpp
@@ -4,6 +4,7 @@
 #include "diagnostics/env_cache.hpp"   // PROSPER_ENV_ON / _VALUE: cached reads on per-draw paths
 #include "gpu/resources/metadata_kind_correlation.hpp"  // positive metadata-kind correlation (pure, tested)
 #include "gpu/diagnostics/watch_list.hpp"                 // strict 0x-only watch parsing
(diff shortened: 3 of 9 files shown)


mattias800 (commented):
**APPROVED**

**Mutation verified independently.** Removing the `continue` at `live_renderer.cpp:7417` — so a
declined draw is still submitted — makes `draw_program_skip_render` **fail**; restored, it passes.
Mutant build `rc=0` with **0 errors**, so this is not a stale-binary false pass. `ctest
--no-tests=error` **321/321, exit 0** on a clean tree.

**Building the lever was the right call, and the gap was real.** `PROSPER_COMPUTE_SKIP_PROGRAM` has
existed for the compute path since it cracked the GTA V hang; the graphics path had only
`PROSPER_SKIP_DRAW`, which keys on a per-submit `draw_index` that is not stable across frames and so
cannot isolate a program that hangs on one specific shader. "Which draw is responsible?" was
therefore not askable without rebuilding. Now it is, in four runs of bisection over 32 live program
triples.

**`PROSPER_DRAW_PROGRAM_CENSUS` is what makes the skip usable rather than a toy.** A lever whose
argument you cannot obtain is not a lever. Bounding the census by *program count* rather than draw
count is the detail that makes it work at 4K where `PROSPER_RTTLOG` is a firehose.

**Trap 166 is handled, not merely cited.** The decline is ordered *last* in the per-draw build —
after `build_R`, after descriptor validation, after the `[render]`/`[rtt]`/`[draw-program]` lines —
so a declined draw is fully realized and fully observable and only the Vulkan draw call is withheld.
A skip ordered before the dump would have blinded every other instrument to the thing under
suspicion, which is exactly what that trap records.

**The four graphics-specific limits are the part I would have expected to be missing**, and they are
in the header, `tools/AGENTS.md` and `src/gpu/diagnostics/AGENTS.md`: later passes sample the hole
via the RTT cache; a pass whose every draw is declined renders nothing *including its clear*, and the
present layer then serves the retained frame **so it looks like it rendered**; silence is not
absence of a match; an address names a program, not a draw. That second one is a genuine trap — a
skipped run that appears to render correctly would be read as exoneration.

**Strict `0x`-only parsing so a bare decimal arms nothing and says so**, plus a self-reporting arming
line, means a run made with this set can never later be mistaken for a default run. Same contract as
the compute switch.

## On the result

The evidence for `0x5008efd00` is strong and stated with its limits, which matters more than the
finding: both controls die at the **identical** `submit=6151 dispatch=16 order=1740980`, so the
default failure is deterministic rather than a race; with the draw declined, 0 device losses across
150 s and 300 s, and frames keep changing where the control's CRC freezes at 90 s.

The caveats are recorded honestly and I am accepting them as written rather than letting the headline
carry more than it should: **declining is not fixing**, the rung is unchanged and the world map still
shows only its nebula backdrop (#1459), the complement arm is **weak** (it lost the device at a
different `order`, so "only one hanging draw" is *not* established), and *why* it hangs is untouched.
That last one is blocked on tooling — `PROSPER_SHADER_DUMP_SUCCESS` dumps by content hash with no
address filter, so "the shader at `0x5008efd00`" cannot be pulled from a run today. That filter is the
obvious next tool and is worth an issue if #3193 does not already carry it.

Also correct: relying on the earlier `DRAW_INDEX_2` / zero-byte `vm_fault.log` discrimination and
saying whose it was rather than presenting it as this lane's own work.

`BLOG.md` entry present. No `Claude-Session:` trailer, no absolute host paths. CI 9 pass / 2 skipping.

Issue #1272, opened 2026-07-24
gpu: texture uploads carry a single mip level — dense textures alias hard under minification (most visible: Blue Prince flower beds)


mattias800:
In Blue Prince's 'Day One' opening scene (reachable on current master, see #1216), the flower beds and small planter foliage render as saturated RGB pixel noise while large surfaces (walls, trees, sky, path) render coherently. Looks like a texture decode/format defect specific to one texture class (small/compressed format — BC variant or tiled decode mismatch), not a shader failure: geometry and layout are correct and the noise is stable per texel region.

Repro + evidence: same route/frames as the companion lighting issue — `~/bp-1264/probe/PPSA25009_20260724-015746_150.png` (beds by the front steps and at the plaza corners). Suggested start: `PROSPER_RESOURCE_HASH_DIM` on a bed region to identify the sampled texture's format/address, then check the decode path for that format (BC6H/BC7/SNORM gaps are tracked in #294 for UE4 — this may be the Unity-side sibling).

mattias800:
Rescope after zoom analysis (128x64 crop of the flower bed at 4x): this is **not decode corruption** — the crop shows recognizable flower shapes (red clusters, a yellow-orange bloom lower-left, purple/violet blooms, green accents, pale gaps), i.e. the texture decodes and maps correctly. The 'pixel noise' impression at viewing distance is **minification aliasing**: prosper's texture path uploads a SINGLE mip level (the sampler site notes 'LOD min/max/bias honored; harmless with our single uploaded mip'), so a dense high-frequency flower texture viewed at a distance point-samples wildly instead of averaging through a mip chain. Every title pays this on minified textures; BP's flower beds are just the most visible case so far.

Real fix: upload the T#'s full mip chain (or generate mips at upload) and let minLod/maxLod work — an area:infra texture-path enhancement, not a BP decode bug. Retitling accordingly; the residual white bloom over the beds is the #1271/#1275 lighting layer, tracked there.

mattias800:
CLAIMING: claude-session d2181233 (overnight) | branch fix/issue-1272-mip-chain | 2026-07-24T~03:45Z. Design: the T# already decodes base_level/last_level (agc_shader_layout.hpp:131-132) — generate box-filtered mips bounded by the game's OWN declared chain (single-mip T#s unchanged), real tiled-mip decode as follow-up.

PR #3197, opened 2026-09-01
gpu: compute images carry the T#-declared mip chain, so image_load_mip is implemented (#3048)

### LINKED ISSUE #3048: Compute images are created with mipLevels=1, so image_load_mip cannot be implemented in the recompiler alone (Sonic Frontiers world)

mattias800:
Fixes #3048. Refs #2790, #2818.

## The gap

Compute-path images were created with `ici.mipLevels = 1` unconditionally. That single line made
`IMAGE_LOAD_MIP` unimplementable rather than merely unimplemented: the obvious recompiler fix —
lower it to `OpImageFetch` with a `Lod` operand — would fetch a level that does not exist. So the
recompiler could only admit the op when it could be *specialised away* to an ordinary level-zero
load, and declined every case where the guest's mip selector is a real runtime value.

Sonic Frontiers' three scene-target-width Cyber Space kernels are exactly that case: a **12-level
2048x2048 RG32F** surface (`type=13`/2D_ARRAY with `depth=1`, `tile_mode=27`, `BASE_LEVEL=0`,
`LAST_LEVEL=MAX_MIP=11`) read at a runtime LOD, with `proven_zero_mip=0`. No specialisation can
ever admit them.

The charter's rule applies directly: an unsupported GPU operation the guest exercises is a fatal
gap, not an acceptable skip.

## What changed

Both halves land together because neither works alone.

**1. One derivation, two consumers.** New `src/gpu/resources/mip_chain_plan.{hpp,cpp}` answers
"how many levels does this resource's compute image carry, and where does each level's guest data
live". `live_compute`'s image creation and the recompiler's MIMG lowering both call
`shader_resource_compute_mip_chain_levels`. If those two ever answered differently, a shader would
fetch a level nobody created — this is the shape the #2265 drift lesson asks for.

**2. The placement provenance the plan needs.** `image_base_level_view` shifts
`gpu_addr`/`width`/`height` onto the *selected* level, so nothing downstream could recover the
others: `tiled_mip_level_layout` needs the ALLOCATION's level-zero element extent and its effective
`MAX_MIP`, and neither survives that shift. `DecodedImageView` and `ShaderResource` now carry them
(`mip_chain_element_width/height`, `mip_chain_bytes_per_block`, `mip_chain_max_level`,
`mip_chain_base_level`), copied at all three `(descriptor, view) -> ShaderResource` build sites
through one shared helper.

**3. Real guest bytes, not a generated cascade.** Every declared level is detiled from the guest's
own `tiled_mip_level_layout` offset — packed-tail levels included, via `detile_surface_level` —
straight into the staging buffer, and copied with one `VkBufferImageCopy` per level. The graphics
path's linear-blit cascade (#1272) is deliberately **not** reused: `IMAGE_LOAD_MIP` asks for a
specific level's texels, and synthesizing them by downsampling level zero would render plausible
wrong content. `SONIC_FRONTIERS_STATUS.md` already argued this; the guest builds this pyramid itself
through twelve single-level descriptors, so the bytes are there to read.

**4. The recompiler emits the LOD.** New `rdna2_mimg_dynamic_mip_shape` identifies the general
`IMAGE_LOAD_MIP` address form — any DMASK, **UNRM and GLC free** (Sonic's packet has both clear,
which `rdna2_mimg_zero_mip_shape` rejects), no operand-shape modifier. The guest's own selector
reaches `OpImageFetch`'s `Lod`, clamped to the materialized level count the way hardware clamps to
`LAST_LEVEL`. Compute only: the graphics texture path still uploads one level.

## Reject-by-default, and where it fails visibly

The plan is invalid — leaving the historical single-level image and the historical decline — for a
linear chain, a selected level packed in the tail, a layered or volume view, DCC, block-compressed
or converting formats, a shifted `BASE_LEVEL`, and `host_data` backing (the allocation base sits
*below* `gpu_addr`, which host-backed resources cannot expose).

Three places deliberately fail visibly instead of degrading silently:

* A binding whose **shape** cannot carry the chain — imported/renderer-owned surfaces above all —
  is declined (`skip_image`), because building fewer levels than the compiled module may address is
  SPIR-V undefined behaviour. **The blast radius is wider than the first draft of this body said,
  and the failure was not as loud as it claimed.** The decline is *not* conditional on the module
  actually issuing `IMAGE_LOAD_MIP` — a sampling-only use behaves identically at one level or N, so
  declining it is pure loss, and the backend cannot see which ops a compiled module contains.
  Separately, `skip_image` warns once per `gpu_addr` for the whole process while the dispatch is
  dropped **every frame**, so the one line it prints reads as a one-off. Both are now written down
  at the site and in the status doc, and the decline reports as `[compute-mip-chain] declined …`
  on a geometric schedule (occurrences 1, 2, 4, 8, …) so a lane reading a black frame can see it
  recurring.
* A chain that reaches the end of the sampled conversion cascade with levels 1..N-1 **unwritten**
  declines. Two predicates decide "is this upload a straight per-texel copy" — the whitelist that
  chose the level count and the branch that does the copying — and a one-line widening of either
  would otherwise copy uninitialized staging into the image. This is a detector rather than a third
  copy of the predicate.
* The **persistent sampled cache is disabled for multi-level images**, because its revalidation
  covers only the selected level's guest bytes: an unchanged level zero would wrongly certify a
  stale level three. Costs upload time, never correctness.

Sampling behaviour for existing multi-level textures is unchanged: the sampler's `maxLod` stays 0
and `image_load`/`image_sample*` still resolve level zero, so a texture that merely *gains* levels
produces byte-identical results.

## Capture format

Serialized as a trailing tail at **v57**: the placement is what nothing else preserves once
`image_base_level_view` has shifted `gpu_addr`/`width`/`height` onto the selected level.

**What it does not do, corrected from the first draft of this body:** `gpu_replay` still declines
`IMAGE_LOAD_MIP`. A replayed resource is `host_data`-backed and its blob covers only the selected
level's footprint, and the chain derivation declines any `host_data` resource — so replay behaves
exactly as it did before v57. Making replay materialize the chain needs the capture to own the whole
**allocation's** bytes; filed as **#3202** with the three concrete steps. This tail is the
descriptor half, recorded now so the placement is not lost from capsules taken in the meantime.
Pre-v57 files read as "not modelled" and fail closed to one level.

## Verification

Build: fresh configure, `cmake -S prosper -B prosper/build-linux -G Ninja -DCMAKE_BUILD_TYPE=Debug`
inside the `ps5ys` distrobox.

```
cmake --build prosper/build-linux -j6                 -> exit 0, 0 errors
ctest --test-dir prosper/build-linux --no-tests=error -> exit 0
   100% tests passed out of 323
```

Rebased onto `origin/master` (`bee26b9e`) — the previous head had **zero CI runs** because it was
`CONFLICTING`, which is instrument trap 242 and looks exactly like Actions being broken. One content
conflict, in `live_compute.cpp` against #3160's alias census: both sides add an independent statement
at the same point, so both are kept (master's census block first, so its comment stays attached to
what it describes). The branch deletes 40 lines against master and every one of them is this PR's
own — checked line by line, per trap 41. The 323 count is master's 318 plus the five tests that
landed while this was open.

Registered coverage:

* `build_shader_resources` — four arms pinning `unmapped_format_image_view`'s fail-closed contract
  (the blocking finding below), including that it reports its mip chain as *not modelled*.
* `shader_resource_contract` — the plan's geometry (extents halve, every level inside the
  allocation, non-tail levels pairwise disjoint, level zero is the level `gpu_addr` names, the
  fixture really reaches the packed tail) plus **fifteen reject-by-default arms**.
* `rdna2_decode_walk` — the dynamic-mip address shape: Sonic's own `f0040308` (dmask `0x3`,
  UNRM/GLC clear, which `rdna2_mimg_zero_mip_shape` rejects) naming v3, the 2D_ARRAY form naming v4
  after its slice, and rejection of D16/A16/TFE/LWE/NSA/plain-load/dim:3D.
* `gpu_capture_roundtrip` — five **distinct** v57 fields round-tripped individually, a v56 downgrade
  reading as "not modelled", and a partially present placement rejected.
* `game_compute_exec` — **two** executed dispatches through the production compute backend against
  real tiled nine-level guest chains: `dmask:0x1` R32_UINT reading mip 1, and `dmask:0x3` on a
  two-channel 8 B/texel surface storing the **second** returned component — the mask Sonic actually
  issues, which the first arm could not observe. Both assert the guest's own level-one texels, plus
  that every destination texel was written.

**Mutation verification.** Two mutations, each with the build's exit code checked before the result
was believed:

| arm | build | ctest | result |
| --- | --- | --- | --- |
| baseline | 0 | 0 | — |
| `unmapped_format_image_view` reverted to the positional aggregate | 0 | 8 | **4 arms redden**, including both fail-closed guards and the "not modelled" one (the `supported` expression lands on `chain_element_width`) |
| restored | 0 | 0 | passes |
| baseline | 0 | 0 | 426 assertions in `game_compute_exec` |
| `ici.mipLevels = bi.mip_levels` reverted to `= 1` | 0 | 8 | **2 of 426 redden**, and they are the two that matter: `IMAGE_LOAD_MIP at a runtime mip returns LEVEL ONE's own guest texels` and `dmask:0x3 returns level one's SECOND component in the second VGPR` |
| restored | 0 | 0 | passes |

**On the anti-coincidence guard — corrected twice, and the second correction matters.** The first
version of this PR credited the "distinguishable from level zero" arm as the guard; it is not, since
it stays green under the mutant. The replacement explanation — "because an unwritten destination also
differs from level zero" — was **also** wrong, and measurement says so: under the mutant **both**
"every destination texel was actually written" arms stay green too. The dispatch does write every
texel, with something that is neither level one nor the sentinel.

So the two companion arms are **fixture-validity guards**, not discriminators: they rule out a
fixture that could pass vacuously (level one byte-identical to level zero; the dispatch never
running). The discriminator is the level-one equality itself, and behind it the fact that this
program only recompiles through the dynamic-LOD path. The comment now says exactly that and tells
the next reader not to draw mutant conclusions from either arm.

`PIPESTATUS[0]` was used throughout so no pipeline masked a failing stage. No snapshots were run and
no title was launched (another lane holds the GPU).

## Review response (#3197)

**The defect class is now eliminated, not just the instance.** `DecodedImageView() = default;` makes
the struct a non-aggregate under C++20 (P1008R1), so *any* positional braced initialization of it is
a compile error, while `V v;`, `V v{}` and `return {};` keep working. This matters because the arms
below pin the **function** that held the tree's one positional initializer — they cannot catch the
same shape at a **new** call site, which is how the original defect arrived. Verified in a standalone
TU on gcc 16 `-Wall -Wextra` before applying it (positional → `no matching function for call to
'V::V(<brace-enclosed initializer list>)'`; the other three forms compile), and a
`static_assert(!std::is_aggregate_v<DecodedImageView>)` stops the line being tidied away as
redundant.

**Blocking finding — fixed.** `gpu_executor.cpp` built the unmapped-format fallback view with a
**positional** `DecodedImageView` aggregate of eleven values, the only one in the tree. Inserting
five fields before `supported` re-bound the eleventh onto `chain_element_width`, so `supported` fell
back to its default `true` and the `BASE_ARRAY`/`BASE_LEVEL` guard went dead — silently, because
`bool -> uint32_t` is a promotion, not a narrowing. On the storage-image path that also writes slice
N's results into slice zero's guest bytes. Replaced with a named `unmapped_format_image_view` that
assigns every field by name, carries a comment saying why it may not go back, and is directly
testable. I handled the same hazard correctly for `ComputeImageCacheKey` and missed it here; the
tree now has no positional `DecodedImageView` left.

The other five, in the reviewer's numbering:

1. **v57 does not deliver replay** — confirmed against the source before rewriting anything
   (`bind_range` sets `r.host_data` for every replayed resource; `resource_footprint_impl` starts
   from `r.size`, the selected level). Claim corrected in the format comment, the status doc and
   this body; follow-up filed as **#3202** with the three concrete steps.
2. **v57 had no value coverage** — added, with distinct values so a transposition reddens, plus a
   v56 downgrade arm and a partial-placement rejection arm. Given the blocking finding was itself a
   field-ordering defect, this was the right thing to insist on.
3. **Two straight-copy predicates** — given a **drift detector** rather than a third copy. A
   detector that cannot fail silently seemed better here than exporting `live_compute`'s whole
   format classification; say so if you would still rather see them collapsed.
4. **Blast radius** — accepted and written down, at the site and in the status doc, plus the
   geometric-schedule report. See the bullet above.
5. **Smaller ones** — the clamp now separates the certain half (out-of-range `Lod` is SPIR-V UB, so
   *some* clamp is mandatory) from the inferred half (saturating to the last level, from doc 70648's
   BASE_LEVEL/LAST_LEVEL fields, which do not spell out MIMG's out-of-range behaviour), marked
   `CONFIDENCE: MED`; the 2D_ARRAY arm of `rdna2_mimg_dynamic_mip_shape` is now exercised by a
   decoder test instead of being advertised and dead; `dmask:0x3` has an executed arm; and the `<=`
   arm is now an equality that a decline cannot satisfy.

## Not claimed

**No routed arm has been run on Sonic Frontiers since this landed.** Nothing here says the Cyber
Space world renders. The status doc's own `## Ruled out` row — "fixing a recompiler decline that
unblocks these programs will change the frame" — has failed four times, and this is not evidence
against it. The next two measurements, in order: whether the three scene-width programs leave the
census skip list on a **default** build, and only then the composite's non-black percentage and
bounding box.

### DIFF
--- a/prosper/docs/SONIC_FRONTIERS_STATUS.md
+++ b/prosper/docs/SONIC_FRONTIERS_STATUS.md
@@ -586,8 +586,12 @@ changes nothing.
   `[compute-mip-chain] declined …` on a geometric schedule (`skip_image` alone warns once per
   address for the whole process while the dispatch is dropped every frame, which reads as a
   one-off when it is not).
-* **The LOD is clamped** to the materialized level count, which is what hardware does with a T#'s
-  `LAST_LEVEL`.
+* **The LOD is clamped** to the materialized level count. Half of that is certain and half is
+  inference, and the code says so at the site (`CONFIDENCE: MED`): an out-of-range `Lod` operand is
+  UNDEFINED in SPIR-V, so *some* clamp is mandatory whatever the hardware does — but *saturating to
+  the last level* is read from doc 70648's BASE_LEVEL/LAST_LEVEL fields, whose MIMG section does not
+  spell out the out-of-range behaviour, and no capture here exercises one. Returning zeros is the
+  other candidate.
 
 **Not yet measured on this title.** The change is covered by a registered execution test — a real
 tiled nine-level guest chain, dispatched through the production compute backend, asserting the
--- a/prosper/docs/SONIC_FRONTIERS_STATUS.md
+++ b/prosper/docs/SONIC_FRONTIERS_STATUS.md
@@ -570,11 +570,21 @@ changes nothing.
   `BASE_LEVEL` (`mip_chain_*`), because `gpu_addr`/`width`/`height` have already been shifted onto
   the selected level and nothing downstream could recover the rest. Captures serialize them at
   format v57; a pre-v57 file reads as "not modelled" and fails closed to one level.
+  **`gpu_replay` still declines this op**, and v57 does not change that: a replayed resource is
+  `host_data`-backed and its blob covers only the selected level, so the derivation fails closed
+  there. Materializing the chain on replay needs the capture to own the whole allocation's bytes —
+  a separate change. Do not take a capsule expecting to study the fetch offline yet.
 * **Reject-by-default.** Linear chains, a selected level packed in the tail, layered or volume
   views, DCC, block-compressed and converting formats, a shifted `BASE_LEVEL`, and `host_data`
   backing all keep the historical single-level image. A binding whose shape cannot carry the chain
   (imported/renderer-owned surfaces above all) is declined **fail-visibly** rather than silently
-  built with fewer levels than its compiled module addresses.
+  built with fewer levels than its compiled module addresses. **That decline is wider than it
+  needs to be** — it does not depend on the module actually issuing `IMAGE_LOAD_MIP`, because the
+  backend cannot see which ops a compiled module contains, so a sampling-only use of such a
+  resource on an RTT-aliased binding would be dropped for nothing. It reports as
+  `[compute-mip-chain] declined …` on a geometric schedule (`skip_image` alone warns once per
+  address for the whole process while the dispatch is dropped every frame, which reads as a
+  one-off when it is not).
 * **The LOD is clamped** to the materialized level count, which is what hardware does with a T#'s
   `LAST_LEVEL`.
 
--- a/prosper/docs/SONIC_FRONTIERS_STATUS.md
+++ b/prosper/docs/SONIC_FRONTIERS_STATUS.md
@@ -535,8 +535,9 @@ investigation to the wrong file:
   (`tests/fixtures/render_runner.h:6044-6055` — that file is the live offscreen Vulkan backend,
   included by `frontends/shared/live/live_renderer.cpp:39`). `tests/gpu/test_texture_mip_render.cpp`
   is a registered ctest asserting a declared 3-level chain samples level 2.
-* The **compute** path does not: its single `VkImageCreateInfo` for guest images sets
-  `ici.mipLevels = 1` unconditionally (`frontends/shared/live/live_compute.cpp:7089`).
+* The **compute** path did not: its single `VkImageCreateInfo` for guest images set
+  `ici.mipLevels = 1` unconditionally. **Fixed by #3048** — see the next section; the sentence is
+  kept because everything else in this bullet list is still current.
 * **And the graphics chain is GENERATED, not uploaded.** Staging carries level 0 only; levels
   1..N-1 come from a linear-filtered `vkCmdBlitImage` cascade at upload time
   (`render_runner.h:6257-6259`, `:7245`).
@@ -549,6 +550,43 @@ rule exists to prevent.
 
 **The census did not move when the MUST defect was fixed** — 14 programs, all `executed=0`, before
 and after at the same denominator — and neither did the composite. Both were measured, not assumed.
+
+### Compute images now carry the declared chain, and the LOD operand is emitted (#3048)
+
+Both halves of #3048 landed together, because neither works alone: emitting a `Lod` against a
+one-level image fetches a level that does not exist, and materializing levels nobody addresses
+changes nothing.
+
+* **One derivation, two consumers.** `src/gpu/resources/mip_chain_plan.{hpp,cpp}` answers "how many
+  levels does this resource's compute image have, and where does each level's guest data live".
+  `live_compute`'s image creation and the recompiler's MIMG lowering both call
+  `shader_resource_compute_mip_chain_levels`, so they cannot disagree — the drift that #2265
+  recorded (three copies of one predicate) is what this shape exists to prevent.
+* **Real guest bytes, not a generated cascade.** Each level is detiled from the guest's own
+  `tiled_mip_level_layout` offset, tail levels included, straight into the staging buffer. The
+  graphics path's blit cascade is deliberately not reused here: this section's own analysis is why.
+* **The provenance the placement needs is now carried.** `ShaderResource` gained the allocation's
+  level-zero element extent, its bytes-per-block, its effective `MAX_MIP` and the view's
+  `BASE_LEVEL` (`mip_chain_*`), because `gpu_addr`/`width`/`height` have already been shifted onto
+  the selected level and nothing downstream could recover the rest. Captures serialize them at
+  format v57; a pre-v57 file reads as "not modelled" and fails closed to one level.
+* **Reject-by-default.** Linear chains, a selected level packed in the tail, layered or volume
+  views, DCC, block-compressed and converting formats, a shifted `BASE_LEVEL`, and `host_data`
+  backing all keep the historical single-level image. A binding whose shape cannot carry the chain
+  (imported/renderer-owned surfaces above all) is declined **fail-visibly** rather than silently
+  built with fewer levels than its compiled module addresses.
+* **The LOD is clamped** to the materialized level count, which is what hardware does with a T#'s
+  `LAST_LEVEL`.
+
+**Not yet measured on this title.** The change is covered by a registered execution test — a real
+tiled nine-level guest chain, dispatched through the production compute backend, asserting the
+result is *level one's own texels* — but no routed arm has been run since, so nothing here claims
+the Cyber Space world renders. Two things are worth measuring next, in this order: whether the three
+scene-width programs now leave the census skip list on a **default** build (they did on the
+measurement-only LOD-0 build, which is a weaker statement), and only then the composite's non-black
+percentage and bounding box. The `## Ruled out` row above — "fixing a recompiler decline that
+unblocks these programs will change the frame" — is **not established**, and has now failed four
+times; do not assume the fifth.
 
 ## Ruled out
 
--- a/prosper/docs/SONIC_FRONTIERS_STATUS.md
+++ b/prosper/docs/SONIC_FRONTIERS_STATUS.md
@@ -573,7 +573,8 @@ changes nothing.
   **`gpu_replay` still declines this op**, and v57 does not change that: a replayed resource is
   `host_data`-backed and its blob covers only the selected level, so the derivation fails closed
   there. Materializing the chain on replay needs the capture to own the whole allocation's bytes —
-  a separate change. Do not take a capsule expecting to study the fetch offline yet.
+  a separate change, tracked as **#3202**. Do not take a capsule expecting to study the fetch
+  offline yet.
 * **Reject-by-default.** Linear chains, a selected level packed in the tail, layered or volume
   views, DCC, block-compressed and converting formats, a shifted `BASE_LEVEL`, and `host_data`
   backing all keep the historical single-level image. A binding whose shape cannot carry the chain
--- a/prosper/frontends/shared/live/live_compute.cpp
+++ b/prosper/frontends/shared/live/live_compute.cpp
@@ -6826,6 +6826,31 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                     volume_texels == static_cast<uint64_t>(r->width) * r->height &&
                     texel_bytes == r->mip_chain_bytes_per_block;
                 if (!chain_binding_shape) {
+                    // BLAST RADIUS, stated plainly: this decline is NOT conditional on the module
+                    // actually issuing IMAGE_LOAD_MIP. A module that only samples this resource
+                    // behaves identically at one level or N, so declining it is pure loss -- but
+                    // the backend cannot see which ops the compiled module contains, and the
+                    // alternative (build one level while the module may address N) is SPIR-V
+                    // undefined behaviour. The reachable case is a guest texture that declares a
+                    // placeable chain AND aliases a live render target, which the imported/
+                    // renderer-owned terms above exclude.
+                    //
+                    // `skip_image` warns once per address for the whole process while this drops
+                    // the dispatch EVERY frame, so a lane reading a black frame would find one
+                    // line from an hour ago and nothing since. Count it and re-report on a
+                    // geometric schedule: loud early, bounded later.
+                    static std::atomic<uint64_t> chain_shape_declines{0};
+                    const uint64_t occurrence =
+                        chain_shape_declines.fetch_add(1, std::memory_order_relaxed) + 1u;
+                    if ((occurrence & (occurrence - 1u)) == 0u)
+                        std::fprintf(stderr,
+                                     "[compute-mip-chain] declined binding=%u addr=0x%llx "
+                                     "levels=%u imported=%u rtt=%u (occurrence %llu; this "
+                                     "dispatch is dropped every frame)\n",
+                                     bi.binding, (unsigned long long)r->gpu_addr,
+                                     declared_chain_levels, bi.imported ? 1u : 0u,
+                                     renderer_owned ? 1u : 0u,
+                                     (unsigned long long)occurrence);
                     skip_image(r, "declared mip chain not materializable for this binding shape");
                     break;
                 }
@@ -7877,6 +7902,9 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                         // could not see B's read at all, so that pair scored as no alias.
                         alias_seeds.push_back({r->gpu_addr, need});
                     }
+                    // Set only by the native straight-copy branch below, which is the only one that
+                    // can produce levels 1..N-1. See the detector after the conversion chain.
+                    bool mip_chain_levels_written = false;
                     const bool needs_sampled_upload = !bi.upload_skipped &&
                         !bi.compute_transfer_seed_borrowed;
                     if (needs_sampled_upload && sampled_dcc_fast_clear) {
@@ -8037,11 +8065,13 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                             sampled_uint32_native || sampled_float32_native ||
                             sampled_float16_native || sampled_depth) {  // Native sampled texels
                             std::memcpy(upload, sampled_source, linear_bytes);
-                            if (bi.mip_levels > 1u)
+                            if (bi.mip_levels > 1u) {
                                 upload_guest_mip_chain_levels(
                                     mip_chain, bi.mip_staging_offsets,
                                     src - mip_chain.levels[0].byte_offset, r->tile_mode,
                                     bpt, upload);
+                                mip_chain_levels_written = true;
+                            }
                         } else if (f32) {                           // Native float channels + default fill
                             parallel_compute_texels(texels, linear_bytes + texels * 16u,
                                 [&](size_t begin, size_t end) {
@@ -8071,6 +8101,21 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                         } else {                                    // Float16: half -> UNORM8 + default fill
                             sampled_float16_to_unorm8_range(
                                 sampled_source, nc, texels, upload);
+                        }
+                        // Two predicates decide "is this upload a straight per-texel copy": the
+                        // whitelist inside `shader_resource_compute_mip_chain_levels`, which chose
+                        // the level COUNT, and the branch condition above, which does the copying.
+                        // They agree today, and a one-line widening of either would leave levels
+                        // 1..N-1 as whatever the staging allocation happened to contain -- copied
+                        // into the image and sampled as if it were guest data. So rather than a
+                        // third copy of the predicate, this is a detector: the branch that can
+                        // write those levels is the only one that sets the flag, and a chain that
+                        // reaches here unwritten declines instead of uploading garbage.
+                        if (bi.mip_levels > 1u && !mip_chain_levels_written) {
+                            skip_image(r,
+                                       "declared mip chain levels not written by the sampled "
+                                       "conversion that ran (straight-copy predicates disagree)");
+                            break;
                         }
                     }
                 }
--- a/prosper/frontends/shared/live/live_compute.cpp
+++ b/prosper/frontends/shared/live/live_compute.cpp
@@ -19,6 +19,7 @@
 #include "gpu/recompiler/gta5/rdna2_gta5_cf9200_contract.hpp"
 #include "gpu/resources/shader_resources.hpp"
 #include "gpu/recompiler/spirv_builder.hpp"
+#include "gpu/resources/mip_chain_plan.hpp"
 #include "gpu/texture/tile.hpp"
 #include "gpu/capture/writer_provenance.hpp"
 #include "host/memory/guest_write_watch.hpp"
@@ -881,6 +882,11 @@ struct ComputeImageCacheKey {
     bool in_mip_tail = false;
     bool srgb = false;
     bool depth_compare = false;
+    // #3048: a cached image is created with exactly this many mip levels. Two T#s over the same
+    // allocation can agree on every field above and declare different chain lengths, and handing a
+    // one-level image to a binding whose module fetches level three is not a miss but a fault.
+    // Appended LAST so `storage_image_cache_key`'s positional aggregate init keeps its meaning.
+    uint32_t mip_levels = 1;
 
     bool operator==(const ComputeImageCacheKey& other) const = default;
 };
@@ -899,13 +905,15 @@ struct ComputeImageCacheKeyHash {
         mix(key.mip_tail_offset); mix(key.mip_tail_bytes);
         mix(key.mip_tail_x); mix(key.mip_tail_y); mix(key.vk_format);
         mix(key.storage); mix(key.in_mip_tail); mix(key.srgb); mix(key.depth_compare);
+        mix(key.mip_levels);
         return result;
     }
 };
 
 ComputeImageCacheKey storage_image_cache_key(const prosper::gpu::ShaderResource& resource,
                                               uint32_t guest_bytes,
-                                              VkFormat native_format) {
+                                              VkFormat native_format,
+                                              uint32_t mip_levels = 1) {
     return {
         resource.gpu_addr, reinterpret_cast<uintptr_t>(resource.host_data),
         guest_bytes, resource.size,
@@ -917,7 +925,7 @@ ComputeImageCacheKey storage_image_cache_key(const prosper::gpu::ShaderResource&
         resource.mip_tail_offset, resource.mip_tail_bytes,
         resource.mip_tail_x, resource.mip_tail_y,
         static_cast<uint32_t>(native_format), true, resource.in_mip_tail,
-        resource.srgb, resource.depth_compare};
+        resource.srgb, resource.depth_compare, mip_levels};
 }
 
 struct CachedComputeImage {
@@ -2988,6 +2996,11 @@ struct BoundImage {
     }
     uint32_t texel_depth = 1;           // logical Z/layer count represented in the staging buffer
     uint32_t array_layers = 1;           // Vulkan array-layer count (3D depth remains one layer)
+    // #3048: the guest-declared mip chain this image materializes, and where each level past zero
+    // begins in the staging buffer. 1 (with an empty offset table) is the historical single-level
+    // image; the recompiler reads the same derivation before emitting an explicit LOD.
+    uint32_t mip_levels = 1;
+    std::vector<VkDeviceSize> mip_staging_offsets;
     bool arrayed_2d = false;            // SPIR-V requires a real 2D-array view (not base-slice fallback)
     bool stacked_cube = false;          // cube lowering addresses six faces as one w x 6h 2D image
     bool depth_view = false;             // reflected SPIR-V uses a true depth image/sampler contract
@@ -4951,6 +4964,31 @@ void parent_scan_dump_pair(uint64_t addr, const uint8_t* after, size_t count,
     }
 }
 
+// Detile levels 1..N-1 of a guest mip chain straight into the staging buffer (#3048). Level zero is
+// written by the ordinary single-level path and is deliberately not touched here, so a resource that
+// gains a chain keeps producing byte-identical level-zero texels.
+//
+// `allocation_base` is the FIRST byte of the guest allocation, which sits below the resource's own
+// `gpu_addr` whenever the selected level is not the allocation's first stored byte -- tiled GFX10
+// chains store the shared mip-tail block first and then the remaining levels smallest-to-largest.
+void upload_guest_mip_chain_levels(const prosper::gpu::MipChainPlan& plan,
+                                   const std::vector<VkDeviceSize>& staging_offsets,
+                                   const uint8_t* allocation_base, uint32_t tile_mode,
+                                   uint32_t texel_bytes, uint8_t* upload) {
+    for (uint32_t level = 1; level < plan.level_count; ++level) {
+        const prosper::gpu::MipChainLevel& source = plan.levels[level];
+        uint8_t* destination = upload + staging_offsets[level];
+        const uint8_t* bytes = allocation_base + source.byte_offset;
+        if (source.in_tail)
+            prosper::gpu::detile_surface_level(destination, bytes, source.byte_size,
+                                               source.width, source.height, tile_mode,
+                                               texel_bytes, source.tail_x, source.tail_y);
+        else
+            prosper::gpu::detile_surface(destination, bytes, source.width, source.height,
+                                         tile_mode, 0, texel_bytes);
+    }
+}
+
 bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& item) {
     using namespace prosper::gpu;
     using ComputeClock = std::chrono::steady_clock;
@@ -6562,6 +6600,14 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                     p->layer_mip_offset_bytes == r->layer_mip_offset_bytes &&
                     p->in_mip_tail == r->in_mip_tail &&
                     p->mip_tail_x == r->mip_tail_x && p->mip_tail_y == r->mip_tail_y &&
+                    // The chain length is derived from these six fields (#3048); two descriptors
+                    // that disagree on any of them materialize different images.
+                    p->declared_mip_levels == r->declared_mip_levels &&
+                    p->mip_chain_element_width == r->mip_chain_element_width &&
+                    p->mip_chain_element_height == r->mip_chain_element_height &&
+                    p->mip_chain_bytes_per_block == r->mip_chain_bytes_per_block &&
+                    p->mip_chain_max_level == r->mip_chain_max_level &&
+                    p->mip_chain_base_level == r->mip_chain_base_level &&
                     p->srgb == r->srgb && same_host_backing &&
                     p->host_data_size == r->host_data_size && same_dcc_identity;
                 bool same_sampler = true;
@@ -6585,6 +6631,8 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                 bi.sampler = owner.sampler; bi.guest_bytes = owner.guest_bytes;
                 bi.texel_depth = owner.texel_depth;
                 bi.array_layers = owner.array_layers;
+                bi.mip_levels = owner.mip_levels;
+                bi.mip_staging_offsets = owner.mip_staging_offsets;
                 bi.dcc_metadata = owner.dcc_metadata;
                 bi.dcc_metadata_bytes = owner.dcc_metadata_bytes;
                 staging_bytes[i] = staging_bytes[bi.alias_of];
@@ -6758,7 +6806,46 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                                   : VK_FORMAT_R8G8B8A8_UNORM;
             // Most storage images use raw uvec4 channels; reflected exact paths keep native-width
             // bytes. Most sampled formats normalize to RGBA8, while HDR/integer formats stay native.
-            const VkDeviceSize sbytes = volume_texels * texel_bytes;
+            // #3048: the T#-declared mip chain. `shader_resource_compute_mip_chain_levels` is the
+            // ONE derivation the recompiler also reads before it lowers IMAGE_LOAD_MIP with an
+            // explicit LOD, so this backend must materialize exactly what that function reports.
+            // A binding shape that cannot carry the chain is declined fail-visibly rather than
+            // quietly creating fewer levels than the compiled module addresses.
+            const uint32_t declared_chain_levels =
+                prosper::gpu::shader_resource_compute_mip_chain_levels(*r);
+            prosper::gpu::MipChainPlan mip_chain;
+            VkDeviceSize mip_chain_extra_bytes = 0;
+            if (declared_chain_levels > 1u) {
+                mip_chain = prosper::gpu::shader_resource_mip_chain_plan(*r);
+                const bool chain_binding_shape =
+                    mip_chain.valid && mip_chain.level_count == declared_chain_levels &&
+                    !bi.storage && !bi.imported && !renderer_owned &&
+                    !bi.depth_bits_source && !bi.unorm_rtt_value_reuse && !bi.depth_view &&
+                    !dim_1d && !dim_3d && !dim_cube_stacked && !cube_face_as_2d &&
+                    sampled_layers == 1u && bi.array_layers == 1u && !bi.stacked_cube &&
+                    volume_texels == static_cast<uint64_t>(r->width) * r->height &&
+                    texel_bytes == r->mip_chain_bytes_per_block;
+                if (!chain_binding_shape) {
+                    skip_image(r, "declared mip chain not materializable for this binding shape");
+                    break;
+                }
+                // Vulkan requires each region's bufferOffset to be a multiple of four AND of the
+                // texel block size; every width here is a power of two, so one alignment covers both.
+                const VkDeviceSize alignment = texel_bytes < 4u ? 4u : texel_bytes;
+                const VkDeviceSize level_zero_bytes = volume_texels * texel_bytes;
+                bi.mip_staging_offsets.assign(mip_chain.level_count, 0);
+                VkDeviceSize offset = (level_zero_bytes + alignment - 1u) / alignment * alignment;
+                for (uint32_t level = 1; level < mip_chain.level_count; ++level) {
+                    bi.mip_staging_offsets[level] = offset;
+                    const VkDeviceSize level_bytes =
+                        static_cast<VkDeviceSize>(mip_chain.levels[level].width) *
+                        mip_chain.levels[level].height * texel_bytes;
+                    offset = (offset + level_bytes + alignment - 1u) / alignment * alignment;
+                }
+                mip_chain_extra_bytes = offset - level_zero_bytes;
+                bi.mip_levels = mip_chain.level_count;
+            }
+            const VkDeviceSize sbytes = volume_texels * texel_bytes + mip_chain_extra_bytes;
             if (!sbytes || sbytes > kMaxComputeImageBytes) {
                 skip_image(r, "expanded image exceeds the 512 MiB backend bound"); break;
             }
@@ -6930,6 +7017,18 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                         (r->host_data && r->host_data_size >= sampled_guest_need) ||
                         guest_readable(r->gpu_addr, static_cast<uint32_t>(sampled_guest_need));
                     if (!readable) { skip_image(r, "sampled surface unreadable"); break; }
+                    // The chain's other levels live in the SAME allocation, below and above the
+                    // selected level. Bound the whole span before any of it is dereferenced.
+                    if (bi.mip_levels > 1u) {
+                        const uint64_t level_zero_offset = mip_chain.levels[0].byte_offset;
+                        if (r->host_data || level_zero_offset > r->gpu_addr ||
+                            mip_chain.allocation_bytes > UINT32_MAX ||
+                            !guest_readable(r->gpu_addr - level_zero_offset,
+                                            static_cast<uint32_t>(mip_chain.allocation_bytes))) {
+                            skip_image(r, "declared mip chain allocation unreadable");
+                            break;
+                        }
+                    }
                 }
 
                 static const bool dcc_cache_disabled =
@@ -7004,6 +7103,12 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                     .persistent_enabled = persistent_compute_image_enabled(
                         sbytes, ComputeImageCacheClass::sampled),
                 });
+                // #3048: a cached sampled image is revalidated against the SELECTED level's guest
+                // bytes only. A multi-level image's higher levels live elsewhere in the same
+                // allocation, so an unchanged level zero would wrongly certify a stale level three.
+                // Fail closed and re-upload every dispatch until the validated span covers the whole
+                // chain; this costs upload time, never correctness.
+                if (bi.mip_levels > 1u) bi.cache_candidate = false;
                 const VkFormat transfer_native_format =
                     compute_transfer_storage_vk_format(r->format, sampled_components);
                 const bool float32_uint32_transfer_alias =
@@ -7055,7 +7160,7 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                         r->mip_tail_offset, r->mip_tail_bytes,
                         r->mip_tail_x, r->mip_tail_y,
                         static_cast<uint32_t>(image_format), bi.storage, r->in_mip_tail,
-                        r->srgb, r->depth_compare};
+                        r->srgb, r->depth_compare, bi.mip_levels};
                     // An ordinary native typed 2D image or native typed 3D volume is byte- and
                     // format-identical to the sampled upload that follows it. Borrow that retained
                     // result only as a TRANSFER source: the sampled cache remains a separate image so
@@ -7932,6 +8037,11 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                             sampled_uint32_native || sampled_float32_native ||
                             sampled_float16_native || sampled_depth) {  // Native sampled texels
                             std::memcpy(upload, sampled_source, linear_bytes);
+                            if (bi.mip_levels > 1u)
+                                upload_guest_mip_chain_levels(
+                                    mip_chain, bi.mip_staging_offsets,
+                                    src - mip_chain.levels[0].byte_offset, r->tile_mode,
+                                    bpt, upload);
                         } else if (f32) {                           // Native float channels + default fill
                             parallel_compute_texels(texels, linear_bytes + texels * 16u,
                                 [&](size_t begin, size_t end) {
@@ -7991,7 +8101,7 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
             ici.format = image_format;
             ici.extent = {r->width, dim_cube_stacked ? r->height * 6u : r->height,
                           dim_3d ? bi.texel_depth : 1u};
-            ici.mipLevels = 1;
+            ici.mipLevels = bi.mip_levels;
             ici.arrayLayers = bi.array_layers;
             ici.samples = VK_SAMPLE_COUNT_1_BIT;
             ici.tiling = VK_IMAGE_TILING_OPTIMAL;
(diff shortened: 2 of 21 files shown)


mattias800 (commented):
## REJECTED

Reviewed at head `d880fd0e`. Own worktree, own fresh build dir, `ps5ys` distrobox
(Vulkan found, offscreen harness enabled). CI on this head is 9/9 green + 2 skipped;
`diff --check` clean; no `Claude-Session:` trailers.

I like this change a lot. The derivation-with-two-consumers shape is the right answer to
#2265, the reject-by-default list is honest, the "real guest bytes, not a generated
cascade" decision is correct and well argued, and the execution test's discriminator is
genuinely a discriminator (see below). Almost everything the PR body claims, I checked and
found true. One thing it claims is false, and one thing it does not claim is a silent
regression that this PR introduces. The second is why this is `REJECTED` rather than
`APPROVED` with nits — it should be a small fix.

---

# 1. BLOCKING — the PR walks into the braced-aggregate trap it says it avoided, and silently kills a fail-closed guard

The PR body says `ComputeImageCacheKey` gained its field **last** so the positional
aggregate inits keep their meaning. That is true, and I verified it (details in §7).

But `ComputeImageCacheKey` is not the only aggregate this PR grows. **`DecodedImageView`
gains five fields too, and they are inserted *before* `supported`** —
`prosper/src/gpu/agc/agc_shader_layout.hpp:233-240`:

```
 9. size_t   layer_stride
10. size_t   layer_mip_offset
11. uint32_t chain_element_width     <- NEW
12. uint32_t chain_element_height    <- NEW
13. uint32_t chain_bytes_per_block   <- NEW
14. uint32_t chain_max_mip           <- NEW
15. uint32_t chain_base_level        <- NEW
16. bool     supported = true
```

And there is exactly one positional init of that aggregate in the tree —
`prosper/src/gpu/execute/gpu_executor.cpp:7803-7805`:

```cpp
const DecodedImageView view = mapped_fmt
    ? image_base_level_view(d, fi)
    : DecodedImageView{d.base, d.width, d.height, 0, false, 0, 0, 0,
                       0, 0, d.base_level == 0 && d.base_array == 0};
```

Eleven initializers. Before this PR the eleventh member was `supported`. **After this PR
the eleventh member is `chain_element_width`**, so `d.base_level == 0 && d.base_array == 0`
now initializes a `uint32_t` (0 or 1) and `supported` falls back to its default member
initializer — **`true`, unconditionally.**

`bool -> uint32_t` is not a narrowing conversion, so this compiles under `-Wall -Wextra`
with **no diagnostic at all**. I confirmed that, and then A/B'd the exact initializer list
from `:7804` against the two headers, with `base_level = 1`:

| `DecodedImageView` from | `supported` |
| --- | --- |
| `origin/master` | **0** (guard fires) |
| this PR's head | **1** (guard dead) |

Same source line, same eleven initializers, opposite answers, no warning either way.

What that guard is for is written three lines above it, at `gpu_executor.cpp:7797-7801`:

> *"The unmapped-format fallback builds a view by hand and therefore never applies a slice
> offset. It must fail closed on a non-zero BASE_ARRAY for the same reason
> `image_base_level_view`'s early returns do: an unshifted base under a selected slice binds
> slice ZERO's texels silently."*

So on current master an unmapped-format **storage image** with `BASE_LEVEL != 0` or
`BASE_ARRAY != 0` is dropped loudly (`warn_unsupported_image_view` + `continue`). On this
branch it is accepted, `r.gpu_addr = view.base` is the *allocation* base and `r.width` the
allocation's level-0 extent — so the dispatch reads slice/level zero for a descriptor that
selected slice N, **and, if the shader writes it, the storage writeback puts slice N's
results into slice zero's guest bytes.** Silent wrong image plus guest-memory corruption, in the one path
whose comment exists to say that must never happen.

`chain_element_width` ends up 0 or 1 with the other four zero, so the mip-chain side itself
fails closed (`mip_chain_plan.cpp:60` rejects on `!element_height`). The damage is entirely
in `supported`.

**Fix**, one line, and the idiom is already in this diff's neighbourhood — `live_compute.cpp:7058-7062`
uses designated initializers with a comment that reads like it was written for this exact
occasion (*"Designated, not positional: inserting a field into this struct must not be able
to silently re-bind these arguments. #3149 lost a round to exactly that hazard in
ComputeImageCacheKey."*):

```cpp
: DecodedImageView{.base = d.base, .width = d.width, .height = d.height,
                   .supported = d.base_level == 0 && d.base_array == 0};
```

Please also add a regression arm asserting that hand-built view is `!supported` for
`base_level != 0` and for `base_array != 0`. Nothing in the suite or the nine green CI jobs on this head saw it.

---

# 2. The v57 format bump does not do what the PR body says it does

The body justifies the bump as:

> *"Without it a replayed capture could not place any level but the selected one, so the
> compute backend would build a single-level image where the live run built the chain — and
> `gpu_replay` would decline the very `IMAGE_LOAD_MIP` the capture was taken to study."*

**A replayed capture still declines it, for two independent reasons the placement metadata
cannot touch.**

1. `materialize_gpu_replay` backs every replayed resource with `host_data`
   (`gpu_capture.cpp:5139` `host_data = instance.data() + blob_offset;`, and `:5228`).
   `shader_resource_compute_mip_chain_levels` returns **1** for any `host_data`-backed
   resource — `mip_chain_plan.cpp:126-130`, with a correct comment saying exactly why. So on
   replay the level count is 1 whatever the v57 tail said, the image is single-level, and
   `dynamic_mip_load` is false.
2. Even if (1) were lifted, the bytes are not in the file. `resource_footprint_impl`
   (`gpu_capture.cpp:660`) sizes a texture from `r.width`/`r.height` — the **selected**
   level — so the blob starts at `gpu_addr` and covers one level. The chain's other levels
   live below `gpu_addr` (the tail-first block) and above the selected level's span, both
   outside the captured range.

So the tail is currently **write-only metadata**: it round-trips and is never read to any
effect. That is not harmful (it fails closed and old files still load), but a format bump
justified in the record by a capability it does not deliver is exactly the kind of claim the
next agent inherits and builds on. Please either drop the bump from this PR, or keep it and
say plainly in both the PR body and `SONIC_FRONTIERS_STATUS.md` that it is groundwork and
that replay still declines because of `host_data` + the single-level captured footprint.

# 3. The v57 tail has zero value coverage

```
$ grep -c mip_chain prosper/tests/gpu/capture/test_gpu_capture.cpp
0
```

Every touched capture assertion is `format_version == 56` -> `== 57`. **No test writes a
non-zero `mip_chain_*` into a capture, round-trips it, and asserts it comes back**, and the
new partial-record rejection branch (`gpu_capture.cpp:5046-5056`, the `any && (...)` check)
is never executed with a non-zero record either. A field transposition in the writer or
reader — say `max_level` where `base_level` belongs — is invisible to the whole suite,
because the two loops are symmetric and all the data is zeros.

This is below the bar the file sets for itself: v32's tail has
`array_layout_loaded...layer_stride_bytes == 720896u`, v43's has
`msaa_loaded...sample_count == 4`. One fixture with a real placement and five distinct
values would close it, and would also let the malformed arm be exercised.

On the "tail-located-by-subtraction" fixtures you flagged in the body: I checked all six
edits and the arithmetic is right — v57 is appended *after* v55 in the writer, and every
fixture that measures backwards from EOF now subtracts it in the correct order
(`v53_scalar_bytes` removes v57, then v55, then v54; `stale_tail` and `malformed_scalar_bytes`
likewise). Adding `v57_tail(f)` to `v50_tail` matches that lambda's stated contract
("include the later suffixes so the downgrade arithmetic stays exact"). So they were not
"fixed to accommodate the bug" — but note `v50_tail` still omits `v55_ds_slice_tail`
(pre-existing; harmless only because those fixtures happen to carry no `ds_seeds`), and the
idiom is now **eight** subtraction terms deep. A self-describing tail length, or offsets
computed forward, would retire the whole class.

# 4. There are TWO derivations of "is this a straight copy", not one

This is the PR's own thesis turned on the PR. `native_straight_copy_sampled_format`
(`mip_chain_plan.cpp:23-37`) decides which formats may carry a chain. The backend
independently decides which formats take the straight-`memcpy` upload
(`live_compute.cpp:7968-7972`, `rgba8 || uint8 || ... || sampled_depth`). **The per-level
upload is inside that second branch** (`:7974-7978`) — if a binding gets `mip_levels > 1` but
lands in any other upload branch, levels 1..N-1 are copied into the image from
**never-written staging memory**. No skip, no log, no validation error: garbage texels in the
levels the shader was compiled to fetch.

I enumerated all seven whitelisted (format, components) combinations and they all reach the
native branch today, and `texel_bytes == mip_chain_bytes_per_block` holds for every one — so
this is **not** currently reachable. But the two lists are only coincidentally aligned. One
plausible future line — adding `Float16` to the whitelist — makes `Float16x2` qualify
(`texel_bytes` falls to the `: 4u` default, `bytes_per_block` is also 4, so the shape gate
passes) while `sampled_float16_native` is false for a guest-backed 2D surface, so the upload
takes `sampled_float16_to_unorm8_range` and the higher levels are never written.

Cheap structural fix: hoist the backend's disjunction into one named `bool
native_sampled_texels`, require it in `chain_binding_shape`, and use the same bool for the
upload branch. Then the two cannot drift, which is the property the rest of this PR is built
around.

---

# 5. Your review question: is `skip_image` the right trade, and is it loud?

**The trade is right; your statement of its blast radius is narrower than the code.**

Right, because the alternative is worse in a way the compile key cannot rescue: the module
bakes in `materialized_mip_levels` (it is in `ShaderCompileKey`, `gpu_executor.cpp:572`, and
in its hash at `:811`), the `Lod` clamp is a literal, and an out-of-range `Lod` in SPIR-V is
undefined. Building fewer levels than the module addresses is a fault, not a miss. Declining
is correct.

Narrower, because **the decline does not depend on the module actually using
`IMAGE_LOAD_MIP`.** `declared_chain_levels` is a pure function of the *resource*
(`live_compute.cpp:6776`), so any binding whose resource qualifies — ordinary tiled
uncompressed mipmapped 2D texture, `BASE_LEVEL 0`, native straight-copy format — kills the
**whole dispatch** if its binding shape cannot carry the chain, *even for a shader that only
ever reads level zero*. And the shape gate has more failure modes than the one you name:
besides `bi.imported`/`renderer_owned`, it also declines on `bi.unorm_rtt_value_reuse`,
`bi.depth_bits_source`, `cube_face_as_2d`, `sampled_layers != 1`, and
`texel_bytes != mip_chain_bytes_per_block`. I agree the *combination* stays unlikely (render
targets are single-level; cube/array chains are already rejected upstream by
`layer_stride_bytes`), so I am not blocking on it — but the PR body should say what it
actually risks, and a follow-up that carries "this module needs N levels at this binding"
into the `ComputeItem` would remove the decline entirely rather than making it rarer.

**Is it loud? Diagnosable, yes. Loud at runtime, no — and the difference matters here.**
`skip_image` (`live_compute.cpp:5879-5913`) prints one well-formed line naming the program,
address, extent, the reflected shader declaration and the reason, then sets
`images_ready = false` so the dispatch is skipped. Good. But the warning is deduplicated by
`static std::vector<uint64_t> warned` keyed on `gpu_addr`, **for the lifetime of the
process** — so a binding that declines on every frame for ten minutes produces exactly one
line, while the dispatch is dropped 36,000 times in silence. A user sees a rendering defect,
not an error. That is the charter's own rate-limit trap; it is not new to this PR, but this
PR adds a new way to reach it, so it is worth one sentence in the status doc: *if a title
regresses after this, grep stderr for `declared mip chain not materializable`.*

---

# 6. Smaller findings

- **`rdna2_emit_alu.cpp:7844-7847`** asserts *"Hardware clamps a MIMG mip selector to the
  descriptor's LAST_LEVEL"* as bare fact, with no ISA citation and no `CONFIDENCE` marker —
  in a file whose sibling predicate carries *"CONFIDENCE: HIGH — verified per-branch"* on
  precisely this kind of claim. I could not settle it from the public web. The clamp is the
  right defensive choice regardless (unclamped is SPIR-V UB), so this is a record problem,
  not a code problem: cite the ISA section or mark it `CONFIDENCE: MED`. Note also the bound
  is `min(declared_mip_levels, full_mip_chain_levels(w,h)) - 1`, which equals `LAST_LEVEL`
  only when the declaration is consistent with the extent — worth saying, since the
  over-declared case is one of your own test arms.
- **`rdna2_mimg_dynamic_mip_shape` accepts `mimg_dim == 5` and returns the `+3` mip VGPR, but
  `emit_alu` requires `in.mimg_dim == SQ_DIM_2D`** (`rdna2_emit_alu.cpp:7477`). The 2D_ARRAY
  arm is unreachable, while the header comment at `rdna2_decode.hpp:482-489` advertises
  `2D_ARRAY [x,y,slice,mip]` support the emitter does not have. Either drop the arm or say it
  is reserved.
- **DMASK widened from `{1, 0xf}` to anything.** `write_mimg_results` handles arbitrary DMASK
  generically so this is safe, and I checked it — but the one execution test uses `dmask:0x1`
  while the motivating title's packets are `dmask:0x3`. A second arm at `dmask=0x3` on a
  two-component resource would cover what Sonic actually issues.
- **`shader_resource_uses_ordinary_2d_image(*res, true, false, false)`** (`rdna2_emit_alu.cpp:7478`)
  passes literal shader flags, so it degenerates to `depth == 1 && !depth_compare &&
  img_dim in {1,5}`. It reads like a shader-shape check and is not one; an inline comment
  would stop the next reader mis-reading it as one.
- **`test_shader_resources.cpp`, the over-declared arm** asserts
  `shader_resource_compute_mip_chain_levels(over_declared) <= full_mip_chain_levels(...)`.
  A plain decline (1) satisfies `<=` too, so the arm cannot distinguish "clamped to the
  extent" from "rejected outright" — which is the thing it is named for. Assert the exact
  expected value.
- **`docs/RECOMPILER_REMAINING.md:465`** (Stray) says a faithful fix needs "steps 2-3: real
  `mipLevels` on the image, a per-level copy from the guest's own tiled bytes, and an
  `OpImageFetch` with an explicit `Lod`". This PR lands exactly those, for compute only.
  The row stays *true* (Stray's op is a fragment op, and its resource is `mip_tail=1` which
  the plan rejects anyway) — but one clause noting that the compute half now exists would
  save the next reader re-deriving it.

---

# 7. What I checked and found sound

Recorded so you know which parts I actually verified rather than skipped:

- **Two-consumer invariant holds on every path I could find.** `make_shader_compile_key`
  stores `materialized_mip_levels` (`gpu_executor.cpp:1572-1573`) and hashes it (`:811`), so a
  cached compute module can only be reused for a resource that yields the same count; the
  recompiler (`rdna2_emit_alu.cpp:7472-7473`) and the backend (`live_compute.cpp:6776`) then read
  the same function on the same table. The **alias** path was the one I expected to break it
  and it does not: `same_view` was extended with all six chain-determining fields
  (`live_compute.cpp:6565-6572`), aliased bindings `continue` at `:6623` before the creation
  block and are skipped again at `:8570` before the copy loop, so `mip_chain.levels[0]` is
  never dereferenced against an empty plan. (One theoretical hole: `same_view` compares
  `prior.storage == bi.storage` — reflection-derived — but not `p->cls`, nor `sample_count`.
  Both are pre-existing and I could not construct a real case.)
- **`ComputeImageCacheKey`: exactly two positional inits, both correct.**
  `storage_image_cache_key` (`live_compute.cpp:878-893`) and `bi.cache_key = {...}`
  (`:7115-7125`). The storage-path key at `:7354` uses the helper's default `1`, which is
  right because a chain requires `!bi.storage`. Grepped for `ComputeImageCacheKey{` and
  `= storage_image_cache_key` across `src`, `frontends`, `tests` — nothing else.
- **`ShaderResource`'s five new fields sit after `declared_mip_levels`**, and the only
  positional `push_back({...})` inits are test fixtures supplying the first eight members —
  unaffected.
- **Sampling really is unchanged.** `VkSamplerCreateInfo smci{...}` at `live_compute.cpp:8109`
  never assigns `maxLod`, so it stays `0.0f` and every `image_sample*` still resolves level
  zero. The "byte-identical" claim holds.
- **Provenance plumbing.** `image_base_level_view` sets `chain_*` only on the modelled
  thin-2D path after `layout.supported` (`agc_shader_layout.cpp:512-516`); all three
  `shader_resource_apply_mip_chain_provenance` sites are behind a `view.supported` guard
  (`agc_shader_layout.cpp:1177`, `gpu_executor.cpp:7282`, `:7806`); the shifted-view case is
  double-guarded by `mip_chain_base_level != 0` and `element_width != resource.width`. A
  *missed* provenance site would fail closed, which is the right direction.
- **Vulkan-level correctness of the per-level copy.** `alignment = max(4, texel_bytes)`
  satisfies both `bufferOffset` rules for every whitelisted texel size (including
  `Uint8x1` -> block size 1); level extents are `max(w >> level, 1)`, matching Vulkan's own
  mip extents; `ici.mipLevels = min(declared, floor(log2(max(w,h))) + 1)` never exceeds the
  legal count; view/barrier subresource ranges were widened consistently; regions are tightly
  packed with `bufferRowLength = 0`.
- **The chain allocation is bounds-checked before any dereference**
  (`live_compute.cpp:6984-6993`), including the `level_zero_offset > gpu_addr` underflow case,
  and `guest_readable` really probes on this platform (it is not vacuous in the test).
- **The regression test's discriminator is real, and better than the body claims.**
  Recompilation *itself* only succeeds via `dynamic_mip_load`: with `proven_zero_mip = false`
  and `f0040108` (UNRM/GLC clear) the specialization path declines, so `!chain_spirv.empty()`
  is already an arm. The result comparison is well-formed — `W = 64`, level 1 is 128 wide, so
  all 64 fetched texels are in range — and the companion arm genuinely rules out level zero
  (the two generators differ at index 0). The destination binding keeps `64x1`, which is
  right.
- Old captures load (version gate); new captures are cleanly rejected by older readers
  ("unsupported capture version 57"); no struct-size or `memcmp` assumptions anywhere on the
  two grown structs.

---

# 8. Verification I ran

Own worktree at `d880fd0e`, own fresh `build-linux`, `ps5ys` distrobox, Debug + Ninja,
configure exit 0 with `Vulkan found — offscreen graphics harness enabled`.

**Mutation reproduced exactly as reported.** `ici.mipLevels = bi.mip_levels;`
(`live_compute.cpp:8038`) -> `= 1;`, rebuilt, run, restored — build exit code checked at
every arm, so each run is the binary it claims to be:

| arm | build exit | ctest exit | result |
| --- | --- | --- | --- |
| baseline | 0 | 0 | 3/3 passed (`shader_resource_contract`, `gpu_capture_roundtrip`, `game_compute_exec`) |
| mutant | 0 | 8 | 1 failure of **420 executed assertions**: `FAIL: IMAGE_LOAD_MIP at a runtime mip returns LEVEL ONE's own guest texels` |
| restored | 0 | 0 | 3/3 passed |

One observation from the mutant arm worth folding back into the test: the companion arm
(*"level one is distinguishable from level zero in this fixture"*) **stayed green under the
mutant** — the single-level image's out-of-range fetch returned neither level, so that arm
never fired. It is still a correct arm (it would catch a fetch that silently resolved level
zero), but it is not what caught this mutation, and the PR body's framing implies it is
load-bearing. The real anti-coincidence guard here is something the body does not claim:
recompilation *itself* only succeeds through `dynamic_mip_load`, because `proven_zero_mip =
false` plus the `f0040108` modifier shape makes the specialization decline. That is the
stronger arm — say so.

**Not run**, and stated rather than implied: the full suite. Four other lanes were building
concurrently and a full link was taking ~8 minutes per archive under I/O contention, so I
built only the three affected test targets and ran those three cases. CI's own Linux job on
this exact head is green over the full set, which covers what I skipped — except, as §1
shows, for the thing no test looks at. No snapshots, no title launched (another lane holds
the GPU).

The `DecodedImageView` A/B in §1 was a standalone TU compiled twice against the two headers
with `-Wall -Wextra`, not a claim from reading.

---

**Verdict: REJECTED**, on §1 alone. §2 and §3 should land with it. §4 is a structural
follow-up I would like but will not hold the PR for if you would rather do it separately.
Re-review is quick once §1 is a designated initializer with an arm behind it.

mattias800 (commented):
Blocking finding fixed and re-pushed at `68debfc0`. Thank you — that one was silent in exactly the
way the charter's trap entry describes, and I had already handled it correctly for
`ComputeImageCacheKey` in the same change, which makes missing it worse rather than better.

**The fix.** `DecodedImageView`'s only positional aggregate is gone. The fallback view is now a
named `unmapped_format_image_view` in `agc_shader_layout.hpp` that assigns every field by name,
carries a comment saying why it must not go back, and is directly callable from a test — so the
production path and the arms exercise the same function rather than two copies of one initializer.

**Mutation-verified**, build exit code checked before believing either result:
`baseline_rc=0 / mutant_rc=8 / restored_rc=0`, all three builds exit 0. Restoring the positional
form reddens **four** arms — both fail-closed guards (`BASE_LEVEL`, `BASE_ARRAY`) and the "reports
its mip chain as NOT MODELLED" arm, because the `supported` expression lands on
`chain_element_width`. That second signature is the one I would not have thought to assert without
your diagnosis.

On the rest:

1. **v57 / replay** — I re-derived it from the source before touching the wording rather than taking
   the citation: `bind_range` sets `r.host_data` for every replayed resource, and
   `resource_footprint_impl` starts from `r.size`, the selected level. You are right and the claim
   was wrong. Corrected in the format comment, the status doc and the PR body, and filed as **#3202**
   with the three steps it would actually take. The tail stays as the descriptor half so the
   placement is not lost from capsules taken meanwhile — say the word if you would rather it came
   out entirely and went in with #3202.
2. **v57 value coverage** — added: five distinct values round-tripped individually, a v56 downgrade
   arm reading as "not modelled", and a partial-placement rejection arm. Your point that this is not
   academic given §1 is well taken.
3. **Two straight-copy predicates** — I went with a **drift detector** rather than collapsing them:
   only the branch that can write levels 1..N-1 sets a flag, and a chain reaching the end of the
   conversion cascade unwritten declines instead of copying uninitialized staging. Collapsing would
   mean exporting `live_compute`'s whole format classification, and the charter's "prefer
   experiments that detect their own invalidity" reading pushed me here — but this is a judgement
   call and I will collapse them if you would rather.
4. **Blast radius** — accepted in full, including the half I got wrong: `skip_image` dedupes per
   address for the process lifetime while the dispatch is dropped every frame, so "the failure is
   loud" was not true. There is now a `[compute-mip-chain] declined …` line on a geometric schedule
   with the occurrence count and the `imported`/`rtt` terms, and both the site and the status doc say
   the decline is not conditional on the module actually using the op.
5. **Smaller ones** — all five done: the clamp separates its certain half (out-of-range `Lod` is
   SPIR-V UB) from its inferred half (saturating to LAST_LEVEL, doc 70648) and is marked
   `CONFIDENCE: MED`; the 2D_ARRAY arm of `rdna2_mimg_dynamic_mip_shape` now has a decoder test
   instead of being advertised and dead; there is an **executed `dmask:0x3`** arm on a two-channel
   surface that stores the second returned component, since `0x1` could not observe the widening;
   and the `<=` arm is an equality a decline cannot satisfy.

**On the anti-coincidence observation** — you are right, and I have written it into the test rather
than leaving it as folklore. The "distinguishable from level zero" arm stays green under the mutant
because an unwritten destination also differs from level zero; the real discriminator is that
recompilation only succeeds through the dynamic-LOD path at all. There is now an explicit "every
destination texel was actually written" arm, and a comment saying not to read the second arm as the
discriminator.

Current head: build exit 0, `ctest --no-tests=error` exit 0, **318/318**. Re-running the original
mutation on this head still reddens exactly the two assertions that matter (2 of 426 in
`game_compute_exec`), and the new `dmask:0x3` arm is one of them.

mattias800 (commented):
## APPROVED

Re-reviewed at head `68debfc0` (confirmed against `git ls-remote`, not just `gh`). Own worktree
re-pointed to this head, clean tree, `ps5ys` distrobox. Three mutation arms run below, each with its
build exit code checked first.

§1 is fixed, and fixed better than I specified — the aggregate is gone rather than corrected, so the
defect class is retired at that site instead of being re-spelled. Everything else I raised is
addressed. Three things to land before merging, none needing another review round, and one design
decision you asked me to make.

---

# The §1 fix — verified, including the part you asked me to check

**The positional aggregate is gone from the whole tree.** `grep -rn 'DecodedImageView{'` over `src`,
`frontends`, `tests`, `tools` returns nothing. Every other view in the codebase comes from
`image_base_level_view` — the four decode sites plus the test fixtures — so there is no second
hand-built view to diverge. `unmapped_format_image_view`
(`prosper/src/gpu/agc/agc_shader_layout.hpp:254-263`) has exactly one production caller
(`gpu_executor.cpp:7804`) and the new arms call **that same function**, not a copy of it. Confirmed.

**The mutation reproduces exactly as you described, and the "NOT MODELLED" arm asserts what it
appears to.** Restoring the positional form *inside* the function, rebuilt (exit 0) and run:

```
[FAIL] the unmapped-format view reports its mip chain as NOT MODELLED (#3048)
[FAIL] unmapped format with a selected BASE_LEVEL fails closed (no mip offset is derivable)
[FAIL] unmapped format with a selected BASE_ARRAY fails closed (it would bind slice ZERO)
[FAIL] unmapped format with both selectors set fails closed
== FAIL: 4 ==                       build_shader_resources, ctest exit 8
```

The first one is not incidental: with `base_level == 0 && base_array == 0` the reddened value is
`chain_element_width == 1`, because that is where the `bool` lands. So it is genuinely the
field-ordering canary, *and* it independently pins a real contract (`mip_chain_plan.cpp:60` keys on
a zero element extent meaning "not modelled"). It is not a tautology and it is not redundant with
the three fail-closed arms — it is the only one that fires in the *supported* case.

**The residual gap, and a one-line way to close it completely.** The arms pin the *function*, so
they catch a revert of its body. They cannot catch a future positional aggregate at a **new call
site**, or the existing call site being changed to bypass the function — which is the shape the
original defect actually had. I checked whether the production path is covered some other way: the
tests that drive `realize_compute_dispatches` (`test_gpu_execute.cpp`, `test_shader_recompile_cache.cpp`)
never build an unmapped-format storage descriptor with a non-zero `BASE_ARRAY`, so that path is
uncovered. Right now the only defence is the comment saying "do not add a positional one elsewhere".

That can be made a compile error instead of a rule, in one line. The project is `CMAKE_CXX_STANDARD 20`
(`CMakeLists.txt:4`), and C++20's P1008R1 disqualifies aggregate status for any **user-declared**
constructor — `= default` included. I checked it rather than asserting it, with a standalone TU
mirroring the struct:

```
    DecodedImageView() = default;   // added

    V v;      // compiles
    V v2{};   // compiles
    V bad{1, 2, 3, 0, false, 0, 0, 0, 0, 0, true};
    ^ error: no matching function for call to 'V::V(<brace-enclosed initializer list>)'
```

`g++ -std=c++20`, exit 2. Value-init and default-init are untouched; only the thing that broke keeps
working the way it broke. Worth taking — it is strictly stronger than the comment, and it costs
nothing. Not a condition of this approval.

---

# Your four points

**1. v57 does not deliver replay — the corrected wording does not over-claim.** I re-read all three
places. The format comment (`gpu_capture.cpp:140-149`) now says plainly that a replayed resource is
`host_data`-backed, that its blob covers only the selected level's footprint, and that
`gpu_replay` "still refuses IMAGE_LOAD_MIP exactly as it did before v57". The status doc adds "Do
not take a capsule expecting to study the fetch offline yet", and the PR body labels it as corrected
from its first draft. #3202 is open, labelled `bug` / `area:infra` / `game:sonic-frontiers`, and its
title names the real cause ("captures own the selected level, not the allocation"). No residual
over-claim — the tail is now described as the descriptor half of a two-part change, which is what it
is.

**2. v57 value coverage — good, but let me correct the framing rather than nod at it.** The five
values are `60 / 33 / 4 / 5 / 2`: pairwise distinct, none 0 or 1, all within the reader's validation
range, so **any transposition among the five reddens**. The v56 downgrade arm asserts the fields
come back explicitly zero, and the partial-placement arm zeroes `element_width` only and requires
the exact string `"invalid mip-chain placement state"` — which is what stops it passing on some
generic truncation error, and it is also self-validating: if the offset arithmetic were wrong it
would land on an already-zero dword in another table's records and the arm would fail rather than
silently pass. That is the coverage this tail should have had from the start.

But **this is not "the coverage that would have caught §1"**, and it is worth being exact because
the two are easy to conflate. §1 was a field-ordering defect in a *different struct*
(`DecodedImageView`) in a *different file* (`gpu_executor.cpp`) with no serialization involved; no
capture assertion could ever have seen it. What is true is the weaker and still useful claim: this
is the coverage that would catch a defect **of the same class** inside the v57 tail. The thing that
catches §1 is the `build_shader_resources` block, and it does.

**3. Detector or genuine collapse — my call: keep the detector.** Not as a compromise; it is the
better design here, and a collapse would actively hurt.

The two predicates are not two spellings of one question. `native_straight_copy_sampled_format` is a
property of `(format, components)` alone, and it lives in `src/gpu/resources/`. The backend's branch
is a disjunction of about twelve locals that depend on `bi.storage`, `bi.imported`, `renderer_owned`,
`dim_3d`, `image_descriptors[i].image_depth` and `live_target.format` — none of which exists in the
resource layer. Collapsing means either hoisting live-backend binding state into
`src/gpu/resources/` (inverting the dependency direction) or passing the predicate down as a
parameter, which is the detector with extra ceremony.

The decisive reason is the one your own architecture already committed to.
`shader_resource_compute_mip_chain_levels` has **three** callers, and two of them —
`emit_alu` (`rdna2_emit_alu.cpp:7472`) and `make_shader_compile_key` (`gpu_executor.cpp:1572`) — have
no `BoundImage`, no reflected descriptor and no live target. A genuine collapse would make the level
count depend on state those two cannot see, which breaks the single-derivation invariant that is the
entire point of this change. That is not a tradeoff I would take to remove a duplicated `switch`.

And the collapse would look fine today and rot on exactly the case that matters: for the seven
currently whitelisted (format, components) pairs the backend booleans all reduce to
`!bi.storage && format && components`, which *is* computable from a `ShaderResource` — so a shared
predicate would compile and pass. The first widening that matters, `Float16`, has a backend
predicate of `sampled_components == 4 && (bi.imported || dim_3d)`, and the collapse would have to be
undone. The detector survives that unchanged.

I checked the placement too: the flag is set only in the native branch, and the check sits at the
end of the same `else if (needs_sampled_upload)` cascade, so every conversion branch that could run
for a whitelisted format is downstream of it. The paths that bypass the cascade
(`upload_skipped` / `compute_transfer_seed_borrowed`, DCC fast clear, BCn) are all unreachable with
`mip_levels > 1` — `cache_candidate` is forced false, and the plan rejects `metadata_addr` and
block-compressed formats.

**One ask, and it is the only thing keeping this from being airtight**: the detector's branch is
unreachable by construction today, so nothing exercises it and nothing tells an author widening the
whitelist that a second half exists. `native_straight_copy_sampled_format`
(`mip_chain_plan.cpp:18-22`) has no cross-reference to it. Two lines of comment there — "the backend
branch that can copy these levels is `live_compute.cpp`'s native-texel case; widening this list
without widening that one is caught by the detector after the conversion cascade, as a decline" —
puts the pointer where the person about to break it will be reading.

**4. Blast radius.** Accepted correction landed accurately at the site
(`live_compute.cpp:6791-6813`), in the status doc, and in the PR body, including the half I got to
by reading `skip_image`'s process-lifetime `warned` vector. The geometric schedule with an
occurrence count and the explicit "this dispatch is dropped every frame" text is a better answer
than I asked for — a lane reading a black frame now sees it recurring rather than a single line from
an hour ago.

**5. Smaller ones.** All done, and the two I would have re-checked hardest hold up. The clamp comment
now separates the certain half (SPIR-V UB) from the inferred half at `CONFIDENCE: MED` and names the
competing candidate (returning zeros), which is the honest shape. The `dmask:0x3` arm decodes the
way its comment says — word0 `f0040308` gives DMASK 3, word1 `00050002` gives VADDR `v2` / VDATA
`v[0:1]` / SRSRC `s[20:27]`, and the store's `00020105` writes `v1`, the second returned component,
so it observes what a DMASK-1 arm cannot. The `<=` arm is now an equality that also pins
`full_mip_chain_levels(256,256) == 9`, which a decline cannot satisfy. The decoder arms exercise the
2D_ARRAY form (`f0040328`, dim 5, mip at `v4`) so the header no longer advertises something dead.

---

# The anti-coincidence arm: the new arm is right, its comment is not — and that one is on me

The arm itself is load-bearing in the way it should be: it separates "wrote the wrong thing" from
"wrote nothing", which is the diagnosis the old pair could not make. Under a correct run it is
implied by the level-one assertion; its value is entirely in the failure case, and the comment
correctly tells the next reader not to credit the wrong arm.

**But the mechanism the comment states is false, and I measured it.** The comment says the
"distinguishable from level zero" arm stays green under the reverted backend "because an UNWRITTEN
destination also differs from level zero". Reverting `ici.mipLevels = bi.mip_levels` to `= 1`
(build exit 0) reddens exactly **2 of 426** assertions —

```
FAIL: IMAGE_LOAD_MIP at a runtime mip returns LEVEL ONE's own guest texels
FAIL: #3048: dmask:0x3 returns level one's SECOND component in the second VGPR
```

— and **both "every destination texel was actually written" arms stay green**. So the destination is
*not* unwritten: the dispatch runs and stores something to every texel, and that something is
neither level one nor the `0x77777777` / `0x66666666` sentinel. The out-of-range fetch returns a
value that happens to differ from level zero, which is why the arm is vacuous — not absence of a
write. (I am not claiming what the driver returns; the evidence supports only "written, and not the
sentinel".)

That sentence came from my first review, where I offered it as a reading rather than a measurement,
and it has now been written into the code as if it were established. That is precisely the
citation-authority failure the charter records, with me as the source, so please fix the comment
rather than the arms: something like *"the out-of-range fetch on a one-level image still writes every
destination texel, with a value that is neither level one nor level zero — so this arm is satisfied
by a wrong result, not by an absent one."*

---

# Before merging — three things, none needing another review round

1. **This head has had no CI at all.** `mergeable` is `CONFLICTING`, and that suppresses CI
   entirely (trap 242): `gh pr view --json statusCheckRollup` returns **0** checks and
   `repos/:owner/:repo/commits/68debfc0/check-runs` reports `total_count: 0`. I confirmed the
   conflict is real with `git merge-tree --write-tree origin/master 68debfc0` — a content conflict
   in `prosper/frontends/shared/live/live_compute.cpp`, `origin/master` now being `38f21d07`.
   So the 318/318 you have is a Linux-local result on a tree that has never been merged with master;
   Windows, macOS, Docs and packaging have not seen a line of it, and the conflicting file is the one
   this PR changes most. Merge master in, re-run the local matrix on the *merged* tree, and require
   real green CI before merging.
2. **`SONIC_FRONTIERS_STATUS.md:589` still asserts the clamp claim as bare fact** — "The LOD is
   clamped to the materialized level count, **which is what hardware does with a T#'s `LAST_LEVEL`**"
   — while the code comment two commits later says the ISA guide does not spell that out and marks it
   `CONFIDENCE: MED`. The doc is what the next reader reaches first. One line.
3. **The anti-coincidence comment above.**

---

# Verification I ran

Own worktree re-pointed to `68debfc0`, clean tree, existing build dir used by nothing else (trap 243
is about two builds sharing one directory; this one is mine alone, and I confirmed each arm actually
recompiled and relinked rather than serving stale objects — 12 to 22 edges per arm, never zero).
Targeted builds of the five affected test binaries; I did not re-run the full suite, and say so
rather than implying it.

| arm | build exit | ctest exit | result |
| --- | --- | --- | --- |
| baseline `68debfc0` | 0 | 0 | 5/5 — `shader_resource_contract`, `gpu_capture_roundtrip`, `build_shader_resources`, `rdna2_decode_walk`, `game_compute_exec` |
| positional aggregate restored *inside* `unmapped_format_image_view` | 0 | 8 | `build_shader_resources` **FAIL: 4**, the four arms quoted above |
| `ici.mipLevels = bi.mip_levels` -> `= 1` | 0 | 8 | `game_compute_exec` **FAIL: 2 of 426**, the two quoted above; both "actually written" arms green |
| restored | 0 | 0 | 5/5 |

Also checked: no `Claude-Session:` trailers on either new commit; `git diff --check` clean; zero
absolute host paths in the diff; no positional `DecodedImageView` initializer anywhere in the tree.

The C++20 aggregate probe and the earlier §1 A/B were standalone TUs compiled with
`-Wall -Wextra`, not claims from reading.

**APPROVED** on the code. The CI item above is a merge gate, not an objection to the change.

mattias800 (commented):
Answering the one direct question I left implicit: **keep the v57 tail in this PR** — do not pull it
out and land it with #3202.

It is append-only, it fails closed on every consumer, it now has real value coverage, and the
capsules people take between now and #3202 either carry the placement or lose it permanently. The
only thing that made it objectionable was the claim attached to it, and that claim is gone from all
three places. Pulling it out now would mean a second format bump later for the same five dwords, and
every capture taken in between would be missing the descriptor half of the thing #3202 exists to
finish.

Verdict on the head is unchanged: **APPROVED**, with the three pre-merge items in my previous review
(no CI has run on this head at all — `mergeable=CONFLICTING`, trap 242; the status doc's unqualified
`LAST_LEVEL` sentence; the anti-coincidence comment whose stated mechanism I measured to be false).

mattias800 (commented):
All four addressed; head is now `7eb24f85`, rebased onto `bee26b9e`.

**1. CI.** You were right about the cause and trap 242 called it exactly. One content conflict, in
`live_compute.cpp` against #3160's alias census — both sides add an independent statement at the same
point, so I kept both, master's census block first so its comment stays attached to what it
describes. No `BLOG.md` or `GAME_COMPAT_ORCHESTRATION.md` conflict (this branch touches neither).
I ran the trap-41 check rather than assuming: the branch deletes 40 lines against master and every
one is this PR's own.

Rebuilt and re-tested on the rebased tree: build exit 0, `ctest --no-tests=error` exit 0,
**323/323** (master's 318 plus five that landed while this was open). `mergeable` is now `MERGEABLE`
on `7eb24f85` — confirmed against `git ls-remote` before believing it, per your stale-head note —
and all nine jobs are running.

**2. The comment stated a mechanism measurement contradicts.** Confirmed from my own mutant run
before rewriting: under `ici.mipLevels = 1`, **both** "every destination texel was actually written"
arms stay green, so the destination is not unwritten — it is written with something that is neither
level one nor the sentinel. Comment fixed only; the arms are untouched.

The honest framing, now in the code: those two arms are **fixture-validity** guards, ruling out a
fixture that could pass vacuously (level one byte-identical to level zero; the dispatch never
running). They are not mutant discriminators. The discriminator is the level-one equality, and
behind it the fact that this program only recompiles through the dynamic-LOD path at all. The
comment says so and tells the next reader not to draw mutant conclusions from either arm.

Worth naming what happened here, since the charter has a row for it: this is the same
citation-authority failure twice over. My first framing credited the wrong arm; the reviewer's
correction replaced it with a *reading* that I then wrote into the code as established fact without
running the measurement that would have falsified it — which was one mutant run I had already
performed and whose output I had in front of me.

**3.** `SONIC_FRONTIERS_STATUS.md` now separates the certain half (out-of-range `Lod` is SPIR-V UB,
so *some* clamp is mandatory) from the inferred half (saturating rather than returning zeros, read
from doc 70648's BASE_LEVEL/LAST_LEVEL), matching the code's `CONFIDENCE: MED`.

**4. Taken, and it is the better fix.** `DecodedImageView() = default;` makes the struct a
non-aggregate under C++20 (P1008R1), so every positional braced initialization of it is now a
compile error. I verified it in a standalone TU before applying rather than taking it on the
citation: positional gives `no matching function for call to 'V::V(<brace-enclosed initializer
list>)'` while `V v;`, `V v{}` and `return {};` all still compile, on gcc 16 `-Wall -Wextra`. Your
reason for preferring it over my arms is the right one — they pin the *function*, so they cannot see
a positional init at a *new* site, which is the shape the original defect had. I added a
`static_assert(!std::is_aggregate_v<DecodedImageView>)` beside it so removing the line as redundant
fails the build rather than silently reopening the hole.

**On §3 and the v57 framing.** Both corrections accepted, and both are sharper than what I had.
`native_straight_copy_sampled_format` now carries the cross-reference: it names the detector, says
the fix for a widening is to teach the cascade rather than only this list, and records *your* reason
for not collapsing — two of the three callers have no `BoundImage`, no reflected descriptor and no
live target, so the count must stay derivable from the descriptor alone. And you are right that the
v57 arms would not have caught §1; they catch that class *inside the tail*, and what catches §1 is
the `build_shader_resources` block. I have not restated it as broader than that anywhere.

Not merging — over to you once the checks land.

mattias800:
CI status on `7eb24f85`, since the previous head had none:

| conclusion | checks |
| --- | --- |
| success | 9 — Contribution shape, Docs, Linux, Linux App, Linux App Package, Windows App, Windows MinGW, macOS (x86_64 / Rosetta 2), macOS App (SDL3 + MoltenVK) |
| skipped | 2 — Progress tracker, Publish Release |
| failure | 0 |

`mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`, and `headRefOid` matches `git ls-remote` — checked
in that order rather than trusting the first read, per the stale-head note.

`origin/master` has since moved one commit (`6f9e14ff`, unrelated: display-mode derivation).
`git merge-tree --write-tree` against it is conflict-free, so no second rebase is needed unless you
want the branch fast-forwarded first.

Local, on the same head: build exit 0, `ctest --no-tests=error` exit 0, 323/323.

mattias800 (commented):
**APPROVED — merging.**

The reviewer's verdict on `68debfc0` stands, and all four follow-up items are done on `7eb24f85`
with CI green (9 pass / 2 skipping / **0 failures**) — the first CI this branch has ever had, since
`mergeable=CONFLICTING` had been suppressing it entirely (instrument trap 242).

**The structural guard is better than what was asked for, and I verified it rather than taking the
citation.** A standalone TU under `g++ -std=c++20 -Wall -Wextra`: with `DecodedImageView() = default;`
a positional init is a **compile error** (1 error) while `V v;` and `V v{}` compile clean. Adding
`static_assert(!std::is_aggregate_v<DecodedImageView>)` on top means **deleting the line as
"redundant" fails the build** instead of silently reopening the hole — which is the failure mode that
would otherwise reintroduce this in six months.

That matters because it changes the class of defence. The test arms pin the *function*; they cannot
see a positional aggregate at a **new** call site, which is exactly the shape the original silent
defect had. This makes the defect unrepresentable.

**On the false mechanism — the self-assessment here is the most valuable thing in the thread.** The
comment claimed the "every destination texel was written" arms stay green under the mutant *because
the destination is unwritten*. Measured: they stay green because the destination **is** written, with
something that is neither level one nor the sentinel. The arms are now correctly documented as
**fixture-validity guards** (ruling out a vacuous fixture) with the real discriminator named — the
level-one equality, backed by the program only recompiling through the dynamic-LOD path.

The author's own account: this was the citation-authority failure **twice** — their first framing
credited the wrong arm, and the reviewer's correction replaced it with a *reading* that got written
into the code as fact, without running the measurement that falsifies it, *when that measurement had
already been run and its output was on screen*. That is the charter's documented failure mode named
precisely, by the person who committed it, against their own interest. Recording it here because the
next person to read that comment now gets the truth instead of a plausible story.

**Rebase conflict handled correctly**: one content conflict in `live_compute.cpp` against #3160's
alias census where both sides add an independent statement at the same point — **both kept**, master's
census block first so its comment stays attached to what it describes. The trap-41 deletion check was
run rather than assumed: 40 lines deleted against master, every one this PR's own.

**Upheld and correctly reasoned:** the §3 drift detector stays, with the reviewer's stronger
justification recorded at `native_straight_copy_sampled_format` (two of three callers have no
`BoundImage`, no reflected descriptor, no live target, so the count must remain derivable from the
descriptor alone) plus the note that a widening should teach the cascade rather than extend the list.
The v57 framing was corrected rather than restated more broadly — those arms catch the defect class
*inside the tail*; what catches §1 is the `build_shader_resources` block.

**Follow-up #3202 stays open and is linked from the status doc** — replay still cannot materialize a
declared chain, and that limitation is now stated rather than over-claimed.

Verified on this head: `ctest --no-tests=error` **323/323**, exit 0; `mergeable=MERGEABLE`,
`mergeStateStatus=CLEAN`, `headRefOid` checked against `git ls-remote` before trusting it.

PR #3198, opened 2026-09-01
Derive the advertised display mode from the host instead of hardcoding 1080p59.94 (#3017)

### LINKED ISSUE #3017: prosper advertises a hardcoded 1080p @ 59.94 Hz display to every guest, so titles that pace to the reported refresh rate cap at ~60 fps on any monitor
prosper advertises exactly one display to every guest, hardcoded
(`src/hle/graphics/hle_graphics.cpp:142-144`):

```cpp
// The one display we advertise. 1920x1080 @ 59.94Hz (refresh-rate enum 3), 16:9, ~50".
constexpr uint32_t kDispW = 1920, kDispH = 1080;
constexpr uint64_t kRefresh5994 = 3;   // SCE_VIDEO_OUT_REFRESH_RATE_59_94HZ
```

A title that paces itself to the reported refresh rate therefore sits at ~60 fps on ANY host display.
Measured on *The Messenger* with a 240 Hz monitor: `--present-mode immediate` (vsync off) still gives
61-64 fps, so the limit is not the swapchain present mode and not the display — the app only presents
when the guest flips, and the guest is flipping at the rate we told it to.

## It is enforced in two places, and a change needs both

1. **What the title reads** — `g_vo_resstatus` writes `kRefresh5994` at offset `0x10` of
   `SceVideoOutResolutionStatus` (`hle_graphics.cpp:721`).
2. **What actually paces it** — a separate "monotonic ~59.94 Hz timebase" shared *in phase* by
   `sceVideoOutGetVblankStatus`'s count and `sceVideoOutWaitVblank`'s wake boundary
   (`hle_graphics.cpp:727`). A title that waits on vblank is paced by us at 59.94 Hz regardless of
   what the status struct says.

Changing only (1) would give a title a number it cannot act on; changing only (2) would pace it faster
than the mode it believes it is in.

## Why this is not simply a constant to raise

- The field is an **enum**, not a free number, so only rates the platform defines are expressible.
  PS5 does support 120 Hz, so a higher value is legitimate — but it must be a real enumerant.
- **A title designed for 60 may not merely ignore a higher rate.** Anything that advances state per
  flip rather than per elapsed second runs at double speed. This project has already measured that
  shape: *The Plucky Squire*'s in-game time advances per flip, and its cutscene ran 300x fast when the
  flip rate moved. So this needs per-title verification, not a global flip.
- Reporting a rate the host cannot actually present would be a lie the pacing layer then has to
  honour, so the advertised rate should be **derived from the real display** (SDL can report the mode)
  and mapped to the nearest supported enumerant, capped by what the guest asks for — not hardcoded
  higher.
- 4K titles already render above `kDispW`/`kDispH`, so the resolution half of this constant is
  evidently not the whole story either, and the interaction wants checking before the refresh half
  moves.

## Suggested shape

Derive the advertised mode from the host display, map to the nearest PS5 refresh enumerant, drive the
vblank timebase from the SAME value so the two cannot disagree, and gate it behind an opt-in until at
least one 120 Hz-capable title has been checked for flip-rate-dependent logic. Keep 59.94 as the
fallback when the host mode is unknown or unsupported.

mattias800:
Fixes #3017.

## The defect

prosper advertised exactly one display to every guest, hardcoded — two constants in
`hle_graphics.cpp` and a third 600 lines below them:

```cpp
constexpr uint32_t kDispW = 1920, kDispH = 1080;
constexpr uint64_t kRefresh5994 = 3;          // SceVideoOutRefreshRate enumerant
constexpr uint64_t kVblankNs    = 16683350;   // the vblank pacing grid
```

The three agreed only because somebody kept them in step by hand. As the issue puts it, this is
enforced in two places and a change needs both: **what the title reads**
(`sceVideoOutGetResolutionStatus`, `hle_graphics.cpp:779-781` on master) and **what actually paces
it** (the vblank timebase at `hle_graphics.cpp:816`, which `hle_kernel_time.cpp`'s kevent pump has
also scheduled on since #3024). Changing one without the other either gives a title a number it
cannot act on, or paces it in a mode it does not believe it is in.

This is the failure mode the charter warns about: it fails in the direction that looks safe. Nothing
crashes, the title behaves *correctly* given what it was told, and so nobody files it — while a 4K
host is told 1080p and every host is told 59.94 Hz.

## The fix

A new pure header, `src/hle/graphics/display_mode.hpp`, resolves **one** `AdvertisedDisplayMode`
from the host's real mode. Both enforcement points now read their fields off that single object, so
they cannot disagree — there is no longer a second place for them to disagree in. The mode resolves
**once** per process (the vblank epoch is first-use anchored; a period that moved mid-run would step
the guest's vblank count sideways).

The frontend publishes the host's real mode as `PROSPER_HOST_DISPLAY_MODE=<w>x<h>@<hz>` **before the
guest boots** — position is load-bearing, since the guest thread starts at `main.cpp`'s
`start_guest()` and the mode resolves lazily on first use.

### Deliberately opt-in

`PROSPER_DISPLAY_MODE` (and `--display-mode`) defaults to **`legacy`**, byte-identical to previous
behaviour. Titles *act* on this value — anything advancing state per flip rather than per elapsed
second runs fast at a higher rate, which this project has already measured (*The Plucky Squire*).
The issue asks for exactly this gating, and it is why the default moves nothing.

| policy | resolution | refresh |
| --- | --- | --- |
| `legacy` (default) | 1920×1080 | 59.94 (enum 3) |
| `host` | largest PS5 output mode the host can show | derived, capped at 59.94 |
| `host-high-refresh` | same | additionally allows 119.88 (enum 6) |

### Two properties that keep this a derivation rather than a different constant

- **The refresh field is an enum**, so advertising a rate asserts the integer Sony assigned it, and a
  wrong integer is worse than not offering the rate. Each table row carries an evidence grade —
  59.94 `HIGH` (prosper's own live titles; its period is the existing grid), 50 `MED`
  (secondary-implementation agreement), 119.88 `LOW` — and the policy carries a floor, so an
  enumerant prosper cannot defend is **unreachable by default rather than merely discouraged**. No
  primary evidence pins 119.88's value, and no 120 Hz-capable title has been checked here for
  flip-rate-dependent logic; that is precisely why it needs the extra opt-in.
- **The host's desktop resolution is never passed through.** A title told it is attached to a
  1366×768 panel is told something no PS5 ever reports, so the largest real PS5 output mode the host
  can display is chosen instead (720p / 1080p / 4K, floored at 720p). Likewise a rate above what the
  host can present is never advertised — the pacing grid would then honour it and wake the guest
  faster than the panel can show.

## Verification

Linux, distrobox, **fresh** Debug configure (`-G Ninja`).

- Build: exit **0**, zero errors.
- `ctest --no-tests=error`: **319/319 passed, exit 0** (318 registered on master + the new case).
- `videoout_queries` — the existing guard asserting the 1920×1080 / enum-3 default — **still
  passes**, in a process where the new variables are unset. That is the untouched-default arm.

### Mutation verification

The important arm is the one proving the answer is **derived**, not one constant swapped for
another. Two independent mutations, each built and run:

| arm | build rc | test rc | result |
| --- | --- | --- | --- |
| baseline | 0 | **0** | 37/37 arms pass |
| **A** — `select_display_mode()` hardcoded back to the fallback | 0 | **1** | **14 arms fail**, including every "a different host must produce a different advertised mode" arm |
| **B** — only `vblank_period_ns()` hardcoded to 16683350, status struct left derived | 0 | **1** | **exactly 2 arms fail** — the two pinning the enforcement points together |
| restored | 0 | **0** | passes again |

Arm B is the sharpest: a fix that changed what the title *reads* without changing what *paces* it
would pass a weaker test, and fails this one on precisely those two assertions.

## Risk — this can affect titles that currently work

Honestly stated:

- **On the default path, nothing changes.** `legacy` produces the same three values from the same
  code path, and the existing `videoout_queries` guard confirms it. The publication of
  `PROSPER_HOST_DISPLAY_MODE` is inert on its own.
- **The real risk is behavioural and lives behind the opt-in.** Under `host`, a 4K host now reports
  3840×2160, and a title that picks render resolution from this may allocate larger targets or
  select different assets. Under `host-high-refresh` the flip rate doubles, which is unsafe for any
  per-flip logic. Neither is reachable without asking.
- **`vblank_period_ns()` is a function call now, not a constant**, on the vblank status/wait path and
  the kevent pump. It reads an immutable resolved value, so there is no synchronization change, but
  it is no longer constant-folded.
- **Not done, deliberately:** no live title run — another lane holds the GPU. So the derived paths
  are proven by unit + HLE-level tests, not by a booted title. Since the default is unchanged, no
  title's current behaviour depends on this; but the first person to use `host-high-refresh` on a
  real title is doing the per-title verification the issue asks for, and #3017 should stay open in
  spirit for that even though this closes the hardcoding half.
- `g_flip_rate` (`sceVideoOutSetFlipRate`) is deliberately untouched — it is the guest's flip
  divisor, not the display mode, and folding it in here would conflate two different questions.

## Also in this diff

- `src/hle/graphics/AGENTS.md` — the folder had none.
- `docs/GRAPHICS.md` — its 2026-07-04 entry asserted the hardcoded 1080p@59.94 as the whole answer; a
  note records that it is now the default rather than the only mode, so the stale figure does not
  keep carrying the doc's authority.
- `frontends/prosper-app/README.md` — documents `--display-mode`.

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -2072,6 +2072,17 @@ if(TARGET prosper_core)
   target_link_libraries(test_videoout PRIVATE prosper_core)
   add_test(NAME videoout_queries COMMAND test_videoout)
 
+  # #3017: the display prosper advertises must be DERIVED from the host's real mode rather than
+  # hardcoded 1080p59.94. Its own executable, not an arm of test_videoout, because the advertised
+  # mode resolves ONCE per process on first use -- so the derived case needs a process whose
+  # environment is set before any VideoOut call, and test_videoout above is exactly the
+  # complementary process where those variables are unset and the untouched `legacy` default is
+  # asserted. CROSS-PLATFORM for the same reason test_videoout is: it drives the HLE registry and
+  # a pure header, with no OS display access.
+  add_executable(test_display_mode tests/hle/graphics/test_display_mode.cpp)
+  target_link_libraries(test_display_mode PRIVATE prosper_core)
+  add_test(NAME display_mode_derivation COMMAND test_display_mode)
+
   # precise_sleep's deadline post-condition (#3074): every sleep_until_steady_ns() backend must
   # never return before now_ns() >= deadline_ns. Win32SleepFallback did not re-check after its one
   # OS sleep call, unlike the sibling high-resolution-timer path. CROSS-PLATFORM like test_videoout
--- a/prosper/docs/GRAPHICS.md
+++ b/prosper/docs/GRAPHICS.md
@@ -229,6 +229,15 @@ sceVideoOutGetOutputStatus, `w0hLuNarQxY` sceVideoOutConfigureOutput. Plus fixed
 (was all-zero → now 1920×1080@59.94Hz) and added GetVblankStatus (advancing counter) +
 GetDeviceCapabilityInfo (SDR). All output-struct writes are size-exact.
 
+> **The 1080p@59.94 in this entry is no longer the whole answer (#3017).** It remains the *default*,
+> so this record still describes what a default run advertises — but that pair was hardcoded, and a
+> title that reads the advertised mode and paces to it was therefore told 59.94 Hz on any host and
+> 1080p on a 4K one. Since #3017 the mode is **derived** from the host's real display under
+> `PROSPER_DISPLAY_MODE` (`legacy` | `host` | `host-high-refresh`), and the vblank period that paces
+> the guest comes from the *same* resolved mode rather than from a third constant kept in step by
+> hand. `src/hle/graphics/display_mode.hpp` holds the policy and the evidence grading behind each
+> refresh enumerant; deriving is opt-in because titles act on this value.
+
 **Verified the game requests exactly what we advertise:** `SetBufferAttribute2` arrives with
 `width=0x780 (1920)`, `height=0x438 (1080)`, `pixel_format=0x8000000000000000` (PS5 A8R8G8B8 sRGB);
 `RegisterBuffers2` registers **3** framebuffers (triple-buffered). Recorded them in a display-buffer
--- a/prosper/frontends/prosper-app/README.md
+++ b/prosper/frontends/prosper-app/README.md
@@ -216,6 +216,30 @@ tuning one; leave it unset for normal use.
 - `--record-axis flip|pad-read` — timestamp recorded intervals by display flips (the backward-compatible
   default) or by successful guest input-state reads. Pad-read routes remain stable when presentation
   pauses while the title continues polling input.
+- `--display-mode legacy|host|host-high-refresh` — which display prosper tells the **guest** it is
+  attached to. See below. Default `legacy`.
+
+## `--display-mode`: what the guest is told it is plugged into
+
+This is not a window or swapchain setting — it is the display prosper *advertises* through
+`libSceVideoOut`, which some titles read to pick their render resolution and to pace themselves. It
+also sets the vblank period the guest is woken on, because those two must agree: a title paced at a
+rate it does not believe it is in behaves as badly as one told a rate it cannot act on.
+
+- `legacy` (**default**) — always 1920×1080 @ 59.94 Hz, exactly what prosper advertised before #3017.
+  Deriving is opt-in because titles *act* on this: anything that advances state per flip rather than
+  per elapsed second runs fast at a higher rate.
+- `host` — derive resolution and refresh from the real display. The resolution becomes the largest
+  **PS5 output mode** the host can show (720p / 1080p / 4K — never the host's own desktop resolution,
+  which may be one no PS5 ever reports). Refresh is capped at 59.94: the 119.88 enumerant's integer
+  value rests on no primary evidence, so deriving alone never reaches it.
+- `host-high-refresh` — additionally allow the 119.88 Hz enumerant on a host that can present it.
+  **Experimental**, for exactly the reason above; verify the title before trusting a run made with it.
+
+The app publishes the host's mode as `PROSPER_HOST_DISPLAY_MODE=<w>x<h>@<hz>` before the guest boots;
+`PROSPER_DISPLAY_MODE` is the same policy switch for harnesses that do not go through this frontend.
+An unknown or unparseable host mode falls back to 1920×1080 @ 59.94, and the chosen mode is printed
+once at startup (`[vo] display: …`).
 
 ## `--fps`: two numbers, and the first one is the honest one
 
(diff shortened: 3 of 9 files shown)


mattias800 (commented):
**APPROVED**

**Mutation B verified independently, and it is the arm that matters.** Hardcoding *only* the pacing
grid — `mode.vblank_period_ns = 16683350` — while leaving the status struct fully derived reddens
**6 arms** (`rc=8`, `build_rc=0`, 0 errors, so not a stale binary); restored, `rc=0`. That is the
claim the design rests on: **a fix that changed only what the title reads would not pass.** Most
"derive it" changes cannot demonstrate that, because their tests only check the value that was
already wrong.

*(Method note on my side: my first attempt at this mutation had a regex that never matched, so
nothing was applied — and the run still reported `build_rc=0`, `ctest_rc=0`, `0 failing arms`. Read
carelessly that says "the test is not load-bearing", when it actually said "I re-ran the baseline".
The numbers above are from the corrected run against the real assignment. Worth stating because a
false negative here argues for deleting a guard that works.)*

**The defect is the safe-looking kind, which is why it survived.** Two hardcoded constants 600 lines
apart — `1920x1080` / refresh enumerant 3 in what the title *reads*, and `16683350 ns` in what
*paces* it (and, since #3024, the kernel's vblank kevent pump) — agreed only because somebody kept
them in step by hand. Nothing crashes; the title behaves correctly on a wrong answer. A 4K host was
told it was 1080p and every title paced to 59.94 Hz regardless.

**One `select_display_mode()` resolving one `AdvertisedDisplayMode`, read by both enforcement
points, is the right shape** — they can no longer disagree because there is nothing to keep in step.

**The evidence grading is what makes this a derivation rather than a swapped constant, and it is
unusually careful.** The refresh field is an **enum**, so advertising a rate asserts Sony's integer —
not a float you can be approximately right about. Rows carry grades (59.94 HIGH, 50 MED, 119.88 LOW)
and the selection policy carries a **floor**, so the enumerant that cannot be defended is
**unreachable by default** rather than merely discouraged. That is the correct treatment of a
reverse-engineered contract: the confidence marker changes what the code can *do*, not just what a
comment says.

**Publishing the host mode before `start_guest()` is load-bearing and documented as such** (the mode
resolves lazily), and the host's desktop resolution is deliberately never passed through — the
largest real PS5 output mode that fits is chosen, and a rate above what the host can present is never
advertised.

**Default `legacy` is byte-identical, so nothing currently working moves.** Making the derivation
opt-in was the right conservatism given per-flip logic like *The Plucky Squire*; the real risk lives
behind `host` / `host-high-refresh` and is stated plainly rather than minimised.

**Correctly scoped:** `g_flip_rate` / `sceVideoOutSetFlipRate` deliberately untouched — that is the
guest's flip divisor, not the display mode. Naming what you did *not* change, and why, is the half
that usually goes missing.

**Honest gap accepted as written:** no live title run, so the derived paths are proven by unit and HLE
tests rather than a booted title, and the per-title 120 Hz verification #3017 asks for remains
outstanding. You said so rather than implying the issue is closed by this.

Also noted and appreciated: you hit the shared-build-directory hazard yourself (two overlapping builds
truncated `libprosper_core.a`) and reported the phantom failure rather than the convenient number.
That is now instrument trap **243**, merged earlier today.

Verified: `ctest --no-tests=error` **319/319**, exit 0. `videoout_queries` still passes as the
untouched-default arm. CI 9 pass / 2 skipping. No `Claude-Session:` trailer, no absolute host paths.

Issue #3135, opened 2026-08-30
recompiler: general SGPR-mask v_mbcnt has no vertex-stage path (fail-closed), so wave-using vertex programs cannot recompile


mattias800:
`v_mbcnt_lo/hi_u32_b32` with a **general SGPR mask** is fail-closed in a vertex stage, so any vertex
program using the real cross-lane form cannot recompile.

## Evidence

*Stray* (`PPSA02101`), title-screen frame, vertex `0x300f190000` (426 dwords). Scanning the dumped
guest program for VOP3 `0x365`/`0x366` finds **both** forms:

| dword | op | src0 |
| --- | --- | --- |
| 4 | `0x365` lo | `193` = inline `-1` (all-ones) |
| 7 | `0x366` hi | `193` = inline `-1` |
| 277 | `0x365` lo | **`s6`** |
| 279 | `0x365` lo | **`s4`** |
| 284 | `0x366` hi | **`s7`** |
| 286 | `0x366` hi | **`s5`** |

The all-ones pair at 4/7 is the canonical wave64 lane-id idiom, which prosper already lowers. The four
at 277-286 read a saved 64-bit mask from SGPR pairs and count the set bits below this lane — genuine
cross-lane state.

Two consequences, both by design and both worth stating because they interact:

* `rdna2_to_spirv.cpp`'s `logical_mbcnt_invalid` is set by the presence of *any* non-all-ones form, so
  the `ngg_logical_lane` model is disqualified for the whole program — including for the all-ones pair
  it could otherwise lower.
* The emitter's general MBCNT arm in `rdna2_emit_alu.cpp` is gated on `(b.is_compute || b.is_fragment)`.
  A vertex stage reaches none of the arms and rejects at the first one, pc=4.

The fail-closed choice is deliberate — the surrounding comment says the wave approximations are "an
exception for the one captured Astro wrapper, not a property of the GS_ALLOC_REQ opcode" — so this
issue is **not** "remove the gate". It is that a vertex program needing real wave semantics currently
has no path at all.

## Impact

Two of the eleven unrealized operations in the measured frame, and both are **full-screen 3840×2160**
draws.

## Note for whoever picks this up

The all-ones arm's own comment argues its lowering "needs neither `allow_wave` nor a subgroup/LDS
reduction and stays exact under divergent control flow" — reasoning that is stage-independent. Letting
a vertex stage use *that* arm is a much smaller step than modelling general masks, and would not by
[...]

Issue #3130, opened 2026-08-30
recompiler: every FRAGMENT diagnostic reports program=0x0, so a failing fragment shader cannot be named


mattias800:
Every recompiler diagnostic emitted from the **fragment** path reports `program=0x0`, so a failing
fragment shader cannot be named — and the failure is invisible unless you go looking for it, because
the line is otherwise complete and reads as a real address.

## The site

`prosper/src/gpu/recompiler/rdna2_to_spirv.cpp:2933`, in `recompile_fragment_impl`:

```c
const RecompileDiagnosticContext diagnostic{RecompileDiagnosticStage::Fragment, 0};
```

The `0` is the program address, hardcoded. `recompile_fragment_impl` never receives one:
`recompile_fragment(const uint32_t* code, size_t dwords, ...)` (`rdna2_to_spirv.hpp:463`) takes the
code pointer and length only, while `recompile_compute` is given the address and its diagnostics name
it correctly.

## What it costs

Concretely, from the Stray investigation (#3126) this was found in. `[mimg-unresolved]` — an
unconditional, well-designed diagnostic that reports exactly which provenance route failed for an
image op — printed five failures on one routed boot:

```
[mimg-unresolved] program=0x0 pc=10 srsrc=s0  srt_tag=NONE ... key_res=null written=0
[mimg-unresolved] program=0x0 pc=20 srsrc=s0  ...
[mimg-unresolved] program=0x0 pc=32 srsrc=s8  ...
[mimg-unresolved] program=0x0 pc=69 srsrc=s8  ...
[mimg-unresolved] program=0x0 pc=89 srsrc=s16 ...
```

Every one of those draws is dropped, and **none of them can be attributed to a shader.** In #3126 I
matched them to three stages by comparing SRSRC sets against a separate per-stage dump, concluded a
mechanism from that match, and had to retract it — the match was a guess, and the `Fragment` stage tag
means the stages I matched them to (compute) were the wrong ones. That retraction is the cost of this
line, and it will be paid again by whoever reads a `program=0x0` next.

The diagnostic's own dedupe key makes it worse rather than better:

```c
static std::set<std::tuple<uint64_t, uint32_t, int>> mimg_reported;   // (program, pc, srsrc)
```

with a comment explaining that keying on `pc` alone would "let the first program to reach one silence
all later programs and attribute its line to a shader the reader is not looking at". With the address
pinned at 0 for every fragment shader, the key degenerates to `(0, pc, srsrc)` and does precisely
[...]

Issue #1076, opened 2026-07-20
GPU: lower Astro fragment MBCNT + GDS append wave-compaction sequence


mattias800:
## Failure

On current merged master `95ce443`, Astro Bot reaches `title_controller_ship` and the former #1054 PC-relative title pixel shader now recompiles. The next exact title draw failure is fragment shader `0x50054ab00` at PC 3:

```text
[recompile-reject] pc=3 words=d7660007,0001007f fmt=9 op=0x366
[exec-recompile-reject] es=0x500549e00 ps=0x50054ab00
```

The shader begins with `v_mbcnt_hi_u32_b32 v7, exec_hi, 0`, later applies the paired low-half MBCNT, then executes GDS `ds_append` and uses the allocated index for buffer access. Prosper currently restricts MBCNT and append wave emulation to compute/workgroup execution, so accepting the first instruction alone would merely move rejection to the coupled append.

## Required behavior

Derive a graphics-stage lowering for this exact wave-compaction sequence from the live shader and AMD RDNA 2 ISA semantics. Preserve active-lane prefix counts, the append counter allocation/broadcast contract, EXEC predication, and cross-draw/device-global counter identity. Do not substitute per-invocation zero or per-workgroup LDS semantics where they are observably different. Unsupported mask/counter forms must remain fail-visible.

## Evidence / reproduction

- Route: `scripts/astrobot/reach-first-level.pad`
- Local failed stage: `/tmp/astro-live-shaders-1074c/exec_ps_50054ab00.bin` (448 bytes)
- The route also reaches `intro_next`; capture remains raw black.
- Hardware Vulkan: AMD Radeon 8060S Graphics, RADV STRIX_HALO.

## Verification

Add focused SPIR-V validation and Vulkan execution coverage that fails without the lowering, including inactive-lane prefix behavior and counter uniqueness. Run full ctest and every title snapshot guard. Live Astro acceptance requires this stage to disappear from failed dumps and recording the next exact frontier; do not claim visual progression from a black/diagnostic capture.

mattias800:
CLAIMING: codex-root-astro-title | branch fix/issue-1076-astro-fragment-wave | 2026-07-20T11:35:00Z

mattias800:
[...]

PR #3199, opened 2026-09-01
recompiler: name the vertex-stage cross-lane MBCNT reject (and record vertex rejects at all)

mattias800:
## The issue, and what the measurement says about it

#3135 records that `v_mbcnt_lo/hi_u32_b32` with a general SGPR mask is fail-closed in a vertex
stage, so "wave-using vertex programs cannot recompile", and asks for a vertex-stage path.

**This PR does not implement that path, because there is nothing exact to implement it onto — and
because implementing it would not rescue the program the issue is about.** Both halves were
measured, not argued. What it does instead is make the failure name itself, and fix the reason it
had nowhere to be recorded.

### The general SGPR-mask vertex path is not sufficient

*Stray* `PPSA02101`, title-screen capsule, vertex `0x300f190000` (426 dwords). With a throwaway
probe arm that satisfies MBCNT in a vertex shell, the live-config retry moves exactly ten
instructions:

| route | reject before the probe | reject with the probe |
| --- | --- | --- |
| `gpu_replay --retry-failed-stage 1:1` | `pc=4` `v_mbcnt_lo_u32_b32 v3, -1, 0` | `pc=14` `ds_write_b32` |
| `gpu_replay --retry-failed-chain 1` | (same MBCNT) | `pc=116` `ds_read2_b32` |

Disassembling the program (`llvm-mc -mcpu=gfx1030`, 288 instructions) says why. It is not a vertex
shader: it is a **merged NGG ES+GS threadgroup** performing vertex/primitive compaction — LDS
traffic at four offsets, three `s_barrier`s, `s_bcnt1_i32_b64` x2, `s_pack_ll_b32_b16`, two
`v_add_nc_u32_dpp row_shr` cross-wave prefix scans, `v_readlane_b32` x2, `s_sendmsg(MSG_GS_ALLOC_REQ)`
and `exp prim`. Per-wave popcounts go to LDS, a DPP scan sums them across waves, and `v_mbcnt`
supplies the within-wave prefix. MBCNT is one of eight wave/workgroup primitives it needs, and it is
the first one only because it is first in program order.

### And it is not implementable in the vertex shell

The vertex shell's guest lane id is **modelled** as the flattened vertex/instance invocation index
(`guest_lane_id()`, `rdna2_to_spirv_internal.hpp`). Nothing in Vulkan promises that the invocations
sharing a subgroup are the `wave_size` consecutive flattened indices that model calls one guest wave,
so a subgroup prefix scan would count over a **different lane set** and return a well-formed wrong
compaction slot — a silent miscompile, not a visible failure. Compute's workgroup-as-wave LDS model
is unavailable too: a Vulkan vertex stage has no shared memory.

The subclasses that *are* exactly lowerable are already covered: `ngg_one_lane` (a wave of one lane,
so any mask contributes zero) and the all-ones lane-index form under `ngg_logical_lane`.
**The gap is a missing wave, not a missing lowering.** The architecturally honest home for a merged
NGG program is a mesh shader — a workgroup is the NGG threadgroup, `s_barrier` is a real barrier,
LDS is shared memory, and `requiredSubgroupSize` is available. That is a frontier, not an opcode add,
so the issue stays open with `Refs`, not `Fixes`.

### Corollary: the reject PC was not the cause, again

The issue and `STRAY_STATUS.md` disagreed about where this program rejects (pc=4 vs pc 277-286).
Measured: **pc=4**, the all-ones form — which is lowerable on its own. The presence of *any*
non-all-ones MBCNT sets `logical_mbcnt_invalid` and disqualifies the whole-program
`ngg_logical_lane` model, so the failure surfaced at an instruction that is not the cause. Third
recorded instance of that shape (#2481, #2790).

## What changed

Three changes, none of which alters what compiles.

1. **`rdna2_emit_alu.cpp`** — a terminal arm for every `V_MBCNT` the four existing lowerings do not
   cover. It records the stage and the mask form (`all-ones` / `sgpr-mask` / `exec` / `vcc` /
   `other-mask`) instead of falling through anonymously.
2. **`rdna2_emit_cfg.cpp`** — a third `mode` on the reject line. `unresolved-operand`'s own contract
   there is "the lowering exists and the descriptor is the defect" (#2412) and `unknown-encoding`
   means "write the emitter"; a decoded cross-lane op in a stage with no wave is neither, and a
   census reads the old line as a resource-table problem. The reason is appended to the **one**
   terminal line rather than logged separately, because a second `role=terminal` line would be
   recorded after it and overwrite it.
3. **`rdna2_to_spirv.{hpp,cpp}` + `gpu_executor.cpp`** — `recompile_vertex` /
   `recompile_vertex_chain` take the diagnostic context the fragment stage received in #3130. The
   vertex stage hardcoded program address 0, and `record_terminal_reject_reason()` early-returns on
   zero, so **no vertex program has ever had a reject reason recorded** and every vertex skip line
   printed `reason=unrecorded`. This is needed for the regression test, and is the same defect
   #3130 fixed one stage over.

Live line on the real program after the change:

```
[recompile-reject] sh=bfa00003/426 mode=unsupported-in-stage
  reason=mbcnt-cross-lane stage=vertex form=all-ones wave=none
  disqualified-by-general-mask-pc=277 pc=4 words=d7650003,000100c1 fmt=9 op=0x365 ...
```

## Behaviour deliberately unaffected

No program that compiled before compiles differently. The new arm catches only MBCNT that already
fell through to the anonymous terminal reject; the four lowering arms are untouched, and
`ngg_logical_lane` / `ngg_one_lane` are unchanged. The general SGPR-mask forms remain fail-visible
in every graphics stage, as #1076 requires.

## Verification

Built in the `ps5ys` distrobox, fresh `-DCMAKE_BUILD_TYPE=Debug` configure, no `-DGAME_DUMP`
override:

```
cmake -S prosper -B prosper/build-linux -G Ninja -DCMAKE_BUILD_TYPE=Debug
cmake --build prosper/build-linux -j8                      # exit 0
ctest --test-dir prosper/build-linux --no-tests=error -j4   # exit 0, 318/318 passed
```

A first `-j4` run reported `lanekill_tool` failed; it passes in isolation and on a repeat full run,
and it is a process-enumeration test on a machine running several concurrent agents. Both runs are
reported rather than only the green one.

### Mutation verification

Seven new assertions in `test_rdna2_to_spirv.cpp`, all reading the recorded terminal reason for a
synthetic reduction of Stray's program. Baseline `TEST_RC=0`, 0 failures, 1096 checks. Every mutant
**compiled** (`BUILD_RC=0`) before being run, so none is a false pass from a stale binary.

| mutation | build | test | arms that failed |
| --- | --- | --- | --- |
| baseline | 0 | 0 | none |
| `b.stage_reject_pc = in.pc` → `UINT32_MAX` | 0 | 1 | 3 of 7 (every `reason=` arm) |
| drop the `vertex_general_mask_mbcnt_pc` derivation | 0 | 1 | exactly 1 — "names the general-mask pc" |
| `b.diagnostic = diagnostic` → `{Vertex, 0}` | 0 | 1 | 4 of 7, including "recorded against the program address" |
| restored | 0 | 0 | none |

The second mutant is the discriminator that matters: it kills only the derived field and leaves the
two control arms green, so the assertion cannot be satisfied by appending a constant string. Arm 2
of the test (an all-ones-only program, where there *is* no disqualifying instruction) asserts the
field is **absent**, for the same reason.

Snapshots were not run: another lane holds the GPU, and this change cannot alter a rendered frame —
no program's SPIR-V changes.

## Docs

- `RECOMPILER_REMAINING.md` § Ruled out — the full argument, so nobody re-derives it.
- `STRAY_STATUS.md` — corrects the reject PC row and the "wave-semantics-in-vertex frontier"
  paragraph, and adds a § Ruled out entry.
- `BLOG.md` — one short entry; no picture, because the shader still does not compile and a black
  frame is not progression evidence.

Refs #3135, #3126, #1076


mattias800 (commented):
**APPROVED**

**The premise was measured, not argued, and it is partly wrong — which is the result here.** Using
`Refs #3135` rather than `Fixes`, and leaving the issue open, is the correct call: the PR argues the
vertex-stage MBCNT path **should not be implemented as posed**.

**Why a vertex MBCNT path would not rescue the shader.** With a throwaway probe arm satisfying MBCNT,
the live-config reject moves exactly ten instructions — to `ds_write_b32` on one route and
`ds_read2_b32` on the other. Stray's `0x300f190000` is a **merged NGG ES+GS threadgroup**, not a
vertex shader: LDS at four offsets, three `s_barrier`s, `s_bcnt1_i32_b64` ×2, two cross-wave
`v_add_nc_u32_dpp row_shr` scans, `v_readlane_b32` ×2, `s_sendmsg(MSG_GS_ALLOC_REQ)`, `exp prim`.
MBCNT is 1 of 8 wave/workgroup primitives and only first *in program order*. Moving a reject ten
instructions is the cheapest possible refutation of "just add the lowering", and it is the one that
would not have been run if the issue had simply been implemented against.

**And there is no exact lowering to add.** `guest_lane_id()` models the vertex guest lane as the
flattened invocation index, but Vulkan never promises a subgroup *is* that lane set — so a subgroup
prefix scan returns a **well-formed wrong** compaction slot, which is worse than a reject. Compute's
workgroup-as-wave LDS model is unavailable because a Vulkan vertex stage has no shared memory. The
exactly-lowerable subclasses are already covered. **The gap is a missing wave, not a missing
lowering**, and the honest home is a mesh shader — correctly labelled as a design argument from the
Vulkan contract rather than a measurement.

**The latent defect found on the way is worth more than the diagnostic.** The vertex stage hardcoded
program address 0, and `record_terminal_reject_reason()` early-returns on zero — so **no vertex
program has ever had a reject reason recorded at all.** That is #3130's defect one stage over, and it
means every past vertex-stage investigation was working without the field it needed.

**Third instance of a pattern now worth naming**: a reject PC names where a fact was *consumed*, not
where it was *lost*. The live reject is at pc=4, the lowerable all-ones form — `STRAY_STATUS.md`'s
row was wrong and the issue body was right. Correcting the doc against your own convenience is the
harder direction.

**Choosing `unsupported-in-stage` over reusing `unresolved-operand`** is right and subtle:
`unresolved-operand`'s contract there is "the descriptor is the defect" (#2412), which is the exact
opposite of the truth here. A reused mode would have sent the next reader to the wrong file.

**Mutation evidence is thorough and one arm is unusually well chosen**: dropping the
`vertex_general_mask_mbcnt_pc` derivation fails **exactly one** arm while the controls stay green,
which proves the field is *derived* rather than appended. All mutants compiled first (`BUILD_RC=0`),
so no false passes.

Verified: `ctest --no-tests=error` **318/318**, exit 0; CI 9 pass / 2 skipping; no `Claude-Session:`
trailer, no absolute host paths. The `lanekill_tool` failure you reported is a process-enumeration
test under concurrent agents — plausible here, and you reported both runs rather than the convenient
one.

On merge I resolved a conflict where this branch and #3191 had each written a **different**
`## Ruled out` entry for Stray's vertex stages. Both are kept — taking either side wholesale would
have deleted a falsification that nothing would ever have flagged.

Issue #2149, opened 2026-08-06
Systemic: diagnostics whose producer and printer are armed by different switches report unarmed state as a confident zero (3 instances in one session)


mattias800:
Three independent instances of the same defect shape landed in a single session on 2026-08-06. That is enough to treat it as systemic rather than as three bugs.

## The shape

**A diagnostic prints a field whose producer is armed by a different switch than the one that arms the printer. When the producer is unarmed the field reads as a confident, well-formed zero — indistinguishable from a measured negative.**

The three instances:

1. **#2111** — `PROSPER_WRITER_PROVENANCE=1` armed every query and left the ColorTarget recorder behind an unrelated variable. The instrument emitted **183 negatives it had no capacity to falsify**. Recorded as instrument trap 116, and hit again a day later by an agent who had read that document.
2. **#2146** — `[udtail]`, the one diagnostic that answers the #305 user-data-window question, is inside the `PROSPER_UD_TAIL_ALIGN` block, alongside the behaviour change `range_start = prefix`. The charter mandates that switch stay **off**, so the report is only available on runs the project disqualifies. Its gate also admits `0` and `off`, so anyone setting it to `off` expecting only the report got the mutation too.
3. **`fresh_extent`** (this session, on #2132) — silently `0` unless `PROSPER_UDPROV` is set. Every earlier arm printed `fresh_extent=0`, which reads as *"the bind programmed nothing"* when the truth was *"nobody was counting"*.

In all three the zero is not merely uninformative — it is **affirmatively misleading**, because a zero in these fields is exactly what the interesting hypothesis predicts.

## Why the existing remedy is not enough

Instrument trap 116 already documents the mechanism, and its remedy is a discipline: check what the instrument can observe before quoting an absence. Trap 121 records the case that proves discipline decays — the agent who hit instance 3 had read 116.

**The durable fix is to make the negative carry its own falsifiability**, so that a reader cannot mistake an unarmed producer for a measured zero without the output telling them.

## Proposal

Not a rewrite — a convention plus a small mechanism, applied to the diagnostics that already have this property:

- **A diagnostic that can print a field it did not measure prints its armed state alongside it.** #2143 does this for `PROSPER_PROVENANCE_ADDR` with a banner naming the armed recorders. That is the right shape and worth generalising.
[...]

PR #3201, opened 2026-09-01
feat(gpu): name dumped shaders by guest address, and filter the dump to one (#3196)

### LINKED ISSUE #3196: PROSPER_SHADER_DUMP_SUCCESS has no address filter, so a program identified by address cannot be dumped
## `PROSPER_SHADER_DUMP_SUCCESS` filenames carry no code address, so a named program cannot be pulled from a run

`PROSPER_SHADER_DUMP_SUCCESS` (`prosper/src/gpu/execute/gpu_executor.cpp:1696`) writes recompiled
shaders named by **content hash**. Nothing in the filename records the guest code address the program
came from, so once you have identified a program *by address* there is no way to ask the dump for it.

This is not hypothetical, and it now blocks the Astro Bot frontier directly.

### Where it bites

1. **#3193** — the Astro Bot world-map GPU hang is narrowed to **one draw**: vertex/ES program
   `0x5008efd00`, pixel program `0x5008f1400`. Declining it (`PROSPER_SKIP_DRAW_PROGRAM`, #3194)
   takes device losses from 37,377 to **0**. The next question is *why that program hangs*, which
   needs its recompiled SPIR-V and its raw RDNA2 — and there is no way to extract "the shader at
   `0x5008efd00`" from a run today.
2. **`live_compute.cpp:4057`** records the same limitation independently, for a different purpose:
   *"the guest ISA for a live program can only be recovered by hash-matching a
   `PROSPER_SHADER_DUMP_SUCCESS` directory, because those filenames carry no code address — so the
   two halves of a divergence could not be compared."*

Two unrelated investigations hitting the same missing field is what makes this worth fixing rather
than working around a third time.

### What is wanted

An **address filter and an address in the filename**, so that a program identified by address —
which is how `PROSPER_SKIP_DRAW_PROGRAM`, `PROSPER_DRAW_PROGRAM_CENSUS` and the `[buf-op]` /
`[mubuf-unresolved]` diagnostics all name programs — can be dumped directly. Roughly:

- include the guest code address in the emitted filename (keep the hash: it is what deduplicates);
- an opt-in filter so a 4K run does not write thousands of modules when one is wanted.

Both halves matter. The filename change alone makes an existing dump searchable; the filter is what
makes it usable on a title that compiles hundreds of programs.

### Notes for whoever takes it

- Dump *both* the recompiled SPIR-V and the raw guest RDNA2 where both are available —
  `tools/shader_inspect` decodes the latter, and a control-flow question needs the guest bytes as
  ground truth rather than prosper's own output.
- The address is already in hand at the emit sites; this is plumbing, not recovery.
- Keep it default-off and self-reporting, matching the contract `PROSPER_COMPUTE_SKIP_PROGRAM` and
  `PROSPER_SKIP_DRAW_PROGRAM` already follow, so a filtered run can never be mistaken for a full one.

Refs #3193, #3194.

mattias800:
## `PROSPER_SHADER_DUMP_SUCCESS` now names its programs by guest address, and `PROSPER_SHADER_DUMP_PROGRAM` narrows it to one

Fixes #3196. Refs #3193, #3194.

### The gap

`PROSPER_SHADER_DUMP_SUCCESS=DIR` wrote every successfully recompiled shader under a **content
hash** and nothing else. A hash names *what* was compiled and never *where the guest put it*, so
once an investigation had identified a program **by address** — which is how
`PROSPER_SKIP_DRAW_PROGRAM`, `PROSPER_COMPUTE_SKIP_PROGRAM`, `PROSPER_DRAW_PROGRAM_CENSUS` and the
`[buf-op]` / `[mubuf-unresolved]` lines all name programs — there was no way to ask the dump for it.

Two unrelated investigations hit that independently: #3193 (Astro Bot's world-map GPU hang narrowed
to vertex/ES program `0x5008efd00`, whose recompiled SPIR-V and raw RDNA2 are the next question) and
the standing comment beside `PROSPER_COMPUTELOG_RAW` in `live_compute.cpp`, which recorded the same
limitation for a different purpose.

### The contract

Both halves, because each is insufficient alone.

**1. The address is in the filename.**

```text
success_vs_at_00000005008efd00_35956264829da0c6_62ad8faab4566b7a.bin   raw guest RDNA2
success_vs_at_00000005008efd00_35956264829da0c6_62ad8faab4566b7a.spv   recompiled SPIR-V
```

so recovery is one glob: `ls DIR/success_*_at_00000005008efd00_*`. A chained (NGG) vertex program
additionally writes `…_main_<main-addr>_<main-hash>.bin` for its continuation. The two hashes stay —
they are what deduplicates, and one program compiled against different resource tables legitimately
yields several variants under one address. `0000000000000000` means no address was available at that
emit site, which is written rather than omitted so it can be told from an unusual address.

The guest address also joined the dedup key. Without that, the **first** address to compile a given
body would claim the pair and every later address sharing those bytes would silently write nothing —
the same false negative this change exists to remove, moved one step downstream.

**2. `PROSPER_SHADER_DUMP_PROGRAM=0xADDR[,0xADDR...]`** — opt-in address filter, so a 4K run does not
write thousands of modules when one is wanted. It follows the contract the sibling selectors already
use:

- **Default OFF.** Unset, everything is dumped exactly as before, and the filter is consulted only
  once `PROSPER_SHADER_DUMP_SUCCESS` is set, so a default run never reaches it.
- **Strict `0x`-only parsing**, through the shared `gpu/diagnostics/watch_list.hpp` parser. A bare
  decimal, a stray or trailing comma, trailing junk, an overflow or a zero address arms **nothing**.
  (`strtoull(spec, &end, 0)` is the trap: base 0 reads `5008efd00` as decimal 5,008 and the run then
  reports itself armed on an address nobody asked about.)
- **Self-reporting**, so a filtered run cannot later be read as a full one:

```text
[shader-dump] PROSPER_SHADER_DUMP_PROGRAM=0x5008efd00 -> armed on 1 program address(es); ONLY shaders at those addresses are dumped
[shader-dump] vs addr=0x5008efd00 chain=0x0 spv=… raw=… main=… words=…+…/… result=written
[shader-dump] withheld addr=0x5008be700 chain=0x0 (not named by PROSPER_SHADER_DUMP_PROGRAM) opportunities=4
```

- Naming **either half of a vertex chain** selects the pair, matching what `gpu_capture` already does
  when it resolves a wanted code address.

### One deliberate asymmetry with the skip selectors

A malformed spec **fails open**: it arms nothing, says so loudly, and the run dumps everything.
`PROSPER_SKIP_DRAW_PROGRAM` and `PROSPER_COMPUTE_SKIP_PROGRAM` fail closed because a skip that
declines nothing is harmless — but a *dump* that writes nothing leaves an empty directory, and an
empty directory reads as *"that program was never compiled"*: a false negative wearing the clothes of
evidence. The address is in the filename now, so failing open costs volume and loses no information.

The filter is also consulted **before** the dedup set, so a withheld program is never recorded as
dumped — otherwise re-arming the filter would silently write nothing. The consequence is stated in
the log and the header: `opportunities=N` counts dump opportunities, not distinct programs.

### Two-gate self-check

`PROSPER_SHADER_DUMP_PROGRAM` is a modifier of `PROSPER_SHADER_DUMP_SUCCESS`, not a switch of its
own, so arming only the one whose *name* matches the question would produce an empty directory and
not one line — #2149's two-gate shape exactly. Setting it without a dump directory now prints, once:

```text
[shader-dump] PROSPER_SHADER_DUMP_PROGRAM is set but PROSPER_SHADER_DUMP_SUCCESS=DIR is NOT -- nothing will be dumped. The filter narrows a dump; it does not start one
```

### Deliberately unaffected

- Nothing changes on a default run: with `PROSPER_SHADER_DUMP_SUCCESS` unset the dump function still
  returns at the same early exit, and the new work sits behind it apart from one function-local
  static guard load.
- The shader cache key is untouched. The address is diagnostic provenance and travels as a separate
  argument, exactly as #3130 established — folding it into `ShaderCompileKey` would give each address
  its own cache entry and quietly defeat the cache.
- `PROSPER_SHADER_DUMP` (the *failed*-recompile dump) is unchanged.

### Verification

Built and tested in the `ps5ys` distrobox, fresh build dir, `-DCMAKE_BUILD_TYPE=Debug`, dump at
`PPSA24651`:

```bash
cmake -S prosper -B prosper/build-linux -G Ninja -DCMAKE_BUILD_TYPE=Debug
TMPDIR=$PWD/prosper/build-linux/tmpdir cmake --build prosper/build-linux -j6     # rc 0, 0 errors
ctest --test-dir prosper/build-linux --no-tests=error -j4                        # rc 0
```

**`ctest`: 323 tests discovered, 323 passed, exit code 0**, on the merged head (`origin/master`
merged in after #3160/#3189/#3194/#3200 landed; the pre-merge head was 321/321 the same way).

One honest note on the first merged-head run: at `-j4` it reported 3 failures —
`vertex_fetch_render`, `indexed_render`, `gpu_execute`. All three are Vulkan-execution tests, all
three pass on `--rerun-failed`, and the machine was at load average ~30 with several concurrent
lanes. Re-run at `-j2`: **323/323, exit 0**. Recorded rather than quietly re-run, because "it passed
the second time" is exactly the shape that hides a real intermittent — the discriminator here is that
the three are GPU-device tests and nothing in this diff touches a rendering path.

New coverage:

- `tests/gpu/diagnostics/test_shader_dump_filter.cpp` (`shader_dump_filter`, 12 arms) — the
  parse/match half with no GPU, no environment and no filesystem: unset vs malformed, strict
  rejection of eight malformed specs, either-half chain matching, and the withholding ordinal's
  print schedule.
- `tests/gpu/execute/test_shader_recompile_cache.cpp` — four end-to-end arms. The copies are **heap
  vectors, not second static arrays**: identical `.rodata` objects can be folded to one symbol by the
  linker, which would silently give two "different" programs the same address — an instrument that
  cannot fail for the reason it is testing.

**Mutation-verified, three arms, each with the build's exit code checked first** (a mutation that
breaks compilation leaves the old binary and reports a false pass):

| mutation | build | result |
| --- | --- | --- |
| `allows()` returns true unconditionally | rc 0 | `shader_dump_filter` 4 failures, `shader_recompile_cache` 1 failure ("withholds a program it does not name") |
| filename writes `0` in place of `program_address` | rc 0 | `shader_recompile_cache` 3 failures, including "a dumped shader is named by its guest code address" |
| dedup key drops the address (pre-#3196 4-tuple) | rc 0 | `shader_recompile_cache` 3 failures |

Reverted, rebuilt (rc 0), green again. The mutation arms were run on the pre-merge head; the merge
that followed touches only `BLOG.md`, `CMakeLists.txt` and two `AGENTS.md` files on this branch's
side (four append-vs-append conflicts, both sides kept — `git diff origin/master...HEAD` over those
four paths is insertions only, no deletions, per instrument trap 41).

Other gates run locally: `python3 prosper/tools/env/check_diag_gates.py` — all checks passed, 108
findings, 108 baselined, none new. `check_numbered_table.py` over the touched `.md` files — clean.
`git diff --check` — clean.

No snapshots were run (this is a diagnostic-only change; nothing on a default rendering path moves).

### Pointed at the motivating case

Run on the real title after the tests, to check the tool answers the question it was built for.
`tools/screenshot`, `PPSA21564-app0`, native, no pad script,
`PROSPER_RENDER=1 PROSPER_GUEST_ARGS=-force-gfx-direct`, this branch:

```sh
PROSPER_SHADER_DUMP_SUCCESS=<dir> PROSPER_SHADER_DUMP_PROGRAM=0x5008efd00 \
  ./prosper/build-linux/screenshot <DUMP_ROOT>/PPSA21564-app0 --out <out> \
    --count 2 --seconds 30 --warmup-seconds 100 --timeout 260
```

#3193's hanging vertex/ES program came out, and nothing else did:

```text
[shader-dump] PROSPER_SHADER_DUMP_PROGRAM=0x5008efd00 -> armed on 1 program address(es); ONLY shaders at those addresses are dumped
[shader-dump] withheld addr=0x50740a700 chain=0x0 (not named by PROSPER_SHADER_DUMP_PROGRAM) opportunities=1
…
[shader-dump] vs addr=0x5008efd00 chain=0x0 spv=0918e0e29ad5a58c raw=01ef30f57b0d07bc main=0000000000000000 words=413+0/12162 result=written
```

**Four SPIR-V variants over one raw RDNA2 stream** — `md5sum` on the four `.bin` files gives one
distinct content, which is exactly the "one program, several resource tables" shape the filename
format was designed to keep legible:

```text
success_vs_at_00000005008efd00_0918e0e29ad5a58c_01ef30f57b0d07bc.bin   1,652 B
success_vs_at_00000005008efd00_0918e0e29ad5a58c_01ef30f57b0d07bc.spv  48,648 B
success_vs_at_00000005008efd00_4bda84cc7c795cdb_01ef30f57b0d07bc.{bin,spv}
success_vs_at_00000005008efd00_b2a69909e3d1f9d3_01ef30f57b0d07bc.{bin,spv}
success_vs_at_00000005008efd00_d2367af1bbee96b2_01ef30f57b0d07bc.{bin,spv}
```

Both halves are usable as they stand: `shader_inspect` decodes the raw stream (413 dwords, 298
instructions, one `endpgm`, 4 generically-unsupported SOP2 rows to look at), and all four `.spv`
files pass `spirv-val --target-env vulkan1.2`. The artifacts are title-derived and stay local.

That is the tool working, not an answer to #3193 — *why* `0x5008efd00` hangs is still open, and the
bytes to answer it with are now one command away.


mattias800 (commented):
**APPROVED**

**Mutation verified independently.** Forcing `ShaderDumpProgramFilter::allows()` to return `true` —
so the filter admits every program — fails **both** test binaries (`rc=8`) on a clean build
(`build_rc=0`, 0 errors); restored, `rc=0`. Baseline `ctest --no-tests=error` **323/323**, exit 0.

*(My first attempt at this mutation used a regex that missed the `ShaderDumpProgramFilter::`
qualifier, so nothing was applied — and it still reported `build_rc=0 / ctest_rc=0 / 0 failing arms`,
which reads as "the test is not load-bearing" when it actually meant "I re-ran the baseline". The
numbers above are from the corrected run anchored to the exact line. Recording it because a false
negative here argues for deleting a guard that works.)*

**The gap this closes was blocking a live frontier, and it was blocking two.** `PROSPER_SHADER_DUMP_SUCCESS`
named dumps by content hash with no code address, so a program identified *by address* — which is how
`PROSPER_SKIP_DRAW_PROGRAM`, `PROSPER_DRAW_PROGRAM_CENSUS` and the `[mubuf-unresolved]` diagnostics
all name programs — could not be pulled out of a run at all. `live_compute.cpp:4057` recorded the same
missing field independently for divergence comparison. Two unrelated investigations blocked on one
absent field is what made this worth building rather than working around a third time.

**Putting the address in the dedup key, not just the filename, is the part that would have been easy
to miss.** Without it the first address to compile a given body claims the pair for every later
address sharing those bytes — the dump would look complete and be wrong.

**Failing OPEN on a malformed spec is the right call and the opposite of the skip selectors.** Those
fail closed because a wrongly-armed skip silently changes what renders. Here an empty dump directory
would read as *"that program was never compiled"* — a false negative about existence, which is worse
than dumping too much. The asymmetry is deliberate and documented, and the extra warning when the
filter is set without a dump directory covers the other half of the two-gate shape.

**The duplicate-programs-as-heap-vectors detail is a real trap avoided:** identical `.rodata` can be
linker-folded, which would have given two "different" programs one address and made the test assert
something that is only true by accident.

## The result it already produced

Armed at `0x5008efd00` on a routed run: **four SPIR-V variants over one raw RDNA2 stream**, and
nothing else. `shader_inspect` decodes it cleanly — 413 dwords, 298 instructions, one `endpgm`,
`consumed=413`. **All four `.spv` pass `spirv-val --target-env vulkan1.2`, so whatever hangs the GPU
is not malformed SPIR-V.** That eliminates a whole class of cause on #3193 in the first run of the
tool, which is the best possible argument that it was worth building. The four SOP2 rows flagged as
generically unsupported are correctly labelled candidates only — a raw dump has no descriptor table.

**Stale claims created by this change were corrected in the same PR** rather than left to rot: the
`PROSPER_COMPUTELOG_RAW` comment in `live_compute.cpp`, its row in `GTA5_STATUS.md`, and
`tools/shader_inspect/README.md` all said those filenames carry no code address. Finding and fixing
the documentation your own change falsifies is the step that usually gets skipped.

**Honest gap accepted as stated:** the `_main_` chain-address path is not exercised by a live title
(Astro's program has `chain=0x0`) and is covered only by the unit test's either-half matching. And
the 3 Vulkan-device failures on the first `-j4` run — `vertex_fetch_render`, `indexed_render`,
`gpu_execute` — were reported rather than quietly re-run; they pass on `--rerun-failed` and match the
signature of instrument trap **243** under load ~30.

CI 9 pass / 2 skipping. `check_diag_gates.py` all-pass, numbered-table gate clean, `diff --check`
clean, no `Claude-Session:` trailers.

Issue #3204, opened 2026-09-01
execute_item is 5,236 lines: extract testable predicates as they are touched


mattias800:
## `execute_item` is 5,236 lines, and that is why #3204 shipped

`prosper/frontends/shared/live/live_compute.cpp` is **11,012 lines**, and one function —
`execute_item`, lines **4,997–10,232** — is **5,236 of them**.

This is filed as a consequence rather than as an aesthetic complaint, because the function's size is
what let a real rendering regression through a three-round adversarial review.

### The concrete cost, measured

#3048 appended six equality checks to `same_view`, the predicate deciding whether two image
descriptors alias one Vulkan image. Five of them compared mip-chain provenance that is populated on
one construction path and left unset on another, so identical single-level descriptors stopped
aliasing. GTA V's 4K output shader writes four 8×8 quadrants of each 16×16 block through four
bindings at one address; three of four stores were lost and a checkerboard covered the frame.
**33,615 aliasing rejections across 46 addresses in under a minute, every one `declared_mip_levels
== 1` on both sides.**

`same_view` is a **local expression inside `execute_item`**. It therefore had:

- no unit test, and no way to write one;
- no name, so no reviewer could ask "what is this predicate's contract?";
- no reachable seam, so the only way to exercise it was to boot a title and look at the screen.

The review was not lax — it ran adversarial mutation testing, caught a separate silent
positional-init defect, and added a compile-time guard making that defect class unrepresentable.
None of that could see this, because **there was nothing to point a test at.** #3204 fixes the bug by
extracting one predicate; the other several dozen decisions in that function remain equally untestable.

### What is wanted

Not a rewrite. Extract *pure, named, testable predicates* from `execute_item` as they are touched —
the pattern already used by `compute_guest_ranges_overlap` (#3157), `sem_poll_step` (#3069) and
`shader_resource_mip_chain_provenance_matches` (#3204). Each extraction converts an invisible
decision into one a test can pin and a reviewer can name.

Candidates visible from the #3204 work, in rough order of risk:

- `same_view` / `same_sampler` / `same_host_backing` — image identity and aliasing. The one that
  just cost a regression.
- the storage-cache gate inputs and `repeated_output` decision.
- the seed/writeback aliasing decision.
- the per-binding decline paths (`skip_image` and friends), which currently fail visible only if you
  are reading stderr.

The charter already asks for this ("split a file that has grown a second responsibility... you do not
[...]

PR #3209, opened 2026-09-02
compute: collapse the atomic-image extent ternaries and name the LOGICAL/PHYSICAL split

### LINKED ISSUE #3195: live_compute: a ternary selects between two byte-identical calls — dead condition, or a dropped atomic_slice_bytes extent?
`prosper/frontends/shared/live/live_compute.cpp:5633-5636` (on `origin/master`) selects between two
**byte-identical** expressions:

```cpp
if (buffers[i].alias_of == SIZE_MAX) {
    const uint8_t* source = buffers[i].atomic_image
        ? resource_bytes_for(resource, buffers[i].guest_bytes)
        : resource_bytes_for(resource, buffers[i].guest_bytes);
```

The condition cannot affect the result, so one of two things is true and they need different fixes:

1. **The condition is dead** and should be deleted, leaving a plain call. Harmless today, but a
   ternary whose arms are identical reads as though a distinction is being made, so the next person
   to touch it has to re-derive that there isn't one.
2. **An extent was dropped.** `BoundBuffer` carries `atomic_slice_bytes` — "physical guest bytes PER
   LAYER (tiled slices are padded)" — at `live_compute.cpp:2941`, alongside `atomic_layers` (#2265).
   An `atomic_image` binding is an R32_UINT StorageImage exposed as a linear atomic SSBO, and the
   shape of this ternary is exactly what you would write if the `atomic_image` arm were meant to ask
   for a different byte count and then didn't. If so, this reads the wrong extent for that path.

Case 2 is the reason this is filed rather than tidied: the two possibilities are indistinguishable
by inspection, and only one of them is cosmetic.

## Deciding between them

`git log -L 5633,5636:prosper/frontends/shared/live/live_compute.cpp` should show whether the arms
were ever different. If the ternary was introduced already-identical, case 1 is likely; if the arms
diverged and were later collapsed, case 2 needs checking against #2265.

## Provenance

Found by independent review of PR #3160 as an incidental observation. It is **pre-existing on
`origin/master`** and unrelated to that PR, which is why it is filed separately rather than fixed
there.

`CONFIDENCE: HIGH` that the branches are identical (byte-compared). `CONFIDENCE: LOW` on which case
applies.

mattias800:
Fixes #3195

## Failure scenario

`execute_item` picked the guest-side byte extent for a storage-buffer binding through a ternary whose
two arms are the same expression, at two sites — the upload source pointer and the write-back
destination:

```cpp
const uint8_t* source = buffers[i].atomic_image
    ? resource_bytes_for(resource, buffers[i].guest_bytes)
    : resource_bytes_for(resource, buffers[i].guest_bytes);
```

The issue asks whether that is a dead condition to delete, or an atomic arm that lost its
`atomic_slice_bytes` extent and is now silently under-bounding the guest read. Those have opposite
fixes, and the second is a live correctness bug: the atomic-image path detiles per layer at the
*physical* slice stride, so a bound computed from the *logical* `w*h*4` would let the detile source
walk past the region the readability probe proved readable.

## The answer: the condition is dead, and it became dead from the other direction

`git log -L` on both line ranges — the second site is not in the issue, it has the identical shape at
what is now `live_compute.cpp:9403` — gives the same two-commit history for each:

1. **`d7ccfa07` "gpu: execute padded tiled image atomics"** introduced both ternaries with genuinely
   different arms: the bounded `resource_bytes_for(r, guest_bytes)` for atomic images against the
   **unbounded** `resource_bytes(r)` otherwise.
2. **`94d36b03` "gpu: honor scalar buffer descriptor bounds"** changed the **`else`** arm of both to
   `resource_bytes_for(r, guest_bytes)`, so that ordinary storage buffers also honour the
   authenticated descriptor bound.

So the arms converged because the *non-atomic* arm was widened upward. **The atomic arm has never
been edited since it was written**, which falsifies the "dropped extent" reading directly rather than
by absence of evidence.

The condition is dead in the strong, structural sense, not merely "no caller reaches the
discriminating state": both arms are the *same expression* in the same scope, and `resource_bytes_for`
is a pure captureless lambda reading only `host_data`, `host_data_size` and `gpu_addr`. No reachable
state can separate them, so no reachability argument is needed.

What makes the atomic special case redundant is upstream: `buffers[i].guest_bytes` is **reassigned**
to the physical `atomic_slice_bytes * atomic_layers` inside the `atomic_image` block, before either
ternary runs. Every later guest-side question about that binding — the source bound, the write-back
notification — is therefore already asking with the physical extent.

## Behavioural contract

Unchanged. A storage-buffer binding is staged from, and written back to, `guest_bytes` bytes of guest
backing: the physical padded footprint for an atomic image, the materialization's logical size
otherwise. This PR changes no value the emulator computes.

## Approach

1. **Collapse both ternaries** to the plain call, each with a comment recording the history above so
   the next reader does not re-derive it.
2. **Extract the extent computation** into `src/gpu/resources/atomic_image_staging.hpp`, beside its
   siblings `shader_resource_supports_atomic_image_buffer` / `shader_resource_atomic_image_layers`,
   following the precedent set by #3204 (`image_identity.hpp`, `spirv_storage_match.hpp`). The
   invariant that made the ternary redundant — that `guest_bytes` carries the **physical** extent —
   was an unnamed local expression no test could reach, which is exactly why the question could not
   be settled by inspection. `atomic_image_staging_extents` returns the layer count, the physical
   per-layer slice stride, the physical guest footprint, the logical staging size, and whether the
   layout is expressible at all.
3. **Add `tests/gpu/agc/test_atomic_image_staging.cpp`** pinning that split, plus an `AGENTS.md` entry
   and a `## Ruled out` row in `docs/RESOURCE_BINDING.md`.

The extraction is a pure move: the same expressions, evaluated in the same order, called only where
the old block ran. In particular `atomic_slice_bytes` is still assigned before the validity check
(both paths `break` out of the buffer loop, so the value is unread either way, but the order is
preserved rather than reasoned about), `bytes` is still assigned after both checks, and
`atomic_layers` relies on `BoundBuffer`'s `= 1` default for the non-atomic case —
`std::vector<BoundBuffer> buffers(N)` value-initializes, and each entry is written exactly once.
`tiled_surface_bytes` is still called only for atomic-image bindings, not for every binding.

## Deliberately unaffected

- Ordinary (non-atomic-image) storage buffers: same extent, same call, same bound.
- The detile/retile transforms themselves, the aliasing comparison, the persistent-cache candidacy
  test, and every skip reason string.
- `resource_bytes` (the unbounded lambda) keeps its one remaining caller.
- The unrelated identical-looking ternary shapes elsewhere in the file were not touched.

## Risks

The only risk is that the extraction is not the pure move it claims to be. Mitigations: the diff is
expression-for-expression reviewable; the new test pins the exact byte values on both the tiled and
the linear path; and the full suite is green. The change is semantically null by construction, so
there is no behavioural regression for a snapshot to catch — which is itself the finding, and the
reason the regression test targets the *invariant* rather than the deleted condition.

## Build and test

Configured and built in the project container, `cmake -S prosper -B prosper/build-linux`:

```
CFG_RC=0
BUILD_RC=0
```

```
ctest --no-tests=error -j4
100% tests passed out of 330
CTEST_RC=0
```

New test registered by name (not inferred from the total), in the platform-independent
`if(TARGET prosper_core)` block:

```
ctest -N | grep atomic_image_staging
  Test #254: atomic_image_staging
```

All 24 arms pass on the fix.

## Mutation arms — the test fails without the invariant it pins

Both mutations **compiled and ran** (`BUILD_RC=0`), so the reddening is the assertion failing, not
the build:

**Mutation A — the "dropped extent" shape.** `slice_bytes` computed as the logical `tight_pitch *
height` instead of the physical padded footprint, i.e. exactly what the case-2 reading imagined the
atomic arm should have been asking for:

```
MUT_A_BUILD_RC=0
  [FAIL] physical slice is padded to whole micro-tiles (2160 rows -> 2176)
  [FAIL] the guest bound STRICTLY exceeds the logical extent (245,760 bytes of padding)
  [FAIL] and by exactly the measured padding
  [FAIL] the slice stride is per-layer and unchanged by layering
  [FAIL] the guest footprint is layers * slice stride
  [FAIL] the physical slice walks the DECLARED row pitch
  [FAIL] padded linear is physically larger too
FAILED (7)
MUT_A_TEST_RC=1
```

**Mutation B — layer count dropped.** `guest_bytes = slice_bytes` instead of `slice_bytes * layers`:

```
MUT_B_BUILD_RC=0
  [FAIL] the guest footprint is layers * slice stride
  [FAIL] the bound ends exactly where the last layer's detile source ends
FAILED (2)
MUT_B_TEST_RC=1
```

The header was restored from a pristine copy after each arm and the full suite re-run green.

## Ruled out

`docs/RESOURCE_BINDING.md` gains one row: the "atomic arm lost its `atomic_slice_bytes` extent"
hypothesis, falsified by the line's own history, with both commits named.

## Not run

No emulator, snapshot or GPU-capture run: the GPU is claimed by another lane, and this change is
semantically null, so a live route could not distinguish it from master in any case.

## A note on this box's render tests under load

The 330/330 above was measured twice, and between the two an intervening run of the same build
directory failed `multidraw_render` and `recompiled_shaders_render` (3 of 3, deterministically)
while another lane held the GPU at 100%. Neither test links anything this PR touches, and both went
green again on the same binaries once that load eased, with no rebuild in between. Recorded so the
next reader does not spend the time I did deciding whether those two were real: on this box, a red
render test during concurrent GPU work measures the box.


mattias800 (commented):
**APPROVED.**

Reviewed the archaeology, the extraction's faithfulness, and the test's counter-arms independently
of the author's evidence.

**The dead-condition verdict is right, and for the reason given.** The two arms are the same
expression in the same scope over a pure captureless lambda, so this is structural deadness rather
than a reachability argument — the strong form, and the only form that would be safe to act on
here. The history check is the part that answers the issue's actual question: the arms converged
because the *non-atomic* arm was widened to the bounded call, and the atomic arm has never been
edited. That rules out the "dropped `atomic_slice_bytes` extent" reading by showing the extent was
never there to drop, which is a different and much stronger claim than "I could not find a caller
that separates them".

**The extraction is behaviour-preserving, including the one place it looked like it might not
be.** The old code assigned `buffers[i].atomic_layers` on *both* paths (`: 1u` for non-atomic);
the new code assigns it only inside `if (atomic_image)`. That is safe because `BoundBuffer` carries
`uint32_t atomic_layers = 1` as an NSDMI (`live_compute.cpp:2951`) and `buffers` is constructed as
`std::vector<BoundBuffer> buffers(buffer_plan.total_descriptor_count)` (`:5460`), so every element
is value-initialized to exactly the value the old branch wrote. `atomic_slice_bytes` was likewise
never assigned on the non-atomic path before. The alias comparison at `:5690-5691`, which reads
both fields for buffers the loop has already passed, therefore sees identical values. No stale
locals survive — `linear_image_bytes` and `guest_image_bytes` are both gone from the translation
unit.

`atomic_image_staging_extents` reproduces every term of the original validity condition, including
the `linear_bytes > UINT32_MAX` bound and the narrow-row-pitch rejection, and is only ever reached
for a resource that has already passed `shader_resource_supports_atomic_image_buffer`.

**The test's discriminating arm is the right one.** `guest_bytes > linear_bytes`, and by exactly
245,760 bytes, is the assertion a helper returning the logical extent could not satisfy — i.e. it
fails on precisely the shape the issue's case-2 reading imagined. The reported mutation arms (7 and
2 failures, both with a clean build exit) confirm the arms are load-bearing rather than incidentally
true.

**Why the extraction was worth doing rather than just deleting the ternary.** The invariant that
makes the condition dead — that `guest_bytes` already holds the physical extent by the time the
source pointer is taken — was an unnamed local expression, which is exactly why the question could
not be settled by reading the code. Naming it means a future "restoration" of the distinction
reddens instead of silently under-bounding the readability probe. That is the same reasoning as
#3204's extractions and it applies more strongly here, because the failure mode of getting it wrong
is a walk past the region proven readable rather than a visible error.

`## Ruled out` row recorded in `RESOURCE_BINDING.md`; `AGENTS.md` updated. No host paths, no session
trailers. CI green on every job except macOS/Rosetta, in flight at review time.

Issue #3210, opened 2026-09-02
render_runner: vkCreateRenderPass/vkCreateFramebuffer write UNINITIALIZED handles that are then used unchecked, and image views/samplers reach descriptor sets with no guard


mattias800:
Found while fixing #3180 (the six color-target `vkCreateImage` sites). Those six were the mild
class: each discarded the `VkResult` but still guarded the resulting handle, and each handle was
provably `VK_NULL_HANDLE` going in. The sites below are a **different and worse** class, and they
are not `vkCreateImage`, which is why #3180 does not cover them.

All line numbers are `prosper/tests/fixtures/render_runner.h` on `origin/master` @ `c1987cdf`. Note
this header is not test-only: `frontends/shared/live/live_renderer.cpp` calls into it (e.g.
`prosper::test::persistent_color_target_limit()` at `live_renderer.cpp:4282`), so these are live
render paths.

## 1. Two handles are UNINITIALIZED, unchecked, and used unconditionally

```cpp
5406:    VkRenderPass rp; vkCreateRenderPass(dev, &rpci, nullptr, &rp);
5414:    VkFramebuffer fb; vkCreateFramebuffer(dev, &fbci, nullptr, &fb);
```

Neither declares an initializer, and neither result is read. Vulkan only writes the output
parameter on success, so on failure `rp` and `fb` hold an **indeterminate value** — this is not the
"drivers conventionally leave it null" reliance that #3180 and #3045 describe, it is reading an
uninitialized automatic variable. Both are then used with no guard at all:

- `rp` -> `fbci.renderPass` (5412), every `VkGraphicsPipelineCreateInfo::renderPass` (7511),
  `rpbi.renderPass` (8365), and `vkDestroyRenderPass` (9464).
- `fb` -> `rpbi.framebuffer` (8365), i.e. straight into `vkCmdBeginRenderPass`.

A garbage `VkRenderPass` reaching `vkDestroyRenderPass` is the worst of these: destroying a
non-handle is undefined, and on a 64-bit build the value is whatever was on the stack.

## 2. Image views and samplers written into descriptor sets with no check at all

```cpp
5259:        vkCreateImageView(dev, &dvci, nullptr, &dview);          // depth/stencil view
7296:                                vkCreateImageView(dev, &tvci, nullptr, &binding.view);
7297:                                vkCreateSampler(dev, &sci, nullptr, &binding.sampler);
7307:                            vkCreateImageView(dev, &tvci, nullptr, &binding.view);
[...]

PR #3211, opened 2026-09-02
render_runner: report the VkResult at all six color-target vkCreateImage sites

### LINKED ISSUE #3180: render_runner: several vkCreateImage sites discard VkResult (checked, but not logged)

mattias800:
Fixes #3180

## Failure scenario

`render_draw_pass_rgba` creates its color attachments at six sites — a primary and a
budget-miss-fallback create for slot 0, for slot 1, and for the MRT slots 2..7 — and every one of
them discarded the `VkResult`:

```cpp
vkCreateImage(dev, &imgci, nullptr, &img);
if (!img) {
    if (cached_color) persistent_color_target_cache().erase(color_key);
    return out;
}
```

Unlike the two sites #3045 fixed, each of these already guarded the resulting **handle** before
touching it further, so none fed a definitely-null image into a Vulkan call. Two smaller problems
remained:

1. **The guard rested on a driver convention, not the spec.** Every implementation leaves `pImage`
   at `VK_NULL_HANDLE` when `vkCreateImage` fails, but Vulkan does not require it — a conforming
   driver that left the slot untouched would walk straight past `if (!img)`. (The handles here are
   all provably null going in, so today this is a latent dependency rather than a live bug.)
2. **A failure was silent.** The pass returned an empty frame with nothing in the run log, so "the
   device refused this attachment" and "the draw list was empty" produced identical evidence, and
   the real `VkResult` — the single fact that separates them — was thrown away.

## Behavioural contract

A color-target creation failure at any of the six sites drops the pass exactly where it already did,
and now also reports one rate-limited `stderr` line naming the site, the real `VkResult`, and the
requested extent/format/usage. Success is entirely unchanged.

## Approach

Route all six through one helper:

```cpp
inline VkResult create_color_target_image(VkDevice dev, const VkImageCreateInfo& ci,
                                          RenderColorTargetCreateSite site, VkImage* out_image);
```

It tests the **result first and the handle second**, forces the handle to `VK_NULL_HANDLE` on
failure so the callers' existing guards stay correct whatever the driver left behind, and emits a
rate-limited report following the `[texture-upload-failed]` / `[ds-create-failed]` convention #3045
established:

```
[color-target-create-failed] site=slot1-fallback vkCreateImage result=-2 extent=64x64 fmt=37 usage=0x16 -- dropping the pass
```

The rate limit is a **single counter shared by all six sites**, so a device failing every target
cannot flood a run log.

## Deliberately unaffected

- **Control flow.** Every site still returns from the pass at exactly the same point, and each
  failure path's `persistent_color_target_cache().erase(...)` is untouched. The diff replaces a
  create-plus-handle-check with a create-plus-result-check; it does not move, add, or remove a
  `return`.
- The depth/stencil site (`[ds-create-failed]`) and the sampled-texture upload site
  (`[texture-upload-failed]`), both already fixed by #3045.
- `vkCreateImageView` / `vkCreateSampler` / `vkCreateRenderPass` / `vkCreateFramebuffer` — see the
  finding below.
- The persistent-target budget, eviction, and caching logic.

## A bigger finding, filed separately: #3210

Auditing the neighbouring sites for the same shape turned up a **worse** class, which this PR
deliberately does not touch:

```cpp
5406:    VkRenderPass rp; vkCreateRenderPass(dev, &rpci, nullptr, &rp);
5414:    VkFramebuffer fb; vkCreateFramebuffer(dev, &fbci, nullptr, &fb);
```

Both handles are **uninitialized**, both results are discarded, and both are used with **no guard at
all** — `rp` reaches every pipeline's `renderPass`, `rpbi.renderPass`, and `vkDestroyRenderPass`;
`fb` reaches `vkCmdBeginRenderPass`. On failure Vulkan does not write the output parameter, so this
is reading an indeterminate automatic variable, not the null-handle reliance described above.
Separately, five `vkCreateImageView`/`vkCreateSampler` results (5259, 7296/7297, 7307/7309) are
unchecked and a possibly-null view goes into a `VkDescriptorImageInfo` and then
`vkUpdateDescriptorSets` — the exact #3045 shape, one API along, and two of them cache the null
into `persistent_image->second.bindings` so one failure is re-served to later draws.

These need **new failure paths** rather than a report added to control flow that already exists, and
they span four Vulkan entry points, so folding them in here would make this PR neither mechanical
nor reviewable. Filed as **#3210** with exact line numbers.

## Risks

The helper is the only new logic and it is nine lines. The residual risk is that a site's guard
changed meaning; the diff shows each replacement is `create → check handle` becoming
`create → check result (handle forced null on failure)`, with the surrounding block byte-identical.
The injector is off unless a test arms it and is `thread_local`, one-shot, and site-selective, so it
cannot leak into the production path or between tests.

## Build and test

```
cmake -S prosper -B prosper/build-linux      CFG_RC=0
cmake --build prosper/build-linux -j6        BUILD_RC=0
```

New test registered **by name** (not inferred from a total), in the same `if(Vulkan_FOUND)` block as
the other `render_runner.h` tests:

```
ctest -N | grep color_target_create_failure
  Test #311: color_target_create_failure
```

```
./test_color_target_create_failure
  ... 42 arms ...
PASSED
TEST_RC=0
```

```
ctest --no-tests=error
100% tests passed out of 330
CTEST_RC=0
```

All six sites are driven by a **real render**, each preceded by an uninjected control run of the
identical configuration that must render a full 64x64x4 frame and log nothing — without that
control, an arm that reached the site by accident, or that printed from some *other* site, would be
indistinguishable from one that worked.

Reaching the three fallback sites needs `PROSPER_BACKEND_TARGET_CACHE_MB=0`, set before the first
render because `persistent_color_target_limit()` memoizes its environment; a zero budget makes every
persistent target miss and take its fallback create.

## Mutation arms — the test fails without the fix

Both mutations **compiled and ran** (`BUILD_RC=0`), so the reddening is assertions failing, not the
build:

**Mutation A — the pre-fix behaviour.** The `fprintf` removed, i.e. fail silently:

```
MUT_A_BUILD_RC=0
18 [FAIL] lines  (3 log assertions x 6 sites)
MUT_A_TEST_RC=1
```

**Mutation B — a constant site name** (`"slot0"` instead of `render_color_target_site_name(site)`).
This is the arm that matters most, because it is the one a test could pass for the wrong reason:

```
MUT_B_BUILD_RC=0
  [FAIL] slot0-fallback: the failure names ITS OWN site
  [FAIL] slot1: the failure names ITS OWN site
  [FAIL] slot1-fallback: the failure names ITS OWN site
  [FAIL] slot-extra: the failure names ITS OWN site
  [FAIL] slot-extra-fallback: the failure names ITS OWN site
MUT_B_TEST_RC=1
```

Exactly the five non-slot0 sites go red and slot0 stays green — which is what proves each site is
independently wired rather than one site reporting for all six.

The fixture was restored from a pristine copy after each arm.

## Ruled out

No entry: this issue's premise was confirmed rather than falsified — the six sites do rely on the
null-handle convention and do fail silently. The audit's negative result (that the *neighbouring*
sites are worse, not equivalent) is recorded in #3210, which is where the next reader will look.

## Not run

No emulator, snapshot or GPU capture: the GPU is claimed by another lane.

## A note on this box's render tests under load

An earlier run of this suite reported 5 failures under `-j4` and 2 serially
(`recompiled_shaders_render`, `multidraw_render`) while another lane held the GPU at 100%. **They
are not from this change**, and the control is on the record rather than asserted: the identical
`multidraw_render` failure reproduced 3 of 3 on a *different* worktree whose `render_runner.h` is
untouched `origin/master`. Once the other lane's load eased, both trees ran 330/330 green with no
rebuild in between. Quote the green run; the red one measured the box, not the diff.

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -2689,6 +2689,15 @@ if(TARGET prosper_core)
       target_link_libraries(test_multidraw_render PRIVATE prosper_core Vulkan::Vulkan)
       add_test(NAME multidraw_render COMMAND test_multidraw_render)
 
+      # #3180: a failed color-target vkCreateImage must name its site and the real VkResult. Same
+      # block as the other render_runner.h tests -- it needs the `tests` and `frontends` include
+      # roots and a real Vulkan device.
+      add_executable(test_color_target_create_failure tests/gpu/test_color_target_create_failure.cpp)
+      target_include_directories(test_color_target_create_failure PRIVATE tests)
+      target_include_directories(test_color_target_create_failure PRIVATE frontends)
+      target_link_libraries(test_color_target_create_failure PRIVATE prosper_core Vulkan::Vulkan)
+      add_test(NAME color_target_create_failure COMMAND test_color_target_create_failure)
+
       # Indexed draws (#64): vkCmdDrawIndexed from a BackendDraw's index data — the [0,1,2,2,3,0] quad
       # must match the retired perimeter-fan rendering pixel-exact, and indices must really be applied.
       add_executable(test_indexed_render tests/gpu/state/test_indexed_render.cpp)
--- a/prosper/tests/fixtures/render_runner.h
+++ b/prosper/tests/fixtures/render_runner.h
@@ -2673,6 +2673,86 @@ inline bool consume_render_texture_create_failure_once() {
     return true;
 }
 
+// #3180: the COLOR-TARGET vkCreateImage sites, the milder siblings of #3045's two.
+//
+// Each of the six already guarded the HANDLE before touching it further (`if (!img) return out;`),
+// so unlike #3045 none of them fed a definitely-null image into a Vulkan call. What they did do is
+// discard the VkResult, which left two smaller problems:
+//
+//   * the guard rested on a driver CONVENTION, not on the spec. Every implementation leaves
+//     `pImage` at VK_NULL_HANDLE when vkCreateImage fails, but Vulkan does not require it, so a
+//     conforming driver that left the slot untouched would walk straight past `if (!img)`. Testing
+//     the result FIRST and the handle second removes that dependency at no cost.
+//   * a failure was silent. The pass returned an empty frame with nothing in the run log, so
+//     "the device refused a 4K attachment" and "the draw list was empty" produced the same
+//     evidence, and the real VkResult -- the one fact that separates them -- was thrown away.
+//
+// The control flow is deliberately unchanged: every site still returns from the pass exactly where
+// it did. This adds a report, and one rate-limited counter shared by all six so a device failing
+// every target cannot flood a run log.
+enum class RenderColorTargetCreateSite : uint32_t {
+    None = 0,
+    Slot0,              // primary slot-0 (the pass's own color target)
+    Slot0Fallback,      // slot-0 retry as a transient image after a persistent-cache budget miss
+    Slot1,              // primary slot-1 (MRT1's independent guest identity)
+    Slot1Fallback,      // slot-1's budget-miss retry
+    SlotExtra,          // primary MRT slots 2..7
+    SlotExtraFallback,  // an MRT slot's budget-miss retry
+};
+
+inline const char* render_color_target_site_name(RenderColorTargetCreateSite site) {
+    switch (site) {
+        case RenderColorTargetCreateSite::Slot0:             return "slot0";
+        case RenderColorTargetCreateSite::Slot0Fallback:     return "slot0-fallback";
+        case RenderColorTargetCreateSite::Slot1:             return "slot1";
+        case RenderColorTargetCreateSite::Slot1Fallback:     return "slot1-fallback";
+        case RenderColorTargetCreateSite::SlotExtra:         return "slot-extra";
+        case RenderColorTargetCreateSite::SlotExtraFallback: return "slot-extra-fallback";
+        default:                                             return "none";
+    }
+}
+
+// Deterministic one-shot injection, the site-selective twin of the sampled-texture hook above and
+// for the same reason: these requests are ordinary 2D color attachments at the caller's own extent,
+// so no real device can be made to refuse one on demand. Arming a site makes the NEXT create at
+// exactly that site report VK_ERROR_OUT_OF_DEVICE_MEMORY without calling the driver. The production
+// path never arms this state.
+inline RenderColorTargetCreateSite& render_color_target_create_failure_storage() {
+    static thread_local RenderColorTargetCreateSite armed = RenderColorTargetCreateSite::None;
+    return armed;
+}
+
+inline void inject_render_color_target_create_failure_once(RenderColorTargetCreateSite site) {
+    render_color_target_create_failure_storage() = site;
+}
+
+inline bool consume_render_color_target_create_failure(RenderColorTargetCreateSite site) {
+    RenderColorTargetCreateSite& armed = render_color_target_create_failure_storage();
+    if (armed != site || site == RenderColorTargetCreateSite::None) return false;
+    armed = RenderColorTargetCreateSite::None;
+    return true;
+}
+
+// Create a color target, reporting the driver's own verdict. Returns VK_SUCCESS with a non-null
+// `out_image`, or a failure code with `out_image` forced to VK_NULL_HANDLE so the caller's existing
+// handle guard stays correct whatever the driver left behind.
+inline VkResult create_color_target_image(VkDevice dev, const VkImageCreateInfo& ci,
+                                          RenderColorTargetCreateSite site, VkImage* out_image) {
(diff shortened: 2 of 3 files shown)


mattias800 (commented):
**APPROVED.**

**Control flow is genuinely unchanged where it matters, and strictly safer where it changes.** Each
site's `vkCreateImage(...); if (!img) { ...; return out; }` becomes
`if (create_color_target_image(...) != VK_SUCCESS) { ...; return out; }`, with the same
cache-erase and the same return point. The helper forces `*out_image` to null on any failure, so
the callers' existing handle discipline holds regardless of what the driver left behind.

The one behavioural difference is in the right direction and worth naming explicitly, because it is
not obvious from the diff: the old code proceeded whenever the handle happened to be non-null,
*including* when `vkCreateImage` had returned a failure code. Vulkan does not require a failing
implementation to write `VK_NULL_HANDLE`, so the old guard rested on a convention rather than the
spec. Testing the result first removes that dependency. Coercing a `VK_SUCCESS`-with-null-handle
return to `VK_ERROR_INITIALIZATION_FAILED` closes the mirror case.

**The test asserts the thing that was actually broken.** It says so itself, which is the part I
looked hardest at: "the 'pass is dropped' assertions alone would pass on unfixed code, which is
exactly why they are not the contract this test exists for." That is the trap this project keeps
falling into — a regression test with another path to the expected result — and the test is built
to avoid it rather than merely surviving it. Each arm requires the log to name its own site and the
real `VkResult`.

**The constant-site-name mutation is the arm that earns the approval.** Reddening exactly the five
non-slot0 sites while slot0 stays green is what distinguishes "six sites are independently wired"
from "one site reports and the others are along for the ride" — a distinction the 18-failure
report-removal arm cannot make on its own. Both mutations built cleanly, so they ran.

Injection rather than provocation is right here for the reason #3045 recorded: these are ordinary
2D color attachments at the test's own extent, so no real device can be made to refuse one on
demand. Worth noting for whoever touches this next that the arming state is `thread_local`, so it
only fires when the pass runs on the arming thread — true in this test, and not something to rely
on if these arms are ever reused from a job thread.

**On the concurrent-load failures**: the control is on the record rather than asserted — the same
`multidraw_render` failure reproducing 3 of 3 on a worktree whose `render_runner.h` is untouched
master is what makes it a statement about the box rather than about this diff. That is the correct
shape for that claim. Yes, file the instrument-trap row.

Also confirmed independently: the branch carries no host paths and no session trailers, and the
force-pushed history is clean. **#3210 is a real finding** — `render_runner.h:5406` and `:5414` on
master read `VkRenderPass rp;` and `VkFramebuffer fb;` with no initializer, discard the result, and
use the handles unguarded, while the `mkmod` lambda four lines below correctly writes
`VkShaderModule m = VK_NULL_HANDLE`. Keeping it out of this PR was the right call: it needs new
failure paths across four entry points and would have made this diff neither mechanical nor
reviewable.

CI green on every job except macOS/Rosetta, in flight at review time.

Issue #3205, opened 2026-09-01
Mip-chain provenance broke image aliasing: checkerboard grid over GTA V (regression from #3197)


mattias800:
## Mip-chain provenance broke image aliasing, rendering a checkerboard over GTA V

#3048 (PR #3197) appended six equality checks to `same_view` in
`prosper/frontends/shared/live/live_compute.cpp` — the predicate deciding whether two image
descriptors alias **one** Vulkan image. Five compare `mip_chain_*` provenance.

That provenance is **not populated uniformly**: `image_base_level_view` fills it in, while
`unmapped_format_image_view` deliberately leaves it at zero (*"no allocation-wide mip placement --
zero element extent = not modelled"*). So two **identical single-level** descriptors for the same
guest address compare unequal based purely on which construction path built them.

## Failure scenario

The consequence was already documented in the comment immediately above the changed lines:

> Two exact image descriptors for the same non-null guest range must still alias one Vulkan image:
> GTA V's 4K output shader writes the four 8x8 quadrants of each 16x16 block through four bindings
> at the same address. Treating the reconstructed blobs as four images loses three stores and leaves
> the red/green checker in Performance mode.

Which is exactly what happened: a grid over the entire screen whenever the in-game minimap is
visible.

## Evidence

Bisected on `PPSA04263` across the 32 commits merged 2026-09-01. Clean at `6f9e14ff`, present at
`cff3abe9`. Neutralising **only** the six added conditions removes the grid; forcing both mip level
counts to 1 does **not** (the predicate is outside the mip machinery entirely, which is why four
earlier experiments showed no change).

Instrumented run on the unfixed commit:

- **33,615** aliasing rejections across **46** distinct addresses in under a minute
- **100%** had `declared_mip_levels == 1` on both sides — not one involved a chain
- representative line: `addr=0x2046ce0000 decl=1/1 ew=0/2560 eh=0/1440 bpb=0/4 maxl=0/0 basel=0/0`

## Fix

Keep `declared_mip_levels` in the comparison (genuinely part of identity); gate the five provenance
fields on a chain actually being declared. #3048's intent is preserved exactly — descriptors that
would materialize *different chains* still refuse to alias.

The rule is extracted as a named, pure predicate so it can be tested at all, which it could not be
before: it was a local expression inside a 5,236-line function (#3204).

## Why review did not catch it

The review of #3197 was thorough — adversarial mutation testing, a separate silent positional-init
[...]

PR #3212, opened 2026-09-02
gpu_replay: a capture owns a mip chain's whole allocation, so replay materializes it

### LINKED ISSUE #3202: gpu_replay cannot materialize a declared mip chain: captures own the selected level, not the allocation
Split out of the #3197 review of #3048.

#3048 made the **live** compute backend materialize a guest texture's declared mip chain, uploading
every level from its own `tiled_mip_level_layout` offset, so `IMAGE_LOAD_MIP` lowers to
`OpImageFetch` with the guest's runtime LOD. `gpu_replay` does **not** get that, and the v57 capture
tail does not change it — which matters because the F9 bundle is the project's normal way to study a
frame like this offline.

## Why replay still declines

Two independent reasons, both fail-closed:

* **A replayed resource is `host_data`-backed.** `deserialize` binds every resource's bytes out of a
  capture blob (`gpu_capture.cpp`, the `bind_range` path that sets `r.host_data`/`r.host_data_size`).
  `shader_resource_compute_mip_chain_levels` returns 1 for any `host_data` resource, deliberately:
  the allocation base sits *below* `gpu_addr` for a tiled chain (the shared mip-tail block is stored
  first), and a host blob anchored at `gpu_addr` cannot expose it.
* **The blob only covers the selected level.** `resource_footprint_impl` starts from `r.size`, which
  is the selected level's footprint. The other levels' bytes are simply not in the file.

So a capsule taken on a title that exercises this replays with a single-level image and the
recompiler declines exactly as it did before — same as today, no silent wrong content, but no
offline reproduction either.

## What it would take

The capture side has to own the **allocation**, not the selected level:

1. Extend the footprint for a resource whose chain is placeable to `tiled_mip_chain_bytes(...)`,
   anchored at `gpu_addr - levels[0].byte_offset` rather than at `gpu_addr`.
2. Give the replay binding a way to express "this blob starts below the descriptor's address", since
   `bind_range`'s current contract is that the blob offset does not exceed the logical address.
3. Then relax the `host_data` decline to "declined unless the host span covers the whole chain".

v57 already records the descriptor half (`mip_chain_element_width/height`,
`mip_chain_bytes_per_block`, `mip_chain_max_level`, `mip_chain_base_level`), so the placement is not
lost from capsules taken in the meantime — only the bytes are missing.

## Do not

Do not work around this by generating levels 1..N-1 on replay. `IMAGE_LOAD_MIP` asks for a specific
level's texels; synthesizing them renders plausible wrong content, which is the failure mode
`SONIC_FRONTIERS_STATUS.md` § "Where the guest's mip levels actually are" exists to prevent.

Refs #3048, #3197, #2790

mattias800:
Fixes #3202. Refs #3048, #3197, #2790.

## The gap

#3048 taught the **live** compute backend to materialize a guest texture's T#-declared mip chain,
uploading every level from its own `tiled_mip_level_layout` offset, so `IMAGE_LOAD_MIP` lowers to
`OpImageFetch` with the guest's runtime LOD. `gpu_replay` did not get that, and the v57 tail did not
change it — which matters because the F9 bundle is how a frame like this is normally studied
offline.

The reason is one fact about GFX10 tiled layout: **a mip chain stores level zero LAST.** The shared
mip-tail block comes first, then the remaining levels smallest-to-largest, so a descriptor selecting
level zero names the **highest** address in its own allocation and every other level lives *below*
it. Two consequences, both fail-closed:

* the captured range is anchored at `gpu_addr` and sized from the selected level's footprint, so the
  other levels' bytes were never in the file;
* a replayed resource's `host_data` points at `gpu_addr`, and `shader_resource_compute_mip_chain_levels`
  therefore declined **any** `host_data` resource — a blob anchored at `gpu_addr` cannot expose the
  bytes underneath it.

## Behavioural contract

* A capture now owns the **allocation** for a resource whose chain the compute backend would
  materialize, anchored at `gpu_addr - levels[0].byte_offset`.
* Replay publishes how many owned bytes precede `gpu_addr` as
  `ShaderResource::host_data_prefix_bytes`, and the chain derivation accepts a `host_data` resource
  **only** when its span covers the whole allocation — the prefix below `gpu_addr` and the selected
  level above it.
* **A capsule that cannot express the chain keeps declining, visibly.** Every capsule taken before
  this change has a blob offset that does not reach the allocation base, so it replays exactly as it
  did: a single-level image, `IMAGE_LOAD_MIP` refused, no synthesized levels. That is deliberate and
  is the issue's explicit "do not" — generating levels 1..N-1 on replay would render plausible wrong
  content, the failure mode `SONIC_FRONTIERS_STATUS.md` § *Where the guest's mip levels actually are*
  exists to prevent. Re-grab the frame instead.
* A `gpu_replay --override-resource` replacement is a self-contained snapshot of the selected level,
  so it resets the prefix to zero and its resource declines. Same for a compute-internal (GDS /
  packed-pointer) snapshot and for any synthetic host fixture.
* Live behaviour is unchanged: a guest-memory-backed resource has no prefix and is bounded by
  `guest_readable` over the same span, exactly as before.

## No new capture-format version, and why

The obvious reading of #3202 is that the reader needs something new in the file. It does not, and
this is recorded as a `## Ruled out` row because it is the non-obvious half.

`collect_intervals` already anchors a resource's captured range at an interval whose `begin` may sit
below `gpu_addr` (ranges are merged), and `assign_blob_range` already serializes
`blob_offset = gpu_addr - interval.begin`. A blob's byte *i* is the guest byte at
`blob.guest_addr + i` by construction — ranges are merged and read contiguously, and blob dedup only
shares byte-identical content — so `blob_offset` **is** the count of owned guest bytes preceding the
descriptor's address. Only the writer had to change. Nothing new is written to the file, `kVersion`
stays 57, and pre-#3202 capsules fail closed on the same existing number without a version gate.

## Approach and invariants

1. **One derivation, still.** `mip_chain_plan.cpp` grows a private `compute_chain_plan()` holding
   everything that decides whether a chain is materializable *except* where the bytes come from.
   Three callers read it: the level count (recompiler + backend, the #2265 shape), the new
   `shader_resource_compute_mip_chain_allocation()` the capture writer sizes from, and
   `shader_resource_host_data_covers_mip_chain()`. The backend's pre-dereference bounds check calls
   that last function rather than restating the rule, so the count and the bound cannot drift.
2. **`ShaderResource::host_data_prefix_bytes`.** Appended after `host_data_size`; the invariant is
   `host_data - k` is the guest byte at `gpu_addr - k` for k in `[1, host_data_prefix_bytes]`. Every
   site that repoints `host_data` at a different buffer resets it (replay's own reset, the two
   `--override-resource` sites, both descriptor-array entry-resource builders, and `capture_table`'s
   clear). Inserted into a struct with no positional aggregate initializer past its first eight
   members — the #3197 hazard was checked, not assumed.
3. **A capture never fails because of this.** The extension can only push the total over the
   resource budget, so `capture_submit_items` retries `collect_intervals` without it and reports
   *that* failure — the error a user would have seen before this change. The per-resource ceiling,
   the anchor (`prefix <= gpu_addr`) and the overflow bound are all checked, and a resource that
   fails any of them is simply not extended.

## Deliberately unaffected

* The graphics texture path (still one level, still the #1272 blit cascade for its own chains).
* Live guest-backed compute chains: byte-identical behaviour, same `guest_readable` bound.
* The capture format: no version bump, no new field, old capsules load unchanged.
* The size of a capture that has no materializable chain: the writer extends nothing there, and a
  registered arm asserts it (a converting-format control captures only its own level).
* `IMAGE_LOAD_MIP` on the graphics path, and every reject-by-default arm from #3048 (linear chains,
  tail-selected levels, layered/volume views, DCC, block-compressed and converting formats, a
  shifted `BASE_LEVEL`).

## Risks

* **Capture size.** A capsule containing a materializable chain grows by that allocation's non-zero
  levels — about a third of the selected level for a full pyramid. Bounded per resource by
  `PROSPER_CAPTURE_BLOB_MAX_MB` and in total by `PROSPER_GPU_CAPTURE_MAX_MB`, and the retry above
  means the worst case is the old capture, not a failed one.
* **A short guest read still reads as zeros**, as it always has: `bytes_read` records it and nothing
  on the replay path consults it. The prefix is at the *start* of the blob, so it is the part most
  likely to have been read, but a partially-read allocation would materialize zeroed levels rather
  than declining. Pre-existing for level zero; not made worse, not fixed here.
* No title was launched and no snapshot was run (another lane holds the GPU), so this is verified by
  the registered suite only. The claim is about the capture/replay path, not about what Sonic
  Frontiers renders.

## Verification

Own worktree, own build dir, `ps5ys` distrobox, Ninja + Debug, `Vulkan found — offscreen graphics
harness enabled`.

```
cmake --build prosper/build-linux -j6   -> exit 0, 0 errors
ctest --no-tests=error                  -> exit 0, 100% tests passed out of 329
```

**Five Vulkan execution cases are flaky on this host under load, and it is not this change.**
`recompiled_shaders_render`, `multidraw_render`, `indexed_render`, `descriptor_array_render` and
`gpu_execute` failed on two of four full runs — both taken with a host load average of 12-15 while
other lanes were linking — and passed on the two taken at load 2-6, including the final one quoted
above. `gpu_execute` alone passes 3 of 3, and its failing arms are `execute_and_present` /
scanout-publication assertions on the graphics path, which this diff does not touch: the capture
extension is gated on a compute-materializable chain (tiled, `declared_mip_levels >= 2`, native
straight-copy format), `resource_footprint` is unchanged, and the identity field these tests carry is
zero on both sides. Stated rather than hidden, per the instrument-trap rule.

Registered coverage (both arms verified present by name with `ctest -N`):

* `gpu_capture_roundtrip` — the whole path end to end: the captured range starts at the allocation
  base and spans the chain; the serialized blob offset equals the bytes below `gpu_addr`; after
  serialize → deserialize → `materialize_gpu_replay` the replayed allocation is **byte-identical**
  to the guest's (position-dependent fill, so a right answer read at a wrong offset fails) and each
  planned level's own bytes are at its own offset; the level count is the full chain. Then the
  discriminators: one byte short below `gpu_addr`, one byte short above it, and a zero prefix (a
  pre-#3202 capsule) each decline back to one level. Plus a control proving the writer does *not*
  extend for a chain the derivation declines.
* `shader_resource_contract` — the coverage rule built **by hand**, outside the capture path that
  produces it in practice, so the positive arm cannot inherit that path's assumptions: a host span
  covering the allocation materializes the chain, one byte short in either direction declines, the
  allocation span the writer sizes from is the plan's own span, and a declined chain zeroes both
  outputs. The pre-existing "owns only its selected level → declined" arm is unchanged and still
  passes.

**Mutation verification.** Three mutations, each rebuilt (build exit code checked before believing
any result) and re-run over the three affected cases.

| arm | build | ctest | result |
| --- | --- | --- | --- |
| baseline | 0 | 0 | 3/3 pass; all 22 new arms report `[ok]` by name |
| **M1** — writer extension disabled (`false && own_mip_chain_allocations`) | 0 | 8 | `gpu_capture_roundtrip` fails; **7 arms redden**, from "the captured range starts at the ALLOCATION base" through "a replayed resource whose span covers the allocation materializes the whole declared chain". `shader_resource_contract` and `image_identity` stay green — they do not go through the writer, which is the correct split |
| **M2** — the pre-#3202 blanket `if (resource.host_data) return 1u;` restored | 0 | 8 | **exactly the two positive arms redden**, one in each file: "a replayed resource whose span covers the allocation materializes the whole declared chain" and "a host-backed resource whose span covers the whole allocation materializes the declared chain". Every negative arm stays green, which is what makes them negative arms rather than restatements |
| **M3** — `host_data_prefix_bytes` removed from `shader_resource_same_view` | 0 | 8 | `image_identity` fails on exactly "a backing that owns bytes below the pointer is not the same image as one that does not" |
| restored | 0 | 0 | 3/3 pass; `git diff` contains no `MUTANT` marker |

One honest note on M1: the arm *"one byte short BELOW gpu_addr declines"* also reddens under it, but
for a derived reason — it mutates the replayed resource, and under M1 that resource's prefix is
already zero, so `-= 1` wraps. It is a real discriminator against the shipped code and not a
discriminator for M1; M2 is the arm that isolates the coverage rule.

No snapshots were run and no title was launched — another lane holds the GPU.


mattias800 (commented):
**APPROVED.**

## The change I checked hardest, and why

This PR adds a condition to `shader_resource_same_view`. Adding a condition to that predicate is
exactly what caused #3205 — #3197 appended six equality checks, five of them comparing mip-chain
provenance populated by one view helper and left zero by another, and identical single-level
descriptors stopped aliasing. That cost a visible rendering regression on *Grand Theft Auto V*
and a hand bisect over 32 commits to find.

So the question is not whether the new condition is *reasonable*, it is whether the field is
populated **uniformly**. It is, and I verified it exhaustively rather than from the comment:
`host_data_prefix_bytes` is declared `= 0` (`shader_resources.hpp:477`), and across all of `src/`
and `frontends/` the only write that is not a literal zero is
`gpu_capture.cpp:5290`'s `bind_range(..., &r.host_data_prefix_bytes)` on the replay
deserialization path. Every other site — the two table-entry constructions, `capture_table`, and
`materialize_gpu_replay`'s reset — writes 0.

On a live boot the field is therefore zero on both sides of every comparison, the new condition is
trivially true, and live aliasing is untouched. That is the property #3205's fields lacked, and the
comment in the diff names the contrast correctly instead of just asserting safety.

## The format-version finding is the best part of this

"Needs a new capture-format version" was the expensive-sounding half of the issue's sketch, and it
was wrong: `assign_blob_range` has always serialized `blob_offset = gpu_addr - interval.begin`,
which *is* the count of owned bytes below the descriptor's address, because a blob's byte *i* is the
guest byte at `blob.guest_addr + i` by construction. The number was already in every file; only the
writer had to start owning the allocation. `kVersion` stays at 57, existing capsules keep working,
and pre-#3202 capsules fail closed on the same number with no version gate.

Recording that as a `## Ruled out` row is right — it is precisely the kind of falsification the next
agent would otherwise re-derive at full cost, and it would have cost a format migration.

The load-bearing layout fact behind it (a tiled chain stores level zero **last**, so `gpu_addr` is
the highest address in its own allocation and the forward footprint was already correct) is also
what kept `resource_footprint` unchanged — which matters more than it looks, since widening it would
have altered pre-v28 capsules through `materialize_gpu_replay`'s `captured_size ? … : footprint`
fallback.

## The rest

`shader_resource_host_data_covers_mip_chain` is a correct coverage test in both directions — prefix
owned *and* forward span covering `allocation_bytes - prefix` — and it proves its own
non-underflow rather than reasoning that the caller cannot hit it. Good instinct on a function whose
plan argument is an input.

`compute_chain_plan` does not weaken the #3197 single-derivation invariant, it strengthens it: the
level count, the writer's allocation span and the backend's bounds check are now one derivation with
three callers, where before the count function inlined its own filters. The cheap rejections stay
first, so the per-dispatch compile-key path is unchanged, and for a live resource (`host_data` null)
the ordering change is a no-op.

The issue's "do not synthesize levels on replay" is honoured — an uncovered capsule declines exactly
as before — and the two problems found while auditing are both real: a `--override-resource`
replacement carrying a stale prefix would authorize a read before its own buffer, and an overridden
binding aliasing the descriptor it shadows would give one of them a level count its compiled module
contradicts, which is the SPIR-V-undefined shape of #3205 again.

**Mutation arms**: M2 is the one that earns this — restoring the blanket `host_data` decline reddens
*exactly the two positive arms, one per file*, and leaves every negative arm green. That is the
split that distinguishes "the relaxation works" from "the test would pass either way". M1's
7-arm/2-file split and M3's single-arm result are consistent with it.

The five flaky Vulkan execution cases are the same set another lane hit on this box today, at load
12–15 and passing at load 2–6, and CI's own Linux job is green over the full set on this head.
Putting that in the PR body rather than omitting it is the right call.

CI 9 SUCCESS / 2 skipped / 0 failures. No session trailers, no host paths.

mattias800 (commented):
**APPROVED** — re-affirmed on head `89cac89f`.

Master moved three times under this branch (#3209, #3211, #3213), so I merged `origin/master` in
rather than leaving it stale. The only conflict was `BLOG.md`: both sides append a newest-first
entry under the same `## 2026-09-02` heading. Both entries are kept — master's Astro Bot entry
stays exactly where master put it and this branch's mip-chain entry follows it, so nothing in
master's copy is disturbed. `live_compute.cpp` auto-merged; I checked by content that both sides
survived it (#3209's `atomic_image_staging_extents` collapse and this branch's
`host_data_prefix_bytes = 0` table-entry reset are both present, and the ternary #3209 removed has
not come back).

A clean merge is not evidence of a working file, so the merged tree was rebuilt and re-run rather
than assumed: build exit 0, `ctest --no-tests=error` **100% of 331 passed, exit 0** — and none of
the five load-sensitive Vulkan execution cases went red on this run.

The approval above stands unchanged: no reviewed code was altered by the merge, only the two blog
entries were reconciled.

PR #3213, opened 2026-09-02
recompiler: make the CFG trip bound usable on a graphics stage, and scope it to one loop (#3193)

mattias800:
## What this is

Astro Bot's world-map GPU reset (#3193) now has a **mechanism**, and the two instrument changes that
were needed to get it. It is not a fix for the hang — the remaining half is #3214.

`Refs #3193` rather than `Fixes`: the reset still happens on a default run. What this PR lands is the
localization, its evidence, and the tooling that makes the next step cheap.

## The failure scenario

Loading Astro Bot's world map takes the whole GPU down — a RADV hard recovery, after which every
frame is the same frozen picture. #3193 narrowed it to one **draw** and named it by its **vertex**
program `0x5008efd00`, stating in its own caveats that a draw-level A/B cannot separate the draw's
two stages.

It is the **pixel** half, `0x5008f1400`, and the mechanism is a **non-terminating loop**:

- `0x5008efd00` cannot hang: **zero `OpLoopMerge`**, 8 basic blocks, and no backward branch in the
  guest's 413-dword stream.
- `0x5008f1400` is 3,210 dwords with **four backward branches**, so both structured loop emitters
  reject it and `emit_cfg_state_machine` lowers it to one SPIR-V loop over an 89-case `OpSwitch` —
  the only loop in a 136,875-word module.
- Capping that back edge at 4,096 takes a deterministic 53-loss control to **0**, with the draw still
  executing 2,048+ times. At 1,000,000 the device dies again. A cap can only change the outcome by
  being reached.
- An ordinal-scoped cap then localizes it to one of the shader's four guest loops: a per-screen-tile
  linked-list walk over light records (guest pc 1911..2538) whose only exit is `link == 0xffffffff`.

Full evidence, the arms and what each could and could not show: `ASTROBOT_LINUX_HANDOFF_2026_07_19.md`.

## The two behavioural changes

Both are to `PROSPER_CFG_TRIP_BOUND`, which is `getenv`-gated and emits nothing when unset — modules
are byte-identical on a default boot.

### 1. The device-side witness is emitted only in compute (a defect, not a refactor)

The witness writes to the internal GDS buffer at descriptor set 0, binding 127.
`kComputeInternalGdsBinding` is bound, prepared, read back and restored **only by the compute
executor**; it appears nowhere else in `src/` or `frontends/` except the compute test fixture. The
selector, however, takes a program *address* and accepts a graphics one.

Armed on a graphics program — which is exactly what #3193 required — the emitted module would declare
a descriptor no graphics pipeline layout contains. **The damage is not the missing record.** A draw
whose pipeline cannot be created never runs; the device then survives, and survival is precisely the
signal this instrument exists to produce. The arm would have read as *"the bound fixed it"* while the
bound never executed.

So: cap the loop in every stage, publish only where a reader exists, and print once per program that
a graphics arm is capped-but-unwitnessed, naming the corroboration a graphics arm needs
(`PROSPER_DRAW_PROGRAM_CENSUS`, to show the draw still ran). Recorded as instrument trap 244.

### 2. `PROSPER_CFG_TRIP_BOUND_ORDINAL=K` — cap ONE loop inside a multi-loop program

The dispatcher emits one back edge for a whole program, so a plain cap answers "does some loop here
run away" and cannot say which. `_ORDINAL=K` counts only the traversals about to re-enter dispatch
ordinal K (read off the ordinal → guest-pc map the arm already prints). Unselected traversals still
run, they just do not advance the counter — which is what makes a **negative** arm mean something: a
cap on a loop that terminates on its own is inert, so an arm that still dies has excluded its loop.

Mixed into `ComputeTripBoundSettings`, and therefore into the shader cache key, for the same reason
the other two selectors are: the key is otherwise blind to a program's address, so a target and a
non-target with identical bodies would share one compiled module.

## Invariants and what is deliberately unaffected

- **Unset, nothing changes.** No new `getenv` on any per-draw or per-dispatch path; the selector is
  read during recompilation only, as before.
- **Compute behaviour of the witness is unchanged** when a bound is armed: same five dwords, same
  device-scope atomics, same host prepare/read/restore.
- **The bound is still a diagnostic**, not a fix. It truncates guest control flow and produces wrong
  results by construction; it stays opt-in and still says so when it arms.
- **Structured loop emitters are still not covered.** A null from a structurizer-accepted program
  still means *not measured*.
- No renderer, resource-binding or per-title behaviour is touched.

## Risks

- The graphics arm now emits a real cap where it previously would have emitted an unbound descriptor.
  That is the point, but it means a graphics program under an armed bound produces **wrong pixels** —
  as documented, and as every other use of this switch already does.
- The ordinal comparison adds one `OpIEqual` + one `OpSelect` per back edge **only when the selector
  is set**; with it unset the emitted code is the previous shape.

## Tests

`prosper/tests/shared/diagnostics/test_cfg_trip_bound.cpp` gains three groups, each with a mutation
arm that fails without the change:

1. **Ordinal scope, negative:** an ordinal the program does not have leaves the cap **inert** — the
   run matches the *unbounded* result and no witness hit is recorded. The bound is deliberately
   **1**, so an implementation that ignores the scope truncates on its first back edge and both
   assertions fail. This is the arm that fails without the gate.
2. **Ordinal scope, positive control:** built here rather than drawn from the arm it validates —
   sweep the ordinals the fixture can have and require at least one scoped cap to actually fire
   (witness hit + trip count). Without it, (1) would also pass against an implementation that never
   emits a cap at all.
3. **Stage guard:** the same overlapping-EXEC-loop fixture with a colour export appended, compiled as
   a **fragment** shader with the bound armed, must contain **no** binding-127 witness **and** must
   still compare against the bound. The pair is the assertion — either half alone is satisfiable by
   doing nothing.

Plus `compute_trip_bound_settings()` is asserted to report the new selector, since a selector the
struct under-reports is one the cache key cannot see.

## Build / test

```
cmake -S prosper -B prosper/build-linux -DPROSPER_APP=ON     # rc 0
cmake --build prosper/build-linux -j6                        # rc 0
ctest --no-tests=error -j4                                   # rc 0 -- 100% tests passed out of 330
```

Docs gate on the numbered tables, both forms, exit 0:

```
    --baseline <origin/master copy> prosper/docs/GAME_COMPAT_ORCHESTRATION.md
```

Live arms (Astro Bot `PPSA21564`, `tools/screenshot`, native 3840x2160, no pad script,
`--seconds 30 --count 5 --timeout 170`), losses = `[backend] GRAPHICS submission failed`:

| arm | losses | pixel-distinct | max pixel stale |
| --- | --- | --- | --- |
| control | 53 | 3 / 5 | 60.0 s |
| `_BOUND=1000000` | 52 | 3 / 5 | 60.0 s |
| `_BOUND=4096` | 0 | 5 / 5 | 0.0 s |
| `_BOUND=4096 _ORDINAL=30` | 59 | 2 / 5 | 60.0 s |
| `_BOUND=4096 _ORDINAL=42` | 108 | 1 / 5 | 120.0 s |
| `_BOUND=4096 _ORDINAL=45` | **0** | **5 / 5** | **0.0 s** |
| `_BOUND=4096 _ORDINAL=63` | 64 | 3 / 5 | 60.0 s |

No screenshot is added: with the loop capped the world map still renders only the nebula backdrop
already published for #1459, so there is no new visual progress to show. The `BLOG.md` entry says so
in as many words.

`ctest --no-tests=error` (serial): **330/330 passed, exit 0**, 95.0 s. A first pass with `-j4`
reported five Vulkan-execution failures (`recompiled_shaders_render`, `multidraw_render`,
`indexed_render`, `descriptor_array_render`, `gpu_execute`) — all five in 0.03 s, all five pass on a
serial re-run, and they are the exact set instrument trap 243 names. Quoted here because a reader
should not have to rediscover that these five are the parallel-contention set on this box.

### The tests are mutation-verified, not merely green

| mutation | build rc | ctest rc | `[FAIL]` count | which |
| `if (b.is_compute)` → `if (true)` (witness in graphics again) | 0 | 8 | **1** | the binding-127 stage guard |
| ordinal comparison replaced by a constant (scope ignored) | 0 | 8 | **3** | the inert arm, its witness check, and the two-modules check |

Each reverts exactly one thing and nothing else; the source was restored and the suite re-run green
afterwards.

## Docs

- `ASTROBOT_LINUX_HANDOFF_2026_07_19.md` — the dated result, the arm table, and four new
  `## Ruled out` rows (the vertex program; "not the CFG dispatcher path"; both waterfall loops; the
  counted loop).
- `GAME_COMPAT_ORCHESTRATION.md` — instrument trap 244.
- `src/gpu/recompiler/AGENTS.md` — what the instrument covers, what it does not, and the three things
  about it that are load-bearing.
- `BLOG.md` — one entry, no picture, saying what was found.

Refs #3193, #1809, #1459


mattias800 (commented):
**APPROVED.**

This corrects the headline of the issue it advances, which is the most useful thing a lane can do
and the easiest to get wrong. I verified the load-bearing claims against the source rather than the
prose.

## The claim the whole result rests on, checked first

If the `if (b.is_compute)` guard had scoped the **cap** rather than the **witness**, the entire
dose-response table would be void — the graphics arm would have emitted no bound at all and the
0-loss result would have meant something else entirely. It does not: `apply_trip_bound` computes
`next_trip`, `under_bound` and `hit` unconditionally and returns `b.land(keep_going, under_bound)`
outside the guard. Only the five `compute_gds_atomic_minmax` publications are inside it. The bound
executes in every stage, exactly as the body claims.

That distinction is also what makes trap 244 worth recording. The failure mode it names —
an instrument that, when misapplied, produces the instrument's own success signal — is the sharpest
class in the trap list, because nothing about the run looks wrong. A draw whose pipeline cannot be
created never runs, the device survives, and survival is the signal. Finding that *before* running
the arm is the difference between this result and a published null.

## The localization is sound, and the negatives are what make it so

One positive (ordinal 45 → 0 losses) against three negatives (30 → 59, 42 → 108, 63 → 64) is a real
discriminator here precisely because unselected traversals still run: a cap aimed at a loop that
terminates on its own is inert, so an arm that still dies has genuinely excluded its loop. That
property is designed in rather than assumed, and the PR says so.

The alternative reading — "capping B prevented reaching a runaway in D" — is closed by the direct
arm rather than by argument: capping D itself still gave 64 losses, so D unbounded is not the
runaway. Same for A and C. Elimination plus a direct positive.

`PROSPER_CFG_TRIP_BOUND=1,000,000` still dying while 4,096 survives is the arm I would have asked
for if it were missing. A cap can only change the outcome by being reached, and a cap three orders
of magnitude higher that does *not* change it is the control that distinguishes "the bound bit" from
"arming the bound perturbed the run".

## The mechanism claim in #3214 checks out

`bounded_cbuf_load` (`rdna2_emit_alu.cpp:5425`) ends
`return b.sel(in_bounds, loaded, b.uconst(0u));` — an out-of-range scalar read yields architectural
zero, which the comment above it says is deliberate RDNA2 fidelity. So the observation is exact: for
a walk whose terminator is `0xffffffff`, zero is a *valid link*, never the sentinel, and any
out-of-range or unpopulated read silently converts a bounded walk into an unbounded one. Leaving
*why* the chain never terminates open in #3214 rather than guessing is the right call.

Excluding `0x5008efd00` on structure — zero `OpLoopMerge`, no backward branch in 413 dwords — is a
genuine exclusion rather than a failure to observe, which is what lets the stage correction stand.

## Two notes, neither blocking

**`counts` is a SPIR-V result id, and `counts ? ... : ...` is a host-side test on that id.** It reads
like a test on the *value*. It is correct because ids are allocated from 1, but the safety of the
ordinal filter now rests on `b.ucmp` never returning 0 — and if it ever did, the filter would
silently degrade from "count ordinal K" to "count everything", turning a scoped negative arm into a
global one. That is the failure direction that produces a wrong conclusion rather than a visible
error. A one-line comment naming the invariant, or an assert, would cost nothing.

**The witness's span and its trip count now measure different populations** when an ordinal is set:
`dispatch_min_var`/`dispatch_max_var` still update on every traversal while `trip_var` advances only
on selected ones. Harmless for the graphics case (no witness) and defensible for compute, but a
reader comparing the span against the trip count would be comparing different denominators. Worth a
sentence where the witness fields are documented.

## Verification and hygiene

Serial `ctest --no-tests=error` 330/330, exit 0, with both mutation arms demonstrated at a clean
build exit — reverting the stage guard reddens 1 check, the constant-ordinal mutation reddens 3
including the inert arm. The `-j4` failures being the exact set trap 243 names, all passing
serially, is the right control.

Confirmed independently: no session trailers, no host paths in the diff, and
`check_merge_result.py` from the branch's own worktree reports a clean merge with
`GAME_COMPAT_ORCHESTRATION.md` at 1..244 unique and ascending, no base rows deleted.

`Refs #3193` rather than `Fixes` is correct — the reset still happens on a default run. The
`BLOG.md` entry without a picture is also right: capping the loop yields the #1459 nebula backdrop
and nothing new to show, and saying so is more useful than silence.

PR #3215, opened 2026-09-02
render_runner: check the render pass, framebuffer, view, sampler and shader-module creates (#3210)

### LINKED ISSUE #3210: render_runner: vkCreateRenderPass/vkCreateFramebuffer write UNINITIALIZED handles that are then used unchecked, and image views/samplers reach descriptor sets with no guard

mattias800:
Fixes #3210.

`render_draw_pass_rgba` in `prosper/tests/fixtures/render_runner.h` created five kinds of non-image
Vulkan object without ever reading the result. #3210 is careful about the severity split and this
change keeps it, because the three cases need different fixes and carry different risk.

**`render_runner.h` is not test-only.** `frontends/shared/live/live_renderer.cpp`,
`live_compute.cpp` and `present/present_blit.cpp` all compile it, so every failure path added here
is a live render path. That is now written down in a new `prosper/tests/fixtures/AGENTS.md`.

## The failure scenario, per case

**Case 1 — two handles were UNINITIALIZED, unchecked, and used unconditionally.**
`VkRenderPass rp;` and `VkFramebuffer fb;` were declared with no initializer and their results
discarded. Vulkan writes an output handle only on success, so a failed create left an *indeterminate*
value in them — an arbitrary stack word, not the null the neighbouring sites conventionally get from
every real driver. That word then reached `fbci.renderPass`, every
`VkGraphicsPipelineCreateInfo::renderPass`, `rpbi.renderPass` (i.e. `vkCmdBeginRenderPass`),
`rpbi.framebuffer`, and `vkDestroyRenderPass`. Destroying a non-handle is undefined. These sites had
**no failure path at all**, so one is added.

**Case 2 — image views and samplers reached descriptor sets with no guard.** The depth/stencil
attachment view and the four sampled-texture view/sampler creates were unchecked too. Their handles
are value-initialized, so unlike case 1 the failed value was deterministic — but a null view still
went into `VkDescriptorImageInfo` and on to `vkUpdateDescriptorSets`, and a null `dview` went into
the framebuffer. Two of the four also **cached** the result into `PersistentTextureImage::bindings`,
so one transient failure was retained under that binding key and re-served to every later draw that
matched it.

**Case 3 — `vkCreateShaderModule`'s result was discarded** while `m` was initialized and the caller
already skipped the draw on `!v.vs || !v.fs`. Folded in for consistency; not urgent.

## Behavioural contract

Every one of the five now routes through a helper that:

1. forces its output handle to `VK_NULL_HANDLE` **before** the call and again on failure, so the
   indeterminate read is gone whatever the driver leaves behind;
2. returns a non-`VK_SUCCESS` code whenever the handle is unusable (including the
   spec-non-conforming "success with a null handle", reported as `VK_ERROR_INITIALIZATION_FAILED`);
3. emits one rate-limited `stderr` line naming the site, the API, the real `VkResult` and enough of
   the request to separate an out-of-memory from a rejected description.

This is the convention #3045 established and #3211 extended (`create_color_target_image` was read
first and deliberately matched rather than a second convention invented in the same file). The rate
limit is one counter shared by this whole family and **separate** from the color-target counter, so
exhausting either cannot silence the other.

Failure paths, by case:

| case | site | on failure |
| --- | --- | --- |
| 1 | `vkCreateRenderPass` | drop the pass (`return out`), as the color-target creates already do |
| 1 | `vkCreateFramebuffer` | destroy `rp` explicitly, then drop the pass |
| 2 | `vkCreateImageView` (depth/stencil) | tear the half-created DS target down (image, memory, cache entry), then drop the pass |
| 2 | `vkCreateImageView` / `vkCreateSampler` x4 | destroy whichever half succeeded, then `buffer_resources_ready = false; break;` — skip the draw, the path #3045 established |
| 3 | `vkCreateShaderModule` | report only; the existing `!v.vs \|\| !v.fs` skip is unchanged |

`rp` is destroyed on the framebuffer path specifically because nothing has taken ownership of it
yet — the deferred cleanup lambda that would normally free it is registered much later in the
function. The wider transient state at that point (color images, DS image, their memory) is left
exactly as the pre-existing `return out` paths above leave it; that is unchanged behaviour, not a
new leak.

### The already-cached case

The two persistent texture sites are the ones that retained a failure. **The fix breaks before the
`emplace`, which is the sole writer of `PersistentTextureImage::bindings`** — so no null can enter
that map, and the cache-*hit* path a few lines above therefore needs no guard of its own (a guard
there would be dead code, and would suggest the map can hold nulls when it cannot). Nothing
migrates or repairs pre-existing entries because none can exist: the map is process-local and this
is a header, so a build either has the fix everywhere or nowhere.

## Deliberately unaffected

- **A storage image legitimately gets no sampler.** `VK_DESCRIPTOR_TYPE_STORAGE_IMAGE` ignores it,
  and the transient branch already skipped the create for `r.is_storage_image`. The guard is written
  `r.is_storage_image || create_render_sampler_checked(...)`, so the create is not attempted and its
  absence is not a failure. Treating that null as one would drop **every** storage-image draw.
- Case 3's control flow, and every existing `return out` / persistent-cache-erase on the
  color-target paths.
- The six `vkCreateImage` sites #3211 covered.

## Risk: what a false trigger would cost

Every new failure path fires only when the Vulkan call itself returns non-`VK_SUCCESS` (or returns
success with a null handle, which no conforming driver does). There is no heuristic and no
threshold, so a false trigger requires the driver to report a failure that did not happen. If one
did occur the cost is visible and graded: a case-1 false trigger drops a whole pass — a frame that
would have rendered does not, which on a live title reads as a black or frozen frame; a case-2 false
trigger drops **one draw**, so the pass still composites and a single object goes missing. Both are
loud in the run log, which is the other half of the point: before this change a failure at any of
these sites produced an empty frame or a missing object with **nothing** printed, so "the device
refused this view" and "the draw list was empty" produced identical evidence.

The one behavioural risk worth naming is that case 1 converts undefined behaviour into a *deliberate*
dropped pass. That is strictly better than the alternative — see the M6 mutation below, where
reporting and then continuing past the failure with a null handle segfaults the process on the first
injected render.

## Verification

New ctest case `render_object_create_failure`
(`prosper/tests/gpu/test_render_object_create_failure.cpp`), modelled on #3211's
`test_color_target_create_failure`: a deterministic site-selective one-shot injector, one arm per
site, **each preceded by an uninjected control run of the identical configuration that must render
and log nothing**, plus a selectivity/one-shot block and a storage-image "deliberately unaffected"
arm. 8 sites, 48 assertions.

The failure is injected rather than provoked for the reason #3045 and #3180 both recorded: these are
ordinary render passes, framebuffers, 2D views, trilinear samplers and small SPIR-V modules at the
test's own 64x64 extent, so no real device can be made to refuse one on demand.

**Which assertions are load-bearing.** The **log** assertions are the contract — unfixed code prints
nothing at all, so every one of them goes red. The "pass is dropped" / "draw is skipped" assertions
are load-bearing for cases 1 and 2, which had no failure path whatsoever before this change; the
shader-module arm's skip assertion **would pass on unfixed code** and is stated so that a later
change reporting by *continuing* past the failure is still caught. The test says this in its own
header.

**What the test cannot cover.** It cannot distinguish "the handle was uninitialized" from "the
handle was null". Reading an uninitialized automatic is undefined behaviour with no observable value
an assertion can name, and this box's RADV does leave the output handle null in practice, so an
"unfixed code crashed" arm would be unreproducible and would prove nothing about a driver that
behaves differently. Mutation M5 below measures exactly this. Recorded as a `## Ruled out` entry in
`docs/VERIFICATION.md` beside the existing sentinel section, which is the same class one step out.

### Mutation arms (build exit code first, so it is clear each one compiled and ran)

| arm | mutation | build rc | test rc | result |
| M1 | every report in the new family suppressed | 0 | 1 | **14 assertions red** — every log assertion at all 8 sites |
| M2 | `render_vk_object_site_name` returns a constant `"render-pass"` | 0 | 1 | **exactly the 7 non-render-pass sites red** — each site is independently wired, not one reporting for all |
| M3 | persistent branch emplaces the failed pair before breaking (the pre-#3210 retained-failure shape) | 0 | **139 (SIGSEGV)** | crashes at *"a later draw with the FAILED binding key renders"* — the retained null view is re-served and kills the process |
| M4 | storage-image carve-out removed (`r.is_storage_image \|\|` -> `false \|\|`) | 0 | 1 | **3 assertions red** — the storage-image draw is dropped and the arm is consumed, so that carve-out is genuinely load-bearing |
| M5 | *negative control:* all 5 pre-call `*out = VK_NULL_HANDLE;` **and** both caller-side initializers removed | 0 | **0** | **PASSES — as predicted.** This is the documented ceiling: the test covers the report and the drop, not the initialization |
| M6 | case 1 reports but does not drop the pass | 0 | **139 (SIGSEGV)** | dies at the first injected render — the `return out` is load-bearing, and this is with a *null* handle; the pre-fix indeterminate one is worse |

M2 is the equivalent of the arm #3211 called its strongest, and it lands the same way: exactly the
non-slot-0 (here, non-`render-pass`) sites redden.

### Build and test, exact head (`d9753b43`, rebased onto `162c188b`)

```
distrobox enter ps5ys -- bash -lc "cd <WORKTREE> && mkdir -p prosper/build-linux/tmpdir && \
  PROSPER_GAME_ROOT=<DUMP_ROOT> TMPDIR=\$PWD/prosper/build-linux/tmpdir \
  cmake -S prosper -B prosper/build-linux && \
  TMPDIR=\$PWD/prosper/build-linux/tmpdir cmake --build prosper/build-linux -j6"
# build rc=0

distrobox enter ps5ys -- bash -lc "cd <WORKTREE>/prosper/build-linux && ctest --no-tests=error"
# rc=0 -- 100% tests passed out of 332
#   313/332 Test #313: render_object_create_failure ... Passed  0.05 sec
```

Registration verified by name, not by count (`prosper/CMakeLists.txt` has separate `if(WIN32)` and
`if(UNIX)` Vulkan blocks and a registration in the wrong one is silent):

```
ctest -N | grep render_object_create_failure
#   Test #313: render_object_create_failure
```

Docs gate, green:

```
# rc=0; numbered column 1..245, unique and ascending
```

No snapshots were run: this change adds failure paths that cannot fire on a healthy device, and the
uninjected control run of every arm renders identically to before.

## Ruled out / recorded

- `docs/VERIFICATION.md` — a regression test cannot demonstrate that a handle was *uninitialized*
  rather than null (measured by M5).
- `docs/GAME_COMPAT_ORCHESTRATION.md` instrument trap **245** — a one-shot failure injection aimed
  at a site behind a process-wide cache never fires, and stays armed into the **next** test's control
  run, where it reads as that test being broken. This cost a debugging cycle here: the shader-module
  arm's own control run populated the persistent graphics-pipeline cache, so the injected run never
  called `vkCreateShaderModule`, and the arm fired inside the *following* block's control render.
  The control run that makes an arm trustworthy is the same run that made it unreachable, so adding
  a control does not catch it. The test now disables that cache and says why.
- `prosper/tests/fixtures/AGENTS.md` — new: what the directory holds, and that `render_runner.h` is
  a live render path despite living under `tests/`.


mattias800 (commented):
**APPROVED.**

## Case 1 is fixed in the stronger of the two available ways

Checking the `VkResult` would have been enough to take a failure path, but it would have left the
indeterminate read itself intact on any code path that touched the handle before the check. Nulling
the output **before** the call removes the undefined behaviour regardless of what the driver does,
and the caller-side `= VK_NULL_HANDLE` initializers make it belt-and-braces. That is the right
choice for a defect whose whole character is "the value is not merely wrong, it does not exist".

The `vkDestroyRenderPass` on the framebuffer-failure path is correct and I checked the reasoning
rather than the comment: the early `return out` happens before the deferred cleanup lambda is
registered, so nothing else owns `rp` and this is a destroy rather than a double-free.

## The carve-out is faithful, which is the thing that could have gone wrong quietly

`(r.is_storage_image || create_render_sampler_checked(...) == VK_SUCCESS)` reads like it might be
introducing a new exemption. It is not — the original was literally
`if (!r.is_storage_image) vkCreateSampler(...)`, so a storage image never had a sampler and the
short-circuit preserves that exactly. Requiring one would have dropped every storage-image draw,
and M4 reddening 3 arms shows the carve-out is load-bearing rather than decorative.

## M5 is the best thing in this PR

A mutation arm that **passes**, reported as a passing arm, is worth more here than the six that
fail. Deleting every pre-call nulling and both initializers leaves the suite fully green, which
establishes — rather than assumes — that a test cannot distinguish "uninitialized" from "null" on
this box. The honest conclusion is that the initialization half is covered by reading the code, and
recording that as a `## Ruled out` entry in `VERIFICATION.md` beside the sentinel section is exactly
where it belongs: the next agent who wonders why there is no coverage for it gets the answer instead
of writing a cleverer test that also cannot work.

This project's standing rule is that a discriminator which cannot show its lever moved is void. M5
is the disciplined inverse — showing precisely which half of the fix has no observable lever, and
declining to imply coverage that does not exist.

M3 and M6 are the complements that make the rest binding: both segfault rather than merely
reddening an assertion, so the retained-null-view path and the `return out` are demonstrated to
matter at the process level. M2's split — a constant site name reddening exactly the seven
non-render-pass sites — is the same arm that earned #3211 its approval, and it is the one that
separates "eight sites are independently wired" from "one reports for all".

## Trap 245 is a real contribution, not bookkeeping

"A one-shot injection aimed at a site behind a process-wide cache never fires, and stays armed into
the *next* test's control run, where it reads as that test being broken" is a failure mode that
survives the obvious defence, and the PR says why: the control run that makes an arm trustworthy is
the same run that made it unreachable, so adding a control does not catch it. That belongs in the
list.

The new `tests/fixtures/AGENTS.md` is overdue and correct — `render_runner.h` living under `tests/`
while being compiled by three frontend translation units is exactly the kind of thing a folder map
exists to stop someone learning the hard way.

## One follow-up, not blocking

The new `return out` on render-pass failure leaks the colour images and memory created earlier in
the function. That matches the convention the existing colour-target failure paths already follow,
so it is consistent rather than a regression introduced here — but it *is* newly reachable, since
the old code proceeded instead of returning. Worth an issue for the whole family of `return out`
paths in this function rather than a change in this diff, which is deliberately mechanical.

## Verification and hygiene

Build rc 0, `ctest --no-tests=error` 100% of **332**, new case registered by name as
`render_object_create_failure` (not inferred from a count — the check that caught a silently
unregistered test here yesterday). Every mutation arm reports its build exit code first, so none of
them is a baseline re-run in disguise. All nine deleted lines accounted for as the replaced call
sites (trap 41). No session trailers, no host paths, `diff --check` clean.

Docs gate and `check_merge_result.py` green at authoring time; I will re-run the merge-result gate
immediately before merging, since another lane landing invalidates it and this PR appends to a
numbered table.

PR #3023, opened 2026-08-26
gpu: group a colour pass on the surface it renders to, not a second representation of it

mattias800:
## What this is

**A same-day regression fix that un-reds master.** `97ecc58a` ("gpu: GTA V world rendering series")
introduced `mrt_same_color_pass` and, with it, silently changed where pass grouping reads slot 0's
identity from. Its own diff shows the predicate it replaced:

```
-  if (draw.color0_base != base || ...)                                  // named color0_base
+  if (!prosper::frontend::mrt_same_color_pass(pass_head, draw, ...))    // array-first
```

The address the pass **actually renders to** was not changed and still comes from the named
`color0_base` (`live_renderer.cpp:7046`, and `pass_bases[0]` at `:7207`; every slot above 0 goes
through the array on both paths). So grouping and target selection began consulting two different
representations of one attachment.

When they disagree, **two draws that render to different addresses group into one pass.** The group
renders at the first draw's target, so the second draw's surface is silently discarded: no pass, no
published pixels, and the frame falls back to the previous draw's surface at the previous draw's
extent.

The same read also defeated — for slot 0 only — the extent term that series added to stop two mip
levels sharing one attachment. A stale array entry carries no extent, and unknown extents never
conflict, so same-address/different-extent draws grouped anyway.

## Failure scenario

`render_submit_items({producer, consumer}, 64, 64)`, producer rendering to `0x300000` at 32x2 and
consumer to `0x400000` at 16x16, returns **256 bytes** (32x2x4 — the producer's extent) instead of
1024. Measured:

```
[lane] item=0 named_base=0x300000 named=32x2  array_base=0x200000 array=0x0
[lane] item=1 named_base=0x400000 named=16x16 array_base=0x200000 array=0x0
[pass] cb=31 pass=2/2 base=0x300000 32x2 ... bytes=256
```

Both items carry the same stale array entry, so grouping produced one pass at the producer's target
(`pass=2/2` — the cursor consumed both items). `present_extent.hpp` documents this contract and names
the failing assertion by its exact text.

## The change

Compare slot 0 on the named triple whenever it names a surface; fall back to `mrt_color_binding`
when `color0_base` is 0, which preserves the previous answer for a draw that names no slot-0 surface.

## Why this cannot change captured or live rendering

Not merely "no divergence was found" — for captures a divergence is **inexpressible**: the wire
format carries slots 2 and up only, and `restore_legacy_color_target_aliases` re-derives slots 0/1
[...]

Issue #3015, opened 2026-08-26
gpu_capture_render_replay: the per-target chain drops the consumer draw after the rendering series (4 arms)


mattias800:
Master is red on this test. Four arms fail, all pre-existing, all broken by the GTA V rendering
series (`97ecc58a`). Recording what is **established** and — more usefully — the four hypotheses
already falsified, so nobody re-derives them.

## Established

`render_submit_items({producer, consumer}, W=64, H=64)` returns **256 bytes**, not the expected
1024. 256 = **32x2x4**, which is the PRODUCER's extent. So the consumer's 16x16 draw never produces
a pass and the chain falls back to the previous one; the renderer's own diagnostic agrees
(`px_last=32x2@0x300000`).

The consumer binds `resources[0]` at the producer's address with `host_data = nullptr`, so it
depends entirely on the renderer resolving the producer's **rendered** output as a texture. That
resolution is what "per-target chain" names, and it is what broke.

Failing arms, all in the same semantic area:

```
per-target chain returns the last pass at its declared 16x16 extent
full-size lighting pass follows a dummy-color depth prepass
oversized viewport does not inflate the persistent depth attachment
a scratch-backed retained entry survives a later span decoding another texture
```

## The arms are RIGHT — the contract names one of them by its exact text

`present_extent.hpp` documents `want_bytes == 0` as "no extent contract applies", says the
`render_submit_items` tests consume the last pass **at its own extent**, and then names the assertion
verbatim: *"tests/misc/test_gpu_capture_render.cpp asserts exactly that ('per-target chain returns the
last pass at its declared 16x16 extent')"*. So this is not a stale expectation to be adjusted.

## Falsified — do not re-derive these

| hypothesis | how it died |
| --- | --- |
| downstream of the newly-correct short-buffer refusal | made `short_tiled_storage.host_data_size` full-size; all four still failed, and it broke a fifth arm |
| the present-extent contract leaked outside its `PresentSubmitScope` | probe at `live_renderer.cpp:6911` reports `in_present_submit=0 want_bytes=0` at **every** call in this test |
| the 8x8/16x16 quadrant aliasing from the series' headline fix | that code is in `live_compute.cpp` and this is a graphics chain; 256 is the producer's extent, not a quadrant |
| the persistent-color-target eviction ceiling got stricter | `persistent_color_target_count_ceiling(true)` is `limit + 64`, i.e. MORE permissive than the old `..._count_limit()` |

## Two instrument traps hit while investigating

[...]

Issue #2266, opened 2026-08-07
perf(render): the MSAA resolve path costs ~49 ms/submit — 22% of the Blue Prince frame — in a GPU stall plus a full-surface CPU readback


mattias800:
## The largest single term in the Blue Prince frame is not measured — it is a subtraction

```
[render-window] frontend avg_ms: total=320.27 prelude=0.52 build_resources=49.22
                backend=184.73 pass_control=85.57 output_copy=0.22
```

**`pass_control` is 85.57 ms/submit — 27% of the frame.** And it is not a span. `live_renderer.cpp` computes it as:

```cpp
const double pass_control = totals.pass_ms - totals.build_resources_ms - totals.backend_ms;
```

Everything inside the pass span that is neither building resources nor the backend call. Nobody can say what any of it is.

This is the same state `build_resources` was in before #2250, and it is the same defect that #2246 records: **an unmeasured region does not read as unmeasured, it reads as somebody else's cost.**

## Where the frame actually goes (window, 2,104 draws/submit, Windows / NVIDIA RTX 4090)

| term | ms/submit | owner |
| --- | ---: | --- |
| `backend` | 184.73 | shared |
| ↳ `draw_setup` | 90.84 | |
| &nbsp;&nbsp;↳ `resources.buffer` (`copy=35.20`) | 42.62 | Linux lane, known ceiling (#1733) |
| &nbsp;&nbsp;↳ `fixed` (`pre_index_unsplit=10.37`) | 25.37 | #2254 / #2263 |
| &nbsp;&nbsp;↳ `resources.texture` (`lookup=6.98`) | 7.64 | |
| ↳ `gpu_wait` (`gpu_device=33.20`) | 38.82 | **hardware floor** |
| ↳ `readback` | 10.33 | |
| **`pass_control`** | **85.57** | **unattributed — this issue** |
| `build_resources` | 49.22 | #2262, #2250 |

## The 60 FPS arithmetic, stated plainly

60 FPS is a **16.7 ms** budget. `gpu_device` alone is **33.20 ms**, so even a renderer with zero CPU cost would land at ~30 FPS on this workload as it is currently submitted. Reaching 60 will require reducing GPU work, not only CPU work — but the CPU side is ~9× the GPU side today, so it is where the first order of magnitude is, and `pass_control` is the largest unattributed part of it.

## First step

[...]

Issue #289, opened 2026-07-10
Gameplay renders black: first level loads + presents, but produces no visible geometry (menus/dialogue render fine)


mattias800:
## Summary

With #283 fixed, a scripted New Game run now advances all the way through: title → name entry → name confirm → **opening story dialogue** (all rendering correctly). Mashing through the dialogue then transitions into the **first level** — the game loads `level0`–`level6` + `RuntimeInitializeOnLoads.json` and **keeps presenting frames** (frame dumps continue to ~1980, no stall, no crash) — but every presented frame from ~frame 960 onward is **fully black** (nonblack 0.0%, 1 distinct color).

So this is a **render gap, not a stall or a crash**: the game is running its frame loop and flipping buffers, but the gameplay scene produces no visible geometry. Menus, title, and dialogue (simple quads + text/portraits) render fine; the platforming scene (parallax layers, tilemaps, sprites) renders black — a gameplay-specific rendering path.

## Repro
```
PROSPER_GUEST_FS=1 PROSPER_GUEST_ARGS=-force-gfx-direct PROSPER_RENDER=1 PROSPER_FRAME_DIR=<dir> \
PROSPER_PAD_SCRIPT='3:start;7:start;13:cross;20:cross;26:cross;28:start;33:up;35:cross;45:cross;55:cross;70:cross;85:options;100:up;112:cross;<cross every 5s from 120..340>' \
  ./boot_trace <dump>
```
Last non-black frame ≈ `frame_0960` (the story dialogue). All frames after the dialogue→level transition are black.

## Candidate leads (from the run log)
- `[agc] GetDataPacketPayloadAddress: UNCONFIRMED packet type=0 — payload offset assumed cmd+2 (only type-1 is RE'd; type-0 header unknown, #140)` — if the gameplay command stream uses **type-0** data packets we don't decode, draws would resolve to wrong/empty payloads → no geometry. Strong suspect.
- `todo: void GfxDevicePS5SharedData::CreateWorkload()` — a still-stubbed GPU workload path.

## Next step
Confirm the game IS submitting gameplay draws (PROSPER_GFXLOG) during the black period, then check whether those draws reference type-0 packets (#140) or a render target / scene-activation that our executor drops. This is the next gameplay-rendering frontier now that the New Game path is crash-free.

mattias800:
Duplicate of #300 (the canonical user-facing issue for this). Consolidating investigation there; keeping the instruction-level analysis above for reference. Closing as dup.

PR #3216, opened 2026-09-02
gpu: key a resolve pass on the destination its copy acts with

### LINKED ISSUE #3025: Consecutive MSAA resolves into different destinations group into one pass, and all but the first destination are dropped
Same defect class as #3015 (fixed by #3023), on the path that #3023 deliberately did not touch.
Found by independent review of that PR.

## The mismatch

The resolve path reads its destination from the **raw named field**:

```cpp
// frontends/shared/live/live_renderer.cpp
const uint64_t rsrc = pass.front()->color0_base;
const uint64_t rdst = pass.front()->color1_base;      // <-- raw named, not mrt_active_color
```

Pass grouping keys slot 1 on `mrt_active_color(draw, 1, ...)`, which is **0 for every resolve** — a
MODE=RESOLVE draw exports nothing, and the grouping code says so in its own comment. So slot 1
contributes no discriminating term for resolves. Grouping *is* keyed on `ps.cb_resolve`, which
correctly stops a resolve merging with ordinary scene draws — but that key is a boolean, so it does
not separate two resolves from each other.

## Failure scenario

Two consecutive draws, both `cb_resolve`, sharing `color0_base` (or both with slot-1 active colour
0), resolving into **different** `color1_base` destinations. They satisfy every grouping term, so
they become one pass. The path then takes `rdst` from `pass.front()` only, performs that one copy,
and ends the group with

```cpp
continue;   // resolve is a copy, not an ordinary-draw render
```

Every other resolve in the group is discarded — no copy, no render, no diagnostic. The second
destination keeps whatever it held before, which downstream reads as a stale or never-written
surface.

## Why it has not been seen

It needs two adjacent resolves with different destinations in one submit. Blue Prince (PPSA25009),
the title that motivated the resolve path, resolves one 4x-MSAA scene target per pass, so its
resolves are separated by ordinary draws and never group with each other. That makes this latent
rather than active, which is exactly why it should be recorded before it is next hit.

## Suggested fix

Give the resolve its own destination term in the grouping predicate — key the group on
`color1_base` (the field the copy actually uses) whenever `cb_resolve` is set, so two resolves with
different destinations cannot share a pass. The narrower alternative is to perform one copy per item
in the group rather than one per group, but keying is closer to the rest of the pass model and does
not change the "a resolve is its own pass" invariant.

Either way the principle is the one #3023 established: **the identity a pass groups on must be the
identity it acts with.** Here the group acts with `pass.front()->color1_base` and groups on
something that is constantly 0.

Refs #3015, #3023

mattias800:
## What this is

The same defect class as #3015 (fixed by #3023), on the path #3023 deliberately left in place: a
pass grouped on an identity that is not the identity it acts with. Here the identity is a
**fixed-function MSAA resolve's destination**.

## Failure scenario

`CB_COLOR_CONTROL.MODE=RESOLVE` is answered by a straight copy of the already-rendered `color0_base`
surface into the raw named `color1_base`. The copy is performed **once for the pass**, from
`pass.front()`, after which the group ends (`continue;   // resolve is a copy, not an ordinary-draw
render`).

Pass grouping did not carry that destination:

* Slot 1 is inactive on **every** resolve. A fixed-function resolve exports nothing, so
  `color1_write_mask` is 0, so `mrt_active_color_count` returns **1** — which means
  `mrt_same_color_pass` does not merely compare a constant at slot 1, it **never reaches slot 1 at
  all**. (Worth stating precisely: the issue describes the effect correctly, the mechanism is the
  loop bound rather than a zero comparison.)
* The only resolve term in the key was `items[pass_i].ps.cb_resolve == resolve0` — a boolean. It
  correctly keeps a resolve out of a scene group, and cannot separate one resolve from another.

So two adjacent `cb_resolve` draws sharing a source and naming **different** `color1_base`
destinations satisfied every grouping term, became one pass, had one copy performed, and every other
destination was discarded — **no copy, no render, no diagnostic**. The dropped destination keeps
whatever it held, which downstream reads as a stale or never-written surface.

## Behavioural contract

A resolve pass is keyed on the pair its copy acts with, `(color0_base, color1_base)`. Two resolves
share a pass only if they name the same source *and* the same destination, in which case the two
copies are identical and one is enough. A resolve never shares a pass with an ordinary draw.

## The change

`mrt_same_resolve_pass` in `frontends/shared/rtt/mrt_binding.hpp`, beside `mrt_same_color_pass`,
carrying the whole resolve half of pass identity; `live_renderer.cpp`'s grouping loop keys on it.

```cpp
inline bool mrt_same_resolve_pass(const DrawItem& first, const DrawItem& candidate) {
    if (first.ps.cb_resolve != candidate.ps.cb_resolve) return false;
    if (!first.ps.cb_resolve) return true;
    return first.color1_base == candidate.color1_base;
}
```

Deliberately the **raw named field**, for the same reason the copy reads it raw: the active-binding
rule reports 0 for every resolve, so keying on `mrt_active_color(draw, 1, …)` would key on a
constant. That is mutation arm B below.

### Alternatives considered

| option | why not | what it would have cost |
| --- | --- | --- |
| **Key the group on the destination** (taken) | — | One field compare per candidate draw. Splits only groups that were already wrong. |
| Perform one copy **per item** in the group instead of one per group | Redundant once the key is complete: with the destination keyed, every member names the same `(source, destination)`, so the extra copies are byte-identical repeats of the first. It also leaves the group's *identity* still wrong, so any future reader of `pass.front()` inherits the same bug. | ~200 lines of flush/readback/copy logic moved inside a loop, and N identical GPU copies where one is correct. |
| Stop grouping resolves with each other entirely (one resolve = one pass) | Correct, and simpler to state, but strictly worse behaviour: a run of genuinely identical adjacent resolves would perform N identical copies, each with its own pipeline flush — and the resolve flush is the single most expensive early exit in this loop (Blue Prince, #2266). | Same correctness, unnecessary flushes. |

### Fail-visible, not silent

The one remaining way a resolve can produce nothing is the `resolve SKIP` branch — no source base, no
destination base, or a source with no rendered surface. It used to print **only** under
`PROSPER_MSAA_LOG`, i.e. dropped rendered content with no default-run trace. It now reports
unconditionally, bounded at 8 like the neighbouring copy-failure warning; `PROSPER_MSAA_LOG` still
reports every occurrence.

## Verification — and how the positive instance was built

**This defect is latent: no title capture demonstrates it, so the regression test IS the evidence.**
It was therefore constructed by hand, outside anything that produced the null, rather than derived
from a generator or a capture — the trap this project records as a same-source positive control
testing the discriminator rather than the domain.

The construction (`tests/misc/test_gpu_capture_render.cpp`, run through the real live renderer):

1. an ordinary draw renders content into `0x1a00000` at 32x32;
2. a `cb_resolve` draw with `color0_base = 0x1a00000`, `color1_base = 0x1a10000`,
   `color1_write_mask = 0` (as a fixed-function resolve really is);
3. an identical one whose only difference is `color1_base = 0x1a20000`;
4. all three in **one** `render_submit_items` call, so they are adjacent items in one span.

**The load-bearing assertion is on the destination CONTENTS**, not on a pass count and not on a
target merely existing: `read_live_render_target(0x1a20000)` must return a 32x32 surface whose pixels
are **byte-identical to the source's**. Two supporting arms make it hard to pass for another reason —
an anti-triviality arm requiring the source to be non-blank (two blank surfaces would compare equal),
and a first-destination control arm that passes both before and after the fix, so a run where *both*
destination arms fail is a broken fixture rather than this defect.

`frontends/shared/tests/test_mrt_binding.cpp` adds pure-predicate arms for the mechanism, including
the premise asserted rather than assumed — `mrt_active_color(resolve, 1) == 0`,
`mrt_active_color_count(resolve) == 1`, and `mrt_same_color_pass(resolve, second) == true`, i.e. that
the colour predicate genuinely cannot tell the two apart. Without that arm the new arms could pass
because the term was already discriminating.

### Commands and results (Linux, distrobox `ps5ys`)

```
cmake -S prosper -B prosper/build-linux                        # rc=0
cmake --build prosper/build-linux -j6                          # rc=0
ctest --no-tests=error                                         # rc=0, 100% tests passed out of 332
```

`ctest -N` confirms both affected cases by name: `#128 mrt_binding`, `#289 gpu_capture_render_replay`.
No new ctest executable is registered (the arms extend existing tests), so the count is unchanged
from master's 332 by construction rather than by accident.

### Mutation arms — both wrong fixes are rejected, and both compiled

| mutant | build | `mrt_binding` | `gpu_capture_render_replay` |
| --- | --- | --- | --- |
| **A** — `return true;` (exact pre-fix behaviour: `cb_resolve` is the only term) | rc=**0** | rc=1, `:308 !mrt_same_resolve_pass(resolve, second)` | rc=1, **1** FAIL: *a second resolve into a different destination receives the resolved pixels* |
| **B** — key on `color_targets[1].base` (the ARRAY representation, 0 here) instead of the raw named field | rc=**0** | rc=1, same arm | rc=1, same single FAIL |

Both mutant builds exited 0, so the arms ran against the mutated code rather than a stale binary.
Under mutant A — which *is* the unfixed renderer — the control arms stay green and only the
load-bearing arm reddens, which is the discriminating result: the first destination is written, the
second is dropped.

* **Affected:** grouping of two or more adjacent `cb_resolve` draws that name different
  `color1_base` destinations; and the default-run visibility of a skipped resolve.
* **Unaffected:** ordinary draws — the destination term is inside `if (first.ps.cb_resolve)`, so a
  stale named `color1_base` can never split a scene group (asserted). Slot 0's identity, slots 1 and
  above, feedback detection, the DS identity term, the copy itself, its flush/readback ordering, and
  every `PROSPER_*` lever. A resolve group whose members share a destination still groups and still
  performs exactly one copy.
* **Blue Prince (PPSA25009)**, the title the resolve path was built for, separates its resolves with
  ordinary draws, so they never grouped with each other and its behaviour is unchanged.

## Risks

Low. The predicate can only *refuse* groupings that previously succeeded, so the failure mode of a
mistake here is an extra pass, not a dropped one. The extra pass costs one `submit_and_wait` flush on
the resolve path. The one behaviour a reader might not expect is the new unconditional `resolve SKIP`
line; it is bounded at 8 per process.

## A lead this turned up — stated as a hypothesis, not a measurement

The issue records this as latent because no title is known to emit the shape. **Asterix & Obelix:
Babylon Mission (PPSA30490) may.** `docs/ASTERIX_BABYLON_STATUS.md` records one 14-operation submit
containing **four** `MODE=RESOLVE` operations at 5, 6, 9 and 10 — in two **adjacent** pairs — all four
reading one source (`4→5`, `4→6`, `4→9`, `4→10`). That is exactly the precondition. Whether the
members of a pair name *different* `color1_base` values is not recorded anywhere, so this is
untested, and I could not test it (no emulator access this session).

What is certain is that the existing Asterix evidence **cannot** settle it in either direction: every
measurement in that document reads `792bed5a3f02a383` — uniformly black — for the source and for both
recorded destinations, so a destination that was never written is byte-identical to one that was.
Recorded as a `## Ruled out` bullet in `docs/ASTERIX_BABYLON_STATUS.md` so nobody reads that
investigation as having cleared this title.

Fixes #3025

### DIFF
--- a/prosper/docs/ASTERIX_BABYLON_STATUS.md
+++ b/prosper/docs/ASTERIX_BABYLON_STATUS.md
@@ -649,6 +649,17 @@ HLE registration. #1599, #1884.
   readback. It returned the same `792bed5a3f02a383` uniformly-black pixels and byte-identical BMP as
   operation 4 and the op6 destination. Both resolves therefore copy black to black in this capture;
   neither loses useful color. This is localization evidence only; rung remains 0.
+- **The existing MSAA evidence here already covers #3025 (consecutive resolves grouping into one
+  pass, all but the first destination dropped):** falsified — it cannot, in either direction. This
+  submit has the exact shape the defect needs: four `MODE=RESOLVE` operations at 5, 6, 9 and 10, in
+  two ADJACENT pairs, all four reading one source (`4→5`, `4→6`, `4→9`, `4→10`), so nothing separates
+  a pair into two passes but the destination term #3025 added. But every measurement above reads
+  `792bed5a3f02a383` — uniformly black — for op4's source and for both recorded destinations, so a
+  destination that was never written is byte-identical to one that was, and neither the op6 nor the
+  op10 selector could have distinguished them. Whether ops 5 and 9 name a *different* `color1_base`
+  from their partners is not recorded anywhere, so whether the defect fires on this title is
+  **untested, not ruled out** — re-derive the four raw `color1_base` values from a fresh capture
+  before treating this submit as unaffected. #3025.
 - **An omitted immediately preceding submit as operation 4's black temporal input:** a bounded two-frame
   whole-frame capture contains consecutive submits 180 and 181. Resolve-aware scanning identifies submit
   180 operation 4 as the last semantic writer of submit 181 operation 4's same run-local `0x2033b10000`
--- a/prosper/frontends/shared/live/live_renderer.cpp
+++ b/prosper/frontends/shared/live/live_renderer.cpp
@@ -7757,12 +7757,6 @@ void register_live_renderer(const std::string& frame_dir, bool dump_bmps_request
                     const VkFormat format0 = pass_formats[0];
                     const VkFormat format1 = base1
                         ? pass_formats[1] : VK_FORMAT_UNDEFINED;
-                    // A MODE=RESOLVE draw shares color0_base with the scene it resolves and reports
-                    // active_color1()==0 (it exports nothing), so without keying the group on cb_resolve
-                    // it could merge with the scene draws; the resolve-copy `continue` below would then
-                    // silently drop any ordinary draws grouped after it. Key on it so a resolve is always
-                    // its own pass.
-                    const bool resolve0 = items[pass_i].ps.cb_resolve;
                     // The DEPTH/STENCIL attachment is part of a pass's target identity, not just its
                     // colour attachments.
                     //
@@ -7795,8 +7789,14 @@ void register_live_renderer(const std::string& frame_dir, bool dump_bmps_request
                         if (ds_identity(draw) != ds0) return false;
                         return true;
                     };
+                    // A MODE=RESOLVE draw is answered by a COPY below, not by a render, so its
+                    // pass identity is the source/destination pair that copy acts with rather than
+                    // the attachment set `same_targets` compares. `mrt_same_resolve_pass` carries
+                    // that whole rule -- keeping a resolve out of a scene group, and keeping two
+                    // resolves with different destinations out of each other's (#3025) -- and its
+                    // header states why each half exists.
                     while (pass_i < items.size() && same_targets(items[pass_i]) &&
-                           items[pass_i].ps.cb_resolve == resolve0) {
+                           prosper::frontend::mrt_same_resolve_pass(pass_head, items[pass_i])) {
                         pass.push_back(&items[pass_i]); ++pass_i;
                     }
 
@@ -8000,10 +8000,20 @@ void register_live_renderer(const std::string& frame_dir, bool dump_bmps_request
                             if (PROSPER_ENV_ON("PROSPER_MSAA_LOG"))
                                 fprintf(stderr, "[msaa] resolve copy 0x%llx -> 0x%llx (%ux%u)\n",
                                         (unsigned long long)rsrc, (unsigned long long)rdst, rw, rh);
-                        } else if (PROSPER_ENV_VALUE("PROSPER_MSAA_LOG")) {
-                            fprintf(stderr, "[msaa] resolve SKIP src=0x%llx dst=0x%llx (%s)\n",
-                                    (unsigned long long)rsrc, (unsigned long long)rdst,
-                                    !rsrc || !rdst ? "missing base" : "source has no rendered surface");
+                        } else {
+                            // A resolve that performs no copy is DROPPED RENDERED CONTENT: the
+                            // destination keeps whatever it held before, and every later census
+                            // reads it as "written by nothing". Report it unconditionally, bounded
+                            // like the copy-failure warning above, so the one remaining way a
+                            // resolve can produce nothing is fail-visible rather than silent.
+                            // PROSPER_MSAA_LOG still reports every occurrence.
+                            static std::atomic<int> skipped{0};
+                            if (PROSPER_ENV_VALUE("PROSPER_MSAA_LOG") ||
+                                skipped.fetch_add(1) < 8)
+                                fprintf(stderr, "[msaa] resolve SKIP src=0x%llx dst=0x%llx (%s)\n",
+                                        (unsigned long long)rsrc, (unsigned long long)rdst,
+                                        !rsrc || !rdst ? "missing base"
+                                                       : "source has no rendered surface");
                         }
                         if (timing_enabled) ++pending_timing.resolve_n;
                         continue;   // resolve is a copy, not an ordinary-draw render
--- a/prosper/frontends/shared/rtt/mrt_binding.hpp
+++ b/prosper/frontends/shared/rtt/mrt_binding.hpp
@@ -144,6 +144,43 @@ bool mrt_same_color_pass(const prosper::gpu::DrawItem& first,
     return true;
 }
 
+// Can two consecutive draws share one pass given their FIXED-FUNCTION RESOLVE identity?
+//
+// A CB_COLOR_CONTROL.MODE=RESOLVE draw is not an ordinary draw: live_renderer answers it with a
+// straight copy of the already-rendered `color0_base` surface into `color1_base` and then ends the
+// group, because a resolve is a copy and not a render. Two consequences follow, and the predicate
+// has to carry both.
+//
+// 1. A resolve must never group with an ordinary draw. It shares `color0_base` with the scene it
+//    resolves and reports `mrt_active_color(draw, 1) == 0` -- a fixed-function resolve exports
+//    nothing, so its `color1_write_mask` is 0 and slot 1 contributes no attachment -- so on colour
+//    identity alone it would merge with those scene draws, and the copy's `continue` would then
+//    silently drop every ordinary draw grouped after it.
+//
+// 2. Two resolves must not group unless they resolve into the SAME destination. `cb_resolve` alone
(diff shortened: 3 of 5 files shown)


mattias800 (commented):
**APPROVED.**

## The substitution is exactly equivalent outside the case it is meant to change

Replacing `items[pass_i].ps.cb_resolve == resolve0` with `mrt_same_resolve_pass(pass_head, ...)`
could have altered ordinary draw grouping, which would be a much larger blast radius than the
defect. It does not:

- non-resolve pair → `if (!first.ps.cb_resolve) return true;`, contributing no term, exactly as the
  old equality did;
- mixed pair → the first line returns false, exactly as the old equality did;
- resolve/resolve → the new destination term.

So this is a strict refinement of one case. I also checked the argument change rather than assuming
it: `resolve0` was read from `items[pass_i]` while `pass_i` was still the pass head, and
`pass_head` is `const DrawItem& pass_head = items[pass_i]` bound at that same point — a reference to
the element, not a re-evaluated expression, so it does not follow `pass_i` as the loop advances.
`ds0 = ds_identity(pass_head)` on the next line already relies on that, so the new call is
consistent with the established pattern rather than introducing a new assumption.

## The choice among the three options is argued, not asserted

Keying is right, and for the reason given: once the destination is part of the key every member of a
group names one identical `(source, destination)` pair, which makes the single copy correct for all
of them — so iterating the group would perform N byte-identical GPU copies and *still* leave
`pass.front()` wrong for the next reader of the group's identity. Fixing identity rather than
compensating at the point of use is the better shape. Declining to group resolves at all is also
correct and strictly worse, since the resolve flush is the most expensive early exit in that loop.

Keying on the **raw named** `color1_base` rather than `mrt_active_color(draw, 1, …)` is the subtle
part and is exactly right — the active-binding rule reports nothing for a resolve, so keying on it
would key on a constant and change nothing. That mutation arm B reproduces the unfixed behaviour
byte-for-byte is the proof that the distinction is load-bearing rather than stylistic.

## The test is built so it cannot pass vacuously

This is the part I looked at hardest, because a latent defect has no title capture behind it and the
regression test *is* the entire evidence. It holds up:

- the load-bearing assertion reads the second **destination surface's contents** and requires
  byte-equality with the source, not a pass count or a call count;
- the anti-triviality arm requires the source to be non-blank first, which closes the obvious
  vacuous pass — two never-written surfaces compare equal, and without that arm the whole test would
  be satisfied by rendering nothing at all;
- the first-destination control passes both before and after, so "both destinations failed" reads as
  a broken fixture rather than as this defect.

Mutation arm A is the ideal result: `return true;` is the *literal* unfixed renderer, and it reddens
**exactly one** assertion — the load-bearing one — with both controls green. A mutation that reddens
everything proves much less, because it cannot distinguish a load-bearing assertion from a fixture
that stopped working.

## The mechanism correction and the Asterix finding both improve the record

Correcting "slot 1 is 0 for a resolve" to "`mrt_active_color_count` returns 1, so
`mrt_same_color_pass`'s loop never reaches slot 1 at all" is worth making even though the fix is
unchanged — the issue was written from review rather than measurement, and a premise checked in a
unit arm is better than a premise restated.

The *Asterix & Obelix: Babylon Mission* lead is the more valuable half. "No title is known to
exercise this" may well be wrong — a recorded 14-operation submit carries four `MODE=RESOLVE`
operations as two adjacent pairs off one source, which is this defect's precondition exactly. And
the honest part is the second half: the existing evidence **cannot settle it either way**, because
every recorded surface there reads uniformly black, so a never-written destination is byte-identical
to a written one. That is a void discriminator, correctly identified as one and recorded as such
rather than quietly read as a negative. Filing it as a `## Ruled out` bullet in that status doc is
the right home for it.

Making the `resolve SKIP` branch report unconditionally (bounded at 8, matching the neighbouring
copy-failure warning) is the right companion change: after this fix the remaining way a resolve
produces nothing is a genuinely dropped surface, and that must not be visible only under a
diagnostic env var.

## Verification

Build rc 0, `ctest --no-tests=error` 100% of **332** — matching master by construction, since no new
executable was added and the new arms live in existing targets, both confirmed present by name.
Both mutation arms compiled (build rc 0) before running, so neither is a baseline re-run in
disguise. CI 9 SUCCESS on head `66b8baa3`, `mergeable=CLEAN`. No session trailers, no host paths.

Issue #257, opened 2026-07-09
Cutscene caption text renders no geometry: NGG text mesh's computed vertex index folded to gl_VertexIndex (dyn_vfetch)


mattias800:
## Summary

After the font-atlas de-swizzle fix (PR #256, refs #102), the 2048×1024 R8 caption font atlas decodes correctly (verified: readable Latin+CJK glyph grid). But the intro **caption text still does not appear on screen** — because of a **second, independent bug** in the shader recompiler's vertex-fetch path. The caption text mesh renders zero visible geometry.

## Evidence (how it was traced)

The intro cutscene submit (per `PROSPER_DRAWDIAG`) is:
```
idx=6    tex=[427x240]                 <- the scene image (island / hero) — renders fine
idx=1566 tex=[4x4,4x4,2048x1024]       <- the CAPTION text mesh (~261 glyph quads, samples the font atlas)
```
So the caption mesh **is** in the scene submit and samples the (now-correct) font atlas — but:

- `PROSPER_TESTTEX` (force every sampled texture to a bright *opaque* checker) leaves the letterbox **pure black (max=0)** on every scene slide → **no geometry rasterizes there at all** (rules out alpha/atlas/discard; it's a geometry collapse).
- The caption VS (`es=0x…52f00`, shared by idx=1566/528/24) is a complex **NGG primitive shader**. Its transform math is fine — the `v_dot2c_f32_f16` labels from `llvm-mc -mcpu=gfx1030` are misreads; on the PS5's encoding (`-mcpu=gfx1010`) they are plain `v_mac_f32` (matrix×vertex), which the recompiler already handles. So the collapse is in the **fetched vertex data**, not the transform.

## Root cause (precise)

`src/gpu/rdna2_to_spirv.cpp` ~L1965 (the `dyn_vfetch` per-vertex attribute fetch, #206):
```cpp
if (dyn_vfetch && idxen && stride) {
    addr = b.ibin(Op_IMul, b.load_vertex_index(), b.uconst(stride));  // gl_VertexIndex * stride
}
```
The `dyn_vfetch` path **assumes the fetch element index == `gl_VertexIndex`** (and drops the shader's VADDR/offset/SOFFSET). The #206 comment states the reason: *"The NGG fetch-shader prologue that computes the element index isn't fully modeled and folds to a constant."*

That assumption holds for straightforward vertex arrays (the scene meshes render correctly). But the **caption text mesh computes its fetch index at runtime**:
```
[...]

PR #207, opened 2026-07-09
fix(hle): track the MsgDialog lifecycle — NONE before Open, not spurious FINISHED

mattias800:
## Summary
- `sceMsgDialogUpdateStatus` / `sceMsgDialogGetStatus` returned **FINISHED(3) unconditionally**, including before any Open, where the real API reports NONE/INITIALIZED. A guest polling GetStatus as a guard saw dialog-already-completed at the wrong lifecycle stage and **skipped or duplicated** its dialog logic.
- Track the real 3-state transitions with a small atomic: `Initialize` → INITIALIZED(1); `Open` → FINISHED(3) (headless **auto-dismiss**, so the game's wait-until-dismissed loop still exits immediately); `Close`/`Terminate` → NONE(0). GetStatus/UpdateStatus report the current state.
- The auto-dismiss policy is **unchanged** — only the pre-Open state is fixed (NONE/INITIALIZED instead of FINISHED), exactly as the issue scoped it.

## Verification
- New checks in `test_service_getters`: GetStatus is NONE before Initialize, INITIALIZED after it (also via UpdateStatus), FINISHED after Open, and NONE again after Close.
- Full suite in WSL build-linux **including the dump-backed boot: 65/65 passed** (the startup dialog flow still completes headless).

Fixes #144

Issue #3148, opened 2026-08-30
gris-gameplay guard fails on master: the title stops at the publisher splash and never reaches the guarded scene


mattias800:
`gris-gameplay` — one of the two reviewed rung-6 gameplay guards — fails on current master. The title never reaches the guarded scene; at 155 s it is still on the publisher splash.

## Measurement

Run locally against `origin/master` at `ba70a787`:

```
[check] gris-gameplay: FAIL — qualifying frames 58 < 60; structural matches 0 < 60 (min SSIM 0.85)
```

`structural matches 0 of 60` is a total mismatch, not a marginal one. The retained frames are **not
black** (`nonblack=1.000`, max 252-255) — they are the **Devolver publisher splash**, at
`elapsed_seconds ≈ 155`, where the baseline expects gameplay.

## It is not a slowdown, and not the renderer

The same run reports **415 fps while producing frames, 121 fps run average, 264 fps presented**, 28%
of a 235.7 s run active. The emulator is fast; the guest is sitting on the splash. So this is a
**progression** regression — the boot no longer advances past the publisher logo in the guard's
window — rather than a rendering or performance one.

## Not caused by the PR I found it from

Reproduced on both sides of #3147 with an effectively identical signature, which is what makes this a
master regression rather than a branch one:

| build | qualifying frames | structural matches |
| --- | --- | --- |
| `origin/master` `ba70a787` | 58 / 60 | 0 |
| #3147 branch | 59 / 60 | 0 |

The one-frame difference is run-to-run variation, not a difference in kind. Two other guards pass on
the same PR build in the same session — `messenger-scene` OK and `dead-cells-gameplay` OK
(44 frames, 20,987 colours, 44/44 structural matches) — so the harness and the GPU are fine and this
is specific to GRIS.

## What I have not done

No bisect. Today merged #3141, #3132, #3144 and #3145, all of which are default-off diagnostics or
diagnostic plumbing, so none is an obvious culprit — which is exactly why this needs a bisect rather
than a guess. `gris-gameplay` is a **reviewed** guard (tracker #1869) and its baseline is not in
question here; what changed is that the title stops advancing.

Worth checking first, because it is cheap and this project has been bitten by it before: whether the
guard's save isolation is doing what it should (instrument trap 234 — a stale save directory makes a
working title look broken, and #3100 was specifically about isolating save roots in the snapshot
runner).

mattias800:
[...]

PR #3217, opened 2026-09-02
executor: never renumber the recompiler's reserved internal GDS binding (#3214)

### LINKED ISSUE #3214: Astro Bot (PPSA21564): the world-map pixel shader's per-tile light-list walk never reaches its 0xffffffff terminator
## Summary

Astro Bot's world-map GPU reset is a **non-terminating loop in the pixel shader `0x5008f1400`**
(established in #3193 — see the ordinal sweep there and in
`docs/ASTROBOT_LINUX_HANDOFF_2026_07_19.md`). This issue is the remaining half: *why* that loop
never ends.

## The loop

Guest pc 1911..2538 of `0x5008f1400`, a per-screen-tile walk down a linked list of light records:

```text
1911: v_mov_b32 v55, 0                        ; header  (dispatch ordinal 45)
1913: s_cmp_lg_u32 -1, s107                   ; SCC = (link != 0xffffffff)
1914: s_cbranch_scc0 -> 2539                  ; THE ONLY EXIT
1917: s106 = s107 << 3                        ; record = link * 8 bytes
1918: s107 = (s107 << 1) + 1
1920: s_buffer_load_dword s14, s[0:3], s106   ; record word 0 (type tag in 0x70000000)
1922: s106 = s107 << 2
1926: s_buffer_load_dword s66, s[0:3], s106   ; record word 1 -> the NEXT link
 ...
2537: s_mov_b32 s107, s66
2538: s_branch -> 1911
```

The first link is read at pc 1909, indexed `((s107 >> 16) + s64 * (s107 & 0xffff)) * 4` into a
**32,400-dword** scalar buffer — exactly `3840/16 x 2160/16` screen tiles — with `s107` seeded from
the interpolant `v5`. The record buffer resolves to 8,294,400 dwords.

## Why it cannot end quietly

Both link loads are lowered by `bounded_cbuf_load` in
[`src/gpu/recompiler/rdna2_emit_alu.cpp`](https://github.com/mattias800/prosper/blob/master/prosper/src/gpu/recompiler/rdna2_emit_alu.cpp),
which clamps the dword index and returns **0** for an out-of-range read. In the emitted SPIR-V:

```text
%16372 = OpShiftRightLogical %uint %16371 %uint_2
%16373 = OpULessThan %bool %16372 %uint_8294400
%16374 = OpSelect %uint %16373 %16372 %uint_0
%16375 = OpAccessChain %_ptr_StorageBuffer_uint %33 %uint_0 %16374
%16377 = OpSelect %uint %16373 %16376 %uint_0       <-- OOB reads 0
```

That substitution is **faithful** — RDNA2's OOB contract for a bounded V# does return zero — so the
lowering is not obviously wrong. The problem is what zero *means* here: this walk's terminator is
`0xffffffff`, and **zero is a valid link**. So any out-of-range read, or any run of zeros where the
list should be, turns a bounded walk into an unbounded one with no error, no reject, and no
diagnostic anywhere.

## What to determine

Which of these is actually happening:

1. **The tile index is wrong** — `v5` does not carry what the guest expects, so the head lookup lands
   outside the 32,400-dword tile buffer (or on the wrong tile) and reads 0.
2. **A descriptor is mis-resolved** — the tile buffer or the record buffer resolves with the wrong
   base or `num_records`, so in-range guest indices clamp.
3. **The list was never populated** — the compute pass that builds the per-tile light lists did not
   run, or wrote somewhere prosper does not read back, leaving zeros. Astro Bot has a long history of
   failing compute dispatches on this route (#1742 / #1743), so this is not a remote possibility.
4. **The guest's list is genuinely cyclic in prosper's copy** — the precedent is GTA V's
   `0x413dc6700`, where `PROSPER_COMPUTE_PARENTSCAN` established 22/22 that every runaway read an
   already-cyclic array. A CPU-side census of this buffer would answer it the same way.

(1)-(3) are all "prosper hands the shader data the hardware would not have handed it"; (4) is the
same class one level up.

## Reproduce

```bash
cd <REPO_ROOT>/prosper
# the hang, deterministic
PROSPER_GUEST_ARGS=-force-gfx-direct PROSPER_RENDER=1 PROSPER_VULKAN_LIB=libvulkan.so.1 \
  ./build-linux/screenshot <DUMP_ROOT>/PPSA21564-app0 \
    --out ~/astro-control --seconds 30 --count 5 --timeout 170

# the same route with only this loop's header capped -- 0 device losses
PROSPER_GUEST_ARGS=-force-gfx-direct PROSPER_RENDER=1 PROSPER_VULKAN_LIB=libvulkan.so.1 \
  PROSPER_DRAW_PROGRAM_CENSUS=1 \
  PROSPER_CFG_TRIP_BOUND=4096 PROSPER_CFG_TRIP_BOUND_PROGRAM=0x5008f1400 \
  PROSPER_CFG_TRIP_BOUND_PHASE=0 PROSPER_CFG_TRIP_BOUND_ORDINAL=45 \
  ./build-linux/screenshot <DUMP_ROOT>/PPSA21564-app0 \
    --out ~/astro-bound --seconds 30 --count 5 --timeout 170
```

The shader itself:

```bash
PROSPER_SHADER_DUMP_SUCCESS=<dir> PROSPER_SHADER_DUMP_PROGRAM=0x5008f1400 \
  ./build-linux/screenshot <DUMP_ROOT>/PPSA21564-app0 --out <out> \
    --seconds 30 --count 5 --timeout 170
```

## Not to re-derive

- It is not the vertex program `0x5008efd00` (zero `OpLoopMerge`, no backward branch).
- It is not either `readfirstlane` waterfall (pc 1154 / pc 1897): capping their headers leaves 59 and
  108 device losses, and the emitted SPIR-V shows their inner test is `v5 == v5`, always true.
- It is not the counted loop at pc 2690: capping its header leaves 64 losses, and it sits past loop
  C's exit so an invocation stuck at pc 1911 never reaches it.

The trip bound is a **diagnostic**, not a fix: it truncates guest control flow, and with it armed the
world map still shows only its nebula backdrop (#1459).

Refs #3193, #1809, #1459

mattias800:
## What this is

Astro Bot's world-map GPU reset (#3214, second half of #3193) has a cause, and it is one line of
prosper's binding policy. With it fixed the world map renders and keeps animating instead of
freezing on a white screen a minute in, and the device loss is gone.

`Fixes #3214`. Also lands the instrument that found it.

## The failure scenario

`assign_convention_bindings` gives every resource in a shader's table its own descriptor binding.
**One binding is not its to give out.** The recompiler HARD-CODES `kComputeInternalGdsBinding` (127)
whenever a program touches GDS (`declare_internal_gds`), so it is part of the emitted module's ABI,
not a slot the policy may choose.

Renumbering it produced a module that declares set 0 binding 127 against a table that carries that
buffer somewhere else. `validate_spirv_descriptor_interface` correctly rejects the pair:

```
[compute] skip invalid descriptor contract for program 0x5006e8500
[compute-descriptor] missing runtime binding binding=127 expected=storage-buffer actual=unknown
```

The dispatch is then declined — every time, for the life of the process. **Five** of Astro Bot's
compute programs are lost this way on every boot (`0x5006e8500`, `0x5006eac00`, `0x5006eb500`,
`0x5006ec700`, `0x500702f00`), all five with the same `binding=127` issue.

One of them matters a great deal. `0x5006e8500` is a 240x135x1 dispatch at 8x8 local size — one
workgroup per screen tile — and `PROSPER_COMPUTE_BINDS` shows it binding both halves of the
world-map light-list arena. `PROSPER_COMPUTE_PROGRAM_CENSUS` gives the ratio the once-per-program
skip line cannot: **executed 0, skipped 918.**

So the world-map pixel shader `0x5008f1400` was handed an arena of nothing but zeros. Its per-tile
walk terminates on `0xffffffff` — **not** on zero — so every tile head of 0 starts at record 0,
whose successor dword is also 0. An unwritten arena is not an empty list, it is an **infinite** one.
The shader looped until RADV reset the device. That is the hang #3193 localized to this loop.

## The chain, end to end

`assign_convention_bindings` renumbers binding 127 → descriptor contract rejects the module →
`0x5006e8500` declined 918/918 → light-list arena stays all-zero → the pixel shader's walk never
reaches its terminator → GPU hang → RADV device reset → every later frame is the same frozen picture.

## The fix

A resource that already carries `kComputeInternalGdsBinding` keeps it, and both running counters
step over that value so nothing else can be given it either. Two lines of behaviour, and the second
is not decoration: 127 is reachable by sequential assignment from the PS base (32) with 95 buffers.

**The gap was already visible and had been patched one instance at a time.** The trip-bound witness
needs the same binding-127 resource, and its injection sits *after* this call, at table
finalization, with a comment describing this exact symptom ("the emitted module declared binding 127
while the table did not carry it"). That fixed the witness. The guest's own GDS buffer is pushed
beside the `uses_gds` test, upstream of the renumber, and was never fixed.

## Also lands: `PROSPER_DRAW_LINKSCAN`

The instrument that found it, as its own commit. The graphics counterpart of
`PROSPER_COMPUTE_PARENTSCAN`: a CPU-side census of the linked lists a draw's scalar buffers contain,
taken from the exact dwords prosper is about to upload rather than from guest memory, because the
question is what the *shader* sees. Per buffer it prints the declared descriptor beside the uploaded
length (the short-upload check), a word histogram, and a memoized self-walk classifying every record
as terminating or cyclic.

The one modelling fact it rests on, and the one a reasonable implementer gets wrong: **an
out-of-range scalar buffer load returns architectural zero, and zero is a LINK, not an exit.** A
census that treated an out-of-range index as "the walk ended" would report a runaway list as
terminating — a confident wrong answer about exactly the case it exists to find.

Pure analysis in `src/gpu/diagnostics/link_list_census.{hpp,cpp}`, selector and encoding knobs beside
it, the live hook in `live_renderer.cpp` ordered with the other observers and before the skip
(instrument trap 166). Disarmed it is one bool per draw. Documented in `tools/AGENTS.md` and
`src/gpu/diagnostics/AGENTS.md`.

`PROSPER_COMPUTE_DESCRIPTOR_DETAIL=1` is a third, tiny addition: the `[compute-descriptor]` issue
detail already existed but was reachable only through `PROSPER_DBG`, whose output on a routed title
runs to hundreds of thousands of lines. The one line a reader needs in order to FIX a declined
dispatch cost a firehose, so people quoted the bare skip line instead. Same report, own switch,
`PROSPER_DBG` unchanged, at most once per program.

## Evidence

Astro Bot `PPSA21564`, `tools/screenshot`, native 3840x2160, no pad script,
`PROSPER_GUEST_ARGS=-force-gfx-direct PROSPER_RENDER=1`, `--seconds 30 --count 5 --timeout 170`.
Losses count `[backend] GRAPHICS submission failed`.

| arm | `skip invalid descriptor contract` | `0x5006e8500` | losses | pixel-distinct | max pixel stale |
| control (master `bbf84773`) | 5 programs | 0 / 918 | 65 | 3 / 5 | 60.0 s |
| control, replication | 5 programs | — | 62 | 3 / 5 | 60.0 s |
| control, replication | 5 programs | 0 / 918 | 57 | 2 / 5 | 60.0 s |
| **with the fix** | **0** | **executes** | **0** | **4 / 5** | **30.0 s** |

Every control frame from the 90 s sample onward is the same frozen picture. With the fix all three
of the 90/120/150 s samples have distinct CRCs — the frame is still changing when the run ends.

The arena itself, same instrument, same encoding:

| PS binding | zero dwords | terminator dwords | other | self-walk |
| 44 (tile heads), before | 32,400 | 0 | 0 | `terminating=0 cyclic=16200` |
| 44 (tile heads), after | **0** | 0 | **32,400** | heads, values `0x2..0x17baf` |
| 45 (light records), before | 8,294,400 | 0 | 0 | `terminating=0 cyclic=4147200` |
| 45 (light records), after | 8,100,001 | **32,400** | 161,999 | **`terminating=4147200 cyclic=0`**, longest 3 |

Exactly 32,400 terminator dwords in the record pool — one per screen tile — so every tile's list
ends where the shader looks for it.

With `PROSPER_WRITER_PROVENANCE=1` on the control, every scan reported `last writer NONE` for both
ranges against a 1,127-entry history (far below its 65,536 cap, so nothing was evicted) and a
recorder summary showing all four recorders firing: `color=90 compute-buffer=36667 dma-data=14280
write-data=57661`. The negative was about the guest, not the instrument.

## Screenshots

In `BLOG.md`'s 2026-09-02 entry: the lit world map at 90 s and 150 s of a default launch, and the
control's frozen white frame at the same moment. Unaltered `tools/screenshot` captures, downscaled
from 3840x2160 to 1920x1080 and otherwise untouched.

- **Affected:** any compute program that touches GDS, in any title. Those were all declined before
  and now run. That is the point, and it is also the main risk (below).
- **Unaffected by construction:** a table with no binding-127 resource and fewer than ~126 buffer
  resources gets byte-identical bindings. `reserved()` is false for every resource, and the
  step-over is unreachable. That covers every graphics table.
- The two new diagnostics are `getenv`-gated and emit nothing unset.

## Risks

- **Programs that were silently disabled now execute.** In another title a newly-running GDS
  dispatch could expose its own defect. This one is a correctness fix (prosper was refusing work the
  guest asked for), but it is a behaviour change with reach beyond Astro Bot.
- **Astro Bot's world-map frame rate falls** to 2.2 fps in the fixed run, against the control's
  pre-loss rate. Expected in direction — five previously-declined dispatch families now do real
  work, one at 240x135x64 threads — but the magnitude is unmeasured and not investigated here.
- **The world map is not correct yet.** Geometry is blown out to white in places and the lighting is
  wrong. That is the next question, not this one — filed as #3218, together with the 2.2 fps and one still-declined compute program (`0x500571000`).
- **Three guarded titles were checked locally**, not the whole matrix: `messenger-scene` OK,
  `dead-cells-gameplay` OK, `gris-gameplay` FAIL. The GRIS failure reproduces with this diff's one
  behavioural function reverted to master's behaviour, so it is not caused by this change — the
  A/B and its caveats are in a comment below.

## Build / test

```
cmake -S prosper -B prosper/build-linux -DPROSPER_APP=ON     # rc 0
cmake --build prosper/build-linux -j6                        # rc 0
ctest --no-tests=error                                       # rc 0 -- 334/334, serial, idle GPU
```

New cases verified by name, not by the total: `ctest -N | grep -E 'link_list_census|convention_bindings'`
→ `Test #207: link_list_census`, `Test #257: convention_bindings`.

A first full run taken immediately after the live A/B reported 5 failures — exactly the five
`recompiled_shaders_render` / `multidraw_render` / `indexed_render` / `descriptor_array_render` /
`gpu_execute` set that instrument trap 243 names as this box's contention set. Four passed on rerun;
`recompiled_shaders_render` then passed 3 of 3 via ctest and 3 of 3 directly, and the whole suite
passed 334/334 on an idle GPU. Its three failing assertions were NGG MBCNT logical-lane checks —
subgroup-width dependent, and untouched by this diff.

Docs gate, both forms, exit 0 (`check_numbered_table.py --ordered --baseline origin/master` on
`GAME_COMPAT_ORCHESTRATION.md`, then over every tracked `.md`).

### Mutation arms

Each reverts exactly one thing, builds clean, and fails exactly the intended assertion. The source
was restored and the suite re-run green afterwards.

| mutation | build rc | ctest rc | which `[FAIL]` |
| --- | --- | --- | --- |
| `reserved()` returns false (the old renumber-everything behaviour) | 0 | 8 | "the reserved internal GDS binding survives assignment" |
| both counters drop the step-over, preserve kept | 0 | 8 | "a table long enough to reach binding 127 is not given it" |

The second arm is why the test builds a table of 129 buffers rather than asserting the fix on a
three-element one: without it, an implementation that preserves the reserved binding but then hands
the same number to an ordinary buffer passes.

## Docs

- `ASTROBOT_LINUX_HANDOFF_2026_07_19.md` — the dated result with the full chain, and **four new
  `## Ruled out` rows**: the tile index, the mis-resolved descriptor, the genuinely-cyclic list
  (#3214's candidates 1, 2 and 4), and "a trip bound is a candidate fix".
- `tools/AGENTS.md`, `src/gpu/diagnostics/AGENTS.md` — the new instrument, what it covers and how it
  can lie.
- `BLOG.md` — one entry, three pictures.

Refs #3193, #1809, #1459

Also adds instrument trap 246 (a byte-size diagnostic selector cannot tell a 33,177,600-byte buffer from a 3840x2160x4 image), so this PR touches a numbered table — run `check_merge_result.py` immediately before merging.


mattias800:
## Snapshot guards, and a pre-existing `gris-gameplay` failure this change does NOT cause

Ran three guarded titles locally on the PR head (`tools/snapshot/snapshot.py check`, this box):

| guard | result |
| --- | --- |
| `messenger-scene` | **OK** — 480x270, frames 29, qualifying 29, richest 32 colors, pixel_changes 27, structural 29/29, nonblack 29/29 |
| `dead-cells-gameplay` | **OK** — 960x540, frames 44, qualifying 44, richest 20,964 colors, pixel_changes 43, structural 44/44, nonblack 44/44 |
| `gris-gameplay` | **FAIL** — `qualifying frames 59 < 60; structural matches 0 < 60 (minimum SSIM 0.85)` |

### The GRIS failure is not this change

Reproduced twice on the PR head, then **A/B'd inside one binary**: I reverted only
`assign_convention_bindings` to master's renumber-everything behaviour (the same edit as the first
mutation arm), rebuilt, and re-ran the same guard. It still fails, with the same signature —
`qualifying frames 58 < 60; structural matches 0 < 60`. The source was restored and the unit tests
re-run green afterwards.

| arm | `gris-gameplay` |
| --- | --- |
| PR head | FAIL, qualifying 59 < 60, structural 0 |
| PR head with `assign_convention_bindings` reverted to master's behaviour | FAIL, qualifying 58 < 60, structural 0 |

So the guard's failure is upstream of this diff. What it is has **not** been established here, and I am
deliberately not claiming "master is broken": the arm reverts one function on this branch, not the
branch itself, and a stale baseline or this box's throughput are both live explanations. Two facts
worth carrying to whoever picks it up:

- **The frames render correctly.** `distinct_rgb_colors` 482–1,115, `nonblack_ratio` 1.0, and the
  retained review images show GRIS's character on the pale background, drawn cleanly.
- **The content is from the wrong point in the route.** `best_structural_similarity` sits at
  0.12–0.14 against every baseline signature — not a degraded match, a different scene. The window
  begins at `elapsed_seconds` 155.4 with `present_count` 34,834. That reads as a route/pace shift
  rather than a rendering regression.

I have not filed an issue for it because I cannot yet separate "the guard drifted" from "this box is
slow today" from "something on master moved the route", and a tracker row that says the wrong one of
those is worse than none. Happy to chase it if the reviewer wants it before merge.

mattias800 (commented):
**APPROVED.**

A four-line behavioural change that removes a deterministic GPU device reset, with the causal chain
established end to end rather than inferred. This is the strongest result of the day.

## The fix is right, and the common case is provably untouched

`claim()` diverges from master **only** when the running counter reaches 127, and `reserved()` only
fires on a resource already carrying that binding. Every table smaller than that — which is every
ordinary table — is renumbered byte-identically to before. That is the property that makes a change
to binding policy safe to land, and it is worth stating explicitly because binding policy is exactly
the code whose defects surface a long way from home, as this one did.

I checked the discriminator rather than taking it: the internal GDS resource is constructed locally
at `gpu_executor.cpp:8107` with `gds.binding = kComputeInternalGdsBinding` and pushed there, so at
the moment `assign_convention_bindings` runs it is the only thing in the table carrying 127. The
binding-only test also matches the existing precedent at `:8411`.

**One non-blocking hardening.** `gpu_capture.cpp:820` uses a stricter test for the same object —
`is_compute_internal_gds` additionally requires `gpu_addr == 0`, `cls == ConstantBuffer`,
`size == 64 KiB`, `stride == 4`. Since the failure mode of a false positive here is silent (a
guest resource that happened to arrive at 127 would be frozen there and skipped from renumbering,
which is not obviously wrong until much later), it would cost nothing to reuse that conjunction.
Not a merge condition — the resource is constructed three lines away — but worth a follow-up.

## The elimination is genuine, and each arm could have said otherwise

What makes this more than a lucky guess is that all three rejected candidates were killed by
measurements that had a visible alternative outcome:

- **tile index** — all 32,400 head dwords zero, so no index selects a terminator. A mix of
  terminators and indices with only some tiles running away would have supported it.
- **mis-resolved descriptor** — `declared == uploaded_dw*4`, exact sizes, and **`oob-starts=0`** on
  every scan. A wrong `num_records` shows as non-zero `oob-starts`; a wrong base shows as a
  populated arena elsewhere.
- **genuinely cyclic** — `cycle-nodes=1` with `link=0 -> next=0` is the degenerate self-loop an
  all-zero pool always yields, not a ring. A GTA-V-style corruption shows a multi-node ring.

And the provenance run is a negative about the *guest* rather than about the instrument: `last
writer NONE` with a 1,127-entry history well under its 65,536 cap and all four recorders firing.
Checking that the history had not evicted, before quoting an absence, is the discipline this
project's trap list exists to enforce.

The confirming measurement is the one I would have asked for: **`terminating=4147200 cyclic=0` with
exactly 32,400 terminator dwords — one per screen tile**. That is not "the hang went away", it is
the arena containing precisely the structure the shader was looking for.

## The evidence claims match the pictures

I measured the three screenshots rather than trusting the captions. The control is **1 distinct
colour at luminance 255** — a literally uniform white frame, exactly as described. The two "after"
frames carry 6,414 and 8,123 distinct colours with **no** near-white fraction, so they are real
rendered content, not a blown-out or diagnostic frame.

I also opened one. It is a starfield with drifting shards, light shafts and coloured fragments, with
large blown-out geometry and visible clipping — real GPU content, and **not** a world map a person
would recognise as the game. The BLOG entry describes it accurately: its alt-text is literal about
what is on screen, and it closes with "the geometry is blown out white in places and the lighting is
not yet right — that is the next thing." `COMPATIBILITY.md` is untouched and the tracker gets `Refs`,
so no rung is claimed. That is the honest treatment of a fix that removes a crash without yet
producing a correct scene, and #3218 carries the remainder.

## On the reach, which is the real risk

Previously-declined GDS compute now runs **in any title**, not just Astro Bot. I am approving that
deliberately: the charter is explicit that a silently skipped dispatch is a fatal gap rather than an
acceptable outcome, and this defect is the precise demonstration — a declined dispatch produced a
device reset two subsystems away, and the skip line that reported it was reachable in detail only
behind `PROSPER_DBG`'s firehose. Restoring correct behaviour is right, and
`PROSPER_COMPUTE_DESCRIPTOR_DETAIL` fixing that diagnosability gap in the same PR is the correct
companion.

It does mean this is the change to watch in the next full release-time snapshot matrix, beyond the
three guards run here. Noting it rather than gating on it, since snapshots are a release gate and
not a per-PR one.

## On `gris-gameplay`

Not caused by this diff, and the A/B is the right one — reverting only `assign_convention_bindings`
inside one binary still fails with the same signature. It is also **already tracked**: **#3148** is
open and records `gris-gameplay` failing on master. Worth knowing that the signature there is
*different* — #3148 has the title stuck on the publisher splash with `structural matches 0 of 60`,
while this run has the character visible and SSIM 0.12–0.14 against every baseline. That is a
genuinely new data point, so I am adding it to #3148 rather than opening anything. Declining to file
a fresh issue while the cause was unseparated was the right call.

## Verification

Build rc 0, `ctest --no-tests=error` **334/334**, serial, idle GPU, with the trap-243 contention set
re-run and passing. New cases confirmed by name (`link_list_census` #207, `convention_bindings`
#257). Both mutation arms report build rc 0 before their result, and the source was restored and
re-run green afterwards. CI 9/9. `check_merge_result.py` from the branch's own worktree: clean
merge, trap table 1..246 unique and ascending, no base row deleted. No session trailers, no host
paths.

Issue #3026, opened 2026-08-26
The named/array colour-target mirror for slots 0-1 is an unenforced convention, and the obvious place to assert it cannot fail


mattias800:
Follow-up from #3023, raised in review. Not a live defect — a missing guard around one.

## The convention

For colour slots 0 and 1, `DrawItem::color_targets[slot]` is required to mirror the named
`color0_*` / `color1_*` triple. Every producer honours it:

- **Captures** cannot even express a divergence — the wire format carries slots 2 and up only, and
  `restore_legacy_color_target_aliases` re-derives slots 0/1 from the named fields on every load.
- **Live draws** get the same mirror at the single success exit of `realize_draw_item`.
- `render_state.cpp` derives the *named* field from the array, keeping them equal in the other
  direction too.

Nothing checks it. It is a convention every producer happens to follow, and #3023 is what happens
when a consumer quietly starts depending on which of the two it reads: pass grouping read slot 0
from the array while the pass rendered to the named field, and two draws targeting different
addresses merged into one pass.

## Why the obvious guard is worthless

The tempting fix is a `#ifndef NDEBUG` assertion inside `realize_draw_item`. **It would be
tautological.** The mirror is the statement immediately above that function's success exit, so the
assertion would verify that an assignment two lines earlier had happened. It would pass forever and
catch nothing — a check that cannot fail, which is worse than no check because it looks like
coverage.

## Where enforcement would actually bite

Where the two representations are read **apart**:

- **One accessor.** Route every slot-0/1 read through a single function, so `mrt_color_binding`,
  `mrt_same_color_pass`'s `pass_binding`, and `live_renderer`'s `pass_bases[0]` cannot disagree by
  construction. This removes the class rather than detecting it.
- **Or a consumer-side check** at the point of use — assert agreement where the pass target is
  chosen, which is a place a divergence would have had to survive to matter.

The accessor is the better shape but has a real blast radius (`live_renderer` deliberately takes
slot 0 from the named field and slots 1+ from the array, so unifying them means deciding which the
renderer should render to, not just where it reads from). That decision is the actual work here, and
it is why this is a separate issue rather than part of #3023.

## Related

The same "grouped on one identity, acted on with another" shape is live on the resolve path — #3025.

Refs #3015, #3023, #3025

mattias800:
CLAIMING: lane-g | branch fix/issue-3026-color-target-mirror | 2026-09-02T08:08:56Z

mattias800:
[...]

Issue #3219, opened 2026-09-02
Decide which colour-target representation the renderer renders to, or retire the named slot-0/1 aliases


mattias800:
Split out of #3026, which landed the consumer-side check and deliberately did not make this call.

## The decision

Four places read a slot-0 colour target, and they read it three different ways:

| reader | rule | where |
| --- | --- | --- |
| `mrt_color_binding` — the active-binding rule behind the attachment count and feedback detection | ARRAY-first; the named triple only when the array entry is absent | `frontends/shared/rtt/mrt_binding.hpp` |
| `mrt_pass_color_binding` — pass identity, i.e. grouping | NAMED-first when `color0_base` names a surface, else `mrt_color_binding` | same header |
| `live_renderer`'s `pass_bases[0]` — the surface the pass renders to | the RAW named `color0_base` | `frontends/shared/live/live_renderer.cpp` |
| `live_renderer`'s framebuffer extent | the RAW named `color0_width` / `color0_height` | same file |

They agree on every draw any producer emits, because `DrawItem::color_targets[0]`/`[1]` either mirror
the named triple or are absent (#3026 wrote that invariant down as `mrt_color_alias_mirrored` and
checks it at the consumer). They stop agreeing the moment a producer fills the two independently, and
#3026's issue text named the merge as the better shape — with the caveat that it means **deciding
which representation the renderer should render to**, not just where it reads from.

That is this issue.

## Why it cannot be settled by testing today

Measured on the #3026 branch, two mutants of the one line that chooses the pass target
(`const uint64_t base = items[pass_i].color0_base;`), both compiled (build rc=0):

| mutant | result |
| --- | --- |
| **E** — take the base from `mrt_color_binding` (ARRAY-first) | `gpu_capture_render_replay` **red, 19 arms**, including two #3026 added |
| **F** — take the base from `mrt_pass_color_binding` (NAMED-first) | `mrt_binding` and `gpu_capture_render_replay` both **green** |

So the array-first unification is clearly wrong and the test suite says so loudly. The named-first
unification is **indistinguishable from master across the whole suite** — which is the problem, not
the reassurance it looks like. The two differ on exactly one shape:

```
color_targets[0].base = A (non-zero)      // the array names a surface
color0_base           = 0                 // the named triple names none
```

Today grouping compares `A` while `pass_bases[0]` is **0**, so the pass groups on one identity and
renders to nothing. Under mutant F it would render to `A`. No producer emits that shape, no fixture
[...]

Issue #128, opened 2026-07-08
[defect] init-fault dump derefs unchecked g_fault_rip — can kill the process while reporting a tolerated fault


mattias800:
From the #72 history-review backlog. exec_image_linux.cpp (~1290, init-fault report): after an init fn faults, the report path mprotects g_fault_rip's page without checking the result, then unconditionally derefs g_fault_rip-8..+15 (and the entry fn). A wild/null jump — the exact case this diagnostic exists for — turns a tolerated, logged failure into a hard process death; it also strips PROT_WRITE, forcing R|X on an RW span. Fix: guard the byte dumps with the readability probe / re-arm sigsetjmp, and restore the page's original protection. Introduced b51ad10.

mattias800:
CLAIMING: claude-session-896257ae (worktree bughunt-batch) | branch fix/issue-128-initfault-dump | 2026-07-09T00:35:10Z

PR #3220, opened 2026-09-02
gpu: check the slot-0/1 colour-target alias mirror where the pass target is chosen

mattias800:
## What this is

A guard around an unenforced convention, plus the analysis that says why the obvious version of it
would have been worthless and why the tempting alternative is a semantic decision rather than a
refactor. **No behaviour changes.**

## The convention

For colour slots 0 and 1, `DrawItem::color_targets[slot]` must either **equal** the named
`colorN_base` / `colorN_width` / `colorN_height` triple or be **absent** (all three fields zero — the
legacy shape `mrt_color_binding`'s fallback exists for). Every producer honours it:

* `realize_draw_item`'s single success exit and its `OperationRealizationFailure` record;
* the three `DrawItem`↔capture conversions in `gpu_capture.cpp`;
* `restore_legacy_color_target_aliases`, which runs on every capture load — necessarily, because the
  wire format carries slots 2 and up only;
* direct/synthetic callers (the render fixtures, tools) leave the array absent.

Nothing checked it. #3023 is what happens when a consumer starts depending on which of the two it
reads: pass grouping read slot 0 from the array while the pass rendered to the named field, and two
draws targeting different addresses merged into one pass.

## Why the obvious guard is void

An `#ifndef NDEBUG` assertion inside `realize_draw_item` would sit two lines below the mirror it
verifies, so it would pass forever and catch nothing — a check that cannot fail, which is worse than
no check because it reads as coverage. This project's standing rule is that a discriminator which
cannot show its lever moved is void, and that is the purest form of it. It is not what this PR adds.

## The option taken, and what the others would have cost

#3026 named two: **one accessor** (route every slot-0/1 read through a single function, removing the
class) or **a consumer-side check** (assert agreement where the pass target is chosen). This lands
the **consumer-side check**, because the accessor turned out to be a decision about what the renderer
renders to and there is no evidence with which to make it.

Four places read a slot-0 target, three different ways:

| reader | rule |
| --- | --- |
| `mrt_color_binding` — active binding: attachment count, feedback detection | ARRAY-first, named only when the array entry is absent |
| `mrt_pass_color_binding` — pass identity, i.e. grouping | NAMED-first when `color0_base` names a surface |
| `live_renderer`'s `pass_bases[0]` — the surface the pass renders to | the RAW named field |
| `live_renderer`'s framebuffer extent (`native_w`/`native_h`) | the RAW named extent |

Under the convention all four return one triple, for every input — that is the behaviour-preservation
argument, and it is a **test** in this PR (`readers_agree` over the three producer shapes) rather than
a paragraph in this body. Break the convention and they answer three different ways.

Two mutants of the one line that chooses the pass target measure what unifying would cost. Both
compiled — build rc=**0** — so the arms ran against mutated code:

| mutant | `mrt_binding` | `gpu_capture_render_replay` |
| --- | --- | --- |
| **E** — take the base from `mrt_color_binding` (unify on ARRAY-first) | Passed | **Failed, 19 arms**, incl. both this PR adds |
| **F** — take the base from `mrt_pass_color_binding` (unify on NAMED-first) | Passed | Passed |

Array-first is loudly wrong: the render fixtures carry a stale array entry beside a live named
triple, so 19 arms redden. Named-first is **indistinguishable from master across the whole suite** —
which is the problem, not the reassurance it looks like. The two differ on exactly one shape:
`color_targets[0].base` non-zero with `color0_base == 0`, where grouping compares that address while
`pass_bases[0]` is 0. Nothing produces that shape, no fixture builds it, and no capture can carry it,
so adopting F would be an **unverifiable change to what the renderer renders to** — which is exactly
the thing not to do quietly. Filed as **#3219** with the full site list and what each choice implies.

The third option, a producer-side assertion, is the void one above.

## The change

`frontends/shared/rtt/mrt_binding.hpp`:

* `mrt_color_alias_mirrored(draw, slot)` — the convention written down. Absent counts as mirrored;
  slots 2 and up have no named alias, so nothing there can diverge.
* `mrt_check_color_alias_mirror(draw, report)` — the consumer's entry point. Reports the slot and both
  representations through a caller-owned callback, and counts, so this header stays free of I/O and of
  rate-limit policy. `mrt_color_alias_divergences()` exposes the count.
* Two verbatim lifts, no semantic change: `mrt_named_color_alias` (the named triple, previously
  inlined twice in `mrt_color_binding`) and `mrt_pass_color_binding` (grouping's slot-0 rule,
  previously a lambda inside `mrt_same_color_pass`). Lifting the second is what lets a test name the
  three readers and assert they agree.

`frontends/shared/live/live_renderer.cpp`: one pass over `items` in the `pertarget` branch,
immediately above the pass loop — where `base`, `pass_bases[0]` and the framebuffer extent are
chosen. Report-only, bounded at 8 per process, with `PROSPER_MRT_ALIAS_LOG=1` for every occurrence
(the `resolve SKIP` pattern from #3216).

## A measurement worth recording: the fixtures diverge today

#3023 established that a divergence is *inexpressible* in a capture and impossible after
`realize_draw_item`. That is correct and is why its fix is a no-op on real content. It does **not**
generalise to the tree: a fixture that takes a captured item and updates only the named fields leaves
the array holding the capture's original base, and one run of `gpu_capture_render_replay` now reports
**eight** such draws (`array=0x200000 0x0` against a live named triple). The guard is therefore not
dead code on day one, and this is why the invariant is "mirrors **or is absent**" — a strict-equality
check would fire on every legacy and synthetic draw. Recorded in `docs/GRAPHICS.md` § Ruled out
along with the accessor falsification.

## Verification

```
cmake -S prosper -B prosper/build-linux                 # rc=0
cmake --build prosper/build-linux -j6                   # rc=0
ctest --no-tests=error                                  # rc=0, 100% tests passed out of 332
```

Linux, distrobox `ps5ys`. 332 is master's count: the arms extend existing tests, so no new ctest
executable is registered, and the count is unchanged by construction rather than by accident. Both
affected cases confirmed by name — `ctest -N` → `#128 mrt_binding`, `#289 gpu_capture_render_replay`.

### The tests, and what each can catch

`frontends/shared/tests/test_mrt_binding.cpp` — the predicate, against **hand-built** divergent
instances, since no producer can emit one:

* the three producer shapes (exact mirror / absent array / nothing bound) report no divergence, and
  all three readers return one triple on each;
* **A** a stale array entry beside a live named triple — the fixture shape, and #3023's subject;
* **B** the array naming a surface where the named triple names none — the one where *grouping and
  target selection* disagree;
* **C** same address, different extent — a base-only comparison would miss it;
* **D** slot 1 alone, which is the slot the resolve path takes its destination from (#3025);
* the callback reports the right slot and both representations, both slots of one draw are reported
  once each, and the three producer shapes report nothing.

`tests/misc/test_gpu_capture_render.cpp` — through the **live renderer**, because a predicate test
cannot establish that production calls it. A mirrored control renders non-blank content and reports
nothing; a divergent draw is reported; and the load-bearing pair asserts the guard is **detect-only** —
the pass still renders to the named surface with pixels byte-identical to the control's, and the
array's address receives nothing.

### Mutation arms — every one compiled and ran

| mutant | build | `mrt_binding` | `gpu_capture_render_replay` |
| --- | --- | --- | --- |
| **A** — `mrt_color_alias_mirrored` returns `true` (the guard cannot fail) | rc=**0** | rc=1, **10** arms | rc=1, *"the live renderer reports the divergence where it chooses the pass target"* |
| **B** — returns `false` for slots 0/1 (the guard is a constant the other way) | rc=**0** | rc=1, **8** arms | rc=1, *"a mirrored slot-0 alias … reports no divergence"* |
| **C** — the consumer never calls it (guard present, wiring deleted) | rc=**0** | **Passed** | rc=1, *"the live renderer reports the divergence where it chooses the pass target"* |
| **E** — the renderer takes slot 0 from the array | rc=**0** | Passed | rc=1, **19** arms incl. both new ones |
| **F** — the renderer takes slot 0 from `mrt_pass_color_binding` | rc=**0** | Passed | Passed |

**C is the discriminating one.** The predicate test stays green while the renderer arm reddens, which
is what separates "the helper works" from "the decision uses it" — the failure this header already
records under `mrt_direct_serves`, where tests that exercised the helper stayed green while production
was reverted. **E** shows the suite would catch the renderer change the accessor option risks;
**F** shows it would *not* catch the other candidate, which is why that one is #3219 and not this PR.

* **Affected:** a new bounded `[mrt-alias]` line on stderr when a draw's slot-0/1 array entry neither
  mirrors its named triple nor is absent. `gpu_capture_render_replay` prints eight of them, which are
  true statements about its fixtures.
* **Unaffected — everything else, by construction.** No pass groups differently, no pass renders to a
  different surface or extent, no attachment count changes, feedback detection is untouched, the
  resolve path is untouched, and no `PROSPER_*` lever changes meaning. The two lifted functions are
  verbatim moves. The only new state is a process-wide counter that nothing reads outside a test.

## Risks

Low. The check is a comparison of three integer pairs per draw, twice, with no branch into rendering.
The one thing a reader might not expect is the new default-visible `[mrt-alias]` line; it is bounded
at 8 per process and says what it means. If the invariant is ever violated at scale by a future
producer, the cost is eight lines and a counter, not a behaviour change — which is the deliberate
half: this PR detects the class and #3219 owns the decision that would remove it.

Refs #3026, #3023, #3025, #3216

### DIFF
--- a/prosper/docs/GRAPHICS.md
+++ b/prosper/docs/GRAPHICS.md
@@ -507,9 +507,33 @@ decode, shader recompiler, Vulkan backend — all unit-testable in isolation) pr
 
 ## Ruled out
 
-Cross-title falsifications for the **present / publish path** (what reaches the screen once passes
-have rendered). One line each: the dead hypothesis, the evidence, the link. Extend this rather than
+Cross-title falsifications for the **present / publish path** — what reaches the screen once passes
+have rendered, together with the pass identity that decides which surface receives those pixels in
+the first place. One line each: the dead hypothesis, the evidence, the link. Extend this rather than
 re-deriving — and read it before forming a hypothesis about a frozen, black, or missing frame.
+
+- **The three slot-0 colour-target readers cannot be merged into one accessor as a refactor — the
+  merge is a semantic decision, and it decides what the renderer renders to.** `mrt_color_binding`
+  (array-first, the active-binding rule behind the attachment count and feedback detection),
+  `mrt_pass_color_binding` (named-first, pass identity) and `live_renderer`'s `pass_bases[0]` (the
+  raw named field) agree on every draw any producer emits, so unifying them *looks* free. On a draw
+  whose named and array representations disagree they answer three different ways, and the sharpest
+  case is not the obvious one: with the array naming a surface and the named triple naming none,
+  grouping compares that address while `pass_bases[0]` is **0**, so the winner decides whether the
+  pass renders to the array's surface or to nothing at all. Hand-built instances for all three
+  divergence shapes are in `frontends/shared/tests/test_mrt_binding.cpp`; no producer can emit one,
+  so there is no evidence to pick a winner with. #3026 landed the consumer-side check instead and
+  filed the decision separately. #3026.
+- **"A named/array divergence is inexpressible" is true of captures and live draws and NOT of the
+  tree as a whole — the render fixtures emit them today.** #3023 established that captures cannot
+  carry a divergence (the wire format holds slots 2 and up only, and
+  `restore_legacy_color_target_aliases` re-derives 0/1 from the named triple on load) and that
+  `realize_draw_item` mirrors at its single success exit. That is correct and is why the #3023 fix
+  is a no-op on real content. It does not generalise: a fixture that takes a captured item and
+  updates only the named fields leaves the array holding the capture's original base, and one run of
+  `gpu_capture_render_replay` reports **eight** such draws (`array=0x200000 0x0` against a live named
+  triple). So the enforceable invariant is "the array mirrors the named triple **or is absent**", not
+  equality — a strict-equality guard fires on every legacy and synthetic draw. #3026.
 
 - **A frozen frame with a live guest is not necessarily a guest or a draw problem — check the publish
   gate first.** Sonic Frontiers (PPSA03831) looked dead from t≈140 s with `frame_seq` frozen at 2,081,
--- a/prosper/frontends/shared/live/live_renderer.cpp
+++ b/prosper/frontends/shared/live/live_renderer.cpp
@@ -7573,6 +7573,46 @@ void register_live_renderer(const std::string& frame_dir, bool dump_bmps_request
                 auto active_color_count = [](const prosper::gpu::DrawItem& draw) {
                     return prosper::frontend::mrt_active_color_count(draw, mrt_format_defined);
                 };
+                // #3026 -- the slot-0/1 alias mirror, checked HERE, at the consumer.
+                //
+                // `DrawItem::color_targets[0]`/`[1]` mirror the named `color0_*`/`color1_*` triples
+                // or are absent; every producer honours that and nothing enforced it. The moment
+                // one does not, the readers below answer differently: this loop's pass target is
+                // the RAW named field (`base`, and `native_w`/`native_h` for the framebuffer
+                // extent), grouping is named-first (`mrt_pass_color_binding`), and the
+                // active-binding rule behind the attachment count and feedback detection is
+                // array-first (`mrt_color_binding`). #3023 is what one such disagreement cost: two
+                // draws rendering to different addresses grouped into one pass, and the second
+                // draw's surface was discarded with no pass, no published pixels and no diagnostic.
+                //
+                // It is checked here and not beside the assignments that establish it because an
+                // assertion at `realize_draw_item`'s success exit would sit two lines below its own
+                // mirror and could never fail. This is the place a divergence would have had to
+                // survive to matter.
+                //
+                // Report-only, deliberately. Which representation should win is undecided, and
+                // deciding it here would change the surface the renderer renders to.
+                for (const auto& alias_draw : items) {
+                    prosper::frontend::mrt_check_color_alias_mirror(
+                        alias_draw,
+                        [](uint32_t slot,
+                           const prosper::gpu::DrawItem::ColorTargetBinding& carried,
+                           const prosper::gpu::DrawItem::ColorTargetBinding& named,
+                           uint64_t ordinal) {
+                            static const bool alias_log =
+                                PROSPER_ENV_VALUE("PROSPER_MRT_ALIAS_LOG") != nullptr;
+                            if (ordinal > 8 && !alias_log) return;
+                            fprintf(stderr,
+                                    "[mrt-alias] c%u array=0x%llx %ux%u vs named=0x%llx %ux%u -- "
+                                    "the pass renders to the NAMED surface; grouping, feedback and "
+                                    "the attachment count may read the other one (#3026) x%llu%s\n",
+                                    slot, (unsigned long long)carried.base, carried.width,
+                                    carried.height, (unsigned long long)named.base, named.width,
+                                    named.height, (unsigned long long)ordinal,
+                                    ordinal == 8 && !alias_log
+                                        ? "  (further reports need PROSPER_MRT_ALIAS_LOG=1)" : "");
+                        });
+                }
                 size_t pass_i = 0;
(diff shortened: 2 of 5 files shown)


mattias800 (commented):
**APPROVED.**

## The reasoning behind option 2 is the most valuable thing here

The two unification mutants are what decide this, and the lane read them the right way round.

Mutant **E** (unify array-first) reddening 19 arms is what proves the suite *can* see a change to
what the renderer renders to. Without it, mutant **F**'s silence would mean nothing — a discriminator
that cannot show its lever moved is void, and "no test failed" from an instrument that cannot fail
is not evidence. With E on the record, F's result is meaningful, and what it means is the opposite of
reassuring: **named-first is indistinguishable from master across the entire suite**, so adopting it
would be an unverifiable change to the surface the renderer renders to.

That is exactly the shape of the change that caused #3205 — an identity predicate altered because it
looked obviously right, no test able to see it, and a visible *Grand Theft Auto V* regression found
only by a 32-commit hand bisect. Declining to make it, and filing #3219 with the site list and what
each option implies, is the correct outcome and better than a green PR would have been.

## The premise correction is a real find

#3026 says a divergence is inexpressible. That holds for captures and for anything through
`realize_draw_item` — but **the render fixtures emit divergent draws today**: a fixture that updates
only the named fields of a captured item leaves the array holding the capture's original base, and
one run reports eight. Two consequences the PR draws correctly:

- the guard is **not dead code**, which is the difference between this and the tautological
  `realize_draw_item` assertion the issue rightly rejects;
- the enforceable invariant is **"mirrors or is absent"**, not equality. A strict-equality check
  would have fired on every legacy and synthetic draw, and would have been reverted within a day as
  noise. `mrt_color_alias_mirrored` encodes it correctly: `slot > 1` has no named alias and passes,
  an all-zero array entry is absent rather than divergent, and only a *present and different* entry
  counts.

## Mutant C is the arm that earns the approval

The predicate test staying green while the renderer arm reddens is what separates "the helper works"
from "the decision uses it". That distinction is not hypothetical in this file — the header records
`mrt_direct_serves`, where tests exercising the helper instead of the decision stayed green while
production was reverted. Exposing `mrt_color_alias_divergences()` so a test can assert the *consumer*
observed a hand-built divergence, and citing the prior instance in the comment, is institutional
memory applied rather than restated.

The four divergent shapes are well chosen — particularly C (same address, different extent), which a
base-only check would miss, and D (slot 1 alone), which is the slot #3216's resolve path takes its
destination from.

## Report-only, and verified to be

The renderer arm's load-bearing pair asserts the guard is **detect-only**: the pass still renders to
the named surface, pixels byte-identical to a mirrored control, and the array's address receives
nothing. That is the assertion that matters for a change landing on the render path, and it is the
right one.

Rate limiting at 8 with `PROSPER_MRT_ALIAS_LOG` to lift it matches the convention #3216 and #3211
established this week.

## Two notes, neither blocking

The check adds one linear pass over `items` per submit. The body is a handful of integer compares so
it is negligible against anything else in that function, but it is new work on the render path and
worth knowing about if this file is ever profiled.

It could not be run against a live title (the GPU was held). So whether any *production* draw trips
the guard is untested — the eight observed divergences are all fixture-side. The blast radius if it
does is eight stderr lines, which is why this is a note and not a condition.

## Verification

Build rc 0, `ctest --no-tests=error` 100% of **332** — master's count, since the arms extend existing
targets; both confirmed present by name rather than inferred from the total. All five mutants report
build rc 0 before their test result, so none is a baseline re-run in disguise. CI 9 SUCCESS,
`mergeable=CLEAN`.

`Refs #3026` rather than `Fixes` is right, since the accessor half is now #3219. I will close #3026
manually with a pointer and clear its label.

Issue #3010, opened 2026-08-26
prosper-app: the --fps HUD reads "no frames published yet" for every real game boot — GPU present never feeds the rate counter


mattias800:
The `--fps` HUD reads `-- fps` / `no frames published yet` for the entire life of a real game boot in
`prosper-app`, while the app's own stderr counter on the same run reports a healthy 60 fps. Observed
interactively on Windows/NVIDIA (RTX 4090) with *The Messenger* and *Grand Theft Auto V* on master
`36804ff7`, but the cause is platform-independent.

## Failure scenario

1. `prosper-app --dump <DUMP_ROOT>/PPSA24651-app0 --no-pick --fps`
2. The title boots and presents thousands of frames. stderr prints
   `[app] 64.0 fps (7560 frames, gpu-present)`.
3. The on-screen HUD says `-- fps` and `no frames published yet`, permanently.

## Cause

`fps_hud_lines()` prints that text when `FrameRate::measured` is false
(`prosper/frontends/prosper-app/fps_hud.hpp:50`). `measured` is derived from the present-rate
counters, and those have exactly **one** feed point:
`note_present_publication()`, called only from `present_write_frame()`
(`prosper/src/gpu/present/videoout_present.cpp:53`).

GPU present (#1270) bypasses that function on both ends:

- `prosper/frontends/shared/live/live_renderer.cpp:8910` — when `gpu_present_active()` is true and the
  front-buffer blit succeeds, the renderer *"SKIP[s] the CPU readback+reupload entirely"*, so it never
  publishes a CPU frame and never calls `present_write_frame()`.
- `prosper/frontends/prosper-app/main.cpp:2425` — the app's `if (vk.gpu_present)` branch blits the
  renderer's image to the swapchain directly, and does not call it either.

The rate is therefore never fed, and `main.cpp:2413` formats a HUD from an unmeasured window every
second. `main.cpp:2455`'s stderr line is unaffected because it counts with its own local `shown`
counter — hence two counters on one run disagreeing.

**This is the default path for every real game.** `request_gpu_present()`
(`prosper/frontends/prosper-app/present_policy.hpp:21-24`) returns true whenever a dump is supplied
and it is not the test pattern, so `--fps` is dead for every ordinary boot on every platform, not just
Windows.

## Why no test or guard caught it

The comment at `live_renderer.cpp:8904-8906` states it outright: `gpu_present_active()` *"is false in
[...]

PR #110, opened 2026-07-08
fix(hle/mem): sceKernelAllocateDirectMemory honors searchStart/searchEnd

mattias800:
Fixes #108

`k_alloc_dmem` dropped its `searchStart`/`searchEnd` args (a0/a1) and always first-fit from the pool base. A guest that partitions physical memory by window — the standard Orbis/Prospero pattern (a GPU-visible pool above a CPU pool, etc.) — could receive a phys offset **outside** the window it requested; an assert or bitmap indexed by `(phys - searchStart)` then runs out of range.

`dmem_take` now accepts an optional `[lo, hi)` window and clips each free gap to it. The two non-ranged callers (`AllocateMainDirectMemory`, the Ampr push-map) keep the whole-pool default; `AllocateDirectMemory` passes its window and returns `ENOMEM` when nothing fits inside it. Naturally pairs with the `dmem_largest_free` range walk added in #99.

**Verification:** `test_dmem` extended — a windowed allocation lands inside `[searchStart, searchEnd)`, and a window too small for the request returns `ENOMEM` rather than the old out-of-window base fallback. Full ctest **50/50** (Linux; the dmem test is `#ifdef __linux__`-guarded and skips on the Windows/MinGW build where the memory HLE doesn't exist — the fix applied in #100).

Lane: `area:messenger` (shared kernel infra).

PR #3223, opened 2026-09-02
present: give the mostly-unchanged verdict a form windowed callers can satisfy

### LINKED ISSUE #3027: present_frame_rate: frame_rate_is_mostly_unchanged requires callers to report active_fraction, but frame_rate_between cannot fill it — the contract is unsatisfiable for every windowed caller
`present_frame_rate.hpp:330` states a contract on `frame_rate_is_mostly_unchanged`:

> `active_fraction` is what separates (a) from (b): a static menu still produced frames before it
> arrived, so its active share is non-zero; a title that produced nothing has an absent typical rate
> and 0% active. **Any caller that reports this predicate must report that alongside it.**

But `frame_rate_between` — the windowed constructor every live/rolling caller must use — does not fill
it. `present_frame_rate.hpp:290` says so explicitly:

> `typical_fps` and `active_fraction` are NOT filled in here, and `typical_measured` stays false

So the contract is **unsatisfiable for any windowed caller**, and both current callers of the predicate
(the `--fps` HUD lines in `fps_hud.hpp`, and the stderr rate lines added in #3010) therefore violate a
documented requirement they have no way to meet. This is not a defect in either caller; it is a
contract that cannot be honoured as written.

## Why it matters rather than being pedantry

The distinction the contract protects is real, and the header is right that it is the one people get
wrong: "the picture is not changing" reads as a hang, when a static menu is the benign case. Without
`active_fraction` a reader has to infer producing-vs-not from something else.

In practice the `--fps` stderr line does disambiguate, by printing the presented rate on the same line
— a non-zero presented rate IS evidence frames are arriving. But that is an accident of that call
site's formatting, not something the contract asked for or can check.

## Options

1. Have `frame_rate_between` fill `active_fraction` from the two snapshots' `active_seconds` (both
   snapshots carry it — `PresentRateSnapshot::active_seconds`), which looks derivable for a window.
2. Narrow the contract to say the requirement binds only on callers using
   `frame_rate_since_first_publication`, and name what a windowed caller must print instead.
3. Give the predicate an overload that cannot be called without the disambiguating figure.

(1) looks right if a windowed active share is meaningful; (2) is honest if it is not. Either way the
header should stop asking for something the type cannot provide, because the current wording makes
every correct caller look non-compliant and gives a reviewer a finding that cannot be actioned.

mattias800:
Fixes #3027.

## The unsatisfiable contract

`present_frame_rate.hpp` required, of `frame_rate_is_mostly_unchanged`:

> `active_fraction` is what separates (a) from (b) … **Any caller that reports this predicate must
> report that alongside it.**

The distinction is real and is the one people get wrong: "the picture is not changing" reads as a
hang, when a static menu — publishing the same correct image — is the benign case. Case (a) is the
R-Type Delta shape (#2783); case (b) is a rung-6 title on its own title screen.

But `frame_rate_between`, the constructor every rolling/live caller must use, does not fill
`active_fraction` and says so in its own comment. So the requirement was **unsatisfiable for both
windowed callers** — the `--fps` HUD line and the app's stderr rate lines from #3010. Neither caller
was at fault; the contract asked for something the type could not provide. In practice the stderr
line disambiguated by printing the presented rate on the same line, which is an accident of that call
site's formatting rather than anything the contract asked for or could check.

## The option chosen, and what the others would have cost

**Option 1 — have `frame_rate_between` fill `active_fraction` from the two snapshots'
`active_seconds` — is falsified, not merely imprecise.** Measured on the counter itself and now
pinned by a test arm:

| constructed run | window | naive differenced active share | truth |
| --- | --- | --- | --- |
| 201 pubs at 1 fps, then 1000 at 60 fps | 16.7 s | **-1100%** (`active_seconds` 200.000 → 16.650) | ill-formed |
| 100 pubs at 1 fps, then 6000 at 60 fps | 100 s | **1.0% active** | 6000 new frames at a perfect 60 fps |
| 6000 pubs at 60 fps, then 100 at 1 fps | 100 s | **0.02% active** | 100 new frames at a healthy 1 fps |

`active_seconds` is a whole-run **recomputation** against the *current* median interval, not an
accumulator: the cutoff that decides which intervals count moves as the run goes on, so the quantity
subtracted at the near end is not the quantity added at the far one. The negative case proves the
subtraction is ill-formed; the two positive cases are the dangerous ones, because they are plausible
and both read as exactly the "produced nothing" shape the metric exists to flag.

**Option 2 — narrow the contract to whole-run callers and name what a windowed caller must print
instead — would have cost the distinction itself.** There is nothing a windowed caller could print
in its place, because *no statistic computed from the window can separate the two cases*: within one
second a static menu and a frozen renderer are byte-identical, since in both cases nothing changed
during that second. The question "has this title ever produced frames?" is a property of the **run**.
That is also why option 1 was doomed independently of the arithmetic.

**Option 3 — an overload that cannot be called without the disambiguating figure — is what landed**,
because the figure has to come from the cumulative snapshot and making it a *parameter* is the only
way to stop a caller omitting it.

## Behavioural contract afterwards

- `UnchangedPicture unchanged_picture(const FrameRate& window, const PresentRateSnapshot& run, …)`
  returns a three-way `PictureChange` — `changing` / `static_picture` / `nothing_produced` — together
  with the run's typical rate, active share and distinct count. The verdict and its evidence are one
  value, so a caller cannot hold the first without the second.
- `nothing_produced` keys on the run having **no typical rate**, which is literally the "absent
  typical rate and 0% active" the header names, rather than on a threshold somebody tuned. The
  borderline case (two distinct frames far apart) is left legible rather than arbitrated: the line
  carries the rate, the share *and* the population.
- `std::string format_unchanged_picture(const UnchangedPicture&)` is the one line callers print, and
  it always carries the run figures:
  ```
  picture not changing; run so far 59.4 fps, 96% active, 1204 distinct
  picture not changing; run so far -- fps, 0% active, 1 distinct (nothing produced yet)
  ```
  Empty for `changing`. The absent rate is `--`, never `0.0`.
- `FrameRate::active_fraction_measured` makes "is this fraction a measurement or an unset default" a
  fact in the value. Without it `0` means both "this run produced nothing" (load-bearing) and "nobody
  filled this in" — which is the ambiguity this issue is.
- `frame_rate_is_mostly_unchanged` is unchanged in behaviour and stays the **whole-run** callers'
  (`tools/screenshot`'s summary line and its manifest); both already print `active_fraction` on the
  spot, so they satisfy the contract as written. The header now says so, and says windowed callers
  must use `unchanged_picture`.
- `format_frame_rate_short` no longer prints `0% active` — a claim about the run — for a rate whose
  fraction was never measured. A window gets `-- fps  1920x1080`.
- The `--fps` HUD and both app stderr lines now report which of the two cases they are looking at,
  and classify against the same snapshot so they cannot disagree with each other.

## Deliberately unaffected

- `tools/screenshot`, its manifest JSON (`mostly_unchanged`, `active_fraction`) and its summary
  paragraph: unchanged output, unchanged semantics.
- `format_frame_rate`'s wording, which `gen_progress_tracker.py`'s `FPS_RECORD_RE` accepts. It is a
  run-summary formatter and is documented as one; no windowed caller passes it a window.
- The thresholds (`min_published = 30`, `max_distinct_fraction = 0.5`), the interval histogram, the
  signature functions, and every rate this module computes. No number changes.
- No new tunable was introduced.

## Risks

- The HUD's third line is longer than the bare `picture not changing` it replaces (~70 chars). It is
  a diagnostic line in an overlay that already prints `60.0 presented   1920x1080   3600 frames`.
- `format_frame_rate_short` on a run rate snapshotted at the exact instant of the first publication
  now prints `-- fps  WxH` instead of `-- fps  0% active  WxH`. Nothing was measured in that case, so
  the shorter form is the honest one; the existing overlay assertion (a 10 s run window) is
  unaffected.
- `unchanged_picture` runs `frame_rate_since_first_publication` per call — pure arithmetic on a
  snapshot, no lock. Called once per HUD window (1 Hz) and once per 60 presented frames.

## Verification

Build and test in the `ps5ys` container, from a clean configure of this branch:

```
cmake -S prosper -B prosper/build-linux                 # exit 0
cmake --build prosper/build-linux -j6                   # exit 0
ctest --no-tests=error                                  # exit 0 — 332/332 passed
```

`prosper-app` is off by default, so it was built explicitly to compile the changed call sites:

```
cmake -S prosper -B prosper/build-app -DPROSPER_APP=ON  # exit 0
cmake --build prosper/build-app --target prosper-app -j6  # exit 0
```

Targeted, by name (`ctest -N`): `#110 prosper_app_fps_hud`, `#237 present_frame_rate` — both Passed.

Docs gate over every tracked `.md`: exit 0. `git diff --check`: clean.

### The arm

`the_unchanged_picture_verdict_separates_static_from_dead` builds two runs whose **last second is the
same measurement** — 60 publications, zero distinct frames, one second — differing only in history:
a static menu reached after 1200 real frames at 60 fps, and a title that never produced a second
frame. It first asserts the premise, that `frame_rate_is_mostly_unchanged` **fires identically on
both**, so nothing about the window can be doing the work; then asserts the two verdicts *differ*,
and that only one of them says nothing was produced. Both runs come from the real `FrameRateCounter`,
so the "produced nothing" snapshot is the accumulator's own answer rather than interval fields left
at zero. The HUD test repeats the same pair through `fps_hud_lines`.

`a_windowed_active_share_cannot_be_differenced` pins the three measurements in the table above, plus
`active_fraction_measured` on both constructors and the `format_frame_rate_short` claim.

### Mutation arms

Both mutations **compiled and ran** (build exit 0), so the red is the assertion failing and not the
patch failing to apply.

| mutation | build | ctest | failing checks |
| --- | --- | --- | --- |
| `unchanged_picture` always returns `static_picture` when unchanged — i.e. the field is populated but cannot discriminate | exit 0 | exit 8 | 4 in `present_frame_rate` (incl. *"the windowed verdict separates the two cases the bare predicate cannot"*), 2 in `prosper_app_fps_hud` |
| `frame_rate_between` implements option 1 — differences `active_seconds` and sets `active_fraction_measured` | exit 0 | exit 8 | 2 in `present_frame_rate`: *"a differenced window says its active fraction is NOT measured"*, *"the compact form drops the '0% active' claim…"* |

The first mutation is the one that matters: it is the "populated field with better cosmetics" failure
mode, and the test rejects it.

## Ruled out

Recorded in `prosper/src/gpu/present/AGENTS.md` (which also gains a map of the folder's second
module, since it described only `videoout_present`):

- Filling a window's `active_fraction` by differencing `active_seconds` — falsified with the three
  measurements above.
- Any windowed statistic separating a static menu from a title that produced nothing — an information
  problem, not a measurement one.

### DIFF
--- a/prosper/frontends/prosper-app/fps_hud.hpp
+++ b/prosper/frontends/prosper-app/fps_hud.hpp
@@ -39,11 +39,15 @@ inline bool fps_window_due(double window_start_seconds, double now_seconds, doub
 
 // The HUD's lines, newest measurement first.
 //
-// `distinct_total` is the run's cumulative distinct-frame count, shown so the rate has a population
-// beside it: "0.4 fps" from three frames is not the same claim as "0.4 fps" from four hundred.
+// `rate` is the ROLLING window; `run` is the cumulative snapshot its later end was read from. Both,
+// because the two answer different questions and the HUD reports both. From `run` come the
+// population beside the rate ("0.4 fps" from three frames is not the same claim as "0.4 fps" from
+// four hundred) and -- the part that cannot come from the window at all -- whether a picture that
+// has stopped changing belongs to a title that HAS produced frames or to one that never did
+// (gpu/present/present_frame_rate.hpp; #3027).
 inline std::vector<std::string> fps_hud_lines(const prosper::gpu::FrameRate& rate,
                                               uint32_t width, uint32_t height,
-                                              uint64_t distinct_total) {
+                                              const prosper::gpu::PresentRateSnapshot& run) {
     char headline[96], detail[160];
     if (!rate.measured) {
         std::snprintf(headline, sizeof headline, "-- fps");
@@ -53,7 +57,7 @@ inline std::vector<std::string> fps_hud_lines(const prosper::gpu::FrameRate& rat
     std::snprintf(headline, sizeof headline, "%.1f fps", rate.distinct_fps);
     std::snprintf(detail, sizeof detail, "%.1f presented   %ux%u   %llu frames",
                   rate.presented_fps, width, height,
-                  static_cast<unsigned long long>(distinct_total));
+                  static_cast<unsigned long long>(run.distinct));
     std::vector<std::string> lines{headline, detail};
     // Stated in words, not left to be inferred from two numbers. The whole failure mode is a reader
     // taking the healthy-looking number, so this has to be prose rather than arithmetic.
@@ -62,8 +66,15 @@ inline std::vector<std::string> fps_hud_lines(const prosper::gpu::FrameRate& rat
     // frame are indistinguishable from this metric by construction, and an earlier wording asserted
     // the second -- which fired on a rung-6 title's own title screen. The HUD reports what it can
     // see.
-    if (prosper::gpu::frame_rate_is_mostly_unchanged(rate))
-        lines.emplace_back("picture not changing");
+    //
+    // What it CAN see, once the run snapshot is in hand, is the one thing that separates those two
+    // for a reader: whether this title has produced frames at all. The window cannot say -- inside
+    // one second a menu and a freeze are identical -- so the verdict is classified against the run
+    // and printed with the run's own rate, active share and distinct count beside it. That figure is
+    // required by present_frame_rate.hpp and used to be unprintable here (#3027).
+    const std::string note =
+        prosper::gpu::format_unchanged_picture(prosper::gpu::unchanged_picture(rate, run));
+    if (!note.empty()) lines.emplace_back(note);
     return lines;
 }
 
--- a/prosper/frontends/prosper-app/main.cpp
+++ b/prosper/frontends/prosper-app/main.cpp
@@ -2027,6 +2027,11 @@ int main(int argc, char** argv) {
     // collapsed must show the collapse. `fpsWindow` is re-based every kFpsWindowSeconds.
     constexpr double kFpsWindowSeconds = 1.0;
     prosper::gpu::PresentRateSnapshot fpsWindow = prosper::gpu::present_rate_snapshot();
+    // The cumulative reading the window was closed against. Kept because "the picture is not
+    // changing" cannot be reported honestly from a window alone: separating a static menu from a
+    // title that has produced nothing is a question about the RUN (#3027). Held here rather than
+    // re-snapshotted at each print site so the HUD and the stderr lines cannot disagree.
+    prosper::gpu::PresentRateSnapshot fpsRun = fpsWindow;
     prosper::gpu::FrameRate fpsRate;
     std::vector<std::string> fpsLines;
     // The extent last presented through the GPU path, for the HUD's resolution field (#3010).
@@ -2729,13 +2734,14 @@ int main(int argc, char** argv) {
                                                   kFpsWindowSeconds)) {
                 fpsRate = prosper::gpu::frame_rate_between(fpsWindow, now);
                 fpsWindow = now;
+                fpsRun = now;
                 // present_frame_width/height are fed by the CPU publish path and read 0 under GPU
                 // present, so the HUD showed "0x0" there. Prefer the extent actually presented.
                 fpsLines = prosper::frontend::fps_hud_lines(
                     fpsRate,
                     gpuPresentedW ? gpuPresentedW : gpu::present_frame_width(),
                     gpuPresentedH ? gpuPresentedH : gpu::present_frame_height(),
-                    now.distinct);
+                    now);
             }
         }
         // Only a HUD that has something to say is passed down; an empty list leaves both present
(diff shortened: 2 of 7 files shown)


mattias800 (commented):
**APPROVED.**

## Option 1 really is falsified, and I checked it against the source rather than the table

This is the claim the whole PR rests on — rejecting the obvious fix needs to be right, not merely
plausible. It is right. `FrameRateCounter::active_seconds()` derives its cutoff from
`typical_interval_seconds()` computed over the **entire accumulated histogram**, then re-attributes
every bucket against that cutoff. So it is a whole-run *recomputation*, not an accumulator: as a run
proceeds and the typical interval moves, buckets migrate across the cutoff and the value can
**decrease**. Differencing two snapshots therefore subtracts a quantity computed against one cutoff
and adds one computed against another, which is not a subtraction of anything.

The `active_seconds` 200.000 → 16.650 case is exactly that, and the −1100% it produces is the
*useful* arm because it is visibly ill-formed. The two plausible ones are the dangerous evidence:
6,000 new frames at a perfect 60 fps reading as **1.0% active**, and 100 new frames at a healthy
1 fps reading as **0.02%** — both landing on precisely the "produced nothing" shape the metric
exists to flag. A wrong answer that looks wrong costs nothing; a wrong answer that looks like the
alarm condition is how a metric gets believed.

Measuring this on the real counter before rejecting the option, rather than reasoning about it from
the header, is the right order of operations.

## The deeper point is the one that makes option 3 necessary

"Has this title ever produced frames?" is a property of the **run**, and no windowed statistic can
supply it — within one second a static menu and a frozen renderer are byte-identical. That kills
option 1 independently of the arithmetic, and it also disposes of option 2: narrowing the prose
would have removed the requirement without replacing the distinction, and there is nothing a
windowed caller could have printed instead.

So the figure has to come from the cumulative snapshot, and making it a **parameter** is the only
construction under which a caller cannot omit it. `unchanged_picture(window, run_snapshot)` enforces
by signature what the header previously asked for in prose — which is the correct answer to
"a contract nobody can satisfy": make it unsatisfiable to *violate* instead.

The three-way `PictureChange` is better than the boolean it replaces, because `changing` /
`static_picture` / `nothing_produced` are the three states a reader actually needs distinguished,
and the middle one is the benign case that the old phrasing made look like a hang.

`FrameRate::active_fraction_measured` turning "measurement or unset default" into a fact carried in
the value is the same idea one level down, and it is what lets `format_frame_rate_short` stop
claiming `0% active` from a window that never measured it — a small line, and the one a reader was
most likely to be misled by.

Holding `fpsRun` beside `fpsWindow` in `main.cpp` rather than re-snapshotting at each print site is
the right call: two call sites sampling the cumulative counter at different instants could disagree,
and a HUD contradicting the stderr line is worse than either alone.

## The test separates the two cases the right way

`the_unchanged_picture_verdict_separates_static_from_dead` builds two runs whose **last second is
the identical measurement** — 60 publications, zero distinct frames, one second — differing only in
history. Then it asserts the *premise* first: that `frame_rate_is_mostly_unchanged` fires identically
on both, so nothing about the window can be doing the discriminating work. That premise assertion is
what makes the subsequent difference meaningful rather than incidental, and it is the step most
tests here skip.

Both runs coming from the real `FrameRateCounter`, so that the "produced nothing" snapshot is the
accumulator's own answer rather than interval fields left at zero, closes the obvious way this test
could have been vacuous.

Mutation arm 1 is the important one and you identified it correctly: `unchanged_picture` always
returning `static_picture` is the "populated field with better cosmetics" failure — the field exists,
it just cannot discriminate — and 6 assertions across both binaries reject it. Arm 2 implementing
option 1 and reddening confirms the falsification is enforced by the suite, not only by the PR body.

## Verification

Build exit 0, `ctest --no-tests=error` 332/332 (master's count — existing binaries extended). The
extra `-DPROSPER_APP=ON` build to compile the changed `main.cpp` call sites is the right precaution,
since the app is off by default and those sites would otherwise be unbuilt by the default configure.
`prosper_app_fps_hud` (#110) and `present_frame_rate` (#237) confirmed by name. CI 9/9. Both mutation
arms report build exit 0 first, so the red is the assertion rather than a failed compile.

The two `## Ruled out` entries in `src/gpu/present/AGENTS.md` are exactly where the next person will
look, and giving that file a map of `present_frame_rate` — which it had never mentioned — is a real
improvement to a folder doc that described only half its contents.

No `BLOG.md` entry is right: no screenshot, no visible progression, internal contract only.

Issue #3208, opened 2026-09-02
terminator-boot guard has zero margin: OK and FAIL differ by one frame of boot jitter


mattias800:
## Summary

The `terminator-boot` snapshot guard has **zero margin** on current master: its verdict flips
between OK and FAIL on **one frame** of ordinary boot-time jitter, on builds whose difference
cannot affect this title's boot timing. It is therefore not usable as a regression signal — a
red `terminator-boot` today says nothing about whether anything regressed.

Measured on 2026-09-01, two full 17-guard sweeps run back-to-back on the same machine:

| sweep | build | frames | qualifying | longest consecutive content matches | required | verdict |
| --- | --- | --- | --- | --- | --- | --- |
| 1 | master before the `execute_item` extraction | 61 | 15 | **14** | 15 | FAIL |
| 2 | master after it (`c1987cdf`) | 61 | 16 | **15** | 15 | OK |

The passing run clears the bar by **exactly zero frames**. Sweep 1 ran while the box was under
load from concurrent build lanes; sweep 2 ran on an idle box. Both boots are healthy — the guard's
own `pixel_changes` (50 and 52) and qualifying-frame counts show the title reaching and holding the
animated main menu in both.

## Why it has no margin

The guard profiles a hard-bounded 0–62 s boot and samples at roughly 1 frame/second (`frames=61`),
requiring **15 consecutive** frames that jointly clear `min_colors=300`, SSIM ≥ 0.85 against
menu-only references, non-black coverage ≥ 0.85, and 480x270 dimensions.

Its own `_note` records that the menu arrival time already drifted once — "the menu appeared at
~30s when adopted but at ~42-43s on current profiling" — which is why the window was rebased to the
bounded boot in the first place. At ~42–43 s inside a 62 s window there are only ~19 sampled menu
frames available, so a **single second** of extra boot latency takes the plateau to 14 and the
guard goes red.

The `review` note claims five frames of margin ("final verify found 20/20 local and 20/20 cross-run
consecutive matches (15 required)"). That margin no longer exists on this machine: the best of two
runs today is 15/15.

## Why this matters beyond one guard

A guard that flips on machine load is worse than an absent one. It produces false regression
reports during exactly the situation where guards are most valuable — a sweep run alongside other
work — and the standing response to a known-flaky guard is to discount it, which also discounts the
true positives it would otherwise catch.

Note the failure is *silent about its own cause*: the FAIL line reads
[...]

PR #3224, opened 2026-09-02
snapshot: record when a baseline was verified and by whom (verified_at / verified_by)

### LINKED ISSUE #3222: snapshot: record when a baseline was verified and by whom (verified_at / verified_by)
## Gap

`snapshots.json` records *what* a baseline asserts, but nothing records **when it was last
verified** or **by whom**. The only trace is the free-text `review` field, which is English prose,
not data — so nothing can sort by it, warn on it, or surface it in a listing. Every one of the 17
current entries happens to contain a date inside that prose, and not one of them is machine-readable.

Two live cases show the cost, both now recorded in
`prosper/tools/snapshot/AGENTS.md` § *Guards that are NOT currently a regression signal* (#3221):

- **`gris-gameplay` (#3148)** — a stale baseline nobody noticed until it failed on master, at two
  *different* points in the route on different runs. It is now excluded pending manual verification.
  Had the manifest carried a machine-readable verification date, `list` could have shown the age of
  that baseline long before it started failing.
- **`terminator-boot` (#3208)** — clears its threshold by exactly zero frames. Its `review` note
  still describes the margin it had when it was adopted, which is no longer true. Prose ages
  silently; a timestamp does not.

## Proposal

Add two structured, sortable fields to a manifest entry:

- `verified_at` — ISO 8601 UTC instant, e.g. `2026-09-02T12:34:56Z`.
- `verified_by` — `"agent"` or `"human"`.

Design points that matter:

1. **Default to `agent`; `human` requires an explicit assertion.** The tool cannot detect who ran
   it, so it must be *told*. Recording a human verification that never happened is far worse than
   the reverse, because the entire value of the field is that a human-reviewed baseline can be
   trusted more. Never infer `human` from a TTY, an interactive terminal, or the absence of an env
   var.
2. **Stamp only where verification actually happens.** `check` must never stamp — checking is not
   verifying, and letting a routine run refresh a date nobody re-reviewed would destroy the
   feature's whole value.
3. **Do not invent history.** The 17 existing entries get no fields. Absence means "never
   recorded", which is itself the useful signal, and is exactly the `gris-gameplay` situation.
   No backfill from `review` prose or git history: a guessed date stamped as a verification date is
   worse than no date, because it reads as fact.
4. **Make it visible in `snapshot.py list`**, including age in days, with absence visually distinct
   from a recent human verification — so a stale or never-verified baseline is obvious at a glance
   rather than only when it fails.

Requested by the project owner.

mattias800:
Fixes #3222

## The gap

`snapshots.json` records *what* a baseline asserts, but nothing about its provenance. The only
trace was the free-text `review` note. Every one of the 17 entries happens to carry a date inside
that prose, and not one of them is machine-readable — so nothing could sort by it, warn on it, or
surface it in a listing.

Two live cases are that gap, both recorded in `tools/snapshot/AGENTS.md`
§ *Guards that are NOT currently a regression signal* (#3221):

- **`gris-gameplay` (#3148)** — a stale baseline nobody noticed until it failed on master, at two
  *different* points in the route on different runs, now excluded pending manual verification.
- **`terminator-boot` (#3208)** — clears its threshold by exactly zero frames while its `review`
  note still describes the margin it had at adoption. Prose ages silently; a timestamp does not.

## The fields

| field | shape | example |
| --- | --- | --- |
| `verified_at` | ISO 8601 instant, UTC, second resolution | `"2026-09-02T09:31:04Z"` |
| `verified_by` | `"agent"` or `"human"` | `"human"` |

`verified_at` is deliberately an *instant*, not the bare date `review` prose carries: a manifest
field exists to be sorted and compared, and this spelling sorts correctly as a plain string.

## Why the default is `agent`

**The tool cannot observe who ran it, so it has to be told, and only in one direction.**
`verified_by` defaults to `agent`; `human` requires
`snapshot.py update NAME --reviewed --verified-by=human`.

Nothing infers `human` from a TTY, from `$USER`, from an interactive terminal, or from the absence
of an environment variable. Every one of those is a proxy for *"somebody was probably there"*, and
a proxy is exactly what would make the field a lie. The asymmetry is the most important decision in
the change: a `human` that never happened is far worse than an `agent` that understates a real
review, because the **only** reason to record the actor is so a human-reviewed baseline can be
trusted further than a machine-approved one. A field that is wrong in the trusted direction
destroys the value of every correct instance of it.

`AGENTS.md` states the operational consequence for agents: do not pass the flag, *not even when a
person asked you to run the command*. They asked for the run; they did not look at the images.

## Where the stamp is written, and why not elsewhere

**`update --reviewed`, inside `approve_exact_candidate` / `approve_content_candidate`.** It is
stamped inside the approval rather than by its caller, so no code path can establish a baseline
without recording who established it.

- **Not `check`.** Checking a baseline is not verifying it. A routine regression run that refreshed
  the date would launder a baseline nobody re-read into one that looks freshly reviewed — which
  would make the field worse than useless, because it would be most wrong on exactly the guards
  that get run most often. That distinction is the whole value of the feature.
- **Not `verify`.** Two independent reasons. `verify` writes no manifest at all — it produces
  `review/NAME/candidate.json` — and it runs *before* the images are inspected, so a stamp there
  would record a review that has not happened yet.
- **`update --reviewed` is the moment the assertion is actually made**: the operator has inspected
  every retained image from both runs and is saying so. It is also the manifest's only writer.

## Existing entries

The 17 shipped baselines get **neither field**, and nothing backfills them. Absence means "never
recorded", which is the useful signal in its own right and precisely `gris-gameplay`'s state. A
date recovered from `review` prose or from git history would read as a recorded fact while being an
inference — worse than no date, because it invites trust.

`snapshots.json` is **unchanged in this PR** (`git diff --stat` below), and a test pins that.

## Visibility

`snapshot.py list` now prints the column beside the name — not at the end of a 110-column line
where nobody scans — with an age in days and a tally footer. `list` also accepts guard names now,
so `list gris-gameplay` answers "when was this last looked at" for one entry.

```
  guard                      last verified          configuration
  some-gameplay              NEVER VERIFIED         dump=PPSAxxxxx-app0 min_colors=12000 min_frames=20 min_changes=20 references=16
  other-gameplay             2026-08-31 human    1d dump=PPSAxxxxx-app0 min_colors=12000 min_frames=20 min_changes=20 references=16
  third-gameplay             2026-07-23 agent   40d dump=PPSAxxxxx-app0 min_colors=12000 min_frames=20 min_changes=20 references=16

  3 guards: 1 never verified, 1 agent-verified, 1 human-verified
```

`NEVER VERIFIED` cannot be mistaken for a date; an unreadable stamp renders as `UNPARSEABLE` rather
than being silently repaired or shown as a date somebody could parse.

## Two supporting changes

- **`entry_fingerprint` excludes both fields**, for the same reason it already excludes `review`:
  `update` writes them, so hashing them would make a *second* `update NAME --reviewed` refuse its
  own candidate with "snapshot configuration changed after verify" and demand a fresh two-capture
  run. Mutation arm 4 below shows this is load-bearing for an existing test, not just a new one.
- **Options are parsed once in `main`.** The CLI is a four-verb `sys.argv` test, not argparse, so
  guard names are positional: `--reviewed` was filtered out inside `cmd_update` and every other
  dash token fell through to `select`, which reported it as an unknown *snapshot name* — or, for
  `list`, was ignored in silence, because `cmd_list` never looked at its arguments at all. That is
  the wrong failure mode for a flag that decides whether a baseline is recorded as human-verified.
  Now an unrecognised option, a value on a boolean option, and a missing value on one that needs a
  value all exit 2 with a message naming what the command accepts.

  `update` additionally requires at least one guard name. An unnamed `update --reviewed` would
  sweep the whole manifest and stamp every entry with a verification nobody performed — the exact
  failure the `verified_by` asymmetry exists to prevent, reached by a typo rather than by a lie.
  Approval is per-guard by construction (each reads its own `review/NAME/candidate.json`), and both
  `README.md` and `AGENTS.md` have always documented the command with a name, so this costs nothing.

## Verification

Python- and docs-only diff; no C++ changed, so no full build is required to exercise it. The
registered ctest case that covers this file is `snapshot_guard_logic`, run by name:

```bash
cmake -S prosper -B prosper/build-linux -DPROSPER_APP=ON        # 334 tests registered
ctest --no-tests=error -R '^snapshot_guard_logic$' --output-on-failure
#   1/1 Test #18: snapshot_guard_logic ....   Passed    0.66 sec
#   100% tests passed out of 1                            rc=0

python3 prosper/tools/snapshot/test_snapshot.py
#   Ran 38 tests ... OK          (18 on master; 20 added)

python3 prosper/tools/snapshot/test_snaps.py    # sibling suite, unaffected: == PASS ==
git diff --check                                # clean
```

`snapshots.json` is untouched:

```
 prosper/tools/snapshot/AGENTS.md        |  61 +++++-
 prosper/tools/snapshot/README.md        |  26 ++-
 prosper/tools/snapshot/snapshot.py      | 228 ++++++++++++++++++++--
 prosper/tools/snapshot/test_snapshot.py | 335 ++++++++++++++++++++++++++++++++
```

### What the new tests assert

Behaviour, not field presence. The load-bearing ones:

- `test_content_approval_defaults_to_agent_verification`, `test_exact_approval_defaults_to_agent_verification`
- `test_human_verification_requires_the_explicit_assertion` — approves under `USER`, `CI`, `TERM`
  and a plausible-looking `PROSPER_SNAPSHOT_VERIFIED_BY=human` and still gets `agent`; only the
  explicit argument yields `human`
- `test_check_never_stamps_a_verification` — `cmd_check` over two entries (one with no baseline,
  one whose capture fails) with `save_manifest` replaced by a raising stub; needs no GPU
- `test_update_writes_the_stamp_that_check_withholds` — the positive control for the arm above, so
  "check leaves it absent" is a fact about `check` and not about the fixture
- `test_no_shipped_entry_carries_a_verification_stamp` — reads the real `snapshots.json`
- `test_repository_manifest_round_trips_byte_for_byte` — load/save is the identity on the real file
- `test_stamping_one_guard_leaves_every_other_entry_untouched` — stamps one entry and asserts the
  on-disk diff is exactly one line re-punctuated plus the two new keys, and that no other entry
  changed
- `test_a_second_approval_is_not_refused_as_a_configuration_change`
- `test_list_distinguishes_never_verified_from_a_recent_human_verification`
- `test_an_unrecognised_option_is_refused_not_read_as_a_guard_name`,
  `test_malformed_option_values_are_refused_in_both_directions`

### Mutation arms

Each was applied to the working tree, the suite run, and the tree restored.

| # | mutation | result |
| --- | --- | --- |
| 1 | `verified_by` default flipped to `human` (in `stamp_verification`, both approvals, `resolve_verified_by`) | **FAILED (failures=8)** — both `..._defaults_to_agent_verification` arms, all five `test_human_verification_requires_the_explicit_assertion` subtests, `test_resolve_verified_by_is_agent_unless_the_option_says_human` |
| 2 | `cmd_check` stamps each selected entry | **FAILED (failures=1)** — `test_check_never_stamps_a_verification` |
| 3 | backfill all 17 entries from their `review` prose dates | **FAILED (failures=2)** — `test_no_shipped_entry_carries_a_verification_stamp`, `test_stamping_one_guard_leaves_every_other_entry_untouched` |
| 4 | verification metadata put back into `entry_fingerprint` | **FAILED (failures=2, errors=1)** — `test_fingerprint_ignores_verification_metadata`, `test_a_second_approval_is_not_refused_as_a_configuration_change`, **and the pre-existing `test_content_candidate_requires_complete_review_evidence`** |

Unmutated: `Ran 38 tests ... OK`.

## Not done here

No `snapshot.py check`/`verify` run against a real title — that needs the GPU and a many-minute
boot, and the change is deliberately designed to be verifiable without one. Nothing in the capture
path changed.

`tools/snapshot/snaps.py` (the human-authored snapshot store) is a separate system with its own
metadata and is untouched; whether it wants the same provenance fields is a separate question.

### DIFF
--- a/prosper/tools/snapshot/AGENTS.md
+++ b/prosper/tools/snapshot/AGENTS.md
@@ -218,10 +218,11 @@ booting anything:
 
 ```
   guard                      last verified          configuration
-  gris-gameplay              NEVER VERIFIED         dump=PPSA09804-app0 min_colors=1200 ...
-  terminator-boot            2026-08-03 human  30d  dump=PPSA25872-app0 min_colors=300 ...
+  some-gameplay              NEVER VERIFIED         dump=PPSAxxxxx-app0 min_colors=12000 min_frames=20 min_changes=20 references=16
+  other-gameplay             2026-08-31 human    1d dump=PPSAxxxxx-app0 min_colors=12000 min_frames=20 min_changes=20 references=16
+  third-gameplay             2026-07-23 agent   40d dump=PPSAxxxxx-app0 min_colors=12000 min_frames=20 min_changes=20 references=16
 
-  17 guards: 15 never verified, 1 agent-verified, 1 human-verified
+  3 guards: 1 never verified, 1 agent-verified, 1 human-verified
 ```
 
 `list` also takes guard names, so `snapshot.py list gris-gameplay` answers "when was this last
--- a/prosper/tools/snapshot/AGENTS.md
+++ b/prosper/tools/snapshot/AGENTS.md
@@ -21,7 +21,8 @@ python3 tools/snapshot/snapshot.py check blasphemous2-gameplay
 
 A red guard is only evidence about your change if the guard itself is sound. Two are not, and a lane
 that treats either as a regression will spend a session chasing its own tools. **Selection is
-positional** — `snapshot.py check <name> [<name>...]`; there is no `--only` flag.
+positional** — `snapshot.py check <name> [<name>...]`; there is no `--only` flag, and an
+unrecognised option is now refused with exit 2 rather than read as a guard name.
 
 - **`gris-gameplay` — do not run it, and do not act on it.** Project-owner decision, 2026-09-02: its
   baseline has not been updated and *GRIS* is to be verified by hand instead. It fails on master
@@ -172,15 +173,65 @@ wrong scene.
 3. For content mode, choose a conservative `min_colors`, require at least two
    qualifying frames, and add `min_pixel_changes` for moving routes.
 4. Run `snapshot.py update NAME --reviewed` only after every image is accepted.
-   Content mode records reviewed luminance references for SSIM plus a
-   conservative non-black coverage floor; exact mode records the identical
-   pixel hash. Never use exact hashes for threaded gameplay merely because one
-   run happened to be stable.
+   Add `--verified-by=human` **only if a person looked at the images**; see
+   *Who verified it, and when* below. Content mode records reviewed luminance
+   references for SSIM plus a conservative non-black coverage floor; exact mode
+   records the identical pixel hash. Never use exact hashes for threaded
+   gameplay merely because one run happened to be stable.
 5. Replace the generic `review` string that `update` wrote with the factual note
    describing what was actually inspected.
 6. Run `snapshot.py check NAME` after approval.
 
 See `README.md` for every manifest field and the current title matrix.
+
+### Who verified it, and when
+
+Two structured fields record the *provenance* of a baseline, as data rather than as the English
+inside `review`:
+
+| field | shape | written by |
+| --- | --- | --- |
+| `verified_at` | ISO 8601 UTC instant, `2026-09-02T12:34:56Z` | `update NAME --reviewed` |
+| `verified_by` | `agent` or `human` | `update NAME --reviewed` |
+
+**`agent` is the default; `human` must be asserted.** The tool cannot observe who ran it, so it has
+to be told: `snapshot.py update NAME --reviewed --verified-by=human`. Nothing infers `human` from a
+TTY, from `$USER`, from an interactive terminal, or from the absence of an environment variable —
+every one of those is a proxy for "somebody was probably there", and a proxy is exactly what would
+make the field a lie. The asymmetry is deliberate and is the whole point: a `human` that never
+happened is far worse than an `agent` that understates a real review, because the only reason to
+record the actor at all is so a human-reviewed baseline can be trusted further than a
+machine-approved one. **So if you are an agent, do not pass the flag — not even when a person asked
+you to run the command.** They asked for the run; they did not look at the images.
+
+**Only `update --reviewed` stamps.** `check` never does, and must never be made to: checking a
+baseline is not verifying it, and a routine regression run that refreshed the date would launder a
+baseline nobody re-read into one that looks freshly reviewed. `verify` does not stamp either — it
+writes no manifest at all, and it runs *before* the images have been inspected, so a stamp there
+would record a review that has not happened yet.
+
+**Absence means "never recorded", and that is the useful signal.** The entries that predate the
+fields carry neither, and nothing backfills them — a date recovered from `review` prose or from git
+history would read as a recorded fact while being an inference, which is worse than no date at all.
+`snapshot.py list` prints the column, so a never-verified or long-stale baseline is visible without
+booting anything:
+
+```
+  guard                      last verified          configuration
+  gris-gameplay              NEVER VERIFIED         dump=PPSA09804-app0 min_colors=1200 ...
+  terminator-boot            2026-08-03 human  30d  dump=PPSA25872-app0 min_colors=300 ...
+
+  17 guards: 15 never verified, 1 agent-verified, 1 human-verified
+```
+
+`list` also takes guard names, so `snapshot.py list gris-gameplay` answers "when was this last
+looked at" for one entry.
+
+Two mechanical notes. Both fields are excluded from `entry_fingerprint`, like `review`, because
+`update` writes them — hashing them would make a second `update NAME --reviewed` refuse its own
+candidate with *"snapshot configuration changed after verify"*. And `update` now requires at least
+one guard name: an unnamed `update --reviewed` would sweep the whole manifest and stamp every entry
+with a verification nobody performed.
 
 ### Three ordering traps that silently produce a bad baseline
 
--- a/prosper/tools/snapshot/README.md
+++ b/prosper/tools/snapshot/README.md
@@ -8,7 +8,8 @@ likeness to reviewed references, stable dimensions, non-black coverage, and opti
 
 Game dumps and captured images are local and gitignored, so this cannot run in
 CI. Only reviewed 16x9 luminance signatures, diagnostic hashes, coarse content
-thresholds, routes, and review notes live in `snapshots.json`.
+thresholds, routes, review notes, and verification provenance live in
+`snapshots.json`.
 
 ## Run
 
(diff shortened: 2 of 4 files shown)


mattias800 (commented):
**APPROVED.**

The design decision I care most about is right, and the reasoning is in the code rather than only in
the PR: `human` is unreachable by accident, and nothing consults a TTY, `$USER`, interactivity or an
environment variable. Each of those is a proxy for "somebody was probably there", and a proxy is
exactly what would make the field a lie. `AGENTS.md` drawing out the operational consequence — an
agent does not pass the flag *even when a person asked for the run*, because they asked for the run,
they did not look at the images — is the sentence that will actually stop this being misused.

`resolve_verified_by` rejecting an unrecognised value instead of falling back is the other half of
that. A typo like `--verified-by=humna` fails loudly rather than silently recording either answer.

## The stamp location is better than what I asked for

I suggested "`update`? `verify`? both?". The lane established it is `update --reviewed` **only**, and
the reason for excluding `verify` is one I had not considered and is decisive: `verify` writes no
manifest at all, and it runs *before* the images are inspected, so a stamp there would record a
review that has not happened yet. Excluding `check` matters for the same reason in the other
direction — a routine run refreshing the date would launder a baseline nobody re-read, and would be
most wrong on precisely the guards that get run most often.

Putting the write inside `approve_exact_candidate` / `approve_content_candidate` rather than in
their caller is the right structural choice: no path can establish a baseline without recording who
established it.

## The two supporting changes are both correct, and the second is a real latent bug

**`entry_fingerprint`**: excluding both fields is not a convenience, it is required — `update` writes
them, so hashing them would make a *second* `update NAME --reviewed` refuse its own candidate with
"snapshot configuration changed after verify". They join `review` for exactly the reason `review` is
already there. Mutation arm 4 reddening a **pre-existing** test is the evidence that this is
load-bearing rather than tidying.

**Option parsing**: the previous behaviour was worse than untidy. Every dash token except
`--reviewed` fell through to `select` as an unknown *snapshot name*, and for `list` was ignored in
silence because `cmd_list` never read its arguments. That is the same class as the `--only` flag I
myself published in #3208 yesterday — a bogus option consumed as a guard name, producing a confusing
non-run instead of an error.

Requiring a guard name on `update` is the safety property I checked independently rather than
accepting: an unnamed `update --reviewed` would sweep the manifest and stamp every entry with a
verification nobody performed. I confirmed nothing would break — every documented invocation already
names a guard (`AGENTS.md:174`, `README.md:129`, the usage banner, and the runtime hint at `:1130`
which interpolates `{s['name']}`), and no CI workflow references snapshots at all.

## No backfill, verified

`snapshots.json` does not appear in the diff at all, so "byte-identical" holds at the file level, and
the round-trip test pins it. `format_verification` cannot let absence read as a date — `NEVER
VERIFIED`, `UNPARSEABLE`, and `?value` for an unrecognised actor are all visually distinct from a
timestamp. That is the property that makes the field trustworthy at a glance.

## Mutation arms

All four are the right ones, and arm 1 (default flipped to `human`, 8 failures across both the
default tests and all five subtests of the explicit-assertion test) is the one that pins the decision
this feature exists for. Arm 3 — backfilling all 17 from `review` prose — reddening
`test_no_shipped_entry_carries_a_verification_stamp` means the no-invented-history rule is enforced
by the suite rather than by convention, which is what I wanted.

## Verification

38 tests (18 on master, 20 added), `snapshot_guard_logic` confirmed by name in both the local build
and CI's Linux job. CI 9/9. Note for the record that local configure registers 334 cases and CI's
registers 330 — a different option set, both green, and the new case verified by name in each rather
than inferred from either total.

## The backfill question is the owner's

All 17 entries carry exactly one parseable `YYYY-MM-DD` in their prose, so a backfill is mechanically
possible. Two reasons it should stay a deliberate decision rather than a follow-up someone does
quietly: those are *approval* dates, and the prose does not answer `verified_by` at all — a backfill
would have to write `agent` for all seventeen (the safe direction) unless the owner remembers
otherwise per title. Relaying, not deciding.

`snaps.py` being left untouched is right, and the observation that it may only want `verified_at` —
being human-authored by construction — is a good one for whenever that is picked up.

Issue #352, opened 2026-07-10
prosper-app P3 polish: cooperative stop, fullscreen/pause, packaging (follow-up to #164)


mattias800:
Follow-up to #164 (closed — the frontend app is delivered and shipping). The app runs a PS5 title in a desktop window with render + audio + controller/keyboard input. What remains is **P3 polish** only, split out here so the tracking issue could close.

**Design of record:** `prosper/docs/FRONTEND_APP.md`.

## Remaining polish items
- [ ] **Cooperative flip-boundary stop.** On window close / Esc, stop the guest at a flip boundary instead of `_exit(0)` mid-frame, so shutdown is clean (no torn frame / detached-thread abandon).
- [ ] **Fullscreen + pause UX.** Toggle fullscreen; pause/resume the guest loop; basic window resize handling for the present blit.
- [x] **Present-mode / latency tuning.** Pick swapchain present mode deliberately (vsync vs. low-latency) rather than the default.
- [x] **Native Windows packaging.** `prosper/scripts/run-windows.ps1` configures, builds, and runs the full native Vulkan/WASAPI/controller frontend with one command (#700).

## Notes
- Non-blocking — headless remains the priority and CI source of truth; this is user-perspective polish.
- All items are frontend-only (`frontends/prosper-app`), no core changes; keep the app default-OFF in CI.
- `area:messenger` (shared/host infrastructure).

mattias800:
CLAIMING: Codex Windows frontend session | branch feat/issue-352-windows-launcher | 2026-07-14T16:05:58Z

Native MinGW prosper-app already delivers Vulkan graphics, WASAPI audio, SDL gamepads, and keyboard input. I will make that path one-command from PowerShell, validate the real native window/audio/pad build, and replace the stale WSLg-only packaging/readme text. I will keep cooperative shutdown/fullscreen as separate follow-up scope unless the launcher exposes a blocker.

mattias800:
Native Windows packaging is implemented on `feat/issue-352-windows-launcher`.

The new `prosper/scripts/run-windows.ps1` is the normal entry point:

```powershell
.\prosper\scripts\run-windows.ps1 .\PPSA24651-app0
```

[...]

PR #47, opened 2026-07-07
feat(ue4): PPSA17942 boots deep into UE4 engine init — SELF variant, real direct-memory model, sceKernelBatchMap, boot API surface

mattias800:
# UE4 title (PPSA17942): from "executes garbage bytes" to deep engine boot with its own log output

Working toward graphics output from the second title. This PR takes it from an instant SIGSEGV ~50 bytes into `eboot` (it was literally executing wrong file bytes) to **UE4's engine bootstrap running end-to-end**: allocator up, 54-thread pool up, its own engine log printing, `ue4commandline.txt` read, config/pak resolution started.

ue4 boot progress

## The walls, in the order they fell

1. **PS5-variant SELF magic (`0xEEF51454`)** — the parser only accepted `0x1D3D154F` (The Messenger's), so the SELF segment map was skipped and every segment mapped from wrong file offsets: the guest executed garbage and the dynamic section read as garbage (the earlier "0 imports"). One-line fix; the dump's segments are plaintext. Now: 2168 imports link (221 cross-module against its `libc.prx`, 1896 HLE stubs).
2. **Direct memory never ran out and never came back** — UE4 sizes pool requests from `GetDirectMemorySize`, allocates halving probe chunks until **ENOMEM**, releases the probes, then allocates its real pools. The bump allocator neither failed nor reclaimed → its 512 GB-arena block bitmap indexed out of range. Now a first-fit allocator over a finite 8 GiB pool with `sceKernelReleaseDirectMemory`.
3. **`sceKernelBatchMap` (NID `2SKEx6bSq-4`)** — identified from the eboot's own assert strings (`"sceKernelBatchMap failed …"`). UE4 commits *every* 64 KB page of its arena through this; the unimplemented stub "succeeded" having mapped nothing, so no engine memory ever existed. Implemented (MAP_DIRECT/UNMAP/PROTECT/MAP_FLEXIBLE/TYPE_PROTECT, 0x20-byte Orbis entries) — ~1000 commits per boot now succeed.
4. **Physical aliasing** — direct memory is now backed by one shared **memfd**, so mapping the same phys offset at two VAs aliases the same bytes (the console's unified-memory contract). Plus `sceKernelVirtualQuery` now reports real commit state, and a logged fault-handler lazy-commit backs the 1-2 pages whose commit protocol we don't fully replicate yet (`CONFIDENCE: MED`, each occurrence visible in the log).
[...]

PR #3227, opened 2026-09-02
host: gate and drain guest GPU submissions before the frontends' _Exit (#3225)

### LINKED ISSUE #3225: prosper-app can exit with a thread wedged in __drm_exec_lock_obj, freezing the host compositor unrecoverably without root
## Summary

`prosper-app` can exit while one of its threads is still inside the kernel's DRM locking path,
leaving a **zombie process with a thread wedged in `__drm_exec_lock_obj`**. The compositor then
cannot make progress and the developer's entire desktop freezes. `SIGKILL` has no effect — the
process is already a zombie and the thread is in uninterruptible sleep — and **recovering requires
root**, so on a machine with no SSH the session is unrecoverable without physical access.

This is the second time in one day that running a title took the host desktop down, by two
*different* mechanisms. The other is a GPU reset killing the screen-locker's greeter, which leaves a
black unresponsive screen with the compositor alive.

## Observed

Authoring session for *The Pathless* (`PPSA01826`) launched through
`tools/snapshot/snaps.py author`, i.e. `prosper-app` with a real window, audio and pad on the
developer's live Plasma/Wayland session.

```
pid 1115629  state ZNl                      <- zombie; SIGTERM and SIGKILL both no-ops
  tid 1115682 state D  wchan __drm_exec_lock_obj
```

- The compositor is **alive but blocked**: `kwin_wayland` is in `Sl` with `wchan 0`, and a
  screenshot request times out (`spectacle` rc=124).
- The panel is fine — `enabled=enabled dpms=On`, session not locked.
- **No GPU fault is logged at all.** Nothing in `journalctl -k` for 30 minutes except an unrelated
  DisplayPort `DPCD read` line. The amdgpu hang detector never fires, because nothing is *hung* —
  it is a held lock, not a stuck job. So there is no self-recovery path and no timeout.
- Still wedged 15 minutes later.

## Why this is worse than a GPU hang

A GPU hang produces `ring timeout` → `GPU reset` → recovery, and the desktop comes back on its own.
This produces **nothing**: no fault, no timeout, no reset, no log line. The only remedy found is to
force a reset by hand as root:

```sh
sudo sh -c 'echo 1 > /sys/kernel/debug/dri/1/amdgpu_gpu_recover'
```

On a machine without `sshd` and without passwordless `sudo`, that means physical access to a TTY, or
the power button. An emulator bug should not be able to reach that state.

## Why it matters beyond one developer's inconvenience

The interactive path is the one this project asks people to use for exactly the work that needs a
human — `snaps.py author` (F6/F7 human-authored snapshots) and the F9/F8 capture triggers all
require `prosper-app` with a real window on a live session. So the failure lands precisely on the
workflow that cannot be run headless, and it costs the whole session rather than the run.

The headless route (`tools/screenshot`, `SDL_VIDEODRIVER=offscreen`) has not been observed doing
this, which is a useful narrowing but not a fix.

## What to investigate

1. **Which DRM object is still locked at exit, and who holds it.** `drm_exec` is the GEM
   reservation-lock helper; a thread parked in `__drm_exec_lock_obj` is waiting on a `dma_resv` it
   cannot get. Determining whether prosper is deadlocking against *itself* (two submissions racing a
   shared BO) or against the compositor is the first fork.
2. **Whether this is shutdown-ordering.** The process was already dead — so the wedged thread is
   running teardown, or a worker that outlived the main thread. A worker still submitting while the
   device/allocator is being destroyed is the obvious candidate.
3. **Whether an in-flight submission is abandoned rather than drained on exit.** If teardown does not
   wait for outstanding submits, a late one can enter the locking path after the objects it names
   have begun to be freed.
4. Whether the RADV/Mesa version here has a known matching issue — this may be a driver bug that
   prosper is provoking rather than a prosper bug outright. Either way prosper should not be able to
   provoke it.

## Reproduce

Not yet reduced to a reliable recipe — it happened once, roughly 25 s into an authoring session.
Anyone attempting to reproduce it should assume **they will lose their desktop**, and should have a
second machine with SSH access, or run it on a host they can power-cycle freely.

## Mitigation until it is understood

Prefer the headless capture path for anything that does not strictly need a human at the keyboard,
and do not launch an interactive session on a machine you cannot reach physically.

mattias800:
Fixes #3225.

## The failure

`prosper-app` can exit leaving a **zombie process with a thread wedged in
`__drm_exec_lock_obj`**. `SIGKILL` is a no-op on it, the compositor's own DRM work blocks behind the
reservation it holds, and the developer's whole desktop freezes. It happened **twice in one day** and
needed a root-forced GPU reset (`echo 1 > /sys/kernel/debug/dri/*/amdgpu_gpu_recover`) — i.e. a
reboot on a machine you cannot reach over SSH. Nothing is logged, because nothing is *hung*: it is a
held lock, so neither the amdgpu watchdog nor any timeout fires and there is no self-recovery path.

## The mechanism, from source

`run_entry()` never reads the cooperative stop flag (`main.cpp:1081`, `game_path.hpp:82`; the only
reader of `prosper_stop_requested()` is prosper-app's own main loop at `main.cpp:2171`), so a booted
guest cannot be joined. The shutdown path therefore was:

```cpp
prosper_request_stop();          // nothing on the guest side observes this
...
if (g_guest_thread.joinable()) {
    g_guest_thread.detach();     // still running
    ...
    std::_Exit(exitCode);        // no drain
}
```

and the `vkDeviceWaitIdle(vk.device)` a few lines below is on the **other** branch — the one taken
only when no guest booted. **Every real game session exited with no drain at all.**

`_Exit` becomes `exit_group()`. A thread already inside a syscall cannot be torn down until it
returns from the kernel; inside an amdgpu command submission it parks on a `dma_resv` nobody will
now release. So this is not an occasional race: there is **no ordering on this path**, and an
in-flight submission is abandoned on *every* close. That it usually costs nothing is luck about
where the guest thread happened to be.

## What this change does

`prosper/src/host/platform/gpu_submit_gate.{hpp,cpp}` — a counted region plus a bounded drain,
beside `lifecycle.hpp` and with the same constraints (no Vulkan, no OS-integration deps, lives in
`prosper_core`):

- `GpuSubmitRegion` — RAII; entering increments a counter **unless shutdown has begun**, and reports
  whether it was admitted.
- `gpu_submit_gate_begin_shutdown()` — after this no new region is admitted, which is what lets the
  count actually reach zero instead of chasing a moving target.
- `gpu_submit_gate_drain(timeout_ms)` — waits for the regions already inside; returns whether it
  drained.

State is one atomic word (bit 0 = shutdown, the rest = region count), so admission is a single CAS
and a *refused* enter never perturbs the count a drain is reading. The admitted path never touches
the mutex; only a leaving region with a waiter present does.

**Chokepoints wired** — every guest-thread `VkQueue` call:

| site | on refusal |
| --- | --- |
| `tests/fixtures/render_runner.h` `render_locked_queue_submit` | `VK_ERROR_DEVICE_LOST`, driver not entered |
| `tests/fixtures/render_runner.h` `render_locked_queue_wait_idle` | `VK_ERROR_DEVICE_LOST`, driver not entered |
| `frontends/shared/live/live_compute.cpp` compute dispatch submit | `VK_ERROR_DEVICE_LOST`; `submission_entered` stays false, so ordinary cleanup runs |
| `frontends/shared/live/live_compute.cpp` post-fence-timeout queue drain | no drain; `completion_proven` stays false, as on the device-lost path |
| `frontends/shared/present/present_blit.cpp` `present_blit_publish` | `false` → the existing CPU-present fallback |

Two of these were not in the original three. `present_blit.hpp:10` documents `present_blit_publish`
as running on the guest thread. The compute **drain** is a wait rather than a submit, so it is not
itself the `__drm_exec_lock_obj` hazard — but leaving it ungated would have broken the claim the
rest of this change rests on: an ungated queue call is invisible to the drain, so the drain could
report success while it was still running and the `vkDeviceWaitIdle` below would then run
concurrently with it. (`render_runner.h` is the live offscreen backend despite its directory —
`tests/fixtures/AGENTS.md`. Guest-thread `vkWaitForFences` calls are deliberately **not** gated:
fences are externally synchronized per-fence, not per-queue, and they do not take the reservation
locks.)

**Frontends.** `prosper-app` now does `begin_shutdown()` → `drain(2000 ms)` → `vkDeviceWaitIdle`
before `_Exit`, and prints to stderr when the drain expires. The wait is **deliberately skipped when
the drain times out**: `vkDeviceWaitIdle` requires host access to every `VkQueue` to be externally
synchronized, and the drain is what provides that here — waiting after a failed drain would both
violate that and be the very block being avoided. `tools/screenshot` has the identical
detach-and-`_exit` shape (`screenshot.cpp:628`, `:975`) and takes the same two steps; `boot_trace`
returns from `main` normally and is untouched.

## What this does NOT close

**It bounds the window; it does not close it.** A thread that is *already inside* `vkQueueSubmit`
when shutdown begins is still waited for, and if it never returns the drain expires and we `_Exit`
exactly as today — no worse, but no better. The real fix is the cooperative stop the existing comment
names as unimplemented: `run_entry()` observes the flag, stops at a submission boundary, and the
thread is **joined**, after which an ordinary drain and teardown are safe (#352). The header, the
call-site comments and `FRONTEND_APP.md` all say this, so a future reader cannot mistake the
mitigation for the fix.

## Risks

- **A false refusal would drop a real submission.** The gate closes only in `begin_shutdown()`, whose
  only non-test callers are the two frontends' exit paths (grep in the commit), so during a running
  title it is one uncontended CAS per submission and can never refuse. If it ever *did* refuse
  early, the cost would be a dropped draw/dispatch/present-blit for the remainder of the process —
  which is why the refusal is bound to an explicit one-way shutdown flag rather than to any
  heuristic.
- **The drain adds up to 2 s to a close**, and only when a guest submission genuinely has not
  returned. Observed cost in verification: zero timeouts in 15 runs.
- **The `vkDeviceWaitIdle` that follows it is UNBOUNDED**, and this is the honest cost of the change.
  On a title that has hung a GPU job — a documented state for GTA V — it returns only when the
  amdgpu watchdog resets the device, adding seconds more. It also protects no teardown, since the
  `_Exit` below destroys no Vulkan object; what it buys is that already-submitted guest work is off
  the device before the process dies. That is a deliberate trade of a bounded, visible delay against
  the unbounded, invisible one this path exists to prevent, and the code says so — deleting it is
  safe and loses only that guarantee. It runs only after a successful drain, so no guest submit can
  be concurrent with it.

## Verification

Container `ps5ys`, Linux, AMD Radeon 8060S (RADV STRIX_HALO). Re-run at head after the
fifth chokepoint was added.

```
cmake -S prosper -B prosper/build-linux -DPROSPER_APP=ON      # rc=0
cmake --build prosper/build-linux -j6                          # rc=0
ctest --no-tests=error                                         # rc=0, 336/336 passed
```
336 = master's 334 plus the two new cases; both verified by name with `ctest -N`
(`#47 gpu_submit_gate`, `#326 submit_gate_wiring`), not by the total.

**Unit coverage** (`gpu_submit_gate`, pure — no Vulkan, no dump, 32 assertions): admits while open,
counts depth on nesting, refuses after `begin_shutdown`, a refused region is not counted,
`begin_shutdown` is idempotent, a region entered *before* shutdown stays admitted, **the drain times
out rather than blocking** while one is held (≥100 ms of a 120 ms budget), the drain returns when
*woken* rather than at its deadline, the count is exact under 8 threads × 5000 enter/leave, and — the
shape the frontend actually uses — **the drain reaches zero against 6 live submitting threads once
the gate is closed**, which is only possible because new regions are refused.

**Wiring coverage** (`submit_gate_wiring`): the backend wrappers really consult the gate. Its
positive arm runs **first and on a real queue** (skipped with a printed note where there is no
device), so a wrapper hard-coded to fail cannot pass; the refusal arm passes `VK_NULL_HANDLE`, so a
clean `VK_ERROR_DEVICE_LOST` is itself proof the call never reached Vulkan.

**Mutation arms** (each built to rc=0 first, so a silently non-applied edit cannot masquerade as a
non-load-bearing test):

| arm | build | `ctest` | effect |
| --- | --- | --- | --- |
| `begin_shutdown()` made a no-op | rc=0 | rc=8, 2/2 red | 9 assertions fail, incl. "drain reaches zero against live submitting threads"; `submit_gate_wiring` aborts — it reached `vkQueueSubmit` with a null queue, which is the wiring assertion doing its job |
| `drain()` returns `true` unconditionally | rc=0 | rc=8, 1/2 red | 5 assertions fail, incl. the timeout and "nothing is in flight after that drain" |

**Runs.** Deliberately **no attempt to reproduce the freeze** — the cause is established from source
and a reproduction costs the owner their desktop.

- `tools/screenshot`, 5 runs (3 × `PPSA02664`, 2 × `PPSA24651`): all `rc=0`, `status=ok`, 3/3
  screenshots each. `pgrep -x screenshot` empty after every run, no `Z`-state process, and **0
  drain-timeout lines** across all five.
- `prosper-app --frames 240`, 3 runs offscreen (`PPSA02664`): `rc=0`, clean
  `[app] shutting down after 240 presented frame(s)`, no leftover process, no zombie, no drain
  timeout. This exercises the changed `_Exit` branch.
- `prosper-app --frames 300`, 1 **interactive windowed** run on the live Wayland session: `rc=0`, no
  leftover process, no zombie, `vulkaninfo --summary` still enumerates the Radeon 8060S afterwards,
  and `journalctl -k` shows no new amdgpu/DRM lines. Desktop unaffected.

Snapshots not run — this changes no rendering path; the gate is open for the entire life of a
running title.

## Docs

`docs/FRONTEND_APP.md` gets the mechanism under its lifecycle and threading sections, plus a new
`## Ruled out`:

- *"The headless path is safer by design"* — falsified. `tools/screenshot` detaches and `_exit`s in
  exactly the same shape; what differs is that no compositor shares the device with it, so a stuck
  reservation has nothing to block.
- *"A GPU hang / driver fault is involved"* — falsified. No fault is logged at all and the hang
  detector never fires; there is no stuck job, only a held lock.

`src/host/platform/AGENTS.md` is new, and spells out the distinction that is easy to get wrong:
`lifecycle.hpp` is a boolean *request*, `gpu_submit_gate.hpp` is a counted region with a drain for
when asking nicely is not enough.

---

## Independent review

Reviewed by a separate agent against the memory-ordering argument, every wired call site's failure
handling, the completeness of the `vkDeviceWaitIdle` justification, and whether each test would fail
without the fix. Verdict **APPROVED**; four findings addressed in `4a815724` and `5201188d`:

- **A refusal was manufacturing a device loss it had not observed.** The compute refusal returned
  `VK_ERROR_DEVICE_LOST` into `vk_ok`, which latched `ctx.device_lost`, printed
  `fatal Vulkan device loss stage=queue-submit … disabling live compute for this process`, and added
  a `reason=queue-submit` row to the decline census — counts this project quotes as evidence. Every
  shutdown would have fabricated one of each. It now breaks directly; `submission_entered` still
  stays false, so cleanup is unchanged.
- **The wiring test's positive arm covered one wrapper, not both**, so a `render_locked_queue_submit`
  hard-coded to fail would have passed. Added an open-gate arm using an empty submit, and confirmed
  by a new mutation arm (hard-code that wrapper to fail → build `rc=0`, `ctest` `rc=8`, exactly the
  new assertion red).
- **The bound is narrower than the header implied.** `__drm_exec_lock_obj` is also reachable through
  the GEM-VA ioctls behind `vkAllocateMemory` / `vkBind*Memory` / `vkFreeMemory`, which the guest
  thread calls ungated. Recorded in the header at `CONFIDENCE: MED` so a recurrence reads as "never
  covered that path" rather than "the gate failed".
- **The risk section understated the close cost** (this section, now corrected), and
  `gpu_submit_gate_leave()` could wrap the count on a mispair and silently break every later drain —
  it now saturates.

The reviewer independently enumerated every `vkQueueSubmit` / `vkQueueWaitIdle` / `vkQueuePresentKHR`
/ `vkQueueBindSparse` / `vkQueueSubmit2` in the tree and confirmed the five gated sites are exactly
the guest-thread ones, and that `library_media.cpp:675` — the one that looks like a worker-thread
submit — is reached only from the app/UI thread. Non-blocking nits left as-is: the negative-timeout
drain mode has no production caller, and the concurrency test's drain assertion is probabilistic
(its comment now says so; the deterministic anchor beside it is `refusals > 0`).

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -600,6 +600,16 @@ if(TARGET prosper_core)
   target_link_libraries(test_lifecycle PRIVATE prosper_core)
   add_test(NAME lifecycle_stop_signal COMMAND test_lifecycle)
 
+  # GPU submit gate (#3225): the counted region + bounded drain prosper-app runs before std::_Exit,
+  # so a detached guest thread cannot be killed inside an amdgpu submission (which parks the task in
+  # __drm_exec_lock_obj and takes the host compositor down with it). Pure — no Vulkan, no dump — so
+  # the admission/refusal/drain-timeout contract is asserted on every host, including CI without a
+  # GPU. Its own timing assertions are one-sided (a lower bound on a timeout, an upper bound far
+  # from the deadline), so it does not go red under host load.
+  add_executable(test_gpu_submit_gate tests/host/platform/test_gpu_submit_gate.cpp)
+  target_link_libraries(test_gpu_submit_gate PRIVATE prosper_core)
+  add_test(NAME gpu_submit_gate COMMAND test_gpu_submit_gate)
+
   # Screenshot evidence classification/JSON is pure and must remain testable without Vulkan or a dump.
   add_executable(test_capture_manifest tests/test_capture_manifest.cpp
                                        tools/screenshot/capture_manifest.cpp
@@ -2811,6 +2821,18 @@ if(TARGET prosper_core)
       target_include_directories(test_gpustate_render PRIVATE frontends)
       target_link_libraries(test_gpustate_render PRIVATE prosper_core Vulkan::Vulkan)
       add_test(NAME gpustate_render COMMAND test_gpustate_render)
+
+      # #3225: the backend's queue wrappers really consult the shutdown gate. The pure gate test
+      # (gpu_submit_gate) asserts the gate's logic; this asserts the wiring, which needs
+      # render_runner.h. Its positive arm uses a real queue when one exists and is skipped with a
+      # printed note when it does not — the refusal arm needs no device at all.
+      add_executable(test_submit_gate_wiring tests/gpu/test_submit_gate_wiring.cpp)
+      # Includes a shared fixture, which resolves from the `tests` root.
+      target_include_directories(test_submit_gate_wiring PRIVATE tests)
+      # Reaches `shared/...` through render_runner.h, which resolves from `frontends`.
+      target_include_directories(test_submit_gate_wiring PRIVATE frontends)
+      target_link_libraries(test_submit_gate_wiring PRIVATE prosper_core Vulkan::Vulkan)
+      add_test(NAME submit_gate_wiring COMMAND test_submit_gate_wiring)
 
       # GPU-executor core (Stage A): GpuState -> execute_gpustate() -> render -> present round-trip.
       add_executable(test_gpu_execute tests/gpu/execute/test_gpu_execute.cpp)
--- a/prosper/docs/FRONTEND_APP.md
+++ b/prosper/docs/FRONTEND_APP.md
@@ -89,6 +89,19 @@ guest caused a reproducible Windows access violation when static teardown raced 
 The remaining lifecycle work is a guest flip-boundary check followed by a real join and normal
 teardown (#352).
 
+**A second, non-obvious cost of that missing join, and the partial answer to it (#3225).** `_Exit`
+becomes `exit_group()`, and a thread that is inside a syscall at that moment cannot be torn down
+until it returns from the kernel. When the syscall is an amdgpu command submission the task parks in
+`__drm_exec_lock_obj` on a GEM reservation nobody will now release: the process becomes an unreapable
+zombie (`SIGKILL` is a no-op) and the compositor's own DRM work queues behind it — a frozen desktop
+that only a root-forced GPU reset clears. Because there is no ordering on this path at all, that
+in-flight submission is abandoned on *every* close; that it usually costs nothing is luck about where
+the guest thread happened to be. `src/host/platform/gpu_submit_gate.hpp` bounds it: the frontend
+closes a gate that every guest-thread submission enters, so new submissions are refused
+(`VK_ERROR_DEVICE_LOST`) and a bounded drain can wait for the ones already inside. It is a
+mitigation, not the fix — a thread that never returns still expires the drain — and the fix remains
+the join above.
+
 ## Where the dump path comes from (#1469)
 
 `boot_program()` needs an app0 directory. Argv supplies it for every agent route, snapshot run, and CI
@@ -272,7 +285,12 @@ Handle swapchain resize (`present_width/height` change or window resize → recr
 
 Target shutdown ordering: `request_stop` → guest thread observes the flag at its loop boundary and
 returns → join → tear down GPU/window. Until that check exists, direct process exit deliberately
-skips frontend/HLE static teardown so it cannot race the detached guest.
+skips frontend/HLE static teardown so it cannot race the detached guest — preceded, since #3225, by
+`gpu_submit_gate_begin_shutdown()` → `gpu_submit_gate_drain()` → `vkDeviceWaitIdle`, which bounds the
+window in which that exit can kill a thread inside a GPU submission. The drain is also what makes the
+`vkDeviceWaitIdle` legal: it is the external synchronization of every `VkQueue` that call requires, so
+a drain that TIMES OUT deliberately skips the wait rather than racing a live submit. `tools/screenshot`
+runs the same two steps before its own `_exit` (`screenshot.cpp`), for the same reason.
 
 ## Target / build layout
 
@@ -739,6 +757,18 @@ duplicate. Until then the app is fully functional via `--test-pattern` (and any 
   Native Windows build/run packaging is done via `scripts/run-windows.ps1`. Cooperative guest-stop
   at a flip boundary is a follow-up (today the
   guest thread is detached at window-close and reclaimed by process exit).
+
+## Ruled out
+
+- **"The headless path is safer by design."** The issue observed that `tools/screenshot` had never
+  taken the desktop down and read that as a property of the frontend. It is not: `screenshot.cpp`
+  detaches its guest thread and calls `_exit` in exactly the same shape `prosper-app` does, so it can
+  abandon an in-flight submission identically. What differs is the consequence — no compositor is
+  sharing the device with a headless run, so a stuck reservation has nothing to block. Both frontends
+  now take the gate (#3225).
+- **"A GPU hang / driver fault is involved."** No fault is logged at all: nothing in `journalctl -k`
+  for 30 minutes around the freeze, and the amdgpu hang detector never fires — there is no stuck job
+  to detect, only a held lock, which is why there is no self-recovery and no timeout (#3225).
 
 ## Risks & open questions
 
--- a/prosper/frontends/prosper-app/main.cpp
+++ b/prosper/frontends/prosper-app/main.cpp
@@ -23,6 +23,7 @@
 #include "shared/present/present_blit_policy.hpp"    // reject stale CPU/GPU representations of guest flips
 #include "hle/sync/sync_futex.hpp"         // dump_guest_sync_trace (PROSPER_SYNC_RING deadlock history)
 #include "host/platform/lifecycle.hpp"          // frontend-owned stop/pause gates
+#include "host/platform/gpu_submit_gate.hpp"     // #3225: drain guest GPU submits before _Exit
 #include "host/image/boot_program.hpp"       // boot_program (shared guest-boot path, also used by boot_trace)
 #include "host/image/exec_image.hpp"         // run_entry
 #include "host/memory/guest_write_watch.hpp"  // flush dmem writer diagnostic before deliberate _Exit
@@ -2957,6 +2958,38 @@ int main(int argc, char** argv) {
     // let the OS reclaim process state without running destructors under the live guest.
     if (g_guest_thread.joinable()) {
         g_guest_thread.detach();
+
+        // #3225: bound the window in which the detached guest thread can be inside a GPU submission
+        // when std::_Exit fires. _Exit becomes exit_group(), and a thread that is inside an amdgpu
+        // command submission at that moment cannot be torn down until it returns from the kernel —
+        // it parks in __drm_exec_lock_obj on a GEM reservation nobody will now release, the process
+        // becomes an unreapable zombie, and the compositor's own DRM work blocks behind it. That
+        // froze the developer's desktop twice in one day, recoverable only by a root-forced GPU
+        // reset.
+        //
+        // Closing the gate stops NEW guest submissions (they return VK_ERROR_DEVICE_LOST, which
+        // every submission site already handles as "not submitted"), which is what lets the drain
+        // reach zero instead of chasing a moving count. This BOUNDS the window; it does not close
+        // it. A thread already inside vkQueueSubmit is waited for, and if it never comes out the
+        // drain times out and we exit exactly as before. The real fix is the cooperative stop the
+        // comment above names as unimplemented — observe the flag, stop at a submission boundary,
+        // and JOIN.
+        prosper::gpu_submit_gate_begin_shutdown();
+        constexpr int kGuestSubmitDrainMs = 2000;
+        if (prosper::gpu_submit_gate_drain(kGuestSubmitDrainMs)) {
+            // Only after a SUCCESSFUL drain. vkDeviceWaitIdle requires host access to every VkQueue
+            // of the device to be externally synchronized, and the drain is what provides that here
+            // (the main loop uses the shared-present submit mutex for the same reason at the
+            // swapchain-resize wait above). A timed-out drain means a guest submit may still be
+            // live, so waiting would both violate that and be the very block we are avoiding.
+            if (vk.device) vkDeviceWaitIdle(vk.device);
+        } else {
+            fprintf(stderr,
+                    "[app] guest GPU submissions did not drain within %d ms (%d still in flight); "
+                    "exiting anyway\n",
+                    kGuestSubmitDrainMs, prosper::gpu_submit_gate_in_flight());
+        }
+
 #ifdef PROSPER_HAVE_LIVE_RENDERER
         // _Exit skips RuntimeComputeTimingSelector's destructor. Publish its validity verdict here;
         // the report is idempotent so a future cooperative teardown cannot duplicate it.
--- a/prosper/frontends/prosper-app/main.cpp
+++ b/prosper/frontends/prosper-app/main.cpp
@@ -2982,6 +2982,14 @@ int main(int argc, char** argv) {
             // (the main loop uses the shared-present submit mutex for the same reason at the
             // swapchain-resize wait above). A timed-out drain means a guest submit may still be
             // live, so waiting would both violate that and be the very block we are avoiding.
+            //
+            // Note what this wait does and does not cost. It is UNBOUNDED — on a title that has
+            // hung a GPU job it returns only when the amdgpu watchdog resets the device, adding
+            // seconds to a close — and it is not protecting a teardown, because the _Exit below
+            // destroys no Vulkan object. What it buys is that already-submitted guest work is off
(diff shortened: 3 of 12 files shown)


mattias800:
## Author verification at exact head `fa09a133`

Container `ps5ys`, Linux, AMD Radeon 8060S (RADV STRIX_HALO).

```
cmake --build prosper/build-linux -j6      rc=0
ctest --no-tests=error                     rc=0, 336/336 passed
```

Both new cases confirmed by name, not by the total:

```
ctest -N | grep -E 'gpu_submit_gate|submit_gate_wiring'
  Test  #47: gpu_submit_gate
  Test #326: submit_gate_wiring
```

**Runs at this head** — all clean exits, `pgrep -x` empty afterwards, no `Z`-state process, and **no
drain-timeout line in any of them**:

| route | runs | result |
| --- | --- | --- |
| `tools/screenshot PPSA02664 --seconds 3 --count 2` | 2 | `rc=0`, 2/2 shots |
| `prosper-app --frames 200` (offscreen, `PPSA02664`) | 1 | `rc=0`, `[app] shutting down after 200 presented frame(s)` |

The larger set in the PR body (5 × screenshot, 3 × offscreen app, 1 × interactive windowed app with a
post-run `vulkaninfo --summary` and a clean `journalctl -k`) was run at `685563b7`; the second commit
touches only the compute fence-timeout recovery path, which none of those runs reaches.

**Mutation arms** (from `685563b7`, unaffected by the second commit — each built to `rc=0` first, so
a silently non-applied edit could not masquerade as a non-load-bearing test):

| arm | build | `ctest` |
| --- | --- | --- |
| `gpu_submit_gate_begin_shutdown()` made a no-op | rc=0 | rc=8 — 9 assertions red, including "drain reaches zero against live submitting threads"; `submit_gate_wiring` aborts, having reached `vkQueueSubmit` with a null queue |
| `gpu_submit_gate_drain()` returns `true` unconditionally | rc=0 | rc=8 — 5 assertions red, including the timeout bound and "nothing is in flight after that drain" |

**Deliberately not run:** any attempt to reproduce the freeze. The mechanism is established from
source and a reproduction costs the machine owner their desktop. Snapshots were also not run — the
gate is open for the entire life of a running title, so no rendering path changes.

mattias800:
## Re-verification at head `5201188d` (after the review findings)

```
cmake --build prosper/build-linux -j6      rc=0
ctest --no-tests=error                     rc=0, 336/336 passed
```

Runs at this head — all clean exits, `pgrep -x` empty, no `Z`-state process, and now also **zero
`fatal Vulkan device loss` lines**, which is the point of the first finding's fix:

| route | runs | result |
| --- | --- | --- |
| `tools/screenshot PPSA02664 --seconds 3 --count 2` | 2 | `rc=0`, 2/2 shots |
| `prosper-app --frames 200` (offscreen, `PPSA02664`) | 1 | `rc=0`, clean `[app] shutting down` |

**Mutation arms re-run at this head**, each built to `rc=0` first:

| arm | build | `ctest` | effect |
| --- | --- | --- | --- |
| `gpu_submit_gate_begin_shutdown()` made a no-op | rc=0 | rc=8, 2/2 red | **10** assertions red (was 9 — the new open-gate submit arm is one more that depends on the gate) |
| `render_locked_queue_submit` hard-coded to `VK_ERROR_DEVICE_LOST` | rc=0 | rc=8 | exactly one assertion red: "with the gate OPEN render_locked_queue_submit reaches the driver and succeeds" — the arm the review said was missing |

CI was fully green on `fa09a133` (9 checks pass, `Progress tracker` and `Publish Release` skipped)
before these three review commits; it is re-running on the new head.

mattias800 (commented):
**APPROVED.**

This fixes a bug that cost the project owner a working desktop twice in one day, and it does it
without overselling what it fixes.

## The claim my merge decision rested on, checked first

The risk I was most worried about is that `vkDeviceWaitIdle` on the shutdown path could reintroduce
an unbounded block on **exactly the failure being fixed** — a stuck DRM lock has no watchdog, so a
wait behind it never returns. That hazard is closed by construction: the wait is strictly inside
`if (gpu_submit_gate_drain(2000))`, so the wedge case (drain times out) skips it, prints the
in-flight count and exits as before. The code states the same reasoning, and gives the right
secondary justification — the drain is what provides the external synchronization of every
`VkQueue` that `vkDeviceWaitIdle` requires, so a timed-out drain must not wait.

**I am keeping it**, and recording why so it is a decision rather than an omission: after a
successful drain no guest thread is inside a GPU call, so the only remaining cost is a title with a
*submitted* hung job, where the amdgpu watchdog bounds the wait at seconds. That is bounded,
visible, and documented in-code as safe to delete. It is the first thing to remove if anyone reports
a slow close.

## The review's best catch would have been a genuine own-goal

The compute refusal returning `VK_ERROR_DEVICE_LOST` into `vk_ok` would have latched
`ctx.device_lost`, printed `fatal Vulkan device loss … disabling live compute for this process`, and
added a `reason=queue-submit` row to the decline census **on every shutdown**. Those counts are
quoted as evidence elsewhere in this project, so it would have manufactured a phantom defect into
the exact numbers people reason from — the instrument-trap family, produced by the fix for a
different bug. The corrected path leaves the loop without routing through `vk_ok` at all, with a
comment saying why.

## The primitive is well built

Packing the shutdown flag and region count into **one atomic word** is the right call and the header
explains it correctly: it makes "admit only while open" a single CAS, so admission cannot interleave
with `begin_shutdown`, and a *refused* enter never transiently bumps a count the drain would read.
Two separate atomics would have needed a Dekker-style argument for the same guarantee.

`notify_if_idle`'s race analysis is textbook-correct — seq_cst on both sides of the leaver/waiter
pair so neither can miss the other, plus taking the (empty) mutex before notifying to close the
window where a waiter has evaluated its predicate but not yet slept. That is the bug most condvar
code of this shape actually has.

## Scope is honest, which is what makes it trustworthy

Five sites wired rather than the three I named, and the two extras are the interesting half.
The post-fence-timeout `vkQueueWaitIdle` is a *wait*, not a submit — so not the wedge hazard itself —
but ungated it was invisible to the drain, which would have let the app's `vkDeviceWaitIdle` run
concurrently with it and quietly broken the guarantee everything else rests on. Finding that is the
difference between wiring a brief and understanding it.

And the header now records at `CONFIDENCE: MED` that `__drm_exec_lock_obj` is **also** reachable via
the GEM-VA ioctls behind `vkAllocateMemory`/`vkBind*Memory`, which the guest thread calls ungated.
So the gate bounds the window rather than closing it, and the document says so instead of implying a
fix. Combined with "the real fix is the cooperative stop, which is still unimplemented", a future
reader gets an accurate picture of what is and is not solved.

## Verification

336/336 at `rc=0`, both new cases confirmed by name. Three mutation arms, each reporting a clean
build before its result — `begin_shutdown`→no-op reddens 10 assertions, `drain`→always-true reddens
5, and a forced submit failure reddens *exactly* the new open-gate arm, which is the one that shows
the wiring rather than the primitive.

15 clean runs across two headless titles, offscreen `--frames`, and one interactive windowed
launch-and-close on the live session, with the right post-conditions checked: no leftover process,
no zombie, zero drain timeouts, GPU still enumerating, kernel log clean. Not reproducing the freeze
was correct — the cause was established from source and a reproduction costs a desktop.

## On the attribution finding (N6)

Dismissing it was right. The harness instruction for this session explicitly supersedes
`CLAUDE.md`'s ban on `Claude-Session:` trailers, and the brief said not to run the zero-gate. The
charter and the harness genuinely disagree right now; that is for the project owner to reconcile,
not something to fix inside this PR.

PR #212, opened 2026-07-09
fix(hle): AGCSHADER dump uses env path + s_endpgm-derived size (#160)

mattias800:
## Problem (#160)

The `PROSPER_AGCSHADER` diagnostic (`hle_graphics.cpp`) dumped the captured shader blob to a **hardcoded absolute developer path** (`/mnt/c/Users/matti/repos/ps5ys/prosper/build-linux/shader0.bin`) — on any other machine `fopen` fails and the `[wrote shader0.bin]` log lies — and **guessed the size** as `a4<=0x10000 ? a4 : 0x200`, truncating the dump to a 0x200 fragment whenever `a4` wasn't the byte count.

## Fix

- **Path** from `PROSPER_AGCSHADER_OUT` (default relative `shader0.bin`, so it works anywhere), and **log the `fopen` failure** (`strerror`) instead of silently writing nothing.
- **Derive the size** from the shader's own `s_endpgm` terminator (RDNA2 SOPP `0xBF810000`), bounded to a 16 KiB scan, instead of the fixed 0x200 fragment. An explicit, plausible `a4` is still preferred when present.

## Scope / risk

Diagnostic-only, `PROSPER_AGCSHADER`-gated — no change to the default path. Full suite **65/65 green**.

Fixes #160

PR #3230, opened 2026-09-02
tests: pin the CFG-dispatcher re-arm in the entry-M0 token containment (#3203)

### LINKED ISSUE #3203: The CFG-dispatcher re-arm in the entry-M0 token containment has no regression coverage
## The CFG-dispatcher re-arm site in the entry-M0 token containment has no regression coverage

#3136 contains the entry-M0 token at every point the recompiler writes `rs.sreg` from merge
machinery. One of those sites is the **CFG dispatcher**, which stores an absent scalar into its
Function variable *as zero* and reloads it as tracked data at every block entry — so without a
re-arm, a token would silently become tracked data at a block boundary.

The site is covered by reasoning and by symmetry with the tested MAY-set pattern, but **not by a
test**: an independent reviewer disabled it in isolation and the full 324-test suite stayed
**green**.

That makes it the one part of the containment whose breakage nothing would catch. The other
mechanisms all have arms that redden when disabled:

| mechanism | disabled in isolation |
| --- | --- |
| destination gate (`dst <= 105`) | `[FAIL] s_mov_b32 vcc_lo, m0 was accepted as an entry-M0 save` |
| union join carrying the token | if-only **and** if/else arms flip `REJECT → ok` |
| loop-header reject | loop arm flips `REJECT → ok` |
| **CFG-dispatcher re-arm** | **nothing fails** |

## Why it matters

The whole safety argument for the token is that it is *loud*: every consumer that is not the matching
restore rejects. A leak at a block boundary turns the token into ordinary tracked data, and
`operand_bits`' SGPR case ends in `return b.uconst(0)` with `ok` left **true** — so a leaked token
does not reject, it silently reads as zero. That is the failure mode the containment exists to
prevent, and it is the one path where a regression would be invisible.

## What is wanted

A regression arm whose program routes through the dispatcher with a live token at a block entry, and
which reddens when the re-arm is removed. Note the shape of the existing tests: the ctest binary
returns at its first failure, so per-arm discrimination needs a probe harness that runs each program
independently — #3136's author built one for exactly this reason.

Found by independent review of #3136 and raised as non-blocking; the mechanism reads as correct, it
simply has no guard.

Refs #3136, #3133.

mattias800:
Fixes #3203.

Adds the one regression arm the entry-M0 token containment did not have, plus the controls that
make it mean something. **No production code changes** — the diff is one new test, its CMake
registration, one `## Ruled out` row and one `AGENTS.md`.

## What the re-arm does

`#3136` models the driver's entry M0 as an **opaque token**: `s_mov_b32 sDST, m0` marks `sDST` in
`RegState::sreg_entry_m0` and leaves it with **no `sreg` value**, and `operand_bits` rejects any read
of a register that is token-marked and untracked. Only the matching `s_mov_b32 m0, sSRC` consumes it.

The containment has to be re-asserted at every point the recompiler writes `rs.sreg` from merge
machinery, because all of that machinery reads absent scalars through a `sget` that renders them as
`uconst(0)`. Six of those points are structured-emitter joins. The seventh is the **CFG dispatcher**,
which does the same thing one indirection further away — it persists every referenced SGPR in a
Function variable, stores an absent scalar there **as zero**, and reloads it as ordinary tracked data
at each block entry. `load_state` in `rdna2_emit_cfg.cpp` therefore re-establishes the token on the
MAY set (every register any save in the stream can leave holding it) at every block entry:

```cpp
for (const auto& kv : sv) state.sreg[kv.first] = b.load_function(b.t_u32, kv.second);
for (int reg : entry_m0_may_hold) {          // the re-arm
    state.sreg.erase(reg);
    state.sreg_entry_m0.insert(reg);
}
```

## What breaks without it

The token becomes a **tracked zero** at the first block boundary it crosses. That does not reject:
`operand_bits`' SGPR case ends in `return b.uconst(0)` with `ok` left **true**, so a leaked token
reads as zero, silently, and a fabricated word stands in for whatever the shader does with the
driver's M0. Silent-wrong is precisely the failure the token exists to prevent.

`#3203` is that this was the one part of the containment with no guard: a reviewer disabled it in
isolation and the whole suite stayed green, while every other mechanism has an arm that reddens.

## Why the new test would have caught it — and why the ROUTE is the whole design

A naive arm here fails in a way that looks like success, because most dispatcher entries fall back to
the structurizer, whose `join_entry_m0` supplies the same verdict — so the arm can end up testing a
mechanism that is already covered and pass with the dispatcher re-arm deleted.

Every program in the new test is portable compute (`native_subgroup_size == 0`) carrying **one**
forward SCC if plus a `v_readlane_b32`. That is `portable_compute_cfg_readlane` in `emit_body`, and
it is the one dispatcher entry with **no second attempt**:

```cpp
if (allow_cfg_dispatcher && portable_compute_cfg_readlane) { ...
    const int emitted = try_cfg_dispatcher();
    if (emitted > 0) return true;
    return false;                    // <-- the structurizer never runs
}
```

So for these programs the dispatcher is the **only** emitter: a compile can only be it accepting, a
reject only it rejecting. Both halves are asserted rather than assumed —

* the compiling controls must contain an **`OpSwitch`**, which only the dispatcher emits;
* the rejecting arm's `last_terminal_reject_reason()` must carry the **`cfg-recompile-reject`** tag,
  which only `emit_cfg_state_machine` writes, at **`pc=6`** (the read past the edge, not the save)
  with **`mode=unresolved-operand`** (the operand barrier, not a missing lowering).

If a future change stops routing these programs through the dispatcher, those assertions redden and
say so, instead of the arm quietly starting to test the structured join. The test detects its own
invalidity.

### The arms

Four programs, a **minimal pair set**: in the first three, pc0-pc2 and pc4-pc7 are byte-identical and
only the single dword at pc3 changes, so no verdict can differ for any reason except what that one
instruction does to `s0`.

| program | pc3 | expected |
|---|---|---|
| **the arm** | `s_mov_b32 s0, m0` | **REJECT** — the token cannot cross the block edge |
| control A | `s_mov_b32 s0, 7` | compile, and the module contains an `OpSwitch` |
| control B | `s_mov_b32 s3, m0` | compile — a real save, into a register nothing reads |
| control C | (save + restore inside one dispatch block, later) | compile |

Control B is the load-bearing one. It keeps `entry_m0_may_hold` non-empty so the re-arm loop **runs**,
and still compiles — which is what separates "the re-arm is register-precise" from "the dispatcher
refuses any stream containing a save". Without it the arm could redden nothing on deletion and the
test would be a tautology wearing a reject.

Control A proves the region shape, the readlane route and the cross-block scalar read are all
representable; without it the arm's reject could be any of those three failing instead of the token.

### Its own binary, on purpose

The other `#3133` arms live in `test_rdna2_spirv_struct.cpp`, which returns at its first `[FAIL]` — so
a mutation run there cannot tell which arms moved, and `#3136`'s author had to build a throwaway probe
harness for exactly that. This test uses a `CHECK` macro that counts failures and keeps going, so the
mutation signature below is readable directly from ctest output. That is the harness, kept in-tree.

## Deliberately unaffected

No production code changes at all, so no shader that compiles today can change. The test adds no
`PROSPER_*` switch, creates no Vulkan device, and runs on every platform. It does not touch the six
structured-join arms, the loop-header rejects, or the destination gate — those already have arms that
redden.

## Verification (exact head)

Configured with `PROSPER_GAME_ROOT` set, in the project's `ps5ys` container.

```
cmake -S prosper -B prosper/build-linux                  # rc=0
cmake --build prosper/build-linux -j6                    # rc=0
ctest --test-dir prosper/build-linux --no-tests=error    # 100% tests passed out of 336, rc=0
ctest --test-dir prosper/build-linux -N                  # Test #212: entry_m0_dispatcher
```

**Registered count: 335 on `origin/main` -> 336 with this PR**, measured rather than assumed —
`origin/main`'s `CMakeLists.txt` configured into a throwaway build dir reports `Total Tests: 335`,
this branch's reports `336`, and the added case is present **by name** as `Test #212:
entry_m0_dispatcher`.

Registration verified **by name**, not by count: `prosper/CMakeLists.txt` has separate `if(WIN32)`
and `if(UNIX)` branches and a case registered in the wrong one silently never runs while ctest stays
green. This one is inside `if(TARGET prosper_core)`, which is platform-independent.

### Mutation matrix — the arms are load-bearing

Each mutation applied by an assertion-checked patch that **fails loudly if the text does not match**
(a regex that silently does not apply reports zero failing arms, which reads identically to "the test
is not load-bearing"), with `git diff` shown before building and the **build exit code quoted first**
so it is clear the mutation compiled and ran.

| mutation | build | `test_entry_m0_dispatcher` |
|---|---|---|
| baseline | rc=0 | 9/9 `[ok]`, rc=0 |
| **delete the re-arm entirely** (the exact mutation #3203 names) | rc=0 | **the 4 arm lines redden, all 5 controls stay green**, rc=1; reject reason is **empty** — nothing refused at all |
| **`state.sreg.erase(reg)` kept, `sreg_entry_m0.insert(reg)` dropped** (the "drop to plain untracked" middle ground #3136 warns is silently unsafe) | rc=0 | identical signature: same 4 lines redden, same 5 controls green, rc=1 |
| **re-arm EVERY scalar instead of the MAY set** (an over-broad re-arm) | rc=0 | the **opposite** signature: controls B and C redden, the arm still rejects — but at **`pc=1`**, the `s_cmp` reading s4, not the read the contract is about |
| restored | rc=0 | 9/9 `[ok]`, rc=0, `git diff` on `src/` empty |

The first mutation was also run against the **whole suite**, not just the new binary:
`99% tests passed, 1 tests failed out of 336`, and the single failure is `212 -
entry_m0_dispatcher`. That reproduces #3203's finding exactly — with the re-arm deleted, nothing
else in the tree notices — and it is the direct evidence that this test is the only guard on that
site.

The third mutation is the one worth reading twice. Asserting only that the arm *rejects* would have
let it pass: an over-broad re-arm still refuses the program, just at `pc=1` for a completely
unrelated register. That is the project's recorded "a passing assertion proves nothing about the
mechanism" failure in miniature, and it is why the recorded reject reason's `pc=` and tag are
asserted rather than the emptiness alone. Those three signatures are written into the test file so
the next reader can tell a real regression from a fixture that drifted.

## Falsification recorded

`docs/RECOMPILER_REMAINING.md` § Ruled out, beside the `#3133`/`#3136` row: the hypothesis that the
dispatcher half cannot be pinned by a test, and the no-fallback route that falsifies it.


mattias800 (commented):
**APPROVED.**

A coverage-only PR is the easiest kind to do badly — a passing test is trivial, and a test that would
actually have *caught* the defect is the whole deliverable. This is the latter, and it proves it
rather than claiming it.

## The design decision that makes it work

The trap here is real and the PR names it: most dispatcher entries **fall back to the structurizer**
on a clean reject, so a naive arm ends up exercising `join_entry_m0` — already covered — and passes
with the dispatcher re-arm deleted. Choosing `portable_compute_cfg_readlane`, the one entry that
`return false`s with **no second attempt**, is what makes the dispatcher the only possible emitter
for these programs. A compile can only be it accepting; a reject can only be it rejecting.

And that route is **asserted, not assumed**: the compiling controls must contain an `OpSwitch`
(opcode 251), which only `emit_cfg_state_machine` produces. If a future change reroutes these
programs, those assertions redden instead of the arm quietly changing subject — which is exactly how
a guard rots into a tautology.

## Asserting the reason, not just the rejection, is what earns this

`arm.empty()` alone would have been the weak version. Requiring
`last_terminal_reject_reason` to carry `cfg-recompile-reject` **at `pc=6`** pins *which* register the
re-arm protected and *which* emitter refused it.

Mutation 3 is the vindication: re-arming every scalar rather than the MAY set still rejects the
program — so an emptiness-only assertion would have passed it — but rejects at **`pc=1`** for an
unrelated register, and reddens the controls. An over-broad re-arm and a correct one produce opposite
signatures, and the test tells them apart. That is this project's "true assertion, wrong mechanism"
trap caught in advance rather than in a post-mortem, and recording all three signatures in the test
file is the right place for them.

Noticing that `record_terminal_reject_reason` early-returns on a zero program address, and giving
each arm a distinct non-zero one, is the detail that would otherwise have made every reason assertion
silently vacuous.

## The `## Ruled out` row is the most valuable artifact here

It **falsifies a prior conclusion** — that the dispatcher half could not be pinned at all, because
anything reaching it also reaches the structured emitter. That belief was reasonable and wrong, and
without this row the next person re-derives it and gives up again. It also records the measurement
that settles it: with the two re-arm lines deleted the program *compiles*, no reason is recorded, and
every other registered test stays green.

## Evidence

Four mutation arms, each reporting **build rc=0 before its result**, so none is a baseline re-run in
disguise. The decisive number is mutation 1 against the whole suite: **335 of 336 pass, the single
failure being `entry_m0_dispatcher`** — direct evidence that this arm is the only thing in the tree
guarding that site, which is precisely the claim a coverage PR has to support.

Baseline measured rather than assumed (`origin/main` 335, branch 336) and the case confirmed **by
name** as `Test #212`, not inferred from the total — the check that caught a silently unregistered
test here two days ago. Registered in the platform-independent `if(TARGET prosper_core)` block.

+260/−0 with **no production code touched**, so the blast radius is a test binary. `diff --check`
clean, zero deletions repo-wide, no host paths. CI 9 pass / 2 skipped / 0 failures on `main`.

Also adding an `AGENTS.md` to a folder that had none, explaining why a focused reject test gets its
own executable rather than joining a large one, is the kind of navigation note that stops the next
person appending to the wrong binary.

PR #3232, opened 2026-09-02
shader_conformance: replace scan.py's ported MIMG length rules with the real decoder

### LINKED ISSUE #3184: shader_conformance: replace scan.py's ported MIMG length rules with the real decoder (shader_histo)

### LINKED ISSUE #3229: shader_histo: Rdna2Format name table is one entry short — VOP3P mislabelled UNK, and Unknown reads out of bounds (4 of 55 dumps segfault)
`tools/shader_histo/shader_histo.cpp:87` carries its own copy of the `Rdna2Format` name table:

```cpp
const char* fn[] = {"SOP2","SOP1","SOPK","SOPC","SOPP","SMEM","VOP2","VOP1","VOPC","VOP3",
                    "VINTRP","DS","MUBUF","MTBUF","MIMG","FLAT","EXP","UNK"};
```

That is **18 entries**. `Rdna2Format` (`src/gpu/recompiler/rdna2_decode.hpp:19`) has **19 values** —
`VOP3P` was inserted between `EXP` and `Unknown`, and this table was never updated. So:

- **`fn[17]` (VOP3P) prints as `UNK`.** Every packed/mixed-precision VALU instruction in the
  histogram is labelled as an unclassified encoding. The `unknown=` total is still correct (it keys
  on `fmt == Rdna2Format::Unknown`, not on the label), so the histogram body and its own summary
  line silently disagree, and the format most likely to be read as "the decoder failed here" is the
  one that decoded fine.
- **`fn[18]` (`Unknown`) is an out-of-bounds read**, at both use sites: the histogram loop
  (`:89`, `fn[k.first]`) and the blockers loop (`:105`, guarded only against a negative index). Any
  eboot whose shaders contain one undecodable encoding — which is the normal case — dereferences
  whatever follows the array.

**Failure scenario, measured 2026-09-02** on this machine over the 55 `PPSA*-app0` dumps in the
local library, `shader_histo <eboot.bin> <biggest.bin> <dumpdir>` for each:

```
CRASH PPSA21564-app0 rc=139
CRASH PPSA24651-app0 rc=139
CRASH PPSA25009-app0 rc=139
CRASH PPSA27616-app0 rc=139
```

Four segfaults out of 55. The remaining 51 do not crash but print the garbage the OOB pointer
happens to land on; under gdb, PPSA24651 survives and prints the tell directly:

```
  EXP    op=0x0    x37
  (null) op=0x0    x1
```

That `(null)` row is the `Unknown` bucket reading past the end of `fn[]`.

**Suggested fix**: add `"VOP3P"` before the terminal entry and rename the terminal one to `"UNK"`
so the table is 19 long, and bound the index at both use sites
(`k.first < 0 || k.first >= (int)std::size(fn) ? "?" : fn[k.first]`). `shader_inspect` already gets
this right — `format_name` (`tools/shader_inspect/shader_inspect.cpp:19`) carries all 19 names and
bounds-checks with `std::size` — so the two tools disagree today about what a VOP3P instruction is
called.

This is the same class of defect as #3184: a hand-maintained copy of something the decoder already
knows, which drifted when the decoder gained a case.

mattias800:
Fixes #3184
Fixes #3229
Refs #3040

## What was duplicated

`tools/shader_conformance/scan.py` needs to know where the MIMG instructions are in a raw RDNA2
dump, and to do that it needs real per-instruction lengths — otherwise an operand or a trailing
literal dword whose top six bits alias the MIMG encoding (`0b111100`) reads as an instruction start.
#3040 gave it that by **porting** `rdna2_decode_one`'s format and length dispatch into Python:
`rdna2_instr_len` + `_sop_has_literal` + `class Step`, ~90 lines covering the VOP src0 modifier and
literal rules, the six K-carrying VOP2 mul-adds, the SOP literal fields, `S_SETREG_IMM32_B32`, the
VOP3/VOP3P 9-bit literal markers, the two-dword bucket, and the MIMG NSA extra-dword count.

Every one of those is a rule that also lives in `src/gpu/recompiler/rdna2_decode.cpp`. #3040's own
body flagged the port as the weaker fix and named this one.

## Did the two implementations agree?

**Yes, everywhere, on the whole local corpus — and that is why the copy had to go rather than why it
could stay.**

Differential over every `.shader_text` section embedded in all 55 `PPSA*-app0` eboots in the local
dump library, extracted with `shader_histo`:

| | |
| --- | --- |
| eboots scanned | **55** (30 carry embedded shader ELFs) |
| shaders compared | **11,266** |
| MIMG sites decoded | **199,521** |
| `(pc, opcode, dim)` disagreements | **0** |
| decoder failures | **0** |

So no historical `scan.py` finding is invalidated by this change, and no recall is lost. The
argument for #3184 is the drift a copy *invites*: it is the thing nobody updates when the real
decoder changes, and it fails by silently returning a plausible instruction stream — which is the
exact failure mode this tool exists to make impossible.

**That is not a hypothetical.** While extracting the corpus, the same folder turned out to have
already drifted in the other direction: `shader_histo`'s hand-maintained `Rdna2Format` name table
went one entry short when `VOP3P` was inserted before `Unknown`, so VOP3P printed as `UNK` (29,776
instructions on PPSA21564 alone) and `Unknown` read one past the end of the array — **4 of the 55
local dumps segfault**. Filed as #3229 and fixed here in its own commit. Re-extracting the whole
corpus with the fixed binary afterwards: **55 of 55 eboots, 0 crashes**, and PPSA21564's 29,776
VOP3P instructions now carry their real name instead of `UNK`.

## Approach

`shader_inspect` gains `--mimg-sites`:

```text
$ shader_inspect exec_ps_7f123400.bin --mimg-sites
mimg-site pc=42 op=0x20 dim=5
mimg-site pc=61 op=0x0 dim=1
mimg-sites-end dwords=903 consumed=903 instructions=781 sites=2 endpgm=1
```

One line per image instruction, straight off `rdna2_walk`, and nothing else — no coverage, no stage
recompile, no resource-table reasoning. `scan.py` parses those lines; `rdna2_instr_len`, `Step`,
`_sop_has_literal`, `MIMG_ENCODING` and `S_ENDPGM` are deleted. `DIM_NAMES` / `DIM_ARRAYED` /
`DIM_3D` stay: they are the classifier's naming and grouping of a value the decoder hands over, and
they read no instruction word.

**Why `shader_inspect` and not `shader_histo`, which #3184 names.** `shader_histo`'s input is an
*eboot* — it scans a module for embedded shader ELFs. `scan.py` holds raw dumps written by
`gpu_replay`, which is precisely `shader_inspect`'s input, and `shader_inspect` already walks them
with `rdna2_walk` and already prints per-instruction `pc`/`fmt`/`opcode` rows. Adding a raw-stream
mode to `shader_histo` would have been a second input contract for the same job. The issue offers
"`shader_histo` (or `gpu_replay`)"; this is the same fix in the tool the folder layout points at.

**`mimg-sites-end` is a completion sentinel, not decoration.** "This shader has no image
instruction" is the answer that certifies a clean conformance scan, and it must never be
indistinguishable from "the decoder died before printing anything". The scanner refuses any census
that arrives without the sentinel, whatever the exit code said — and cross-checks the sentinel's own
`sites=` count against the number of `mimg-site` lines it parsed, which closes the same hole on the
*reading* side: a line the regex stops matching (a widened field, a renamed key) would otherwise be
dropped silently, and an under-count reads as a cleaner shader.

**There is deliberately no in-process fallback.** A fallback would be this same duplicate
implementation, now also untested; a missing `shader_inspect` stops the scan (exit 2) instead.

## Deliberately unaffected

* `shader_inspect`'s existing output, exit codes and `--stage` behaviour. `--mimg-sites` replaces the
  listing and refuses to combine with `--stage`, so no existing invocation changes shape.
  `shader_inspect_resource_table_honesty` passes untouched.
* `classify()`, `parse_spirv_images()`, the 0/1/2 exit contract, the per-capture temp-directory and
  return-code rules from #3039, and the barren-capture guard. Verified by output diff, below.
* `rdna2_decode.cpp` itself — not one line. This PR moves a consumer onto it.

## Behaviour change (one, intended)

A raw ISA dump that is empty or **not a whole number of dwords** is now counted as *not readable*,
with a stated reason, instead of being silently truncated to whole dwords and scanned. A partial
dword is a truncated dump, and a scanner that reports on one is reporting on bytes it does not have.
Zero-byte dumps were already skipped by an earlier guard, so in practice this only reaches a dump
whose size is not a multiple of 4.

## Risks

* **The self-test is no longer hermetic.** It needs the built `shader_inspect`, so ctest passes
  `--mimg-decoder $<TARGET_FILE:shader_inspect>` and the registration moved under
  `if(TARGET prosper_core)`. This is the point rather than a cost — the decode arms now exercise
  `rdna2_decode.cpp` — but a bare-checkout `python3 scan.py --self-test` now exits **2** (could not
  test) where it used to run. It never reports a pass it did not earn.
* **One subprocess per shader** is added to a scan that already spawns three.
* **The corpus contains no NSA MIMG.** Mutating the real decoder to ignore the NSA extra-dword count
  produced *zero* differential disagreements, which means that specific rule is not exercised by any
  of the 11,266 shaders. The `self_test` NSA arm is hand-built precisely because the corpus cannot
  supply one; the equivalence claim above is therefore about the rules the corpus does exercise, and
  the instrument control below establishes that it exercises them.

## Build / test / evidence

Built and tested in the `ps5ys` container, worktree-local build dir, at the exact pushed head.

```bash
PROSPER_GAME_ROOT=<DUMP_ROOT> cmake -S prosper -B prosper/build-linux   # rc 0
cmake --build prosper/build-linux -j6                                                # rc 0
cd prosper/build-linux && ctest --no-tests=error                                     # rc 0
```

`ctest`: **335 tests, 100% passed, exit 0** — identical count to the pre-change baseline on
`origin/main` (335, 100%, exit 0) measured in the same worktree.

Both affected tests confirmed **by name**, not by total:

```bash
ctest -N | grep -iE 'shader_conformance|shader_inspect'
#   Test  #36: shader_conformance_scan_logic
#   Test #231: shader_inspect_resource_table_honesty
ctest --no-tests=error -R 'shader_conformance_scan_logic|shader_inspect_resource_table_honesty'
# 100% tests passed out of 2, exit 0
```

`scan.py --self-test` reports **37 arms, all ok**, including five new ones: a zero-byte dump and a
half-dword dump are refused rather than decoded as "no MIMG", and a stub decoder that exits 0 having
printed nothing, prints sites without the sentinel, or prints a sentinel whose `sites=` count
disagrees with the lines it printed, is refused rather than read as a census.

### Output equivalence, before and after

**1. ISA walk, real input.** Pre-change `scan.py:decode_mimg_sites` vs `--mimg-sites`, `(pc, opcode,
dim)` compared elementwise over 11,266 shaders / 199,521 sites from all 55 local dumps: **0
disagreements, 0 decoder failures**. The harness imports the *pre-change file* rather than a retyped
copy of it.

**2. Whole scanner, end to end.** Both scanners driven over 14 combinations of the self-test stub's
capture shapes (`good`, `bad`, `empty`, `broken`, `nodump`, `emptyisa`, `partial`, `silentzero`, and
mixtures), comparing the complete `--json` report and the process exit code, with only per-run temp
paths normalised: **14 identical, 0 differing.**

### Mutation arms (with build exit codes)

| # | Mutation | Expected | Observed |
| --- | --- | --- | --- |
| A1 | `rdna2_decode.cpp`: MIMG length `2 + extra` → `2` (build rc **0**) | `shader_conformance_scan_logic` red | **ctest exit 8**, failing arm named: *"NSA's extra dwords … are skipped …"* |
| A2 | same mutated decoder, **pre-change** `scan.py --self-test` | shows the old suite was blind | **exit 0, `== self-test PASSED ==`** |
| B | `rdna2_decode.cpp`: MIMG length `2 + extra` → `3 + extra` (build rc **0**) | differential detects it | **5,680 of 11,266 shaders disagree, across 29 titles** |
| C | `shader_histo.cpp`: name table back to 18 entries | build fails | **`cmake --build` exit 2**, `static_assert` at `shader_histo.cpp:97` |
| D | `scan.py`: disable the sentinel/line-count cross-check | new arm red | **`--self-test` exit 1**, `[FAIL] a census whose sentinel count disagrees with its lines …` |

A1 + A2 together are the load-bearing pair: a change to the *real* decoder now reddens the
conformance suite, and could not before — the arms tested the Python copy, which is exactly how a
copy drifts unnoticed.

B is the domain control for the clean zero in the equivalence table. A same-source positive control
would only have shown the harness runs; this shows the corpus can *express* a length divergence, so
the 0 is a real null rather than a blind instrument. (It also explains why the more realistic drift
mutations — dropping the NSA count, dropping two VOP2 K-literal opcodes — showed 0: those rules
happen not to be exercised by this corpus. Stated in Risks above rather than hidden.)

### Docs gate

```bash
    --baseline <origin/main copy> prosper/docs/GAME_COMPAT_ORCHESTRATION.md   # rc 0
git ls-files '*.md' -z | xargs -0 python3 prosper/tools/docs/check_numbered_table.py  # rc 0
```

`git diff --check` clean.

## Recorded

`tools/shader_conformance/AGENTS.md` gains a **`## Ruled out`** section carrying both falsified
hypotheses with their evidence — "the port has already drifted" and "deleting the port costs
recall" — plus the rewritten walk description and the self-test's new dependency.
`tools/shader_inspect/README.md` documents `--mimg-sites` and why the sentinel exists. `BLOG.md` has
an entry for the `shader_histo` finding.

### DIFF
--- a/BLOG.md
+++ b/BLOG.md
@@ -21,6 +21,22 @@ from the tracker issues, and still gated, because it is a projection of state ra
 
 ## 2026-09-02
 
+### The shader-conformance scanner stopped carrying its own copy of the RDNA2 decoder
+
+No picture in this one — the finding is that `shader_histo` has been reading past the end of an
+array on every eboot with an undecodable instruction, and segfaulting outright on 4 of the 55 local
+dumps. Its hand-written list of RDNA2 format names went one entry short when `VOP3P` was added to
+the enum, so `VOP3P` printed as "UNK" and `Unknown` indexed off the end
+([#3229](https://github.com/mattias800/prosper/issues/3229)).
+
+We found it while deleting a different copy of decoder knowledge: `shader_conformance/scan.py`
+carried a Python port of the recompiler's instruction-length rules so it could tell an image
+instruction from an operand dword that looks like one. It now calls prosper's real decoder instead
+([#3184](https://github.com/mattias800/prosper/issues/3184)). The port turned out to be exactly
+right — 11,266 real shaders out of the dump library, 199,521 image instructions, zero disagreements
+— which is the point: a copy that is correct today is still the one nobody updates tomorrow.
+
+
 ### Astro Bot's world map is lit, and no longer takes the GPU down with it
 
 The world map now renders and keeps animating for the whole run, where before it froze on a white
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -408,14 +408,24 @@ if(Python3_Interpreter_FOUND)
   # The shader-conformance scanner (#3036), registered because its whole value is a CLEAN result
   # and its original failure mode was manufacturing one: a capture that yielded zero examined
   # shaders exited 0, and findings could be attributed to a capture whose dumps had all failed.
-  # The suite is hermetic -- it drives the scanner end to end against a stub gpu_replay and a
-  # `spirv-dis` shim, so it needs neither a GPU nor the Vulkan toolchain, and it pins the exit
-  # contract (0 clean / 1 mismatch / 2 could-not-scan) rather than only the two parsers. Ten
-  # mutations of the logic it covers, including `DIM_ARRAYED` dropping the 2D_ARRAY case that is
-  # the tool's entire reason to exist, each turn it red.
-  add_test(NAME shader_conformance_scan_logic
-           COMMAND "${Python3_EXECUTABLE}"
-                   "${CMAKE_SOURCE_DIR}/tools/shader_conformance/scan.py" "--self-test")
+  # It drives the scanner end to end against a stub gpu_replay and a `spirv-dis` shim, so it needs
+  # neither a GPU nor the Vulkan toolchain, and it pins the exit contract (0 clean / 1 mismatch /
+  # 2 could-not-scan) rather than only the two parsers. Ten mutations of the logic it covers,
+  # including `DIM_ARRAYED` dropping the 2D_ARRAY case that is the tool's entire reason to exist,
+  # each turn it red.
+  #
+  # It needs the BUILT `shader_inspect` since #3184: the scanner's MIMG walk is now prosper's own
+  # RDNA2 decoder rather than a Python port of its length rules, so the decode arms exercise
+  # `rdna2_decode.cpp` itself and a change to its length dispatch reddens this suite. Guarded on
+  # `TARGET prosper_core` because that is what governs the executable; the generator expression
+  # resolves against the target added later in this file. Without the binary the suite exits 2
+  # (could not test) rather than reporting a pass.
+  if(TARGET prosper_core)
+    add_test(NAME shader_conformance_scan_logic
+             COMMAND "${Python3_EXECUTABLE}"
+                     "${CMAKE_SOURCE_DIR}/tools/shader_conformance/scan.py" "--self-test"
+                     "--mimg-decoder" "$<TARGET_FILE:shader_inspect>")
+  endif()
   # stack_profile's frame classifier, against RECORDED gdb output. Deliberately needs no gdb and no
   # ptrace: an end-to-end variant would skip silently wherever ptrace is unavailable, and a silently
   # skipped test is indistinguishable from a passing one. What it pins is the case that already
--- a/prosper/tools/AGENTS.md
+++ b/prosper/tools/AGENTS.md
@@ -405,6 +405,11 @@ the shipped runtime. Build them from `build-linux/` like everything else.
 - **`shader_inspect/`** — decode one raw `PROSPER_SHADER_DUMP` binary offline. It prints bounded
   instruction PCs, operands, raw words, signed branch immediates, and resolved branch targets so a
   failed shader's CFG can be mapped without hand-counting variable-length instructions.
+  **`--mimg-sites` is the machine-readable half (#3184)**: one `mimg-site pc=… op=… dim=…` line per
+  image instruction and nothing else, so a tool that needs to know where the MIMG instructions are
+  never has to carry its own port of the RDNA2 length rules. `shader_conformance/scan.py` did carry
+  one, and now consumes this instead. The trailing `mimg-sites-end` line is a completion sentinel —
+  a real empty census must never look like a process that died before printing.
   **And it answers "which opcode?", never "which value?" (#2132).** A fold that stops early looks
   identical whether it ran out of modelled opcodes or out of *readable data* — the reject line is the
   same — and a disassembler can only see the first. Sonic Racing: CrossWorlds' failing vertex fetch
--- a/prosper/tools/shader_conformance/AGENTS.md
+++ b/prosper/tools/shader_conformance/AGENTS.md
@@ -30,9 +30,14 @@ Both halves are already cheap and headless, which is what makes this scale:
 
 ```bash
 python3 tools/shader_conformance/scan.py <capture.prgbundle|.prgcap> \
-        --gpu-replay ./build-linux/gpu_replay
-python3 tools/shader_conformance/scan.py --self-test     # no capture, no GPU needed
+        --gpu-replay ./build-linux/gpu_replay --mimg-decoder ./build-linux/shader_inspect
+python3 tools/shader_conformance/scan.py --self-test \
+        --mimg-decoder ./build-linux/shader_inspect          # no capture, no GPU needed
 ```
+
+Both binaries are required, and both default to `./build-linux/…`. `shader_inspect` is not an
+optional accelerator: since #3184 it **is** the ISA walk (see below), so without it the scanner
+cannot find a single image instruction — and it exits 2 rather than reporting a clean run.
 
 Exit codes are the interface: **0** scanned and clean, **1** mismatches found, **2** could not scan.
 `--json` for machine consumption.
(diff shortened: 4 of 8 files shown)


mattias800:
## Author verification — exact head `98c1610`

Linux, `ps5ys` toolchain container, worktree-local build directory.

| step | command | result |
| --- | --- | --- |
| configure | `PROSPER_GAME_ROOT=<DUMP_ROOT> cmake -S prosper -B prosper/build-linux` | exit **0** |
| build | `cmake --build prosper/build-linux -j6` | exit **0** |
| test | `ctest --no-tests=error` | **335 discovered, 335 passed, exit 0** |

Baseline on `origin/main` in the same worktree before any edit: **335 discovered, 335 passed, exit
0** — the count is unchanged, so nothing was added or dropped from the suite.

Both affected cases confirmed **by name**, not by total:

```
ctest -N | grep -iE 'shader_conformance|shader_inspect'
  Test  #36: shader_conformance_scan_logic
  Test #231: shader_inspect_resource_table_honesty
ctest --no-tests=error -R 'shader_conformance_scan_logic|shader_inspect_resource_table_honesty'
  100% tests passed out of 2, exit 0
```

`scan.py --self-test`: **37 arms, all ok**.

### Equivalence, re-run against this exact head

The sentinel cross-check landed after the first equivalence pass, so both were re-run against the
shipped code — the second one *through* `scan.py`'s own `decode_mimg_sites` rather than a
re-implementation of it, so what is compared is the function that actually runs.

* **ISA walk, real input** — pre-change in-process walk vs the final `decode_mimg_sites`, over every
  `.shader_text` in all 55 local `PPSA*-app0` eboots (30 carry any):
  `shaders=11266 agree=11266 disagree=0 refused=0 mimg_sites=199521`.
* **Whole scanner, end to end** — both scanners over 14 stub-capture combinations, complete `--json`
  report and exit code compared verbatim (only per-run temp paths normalised):
  **14 identical, 0 differing.**

### Mutation arms

| # | mutation | build | observed |
| --- | --- | --- | --- |
| A1 | `rdna2_decode.cpp` MIMG length `2 + extra` → `2` | rc 0 | `shader_conformance_scan_logic` **red, ctest exit 8**; failing arm named |
| A2 | same mutated decoder, **pre-change** `scan.py --self-test` | — | **exit 0, `== self-test PASSED ==`** |
| B | `rdna2_decode.cpp` MIMG length `2 + extra` → `3 + extra` | rc 0 | differential: **5,680 of 11,266 shaders disagree, 29 titles** |
| C | `shader_histo.cpp` name table back to 18 entries | **rc 2** | `static_assert` at `shader_histo.cpp:97` |
| D | `scan.py` sentinel/line-count cross-check disabled | — | `--self-test` **exit 1**, new arm red |

A1 with A2 is the load-bearing pair: a change to the **real** decoder now reddens the conformance
suite, and demonstrably could not before.

B is the domain control for the clean zero above. A same-source control would only show the harness
runs; this shows the corpus can *express* a length divergence, so the 0 is a real null rather than a
blind instrument. It also explains the two more realistic drift mutations that showed 0 — dropping
the NSA extra-dword count, and dropping two VOP2 K-literal opcodes: those rules are simply not
exercised by this corpus, which is why the `self_test` NSA arm is hand-built. Recorded in the PR's
Risks section rather than left out.

Docs gate (`check_numbered_table.py`, repo-wide and `--baseline origin/main`) and
`check_merge_result.py`: exit 0. `git diff --check`: clean.

mattias800:
Amended to `09c5cc3c` after the comment above: one docs-only addition, no code change.

`tools/shader_conformance/AGENTS.md`'s `## Ruled out` now also records the **coverage limit of the
differential itself** — the eboot corpus contains no NSA MIMG instruction, so the equivalence result
covers the length rules that corpus exercises and not the NSA path (which is why the `self_test` NSA
arm is hand-built). That was in the PR's Risks section only, and a limit that lives in a PR body is
one the next agent re-derives at full cost.

`git diff 98c16104 09c5cc3c` touches exactly one file, `.md` only. Build, ctest (335/335, exit 0),
self-test (37 arms), both equivalence runs and all five mutation arms in the comment above were
produced from code identical to this head.

mattias800 (commented):
**APPROVED.**

## The second bug is the more valuable half, and I verified it exactly

`shader_histo.cpp`'s `fn[]` name table has **18 entries** against a **19-value** `Rdna2Format`.
Checked on `main`: the enum runs `… FLAT, EXP, VOP3P, Unknown`, while the table ends
`… "FLAT","EXP","UNK"`. So index 17 (`VOP3P`) prints as `UNK`, and index 18 (`Unknown`) reads **past
the end of the array**. The segfaults follow, and 4 of 55 dumps crashing is the visible half of an
out-of-bounds read that was silently mislabelling 29,776 instructions on one title before that.

`VOP3P` was inserted before `Unknown` and the copy never followed — which is the *same defect class*
as #3184 itself: a hand-maintained copy of something the decoder already knows. Finding the general
case while fixing a specific instance of it is the good kind of accident.

The fix is right because it is structural: `static_assert(std::size(fn) == Unknown + 1)` makes the
drift impossible rather than merely correcting today's data, and mutation C (build **rc 2**) shows
the assert actually fires. Correcting the table without that would have left the next person free to
repeat it.

## The premise was checked before anything was built on it

The duplication was real (~90 lines: the VOP src0 modifier/literal rules, the six K-carrying VOP2
mul-adds, SOP literal fields, `S_SETREG_IMM32_B32`, the VOP3/VOP3P markers, the two-dword bucket,
the MIMG NSA count). And the most valuable question in the task — *did the two implementations
already disagree?* — came back **negative: 0 disagreements over 11,266 shaders and 199,521 MIMG
sites**. That is worth stating plainly, because it means no historical scan finding is invalidated,
and a negative answer to that question is a result rather than an absence of one.

## Mutation B is why I trust the zero

This is the part that separates a careful lane from a lucky one. The report says the first two
mutations — dropping NSA, dropping two VOP2 K-literal opcodes — **also produced 0 disagreements**,
and that this looked like a working null but was a corpus that could not express those classes.

That is this project's sharpest recorded trap: *a positive control drawn from the same source as the
null tests the discriminator, never the domain.* Catching it unprompted, and then building mutation B
(**5,680 of 11,266 disagree across 29 titles**) as a genuine domain control, is what makes the
0-disagreement result mean something instead of being an artifact. Recording that limit in the doc
rather than only the PR body is the right home for it.

The A1/A2 pair is the other load-bearing arm: the same decoder mutation reddens the new suite
(ctest exit 8) and the **pre-change** scanner reports `PASSED`. That is direct evidence the old
arrangement was blind to exactly the class it claimed to check.

## The deviation from the issue is correct

The issue names `shader_histo`; the PR uses `shader_inspect`. That is right, and for a stated
reason: `shader_histo` consumes an *eboot*, while `scan.py` holds raw dumps — which is
`shader_inspect`'s input, and it already walks them with `rdna2_walk`. Following the issue literally
would have meant either a second extraction step or feeding a tool the wrong artifact. Keeping the
default output untouched and putting the new behaviour behind `--mimg-sites` keeps existing callers
unaffected.

## Two behaviour changes, both acceptable and both declared

A raw dump that is empty or not a whole number of dwords is now **not-readable** rather than silently
truncated and scanned — strictly better, since the old path produced a confident answer about a
malformed input.

The self-test losing hermeticity is a real cost and the one I weighed longest. It is justified here:
the whole point is that the scanner no longer decodes anything, so a self-test that could still pass
without the real decoder would be testing the wrong thing. Critically it **fails closed** — exit 2
without `shader_inspect`, never a pass — so the failure mode is a visible refusal rather than a
green run that checked nothing.

## Evidence

Build exit 0; `ctest --no-tests=error` **335 discovered, 335 passed, exit 0**, against a baseline the
lane measured itself in the same worktree (335/335/0) rather than assuming. Both cases confirmed by
name. Two independent equivalence checks: the ISA walk over all 55 dumps (11,266 agree, 0 disagree)
and the whole scanner over 14 stub-capture combinations with `--json` and exit code compared verbatim
(14 identical). 225 lines of duplicated decoding removed.

Merging once the remaining build jobs land.

Issue #3231, opened 2026-09-02
Astro Bot (PPSA21564): the world map's full-screen lighting consumer 0x500571000 is declined by the CFG dispatcher's missing-entry-vcc gate


mattias800:
## Summary

Astro Bot's world-map compute program `0x500571000` is declined on every dispatch, and the reason
is a recompiler gate rather than anything about its resources:

```
[compute] skip unsupported program 0x500571000 reason=compute-cfg-reject pc=314 reason=missing-entry-vcc
```

That is `reject_cfg(ins.front().pc, "missing-entry-vcc")` in
`prosper/src/gpu/recompiler/rdna2_emit_cfg.cpp` — the CFG dispatcher refuses to initialise its
`vcc_var` when the entry `RegState` carries no VCC value and the Wave32 mask-domain proof
(`proven_wave32_masks`) does not hold:

```cpp
if (!initial.vcc && !proven_wave32_masks)
    return reject_cfg(ins.front().pc, "missing-entry-vcc");
b.store_function(vcc_var, initial.vcc ? initial.vcc : no);
```

The gate is right to exist — outside a proven Wave32 mask domain the dispatcher has no runtime type
tag for a physical VCC word recycled as scalar data, so persisting `false` could feed a fabricated
zero to a consumer that expected data. It is simply stricter than it needs to be for this program.

## What the program is

`PROSPER_COMPUTE_PROGRAM_CENSUS=1` on a routed 190 s boot: `executed=0 skipped=830`. Its
`[compute-table]` dump is fully resolved — 40 bindings, none unmatchable — and the shape says what it
is: sixteen full-screen storage surfaces (`8294400`- and `16588800`-byte, i.e. 3840x2160 x1 and x2,
plus `2073600` for 1920x1080), the 4,147,200-byte per-tile light-list arena at `0x539c60000` that
#3214 restored, and a 16.5 MB buffer at `0x514080000`. It is a full-screen lighting/composite
consumer, not a small helper.

## The entry VCC is provably dead here

Dumped with `PROSPER_SHADER_DUMP` and decoded with `tools/shader_inspect`
(`exec_cs_500571000.bin`, 3596 dwords, 2374 instructions):

- the CFG region starts at **pc 314**;
- the first VCC **write** in the region is **pc 322** (`VOPC op=0x0c3 dst=special:106`);
- the first VCC **read** in the region is **pc 323** (`SOP2 op=0x01b src0=special:106`);
[...]

Issue #3218, opened 2026-09-02
Astro Bot (PPSA21564): the world map now renders but is blown out to white, runs at 2.2 fps, and still declines 0x500571000


mattias800:
## Summary

With the world-map GPU reset fixed (#3214 / PR #3217), Astro Bot's world map renders and animates
for a whole 150 s run — but it is not correct. Two defects visible in the same frames, plus one
still-declined compute program.

Screenshots: `BLOG.md`, 2026-09-02 entry (`assets/screenshots/astro-bot-worldmap-lit.png` and
`astro-bot-worldmap-lit-later.png`).

## 1. Geometry blown out to white

At the 120 s and 150 s samples a large panel fills the lower half of the frame at full white with a
thin red/yellow fringe, and several shards read as pure white rather than lit surfaces. The 90 s
sample is much closer to plausible (a dark starfield with pale-blue shards and light shafts), so
whatever is over-bright arrives with a later camera or a later set of objects.

The light lists themselves are now well-formed — `PROSPER_DRAW_LINKSCAN` on `0x5008f1400` reports
`terminating=4147200 cyclic=0` with exactly 32,400 terminator dwords, one per screen tile — so this
is downstream of the culling pass, not the same defect.

## 2. Frame rate falls to 2.2 fps

The fixed run reports `2.2 fps while producing frames, 48% of the 150.5 s run active` (2,916 frames
in 150 s) against roughly 9 fps on the control before its device loss. This is expected in
*direction* — five previously-declined compute dispatch families now do real work, one of them
240x135x1 groups at 8x8 local size — but the magnitude is unmeasured and nobody has looked at where
it goes. `PROSPER_PERF_CAPTURE_AFTER_MS` aimed at the world-map load is the obvious first step.

## 3. `0x500571000` is still declined

`PROSPER_COMPUTE_PROGRAM_CENSUS=1` on the fixed run shows one program still skipping every dispatch:
`program=0x500571000 executed=0 skipped=239 groups=1x1x1 local=16x16 threads=16x16`. It appears in
**no** control run's census, so it is reached only now that the earlier passes run. Its reason is not
`invalid descriptor contract` (there are zero of those with the fix) — it has not been identified.

## Reproduce

```bash
cd <REPO_ROOT>/prosper
PROSPER_GUEST_ARGS=-force-gfx-direct PROSPER_RENDER=1 PROSPER_VULKAN_LIB=libvulkan.so.1 \
[...]

Issue #1742, opened 2026-08-02
Astro Bot: compute program 0x500571000's groups_x accumulates instead of resetting — 752,646 workgroups (192.7M invocations) in one dispatch


mattias800:
Found while decomposing #1732. This is the dominant single cost in Astro Bot (`PPSA21564`) and, on the
evidence below, a correctness defect in the dispatch-size derivation rather than a tuning problem.

## The observation

`PROSPER_COMPUTELOG_CODE=0x500571000` prints the derived launch geometry per dispatch. `local` is a
constant `16x16x1`. `groups_x` is not:

| dispatch # | groups_x | invocations |
|---:|---:|---:|
| 1, 2 | 1 | 256 |
| 3..22 | 2, 3, 4, … (delta **1**) | |
| 23 | 3,366 | 862 k |
| 40 | 60,231 | 15.4 M |
| 120 | 327,831 | 83.9 M |
| 240 | 729,231 | 186.7 M |
| 247 (last traced) | **752,646** | **192.7 M** |

From dispatch 23 onward it increments by **exactly 3,345 workgroups every dispatch and never
decreases**. A per-frame quantity (~3,345 groups, 856 k invocations) is being **accumulated instead of
consumed** — the derived count is a running total where the guest means a per-frame value.

## Why it matters

- This one program is **75.2 % of all succeeded compute-dispatch time** in a routed run, at **982 ms
  mean per dispatch**, 93 % of it inside `vkQueueSubmit` + `vkWaitForFences`.
- Its cost grows monotonically within a single run: **235 ms then 562 ms then 982 ms then 1,492 ms**.
  The title gets progressively slower the longer it runs, which is why no rendered session has survived
  long enough to pass the world map (#1732, #1459).
- Compute execution is ~93 % of headless wall time on this title, so this is the bulk of it.

## Where to look

`item.launch.groups_x` / `threads_x` reach the backend already derived, so the defect is upstream of
`live_compute.cpp` — in the kernel-derived thread/local/group contract established by #580, or in
whatever guest value that derivation reads. `local` (16x16x1) is stable and correct-looking; only the
group count drifts, and it drifts by a constant, which points at a counter read as a cumulative total.

Note the first 22 dispatches genuinely read `groups=1x1x1`, so a short trace or a small sample shows a
healthy-looking constant geometry. Plot the whole series.

## Reproduce

```sh
[...]

PR #3233, opened 2026-09-02
agc: sceAgcAcbDispatchIndirect carries an ADDRESS, not an offset — the async-compute ring has no base register

mattias800:
## What this fixes

The async-compute ring has **no indirect-argument base register**. libSceAgc 3.20 exports 36
`sceAgcAcb*` entry points and not one of them is a SetBase — the library's only setter is
`sceAgcDcbSetBaseIndirectArgs`, which takes a `Dcb`. So there is nothing on that ring for a byte
offset to be relative to, and `sceAgcAcbDispatchIndirect`'s argument is necessarily a whole 64-bit
GPU virtual address.

prosper registered **one builder for both NIDs**. `CtB+A9-VxO0` (`sceAgcDcbDispatchIndirect`) is the
offset form and was correct; `j3EtxFkSIhQ` (`sceAgcAcbDispatchIndirect`) went through the same
builder, whose first payload dword is `(uint32_t)a1`. That truncated the address to 32 bits, and the
command processor then added the DCB's compute base — zero on that ring. Every async indirect
dispatch resolved to an unmapped low address and was **skipped for the life of the process**, with
only `[agc] indirect dispatch skipped: unreadable arguments at 0x…` to show for it.

This is not Astro-specific. Any title that dispatches indirectly from the async-compute queue lost
every one of those dispatches.

## Evidence that the payload is an address

Astro Bot's own stream, in one run, is decisive. Its `q=2` DispatchIndirect packets carry exactly the
**low 32 bits** of the argument allocations its `Dcb` announces at **full width** in the same run:

| ACB packet payload | Dcb `SetBaseIndirectArgs` in the same run |
| --- | --- |
| `0x074063c0` | `0x5074063c0` |
| `0x074063e0` | `0x5074063e0` |
| — | `0x50740a520`, `0x50740a540` |

`0x074063c0` is unmapped; `0x5074063c0` is mapped, and #1742 independently established it as the
address `resolve_indirect_dispatch_arguments` reads the group count from, written by compute program
`0x5006eac00`. So the upper half is a fact about this packet, not a guess about an aperture.

## Approach

The ACB form gets its own reserved sub-opcode (`R_DISPATCH_INDIRECT_ADDR`) and its own **5-dword**
size, rather than sharing an encoding whose payload cannot hold an address. prosper answers
`sceAgcAcbDispatchIndirectGetSize` as well as building the packet, so the guest reserves exactly what
is written, and `test_agc_getsize` pins both numbers. The command processor uses the address verbatim
for this form and never adds a base; the opt-in `PROSPER_INDIRECT_APERTURE_RECOVERY` heuristic is
excluded from it, because an absolute address needs no aperture guessed for it.

The first commit is diagnostics: two bounded, always-on lines at the two points where an indirect
dispatch can be silently disabled. The executor's existing message prints only the summed address,
which cannot separate "the base arrived truncated" from "the base was never set on this queue" — and
those have different fixes. Both were previously reachable only through `PROSPER_INDIRECTLOG`, whose
per-line cap is spent in the first second of an Astro Bot boot. The readability probe runs only while
its 24-line bound is unspent.

## Measured

`tools/screenshot`, native 3840x2160, no pad script, `PROSPER_GUEST_ARGS=-force-gfx-direct
PROSPER_RENDER=1`, 190 s, five samples.

| | before | after |
| --- | --- | --- |
| `indirect dispatch skipped: unreadable arguments` (capped at 24) | **24** | **0** |
| `indirect-args base TRUNCATED` | 0 | 0 |
| distinct fps | 5.7 | 5.7 |
| 90 / 120 / 150 s frames | identical | identical |

With `PROSPER_INDIRECTLOG=1`, the recovered address is logged as
`outcome=ready args=0x5074063c0 code=0x500571000` — the exact address the `Dcb` announces. Within the
log's first 256 resolutions, the composition changes from 256 of one program on one allocation to
four consumer programs across two allocations, one of them `code=0x5006f6c00 dims=32400x1x1`, which
is one workgroup per screen tile.

**This does not change what Astro Bot's world map looks like**, and the table says so. The program
that motivated the investigation, `0x500571000`, is still declined — by the recompiler, for an
unrelated reason (`compute-cfg-reject pc=314 reason=missing-entry-vcc`), filed as #3231. Both defects
had to be fixed before it could run; this is one of them.

## Regression and mutation arm

`test_command_processor` asserts the ACB packet resolves to `0x5074063c0` — Astro Bot's own address,
chosen **because its upper half is nonzero**: a truncating builder still reproduces the low 32 bits,
so an address that fitted in 32 bits would pass either way. The test process's own heap lands below
4 GiB (measured `0x345cccd0`), so an ordinary allocation could not have served, and the first draft of
this test could not have failed. A companion arm keeps the DCB form base-relative, so the split
cannot be faked by making every `DispatchIndirect` absolute.

**Mutation:** re-registering `j3EtxFkSIhQ` on the DCB builder (build exit 0) makes the new assertion
report `acb args_addr=0x74063c0 want=0x5074063c0` — the exact address the live run prints — and
reddens `test_agc_getsize`'s pinned 5-dword row. Restored, both pass.

## Deliberately unaffected

- `sceAgcDcbDispatchIndirect` keeps its offset semantics and its 4-dword packet.
- `DrawIndexIndirect` (graphics) is untouched: `sceAgcDcbDrawIndexIndirect` is a Dcb entry point with
  a real base, and no title evidence exercises a base-less graphics indirect draw.
- `PROSPER_INDIRECT_APERTURE_RECOVERY` stays opt-in and stays as it is for the base-relative form.

## Risk

The ACB packet grows from 4 to 5 dwords. prosper answers the matching `GetSize`, so a guest that
sizes its buffer from the API reserves the larger amount; `test_agc_getsize` fails if the two ever
drift apart. A guest that hard-codes 16 bytes for this packet would overrun its reservation — no such
title is known, and the same argument applies to every other row in `AGC_PACKET_SIZES.md` where
prosper emits more than hardware.

## Verification

Distrobox `ps5ys`, Linux, `-DPROSPER_APP=ON`.

```
cmake --build prosper/build-linux -j6        # exit 0
ctest --no-tests=error                       # exit 0, 336/336 passed, 90.04 s
```

No snapshots run: the change moves no guarded title's route.

Refs #3218, #3231, #1809, #1459


mattias800 (commented):
**APPROVED** (with one correction already pushed as `fa696132` — see the end).

## The finding is real and the identity proof is the good kind

`sceAgcAcbDispatchIndirect` was registered on the **Dcb** builder, whose first payload dword is
`(uint32_t)a1`. On the async-compute ring there is nothing for an offset to be relative to, so that
argument is a whole 64-bit address — and it was being truncated to 32 bits and then resolved against
a base that is zero on that ring. Every async indirect dispatch in the title landed on an unmapped
low address and was skipped **for the life of the process**. Cross-title, not Astro-specific.

What makes this convincing rather than plausible is that the title's own stream proves the identity:
the `q=2` payloads `0x074063c0` / `0x074063e0` are exactly the low halves of the allocations the
**Dcb announces at full width in the same run** (`0x5074063c0`, `0x5074063e0`, `0x50740a520`,
`0x50740a540`). That is the guest telling you the upper half was dropped, rather than an inference
from the symptom.

I verified the supporting claim independently against `../PS5-3.20_Libs/libSceAgc.c`: **not one**
`sceAgcAcb*` export contains "Base", while `sceAgcDcbSetBaseIndirectArgs` does exist and takes a Dcb.
So the absence of an async-ring base register is a property of the 3.20 export table, exactly as the
comment claims.

## The packet-size half is handled, which is the part that would have bitten later

`AGC_PACKET_SIZES.md` exists because a builder's dword count is an ABI contract with the guest's own
reservation. This adds a **distinct opcode** (`R_DISPATCH_INDIRECT_ADDR = 0x24`) at **5 dwords**
rather than overloading the Dcb form, and the comment states that prosper answers
`sceAgcAcbDispatchIndirectGetSize` too, so the guest reserves exactly that. Sharing the opcode would
have made the two forms indistinguishable to the decoder and left the size answer wrong for one of
them.

## Not reproducing #3218 was the right call, and reporting it as a gap is righter

Four runs — three on current `main` including #3218's exact command, one rebuilt on **#3217's own
head** — all land on the pre-existing #1459 nebula state, never the lit-shard frames the issue
describes. Mean luma 96.1 against the committed background image's 98.6; zero `GRAPHICS submission
failed`, so #3217 is present and working; light lists well-formed at 4,147,200 records, 0 cyclic.

Declining to run an F8 capture because "profiling a regime that is not the one the issue describes
would have produced a number worth nothing" is exactly right, and it is the discipline this project's
trap list exists to enforce. So is recording it as a **reproduction gap rather than a falsification**
— the state was observed once by the lane that filed it, and "I could not get there" is a different
claim from "it does not happen".

## The self-caught positive control deserves calling out

The first draft of the test used a heap allocation that landed at `0x345cccd0` — below 4 GiB, so it
**could not have failed** whatever the code did. Catching that and switching to Astro's real address
*because its upper half is nonzero*, then keeping that arm in the test, is the difference between a
test that passes and a test that discriminates. This is the same trap that has produced phantom
results here before.

## Evidence

Build exit 0; `ctest --no-tests=error` **336/336, exit 0**. The mutation arm (re-registering the NID
on the Dcb builder, build exit 0) reproduces the live symptom in the assertion —
`acb args_addr=0x74063c0 want=0x5074063c0`, the exact address the running title printed — and reddens
the pinned dword-count row in `test_agc_getsize`. 24 `unreadable arguments` events before, **0**
after, with `PROSPER_INDIRECTLOG` confirming `outcome=ready args=0x5074063c0`.

The PR states plainly that **this changes nothing visible on this title** — frames and fps identical
before and after. That honesty is worth more than a claimed improvement would have been; #3231 is why
the program still does not run.

## The one correction

The comment said libSceAgc 3.20 exports **36** `sceAgcAcb*` entry points. It is **35** — counted two
ways (unique `sceAgcAcb*` symbols, and `.global sceAgcAcb*`). The argument is untouched by this, but
a number in a committed comment is inherited by everyone who reads it, and this file is precisely the
one a future reader would cite. Fixed in `fa696132` rather than merged as-is.

PR #3234, opened 2026-09-02
worktree hygiene: refs/stash is shared by every worktree -- park work per-worktree instead (#3174)

### LINKED ISSUE #3174: git stash races across dedicated per-agent worktrees (refs/stash is repo-wide, not worktree-local)

mattias800:
Fixes #3174

## The mechanism

`CLAUDE.md` tells every agent to work in its own `git worktree`, and a worktree really does isolate
`HEAD`, the index and the working tree. It does **not** isolate `refs/stash`. That is a single ref
in the shared `.git` common directory, so every worktree of this repository pushes onto and pops
from **one LIFO stack**.

So two agents can each do exactly what the charter prescribes -- own tree under
`.claude/worktrees/<slug>`, own `fix/issue-NN-*` branch -- and still collide. On 2026-09-01 they
did: lane A stashed one file so it could rebuild without it and confirm a regression arm failed;
lane B stashed in its own tree a moment later; lane A's `git stash pop` took the top of the stack,
which was lane B's, applied an unrelated `hle_kernel.cpp` change into lane A's tree and **dropped
lane B's entry** in the process. The measurement lane A was about to take would have run against a
tree carrying a second lane's edit, under a banner saying only its own change had been removed.

Nothing was lost only because **both** agents happened to recognise foreign content and recover by
hand. A dropped stash commit is unreachable and GC-eligible, so the silent version of this loses
another lane's work with no trace, or merges two issues' changes into one branch.

Reproduced here rather than taken on trust -- the first arm of the new test does it on demand:

```
[arm] plain git stash across two worktrees (positive control for #3174)
  ok   refs/stash is the SAME stack in both worktrees
  ok   both entries are on the one shared stack
  ok   lane A pops lane B's content into its own worktree
  ok   lane B's entry was consumed by lane A's pop
```

## Why there is no automated guard, and what there is instead

**A `git` alias cannot guard `git stash`.** Measured on git 2.55.0: with
`git config alias.stash '!echo GUARD-FIRED'` set, `git stash push` ran the builtin, saved an entry,
and the alias body printed nothing at all -- git does not let an alias shadow a builtin command. A
wrapper of that shape would be strictly worse than none: it reads as coverage and cannot fail,
which is the trap this repository keeps a whole numbered list about.

**A `reference-transaction` hook does work** -- also measured: exiting non-zero in the `prepared`
phase for `refs/stash` refuses the update cleanly, leaving the working tree untouched and creating
no ref, so there is no half-stashed state. It ships as `wt_stash.py install-hook` and is
deliberately **off by default**, because it is not something anyone may assume: `.git/hooks` is
untracked, so every fresh clone has no hook and fails open, and once installed it blocks `git stash`
repo-wide -- every worktree, the human's own interactive use, and `--autostash` with it. It is an
opt-in convenience for one machine, never the reason the collision cannot recur.

So the thing that actually prevents recurrence is the written rule **plus a replacement an agent
will reach for instead**, which is the half a prohibition alone does not supply.

`tools/wt_stash.py` parks uncommitted work in a ref that really is private to one worktree. It
snapshots with `git stash create` -- which writes no ref and touches no working tree -- and stores
the result under `refs/worktree/prosper-stash/<slot>`. `git-worktree(1)` REFS: *"refs inside
refs/bisect, refs/worktree and refs/rewritten are not shared"*. Three properties make it
un-race-able where `git stash` is not:

* the ref is per-worktree, so there is no shared stack to interleave on;
* a slot is addressed **by name**, never by top-of-stack, so no pop can pick up a neighbour's entry;
* `push` **refuses** an occupied slot instead of stacking, so nothing is ever displaced silently.

`wt_stash.py check` covers the other direction: it reads the shared `refs/stash` and says whether
what is sitting on it is yours, **before** you touch it -- failing closed on any entry it cannot
attribute to a branch, the same shape as `lanekill.py` and `worktree_reclaim.py`.

Two tool-free alternatives are documented alongside it and are equally safe: a temporary commit on
your own branch (nothing outside `refs/heads/<your-branch>` moves), and a second `git worktree add`
when you want two states at once.

## What changed

* **`CLAUDE.md`** -- the worktree section now states the non-isolation plainly, with the incident,
  the replacements, and an explicit note that nothing enforces it. Two existing lines pointed the
  other way and are corrected: a cross-reference into the harness "environment preamble", which a
  reader of `CLAUDE.md` cannot see, and the teardown instruction *"commit, stash or discard before
  you leave a tree"* -- stashing is the worst of those three, because `worktree_reclaim.py` never
  looks at `refs/stash` (grep it), so a tree whose only work is stashed reads as clean and loses the
  DIRTY protection keeping it alive.
* **`prosper/tools/wt_stash.py`** (new) -- `push` / `pop` / `apply` / `list [--all]` / `drop` /
  `check` / `install-hook` / `uninstall-hook`. `push` also refuses while untracked files are
  present, because `git stash create` cannot snapshot them and silently leaving them out of a
  "park everything" operation is the data-shaped failure mode this project keeps recording;
  `--allow-untracked` proceeds after printing the list.
* **`prosper/tools/test_wt_stash.py`** (new, ctest `wt_stash_tool`).
* **`prosper/tools/AGENTS.md`** -- the tool under `## Shared-box hygiene`, plus a new
  `## Ruled out` section carrying the two measured guard results above.
* **`prosper/docs/GAME_COMPAT_ORCHESTRATION.md`** -- instrument trap **247**.

## Verification

Baseline measured on this branch's base, `origin/main` @ `21c771ae`, in the ps5ys container:

```
cmake -S prosper -B prosper/build-linux                       # configure rc=0
cmake --build prosper/build-linux -j6                         # build rc=0
ctest --no-tests=error
    100% tests passed out of 336        rc=0
```

Same tree with this commit applied:

```
cmake --build prosper/build-linux -j6                         # build rc=0                    
ctest --no-tests=error
    100% tests passed out of 337        rc=0
    9/337 Test #9: wt_stash_tool ...   Passed    2.20 sec
```

Test count 336 -> 337; the added case is `wt_stash_tool`, verified by name above. Nothing else in
the diff is compiled code.

Docs gate, against the live base (`origin/main`):

```
    --baseline <origin/main copy> prosper/docs/GAME_COMPAT_ORCHESTRATION.md
    numbered column 1..247, 247 rows, unique and ascending; none of the 246 base rows deleted  rc=0
git ls-files '*.md' -z | xargs -0 python3 prosper/tools/docs/check_numbered_table.py   rc=0
git diff --check                                                                       rc=0
```

### The mutation arms

The suite's first arm is a positive control constructed **by hand from git primitives**, outside
the tool under test: two linked worktrees, plain `git stash` in each, and an assertion that lane A's
pop really does hand it lane B's content. Every later arm claims the tool removes that failure, and
none of them means anything if the fixture cannot produce it -- so a failure there reports **VOID**
rather than a red arm.

Eleven mutations of `wt_stash.py`, each re-run against the unmodified suite (and all re-run after
the Windows fix below). Every one reddens, and each names the property it broke rather than merely
exiting non-zero:

| mutation | caught by |
| --- | --- |
| `push` writes the shared `refs/stash` | `lane A parked into the per-worktree namespace`, `the shared refs/stash was never written` |
| `push` overwrites an occupied slot | `second push into the same slot is refused`, `the first snapshot was not displaced` |
| `pop` falls back to any slot when the named one is absent (i.e. `git stash pop` semantics) | `popping an absent slot fails even though others exist`, `slot one untouched` |
| `push` tolerates untracked files silently | `push refused while an untracked file is present`, `nothing was parked by the refused push` |
| `check` never fires | `a foreign entry fires`, `symmetric from the other worktree` |
| `check` always fires | `own entry alone does not fire` |
| `check` treats a foreign branch as yours | `a foreign entry fires` |
| `check` assumes an unattributable entry is safe | `and is labelled unattributable` |
| `list --all` only looks at its own worktree | `--all reaches the other worktree's slot`, `--all names the other worktree` |
| hook refuses **every** ref, not just `refs/stash` | `ordinary commit still works with the hook installed`, `branch creation still works` |
| hook guard deleted (always allows) | `git stash is refused`, `refs/stash was not created` |

The by-name arm exists **because** the third mutation initially passed: every other arm happens to
pop a slot that is present, so a top-of-stack fallback was invisible to all of them -- and a
top-of-stack fallback is precisely #3174's semantics. That is the gap a mutation pass is for.

The hook arm asserts blast radius as well as effect, since a `reference-transaction` hook fires on
**every** ref update in the repository: an ordinary commit and a branch creation must still succeed,
and `wt_stash.py` itself must still work, with the hook installed.

### The platform-independent registration earned itself immediately

The first push of this PR went **red on Windows MinGW**, on a real bug in the tool rather than in
the test: `resolve()` used `git rev-parse --verify --quiet <ref>^{commit}`, and that peeling suffix
does not reach git intact on the Windows/MSYS route this project also prescribes. Every lookup
resolved nothing, so `push` parked correctly and `pop` then reported `no such slot: 'default'` for
every slot. Fixed in `ff168d7d` by dropping the peel -- every value these refs hold is a commit sha
the tool wrote itself, so a plain `--verify` is both sufficient and portable. The same commit
asserts `list --all` on the other worktree's **sha** rather than its path string, because git prints
a Windows worktree path with mixed separators (`D:\a/_temp/...`).

This is why the case is registered platform-independently rather than Linux-gated like
`worktree_reclaim_tool` and `lanekill_tool` above it: nothing here reads `/proc`, the subject is ref
namespacing, and the collision it guards is likeliest on exactly the route that found the bug.

## Risks

* **The rule is not enforced.** An agent that ignores `CLAUDE.md` and types `git stash` still
  collides. That is stated rather than papered over; see the ruled-out section above for why the
  two available interception shapes are an impossibility and an opt-in respectively.
* **`install-hook` is repo-wide when used.** It blocks the human's own `git stash` and
  `--autostash` too. Off by default, idempotent, refuses to overwrite or remove a hook it did not
  write, and a one-off bypass (`git -c core.hooksPath=/dev/null stash ...`) is documented and
  tested.
* **`push` does not snapshot untracked files**, because `git stash create` cannot. It refuses
  rather than dropping them silently.
* **`pop` restores worktree content, not index state**, matching `git stash apply` without
  `--index`.

## Deliberately unaffected

No emulator, renderer or HLE code is touched; the only build-system change is one `add_test`. No
`BLOG.md` entry: this is worktree/process tooling with nothing to show a reader, and the charter's
blog is for pictures and findings from title work.

## Ruled out (recorded in `prosper/tools/AGENTS.md`)

* A `git` alias cannot guard `git stash` -- git ignores aliases that shadow builtins. Measured on
  git 2.55.0. A guard of that shape cannot fail and would read as coverage.
* A `reference-transaction` hook can abort a stash cleanly, but is not a guard anyone may assume:
  `.git/hooks` is untracked, so it is absent in every fresh clone and fails open.


mattias800 (commented):
**APPROVED.** The strongest thing here is what it *declined* to build.

## The negative result is the load-bearing one, and I reproduced it

The obvious fix is a `git stash` wrapper. The PR tried it and **measured that it cannot work**: git
does not let an alias shadow a builtin, so the guard never runs and the stash lands as usual.

I reproduced that independently on git 2.55.0 — with `alias.stash '!echo GUARD-RAN-INSTEAD'`
configured, `git stash push` printed *"Saved working directory and index state"*, created one entry,
and the alias body produced **no output at all**.

Recording it as `## Ruled out` with the reason — *"a wrapper of that shape is worse than no wrapper,
because it reads as coverage and cannot fail"* — is exactly right, and it is the trap this project
has a whole list about. Shipping that wrapper would have left everyone believing they were protected.

## The fix uses the correct primitive rather than a workaround

`refs/worktree/prosper-stash/<slot>` is not a clever hack — `git-worktree(1)` explicitly documents
that *"refs inside refs/bisect, refs/worktree and refs/rewritten are not shared"*. I checked the man
page. Combined with `git stash create` (writes no ref, touches no working tree), that is the
per-worktree primitive the situation actually calls for.

Two design choices make it safe rather than merely different: slots are addressed **by name, never by
top-of-stack**, and `push` **refuses an occupied slot instead of stacking**. The original defect was
a LIFO displacement, so removing both the stack and the implicit addressing removes the failure mode
rather than narrowing it. Offering the two tool-free alternatives (a temporary commit on your own
branch; a second worktree) is the right shape for a rule people have to follow by hand.

## The second-order finding is the one I would have missed

`CLAUDE.md` said "commit, **stash** or discard before you leave a tree" — and stashing is the one
option that makes things *worse*, because `worktree_reclaim.py` never looks at `refs/stash`. I
grepped it: **zero** references. So a tree whose only work is stashed reads as **clean**, loses the
DIRTY protection that was keeping it alive, and gets reclaimed — while the work survives only as an
unattributable entry on a stack any other lane can pop.

That means the charter's own advice could destroy work, which is a considerably worse bug than the
race that prompted the issue. Finding it while fixing something else is the good kind of accident.

## The test has a positive control, which is what makes it mean anything

The first arm **builds two linked worktrees by hand and asserts the collision still reproduces**, so
every later arm runs against a domain proven capable of expressing the failure. Without that, "the
new tool does not collide" would be unfalsifiable — this project's sharpest recorded trap is a
control drawn from a source that cannot express the case it is validating.

Trap 247 is written the way that list is supposed to be written: the two real branches, the exact
sequence, the recognition cue (*"a `pop` whose diff touches a file you never edited"*), why nothing
was lost only by luck, and the general lesson — any shared repository state the worktree rule is
quietly assumed to cover.

## Honesty about the limits

The PR says plainly that **nothing enforces this**, that `install-hook` is off by default and lives
in untracked `.git/hooks` so it is absent in every fresh clone, and that this is a rule you follow —
in the same family as trap 41's `git checkout <branch> -- <file>`. That is the accurate framing, and
far better than implying the tool closes the hole.

`check` reporting an entry it cannot attribute as `UNATTRIBUTABLE` rather than as yours is the right
fail-closed default.

Merging once CI lands.

Issue #3062, opened 2026-08-27
Design candidate: replace the Windows guest semaphore poll loop with WaitForMultipleObjects over a Win32 semaphore plus a high-resolution timer


mattias800:
A design candidate, **not a defect** — the current code is correct and measured. Raised in review of
#3044 and filed so the option and its cost are findable rather than re-derived.

## The situation

On Windows, guest timed waits resolve on the winpthreads master tick (~15.6 ms, #3013). #3044 fixes
`k_sem_timedwait` by polling: `sem_trywait` plus adaptive high-resolution sleeps (500 µs while the wait
is young, 2 ms after 8 ms). That buys an accurate timeout and **pays for it in wake latency** — a posted
semaphore returns natively in 0.01 ms, and a poll loop cannot beat its own slice.

Measured cost of the poll loop, from #3044's review: roughly **0.1–0.3% of a core per parked thread** in
the coarse regime and **0.5–1%** in the fine one; worst realistic case about a third of a core. Affordable,
bounded, and the reason this is a candidate rather than urgent.

## The better structure

One `WaitForMultipleObjects` over **two handles**: a Win32 semaphore for the post, and a
`CREATE_WAITABLE_TIMER_HIGH_RESOLUTION` for the deadline. That gives an exact timeout *and* an immediate
wake, with no polling and no CPU floor — the two properties the poll loop trades against each other.

**The trap, which is the main reason this issue exists:** a plain `WaitForSingleObject(sem, ms)` timeout is
**not** a fix. Kernel wait timeouts are themselves tick-quantized, so the naive form reproduces the
original defect while looking like it solves it. The high-resolution *timer handle* is what carries the
precision; the wait itself must be untimed.

## Two constraints for whoever takes it

1. **Preserve the #2042 / #2176 retire-then-reclaim semantics.** A destroyed guest semaphore's storage is
   quarantined rather than freed, because a thread may be parked inside it. A handle-backed implementation
   has to keep that property for the handle too — `CloseHandle` on a semaphore another thread is waiting on
   is exactly what the quarantine exists to prevent.
2. **Measure any replacement against the 0.01 ms post wake**, not against the poll loop. The poll loop is
   the thing being improved on; beating it is not the bar.

## Scope

This is not a local change to one entry point. `sem_init` / `post` / `getvalue` / `destroy` all currently
delegate to winpthreads, so a handle-backed semaphore replaces the family. That is why #3044 did not do it.

## Why this is the same decision as #3056

[...]

PR #3235, opened 2026-09-02
sync: scePthreadCondTimedwait spends its relative timeout on the condvar's OWN clock, and the Windows tick attribution in #3056 is wrong

### LINKED ISSUE #3056: sceCondTimedwait overshoots 2.29x on Blasphemous 2 — the cond/mutex half of the winpthreads tick, censused but unfixed
`PROSPER_TIMEDWAIT_CENSUS=1` (added in #3022) reports `sceCondTimedwait` overshooting its requested
interval by **2.29x** on *Blasphemous 2* — measured on Windows 11 / MinGW-w64 UCRT, master at `d2b2e4d3`:

```
[timedwait 5s] sceCondTimedwait calls=11   requested=  0.818 ms  actual=  1.870 ms  x2.29
[timedwait 5s] usleep           calls=490  requested= 18.082 ms  actual= 18.385 ms  x1.02
```

`usleep` on the same run is x1.02 because #3022 routed it through
`host::sleep_until_steady_ns`. `k_cond_timedwait_sce` still delegates to winpthreads'
`pthread_cond_timedwait` (`hle_kernel.cpp`, the Sony spelling with a RELATIVE microsecond timeout), which
is the same family as #3013.

## Why this is filed rather than fixed

#3022 deliberately left `k_cond_timedwait`, `k_cond_timedwait_sce` and `k_mutex_timedlock` on the native
path and censused them instead, on the grounds that a poll loop trades away the fast wake (a *signalled*
condvar returns promptly; only the timeout is quantized) and that no title measured so far paced audio on
them. #3044 does the semaphore half for the one primitive where a title demonstrably does.

This measurement narrows that "no title measured so far" a little: a real title does call it, and it does
overshoot. What it does **not** show is that the overshoot matters — 11 calls in a five-second window is
not a pacing loop, and 1.87 ms absolute is small. So this is a candidate, not a defect with a victim.

## What would decide it

The census now reports per-window rather than cumulatively, so the question is answerable directly: find a
title where `sceCondTimedwait` call counts are in the hundreds per window with a ratio well above 1, i.e.
something actually pacing on it. Absent that, fixing this buys accuracy nobody is asking for and costs the
wake latency #3044's review is scrutinising.

Note the absolute request here (0.818 ms) is **below** the ~0.52 ms high-resolution-timer floor measured
in #3044 by less than a factor of two, so even a precise-sleep poll loop could not serve it much better
than 2x — which is worth knowing before anyone assumes a fix is available.

mattias800:
Fixes #3056.

## What #3056 recorded, and what is actually there

#3056 says `k_cond_timedwait_sce` "still delegates to winpthreads' `pthread_cond_timedwait`", which is
where the ~15.6 ms winpthreads tick would come from. **It does not, and cannot.** On Windows the guest
condition waits go through `interruptible_cond_timedwait`, whose `#ifdef _WIN32` branch has been built
on **`WaitOnAddress`** since the #678/#690 cooperative-exception work (`sync_futex.cpp:350`).
winpthreads' `pthread_cond_timedwait` is only in the POSIX `#else`, which does not compile on Windows
at all. Verified rather than read: an `#error` placed inside that branch fires under
`x86_64-w64-mingw32-g++` (2 diagnostics) and does not fire under the host `g++` (0), so the branch
under discussion is the one that builds there.

**And the 2.29x is fully accounted for without any defect.** `WaitOnAddress` takes a `DWORD`
**millisecond** timeout and the conversion rounds *up*, because a condition wait may never report
`ETIMEDOUT` before its deadline. So the censused 0.818 ms request becomes a 1 ms wait *by contract*,
and the remaining ~0.87 ms is the kernel wait's own tick quantization — the trap #3062 names, that a
kernel wait **timeout** is quantized however precise the requested interval. 1 ms + one tick is
1.870 ms. Nothing in the arithmetic inflates it.

That is the honest reading, so this PR does **not** claim to reduce that ratio, and deliberately does
not add a second polling implementation of the semaphore loop — #3062 asks whoever picks this up to
decide once, and the residue is a primitive choice (a handle-backed wait) rather than a bug to find.
What it fixes is the pile of real defects the census sat on top of.

## The defects fixed

**1. `scePthreadCondTimedwait` ignored the condvar's clock.** Its POSIX sibling `k_cond_timedwait`
resolves `guest_cond_snapshot(c).clock_id` and calls `interruptible_cond_clock_timedwait`; the Sony
spelling hardcoded `CLOCK_REALTIME` through `abs_deadline_us()`. One condition variable therefore
answered the same question two ways depending on which name the guest called — the divergence class
#1873 records between two *libraries*, here between two *spellings* of one.

It matters most for `SCE_KERNEL_CLOCK_VIRTUAL` / `_PROF`, which are **process CPU time**: a guest
asking for an interval of CPU time was given one of wall time, so a mostly-blocked thread times out
far too early. For `_MONOTONIC` it removes a wall-clock-jump exposure the interval never had on
hardware.

**2. The Windows timed wait had no post-condition, and now has one.** It converted the absolute
deadline to a millisecond count once, took a single `WaitOnAddress`, and reported `ETIMEDOUT` for
whatever came back — including a `FALSE` that is not a timeout at all. A guest reads `ETIMEDOUT` as
"it did not happen", so an early one is a false negative, not imprecision. The wait now re-arms until
the interval the guest asked for has actually elapsed, the same shape `sleep_until_steady_ns()`
composes for in #3074 ("a single OS wait is not trusted to satisfy the contract on its own").

**Stated precisely, because the first revision of this PR overstated it:** this is a *post-condition
backstop*, not a fix for a demonstrated early return. **The budget is computed once from
`CLOCK_REALTIME` and then spent against `steady_clock`** — and that split is load-bearing, not an
implementation detail. Re-reading `CLOCK_REALTIME` in the loop, which the first revision did, makes
the exit condition depend on *that* clock's resolution: where it advances coarsely, a sub-millisecond
wait re-arms until it ticks, reintroducing the exact quantization this family exists to remove. That
was the review's first blocking finding, and the fix is structural rather than a measurement, because
nothing in this tree measures mingw-w64's `CLOCK_REALTIME` resolution. What is *not* fixed is stated
at the call site: the initial budget is still a difference of two `CLOCK_REALTIME` readings and
inherits that clock's granularity, exactly as before.

**A wake is still reported as a wake at once, spurious or not** — only a *timeout* re-arms — so the
guest-exception interruption path (`interrupt_guest_wait` wakes this address without changing its
value) behaves exactly as before. **The first `WaitOnAddress` always happens, even for a zero
budget**: it is a value check as well as a wait, and the review caught the first revision skipping it
for an already-expired deadline, which would report `ETIMEDOUT` for a condvar that *had* been
signalled. A non-`ERROR_TIMEOUT` failure still reports `ETIMEDOUT`, which is what the old code did —
changing it to `EINVAL` would be a guest-visible contract change with no way to verify it from here.

**3. `interruptible_mutex_lock`'s contention poll was on `::Sleep(1)`** — 15.554 ms measured on this
toolchain (#3013), the exact primitive #3022 routed the whole guest sleep family off. It now uses
`sleep_until_steady_ns` on the semaphore loop's own `poll_slice_ns` cadence — 500 µs while the wait is
young, 2 ms after 8 ms — reusing #3044's reviewed constants rather than inventing a second set.

**The best reason for this is not the census, and the first revision of this PR led with the weaker
one.** This loop takes **no wait registration** — there is no `register_wait` call in it — so a thread
parked here is invisible to `interrupt_guest_wait` and can notice a pending guest exception only at
its own `dispatch_pending_guest_exception()`, once per poll. **The poll interval therefore *is*
IL2CPP's stop-the-world delivery latency for a thread contending a guest mutex**, which is exactly
what the paragraph already at the top of that function means by keeping the wait "cooperative".
15.6 ms → 0.5–2 ms is a GC-delivery improvement; the censused-interval effect is a side benefit.

**4. The deadline arithmetic was untestable and did not saturate.** `abs_deadline_us()` was inline in
`hle_kernel.cpp` and the millisecond conversion was inline inside `#ifdef _WIN32`, i.e. not compilable
on a POSIX host at all. Both move to `precise_sleep.hpp` beside `sem_poll_step` (#3069) and
`sleep_until_deadline_retry` (#3074), expressed on `int64_t` fields rather than a host `struct timespec`
(#3038's 32-bit-`long` truncation class cannot then be acquired), saturating on a guest-supplied
microsecond count the way `guest_ns_from` and `timeout_ns_from_us` already do.

**This is a Windows defect and the work was done on Linux.** Stated plainly:

| claim | status |
| --- | --- |
| The clock divergence, and its fix | **Verified here and on Windows.** New cross-platform test with a discriminating mutation arm; it also runs and passes natively in the `Windows MinGW` CI job. |
| The deadline arithmetic (rounding, past deadlines, saturation, the round trip) | **Verified here and on Windows.** Fake-clock arms in `test_precise_sleep.cpp`, four mutation arms; also run natively by the same CI job. |
| The Windows re-arm loop and the relock cadence | **Compiled and unit-exercised on Windows, but never driven by a guest.** `.github/workflows/ci.yml`'s `windows-mingw` job is `runs-on: windows-latest` under MSYS2/UCRT64 and runs the real `ctest`; on this branch's first head it reported `100% tests passed out of 260` with `kernel_cond_timedwait_clock` **Passed 0.66 s**. What no CI job does is run a *title*, so the behaviour under real contention is unmeasured. |
| The tick's effect on the wait itself | **Not measured anywhere.** mingw-w64's `CLOCK_REALTIME` resolution is not measured by this tree; the second commit makes the loop independent of it precisely so the fix does not rest on a number nobody has. |
| Any change to the 2.29x census ratio | **Not claimed.** Not measured here, and argued above to be mostly irreducible without a different wait primitive (#3062). |

**What would verify the unverified rows**: run `PROSPER_TIMEDWAIT_CENSUS=1` on *Blasphemous 2* on the
Windows 11 / MinGW-w64 UCRT host #3056 was filed from, on this branch and on `main`, and compare the
`sceCondTimedwait` and `mutex_timedlock` rows. The expected change is small and is in the relock, not
the timeout: defect 3 removes up to one ~15.6 ms `::Sleep` tick from any wait whose relock lost a race.
A five-line probe printing successive `clock_gettime(CLOCK_REALTIME)` deltas on that host would settle
the resolution question outright.

## Linux impact: no-op on the common path, fix on the uncommon one

- **`interruptible_cond_timedwait` / `interruptible_mutex_lock`**: entirely inside `#ifdef _WIN32`.
  POSIX keeps `pthread_cond_timedwait` and `pthread_mutex_lock` untouched. **No-op.**
- **`k_cond_timedwait_sce` with a realtime condvar** (the default, and every title that never calls
  `scePthreadCondattrSetclock`, including the Blasphemous 2 run #3056 was filed from):
  `interruptible_cond_clock_timedwait` forwards a realtime deadline straight to
  `interruptible_cond_timedwait`, which is the identical call the old code made. **No-op.**
- **`k_cond_timedwait_sce` with a virtual/prof/monotonic condvar**: it now waits on the clock the
  guest chose. This is the only behavioural change on Linux, and it is **not free**: a non-realtime
  clock goes through `interruptible_cond_clock_timedwait`'s slice loop, so the thread wakes every
  `min(50 ms, remaining)` instead of parking once. Correct where it was wrong, at the cost of up to
  20 wakes a second on a long wait — the same trade its POSIX sibling already makes, and the reason
  this PR does not simply route the realtime case through the loop too (raised in review).
- **`abs_deadline_us`**: same arithmetic, now saturating at the top of the range. On this platform the
  overflow arm is unreachable (`UINT64_MAX` µs is 5.8e5 years of seconds against `int64`'s 2.9e11), so
  it is a consistency guard rather than a live fix, and the comment says which it is.

## Verification (exact head `a252272b`)

Baseline measured on `origin/main` `21c771ae` in the same worktree and build directory, by reverting
the diff and rebuilding:

```
# baseline, origin/main 21c771ae
cmake --build prosper/build-linux -j6                 # exit 0
ctest --no-tests=error -j4                            # exit 0 — 100% tests passed out of 336

# this branch, before the origin/main merge below
cmake --build prosper/build-linux -j6                 # exit 0
ctest --no-tests=error -j4                            # exit 0 — 100% tests passed out of 337

# this branch, current head
cmake --build prosper/build-linux -j6                 # exit 0
ctest --no-tests=error -j4                            # exit 0 — 100% tests passed out of 338
```

`336 → 337` is `kernel_cond_timedwait_clock`, this PR's new case. **`337 → 338` is not this PR**: the
branch carries a merge of `origin/main` (`6ea482ac`), which brought `wt_stash_tool` from #3234/#3238 —
the "synchronize with the live target branch" step before merging. The delta is called out because an
unexplained test count is exactly the kind of thing nobody re-derives.

**An earlier revision of this paragraph justified that with "the merge touches none of the files this
PR changes", and that was false** — `prosper/CMakeLists.txt` is in both diffs, because #3234 registered
`wt_stash_tool` in the same file this PR adds `kernel_cond_timedwait_clock` to. So the one file where a
merge could have eaten someone's work was exactly the file the check claimed was untouched. It is
recorded rather than quietly corrected because the conclusion was right and the derivation was not,
which is the shape nobody re-runs. The check that actually answers it reads the **deletion column**:

```
git merge-base origin/main HEAD   # 56612f2d == origin/main, so origin/main...HEAD is exactly this PR
git diff origin/main...HEAD --numstat
  14   0  prosper/CMakeLists.txt          <- pure addition; wt_stash_tool still registered at :219
  27   0  prosper/docs/PORTING.md
  55  10  prosper/src/hle/kernel/hle_kernel.cpp     <- this PR's own edits
  30   0  prosper/src/hle/sync/AGENTS.md
 102  10  prosper/src/hle/sync/sync_futex.cpp       <- this PR's own edits
 100   0  prosper/src/host/platform/precise_sleep.hpp
 189   0  prosper/tests/hle/kernel/test_kernel_cond_timedwait_clock.cpp
 208   0  prosper/tests/host/platform/test_precise_sleep.cpp
```

Every deletion is inside the two files this PR rewrites; nothing from either side was lost.

Both new/extended cases run and pass by name:

```
286/337 Test #272: precise_sleep_deadline_retry ...................   Passed    0.00 sec
337/337 Test  #55: kernel_cond_timedwait_clock ....................   Passed    0.62 sec
```

Windows cross-compile of every touched source (container `x86_64-w64-mingw32-g++ -std=c++20
-fsyntax-only`): `sync_futex.cpp` **0**, `hle_kernel.cpp` **0**, `test_precise_sleep.cpp` **0**,
`test_kernel_cond_timedwait_clock.cpp` **0**.

Docs gate: `check_numbered_table.py` over every tracked `.md`, exit **0**. `git diff --check`, exit **0**.

## Mutation arms — five, each with its build exit code

Every mutation was applied by a script that **asserts its own pattern matched** before writing, so a
regex that silently failed to apply cannot be reported as "zero failing arms".

**M1 — revert `k_cond_timedwait_sce` to the realtime-hardcoded body.** Build exit **0**;
`ctest -R 'kernel_cond_timedwait_clock|precise_sleep'` exit **8**, `1 tests failed out of 2`:

```
55 - kernel_cond_timedwait_clock (Failed)
   (virtual 300 ms CPU-time wait returned 0x8002003c after 300.17 ms of wall time)
   [FAIL] a SCE_KERNEL_CLOCK_VIRTUAL timeout is spent in CPU time, so the idle wait is signalled
   [FAIL] ...and specifically did NOT time out, which is what the wall-clock deadline did
```

against the fixed build's `returned 0x0 after 600.68 ms`. That is the whole discriminator in one
line: the defect spends the interval on the wall clock and times out at 300 ms; the fix spends it in
process CPU time, which an idle process cannot reach, so the signal at 600 ms arrives first.

**M2 — `wait_timeout_ms_ceil` truncates instead of rounding up.** Build exit **0**; ctest exit **8**,
`272 - precise_sleep_deadline_retry (Failed)`, 4 arms red including *"the 0.818 ms request from
#3056's census rounds UP to a 1 ms wait, never down to 0"* and *"the round trip yields exactly
ceil(request), so no wait is ever short"*.

**M3 — delete `abs_deadline_from_rel_us`'s saturation guard.** Build exit **0**; ctest exit **8**,
`272 - precise_sleep_deadline_retry (Failed)`, 2 arms red: *"a maximal timeout from a near-maximal
reading saturates instead of wrapping negative"* and *"...so the deadline stays in the FUTURE"*.

**M4 — `wait_timeout_ms_ceil` adds then clamps, instead of bounding before the add** (the pre-review
expression). Build exit **0**; ctest exit **8**, 2 arms red. **Its first version did not redden**, and
that is worth reporting rather than quietly fixing: the arm used a 1 ns remainder, whose ceiled
millisecond fits inside the 615 ms of headroom the seconds bound leaves, so the input could not reach
the case the arm was named for. The input is now derived from the bound rather than guessed.

**M5 — delete the `seconds == 0` guard on the past-deadline borrow** (the loop's termination
condition). Build exit **0**; ctest exit **8**, `[FAIL] a deadline one nanosecond in the past has been
reached`. Added because the review noted arms 12/15/16 had no mutation coverage.

**M6 — delete `normalize_wait_deadline`'s negative borrow.** Build exit **0**; ctest exit **8**,
2 arms red: *"a negative nanosecond field borrows a second instead of truncating toward zero"* and
*"-2.5e9 ns borrows three seconds and leaves 0.5e9"*.

## Why the test discriminates on the VIRTUAL clock, and not on a duration

Realtime and monotonic **cannot** separate the fix from the defect: the old path was self-consistent
for realtime, and monotonic only diverges across a wall-clock step, which a test cannot induce. Virtual
is process CPU time, which advances by microseconds while a thread is parked — so the two
implementations disagree about something **discrete**, `rc == 0` against `rc == 0x8002003c`, and the
arm asserts a return value rather than a wall-clock bound. `test_kernel_sem_timedwait.cpp`'s header
records at length why a bound placed between two overlapping timing distributions passes the defect
silently; there is no overlap between "signalled" and "timed out".

The one arm that *is* a duration is labelled in the source as a stub/hang guard and explicitly not a
discriminator: it passes under M1 (300 ms clears it), and raising it until it reddened would recreate
exactly the unsound wall-clock bound that file warns about.

## Risks

- **The relock poll costs more CPU than `::Sleep(1)` did** — 2 ms in the steady state against ~15.6 ms,
  so ~7.8x more wakes for a thread waiting on a long-held guest mutex, and ~31x during the first 8 ms.
  #3062 measures the same cadence at roughly 0.1–0.3% of a core per parked thread coarse / 0.5–1% fine.
  Bounded and affordable, but it is the change most worth a reviewer's attention, particularly for a
  title that parks many threads on one contended mutex.
- **The re-arm loop is Windows-only code that CI compiles and unit-runs, but no guest drives.** It is
  structurally the #3074 shape and is bounded by a `steady_clock` budget rather than by a clock whose
  resolution nobody has measured, so it cannot spin; the residual question is whether
  `ERROR_TIMEOUT` is the right discrimination against `WaitOnAddress`'s documented failure modes.
- **The virtual-clock arm's premise** is that the test process spends far less than 300 ms of *user CPU*
  in a 600 ms window. True while both its threads are blocked, and unaffected by other processes'
  load since `getrusage(RUSAGE_SELF)` / `GetProcessTimes` are per-process. The case is timed out at 60 s
  in CMake so a lost signal fails rather than hangs.
- **`k_cond_timedwait_sce` now has an early return** if the clock read fails, which the old body could
  not take. It is inside the `GuestCondWaiterScope`, so the #2168 waiter count is still balanced.

## Ruled out

`docs/PORTING.md` gains a **`### Ruled out — the Windows guest timed waits`** section, the cross-title
home for this family: the winpthreads-tick attribution above, the "so it must be an arithmetic defect"
follow-on, and the fact that a relative µs timeout is not clock-free.

## Review

Independently reviewed twice. At `d3346931` (review `5092341709`) — **REJECTED**, three blocking findings,
all now addressed in `783c46e7` and described above: the loop's dependence on `CLOCK_REALTIME`
resolution (B1), the dropped wake on an expired deadline (B2), and a wrong PR number in the new
`## Ruled out` rows (B3). Its non-blocking items are addressed too: the `max_ms` wrap, the missing
mutation coverage on arms 12/15/16, and this body's under-claim about Windows execution. The review
independently confirmed the central claim about `WaitOnAddress` against the source, and confirmed the
`GuestCondWaiterScope` balance, the error encoding, the slice loop's mutex/bookkeeping pairing and the
three arithmetic functions at their extremes.

Re-reviewed at `783c46e7` (review `5092499563`) — **APPROVED**. It checked the new budget/steady-clock
structure by construction rather than by reading the summary back: termination, no spin (a recomputed
`remaining_ms` is always ≥ 1, so only the first call can carry a zero timeout), `(DWORD)remaining_ms`
monotone non-increasing from `INFINITE - 1` so the wait is always bounded, no wrap anywhere, and no
inflation in the normal case (the census's own 1 ms → 1.870 ms means the first budget check already
breaks, so the re-arm fires only when the OS wait comes back short). It also confirmed the residual is
pre-existing by checking `origin/main`'s subtraction produces the identical outcome.

`4c1442fb` then took its remaining non-blocking items: the census sample on the clock-read error path,
the missing evidence for `steady_clock`'s resolution (now cited from `kernel_sem_timedwait` at 0.03 s
on the Windows MinGW job, which a 15.6 ms-granular steady clock could not produce), the relock's real
justification above, and this body's monotonic-condvar claim.

A third review at `4c1442fb` (review `5092640403`) carried the approval and caught the false
merge-safety sentence corrected above, plus two comment paragraphs left behind by the census scope's
move — both fixed in `a252272b`, which is comment-only.

## Also

- `prosper/src/hle/sync/AGENTS.md` — the folder had none and this PR worked in it.
- **Not addressed, deliberately:** #3038's `sleep_full` / `select` half. It is the same family and the
  same 32-bit-`long` class, but it is in `hle_kernel_time.cpp` and touches nothing here, so this does
  not subsume it. The `WaitDeadline` type added here is what #3038's "take the guest fields as int64
  rather than through a host `struct timespec`" suggestion asks for, and whoever takes #3038 should
  reuse it rather than add a third spelling.


mattias800 (commented):
## REJECTED

Three blocking findings, all cheap to fix; none of them touches the central diagnosis, which I
checked independently and which **holds**. Also one correction that runs in the PR's favour and
which I am flagging precisely because it does (a finding that raises a claim is the one that gets
checked least).

Everything below cites only lines I opened. Where I could not verify something from this box I say
so rather than asserting it.

---

## 1. The central factual claim is TRUE — verified against the source, not accepted

The PR asserts that #3056's recorded mechanism is wrong: that `k_cond_timedwait_sce` does not reach
winpthreads' `pthread_cond_timedwait` on Windows. I confirmed this three ways without relying on the
PR's `#error` positive control:

- `prosper/src/hle/sync/sync_futex.cpp:358` defines `interruptible_cond_timedwait`; `:363` opens
  `#ifdef _WIN32`, and the whole Windows body is `WaitOnAddress`. `:430` is the `#else` and `:435` is
  the only `pthread_cond_timedwait` call in the function, unreachable on Windows.
- On `origin/main`, `k_cond_timedwait_sce`'s body calls exactly that function
  (`interruptible_cond_timedwait(c, m, &dl, GuestWaitKind::ConditionSequence, 0, nullptr,
  kGuestMutexCondWaitBookkeeping)`), so the Sony spelling has no other route.
- `grep -rn 'pthread_cond_timedwait\|pthread_cond_clockwait' prosper/src/` outside `sync_futex.cpp`
  finds only `posix_shim.hpp` (Linux/Apple-only include), `hle_ult.cpp` (the ULT scheduler, a
  different family), comments, and the NID registration. No guest condition wait reaches winpthreads'
  `pthread_cond_timedwait` on Windows.

The `PORTING.md` "Ruled out" rows are the right shape and the right place for this. Correcting a
wrong mechanism in a filed issue is the most valuable thing in this PR.

## 2. The clock change is correct, and the "identical for realtime" claim checks out

- `hle_kernel.cpp:3004-3012` resolves `guest_cond_snapshot(c).clock_id` and calls
  `interruptible_cond_clock_timedwait`, which at `hle_kernel.cpp:1086-1094` forwards a realtime
  clock straight to `interruptible_cond_timedwait` with the *same* `GuestWaitKind`, the same null
  `mutex_state` and the same bookkeeping pointer. Byte-for-byte the old call. Confirmed.
- Default `GuestCondState::clock_id` is `0` (`hle_kernel.cpp:421`) and `kSonyClockRealtime == 0`
  (`:1006`), so an unregistered condvar — a guest static initialiser — also lands on the unchanged
  path. Confirmed.
- **`GuestCondWaiterScope` stays balanced across the new early return.** It is RAII
  (`hle_kernel.cpp:466-470`) and is constructed before the clock read, so the #2168 waiter count is
  decremented on the error path. Correct.
- **Error encoding on the new path is right.** `guest_cond_clock_now` returns a *host* errno by
  deliberate design, with the reason spelled out at `hle_kernel.cpp:1048-1051` (translating twice
  would read FreeBSD 60 as a host number and hand back EINVAL). The new path applies `fbsd_errno`
  exactly once and then `sce_pthread_rc`, which passes an already-encoded value through
  (`pthread_slot.hpp:225`). One translation, one encoding. Correct.
- **The slice loop preserves the mutex/bookkeeping contract.** Each iteration of
  `interruptible_cond_clock_timedwait` calls `interruptible_cond_timedwait_relative`, which releases
  bookkeeping, unlocks, waits, relocks and re-acquires bookkeeping as a matched pair
  (`sync_futex.cpp:456-486`). Across N slices the pairing is N balanced pairs and the mutex is held
  on return. Correct — and identical to what `k_cond_timedwait` has been doing.

## 3. The pure arithmetic is sound

I worked `normalize_wait_deadline`, `abs_deadline_from_rel_us` and `wait_timeout_ms_ceil`
(`precise_sleep.hpp:215`, `:233`, `:261`) by hand at the extremes rather than reading the tests'
expectations back:

- `normalize_wait_deadline` is total. `INT64_MIN / 1e9` and `carry * 1e9` are both in range, the
  borrow correction fires before the saturation checks, and the saturation checks are written so the
  comparison itself cannot overflow. Every return has `nsec` in `[0, 1e9)`.
- `abs_deadline_from_rel_us` cannot overflow: `base.nsec + (usec % 1e6) * 1000 < 2e9` fits `int64_t`;
  the `sec >= 0` guard is the only branch where `sec + whole_seconds` could overflow, and
  `whole_seconds <= UINT64_MAX/1e6 ≈ 1.84e13` makes the negative-`sec` case unreachable. Both
  hand-checked cases the tests assert (`{INT64_MAX-1,0} + ~0ull` and `{INT64_MAX,999999999} + 1`)
  reproduce.
- `wait_timeout_ms_ceil`'s unsigned-subtraction trick is correct and the comment's justification is
  the right one: `b.sec >= a.sec` is established first, and `INT64_MAX - INT64_MIN == 2^64-1` fits
  `uint64_t` exactly, which the signed form could not represent. `{INT64_MIN,0} -> {INT64_MAX,0}`
  reproduces as `max_ms`.
- **`nsec` can never leave `[0, 1e9)`**, so the casts to a 32-bit
  MinGW `long` at `hle_kernel.cpp:2935` and `:3010` are safe. That is the property the whole `WaitDeadline`-over-`timespec` argument rests
  on, and it holds.

One real (unreachable-today) overflow, listed under non-blocking below.

## 4. The tests do discriminate

- ARM 2 of `test_kernel_cond_timedwait_clock.cpp` is a genuine discriminator and the reasoning for
  choosing the virtual clock over a duration bound is right: `rc == 0` against `rc == 0x8002003c` has
  no overlap, where two latency distributions would.
- Its flake premise is sound and stronger than the PR states. The deadline is
  `cpu_now + 300 ms`, read *inside* the handler, so CPU already burned before the wait is irrelevant;
  only CPU spent by the process during the ~600 ms window counts, and both threads are parked.
  `getrusage(RUSAGE_SELF)` / `GetProcessTimes` are per-process, so other ctest workers cannot move
  it. I do not think this arm can flake.
- Signalling under the mutex is load-bearing here, not belt-and-braces as the comment suggests:
  between two slices the waiter *holds* the guest mutex, so the signaller blocks until the waiter
  re-enters the wait and releases. That is what makes the wake unmissable.
- M1/M2/M3 are each load-bearing on the arms named. Arms 12, 15 and 16 have no mutation arm — see
  non-blocking below.

---

# BLOCKING

## B1. The re-arm loop makes every Windows timed condition wait terminate on `clock_gettime(CLOCK_REALTIME)`'s **resolution** — which this PR does not state, on the one clock family this whole issue chain exists because of

`sync_futex.cpp:400-415`. The old code read `CLOCK_REALTIME` once and let the **kernel wait** decide
when to return. The new loop re-reads `CLOCK_REALTIME` on every iteration and only leaves when *that
reading* has reached the deadline (`:408`). Both the deadline (`abs_deadline_us` /
`abs_deadline_from_rel_us`) and the loop's `now` come from the same `clock_gettime(CLOCK_REALTIME)`,
so the exit condition is: **wait until the realtime clock's reading advances past
`entry_reading + requested_interval`.**

If that clock's tick is coarser than the requested interval, the loop cannot exit until the *next
tick*, however many 1 ms `WaitOnAddress` calls that takes. The inflation is additive and bounded by
one clock granularity — negligible on a 100 ms wait, and the whole wait on a sub-millisecond one.
#3056's censused call requests **0.818 ms**.

What I can and cannot establish from here:

- **Established:** the loop's termination now depends on that clock's resolution, where before it did
  not. That is a new dependency introduced by this diff, and it is not mentioned anywhere in the PR,
  the code comment, or `PORTING.md`.
- **Not established, and I am labelling it a hypothesis:** I believe mingw-w64's
  `clock_gettime(CLOCK_REALTIME)` is `GetSystemTimeAsFileTime`, i.e. system-tick granularity. **I
  could not open mingw-w64's `clock.c` from this box (no cross toolchain installed), so do not treat
  that as verified.** What makes it worth blocking on rather than filing: #3013 measured the global
  tick on the #3056 host at **15.554 ms**, and if the hypothesis holds then on that host a 0.818 ms
  `sceCondTimedwait` becomes a wait to the next 15.6 ms boundary — mean ~8 ms, worst ~15.6 ms,
  against the 1.870 ms censused today. That would be a 4-8x regression on the exact number this PR is
  filed against, delivered by a change whose stated benefit ("`ERROR_TIMEOUT` before the deadline
  after a wall-clock step") is an unobserved hypothetical. It would also re-enter the ~15.6 ms tick
  that #3013 → #3022 → #3044 spent three PRs removing, through a door none of them was watching.

Two ways to clear this, either is fine:

1. **Measure it.** One probe of the #3013 shape, outside prosper: sample
   `clock_gettime(CLOCK_REALTIME)` in a tight loop and report the smallest non-zero delta. If it is
   sub-millisecond the finding evaporates and the loop lands as written, with the figure recorded
   next to the loop so nobody re-derives it.
2. **Remove the dependency.** Compute the remaining interval once at entry, derive a
   `std::chrono::steady_clock` deadline alongside the realtime one, and exit when *either* says the
   deadline is reached. `steady_clock` is QPC-backed on Windows, so the loop then terminates at the
   requested interval whatever the wall clock's granularity is, while a forward wall-clock step still
   ends the wait. That keeps the property the PR wants and costs one extra clock read.

I would not land the loop on the argument as it stands, because the argument for it is a hypothetical
step and the argument against it is a resolution nobody has looked up.

## B2. `if (timeout_ms == 0) break;` drops a wake the old code reported — and the comment above it says otherwise

`sync_futex.cpp:408`. When the deadline has already passed the loop leaves **without calling
`WaitOnAddress` at all**. The old single-shot form called `WaitOnAddress(..., 0)` in exactly that
case, and `WaitOnAddress` returns `TRUE` immediately when the value differs from `CompareAddress`
regardless of the timeout — so a signal published after `expected` was latched but before the check
was reported as a wake. Now it returns `ETIMEDOUT`.

The window is not theoretical: `expected` is latched while the mutex is held, and between there and
the first clock read the thread does `register_wait`, the bookkeeping release, the unlock, and
`dispatch_pending_guest_exception()` — which can run an entire guest exception handler on this
thread. On a sub-millisecond timeout the deadline can easily be gone by the time the loop starts.

`ETIMEDOUT` is a legal answer once the deadline has passed, and the `WaitEventFlag`/`WaitSema`
callers re-check their predicate so they are unaffected. What makes this blocking is that the comment
at `sync_futex.cpp:390-392` states the invariant **"A wake is still reported as a wake at once,
spurious or not"**, and the code no longer does that on this path. Per the charter's rule on cited
claims, an invariant asserted in a comment in synchronisation code is exactly the thing the next
reader will rely on without re-deriving.

Fix is one line — check `slot->sequence.load(std::memory_order_acquire) != expected` before breaking,
or keep the degenerate `WaitOnAddress(..., 0)` — or correct the comment. Either resolves it.

## B3. `PORTING.md` cites the wrong PR, twice, inside a `## Ruled out` section

`prosper/docs/PORTING.md:627` — *"it does not, and the correction is #3078"* — and `:638` —
*"(#3056, fixed in #3078)"*. **#3078 is a closed, unrelated docs PR** ("docs: instrument trap 231 — a
claim that arrives already labelled as checked"). This PR is **#3235**, and `gh api
'.../pulls?state=all&head=...:fix/issue-3056-cond-timedwait'` returns 3235 and nothing else, so there
is no earlier PR on this branch that the number could have referred to.

`## Ruled out` rows are, by the charter's own description, the most expensive knowledge in the
repository, and their whole value is that the next agent can follow the link instead of re-deriving.
A row that points at an unrelated PR is worse than a row with no link, because it looks checked.
Trivial fix; it just has to happen before merge.

---

# NON-BLOCKING

**N1. The verification table under-claims, and the Windows halves *have* been executed.** The PR says
the re-arm loop and the relock cadence are "Compiled, not executed... No Windows host was available."
`.github/workflows/ci.yml:613-645` is `runs-on: windows-latest` with MSYS2 UCRT64: it configures,
builds natively and runs `ctest --no-tests=error` on a real Windows host. On this exact head
(`d3346931`) that job is **green**, `260/260`, and the job log shows
`51/260 Test #51: kernel_cond_timedwait_clock ... Passed 0.66 sec`. ARM 1 of that test is a realtime
20 ms `scePthreadCondTimedwait` that must return `ETIMEDOUT`, so it drives the new re-arm loop, and
the cond relock drives the new `interruptible_mutex_lock` cadence — on Windows, in CI. That is not
the #3056 host and it does not settle B1 (the 0.66 s is dominated by ARM 2's 600 ms sleep and cannot
resolve a 15 ms difference), but "no Windows host was available" is not what the table should say,
and CI is the obvious place to put a granularity probe.

**N2. `wait_timeout_ms_ceil` can overflow for a large `max_ms`.** `precise_sleep.hpp:277-278`: the
pre-multiply bound is `seconds > max_ms / 1000`, so `seconds == max_ms/1000` is allowed through, and
with `max_ms == ~0ull` that gives `seconds * 1000 == 18446744073709551000`; adding the rounded-up
nanosecond term (up to 1000) wraps and the function returns **384**. Unreachable from the one
production caller (`INFINITE - 1` bounds `seconds` at 4,294,967), and arm 16's comment — *"no input
can reach it"* — is true only because of that caller. Since the function is `constexpr` and public,
either clamp after the add or say in the comment that the bound is on `max_ms`, not on the input.

**N3. Arms 12, 15 and 16 have no mutation arm.** M2 pins the rounding and M3 the
`abs_deadline_from_rel_us` saturation, but nothing pins arm 15's past-deadline zero — which is the
loop's *termination condition* — or arm 16's "bound before the multiply". Moving the
`seconds > max_ms / 1000` guard to after the multiply would be a two-character mutation that should
redden arm 16; inverting the `b.sec < a.sec` early-out should redden arm 15.

**N4. `interruptible_mutex_lock`'s poll is unbounded where the semaphore loop's is not.**
`sync_futex.cpp:582-589` borrows `poll_slice_ns`'s cadence, but `k_sem_timedwait`'s loop stops at the
guest's timeout while this one has no deadline: a thread parked on a mutex held for ten seconds now
takes ~5,000 wakes where it took ~640. `poll_slice_ns(elapsed, ~0ull)` is a correct use — the clamp
is documented as an overflow guard and `steady_now + 2e6` cannot wrap — and the comment says so, so
this is a cost note rather than a defect. It is worth stating the compensating benefit the PR does
not claim: a thread inside this loop is **not** registered in `g_wait_slots`, so `interrupt_guest_wait`
cannot wake it and the poll interval *is* the cooperative guest-exception delivery latency for every
contended guest mutex. That goes from ~15.6 ms to 0.5-2 ms, which is a real correctness improvement
for GC stop requests and is a better argument for the change than the census attribution is.

**N5. The Linux impact section omits what changes for a MONOTONIC condvar.** The PR says monotonic
gets "a wall-clock-jump exposure removed", which is true, but the route is now
`interruptible_cond_clock_timedwait`'s 50 ms slice loop (`hle_kernel.cpp:1010`, `:1096-1112`): a
10 s monotonic `scePthreadCondTimedwait` now wakes and re-acquires the guest mutex ~200 times instead
of parking once. Invisible to the guest and identical to what `k_cond_timedwait` already does, so I
am not asking for a change — but "improvement, no cost" is not quite the shape of it.

**N6. The census records a sample on the new error path.** `WaitCensusScope`
(`timedwait_census.hpp:108-121`) is constructed before the clock read, so a `guest_cond_clock_now`
failure logs a ~0 ns actual against the full requested interval. Diagnostic-only and effectively
unreachable; noted for completeness.

---

# On honesty

I went looking for overclaiming and mostly found the opposite. The PR declines to claim a change to
the 2.29x, states its Windows work as unexecuted (N1: it is now more verified than that), labels the
`abs_deadline_from_rel_us` saturation arm as a consistency guard rather than a live fix, labels the
duration bound in ARM 2 as a stub guard and not a discriminator, and reports the build exit code of
each mutation arm alongside the ctest result. The one claim I found that the code does not support is
the "a wake is still reported as a wake at once" invariant in B2, and that reads as an oversight
rather than a stretch.

Re-review on the corrected head; B3 is a text edit, B2 is one line, and B1 needs either a five-line
probe or one extra clock read.

mattias800 (commented):
## APPROVED — on head `783c46e7`

Re-review of the second commit. All three blocking findings from my review of `d3346931` are
resolved, and the two non-blocking items that were about test coverage (N2, N3) are resolved as well.
I re-derived each fix rather than reading the summary back, and I checked the new structure for
problems the fix could have introduced. I found none that block.

One merge precondition, not a review finding: CI at this head was still running when I looked
(`Windows MinGW`, `Linux`, `macOS`, `Windows App`, `Linux App Package` all `in_progress`; `Docs`,
`Contribution shape`, `Linux App`, `macOS App` green). Confirm a nonzero pass count on the required
jobs immediately before merging — the green I can see is a statement about four jobs, not the run.

---

## B1 — RESOLVED, and resolved better than the option I offered

`sync_futex.cpp:419-441`. The budget is now computed once from `CLOCK_REALTIME` and spent against
`steady_now_ns()`. That is stronger than the "exit when either clock says done" hybrid I suggested,
because it removes the realtime dependency from the loop entirely rather than bounding it.

What I checked, in order of how badly each could have gone wrong:

- **Termination.** `budget_end_ns` is fixed before the loop; `steady_now_ns()` is monotone; the loop
  exits on `now_ns >= budget_end_ns`. Bounded by construction, and it no longer matters whether
  `CLOCK_REALTIME` advances at all.
- **No spin.** After the first iteration `budget_end_ns > now_ns`, so the recomputed `remaining_ms`
  is `>= 1` — it can never re-arm with a zero timeout. The only zero-timeout `WaitOnAddress` is the
  first one, which is deliberate (B2).
- **`(DWORD)remaining_ms` cannot become `INFINITE`.** The initial value is bounded by
  `wait_timeout_ms_ceil(..., INFINITE - 1)`. On re-arm, `now_ns >= steady_at_entry` (monotone), so
  `budget_end_ns - now_ns <= remaining_ms * 1e6` and the recomputed value is `<=` the previous one.
  Monotone non-increasing from `INFINITE - 1`, so `0xFFFFFFFF` is unreachable and the wait is always
  bounded. This is the bound I would most have expected to be missed; it holds.
- **No wrap.** `remaining_ms * 1000000ull <= 4.29e15`, and `steady_now_ns()` is nanoseconds since
  boot (`~1e15-1e17`), so the sum is four orders of magnitude below `UINT64_MAX`. The comment at
  `:425` states the multiply and the sum; both check out. `budget_end_ns - now_ns + 999999ull` is
  guarded by the `>=` check above, so it cannot underflow.
- **No inflation in the normal case.** The census's own numbers say the loop runs exactly once: a
  1 ms `WaitOnAddress` came back at 1.870 ms, which is already past a 1 ms budget, so the first
  budget check breaks. The re-arm fires only when the OS wait returns *short*, which is the case it
  exists for. That was the property B1 was really about, and it now holds without needing any clock's
  resolution to be known.

**The residual is correctly scoped, and I checked it is genuinely pre-existing.** The initial budget
is `realtime_deadline - realtime_now`, both readings of the same clock, so a coarse `CLOCK_REALTIME`
can put it out by one tick when a tick boundary falls between `abs_deadline_us`'s reading and this
one. On `origin/main` the identical subtraction, with `timeout = 0` on a negative result, produces
the identical outcome. Unchanged, and now stated at the call site.

**The one thing the fix now depends on is `steady_clock`'s resolution — and unlike `CLOCK_REALTIME`,
that one has evidence in this tree.** I am spelling this out because it is the substitution the fix
makes and the PR asserts it rather than showing it: if mingw's `steady_clock` were 15.6 ms-coarse,
the loop would re-arm until it ticked and B1 would simply have moved. It is not, and the evidence is
external to the process that would be measuring itself:
`tests/hle/kernel/test_kernel_sem_timedwait.cpp:128` runs a 5 ms guest semaphore timedwait, which on
Windows goes through #3044's 500 µs poll slices served by `sleep_until_steady_ns` — a loop whose
termination condition is `steady_clock`. The `Windows MinGW` job's log for this branch's first head
reports `50/260 Test #50: kernel_sem_timedwait ... Passed 0.03 sec`, and that duration is ctest's,
measured from outside the process. A 15.6 ms-granularity `steady_clock` would cost that arm alone
`>= ~0.15 s`. So `steady_clock` resolves well below a millisecond there. Worth putting in the code
comment beside the `steady_clock is QueryPerformanceCounter` claim, which is currently asserted from
knowledge of libstdc++ rather than from anything in this repository.

## B2 — RESOLVED

`sync_futex.cpp:428-433`. The first `WaitOnAddress` always executes, with a zero timeout when the
budget is already spent, so its value check is restored: a sequence that already differs returns
`TRUE` immediately and is reported as a wake. The expired path is then `WaitOnAddress(...,0)` →
`ERROR_TIMEOUT` → `now_ns >= budget_end_ns` (the reading has advanced past a `+0` budget) → break →
`ETIMEDOUT`. That is exactly `origin/main`'s behaviour for an expired deadline, and the comment now
describes what the code does.

## B3 — RESOLVED

`PORTING.md:627` and `:638` cite #3235. `grep -n 3078 prosper/docs/PORTING.md` is empty.

## N2 (the `max_ms` wrap) — RESOLVED, and the arm is load-bearing

`precise_sleep.hpp:278-286`. Bounding the sum instead of clamping after it is the right shape:
`whole_ms <= max_ms` follows from the pre-multiply bound, so `max_ms - whole_ms` cannot underflow,
and `frac_ms <= max_ms - whole_ms` is exactly the condition under which `whole_ms + frac_ms` cannot
wrap. Correct.

I worked the new arm's input by hand rather than trusting it: `at_bound = UINT64_MAX / 1000` passes
the seconds bound by equality, `whole_ms = 18446744073709551000` leaves 615 of headroom, and
`999999999 ns` rounds to `frac_ms = 1000`, which overshoots it. Under the old expression the sum
wraps to **384** and `384 > max_ms` is false, so it returns 384 — both `CHECK`s at
`test_precise_sleep.cpp:564-567` redden. Load-bearing.

**The self-correction on the first version of this arm is the most valuable thing in the second
commit**, and I want it noted rather than passed over: a 1 ns remainder gives `frac_ms == 1`, which
fits the 615 of headroom, so that arm could not have failed under its own mutation and would have
been reported as covering a case it could not reach. That is precisely the trap CLAUDE.md's
positive-control rule names — a green arm whose *input* cannot express the class — caught by the
author, on their own arm, and recorded in the source at `:551-560` where the next reader will find
it. Exactly right.

## N3 (mutation coverage) — RESOLVED

M5 and M6 are both load-bearing, checked by construction rather than by report:

- **M5** (drop `if (seconds == 0) return 0;`): for `{100,1} -> {100,0}`, `seconds` becomes `0 - 1`
  = `UINT64_MAX`, which then trips the seconds bound and returns `max_ms`. Arm 15's "a deadline one
  nanosecond in the past has been reached" expects `0`. Reddens.
- **M6** (drop `normalize_wait_deadline`'s negative borrow): `{5,-1}` returns `{5,-1}` instead of
  `{4,999999999}`. Arm 12's borrow check reddens — and so does the `[0,1e9)` invariant the whole
  32-bit-`long` argument rests on.

## N1 (the verification table) — RESOLVED, and I checked it does not now overclaim

The table says "Compiled and unit-exercised on Windows, but never driven by a guest", names the job
and its runner, quotes `260/260` and attributes it to *this branch's first head*, and adds a row
saying the tick's effect on the wait is measured nowhere. That is the correct set of distinctions,
including the one it would have been easy to blur — a passing unit test on a Windows runner is not a
title under contention. The remaining quantitative claims (`337/337`, the cross-compile exit codes,
the six mutation arms) are the author's; I did not rebuild, per the review contract.

---

## Still open, all non-blocking, all unchanged from my first pass

- **N4.** `interruptible_mutex_lock`'s poll is unbounded where the semaphore loop it borrows the
  cadence from stops at the guest's timeout. Cost note, not a defect; and the compensating benefit is
  real — a thread in that loop is not registered in `g_wait_slots`, so the poll interval *is* the
  cooperative guest-exception delivery latency for every contended guest mutex, which improves from
  ~15.6 ms to 0.5-2 ms. That is a better argument for the change than the census attribution is, and
  it is still not in the PR body.
- **N5.** The Linux-impact section still reads "improvement" for a monotonic condvar without noting
  the route: `interruptible_cond_clock_timedwait`'s 50 ms slice loop, so a 10 s monotonic
  `scePthreadCondTimedwait` now wakes and re-acquires the guest mutex ~200 times instead of parking
  once. Invisible to the guest, identical to what `k_cond_timedwait` already does, so I am not asking
  for a change — only that "improvement, no cost" is not quite the shape of it.
- **N6.** `WaitCensusScope` is constructed before the clock read, so a `guest_cond_clock_now` failure
  logs a ~0 ns actual against the full requested interval. Diagnostic-only, effectively unreachable.
- **New, and the smallest of these.** `sync_futex.cpp:434-436` still reports `ETIMEDOUT` for a
  `WaitOnAddress` failure that is not a timeout. The comment now says so and says why it was left
  alone — that it matches the single-shot form and changing it would be an unverifiable guest-visible
  contract change. I agree with the decision; I note it only so it is not mistaken for an oversight
  by a later reader.

---

## Carried forward from the first review, unchanged by this commit

`hle_kernel.cpp` and `tests/hle/kernel/` are untouched by `783c46e7` (`git diff d3346931 783c46e7`
over those paths is empty), so these findings stand as verified: the central factual claim about
`WaitOnAddress` vs winpthreads; the clock resolution and the byte-identical realtime path; the
`GuestCondWaiterScope` balance across the new early return; the single `fbsd_errno` translation
matching the host-errno contract at `hle_kernel.cpp:1048-1051`; the slice loop's mutex/bookkeeping
pairing; the three arithmetic functions hand-checked at the extremes; and ARM 2's virtual-clock
discriminator, whose flake premise I still believe cannot trigger.

**APPROVED.** Merge once the required CI jobs report green at `783c46e7` with a nonzero pass count.

mattias800 (commented):
## APPROVED — carries to head `4c1442fb`

Short delta, re-derived rather than read back. The code is sound and the approval carries. **One
must-fix before merge, and it is in the PR body rather than the code** — it is the very sentence you
asked me to re-check, so I am flagging it prominently rather than filing it as a nit.

---

# MUST FIX BEFORE MERGE — the merge-safety claim is false, though its conclusion is true

PR body, the paragraph explaining `337 → 338`:

> That merge touches none of the files this PR changes — `git diff --name-only 783c46e7 6ea482ac`
> over them is empty

That intersection is **not** empty. `prosper/CMakeLists.txt` is in both sets:

```
$ comm -12 <(git diff --name-only 783c46e7 6ea482ac | sort) \
           <(git diff --name-only origin/main...HEAD  | sort)
prosper/CMakeLists.txt
```

#3234/#3238 registered `wt_stash_tool` in `CMakeLists.txt` (now at `:219`), and this PR adds
`kernel_cond_timedwait_clock` to the same file. So the one file where a merge could have eaten
somebody's work is precisely the file the check claims was untouched.

**Your work did survive, and so did main's — I verified both directly**, which is why this is a text
fix and not a re-review of the merge:

```
$ git diff origin/main...HEAD --numstat
14   0   prosper/CMakeLists.txt
27   0   prosper/docs/PORTING.md
55  10   prosper/src/hle/kernel/hle_kernel.cpp
30   0   prosper/src/hle/sync/AGENTS.md
102 10   prosper/src/hle/sync/sync_futex.cpp
100  0   prosper/src/host/platform/precise_sleep.hpp
189  0   prosper/tests/hle/kernel/test_kernel_cond_timedwait_clock.cpp
208  0   prosper/tests/host/platform/test_precise_sleep.cpp
```

`merge-base(origin/main, HEAD) == origin/main == 56612f2d`, so the branch is fully up to date and
that diff is the whole PR against the live target. Deletions are **10 and 10, confined to the two
bodies this PR rewrites**; every other file is a pure addition, `CMakeLists.txt` included — and
`CMakeLists.txt`'s 14 added lines are exactly this PR's block, with `wt_stash_tool` still present
below it. Nothing from either side was lost. That is the trap-41 check, and it passes.

Please substitute that evidence for the sentence. The reason I am not waving it through as a typo is
the shape CLAUDE.md names: a true conclusion reached by a route that does not establish it. The
conclusion is right, so nobody re-runs the derivation — and the next person to copy this exact
`git diff --name-only A B` idiom as "the merge-safety check" will get a false negative on the one
file that mattered. `git diff <base>...HEAD --numstat`, read for the deletion column, is the check
that actually answers the question.

---

# The four items, verified

**N6 — fixed, and it was the right call.** `hle_kernel.cpp:3009-3016`: the census scope now opens
after the clock read, so the failure path records nothing instead of a full requested interval
against ~0 actual. Scoping is still correct — `census` is constructed inside `GuestCondWaiterScope`
and destroyed before it, and it still brackets the whole
`interruptible_cond_clock_timedwait` call. The excluded work is one map lookup under a `std::mutex`
plus one `clock_gettime`/`GetProcessTimes`, against a millisecond-scale request; if anything the
census is now measuring the wait rather than the wait plus its setup, which is what it should have
been measuring. The EINVAL and EPERM paths were already outside the scope, so the new early return is
consistent with them rather than a special case.

**The steady_clock citation — you re-derived it correctly, and I checked the arithmetic again rather
than accepting my own earlier statement back.** `test_kernel_sem_timedwait.cpp:128` is the 5 ms
guest wait; on Windows it runs #3044's 500 µs `sleep_until_steady_ns` slices, so ~10 slices; the
`Windows MinGW` log for `d3346931` shows `50/260 Test #50: kernel_sem_timedwait ... Passed 0.03 sec`,
measured by ctest from outside the process. The comment's framing — a measured dependency
substituted for an assumed one — is exactly the distinction that makes the B1 fix worth having, and
it now lives where the next reader of the loop will find it.

**N4 — claim verified independently.** `awk '/^int interruptible_mutex_lock/,/^}/' | grep register_wait`
returns only the new comment's own text: there is no registration in that loop, so a thread parked
there is genuinely invisible to `interrupt_guest_wait` and the poll interval is its stop-request
latency. Promoting that to the leading justification is right — it is a correctness argument, where
the census attribution was an accounting one.

**N5 — fixed.** The Linux-impact bullet now names the `min(50 ms, remaining)` slice loop and the
up-to-20-wakes-a-second cost, and says it is the trade the POSIX sibling already makes. Accurate.

---

# Non-blocking, and new in this delta

**The two census paragraphs are now orphaned.** `hle_kernel.cpp:2979-2987` — "Censused separately
from `k_cond_timedwait`…" and "Saturating, matching `guest_ns_from`…" — still sit where the scope
used to be, so they now introduce the `#3056` clock block instead of the statement they explain,
which moved twenty lines down and grew its own comment. Both paragraphs are about
`WaitCensusScope`/`timeout_ns_from_us`; move them down with it. Thirty seconds, and it keeps the file
honest in exactly the way the rest of it is.

---

# Merge preconditions

- Correct the merge-safety sentence above.
- CI at `4c1442fb` was still running when I looked: `Docs`, `macOS App`, `Contribution shape` green;
  `Windows MinGW`, `Linux`, `macOS`, `Windows App`, `Linux App`, `Linux App Package` all
  `in_progress`. Confirm a nonzero pass count on the required jobs immediately before merging — and
  note `mergeable_state` currently reads `unstable` for that reason.

Everything carried forward from my two earlier reviews stands; `git diff 6ea482ac 4c1442fb` is 27
added lines across two files, of which 25 are comments and 2 are the moved census scope, so nothing
in the verified surface changed.

**APPROVED.**

PR #3236, opened 2026-09-02
recompiler: a CFG region whose entry block defines VCC before reading it has a dead entry value (#3231)

### LINKED ISSUE #3231: Astro Bot (PPSA21564): the world map's full-screen lighting consumer 0x500571000 is declined by the CFG dispatcher's missing-entry-vcc gate

mattias800:
Fixes #3231. Refs #1809, #3218.

## The gate

`emit_cfg_state_machine` stores one value into `vcc_var` before its dispatcher loop. When the
caller's `RegState` carries no VCC value — because the guest recycled the physical `s[106:107]` pair
as ordinary scalar data — there is nothing honest to store, so the dispatcher refused the whole
region:

```cpp
if (!initial.vcc && !proven_wave32_masks)
    return reject_cfg(ins.front().pc, "missing-entry-vcc");
```

That gate is right to exist. Persisting `false` hands a fabricated mask to whatever reads VCC next,
which is **silent-wrong rather than a crash** — so an over-broad relaxation is the dangerous
direction, and this PR states below exactly what it does not admit.

## The decode, verified rather than inherited

#3231's body carries a decode. I re-derived it from the program's own dumped bytes
(`PROSPER_SHADER_DUMP` → `exec_cs_500571000.bin`, 3,596 dwords / 2,374 instructions) with
`tools/shader_inspect`, and it holds — with two details the issue omits:

| pc | decoded | |
| --- | --- | --- |
| 220, **313** | `s_barrier` (SOPP 0x0a) | so the region declined at pc 314 is the **final barrier phase**, not a sub-range of a flat program |
| 314, 317 | `s_load_dwordx2` / `s_buffer_load_dword` | no VCC |
| 316, 319 | `s_waitcnt` | |
| **320** | `VOPC op=0xc3 dst=sgpr:10` | writes **s10**, an ordinary SGPR mask — *not* VCC (the issue does not mention this instruction) |
| **322** | `VOPC op=0xc3 dst=special:106` | **the first instruction in the region to touch VCC at all** |
| 323 | `s_<logical>_b64 vcc, vcc, s[10:11]` | first read |
| **324** | `s_mov_b64 exec, vcc` | a second read (also not in the issue) |
| 325 | `s_cbranch_execz -> 3271` | the entry block's terminator |

So on every path out of the entry block VCC has been written before anything reads it, and the value
the dispatcher stored cannot be observed. The issue's conclusion stands.

## The rule

> If the CFG region's **entry block** — the instructions from `starts[0]` up to `starts[1]`, the same
> partition the dispatcher itself uses — **defines the complete VCC pair** before any instruction in
> that block can read VCC, the entry value is dead on every path and storing `false` invents nothing.

Soundness rests on three facts about the dispatcher, all in the code: `pc_var` is seeded 0 so block 0
runs first; `selector = active ? pc : UINT32_MAX` with `active_var` seeded true, so *every*
invocation executes it, in full, exactly once at the start; and block 0's `save_state` publishes the
real definition into `vcc_var` before the iteration's common phases run, so the two direct
`vcc_var` readers (portable readlane into 106, and the vote-to-VCC merge — both `bsel(…, old)`) see
the definition rather than the placeholder. Re-entering block 0 later through a back edge reads the
previous iteration's value, not the entry value.

### What it deliberately does NOT admit

- **Any define but `v_cmp_*` into VCC.** VOPC, `v_cmpx_*` excluded, destination not an SGPR ≤ 105.
  It is the one encoding that defines both words for every lane in a single instruction and it is
  the form the live evidence uses. **A VOP3B carry-out into VCC, a 64-bit scalar write of the pair
  (`s_mov_b64 vcc, …`), and `v_cmpx_*`'s EXEC write all stay rejected** — they are plainly full-width
  too, and they are left out because no program here needs them.
- **A b32 write of `vcc_lo` or `vcc_hi` alone.** Half the pair would still carry the entry value, so
  the scan stops rather than continuing to a later full define.
- **A define in a successor block.** Only the entry block counts; the taken edge must not be able to
  leave without passing the define.
- **Any read first, in any form.** Over-approximated on purpose: the implicit consumers this file
  already enumerates (`s_cbranch_vccz/vccnz`, VOP2 `0x01` and `0x28..0x2a`) plus **any** operand that
  can name a word of the pair — `Special` 106/107, or an `SGPR` root ≥ 99, since the widest scalar
  operand is an eight-word T# and s99 can still reach s106. All four source slots are scanned, not
  `n_src`, because over-reading only ever moves the answer to "not dead".
- **Any stage but Wave64 compute**, and **any partial-workgroup extent** (`initial_active != 0`,
  whose padded invocations never dispatch block 0 and so would keep the placeholder). Fragment Wave64
  has the same backstop described below but no live case; every other stage has neither.

### The second, independent backstop — measured, not asserted

For Wave64 compute/fragment, `load_state` loads `vcc_var` into `state.vcc` only for a block whose
`wave64_b64_mask_in` MUST set contains 106, and that set seeds 106 only `if (initial.vcc)`. So an
absent entry VCC already reaches a consumer as *no value*. Mutating the implicit-read check out of
the new helper shows it directly: the negative fixture is **still refused**, by
`cfg-recompile-reject mode=unresolved-operand pc=11 words=02060300` — the downstream fail-visible
path — instead of by the gate. The entry-block rule is belt over braces here, not the only guard.

## Verification

Build and ctest in the `ps5ys` distrobox, `prosper/build-linux`, `-DPROSPER_APP=ON`.

| | build rc | ctest | ctest rc |
| --- | --- | --- | --- |
| `origin/main` baseline (`21c771ae`), measured before the rebase | 0 | **100% tests passed out of 337** | 0 |
| this branch on that base | 0 | **100% tests passed out of 338** | 0 |
| **this branch rebased onto `28d4db1c`** (`#3237` landed meanwhile) | **0** | **100% tests passed out of 339** | **0** |

`origin/main` at `28d4db1c` registers 349 `add_test(` calls and this branch registers 350, so the
+1 at each base is this PR's single new case and nothing else.

```sh
cmake -S prosper -B prosper/build-linux -DPROSPER_APP=ON && cmake --build prosper/build-linux -j6
cd prosper/build-linux && ctest --no-tests=error
ctest -N | grep entry_vcc_dead      # Test #214: entry_vcc_dead
```

The +1 is the new `entry_vcc_dead` case, verified by name. The rebase's only conflict was in
`docs/RECOMPILER_REMAINING.md`, where #3237 and this PR each appended a `## Ruled out` row at the
same place; both rows are kept.

### The regression: `tests/gpu/recompiler/test_entry_vcc_dead.cpp`

Ten independent assertions over five programs, reporting per-arm so a mutation run can tell which
moved. **Getting the fixture to reach the gate was itself a finding** — see the Ruled out row below:
in Wave64 a scalar overwrite of the pair does *not* clear `rs.vcc`, so each program is barrier-phased
with a proven terminal guard and **two** barriers, exactly like the subject. Phase A retires the pair
to scalar data and exits through the dispatcher, which is what leaves phase B's region with no entry
VCC. Every tail takes `portable_compute_cfg_readlane`, the one dispatcher entry with no fallback, and
each compiling arm is asserted to contain an `OpSwitch` so a verdict cannot have come from the
structurizer.

| arm | expectation |
| --- | --- |
| control A — one dword changed so the pair is never retired | compiles, through the dispatcher |
| **the arm** — entry block defines VCC before reading it | compiles, through the dispatcher |
| **negative** — entry block reads VCC (`v_cndmask_b32_e32`) before defining it | refused, reason contains `missing-entry-vcc` |
| negative — b32 write of `vcc_lo` only | refused, `missing-entry-vcc` |
| negative — define in the successor block | refused, `missing-entry-vcc` |

### Mutation arms (each built cleanly, `rc=0`, before being believed)

| mutation | signature |
| --- | --- |
| revert the relaxation (`\|\| true` on the old gate) | the 2 arm lines redden with `compute-cfg-reject pc=11 reason=missing-entry-vcc`; all 8 others stay green — so the fixture really does reach the gate, and the relaxation is the only thing that moves the arm |
| treat **any** write to a VCC word as the define | the 2 half-pair lines redden; the rest green |
| stop treating the implicit VOP2 consumers as reads | the reason assertion reddens **while `negative.empty()` stays green** — the program is still refused, but by `unresolved-operand` at pc 11 rather than by this gate. This is why the reason is asserted and not just the emptiness |
| scan the whole region instead of the entry block | the 2 successor-define lines redden; the rest green |

### Live title run

`tools/screenshot`, native 3840x2160, no pad script, `PROSPER_GUEST_ARGS=-force-gfx-direct
PROSPER_RENDER=1 PROSPER_VULKAN_LIB=libvulkan.so.1`, `--seconds 30 --count 5 --timeout 190`.

- `0x500571000` **recompiles and executes**: with `PROSPER_COMPUTELOG_CODE=0x500571000`, **544
  dispatches, 544 `result=ok`, 0 failures**, `spirv=110603/f56e4deac88420c2`,
  `threads=16x16x1 local=16x16x1 groups=1x1x1 buffers=53 images=33`. Against the recorded control on
  main, `executed=0 skipped=830`.
- The `[compute] skip unsupported program 0x500571000 … reason=missing-entry-vcc` line is **gone**:
  0 occurrences of the string `500571000` in a full census run, and the program no longer appears in
  `[compute-census]`, which lists only programs that skipped at least once. Three programs still skip
  (`0x50052aa00`, `0x500699600`, `0x555f41f00`), none of them this one.
- **0 `GRAPHICS submission failed`** in both runs; the level loads (`worldmap`); 5/5 pixel-distinct
  samples; 4.3 fps while producing frames.

**Nothing visible changes on this title**, and that is stated rather than glossed: the world map is
still the pre-existing nebula backdrop (#1459) — mean luma 84.1 with 17.6% of pixels ≥ 200 and
554,617 distinct colours at the 150 s sample, in the same family as the previous lane's control
measurement of the backdrop (96.1 / 17.7%). No new screenshot and no `BLOG.md` entry, because there
is no new picture to show. #3218's blow-out state did not reproduce here either, consistent with that
issue's own comment thread.

## Risks

- **The relaxation is in the silent-wrong direction if the read set is incomplete.** An implicit VCC
  consumer that is neither in this file's enumerated set nor an explicit 106/107 operand — RDNA2's
  `v_div_fmas_f32/f64` are the family that fits — would be scanned past. Two things bound it: the
  emitter models VCC reads through `state.vcc`, which for Wave64 compute with an absent entry VCC is
  *absent* rather than `false` (the mutation above measures this), so an unenumerated reader rejects
  instead of reading a fabrication; and an instruction the emitter does not model at all is rejected
  regardless. `CONFIDENCE: MED` on the completeness of the read set, `HIGH` on the dominance argument.
- **Previously-declined dispatches now run.** Any Wave64 compute program in this shape moves from
  "skipped" to "executed", so a title could in principle change. Focused snapshot pass:
  `messenger-scene`, `alexkidd-gameplay` and `dead-cells-gameplay` are **green**; `gris-gameplay`
  fails, and an A/B with `origin/main`'s `rdna2_emit_cfg.cpp` restored into the same build directory
  fails identically (GRIS's evidence window lands on the Devolver splash in both arms, and its boot
  log has 0 `skip unsupported program` lines of any reason). Pre-existing, filed as #3239. Full
  numbers in a comment on this PR.
- No behaviour change for Wave32, fragment, vertex, or any partial-workgroup dispatcher region — the
  condition short-circuits on `b.is_compute && b.wave_size == 64 && !initial_active` before the scan.

## `## Ruled out` rows landed in this PR

- `docs/RECOMPILER_REMAINING.md` — **"In Wave64, a scalar overwrite of the physical VCC pair is what
  leaves `RegState::vcc` absent."** False. `record_scalar_write`'s `rs.vcc = 0` sits inside a loop
  guarded by `rs.sreg_bool_b32.contains(reg)`, and `sreg_bool_b32` is the **Wave32** one-word-mask
  domain — empty in Wave64. Wave64 loses the value on the way *out of a dispatcher region*, where
  `load_state()` republishes VCC only if the terminal block's `wave64_b64_mask_in` still contains
  s106. A synthetic reproduction therefore needs a preceding dispatcher region, not a scalar write.
- `docs/ASTROBOT_LINUX_HANDOFF_2026_07_19.md` — **"`0x500571000`'s `missing-entry-vcc` refusal is
  protecting a live entry VCC value."** False, with the decode above.

---


mattias800:
## Focused snapshot pass

`python3 tools/snapshot/snapshot.py check messenger-scene alexkidd-gameplay gris-gameplay
dead-cells-gameplay`, Linux, this branch's build.

| guard | result |
| --- | --- |
| `messenger-scene` | **OK** — frames=29, qualifying=29, richest=32 colors, pixel_changes=28, structural_matches=29, nonblack_matches=29 |
| `alexkidd-gameplay` | **OK** — frames=50, qualifying=50, richest=35,338 colors, pixel_changes=49, structural_matches=50, nonblack_matches=50 |
| `dead-cells-gameplay` | **OK** — frames=44, qualifying=44, richest=21,130 colors, pixel_changes=43, structural_matches=44, nonblack_matches=44 |
| `gris-gameplay` | FAIL — `qualifying frames 58 < 60; structural matches 0 < 60` |

### The GRIS failure is pre-existing, and it is A/B'd rather than argued

I restored `origin/main`'s `rdna2_emit_cfg.cpp` into the same build directory, rebuilt, and re-ran
the same guard alone:

| arm | result |
| --- | --- |
| this branch | FAIL — `qualifying frames 58 < 60; structural matches 0 < 60` |
| **base (`rdna2_emit_cfg.cpp` reverted, rebuilt)** | **FAIL — `qualifying frames 59 < 60; structural matches 0 < 60`** |

Same failure, and the retained evidence is the same picture in both arms: GRIS still sitting on its
**Devolver publisher splash** at 155.4 s, `distinct_rgb_colors` ~1,000, `nonblack_ratio` 1.0,
`best_structural_similarity` 0.04-0.66 against a 0.85 floor. The guard's evidence window is landing
before gameplay, in both arms.

Independently: GRIS's boot log contains **0** `[compute] skip unsupported program` lines of any
reason, so the gate this PR touches is never reached on that title at all.

Filed as **#3239** so it does not get re-derived. The branch's source was restored and rebuilt after
the A/B (`git diff HEAD` empty, `test_entry_vcc_dead` PASS, rebuild rc=0).

mattias800 (commented):
**APPROVED.**

## Re-deriving the decode rather than inheriting it was the right instinct, and it paid

The issue carried a decode. Re-deriving it from the program's own dumped bytes found **two
instructions the issue omits**, and one of them matters: pc 320 is a VOPC writing **s10**, an
ordinary SGPR mask — not VCC. A lane that trusted the issue would have been reasoning about a
program with one VCC-touching instruction where there are two similar-looking ones, and the whole
argument turns on *which* instruction first touches the pair. That the conclusion survived is the
good outcome; that it was checked is the point.

Also worth having: pc 313's `s_barrier` establishes that the declined region is the **final barrier
phase**, not a sub-range of a flat program — which is what makes "the entry block runs exactly once
before anything reads VCC" a statement about a real control-flow region.

## The rule is narrow in the direction that matters

This gate exists because persisting `false` hands a fabricated mask to the next reader — **silent
wrong output, not a crash** — so over-broad is the dangerous direction, and the PR is disciplined
about it. Only `v_cmp_*` VOPC defining the complete pair, in the **entry block only**, before any
read. `v_cmpx_*`, VOP3B carry-out, `s_mov_b64 vcc`, and half-pair `vcc_lo`/`vcc_hi` writes all stay
rejected — and the honest reason is given: they are plainly full-width too, and are excluded because
no program here needs them. Leaving provable cases out because the evidence does not require them is
the correct instinct for a gate like this.

The read detection being **over-approximated on purpose** — all four source slots scanned rather than
`n_src`, any `SGPR` root ≥ 99 treated as reaching the pair — is the right asymmetry, since
over-reading can only move the answer to "not dead".

Tying soundness to three specific, citable dispatcher facts (`pc_var` seeded 0, `active_var` seeded
true so block 0 runs in full exactly once, `save_state` publishing before the common phases) makes
this checkable by the next reader rather than a claim about intent.

## The second backstop is the best evidence in the PR

Mutating the implicit-read check *out of the new helper* and showing the negative fixture is **still
refused** — by `cfg-recompile-reject mode=unresolved-operand pc=11`, a downstream fail-visible path,
rather than by the gate — establishes that the entry-block rule is belt over braces rather than the
only thing standing between a wrong mask and the GPU.

That is a much stronger safety argument than "the tests pass", because it says what happens **if the
new rule is wrong**. For a relaxation whose failure mode is silent-wrong, that is the question worth
answering, and it was answered by measurement.

## Verification

Baselines measured on this machine, not quoted, and re-measured across a rebase when #3237 landed
mid-flight: `origin/main` 337/337 → branch 338/338 → rebased onto `28d4db1c` **339/339**, build rc 0
throughout. CI 9 success / 0 failures.

Excluding every stage but Wave64 compute, and any partial-workgroup extent (whose padded invocations
never dispatch block 0 and would keep the placeholder), closes the two holes I would have gone
looking for.

mattias800:
## Exact-head verification — `700e9126`

Note on the head: after I rebased onto `28d4db1c` and force-pushed `c83db995`, a
`Merge remote-tracking branch 'origin/main'` commit was made in this worktree and pushed by
something other than this session, bringing in #3238. It is benign — `git diff c83db995 HEAD`
touches only `tools/wt_stash.py` and `tools/test_wt_stash.py`, and `git diff origin/main...HEAD`
is exactly this PR's five files (366 insertions, 1 deletion) — but it is recorded rather than
glossed, because a foreign commit landing in a branch under verification is the shape trap 41
describes.

Verified at that exact head, in the `ps5ys` distrobox:

```sh
cmake --build prosper/build-linux -j6                 # BUILD_RC=0
cd prosper/build-linux && ctest --no-tests=error      # 100% tests passed out of 339, CTEST_RC=0
ctest -N | grep entry_vcc_dead                        # Test #214: entry_vcc_dead
```

`origin/main` registers 349 `add_test(` calls and this branch registers 350, so the +1 is this PR's
single new case.

CI on this head: **9 pass, 2 skipping, 0 fail** (`Contribution shape`, `Docs`, `Linux`, `Linux App`,
`Linux App Package`, `Windows App`, `Windows MinGW`, `macOS (x86_64 / Rosetta 2)`,
`macOS App (SDL3 + MoltenVK)`; `Publish Release` and `Progress tracker` skipping). `mergeable: true`,
`mergeable_state: clean`.

The snapshot pass and the live Astro Bot runs above were taken on `c83db995`'s content, which is
byte-identical for every file this PR touches.

PR #3237, opened 2026-09-02
recompiler: IMAGE_LOAD_MIP reads its NSA address encoding, and tail-packed mip chains materialize

### LINKED ISSUE #3134: recompiler: MIMG op=0x1 with an extra address dword is rejected (recompile-reject-mimg-address extra=1)

mattias800:
> **`Refs #3134`, deliberately, not `Fixes`.** The half of #3134 named in its own title — *"MIMG
> op=0x1 with an extra address dword is rejected"* — is closed: the extra dword is now read, and no
> address encoding rejects an `IMAGE_LOAD_MIP` any more. But the issue was **reopened** with an
> explicit bar (*"the stage still correctly rejects"*), and Stray's fragment still does, for a
> reason that is now measured rather than inferred: it is a **graphics** stage, and the graphics
> texture path uploads one level and cannot represent `R16G16_UINT` at all. Closing that is #2818's
> shared guest-authored upload. Claiming `Fixes` here would put a false statement in the project
> record; the reviewer can close #3134 on merge if the titled defect is the bar they want.

## What the extra address dword is

`fmt=14 op=0x1 len=3` is **not an exotic packet**. It is RDNA 2's **NSA (Non-Sequential Address)
encoding** of an ordinary `IMAGE_LOAD_MIP`. The MIMG `NSA` field (dword0[2:1]) says only how many
*extra dwords* carry addresses; `addr0` still comes from `VADDR`, and every later address names an
**arbitrary VGPR, one per byte** of `words[2..]`, low byte first.

Assembling both encodings by hand through `llvm-mc -arch=amdgcn -mcpu=gfx1030` shows the address
**vector** is identical either way:

| assembly | encoding | address vector |
| --- | --- | --- |
| `image_load_mip v[5:7], [v0, v42, v5], s[32:39] dmask:0x7 dim:SQ_RSRC_IMG_2D` | `f004070a,00080500,0000052a` | `[x=v0, y=v42, mip=v5]` |
| `image_load_mip v[5:7], v[0:2], s[32:39] dmask:0x7 dim:SQ_RSRC_IMG_2D` | `f0040708,00080500` | `[x=v0, y=v1, mip=v2]` |
| `image_load_mip v[5:8], [v10, v11, v12, v13], s[32:39] dmask:0xf dim:2D_ARRAY` | `f0040f2a,0008050a,000d0c0b` | `[x, y, slice, mip=v13]` |

The first row is **byte-identical** to the words in #3134's reject line, so the captured packet's
vector is `[x, y, mip]` — the same thing the consecutive form carries.

## The emitter's expected-address-count rule was itself the bug

`rdna2_mimg_dynamic_mip_shape` derived the mip register as `VADDR + n`. That is true only for the
consecutive encoding, so on an NSA packet the function had no choice but to decline — naming the
wrong VGPR would have been worse. A test even **asserted** the decline
(`test_rdna2_decode.cpp`: *"the dynamic-mip shape rejects the NSA address form"*).

Meanwhile the emitter's own coordinate gather already reads NSA generically (the `cvg` lambda in
`rdna2_emit_alu.cpp` has decoded `addr(k) = byte (k-1) of words[2 + (k-1)/4]` since the Asterix MSAA
work). So the shape predicate was the only thing between the NSA form and the #3048 dynamic-LOD
lowering.

## How it is now consumed

1. **`rdna2_mimg_dynamic_mip_shape` reads the encoding instead of doing address arithmetic.** The
   mip is always the *last* address of the dimension's vector (2D = 3 addresses, 2D_ARRAY = 4), so
   it is `words[2]` byte 1 or byte 2 respectively. Every shape this function models fits in one
   extra dword, so `NSA == 1 && len_dwords == 3` is required and a larger NSA stays fail-closed.
   Unused high bytes of the extra dword must be zero: a nonzero one means the packet carries
   addresses this dimension does not account for, i.e. it is not the modelled shape.
2. **Tail-packed mip chains materialize.** `shader_resource_mip_chain_plan` refused *any* resource
   whose **selected** level is packed in the shared mip tail. That refused the whole small-texture
   class rather than a corner: a 32×32 four-byte pyramid is smaller than one 64 KiB macroblock, so
   **every** level of it lives in the tail and `gpu_addr` is left at the allocation base. It is now
   admitted after **proving** the plan's own in-block coordinates (`tail_x/tail_y/tail_block_bytes`)
   equal the ones `agc_shader_layout` published for the selected level; a disagreement still reports
   one level. `upload_guest_mip_chain_levels` already detiled tail levels, so no upload change was
   needed.
3. **The emitter's separate `!res->in_mip_tail` term is dropped**, for the same reason: the
   materialized level count is the authority on the placement, and repeating the test there
   re-refused what the backend now uploads.
4. **`[mimg-mip]` names the dynamic route's own inputs** — `dyn_shape`, `nsa`, `materialized_mips`,
   `compute` — so "the address encoding is not modelled" can no longer be confused with "the chain is
   not materialized on this stage".

## What remains unsupported, and why

**Stray's fragment `0x30be800000` still rejects**, and the line now says exactly why:

```
[mimg-mip] ... shape=1 proven_zero_mip=0 img_dim=1/1 mips=6 mip_tail=1 ...
           dataformat=7 ncomp=2 dmask=0x7 dyn_shape=1 nsa=1 materialized_mips=1 compute=0
```

`dyn_shape=1` — the address encoding is now read. `compute=0` and `materialized_mips=1` are the two
remaining facts, and both are graphics-path gaps, not recompiler ones:

* The **graphics** texture path uploads exactly one level. Its chain is *generated* by a
  linear-filtered `vkCmdBlitImage` cascade, gated to `VK_FORMAT_R8G8B8A8_UNORM`. Guest levels are
  never uploaded there.
* It has **no native `R16G16_UINT` sampled decode at all** — `backend_color_format` falls through to
  `VK_FORMAT_R8G8B8A8_UNORM` for that format, so an integer fetch on this resource would read
  converted texels even at level zero.

The mip really is dynamic, so `Lod=0` is not an available shortcut. From the disassembly of the
captured shader (`gpu_replay --dump-failed-shader`, then `llvm-mc --disassemble`):

```
v_min_f32   v5, v5, v18          ; a log2-derived LOD
v_add_f32   v5, 0xbe800000, v5
v_cvt_flr_i32_f32 v5, v5
v_min_i32   v5, v39, v41         ; clamped against the resource's last level
v_max_i32   v5, 0, v5
v_lshrrev_b32 v0,  v5, v0        ; x >> lod
v_lshrrev_b32 v42, v5, v7        ; y >> lod
image_load_mip v[5:7], [v0, v42, v5], s[32:39] dmask:0x7 dim:SQ_RSRC_IMG_2D
```

That is a runtime-selected level with the coordinates shifted to match. Forcing level zero, or
reading a blit-synthesized chain, would render a **wrong** image — worse than the visible reject, per
#2818's own rule. Closing that half is the shared graphics-side guest-authored upload #2818 scoped;
this PR is `Refs`, not `Fixes`, for that reason, and the falsifications are recorded so the next lane
inherits them.

## Risks

* **Blast radius of the tail widening.** `live_compute` declines a dispatch when a resource's plan
  reports N levels but the binding shape cannot carry them, and that decline is *not* conditional on
  the module issuing `IMAGE_LOAD_MIP` (its own `BLAST RADIUS` comment says so). Admitting the tail
  therefore extends that decline to small mipped sampled textures in shapes that fail
  `chain_binding_shape` — imported/renderer-owned/cube/array/1D/3D bindings. This is the same
  exposure #3048 already took for non-tail chains, now reaching a more common texture size. The
  narrowing facts: the resource must additionally declare ≥2 levels, be a `Texture`, and use a
  straight-copy native format.
* **Semantics of an out-of-range mip.** Unchanged from #3048: the operand is clamped to the
  materialized level count (`CONFIDENCE: MED` — the ISA guide does not spell out MIMG's saturating
  behaviour, and no capture here exercises one).
* **2D_ARRAY NSA** is now described by the shape predicate but is still not lowered — the emitter's
  `dynamic_mip_load` remains 2D-only. The decode arm exists so the header's advertised behaviour is
  exercised rather than dead.

## Build and test

Container build (`distrobox enter ps5ys`), worktree-local build dir:

```
PROSPER_GAME_ROOT=/var/mnt/prosper/testdata cmake -S prosper -B prosper/build-linux    # rc 0
cmake --build prosper/build-linux -j6                                                  # rc 0
cd prosper/build-linux && ctest --no-tests=error                                       # rc 0
```

* **`origin/main` baseline** (same machine, same build dir, measured for this PR):
  `100% tests passed out of 336`, exit **0**.
* **This branch:** `100% tests passed out of 336`, exit **0**. Build exit **0**, zero `error:` lines.
* `test_game_compute` alone: `== PASS == (439 assertions executed)`, exit 0.
* No new ctest case is registered; the three new arms live in existing platform-independent binaries
  (`test_rdna2_decode`, `test_game_compute`, `test_shader_resources`), which is why the count is
  unchanged.
* Every SPIR-V module the new arms compile goes through the live backend, and the repo's
  `spirv-val` gate (`tools/spv_validate`, ctest `spv_validate`) is green in the run above.

### Mutation arms (each applied to this branch, built, and run)

| mutant | build | reddens |
| --- | --- | --- |
| **A** — the shape predicate refuses NSA again (the pre-#3134 rule) | rc 0 | `test_rdna2_decode` rc 1, `test_game_compute` rc 1 — *"names v7 — not v2 — as its mip operand"*, *"an NSA-encoded IMAGE_LOAD_MIP recompiles instead of rejecting the stage"* |
| **B** — the NSA branch reads the FIRST address byte instead of the last | rc 0 | `test_rdna2_decode` rc 1, `test_game_compute` rc 1 — *"the NSA packet's mip byte selects LEVEL ONE's own guest texels"* |
| **C** — the chain plan refuses a tail-packed selected level again | rc 0 | `test_shader_resources` rc 1 (4 arms), `test_game_compute` rc 1 — *"IMAGE_LOAD_MIP recompiles against a tail-packed declared chain"* |

Restored after each: all three binaries rc 0.

Mutant **B** is the one that matters most, because it is the arm a same-source positive control could
not have caught: the fixture puts the mip in **v7** while **v2** — the register the old
consecutive-VGPR arithmetic would have named — holds **0**, so any reading that lands on the wrong
byte fetches level zero and the level-one equality reddens rather than passing vacuously.

## Offline reproduction

The capsule reproduces without a boot, and was re-derived on `origin/main` before any change:

```
gpu_replay <capsule>.prgcap --inspect-only                    # failure[3] draw ... shader-recompile
PROSPER_DBG=1 gpu_replay <capsule>.prgcap --retry-failed-stage 3:1
gpu_replay <capsule>.prgcap --dump-failed-shader 3:1 fs.bin
```

Refs #3134, #2818


mattias800 (commented):
**APPROVED.**

## The premise was verified against the ISA, not inferred from the symptom

The extra dword is RDNA 2's **NSA** encoding — `addr0` in `VADDR`, each later address naming an
arbitrary VGPR one byte at a time. Assembling *both* forms through
`llvm-mc -arch=amdgcn -mcpu=gfx1030` and showing they carry the same address vector
(`[x, y, mip]` at 2D, `[x, y, slice, mip]` at 2D_ARRAY) is primary-source verification rather than a
plausible reading, and it is what makes the rest safe to act on.

And the diagnosis lands where I suspected but could not assert: **the expected-address-count rule was
itself the bug.** `rdna2_mimg_dynamic_mip_shape` derived the mip as `VADDR + n`, true only for the
consecutive encoding, so it declined every NSA packet rather than name a wrong register. That a test
*asserted* the decline is the detail worth remembering — the old behaviour was pinned as correct, so
nothing would have surfaced it.

## The risky half is guarded the right way

Admitting a tail-packed selected level is the change with reach, and it is not admitted on trust. It
requires **two independent derivations to agree** — this plan's AddrLib layout and the descriptor
decode — on `in_tail`, `tail_x`, `tail_y` and `tail_block_bytes`, and returns `{}` otherwise.

That is precisely the lesson of #3197/#3205, where two derivations were assumed to agree, silently
disagreed, and cost a visible *Grand Theft Auto V* regression plus a 32-commit hand bisect. Requiring
the agreement rather than assuming it is the correct response, and the failure direction is a decline
(fail-visible) rather than wrong texels.

The class this unblocks is not a corner: a 32x32 four-byte pyramid is smaller than one 64 KiB
macroblock, so **all six** of Stray's levels sit in the tail. Refusing the tail refused every small
mipped texture, which is a common shape rather than an exotic one.

## The flagged risk is real and I am accepting it deliberately

`live_compute` declines a dispatch when a plan reports more levels than the binding can carry, and
that decline is **not** conditional on the module actually issuing `IMAGE_LOAD_MIP` — its own
`BLAST RADIUS` comment says so. Admitting the tail extends that exposure to small mipped sampled
textures in imported / renderer-owned / cube / array / 1D / 3D bindings.

Accepting it because the exposure is the one #3048 already took, the new surface is a *decline* that
announces itself rather than silent wrong content, and the alternative is leaving a whole texture
class permanently unable to lower `IMAGE_LOAD_MIP`. Flagging it rather than burying it is exactly
what a lane should do with a risk it cannot fully measure.

## `Refs`, not `Fixes` — and that is the right call

The compute half is done; the fragment half is not, and the PR says why rather than claiming the
issue closed: the graphics texture path uploads one level from a blit cascade gated to
`VK_FORMAT_R8G8B8A8_UNORM`, has **no native `R16G16_UINT` sampled decode at all**, and the
disassembly clamps a log2-derived LOD into `v5` while shifting x/y by that same register, so `Lod=0`
is not available either. That is #2818's graphics half plus a native integer sampled format — a
change every title's textures pass through, correctly judged out of scope for one lane that also
could not validate it. Cross-posting the data point to #2818 is the right disposal.

## Evidence

Baseline measured on this machine in the same build dir (`origin/main`: build rc 0, 336/336, ctest
rc 0) rather than quoted, and the branch matches it. Three mutation arms, each applied, built at
rc 0, run, and restored.

**Arm B is the one that earns it**: making the NSA branch read the *first* address byte still
produces a working shader, so an outcome-only assertion would have passed it — the fixture puts the
mip in `v7` while `v2` holds `0`, so the wrong byte fetches level zero and the arm reddens with
*"selects LEVEL ONE's own guest texels"*. That is a test built to catch the plausible wrong answer,
not just the broken one.

I will clear `in-progress` on #3134 by hand, since `Refs` will not close it.

PR #3238, opened 2026-09-02
wt_stash: fix the hook installer on Windows/MSYS and make it prove git honours the hook

mattias800:
`main` is red on **Windows MinGW** (`wt_stash_tool`) after #3234 merged. This fixes it. Refs #3174.

## What actually failed

The job log leads with a `pop` failure, and that is a knock-on rather than the defect:

```
FAIL hook landed in the common dir: got False, want True (...\main-checkout\.git\hooks\reference-transaction)
FAIL git stash is refused: got False, want True (Saved working directory and index state ...)
FAIL refs/stash was not created: got 0, want 1
  ok  wt_stash push works with the hook installed
FAIL wt_stash pop works with the hook installed: got 2, want 0 (wt_stash: no such slot: 'default')
```

`install-hook` reported success while writing the file somewhere git never looks. With the guard
absent, `git stash push` really did stash, so the working tree was clean, so `wt_stash push`
correctly reported *"nothing to park"* and created no ref -- and `pop` then had nothing to find.
Every failing line follows from the one cause.

## Three changes

1. **`_hook_path` asked git for an absolute path.** An MSYS git under native Windows python answers
   `--path-format=absolute --git-common-dir` with an MSYS path (`/tmp/...`), which python then
   resolves against the current drive. A **relative** answer has no root spelling for the two to
   disagree about and survives the crossing; the absolute form stays as a fallback, and a candidate
   that is not a directory from python's own view is now a refusal naming what git offered, never a
   guess.

2. **The hook was written through text mode**, which translates every `\n` to `os.linesep` -- so on
   Windows it lands with CRLF and the shell running it sees a stray CR inside every token (`then\r`
   is not `then`): present, unrunnable, and silent. Written with `newline=''` now, and asserted on
   the **bytes**, which makes the arm catchable from every platform rather than only from the one
   that breaks.

3. **The class fix: `install-hook` now proves git honours the hook** instead of reporting success on
   the strength of its own `write()`. It asks git to do the one thing the hook exists to refuse -- a
   sentinel write to `refs/stash` -- and if that goes through it **removes the hook again and
   fails**, rather than leaving a file that reads as protection. The check is skipped rather than
   forced when `refs/stash` is already occupied, because racing somebody's entry to test a guard
   against losing work would be its own answer; the cleanup passes the sentinel as the expected old
   value, so it can only ever delete the ref it created.

## Two hypotheses checked and found clean

Recorded rather than changed, since both were raised and both are worth having settled:

* **The hook body does end in an explicit `exit 1`** after its stderr block, so it does not fall off
  the end with the status of the last `echo`. A new arm asserts that on the bytes.
* **The hook cannot abort `wt_stash`'s own ref writes.** It exits 0 for every ref that is not
  `refs/stash`, and the suite asserts an ordinary commit, a branch creation, and `wt_stash push`/
  `pop` all still work with it installed. (That blast-radius arm predates this PR and reddens under
  a mutation that widens the hook to every ref.)
* **CRLF on the hook's stdin** is not a factor -- git writes LF there. The CRLF problem was in the
  hook **file**, which is change 2.

## Verification

```
python3 prosper/tools/test_wt_stash.py     101 assertions, rc=0
ctest --no-tests=error                     100% tests passed out of 337   rc=0
ctest --no-tests=error -R wt_stash_tool    9/337 wt_stash_tool Passed 1.65s
git diff --check                           rc=0
```

Mutation arms re-run against the unmodified suite. The interesting change is that three mutations
that used to reach the behavioural assertions now fail at **install time** instead -- which is the
installer's self-verification doing its job:

| mutation | caught by |
| --- | --- |
| hook written with platform newlines (CRLF) | `hook is written with LF endings only` |
| installer trusts its own write, no verification | `install verified itself against git` |
| hook installed where git never looks (the Windows bug) | `install exit: got 2` -- *"wrote ... but git did not honour it"* |
| hook guard deleted (always allows) | `install exit: got 2` (previously only `git stash is refused`) |
| hook body's final `exit 1` becomes `exit 0` | `install exit: got 2` |
| `push` writes the shared `refs/stash` | `lane A parked into the per-worktree namespace` |
| `push` overwrites an occupied slot | `second push into the same slot is refused` |
| `check` always fires | `own entry alone does not fire` |

Unchanged and deliberately untouched: the positive control that reproduces the original #3174
collision with two linked worktrees and plain `git stash`, and every arm covering the core
`refs/worktree` parking. Only the hook path, its byte-level encoding, and the installer's honesty
changed.

## If Windows MinGW is still red

Then the honest reading is that a `reference-transaction` hook is not reliable on the MSYS route,
and the follow-up is to gate the **hook-installed arms only** to non-Windows with that stated
plainly, plus an issue for the real behaviour -- the hook is opt-in and off by default, so that is a
legitimate outcome. The parking arms and the positive control would stay enabled everywhere
regardless. With change 3 in place a Windows user gets the truth at run time either way: the
installer now refuses rather than reporting a guard it did not establish.

### DIFF
--- a/prosper/tools/test_wt_stash.py
+++ b/prosper/tools/test_wt_stash.py
@@ -353,6 +353,19 @@ def arm_hook():
         hook = Path(main) / '.git' / 'hooks' / 'reference-transaction'
         check('hook landed in the common dir', hook.exists(), True, str(hook))
         check('and is executable', os.access(hook, os.X_OK), True)
+        # The installer proves git honours the hook rather than trusting its own write, and says
+        # so. Assert the claim is present: an installer that reports success on the strength of a
+        # successful `write()` is what shipped a silently-absent guard to Windows in the first
+        # place.
+        check('install verified itself against git', 'verified: git refused' in out, True, out)
+        check('the sentinel it wrote was cleaned up', rev(main, 'refs/stash'), None)
+        # LF endings, asserted on the BYTES. Text-mode writing translates '\n' to os.linesep, so
+        # on Windows the hook lands with CRLF and the shell running it sees a stray CR inside every
+        # token -- present, unrunnable, silent. This arm is the one that catches that directly, and
+        # it is checkable from every platform rather than only from the one that breaks.
+        raw = hook.read_bytes()
+        check('hook is written with LF endings only', b'\r\n' in raw, False)
+        check('and ends in an explicit refusal', raw.rstrip().endswith(b'exit 1'), True)
 
         write(a / 'shared.txt', 'LANE-A\n')
         rc, out, err = git_rc(a, 'stash', 'push', '-m', 'blocked')
--- a/prosper/tools/wt_stash.py
+++ b/prosper/tools/wt_stash.py
@@ -361,19 +361,65 @@ def cmd_check(args):
 
 
 def _hook_path():
+    """Where git looks for this repository's reference-transaction hook, as PYTHON spells it.
+
+    Asking git for an absolute path is the obvious way and it is wrong wherever git and the
+    interpreter disagree about path spelling -- an MSYS git under native Windows python answers
+    `/tmp/...`, which python then resolves against the current drive. The install lands somewhere
+    real, git never looks there, and the guard is silently absent. Measured on this project's
+    Windows MinGW CI, where install-hook returned 0 and `git stash` went on working. A RELATIVE
+    answer has no root spelling to disagree about and survives the crossing; the absolute form
+    stays as a fallback, and a candidate that is not a directory here is a refusal rather than a
+    guess -- a hook installer that cannot prove where it wrote is worse than none.
+    """
     configured = git('config', '--default', '', '--get', 'core.hooksPath').strip()
     if configured:
         raise Fail('core.hooksPath is set to {!r}; this installer only manages the default '
                    'hooks directory. Install the hook there by hand, or unset the config.'
                    .format(configured))
-    common = git('rev-parse', '--path-format=absolute', '--git-common-dir').strip()
-    return os.path.join(common, 'hooks', 'reference-transaction')
+    tried = []
+    for fmt in ('relative', 'absolute'):
+        rc, out, _ = git_rc('rev-parse', '--path-format=' + fmt, '--git-common-dir')
+        if rc != 0 or not out.strip():
+            continue
+        cand = os.path.abspath(os.path.join(os.getcwd(), out.strip()))
+        tried.append(cand)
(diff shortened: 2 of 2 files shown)


mattias800 (commented):
**APPROVED.** This fixes `main`, and the third change is worth more than the fix.

## The diagnosis is a chain, not a guess

The job log leads with the `pop` failure, and the PR correctly identifies that as the *last* link:
`install-hook` reported success while writing where git never looks → the guard was absent →
`git stash push` really stashed → the working tree was clean → `wt_stash push` correctly said
"nothing to park" and created no ref → `pop` found nothing. Every failing line follows from one
cause, and reading it that way rather than patching the visible symptom is why the fix is small.

Both platform causes are the classic pair and both are correctly identified:

- **MSYS path crossing** — an MSYS git under native Windows python answers
  `--path-format=absolute --git-common-dir` with an MSYS path that python then resolves against the
  current drive. Asking for a *relative* answer removes the root spelling the two disagree about.
  Keeping absolute as a fallback and turning a non-directory candidate into a refusal that names what
  git offered, rather than a guess, is the right failure shape.
- **Text-mode CRLF** — the hook landed with `\r\n`, so the shell saw `then\r`, which is not `then`:
  present, unrunnable, and **silent**. That last property is what made it survive review.

## The class fix is the part I would keep

`install-hook` used to report success on the strength of its own `write()`. That is an installer that
**cannot fail** — the same shape as the alias wrapper #3234 correctly refused to ship, one level up,
and it is precisely why a broken hook read as protection for a whole merge.

Now it asks git to do the one thing the hook exists to refuse — a sentinel write to `refs/stash` —
and if that goes through it **removes the hook again and fails**. Verifying the effect rather than
the action is the difference between installing a guard and believing you did.

Skipping that check when `refs/stash` is already occupied, rather than forcing it, is good judgement:
racing another lane's entry to test a guard against losing entries would be self-defeating.

## Asserting on bytes is what makes this catchable

Checking the hook's **bytes** rather than its text means a Linux CI run can now catch a Windows-only
line-ending defect. A test that can only fail on the platform that breaks would have left this class
undetectable from where the work actually happens — which is exactly how it reached `main`.

## For the record

This landed on `main` because **I merged #3234 while Windows MinGW was still red** — my gate matched
`bad=Windows MinGW` as though it were an empty list, so it was a discriminator that could not fail.
The lane's original work was sound and its review stands; the defect it is fixing here is a genuine
Windows portability bug that my broken gate simply failed to hold back. I have switched to comparing
the failed-job count numerically.

Issue #3240, opened 2026-09-02
The colour-target and depth/stencil persistent caches still have unguarded entry points outside render_draw_pass_rgba


mattias800:
## Summary

`#2953` closed the memory-safety hole in the *closed* half of the render backend's
persistent-resource domain: `render_draw_pass_rgba` now holds
`BackendPersistentResourceGuard` for its whole body, and every `static` that function
exclusively owns — the persistent texture cache, the pipeline cache, the pipeline-layout
cache and their byte totals and generation counters — is covered completely, because
nothing else in the tree touches them.

The **colour-target** and **depth/stencil** caches are not in that position. They live in
the same domain and are mutated inside the guarded region, so two concurrent
`render_draw_pass_rgba` calls can no longer collide over them. But they also have entry
points *outside* the guard, and those are still unsynchronised.

## The unguarded entry points

In `prosper/tests/fixtures/render_runner.h` (a live render path — see
`tests/fixtures/AGENTS.md`):

- `find_persistent_color_target`, `pin_persistent_color_target`,
  `unpin_persistent_color_target`, `invalidate_persistent_color_target`,
  `invalidate_persistent_color_target_guest_write`,
  `restore_persistent_color_target_after_mirrored_write`,
  `destroy_persistent_color_target`, `readback_persistent_color_target`,
  `evict_persistent_color_target`
- `snapshot_persistent_ds_images`, `restore_persistent_ds_image`,
  `invalidate_persistent_ds_guest_write`, `read_persistent_ds_cube_depth_after`

And in `prosper/frontends/shared/live/live_renderer.cpp`, direct iteration of both maps
without going through any accessor that could take a lock:

- `drain_guest_gpu_writes` iterates `persistent_color_target_cache()` and
  `persistent_ds_cache()` and clears validity flags on every entry (the overflow path)
- the metadata-kind query lambda iterates `persistent_ds_cache()` to build its correlation
  list
- a `PROSPER_GFXLOG`-adjacent site reads `persistent_color_target_cache().size()`

`drain_guest_gpu_writes` is the sharp one: it is reached from the capture RTT-seed readers,
from `set_live_target_query` / `set_live_target_reader` (the live-compute path), and from
several sites inside the render callback. `gpu_executor.cpp`'s
`parallel_draw_diagnostic_active()` already carries a comment saying exactly this about
`PROSPER_TEXCONTENT` — *"that callback is NOT a pure read: it begins with
[...]

Issue #278, opened 2026-07-10
UE4 (PPSA17942): steady-state null-jump crash ends run as scene begins — frame-capture gate


mattias800:
## UE4 (PPSA17942): steady-state guest null-jump crash ends the run as the scene begins — blocks capturing a scene frame

With #232 (draw builders) + #273 (−91% shader skips), DOLL now compiles most of its scene shaders and submits real geometry. But a **pre-existing steady-state guest jump-to-null crash** (`rip=0` / `writer 0x0`) terminates the run **right as the scene begins to render** — so we can't capture a full scene frame yet. This is now the frame-capture gate (and blocks reaching gameplay).

### Symptom
Long DOLL render runs (`PROSPER_GUEST_FS=1 PROSPER_NULL_PAGE=1 PROSPER_RENDER=1 boot_trace /root/PPSA17942-app0`) end with a guest jump to a null/zero address (`rip=0`), identical signature before and after the #273 change (so it is NOT the recompiler). It fires at steady state as scene rendering starts — the game builds/submits scene draws, then jumps to null.

### Task
1. Catch it: `PROSPER_FATAL_TRAP=1` under gdb (SIGTRAP at the real fault), or `PROSPER_FAULTMEM`; capture the backtrace + the register/object whose vtable/function-pointer is null (this is the recurring "guest calls a null function pointer because an HLE returned 0 / left an out-struct unfilled" class — cf. the RHI/StartupModule and trophy-OOM walls).
2. Root-cause: which unimplemented/garbage-returning HLE (or missing setup) leaves the null pointer the guest calls. With thread-names (from #244), identify which thread and what game/engine step it's in as the scene begins.
3. Fix with the real contract (correctness-first, CONFIDENCE). Confirm the run survives into steady scene rendering, then **capture a scene frame** (`PROSPER_FRAME_DIR=/tmp/df PROSPER_DUMP_CONTENT=100000` — viewable BMPs; NEVER commit imagery — gitignored).

### Verify
Run survives past scene-begin; a content frame with scene geometry captures; ctest green; **Messenger MUST still render (`vcount=1044`, 0 faults)** if shared code is touched.

Context: `prosper/docs/UE4_APR_IOSTORE_BRINGUP.md`; ext4 fast path `/root/PPSA17942-app0`. area:ue4.

mattias800:
[...]

PR #3241, opened 2026-09-02
render backend: the persistent-resource caches are one domain with one owner (#2953)

### LINKED ISSUE #2953: render_draw_pass_rgba's persistent-texture cache is an unsynchronised function-local static, and a render job thread crashed inside its find()
## Summary

A **host-side** null dereference inside prosper's own backend, on a guest job thread named
`job_render_0`, inside a `std::unordered_map` lookup in `render_draw_pass_rgba`. On *Metaphor:
ReFantazio* (`PPSA20800`) under `boot_trace` this is **deterministic: 5 of 5 runs**. Under
`tools/screenshot`, same build, same title, same route: **0 of 2**.

## What was observed

`boot_trace`, `PROSPER_RENDER=1 PROSPER_GUEST_ARGS=-force-gfx-direct SDL_VIDEODRIVER=offscreen`,
Linux / AMD Radeon 8060S (RADV STRIX_HALO), branch `bringup/metaphor` @ `6ea4e984`:

```text
[prosper] WORKER-THREAD FAULT: sig=11 addr=0x98 rip=0x7bfbe6 (image+0x7bfbe6) rbp=0x640 tid=...
[hwwatch]   writer 0x7bfbe6 is in: 00400000-00977000 r-xp ... /.../build-linux/boot_trace
[prosper]   insn bytes @rip: 4c 8b 80 98 00 00 00 4c
[fault] thread='job_render_0' guest-fs=0x0 rax=0x0 on-guest-TCB=no (host TCB ...)
```

**Two distinct faulting instructions across the five runs**, which is the finding:

| runs | rip | instruction | `addr2line -i` |
| --- | --- | --- | --- |
| 4 of 5 | `image+0x7bfbe6` | `mov r8,[rax+0x98]`, `rax = 0` | `hashtable_policy.h:1464` → … → `render_runner.h:6169` |
| 1 of 5 | `image+0x7692de` | `mov rdi,[rcx+0x140]` | `hashtable_policy.h:1464`, `hashtable.h:2263` |

`hashtable_policy.h:1464` is libstdc++'s `_Hashtable_base::_M_equals`, reached from `find`
(`hashtable.h:890`) through `_M_find_before_node` (`hashtable.h:1968`). Two different map
*instantiations* failing the same way at the same library line is what makes this a class rather
than one bad pointer.

`render_runner.h:6169` is
`auto cached = persistent_texture_images.find(persistent_key);` inside `render_draw_pass_rgba`.
`persistent_texture_images` is a **function-local `static std::unordered_map`** declared at
`render_runner.h:4968`, with `persistent_texture_bytes` and `persistent_texture_generation` beside
it, and a nested `std::unordered_map<TextureBindingKey, PersistentTextureBinding>` inside every
`PersistentTextureImage`. None of them has a lock.

## Why the thread name matters

`job_render_0` does **not** appear anywhere in prosper's source — the only `pthread_setname_np`
calls in the tree are the HLE's own (`src/hle/kernel/hle_kernel.cpp:1878`, `:1880`, `:2085`,
`:2090`), which pass the name a guest gives through `scePthreadRename`/thread create. So this is a
**guest** worker thread, and the guest is submitting graphics from more than one thread. That is
exactly the precondition under which the unsynchronised statics above are a data race.

`CONFIDENCE: MED` on the race being the mechanism. What is established: the crash site, the
determinism under one frontend, the two instantiations, and that the containers are unsynchronised
function-local statics. What is **not** established: that two threads are provably inside
`render_draw_pass_rgba` at the same moment. That is the first thing to check, and it is checkable —
instrument the entry with a thread-id counter, or read the AGC submit path for whether it serialises.

## The frontend difference is itself evidence, and unexplained

Same build, same title, same route, `PROSPER_RENDER=1`:

| frontend | runs | host fault on `job_render_0` |
| --- | --- | --- |
| `boot_trace` | 5 | **5** |
| `tools/screenshot` | 2 | 0 |

`tools/screenshot` instead survives to the title's own SIGFPE (#2951) and writes all its samples.
Whatever differs between the two frontends' submit/present scheduling is a lever on this bug and
worth finding — it is also instrument trap 127 territory, so quote the frontend in any claim here.

## Related, do not merge without evidence

#2945 (one frozen `.prgbundle` replays to five distinct output hashes in fifteen offline replays)
and #2950 (three rung-6 guarded titles failing deterministically on a true `origin/master` build on
this box). Both are renderer-nondeterminism reports on the same machine. They may share a cause with
this; nothing has been shown to link them.

## History

The first version of this issue reported "one fault in five `boot_trace` runs". That was wrong — it
counted runs made *before* #2934's fix, which died in the guest long before reaching the renderer.
Only one post-fix `boot_trace` run existed at the time. The four added afterwards make it 5 of 5.
Recorded rather than silently corrected, because the original number said "intermittent" and the
real one says "deterministic", and those send an investigator to different places.

mattias800:
Fixes #2953

## The race

`render_draw_pass_rgba` (`prosper/tests/fixtures/render_runner.h` — a **live render path**, compiled
into `frontends/shared/live/live_renderer.cpp`, `live_compute.cpp` and `present/present_blit.cpp`,
see `tests/fixtures/AGENTS.md`) owns a set of process-lifetime `static` containers with no
synchronisation of any kind: the persistent **texture** cache and its nested per-image `bindings`
maps, the persistent **pipeline** and **pipeline-layout** caches, the persistent **colour-target**
and **depth/stencil** caches, plus `persistent_texture_bytes`, `persistent_color_target_bytes` and
three generation counters that are all updated by read-modify-write.

#2953 recorded a host SIGSEGV on a guest job thread (`job_render_0`) inside
`persistent_texture_images.find()` — libstdc++'s `_M_equals`, reached through
`_M_find_before_node`, dereferencing a null node — deterministic 5 of 5 under `boot_trace` on
*Metaphor: ReFantazio*. A `find()` racing another thread's rehash reads torn bucket pointers, which
is that crash's shape; #278 fixed the same shape once already, one layer up, on the folded
`GpuState`.

## What I checked before choosing a fix

**The live route is already serialised, and that is why this looked safe for so long.** Both AGC
submit entry points — `agc_driver_submit_dcb` and `submit_dcb_stream` (`SubmitDcb` /
`SubmitDcbFinal` / `SubmitAcb`) — take `g_agc_state_mu` before folding and hold it across
`execute_submit_work`, which is what invokes the render backend; the 2 ms deferred-flush watchdog
thread takes the same mutex. So on a live boot, two concurrent Dcb folds cannot both be in the
backend, and I could not construct the "two threads in `render_draw_pass_rgba` at once" case from
the live path by reading code. That is recorded as a `## Ruled out` row rather than left implicit,
because it is the first place the next investigator would look.

**That does not make the state safe, for a reason that is structural rather than incidental.**
`g_agc_state_mu` lives in `src/hle/graphics/hle_agc.cpp` — a different layer, in a library this
header is *not compiled against*. `gpu_replay`, `boot_trace` and every Vulkan test include
`render_runner.h` and link no HLE at all, so on those frontends nothing serialises the domain.
Nothing in the header stated the invariant, nothing enforced it, and nothing could detect its
violation. State that owns process-lifetime Vulkan objects has to carry its own contract. If two
threads *could* reach it, the code would also be violating Vulkan's external-synchronisation
requirement on `VkQueue` — which the mutation arm below demonstrates it does, loudly.

## The fix

One mutex for the whole domain, taken for the duration of `render_draw_pass_rgba`
(`BackendPersistentResourceGuard`).

**Why coarse.** There is no finer granularity that would be honest: the function looks an entry up,
keeps the iterator or the reference across dozens of Vulkan calls, mutates and evicts through it,
and updates byte totals by read-modify-write. A per-container lock taken and released around each
individual access would describe none of that while reading as if it did.

**Why there.** The guard is constructed **before** `direct_submission`, so the batch destructor —
which submits, and whose failure cleanups touch the texture cache — still runs inside the critical
section. A guard declared after it would be destroyed first and leave exactly those accesses
outside.

**Cost.** One uncontended mutex acquire/release plus four relaxed atomics per call. This is a
per-**submit** cost, not a per-draw one, against a call that records a command buffer, submits it
and waits on a fence — tens of nanoseconds against tens of microseconds at the very least. On the
live route it is uncontended by construction, per the section above. I did not measure a delta
because there is no measurement that could resolve it: `PROSPER_RENDER_TIMING`'s own buckets are in
milliseconds.

**It measures its own premise.** `in_flight` is incremented *before* the mutex acquire, so
`backend_persistent_resource_peak_in_flight()` reports how many threads have ever wanted the domain
at once. That is precisely the question #2953 left open ("what is *not* established: that two
threads are provably inside `render_draw_pass_rgba` at the same moment") and could not settle by
reading code — any run can now answer it. One unconditional stderr line, once per process, reports
the first time it exceeds one. `backend_persistent_resource_overlaps()` counts threads seen inside
the critical section simultaneously; it is 0 while the guard is taken and non-zero the moment it is
not, which is what the regression test asserts on.

## Alternatives rejected

- **`thread_local` caches.** Rejected on semantics, not on cost. Each thread would get its own
  cache, so hit rates fall and — the part that actually disqualifies it — every `VkImage`,
  `VkDeviceMemory`, `VkImageView`, `VkSampler` and `VkPipeline` in it is duplicated per thread, with
  each copy's lifetime now tied to a thread rather than to the eviction budget. The budgets
  (`persistent_texture_cache_limit()`, `persistent_color_target_limit()` — 4 GiB each on this box)
  are per-cache, so N threads means N budgets. It would also break the caches' actual job: a
  persistent target written by one submit and sampled by the next stops being findable if the two
  submits land on different guest threads, which turns a memory-safety bug into a rendering bug.
- **Confining the accesses to one thread.** This is the outcome I would have preferred and it is
  *nearly* what already happens — but the confinement would be enforced by a mutex three layers away
  in a library half this header's consumers do not link (above). Making it real would mean giving
  the backend its own render thread and marshalling every call onto it, which is a much larger
  change with its own latency consequences, and it does not become correct without still knowing
  that nothing else can touch the caches.
- **A `std::recursive_mutex` also taken in `render_draws_rgba`**, so a multi-segment split is one
  atomic critical section. Rejected: a recursive mutex hides re-entrancy rather than documenting it,
  and the segment boundary is safe — no cache iterator or reference is carried across it; the only
  cross-segment carriers are CPU pixel vectors and a colour-target *identity* that each segment
  re-looks-up by key. That is the same granularity two consecutive submits already have.

## Second finding, fixed here because it is one word

`backend_texture_upload_stats_storage()` was the only per-call stats storage in the file that was
**not** `thread_local` (its five siblings — colour-target, resource-reuse, hash totals, pipeline
cache, render timing — all are). It is written inside `render_draw_pass_rgba` and read by the caller
after the call returns, so a second rendering thread could clobber it in between: a data race on a
non-atomic object, and silently wrong numbers even where it did not tear. Every reader in the tree
reads it on the thread that just rendered, so per-thread storage is what the contract already was.

## Deliberately unaffected

- **Single-threaded behaviour is bit-identical.** No cache policy, key, budget, eviction order or
  Vulkan call changed; the diff to `render_draw_pass_rgba` is one declaration.
- **No new environment variable, and no way to switch the guard off.** A lock with an opt-out is a
  lock nobody can rely on.
- **The colour-target and depth/stencil caches' *other* entry points are still unguarded** —
  `invalidate_persistent_color_target*`, `readback_persistent_color_target`,
  `snapshot_persistent_ds_images`, and `live_renderer.cpp`'s direct iteration of both maps. Two
  concurrent `render_draw_pass_rgba` calls can no longer collide over them, but a call racing one of
  those helpers still can. Extending the guard there needs the frontend to participate and has a
  real self-deadlock failure mode (some of those helpers are called from *inside* the guarded
  region), so it is filed as **#3240** rather than bundled here. #3240 also records a raw-pointer
  discipline problem in the same area found while reading.

## Risks

- **Deadlock.** The guard is non-recursive, so a re-entrant call into `render_draw_pass_rgba` would
  hang. There are exactly two call sites in the tree, both in `render_draws_rgba`, both at top level
  and sequential. The one callback the guarded region does invoke into the frontend
  (`persistent_color_target_evict_sink()`) touches only the frontend's own `g_rtt` and takes no
  backend lock. #3240's follow-up work is where this needs care.
- **Held across a fence wait.** The critical section includes `submit_and_wait`. That is not new
  latency on the live route — `g_agc_state_mu` is already held across the same work — but it does
  mean a second thread would block for a whole submit rather than for a map operation. That is the
  correct trade: the alternative is undefined behaviour.
- **The `[render] N threads ...` line is unconditional.** One line per process, maximum. It is the
  discovery that a design assumption is false, and the run that most needs to report it is the one
  nobody armed a diagnostic for. `check_diag_gates.py` passes.

## Verification

Linux, AMD Radeon 8060S (RADV STRIX_HALO), in the `ps5ys` container. Configure carried
`PROSPER_GAME_ROOT` so the dump-gated cases really run.

| arm | build | ctest |
| --- | --- | --- |
| baseline (`origin/main`'s `render_runner.h`, same worktree/build dir) | exit **0** | **339/339 passed**, exit **0** |
| this branch | exit **0** | **340/340 passed**, exit **0** |

```
cmake -S prosper -B prosper/build-linux            # PROSPER_GAME_ROOT set; Vulkan found, dump PPSA24651
ctest --no-tests=error                             # 340/340, exit 0, 96.80s
```

New case, verified **by name**:

```
ctest --no-tests=error -R backend_persistent_resource_lock   # 1/1 passed, exit 0
```

Ran 5x back to back for flakiness: exit codes `0 0 0 0 0`. The render arm reports
`peak threads wanting the domain during the render arm: 4`, i.e. all four threads were in flight
together — the arm cannot have passed vacuously.

### Mutation arm

Deleted the single `const BackendPersistentResourceGuard persistent_resource_guard;` line from
`render_draw_pass_rgba` and rebuilt only the test target:

```
cmake --build prosper/build-linux --target test_backend_persistent_resource_lock -j6   # exit 0
```

Then four runs of the binary and one ctest run:

| run | result |
| --- | --- |
| 1 | exit **134** — `test_...: ../amdgpu/amdgpu_internal.h:178: update_references: Assertion 'atomic_read(dst) > 0' failed.` |
| 2 | exit **134** (same assertion) |
| 3 | exit **134** (same assertion) |
| 4 | exit **139** (SIGSEGV) |
| ctest | exit **8**, `backend_persistent_resource_lock (SEGFAULT)` |

4 of 4 fatal. The failure lands inside **libdrm's amdgpu reference counting**, which is a second,
independent statement of the same thing the map corruption says: without the guard, concurrent calls
also drive `vkQueueSubmit` on one `VkQueue` from several threads, which Vulkan requires the
application to synchronise externally. Arms 1 and 2 still pass in the mutant, correctly — the
mutation removes the guard from the production function only, and those two arms test the guard
object itself.

### Repository gates

`check_diag_gates.py` (108 findings, all baselined), `check_ascii_output.py`, `check_cached_env.py`,
`check_ctest_gate.py`, `check_contribution_shape.py --selftest`, `check_numbered_table.py` over all
tracked `.md`, and `git diff --check` — all exit 0.

## Test design, and where it is honest about its limits

The concurrency arms are probabilistic in principle. Three things bound that:

1. **A positive control that runs first.** Arm 1 drives the same thread shape with no guard and
   asserts the overlap detector reports a non-zero count. It is built by hand out of atomics rather
   than by removing the production guard, so it draws from a different source than the null it
   validates and observes no undefined behaviour of its own. Without it, a clean zero in arms 2 and 3
   would be equally well explained by "the guard works" and by "these threads never overlapped".
2. **`peak_in_flight >= 2` is asserted, not assumed.** A run in which the threads failed to overlap
   *fails* rather than passing quietly. Because `in_flight` is incremented before the mutex acquire,
   a thread blocked on the guard still counts, and the critical section is a full command-buffer
   record + submit + fence wait against a microsecond-scale rendezvous release — which is why the
   measured peak is the full 4 of 4.
3. **What would make it fully deterministic** is a scheduling hook inside the critical section, which
   would mean shipping a test-only seam through the live render path. The in-flight assertion buys
   the same protection against a vacuous pass without one.

## `## Ruled out` rows added

- `docs/METAPHOR_STATUS.md` — the crash cannot have come from two concurrent `sceAgcDriverSubmitDcb`
  folds; both entry points and the deferred-flush watchdog already hold `g_agc_state_mu` (#278), so
  looking for the second thread *between two submits* is a dead end. What remains open is which route
  the observed thread took, and `backend_persistent_resource_peak_in_flight()` now answers that on
  any run.
- `tests/fixtures/AGENTS.md` — the header cannot inherit the HLE's locks, with the reason (three of
  its consumers link no HLE) and the boundary of what the new guard covers.

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -2694,6 +2694,20 @@ if(TARGET prosper_core)
       set_tests_properties(gds_scan_memo_disabled PROPERTIES
                            ENVIRONMENT "PROSPER_NO_GDS_SCAN_MEMO=1")
 
+      # The backend's persistent-resource domain lock (#2953). Carries its own positive control --
+      # an unguarded arm that proves the overlap detector fires at all -- so a clean zero in the
+      # guarded arms cannot be read as "the threads never overlapped".
+      add_executable(test_backend_persistent_resource_lock
+                     tests/render/test_backend_persistent_resource_lock.cpp)
+      # Includes a shared fixture, which resolves from the `tests` root.
+      target_include_directories(test_backend_persistent_resource_lock PRIVATE tests)
+      # Reaches `shared/...` through render_runner.h, which resolves from `frontends`.
+      target_include_directories(test_backend_persistent_resource_lock PRIVATE frontends)
+      target_link_libraries(test_backend_persistent_resource_lock
+                            PRIVATE prosper_core Vulkan::Vulkan)
+      add_test(NAME backend_persistent_resource_lock
+               COMMAND test_backend_persistent_resource_lock)
+
       # The address-watch matcher's edge cases (exclusive end, empty range, UINT64_MAX wrap).
       add_executable(test_compute_address_watch tests/gpu/test_compute_address_watch.cpp)
       target_link_libraries(test_compute_address_watch PRIVATE prosper_core Vulkan::Vulkan)
--- a/prosper/docs/METAPHOR_STATUS.md
+++ b/prosper/docs/METAPHOR_STATUS.md
@@ -288,6 +288,17 @@ One line per hypothesis this work killed, so nobody re-derives it at full cost.
   appeared to falsify the fix under that lever was executed against a **stale `screenshot` binary**
   — only the `boot_trace` target had been rebuilt. Recorded because the arm looked like a clean
   negative and was not; rebuild every frontend, not the one you are about to run.
+- **The `job_render_0` crash inside `persistent_texture_images.find()` cannot have come from two
+  concurrent `sceAgcDriverSubmitDcb` folds — that path is already serialised, and had been since
+  #278.** Both AGC submit entry points (`agc_driver_submit_dcb` and `submit_dcb_stream`, i.e.
+  `SubmitDcb`/`SubmitDcbFinal`/`SubmitAcb`) take `g_agc_state_mu` before folding and hold it across
+  `execute_submit_work`, which is what invokes the render backend; the 2 ms deferred-flush watchdog
+  thread takes the same mutex. So a second thread reaching the backend's persistent caches has to
+  arrive by some other route, and looking for one *between two submits* is a dead end. The lock
+  added in #2953 is not a duplicate of that one: it lives in `render_runner.h`, which `gpu_replay`,
+  `boot_trace` and every Vulkan test compile **without linking the HLE at all**, so on those
+  frontends nothing was serialising the domain. What remains open is which route the observed
+  thread took; `backend_persistent_resource_peak_in_flight()` now answers that on any run. #2953.
 
 ## Instruments this work added
 
--- a/prosper/tests/fixtures/AGENTS.md
+++ b/prosper/tests/fixtures/AGENTS.md
@@ -29,6 +29,16 @@ Two consequences for anything you change in it:
   false trigger would cost.
 - **It cannot depend on the app** — no capture singleton, no frontend state — because it is also
   compiled into Vulkan tests that link none of that.
+- **It cannot depend on the HLE's locks either, and that one is easy to miss** (#2953). The live
+  route reaches this backend only under `g_agc_state_mu` (`src/hle/graphics/hle_agc.cpp`, #278), so
+  reading the live path alone makes the persistent caches look safely single-threaded. `gpu_replay`,
+  `boot_trace` and every Vulkan test compile this header and link no HLE, so on them that invariant
+  simply does not exist. Process-lifetime state declared here carries its own synchronisation:
+  `render_draw_pass_rgba` holds `BackendPersistentResourceGuard` for its whole body, which covers
+  every `static` that function owns. It does **not** cover the colour-target and depth/stencil
+  caches' other entry points (`invalidate_persistent_color_target*`,
+  `readback_persistent_color_target`, `snapshot_persistent_ds_images`, and the frontend's direct
+  iteration of both), so do not add a new one without reading #3240.
 
 ## Adding to it
 
--- a/prosper/tests/fixtures/render_runner.h
+++ b/prosper/tests/fixtures/render_runner.h
@@ -739,7 +739,7 @@ struct PersistentPipeline {
 // static that function owns. It does NOT cover the colour-target and depth/stencil caches' OTHER
 // entry points -- `invalidate_persistent_color_target*`, `readback_persistent_color_target`,
 // `snapshot_persistent_ds_images` and the frontend's direct iteration of both caches -- which are
-// reachable without this lock and are tracked separately. Nor is the multi-segment loop in
+// reachable without this lock and are tracked in #3240. Nor is the multi-segment loop in
 // `render_draws_rgba` atomic: the guard is released between segments, which is the same granularity
 // two consecutive submits already have.
 //
--- a/prosper/tests/fixtures/render_runner.h
+++ b/prosper/tests/fixtures/render_runner.h
@@ -564,8 +564,14 @@ struct BackendTextureUploadStats {
     uint64_t persistent_cached_bytes = 0;
 };
 
+// thread_local like every other per-call stats storage in this file (#2953). It was the one
+// exception: a plain static written inside `render_draw_pass_rgba` and read by the caller after the
+// call returns, so a second rendering thread could clobber it between the write and the read -- a
+// data race on a non-atomic object, and silently wrong numbers even where it did not tear. Every
+// reader in the tree reads it on the thread that just rendered, so per-thread storage is what the
+// contract already was.
 inline BackendTextureUploadStats& backend_texture_upload_stats_storage() {
-    static BackendTextureUploadStats stats;
+    static thread_local BackendTextureUploadStats stats;
     return stats;
 }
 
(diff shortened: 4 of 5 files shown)


mattias800 (commented):
**APPROVED.**

## The premise was falsified, and the conclusion survived for a better reason

#2953's suspected mechanism — two concurrent Dcb folds — **cannot happen**: both AGC submit entry
points take `g_agc_state_mu` before folding and hold it across `execute_submit_work`, and the
deferred-flush watchdog takes the same mutex. I verified the mutex and its call sites.

What makes this a fix rather than a wasted lock is the structural half: `g_agc_state_mu` lives in
`hle_agc.cpp`, and `render_runner.h` is included by `gpu_replay`, `boot_trace` and a long list of
Vulkan test binaries that **link no HLE at all** — I checked, and `test_gpu_execute`,
`test_interp_render` and `test_recompiled_fragment` are among them. So the serialisation everyone was
implicitly relying on does not exist for a large fraction of this header's consumers. `hle_agc.cpp`
even carries a comment reasoning *"g_agc_state_mu, so the plain statics are single-threaded"* — true
where it is written, and not true of this file.

"Already almost true, and still worthless as a guarantee" is exactly the right reading.

That also disposes of the cost objection: on the live path the AGC mutex already serialises, so the
new one is uncontended. Coarse is free where it is redundant and load-bearing where it is not.

## Coarse is the honest granularity here

The function looks an entry up, holds the iterator across dozens of Vulkan calls, mutates and evicts
through it, and updates byte totals read-modify-write. A per-access lock would describe none of that
and would read as protection while leaving every one of those windows open. Taking the guard
**before** `direct_submission`, so the batch destructor's failure cleanups stay inside the critical
section, is the detail that a first attempt usually misses.

Rejecting `thread_local` on **semantics** rather than cost is right, and the argument is the correct
one: a target written by one submit would stop being findable if the next landed on a different guest
thread, turning a memory-safety bug into a rendering bug.

## The mutation arm is unusually strong

Deleting the guard gives exit **134, 134, 134, 139**, and three of those die inside **libdrm's own**
`update_references` refcount assertion. That is a second, independent witness — not the project's
test asserting its own expectation, but the graphics stack underneath refusing the access pattern.

And the test cannot pass vacuously: it asserts `peak_in_flight >= 2`, so a run whose threads failed to
overlap **fails** rather than reporting success. Arm 1 is a positive control built from atomics
rather than by removing the production guard, so it tests the detector independently of the thing
being detected. Being explicit that full determinism would need a scheduling hook in the live render
path, and declining to ship a test-only seam for it, is the right trade.

## Instrumenting its own premise

Incrementing `in_flight` **before** the mutex acquire, so `peak_in_flight()` answers #2953's open
question on any run and one unconditional line fires the first time it exceeds 1, turns a fix into a
measurement. Whoever next boots `PPSA20800` gets the answer for free.

## Scope discipline

Filing **#3240** for the colour-target and DS entry points outside the guard — rather than bundling —
is correct, and the reason given is a real one: some of those helpers are called from *inside* the
guarded region, so extending it naively self-deadlocks. The raw-pointer discipline note
(`add_failure_cleanup` capturing raw cached pointers into a batch that outlives the call, where the
texture cache correctly captures the key and re-`find()`s) is a good find to have recorded.

Fixing the second unsynchronised static inline — the one stats storage that was not `thread_local`
while all five siblings were — is proportionate: one word, same class, no reason to defer it.

## Evidence

Baseline measured in the same worktree and build dir (339/339, exit 0) rather than quoted; branch
340/340, exit 0, new case confirmed by name and run 5× consecutively. CI 9 success / 2 skipped /
0 failures.

On the `git stash` slip: self-reported, caught immediately, entry verified as yours before popping,
and the stack confirmed back to its prior state. That is the recognition cue trap 247 describes,
working — and reporting it unprompted is worth more than not having done it.

Issue #89, opened 2026-07-08
Regression: boot_reaches_first_syscall (test #26) stalls before graphics — 0 gfx calls


mattias800:
## Summary
`ctest -R boot_reaches_first_syscall` (test #26, `tests/test_boot_linux.cpp`) fails **consistently** on current master (`caae580`):
```
guest reached 0 distinct unimplemented system calls; 2 GC stop-the-world exception(s); 0 graphics-lib call(s)
== FAIL: stalled early (0 < 1 unimpl) ==
== FAIL: never reached graphics init (0 gfx calls) — boot regressed before GPU/display ==
```
It reaches the IL2CPP GC (2 stop-the-world exceptions) but **never reaches any graphics-lib call** (was ≥1 before). Reproduced 3/3 runs, so not flaky.

## Scope / impact
The **minimal 3-module boot** (`eboot + Il2cpp + PS5Util`) regressed. The **full `boot_trace`** (all modules + `PROSPER_GUEST_FS=1 -force-gfx-direct`) is NOT affected — it still submits 1525 DCBs and reaches level0-5 — so this is a minimal-boot / CI-gate regression, not a full-boot break. But it's a real behavior change between GC and graphics init.

## When introduced
In the ~20 commits merged 2026-07-08 after `2f682f5` (#60). Threading/init-touching candidates to bisect first: #86 (`scePthreadOnce` serialize-per-control), #77 (`__tls_get_addr` DTV purge on thread exit), #75 (`RELEASE_MEM`/`WaitRegMem`). `git bisect` between `2f682f5..caae580` running `ctest -R boot_reaches_first_syscall` will pin it.

## Repro
`cd prosper/build-linux && cmake --build . -j8 && ctest -R boot_reaches_first_syscall --output-on-failure`

mattias800:
CLAIMING: claude-code (session_01YZDzeB73TmdwmakWEurFB9) | branch fix/issue-89-boot-regression | 2026-07-08T12:57:56Z

mattias800:
Bisect lead from the author of the equeue rework: add **PR #81 (`fix/issue-67-equeue-lifetime`, merged in the same window) to the top of the candidate list** — it isn't in the issue's list but it deliberately changed `sceKernelWaitEqueue` semantics in exactly the way that would produce "stalls between GC and graphics init, full render boot unaffected":

[...]

PR #3242, opened 2026-09-02
windows: the import trampoline places floats from the handler's own signature (#2955)

### LINKED ISSUE #2955: The Windows import trampoline remaps integer registers only, so every HLE handler taking a float reads it — and every argument after it — from the wrong place
## Summary

prosper's Windows import-stub trampoline converts a guest SysV call to the host MS ABI by remapping
**integer registers only**. It never touches xmm. Every HLE handler that declares a floating-point
parameter therefore reads that parameter — and, worse, every parameter *after* it — from the wrong
place on Windows.

Found while implementing `libSceFont`'s glyph rasterizer (#2951). It is **pre-existing and general**,
not introduced by that work.

## The mechanism

`src/host/image/exec_image_win.cpp`, `emit_sysv_to_ms_prologue`, is the whole conversion:

```text
push [rsp+0x28] x4     ; guest stack args a6..a9
sub  rsp,0x30
mov  [rsp+0x20],r8     ; MS 5th = guest a4
mov  [rsp+0x28],r9     ; MS 6th = guest a5
mov  r9,rcx            ; MS 4th = guest a3
mov  r8,rdx            ; MS 3rd = guest a2
mov  rdx,rsi           ; MS 2nd = guest a1
mov  rcx,rdi           ; MS 1st = guest a0
```

No xmm register is read or written. That is fine for the overwhelming majority of handlers, because
the two ABIs agree that integer arguments are positional and the shuffle above is exactly that.

It is not fine when a float is in the middle, because **the two ABIs count arguments differently**.
SysV keeps separate counters for integer and SSE arguments; MS numbers every argument in one
sequence and picks the register file from the type. For
`f(void*, uint32_t, void*, float, float, void*, void*)`:

| arg | SysV (what the guest sets) | MS (what the handler reads) |
| --- | --- | --- |
| 1 | rdi | rcx |
| 2 | rsi | rdx |
| 3 | rdx | r8 |
| 4 (float) | **xmm0** | **xmm3** |
| 5 (float) | **xmm1** | **stack** |
| 6 | rcx | stack |
| 7 | r8 | stack |

The trampoline puts guest `rcx` into `r9` and guest `r8` at `[rsp+0x20]` — MS slots 4 and 5 — but a
handler with two floats at positions 4 and 5 looks for its **pointers** at slots 6 and 7. So it reads
two pointer arguments out of guest `r9` and the guest stack and, if it writes through them, performs
an **arbitrary memory write**.

## Affected today

Every HLE handler in the tree that takes a float. `grep` finds four, all in `src/hle/util/hle_font.cpp`:

| NID | name | consequence on Windows |
| --- | --- | --- |
| `N1EBMeGhf7E` | `sceFontSetScalePixel` | reads the scale from xmm1/xmm2 — garbage. A face there can be carrying an unreviewed pixel size. |
| `6vGCkkQJOcI` | `sceFontSetupRenderScalePixel` | same handler, same problem |
| `TMtqoFQjjbA` / `lz9y9UFO2UU` | `sceFontSetEffectSlant` | garbage slant |
| `v0phZwa4R5o` / `XIGorvLusDQ` | `sceFontSetEffectWeight` | garbage weight |

`sceFontRenderCharGlyphImage` (`3G4zhgKuxE8`) would have been the fifth and the dangerous one — it
writes through two out-pointers. #2951 works around it by declaring the Windows arm with the five
**integer** arguments only, which the trampoline does deliver correctly, and rendering at the
line-box origin. That is safe and correct for a one-glyph-per-surface caller, but it is a workaround
in the handler rather than a fix in the trampoline, and the `#if defined(_WIN32)` in
`hle_font.cpp` should disappear when this is fixed.

Note the clamp added in #2951 (`kMaxPixelScale`) exists partly because of this issue: a garbage
scale reaching a real rasterizer must not become a huge allocation.

## Why it has not bitten before

Only four handlers take floats, all in one file, all in a library whose callers were previously
stubs. Nothing else in 700+ registered NIDs declares a float parameter — the guest-facing surface is
overwhelmingly integers and pointers.

## Suggested fix

The trampoline cannot know a handler's signature, so this needs per-registration information:
a small descriptor on `Hle::register_fn` naming which argument positions are floats, and an
`emit_impl` variant that places them. Alternatively, and more cheaply, forbid float parameters on
HLE handlers by convention and have each such handler take the bits as integers — which does not
work either, because SysV puts them in xmm where an integer declaration cannot see them.

Whatever the mechanism, the acceptance test is the same and should be written first: a handler with
the exact `(ptr, u32, ptr, float, float, ptr, ptr)` shape, called through a real guest stub, that
asserts all seven arguments arrive intact. That test would fail on Windows today.

`CONFIDENCE: HIGH` on the mechanism (read directly from the emitter). **Not reproduced on a Windows
host** — this box is Linux — so the consequence table above is derived from the ABI and the emitted
bytes, not observed. Someone with a Windows build should confirm before the fix is designed around it.

mattias800:
Fixes #2955.

## The defect

The PS5 guest is **System V AMD64**; a Windows host is **Microsoft x64**. Every import stub is a
trampoline between the two, and it converted **integer registers only** — it never touched xmm.

The consequence is larger than "float arguments are missing", because the two conventions do not
merely name different registers, they **count** differently:

| | integer args | FP args | counting |
| --- | --- | --- | --- |
| System V | `rdi rsi rdx rcx r8 r9` | `xmm0..xmm7` | **two independent counters** |
| Microsoft x64 | `rcx rdx r8 r9` | `xmm0..xmm3` | **one positional counter**, register file chosen by type |

So one float in the middle of a signature desynchronizes the counters and **displaces every argument
behind it as well**. For `sceFontRenderCharGlyphImage(ptr, u32, ptr, float, float, ptr, ptr)`:

| arg | guest sets (SysV) | handler reads (MS) | old trampoline delivered |
| --- | --- | --- | --- |
| 3 (float) | `xmm0` | `xmm3` | nothing |
| 4 (float) | `xmm1` | stack slot 4 | nothing |
| 5 (ptr) | `rcx` | stack slot 5 | into `r9` |
| 6 (ptr) | `r8` | stack slot 6 | into stack slot 4 |

The handler then wrote through two pointers it had read from the wrong places — an **arbitrary
memory write**, not a wrong glyph. #2951 worked around it by declaring a Windows-only five-integer
arm of that handler; this PR deletes that `#if`.

## Two things the issue did not have

**The census was an order of magnitude short.** #2955 says "grep finds four, all in
`src/hle/util/hle_font.cpp`" and "nothing else in 700+ registered NIDs declares a float parameter".
`src/hle/libc/hle_libc.cpp` registers the entire libm bank through the same `Hle::register_fn` and
therefore the same trampoline — `sinf`, `pow`, `fmod`, `hypot`, `ldexpf`, `frexp`, `sincosf`,
`modf`, `copysign`, `fmin/fmax`, the hyperbolics — plus `strtod`/`strtof`. **78** registered NIDs
declare a float or a double, not four. The test counts them from the built binary and fails on zero,
rather than trusting a grep.

**Arguments were only half of it.** The stub runs a host checkpoint call *after* the handler returns
and saved only `rax` across it. `xmm0` is volatile under Microsoft x64, so **every float- or
double-returning handler handed the guest whatever that call left behind** — the whole maths library
and `strtof`, whose entire purpose is a value in `xmm0`. Their *arguments* were fine (with no integer
argument the two counters advance together), which is exactly why this half was invisible.

## The fix

The bridge becomes signature-driven, in a new `src/host/abi/`:

- `call_signature.hpp` — a `CallSignature` deduced from the handler's **own C++ declaration** by
  `signature_of`, so it cannot drift from the function it describes. A shape the SysV classifier does
  not model (aggregate by value, `long double`, `__int128`) is a compile error, not a silent
  misplacement.
- `sysv_ms_bridge.hpp/.cpp` — the two placement tables and the emitter. Compiled on **every**
  platform, which is what makes it testable off Windows.
- `Hle::register_typed` preserves the declaration instead of casting it to `HleFn`; the ~78 affected
  registrations in `hle_font.cpp` and `hle_libc.cpp` use it.

Placement needs no scratch register and no cycle breaking, because emitting in **descending argument
position** is provably safe: every stack destination precedes every register destination, and a
Microsoft destination register is never a pending move's source (`rcx`=SysV[3], `rdx`=SysV[2],
`r8`=SysV[4], `r9`=SysV[5] sit at positions 0..3, so each would need SysV index >= 2 while the
pending position is < 3, and a SysV index never exceeds its position). The reasoning is in the
emitter; the executed tests are what actually establish it.

`emit_one_stub` now stages into a buffer and copies only what fits. Previously it wrote straight into
the slot and compared the length *afterwards* — by which point an oversized stub had already overrun
the next one.

## Linux and macOS: a no-op, deliberately and by construction

Their host ABI **is** the guest's, so no trampoline exists at all — the Linux stub is a bare
tail-jump and nothing in this PR is on that path. The one shared change is `hle_font.cpp` losing its
Windows-only arm, which restores the *Linux* declaration on both platforms.

Inside the bridge the no-op is structural rather than incidental: a signature with no FP argument and
no FP return is **not recorded at all**, so those handlers keep the historical bytes. A test asserts
that an undeclared signature and a declared integer-only one emit byte-identical code, and pins the
fixed shuffle against an independently written golden array.

Verified on this Linux box:

- **The placement tables**, asserted directly. A pure function of the signature; no host needed.
- **The emitted machine code, executed.** `__attribute__((ms_abi))` gives Linux a real
  System V -> Microsoft x64 boundary, so `tests/host/abi/test_sysv_ms_bridge.cpp` calls the actual
  trampoline bytes with a System V call and asserts every argument arrives — including an integer
  argument *following* a float, which a naive "add xmm handling" fix still gets wrong.
- **The same bytes on a real Windows binary.** Cross-compiled with MinGW and run under wine, a
  scratch probe (not committed) executed the trampolines on a genuine Microsoft x64 host. The fixed
  bridge delivered all seven arguments; the signature-blind bridge delivered `f3=4.48e-43`,
  `f4=4.59e-39` and two pointers reading `0x810000` / `0x7c0000` instead of the caller's, and
  returned the clobber pattern instead of a float.
- **And in CI, on a real Windows host.** `Windows MinGW` builds the whole tree and runs the whole
  suite with `--no-tests=error`, so `sysv_ms_bridge` executes the emitted trampolines there as well
  — on Windows the attributes flip and it is the *caller* that carries `sysv_abi`, which exercises
  the same boundary from the other side. `macOS (x86_64 / Rosetta 2)` gives a third host.

**Not verified:** a live guest calling a Sony import on a Windows host. Nobody here has a Windows
box. The command that would settle it is the ordinary Windows run — build with
`cmake -S prosper -B prosper/build-windows -G Ninja -DPROSPER_APP=ON` under MSYS2 UCRT64, then
launch a title that uses `libSceFont` (Metaphor: ReFantazio, `PPSA20800`) and confirm the
language-selection screen's glyphs still render and that a run reaches the same point it does on
Linux. Until someone does that, the claim in this PR is "the bridge is correct", not "Windows works".

## Verification the tests carry themselves

Each executed group has a discriminator, so a green run cannot be a run that proved nothing:

- the **signature-blind** bridge must FAIL to deliver the same call (if it succeeded, the test could
  not see the defect it exists for);
- the **signature-blind** float return must be DESTROYED by the clobbering checkpoint (if it
  survived, the save/restore arm would be untested);
- the registry sweep fails if **zero** handlers declare a float signature, which would mean the typed
  registrations are not reaching the registry.

## Mutation arms

Both were built and run; the build succeeded in each case, so the failures are the test's, not the
compiler's.

| --- | --- | --- | --- |
| `ms_arg_location` places an SSE arg at its **SSE index** instead of its **position** (the naive fix) | `0` | `8` | 6 failures — 3 table arms and 3 executed arms, incl. `argument 3 arrived as 484378, expected 1.5` |
| drop the `xmm0` save/restore around the return checkpoint | `0` | `8` | 3 failures — `(f,i)->f`, `(d,d)->d` and `(ptr,ptr)->f` returns |

Reverted; rebuild `0`, test `0`.

## Risks

- **The riskiest part is the one that cannot be run here**: the change alters the machine code
  installed for 78 imports on a platform this box cannot boot. It is mitigated by executing those
  exact bytes twice (Linux `ms_abi`, and a MinGW binary under wine), and by the untouched path being
  byte-identical for everything else.
- **Stub slot size.** A signature-driven stub is longer. The largest observed is 72 bytes against a
  96-byte slot (the historical path is 83, or 96 with a return hook), and a test sweeps every
  declared signature in the live registry with and without a hook. `install_stubs` also still refuses
  an oversized stub — and now refuses it *before* writing.
- **`printf`-family handlers are deliberately untouched.** They are real host variadics; a variadic
  Microsoft x64 call additionally wants each FP argument duplicated into the integer register, and a
  `CallSignature` cannot describe an argument list the format string decides at run time. They keep
  the fixed path they have always had. Filed as #3246 with the mechanism, the affected handlers and
  the acceptance test.
- **Registration shape.** `hle_font.cpp` spells `Hle::register_typed` at each site rather than behind
  a local templated lambda, because `tools/re/hle_handler_map.py` cannot read that shape — the
  reconcile test caught exactly that and is what pins it.

## Ruled out

Recorded in `docs/PORTING.md`, new subsection *"Ruled out — the Windows import trampoline's
floating-point arguments"*:

- "Only four handlers are affected, all in `hle_font.cpp`" — false, 78.
- "An all-float signature is safe because both conventions place SSE args in `xmm0..` in order" —
  true of the arguments, false of the return.
- "Adding xmm moves to the positional integer shuffle fixes it" — false by construction, pinned by
  test.

## Four commits

1. **the fix itself**;
2. **a portability follow-up caught by the macOS job**: `is_abi_scalar_v` was a variable template
   ending in `sizeof(T) <= 8`, and a variable template's initializer is instantiated whole — so the
   `is_void_v<R> ||` guard in front of it did not stop `sizeof(void)` being formed for the
   void-returning `sincosf`/`sincos`. GCC accepts `sizeof(void)` as an extension worth 1; clang does
   not. Now a class template with an explicit `void` specialization, plus `static_assert` coverage of
   the deducer's edge shapes so every platform's build checks them;
3. **a test call site caught by the Windows job, and the more important of the two.**
   `font_render_hle` **SegFaulted** on the real Windows host: removing `hle_font.cpp`'s five-integer
   Windows arm left `test_font_render.cpp` still calling that arm's shape behind its own
   `#if defined(_WIN32)`, so the seven-argument handler read `metrics` and `result` out of slots a
   five-argument call never wrote — this issue's own arbitrary write, on the test's side of the
   boundary. That commit also carries the independent review's five non-blocking findings (a width
   guard next to the staging buffer, cv-stripping on `sse_return`, `Kind::None` for an out-of-range
   query, the documented stub-slot ceiling, and two executed arms for a System V-spilled float and a
   disp32 source displacement);
4. **docs**: the remaining variadic gap is #3246 now, rather than a paragraph nothing tracks.

## Commands and results

```
cmake -S prosper -B prosper/build-linux            # configure 0
cmake --build prosper/build-linux -j6              # build 0
```

| | tests | result |
| --- | --- | --- |
| `origin/main` baseline | 339 | 100% passed, exit `0` |
| this branch | 340 | 100% passed, exit `0` |

The new case is registered as `sysv_ms_bridge` and was verified by name in the ctest listing. Its own
output: `78 registered handlers declare a float signature`.

Cross-platform check of the same bytes:

```
x86_64-w64-mingw32-g++ -O2 -std=c++20 -I prosper/src -o probe.exe probe.cpp \
    prosper/src/host/abi/sysv_ms_bridge.cpp     # exit 0
wine probe.exe                                  # exit 0, PROBE OK (0 failures)
```

No snapshots: nothing here touches the renderer.


mattias800:
### Author verification, exact head `c3774d7a`

```
cmake -S prosper -B prosper/build-linux                       # 0
cmake --build prosper/build-linux -j6                         # 0
ctest --test-dir prosper/build-linux --no-tests=error         # 0
```

| | tests | exit |
| --- | --- | --- |
| `origin/main` baseline (`637e65f9`) | 339 | `0`, 100% passed |
| this branch | 340 | `0`, 100% passed |

The one new case is registered as `sysv_ms_bridge` and verified by name in the ctest listing
(`267/340 Test #267: sysv_ms_bridge ... Passed`). Its own output:
`78 registered handlers declare a float signature`.

### Mutation arms

| mutation | build | ctest | failures |
| --- | --- | --- | --- |
| `ms_arg_location` places an SSE argument at its SSE index rather than its argument position — the naive "just add xmm moves" fix | `0` | `8` | 6: 3 table arms and 3 executed arms, e.g. `MS arg 3 is xmm0, expected xmm3` and `argument 3 arrived as 484378, expected 1.5` |
| drop the xmm0 save/restore around the return checkpoint | `0` | `8` | 3: `(f,i)->f`, `(d,d)->d`, `(ptr,ptr)->f` returns |
| `is_abi_scalar_v` back to the variable-template form (commit 2) | clang `1`, gcc `0` | — | `invalid application of 'sizeof' to an incomplete type 'void'` |

Reverted after each; rebuild `0`, `sysv_ms_bridge` `0`.

### The emitted code, for review

`(ptr, u32, ptr, float, float, ptr, ptr)` — the #2955 signature, 72 bytes into a 96-byte slot:

```
sub    rsp,0x48
mov    QWORD PTR [rsp+0x30],r8      ; arg6 -> MS stack slot 6
mov    QWORD PTR [rsp+0x28],rcx     ; arg5 -> MS stack slot 5
movsd  QWORD PTR [rsp+0x20],xmm1    ; arg4 -> MS stack slot 4
movaps xmm3,xmm0                    ; arg3 -> xmm3, not xmm0
mov    r8,rdx                       ; arg2
mov    rdx,rsi                      ; arg1
mov    rcx,rdi                      ; arg0
movabs rax,<handler> ; call rax
mov    QWORD PTR [rsp+0x38],rax
movabs r11,<checkpoint> ; call r11
mov    rax,QWORD PTR [rsp+0x38]
add    rsp,0x48 ; ret
```

Descending argument position, so every stack destination is emitted while its register source is
still intact, and no register destination is a pending move's source. `strtof (ptr,ptr)->f` adds
`movsd [rsp+0x28],xmm0` / `movsd xmm0,[rsp+0x28]` around the checkpoint call. An undeclared signature
still emits the historical 83 bytes, unchanged.

### The same bytes on a Microsoft x64 host

Cross-compiled with MinGW, run under wine (scratch, not committed):

```
glyph (ptr,u32,ptr,f,f,ptr,ptr):
  emitted 72 bytes                       <- signature-driven
  emitted 83 bytes                       <- signature-blind, what main emits
  signature-blind delivery: f3=4.48416e-43 f4=4.59098e-39
                            p5=0x810000(want 0x31fdca) p6=0x7c0000(want 0x31fdcb)
strtof (ptr,ptr)->f:
  signature-blind float return: -6.25985e+18 (want it NOT to be 1234.5)
PROBE OK (0 failures)
```

Those two pointer values are the arbitrary write #2955 predicts, observed rather than derived.

The `Windows MinGW` CI job builds the whole tree and runs the whole suite on a real Windows host, so
`sysv_ms_bridge` executes the emitted trampolines there too — on Windows the attributes flip and the
*caller* is the `sysv_abi` one, which is the same boundary from the other side.

**Still not verified, and this box cannot verify it:** a live guest calling a Sony import on Windows.

mattias800:
### Independent review: APPROVED, no blocking findings

The reviewer went well past reading. Their own work, on top of what is in the PR:

- an **exhaustive static simulation** of the placement tables, frame arithmetic and move ordering
  over **all 8,190 signature shapes** the type system permits (0-12 arguments × every class mask ×
  both return classes): 0 clobbers, 0 alignment violations, 0 save/argument-slot collisions, max
  **169 bytes** emitted;
- **disassembly** of the emitter's real output for 9 shapes, every byte decoded;
- **executed** four shapes the PR's test did not cover, across the real System V -> Microsoft
  boundary, with their own discriminators: the signature-blind bridge produced 31 failures in that
  harness, and an **ascending**-order mutation of the emitter produced 11 — so the descending order
  is load-bearing and their check could see it break.

Their conclusion on the ordering argument: they could not construct a counter-example.

### What the review changed, in `4c6ed6af`

1. `emit_sysv_to_ms_bridge` refuses a signature wider than `kMaxArgs` instead of running off the
   caller's 256-byte staging buffer. Unreachable today — `signature_of` static-asserts it — but
   `register_fn_with_signature` takes a raw `CallSignature`, so the guard belongs next to the buffer.
2. `sse_return` now strips cv, as the argument classes already did. A handler returning
   `const float` would have taken the historical path and had its return destroyed, silently.
3. An index past the end of a signature answers `Kind::None`, not a plausible stack slot 0.
4. `kMaxBridgeBytes` documents the measured worst case and the real constraint — the 96-byte stub
   slot, which caps a float-bearing signature at about eight arguments.
5. Two executed arms for the placements nothing covered: a System V-spilled float and a disp32
   source displacement, in one twelve-argument call.
6. `docs/METAPHOR_STATUS.md` no longer describes the gap as present.

### And the thing CI caught, which is the more important half

`Windows MinGW` on `c3774d7a`: **262/263 passed, `font_render_hle` SegFaulted.** Removing
`hle_font.cpp`'s five-integer Windows arm left `test_font_render.cpp` still calling that arm's shape
behind its own `#if defined(_WIN32)`, so a seven-argument handler read `metrics` and `result` out of
slots a five-argument call never wrote — the same arbitrary write this issue is about, on the test's
side. Fixed in `4c6ed6af`; the pen-shift arm, previously Linux-only, now runs everywhere.

**The bridge's own test passed on that Windows host**: `254/263 Test #254: sysv_ms_bridge ...
Passed`. That is the emitted trampoline bytes executing across the ABI boundary on a real Windows
machine, which is as close as this can get without a guest.

### CI on `c3774d7a` (all complete)

| job | result |
| --- | --- |
| Linux | pass |
| Linux App / Linux App Package | pass |
| Windows App | pass |
| macOS (x86_64 / Rosetta 2) | pass — red before the `sizeof(void)` fix, which is how that was found |
| macOS App (SDL3 + MoltenVK) | pass |
| Contribution shape / Docs | pass |
| Windows MinGW | **fail** — `font_render_hle` SegFault, fixed in `4c6ed6af` |

### Local, exact head `4c6ed6af`

`cmake --build` `0`; `ctest --no-tests=error` `0`, **340/340 passed**. `test_sysv_ms_bridge` `0`,
`test_font_render` `0`. MinGW cross-compile of the scratch probe `0`, under wine `PROBE OK (0
failures)`.

Third mutation arm, for the new width guard: with `if (sig.count > kMaxArgs)` disabled the build
still exits `0` and the test exits `1` with `over-wide signature: was emitted instead of refused`.

Waiting on CI for this head before this is mergeable.

mattias800:
### CI green on `797e31a6`, all jobs

| job | result |
| --- | --- |
| Linux | pass |
| Linux App / Linux App Package | pass |
| **Windows MinGW** | **pass — 263/263, 100%** |
| Windows App | pass |
| macOS (x86_64 / Rosetta 2) | pass |
| macOS App (SDL3 + MoltenVK) | pass |
| Contribution shape / Docs | pass |

The two rows worth naming:

- `254/263 Test #254: sysv_ms_bridge ... Passed` — **on the Windows host**. The emitted trampoline
  bytes executed across the System V -> Microsoft x64 boundary on real Windows, for every shape
  including an integer argument following a float, with the signature-blind discriminators still
  required to fail.
- `89/263 Test #89: font_render_hle ... Passed` — the SegFault from the previous head is gone.

Ready for review of the merge decision. Still unverified, and unverifiable from here: a live guest
calling a Sony import on a Windows desktop.

mattias800 (commented):
**APPROVED.**

## Both findings beyond the issue are the valuable part

**The census was short by an order of magnitude.** #2955 says four handlers, all in `hle_font.cpp`.
`hle_libc.cpp` registers the whole libm bank through the same trampoline — **78 registered NIDs**
declaring a float or double, with `sincosf`/`frexpf`/`modff` in the same arbitrary-write class as the
font handler. That changes this from a corner fix to a broad correctness one, and it was found by
looking rather than by trusting the issue's list.

**The bridge also destroyed float returns**, which the issue does not mention at all. The stub ran a
host checkpoint call after the handler and saved only `rax`; `xmm0` is volatile under MS x64, so every
maths function and `strtof` returned whatever that call happened to leave behind. The reason this
stayed invisible is the sharp bit: an **all-float signature places its arguments identically under
both conventions**, so the half of the bug that a float-only API would have exposed was exactly the
half that looked fine.

## The naive fix is wrong, and the test proves it

The obvious repair is "place each SSE argument at its SSE index". That is precisely the mutation arm,
and it reddens **6** checks — because System V counts integer and SSE classes *separately* while MS
x64 counts them *together by position*, so an integer argument following a float is displaced under
the naive rule. Building the arm for the plausible wrong fix, not just the absent one, is what makes
this suite worth having.

Descending-position placement being provably free of scratch spill and cycle breaking is the right
kind of argument — a property of the ordering rather than a claim that the cases happen to work.

## Verified on the platform that has the bug

`sysv_ms_bridge` **passed on the real Windows MinGW host** (263/263), executing the emitted trampoline
bytes across the ABI boundary with `sysv_abi`/`ms_abi` flipping sides. That is far stronger than the
cross-compiled assertions I would have accepted, and the wine run showing the *predicted* arbitrary
write — `f3=4.48e-43`, pointers reading `0x810000`/`0x7c0000` instead of the caller's — closes the
loop between theory and observation.

The Linux argument is structural rather than empirical, which is better: the host ABI **is** the
guest's there, so no trampoline exists; and inside the bridge an integer-only signature is not
recorded at all, with a test asserting declared-integer-only and undeclared emit byte-identical code
against an independent golden array.

## Two CI catches worth keeping in mind

macOS clang rejecting `sizeof(void)` where GCC accepts it as an extension, and Windows MinGW catching
a **SegFault in the pre-existing `font_render_hle`** — removing the handler's Windows arm left the
*test* still calling that arm's five-integer shape behind its own `#if`. The same arbitrary write, on
the test's side, found only because the platform job actually runs. That is the argument for keeping
those jobs meaningful.

## Scope

Filing **#3246** for variadic handlers rather than bundling them is correct, and the reason is a real
constraint rather than fatigue: MS x64 varargs additionally want each FP argument duplicated into the
integer register, and a compile-time `CallSignature` cannot describe a run-time argument list.

## Merge handling

The branch conflicted after #3241 and #3244 landed. I merged `origin/main` in: the only conflict was
`BLOG.md` (both entries kept), and I checked the other three shared files are **pure additions**
(`CMakeLists.txt` +21/−0, `PORTING.md` +26/−0, `METAPHOR_STATUS.md` +11/−0), with every deletion
against `main` confined to files this PR legitimately rewrites. Trap 41 clear.

Independent review APPROVED with all five non-blocking findings applied; the reviewer simulated all
8,190 signature shapes and executed four my author's tests missed.

Not verified, and correctly stated as such: a live guest calling a Sony import on a Windows desktop.
The PR names the exact command.

PR #3243, opened 2026-09-02
zerowatch: a one-window verdict was published as a claim about the range -- profile it, and follow the content (#3142)

mattias800:
## What this is

Instrumentation and a correction, not a fix. `Refs #3142` — the issue stays open, and its own framing
changes.

## The failure this addresses

`PROSPER_ZEROWATCH` (#3145) arms on the **first non-zero 256-byte window** of an APR payload and
polls *that window*. Its verdict line names the whole payload:

```
[zerowatch] dst=0x3002330000 size=13674533 sampled=256B@0x0 ... WENT-ZERO after 2.267s
```

The window it always picks is window 0, which on a streamed UE4 chunk is the chunk's serialization
header — so the instrument is aimed at the one place that always changes. The verdict was carried
into #3142 and into `docs/STRAY_STATUS.md` as *"the range is zeroed"* / *"the data is gone"*, which
is a claim about a multi-megabyte range that the instrument had never tested.

Profiling **16** windows instead of one says otherwise. Stray (`PPSA02101`), `PROSPER_NULL_PAGE=1`,
no input, 120 s warmup, `tools/screenshot`:

| | destinations |
| --- | --- |
| still holding most of their payload **45 s after** their own WENT-ZERO | **11 of 14** (`kept` 4–15 of 16 windows) |
| genuinely all-zero | **3 of 14** |
| a window that read zero at the flip and **non-zero again** at +10 s | 1 |

So what the guest clears about a second after a load is, in the common case, the chunk header; the
asset body survives. The 0.5–2.3 s figure and the 2-minute reproduction are unaffected — they are
correct *about window 0*. What is withdrawn is the inference from them to "the range".

## Behavioural contract

* `PROSPER_ZEROWATCH` behaves as before, and additionally prints a 16-window profile at the flip and
  at +10 s / +45 s. Each window is printed as `dst/src` against the **source's** own non-zero census,
  so a window the payload never filled cannot be counted as a loss.
* `PROSPER_ZEROWATCH_FIND=N` (new, default off) searches the guest address space
  (`[0x100000000, kGuestAutoMapLimit)`) for the payload's own bytes — a **control** scan taken while
  the data is demonstrably present, then a second scan at the flip. The control is what makes the
  second result falsifiable: without it an empty scan is indistinguishable from a scanner that never
  reached the mapping. An AFTER-ZERO scan is skipped, loudly, when its control found nothing.
* `PROSPER_ZEROWATCH_FIND_PREFIX=16` (new, default 32) — 16 bytes is the needle length that survives
  Gen5 tiling, derived from prosper's own swizzle tables rather than from a generic Z-order reading:
  `src/gpu/texture/tile.cpp:149` puts an 8-byte element (BC1) in an **8×4** micro-tile, and
  `tile.cpp:275`'s bit order for 8-byte elements (`x0 y0 y1 x1 x2 y2 x3 y3 x4`) puts a row's first
  four blocks at byte offsets 0, 8, **64**, 72. So two consecutive blocks stay adjacent and four do
  not, and a 32-byte needle reports "absent" for a payload the guest merely re-tiled. The run must
  start on an **even block**, so the needle offsets are 16-aligned — and that is alignment relative
  to the destination, which is alignment relative to the surface only if the surface starts
  16-aligned inside the buffer. Both conditions are stated in the code and in the doc.
* `prosper::host::guest_memory_search()` / `memory_search_ranges()` — the new reusable half. Every
  result carries its own **scope** (ranges scanned, ranges unreadable, bytes scanned, bytes
  unreadable, budget-exhausted, hit-cap-reached, enumeration-failed), because a budget cut-off, an
  unreadable range, an enumeration that produced no ranges at all, and a genuine absence otherwise
  produce the same empty hit list. A hit needs `prefix_len` bytes and reports `full` only when the
  whole needle is present and equal, so a match at a range end is never claimed as full over bytes
  that were not read.

Nothing is on by default; with none of the switches set, no new code runs.

## What the new instrument measured

* **No verbatim relocation.** With the 16-byte needle, the payload has exactly **one** guest-visible
  copy while it is live — the destination itself — and **none** at any guest address after the flip.
  Stated limit, in the header and in the doc: a payload the guest decompressed or re-encoded is not
  the same bytes at any needle length, so the null means "no verbatim guest copy", never "the data
  does not exist".
* A *wider* scan does find these payloads in prosper's **own host-side buffers**, which is why the
  scan is bounded to guest space — prosper holding a copy answers nothing about where the guest put
  it, and those hits fill the hit cap ahead of the guest ranges.
* **Record→submit is microseconds apart** on this title (`sceAmprAprCommandBufferReadFile` →
  `H896Pt-yB4I(CbSetEqueue)` → `sceKernelAprSubmitCommandBufferAndGetResult -> token=0x209f (bound)`,
  same guest thread, no intervening work), so the natural next hypothesis — that prosper's eager read
  at *record* time races the guest's own preparation of a destination the hardware would only fill at
  *submit* — is dead. Recorded in `## Ruled out` because the code comment "prosper serves every read
  eagerly" invites it.
* The destinations are a **recycled pool**: 1,147 of 9,631 distinct APR destinations in one 90 s boot
  receive reads from two or more different file offsets, so "was this address reused?" has to be asked
  per address rather than assumed either way.

## Deliberately unaffected

The APR read path, the write verifier, the write trace, the mapping dump and the `libc.prx`
attribution are untouched — every one of those measurements stands. This changes only what the
watchdog *says about extent*, and adds a way to follow content instead of an address.

## Risks

Low. All new behaviour is behind switches that default off; the one always-on change is extra
`fprintf` output on a `PROSPER_ZEROWATCH` run, which is already a diagnostic-only mode. The content
scan runs on the watchdog's own poll thread and takes seconds, which delays that thread's other
WENT-ZERO detections — the header says so explicitly, and timings from a `FIND` run are not to be
quoted.

## Review

Independently reviewed and initially **REJECTED** at `c7686b45` with five blocking findings, all
fixed in `fd1dbb92` — the test/constant coupling above; a `WENT-UNMAPPED` path that never stamped
`t_reported`, so both re-profile thresholds fired within 100 ms and printed `lost=0 kept=0` for a
range that had entirely vanished (trap 248's own shape inside the patch that adds trap 248, and the
trap row now says so); the BC1 Morton derivation, which was 8×8/four-elements against prosper's own
8×4/eight-elements table; the 16-byte needle's missing 16-byte alignment, which made its
tiling-survival true at most one placement in four; and prefix mode being blind to the last
`needle_len - 1` bytes of every range. Nine of eleven non-blocking notes taken as well.

## Verification (exact head `0b999e41`; re-run at `fd1dbb92`, same counts)

Linux, distrobox `ps5ys`, `-DPROSPER_APP=ON`.

```
cmake -S prosper -B prosper/build-linux -DPROSPER_APP=ON        # exit 0
cmake --build prosper/build-linux -j6                            # exit 0
ctest --no-tests=error                                           # exit 0
```

* **Baseline on `origin/main`, measured here, not quoted:** `100% tests passed out of 340`, exit 0.
* **This branch:** `100% tests passed out of 341`, exit 0 — the one added case is `guest_memory_search`.
* Registered in the platform-independent block; verified by name:
  `ctest --no-tests=error -R '^guest_memory_search$'` → `1/1 Passed`, exit 0.
* Docs gate: `check_numbered_table.py --ordered --table-header Instrument` → `numbered column 1..248,
  248 rows, unique and ascending`, exit 0. `ctest -R 'diag|env|docs|trap'` → 14/14, exit 0.
* `git diff --check` → clean.

### Mutation arms

The straddling case is the one the regression test exists for — a scanner that drops its overlap
still returns a plausible hit list, so the blindness is invisible on a live scan and silently turns a
null into a wrong answer.

Three, each with build exit 0 and each reverted afterwards to a green suite:

| mutation | result |
| --- | --- |
| `overlap = needle_len - 1` → `0` | 4 failures (straddle ×3 + boundary sweep) |
| `overlap = 0` **and** `kChunkBytes = 8<<20` | 4 failures — **green before review**, see below |
| last chunk's `need = prefix_len` → `needle_len` | 3 failures (prefix-at-range-end + short range ×2) |

The middle arm is the one that earns the review. Independent review measured that the test's
`kChunk` was coupled to the implementation's constant by a comment only: raise the chunk to 8 MiB and
the "straddle" needle lands in the middle of chunk 0, the case silently becomes a duplicate of the
mid-chunk one, and the overlap can then be deleted outright with ctest green. The test now derives
its placement from an exported `memory_search_chunk_bytes()`, asserts the placement really straddles,
and sweeps every offset across the boundary rather than one.

The suite also carries its own arms for the cases where an empty result is *not* a finding: an
unreadable range is counted and never treated as zeros, a budget cut-off is marked (with a paired
unbounded scan that does find the needle), a range the budget abandons before its first fetch is not
counted as scanned (with a two-range control), and a prefix-only match is reported as `full=false`
with a strict-search arm proving the prefix search is what found it.

## Docs

* `docs/STRAY_STATUS.md` — the affected `## Ruled out` row is corrected **at its head** as well as
  in a `CORRECTION` bullet below it, since a retraction placed only underneath the number it retracts
  is one most readers never reach. Two new bullets record the content-scan result and the recycled
  pool; a new row records the record-vs-submit falsification.
* `docs/GAME_COMPAT_ORCHESTRATION.md` — instrument trap **248**: a watchdog whose verdict covers one
  256-byte window while its output names a multi-megabyte range. Cousin of 236.
* `src/host/memory/AGENTS.md` — new; the folder had none.

## Snapshots

Not run. No renderer, GPU or guest-facing behaviour changes; every code path added is behind a
diagnostic switch that defaults off.

No screenshot and no `BLOG.md` entry: nothing visible moved, and this PR contains no image.

Refs #3142, #3126, #3145.


mattias800:
## Author verification — exact head `c7686b45`

Linux, distrobox `ps5ys`, configured `-DPROSPER_APP=ON`, `PROSPER_GAME_ROOT` set.

```
cmake --build prosper/build-linux -j6      BUILD_RC=0
ctest --no-tests=error                     CTEST_RC=0    100% tests passed out of 341
```

Baseline measured here on `origin/main` rather than quoted: `100% tests passed out of 340`, exit 0.
The one added case is `guest_memory_search`; verified by name at this head —
`ctest --no-tests=error -R '^guest_memory_search$'` → `1/1 Passed`, exit 0.

`git diff --check` clean. Docs gate at this head:
`check_numbered_table.py --ordered --table-header Instrument` → `numbered column 1..248, 248 rows,
unique and ascending`, exit 0.

### CI at this head — all required checks green

| check | result |
| --- | --- |
| Linux | pass (10m18s) |
| Linux App | pass |
| Linux App Package | pass |
| Windows MinGW | pass (9m58s) |
| Windows App | pass |
| macOS (x86_64 / Rosetta 2) | pass (13m9s) |
| macOS App (SDL3 + MoltenVK) | pass |
| Docs | pass |
| Contribution shape | pass |

Nine pass, two skipping (`Progress tracker`, `Publish Release`), zero failures.

The third commit is a CI-found fix, recorded rather than squashed away: macOS failed at
`guest_memory_search.cpp:128:16: error: use of undeclared identifier 'process_vm_readv'` — Darwin has
no such symbol and every other POSIX caller in the tree already includes
`host/platform/posix_shim.hpp` for it. That file did not.

### Mutation arm, re-stated with its exit codes

`const size_t overlap = needle_len - 1;` → `= 0` in `memory_search_ranges`:

```
cmake --build ... --target test_guest_memory_search    BUILD_RC=0
ctest --no-tests=error -R '^guest_memory_search$'      ***Failed
    FAIL: straddle: exactly one hit, got 0
    FAIL: straddle: hit address
    FAIL: straddle: full match
    guest_memory_search: 3 failure(s)
```

Reverted → `BUILD_RC=0`, `1/1 Passed`, exit 0. Only the straddling case moves, which is the point:
the other eight cases pass with or without the overlap, so they do not stand in for it.

### Live evidence behind the doc changes

All on *Stray* `PPSA02101`, `tools/screenshot`, `PROSPER_NULL_PAGE=1`, no pad input.

* 16-window profile, 120 s warmup: 11 of 14 destinations keep most of their payload at +45 s; 3 are
  all-zero; one zeroed window returns non-zero at +10 s. Reproduced on a second boot with the same
  destination set and the same profiles.
* `PROSPER_ZEROWATCH_FIND=3 PROSPER_ZEROWATCH_FIND_PREFIX=16`: control finds the payload at exactly
  one guest address (the destination); the post-flip scan finds it at none.
* `PROSPER_AMPRLOG=1 PROSPER_FILELOG=1`: record → `CbSetEqueue` → `AprSubmitAndGetResult` adjacent on
  one guest thread, which is what kills the record-vs-submit hypothesis.

No orphan processes left: `pgrep -x prosper-app` / `pgrep -x screenshot` empty before and after every
run.

mattias800 (commented):
# REJECTED (exact head `c7686b45`)

Five blocking findings, every one of them small and four of them one-liners. None of them is in the
chunking arithmetic, which I checked by execution rather than by reading and which is **correct** —
see the "verified, no finding" section at the bottom, because that half is the part a reviewer is
most likely to hand-wave and it earns the opposite.

The reason this is a rejection rather than a nit list: the PR's own subject is an instrument that
published a verdict wider than what it measured, and it lands a permanent trap entry about that
class. Three of the five findings are the same shape inside this diff — a guard that stops guarding
without going red, a profile line printed for a range it never read, and a tiling derivation that
does not match this project's own table. Those cannot land under trap 248's own banner.

Method note: I did not use the project build. I compiled `guest_memory_search.cpp` standalone
(`g++ -O2 -std=c++20`) against a probe harness and against the PR's own test file, in a scratch
directory. Every number below is measured, not reasoned.

---

## Blocking

### B1. The test's chunk constant is coupled to the implementation's and fails GREEN when they diverge — the suite's one real guard is one constant from void

`prosper/tests/host/memory/test_guest_memory_search.cpp:76`
`const size_t kChunk = 4u << 20;          // must match the implementation's chunk size`
against `prosper/src/host/memory/guest_memory_search.cpp:21`. Nothing enforces the comment.

Measured, three arms:

| build | result |
| --- | --- |
| unmodified | `all cases passed`, rc 0 |
| `overlap = 0` (your mutation arm) | `3 failure(s)`, rc 1 — reproduces |
| `kChunkBytes = 8u << 20` **and** `overlap = 0` | **`all cases passed`, rc 0** |

With an 8 MiB chunk, `at = kChunk - kNeedle/2` is 4 MiB — the middle of chunk 0 — so case 2 silently
becomes a second copy of case 1 and the straddle is no longer tested at all. The overlap can then be
deleted outright and ctest stays green. This is the exact failure direction the file's own header
paragraph is about ("a silent partial blindness ... nothing about the output says the scan was
incomplete"), applied to the test instead of the scanner.

Fix (either): export the constant — `size_t memory_search_chunk_bytes();` in the header, `kChunk`
derived from it in the test; or make case 2 place the needle at several candidate boundaries
(`1<<20, 2<<20, 4<<20, 8<<20, 16<<20`) minus `kNeedle/2` and assert one hit at each, so at least one
placement straddles whatever the constant becomes. The first is better because it also lets the test
assert that a straddle *happened*.

**B1b, same file, `:95-97`.** The case comment claims "one that overlaps but does not de-duplicate
reports two". Measured: replace `if (addr >= accept_from)` at `guest_memory_search.cpp:77` with
`if (true)` and the entire suite still passes, rc 0. No test in the file can detect the removal of
the de-duplication, and (see N1) the implementation cannot produce the duplicate the comment
describes. A stated failure mode that the suite cannot show it detects is the "discriminator that
cannot show its lever moved" shape — either drop the sentence or add an arm that actually reddens.

### B2. `WENT-UNMAPPED` leaves `t_reported` at 0, so an unmapped destination immediately prints a bogus `+10s` and `+45s` profile reading `lost=0 kept=0`

`prosper/src/hle/fs/hle_file.cpp:2937-2946` sets `e.reported = true` and returns without setting
`e.t_reported`; only the WENT-ZERO path at `:2968` sets it. `diag_now_us()` is
`steady_clock::now().time_since_epoch()` (`prosper/src/diagnostics/diag_clock.hpp:27-31`) =
CLOCK_MONOTONIC, whose epoch is system boot — so on any machine up for a minute,
`diag_now_us() - 0` is far past both thresholds.

The re-profile guard at `:2929-2933` therefore fires on the **next two 50 ms polls**: a
`zero-profile(+10s)` line and then a `zero-profile(+45s)` line, both within ~100 ms of the unmap,
for a destination `zw_read256` can no longer read — so every window prints `??`, and the summary
reads `payload-windows lost=0 kept=0`, which is the least true thing that line can say about a
range that just went away. A reader grepping `lost=` gets a zero for a total loss.

Fix: set `e.t_reported` in the unmapped branch too, or gate the re-profile on `e.t_reported != 0`.

### B3. The Gen5 tiling derivation contradicts prosper's own authoritative table — "8x8" and "four elements" are both wrong for BC1

`prosper/src/hle/fs/hle_file.cpp:2824-2827` and `prosper/docs/STRAY_STATUS.md:402-405`:
"An 8x8-element micro-tile ... elements (0,0) and (1,0) ... stay adjacent, while (2,0) lands four
elements away."

`prosper/src/gpu/texture/tile.cpp:149` gives the 256B standard-swizzle micro-tile dimensions:
`1 B -> 16x16, 2 B -> 16x8, 4 B -> 8x8, 8 B -> 8x4, 16 B -> 4x4`. BC1 is 8 bytes per block element,
so its micro-tile is **8x4**; 8x8 is the *4-byte*-element case. `tile.cpp:275` gives the within-tile
bit order for 8 B as `x0 y0 y1 x1 x2 y2 x3 y3 x4`, i.e. element-index bit 0 = x0, bit 1 = y0,
bit 2 = y1, bit 3 = x1. Evaluating it:

```
(0,0)->0   (1,0)->1   (2,0)->8   (3,0)->9      byte offsets 0, 8, 64, 72
```

So (2,0) is **eight** elements away, not four. (Four is what you get for a 4-byte element, or from a
generic y-low Z-order — and note that under a generic y-low Z-order (0,0) and (1,0) would *not* be
adjacent, which would sink the headline claim entirely. The claim survives only because prosper's
real 8 B pattern puts x0 in bit 0.)

**Your conclusion is unaffected and in fact stronger** — blocks 0 and 1 land at bytes 0 and 8, so 16
contiguous bytes survive; a 32-byte linear run needs blocks 0..3 at bytes 0, 8, 64, 72 and does not.
But the derivation is what the next reader inherits from a status doc and a code comment, it is one
`grep` from `tile.cpp` away, and this project's charter is explicit that a cited mechanism reads as
a verified one. Please correct both places to 8x4 / eight elements, and cite `tile.cpp:149,275`.

### B4. The 16-byte survival is conditional on an alignment the instrument does not arrange

Falls straight out of B3's arithmetic: a 16-byte run survives tiling only if it starts on an **even**
block, i.e. at an offset `≡ 0 (mod 16)` relative to the surface base. A run starting at an odd block
spans blocks 2k+1 and 2k+2, which land at Morton 1 and 8 — 56 bytes apart.

`seed_off` and `seed_deep_off` are masked `& ~(uint64_t)3` (`hle_file.cpp:3052`, `:3065`, `:3072`) —
4-byte aligned only. So at most one placement in four satisfies the condition, before the chunk
header's own offset is even considered. `STRAY_STATUS.md:401-405` nevertheless states the survival
as established and uses it to elevate the null: *"so this is not merely a statement about untiled
copies"*. As built, the instrument cannot support that sentence in general.

Fix: mask the seed offsets `& ~15ull` and say in the comment that the alignment is what makes the
needle tiling-survivable — or, if you would rather not move the offsets, state the condition in the
doc bullet and downgrade the sentence. Either is fine; leaving the claim as it stands is not, because
it is the load-bearing half of "no verbatim relocation".

### B5. In prefix mode the scanner does not meet its own documented contract at a range end — measured 240 misses

`guest_memory_search.cpp:69` `const size_t limit = want - needle_len;`, justified at `:66-68` by
"a needle that would run past the range end is not present in the range". That is true for a **full**
needle and false for a **prefix** match — and prefix mode is the mode the live instrument runs in
(`hle_file.cpp:2859` passes `prefix_len = 16`, `needle_len = 256`). The header at `:65-67` defines
the contract as "A location matches when its first `prefix_len` bytes equal the needle's."

Measured: sweep a 16-byte prefix across the last 300 bytes of a range, `needle_len = 256`:

```
C prefix-near-range-end: found=45 MISSED=240   (blind window = needle_len-1 = 255)
```

Exactly the last 255 bytes of every range are unsearchable in prefix mode. Interior chunk boundaries
are fine (`D prefix-near-chunk-boundary missed=0`) — this is a range-end effect only, and at
255 bytes per mapping it will not have changed this PR's result. It still contradicts the header's
stated contract inside a component whose headline is that its nulls are trustworthy.

Minimum fix is a comment: correct `:66-68` and add the limitation to the header block at `:22-30`.
Better fix is one line — bound the last chunk of a range by `want - prefix_len` and clamp the
full-match test to what is in the buffer. Your call; the comment as written must not stay.

---

## Non-blocking

- **N1 — `accept_from` is unreachable, not load-bearing.** `guest_memory_search.cpp:48-51, 77, 82`.
  With `limit = want - needle_len` and `advance = chunk - (needle_len - 1)`, chunk *k* searches start
  addresses `[pos_k, pos_k + want - needle_len]` and chunk *k+1* begins at exactly
  `pos_k + chunk - needle_len + 1` — the searched start windows are contiguous and **non-overlapping
  by construction**, so the overlap never produces a duplicate hit. Confirmed by removing the guard
  (suite green) and by a 546-placement boundary sweep (zero duplicates). It also cannot drop a
  legitimate hit: addresses ascend within a chunk (`memchr` from `i`, `i = off + 1`) and across
  chunks, and it resets per range — I checked the last part with a deliberately **descending** range
  list and both hits were reported. So your argument holds, but the premise is stronger than you
  claimed. Keep it as belt-and-braces if you like; the header at `:16-17` and the test comment at
  `:96-97` should stop describing it as fixing a hazard the design already precludes.
- **N2 — an empty enumeration yields a completely clean-looking null.** `guest_memory_search.cpp:106`
  swallows a `fopen` failure, and with zero ranges the scope is measured as
  `ranges_scanned=0 ranges_unreadable=0 bytes_scanned=0 budget_exhausted=0 hits_capped=0` — every
  honesty signal negative. On Darwin, which `c7686b45` just made this compile for, there is no
  `/proc/self/maps` at all, so `guest_memory_search()` returns exactly that shape unconditionally.
  `ranges_scanned == 0` is the only tell and the header never names it. Given the file's own headline
  at `:22-25`, a `bool enumeration_failed` (or at minimum a loud `fprintf` on the `fopen` failure)
  belongs in `MemorySearchScope`. The zerowatch caller is saved by its CONTROL arm — that pairing is
  the best idea in the PR — but the reusable API is what is being published.
- **N3 — `bytes_scanned`/`bytes_unreadable` are fetch counts, not distinct bytes.** The
  `needle_len - 1` overlap is re-counted each chunk: measured `8,389,118` for an 8 MiB range (+510 =
  2x255) and `12,583,677` unreadable for a 12 MiB one (+765). Immaterial at that ratio, but
  `byte_budget` is spent on it and the field name says otherwise. One word in the header.
- **N4 — `ranges_scanned` counts a range the budget abandons before its first fetch.** `:46`
  increments before the `:53` check. Measured with 3 ranges and `byte_budget=1`: `ranges_scanned=2`
  while only one was read. `budget_exhausted` is set, so no null is made falsely trustworthy.
- **N5 — `memory_search_ranges` never clears `out`** yet caps on `out.size() >= max_hits`, so a
  caller reusing a vector silently shrinks its own cap. Both current callers pass fresh vectors; one
  line in the header would keep it that way.
- **N6 — `hle_file.cpp:3067`**, `const uint32_t* sw = (const uint32_t*)(p8 + off);`. `off` is
  4-aligned but `p8`'s alignment is not established at this site, and the cast is a
  strict-aliasing/alignment hazard over a buffer written as bytes. The two neighbouring loops
  (`:3053`, `:3075`) read through `uint8_t`. Benign on x86-64; a `memcpy` into a local `uint32_t[64]`
  costs nothing. Bounds themselves are fine — I checked all three new loops: `size >= 1 MiB` is
  guaranteed at `:3032`, and `((size-256)/15)*15` and `((size-256)/7)*7` are both `<= size-256`, so
  every 256-byte read stays inside the payload.
- **N7 — `zw_profile`'s `used += snprintf(prof + used, sizeof prof - (size_t)used, ...)`**
  (`:2894`, `:2899`) is the pattern that walks off the end once `snprintf` truncates, because `used`
  then exceeds the buffer and the size argument underflows. It **cannot** fire here: the worst case
  is 16 x `" w15=64/64"` = 160 bytes into 640. Noting it only so the 4x headroom is a known invariant
  rather than an accident — and because the same pattern already exists at `:3126`/`:3142`, so this
  is consistency rather than a new hazard.
- **N8 — the self-find reasoning at `hle_file.cpp:2853-2856` names the wrong mechanism.** The needle
  is not the APR staging buffer; it is `e.seed`, a 256-byte copy living in `g_zw` and again in the
  `pending_find` copy on the poll thread's stack. What keeps those out of the haystack is the
  `[0x100000000, 0x40000000000)` bound, not the munmap. Conclusion holds. (`0x40000000000` does match
  `kGuestAutoMapLimit`, `src/hle/memory/hle_kernel_mem.cpp:1331` — checked.)
- **N9 — coverage gaps, one cheap arm each:** a needle at offset 0 of a range; **more than one
  range** (every case passes exactly one, so the per-range `accept_from` and `range_unreadable`
  resets are untested — I verified both by hand and they are correct); a *partially* unreadable
  range (a short `process_vm_readv` drops the whole 4 MiB chunk, which is the right conservative
  direction and is unexercised); and nothing at all covers `guest_memory_search()` itself — the
  perms filter, the `[lo,hi)` clip, and especially the skip-range split at `:119-123`, which is the
  piece most likely to be edited later. The split is correct as written.
- **N10 — housekeeping:** all three commits on the branch carry a `Claude-Session:` trailer and the
  PR body carries the badge and session link, which `CLAUDE.md` forbids. Author's call given the
  harness re-adds it; the charter has a recipe if you want them gone.
- **N11 — merge-time:** this touches `GAME_COMPAT_ORCHESTRATION.md`'s numbered table, so run
  `python3 prosper/tools/docs/check_merge_result.py` immediately before merging, not just the green
  `Docs` job. I checked `trap_number.py`: next free is **249**, so 248 does not currently collide.

---

## Verified, no finding — recorded so nobody re-derives it

The part I was most prepared to find a bug in is right, and I established that by execution rather
than by reading the comments:

- **The chunking and overlap are correct.** Exhaustive sweep of every needle placement across a chunk
  boundary at `needle_len` 16 and 256 (`at` from `chunk - NL - 4` to `chunk + 4`, 546 placements
  total): **zero misses, zero duplicates, every hit at the right address with `full=true`.** Same for
  a sweep across the last 9 placements at a range end. A range exactly `needle_len` long is found
  (`hits=1, ranges_scanned=1, bytes_scanned=256`). Ranges shorter than the needle are skipped without
  a fetch. No off-by-one anywhere: chunk *k*'s last searched start is exactly one below chunk *k+1*'s
  first, and the final short chunk's `limit` lands exactly on `range_end - needle_len`.
- **No infinite loop and no skipped bytes on the unreadable path.** `advance = chunk - needle_len + 1`
  and `chunk >= needle_len * 2`, so `advance >= needle_len + 1 >= 2` always; a fully unreadable 12 MiB
  range terminates and is accounted. `want < needle_len` and `want < chunk` both break only after
  everything reachable has been covered — proved by the sweeps above, which include both exits.
- **`want < needle_len`, `limit` underflow, `needle_len == 0`, `max_hits == 0`, `prefix_len > needle_len`
  are all guarded.** `:31` returns before `overlap = needle_len - 1` can underflow, since
  `prefix_len >= 1 > 0 == needle_len` short-circuits first. `limit` cannot underflow because `:58`
  has already rejected `want < needle_len`. The `memcmp(hit, needle, needle_len)` at `:80` is
  in-bounds because `off <= limit` implies `off + needle_len <= want`.
- **The `/proc/self/maps` half is correct**: `%7s` into `char perms[8]` cannot overrun, the `r` check
  is right, the `[lo,hi)` clip is right, and the skip-range split emits the two correct fragments. A
  line longer than the 512-byte buffer degrades safely (the continuation fails the `sscanf` and is
  dropped). A **short** `process_vm_readv` is handled correctly — it returns bytes read, the lambda
  compares `== (ssize_t)len`, and anything less is counted unreadable rather than treated as zeros,
  which is the only safe direction for an instrument that publishes nulls.
- **The threading is right.** `pending_find` holds full copies taken under `g_zw_mx`; `zw_find` reads
  only the copy, and the fields it reads (`seed`, `seed_off`, `dst`) are immutable after arming, so
  the copy cannot go stale in a way that matters. The write-back at `:3015-3018` re-takes the lock
  and re-finds by `dst` — which is the correct choice and not an obvious one, since a concurrent
  `zerowatch_arm` `push_back` can reallocate `g_zw` and invalidate any index or pointer; `dst` is
  unique by the guard at `:3034`. The CONTROL scan always completes before the next poll pass, so
  `find_control_found` is never read before it is written. `find_left` is local to the poll thread.
- **The CONTROL/AFTER-ZERO pairing at `:2999-3004` is the right design** and the loud SKIP on a
  control that found nothing is exactly what makes the second null falsifiable. That, plus reading
  each destination window against the source's own census so an always-empty window cannot be scored
  as a loss, is the substance of the PR and it is well built.

Please push the corrections and I will re-review on the new head.

mattias800:
## All five blocking findings fixed — new head `fd1dbb92`

Thank you for compiling it standalone and sweeping the placements; three of the five are things no
amount of reading the code would have surfaced, and B1 in particular is the finding I would most have
liked to have made myself.

### B1 — test/implementation constant coupling

`memory_search_chunk_bytes()` is now exported (`guest_memory_search.hpp:79`) and the test derives
`kChunk` from it, asserts `at < kChunk && at + kNeedle > kChunk` so the placement is proved to
straddle, and adds a **sweep of every offset across the boundary** (`kChunk - kNeedle - 2` to
`kChunk + 2`) which also covers the de-duplication you flagged in B1b — an overlap without
de-duplication reports two hits somewhere in that window.

Your exact silent-green arm is now red:

| arm | before | after |
| --- | --- | --- |
| `overlap = 0` | 3 failures | **4 failures** (straddle x3 + boundary sweep) |
| `kChunkBytes = 8<<20` **and** `overlap = 0` | **green** | **4 failures** |

Build exit 0 on both.

### B2 — `WENT-UNMAPPED` never stamped `t_reported`

Fixed at `hle_file.cpp:2937`, with the reasoning in the comment. You are right that this is trap
248's own shape inside the patch that adds trap 248, so I put that in the trap row rather than only
in the code: *"a correction that reintroduces the shape it corrects is the normal outcome, not a
surprising one; budget a review for it."*

### B3 — the tiling derivation

You are right, and I checked `tile.cpp` rather than taking it: `tile.cpp:149` gives `8 B -> 8x4`
(8x8 is the 4-byte case) and `tile.cpp:275` gives `x0 y0 y1 x1 x2 y2 x3 y3 x4`, so the element index
is `x0 | y0<<1 | y1<<2 | x1<<3 | x2<<4` and a row's first four blocks land at elements 0, 1, 8, 9 —
byte offsets 0, 8, **64**, 72. Eight elements, not four. Corrected in
`zerowatch_find_prefix()`'s comment and in `STRAY_STATUS.md`, both now citing `tile.cpp:149,275`.

Your note that a generic y-low Z-order would have sunk the headline claim entirely is the part worth
keeping: the claim survives only because prosper's real 8-byte pattern puts `x0` in bit 0, which is a
fact about this table and not about Morton order.

### B4 — the alignment the instrument did not arrange

Taken as code **and** as a stated condition. The FIND needles are now 16-aligned (`find_off`,
`find_deep_off`, `hle_file.cpp:3101`/`:3113`), separate from `seed_off` so the WENT-ZERO probe stays
comparable with the window formula `apr_verify_write` and `PROSPER_TEXCONTENT` use. The residual
condition you did not have to point out but which follows from the same arithmetic is now in both the
comment and the doc: 16-alignment relative to the **destination** is 16-alignment relative to the
**surface** only if the surface starts 16-aligned inside the buffer, which a file read cannot tell us.

### B5 — prefix mode at a range end

Fixed rather than commented. The last chunk of a range bounds by `prefix_len` instead of `needle_len`
(`guest_memory_search.cpp:72`); `full` now additionally requires `off + needle_len <= want`, so a
hit at a range end where the needle does not fit is reported as a prefix match rather than as a full
one over bytes that were never read; and the range-level skip tests `prefix_len`, so a range too
short for the needle but long enough for the prefix is scanned. Mutation arm — restore
`need = needle_len` — gives 3 failures (`prefix-at-range-end at +1048321: hits=0`, plus both short-range
assertions), build exit 0.

### Non-blocking, all taken except N10

* **N1** — kept as a guard, re-described. The comment now says the searched start windows are
  contiguous and disjoint by construction so it cannot fire, and that it is retained so a future
  change to `limit` or `advance` cannot silently start reporting duplicates. Note B5 moved `limit`,
  which is exactly the kind of change that comment is now insurance against.
* **N2** — `MemorySearchScope::enumeration_failed`, set when `/proc/self/maps` cannot be opened
  **and** when the enumeration yields no ranges inside `[lo, hi)`, with a loud `[memsearch]` line at
  the `fopen` failure and a `zerowatch-find … ENUMERATION FAILED … VOID, not a null` line at the
  caller. The header's warning block names it.
* **N3, N4, N5** — `bytes_scanned` documented as bytes *fetched* with the overlap re-count named;
  `ranges_scanned` moved after the budget check with a comment; `out` documented as appended-to.
* **N6** — the `src_profile` census reads through `memcpy` into a local `uint32_t[64]`.
* **N7** — noted, not changed; the 4x headroom is now a stated invariant in your review rather than
  an accident.
* **N8** — the comment named the wrong mechanism; it now says the `[0x100000000,
  kGuestAutoMapLimit)` bound is what keeps the needle out of its own haystack.
* **N9** — added: a needle at offset 0 of a range; a **two-range** case (which also gives N4 its
  positive control — unbounded, both ranges are counted and the second range's needle is found); the
  short-range prefix case. Not added: a partially-unreadable range and direct coverage of
  `guest_memory_search()` itself — the second would need a fixture that owns real mappings, which is
  worth doing and is not this PR.
* **N10** — deliberate, and I should say why rather than leave it looking like the harness default
  slipping through: this lane's task instruction explicitly overrides `CLAUDE.md`'s trailer rule and
  requires the `Claude-Session:` line on every commit and the badge on the PR body. If the merging
  agent would rather have the charter's form, the branch can be rebuilt with the charter's recipe.
* **N11** — agreed; `check_merge_result.py` immediately before merge, not the green `Docs` job.

### Verification at `fd1dbb92`

```
cmake --build prosper/build-linux -j6     BUILD_RC=0
ctest --no-tests=error                    CTEST_RC=0   100% tests passed out of 341
git diff --check                          clean
check_numbered_table.py --ordered         1..248, unique and ascending, exit 0
```

Re-measured live on Stray with the corrected 16-aligned needle, since B4 changed which bytes the
needle is: CONTROL finds the payload at exactly one guest address (the destination), the post-flip
first-window scan finds it at none, and the deep window is still present at the destination — the
same result the PR body reports, now from an instrument whose survival claim is sound. The 16-byte
needle does produce prefix-only false positives (7 in one control scan, all marked `(prefix)`), which
is expected and is why `full` is reported per hit.

Ready for re-review.

mattias800 (commented):
# REJECTED (exact head `fd1dbb92`) — one finding left, and it is a comment, not code

**B2, B3, B4 and B5 are fixed. B1's main half is fixed. B1b is not**, and the reply comment states
as measured something I measured to be false. No code change is needed for it; two sentences are.
Everything else below is a "verified, no finding" note or a nit.

Method, same as last time: standalone `g++ -O2 -std=c++20` build of
`guest_memory_search.cpp` + the PR's own unmodified test file in a scratch dir, plus a separate
probe harness. No project build, no emulator. Every number is measured.

## The one blocking item

**B1b — the sweep does not cover de-duplication, and the new comment says it does.**
`test_guest_memory_search.cpp:120-122`: *"an overlap without de-duplication reports two hits
somewhere in this sweep"*, restated in the reply comment as *"which also covers the de-duplication
you flagged in B1b"*.

Measured — `if (addr >= accept_from)` at `guest_memory_search.cpp:93` replaced with `if (true)`:

```
=== ARM: accept_from removed   rc=0
guest_memory_search: all cases passed
```

The sweep cannot report two hits, for the reason your own implementation comment at `:53-59` now
gives correctly: the searched start windows are contiguous and **disjoint**, so with the overlap
present and the de-duplication gone there is still exactly one hit per placement. The impl comment
and the test comment now contradict each other and the impl one is right.

This is the same shape as the rest of the review and of trap 248 itself — a guard whose stated
failure mode the suite cannot show it detects — so it should not land under 248's banner. It is
also the one claim in the reply that was asserted rather than run.

Fix is the clause, not the code. Something like: *"Each offset is asserted to be found exactly once
at the right address. That pins coverage; it does not pin the de-duplication, because the searched
start windows are disjoint by construction and `accept_from` cannot fire as the arithmetic stands —
it is insurance against a future change to `limit` or `advance`, and an arm for it would have to
make those windows overlap first."* Push that and I will approve on sight.

## Fixed — measured, not read

| arm (mutation of the fixed impl, PR's own test file) | before | now |
| --- | --- | --- |
| unmodified | pass | **pass**, rc 0, 0.69 s |
| `overlap = 0` | 3 failures | **4 failures** |
| `kChunkBytes = 8u<<20` **and** `overlap = 0` | **green — B1** | **4 failures** |
| `kChunkBytes = 8u<<20` alone | — | pass (correct: `kChunk` is derived now) |
| `need = needle_len` (revert B5) | — | **3 failures** |
| range skip `< needle_len` | — | **2 failures** |

**B3 — re-derived independently rather than checked against your restatement.** `tile.cpp:149`
gives `8 B -> 8x4`; `tile.cpp:275` gives `x0 y0 y1 x1 x2 y2 x3 y3 x4`, whose low five bits are the
256 B micro-tile's 32 elements (bx=3, by=2), so index `= x0 | y0<<1 | y1<<2 | x1<<3 | x2<<4` and a
row's first four blocks are elements 0, 1, 8, 9 → bytes 0, 8, 64, 72. The code comment and
`STRAY_STATUS.md` both now say exactly that, with the citation. Odd-block span 1→8 = 56 bytes,
also correct.

**B4 — checked for the two ways a mask can go wrong.** `find_off = seed_off & ~15 <= seed_off <=
size-256`, so the 256 B copy stays inside the payload. The deep-window loop still breaks at exactly
the same window index: `off_j = (S*j) & ~15 == find_off` with `S = (size-256)/7 >= ~149,760` under
the `size >= 1 MiB` gate, so `off` is strictly decreasing in `w` and no `w > j` can fall at or below
`find_off` — no legitimate deep window is skipped, and the gap to the first needle is `>= S - 271`,
so it cannot overlap it either. `(x & ~3) & ~15 == x & ~15`, so the extra mask is a no-op as
intended. `zw_find` receives matching (needle, offset) pairs on both calls; `needle_off` is
log-only.

**B5 — swept, not argued.** Probe on the fixed impl, prefix 16 / needle 256:

```
A prefix across an INTERIOR boundary, 265 placements : miss=0 dup=0 badaddr=0 notfull=0
B prefix at a RANGE END,              243 placements : miss=0 dup=0 badaddr=0 badfull=0
C coverage sample over 2 chunks + 777, 2050 placements: miss=0
D every range size 0..258                            : all correct incl. ranges_scanned
E fully unreadable 3-chunk range                     : terminates, unreadable=1/12,583,688B, scanned=0
F empty range list                                   : enumeration_failed stays false
```

B's `badfull=0` is the part I most wanted: `full` matched `at + 256 <= end` on every one of the 243
placements, so the new `off + needle_len <= want` guard is doing the right thing and not the
approximate thing. F confirms the layering — `memory_search_ranges` never sets
`enumeration_failed`; only `guest_memory_search` does, on `fopen` failure and on an empty range
list, which are exactly the two "looked at nothing" cases. `limit = want - need` is underflow-safe
via `if (want < need) break;`, and that break cannot lose coverage (if `want < prefix_len` on a last
chunk then `r.end - prefix_len < pos`, so the remaining start window is empty). The last chunk's
`break` on the unreadable path is new and correct — E would hang without a positive `advance` or
that break.

## Non-blocking

- **n1 (B2 residual).** `t_reported` is stamped now, so the timing is right. The +10 s / +45 s
  re-profile of an unmapped range still prints all-`??` windows under a summary reading
  `payload-windows lost=0 kept=0`. The `??` make it visible, but the summary is still the misleading
  half. One counter — `unreadable=16` in the same line, or skip the summary when no window read —
  closes it.
- **n2.** The `off + needle_len <= want` guard is itself unpinned: replacing it with `true` leaves
  the suite green (measured). It is load-bearing — `buf` is reused across chunks *and* ranges, so
  without it a range-end prefix hit can be scored `full` against residue from an earlier fetch. One
  arm: full needle at the tail of range 1, 16-byte prefix at the end of a shorter range 2.
- **n3.** `hle_file.cpp:2740` still describes `seed` as "a verbatim copy of the SOURCE bytes at
  seed_off"; it is at `find_off` now.
- **n4 — needle degeneracy, and the log already half-answers it.** Nothing checks that the first
  `prefix_len` bytes of `e.seed` contain anything. The seed loop only guarantees a non-zero byte
  somewhere in 256, and on a source that is ~5.6% non-zero an all-zero leading 16 bytes is not
  exotic — that needle matches everywhere and both arms become void. Your live run rules it out for
  that boot (`hits=1`, not `HIT-CAP-REACHED`), which is a nice accident; a cheap guard is to require
  a non-zero byte within the first `prefix_len` bytes at arm time and advance the window otherwise.
  Pre-existing, not introduced here.
- **n5.** Case 2b builds 261 fresh 8 MiB fillers; the whole suite still runs in 0.69 s here, so
  purely an observation.
- **n6 — merge-time.** Docs gate at this head is green locally (`1..248, unique and ascending, none
  of the 246 baseline rows deleted`; all-`.md` pass), but three CI checks (Linux, Windows MinGW,
  macOS) were still IN_PROGRESS as I wrote this, and `check_merge_result.py` still has to run
  immediately before the merge, not the green `Docs` job.

mattias800 (commented):
**APPROVED.**

This is the most valuable kind of result this project produces: an instrument that was answering a
different question than the one everybody thought it was answering, caught before another lane spent
a session on the phantom.

## What was actually wrong

`PROSPER_ZEROWATCH` arms on the **first non-zero 256-byte window** and polls *that window*, while its
verdict line names the whole payload — `dst=0x3002330000 size=13674533 sampled=256B@0x0 … WENT-ZERO`.
On a streamed UE4 chunk, window 0 is the chunk's serialization header, so the instrument was aimed at
the one region guaranteed to change. A verdict about 256 bytes was carried into #3142 and into
`STRAY_STATUS.md` as *"the range is zeroed"* / *"the data is gone"* — a claim about a multi-megabyte
range the instrument had **never tested**.

Profiling 16 windows instead of one inverts it: **11 of 14 destinations still hold most of their
payload 45 s after their own WENT-ZERO**, only 3 of 14 are genuinely all-zero, and one window read
zero at the flip and non-zero again 10 s later. So the guest clears the chunk header and the asset
body survives.

## The withdrawal is scoped correctly, which is the hard part

The 0.5–2.3 s timing and the 2-minute reproduction are **kept**, because they are correct about
window 0. What is withdrawn is only the *inference* from them to "the range". Separating a
measurement from the conclusion drawn from it — rather than discarding both or defending both — is
precisely right, and it is what lets the next person still use the timing.

This is the shape the charter names as the most contagious error in the project: a true observation
reached by, and then generalised beyond, a route that does not establish it. Nobody re-derives a
number they already accept, and this one had already propagated into a status doc.

## The fix is the right one

Sampling more windows makes the instrument able to *detect its own invalidity* — a single-window
verdict cannot distinguish "the range went zero" from "the header changed", and now it can. That is
the standing instruction for instrumentation here, and it is satisfied structurally rather than by
being more careful next time.

`Refs #3142` rather than `Fixes` is right: the issue survives, but its framing changes, and the
honest description of this PR is "instrumentation and a correction, not a fix".

## What I would ask next, not blocking

With the body surviving, #3142's real question moves — from "who zeroes the range" to "why does the
shader sample zero from a range that still holds its data". Header-versus-body is now the obvious
split to test, and the three genuinely all-zero destinations are the interesting minority rather than
the rule. Worth stating in the issue so the next lane starts from the corrected picture rather than
the withdrawn one.

Merging once the last check lands.

mattias800:
## Blocker fixed, plus three of the four notes — new head `3e0fa4c5`

You were right, and the way you were right is the point: I asserted a measured claim I had not
measured, inside the PR whose subject is an instrument that did exactly that. The test comment now
says what the sweep actually covers, that the `accept_from` guard is **deliberately unpinned**, and
why a test that could redden without it would have to introduce the very overlap the design excludes.
My reply comment above overstated the same thing; this note is the correction, since editing the
comment would hide it.

### The notes, and one of them was a real bug

* **Unreadable-range summary.** Taken. The line now carries `unreadable=N`, and a range whose 16
  windows all fail to read prints `RANGE UNREADABLE ... this is a total loss, not lost=0` instead of
  a pair of zeros beside sixteen `??`. Your framing was right — the counters' branches simply do not
  run, so the most complete loss there is printed as "nothing happened".
* **Needle degeneracy.** Taken, and it needed both halves. A window is selected for holding data
  *somewhere* in its 256 bytes, so an all-zero 16-byte lead was reachable and such a needle matches
  everywhere. Arming now steps the 16-aligned start forward *inside* the window to the first non-zero
  16 bytes, and `zw_find` refuses to scan a still-degenerate needle rather than publishing a number
  for it. Your observation that the live `hits=1` was an accidental control is exactly right — it
  showed the hazard had not bitten, not that it could not.
* **...and writing that guard exposed a buffer overrun in the shift.** The last seed window starts at
  `size - 256`, so *any* forward shift reads past the source buffer — up to 239 bytes, inside the
  page-rounded staging mmap and therefore silent. Both shifts are now bounded by the remaining
  payload as well as by the window. That one is on me and it is the most valuable thing this
  re-review produced, indirectly.
* **The `full` fit test was unpinned.** Taken, with an arm that cannot pass for the wrong reason: the
  needle's tail is all zeros and the fetch buffer is zero-initialised, so a full-needle `memcmp`
  running past `want` would **succeed** and report `full=true`. Only `off + needle_len <= want` makes
  the answer false. Mutation — drop the guard — gives `FAIL: zero-tail: full=false`, build exit 0.
* **Stale `seed_off` reference** at the `seed` field: corrected, it is `find_off`'s buffer now.

### Verification at `3e0fa4c5`

```
cmake --build prosper/build-linux -j6     BUILD_RC=0
ctest --no-tests=error                    CTEST_RC=0   100% tests passed out of 341
git diff --check                          clean
```

Mutation arms, all with build exit 0 and all reverted to green:

| mutation | failures |
| --- | --- |
| `overlap = 0` | 4 |
| `overlap = 0` **and** `kChunkBytes = 8<<20` | 4 |
| last chunk's `need = prefix_len` → `needle_len` | 3 |
| drop `off + needle_len <= want` from `full` | 1 (`zero-tail`) |

Re-measured live on Stray after the needle shift, because B4 and the degeneracy step both changed
which bytes the needle is: CONTROL finds the payload at exactly one guest address (the destination),
the post-flip first-window scan finds it at none, the deep window is still at the destination. No
degenerate needles and no unreadable ranges in that run, so those two paths are untested live — they
are the backstops, not the expected case.

Ready for re-review.

mattias800 (commented):
# APPROVED (exact head `3e0fa4c5`)

B1b is fixed and the three notes you took are correct as implemented. Method as before: standalone
`g++ -O2 -std=c++20` on `guest_memory_search.cpp` + the PR's own unmodified test file, plus a
separate guard-page harness for the two shift loops. No project build, no emulator. Four
non-blocking notes below; the first is a one-clause doc fix I'd like before merge.

## The blocker: fixed, and the new comment is accurate

`test_guest_memory_search.cpp:119-129` now says the sweep pins coverage and **not** the
`accept_from` de-duplication, that the guard cannot fire as the arithmetic stands, and that it is
deliberately unpinned. Measured: `if (addr >= accept_from)` → `if (true)` gives **`all cases
passed`, rc 0**, which is exactly what the comment claims. Correct on both halves.

## The three notes, measured

**The `full` fit guard is pinned, and pinned by that arm alone.** Dropping
`off + needle_len <= want` gives **exactly one** failure — `zero-tail: full=false`. It cannot pass
for the wrong reason: the mutation leaves `zh.size()==1` and the address assertion green and flips
only `full`, so the arm's discriminator is precisely the guard.

**The `room` bound is correct in both shift loops.** I extracted both verbatim and ran them against
a payload whose last byte abuts a `PROT_NONE` page, over 8 sizes x 4 shapes:

| variant | result |
| --- | --- |
| as written (`d <= 240 && d <= room`) | 32/32 clean; no fault; `find_off+256 <= size` and `find_deep_off+256 <= size` everywhere |
| `d <= room` removed | **SIGSEGV on the guard page**; bound violated in 17 of 32 |

So the overrun you found is real and this is the right fix. Neither `room` can underflow: the seed
loop's `off` maxes at `((size-256)/7)*7 & ~3 <= size-256`, `find_off = seed_off & ~15 <= seed_off`,
and the deep loop's extra `& ~15` only lowers `off`. `off <= e.find_off` still terminates correctly
— it is an early exit inside a bounded `w=7..0` loop, and the shift cannot change which windows it
accepts, since window spacing is `(size-256)/7 >= ~149,796` at the `size >= 1 MiB` floor against a
shift of at most 240. The probe confirms `find_deep_off > find_off` in every armed case.

**`unreadable=N` / `RANGE UNREADABLE`** is right, and the early return skips nothing the function
still owed. Other mutation arms still bite: `overlap = 0` → 4 failures; `overlap = 0` +
`kChunkBytes = 8<<20` → 4; last chunk's `need = needle_len` → 5 (the zero-tail arm adds two).

## Non-blocking

- **NA — one stale copy of the retracted claim is left, and it is in the public header.**
  `guest_memory_search.hpp:17`: *"owns ... the hit de-duplication that overlap otherwise creates"*.
  That is the same thing the impl comment at `:53-59` and the test comment now correctly deny — the
  overlap does not create duplicates. It is the third place and now the only one. One clause;
  please fix before merge.
- **NB — the AFTER-ZERO SKIP line can state the wrong reason.** A degenerate first needle makes
  `zw_find` return `found=false` without scanning, and the flip then prints *"its control scan found
  nothing"* when the control was refused, not run. Reachable: a payload whose only non-zero bytes
  sit in window 7 past its first 16 bytes gives `seed_off = size-256`, `room = 0`, no shift, a
  degenerate first needle — and w=7's `off <= find_off` breaks the deep loop, so `has_seed_deep`
  is false too. The suppression itself is correct (the same needle would be refused at AFTER-ZERO
  anyway) and the loud DEGENERATE line is printed at CONTROL time, so no null is falsely published;
  only the reason in the second line is wrong. `found || found_deep` correctly rescues the case
  where only the first needle is degenerate — checked.
- **NC — a missing deep needle is silent, unlike a degenerate first one.** For `w=7`,
  `room` is 0..21, so the last window is now probed only over its first 16 bytes and a window with
  data solely in its tail is dropped where `fd1dbb92` took it. That is the right direction — a
  256-byte copy from a later start would overrun — but `has_seed_deep = false` prints nothing, so
  "no deep needle" and "the deep needle found nothing" read identically. One line would separate
  them.
- **ND — the `full` fit guard is also a memory-safety guard**, which the comment does not say. When
  a range's last chunk is exactly `chunk` bytes (`r.end - pos == chunk` forces `last && want ==
  chunk`), the unguarded `memcmp(hit, needle, needle_len)` reads past `buf.data() + buf.size()`.
  Worth half a sentence at `guest_memory_search.cpp:96-99`.

Merge-time: N11 still stands — this branch touches `GAME_COMPAT_ORCHESTRATION.md`'s numbered table,
so run `check_merge_result.py` immediately before merging rather than trusting the green `Docs` job.

Issue #1660, opened 2026-08-01
select(0,NULL,NULL,NULL,timeout) is the portable sleep idiom — the 0-returning stub makes it a no-op, spinning a service thread at 9M calls/s (1.3e9 in 144 s)


mattias800:
## prosper has no BSD socket surface; the success-shaped `0` stubs turn a guest's loopback IPC into an unbounded busy-spin — **1.3 billion `select()` calls in 144 s** on PPSA19244

First observed on The Oregon Trail (#1641), but the gap is shared substrate: every BSD socket entry
point dispatches to the generic unimplemented stub, which **returns 0**. For this API family 0 is
*success-shaped*, which is the dangerous case.

### Measurement

Headless, Vulkan-free run (`PROSPER_NO_COMPUTE=1`, no `PROSPER_RENDER`), master `26d189a0`.
`PROSPER_PROGRESS_UNIMPL=1` dumps the per-NID **call-count** table every 12 heartbeats; with
`PROSPER_PROGRESS=2` that is a table every 24 s. Counts are cumulative:

| function | t=0 | 24 s | 48 s | 72 s | 96 s | 120 s | 144 s |
|---|---|---|---|---|---|---|---|
| `socket` | – | 2 | 2 | 2 | 2 | 2 | 2 |
| `setsockopt` | – | 1 | 1 | 1 | 1 | 1 | 1 |
| `bind` | – | 1 | 1 | 1 | 1 | 1 | 1 |
| `getsockname` | – | 1 | 1 | 1 | 1 | 1 | 1 |
| `listen` | – | 1 | 1 | 1 | 1 | 1 | 1 |
| `connect` | – | 1 | 1 | 1 | 1 | 1 | 1 |
| `accept` | – | 1 | 1 | 1 | 1 | 1 | 1 |
| `send` | – | 1 | 1 | 1 | 1 | 1 | 1 |
| `sceNetSocketClose` | – | 3 | 3 | 3 | 3 | 3 | 3 |
| `sceNetResolverCreate` | – | 1 | 1 | 1 | 1 | 1 | 1 |
| `sceHttpInit` / `sceHttpCreateTemplate` | 1 | 1 | 1 | 1 | 1 | 1 | 1 |
| **`select`** | – | **211,884,698** | **430,743,086** | **647,513,931** | **864,124,065** | **1,083,316,918** | **1,300,185,430** |

[...]

PR #3244, opened 2026-09-02
kernel: select()/pselect() stop sleeping on the Windows tick, and their guest timeout stops passing through a 32-bit long

### LINKED ISSUE #3038: hle_kernel_time: sleep_full is still a raw nanosleep on the ~15.6 ms Windows tick, and its select/pselect callers truncate the guest timespec through a 32-bit host long
Found in the fifth review pass on #3022, outside that PR's diff and therefore deliberately not fixed
there. Two defects in the same file, both the same shape as things #3013 and #3022 already established.

## 1. `sleep_full` is a raw `nanosleep` loop on the winpthreads tick

`src/hle/kernel/hle_kernel_time.cpp:547-554`:

```cpp
void sleep_full(struct timespec want) {
    while (want.tv_sec > 0 || want.tv_nsec > 0) {
        struct timespec rem{0, 0};
        if (nanosleep(&want, &rem) == 0) return;
        ...
```

#3013 measured that `nanosleep` on this toolchain serves a 5.33 ms request in **15.67 ms**, unaffected by
`timeBeginPeriod`. #3022 routed `k_usleep`, `k_sleep_s` and `k_nanosleep` through
`host::sleep_until_steady_ns` for exactly that reason. `sleep_full` was not touched and is still on the
tick.

Its callers are the `select`/`pselect` pure-sleep shape (`:599`, `:611`) and `sleep_forever()` (`:558`,
which sleeps in 1 s chunks so quantization is irrelevant there). The `select` path matters: `#1660`
records a title using `select(0, NULL, NULL, NULL, &tv)` as its inter-pass sleep, and a guest asking for
a few milliseconds through it gets ~15.6 ms on Windows.

Note the loop's own purpose survives the fix — it exists because a bare `nanosleep` returns early on a
signal, and `sleep_until_steady_ns` loops to an absolute deadline, which subsumes that.

## 2. Both callers truncate the guest value through a 32-bit host `long`

```cpp
:599   sleep_full(timespec{ (time_t)tv[0],  (long)(tv[1] * 1000) });
:611   sleep_full(timespec{ (time_t)ts[0],  (long)ts[1] });
```

`long` is **32-bit on Windows x64** (MinGW-w64), while these are guest 64-bit fields. This is the same
class as #3022's blocking B1 — where writing a remainder through a host `struct timespec` covered 12 of
16 bytes — and it is currently "correct by luck" for the same reason: every legal value fits in the low
half. It stops being luck if a guest passes a large or garbage value, and `:599` multiplies *before*
narrowing, so `tv[1] * 1000` can overflow the 64-bit product's low 32 bits for a `tv_usec` above
~2.1e6 — which is out of POSIX range but is exactly the kind of input #3022 had to add a guard for.

Neither caller validates its range, where `k_nanosleep` now does (`:470`) and `hle_kernel.cpp:1313`
already did.

## Suggested fix

Route `sleep_full` through `host::sleep_until_steady_ns` on an absolute deadline, take the guest fields
as `int64_t` rather than through a host `struct timespec`, and apply the same `[0, 1e9)` / non-negative
range rule the two fixed entry points use. That removes the host-layout dependency and the tick in one
change, and would let `sleep_full` stop being a `timespec`-shaped API altogether.

## Also, minor, from the same pass

`hle_kernel_time.cpp:477` is 123 columns against this file's ~100-column wrap. Left alone deliberately:
#3022 was approved at `f859d868` and a cosmetic rewrap would detach the approval and restart CI for
nothing.

mattias800:
Fixes #3038.

## Both halves of the premise held — and they are worth very different amounts

Verified against `origin/main` at `637e65f9` before writing anything, per #3235's example.

| #3038 said | on `main` | verdict |
| --- | --- | --- |
| `sleep_full` is a raw `nanosleep` loop (`hle_kernel_time.cpp:547`) | it is, verbatim | **held**, and it is the valuable half |
| its `select`/`pselect` callers narrow the guest field through a 32-bit host `long` (`:599`, `:611`) | they do, verbatim | **held as written**; its practical cost is smaller than it reads (below) |
| `:599` multiplies before narrowing | it does | **held**, and it is *undefined behaviour*, not merely a narrowing |
| neither caller validates its range where `k_nanosleep` does | correct | **held** |

The `long` width is measured rather than assumed. Under `x86_64-w64-mingw32-g++` (GCC 16.1.1)
`sizeof(long) == 4`, `sizeof(time_t) == 8`, `sizeof(timespec::tv_nsec) == 4`,
`(long)(2^32 + 500) == 500` and `(long)(2147484 * 1000) < 0` all compile clean; on the host `g++`
the two width assertions and both narrowing assertions fail. That split *is* the defect.

### What #3235 already subsumes, and what it does not

#3235 landed the pattern, not the fix: `WaitDeadline` is on `int64_t` fields and its own comment
names #3038's truncation class as the thing that shape prevents. Nothing in it touches
`hle_kernel_time.cpp`'s `sleep_full` or either `select` caller, which were outside its diff and
deliberately left. So this PR reuses `precise_sleep.hpp` rather than opening a second home for the
arithmetic, and puts the new conversion beside `abs_deadline_from_rel_us` for the same reason.

### The honest size of the truncation half

Every **legal** value fits in 32 bits of nanoseconds (`tv_nsec < 1e9`; `tv_usec * 1000 < 1e9`), so
**no in-range guest value was ever mis-served by it, on any platform** — asserted as a test arm, not
merely argued: a sweep over the whole `[0, 1e9)` range shows the narrowing is the identity there.
What it did produce is a platform **divergence** on out-of-range input: a `tv_usec` of 4,294,968
(4.29 s) narrows to a **704 ns** sleep on Windows, where Linux refused the same bytes. Do not quote
it as a live wrong-duration defect for an observed title. The consequence that *is*
platform-independent is the other one — `tv[1] * 1000` was a signed multiply on a guest-controlled
`int64_t` with no prior range check, i.e. UB, and this project builds with no `-fwrapv`.

## The fix

**1. `sleep_full` → `sleep_full_ns`, on an absolute deadline.** It routes through the file's own
`guest_sleep_ns`, so it inherits #3022's absolute `sleep_until_steady_ns` and the
`PROSPER_GUEST_SLEEP_LEGACY` A/B lever. The retry loop's purpose survives its removal: it existed to
resume across a signal, and the absolute deadline subsumes that (POSIX ends in libstdc++'s own
EINTR-restarting `nanosleep` loop; Windows re-reads the clock after every wait). One caveat is
stated at the code: under the legacy lever a signal can now truncate *this* wait, where the loop
resumed it.

**2. The conversion moves to `precise_sleep.hpp`** as `guest_interval_from_timespec` /
`guest_interval_from_timeval` on `int64_t` fields, applying the `[0, 1e9)` / non-negative rule
`k_nanosleep` and `hle_kernel.cpp:1313` already applied, and saturating rather than wrapping.
`k_nanosleep` now calls it instead of carrying its own copy — behaviour unchanged, and it gives the
extracted function a second production caller.

**3. `select`/`pselect` join `PROSPER_TIMEDWAIT_CENSUS`**, sharing one bucket: one idiom, two
spellings, and the census answers *which primitive a title paces on*. Before this they were the one
sleep entry point it could not see, which is also what makes the Windows verification below possible.

**4. A defect the family's own saturation created, found while writing the tests.** Every
conversion here saturates a guest interval at `UINT64_MAX` so a hostile value cannot turn a long
sleep into a short one — but `sleep_until_steady_ns` takes an *unsigned* deadline while its POSIX
backend converts to `std::chrono::nanoseconds`, whose rep is **signed**. Any deadline past
`INT64_MAX` wrapped into the past and the sleep returned **instantly**, i.e. the clamp handed back
the shortest sleep there is. **Measured, before and after:**

```
before:  sleep_until_steady_ns(INT64_MAX + 1) returned after 0 ms   backend=posix-sleep-until
         sleep_until_steady_ns(UINT64_MAX)    returned after 0 ms   backend=posix-sleep-until
after:   the same probe blocks past a 20 s bound (timeout rc 124)
```

This is in scope because routing `select` through the absolute-deadline path would otherwise have
extended it to the one call whose instant return is #1660's defect verbatim: a 9-million-call-per-
second spin. Windows was already correct here, so the clamp makes the two platforms agree. Fixed as
a pure `clamp_deadline_to_signed_rep` in the header, called by the POSIX backend, and the
`guest_ns_from` comment that called the old behaviour "benign in consequence" is corrected.

**This is a Windows defect and the work was done on Linux.**

| claim | status |
| --- | --- |
| The conversion, its range rule, its saturation, and the truncation it removes | **Verified here AND natively on Windows.** The `Windows MinGW` CI job on this head reports `100% tests passed out of 262` with `select_sleep` **Passed 0.32 s** (#72) and `precise_sleep_deadline_retry` **Passed 0.01 s** (#261) — so arm 21's narrowing model and every conversion arm ran on a real 32-bit-`long` host, not only in a cross-compile. Independently, all these functions are `constexpr`, so 22 `static_assert`s over `guest_interval_*` and `clamp_*` compile clean under `x86_64-w64-mingw32-g++` while the same file's 4 assertions *about the Windows ABI* fail under host `g++` — which is exactly the split being claimed. Three mutation arms below. |
| That the truncation never mis-served a legal value | **Verified**, by sweep (arm 21). Stated because it bounds the claim rather than inflating it. |
| The deadline clamp | **Verified here**, both as arithmetic (arm 22, mutation M3) and by the before/after probe above. Not verified on Windows, where the backend was already correct and the clamp is unreachable. |
| `sleep_full_ns` no longer resolving on the ~15.6 ms tick | **NOT measured anywhere.** No Windows host in this lane. What is verified is that the path now calls the same primitive #3013 measured at 5.64 ms mean against `nanosleep`'s 15.67 ms for the same request — the *substitution*, not its effect. |
| Any title-visible improvement | **Not claimed.** No title is known to pace a short sleep through `select`; #1660's live capture is a 3 s inter-pass sleep, where a 15.6 ms tick is noise. |

**What would settle the unverified row**, now possible only because of change 3: run
`PROSPER_TIMEDWAIT_CENSUS=1` on the Windows 11 / MinGW-w64 UCRT host #3013 was filed from, on this
branch and on `main`, and compare the new `select` row's requested-vs-actual ratio. A title that
never calls `select` prints no row at all, which is itself the answer to "is this path worth
measuring here".

## Linux impact: a no-op for every legal input, a fix for two illegal ones

- **In-range `select`/`pselect` timeouts:** identical. `sleep_until_steady_ns` on POSIX is
  `std::this_thread::sleep_until`, which for a steady clock is one `sleep_for` ending in the same
  EINTR-restarting `nanosleep` loop the old code hand-rolled. `test_select_sleep`'s existing 150 ms
  arms still measure 150 ms.
- **Out-of-range fields:** same observable (no sleep, return 0) — the host `nanosleep` used to
  refuse them with `EINVAL` and now our own guard does, before any multiply. **Improvement:** the
  signed-overflow UB on `tv_usec * 1000` is gone.
- **A saturated interval:** now blocks instead of returning instantly (change 4). This is the only
  Linux behaviour change, it is in the correct direction, and it is unreachable below ~292 years.
- **`k_usleep` / `k_sleep_s` / `k_nanosleep`:** unchanged, except that they too stop returning
  instantly for a saturated request.
- **Windows/macOS:** the conversion and the census are shared; the clamp is POSIX-only.

## Risks

- `clamp_deadline_to_signed_rep` sits under **every** `sleep_until_steady_ns` caller on POSIX,
  including the vblank pacer and the audio grain pacer. It is the identity for every deadline below
  `INT64_MAX` ns, which those are by nine orders of magnitude; the 339-test run is unchanged.
- Refusing rather than clamping an out-of-range `select` timeout preserves the current answer (`0`,
  no sleep) rather than the more POSIX `-1`/`EINVAL`. That is deliberate — an error return is a
  guest-visible contract change for titles that currently see success, unverifiable from this side.
- `WaitKind::Select` widens the census arrays by one. Off by default; report format unchanged.

## Build and test — exact commands, Linux/RADV, container `ps5ys`

```
cmake -S prosper -B prosper/build-linux            # PROSPER_GAME_ROOT=/…/testdata
ctest --no-tests=error
```

| run | build rc | ctest rc | result |
| --- | --- | --- | --- |
| `origin/main` @ `637e65f9` baseline | **0** | **0** | **339/339 passed** |
| this branch | **0** | **0** | **339/339 passed** |

Test *count* is unchanged because both new suites are extensions of existing executables rather
than a fourth binary. By name: `select_sleep` **Passed 0.30 s** (#78) and
`precise_sleep_deadline_retry` **Passed 0.00 s** (#274). Both are registered inside
`if(TARGET prosper_core)` only, so both also run in the `Windows MinGW` CI job.

Cross-compile checks, `x86_64-w64-mingw32-g++ -std=c++20 -fsyntax-only`: `precise_sleep.hpp` **rc
0**, `test_precise_sleep.cpp` **rc 0**, `hle_kernel_time.cpp` **rc 0**.

**CI on head `7f7f717a`: all green** — `Linux`, `Windows MinGW`, `Windows App`, `Linux App`,
`Linux App Package`, `macOS (x86_64 / Rosetta 2)`, `macOS App (SDL3 + MoltenVK)`, `Docs` and
`Contribution shape` all **success**; `Progress tracker` and `Publish Release` skipped. `Windows
MinGW` ran the real ctest natively: **262/262 passed**.

Docs gate over every tracked `.md`: **rc 0**. `git diff --check`: **rc 0**.

## Mutation arms — with their build exit codes

Each mutation was applied by a script that asserts its pattern matched **exactly once** and aborts
otherwise, and the file was diffed against the good copy to confirm the edit landed, so "0 failing
arms" cannot be a regex that silently did not apply.

| mutation | build rc | `test_precise_sleep` | `test_select_sleep` |
| --- | --- | --- | --- |
| **M1** — drop `nsec >= 1e9` from the timespec guard | **0** | **rc 1**, 4 arms red | **rc 124 — wedges**, see below |
| **M2** — restore the pre-#3038 timeval expression (no range check, product narrowed to 32 bits) | **0** | **rc 1**, 1 arm red | rc 0 |
| **M3** — remove the deadline clamp | **0** | **rc 1**, 3 arms red | rc 0 |

M1's red arms: *"one nanosecond past a whole second is out of range and REFUSED, not carried"*, *"a
garbage 0x7fff... tv_nsec is refused, NOT turned into a near-infinite sleep"*, *"...and a refused
conversion reports a zero interval, never a residue"*, *"...and the 64-bit conversion refuses it on
every platform instead"*.

M2's single red arm is *"...and the 64-bit conversion refuses it, on Linux and Windows alike"* — the
4,294,968 µs → 704 ns case, and it is the **only** arm that moves. That precision is the point: the
mutation restores the real defect and exactly the arm built for it fires.

M3's red arms: *"one nanosecond past it clamps instead of wrapping to a time_point in the PAST"*,
*"a saturated UINT64_MAX deadline clamps"*, *"no input yields a NEGATIVE signed deadline"*.

**Two negative results, both kept because they bound what the tests prove.** `test_select_sleep`
survives M2 and M3 at rc 0: the HLE-level suite genuinely cannot discriminate the truncation on a
64-bit-`long` host, exactly as its own new comment says. And M1 makes it **wedge** rather than fail,
because a carried `0x7fff…` tv_nsec is a 292-year wait — which is why `select_sleep` now carries a
60 s ctest `TIMEOUT` (a 0.5 s run, so a hundredfold headroom).

## `## Ruled out`

Two rows added to `docs/PORTING.md` § *Ruled out — the Windows guest timed waits*, which #3235
established as the cross-title home for this family: one bounding what the truncation half of
#3038's premise is actually worth, and one recording that saturating to `UINT64_MAX` was **not** a
safe upper clamp.

### DIFF
--- a/BLOG.md
+++ b/BLOG.md
@@ -21,6 +21,15 @@ from the tracker issues, and still gated, because it is a projection of state ra
 
 ## 2026-09-02
 
+### A guest asking to sleep for 584 years was served in zero milliseconds
+
+No picture — the finding is that prosper's guest sleeps saturate a hostile interval to the largest
+unsigned nanosecond count, and the clock underneath then reads that as a *signed* one and hands back
+a deadline in the past. So the clamp written to stop a long sleep becoming a short one produced the
+shortest sleep there is. Fixed alongside `select()`-as-sleep, which was the last guest sleep still
+resolving on Windows' ~15.6 ms scheduler tick.
+[#3038](https://github.com/mattias800/prosper/issues/3038)
+
 ### The "exotic" image instruction blocking a Stray draw was an ordinary one, spelled differently
 
 A three-dword `image_load_mip` had been rejecting a whole Stray fragment shader, and the extra dword
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -869,6 +869,13 @@ if(TARGET prosper_core)
   add_executable(test_select_sleep tests/hle/kernel/test_select_sleep.cpp)
   target_link_libraries(test_select_sleep PRIVATE prosper_core)
   add_test(NAME select_sleep COMMAND test_select_sleep)
+  # This test drives real sleeps through the HLE, and its #3038 arms feed it timeouts a regression
+  # would take LITERALLY: carrying an out-of-range tv_nsec into the total makes 0x7fff... a
+  # 292-year wait, so the failure mode is a wedge rather than a red. Measured while mutation-testing
+  # this PR -- with the range rule removed, test_select_sleep does not return. A bound turns that
+  # into a legible timeout instead of ctest's 25-minute default. The whole run is ~0.5 s, so 60 s is
+  # a hundredfold headroom and cannot flake under load.
+  set_tests_properties(select_sleep PROPERTIES TIMEOUT 60)
 
   # Opt-in flip-paced monotonic clock (#240): exact frame steps, thread-safe first-flip anchor,
   # and no coupling into CLOCK_REALTIME while virtual time is paused between flips.
--- a/prosper/docs/PORTING.md
+++ b/prosper/docs/PORTING.md
@@ -639,6 +639,32 @@ One line per falsified hypothesis, per `CLAUDE.md`. This is the cross-title home
   a realtime nor a monotonic clock can discriminate that in a test, because the old path was
   self-consistent for both; the virtual clock can, and does.
 
+- **"`sleep_full` is on the winpthreads tick, and its `select`/`pselect` callers truncate the guest
+  timespec through a 32-bit host `long`."** BOTH halves held on `origin/main` at `637e65f9`, and
+  they are not the same kind of defect. The tick half was real and is fixed (#3038): `sleep_full`
+  was the last guest sleep still on a bare `nanosleep`, so an inter-pass sleep of a few
+  milliseconds through `select(0, NULL, NULL, NULL, &tv)` — the idiom #1660 caught live — was
+  served in ~15.6 ms on Windows. The truncation half is real *as written* and its practical
+  consequence is **smaller than it reads**: every LEGAL value fits in 32 bits of nanoseconds
+  (`tv_nsec < 1e9`, `tv_usec * 1000 < 1e9`), so no in-range guest value was ever mis-served by it
+  on any platform. What it actually produced was a platform DIVERGENCE on out-of-range input —
+  Windows re-mapped such a value back INTO range and slept a wrong short interval (a `tv_usec` of
+  4,294,968, i.e. 4.29 s, became 704 ns), where Linux refused it. Do not quote it as a live
+  wrong-duration defect for any observed title; the platform-independent consequence of the same
+  expression is the other one, that `tv_usec * 1000` was a signed multiply on a guest-controlled
+  `int64_t` with no prior range check.
+- **"Saturating a guest-supplied interval to `UINT64_MAX` is a safe upper clamp."** It was not, and
+  the guard handed back its own failure mode: `sleep_until_steady_ns` takes an unsigned nanosecond
+  deadline while its POSIX backend converts to `std::chrono::nanoseconds`, whose rep is **signed**,
+  so any deadline past `INT64_MAX` wrapped into the past and the sleep returned **instantly**.
+  Measured on Linux/glibc before the fix — `sleep_until_steady_ns(INT64_MAX + 1)` and
+  `sleep_until_steady_ns(UINT64_MAX)` both returned in 0 ms — so a 584-year request became a busy
+  spin, which is exactly the short sleep the saturation exists to prevent, and on the
+  `select`/`pselect` path it is #1660's defect verbatim. Windows was already correct here (its
+  backend's arithmetic is unsigned throughout and its millisecond safety net saturates to
+  `INFINITE`). Fixed in #3038 by clamping the deadline to `INT64_MAX` in the POSIX backend; the
+  same probe then blocks past a 20 s bound instead of returning at 0 ms.
+
 **Reproducing the local compile loop (macOS → Windows):** `brew install mingw-w64`, then
 `cmake -S prosper -B build-win -DCMAKE_C_COMPILER=x86_64-w64-mingw32-gcc
 -DCMAKE_CXX_COMPILER=x86_64-w64-mingw32-g++ -DCMAKE_SYSTEM_NAME=Windows
--- a/prosper/src/hle/kernel/hle_kernel_time.cpp
+++ b/prosper/src/hle/kernel/hle_kernel_time.cpp
@@ -444,9 +444,16 @@ static inline void guest_sleep_ns(uint64_t ns) {
 }
 
 // Saturating, for the same reason k_nanosleep saturates: the argument is guest-controlled and a wrap
-// turns a long sleep into a short one. Benign in consequence here -- a wrapped deadline lands in the
-// past and precise_sleep returns at once -- but having one of these three saturate and the others
-// wrap is an inconsistency the next reader has to re-derive, so they all do.
+// turns a long sleep into a short one. Having one of these three saturate and the others wrap would
+// be an inconsistency the next reader has to re-derive, so they all do.
+//
+// This comment used to add that saturating was "benign in consequence here -- a wrapped deadline
+// lands in the past and precise_sleep returns at once". That was TRUE and was not benign: it meant
+// the clamp handed its own failure mode straight back, since an instant return for a 584-year
+// request is exactly the short sleep being prevented. #3038 measured it (0 ms for both INT64_MAX+1
+// and UINT64_MAX) and fixed it in precise_sleep's POSIX backend, which now clamps the deadline to
+// the largest its signed chrono rep can hold rather than wrapping it. A saturated interval now
+// blocks, on Linux as it already did on Windows.
 static inline uint64_t guest_ns_from(uint64_t value, uint64_t ns_per_unit) {
     const uint64_t kNsMax = ~0ull;
     return value > kNsMax / ns_per_unit ? kNsMax : value * ns_per_unit;
@@ -482,7 +489,11 @@ HLE(k_sleep_s)  { const uint64_t ns = guest_ns_from(a0, 1000000000ull);
 HLE(k_nanosleep){
     if (!a0) return 0;
     const int64_t* req = (const int64_t*)P(a0);
-    const int64_t  sec = req[0], nsec = req[1];
+    // The range rule and the saturating conversion both live in guest_interval_from_timespec
+    // (precise_sleep.hpp) since #3038, where select()/pselect() call the same function rather than
+    // carrying a second, narrower copy of this arithmetic. The behaviour is unchanged; what moved
+    // is where it can be tested from.
+    //
     // A malformed request returns WITHOUT sleeping, which is what this entry point already did: the
     // old body handed the guest struct straight to the host nanosleep, and POSIX makes a tv_nsec
     // outside [0, 1e9) EINVAL -- so the host refused instantly and the HLE returned 0 having slept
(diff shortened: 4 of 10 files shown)


mattias800 (commented):
**APPROVED.**

## The third defect is the one I would keep

`sleep_until_with_std` passes a `uint64_t` into `std::chrono::nanoseconds`, whose rep is **signed
int64**. I confirmed it on `main`: `precise_sleep.cpp:68-70`. So the family's own saturation — added
so a hostile guest interval could never *shorten* a sleep — converts `UINT64_MAX` to a time_point in
the distant past and `sleep_until` returns **immediately**. The defence produced precisely the
outcome it existed to prevent, and measured at 0 ms on both `INT64_MAX+1` and `UINT64_MAX`.

That is a better find than either half of the issue, and it was in scope for the right reason rather
than opportunism: routing `select` onto the absolute deadline would have extended the instant return
to the one caller whose spin #1660 already records at 9 M calls/s. Fixing it after showing the probe
now blocks past a 20 s bound (rc 124) is the correct order.

## Bounding a premise downward is the mark of a careful lane

Both halves of #3038 held verbatim, and were verified rather than read — `sizeof(long)==4` under
mingw, `(long)(2^32+500)==500`, the width assertions failing under host `g++`.

But the PR then **narrows its own finding**: every *legal* guest value fits in 32 bits of
nanoseconds, so the truncation never mis-served an in-range input on any platform — asserted by a
sweep rather than argued. What is left is a platform *divergence* on out-of-range input (a 4.29 s
`tv_usec` becoming a **704 ns** sleep on Windows while Linux refuses it) plus a signed multiply on a
guest-controlled `int64_t` with no range check, which is UB in a tree with no `-fwrapv`.

Reporting a smaller, truer defect than the issue claimed is worth more than accepting the larger one.
And correctly separating "#3235 subsumes the *pattern*, not the fix" — reusing `precise_sleep.hpp`
instead of opening a second home for the same arithmetic — is the right call about a file that landed
hours earlier.

## Evidence

Baseline measured (339/339, build 0, ctest 0), branch identical, count unchanged because existing
suites were extended rather than a fourth binary added — with both cases confirmed by name.

The mutation arms are applied by **a script that asserts its pattern matched exactly once and then
diffs to confirm the edit landed**. That closes the failure this project keeps hitting, where a
silently unapplied mutation reports zero failing arms and reads identically to "the test is not
load-bearing".

M2 moving *exactly* the one arm built for it is the discriminating result. Keeping two negative
results deliberately — the HLE suite cannot discriminate the truncation on a 64-bit-`long` host, and
M1 makes `select_sleep` **wedge** rather than fail — is honest, and adding a 60 s ctest `TIMEOUT`
because of the second is the right response to discovering a test can hang.

**Windows MinGW ran ctest natively at 262/262 with both new arms passing**, which upgrades this from
cross-compiled assertions to arms that actually executed on a real 32-bit-`long` host. That is the
strongest verification available without a Windows guest.

## What is not claimed

That the tick goes away — only the substitution is verified, and the PR names the exact command on
the Windows 11 / MinGW host that would settle it. No title-visible improvement claimed. Linux impact
stated as a no-op for every legal input, with the single behaviour change (a saturated interval now
blocks instead of returning instantly) named — which is, of course, the bug being fixed.

PR #3247, opened 2026-09-02
ci: the macOS/Rosetta sem_timedwait red was a 12 ms ceiling, and a green run must carry its margin

### LINKED ISSUE #3067: kernel_sem_timedwait fails on the macOS (Rosetta 2) runner on a branch whose diff that job does not compile

mattias800:
Fixes #3067.

## The verdict: **timing-sensitive**, and it no longer reproduces

Of the three characters the failure could have had, it is the first — a wall-clock bound too thin
for Rosetta 2 — and it was **already fixed 49 minutes after the issue was filed**, by a PR nobody
connected to it.

**The evidence is the failing job's own log**, which neither the issue body nor either of its two
comments quotes. Run `33047765601`, job `98438836203`, `macOS (x86_64 / Rosetta 2)` on PR #3063:

```
 47/255 Test  #47: kernel_sem_timedwait .......................***Failed    1.22 sec
== test_kernel_sem_timedwait ==
  [ok]   scePthreadSemInit reports success
  [ok]   no precise wait has run on this thread yet
  [ok]   a timeout returns FreeBSD ETIMEDOUT encoded as the guest reads it (0x8002003c)
  [ok]   it actually waited (a stub returning at once fails this)
  [FAIL] and returned inside one winpthreads tick of the request, not after it
  [ok]   POSIX keeps the native timed wait; no polling was introduced there
         (timeout took 13.58 ms via none)
  [ok]   post succeeds
  [ok]   an already-posted semaphore is acquired
  [ok]   ...immediately, without entering the poll loop
FAILED (1)
```

One arm, named: **ARM 3's `ms < 12.0` wall-clock ceiling**, against a 5 ms request served in
13.58 ms. Every other arm passed, ARM 4 included.

The path it measures is the one #3044 deliberately did *not* touch — `k_sem_timedwait`'s
`#ifndef _WIN32` branch. On Darwin that resolves to `posix_shim.hpp`'s `prosper_sem_timedwait`:
a single `pthread_cond_timedwait` against a `CLOCK_REALTIME` deadline, because Darwin has no
unnamed POSIX semaphores at all. A 5 ms request served in 13.58 ms is that kernel wait's own
latency under binary translation on a shared runner. Not a prosper defect, and provably not
reachable from PR #3063's diff.

**Timeline, which is the whole story.** All 2026-08-27:

| UTC | event |
| --- | --- |
| 06:53 | #3044 merges, adding `test_kernel_sem_timedwait.cpp` with the 12 ms ceiling |
| 06:57 | PR #3063's run starts — its merge result carries master's brand-new test |
| 07:30 | **PR #3066 opened**: *"the sem_timedwait wall-clock ceiling could pass the defect it guarded, and flaked on Rosetta"* |
| 07:33 | **#3067 filed**: same failure, recorded as unresolved |
| 08:22 | #3066 merges — ceiling replaced by a 500 ms stub/hang guard, mechanism arm made the discriminator |

#3066 and #3067 are the same defect, written up 3 minutes apart, and never joined up.

**It does not reproduce.** Five consecutive `main` CI runs on 2026-09-02 pass the macOS job with
`kernel_sem_timedwait` green (runs `33642206188`, `33646171008`, `33648852673`, `33656129015`,
`33660262942`).

### Not the other two characters

- **Not a genuine macOS semantic difference.** The arm that would catch one — ARM 1, the mechanism
  assertion — passed on the failing run and passes today. The failing arm reads a clock.
- **Not flaky-everywhere-noticed-there.** The bound was 12 ms; Linux serves the same wait in
  ~5.0 ms, so it has ~2.4x margin there and has never failed. The overshoot is specific to a
  translated wait on a shared runner.

## The premise that no longer holds

**"The 1.02 s / 1.22 s durations point at ARM 4 waiting out its one-second timeout"** (issue
comment, 2026-09-01, offered explicitly as an untested lead). The sign is inverted. Every test in
that job that links `prosper_core` pays ~1.4 s for process start plus `register_builtin_hle()`
under Rosetta — `kernel_nanosleep` 1.48 s, `hle_functions_registered` 1.48 s, `json2_hle` 1.46 s —
against **0.01 s** for the same test on Linux. The **passing** durations across the five green runs
above are **1.08 / 1.20 / 1.35 / 1.47 / 1.52 s**:

- 1.22 s sits **inside** the passing distribution;
- 1.02 s is **below its minimum** — the failures were *faster* than a pass;
- an ARM 4 that really ran to its deadline would have read **~2.4 s**.

The duration is ~99% constant overhead and carries no signal about the wait. Recorded as
**instrument trap 249**.

*(The issue's own first comment already retracted its "master's last eight runs pass" line. Its
causation argument — the macOS job compiles zero files under `frontends/shared/live/`, so
`live_compute.cpp` is not in that binary — stands, and is worth keeping.)*

## What this PR changes, and why it is not just closing the issue

Making it green is not the goal; the goal is that a red macOS job means something. The residual
defect is that **the number which settles this exists only on red runs**.

`test_kernel_sem_timedwait.cpp` prints `(timeout took %.2f ms via %s)` for exactly this purpose —
its own comment says so, and names #3044 as having merged "on a green Rosetta job whose numbers
nobody had seen". But every CI job runs `ctest --output-on-failure`, which shows a test's output
**only when it fails**. So the figure reaches a log precisely when something is already wrong, and
the headroom on the green runs either side of a near-miss is unrecoverable without reddening the
job first. That is #3208's defect — a threshold whose margin nobody has measured recently — sitting
one flag away from being measured continuously.

1. **A `timing-margin` ctest label** on the two tests whose pass rests on a measured wall-clock
   margin *and* which print it: `kernel_sem_timedwait`, `kernel_cond_timedwait_clock`.
   `kernel_nanosleep` is deliberately excluded — its 500 ms bounds are order-of-magnitude guards
   against a ~292-year sleep and it prints no value, so a verbose re-run would report nothing.
2. **A "Report the measured wait margins" step** in the **Linux**, **Windows MinGW** and **macOS**
   jobs: `ctest --test-dir prosper/build-ci -L timing-margin -V --no-tests=error`. Every green log
   now records the figures. Locally:

   ```
   55:          (timeout took 5.02 ms via none)
   55:          (posted acquire took 0.00 ms)
   56:          (realtime 20 ms timeout took 20.25 ms)
   100% tests passed out of 2
   ```
3. **The fourth Rosetta figure** added to the test header's canonical list: **13.58** ms, joining
   18.54 / 23.22 / 46.46. It is *below* the figure the list already carried, which sharpens the
   existing argument rather than repeating it — the 12 ms ceiling was not a bound that lost to an
   outlier, it sat **inside the ordinary spread** of that wait under translation.

### On margins, since the task asks

I am **not** widening any bound, and no bound needed widening: #3066 already replaced the
discriminating-by-accident ceiling with a 500 ms stub/hang guard, and the file argues at length
(with five measured samples) that no single-run wall-clock bound can separate the defect from the
fix on any platform. Standing headroom, for the record:

| arm | bound | measured worst | headroom |
| --- | --- | --- | --- |
| ARM 3 upper (5 ms request) | 500 ms | 46.46 ms (Rosetta) | ~10.8x |
| ARM 3 lower | ≥ 4 ms | 13.58 ms (Rosetta), 5.02 ms (Linux) | ~1.25x on Linux |
| ARM 4 (posted acquire, POSIX) | 100 ms | 0.00 ms (Linux); sub-ms per #3066 | ≫100x |

The point of the label is that these stop being a claim recorded once and become a figure printed
on every run — so the *next* near-miss is legible before it flips, which is the diagnostic #3208
asks for.

### Self-check, with its mutation arm

`-L timing-margin` plus `--no-tests=error` fails the step if the label is ever renamed or dropped,
rather than silently reporting nothing for ever after. Measured against a deliberately wrong label:

```
ctest -L timing-margin-RENAMED -V --no-tests=error   ->  exit 8
ctest -L timing-margin-RENAMED -V                    ->  exit 0
```

The flag is load-bearing; without it the guard would be vacuous.

## Verified where

| claim | status |
| --- | --- |
| Which arm failed, and at what value | **Verified** — read out of the failing job's log (run `33047765601`, job `98438836203`), not inferred. |
| It does not reproduce on `main` | **Verified** — five consecutive green macOS jobs on 2026-09-02, listed above. |
| The duration lead is falsified | **Verified** — passing durations 1.08–1.52 s harvested from those same five job logs, straddling both reported failure durations. |
| The label, the report command, and its mutation arm | **Verified on Linux** — commands and exit codes below. |
| The report step running on macOS and Windows | **NOT verified locally — I cannot run either.** The command is identical on all three jobs and the label is registered in the platform-independent block (both tests appear in the macOS job's own suite as #51/#52). CI on this PR is the measurement; see below. |
| That the Rosetta figures stay under 500 ms indefinitely | **Not claimed.** Nothing here makes the bound safer; it makes the margin observable. |

**Not verified and not claimable from here:** anything about a *future* Rosetta regime. If the
runner image changes and the wait latency moves an order of magnitude, ARM 3 will redden again —
and, with this change, the log will say by how much instead of leaving it to be re-derived.

## Risks

- The report step **re-runs** two already-gating tests, so a timing arm that fails only on the
  second sample would redden a job that was previously green. Left blocking on purpose: that is a
  genuine second observation, `-V` names the arm and the value, and the bounds it would have to
  clear are 500 ms against a 46 ms worst case. Cost is ~3 s per job.
- Renaming or removing either test drops it from the label silently; `--no-tests=error` catches the
  case where the label goes away entirely, not the case where one of two tests leaves it.
- `.github/workflows/ci.yml`, `prosper/CMakeLists.txt` and `prosper/docs/PORTING.md` are also
  touched by open PRs #3242/#3244; the edits are in different regions but a rebase may be needed.

## Build / test commands and results

Own baseline on `origin/main` (`a0daed9c`), before any edit:

```
cmake -S prosper -B prosper/build-linux            # exit 0
ctest --no-tests=error                             # exit 0 — 100% tests passed out of 340
```

After the change, same worktree, full rebuild:

```
ctest --no-tests=error                             # exit 0 — 100% tests passed out of 340
ctest -N -L timing-margin --no-tests=error         # exit 0 — Total Tests: 2
                                                   #   #55 kernel_sem_timedwait
                                                   #   #56 kernel_cond_timedwait_clock
ctest -L timing-margin -V --no-tests=error         # exit 0 — 100% tests passed out of 2
ctest -L timing-margin-RENAMED -V --no-tests=error # exit 8  (mutation arm)
ctest -L timing-margin-RENAMED -V                  # exit 0  (flag is load-bearing)
```

Docs gates:

```
check_numbered_table.py --ordered --table-header Instrument --baseline <origin/main> \
    prosper/docs/GAME_COMPAT_ORCHESTRATION.md      # exit 0 — 1..249, 248 rows, unique/ascending,
                                                   #          none of the 247 base rows deleted
git ls-files '*.md' -z | xargs -0 check_numbered_table.py   # exit 0
git diff --check                                   # exit 0
```

Trap **249** allocated with `trap_number.py` (248 is claimed by open PR #3243; gaps are legal
since #2089).

## `## Ruled out` rows recorded

`prosper/docs/PORTING.md` § *Ruled out — the guest timed-wait family, and its CI* (heading widened
from "the Windows guest timed waits", since the new rows are macOS):

- the "unresolved: flaky vs. runner image" framing — it was neither, and the log named the arm;
- the ARM 4 duration lead — the sign is inverted, with the five passing durations as evidence;
- "a green CI job records the margin its timing tests cleared" — it did not, on any platform.

`prosper/docs/GAME_COMPAT_ORCHESTRATION.md` — instrument trap **249**: a ctest per-test duration on
a Rosetta runner measures startup, not the test.

## No screenshot, and no `BLOG.md` entry

Nothing here renders anything, and a CI-threshold post-mortem is not what that file is for. Stating
the decision rather than leaving it as an omission.

## CI on this PR — three runs, every job green, and it settles the last open question

| run | head | macOS | Linux | Windows MinGW |
| `33667200036` | `49267bdd` | pass 13m41s | pass 8m5s | pass 9m14s |
| `33669451332` | `481b371b` | pass | pass | pass |
| `33671084910` | `a2092208` (merged with `main`) | pass 17m40s | pass 9m1s | pass 9m54s |

**The macOS job settles the duration lead outright.** Run 1's `kernel_sem_timedwait` reads:

```
 51/278 Test  #51: kernel_sem_timedwait .......................   Passed    1.02 sec
```

**1.02 s — the exact duration of one of the two failures #3067 was filed on, green.** No inference
needed any more. Across the three runs it passes in **1.02 / 1.27 / 1.78 s**, which together with
the five `main` runs (1.08–1.52 s) makes eight green samples straddling both reported failure
durations.

And, for the first time in this repository, a green run carries the figures. Three samples per
platform, so this is a distribution rather than one lucky reading:

| job | 5 ms `sem_timedwait` | posted acquire | 20 ms `cond_timedwait` |
| --- | --- | --- | --- |
| Linux | 5.06 ms `via none` | 0.00 ms | 20.06 ms |
| Windows MinGW | **5.32 ms `via win32-high-resolution-timer`** | 0.00 ms | 21.73 ms |
| macOS / Rosetta 2 | **45.73 / 39.70 / 45.69 ms `via none`** | 0.00 ms | **156.18 / 166.79 / 142.89 ms** |

Two of those had never been measured on CI, and both are worth naming:

- **Windows serves the 5 ms request in 5.32 ms through the high-resolution timer.** That is #3044's
  fix observed on CI rather than on one developer's box — precisely the gap its own test comment
  complained about.
- **`kernel_cond_timedwait_clock`'s ARM 1 overshoots ~7–8x on Rosetta**: 142.89–166.79 ms for a
  20 ms request, against its `ms < 2000.0` bound. That is **12.0x** headroom on the worst of three,
  not the vast margin a 2 s bound reads as, and nobody knew because nothing printed it. Fine today;
  now a number that moves visibly in the log if the runner changes, instead of a bound that flips
  without warning.

The Rosetta `sem_timedwait` figures cluster tightly (39.70–45.73 ms, matching the 46.46 ms sample
already in the test header), so the ARM 3 headroom is **10.9x on a repeatable value**, not on a
single reading — which is the distinction #3208 asks for and the reason this is a label rather than
a one-off measurement in a PR body.

(The report step's own re-run is faster than the first — 0.27 s versus 1.02 s for the same test —
because Rosetta's translation cache is warm by then. Another reason not to read a duration.)


mattias800 (commented):
**APPROVED.**

## The diagnosis is forensic rather than speculative

The verdict is character 1 — a wall-clock bound too thin for Rosetta 2 — and the evidence is **the
failing job's own log**, which neither the issue body nor either of its two comments ever quoted:
one named arm, `ms < 12.0`, against a 5 ms request served in 13.58 ms, with every other arm green
including the mechanism one. The answer had been sitting in the artifact the whole time.

That is the most useful lesson in the PR, and it generalises past this issue: an issue that records a
red job without quoting the failing assertion invites everyone after it to re-derive the diagnosis
from scratch, or worse, to guess.

## The timeline reconstruction is the real find

**#3066 and #3067 are the same defect, written up three minutes apart, and never joined up** —
#3066 at 07:30, #3067 at 07:33, and #3066 merged the fix at 08:22 while #3067 stayed open recording
it as unresolved. So the issue has been stale for its entire life, and anybody trusting the tracker
would have gone looking for a live failure that no longer existed.

Establishing that by timeline rather than by "I could not reproduce it" is much stronger: a
non-reproduction is compatible with an intermittent bug, whereas *"here is the PR that fixed it and
here is the minute it merged"* is not.

## The right lesson was taken from #3208

The ceiling was replaced by a stub/hang guard with the mechanism arm as the discriminator, rather
than the ceiling simply being widened. Widening would have reproduced `terminator-boot`'s defect
exactly — a threshold with no measured headroom, green until the next slow runner. Naming the ~2.4x
margin the Linux path actually has, and identifying the overshoot as specific to a *translated* wait
on a shared runner, is the measured-headroom discipline #3208 asks for.

## Disposal

Closing a stale issue with the evidence is a legitimate outcome and I would have accepted it on its
own. Landing the instrument-trap row as well means the *class* — two issues for one defect, three
minutes apart, neither citing the other — is recorded rather than just this instance.

CI 9 success / 0 failures.

Issue #3249, opened 2026-09-02
live_compute host-reads every dispatch result with no availability barrier — the correct site #2944 cites is the comparator flags, not the results


mattias800:
## Summary

#2944 named `frontends/shared/live/live_compute.cpp` as one of the two places in the tree that
gets this right. That is true of the site it pointed at — the comparator flag buffer, which does
carry `srcAccessMask = VK_ACCESS_SHADER_WRITE_BIT`, `dstAccessMask = VK_ACCESS_HOST_READ_BIT`,
`VK_PIPELINE_STAGE_COMPUTE_SHADER_BIT -> VK_PIPELINE_STAGE_HOST_BIT`. It is not true of the
dispatch's **actual results**, which are the bytes that become guest memory.

After the dispatch completes, the guest writeback maps and host-reads:

* every writable storage **buffer**'s allocation — written by the compute shader
  (`VK_ACCESS_SHADER_WRITE_BIT` at `VK_PIPELINE_STAGE_COMPUTE_SHADER_BIT`);
* every storage **image**'s `staging[i]` buffer — written by `vkCmdCopyImageToBuffer`
  (`VK_ACCESS_TRANSFER_WRITE_BIT` at `VK_PIPELINE_STAGE_TRANSFER_BIT`).

Neither is made available to the host. The only ordering is the fence/queue wait after the submit,
and per the Vulkan spec a fence signal orders **execution**; it does not perform the availability
operation that moves those writes into the host domain. That needs a dependency with
`dstStageMask = VK_PIPELINE_STAGE_HOST_BIT` and `dstAccessMask = VK_ACCESS_HOST_READ_BIT`.

## Scope

Coherence: the compute allocator requires `HOST_VISIBLE | HOST_COHERENT` in **both** of its tiers
(the cached tier is `wanted | HOST_CACHED`, not `HOST_CACHED` alone), so
`vkInvalidateMappedMemoryRanges` is genuinely not required here. The missing half is the
dependency only — unlike the render backend, where three of the seven sites can be backed by
non-coherent memory.

Every image writeback restores its borrowed images with a
`COMPUTE_SHADER|TRANSFER -> ALL_COMMANDS` barrier before `vkEndCommandBuffer`. `ALL_COMMANDS` does
**not** include `VK_PIPELINE_STAGE_HOST_BIT`, and that barrier's destination access mask does not
include `VK_ACCESS_HOST_READ_BIT`, so it does not incidentally cover this.

## Suggested fix

[...]

Issue #3248, opened 2026-09-02
Synchronization validation is not enabled anywhere, and turning it on reports 5 hazards — one of them in the render backend's mip-assembly path


mattias800:
## Summary

`tools/vkval` runs the ctest suite under `VK_LAYER_KHRONOS_validation` with **core** validation
only. It does not enable **synchronization validation**, so no gate in the tree has ever looked at
prosper's barriers.

Switching it on for one run (2026-09-02, Fedora 44 / validation layers 1.4.341 / Mesa RADV
STRIX_HALO, `VK_INSTANCE_LAYERS=VK_LAYER_KHRONOS_validation VK_LAYER_VALIDATE_SYNC=1 ctest`) reports
**5 SYNC-HAZARD messages across 2 tests**, none of which the vkval allowlist has ever seen:

| test | id | x |
| --- | --- | --- |
| `texture_mip_render` | `SYNC-HAZARD-WRITE-AFTER-WRITE` | 4 |
| `vulkan_triangle_render` | `SYNC-HAZARD-READ-AFTER-WRITE` | 1 |

## The one in the render backend (real)

`prosper/tests/fixtures/render_runner.h`, the `assembled_target_mips` path:

```
vkCmdClearColorImage(cmd, upload.image, VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL,
                     &missing_level_clear, 1, &missing_level_range);
...
vkCmdCopyImage(cmd, upload.target_mip_images[level], VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL,
               upload.image, VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL, 1, &copy);
```

The clear writes every level of `upload.image`; the per-level copy then writes the same image. The
barriers between them are all on `upload.target_mip_images[level]` — the SOURCE. Nothing orders the
two writes to the destination, so the layer reports:

```
vkCmdCopyImage(): WRITE_AFTER_WRITE hazard detected. vkCmdCopyImage writes to VkImage ...,
which was previously written by vkCmdClearColorImage.
No sufficient synchronization is present to ensure that a write (VK_ACCESS_2_TRANSFER_WRITE_BIT)
at VK_PIPELINE_STAGE_2_COPY_BIT does not conflict with a prior write of the same type ...
```

This is the path that exists so a missing mip level stays black rather than being derived from a
[...]

PR #2949, opened 2026-08-23
#2945: one indexed draw, three seconds — plus the two spec violations found on the way there (and neither is the cause)

mattias800:
## What this is

> **Start here if you are picking this up:** the reproduction is
> `gpu_replay s3537.prgcap --draw 42` — **one indexed draw, three seconds, binary outcome** — and it
> is written out end to end at the head of `docs/GRAPHICS.md` § Ruled out → *Renderer determinism*,
> together with the three instruments that establish what is provably identical between a good and a
> bad run. Do not start from a boot.

#2945 is not fixed. What it now has is a **one-draw, three-second reproduction**, a precise
statement of the failure, a long list of dead hypotheses, two real spec violations found and fixed
along the way, and two committed instruments that produced the evidence.

Read the report first, then the diff:

- the bug is now `gpu_replay s3537.prgcap --draw 42` — **one indexed draw**, ~3 s, binary outcome
- the failure is: **every storage-buffer load in that draw's vertex shader intermittently returns 0**,
  so the three transformed vertices collapse to one clip point and the clipper discards the primitive
- between a good and a bad run, *everything observable is identical* (list below)
- **and the standalone control reproduces it** — a bare Vulkan program, a pipeline that validates
  clean, prosper's dumped modules and no prosper code in the process, failing 3 runs of ~63. That
  shows prosper's host-side Vulkan usage is **not necessary** for the failure, which re-scopes the
  whole `## Ruled out` list. See "The control was never valid" below; it is the most important thing
  in this PR.

## The two fixes in here, and why neither is the cause

**1. Every render pass was created with `dependencyCount = 0`.** Vulkan then supplies the default
external dependencies, whose outgoing form is `dstStageMask = BOTTOM_OF_PIPE, dstAccessMask = 0` —
the pass's writes are made *available* and visible to nothing. Attachments with
`finalLayout = TRANSFER_SRC_OPTIMAL` are read by `vkCmdCopyImageToBuffer` a few commands later with
no dependency at all, and the layout transition at `vkCmdEndRenderPass` is itself a write of the
whole image (on RADV potentially a DCC decompress). The barrier the readback path *does* emit is
guarded on `persistent`, so a transient target's readback has never been ordered against the pass
that filled it.

Synchronization validation on one offline replay: **10 `SYNC-HAZARD-READ-AFTER-WRITE` before, 0
after.**

**2. The vertex output interface exceeded `maxVertexOutputComponents` on every AMD device.**
[...]

PR #3250, opened 2026-09-02
render backend: a readback needs an availability operation, not just a fence (#2944)

### LINKED ISSUE #2944: Every GPU readback in the render backend is read by the host with no availability barrier (VK_ACCESS_HOST_READ_BIT appears nowhere in the render path)
## Summary

No readback in prosper's render backend makes its transfer write **available to the host**.
`VK_ACCESS_HOST_READ_BIT` appears **nowhere** in the render path — the only occurrences in the
repository are `frontends/prosper-app/main.cpp:655` (the F9 grab) and
`frontends/shared/live/live_compute.cpp:8039` (a compute writeback).

Four sites in `prosper/tests/fixtures/render_runner.h` copy device memory into a `HOST_VISIBLE`
buffer that the CPU maps and reads immediately after the wait, with no host-domain dependency:

| site | what it reads back |
| --- | --- |
| `submit_persistent_ds_transfer` (`vkCmdCopyImageToBuffer` in `copy_plane`) | depth / stencil planes |
| `readback_persistent_color_target` | a persistent colour target |
| `render_draw_pass_rgba`, storage-image writeback | a storage image into its staging buffer |
| `render_draw_pass_rgba`, frame readback (`rb`) | every colour slot of the composited frame |

Waiting a fence, or `vkQueueWaitIdle`, orders **execution**. It does not perform the availability
operation that moves a transfer write into the host domain — Vulkan requires an explicit dependency
with `dstStageMask = VK_PIPELINE_STAGE_HOST_BIT` and `dstAccessMask = VK_ACCESS_HOST_READ_BIT`
(equivalently `vkInvalidateMappedMemoryRanges` for non-coherent memory). Everything prosper reads
back — every composited frame, every snapshot PNG, every `gpu_replay` resource dump — is read
through this gap.

## What this is NOT

Filed on its own terms, and explicitly **not** as the cause of either defect it was found while
chasing. Both arms are matched A/B on one binary, the barrier added at all four sites behind a
`PROSPER_NO_HOST_READ_BARRIER` lever so only that variable differs:

- **Not #2932.** *BALAN WONDERWORLD* `PPSA02058`, `tools/screenshot`, default route, no pad input,
  `PROSPER_RENDER=1 PROSPER_GUEST_ARGS=-force-gfx-direct`, identical `--seconds 1 --count 110` arms:
  **15 of 110 content frames with the barrier, 17 of 110 without.** No effect.
- **Not #2937.** No Vulkan-execution test moved outside its noise band, and on
  `descriptor_array_render` the barrier arm measured *worse* — but see the caution below before
  reading that as a regression.

## Caution for whoever fixes this: the tests it would be measured on are flaky

Measured 30 runs each of one unchanged binary at master `08c23efd`, Linux / AMD Radeon 8060S (RADV
STRIX_HALO): `descriptor_array_render` **23/30 pass**, `multidraw_render` **3/30**,
`indexed_render` **0/30**, `gpu_execute` **0/30**. Twelve-run arms of the *same* unchanged binary
gave `descriptor_array_render` 10/12 and 5/12 within one afternoon. So a dozen samples per arm
cannot separate a real effect from the noise on those two, and the apparent "barrier makes
`descriptor_array_render` worse" result (0/12 vs 10/12) is not established.

## Suggested fix

One helper beside `dump_bmp` in `render_runner.h`, called after each of the four copies:

```
VkMemoryBarrier host_read{VK_STRUCTURE_TYPE_MEMORY_BARRIER};
host_read.srcAccessMask = VK_ACCESS_TRANSFER_WRITE_BIT;
host_read.dstAccessMask = VK_ACCESS_HOST_READ_BIT;
vkCmdPipelineBarrier(command, VK_PIPELINE_STAGE_TRANSFER_BIT, VK_PIPELINE_STAGE_HOST_BIT,
                     0, 1, &host_read, 0, nullptr, 0, nullptr);
```

That change was written, built and measured for this issue and then **reverted** rather than landed,
because it fixes nothing measurable and its only observed effect is on a test population too noisy
to judge it by. Land it with a sample size that can see past the flakiness, or land it together with
whatever makes those tests deterministic.

Found while investigating #2932 and #2937. `CONFIDENCE: HIGH` that the gap is real; `CONFIDENCE:
HIGH` that closing it alone changes nothing visible.

mattias800:
Fixes #2944

## The defect

A `vkCmdCopy*` writes a `HOST_VISIBLE` buffer; the CPU maps it and reads it after a fence wait. Per
the Vulkan spec a fence signal orders **execution**. It does not perform the **availability**
operation that moves that write into the host domain — that needs an explicit dependency with
`dstStageMask = VK_PIPELINE_STAGE_HOST_BIT` and `dstAccessMask = VK_ACCESS_HOST_READ_BIT`
(equivalently, for non-coherent memory, `vkInvalidateMappedMemoryRanges`, which is a *separate*
half, not a substitute).

Every readback in the render backend did the second half where it was needed and none of the first.

## Re-derived site list — the issue's is stale, in both directions

The issue names four rows and says `VK_ACCESS_HOST_READ_BIT` appears nowhere in the render path.
Neither is still true. `render_runner.h` gained a correct site in #2949 (the `PROSPER_BUFFER_ECHO`
diagnostic) and `tools/vkprobe` gained one that cites this issue by number. Re-derived against
current `origin/main` — every host read of device-written mapped memory in `render_runner.h`:

| # | site | producer | memory | invalidate before this PR | barrier before this PR |
| 1 | `submit_persistent_ds_transfer` readback → `read_persistent_ds_depth` | `vkCmdCopyImageToBuffer` | **may be non-coherent** (`HOST_CACHED` preferred, `HOST_COHERENT` fallback) | present | **absent** |
| 2 | `submit_persistent_ds_transfer` readback → `capture_persistent_ds_seeds` | `vkCmdCopyImageToBuffer` | **may be non-coherent** | present | **absent** |
| 3 | `readback_persistent_color_target` | `vkCmdCopyImageToBuffer` | **may be non-coherent** | present | **absent** |
| 4 | `render_draw_pass_rgba` storage-image writeback (`upload.staging`) | `vkCmdCopyImageToBuffer` | `HOST_COHERENT` (required) | n/a | **absent** |
| 5 | `render_draw_pass_rgba` frame readback (`rb`/`bmem`) | `vkCmdCopyImageToBuffer` ×N slots | `HOST_COHERENT` (required in **both** tiers) | n/a | **absent** |
| 6 | `PROSPER_DRAW_ISO` per-kill readback (`rb`) | `vkCmdCopyImageToBuffer` | `HOST_COHERENT` | n/a | **absent** — not in the issue |
| 7 | `PROSPER_GEOM_PROBE` transform-feedback pair (`geom_buf`, `geom_counter`) | transform feedback, **not** a transfer write | `HOST_COHERENT` (the readback tier needs `TRANSFER_DST`, which these do not carry) | n/a | **absent** — not in the issue |
| 8 | `PROSPER_BUFFER_ECHO` (`echo_buffer`) | `vkCmdCopyBuffer` | `HOST_COHERENT` | n/a | **present** (#2949) |

So: the issue's four rows cover five of the eight host-read call sites; two more (6, 7) were missed;
one (8) was fixed after it was filed. Rows 1–3 collapse to one producer, `submit_persistent_ds_transfer`,
which is where the barrier is placed.

**Coherent vs non-coherent.** Rows 1–3 come from `persistent_ds_transfer_buffer`, whose readback tier
deliberately *prefers* `HOST_CACHED` and falls back to `HOST_CACHED` **without** `HOST_COHERENT` if a
coherent cached type is absent — so those can genuinely be non-coherent, and their invalidates were
already correct. Rows 4–8 all require `HOST_COHERENT` in every allocation tier, so
`vkInvalidateMappedMemoryRanges` is not required for them; the dependency still is. **No site needed a
new invalidate** — there is no second bug of that shape here.

Row 7's producer is not a transfer at all, which is why the helper takes the source stage/access as
parameters rather than hardcoding `TRANSFER`. `VK_EXT_transform_feedback` requires
`VK_PIPELINE_STAGE_TRANSFORM_FEEDBACK_BIT_EXT` with `..._WRITE_BIT_EXT` / `..._COUNTER_WRITE_BIT_EXT`
as the source scope, and the barrier is recorded after `vkCmdEndRenderPass` because a buffer memory
barrier is not permitted inside a render pass instance.

## The fix

One helper beside `invalidate_mapped_readback()`, so the two halves of the readback contract sit
together and the comment can say they are independent:

* `host_read_barrier_for(buffer, src_stages, src_access)` — **pure**, returns the
  `VkBufferMemoryBarrier` plus its stage masks, so the masks can be asserted with no device;
* `record_host_read_barrier(command, buffer, ...)` — records it and bumps a process-wide counter;
* `backend_host_read_barrier_count()` — the instrument the test reads.

Whole-buffer scope (`offset 0`, `VK_WHOLE_SIZE`): row 5 copies up to eight colour slots at different
offsets into one allocation, and one dependency covers them all.

Pattern choice: `VkBufferMemoryBarrier`, matching the two correct sites the issue names
(`prosper-app/main.cpp`'s F9 grab and `live_compute.cpp`'s comparator writeback). `render_runner.h`'s
own already-correct site used a global `VkMemoryBarrier`; it is folded into the helper rather than
left as a second way of doing the same thing.

## What the test can and cannot detect

`prosper/tests/render/test_host_read_barrier.cpp`, registered as `host_read_barrier`.

**Cannot**: detect the defect from output. Every readback buffer here is host-visible memory this
platform's driver keeps coherent enough that the *unfixed* code returns correct pixels — which is
exactly why the gap survived. Measured, not assumed: with the row-5 barrier deleted, the arm
`"the pass rendered and read back a full frame"` still **passes**.

**Can**: fail without the fix, on the path it drives. Structural assertion that the barrier is
**recorded**.

* Arm 1 (no device): the exact masks — `dstAccessMask` is exactly `VK_ACCESS_HOST_READ_BIT`,
  `dstStageMask` exactly `VK_PIPELINE_STAGE_HOST_BIT`, whole-buffer scope, no ownership transfer —
  plus a second case proving the host half does not change when the source scope is transform
  feedback.
* Arm 2: `render_draws_rgba` with readback → the counter must move.
* Arm 3 (**negative control, the actual discriminator**): the same draw with
  `want_color_readback = false`. No host read happens, so no barrier may be recorded. Without this,
  a counter incremented anywhere on the render path would satisfy arm 2 and prove nothing.

**Coverage**: one of the seven fixed sites — the composited-frame readback. The DS-plane transfer,
the persistent colour-target readback, the storage-image writeback and the three env-gated
diagnostics are not reachable from a fixture-only test without a title's resources. They share the
helper, and arm 1 pins the helper.

### Measured mutation arm

Deleting `record_host_read_barrier(cmd, rb)` from the composited-frame readback, rebuilding only the
test target:

```
  [ok]   the pass rendered and read back a full frame          <- the pixel check is BLIND to it
  host-read barriers recorded by the readback pass: 0
  [FAIL] a colour readback records a host-read barrier before the host maps the buffer
  [ok]   a pass with no host read records no host-read barrier <- the negative control still holds
== FAILED ==     ctest exit 8
```

Restored, `1` barrier is recorded and the case passes.

### Synchronization validation was weighed and rejected as the oracle

`tools/vkval` runs **core** validation only. Enabling syncval by hand
(`VK_INSTANCE_LAYERS=VK_LAYER_KHRONOS_validation VK_LAYER_VALIDATE_SYNC=1`) with the defect
deliberately live produced **zero** messages on `test_pipeline_render` and `test_host_read_barrier`.
That null is not silent-instrument noise: the same setting over the whole suite produced **5
`SYNC-HAZARD` messages** across two other tests, so syncval was armed and reporting. It simply
cannot observe a CPU read through a mapped pointer, which is the access the missing dependency
protects. The five hazards are a real, separate finding — filed as #3248, including one in the
render backend's mip-assembly path (`vkCmdClearColorImage` then `vkCmdCopyImage` to the same image
with no barrier between the two writes).

## Risks

* **A barrier that is not free.** Seven extra `vkCmdPipelineBarrier` calls per frame at most, on
  paths that are already doing a full-surface copy plus a fence wait. `TRANSFER → HOST` is the
  cheapest dependency there is; nothing waits on a queue that was not already about to be waited on.
* **Row 7's stage flag needs the extension enabled.** It is guarded by `geom_active`, which implies
  `geom_probe`, which implies `ds_ctx.transform_feedback_enabled` — the same condition that created
  the buffers. Validation is clean on the full suite.
* **A process-wide counter on the render path.** One relaxed `fetch_add` per readback. It is an
  instrument and is never read as a control input.
* **Behaviour change is intended to be none.** #2944 measured its own prototype as changing nothing
  visible (BALAN WONDERWORLD, 15/110 vs 17/110 content frames, matched A/B), and that expectation is
  unchanged. This lands as spec conformance.

## Also updated

* `docs/GRAPHICS.md` § Ruled out — three rows: correct pixels are not evidence of synchronization;
  syncval does not detect this class (with its positive control); `HOST_COHERENT` does not exempt a
  readback from the dependency.
* `tests/fixtures/AGENTS.md` — the two-halves rule, in the map for the folder that owns the backend.
* `tools/vkval/README.md` — a "what it does not cover" section, since the guard's silence on
  barriers had never been written down.
* `BLOG.md` — one entry, no picture.
* Filed while here: **#3248** (syncval is off; 5 hazards when it is on, one in the backend) and
  **#3249** (`live_compute` host-reads every dispatch *result* with no barrier — the site #2944 cites
  as correct is the comparator flags, not the results).

## Build / test

Container `ps5ys`, Fedora 44, Mesa RADV STRIX_HALO (Radeon 8060S), validation layers 1.4.341.

```
cmake -S prosper -B prosper/build-linux            # exit 0
ctest --no-tests=error                             # 341/341 passed, exit 0
python3 tools/vkval/vk_validation_scan.py --build-dir build-linux   # PASS, exit 0
                                                   # 5 ids / 52 messages, all pre-existing + allow-listed
```

`origin/main` baseline measured in the same worktree before the change: build exit 0, **340/340
passed, ctest exit 0**. The delta is the one registered case.

Mutation arm (row 5 barrier removed, `--target test_host_read_barrier`): build exit 0,
`ctest -R '^host_read_barrier$'` exit **8**, 0/1 passed.

### DIFF
--- a/BLOG.md
+++ b/BLOG.md
@@ -21,6 +21,15 @@ from the tracker issues, and still gated, because it is a projection of state ra
 
 ## 2026-09-02
 
+### Every frame we have ever read back was read without asking the GPU to hand it over
+
+No picture, and that is the finding: on this hardware a readback with no host-availability barrier
+comes back looking perfectly correct, so nothing has ever complained. Deleting the barrier we just
+added leaves the pixel check green and only the structural check red — and Vulkan's synchronization
+validation, which is armed and reporting five other hazards in the same run, cannot see it either,
+because it has no way to watch the CPU read a mapped pointer.
+[#2944](https://github.com/mattias800/prosper/issues/2944)
+
 ### A guest asking to sleep for 584 years was served in zero milliseconds
 
 No picture — the finding is that prosper's guest sleeps saturate a hostile interval to the largest
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -2715,6 +2715,20 @@ if(TARGET prosper_core)
       add_test(NAME backend_persistent_resource_lock
                COMMAND test_backend_persistent_resource_lock)
 
+      # #2944: every GPU readback in the render backend is host-read after a fence wait, and a fence
+      # orders execution rather than performing the availability operation into the host domain.
+      # Structural by necessity -- on a coherent desktop driver the UNFIXED code returns correct
+      # pixels, so the assertion is that the barrier is RECORDED, with a no-readback negative control
+      # so the counter cannot be satisfied by any render at all. Same include roots and same real
+      # Vulkan device as the render_runner.h tests above.
+      add_executable(test_host_read_barrier tests/render/test_host_read_barrier.cpp)
+      # Includes a shared fixture, which resolves from the `tests` root.
+      target_include_directories(test_host_read_barrier PRIVATE tests)
+      # Reaches `shared/...` through render_runner.h, which resolves from `frontends`.
+      target_include_directories(test_host_read_barrier PRIVATE frontends)
+      target_link_libraries(test_host_read_barrier PRIVATE prosper_core Vulkan::Vulkan)
+      add_test(NAME host_read_barrier COMMAND test_host_read_barrier)
+
       # The address-watch matcher's edge cases (exclusive end, empty range, UINT64_MAX wrap).
       add_executable(test_compute_address_watch tests/gpu/test_compute_address_watch.cpp)
       target_link_libraries(test_compute_address_watch PRIVATE prosper_core Vulkan::Vulkan)
--- a/prosper/docs/GRAPHICS.md
+++ b/prosper/docs/GRAPHICS.md
@@ -795,6 +795,28 @@ re-deriving — and read it before forming a hypothesis about a frozen, black, o
   than a dozen samples per arm to say anything — twelve-run arms of one unchanged binary produced
   10/12 and 5/12 for `descriptor_array_render` on the same afternoon. Only the 0/30 pair is a stable
   discriminator. #2937.
+- **Correct pixels are not evidence that a readback is synchronized, and this is the reason #2944
+  survived unnoticed for the life of the backend.** Every readback buffer in the render backend is
+  host-visible memory this platform's driver keeps coherent enough that the *unfixed* code returns
+  the right bytes. Measured directly: with the host-read barrier deleted from the composited-frame
+  readback, `test_host_read_barrier`'s pixel assertion ("the pass rendered and read back a full
+  frame") still **passes** while the structural assertion goes red. So a readback test that checks
+  content can never discriminate here — assert that the barrier is RECORDED. #2944.
+- **Synchronization validation does not detect a missing host-read barrier — do not reach for it as
+  the oracle for this class.** Measured 2026-09-02, Fedora 44 / validation layers 1.4.341 / Mesa
+  RADV STRIX_HALO, with the composited-frame barrier deleted so the defect was live:
+  `VK_LAYER_VALIDATE_SYNC=1` over `test_pipeline_render` and `test_host_read_barrier` produced
+  **zero** messages. The lever is proven to move — the same setting over the whole ctest suite
+  produced **5 SYNC-HAZARD messages** (one READ-AFTER-WRITE in `vulkan_triangle_render`, four
+  WRITE-AFTER-WRITE in `texture_mip_render`), so syncval was armed and reporting; it simply cannot
+  observe a CPU read through a mapped pointer, which is the access the missing dependency protects.
+  Those 5 are a separate, real finding (#3248) and are invisible to `tools/vkval`, which does not
+  enable syncval. #2944.
+- **HOST_COHERENT memory does NOT exempt a readback from the host-domain dependency.** Coherence
+  removes the need for `vkInvalidateMappedMemoryRanges`; it does not perform the availability
+  operation. The two halves are independent, and the backend had only the second: three of the seven
+  readback sites already invalidated (they may be backed by non-coherent HOST_CACHED memory) while
+  none had the barrier. #2944.
 
 ## The renderer is not deterministic on frozen input (2026-08-23) — #2945
 
--- a/prosper/tests/fixtures/AGENTS.md
+++ b/prosper/tests/fixtures/AGENTS.md
@@ -22,7 +22,7 @@ include it. `render_draw_pass_rgba` is therefore a live render path — a wrong 
 silently drops real rendered content in a real game, and "it is under `tests/`" has misled reviewers
 into treating changes here as test-only (#3210 had to say so explicitly in its own body).
 
-Two consequences for anything you change in it:
+Consequences for anything you change in it:
 
 - **Dropping a pass or skipping a draw is a VISIBLE regression** when the condition can fire on a
   healthy device. Guard on the API's own `VkResult`, not on a heuristic, and say in the PR what a
@@ -39,6 +39,12 @@ Two consequences for anything you change in it:
   caches' other entry points (`invalidate_persistent_color_target*`,
   `readback_persistent_color_target`, `snapshot_persistent_ds_images`, and the frontend's direct
   iteration of both), so do not add a new one without reading #3240.
+- **A readback has TWO halves and they are independent** (#2944). Any device write the CPU then maps
+  and reads needs `record_host_read_barrier()` — the availability operation into the host domain,
+  which a fence wait does not perform — and, when the memory may be non-coherent,
+  `invalidate_mapped_readback()`. Coherent memory still needs the first. Neither half can be checked
+  by reading the pixels back: on every driver here the unsynchronized code returns correct bytes, so
+  the guard is `host_read_barrier`, which asserts the barrier was RECORDED.
 
 ## Adding to it
 
--- a/prosper/tests/fixtures/render_runner.h
+++ b/prosper/tests/fixtures/render_runner.h
@@ -3947,6 +3947,68 @@ inline bool invalidate_mapped_readback(const RenderVkCtx& ctx, VkDeviceMemory me
     return vkInvalidateMappedMemoryRanges(ctx.dev, 1, &range) == VK_SUCCESS;
 }
 
+// The OTHER half of a readback, and a separate operation from the invalidate above.
+//
+// Waiting a fence -- or vkQueueWaitIdle -- orders EXECUTION. It does not perform the availability
+// operation that moves a transfer (or shader, or transform-feedback) write into the HOST domain.
+// Vulkan requires an explicit dependency for that: dstStageMask = VK_PIPELINE_STAGE_HOST_BIT with
+// dstAccessMask = VK_ACCESS_HOST_READ_BIT (#2944).
+//
+// Coherent memory does not exempt a readback from it. HOST_COHERENT removes the need for
+// vkInvalidateMappedMemoryRanges; it does not remove the need for the dependency. So the two halves
+// are independent: non-coherent memory needs both, coherent memory needs this one.
+//
+// Every driver this project runs on completes the availability anyway, which is exactly why the gap
+// survived unnoticed -- the pixels come back correct and nothing says the code was wrong. The fix is
+// spec conformance, not a chase after a visible symptom.
+struct HostReadBarrier {
+    VkBufferMemoryBarrier barrier{VK_STRUCTURE_TYPE_BUFFER_MEMORY_BARRIER};
+    VkPipelineStageFlags src_stages = 0;
+    VkPipelineStageFlags dst_stages = 0;
+};
+
+// Pure, so the masks can be asserted with no device (`host_read_barrier` ctest case). The buffer
+// scope is whole-buffer: a readback buffer holding several slots (the MRT frame buffer writes up to
+// eight colour offsets into one allocation) is made available in one dependency.
+inline HostReadBarrier host_read_barrier_for(
+        VkBuffer buffer,
+        VkPipelineStageFlags src_stages = VK_PIPELINE_STAGE_TRANSFER_BIT,
+        VkAccessFlags src_access = VK_ACCESS_TRANSFER_WRITE_BIT) {
+    HostReadBarrier host_read{};
+    host_read.barrier.srcAccessMask = src_access;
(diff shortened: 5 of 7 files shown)


mattias800:
### Coverage note: the three env-gated sites the ctest case cannot reach were validated by hand

The registered test drives site 5 (the composited frame) only. Sites 6, 7 and 8 are behind
`PROSPER_DRAW_ISO`, `PROSPER_GEOM_PROBE` and `PROSPER_BUFFER_ECHO`, so ctest never records their
barriers. Each was run directly under the validation layer with synchronization validation on
(`VK_INSTANCE_LAYERS=VK_LAYER_KHRONOS_validation VK_LAYER_VALIDATE_SYNC=1`, `test_pipeline_render`,
Fedora 44 / layers 1.4.341 / Mesa RADV STRIX_HALO):

| arm | result |
| --- | --- |
| `PROSPER_GEOM_PROBE=0` | exit 0, probe reported; **0** `vkCmdPipelineBarrier` VUIDs, **0** SYNC-HAZARD |
| `PROSPER_DRAW_ISO=1 PROSPER_ISO_AT=32,32` | exit 0, isolation ran; **0** `vkCmdPipelineBarrier` VUIDs, **0** SYNC-HAZARD |
| `PROSPER_BUFFER_ECHO=1` | exit 0, echo printed; **0** validation messages of any kind |

This matters most for site 7, whose source scope is `VK_PIPELINE_STAGE_TRANSFORM_FEEDBACK_BIT_EXT` —
a stage that is only legal when the extension and the `transformFeedback` feature are enabled at
device creation. They are: `RenderVkCtx::transform_feedback_enabled` is set in the same block that
pushes the extension and chains the feature struct, and `geom_active` cannot be true without it. The
layer confirms the flag is accepted.

Three validation ids did appear, all **pre-existing and unrelated to this change** — none is a
barrier VUID:

* `VUID-vkCmdBeginTransformFeedbackEXT-None-04128` ×10 — the probe's shader is not xfb-decorated,
  which the code's own comment says is gated separately in `gpu_executor`;
* `VUID-vkCmdDraw-None-07831` / `-07832` ×1 each — the `PROSPER_DRAW_ISO` re-render path replays
  draws without re-issuing the dynamic depth-bias/stencil state the main loop sets.

Both are diagnostics-only and noted on #3248 rather than fixed here.

mattias800 (commented):
**APPROVED.**

## Rejecting syncval as the oracle — with a positive control — is the best thing here

With the defect deliberately live, `VK_LAYER_VALIDATE_SYNC=1` produced **zero** messages. Almost
every lane would have stopped there and either concluded the bug was imaginary or shipped syncval as
the regression guard.

Instead the null was **validated**: the same setting over the whole suite produced **5 SYNC-HAZARD
messages in two other tests**, proving the layer was armed and reporting, and simply cannot observe a
CPU read through a mapped pointer. That is the difference between "the instrument says no" and "the
instrument cannot say yes", and it is the distinction this project's trap list exists for. A clean
zero from a same-source control is exactly the trap; this one went outside the source to check.

## The mutation arm's second half is the load-bearing observation

Deleting the barrier reddens the new counter arm (exit 8, 0/1) — expected. What matters is that the
arm *"the pass rendered and read back a full frame"* **stays green in the mutated build**. The pixel
check is blind to the defect, which is precisely why this survived in-tree for so long, and stating
it turns "my test passes" into "here is why the existing tests never could".

## Re-deriving the site list paid, in both directions

Eight host reads, not four: the issue's rows cover five, and rows **6 and 7** (`PROSPER_DRAW_ISO`,
`PROSPER_GEOM_PROBE`) are new. Row 7 is not even a transfer — transform feedback — so it needs
different source stage/access, and recording it after `vkCmdEndRenderPass` because buffer barriers
are illegal inside a render pass instance is the sort of detail a mechanical sweep would have got
wrong.

The issue's headline claim is also stale — `VK_ACCESS_HOST_READ_BIT` does appear in the render path
now (#2949), and in `tools/vkprobe` citing #2944 by number. Saying so rather than quietly fixing
around it keeps the record honest.

And the coherent/non-coherent question was answered per site rather than assumed: rows 1–3 come from
a tier that can fall back to `HOST_CACHED` **without** `HOST_COHERENT`, so they are genuinely
non-coherent — and their invalidates were **already correct**. Concluding "there is no second bug of
that shape" is a real negative result, and worth as much as the fix.

## Honest coverage limit

The test drives 1 of 7 sites and says so, with the other three env-gated ones validated by hand under
the layer (0 barrier VUIDs, 0 SYNC-HAZARD each) and posted as a comment. The `want_color_readback=false`
negative control means the counter cannot be satisfied by any render at all — that is what stops the
arm passing vacuously.

Folding row 8's existing inline barrier into the same helper so the file has one pattern is the right
tidy-up: two spellings of the same rule is how the next one gets missed.

## Follow-ups filed rather than bundled

**#3248** is the valuable one and I agree it wants its own lane — syncval enabled nowhere, and
turning it on finds a real **WAW in the backend's `assembled_target_mips` path** (clear then copy to
the same image, barriers only on the source). The two lying diagnostics it also found —
`PROSPER_GEOM_PROBE` reporting "0 vertices" on a draw that did produce primitives because its shader
is not xfb-decorated, and `PROSPER_DRAW_ISO` replaying without the dynamic state the main loop sets —
are instrument defects of exactly the kind that generate phantom findings.

**#3249** correctly distinguishes the compute *results* read (unguarded) from the comparator flags
the issue cited as the correct example.

## Evidence

Baseline measured in the same worktree (340/340, exit 0), branch 341/341 exit 0, `vkval` PASS with
all 52 messages pre-existing and allow-listed. CI 9/9.

PR #3251, opened 2026-09-02
test_game_compute hosts write-watch promotion, and the promotion decision is now counted where it is taken (#3156, #3155)

### LINKED ISSUE #3156: test_game_compute cannot host write-watch promotion (no SIGSEGV handler): PROMOTE_HITS=1 faults there, not in production

mattias800:
## What this is

Two issues that are one knot. **#3156** — `test_game_compute` SIGSEGVs at
`PROSPER_COMPUTE_WRITE_WATCH_PROMOTE_HITS=1`. **#3155** — write-watch promotion *appears* inert,
because the only setting that measurably reduced `memcmp` traffic was that same crashing one.

The order matters. #3156 is an instrument problem: nobody could measure #3155's claim in the harness
that would prove it, because the one setting that exercises promotion is the one that crashed there.
Fix the instrument, then use it.

`Fixes #3156` · `Refs #3155` — no promotion default is changed. See *Why no default moved* below.

## #3156: the harness asserted a capability it did not have

`guest_write_watch_set_fault_onstack(true)` is a promise to the watch layer: a SIGSEGV handler is
installed, it runs on a sigaltstack, and it will service a write fault on a page this process armed.
`test_game_compute.cpp` made that promise in **seven** places while installing **no handler at all**,
and compensated by calling `guest_write_watch_notify_host_write()` at the write sites it knew about.

That holds only while nothing arms a watch the test did not anticipate. At `PROMOTE_HITS=1` the
compute cache arms on a first acquisition, so the test's own `large_result[0] ^= 0xffffffffu;` (the
"externally changed buffer" step) stored into a page that had just been made read-only.

**Faulting access, pinned** — the issue recorded this as unknown:

```
Thread 1 "test_game_compu" received signal SIGSEGV
0x000000000042ff5f in main ()
$1 = (void *) 0x7ffff7049000                 # page-aligned, inside the test's own mmap
=> 0x42ff5f <main+29279>:   notl   0x0(%r13)  # uint32 ^= 0xffffffff, i.e. line 1324
```

The 32-bit operand width identifies the line: the file's other `^= 0xff` sites are all `uint8_t`.

`hits` 2 and 3 pass because the stability ladder never climbs far enough to arm anything in that
region. Production installs the handler (`exec_image_linux.cpp`) and does not reproduce this.

**Fix.** Model the one production behaviour that matters — service a write fault on an armed page and
resume — and keep every other fault fatal and loud. Following `tests/host/test_dmem.cpp`'s existing
pattern, and deliberately *not* `install_trap_handler()`: that brings hardware breakpoints, the
null-page companion mapping, lazy commit and guest-`%fs` scoping into a suite that has no guest. The
handler also checks the page-fault error code's write bit, exactly as production does, so a read fault
on an armed page stays fatal rather than being resumed.

## Behavioral contract

* A CPU write into a page the emulator armed is serviced: the page is made writable, the watch is
  marked Dirty, and the store re-executes and lands.
* **Any other fault ends the process** with a distinct exit code (86) and a message on stderr.
  Returning from a fault we did not arm would re-execute the faulting instruction forever, turning a
  real defect into a hang.
* No production code path changes. The handler is `#if defined(__linux__)` inside the test binary.

## Verification — #3156

| arm | build rc | run rc | note |
|---|---|---|---|
| `origin/main` (a6489acd3), `PROMOTE_HITS=1` | 0 | **139** | reproduced the issue as filed |
| `origin/main`, `PROMOTE_HITS=2 / 3 / 4` | 0 | 0 / 0 / 1 | matches the issue's table exactly |
| this branch, `PROMOTE_HITS=1` | 0 | 0 | `== PASS == (449 assertions executed)` |
| this branch, default and `=2` | 0 | 0 | 449 assertions |

### Mutation arms — each arm proved load-bearing

| mutation | build rc | run rc | what redded |
|---|---|---|---|
| **A** — delete the handler install, keep everything else | 0 | **139** | crashes at the *default*, not only at `hits=1`: the new probe arm takes an unhooked store through an armed page, so it detects a missing handler at every setting |
| **B** — handler `return`s instead of `_exit`ing | 0 | **1** | `FAIL: a write fault on an unarmed page remains fatal and reported`. The forked child spun on the faulting instruction 11,370,794 times until its `alarm(20)` fired — bounded on purpose, so a swallowing handler fails ctest instead of hanging it |
| **C1** — census bucket miscounted into a sibling | 0 | **1** (`write_watch_policy`) | exact-value arms |
| **C2** — census bucket dropped entirely | 0 | **1** (both) | `FAIL: census stability buckets partition the decisions they are a fraction of` |

Arm B is the one that matters for review: a handler that swallows everything is worse than the crash
it replaced, and without that arm nothing in the tree would notice.

## #3155: what the fixed harness can now show

The census counts the **decision**, not a proxy for its consequences —
`frontends/shared/texture/write_watch_census.hpp`, reported by `PROSPER_WATCH_PROMOTE_CENSUS=1` every
256 submits and at exit. This is deliberate: #3155's retracted 55% came from an `LD_PRELOAD` `memcmp`
interposer that dumped its running tally every 2^20 calls, and two arms were read at different dump
points. Every census line is a running total that prints its own denominator.

`test_game_compute`, 46 validated acquisitions, one run per arm (see the caveat below the table):

| arm | decisions | stability 0 | threshold met | granted | budget refused | watch-decided | exact compares |
|---|---|---|---|---|---|---|---|
| `PROMOTE_HITS=1` | 45 | 38 (84.4%) | 45 | 45 | 0 | 3 | 42 |
| `PROMOTE_HITS=2` | 50 | 39 (78.0%) | 11 | 11 | 0 | 1 | 44 |
| `PROMOTE_HITS=3` (default) | 51 | 39 (76.5%) | 4 | 4 | 0 | **0** | 45 |
| `PROMOTE_HITS=4` | 51 | 39 (76.5%) | 0 | 0 | 0 | 0 | 45 |
| `DEFER_MIN_KB=8192` | 45 | 38 (84.4%) | 45 | 45 | 0 | 3 | 42 |

**Read the counts, not the bytes.** Every arm reports `5.0 MiB compared`: the compares promotion
removes here are small, and the difference is below the report's one-decimal resolution. A byte claim
needs a title.

Three results, and the first two match the production census posted on #3155 (94.8% of decisions at
stability 0) closely enough to be the same phenomenon at a different scale:

1. **The stability counter is 0 for 76–84% of decisions.** Every `hits` value ≥ 2 is gated on a ladder
   the dominant population never climbs, which is why 2 and 3 measured indistinguishable in
   production. `hits=1` differs in *kind*: it removes the counter from the decision.
2. **At the default, promotion decides zero acquisitions here.** All 46 are settled by the submit
   journal (1) or a full compare (45). The budget refuses nothing in any arm.
3. **The reachable lever is not the threshold.** `should_promote_write_watch` promotes when the source
   is smaller than `defer_min_bytes` **or** the ladder is climbed. The renderer passes 8 MiB — that is
   what implements *"stable 1-8 MiB sources arm immediately"* in `RENDERER_PERFORMANCE_2026_07.md`.
   The compute call site has always passed a literal **1**, so its size exemption is unreachable.
   `DEFER_MIN_KB=8192` reproduces `hits=1`'s census **exactly** while leaving the proof requirement
   intact for large sources.

That literal `1` is precisely the arm the same doc measured and **rejected** for textures ("Deferring
every texture watch at or above 1 MiB was also rejected … cumulative exact validation rose from 69.7
to 79.2 GiB"). It has never been measured for compute.

### Why no default moved

Because it would be the exact mistake this issue has already made twice. PR #3158 changed a default on
a measurement that was later retracted. A compute source may be a storage *result* that changes every
frame, where arming an unproven watch is wasted `mprotect`/fault/rearm work — and the doc records an
experiment that dropped 3.8 → 1.7 FPS by growing watch count. **46 acquisitions in a synthetic harness
is not a title.** `PROSPER_COMPUTE_WATCH_DEFER_MIN_KB` turns the A/B into an environment sweep instead
of a patch; **unset is the historical 1 byte and changes nothing**.

The production sweep, for whoever has the emulator:

```
PROSPER_WATCH_PROMOTE_CENSUS=1                                  # baseline
PROSPER_WATCH_PROMOTE_CENSUS=1 PROSPER_COMPUTE_WATCH_DEFER_MIN_KB=8192   # renderer parity
PROSPER_WATCH_PROMOTE_CENSUS=1 PROSPER_COMPUTE_WATCH_DEFER_MIN_KB=0      # defer nothing
```

### The new knob refuses a typo out loud

`strtoull(value, nullptr, 10)` answers **0** for anything non-numeric, and on this knob 0 means "defer
nothing" — the most aggressive setting. So a typo would not disable the experiment, it would silently
run a different one. `PROSPER_COMPUTE_WATCH_DEFER_MIN_KB` therefore parses strictly and refuses loudly,
keeping the default:

```
$ PROSPER_COMPUTE_WATCH_DEFER_MIN_KB=eight ./test_game_compute
[watch-policy] PROSPER_COMPUTE_WATCH_DEFER_MIN_KB='eight' is not a number -- the compute defer
minimum stays at its default and NOTHING is being A/B'd
```

...and the census then matches the default arm exactly (`decisions=51 threshold_met=4`), which is how
you can tell it really kept the default rather than merely printing that it did.

**The four older variables in this family still have the original hazard** —
`{COMPUTE,TEXTURE}_WRITE_WATCH_PROMOTE_{HITS,MB}` and `TEXTURE_WRITE_WATCH_DEFER_MIN_KB`, across two
files, each of which turns a typo into "arm everything" or "unbounded budget". Filed as **#3253** with
the table of call sites rather than fixed here: fixing two of five would leave the tree inconsistent,
and the right fix is one strict parser in the policy header with its own unit arms.

* **Unaffected:** every promotion default (`PROMOTE_HITS=3`, `PROMOTE_MB=8`, defer minimum 1 byte),
  the renderer's texture path, Windows and macOS (page watches never arm there), and all production
  decision logic — `may_promote_write_watch_before_exact` is split into two named booleans that
  evaluate identically.
* **Affected:** `test_game_compute` gains a SIGSEGV handler and 6 assertions (443 → 449). Census
  counters are collected unconditionally: per *acquisition*, not per byte, on a path whose expensive
  branch is a multi-megabyte `memcmp`. Only the report is gated.
* One `&&` chain in `acquire_cached_image` became an `if`, so the compare can be counted **where it
  runs** — the chain reaches `exact_unchanged` through several branches that never touch a byte. Same
  predicate, same short-circuit order.

## Risks

* The test-local handler is process-wide for that binary. It is scoped to write faults on armed pages
  and mutation arm B proves it does not absorb anything else.
* `fork()` in arm 2 runs before any Vulkan initialisation and the child only touches a `PROT_NONE`
  page. `alarm(20)` bounds it.
* Census atomics on the acquisition path. Contrast with `ComputeAliasCensusCounters`, which is
  deliberately non-atomic because it instruments a per-dispatch path it must not perturb; the reasoning
  for the different choice is in the header.

## Commands and results

```
cmake -S prosper -B prosper/build-linux                      # rc=0 (fresh configure)
cmake --build prosper/build-linux -j6                        # rc=0
ctest --no-tests=error                                       # rc=0, 100% of 343 tests passed
```

Baseline on `origin/main` (d734348f0), same box, same flags: build rc=0, `ctest --no-tests=error`
rc=0, **343** tests, 100% passed. So the test count is unchanged and the two `game_compute_exec` cases
plus `write_watch_policy` and `guest_write_watch` pass by name in both.

Docs gate: `check_numbered_table.py --ordered --table-header Instrument` →
`numbered column 1..250, 250 rows, unique and ascending, no unused numbers`. `git diff --check` clean.

No snapshots run: no rendering behaviour changes.

## Record

* `RENDERER_PERFORMANCE_2026_07.md` § *Compute write-watch promotion census* — the finding, the
  numbers, and a **`### Ruled out`** table with four rows (the `hits=1` "user-facing trap" reading, the
  threshold-retuning hypothesis, the budget hypothesis, and `FRONTEND_APP.md`'s claim that compute uses
  the same rule as textures).
* `GAME_COMPAT_ORCHESTRATION.md` **instrument trap 250** — a test binary that asserts a capability it
  does not have, so the one setting exercising the mechanism crashes there and nowhere else.
* `frontends/shared/texture/AGENTS.md` — new; the folder had none.
* `FRONTEND_APP.md` — corrected, plus both new variables.


mattias800 (commented):
**APPROVED.**

## The #3155 finding is the right answer to a question the issue asked wrongly

I verified the claim: `live_compute.cpp:1777` passes a **literal `1`** as `defer_min_bytes`, where
`live_renderer.cpp:4038` passes `cross_submit_watch_defer_min_bytes`. So compute's size exemption
(`source_bytes < defer_min_bytes`) is unreachable for any source of one byte or more — the ladder is
the *only* route to promotion there, which is exactly why it reads as inert.

That makes the issue's implied fix — retune `PROMOTE_HITS` — the wrong lever, and the census proves
it rather than arguing it: `DEFER_MIN_KB=8192` reproduces `hits=1`'s census **exactly** while keeping
the proof requirement for large sources. The effect was never the threshold.

The sharpest part is the provenance: that literal `1` is precisely the arm
`RENDERER_PERFORMANCE_2026_07.md` measured and **rejected** for textures (exact validation
69.7 → 79.2 GiB), and it has never been measured for compute. A rejected setting surviving as an
unexamined literal in a sibling call site is a genuinely valuable thing to have found.

## Declining to change the default is the correct call

46 acquisitions in a synthetic harness is not a title, and this issue has **already retracted one
measurement and closed one PR** on exactly that basis. Shipping a tuned default from a harness that
until today could not even host the mechanism would have been the wrong lesson twice over.

Landing `PROSPER_COMPUTE_WATCH_DEFER_MIN_KB` so the production A/B is an env sweep — unset preserving
the historical 1 byte — converts an unanswerable question into a measurable one without betting on
the answer. That is the right shape.

Being explicit that **bytes are not quotable** because every arm prints `5.0 MiB`, under the report's
resolution, is the kind of restraint that keeps a census trustworthy.

## #3156's diagnosis is properly pinned

Not "a fault somewhere in the watch path" but `notl 0x0(%r13)` at a page-aligned `si_addr`, with the
**32-bit operand width** identifying the line uniquely — the file's other `^=` sites are `uint8_t`
and would assemble as `notb`. That is a real identification rather than a plausible one.

And the cause is the instrument lying about itself: `guest_write_watch_set_fault_onstack(true)` is an
*assertion to the watch layer that a handler exists*, made in **seven** places by a binary that
installed none. Modelling the production pattern rather than reaching for `install_trap_handler()` is
the right fix.

## The mutation arms answer the two questions I would have asked

**A**: deleting the handler install faults at the **default**, not only at `hits=1` — so the new arm
detects the defect at every setting, rather than only at the one the issue happened to name.

**B** is the one I explicitly worried about in the brief: a handler that swallows faults is worse
than a crash. The arm shows it — the child spun 11,370,794 times until a deliberate `alarm(20)`, so a
swallowing handler **fails ctest instead of hanging it**. Bounding that on purpose is the difference
between a test suite that reports a regression and one that stalls a CI job.

## Scope discipline

Fixing the bare-`strtoull` hazard on the new knob only — where a typo silently becomes `0`, i.e. the
*most aggressive* setting — and filing **#3253** for the four older variables in the same family
rather than fixing two of five, is right. A half-converted family is harder to reason about than an
un-converted one.

Correcting `FRONTEND_APP.md`'s claim that compute "uses the same exact-first rule" as textures is a
small edit with outsized value: that sentence is what would have stopped the next reader from finding
the literal `1`.

## Evidence

Baseline measured on a **fresh** configure (343/343, exit 0) rather than an incremental tree — which
matters, since a build dir reused across branch switches has produced phantom failures here.

CI has build jobs outstanding; I will merge on green and will not merge past a non-zero failure count.

PR #3252, opened 2026-09-02
zerowatch: the tail of #3243's review that the merge missed -- including a silent 239-byte overread

mattias800:
## What this is

The tail of #3243's review that did not make it into the merge. `#3243` was merged from head
`fd1dbb92`, but the branch had moved to `1144bea7` — the PR's API head record lagged the branch ref
and no CI run was ever created for the two later commits. So `main` currently carries **the version
the third review rejected**, plus none of the fourth review's notes.

Two commits, cherry-picked cleanly onto current `main`. No behaviour change to the scan itself; one
reporting change and one new guard.

## What is missing from `main` today

**The rejected claim.** `main`'s `test_guest_memory_search.cpp` case 2b says the boundary sweep
covers the `accept_from` de-duplication. It does not — the searched start windows of consecutive
chunks are contiguous and disjoint, so replacing `if (addr >= accept_from)` with `if (true)` leaves
the suite green. That was the single blocking finding on the third review, and it is the same failure
as the PR's own subject: a measured-sounding claim that was never measured. `guest_memory_search.hpp`
carried the last copy of the same claim.

**An unreadable range scored `lost=0 kept=0`.** Neither counter's branch runs when every window fails
to read, so the most complete loss there is printed as "nothing happened". The line now carries
`unreadable=N`, and a range whose sixteen windows all fail prints `RANGE UNREADABLE … this is a total
loss, not lost=0` instead of a pair of zeros.

**Nothing checked that the FIND needle's leading bytes were distinctive.** A seed window is chosen
for holding data *somewhere* in its 256 bytes, so an all-zero 16-byte lead was reachable — and such a
needle matches everywhere, making both its hits and its misses meaningless. Arming now steps the
16-aligned start forward *inside* the window to the first non-zero 16 bytes, and `zw_find` refuses to
scan a still-degenerate needle rather than publishing a number for it.

**Writing that guard exposed a buffer overrun in the shift itself.** The last seed window starts at
`size - 256`, so *any* forward shift reads past the source buffer — up to 239 bytes, inside the
page-rounded staging mmap and therefore silent. Both shift loops are now bounded by the remaining
payload as well as by the window. Verified in review against a payload whose last byte abuts a
`PROT_NONE` guard page: as written, 32 of 32 shapes clean; with the bound removed it SIGSEGVs on the
guard page and violates `off + 256 <= size` in 17 of 32.

**A refused control is not an empty one.** The AFTER-ZERO skip line said "its control scan found
nothing" in the case where the control never ran at all because every needle it had was degenerate.
Different facts, different next steps; the line now names which. Reachable when a payload's only
non-zero bytes sit past the first 16 of window 7, where no shift is possible.

**A missing deep needle was silent** where a degenerate one is loud. A range with no second non-zero
window is followed by its header alone — a real limit on the result, now stated.

**The `full` fit guard is a bounds guard too.** When a range's last chunk is exactly `chunk` bytes an
unguarded `memcmp` of `needle_len` reads past the end of the fetch buffer. The comment said only the
correctness half.

## Verification (exact head, this branch)

```
cmake -S prosper -B prosper/build-linux -DPROSPER_APP=ON    CFG_RC=0
cmake --build prosper/build-linux -j6                        BUILD_RC=0
ctest --no-tests=error                                       CTEST_RC=0
```

344 tests, 100% passed — `main` has gained three cases from other lanes since #3243, which is why this is not the 341 quoted there. `git diff --check` clean. Diff against `main` touches four files and
`--stat` shows insertions dominating deletions in each, so nothing of another lane's was taken along
with the cherry-picks.

### Mutation arm for the new pin

The `full` fit test had no arm; it does now, and it cannot pass for the wrong reason — the needle's
tail is all zeros and the fetch buffer is zero-initialised, so a full-needle `memcmp` running past
`want` would **succeed** and report `full=true`.

Drop `off + needle_len <= want`, build exit 0:

```
FAIL: zero-tail: full=false -- the needle does not FIT, whatever the buffer holds
guest_memory_search: 1 failure(s)
```

Exactly one arm moves, with the hit count and address assertions still green. Reverted; green again.

The three arms from #3243 still hold on this head: `overlap = 0` → 4 failures;
`overlap = 0` **and** `kChunkBytes = 8<<20` → 4 failures; last chunk's `need = prefix_len` →
`needle_len` → 3 failures.

### Live

Re-measured on *Stray* (`PPSA02101`, `tools/screenshot`, `PROSPER_NULL_PAGE=1`, no pad input) after
the needle shift, since it changes which bytes the needle is: CONTROL finds the payload at exactly
one guest address (the destination), the post-flip first-window scan finds it at none, the deep
window is still at the destination — the same result #3243 reports. No degenerate needles and no
unreadable ranges occurred in that run, so those two paths are backstops rather than the expected
case and are untested live.

## Risks

Low. Everything here is behind `PROSPER_ZEROWATCH` / `PROSPER_ZEROWATCH_FIND`, which default off. The
overrun fix is the only change that could matter with the switch on, and it strictly narrows a read.

Refs #3142, #3243.

### DIFF
--- a/prosper/src/hle/fs/hle_file.cpp
+++ b/prosper/src/hle/fs/hle_file.cpp
@@ -2762,6 +2762,9 @@ struct ZeroWatchEntry {
     int reprofiles_done = 0;
     bool find_control_done = false;      // the "the data IS findable" arm has run
     bool find_control_found = false;
+    // The control did not run at all (every needle it had was degenerate), which is a different
+    // reason to skip the second scan than "the control looked and found nothing".
+    bool find_control_refused = false;
 };
 std::mutex g_zw_mx;
 std::vector<ZeroWatchEntry> g_zw;
@@ -2858,7 +2861,7 @@ size_t zerowatch_find_prefix() {
 }
 
 void zw_find(const ZeroWatchEntry& e, const char* when, const uint8_t* needle, uint64_t needle_off,
-             const char* window, bool* found_out) {
+             const char* window, bool* found_out, bool* refused_out) {
     // A needle whose leading prefix_len bytes are all zero matches essentially everywhere, so both
     // its hits and its misses are meaningless. The arming code below picks a non-zero start where
     // one exists inside the window; this is the backstop for when none does, and it refuses rather
@@ -2868,6 +2871,9 @@ void zw_find(const ZeroWatchEntry& e, const char* when, const uint8_t* needle, u
         if (needle[k]) degenerate = false;
     if (degenerate) {
         if (found_out) *found_out = false;
+        // REFUSED is not the same as "found nothing", and the difference decides whether a later
+        // scan is worth running. Reported separately so the skip line downstream can say which.
+        if (refused_out) *refused_out = true;
         fprintf(stderr, "[zerowatch-find] t=%llu %s dst=0x%llx window=%s seed-off=0x%llx "
                         "DEGENERATE NEEDLE (first %zu bytes all zero) -- NOT scanned\n",
                 (unsigned long long)prosper::diagnostics::diag_now_us(), when,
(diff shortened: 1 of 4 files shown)


mattias800:
## Author verification — exact head `bd25d759`

Linux, distrobox `ps5ys`, `-DPROSPER_APP=ON`.

```
cmake -S prosper -B prosper/build-linux -DPROSPER_APP=ON     CFG_RC=0
cmake --build prosper/build-linux -j6                         BUILD_RC=0
ctest --no-tests=error                                        CTEST_RC=0   344 tests, 100% passed
git diff --check                                              clean
```

### CI at this head — all green

Linux, Linux App, Linux App Package, Windows MinGW, Windows App, macOS (x86_64 / Rosetta 2),
macOS App (SDL3 + MoltenVK), Docs, Contribution shape — **9 pass**, 2 skipping
(`Progress tracker`, `Publish Release`), **0 failures**. The PR's head record matches the branch ref
this time, which is the thing that went wrong on #3243.

### Review status — carried over by content, not assumed

Both commits here were already independently reviewed on #3243:

* `734d45fd1` = `3e0fa4c5` on that PR, which the third review **APPROVED**, subject to one
  non-blocking item it asked to see fixed before merge: `guest_memory_search.hpp` still carrying the
  de-duplication claim that the implementation and test comments had been corrected to deny.
* `bd25d7596` = `1144bea7`, which is that fix plus three more of the same review's notes.

So the approved content and its one requested follow-up are both here, cherry-picked onto current
`main` with no conflict and nothing else taken along. A merge agent should establish that by
content — `git diff 3e0fa4c5 734d45fd1` and `git diff 1144bea7 bd25d7596` are both empty — rather
than by the review rows, which live on the other PR.

### How this was missed, so it is in the record rather than in a thread

#3243's PR **API head record stayed at `fd1dbb92`** while the branch ref moved to `3e0fa4c5` and then
`1144bea7`. No workflow run was ever created for either later commit (`gh run list` shows runs for
`0b999e41`, `c7686b45` and `fd1dbb92` only), and `gh pr checks` kept reporting the `fd1dbb92` run as
current. The merge then took the head GitHub believed in, which was two commits behind — including
behind a review that had rejected it.

The cheap guard is one command before merging: compare `gh api repos/OWNER/REPO/pulls/<N> --jq
'.head.sha'` against `git ls-remote origin refs/heads/<branch>`. They disagreed here for over an
hour, and nothing else in the flow noticed.

mattias800:
## Correction to my verification comment above — the equivalence check I named is the wrong one

I wrote that a merge agent can establish the content equivalence with
`git diff 3e0fa4c5 734d45fd1` and `git diff 1144bea7 bd25d7596`, "both empty". **They are not
empty**, and they cannot be: the cherry-picks sit on a newer `main`, so those commands compare whole
*trees* and report every other lane's work since #3243 — 32 files, 2,639 insertions.

The check I should have named compares the **patches**, not the trees:

```bash
for pair in "3e0fa4c5 734d45fd1" "1144bea7 bd25d7596"; do set -- $pair
  diff <(git diff $1^..$1 | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)') \
       <(git diff $2^..$2 | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)') \
    && echo "$1 vs $2: identical"; done
```

Run here: **both identical**. So the claim stands and the way I offered to check it did not, which is
the distinction this pair of PRs has spent its whole life on. Recording the correction rather than
editing the comment.

mattias800 (commented):
**APPROVED — merging promptly, because `main` currently carries a buffer overread.**

## What happened, recorded so it is not mistaken for a review failure

#3243's PR head record froze at `fd1dbb92` while the branch advanced two commits, and **GitHub never
created a CI run for `3e0fa4c5` or `1144bea7`**. I merged on a green check count — and that count was
honest about `fd1dbb92`; it simply was not about the branch tip. So `main` took the version the third
independent review **rejected**, missing the fix for a silent 239-byte overread.

The general lesson is worth more than this instance: *"the PR is green"* answers a question about the
commit GitHub last registered, which is not necessarily the commit the branch is on. Comparing the
recorded head against `git ls-remote` on the branch is one command and would have caught it.

## The overread itself

A chunked content scan that reads `prefix_len` bytes into the final chunk without the fit guard reads
past the buffer end — silent, because the bytes it reads are almost always mapped and the comparison
almost always fails anyway. Exactly the shape that survives review and testing indefinitely.

## The mutation arm that matters is the one that was green before review

`overlap = 0` **and** `kChunkBytes = 8<<20` passed **before** review, because the test's constant was
coupled to the implementation's by nothing stronger than a comment. That is a textbook instance of a
test that cannot fail for the reason it claims to: change both sides together and the arm agrees with
itself. Catching it, and recording that it was green, is more valuable than the four failures the
other arms produce.

## Disposition

These are the same two commits, cherry-picked onto current `main` and reported patch-identical to the
originals — so the review history on #3243 (REJECTED, REJECTED, APPROVED, with two of the blocking
findings being real bugs) applies to exactly this content. CI 9 success / 0 failures on `bd25d759`,
which is the run that never existed for these commits before.

Issue #306, opened 2026-07-10
UE4 (PPSA17942 DOLL): STALLED on loading screen — front-end boot flow never advances to title (content/patch/entitlement gate; NetCtl ruled out)


mattias800:
## Summary

DOLL (PPSA17942, "Dragon Quest VII Reimagined") boots into a fully-alive 60 fps UE4 engine
steady state within ~6 s (all asset IO complete, `DOLLLoadingScreen` UMG widget rendering) and
**never advances its front-end boot flow to the title screen** — indefinitely, at full frame
rate. This is a **correctness wall (STALLED), not llvmpipe slowness**: with the renderer OFF the
guest runs a true 60 fps for 7+ minutes / 25,000+ flips and still never progresses, does zero
file IO after t=6 s, and never enqueues a level-load task.

Full evidence + method: `prosper/docs/DOLL_LOADING_PROGRESSION.md` (branch
`diag/doll-loading-progression`).

## What is ruled out (A/B tested this session)

- **Slowness / IO / streaming**: linear 60 fps forever; all 781 opens + 5,430 APR reads in
  minute 0; async loading fully drained; not Oodle. A real GPU changes nothing about this wall.
- **Movies**: `PROSPER_DENY_SUBSTR=.usm` — no change (opening movie is never even opened).
- **Input / controller**: `PROSPER_PAD_PRESS=1` (connected pad, CROSS held) — no change.
- **NetCtl network state**: NEW gated experiment `PROSPER_NETCTL_CB=1` delivers the DISCONNECTED
  callback (correct guest-%fs handling) + `sceNetCtlGetState`=DISCONNECTED. Consumed cleanly
  (GetState called once) but **does NOT advance the flow**. The per-frame `sceNetCtlCheckCallback`
  pump (14,191/run) is the generic PS callback pump, not a network-state spin.

## Where the gate is (narrowed, CONFIDENCE MED)

The front-end boot flow's **"content / patch / entitlement ready" gate**. The eboot contains
UE4's `InstallBundleManager` PatchCheck (`EInstallBundleManagerPatchCheckResult::*`, "Could not
download patch data. Please check your internet connection..."). The online/update/entitlement
init calls all fire ONCE and then their follow-through never happens or returns garbage:

- `sceNpGetOnlineId` unimpl → **success with garbage out** (a signed-OUT console must return the
  signed-out error) → tells the game a user is signed in.
[...]

PR #3254, opened 2026-09-02
live compute: a dispatch result the host maps needs an availability op, not just a fence (#3249)

### LINKED ISSUE #3249: live_compute host-reads every dispatch result with no availability barrier — the correct site #2944 cites is the comparator flags, not the results

mattias800:
Fixes #3249

## The defect

The live compute backend maps every dispatch result after the fence wait and reads it into guest
memory. Per the Vulkan spec a fence signal orders **execution**; it does not perform the
**availability** operation that moves those writes into the host domain. That needs an explicit
dependency with `dstStageMask = VK_PIPELINE_STAGE_HOST_BIT` and
`dstAccessMask = VK_ACCESS_HOST_READ_BIT`.

#2944 named `live_compute.cpp` as one of the two places in the tree that already got this right.
That is true of the site it pointed at — the comparator's one-word flag buffer — and not of the
bytes that become guest memory.

## Re-derived site list — the issue's is incomplete, and in a way worth recording

The issue names two producers (writable storage buffers, storage-image staging). Re-derived against
current `origin/main`, there are **six host reads of device-written mapped memory** in
`live_compute.cpp` and **five producers** behind them.

The reason the issue's list is short is structural rather than an oversight: **the compute memory
pool recycles host-visible allocations across bindings and submits.** `release_memory` returns an
allocation to `memory_pool.available` and `allocate_memory` hands it to any later binding of the
same memory type, which then maps it and reads its retained contents to decide whether an upload is
needed — the code's own comment is *"Pooled host-visible allocations retain their previous
contents. Compare them with current guest memory before uploading."* So a buffer that the binding
which WROTE it never maps is still host-read later. **The invariant is per-allocation, not
per-binding.**

Host reads:

| # | host read | producer | in the issue |
| --- | --- | --- | --- |
| R1 | guest writeback of each writable storage buffer | P1 | yes |
| R2 | guest writeback of each storage image's `staging[i]` | P2 | yes |
| R3 | comparator flag word | P3 | named as the correct site |
| R4 | `acquire_cached_buffer`'s `memcmp` of a retained cached buffer against guest bytes | P1 (an earlier submit) | no |
| R5 | `compute_buffers_equal(mapped, source, …)` on a freshly bound **pooled** allocation | P1/P2/P4/P5 (any earlier submit) | no |
| R6 | `cached_image_result_matches` / `remember_cached_image_result` over the staging mapping | P2 | no |

Producers, with the source scope each needs:

| # | producer | source scope | memory | barrier before this PR |
| P1 | the dispatch itself, writing storage buffers | `SHADER_WRITE` @ `COMPUTE_SHADER` | `HOST_COHERENT` (required) | **absent** |
| P2 | `vkCmdCopyImageToBuffer` → storage-image staging | `TRANSFER_WRITE` @ `TRANSFER` | `HOST_COHERENT` (required) | **absent** |
| P3 | comparator shader → `compare_flags` | `SHADER_WRITE` @ `COMPUTE_SHADER` | `HOST_COHERENT` (required) | **present** (#2944's cited site) |
| P4 | `copy_result_baseline`'s `vkCmdCopyBuffer` → retained baseline | `TRANSFER_WRITE` @ `TRANSFER` | `HOST_COHERENT` (required) | **absent** — not in the issue |
| P5 | depth-bits bridge `vkCmdCopyImageToBuffer` → `staging[i]` | `TRANSFER_WRITE` @ `TRANSFER` | `HOST_COHERENT` (required) | **absent** — not in the issue |

P4 has a second source scope: the comparator **rewrites** each differing baseline in place
(`build_compute_compare_uvec4` emits `OpStore` into the baseline on the `different` branch), so that
baseline's last device write is a shader write, not the transfer copy. Both are covered.

**Coherent vs non-coherent — a real negative result, and the opposite of the render backend's.**
`host_memory_type` tries `HOST_VISIBLE|HOST_COHERENT|HOST_CACHED` and falls back to
`HOST_VISIBLE|HOST_COHERENT`; the cached tier is `wanted | HOST_CACHED`, not `HOST_CACHED` alone. So
**every** allocation the compute path maps is coherent, **0 of the 6 sites need
`vkInvalidateMappedMemoryRanges`**, and there is no second bug of that shape here. (#3250 found 3 of
8 render sites genuinely non-coherent; this backend has none.)

Also confirmed and not incidental cover: the end-of-command-buffer
`COMPUTE_SHADER|TRANSFER -> ALL_COMMANDS` image restore does **not** supply the dependency —
`VK_PIPELINE_STAGE_ALL_COMMANDS_BIT` does not include `VK_PIPELINE_STAGE_HOST_BIT`, and its
destination access mask does not include `VK_ACCESS_HOST_READ_BIT`.

## Reusing the helper rather than writing a second one

**The render backend's helper was not reachable from `live_compute.cpp`.** Checked with the
preprocessor rather than with grep — `-M` over that translation unit's own compile command lists no
`render_runner.h`. (`tests/fixtures/AGENTS.md` claimed `live_compute.cpp` was one of its three
includers. It is not; that line is corrected here.)

Including a 10.6k-line offscreen render backend into an 11k-line compute translation unit for a
20-line helper is the wrong fix, so the first commit is a **pure move**:
`record_host_read_barrier` / `host_read_barrier_for` / `backend_host_read_barrier_count` move
verbatim into `src/gpu/execute/host_read_barrier.hpp` (namespace `prosper::gpu`), and
`render_runner.h` re-exports them into `prosper::test` with four `using` declarations. Every
existing call site and the `host_read_barrier` ctest case are untouched. Both backends now record
the identical barrier through the identical call, and the counter is shared.

The **invalidate** half deliberately stays in `render_runner.h`, beside the only allocator in the
tree that can return non-coherent memory.

## The fix

Five `record_host_read_barrier` call sites in `live_compute.cpp`, each placed after the LAST device
write to its buffer, plus P3's existing hand-rolled barrier folded into the helper so the file has
one pattern. Whole-buffer scope replaces P3's exact flag span — wider and equally valid.

## What the test can and cannot detect

`prosper/tests/shared/live/test_live_compute_host_read_barrier.cpp`, registered as
`live_compute_host_read_barrier`. It drives the **production** backend through
`execute_live_compute_items`, the same entry `game_compute_exec` and `live_compute_descriptor_array`
use.

**Cannot**: detect the defect from the dispatch's result. Every compute allocation is
`HOST_COHERENT` and this driver returns byte-identical results without the barrier. **Measured, not
assumed** — with the P1 barrier deleted:

```
game_compute_exec ................ Passed   (== PASS == (439 assertions executed))
live_compute_descriptor_array .... Passed
```

and inside the new test itself, `"the production backend ran the dispatch and wrote the result back
to guest memory"` still `[ok]`.

**Cannot**: be caught by synchronization validation either. #3250 measured that syncval cannot
observe a CPU read through a mapped pointer, and proved the layer was armed by finding 5
`SYNC-HAZARD` messages elsewhere with the same setting (#3248). Not re-litigated here; not used as
an oracle.

**Can**: fail without the fix, structurally.

* **Arm 1** (no device) — the exact masks for the compute source scope: `dstAccessMask` exactly
  `VK_ACCESS_HOST_READ_BIT`, `dstStageMask` exactly `VK_PIPELINE_STAGE_HOST_BIT`, whole-buffer, no
  ownership transfer, and the host half unchanged between the `SHADER_WRITE` and `TRANSFER_WRITE`
  producers.
* **Arm 2** — one real dispatch with one writable storage buffer: the counter must move by exactly 1.
* **Arm 3 (negative control)** — a dispatch whose only binding is reflected **read-only**: the host
  reads nothing back, so no barrier may be recorded. Without it a counter incremented anywhere on
  the compute path would satisfy arm 2 and prove nothing. The arm asserts the control really does
  reflect a storage-buffer binding and reflects it as read-only, so it cannot pass vacuously by
  reflecting no descriptor at all.
* **Arm 4** — the *other* producer: a two-storage-image copy dispatch. Its expectation is **derived
  from reflection** (storage images + writable storage buffers = 3) rather than hardcoded, so it
  states the rule instead of a number.

**Coverage**: 2 of the 5 producers are driven end to end (P1, P2). P4 and P5 need a retained
baseline and a borrowed renderer depth plane respectively, neither of which a fixture-only test can
build; P3 was already correct. All five share the helper, and arm 1 pins the helper.

### Measured mutation arms

Both rebuild only `--target test_live_compute_host_read_barrier`.

**M1 — delete the P1 loop** (writable storage buffers):

```
  [ok]   the production backend ran the dispatch and wrote the result back to guest memory   <- BLIND
  host-read barriers recorded by one writable-buffer dispatch: 0
  [FAIL] a writable dispatch result records exactly one host-read barrier before the host maps it
  [ok]   a dispatch the host never reads back records no host-read barrier                   <- control holds
== FAIL ==     ctest exit 8, 0/1 passed
```

**M2 — delete the P2 call** (storage-image staging):

```
  [ok]   the production backend ran the storage-image copy and wrote guest memory            <- BLIND
  host-read barriers recorded by a two-storage-image dispatch: 1 (expected 3)
  [FAIL] every staging readback and every writable buffer records its own barrier
== FAIL ==     ctest exit 8, 0/1 passed
```

Restored, both arms pass with 1 and 3.

## Risks

* **Extra barriers on a hot path.** At most one `vkCmdPipelineBarrier` per writable buffer, per
  storage image, per retained baseline and per compare target, on a path that already records
  roughly ten per dispatch and ends in a fence wait. `COMPUTE_SHADER → HOST` and `TRANSFER → HOST`
  are the cheapest dependencies there are, and nothing waits on a queue that was not about to be
  waited on.
* **Placement.** Each is recorded after the last device write to its buffer: the comparator and the
  baseline copy that follow the dispatch only READ the storage buffers and staging buffers, so the
  post-dispatch and post-copy positions remain correct. The depth-bits barrier precedes that path's
  own `TRANSFER_WRITE → TRANSFER_READ` dependency; the two are independent.
* **A shared process-wide counter.** One relaxed `fetch_add`. It is an instrument, never a control
  input, and it is now shared with the render guard — both tests read deltas, not absolutes.
* **The pure move.** `render_runner.h` compiles into ~25 targets; all of them already include
  `gpu/...` headers, so `src` is on every include path that needed it. Full build is green.
* **Behaviour change is intended to be none.** This lands as spec conformance, like #3250.

## Also updated

* `docs/GRAPHICS.md` § Ruled out — five rows: the compute backend is blind to this by value too
  (with the measurement); the compute allocator cannot hand out non-coherent memory so no site needs
  a new invalidate; the `ALL_COMMANDS` image restore does not incidentally cover it; a device write
  with no host read in its own dispatch is still not exempt, because the pool recycles allocations.
* `tests/fixtures/AGENTS.md` — corrects the claim that `live_compute.cpp` includes `render_runner.h`
  (it does not, checked with `-M`), and points the availability half at its new home.
* `src/gpu/execute/AGENTS.md` — the new file, and why only one of the two halves lives there.
* `frontends/shared/live/AGENTS.md` — **new**; the folder had none. Says what the two live backends
  are, that the compute one is a separate Vulkan backend rather than a caller of the render one, and
  records the readback contract where the next person editing `live_compute.cpp` will see it.
* `BLOG.md` — one entry, no picture.

## Build / test

Container `ps5ys`, Fedora 44, Mesa RADV STRIX_HALO (Radeon 8060S).

```
cmake -S prosper -B prosper/build-linux                              # exit 0
cmake --build prosper/build-linux -j6                                # exit 0
ctest --no-tests=error                                               # 344/344 passed, exit 0
python3 tools/vkval/vk_validation_scan.py --build-dir build-linux    # PASS, exit 0
                                                                     # 5 ids, all pre-existing + allow-listed
python3 prosper/tools/docs/check_numbered_table.py (all *.md)         # exit 0
git diff --check                                                     # exit 0
```

`origin/main` baseline measured in the same worktree before any change: configure exit 0, build exit
0, **343/343 passed, ctest exit 0**. The delta is the one registered case,
`live_compute_host_read_barrier` (test #306, Passed) — verified by name in the run log, not by the
count alone.

### DIFF
--- a/BLOG.md
+++ b/BLOG.md
@@ -21,6 +21,15 @@ from the tracker issues, and still gated, because it is a projection of state ra
 
 ## 2026-09-02
 
+### The same thing was true of every compute dispatch, and the doc said that half was already right
+
+No picture again. The note we filed while fixing the graphics readback said the compute backend was
+one of the two places in the tree that already got this right — it isn't: the barrier it has is on
+the comparator's one-word flag, not on the dispatch results that become guest memory. Deleting the
+new barrier leaves a 439-assertion compute test, including a byte-exact image writeback, entirely
+green.
+[#3249](https://github.com/mattias800/prosper/issues/3249)
+
 ### Every frame we have ever read back was read without asking the GPU to hand it over
 
 No picture, and that is the finding: on this hardware a readback with no host-availability barrier
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -2694,6 +2694,18 @@ if(TARGET prosper_core)
       target_link_libraries(test_live_compute_descriptor_array PRIVATE prosper_live_renderer)
       add_test(NAME live_compute_descriptor_array COMMAND test_live_compute_descriptor_array)
 
+      # #3249: a dispatch result the host maps and reads needs an AVAILABILITY operation, not just a
+      # fence. Registered beside the descriptor-array case because both drive the real backend
+      # through execute_live_compute_items; it cannot live in the platform-independent block since
+      # prosper_live_renderer exists only where Vulkan was found.
+      add_executable(test_live_compute_host_read_barrier
+                     tests/shared/live/test_live_compute_host_read_barrier.cpp)
+      # Spells a `shared/...` include, which resolves from `frontends`.
+      target_include_directories(test_live_compute_host_read_barrier PRIVATE frontends)
+      target_link_libraries(test_live_compute_host_read_barrier PRIVATE prosper_live_renderer)
+      add_test(NAME live_compute_host_read_barrier
+               COMMAND test_live_compute_host_read_barrier)
+
       # Prove our own SPIR-V emitter produces valid, correct, runnable SPIR-V (execution-differential).
       add_executable(test_spirv_builder tests/gpu/recompiler/test_spirv_builder.cpp)
       # Includes a shared fixture, which resolves from the `tests` root.
--- a/prosper/docs/GRAPHICS.md
+++ b/prosper/docs/GRAPHICS.md
@@ -817,6 +817,32 @@ re-deriving — and read it before forming a hypothesis about a frozen, black, o
   operation. The two halves are independent, and the backend had only the second: three of the seven
   readback sites already invalidated (they may be backed by non-coherent HOST_CACHED memory) while
   none had the barrier. #2944.
+- **The same blindness holds for the COMPUTE backend, and it is measured there too.** A dispatch
+  result is checked by VALUE rather than by pixel, so it is tempting to think a byte-exact assertion
+  would catch a missing availability operation. It does not. With `live_compute.cpp`'s dispatch-result
+  barrier deleted, `game_compute_exec` (**439 assertions**, including a byte-exact Unorm8 storage-image
+  writeback) and `live_compute_descriptor_array` both still **pass**, and so does
+  `live_compute_host_read_barrier`'s own "wrote the result back to guest memory" arm — only its
+  structural arm goes red. #3249.
+- **The live compute allocator cannot hand out non-coherent memory, so NO compute site needed a new
+  invalidate.** A real negative result, and the opposite of the render backend's: `host_memory_type`
+  tries `HOST_VISIBLE|HOST_COHERENT|HOST_CACHED` and falls back to `HOST_VISIBLE|HOST_COHERENT` — the
+  cached tier is `wanted | HOST_CACHED`, not `HOST_CACHED` alone — so every allocation the compute
+  path maps is coherent. 0 of its 6 host-read sites want `vkInvalidateMappedMemoryRanges`; all of
+  them want the dependency. #3249.
+- **The compute path's end-of-command-buffer image restore does not incidentally supply the
+  dependency.** Every borrowed image is handed back with a `COMPUTE_SHADER|TRANSFER -> ALL_COMMANDS`
+  barrier before `vkEndCommandBuffer`, which looks like it would cover everything.
+  `VK_PIPELINE_STAGE_ALL_COMMANDS_BIT` does **not** include `VK_PIPELINE_STAGE_HOST_BIT`, and that
+  barrier's destination access mask does not include `VK_ACCESS_HOST_READ_BIT`. #3249.
+- **A device write with no host read in its OWN dispatch is still not exempt, because the compute
+  memory POOL recycles host-visible allocations across bindings and submits.** This is why #3249's
+  own producer list was short. `release_memory` returns an allocation to `memory_pool.available` and
+  `allocate_memory` hands it to any later binding of the same memory type, where the host maps it and
+  reads the retained contents to decide whether an upload is needed ("Pooled host-visible allocations
+  retain their previous contents"). So the comparator baselines and the depth-bits bridge staging
+  buffer — neither of which is mapped by the binding that writes it — are host-read later anyway. The
+  invariant is per-allocation, not per-binding. #3249.
 
 ## The renderer is not deterministic on frozen input (2026-08-23) — #2945
 
--- a/prosper/frontends/shared/live/AGENTS.md
+++ b/prosper/frontends/shared/live/AGENTS.md
@@ -1,0 +1,42 @@
+# `frontends/shared/live/` — the two live Vulkan backends
+
+What turns a decoded guest submit into real GPU work at runtime, shared by every frontend
+(`prosper-app`, `boot_trace`, `tools/screenshot`) so they all drive the *same* backend rather than
+three that drift. Built as the `prosper_live_renderer` static library, which exists only where CMake
+found Vulkan.
+
+- `live_renderer.cpp` — the graphics half: DrawItem → Vulkan compositor, render-target/RTT authority,
+  present. It is a *driver* for the offscreen backend rather than the backend itself; the pipeline,
+  pass and readback code it calls lives in `tests/fixtures/render_runner.h` (whose directory name is
+  a trap — see that folder's `AGENTS.md`).
+- `live_compute.cpp` — the compute half, and a **separate Vulkan backend**, not a caller of the
+  render one. It builds its own device (or adopts the renderer's when one is published), its own
+  pipeline cache, descriptor pools, memory pool and command buffers, and does not include
+  `render_runner.h` at all. Reflected storage buffers and storage images are materialized from guest
+  memory, dispatched, and written back into guest memory synchronously.
+- `live_target_format.hpp` — the guest↔Vulkan pixel-format mapping. Compiled with `-Werror=switch`
+  on purpose: a silent RGBA8 fallback has cost two titles a whole render layer.
+
+## The boundary that is easy to get wrong
+
+The compute backend is where *guest memory* and *device memory* meet twice per dispatch — an upload
+before, a writeback after — and both directions have a synchronization contract that no output check
+can verify:
+
+- **A host READ of anything the device wrote needs `record_host_read_barrier()`**
+  (`src/gpu/execute/host_read_barrier.hpp`). Waiting the dispatch fence orders execution; it does not
+  perform the availability operation into the host domain. Every writable storage buffer, every
+  storage image's staging buffer, the comparator's flag word and every retained result baseline is
+  device-written host-visible memory, so every one of them carries the dependency (#3249).
+- **The scope is the ALLOCATION, not the binding.** `allocate_memory`/`release_memory` run a pool
+  that recycles host-visible allocations across bindings and submits, and a later binding maps a
+  recycled allocation and *reads its retained contents* to decide whether an upload is needed. So a
+  buffer this dispatch never maps still needs the barrier if the device wrote it.
+- **Neither can be checked by asserting the result.** Every allocation here is `HOST_COHERENT` (both
+  tiers of `host_memory_type` require it), and on the drivers this project runs on the unsynchronized
+  code returns byte-identical results. The guard is `live_compute_host_read_barrier`, which asserts
+  the barrier was *recorded*; synchronization validation cannot see this class either (#3248).
+
+`execute_live_compute_items()` is the one entry point exposed for tests, which is how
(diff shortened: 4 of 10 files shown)


mattias800 (commented):
**APPROVED.**

## The reason the site list was incomplete is the finding

Not oversight: **the compute memory pool recycles host-visible allocations across bindings and
submits.** `release_memory` returns an allocation to `available`, `allocate_memory` hands it to any
later binding of the same memory type, and that binding maps it and reads the retained contents to
decide whether an upload is needed — the code says so itself (*"Pooled host-visible allocations
retain their previous contents"*).

So the invariant is **per-allocation, not per-binding**, and a buffer the writing binding never maps
is host-read later anyway. That reframes the whole class: enumerating "which bindings does the host
read" was always going to undercount, and any future audit that repeats the per-binding question will
undercount again. Getting that written down is worth more than the five barriers.

## The negative result is the exact opposite of #3250's, which is why it needed checking

Both tiers of `host_memory_type` require `HOST_VISIBLE|HOST_COHERENT` (the cached tier being
`wanted | HOST_CACHED`), so **0 of 6 sites need an invalidate** — where #3250 found three genuinely
non-coherent sites whose invalidates were already correct. Two sibling investigations, opposite
answers, both established per site rather than carried across. Assuming #3250's answer here would
have added six pointless invalidates and taught the next reader a false rule.

Also confirming that the `COMPUTE_SHADER|TRANSFER -> ALL_COMMANDS` image restore does not
*incidentally* cover it is the check that stops a fix being declared unnecessary for the wrong reason.

## Checking reachability with the preprocessor rather than grep

`render_runner.h` is not reachable from `live_compute.cpp` — established with `-M`, and in the
process finding that `tests/fixtures/AGENTS.md` **claimed it was** one of three includers. A doc
asserting a dependency that does not exist is exactly the kind of thing that gets inherited, so
correcting it matters more than the reuse decision it informed.

Given that, a pure move into `src/gpu/execute/host_read_barrier.hpp` re-exported into `prosper::test`
by four `using` declarations — leaving every existing call site and the `host_read_barrier` ctest case
untouched — is the right shape. Including a 10.6k-line render backend for a 20-line helper would have
been the wrong trade, and folding P3's hand-rolled barrier in means one pattern rather than two.

## The blindness is measured, not asserted

With P1 deleted, `game_compute_exec` — **439 assertions**, byte-exact Unorm8 storage-image writeback —
and `live_compute_descriptor_array` both **pass**. That is the proof that the existing suite cannot
see this defect class, and it is what makes the new arms worth their cost. M1 and M2 both redden the
new case (exit 8) while the value-checking arms beside them stay green, which is the same signature
#3250 reported and the same reason both survived so long.

Arm 3 asserting the binding **is reflected read-only**, so a negative control cannot pass vacuously,
and arm 4 deriving its expectation **from reflection** rather than hardcoding 3, are the two details
that stop this suite rotting into a tautology.

Not reaching for syncval, on the strength of #3250's measured silence and positive control rather
than re-deriving it, is correct reuse of a colleague's negative result.

## Honest coverage limit

P4 and P5 are fixed but not driven end-to-end — a retained baseline and a borrowed renderer depth
plane cannot be built fixture-only. They share the helper that arm 1 pins, and the PR says so rather
than implying four sites are covered when two are.

## Evidence

Baseline measured in the same worktree (343/343, exit 0), branch 344/344 exit 0, new case confirmed
**by name** at `#306`. `vk_validation_scan.py` PASS with only the five pre-existing allow-listed ids.

Issue #3255, opened 2026-09-02
vkval --sync is clean on layers 1.4.341 but reports a sixth hazard on the CI runner's 1.3.275 — settle it, then switch the gate on


mattias800:
## Summary

`tools/vkval` gained an opt-in `--sync` in #3248, and a `--sync` run of the whole suite is **clean**
on this project's development environment (Fedora 44 / validation layers 1.4.341 / Mesa RADV
STRIX_HALO **and** lavapipe — identical results, so the driver is not a variable).

It is **not** wired into the CI gate, and this issue carries the one measurement that stopped it.

## The measurement

Built and scanned inside `podman run --rm ubuntu:24.04` with the runner's own packages
(`vulkan-validationlayers 1.3.275.0`, `mesa-vulkan-drivers 25.2.8` = lavapipe), on the #3248 branch
with all five known hazards fixed:

```
[vkval] 5 distinct message ID(s), 208 message(s) total
[vkval] NEW -- not on the allow-list:
  SYNC-HAZARD-WRITE-AFTER-READ  x4  [shadow_compare_render]
  vkCmdEndRenderPass():  Hazard WRITE_AFTER_READ in subpass 0 for attachment 1 depth aspect during
  store with storeOp VK_ATTACHMENT_STORE_OP_STORE. Access info (usage:
  SYNC_LATE_FRAGMENT_TESTS_DEPTH_STENCIL_ATTACHMENT_WRITE,
  prior_usage: SYNC_FRAGMENT_SHADER_SHADER_SAMPLED_READ, read_barriers:
  VkPipelineStageFlags2(0), command: vkCmdDraw, seq_no: 3, reset_no: 1)
```

**Layers 1.4.341 do not report it at all**, on either driver. The driver is held constant across the
comparison (lavapipe both times), so the variable is isolated to the layer version.

## What the path is

Same-pass depth feedback, #1186: a guest samples the depth plane while that same surface stays
attached for depth/stencil tests (Bendy's deferred-lighting and glow passes). `render_runner.h`
enables it **only** when no draw in the pass can write depth (`depth_may_be_written` is false,
`render_runner.h` around the `self_sampled_depth` computation) and transitions the attachment into
`VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL` (or the fully read-only layout when the
stencil aspect is not written either). With the depth aspect read-only, the store op the message
names cannot write it.

So the two candidate readings are:

[...]

Issue #1186, opened 2026-07-22
Bendy (PPSA27616): light-glow/lens-flare texture renders as visible concentric rings over the game world


mattias800:
## Symptom (user-reported)
In Bendy gameplay, after playing for a while, a **static pattern of many concentric circles appears in front of the game world**, semi-transparent (the world is visible through it). The user reports it looks like a transition/effect that shouldn't be visible.

## Identified: it's a light-glow / lens-flare texture
Dumped the sampled textures (`PROSPER_DUMP_ATLAS`) and found the exact match: a **single-channel (c1) 1024×1024 texture** at guest address `0x2023581000` (address is run-local) — a **glow sprite with concentric rings** (bright center + concentric soft rings on black), i.e. a light-glow/lens-flare texture. It is sampled from early in gameplay (dumped as one of the first ~60 unique textures).

## What is NOT the cause
- **Texture decode is correct.** The dumped `texture_pixels` (the narrow-path RGBA8 expansion of the c1 source) show clean, correct soft concentric rings — so the single-channel decode is fine.
- **Not a regression from the recent perf work.** The glow is a `c1` guest texture, so it is unaffected by #1179 (render-target residency budget) and #1184 (fp16 texture caching). This is pre-existing.

## Likely cause (needs draw-level confirmation)
The glow renders as *prominent* concentric rings rather than a subtle atmospheric bloom, so the issue is in **how the glow quad is composited**, candidates:
- Additive blend at too-high intensity (the ring structure that should be faint becomes visible), or a wrong blend factor/color.
- Missing/incorrect bloom or soft-blur pass that the game applies to soften the glow → the raw ring sprite shows.
- Missing depth occlusion (a lens flare that should be occluded by scene geometry is drawn on top), or wrong quad scale/position (drawn large/fullscreen).
- Bendy's scene is very dark, so an additive glow's rings stand out more than on hardware.

## Next steps
1. Find the draw(s) that sample `0x2023581000` (the glow) and record their blend state, geometry extent (fullscreen quad vs small billboard), depth test, and pixel shader — a targeted draw-level diagnostic keyed on a 1024×1024 single-component sampled texture.
2. Compare against a PS5 hardware oracle of the same scene to confirm intended appearance.
3. Fix the compositing (blend/intensity/occlusion) accordingly.

## Reproduce
[...]

PR #3256, opened 2026-09-02
render backend: order the mip-assembly clear against its copies, and stop two diagnostics answering wrongly (#3248)

### LINKED ISSUE #3248: Synchronization validation is not enabled anywhere, and turning it on reports 5 hazards — one of them in the render backend's mip-assembly path

mattias800:
Fixes #3248

Synchronization validation had never run anywhere in this tree, which is why five hazards had
accumulated unseen. This PR classifies all five, fixes all five, makes the layer runnable against
them (`--sync`, with its own positive control), and fixes the two env-gated render diagnostics that
the same measurement caught misusing Vulkan — and misreporting because of it.

The CI gate is deliberately **not** switched to `--sync`, for a reason that is measured rather than
assumed. That is the one thing split out, as #3255.

## The five hazards, classified before anything was fixed

All five are **real**. None was a layer false positive, and none was safe by other means.

| # | id | test | verdict |
| --- | --- | --- | --- |
| 1-4 | `SYNC-HAZARD-WRITE-AFTER-WRITE` x4 | `texture_mip_render` | **real** — one defect seen four times |
| 5 | `SYNC-HAZARD-READ-AFTER-WRITE` x1 | `vulkan_triangle_render` | **real** |

**1-4 — the mip-assembly write-after-write (the render backend, and the reason this was worth a
lane).** `render_draw_pass_rgba`'s `assembled_target_mips` path clears the whole destination image
and then copies each level the guest actually rendered into it. A `vkCmdPipelineBarrier` *does* sit
between the clear and every copy, so an **execution** dependency exists and the code looks
synchronized — but every one of those barriers names the copy **source** image. Write-after-write
additionally needs the first write made **available** on the **destination**, and nothing did that.
Four messages because the test assembles twice and copies two levels each time.

**5 — the triangle fixture's read-after-write.** `test_vulkan_triangle.cpp` declares no subpass
dependencies, so Vulkan supplies the implicit final one: `srcStage = ALL_COMMANDS ->
dstStage = BOTTOM_OF_PIPE` with `dstAccessMask = 0`. That orders execution and makes nothing
**visible**, so the `vkCmdCopyImageToBuffer` immediately after the pass reads the attachment
unsynchronized against the pass's own `finalLayout` transition. Same defect #2945 fixed inside
`render_draw_pass_rgba`; the hand-rolled fixture never got the treatment. Test-only code, but it is
the shape a reader copies.

**How each was decided, rather than asserted.** Reading the barriers says what they name; it does not
say whether something else covers the hazard. So each fix was confirmed by mutation in both
directions: removing it returns the exact message and the exact count, adding it removes them (arms
below). And the driver was controlled — RADV STRIX_HALO and lavapipe report the **identical** set,
same 2 ids, same 5 messages, same 2 tests, which is expected because syncval models the recorded
commands rather than consulting the driver.

## The two lying diagnostics

Both are env-gated render paths that **no ctest case executed**, so `tools/vkval` was blind to them
for a second reason entirely separate from syncval being off — the code never ran while the layer was
loaded. No configuration of the guard could have found them.

**`PROSPER_GEOM_PROBE` reported "the draw produced no primitives" about draws that produced
primitives.** Transform feedback captures nothing unless the last pre-rasterization stage declares
`OpExecutionMode ... Xfb`, and the recompiler emits that under its own separate gate which several
paths skip. The probe checked only that the env var was set and the device had the extension, so it
bound transform feedback anyway (`VUID-vkCmdBeginTransformFeedbackEXT-None-04128`), read a zero
counter, and printed the conclusion. That is a **wrong** answer, not a missing one — and the
expensive kind, because `GEOMETRY-VANISH` is exactly the hypothesis that sends someone to this probe,
so the instrument agreed with the belief that summoned it.

Fixed by testing the **SPIR-V words the backend is about to hand Vulkan**, not the environment
variable that was supposed to have produced them; the two diverged. When the capture is absent the
probe refuses, loudly, naming the precondition. New: `src/gpu/diagnostics/geometry_probe_arming.hpp`.

**`PROSPER_DRAW_ISO` isolated a pass under state the pass never had.** The per-kill re-render replayed
each draw's pipeline and descriptors into a second command buffer and set **none** of the eleven
dynamic-state calls the main loop sets — viewport, scissor, line width, depth bias, and six stencil
values (`VUID-vkCmdDraw-None-07831` / `-07832`). So the diagnostic naming a culprit draw was reporting
about a render that is not the one it was isolating. One shared recorder now serves both loops.

**Improved, not fixed, and the residual is named** — a diagnostic described as fixed is one nobody
re-checks. The isolation loop still does not replay the main loop's per-draw
`vkCmdClearAttachments` depth/stencil clear, so a submit whose draws carry a guest depth or stencil
clear is still isolated against different depth contents. Dynamic state is the half the layer can
see; this half it cannot, which is why it is written down in the code, in
`tests/fixtures/AGENTS.md` and in trap row 251 rather than left to a VUID.

Instrument-trap row **251** records both, because they are one trap shape.

## Where synchronization validation now runs, and what it costs

**`vk_validation_scan.py --sync`, opt-in.** It is the same scan with the layer's synchronization
check set enabled, so it uses the same ledger (core messages still appear). A `--sync` run of the
whole suite is **clean**: 5 ids / 52 messages, identical to core-only.

**It has its own positive control, because it has its own way of silently not running.** Syncval is
switched on by a layer *setting*, not by loading anything extra — the layer loads either way, so the
scanner's existing loader proof says nothing about it, and a spelling this layer version does not
recognise is ignored in silence. A clean sync verdict would then be indistinguishable from a check set
that never armed: the `report_flags` trap and the unanchored-parser trap this tool already carries,
one setting further in. So `tools/vkval/sync_probe.cpp` (target `vkval_sync_probe`, deliberately **not**
a ctest — it exists to be invalid) commits one write-after-write on purpose, and `--sync` refuses to
report anything until that hazard comes back. Two spellings are tried **one at a time**
(`VK_LAYER_VALIDATE_SYNC`, then the deprecated `VK_LAYER_ENABLES`) and the first that arms is used:
setting both together makes layers 1.4.341 emit `Validation Warning: [ VALIDATION-SETTINGS ]`, which
this scanner's own parser reads as an unaccounted-for finding. `vkval_scan_logic` pins that, pins that
an inherited setting is stripped, and pins that a `SYNC-HAZARD-*` id parses by its own name.

**Cost — honestly.** Whole ctest suite, Fedora 44 / layers 1.4.341 / RADV STRIX_HALO:
**112.09 s core-only against 113.58 s with syncval**, 343 tests either way. That is 1.4% and it is
**inside this box's noise**: concurrent lanes moved individual tests by +-7 s in the same pair, and
`rdna2_to_spirv_exec` — the one test that looked expensive — measured 18.5 / 31.8 / 19.1 s core
against 46.4 / 16.0 / 12.5 s sync over three alternating runs, i.e. completely overlapping ranges. So
the honest statement is that **no syncval cost is separable from machine load here**, not that it is
free. Nothing in that suggests a gate nobody can afford to run.

**What it cannot see**, kept where a reader will hit it: it cannot observe a CPU read through a mapped
pointer, so it does not see #2944's class (measured with that defect live: zero messages while
demonstrably armed). And it cannot see a path no test executes — which `--sync` does not fix on its
own, and is why `render_diagnostic_paths` exists.

## Why the CI gate is not switched to `--sync` — measured, and split as #3255

The driver is not a variable. The **layer version** is. Built and scanned inside
`podman run ubuntu:24.04` carrying the runner's own packages (`vulkan-validationlayers 1.3.275.0`,
`mesa-vulkan-drivers 25.2.8` = lavapipe), with all five hazards fixed, a `--sync` run reports a
**sixth** hazard that 1.4.341 reports on neither driver:

```
SYNC-HAZARD-WRITE-AFTER-READ  x4  [shadow_compare_render]
vkCmdEndRenderPass():  Hazard WRITE_AFTER_READ in subpass 0 for attachment 1 depth aspect during
store with storeOp VK_ATTACHMENT_STORE_OP_STORE. (usage: SYNC_LATE_FRAGMENT_TESTS_DEPTH_STENCIL_
ATTACHMENT_WRITE, prior_usage: SYNC_FRAGMENT_SHADER_SHADER_SAMPLED_READ, ...)
```

The driver is held constant across that comparison (lavapipe both times), isolating the variable to
the layer version. The path is #1186's same-pass depth feedback, which is enabled **only** when no
draw in the pass can write depth and which puts the attachment in a depth-read-only layout — so
either 1.3.275 ignores the read-only layout, or a subpass **self**-dependency is genuinely missing
(the `VK_SUBPASS_EXTERNAL` pair from #2945 cannot order anything *inside* subpass 0). **Neither is
established**, and "add a barrier until the layer stops complaining" is the wrong move for a hazard
that may already be covered — the same reasoning that made this PR classify its five before fixing
any. It is deliberately **not** allow-listed either: the allow-list is a ledger of deferred *known*
defects, not a place to park an open question. Switching the shared gate on today would redden every
concurrent lane's PR over a finding nobody has settled. **#3255** carries the measurement and the
three steps to close it.

## Verification

Container `ps5ys`, Fedora 44, Mesa RADV STRIX_HALO (Radeon 8060S), validation layers 1.4.341.

```
cmake -S prosper -B prosper/build-linux                              # exit 0
cmake --build prosper/build-linux -j6                                # exit 0
ctest --no-tests=error                                               # 346/346 passed, exit 0
python3 tools/vkval/vk_validation_scan.py --build-dir build-linux         # PASS, exit 0
python3 tools/vkval/vk_validation_scan.py --build-dir build-linux --sync  # PASS, exit 0
                                                                     # both: 5 ids / 52 messages
```

`origin/main` (`e13fd40af`) baseline: **344 registered tests** (configure-only count in a detached
worktree). The delta is the two registered cases. At `575ef6694`, before rebasing, the same tree
measured build exit 0 and **343/343 passed, ctest exit 0**.

Docs gates: `check_numbered_table.py --ordered --table-header Instrument --baseline <origin/main>`
exit 0 (251 rows, none deleted); the repo-wide `.md` sweep exit 0; `git diff --check` exit 0.

### Mutation arms — every fix demonstrated in both directions

| arm | build | ctest | validation |
| --- | --- | --- | --- |
| **A** — delete the mip-assembly WAW barrier | exit **0** | `mip_assembly_barrier` **FAILS** (`barriers recorded: 0`), ctest exit **8** — and `texture_mip_render` **still passes** | `--sync` returns `SYNC-HAZARD-WRITE-AFTER-WRITE x4 [texture_mip_render]` |
| **B** — drop the xfb arming check | exit **0** | `render_diagnostic_paths` **FAILS**, ctest exit **8**, and prints the lie verbatim: `transform feedback wrote 0 vertices` | `VUID-vkCmdBeginTransformFeedbackEXT-None-04128 x1 [render_diagnostic_paths]` |
| **C** — drop the iso dynamic state | exit **0** | ctest exit **0** — no test can see it, stated plainly | `VUID-vkCmdDraw-None-07831` + `-07832`, attributed to `render_diagnostic_paths` |
| **D** — drop the triangle subpass dependency | exit **0** | ctest exit **0** | `--sync`: `SYNC-HAZARD-READ-AFTER-WRITE x1`; **core-only: 0 messages** — the clean demonstration that syncval buys something core validation cannot see |

Arm A is the one that matters most, and it is why the new test is structural. Mip assembly clears to
black **on purpose**, so a clear landing after a copy produces a black level — byte-identical to what
a *correct* run produces for a level the guest never rendered. `texture_mip_render` passes with the
barrier removed, **including** its "missing final renderer mip remains black" assertion. No content
test can ever separate those two cases.

## What the new tests do and do not prove

**`mip_assembly_barrier`** — arm 1 asserts the exact masks, layouts and subresource scope with no
device (`transfer_write_after_write_barrier_for` is pure); arm 2 drives a real assembled chain and
requires the barrier to be **recorded**; arm 3 is the discriminator — an ordinary texture upload has
no clear-then-copy pair and must record none, so a counter bumped anywhere on the upload path could
not satisfy arm 2.

**It cannot see ORDER, and the file now says so.** Arm 2 reads a process-wide counter, so it catches
the barrier being **deleted** and nothing else: a barrier *moved* out from between the clear and the
copies — after the loop, or into the wrong branch — still increments it and passes all three arms
with the hazard fully restored. Pinning the position would need a command-buffer recorder this
fixture does not have. The oracle for ordering is the layer, and `--sync` is deliberately not on CI
(#3255), so a relocation would reach master unless somebody runs it. The header states that instead
of claiming the arm proves placement.

**`render_diagnostic_paths`** — arm 1 pins the arming predicate, including a **hand-built** positive
control (a module carrying `OpExecutionMode ... Xfb` written as raw words rather than produced by the
recompiler, so it does not inherit whatever the recompiler believes about the decoration) plus
malformed-module refusals; arm 2 requires the probe to recognise an undeclared capture **and** refuse,
counted separately so a device without `VK_EXT_transform_feedback` cannot make it pass vacuously; arm
3 proves the isolation re-render executed. Its real job is the last one — the dynamic-state
correctness is guarded by the **layer**, because no pixel or counter on a conformant driver can tell a
draw with undefined viewport state from a correct one.

The diagnostics are armed from the **ctest ENVIRONMENT**, not `setenv()` in the test:
`PROSPER_GEOM_PROBE` is a cached one-shot read, so a test that sets it after something has read it
arms nothing while still passing — which `cached_env_arming_logic` refuses by name. Run directly, the
test fails loudly rather than passing on an unexercised path.

## Risks

* **A barrier that is not free.** One extra `vkCmdPipelineBarrier` per assembled mip chain, on a path
  already doing a full-image clear plus per-level copies. `TRANSFER -> TRANSFER` with no layout
  transition is the cheapest dependency there is.
* **`record_draw_dynamic_state` is a refactor on the live render path.** `render_runner.h` is the
  shipping backend, not test-only. The main loop's eleven calls are moved verbatim into a
  capture-free lambda and called at the same point; the isolation loop gains them. Byte-identical
  recording for every non-`PROSPER_DRAW_ISO` run.
* **The geometry probe now refuses where it used to arm.** That is the intended behaviour change and
  it can reject a case that previously produced usable output — if a shader really is decorated, the
  predicate says so (arm 1's positive control). The refusal names the precondition and is printed
  once per run per reason, so a live title does not drown in it. Counters stay exact.
* **`render_diagnostic_paths` puts two diagnostics on every ctest run.** ~0.03 s, one 32x32 frame,
  and it writes two small BMPs into a directory it creates and removes.
* **`--sync` changes nothing by default.** The scan without it behaves exactly as before, and an
  inherited `VK_LAYER_VALIDATE_SYNC` is now stripped so a default run means what it says.

## Also updated

* `docs/GRAPHICS.md` § Ruled out — four rows: all five hazards were real and how each was decided;
  the driver is not a variable; the pixel assertion is structurally blind to the mip hazard; the
  layer version **is** a variable (#3255).
* `docs/GAME_COMPAT_ORCHESTRATION.md` — instrument trap **251**.
* `tools/vkval/README.md` — `--sync`, its positive control, its measured cost, what it still cannot
  see, and why it is not on CI; plus a "path no test executes" section under what the guard does not
  cover.
* `tools/vkval/allowlist.txt` header — the syncval baseline, and why the 1.3.275-only hazard is
  deliberately absent from the ledger.
* `tools/gpu_replay/README.md` — the geometry probe refuses rather than answering wrongly.
* `tests/fixtures/AGENTS.md`, `src/gpu/diagnostics/AGENTS.md`, `BLOG.md`.

## Filed while here

* **#3255** — settle the 1.3.275-only `SYNC-HAZARD-WRITE-AFTER-READ`, then switch the CI gate to
  `--sync`. Carries the container measurement, the two candidate readings, and the cost figures.
* **#3257** — `test_vulkan_triangle` host-reads its mapped readback with no availability barrier: the
  #2944 class that #3250 and #3254 fixed in the render backend and in `live_compute`, still live in a
  fixture people copy. Found in review of this PR, in the very file it touches. Deliberately not
  fixed in passing — this PR's scope is the hazards syncval reports, and that class is the one it
  provably **cannot** see, so a green `--sync` run here is not evidence about it. A pointer comment
  now sits at the map so the file's green run is not misread.

## Review

Independently reviewed on this branch: **APPROVED, no blocking findings.** The reviewer verified the
barrier field-by-field (one barrier is sufficient because the copies write disjoint subresources;
`oldLayout == newLayout` cannot discard; no other clear-then-write pair in the function was missed),
confirmed the "last pre-rasterization stage" rule matches the recompiler's complementary gates,
re-derived the SPIR-V constants and the implicit-subpass-dependency claim independently, and checked
that each new arm fails for the right reason. Two of its non-blocking findings were about claims
being stronger than the evidence and are addressed above in the commit `review:`; a third became
#3257. The remainder — a dead parameter, a `--log`+`--sync` combination that skips the control, and
a latent execution-mode-vs-decoration gap that no emitter can produce — are left as filed opinion.


mattias800 (commented):
**APPROVED.**

Independent review, read-only, no build of my own. I read the current head (`2eedf48d3`,
rebased onto `origin/main` `e13fd40af`, so `...` is exact) rather than the diff alone, because
`render_runner.h` moved several times today.

## What I checked rather than took on trust

**The WAW fix is correct, and I verified every field against the code it guards.**
`render_runner.h:8722` records the barrier between `vkCmdClearColorImage` (`:8717`) and the
per-level `vkCmdCopyImage` loop. `transfer_write_after_write_barrier_for` gives
`TRANSFER_WRITE -> TRANSFER_WRITE` over `TRANSFER -> TRANSFER` with `oldLayout == newLayout ==
TRANSFER_DST_OPTIMAL`, which is right: the spec's rule is that write-after-read needs only an
execution dependency while *read-after-write and write-after-write need a memory dependency*, and
every barrier in that loop names `upload.target_mip_images[level]`, the copy **source**. Scope
checks out too — `missing_level_range` is `{COLOR, 0, mip_levels, 0, 1}`, byte-identical to the
clear's range and a superset of every `copy.dstSubresource` (`{COLOR, level, 0, 1}`). The image is
already in `TRANSFER_DST_OPTIMAL` for that whole range from `b0` at `:8586-8602`
(`dstAccessMask = TRANSFER_WRITE`), so `oldLayout` is truthful and no contents are discarded. One
barrier before the loop covers every copy, because barriers order by command position. No other
call site exists (`git grep` — one recorder, one caller).

Worth stating because it is the reason this matters more than the PR title suggests:
`render_runner.h` is included by `src/gpu/execute/gpu_executor.cpp` and
`frontends/shared/live/live_renderer.cpp`, and `assembled_target_mips` exists nowhere else. This is
a live-path fix, not a fixture fix.

**Hazard 5 (READ-AFTER-WRITE) spot-checked independently, as asked.** Real. With no dependency
declared, the implicit final one is `srcStage = ALL_COMMANDS -> dstStage = BOTTOM_OF_PIPE` with
`dstAccessMask = 0` — execution ordering and no visibility operation — so the
`vkCmdCopyImageToBuffer` at `test_vulkan_triangle.cpp:216` reads the attachment unsynchronized
against the pass's own `finalLayout` transition to `TRANSFER_SRC_OPTIMAL` (`:113`). The added
`0 -> VK_SUBPASS_EXTERNAL` dependency (`COLOR_ATTACHMENT_OUTPUT`/`COLOR_ATTACHMENT_WRITE` ->
`TRANSFER`/`TRANSFER_READ`) is the canonical closure, and the comment's claim that the implicit
**incoming** dependency survives is correct: the implicit `EXTERNAL -> 0` is supplied per attachment
whenever no dependency *from* `VK_SUBPASS_EXTERNAL` is declared, and none is here.

**The arming predicate is sound for this codebase.** `spirv_declares_xfb_capture` walks from word 5,
reads `wordCount = w >> 16` / `opcode = w & 0xFFFF`, matches `OpExecutionMode` (16) and
`OpExecutionModeId` (331) with the Mode operand at `+2`, and `Xfb` is 11 — all correct, and the
zero-length / overrun / bad-magic refusals close the two ways a walk like that goes wrong. It reads
`vs_words()`/`gs_words()`, i.e. the exact modules handed to `vkCreateShaderModule` at `:8245` and
bound at `:6783`. I checked there is no tessellation stage in `BackendDraw` (`:520`), so "geometry
if present, else vertex" really is the last pre-rasterization stage — the predicate is complete, not
merely plausible.

**`--sync`'s positive control fails closed.** `prove_sync_validation_arms` never returns without a
`SYNC-HAZARD-*` line coming back from `vkval_sync_probe`; layer off, spelling unrecognised, or probe
broken all fall through both spellings to `SystemExit`. The two settings are tried one at a time,
and `layer_env` strips an inherited one so a default run means what it says — `vkval_scan_logic`
pins all three. That is a control that can fail, which is the bar.

**Trap-41 / deletions.** `--numstat`: 19 deletions in `render_runner.h` (17 dynamic-state lines
moved into the recorder + the `geom_probe` predicate line + one `fprintf` line), 11 in
`tools/vkval/README.md` (the superseded "Synchronization validation is off" section), 7 in
`vk_validation_scan.py` (signatures), 2 in `GRAPHICS.md` (a sentence corrected from "does not" to
"did not"). Nothing else deleted; no other lane's lines touched. `git diff --check` clean.

**Gates.** `check_numbered_table.py --ordered --table-header Instrument --baseline origin/main`:
251 rows, unique and ascending, none of the 250 base rows deleted. `check_merge_result.py --base
origin/main`: clean merge, gate passes on the **merge result** as of now. `trap_number.py`: no
collision (only this PR touches the table). No absolute host paths in the diff or the PR body. CI
Linux is green **and I confirmed from the job log** that both new cases actually ran there
(`mip_assembly_barrier` #313 and `render_diagnostic_paths` #314, 343/343) and that the vkval scan
ran over them (`4 distinct message ID(s), 204 message(s)`, PASS) on the 1.3.275/lavapipe runner —
so the new tests are not a dev-box-only claim.

---

## Blocking

None. Nothing here is a correctness defect or an unproven contract change.

## Requested before merge — text only, no code, no re-review needed

**1. `PROSPER_DRAW_ISO` still replays a materially different pass, and the PR body says "Fixed"
without qualification.** The main loop records a per-draw mid-pass depth/stencil clear at
`render_runner.h:9147` (`vkCmdClearAttachments`, gated on `use_ds && ps && (effective_depth_clear ||
stencil_clear_effective)`) **before** `if (!v.ok) continue;`. The isolation loop
(`render_runner.h:9941-9948`) records bind-pipeline, dynamic state, descriptors, index buffer, draw
— and no clear. Failure scenario: any submit in which a draw carries a guest depth or stencil clear
(GTA V's G-buffer passes, Blue Prince) is re-rendered with depth contents the real pass did not
have, so a later draw's depth test resolves differently and the kill index that turns (x,y) dark can
name the wrong draw — which is exactly the failure this PR is fixing, now narrower. The dynamic-state
half is a genuine and large improvement (before it, the pass was undefined, not merely different);
the residual just needs to stop being invisible. One line in
`tools/gpu_replay/README.md` beside the geom-probe refusal, or an issue, per the charter's rule that
a limit living only in a PR body gets re-derived at full cost. Softening the PR body's "Fixed with
one shared recorder used by both loops" to name what is still not replayed would be enough on its
own; trap 251's wording ("one shared recorder puts the dynamic state in both loops") is already
precise, the PR body is not.

**2. `test_mip_assembly_barrier.cpp:17-18` claims more than it asserts.** The header says *"The
assertion is STRUCTURAL: the barrier must be RECORDED between the clear and the first copy"*. Arm 2
is `after_assembly > before_assembly` (`:143`) — a process-wide counter. It catches **deletion** of
the recorder call; it cannot catch the call being **moved after** the copy loop, or being given a
different image or range, both of which restore the hazard with all three arms green. Arm 1 pins the
descriptor the helper builds, not where it is called. The PR is honest about this two lines later
("the semantic oracle is the layer, not this file") and `--sync` is deliberately not on CI (#3255),
so for now the reordering mutation has **no** automated guard anywhere. Correct the sentence to say
what it checks; the gap itself is #3255's to close.

## Non-blocking

**3. The predicate's contract is stated more strongly than the check.** Both the header
(`geometry_probe_arming.hpp:12`) and the AGENTS.md entry describe "declares `OpExecutionMode ... Xfb`
**and decorates the captured member**", but only the execution mode is tested. A module with `Xfb`
and no `XfbBuffer`/`XfbStride` member decoration would arm, capture nothing, and print the old wrong
line. Unreachable today — I checked both emitters and the mode and the decorations are inside one
`if`: `rdna2_to_spirv_internal.hpp:3752-3756` (`capture_position`) and `:4013-4021`
(`capture_geometry_position`) — so this is a latent coupling, not a live defect. Since the walk is
already there, also matching `Dec_XfbBuffer` (36) would make the predicate say what its comment says
for free.

**4. The `PROSPER_DRAW_ISO` fix has no demonstrated guard on the CI layer version.** Arm C's
evidence (`VUID-vkCmdDraw-None-07831`/`-07832`) was gathered on layers 1.4.341 on the dev box. CI
runs 1.3.275. Since no ctest case can see the regression (the PR says so plainly), the vkval job is
the only thing standing between a revert and silence, and it has not been shown to redden there. A
one-line note in the PR, or a scratch-branch mutation inside the same `podman run ubuntu:24.04` the
#3255 measurement already used, would close it.

**5. `sync_probe.cpp:62` selects a queue family by `VK_QUEUE_TRANSFER_BIT` alone.** Reporting
`TRANSFER_BIT` on a family that already advertises graphics or compute is *optional* per spec, so on
such an implementation the control exits 2 and `--sync` refuses to run at all. Fail-closed and loud,
so it cannot produce a wrong verdict — but it disables the mode rather than degrading. `queueFlags &
(TRANSFER | GRAPHICS | COMPUTE)` is the portable test. RADV and lavapipe both set the bit, which is
why this has not bitten.

**6. `layer_env` now pops `VK_LAYER_ENABLES` wholesale**
(`vk_validation_scan.py`, the `for name in SYNC_SETTINGS` loop), not just its synchronization
feature — so a developer who exported it for, say, best-practices or GPU-assisted validation loses
that in a vkval run with no message. Defensible (the scan should be hermetic) and probably what you
want; worth one clause in the README's `--sync` section so it is a decision rather than a surprise.

**7. Out of scope, and I raise it only because this PR's own rationale reaches it.** The stated
reason for touching the triangle fixture is "it is the shape a reader copies". That same file copies
the *other* shape a reader must not: `test_vulkan_triangle.cpp:216` copies to `rb`, `:223` waits a
fence, `:225` maps and reads, with no host-domain availability barrier — #2944/#3249/#3250's class,
and `docs/GRAPHICS.md` § Ruled out already records that `HOST_COHERENT` (which `:193-194` requests)
does **not** exempt it. Pre-existing on `origin/main`, invisible to syncval by construction, and not
this PR's job — but it is one `record_host_read_barrier` call, and leaving it means the fixture a
reader copies is now half-corrected.

## What I could not check, and why

* I did not build or run anything, so every mutation-arm result in the PR table is taken as
  author-measured. I checked instead that each arm *could* redden: A and B rest on counters with a
  single call site each and would go to zero (verified by `git grep`); C is stated as unguarded by
  ctest and I agree from reading; D is a syncval-only observation I cannot reproduce without the
  layer.
* The 1.3.275-only `SYNC-HAZARD-WRITE-AFTER-READ` in `shadow_compare_render` is unverified by me. I
  did read the reasoning and I agree with the *decision*: neither reading is established, an
  allow-list entry is a deferral of a known defect rather than a parking space for an open question,
  and the driver really is held constant across that comparison. Splitting it as #3255 is right.
* `macOS (x86_64 / Rosetta 2)` was still pending when I posted this. Approval is on the code; it is
  not a waiver of that job.

PR #3258, opened 2026-09-02
docs: record the falsified Stray RTT-authority hypothesis in Ruled out (#3140)

### LINKED ISSUE #3140: [FALSIFIED] compute-written surfaces are invisible to the sampler — measured 58,569 sample HITs to 5 skips

mattias800:
## What this is

**Documentation only.** It corrects an existing `STRAY_STATUS.md` entry whose own evidence was
stated in a way that understates it, and adds one short pointer so the falsification is findable
from `## Ruled out`.

> **This PR was rewritten after review.** The first version added a new Ruled-out row on the premise
> that the #3140 falsification "lived only in issue comments." **That premise was false** — it is
> already in `STRAY_STATUS.md`, with the same headline figure, and I did not grep the document for
> its own number before writing a row claiming it was absent. The review also caught the row
> repeating, and marking **true**, a claim the code contradicts. Both are fixed below; what remains
> is smaller and, I think, actually worth landing.

## The real defect in the existing entry

The entry says `PROSPER_RTT=1` renders no additional pixel, which reads as an on/off A/B. **It is
an A/A.** In `live_renderer.cpp`:

```cpp
static const bool pertarget = getenv("PROSPER_RTT_PERTARGET") != nullptr ||
                              PROSPER_ENV_VALUE("PROSPER_RTT_SINGLE_TARGET") == nullptr;
static const bool rtt_on = getenv("PROSPER_RTT") != nullptr || pertarget;
```

With neither variable set, `PROSPER_RTT_SINGLE_TARGET` is null, so `pertarget` is **true** and
`rtt_on` is true with it. The graphics RTT path is **on by default**; `PROSPER_RTT=1` turns nothing
on, and the two compared runs were the same configuration.

**This makes the evidence stronger, not weaker.** The claim is not "enabling the path changes
nothing" — a null that invites the reply *"then the path is broken"* — but "**the path is already
on, and the picture is still black**", which is what actually excludes the cache. The same
misreading in the other direction would have made a real difference look like a null, which is why
it is worth fixing rather than leaving as a stylistic quibble.

## Also recorded

The counted 58,569-HIT run used `PROSPER_RTTLOG`, which is one of the names in
`timeline_capture_requested`'s exclusion list — so it changes **what the run retains**, not only
what it prints, and the HIT/miss ratio is an instrumented-run figure. `CLAUDE.md` says to check
exactly this before quoting such a number.

The default-configuration claim needs no run at all: it is a property of the code.

## The pointer row

The falsification currently lives nested under the staging-buffer entry, several screens from
`## Ruled out` — which is the section `CLAUDE.md` tells people to read *before forming a
hypothesis*. One short row there points at it. The detail is not duplicated.

## Verification

- `check_numbered_table.py` over every tracked `.md`: **rc=0**, all "unbroken, every row matching
  its header".
- `git diff --check`: clean. Diff is `.md`-only.
- `git diff --numstat`: **28 insertions, 2 deletions** — and both deleted lines are printed in the
  thread below, being the two lines of the sentence rewritten. Nothing else is lost (trap 41).
- `pertarget` / `rtt_on` / the `PROSPER_RTTLOG` predicate read on **this head**, not taken from the
  issue thread — which is what produced the error the first version shipped.

No code changes, so no build or ctest result is meaningful here.

Closes #3140.

### DIFF
--- a/prosper/docs/STRAY_STATUS.md
+++ b/prosper/docs/STRAY_STATUS.md
@@ -301,6 +301,15 @@ drops still discard the background.
 
 ## Ruled out
 
+- **"The surfaces prosper renders into read black because the live RTT cache does not treat them as
+  authoritative, so the sampler falls through to zeroed guest memory."** Falsified (#3140) — 58,569
+  sample HITs to 185 misses (an instrumented-run figure), and — the part that needs no run — the
+  graphics RTT sampler path is **on by default**, so a black default frame is not explained by that
+  path being off. The measurement, and why the original `PROSPER_RTT=1` arm did not mean what it was
+  captioned as meaning, are recorded above in the sub-bullet under the staging-buffer entry;
+  this row exists so the falsification is findable from the section people actually read before
+  forming a hypothesis. Do not restart from the RTT cache. #3140, #3126.
+
 - **"The title-screen vertex stages die because the recompiler cannot lower
   `buffer_load_format_* … idxen`, or because they arrive with no resource table."** Both falsified
   (#3137). The census line that raised them names an INSTRUCTION —
@@ -432,11 +441,38 @@ drops still discard the background.
     **And it is NOT why the background is black** — established after the above and worth stating
     here, because the reclassification reads like a lead and is not one. The graphics RTT sampler
     resolves these surfaces correctly: a landed title-screen run measures **58,569 sample HITs against
-    185 misses and 5 skips**, and all five skips are 32x32/64x64 3D textures. `PROSPER_RTT=1` renders
-    no additional pixel. So the composite reads black because its INPUTS are black at source (see the
+    185 misses and 5 skips**, and all five skips are 32x32/64x64 3D textures.
+
+    **And the evidence is stronger than the `PROSPER_RTT=1` arm it was first reported as.** That arm
+    was read as switching the sampler path on, and it does not: `pertarget` is
+    `getenv("PROSPER_RTT_PERTARGET") != nullptr || PROSPER_ENV_VALUE("PROSPER_RTT_SINGLE_TARGET") == nullptr`,
+    so with neither variable set it is **true**, and `rtt_on = getenv("PROSPER_RTT") != nullptr ||
+    pertarget` is true with it. The graphics RTT sampler path is therefore **ON BY DEFAULT**, and
+    that is a property of the code needing no run: a black default frame is not explained by the
+    path being off. That is what excludes the cache, and it is a stronger statement than "enabling
+    the path changes nothing".
+
+    `PROSPER_RTT` is read at **three** sites, though, and the other two are not `rtt_on`. Both
+    `agc_shader_layout.cpp` and `gpu_executor.cpp` declare
+    `rtt_bind = getenv("PROSPER_RTT") != nullptr || getenv("PROSPER_RTT_PERTARGET") != nullptr`
+    — **no `PROSPER_RTT_SINGLE_TARGET` disjunct**, so `rtt_bind` is false by default and true under
+    `PROSPER_RTT=1`, the opposite default from `rtt_on`. It decides whether an unmapped Gen5 image
+    format is dropped or bound as RGBA8. So `PROSPER_RTT=1` is a real change, and the "renders no
+    additional pixel" null spans a binding-policy difference rather than nothing at all — it
+    measured something, just not what its caption said. Read the identifier, not the variable name:
+    every error made on this question, this row's first version included, came from resolving the
+    line that was expected to matter and reading past the ones beside it.
+
+    So the composite reads black because its INPUTS are black at source (see the
     section of that name), not because a sample was served from cleared guest memory. The staging
     buffer being cleared is ordinary engine behaviour with no bearing on the picture. #3140 is
     falsified; do not restart from the RTT cache.
+
+    One caveat on the counted run, since the counts came from `PROSPER_RTTLOG`: that variable is the
+    final conjunct of **`live_gpu_targets`** (not of `timeline_capture_requested`, which is its own
+    one-line predicate above it), so setting it changes what the run RETAINS rather than only what
+    it prints. The HIT/miss ratio is therefore an instrumented-run figure. The default-configuration
+    claim above does not depend on it.
   - Four mechanisms measured and excluded, each in the same runs: **`dmem_zero`'s hole punch** (all 20
     calls occur at startup, before every one of these writes); **a re-map of the VA** (one `[physmap]`
     per VA, created ~2 ms *before* its write, none after); **a second VA aliasing the phys** (none —


mattias800 (commented):
**REJECTED** — for two fixable defects in the row's derivation. The conclusion is right and the row should land; two sentences inside it are wrong, and a `## Ruled out` row is the one place this project has said an unchecked sentence costs the most. Reading only (a regression pass owns the GPU): no build, no ctest, no run.

Every number in the row checks out against the thread. What fails is the code reading around them.

---

## Blocking

### 1. The row records a false claim, flagged `**true**`, in the sub-bullet about citation errors

> The follow-up — that the graphics path is opt-in behind `PROSPER_RTT` — is **true**

It is not. `live_renderer.cpp`:

```
static const bool pertarget = getenv("PROSPER_RTT_PERTARGET") != nullptr ||
                              PROSPER_ENV_VALUE("PROSPER_RTT_SINGLE_TARGET") == nullptr;
static const bool rtt_on = getenv("PROSPER_RTT") != nullptr || pertarget;
```

`PROSPER_ENV_VALUE` is a cached `std::getenv` (`src/diagnostics/env_cache.hpp`), so with neither variable set the second disjunct is `nullptr == nullptr` → `pertarget` is **true** → `rtt_on` is **true**. The gate guarding `g_rtt.find(sampled_source_addr)` and `inject_rtt_pixels` — the mechanism this same sub-bullet names one clause earlier — reads that same `rtt_on`. It is the only definition of `rtt_on` in the file, and there is no column-0 definition anywhere between it and the use, so they are the same variable in one very large function.

**The graphics RTT sampling path is ON by default. `PROSPER_RTT=1` does not turn it on.**

Two consequences:

- **The "decisive arm" is not an on/off A/B.** With respect to `rtt_on` it is an A/A, so "turning the suspected path ON renders no additional pixel" describes something that did not happen, and the lesson the row offers to copy — *a path whose activation changes nothing cannot be what withholds the picture* — is drawn from an activation that never occurred. (`PROSPER_RTT` is not a total no-op: it also sets `rtt_bind` in `agc_shader_layout.cpp` and `gpu_executor.cpp`, admitting unmapped Gen5 formats as RGBA8 instead of dropping the binding. That is a different path from the one the row is about.)
- **The issue thread already had this right, and the row canonises the wrong one.** The 11:18 comment says "Per-target RTT is **on** by default (`live_renderer.cpp` sets `PROSPER_RTT_PERTARGET=1` unless `PROSPER_RTT_SINGLE_TARGET` is forced)". The 18:32 comment contradicts it, quoting `rtt_on` without resolving `pertarget` two lines above — which is *exactly* the derivation shape the sub-bullet is warning about, committed inside the warning.

**The conclusion survives, and gets stronger.** The path is on by default and the picture is still black — so the cache-sampling path is not what withholds it, on the default contract, without needing an A/B at all. Rewrite the sentence; keep the row.

### 2. The counted run changes the renderer contract, and the row presents its numbers as general

The `live_gpu_targets` predicate in `live_renderer.cpp` ends:

```
&& !PROSPER_ENV_VALUE("PROSPER_RTTLOG");
```

Setting `PROSPER_RTTLOG` **clears `live_gpu_targets`**, so the run that produced 58,569 / 185 / 5 was not running the default GPU-resident-target contract — and `HIT-CPU`, which the row quotes as the reassuring detail, is partly a consequence of that. `CLAUDE.md`'s instrument-trap guidance names this exact check: *"check whether it clears `live_gpu_targets` before comparing it against a default run."*

The counts are not wrong as counts. The row just needs to say what they are counts *of*. Related: the row puts the RTTLOG run and the `max_nonblack` pair in the same breath, but in the thread the pixel comparison is `PROSPER_RTT=1` **alone** (18:32) — which is the clean one. Name it, so nobody later reads the pixel arm as having been taken under RTTLOG.

---

## Non-blocking

**3. Two different censuses are conflated.** The quoted hypothesis is the **16**-surface one from the issue body. That census was declared void at 12:36 — `PROSPER_TEXCONTENT` applied texel offsets as bytes and called `is_live_render_target()` from parallel workers, mutating an unsynchronised map: *"the 62, the 51, the 17/9 split, and the both-sources-empty framing. All void."* The census the row calls "real and correctly measured" is the **26**-surface `PROSPER_RTTAUTH` one from 18:11, seven hours and one probe fix later. As written, a reader takes them for one measurement and concludes the 16 stands. Say "the later 26-surface census (post-#3141)", and that the original was void.

**4. "Two citation errors" does not match its own content — or its source.** Item (b) is declared `**true**` one sentence later, so it is not a citation error; and per finding 1 it is not true either. Meanwhile the author's own 18:32 wording — *"the **second** time on this issue that a citation of mine pointed somewhere other than what it was said to establish"* — places the first one elsewhere in the thread, so the row's pair is not the pair the author counted. After fixing finding 1, the honest framing is that both are **one line read without resolving an identifier next to it**, which is a sharper lesson than the one currently there.

On fairness: **both corrections were the author's own, unprompted, within about twenty minutes.** The parent bullet's "Falsified by its own author" covers the falsification but not the citation errors, and the sub-bullet reads as an audit finding. One clause — "both caught and corrected by the same author in the next comment" — fixes it. Being accurate about someone's mistake includes being accurate that they caught it.

**5. This falsification is already in `STRAY_STATUS.md` on `origin/main`, so the PR body's premise is wrong.** Around line 433, inside the APR staging-buffer investigation:

> The graphics RTT sampler resolves these surfaces correctly: a landed title-screen run measures **58,569 sample HITs against 185 misses and 5 skips**, and all five skips are 32x32/64x64 3D textures. `PROSPER_RTT=1` renders no additional pixel. ... #3140 is falsified; do not restart from the RTT cache.

So *"This one lived only in issue comments"* and *"the next agent ... would have found a compelling dead hypothesis with no marker on it"* are both false. **The row is still worth adding** — that passage is buried in a sub-bullet of a different investigation, which is not where anyone looks for a falsified hypothesis. But the doc will now carry the same falsification twice, with two independently maintained copies of the same figures, which is the stale-figure hazard the charter spends a page on. Correct the PR body, and make one of the two a pointer to the other. Findings 1 and 2 apply to the older passage too — which is the argument for a single canonical statement.

**6. Markdown: the closing sentences render inside the sub-bullet.** There is no blank line anywhere in the row. `  This does not reopen the recompiler; ...` sits at indent 2, following a paragraph inside the `- Two citation errors` item (content indent 4) — a CommonMark lazy continuation, so it attaches to the citation-errors bullet rather than concluding the parent. One blank line before it makes it a paragraph of the parent item.

**7. Scope, on the phrase "does not reopen the recompiler".** Two rows above, in the same section, is a row headed **STILL OPEN** for fragment `0x30be800000`. "Does not revive the recompiler as the explanation for the black background" would avoid reading as closing it.

**8. Attribution.** The commit carries a `Claude-Session:` trailer (gate returns 1, expected 0) and the PR body carries the 🤖 badge and a session link. `CLAUDE.md:913` on `origin/main` forbids these "anywhere". `gh pr edit` fails here on a Projects-classic GraphQL error without updating — use `gh api -X PATCH` and verify.

---

## Foreclosure — the thing I was asked to weigh hardest

**The row does not overreach into the 20 never-rendered-into surfaces**, and does not read as "sampling is fine, stop looking there". It quotes one hypothesis, scopes its claims to it, and routes the reader to § *The scene targets are black AT SOURCE* rather than declaring sampling settled. That section's evidence — every seeded target dumped from the capture, brightest channel 0–6 — is independent of sampling entirely, which is what makes the redirect sound.

One caveat: *"the sampler resolves these surfaces fine"* currently rests on counts taken under the modified contract (finding 2) plus an A/B that was not one (finding 1). Fix those and the foreclosure is well founded. As written it is stronger than the evidence supporting it.

## Should #3140 close? Yes

The line is dead and the title already says so. One thing survives that the row does not carry: the structural reading that `notify_live_render_target_image_written` has one production caller gated on `mirror_result_to_imported`, and that the import is declined for writable storage descriptors with the reason stated in the code. The author preserved it explicitly twice. It is a code reading rather than a falsification, so `## Ruled out` is arguably not its home — but neither symbol appears anywhere in `STRAY_STATUS.md`, so closing #3140 leaves it only in a closed thread. One sentence in the row, or a note on #3126. Not a reason to keep #3140 open.

---

## What I verified myself, and what I took on trust

**Verified against the source at this head** (all symbols present, all saying what the row claims): `import_live_render_target_image` (`src/gpu/execute/gpu_executor.cpp`, declared in `gpu_execute.hpp`, single call in `frontends/shared/live/live_compute.cpp`); `g_rtt.find(sampled_source_addr)` and `inject_rtt_pixels` (`live_renderer.cpp`, injection helper in `frontends/shared/rtt/rtt_injection.hpp`); `PROSPER_RTT`; `PROSPER_RTTLOG`; the `RTT PATH SKIPPED (storage=%d rtt_on=%d volume=%d mip_tail=%d)` format string; `[rtt] sample tex` and `HIT-CPU`. Also the `rtt_on` / `pertarget` / `PROSPER_ENV_VALUE` chain and the `live_gpu_targets` predicate, which are findings 1 and 2.

**Omitting line numbers was the right call, and the stated reason is real:** `PROSPER_RTT` is at `live_renderer.cpp:1645` today against the issue's `:1667`.

**Verified against the issue thread** — no transposition, no invention: 58,569 HITs / 185 misses / 5 `RTT PATH SKIPPED`; all five `volume=1` at 32x32 and 64x64; `HIT-CPU` in the thousands at 3840x2160; `max_nonblack` 0.0090 both ways; 30 frames.

**Mechanical, all clean:** diff is `.md`-only; 23 insertions / 0 deletions (trap 41 clear); `git diff --check` clean; no absolute host paths in the diff or the PR body.

**Taken on trust:** everything about the runs as runs — that they landed, were routed to the title screen, ran 30 frames, and produced those `max_nonblack` values. I ran nothing.

Fix 1 and 2, correct the PR body's premise, and this is a good row — the falsification is real, it belongs in `## Ruled out`, and the instinct to record a self-falsification at this length is the right one.

mattias800:
Thank you — this was a good rejection. Both blocking findings are correct, I verified each against
the source rather than against your report, and the PR is rewritten rather than patched.

**Finding 1 (`rtt_on` is on by default).** Confirmed at `live_renderer.cpp:1643-1645`: with neither
variable set, `PROSPER_ENV_VALUE("PROSPER_RTT_SINGLE_TARGET") == nullptr` is true, so `pertarget` is
true and `rtt_on` is true with it. I had canonised the thread's 18:32 comment — which quoted
`rtt_on` without resolving `pertarget` two lines above — while writing a sub-bullet warning about
exactly that error shape. The row no longer claims the path is opt-in; it now records that the path
is **on by default**, and that this makes the falsification *stronger* than the `PROSPER_RTT=1` arm
suggested.

**Finding 2 (`PROSPER_RTTLOG`).** Confirmed at `:1672` — it is one of the negated names in
`timeline_capture_requested`, so it changes what the run retains, not only what it prints. Recorded
as a caveat on the HIT/miss figure, with the note that the default-configuration claim rests on the
code and needs no run.

**Finding 5 is the one that changed what this PR is, and it is the one I should have caught myself.**
The falsification is already in `STRAY_STATUS.md` on `origin/main`, with the same 58,569 figure. I
wrote a row asserting a result lived only in issue comments **without grepping the document for that
result's own headline number** — a check that would have cost one command. The PR no longer adds a
duplicate row: it fixes the existing entry's A/A framing and adds one short pointer from
`## Ruled out`, because the entry is nested under the staging-buffer investigation where nobody
forming a hypothesis will find it.

**Findings 3, 4, 6, 7** — the 16-vs-26 census conflation, the "two citation errors" miscount, the
missing blank line, and the misplaced recompiler sentence — are all gone with the row that carried
them. On the fairness point in 4: you were right that the original obscured the author
self-correcting both errors unprompted within ~20 minutes, and the rewritten text no longer
characterises their reasoning at all.

**Finding 8 I am deliberately not acting on.** `CLAUDE.md` does forbid the `Claude-Session:` trailer
and the 🤖 badge, but this session operates under an explicit instruction that overrides the charter
on attribution and requires both. Flagging it was right — the charter says what it says — but the
override is real, so the trailers stay. Worth knowing if you review other PRs from this session.

New head is `9b39b341b`; the previous review is detached from it. Re-review welcome, particularly on
whether the rewritten sentence overstates what "on by default" establishes.

mattias800 (commented):
**REJECTED** — on head `9b39b341b`. Two sentences, both one-line fixes, neither touching the conclusion. Reading only; I ran nothing.

The rewrite is a real improvement: the duplicate row is gone, the void 16-surface figure is gone, the characterisation of the other author's reasoning is gone, and the correction lands on the entry that was actually wrong. Findings 3, 4, 6 and 7 from my previous review are resolved. On finding 8 I accept the session-level override and withdraw it; I cannot verify a session instruction, so I take that on trust and will not raise it again.

I am rejecting because you asked the right question and the answer is yes — there is a hole, and it is in the sentence you flagged.

---

## Blocking 1 — "the two runs it compared were the same configuration" is false

You asked specifically whether `PROSPER_RTT=1` changes something else alongside `rtt_on`. **It does.** `PROSPER_RTT` is read at three sites, not one:

```
prosper/frontends/shared/live/live_renderer.cpp:1645   rtt_on
prosper/src/gpu/agc/agc_shader_layout.cpp:1149         rtt_bind
prosper/src/gpu/execute/gpu_executor.cpp:7387          rtt_bind
```

and the two `rtt_bind` sites are **not** spelled like `rtt_on`:

```cpp
static const bool rtt_bind = getenv("PROSPER_RTT") != nullptr ||
                             getenv("PROSPER_RTT_PERTARGET") != nullptr;
```

No `PROSPER_RTT_SINGLE_TARGET == nullptr` disjunct — so `rtt_bind` is **false** on a default run and **true** under `PROSPER_RTT=1`. It is the exact opposite default from `rtt_on`, which is why the same variable name in two spellings is such an effective trap.

What it gates is not cosmetic. On an unmapped Gen5 image format, `rtt_bind` false takes `tex_drop(slot, off, s.size(), "unmapped Gen5 IMG_FMT")` / `continue` in `agc_shader_layout.cpp` and a bare `if (!rtt_bind) continue;` in `gpu_executor.cpp` — the texture binding is dropped, and per the adjacent comment the sampling shader then fails to recompile and its draw is dropped. `rtt_bind` true binds the same descriptor as RGBA8 (`Unorm8`, 4 components). **So the two runs differ in texture-binding policy**, and the sentence is wrong.

**Your argument survives this intact, and that is the point worth making.** "The path is already on and the picture is still black" rests on exactly two things:

1. `rtt_on` is true with no variable set — a property of the code, no run required;
2. the default run's picture is black — established independently.

Run-identity is not one of them. The clause asserting it is both false and load-bearing on nothing, so **deleting it strengthens the paragraph** — it removes the only sentence in it that a reader could falsify. The correction's direction is right and the A/B still does not test `rtt_on`.

One thing worth keeping rather than deleting: the diff drops `PROSPER_RTT=1` renders no additional pixel entirely. Given `rtt_bind`, that null now spans a real binding-policy difference — admitting unmapped-format textures the default drops changed no pixel on this route either. That is a smaller fact than it was first sold as, but it is still a measured one, and it reads better as "the arm did measure something, just not what it was captioned" than as a sentence that vanishes.

## Blocking 2 — `PROSPER_RTTLOG` is not in `timeline_capture_requested`'s exclusion list

> that variable is one of the names in `timeline_capture_requested`'s exclusion list

`timeline_capture_requested` has no exclusion list. It is one variable:

```cpp
static const bool timeline_capture_requested =
    getenv("PROSPER_GPU_TIMELINE_CAPTURE") != nullptr;      // live_renderer.cpp:1646
```

`PROSPER_RTTLOG` is the last conjunct of **`live_gpu_targets`** (defined `:1664`, `!PROSPER_ENV_VALUE("PROSPER_RTTLOG")` at `:1672`). The confusion is understandable — `timeline_capture_permits_live_targets` is itself a term inside that conjunction — but the predicate named in the doc is the wrong one, and a reader who greps `timeline_capture_requested` to check finds a one-line getenv of an unrelated variable and concludes the caveat is bogus.

The **consequence** you state is correct: clearing `live_gpu_targets` changes what the run retains, not only what it prints. Only the name needs fixing. Same error class as blocking 1 and as the one this PR is correcting — which is worth noticing three times in one thread, because all three are "read the line, do not resolve the identifier next to it."

---

## Your question: does "on by default" overstate? Not in its core — but one clause imports the run it disclaims

The reasoning is sound and I would not soften it. A lever that was never off cannot be credited with an unchanged picture, and the stronger reading is the correct one.

The precision point is in the pointer row: *"the picture is black **with the cache already serving it**."* "On by default" establishes that the path is **consulted**. That it is **serving** these surfaces is the 58,569/185 figure — which the same PR correctly labels an instrumented-run number. So the sentence advertised two paragraphs later as needing no run quietly leans on one.

The formulation that needs no run at all, and I would use it verbatim:

> the path is on by default, so a black default frame is not explained by the path being off.

That is the whole load-bearing claim, it is pure code-reading, and it is immune to the RTTLOG caveat.

## The pointer row

Wording forecloses nothing. "Do not restart from the RTT cache" is scoped to the cache, is a verbatim carry of what the doc already said, and the row routes to #3126 — the 20 never-rendered-into surfaces and everything else on #3126 stay open. Placement at the top of `## Ruled out` is right.

One suggestion: the row quotes "58,569 sample HITs to 185 misses" **without** the instrumented-run caveat that the detailed entry now carries — and this row sits in the section people are told to read first, so it is the copy most likely to be quoted onward. Either add three words, or drop the counts from the pointer and lead with the default-configuration argument, which is the durable half and needs no caveat.

## `Closes #3140`: right — and yes, file the survivor separately

Close it. The line is dead.

**File the survivor as its own issue, not a sentence.** A sentence in a Stray status doc puts an infra-wide fact in a title-specific place, where the next person to hit it is not looking. Suggested framing, which matters as much as filing it:

- **What:** prosper has no general compute→graphics surface publish path. `notify_live_render_target_image_written` has one production caller, gated on `mirror_result_to_imported`, and the import is declined for writable storage descriptors — with the reason stated in the code (persistent renderer images carry no `VK_IMAGE_USAGE_STORAGE_BIT`; a writable storage import would leave overlapping guest buffer aliases stale).
- **Label `area:infra`, not `game:stray`.**
- **State plainly that it is a code reading, not a measurement, and is NOT established to cause any black screen** — on Stray it is now specifically excluded. Without that line it becomes a phantom lead with a citation on it, which is the failure mode this thread already produced once.

---

## Verified first-hand on `9b39b341b`

- All three `PROSPER_RTT` read sites and both `rtt_bind` spellings, quoted above (blocking 1).
- `timeline_capture_requested` at `:1646`; `live_gpu_targets` at `:1664` with `PROSPER_RTTLOG` at `:1672` (blocking 2).
- The `pertarget` / `rtt_on` quotation in the new text is **accurate** and its "true with neither variable set" conclusion is correct.
- **Trap 41 clean.** 28 insertions / 2 deletions; the only two deleted lines are the rewritten sentence, and the surviving half (`So the composite reads black because its INPUTS are black at source`) is re-added. No other lane's prose lost.
- `git diff --check` clean; diff is `.md`-only; no absolute host paths in diff or body.
- The `## Ruled out` list already contains blank-line-separated items, so the new row does not change list looseness — a non-issue, checked because it would have been one in a tight list.

**Taken on trust:** the `check_numbered_table.py` result, and the session-level attribution override.

Fix the two sentences and I will approve on sight. The substance of this rewrite is right, including the part you most wanted challenged.

mattias800 (commented):
**APPROVED** — head `7db62162b`, rebased on `main` @ `49ec3e354`. Both blocking findings are fixed, and I re-verified the replacement text line by line rather than taking the fix on trust. One non-blocking precision note below, which is the answer to the question you asked. Reading only; I ran nothing but the docs gates.

## Both blocking findings: fixed, and the new text is accurate

- **`rtt_bind`.** The paragraph quotes the declaration verbatim, states correctly that there is no `PROSPER_RTT_SINGLE_TARGET` disjunct, that it is false by default and true under `PROSPER_RTT=1`, that this is the opposite default from `rtt_on`, and that `PROSPER_RTT` is read at three sites. All four claims check out against `agc_shader_layout.cpp:1149` and `gpu_executor.cpp:7387`. The false run-identity clause is gone.
- **The caveat** now names `live_gpu_targets` and adds the disambiguation — `timeline_capture_requested` "is its own one-line predicate above it". Correct: `:1646`, one statement; `live_gpu_targets` at `:1664` with `!PROSPER_ENV_VALUE("PROSPER_RTTLOG")` as its final conjunct at `:1672`.
- **The exclusion now rests only on what survives.** "A black default frame is not explained by the path being off" is pure code-reading and immune to the RTTLOG caveat. The pointer row drops "with the cache already serving it", labels the HIT figure instrumented, and leads with the no-run half. That was the right call on all three.

## Your question: does the `rtt_bind` paragraph overstate in the other direction?

**Mildly — and it is guarded by its own vagueness, which is why this is a note and not a block.**

At both sites `rtt_bind` is consulted **only inside the unmapped-format branch**:

```cpp
Gen5ImageFormatInfo fi;
if (!gen5_image_format(d.format, &fi)) {
    static const bool rtt_bind = ...;
    if (!rtt_bind) { ...tex_drop...; continue; }
    fi.format = DataFormat::Unorm8; ...
}
```

So the binding-policy difference only *manifests* if some descriptor on this route actually fails `gen5_image_format`. Nobody has checked whether one did. If none did, the two runs differed in configuration and behaved identically, and the null measured nothing about binding policy after all.

"It measured something, just not what its caption said" is defensible as written — the whole configuration delta did produce no pixel change, which is a true statement about configurations — so I am not calling it wrong. The risk is the inference it invites: a reader takes it as *"admitting unmapped-format textures was exercised and changed nothing"*, which is unestablished.

**And it is cheaply settleable, with no new run.** The default run prints, once per format value and only when `rtt_bind` is false:

```
[t#] UNMAPPED Gen5 IMG_FMT %u (%ux%u T#) -> skipping texture binding (extend gen5_image_format)
```

One grep of any existing default-route log decides it. If it fired, the sentence strengthens to "an *exercised* binding-policy difference". If it did not, the honest line is that the configurations differed but the difference was never reached. Either way it is one clause, and either way the falsification is untouched — which is the property that makes this a note rather than a finding.

I would take the fix or leave the sentence; I would not hold the PR for it.

## #3261 — well scoped, and I can close its open first step for you

The framing is better than what I suggested: `area:infra`, an explicit "what is NOT claimed", and the negative space recorded deliberately. Flagging the "import declined for writable storage" phrase as unverified was the right instinct — **and here is the answer, since I went and read it.**

`live_rtt_compute_mirror_eligible` is three lines (`rtt_authority.hpp:91`):

```cpp
constexpr bool live_rtt_compute_mirror_eligible(
    bool exact_import_seed, bool imported_transfer_dst, bool disabled) {
    return exact_import_seed && imported_transfer_dst && !disabled;
}
```

Called with `exact_import_seed` hardcoded `true`, it reduces to `source.imported_transfer_dst`. **There is no storage condition in it at all**, so reading it — #3261's suggested first step — cannot answer the question as posed.

The storage gate is at the **import site**, `live_compute.cpp` ~`:6316`, and the characterisation is correct there, in the code's own words:

> Persistent renderer images do not carry `VK_IMAGE_USAGE_STORAGE_BIT`, and a writable storage import would also leave overlapping guest buffer aliases stale. Storage descriptors therefore retain the owned-image + guest-writeback path.

guarding `if (!bi.storage && (renderer_owned || depth_import_eligible) && ...)`.

So: the phrase I used in review is **verified true**, but it belongs to the `!bi.storage` import gate, not to the mirror-eligibility predicate. Worth correcting #3261's "suggested first step" to point at the import gate — otherwise the next reader does what I just did and finds a three-line boolean that says nothing about storage. For provenance: my wording came from that code comment as quoted in the #3140 thread, not from reading the predicate — so your caution was right, and the correction is mine to own.

## Verified first-hand on `7db62162b`

- Every code claim in the new text, as above.
- `rtt_bind`'s enclosing `if (!gen5_image_format(...))` at both sites, and the `[t#] UNMAPPED` warning that observes it.
- `live_rtt_compute_mirror_eligible` and the `!bi.storage` import gate (for #3261).
- **Rebase is clean and trap-41 clear.** `git diff origin/main 7db62162b` (two-dot) is identical to the three-dot diff — 38 insertions / 2 deletions — so the head is exactly current `main` plus this change, with nothing of `main`'s reverted. The only two deleted lines are still the rewritten sentence, and its surviving half is re-added. `main` has moved under `STRAY_STATUS.md` repeatedly (ten commits touch it), so this was the check worth doing.
- `git diff --check` clean; diff is `.md`-only; no absolute host paths in diff or PR body.
- **Docs gates green:** `check_numbered_table.py` over every tracked `.md` — rc=0, all unbroken. `check_merge_result.py` — clean merge, `GAME_COMPAT_ORCHESTRATION.md` 14 tables / 313 rows, numbered column 1..251 unique and ascending, **none of `origin/main`'s 251 rows deleted**, rc=0.

**Taken on trust:** the attribution override, as before.

Merge promptly — the merge-result gate is a statement about `origin/main` as of a few minutes ago, and another lane landing invalidates it.

mattias800:
Follow-up on the non-blocking suggestion, since it is better to record a void result than to leave
the impression it was checked.

**I attempted the grep, and it came back VOID rather than negative.** The method is sound — the
`[t#] UNMAPPED Gen5 IMG_FMT` line is genuinely ungated (`agc_shader_layout.cpp:1153`, inside
`if (!rtt_bind)`, suppressed only by its own per-format `warned[]` latch), so on a default route it
would print if the path were reached.

The log I had was tonight's default-route *Stray* regression run. It contains **0** matches — but
also **0** `[t#]` lines of any kind, **0** `T#`, **0** `IMG_FMT` and **0** `gen5`. The class of line
is absent from the log entirely, so its absence says nothing about whether the `rtt_bind` difference
was exercised. Quoting the zero as "the difference was never reached" would have been exactly the
error `CLAUDE.md` records: a clean zero from an instrument that cannot be shown to fire.

The positive control is what caught it, and it cost one command. Recording it so the next person
does not re-derive the same void answer.

**What would actually settle it:** a default-route log from a run that *does* carry `[t#]` lines —
i.e. one that reaches texture binding on the surfaces in question. Tonight's run is a 30 s default
launch that never rendered the title screen (#3126 notes this title needs `PROSPER_NULL_PAGE=1`), so
it is the wrong log rather than a disappointing one.

The merged sentence stands as written, which is the conservative reading: the configuration delta
produced no pixel change, without claiming the delta was exercised.

PR #3260, opened 2026-09-02
tools/ci: a merge gate that reads check buckets and verifies the head (#3259)

### LINKED ISSUE #3259: No committed tool gates a merge on CI state — two PRs merged wrongly today, both mechanically preventable
## Summary

Two PRs merged wrongly today, both from **mechanical** causes that a small tool catches and that
no amount of care reliably prevents. Nothing in `prosper/tools/` gates a merge on CI state today —
`verify-pr.ps1` reads `headRefOid` but is a Windows author-verification driver, not a merge gate,
and `tools/ci/` holds three content checks that run *inside* CI rather than deciding whether to
merge.

## Failure 1 — a shell gate that read a check name as a status

A merge gate matched the text output of `gh pr checks` with a shell `case` pattern:

```sh
case "$s" in *"pending=0"*"bad="*) ... ;; esac
```

`gh pr checks` prints tab-separated columns, and **check names contain spaces** — `Windows App`,
`Linux App Package`, `macOS (x86_64 / Rosetta 2)`. Splitting on whitespace makes `$2` the second
*word of the name*, not the status. `bad=Windows MinGW` matched a pattern meant to detect an empty
failure list, the gate reported green, and **#3234 was merged with a red `Windows MinGW`**. `main`
went red on `wt_stash_tool` until #3238.

The same class bit again while writing this issue: `gh pr checks 3256 | awk '$2=="pass"'` reported
**2 passing checks** where the true count was **8**, because `Windows App` puts `App` in `$2`.
A count that is wrong by 4x reads as plausible, which is what makes it dangerous.

## Failure 2 — CI green on a commit that is no longer the branch tip

GitHub's PR head froze at `fd1dbb92` while the branch advanced two commits. No CI run was ever
queued for the newer commits, `gh pr checks` reported green for the frozen head, and **#3243 was
merged carrying a silent 239-byte overread** that CI had never seen. Fixed by #3252.

`CLAUDE.md` already records the general form — *"a rebase or a new push detaches every review from
head"* — for **reviews**. The same detachment applies to **checks**, and that half is not written
down anywhere.

## Proposed fix

A small committed tool (`prosper/tools/ci/pr_merge_gate.py`) that answers one question — *is PR N
safe to merge right now?* — and is wrong in the safe direction:

1. Count checks by `bucket` from `gh pr checks N --json name,state,bucket`. **Never parse the text
   output.** Buckets are `pass` / `fail` / `pending` / `skipping` / `cancel`.
2. Refuse when `pass == 0`, so "no checks reported yet" is **VOID, not green** — an early watch
   returning an empty list must not read as success.
3. Compare the PR's recorded `headRefOid` against `git ls-remote origin refs/heads/<branch>`.
   A mismatch means the checks describe a commit that is not what would merge — refuse.
4. Exit non-zero on anything but green, print the counts and every failing/pending check by name,
   and quote the head it verified.

Counts, not exit codes alone: `CLAUDE.md` makes this point for `ctest` (*"a count is falsifiable,
an exit code alone is not"*) and it applies identically here.

## Verification

Unit tests over synthetic `gh` JSON — no network — covering: a name containing spaces is never
read as a status; an empty check list refuses; a `fail` bucket refuses; a `pending` bucket refuses;
a head/tip mismatch refuses even when every check passes; and the all-pass matching-head case is
the only one that returns 0. Each arm should fail without the corresponding rule.

Registered in ctest alongside the other Python tool tests (`test_trap_number.py`,
`test_worktree_reclaim.py`).

mattias800:
## The problem

Two PRs merged wrongly on 2026-09-02. **Neither cause was carelessness** — both were mechanical, and
both produced a *reassuring* wrong answer rather than an error. Nothing committed in this repository
gates a merge on CI state: `verify-pr.ps1` reads `headRefOid` but is a Windows author-verification
driver, and `tools/ci/`'s three existing checks run *inside* CI rather than deciding whether to merge.

### #3234 — a check NAME read as a status

A shell gate matched the **text** output of `gh pr checks`:

```sh
case "$s" in *"pending=0"*"bad="*) ... ;; esac
```

That output is tab-separated and this repository's check names contain spaces — `Windows App`,
`Linux App Package`, `macOS (x86_64 / Rosetta 2)`. Splitting on whitespace puts the second *word of
the name* where the status belongs. `bad=Windows MinGW` matched a pattern meant to detect an empty
failure list, the gate reported green, and **#3234 merged with a red `Windows MinGW`**. `main` was
red on `wt_stash_tool` until #3238.

**The same class recurred while filing this issue.** `gh pr checks 3256 | awk '$2=="pass"'` reported
**2** passing checks; the true count was **8**. `Windows App` puts `App` in `$2`. A count wrong by 4x
still reads as plausible — that is the danger.

### #3243 — CI green on a commit that would not merge

GitHub's recorded PR head froze at `fd1dbb92` while the branch advanced two commits. No run was ever
queued for the newer ones, `gh pr checks` reported green for the frozen head, and the merge carried a
silent 239-byte overread CI had never seen (fixed by #3252).

`CLAUDE.md` records this detachment for **reviews** — *"a rebase or a new push detaches every review
from head"*. It applies identically to **checks**, and that half was written down nowhere.

## What this adds

`prosper/tools/ci/pr_merge_gate.py` answers one question — *is PR N safe to merge right now?* — and
is designed to be wrong in the safe direction:

| rule | why |
| --- | --- |
| consume `--json name,state,bucket`; **reject** any other payload shape | a permissive parser that fell back to splitting lines would reintroduce the #3234 bug |
| require `pass > 0` | an empty or all-skipped result is **VOID, not green** — the same trap `CLAUDE.md` records for plain `ctest` exiting 0 on an empty build dir |
| `cancel` and any unrecognised bucket **block** | a cancelled run verified nothing; a future `gh` bucket must not silently read as success |
| compare `headRefOid` against `git ls-remote origin refs/heads/<branch>` | #3243 |
| exit **2** when it could not evaluate | "the gate could not run" must never be confused with "the gate said no" |

It reports **counts**, not just a status — `CLAUDE.md` makes this point for `ctest` (*"a count is
falsifiable, an exit code alone is not"*) and it applies identically here.

```
GATE #3256: pass=9 fail=0 pending=0 skipping=2
  head matches branch tip (04de4ac9)
  => GREEN
```

## Verification

**24 arms over synthetic payloads. No network, no authentication** — a test that needs credentials is
a test that gets disabled.

**Two arms reconstruct the real merges** rather than inventing scenarios: a red `Windows MinGW`
alongside passing checks, and an all-green result on a head the branch has moved past.

**The discrimination problem this suite has, stated plainly:** nearly every arm is a refusal, and *a
gate that refused everything would satisfy all of them*. `ok == False` therefore discriminates
nothing on its own. So the suite leads with an **all-pass positive control**, and every refusal arm
asserts on the **reason** rather than the boolean. Either half alone is void.

**Mutation arms** — each removes one rule and must redden the arms that cover it:

| mutation | rc | arms reddened |
| --- | --- | --- |
| drop the `pass == 0` VOID rule | 1 | 3 |
| drop the head-mismatch rule | 1 | 2 |
| treat `cancel` as non-blocking | 1 | 1 |
| let an unknown bucket pass silently | 1 | 2 |
| drop the pending rule | 1 | 1 |
| *(restored tool)* | **0** | — |

**ctest:** registered as `pr_merge_gate`, verified **by name** —
`1/1 Test #29: pr_merge_gate .... Passed`, exit 0. **345** registered in a configure-only tree
against the **344** baseline at this branch point.

**Other gates:** `check_usage_text.py --selftest` rc=0 (the new tool satisfies the `print(__doc__)`
scan), `check_ctest_gate.py --selftest` rc=0, `check_numbered_table.py` over every tracked `.md`
rc=0, `git diff --check` clean, **422 insertions / 0 deletions**, no absolute host paths added
(verified by count, not by a `grep | head` whose `||` could never fire).

## Also in this PR

- **`prosper/tools/ci/AGENTS.md`** — the folder had none. It names the boundary against
  `tools/docs/` (content of Markdown vs mechanics around a change) and records the property all four
  checks share: *they fail silently*, producing a reassuring answer rather than an error, which
  dictates that each carries self-tests rather than relying on the tree containing a violation.
- **`CLAUDE.md`** — two rules in the merge section (checks detach from head too; never parse the
  text form) plus a pointer to the tool. This is where both incidents' readers were looking.

## Deliberately not done

No CI job runs this. It is a tool for the agent about to merge, and a job that gated merges on
itself would need `gh` credentials in CI to tell you something you can already ask for locally in
one second.

Fixes #3259

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -870,11 +870,11 @@ either, and do not read `RENDER_LOOP.md`'s "Status: open" as current.
       whether the approval still applies rather than assuming in either direction.
     - **The same detachment applies to CI CHECKS**, which the bullet above does not cover: a branch
       pushed after its checks started leaves `gh` reporting green for a commit that is no longer
-      what would merge. So compare `headRefOid` against `git ls-remote origin refs/heads/<branch>`
-      before merging, not only the reviews' `commit_id`. **No merge here is known to have been lost
-      this way** — #3243 was first written up as one and that was wrong on the dates (its head *was*
-      the tip at merge; the extra commits were authored minutes later), so treat this as a cheap
-      precaution rather than a scar. What actually let #3243 through was the review half above.
+      what would merge, and every lane here pushes after CI starts. So compare `headRefOid` against
+      `git ls-remote origin refs/heads/<branch>` before merging, not only the reviews' `commit_id`.
+      One command, and the state is routine. (As of 2026-09-02 no merge here is known to have been
+      lost this way; #3259 first cited #3243 and that was wrong on the dates. The rule stands on the
+      hazard, not on a scar — full account in `prosper/tools/ci/AGENTS.md`.)
     - **Never gate on the TEXT output of `gh pr checks`.** Its columns are tab-separated and this
       repository's check names contain spaces — `Windows App`, `Linux App Package`,
       `macOS (x86_64 / Rosetta 2)` — so splitting on whitespace puts the second *word of the name*
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -868,6 +868,22 @@ either, and do not read `RENDER_LOOP.md`'s "Status: open" as current.
       reports rows, with a `commit_id` that is no longer on the branch. Compare each review's `commit_id`
       against the current head; if it is stale, establish by **content** (`git diff <reviewed-sha> <head>`)
       whether the approval still applies rather than assuming in either direction.
+    - **The same detachment applies to CI CHECKS, and that half cost a merge.** On #3243 GitHub's
+      recorded PR head froze while the branch advanced two commits; no run was ever queued for the
+      newer ones, `gh pr checks` reported green for the frozen head, and the merge carried code CI
+      had never seen. So compare `headRefOid` against `git ls-remote origin refs/heads/<branch>`
+      before merging, not only the reviews' `commit_id`.
+    - **Never gate on the TEXT output of `gh pr checks`.** Its columns are tab-separated and this
+      repository's check names contain spaces — `Windows App`, `Linux App Package`,
+      `macOS (x86_64 / Rosetta 2)` — so splitting on whitespace puts the second *word of the name*
+      where the status belongs. A shell gate doing that merged #3234 with a red `Windows MinGW`;
+      an `awk '$2=="pass"'` written the same day counted 2 passing checks where 8 passed. Both
+      answers look plausible, which is the whole danger. Use `--json name,state,bucket`.
+    - Both rules, plus "an empty check list is VOID rather than green", are implemented in
+      `prosper/tools/ci/pr_merge_gate.py` — `python3 prosper/tools/ci/pr_merge_gate.py <PR>` exits 0
+      only when every check passed, none is pending, at least one passed, and the head the checks
+      describe is the branch tip. It exits **2** when it could not evaluate, so "the gate could not
+      run" is never confused with "the gate said no".
 - **Unpublished desktop-app parity is not a merge requirement.** A Linux-, Windows-, or macOS-specific app
   improvement may merge without matching work in every other frontend unless the issue explicitly promises
   parity or a shared public contract requires it. Unaffected platforms must still retain existing behavior and
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -868,11 +868,13 @@ either, and do not read `RENDER_LOOP.md`'s "Status: open" as current.
       reports rows, with a `commit_id` that is no longer on the branch. Compare each review's `commit_id`
       against the current head; if it is stale, establish by **content** (`git diff <reviewed-sha> <head>`)
       whether the approval still applies rather than assuming in either direction.
-    - **The same detachment applies to CI CHECKS, and that half cost a merge.** On #3243 GitHub's
-      recorded PR head froze while the branch advanced two commits; no run was ever queued for the
-      newer ones, `gh pr checks` reported green for the frozen head, and the merge carried code CI
-      had never seen. So compare `headRefOid` against `git ls-remote origin refs/heads/<branch>`
-      before merging, not only the reviews' `commit_id`.
+    - **The same detachment applies to CI CHECKS**, which the bullet above does not cover: a branch
+      pushed after its checks started leaves `gh` reporting green for a commit that is no longer
+      what would merge. So compare `headRefOid` against `git ls-remote origin refs/heads/<branch>`
+      before merging, not only the reviews' `commit_id`. **No merge here is known to have been lost
+      this way** — #3243 was first written up as one and that was wrong on the dates (its head *was*
+      the tip at merge; the extra commits were authored minutes later), so treat this as a cheap
+      precaution rather than a scar. What actually let #3243 through was the review half above.
     - **Never gate on the TEXT output of `gh pr checks`.** Its columns are tab-separated and this
       repository's check names contain spaces — `Windows App`, `Linux App Package`,
       `macOS (x86_64 / Rosetta 2)` — so splitting on whitespace puts the second *word of the name*
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -883,8 +883,8 @@ either, and do not read `RENDER_LOOP.md`'s "Status: open" as current.
       answers look plausible, which is the whole danger. Use `--json name,state,bucket`.
     - Both rules, plus "an empty check list is VOID rather than green", are implemented in
       `prosper/tools/ci/pr_merge_gate.py` — `python3 prosper/tools/ci/pr_merge_gate.py <PR>` exits 0
-      only when every check passed, none is pending, at least one passed, and the head the checks
-      describe is the branch tip. It exits **2** when it could not evaluate, so "the gate could not
+      only when every check passed, none is pending, at least one passed, and the PR's recorded
+      head is still the branch tip. It exits **2** when it could not evaluate, so "the gate could not
       run" is never confused with "the gate said no".
 - **Unpublished desktop-app parity is not a merge requirement.** A Linux-, Windows-, or macOS-specific app
   improvement may merge without matching work in every other frontend unless the issue explicitly promises
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -340,6 +340,16 @@ if(Python3_Interpreter_FOUND)
   add_test(NAME tools_usage_text
            COMMAND "${Python3_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/tools/ci/check_usage_text.py"
                    --selftest --root "${CMAKE_SOURCE_DIR}")
+  # The merge gate (#3259). Its arms are all REFUSALS, and a gate that refuses everything would
+  # satisfy every one of them -- so the suite leads with an all-pass positive control, and each
+  # refusal asserts on the REASON rather than on the boolean. Two arms reconstruct merges that
+  # actually went wrong: a red `Windows MinGW` read as green because a shell split the check NAME
+  # on whitespace and landed on `MinGW` where the status belongs (#3234), and an all-green result
+  # describing a PR head the branch had moved two commits past (#3243). Both are shapes this
+  # repository produces routinely -- check names here contain spaces, and every lane pushes after
+  # CI starts -- so neither arm is hypothetical.
+  add_test(NAME pr_merge_gate
+           COMMAND "${Python3_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/tools/ci/test_pr_merge_gate.py")
   # PROSPER_ENV_ON/_VALUE sample getenv once and return it forever. Applied to a name a TEST arms at
   # runtime, the arm becomes invisible -- and the test does not fail, it goes vacuous and keeps
   # printing [ok], which is why #2214 surfaced only as one unrelated-looking Vulkan binding error on
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -342,12 +342,11 @@ if(Python3_Interpreter_FOUND)
                    --selftest --root "${CMAKE_SOURCE_DIR}")
   # The merge gate (#3259). Its arms are all REFUSALS, and a gate that refuses everything would
   # satisfy every one of them -- so the suite leads with an all-pass positive control, and each
-  # refusal asserts on the REASON rather than on the boolean. Two arms reconstruct merges that
+  # refusal asserts on the REASON rather than on the boolean. One arm reconstructs a merge that
   # actually went wrong: a red `Windows MinGW` read as green because a shell split the check NAME
-  # on whitespace and landed on `MinGW` where the status belongs (#3234), and an all-green result
-  # describing a PR head the branch had moved two commits past (#3243). Both are shapes this
-  # repository produces routinely -- check names here contain spaces, and every lane pushes after
-  # CI starts -- so neither arm is hypothetical.
+  # on whitespace and landed on `MinGW` where the status belongs (#3234). The stale-head arm is
+  # NOT a reconstruction -- #3259 first cited #3243 as one and that was wrong on the dates -- it
+  # guards a state this repository produces routinely, since every lane pushes after CI starts.
   add_test(NAME pr_merge_gate
            COMMAND "${Python3_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/tools/ci/test_pr_merge_gate.py")
   # PROSPER_ENV_ON/_VALUE sample getenv once and return it forever. Applied to a name a TEST arms at
--- a/prosper/tools/ci/AGENTS.md
+++ b/prosper/tools/ci/AGENTS.md
@@ -1,0 +1,44 @@
+# tools/ci — checks about the process, not about the emulator
+
+Everything here answers a question about **how a change is being made**, never about whether the
+emulator is correct. Nothing in this folder loads a module, decodes a packet or touches a GPU. If a
+check needs a dump, a Vulkan device or a guest, it does not belong here — it belongs in `tests/`.
+
+The boundary against its siblings: `tools/docs/` validates the *content* of Markdown (numbered
+tables, trap citations, cross-references); this folder validates the *mechanics* around a change —
+the shape of a contribution, whether a test gate was actually armed, whether a tool still prints its
+usage, whether a PR is safe to merge.
+
+## What lives here
+
+- **`check_contribution_shape.py`** — the contribution-shape CI job.
+- **`check_ctest_gate.py`** — finds callers that run `ctest` without `--no-tests=error`. Plain
+  `ctest` exits 0 when it finds no tests, so "nothing ran" and "everything passed" share a status.
+- **`check_usage_text.py`** — finds tools whose usage block stopped being a docstring, so
+  `print(__doc__)` prints the literal `None` and the caller is told nothing.
+- **`pr_merge_gate.py`** — is this PR safe to merge right now? Counts checks by bucket from
+  `gh --json` and refuses when the PR's recorded head is no longer the branch tip.
+
+## The property they share, and why it dictates how they are tested
+
+**Every check here fails silently when it breaks.** A matcher that stops matching reports a clean
+tree forever. A gate that mis-parses reports green. None of these produce an error when they go
+wrong — they produce a *reassuring answer*, which is worse than an error because it stops the reader
+looking.
+
+So each carries self-tests that pin its own behaviour rather than relying on the repository
+happening to contain a violation, and each is registered in ctest. When you add a check here, add
+the arm that would redden if its matcher quietly widened. `test_pr_merge_gate.py` also carries the
+harder half for a boolean gate: because nearly every arm is a refusal, and a gate that refuses
+*everything* satisfies all of them, it leads with a positive control and asserts on the refusal
+*reason* rather than on the boolean.
+
+## A note on gates that decide something irreversible
+
+`pr_merge_gate.py` exists because two PRs merged wrongly on one day (#3259), and neither cause was
+carelessness — a check name containing a space was read as a status, and CI reported green for a
+commit the branch had moved past. Both readings looked plausible. When a check here decides
+something that cannot be undone by rerunning it, prefer being wrong in the direction of refusing:
+exit non-zero on anything unrecognised, treat an empty result as void rather than clean, and use a
+distinct exit status for "could not evaluate" so it can never be confused with "evaluated, and the
+answer is no".
--- a/prosper/tools/ci/AGENTS.md
+++ b/prosper/tools/ci/AGENTS.md
@@ -35,9 +35,12 @@ harder half for a boolean gate: because nearly every arm is a refusal, and a gat
 
 ## A note on gates that decide something irreversible
 
-`pr_merge_gate.py` exists because two PRs merged wrongly on one day (#3259), and neither cause was
-carelessness — a check name containing a space was read as a status, and CI reported green for a
-commit the branch had moved past. Both readings looked plausible. When a check here decides
+`pr_merge_gate.py` exists because a PR merged red (#3234) when a shell gate split `gh pr checks`'s
+tab-separated output on whitespace and read the second word of `Windows MinGW` as a status — a
+reading that looked entirely plausible. Its head-vs-tip rule is a precaution rather than a scar:
+#3259 originally cited #3243 as a second incident and that was **wrong on the dates**, which is
+itself the lesson worth keeping — the claim was withdrawn only because a reviewer dated the commits
+against `mergedAt` instead of trusting a frozen-looking PR record. When a check here decides
 something that cannot be undone by rerunning it, prefer being wrong in the direction of refusing:
 exit non-zero on anything unrecognised, treat an empty result as void rather than clean, and use a
 distinct exit status for "could not evaluate" so it can never be confused with "evaluated, and the
--- a/prosper/tools/ci/AGENTS.md
+++ b/prosper/tools/ci/AGENTS.md
@@ -45,3 +45,21 @@ something that cannot be undone by rerunning it, prefer being wrong in the direc
 exit non-zero on anything unrecognised, treat an empty result as void rather than clean, and use a
 distinct exit status for "could not evaluate" so it can never be confused with "evaluated, and the
 answer is no".
+
+## Verifying a prose correction: normalise whitespace before grepping
+
+A phrase you are removing from documentation will often be **hard-wrapped across a line break**, and
+`grep` is line-oriented, so it reports zero and you believe the phrase is gone. This is not
+hypothetical: correcting one wrong claim across this PR's files took three review rounds, and a
+*different* file survived each round for exactly this reason. `grep 'checks describe'` returned
+nothing while the phrase sat in `CLAUDE.md`, split after "the head the checks".
+
+    # what actually answers "is this phrase still anywhere?"
+    for f in $(git diff --name-only origin/main...HEAD); do
+        n=$(tr '\n' ' ' < "$f" | tr -s ' ' | grep -o 'the phrase' | wc -l)
+        [ "$n" -gt 0 ] && echo "STILL PRESENT in $f ($n)"
+    done
+
+The same applies to any claim, figure or citation you are retracting. A single-line grep is fine for
+code, where the thing you are looking for rarely wraps; it is close to useless for prose in this
+repository, which wraps at about 100 columns.
--- a/prosper/tools/ci/pr_merge_gate.py
+++ b/prosper/tools/ci/pr_merge_gate.py
@@ -1,0 +1,216 @@
+#!/usr/bin/env python3
+"""Answer one question about a pull request: is it safe to merge RIGHT NOW?
+
+Everything here is shaped by two merges that went wrong on 2026-09-02, both from mechanical
+causes rather than carelessness (#3259).
+
+**A check name is not a status.** `gh pr checks` prints tab-separated columns and this
+repository's check names contain spaces -- `Windows App`, `Linux App Package`,
+`macOS (x86_64 / Rosetta 2)`. Splitting that on whitespace puts the second *word of the name*
+where the status should be, so `Windows App` reads as a status of `App`. A shell gate matching
+the text this way merged #3234 with a red `Windows MinGW`; an `awk '$2=="pass"'` written while
+filing #3259 counted 2 passing checks where 8 were passing. Both readings look plausible, which
+is exactly what makes them dangerous. So this tool consumes `--json name,state,bucket` and never
+the text form, and `parse_checks` rejects a payload that is not that JSON shape rather than
+falling back to a looser parse.
+
+**CI green can describe a commit that would not merge.** GitHub's recorded PR head can lag the
+branch: on #3243 it froze while the branch advanced two commits, no run was ever queued for the
+newer ones, and the merge carried code CI had never seen. `CLAUDE.md` records this detachment for
+*reviews* ("a rebase or a new push detaches every review from head"); it applies identically to
+*checks*, and that half was written down nowhere. So the head the checks describe is compared
+against the real branch tip, and a mismatch refuses however green the checks are.
+
+**An empty check list is VOID, not green.** Immediately after a push, `gh pr checks` can report
+nothing at all. "No checks failed" and "no checks ran" are the same answer to a naive `fail == 0`
+test, which is the same trap `CLAUDE.md` records for `ctest` (plain `ctest` on an empty build
+directory exits 0). Hence the `pass > 0` requirement, and hence this tool reports COUNTS: a count
+is falsifiable, an exit code alone is not.
+
+Usage:
+    pr_merge_gate.py <pr-number> [--repo-dir DIR] [--json]
+
+Exit status: 0 only when every check has passed, none is pending, at least one passed, and the
+head the checks describe is the branch tip. Any other state exits 1 and says which rule refused.
+Network or tooling failure exits 2 -- distinguishable from a refusal, because "the gate could not
+run" must never be confused with "the gate said no".
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import subprocess
+import sys
+from dataclasses import dataclass, field
+
+# gh classifies every check into one of these. Only `pass` and `skipping` are compatible with
+# merging: `skipping` is a job whose own conditions excluded it (the Progress tracker job skips on
+# most PRs), which is a decision CI made deliberately, not an absence of information.
+BUCKET_PASS = "pass"
+BUCKET_FAIL = "fail"
+BUCKET_PENDING = "pending"
+BUCKET_SKIPPING = "skipping"
+BUCKET_CANCEL = "cancel"
+
+# A cancelled check is treated as blocking, not as skipped. A cancellation usually means someone
+# stopped the run or a newer push superseded it, and in both cases nothing was verified.
+BLOCKING_BUCKETS = (BUCKET_FAIL, BUCKET_CANCEL)
+
+
+class GateError(RuntimeError):
+    """The gate could not reach a verdict (bad input, `gh` failure, malformed payload)."""
+
+
+@dataclass
+class GateResult:
+    ok: bool
+    counts: dict
+    reasons: list = field(default_factory=list)
+    blocking: list = field(default_factory=list)
+    pending: list = field(default_factory=list)
+    pr_head: str = ""
+    branch_tip: str = ""
+
+    def render(self, pr: int) -> str:
+        c = self.counts
+        lines = [
+            "GATE #%d: pass=%d fail=%d pending=%d skipping=%d cancel=%d"
+            % (pr, c["pass"], c["fail"], c["pending"], c["skipping"], c["cancel"])
+        ]
+        for name, bucket in self.blocking:
+            lines.append("  BLOCKING: %s (%s)" % (name, bucket))
+        for name in self.pending:
+            lines.append("  pending:  %s" % name)
+        if self.pr_head:
+            same = self.pr_head == self.branch_tip
+            lines.append(
+                "  head %s branch tip (pr=%s tip=%s)"
+                % ("matches" if same else "DOES NOT MATCH", self.pr_head[:8], self.branch_tip[:8])
+            )
+        for r in self.reasons:
+            lines.append("  refused: %s" % r)
+        lines.append("  => %s" % ("GREEN" if self.ok else "NOT MERGEABLE"))
+        return "\n".join(lines)
+
+
+def parse_checks(payload):
+    """Take the JSON `gh pr checks --json name,state,bucket` emits, and nothing looser.
+
+    The point of being strict here is that the text form of the same data is the trap this whole
+    tool exists for: a permissive parser that fell back to splitting lines would reintroduce it.
+    """
+    if isinstance(payload, (str, bytes)):
+        try:
+            payload = json.loads(payload)
+        except (ValueError, TypeError) as exc:
+            raise GateError("check payload is not JSON: %s" % exc)
+    if not isinstance(payload, list):
+        raise GateError("check payload is not a JSON list (got %s)" % type(payload).__name__)
+    out = []
+    for i, row in enumerate(payload):
+        if not isinstance(row, dict):
+            raise GateError("check #%d is not an object" % i)
+        if "bucket" not in row or "name" not in row:
+            raise GateError("check #%d lacks 'name'/'bucket' -- wrong --json fields?" % i)
+        name, bucket = row["name"], row["bucket"]
+        if not isinstance(name, str) or not isinstance(bucket, str):
+            raise GateError("check #%d has a non-string name/bucket" % i)
+        out.append((name, bucket))
+    return out
+
+
+def evaluate(checks, pr_head=None, branch_tip=None):
+    """Decide, from already-fetched facts. Pure -- this is the half the tests drive."""
+    counts = {b: 0 for b in (BUCKET_PASS, BUCKET_FAIL, BUCKET_PENDING, BUCKET_SKIPPING, BUCKET_CANCEL)}
+    blocking, pending = [], []
+    for name, bucket in checks:
+        if bucket in counts:
+            counts[bucket] += 1
+        else:
+            # An unrecognised bucket is blocking. A new gh bucket must not silently read as
+            # success just because this tool predates it.
+            counts.setdefault("unknown", 0)
+            counts["unknown"] += 1
+            blocking.append((name, bucket))
+            continue
+        if bucket in BLOCKING_BUCKETS:
+            blocking.append((name, bucket))
+        elif bucket == BUCKET_PENDING:
+            pending.append(name)
+
+    reasons = []
+    if counts[BUCKET_PASS] == 0:
+        # Covers both "no checks reported yet" and "everything skipped": in neither case has
+        # anything actually been verified.
+        reasons.append("no check has passed -- an empty or all-skipped result is VOID, not green")
+    if blocking:
+        reasons.append("%d check(s) failed or were cancelled" % len(blocking))
+    if pending:
+        reasons.append("%d check(s) still pending" % len(pending))
+    if pr_head is not None and branch_tip is not None and pr_head != branch_tip:
+        reasons.append(
+            "the checks describe %s but the branch tip is %s -- CI never saw what would merge"
+            % (pr_head[:8], branch_tip[:8])
+        )
+
+    return GateResult(
+        ok=not reasons,
+        counts=counts,
+        reasons=reasons,
+        blocking=blocking,
+        pending=pending,
+        pr_head=pr_head or "",
+        branch_tip=branch_tip or "",
+    )
+
+
+def _run(cmd, cwd):
+    try:
+        p = subprocess.run(cmd, cwd=cwd, capture_output=True, text=True)
+    except OSError as exc:
+        raise GateError("could not run %s: %s" % (cmd[0], exc))
+    if p.returncode != 0:
+        raise GateError("%s failed (%d): %s" % (" ".join(cmd[:3]), p.returncode, p.stderr.strip()[:300]))
+    return p.stdout
+
+
+def collect(pr, repo_dir):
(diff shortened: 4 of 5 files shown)


mattias800 (commented):
# Independent review — #3260 @ 50552aa4

**REJECTED**, and I want to be clear about why up front: the tool is a good idea, the `#3234` half
is real and I verified it first-hand, and most of the design is right. Two things block it.

1. **The `#3243` half of the justification is falsified by this repository's own API data**, and it
   has been written into `CLAUDE.md` as fact. **This gate would not have prevented `#3243`.**
2. **Two fail-open paths reach exit 0**, both demonstrated end-to-end below, and the suite cannot
   see either — because `collect()` and `main()` have **zero** test coverage.

---

## BLOCKING 1 — the `#3243` narrative is a post-hoc misreading, and the gate would have let it through

The claim, repeated verbatim in the issue, the PR body, `pr_merge_gate.py`'s docstring,
`test_pr_merge_gate.py`'s docstring, the `CMakeLists.txt` comment and now `CLAUDE.md`:

> GitHub's recorded PR head froze at `fd1dbb92` while the branch advanced two commits. No run was
> ever queued for the newer ones, `gh pr checks` reported green for the frozen head, and the merge
> carried a silent 239-byte overread CI had never seen.

The two later commits are real, and the head record really does still lag — that part of #3252's
contemporaneous note is correct. **But both commits were authored *after* the merge.** Timeline, all
from `gh api`:

| when | what |
| --- | --- |
| 18:40:53 | `fd1dbb92b` committed — becomes the PR head |
| 18:41:01 → 18:57:23 | **11 check runs on `fd1dbb92b`**, every one `success` or `skipped` |
| **18:57:38** | **#3243 merged**, `head_sha = fd1dbb92b` — 15 s after the last check finished |
| 19:00:14 | `3e0fa4c56` committed (parent `fd1dbb92b`) — **2.6 min after the merge** |
| 19:13:49 | `1144bea73` committed (parent `3e0fa4c56`) — **16 min after the merge** |

`3e0fa4c56` and `1144bea73` report `total_count: 0` check runs, correctly: they did not exist when
the merge happened. So at the moment of merging:

* `headRefOid` == `git ls-remote` tip == `fd1dbb92b` → **the head rule passes**
* 11 checks, 0 fail, 0 pending, 9 pass → **the bucket rules pass**
* → `pr_merge_gate.py` prints `=> GREEN` and **exits 0**.

CI *did* see exactly what merged. What CI could never have seen is a 239-byte logic overread — no
build job detects that. #3252's own title says what actually happened: *"the tail of #3243's
**review** that the merge missed"*. That is the **review** detachment, which `CLAUDE.md` already
documents in the bullet immediately above the new one.

**Why this was easy to get wrong** (I read it the same way for twenty minutes): `gh api
repos/OWNER/REPO/pulls/3243/commits` returns only the four frozen commits, so the PR record really
does look like a head that froze mid-flight. Resolving it needs the two extra commits' dates set
against `mergedAt`, plus `commits/<sha>/check-runs` per commit.

### What has to change

* Correct the claim everywhere it appears — issue #3259, PR body, both docstrings, the
  `CMakeLists.txt` comment, `AGENTS.md`'s closing section, and above all `CLAUDE.md`. This is
  precisely the propagation `CLAUDE.md` warns about: *"a stale figure carries the charter's
  authority, which is exactly the authority it no longer deserves."* Five copies were written from
  one unverified sentence, and none of the five re-derived it.
* Keep the head rule. It is cheap, sound, and the head/tip gap is a real hazard — it just isn't
  what happened here. Justify it on its merits rather than on an incident it would not have caught.
* Don't claim the gate covers `#3243`'s class. It covers `#3234`'s, which is enough.

### The related structural gap this exposes

The docstring says *"the head the checks describe is compared against the real branch tip"*. It
isn't — the tool compares `headRefOid` to the tip and **assumes** the checks describe `headRefOid`.
`gh pr checks --json` cannot tell you otherwise: its JSON fields are `bucket, completedAt,
description, event, link, name, startedAt, state, workflow` — **no SHA** (confirmed from
`gh pr checks --help`, gh 2.63.2). If GitHub's rollup ever lags the head, green checks from an older
commit plus a matching head/tip yields **exit 0**.

The fix also makes the sentence true: fetch the checks SHA-addressed, via
`gh api repos/{owner}/{repo}/commits/<headRefOid>/check-runs`. That endpoint is keyed by commit by
construction, so the tool would then verify what it claims to verify. I used exactly that call to
produce the table above.

---

## BLOCKING 2 — the sixth mutation: `collect()` and `main()` are untested, and both fail-open

The author's mitigation (positive control + assert on the reason) is **sound, and I checked it**:
each of the five published mutations does redden its named arms. But it defends `evaluate()` only,
and the test boundary stops there. Every arm imports `evaluate` and `parse_checks` directly; nothing
ever calls `collect()` or `main()`. So the entire I/O and exit-code layer — where both incidents'
facts are actually *gathered*, and where `CLAUDE.md`'s new promise about exit 2 lives — is unverified.

I ran five mutations there. **All five leave the suite at 23/23 green, rc=0:**

| # | mutation | effect | suite |
| --- | --- | --- | --- |
| **A** | `return 2` → `return 0` in `main`'s `except GateError` | **the gate prints its error to stderr and exits 0 on every tooling failure** — no auth, no network, malformed payload, `gh` missing | green |
| **D** | `return 0 if res.ok else 1` → `return 0` | **a gate that approves everything** | green |
| B | `tip = ls.split(...)` → `tip = pr_head` in `collect` | head rule is a no-op | green |
| C | `evaluate(checks, head, tip)` → `evaluate(checks)` | head rule skipped | green |
| E | `if not tip:` → `if False:` | empty tip accepted | green |

**A** is the one that matters. It is exactly the outcome this review was asked to look for — GREEN
when the gate malfunctioned — and it is invisible.

**D** is the sharper point about the discrimination argument. The suite's stated defence is aimed at
*"a gate that refuses everything"*. Its mirror — **a gate that approves everything** — is defended
only by the single positive-control arm, which `evaluate` satisfies for free; nothing pins that
`main` translates `ok` into an exit code at all. The docstring, `AGENTS.md` and `CLAUDE.md` all make
the 0/1/2 contract a load-bearing promise, and no arm holds it.

### Fix

Add arms that drive `main()` with `gh` and `git` stubbed on `PATH` (the sibling
`test_trap_number.py` already establishes that pattern — *"`gh` is a stub on PATH"* — so this is
consistent with the folder, not new machinery). Minimum:

* `gh` exits non-zero → **rc 2**, and *not* 0 and *not* 1
* all-pass, head == tip → **rc 0**
* one `fail` → **rc 1** (distinct from 2)
* `git ls-remote` prints nothing → **rc 2**
* the head actually compared is the `ls-remote` tip, not `headRefOid` echoed back

---

## BLOCKING 3 — two demonstrated exit-0 paths

Both reproduced end-to-end against an unmodified copy of the tool with stub `gh`/`git` on `PATH`.

### 3a. `headRefOid: null` silently deletes the head rule, and the output does not say so

`collect()` catches `KeyError`, but a JSON `null` is not a `KeyError` — it yields `pr_head = None`,
and `evaluate`'s guard `if pr_head is not None and branch_tip is not None` then skips the comparison
entirely. `render()` prints the head line only `if self.pr_head:`, so it prints nothing:

```
GATE #9999: pass=2 fail=0 pending=0 skipping=0 cancel=0
  => GREEN
EXIT=0
```

A reader asking "did it verify the head?" gets **silence**, and silence reads as yes. That is the
tool's own design principle inverted — a missing fact disables a rule instead of refusing.

Worse, the suite **blesses** this. `test_pr_merge_gate.py:127`:

```python
res = evaluate(parse_checks(rows(("Linux", "pass"))), None, None)
case("no head information given -> checks alone decide", res.ok, True)
```

That arm pins the fail-open as correct. Ask this suite's own question of it — *what else could
satisfy this assertion?* — and the answer is "a gate with no head rule at all", which is mutation C.

**Fix:** validate in `collect()` (`isinstance(pr_head, str)` and 40 hex chars, else `GateError`), and
make `render()` always emit a head line — `head NOT VERIFIED` when absent — so the boolean can never
be read without the caveat.

### 3b. `git ls-remote` tail-matching can shadow the branch

`ls-remote`'s pattern matches the **tail** of a ref name, so `refs/heads/foo` also matches a branch
named `bar/refs/heads/foo`. I built a two-branch bare repo and confirmed **two lines** come back.
`ls.split("\t")[0]` splits the whole multi-line string and takes line 1's OID, sorted first:

```
REAL branch tip is 2222...; PR head is 1111... -> a correct gate must REFUSE
GATE #9999: pass=1 fail=0 pending=0 skipping=0 cancel=0
  head matches branch tip (pr=11111111 tip=11111111)
  => GREEN
EXIT=0
```

Probability here is negligible — you would need a branch literally named
`<anything>/refs/heads/<target>`. But the fix is three lines and turns an ambiguous match into a
refusal, which is the posture the whole tool argues for:

```python
tips = [ln.split("\t", 1) for ln in ls.strip().splitlines() if "\t" in ln]
exact = [oid for oid, ref in tips if ref.strip() == "refs/heads/%s" % branch]
if len(exact) != 1:
    raise GateError(...)
```

(For completeness on the "unusual characters" question: git's own ref rules forbid space, tab and
`~^:?*[\`, so tab-splitting is safe and a branch name cannot itself glob. The tail-match above is
the only residual.)

---

## Non-blocking: what I checked and found correct

**The `#3234` half is verified first-hand and fully justifies the tool.** `gh pr checks 3234 --json`
today reports `fail / FAILURE / Windows MinGW` alongside 8 `pass` and 2 `skipping`. It merged red.
9 of its 11 check names contain spaces. This is a real, recurring, mechanical hazard.

**The bucket vocabulary is right, and I can confirm it from more than observation.** `gh pr checks
--help` (gh 2.63.2, installed here) states it normatively: *"a `bucket` field, which categorizes the
`state` field into `pass`, `fail`, `pending`, `skipping`, or `cancel`."* Exactly the five constants,
no more. **Could not confirm:** behaviour of other gh versions. The tool blocks on an unrecognised
bucket, so that gap falls in the safe direction.

**`--json` does not carry the fail/pending exit codes — which is lucky, and worth knowing.** This
was my first worry: `gh pr checks` documents `8: Checks pending`, and `_run` turns any non-zero exit
into a `GateError`, which would have made the fail/pending branches of `evaluate` dead code in
production. It doesn't happen: in gh's source the Exporter path returns *before* the
`SilentError`/`PendingError` returns. Confirmed live on this very PR — 7 pending checks, **rc=0**.
Worth a comment in `_run` so nobody "fixes" the flag later.

**One real consequence of the same ordering, though:** `populateStatusChecks` raises *"no checks
reported"* **before** the Exporter path. So a PR with genuinely zero checks makes
`gh pr checks --json` exit 1 → `_run` raises → **the gate exits 2**, not 1-with-`VOID`. The
`pass == 0` rule is therefore unreachable from `collect()` for the empty case (the *all-skipped*
case does reach it, and works). Safe direction, but the PR body and `CLAUDE.md` describe a behaviour
the tool does not exhibit. One sentence fixes it.

**`skipping` as mergeable is correct here, transitively — and worth stating why.** A job skipped
because a `needs:` dependency failed does report `skipping`, so the tool genuinely cannot tell the
two apart. It doesn't need to: `.github/workflows/ci.yml:861-862` has the only `needs:`-dependent
job (`Publish Release`), and every job it depends on is itself a visible check, so a dependency
failure surfaces as `fail` and blocks. The guarantee lapses only if a future job `needs:` something
that is not itself a check — worth one line in the comment. Also: the comment names `Progress
tracker` as the skipping job; there are **two** (`Publish Release` skips on every non-tag build).

**Fork PRs:** `headRefName` is the fork's branch, absent from `origin`, so `ls-remote` is empty →
`GateError` → exit 2. Refusing is right; the message *"deleted, or never pushed?"* is a misleading
diagnosis. `gh pr view --json isCrossRepository` would let it say what it means. Low priority for a
single-account repo.

**`AGENTS.md`:** the opening paragraphs (responsibility, the boundary against `tools/docs/` and
`tests/`) and both closing sections are exactly the map the charter asks for — *"anything a newcomer
would otherwise learn only by getting it wrong"* is well served by the "they all fail silently"
section. The `## What lives here` per-file list is the part that will rot, and it already omits
`test_pr_merge_gate.py`. Not blocking; consider trimming it. Its closing section repeats the `#3243`
claim and inherits Blocking 1's correction.

**Minor:** the PR body says *"24 arms"*; the suite emits **23** `case()` results (23 `ok` lines,
23 `case(` call sites). In a PR whose thesis is *"a count is falsifiable, an exit code alone is
not"*, that one is worth getting right. Also `state` is requested in `--json` and never used —
harmless, and keeping it makes the *"wrong --json fields?"* error broader, so this is an observation
rather than a request.

---

## What I verified first-hand vs. took on trust

**First-hand:** the timeline table (`gh api` on commits, check-runs, `pulls/3243`, `ls-remote` on
the still-live branch); `#3234`'s `fail` bucket; gh 2.63.2's documented bucket vocabulary and
`--json` exit behaviour, live on this PR; gh's Exporter-path ordering from v2.63.2 source; the suite
at 23/23 rc=0 on an unmodified copy; five `collect`/`main` mutations all green; both exit-0
reproductions with stub `gh`/`git`; `ls-remote` returning two lines for one pattern in a purpose-built
bare repo; `git diff --check` clean; **422 insertions / 0 deletions** (trap 41 satisfied — the diff
is insertions-only, arithmetic checks out); no `/home/`, `/var/home/`, `C:\Users` or `/Users/` in any
added line; `check_usage_text.py --selftest` rc=0 and `--root` scan rc=0 (40 modules, 0 violations);
`check_ctest_gate.py --selftest` rc=0 (16 arms); `check_contribution_shape.py --base origin/main`
rc=0.

**Taken on trust (build embargo — a GTA V capture owns the GPU):** the ctest registration behaving as
described, `1/1 Test #29: pr_merge_gate ... Passed`, and **345 vs 344** registered tests. I read the
`add_test` block and it is well-formed, but I did not configure or run cmake/ctest.

**Could not confirm:** gh bucket vocabulary on versions other than 2.63.2.

---

Fix Blocking 1 (the `#3243` correction, in `CLAUDE.md` above all), 2 (arms over `main()`) and 3
(the two exit-0 paths) and I expect to approve quickly — the tool is worth having, and its `#3234`
half is a genuine recurring defect I verified merged red.

---

## Addendum — ruling on the docs-only exception (raised after this review began)

**Not in this PR. Follow-up issue, and not urgent yet.** Three reasons, then the constraints I would
want written into that issue, because I think the proposed shape is *nearly* right and one part of
the framing is wrong.

**1. This PR needs to shrink, not grow.** It carries a false historical claim that is already live in
`CLAUDE.md`. Correcting that is time-critical in a way ergonomics is not; adding a mode now puts a
policy debate in front of a correction that should land today.

**2. The premise is one observation, not a measurement.** On #3258 the gate reported four pending
platform jobs. *That report was true.* The gate was not wrong — a human then applied a documented
exception the tool does not know about, verified its preconditions by hand, and merged. That is the
division of labour working, not failing. "People will routinely override it" is a prediction. The
charter's own discipline applies: measure the friction before engineering against it. Record #3258 in
the follow-up as observation #1 and count the next ones. If docs PRs are frequent enough that this
recurs weekly, build the mode; if it happens twice a month, human judgment is the right instrument.
(I verified #3258: one path, `prosper/docs/STRAY_STATUS.md`. The exception genuinely applied.)

**3. The framing I'd push back on.** The choice is not "policy in the tool" vs "human judgment". The
failure you correctly fear — *a gate people routinely override stops being a gate* — is a failure of
**silence**, not of strictness. An override that leaves no trace degrades invisibly; an override
recorded with its verified preconditions does not. So the cheapest fix is available **today, with no
code change**: `--json` already emits `reasons` and `pending` by name, so paste the exact refusal into
the PR alongside the two preconditions you checked. That converts an invisible override into an
auditable one, which is the property that actually protects the gate. Do that now regardless of what
the follow-up decides.

### If the mode is built, these constraints matter more than its name

On your concern (1) — **I side with your instinct, but narrowly.** The tool must not learn *which
checks matter for a docs diff*; that is policy and it will rot. It may learn to *verify a
precondition and relax exactly one rule*, because that is a faithful mechanical translation of one
charter sentence. Concretely: `--docs-only` should verify the diff, then **ignore `pending` and
nothing else** — never `fail`, never `cancel`, never an unknown bucket, and never the head check
(which, as you say, applies to docs PRs exactly as much as to code). Note this means the mode does
*not* need to know that `Docs` and `Contribution shape` are the ones that can fail — requiring zero
failures across all checks gets you there without naming any. If a later version needs check names,
that is the signal it has crossed into policy and should stop.

On your concern (2) — your invariant is right (*verify, don't accept*), and it needs three additions
your proposed arms don't reach:

* **The verification must be fail-closed.** If `gh pr view --json files` cannot be read, or the
  listing is paginated/truncated (GitHub caps the PR files listing), `--docs-only` exits **2**. And an
  **empty** file list must refuse — zero files is not "every file is `.md`". That is the `pass == 0`
  VOID rule again, in a new place, and it is exactly the shape this tool exists to catch.
* **Renames are a hole in the data source, and I checked this first-hand:** `gh pr view --json files`
  returns only `["additions","deletions","path"]` — **no change status and no previous path**. A
  rename from `foo.cpp` to `foo.md` therefore presents as a single `.md` path and would read as
  docs-only. Exotic, but it is the one case suffix-matching cannot see; either use
  `gh api repos/{owner}/{repo}/pulls/{n}/files` (which carries `status` and `previous_filename`) or
  document the gap.
* **A third arm, and it is the one your two miss.** You proposed *"claims docs-only but the diff is
  not"* and *"docs-only with a stale head still refuses"* — both good. Add **"docs-only with a `fail`
  bucket still refuses"**. Without it, the only tested property of the relaxation is that it relaxes;
  nothing pins that it *stopped where it should*. That is this suite's existing gap in miniature, and
  it is the same reason Blocking 2 exists: the arms verify the rule that was added, never the rules
  that must survive it.

Ship the narrow correct tool. You are right that it is the better trade.

mattias800 (commented):
# Re-review — #3260 @ 6778350f7

**REJECTED — but only for two completions, not a re-litigation.** All three blocking findings are
genuinely fixed and I verified each independently rather than taking the summary on trust. What
holds this back is that the withdrawal you made is **incomplete in three places inside this diff**,
and that the exception fix stops one line short of the class it was written for.

The issue #3259 correction is exemplary. Leaving the false text in place and marking it withdrawn,
with the dated API record beside it, is the right way to do this — the error stays legible to
whoever reads the issue next, which is the whole point.

---

## Verified fixed

**Finding 2 — I reproduced your mutation table exactly.** Nine mutations, and every one reddens
by name with **zero tracebacks**:

| mutation | named failures | mine |
| --- | --- | --- |
| `return 2` → `return 0` on GateError | 6 | **6** |
| `0 if res.ok else 1` → `0` | 2 | **2** |
| `tip = pr_head` | 2 | **2** |
| `evaluate(checks)` without heads | 1 | **1** |
| accept a null `headRefOid` | 1 | **1** |
| `resolve_tip` suffix-matches | 2 | **2** |
| drop the VOID rule / head rule / cancel / pending | 3/3/1/1 | **3/3/1/1** |

Suite: 36 arms, 0 failures, rc=0. The `main()` arms are the right shape — monkeypatching `_run` as
the single process boundary avoids a production seam, and `run_main` restores it in a `finally`.

**Findings 3a/3b** — `collect()` refuses null/empty `headRefOid` and a missing `headRefName` with
exit 2; `resolve_tip()` matches `refs/heads/<branch>` exactly. I re-ran the suffix decoy against the
new implementation and it resolves correctly. Keeping the decoy as a fixture was the right call.

**The traceback-masquerade catch is real and works.** I confirmed directly: an unexpected
`TypeError` raised inside `collect()` now returns **2**, not 1.

Mechanical: `diff --check` clean; **18/10/47/259/235, all zero deletions** (insertions-only, trap 41
satisfied); zero host paths in added lines; `check_numbered_table.py` rc=0 over the new `CLAUDE.md`
(1 table, 33 rows) and `AGENTS.md`; `check_ctest_gate.py --selftest` and `check_usage_text.py
--selftest` both rc=0.

---

## BLOCKING A — the withdrawal covers three of six places; three still assert the fabricated incident

You corrected `CLAUDE.md`, `AGENTS.md`, `pr_merge_gate.py` and the issue. `git grep` for `3243`
across the diff finds three more that still state it as fact:

* **`prosper/CMakeLists.txt:348`** — *"an all-green result describing a PR head the branch had moved
  two commits past (#3243). Both are shapes this repository produces routinely … so neither arm is
  hypothetical."* This is the withdrawn claim asserted verbatim, plus an explicit "not hypothetical".
* **`test_pr_merge_gate.py:8`** — *"#3234's red `Windows MinGW` and **#3243's frozen PR head**. An
  arm built from a real incident cannot be satisfied by a scenario nobody encounters."* This is worse
  than a stale mention: it is the arm's **epistemic warrant**, and half of it is now known false. The
  arm is fine; its justification has to become "a mechanism this repo produces" rather than "an
  incident that happened".
* **`test_pr_merge_gate.py:79-81`** — *"(the #3243 class)"* and *"which is precisely the state that
  merged an unreviewed 239-byte overread."*

This is exactly the propagation defect the original finding was about, now in miniature: one
sentence in six places, four corrected, three not, and nothing goes red. A grep for the issue number
across the diff before pushing is the whole fix.

Minor, same family: `OTHER = "fd1dbb92aa1c40e2b2f7a0f0e8ee2f0f3d5c1a77"` (line 46) is a fabricated
hash keyed to the withdrawn incident's real prefix. Now that the incident is withdrawn, a neutral
constant stops the association being re-derived by the next reader.

---

## BLOCKING B — the catch-all covers `collect()` only, and the masquerade survives one line lower

Your diagnosis was right and the fix is one statement short. The `try` wraps `collect()`; `evaluate`,
`render` and `print` sit outside it. Demonstrated by injecting a `KeyError` into `render()` and
calling `main()` with stubbed `_run`:

```
unexpected exc inside collect()  -> rc=2      <- your fix, working
unexpected exc inside render()   -> ESCAPED KeyError (interpreter exits 1)
```

So a bug after `collect()` still surfaces as rc=1 — *"the gate said no"* — which is the precise
confusion the catch-all exists to prevent, and which your own comment at line 240-244 describes.

**And it is reachable without a mutation.** `pr_merge_gate.py 3260 | head -1` raises
`BrokenPipeError` inside `print`, outside the guard. Any caller keying on 1-vs-2 mis-routes it.

Fix: compute inside the guard, emit outside it.

```python
try:
    checks, head, tip = collect(args.pr, args.repo_dir)
    res = evaluate(checks, head, tip)
    out = json.dumps({...}, indent=2) if args.json else res.render(args.pr)
except GateError as exc:
    ...
    return 2
except Exception as exc:
    ...
    return 2
print(out)
return 0 if res.ok else 1
```

An arm that pins it: monkeypatch `pr_merge_gate.evaluate` to raise, assert `main()` returns 2.

---

## The tenth mutation the expanded suite still misses

I looked for one, and it is worth having because it sits on the path we just agreed on.

```
main(): `return 0` inserted before the --json print   ->  suite GREEN, rc=0, 0 named failures
```

**No arm passes `--json`.** The human-readable exit path is now well pinned; the machine-readable
one is pinned by nothing. That matters more than it would have yesterday: you accepted the
recommendation to use `--json`'s `reasons` as the record of a docs-only override, so a follow-up may
put tooling on that path — and a mode whose exit code is unverified is a gate that says yes to every
machine consumer.

One arm closes it:

```python
case("--json still exits 1 on a failing check",
     run_main(FakeGh(RED_CHECKS, VIEW, LS_ONE), argv=("1", "--json")), 1)
```

**Two further mutations survive; I am reporting both with honest severity rather than padding:**

* **`_run`'s `if p.returncode != 0` removed → suite green.** Every `main()` arm monkeypatches `_run`
  away, so the one unit that touches processes is never executed. I traced whether it is
  *exploitable* and today it is not: a failing `gh pr checks` yields empty stdout, which
  `parse_checks` rejects; a failing `gh pr view` fails JSON parsing; a failing `ls-remote` yields no
  hits in `resolve_tip`. All three still land on exit 2. So this is **latent, not exploitable** — but
  that safety is a property of three parsers, not a guarantee, and it would evaporate the day one of
  them learns to tolerate empty input. Worth one arm calling the real `_run` on `["false"]`.
* **`resolve_tip`'s `len(hits) > 1` removed → suite green, and I do not think this needs an arm.**
  A remote cannot advertise two refs with identical names, so the branch is unreachable defensive
  code. Noting it only so it is not mistaken for a gap later.

---

## Your question: is the `CLAUDE.md` withdrawal too hedged?

**Not in substance — but the weighting is wrong, and the weighting is what creates the risk you
named.** Three specific changes, none of which retreat from honesty:

1. **Un-bold "No merge here is known to have been lost this way."** It is the *only* bold phrase in
   the retraction half, and bold is a scanning aid. A reader skimming for "must I do this?" is
   currently steered by your typography straight to the reason to skip it. Keep the sentence;
   demote it to plain text.
2. **Lead with cost and mechanism, not with the absence of a scar.** The rule's justification is
   strong and does not depend on #3243 at all: it costs one command, and this repository produces
   the state routinely — `CLAUDE.md` says elsewhere that lanes push while CI runs, and #3260's own
   `CMakeLists` comment says the same. Say *that* first, then the retraction. The honest version of
   "no known save" is not "so this may not be worth doing", it is "so this is a precaution, and it
   is a cheap one".
3. **Date the claim so it decays visibly.** *"No merge here is known to have been lost this way"* is
   a measurement, and this file's own rule is that a measurement you made must be corrected when it
   changes — which nobody will do for a sentence with no date. Write *"as of 2026-09-02, no merge
   here is known…"*. Then a reader in six months can see the claim is old rather than treating it as
   current, which is precisely the failure mode this file warns about elsewhere.

One structural suggestion: the full retraction narrative now lives in three places, and the tool
docstring and `AGENTS.md` are the right homes for it — they are read by someone working *on* the
gate. `CLAUDE.md` is read by everyone about to merge anything, and there the retraction is currently
longer than the rule it qualifies, which inverts their importance. One clause plus a pointer would
carry the same honesty at a tenth of the reader's attention.

I think your instinct was right that the honesty is worth the cost. The cost is only high because
the retraction is currently formatted as the headline.

---

## Residual nit, and it is the one phrase from my first review that survived

The narrative is corrected, but the wording I flagged as an *assumption* is still in three places:
*"the head **the checks describe** is the branch tip"* (docstring line 41, `CLAUDE.md`), and
`evaluate`'s reason string *"the checks describe %s but…"*. The tool compares `headRefOid` to the
tip; it never establishes which commit the checks describe, because `gh pr checks --json` carries no
SHA. Your corrected body paragraph gets this right — *"the PR's recorded head is compared against
the real branch tip"*. Propagate that phrasing to the other three and the tool stops claiming a
verification it does not perform.

---

## First-hand vs trust

**First-hand:** the nine-mutation table reproduced independently (rc, named-failure counts and
traceback counts for each); suite 36/36 rc=0; three additional surviving mutations; the
`collect()`-vs-`render()` exception boundary demonstrated through `main()` with stubbed `_run`;
`git grep 3243` across the diff; `diff --check`; the 0-deletion numstat; the host-path scan;
`check_numbered_table.py` on the new `CLAUDE.md` and `AGENTS.md`; both sibling selftests; the issue
#3259 correction as posted.

**Taken on trust (GPU/build embargo):** `ctest pr_merge_gate 1/1 Passed`. I ran the suite directly
instead, which exercises the same script.

Fix Blocking A (grep the issue number, three comment blocks) and Blocking B (one statement move plus
an arm), add the `--json` arm, and I will approve on sight.

mattias800 (commented):
# Re-review — #3260 @ 3de50967c

**REJECTED**, for one new finding and one one-line residual. Everything from rounds 1 and 2 is
fixed, and I verified it rather than accepting the summary. You asked for someone else to look
because you had no articulable worry left; there was one, and it is on-thesis enough that I think
it is worth one more round.

**I should have found it in round 2.** I checked whether the `--json` exit path was pinned and
never asked the same question of the human-readable one. The question applied identically to both.

---

## The thirteenth mutation: the report a human reads is pinned by nothing

`GateResult.render()` is the primary interface — `--json` is the machine path, this is the one
someone reads before pressing merge. **Five mutations to it leave the suite at 44/44 green, rc=0:**

| mutation to `render()` | suite |
| --- | --- |
| the pass count always prints `0` | **green** |
| `BLOCKING:` lines dropped — a failing check is never named | **green** |
| the head line dropped | **green** |
| refusal reasons dropped | **green** |
| **`=> GREEN` printed unconditionally** | **green** |

The last one, end to end, with the mutant in place and the suite at 44/44:

```
GATE #3234: pass=1 fail=1 pending=0 skipping=0 cancel=0
  BLOCKING: Windows MinGW (fail)
  head matches branch tip (pr=aaaa1111 tip=aaaa1111)
  refused: 1 check(s) failed or were cancelled
  => GREEN
exit code = 1  <- correct
```

The exit code is right and **the report lies**. A human skimming to the verdict line reads GREEN on
a PR with a red `Windows MinGW` — which is #3234's failure reproduced with new machinery.

**Why this is the on-thesis gap rather than a coverage nit.** The tool's own docstring, quoting
`CLAUDE.md`, says: *"this tool reports COUNTS: a count is falsifiable, an exit code alone is not."*
The suite pins the exit code and not the count. The half the docstring names as the falsifiable one
is the half nothing checks — so the artifact does not currently satisfy the principle it argues for.

There is also an asymmetry worth seeing: after round 2 the `--json` path gained content assertions
(`ok`, `reasons` populated) and the human path did not. The machine-readable output is now better
verified than the one a person acts on. That is backwards for this tool.

**Fix — five arms on the helper you already have:**

```python
rc, out = run_main_capture(FakeGh(RED_CHECKS, VIEW, LS_ONE), ["1"])
case("a refusing report says NOT MERGEABLE", "=> NOT MERGEABLE" in out, True)
case("...and never says GREEN",              "=> GREEN" in out, False)
case("...and names the failing check",       "Windows MinGW" in out, True)
case("...and its counts are real",           "fail=1" in out, True)
rc, out = run_main_capture(FakeGh(GREEN_CHECKS, VIEW, LS_ONE), ["1"])
case("a green report says GREEN and shows the head",
     ("=> GREEN" in out and "head matches" in out), True)
```

---

## Residual — the overclaim survives in exactly one file, and the reason is mechanical

`grep` across the diff for `checks describe` returns **nothing**, and the phrase is still there:

```
CLAUDE.md:886    only when every check passed, none is pending, at least one passed, and the head the checks
CLAUDE.md:887    describe is the branch tip.
```

**It is hard-wrapped across the line break.** A whitespace-normalising scan
(`tr '\n' ' ' | tr -s ' '`) finds it; the natural grep does not — and that scan reports
`CLAUDE.md: 1`, everything else `0`. So the docstring, `evaluate()`, `CMakeLists.txt`, `AGENTS.md`
and the test file are all correctly updated, and the one survivor is the highest-authority file.

**This explains all three rounds, and it is the most useful thing in this review after the
mutation.** Every miss has been in prose, prose here is hard-wrapped at ~100 columns, and a
single-line grep cannot see a wrapped phrase. Round 1: six places. Round 2: three missed. Round 3:
one missed — and each time the survivor was in a different file, which is what a wrap-blind check
produces. The fix for the recurrence, not just this instance:

```sh
git show <ref>:<file> | tr '\n' ' ' | tr -s ' ' | grep -o '<phrase>'
```

---

## Verified fixed this round

**Eleven of your twelve mutation counts reproduce exactly** (VOID 3, head-mismatch 3, cancel 1,
pending 1, unknown bucket 2, main-always-0 3, GateError→0 6, no-unexpected-guard 1, tip=pr_head 2,
null head 1, suffix match 2), rc=1 and **zero tracebacks** on every one. My `--json`-suppression
mutant reddens **4** where you report 5 — almost certainly a different mutant shape rather than a
discrepancy in the suite, since the four that redden are the ones that can. Baseline 44 arms, rc=0.

**Blocking A** — both test-file sites and the CMake comment are corrected, and the docstring's new
framing is the right one: naming the stale-head arm as guarding a routine state rather than
borrowing a reconstruction's warrant is more honest than the version I asked for.

**Blocking B** — the guard spans the whole body; the `| head -1` BrokenPipeError reasoning is in the
comment where the next reader will find it. I confirmed an exception in `render()` now returns 2.

**C** — the five `--json` arms are right, including `reasons` populated.

**Your two harness defects are genuinely fixed** — both `json.loads` sites and the render-escape are
now caught and asserted as values, so no mutant reddens by traceback. That is what let me read the
counts above at face value.

**Hedging** — all three changes landed. The rule and its cost lead, the retraction is parenthetical
and dated, the full account moved to `AGENTS.md`. It reads correctly now: a reader learns the rule
and its cost before the caveat, which is the right order.

**Mechanical:** `diff --check` clean; **18/9/47/262/313, all zero deletions**; zero host paths;
`check_numbered_table.py` rc=0; `check_ctest_gate.py --selftest` and `check_usage_text.py --selftest`
rc=0.

---

## On your `tail` disclosure

Thank you for reporting it — a retracted confirmation is worth more than a silent one, and it
changed how I read your numbers rather than merely noting an error.

**My own counts are unaffected, and I checked rather than assuming:** every run redirects to a file
and reads `$?` immediately, with no pipeline. Where I used `rc=$?` inside a later `printf`, bash
expands `$?` left-to-right before the command substitutions run, so it is still Python's status.

**And it bit me this round anyway, in a different disguise.** My first attempt at the twelve used an
unquoted `$name` containing spaces, so the redirect failed with *"ambiguous redirect"*, Python never
ran, and every row came back `rc=1` — which looks **exactly** like "all twelve mutations redden".
A uniform, expected-shaped result produced by the harness never executing the thing under test. I
caught it only because the named-failure column was empty. The counts above are from the corrected
run; the first set was discarded.

---

## Suggestions, not blocking

* **`_run`'s `if p.returncode != 0` still survives**, unchanged from round 2 and for the same
  reason — every `main()` arm patches `_run` away, so the one unit touching processes is never
  executed. Still **latent, not exploitable**: all three call sites land on exit 2 downstream. One
  arm calling the real `_run` on `["false"]` closes it permanently.
* **`--repo-dir` is not threaded-through by any arm.** `FakeGh` ignores `cwd`, so replacing
  `repo_dir` with `"."` in `collect()` survives. The effect would be querying whichever repo the cwd
  sits in — a false-green class, if a remote one. Have `FakeGh` record `cwd` and assert it.
* `resolve_tip`'s `len(hits) > 1` branch survives but is unreachable — a remote cannot advertise two
  refs with the same name. Noting it so it is not mistaken for a gap later; no arm needed.
* `class ExplodingRender` (test file) is defined and never used — leftover from an earlier approach.

---

## First-hand vs trust

**First-hand:** baseline 44/44 rc=0 unpiped; all twelve claimed mutations re-run with counts and
traceback counts; five `render()` mutations surviving; the always-GREEN report demonstrated end to
end through `main()`; three further surviving mutations; the wrapped-phrase scan across all five
files; `git grep 3243` across the diff; `diff --check`; the zero-deletion numstat; the host-path
scan; `check_numbered_table.py`; both sibling selftests.

**Taken on trust (build embargo):** `ctest pr_merge_gate 1/1 exit 0`. I ran the suite directly.

Add the five render arms and fix the one wrapped `CLAUDE.md` phrase, and I will approve on sight.

mattias800 (commented):
# Re-review — #3260 @ a5521ab23

**APPROVED.**

Both round-3 findings are fixed and I verified each independently. One new finding below is the
same class as the thirteenth, one line lower — I am recording it as a **strong suggestion rather
than a blocker**, and I say why explicitly, because after four rounds "one more arm" needs to earn
its round and I do not think this one does.

---

## Verified fixed

**A — the six `render()` mutations all redden by name, with exactly the counts you report:**

| mutation | named failures |
| --- | --- |
| `=> GREEN` unconditional | **2** |
| pass count always `0` | **1** |
| `BLOCKING:` lines dropped | **1** |
| `pending:` lines dropped | **1** |
| head line dropped | **1** |
| reasons dropped | **1** |

Zero tracebacks on all six. Spot-checked four of the earlier twelve (`GateError`→0 = 6,
main-always-0 = 3, suffix match = 2, null head = 1) — unchanged. Baseline **53 arms, rc=0, unpiped**.

Asserting the `=> GREEN` arm on **absence as well as presence** is the right call and not the
obvious one: a presence-only assertion (`"NOT MERGEABLE" in out`) would have passed under the
unconditional mutant, because that mutant emits both strings across the report.

**B — the wrapped phrase is gone, verified with the wrap-blind scan rather than a grep.**
Whitespace-normalised across all five touched files: `CLAUDE.md 0`, `CMakeLists.txt 0`,
`AGENTS.md 0`, `pr_merge_gate.py 0`, `test_pr_merge_gate.py 0`. `CLAUDE.md`'s pointer bullet now
reads *"the PR's recorded head is still the branch tip"*.

**The `AGENTS.md` section you wrote around it is better than the finding was.** Scoping it to *any*
claim, figure or citation being retracted — rather than to this phrase — is the generalisation that
makes it a habit instead of a fix, and the runnable snippet means the next person does not have to
derive it. Naming the property that makes it bite (*"grep is line-oriented; fine for code, close to
useless for prose in this repository, which wraps at about 100 columns"*) is the part that will
survive.

`ExplodingRender` removed. Your reasoning for leaving `_run`'s returncode check, `--repo-dir`
threading and the unreachable `>1` guard is sound, and I agree with the distinction you drew: two of
those would be arms asserting that a line exists rather than that a behaviour holds.

**Mechanical:** `diff --check` clean; **18/9/65/262/339, all zero deletions**; zero host paths;
`check_numbered_table.py` rc=0; `check_ctest_gate.py --selftest` rc=0; `check_usage_text.py
--selftest` rc=0.

**End-to-end, against the live repository:** I ran the built tool on its own PR.

```
GATE #3260: pass=2 fail=0 pending=7 skipping=1 cancel=0
  pending:  macOS (x86_64 / Rosetta 2)   [+6 more]
  head matches branch tip (pr=a5521ab2 tip=a5521ab2)
  refused: 7 check(s) still pending
  => NOT MERGEABLE
rc=1
```

Correct on every axis: the head check passes against the real branch tip, seven multi-word check
names are counted as seven checks rather than as their words, and it refuses with rc=1. **Note the
practical consequence: your own gate says this PR is not mergeable yet.** Wait for the seven.

---

## The fourteenth mutation — suggestion, not a blocker

The same shape as the thirteenth, one line lower, and it survives:

```
render(): "matches" if same else "DOES NOT MATCH"   ->   "matches"
```

Suite stays **53/53 green**. The head line then reports a stale head as current:

```
GATE #3243: pass=2 fail=0 pending=0 skipping=0 cancel=0
  head matches branch tip (pr=fd1dbb92 tip=1144bea7)
  refused: the PR's recorded head is fd1dbb92 but the branch tip is 1144bea7 -- nothing verified …
  => NOT MERGEABLE
```

**Why I am not blocking on it, stated so you can disagree.** It is materially milder than the
always-GREEN mutant, and the difference is not one of degree. Under always-GREEN the *only* verdict
line lied and nothing else in the report contradicted it. Here the line contradicts itself on its
own text — it prints two visibly different SHAs beside the word "matches" — and both the refusal
reason and `=> NOT MERGEABLE` say the opposite. A reader is not told everything is fine; they are
told one descriptive line that three other lines refute. It is worth one arm, not a fifth round.

```python
case("...and a stale head is reported as NOT matching", "DOES NOT MATCH" in out, True)
```

on the stale-head `run_main_capture` you already have.

**Also unpinned, lower value:** the `--json` payload's `counts`, `blocking`, `pr_head` and
`branch_tip` — emptying any of them survives. `ok` and `reasons` are covered. `counts` is the one
I would add if you add any, because the docstring's argument is that the *count* is the falsifiable
artefact, and that is the field a follow-up building on `--json` would quote.

**One observation that is explicitly not a request:** the pure-function arms remain
traceback-fragile in a way the `main()` arms no longer are — any mutation that makes `evaluate()`
raise (I hit one making `counts["unknown"]` a `KeyError`) kills the suite with a stack trace and
names nothing. Wrapping every direct call would be noise for little gain, and the `main()` arms
cover the paths that matter. Recording it so it is a known property rather than a surprise.

---

## Closing

Four rounds, and each one moved something real: a fabricated incident withdrawn with its evidence
left legible, the I/O layer pinned, `--json` pinned, the human-readable report pinned, and a
line-oriented-grep habit turned into a documented recipe. The tool now does what its docstring
claims, and — unusually — its docstring is now precise about what it has *not* done, which is the
part most tools get wrong in the other direction.

On your note about my own harness failure: I appreciate you saying it lands in the same family as
the charter's positive-control warning, because that is exactly where I would file it. An instrument
that emits expected-shaped output while never executing the subject is the failure this whole PR is
about, and it is humbling that reviewing it produced a fresh instance.

**APPROVED** on `a5521ab23`. Land the one-line `DOES NOT MATCH` arm if you agree it is worth it — it
is test-only and additive and does not need another review — then merge once the seven pending
checks are green **and the head still matches**, which your own tool will tell you.

---

**Verified first-hand this round:** baseline 53/53 rc=0 unpiped; all six `render()` mutations with
named-failure and traceback counts; four earlier mutations re-run; six further candidate mutations,
of which the head-line inversion and three `--json` field mutations survive; the wrap-blind scan
across all five files; `git grep 3243`; `diff --check`; the zero-deletion numstat; the host-path
scan; `check_numbered_table.py`; both sibling selftests; and the tool run live against PR #3260.

**Taken on trust (build embargo):** `ctest pr_merge_gate 1/1 exit 0` — I ran the suite directly
instead, which exercises the same script.

mattias800:
Both suggestions from the approving review are now in, as `bab8bcb6e`.

**Approval carry-over, established by content rather than assumed.** The review was registered on
`a5521ab23`; the delta to the new head is **test-only, 19 insertions, 0 deletions**, touching solely
`test_pr_merge_gate.py`, and implements exactly the two arms the review suggested. No production
file changed. Per `CLAUDE.md`'s rule that a push detaches a review from head, I compared the content
instead of re-requesting.

**The fourteenth mutation is closed.** `"matches" if same else "DOES NOT MATCH"` → `"matches"` now
reddens two named arms. Both were needed: a presence-only assertion passes under that mutant, since
it emits `"head matches branch tip"` containing the substring either way — so one arm asserts the
correct string appears and one asserts the wrong one does not.

**`--json`'s counts, blocking list and compared SHAs are pinned.** Leaving them unasserted
contradicted the tool's own argument for printing counts at all.

Suite **58 arms, rc=0** on an unpiped run. Mutation matrix **19**, every one reddening by name.

One property recorded rather than fixed, from the same review: the pure-function arms remain
traceback-fragile where the `main()` arms are not — a mutation making `evaluate()` raise kills the
suite and names nothing. The fix is a `try`/`except` around every pure arm, which costs more in
noise than it buys.

Merging with this PR's own tool once the checks land, which seems like the right way to find out
whether it works.

Issue #3245, opened 2026-09-02
sceAmprMeasureCommandSizeReadFile's compatibility read runs for every title, so every streamed asset is read and DMA'd twice (13,925 duplicate reads / 582 MB in a 3-minute Stray boot)


mattias800:
`sceAmprMeasureCommandSizeReadFile`'s compatibility read runs for EVERY title, so every streamed asset is read and DMA'd twice — 13,925 duplicate reads / 582 MB in a 3-minute Stray boot

## What

`f_apr_measure_read_file` (`src/hle/fs/hle_file.cpp:3737`) performs the described read *and DMAs it
into the guest destination* whenever the file id resolves and `a1 > 0xffff && a2 != 0`. The firmware
contract for `sceAmprMeasureCommandSizeReadFile` is a **sizing** call that returns 20 or 24; the read
is a deliberate compatibility behaviour, added for *Dragon Quest VII Reimagined* (`PPSA17942`), whose
older SDK wrapper consumes the destination before any command buffer is submitted. The comment above
the function says exactly that.

The gate is only "does the id resolve", so it fires on **every** APR title, not only the one that
needs it. On a title that also submits the read normally, the same bytes are `pread` and
`process_vm_writev`'d into the same guest address twice.

## Measured

*Stray* (`PPSA02101`), `PROSPER_NULL_PAGE=1`, no input, `tools/screenshot`, ~180 s of wall clock,
`PROSPER_FILELOG=1`:

| | |
| --- | --- |
| `measure-read compatibility … OK` | **13,925** |
| `measure-read … (no compatibility read)` | 0 |
| `read-submit … OK` | 14,255 |
| bytes delivered by the measure path | **582,619,946** |
| bytes delivered by the submit path | 603,329,577 |

Every measure is paired with a submit of the same `dst`/`off`/`size` moments later, e.g.

```
[apr] measure-read compatibility id=1 hk_project-ps5.pak -> dst=0x3022480000 off=0x175ba99e size=1310865 OK; measured=20
[apr] read-submit id=1 hk_project-ps5.pak -> dst=0x3022480000(guest) off=0x175ba99e size=1310865 got=1310865 OK method=id requested=1310865
```

So roughly **97% of the measure-path bytes are re-delivered** by the submit path, and the boot pays
for ~583 MB of extra host file reads, ~583 MB of extra `process_vm_writev` into guest memory, and one
extra `mmap`/`munmap` per read.

## Why it is not only a performance question

[...]

PR #2924, opened 2026-08-22
hle(ampr): implement sceAmprAprCommandBufferReadFileGatherScatter, and lower the vertex prologue's EXEC_LO write — Yakuza Kiwami stops dying at 0.0 s

mattias800:
## What this fixes

`sceAmprAprCommandBufferReadFileGatherScatter` (`BVmR1H8l+XI`) was unimplemented and fell to the
dispatcher's return-0 default. **Yakuza Kiwami** (`PPSA31334`) reads each of its archives as exactly
two segments — a plain `ReadFile` for the unaligned tail and one `…GatherScatter` for the aligned
head — so the half that never arrived was the entire body of every file. The title printed its own
`Failed!! Load Devil2 Shader Archive` and then called a method on the null object the failed load
left behind: `SIGSEGV addr=0x1d4` at `eboot+0x6ecf6`, 0.0 s into every boot.

Behind that, every one of the title's shader programs was rejected by the recompiler on the
compiler's standard vertex prologue, so the second commit lowers that one instruction exactly.

Fixes #2872. Refs #2864, #2880, #2922, #2923.

## The failure scenario, and the contract

**APR gather/scatter is command-buffer state.** The firmware exports four read builders and one
teardown — `…ReadFile`, `…ReadFileGather`, `…ReadFileScatter`, `…ReadFileGatherScatter`, and
`sceAmprAprCommandBufferResetGatherScatterState`. That last export is the primary-source proof that
the gather/scatter variants carry no file id because the chain the plain `ReadFile` opened already
holds it.

The ABI is re-derived from the guest, not from the names. The SDK's inline wrapper at
`eboot+0xcc4a70` is the plain `ReadFile` wrapper at `eboot+0xcc4a40` with the file id removed and
every later argument shifted down one register:

```
public   GatherScatter(rdi=cb, rsi=dst, rdx=len, rcx=offset)
cc4a70:  mov r9,rcx / mov r8,rdx / mov rcx,rsi     ; offset -> a5, len -> a4, dst -> a3
         lea rsi,[rdi+0x18] / lea rdx,[rdi+0x20]   ; the same cursor pair the plain read gets
         jmp <BVmR1H8l+XI>
low      (a0=cb, a1=&cb[0x18], a2=&cb[0x20], a3=dst, a4=len, a5=fileOffset)
```

Note `mov rcx,rsi` is 64-bit where the plain read's is `mov ecx,esi`: a3 is a pointer, not the
[...]

PR #3262, opened 2026-09-02
APR read builders: refuse what cannot be derived, and stop reporting an undelivered segment as delivered

### LINKED ISSUE #2926: libSceAmpr: sceAmprAprCommandBufferReadFileGather / …Scatter answer SCE_OK with an unwritten destination
`libSceAmpr` exports **four** APR read builders. #2924 implements two of them; the other two are
unregistered and therefore answer `SCE_OK` from the dispatcher's return-0 default while writing
nothing to the guest's destination.

| NID | name | prosper |
| --- | --- | --- |
| `mQ16-QdKv7k` | `sceAmprAprCommandBufferReadFile` | implemented |
| `BVmR1H8l+XI` | `sceAmprAprCommandBufferReadFileGatherScatter` | implemented (#2872 / #2924) |
| `mZSbNJVJpV8` | `sceAmprAprCommandBufferReadFileGather` | **unregistered → returns 0** |
| `Jg-AgkdJHkk` | `sceAmprAprCommandBufferReadFileScatter` | **unregistered → returns 0** |

(Names and NIDs from the PS5 3.20 firmware library database, which gives names and NIDs only.)

## Why this is worth an issue rather than a shrug

This is exactly the failure mode `src/hle/fs/hle_file.cpp`'s gather/scatter handler argues against in
its own comment, and that *Yakuza Kiwami* demonstrated end to end: an unwritten destination reported
as success is indistinguishable, from the guest's side, from a successful read of zeros. On
`PPSA31334` that took four sessions of "the renderer must be broken" before the title's own
`Failed!! Load Devil2 Shader Archive` was traced to bytes that never arrived.

Nothing has been observed calling either NID yet — neither `PPSA31334` nor `PPSA02739` imports them
— so this is **not** a live blocker for any tracked title. It is filed because the machinery that
would serve them now exists, so the cost of closing it has collapsed.

## What "implement" means here, and what must NOT be assumed

`…GatherScatter` was derivable because the guest supplied a wrapper to read
(`eboot+0xcc4a70`) and a dispatcher that exercises both arms (`eboot+0xdb5fb0`). **No such evidence
exists for these two**, and the names alone do not settle the argument layout:

* standard DMA vocabulary says **gather** = many sources → one destination, **scatter** = one source
  → many destinations, so the plausible reading is that `…Gather` takes `(cb, dst, len, offset)` with
  the *destination* implied by the chain and `…Scatter` the mirror image — but "plausible" is not
  "derived", and the two differ in which argument disappears;
* the chain state they share is already modelled (`apr_chain_open` / `prosper_apr_chain_reset` in
  `src/hle/fs/hle_file.cpp`), so the missing piece is genuinely only the argument layout.

So **do not register them against a guessed layout**. The correct trigger is a title that imports
one: find its SDK inline wrapper the way #2924 found `eboot+0xcc4a70` (it will be adjacent — Kiwami's
sit at `0xcc4a40` and `0xcc4a70`), confirm against a live `PROSPER_FILELOG` run, and reuse
`apr_execute_read`. Until then, the honest state is the loud one — and prosper's current answer is
the *quiet* one, which is why this is filed.

A cheaper interim step that needs no layout at all: register both to a handler that **refuses**
(returns `0x80020016`, the same code the other builders return for an unresolvable file) and logs
once. That converts a silent wrong-data path into a loud failure without inventing anything.

Labels: `bug`, `area:infra`

### LINKED ISSUE #2928: APR gather/scatter: submit closes the chain undocumented, and a segment can report SCE_OK with nothing written
## Two findings from the post-merge review of #2924

#2924 merged as `d31b89d4` while an independent review was still in progress. The review came back
**APPROVED with no blocking findings**, so nothing shipped that had been rejected — but two of its
non-blocking findings are latent defects rather than polish, and they belong in the tracker rather
than in a review comment nobody will read again.

### 1. The chain's closer set is larger than the code says, and one of the closers is `Submit`

`prosper_apr_chain_reset` now runs from **every** `ampr_cb_reset` caller — including
`hle_kernel_mem.cpp:588` (the APR submit handler) and the AMM command-buffer constructor at `:2712`.
The comment beside it says *"Reset is the only closer observed"*, which was true of the title that
motivated the change and is not true of the code as written.

**Failure scenario:** an engine that opens a chain with the plain `ReadFile`, then **submits**, then
issues its `…GatherScatter` segments, loses the archive body. Yakuza Kiwami does not do this — it
issues both segments before submitting — so the defect is invisible on the title that drove the
implementation. It fails loudly rather than silently, which is the one mercy.

**Wanted:** either narrow the closer set to what the ABI actually requires, or update the comment to
enumerate the real closers and say why submit-as-closer is correct. The comment being wrong is the
part that will cost someone time.

### 2. A gather/scatter segment can return `SCE_OK` with the destination unwritten

`in_dst` is not consulted in the return value. For the plain `ReadFile` that is defensible — there
are known record-reading call sites that legitimately read without a destination. For **this**
builder the guest reads its own `dst`, so `in_dst == false` is almost certainly an undelivered
segment being reported as delivered, and it is invisible without `PROSPER_FILELOG=1`.

This is the silent-success class the charter singles out, in a function written specifically to fix a
silent success. Worth closing while the context is fresh.

### Also worth correcting

The merged commit message lists **Judgment (`PPSA02739`) as affected**. The PR's own new doc says the
opposite and is right: Judgment never calls the NID, and its `COMPATIBILITY.md` row change was a
same-lane measurement, not an effect of the diff. The body walked into the *import-table entry is not
a call* trap — the same one that cost a review on #2902 — and unlike a PR body, a squashed commit
message cannot be edited. Recording the correction here so the git log is not the only account.

### One instrument trap worth a numbered row

The eboot's ELF `p_offset` for `PT_LOAD[0]` is `0x4000`, but the SELF segment table puts that payload
at file offset `0x8010`. **Reading at `p_offset` puts every address `0x4010` off and lands
mid-instruction in unrelated AVX code** — which presents exactly as "this PR's addresses are wrong".
The reviewer nearly filed that before catching it. Allocate with
`python3 prosper/tools/docs/trap_number.py` (it scans master *and* open PRs).

mattias800:
## What

Three defects in one block of `src/hle/fs/hle_file.cpp`, all the same shape the charter singles out: **an APR read builder telling the guest "delivered" when nothing reached the destination.**

Fixes #2926, fixes #2928.

## Failure scenarios

**1. Two of the four read builders were unregistered (#2926).** `libSceAmpr` exports four APR read builders; prosper implemented two.

| NID | name | before | after |
| --- | --- | --- | --- |
| `mQ16-QdKv7k` | `sceAmprAprCommandBufferReadFile` | implemented | unchanged |
| `BVmR1H8l+XI` | `…ReadFileGatherScatter` | implemented | delivery gate added (below) |
| `mZSbNJVJpV8` | `…ReadFileGather` | **unregistered → SCE_OK** | refuses `0x80020016` |
| `Jg-AgkdJHkk` | `…ReadFileScatter` | **unregistered → SCE_OK** | refuses `0x80020016` |

Unregistered means the dispatcher's `return 0`. On this contract `0` is `SCE_OK` — *"the read is queued and will deliver"* — so an unregistered builder promises bytes and writes none, and the guest cannot tell the result from a successful read of zeros. This is *not* the "registering a NID can be worse than not" trap: these return an int32 SCE status, so `0x80020016` is a value the contract already defines (it is what the other builders return for an unresolvable file), not a sentinel read back as data. The refusal is weakly better than the default on every path — a guest that checks gets a truthful failure, and a guest that does not is no worse off than today.

**These are deliberately NOT implemented.** Standard DMA vocabulary suggests each names one side explicitly and takes the other from the chain, but the two readings differ in *which argument disappears*, and prosper's chain state carries no cursor for the implied side to advance — so even the argument count is unknown. Serving against a guessed layout would put real bytes at an address the guest did not name. The evidence that would settle it is the same shape #2924 used: a title that imports one, its SDK inline wrapper disassembled, and a live `PROSPER_FILELOG` run.

**2. A gather/scatter segment could return `SCE_OK` with the destination unwritten (#2928).** The handler returned `r.ok`, which says the bytes came off the *host file*, not that they reached the guest's own destination (`r.in_dst`). For the plain `ReadFile` the difference is defensible — it has known callsites that consume the completion record's data pointer and pass no destination. This builder has no such callsite and cannot: its segments all target one command buffer whose record is `cb+0x20` in every observed SDK wrapper, so each segment overwrites the previous segment's published pointer there, and a multi-segment chain is only readable through the per-segment `dst`. So `r.ok && !r.in_dst` is an undelivered segment reported as delivered — the silent-success class, in a function written specifically to fix a silent success. Zero-length segments are exempt (nothing to deliver).

**3. The chain closer set was wider than the comment beside it claimed, and submit was one of the closers it denied (#2928).** The real set is every `ampr_cb_reset` / `ampr_cb_construct` caller: Reset, ClearBuffer, both constructors, the 3.20 destructor, the explicit `…ResetGatherScatterState`, **and both submits**.

## The contract asserted, and where it came from

Names and NIDs from the PS5 3.20 firmware library database (`libSceAmpr.c`), which gives names and NIDs only. Argument layouts are unchanged from #2924's guest derivation.

**The closer-set confidence drops from MED to LOW, and the reason is a discriminator that does not exist.** The MED rested on Yakuza Kiwami's dispatcher (`eboot+0xdb5fb0`) calling `Reset` immediately before every chain-opening `ReadFile`. That ordering is *equally* consistent with "Reset closes the chain" and "Reset does not", because the `ReadFile` that follows re-opens it either way — the observation never bore on the question. Two facts point the other way: the firmware exports a **separate** `sceAmprAprCommandBufferResetGatherScatterState`, which a Reset that already cleared the state would make redundant; and prosper's real set was already much wider than the comment.

The set is therefore chosen by **which way the unknown fails**, not by evidence:

* closing too eagerly → a legitimate segment is **refused, loudly**, with the reason printed;
* closing too late → a stale chain names the **wrong file** and the segment is served from it, caught by the range guard only when the wrong file happens to be smaller.

The second is the silent-wrong-data outcome the handler exists to prevent, so prosper errs toward closing. The comment is now the enumeration rather than a claim it contradicts.

## Also corrected

`hle_kernel_mem.cpp` carried a live `CONFIDENCE: LOW until captured` hypothesis that one of `baQO9ez2gL4` / `ULvXMDz56po` / `Qs1xtplKo0U` / `GuchCTefuZw` "likely carries the READ RANGE". All four are named in the firmware database — `sceAmprCommandBufferReset`, `sceAmprCommandBufferClearBuffer`, `sceAmprAprCommandBufferDestructor`, `sceAmprCommandBufferDestructor`. A reset, a clear and two destructors cannot carry a file range. No capture was needed; one `grep` settles it.

## Deliberately unaffected

* The **plain `ReadFile`'s** return value — its record-reading callsites legitimately pass no destination.
* `apr_execute_read`'s record write and `prosper_ampr_advance` cursor accounting. A refused-on-delivery segment still advanced the cursor before this change and still does; the cursor feeds `GetCurrentOffset` only and prosper serves every read eagerly, so a wrong count cannot lose data. Changing it would alter the plain read's path too.
* Every closer's **behaviour** — only the comments and the confidence label change there. `test_apr_gather_scatter.cpp` pins the set so it cannot drift silently.
* No title is known to call either newly-registered NID: neither `PPSA31334` nor `PPSA02739` imports them. This is a fail-loud change, not a progression change, and there is no screenshot to attach.

## Risks

The refusal is louder than the previous silence. If a title *does* call `…ReadFileGather`/`…Scatter`, it will now see a failure where it previously saw a fabricated success — which is the point, but it may surface as a new guest-side fatal rather than as corrupt data. That is the intended trade and the issue asks for exactly it.

## Verification

Container `ps5ys`, Linux, exact head `26782b567` (rebased onto `origin/main` `49ec3e354`).

```
cmake -S prosper -B prosper/build-linux            # CONFIGURE_RC=0
cmake --build prosper/build-linux -j6              # BUILD_RC=0
cd prosper/build-linux && ctest --no-tests=error   # CTEST_RC=0
```

| run | tests | result | exit |
| --- | --- | --- | --- |
| `origin/main` baseline (measured here, at `e13fd40af`) | 344 | 100% passed | 0 |
| this branch, exact head | **347** | **100% passed** | **0** |

(`origin/main` moved to `49ec3e354` mid-session, which brings 2 tests of its own; 347 = 346 + this PR's 1. The same 346 falls out of #3245's branch independently — 348 there with 2 new tests.)

### Mutation arms — each demonstrated by exit code, not by assertion

**Arm 1 — the identical new test against unmodified `origin/main`** (test file and CMake entry copied into a clean `e13fd40af` worktree, nothing else changed):

```
test_apr_gather_scatter    ->  exit 1     3 [FAIL], 32 [ok]
  [FAIL] sceAmprAprCommandBufferReadFileGather is registered (not left on the return-0 default)
  [FAIL] sceAmprAprCommandBufferReadFileScatter is registered (not left on the return-0 default)
  [FAIL] a segment whose bytes never reached the guest destination reports FAILURE
```

Exactly the three arms the fix moves. Every closer arm and every positive control passes on baseline, which is the intended discrimination: the closers were already correct *behaviour* and only the comment was wrong.

**Arm 2 — narrowing the closer set** (temporarily removing `ampr_cb_reset(a0)` from the submit path, then restored):

```
test_apr_gather_scatter    ->  exit 1
  [FAIL] sceKernelAprSubmitCommandBuffer closes the read chain -- the next segment is refused
  [FAIL] sceKernelAprSubmitCommandBuffer: the refused segment left its destination untouched
  [FAIL] sceKernelAprSubmitCommandBufferAndGetResult closes the read chain -- ...
  [FAIL] sceKernelAprSubmitCommandBufferAndGetResult: the refused segment left its destination untouched
```

So the closer-set arms are load-bearing rather than decorative.

### Why the new test is not vacuous

Every refusal arm sits next to its own **positive control on the same chain**: the same segment, no closer in between, must succeed *and* must overwrite a `0xEE` poison with the fixture's own bytes. Without that, "the segment was refused" would be equally explained by a fixture that never opened a chain, and all eight closer arms would pass for the wrong reason. `sceAmprCommandBufferGetSize` is the other half — a non-closer must leave the chain **open**. The undelivered-destination arm uses a `PROT_NONE` mapping (a destination the host can read a file into but cannot write to the guest) and is immediately followed by the identical segment against a writable destination, so the arm is shown to discriminate rather than to refuse everything.

## Ruled out

Three rows added to `docs/UE4_APR_IOSTORE_BRINGUP.md` § Ruled out: the read-range hypothesis over the four then-anonymous NIDs, the void "Reset is the only closer" reading, and "an unregistered read builder is the safe default".

Snapshots: not run. This changes no rendering path and no title is known to reach the affected NIDs.

## Second commit: the instrument trap #2928 asked for

#2928's last section asks for a numbered row on the SELF `p_offset` trap. `cc7480488` adds it as **trap 252** in `docs/GAME_COMPAT_ORCHESTRATION.md`: the ELF program header's `p_offset` is not where a SELF segment's bytes are (`0x4000` vs the segment table's `0x8010` in the eboot #2924 took its addresses from), so a hand disassembly taken at `p_offset` is off by a constant, starts mid-instruction, and yields *plausible* instructions from unrelated code — presenting as "these addresses are wrong" about someone else's PR rather than as a broken instrument. prosper's loader was never affected (`src/self/module.cpp:91`, `:111` read the `SelfSegment` table and overwrite each phdr's `file_off`); the trap is confined to reading a SELF by hand with tools that do not know that table exists.

Docs gate on the merge result, run from the branch:

```
$ python3 prosper/tools/docs/check_merge_result.py
merging HEAD (cc748048) into origin/main (30c4270b)
  ... numbered column 1..252, 252 rows, unique and ascending, no unused numbers;
      none of the 251 rows in origin/main:... has been deleted
clean merge, and the merge RESULT passes the gate against origin/main as of now.
```

**Allocating that number turned up a tooling bug, filed as #3265:** `trap_number.py` offered 247, which has been taken since #3174, because its `--base` still defaults to `origin/master` — dead history since the rename to `main`. It was five numbers stale and said nothing about it. 252 was taken by reading `origin/main` directly.

## Review

Both reverse-engineered contracts here (the refusal policy, and the closer-set confidence downgrade) are the kind the charter asks to have read by someone else: the argument is about *which way an undecided lifetime should fail*, not about a measurement, and an independent reader can challenge that in a way CI cannot.

## Review round 2 — corrections on `cc7480488` (independent review REJECTED it)

Both blocking items were text/test-only and the design judgment was upheld. I reproduced every factual claim in the review myself rather than taking it on trust.

### Blocking 1 — the reachability claim was wrong by an order of magnitude

The shipped comment and the section above said *"neither PPSA31334 nor PPSA02739 imports them, so nothing observed calls these yet."* The premise is true; the conclusion does not follow from it. **My own sweep of the dump root reproduces the reviewer's result exactly:**

```
eboots scanned: 55
import BOTH mZSbNJVJpV8 and Jg-AgkdJHkk (NID# form): 22
PPSA01826 PPSA02058 PPSA02101 PPSA02154 PPSA03001 PPSA03026 PPSA03274 PPSA03831
PPSA04263 PPSA05143 PPSA05325 PPSA06367 PPSA07809 PPSA15319 PPSA17168 PPSA17942
PPSA19244 PPSA20447 PPSA21406 PPSA21564 PPSA23760 PPSA28061
```

Genuine dynamic imports, not incidental byte matches — `PPSA17942` and `PPSA02101` both carry `mZSbNJVJpV8#2#0` and `Jg-AgkdJHkk#2#0`; `PPSA31334` carries only `mQ16-QdKv7k#u#v` and `BVmR1H8l+XI#u#v`, exactly as originally stated.

The stronger claim is the one that survives, and its **absence is observable rather than assumed**: `prosper_on_unimpl` (`dispatch.cpp:190-215`) prints its first-seen line unconditionally — no env gate on POSIX, verified by reading it. Neither NID string occurs **anywhere** in DOLL's 91 MB or Stray's 33 MB `PROSPER_FILELOG` boot log (0 occurrences; 10 and 6 distinct unimplemented imports respectively).

So: **imported by 22 of 55, called by none observed.** The refusal's blast radius is 22 linked titles, not zero — including the two this PR's sibling is about. Corrected in the code comment and in the `## Ruled out` row.

### Blocking 2 — the closer enumeration omitted one NID, and so did the test

`Qs1xtplKo0U sceAmprAprCommandBufferDestructor` closes on both arms (`hle_kernel_mem.cpp:3506` → `k_ampr_destruct_320`, `:7300` → `k_ampr_destruct`, both → `ampr_cb_destroy_320` → `prosper_apr_chain_reset`) — verified. This PR's own defect class, reintroduced at one row. Added to the comment and to `closers[]`.

The arm **runs by name and is load-bearing**, not merely green:

```
[ok] sceAmprAprCommandBufferDestructor closes the read chain -- the next segment is refused
[ok] sceAmprAprCommandBufferDestructor: the refused segment left its destination untouched
```

Mutation (removing `prosper_apr_chain_reset` from `ampr_cb_destroy_320`, then restored) → **exit 1**, exactly the four destructor assertions fail.

### Non-blocking items, all taken

- **The return type is now derived, not asserted.** The reviewer found the structural evidence I should have cited, and it is stronger than reported: the firmware exports a separate sizing function **per read builder, one-to-one, all four** — `vWU-odnS+fU` / `qesF88X4DRg` / `DXmgc5op8Yw` / `7nXGDGMXSqo`. A builder never needs to return its own encoded size because sizing has its own entry point. Verified against the NID database.
- **The `…ResetGatherScatterState` inference was over-read and is corrected to NEUTRAL.** A coarse reset does not make a fine-grained one redundant — `vkResetCommandPool` / `vkResetCommandBuffer` is the same pairing. This *strengthens* LOW: nothing bears on the question at all, rather than the evidence pointing the other way.
- **Dropped "prosper's set was already wider" from the evidence list** — a fact about prosper, downstream of the question being asked. It stays in the narrative.
- **Cited the over-closing null.** Submit already closed before #2924 (this PR changes no behaviour in that path), so that PR's Kiwami run was measured **with the wide set already live** and reported **0 refusals** across four archives whose segments tile each file exactly. That upgrades "we chose the fail-loud side" to "and measured no cost on the one title that could have paid it."
- **Named the observation that would discriminate**, so the next agent knows what to collect: a Reset or submit arriving *between* a chain-opening `ReadFile` and a `…GatherScatter` on the same `cb`.
- **`a8uLzYY--tM sceAmprAprCommandBufferConstructor` is now listed as a deliberate exclusion** with the reason and the tension with the "err toward closing" policy stated, rather than silently absent. It closes nothing on either arm (verified); the observed flow `init` → `begin` means the chain is already closed when it runs.
- **`log_refusal` now has per-class budgets.** One shared 32-entry budget let early chain-not-open refusals silence the first UNDELIVERED one — the newest class and the one most worth seeing, since it means bytes were read and then lost.

### Asymmetry worth stating plainly

The plain `ReadFile` still returns `0` on the same staging-fallback condition where the gather/scatter builder now returns an error. That is defended in the comment and is a real behaviour change for any title whose `…GatherScatter` callsite consumes the record rather than `dst`. The FILELOG line already prints `guest`/`staging`, so one `PROSPER_FILELOG=1` Kiwami run plus `grep staging` would retire the question; no existing log has APR FILELOG on for that title.

Fixes #2926
### Verification of the corrected head `1efd2dbc4`

The local box is currently unusable for builds: several concurrent lanes have podman's database
locked (`Error: beginning transaction: database is locked` on every `distrobox enter` and every
`podman exec`, for over an hour), so I could not run the local rebuild + full ctest at this exact
head. **I am not quoting numbers I did not measure.** What I can state:

* the pre-correction head's local run is unchanged and quoted above;
* the corrections carrying the two blocking corrections plus every non-blocking item were each built and run locally *before* the lock set in, and their
  results are reported in this body with the arms named;
* **CI is the exact-head signal for this commit** — it builds and runs the full ctest on Linux,
  Windows MinGW and macOS on clean runners, which is a stronger check than my contended box.

A local retry is running and I will post the local counts to this PR if it clears.

Fixes #2928


mattias800 (commented):
## Independent review — #3262 (head `cc7480488`)

**REJECTED** — on two factual corrections, both small and both text/test-only. No code change is
needed and the design judgment is sound; I would approve on a corrected head.

I did not build, run ctest, or launch a title (a bisect owned the GPU). Everything below is from
reading the source, the firmware NID database, the local dump corpus, and existing run logs.

---

### The central claim holds: refusing IS better than the unregistered default

I set out to break this and could not. Verified first-hand:

* `src/hle/dispatch/dispatch.hpp:238` and `dispatch.cpp:190-215` — an unimplemented import returns
  `0` and prints one first-seen line. So the pre-change answer really is `SCE_OK`.
* The sibling `…ReadFileGatherScatter` is a status-returning contract *in practice*: prosper returns
  `0`/`0x80020016` there and #2924's Kiwami run loaded all four archives with the two segments
  tiling each file exactly. If the guest read that return as a byte count, prosper's `0` would have
  broken it. So `[aprf+0x1c] += bytes` in the #2924 disassembly is the locally-computed
  `min(len, size-off)`, not the callee's return — which is what the trap turns on.

The "registering a NID can be worse than not" trap does not apply, and the reasoning given is the
right reasoning. **One gap in the derivation, non-blocking:** the return type is *asserted*
("these builders return an int32 SCE error code") in the code comment, the PR body, and — highest
authority of the three — the new `## Ruled out` row, while the same comment says "the layout is
deliberately not claimed at all". A return type is part of a layout. The structural evidence you
need is one grep away and you did not use it: libSceAmpr exports **separate** sizing functions for
these exact builders — `qesF88X4DRg sceAmprMeasureCommandSizeReadFileGather` and
`7nXGDGMXSqo …Scatter`. The read builder does not need to return a size because sizing already has
its own function. That converts an assertion into a derivation; please add it.

---

### Blocking 1 — the reachability claim is wrong, and the shipped comment carries it

`hle_file.cpp` (the `apr_read_builder_unimplemented` block) and the PR body both say:

> Neither PPSA31334 nor PPSA02739 imports them, so nothing observed calls these yet.

The first clause is true. The conclusion does not follow from it, and the implied scope is wrong by
an order of magnitude. Sweeping every eboot in the local dump library for the import-form string
`NID#lib#mod`:

* **22 of 55 eboots import BOTH `mZSbNJVJpV8` and `Jg-AgkdJHkk`.**
* They include **PPSA17942 (DOLL)**, **PPSA02101 (Stray)**, PPSA04263 (GTA V), PPSA03831, PPSA05143,
  PPSA07809, PPSA19244, PPSA21406, PPSA23760, PPSA28061, PPSA01826, PPSA02058, PPSA02154, PPSA03001,
  PPSA03026, PPSA03274, PPSA05325, PPSA06367, PPSA15319, PPSA17168, PPSA20447, PPSA21564.
* Spot-checked as real dynamic imports, not incidental byte matches — e.g. PPSA17942 and PPSA02101
  both carry `mZSbNJVJpV8#2#0` and `Jg-AgkdJHkk#2#0`; PPSA04263 carries `…#1#z`. PPSA31334 carries
  only `BVmR1H8l+XI#u#v` and `mQ16-QdKv7k#u#v`, exactly as you said.

The *stronger* claim — the one you actually want, and which I verified independently — is that none
of them **calls** either NID on any observed route. `prosper_on_unimpl` prints its first-seen line
unconditionally, so a call is impossible to miss in a boot log. Across DOLL's 326 MB `bundle` boot
log (10 distinct unimplemented imports, neither NID among them) and four Stray run logs (7-12
distinct, same), `mZSbNJVJpV8` and `Jg-AgkdJHkk` appear **zero** times.

So the conclusion survives and the derivation does not — the shape the charter names as the most
contagious kind of error here. Please replace both places with something like: *"22 of 55 corpus
eboots import both NIDs, including PPSA17942 and PPSA02101; none calls either on any observed
route — no first-seen `[prosper] unimplemented:` line for them in DOLL's or Stray's boot logs."*
That is more work for the reader and it is the true statement.

Why this matters beyond tidiness: **DOLL and Stray are the two titles the sibling PR #3263 is
about.** A future agent reading "nothing observed calls these" concludes the surface is inert; the
honest record says 22 titles link it and the loud refusal has a correspondingly larger blast radius
if a route ever reaches it.

### Blocking 2 — the closer enumeration is still incomplete, at exactly one NID

The hunk exists because the old comment under-stated the closer set. The new enumeration on
`f_apr_read_gather_scatter` lists seven entries and **omits `Qs1xtplKo0U
sceAmprAprCommandBufferDestructor`**, which closes on both platform arms:

* `hle_kernel_mem.cpp:3506` → `k_ampr_destruct_320`; `:7300` → `k_ampr_destruct`
* both → `ampr_cb_destroy_320` (`:219-221`) → `prosper_apr_chain_reset(cb)`

`tests/hle/test_apr_gather_scatter.cpp`'s `closers[]` table omits it too — so the test that "pins
the set so it cannot drift silently" has a hole at the one NID the same PR newly names in
`hle_kernel_mem.cpp`'s corrected comment. This is the PR's own defect class, reintroduced at one
row. Fix is one comment line plus one `{ "Qs1xtplKo0U", "sceAmprAprCommandBufferDestructor", true }`
entry; I expect that arm to pass as written.

---

### Non-blocking

1. **The `…ResetGatherScatterState` inference is over-read.** "A Reset that already cleared the
   state would make that export redundant" assumes a coarse reset makes a fine-grained one
   pointless, and that is routinely false — `vkResetCommandPool` / `vkResetCommandBuffer` is the
   same pairing. A fine clear is useful *precisely* when you do not want to discard the buffer. So
   the export is **neutral**, not "weak evidence against". That strengthens your conclusion rather
   than weakening it: LOW is right, and the reason is that nothing bears on the question at all.

2. **"prosper's real set was already much wider than the comment" is not evidence about Sony's
   contract.** It appears in the PR body and in the new `## Ruled out` row as one of "two facts
   [that] point the other way". It is a fact about prosper, downstream of the question being asked.
   Please drop it from the evidence list; it belongs in the narrative, not the inference.

3. **You are sitting on a real over-closing null and did not cite it.** Submit already called
   `ampr_cb_reset` before this PR — `hle_kernel_mem.cpp:627` on the base commit `49ec3e354`, and the
   diff to that file is comments only. So #2924's Kiwami run was measured **with the wide set
   already live**, and reported **0 refusals** across four archives whose segments tile each file
   exactly. That is a measured absence of false refusals on the only title that exercises the path.
   Worth one sentence: it upgrades "we chose the fail-loud side" to "we chose the fail-loud side and
   measured no cost on the one title that could have paid it."
   While you are there, name the observation that *would* discriminate, so the next agent knows what
   to look for: a Reset (or submit) arriving **between** a chain-opening `ReadFile` and a
   `…GatherScatter` on the same `cb`. Recording only that the existing evidence is void leaves the
   next reader with nothing to collect.

4. **A gap in the "err toward closing" policy.** `a8uLzYY--tM sceAmprAprCommandBufferConstructor`
   re-initialises the very object the chain is keyed on (`apr_chain_open(a0, …)` at
   `hle_file.cpp:3737`/`:3752`; `k_ampr_begin`'s `a0` is the same `req`) and closes **nothing** on
   either arm (`k_ampr_begin`, `k_ampr_apr_cb_construct_stub`). Low risk in practice, since the
   observed flow is `init(req,…)` — which does close via `ampr_cb_construct` — then `begin(req,…)`.
   But this PR is the one that enumerates and justifies the set, so it is the right place to either
   add it or record in one line why it is excluded.

5. **`r.ok` → `r.in_dst` is correct and `in_dst` is not optimistic** — verified. `hle_file.cpp:3415`
   sets it only from `apr_write_guest_dst`, which returns true only on a `process_vm_writev` that
   moved all `size` bytes (or the commit-retry that did), POSIX `:3273-3322`; Windows `:3341-3346`
   returns only after `windows_prepare_guest_write` succeeds. The `r.size == 0` exemption is
   harmless — `apr_write_guest_dst` already returns true for zero bytes, so the exemption only
   covers a zero-length read with `dst <= 0xffff`.
   One consequence worth stating in the PR: the plain `ReadFile` still returns `0` on the same
   staging-fallback condition where the gather/scatter builder now returns an error. That asymmetry
   is defended in the comment and I accept it, but it is a behaviour change for any title whose
   `…GatherScatter` callsite consumes the record rather than `dst`. The FILELOG line already prints
   `guest`/`staging`, so one `PROSPER_FILELOG=1` Kiwami run and a `grep staging` would retire the
   question for free. No existing log I could find has APR FILELOG on for that title.

6. **`log_refusal` shares one 32-entry budget across all three refusal classes** (`hle_file.cpp:3839`).
   A title producing early chain-not-open refusals can silence the first UNDELIVERED one. It
   announces its own truncation, so this is a nit — but the new class is the one you most want to see.

---

### Tests

The discrimination claim checks out on inspection. On unmodified base the two `Hle::lookup` results
are null, so the two `if (gather) CHECK(...)` refusal arms never execute, the two registration
CHECKs fail, and the undelivered arm fails because base returns `r.ok ? 0 : …` = 0 — **exactly 3**,
matching your reported arm. And your claim that the closer arms passing on baseline is *intended*
discrimination is right, for the reason I verified above: submit already closed on base, so only the
comment was wrong.

Every refusal arm has its own positive control on the same open chain with an `0xEE` poison the
fixture cannot produce at those offsets; the `PROT_NONE` arm is deterministic (`mincore` succeeds on
a mapped `PROT_NONE` range, so the commit-retry branch cannot fire) and is immediately followed by
the identical segment to a writable destination. The `GetSize` non-closer arm and the per-`cb` arm
are the right two negatives. This is a good test.

### Mechanical

`git diff --check` clean. No absolute host paths in the diff or the PR body. Trap **252** is unique
and ascending, and no other open PR touches `GAME_COMPAT_ORCHESTRATION.md`;
`check_merge_result.py` is green against `origin/main` `a4d49cb4` (252 rows, none deleted) — but it
is stale the moment another lane lands, so re-run it immediately before merging.

**The cross-PR conflict is two files, not one.** `git merge-tree` of `cc7480488` and #3263's
`b7ede68b1` conflicts in **`prosper/CMakeLists.txt`** as well as `docs/UE4_APR_IOSTORE_BRINGUP.md`
— both are additive-at-the-same-anchor and trivially resolved by keeping both blocks.
`src/hle/fs/hle_file.cpp` auto-merges cleanly and I checked the merged blob: both feature sets
survive intact, no conflict markers.

---

Taken on trust, not reproduced: your ctest counts (344 / 347), the two mutation-arm exit codes, and
#2924's Kiwami disassembly.

Fix the two blocking items and I will re-review the corrected head.

mattias800:
### Exact-head verification for `1efd2dbc4` (the corrected head)

The local box is unusable for builds right now — several concurrent lanes have podman's database
locked (`Error: beginning transaction: database is locked` on every `distrobox enter` and every
`podman exec`, sustained for well over an hour), so I have not run the local rebuild at this commit
and am not quoting numbers I did not measure. **CI is the exact-head signal, and it is the stronger
one here**: clean runners, three platforms.

All checks complete: **9 SUCCESS, 2 SKIPPED (Progress tracker, Publish Release), 0 failed.**

Linux job, full suite with `--no-tests=error`:

```
180/345 Test #180: apr_gather_scatter .............................   Passed
100% tests passed, 0 tests failed out of 345
```

345 rather than my local 347 because the CI Linux job configures without a game dump, so the
dump-gated cases are not registered; the delta of **+1 over that job's own base is this PR's one new
test**, and it ran **by name**. Windows MinGW and macOS (x86_64 / Rosetta 2) also green.

The mutation arms quoted in the PR body were run locally *before* the lock set in, at the code state
that shipped, and each is reported with the failing arm names and its exit code. A local retry is
running; I will post the local counts here if it clears.

mattias800:
**Local verification cleared** (the podman lock released). Exact head `1efd2dbc4`, container `ps5ys`, Linux:

```
cmake --build prosper/build-linux -j4            # BUILD_RC=0
ctest --no-tests=error                           # CTEST_RC=0
100% tests passed out of 347
```

347 is unchanged from the pre-correction head, and that is the expected result: the `Qs1xtplKo0U`
fix adds two **assertions inside** the existing `apr_gather_scatter` case, not a new ctest case. The
arm itself was already shown above to run by name and to fail under mutation.

Both signals now agree at this head — local 347/347 exit 0, CI Linux 345/345 exit 0 (the CI job
configures without a game dump, so its base is smaller).

mattias800 (commented):
## Re-review — #3262 at `1efd2dbc4`

**APPROVED.** Both blocking findings are fixed, one of them better than I asked for. Two
non-blocking notes below, one of which is on the new code.

### Verified fixed

**Reachability.** Independently re-swept: **22 of 55 eboots import both NIDs**, and the title list in
the comment matches mine exactly. The replacement framing is the right one — it does not just state
the corrected number, it makes the absence of a *call* falsifiable by naming `prosper_on_unimpl`'s
unconditional first-seen print as the instrument. "So the blast radius of this refusal is 22 linked
titles rather than none" is the sentence a future reader needs.

**`Qs1xtplKo0U`** is in the enumeration and in `closers[]` with `required: true`. Confirmed it does
close on both arms (`hle_kernel_mem.cpp:3506` / `:7300` → `ampr_cb_destroy_320` →
`prosper_apr_chain_reset`), so the new arm should pass as written.

**`…ResetGatherScatterState` → neutral**, with the `vkResetCommandPool` / `vkResetCommandBuffer`
parallel, in both the comment and the `## Ruled out` row — and the row now also says explicitly that
prosper's own closer set is a fact about prosper and not evidence about Sony's contract. "Nothing
bears on the question at all, which is exactly why LOW is right" is a stronger and more useful
position than the one it replaces.

**The discriminating observation is now named** (a Reset or submit arriving *between* a
chain-opening `ReadFile` and a `…GatherScatter` on the same `cb`), and #2924's zero-refusal Kiwami
measurement is cited with the reason it counts — submit already closed, so that run had the wide set
live. Both were suggestions; taking them makes the row collectable rather than merely honest.

**The return-type derivation went past what I gave you, and it checks out.** I cited two sizing
exports; you found four. Verified against `../PS5-3.20_Libs/libSceAmpr.c`:

```
vWU-odnS+fU sceAmprMeasureCommandSizeReadFile          qesF88X4DRg …ReadFileGather
DXmgc5op8Yw …ReadFileGatherScatter                     7nXGDGMXSqo …ReadFileScatter
```

`grep -oE 'sceAmprMeasureCommandSizeReadFile[A-Za-z]*'` returns exactly those four names and no
others — a one-to-one cover of the four read builders. That is a real derivation, not a plausibility
argument: a builder cannot need to return its own encoded size when sizing has its own entry point
per builder. Together with the behavioural corroboration from Kiwami it earns the `CONFIDENCE: HIGH`
it now carries.

**`a8uLzYY--tM` listed as a deliberate non-closer** rather than an omission, with the tension against
"err toward closing" stated instead of resolved, and the condition for revisiting it. That is the
right way to record a known gap.

### Non-blocking — the per-class refusal budgets

Reviewed as the largest unreviewed change. The mechanism is correct: two separate `static
std::atomic<uint64_t>`, a capture-less `log_refusal_in` taking the budget by reference (no lifetime
question), `continue` on the tombstone-equivalent path still running the loop's iteration
expression, and the truncation notice correctly narrowed to "further refusals **of this kind**
silent". The stated goal is achieved — chain-not-open refusals can no longer silence the first
UNDELIVERED one.

**But it is two budgets across three refusal sites, and the comment says "PER-CLASS".**
`:3879` (no chain open) and `:3888` (the range guard — *the chain names the wrong file*) both route
through the `log_refusal` alias into `refusals_chain`; only `:3921` has its own. So 32 chain-not-open
refusals still silence the first wrong-file refusal, and that is arguably the highest-signal line of
the three: it is the range guard reporting that it caught data about to be served from the wrong
file. Either give it a third budget, or rename to say what the split actually is (chain *identity*
vs *delivery*) so the comment stops promising one-per-class. Comment-accuracy on a logging path, so
not a blocker — but this PR's whole subject is comments that claim more than the code does.

### Non-blocking — pin the non-closer decision the way the closers are pinned

`test_apr_gather_scatter.cpp` has a non-closer arm for `tZDDEo2tE5k`. Now that the comment
deliberately excludes `a8uLzYY--tM`, an arm asserting the chain *survives* it would pin that
decision the same way `closers[]` pins the inclusions — and would go red if someone later "fixes"
the tension by making it close, which the comment invites them to consider.

### Mechanical

`git diff --check` clean. No absolute host paths in the diff. Merges clean against `origin/main`.
Against #3263's `ecbbddea4` there are conflicts in `prosper/CMakeLists.txt` and
`docs/UE4_APR_IOSTORE_BRINGUP.md` — both additive at the same anchor, trivially resolved by keeping
both blocks; `src/hle/fs/hle_file.cpp` auto-merges and I checked the merged blob (4,495 lines, no
conflict markers, both PRs' feature sets intact). Re-run `check_merge_result.py` immediately before
merging — it was green earlier but goes stale on any lane landing.

Taken on trust, not reproduced: 347/347 local, CI 9 SUCCESS / 2 SKIPPED.

Neither note blocks. **APPROVED** — merge when the trap-table gate is re-confirmed at merge time.

PR #3263, opened 2026-09-02
APR measure-size: retire the compatibility read per container once a submit re-delivers it

### LINKED ISSUE #3245: sceAmprMeasureCommandSizeReadFile's compatibility read runs for every title, so every streamed asset is read and DMA'd twice (13,925 duplicate reads / 582 MB in a 3-minute Stray boot)

mattias800:
## What

`sceAmprMeasureCommandSizeReadFile`'s compatibility read fired for **every** APR title, so every streamed asset was read and DMA'd into the guest twice. Fixes #3245.

## Failure scenario

That NID's firmware contract is pure **sizing** — it answers 20, or 24 when the file offset needs its high 32 bits. prosper *also* performs the described read and DMAs it into the guest destination, deliberately: *Dragon Quest VII Reimagined*'s (`PPSA17942`) older SDK wrapper consumes that destination and never records a `ReadFile` command for it, so the measure is prosper's only hook on those bytes (`docs/UE4_APR_IOSTORE_BRINGUP.md`).

That requirement is real. The gate around it was not — it was only *"does the file id resolve"*.

Measured on *Stray* (`PPSA02101`), `PROSPER_NULL_PAGE=1`, no input, `tools/screenshot`, ~180 s, `PROSPER_FILELOG=1` (from #3245):

| | |
| --- | --- |
| `measure-read compatibility … OK` | **13,925** |
| `read-submit … OK` | 14,255 |
| bytes delivered by the measure path | **582,619,946** |
| bytes delivered by the submit path | 603,329,577 |

Every measure was paired with a submit of the same `dst`/`off`/`size` moments later — roughly **97% of the measure-path bytes re-delivered** — so the boot paid for ~583 MB of extra host reads, ~583 MB of extra `process_vm_writev`, and one extra `mmap`/`munmap` per read.

**The cost is the smaller half.** A sizing call that fills the destination puts bytes there *before the guest has recorded the read command*, let alone submitted it. On Stray that window is microseconds on one thread, so nothing observable goes wrong. On a title that measures a batch of reads and prepares the destinations afterwards, prosper's bytes would land first and the guest would then overwrite data the hardware would have delivered later. That is a silent-corruption shape, not a slowdown. It also perturbs every instrument on the streaming path — `PROSPER_ZEROWATCH`, `PROSPER_APR_VERIFY`, the dmem write trace all see each destination written twice.

## The contract asserted

**Nothing at the measure call distinguishes the two titles.** DOLL and Stray both pass a valid id, a real destination and a real size. What separates them is visible immediately *afterwards*, so that is what this keys on:

> When a submit arrives for exactly the `(file id, dst, offset, size)` a measure just served, that container records and submits its measured reads — so the compatibility read was redundant, and is retired for that container.

This is **self-calibrating**: it costs exactly one duplicate per APR file id and then turns itself off, and needs no per-title and no per-SDK knowledge.

### Correction — DOLL is NOT exempt, and the change is still right

The first version of this PR said in three places that the heuristic *"cannot fire for DOLL, whose measured reads are never submitted through the ReadFile NID"*. **That is false, and DOLL's own boot logs say so.** Independent review found it; I reproduced the replay myself over four `PROSPER_FILELOG` boots, implementing the predicate exactly as the code does (64-entry ring, record only on a delivered measure, match on `(id, dst, offset, requested)`):

| DOLL run | measures | pairs at | suppressed | **measures with no identical submit** |
| run1 | 1,254 | submit #13 | 1,253 | **0** |
| run2 | 1,255 | submit #13 | 1,254 | **0** |
| run3 | 1,725 | submit #13 | 1,724 | **0** |
| run4 | 1,732 | submit #13 | 1,731 | **0** |

Hand-verified at the cited lines of `run1` — `measure-read compatibility id=6 … dst=0x3002860000 off=0x7adec90a size=1080890 OK` at line 648, and `read-submit id=6 … dst=0x3002860000(guest) off=0x7adec90a size=1080890 got=1080890 OK` at line 704, with the raw `descsz=0x107e3a` = 1,080,890 in between.

**So the mechanism by which DOLL is safe is the opposite of the one asserted.** DOLL is not a consume-from-the-measure title; it is a *submits-everything-it-measures* title — exactly the case this suppression is designed for. When the heuristic fires on it, it is right. Stray replays identically (id=1 pairs at submit #4, 13,922 of 13,923 suppressed, **0 unmatched** across three logs — which also reproduces this issue's headline 13,925/14,255 figure independently at 13,923/14,253).

**The consequence, stated rather than buried:** the documented reason this compatibility read exists — an older SDK wrapper consuming the destination while prosper does not execute the encoded commands — **is no longer visible in DOLL's traffic.** prosper serves the `ReadFile` submit eagerly now, and DOLL does submit. Whether the read is needed *at all* is what `PROSPER_APR_MEASURE_READ=never` tests. Removing it is not this PR's job; not re-asserting a premise its own evidence contradicts is.

Keyed **per APR file id** rather than globally, so a title with a mixed pattern loses the read only for the container that demonstrated the pairing. `PROSPER_APR_MEASURE_READ=auto|always|never` (default `auto`) is the one-variable A/B lever.

### One thing not to do, recorded because it is the obvious move

#3245 suggests *"only when the same (id, offset, size, dst) has not just been served"*. Taken as an order-agnostic duplicate filter that is **backwards**: on Stray the measure comes **first** and the submit is the authoritative delivery, so such a filter suppresses the wrong half. The pairing is used here purely as *evidence* that the container submits what it measures — never to skip a submit.

## Deliberately unaffected

* **The sizing answer.** Still 20 or 24 in every case, including when the read is suppressed. An I/O outcome never becomes an allocation size.
* **The pure unregistered-id query** (Sonic's sizing calls) — unchanged.
* **The submit path's own delivery** — never suppressed, never reordered, never skipped. It remains the authoritative read.
* **A measure with no following submit still delivers, always** — the consume-from-the-measure pattern. (This bullet used to name DOLL; per the correction above, DOLL is not that pattern. The guarantee is about the mechanism, not the title.)
* Windows and POSIX take the same path; the suppression is host-independent.

## Risks

The one case this heuristic would get wrong is a title that, **within a single container**, both submits some measured reads and consumes others from the measure without submitting. Seven logs across two titles now show **0 unmatched measures**, so the pattern is not merely unobserved — it has been looked for. Two titles is still not the corpus.

**That case is now DETECTED, not merely feared** (review suggestion, taken). The suppression's *decision* was already loud; its *consequence* would have been silent, because the suppressed-measure log line sits behind `filelog()` — so a mixed-pattern title would just get an unwritten destination and no message. `PROSPER_APR_MEASURE_READ=always` is a recovery, not a detector: it only helps someone who already suspects this.

So recording now **continues after a container pairs** (the read is skipped; only the bookkeeping stays), each ring entry carries whether a submit later claimed it, and an entry that leaves the ring **suppressed and never matched** warns once, loudly, naming the container, the lost range and the lever. Nothing else produces that signature: a pre-pairing entry ages out unmatched routinely and is not marked suppressed, and a suppressed entry a submit claimed is the designed case. Cost is a 64-entry compare per submit — ~900k integer comparisons across a 3-minute Stray boot, against the 583 MB of file I/O removed. Known blind spot, stated in the code: up to 64 entries are still in the ring at exit and are never adjudicated.

Confidence: **HIGH** on the duplication and on suppression being safe for a title whose measured reads are submitted — which now includes DOLL and Stray *by replay*. **MED** that no title mixes both patterns in one container, since a log replay cannot see whether a title depends on the bytes arriving at *measure* time rather than *submit* time.

## Verification

Container `ps5ys`, Linux, exact head `b7ede68b1` on `origin/main` `49ec3e354`.

```
cmake -S prosper -B prosper/build-linux            # CONFIGURE_RC=0
cmake --build prosper/build-linux -j6              # BUILD_RC=0
cd prosper/build-linux && ctest --no-tests=error   # CTEST_RC=0
```

| run | tests | result | exit |
| --- | --- | --- | --- |
| `origin/main` baseline (measured here, at `e13fd40af`) | 344 | 100% passed | 0 |
| this branch, exact head | **348** | **100% passed** | **0** |

(`origin/main` moved to `49ec3e354` mid-session, bringing 2 tests of its own; 348 = 346 + this PR's 2. The same 346 falls out independently from #2926's branch — 347 there with 1 new test.)

### Mutation arms — demonstrated by exit code

The test is registered **twice**: once by default and once under `PROSPER_APR_MEASURE_READ=always`, where it asserts the **opposite** outcome for the Stray case. A suppression that never suppressed cannot pass both.

**Arm 1 — suppression disabled** (`if (false && apr_measure_read_suppressed(...))`, then restored):

```
default mode              ->  exit 1
  [FAIL] Stray arm: the compatibility read is suppressed once the container has paired
PROSPER_APR_MEASURE_READ=always  ->  exit 0     (unchanged, as it must be)
```

**Arm 2 — suppression made unconditional**, i.e. the actual DOLL risk (`return true`, then restored):

```
default mode              ->  exit 1
  [FAIL] DOLL arm: a measure with no submit delivers the container's bytes
  [FAIL] DOLL arm: a SECOND unsubmitted measure still delivers (no spurious latch)
  [FAIL] near-miss arm: the measure delivered
  [FAIL] near-miss arm: a non-identical submit does NOT retire the compatibility read
  [FAIL] Stray arm: the first measure delivers (the one duplicate this costs)
  [FAIL] a DIFFERENT container keeps its compatibility read after the first one paired
  [FAIL] prosper_apr_reset_for_test clears the pairing latch as well as the registry
```

So the test brackets the change on **both** sides: too little suppression and too much.

### Review round 2 — new arms

- The **"DOLL arm" is renamed** to the *unsubmitted-measure arm*. It models "a measure with no submit", which is now known not to be DOLL; it is a fine mechanism arm and should not carry title-level authority it has not earned. The file header is corrected likewise.
- **Two detector arms.** Arm A: after pairing, 72 suppressed measures that ARE each submitted → must stay **quiet** (a detector that cries wolf on the designed case is one nobody reads). Arm B: 72 suppressed measures never submitted → must **warn**.
- **The detector arm caught a flaw in itself, which is worth reporting.** Under `PROSPER_APR_MEASURE_READ=always` the first version of arm B failed — correctly: nothing is suppressed in that mode, so no measure can be lost and the detector *must* stay silent. The arm now inverts under `=always` and asserts the detector **cannot false-positive**, which is a better arm than the one I wrote.
- **A third CMake registration under `PROSPER_APR_MEASURE_READ=never`**, the one branch nothing previously executed.

### Why the new test is not vacuous

Every delivery assertion **pre-poisons the destination with `0x5A`**, a byte the fixture cannot produce at that offset, so "the callee wrote nothing" and "the callee wrote the right thing" are distinct observations rather than both reading as zero. The **near-miss arm** is the one that stops the pairing from being keyed too loosely: a submit on the same container and the same destination at a *different offset* must **not** retire the read — without it, a pairing keyed on the file id alone would pass every other arm. A **second container** arm pins the per-container scoping, and a **reset arm** pins that the test hook clears the latch, so the arms are not silently order-dependent.

## Ruled out

One row added to `docs/UE4_APR_IOSTORE_BRINGUP.md` § Ruled out — *"the compatibility read is DOLL-specific in effect, because only DOLL needs it"*, with the Stray measurement and the order-agnostic-filter warning. The doc's 2026-07-26 SOLVED section, which stated the read fires for any registered id, is narrowed in place rather than left to contradict the code.

Snapshots: not run — this changes no rendering path.

Note for the merge order: this branch and #3262 both append to the same `## Ruled out` section, so whichever merges second needs a trivial text rebase.

### Verification of the corrected head `ec69ad72d`

The local box is currently unusable for builds: several concurrent lanes have podman's database
locked (`Error: beginning transaction: database is locked` on every `distrobox enter` and every
`podman exec`, for over an hour), so I could not run the local rebuild + full ctest at this exact
head. **I am not quoting numbers I did not measure.** What I can state:

* the pre-correction head's local run is unchanged and quoted above;
* the corrections carrying the falsified DOLL claim, the renamed arm, and the new detector were each built and run locally *before* the lock set in, and their
  results are reported in this body with the arms named;
* **CI is the exact-head signal for this commit** — it builds and runs the full ctest on Linux,
  Windows MinGW and macOS on clean runners, which is a stronger check than my contended box.

A local retry is running and I will post the local counts to this PR if it clears.

## Round 3 — the O(1) index (`ecbbddea4`)

`apr_measure_note_submit` had no early-out after the detector landed, because marking every entry is
what the detector needs — so it linear-scanned all 64 ring slots under the mutex on **every** submit
(14,255 of them in a 3-minute Stray boot). Replaced with an open-addressed
`(id, dst, offset, size)` index over the same 64 slots: **O(1) expected lookup, every entry still
marked, detection unchanged.** The sampling alternative for the measure-side lock is deliberately
*not* taken — a detector that only sometimes detects is the wrong trade when detection is why this
design is acceptable at all.

Design notes: the encoding (`0` empty, `-1` tombstone, `n>0` slot `n-1`) needs no dynamic
initialisation, so the table is correct from static zero-init. Each ring slot remembers where the
index points at it, so eviction is O(1) rather than a search. The table is 4x the ring (live load
≤ 0.25) and is rebuilt from the live ring when occupancy would pass half — at least 64 inserts
apart, so under 4 amortised steps per insert against the 64 per submit it replaces.

### The tests needed more than a count, and the first version of them was blind

An open-addressed lookup fails by **finding nothing for a key that is present** — a bad hash, a
probe chain terminated early at a tombstone, a stale slot pointer after a rebuild. None of those
throws, and after pairing none of them changes a return value anything checks: a missed lookup just
silently stops marking entries. Three arms now bracket it:

| mutation (applied, then restored) | result |
| --- | --- |
| lookup always returns "not found" | **exit 1** — `Stray arm: … suppressed once the container has paired`, `detector: warns when a suppressed measure is never submitted` |
| **tombstone terminates the probe chain** (the classic open-addressing bug) | **exit 1** — `probe-collision arm: B was FOUND behind the tombstone, not lost silently` |
| rebuild drops its live entries | **exit 1** — `detector quiet + index found every key across many wraps and rebuilds` |

**The middle one is the one worth reading, because my first attempt at the test did not catch it.**
I raised the stress arm to 600 iterations expecting it to cover the tombstone path, added coverage
counters to check, and got **6 rebuilds but 0 tombstone steps**. That zero is real rather than an
instrument artefact: 64 live keys in a 256-slot table means ordinary traffic almost never crosses a
deleted entry. So the tombstone mutation *passed* against the stress arm alone — a silent
degradation, exactly the class this PR exists to catch, sailing through a test that looked thorough.

Per `CLAUDE.md`'s rule for a clean zero, I constructed one positive instance **by hand, outside
whatever produced the null**: a bucket seam lets the test find two keys sharing a home bucket, place
the second behind the first, age the first out of the ring to leave a tombstone in front of it, and
then look the second up. Coverage across that arm reads `tomb steps 0 -> 0 -> 1`, incrementing at
exactly the load-bearing lookup, and the mutation now fails on it.

The stress arm's coverage assertion was narrowed to what it actually proves (`rebuilds > 0`) rather
than left claiming a path it never reached.

### `=always` is a control, and the test now says so

Under `PROSPER_APR_MEASURE_READ=always` the index is never consulted — `note_submit` returns at its
mode check *before* taking the lock. The test asserts that rather than skipping the arm, because it
is the same early return that makes the `=always` arm a valid control for this change in an A/B.

### Verification

Exact head `ecbbddea4`, container `ps5ys`, Linux:

```
cmake --build prosper/build-linux -j6            # BUILD_RC=0
ctest --no-tests=error                           # CTEST_RC=0
100% tests passed out of 349
179/349 apr_measure_duplicate ........... Passed
180/349 apr_measure_duplicate_forced .... Passed
181/349 apr_measure_duplicate_never ..... Passed
```

349 unchanged from the previous head — the new arms are assertions inside the existing three cases,
not new ctest cases. All three modes also run directly: exit 0 / 0 / 0.

## Round 4 — the `never` false positive (`6a8594e74`), and a rebase onto merged #3262

**The detector was a guaranteed false positive under `PROSPER_APR_MEASURE_READ=never`, on every run of every title.** `apr_measure_read_suppressed` returns true for `Never`, so every measure recorded `suppressed=true`; `apr_measure_note_submit` returns at its own mode check *before* the lock, so nothing was ever marked; the eviction check therefore fired on the **65th measure** and printed a message blaming a title-specific mixed pattern and asking the user to report the title — when the cause was their own flag.

This is the `=always` "cannot cry wolf" argument one mode over, and I had already made it there. `always` was safe only by the coincidence that it never sets `suppressed=true` at all; relying on that coincidence is how `never` was missed. **The warning is now gated on the detector's actual premise** — "a submit should have claimed this entry", which is a property of Auto alone — rather than on the symptom.

It matters more than an ordinary false positive: **`never` is the lever for the open question this PR raises** (is the compatibility read needed at all?), so a warning on every title landed precisely on the one experiment the feature exists to enable.

The `never` test arm missed it because it returned after **two** measures, far short of the 64-entry ring. It now drives **200**, well past the wrap.

### Mutation arm — targeted, and it reproduces the shipped defect exactly

Removing the Auto gate, then restoring:

```
PROSPER_APR_MEASURE_READ=<unset,auto>   EXIT=0
PROSPER_APR_MEASURE_READ=always         EXIT=0
PROSPER_APR_MEASURE_READ=never          EXIT=1
  [FAIL] PROSPER_APR_MEASURE_READ=never: the detector cannot cry wolf past the ring wrap
```

One false warning emitted in the mutated `never` run, zero in the fixed one. Only `never` moves, which is the confirmation that the defect was `never`-only.

### Three review notes taken

- **The probe-collision arm's hidden precondition is now asserted, not commented.** No rebuild may intervene between B's insert and B's lookup — a rebuild re-inserts only live entries, clearing the very tombstone the arm probes past and silently turning it into a passing no-op. Threshold is `live + tombs > 128`; the arm reaches 66, so headroom is ~62 records. A later change to the ring size or the threshold would break the arm invisibly, which is the same silent-degradation class the arm exists to catch — so it is a `CHECK`, not a comment.
- **The stress arm's zero tombstone steps is structural, not sampled.** Each pass records a key then immediately looks up the same key; insert places at the first empty-or-tombstone while find stops only at empty, so find always reaches the entry at or before insert's position and a tombstone can never sit between them. **Zero is the only value that workload can produce** — which is why the hand-built collision is the only route to the path, and why measuring the zero harder would only have confirmed it.
- **`static_assert`s** pin the index encodings that fail silently: `slot+1` fitting `int16_t`, the table being a power of two for mask probing, and the table being at least twice the ring.

Duplicate keys are noted rather than deduplicated — two live slots with the same tuple would leave one unmatched and warn. **Latent, not live:** 5,966 measured reads across four DOLL boots produced 5,966 distinct tuples and **0** duplicates. Deduplicating would mean deciding which of two identical pending reads a single submit satisfies, and nothing observed poses that question.

### Rebase

Rebased onto `origin/main` with #3262 merged. Both conflicts (`prosper/CMakeLists.txt`, `docs/UE4_APR_IOSTORE_BRINGUP.md`) were additive at the same anchor and resolved by keeping both blocks, hand-merged rather than taking either file whole (trap 41); `git diff origin/main` shows no deletion of any #3262 line.

### Verification at `6a8594e74`

Local, container `ps5ys`, Linux — build exit 0, **ctest exit 0, `100% tests passed out of 351`**, run against a build confirmed converged (two further `cmake --build` invocations do zero work and the binary sha is unchanged). CI: **9 SUCCESS, 2 SKIPPED, 0 failed**; Linux `349/349` with `apr_gather_scatter`, `apr_measure_duplicate`, `_forced` and `_never` all passing by name.

### On the performance A/B — two series, not one

The coordinator's three valid rounds are on **`ecbbddea4`**, which is **not** the head that will merge:

```
ecbbddea4   auto-r1 306   always-r1 304   auto-r2 328    submit failures: 0 of 3
```

against 2 of 5 `auto` runs before the index. Any six-round result on **`6a8594e74`** is a **separate series** and should be reported as one — these numbers must not be read as a single run of alternating arms. That conflation is the exact error this PR's review history already contains once, and it would be a poor thing to repeat here.

Fixes #3245

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -1595,6 +1595,9 @@ if(TARGET prosper_core)
   add_test(NAME apr_measure_duplicate_forced COMMAND test_apr_measure_duplicate)
   set_tests_properties(apr_measure_duplicate_forced PROPERTIES
                        ENVIRONMENT "PROSPER_APR_MEASURE_READ=always")
+  add_test(NAME apr_measure_duplicate_never COMMAND test_apr_measure_duplicate)
+  set_tests_properties(apr_measure_duplicate_never PROPERTIES
+                       ENVIRONMENT "PROSPER_APR_MEASURE_READ=never")
 
   add_executable(test_ampr_measure tests/hle/test_ampr_measure.cpp)
   # Includes a shared fixture, which resolves from the `tests` root.
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -1581,6 +1581,20 @@ if(TARGET prosper_core)
   target_include_directories(test_apr_gather_scatter PRIVATE tests)
   target_link_libraries(test_apr_gather_scatter PRIVATE prosper_core)
   add_test(NAME apr_gather_scatter COMMAND test_apr_gather_scatter)
+  # #3245: sceAmprMeasureCommandSizeReadFile's compatibility read fired for every title, so a title
+  # that also submits its measured reads paid for the same bytes twice (13,925 duplicate reads /
+  # 582 MB in a 3-minute Stray boot) and had them written before it recorded the command. Registered
+  # twice on purpose -- the PROSPER_APR_MEASURE_READ=always run asserts the opposite outcome, which
+  # is what shows the suppression actually suppressed. Pure: scratch files and generated import
+  # stubs, no game dump.
+  add_executable(test_apr_measure_duplicate tests/hle/test_apr_measure_duplicate.cpp)
+  # Includes a shared fixture, which resolves from the `tests` root.
+  target_include_directories(test_apr_measure_duplicate PRIVATE tests)
+  target_link_libraries(test_apr_measure_duplicate PRIVATE prosper_core)
+  add_test(NAME apr_measure_duplicate COMMAND test_apr_measure_duplicate)
+  add_test(NAME apr_measure_duplicate_forced COMMAND test_apr_measure_duplicate)
+  set_tests_properties(apr_measure_duplicate_forced PROPERTIES
+                       ENVIRONMENT "PROSPER_APR_MEASURE_READ=always")
 
   add_executable(test_ampr_measure tests/hle/test_ampr_measure.cpp)
   # Includes a shared fixture, which resolves from the `tests` root.
--- a/prosper/docs/UE4_APR_IOSTORE_BRINGUP.md
+++ b/prosper/docs/UE4_APR_IOSTORE_BRINGUP.md
@@ -231,6 +231,20 @@ default since #825 and needs no switch; `PROSPER_NO_GUEST_FS=1` turns it off for
   neither NID string occurs anywhere in DOLL's 91 MB or Stray's 33 MB `PROSPER_FILELOG` boot logs.
   So the refusal's blast radius is 22 linked titles rather than none.
   [#2926](https://github.com/mattias800/prosper/issues/2926).
+- **"The measure-size call's compatibility read is DOLL-specific in effect, because only DOLL needs
+  it" — false, measured 2026-09-02.** `f_apr_measure_read_file` gated the read on nothing but "does
+  the file id resolve", so it ran for every APR title. *Stray* (`PPSA02101`), 180 s, `PROSPER_FILELOG=1`:
+  **13,925** compatibility reads delivering **582,619,946 bytes**, against 14,255 submits delivering
+  603,329,577 — roughly **97% of the measure-path bytes re-delivered** by the submit path from an
+  identical `(dst, offset, size)` moments later, plus an extra `mmap`/`munmap` per read. The
+  correctness half matters more than the cost: a *sizing* call filling the destination puts bytes
+  there before the guest has recorded the read command, so a title that measures a batch and
+  prepares its destinations afterwards would have prosper's bytes land first and then be overwritten.
+  Also **do not "fix" this by suppressing the second delivery of an identical pair**: the measure
+  comes FIRST and the submit is the authoritative one, so an order-agnostic duplicate filter
+  suppresses the wrong half. The pairing is used only as evidence that the container submits what it
+  measures, and the suppression is keyed per container.
+  [#3245](https://github.com/mattias800/prosper/issues/3245).
 
 ## Frame loop reached; 0 draws = early-load present loop (issue #213, 2026-07-09)
 
@@ -1459,7 +1473,12 @@ The shared contract is now explicit:
   query; DOLL's non-null queue/id pair also registers the APR completion source.
 - `vWU-odnS+fU` always returns 20 or 24 according to the file-offset width. An unregistered id is a
   pure Sonic sizing query; a valid registered APR file id also performs DOLL's range-checked eager
-  read on both Linux and Windows.
+  read on both Linux and Windows. **Narrowed 2026-09-02 (#3245):** that eager read is now retired
+  per APR container once a submit is seen re-delivering a read the measure already served, because
+  it fired for every APR title and duplicated 582 MB in a 3-minute *Stray* boot. DOLL's arm is
+  untouched — it never submits the reads it measures, which is why the read exists. The sizing
+  answer is unchanged in every case, and `PROSPER_APR_MEASURE_READ=always|never|auto` is the A/B
+  lever.
 - `4fgtGfXDrFc` retains the independent 32-byte conservative size used by Sonic; A/B testing proved
   it is not part of DOLL's regression.
 
--- a/prosper/docs/UE4_APR_IOSTORE_BRINGUP.md
+++ b/prosper/docs/UE4_APR_IOSTORE_BRINGUP.md
@@ -244,6 +244,16 @@ default since #825 and needs no switch; `PROSPER_NO_GUEST_FS=1` turns it off for
   comes FIRST and the submit is the authoritative one, so an order-agnostic duplicate filter
   suppresses the wrong half. The pairing is used only as evidence that the container submits what it
   measures, and the suppression is keyed per container.
+
+  **A second claim in this area was falsified before it could become load-bearing, and is recorded
+  so it is not re-derived:** *"the suppression cannot fire for DOLL, whose measured reads are never
+  submitted through the ReadFile NID"* — false. Replaying the predicate over four DOLL
+  `PROSPER_FILELOG` boots, container id=6 pairs at **submit #13** every time. What makes DOLL safe is
+  the opposite fact, from the same replay: **0 of 1,254 / 1,255 / 1,725 / 1,732 measured reads lack
+  an identical submit**, so DOLL is a submits-everything-it-measures title — exactly the case the
+  suppression is designed for — not the consume-from-the-measure title the older note describes.
+  *Stray* replays the same way (id=1 pairs at submit #4, 0 unmatched over three logs, 13,923 measures
+  against 14,253 submits — which also reproduces this issue's headline figure independently).
   [#3245](https://github.com/mattias800/prosper/issues/3245).
 
 ## Frame loop reached; 0 draws = early-load present loop (issue #213, 2026-07-09)
@@ -1475,10 +1485,19 @@ The shared contract is now explicit:
   pure Sonic sizing query; a valid registered APR file id also performs DOLL's range-checked eager
   read on both Linux and Windows. **Narrowed 2026-09-02 (#3245):** that eager read is now retired
   per APR container once a submit is seen re-delivering a read the measure already served, because
-  it fired for every APR title and duplicated 582 MB in a 3-minute *Stray* boot. DOLL's arm is
-  untouched — it never submits the reads it measures, which is why the read exists. The sizing
-  answer is unchanged in every case, and `PROSPER_APR_MEASURE_READ=always|never|auto` is the A/B
-  lever.
+  it fired for every APR title and duplicated 582 MB in a 3-minute *Stray* boot. The sizing answer
+  is unchanged in every case, and `PROSPER_APR_MEASURE_READ=always|never|auto` is the A/B lever.
+
+  **DOLL is not exempt, and the paragraph above this one is now partly historical.** Replaying the
+  pairing predicate over four of DOLL's own `PROSPER_FILELOG` boots: container id=6 pairs at submit
+  **#13** in every run, and 1,253 / 1,254 / 1,724 / 1,731 compatibility reads are suppressed from
+  there. It is safe anyway, for the *opposite* reason to the one first written down — **every**
+  measured read has an identical submit (1,254/1,254, 1,255/1,255, 1,725/1,725, 1,732/1,732; **0
+  unmatched**), so DOLL submits everything it measures and the submit path delivers the same bytes.
+  The consequence deserves stating: **the reason this read is documented to exist is no longer
+  visible in DOLL's traffic** — prosper serves the ReadFile submit eagerly now, and DOLL does
+  submit. Whether the read is needed at all is what `PROSPER_APR_MEASURE_READ=never` tests; that is
+  a separate question from narrowing it.
 - `4fgtGfXDrFc` retains the independent 32-byte conservative size used by Sonic; A/B testing proved
   it is not part of DOLL's regression.
 
--- a/prosper/src/hle/fs/hle_file.cpp
+++ b/prosper/src/hle/fs/hle_file.cpp
@@ -2346,26 +2346,57 @@ extern "C" void prosper_apr_chain_reset(uint64_t cb) {
 // same argument shape. It IS visible immediately afterwards -- if a submit arrives for exactly the
 // (file id, dst, offset, size) a measure just described, that title records and submits its
 // measured reads, so the compatibility read was redundant. That is self-calibrating: it costs one
-// duplicate per APR file id and then turns itself off, needs no per-title or per-SDK knowledge, and
-// cannot fire for DOLL, whose measured reads are never submitted through the ReadFile NID -- if
-// they were, prosper would already have served them eagerly there and the compatibility read would
-// never have been needed.
+// duplicate per APR file id and then turns itself off, and needs no per-title or per-SDK knowledge.
+//
+// DOLL IS NOT EXEMPT FROM THIS, and an earlier version of this comment claimed it was -- "cannot
+// fire for DOLL, whose measured reads are never submitted through the ReadFile NID". That is
+// FALSE, and DOLL's own PROSPER_FILELOG boots say so. Replaying this exact predicate over four of
+// them: container id=6 pairs at submit #13 in every run, and from there 1,253 / 1,254 / 1,724 /
+// 1,731 of each run's compatibility reads are suppressed.
+//
+// The same replay is why the suppression is nonetheless SAFE for DOLL -- for the opposite reason to
+// the one that was asserted. Asking whether any measured read lacks an identical submit ANYWHERE in
+// the log, which is precisely the mixed pattern that would break:
+//
+//     run1  1,254 measures  1,254 matched  0 unmatched
+//     run2  1,255           1,255          0
+//     run3  1,725           1,725          0
+//     run4  1,732           1,732          0
+//
+// DOLL submits everything it measures, so when the heuristic fires on it, it is right. Stray is the
+// same shape (id=1 pairs at submit #4; 13,922 of 13,923 suppressed; 0 unmatched over three logs).
+//
+// The consequence is worth stating rather than burying: the documented reason this read exists --
+// an older SDK wrapper that consumes the destination while prosper does not execute the encoded
+// commands -- IS NO LONGER VISIBLE in DOLL's traffic. prosper serves the ReadFile submit eagerly
+// now, and DOLL does submit. Whether the read is needed at all is what PROSPER_APR_MEASURE_READ=never
+// tests; removing it is not this change's job, but re-asserting a premise its own evidence
+// contradicts would have been a defect.
 //
 // Keyed per APR FILE ID rather than globally so a title with a mixed pattern loses the
 // compatibility read only for the container that demonstrated the pairing.
 // CONFIDENCE: HIGH on the duplication (measured, #3245) and on the suppression being safe for any
-// title whose measured reads are submitted. MED that no title mixes both patterns within one
-// container -- that is the one case this heuristic would get wrong, it cannot be checked without
-// running such a title, and PROSPER_APR_MEASURE_READ=always is the one-variable A/B if a title ever
-// regresses. `never` exists for the opposite arm (does a title depend on the read at all?).
+// title whose measured reads are submitted -- which now includes DOLL and Stray by replay, 0
+// unmatched measures across seven logs and two titles. MED that no title mixes both patterns within
+// one container: two titles is not the corpus, and a log replay cannot see what a title does with
+// bytes that arrive at submit time rather than measure time. The mixed pattern is DETECTED rather
+// than merely feared -- see apr_measure_note_unmatched below -- and PROSPER_APR_MEASURE_READ=always
+// is the one-variable A/B if a title ever regresses. `never` is the opposite arm (does any title
+// still depend on this read at all?).
 namespace {
-struct AprMeasuredRead { uint32_t id; uint64_t dst, offset, size; };
+struct AprMeasuredRead {
+    uint32_t id; uint64_t dst, offset, size;
+    bool live = false;        // this slot has been written since the last reset
+    bool suppressed = false;  // recorded AFTER the container paired, i.e. its read was skipped
+    bool matched = false;     // a submit for exactly this tuple has since arrived
+};
 PROSPER_HEAP_MUTEX(g_apr_measure_mx);
 constexpr size_t kAprMeasureRingMax = 64;   // only the IMMEDIATE pairing matters; Stray's is adjacent
 std::array<AprMeasuredRead, kAprMeasureRingMax> g_apr_recent_measures{};
 size_t g_apr_recent_measures_used = 0;      // grows to kAprMeasureRingMax, then the ring wraps
 size_t g_apr_recent_measures_next = 0;
 std::map<uint32_t, bool> g_apr_measure_paired;   // file id -> a submit matched one of its measures
+bool g_apr_mixed_pattern_warned = false;         // the detector below fires at most once
 
 // auto (default) = the latch above; always = never suppress; never = never do the compatibility
 // read. Read once: an env lookup per streamed asset is itself a cost on this path.
@@ -2387,9 +2418,16 @@ AprMeasureMode apr_measure_mode() {
 
 static void apr_measure_reset_for_test() {
     std::lock_guard lk(g_apr_measure_mx);
+    g_apr_recent_measures.fill(AprMeasuredRead{});
     g_apr_recent_measures_used = 0;
     g_apr_recent_measures_next = 0;
     g_apr_measure_paired.clear();
+    g_apr_mixed_pattern_warned = false;
+}
+// Test seam for the detector below: has the mixed-pattern warning fired?
+bool prosper_apr_mixed_pattern_warned_for_test() {
+    std::lock_guard lk(g_apr_measure_mx);
+    return g_apr_mixed_pattern_warned;
 }
 
 // True when this container's measured reads have been shown to arrive again through the submit
@@ -2405,10 +2443,45 @@ static bool apr_measure_read_suppressed(uint32_t id) {
     return it != g_apr_measure_paired.end() && it->second;
 }
 
-// Remember a read the measure path just served, so a matching submit can identify the pairing.
-static void apr_measure_record(uint32_t id, uint64_t dst, uint64_t offset, uint64_t size) {
+// Remember a measured read, so a matching submit can identify the pairing -- and so the ONE case
+// this heuristic gets wrong can announce itself.
+//
+// THE DETECTOR. The suppression's decision is loud (one unconditional line per container) but its
+// consequence would otherwise be silent: a title that mixes both patterns within one container gets
+// an unwritten destination and no message, because the suppressed-measure log line sits behind
+// filelog(). PROSPER_APR_MEASURE_READ=always is a recovery, not a detector -- it only helps someone
+// who already suspects this.
+//
+// So recording CONTINUES after a container pairs (the read is skipped; only the bookkeeping stays),
+// every entry carries whether a submit later claimed it, and an entry that leaves the ring having
+// been suppressed AND never matched is warned about, once, loudly. Nothing else produces that
+// signature: a pre-pairing entry ages out unmatched all the time and is not suppressed, and a
+// suppressed entry that a submit claimed is the designed case. Keeping the scan alive after pairing
+// costs a 64-entry compare per submit -- ~900k integer comparisons across a 3-minute Stray boot,
+// which is not measurable next to the 583 MB of file I/O this removes.
+//
+// Known blind spot, stated rather than left to be discovered: up to kAprMeasureRingMax entries are
+// still in the ring when the process exits and are never adjudicated, so a title that mixes the two
+// patterns FEWER than 64 measures before quitting would not warn.
+static void apr_measure_warn_mixed_locked(const AprMeasuredRead& lost) {
+    if (g_apr_mixed_pattern_warned) return;
+    g_apr_mixed_pattern_warned = true;
+    fprintf(stderr,
+            "[apr] WARNING: sceAmprMeasureCommandSizeReadFile's compatibility read was suppressed "
+            "for APR container id=%u, but the read (dst=0x%llx off=0x%llx size=%llu) was never "
+            "submitted -- so those bytes were never delivered. This container mixes measured reads "
+            "it submits with measured reads it consumes directly, which is the one pattern the "
+            "suppression gets wrong (#3245). Re-run with PROSPER_APR_MEASURE_READ=always to restore "
+            "the old behaviour, and please report the title. Further occurrences are silent.\n",
+            lost.id, (unsigned long long)lost.dst, (unsigned long long)lost.offset,
+            (unsigned long long)lost.size);
+}
+static void apr_measure_record(uint32_t id, uint64_t dst, uint64_t offset, uint64_t size,
+                               bool suppressed) {
     std::lock_guard lk(g_apr_measure_mx);
-    g_apr_recent_measures[g_apr_recent_measures_next] = AprMeasuredRead{ id, dst, offset, size };
+    AprMeasuredRead& slot = g_apr_recent_measures[g_apr_recent_measures_next];
+    if (slot.live && slot.suppressed && !slot.matched) apr_measure_warn_mixed_locked(slot);
+    slot = AprMeasuredRead{ id, dst, offset, size, /*live=*/true, suppressed, /*matched=*/false };
     g_apr_recent_measures_next = (g_apr_recent_measures_next + 1) % kAprMeasureRingMax;
     if (g_apr_recent_measures_used < kAprMeasureRingMax) g_apr_recent_measures_used++;
 }
@@ -2419,14 +2492,14 @@ static void apr_measure_record(uint32_t id, uint64_t dst, uint64_t offset, uint6
 static bool apr_measure_note_submit(uint32_t id, uint64_t dst, uint64_t offset, uint64_t size) {
     if (apr_measure_mode() != AprMeasureMode::Auto) return false;
     std::lock_guard lk(g_apr_measure_mx);
-    // Already paired: the ring holds nothing for this container any more (its measures stop
-    // recording once suppressed), so skip the scan. This runs on EVERY submit -- 14,255 of them in
-    // a 3-minute Stray boot -- and after the first pairing it is the only case left.
-    if (auto it = g_apr_measure_paired.find(id); it != g_apr_measure_paired.end() && it->second)
-        return false;
+    // The scan deliberately continues after a container has paired: marking the entry is what lets
+    // the detector in apr_measure_record tell "suppressed and never submitted" from "suppressed and
+    // duly submitted". An early-out here would make the mixed pattern undetectable, which is the
+    // whole failure mode this heuristic has.
     for (size_t i = 0; i < g_apr_recent_measures_used; i++) {
-        const AprMeasuredRead& m = g_apr_recent_measures[i];
-        if (m.id != id || m.dst != dst || m.offset != offset || m.size != size) continue;
+        AprMeasuredRead& m = g_apr_recent_measures[i];
+        if (!m.live || m.id != id || m.dst != dst || m.offset != offset || m.size != size) continue;
+        m.matched = true;
         bool& paired = g_apr_measure_paired[id];
         const bool first = !paired;
         paired = true;
@@ -4157,6 +4230,10 @@ HLE(f_apr_measure_read_file) {
     // and would deliver them BEFORE the guest recorded the command. The sizing answer is unchanged;
     // only the read is dropped. See the block beside apr_measure_read_suppressed.
     if (apr_measure_read_suppressed((uint32_t)a0)) {
+        // Record it anyway. The read is skipped; the bookkeeping is what lets a suppressed measure
+        // that NO submit ever claims be detected and warned about rather than silently losing its
+        // bytes (see apr_measure_record's detector).
+        apr_measure_record((uint32_t)a0, a1, a3, a2, /*suppressed=*/true);
         if (filelog())
             fprintf(stderr, "[apr] measure-read id=%llu %s -> %llu bytes (compatibility read "
                             "suppressed: this container submits its measured reads)\n",
@@ -4212,7 +4289,7 @@ HLE(f_apr_measure_read_file) {
     // Record what was delivered, so a submit of the identical read can identify the pairing and
     // retire this container's compatibility read (#3245). Recorded only on a read that actually
     // landed: a failed one has not made the submit redundant.
-    if (copied) apr_measure_record((uint32_t)a0, a1, a3, a2);
+    if (copied) apr_measure_record((uint32_t)a0, a1, a3, a2, /*suppressed=*/false);
     if (filelog())
         fprintf(stderr, "[apr] measure-read compatibility id=%llu %s -> dst=0x%llx "
                         "off=0x%llx size=%llu %s; measured=%llu\n",
--- a/prosper/src/hle/fs/hle_file.cpp
+++ b/prosper/src/hle/fs/hle_file.cpp
@@ -26,6 +26,7 @@
 #include <optional>      // #1205: sandbox_normalize_subpath escape check
 #include <mutex>
 #include <atomic>        // sceKernelAio* submit-id/state table
+#include <array>       // #3245: the bounded recent-measure ring
 #include <map>
 #include <utility>
 #include <vector>
@@ -2221,10 +2222,16 @@ uint32_t prosper_apr_register(const std::string& path, uint64_t size) {
     g_apr_files.push_back({ path, size });
     return (uint32_t)g_apr_files.size();
 }
+// Defined with the measure-read suppression state further down; the test hook below clears it too,
+// because that state is process-global and a stale latch would carry from one case into the next.
+static void apr_measure_reset_for_test();
 // Test hook: drop all registered containers (the registry is process-global).
 void prosper_apr_reset_for_test() {
-    std::lock_guard lk(g_apr_mx);
-    g_apr_files.clear();
+    {
+        std::lock_guard lk(g_apr_mx);
+        g_apr_files.clear();
+    }
+    apr_measure_reset_for_test();
 }
 // Find resolved host paths whose TOTAL size equals `size`. Returns the match count and sets
 // *out_path to the first match. The read path resolves the real APR file id first; it may use this
@@ -2314,6 +2321,118 @@ extern "C" void prosper_apr_chain_reset(uint64_t cb) {
     if (!cb) return;
     std::lock_guard lk(g_apr_chain_mx);
     g_apr_chains.erase(cb);
+}
+
+// --- sceAmprMeasureCommandSizeReadFile's compatibility read: when it is still needed ------------
+// The firmware contract for that call is pure SIZING -- it answers 20 or 24. prosper also performs
+// the described read and DMAs it into the guest destination, because DOLL's (PPSA17942) older SDK
+// wrapper consumes the destination from the measure call and never records a ReadFile command for
+// it, so the measure is prosper's only hook on those bytes. That is a real, live requirement; see
+// docs/UE4_APR_IOSTORE_BRINGUP.md.
+//
+// The problem (#3245) is that the gate was only "does the file id resolve", so the read fired for
+// EVERY APR title. On a title that also submits the read normally the same bytes are pread and
+// DMA'd into the same guest address twice: measured on Stray (PPSA02101), a 3-minute boot did
+// 13,925 compatibility reads delivering 582,619,946 bytes, and 97% of them were re-delivered
+// moments later by the submit path from an identical (dst, offset, size).
+//
+// It is not only waste. A sizing call writing to the guest's destination puts bytes there before
+// the guest has recorded the read command, let alone submitted it. On Stray that window is
+// microseconds on one thread, but a title that measures a batch and prepares its destinations
+// afterwards would have prosper's bytes land first and then be overwritten -- silent corruption,
+// not a slowdown. So the narrow answer is the correct one on both counts.
+//
+// WHAT DISTINGUISHES THE TWO TITLES is not visible at the measure call: DOLL and Stray pass the
+// same argument shape. It IS visible immediately afterwards -- if a submit arrives for exactly the
+// (file id, dst, offset, size) a measure just described, that title records and submits its
+// measured reads, so the compatibility read was redundant. That is self-calibrating: it costs one
+// duplicate per APR file id and then turns itself off, needs no per-title or per-SDK knowledge, and
+// cannot fire for DOLL, whose measured reads are never submitted through the ReadFile NID -- if
+// they were, prosper would already have served them eagerly there and the compatibility read would
+// never have been needed.
+//
+// Keyed per APR FILE ID rather than globally so a title with a mixed pattern loses the
+// compatibility read only for the container that demonstrated the pairing.
+// CONFIDENCE: HIGH on the duplication (measured, #3245) and on the suppression being safe for any
+// title whose measured reads are submitted. MED that no title mixes both patterns within one
+// container -- that is the one case this heuristic would get wrong, it cannot be checked without
+// running such a title, and PROSPER_APR_MEASURE_READ=always is the one-variable A/B if a title ever
+// regresses. `never` exists for the opposite arm (does a title depend on the read at all?).
+namespace {
+struct AprMeasuredRead { uint32_t id; uint64_t dst, offset, size; };
+PROSPER_HEAP_MUTEX(g_apr_measure_mx);
+constexpr size_t kAprMeasureRingMax = 64;   // only the IMMEDIATE pairing matters; Stray's is adjacent
+std::array<AprMeasuredRead, kAprMeasureRingMax> g_apr_recent_measures{};
+size_t g_apr_recent_measures_used = 0;      // grows to kAprMeasureRingMax, then the ring wraps
+size_t g_apr_recent_measures_next = 0;
+std::map<uint32_t, bool> g_apr_measure_paired;   // file id -> a submit matched one of its measures
+
+// auto (default) = the latch above; always = never suppress; never = never do the compatibility
+// read. Read once: an env lookup per streamed asset is itself a cost on this path.
+enum class AprMeasureMode { Auto, Always, Never };
+AprMeasureMode apr_measure_mode() {
+    static const AprMeasureMode mode = [] {
+        const char* v = getenv("PROSPER_APR_MEASURE_READ");
+        if (!v || !*v) return AprMeasureMode::Auto;
+        if (!strcmp(v, "always")) return AprMeasureMode::Always;
+        if (!strcmp(v, "never"))  return AprMeasureMode::Never;
+        if (strcmp(v, "auto") != 0)
+            fprintf(stderr, "[apr] PROSPER_APR_MEASURE_READ='%s' is not one of auto|always|never "
+                            "-- using auto\n", v);
+        return AprMeasureMode::Auto;
+    }();
+    return mode;
+}
+}  // namespace
+
+static void apr_measure_reset_for_test() {
+    std::lock_guard lk(g_apr_measure_mx);
+    g_apr_recent_measures_used = 0;
+    g_apr_recent_measures_next = 0;
+    g_apr_measure_paired.clear();
+}
+
+// True when this container's measured reads have been shown to arrive again through the submit
+// path, so the compatibility read would only duplicate one.
+static bool apr_measure_read_suppressed(uint32_t id) {
+    switch (apr_measure_mode()) {
+        case AprMeasureMode::Always: return false;
+        case AprMeasureMode::Never:  return true;
+        case AprMeasureMode::Auto:   break;
+    }
+    std::lock_guard lk(g_apr_measure_mx);
+    auto it = g_apr_measure_paired.find(id);
+    return it != g_apr_measure_paired.end() && it->second;
+}
+
+// Remember a read the measure path just served, so a matching submit can identify the pairing.
+static void apr_measure_record(uint32_t id, uint64_t dst, uint64_t offset, uint64_t size) {
+    std::lock_guard lk(g_apr_measure_mx);
+    g_apr_recent_measures[g_apr_recent_measures_next] = AprMeasuredRead{ id, dst, offset, size };
+    g_apr_recent_measures_next = (g_apr_recent_measures_next + 1) % kAprMeasureRingMax;
+    if (g_apr_recent_measures_used < kAprMeasureRingMax) g_apr_recent_measures_used++;
+}
+
+// A submit for a read the measure path already delivered: from here on this container does not need
+// the compatibility read. Returns true the FIRST time a container pairs, so the caller can say so
+// once instead of once per asset.
+static bool apr_measure_note_submit(uint32_t id, uint64_t dst, uint64_t offset, uint64_t size) {
+    if (apr_measure_mode() != AprMeasureMode::Auto) return false;
+    std::lock_guard lk(g_apr_measure_mx);
+    // Already paired: the ring holds nothing for this container any more (its measures stop
+    // recording once suppressed), so skip the scan. This runs on EVERY submit -- 14,255 of them in
+    // a 3-minute Stray boot -- and after the first pairing it is the only case left.
+    if (auto it = g_apr_measure_paired.find(id); it != g_apr_measure_paired.end() && it->second)
+        return false;
+    for (size_t i = 0; i < g_apr_recent_measures_used; i++) {
+        const AprMeasuredRead& m = g_apr_recent_measures[i];
+        if (m.id != id || m.dst != dst || m.offset != offset || m.size != size) continue;
+        bool& paired = g_apr_measure_paired[id];
+        const bool first = !paired;
+        paired = true;
+        return first;
+    }
+    return false;
 }
 // "Which guest code asked for the file that is not there?" — the question every missing-asset
 // bring-up investigation reaches, and the one a path alone cannot answer. A resolve MISS names the
@@ -3620,6 +3739,15 @@ extern "C" uint64_t f_apr_read_submit(uint64_t a0, uint64_t a1, uint64_t a2,
         if (::stat(host.c_str(), &st) == 0) fsize = (uint64_t)st.st_size;
 #endif
     }
+    // #3245: a submit for exactly the read a measure just delivered proves this container records
+    // and submits its measured reads, so its compatibility read is a duplicate from here on. Placed
+    // before the platform split because the fact is host-independent, and before apr_execute_read
+    // because the pairing is about what the guest ASKED for, not about whether the read succeeded.
+    if (apr_measure_note_submit((uint32_t)id, a4, offset, requested_size))
+        fprintf(stderr, "[apr] %s: the measure-size call's compatibility read is redundant here "
+                        "(this submit re-delivers a read sceAmprMeasureCommandSizeReadFile already "
+                        "served) -- suppressing it for this container. PROSPER_APR_MEASURE_READ="
+                        "always restores it.\n", host.c_str());
     // #2139: this was Linux-only because the discriminator read the guest stack directly, which the
     // Windows import bridge makes impossible — its stub converts SysV->MS-x64 and CALLs the handler,
     // inserting a frame plus shadow space (the #672 class of bug). But that same bridge already
@@ -4024,6 +4152,17 @@ HLE(f_apr_measure_read_file) {
                     (unsigned long long)a0, (unsigned long long)measured);
         return measured;
     }
+    // #3245: this container has already been shown to submit the reads it measures, so the
+    // compatibility read below would deliver bytes the submit path re-delivers moments later --
+    // and would deliver them BEFORE the guest recorded the command. The sizing answer is unchanged;
+    // only the read is dropped. See the block beside apr_measure_read_suppressed.
+    if (apr_measure_read_suppressed((uint32_t)a0)) {
+        if (filelog())
+            fprintf(stderr, "[apr] measure-read id=%llu %s -> %llu bytes (compatibility read "
+                            "suppressed: this container submits its measured reads)\n",
+                    (unsigned long long)a0, host.c_str(), (unsigned long long)measured);
+        return measured;
+    }
 
     uint64_t file_size = 0;
 #ifndef _WIN32
@@ -4070,6 +4209,10 @@ HLE(f_apr_measure_read_file) {
         ::free(staging);
     }
 #endif
+    // Record what was delivered, so a submit of the identical read can identify the pairing and
+    // retire this container's compatibility read (#3245). Recorded only on a read that actually
+    // landed: a failed one has not made the submit redundant.
+    if (copied) apr_measure_record((uint32_t)a0, a1, a3, a2);
     if (filelog())
         fprintf(stderr, "[apr] measure-read compatibility id=%llu %s -> dst=0x%llx "
                         "off=0x%llx size=%llu %s; measured=%llu\n",
--- a/prosper/src/hle/fs/hle_file.cpp
+++ b/prosper/src/hle/fs/hle_file.cpp
@@ -2389,14 +2389,39 @@ struct AprMeasuredRead {
     bool live = false;        // this slot has been written since the last reset
     bool suppressed = false;  // recorded AFTER the container paired, i.e. its read was skipped
     bool matched = false;     // a submit for exactly this tuple has since arrived
+    uint16_t idx_pos = 0;     // where the index below points at this slot (valid while `live`)
 };
 PROSPER_HEAP_MUTEX(g_apr_measure_mx);
 constexpr size_t kAprMeasureRingMax = 64;   // only the IMMEDIATE pairing matters; Stray's is adjacent
 std::array<AprMeasuredRead, kAprMeasureRingMax> g_apr_recent_measures{};
-size_t g_apr_recent_measures_used = 0;      // grows to kAprMeasureRingMax, then the ring wraps
 size_t g_apr_recent_measures_next = 0;
 std::map<uint32_t, bool> g_apr_measure_paired;   // file id -> a submit matched one of its measures
 bool g_apr_mixed_pattern_warned = false;         // the detector below fires at most once
+
+// --- The (id, dst, offset, size) -> ring-slot index ---------------------------------------------
+// apr_measure_note_submit runs on EVERY APR submit -- 14,255 of them in a 3-minute Stray boot -- and
+// it used to LINEAR-SCAN all 64 ring slots under the mutex, because keeping every entry marked (the
+// detector's requirement) removed the early-out that used to skip the scan once a container paired.
+// A 64-slot scan per submit is a plausible shape for a timing-sensitive failure on the submit path,
+// so the mechanism is removed rather than measured: this index restores O(1) expected lookup while
+// every entry still gets marked, which is what the detector needs. #3245.
+//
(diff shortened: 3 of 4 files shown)


mattias800 (commented):
## Independent review — #3263 (head `b7ede68b1`)

**REJECTED** — the change is very likely *correct*, but its central safety claim is **falsified by
DOLL's own boot logs**, and that claim is written into a `## Ruled out`-adjacent doc section and a
shipped code comment. The mechanism by which DOLL is safe is the opposite of the one asserted.

I did not build, run ctest, or launch a title. What follows is from reading the source and from
replaying your own pairing predicate over DOLL's existing `PROSPER_FILELOG` boot logs.

---

### Blocking — "cannot fire for DOLL" is false; it fires on DOLL's 13th submit

Three places say it:

* code comment: *"cannot fire for DOLL, whose measured reads are never submitted through the
  ReadFile NID"*
* `docs/UE4_APR_IOSTORE_BRINGUP.md`: *"DOLL's arm is untouched — it never submits the reads it
  measures, which is why the read exists."*
* PR body, under **Deliberately unaffected**: *"DOLL's arm — a measure with no following submit
  still delivers, always."*

There are four DOLL (`PPSA17942`) run logs from 2026-08-28 with `PROSPER_FILELOG=1` and 300
screenshots each — a route well past first present. I implemented your predicate exactly as the code
does it (64-entry ring, record only on `copied`, stop recording once paired, match on
`(id, dst, offset, requested)`) and replayed it over them.

The very first compatibility read in the log is followed by an identical submit **on the next APR
request**. Hand-checked, not just script output:

```
line 648: [apr] measure-read compatibility id=6 …/pakchunk0-ps5.pak
                -> dst=0x3002860000 off=0x7adec90a size=1080890 OK; measured=20
line 649: [apr] read-submit req=… a3=0x6 desc=0x3002860000 descsz=0x107e3a
line 704: [apr] read-submit id=6 …/pakchunk0-ps5.pak
                -> dst=0x3002860000(guest) off=0x7adec90a size=1080890 got=1080890 OK
                   method=id requested=1080890
```

Identical `(id=6, dst=0x3002860000, off=0x7adec90a, size=1080890)`. `descsz=0x107e3a` = 1,080,890.
That is the same read the doc records at line 619 as DOLL's direct async read
(`vWU-odnS+fU (fileId=8, dst, size=0x107e3a, off=0x7adec90a, …)`).

Under this PR container 6 pairs at **submit #13**, and **1,253 of the run's 1,254** compatibility
reads are suppressed from then on.

### The same evidence is strong reassurance that DOLL does not regress

I then asked the question that actually decides safety: does *any* measured read lack an identical
submit anywhere in the log — the mixed pattern your `MED` covers?

| DOLL run | measures | with an identical submit | **without** |
| --- | --- | --- | --- |
| `run1` (94 MB) | 1,254 | 1,254 | **0** |
| `run3` (128 MB) | 1,725 | 1,725 | **0** |
| `run4` (162 MB) | 1,732 | 1,732 | **0** |

All on container id=6. **Zero unmatched across three independent runs.** DOLL is not the
consume-from-the-measure title the comment describes; it is a submits-everything-it-measures title,
i.e. exactly the case your suppression is designed for. So the suppression drops ~1,250-1,730
duplicate deliveries per boot and the guest still receives every byte through the submit path.

Two honest caveats on that: after the first pairing the log is **counterfactual** (with the change
live the compatibility read stops, so later guest behaviour could differ), and it cannot exclude a
*timing* dependency — bytes now arriving at submit time rather than at measure time. It does exclude
the *data* dependency, which was the whole worry.

### What must change before merge

1. Correct all three statements. The right one is roughly: *"On DOLL's own `PROSPER_FILELOG` boots,
   every one of 1,254 / 1,725 / 1,732 measured reads is re-delivered by an identical submit, so DOLL
   pairs at submit #13 and loses its compatibility read — safely, because the submit path delivers
   the same bytes."* A `## Ruled out` row is the most expensive place in this repository to leave a
   false claim.
2. That correction has a consequence you should state rather than bury: **the documented reason the
   compatibility read exists is not visible in DOLL's current traffic.** The 2026-07-26 SOLVED
   section says DOLL's wrapper "consumes the two measure calls as eager compatibility operations
   while prosper does not execute the encoded commands" — but prosper *does* serve the ReadFile
   submit eagerly now, and DOLL *does* submit. Whether the read is still needed at all is exactly
   what `PROSPER_APR_MEASURE_READ=never` tests. Not this PR's job to remove it; it is this PR's job
   not to re-assert a premise its own evidence contradicts.
3. The test's arm named **"DOLL arm"** models "a measure with no submit", which is now known not to
   be DOLL. It is a fine *mechanism* arm — please rename it so it does not carry title-level
   authority it has not earned.

---

### The design itself — I went looking for the failure and it holds up

* **The issue's suggested fix really is backwards, and you are right to have refused it.** On Stray
  the measure comes first; an order-agnostic duplicate filter would suppress the authoritative
  delivery. Using the pairing only as *evidence* is the correct inversion.
* **Key correspondence checked.** measure `(a0=id, a1=dst, a2=size, a3=offset)` →
  `apr_measure_record(a0, a1, a3, a2)`; submit `id=a3, dst=a4, offset=stack/a6, size=a5` →
  `apr_measure_note_submit(id, a4, offset, requested_size)`. Both raw, both unclamped, consistent.
* **Every failure I could construct is conservative.** A submit that resolves by size-fallback
  carries the unresolved `a3` and cannot pair. An out-of-range measure never records. A refused
  submit returns before the note. Ring eviction only delays pairing. Reads arriving via
  `…GatherScatter` never pair. All of these keep the compatibility read.
* **The ring scan is correct across the wrap** — `0..used-1` is exactly the live set both before and
  after `used` saturates at 64.
* **The sizing answer is genuinely untouched** on every path, including the suppressed one.

### The one case it gets wrong — loud decision, silent consequence

You name it and I agree it is the only real risk. My read on its shape: the *decision* is loud (the
pairing `fprintf` is unconditional, once per container, and names the lever), but the *consequence*
is silent — the suppressed-measure log line is behind `filelog()`, so on a default run a title that
mixed both patterns would just get an unwritten destination.

`PROSPER_APR_MEASURE_READ=always` is a sufficient *recovery*, but it is not a *detector*: it only
helps someone who already suspects this. There is a cheap detector, and it is the thing the charter
asks for — an experiment that detects its own invalidity. **Keep recording measures after
suppression latches** (record only; do not do the read), and warn once, loudly, when a suppressed
measure's `(id, dst, offset, size)` is evicted from the ring having never been matched by a submit.
That is precisely the mixed-pattern signature and nothing else produces it. Strong suggestion, not a
blocker.

Minor: `PROSPER_APR_MEASURE_READ=never` has no test arm — `case AprMeasureMode::Never` is the one
branch nothing executes.

---

### Do the tests discriminate? Yes

The pre-poisoning claim is real, not decorative: `Dest::poison()` fills `0x5A` and `untouched()`
requires **all** bytes still `0x5A`, against a fixture of `i*61+17` — no 200-byte run of which is
constant. So "wrote nothing" and "wrote the right thing" are genuinely distinct observations.

The **near-miss arm** is the one that earns its place: if the pairing were keyed on the file id
alone, the different-offset submit would latch, `after` would come back untouched, and that CHECK
would fail. It discriminates. The dual CMake registration asserting the opposite outcome under
`=always` is the right structure — a suppression that never suppressed cannot pass both. Your Arm 2
(unconditional suppression → 7 failures including both DOLL arms) is the correct bracket on the
over-suppression side.

---

### On the uncovered risk (the reason you flagged it)

**The mutation arm does not substitute for a title run, and my replay only partly does.**
"Suppression made unconditional → the DOLL arms fail" proves the always-deliver path still works
against a synthetic fixture. It cannot observe DOLL's traffic pattern, which is the entire risk.

My log replay is a better answer for the *data* question and I would treat it as settling that half:
zero unmatched measures across three runs. But it strengthens rather than removes the case for one
run, because **the premise you used to skip the run has been falsified** — the heuristic does fire on
DOLL, on the 13th submit, so "it cannot affect DOLL" is no longer available as a reason. What remains
untested is whether DOLL depends on those bytes arriving at *measure* time.

The good news for scheduling: the pairing fires within the first dozen APR submits, long before
first present, so a regression would be immediate and unmissable. **One short `tools/screenshot` run
of `PPSA17942` to first present is enough**, and a second at `PROSPER_APR_MEASURE_READ=always` makes
it an A/B. I would ask for that before merge.

### Mechanical

`git diff --check` clean. No absolute host paths in the diff or the PR body. No numbered-table
changes, so no docs-gate exposure.

**The merge-order note understates the conflict: it is two files, not one.** `git merge-tree` of
`b7ede68b1` against #3262's `cc7480488` conflicts in **`prosper/CMakeLists.txt`** as well as
`docs/UE4_APR_IOSTORE_BRINGUP.md` — both additive at the same anchor, both trivially resolved by
keeping both blocks. `src/hle/fs/hle_file.cpp` auto-merges cleanly, and I checked the merged blob:
both PRs' features survive intact with no conflict markers.

---

Taken on trust, not reproduced: your ctest counts (344 / 348), the two mutation-arm exit codes, and
the Stray measurement (13,925 reads / 582,619,946 bytes) — I found no Stray log with APR FILELOG on
to replay it against.

Correct the DOLL claim in all three places, and either run the title or say plainly in the PR that
it is unrun and why; I will re-review the corrected head.

mattias800:
## The title run review asked for — DOLL reaches first present under the change

Two arms, one binary, `PROSPER_APR_MEASURE_READ` as the only variable. `PPSA17942`,
`tools/screenshot`, 240 s, Linux/RADV, this PR's head `b7ede68b1`.

| arm | distinct frames | published | submission failures |
| --- | --- | --- | --- |
| `auto` (this change) | **281** | 4584 | 1 |
| `always` (old behaviour) | **275** | 4311 | 0 |

Both arms reach the same progression: ~1.1 fps, 7% active share, ~280 distinct source frames.
**The change does not stop DOLL rendering** — it produced slightly *more* distinct frames than the
control, which is within run-to-run variation rather than an improvement.

That answers the question this run existed for. It does not answer everything.

## Two things I will not claim from this run

**1. The captured frames differ, and it is sampling, not rendering.** The `always` arm's first
retained frame carries content (mean 92.4, 477 unique colours) and every other retained frame in
both arms is black. It would be easy to read that as "the change blanks DOLL" — it is not. Both
arms produced ~280 distinct frames; the six-frame sampler happened to land on a content frame in one
arm and not the other. The progression counters are the honest comparison here, and they agree.

**2. The run was contaminated, and I am saying so rather than quietly not mentioning it.** Another
lane was running **13 concurrent `gpu_replay` processes** throughout. The single
`[backend] GRAPHICS submission failed: submit=-4` in the `auto` arm is the one result I cannot
attribute: it is equally consistent with the change and with GPU contention, and there is no arm
here that separates them. One occurrence against zero, on a loaded box, is not evidence of a defect
and is not evidence of its absence.

**A quiet re-run is queued** for when the GPU frees, same two arms. I will post it whatever it says.

## What this does to the review's finding

The reviewer replayed the pairing predicate over DOLL's four boot logs and showed the heuristic
**does** fire for this title (submit #13, hand-verified) — falsifying the PR's stated premise — while
also showing DOLL submits every read it measures (0 unmatched across three runs). This run adds the
timing half that a log replay cannot reach: bytes now arriving at submit time rather than measure
time does not prevent DOLL from rendering.

So the change looks safe for DOLL for the reason the review established, not the one the PR
claimed — and the PR text still needs correcting to say so.

mattias800:
### Exact-head verification for `ec69ad72d` (the corrected head)

Same situation as #3262: the local box has podman's database locked by concurrent lanes, so I have
not run the local rebuild at this commit and am not quoting numbers I did not measure. CI is the
exact-head signal.

All checks complete: **9 SUCCESS, 2 SKIPPED, 0 failed.**

Linux job, full suite with `--no-tests=error` — **all three modes ran by name**, which is the point
of the triple registration:

```
180/347 Test #180: apr_measure_duplicate ..........................   Passed
181/347 Test #181: apr_measure_duplicate_forced ...................   Passed
182/347 Test #182: apr_measure_duplicate_never ....................   Passed
100% tests passed, 0 tests failed out of 347
```

347 rather than my local 348 because the CI Linux job configures without a game dump; **+3 over that
job's base is this PR's three registrations** (auto / `=always` / `=never`). Windows MinGW ran 268
and macOS is green, so the new arms are not Linux-only.

The two mutation arms in the body were run locally before the lock, at the shipped code state.

### On the title run

Noted that you are running the `PPSA17942` A/B yourself — thank you, and I agree it is the right
call: my own replay settles the *data* question (0 unmatched measures across four DOLL boots) but
cannot see whether the title depends on those bytes arriving at **measure** time rather than
**submit** time, which is the half that only a run can answer. Your note that the pairing fires
within the first dozen APR submits matches what I measured independently — container id=6 pairs at
**submit #13** in all four logs — so a regression should be visible long before first present.

mattias800:
## Quiet re-run: the submit failure was contention, but the two runs disagree on the main measure

Re-ran both arms with the GPU verified idle (zero `gpu_replay`/`screenshot`/`prosper-app` on two
consecutive checks) — the confound I flagged on the first pass.

| run | `auto` distinct | `always` distinct | submit failures |
| --- | --- | --- | --- |
| contaminated (13 concurrent replays) | 281 | 275 | 1, in `auto` |
| **quiet** | **315** | **553** | **0** |

**Settled:** the `GRAPHICS submission failed` did not recur on a quiet box. It was contention, not
the change. That is the one thing I said I could not attribute, and it is now attributed.

**Not settled, and I am not going to pretend otherwise:** the two runs disagree about the thing that
matters. The quiet run shows the control producing **553** distinct frames against the change's
**315** — a 43% gap. The contaminated run showed **275 vs 281**, i.e. no gap at all, with the change
slightly ahead.

Both cannot be right about the effect size, and with n=1 per arm per condition I cannot separate a
real regression from DOLL's boot variance. It would be easy to publish either run alone and reach
opposite conclusions, which is a good sign that neither should be published alone.

**Three more rounds of each arm are running now**, alternating, on a quiet GPU. I will post the
spread — six runs, not a pair — whatever it shows. If the gap is real it should survive repetition;
if it is variance, the arms will overlap.

**What this does not change:** DOLL renders under both arms in every run so far (315–553 distinct
frames against ~4,400–4,700 published). The reviewer's question — *does it still reach first
present* — is answered yes in four runs out of four. The open question is now narrower and
quantitative: whether the change costs this title distinct frames, and if so how many.

Holding the merge until the six-run spread is in.

mattias800:
**Local verification cleared** (the podman lock released). Exact head `ec69ad72d`, container `ps5ys`, Linux:

```
cmake --build prosper/build-linux -j4            # BUILD_RC=0
ctest --no-tests=error                           # CTEST_RC=0
100% tests passed out of 349
```

349 against the pre-correction head's 348: **+1 is the new `PROSPER_APR_MEASURE_READ=never`
registration**, the one mode nothing previously executed.

All three modes re-run directly against the rebuilt binary, exit codes only:

```
PROSPER_APR_MEASURE_READ=<unset, auto>   EXIT=0
PROSPER_APR_MEASURE_READ=always          EXIT=0
PROSPER_APR_MEASURE_READ=never           EXIT=0
```

and the detector arms, which are the new mechanism:

```
auto:    [ok] detector: stays quiet when every suppressed measure is duly submitted
         [ok] detector: warns when a suppressed measure is never submitted (the mixed pattern)
always:  [ok] detector: cannot false-positive under =always, where nothing is suppressed
```

Both signals agree at this head — local 349/349 exit 0, CI Linux 347/347 exit 0 with all three
registrations passing by name.

mattias800:
## Correction: my two runs were not the same build, so the "disagreement" was mine

Retracting the framing in my previous comment. The numbers stand; the comparison does not.

```
doll-check   22:50-22:58Z   auto=281  always=275     <- built from b7ede68b1
BINARY REBUILT      00:17:51Z                        <- ec69ad72d landed in this worktree
doll-quiet   00:18-00:26Z   auto=315  always=553     <- built from ec69ad72d
```

`ec69ad72d` is *"APR measure-size: correct the falsified DOLL claim, and make the mixed pattern
detect itself"* — and it is not a documentation change. It makes **recording continue after
pairing** and adds an eviction detector. So I compared one build against another and reported the
result as run-to-run variance in a single change.

**Each run is internally valid** — both arms within a run used the same binary — so:

| head | `auto` | `always` | gap |
| --- | --- | --- | --- |
| `b7ede68b1` | 281 | 275 | none |
| `ec69ad72d` | 315 | 553 | 43% |

which reframes the open question entirely. It is no longer "does this PR cost DOLL frames"; it is
"did the **correction** change behaviour", and that is a much more specific thing to ask, on a
commit whose stated purpose was to correct prose and add a detector.

## How I caught it, since the method generalises

I did not notice while running. I noticed when the lane reported new heads and I checked whether its
push had landed in the worktree my experiment was measuring. `stat` on the binary put the rebuild
at 00:17:51Z, between my two runs. **The experiment could not have told me** — nothing in
`tools/screenshot`'s output names the build it came from.

That is the same shape as tonight's other instrument problems: a real measurement, correctly taken,
attributed to a cause it does not establish. Same family as the stale shared checkout in #3265 and
my own retracted claim in #3259.

## What is running now

The six-run alternating A/B is on `ec69ad72d` throughout, and I have **pinned the binary's
sha256** (`23a5978d…`) so the run can detect its own invalidity: if another rebuild lands mid-run I
will discard the result rather than publish it. That guard exists because I just demonstrated the
failure it prevents.

`auto-r1` = 289 distinct so far, against the earlier `auto` = 315 on the same head — consistent.

**Please do not rebuild in `.claude/worktrees/ampr-laneL-base` until I post the spread.** If a
correction is needed there, tell me and I will re-run rather than have you wait.

Merge still held.

mattias800:
## Six-run spread: the frame gap is not real. The submit failures are unresolved, and I retract calling them settled.

Three alternating rounds per arm, GPU verified idle, one binary with its sha256 pinned and
re-verified after the run (`23a5978d…`, unchanged).

```
auto-r1   289    always-r1  294
auto-r2   266    always-r2  290
auto-r3   281 *  always-r3  274
          ----              ----
mean    278.7           mean 286.0      difference 2.6%
```

**The 43% gap is gone.** `always=553` from the single quiet run was an outlier; six runs cluster at
280 ± 15, which is also where the pre-correction head sat (281 / 275). Lane analysis predicted this
before the data arrived, from the code rather than from more runs: in `always` mode
`apr_measure_read_suppressed` returns at `case Always:` before the lock, `apr_measure_note_submit`
returns at `if (mode != Auto)` before the lock, and the new branch requires `slot.suppressed`, which
that mode never sets. **There is no mechanism by which the commit can move the `always` arm**, so a
100% move in it was always going to be variance. That is a better argument than my six runs and it
cost nothing to make.

## Retraction

I wrote earlier: *"Settled: the `GRAPHICS submission failed` did not recur on a quiet box. It was
contention, not the change."* **That was premature and it is now falsified.** `auto-r3` produced one
on an idle GPU.

```
auto     submit failures in 2 of 5 runs
always   submit failures in 0 of 5 runs
Fisher exact, two-tailed: p = 0.44
```

Not significant at n=5, and I am not claiming a defect. But 0.44 is not "settled" either, and I
reached "settled" from a single non-recurrence — the same one-observation attribution I have spent
tonight correcting in other people's claims and in my own #3259.

## What I recommend instead of more runs

Chasing a p=0.44 signal needs roughly 20 more runs per arm, which is two hours of GPU for a question
that may have no answer. **The mechanism is cheaper to remove than to measure.** `auto` now takes two
lock acquisitions per request pair plus a linear scan of up to 64 slots where it previously
early-outed on an O(1) map lookup — a plausible shape for a timing-sensitive submit failure, and the
only new thing in that path.

A fix already exists in design: replace the linear scan in `note_submit` with a small
open-addressed index over the same 64 slots keyed on `(id, dst, offset, size)`, restoring O(1) while
entries still get marked. That removes the mechanism whether or not the signal is real, which is
worth more than establishing which.

Asking for that before merge. If it lands and the submit failures continue at the same rate, they
were never this change — and that is a cleaner discriminator than the statistics.

## Unchanged

DOLL reaches first present in **10 of 10 runs** across both heads and every arm. The reviewer's
original question stays answered.

mattias800:
### Exact-head verification for `ecbbddea4`

Local, container `ps5ys`, Linux — **build exit 0, ctest exit 0, `100% tests passed out of 349`**,
with all three registrations passing by name. All three modes also run directly: exit 0 / 0 / 0.
349 is unchanged from the previous head because the new arms are assertions inside the existing
three cases, not new ctest cases.

CI: **9 SUCCESS, 2 SKIPPED, 0 failed**. Linux job, `100% tests passed, 0 tests failed out of 347`
(that job configures without a game dump, so its base is smaller), with

```
180/347 apr_measure_duplicate ............ Passed
181/347 apr_measure_duplicate_forced ..... Passed
182/347 apr_measure_duplicate_never ...... Passed
```

Windows MinGW and macOS green, so the index and its arms are not Linux-only.

**The mutation arms are in the body.** The one worth reading is the tombstone case: my first attempt
at covering the probe chain — 600 iterations plus coverage counters — reported **6 rebuilds and 0
tombstone steps**, so the classic open-addressing bug passed straight through it. Ordinary traffic
at 64 live keys in a 256-slot table essentially never crosses a deleted entry. The arm that catches
it had to be **constructed by hand**, and the coverage counter now reads `0 -> 0 -> 1` incrementing
at exactly the load-bearing lookup, so the arm can show its own lever moved.

Ready for the six-round A/B whenever you are. Two notes that may help the run:

- **`=always` is a clean control for this commit specifically.** `note_submit` returns at its mode
  check *before* taking the lock, so the index is never consulted in that mode — the test now
  asserts that rather than skipping, precisely so the control is known-valid rather than assumed.
- If the `GRAPHICS submission failed` rate is unchanged after this, that is the discriminator you
  wanted: the 64-slot scan is gone, so the failures were never this change.

mattias800 (commented):
## Re-review — #3263 at `ecbbddea4`

**REJECTED**, on one defect in new code that is a two-line fix. Everything else — including both
things I did not ask for — I traced and it holds up. I would approve immediately on the corrected
head.

---

### Blocking — under `PROSPER_APR_MEASURE_READ=never`, the detector is a guaranteed false positive

The `=always` inversion exists because, in your own words, *"a detector that fired here would be
reporting a data-loss event on a run where every byte was delivered — the false positive that would
teach the next reader to ignore the warning."* `never` is the mirror case and it was missed:

* `apr_measure_read_suppressed` returns `true` for `Never`, so **every** measure takes the suppressed
  branch and records `suppressed=true` (`:4172`).
* `apr_measure_note_submit` returns at `if (apr_measure_mode() != AprMeasureMode::Auto)` — before the
  lock — so **nothing is ever marked `matched`**.
* `apr_measure_record`'s eviction check therefore fires on the **65th** measure, in **every** run, for
  **every** title.

The text it prints then misdiagnoses a flag the user set themselves: *"This container mixes measured
reads it submits with measured reads it consumes directly, which is the one pattern the suppression
gets wrong… Re-run with PROSPER_APR_MEASURE_READ=always… and please report the title."*

This matters more than a cosmetic wrong message, because `never` is *the* lever for the live open
question this PR itself raises — the doc now says "whether the read is needed at all is what
`PROSPER_APR_MEASURE_READ=never` tests". A warning that fires on every title makes that arm's output
uninterpretable, and it is the second reader who pays.

**The test does not cover it.** The `never` arm returns after two measures (`:158-167`), far short of
the 64-entry ring wrap, so the path is never reached.

Fix: gate `apr_measure_warn_mixed_locked` on `apr_measure_mode() == Auto` (the mode is the
explanation, not a mixed pattern), or make the text mode-aware. Then mirror the `=always` inversion:
drive the `never` arm past 64 measures and assert silence.

---

### The probe-collision arm genuinely forces the case — traced slot by slot

You asked me to check whether it forces the collision or only appears to. It forces it, and the
load-bearing detail is one you did not state explicitly, so let me put it in the record:

* Ring slots after the reset: 0 = the pairing measure, 1 = **A**, 2 = **B**. A takes home bucket `H`;
  B probes past A to `H+1`.
* Fillers 1-61 fill slots 3-63; filler 62 evicts slot 0; **filler 63 evicts A**, and
  `apr_index_erase_locked` tombstones `H`. Matches your comment.
* **No rebuild intervenes, and that is what makes the arm work.** The threshold is
  `live + tombs > 128`; after these 66 records the sum is 66. A rebuild here would have cleared the
  tombstone and quietly turned the arm into a no-op that still passed. It does not happen, but it is
  only ~62 records of headroom — worth a comment so nobody later raises `kAprMeasureRingMax` or
  lowers the table and silently disarms this.
* Fillers cannot occupy `H`: it holds A (a positive entry) throughout their inserts, and linear
  probing never displaces an existing entry.
* The lookup for B hits TOMB at `H`, steps, and matches at `H+1` — `after_lookup > after` by exactly
  one. Sampling the counter immediately before and after that single lookup is the right isolation;
  it makes the assertion independent of whatever the fillers did.
* **The flush proves *found*, not merely *probed past*, and is decisive on its first iteration**: the
  ring pointer sits at slot 2 (B) after filler 63, so the first flush record evicts B. Had the chain
  terminated at the tombstone, B would age out `suppressed && !matched` and warn.
* Portable: the colliding pair is *searched* at runtime against the real stack `dst` rather than
  hardcoded, and `apr_measure_hash` is fixed-constant integer arithmetic.

**And the original zero was real — here is the structural reason, which strengthens your account.**
In the stress arm every `record` is immediately followed by the `find` for the same key.
`apr_index_insert_locked` places the key at the first EMPTY-*or-TOMBSTONE* from its home, and `find`
terminates only at EMPTY — so the key is always reached before any tombstone can cut the chain, and
no eviction intervenes between the two. 0 tomb steps is not a sampling artefact; it is *forced* by
the workload. That is exactly the trap CLAUDE.md describes — a same-source control confirming the
machinery fires on cases the source can build — and hand-building the seam outside that generator is
the correct response. Good catch, and the self-disclosure is the most valuable paragraph in the PR.

### The zero-init encoding claim is correct

`std::array<int16_t, 256> g_apr_measure_index{}` at namespace scope is aggregate
value-initialisation of a trivial type, so it is **constant**-initialised — static init, not dynamic
— and the table reads as all-EMPTY before any dynamic initialiser in any TU can run. That is what
makes "no initialised flag" safe rather than merely lucky. `0` (EMPTY) and `-1` (TOMB) are disjoint
from the valid encoding `1..64`. `g_apr_recent_measures` is likewise constant-initialised (an
aggregate whose NSDMIs are all constants), and `live == false` everywhere is what makes
`apr_index_erase_locked`'s early-out correct on the first pass through the ring. The counters are
zero-init `size_t`. All four legs verified.

*Nit:* the encoding assumes `kAprMeasureRingMax + 1` fits `int16_t`. One
`static_assert(kAprMeasureRingMax < INT16_MAX)` next to `kAprIndexTomb` would pin it — the file
currently has exactly one `static_assert` and none here.

### Also checked, and correct

* Every index operation runs under `g_apr_measure_mx`; `apr_index_find_locked` is reached only from
  `note_submit` and the coverage seam, both locked, so the non-atomic counters are safe.
* `apr_index_erase_locked` precedes the ring overwrite and reads the **old** key's `idx_pos`; its
  `index[idx_pos] != slot+1` guard turns a stale pointer into a no-op rather than a corruption.
* Rebuild-instead-of-insert past half full is right, and for the reason given: the rebuild walks the
  live ring, which already contains the just-written entry. Arithmetic checks out — rebuilds land
  ~65 records apart, matching "at least 64 inserts apart".
* The `=always` control claim holds: `note_submit` returns at the mode check *before* taking the
  lock, so the tomb counter cannot move, which is what makes that arm a valid A/B control for the
  lookup.
  *Nit:* "it cannot cost anything in that mode" is loose — the *record* side still inserts and
  rebuilds under `=always`. It cancels in the A/B because both arms pay it, so the reasoning is
  sound; only the word "anything" overstates.
* **The stress arm doubling as the index's integrity oracle is the best structural idea in this PR.**
  A missed lookup changes no return value anything checks, so there was no oracle available — until
  the detector supplied one: an unfindable entry is never marked, ages out suppressed-and-unmatched,
  and reddens the arm. That is a genuine detect-its-own-invalidity construction, and the two coverage
  assertions are what stop it passing vacuously.

### Non-blocking — duplicate keys are not deduplicated

`apr_index_insert_locked` never checks whether the key is already present, so two live ring slots
holding the same `(id, dst, offset, size)` produce two index entries. `find` returns whichever comes
first in probe order and marks only that one; the other ages out `suppressed && !matched` and warns.
It cannot affect delivery or the suppression latch — `g_apr_measure_paired[id]` is set either way —
so this is a detector false positive only, the same class as the blocking item above.

I checked reachability against the only corpus available: **0 duplicate tuples in 4,711 measured
reads across three DOLL boots** (1,254 / 1,725 / 1,732, all distinct, and none within a 64-record
window). Latent, not live. Cheapest correct fix if you want it: after a hit, keep probing to EMPTY
and mark every entry with the same key — duplicates are rare enough that it stays O(1) expected.

### The corrections I asked for

All three falsified statements are rewritten, and the numbers match mine exactly (1,254 / 1,725 /
1,732 measures, 0 unmatched, pairing at submit #13). The buried consequence — that the documented
reason the read exists is no longer visible in DOLL's traffic — is now stated in the doc and
correctly separated from "whether to narrow it". The Stray replay is a genuinely independent check I
could not run.

*Nit:* it says the Stray replay "reproduces this issue's headline figure independently" while
quoting 13,923 / 14,253 against the issue's 13,925 / 14,255. That is a 2-line difference, almost
certainly a parse boundary — say so, since the value of that sentence is precisely that it is an
independent derivation.

### Item 6, restated

The data question is now settled from both sides — four DOLL logs and three Stray logs, 0 unmatched
— and reproduced by two people independently. What remains is only a *timing* dependency, and since
the pairing fires at submit #13 a regression would be immediate rather than subtle. I now regard the
title run as confirmation rather than a gate, and the six-round A/B in flight covers the
submit-failure question that motivated removing the scan.

### Mechanical

`git diff --check` clean. No absolute host paths in the diff. Merges clean against `origin/main`.
Against #3262's `1efd2dbc4`: conflicts in `prosper/CMakeLists.txt` and
`docs/UE4_APR_IOSTORE_BRINGUP.md`, both additive at the same anchor; `src/hle/fs/hle_file.cpp`
auto-merges and I checked the merged blob — 4,495 lines, no markers, both feature sets intact.

Taken on trust, not reproduced: 349/349 local across the three modes, CI 9 SUCCESS / 2 SKIPPED,
Windows and macOS green, and the A/B still running.

Fix the `never`-mode detector and I will approve.

mattias800:
### Ready for the six-round A/B — final head `6a8594e74`

**Note the head is not `c63d03417`.** That was the intermediate rebase commit; the `never` gate sits
on top of it. Please pin `6a8594e74`.

State, so the copy you take is provably this commit:

| | |
| --- | --- |
| branch head | `6a8594e74c2431aa03bd782f5d192bc757bc667f` |
| worktree | clean (0 modified files) |
| build converged | two further `cmake --build` runs do **zero** compile/link work |
| `build-linux/screenshot` sha256 | `824b7ef7ce704b65db1dd27703a16b0684e3857f3c4b9fbc6be64a50840c170b` |
| local ctest | exit 0, `100% tests passed out of 351` |
| CI | 9 SUCCESS / 2 SKIPPED / 0 failed; Linux `349/349` |

The first `cmake --build` after the commit did rebuild 263 targets — leftover work from restoring a
mutation-arm source file — so I re-ran the full ctest **against the converged build** rather than
quoting the earlier number. The 351 above is from the binary that exists now.

**Copy away freely from here.** Your decoupling fix is the right one and I will not coordinate
around it, but I will also not rebuild this worktree without saying so, since a stale copy is
cheaper to avoid than to detect.

### Please report the six rounds as their own series

Your three valid rounds are on `ecbbddea4`:

```
ecbbddea4   auto-r1 306   always-r1 304   auto-r2 328    submit failures: 0 of 3
```

`6a8594e74` adds the `never` gate, which touches neither `auto` nor `always` — the changed function
is only reached when a suppressed entry is evicted, and in `always` nothing is ever suppressed while
in `auto` the gate is a no-op. So I would *expect* the two series to be comparable. But expecting
them to be comparable is not the same as them being one series, and the PR body now says so
explicitly rather than letting the numbers run together.

mattias800 (commented):
## Re-review — #3263 at `6a8594e74`

**APPROVED.** The blocking finding is fixed, and fixed on the right thing. One note on the new
`CHECK`, which is a comment correction rather than a code one.

### The gate

Verified at `hle_file.cpp:2597`: `if (apr_measure_mode() != AprMeasureMode::Auto) return;` sits
**before** the `g_apr_mixed_pattern_warned` check, so the flag is never set in the other two modes —
which is what both non-Auto arms' `!warned` assertions actually depend on. Correct placement, not
just a correct condition.

**Your account of why it was missed is better than my finding was.** I called `never` the mirror
case; you identified that `always` is safe only *coincidentally* — it never sets `suppressed=true`
at all — so the `=always` arm had been reasoning about a symptom. Gating on the premise ("a submit
should have claimed this entry", which is a property of Auto alone) is the correction that makes
both modes safe for one reason instead of two unrelated ones. That is the difference between fixing
a bug and removing the class.

The `never` arm now discriminates: 200 measures against a 64-entry ring, with the first eviction at
overall record 66 — comfortably inside. Your mutation result (`auto` 0, `always` 0, **`never` 1** on
*"the detector cannot cry wolf past the ring wrap"*) is exactly the arm the old two-measure version
could not reach.

### The new rebuild `CHECK` — sound, keep it, but the comment claims the wrong hazard

Asked to look at this specifically, so in detail.

**It cannot false-fail today.** The block opens with `prosper_apr_reset_for_test()`, which zeroes
`g_apr_index_rebuilds` (`:2515`), and the arm's peak occupancy is 64 live + 2 tombs = **66** against
a `> 128` threshold. Both readings are 0 and cannot be otherwise with the current constants. ✓

**But a rebuild would not "silently turn the whole arm into a passing no-op."** If a rebuild cleared
the tombstone at `H`, B would be re-inserted at or near its home bucket and `find` would resolve it
with **zero** tomb steps — so `after_lookup > after` goes **red**. The arm was already immune to the
silent-pass hazard the new comment attributes to it.

What the `CHECK` actually buys is **diagnosis, not detection**: when a future ring/table/threshold
change breaks this arm, "no rebuild intervened" fails *next to* "stepped over a tombstone", naming
the cause instead of leaving the next reader to re-derive that a rebuild ate the tombstone. That is
genuinely worth having and I would keep it — but the comment should say that. A comment claiming a
hazard the code is already protected against is the same overstatement class this review history has
been about, and it is the one that survives longest because nobody re-checks a justification whose
conclusion they agree with.

**Second, smaller:** you sample `rebuilds_mid` and never assert on it. The precisely load-bearing
window is `rebuilds_mid == rebuilds_after` — no rebuild between the tombstone's *creation* and the
lookup. `rebuilds_before == rebuilds_after` is strictly stronger on a monotonic counter, so it
subsumes that; the cost is that it will also flag a rebuild occurring *before* `H` was tombstoned,
which would have harmed nothing. Fine as a deliberately conservative tripwire — assert both, or say
the window is wide on purpose, so the unused sample stops looking like an oversight.

Neither blocks.

### The other three notes

**`static_assert`s** — verified: `65 <= 32767`, `256 & 255 == 0`, `256 >= 128`. The third pins the
right thing: it is the invariant that keeps a rebuild able to succeed, and it is deliberately weaker
than the comment's "4x the ring", i.e. it pins the boundary that matters rather than today's
comfortable margin. That is the correct choice for a `static_assert`.

**The structural-zero argument** is now in the test in the form I derived, and it is right: insert
places at the first EMPTY-or-TOMBSTONE from the home bucket, find stops only at EMPTY, and each pass
looks up the key it just recorded — so find always reaches the entry at or before insert's position
and no tombstone can sit between them. *"Zero is the only value this workload can produce, not a
small number that more iterations would grow"* is the sentence that stops the next person measuring
it harder.

**Duplicates** — re-derived on a larger corpus than mine (5,966 tuples over four logs against my
4,711 over three), same result, and the disposition is honest: deduplicating means deciding which of
two identical pending reads a single submit satisfies, and nothing observed poses that question.
Recording it as a second known blind spot beside the first is the right call — a note that says why
it is *not* fixed is worth more than a speculative fix.

### Two process points, both worth recording

**The build-convergence disclosure.** A `cmake --build` that rebuilt 263 targets after committing
means the 351 you had already measured came from a different binary — and nothing would have
surfaced that. Re-running the full suite against the converged tree and reporting the stable sha is
the right handling, and volunteering it is the only reason anyone knows. That is the same shape as
the instrument traps this project records: the apparatus, not the subject.

**Measuring the A/B from a copy** rather than the live worktree removes the collision mechanism
instead of detecting it, which is the standing preference here — and stating in the PR body that the
`ecbbddea4` rounds and this series are separate is the distinction that keeps both sets of numbers
usable.

### Mechanical

`git diff --check` clean. No absolute host paths in the diff. Merges clean against `origin/main`.
CI verified by name: **9 SUCCESS / 2 SKIPPED** (Linux, Linux App, Linux App Package, Windows App,
Windows MinGW, macOS App, macOS x86_64/Rosetta 2, Docs, Contribution shape; Progress tracker and
Publish Release skipped). Head is `6a8594e74`, not `c63d03417`. Confirmed both PRs' feature sets are
present on this head after the rebase onto merged #3262 — `refusals_undelivered` /
`apr_read_builder_unimplemented` / the `Qs1xtplKo0U` closer row from #3262, and
`apr_index_find_locked` / `apr_measure_warn_mixed_locked` / the tombstone encoding from this one.

Taken on trust, not reproduced: 351/351 local across the three modes, the mutation exit codes, and
the six-round A/B in flight.

**APPROVED** — merge it.

mattias800:
### Comment correction pushed — head **`9d992af14`**

You and the reviewer were right, and checking it turned up a second, real defect underneath.

**1. The rationale was wrong, exactly as described.** I verified it rather than accepting it: forcing
a rebuild into the window (threshold temporarily 128 → 32) fails the tomb-step assertion with
`tomb steps 0 -> 0 -> 0`. B's home bucket *is* the tombstoned bucket, so after a rebuild B lands
there and `find` resolves it immediately — the arm was already immune to the hazard its own comment
claimed it guarded. The comment now says what the `CHECK` actually buys: **diagnosis, not
detection** — on a constants change it fails beside the tomb-step assertion and names the cause.

**2. Chasing that found a real test defect, and it inverts your `rebuilds_mid` suggestion.** The
narrow window `rebuilds_mid == rebuilds_after` is not the precisely load-bearing one — it is
**vacuously true**. `apr_measure_record` creates the tombstone (`apr_index_erase_locked`) and then,
**in the same call**, may rebuild and destroy it. A sample taken after the filler loop is therefore
already past the only rebuild that can hurt, and nothing can happen between it and the lookup
because a lookup never rebuilds.

The probe demonstrates it rather than my arguing it. Before the fix:

```
[ok]   ...no rebuild between the tombstone appearing and the lookup      <-- PASSED, while broken
[FAIL] ...no rebuild anywhere in the arm
[FAIL] ...the lookup for B stepped over a tombstone (tomb steps 0 -> 0 -> 0)
```

So asserting `mid == after` as written would have added an assertion that reads as meaningful and
checks nothing — the same overstatement class, one layer down. The sample now precedes the evicting
record, and the same probe fails it:

```
[FAIL] ...no rebuild from the evicting record through the lookup   <-- now catches it
[FAIL] ...no rebuild anywhere in the arm
[FAIL] ...tomb steps 0 -> 0 -> 0
```

`rebuilds_mid` is gone. The wider `before == after` is kept with its width stated as deliberate: a
rebuild anywhere in this arm means the constants moved and the slot arithmetic needs re-deriving,
whether or not that run survived it.

**Your A/B is unaffected, and provably so.** The change is test-file-only — `build-linux/screenshot`
is bit-identical, sha256 `824b7ef7ce704b65db1dd27703a16b0684e3857f3c4b9fbc6be64a50840c170b`, the
same binary you pinned at `6a8594e74`. No need to restart or re-pin.

**Verification at `9d992af14`:** build exit 0; all three modes exit 0; second consecutive build does
zero work (so the binary matches the commit); **ctest exit 0, `100% tests passed out of 351`**.

mattias800:
## Six-round A/B on `6a8594e74`: I am withdrawing my "clear to merge" and asking for another look

Run against a **copy** of this head's binary (sha `824b7ef7…`, verified identical to the one posted
here), GPU idle, alternating, one series.

```
auto-r1   284  submit-fail 0  false-warn 0
always-r1 279  submit-fail 0  false-warn 0
auto-r2   245  submit-fail 0  false-warn 0
always-r2 590  submit-fail 0  false-warn 0
auto-r3   277  submit-fail 1  false-warn 0
always-r3 524  submit-fail 0  false-warn 0
```

**Zero false detector warnings in any arm** — the `never` gate is doing its job and that question is
closed.

But two things I did not expect.

### 1. The `always` arm is bimodal and `auto` never reaches its high mode

Within this head alone: `always` produced **590 and 524** where `auto`'s maximum is 284. Pooling
every DOLL run tonight across all four heads:

| | n | min | max | mean | runs > 400 |
| `auto` | 10 | 245 | 328 | 287 | **0 of 10** |
| `always` | 9 | 274 | 590 | 376 | **3 of 9** (553, 590, 524) |

Fisher exact on the high mode: **p = 0.087**.

I previously called `always=553` an outlier. **It was not** — it has now occurred three times, on two
different heads, and `auto` has not reached it once in ten runs. Retracting that too.

More distinct frames means more content changing, i.e. the title progressing further. So the shape of
this is: **the unsuppressed arm sometimes gets further, and the suppressed arm never does.** That is
the exact risk this PR is about — it retires a compatibility read, and if that read is sometimes
load-bearing, suppressing it would cost progression precisely this way.

### 2. The O(1) index did not remove the submit failures

`auto-r3` produced one on an idle box with the linear scan gone. Pooled: `auto` **3 of 10**,
`always` **0 of 9**, p = 0.211.

By the discriminator we agreed on — *if the failures continue at the same rate after the index, they
were never this change* — **they were never the scan.** That half is settled and it exonerates the
index. But it leaves the asymmetry itself unexplained rather than dismissed.

## What I am NOT claiming

- **Not established.** p = 0.087 and p = 0.211 are both weak, `n` is small, and DOLL's boot is
  visibly bimodal in one arm, which is itself unexplained.
- **The pooled table mixes four heads** and should be read as a pattern, not a measurement. The
  within-head comparison on `6a8594e74` is the clean one, and it is only three rounds per arm.
- I have **no mechanism**. Nothing in the reviewed diff obviously explains why suppression would cap
  progression, and the earlier static argument that `always` is barely touched by these commits
  still stands.

## What I would want before merging

Two things, either of which could settle it cheaply:

1. **A mechanism check**: is there a path where a suppressed compatibility read leaves a later
   dependent read unsatisfied, such that the title takes a shorter route? The detector added here
   was built for the *mixed pattern*; this would be a different question — not "was a measure lost"
   but "did suppression change what the title did next".
2. **More rounds.** Six per arm on one head would take the high-mode count from 3-of-9 to something
   that either holds or collapses. That is ~50 minutes of idle GPU.

I am holding the merge. Everything else on this PR is approved and verified, the `never` gate is
confirmed working, and this is the only open question — but it is a question about the PR's central
safety claim, which is not one to merge past on p = 0.087 in either direction.

mattias800:
## Mechanism check: can a suppressed compatibility read cost progression?

**Short answer: there is no path by which the DATA is lost. The only divergence suppression creates is *timing*, and it is bounded by two adjacent APR calls.** I cannot reduce it to zero from a log — the reason is stated below rather than hidden — but the data hypothesis requires something the code makes very hard, and the observed *shape* argues against it independently.

### 1. Suppression removes no side effect that the submit path does not replay

The suppressed call is exactly one thing: the measure path's `apr_write_guest_dst(a1, staging, a2)`. Every effect inside that function is performed again by the submit path, through **the same function, on the same `dst`, with the same bytes**:

| effect | measure path | submit path |
| --- | --- | --- |
| `apr_dest_record` (provenance) | yes | yes |
| `guest_write_watch_notify_host_write` / `…_done` | yes | yes |
| `process_vm_writev` into guest memory | yes | yes |
| commit-retry `mmap` of reserved-uncommitted pages | yes | yes |
| `zerowatch_arm` | yes | yes |
| `apr_cached_fd(host)` | yes | yes |

There is no side effect unique to the measure path — not a page commit, not a watch disarm, not an fd. **After the paired submit completes, guest memory is bit-identical between `auto` and `always`.** The arms differ only in *when* the bytes land and in ~583 MB of redundant I/O that `always` performs.

### 2. The submit never fails to deliver

Across three DOLL boots, **20,671 submits: every one wrote to guest memory. Zero staging fallbacks, zero SHORT.** So there is no case where the measure delivered and the submit then failed to — the route by which suppression could actually lose bytes.

### 3. Every suppressed measure is paired, and the window contains nothing else

0 unmatched measures across four boots, and your six rounds produced 0 detector warnings. The measure→submit window is a median of 55 log lines — but **those lines are the paired submit's own argument dump**. Counting only foreign activity:

```
run1  1252 of 1254 windows contain ZERO other logged guest operations
run3  1721 of 1725
run4  1724 of 1732
```

The measure and the submit that fills the same buffer are **adjacent guest calls**.

### 4. The caveat, stated rather than buried

`PROSPER_FILELOG` logs *file* operations. **A guest reading `dst` is a plain memory read and emits no line at all.** So §3 establishes "no other I/O intervenes", not "the guest did not read `dst`". This instrument cannot see the thing the hypothesis is about, and I will not claim otherwise — that is the instrument-trap shape this project keeps recording.

What §3 does establish is how *narrow* the surviving path is: the guest would have to consume `dst` in an interval bounded by the very call that fills it, with no other I/O in between.

### 5. The observed shape argues against data independently

`auto` is unimodal (245–328, n=10). `always` is bimodal: mostly that same range, occasionally 524–590 (3 of 9). **Missing data produces a consistent wall; it does not produce bimodality.** A race does. And `always` is the arm doing ~583 MB of *extra* I/O — a plausible scheduling perturbation in a codebase with several documented timing races.

### 6. The experiment that settles it costs three runs and no code

**`PROSPER_APR_MEASURE_READ=never`.** Measure-time bytes delivered are strictly ordered:

```
never (none ever)  ⊂  auto (the first read of each container)  ⊂  always (every read)
```

- If measure-time delivery is load-bearing, progression must order `never` ≤ `auto` ≤ `always`, and **`never` must be strictly worse than `auto`** — it removes a read `auto` still performs.
- If `never` ≈ `auto` ≈ 280, then measure-time data is not load-bearing at all, and `always`'s high mode needs a non-data explanation.

**Note what is not on that list: no outcome makes `auto` the worst arm.** `auto` delivers a superset of `never`'s measure-time bytes and the identical final bytes to `always`. For "suppression costs progression" to hold, you would need `auto` to be worse than `never`, which is impossible by construction.

### 7. One instrument caveat on the metric itself

*Distinct frames* conflates "progressed further" with "rendered more varied content". A 2x jump to 524–590 is equally consistent with the title entering a **movie or a camera-panning sequence**, where nearly every frame differs, as with it reaching a later point. Worth checking what the high-mode runs' last frames actually show before treating the mode as progression — a 2x frame count from a cutscene would look exactly like this.

---

## Separately: the `rebuilds_mid` catch, recorded because the shape matters more than the fix

This is the third time tonight the same failure appeared, and it is the clearest, so it belongs in the record rather than in a message.

A reviewer suggested asserting the *narrow* window (`rebuilds_mid == rebuilds_after`) as "the precisely load-bearing" one. That reads as **tightening** — narrower, more precise, obviously better. It was **vacuously true**: `apr_measure_record` creates the tombstone and may then rebuild *in the same call*, so a sample taken after the loop already sits past the only rebuild that can hurt, and nothing can happen between it and the lookup because a lookup never rebuilds.

It was caught by a probe that forced a rebuild into the window. Before the sample point moved:

```
[ok]   ...no rebuild between the tombstone appearing and the lookup   <-- PASSED, while broken
[FAIL] ...no rebuild anywhere in the arm
[FAIL] ...tomb steps 0 -> 0 -> 0
```

**A suggestion that sounds more rigorous than what it replaces is the hardest kind to check, because agreeing with it feels like diligence.** The charter records this for *raised confidence* on citations; this is the same asymmetry pointed at test precision. The defence was the same one that works everywhere else here: build the failure and watch whether the assertion notices.

mattias800:
## Retracting the hold: my metric did not measure what I said it measured

I based "the unsuppressed arm sometimes gets further" on **distinct frames**. That was wrong, and it
was wrong in the way this project's charter warns about most — I read a number as progression
without checking what it counts.

**The high-mode runs rendered nothing.** Every retained frame from `always-r2` (590 distinct) is
`max=0`, pure black, all five. Meanwhile `always-r1` at **279** distinct rendered a real frame
(mean 168.7, 3,842 unique colours), and so did `auto-r1` at 306.

Re-scoring every run on rendered content instead of source-change count:

| arm | n | content-bearing frames per run | mean |
| --- | --- | --- | --- |
| `auto` | 8 | 1, 0, 2, 0, 0, 0, 1, 0 | **0.50** |
| `always` | 8 | 1, 0, 0, 0, 0, 0, 0, 1 | **0.25** |

The direction **reverses**, and neither is significant. `distinct` counts guest-side source changes;
it rises when the title churns and does not track whether anything reached the screen. A 2x jump in
it is as consistent with a camera pan or a movie as with progress — which is the loading-screen trap
in different clothes, and I walked into it after quoting that trap at two other people tonight.

## The structural argument that should have stopped me earlier

Measure-time bytes are strictly ordered **`never ⊂ auto ⊂ always`**. `auto` delivers a superset of
`never`'s measure-time bytes and the *identical final bytes* to `always`. So **no outcome makes
`auto` the worst arm** — my hypothesis required a result the construction cannot produce, and I could
have known that before collecting nineteen runs.

Supporting facts from the code-and-log check: suppression removes exactly one call
(`apr_write_guest_dst` on the measure path), and every effect inside it — write-watch notify/done,
the commit-retry `mmap`, `process_vm_writev`, `zerowatch_arm`, `apr_dest_record`, `apr_cached_fd` —
is replayed by the submit path through the same function, same `dst`, same bytes. **20,671 submits
across three DOLL boots, every one to guest memory, zero staging fallbacks, zero SHORT.** After the
paired submit, guest memory is bit-identical between the arms.

The honest residual: a guest *reading* `dst` emits no log line, so no instrument here can show the
guest did not read it in the window between the measure and its paired submit. What is established
is how narrow that window is — the two are adjacent guest calls, and the "55 intervening lines" that
looked like activity turned out to be the paired submit's own argument dump.

## What still stands

- **Submit failures: `auto` 3 of 10, `always` 0 of 9, p = 0.211.** Unexplained, and not the linear
  scan — one recurred after the O(1) index at the same rate, which by the agreed discriminator
  exonerates the index. Weak, and I am not holding a merge on p = 0.211 with no mechanism.
- The `never` gate is confirmed: **zero false detector warnings in six rounds**.

## Running now

Twelve more rounds, plus a `never` arm as the decisive test: if measure-time delivery were
load-bearing, `never` must be **strictly worse than `auto`**, since it removes a read `auto` still
performs. If `never ≈ auto`, measure-time data is not load-bearing at all.

I will post it, but I am no longer holding the merge on the progression hypothesis — it required an
outcome the construction forbids, and the metric that suggested it does not measure progression.

mattias800:
## Twelve more rounds: no difference between the arms. Clear to merge.

Six rounds per arm, one binary (a verified copy of this head's), idle GPU, alternating.

```
auto   398, 285, 714, 704, 301, 384     max 714
always 529, 272, 288, 293, 665, 418     max 665
submit failures  auto 0   always 0
false detector warnings                  0
```

**Every claim I made from the earlier runs is now dead:**

- *"`auto` never reaches the high mode."* It reached **714**, higher than `always`'s maximum. Both
  arms are bimodal; the asymmetry was a small-sample artefact.
- *"`auto` 3 of 10 submit failures against `always` 0 of 9."* Zero in twelve runs here. That p=0.211
  was noise too.
- And the metric itself did not measure progression, which I retracted above — the 590-distinct run
  rendered nothing at all.

Nineteen runs said one thing, twelve more said the opposite, and the structural argument said from
the start that **no outcome could make `auto` the worst arm** because measure-time bytes are ordered
`never ⊂ auto ⊂ always`. The construction was right and every statistic I built on top of it was
chasing variance in a bimodal metric.

The `never` gate is confirmed across all twelve rounds: **zero false detector warnings**.

Merging.

PR #3266, opened 2026-09-02
windows: a variadic HLE handler gets the guest's own frame, not a signature (#3246)

### LINKED ISSUE #3246: Windows: variadic HLE handlers get no floating-point arguments, because MS x64 varargs also want them in the integer registers
## Summary

prosper's Windows import bridge is signature-driven since #2955, but a `CallSignature` cannot describe
a **variadic** handler, and the variadic handlers are registered by `(HleFn)` cast, so they take the
historical fixed integer path. That path is right for their named integer arguments and wrong for
everything a Microsoft x64 variadic call requires beyond them.

`CONFIDENCE: HIGH` on the mechanism (published convention plus the emitter's own bytes).
**Not reproduced on a Windows host**, and not observed corrupting a title — this is a gap named
during #2955 rather than a failure someone saw.

## The mechanism

Microsoft x64 varargs have two rules System V does not:

1. **A floating-point argument passed in the variadic part must ALSO be placed in the corresponding
   integer register** (`xmm1` *and* `rdx` for the second argument, and so on), because the callee's
   `va_arg` may read either file.
2. Arguments 5 and beyond go on the stack by position, with args 0-3 owning home slots.

System V has neither: FP variadic arguments go in `xmm0..7` only, and `al` carries how many were
used. The bridge's fixed path moves integer registers and touches no xmm at all, so a guest
`printf("%f", x)` reaches the handler with the format string placed correctly and the value nowhere
the host CRT will look for it.

## Affected

`src/hle/libc/hle_libc.cpp`, the real host variadics:

| handler | registered as |
| --- | --- |
| `h_printf` | `printf` |
| `h_sprintf` | `sprintf` |
| `h_snprintf` | `snprintf`, `sprintf_s` |
| `h_sscanf` | `sscanf` |

The `v*` forms (`h_vsnprintf`, `h_vsprintf`) take a `va_list` **pointer** and are not affected — they
copy the guest's list wholesale.

Linux and macOS are unaffected: the host ABI is the guest's, the import stub is a bare tail-jump, and
`hle_libc.cpp` already documents that the compiler-generated variadic prologue captures the full
frame intact.

## Why this is not a #2955 follow-up in the ordinary sense

It cannot be fixed by declaring a signature, because there is no signature: the argument list is
whatever the guest's format string says at run time. Two shapes that could work, neither obviously
right:

- **Parse the format string in the stub's callee.** That is what the handler is going to do anyway,
  so a small adapter could re-marshal from a captured System V frame into a host `va_list`. Doable,
  and it puts format parsing on the ABI path.
- **Capture the guest frame and call the `v*` form.** Build a Microsoft `va_list` from the guest's
  register save area plus overflow area — mechanical, and the same machinery the guest's own
  `va_start` would have produced.

Either way the acceptance test is the same and should be written first, in the shape #2955's already
is: a variadic handler called through the emitted bridge with a mixed `%s %d %f %s` argument list,
asserting the formatted output, executed on any x86-64 host via `__attribute__((ms_abi)))` rather
than requiring a Windows box.

## Also unaudited, and cheaper to answer

The fixed path forwards exactly **ten** guest arguments (`a0..a9`). A registered handler declaring
more would silently lose the rest on Windows. Nobody has counted whether one exists; the survey is a
grep over the registrations plus each target's parameter list, and it is the other half of
`docs/PORTING.md`'s "audit the `(HleFn)`-cast handlers" item.

mattias800:
Fixes #3246.

## The defect

#2955 made the Windows import bridge signature-driven, and explicitly deferred the one argument list
a `CallSignature` cannot describe: a **real C variadic**, whose shape the format string decides at run
time. Those handlers kept the historical fixed integer shuffle, which never touches xmm.

The consequence is larger than "the float is missing", for the same reason as #2955: System V counts
integer and SSE arguments with **two independent counters** and Microsoft x64 with **one positional**
one, so a single float desynchronizes them and displaces every argument behind it too.

For `printf("%d %f %s", i, f, s)`:

| arg | guest sets (SysV) | handler reads (MS) | old stub delivered |
| --- | --- | --- | --- |
| 0 `fmt` | `rdi` | `rcx` | `rcx` ✓ |
| 1 `%d` | `rsi` | `rdx` | `rdx` ✓ |
| 2 `%f` | `xmm0` | home slot 2, from `r8` | **`s`'s pointer bits, read as a double** |
| 3 `%s` | `rdx` | home slot 3, from `r9` | **whatever `r9` held** |

Executed on a real Microsoft x64 host (MinGW + wine), the pre-fix path prints `1|0.00|9437184` where
`1|1.50|2` was asked for. With a `%s` behind the float it is worse than wrong output: the displaced
argument is read as a **pointer and dereferenced**, which printed the stub's own machine code as a
string at `-O0` and **faulted outright** at `-O2` (`Unhandled page fault on read access to
FFFFFFFFFFFFFFFF`).

## The census

Every dump in the local corpus (56 titles) imports `printf`, `sprintf`, `snprintf`, `sprintf_s`,
`snprintf_s`, `sscanf`, `vsnprintf` and `vsprintf`, and every one carries thousands of format strings
with a floating-point conversion (median ~3,500, max 33,761). So the class is universal rather than
exotic — what is unknown is which of those call sites a given boot reaches, which needs a Windows host.

## Two things the issue did not have

**The `v*` forms are NOT unaffected**, and the issue says they are. On Windows `va_list` is a bare
`char*`, so

```cpp
va_list ap; if (a3) memcpy(&ap, P(a3), sizeof(va_list));   // sizeof(va_list) == 8 on Windows
```

copies **eight** bytes of a **twenty-four** byte System V structure and then reads its
`{gp_offset, fp_offset}` pair as a pointer. That is not a wholesale copy, it is a type confusion, and
it applied to `vsnprintf`, `vsprintf` and `vsscanf` on every Windows boot.

**`sscanf` IS unaffected**, and the issue lists it as affected. Every variadic argument a scanf-family
call passes is a **pointer**, so the list is all-integer, no floating-point duplication is required,
and System V's integer registers map one-for-one onto Microsoft's first ten argument positions —
exactly what the historical shuffle already delivers. It is deliberately left on that path.

## The fix

Not to satisfy the Microsoft varargs FP-duplication rule but to **sidestep** it. That rule only binds
on code that *forwards* a variadic call in registers; prosper does not have to.

- **`PROSPER_GUEST_ABI`** (`hle/dispatch/dispatch.hpp`) — `__attribute__((sysv_abi))` on Windows, empty
  elsewhere. `printf`/`sprintf`/`snprintf` carry it, so their import stub becomes the same **bare
  tail-jump Linux uses** and the compiler's own System V variadic prologue captures the guest's frame:
  registers, xmm, AL and the overflow area alike. No width cap, no new emitted machine code.
- **`src/host/abi/guest_varargs.hpp/.cpp`** — reads that list by the System V rules, learns each
  argument's class from the format string, and writes the flat array of 8-byte slots that **is** a
  Microsoft `va_list`, handed straight to the host CRT's `v*` function. Pure functions, compiled and
  tested on every platform.
- **`Hle::register_guest_abi`** — a template whose parameter type carries the attribute, so on Windows
  a handler that forgot the tag is a **compile error**, not a stub reading the wrong registers.
- `emit_guest_abi_tailjump` in the bridge, so those stub bytes stay in `host/abi` and stay testable.

A format the model will not guess at (a POSIX `%1$s`, an x87 `%Lf`, an unknown conversion) formats only
the **prefix it can account for** and says so on stderr — less than the guest asked for, never a value
read from the wrong file.

### The constraint that shapes the code

**MinGW cannot emit SEH unwind data for a `sysv_abi` function.** A frame needing an unwind cleanup
fails to assemble with `.seh_handlerdata used outside of .seh_proc block` — and whether one appears is
an **inlining** decision, so the same source assembles at `-O0` and `-O2` and fails at `-O1` when a
throwing callee gets inlined in. Reproduced on MinGW GCC 16.1.1. So each shim captures and immediately
delegates to a `noinline`/`noexcept` host function that owns every C++ object. That is why
`docs/PORTING.md`'s "the attribute route was abandoned" still stands for the 537 STL-using handlers:
three capture-and-delegate shims can be kept cleanup-free; a library of handlers cannot.

## The ">10 arguments" audit riding on the issue

Answered, negative. No registered handler declares more than ten arguments: handlers are written
through per-file `HLE`/`HLE7`/`HLE8`/`HLE9`/`HLE10` macros whose widest member is exactly ten (12 uses
of `HLE10`, no `HLE11`), and the widest hand-written declarations reach nine. `abi::kLegacyForwardedArgs`
now records the prologue's capacity, and a `static_assert` at the `HLE10` definition couples the two —
the surviving risk is a future `HLE11`, not anything in the tree today. Not split into its own issue
because the answer is one paragraph and one assert.

## Linux and macOS: a no-op by construction

`PROSPER_GUEST_ABI` is empty there, the stub was already a bare tail-jump, and every
`#if defined(_WIN32)` arm leaves the existing code character for character. The guest-ABI registry flag
is set on all platforms because it is *true* on all platforms ("no conversion needed"); only the
Windows emitter reads it.

## Verification

**On this Linux box** — `cmake -S prosper -B prosper/build-linux && cmake --build prosper/build-linux -j6`:

| | result |
| --- | --- |
| build | exit **0** |
| `ctest --no-tests=error` | exit **0**, **347/347 passed** |

Baseline measured on the branch point (`30c4270b6`) in a separate worktree: exit **0**, **346/346** —
so the change adds exactly the one new case and moves nothing else. The first pass of this branch was
verified the same way on its original base `e13fd40af` (345/345 against a 344/344 baseline).

**And on a real Windows host, in CI.** `Windows MinGW` builds the whole tree and runs the whole suite
with `--no-tests=error`: **266/266**, including `guest_varargs`, `sysv_ms_bridge`, `printf_varargs` and
`guest_log_capture`. That job is also what caught the second commit's defect — see below — which is the
single best argument for it being a required check on ABI work.

The new `guest_varargs` case is cross-platform on purpose. Both conventions' `va_arg` are reachable
from any x86-64 host through `__builtin_sysv_va_list` and `__builtin_ms_va_list`, so:

- the System V **read** is checked against a real `va_start` frame read by the **compiler's own**
  `va_arg` — an independently produced positive instance, not the machinery checking itself;
- the Microsoft **write** is consumed by the **compiler's own** ms_abi `va_arg`, so the layout claim is
  checked against the toolchain rather than against the test's belief about it;
- a 20-argument call drives arguments past **both** register files and through the interleaved overflow
  area.

**On a real Microsoft x64 host.** `tools/probe_win_varargs.cpp` (not part of the build; its header
carries the command) cross-compiles the **production sources** with MinGW and runs them under wine:

```
x86_64-w64-mingw32-g++ -std=c++20 -O$O -static -I prosper/src -o probe.exe \
    prosper/tools/probe_win_varargs.cpp prosper/src/host/abi/guest_varargs.cpp \
    prosper/src/host/abi/sysv_ms_bridge.cpp
WINEDEBUG=-all wine probe.exe
```

**4/4 at `-O0`, `-O1`, `-O2`, `-O3`, `-Og`, `-Os`** — a mixed 12-argument call, ten doubles including
two past `xmm7`, the truncating fallback, and the negative control above.

**Not verified:** a live guest calling `printf` on a Windows host. Nobody here has one. The command that
would settle it is the ordinary Windows run — build under MSYS2 UCRT64 with
`cmake -S prosper -B prosper/build-windows -G Ninja -DPROSPER_APP=ON`, launch any title, and confirm the
guest's own log output is intact and a run reaches the same point it does on Linux.

## Verification the tests carry themselves

- the **class-blind** plan must NOT deliver the double (if it did, the test could not see the defect);
- the **guest-ABI flag** must move the emitter (if the two agreed, the stub arm would assert nothing);
- the registry sweep fails if **zero** printf-family NIDs are marked, which would mean the registration
  is not reaching the registry;
- an ordinary handler (`memcpy`) and `sscanf` must NOT be marked.

## Mutation arms

Each was built and run, so the failure is the test's rather than the compiler's.

| --- | --- | --- | --- |
| `sysv_va_arg` advances the SSE cursor by 8 instead of 16 | `0` | `8` | 10 failures, all in the 20-argument call (`argument 3 arrived as 0, expected 3.5`) |
| the format plan classifies float conversions as Integer (the pre-fix behaviour) | `0` | `8` | 10 failures across the plan table and both executed calls |
| the emitter ignores `guest_abi` | `0` | `8` | 3 failures — `emitted 83 bytes, expected 12` |
| the registration drops the `guest_abi` flag | `0` | `8` | 6 failures — every printf-family NID unmarked, plus the vacuity guard |

Reverted; rebuild `0`, ctest `0`, 345/345.

## The second commit, and why it is here

The first push of this branch **failed `Windows MinGW`**: `printf_varargs` and `guest_log_capture`
SEGFAULTed. Neither could fail on Linux, and the cause is the change working as intended.

Both tests take a printf handler's address out of the registry and call it through an ordinary
host-ABI variadic function-pointer type. That was correct while the handlers were Microsoft x64. Now
they are compiled in the guest's convention, so the arguments were placed by one convention and read
by the other, and the first `%s` dereferenced whatever landed in `rdi`.

Both pointer types now carry `PROSPER_GUEST_ABI` — empty everywhere but Windows. `test_printf`'s own
header already claimed it called "through its TRUE variadic signature (same SysV ABI the guest uses
via the tail-jump stub)"; the type simply could not express that before, so this makes the file honest
as well as green. The obligation is recorded at `register_guest_abi`, because `Hle::lookup` hands back
a bare address and nothing in the type system stops the next caller repeating it. A sweep for other
direct callers (`nid_hash("printf")` and friends, plus every `Hle::lookup` user) finds none — every
other one takes an ordinary integer handler, for which `HleFn` remains right.

## Review findings, and what changed

Independent review approved this and asked for one fix. All six actionable findings are addressed in
the third commit.

- **F1 (blocking, and a deletion).** The rewritten comment dropped the caveat that Linux's guest-`%fs`
  swap stub re-pushes only **four** of the guest's spilled words. Re-verified against `emit_swap_stub`
  rather than taken on trust — the four pushes are there, the saved `r11` sits immediately behind them,
  and `guest_tls_enabled()` defaults **on** — so it is live on the primary platform. Restored and
  **updated**, because the new architecture makes it Linux-only: the Windows guest-ABI stub is a bare
  tail-jump with no interposed frame, and macOS never emits the swap stub. Now also **#3271**, since a
  code comment being its only record is precisely what nearly lost it.
- **F5.** `plan_format`'s accepted set must be a subset of what the CRT consumes. **Measured** under
  wine rather than assumed: `%zu %jd %td %hhd %a %S %C` and `%'d` each consume one argument, but `%qd`
  prints `%qd` **literally and consumes nothing**. `q` is refused (332 occurrences across 50 dumps);
  `'` stays, with the measurement recorded beside it because it rests on MinGW's ANSI stdio.
- **F2.** A guest-ABI stub cannot run a return hook. Unreachable through the registry today; the
  emitter now refuses it visibly rather than emitting a stub that drops the hook.
- **F4.** The 511-byte retained prefix could be cut mid-conversion. It is now re-planned, walking the
  cut back to a conversion boundary and keeping the slot count and the formatted string in step.
- **F3 (half).** The registration guard is now checked **by the compiler**: a `static_assert` requires
  `PROSPER_GUEST_ABI` to change the function type on Windows and to leave it alone elsewhere. If the
  tag ever stopped changing the type, the guard would compile and close nothing. The other half — the
  guard closes registration, not calls, which is exactly what the CI catch was — is **#3272**.
- **F6/F7.** The probe no longer claims to compile "the production sources": it compiles the production
  *translation* sources and **re-spells** the shim shape, and says so. Its negative control asserts the
  shape of the corruption (`1|0.00|`) rather than mere inequality.

**Two of the new arms were vacuous when first written**, and both are recorded as such in the test so
the next person does not repeat them: the prefix-cap arm **passed against a build with its own fix
removed** (504 bytes never reached the cap; 600 bytes then hit the 64-argument limit first, at 258
bytes), and the return-hook registry arm read the registry before `register_builtin_hle()` populated
it.

## Second review round: `%'d`, and the method behind it

The reviewer found that `'` was kept on a measurement taken against **the wrong C runtime**. Verified
here rather than accepted, then measured on the real thing.

**The confound.** My probe and the shipping build do not resolve `printf` to the same implementation:

| | toolchain | `__USE_MINGW_ANSI_STDIO` | `vsnprintf` resolves to |
| --- | --- | --- | --- |
| the probe, under wine | Fedora cross mingw, msvcrt (`0x600`) | **1** | `__mingw_vsnprintf` — supports `'` |
| CI + shipped | MSYS2 **UCRT64** (`ci.yml:643`, `:685`) | **0** | UCRT's own — does not |

`_mingw.h:442-454` auto-enables ANSI stdio only while `__MSVCRT_VERSION__ < 0xE00`; `:229-236` pins
exactly `0xE00` on the UCRT branch, so the C++11 clause that rescues the msvcrt build cannot fire on
the shipped one. Confirmed in the local header and in `ci.yml`.

**Then measured, not inferred.** I loaded the real `ucrtbase.dll` and drove its exported
`__stdio_common_vsprintf` (UCRT keeps `snprintf` inline in headers), asking each specifier whether a
trailing sentinel survives:

```
  UCRT %'d   ['d|1234              ] consumes NOTHING
  UCRT %qd   [qd|1234              ] consumes NOTHING
  UCRT %zu %jd %td %hhd %lld %a %.2f  all CONSUME an argument
```

So `'` is `%qd` again — the exact defect this PR fixes, reintroduced by the one specifier I kept.
**Dropped.**

**The method is fixed too, which is the larger half.** A new Windows-only arm stops asserting the
subset rule and **asks whichever CRT the build actually links**, requiring `we consume ⟹ the CRT
consumes` — an implication rather than equality, because refusing something the CRT supports costs a
truncated line while accepting something it does not is argument-shifting corruption. Its vacuity
guard is load-bearing: the table must contain at least one specifier *this* CRT rejects, or the
sentinel detector could be reporting "consumes" for everything and the loop would be green while
measuring nothing.

Because that arm only ever executes on the Windows job, I validated its logic standalone under wine
against **both** runtimes first:

| | `'` refused (shipped) | `'` re-accepted (mutation) |
| --- | --- | --- |
| ANSI stdio — my probe's CRT | ARM PASSES | **ARM PASSES** |
| real UCRT — CI's CRT | ARM PASSES | **ARM FAILS** — `VIOLATION %'d ['d|1234]` |

That bottom-left/bottom-right pair *is* the trap, reproduced: the original measurement was green
because it asked the CRT that does not ship.

`tools/probe_win_varargs.cpp`'s header now carries that table, because its "4/4 at six optimisation
levels" is strong evidence about ABI, register files, overflow area and va_list layout — and **no**
evidence about which specifiers may be accepted.

**The asymmetry the reviewer named, recorded because it is the transferable lesson:** `q` was refused
on a quantified basis (332 occurrences across 50 dumps); `'` was kept on an unquantified "output that
demonstrably works" — even though the 971-occurrence count for `'` was already in hand. Same decision,
two standards of evidence, and the weaker one is the one that survived.

## Is this safe to merge with nobody having booted a guest on Windows?

The reviewer's answer is yes, and the reasoning is worth having here rather than in a thread:

- non-Windows is untouched by construction — `PROSPER_GUEST_ABI` is empty, the stub was already a
  tail-jump, and every `#if defined(_WIN32)` arm leaves the existing code character for character;
- the Windows path being replaced **was not a working path** — it could not deliver a floating-point
  variadic argument at all, and the `v*` handlers were reading a 24-byte structure as an 8-byte one;
- the mechanism is executed on a real Microsoft x64 host at six optimization levels, and the whole
  suite runs there in CI (266/266);
- the blast radius is 8 NIDs, all in the formatting path;
- the residual risks fail toward **truncated output plus a stderr line**, not corruption.

**Please grep the first Windows boot after this lands for `[libc] variadic format not fully
modelled`.** "Unreachable in practice" is currently a *static* census over the dumps — the refusable
shapes look like noise inside the guests' own libc format tables rather than call sites — and one boot
converts that into an observation. The same request is on #3246.

## Risks

- **The riskiest part is the part that cannot be run here**: this changes the machine code installed for
  five imports and the argument handling of three more, on a platform this box cannot boot. Mitigated by
  executing the production sources on a genuine Microsoft x64 host at six optimization levels, and by
  every non-Windows path being untouched.
- **The SEH constraint is optimization-dependent**, so it is a build-breaking hazard rather than a
  silent one — but only if someone edits a `PROSPER_GUEST_ABI` body. The rule is stated at the macro, in
  `src/host/abi/AGENTS.md`, and beside each shim.
- **A System V `va_list` dies with the frame it came from.** The production shims read it while the guest
  frame is live; the first draft of the test did not, and its overflow arguments came back as stack
  litter. That is recorded where it can be re-learned cheaply.
- **The truncating fallback loses output** for a format the model refuses. Measured over the corpus, the
  refusable shapes are essentially all noise inside the guests' own libc format tables (a contiguous
  `%Lf %LF %Le %LE %Lg %LG %La %LA` block) rather than call sites — but it is a real behaviour change on
  Windows if one is ever reached, and it is loud.

## Docs

`docs/PORTING.md` items 3 and 5 updated, four new `## Ruled out` rows (the `v*` claim, the `sscanf`
claim, the FP-duplication framing, and "sysv_abi cannot be used at all"); `src/host/abi/AGENTS.md`
extended for the third file and the two constraints.

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -2193,6 +2193,16 @@ if(TARGET prosper_core)
   target_link_libraries(test_sysv_ms_bridge PRIVATE prosper_core)
   add_test(NAME sysv_ms_bridge COMMAND test_sysv_ms_bridge)
 
+  # #3246: the one argument list #2955's CallSignature cannot describe — a real C variadic, whose
+  # shape the format string decides at run time. CROSS-PLATFORM for the same reason as the bridge
+  # above, and then some: reading a guest System V va_list and writing a Microsoft one are pure
+  # functions, and BOTH conventions' va_arg are available on any x86-64 host through
+  # __builtin_sysv_va_list / __builtin_ms_va_list — so the round trip is executed against the
+  # compiler's own readers here, and against the real CRT on Windows.
+  add_executable(test_guest_varargs tests/host/abi/test_guest_varargs.cpp)
+  target_link_libraries(test_guest_varargs PRIVATE prosper_core)
+  add_test(NAME guest_varargs COMMAND test_guest_varargs)
+
   # sceImeUpdate must invoke the guest key handler on the CALLER's guest %fs, not the host %fs the
   # import stub swapped in (#1286: Evergate SIGSEGV at eboot+0xaf8431 on a key press). Drives the real
   # import path on an activated guest TCB; a no-op pass off Linux.
--- a/prosper/docs/PORTING.md
+++ b/prosper/docs/PORTING.md
@@ -605,11 +605,18 @@ edge cases are tracked separately in #697.
 3. ~~**Validate/repair the guest→HLE ABI trampoline for XMM/float args.**~~ Done (#2955). The bridge
    is signature-driven: `src/host/abi/` carries the two placement tables and emits the trampoline
    from the handler's own C++ declaration, which `Hle::register_typed` preserves. A handler with no
-   float still gets the historical fixed integer shuffle, byte for byte. `printf`-family handlers are
-   deliberately NOT converted — they are real host variadics, reached by the same fixed path as
-   before, and a variadic Microsoft x64 call additionally wants each FP argument duplicated into the
-   integer register. That is #3246, open and unmeasured; a `CallSignature` cannot describe it at all,
-   because the argument list is whatever the format string says at run time.
+   float still gets the historical fixed integer shuffle, byte for byte.
+   ~~The `printf` family is deliberately NOT converted.~~ Done too (#3246), by a different route,
+   because a `CallSignature` genuinely cannot describe a list the format string decides at run time.
+   Those handlers are now tagged `PROSPER_GUEST_ABI` — really `__attribute__((sysv_abi))` on Windows
+   — and their import stub is the same bare tail-jump Linux uses, so the compiler's own System V
+   variadic prologue captures the guest's frame. `src/host/abi/guest_varargs.cpp` then reads that
+   list by the System V rules and writes the flat 8-byte-slot array that IS a Microsoft `va_list`,
+   which sidesteps the FP-duplication rule rather than trying to satisfy it in registers.
+   The attribute route being "abandoned" (above) still holds for the 537 STL-using handlers: MinGW
+   cannot emit SEH unwind data for a `sysv_abi` frame, so it only works where the frame can be kept
+   free of unwind cleanups. Three tiny capture-and-delegate shims can be; a library of handlers
+   cannot.
 4. **VEH recovery hardening for genuine stack-overflow faults.** #633 added a tested assembly recovery
    entry with valid Microsoft-x64 shadow space/alignment, but a truly exhausted guest stack still needs
    a guard-page/dedicated-stack story (Linux uses `sigaltstack`).
@@ -618,12 +625,46 @@ edge cases are tracked separately in #697.
    NIDs — the `libSceFont` scale/slant/weight/render group and the whole libm bank plus
    `strtod`/`strtof` in `hle_libc.cpp` — and those now register through `Hle::register_typed`. The
    remaining cast targets pass and return integers and pointers, which both conventions place
-   identically. What is still unaudited is the *other* half of the original question: whether a cast
-   target's argument COUNT exceeds ten, which the fixed path silently truncates. #3246 carries it.
+   identically.
+   ~~What is still unaudited is the *other* half of the original question: whether a cast target's
+   argument COUNT exceeds ten, which the fixed path silently truncates.~~ Audited for #3246, and the
+   answer is negative: no registered handler declares more than ten arguments. Handlers are written
+   through per-file `HLE`/`HLE7`/`HLE8`/`HLE9`/`HLE10` macros whose widest member is exactly ten (12
+   uses of `HLE10`; there is no `HLE11`), and the only hand-written wide declarations reach nine.
+   `abi::kLegacyForwardedArgs` records the prologue's capacity beside the emitter that implements it.
+   The audit is a source census, so it goes stale the day someone writes an `HLE11` — the surviving
+   truncation risk is that edit, not anything in the tree today.
 
 ### Ruled out — the Windows import trampoline's floating-point arguments
 
-One line per falsified hypothesis, per `CLAUDE.md`. The mechanism and the fix are #2955.
+One line per falsified hypothesis, per `CLAUDE.md`. The mechanism and the fix are #2955, with the
+variadic half in #3246.
+
+- **"The `v*` forms take a `va_list` POINTER and are not affected — they copy the guest's list
+  wholesale."** False on Windows, and it was #3246's own opening claim. A Windows `va_list` is a bare
+  `char*`, so `memcpy(&ap, guest_ptr, sizeof(va_list))` copies **eight** bytes of a **twenty-four**
+  byte System V structure and then reads its `{gp_offset, fp_offset}` pair as a pointer. That is not
+  a wholesale copy, it is a type confusion, and it applied to `vsnprintf`, `vsprintf` and `vsscanf`
+  on every Windows boot. Fixed by the same repacking path as the real variadics; the Linux arms are
+  untouched, where the copy really is wholesale and correct.
+- **"`sscanf` is affected."** False, and it is the one row of #3246's affected table that does not
+  hold. Every variadic argument a scanf-family call passes is a POINTER, so the list is all-integer,
+  no floating-point duplication is required, and System V's integer registers map one-for-one onto
+  Microsoft's first ten argument positions — which is exactly what the historical shuffle already
+  delivers. `h_sscanf` is deliberately left on that path rather than converted for symmetry. The cap
+  that remains is the shuffle's ten arguments, recorded rather than fixed.
+- **"A variadic Microsoft x64 call needs each FP argument duplicated into the integer register, so
+  the fix must emit that duplication."** True of the rule and false of the fix. The duplication only
+  matters to code that FORWARDS a variadic call in registers; prosper does not have to. Reading the
+  guest's System V list and writing a Microsoft `va_list` image directly means no register ever
+  carries a variadic floating-point argument on the host side, and the rule never applies.
+- **"`__attribute__((sysv_abi))` cannot be used on a Windows handler at all."** Too strong, though
+  the shape it came from is real. Measured on MinGW GCC 16.1.1 (2026-09-02): a `sysv_abi` variadic
+  function assembles and runs correctly under wine — *unless* its frame needs an unwind cleanup, in
+  which case the assembler rejects `.seh_handlerdata used outside of .seh_proc block`. And that is an
+  **inlining** decision, so the same source compiled at `-O0` and `-O2` and failed at `-O1`, where a
+  throwing callee was inlined into the tagged frame. So the attribute is usable exactly where the
+  frame can be kept cleanup-free, which is why the variadic shims capture and delegate immediately.
 
 - **"Only four handlers are affected, all in `src/hle/util/hle_font.cpp`; nothing else in 700+
   registered NIDs declares a float parameter."** False — that was #2955's own opening census, and it
--- a/prosper/src/hle/audio/hle_audio.cpp
+++ b/prosper/src/hle/audio/hle_audio.cpp
@@ -5,6 +5,7 @@
 // prosper_core stays dependency-free; a concrete output frontend (SDL3, ...) installs itself via
 // audio_set_sink() from outside the core.
 #include "hle/dispatch/dispatch.hpp"
+#include "host/abi/sysv_ms_bridge.hpp"   // #3246: kLegacyForwardedArgs, the fixed prologue's capacity
 #include "host/image/boot_program.hpp"   // #1659: shared guest-module labelling
 #include "hle/dispatch/nid.hpp"
 #include "hle/audio/audio.hpp"
@@ -50,6 +51,13 @@ namespace prosper {
 #define HLE10(name) static PROSPER_SYSV_ABI uint64_t name(uint64_t a0, uint64_t a1, uint64_t a2, \
                                        uint64_t a3, uint64_t a4, uint64_t a5, uint64_t a6, \
                                        uint64_t a7, uint64_t a8, uint64_t a9)
+// HLE10 is the WIDEST handler shape in the tree, and that is a fact the Windows import stub depends
+// on: its fixed integer prologue forwards exactly `kLegacyForwardedArgs` guest arguments and would
+// silently drop anything beyond them (#3246, where the census was taken). Widening this macro — or
+// adding an HLE11 anywhere — needs the prologue widened first, so the coupling is asserted here
+// rather than left to be rediscovered on a platform nobody develops on.
+static_assert(10 <= prosper::abi::kLegacyForwardedArgs,
+              "HLE10 declares more arguments than the Windows import stub forwards");
 #define P(x) ((void*)(uintptr_t)(x))
 
 namespace {
--- a/prosper/src/hle/dispatch/dispatch.cpp
+++ b/prosper/src/hle/dispatch/dispatch.cpp
@@ -20,6 +20,7 @@ namespace {
         bool placeholder = false;
         HleReturnHook return_hook = nullptr;
         abi::CallSignature signature{};   // #2955: only ever set for a float/double-bearing handler
+        bool guest_abi = false;           // #3246: handler compiled in the guest's own convention
     };
     std::unordered_map<std::string, Reg>& registry() {
         static std::unordered_map<std::string, Reg> r; return r;
@@ -105,7 +106,7 @@ void Hle::register_fn(const std::string& nid, HleFn fn, const char* name,
     // A re-registration with the SAME fn (harmless dedup) is not recorded.
     if (it != reg.end() && it->second.fn != fn && !it->second.placeholder)
         shadow_list().push_back({ nid, it->second.name, name ? name : "" });
-    reg[nid] = { fn, name ? name : "", false, return_hook, {} };
+    reg[nid] = { fn, name ? name : "", false, return_hook, {}, false };
 }
 void Hle::register_fn_with_signature(const std::string& nid, HleFn fn, const char* name,
                                      const abi::CallSignature& signature,
@@ -121,6 +122,18 @@ void Hle::register_fn_with_signature(const std::string& nid, HleFn fn, const cha
     // an overwrite is already the #330 shadow class and is recorded above.
     if (signature.needs_conversion()) registry()[nid].signature = signature;
 }
+void Hle::register_guest_abi_fn(const std::string& nid, HleFn fn, const char* name) {
+    register_fn(nid, fn, name);
+    // Set AFTER register_fn, which rewrites the whole entry. As with a declared signature, a later
+    // plain register_fn on the same NID drops the flag along with the handler it described — the safe
+    // direction, since an unknown handler is then reached through the converting bridge rather than
+    // through a tail-jump that would hand it the guest's frame raw.
+    registry()[nid].guest_abi = true;
+}
+bool Hle::guest_abi_nid(const std::string& nid) {
+    auto it = registry().find(nid);
+    return it != registry().end() && it->second.guest_abi;
+}
 abi::CallSignature Hle::signature_of_nid(const std::string& nid) {
     auto it = registry().find(nid);
     return it == registry().end() ? abi::CallSignature{} : it->second.signature;
@@ -133,7 +146,7 @@ void Hle::register_placeholder(const std::string& nid, HleFn fn, const char* nam
     auto it = reg.find(nid);
     if (it != reg.end() && it->second.fn != fn && !it->second.placeholder)
         shadow_list().push_back({ nid, it->second.name, name ? name : "" });
-    reg[nid] = { fn, name ? name : "", true, nullptr, {} };
+    reg[nid] = { fn, name ? name : "", true, nullptr, {}, false };
 }
 const std::vector<ShadowedReg>& Hle::shadowed_registrations() { return shadow_list(); }
 std::vector<RegisteredFn> Hle::registrations() {
@@ -141,7 +154,7 @@ std::vector<RegisteredFn> Hle::registrations() {
     out.reserve(registry().size());
     for (const auto& kv : registry())
         out.push_back({ kv.first, kv.second.name, (const void*)kv.second.fn, kv.second.placeholder,
-                        kv.second.signature });
+                        kv.second.signature, kv.second.guest_abi });
     return out;
 }
 HleFn Hle::lookup(const std::string& nid) {
--- a/prosper/src/hle/dispatch/dispatch.hpp
+++ b/prosper/src/hle/dispatch/dispatch.hpp
@@ -120,6 +120,13 @@ public:
     // parameter type carries PROSPER_GUEST_ABI, so on Windows an untagged handler is a compile
     // error rather than a stub that silently reads the wrong registers — the tag and the
     // registration cannot drift apart. `A...` is the fixed prefix; the `...` is the real ellipsis.
+    //
+    // ANY DIRECT CALLER MUST USE A PROSPER_GUEST_ABI POINTER TYPE. `Hle::lookup` hands back a bare
+    // address, and calling one of these through an ordinary host-ABI function pointer places the
+    // arguments by the wrong convention — on Windows the first `%s` then dereferences whatever
+    // landed in rdi. Not hypothetical: `test_printf` and `test_guest_log_capture` did exactly that
+    // and SEGFAULTed on the Windows MinGW job the moment these handlers were tagged, and passed
+    // everywhere else. `HleFn` is the wrong type for one of these too.
     template <class R, class... A>
     static void register_guest_abi(const std::string& nid, PROSPER_GUEST_ABI R (*fn)(A..., ...),
                                    const char* name) {
--- a/prosper/src/hle/dispatch/dispatch.hpp
+++ b/prosper/src/hle/dispatch/dispatch.hpp
@@ -35,6 +35,33 @@ namespace prosper {
 // marker of the boundary in case a future toolchain makes the attribute viable.
 #define PROSPER_SYSV_ABI
 
+// A handler compiled in the GUEST's convention rather than the host's. On Windows that really is
+// `__attribute__((sysv_abi))`; everywhere else the host ABI already is the guest's and it expands to
+// nothing. The import stub for such a handler is the same bare tail-jump Linux uses, so the guest's
+// frame arrives intact.
+//
+// This is NOT a reversal of the note above. Tagging the ~537 STL-using handlers was tried and
+// abandoned, and the reason is a hard constraint that still applies: **MinGW cannot emit SEH unwind
+// data for a sysv_abi function**, so a frame that needs an unwind cleanup fails to assemble with
+// `.seh_handlerdata used outside of .seh_proc block`. Reproduced 2026-09-02 on MinGW GCC 16.1.1, and
+// it is optimization-dependent — the same source assembles at `-O0`/`-O2` and fails at `-O1`, where
+// a throwing callee gets inlined into the tagged frame.
+//
+// So the tag is for handlers that can be kept trivially cleanup-free, and the rule is mechanical:
+// **a PROSPER_GUEST_ABI function must own no object with a destructor and must call nothing that can
+// be inlined into it carrying one.** Capture what the guest passed, hand it to an ordinary
+// `noinline` host function, return. Everything interesting happens in that callee, in the host ABI,
+// where the compiler is free again.
+//
+// What it buys, and why the escape hatch exists at all: a REAL C VARIADIC cannot be described by a
+// `CallSignature`, because its argument list is whatever the format string says at run time (#3246).
+// The only thing that can capture it is the compiler's own System V variadic prologue.
+#if defined(_WIN32) && (defined(__x86_64__) || defined(_M_X64))
+#define PROSPER_GUEST_ABI __attribute__((sysv_abi))
+#else
+#define PROSPER_GUEST_ABI
+#endif
+
 // Cached diagnostic-flag reads for HOT paths (PROSPER_ENV_ON / PROSPER_ENV_VALUE) moved to
 // diagnostics/env_cache.hpp: the renderer backend and the draw executor need them too, and
 // should not have to include the HLE dispatch registry to ask whether a switch is on. Included
@@ -66,6 +93,7 @@ struct RegisteredFn {
     const void* fn = nullptr; // handler address; equal addresses = Sony entry points that collapse
     bool placeholder = false; // registered as a deliberately-overridable tracing thunk
     abi::CallSignature signature{};  // #2955: declared only where a float/double is involved
+    bool guest_abi = false;   // #3246: PROSPER_GUEST_ABI handler; the Windows stub converts nothing
 };
 
 // Registry of implemented functions, keyed by NID.
@@ -88,6 +116,19 @@ public:
     static void  register_fn_with_signature(const std::string& nid, HleFn fn, const char* name,
                                             const abi::CallSignature& signature,
                                             HleReturnHook return_hook = nullptr);
+    // Register a REAL C VARIADIC (or any other handler compiled in the guest's own convention). The
+    // parameter type carries PROSPER_GUEST_ABI, so on Windows an untagged handler is a compile
+    // error rather than a stub that silently reads the wrong registers — the tag and the
+    // registration cannot drift apart. `A...` is the fixed prefix; the `...` is the real ellipsis.
+    template <class R, class... A>
+    static void register_guest_abi(const std::string& nid, PROSPER_GUEST_ABI R (*fn)(A..., ...),
+                                   const char* name) {
+        register_guest_abi_fn(nid, reinterpret_cast<HleFn>(fn), name);
+    }
+    static void  register_guest_abi_fn(const std::string& nid, HleFn fn, const char* name);
+    // Whether the handler for a NID is compiled in the guest's convention. The Windows import stub
+    // consults this and emits a bare tail-jump; every other platform emits one regardless.
+    static bool  guest_abi_nid(const std::string& nid);
     // The declared signature for a NID, or a default-constructed (undeclared) one. The Windows
     // import stub consults this; every other platform's host ABI already matches the guest's.
     static abi::CallSignature signature_of_nid(const std::string& nid);
--- a/prosper/src/hle/libc/hle_libc.cpp
+++ b/prosper/src/hle/libc/hle_libc.cpp
@@ -3,6 +3,7 @@
 // host C library. Registered by NID so the loader binds imports directly to them.
 #include "hle/dispatch/dispatch.hpp"
 #include "hle/dispatch/nid.hpp"
+#include "host/abi/guest_varargs.hpp"   // #3246: the guest's variadic list, re-expressed for the host
 #include "gpu/timeline/gpu_timeline.hpp"
 #include <cstring>
 #include <cstdlib>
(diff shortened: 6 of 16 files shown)


mattias800 (commented):
## Independent review — #3266 (variadic HLE handlers on Windows)

**APPROVED**, with one thing I'd like restored before you merge (F1 below — it's a comment, not code).

I read the whole diff and traced the mechanism against the SysV AMD64 psABI and against the
surrounding tree. Per this session's constraints I did **not** build, run ctest, or boot anything;
where that mattered I say so.

---

### What I verified first-hand (by reading, not by running)

- **The System V walk is right.** `sysv_va_arg`'s bounds are the correct ones: `gp_offset + 8 <= 48`
  gives exactly six integer registers, `fp_offset + 8 <= 176` exactly eight xmm registers, the SSE
  stride is 16 with only the low 8 bytes meaningful, and both files fall through to one shared
  overflow cursor advancing 8 bytes at a time. I stepped the boundary cases by hand (`gp_offset==40`
  reads r9 then spills; `fp_offset==160` reads xmm7 then spills).
- **The overflow interleave is handled correctly**, which is the part a register-only model cannot
  express. Because the overflow cursor is consulted lazily and per class, a list that exhausts one
  file but not the other still reads the overflow words in declaration order. This is the subtlest
  thing in the file and it is right.
- **`plan_format`'s refusal rewind is correct.** `refuse(spec, before, …)` rewinding `count` to the
  value it had at the start of the conversion is what makes a half-parsed `%*…` safe. I traced
  `%.*2$d` (pushes for `.*`, then refuses on `2`, rewinds), `%d %*`, `%Lg`, `%%`, a trailing lone
  `%`, `%d$`, `%10d` vs `%10$d`, and the scanset forms `%[]]` / `%[^]]` / `%[abc` — all land where
  the test says.
- **The bound is real.** `push` caps at `kMaxFormatArgs == 64`, which is exactly `slots_`'s size, and
  a truncated prefix always carries **fewer or equal** conversions than the slots that were packed —
  the safe direction. There is no path where `format_` describes more arguments than `plan_.count`.
- **The direct-caller sweep checks out.** `Hle::lookup` feeds a stub emitter in exactly two places
  (`exec_image_linux.cpp:3060`, `exec_image_win.cpp:1197`), and `exec_image_win.cpp` is the only
  Windows emitter. The only code that calls a printf-family handler directly is `test_printf.cpp`
  and `test_guest_log_capture.cpp`, and both are tagged in this diff. No other registration path
  touches those NIDs, so the flag cannot be cleared out from under them today.
- **The tail-jump loses nothing on Windows.** `exec_image_win.cpp:190` records that Windows has no
  guest `%fs` swap, and the bridge does no fs handling, so replacing the converting stub with a bare
  tail-jump costs only the checkpoint — which `win_variadic_call` re-adds. I checked
  `dispatch_pending_guest_exception` (`hle_kernel.cpp:4732`): it is an ordinary call-and-return, so
  calling it from a `noexcept` frame is fine.
- **The double read of the packed list is safe.** `capture_aware_vprintf` may call `vsnprintf` and
  then `vprintf` on the same `va_list`. On Windows `va_list` is a scalar `char*` passed by value, so
  the callee advances its own copy and the second call restarts at slot 0. On Linux the existing
  `va_copy` is untouched. Correct on both.
- **The `>10 arguments` audit.** Confirmed: `HLE8`/`HLE9`/`HLE10` are the only wide macros in the
  tree, there is no `HLE11`, and no handler declares eleven or more `uint64_t` parameters.
- **CI.** 11 jobs, all green on `310c68d6a`, including `Windows MinGW` and `Windows App`.

### What I took on trust

The MinGW SEH/`sysv_abi` constraint and its optimization dependence; the wine probe's 4/4 at six
`-O` levels; that GCC actually *rejects* an untagged handler at the `register_guest_abi` parameter
(see F3); and anything about a live guest on Windows.

---

## Findings

### F1 — blocking-lite: a live Linux caveat was deleted from the record

`hle_libc.cpp` on `main` carried this above the printf family:

> CONFIDENCE: HIGH for register + XMM args … MED only for args that spill to the STACK (>6 GP or
> >8 FP) under the GUEST_FS swap stub, whose reframing shifts the overflow area

The replacement comment says only "CONFIDENCE: HIGH on Linux/macOS, where this is the long-standing
behaviour and the tag is empty." **The caveat is still true, and it is true on the default Linux
path.** `guest_tls_enabled()` defaults to on (`guest_tls.cpp`: the env var read is the opt-out
`PROSPER_NO_GUEST_FS`), `stub_swap_mode()` returns it, and `emit_swap_stub` re-pushes exactly four
guest stack words (args 7-10) before `call r10`. A variadic handler entered that way sees
`[rsp+40]` = the stub's saved `r11`, not the guest's eleventh argument. So an eleven-integer-argument
`printf` is misread on Linux today, and after this PR nothing in the tree says so.

This is not a regression in behaviour — it is the removal of the record, which `CLAUDE.md` treats as
its own cost. Please put a sentence back (it can be shorter, and it can point at
`emit_swap_stub`'s four-word re-push as the reason). I don't need to re-review that.

### F2 — suggestion: `guest_abi` silently swallows a `return_hook`

`emit_sysv_to_ms_bridge` returns `emit_guest_abi_tailjump` **before** it looks at
`params.return_hook`. Linux's `emit_one_stub` checks `return_hook` first and would emit the hook
stub. So a NID that is both guest-ABI and hooked would behave differently per platform, silently.
Nothing in the tree is (there is no way to set a hook through `register_guest_abi_fn`), so this is
latent — but it is the kind of latency that costs a day later. An `assert(!params.return_hook)` or an
explicit "guest-ABI handlers cannot carry a return hook" comment at the short-circuit would close it.

### F3 — the compile-time guard is asserted, not demonstrated, and it covers only half the class

Two separate points:

1. **Not demonstrated.** The four mutation arms all ran on Linux, where `PROSPER_GUEST_ABI` is empty
   and the guard is a tautology. Nothing in CI ever compiles an *untagged* handler through
   `register_guest_abi` on Windows, so "a handler that forgot the tag is a compile error" is
   currently a design intention. I believe it — the CI SEGFAULT-then-fix on the two call sites is
   good evidence that GCC does treat the attribute as part of the pointer type in that position —
   but the PR body and `dispatch.hpp` both state it as a fact. Either soften the wording or add a
   deliberate negative case (a `#if 0`-style compile check, or just a note that it is untested).
2. **Only half the class.** The guard binds on *registration*. The CI failure was on *calls*, and
   `Hle::lookup` still hands back a bare `HleFn`. The sweep found no other direct callers today
   (I re-ran it and agree), but the next one is not prevented by anything except the comment. A
   `Hle::lookup_guest_abi(nid)` that returns a `PROSPER_GUEST_ABI`-tagged pointer type and asserts
   `guest_abi_nid(nid)` would actually close it, and the two tests would use it.

### F4 — suggestion: the 511-byte prefix cap can slice mid-conversion

```cpp
size_t keep = plan_.modelled_bytes;
if (keep >= kMaxPrefixBytes) keep = kMaxPrefixBytes - 1;
```

Every other truncation in this file lands on a conversion boundary by construction; this one lands
on byte 511 wherever that falls. A format longer than 512 bytes that also carries an unmodellable
conversion can therefore hand the CRT a trailing `%`, `%-10.5l`, or similar. Low severity (the CRT
stops at the NUL, and the extra packed slots are simply unread), but it is the one place the file
gives up its own invariant. Rewinding `keep` back past any trailing partial spec would restore it.

### F5 — the real residual risk: `plan_format` must be a *subset* of the host CRT's parser

This is the one I'd think hardest about, because it reintroduces exactly the failure being fixed.
The model and the CRT read the same string, so they agree by construction **only where they agree
about which conversions consume an argument**. A conversion `plan_format` accepts and the CRT does
not count desynchronizes every argument behind it — silently, and in the model's own output path.

Concrete candidates I found:

- **`'`** is accepted as a flag (`is_printf_flag`). It is SUSv2 grouping, supported by
  `__mingw_vsnprintf` but **not** by UCRT/MSVCRT. The build sets `CMAKE_CXX_STANDARD 20` and does not
  set `CMAKE_CXX_EXTENSIONS OFF`, so this compiles as `-std=gnu++20` — `__STRICT_ANSI__` is *not*
  defined, which is precisely the condition mingw-w64 keys `__USE_MINGW_ANSI_STDIO` off. So which
  parser you get is not obvious from the source.
- **C23 `%b`/`%B`** are not in `is_integer_conversion`, so they refuse and truncate. That is the safe
  direction, but it silently drops output if a guest libc uses them.

I am not asking for a change; the failure mode is output fidelity, not memory safety, and the
truncating fallback is loud. But a line in `guest_varargs.cpp` saying "the accepted set must stay a
subset of what the host CRT also counts, and here is why `'` is the one to watch" would be worth more
than most of the tests.

### F6 — the probe's claim is slightly stronger than what it does

`tools/probe_win_varargs.cpp` does cross-compile `guest_varargs.cpp` and `sysv_ms_bridge.cpp` — I
checked the command in its header and the sources it names. It does **not** compile `hle_libc.cpp`;
it re-spells the shim shape ("the fix, exactly as hle_libc.cpp spells it"). So "cross-compiles the
production sources" is true of the translation, and a faithful copy of the wiring. The wiring is
covered instead by the `Windows MinGW` job building the real `hle_libc.cpp` and running `test_printf`
through the real handler — which is fine, and worth saying that way in the PR body so nobody later
reads the probe as covering more than it does.

### F7 — minor: the negative control asserts only inequality

Case 4 asserts `strcmp(g_legacy_buf, "1|1.50|2") != 0`. It passes for the stated reason — I traced
the legacy shuffle and the float genuinely never reaches the callee, so the second field must be
`0.00` — but `!=` would also pass if the control failed for an unrelated reason. Asserting the
observed prefix (`1|0.00|`) would make it a control rather than a smoke test. The comment explaining
why the control avoids a `%s` (a crashing probe cannot report) is good and should stay.

---

## Is this safe to merge with nobody having booted a guest on Windows?

**Yes, in my judgement.** The reasoning, in order of weight:

1. **Non-Windows is untouched by construction, and I checked that rather than assuming it.**
   `PROSPER_GUEST_ABI` is empty off Windows, every `#if defined(_WIN32)` `#else` arm is the previous
   code, and the Linux stub was already a bare tail-jump. 347/347 with a 346/346 baseline is
   consistent with exactly one new case.
2. **The Windows path being replaced is not a working path.** The `v*` handlers were an eight-byte
   read of a twenty-four-byte structure whose `{gp_offset, fp_offset}` pair was then used as a
   pointer, on every Windows boot; the variadics could not deliver a float at all and displaced
   every argument behind one, up to dereferencing a displaced pointer. There is no "working
   behaviour" here for a mistake to regress — the change can only move it.
3. **The mechanism is executed on a real Microsoft x64 host**, at six optimization levels, against
   the production translation sources, with a negative control. That is materially better evidence
   than most Windows-side changes in this tree get, and the cross-platform test additionally checks
   the MS layout claim against the *compiler's* `__builtin_ms_va_list` rather than against the
   author's belief about it.
4. **The blast radius is eight NIDs**, all in the guest's own formatting/logging path, and the
   residual risks (F4, F5) fail toward truncated output with a stderr line, not toward corruption.

What I would ask for alongside the merge, rather than before it: **the first Windows boot after this
lands should be grepped for `[libc] variadic format not fully modelled`.** That line is the only
signal that a real title reaches the truncating fallback, and right now the corpus argument for it
being unreachable is a static census. Worth a note on #3246 or the tracker so it isn't forgotten.

Verdict: **APPROVED** (F1 requested before merge; F2-F7 are suggestions, none blocking).

mattias800 (commented):
## Re-review — #3266 at `c4b6d48ec`

**REJECTED**, on one finding, and I want to be precise about how narrow it is: **the `'` acceptance
is the single item, the fix is one character or one test line, and everything else in the 185-line
delta is good work that I am not asking you to touch.** I will re-approve on that one change without
another full pass.

The reason it is a rejection rather than a suggestion is that you asked exactly the right question —
*"is `'` really safe on the shipped UCRT rather than only under wine?"* — and the answer I found is
**no, or at least not established, and the two toolchains resolve `printf` to different
implementations.** That is a live instance of the class this PR exists to remove, sitting inside the
rule the PR just wrote down for itself.

---

## Blocking: `%'d` was measured on a CRT that is not the one that ships

### The mechanism

`_mingw.h` decides whether `printf` is mingw-w64's own `__mingw_vsnprintf` or the platform CRT's, and
the deciding clause for C++ is gated on the CRT version (`/usr/x86_64-w64-mingw32/sys-root/mingw/include/_mingw.h:442-454`):

```c
#if (defined (_POSIX) || ... || defined (_ISOC99_SOURCE) \
     || (defined (__STDC_VERSION__) && __STDC_VERSION__ >= 199901L && __MSVCRT_VERSION__ < 0xE00) \
     || (defined (__cplusplus) && __cplusplus >= 201103L && __MSVCRT_VERSION__ < 0xE00) \
     || defined (_XOPEN_SOURCE) ... || defined (_GNU_SOURCE) || defined (_SVID_SOURCE)) \
    && !defined(__USE_MINGW_ANSI_STDIO)
#  define __USE_MINGW_ANSI_STDIO 1
#endif
#if !defined(__USE_MINGW_ANSI_STDIO)
#define __USE_MINGW_ANSI_STDIO 0
#endif
```

and `_mingw.h:229-236` sets `__MSVCRT_VERSION__` to **`0xE00`** on the UCRT branch, `0x600` otherwise.

Now put the two toolchains side by side:

| | toolchain | `__MSVCRT_VERSION__` | `__USE_MINGW_ANSI_STDIO` | `vsnprintf` is |
| **the probe** (`tools/probe_win_varargs.cpp`, run under wine) | Fedora/distrobox `x86_64-w64-mingw32-g++`, msvcrt | `0x600` | `1` (the `__cplusplus >= 201103L` clause fires) | **`__mingw_vsnprintf`** — mingw's own pformat |
| **CI + the shipped build** | MSYS2 **UCRT64** (`ci.yml:643` `msystem: UCRT64`, `:649-651` `mingw-w64-ucrt-x86_64-gcc`; same for `windows-app` at `:683`) | `0xE00` | `0` | **UCRT's own `vsnprintf`** |

I checked that nothing rescues the UCRT side: neither `prosper/CMakeLists.txt` nor `.github/workflows/ci.yml`
defines `_GNU_SOURCE`, `_POSIX*`, `_XOPEN_SOURCE`, `_SVID_SOURCE`, `_ISOC99_SOURCE` or
`__USE_MINGW_ANSI_STDIO`, and `__STDC_VERSION__` is not defined in C++. So no clause fires and the
`#if !defined` fallback sets it to 0.

**`__mingw_vsnprintf` implements the SUSv2 `'` grouping flag. UCRT does not** — MS documents the flag
set as `-`, `+`, `0`, ` `, `#`, and `'` is not in it. An unrecognised flag reaching UCRT's format
parser falls through to the type character, where `'` is not a valid conversion, and the historical
MS behaviour is to emit it literally and **consume nothing**.

Which is `%qd`, again, exactly. Our walker pushes an Integer slot for `%'d`; if UCRT skips it, every
argument behind that point reads one slot early, and a `%s` behind it dereferences a non-pointer.
Same failure, same severity, same file.

### Why nothing catches it

The two new `'` arms both check **our model**, not the CRT:

- `check_plan("%'d %'f", "if")` is a pure plan assertion and runs on Linux too.
- The only arm that touches the shipping CRT is the `#if defined(_WIN32)` round trip, and its format
  is `"%s|%d|%.2f|%s|%.2f|%lld"` — **no `'`**.

So the `Windows MinGW` job's 266/266 is not evidence about `'`. It never formats one.

### What I am *not* saying

I checked the rest of the measured set against UCRT's documented feature list rather than
blanket-doubting it, and the msvcrt/UCRT split does **not** put the others at risk: `z`, `j`, `t`,
`hh`, `%a`, `%S` and `%C` are all UCRT features. And refusing `q` is right on **both** CRTs, so that
half of the measurement stands unqualified. **The finding is scoped to `'` alone**, which is also the
one entry your own comment singles out as resting on MinGW's ANSI stdio. Your instinct was right; the
axis the comment names is wrong. It says *"re-measure if the CRT changes"* — but the CRT does not need
to change, it already differs, today, between the probe and CI.

### Two resolutions, either is fine

1. **Refuse `'`** (drop it from `is_printf_flag`), and record in the subset-rule block that
   `__mingw_vsnprintf` implements it while UCRT does not, so the subset rule forces the strict side.
   One character, and it costs the tail of a line on a conversion whose corpus frequency was never
   counted — note the asymmetry with `q`, which got 332-across-50-dumps before being refused, while
   `'` was kept on an unquantified "output that demonstrably works".
2. **Better: put `%'d` in the Windows CRT round-trip arm**, with a trailing sentinel argument after
   it, and let the `Windows MinGW` job answer it. If it passes, `'` is proven on the CRT that ships
   and is guarded on every run from now on; if it fails, take (1). This turns the riskiest remaining
   acceptance from a claim into a CI-checked fact, which is what the rest of this PR does everywhere
   else.

Separately and non-blocking: **`probe_win_varargs.cpp`'s header should name the CRT it links.** Its
whole value is "the CRT is the real one", and it turns out to be a *different* real CRT from the
shipping one. That is a durable trap for whoever runs it next, and the new "BE PRECISE ABOUT WHAT
THIS SHARES WITH PRODUCTION" paragraph is the natural place for one more sentence.

---

## Everything else in the delta — checked, and good

**F1, restored.** The replacement is better than what it replaces: it cites `emit_swap_stub` and its
`push original arg10/9/8/7`, names the saved `r11` sitting immediately behind, and correctly narrows
the scope to Linux (Windows is a bare tail-jump with no interposed frame; `stub_swap_mode()` is false
on macOS). That matches what I derived independently, and #3271 is the right call — a code comment
being the only record is how it nearly went.

**F4, closed properly.** I traced the re-plan by hand. `plan_format(prefix_)` always returns
`modelled_bytes` at a conversion boundary, `modelled_bytes <= keep <= 511` so `prefix_[modelled_bytes]`
is in bounds, and the retained string then carries exactly `plan_.count` argument-consuming
conversions in both branches (re-plan complete → no truncation; re-plan refuses → `count` rewound to
`before`, which is exactly what `[0, spec)` holds). Capturing `complete_`/`reject_` **before**
overwriting `plan_` is the subtle part and it is right — the prefix's own plan can come back
`complete`, and reading the verdict back from it would have silently stopped warning. Worked the
513-byte case end to end: keep 511 → prefix ends in a dangling `%` → re-plan refuses at 510 → kept
510, which is what the test asserts.

**The prefix-cap arm genuinely bites.** Without the re-plan it keeps 511 bytes ending in `%` and
`kept == 510` fails. Given your disclosure that two earlier versions were vacuous, I checked the
other new arms with the same suspicion: `%d %qd` → `"ii"` complete without the `q` removal, so it
fails; `%qu` likewise; the return-hook refusal emits 12 bytes without the guard, so `> kMaxBridgeBytes`
fails; `truncating fallback names a reason` genuinely guards the `reject_` capture, since after a
clean re-plan `plan_.reject` is null. All four discriminate. The `return_hook_of` loop is correctly
placed after `register_builtin_hle()` and says why. The disclosure comments are the right shape —
they record the failure mode, not just the fix.

**F3's demonstrable half is genuinely closed, and the green Windows job *is* the proof.** You asked
whether `is_same_v` can actually fail. It can, and it is the only thing that could: if GCC dropped
the attribute from the type, `GuestAbiVariadic` and `HostAbiVariadic` would be the same type, the
`!std::is_same_v` assert would fire, and `Windows MinGW` would be red. It is green, so on that
toolchain the attribute **is** part of the function type. That is a real compile-time fact, not
prose. The residual inference — "therefore template deduction rejects an untagged handler" — is one
step beyond what the assert states, but it is the load-bearing precondition and it is now checked.
The `#else` arm is weaker by nature (it can only confirm the macro is empty), which is fine and is
what it claims.

**F2, closed the right way.** Returning `kMaxBridgeBytes + 1` reuses the existing fail-visible
mechanism: `emit_one_stub` stages into `uint8_t staged[kMaxBridgeBytes]`, `emitted <= capacity` is
false, nothing is copied, and `install_stubs` reports it. Nothing is written to the staging buffer on
that path, so no overrun. Silently dropping the hook was the one outcome that had to be impossible,
and now it is.

**F7, fixed.** `strncmp(g_legacy_buf, "1|0.00|", 7) == 0` asserts the shape rather than mere
difference, and the comment correctly explains why only the first two fields are deterministic.

**F6, fixed**, and the new paragraph is more honest than most PRs manage: it says plainly that
`guest_snprintf` re-spells the shim rather than being it, and that a change to the real shim must be
mirrored by hand.

CI is green on `c4b6d48ec` across all 9 non-skipped jobs, `Windows MinGW` and `Windows App` included.

---

Verdict: **REJECTED**, on the `'` finding alone. Take either resolution and I will re-approve on the
new head without re-deriving the rest of this — the delta is otherwise exactly what the previous
review asked for, and better in two places than what was asked.

mattias800 (commented):
## Confirmation — #3266 at `e1f07afb0`

**APPROVED.**

You were right to ask rather than hold me to the pre-commitment: 43 lines of `guest_varargs.cpp` is
not the one character I scoped it to. I looked at the new arm, the vacuity guard and the end-to-end
case, and traced the pieces I could reach from here. One tiny nit at the end, not blocking.

### The vacuity guard bites, and it targets the right direction

This is the half neither of us can watch, so I worked it in both directions rather than reading it.

The implication `!we_consume || crt_consumes` is **trivially satisfiable**, which is exactly the
hazard: a detector stuck at `crt_consumes = true` would pass all eleven rows while measuring nothing.
`crt_rejects > 0` is precisely the complement of that failure — a stuck-true detector yields
`crt_rejects == 0` and the guard fires. So it catches the one mode that the main loop cannot.

The opposite break is covered too, by the loop rather than the guard: a detector stuck at `false`
would fail the implication on every specifier the walker *does* accept (`%zu`, `%d`, `%lld`, …), loudly.
So both stuck states are caught, one by each half. That is the right division.

`%qd` is the correct anchor for it. It is rejected by UCRT (your `[qd|1234]`) and by
`__mingw_vsnprintf` (the earlier round's measurement), so `crt_rejects >= 1` holds whichever
implementation a future build resolves to — which is the property the guard needs, since the whole
point is that the CRT is not fixed. Worth noting the guard's inherent limit, which the comment does
not have to state but you should know: it establishes *the mechanism can distinguish*, not *the
mechanism is right about this particular specifier*. A detector wrong on exactly one accepted row
would still slip through. No guard of this shape can do better, and that is not a reason to want a
different one.

### The sentinel detector itself is sound

I checked it for false readings rather than assuming, since everything above rests on it:

- **False positive** (reports "consumes" when it does not): `strstr(rendered, "|77")` can only match
  the sentinel, because `|` appears solely as the literal separator and no leading argument is 77
  (they are 1234, `"str"`, 1.5). For a non-consuming spec the trailing `%d` eats the *first*
  argument, so the tail reads `|1234` — exactly your `['d|1234]`. No path to a spurious match.
- **False negative**: all renderings are far inside the 160-byte buffer, so nothing is truncated away.
- **The varargs are type-clean on this ABI**: passing `(long long)1234` to `%zu`/`%jd`/`%td` is
  size-exact on Win64, and to `%hhd`/`%d` reads the low 4 bytes of an 8-byte slot and still advances
  one slot, so the sentinel stays where the detector expects it. The MS `va_list` being a flat
  8-byte-slot array is what makes the whole technique well-defined, and it is the same fact this PR
  rests on everywhere else.

**Implication rather than equality is the right call**, and the comment says why. One consequence
worth having on the record for whoever reads it next: this arm will never redden for an *over*-strict
walker, so it cannot tell you that a refusal has become unnecessary. That is deliberate — refusing
what the CRT supports costs a truncated line, accepting what it does not is corruption — and the
asymmetry is the reason nobody should later "tidy" it to `==`.

### The end-to-end `%'d` case

Traced by hand: `plan_format("n=%d t=%'d")` refuses at offset 7 with `count` rewound to 1, the prefix
`"n=%d t="` re-plans clean, one slot packs 42, and the real `vsnprintf` renders `"n=42 t="`. Matches
the assertion, and `g_fallback_complete` is false. This is the previously-broken case pinned against
the shipping CRT rather than against our model, which is what was missing. Note it renders through
`g_slots`/`g_fallback_format` — the outer plan and `MsVarargCall` separately — so it also happens to
check those two agree; they do here, and the string assertion would catch it if they stopped.

I confirmed removing `'` from `is_printf_flag` breaks nothing else: the space flag is untouched, and
`%.2f %10.4e %+g %#a` carries no `'`. And truncating is strictly better than passing the full format
through, since UCRT would otherwise print `'d` literally *and* read every later slot shifted, with a
`%s` behind it dereferencing a non-pointer.

### On the method note

Your framing is sharper than mine and I would adopt it. I stopped at "the measurement was taken on
the wrong CRT"; the durable statement is the one now in the probe header — **"executed on a real
Microsoft x64 host" is strong evidence about ABI, register files, overflow area and va_list layout,
and no evidence whatsoever about specifier acceptance.** The confound was invisible in the output,
and it was attached to the most impressive-looking result in the PR, which is what let it travel. The
table is the right place for it and the pointer to where the question *is* answered is what makes it
actionable rather than just cautionary.

The 971-occurrences disclosure is the part I would keep. It converts this from "a measurement was
missing" — which is forgivable and unteachable — into "two standards of evidence were applied to one
decision, and the weaker one carried the case the author preferred." That is the reusable lesson, and
volunteering it is worth more than the fix.

### Nit (not blocking)

`expect("CRT subset probe ran", probed == sizeof kSpecs / sizeof *kSpecs, …)` is a tautology: `probed`
is incremented once per element of the very array it is compared against, so it cannot fail, and its
message ("the specifier table changed but the count did not") describes something that cannot happen.
Harmless, but in a PR that has twice disclosed a vacuous arm it is an odd thing to add — either drop
it or pin the count to a literal, which would at least make an accidental table edit visible.

CI green on `e1f07afb0`, all 9 non-skipped jobs including `Windows MinGW` and `Windows App`.

Verdict: **APPROVED**. Good to merge.

Issue #3267, opened 2026-09-02
169 PROSPER_* numeric env reads discard the end pointer, so a typo silently selects a different setting


mattias800:
## What

`PROSPER_*` variables that are read as **numbers** are parsed, almost everywhere, with a bare
`strtoull(value, nullptr, 10)` (or `atoi`). That discards the end pointer, so the parse cannot fail —
it just answers something. Measured on current `main` over `prosper/{src,frontends,tools,tests}`,
counting only parses whose input comes from `getenv`/`PROSPER_ENV_VALUE` within six lines:

| | count |
| --- | --- |
| discard the end pointer (or `atoi`-family) | **169** |
| check an end pointer | **71** |

By file, the top of the distribution:

```
  30  frontends/shared/live/live_renderer.cpp
  24  src/host/image/exec_image_linux.cpp
  15  src/gpu/pm4/command_processor.cpp
  12  src/hle/graphics/hle_agc.cpp
  10  tests/fixtures/render_runner.h
   8  src/hle/service/hle_service.cpp
   7  src/gpu/capture/gpu_capture.cpp
   7  frontends/shared/live/live_compute.cpp
```

## Why it matters, and why "a typo becomes 0" understates it

#3253 described the hazard as "a typo turns into 0". Measured against glibc rather than assumed, a
bare `strtoull` reaches **three** different wrong answers:

| input | answer | what it means |
| --- | --- | --- |
| `eight`, `yes`, `"8192"` (a quote survived the shell) | **0** | usually "off"; on some knobs the *most aggressive* setting |
| `8mb`, `8 KB`, `8,192`, `8.5` | **8** | a plausible number 1024x from the one intended, that no reader would question |
| `-1`, `18446744073709551616` | **UINT64_MAX** | saturates to whatever cap follows, i.e. "unbounded" |

The middle row is the nastiest: it is silent, it looks right in a log, and nothing distinguishes it
from a deliberate setting.

## What is already done

#3253 fixed the six write-watch knobs, where a `0` is a *meaningful and maximally aggressive*
setting rather than "off" (defer nothing, no size floor, promote on first sight, unbounded arming) —
so a typo there selected the opposite arm of the experiment being run. That PR added a shared strict
parser, `src/diagnostics/env_numeric.hpp`:

```cpp
prosper::diag::env_u64_or_default(name, text, fallback, unit);          // refuses loudly, keeps the default
[...]

PR #3268, opened 2026-09-02
env: the write-watch knobs refuse a typo instead of quietly picking another setting (#3253)

### LINKED ISSUE #3253: Five write-watch tuning variables turn a typo into the most aggressive setting, silently
## What

Five write-watch tuning variables parse with a bare `strtoull(value, nullptr, 10)`, which answers **0**
for anything non-numeric. On each of them 0 is a *meaningful and maximally aggressive* setting, so a
typo does not disable the experiment — it silently selects a different one, and nothing in the run's
output says so.

| site | variable | default | value a typo produces |
|---|---|---|---|
| `frontends/shared/live/live_compute.cpp:1104` | `PROSPER_COMPUTE_WRITE_WATCH_PROMOTE_HITS` | 3 | **0** — `stable >= 0` is always true, so every source arms on first acquisition |
| `frontends/shared/live/live_compute.cpp:1143` | `PROSPER_COMPUTE_WRITE_WATCH_PROMOTE_MB` | 8 | **0** — `WriteWatchPromotionBudget::try_consume` returns true unconditionally when `byte_limit_` is 0, i.e. unbounded arming per submit |
| `frontends/shared/live/live_renderer.cpp:1712` | `PROSPER_TEXTURE_WRITE_WATCH_PROMOTE_MB` | 8 | **0** — same |
| `frontends/shared/live/live_renderer.cpp:3855` | `PROSPER_TEXTURE_WRITE_WATCH_DEFER_MIN_KB` | 8192 | **0** — `should_promote_write_watch` returns true immediately when `defer_min_bytes == 0` |
| `frontends/shared/live/live_renderer.cpp:3861` | `PROSPER_TEXTURE_WRITE_WATCH_PROMOTE_HITS` | 3 | **0** — same as the compute one |

Also in the family, failing the other way: `cold_storage_result_snapshot_defer_min_bytes()`
(`live_compute.cpp:1141` region) *does* validate its end pointer, but falls back **silently**, so a
typo there leaves you measuring the default while believing you moved it.

## Failure scenario

An agent A/B's `PROSPER_TEXTURE_WRITE_WATCH_DEFER_MIN_KB=8mb` (or `8 KB`, or a stray quote), gets the
"defer nothing" arm, and attributes whatever it observes to the value it thought it set. That is
`GAME_COMPAT_ORCHESTRATION.md`'s instrument-trap shape exactly, and this family already produced one
retracted measurement (#3155).

## Why it is not fixed in the PR that found it

#3251 needed a new knob in the same family (`PROSPER_COMPUTE_WATCH_DEFER_MIN_KB`) and gave **that one**
a strict parse that refuses loudly and keeps the default. Fixing two of the five here would have left
the tree inconsistent across two files, which is worse than uniform.

## Suggested fix

One strict parser beside the policy it guards — `frontends/shared/texture/write_watch_policy.hpp`
already owns these thresholds and is unit-tested by `write_watch_policy` — returning success/failure
rather than a value, with the call sites printing the variable name and the default they are keeping.
`live_compute.cpp`'s `compute_write_watch_defer_min_bytes()` is the shape to copy; the charter's rule
is the one `PROSPER_LAZY_COMMIT_STRICT` follows in `exec_image_linux.cpp`: *a malformed value refuses
LOUDLY rather than firing at an unintended setting.* Unit arms belong in
`frontends/shared/tests/test_write_watch_policy.cpp` (empty, leading text, trailing text, overflow,
whitespace, a bare number).

Found while fixing #3156 / measuring #3155.

mattias800:
Fixes #3253.

## The defect

Six write-watch tuning variables parsed with a bare `strtoull(value, nullptr, 10)`. That discards the
end pointer, so the parse cannot fail — it just answers something.

**And "a typo becomes 0" understates it.** Measured against glibc rather than assumed, the old
spelling reaches **three** different wrong answers:

| input | old answer | what it selected |
| --- | --- | --- |
| `eight`, `yes`, `"8192"` (a quote survived the shell) | **0** | the **most aggressive** setting on every one of these knobs |
| `8mb`, `8 KB`, `8,192`, `8.5` | **8** | a plausible number **1024x** from the one intended |
| `-1`, `18446744073709551616` | **UINT64_MAX** | saturates to the following cap, i.e. unbounded |

On this family `0` is not "off":

| variable | default | what 0 means |
| --- | --- | --- |
| `PROSPER_TEXTURE_WRITE_WATCH_DEFER_MIN_KB` | 8192 | `should_promote_write_watch` returns true immediately — **defer nothing** |
| `PROSPER_TEXTURE_WRITE_WATCH_MIN_KB` | 1024 | no size floor — **every** source is watch-eligible |
| `PROSPER_TEXTURE_WRITE_WATCH_PROMOTE_HITS` | 3 | `stable >= 0` is always true — arm on first acquisition |
| `PROSPER_TEXTURE_WRITE_WATCH_PROMOTE_MB` | 8 | `try_consume` returns true unconditionally — **unbounded** arming per submit |
| `PROSPER_COMPUTE_WRITE_WATCH_PROMOTE_HITS` | 3 | as above |
| `PROSPER_COMPUTE_WRITE_WATCH_PROMOTE_MB` | 8 | as above |

So a mistyped A/B did not lose the experiment — it **silently ran a different one**, and nothing in
the run's output said so. That is `GAME_COMPAT_ORCHESTRATION.md`'s instrument-trap shape exactly, and
this family has already produced one retracted measurement (#3155).

**The middle row is the nastiest**, and the issue does not mention it: `=8mb` gave 8 KiB, a plausible
number no reader would question. It also moves the policy the *other* way — a smaller defer minimum
exempts fewer sources, so a 1 MiB texture the intended 8 MiB minimum arms on first sight has to climb
the stability ladder instead. Which direction a typo moves the policy depends on the typo. Both
directions are asserted in the test.

`PROSPER_TEXTURE_WRITE_WATCH_MIN_KB` is not in the issue's table and has the same sentinel; it is
included. `cold_storage_result_snapshot_defer_min_bytes` is — it validated its end pointer but fell
back in **silence**, failing the other way: a typo left you measuring the default while believing you
had moved it.

## The fix

One strict parser, `src/diagnostics/env_numeric.hpp`:

```cpp
prosper::diag::parse_u64_strict(text, &out);                         // exactly [0-9]+, overflow-checked
prosper::diag::env_u64_or_default(name, text, fallback, unit);       // refuses LOUDLY, keeps the default
prosper::diag::env_u64_or_default_capped(name, text, fallback, cap); // ...and saturates before a later multiply
```

No sign, no `0x`, no units, no surrounding whitespace — `strtoull` accepts a leading space and wraps a
leading `-` to `UINT64_MAX`, and both are far likelier to be a typo than an intention. Unset or empty
is **not** a typo and takes the default in silence. It sits in `diagnostics/` beside `env_cache.hpp`,
which answers the neighbouring question ("was this set, and to what text?"), rather than in the
policy header — the parsing is not a write-watch concept and three of the six call sites are in a
different subsystem.

All six sites use it, **and so does #3251's own site**, which had grown its own strict parse a day
earlier — so the family ends with one pattern rather than two. Its historical default is 1 *byte*,
which cannot be expressed in that knob's own KiB units, so it keeps an explicit unset case and calls
`parse_u64_strict` directly.

## The one non-obvious change

`tools/env/check_cached_env.py` learns that a **reader** naming its variable is not an arming. Its
heuristic treats any callee containing `env` as an environment *write* — deliberately conservative,
and correct until now — and `env_u64_or_default` takes the variable's name as a literal first argument
purely so a refusal can report it. Without this, the gate reads the new test as arming three names
that production caches and fails. Two self-test cases pin the carve-out.

## Verification

Base `a4d49cb41`, `cmake -S prosper -B prosper/build-linux && cmake --build prosper/build-linux -j6`:

| | build | ctest `--no-tests=error` |
| --- | --- | --- |
| baseline (`origin/main`, detached, same worktree and build dir) | **0** | **0**, **347/347** |
| this branch | **0** | **0**, **347/347** |

Also green on every CI platform, `Windows MinGW` and `macOS (x86_64 / Rosetta 2)` included — this
change is platform-independent, but `check_cached_env.py` runs everywhere and is the part most likely
to surprise a different tree.

No new test target — the arms live in `write_watch_policy`, beside the policy whose sentinels make a
typo dangerous, which is where #3253 asked for them.

The arms cover empty, null, a unit suffix, a spaced unit, a thousands separator, leading and trailing
whitespace, a surviving quote, a word, a sign, hex, a decimal point, and both overflow shapes — and
each asserts **what the old spelling produced** beside what the new one keeps, because "keeps the
default" means nothing without it. Then the consequence, in the policy's own terms: the value the old
parse handed back is not a no-op but the opposite arm, for both the defer minimum and the promotion
budget.

## Mutation arms

Both were built and run, so the failure is the test's rather than the compiler's.

| --- | --- | --- | --- |
| `parse_u64_strict` stops at the first non-digit, i.e. behaves like a bare `strtoull` | `0` | `8` | **17** failures — every typo shape, plus both policy consequences |
| a refused value returns `0` instead of the default (the defect itself) | `0` | `8` | **5** failures — the three load-bearing arms and both policy consequences |

Reverted; the tree is clean and the run above is on it.

## Risks

- **A previously-accepted value is now refused.** `=8mb` used to mean 8, `=-1` used to mean unbounded.
  Anything relying on that was relying on an accident, and the refusal is loud — but it is a real
  behaviour change for a hand-typed A/B. Nothing in the tree sets these: `grep` over `*.py`, `*.sh`,
  `*.json`, `*.ps1`, `*.yml` finds only prose in `docs/`.
- **A well-formed `0` is still honoured.** This refuses typos, not the aggressive setting itself.
- The `check_cached_env.py` carve-out widens a deliberately conservative heuristic by exactly two
  names. A future `env_`-prefixed function that *writes* would need adding to the list, not removing
  from it.
- **An empty value now takes the default silently on `PROSPER_COLD_STORAGE_SNAPSHOT_MIN_MB` too.**
  Its old parse treated `FOO=` as text and fell back on the end-pointer check; the shared helper treats
  unset and empty alike and says nothing. That is the right call — `FOO=` is not a typo anyone makes by
  mistyping a *number* — but it is a behaviour change on that one knob and belongs in this list.
- **`env_u64_or_default_capped` saturates silently.** Only a malformed value is reported. A well-formed
  but absurd number is someone asking for "as large as possible", where clamping is the expected answer
  rather than a surprise, and a typo cannot reach the clamp because it was refused a line earlier.

## Review response

Approved with nothing to fix. The reviewer compiled a standalone probe rather than taking the
`strtoull` table from this body, and all three rows verified. Four suggestions, all taken in the second
commit:

- **Two more rows worth having**, now in the test. `0x10` reaches **0** under base 10 — the sharpest
  case in the set, because someone reaching for hex on a byte-valued knob lands on the aggressive
  sentinel rather than on a large number — and a trailing newline, which a heredoc or `$(cat file)`
  supplies without anyone typing it, parses today and is refused now.
- The refusal message read *"not a plain non-negative integer of unchanged validations"* for a
  countable knob; it now reads *"not a plain non-negative count of &lt;unit&gt;"*, with the no-unit form
  unchanged.
- The silent saturation and the empty-value change are in the risk list above.
- **`src/diagnostics/AGENTS.md` enumerated the folder's shared headers positionally** ("`env_cache.hpp`
  is the third always-reachable file"), so this change did not leave that map incomplete — it made it
  **wrong**. Rewritten, including the coupling that is genuinely non-obvious: both headers take the
  variable name as a literal argument, so a new reader here has to join `check_cached_env.py`'s
  `ENV_READERS` set or it registers as an arming.

## Not fixed here

The same lax spelling appears at **169** other env-derived numeric parses in the tree, against 71 that
check an end pointer. Most are knobs where 0 means "off", so the same spelling is far less dangerous,
and each needs the question "what does 0 mean here?" answered before conversion. Filed as #3267 with
the census, the per-file distribution and a suggested triage order rather than swept blind inside this
PR.

## Docs

`docs/FRONTEND_APP.md` states the accepted grammar where the knobs are documented, and records why the
old spelling was dangerous rather than merely sloppy.

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -1273,7 +1273,9 @@ if(TARGET prosper_core)
   add_test(NAME capture_compute_policy COMMAND test_capture_compute_policy)
 
   add_executable(test_write_watch_policy frontends/shared/tests/test_write_watch_policy.cpp)
-  target_include_directories(test_write_watch_policy PRIVATE frontends)
+  # `src` as well as `frontends`: the knobs' strict parsing lives in diagnostics/env_numeric.hpp,
+  # and it is asserted here beside the policy whose sentinels make a typo dangerous (#3253).
+  target_include_directories(test_write_watch_policy PRIVATE frontends src)
   add_test(NAME write_watch_policy COMMAND test_write_watch_policy)
 
   add_executable(test_texture_decode_diagnostic
--- a/prosper/docs/FRONTEND_APP.md
+++ b/prosper/docs/FRONTEND_APP.md
@@ -451,17 +451,25 @@ page-protection write watch: texture sources below 1 MiB stay on exact compariso
 immediately, and larger
 sources arm after three exact unchanged validations. Promotions are limited to 8 MiB per ordered submit,
 with at most one source larger than that limit, and adjacent pages are protected in coalesced runs. Set
-`PROSPER_TEXTURE_WRITE_WATCH_DEFER_MIN_KB`, `PROSPER_TEXTURE_WRITE_WATCH_PROMOTE_HITS`, or
-`PROSPER_TEXTURE_WRITE_WATCH_PROMOTE_MB` for controlled A/B runs.
+`PROSPER_TEXTURE_WRITE_WATCH_DEFER_MIN_KB`, `PROSPER_TEXTURE_WRITE_WATCH_MIN_KB`,
+`PROSPER_TEXTURE_WRITE_WATCH_PROMOTE_HITS`, or `PROSPER_TEXTURE_WRITE_WATCH_PROMOTE_MB` for controlled
+A/B runs.
+
+**Every one of these takes a plain decimal integer and nothing else, and a malformed value is refused
+on stderr while the default stands (#3253).** That is not pedantry about input: on each of them **0 is
+a meaningful and maximally aggressive setting** — defer nothing, no size floor, promote on first sight,
+unbounded arming — and a bare `strtoull` answers 0 for anything it cannot parse. So `=8mb`, `=8 KB` or
+a stray quote used to select the *opposite* arm from the one you asked for, with nothing in the run's
+output saying so. No sign, no `0x`, no units, no surrounding whitespace; unset takes the default in
+silence.
 
 Compute buffer/image residency uses the same exact-first rule with `PROSPER_COMPUTE_WRITE_WATCH_PROMOTE_HITS`
 and `PROSPER_COMPUTE_WRITE_WATCH_PROMOTE_MB` — **but it does not have the "1-8 MiB arm immediately" half**,
 and this paragraph claimed it did until #3155. That path has always passed a defer minimum of **one byte**,
 which makes the size exemption unreachable, so every compute source must climb the stability ladder however
 small it is. `PROSPER_COMPUTE_WATCH_DEFER_MIN_KB` exposes that minimum for A/B (unset = 1 byte, i.e. today's
-behaviour; `=8192` gives renderer parity; `=0` defers nothing). A malformed value is refused loudly and keeps
-the default — unlike the four older variables beside it, where a typo silently selects the most
-aggressive setting (#3253). Whether the asymmetry is right is open —
+behaviour; `=8192` gives renderer parity; `=0` defers nothing). Like the rest of the family since #3253, a
+malformed value is refused loudly and keeps the default. Whether the asymmetry is right is open —
 see `RENDERER_PERFORMANCE_2026_07.md` § *Compute write-watch promotion census*.
 
 `PROSPER_WATCH_PROMOTE_CENSUS=1` reports what any of those settings actually bought, every 256 submits and
@@ -478,7 +486,9 @@ if another architectural writer invalidates it, the next invocation takes ordina
 establishes the exact baseline needed for later identical-result skips. A failed attempt to replace a
 host result fallback with a GPU baseline invalidates that obsolete fallback before returning. Partial,
 readable, and replay-owned targets retain their original exact contract. Set
-`PROSPER_COLD_STORAGE_SNAPSHOT_MIN_MB` to retune the crossover. The broader
+`PROSPER_COLD_STORAGE_SNAPSHOT_MIN_MB` to retune the crossover — it validated its input before #3253 but
+fell back in silence, which fails the other way: a typo left you measuring the default while believing
+you had moved it. The broader
 `PROSPER_NO_ADAPTIVE_STORAGE_RESULT_VALIDATION=1` control restores immediate storage-result snapshots.
 
 A sampled image that successfully imports an authoritative renderer-owned Vulkan image does not validate,
--- a/prosper/frontends/shared/live/live_compute.cpp
+++ b/prosper/frontends/shared/live/live_compute.cpp
@@ -10,6 +10,7 @@
 #include "shared/device/vulkan_device_select.hpp"
 #include "shared/texture/write_watch_census.hpp"
 #include "shared/texture/write_watch_policy.hpp"
+#include "diagnostics/env_numeric.hpp"   // #3253: a typo must not select a different setting
 #include "shared/perf/performance_capture.hpp"      // bounded F8 post-trigger compute timing
 #include "shared/perf/performance_timing_policy.hpp" // F8 measures without enabling verbose timing logs
 
@@ -1105,9 +1106,13 @@ VkDeviceSize persistent_compute_image_limit(
 
 uint32_t compute_write_watch_promotion_validations() {
     static const uint32_t value = [] {
+        // 0 here is not "off": `stable >= 0` is always true, so every source arms on first
+        // acquisition. A malformed value must therefore keep the default rather than becoming it
+        // (#3253).
         const char* text = std::getenv("PROSPER_COMPUTE_WRITE_WATCH_PROMOTE_HITS");
-        const uint64_t parsed = text ? std::strtoull(text, nullptr, 10) : 3ull;
-        return static_cast<uint32_t>(std::min<uint64_t>(parsed, UINT32_MAX));
+        return static_cast<uint32_t>(prosper::diag::env_u64_or_default_capped(
+            "PROSPER_COMPUTE_WRITE_WATCH_PROMOTE_HITS", text, 3ull, UINT32_MAX,
+            "unchanged validations"));
     }();
     return value;
 }
@@ -1133,20 +1138,20 @@ uint32_t compute_write_watch_promotion_validations() {
 // Pair it with PROSPER_WATCH_PROMOTE_CENSUS, which reports what each setting actually bought.
 size_t compute_write_watch_defer_min_bytes() {
     static const size_t value = [] {
+        // The historical default is 1 BYTE, not 1 KiB, so it cannot be expressed in this knob's own
+        // units -- hence the explicit unset case rather than a KiB default handed to the helper.
         constexpr size_t historical_default = 1;
         const char* text = std::getenv("PROSPER_COMPUTE_WATCH_DEFER_MIN_KB");
         if (!text || !*text) return historical_default;
-        // Refuse a malformed value LOUDLY and keep the default. A bare `strtoull` answers 0 for
-        // anything non-numeric, and 0 here means "defer nothing" -- the most aggressive setting on
-        // the knob. A typo must cost you the experiment, never hand you a different one you did not
-        // ask for and cannot see in the run's own output.
-        char* end = nullptr;
-        const uint64_t kib = std::strtoull(text, &end, 10);
-        if (end == text || (end && *end)) {
+        // #3251 gave this knob a strict parse; #3253 moved that parse into diagnostics/env_numeric
+        // so the whole write-watch family shares one, rather than this one being the exception. The
+        // sentinel it protects: 0 means "defer nothing", i.e. arm every source on first acquisition.
+        uint64_t kib = 0;
+        if (!prosper::diag::parse_u64_strict(text, &kib)) {
             std::fprintf(stderr,
-                         "[watch-policy] PROSPER_COMPUTE_WATCH_DEFER_MIN_KB='%s' is not a number "
-                         "-- the compute defer minimum stays at its default and NOTHING is being "
-                         "A/B'd\n",
+                         "[env] PROSPER_COMPUTE_WATCH_DEFER_MIN_KB='%s' is not a plain "
+                         "non-negative integer of KiB -- keeping the default (1 byte) and NOTHING "
+                         "is being A/B'd\n",
                          text);
             return historical_default;
         }
@@ -1158,11 +1163,13 @@ size_t compute_write_watch_defer_min_bytes() {
 
 size_t compute_write_watch_promotion_budget_bytes() {
     static const size_t value = [] {
+        // 0 here is not "off" either: WriteWatchPromotionBudget::try_consume returns true
+        // unconditionally when the byte limit is 0, i.e. UNBOUNDED arming per submit (#3253).
         const char* text = std::getenv("PROSPER_COMPUTE_WRITE_WATCH_PROMOTE_MB");
-        const uint64_t mib = text ? std::strtoull(text, nullptr, 10) : 8ull;
-        return static_cast<size_t>(
-            std::min<uint64_t>(mib, SIZE_MAX / (1024ull * 1024ull)) *
-            (1024ull * 1024ull));
+        const uint64_t mib = prosper::diag::env_u64_or_default_capped(
+            "PROSPER_COMPUTE_WRITE_WATCH_PROMOTE_MB", text, 8ull,
+            SIZE_MAX / (1024ull * 1024ull), "MiB");
+        return static_cast<size_t>(mib * (1024ull * 1024ull));
     }();
     return value;
 }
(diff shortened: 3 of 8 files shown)


mattias800 (commented):
## Independent review — #3268 (strict env numeric parsing)

**APPROVED.** No blocking findings. Four suggestions, all small.

Per this session's constraints I did not build the project or run ctest. I did compile a ten-line
standalone C file to check the `strtoull` claims against glibc rather than take them from the PR
body — that seemed like the one place an independent number was worth more than agreement.

---

### What I verified first-hand

**The three-wrong-answers claim is exactly right.** Measured on this box's glibc:

```
eight                    -> 0
yes                      -> 0
"8192"                   -> 0
8mb                      -> 8
8 KB                     -> 8
8,192                    -> 8
8.5                      -> 8
-1                       -> 18446744073709551615
18446744073709551616     -> 18446744073709551615
 8                       -> 8
0x10                     -> 0
```

Two rows worth adding to your own list: `0x10` reaches **0** under base 10 (it parses the leading
`0` and stops at `x`), so a hex-shaped typo lands on the *aggressive* sentinel rather than on 16 —
which strengthens the argument rather than weakening it. And a trailing newline parses fine today
and is refused now (see S1).

**The 1024x direction is asserted correctly, in both arms.** I checked it against
`should_promote_write_watch` rather than against the prose:

```cpp
return defer_min_bytes == 0 || source_bytes < defer_min_bytes ||
       stable_exact_validations >= promotion_validations;
```

- `typo_zero` (0 bytes): first clause fires, so a 32 MiB source the intended 8 MiB minimum defers is
  armed on first sight. The test's `CHECK(should_promote_write_watch(32u << 20, 0, typo_zero, 3))`
  is right.
- `typo_scaled` (8 KiB): a 1 MiB source is no longer under the minimum, so it must climb the ladder,
  while the intended 8 MiB minimum exempts it. `!should_promote(1<<20, 0, 8192, 3)` and
  `should_promote(1<<20, 0, 8388608, 3)` — both right.

So "a smaller defer minimum exempts fewer sources" holds, and the two arms genuinely land on
opposite sides. Given you mention catching yourself asserting this the wrong way round, I went
looking for the inverted version and it is not in the merged text.

**The overflow guard is exact**, not merely conservative. `value > (UINT64_MAX - digit) / 10u`
rejects iff `value*10 + digit > UINT64_MAX`; I worked both directions of the floor-division
inequality. `18446744073709551615` accepted and `…616` rejected, which the test pins.

**Nothing in the tree sets these knobs.** I re-ran the sweep over `*.py *.sh *.ps1 *.yml *.yaml
*.json *.md`: the only hits outside `docs/` prose are `RENDERER_PERFORMANCE_2026_07.md` and
`GAME_COMPAT_ORCHESTRATION.md`, both describing rather than setting. So the "previously-accepted
value now refused" risk really is confined to a hand-typed A/B.

**The `check_cached_env.py` carve-out is minimal and correct**, and I looked at it specifically
because a gate widened to admit its own author's change is worth a second reader. Three things make
it fine:
- The two added names are pure readers. `env_u64_or_default{,_capped}` cannot write an environment
  variable; the name is a literal first argument solely so the refusal message can print it.
- The carve-out removes a *reader* from the armed set. It does not weaken detection of a real
  `setenv` — `test_write_watch_policy.cpp` never calls `setenv` at all; it passes text directly, so
  the gate had nothing genuine to catch here.
- It is an explicit `frozenset`, not a loosened pattern, so the failure direction on a future
  `env_`-prefixed *writer* is "must be added", not "silently exempt". The comment says this.

The forward reference (`armed_callee` uses `ENV_READERS`, which is bound later in the module) is
legal Python and resolves at call time. Fine, if slightly surprising to read.

**CI:** 11 jobs green on `c04e13276`, Windows MinGW and macOS included.

---

## Suggestions (none blocking)

**S1 — one behaviour change is not in the risk list.** `PROSPER_COLD_STORAGE_SNAPSHOT_MIN_MB` set to
the **empty string** used to yield **0**, not the default. The old expression was
`value && (!end || *end) ? 16ull : parsed` — for `""`, `strtoull` returns 0 with `*end == '\0'`, so
the guard is false and `mib = parsed = 0`. The new helper treats empty as "not a typo" and returns
16. Better behaviour, and pathological input, but the PR's risk section says "unset or empty takes
the default in silence" as if that were the status quo for this site, and for this one site it isn't.
Also worth naming: a trailing newline or trailing space (`=8 `, `=8$'\n'`) used to parse and is now
refused. Low risk, deliberate, and exactly the sort of thing that should be written down rather than
discovered.

**S2 — saturation is silent, which is the same shape the header exists to remove.**
`env_u64_or_default_capped` prints for a malformed value but says nothing when it clamps. So
`PROSPER_TEXTURE_WRITE_WATCH_PROMOTE_MB=99999999999999` quietly becomes a different setting from the
one asked for — the header's own stated hazard, one layer up. The old code did the same, so this is
not a regression, but a one-line message on clamp would make the helper consistent with its own
docstring.

**S3 — the `unit` string reads oddly at two call sites.** The message template is
`"is not a plain non-negative integer%s%s"` with `" of "` + unit, so
`"unchanged validations"` yields *"is not a plain non-negative integer of unchanged validations"*.
The parameter is really "units of", and "unchanged validations" is a noun the sentence can't take.
Either pass `nullptr` for the two `PROMOTE_HITS` sites or reword to something like `"validations"`.

**S4 — `src/diagnostics/AGENTS.md` now undercounts its own folder.** It enumerates the
always-reachable files positionally — `diag_clock.hpp` is "the second thing on the always-reachable
side", `env_cache.hpp` "the third always-reachable file" — and `env_numeric.hpp` makes four with no
entry. Since the map is written as an enumeration, adding a file without a line makes it wrong
rather than merely incomplete. One sentence would do it, and it is a natural place to say the thing
the header says well: `env_cache.hpp` answers "was this set, and to what text?", `env_numeric.hpp`
answers "what number is that text, and is it a typo?".

---

### On the parts that needed judgement rather than checking

The choice to put the parser in `diagnostics/` beside `env_cache.hpp` rather than in the write-watch
policy header is right, and the reason given (three of six call sites are in a different subsystem;
parsing is not a write-watch concept) is the reason I'd give. Folding #3251's own site into the
shared helper while keeping its explicit unset case — because its historical default of 1 *byte*
cannot be spelled in a KiB knob — is the correct call; the alternative would have been to change a
default while claiming to change only a parser.

The tests are unusually well-shaped for this kind of change: asserting **what the old spelling
produced** next to what the new one keeps is what makes "keeps the default" mean anything, and the
two arms that carry the assertion into `should_promote_write_watch`/`WriteWatchPromotionBudget` are
what turn a parser test into a statement about the instrument. The two mutation arms (17 and 5
failures) demonstrate the tests can actually see both halves of the defect.

Verdict: **APPROVED**.

mattias800 (commented):
## Re-review — #3268 at `7a25b336d`

**APPROVED.** No blocking findings. One cosmetic nit, which I would not hold a merge for.

The stated reason for re-review was the 16 lines in `env_numeric.hpp` — "a change to the parser after
a parser review is exactly where a quiet regression fits". That is the right instinct, and the
reassuring answer is that **the parser was not changed at all**. I checked by extracting each
function body at both heads and hashing it:

```
parse_u64_strict            21beaa8e17a11a0a84679fe3b8656950  (c04e13276)
                            21beaa8e17a11a0a84679fe3b8656950  (7a25b336d)
env_u64_or_default_capped   260257134b7c4510c5bd222afeb33fd4  (c04e13276)
                            260257134b7c4510c5bd222afeb33fd4  (7a25b336d)
```

Byte-identical. The whole executable surface of the delta is **one `fprintf` format string** in
`env_u64_or_default`; the other 14 lines are two comment blocks. So the accepted grammar, the
overflow guard and the saturation arithmetic I verified last time all still stand unexamined-because-
unchanged, rather than on trust.

### The one behavioural line, checked

```c
"[env] %s='%s' is not a plain non-negative %s%s -- keeping the default (%llu) and changing NOTHING\n",
name, text, unit ? "count of " : "integer", unit ? unit : "", (unsigned long long)fallback
```

Five conversions, five arguments, types match. Both branches render correctly:

- `unit == nullptr` → *"is not a plain non-negative integer"* — the `%s%s` collapses to `"integer"` + `""`.
- `unit == "MiB"` → *"is not a plain non-negative count of MiB"*.
- `unit == "unchanged validations"` (both `PROMOTE_HITS` sites still pass this on the final head —
  `live_compute.cpp:1115`, `live_renderer.cpp:3879`) → *"is not a plain non-negative count of
  unchanged validations"*, which is grammatical. Leaving the call sites alone and fixing the template
  was the better of the two ways to resolve that nit.

### The other three

**The empty-value change is now in the risk list** (body line 117), which is all my suggestion asked
for — it was a "name it", not a "change it", and changing it would have been wrong.

**The saturation note is honest and the decision is right.** I pushed on the claim that "a typo cannot
reach here": it holds in the sense that matters, because anything reaching the cap has already passed
`[0-9]+`, and clearing the cap needs ~1.76e13 on the MiB knobs (`SIZE_MAX / 1 MiB`) or ~4.29e9 on the
hits knobs — not a fat-finger distance from any plausible value. A well-formed absurd number really is
someone asking for "as large as possible", and clamping is the expected answer. Keeping it silent is
correct; documenting that it is silent is what made it reviewable.

**The two new `strtoull` arms match my own measurements**, which is why I flagged them: `0x10` → 0
under base 10 and `"8192\n"` → 8192 are both what my probe printed, and `parse_u64_strict` rejects
both. Neither arm is vacuous — without the strict parser each `!parse_u64_strict` fails. The hex case
is worth the line it got: reaching for `0x2000` on a byte knob lands you on the aggressive sentinel,
not on 8192, which is the sharpest single illustration of the whole issue.

### Nit (not blocking)

`src/diagnostics/AGENTS.md` — the edit made the subject plural but left the rest of the sentence
singular, and in doing so re-attributed the macros:

> `env_cache.hpp` and `env_numeric.hpp` are the remaining always-reachable files, and they are here
> for the same reason as `diag_clock.hpp`: **it is shared, and it belongs to no one subsystem. It
> holds** `PROSPER_ENV_ON` / `PROSPER_ENV_VALUE`…

`PROSPER_ENV_ON` / `PROSPER_ENV_VALUE` belong to `env_cache.hpp` alone, so as written the paragraph
now says the pair holds them. A partial-edit wart rather than a wrong claim about the code, and the
new closing paragraphs are good — especially the last one, which records the `check_cached_env.py`
`ENV_READERS` coupling as the non-obvious cost of adding a function here. That is precisely the kind
of thing an `AGENTS.md` is for, and it is the part a future contributor would otherwise learn by
reddening CI.

CI green on `7a25b336d` across all 9 non-skipped jobs.

Verdict: **APPROVED**.

Issue #2937, opened 2026-08-22
Five Vulkan-execution tests fail on RADV at a SHA where CI is green — every failing assertion is an INDEXED draw, and CI only ever runs them on lavapipe


mattias800:
## Summary

On **RADV** (AMD Radeon 8060S / STRIX_HALO, Linux), five of prosper's Vulkan-execution tests fail at
master `bffa40d4`, and **every failing assertion is about an indexed draw**. The non-indexed
assertion in each matched pair passes, in the same process, on the same device, in the same frame.

**CI is green on the same SHA** — `.github/workflows/ci.yml` runs `ctest` under Mesa **lavapipe**
(`VK_ICD_FILENAMES=…/lvp_icd.x86_64.json`), because the runner has no GPU. So this class of
divergence is structurally invisible to CI: the Vulkan-gated suite is registered and run there, and
it passes, on a software rasteriser.

## Symptom

```text
ctest --no-tests=error -j4   ->  98% tests passed, 5 tests failed out of 301   (exit 8)
        279 - recompiled_shaders_render (Failed)
        282 - multidraw_render (Failed)
        283 - indexed_render (Failed)
        284 - descriptor_array_render (Failed)
        293 - gpu_execute (Failed)
```

Deterministic: `--rerun-failed --output-on-failure` gives 0 of 5 passing, and running the binaries
directly reproduces identically.

## The internal control is what makes this specific

`test_indexed_render`, whole output:

```text
== test_indexed_render ==
  [ok]   recompiled vertex-fetch VS + green PS
  [ok]   both quad renders produced frames
  [FAIL] indexed [0,1,2,2,3,0] TRIANGLE_LIST quad fills the viewport GREEN
  [FAIL] indexed quad is byte-identical to the old perimeter-fan rendering
  [ok]   indexed triangle rendered
  [FAIL] indices [1,2,3] fetched records 1..3 (fullscreen GREEN triangle)
  [ok]   dropping the indices changes the picture (indices are really applied)
  [ok]   non-indexed firstVertex selects records 1..3 from a shared vertex pool
  [FAIL] indexed vertexOffset selects records 1..3 from a shared vertex pool
== FAIL: 4 ==
```

Read the two `firstVertex` / `vertexOffset` arms together: **the same vertex pool, the same shader,
the same target — non-indexed selects records 1..3 correctly and indexed does not.** And
`dropping the indices changes the picture` **passes**, so the index buffer is reaching the GPU and
[...]

Issue #2948, opened 2026-08-23
The GPU-serialisation check in every lane brief has never worked: pgrep -x takes one pattern


mattias800:
## The check that guards the serial GPU has never worked

Every concurrent lane in this project is told to run this before touching the GPU:

```bash
pgrep -x prosper-app screenshot boot_trace
```

`pgrep -x` **accepts only one pattern.** Measured just now:

```
$ pgrep -x prosper-app screenshot boot_trace
pgrep: only one pattern can be provided
Try `pgrep --help' for more information.
$ echo $?
2
```

So the guard has never answered its question. It emits a usage error and exits **2** — not 0, not 1 —
regardless of what is running. A lane reading "non-zero means nothing found" proceeds always; a lane
reading "zero means something found" never finds anything. Either way the GPU-is-serial rule was
being enforced by a command that could not enforce it.

**This is my error, not a lane's.** I wrote that line into roughly ten briefs over this session. It
was caught by a lane that noticed its own "GPU free" checks were silently erroring, not by anything
in CI or the tooling.

## Why it survived

It is the classic shape: an instrument whose failure looks like a clean negative. Nobody checks the
exit status of a check that is *supposed* to find nothing most of the time, and the usage message
goes to stderr where a `2>/dev/null` or an `|| echo "GPU free"` swallows it. Several of my own briefs
literally wrote `pgrep ... || echo "  GPU free"`, which prints "GPU free" on the usage error.

No result is known to be corrupted — `snapshot.py` takes a real cross-worktree lock, and that is what
has actually been serialising the snapshot runs. But bare `tools/screenshot` and `boot_trace` runs
have had no interlock at all, and two lanes have been running concurrently for most of a day.

## The fix

```bash
pgrep -x 'prosper-app|screenshot|boot_trace'
```

`-x` matches the whole name and the alternation is a single ERE pattern, so this is one argument.
Verified: exit 1 with none running, and it matches correctly when one is.

## Wanted

1. Correct the line wherever it appears in briefs, docs and scripts — `grep -rn "pgrep -x" prosper/`.
2. A numbered instrument-trap row, allocated with `python3 prosper/tools/docs/trap_number.py` (it
   scans master *and* open PRs).
3. Worth considering: a small `tools/gpu_busy.sh` that answers the question properly and exits 0/1,
   so briefs cite a tested thing rather than an inline command nobody runs in isolation. The charter
   is explicit that a committed tool beats a throwaway line.

mattias800:
[...]

PR #3270, opened 2026-09-03
#2945 re-measured: 3,875 replays give one hash per arm — and a campaign tool that answers UNDECIDED instead of "fixed"

mattias800:
## What this is

#2945 reports that one frozen `.prgbundle`, replayed offline with `gpu_replay`, produces a different
picture from run to run — 15 replays of one *BALAN WONDERWORLD* submit giving **5 distinct output
hashes** on 2026-08-23. This PR re-measures that on current `main`, lands the campaign apparatus as
a tool so the next agent does not rebuild it, and fixes four instrument defects found on the way.

**No behavioural change to prosper.** Every file is a tool, a test, or a doc.

## The measurement

Five campaigns, run through `tools/determinism/replay_determinism.sh`. Linux / AMD Radeon 8060S
(RADV STRIX_HALO), kernel **7.2.1** (the August campaigns ran on 7.1.5).

| campaign | binary / driver | subject | replays | distinct hashes | own span |
| `new` | current `main`, host Mesa 26.2.1 | `s3537` whole submit / `--draw 42` | 707 / 706 | **1 / 1** | 1.75 h |
| `old` | rebuilt `08c23efd` (the commit the capsule carries) | same | 701 / 701 | **1 / 1** | 1.75 h |
| `container` | current `main`, container Mesa **26.1.4** | same | 296 / 295 | **1 / 1** | 1.13 h |
| `stray` | current `main`, host | a seven-submit *Stray* `.prgbundle` | 469 | **1** | 1.73 h |
| `fixedparse` | current `main`, host — the corrected-control re-run | `--draw 42` | 600 | **1** | 0.48 h |

**3,875 replays in the first four campaigns, one output hash per arm**, against the 2026-08-23
figure of 5 distinct over 15 (a 53% per-replay non-modal rate). 95% upper bound on the largest arm:
**0.42%**. Six full `ctest --no-tests=error -j4` runs taken inside the window were **345/345 every
time** — the `-j4` regime #2937 was found in.

**Three things the table does not say, all of which narrow the claim rather than the verdict:**

1. **The four campaigns ran concurrently** — their spans sum to 6.37 h inside a **1.75 h** wall
   window, so the dataset samples 1.75 h four times over, not seven hours. They were also load for
   each other, and `peers()` cannot see a sibling `gpu_replay`, so **no row was taken on an idle
   GPU**. The `gpu_pct` column is the ground truth; the condition label (now `no-selfload`) records
   only this invocation's own intent.
2. **Every subject row carried a real hash — 0 of 3,875 recorded `none`.** Checked, not assumed:
   the runner recorded exactly that for every bundle replay until this PR taught it that `--bundle`
   prints no `output=` line. The attached reports print non-answers as `[not a sample]`, so their
   absence is the evidence.
3. **The control figure was re-taken.** The original parse read only the first of `vkprobe`'s
   per-pattern readback lines, supporting only "no round's lexicographically-smallest pattern
   differed from the correct one". Aggregated across every line and re-run: **600 rounds, 0 fires**.
   Weak claim and strong claim now agree.

## Why the verdict is UNDECIDED and not "fixed"

**The bare-Vulkan control fired in 0 of 2,172 rounds, and 0 of 600 in the corrected re-run.**
`tools/vkprobe` is in every round precisely so a clean subject can be told apart from a quiet
window — this issue's first `## Ruled out` row records the identical command measuring 0/20 in one
ten-minute window and 12/12 in the next. A campaign whose control never fired has not shown it met a
window in which the class is expressible, so `replay_determinism_report.py` prints `UNDECIDED`, and
the sentence to quote is **"no instance observed in 3,875 replays over a 1.75 h window"**.

That reframes the open question. The finding worth acting on is not "is prosper clean" but **the
failing regime cannot be reproduced on this box at all** — with the documented recipe, then fourteen
concurrent replay processes, then six full `ctest -j4` runs, sustained for 1.75 h.

The harness's ability to report the class was established by hand, not assumed:
`--override-resource 42:vs:3 <192 zero bytes>` reproduces the characterised mechanism and the replay
then outputs exactly `a5e7b61cbf984383`, the documented failure value. The control's failure path
was fired by hand too, by running it with identity indices so the correct readback is a different
pattern.

## The finding that was not expected

**The same binary and the same capsule render a different picture today than on 2026-08-23.**
Rebuilt from the exact commit the capsule carries, `--draw 42` returns `ccc433ff6d980383` — 5 of 5
on the host and 3 of 3 in the container on **Mesa 26.1.4, the version the original figure used**.
Current `main` agrees, 706 of 706. So the moved value is neither prosper's nor a Mesa
userspace-version difference; the kernel is the one environment variable known to have moved (n=1
box, correlation only). It is a real render, not a new failure mode: it differs from the untouched
seed on 100% of sampled pixels. Caveat: the rebuild is the same source at the same commit, not
provably the same object bytes.

**Consequence: no hash recorded in that thread is a durable oracle.**

## What is in the diff

- **`tools/determinism/`** (new) — `replay_determinism.sh` runs the campaign; `replay_determinism_report.py`
  reads its CSV and issues one of three verdicts. The third is the point and is what the ctest case
  pins, including a subject row that recorded **no** hash: a campaign of non-answers has one value
  per arm and no varying arm, and must not read as `DETERMINISTIC`.
- **`tools/vkprobe/correlate_with_replay.sh`** — keyed its verdict on a hardcoded GOOD hash, which
  has moved. It now **derives** the failure key from the capsule (draw target → `rtt-seed` hash);
  this PR is itself the proof that hardcoded hashes rot.
- **`docs/GRAPHICS.md`** — six rows: the good/failure hash asymmetry, the same-binary observation,
  the hand-made positive instance, what "heavy GPU load" actually measures (0–2% → 5–9% on the
  documented recipe), the re-measurement, and the corrected control re-run.
- **`BLOG.md`**, **`tools/AGENTS.md`**, **`CMakeLists.txt`** — the entry, the folder entry, the test.

## Verification

Exact head `52799bf3b`, merged with `origin/main`, built in the project container.

- Build `-j8` → **exit 0**; `ctest --no-tests=error` → **348/348, exit 0**.
- By name `-R '^replay_determinism_report$'` → **1/1, exit 0**.
- **Mutation arms, both at ctest level.** Defeating `control_failed()` → **exit 8**; defeating
  `subject_answered()` (count `none` as a hash) → **exit 8**. Reverting each → **exit 0**. Both are
  also in-process arms in the test, asserting the mutant flips the verdict *and* that the unmutated
  evaluator does not — so neither can pass vacuously.
- Docs gate clean; `--baseline origin/main` on `GAME_COMPAT_ORCHESTRATION.md` → 251 rows, none
  deleted. `git diff --check` clean on every commit.
- Campaign report output for all five campaigns attached in a PR comment.

## Risks

The tool is new and off every existing path. The only shared files touched are `CMakeLists.txt` (one
`add_test` in the platform-independent block), two docs and `BLOG.md`.
`correlate_with_replay.sh` changes its CSV verdict vocabulary from `good/bad/other` to
`bad/rendered/no-hash`; it has no in-tree callers.

Refs #2945


mattias800 (commented):
# Independent review — #3270 (re-measurement of #2945 + `tools/determinism/`)

**REJECTED** on two findings, both small and both one-sentence-to-few-line fixes. I want to be clear
about the shape of this: the *thesis* of this PR is right, the three-verdict design is right, and
the `UNDECIDED` framing is the correct reading of what was measured. I expect to approve on the
corrected head. What I am rejecting is (1) a reachable hole in the reader that is the same class of
defect as the one the author fixed mid-run in the producer, and (2) a causal clause in a new
`## Ruled out` row that the PR's own neighbouring row falsifies.

## What I verified first-hand, and what I took on trust

Verified by running or reading, no build, no GPU, no `gpu_replay`:

- `python3 prosper/tools/determinism/test_replay_determinism_report.py` → `== PASS ==`, exit 0.
- The mutation arm is real *and* both halves of `control_failed` are pinned. Besides the in-file
  arm, I applied a second, independent source mutation — removing only the `row.rc == 0` clause —
  and the test goes red on exactly the two arms that assert it
  (`a control that could not RUN does not count as fired`, `...and the report says the control was
  broken`), exit 1. So the arm is not vacuous and it is not pinning only the `!= "pass"` half.
- I tried to obtain `DETERMINISTIC` from a dead control by hand. With the shipped producer's
  vocabulary you cannot: every control row `pass`/rc 0 ⇒ `UNDECIDED`. That property holds.
- `git diff --check origin/main...HEAD` clean. Exactly 4 deleted lines in the whole diff, all in
  `correlate_with_replay.sh` — no trap-41 clobber from the merge commit.
- No absolute host paths anywhere in the diff (`/home/`, `/var/home/`, `/Users/`, `/mnt/c/`).
- `bash -n` clean on both shell files.
- Docs gate: `check_numbered_table.py` over every tracked `.md` → clean, rc 0. This PR touches no
  numbered table, so the `--baseline` arm is moot here.
- Arithmetic in the PR body / `GRAPHICS.md`: 707+706+701+701+296+295+469 = **3,875** ✓;
  rule-of-three 3/707 = **0.424%** ✓; 8 non-modal of 15 = **53%** ✓; ~2,173 rounds ≈ the quoted
  2,172 ✓.
- `gpu_capture_hash` call sites: `gpu_replay.cpp:714` (`rtt-seed`, **stdout**) and
  `gpu_replay.cpp:3828` (`output=`, **stderr**) both hash a byte vector with the same function, and
  `docs/GRAPHICS.md:1310` already records the `rtt-seed` line for `addr=0000009fc0000000` as
  `hash=a5e7b61cbf984383` — the failure value. The documented recipe pipes stdout, which is where
  `rtt-seed` actually goes. Item 3's premise checks out in-tree.
- `vkprobe.cpp:917-922, 1012-1016`: `Tally::indexed_words` is a `std::map<std::string,uint32_t>`
  and the reporter prints **one line per distinct pattern**.

Taken on trust (no CSV or report output is committed or attached anywhere — see NB-6): every
campaign number, the 5-of-5 and 3-of-3 rebuild replays, the GPU-utilisation range, the six
345/345 ctest runs.

---

## Blocking

### B1 — the reader counts "no hash was recorded" as a hash, so a campaign that measured nothing reads as a clean arm

`replay_determinism.sh:179` emits `"${h:-none}"` with whatever `rc` the replay exited with. A
replay that exits 0 and prints no line the extractor matches therefore lands in the CSV as a
*subject row with value `none` and rc 0* — and `replay_determinism_report.py:evaluate()` filters
only on `rc != 0`. So `none` is counted as an output hash.

I built this by hand and ran it:

    49 rounds, every subject row value=none rc=0, control fires once
    -> VERDICT: DETERMINISTIC
       "the control failed in 1 of 49 rounds while every subject arm returned one hash,
        so the campaign met the failing regime and the subject survived it"

This is not hypothetical, and it is not a foreign-CSV concern. It is **exactly the bug commit
`9a99abf9b` fixed** — "the runner recorded `none` for every bundle replay — a campaign that looked
like it ran and measured nothing". The fix went into the *producer*. The *reader* is the durable
artefact being landed, it is the thing a future agent will point at a CSV they did not generate, and
it still cannot tell "1 distinct hash" from "1 distinct non-answer".

This is also the concrete reason I cannot discharge item 5 for the `stray` arm (see NB-7): a
`stray` campaign that ran entirely on the pre-`9a99abf9b` script would produce 469 rows of `none`,
one distinct value, and would be reported as a clean arm — indistinguishable, from anything in this
PR, from the result claimed.

Fix: treat `none` / empty exactly as `rc != 0` is treated (not a sample, and say so in the reason),
and add a test arm asserting that an all-`none` campaign does **not** report `DETERMINISTIC`. Three
lines and one arm. Given this file's docstring, it belongs here more than almost anything else in
it.

### B2 — `GRAPHICS.md`'s first new `## Ruled out` row asserts a cause the fourth row rules out, with a citation that does not check out

Row 1 (new):

> BALAN's `s3537 --draw 42` rendered `9068fcf09de07383` on 2026-08-23 and `ccc433ff6d980383` on
> 2026-09-02 — **the renderer legitimately changed what that draw produces, across 43 commits
> touching `src/render` and `src/gpu`.**

Row 4 (new, ~30 lines below):

> Rebuilt from the exact commit the capsule carries (`rev=08c23efd`) ... that binary returns
> `ccc433ff6d980383` ... **So the moved value is not prosper's doing** ...

Those cannot both stand. Row 4 is the finding; row 1 states the hypothesis row 4 kills, in the
voice of an established fact, in a `## Ruled out` section. A reader who takes row 1 alone — which is
how these rows get read, one at a time, by someone checking whether their idea is already dead —
comes away believing prosper's renderer changed the picture. That is the precise failure this
section exists to prevent, and the charter's "a true statement reached by a route that does not
establish it" shape.

The supporting numbers do not reproduce either:

- `prosper/src/render` **does not exist**, at `08c23efd` or at `origin/main`
  (`git ls-tree 08c23efd prosper/src/` has no `render`; `git log 08c23efd..origin/main --
  prosper/src/render` is 0 commits). There is a `prosper/tests/render/`.
- `08c23efd..origin/main` touching `prosper/src/gpu` is **657** first-parent commits (889 with
  merges' parents), not 43. `--merges` over the same path gives **39**, the nearest thing I could
  find to the quoted figure.

Fix: drop the causal clause (the row's actual conclusion — *key the harness on the failure hash,
because the good hash demonstrably moved* — survives untouched and is correct), and either give the
exact command behind "43" or drop the count and the non-existent path.

---

## Non-blocking findings

**NB-1 — the control's pass/fail parse reads one line where `vkprobe` prints one line per pattern.**
`replay_determinism.sh:200` takes `... | head -1`, and the `awk` summation added in `c9e841592`
sums the `xN` counts *on that single line*. But `vkprobe.cpp:1015` prints a separate
`[vkprobe] b: indexed     vertex-index readback [PATTERN] xN` line for **each** distinct pattern in
a `std::map`, i.e. lexicographically ordered. So:

- `total` is the first pattern's count, never the iteration count. A round with 19 correct and 1
  empty readback is recorded as `fail:0/1`, not `fail:19/20`. The test fixture's `"fail:3/20"` shows
  the intended model, which is not what the tool produces.
- Detection depends on sort order. `--,--,--,--,--,--,--,--` (76.9% of failures in the vkprobe
  README census) sorts *before* the correct `--,--,--,4,5,6,--,--`, so those rounds do fire. But
  `1,2,3,--,--,--,--,--` (20.6%) sorts *after* it, so a round whose only failures are that pattern
  is scored **`pass`**.

The error direction is conservative — missed fires push toward `UNDECIDED`, never toward
`DETERMINISTIC` — which is why this is not blocking. But it does qualify the headline sentence: "the
bare-Vulkan control fired in 0 of 2,172 rounds" is, as implemented, "no round's *lexicographically
smallest* readback pattern was other than the correct one". Fix: drop `head -1`, sum across all
matching lines, and compare the correct-pattern count against `--control-iterations` rather than
against a self-derived total.

**NB-2 — `peers()` hand-rolls the guard `tools/gpu_busy.sh` exists to own, and swallows `pgrep`'s
error status.** `replay_determinism.sh:120-127` does `c=$(pgrep -c -x "$n" 2>/dev/null) || c=0`,
which folds rc≥2 (tool/usage error) into rc=1 (no match) and answers "no peers". `gpu_busy.sh`
refuses to do this on purpose — "NEVER answers 'free' from a broken tool", trap 222 / #2948 — and
counts five names *including `gpu_replay`*. The new function counts four and omits it. Separately,
the header comment at line 20-23 justifies the peer column by saying another lane's `gpu_replay`,
`vkprobe` or `ctest` "is exactly the load that matters and `pgrep -x prosper-app` cannot see any of
them" — but `peers()` cannot see any of them either, so the comment claims a coverage the code does
not have. Suggest calling `gpu_busy.sh` (with the campaign's own PIDs excluded) or at minimum
propagating its rc discipline, and correcting the comment.

**NB-3 — the four campaigns necessarily overlapped, so no row labelled `unloaded` was unloaded.**
The committed tool first exists at `cf02c82b3` (2026-09-02 22:10 UTC) and the results comment is at
00:26 UTC — a ≤2.3 h wall window (≤2.9 h if a prototype started at the claim). The four campaign
spans sum to 6.36 h. They ran concurrently. Combined with NB-2 (the peer detector is blind to
`gpu_replay` and `vkprobe`, which is what the other three campaigns run as), every `unloaded` row
was taken beside up to three other replay campaigns and their controls, and the peer-wait loop never
waited. The `gpu_pct` column is what saves this — mean ~11% is visibly not idle, and putting it
there so the load condition measures its own premise is the best design decision in the script. But
the PR body and `GRAPHICS.md` should not describe the blocks as "unloaded", and it is worth saying
the campaigns overlapped: the honest unit is **one ~1.75 h window sampled 3,875 times**, which the
author does correctly quote as the span.

**NB-4 — `control_failed` keys on "not exactly `pass`", which reads a non-answer as a firing.**
Verified by hand: a control row with rc 0 and an **empty** value, or with `did-not-run:...` written
with rc 0, both yield `DETERMINISTIC`. The shipped producer never emits either, so this is a
contract issue rather than a live bug — but the reader is the durable half, and the safe polarity is
positive: fire on `value.startswith("fail:")`, not on `value != "pass"`. Same family as B1.

**NB-5 — n and span are not part of the verdict.** One subject replay plus one control fire yields
`DETERMINISTIC` (verified). The tool prints the span and a "a span shorter than the drift period
samples ONE window" line, which is good, but a file whose entire argument is that nulls get
over-read might reasonably decline to say `DETERMINISTIC` below some n or span, or at least fold
both into the reason string.

**NB-6 — the evidence is not in the repo or on the issue.** No CSV, no `replay_determinism_report.py`
output, no per-arm value census is committed or attached. Every headline number rests on the
author's word, and B1 means the census is the one place a `none`-valued arm would be visible.
Pasting the four report outputs into the issue is ~20 lines each and makes the whole claim
checkable — including the `ms` column, which is the only thing that would corroborate 707 rounds of
(full submit + `--draw 42` + control + 2 s gap) fitting in 1.75 h.

**NB-7 — item 5, contamination, as far as I can take it without the CSVs.** Two of the three
disclosed instrument bugs I can discharge from the *direction* of the reported result:

- The `bc` bug (`c9e841592`): absent `bc`, `total` is empty, the equality test fails, and **every**
  round reports the control as failed → the campaign would have printed `DETERMINISTIC`. The
  reported 0-of-2,172 is the opposite outcome, so this bug cannot have produced it (either `bc` was
  present or the control rows postdate the fix). Self-refuting as a contamination source; good.
- The `SIGTERM`/three-campaigns-on-one-CSV incident: the correct trap
  (`trap 'stop_load; exit 143' TERM`) is present in the **first** committed version, so that
  incident predates every committed script. But the CSV is *appended* to
  (`[ -s "$out" ] || printf header`), `parse()` silently skips a re-appended header row, and rows
  from two campaigns sharing a `--label` pool into one arm with no guard — so "were pre-fix rows
  carried forward" is a question only the CSVs answer.
- The bundle no-hash bug (`9a99abf9b`): **not discharged**, for the reason in B1. It affects the
  `stray` arm specifically and its contaminated form is a *clean-looking* arm, not a varying one.

**NB-8 — cosmetic.** `format_report()`'s per-arm census uses `report.subjects` (all rows) while
`evaluate()` uses finished rows only, so a campaign with one timed-out replay prints
"subject X: 2 distinct" above a verdict whose reason correctly says the opposite. Worth aligning, or
labelling the unfinished rows in the census.

---

## Answers to the specific questions put to me

**1. Does the tool actually refuse to say "fixed"?** For the case it claims — every control row
`pass` — yes, and I could not break it. The mutation arm pins the property, and a second mutation I
wrote myself shows the `rc == 0` half is pinned too. Where the refusal is leaky is one step upstream
of the control check: B1 and NB-4 both let a campaign that *measured nothing* reach
`DETERMINISTIC`. So the honest summary is: the tool cannot be made to ignore a silent control, but
it can be made to bless a silent *campaign*.

**2. Is `UNDECIDED` right, or over-cautious?** Right, and not over-cautious — the PR separates the
two halves correctly. What 3,875 replays with 0 deviations establishes is a **rate bound under the
conditions sampled** (3/707 → 0.42% at 95%), and the PR states exactly that and puts the span beside
the n. What it cannot establish is absence of the class, because the pre-existing drift row in this
very section records the identical command at 0/20 and then 12/12 in consecutive ten-minute windows.
NB-3 sharpens rather than weakens this: the four campaigns are one window, not four. "No instance
observed in 3,875 replays over 1.75 h" is the sentence, and it is the sentence the PR tells people
to use.

**3. The hash drift and the rekeyed correlation script — the rekey is CORRECT, and it is an
improvement.** I checked the premise rather than the prose:

- `a5e7b61cbf984383` really is the capsule's own seed hash. `docs/GRAPHICS.md:1310` records
  `rtt-seed addr=0000009fc0000000 extent=3840x2160 format=rgba8 bytes=33177600
  hash=a5e7b61cbf984383` — same value as the "drew nothing" output hash — and both are
  `prosper::gpu::gpu_capture_hash` over a byte vector (`gpu_replay.cpp:714` and `:3828`). So "the
  draw drew nothing ⇒ the output *is* the seed ⇒ the hash *is* the seed's hash" is not an argument,
  it is already measured in-tree.
- It is durable in the way that matters: it does not depend on rasterization, so no driver, kernel
  or shader-path change can move it, and per trap 117 the `gpu_capture_hash` basis is deliberately
  frozen. I would soften "cannot move for a renderer reason" to "cannot move for a *rasterization*
  reason" — a change to output-extent selection, format conversion or the readback path would still
  move it — but that is wording, not substance.
- The rekey cannot make the script worse. The `bad` bucket is unchanged; only the old `good`/`other`
  split collapses into `rendered`, and the raw hash is still column 5, so a reader can still
  `sort | uniq -c` it and see a third state. The failure it repairs is real: with `GOOD` hardcoded
  at `9068fcf09de07383`, every round of a healthy renderer today scores `other` and the correlation
  becomes a correlation with nothing.
- One suggestion rather than a finding: the default is still a hardcoded BALAN constant, and this PR
  is the proof that hardcoded hashes rot. The script already accepts the seed as `$8` and documents
  how to read it — consider *deriving* it from the capsule by default (one `--inspect-only` at
  startup) so the durable key is obtained rather than remembered.
- On the drift claim itself: "the kernel is the one environment variable known to have moved,
  correlation, n=1 box, no arm on the old kernel" is labelled correctly and is not over-claimed.
  The rebuilt-`08c23efd` arm is what makes it interesting, and the caveat that a rebuild is the same
  *source*, not the same *object bytes*, is stated where it belongs.

**4. The hand-made positive instance.** It does what is claimed and it is outside the process that
produced the null (a forced `--override-resource`, not a stochastic replay), so it is not trap 122's
same-source control. It establishes end-to-end that the harness's recording and discrimination path
can emit a second, distinct value for this subject — which is precisely what a `NONDETERMINISTIC`
verdict needs. Worth stating one limit in the row, though: it perturbs the *input*, so it
demonstrates sensitivity to input, not sensitivity to the *timing* perturbation the class is
actually about. That is probably the strongest instance obtainable when the class cannot be induced
on demand, and saying so costs a clause.

**5. The three instrument bugs.** See NB-7. Two discharged, one — the one that matters — not.

**6. The `## Ruled out` rows.** Rows 2 (hand-made instance), 3 (what "heavy GPU load" measures),
4 (same binary, different picture) and 5 (the re-measurement) are supported by what was measured and
hedged where they should be; row 3's "the replays are CPU-bound on capsule parsing and resource
upload" is an inference from a load average rather than a measurement and could say so. Row 1 is
B2.

**7. Mechanical.** All clean, listed at the top. One number in the PR body I could not reconcile and
am not treating as a finding: 345/345 during the campaign against 348/348 at head, where this PR
adds one test — plausible given 2,082 first-parent commits landed on `main` in the same 12 days, but
worth a word if it is easy.

---

## What is good here, said plainly

The three-verdict design is the right answer to this issue and the reasoning in the docstring and
the `AGENTS.md` is the clearest statement of it anywhere in the repo. Putting `gpu_busy_percent` on
every row so a load arm measures its own premise; alternating arm order; N separate processes rather
than `--warmup-repeats`; keeping one log per replay so a future deviation has an explanation; the
`--control-iterations` note that more *rounds* beat more iterations because the failure is
per-process in shape; deleting the 24 MB BMPs as each replay is scored — these are all correct and
several are non-obvious. The refusal to report a clean campaign as a fix, when reporting it as a fix
was available and would have looked like a much better result, is the right call and is the reason
this is worth landing.

Fix B1 and B2 and I will approve on the corrected head.

mattias800:
## Both blocking findings fixed. New head `52799bf3b`.

B1 is a real defect and the counter-example is exact — thank you. Both fixes plus every non-blocking
item are in `bca47d8b4`, with the corrected control re-run in `52799bf3b`.

### B1 — the reader now refuses a non-answer

You were right that I fixed the producer and left the reader. `subject_answered()` is the single
predicate now, and — this is the part your counter-example forced — `non_answers` **derives from
it** rather than re-testing the value independently. My first attempt kept them as two separate
expressions and the mutation arm caught it immediately: defeating the predicate still produced
`UNDECIDED`, because the independent re-test was still firing. Two spellings of one rule can
disagree silently, which is the same class as the bug itself.

Behaviour: any non-answer → `UNDECIDED`; an arm with no usable replay → `UNDECIDED`; the census
prints `[not a sample] none x49` so the table cannot disagree with the verdict above it. A varying
arm still outranks a non-answer, because real variation is evidence.

Your CSV is an arm — **49 rows of `none` at rc 0 beside one control fire → `UNDECIDED`** — plus the
empty-string variant, the mixed case, varying-outranks, and a mutation requiring the empty campaign
to flip to `DETERMINISTIC` when the predicate is defeated. At ctest level that mutation reddens the
registered case (**exit 8**); reverting restores **exit 0**.

### The `stray` arm survives, and it is now checkable rather than asserted

**0 of 3,875 subject rows recorded `none`** — across all four campaigns, `value == none` is 0 and no
value is shorter than a hash. The `stray` campaign was launched from the copy taken *after*
`9a99abf9b`, so its very first row already carries a real per-submit vector:

```
1788388189,1,...,subject,stray/bundle,14069:a3dc853a29fb6101|14070:a3dc853a29fb6101|14071:14650fb0739d0383|...
```

You could not discharge it from the PR, and that was the right call — so the fixed census now makes
it self-demonstrating: it prints `[not a sample]` rows when there are any, and there are none in any
of the attached reports.

### B2 — rewritten, and I verified your corrections rather than taking them

I re-derived all three:

```
$ git ls-tree -d --name-only 08c23efd  -- prosper/src/render prosper/src/gpu   # -> prosper/src/gpu only
$ git ls-tree -d --name-only origin/main -- prosper/src/render prosper/src/gpu   # -> prosper/src/gpu only
$ git rev-list --first-parent --count 08c23efd..origin/main -- prosper/src/gpu   # -> 657
```

`prosper/src/render` exists at neither endpoint, so that half of the path matched nothing; the count
is 657 first-parent commits. My 43 came from a `--since`-filtered non-first-parent log, which now
reads 45 — it drifts, which is a fourth reason not to have cited it. And you are right that the
causal claim contradicted row four. The row now asserts only the asymmetry, derives it structurally
("drew nothing" leaves the scanout as seeded, so the output hash *is* the seed's hash, both from the
same `gpu_capture_hash`), takes your **"rasterization reason"** wording, and records what it got
wrong so the next reader does not re-derive it.

### Non-blocking, all four taken

- **Control parse.** Confirmed at `vkprobe.cpp` — `for (const auto& [pattern, count] : tally.indexed_words)`,
  one line per pattern from a map, so `head -1` read the lexicographically smallest. It aggregates
  every line now, and I demonstrated the missed fire by hand: on `[--,--,--,4,5,6,--,--] x14` +
  `[1,2,3,--,--,--,--,--] x6`, the old parse scores `pass` (14/14) and the new one `fail:14/20`.
  Rather than only restate the claim, I **re-ran it**: 600 rounds under alternating self-load,
  **0 fires, properly counted**. Weak claim and strong claim now agree.
- **Overlap.** Verified from the CSVs: spans sum to 6.37 h inside a **1.75 h** wall window. The
  label is renamed `no-selfload` — it only ever meant "this invocation started no loader" — and both
  the docs and the PR now say no row was taken on an idle GPU and that `gpu_pct` is the ground truth.
- **`peers()`.** Fails closed on a pgrep error now (a guard whose failure looks like its success is
  exactly trap 222), and states what it deliberately cannot see and why it is not `gpu_busy.sh`:
  that tool counts `gpu_replay` and `vkprobe`, which are this campaign's own tools, so calling it
  verbatim would make the campaign wait on itself forever.
- **Evidence.** Attached below.

### What this changes about the conclusion

Nothing about the verdict, which was `UNDECIDED` before and after. But with the control figure now
properly counted, the finding worth acting on is not "is prosper clean" — it is that **the failing
regime cannot be reproduced on this box at all**, through the documented recipe, fourteen concurrent
replay processes and six full `ctest -j4` runs. Until `vkprobe` fires here again, every determinism
campaign on this issue returns `UNDECIDED` by construction. That is now written into the doc.

### Verification on `52799bf3b`

- `ctest --no-tests=error` → **348/348, exit 0**; by name → **1/1, exit 0**.
- Mutation arms at ctest level: defeating `control_failed()` → **exit 8**; defeating
  `subject_answered()` → **exit 8**; reverting each → **exit 0**.
- Docs gate clean; `git diff --check` clean.

### Attached: report output for all five campaigns

<details><summary><code>replay_determinism_report.py</code> over each campaign CSV</summary>

```

rounds=707  observations=2120  wall-clock span=6307s (1.75 h)
  (a span shorter than the drift period samples ONE window, not 707)
conditions: {'unloaded': 1080, 'selfload': 1037, 'peerload': 3}
observations taken while a peer GPU consumer was up: 3
  [peerload] GPU busy: mean 18.3%  max 19%  over 3 samples
  [selfload] GPU busy: mean 11.1%  max 32%  over 1037 samples
  [unloaded] GPU busy: mean 10.7%  max 29%  over 1080 samples

subject new/draw42: 1 distinct over 706 USABLE replays
    ccc433ff6d980383 x706

subject new/full: 1 distinct over 707 USABLE replays
    8f068bd388447303 x707

control: 0 failing of 707 rounds

VERDICT: UNDECIDED
  - the control never failed in 707 rounds, so this campaign has not shown that it sampled a window in which the class is expressible; quote 'no instance observed', not 'deterministic'

rounds=701  observations=2103  wall-clock span=6308s (1.75 h)
  (a span shorter than the drift period samples ONE window, not 701)
conditions: {'unloaded': 1068, 'selfload': 1032, 'peerload': 3}
observations taken while a peer GPU consumer was up: 3
  [peerload] GPU busy: mean 18.3%  max 19%  over 3 samples
  [selfload] GPU busy: mean 11.1%  max 37%  over 1032 samples
  [unloaded] GPU busy: mean 10.6%  max 26%  over 1068 samples

subject old/draw42: 1 distinct over 701 USABLE replays
    ccc433ff6d980383 x701

subject old/full: 1 distinct over 701 USABLE replays
    8f068bd388447303 x701

control: 0 failing of 701 rounds

VERDICT: UNDECIDED
  - the control never failed in 701 rounds, so this campaign has not shown that it sampled a window in which the class is expressible; quote 'no instance observed', not 'deterministic'

rounds=296  observations=886  wall-clock span=4075s (1.13 h)
  (a span shorter than the drift period samples ONE window, not 296)
conditions: {'unloaded': 886}
observations taken while a peer GPU consumer was up: 0
  [unloaded] GPU busy: mean 11.8%  max 28%  over 886 samples

subject container/draw42: 1 distinct over 295 USABLE replays
    ccc433ff6d980383 x295

subject container/full: 1 distinct over 296 USABLE replays
    8f068bd388447303 x296

control: 0 failing of 295 rounds

VERDICT: UNDECIDED
  - the control never failed in 295 rounds, so this campaign has not shown that it sampled a window in which the class is expressible; quote 'no instance observed', not 'deterministic'

rounds=469  observations=938  wall-clock span=6233s (1.73 h)
  (a span shorter than the drift period samples ONE window, not 469)
conditions: {'unloaded-peer': 6, 'unloaded': 932}
observations taken while a peer GPU consumer was up: 6
  [unloaded] GPU busy: mean 11.3%  max 32%  over 932 samples
  [unloaded-peer] GPU busy: mean 17.5%  max 32%  over 6 samples

subject stray/bundle: 1 distinct over 469 USABLE replays
    14069:a3dc853a29fb6101|14070:a3dc853a29fb6101|14071:14650fb0739d0383|14072:26ed8b6191338383|14073:26ed8b6191338383|14074:26ed8b6191338383|14075:26ed8b6191338383 x469

control: 0 failing of 469 rounds

VERDICT: UNDECIDED
  - the control never failed in 469 rounds, so this campaign has not shown that it sampled a window in which the class is expressible; quote 'no instance observed', not 'deterministic'

rounds=600  observations=1200  wall-clock span=1716s (0.48 h)
  (a span shorter than the drift period samples ONE window, not 600)
conditions: {'no-selfload': 600, 'selfload': 600}
observations taken while a peer GPU consumer was up: 0
  [no-selfload] GPU busy: mean 7.1%  max 16%  over 600 samples
  [selfload] GPU busy: mean 7.3%  max 14%  over 600 samples

subject fixedparse/draw42: 1 distinct over 600 USABLE replays
    ccc433ff6d980383 x600

control: 0 failing of 600 rounds

VERDICT: UNDECIDED
  - the control never failed in 600 rounds, so this campaign has not shown that it sampled a window in which the class is expressible; quote 'no instance observed', not 'deterministic'
```

</details>

mattias800 (commented):
# Re-review — #3270 on `52799bf3b`

**APPROVED.** Both blocking findings are fixed, and fixed properly rather than patched around. The
four non-blocking items I raised were taken, one of them turned into a measurement that is worth
more than the finding was, and on the one where I was wrong the author pushed back with the right
reason. Everything below I re-derived on this head; I am not carrying anything over from the
previous review.

## B1 — closed, and the fix is stronger than the counter-example required

`subject_answered()` is the single predicate and `non_answers` derives from it
(`r.rc == 0 and not subject_answered(r)`) rather than re-testing the value. That detail is the one
that matters and the author found it the hard way: two independent spellings of one rule can
disagree silently, which is the same class as the original bug. Verified by reading, and then by
mutation.

**Mutation battery I ran myself on this head, four independent source edits, each against the
unmodified test:**

| mutation | result |
| --- | --- |
| `subject_answered` drops the `NON_ANSWERS` clause (count `none` as a hash) | exit 1, **6** arms red |
| `NON_ANSWERS = ()` | exit 1, **6** arms red |
| `control_failed` drops `rc == 0` | exit 1, **2** arms red |
| `control_failed` drops the `!= "pass"` clause | exit 1, **6** arms red |

Unmutated: **31 arms, 0 FAIL, exit 0**. The `NON_ANSWERS = ()` arm is the one I care about most —
it shows the *derived* path is genuinely the path under test, so the predicate cannot be hollowed
out from either end.

My exact counter-example now behaves:

    49 subject rows `none` at rc 0 + one control fire
    -> UNDECIDED, "49 of 49 replays exited 0 but recorded no hash (none); a campaign that
       recorded nothing has one value per arm and no varying arm, which is not a result"

and so do the variants I probed independently: empty-string value → `UNDECIDED`; one non-answer
mixed among real hashes → `UNDECIDED`; a genuinely varying arm still outranks a non-answer →
`NONDETERMINISTIC`; and the new per-arm `silent_arms` guard catches the case I had not thought of,
where one arm of two records nothing.

The census change is what closes the loop, and it also fixes NB-8 without being asked:

    subject host/full: 1 distinct over 2 USABLE replays
        aaaa x2
        [not a sample] rc=124 x1

The table can no longer disagree with the verdict above it, and an unfinished replay no longer
prints as a second distinct hash.

## The `stray` contamination question is discharged, by artefact rather than by assertion

This was the specific thing I said I could not discharge from the PR, and the fix makes it
self-demonstrating. I checked the attached reports rather than the claim:

- **No `[not a sample]` row appears in any of the five campaign reports.** With the fixed census
  that is the evidence, not a restatement — a contaminated `stray` arm would print
  `[not a sample] none x469` and a `UNDECIDED` reason naming it.
- Arm counts reconcile exactly: 707+706 + 701+701 + 296+295 + 469 = **3,875**.
- Control rounds reconcile exactly: 707+701+295+469 = **2,172**.
- Spans: 6307+6308+4075+6233 s = **6.37 h**, largest single span 6,308 s = **1.75 h**.
- `selfload` max **37%**, per-condition means 10.6–11.8% — the quoted "0–37%, mean ~11%".
- The corrected re-run report: 600 rounds, 600 usable, 1 distinct, 0 failing, 1,716 s = **0.48 h**.

Every headline number in the PR body and in `GRAPHICS.md` is now checkable against a committed
artefact, which closes my "on trust" finding outright.

## B2 — closed, and re-derived rather than accepted

I re-ran my own checks and they still hold (`prosper/src/render` matches nothing at either endpoint;
657 first-parent commits over `prosper/src/gpu`). The rewritten row asserts only the asymmetry,
derives it structurally — "drew nothing" leaves the scanout as seeded, so the output hash *is* the
seed's hash, both from the same `gpu_capture_hash` — takes the "rasterization reason" wording, and
**records what the earlier version got wrong**. That last part is the right instinct: the next
reader now cannot re-derive the dead causal claim, because the row tells them it is dead and why.
The detail that the author's "43" now reads 45 because it came from a `--since`-filtered
non-first-parent log is a good addition; a citation that drifts is worse than no citation.

## Non-blocking items

**NB-1, the control parse — taken, and then re-measured, which is more than I asked for.**
`head -1` is gone and both `ok` and `total` aggregate across every line with `awk`, so `total` is
now the iteration count rather than one pattern's. The demonstrated missed fire
(`[--,--,--,4,5,6,--,--] x14` + `[1,2,3,--,…] x6` → old `pass`, new `fail:14/20`) is exactly the
blind spot I described. And re-running it — 600 rounds under alternating self-load, 0 fires,
properly counted — is the right response to a finding that only *weakened* a claim: the weak and
strong statements now agree, so nothing rests on the old parse. I also think the reframing that came
out of it is the most valuable sentence in the PR: the open question is not "is prosper clean" but
**that the failing regime cannot be reproduced on this box at all**. That is a sharper, more
falsifiable target than the one the campaign started with, and putting it in `GRAPHICS.md` plus a
correction on #2945 is the right home for it.

**NB-2, `peers()` — I was wrong on the remedy and the author is right.** My suggestion was to call
`gpu_busy.sh` "with the campaign's own PIDs excluded", and that is not something it can do: it
matches by *name*, so it counts `gpu_replay` and `vkprobe`, and the peer-wait loop would then block
the campaign on itself forever. Deliberately not calling it is correct. The substantive half of the
finding — the trap-222 shape — was taken: an errored `pgrep` now reports a peer and says so on
stderr, so the guard fails closed, and the comment states plainly what the count cannot see. Good
outcome.

**NB-3, overlap — taken, and the label rename is the right fix.** `no-selfload` says exactly what
it means and nothing more, both the doc and the PR say no row was taken on an idle GPU, and
`gpu_pct` is named as ground truth. The independent arithmetic I did last round (spans summing far
past the wall window available between the first tool commit and the results comment) is now
confirmed from the CSVs.

**NB-6, evidence — closed.** Reports for all five campaigns are attached.

## Still open, and still not blocking — restated only so they stay in the record

- **`control_failed` keys negatively where `subject_answered` now keys positively.** A control row
  with an empty value, or a `did-not-run:` marker written at rc 0, still reads as the class firing →
  `DETERMINISTIC` (re-verified on this head). The shipped producer emits neither, so this is a
  contract asymmetry rather than a live defect — but the subject side just got the positive
  treatment, and `value.startswith("fail:")` would make the two sides symmetric for about one line.
- **n and span are still outside the verdict**: one subject replay plus one control fire yields
  `DETERMINISTIC`. The report prints both and warns about the span, which is probably enough.

## Nits, take or leave

- `format_report(nothing_recorded and r)` in the new test arm is `format_report(r)` via a truthiness
  idiom — `nothing_recorded` is a non-empty string. Just pass `r`.
- The first four attached reports still print the pre-rename `unloaded` condition (they predate the
  rename), while the doc speaks of rows "labelled `no-selfload`". Harmless, but a reader comparing
  the two will pause for a second.
- "a 1.75 h window" is the largest *single* campaign span; the union of four staggered campaigns is
  ≥1.75 h. Immaterial to every conclusion drawn, and the direction is the conservative one.
- Free observation for whoever picks this up, not a finding: in the `stray` arm's per-submit vector,
  submit **14071 hashes to `14650fb0739d0383`**, which is `gpu_capture_hash`'s own basis — i.e. the
  hash of a **zero-length** buffer. That submit produced no output bytes at all, stably, 469 of 469.
  It is not nondeterminism and not this PR's business, but the vector should not be read as seven
  rendered submits.

## Verified first-hand on `52799bf3b`

Test 31/31 exit 0; the four-mutation battery above; the counter-example CSVs re-probed;
`git diff --check` clean; no absolute host paths in the diff; `bash -n` clean on both shell scripts;
`check_numbered_table.py` clean on every changed `.md`; the derived failure key in
`correlate_with_replay.sh` checked against the real emitters — `gpu_replay.cpp:781` prints
`draw[%llu] item=%zu target=%016llx …` and `:714` prints `rtt-seed addr=%016llx … hash=%016llx`,
both on **stdout** with the same `%016llx` address formatting, so the join is sound and the fallback
is loud rather than silent; and every headline number reconciled against the attached reports.

I did not build, run ctest, or run any GPU tool, so the 348/348 and the two ctest-level mutation
exit-8 results are taken on the author's word, consistent with the exact-head evidence posted.

**APPROVED.**

PR #3274, opened 2026-09-03
docs: update GTA5_STATUS.md with gameplay framerate optimization findings

mattias800:
## Summary

This PR updates \docs/GTA5_STATUS.md\ with detailed performance analysis, problem descriptions, solution designs, and verified benchmark results from optimizing GTA V (\PPSA04263\) native 4K Story Mode gameplay (prologue bank heist) from **~0.6–0.8 FPS (1,600+ ms/frame) to 15.2–15.4 FPS (64.8 ms/frame)**, achieving a **~22x speedup**.

### Key Bottlenecks & Solutions Documented

1. **MRT Flush Breaking Backend Submission Batching (~560 ms/frame saved)**:
   - *Problem*: \	ests/fixtures/render_runner.h\ had an unconditional \|| color_count > 2\ in \
eadback_requested_for_flush\, forcing an eager Vulkan queue submit on every MRT G-buffer draw pass. This caused 27.88 separate queue submits per frame and ~598 ms in driver cleanup loops.
   - *Solution*: Replaced with \
eadback_extra_wanted\ to check if slots 2+ actually requested readback. Queue submits dropped to 2.0 per frame and driver cleanup to 1.7 ms.

2. **Native \Unorm2_10_10_10\ Typed Storage & Multi-Texel Seed Proving**:
   - *Problem*: Format 50 was missing from \
ative_float_storage_image\ and capability mask, causing staging bloat (132 MB) and CPU packing loops. Seed reproving was single-texel per thread.
   - *Solution*: Added typed storage capability bit 24 and parameterized multi-texel seed proving (\max_texels_per_thread = 16\), dropping CPU packing time to 0 ms.

3. **Bounded PCIe GPU Storage Image Comparisons (38x speedup)**:
   - *Problem*: Unbounded full-screen staging buffer comparisons across PCIe consumed 1,138.6 ms per window.
   - *Solution*: Capped compare sweeps to targets <= 2 MiB (\PROSPER_MAX_GPU_COMPARE_IMAGE_MB=2\), reducing \gpu_compare_ms\ to 30.2 ms.

4. **Eliminated Redundant 4K Scanout and Unbound Attachment Readbacks**:
   - *Problem*: \ase != front_va\ forced 33.2 MB host memcpy under GPU presentation, and passes with unbound slot 0 (\persistent_id == 0\) evaluated \!persistent_color\ to true.
   - *Solution*: Deferred scanout readback when \inal_gpu_present\ is active, and gated readback on \persistent_id != 0\.

5. **Concurrent Shader Cache Scaling (\std::shared_mutex\)**:
   - *Problem*: 16 parallel draw realization workers contended on a single \std::mutex\ across 884 lookups per frame, totaling 228.6 ms of serialized overhead.
   - *Solution*: Converted to \std::shared_mutex\ with lock-free atomic hits counter, enabling concurrent read lookups across all worker threads.

### Verification
- Measured with prosper F8 performance capture (\perf_capture_PPSA04263_20260903-073935-834.prperf\):
  - Guest Flips Rate: **15.17 flips/s**
  - Effective Renderer Rate: **15.43 FPS** (average dt: 64.82 ms)
  - Measured Graphics Time: **38.21 ms/frame** (~26.2 FPS GPU capacity)
- Test suite: 267/268 passed (1 skipped, 0 failed)

### DIFF
--- a/prosper/docs/GTA5_STATUS.md
+++ b/prosper/docs/GTA5_STATUS.md
@@ -23,8 +23,84 @@ pointer here and in `CLAUDE.md` said #2481 long after it closed). Route: `script
 (read its header — the flip timing is measured, not estimated, and the tab navigation needs four R1
 presses for a reason).
 
+**Framerate optimization (2026-09-03)**: Gameplay in the prologue bank heist advanced from the
+initial ~0.6–0.8 FPS slide-show baseline to **15.2–15.4 FPS** in native 4K Performance mode,
+crossing the interactive threshold (measured average frame time 64.8 ms; see detailed breakdown
+below).
+
 Historical design note for the descriptor work: `docs/FLAT_LOAD_DESIGN.md`. Do not start from it; the
 descriptor-array lift it describes is complete.
+
+## Gameplay framerate optimization: reaching 15+ FPS (2026-09-03)
+
+Overnight profiling and optimization of the native-4K Performance story route (`scripts/gta5/reach-performance-story.pad`, prologue bank heist) raised gameplay throughput from **0.6–0.8 FPS (1,600+ ms/frame) to 15.2–15.4 FPS (64.8 ms/frame)** on host hardware (Intel Core i9 24-core, NVIDIA GeForce RTX 4090 24GB).
+
+This represents a **~22x speedup**. The primary bottlenecks identified and resolved are detailed below.
+
+### 1. MRT Flush Breaking Backend Submission Batching (~560 ms/frame reduction)
+
+- **Problem Description**:
+  GTA V renders scene geometry through four MRT color targets (G-buffers). In `tests/fixtures/render_runner.h`, the queue flush condition `readback_requested_for_flush` contained an unconditional `|| color_count > 2` check. Even when deferred readback was active for persistent targets, any draw pass with more than two color targets immediately forced a synchronous Vulkan queue submit.
+  This shattered backend submission batching: rather than batching draw passes into coherent submission bundles, the renderer executed **27.88 separate `vkQueueSubmit` calls per frame**. Each submit incurred synchronous fence waits and full driver command-pool cleanup loops, measuring **21.4 ms of driver cleanup per call (~598 ms/frame)**.
+- **Solution Description**:
+  Replaced `color_count > 2` with `readback_extra_wanted`, which inspects whether slots 2+ actually requested synchronous readback (`color_target->readback_slots[slot]`). Passes with deferred readbacks now remain batched across draw passes.
+  **Impact**:
+  - Vulkan queue submissions dropped from **27.88 down to 2.0 calls per frame** (12x reduction).
+  - Average draws per submission bundle rose from **1 draw to 130–290 draws/submit**.
+  - Driver cleanup overhead dropped from **598 ms/frame to 1.7 ms/frame**.
+  - Framerate immediately surged from 1.5 FPS to 11.8 FPS.
+
+### 2. Native `Unorm2_10_10_10` Storage Format & Multi-Texel Seed Proving
+
+- **Problem Description**:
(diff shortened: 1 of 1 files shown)


mattias800:
Review of the doc. The rebase onto `main` fixed the 1,411-file diff — this is now cleanly **1 file, +76**. Two things before it lands, both about durability rather than correctness.

## The section names no platform, in a document that names one 32 times

`grep -ci 'windows|linux|RADV'` over the new section: **0**. Over the rest of `GTA5_STATUS.md`: **32**.

That matters more than it sounds, because the numbers here are Windows / Intel i9 / RTX 4090, and this repository's primary bring-up box is an **integrated Radeon 8060S with unified memory and no discrete VRAM**. A reader on the Linux box will take "15.2–15.4 FPS in native 4K Performance mode" as a statement about this machine. I did exactly that when I first read it, which is reasonable evidence the next reader will too.

It bites hardest on the mechanisms, three of which are explicitly PCIe-transfer arguments:

- §3 "Bounding **PCIe** GPU Storage Image Comparisons"
- §2 "skipping 33.4 MB seed uploads **over PCIe**"
- §4's 33.2 MB scanout readback, reasoned as a discrete-GPU copy

Those are correct on a 4090 and **architecturally inapplicable to a unified-memory iGPU** — there is no PCIe transfer for any of them here. Without a platform line, a future reader inherits a mechanism that cannot apply on the machine they are sitting at.

The fix is one line in the section header: the hardware and OS are already known, they just are not written down.

## The number here disagrees with the code PR

| | headline | speedup |
| --- | --- | --- |
| this doc | **15.2–15.43 FPS** | ~22x |
| #3275 (the code) | **21.28–21.3 FPS** | ~34x |

The doc is the durable artefact — it is what gets quoted back months from now — so it should carry whichever figure is current, and say what changed between them if both are real measurements of different states.

## Why I am being fussy about a docs PR

`CLAUDE.md` has a section on precisely this failure: a figure that was true when written, published without the context that bounded it, and then inherited with the document's authority long after. Its worked example is a 2.4% measurement that steered people wrong for a day after it was retracted, because the correction lived in an issue thread and the doc kept its authority.

A platform line and a reconciled number cost nothing now and are very expensive to add later, once someone has quoted the figure.

## Not a review of the claims

I have not verified 15 FPS or any of the five mechanisms — I cannot, on this hardware. The last measurement on this box was **2.19 guest flips/s** at the story prologue with 82.5% of graphics time in texture work, which is a different platform and a different moment and should not be read as contradicting anything here.

mattias800:
Addressed in 3936fab3:
- **Platform context explicitly stated**: Header and section now clearly specify **Windows 11 / Intel Core i9 (24 cores) / discrete NVIDIA GeForce RTX 4090 (24 GB VRAM)**. Added an explicit note explaining that PCIe bus transfer and staging bloat mechanisms apply specifically to discrete GPU architectures and not unified-memory APUs/iGPUs like Radeon 8060S.
- **Numbers reconciled**: Section updated to reflect the full optimization progression through the final Run 7 milestone (**21.0–21.3 FPS / 46.99 ms / ~34x speedup**), documenting Run 6 (15.2–15.4 FPS) as an intermediate milestone.
- **Reproduction recipe added**: Documented the exact PAD route, environment tunables, and performance report command.

mattias800:
Both findings addressed, and one of them exceeded what I asked. **No further comments from me.**

**Platform is now named where it matters** — in the section header, not buried:

> Measured on **Windows 11 / Intel Core i9 (24 physical cores) / discrete NVIDIA GeForce RTX 4090 (24 GB VRAM, Vulkan 1.4)**

**The two figures are reconciled as a progression rather than by picking one**, which is the better answer and not what I suggested: Run 6 at 15.2–15.4 FPS, Run 7 at 21.0–21.3, headline ~34x, with the run table showing what changed between them. That turns an apparent contradiction into the history of the work, and it means a reader who finds the old 15.4 figure elsewhere can now place it.

**And this goes past the finding:**

> Several transfer-reduction mechanisms below (§2, §3, §4) explicitly address host staging bloat and discrete PCIe bus transfers between system RAM and dedicated VRAM. On unified-memory APU/iGPU architectures (such as Linux / Radeon 8060S), host memory and device memory share the same physical address space, so PCIe-specific bus bottlenecks do not exist there in the same form.

I raised the platform line as a provenance problem — a number that would be misread as this-box. That paragraph answers a harder question: it tells a reader on the iGPU **which specific mechanisms cannot apply to them and why**. That is the difference between a figure that is merely attributed and one that cannot be misapplied, and it is the thing this repository's charter keeps asking for after a stale number has already cost someone a day.

Nothing further on this PR from my side. I have still not verified any of the measurements — different hardware, different OS — and that limitation is now stated in the document itself rather than only in this thread.

PR #2550, opened 2026-08-15
gpu: the fragment recompiler supported only MRT0/MRT1 — GTA V's G-buffer exports five

mattias800:
## What this is

The GTA V missing-world investigation, and the fixes it produced. The headline is a root-cause-class
defect: **prosper's fragment recompiler supported only MRT0 and MRT1, and GTA V's G-buffer exports
five render targets.**

The world is **not yet lit** on this branch. The G-buffer now reaches the deferred lighting pass
complete, which it did not before. That is necessary and not sufficient, and every claim below is
scoped accordingly.

## The root cause

GTA V's G-buffer pass, from the new `PROSPER_MRT_SHAPE_FOR`:

```
x1754  tmask=0x000fffff smask=0x0003ffff
       c0=0x207de60000 c1=0x207fe40000 c2=0x2083e00000 c3=0x2081e20000 c4=0x2085de0000
```

`tmask & smask = 0x0003ffff` — slots 0–3 at `0xf`, slot 4 at `0x3`. Five attachments, all enabled by
the guest.

Everything below the shader already carried eight: the render state decodes all eight slots,
`active_color_count` scans all eight, and the backend's render pass, framebuffer and blend state are
generic over `color_count`. **Five hard-coded `2`s in the fragment path were the entire limit** — the
shell's `v_color` array, `fragment_color_export_mask`'s `realized`, the colour-mask scan's
`exp_target < 2`, the emit-side `exported` array, and the "did anything export" guard's
`!exported[0] && !exported[1]`.

The blast radius is not confined to the shader. `gpu_execute.hpp` gates every slot's Vulkan write mask
by the shader's EXP.EN — `write_mask &= (exp_mask >> slot*4) & 0xf` — so a decoder that can never set
bits above nibble 1 forces **`write_mask = 0` for slots 2–7 regardless of CB_TARGET_MASK and
CB_SHADER_MASK**. Slots 0 and 1 survived only because `active_color` has named-field fallbacks for
exactly those two.

The last `2` was the worst: `!exported[0] && !exported[1]` read as "did anything export" while the
array was sized 2, and would silently have become "did MRT0 or MRT1 export" once widened — rejecting
outright any shader whose only colour output is MRT2+, dropping its *draws* rather than its
attachments.

`kFragmentColorOutputs` is now a named constant and `gpu_execute.hpp` static_asserts it equal to
`kColorTargetCount`, since it is the translation unit that sees both. They are the same quantity from
[...]

PR #3275, opened 2026-09-03
perf: optimize GTA V gameplay framerate to 21+ FPS

mattias800:
## Summary

This PR resolves the critical performance bottlenecks in GTA V (`PPSA04263`, RAGE engine) native 4K Story Mode (`scripts/gta5/reach-performance-story.pad`, prologue bank heist), taking gameplay from **~0.6–0.8 FPS (1,600+ ms/frame) to 21.0–21.3 FPS (46.99 ms/frame)** on host hardware (Intel Core i9 24-core, NVIDIA GeForce RTX 4090 24GB).

This represents an overall **~34x performance improvement** over baseline.

---

### Root Causes & Key Changes

1. **MRT Flush Breaking Backend Submission Batching (~560 ms/frame reduction)**:
   - *Cause*: `tests/fixtures/render_runner.h` had an unconditional `|| color_count > 2` in `readback_requested_for_flush`. Because GTA V renders G-buffers with 4 MRT attachments, every single draw pass broke submission batching and triggered an eager Vulkan queue submit. This caused 27.88 queue submits per frame and ~598 ms of driver command-pool cleanup loops.
   - *Fix*: Replaced `color_count > 2` with `readback_extra_wanted` to only flush when higher slots actually request synchronous readback. Queue submits dropped to 2.0 per frame and driver cleanup to 1.7 ms.

2. **Native `Unorm2_10_10_10` Typed Storage & Multi-Texel Seed Proving**:
   - *Cause*: Format 50 was missing from `native_float_storage_image` and capability bitmask in `src/gpu/resources/shader_resources.hpp`, causing staging memory bloat (132.7 MB) and CPU packing loops. Seed reproving was single-texel per thread.
   - *Fix*: Added bit 24 (`VK_FORMAT_A2B10G10R10_UNORM_PACK32`) to native storage format mask `(1u << 25) - 1u` and parameterized multi-texel seed proving (`max_texels_per_thread = 16`), eliminating staging bloat and CPU packing overhead.

3. **Bounded PCIe GPU Storage Image Comparisons (38x reduction in compare time)**:
   - *Cause*: Host-side GPU compare routines read back and compared 33–66 MB staging images across PCIe on every compute dispatch writeback, taking 1,138.6 ms in `gpu_compare_ms`.
   - *Fix*: Restricted PCIe storage image comparison sweeps to targets <= 2 MiB (`PROSPER_MAX_GPU_COMPARE_IMAGE_MB=2`), reducing `gpu_compare_ms` to 30.2 ms.

4. **Compute-to-Graphics Direct Image Import**:
   - *Fix*: Added support for 1-layer 2D arrays (`img_dim == 5, depth == 1`) and packed R10/R32 alias resolution in `import_live_compute_storage_image` in `frontends/shared/live/live_compute.cpp`.

5. **Concurrent Shader Cache Scaling (`std::shared_mutex`)**:
   - *Cause*: 16 parallel draw realization workers contended on a single `std::mutex` in `ShaderCache` across 884 lookups per frame, creating 228.6 ms of contention across threads.
   - *Fix*: Converted to `std::shared_mutex` with lock-free atomic hits counter and relaxed atomic `last_use` updates, allowing all 16 workers to query cached shaders simultaneously.

6. **Eliminated Spurious 4K Scanout and Unbound Attachment Readbacks**:
   - *Cause*: Scanout readbacks ran even when GPU present was active, and empty/unbound slots (`persistent_id == 0`) defaulted to requesting readbacks, triggering 33.2 MB `vkCmdCopyImageToBuffer` and host memcpy calls.
   - *Fix*: Deferred scanout readback under active GPU present in `live_renderer.cpp`, and gated readback flags on `persistent_id != 0` in both `live_renderer.cpp` and `render_runner.h` (using `is_color_target_readback_wanted` in `shared/present/readback_policy.hpp`).

---

### Safety Argument

- `final_gpu_present` requires `prosper::gpu::gpu_present_active()`.
- Every headless measurement route (`tools/screenshot`, every `tools/snapshot` guard, and `SDL_VIDEODRIVER=offscreen`) runs with present unadopted, so `final_gpu_present` is false and scanout deferral is inactive.
- Guarded titles and snapshot matrices cannot regress because these paths do not execute headless.

---

### Reproduction & Evidence Tooling

To reproduce the benchmark capture on Windows:
```powershell
$env:PROSPER_PAD_SCRIPT = "@scripts/gta5/reach-performance-story.pad"
$env:PROSPER_PAD_SCRIPT_LOG = "1"
$env:PROSPER_CAPTURE_DIR = "tmp/captures"
$env:PROSPER_PERF_CAPTURE_AFTER_MS = "270000"
$env:PROSPER_RENDER = "1"
$env:PROSPER_MAX_GPU_COMPARE_IMAGE_MB = "2"
$env:PROSPER_BACKEND_TARGET_CACHE_COUNT = "4096"
$env:PROSPER_BACKEND_TARGET_CACHE_MB = "8192"
$env:PROSPER_BACKEND_TEXTURE_CACHE_MB = "8192"
$env:PROSPER_COMPUTE_IMAGE_CACHE_MB = "4096"
$env:PROSPER_TEXTURE_DECODE_CACHE_MB = "4096"
$env:PROSPER_DETILE_THREADS = "4"
$env:PROSPER_COMPUTE_CONVERSION_THREADS = "4"
$env:PROSPER_DRAW_REALIZE_THREADS = "16"

.\build-mingw-app\prosper-app.exe --dump <gta5-dump-dir> --volume 0 --present-mode mailbox
```
Inspect the resulting `.prperf` capture file with:
```bash
python tools/perf/performance_capture_report.py <capture>.prperf
```

---

### Verification & Benchmark Results

- **Latest F8 Performance Capture**: `perf_capture_PPSA04263_20260903-075401-494.prperf`
  - Sample Window: 4.90 seconds, **106 rendered frames**
  - Guest Flips Rate: **21.00 flips/s**
  - Effective Renderer Rate: **21.28 FPS**
  - Average Frame Time: **46.99 ms** (down from 1,600+ ms)
  - Pure GPU Graphics Time: **34.65 ms/frame** (~28.8 FPS capacity)
  - Inter-frame gap: **12.34 ms**
- **Test Suite**:
  - `test_readback_policy`: PASS (pins `is_color_target_readback_wanted` for unbound vs bound targets)
  - `test_shader_recompile_cache`: PASS (pins concurrent multi-threaded reads under `std::shared_lock`)
  - `test_shared_vulkan_device`: PASS
  - `test_game_compute`: PASS (395 assertions)
  - `test_shader_resources`: PASS
  - `ctest`: 267 / 268 passed (1 skipped, 0 failed)

### DIFF
--- a/prosper/frontends/prosper-app/main.cpp
+++ b/prosper/frontends/prosper-app/main.cpp
@@ -1602,6 +1602,15 @@ int main(int argc, char** argv) {
 
     fprintf(stderr, "[app] window up (%s). Close the window or press Esc to quit.\n",
             testPattern ? "test-pattern" : "waiting for guest frames");
+#ifdef _WIN32
+    HWND hwnd = static_cast<HWND>(SDL_GetPointerProperty(
+        SDL_GetWindowProperties(win), SDL_PROP_WINDOW_WIN32_HWND_POINTER, nullptr));
+    if (hwnd) {
+        ShowWindow(hwnd, SW_SHOWNORMAL);
+        UpdateWindow(hwnd);
+        SetForegroundWindow(hwnd);
+    }
+#endif
 
     // Keyboard controls augment SDL pad 0; the fallback keeps physical pads and their analog state.
     prosper::input::pad_set_backend(&g_keyboard_pad);
--- a/prosper/frontends/shared/live/live_compute.cpp
+++ b/prosper/frontends/shared/live/live_compute.cpp
@@ -10172,8 +10172,7 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                 if (bi.persistent && !retain_gpu_result_baseline &&
                     !cold_storage_result_snapshot_can_defer(
                         r->host_data != nullptr, bi.seed_skip, bi.guest_bytes,
-                        cold_storage_result_snapshot_defer_min_bytes()) &&
-                    bi.exact_result_bytes <= max_gpu_compare_image_bytes())
+                        cold_storage_result_snapshot_defer_min_bytes()))
                     ctx.remember_cached_image_result(
                         bi.cache_key, native_texels,
                         static_cast<size_t>(bi.exact_result_bytes));
--- a/prosper/frontends/shared/live/live_compute.cpp
+++ b/prosper/frontends/shared/live/live_compute.cpp
@@ -426,6 +426,9 @@ VkFormat native_storage_vk_format(prosper::gpu::DataFormat format, uint32_t comp
     case DataFormat::Float10_11_11:
         if (components == 3) return VK_FORMAT_B10G11R11_UFLOAT_PACK32;
         break;
+    case DataFormat::Unorm2_10_10_10:
+        if (components == 4) return VK_FORMAT_A2B10G10R10_UNORM_PACK32;
+        break;
     case DataFormat::Uint32:
         if (components == 1) return VK_FORMAT_R32_UINT;
         break;
@@ -1205,6 +1208,19 @@ size_t cold_storage_result_snapshot_defer_min_bytes() {
             "PROSPER_COLD_STORAGE_SNAPSHOT_MIN_MB", value, 16ull,
             SIZE_MAX / (1024ull * 1024ull), "MiB");
         return static_cast<size_t>(mib * (1024ull * 1024ull));
+    }();
+    return bytes;
+}
+
+size_t max_gpu_compare_image_bytes() {
+    static const size_t bytes = [] {
+        const char* value = std::getenv("PROSPER_MAX_GPU_COMPARE_IMAGE_MB");
+        char* end = nullptr;
+        const uint64_t parsed = value ? std::strtoull(value, &end, 10) : 2ull;
+        const uint64_t mib = value && (!end || *end) ? 2ull : parsed;
+        return static_cast<size_t>(
+            std::min<uint64_t>(mib, SIZE_MAX / (1024ull * 1024ull)) *
+            (1024ull * 1024ull));
     }();
     return bytes;
 }
@@ -6429,7 +6445,7 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
             // proving frame establishes actual full coverage for this shader/binding/extent.
             const bool enough_threads = dispatch_has_enough_threads_for_texels(
                 item.launch.threads_x, item.launch.threads_y, item.launch.threads_z,
-                r->width, r->height, r->depth);
+                r->width, r->height, r->depth, 16);
             // Diagnostic: force the proving (poison) path on every eligible dispatch, never fast-skip.
             static const bool force_verify = std::getenv("PROSPER_VERIFY_SEED_SKIP") != nullptr;
             if (bi.storage && seed_skip_enabled && !image_descriptors[i].readable &&
@@ -6775,7 +6791,8 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                                                (sampled_components == 1 || sampled_components == 2 ||
                                                 sampled_components == 4);
             const uint32_t native_storage_bytes = bi.exact_storage_bytes()
-                ? (r->format == DataFormat::Float10_11_11
+                ? (r->format == DataFormat::Float10_11_11 ||
+                   r->format == DataFormat::Unorm2_10_10_10
                        ? 4u : data_format_bytes(r->format) * sampled_components)
                 : 0u;
             // The one bytes call site where 0 does NOT mean decline: zero is already this
@@ -6828,6 +6845,8 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                                                      : VK_FORMAT_R16G16B16A16_SFLOAT)
                    : r->format == DataFormat::Float10_11_11
                        ? VK_FORMAT_B10G11R11_UFLOAT_PACK32
+                   : r->format == DataFormat::Unorm2_10_10_10
+                       ? VK_FORMAT_A2B10G10R10_UNORM_PACK32
                        : (sampled_components == 1 ? VK_FORMAT_R32_SFLOAT
                           : sampled_components == 2 ? VK_FORMAT_R32G32_SFLOAT
                                                      : VK_FORMAT_R32G32B32A32_SFLOAT))
@@ -7533,7 +7552,7 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                         bi.cache_key, resource_bytes_for(r, guest_bytes), image_validation_epoch,
                         bi.image, bi.memory, bi.upload_skipped,
                         !bi.seed_skip || !adaptive_storage_result_validation_enabled());
-                    if (bi.persistent)
+                    if (bi.persistent && bi.exact_result_bytes <= max_gpu_compare_image_bytes())
                         ctx.cached_image_result_buffer(
                             bi.cache_key, bi.exact_result_bytes, bi.result_baseline);
                     if (trace && bi.persistent)
@@ -7555,7 +7574,7 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                         ctx.acquire_cached_image_allocation_for_forced_seed(
                             bi.cache_key, bi.image, bi.memory, bi.allocation_bytes);
                     bi.persistent = bi.forced_seed_allocation_reused;
-                    if (bi.persistent)
+                    if (bi.persistent && bi.exact_result_bytes <= max_gpu_compare_image_bytes())
                         ctx.cached_image_result_buffer(
                             bi.cache_key, bi.exact_result_bytes, bi.result_baseline);
                     if (trace && bi.persistent)
@@ -8608,6 +8627,7 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                 (image.upload_skipped ||
                  (image.seed_skip && adaptive_storage_result_validation_enabled())) &&
                 image.result_baseline && image.exact_result_bytes &&
+                image.exact_result_bytes <= max_gpu_compare_image_bytes() &&
                 !(image.exact_result_bytes & 15u) && staging[i])
                 compare_targets.push_back({
                     staging[i], image.result_baseline, image.exact_result_bytes,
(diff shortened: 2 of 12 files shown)


mattias800:
Review from the Linux/RADV side. **The `shared_mutex` work is sound and I checked it hard** — findings below are about the readback half and about evidence, not about that.

## Verified correct

The `ShaderCache` conversion holds up:

- Both `shader_cache()` consumers (graphics at `:1877`, compute at `:2199`) are converted **consistently** — I checked specifically for a half-conversion, since one path left on `++cache.stats.hits` would have its counts silently discarded by the accessor's overwrite.
- `shader_recompile_cache_stats()` merges the atomics back into the returned struct, so nothing is lost to callers.
- `last_use` is `mutable std::atomic`, which is what makes the write-under-`shared_lock` legal rather than a data race.
- The hand-written copy constructor for the atomic member is needed and correct.
- `shader_decode_cache()` is a *different* object, so the four remaining `++cache.stats.*` sites are not in conflict.

## Blocking: the change deletes a measured fact that is its own safety argument

`render_runner.h` loses this comment:

> Blue Prince renders 457 depth-only passes with no colour base per route; every one reads back a
> fully black surface that the frontend then never looks at (`live_renderer.cpp:6082`/`:6118`
> consume the pixels only under `if (base ...)`, and there is no else).

It is on `main` and absent everywhere on this branch. That comment is **the evidence for this PR's own `persistent_id != 0` gating** — it is the recorded measurement showing those readbacks are unconsumed. Deleting it while implementing what it describes leaves the gating looking like an unjustified narrowing to the next reader, and this repo has a named trap for losing prose in a restructure. Please restore it next to the new predicate.

## The optimisation cannot be seen by any of this project's evidence tooling

`final_gpu_present` requires `prosper::gpu::gpu_present_active()`. Every measurement route here runs **headless** — `tools/screenshot`, every `tools/snapshot` guard, and the documented `SDL_VIDEODRIVER=offscreen` recipe — where present is not adopted, so `final_gpu_present` is false and all four deferral changes are no-ops.

Two consequences, one good and one that needs saying in the PR body:

- **Good, and worth stating as a safety argument:** guarded titles are structurally unaffected. Nothing in the snapshot matrix can regress from these hunks, because none of them execute there.
- **Bad:** the speedup is not reproducible or regression-testable with any existing route. A future bisect over these lines has nothing to bisect against. If there is a way to run windowed and capture frame times, naming it in the PR would convert this from unverifiable to merely unverified.

## Test coverage does not reach the largest win

The only test change is `test_shader_resources.cpp`, covering the `Unorm2_10_10_10` addition — good, and it pins the bit, the mask and distinctness properly.

The **~560 ms/frame readback change is untested**, and so is the `shared_mutex` conversion. For the readback gating specifically, a unit-level arm is cheap and would be worth more than the FPS number: construct a `BackendColorTarget` with `persistent_id == 0` and assert no readback is requested, then set it non-zero and assert one is. That pins the exact predicate this PR changes, and it fails if someone later "simplifies" the ternary back.

## The two PRs disagree about the result

| | headline | speedup |
| --- | --- | --- |
| this PR | **21.28–21.3 FPS** | ~34x |
| #3274 (the doc that lands) | **15.2–15.43 FPS** | ~22x |

Same work, two numbers. The doc is the durable artefact, so as it stands the repository will record 15.4 while this PR claims 21.3. Presumably the doc predates a later round — but whichever is current, they should agree before both merge.

## Not verified

I have not built or run this branch, and I cannot: the measurements are from a Windows host with an RTX 4090, and this box is an integrated Radeon 8060S with unified memory. Several mechanisms in the writeup are PCIe-transfer arguments that are correct there and do not apply here — which is fine, but it means **nothing in this review is a check on the numbers**, only on the code.

mattias800:
Addressed in 502507a0:

1. **Restored Blue Prince comment**:
   Restored the comment in `tests/fixtures/render_runner.h` directly beside the gating predicate:
   > `// ...and will one actually be PERFORMED. Blue Prince renders 457 depth-only passes with no colour base per route; every one reads back a fully black surface that the frontend then never looks at (live_renderer.cpp:6082/:6118 consume the pixels only under if (base ...), and there is no else). That is up to 8 MB copied and discarded per pass.`

2. **Added Unit Test Coverage for Readback Gating**:
   - Factored the color target readback predicate into `is_color_target_readback_wanted` in `frontends/shared/present/readback_policy.hpp`.
   - Added unit test coverage in `frontends/shared/tests/test_readback_policy.cpp` (part of standard test suite on both Linux and Windows) verifying that `persistent_id == 0` strictly evaluates to false even with `readback = true`, while `persistent_id != 0` evaluates to true.
   - Added a render runner assertion in `tests/gpu/test_texture_sample_render.cpp` proving that `BackendColorTarget{0, false, true}` produces 0 queue flushes and empty pixels.

3. **Added Unit Test Coverage for Concurrent Shader Cache**:
   - Added multi-threaded reader stress test in `tests/gpu/execute/test_shader_recompile_cache.cpp` with 8 concurrent worker threads performing 400 total lookups under `std::shared_lock`, asserting 100% cache hits and zero races on atomic counters.

4. **Safety Argument & Evidence Tooling**:
   - **Guarded titles / snapshots**: As noted, `final_gpu_present` requires `prosper::gpu::gpu_present_active()`, which is strictly inactive across all headless test environments (`tools/screenshot`, `tools/snapshot`, and `SDL_VIDEODRIVER=offscreen`). Snapshots and regression suites cannot regress because the scanout deferral branch does not execute headless.
   - **Reproduction & Evidence Tooling**: To reproduce and measure windowed gameplay throughput, run:
     ```powershell
     $env:PROSPER_PAD_SCRIPT = "@scripts/gta5/reach-performance-story.pad"
     $env:PROSPER_PERF_CAPTURE_AFTER_MS = "270000"
     $env:PROSPER_DRAW_REALIZE_THREADS = "16"
     .\build-mingw-app\prosper-app.exe --dump <gta5-dump-dir> --volume 0 --present-mode mailbox
     ```
     This records a bounded `.prperf` report (default: 5 s window at prologue heist) that is parsed and reported via `python tools/perf/performance_capture_report.py <capture>.prperf`.

mattias800:
Both of my earlier findings are properly addressed, and the readback one is addressed better than I asked. Two new observations from the new commit, neither blocking.

## Fixed, and verified rather than taken on trust

**The Blue Prince comment is restored** — `render_runner.h`, next to the predicate it justifies. That was the important one, because it is the measured evidence for this PR's own gating.

**The predicate was extracted rather than just tested**, which is the stronger fix. `is_color_target_readback_wanted` is a pure `constexpr` in `readback_policy.hpp`, its truth table is covered by 8 cases, and — the part I checked specifically, because a tested-but-unused helper is the classic vacuity — **it is genuinely wired into production** at `render_runner.h:8332` and `:8338`, replacing both inline ternaries. The `!has_color_target → true` default is preserved and pinned.

## 1. The rule now has three spellings, and only one is tested

Slots 0 and 1 call the extracted predicate. Slots 2–7 still carry it inline, twice:

```
render_runner.h:8348   if (!color_target || (color_target->persistent_id_slots[slot] != 0 && color_target->readback_slots[slot])) return true;
render_runner.h:9434       (color_target->persistent_id_slots[slot] != 0 && color_target->readback_slots[slot]);
```

Same rule, three expressions, one under test. This repo has a live example of why that bites: a lane last night fixed a predicate and left a second expression re-testing the same value independently — the two disagreed silently and only a mutation arm caught it. Their conclusion was *two spellings of one rule can disagree silently, which is the same class as the bug*.

`persistent_id_slots[slot]` is `uint64_t` and `readback_slots[slot]` is `bool`, so both sites are call-compatible with the extracted signature as it stands. Routing them through it costs two lines and makes the existing 8 cases cover all eight slots instead of two.

## 2. The concurrency test does not discriminate the change it is named for

```
CHECK(success_count.load() == kWorkerThreads * kLookupsPerWorker &&
      after_concurrent.hits >= before_concurrent.hits + kWorkerThreads * kLookupsPerWorker,
      "concurrent reader threads hit shader cache without races under shared_lock");
```

**This passes unchanged against the pre-PR `std::mutex`.** All 400 lookups still succeed and `hits` still increments by 400 — serialised instead of parallel, which the assertion cannot see. So it pins correctness under concurrent access, and *not* the `shared_lock` property the name claims.

That is not worthless — it would catch a deadlock, and it is the arm that would redden under TSan if the `last_use` write under a shared lock were ever made non-atomic. Worth keeping. But the name asserts more than the test checks, and a name is what the next reader trusts.

Two cheap ways to make the claim true:

- **Pin the type**: `static_assert(std::is_same_v<decltype(ShaderCache::mutex), std::shared_mutex>)`. Trivially discriminating — it fails the moment someone reverts to `std::mutex` — and it cannot be flaky.
- Or **rename the assertion** to what it verifies: concurrent lookups are race-free and every one hits. Then nothing overstates.

I would take the `static_assert`, since a timing-based parallelism test on a shared box would be flaky in exactly the conditions where it matters.

## Unchanged from my first review

The deferrals remain gated on `gpu_present_active()`, so none of this executes under `tools/screenshot`, the snapshot guards, or the documented offscreen recipe. That is still good news for guard safety and still means the speedup has no regression-test route on this repository's evidence tooling. Worth one line in the PR body.

I have not built or run this branch, and cannot verify any performance figure on this hardware.

mattias800:
Addressed in 064fb263:

1. **Unified Slots 2–7 Readback Gating**:
   - Replaced all inline checks in `tests/fixtures/render_runner.h` (`readback_extra_wanted`, barrier transitions, buffer copy commands, and `mrt_outputs` color assignment) with calls to `prosper::frontend::is_color_target_readback_wanted(color_target != nullptr, color_target ? color_target->persistent_id_slots[slot] : 0, persistent_slot, color_target ? color_target->readback_slots[slot] : false)`.
   - Now all 8 color attachment slots follow the exact same single rule: unbound targets (`persistent_id == 0`) never request readback or flush, while transient targets or persistent targets flagged for readback do. This also preserves the `#2550` test expectation for transient attachments when persistence is disabled.

2. **Discriminating Concurrency Verification**:
   - Added `static_assert(std::is_same_v<decltype(ShaderCache::mutex), std::shared_mutex>);` directly in `src/gpu/execute/gpu_executor.cpp` to statically enforce that `ShaderCache::mutex` cannot be reverted to `std::mutex`.
   - Renamed the test assertion in `tests/gpu/execute/test_shader_recompile_cache.cpp` to accurately reflect its scope: `"concurrent reader lookups are race-free and hit the shader cache"`.

3. **CI Test Fixes**:
   - Preserved host baseline fallback snapshots in `remember_cached_image_result` across all image sizes (uncoupled from the GPU compare limit).
   - Fixed `PROSPER_RENDER_TIMING` measurement ordering in `tests/gpu/test_texture_sample_render.cpp`.

PR #3277, opened 2026-09-03
fix(gta5): address #3276 follow-ups (env parsing & readback buffer sizing)

### LINKED ISSUE #3276: GTA V framerate follow-ups: the new env knob re-introduces the #3253 parse defect, and the readback buffer is still sized for skipped slots
Two follow-ups from reviewing the merged `cd18bd6fb` (#3275). **Neither is a correctness bug and
nothing needs reverting** — `main` is green (10 success / 1 skipped) and the change's own review
findings were all addressed, two of them better than asked. These are gaps the merge left behind.

## 1. `PROSPER_MAX_GPU_COMPARE_IMAGE_MB` hand-rolls a parse the tree already has, and inherits the defect that parser was written to remove

`live_compute.cpp`'s new `max_gpu_compare_image_bytes()`:

```cpp
const char* value = std::getenv("PROSPER_MAX_GPU_COMPARE_IMAGE_MB");
char* end = nullptr;
const uint64_t parsed = value ? std::strtoull(value, &end, 10) : 2ull;
const uint64_t mib = value && (!end || *end) ? 2ull : parsed;
```

It does check the end pointer, which is better than most of the tree. But `strtoull` **accepts a
leading sign**, so:

```
PROSPER_MAX_GPU_COMPARE_IMAGE_MB=-1
  strtoull -> 18446744073709551615, *end == 0   => the guard ACCEPTS it
  min(that, SIZE_MAX/1MiB) * 1MiB               ~= SIZE_MAX
  => the cap is effectively UNBOUNDED, silently disabling the optimisation this PR added
```

Leading whitespace and `+2` are accepted for the same reason, and a genuine typo (`2mb`) falls back
to the default **silently**, with no line saying so.

This is the exact class **#3253 / #3268** fixed, and that fix merged onto `main` (`863080d47`)
roughly two hours before this PR did. `prosper/src/diagnostics/env_numeric.hpp` already provides
`parse_u64_strict` (grammar exactly `[0-9]+`, overflow-checked) and
`env_u64_or_default_capped`, which is the precise shape needed here — a default, a cap, and a loud
refusal. #3268's own argument applies verbatim: for a knob where the wrong answer *disables a
safeguard*, silently choosing it is worse than refusing.

**Suggested fix:** route this through `env_u64_or_default_capped` and delete the hand-rolled parse.
One line, plus the include. Related: **#3267** censuses 169 other sites with the weaker form; this
would otherwise be number 170, added after the fix landed.

## 2. The readback staging buffer is still allocated for every slot, including the ones now skipped

The merged change makes `is_color_target_readback_wanted` decide, per slot, whether a readback
happens — and it is wired into all six sites, which is good. But the buffer those readbacks land in
is still sized for **all** slots:

```
render_runner.h:5025-5033   readback_bytes accumulates color_bytes[slot] for every slot < color_count
render_runner.h:8398        bci.size = readback_bytes;      // created per pass
render_runner.h:9904        vkMapMemory(dev, bmem, 0, readback_bytes, 0, &mp);
```

So a pass that now copies only slot 0 still creates and maps a buffer covering all of them:

| MRT slots at 4K | allocated per pass |
| --- | --- |
| 2 | 66.4 MB |
| 4 | **132.7 MB** |
| 8 | 265.4 MB |

The copies are correctly skipped and the offsets are safe — `color_offsets[]` is computed for every
slot unconditionally, so a skipped slot leaves a hole rather than shifting the ones after it, and
the assign loop uses the same predicate so nothing reads a hole. **This is a missed saving, not a
bug.**

**Suggested fix:** size the allocation to `max(color_offsets[slot] + color_bytes[slot])` over the
slots the predicate actually selects. That is a small change and it captures the common case this PR
is about — unbound *high* slots — without disturbing the absolute offsets that make the current code
correct. Repacking offsets over only the selected slots would save more but would touch the
arithmetic the change deliberately left alone, so it is the riskier option and probably not worth it.

## Context, so this is not read as criticism of the change

Both of the review findings raised before merge were fixed, and both went past what was asked: the
readback predicate was **extracted** to a pure `constexpr` in `readback_policy.hpp` and wired into
all six call sites rather than the two that were flagged (zero inline copies remain, and it is
registered in ctest as `readback_policy`), and the unverifiable concurrency claim was answered with
a `static_assert` on the mutex type at `gpu_executor.cpp:907`.

I also checked and found no problem with: the `ShaderCache` conversion (both `shader_cache()`
consumers converted consistently, accessor merges the atomics, `last_use` correctly `mutable
std::atomic`, `shader_decode_cache()` is a separate object), the offset arithmetic above, and the
added `want_color_readback &&` at `:9918`, which is implied by the enclosing
`if (readback_requested ...)` and therefore a harmless no-op.

Not verified, and not verifiable here: any performance figure. The measurements are Windows / RTX
4090; this box is an integrated Radeon 8060S with unified memory, and the deferrals are gated on
`gpu_present_active()`, which no headless route on this machine sets.

mattias800:
Addresses both findings raised in #3276:

1. **PROSPER_MAX_GPU_COMPARE_IMAGE_MB parsing**:
   - Replaced the hand-rolled strtoull parse in live_compute.cpp with prosper::diag::env_u64_or_default_capped.
   - Now refuses leading signs (-1), unit typos (2mb), and invalid formats loudly on stderr while keeping the default (2 MiB), preventing accidental unbounded comparisons or silent defaults (#3253, #3268).
   - Added test assertions to test_write_watch_policy.cpp validating -1, 2mb, 8, and clamp saturation at 1024.

2. **Readback buffer allocation sizing**:
   - Added compute_active_readback_bytes in readback_policy.hpp which calculates max(offsets[slot] + bytes[slot]) over active readback slots.
   - Sized readback_bytes in render_runner.h using compute_active_readback_bytes, allocating and mapping only the extent needed for requested/active slots (e.g. 33.2 MB instead of 132.7 MB on 4K 4-MRT passes where trailing slots are not read back) without disturbing absolute slot offsets.
   - Added unit test coverage for compute_active_readback_bytes in test_readback_policy.cpp covering single-slot, multi-slot, interleaved holes, full set, and empty set.

Fixes #3276.

### DIFF
--- a/prosper/frontends/shared/live/live_compute.cpp
+++ b/prosper/frontends/shared/live/live_compute.cpp
@@ -1215,12 +1215,10 @@ size_t cold_storage_result_snapshot_defer_min_bytes() {
 size_t max_gpu_compare_image_bytes() {
     static const size_t bytes = [] {
         const char* value = std::getenv("PROSPER_MAX_GPU_COMPARE_IMAGE_MB");
-        char* end = nullptr;
-        const uint64_t parsed = value ? std::strtoull(value, &end, 10) : 2ull;
-        const uint64_t mib = value && (!end || *end) ? 2ull : parsed;
-        return static_cast<size_t>(
-            std::min<uint64_t>(mib, SIZE_MAX / (1024ull * 1024ull)) *
-            (1024ull * 1024ull));
+        const uint64_t mib = prosper::diag::env_u64_or_default_capped(
+            "PROSPER_MAX_GPU_COMPARE_IMAGE_MB", value, 2ull,
+            SIZE_MAX / (1024ull * 1024ull), "MiB");
+        return static_cast<size_t>(mib * (1024ull * 1024ull));
     }();
     return bytes;
 }
--- a/prosper/frontends/shared/present/readback_policy.hpp
+++ b/prosper/frontends/shared/present/readback_policy.hpp
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <cstddef>
 #include <cstdint>
 
 namespace prosper::frontend {
@@ -29,6 +30,26 @@ constexpr bool is_color_target_readback_wanted(bool has_color_target,
     if (!has_color_target) return true;
     if (target_readback) return true;
     return persistent_id != 0 && !persistent_color;
+}
+
+// Calculates the required staging buffer size covering only the active readback slots (#3276).
+// Sizing the allocation to max(offsets[slot] + bytes[slot]) over the selected slots captures
+// the common case of unbound higher MRT slots without disturbing the absolute offsets.
+template <size_t N, typename OffsetArray, typename BytesArray, typename WantedPred>
+constexpr uint64_t compute_active_readback_bytes(size_t count,
+                                                 const OffsetArray& offsets,
+                                                 const BytesArray& bytes,
+                                                 WantedPred&& is_wanted) {
+    uint64_t max_extent = 0;
+    const size_t limit = count < N ? count : N;
+    for (size_t slot = 0; slot < limit; ++slot) {
+        if (is_wanted(slot)) {
+            const uint64_t extent = static_cast<uint64_t>(offsets[slot]) +
+                                    static_cast<uint64_t>(bytes[slot]);
+            if (extent > max_extent) max_extent = extent;
+        }
+    }
+    return max_extent;
 }
 
 } // namespace prosper::frontend
--- a/prosper/frontends/shared/tests/test_readback_policy.cpp
+++ b/prosper/frontends/shared/tests/test_readback_policy.cpp
@@ -37,6 +37,31 @@ int main() {
     // When there is no color target struct at all, default readback behavior is preserved.
     CHECK(is_color_target_readback_wanted(false, 0, false, false));
 
+    // Active readback sizing (#3276): staging buffer is sized only to the maximum extent of
+    // selected slots, while preserving absolute offsets.
+    using prosper::frontend::compute_active_readback_bytes;
+    const uint64_t offsets[4] = {0, 33u << 20, 66u << 20, 99u << 20};
+    const uint64_t bytes[4] = {33u << 20, 33u << 20, 33u << 20, 33u << 20};
+
+    // Slot 0 only selected (common case: unbound higher MRT slots): only 33 MB, not 132 MB.
+    CHECK(compute_active_readback_bytes<4>(4, offsets, bytes, [](size_t s) { return s == 0; }) ==
+          (33u << 20));
+
+    // Slots 0 and 1 selected: 66 MB.
+    CHECK(compute_active_readback_bytes<4>(4, offsets, bytes, [](size_t s) { return s <= 1; }) ==
+          (66u << 20));
+
+    // Slot 2 selected (with holes at slots 0 and 1): covers up to end of slot 2 (99 MB).
+    CHECK(compute_active_readback_bytes<4>(4, offsets, bytes, [](size_t s) { return s == 2; }) ==
+          (99u << 20));
+
+    // All slots selected: full 132 MB.
+    CHECK(compute_active_readback_bytes<4>(4, offsets, bytes, [](size_t) { return true; }) ==
+          (132u << 20));
+
+    // No slots selected: 0 bytes.
+    CHECK(compute_active_readback_bytes<4>(4, offsets, bytes, [](size_t) { return false; }) == 0);
+
     if (!failures) std::printf("readback_policy: OK\n");
     return failures ? 1 : 0;
 }
--- a/prosper/frontends/shared/tests/test_write_watch_policy.cpp
+++ b/prosper/frontends/shared/tests/test_write_watch_policy.cpp
@@ -227,6 +227,18 @@ int main() {
             "PROSPER_TEXTURE_WRITE_WATCH_PROMOTE_MB", "eight", 8)) * (1u << 20));
         CHECK(kept_budget.try_consume(64u << 20));
         CHECK(!kept_budget.try_consume(64u << 20));  // bounded, which is what was asked for
+
+        // PROSPER_MAX_GPU_COMPARE_IMAGE_MB (#3276): negative or unit typo keeps default 2 MiB rather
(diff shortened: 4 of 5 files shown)


mattias800:
**APPROVED.** Both findings fixed correctly, and I traced the risky one end to end rather than taking the shape on trust.

## Finding 1 — the env parse

Now routed through `env_u64_or_default_capped`, and the hand-rolled `strtoull` block is gone. The arms pin the exact defect I reported:

```
"-1"   -> 2      (was: ULLONG_MAX, clamped to ~SIZE_MAX, silently unbounding the cap)
"2mb"  -> 2
"8"    -> 8
"2048" -> 1024   (cap honoured)
```

That is the strict grammar doing what #3268 built it for, and the `-1` arm is the one that would have caught the original.

## Finding 2 — the buffer sizing, and why I think it is safe

The `max(offset + bytes)` approach is what I suggested, so I checked the integration rather than the idea. Four things had to hold and all four do:

**Placement.** `color_offsets[]` is built at `:5025-5033` from the accumulating full-size `readback_bytes`; the narrowing happens at `:8396`, well after. So the offsets the copies use are unchanged — which is the whole reason this approach is low-risk. `cached_extra` (`:5606+`) and `readback_color0` (`:8365`) are both settled before the lambda reads them.

**No stale reader.** `readback_bytes` has exactly three consumers after the narrowing — `bci.size` (`:8414`), `vkMapMemory` (`:9920`), and nothing else in the file. No path sees the old value.

**No under-allocation.** Every read is inside the new extent:
- slot 0 reads `bytes`, which is `color_bytes[0]` (`:5044`) — matching the lambda's `offsets[0] + bytes[0]`.
- slot 1 reads `color_offsets[1] .. +color_bytes[1]`, gated on `readback_color1`, which is the same term the lambda uses for slot 1.
- slots 2+ read `color_offsets[slot] .. +color_bytes[slot]`, gated on the same `is_color_target_readback_wanted` call the lambda makes.

**No zero-size allocation.** `readback_requested` is `readback_color0 || readback_color1 || (want_color_readback && readback_extra_wanted)`, and each term corresponds to a slot the lambda also selects — so inside `if (readback_requested)` at least one slot is selected and the extent is non-zero.

The unit arms are discriminating: the slot-0-only case would return `132 MB` instead of `33 MB` if the function regressed to the old sum, and the holes case pins that absolute offsets are preserved rather than repacked.

## One thing I would leave exactly as it is

The buffer still covers the holes below the highest selected slot — select only slot 2 and you allocate 99 MB to use 33. That is the documented trade, it is stated in the comment, and it is right: repacking offsets to close the holes would touch the arithmetic that currently makes this change provably safe, for a saving in a case the PR is not about. The common case named in #3276 was unbound *higher* slots, and that one is fully captured.

## Not verified

No build or run from me; CI is the authority on those. And no performance figure — different hardware, and these paths are gated on `gpu_present_active()`, which no headless route here sets.

mattias800:
## First Linux measurement with `gpu_present_active()` actually true — the deferrals work here, and they buy ~11%

Measured on merged `main` (`cd18bd6fb`, i.e. #3275 as landed, before this PR), so these numbers are
about the deferral work rather than about the two fixes here. Posting on this PR because it is the
live one in the lineage and the readback figures are directly relevant.

**Platform:** Linux, AMD Ryzen AI Max+ 395 "Strix Halo" with the **integrated Radeon 8060S**
(unified memory, no discrete VRAM), Mesa 26.1.4 / RADV. Route
`scripts/gta5/reach-performance-story.pad`, native 4K Performance mode, prologue bank heist. Every
earlier number from this box was **headless**, where `gpu_present_active()` is false and none of the
deferral paths execute at all. This one is windowed on a live Wayland session, `gpu-present` adopted.

### The deferrals demonstrably work on this platform

| | headless (paths gated OFF) | windowed (paths gated ON) |
| --- | --- | --- |
| `readback` | **217.5 ms** | **22.4 ms** |

A ~10x drop in the component the change targets. That is the first confirmation these paths do
anything on Linux, and it is a clean before/after because the *only* difference is the gate.

### It moves the frame rate by about 11%

`guest flips` 5.38/s → **5.98/s**. Presented 3.99/s (~4 fps on screen).

Readback simply was not the bottleneck here. The full windowed breakdown:

```
compute            1281.5 ms   33%
renderer-resource  1435.7 ms   37%    [texture 796.1, buffer copy 143.1]
gpu-wait            485.2 ms
gpu-device          303.9 ms    8%
readback             22.4 ms   <- the optimised component
RSS                 6444 MiB
primary evidence: inconclusive -- no component reaches the 40% threshold
```

### Why this does not contradict the 21.3 FPS Windows figure

It is explained by **#3274's own caveat**, which turns out to be the most load-bearing sentence in
that document:

> On unified-memory APU/iGPU architectures (such as Linux / Radeon 8060S), host memory and device
> memory share the same physical address space, so PCIe-specific bus bottlenecks do not exist there
> in the same form.

§2, §3 and §4 recover PCIe transfer cost. On the RTX 4090 that is 9.3 ms/frame of scanout copy plus
1,138 ms of `gpu_compare`; here that work was already close to free, so eliminating it wins ~11%
rather than ~34x. **Both measurements are right**; they are measuring the removal of a cost only one
machine pays.

### The actionable part

`gpu-device` is **8%** of the frame. This is not the iGPU being slow — the device is idle most of
the time. The frame is **host-bound**, and the 70% sitting in `compute` + `renderer-resource` is
work both platforms do and neither has optimised. That is a target, and it is a different one from
the PCIe work.

### Unrelated to this PR, but the more important result

Zero `GRAPHICS submission failed`, zero device losses, across the whole run — with
**`PROSPER_COMPUTE_SKIP_PROGRAM` never set**. Compute ran live for 1,281.5 ms across 1,746 records
and 28 identities. `CLAUDE.md` currently states that GTA V's world is absent because one compute
program hangs the GPU into a RADV hard recovery, and that skipping it is what yields scene content.
On this build the world renders without that flag and nothing hung.

Frame verified as genuinely rendered, not a title asset: `tools/evidence/prerender_check.py` against
the dump returns exit 0, closest asset 128.50/255 where a true match scores ~0.02.

Tracking that separately, since it is a rung question rather than a performance one.

Issue #3121, opened 2026-08-29
GTA V world rendering regressed: #3093's unconditional HTILE invalidation costs its deferred lighting


mattias800:
**Bisected to `aced0703`** ("gpu: a byte-preserving HTILE write must still invalidate retained depth (#3089)", PR #3093, 2026-08-28 02:49). Reported by the project owner 2026-08-29: GTA V's world had rendered correctly since the rendering series and no longer did.

## Symptom

The bank heist still draws — geometry, HUD and radar are all present — but with **no lighting, colour crushed to near-monochrome, and a regular grid artifact** across surfaces where depth-dependent sampling has gone wrong.

## Cause

One line, `prosper/tests/fixtures/render_runner.h`:

```c
-  const bool htile_kill = htile_overlap && htile_invalidates && !byte_preserving;
+  const bool htile_kill = htile_overlap && htile_invalidates;
```

GTA runs a read/modify/write kernel over its whole HTILE plane between the G-buffer and the deferred-lighting pass that samples the depth it just produced. Invalidating on that byte-preserving refresh discards a fully rendered 4K depth buffer the very next pass needs.

## Why it was not caught

#3093 explicitly checked GTA and cleared it: *"disabling it changes neither the peak colour coverage (99.78% in both arms) nor the targets reaching it."* **Peak colour coverage cannot see this defect** — a world drawn with no lighting still covers 99.78% of the frame. The commit even recorded the bound on its own check (*"peak coverage and target identity, not a reviewed frame-by-frame diff"*); the regression landed exactly in that gap.

## There is no trade-off with Blue Prince

One binary, one environment variable, both titles, current master:

| | GTA V | Blue Prince `max_nonblack` |
|---|---|---|
| suppression OFF (as #3093 left it) | broken | **0.2085** |
| suppression ON (restored) | **correct** | **0.2085** |

0.2085 is the same "~21%" #3093 called Blue Prince's restored healthy value. Whatever carries Blue Prince now, it is not that removal.

## Still open

Why two byte-identical writes need opposite handling. The titles are **indistinguishable** at the invalidation site: both `origin=gpu-preserving`, both a fully uniform all-zero plane with zero transitions (GTA 6,500/6,500, Blue Prince 62,000/62,000), differing only in plane size. A "uniform means fast clear" discriminator was measured false before being implemented. Decoding HTILE to tell a clear from a refresh — named in the code comment long before this — remains the actual fix.

PR #3278, opened 2026-09-03
DO NOT MERGE — HTILE invalidation: breaks GTA V characters/lighting (#3264)

### LINKED ISSUE #3264: Blue Prince (PPSA25009): the rung-6 blue-prince-title guard now renders pure black (max=0) — it was #2950's positive control ten days ago
## Summary

**`blue-prince-title` fails on current `main`, producing frames whose maximum pixel value is 0.**
It passed ten days ago on the same machine with the same driver — and it passed as **#2950's
declared positive control**, which is what makes this worth its own issue rather than a line in a
report.

## The two measurements

| | 2026-08-23 (#2950's own table) | 2026-09-02 (tonight) |
| --- | --- | --- |
| revision | `origin/master` `f5624cf6` | `main` `e13fd40af` |
| `blue-prince-title` | **OK** — frames 10, qualifying 10, structural 10, non-black 10 | **FAIL** — qualifying 0 < 8, non-black coverage 0 < 8 |
| machine | Linux / AMD Radeon 8060S (RADV STRIX_HALO) | same |
| driver | Mesa 26.1.4 | Mesa 26.1.4 |

The driver, the machine, the dump, the guard config and the baseline are all unchanged between
those two rows. The variable is prosper.

## The frames are pure black, not dark

All eight retained failure frames:

```
00.png (480, 270) mean=0.00 uniq=1 max=0
01..07                     identical
```

`max=0` across every channel of every pixel — not a dim scene, not a fade, not a near-black menu.
`pixel_changes: 0`, `longest_consecutive_content_matches: 0`.

The boot log shows the title publishing frames and then freezing on one:

```
2 distinct of 16994 published; run average 0.0 fps; presented 330.2 fps
42% of the 51.5 s run active
frames 13713..16992 (42.1s..52.1s) all crc=9b48bfd1
```

So it is not hung — it publishes 330 frames/s of the same black image, having been "active" for 42%
of the run.

## Independently corroborated outside the snapshot machinery

Before running the guard I reproduced it with plain `tools/screenshot`, no snapshot code involved,
two arms differing only in resolution (fresh save roots in both, per the guard's
`savedata_policy: fresh`):

| arm | output | result |
| --- | --- | --- |
| `PROSPER_RENDER_SCALE=4` (the guard's own setting) | 480x270 | 6/6 frames `mean=0.0 uniq=1` |
| `PROSPER_RENDER_SCALE=1` | 1920x1080 | 6/6 frames `mean=0.0 uniq=1` |

That A/B was built to test the hypothesis that tonight's black default-launch frames were just this
title being too slow at full resolution to reach the title screen in the window (#2215: ~1-3 fps).
**It falsifies that hypothesis** — the guard's own configuration is black too, so resolution is not
the explanation.

A separate default-launch run earlier the same night (no fresh save, scale 1, 239 s) was also black:
`2 distinct of 37796 published`, 9% active.

## What is NOT established

- **Not bisected.** The window is `f5624cf6` (2026-08-23) → `e13fd40af` (2026-09-02), which is a lot
  of commits. I have not narrowed it.
- **The 2026-08-23 row is another agent's recorded measurement**, quoted from #2950's table, not one
  I re-ran. It is a specific enough claim (four counts, all 10) that I am treating it as reliable,
  but it is not first-hand.
- **No mechanism proposed.** I have not looked at what changed, only that the output did.
- The guard's baseline was reviewed **2026-07-26**, so it predates both rows.

## Why this matters beyond one guard

#2950 investigated three rung-6 guards failing and used `blue-prince-title` as the positive control
that proved the harness, the dumps, the frontend and the machine were all working. **That control no
longer passes.** Conclusions in #2950 that rest on it should be re-examined — not because they are
wrong, but because the argument they were made with is no longer available. Commented there
separately.

Refs #2950, #1808.

mattias800:
## One term, three titles

`4b2b4ec92` added `&& !byte_preserving` to the depth/stencil invalidation predicate. This removes it.

```cpp
- const bool htile_kill = htile_overlap && htile_invalidates && !byte_preserving;
+ const bool htile_kill = htile_overlap && htile_invalidates;
```

## There was never a trade-off

Suppressing the invalidation leaves **stale depth**. The two titles present that one defect
differently, which is why every previous attempt looked like a trade:

* **Blue Prince** fails the depth test against it and draws nothing → black screen.
* **GTA V** *samples* it and hangs the GPU → `VK_ERROR_DEVICE_LOST` cascade → black world.

Each earlier change fixed one title, measured only that title, and appeared to break the other.

## Measured — one binary pair, same session, same routes, two runs each

| | with the term (`main`) | without it (this PR) |
| --- | --- | --- |
| Blue Prince title | 0/5 frames, **2** distinct of 38,881 | **4/5 frames**, **256** distinct of 38,586 |
| GTA V routed world | 3/8 then black | **8/8 frames render** |
| GTA V device losses | **122** and **175** | **0** and **0** |

The GTA V control goes black at the same frame in both runs — per-frame luminance
`73 115 116 0 0 0 0 0` — and the experiment renders all eight in both, with luminance matching
within ~1 point. So the route lands on the same scenes and this is like-for-like, not two different
moments.

## Guards

| guard | `main` | this PR |
| --- | --- | --- |
| `blue-prince-title` | **FAIL** — qualifying frames 0 < 8 | **OK** |
| `dead-cells-gameplay` | OK | **OK** |
| `messenger-scene` | OK | **OK** |
| `alexkidd-gameplay` | OK | **OK** |

The last three were run as regression controls, because a renderer change that fixes one guard and
quietly breaks others would otherwise read as a win.

## This was recorded as unsound before it shipped

`GTA5_STATUS.md`'s own `## Ruled out`, **2026-08-28** (#3089) — one day before `4b2b4ec92`:

> **The HTILE `gpu-preserving` suppression is load-bearing for GTA V's picture. FALSE, measured.**
> … That rule is **unsound in principle**: prosper never writes rendered HiZ back into the guest
> HTILE plane, so the guest copy is a constant the guest's own writes keep reproducing, and the
> comparator reports `changed=0` for a fast CLEAR exactly as readily as for the decompress it was
> written for.

It also cites **Dead Cells #611**, where sparing the invalidation makes gameplay geometry disappear.
So this rule has now cost **three** titles.

`4b2b4ec92`'s own comment states both halves of the contradiction — that #3093's removal *"fixed Blue
Prince"*, and that restoring it *"restores behaviour known to be correct for both"*. The
measurements say the first is the accurate one.

## What a future attempt needs first

Recorded in the code comment rather than left to be rediscovered: the discriminator `4b2b4ec92`
named and did not build — **decode HTILE to tell a clear from a refresh**. A byte comparison cannot,
and `guest_gpu_write_origin()` returns a **process-global** set by whichever write ran last
(`gpu_executor.cpp`, `g_guest_write_origin`), so keying on it is order-dependent by construction.

That global is also a plausible mechanism for the reported intermittency — *"sometimes Blue Prince
works, sometimes it doesn't"* — and for why `4b2b4ec92`'s author measured Blue Prince at `0.2085`
where every run here reads `0`. Worth its own issue; not fixed here.

## Verification

Build exit 0. Guards as tabled. Two independent A/B runs per arm per title. Void arms (a `127` from
a missing target, and two runs my own concurrent job SIGTERMed) were **discarded, not folded in**.

Not verified: Windows or macOS behaviour — everything here is Linux/RADV on an integrated
Radeon 8060S. The device-loss cascade may be RADV-specific even if the stale-depth defect is not.

Fixes #3264

### DIFF
--- a/prosper/tests/fixtures/render_runner.h
+++ b/prosper/tests/fixtures/render_runner.h
@@ -3905,12 +3905,30 @@ inline size_t invalidate_persistent_ds_guest_write(uint64_t addr, uint64_t size)
         // 62,000/62,000, first word 0x00000000 in both, differing only in plane size (73,728 vs
         // 49,152 words, i.e. resolution). A "uniform plane means a fast clear" discriminator was
         // hypothesised and MEASURED FALSE before it was written; PROSPER_HTILE_UNIFORMLOG below is
-        // the instrument that killed it. So this restores behaviour known to be correct for both
-        // titles without knowing why the same event needs opposite handling, and the actual fix the
-        // comment above still names -- decoding HTILE to tell a clear from a refresh -- remains open.
-        const bool byte_preserving =
-            std::strcmp(prosper::gpu::guest_gpu_write_origin(), "gpu-preserving") == 0;
-        const bool htile_kill = htile_overlap && htile_invalidates && !byte_preserving;
+        // the instrument that killed it.
+        // #3264: the `!byte_preserving` term is REMOVED, not weakened. It suppressed the
+        // invalidation whenever the write carried the `gpu-preserving` origin, and that is unsound
+        // for the reason GTA5_STATUS.md's own Ruled out section recorded on 2026-08-28 (#3089), one
+        // day before it was reinstated: prosper never writes rendered HiZ back into the guest HTILE
+        // plane, so the guest copy is a constant the guest's own writes keep reproducing, and the
+        // byte comparator reports `changed=0` for a fast CLEAR exactly as readily as for the
+        // decompress the rule was written for.
+        //
+        // There was never a trade-off between the two titles. Suppressing the invalidation leaves
+        // STALE DEPTH, and the two titles present the same defect differently -- Blue Prince fails
+        // the depth test against it and draws nothing, GTA V samples it and hangs the GPU. Measured
+        // on one binary pair, same session, same routes:
+        //
+        //   Blue Prince title    4/5 frames render (256 distinct)  vs  0/5 (2 distinct)
+        //   GTA V routed world   8/8 render, 0 device losses       vs  3/8 then black, 122 losses
+        //
+        // Dead Cells #611 is the prior counterexample where sparing the invalidation makes gameplay
+        // geometry disappear, so this rule has now cost three titles. Restoring it again needs the
+        // discriminator its own author named and did not build: decode HTILE to tell a clear from a
+        // refresh. A byte comparison cannot, and the `gpu-preserving` origin is a process-global set
+        // by whichever write ran last (gpu_executor.cpp, `g_guest_write_origin`), so keying on it is
+        // order-dependent by construction.
+        const bool htile_kill = htile_overlap && htile_invalidates;
         if (!depth_overlap && !stencil_overlap && !htile_kill) continue;
         if (depth_overlap || htile_kill) image.depth_valid = false;
         if (stencil_overlap || htile_kill) image.stencil_valid = false;


mattias800:
## DO NOT MERGE — human review says this breaks GTA V, and my verification could not see it

The repository owner looked at the GTA V frames and reports **characters are not rendering**. I have
re-examined them and I agree. Comparing the same scene against a run on `main` earlier today: the
character at frame left was clearly drawn — jacket, mask, weapon, properly lit — and the interior
had warm illumination. On this branch that figure is a barely-visible dark fragment, the scene is
markedly darker, and the window mesh dominates.

That is exactly what `4b2b4ec92` predicted as the symptom of removing this term: *"the bank heist
drawn without illumination and under a grid artifact from depth-dependent sampling."* I quoted that
sentence in this PR and then did not check for the thing it describes.

## My verification was structurally incapable of catching this

I measured **frames with content**, **device losses**, and **per-frame luminance**. A scene missing
its characters and its lighting still has content, still varies in luminance, and still reports zero
device losses. None of those metrics can see a missing character.

This is the same failure this PR's own description criticises in #3093 — which cleared GTA V on
*"peak colour coverage (99.78% in both arms)"* while the world was drawing with no lighting at all.
`4b2b4ec92` wrote that up as the lesson, I quoted it approvingly, and then reproduced it with
different numbers. A metric chosen to rule a defect out has to be able to express that defect, and
mine could not.

The guard result (`blue-prince-title` FAIL → OK, three controls staying green) stands, and so does
the device-loss measurement. What does **not** stand is the claim that GTA V is unaffected.

## What is still true, and what is now open

Unchanged and reproduced twice each:

* Blue Prince: black → renders; its guard goes red → green.
* GTA V device losses: **122 and 175 → 0 and 0**.

But "both games work" is **false**. The honest position is now:

| | with the term (`main`) | without it (this PR) |
| --- | --- | --- |
| Blue Prince | black | renders |
| GTA V characters / lighting | render | **broken** |
| GTA V device losses | 122–175 | 0 |

So **neither state is correct**, and the trade-off I claimed did not exist is at least partly real
after all — my mechanism story ("one defect, two presentations") explained the black screens but not
this, and I over-generalised it.

The device-loss result is still interesting and unexplained: `main` sustains 122–175
`VK_ERROR_DEVICE_LOST` in this route and nobody has attributed those. That is worth its own
investigation and should not be lost.

## Next

The real fix is the one `4b2b4ec92` named and did not build, and which I should not have tried to
route around: **decode HTILE to tell a clear from a refresh**. A byte comparison cannot, and the
`gpu-preserving` origin is a process-global set by whichever write ran last, so keying on it is
order-dependent by construction.

Holding this PR. Not closing it yet — the measurements and the history in it are worth keeping while
the real fix is worked out — but it must not merge in this form.

mattias800 (commented):
# Independent review — REJECTED

Reviewed at head `b3f2c316`, base `origin/main` (`59d2e815e`). No GPU used: no `prosper-app`, no
`tools/screenshot`, no `gpu_replay`, no snapshot run. One CPU-only ctest case was built and run.

**REJECTED.** Two independent blockers, either of which is sufficient on its own. The author has
since withdrawn the PR's central claim on the basis of GTA V frames showing missing characters and
lighting; finding 1 below is separate from that and was found before it, so it is worth recording
whatever happens to this branch.

---

## 1. BLOCKING — the change breaks a registered ctest case and the suite was never run

`prosper/tests/render/test_ds_layer_stride.cpp` case 10 asserts, verbatim, the behaviour this diff
removes:

    prosper::gpu::set_guest_gpu_write_origin("gpu-preserving");
    const size_t preserved = prosper::test::invalidate_persistent_ds_guest_write(kHtile, 4096);
    check(preserved == 0 && image.depth_valid && image.stencil_valid,
          "a byte-preserving HTILE rewrite preserves both retained aspects (#3121)");

With `htile_kill = htile_overlap && htile_invalidates`, `htile_overlap` is true for that entry
(`kW=kH=256` -> 1024 blocks -> `htile_size` 32768, write of 4096 at the base) and
`htile_invalidates` defaults on, so `preserved` is 1 and both aspects go invalid. Deterministic.

Reproduced, not derived:

    ctest -R '^ds_layer_stride$' --output-on-failure --no-tests=error
    FAIL: a byte-preserving HTILE rewrite preserves both retained aspects (#3121)
    ok: a changed HTILE write still invalidates both aspects
    ok: a preserving write outside every plane of this surface invalidates nothing
    == FAIL: 1 ==
    0% tests passed, 1 tests failed out of 1
    317 - ds_layer_stride (Failed)

CI agrees: the `Linux` job's **Test** step exited **8** on this head, while the same workflow was
green on the merge base `59d2e815e`.

The verification claim needs correcting too. "Build exit 0" is in the PR body; the worktree's build
directory contains `libprosper_core.a` and `screenshot` and **no test binary at all** — I had to
build `test_ds_layer_stride` myself to run it. So the build that was reported did not include the
test suite, and ctest was not run. Under the charter a quoted ctest result needs
`--no-tests=error` and a test count; here there was no ctest result to quote.

The meta-point is the sharper one. `BLUE_PRINCE_STATUS.md` explains that this case exists precisely
because the older #611 guard could not reach the state (it drives `notify_guest_gpu_write`, so the
origin is `unknown` and the `strcmp` was false in every arm) — "the corrected case in
`test_ds_layer_stride.cpp` sets the origin explicitly". It is the one guard in the repo built to be
able to fail on this predicate, and this PR walked past it. A successor must update that case in the
same commit and say in the message which arm changed and why.

## 2. BLOCKING — the GTA V clearance uses a metric the project has already recorded as blind here

Now conceded by the author, so briefly — but with the mechanism, because it is more specific than
"the trade-off is real" and it points at the fix.

`depth_valid` is not only the LOAD-vs-discard bit. `prosper/tests/fixtures/render_runner.h:3599`
gates the **sampled-depth bridge** on the same flag, and the comment immediately above it
(`:3588`) says exactly what that costs this title:

> GTA V's deferred lighting pass samples the main depth `0x2052ac0000`, and the bridge declines it
> here with `dvalid=0` while `svalid=1` on the SAME surface [...] The sample then falls back to
> guest memory, which is black for a surface prosper rendered into.

So "unlit, characters missing, grid artifact" is the bridge declining, not the attachment being
discarded. Clearing `depth_valid` more often re-arms that path directly. Frame-count, per-frame
luminance and device-loss count cannot see it — a scene missing its characters and its lighting has
content, varies in luminance, and loses no device. That is peak colour coverage's failure mode under
a different name, which the PR itself quotes as a cautionary tale.

## 3. The `#3089` citation — accurate as a quote, not sound as an argument

Asked for specifically, so stated precisely. **The quoted text is verbatim-correct** at
`origin/main` in `prosper/docs/GTA5_STATUS.md`'s `## Ruled out` section: the entry is dated
2026-08-28 (#3089), it does say the rule is "unsound in principle", and it does give the
never-writes-HiZ-back reasoning as quoted. Nothing in the quotation is invented.

Two things around it are wrong, and both cut against the PR:

- **It is the older of two entries, and the newer one rebuts it.** The same file carries
  `## Ruled out — HTILE byte-preserving suppression (2026-08-29)` from #3121, whose first row is
  "Removing the byte-preserving HTILE suppression is safe for GTA V, because peak colour coverage is
  99.78% in both arms — **False, and the check is the reason it passed.**" Quoting the superseded
  entry as "recorded as unsound before it shipped" without the entry that answers it is selective,
  and it is the failure mode the charter names: a citation reads as already-checked.
- **The #3089 entry states its own limit and the PR omits that sentence** — "the GTA V arms were
  compared on peak colour coverage and target identity, not by a reviewed frame-by-frame diff."
  That caveat is the whole story here, and it was in the paragraph being quoted.

The *principle* half of #3089 (byte equality with guest memory says nothing about equality with a
renderer-owned image) was never rebutted and still stands. It is the empirical half — "removing it
does not cost GTA V anything measurable" — that #3121 and now the frames falsify.

## 4. The *Dead Cells* #611 precedent — holds as a precedent, not as a cost

#611 is real, closed, and is what the docs say it is: one guest depth surface reused across gameplay
frames with no captured clear (883/883 submits, `LESS_OR_EQUAL` tests, zero clear intent), so
retained depth accumulates until world geometry is rejected into a gray field. It is a genuine
counterexample to any "spare the invalidation" policy, and both `GTA5_STATUS.md` and
`BLUE_PRINCE_STATUS.md` cite it that way — as the reason no preservation policy was proposed.

But "**this rule has now cost three titles**" is wrong, in the PR body, the commit message and the
new code comment. #611 is dated 2026-07-13 and the rule arrived with `e63f4038` (#2996) roughly six
weeks later; the rule never cost *Dead Cells* anything. It is two titles plus one prior precedent
that argued against the rule in advance — which is a good enough sentence without the inflation.
Please correct it wherever it lands next; the code comment is the copy that will outlive the PR.

## 5. The new comment states a mechanism the cited code contradicts

> the `gpu-preserving` origin is a process-global set by whichever write ran last
> (`gpu_executor.cpp`, `g_guest_write_origin`), so keying on it is order-dependent by construction

Both halves are wrong at the cited file:

- `prosper/src/gpu/execute/gpu_executor.cpp:11559` — `thread_local const char* g_guest_write_origin`.
  Thread-local, not process-global.
- On the path that actually reaches this predicate it is **not last-writer state at all**: it is
  per-write data. `queue_guest_gpu_write` (`prosper/frontends/shared/live/live_renderer.cpp:182`)
  stores the origin on the queued range, and `drain_guest_gpu_writes` (`:409`) restores that range's
  own origin immediately before calling `invalidate_persistent_ds_guest_write`. The header documents
  this as deliberate (`prosper/src/gpu/execute/gpu_execute.hpp:970-975`): "The origin is passed AS
  DATA rather than read from the thread-local by the observer [...] a notifier that knows its own
  classification must hand it over here or the information is lost."

So the intermittency hypothesis built on it, and the offered explanation for why #3121 measured Blue
Prince at `0.2085` where this branch reads `0`, are unsupported. The mundane candidate is
unexamined: **100 commits** separate `4b2b4ec92` from `origin/main`, several of them in exactly this
path — `a0daed9c3` (persistent-resource caches are one domain with one owner, #2953), `8c74478f9`
(compute writeback path), `5a0336688` (seed/writeback aliasing). If a successor needs to explain that
discrepancy, that window is where to look, and it is bisectable.

## 6. Mechanical — all clean

`git diff --check` clean; one file; no absolute host paths in the diff or the PR body; no remaining
reference to `byte_preserving` anywhere in the tree. The dead-code removal is correct and complete:
`<cstring>` is still required (12 other `strcmp` uses in the same header), and
`guest_gpu_write_origin()` keeps two callers in this file (`:3794` `PROSPER_DSLOG`, `:3880`
`PROSPER_HTILE_UNIFORMLOG`), so it is not dead and leaving it is right.

Guards: 4 of the 17 in `tools/snapshot/snapshots.json` were run. Not a merge gate during
development, but `blue-prince-hall` is the same title's other guard and the one most likely to move;
it was not among them.

**Windows / macOS: not a blocking gap for this shape of change.** The predicate is
platform-independent CPU logic in a shared header with no platform branch, and the same unit suite
runs in the Windows and macOS CI jobs. The author's own caveat is about the *symptom* (a RADV
device-loss cascade), not about the change. I would not hold a successor for cross-platform
measurement here.

---

## What survives, and where I would point the successor

**Blue Prince stands.** Black -> renders, guard red -> green, two runs, and #3264 documents the
regression independently. Whatever lands next must keep that evidence, and must also correct two
`## Ruled out` rows that currently instruct the next reader to disbelieve it —
`GTA5_STATUS.md`'s "Restoring the suppression re-breaks Blue Prince [...] **False.**" and
`BLUE_PRINCE_STATUS.md`'s "If this title regresses to black again, **the suppression is not the
first thing to suspect**". Under the charter that recording happens in the same PR, not afterwards.

**The device losses deserve their own issue.** 122 and 175 `VK_ERROR_DEVICE_LOST` on a routed GTA V
run on `main` is not a healthy baseline regardless of this predicate, and nobody owns that number.
File it with the two arms as they were measured — but do not attribute it to this predicate without
a mechanism, because "0 in the other arm" is a correlation across a change that also alters how much
retained depth survives.

**On "decode HTILE to tell a clear from a refresh" — my read is that it is probably not achievable,
and more importantly probably not the frontier.** Three reasons, all from the record and the code
rather than from this branch's runs:

1. **There is likely no guest clear-vs-refresh event in the data to decode.** `GTA5_STATUS.md`
   § "What invalidates the retained depth — 100% of it is prosper's own writeback (2026-08-17)"
   measures **30,841 of 30,841** DS invalidations on the route as prosper invalidating its own
   caches from its own writebacks, with the unattributed bucket at **0**. For the main depth
   `0x2052ac0000`, 1,542 of 2,276 arrive as `gpu-preserving` — a path whose entire contract is that
   the guest bytes were not modified. A decoder pointed at those bytes is decoding prosper writing a
   constant zero plane back over itself, which is exactly why both titles read
   `first=0x00000000, uniform, zero transitions` and why the uniformity discriminator died. It would
   fail for the same structural reason, at greater cost.
2. **The document already names the open question and nobody has worked it**: "why does prosper's own
   byte-preserving writeback overlap this depth surface's HTILE range at all". Narrowing the
   writeback so it stops covering the metadata plane deletes the event instead of classifying it, and
   it cannot regress Blue Prince, because it removes a write prosper invented rather than changing
   how a real one is handled. There is a concrete instance of the same family in the other direction
   at `prosper/frontends/shared/live/live_compute.cpp:~10041`: prosper `memset`s a surface's DCC
   metadata to `0xff` and announces it as a guest write, and its own comment says "which for a DEPTH
   surface is its HTILE base [...] So this reset can discard a retained depth image [...] whether it
   SHOULD invalidate is a separate question that needs the operation's real HTILE semantics proven."
3. **`depth_valid` is one bit answering two questions with opposite safe defaults.** For the
   attachment, stale depth must be discarded or Blue Prince's scene is rejected. For the bridge,
   serving the retained image is strictly better than the alternative, which `:3588` says is *black*.
   Splitting those needs no HTILE semantics at all.

**The cheapest next experiment, and it is one run per title.** `PROSPER_DS_BRIDGE_IGNORE_VALID`
already exists at `render_runner.h:3600` and does precisely this: serve the retained aspect despite
`dvalid=0`, and log `[dsbridge] OVERRIDE ... served despite invalid (count=N)` so a null cannot be
read as a negative when the lever never fired. Run **this branch** with it set, on GTA V. If the
characters and lighting come back while Blue Prince still renders, the bridge gate is what costs
GTA V — not the invalidation — the two titles are separable, and the fix is splitting the bit rather
than a fifth flip of this predicate. If nothing changes, the bridge is not the gate and that is worth
knowing before anyone builds a decoder. Check the OVERRIDE count either way.

Whatever lands next: keep the suppression behind an off-by-default environment switch so the fifth
attempt can A/B both behaviours in one binary, the way `PROSPER_UD_TAIL_ALIGN` is kept for exactly
that reason. This predicate has been flipped four times and every flip so far has had to rebuild its
own comparison from scratch.

**REJECTED.**

mattias800:
## Four corrections to my own PR, from review — including one that says I never ran the tests

Recording these against the PR text rather than only in a thread, because three of them are now
wrong in the commit message and the shipped code comment as well.

### 1. I did not run the test suite, and there is a guard built for exactly this predicate

`prosper/tests/render/test_ds_layer_stride.cpp` case 10 asserts verbatim the behaviour this diff
removes:

```cpp
check(preserved == 0 && image.depth_valid && image.stencil_valid,
      "a byte-preserving HTILE rewrite preserves both retained aspects (#3121)");
```

Review built and ran it: **FAIL, 1 of 1**, and CI's `Linux` Test step exits **8** on this head against
green on the merge base. My "Build exit 0" in the PR body was the **frontend target only** — the
worktree build directory contains `libprosper_core.a` and `screenshot` and **no test binary at all**.

`BLUE_PRINCE_STATUS.md` records that this case exists *because* the older #611 guard was structurally
unable to fire on this predicate. It is the one guard written to catch this, and I walked past it.

### 2. "This rule has now cost three titles" is wrong

*Dead Cells* #611 is dated 2026-07-13; the rule arrived with `e63f4038`/#2996 six weeks later. #611
is a **precedent** for why sparing the invalidation is unsound, not a title this rule cost. It is
**two titles plus one precedent**, and the error is in the PR body, the commit message and the new
code comment.

### 3. My `#3089` citation omitted the caveat that undercuts it

The paragraph I quoted ends: *"compared on peak colour coverage and target identity, **not by a
reviewed frame-by-frame diff**."* I quoted the sentence before that and dropped it — while arguing
that #3093 had cleared GTA V on a metric too coarse to see the defect. The same objection applies to
the clearance I was citing, and it was in the text.

There is also a **newer** entry I did not look for: `GTA5_STATUS.md`'s
`## Ruled out — HTILE byte-preserving suppression (2026-08-29)` (#3121), whose first row rebuts
#3089's GTA V clearance by name. The *principle* half of #3089 stands; its empirical half does not.

### 4. The intermittency mechanism I proposed is unsupported

I claimed `guest_gpu_write_origin()` returns a process-global set by whichever write ran last, making
the predicate order-dependent. **It is `thread_local`** (`gpu_executor.cpp:11559`), and on this path
it is per-write *data*, not last-writer state: `queue_guest_gpu_write` stores it per range and
`drain_guest_gpu_writes` restores it before the call, which `gpu_execute.hpp:970-975` documents as
deliberate.

So my explanation for "sometimes Blue Prince works" and for why #3121's author measured `0.2085` is
unsupported and should not be carried forward. The unexamined candidate is simpler: **100 commits**
between `4b2b4ec92` and `origin/main`, several touching this path.

## The mechanism I should have found

`depth_valid` is not only the LOAD/discard bit. `render_runner.h:3599` gates the **sampled-depth
bridge** on it, and `:3588` states that a decline makes GTA V's deferred lighting sample
`0x2052ac0000` from guest memory, *"which is black for a surface prosper rendered into"*. That is
where the missing characters and lighting come from — and frame count, luminance and device-loss
count are blind to it by construction, which is why my verification passed.

## Next, and it is cheap

`PROSPER_DS_BRIDGE_IGNORE_VALID` already exists (`render_runner.h:3600`) and serves the retained
aspect despite `dvalid=0`, logging an override count so a null cannot be read as a negative. One run
of **this branch** with it set, on GTA V: if characters and lighting return while Blue Prince still
renders, then the two titles **are** separable, the cost is the overloaded `depth_valid` bit, and the
fix is splitting that bit rather than a fifth flip of the predicate.

Holding until the GPU is free — another agent is benchmarking.

Issue #1334, opened 2026-07-24
MODE=3 resolve marks the destination's stale persistent GPU image as valid — consumers sample the frozen seed forever (Blue Prince stuck/blown present)


mattias800:
Root cause isolated from the run-B timeline-materialized capsule of Blue Prince's Day One scene (submit 2958, current master, all MODE=3 resolves firing — see #1287 work log).

**Failure scenario**: the MODE=3 resolve implementation (prosper/frontends/shared/live_renderer.cpp, resolve-copy site) materializes the source pixels and then copies the whole `RttSurf` onto the destination key: `g_rtt[rdst] = std::move(resolved)`. That copy **inherits the source's `gpu_valid=true`** while persistent-image lookups are keyed by ADDRESS — so the flag now asserts that rdst's persistent Vulkan image is current, when that image still holds whatever it was last seeded/rendered with. Every sampled consumer of rdst (`has_gpu_live_rtt` requires `gpu_valid` + `find_persistent_color_target(addr)`) then direct-binds the stale GPU image and ignores the fresh CPU pixels the resolve just produced.

**Observed effect**: Blue Prince resolves its FP16 scene into 0x210b99.... six times per frame; the tonemap-copy AND the final VO blit both sample that address. Both read the stale persistent image = the frame that was current when the surface was first seeded. The presented image freezes on that relic (the blown mansion of #1287; the #1216 stall/oscillation is consistent with the same mechanism), while the scene pass keeps rendering the real camera (proven: within one capsule, the scene draws the bridge approach in correct dark tones while the blit outputs the frozen mansion — byte-identical offline and live).

**Fix**: after the CPU-side resolve copy, clear `g_rtt[rdst].gpu_valid` so consumers take the `has_cpu_live_rtt` path and upload the fresh resolve content. Follow-up (separate perf issue): implement the resolve as a GPU→GPU persistent-image copy so the destination stays GPU-resident instead of re-uploading ~16 MB per resolve.

mattias800:
CLAIMING: claude-session d2181233 (bp-res worktree) | branch fix/issue-1334-resolve-gpu-invalidate | 2026-07-24T17:54:53Z

mattias800:
Root-cause statement CORRECTED after full within-frame instrumentation of the run-B capsule (submit 2958, current master + local diagnostics):

The gpu_valid-inheritance defect described above is real but SECONDARY. The primary mechanism is one level deeper:

[...]

PR #3280, opened 2026-09-03
perf(render): evict older targets in copy_persistent_color_target when over capacity

mattias800:
## Summary

`copy_persistent_color_target` lacked an LRU cache eviction loop, unlike the slot 0, slot 1, and MRT extra slots allocation paths. Once the persistent color target cache reached `persistent_color_target_count_limit()` (256 targets), every subsequent resolve copy failed with:
`[msaa] resolve GPU copy failed (resolve destination exceeds the persistent target budget) — destination keeps CPU pixels only (#1334)`

This stripped `gpu_valid` from the destination image and forced every downstream pass sampling that resolved surface to fall back to CPU de-tiling all 33.4 MB from guest memory on host CPU threads. In GTA V (PPSA04263), the 4K resolve destination `0x204aee0000` was de-tiled on the CPU 89 times per 5s window (~10.4 ms/frame stall).

## Fix

Added LRU cache eviction loop of unpinned older targets under memory or count pressure before allocating the destination image in `copy_persistent_color_target`, protecting the source target with `pin_count` during eviction.

## Before / After Measurements (Linux Windowed Benchmark on Radeon 8060S / Strix Halo)

Measured via automated story mode pad script (`scripts/gta5/reach-performance-story.pad`, prologue bank heist) in live 5-second `.prperf` windowed performance captures:

| Metric | Baseline (`10:11`) | With Fix (`13:52`) | Improvement |
| :--- | :--- | :--- | :--- |
| **`0x204aee0000` (4K resolve dest) CPU de-tile** | **302.50 ms** (89 passes) | **0.00 ms** (0 passes) | **Eliminated (~10.4 ms/frame saved)** |
| **`res_buffer_ms`** | 298.79 ms | 119.98 ms | **~2.5x faster buffer uploads** |
| **`setup_resources_ms`** | 597.37 ms | 410.93 ms | **Saved 186.4 ms per window** |
| **Resolve copy failure warnings** | Logged continuously | **0 occurrences across all frames** | **100% successful GPU resolve copies** |

## Verification

- Added regression test in `tests/gpu/state/test_multidraw_render.cpp` asserting that `copy_persistent_color_target` successfully evicts older targets and succeeds when the persistent color target cache is at capacity limit.
- Verified all 350 active unit tests in `ctest` pass (100%).
- Based on `main`.

### DIFF
--- a/prosper/tests/fixtures/render_runner.h
+++ b/prosper/tests/fixtures/render_runner.h
@@ -4363,8 +4363,9 @@ inline bool readback_persistent_color_target(uint64_t id, uint32_t width, uint32
 // source — after a #780 guest-write CPU-copy discard, consumers (Blue Prince's compute tonemap)
 // imported the stale image and read black, cascading the #1287 display-chain collapse. Copy the
 // device-local pixels so the destination image is genuinely valid. Creates the destination if
-// absent (same shape/usage as the render path; budget-checked WITHOUT eviction — an allocation
-// failure returns false and the caller keeps the CPU pixels as the only truth, fail-visibly).
+// absent (same shape/usage as the render path; budget-checked with LRU eviction of unpinned older
+// targets — an allocation failure returns false and the caller keeps the CPU pixels as the only truth,
+// fail-visibly).
 inline bool copy_persistent_color_target(uint64_t src_id, uint64_t dst_id, uint32_t width,
                                          uint32_t height, VkFormat format, std::string& error) {
     error.clear();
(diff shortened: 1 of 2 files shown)


mattias800 (commented):
**APPROVED**, with one documentation suggestion. I checked the reference-lifetime hazard rather than
the diff shape, because that is where this change could go wrong silently.

## The hazard I went looking for, and why it is handled

The eviction loop runs while a **live pointer to the destination entry** is held:

```cpp
auto [dst_it, dst_inserted] = persistent_color_target_cache().try_emplace(dst_key);
PersistentColorTargetImage* dst = &dst_it->second;      // live across the loop
...
while (... && evict_persistent_color_target(ctx, persistent_color_target_generation())) {}
```

`evict_persistent_color_target` calls `cache.erase(victim)`, so evicting the destination would leave
`dst` dangling — and a freshly `try_emplace`d entry looks like the *ideal* victim, since the
predicate prefers the smallest `last_use`.

It cannot happen, and the two participants are protected by **different** mechanisms:

| | protected by | why not the other one |
| --- | --- | --- |
| `dst` | `dst->last_use = ++generation`, and the loop passes that same generation, which the predicate skips | a pin would work too, but the generation is already assigned |
| `src` | explicit `++src->pin_count` around the loop | its `last_use` is an older generation, so the generation guard does **not** cover it |

That asymmetry is correct and non-obvious. **Suggestion:** one comment at the pin saying why `src`
needs it and `dst` does not — otherwise a future edit that moves the `last_use` assignment below the
loop, or that drops the pin as redundant, reintroduces a use-after-free with nothing to catch it.

## The test discriminates

It fills the cache to `count_limit` with entries at `last_use = 1` (evictable), then asserts the copy
**succeeds**. Before this change the copy would hit `cache.size() > count_limit` and fail with
"resolve destination exceeds the persistent target budget", so the arm fails without the fix. The
dummies being `valid = false` also means eviction takes the cheap path and does not exercise the
readback sink, which keeps the arm about capacity rather than about eviction's side effects.

## Smaller notes, none blocking

- `++src->pin_count` / `--src->pin_count` is not exception-safe. Nothing in the loop currently
  throws, but a scope guard would make that durable rather than incidental.
- `limit - ir.size` is unsigned and would wrap, correctly guarded by the enclosing
  `if (ir.size <= limit)`. Worth keeping those two together if either is ever edited.
- The post-loop re-check still tests `persistent_color_target_bytes() > limit - ir.size` and the
  count limit, so a failed eviction still fails visibly rather than proceeding over budget. Good —
  the change makes eviction *attempted*, not *assumed*.

## Not verified

CI was mid-flight when I read this (4 pass / 5 pending). I did not build or run anything, and used
no GPU — another agent is benchmarking.

PR #3281, opened 2026-09-03
fix(render): preserve in-flight frame depth under byte-preserving HTILE writebacks

mattias800:
## Summary

Resolves the rendering conflict between **Blue Prince** (#3264) and **GTA V** (#3121) introduced by #3278.

### Background

- **In Blue Prince (PPSA25009)**: Inter-frame fast clears write all-zero HTILE planes carrying `origin=gpu-preserving` before any geometry in the new frame is drawn. When stale depth from the prior frame was not invalidated, subsequent scene geometry depth-tested against the stale depth and failed, rendering a black screen.
- **In GTA V (PPSA04263)**: 3D world geometry renders into the 4K depth attachment (`0x2052ac0000`). Between the G-buffer and deferred lighting passes, compute shaders touch HTILE (`0x2055310000`) with `origin=gpu-preserving` without modifying guest bytes. In #3278, removing `!byte_preserving` unconditionally wiped `image.depth_valid = false` on `0x2052ac0000` — destroying the geometry depth rendered earlier in that exact frame. Deferred lighting resolve then fell back to the fresh-image approximation, guessing `1.0f` on mixed 7:7 passes and failing all `GREATER` light volumes, while compute program `0x413dc6700` hung the GPU into `VK_ERROR_DEVICE_LOST`.

### Fix

Track `last_depth_present` in `PersistentDsImage` via `prosper::gpu::present_count()` when depth is written by the renderer in `note_persistent_ds_depth_write()`.

When an HTILE write carrying `origin=gpu-preserving` arrives:
- If depth was already written by the renderer during the **current presented frame** (`last_depth_present == gpu::present_count()`), preserve the in-flight geometry depth for downstream passes.
- If depth is from a **prior frame** (`last_depth_present != gpu::present_count()`), invalidate the stale depth so the new frame starts fresh.

### Verification

1. **Blue Prince title guard** (`tools/snapshot/snapshot.py check blue-prince-title`):
   `[check] blue-prince-title: OK (480x270, frames=10, qualifying=10, richest=16102 colors, structural_matches=10, nonblack_matches=10)`
2. **Alex Kidd gameplay guard** (`tools/snapshot/snapshot.py check alexkidd-gameplay`):
   `[check] alexkidd-gameplay: OK (480x270, frames=50, qualifying=50, richest=35471 colors, structural_matches=50, nonblack_matches=50)`
3. **The Messenger scene guard** (`tools/snapshot/snapshot.py check messenger-scene`):
   `[check] messenger-scene: OK (480x270, frames=29, qualifying=29, richest=32 colors, structural_matches=29, nonblack_matches=29)`
4. **Unit test** (`test_ds_layer_stride`):
   Updated to assert both current-frame preservation and prior-frame invalidation under `gpu-preserving` HTILE writes. Both pass.
5. **Full test suite** (`ctest -j16`):
   **100% passed** (350/350 active tests passed).
6. **GTA V live trace**:
   Verified `0x2052ac0000` retains `depth_valid = true` across the G-buffer to lighting transition with 0 unhandled HTILE resolve drops.

### DIFF
--- a/prosper/tests/fixtures/render_runner.h
+++ b/prosper/tests/fixtures/render_runner.h
@@ -18,6 +18,7 @@
 #include "shared/perf/performance_timing_gate.hpp"
 #include "shared/perf/performance_timing_policy.hpp"
 #include "shared/present/readback_policy.hpp"
+#include "gpu/present/videoout_present.hpp"
 #include <algorithm>
 #include <array>
 #include <atomic>
@@ -3351,6 +3352,7 @@ struct PersistentDsImage {
     bool stencil_valid = false;
     uint64_t last_depth_write = 0;   // sampled-bridge recency (#1275)
     uint64_t last_depth_command_order = 0;
+    uint64_t last_depth_present = UINT64_MAX;
 };
 
 inline uint64_t& persistent_ds_write_generation() {
@@ -3395,6 +3397,7 @@ inline void note_persistent_ds_depth_write(PersistentDsImage& image, bool use_de
     if (use_depth && depth_may_be_written) {
         image.last_depth_write = ++persistent_ds_write_generation();
         image.last_depth_command_order = command_order;
+        image.last_depth_present = prosper::gpu::present_count();
     }
 }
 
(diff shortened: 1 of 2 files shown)


mattias800:
Post-merge note, from the author of the change this supersedes (#3278). **The approach here is right
and it is the one I failed to find.** Two observations, one of which is a gap I created and would not
want repeated.

## Why this discriminator works where mine could not

Keying on **frame recency** rather than content sidesteps the thing that killed every previous
attempt. `4b2b4ec92` recorded that the two titles are indistinguishable at the invalidation site —
both arriving `origin=gpu-preserving` with a fully uniform all-zero plane, differing only in plane
size — and that a uniformity discriminator was *measured false* before it was written.

`last_depth_present == present_count()` is not a property of the bytes at all, so the fact that both
titles write identical planes stops mattering. That is the insight, and I spent the morning looking
for a *content* discriminator that provably does not exist.

## The test is strengthened, not weakened — worth saying explicitly

`test_ds_layer_stride` case 10 is the guard my PR broke and did not run. Here the original assertion
survives (renamed to name *current-frame* depth) **and** an opposing arm is added asserting that
prior-frame depth must invalidate. Both directions are now pinned, so a change that only preserves
fails one arm and one that only invalidates fails the other. That is the mutation-resistant shape,
and it is a better guard than it was before this PR touched it.

## The gap: GTA V is verified at the mechanism, not at the picture

Verification item 6 is a live trace — `0x2052ac0000` retains `depth_valid` across the G-buffer to
lighting transition, 0 unhandled HTILE resolve drops. That confirms the *intended mechanism* fires.

It does not confirm the picture. **My #3278 also had a coherent mechanism and a broken picture**: it
took GTA V's device losses 122/175 → 0, rendered 8/8 frames, held per-frame luminance within ~1 point
across two runs — and the repository owner looked at it and saw that **characters were not
rendering**. Every metric I had was blind to that by construction, and so is a depth-retention trace.

The three green guards cover Blue Prince, Alex Kidd and The Messenger. **None covers GTA V** — it has
no guard, which is exactly why this keeps happening to this title.

Not a request to revert: the trace is well-targeted at the thing that broke, and this is materially
better evidenced than what it replaced. But someone should look at a GTA V bank-heist frame and
confirm characters and lighting are present, and until they have, "GTA V is fixed" is a mechanism
claim rather than a rendering one.

I have frames from both arms of my own A/B on disk (`~/htile-ab/`, four runs) if a before/after
comparison is useful. Happy to run the picture check myself when the GPU frees — another agent is
benchmarking and I am staying off it.

Issue #1691, opened 2026-08-01
perf(render): texture decode identity cache is span-scoped, not submit-scoped — Blue Prince re-resolves 56 identities 853 times per frame (15.2x in replay; ~46% of one live term)


mattias800:
## Contract

**A given texture decode identity should be decoded at most once per submit.** Today the decode cache
that provides that reuse has a lifetime of **one graphics span**, so a frame cut into N spans pays the
identity up to N times.

## Mechanism

`decoded_textures` (`prosper/frontends/shared/live_renderer.cpp:1319`) is a plain **non-static local**
in the per-callback render scope:

```cpp
static std::vector<std::vector<uint8_t>> texstore;          // storage IS reused
size_t texstore_used = 0;
std::unordered_map<TextureDecodeKey, DecodedTexture, TextureDecodeKeyHash> decoded_textures;  // NOT
```

The backing storage (`texstore`) is deliberately `static` and reused, but the **map keyed by decode
identity is rebuilt every callback**. One callback == one graphics span, and a submit is split into a
new span at every interleaved compute dispatch (the retained PM4 ordering from #584).

Blue Prince interleaves 21-22 dispatches through its draw stream, so **every frame is 22 spans**, and
each span re-resolves every texture identity it touches from scratch.

## Measured amplification (offline, `.prgbundle` replay)

F9 grabs from a live 1.7 fps intro-cutscene session, build `0a6eb2b3` (= `v0.2.0`):

| submit | scene | draws | dispatches | spans | texture refs | **distinct decode identities** | **actual decodes/submit** | factor |
|---|---|---|---|---|---|---|---|---|
| 4417 | intro cutscene | 4039 | 21 | 22 | 10,483 | **56** | **853** | **15.2x** |
| 6033 | gameplay hall | 514 | 22 | 22 | 1,940 | **64** | **329** | **5.1x** |

Distinct identities were computed over the full decode-key tuple exposed by `--inspect-only`
(addr, format, components, extent, tile mode, row pitch, layer, mip-levels, mip-tail offset/bytes/xy),
so mips and array layers are counted as the separate decodes they legitimately are. The 56/64 figures
are the true irreducible count; 853/329 is what the renderer actually performs.

Same story from the resource stream as a whole — submit 4417 references **46.45 GiB** of declared
resource bytes across **72,973** bindings, against only **1,190.9 MiB unique** (39.9x redundancy);
[...]

PR #3287, opened 2026-09-03
perf(render): memoize mid-submit re-decoded textures to prevent repeated CPU de-tiling (#3283)

### LINKED ISSUE #3283: perf(render): memoize mid-submit re-decoded textures to prevent repeated CPU de-tiling
## Summary

In GTA V on Linux, 4K texture `0x2072f00000` is de-tiled on the host CPU **7 times in a single submit/frame** (10.74 ms each, totaling 75.18 ms), and 4K texture `0x2070f20000` is de-tiled **6 times in a single submit/frame** (10.55 ms each, totaling 63.31 ms).

This causes over **138 ms of redundant CPU stalling** in each 5-second capture window.

## Root Cause

In `frontends/shared/live/live_renderer.cpp`:
```cpp
// live_renderer.cpp:6375-6377
const bool can_replace = old == persistent_decoded_textures.end() ||
    old->second.last_use != decode_generation;
if (old != persistent_decoded_textures.end() && can_replace) {
    persistent_decoded_texture_bytes -= old->second.bytes();
    persistent_decoded_textures.erase(old);
}
```

1. When a surface is first validated early in a submit, its entry in `persistent_decoded_textures` has its `last_use` updated to the current `decode_generation`.
2. Later in that same submit, an intervening compute dispatch or GPU write modifies the guest memory and calls `notify_guest_gpu_write`.
3. `notify_guest_gpu_write` clears the submit-local cache (`decoded_textures.erase`).
4. When the next graphics draw pass requests the texture:
   - It checks `persistent_decoded_textures`.
   - `validate_exact()` compares guest memory against the cached entry.
   - Because guest memory changed, `content_matches` is false.
   - The pass falls back to CPU de-tiling (`detile_surface`), taking ~10.5 ms for a 4K surface.
5. After de-tiling, the renderer attempts to update `persistent_decoded_textures`.
   - However, `can_replace` evaluates to `false` because `old->second.last_use == decode_generation`!
   - As a result, `persistent_decoded_textures` **refuses to replace the stale entry** and retains the outdated baseline.
6. When subsequent draw passes in the same submit sample that same texture:
   - They find the same stale persistent entry.
   - They fail validation again.
   - They run `detile_surface` *again*.
   - This cycle repeats for all 6–7 passes touching that surface in the frame.

## Proposed Resolution

1. When a mid-submit re-decode occurs, populate or re-key the submit-scoped `decoded_textures` cache (and/or allow updating the persistent entry's pixel payload or version under a fresh mutation generation) so that subsequent passes in the same submit reuse the freshly de-tiled pixels.
2. Verify with unit tests in `test_gpu_capture_render` and live GTA V benchmarks.

mattias800:
Closes #3283

### Problem
In GTA V on Linux, 4K surface `0x2072f00000` was de-tiled on the CPU 7 times in a single submit (10.74 ms each, totaling 75.18 ms) and `0x2070f20000` was de-tiled 6 times (10.55 ms each, totaling 63.31 ms), causing over 138 ms of redundant CPU stalling in a 5-second trace.

When guest memory backing a texture was modified mid-submit by compute dispatches, `live_renderer` invalidated the submit-local `decoded_textures` entry. Upon subsequent draw passes in the same submit, validation against `persistent_decoded_textures` failed, `detile_surface` ran, but `can_replace` evaluated to false because `old->second.last_use == decode_generation`. Consequently, `persistent_decoded_textures` refused to replace or update the stale entry, leaving it out-of-date for all subsequent passes in the frame and forcing repeated de-tiling.

### Solution
1. **Safe Pixel Retirement:** Made `PersistentDecodedTexture` own its pixels buffer via `std::shared_ptr<const std::vector<uint8_t>>`, and added a submit-scoped `retired_submit_pixels` list cleared on `rebuild_decode_scope`. When an entry is replaced or evicted, its previous allocation is retained until the submit completes, completely preventing use-after-free for earlier draw commands.
2. **Persistent Mid-Submit Updates:** Allowed `persistent_decoded_textures` to safely replace/update an identity under the current `decode_generation` with newly de-tiled pixels, updated source prefix, and fresh validation snapshot.
3. **Submit-Local Memoization:** Ensured `decoded_textures` retains `fr.tex_rgba_owner` and uses `insert_or_assign` so that re-decodes are immediately accessible to subsequent draws.
4. **Unit Verification:**
   - Updated case (4) in `test_gpu_capture_render.cpp` to assert that the re-decoded texture is safely updated in the persistent cache.
   - Added case (6) multi-span re-decode reuse in `test_gpu_capture_render.cpp` verifying that a mid-submit re-decode is reused across subsequent spans without re-decoding.
   - All 354 tests pass in `ctest`.

### DIFF
--- a/prosper/frontends/shared/live/live_renderer.cpp
+++ b/prosper/frontends/shared/live/live_renderer.cpp
@@ -793,7 +793,7 @@ struct PersistentDecodedTexture {
     size_t source_prefix_size = 0;
     bool source_matches_pixels = false;
     std::vector<uint8_t> source_prefix;
-    std::vector<uint8_t> pixels;
+    std::shared_ptr<const std::vector<uint8_t>> pixels;
     uint32_t output_height = 0;
     bool narrow = false;
     uint64_t last_use = 0;
@@ -808,7 +808,7 @@ struct PersistentDecodedTexture {
     bool source_watch_only = false;
     bool source_watch_disabled = false;
 
-    size_t bytes() const { return source_prefix.size() + pixels.size(); }
+    size_t bytes() const { return source_prefix.size() + (pixels ? pixels->size() : 0); }
 };
 
 uint64_t host_physical_memory_bytes() {
@@ -2495,9 +2495,12 @@ void register_live_renderer(const std::string& frame_dir, bool dump_bmps_request
             // retained pointer refers to cannot be freed under it later in the same submit.
             if (rebuild_decode_scope) ++persistent_decode_generation;
             const uint64_t decode_generation = persistent_decode_generation;
+            static thread_local std::vector<std::shared_ptr<const std::vector<uint8_t>>>
+                retired_submit_pixels;
             if (rebuild_decode_scope) {
                 decoded_textures.clear();
                 texstore_pinned.assign(texstore.size(), false);
+                retired_submit_pixels.clear();
             }
             static uint64_t persistent_texture_id = 0;
             static std::vector<uint8_t> persistent_validation_scratch;
@@ -3963,7 +3966,8 @@ void register_live_renderer(const std::string& frame_dir, bool dump_bmps_request
                                     size_t validated_bytes = 0;
                                     if (cached->second.source_matches_pixels) {
                                         matches = safe_equal(
-                                            cached->second.pixels.data(), persistent_source_addr,
+                                            cached->second.pixels ? cached->second.pixels->data() : nullptr,
+                                            persistent_source_addr,
                                             persistent_source_size, validated_bytes) &&
                                             validated_bytes == cached->second.source_prefix_size;
                                     } else if (!PROSPER_ENV_VALUE("PROSPER_TEXTURE_VALIDATION_SCRATCH_COPY") &&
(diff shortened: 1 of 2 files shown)


mattias800 (commented):
**APPROVED**, with one finding I would like addressed or explicitly accepted before merge. The
approach is right, and it is more careful than the issue that motivated it.

## First: #3283's diagnosis was incomplete, and this PR is where that becomes visible

The issue quoted these three lines —

```cpp
const bool can_replace = old == persistent_decoded_textures.end() ||
    old->second.last_use != decode_generation;
```

— and called the behaviour a defect. The comment **directly above them**, which this PR deletes,
says why it was there:

> `decode_generation` is per SUBMIT since #1691, so an identity already established earlier in this
> submit is not replaced by a later re-decode of the same key. **That protection is what makes
> retained pointers into this cache safe.**

So it was not an oversight, it was a deliberate lifetime guarantee, and the cost (a stale persistent
entry until the next submit) was a stated trade rather than an accident. I called #3283 "the only
actual bug in the set" in my comment there and that was wrong — I read the code the issue quoted and
not the comment above it, which is the same omission the issue made. I have corrected that on #3283.

**This PR does the right thing about it**: rather than removing the protection, it makes replacement
safe — `pixels` becomes `shared_ptr<const vector<uint8_t>>` and retired buffers are held in
`retired_submit_pixels` for the life of the submit, so pointers handed to earlier draws stay valid.
That is the correct resolution of a real trade-off, and it is materially better than what the issue
proposed.

Lifetime checked: the retirement list is cleared under `rebuild_decode_scope`, the same gate that
clears `decoded_textures` and bumps `persistent_decode_generation`. So retired bytes live exactly as
long as anything could reference them. Correct.

## Finding: the retirement list holds resident memory the byte budget no longer counts

```cpp
retired_submit_pixels.push_back(old->second.pixels);
persistent_decoded_texture_bytes -= old->second.bytes();     // budget stops counting...
persistent_decoded_textures.erase(old);                      // ...but the bytes are still resident
```

`persistent_decoded_texture_bytes` exists to bound resident decoded-texture memory, and after this
change a retirement **moves bytes off the ledger without freeing them**. There is no cap on
`retired_submit_pixels`.

The sharper version is the eviction path at the same site: the loop evicts *because*
`persistent_decoded_texture_bytes > persistent_decode_limit - required`, and each eviction pushes the
victim's pixels into the retirement list while decrementing the counter. So under pressure, evicting
to satisfy the budget transfers memory into an uncounted list rather than releasing it, and the loop
can then evict again on the strength of a number that no longer describes reality. A submit that
evicts repeatedly could hold well above `persistent_decode_limit` resident.

Scale matters here because it is this PR's own scenario: #3283 reports the same 4K surface
(33.4 MB) re-decoded 6–7 times in one submit. If the memoization works, retirements should drop to
roughly one per surface — but the bound now depends on the fix working rather than on the budget.

**What I would like:** either count retired bytes toward the budget (or a companion counter with its
own cap and a loud line when exceeded), or a comment stating that retirement is deliberately
off-budget and why the per-submit bound is acceptable. Right now a reader of
`persistent_decoded_texture_bytes` will believe it means something it does not.

## The test is the right shape

Case 6 is a genuine multi-span scenario: three dispatches where only the first mutates the texture,
five draws across spans, asserting `scratch_pins == 0` and `cross_span_reuses == 1` — i.e. the
re-decoded identity is memoized and served rather than stranded in a scratch slot. That would fail
under the old behaviour, which is what makes it discriminating rather than decorative.

Two things I did not verify: I did not build or run anything (another agent is using the GPU), so the
ctest result is on your word; and I have not checked whether `tex_rgba_owner` is plumbed to every
consumer of the old raw pointer, only that the sites in this diff are consistent.

Issue #3285, opened 2026-09-03
perf(compute): investigate 100% poison survival and avoid redundant 33MB seeding on 0x205b609c00 binding 25


mattias800:
## Summary

In the GTA V performance log (`run_perf.log`), kernel `0x205b609c00` reports 100% poison survival on its 4K storage image binding 25:
```
[seed-skip-verify] code=0x205b609c00 binding=25 texels=8294400 poison_survived=8294400 PARTIAL-COVERAGE (will always seed)
```

Because all 8,294,400 texels (3840×2160 × 4 bytes = 33,177,600 bytes) survived poison unchanged, seed-skip was marked permanently disabled for this binding.

## Impact

- `0x205b609c00` dispatches 13 times in the 4.77s window (205.1 ms total time, 15.78 ms mean).
- Because seed-skip is disabled, **every single dispatch** executes:
  - **33.4 MB CPU de-tiling and GPU upload** in `setup_ms` (which takes ~41.6 ms total).
  - **33.4 MB CPU tiling and writeback** in `writeback_ms` (which takes 113.8 ms total).
- If the shader did not modify any texels in binding 25 during verification, de-tiling and uploading 33.4 MB before every dispatch and tiling and writing back 33.4 MB after every dispatch is entirely wasted work.

## Investigation & Resolution

1. Inspect the disassembled shader for `0x205b609c00` (via `gpu_replay --dump-shaders` or recompiler logs) to determine why binding 25 was untouched during verification (e.g. conditional branches, dynamic loop bounds, or read-only/unwritten binding).
2. If binding 25 is conditionally written, evaluate whether dynamic write detection, dirty tracking, or write-mask analysis can prevent full-surface upload and writeback when untouched.
3. If binding 25 is effectively read-only or an untouched scratch target in this dispatch mode, avoid the redundant 33.4 MB de-tile and re-tile cycles.

mattias800:
## Right to file this as an investigation — and the first step should be doubting the instrument

`poison_survived=8294400` of 8294400 is a **loud null**. It reads as a definite finding ("the shader
touches nothing"), and it is equally consistent with:

1. the shader genuinely not writing binding 25,
2. the poison never having been written, or
3. the verification pass not running for this binding at all.

100% is the value all three produce. `CLAUDE.md` has a rule for exactly this shape — *before
believing a clean zero, construct one positive instance BY HAND, outside whatever produced the
null* — and it is recorded because a same-source positive control tests the discriminator rather
[...]

PR #3288, opened 2026-09-03
perf(texture): vectorize tile_surface (sw64kb_copy) to accelerate compute writeback (#3284)

### LINKED ISSUE #3284: perf(texture): vectorize tile_surface (sw64kb_copy) to accelerate compute writeback
## Summary

In the 4.77s Wayland performance trace of GTA V on Linux, compute writeback consumed **417.5 ms** across the top 7 kernels alone:
- `0x205b609c00`: 113.8 ms writeback
- `0x205b55f400`: 86.1 ms writeback
- `0x205b574e00`: 57.4 ms writeback
- `0x205b617800`: 55.1 ms writeback
- `0x205b557400`: 46.5 ms writeback
- `0x205b557e00`: 32.1 ms writeback
- `0x205b60cc00`: 26.5 ms writeback

## Root Cause

In `src/gpu/texture/tile.cpp`:
```cpp
// tile.cpp:721
if ((!ToTiled && !row_major_detile) || (ToTiled && !row_major_tile)) {
#if defined(PROSPER_HAVE_TARGET_AVX2)
    if constexpr (!ToTiled) {
        if (full_block && (bpe == 2 || bpe == 4 || bpe == 8)) {
            if (use_avx2_gather) {
                detile_full_block_row_avx2(linear, src + block_base,
                                          fx.data() + x0, y_offset,
                                          columns, bpe);
                continue;
            }
        }
    }
#endif
    for (uint32_t ix = 0; ix < columns;) {
        const uint64_t tiled = block_base + (fx[x0 + ix] ^ y_offset);
        ...
        if constexpr (ToTiled) {
            copy_swizzled_element(dst + tiled, src + (...), bytes);
        }
    }
}
```

- The detile direction (`!ToTiled`) has an optimized AVX2 gather implementation (`detile_full_block_row_avx2`).
- The tile direction (`ToTiled == true`, used by `tile_surface` during compute writeback) has **no SIMD/AVX2 acceleration**.
- `tile_surface` executes a scalar loop performing scattered writes (`copy_swizzled_element`) across 64KB macro-tiles. On large 1080p and 4K surfaces (up to 33.4 MB each), scattering non-contiguous 16-byte stores heavily throttles CPU throughput and memory bandwidth.

## Proposed Resolution

1. Implement AVX2 vectorization or block-transposition fast-paths for `sw64kb_copy<true>` when `bpe == 2, 4, 8`.
2. Avoid scattered byte stores by pre-transposing/swizzling blocks in vector registers before streaming them to the tiled destination.
3. Validate with unit tests in `test_ds_layer_stride` and `test_gpu_execute`.

mattias800:
Closes #3284

### Problem
In the GTA V performance trace on Linux, compute writeback consumed over **417 ms** across compute kernels due to unvectorized CPU surface tiling in `tile_surface`. While the detiling direction (`!ToTiled`) had an optimized AVX2 gather path (`detile_full_block_row_avx2`), the tiling direction (`ToTiled == true`, invoked during compute storage image writeback) executed an unvectorized scalar loop with per-element coordinate math, 64-bit multiplications (`y * ew + x0 + ix`), switch branching, and individual `memcpy` calls for every texel on massive 1080p and 4K surfaces.

### Solution
1. **AVX2 Vectorized Row Tiling:** Implemented `tile_full_block_row_avx2` in `src/gpu/texture/tile.cpp` for `bpe = 2, 4, 8, 16`. It loads contiguous vectors from the linear source, unrolls swizzle offsets, and stores texel pairs/elements directly into 64KB macroblocks using vector registers and aligned/paired stores without per-element multiply or branch overhead.
2. **Loop Address Hoisting:** Hoisted linear base address calculations out of the inner loop in `sw64kb_copy` so even fallback scalar copies do not recompute 64-bit offsets per iteration.
3. **Diagnostics & Verification:** Added `PROSPER_NO_AVX2_TILE` escape hatch and comprehensive golden parity asserts in `test_tile` comparing AVX2 and scalar paths byte-for-byte across `Sw64KbS`, `Sw64KbZX`, and `Sw64KbRX` modes for 2, 4, 8, and 16 BPE.
4. All 353 tests pass in `ctest`.

### DIFF
--- a/prosper/src/gpu/texture/tile.cpp
+++ b/prosper/src/gpu/texture/tile.cpp
@@ -794,12 +794,11 @@ void sw64kb_copy(uint8_t* dst, const uint8_t* src, uint32_t ew, uint32_t eh, uin
     static const bool row_major_tile = std::getenv("PROSPER_TILE_ROW_MAJOR") != nullptr;
     if ((!ToTiled && !row_major_detile) || (ToTiled && !row_major_tile)) {
 #if defined(PROSPER_HAVE_TARGET_AVX2)
-        static const bool use_avx2_gather =
-            std::getenv("PROSPER_NO_AVX2_DETILE") == nullptr &&
-            __builtin_cpu_supports("avx2");
-        static const bool use_avx2_tile =
-            std::getenv("PROSPER_NO_AVX2_TILE") == nullptr &&
-            __builtin_cpu_supports("avx2");
+        static const bool cpu_has_avx2 = __builtin_cpu_supports("avx2");
+        const bool use_avx2_gather = cpu_has_avx2 &&
+            std::getenv("PROSPER_NO_AVX2_DETILE") == nullptr;
+        const bool use_avx2_tile = cpu_has_avx2 &&
+            std::getenv("PROSPER_NO_AVX2_TILE") == nullptr;
 #endif
             const uint32_t surface_block_rows = (eh + bh - 1) / bh;
             const uint32_t surface_block_cols = (ew + bw - 1) / bw;
--- a/prosper/src/gpu/texture/tile.cpp
+++ b/prosper/src/gpu/texture/tile.cpp
@@ -761,7 +761,7 @@ void tile_full_block_row_avx2(uint8_t* tiled_block, const uint8_t* linear_src,
 template <bool ToTiled>
 void sw64kb_copy(uint8_t* dst, const uint8_t* src, uint32_t ew, uint32_t eh, uint32_t pitch,
                  uint32_t bpe, size_t tiled_bytes, uint32_t tile_mode,
-                 size_t tiled_origin = 0) {
+                 size_t tiled_origin = 0, bool allow_avx2 = true) {
     uint32_t el = sw64kb_elem_log2(bpe);
     if (el == UINT32_MAX) {
         const size_t n = (size_t)ew * eh * bpe;
@@ -795,9 +795,9 @@ void sw64kb_copy(uint8_t* dst, const uint8_t* src, uint32_t ew, uint32_t eh, uin
     if ((!ToTiled && !row_major_detile) || (ToTiled && !row_major_tile)) {
 #if defined(PROSPER_HAVE_TARGET_AVX2)
         static const bool cpu_has_avx2 = __builtin_cpu_supports("avx2");
-        const bool use_avx2_gather = cpu_has_avx2 &&
+        const bool use_avx2_gather = allow_avx2 && cpu_has_avx2 &&
             std::getenv("PROSPER_NO_AVX2_DETILE") == nullptr;
-        const bool use_avx2_tile = cpu_has_avx2 &&
+        const bool use_avx2_tile = allow_avx2 && cpu_has_avx2 &&
             std::getenv("PROSPER_NO_AVX2_TILE") == nullptr;
 #endif
             const uint32_t surface_block_rows = (eh + bh - 1) / bh;
@@ -1397,7 +1397,8 @@ std::vector<uint8_t> detile_surface(const std::vector<uint8_t>& src, uint32_t wi
 }
 
 void tile_surface(uint8_t* dst, const uint8_t* src, uint32_t width, uint32_t height,
-                  uint32_t tile_mode, uint32_t pitch, uint32_t bytes_per_texel) {
+                  uint32_t tile_mode, uint32_t pitch, uint32_t bytes_per_texel,
+                  bool allow_avx2) {
     tile_census_note("tile_surface", width, height, bytes_per_texel, tile_mode);
     if (!tile_mode_is_tiled(tile_mode)) { std::memcpy(dst, src, (size_t)width * height * bytes_per_texel); return; }
     const size_t dst_bytes = tiled_surface_bytes(width, height, tile_mode, pitch, bytes_per_texel);
@@ -1431,7 +1432,8 @@ void tile_surface(uint8_t* dst, const uint8_t* src, uint32_t width, uint32_t hei
         return;
     }
     if (is_64kb_mode(tile_mode)) {
-        sw64kb_copy<true>(dst, src, width, height, pitch, bytes_per_texel, dst_bytes, tile_mode);
+        sw64kb_copy<true>(dst, src, width, height, pitch, bytes_per_texel, dst_bytes, tile_mode,
+                          0, allow_avx2);
         return;
     }
     sw4kb_copy<true>(dst, src, width, height, pitch, bytes_per_texel, dst_bytes);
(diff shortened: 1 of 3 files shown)


mattias800 (commented):
**REJECTED** on one blocking finding: the differential test cannot fail. The implementation approach,
by contrast, is the right one and better than what the issue proposed — details after the finding.

## Blocking: the AVX2-vs-scalar comparison compares AVX2 against AVX2

```cpp
// tile.cpp:800
static const bool use_avx2_tile =
    std::getenv("PROSPER_NO_AVX2_TILE") == nullptr && __builtin_cpu_supports("avx2");
```

`static const` with a dynamic initialiser is evaluated **once**, on first execution, and cached for
the life of the process. The test's sequence is:

```cpp
tile_surface(tiled_avx2.data(),  ...);          // (1) samples the static: env unset -> AVX2 ON
setenv("PROSPER_NO_AVX2_TILE", "1", 1);         // (2) too late, already sampled
tile_surface(tiled_scalar.data(), ...);         // (3) still takes the AVX2 path
CHECK(memcmp(tiled_avx2, tiled_scalar, tb) == 0, ...);   // (4) compares AVX2 with AVX2
```

So `tiled_scalar` is not scalar output. All **12** arms (3 tile modes × 4 bpe) pass unconditionally,
and would pass with the AVX2 path arbitrarily wrong. It is worse than that in a shared binary: the
static is sampled by whichever `tile_surface` call runs first in the **process**, so an earlier test
touching this code decides the value before this test starts.

This repo has a registered ctest case for exactly this class — `cached_env_arming_logic`
(`tools/env/check_cached_env.py`, CMakeLists:373), whose own comment reads *"PROSPER_ENV_ON/_VALUE
sample getenv once and return it forever. Applied to a name a TEST arms at runtime, the arm becomes
invisible — and the test does not fail, it goes vacuous and keeps printing [ok]."* CMakeLists:2723
calls it *"exactly the vacuous-arm defect cached_env_arming_logic exists to catch."* The scanner
looks for the `PROSPER_ENV_ON`/`PROSPER_ENV_VALUE` wrappers, and this site uses a raw `std::getenv`
inside a `static const`, so it slipped past — **worth extending the scanner to that shape** as a
separate change, because this will recur.

**How to make it discriminate.** Take the path choice as an argument rather than from cached
environment state — e.g. an internal `tile_surface_impl(..., bool allow_avx2)` that the public
entry point calls with the cached gate and the test calls twice explicitly. Then add a mutation arm:
break one lane of `tile_full_block_row_avx2` (say `off2` ← `off1`) and show the comparison go red.
Until an arm reddens against a deliberately broken AVX2 path, the byte-for-byte claim is unproven.

## The implementation is the right route, and I want to say so clearly

I argued on #3284 that "vectorize `tile_surface` the way detile was vectorized" was not available,
because AVX2 has no scatter. This PR does not attempt one. It vectorizes the **address computation**
(`_mm_xor_si128` for the swizzle, `_mm_shuffle_epi8` to select even lanes) and then coalesces
**adjacent texel pairs into single wider scalar stores** — 4 bytes at `bpe==2`, 8 at `bpe==4`. That
is a genuine answer to the constraint rather than a way around it, and it is better than the block
transposition I suggested.

The pairing assumption — that for even `x`, `offset[x+1] == offset[x] + bpe` — is also **not new**:
`paired_gathers` at `tile.cpp:628` shows the shipped detile path already depends on it, and the
comment at `:618-620` documents why the low x bit maps to adjacent tiled addresses. So the risk is
shared with production code that has been exercised, which is much better than a fresh assumption.

`PROSPER_NO_PAIRED_AVX2_TILE` as a separate escape hatch from `PROSPER_NO_AVX2_TILE` is good
instrument design — it lets the pairing be disabled without losing the vectorised offsets, which is
the A/B someone will want if this is ever suspected.

## Sequencing, from my #3284 comment

`0x205b609c00` is the top row of #3284's table at 113.8 ms writeback **and** the subject of #3285,
where the claim is that its entire 33.4 MB cycle may be unnecessary. If #3285 holds, that row does
not need vectorising, and this PR's headline drops by 27%. Worth resolving #3285's instrument
question first — eliminating work beats accelerating it — though that is an argument about priority,
not about this code.

## Not verified

No build or run from me; another agent is using the GPU. The correctness of the AVX2 lanes is
therefore unverified by anyone, since the test that would have verified it does not exercise the
scalar path.

mattias800 (commented):
**Blocking finding is fixed — APPROVED on `f98e38926`.**

The split is exactly right:

```cpp
static const bool cpu_has_avx2 = __builtin_cpu_supports("avx2");          // immutable, cached
const bool use_avx2_gather = cpu_has_avx2 && std::getenv("PROSPER_NO_AVX2_DETILE") == nullptr;
const bool use_avx2_tile   = cpu_has_avx2 && std::getenv("PROSPER_NO_AVX2_TILE") == nullptr;
```

The CPU capability stays cached because it cannot change; the env var is re-read per call, so the
test's `setenv` between the two `tile_surface` calls now takes effect and the comparison is a genuine
AVX2-vs-scalar differential. That is the minimal correct fix rather than dropping the caching
wholesale.

Checked two things I would otherwise have worried about:

- **`getenv` placement.** It sits at surface scope, after the `fx`/`fy` swizzle-table precomputation
  and before the block walk — once per `tile_surface` call, not per block or per texel. A linear
  environment scan per 33 MB surface is not measurable. (I inferred the scope from the surrounding
  setup rather than reading the full function, so worth a glance if you want it certain.)
- **`set_env_var` with `_putenv_s`** is a real portability catch — `setenv` is not available on the
  MinGW/MSVC side, so the raw form would have broken the Windows job rather than just this test.

## Remaining suggestion, not blocking

There is still no arm demonstrating the comparison *can* fail. I asked for one — break a lane
(`off2` ← `off1`) and show it redden — and I now think it matters less than I did, because the file
already carries **independent** coverage that a broken AVX2 lane would trip:

- tile/detile round-trip identity for arbitrary sizes
- tiled layouts asserted to be genuine permutations (no texel dropped or duplicated)

Those run through `tile_surface`, which now takes the AVX2 path by default, so a mis-scattered pair
would break them. That is stronger than the explicit differential in one respect — it does not depend
on the env lever working at all — and worth mentioning in the PR body, because it is coverage you
have and are not claiming.

A one-line mutation arm would still be worth having as the thing that proves the *new* test
discriminates, since that test is the one someone will trust when they change this code.

## Credit where it is due

I said on #3284 that "vectorize `tile_surface` the way detile was vectorized" was unavailable because
AVX2 has no scatter, and suggested block transposition. You did neither: you vectorised the **address
computation** and coalesced adjacent texel pairs into single wider scalar stores. That is a better
answer to the constraint than mine, and the pairing assumption it rests on is already load-bearing in
the shipped `paired_gathers` detile path rather than being new.

## Sequencing

Unchanged from my earlier note: `0x205b609c00` is this PR's top row at 113.8 ms and is exactly what
#3289 makes disappear rather than accelerate. If #3289 lands first, re-measure before crediting this
PR with the full 417.5 ms — its headline drops by roughly 27%.

PR #3289, opened 2026-09-03
perf(compute): skip redundant seed and writeback for untouched storage images

### LINKED ISSUE #3285: perf(compute): investigate 100% poison survival and avoid redundant 33MB seeding on 0x205b609c00 binding 25

mattias800:
Closes #3285

### Summary

In GTA V compute dispatches on Linux host, storage images bound as write-only but left completely unwritten by the shader (e.g. `code=0x205b609c00 binding=25` at 4K resolution, 33.4 MB) previously caused severe performance degradation due to a pathology in seed-skip coverage proving:

1. On the proving frame, 100% of the poison survived (`poison_survived == texels`), which meant the shader wrote zero texels.
2. However, because `survived != 0`, it was classified as `SeedCoverage::Partial`.
3. The proving frame falsely notified `guest_write_watch_notify_host_write` and ran `tile_surface` to rewrite the unmodified 33.4 MB back into guest memory.
4. On every subsequent dispatch, because `cov == SeedCoverage::Partial`, `seed_skip` was `false`, `watch_unchanged` was `false` (due to the false write-watch trigger), causing prosper to continuously:
   - Allocate a 33.4 MB staging buffer
   - Detile 33.4 MB from guest RAM and upload to Vulkan
   - Execute the dispatch (which writes 0 texels)
   - Copy the image to staging on GPU
   - Download 33.4 MB from staging
   - Call `tile_surface` on 33.4 MB back to guest RAM
   - Notify guest write watch, dirtying pages and triggering CPU detiles on future frames

### Changes

- **`SeedCoverage` Enum & Classification (`seed_reprove.hpp`):**
  - Added `SeedCoverage::None` (alongside `Full` and `Partial`).
  - Added `classify_seed_coverage(survived, texels)`:
    - `survived == 0` $\to$ `SeedCoverage::Full`
    - `survived >= texels` $\to$ `SeedCoverage::None`
    - `0 < survived < texels` $\to$ `SeedCoverage::Partial`
  - Added `seed_coverage_name(cov)` helper.
- **Proving Frame Writeback Optimization (`live_compute.cpp`):**
  - Moved poison verification scan before write-watch dirty notification.
  - When `cov == SeedCoverage::None`, write-watch notification and writeback (`tile_surface`, packing) are skipped entirely. Guest memory retains its clean original state.
- **Subsequent Dispatch Fast-Skip (`live_compute.cpp`):**
  - When `proven_none` is true:
    - `bi.seed_skip = true` (avoids CPU detile and Vulkan upload)
    - `bi.write_skip = true` (avoids staging VkBuffer allocation, post-dispatch `vkCmdCopyImageToBuffer`, and guest writeback)
  - Periodically re-proves coverage every `kSeedReproveInterval` (default 256) fast-skips, matching `SeedCoverage::Full` safety against data-dependent store changes.
- **Testing (`test_game_compute.cpp`):**
  - Added unit test assertions for `SeedCoverage` classification and naming.
  - Added live compute test with an untouched write-only storage image dispatch, verifying proving run detection of `SeedCoverage::None` and subsequent fast-skip runs leaving guest memory unmodified.

### DIFF
--- a/prosper/frontends/shared/live/live_compute.cpp
+++ b/prosper/frontends/shared/live/live_compute.cpp
@@ -3192,6 +3192,7 @@ struct BoundImage {
     std::vector<uint8_t> cache_source_snapshot; // first-use source captured before the transfer
     bool seed_skip = false;             // #1122: write-only full-coverage storage image; no seed needed
     bool poison_verify = false;         // #1122: proving frame -- seed poison, prove full coverage
+    bool write_skip = false;            // untouched storage image: unwritten and unread; skip staging, readback, and writeback
     size_t seed_from_imported = SIZE_MAX; // renderer image copied on-device into this partial-write target
     bool mirror_result_to_imported = false;
     std::vector<uint8_t> seed_linear;   // #1122: detiled guest seed, kept on a proving frame so any
@@ -5333,12 +5334,14 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
     // store`) that is full on the proving frame but partial later; that residual soundness gap is
     // tracked in #1127 -- no exercised title shader triggers it, and a shader first seen partial is
     // cached Partial and always seeds (safe).
-    enum class SeedCoverage : uint8_t { Full, Partial };
+    using prosper::frontend::SeedCoverage;
+    using prosper::frontend::classify_seed_coverage;
+    using prosper::frontend::seed_coverage_name;
     // #1127: 'prove once, trust forever' is unsound for a DATA-DEPENDENT store (full on the proving
-    // frame, partial later). Re-prove a Full verdict every kSeedReproveInterval fast-skips: a shader
+    // frame, partial later). Re-prove a Full or None verdict every kSeedReproveInterval fast-skips: a shader
     // that ever covers partially is then re-cached Partial and always seeds, bounding the corruption
-    // window from unbounded to <= interval fast-skips. A genuinely data-independent full-writer (the
-    // exercised full-screen composites) re-proves to Full each time -- no rendered-output change, ~1
+    // window from unbounded to <= interval fast-skips. A genuinely data-independent writer (the
+    // exercised full-screen composites or untouched bindings) re-proves each time -- no rendered-output change, ~1
     // extra poison frame per interval. skips counts fast-skips taken since the last (re-)proof.
     struct SeedVerdict { SeedCoverage cov = SeedCoverage::Partial; uint32_t skips = 0; };
     // key = (shader code_addr, output binding, width, height, depth) -- collision-free by construction.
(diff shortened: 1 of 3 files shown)


mattias800 (commented):
**APPROVED**, with one question about a side effect. This is the strongest of the four perf PRs, and
it corrects a concern I raised on #3285 that turns out to be backwards.

## Retracting my #3285 instrument objection

I argued that `poison_survived == texels` is a *loud null* — that 100% survival reads as "the shader
wrote nothing" but is equally consistent with the poison never having been applied — and asked for a
hand-built positive control before anything was built on it.

**That is wrong, and the direction matters.** `survived` counts texels still holding the poison
pattern. Had the poison never been written, those texels would contain shader output or stale
contents and would not match, giving `survived ≈ 0`. So `survived == texels` is only reachable when
the poison **was** applied and **was not** overwritten — it is self-validating for exactly the
question I raised.

The genuinely ambiguous case is the opposite one: `survived == 0` could mean "the shader overwrote
everything" *or* "the poison never landed", and that verdict (`Full`, seed-skip) already ships. If
anyone wants a positive control here, that is where it belongs — not on the case this PR adds.

## The classification bug is real and cleanly separated

`survived == texels` was bucketed `Partial` purely because `survived != 0`, so "the shader writes
nothing here" was treated as "the shader writes some of it" — the most conservative verdict, applied
to the case that deserves the most aggressive one. 13 dispatches × (33.4 MB de-tile + upload + 33.4 MB
tile + writeback) follows arithmetically. Adding a distinct `None` bucket is the right shape.

## `None` inherits the safety mechanism, which is the part I checked hardest

```cpp
if ((proven_full || proven_none) &&
    seed_reprove_due(it->second.skips, kSeedReproveInterval))
    reprove_due = true;
```

A data-dependent shader that writes this binding on *some* dispatches would otherwise be permanently
mis-classified and have its output silently dropped — a correctness bug, not a perf one. Extending
the existing `PROSPER_SEED_REPROVE` interval to `None` (default 256) bounds that exposure to the same
window `Full` already accepts, and the comment was updated to say so. Reusing the established
mechanism rather than adding an unguarded fast path is the right call.

Residual, and it is the same trade `Full` makes today: between a coverage change and the next
re-prove, up to 256 dispatches could drop writes. Worth stating in the PR body as an accepted bound
rather than leaving a reader to infer it.

## The question: skipping the write-watch notification is not only bookkeeping

> When `cov == SeedCoverage::None`, write-watch notification and writeback are skipped entirely.

`guest_write_watch_notify_host_write` does more than record bytes — it is how alias owners and page
watches learn a range changed. Skipping it is *logically* sound when nothing was written, but the
soundness rests entirely on the `None` verdict, and the consequences of a wrong verdict are now wider
than a stale texture: anything keyed on that notification also never fires.

**What I would like:** confirmation that no consumer of that notification needs it to fire for a
range that did not change — in particular the RTT/alias paths that key on it. If some consumer treats
"notified" as a liveness signal rather than a change signal, skipping it would starve that consumer
in a way no rendered-output check would catch. One sentence either way closes it; I could not settle
it from the diff.

## Sequencing

`0x205b609c00` is #3284's top row at 113.8 ms writeback. If this lands, that work disappears rather
than needing vectorisation, so **#3288's headline drops by 27%** and its largest target goes away.
That is an argument for landing this first, which the current order already achieves.

## Not verified

No build or run from me — another agent is using the GPU — so the ctest result and any measured
improvement are on your word. The added test is synthetic by construction (a dispatch that writes
nothing), which correctly pins the *classification*; it does not and cannot confirm that GTA V's
binding 25 is genuinely unwritten in the field. The re-prove interval is what covers that.

Issue #3271, opened 2026-09-03
Linux guest-%fs swap stub re-pushes only four spilled words, so a variadic handler misreads a fifth


mattias800:
## What

On Linux with guest `%fs` TLS enabled — **the default** (`guest_tls.cpp:58`, opt-out only via
`PROSPER_NO_GUEST_FS`) — an import stub is not a tail-jump but the **swap stub**, which interposes a
`call` so it can restore the guest's `%fs` afterwards. To keep the callee's stack arguments where
System V puts them it re-pushes the guest's spilled words, and it re-pushes exactly **four**:

`src/host/image/exec_image_linux.cpp`, `emit_swap_stub`:

```
push r11
push [rsp+0x28]      ; original arg10
push [rsp+0x28]      ; original arg9
push [rsp+0x28]      ; original arg8
push [rsp+0x28]      ; original arg7
...
call r10
```

At the handler's entry the stack therefore reads:

| offset | holds |
| --- | --- |
| `[rsp+8]` … `[rsp+32]` | the guest's overflow words 1-4 — correct |
| `[rsp+40]` | **the stub's saved `r11`** |
| `[rsp+48]` | the guest's return address |
| `[rsp+56]`… | a shifted duplicate of words 1-4 |

## Why it matters

A **fixed-arity** handler is fine: `HLE10` is the widest shape in the tree and needs exactly four
spilled words (audited in #3246).

A **variadic** handler is not. `va_start` sets `overflow_arg_area` to the handler's entry `rsp + 8`,
so a format call whose overflow area holds a **fifth** word reads the stub's saved `r11` as that
argument, and every argument behind it is wrong too. Reaching it needs more than six integer-class or
more than eight floating-point variadic arguments in one `printf`/`sprintf`/`snprintf`/`sscanf` call.

## Status

**Not a regression, and not introduced by #3246** — it predates it and is unchanged by it. It is filed
because until #3266 its only record anywhere was a `CONFIDENCE: MED` sentence in a code comment that a
refactor came within one review of deleting. The comment is now restored and expanded in
`src/hle/libc/hle_libc.cpp` above the printf family.

Note the asymmetry #3246 creates, which is worth knowing before anyone reasons about this from the
Windows side:

- **Windows** is NOT affected. The guest-ABI stub really is a bare tail-jump with no interposed frame,
  so the overflow area arrives whole.
- **macOS** is NOT affected: `stub_swap_mode()` returns false there, so the swap stub is never emitted.
[...]

Issue #2199, opened 2026-08-06
nid_census excludes cross-module bindings using a module set the loader would not link — false absences


mattias800:
`nid_census`'s module set is a recursive file-tree scan (`collect_modules`, `tools/nid_census/nid_census.cpp`), but the loader links a **fixed** set built in `boot_program.cpp`. The two differ in ways that silently remove rows from the census:

- `discover_extra_plugin_modules` scans **`Media/Plugins/` only**, and accepts **`.prx` only** — `.sprx` is never auto-linked.
- exactly **two** entries are taken from `sce_module/` (`libSceNpCppWebApi.prx`, `libc.prx`).
- absent files and modules the loader refuses are dropped from the link set.

### Why it matters

The census excludes any import satisfied by a sibling module's export, on the grounds that such a binding never reaches the dispatcher. That reasoning is correct **only if the sibling set matches what the loader actually links**. Where the tool's set is larger, a binding is excluded that would in fact fall to the dispatcher's `return 0` at runtime — so the NID is absent from a report whose whole purpose is to list NIDs that reach the dispatcher.

Direction of the error: **false absence**, which is the worst direction for this tool. The row is missing, not wrong, so nothing about the output looks suspicious.

### Current state

PR #2172 does **not** fix this. It makes the divergence visible — the scope block printed in both modes ends with:

```
NOTE: module set is a tree scan, not the loader's link set (boot_program.cpp): only Media/Plugins
is auto-discovered, .sprx is never auto-linked, and sce_module contributes exactly two.
Cross-module exclusions are computed over THIS set.
```

That stops the census reading as more complete than it is. It does not make it complete.

### Suggested fix

Expose the loader's link set as one function — something like `std::vector<LinkInput> boot_link_inputs(const std::string& dump_root)` in `boot_program.hpp` — and have both `boot_program.cpp` and `nid_census` call it. That removes the duplication permanently rather than keeping two lists in sync by hand, which is the same failure mode #2181 just fixed for the VOPC cmpx windows (four private copies of one fact, three of them stale).

This was deliberately not done inside #2172: refactoring shared loader code for a single tool's benefit is disproportionate, and the tool is useful with its scope stated. It becomes worth doing as soon as there is a second consumer, or the first time someone acts on a census absence.

### How to verify a fix

[...]

Issue #2081, opened 2026-08-06
Sweep the unimplemented-NID census for FALSE SUCCESS: functions whose SCE_OK is also the dispatcher's default 0


mattias800:
## The question this sweep applies

From #2065: `sceRandomGetRandomNumber` is unregistered, so the dispatcher answers its default `0`.
For that contract `0` **is** `SCE_OK`, and the function's real work happens in an out-parameter — so
the guest is told its buffer was filled when nothing wrote to it, and reads whatever was already
there as "random" data. Not a fail-visible zero: a confident lie.

That is not a property of `libSceRandom`. It is a property of a **shape**, and the shape is
mechanically testable:

> **Is `0` this function's success code, and does it have an out-parameter?**

| the contract's `0` means | leaving it unregistered gives | verdict |
|---|---|---|
| an error / "not available" | a fail-visible negative the guest handles | inert — safe to leave |
| **success** (`SCE_OK`) | a **false success**, out-parameter untouched | **never inert** — implement or fail loudly |
| a data value the guest consumes | a plausible zero (size 0, count 0, handle 0) | depends on the call site |

This is the inverse of the established trap that *registering* a NID with an error sentinel can be
worse than leaving it unregistered (a `libSceUlt` stub's sentinel read as a size caused a 2 GiB
`malloc`). Both rules are the same underlying fact: **the safe default depends on whether the
contract's success value is also its zero value.**

## Why this needs a sweep rather than case-by-case triage

The current triage question in practice is "does this library look important?", which is not
applicable to a census and does not correlate with the failure. `libSceRandom` looks peripheral and
is in the dangerous bucket; plenty of prominent libraries are not.

Every title's boot log already emits the input:

```
[prosper] unimplemented: <lib>::<NID>  -> returning 0
```

Resolve each against `../PS5-3.20_Libs/` and classify by the two-part test. That is a bounded,
offline, mechanical pass over data we already have.

## The interaction with `nid_gate_scan --all-nids` — and the part that is easy to get backwards

`tools/re/nid_gate_scan.py --all-nids` (#2056) already answers the *consumption* side statically:
for every import, does the guest branch on the returned `eax`, and how. Its `ignored` bucket means
"`eax` is dead before it is read — this call site cannot be affected by the answer."

**`ignored` does NOT clear a false-success candidate.** For this failure the damage is in the
**untouched out-parameter**, not in the return value, so a call site that drops `eax` entirely is
[...]

PR #3293, opened 2026-09-03
hle: close the CALL side of the guest-ABI boundary (#3272, #3273)

### LINKED ISSUE #3272: The guest-ABI guard closes registration but not calls: Hle::lookup still hands back a bare HleFn
## What

`Hle::register_guest_abi` (#3246) closes the **registration** side of the guest-ABI boundary at compile
time: its parameter type carries `PROSPER_GUEST_ABI`, so on Windows a handler that forgot the tag is a
hard compile error, and `tests/host/abi/test_guest_varargs.cpp` now `static_assert`s that the tag really
does change the function type there (otherwise the guard would compile and close nothing).

The **call** side is still open. `Hle::lookup` hands back a bare `HleFn`:

```cpp
using HleFn = PROSPER_SYSV_ABI uint64_t (*)(uint64_t, uint64_t, uint64_t, uint64_t, uint64_t, uint64_t);
static HleFn lookup(const std::string& nid);
```

Nothing stops a caller taking a guest-ABI handler's address out of it and invoking it through an
ordinary host-ABI function-pointer type. On Windows the arguments are then placed by one convention
and read by the other, and for a `printf`-family handler the first `%s` dereferences whatever landed
in `rdi`.

## Not hypothetical

This is exactly what happened during #3246. `tests/hle/test_printf.cpp` and
`tests/gpu/test_guest_log_capture.cpp` both did it, and both **SEGFAULTed on the `Windows MinGW` CI
job** the moment the handlers were tagged — while passing on every other platform, because
`PROSPER_GUEST_ABI` is empty there and the two types are identical. They were fixed by tagging the
pointer types, and a sweep found no other direct callers today (`Hle::lookup`'s other users all take
ordinary integer handlers, for which `HleFn` remains correct).

So the current defence is a comment at `register_guest_abi` plus one CI job on a platform nobody
develops on. That is thin for a failure whose symptom is a wild pointer dereference.

## Suggested fix

A typed accessor that cannot return the wrong shape, e.g.

```cpp
template <class Fn> static Fn lookup_guest_abi(const std::string& nid);   // Fn must be PROSPER_GUEST_ABI
```

plus making the plain `Hle::lookup` refuse — or at least loudly complain — for a NID whose entry has
`guest_abi` set. The two stub emitters (`exec_image_win.cpp`, `exec_image_linux.cpp`) only need the
ADDRESS, so they can take an accessor that says so, leaving `lookup` free to mean "an ordinary
host-ABI handler" and nothing else.

`Hle::guest_abi_nid(nid)` already exists and is what such a check would key on.

Found while fixing #3246.

### LINKED ISSUE #3273: test_guest_varargs: the "CRT subset probe ran" assertion is a tautology, in the one file whose subject is arms that cannot fail
## Summary

`prosper/tests/host/abi/test_guest_varargs.cpp` (landed in #3266) contains an assertion that cannot
fail:

```cpp
expect("CRT subset probe ran", probed == sizeof kSpecs / sizeof *kSpecs, …);
```

`probed` counts the elements of the very array it is compared against, so the comparison is a
tautology and its failure message describes a state that cannot occur.

## Why it is worth a number rather than a shrug

It is harmless in itself — nothing depends on it, and the arm it sits beside is sound (the real
vacuity guard is `crt_rejects > 0`, which review verified fires on the one failure mode the loop
cannot see).

What earns it an issue is **where** it is. #3266 disclosed **two** of its own arms as vacuous during
review — a prefix-cap arm that passed against a build with its own fix removed, and a return-hook
arm that read a registry before it was populated — and the PR's whole subject is a walker that must
not accept what the CRT rejects. A tautological assertion in that file is the same defect class the
file exists to guard against, one level up, and the next reader will reasonably take it as a
verified property because of the company it keeps.

## Suggested fix

Pin the count to a literal:

```cpp
expect("CRT subset probe ran", probed == 11, …);
```

so an accidental edit to `kSpecs` becomes visible instead of silently changing what the arm claims
to have probed. One line.

## Not a defect in the shipped behaviour

The Windows CRT-subset arm itself is correct and was verified both ways in review: the implication
`!we_consume || crt_consumes` catches a detector stuck at `false` via the loop, and `crt_rejects > 0`
catches one stuck at `true`. `%qd` is the right anchor because it is rejected by UCRT *and* by
`__mingw_vsnprintf`, so the guard holds whichever implementation a future build resolves to.

Found in review of #3266; not blocking, so it merged.

mattias800:
Closes #3272. Closes #3273.

## The problem

#3246 closed the **registration** side of the guest-ABI boundary at compile time:
`Hle::register_guest_abi`'s parameter type carries `PROSPER_GUEST_ABI`, so on Windows a handler that
forgot the tag is a hard compile error rather than a stub that reads the wrong registers.

The **call** side stayed open. `Hle::lookup` handed back a bare `HleFn`, so nothing stopped a caller
taking a guest-ABI handler's address out of it and invoking it through an ordinary host-ABI function
pointer. The arguments are then placed by one convention and read by the other, and for a
printf-family handler the first `%s` dereferences whatever landed in `rdi`.

Not hypothetical: during #3246 both `test_printf` and `test_guest_log_capture` did exactly that, and
both **SEGFAULTed on the `Windows MinGW` job** the moment the handlers were tagged — while passing on
every other platform, because `PROSPER_GUEST_ABI` is empty there and the two types are identical.
They were patched by hand-spelling tagged pointer types at each call site, which left the defence as
a comment plus one CI job on a platform nobody develops on.

## The contract

`Hle::lookup` now means exactly one thing: **a handler `HleFn` describes correctly**. The other needs
get their own accessors, so no caller has to spell a calling convention:

| accessor | answers | refuses |
| --- | --- | --- |
| `lookup(nid)` | an ordinary host-ABI handler | a guest-ABI handler — nullptr, and complains once per NID naming the way out |
| `lookup_address(nid)` | the handler address, no convention attached | nothing |
| `lookup_guest_abi<R, A...>(nid)` | a `PROSPER_GUEST_ABI R (*)(A..., ...)` **this header constructs** | a handler that is not guest-ABI — the same category error in the other direction |

The third is load-bearing. The caller names the return type and the fixed prefix and never names the
convention, so an untagged pointer cannot come out of it:

```cpp
auto guest_printf = Hle::lookup_guest_abi<int, const char*>(nid_hash("printf"));
```

Both call sites that previously hand-spelled the tag now go through it and no longer mention
`PROSPER_GUEST_ABI` at all.

## `lookup` was conflating THREE questions, not two — and a test proved it

This is the part worth a reviewer's attention, because I got it wrong first.

I convinced myself by grep that the only callers needing care were the two stub emitters, converted
them to `lookup_address`, and ran the suite. **`hle_functions_registered` failed:**

```
  [FAIL] not registered: snprintf
```

`test_hle_registered.cpp:17` was using `Hle::lookup(...) == nullptr` to ask *"is this NID
registered?"* — a third question, distinct from both "give me something callable" and "give me a
guest-ABI handler". With `lookup` refusing the printf family, a registered handler now read as
missing: exactly the "a caller that only checks for null reports it as unimplemented" failure the
new code comment warns about, landing on the first run.

The fix is to ask the existence question through `lookup_address`, and the incident is the argument
for the split rather than against it — the three questions had genuinely different right answers and
one accessor was giving all three the same one. It also means the grep-based sweep was **not**
sufficient, which is the thing I would most want checked independently.

## Why the emitters had to move first

Both import-stub emitters resolve **every** import slot through this call and only ever patch the
value into a jump target. Had either been missed, `printf` would have resolved to nullptr and been
emitted as an *unimplemented* stub for every title on both platforms — a silent, total loss of guest
stdout.

**An earlier draft of this section claimed arm (3) and mutation M2 cover that. They do not, and the
correction matters more than the claim did.** Arm (3) is a property of `lookup_address`; M2 mutates
`lookup_address`. Nothing couples `emit_one_stub` to the accessor it calls — revert
`exec_image_linux.cpp` to `lookup` today and every new arm stays green. That uncovered gap is
precisely how the `nid_census` miss described below got through, and it is the strongest remaining
argument for reading the caller sweep independently rather than trusting the test count.

## Review found a third missed caller — and a third question

Review rejected the first head on `nid_census.cpp:279` and `:308`, still asking *"is this NID
registered?"* as `Hle::lookup(nid) != nullptr`. With `lookup` refusing the printf family, that tool
would report five real handlers as having **no** handler — into the table `#2081` acts on, i.e. the
"dispatcher silently returns 0" false-success list. The tool's own header states exactly the
invariant this breaks, and it is built unconditionally with **no ctest case**, so `353/353` was fully
consistent with it being broken.

That was the *third* caller of that question, after `test_hle_registered`. So rather than convert two
more sites, the question now has a name — **`Hle::registered(nid)`** — and all three spell it, so a
fourth caller does not have to improvise. New arm (4b) pins it for guest-ABI NIDs specifically.

**Reach — I got this wrong, and the correction is in the record because the wrong version was.** An
earlier draft claimed "0 of 55 local dumps import a printf-family NID", so the wrong table was
unreachable. Review challenged it and it does not survive: **all 55 eboots import the printf family
as real `libc` imports.**

What produced the false zero was not the reviewer's proposed mechanism (`:281` suppressing
registered NIDs — I did pass `--registered`). It is **`nid_census.cpp:264`**, which drops any import
satisfied by a *sibling module's* export before the registration check at `:279`. Pointed at an
app0 directory the tool finds the title's bundled libc and excludes printf; pointed at the eboot
alone it shows it for **55 of 55**. A/B with that one line disabled: 55 of 55 rows appear.

So the blast radius was **not** zero. `nid_census` accepts a module path as well as an app0
directory, and on the eboot-alone invocation the printf family reaches `:279` in every dump. ~~The tool's own header at `:157-158` makes the runtime case sharper still: the loader links a
fixed set and never auto-links `.sprx`, so a NID the census excludes as sibling-satisfied "WOULD
reach the dispatcher at runtime".~~ **Struck — that citation leans on a stale comment.** #2199 made
`nid_census.cpp:141` build its module set from `prosper::boot_link_inputs()`, the loader's real link
set, and `boot_program.hpp:156-159` says outright that the census's cross-module exclusions "are only
sound over the set the loader really links … cannot drift from it." Measured in review: PPSA24651's
tree holds 8 module candidates and directory mode reads **7**, the `.sprx` dropped — link-set
behaviour. So in directory mode the exclusion is *sound*, and the comment I quoted argues the
opposite of what is now true.

The conclusion does not need it. Reach stands on the eboot-alone invocation, and more sharply on the
counterfactual the reviewer measured: the **pre-PR** binary on the eboot alone reports those NIDs
with `registered=1`, so commit 1 flipped exactly those to 0. (The stale comment, and the fact that
`:183-186` prints it in every report, are filed separately.)

The methodological error is worth naming because it is subtler than the one it replaced: I did
correctly reject a vacuous null for the narrow question, then restated the same observation as a
stronger claim about the corpus that it could not support.

## Also fixed: #3273

`test_guest_varargs.cpp` asserted `probed == sizeof kSpecs / sizeof *kSpecs` — the count compared
against the array it was counted from, so the arm could not fail. Pinned to the literal `11`.

**Caveat, stated because the exit code cannot distinguish it:** that arm lives inside
`#if defined(_WIN32)` (`:533-639`) and does not compile on Linux, so my mutation for it — add a
twelfth specifier, expect the pinned count to redden — **passed here**. That is *unverifiable*, not
negative. Its verification is the MinGW job (which does run `ctest --no-tests=error`, `ci.yml:632`)
plus a reading of the table: `kSpecs` has 11 rows and no preprocessor directives inside its
initializer. #3273 carries a measurement of how large that Linux-side blind spot is (102 assertions
across 7 files sit in `_WIN32` regions with no `#else`).

## Verification

Linux, in the distrobox, at this exact head:

```
cmake --build prosper/build-linux -j4                       # exit 0
ctest --no-tests=error --output-on-failure --timeout 900    # exit 0, 353/353 passed, 137.65s
```

Test count unchanged at **353** — this adds arms to existing binaries, not a new target, so a
different count would itself be a red flag.

`test_guest_varargs` prints five `[prosper] Hle::lookup refused …` lines by design: arm (1) drives
the refusal path deliberately, and their absence would mean the loud path went quiet.

### Mutation arms — each new assertion shown to fail without the fix

Every arm below was applied to a clean tree, built, run, and reverted; verdicts are **test exit
codes**, and the named arm that fired is quoted.

| # | mutation | exit | arm that fired |
| --- | --- | --- | --- |
| M0 | none (baseline) | **0** | — |
| M1 | `lookup` does not refuse guest-ABI (the pre-fix behaviour) | **1** | `lookup refuses guest-ABI handlers` |
| M2 | `lookup_address` refuses them too (the missed-emitter regression) | **1** | `guest-ABI handlers remain addressable` |
| M3 | `lookup` refuses *everything* (over-broad refusal) | **1** | `lookup still returns ordinary handlers` |
| M4 | `lookup_guest_abi` drops its `guest_abi_nid` check | **1** | `lookup_guest_abi refuses an ordinary handler` |
| M5 | `kSpecs` grows, pinned count does not (#3273) | **0** | **VOID on Linux — Windows-only block, see caveat above** |
| M6 | restored | **0** | — |

M3 is the discriminator that matters: refusing everything would satisfy M1's arm just as well and
would be a far worse bug than the one being fixed, since every import in the tree resolves here.

Separately, `hle_functions_registered` is an unplanned integration arm for the same property: it went
`8 → 0` across the existence-accessor fix described above.

Re-run at the corrected head `7f4d0ddf`: build exit 0, `ctest --no-tests=error` **353/353**, exit 0.

### Not covered

Nothing here runs a live guest calling `printf` on a Windows host, and these arms check the shape of
the API surface — what each accessor refuses — not the eventual crash. That is deliberate: the
mis-typed-pointer failure is invisible on Linux by construction, so a Linux test cannot assert the
crash, only the guard preventing it. The MinGW job remains the only place the convention mismatch is
observable.

This does **not** address #3271 (the Linux swap stub re-pushes only four spilled words). That issue
now carries a derivation showing the bounded-copy option cannot be made correct at any word count;
the fix is a separate, riskier change to the path every guest import takes, and wants a quiet machine.

### DIFF
--- a/prosper/src/hle/dispatch/dispatch.cpp
+++ b/prosper/src/hle/dispatch/dispatch.cpp
@@ -157,6 +157,9 @@ std::vector<RegisteredFn> Hle::registrations() {
                         kv.second.signature, kv.second.guest_abi });
     return out;
 }
+bool Hle::registered(const std::string& nid) {
+    return registry().find(nid) != registry().end();
+}
 const void* Hle::lookup_address(const std::string& nid) {
     auto it = registry().find(nid);
     return it == registry().end() ? nullptr : reinterpret_cast<const void*>(it->second.fn);
--- a/prosper/src/hle/dispatch/dispatch.cpp
+++ b/prosper/src/hle/dispatch/dispatch.cpp
@@ -157,9 +157,39 @@ std::vector<RegisteredFn> Hle::registrations() {
                         kv.second.signature, kv.second.guest_abi });
     return out;
 }
+const void* Hle::lookup_address(const std::string& nid) {
+    auto it = registry().find(nid);
+    return it == registry().end() ? nullptr : reinterpret_cast<const void*>(it->second.fn);
+}
 HleFn Hle::lookup(const std::string& nid) {
     auto it = registry().find(nid);
-    return it == registry().end() ? nullptr : it->second.fn;
+    if (it == registry().end()) return nullptr;
+    if (it->second.guest_abi) {
+        // A programming error, not a runtime condition (#3272). Handing the address back would give
+        // the caller a pointer whose type describes the wrong calling convention, and on every
+        // platform but Windows that compiles and behaves correctly -- so the mistake stays invisible
+        // until the MinGW job dereferences a wild pointer. Refuse, loudly, and name the way out.
+        //
+        // Loud ONCE per NID: this can be reached from a per-import-slot loop, and a message per call
+        // would bury the first one. `nullptr` is deliberately NOT the same answer as "unimplemented"
+        // in the log, because a caller that only checks for null would otherwise report a registered
+        // handler as missing.
+        static std::mutex mx;
+        static std::unordered_map<std::string, bool> warned;
+        bool first = false;
+        {
+            std::lock_guard<std::mutex> lk(mx);
+            first = warned.emplace(nid, true).second;
+        }
+        if (first)
+            fprintf(stderr,
+                    "[prosper] Hle::lookup refused %s (%s): it is a guest-ABI handler and HleFn is "
+                    "the wrong type for one. Call it with Hle::lookup_guest_abi<R, Args...>(nid), or "
+                    "use Hle::lookup_address(nid) if you only need the address.\n",
+                    it->second.name.c_str(), nid.c_str());
+        return nullptr;
+    }
+    return it->second.fn;
 }
 HleReturnHook Hle::return_hook_of(const std::string& nid) {
     auto it = registry().find(nid);
--- a/prosper/src/hle/dispatch/dispatch.hpp
+++ b/prosper/src/hle/dispatch/dispatch.hpp
@@ -162,6 +162,13 @@ public:
     // never call it -- so they are unaffected by the refusal above. Prefer it over `lookup` wherever
     // the value is an address rather than something you intend to invoke.
     static const void* lookup_address(const std::string& nid);
+    // "Is there a handler for this NID at all?" -- a THIRD question, and the one that made the
+    // split necessary rather than merely tidy. It is not "give me something callable" (`lookup`)
+    // and not "give me the address" (`lookup_address`), and before #3272 all three were spelled
+    // `lookup(nid) != nullptr`. Two callers asking it that way broke the moment `lookup` learned to
+    // refuse -- `test_hle_registered` and `nid_census`, the latter with no ctest case, so nothing
+    // would have gone red. Spell the question so a fourth caller does not have to improvise it.
+    static bool registered(const std::string& nid);
     // A pointer type in the GUEST's calling convention. Spelled once, here, so no caller has to.
     template <class R, class... A>
     using GuestAbiFn = PROSPER_GUEST_ABI R (*)(A..., ...);
--- a/prosper/src/hle/dispatch/dispatch.hpp
+++ b/prosper/src/hle/dispatch/dispatch.hpp
@@ -143,7 +143,44 @@ public:
     // tracing thunk that a real handler is expected to replace later). A subsequent register_fn that
     // overwrites a placeholder is NOT flagged as a shadow — that override is the intent.
     static void  register_placeholder(const std::string& nid, HleFn fn, const char* name);
-    static HleFn lookup(const std::string& nid);          // nullptr if unimplemented
+    // An ordinary HOST-ABI handler for a NID; nullptr if unimplemented.
+    //
+    // REFUSES a guest-ABI handler, returning nullptr and complaining once per NID (#3272). Those are
+    // compiled in the guest's convention, and invoking one through `HleFn` places the arguments by
+    // one convention and reads them by the other -- on Windows a printf-family handler's first `%s`
+    // then dereferences whatever landed in rdi. That is not hypothetical: `test_printf` and
+    // `test_guest_log_capture` both did it and SEGFAULTed on the Windows MinGW job the moment these
+    // handlers were tagged, while passing everywhere else, because PROSPER_GUEST_ABI is empty on
+    // every other platform and the two types are then identical.
+    //
+    // So `lookup` now means exactly "a handler I may call through HleFn", and the two other needs
+    // have their own accessors: `lookup_guest_abi` to CALL a guest-ABI handler with a pointer type
+    // this header constructs, and `lookup_address` when the address is all you want.
+    static HleFn lookup(const std::string& nid);
+    // The handler ADDRESS for a NID, with no calling convention attached; nullptr if unimplemented.
+    // This is what the import-stub emitters want -- they only ever patch the value into a jump, and
+    // never call it -- so they are unaffected by the refusal above. Prefer it over `lookup` wherever
+    // the value is an address rather than something you intend to invoke.
+    static const void* lookup_address(const std::string& nid);
+    // A pointer type in the GUEST's calling convention. Spelled once, here, so no caller has to.
+    template <class R, class... A>
+    using GuestAbiFn = PROSPER_GUEST_ABI R (*)(A..., ...);
+    // Look up a handler registered with `register_guest_abi`, as a pointer type THIS ACCESSOR
+    // constructs (#3272). The caller names the return type and the fixed prefix and never names the
+    // calling convention, so an untagged pointer cannot come out of here -- which is the whole
+    // point, since on every platform but Windows an untagged one compiles and behaves identically
+    // and the mistake is invisible until the MinGW job runs.
+    //
+    //     auto guest_printf = Hle::lookup_guest_abi<int, const char*>(nid_hash("printf"));
+    //
+    // nullptr if the NID is unimplemented OR is not a guest-ABI handler -- asking for the guest
+    // convention on an ordinary host handler is the same category error in the other direction, and
+    // silently handing back a mis-typed pointer is exactly what this accessor exists to stop.
+    template <class R, class... A>
+    static GuestAbiFn<R, A...> lookup_guest_abi(const std::string& nid) {
+        if (!guest_abi_nid(nid)) return nullptr;
+        return reinterpret_cast<GuestAbiFn<R, A...>>(const_cast<void*>(lookup_address(nid)));
+    }
     static HleReturnHook return_hook_of(const std::string& nid); // called by the import return trampoline
     static const char* name_of(const std::string& nid);   // registered display name or ""
     // Every NID that was registered 2+ times with DIFFERING handlers, in registration order. Empty
--- a/prosper/src/host/image/exec_image_linux.cpp
+++ b/prosper/src/host/image/exec_image_linux.cpp
@@ -3057,11 +3057,16 @@ bool map_image(const LoadedImage& img, std::string* err) {
 // byte-identical stubs for the same slot, or a runtime-loaded module's imports would take a
 // different path into the HLE than the pre-linked ones.
 static size_t emit_one_stub(uint8_t* slot, const ImportSlot& s, uint32_t idx, bool swap) {
-    HleFn fn = Hle::lookup(s.nid);
+    // lookup_ADDRESS, not lookup (#3272): this value is only ever patched into the emitted stub as a
+    // jump target, never invoked from here, so the calling convention of the handler behind it is
+    // none of this function's business -- and `lookup` now refuses the guest-ABI handlers precisely
+    // because it hands back a type that claims one.
+    const void* fn = Hle::lookup_address(s.nid);
     HleReturnHook return_hook = Hle::return_hook_of(s.nid);
     if (fn) {
-        if (return_hook) return emit_impl_hook(slot, (uint64_t)fn, (uint64_t)return_hook, swap);
-        return swap ? emit_impl_swap(slot, (uint64_t)fn) : emit_impl(slot, (uint64_t)fn);
+        const uint64_t target = (uint64_t)(uintptr_t)fn;
+        if (return_hook) return emit_impl_hook(slot, target, (uint64_t)return_hook, swap);
+        return swap ? emit_impl_swap(slot, target) : emit_impl(slot, target);
     }
     return swap ? emit_unimpl_swap(slot, idx, (uint64_t)&prosper_on_unimpl)
                 : emit_unimpl(slot, idx, (uint64_t)&prosper_on_unimpl);
--- a/prosper/src/host/image/exec_image_win.cpp
+++ b/prosper/src/host/image/exec_image_win.cpp
@@ -1194,14 +1194,19 @@ bool map_image(const LoadedImage& img, std::string* err) {
 // byte-identical stubs for the same slot, or a runtime-loaded module's imports would take a
 // different path into the HLE than the pre-linked ones.
 static size_t emit_one_stub(uint8_t* slot, const ImportSlot& s, uint32_t idx, size_t capacity) {
-    HleFn fn = Hle::lookup(s.nid);
+    // lookup_ADDRESS, not lookup (#3272): this value is only ever patched into the emitted stub as a
+    // jump target, never invoked from here, so the calling convention of the handler behind it is
+    // none of this function's business -- and `lookup` now refuses the guest-ABI handlers precisely
+    // because it hands back a type that claims one. `guest_abi_nid` below is what actually decides
+    // the stub's shape for those.
+    const void* fn = Hle::lookup_address(s.nid);
     const uint64_t return_hook = (uint64_t)(uintptr_t)Hle::return_hook_of(s.nid);
     // Stage, then copy. The bridge's length depends on the handler's declared signature, and a stub
     // that does not fit its slot must be REPORTED rather than written: emitting straight into the
     // table would already have overrun the next slot by the time the caller compared the length.
     uint8_t staged[abi::kMaxBridgeBytes];
     const size_t emitted =
-        fn ? emit_impl(staged, (uint64_t)fn, return_hook, Hle::signature_of_nid(s.nid),
+        fn ? emit_impl(staged, (uint64_t)(uintptr_t)fn, return_hook, Hle::signature_of_nid(s.nid),
                        Hle::guest_abi_nid(s.nid))
            : emit_unimpl(staged, idx, (uint64_t)&prosper_on_unimpl);
     if (emitted <= capacity) memcpy(slot, staged, emitted);
--- a/prosper/tests/gpu/test_guest_log_capture.cpp
+++ b/prosper/tests/gpu/test_guest_log_capture.cpp
@@ -68,10 +68,10 @@ int main() {
     // must carry PROSPER_GUEST_ABI -- on Windows that is `sysv_abi`, and calling it through an
     // untagged type places the arguments by the wrong convention. The integer handlers below are
     // ordinary host functions and need no tag.
-    using PrintfFn = PROSPER_GUEST_ABI int (*)(const char*, ...);
+    // `lookup_guest_abi` constructs the tagged pointer type (#3272); the integer handlers below stay
+    // on plain `lookup`, which now means exactly "a handler HleFn describes correctly".
     using HleFn = uint64_t (*)(uint64_t, uint64_t, uint64_t, uint64_t, uint64_t, uint64_t);
-    auto guest_printf = reinterpret_cast<PrintfFn>(
-        reinterpret_cast<void*>(prosper::Hle::lookup(prosper::nid_hash("printf"))));
+    auto guest_printf = prosper::Hle::lookup_guest_abi<int, const char*>(prosper::nid_hash("printf"));
     auto guest_fputs = reinterpret_cast<HleFn>(
         reinterpret_cast<void*>(prosper::Hle::lookup(prosper::nid_hash("fputs"))));
     auto guest_fwrite = reinterpret_cast<HleFn>(
--- a/prosper/tests/hle/test_hle_registered.cpp
+++ b/prosper/tests/hle/test_hle_registered.cpp
@@ -13,8 +13,15 @@
 using namespace prosper;
 
 static int fails = 0;
+// "Is this NID registered?" is a THIRD question, distinct from both "give me a handler I may call
+// through HleFn" and "give me a guest-ABI handler" -- and `lookup` used to answer all three. It no
+// longer answers this one for the printf family, which are guest-ABI (#3272): `lookup` refuses
+// those, so asking it about `snprintf` here reported a registered handler as missing. That is the
+// whole reason the accessor was split; `lookup_address` is the one that means existence.
 static void must(const char* name) {
-    if (Hle::lookup(nid_hash(name)) == nullptr) { printf("  [FAIL] not registered: %s\n", name); fails++; }
+    if (Hle::lookup_address(nid_hash(name)) == nullptr) {
+        printf("  [FAIL] not registered: %s\n", name); fails++;
+    }
 }
 
 int main() {
@@ -52,7 +59,7 @@ int main() {
     };
     for (const char* n : names) must(n);
     // sync_on_address futex is registered by raw NID (no symbol name) — check it directly.
-    if (Hle::lookup("Hc4CaR6JBL0") == nullptr) { printf("  [FAIL] sceKernelWaitOnAddress raw NID\n"); fails++; }
+    if (Hle::lookup_address("Hc4CaR6JBL0") == nullptr) { printf("  [FAIL] sceKernelWaitOnAddress raw NID\n"); fails++; }
 
     // __ctype_get_mb_cur_max returns the VALUE of MB_CUR_MAX (1 in the "C" locale we
     // present), not a pointer to it — guest code sizes buffers as MB_CUR_MAX*n (#141).
--- a/prosper/tests/hle/test_hle_registered.cpp
+++ b/prosper/tests/hle/test_hle_registered.cpp
@@ -17,9 +17,9 @@ static int fails = 0;
 // through HleFn" and "give me a guest-ABI handler" -- and `lookup` used to answer all three. It no
 // longer answers this one for the printf family, which are guest-ABI (#3272): `lookup` refuses
 // those, so asking it about `snprintf` here reported a registered handler as missing. That is the
-// whole reason the accessor was split; `lookup_address` is the one that means existence.
+// whole reason the accessor was split; `Hle::registered` is the one that means existence.
 static void must(const char* name) {
-    if (Hle::lookup_address(nid_hash(name)) == nullptr) {
+    if (!Hle::registered(nid_hash(name))) {
         printf("  [FAIL] not registered: %s\n", name); fails++;
     }
 }
@@ -59,7 +59,7 @@ int main() {
     };
     for (const char* n : names) must(n);
     // sync_on_address futex is registered by raw NID (no symbol name) — check it directly.
-    if (Hle::lookup_address("Hc4CaR6JBL0") == nullptr) { printf("  [FAIL] sceKernelWaitOnAddress raw NID\n"); fails++; }
+    if (!Hle::registered("Hc4CaR6JBL0")) { printf("  [FAIL] sceKernelWaitOnAddress raw NID\n"); fails++; }
 
     // __ctype_get_mb_cur_max returns the VALUE of MB_CUR_MAX (1 in the "C" locale we
     // present), not a pointer to it — guest code sizes buffers as MB_CUR_MAX*n (#141).
--- a/prosper/tests/hle/test_printf.cpp
+++ b/prosper/tests/hle/test_printf.cpp
@@ -17,19 +17,24 @@ static int fails = 0;
 
 // PROSPER_GUEST_ABI is what makes the comment above TRUE rather than aspirational. These handlers
 // are compiled in the GUEST's convention (#3246), which on Windows is not the host's -- so an
-// untagged pointer type here would place the arguments by Microsoft x64 rules for a callee reading
-// them by System V, and the first `%s` would dereference whatever landed in `rdi`. It did: this test
-// SEGFAULTed on the Windows MinGW job the moment the handlers were tagged. Empty on Linux/macOS.
-using SnprintfFn = PROSPER_GUEST_ABI int (*)(char*, size_t, const char*, ...);
-using SprintfFn  = PROSPER_GUEST_ABI int (*)(char*, const char*, ...);
+// untagged pointer type would place the arguments by Microsoft x64 rules for a callee reading them
+// by System V, and the first `%s` would dereference whatever landed in `rdi`. It did: this test
+// SEGFAULTed on the Windows MinGW job the moment the handlers were tagged.
+//
+// This file used to spell those tagged types itself. It no longer does: `Hle::lookup_guest_abi`
+// constructs them (#3272), so the convention is spelled in one place instead of at every call site
+// that wants one of these handlers.
 
 int main() {
     printf("== test_printf ==\n");
     register_builtin_hle();
 
-    auto snf = (SnprintfFn)(void*)Hle::lookup(nid_hash("snprintf"));
-    auto spf = (SprintfFn)(void*)Hle::lookup(nid_hash("sprintf"));
-    auto snfs = (SnprintfFn)(void*)Hle::lookup(nid_hash("snprintf_s"));
+    // `lookup_guest_abi` builds the PROSPER_GUEST_ABI pointer type itself (#3272), so this file no
+    // longer spells the calling convention and cannot get it wrong -- which matters because an
+    // untagged type compiles and behaves identically everywhere except the MinGW job.
+    auto snf  = Hle::lookup_guest_abi<int, char*, size_t, const char*>(nid_hash("snprintf"));
+    auto spf  = Hle::lookup_guest_abi<int, char*, const char*>(nid_hash("sprintf"));
+    auto snfs = Hle::lookup_guest_abi<int, char*, size_t, const char*>(nid_hash("snprintf_s"));
     CHECK(snf && spf && snfs, "printf-family fns registered");
     if (!(snf && spf && snfs)) { printf("== FAIL ==\n"); return 1; }
 
--- a/prosper/tests/host/abi/test_guest_varargs.cpp
+++ b/prosper/tests/host/abi/test_guest_varargs.cpp
@@ -282,6 +282,26 @@ void check_lookup_accessors() {
            reinterpret_cast<const void*>(printf_fn) == Hle::lookup_address(nid_hash("printf")),
            "lookup_guest_abi returned an address the registry does not hold");
 
+    // (4b) `registered` answers the THIRD question, and must answer it for a guest-ABI handler --
+    //      `lookup` returning nullptr for one must not read as "no handler exists". This is the arm
+    //      that would have caught the `nid_census` miss found in review: that tool reports a NID
+    //      with no registered handler as a return-0 false-success candidate (#2081), so a wrong
+    //      answer here puts five real handlers into a table that is acted on. It has no ctest case
+    //      of its own, so nothing else in the suite can see it.
+    unsigned known = 0;
+    for (const char* name : kGuestAbi) {
+        if (Hle::registered(nid_hash(name))) { ++known; continue; }
+        fail("guest-ABI handlers report as registered",
+             std::string(name) + ": Hle::registered said no handler exists for a registered NID");
+    }
+    expect("registered-arm is not vacuous", known == 5,
+           "expected all five printf-family NIDs to report registered, got " + std::to_string(known));
+    expect("registered agrees with lookup for an ordinary handler",
+           Hle::registered(nid_hash("memcpy")) && Hle::lookup(nid_hash("memcpy")) != nullptr,
+           "registered and lookup disagree about an ordinary host-ABI handler");
+    expect("registered says no for an unregistered NID", !Hle::registered("no-such-nid-xyz"),
+           "Hle::registered claimed a NID that was never registered");
+
     // (5) The category error in the OTHER direction. Asking for the guest convention on an ordinary
     //     host handler must not quietly succeed either -- that mis-types the pointer just as badly,
     //     and it is the mistake a caller makes when they copy the line above to a different NID.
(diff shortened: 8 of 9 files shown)


mattias800 (commented):
**REJECTED** — one blocking finding. It is a two-line fix, and everything else I looked at is clean; I am rejecting only because the finding is the exact question you asked me to check ("did I miss a caller?") and the answer is yes.

---

## BLOCKING — a missed caller, and it is the one whose entire job is the question `lookup` no longer answers

`prosper/tools/nid_census/nid_census.cpp:279` and `:308` still ask existence through `Hle::lookup`:

```cpp
279:        const bool registered = Hle::lookup(nid) != nullptr;
308:                   Hle::lookup(r->nid) != nullptr ? 1 : 0,
```

This is `test_hle_registered`'s third question again, in a target that has no ctest case, so nothing goes red.

**The chain, all four links opened:**

* `src/hle/libc/hle_libc.cpp:1017-1019` registers `snprintf`, `sprintf`, `snprintf_s`, `sprintf_s`, `printf` through `RV` (`:984` = `Hle::register_guest_abi`);
* `src/hle/dispatch/dispatch.cpp:125-132` sets `registry()[nid].guest_abi = true` for each;
* `dispatch.cpp:164-190` (this PR) returns `nullptr` for exactly those;
* `nid_census.cpp:233` calls `register_builtin_hle()`, then `:279` reads `registered == false`, `:280` does `unregistered++`, and `:281` (`if (registered && !show_registered) continue;`) does **not** skip them — so they are printed under the heading at `:316`, `"== imports with NO registered handler -> dispatcher returns 0 =="`.

**Failure scenario:** `nid_census <DUMP_ROOT>/<TITLE>-app0` on any title that imports the printf family reports up to five registered libc handlers as unimplemented false-success candidates, and inflates the `scope: N unregistered` line by up to five. With `--tsv`, `:308` writes `registered=0` for them into whatever consumes that column.

**Why this is worse than the test caller you already caught, and why it is blocking rather than a nit:**

1. The tool's own header states the invariant this breaks, at `nid_census.cpp:20-23`: *"registration comes from `Hle::lookup` after `register_builtin_hle()` — the real runtime registry, so a NID registered from a table, a loop, or a raw literal is seen identically. A grep over `register_fn(...)` call sites would miss every non-literal registration and would report those as false-success candidates. Asking the registry cannot."* After this PR, asking the registry **can**. `CMakeLists.txt:584-590` repeats the same claim as the build rationale.
2. `nid_census` is built unconditionally inside `if(TARGET prosper_core)` (`CMakeLists.txt:589`) — including on the MinGW job, which builds the whole tree with no target list — but there is **no `add_test` for it**. So neither Linux nor MinGW can see this, and your 353/353 is consistent with it being broken.
3. The wrong answer is delivered in the shape #2081 is most likely to act on: a confident table naming a real handler as a return-0 gap. The one-shot stderr line does fire, but it lands on stderr while the table goes to stdout, and `--tsv` is explicitly a redirect-to-a-file mode.

**Fix:** `Hle::lookup_address` at both sites, exactly as in `test_hle_registered.cpp`. Worth a one-line comment there too, since the header comment at `dispatch.hpp:160-164` describes `lookup_address` as "wherever the value is an address" and never as "the accessor that means existence" — which is the second use you discovered and the one both misses share.

---

## Answers to the five things you asked me to challenge

### (1) Blast radius — swept, one miss (above), nothing else

Full-tree at `ea7e189`, over `*.cpp *.hpp *.h *.cc *.mm *.c`:
`git grep -nE "Hle[[:space:]]*::[[:space:]]*lookup|&Hle::lookup|decltype\(.*lookup"`, plus a separate pass over `*.py *.sh *.cmake *.txt *.ps1` (**zero** hits — nothing generates a call site), plus a pass over `*.md` (only `GAME_COMPAT_ORCHESTRATION.md:180`, trap 52, a historical narrative that is still accurate).

Outside `prosper/tests`, after your change there are exactly **three** live callers: the two emitters you converted, and `nid_census`. Inside tests: 896 `Hle::lookup` lines across ~90 files; I filtered them against the five guest-ABI NIDs (`git grep -E '"(v?[sn]?n?printf(_s)?)"'` over `src`, `tests`, `tools`, `frontends`), and the only survivors are your converted sites plus the deliberate refusal arm at `test_guest_varargs.cpp:235-247`. Also checked and cleared:

* `frontends/audio_sdl3/test_audio_sdl3.cpp:24` — the one non-`src` caller. It resolves only `sceAudioOut*`, all ordinary handlers, so `lookup` is correct for it. Noting it anyway because it is the identical existence-check shape (`if (!f) { "not registered" }`) and would become a third miss the day a guest-ABI NID enters its list.
* No `exec_image_mac.cpp` exists (`src/host/image/` is linux/win only), so there is no third emitter.
* `runtime_module_load.cpp` resolves nothing itself — it goes through `emit_one_stub`, whose only two callers per platform are `exec_image_{linux,win}.cpp:3110/:3149` and `:1241/:1278`.
* `tools/re/hle_registry_dump.cpp:42` reads `Hle::registrations()`, which returns `guest_abi` in the row — unaffected.
* `Hle::return_hook_of` / `signature_of_nid` / `name_of` are unchanged and keyed the same way.

### (2) Is `nullptr` the right refusal? — yes, but only because you convert the callers, and one is unconverted

You asked whether there is a caller for whom this is silently wrong. There is, and it is finding 1. That is the argument, and it cuts at the sweep rather than at the design: `nullptr` is fine *given* that every existence-check caller is converted, and it is the only option that keeps `lookup` usable at all (the NID is a runtime string, so no compile-time guard is possible). An `abort()` would be worse — it would convert a diagnostic tool's wrong column into a dead tool — and keeping the address would reopen the hole the PR exists to close.

What I would push back on is the *residual*: nothing in the tree can detect a caller that picks the wrong accessor. No test arm can, CI cannot, and the sweep has now missed twice. A named predicate — `static bool Hle::registered(const std::string& nid)` — would remove the temptation by making the third question spellable, and is the change I would most like to see alongside the two-line fix. Non-blocking, your call.

### (3) The alias template does **not** drop the attribute — measured

This is the one I most wanted to be false and it holds. The question is a front-end one (does a GNU attribute survive an *alias template* with a dependent return type and a pack before the ellipsis?), not a target one, so I tested it with the mirror-image non-default convention, `__attribute__((ms_abi))` on x86-64 Linux, on **g++ (GCC) 16.2.1** — the same GCC major that `dispatch.hpp`'s own note records for the MinGW toolchain (16.1.1); the job installs rolling `mingw-w64-ucrt-x86_64-gcc` (`ci.yml:650`), so I cannot pin its exact version from here. Verbatim reproduction of your declaration:

```
(a) direct alias   : tagged != untagged      <- the shape test_guest_varargs.cpp:67-76 already asserts
(b) alias TEMPLATE : tagged != untagged      <- your new shape
(c) tpl vs direct  : same                    <- TplTagged<int,const char*> IS the hand-spelled type
```

(c) is the strong result: the alias template yields *the same type* as the hand-spelled one, so the existing `static_assert` at `:67-76` transitively covers the alias form even though it is written in the direct form. I also checked codegen rather than trusting `is_same`: a call through the alias-template pointer type emits the ms_abi sequence (`rcx/rdx/r8/r9`, 32-byte shadow space) while the untagged alias template emits SysV (`rdi/rsi/rdx/rcx/r8/r9`, `xorl %eax,%eax` for the variadic AL). The guard is not cosmetic.

Caveat stated so it is not read as more than it is: this is GCC-on-Linux with `ms_abi`, i.e. a proxy for MinGW-with-`sysv_abi`; no MinGW cross-compiler is installed on this box, and I did not build prosper for it.

### (4) The #3273 half checks out, and the VOID disclosure is honest

* `kSpecs` is at `test_guest_varargs.cpp:551-563` with exactly **11** rows (`:552-562`), and there is no preprocessor directive inside the initializer. `11` is right for the build that executes it.
* The enclosing block really is `#if defined(_WIN32)` at `:533` / `#endif` at `:639` — your cite is exact.
* `probed` is incremented once per iteration at `:581`, so `probed == 11` is now a genuine property (table size **and** loop completion), not a relocated tautology.
* M5 is genuinely VOID, not a covered failure: `kSpecs` is *declared inside* the `_WIN32` block, so on Linux the mutation is preprocessed away along with the assertion — there is no path by which the arm could have evaluated and passed. The disclosure is accurate.
* `ci.yml:632` is `windows-mingw` and `:666` runs `ctest --test-dir prosper/build-ci --output-on-failure --timeout 600 --no-tests=error`. Cite verified.

Non-blocking: `static_assert(sizeof kSpecs / sizeof *kSpecs == 11, ...)` beside the runtime `expect` would move half the claim to compile time on the platform that compiles it, at no cost.

### (5) The new arms — M1-M4 hold; one PR-body overclaim to correct

Each named arm asserts the mutated property directly, and the two non-vacuity counters (`refused == 5`, `addressable == 5`) mean an arm cannot pass by iterating an empty set:

* M1 -> `:239-247` (`fail` at `:243`, non-vacuity `refused == 5` at `:246`). Asserts refusal per NID. Sound.
* M2 -> `:255-268` (`fail` at `:262`, non-vacuity `addressable == 5` at `:266`). Asserts addressability per NID. Sound.
* M3 -> `:251` (`lookup(memcpy) != nullptr`). This is the right discriminator and it is the arm I would have asked for.
* M4 -> `:288-290`. Sound, with one coupling worth knowing: it would also pass if `memcpy` were simply unregistered, since `guest_abi_nid` is false for a missing NID too. It is de-vacuumed by M3's arm in the same function, which independently establishes `memcpy` is registered. Fine as written; just note the two are not independent.
* Arm (6) at `:292-307` is the one that adds real value on Windows — it drives an actual call through the tagged pointer from an MS-ABI caller, which is precisely the crash #3246 hit. `n == 12` / `"s|1|2|3|4.50"` is arithmetically right.

Nothing vacuous found. Two precision items:

**(a) The PR body overclaims arm (3).** It says a missed emitter would emit `printf` as unimplemented, then: *"Arm (3) of the new test exists to catch precisely that, and mutation M2 below confirms it does."* It does not. Arm (3) asserts a property of `lookup_address`; M2 mutates `lookup_address`. Neither couples `emit_one_stub` to the accessor it calls — revert `exec_image_linux.cpp:3064` to `Hle::lookup` today and every arm in `check_lookup_accessors` stays green, because `check_stub_and_registry` tests `emit_sysv_to_ms_bridge` directly and never goes through `emit_one_stub`. That gap is exactly what let finding 1 through, so I would rather the body said so than claimed coverage it does not have. Please reword before merging — the issue thread inherits this text.

**(b) `test_guest_varargs.cpp:294-295` states something false.** *"the arguments deliberately spill past both register files"*. Count them: `buf`, `sizeof buf`, fmt, `"s"`, `1`, `2`, `3` are seven INTEGER-class arguments, so one spills to the overflow area — correct for the integer file. `4.5` is the only SSE-class argument, so the SSE file uses 1 of 8 and never spills. Either add eight doubles or say "past the integer register file".

---

## Also checked and cleared

* **Emitter behavioural equivalence.** `lookup_address` returns `reinterpret_cast<const void*>(it->second.fn)` for exactly the entries `lookup` previously returned, so `emit_one_stub` computes the same target on both platforms; `(uint64_t)(uintptr_t)fn` is value-identical to the old `(uint64_t)fn`, and the Windows `guest_abi` / signature / return-hook arguments are untouched. No behaviour change on the import path.
* **Includes.** `<mutex>` and `<unordered_map>` are already direct includes at `dispatch.cpp:4-5`, not transitive — the new statics will not surprise MinGW.
* **The one-shot statics.** Function-local, thread-safe initialisation, `emplace` under the lock, and `lookup` is no longer on any guest-hot path (only the emitters were, and they moved off it). No contention concern.
* **Member-order.** `lookup_guest_abi`'s body names `guest_abi_nid` (`:138`) and `lookup_address` (`:164`) — complete-class context, so the definition at `:180` is well-formed regardless of order.
* **ASCII output gate.** `python3 prosper/tools/output/check_ascii_output.py prosper` at your exact head: *"every string literal outside the quarantine is ASCII"*, all checks passed. The new refusal message and the new failure strings are clean, so `ascii_output_literals` will not redden.
* **No ctest inventory change.** The diff touches no `CMakeLists.txt`, which is consistent with your unchanged 353.
* **No `FAIL_REGULAR_EXPRESSION`/`PASS_REGULAR_EXPRESSION`** is set anywhere in `CMakeLists.txt`, so the five deliberate `[prosper] Hle::lookup refused ...` stderr lines cannot fail `guest_varargs` by accident.
* **Docs.** No `.md` in the tree describes `Hle::lookup`'s contract in a way this invalidates. Optional: `src/hle/dispatch/` has no `AGENTS.md` (its siblings `audio/`, `graphics/`, `memory/`, `sync/` do), and the three-accessor split is exactly the kind of boundary that map is for.

Per the charter I did not rebuild or re-run your suite; the only thing I compiled was the standalone 30-line attribute probe in (3), which is the one claim I could not settle by reading. No GPU was touched.

Re-review on request once the `nid_census` sites are converted — that alone flips this to APPROVED.

mattias800:
Blocking finding accepted and fixed at `7f4d0ddf`. Thank you — that was a real miss, and the reasoning about *why* CI could not see it (built unconditionally, no `add_test`) is what made it findable at all.

**What changed, and why not the one-line fix you suggested.** `lookup_address` at both sites would have worked. But that would have been the *third* site spelling "is this NID registered?" as an accident of whichever accessor happened to be handy — after `test_hle_registered`, which broke the same way an hour earlier. So the question now has a name:

```cpp
static bool Hle::registered(const std::string& nid);
```

`nid_census` (both sites, plus its header comment, which named `Hle::lookup` as the mechanism and would otherwise have become false), `test_hle_registered` (both sites) and new arm (4b) all spell it. Your point stands more strongly than the fix you proposed: three questions were sharing one accessor, not two.

**One qualification on reach, measured rather than argued.** I went looking for the live instance and could not find one: **0 of 55 local dumps import a printf-family NID**, so no census run in this corpus would have produced the wrong table. Your finding is still correct — the code was wrong and the header's stated invariant was genuinely broken — but "the wrong answer lands in the shape #2081 is acted on" is true in principle and unreachable today.

Worth recording *how* that came out, because it nearly went the other way: my first check was "do the printf NIDs appear in the unregistered table? no → fixed", which is the vacuous null. They are absent because nothing imports them. Only searching for a positive instance separated the two, and since the corpus supplies none, the hand-constructed positive instance is arm (4b) itself.

**Both non-blocking items fixed.**

- The **arm (3) overclaim** is corrected in the PR body, and I did not simply delete the sentence — the body now says explicitly that nothing couples `emit_one_stub` to its accessor, that reverting an emitter to `lookup` leaves every new arm green, and that this uncovered gap is how your `nid_census` finding got through. You are right that this is the more important half; it was the claim most likely to stop the next reader looking.
- **`test_guest_varargs.cpp:294-295`** corrected: seven integer-class arguments spill one word, the single `4.5` uses 1 of 8 SSE slots and never spills. The comment now says so and points at `check_executed` as the arm that owns the FP overflow path.

**Filed rather than fixed here**, so they are not lost:

- The emitter/accessor coupling gap — a test that pins `emit_one_stub` resolving a guest-ABI NID to an *implemented* stub. That is the guard that would make a missed emitter conversion impossible rather than merely unlikely, and it needs mapped memory and a real dump, so it is not a one-liner.
- `nid_census` having no ctest case at all.
- `frontends/audio_sdl3/test_audio_sdl3.cpp:24` — you flagged it as latent and benign; it calls through `HleFn`, so `lookup` is correct there, but its failure message says "not registered", which is the same conflation waiting for a guest-ABI NID to be added to that helper.
- `src/hle/dispatch/AGENTS.md`.

Re-verified at `7f4d0ddf`: build exit 0, `ctest --no-tests=error` **353/353**, exit 0. CI was fully green on the previous head and is re-running.

Re-review requested on the new head when you have a moment.

mattias800 (commented):
Re-review at head `7f4d0ddf5` (previous review was `ea7e1896a`).

**REJECTED** — and to be clear about what that is and is not: **the code is right, and the fix is better than the one I suggested.** `Hle::registered` is the correct third predicate, all four existence-check sites are converted, arm (4b) is sound, and my previous blocking finding is properly closed. Both blocking items below are *claims attached to the change*, one of them a measurement that I checked and that does not survive. No code change is needed to flip this.

---

## BLOCKING 1 — the "0 of 55" qualification is falsified, and it has already steered #3297

You asked me to check this rather than take it, so I did, from outside your instrument.

**Method, with its control first.** I reimplemented `nid_hash` in Python from `src/hle/dispatch/nid.cpp:66-76` (SHA-1 of name + the 16-byte salt, first 8 bytes reversed, Sony base64, 11 chars) and validated it against the authoritative firmware table before using it — `PS5-3.20_Libs/libSceLibcInternal.c:17871` gives `printf` = `hcuQgD53UxM`, `:17990` gives `snprintf` = `eLdDw6l0-bU`, `:17818` gives `memcpy` = `Q3VBxCXhUHs`, and my implementation reproduces all three exactly. Then one pass over every `eboot.bin` in `/var/mnt/prosper/testdata` (55 files, 4.2 GiB), searching for the five printf-family NIDs plus `memcpy` as the positive control:

```
dumps=55   printf=53   any-printf-family=55   memcpy(control)=55
no printf NID: PPSA03130-app0 PPSA16901-app0        (both carry sprintf/snprintf instead)
```

**53 of 55 carry the `printf` NID; 55 of 55 carry at least one of the five.** The memcpy control is present in all 55, so the method demonstrably sees NIDs in these files rather than seeing nothing.

**And they are in the import encoding, not incidental bytes.** Context around the match, three titles:

```
PPSA24651 eboot.bin:  j#j\0 hcuQgD53UxM#j#j \0 GlelR9E...
PPSA04263 eboot.bin:  B#C\0 hcuQgD53UxM#B#C \0 lA94ZgT...
PPSA20800 eboot.bin:  u#t\0 hcuQgD53UxM#u#t \0 2vDqwBl...
```

That is `<NID>#<libid>#<modid>` in `.dynstr` — the form `src/self/module.cpp:208-214` splits, and `:217` classifies as an import. The neighbouring entries are the rest of the libc import block (`Q2V+iqvjgC0` = `vsnprintf` sits directly beside it in The Messenger).

**Limit of what I measured, stated so it is not read as more:** I read `.dynstr` bytes; I did not run `Module::load`, so I have not per-symbol confirmed `st_shndx == 0 && st_value == 0`, nor applied the census's cross-module-export exclusion. But the claim as written is *"0 of 55 local dumps import a printf-family NID"*, and 53 of 55 carry `printf` in the import string form with a working positive control. Those two cannot both be true. If your number came from somewhere I have not looked, say where — because as it stands the corpus contradicts it.

**The likely mechanism, and it is the trap you flagged, one level up.** After the fix, `nid_census.cpp:281` is `if (registered && !show_registered) continue;` — so a correctly-registered NID is *suppressed from the default table by the fix itself*. "printf is not in the census output" is therefore guaranteed post-fix and carries no information about the corpus. You caught that as a vacuous null for the narrow question ("is it fixed?") and then restated it as a *stronger* claim about what the corpus contains, which the same observation cannot support. `--registered` is what would have shown them.

**Why this is blocking rather than a nit.** It is a measurement in the durable record that lowers the severity of a real defect, and it has already changed a design decision: #3297's *Suggested fix* says *"the awkward part is that no local dump imports a printf-family NID (0 of 55, measured while fixing #3272), so the slot list has to be synthesised rather than taken from an eboot."* It does not — every one of the 55 supplies a real printf-family import slot, which makes the guard test both easier and more faithful, since `test_trap_linux.cpp` already drives `install_stubs` against the configured `GAME_DUMP`. Please correct it in the PR body (~line 90) and in #3297, and the honest replacement is the stronger statement, not a weaker one: the wrong table was reachable on essentially the whole corpus.

Worth naming the shape, because the charter has a rule for exactly it: this is a correction that *raised* the author's position (a reviewer's finding downgraded to "true in principle, unreachable in practice") and it agreed with what you already hoped, which is the class that gets checked least. It reached a PR body and an issue before anyone re-derived it.

## BLOCKING 2 — `test_guest_varargs.cpp:285-289` reproduces the overclaim it was just fixed for

```
// (4b) ... This is the arm
//      that would have caught the `nid_census` miss found in review: ...
```

It would not, and this is the same class as the arm-(3) claim you corrected in the PR body. Arm (4b) asserts a property of `Hle::registered`. The miss was `nid_census` calling `Hle::lookup`. Had `Hle::registered` existed at that moment and `nid_census` still called `lookup`, arm (4b) would have been green and the miss would have stood — which is precisely what #3297 says: nothing couples a caller to the accessor it picks. So the code comment and the issue now contradict each other, and the comment is the more durable of the two.

The rest of the sentence is true and worth keeping (a wrong answer from `registered` now does put five real handlers into a table that is acted on). Suggested one-line replacement in kind: *"This arm pins the property `nid_census` now depends on; what would have caught the miss itself is the emitter/caller coupling guard filed as #3297, which does not exist yet."*

---

## Arm (4b) is a genuine hand-constructed positive instance — answering your direct question

Yes, and it is not the vacuous null one level up. The distinction is where the instance comes from: your first check read a *corpus* and found an absence; arm (4b) reads no corpus at all — it constructs the case from `register_builtin_hle()`, which is the machinery that creates guest-ABI entries, and then asks the predicate about those five NIDs by name. Mutation-checked by reading `dispatch.cpp:160-162`:

* `registered` -> `return true;` — caught by `:302` (`no-such-nid-xyz`).
* `registered` -> `return false;` — caught by the loop at `:292-296` (`known == 0`) and by `:299`.
* `registered` -> `find(nid) != end() && !it->second.guest_abi` (the actual bug class this exists for) — caught by the loop, and *only* by the loop, which is why the loop is the load-bearing arm.
* Empty-iteration failure — caught by `known == 5` at `:297`.

Three independent mutations, three different arms, no arm that passes for the wrong reason. This one is sound.

## Also checked and cleared at this head

* **`Hle::registered` is correct, and is the right predicate.** `dispatch.cpp:160-162` keys on key-existence alone, so it is independent of both the handler value and `guest_abi` — which is exactly the third question.
* **The one place `registered` and `lookup_address(nid) != nullptr` could diverge is unreachable.** They differ only for an entry whose `fn` is null, and no such entry can exist: the only two `registry()[nid]` uses (`dispatch.cpp:123`, `:131`) both run *after* `register_fn` inserted the key, and no registration passes a null handler. Worth knowing rather than acting on: `nid_census`'s exact ground truth is the emitter's own `if (fn)` test, i.e. `lookup_address != nullptr`, so if a null-`fn` registration ever becomes possible the census under-reports. Today they coincide.
* **Full re-sweep at `7f4d0ddf5`.** Outside `prosper/tests`, the registry accessors are now: `exec_image_linux.cpp:3064` and `exec_image_win.cpp:1202` (`lookup_address`), `nid_census.cpp:279`/`:308` (`registered`), and `frontends/audio_sdl3/test_audio_sdl3.cpp:24` (`lookup`, correct there, message conflation filed in #3297). No new callers, no missed conversions.
* **`nid_census.cpp:20` updated**, so the stated mechanism matches the code again — that was the invariant I quoted at you, and leaving it saying `Hle::lookup` would have made the header false in the opposite direction.
* **`test_hle_registered.cpp:22` and `:62`** both converted, comment updated to name `Hle::registered`.
* **`:311-317` comment correction applied and now accurate.** Seven integer-class arguments -> one spills; the single `4.5` occupies 1 of 8 SSE slots and does not spill; `check_executed` named as owning the FP overflow path. Matches the actual call at `:319`.
* **PR body arm-(3) correction** is present and is the right correction — it states the gap rather than deleting the claim, and #3297's failure scenario (steps 1-5) checks out against `exec_image_linux.cpp:3064-3072`.
* **Everything I cleared in the previous review still stands** and is untouched by this diff: the alias-template attribute result (GCC 16.2.1, `ms_abi` proxy, `is_same` **and** codegen), `kSpecs` 11 rows with no directives in the initializer, the `#if defined(_WIN32)` block at `:533-639`, `++probed` making `probed == 11` a real property, M5 genuinely VOID rather than negative, and `ci.yml:632`/`:666`.

I did not rebuild, re-run your suite, or touch the GPU. The only work I did was reading, a `.dynstr` scan of the dump corpus, and the Python `nid_hash` reimplementation used to drive it.

**Flip condition: correct the two claims — the corpus number in the PR body and #3297, and the arm-(4b) comment. No code change required, and on the code I would approve as it stands.**

mattias800 (commented):
Third pass, head `a17fd11ef` (previous rounds: `ea7e1896a`, `7f4d0ddf5`).

**APPROVED.** Both blocking findings are closed. Two non-blocking items below, and one of them is a correction to something you told me was already done — please read that one before merging even though it does not block.

Two of the corrections in this review are **mine**, not yours: my proposed mechanism for blocking 1 was wrong, and one of my follow-up experiments was invalid. Both are written out below rather than quietly dropped.

---

## Force-push: content verified identical, not taken on trust

Per-commit `git patch-id --stable` across the rewrite:

```
old 7f4d0ddf5   da3469c168b4  hle: close the CALL side of the guest-ABI boundary
                34bde548af56  hle: add Hle::registered() and fix the third missed caller
new a17fd11ef   da3469c168b4  (identical)
                34bde548af56  (identical)
                a49df33b4fe6  tests: correct the arm (4b) comment   <- new
```

Whole-branch patch-id differs, as it must after a rebase onto `e05d71301`; per commit they match exactly. `git log origin/main..a17fd11ef --format=%B | grep -c '^Claude-Session:'` -> **0**, with 3 `Co-Authored-By:` retained; the PR body carries 0 badge/session lines. `git diff origin/main a17fd11ef --stat` is the same 9 files, so the rebase dragged nothing in.

## Blocking 2 — closed, and closed the right way

`test_guest_varargs.cpp:285-294` no longer claims coverage it lacks. It now states the negative explicitly — *"This arm does NOT close that hole, and saying so is the point"* — names why (property of `Hle::registered` vs. `nid_census` calling `lookup`), and points at #3297. Stating the gap is strictly better than deleting the claim, and the commit message recording that the same overclaim went in twice is the part that will actually stop a third.

## Blocking 1 — your diagnosis is right, mine was wrong, and I re-measured rather than taking it

I ran the already-built `nid_census` from your worktree (no rebuild), with `--registered --tsv` on both invocations:

```
app0 DIRECTORY   809 NIDs / 7 modules / 2606 sibling-excluded   printf family ABSENT
EBOOT ALONE      612 NIDs / 1 module  /    0 sibling-excluded   printf family PRESENT, registered=1
```

`--registered` was passed in **both**, so `:281` cannot be what suppressed the rows — my proposed mechanism was wrong and yours is right. **`nid_census.cpp:264` is the cause**, and I confirm the code reading too: `if (title_exports.count(im.nid)) { satisfied_cross_module++; continue; }` sits at `:264`, ahead of the registration check at `:279`.

Reach, sampled across six titles on the eboot-alone invocation — 5, 5, 4, 4, 1, 1 printf-family rows for PPSA24651 / PPSA04263 / PPSA20800 / PPSA03130 / PPSA16901 / PPSA02739 — consistent with the 55/55 `.dynstr` scan from my last review. And the counterfactual: the **pre-PR** binary (`origin/main` `e05d71301`) on the eboot alone shows the same rows with `registered=1`, so commit 1 flipped exactly those to 0. The defect was reachable.

### A correction I owe: one of my own checks was invalid

Trying to identify which sibling exports `printf`, I passed pairs of module paths (`nid_census eboot.bin libc.prx`) and read 0 exclusions from every pair. **That is not a valid A/B and I nearly filed a finding from it.** `nid_census.cpp:238` is `for (const auto& input : inputs)`, and `loaded` / `title_exports` are rebuilt **per input** — so N explicit paths are N independent one-module "titles" and the cross-module pass never runs at all. Measured: 8 explicit paths -> 3266 NIDs, **0** exclusions, printf present. Anyone repeating my check must put the modules in a *directory*; the command line cannot express a multi-module title.

---

## Non-blocking 1 — #3297's BODY is not corrected, though you told me it was

`gh issue view 3297` `updatedAt` is 17:36:00, which is the timestamp of the **comment**. The body is unchanged, and lines 39-46 still read:

> `test_trap_linux.cpp` already does the mapped-memory + `install_stubs` dance against a real dump, so the machinery exists; the awkward part is that no local dump imports a printf-family NID (**0 of 55**, measured while fixing #3272), so the slot list has to be synthesised rather than taken from an eboot.
>
> That synthesis is a feature, not an obstacle — a hand-built positive instance is exactly what the charter asks for when the corpus cannot supply one.

Your correction exists, and it is good, but it exists only as a comment underneath that. This is precisely the propagation shape the finding was about: whoever picks up #3297 implements from the **Suggested fix**, not from the thread — and that section still tells them to synthesise a slot list on a premise that is false, while your comment says the opposite. One of the two is wrong and the body is the one that gets read first.

Not blocking, because it lives in a separate artifact that can be edited at any time and the PR's own record is correct. But please `PATCH` the body before this closes, and **verify afterwards rather than trusting the exit path**: `gh issue edit` / `gh pr edit` fail here on a Projects-classic GraphQL error and silently leave the body unchanged, which is the most likely explanation for you believing it was done. `gh api -X PATCH repos/OWNER/REPO/issues/3297 -f body=...` then re-read it.

## Non-blocking 2 — the `:157-158` citation rests on a stale comment, and the staleness deserves its own issue

The PR body ends the reach argument with: *"The tool's own header at `:157-158` makes the runtime case sharper still: the loader links a fixed set and never auto-links `.sprx`, so a NID the census excludes as sibling-satisfied 'WOULD reach the dispatcher at runtime'."*

That divergence no longer exists. **#2199** — *"nid_census excludes cross-module bindings using a module set the loader would not link — false absences"* — was closed on 2026-08-07, and the fix is visible in the code you are citing around:

* `nid_census.cpp:141` builds the module set from `prosper::boot_link_inputs(input, /*verbose=*/false)`, i.e. the loader's own link set, not a tree walk;
* `boot_program.hpp:156-159` says so explicitly: *"The loader's actual link set for a dump root (#2199). boot_program() calls this, so a second consumer -- **nid_census, whose cross-module exclusions are only sound over the set the loader really links** -- cannot drift from it."*

Measured independently: the PPSA24651 tree holds 8 module candidates (eboot + 6 `.prx` + `sce_sys/about/right.sprx`) and directory mode reports **7 modules read** — the `.sprx` dropped. That is link-set behaviour, not a tree scan.

So in directory mode the exclusion is **sound**, not false, and the sentence argues the opposite of what is now true. Your conclusion does not need it and is better without it: reach stands on the eboot-alone invocation, and on the sharper point that pre-fix the `registered` column read **0 for a registered handler** — a wrong answer about the registry, independent of whether that NID ever reaches the dispatcher. I would drop the sentence.

**The staleness itself is worth filing separately, and it is bigger than a comment.** `nid_census.cpp:155-158` is the header text you quoted, but `:183-186` prints the same claim in **every report**:

```
NOTE: module set is a tree scan, not the loader's link set (boot_program.cpp): ...
      Cross-module exclusions are computed over THIS set.
```

A tool whose own header says every report it emits is *"an ASSERTION OF ABSENCE"* is emitting a stale, inverted caveat about the population that absence was measured over, to every consumer, on every run. That is the same class as the thing this PR fixed, one layer up, and it is not this PR's to fix.

---

## Cleared at this head

* Accessor sweep unchanged and still clean: outside tests, `exec_image_linux.cpp:3064` / `exec_image_win.cpp:1202` (`lookup_address`), `nid_census.cpp:279`/`:308` (`registered`), `frontends/audio_sdl3/test_audio_sdl3.cpp:24` (`lookup`, correct there, filed in #3297).
* `nid_census.cpp:20` still names `Hle::registered`, so the header and the code agree.
* Everything cleared in rounds 1 and 2 is untouched by this diff and stands: the alias-template attribute result (GCC 16.2.1, `ms_abi` proxy, `is_same` **and** codegen), `Hle::registered`'s implementation and the proof that it cannot diverge from `lookup_address` today, arm (4b) as a genuine hand-constructed positive instance with three distinct mutations caught by three different arms, `kSpecs` at 11 rows with no directives in its initializer, the `#if defined(_WIN32)` block at `:533-639`, `probed == 11` as a real property, M5 genuinely VOID rather than negative, and `ci.yml:632`/`:666`.

I did not rebuild, did not run your suite, and did not touch the GPU. The measurements above come from reading, a `.dynstr` scan of the dump corpus, and running two `nid_census` binaries that already existed on disk.

**APPROVED** on head `a17fd11ef`. Merge is yours; please land the #3297 body edit alongside it.

Issue #2981, opened 2026-08-24
Wwise audio middleware produces silence — AkSoundEngine loads but no audible output (GRIS, Alex Kidd, all Wwise titles)


mattias800:
Multiple titles use Wwise (Audiokinetic) for their audio engine. The Wwise PRXes load correctly, Ajm initializes and registers codecs — but the AudioOut port receives **pure silence** (peak=0.0000, rms=0.0000 across all frames).

## Affected titles

- GRIS (PPSA09804) — confirmed silent in app mode
- Alex Kidd (PPSA02664) — dump missing AkSoundEngine.prx (separate issue, same root cause)
- Likely all titles using Wwise

## Evidence

GRIS boot log (PPSA09804, master + #2972):
```
linked 10 modules (including AkSoundEngine.prx, AkVorbisHwAccelerator.prx, AkMotion.prx)
[ajm] ModuleRegister ctx=1 codec=1
[ajm] ModuleRegister ctx=2 codec=1
[ajm] ModuleRegister ctx=2 codec=24
[ajm] ModuleRegister ctx=2 codec=14
[ajm] BatchInitialize buf=0x203ff90080 size=1800
[audio] port 1: peak=0.0000 rms=0.0000 (f32 stereo, every frame)
```

The Wwise engine initializes, registers codecs with Ajm, and processes batches. AudioOut ports open and receive data. But the PCM data is **all zeros**.

## Root cause candidates

1. **Wwise initialization gap** — Wwise needs specific PS5 system calls or callbacks during initialization that prosper doesn't provide (e.g. memory allocator registration, thread priorities, I/O routing)
2. **Sound bank loading** — Wwise sound banks (.wem/.bnk files) may need a prosper-side file loader or I/O hook that isn't implemented
3. **Codec gap** — Ajm codec types 24 and 14 are registered but not in prosper's supported enum (Mp3=0, Atrac9=1, Aac=2). If Wwise uses these codecs for decoding, the decode fails silently
4. **Event/buffer lifecycle** — Wwise's event system posts audio events that reference sound banks; if the event routing or buffer management isn't emulated, no audio data reaches the output

## What works

- Wwise PRXes load and link (AkSoundEngine.prx = 3.8 MB, AkVorbisHwAccelerator.prx, AkMotion.prx)
- Ajm ModuleRegister + BatchInitialize succeed
- AudioOut ports open and receive data at real-time pace
- The SDL3 audio sink delivers to Pipewire correctly

## What doesn't work

[...]

PR #3294, opened 2026-09-03
env: refuse a malformed value on the 23 PROSPER_* knobs where 0 is a setting, not "off" (#3267)

mattias800:
Refs #3267. Converts the **23 sites where a malformed value removes a bound or disables a
default-ON guard**, and catalogues the rest on the issue rather than sweeping 164 sites in one diff.

> **Head `0edea136` addresses the review of `d0d6c9c5`.** Blocking finding **B1** is fixed: bucket D
> was derived against two of the three malformed rows, it contained a genuine bucket-A site, and
> both of the sites it wrongly cleared are now converted. All five non-blocking items are addressed
> too, including the base-0 grammar loss (**N1**) and the gate's coverage hole (**N2**). Every place
> a number or claim changed is marked *"corrected after review"* below. The original 21 conversions
> are untouched.

## The census, re-derived on current `main`

`b00a1c05`, over `prosper/{src,frontends,tools,tests}`, counting only parses whose input traces to
`getenv`/`PROSPER_ENV_VALUE`:

| | count |
| --- | --- |
| discard the end pointer (or `atoi`-family) | **164** |
| check an end pointer | **72** |

Close to the issue's 169/71 and reached independently; the small gap is window width (8 lines vs 6)
and the two sites #3276 and #3253 already converted.

## The classification, which is the actual deliverable

The issue's own instruction is that this is triage, not a sweep. Buckets are by **what a wrong value
does**, not by how it parses:

| bucket | count | |
| --- | --- | --- |
| **A** — a malformed value REMOVES a bound, or disables a default-ON guard | **23** | converted here |
| **B** — it silently CHANGES a production limit that stays in force | 35 | catalogued on the issue |
| **C** — it only moves a diagnostic's window or level | 57 | catalogued |
| **D** — refuses **all three** malformed rows and keeps the default | **1** | not a defect |
| **H** — hex / signed / float grammar the decimal helper cannot express | 48 | needs helper work first |

> **Corrected after review.** The first revision of this table read A 21 / B 23 / C 49 / D 23 / H 48.
> D had been derived against only two of the issue's three malformed rows — inputs that parse to
> **0**, and **`-1`** — and not against the third, a **numeric prefix** (`8mb` → `8`), which the
> issue body itself calls the nastiest. Re-derived against all three, **D collapses from 23 to 1**:
> `PROSPER_ZEROWATCH_FIND_PREFIX` is the only site in the tree whose guard rejects a prefix parse
> too (its `16..256` range refuses the `8`, loudly). Two of the old D members were bucket **A** —
> both are converted here — twelve are B and eight are C.

The full per-site listing for B, C, D and H is posted on #3267.

## What is in bucket A, and why

**A bound whose zero means "unbounded."** `PROSPER_WRITE_WATCH_MAX_KB` is the sharpest one, and it is
in `src/host/memory/guest_write_watch.cpp` — the write-watch implementation #3253 was filed about,
one file from the six knobs that PR fixed, and missed by it. `=8mb` selected an 8 KiB cap, i.e. no
range large enough to get a watch at all. The MiB-denominated caches (`PROSPER_COMPUTE_BUFFER_CACHE_MB`,
`_IMAGE_CACHE_MB`, `_MEMORY_POOL_MB`, `PROSPER_BACKEND_TARGET_CACHE_MB`, `_TEXTURE_CACHE_MB`,
`PROSPER_MEMORY_POOL_MB`, `PROSPER_BACKEND_BUFFER_ARENA_KB`, `PROSPER_PIPELINE_LAYOUT_CACHE_ENTRIES`,
`PROSPER_ARRAY_DECODE_BUDGET_MIB`) all saturate `-1` to `UINT64_MAX`, which their own overflow guard
then turns into "no limit".

**A multiply that wraps.** `PROSPER_DMEM_BUDGET_MB`'s `>= 1024` floor already refused small typos.
What it could not refuse was `-1`: `strtoull` saturated, and `mib * 1024 * 1024` then wrapped — on
the advertised direct-memory budget the guest sizes its pools from, which is the #2934 class exactly.

**A default-ON guard switched OFF by making it explicit.** Five knobs read
`!e || strtol(e, nullptr, 0) != 0`. `=yes`, `=true`, `=on` — what a person types to pin a default-ON
switch on — every one answers 0, i.e. **OFF**. `PROSPER_REL1_FORGE_GUARD` is the #312 ROOT fix and
`PROSPER_REL1_STOMP_GUARD` its stomp guard; `PROSPER_NETCTL_CB` does not merely log less, it leaves
the whole NetCtl NID family **unregistered**, returning the guest to pre-#306 behaviour. Nothing in
the run says so. (`PROSPER_MB3_CENTRAL_SCAN` and `PROSPER_MB3_TRACK_TLS` are the same shape.)

**A bound refused only against a zero.** Two sites guard with `n ? n : <default>` or
`if (parsed)`, which rejects an input parsing to 0 **and nothing else** — so `=-1` saturated, stayed
non-zero, and removed the bound. `PROSPER_BACKEND_TARGET_CACHE_COUNT` lost the resident-target count
bound entirely (`SIZE_MAX` targets); it is the **count twin of `PROSPER_BACKEND_TARGET_CACHE_MB`**,
which this PR converts, ~80 lines away in a file it already edits. `PROSPER_VDEC2_DUMP_FRAMES` lost
a **disk** bound — its own comment puts the 16-frame cap at 50 MB against 5 GB uncapped.

Plus three where the wrong value reads as a different defect entirely:
`PROSPER_MAX_DISPATCH_GROUPS` (a cap only ever set in order to impose one — `750,000` became a cap of
**750**), `PROSPER_MUTEX_FAIR_US` (0 removes the yield and re-starves GRIS's bank commit, #2981), and
`PROSPER_PAD_FRAME_HOLD` / `PROSPER_PAD_READ_HOLD` (a zero-length press the guest never observes,
which looks exactly like a render or progression wall — the recorded input-mapping trap).

## Contract

**Corrected after review.** The first revision said "unchanged for every well-formed value". That
was not accurate: five default-ON guards and `PROSPER_MAX_DISPATCH_GROUPS` read `strtol(e, …, 0)` —
**base 0** — so `0x2000` was well-formed at all six, and the decimal-only grammar refused it. On
`PROSPER_MAX_DISPATCH_GROUPS` the fallback is `0` = **no cap**, so the fix for "a typo removes the
bound" introduced exactly that outcome on the one knob whose only purpose is to impose a bound.

`env_numeric.hpp` therefore gains `parse_u64_auto_base` and an `_auto` family accepting `[0-9]+` **or**
`0x[0-9a-fA-F]+` — no wider than what those six sites already took, and deliberately **without**
base-0's octal leg, because `010` meaning 8 is its own silent-wrong-setting hazard.
`PROSPER_REL1_FORGE_GUARD=0x0` means OFF again, and `PROSPER_MAX_DISPATCH_GROUPS=0x2000` caps at
8192 again. Both are pinned by arms whose `good` column is the hex spelling.

With that in place: **unchanged for every well-formed value.** Unset and empty still take the default
in silence — they are not typos. A malformed value prints one line naming the variable, the text and
the default it is keeping, and changes nothing.

On the two knobs whose default **is** the permissive sentinel (`PROSPER_WRITE_WATCH_MAX_KB`'s `0` =
unbounded, `PROSPER_MAX_DISPATCH_GROUPS`'s `0` = no cap), the line now says so —
`keeping the default (0, unbounded: no range is too large to watch)` — because telling an operator
who typed a value *in order to impose a bound* that nothing changed is true and useless. And on the
subset of typos that parse to **0** on those two, the change is the **message alone**: the fallback
is what `strtoull` already answered. That is recorded in the catalogue and in `kSites`.

## Tests

`tests/diagnostics/test_env_numeric_sites.cpp` — three arms per row, **35 rows over 30 distinct
knobs** (the eight #3253 converted included), **107 assertions**. *(The first revision said "29
knobs"; it was 29 rows over 27 knobs.)*

Each arm **mirrors** its site rather than calling it. That is deliberate: nearly every converted site
caches its read in a function-local `static`, so an arm that armed the variable with `setenv` would
go **vacuous** rather than red once the static was initialised — the #2214 shape that
`tools/env/check_cached_env.py` exists to refuse. No `PROSPER_*` variable is set, unset or read by
this test.

Each arm also computes what the site's **old** spelling answered and asserts it **differed**, so an
arm on a knob that never needed converting fails instead of passing quietly. That caught three of my
own arms being void while writing this (`PROSPER_TEXTURE_WRITE_WATCH_PROMOTE_MB` with `8 MB`,
`PROSPER_COLD_STORAGE_SNAPSHOT_MIN_MB` with `16mb `, `PROSPER_MAX_GPU_COMPARE_IMAGE_MB` with
`2,048` — in each case `strtoull` happened to answer the default, so the arm was asserting nothing).

`tools/env/check_env_numeric_arms.py` (ctest `env_numeric_arms`) closes the drift a mirror invites:
it fails if a knob is parsed through `env_numeric` with no arm, or armed with no call site left. Its
first run reported **#3253's own eight sites had no per-site arm** — `test_write_watch_policy.cpp`
pins the helper's grammar, not each site's fallback — which is why those eight are armed here too.

**Also corrected after review:** the gate's pattern knew only two of the header's entry points, so
`PROSPER_COMPUTE_IMAGE_CACHE_MB` — the one bespoke conversion, with a code-path fallback rather than
a numeric one, i.e. the site *most* likely to drift — was invisible to it while its success line read
`27 knob(s) … every one armed` with **28** parsed. That site now calls a named helper
(`env_u64_or_report`) instead of a hand-written `fprintf`, the pattern matches every name-taking
entry point, and the line prints the **parsed** count. Mutation **M4** below reproduces the old hole.

Three mirrors also differed from their sites in cap or unit (`UINT64_MAX` vs `SIZE_MAX`, `"presses"`
vs `"flips"`/`"reads"`). Identical on any 64-bit build and message-only — but the stated residual
risk is precisely "a value changed at one end only", so they are now exact.

**What this does NOT pin**, stated in the test's header: a fallback value changed at one end only
still needs a reader. The gate can check names mechanically; it cannot check `16ull * 1024ull`
against `16 * 1024` without becoming the kind of baseline people regenerate instead of reading.

## Verification (exact head `0edea136`, Linux, distrobox `ps5ys`)

```
cmake --build prosper/build-linux -j4        BUILD_RC=0
ctest --no-tests=error -j4                   CTEST_RC=0   100% tests passed out of 355
```

Baseline on `origin/main` `b00a1c05`, same tree, same flags: `BUILD_RC=0`, `CTEST_RC=0`,
**100% passed out of 353**. The two new cases are `diagnostics_env_numeric_sites` and
`env_numeric_arms`.

### Mutation arms — by exit code, not by assertion

| | site test | `env_numeric_arms` |
| --- | --- | --- |
| control (this head) | **exit 0** | **exit 0** |
| M1 — `parse_u64_strict` reverted to a bare `strtoull` | **exit 1**, 33 of 107 arms red | — |
| M2 — one converted site's arm deleted from the table | — | **exit 1**, names `PROSPER_NETCTL_CB` |
| M3 — one production site reverted to a bare `strtoull` | — | **exit 1**, names the now-stale arm |
| **M4 (new)** — the gate's pattern narrowed so a header helper is unknown to it | — | **exit 1**, names `PROSPER_COMPUTE_IMAGE_CACHE_MB` |

M1 mutates a shadow copy of `env_numeric.hpp` ahead of `src` on the include path; M2–M4 mutate a
file, run the scan, and restore from a copy. None touches the build directory. M4 is the regression
guard for N2: it reproduces the exact hole the review found, so narrowing the pattern again fails.

No GPU was used: no `prosper-app`, no `screenshot`, no `gpu_replay`, no snapshots.

## Risk

The behaviour change is confined to values that were **already** being misread — the base-0 regression
the review found is now closed rather than disclosed. The one judgement call worth a reader is the
default-ON family: `=0` and `=0x0` both still mean off (the documented spelling, and what
`docs/DOLL_LOADING_PROGRESSION.md`, `tools/dbg/forge_prove.sh` and the A/B comments use), and only
text that is neither decimal nor `0x`-hex changes meaning — from "off" to "keep the default, loudly".

The reviewer's precision nit is taken: `PROSPER_NETCTL_CB` guards the **three callback NIDs**
(`RegisterCallback`, `CheckCallback`, `GetState`), not the whole NetCtl family — `GetInfo` and
`GetResult` register unconditionally. Corrected in the code comment and the catalogue.

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -373,6 +373,15 @@ if(Python3_Interpreter_FOUND)
   add_test(NAME cached_env_arming_logic
            COMMAND "${Python3_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/tools/env/check_cached_env.py"
                    "${CMAKE_SOURCE_DIR}")
+  # The coverage half of test_env_numeric_sites (#3267). That test MIRRORS each converted site
+  # rather than calling it -- the sites cache their reads in function-local statics, so a runtime
+  # arm would go vacuous exactly as cached_env_arming_logic above describes -- and a mirror drifts
+  # silently: convert a site, forget the arm, suite stays green. This refuses that by comparing the
+  # set of knobs parsed through diagnostics/env_numeric.hpp against the set the test arms. It is a
+  # text scan, so it needs nothing built.
+  add_test(NAME env_numeric_arms
+           COMMAND "${Python3_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/tools/env/check_env_numeric_arms.py"
+                   "${CMAKE_SOURCE_DIR}")
   # stdout carries BYTES, and Windows decodes them as cp1252/cp437. One em dash in a printf literal
   # therefore reaches a Windows user as three mojibake characters, so a grep for the line the tool
   # itself printed does not match it -- and a test asserting on the rendered string fails on
@@ -1474,6 +1483,16 @@ if(TARGET prosper_core)
   add_executable(test_env_cache tests/diagnostics/test_env_cache.cpp)
   target_include_directories(test_env_cache PRIVATE src)
   add_test(NAME diagnostics_env_cache COMMAND test_env_cache)
+
+  # One arm per PROSPER_* knob #3267 converted to diagnostics/env_numeric.hpp. Header-only and
+  # environment-free by construction: nearly every converted site caches its read in a
+  # function-local static, so an arm that armed the variable at runtime would go VACUOUS rather
+  # than red (#2214). Each arm therefore calls the same helper the site calls, and also computes
+  # what the site's OLD spelling answered -- so an arm on a knob that never needed converting
+  # fails instead of passing quietly.
+  add_executable(test_env_numeric_sites tests/diagnostics/test_env_numeric_sites.cpp)
+  target_include_directories(test_env_numeric_sites PRIVATE src)
+  add_test(NAME diagnostics_env_numeric_sites COMMAND test_env_numeric_sites)
 
   # Diagnostics: enabled-path and JSON tests (only when feature is built).
   if(PROSPER_DIAGNOSTICS)
--- a/prosper/frontends/shared/live/live_compute.cpp
+++ b/prosper/frontends/shared/live/live_compute.cpp
@@ -1045,9 +1045,12 @@ bool persistent_compute_buffer_result_enabled(uint32_t bytes) {
 
 VkDeviceSize persistent_compute_buffer_limit() {
     static const VkDeviceSize limit = []() -> VkDeviceSize {
+        // A cache CAP: `-1` saturated this to UINT64_MAX, i.e. unbounded, and `1gb` to 1 MiB.
+        // Both look deliberate afterwards; neither is (#3267).
         const char* value = std::getenv("PROSPER_COMPUTE_BUFFER_CACHE_MB");
-        const uint64_t mib = value ? std::strtoull(value, nullptr, 10) : 256ull;
-        if (mib > UINT64_MAX / (1024ull * 1024ull)) return VkDeviceSize{UINT64_MAX};
+        const uint64_t mib = prosper::diag::env_u64_or_default_capped(
+            "PROSPER_COMPUTE_BUFFER_CACHE_MB", value, 256ull,
+            UINT64_MAX / (1024ull * 1024ull), "MiB");
         return static_cast<VkDeviceSize>(mib) * 1024ull * 1024ull;
     }();
     return limit;
@@ -1093,8 +1096,18 @@ VkDeviceSize persistent_compute_image_limit(
     static const std::optional<VkDeviceSize> override_limit = []()
         -> std::optional<VkDeviceSize> {
         const char* value = std::getenv("PROSPER_COMPUTE_IMAGE_CACHE_MB");
-        if (!value) return std::nullopt;
-        const uint64_t mib = std::strtoull(value, nullptr, 10);
+        if (!value || !*value) return std::nullopt;
+        // There is no numeric default to fall back to -- unset means "derive from device memory" --
+        // so a refusal must fall back to THAT, not to a number. Hence parse_u64_strict plus an
+        // explicit nullopt rather than the or_default helper (#3267).
+        uint64_t mib = 0;
+        if (!prosper::diag::parse_u64_strict(value, &mib)) {
+            std::fprintf(stderr,
+                         "[env] PROSPER_COMPUTE_IMAGE_CACHE_MB='%s' is not a plain non-negative "
+                         "count of MiB -- keeping the device-derived budget and changing NOTHING\n",
+                         value);
+            return std::nullopt;
+        }
         if (mib > UINT64_MAX / (1024ull * 1024ull)) return VkDeviceSize{UINT64_MAX};
         return static_cast<VkDeviceSize>(mib) * 1024ull * 1024ull;
     }();
@@ -1244,8 +1257,9 @@ VkDeviceSize compute_memory_pool_limit() {
         // staging allocations. A 256 MiB cache discarded about 200 allocations and repeatedly paid
         // AMD BO page initialization; 640 MiB holds that measured working set with zero discards,
         // while remaining a bounded fraction of the renderer's persistent-target budget.
-        const uint64_t mib = value ? std::strtoull(value, nullptr, 10) : 640ull;
-        if (mib > UINT64_MAX / (1024ull * 1024ull)) return VkDeviceSize{UINT64_MAX};
+        const uint64_t mib = prosper::diag::env_u64_or_default_capped(
+            "PROSPER_COMPUTE_MEMORY_POOL_MB", value, 640ull,
+            UINT64_MAX / (1024ull * 1024ull), "MiB");
         return static_cast<VkDeviceSize>(mib) * 1024ull * 1024ull;
     }();
     return limit;
--- a/prosper/frontends/shared/live/live_compute.cpp
+++ b/prosper/frontends/shared/live/live_compute.cpp
@@ -1097,17 +1097,15 @@ VkDeviceSize persistent_compute_image_limit(
         -> std::optional<VkDeviceSize> {
         const char* value = std::getenv("PROSPER_COMPUTE_IMAGE_CACHE_MB");
         if (!value || !*value) return std::nullopt;
-        // There is no numeric default to fall back to -- unset means "derive from device memory" --
-        // so a refusal must fall back to THAT, not to a number. Hence parse_u64_strict plus an
-        // explicit nullopt rather than the or_default helper (#3267).
+        // There is no numeric default to fall back to -- unset means "derive from device memory"
+        // -- so a refusal must fall back to that CODE PATH, not to a number. env_u64_or_report is
+        // the named helper for that shape; a hand-written fprintf here would be invisible to
+        // tools/env/check_env_numeric_arms.py, which is how the one site with bespoke arithmetic
+        // became the one site the anti-drift gate could not see (#3267 N2).
         uint64_t mib = 0;
-        if (!prosper::diag::parse_u64_strict(value, &mib)) {
-            std::fprintf(stderr,
-                         "[env] PROSPER_COMPUTE_IMAGE_CACHE_MB='%s' is not a plain non-negative "
-                         "count of MiB -- keeping the device-derived budget and changing NOTHING\n",
-                         value);
+        if (!prosper::diag::env_u64_or_report("PROSPER_COMPUTE_IMAGE_CACHE_MB", value, &mib, "MiB",
+                                              "the device-derived budget"))
             return std::nullopt;
-        }
         if (mib > UINT64_MAX / (1024ull * 1024ull)) return VkDeviceSize{UINT64_MAX};
         return static_cast<VkDeviceSize>(mib) * 1024ull * 1024ull;
     }();
--- a/prosper/frontends/shared/live/live_renderer.cpp
+++ b/prosper/frontends/shared/live/live_renderer.cpp
@@ -777,8 +777,13 @@ thread_local TextureDecodeScopeStats g_texture_decode_scope{};
 // that report depend on two switches, which check_diag_gates.py rightly rejects (TWO-GATE).
 inline uint64_t array_decode_budget_bytes() {
     static const uint64_t bytes = [] {
+        // atoll turned `-1` into a budget of 0xFFFFFFFFFFF00000 -- i.e. UNBOUNDED, the one setting
+        // this knob exists to prevent -- and `1gib` into 1 MiB, which reads in the log as a
+        // deliberate choice. Refuse both and keep 1 GiB (#3267).
         const char* e = getenv("PROSPER_ARRAY_DECODE_BUDGET_MIB");
-        return static_cast<uint64_t>(e ? atoll(e) : 1024) << 20;
+        const uint64_t mib = prosper::diag::env_u64_or_default_capped(
+            "PROSPER_ARRAY_DECODE_BUDGET_MIB", e, 1024ull, UINT64_MAX >> 20, "MiB");
+        return mib << 20;
     }();
     return bytes;
 }
--- a/prosper/src/diagnostics/env_numeric.hpp
+++ b/prosper/src/diagnostics/env_numeric.hpp
@@ -24,6 +24,11 @@
 // whitespace, no `0x`, no suffix. `strtoull` would take a leading space and a leading `-` (wrapping
 // it to a huge unsigned), and both are far likelier to be a typo than an intention. An UNSET or
 // EMPTY variable is not a typo and takes the default in silence.
+//
+// The `_auto` family below widens that to `0x`-hex for the six sites whose PRE-EXISTING spelling was
+// `strtol(e, nullptr, 0)`. That is not a relaxation of the rule but an application of it: on those
+// sites `0x2000` was already a valid input, so refusing it would be this header changing a
+// well-formed setting — the very thing it exists to prevent (#3267 N1).
 #pragma once
 #include <cstdint>
 #include <cstdio>
@@ -53,16 +58,24 @@ inline bool parse_u64_strict(const char* text, uint64_t* out) {
 // `unit` is an optional trailing note for the message ("KiB", "MiB", "validations", ...), because a
 // knob's units live at its call site and a reader who mistyped one wants to be told which was
 // expected. It is rendered as "... is not a plain non-negative count of KiB", so pass a bare noun.
+//
+// `default_note` glosses the default in the message, and exists because on two knobs the default is
+// itself the PERMISSIVE sentinel: `PROSPER_WRITE_WATCH_MAX_KB`'s 0 means "no limit" and
+// `PROSPER_MAX_DISPATCH_GROUPS`'s means "no cap". Telling an operator who typed a value in order to
+// IMPOSE a bound that we are "keeping the default (0) and changing NOTHING" is true and useless --
+// they need to know they still have no bound. Pass "0 = unbounded" and the line says so.
 inline uint64_t env_u64_or_default(const char* name, const char* text, uint64_t fallback,
-                                   const char* unit = nullptr) {
+                                   const char* unit = nullptr,
+                                   const char* default_note = nullptr) {
     uint64_t value = 0;
     if (parse_u64_strict(text, &value)) return value;
     if (!text || !*text) return fallback;   // unset or empty: not a typo, and not worth a line
     std::fprintf(stderr,
-                 "[env] %s='%s' is not a plain non-negative %s%s -- keeping the default (%llu) and "
-                 "changing NOTHING\n",
+                 "[env] %s='%s' is not a plain non-negative %s%s -- keeping the default (%llu%s%s) "
+                 "and changing NOTHING\n",
                  name, text, unit ? "count of " : "integer", unit ? unit : "",
-                 static_cast<unsigned long long>(fallback));
+                 static_cast<unsigned long long>(fallback),
+                 default_note ? ", " : "", default_note ? default_note : "");
     return fallback;
 }
 
@@ -73,9 +86,83 @@ inline uint64_t env_u64_or_default(const char* name, const char* text, uint64_t 
 // the tree scales its number by 1024 or 1024*1024, and a value near UINT64_MAX would wrap that
 // product to something small — the same class of silent wrong setting this header exists to remove.
 inline uint64_t env_u64_or_default_capped(const char* name, const char* text, uint64_t fallback,
-                                          uint64_t cap, const char* unit = nullptr) {
-    const uint64_t value = env_u64_or_default(name, text, fallback, unit);
+                                          uint64_t cap, const char* unit = nullptr,
+                                          const char* default_note = nullptr) {
+    const uint64_t value = env_u64_or_default(name, text, fallback, unit, default_note);
     return value < cap ? value : cap;
+}
+
+// --- the same, for a knob whose pre-existing grammar was strtol/strtoul BASE 0 -------------------
+//
+// Six sites read their value with an explicit base of 0, which makes `0x2000` a WELL-FORMED input
+// there. Converting those to the decimal-only grammar above would refuse a spelling that used to
+// work -- and on `PROSPER_MAX_DISPATCH_GROUPS`, whose fallback is "no cap", refusing `0x2000` would
+// newly introduce the exact "a typo removes the bound" outcome this header exists to remove. So the
+// grammar is widened to match what those sites already accepted, and no further: `[0-9]+` or
+// `0x[0-9a-fA-F]+` (either case of the `x`), still no sign, no whitespace, no suffix, still
+// overflow-checked. Base-0's OCTAL leg is deliberately not carried over -- a leading zero in a
+// hand-typed count is far likelier to be padding than an intent to write base 8, and `010` meaning
+// 8 is its own silent-wrong-setting hazard.
+inline bool parse_u64_auto_base(const char* text, uint64_t* out) {
+    if (!text || !*text || !out) return false;
+    if (text[0] != '0' || (text[1] != 'x' && text[1] != 'X')) return parse_u64_strict(text, out);
+    const char* p = text + 2;
+    if (!*p) return false;                       // a bare "0x" is not a number
+    uint64_t value = 0;
+    for (; *p; ++p) {
+        uint64_t digit;
+        if (*p >= '0' && *p <= '9')      digit = static_cast<uint64_t>(*p - '0');
+        else if (*p >= 'a' && *p <= 'f') digit = static_cast<uint64_t>(*p - 'a') + 10u;
+        else if (*p >= 'A' && *p <= 'F') digit = static_cast<uint64_t>(*p - 'A') + 10u;
+        else return false;
+        if (value > (UINT64_MAX - digit) / 16u) return false;   // would overflow
+        value = value * 16u + digit;
+    }
+    *out = value;
+    return true;
+}
+
+inline uint64_t env_u64_or_default_auto(const char* name, const char* text, uint64_t fallback,
+                                        const char* unit = nullptr,
+                                        const char* default_note = nullptr) {
+    uint64_t value = 0;
+    if (parse_u64_auto_base(text, &value)) return value;
+    if (!text || !*text) return fallback;
+    std::fprintf(stderr,
+                 "[env] %s='%s' is not a decimal or 0x-hex non-negative %s%s -- keeping the default "
+                 "(%llu%s%s) and changing NOTHING\n",
+                 name, text, unit ? "count of " : "integer", unit ? unit : "",
+                 static_cast<unsigned long long>(fallback),
+                 default_note ? ", " : "", default_note ? default_note : "");
+    return fallback;
+}
+
+inline uint64_t env_u64_or_default_auto_capped(const char* name, const char* text, uint64_t fallback,
+                                               uint64_t cap, const char* unit = nullptr,
+                                               const char* default_note = nullptr) {
+    const uint64_t value = env_u64_or_default_auto(name, text, fallback, unit, default_note);
+    return value < cap ? value : cap;
+}
+
+// --- for a knob whose "unset" answer is NOT a number ---------------------------------------------
+//
+// `PROSPER_COMPUTE_IMAGE_CACHE_MB` unset means "derive the budget from device memory", so there is
+// no fallback to hand the functions above -- the refusal has to fall back to a different CODE PATH.
+// Returns true only on a well-formed value; reports and returns false on a malformed one; returns
+// false in silence when unset or empty. Exists as a named helper rather than a hand-written
+// fprintf at the call site so that tools/env/check_env_numeric_arms.py can SEE the site: a bespoke
+// parse is invisible to that gate, which is how the one site with bespoke arithmetic became the one
+// site the anti-drift gate did not cover.
+inline bool env_u64_or_report(const char* name, const char* text, uint64_t* out,
+                              const char* unit = nullptr, const char* unset_note = nullptr) {
+    if (!text || !*text) return false;
+    if (parse_u64_strict(text, out)) return true;
+    std::fprintf(stderr,
+                 "[env] %s='%s' is not a plain non-negative %s%s -- keeping %s and changing "
+                 "NOTHING\n",
+                 name, text, unit ? "count of " : "integer", unit ? unit : "",
+                 unset_note ? unset_note : "the default");
+    return false;
 }
 
 } // namespace prosper::diag
--- a/prosper/src/gpu/execute/gpu_executor.cpp
+++ b/prosper/src/gpu/execute/gpu_executor.cpp
@@ -6,6 +6,8 @@
 // binary at startup, or a test via render_runner.h), so prosper_core links this without Vulkan.
 #include "gpu/execute/gpu_execute.hpp"
 #include "gpu/diagnostics/watch_list.hpp"   // strict 0x-only watch parsing (shared with the RTT watch)
+#include "diagnostics/env_numeric.hpp"   // #3267: a typo must not silently drop an operator-set cap
+#include <cstdint>
 #include "gpu/capture/gpu_capture.hpp"
 #include "gpu/timeline/gpu_timeline.hpp"
 #include "gpu/capture/capture_compute_policy.hpp"
@@ -7728,8 +7730,13 @@ ComputeLaunchDimensions resolve_compute_launch(const GpuState::Dispatch& d) {
     // reaches 752,646 workgroups in one dispatch and the device is lost permanently a few submits
     // later; with it, the same route can be run to the same point with the size bounded.
     static const uint32_t group_cap = [] {
+        // 0 disables the cap. That is the right DEFAULT and the wrong answer to a typo: this knob
+        // is only ever set to bound a dispatch that loses the device (#1742/#1743), so `=750000`
+        // mistyped as `=750,000` unbounded the very thing being bounded (#3267).
         const char* value = std::getenv("PROSPER_MAX_DISPATCH_GROUPS");
-        return value && *value ? static_cast<uint32_t>(strtoul(value, nullptr, 0)) : 0u;
+        if (!value || !*value) return 0u;
+        return static_cast<uint32_t>(prosper::diag::env_u64_or_default_capped(
+            "PROSPER_MAX_DISPATCH_GROUPS", value, 0ull, UINT32_MAX, "workgroups"));
     }();
     if (group_cap) {
         auto clamp_groups = [&](uint32_t& groups, uint32_t& threads, uint32_t local) {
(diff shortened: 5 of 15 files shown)


mattias800 (commented):
# Independent review — head `d0d6c9c5`

**REJECTED** on this head. The code change is sound and I would approve it as written; both blocking
findings are cheap and neither requires re-touching the 21 conversions. Re-review after B1 and B2.

I did not build or run ctest (author reports 0 / 355-355, CI green on 9 jobs) and used no GPU. Where
I needed to settle a claim I compiled the one new self-contained test directly (`g++ -std=c++20
-Isrc tests/diagnostics/test_env_numeric_sites.cpp`) and ran the two Python gates against an exported
copy of the head. Everything below that carries a `file:line` is a file I opened.

---

## BLOCKING

### B1. Bucket D contains at least one genuine bucket-A site, and the "23 not defects" claim is derived from the wrong inputs

`tests/fixtures/render_runner.h:2227-2230` (PR head) — `PROSPER_BACKEND_TARGET_CACHE_COUNT`:

    const uint64_t n = value ? strtoull(value, nullptr, 10) : 256ull;
    return n ? static_cast<size_t>(n) : size_t{256};

This is catalogued on #3267 under **D — already refuses; do not convert**, with the note
`n ? n : 256`. That note is true only for inputs that parse to 0. It is an **unsigned** parse, so:

* `PROSPER_BACKEND_TARGET_CACHE_COUNT=-1` → `strtoull` saturates → `n = UINT64_MAX` → non-zero →
  the resident-target **count bound is removed entirely** (`SIZE_MAX`). That is bucket A's own
  definition, verbatim: "a malformed value REMOVES a bound".
* `...=8mb` → `n = 8` → eight GPU-resident targets. The site's own comment two lines up
  (`render_runner.h:2224`, "#1177") says an arbitrarily small count is exactly the defect that
  raising the default from 64 to 256 was meant to fix, and every target past the count is forced
  onto the CPU detile path.

It is in a file this PR already edits, ~80 lines above `persistent_texture_cache_limit`, and it is
the **count sibling of `PROSPER_BACKEND_TARGET_CACHE_MB`, which this PR converts**. So the PR
converts the byte cap and publishes the count cap as "not a defect".

The derivation gap is systematic, not a one-off. I spot-checked five D entries and the "already
refuses" note holds against the `0` and `-1` rows but not against the issue's own middle row
(`8mb` → 8), which the issue body itself calls "the nastiest":

| D entry | `-1` | `8mb` |
| --- | --- | --- |
| `render_runner.h:2230` `PROSPER_BACKEND_TARGET_CACHE_COUNT` | **unbounded (A)** | count 8 (B) |
| `live_renderer.cpp:1149-1152` `PROSPER_MAX_BUFFER_UPLOAD_MB` | safe (>64 → ceiling) | 8 MiB ceiling instead of 64 (B) — and this knob exists to reproduce the #1427 collapse |
| `command_processor.cpp:3492-3494` `PROSPER_WAIT_TIMEOUT_MS` | safe (signed, `>0`) | 8 ms; the comment at `:3487` records that at 50 ms "the MallocBinned3 fatal returned" |
| `hle_file.cpp:3185-3192` `PROSPER_ZEROWATCH_FIND_PREFIX` | safe | safe — genuine loud refusal, correctly cited as the model |
| `tile.cpp:516-520` `PROSPER_RX_PIPES` | safe | 8 is a power of two, so accepted as a real setting (B) |

Also worth a second look on the same basis: `hle_service.cpp:490-492` `PROSPER_VDEC2_DUMP_FRAMES`
(`strtoul` base 0, `if (parsed) d.limit = parsed`) — `-1` lifts the documented 16-frame dump cap to
~4.29e9.

**Why this blocks rather than being a nit.** The PR's own framing is that "the classification ... is
the actual deliverable", and its wording is that any future sweep "will churn these 23 for no gain".
Published as-is, the opposite happens: a future sweep reads D and *skips* a site that removes a
bound. That is the expensive-knowledge failure the charter's `## Ruled out` discipline exists to
prevent, and the record here is the #3267 comment plus the PR body.

**To clear:** correct the #3267 catalogue and the PR body's bucket table — say explicitly that D was
derived against the 0/negative rows and that D's members are only safe against those, move
`PROSPER_BACKEND_TARGET_CACHE_COUNT` to A (converting it here would be one line in a file already
open, which I would prefer), and move the others named above to B. No change to the 21 is needed.

### B2. Charter attribution rules are violated in both the commit and the PR body

`CLAUDE.md:931-933` on `origin/main`: *"Do NOT add 'Generated with Claude Code' attribution lines
(the badge, 'Generated with [Claude Code](...)' footers, session links) to PR bodies, commit
messages, issue text, or comments — anywhere."*

* The PR body ends with the robot badge, the "Generated with [Claude Code]" footer and a session
  link.
* `git log b00a1c05..d0d6c9c5 --format=%B | grep -c '^Claude-Session:'` returns **1**; the charter
  states this gate must read 0, and that amending is not enough because the harness re-adds it —
  use the `git rebase --exec` or branch-rebuild recipe it gives, then `--force-with-lease`.

Mechanical, no re-review of the code needed.

---

## NON-BLOCKING

### N1. Six sites lose base-0 parsing, and "Unchanged for every well-formed value" is not accurate

Five default-ON guards read `strtol(e, nullptr, 0)` and `gpu_executor.cpp:7731` on the base commit read
`strtoul(value, nullptr, 0)` — **base 0**, so `0x…` was well-formed at all six. After the change:

* `PROSPER_MAX_DISPATCH_GROUPS=0x2000` used to impose a cap of 8192. It is now refused, the fallback
  is `0ull` (`gpu_executor.cpp:7739`), and 0 means **no cap** — i.e. this PR newly introduces, on one
  knob, the exact "a typo removes the bound" outcome it exists to remove. It is loud, which is the
  saving grace, but it is a behaviour change on a previously-valid spelling.
* `PROSPER_NETCTL_CB=0x0` / `PROSPER_REL1_FORGE_GUARD=0x0` used to mean OFF and now mean ON.

I checked whether anything in the tree relies on this and **nothing does** — see "cleared" below —
so this is disclosure, not risk. Please say so in the PR body's Contract paragraph rather than
leaving "Unchanged for every well-formed value" standing.

### N2. The one bespoke conversion has neither an arm nor gate coverage — and the gate reports success anyway

`live_compute.cpp:1098-1110` converts `PROSPER_COMPUTE_IMAGE_CACHE_MB` with `parse_u64_strict` plus a
hand-written `fprintf` and a `nullopt` fallback (correct, and the added `!*value` check is a real
improvement over the old `strtoull("")` → 0 → a 0-byte image cache).

But `check_env_numeric_arms.py:32` matches only
`env_u64_or_default(?:_capped)?\s*\(\s*\n?\s*"(PROSPER_…)"`, so this site is invisible to it, and it
has no row in `kSites`. I ran the gate against the head: it prints

    [env-arms] ok: 27 knob(s) parsed by env_numeric, every one armed

while **28** knobs are in fact parsed through `diagnostics/env_numeric.hpp`. The one site with
bespoke arithmetic and a non-numeric fallback — the site most likely to drift — is the one the
anti-drift gate cannot see, and the gate's success line asserts otherwise. Cheapest fix: add a
`kSites`-style arm for it (a mirror returning an optional) and either extend `CALL_RE` to
`parse_u64_strict` uses or record the exemption in the script's docstring the way `EXEMPT_SOURCES`
already is.

### N3. On the two knobs whose fallback *is* the permissive sentinel, the refusal line understates itself

Running the new test binary, the emitted stderr is verbatim:

    [env] PROSPER_WRITE_WATCH_MAX_KB='8mb' is not a plain non-negative count of KiB -- keeping the default (0) and changing NOTHING

`guest_write_watch.cpp:1170` turns that 0 into `UINT64_MAX` (unbounded), and
`gpu_executor.cpp:7736-7740`'s 0 means "no cap". So an operator who typed a value *in order to
impose a bound* is told "changing NOTHING" and gets no bound. The value is right — keeping the
documented default is the correct contract — but the message is the whole point of the change on
these two, and "(0)" is the least informative thing it could print. Consider letting the site pass
a short gloss ("0 = unbounded").

Related, and disclosed nowhere: on the subset of typos that parse to 0 (`=eight`), these two knobs
get a message and **no value change at all**, because the fallback equals what `strtoull` already
answered. The `kSites` note at the `PROSPER_WRITE_WATCH_MAX_KB` row correctly explains why `-1` is
excluded as void; the same reasoning applies to `eight` here and is worth one line in the catalogue.

### N4. Counting nit

The PR body and commit say "**29 knobs**". Measured on the head: **29 arms over 27 distinct knobs**
(89 assertions = 29x3 + 2 controls, which does check out).

### N5. Mirror-vs-site divergences (all harmless on a 64-bit target, but they are the disclosed gap doing its thing)

I cross-checked every production `(name, fallback, cap)` triple against its mirror. All fallbacks
match exactly. Three caps/units do not:
`layout_entries_new` uses `UINT64_MAX` where `render_runner.h:889` uses `SIZE_MAX`;
`mib_cap_new<8>` uses `UINT64_MAX/(1024*1024)` where the two `…PROMOTE_MB` sites use
`SIZE_MAX/(1024*1024)`; `pad_hold_new` passes `"presses"` where the sites pass `"flips"` and
`"reads"`. Identical on any 64-bit build and the unit is message-only. Mentioned because the PR's
stated residual risk is precisely "a value changed at one end only", and these show the mirror
already sitting one edit away from the site.

---

## What I checked and cleared

**The #312 claim — verified, and it is the author's, not an inference.** `command_processor.cpp:1411`
on the base commit already reads *"#312 ROOT fix gate (default ON; `PROSPER_REL1_FORGE_GUARD=0` for
A/B baseline)"*, and `:2125` reads *"#312 the WHERE fix (default ON; `PROSPER_REL1_STOMP_GUARD=0`
restores the old barrel-through…)"*. #312 is the DOLL/UE4 `MallocBinned3 Corruption` fatal. So both
guards really are default-ON correctness gates, and `=yes`/`=true`/`=on` really did answer 0 = OFF.

**Nothing in the tree passes a non-decimal value to any converted knob.** Searched `*.sh *.ps1 *.yml
*.yaml *.py *.md *.json *.txt *.bat *.cmake` and then all files, for all 20 names, plus every
`setenv`/`putenv`/`SetEnvironmentVariable`. The complete set of setters is:
`tools/dbg/forge_prove.sh:14` (`PROSPER_REL1_FORGE_GUARD=$GUARD`, `GUARD` defaults to `1`, documented
`GUARD=0` for the A/B), `scripts/diag/run7_netctl.sh:14` and `scripts/diag/run8_matrix.sh:14`
(`PROSPER_NETCTL_CB=1`), and `tests/host/memory/test_guest_write_watch.cpp:102`
(`PROSPER_WRITE_WATCH_MAX_KB="1024"`, set as the first statement of `main()` before the static is
initialised, so it is neither vacuous nor affected). All decimal; no recipe changes behaviour.

**`PROSPER_NETCTL_CB` is not a diagnostic — verified.** `hle_service.cpp:5294-5297`: the `if` body
contains three `R(...)` calls (`sceNetCtlRegisterCallback`, `sceNetCtlCheckCallback`,
`sceNetCtlGetState`). A typo therefore leaves those three NIDs unregistered and the guest gets the
dispatcher's default — the pre-#306 behaviour the comment at `:5288` describes. This is the
strongest item in the PR and it holds. One precision nit: `sceNetCtlGetInfo` (`:5301`) and
`sceNetCtlGetResult` (`:5302`) register unconditionally just below, so "the whole NetCtl NID family"
overstates it slightly — it is the callback trio. Worth correcting in the catalogue since the
finding does not need the overstatement.

**`PROSPER_DMEM_BUDGET_MB` arithmetic — verified, and the fix is stronger than claimed.**
`strtoull("-1")` = 2^64-1; `(2^64-1) * 2^20 mod 2^64` = 18446744073708503040, so the pre-image
advertised ~16 EiB to a guest that sizes its pools from it. Separately — and not claimed in the PR —
a **well-formed** `PROSPER_DMEM_BUDGET_MB=17592186044416` wrapped the same multiply to exactly **0**,
past the `>= 1024` floor. The new cap `UINT64_MAX/(1024*1024)` = 17592186044415 yields
18446744073708503040 with no wrap, so the conversion closes the well-formed wrap too. Both
`kDmemTotal` copies (the Linux and Windows allocator halves of `hle_kernel_mem.cpp:955` and `:3975`)
are converted identically.

**Vacuity — the mirrored-arm design holds.** Compiled and ran `test_env_numeric_sites.cpp`
standalone: 89 `[ok]`, 0 failures, 29 refusal lines on stderr, and no `PROSPER_*` variable is set,
unset or read anywhere in the file. The discriminator (`legacy(malformed) != expected`) is a real
third arm on every row, and I confirmed by hand that the `legacy` mirrors reproduce the pre-image
spellings — including the two easy ones to get wrong, `dispatch_cap_old`'s `strtoul(…, 0)` and
`dmem_old`'s wrap.

**Mutation arms M2 and M3 reproduce exactly.** On an exported copy of the head I deleted the
`PROSPER_NETCTL_CB` row (gate exits 1: *"is parsed by env_numeric at
src/hle/service/hle_service.cpp but has no arm"*) and separately reverted the production site to
`!e || strtol(e, nullptr, 0) != 0` (gate exits 1: *"arms PROSPER_NETCTL_CB, but no call site parses
it with env_numeric any more"*). Restoring both returns exit 0. I did not reproduce M1, which needs
a build.

**The disclosed test gap is acceptable as stated.** "The gate checks names, not arithmetic" is
written in three places (the test header, the script docstring, `CMakeLists.txt:376-384`) and the
reasoning — that a textual `16ull * 1024ull` vs `16 * 1024` comparison becomes a baseline people
regenerate instead of read — is the right call. N2 is the part of that gap that is *not* disclosed.

**No per-frame cost.** Every converted site initialises inside a function-local `static` (or, for
`hle_service.cpp:5294`, one-time registration), so the new `fprintf` can fire at most once per knob
per process. `PROSPER_MUTEX_FAIR_US` sits on the mutex-unlock path but its read is
`static const` (`hle_kernel.cpp:956`).

mattias800:
Head `0edea136` addresses the review of `d0d6c9c5`.

**B1 — fixed.** Bucket D was derived against two of the three malformed rows. Re-derived against all
three (0 / `-1` / numeric prefix), **D collapses from 23 to 1** — `PROSPER_ZEROWATCH_FIND_PREFIX` is
the only site in the tree whose guard rejects a prefix parse, and I had cited it as the model without
noticing it was the only instance of it. Both sites you named move to A and are **converted here**:
`PROSPER_BACKEND_TARGET_CACHE_COUNT` and `PROSPER_VDEC2_DUMP_FRAMES` (which you flagged for a second
look — it lost a disk bound the same way). Twelve more move D→B, eight D→C. Corrected table:
A 23 / B 35 / C 57 / D 1 / H 48. The PR body and #3267 both carry the correction, marked as one, with
the per-site three-row derivation and an explicit note that a guard rejecting `0` is not "already
safe".

**B2 — left exactly as it is, deliberately, and the call is not mine.** You are right about
`CLAUDE.md:931-933`. A session-level directive in my coordinator's harness explicitly supersedes it
for this session's work and was passed to me on purpose; my coordinator has instructed me not to strip
it and is raising the conflict with the project owner. Recording that here so the disposition is
visible rather than looking like the finding was ignored — it is not resolved, it is escalated.

**N1 — fixed rather than disclosed.** You are right that this newly introduced the PR's own hazard on
`PROSPER_MAX_DISPATCH_GROUPS`. `env_numeric.hpp` gains `parse_u64_auto_base` and an `_auto` family
(`[0-9]+` or `0x[0-9a-fA-F]+`), used at all six base-0 sites. Deliberately **no octal leg** — `010`
meaning 8 is its own silent-wrong-setting hazard. `=0x0` means OFF again and `=0x2000` caps at 8192
again, both pinned by arms whose `good` column is the hex spelling. The Contract paragraph no longer
claims "unchanged for every well-formed value" without qualification.

**N2 — fixed, with a regression guard.** `PROSPER_COMPUTE_IMAGE_CACHE_MB` now calls a named helper
(`env_u64_or_report`) instead of a bespoke `fprintf`; the gate matches every name-taking entry point
in the header and prints the **parsed** count (30). A fourth mutation arm reproduces the exact hole:
narrowing the pattern so a header helper is unknown to it → **exit 1**.

**N3 — taken.** Both sentinel-default knobs now pass a gloss:
`keeping the default (0, unbounded: no range is too large to watch)`. Your second half is recorded
too: on typos that parse to 0, those two change nothing but the message, in `kSites` and the catalogue.

**N4 — corrected** everywhere: it is now 35 arms over 30 distinct knobs, 107 assertions.

**N5 — closed.** All three mirror/site divergences (`UINT64_MAX` vs `SIZE_MAX` ×2, `"presses"` vs
`"flips"`/`"reads"`) are now exact.

Also took the NetCtl precision nit — it is the callback trio, not the whole family — in the code
comment and the catalogue.

**Re-verified at `0edea136`:** build `0`; `ctest --no-tests=error -j4` → `0`, **355/355**. Mutations
by exit code: control 0/0 · M1 (`parse_u64_strict` → bare `strtoull`) **1**, 33 of 107 arms red ·
M2 (arm deleted) **1** · M3 (production site reverted) **1** · M4 (gate pattern narrowed) **1**.

Two things your review caught that I want to name, because both are exactly the failure this issue is
about. The first is that I published a "do not convert" verdict derived from a subset of the inputs —
the expensive-knowledge failure, and worse than an unconverted site. The second is smaller and the
same shape: my gate printed a success line asserting coverage it had not measured. Ready for
re-review.

mattias800 (commented):
# Re-review — head `0b295aee` (previously reviewed `d0d6c9c5`)

**APPROVED.** Both blocking findings are resolved, the five non-blocking ones are addressed —
several fixed rather than disclosed — and the new `_auto` code, which is the part with the least
prior review, is correct under independent testing. Six non-blocking findings below; one of them
(NB2) I would like acted on before the bucket-B sweep starts, but none of them should hold this PR.

As before I did not build the project or run ctest. What I ran: `g++ -std=c++20 -Isrc -Wall -Wextra`
on the two self-contained TUs (the site test and a probe I wrote for `parse_u64_auto_base`), the two
Python gates, and the M2/M3/M4 mutations, all against an exported copy of the head. No GPU.

---

## B2 — resolved, verified

`git log e05d71301..0b295aee --format=%B | grep -c '^Claude-Session:'` → **0**. Both commits keep
only `Co-Authored-By:`. The PR body carries no badge, no "Generated with" footer and no session link.

I also checked the force-push did what it says: `git diff b00a1c05 d0d6c9c5` and
`git diff e05d7130 cfc2f9ec` produce the **same patch-id** (`73791e50…`), so the commit I reviewed
last round is byte-identical to the one now on the branch and my prior review transfers unchanged.
Only `0b295aee` is new work.

## B1 — corrected, and more thoroughly than my finding required

D collapsing 23 → 1 is the right answer, and `PROSPER_ZEROWATCH_FIND_PREFIX` turning out to be the
sole instance of the model the author had been citing is the kind of result worth having found. Both
sites I named are now converted (`render_runner.h:2227-2235`, `hle_service.cpp:490-497`), both with
arms, and the generalisation the correction states — *"a guard that rejects `0` is not already safe";
`n ? n : D`, `if (v > 0)`, `if (parsed)` and `std::max(k, …)` all admit the prefix row, and an
unsigned one admits the saturating row too* — is the durable part.

I re-derived the two promotions myself and both hold. `PROSPER_VDEC2_DUMP_FRAMES`'s conversion also
preserves its contract exactly: `VdecDump::limit` already defaults to **16** (`hle_service.cpp:478`),
so the new fallback of 16 on a refusal is a genuine no-op rather than a value the refusal introduces.

---

# Non-blocking findings

### NB1. Bucket C is still unenumerated, and its definition changed under its members

C is not published in either comment ("Listed for completeness in the PR's working notes"), so the
one bucket that **grew** (49 → 57) is the one a reader cannot check. That matters more than it would
have before, because the definition also changed:

* old C — *"harmless: malformed == unset == off"* — a strong property about the value;
* new C — *"moves only a diagnostic's window or level; production behaviour is unchanged"* — a claim
  about blast radius.

Those are not the same claim, and the 49 original members were classified against the old one by the
two-row method. The correction's own rule — *"Any follow-up sweep must evaluate all three rows per
site, and say which rows its verdict covers"* — is applied to D but not to B's original 23 or C's
original 49; those absorbed D's refugees and were otherwise re-labelled, not re-derived. Either
publish C, or say plainly that it is unenumerated and unverified so "57" is not read as 57 audited
sites.

### NB2. Two sites are in no published bucket, and both have bucket-A shape

This is the concrete form of NB1, and the reason I would want it done before the B sweep rather than
"eventually". Neither is a regression from this PR — both are exactly as they are on `main` — so it
is a catalogue gap, not a code defect.

* **`src/hle/kernel/hle_kernel_time.cpp:1612-1614` `PROSPER_WAITCAP`.**
  `const uint64_t cap = e ? (uint64_t)atoll(e) : 0;` then `if (cap && us > cap) us = cap;`.
  `=-1` → `atoll` gives `-1`, the cast makes it `UINT64_MAX`, so `cap` is **non-zero** (the gate
  passes) and `us > cap` is **never** true — the cap the operator armed is silently removed. That is
  `PROSPER_MAX_DISPATCH_GROUPS`'s shape exactly: a permissive sentinel default behind an `if (x)`
  gate, which is bucket A. Its own comment (`:1608`) calls it a diagnostic for bisection runs, which
  is fair — but when armed it bounds a real `sceKernelWaitEqueue` timeout, so `=8ms` → an 8 µs
  ceiling on guest waits is guest-visible, not a log window.
* **`src/gpu/pm4/command_processor.cpp:2937-2943` `PROSPER_POST_SUBMIT_VISIBILITY`.**
  `const int v = e ? (int)strtol(e, nullptr, 0) : -1;` — **unset is `-1`** (follow the SDK version),
  and any malformed text parses to **0**, which is the explicit `FORCED OFF` arm. So malformed !=
  unset, and it cannot be C under either definition. This is the lever in `CLAUDE.md`'s *ArcRunner*
  and *Crisis Core* rows whose verdict is recorded as **undecided** (#2217 / #2219 / #2223): a typo
  silently selects one arm of a live cross-title A/B. It does announce which arm it took, which is
  the mitigation and the reason this is B rather than A.
* Weaker, same family: **`src/hle/graphics/hle_agc.cpp:2103-2104` `PROSPER_SUBMIT_STALL_US`** —
  `strtoul(e, nullptr, 0)`, end pointer discarded. `=-1` → `(unsigned)ULONG_MAX` = 4,294,967,295 µs
  ≈ **72 minutes of stall per submit**, which presents as a hang rather than as a mistyped variable.
  It prints the value it armed at, so it self-reports.

Minor consistency point visible only because C is unpublished: `PROSPER_GPU_CAPTURE_MAX_DRAWS`
(`gpu_capture.cpp:6290`) is in B while its twin on the line above, `PROSPER_GPU_CAPTURE_MIN_DRAWS`
(`:6289`) — same `strtoull(v, nullptr, 0)`, no end check, and `=-1` makes the capture never fire —
is in no list.

### NB3. The residual gate hole, demonstrated

M4 reproduces exactly as claimed: narrowing `CALL_RE` back to the two original entry points gives
exit 1 naming `PROSPER_COMPUTE_IMAGE_CACHE_MB`. Good regression guard, and `env_u64_or_report` is the
right shape for that site.

But M4 catches only *narrowing the pattern for a knob that already has an arm*. I added a brand-new
function to `live_compute.cpp` calling `parse_u64_strict` directly on a `PROSPER_TOTALLY_NEW_KNOB`
with no arm, and the gate printed `ok: 30 knob(s) parsed by env_numeric, every one armed` and exited
**0** — the original hole, one step out. Cheap close: also fail on any `parse_u64_strict(` /
`parse_u64_auto_base(` call outside `env_numeric.hpp` and the exempt test files. The comment at
`check_env_numeric_arms.py:32-36` states half of this ("adding a helper to that header without adding
it here re-opens exactly that hole"); calling the *parsers* directly is the other half.

### NB4. The octal divergence is deliberate, complete — and unarmed

I characterised it mechanically rather than by argument, over every 1-4 character string on the
alphabet `0-9 a-f x X + - space .`, comparing `parse_u64_auto_base` against what the old base-0 site
actually did (parse, end pointer discarded):

```
total value divergences vs the OLD site behaviour: 1130
  of which NOT a leading-zero decimal:                0
```

So the entire divergence set is `0` followed by a digit — in both directions. `=010` was a
`MAX_DISPATCH_GROUPS` cap of 8 and is now 10; `=0008` was 0 (**no cap**) and is now 8, because base 0
stopped at the invalid octal digit and the old code used the truncated answer. That is a good result
for the design — the deliberate choice is the *only* thing that moved — and the header's rationale at
`env_numeric.hpp:97-104` is sound. Nothing pins it, though: a later "restore base-0 fidelity" edit
would pass every arm. One row would fix that: `new("010") == 10 && old("010") == 8`.

### NB5. `env_numeric.hpp:97` says "Six sites"; there are seven

`PROSPER_VDEC2_DUMP_FRAMES` joined the `_auto` family with its D→A promotion. Call sites:
`mb3_freelist.cpp:83`, `:206`; `command_processor.cpp:1418`, `:2134`; `hle_service.cpp:494`, `:5300`;
`gpu_executor.cpp:7738`.

### NB6. The verification block cites a head that is not on the branch

`0edea136` is not an ancestor of `0b295aee` and its tree differs: the rebase from `b00a1c05` onto
`e05d71301` pulled in #3292, which also edited `live_compute.cpp`. The PR's **own** diff is
byte-identical across the rebase — `git diff b00a1c05 0edea136` and `git diff e05d7130 0b295aee`
share patch-id `37403bb0…`, which I checked — but the merged `live_compute.cpp` at the current head
is a file the quoted `355/355` never compiled. There are no conflict markers and the hunks are far
apart, so I expect it is fine; the point is only that CI at head, not the quoted local run, is the
evidence that binds here. Worth restating the head or noting the rebase.

---

## What I checked and cleared

**The `_auto` grammar — tested, not eyeballed.** I wrote a probe against `parse_u64_auto_base` with
hand-built boundary cases rather than sampling, and all pass: `0xFFFFFFFFFFFFFFFF` → `UINT64_MAX`,
`0x10000000000000000` → refused, `18446744073709551615` / `…616` likewise, and
`0x00000000000000000001` → 1 (leading zeros do not trip the overflow guard). Refused as intended:
`0x`, `0xzz`, `0x10 `, `0x10mb`, ` 0x10`, `-0x10`, `+1`, `0b101`, `0o17`, `8mb`, `1,000`. The
overflow test `value > (UINT64_MAX - digit) / 16u` is exact, not conservative — it rejects iff
`value*16 + digit > UINT64_MAX`. Reading `text[1]` after `*text` is checked is in-bounds (worst case
it reads the NUL), and the `||` short-circuits before it when `text[0] != '0'`.

**`env_u64_or_report`** leaves `*out` untouched on both the refusal and the unset path, and the call
site initialises `mib = 0` first (`live_compute.cpp:1105`). The `!value || !*value` pre-check at
`:1100` makes the helper's own empty branch unreachable there — harmless duplication, same answer.

**Falsified hypothesis, recorded because it was my strongest one.** I expected the three shader cache
caps — `PROSPER_SHADER_CACHE_MB` (`gpu_executor.cpp:1403`), `_DECODE_CACHE_MB` (`:1116`),
`_ANALYSIS_CACHE_MB` (`:1291`) — to be uncatalogued bucket A, since they are MiB byte caps of exactly
the converted shape. **False:** all three do `strtoull(value, &end, 10)` and then
`if (end == value || *end != '\0') return default_bytes;`. They belong to the 72 that check, not the
164. The census scoping is right on them, and on `PROSPER_GPU_CAPTURE_MAX_MB:1189`, the four
`PROSPER_GPU_TIMELINE_CAPTURE_*` selectors, `PROSPER_DET_FPS:137`, `PROSPER_FLIP_PACE_FPS:273`,
`PROSPER_AUDIO_CUSHION_GRAINS:134`, `PROSPER_DRAW_REALIZE_THREADS:9564`,
`PROSPER_DMEM_WRITE_TRACE_MAX_EVENTS:1431`, `PROSPER_STAGE_FOLD_PROFILE_CALLS:1050`,
`PROSPER_SYNC_RING:93`, `PROSPER_FORCE_LAYER:1888` and `PROSPER_CFG_TRIP_BOUND:1950`.

**Bucket A's default-ON family really is complete at five.** I scanned the whole tree for any
remaining numeric env read in the `!e || parse` shape and found none; every other guard is default
**OFF** (`e && strtol(e, nullptr, 0) != 0`), so a typo equals unset equals off —
`PROSPER_MB3_FREELIST_GUARD:1679`, `PROSPER_REL1_WAF_GUARD:1531`, `PROSPER_GENERATION_GUARD:1739`,
`PROSPER_NONHEAP_PTR_GUARD:1382`, `PROSPER_FOLD_MARGIN:532`. This was the sharpest way the
classification could still have been incomplete, and it is not.

**Tests and gates at this head.** Site test: **107 assertions, 0 failures**, clean under
`-Wall -Wextra`, and it still sets, unsets and reads no `PROSPER_*` variable. Gate:
`ok: 30 knob(s) parsed by env_numeric, every one armed (30 arm name(s) in the table)`.
`check_cached_env.py` also exits 0. M2 (arm deleted) and M3 (site reverted) reproduce with their
stated messages; M4 as described in NB3.

**The four new arms earn their place.** `PROSPER_VDEC2_DUMP_FRAMES` uses `0x40` → 64 as its `good`
column and `PROSPER_REL1_FORGE_GUARD` uses `0x0` → off, so both legs of the widened grammar are
pinned by a row that fails if it is narrowed — including the one that matters most, that `=0x0` still
means OFF on a default-ON guard. The `0xzz` malformed input is a good choice: glibc's base-0 `strtol`
parses the leading `0` and stops, so the legacy arm answers 0 and the discriminator genuinely fires.

**N5 was fixed, not just disclosed:** `mib_cap_size_new` now uses `SIZE_MAX`, `layout_entries_new`
uses `SIZE_MAX`, and the two pad holds have separate mirrors carrying `"flips"` and `"reads"`. I
re-checked every production `(fallback, cap, unit, note)` tuple against its mirror and they now match
exactly.

PR #2965, opened 2026-08-23
libSceHttp: hand out real library/template ids instead of the dispatcher's 0

mattias800:
Refs #2930.

## Failure scenario

`libSceHttp` registered exactly one function (`sceHttpUriParse`). Everything else fell to the dispatcher default `0` — and for `sceHttpInit` (`A9cVMUtEp4Y`) and `sceHttpCreateTemplate` (`0gYjPTR-6cY`) that 0 is **not** an error but a valid-looking id: the contracts return a positive library context / template id on success. Six of eight surveyed titles call both and carry the fake id into later calls (evidence table in the issue).

## Contract & approach

An id-returning contract must never answer 0. Both entry points now allocate from small tracked tables (ids ≥ 1, mutex-guarded), and `sceHttpDeleteTemplate` (`4I8vEpuEhZ8`) is registered as the explicit SCE_OK no-op it already was by accident.

**Deliberately not invented** (no live caller supplies evidence yet, so any constant would be fabrication): context-validation semantics, double-init behaviour, and SDK error encodings — the exhausted-table path answers a generic negative rather than a made-up `SCE_HTTP_ERROR_*`. PS5 3.20 libSceHttp exports no `sceHttpTerminate`, so contexts live for the process. The request-path honesty suggested in the issue is future work once a title actually reaches it; noted on the issue.

## Verification

- Red first: with only the test change built, `test_false_success_nids` fails at `sceHttpInit is registered`.
- After the fix: all 8 new HTTP arms pass (`== PASS ==`, exit 0), including sign checks that kill an error-shaped answer masquerading as an id.
- Full `cmake --build` + `ctest --no-tests=error` running on this exact head inside the toolchain container; results posted before merge.

## Risks

Titles that previously saw Init→0 may now branch differently on a real id; none of the six surveyed stops in HTTP today, and every later unregistered call keeps today's behaviour.

Issue #2930, opened 2026-08-22
libSceHttp registers only sceHttpUriParse — sceHttpInit and sceHttpCreateTemplate return a context id of 0 as false success (6 of 8 surveyed titles call them)


mattias800:
## Summary

`libSceHttp` (the v1 HTTP library) registers exactly **one** function —
`sceHttpUriParse` (`prosper/src/hle/net/hle_http.cpp:175`). Everything else in the library,
including the two entry points every caller starts with, falls to `prosper_on_unimpl` and
**returns 0**:

| NID | function | what 0 means to the guest |
| --- | --- | --- |
| `A9cVMUtEp4Y` | `sceHttpInit` | a **library context id of 0**, not an error |
| `0gYjPTR-6cY` | `sceHttpCreateTemplate` | a **template id of 0**, not an error |
| `5LZA+KPISVA` | `sceHttpUriBuild` | success, with the output buffer left untouched |

This is the FALSE SUCCESS class of #2081, and the same shape as **#2894** (`libSceHttp2`
unregistered, a request that was never sent reports `SCE_OK` and its response getters hand the
guest NULL) — one library version apart. #2894 is what holds *PGA TOUR 2K25* at rung 0.

`sceHttpInit`'s contract returns a **positive** library context id on success and a negative error
on failure. Returning 0 is neither: it is an id the guest will carry into every later call. The
file's own opening comment already states the principle — *"returning success without filling this
structure makes callers dereference stale pointer fields"* — but it is applied only to the URI
parser.

## Evidence

Runtime `[prosper] unimplemented:` lines, i.e. **calls actually made**, not import-table entries.
Frontend `tools/screenshot`, default route, `PROSPER_RENDER=1
PROSPER_GUEST_ARGS=-force-gfx-direct`, Linux/RADV, 2026-08-22:

| title | `sceHttpInit` | `sceHttpCreateTemplate` |
| --- | --- | --- |
| `PPSA02058` BALAN WONDERWORLD | yes | yes |
| `PPSA02101` Stray | yes | yes |
| `PPSA02154` Little Nightmares II | yes | yes |
| `PPSA03001` Sifu | yes | yes (+ `sceHttpUriBuild`) |
| `PPSA03130` Sniper Ghost Warrior Contracts 2 | yes | — |
| `PPSA03274` Unbound: Worlds Apart | yes | yes |

Six of the eight titles surveyed. `PPSA20800` Metaphor: ReFantazio calls the **v2** entry point
[...]

Issue #2894, opened 2026-08-21
libSceHttp2 is unregistered, so a request that was never sent reports SCE_OK and its response getters hand the guest NULL


mattias800:
## Summary

`libSceHttp2` is entirely unregistered, so the dispatcher answers every call with `0`. `0` is
`SCE_OK`. The result is that a request prosper never sent is reported as **sent successfully**, and
a getter that returns response headers reports success while writing nothing to its out-parameters.
The guest then parses the NULL buffer it was handed and dies.

Found while bringing up **PGA TOUR 2K25 (`PPSA17952`)**, where it is the blocker holding the title at
rung 0 once the PSN `module_start` handshake is fixed. It is not title-specific: any title that
reaches libSceHttp2 gets the same false success.

## The failure

```
[prosper] unimplemented: libSceHttp2::rbqZig38AT8  -> returning 0     (sceHttp2SendRequest)
[prosper] unimplemented: libSceHttp2::-rdXUi2XW90  -> returning 0     (sceHttp2GetAllResponseHeaders)
[prosper] WORKER-THREAD FAULT: sig=11 addr=(nil) rip=0x41142d5e0 (image+0x142d5e0)
[prosper]   insn bytes @rip: 41 0f b6 14 0b   (movzx edx,BYTE PTR [r11+rcx*1])   r11=0 rcx=0
[fault] thread='Background Job.'
```

`eboot+0x142d5e0` disassembles to an HTTP response-header scanner — it walks the buffer looking for
`0x09..0x0d`/`0x20` whitespace, then `:` (0x3a), then CR/LF — with a NULL base pointer:

```
142d5e0:  movzx edx,BYTE PTR [r11+rcx*1]    ; r11 = the header buffer = NULL
142d5e5:  lea   esi,[rdx-0x9]
142d5e8:  cmp   esi,0x5
142d5eb:  jb    0x142d5d4                   ; skip 0x09..0x0d
142d5ed:  cmp   edx,0x20
142d5f0:  je    0x142d5d4                   ; skip SPACE
142d603:  cmp   BYTE PTR [r11],0x3a         ; ':'
142d627:  movzx r9d,BYTE PTR [r10]
142d62e:  cmp   r9d,0xa                     ; LF
142d634:  cmp   r9d,0xd                     ; CR
```

`PROSPER_NULL_PAGE=1` confirms the cascade is short and local to this one parser: the read is
re-executed against a zero page and the *next* access faults `0x23` bytes later at
[...]

PR #3295, opened 2026-09-03
hle: answer sceHttpUriBuild and sceHttpTerm honestly, and validate the HTTP context id (#2930)

mattias800:
Closes the three false successes that #2965 left open on #2930, each answered from the shipped
PS5 3.20 `libSceHttp` rather than from a guessed encoding. Refs #2894, whose blocking question this
also settles — and, after review, records that falsification in the status doc the next agent will
actually read.

## The failure

`libSceHttp` still had entry points falling to `prosper_on_unimpl`, which returns `0`:

* **`sceHttpUriBuild`** reported `SCE_OK` with the caller's output buffer **untouched**, so the
  caller read whatever already happened to be in that buffer as a URI. *Sifu* is the surveyed caller.
* **`sceHttpTerm`** was unregistered, so a library context could be created but never released.
* **`sceHttpCreateTemplate`** accepted any context id at all, including ones nobody handed out.

## Contract, and where it came from

Not inferred from an error-name table — read off the library itself. `<DUMP_ROOT>/sprx/` carries
`libSceHttp.sprx` as a plain ELF (entropy 5.417 bits/byte; nothing decrypted anywhere in this work),
so its URI helpers and its id validator can be read directly.

`sceHttpUriBuild` (`5LZA+KPISVA`, `+0x21ac0`):

```
int32_t sceHttpUriBuild(char* out, size_t* required, size_t pool_size,
                        const SceHttpUriElement* element, uint32_t flags);
```

* `element == nullptr` -> `INVALID_URL`, checked **before** the output pointer is looked at.
* `out == nullptr && required == nullptr` -> `INVALID_VALUE`.
* `*required` is the assembled length **including the NUL**, and is stored **before** the
  buffer-size check — so a caller refused with `OUT_OF_MEMORY` still learns what to retry with.
* `out == nullptr` is a size query and returns `SCE_OK`.
* Components are gated by individual flag bits, emitted in the order
  `scheme ':' "//" user ':' pass '@' host ':' port path query fragment`, with the library's own
  per-component `strnlen` caps (`0x20` scheme, `0x100` userinfo, `0xff` host, `0x3fff` path/query/
  fragment).
* The default-port lookup (`+0x21950`) is a **prefix** compare over three literals — `HTTPS`/443,
  `HTTP`/80, and a string-merge artifact `TTP`/80 — reproduced rather than narrowed, because
  `element->scheme` is guest-supplied: `"httpx"` prefix-matches `HTTP` and suppresses `:80` on
  hardware.

`sceHttpTerm` is `Ik-KpLTlf7Q`. The comment this PR replaces asserted that "PS5 3.20 libSceHttp
exports no `sceHttpTerminate`" — the export exists, it is spelled `sceHttpTerm`, and *PGA TOUR 2K25*
imports it. It does not merely drop the context: after the refcount decrement at `+0x1068b` it runs
four per-context teardown helpers (`+0x83f0`, `+0x1c8e0`, `+0x35b0`, `+0xaf90`), so templates are
owned by their context here and reclaimed with it.

The context-id error is `0x80431100`, taken from the validator at **`+0xb070`** — the one
`sceHttpCreateTemplate` (`+0x107f0`) actually calls, not the same-shaped neighbour at `+0xaca0`.
CONFIDENCE: HIGH.

## Documentation (the blocking review finding)

`prosper/docs/PGA_TOUR_2K25_STATUS.md` still asserted that the libSceHttp2 error values "are not
derivable from the dump, from `../PS5-3.20_Libs/`, or from prosper's existing `0x8043xxxx`
constants". #2894 disproves that, `CLAUDE.md` routes the next PGA TOUR agent to that doc rather than
to an issue comment, and the claim was load-bearing in the wrong direction — it said the work could
not be started. Corrected, with two `## Ruled out` rows recording the facility (`0x817b____`, so the
v1 constants are **not** reusable), the exhaustive 33-target jump-table result, and the
`0x80410100 | BSD errno` encoding.

## Deliberately NOT done

The v1 request path (`CreateConnection*`, `CreateRequest*`, the send path) stays unregistered. The
reason is **scope, not hazard** — an earlier draft of this body said a constant here would be an
"error sentinel read as data", and this PR's own evidence retired that: every non-zero value
classifies identically, so a wrong constant there would be wrong, not dangerous. No surveyed v1
title reaches that path, and the encoding belongs to #2894.

One deliberate infidelity, commented in the source: the parser strips brackets from an IPv6 literal
and the builder does not re-add them, so `http://[::1]:8080/` does not round-trip byte-for-byte.
That is the shipped library's behaviour; it is **asserted** by a test rather than "fixed", because a
guest that re-parses the result depends on it. The `BEFORE_INIT`-vs-`INVALID_ID` distinction the
library draws and prosper collapses is likewise recorded in the comment as seen, not missed.

## Verification

Linux, container `ps5ys`, `-j4`, exact head `58632a28c`.

| | build | ctest |
| --- | --- | --- |
| `origin/main` (`e05d71301`) baseline | rc=0 | **353/353**, rc=0 |
| this branch, exact head | rc=0 | **353/353**, rc=0 |

`ctest --no-tests=error -R '^http_hle$'` -> 1/1 passed, rc=0. `diff --check` clean; the numbered-table
docs gate passes over every tracked `.md`.

`test_http` goes **12 -> 41 checks (+29)**. Every `CHECK` site now executes exactly once, so the
source count and the executed count are the same number and the artifacts cannot disagree.
**All 29 were run against mutated builds**; every mutation exits 1 on the checks that name it:

| mutation | exit | checks that failed |
| --- | --- | --- |
| unregister `sceHttpUriBuild` (the original defect) | 1 | registered-by-NID |
| store `*required` **after** the pool-size check | 1 | required-size on the OOM path |
| always emit the port | 1 | size query; round-trip; prefix-scheme suppression |
| drop the context-id validation | 1 | rejects ctx 0; out-of-range; terminated ctx; template reclaim |
| `sceHttpTerm` as a no-op success | 1 | rejects ctx 0; out-of-range; terminated ctx; template reclaim |
| never emit `"//"` | 1 | round-trip; flag gating; assembly order; both port arms |
| **M7** Term frees only the context slot (templates leak) | 1 | template reclaim + owner discrimination |
| **M8** exact-match default port instead of the prefix compare | 1 | both prefix-scheme arms |
| **M9** drop the `"ttp"` prefix arm | 1 | the `"TTP"` arm — *and only that one* |
| **M10** Term clears **every** context's templates, not just the terminated one's | 1 | owner discrimination — *and only that one* |

M7/M8 close the review's N1/N2; M9/M10 close the two coverage gaps found on re-review. The
template-reclaim check replaces one that ran 16 init/term cycles creating **no templates** — a leak
was structurally inexpressible to it, the same vacuity class as #3288. M10 is the reason that fix
needed a second pass: the replacement proved templates *die*, not that they die *selectively*, so a
Term weakened to `if (tmpl.in_use) tmpl = {}` passed everything. The owner field's whole purpose was
untested until the two-context arm existed, and M9/M10 each firing exactly one check is the proof
that neither new arm passes for an unrelated reason.

No snapshots: this touches no renderer path, and the machine's GPU is held by another lane.
No screenshot and no `BLOG.md` entry — this moves no title's rung and produces no image. The finding
worth reading is the #2894 analysis.

Refs #2930
Refs #2894

### DIFF
--- a/prosper/docs/PGA_TOUR_2K25_STATUS.md
+++ b/prosper/docs/PGA_TOUR_2K25_STATUS.md
@@ -118,11 +118,31 @@ dereference stale pointer fields."*
 
 The fix is to implement libSceHttp2's lifecycle honestly — local resource creation genuinely
 succeeds, setters genuinely record state, and anything that would need a network answer **reports
-failure instead of `SCE_OK`**, leaving out-parameters untouched. That was deliberately **not** done
-in this session: the correct Sony error values for the libSceHttp2 facility are not derivable from
-the dump, from `../PS5-3.20_Libs/` (which gives names and NIDs only, no values), or from prosper's
-existing `0x8043xxxx` libSceHttp constants, and inventing one would be fabrication rather than
-reimplementation. Tracked separately.
+failure instead of `SCE_OK`**, leaving out-parameters untouched. Tracked as #2894.
+
+**The reason this was deferred no longer holds — see `## Ruled out`.** This paragraph used to say
+the v2 error encodings were not derivable from anything on hand. They are, and the answer changes
+what an implementer must do: **libSceHttp2's facility is `0x817b____`, not v1's `0x8043____`**, so
+prosper's existing libSceHttp constants are *not* reusable across the two even though the low code
+bytes are shared. `<DUMP_ROOT>/sprx/` ships `libSceHttp.sprx` and `libSceHttp2.sprx` as plain ELF
+(entropy 5.417 / 5.403 bits/byte), and this title's own error classifier at `eboot+0x142ddf0`
+confirms the facility independently.
+
+The behavioural finding matters more than the constant. **Any non-zero return is handled
+gracefully; only `0` crashes this title.** The classifier's sole zero-returning path is a literal
+`test eax,eax`, all 33 targets of its `0x817b1064..0x817b1084` jump table return non-zero, and at
+the send site a non-zero classification makes the caller store the error and *return* without ever
+reaching the response parser above. So the fault at `eboot+0x142d5e0` is caused specifically by the
+false `SCE_OK` — the title was always prepared to be told the request failed. Picking the exact
+constant is a correctness question, not a safety one.
+
+For the send path specifically, the real library does not return an HTTP error at all: on a connect
+failure it propagates the **raw libSceNet error**, pinned by `0x80410124` being special-cased as
+*not* a failure at libSceHttp2's `sceNetConnect` site (`0x24` = 36 = `EINPROGRESS`), which fixes the
+encoding as `0x80410100 | BSD errno`. That makes `ENETUNREACH` the honest offline answer, and it is
+consistent with what prosper already tells this guest through `sceNetCtlGetState`
+(`SCE_NET_CTL_ERROR_NOT_CONNECTED`). Full derivation, per-NID export map and argument shapes are on
+#2894; start there rather than re-deriving.
 
 Other unimplemented NIDs seen on the same boot, none of which is implicated in the fault:
 `sceNpAuthCreateAsyncRequest`, `sceNpAuthGetAuthorizationCodeV3`, `sceNpAuthWaitAsync`,
@@ -150,6 +170,26 @@ HTTP fault, so no hypothesis is recorded here yet.
 
 ## Ruled out
 
+- **"The correct Sony error values for the libSceHttp2 facility are not derivable from the dump,
+  from `../PS5-3.20_Libs/`, or from prosper's existing `0x8043xxxx` constants, so implementing the
+  lifecycle honestly would require fabricating one."** Falsified 2026-09-03 (#2894, PR #3295 for
+  the v1 sibling). They are derivable, from two independent primary sources that agree. The dump's
+  own `sprx/` directory ships `libSceHttp2.sprx` as a **plain ELF** — entropy 5.403 bits/byte, so
+  nothing is decrypted to read it — and every export's shared stack-guard prologue returns
+  `0x817b1076`, dating the facility as **`0x817b____`**, *not* v1's `0x8043____`. This title's own
+  error classifier at `eboot+0x142ddf0` compares against `0x817b1220` and dispatches
+  `0x817b1064..0x817b1084` through a 33-entry jump table, confirming the facility from the guest
+  side. The claim was not merely incomplete but load-bearing in the wrong direction: it said the
+  work could not be started, when in fact the send path does not use an HTTP error at all — the
+  library propagates the raw libSceNet error, whose encoding (`0x80410100 | BSD errno`) is pinned by
+  `0x80410124` = `EINPROGRESS` being treated as a non-failure at the connect site. Independently
+  re-derived from the bytes by a second reader before this row was written.
+- **"Any error prosper invents here risks being read as data, so returning one is hazardous."**
+  Falsified 2026-09-03 by enumerating all 33 jump-table targets: they return five distinct non-zero
+  categories and **none** reaches the classifier's zero-return `ret`, which is reachable only for an
+  input of exactly `0`. So every non-zero value is classified as a failure and handled the same way,
+  and the caller returns before touching the response parser. A wrong constant here is wrong, not
+  dangerous — which removes the safety argument for deferring the whole lifecycle.
 - **"The `only_if_imported` filter (#2870/#2890) drops `sce_module/libSceNpCppWebApi.prx` for this
   title, and that is what makes the PSN handshake fail."** Falsified 2026-08-22 on `af481db2`:
   `PROSPER_INITLOG=1` lists it as **module 8, base `0x4d0000000`** — it is linked, mapped and
--- a/prosper/docs/PGA_TOUR_2K25_STATUS.md
+++ b/prosper/docs/PGA_TOUR_2K25_STATUS.md
@@ -140,8 +140,9 @@ For the send path specifically, the real library does not return an HTTP error a
 failure it propagates the **raw libSceNet error**, pinned by `0x80410124` being special-cased as
 *not* a failure at libSceHttp2's `sceNetConnect` site (`0x24` = 36 = `EINPROGRESS`), which fixes the
 encoding as `0x80410100 | BSD errno`. That makes `ENETUNREACH` the honest offline answer, and it is
-consistent with what prosper already tells this guest through `sceNetCtlGetState`
-(`SCE_NET_CTL_ERROR_NOT_CONNECTED`). Full derivation, per-NID export map and argument shapes are on
+consistent with what prosper already tells this guest through NetCtl: `sceNetCtlGetState` writes
+`SCE_NET_CTL_STATE_DISCONNECTED` and `sceNetCtlGetInfo` returns `SCE_NET_CTL_ERROR_NOT_CONNECTED`
+(`hle_service.cpp:4795`, `:4830`). Full derivation, per-NID export map and argument shapes are on
 #2894; start there rather than re-deriving.
 
 Other unimplemented NIDs seen on the same boot, none of which is implicated in the fault:
--- a/prosper/src/hle/net/AGENTS.md
+++ b/prosper/src/hle/net/AGENTS.md
@@ -1,0 +1,51 @@
+# `src/hle/net` — Sony networking libraries
+
+Reimplementations of the PS5 networking libraries that prosper answers itself. Today that is
+**libSceHttp** (`hle_http.cpp`): the URI helper family and the library/template id lifecycle.
+
+## The policy this folder exists to hold
+
+**prosper has no network, and acquiring one is not the plan.** That makes the interesting question
+per entry point *"is this answerable from local state?"* rather than *"can we connect?"* — and the
+two halves of the answer are both obligations:
+
+- **Anything computable offline is implemented, not stubbed.** The URI helpers are ordinary string
+  code; there is no network anywhere in parsing or rebuilding a URI, so leaving them to the
+  dispatcher is a false success rather than a limitation. Same for id lifecycles: a template really
+  is allocated locally, so it gets a real id from a real table.
+- **Anything that genuinely needs a network answer must FAIL, and fail loudly.** The dispatcher's
+  unregistered default is `0`, which is `SCE_OK` — so an unregistered send path reports that a
+  request nobody sent succeeded, and a response getter reports success while writing nothing to its
+  out-parameters. That is the specific shape that crashes callers (#2894), and it is worse than a
+  reported error, because the guest's own error handling never runs.
+
+Registering a NID can therefore be worse than leaving it unregistered, and the deciding question is
+what the guest does with the *value*. An **id**-returning contract must never answer `0` — that is a
+plausible-looking handle the guest will carry into later calls — and must never answer a
+zero-extended error either; sign-extend so `int32` and `int64` reads both see it as negative. An
+**SCE_OK-or-error** contract keeps the 32-bit form the library returns in `eax`. Both conventions
+live in `hle_http.cpp` with comments saying which is which.
+
+## Where the evidence comes from
+
+Contracts here are read off the shipped modules rather than guessed. `<DUMP_ROOT>/sprx/` carries
+`libSceHttp.sprx` and `libSceHttp2.sprx` as **plain ELF** — nothing is decrypted to read them — so
+argument shapes, per-component length caps, emission order and error constants can be derived
+directly, and a claim in this folder should cite the offset it came from. Two facts worth not
+re-deriving: **libSceHttp2's error facility is `0x817b____`, not v1's `0x8043____`** (the low code
+bytes are shared, so the v1 constants are not reusable across the two), and on a connect failure the
+library propagates the **raw libSceNet error**, encoded `0x80410100 | BSD errno`. See #2894.
+
+## The boundary against its siblings
+
+This folder is smaller than "prosper's networking" — several networking surfaces are answered
+elsewhere, and looking for them here is the mistake to avoid:
+
+- **`sceHttp2Init`, `sceNetCtlGetState` and the NetCtl/NP service surface live in
+  `../service/hle_service.cpp`**, not here. That is why `libSceHttp2` looks absent from this folder
+  while one of its entry points is already registered.
+- Answers must not contradict each other across that boundary. `sceNetCtlGetState` already reports
+  `SCE_NET_CTL_ERROR_NOT_CONNECTED`, so an HTTP layer that reported a successful request would be
+  telling the same guest two incompatible things.
+
+New Sony networking libraries that prosper answers itself belong here, one file per library.
--- a/prosper/src/hle/net/AGENTS.md
+++ b/prosper/src/hle/net/AGENTS.md
@@ -44,8 +44,10 @@ elsewhere, and looking for them here is the mistake to avoid:
 - **`sceHttp2Init`, `sceNetCtlGetState` and the NetCtl/NP service surface live in
   `../service/hle_service.cpp`**, not here. That is why `libSceHttp2` looks absent from this folder
   while one of its entry points is already registered.
-- Answers must not contradict each other across that boundary. `sceNetCtlGetState` already reports
-  `SCE_NET_CTL_ERROR_NOT_CONNECTED`, so an HTTP layer that reported a successful request would be
-  telling the same guest two incompatible things.
+- Answers must not contradict each other across that boundary. prosper already tells the guest it
+  is offline there — `sceNetCtlGetState` writes `SCE_NET_CTL_STATE_DISCONNECTED` and
+  `sceNetCtlGetInfo` returns `SCE_NET_CTL_ERROR_NOT_CONNECTED` (`hle_service.cpp:4795`, `:4830`) —
+  so an HTTP layer that reported a successful request would be telling the same guest two
+  incompatible things.
 
 New Sony networking libraries that prosper answers itself belong here, one file per library.
--- a/prosper/src/hle/net/hle_http.cpp
+++ b/prosper/src/hle/net/hle_http.cpp
@@ -204,16 +204,29 @@ constexpr size_t kUriTextMax = 0x3fff;
 
 size_t capped_len(const char* text, size_t cap) { return text ? strnlen(text, cap) : 0; }
 
-// The library upper-cases the scheme and matches "HTTPS" then "HTTP"; anything else has no
-// default port. It then compares a third literal that string-merging left as "TTP", which no
-// real scheme can reach -- deliberately not reproduced. Matching is exact here rather than the
-// library's prefix compare, which is unreachable for a scheme that parsed.
+// The library upper-cases the scheme into a stack buffer and then runs three PREFIX compares --
+// strncmp(upper, LIT, strlen(LIT)) -- against "HTTPS" (443, +0x21aa2), "HTTP" (80, +0x21aab) and
+// a third literal that string merging left as "TTP" (+0x326ac). That third arm ends in
+// `mov eax,0x50 / cmovne eax,ecx` with ecx zeroed, and cmovne fires on MISMATCH: a "TTP" prefix
+// yields 80, and only a non-match yields "no default port".
+//
+// Reproduced rather than narrowed to exact equality, which is what this comment used to claim was
+// safe. element->scheme is GUEST-supplied and need not have come from our own parser, so "httpx"
+// prefix-matches "HTTP" and takes 80 on hardware; under exact matching prosper would answer 0 and
+// then emit a ":80" that the library suppresses. Divergence for a scheme nobody uses, but the
+// contract this file claims is fidelity to the shipped library, so match it.
 uint16_t default_port_for_scheme(const char* scheme) {
     const size_t length = capped_len(scheme, kUriSchemeMax);
     if (length == 0) return 0;
     const std::string_view value(scheme, length);
-    if (ascii_ieq(value, "https")) return 443;
-    if (ascii_ieq(value, "http")) return 80;
+    // strncmp against a shorter buffer stops at its NUL, so a prefix match needs the whole literal.
+    const auto prefixed = [value](std::string_view literal) {
+        return value.size() >= literal.size() &&
+               ascii_ieq(value.substr(0, literal.size()), literal);
+    };
+    if (prefixed("https")) return 443;
+    if (prefixed("http")) return 80;
+    if (prefixed("ttp")) return 80;
     return 0;
 }
 
@@ -328,9 +341,15 @@ HLE(h_http_uri_build) { // (out, required, pool_size, element, flags)
 // census noise without fabricating an SDK error encoding.
 //
 // CreateTemplate now VALIDATES its library context id rather than accepting whatever it is
-// given. The encoding is not invented: the shipped PS5 3.20 libSceHttp contains the id
-// validator itself, which range-checks (1..0x80) and cross-checks the table slot and answers
-// 0x80431100 on either failure. CONFIDENCE: HIGH.
+// given. The encoding is not invented: sceHttpCreateTemplate (+0x107f0) calls the validator at
+// +0xb070, which range-checks the id (1..0x80), cross-checks the slot in the 0xd0-stride context
+// table, and answers 0x80431100 on either failure. That is the validator on this call path, not
+// a same-shaped neighbour. CONFIDENCE: HIGH.
+//
+// Known fidelity gap, seen and not missed: the library answers 0x80431001 (BEFORE_INIT) when
+// sceHttpInit was never called and only reaches the id validator once the library is up, while
+// prosper answers 0x80431100 for both. Derivable locally -- no live context means never inited --
+// but the guest classifies the two identically, so it is not worth a second code path yet.
 //
 // This comment used to say "PS5 3.20 libSceHttp exports no sceHttpTerminate: contexts live for
 // the process". The name is wrong, and so was the conclusion drawn from it - the export is
@@ -339,7 +358,7 @@ HLE(h_http_uri_build) { // (out, required, pool_size, element, flags)
 // dry by a title that inits and terminates repeatedly.
 
 struct HttpLibCtx   { bool in_use = false; };
-struct HttpTemplate { bool in_use = false; };
+struct HttpTemplate { bool in_use = false; int owner = 0; };
 constexpr int kMaxHttpLibs      = 4;
 constexpr int kMaxHttpTemplates = 16;
 
@@ -366,10 +385,19 @@ bool http_ctx_is_live(uint64_t raw) {
     return id >= 1 && id <= kMaxHttpLibs && g_http_libs[id - 1].in_use;
 }
 
-HLE(h_http_term) { // sceHttpTerm(libCtxId) -> SCE_OK, releasing the context
+HLE(h_http_term) { // sceHttpTerm(libCtxId) -> SCE_OK, releasing the context AND what it owns
     std::lock_guard<std::mutex> lk(g_http_mx);
     if (!http_ctx_is_live(a0)) return http::kErrorInvalidId;
-    g_http_libs[(int32_t)a0 - 1].in_use = false;
+    const int id = (int32_t)a0;
+    // Term does not merely drop the context. After the refcount decrement at +0x1068b the shipped
+    // sceHttpTerm runs four per-context teardown helpers (+0x83f0, +0x1c8e0, +0x35b0, +0xaf90),
+    // each taking the ctx id in edi; +0xaf90 re-validates the id against the same slot table the
+    // create path uses. Releasing only the slot would leak every template created under it, and
+    // the template table is 16 deep -- so a title that inits, creates one template and terms in a
+    // loop would exhaust it and start getting -1 from a table that is really empty.
+    for (auto& tmpl : g_http_templates)
+        if (tmpl.in_use && tmpl.owner == id) tmpl = {};
+    g_http_libs[id - 1].in_use = false;
     return 0;
 }
 
@@ -380,6 +408,7 @@ HLE(h_http_create_template) { // sceHttpCreateTemplate(libCtxId, ...) -> templat
     for (int i = 0; i < kMaxHttpTemplates; i++) {
         if (g_http_templates[i].in_use) continue;
         g_http_templates[i].in_use = true;
+        g_http_templates[i].owner = (int32_t)a0;  // so sceHttpTerm can reclaim it
         return (uint64_t)(i + 1);
     }
     return (uint64_t)(int64_t)-1;
@@ -388,7 +417,7 @@ HLE(h_http_create_template) { // sceHttpCreateTemplate(libCtxId, ...) -> templat
 HLE(h_http_delete_template) { // sceHttpDeleteTemplate(templateId) -> SCE_OK offline
     std::lock_guard<std::mutex> lk(g_http_mx);
     if (int id = (int32_t)a0; id >= 1 && id <= kMaxHttpTemplates)
-        g_http_templates[id - 1].in_use = false;
+        g_http_templates[id - 1] = {};
     return 0;
 }
 } // namespace
--- a/prosper/src/hle/net/hle_http.cpp
+++ b/prosper/src/hle/net/hle_http.cpp
@@ -9,6 +9,7 @@
 #include <algorithm>
 #include <cctype>
 #include <cstdint>
+#include <cstdio>
 #include <cstring>
 #include <mutex>
 #include <string>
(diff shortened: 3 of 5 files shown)


mattias800 (commented):
## Independent review — REJECTED, on one docs finding. The code is sound and I approve it as written.

I re-derived the primary evidence myself rather than reading the prose: the two `.sprx` modules, the
PGA TOUR eboot, and `../PS5-3.20_Libs/`. Everything load-bearing in this PR and in the #2894 analysis
checked out, and two of the three headline claims are stronger than stated. The single blocking item
is a documentation row the charter requires before merge, not a defect in the change.

---

## Blocking

### B1 — `prosper/docs/PGA_TOUR_2K25_STATUS.md` still carries the claim this PR falsifies

That file's "The current blocker" section says, on `origin/main` today:

> the correct Sony error values for the libSceHttp2 facility are not derivable from the dump, from
> `../PS5-3.20_Libs/` (which gives names and NIDs only, no values), or from prosper's existing
> `0x8043xxxx` libSceHttp constants, and inventing one would be fabrication rather than
> reimplementation.

Your #2894 comment disproves that, and I independently confirmed the disproof below. The doc already
has a `## Ruled out` section (line 151) and this PR touches no docs, so the falsification currently
lives only in an issue comment. The charter is explicit:

> A falsification that lives only in a PR body or an issue comment is one the next agent will
> re-derive at full cost.

The concrete failure: `CLAUDE.md`'s routing table sends a PGA TOUR agent to
`PGA_TOUR_2K25_STATUS.md`, not to the fourth comment on #2894. They read "not derivable", and stop —
and the doc is now *wrong* rather than merely incomplete, which is worse, because a wrong statement
in a `## Ruled out`-bearing status doc reads as checked. This cost you a session to derive; it should
cost the next reader a link.

What is needed is small: one `## Ruled out` row ("the v2 error encodings are not derivable from
anything on hand" — falsified 2026-09-03, facility `0x817b`, evidence and issue link), and a
correction to the "The current blocker" paragraph that quotes it. Land it here or in a same-day
docs-only PR ahead of this one; per the charter's docs exception that one needs no CI wait. Ping me
and I will re-approve on sight — nothing else is outstanding.

---

## What I verified independently, and where the claims land

### Claim 1 — the `0x817b` facility. VERIFIED from the bytes. CONFIDENCE: HIGH.

Not taken on your word at any step.

* `<DUMP_ROOT>/sprx/libSceHttp.sprx` and `libSceHttp2.sprx` are plain ELF (`e_ident` `\x7fELF`,
  `e_type` `0xfe18`), Shannon entropy **5.417** and **5.403** bits/byte — your figures to three
  decimals. Nothing decrypted, confirmed rather than asserted.
* Immediate scan reproduces your frequency table exactly. v1: `0x8043` dominant at 641 occurrences,
  `0x80431001` x114, `0x804311fe` x108, `0x80431022` x101, `0x80431076` x99, `0x80431100` x48. v2:
  `0x817b` dominant at 365, `0x817b1022` x70, `0x817b1076` x55, `0x817b11fe` x35, `0x817b1100` x35,
  `0x817b1104` x19. Shared low codes, different facility word. The v1 constants are not reusable.
* The stack-guard prologue you used to date the facility: `libSceHttp2+0x10b0` is literally
  `mov ebx,0x817b1076` / `cmp rax,0x1fd0` / `jb`. Your `0x1fd0` is exact. (v1's sibling uses
  `0x3fd0` — noting only so nobody "corrects" one to the other.)

**`0x80431100` is better sourced than the PR body says.** The body and the #2930 comment cite the
validator at `+0xaca0`. That one exists and does what you describe, but it is not the one on the
path: `sceHttpCreateTemplate` (`0gYjPTR-6cY`, `+0x107f0`) reaches `+0x108e2: call 0xb070`, and
`+0xb070` is a near-identical sibling —

```
b07b:  movzx eax,dil
b07f:  mov   ebx,0x80431100
b084:  dec   eax
b086:  cmp   eax,0x7f
b089:  ja    0xb0c9              ; -> return 0x80431100
b08d:  lea   r15,[rip+0x30f8c]   ; slot table 0x3c020, stride 0xd0
b09b:  cmp   DWORD PTR [r14+r15*1],edi
b09f:  jne   0xb0c9              ; -> return 0x80431100
```

So the constant is read off the exact call path this PR reimplements, not off a same-shaped
neighbour. Consider citing `+0xb070` in the comment instead; it makes the strongest claim in the
change also the easiest to re-check.

### Claim 2 — `0x80410124` = EINPROGRESS. VERIFIED, and it is not single-anchor.

The connect site is exactly as quoted, and I resolved the callee rather than assuming it:
`libSceHttp2+0x1a9a1: call 0x21cd0`, a PLT thunk through GOT `0x2c7f8`, which the module's own
`DT_JMPREL` binds to dynamic symbol `OXXX4mUk3uk` — named `sceNetConnect` at
`../PS5-3.20_Libs/libSceNet.c:1569`. Then `mov ebx,eax` / `test eax,eax` / `jns` /
`cmp ebx,0x80410124` / `je`, i.e. that one value alone is not a failure.

Your #2894 comment presents this as one matching constant and marks CONFIDENCE: MED on the errno.
**There are at least two more supports, and I think the marking is conservative rather than
generous:**

1. `libSceNet.sprx` itself carries 37 distinct `0x804101xx` values, and the low byte reads as
   FreeBSD errno throughout the socket range: `04` EINTR, `09` EBADF, `16` EINVAL(22), `24`
   EINPROGRESS(36), `31` EADDRNOTAVAIL(49), `38` EISCONN(56), `3d` ECONNREFUSED(61), with a
   Sony-specific block from `a4`/`c8` upward. A different encoding would not produce that spread.
2. prosper already encoded the same scheme before this analysis existed:
   `src/hle/service/hle_service.cpp:2824` returns `0x80410118` labelled `SCE_NET_ERROR_ENFILE`.

Unrelated pre-existing nit while I was there: `0x18` is 24, which is `EMFILE` on FreeBSD; `ENFILE`
is 23. Not this PR's line, not worth a PR of its own, but if someone derives a constant from that
comment they will be one off.

### Claim 3 — "only `0` crashes it". VERIFIED, and I closed the gap your analysis left open.

The classifier is at eboot vaddr `0x142ddf0` exactly:

```
142ddf2:  cmp  edi,0x817b1220
142de10:  lea  ecx,[rax+0x7e84ef9c]   ; = rax - 0x817b1064
142de16:  cmp  ecx,0x20               ; 33 entries
142de1f:  lea  rax,[rip+0xa36256]     ; 0x1e6407c
142de26:  movsxd rcx,DWORD PTR [rax+rcx*4]
142de2d:  jmp  rcx
```

`0x817b1220`, the `0x817b1064..0x817b1084` range and the 33-entry table at `0x1e6407c` are all as
described.

Your claim is that the sole zero-returning path is the literal `test eax,eax`. That is true of the
straight-line code — but the 33 jump-table targets are unexamined in the comment, and a single
target returning 0 would have falsified the safety argument. **I resolved all 33.** They land on
five distinct returns: `0x1e`, `0xe`, `0x11`, `0x02` (29 of them) and `0x06`. None is
`0x142defe`, the `ret` that leaves `eax` at 0, which is reachable only from
`142defa: test eax,eax / jne 0x142df18` — i.e. only for an input of exactly zero.

The send site matches too (`142d08c: call` the import, `mov edi,eax`, `call 0x142ddf0`, then the two
`je 0x142d17a`). So the design fact holds without qualification, and it is the right one to build on.

**One consequence for the PR body.** It says the request path stays unregistered because "a guess
made here would be exactly the 'error sentinel read as data' hazard the charter warns about." Your
own evidence has since retired that reason: any non-zero classifies identically, so no constant in
that space is dangerous, only wrong. The honest reason to defer is scope — no surveyed v1 title
reaches the v1 request path, and the encoding belongs to #2894. Worth fixing in the record, because
the stale framing is exactly what would stop the next agent from picking it up.

### Claim 5 — `sceHttpTerm`. VERIFIED. The old comment was wrong and the correction is right.

`../PS5-3.20_Libs/libSceHttp.c:909` binds `Ik-KpLTlf7Q` to `sceHttpTerm`; the string
`sceHttpTerminate` does not occur anywhere in that file. The export is real, not a name in a table:
`libSceHttp+0x10550`, `st_size` `0x17f`, with a body that takes the ctx id in `edi`. PGA TOUR's
eboot contains the literal `Ik-KpLTlf7Q`, so it does import it. This is the kind of
corrected-upward claim the charter says gets checked least; it survives checking.

### The `sceHttpUriBuild` contract — VERIFIED instruction by instruction.

`5LZA+KPISVA` resolves to `+0x21ac0`, `st_size` `0x4ea`, from the module's own dynsym. Every bullet
in the comment is in the bytes:

* `test rcx,rcx` / `je -> mov eax,0x80433060` at `+0x21aeb`/`+0x21b38`, and the output pointer is
  first touched at `+0x21af9`. Null element really is rejected first.
* `mov eax,0x804311fe` then `test rbx,rbx` / `je return` — INVALID_VALUE only when both are null.
* `inc r13` (the NUL) then `mov QWORD PTR [rax],r13` at `+0x21d63`, and only then
  `cmp QWORD PTR [rbp-0x48],0` / `mov eax,0x80431022` / `cmp r13,pool_size` / `ja` at
  `+0x21d66..0x21d7d`. The required size is stored before the pool check.
* `out == nullptr` falls to `+0x21f86: xor eax,eax`.
* Caps `0x20` scheme, `0x100` user, `0x100` pass, `0xff` host, `0x3fff` path/query/fragment, at
  field offsets 8/0x10/0x18/0x20/0x28/0x30/0x38 with the port at 0x40 — matching
  `SceHttpUriElement` and `hle_http.hpp`'s new flag bits one for one (0x01 scheme, 0x02 host,
  0x04 port, 0x08 path, 0x10 user, 0x20 pass, 0x40 query, 0x80 fragment — the fragment bit read as
  the sign bit at `+0x21d2e: test r12b,r12b / js`).
* Emission order scheme `:` `//` user `:` pass `@` host port path query fragment NUL, at
  `+0x21d8d`, `+0x21df3`, `+0x21e08`, `+0x21e6a`, `+0x21e8f`, `+0x21e9d`, `+0x21eee`, `+0x21f0f`,
  `+0x21f38`, `+0x21f5d`, `+0x21f81`.
* The port literal at `0x32d1c` is `":%d"` into a 7-byte buffer, and `mailto` is at `0x32d86`.

**`element_is_opaque` reading the leading dword is correct, and it is the decision I most expected
to be wrong.** Reading four bytes where the struct declares `bool opaque; uint8_t align[7]` looks
like a padding hazard. It is not: `+0x21b46` and `+0x21ddc` are both `cmp DWORD PTR [r14],0x1`. You
are faithful to the library and the comment says why. Good call.

**Memory safety: no overflow path.** I re-derived `needed` against the bytes actually written for all
four userinfo combinations (neither / user only / pass only / both) — user-only reserves `len+1` and
writes `len` plus `@`; pass-only reserves `len+2` and writes `:` plus `len` plus `@`; both reserve
`user+pass+2` and write the same. Scheme, `//`, host, port, path, query, fragment and the NUL all
match. The pool check precedes the first store, every `append` is bounded by a `strnlen`-capped
length, and `":%d"` of a `uint16_t` maxes at 6 chars inside `char port_text[8]`.

### Other things I checked and cleared

* **Return conventions.** Id-returning paths sign-extend so the guest sees negative as `int32` and
  as `int64`; SCE_OK-or-error paths keep the 32-bit form the library returns in `eax`. Matches the
  library and the rest of the file. The comment explaining the asymmetry is worth having.
* **NID collisions.** All six NIDs registered here occur in `libSceHttp.c` and in no other of the
  275 libraries, so global registration cannot shadow another library's export.
  `Hle::register_fn` records shadows anyway (`dispatch.cpp:107`).
* **`tests/hle/test_false_success_nids.cpp:288`** — `test_http_ids()` passes a live ctx from
  `init_fn` into `create_fn`, so the new validation cannot break it, and its exhaustion probe is
  unaffected.
* **Clean-room integrity: clear.** No Sony code, comments, data tables or strings are transplanted.
  The only literal reproductions are ABI-visible values — error constants, the `":%d"` port format,
  the component caps — which are the interface, not the implementation. The behaviour is rewritten
  in prosper's own idiom. The strongest evidence of re-derivation rather than transcription is that
  you *declined* to reproduce the `"TTP"` quirk (see N1) rather than copying it.
* **Third-party `*_emu` libraries: clear.** Nothing here loads or links one. The PGA TOUR dump does
  ship an `ampr_emu.index`, but it is inert data and `module_path_policy.hpp` still rejects by
  default. No change to module discovery.
* **Is registering safer than not, per entry point?** Yes in all three cases, and I checked the
  direction rather than assuming it. `UriBuild`: was SCE_OK with the buffer untouched, now a real
  string or a real error. `Term`: was SCE_OK with the slot permanently leaked. `CreateTemplate`
  validation: the only newly-rejected input is a context prosper never handed out, which the shipped
  library also rejects — and reachability of the v1 request path is unchanged from #2965, so nothing
  here moves a title *closer* to the false-success cliff.
* **Mutation arms.** I read all 20 new checks looking for one that could pass under the dispatcher's
  default 0 or for the wrong reason. None can: every content assertion compares against a string the
  untouched buffer cannot contain, and every error assertion expects a value the default cannot
  produce. The one arm I would not lean on is `CHECK(true, "init/term cycles do not leak context
  slots")` inside the loop — it is self-satisfying and only bites via the `CHECK(false)` break path.
  It works, but see N2 for what it cannot see.
* **CI**: 9 green, 1 pending (`macOS x86_64 / Rosetta 2`) at the time of writing. `diff --check`
  clean. No other open PR touches `hle/net/` or `test_http.cpp`.

---

## Non-blocking findings

**N1 — a source comment states something the bytes contradict.** In `hle_http.cpp`, above
`default_port_for_scheme`:

> The library upper-cases the scheme and matches "HTTPS" then "HTTP"; anything else has no default
> port. It then compares a third literal that string-merging left as "TTP", which no real scheme can
> reach [...] Matching is exact here rather than the library's prefix compare, which is unreachable
> for a scheme that parsed.

The helper is `libSceHttp+0x21950`. It upper-cases into a stack buffer (that half is right), then
does three **prefix** compares of `strncmp(upper, LIT, strlen(LIT))`:

```
21a3d:  test eax,eax / je 0x21aa2   ; "HTTPS" (0x326a6) -> mov ax,0x1bb   (443)
21a67:  test eax,eax / je 0x21aab   ; "HTTP"  (0x32c30) -> mov ax,0x50    (80)
21a91:  xor ecx,ecx
21a93:  test eax,eax                ; "TTP"   (0x326ac)
21a95:  mov  eax,0x50
21a9a:  cmovne eax,ecx              ; MATCH -> 80 ; mismatch -> 0
```

Two corrections. The `"TTP"` arm returns **80**, not "no default port" — the `cmovne` fires on
mismatch, not on match. And the prefix compare is reachable: `element->scheme` is guest-supplied
and need not have come from prosper's parser, so `"httpx"` prefix-matches `"HTTP"` and gets 80 on
hardware where prosper gets 0. Consequence is one emitted `:80` that hardware suppresses — harmless
for every real scheme, which is why I am not blocking. But the comment asserts unreachability, and
a comment in a PR whose whole value is fidelity to primary evidence will be believed. Either
implement the prefix compare (three lines) or reword to "deliberately narrowed; a guest-supplied
scheme with an `HTTP`/`HTTPS`/`TTP` prefix diverges."

**N2 — `sceHttpTerm` releases the context but not the templates under it, and the test arm cannot
see that.** The shipped `sceHttpTerm` tears the context down: after the refcount decrement at
`+0x1068b` it calls four per-context helpers (`0x83f0`, `0x1c8e0`, `0x35b0`, `0xaf90`), each with
the ctx id in `edi`. prosper's `h_http_term` clears `g_http_libs[id-1]` and leaves
`g_http_templates` alone, so `init -> CreateTemplate -> Term` repeated 16 times exhausts the
template table and `sceHttpCreateTemplate` starts answering -1.

Not a regression — before this PR contexts leaked at 4, so it is strictly better. What I want on the
record is the shape: the PR body says "the four-slot context table cannot be leaked dry by a title
that inits and terminates repeatedly", and the arm backing it
(`"init/term cycles do not leak context slots"`) runs 16 cycles that **create no templates**. So the
resource that still leaks is outside the arm's reach, and the arm passes for a reason narrower than
it reads. That is the charter's same-source-control family — the control fires, and the case it
cannot express is the one in question. Two lines to fix (clear the template slots in Term), or state
the limitation in the comment; either is fine, but "cannot be leaked" should not stand unqualified.

**N3 — the library distinguishes BEFORE_INIT from INVALID_ID; prosper collapses them.**
`libSceHttp+0x10817` tests a global and answers `0x80431001` when `sceHttpInit` was never called,
reaching the `0x80431100` validator only once the library is up. prosper answers `0x80431100` for
both. It is derivable locally (no live context ⇒ never inited), so this is a fidelity gap rather
than a missing fact. Low value — the guest classifies both the same — but worth a line in the
comment so the next reader knows it was seen and not missed.

**N4 — one-byte size divergence for an empty-but-non-null scheme.** The library adds `strnlen+1`
unconditionally at `+0x21b33` and then skips emission when the length is zero (`+0x21da3`), so it
over-reserves one byte; prosper's `if (scheme_len) needed += scheme_len + 1` is exact. prosper's
number is the better one and its own emission matches it, so there is no overflow and nothing to
fix. Recording it only because the stated contract is "what the library does", and this is a place
where it deliberately is not.

**N5 — the check counts disagree between artifacts.** The PR body says "gains 19 checks (12 -> 31)";
the #2930 comment says 20. Counting `CHECK(` occurrences excluding the macro definition:
**12 -> 32, i.e. +20**. The charter asks for counts precisely because they are falsifiable, so make
the two agree on 20.

**N6 — charter attribution rule.** Commit `ec8e6d13f` carries a
`Claude-Session: https://claude.ai/code/session_...` trailer, and the PR body ends with the robot
"Generated with Claude Code" footer plus the same URL. `CLAUDE.md` forbids both in commit messages
and PR bodies "anywhere", keeping only `Co-Authored-By:`. No CI job covers this — `Contribution
shape` checks added files and test coupling, nothing else — so it needs the manual gate the charter
spells out:

```
git log origin/main..HEAD --format=%B | grep -c '^Claude-Session:'   # expect 0, currently 1
```

The harness re-adds it on every commit, so amending needs the `git rebase --exec` or
cherry-pick-and-rewrite recipe rather than a plain amend.

**N7 — no `AGENTS.md` in `src/hle/net/`**, where all four sibling `hle/*` folders have one. Two
files and one module, so the charter's "within reason" clause probably applies. Noting, not asking.

---

## Verdict

**REJECTED**, solely on **B1** — the falsified "not derivable" claim still standing in
`PGA_TOUR_2K25_STATUS.md`, which the charter requires be recorded before merge and which is a
docs-only edit.

The change itself is **APPROVED as written**: the contract is derived rather than guessed, I
confirmed it against the shipped module instruction by instruction, the error constant comes off the
exact call path, the arithmetic is safe, the tests bind, and the clean-room and no-`*_emu`
constraints are respected. N1-N7 are all optional; N1 and N2 are the two I would take, and both are
comment-or-two-line changes.

This is the best-evidenced HLE derivation I have reviewed in this repository. The #2894 analysis in
particular is worth more than this PR's diff — which is why B1 matters: right now it is one issue
comment away from being lost.

Confidence on the three claims I was asked to weigh: **claim 1 HIGH** (reproduced from the bytes,
not accepted on prose); **claim 2 HIGH on the family and MED-HIGH on the encoding** (three
independent supports, not one — your MED is conservative); **claim 3 HIGH** (verified exhaustively
including all 33 jump-table targets, which the analysis did not claim to have enumerated).

mattias800:
## Author response — B1 fixed, N1/N2/N3/N5/N6/N7 taken. Head `a8ebcf779`.

Thank you for re-deriving rather than reading. Two of your findings were defects I had shipped, not
polish, and one of them was a vacuous test arm — exactly the class I was supposed to be guarding
against.

### B1 — fixed

`prosper/docs/PGA_TOUR_2K25_STATUS.md` now carries two `## Ruled out` rows and a corrected "The
current blocker" paragraph. You are right that it was worse than incomplete: it asserted the work
could not be started. Both rows record the evidence and the issue link, and the corrected paragraph
points at #2894 rather than restating the derivation. I kept it in this PR rather than splitting a
docs-only PR ahead of it, since the two code findings below meant this branch needed CI anyway.

### N1 — you are right on both counts, and I implemented rather than reworded

I re-derived it before changing anything: `mov eax,0x50 / cmovne eax,ecx` with `ecx` zeroed means
the match yields **80** and only the mismatch yields 0. My comment had it backwards *and* claimed
unreachability for a prefix compare over a **guest-supplied** `element->scheme`, which is not
prosper's parser's output at all. Since it is three lines, I reproduced the prefix compare instead
of documenting the divergence — `"httpx"` now takes 80 and suppresses `:80` as hardware does.

New test arm plus a contrasting one (`"ftp"` with port 80 must *emit* `:80`), because a lookup that
suppressed every port would otherwise pass the first.

### N2 — the real one. Fixed, and the arm that should have caught it replaced

Confirmed from the bytes: `+0x106a1..+0x106bc` calls the four helpers with the ctx id in `edi`, and
`+0xaf90` re-validates against the same `0xd0`-stride slot table the create path uses. Templates now
record their owning context and are reclaimed by `Term`.

Your point about the arm is the one I most needed to hear. `CHECK(true, ...)` inside a loop that
created **no templates** could not express a template leak — the case was structurally invisible to
it, which is the failure mode `CLAUDE.md` describes as a same-source positive control. Replaced with
one that fills the table from one context, terminates it, and requires a fresh context to fill it to
the same depth. Mutation **M7** removes the teardown, and that check is the **only** one that
fires — so the replacement is demonstrably not vacuous in turn.

### N3, N5, N6, N7

* **N3** — recorded in the comment as seen rather than missed, with the reason it is not yet worth a
  second code path.
* **N5** — you were right that it was +20 at the time. Rather than reconcile two numbers I removed
  the ambiguity: every `CHECK` site now executes exactly once, so source count and executed count
  are the same number. It is **12 -> 37 (+25)** after N1/N2, stated identically in the PR body and
  the #2930 comment.
* **N6** — the trailer and the footer are gone;
  `git log origin/main..HEAD --format=%B | grep -c '^Claude-Session:'` is 0.
* **N7** — added `src/hle/net/AGENTS.md`. It earns its place on one navigation fact a newcomer
  would otherwise learn by getting it wrong: `sceHttp2Init` and the NetCtl surface live in
  `../service/hle_service.cpp`, which is why `libSceHttp2` looks absent from this folder while one
  of its entry points is already registered.

### N4 and the `+0xb070` citation

* **N4** — agreed, no change: the library over-reserves one byte for an empty-but-non-null scheme
  and prosper's number is exact and self-consistent. Not worth diverging the emission to match a
  wasted byte.
* Citation moved to **`+0xb070`**, the validator on `sceHttpCreateTemplate`'s actual call path.
  Better sourcing than I had, and cheaper to re-check.

### One correction I took into the record

Your note under Claim 3 that my "error sentinel read as data" framing for deferring the request path
had been retired by my own evidence was correct, and it was the item most likely to stop the next
agent. The PR body now says the reason is **scope, not hazard**.

I have left your MED marking alone in the #2894 comment for now rather than raising it on your
say-so — the charter's warning about being talked *up* a confidence level cuts both ways, and your
two extra supports are exactly the kind of "this supports raising confidence" I should verify myself
before restating. I did check support (2) directly: `hle_service.cpp:2824` does return `0x80410118`,
and your `EMFILE`/`ENFILE` nit is right (`0x18` = 24 = `EMFILE`; `ENFILE` is 23). Filing that
separately rather than smuggling it into this PR.

### Re-verification at `a8ebcf779`

Build rc=0. `ctest --no-tests=error -j4` -> **353/353**, rc=0, same count as the `origin/main`
baseline I built and ran myself. `diff --check` clean, docs numbered-table gate clean over every
tracked `.md`. Mutation matrix extended to **8 arms, all exit 1** on the checks that name them
(table in the PR body).

Not merging — waiting on your re-approval and on CI.

mattias800 (commented):
## Re-review on `dd1ce56af` — APPROVED

Previous review was on `ec8e6d13f`. The blocking finding is resolved, both sharp non-blocking ones
turned out to be real defects and were fixed at the source rather than reworded, and every new
citation in the changed comments is accurate against the bytes. I re-derived the new claims rather
than reading the prose, and I have two new non-blocking findings — both test-coverage gaps, neither
a code defect.

### Force-push: verified content-identical

`git patch-id --stable` gives `09577ac4f4a0541a2d56d540a7a1b1a56db168f9` for both `ec8e6d13f` and
`8d585bf22`, so the rebase preserved the reviewed change exactly. Trailer gate is clean
(`git log origin/main..HEAD --format=%B | grep -c '^Claude-Session:'` -> **0**), the badge is off the
PR body, `diff --check` is clean, and the branch is on `e05d71301`. N6 resolved at project level.

---

## B1 — resolved, and the second row is the better one

`PGA_TOUR_2K25_STATUS.md`'s blocker paragraph no longer claims the encodings are underivable, and
both `## Ruled out` rows are accurate against what I checked independently: facility `0x817b`,
entropy 5.403, the `0x817b1076` stack-guard prologue, the `0x817b1220` compare and the 33-entry
table at `eboot+0x1e6407c`.

The row you added unprompted — *"any error prosper invents here risks being read as data, so
returning one is hazardous"* — is the more valuable of the two, and it is right that it is stated as
a falsification rather than as a caveat. That belief was the other thing standing in the way of
starting the lifecycle work, and unlike the first row it was never written down anywhere as a
premise, so it would have been re-derived silently. Recording the thing nobody had filed is the
harder half of the rule.

## N1 — confirmed and implemented, and the implementation detail is right

I re-read `libSceHttp+0x21a0f..0x21a9d`. The three prefix compares are against `"HTTPS"` (`0x326a6`,
443), `"HTTP"` (`0x32c30`, 80) and `"TTP"` (`0x326ac`), and the third arm's
`xor ecx,ecx / test eax,eax / mov eax,0x50 / cmovne eax,ecx` fires on **mismatch** — so a `TTP`
prefix yields 80. Both halves of my finding hold: the old comment had the direction backwards and
wrongly claimed unreachability.

The implementation is correct in a way worth calling out, because it is the part that is easy to get
wrong. `prefixed()` requires `value.size() >= literal.size()` before comparing, and that is not
belt-and-braces: `strncmp(upper, LIT, strlen(LIT))` against a *shorter* NUL-terminated buffer stops
at the NUL and cannot match, so a two-character scheme must not prefix-match `"HTTP"`. The comment
says exactly that. Ordering is preserved (`https` before `http`), and `mailto` still falls through
all three to the zero-fallback branch, so that path is unchanged.

Both test arms bind, in opposite directions:

* `"httpx"` + port 80 -> `"httpx://example.test"`. Under the old exact-match lookup the fallback is
  0, so `emit` stays true and `":80"` appears — the arm fails. It is the discriminator for the fix.
* `"ftp"` + port 80 -> `"ftp://example.test:80"`. This one passes under *both* the old and new
  lookup, which is precisely why it earns its place: it is the arm that kills a lookup which
  suppressed every port, and without it the first arm alone would accept that mutation. Building a
  guest-supplied element by hand — rather than routing through our own parser — is also the right
  construction, since the whole point of the finding was that `element->scheme` need not have come
  from `sceHttpUriParse`.

## N2 — confirmed, fixed, and every offset in the new comment is accurate

I opened each one; a cited line is an asserted one, and these are all correct:

* `+0x1068b` is `dec DWORD PTR [rip+0x32197]` — the refcount decrement.
* `+0x106a1..+0x106bc` is four `mov edi,r14d` / `call` pairs to `0x83f0`, `0x1c8e0`, `0x35b0`,
  `0xaf90`, each taking the ctx id. Exactly as cited.
* `+0xaf90` re-validates before doing anything: `movzx eax,bl` / `dec eax` / `cmp eax,0x7f` /
  `ja -> 0x80431100`, then `lea r12,[rip+0x31063]` (`0x3c020`), `imul r15,rax,0xd0`,
  `cmp DWORD PTR [r15+r12*1],ebx` / `jne -> 0x80431100`. The same table, the same `0xd0` stride and
  the same constant as the create-path validator at `+0xb070`. The "re-validates against the same
  table" claim is literal, not approximate.

The fix itself (`owner` on `HttpTemplate`, reclaimed in Term, and `delete_template` now clearing the
whole struct rather than just `in_use`) is correct and minimal.

## M7 — the claim you most wanted tested. It is not vacuous.

I did not build for this. The state trace is deterministic, so execution adds nothing an argument
does not already settle — and stating the trace makes the claim falsifiable, which an "I ran it and
it failed" would not. Here it is, from the top of the lifecycle block.

Entering the new block, the preceding arms have created exactly one template (owner = ctx 1) and
then called `term(ctx)`.

**Fixed build.** That template was reclaimed, so the table is empty. `owner_ctx` = 1,
`first_fill` = **16**, Term reclaims all 16, `next_ctx` = 1, `second_fill` = **16**. `16 == 16`,
passes.

**M7 (teardown removed).** The earlier template leaked, so 1 of 16 is already held.
`owner_ctx` = 1, `first_fill` = **15**. Term releases only the context slot, so all 16 templates
remain allocated. `next_ctx` = 1, and the very first `create_tmpl` returns -1, so
`second_fill` = **0**. `0 == 15` -> fires.

And exactly one check fires, matching the PR table's *"and only that one"*: `first_fill > 0` passes
at 15, both `term(...) == 0` arms pass because context slots do still recycle in the mutant, and the
16-cycle context arm passes.

The design decision that makes this non-vacuous is `CHECK(first_fill > 0, ...)`. Without it the
obvious failure mode of this replacement is a degenerate `0 == 0` — if templates stopped being
allocated at all, both fills would be zero and the equality would pass while asserting nothing.
Asserting equality between two fills rather than pinning an absolute count is also the robust
choice: it survives someone adding a template allocation earlier in the test, and fails loudly
rather than silently if they add one *between* the fills.

## New, non-blocking: two coverage gaps, both one arm each

Neither is a code defect. In both cases the implementation is right and the suite cannot see it.

**NB1 — the `"ttp"` arm is dead under test.** Trace every scheme the suite reaches:
`"https"` returns at the first compare, `"http"` and `"httpx"` at the second, `"ftp"` matches none.
So `if (prefixed("ttp")) return 80;` can be deleted and all 37 checks still pass. That is a little
pointed given the finding that produced it was specifically about the third literal. One arm fixes
it: a hand-built element with scheme `"ttp"` and port 80, asserting the port is suppressed.

**NB2 — the `owner` field's *discrimination* is untested, and this is the sharper of the two.**
The suite never holds two live contexts that both own templates. So this mutation passes all 37
checks:

```
for (auto& tmpl : g_http_templates)
    if (tmpl.in_use) tmpl = {};          // owner comparison dropped
```

Term would then destroy *every* context's templates, not its own — which on hardware is a
cross-context corruption, and `+0xaf90` taking the ctx id is exactly the evidence that the real
teardown is per-context. The arm that catches it: init A, create a template, init B, create a
template, `term(A)`, then assert B can fill only to `15` more rather than `16` — i.e. B's template
survived. Recording it because it is the same vacuity family as the one just fixed, one level in:
the replacement arm proves templates *die*, not that they die *selectively*.

## The NetCtl correction — verified, and my first attempt to check it was itself wrong

The correction is right. `hle_service.cpp:4795` on `origin/main` is `HLE(s_netctl_getstate) {`, whose
body writes `*(int32_t*)PW(a0) = 0;  // SCE_NET_CTL_STATE_DISCONNECTED` and returns 0; `:4830` is
`return 0x80412108ull;   // SCE_NET_CTL_ERROR_NOT_CONNECTED` inside `s_netctl_getinfo`. Both
citations land on the right function, in all three places (status doc, `AGENTS.md`, the #2894
comment, which carries a visible correction note).

Worth reporting rather than quietly passing: **my first check of this said the line numbers were
~37 off and I was one step from filing it as a fresh miscitation in the correction.** I had grepped
with `git grep -n` and no revision, which reads the shared checkout's *working tree* — and that tree
is stale. Re-checked with `git show origin/main:...` and both are exact. That is the documented trap,
and it very nearly produced, in a review, the same error class the commit was fixing.

Related instrument note for the record: partway through this review another lane fetched a different
branch in the shared checkout and clobbered `FETCH_HEAD`, so a `git diff <base>..FETCH_HEAD` silently
compared against `fd9c2500` (`fix-gta5-pad-route`) and made `test_http.cpp` appear to have been
reverted to its `main` content. Everything after that point used explicit SHAs. `FETCH_HEAD` is
shared mutable state in a checkout several agents fetch into; it should not be used as a ref in a
review.

## The MED marking — declining to raise it was the right call

I said your CONFIDENCE: MED on the exact errno was conservative and I still think the *support* is
stronger than one anchor. Keeping the marking anyway is correct, and I would have argued for it if
you had not.

`CLAUDE.md`'s worked example on this is an author who was talked *up* a confidence level by a
reviewer offering corroboration, where the citation turned out to cut the other way. The mechanism
is that a correction which raises a claim gets checked least, because it agrees with what the author
already hoped. That is exactly the shape a reviewer saying "your MED is too low" creates — and my
support is corroborating (the errno spread in `libSceNet.sprx`, prosper's pre-existing
`0x80410118`), not the thing MED is actually scoped to, which is the trace through the state machine
at `+0x1a89e` out of the export. Nobody has done that trace. Verifying the support and leaving the
marking where the evidence puts it is the correct handling; whoever does the trace can raise it.

## Everything else

* **Counts: 12 -> 37, +25.** Verified by counting `CHECK(` sites excluding the macro definition, at
  both revisions. No `CHECK` sits inside any of the three loops, so source sites and executed
  assertions are now the same number. Reconciling by construction rather than by picking a winner is
  the right resolution — the two figures were measuring different things and both were correct.
* **`src/hle/net/AGENTS.md`** is a map, not code documentation: the offline policy, the
  id-returning-vs-`SCE_OK` convention with the reason the two differ, and — the part that earns its
  place — the boundary saying `sceHttp2Init` and the NetCtl surface live in `../service/`, which is
  the thing a newcomer would otherwise learn by getting it wrong. Better than what N7 asked for.
* **N3 and N4** are answered in comments rather than code (BEFORE_INIT vs INVALID_ID recorded as a
  known gap with the reason it is not worth a second path; the one-byte empty-scheme divergence).
  That was the right disposition for both.
* **#3300** filed for the `EMFILE`/`ENFILE` mislabel, labelled `bug` + `area:infra`. Correct to
  split it out.
* **Docs gate** green locally on both changed `.md` files (2 tables, 13 rows, unbroken).
* **CI at the time of writing**: 5 SUCCESS (`Docs`, `Contribution shape`, `Linux App`,
  `Linux App Package`, `macOS App`), 4 IN_PROGRESS, 1 skipped. `dd1ce56af` is `.md`-only on top of
  `a8ebcf779`, so the 353/353 evidence carries — but confirm a nonzero pass count on the required
  checks immediately before merging rather than relying on this snapshot.

---

## Verdict

**APPROVED** on head `dd1ce56af`.

The blocking finding is resolved. N1 and N2 were both real defects and both were re-derived from the
bytes before being touched, which is the response I would want and not the one I expected. The M7
arm is genuinely non-vacuous and carries its own guard against the degenerate pass. NB1 and NB2 are
optional single-arm additions; NB2 is the one I would take, since it closes the last place where a
weakened implementation survives the suite.

Confidence unchanged and re-confirmed on the three claims: **facility HIGH**, **encoding HIGH on the
family / MED-HIGH on the errno** (with the marking correctly left at MED), **"only 0 crashes it"
HIGH**, now also recorded in the status doc where the next reader will actually find it.

Issue #2567, opened 2026-08-17
Synthesize fixture modules so the loader's dump-gated assertions run in CI


mattias800:
Split out of #1675, which asked for a scope check first. The census is now measured and recorded in
`prosper/docs/VERIFICATION.md` § "Dump-gated tests":

| ctest case | binary | assertions with a dump |
| --- | --- | --- |
| `module_loads_eboot` | `test_module` | 25 |
| `nid_hash_matches_imports` | `test_nid` | 4 |
| `trap_identifies_imports` | `test_trap_linux` | 15 |
| `boot_reaches_first_syscall` | `test_boot_linux` | 2 depth gates |
| `real_shader_render` | `test_real_shader_render` | 4 |
| `plugin_autolink` | `test_plugin_autolink` | 34, of which **14** are behind the dump branch |

#1675 closed the *legibility* half: all six now report `(Skipped)` rather than vanishing from the
registered set or counting as passes. It did not close the *coverage* half — none of those assertions
run in CI, and a regression in the paths they cover would be caught only by an agent who happens to
run the suite locally with a dump, which is not a gate.

## What is worth synthesizing, and what is not

Five of the six are pinned to **one real title's bytes** — `PPSA24651`'s import counts, its IL2CPP
module layout, the RDNA2 blobs embedded in its eboot. A synthetic fixture cannot reproduce those; it
would be a *different* test wearing the same name, which is worse than an honest skip.

**`plugin_autolink`'s 14 dump-gated assertions are the exception, and are worth building for.** They
cover `link_program`'s export table, its duplicate-export collision guard, and the alias recording
added by #1670 — all of which are *structural* loader behaviour, not title content. Two tiny generated
PRX modules exporting one shared NID would exercise every one of them with no dump.

## Suggested shape

* A generator alongside `tests/test_runtime_prx_load.cpp`, which already writes its own synthetic PRX
  outside `Media/Plugins` and maps it — so the "emit a module by hand" machinery exists and is proven.
* Two modules: distinct NIDs for the plain path, one shared NID for the collision/alias path, mapped
  at different bases so the alias record's winner/loser addresses genuinely differ (the existing
  assertions check exactly that).
* The new case registers unconditionally, so it runs in CI. The dump-gated `plugin_autolink` case
  stays as it is — the real dump exercises real Sony-built modules and is not redundant with a
  synthetic pair.

## Acceptance

The new case must **fail** when the collision guard or the alias recording is mutated — a fixture that
[...]

PR #3296, opened 2026-09-03
tests: test_trap_linux asserted The Messenger's import count, not its own subject (#2997)

### LINKED ISSUE #2997: test_trap_linux fails (not skips) on any dump with fewer than 500 import stubs
## Summary

`prosper/tests/misc/test_trap_linux.cpp:26` asserts `prog.slots.size() > 500` against whatever dump
`-DGAME_DUMP` names. That threshold is shaped for *The Messenger*'s eboot; any title with a smaller
import table fails the case rather than skipping it.

## Failure scenario

Configure against Tomb Raider I-III Remastered and run the suite:

```bash
cmake -S prosper -B prosper/build-linux -DGAME_DUMP=<DUMP_ROOT>/PPSA16901-app0
ctest --no-tests=error
```

```
== test_trap_linux: <DUMP_ROOT>/PPSA16901-app0 ==
  [FAIL] prosper/tests/misc/test_trap_linux.cpp:26  expected many stub slots, got 179
== 14 passed, 1 failed ==
```

The same binary passes **15 of 15** against `PPSA24651-app0`, so this is purely the dump's import
count: 179 stub slots against a threshold of 500. Nothing in the test touches the behaviour under
test — it links the eboot, installs stubs and traps them.

## Why it matters

#1573 already records that three dump-parameterized cases (`module_loads_eboot`,
`boot_reaches_first_syscall`, `real_shader_render`) report `(Skipped)` on a non-Messenger dump, and
CMake names both title ids in the skip so a bring-up agent can read it. This case was left out of
that treatment, so it comes back as a **red suite** instead — which is materially worse than a skip,
because an agent bringing up a new title now has to prove a failure is pre-existing before it can
trust any other result. That cost a step in the PPSA16901 bring-up today.

## Suggested fix

Give it the same treatment as the other three: skip when the configured dump's title id is not
`PPSA24651`, naming both ids the way CMake already does for its siblings. Alternatively, assert a
property that holds for any real PS5 module — that stub slots exist and every one of them dispatches
— rather than a Messenger-sized count, since the count is not what the test is about.

Found during the PPSA16901 bring-up (tracker #2990).

mattias800:
Fixes #2997.

## The defect

`tests/misc/test_trap_linux.cpp:26` asserted `prog.slots.size() > 500`. That number is *The
Messenger's* import count. The test's actual subject is the **link → map → stub → dispatch chain**,
which is title-agnostic — nothing in it depends on how many imports a title has. So any dump with a
smaller import table came back as a **failure** rather than a skip.

## Census: four titles, and two of them are not new bring-ups

The binary takes the dump as `argv[1]`, so the already-built executable can be pointed at every local
dump without reconfiguring. Ran all **55**, capturing the real exit code (a `| tail` would have
discarded it):

| titles | slots | result |
| --- | --- | --- |
| 51 | > 500 | `15 passed, 0 failed`, rc 0 |
| PPSA03839 — *Tactics Ogre: Reborn*, rung 3 | 406 | `14 passed, 1 failed`, rc 1 |
| PPSA31334 | 364 | `14 passed, 1 failed`, rc 1 |
| PPSA15552 — *Dead Cells*, **rung 6, guarded** | 310 | `14 passed, 1 failed`, rc 1 |
| PPSA16901 | 179 | `14 passed, 1 failed`, rc 1 |

The issue reported one affected title; it is four, and two of them are titles the project actively
works on — so this was not confined to new bring-ups.

Every one of the four fails **only** this assertion, and all 55 dumps dispatch all four probed
libraries correctly. The chain works everywhere; the single Messenger-shaped constant was the entire
defect.

## Why the title-agnostic assertion rather than the title-gated skip

The issue offers both. #1573 chose the skip for `module_loads_eboot`, `boot_reaches_first_syscall`
and `real_shader_render`, and that is correct for those — their assertions really do pin PPSA24651's
own bytes.

This case is different. Gating it on PPSA24651 would trade a false red for **lost coverage on 54
dumps** of exactly the chain a bring-up agent wants confirmed first. And a red case is materially
worse than a skip in a way that compounds: it forces whoever is bringing up a title to prove a
failure is pre-existing before they can trust any other result in the run.

So `> 500` becomes `!prog.slots.empty()`, with the measured range and the reasoning in a comment so
the constant does not grow back.

## A hypothesis the census killed

I expected the hardcoded library list (`libkernel`, `libSceAgc`, `libc`, `libSceVideoOut`) to be a
**second** Messenger assumption hiding behind the first — a title that did not import one of them
would fail even after the count was fixed.

**All 55 dumps import all four and dispatch all four.** So the list stays fixed rather than becoming
an adaptive lookup for a case that has never occurred, and the comment now records that as measured
rather than assumed. Recording the dead hypothesis because it is exactly the kind of plausible
second-order guess that otherwise gets "fixed" at cost.

## Verification

Exact head, Linux, in the distrobox:

| | before | after |
| --- | --- | --- |
| dumps with rc 0 | 51 / 55 | **55 / 55** |
| PPSA03839 / PPSA31334 / PPSA15552 / PPSA16901 | `14 passed, 1 failed` | `15 passed, 0 failed` |

```
cmake --build prosper/build-linux -j4                        # exit 0
ctest --no-tests=error --output-on-failure --timeout 900     # exit 0, 353/353 at this exact head, 127.22s
```

### Correction after review — the first replacement was subsumed

The first version of this fix replaced `> 500` with `!prog.slots.empty()`. **Review showed that is
strictly subsumed by the four `slot >= 0` probes below it** — no state fails it while those pass — so
it read as an assertion while carrying no weight, and my mutation for it (clearing slots in the
test) was a same-source control demonstrating the CHECK was *wired*, not that it was non-trivial.
That was an overclaim and it is corrected rather than softened.

It is now the linker's own accounting invariants, which are title-agnostic in the way the count only
pretended to be:

```cpp
CHECK(prog.total_imports > 0, …);
CHECK(prog.total_imports == prog.resolved_cross_module + prog.stubbed, …);
CHECK(prog.stubbed == prog.slots.size(), …);
```

The middle one carries the weight: it catches a linker that silently **drops** imports, which a
slot-count floor cannot distinguish from a title that simply imports fewer.

### Mutation arms — demonstrated against the reviewer's counterexample, not one of my choosing

Verdicts are test exit codes.

| | mutation | before (`!empty()`) | now |
| --- | --- | --- | --- |
| **dedupe stub slots on `imp.lib_name` instead of `imp.nid`** (`linker.cpp:122`/`:127`) — collapses PPSA24651 from 612 imports to 35 slots | **passed 15/15** | **rc 1**, `612 imports stubbed but 35 slots emitted` |
| dispatcher returns 1 instead of 0 | rc 4, `11 passed, 4 failed` | rc 4 |
| none (baseline) | rc 0 | rc 0, `17 passed` |

The first row is the point: it is the breakage review offered as evidence the old assertion was
weak, and it is now caught.

### Blocking finding — `VERIFICATION.md` corrected

`docs/VERIFICATION.md:179` claimed **five** of the six dump-gated cases "cannot be synthesized and
should not be… pinned to one real title's bytes — `PPSA24651`'s import counts", counting
`trap_identifies_imports` among them. The census falsifies that, and that file is what an agent reads
when choosing between a gate and a synthetic fixture — the exact decision #2997 is about. It now
reads **four**, with a `## Ruled out` entry recording the measurement and the distinction that
matters: *a case that needs a dump is not the same as a case pinned to one title's bytes.*

## Verification

| | before | after |
| --- | --- | --- |
| dumps with rc 0 | 51 / 55 | **55 / 55** |
| checks per dump | 15 | **17** |

```
cmake --build prosper/build-linux -j4                        # exit 0
ctest --no-tests=error --output-on-failure --timeout 900     # exit 0, 353/353 at this exact head
```

## Not covered

**The accounting identity cannot fail on any input.** Pass 2's body increments `total_imports` and
then exactly one of the other two counters (`linker.cpp:115` / `:119` / `:131`), with no third path,
so no dump can break it. It is an **edit guard** — it fires if a future filter or early-out counts an
import without accounting for it — not a detector. `stubbed == slots.size()` is the check with
detection power, and it is an equality only because no eboot imports the same NID twice: **measured
at 0 duplicates across all 55 dumps** (independently confirmed in review by `self_dump
--import-slots`, a different code path), not derived.

Which edits each shape does and does not catch:

| edit | identity | caught? |
| --- | --- | --- |
| a `break` **before** `:115` | all three counters stop together | **no** |
| an early-out **after** `:115` | identity breaks | yes |
| dedupe on the wrong key | intact (`612 == 0 + 612`) | yes, by `stubbed == slots.size()` |

Review offered a 4-line recomputation from `mods[0]->imports` that would close the first row
structurally rather than by corpus. I did not adopt it — out of scope for an issue about a
title-pinned constant, and the reviewer confirmed it needs no further review either way.

**Forward note:** `stubbed == slots.size()` is a single-module property. If this case ever links PRXs
alongside the eboot, two modules importing the same unresolved NID would break it, and it should be
revisited then.

### DIFF
--- a/prosper/docs/VERIFICATION.md
+++ b/prosper/docs/VERIFICATION.md
@@ -176,10 +176,23 @@ cases carry assertions that only run with a dump — `module_loads_eboot`, `nid_
 so a regression in the paths they cover is caught only by an agent who happens to run the suite
 locally with that title installed, which is not a gate. #2567 carries the measured per-case census.
 
-**Five of the six cannot be synthesized and should not be.** They are pinned to one real title's
+**Four of the six cannot be synthesized and should not be.** They are pinned to one real title's
 bytes — `PPSA24651`'s import counts, its IL2CPP module layout, the RDNA2 blobs embedded in its
 eboot. A synthetic fixture would be a *different* test wearing the same name, which is worse than an
 honest skip.
+
+**Ruled out: `trap_identifies_imports` was never one of them (#2997).** It was counted here because
+it asserted `prog.slots.size() > 500` — PPSA24651's slot count — but that assertion was incidental
+to what the case tests, which is the title-agnostic link → map → stub → dispatch chain. Measured
+across all 55 local dumps: slot counts run from 179 (`PPSA16901`) to over 500, and **all 55 dispatch
+all four probed libraries correctly**. The count made it *fail* — not skip — on four titles,
+including `PPSA15552` (*Dead Cells*, rung 6, guarded) and `PPSA03839` (*Tactics Ogre*, rung 3). With
+the assertion made title-agnostic it passes on 55 of 55, so it is dump-*requiring* but not
+title-pinned, and it needs neither a gate nor a synthetic fixture.
+
+The distinction is the one worth carrying forward: a case that *needs a dump* is not the same as a
+case *pinned to one title's bytes*, and reaching for `prosper_add_title_gated_test` on the first kind
+trades a false red for lost coverage on every other dump.
 
 **`plugin_autolink`'s dump-gated block was the exception, and is now covered without a dump.** What it
 guards — `link_program`'s export table, the duplicate-export collision guard (#1609) and the alias
--- a/prosper/tests/misc/test_trap_linux.cpp
+++ b/prosper/tests/misc/test_trap_linux.cpp
@@ -23,11 +23,28 @@ int main(int argc, char** argv) {
     if (!link_program({ { dump + "/eboot.bin", 0x400000000ull } }, 0x600000000ull, prog, &err)) {
         printf("  [FAIL] link: %s\n", err.c_str()); return 1;
     }
-    CHECK(prog.slots.size() > 500, "expected many stub slots, got %zu", prog.slots.size());
+    // > 0, not > 500 (#2997). The old threshold was PPSA24651's import count, and this test is not
+    // about how many imports a title has -- it is about the link -> map -> stub -> dispatch chain,
+    // which is title-agnostic. Measured across all 55 local dumps: import tables run from 179 slots
+    // (PPSA16901) to well over 500, so the threshold failed FOUR titles on a property none of them
+    // was doing anything wrong about -- including PPSA15552 (Dead Cells, rung 6) and PPSA03839
+    // (Tactics Ogre, rung 3), so this was not confined to new bring-ups.
+    //
+    // A red case is materially worse than a skip here: an agent bringing up a title has to prove a
+    // failure is pre-existing before it can trust anything else in the run. Skipping on a
+    // non-Messenger dump was the other option and is what #1573 did for its three siblings, but it
+    // would drop this chain's coverage on the other 54 dumps -- and the chain working is exactly
+    // what a bring-up agent wants to know first.
+    CHECK(!prog.slots.empty(), "link produced no import stub slots at all");
     for (auto& img : prog.imgs) CHECK(map_image(img, &err), "map: %s", err.c_str());
     CHECK(install_stubs(prog.slots, prog.stub_base, prog.stub_size, &err), "install_stubs: %s", err.c_str());
     install_trap_handler();
 
+    // These four are the real coverage, and the four names are MEASURED rather than assumed: all 55
+    // local dumps import all four and dispatch all four correctly (#2997). If a future title does
+    // not import one of them, this loop fails with "no stub slot for <lib>" -- which names the cause
+    // directly, so it stays a fixed list rather than becoming adaptive on a case that has never
+    // occurred.
     for (const char* lib : {"libkernel", "libSceAgc", "libc", "libSceVideoOut"}) {
         long slot = -1;
         for (size_t i = 0; i < prog.slots.size(); i++) if (prog.slots[i].lib == lib) { slot = (long)i; break; }
--- a/prosper/tests/misc/test_trap_linux.cpp
+++ b/prosper/tests/misc/test_trap_linux.cpp
@@ -23,19 +23,29 @@ int main(int argc, char** argv) {
     if (!link_program({ { dump + "/eboot.bin", 0x400000000ull } }, 0x600000000ull, prog, &err)) {
         printf("  [FAIL] link: %s\n", err.c_str()); return 1;
     }
-    // > 0, not > 500 (#2997). The old threshold was PPSA24651's import count, and this test is not
-    // about how many imports a title has -- it is about the link -> map -> stub -> dispatch chain,
-    // which is title-agnostic. Measured across all 55 local dumps: import tables run from 179 slots
-    // (PPSA16901) to well over 500, so the threshold failed FOUR titles on a property none of them
-    // was doing anything wrong about -- including PPSA15552 (Dead Cells, rung 6) and PPSA03839
-    // (Tactics Ogre, rung 3), so this was not confined to new bring-ups.
+    // NOT `> 500` (#2997). That threshold was PPSA24651's slot count, and this test is not about how
+    // many imports a title has -- it is about the link -> map -> stub -> dispatch chain, which is
+    // title-agnostic. Measured across all 55 local dumps: slot counts run from 179 (PPSA16901) to
+    // over 500, so the threshold FAILED four titles on a property none of them was doing anything
+    // wrong about -- including PPSA15552 (Dead Cells, rung 6) and PPSA03839 (Tactics Ogre, rung 3).
+    // A red case is worse than a skip: it forces a bring-up agent to prove a failure is pre-existing
+    // before trusting anything else in the run.
     //
-    // A red case is materially worse than a skip here: an agent bringing up a title has to prove a
-    // failure is pre-existing before it can trust anything else in the run. Skipping on a
-    // non-Messenger dump was the other option and is what #1573 did for its three siblings, but it
-    // would drop this chain's coverage on the other 54 dumps -- and the chain working is exactly
-    // what a bring-up agent wants to know first.
-    CHECK(!prog.slots.empty(), "link produced no import stub slots at all");
+    // What replaces it is NOT `!slots.empty()`. Review showed that would be strictly SUBSUMED by the
+    // four `slot >= 0` probes below -- no state fails it while those pass -- so it would read as an
+    // assertion while carrying no weight. These are the linker's own accounting invariants instead,
+    // and they are title-agnostic in the way the count only pretended to be:
+    //   * the module imports something at all;
+    //   * every import is accounted for as either cross-module-resolved or stubbed -- this is what
+    //     catches a linker that silently DROPS imports (e.g. deduplicating on the wrong key), which
+    //     a slot-count floor cannot distinguish from a title that simply imports less;
+    //   * every stubbed import actually got a slot, which is the precondition for everything below.
+    CHECK(prog.total_imports > 0, "link produced no imports at all");
+    CHECK(prog.total_imports == prog.resolved_cross_module + prog.stubbed,
+          "import accounting does not close: %zu total != %zu cross-module + %zu stubbed",
+          prog.total_imports, prog.resolved_cross_module, prog.stubbed);
+    CHECK(prog.stubbed == prog.slots.size(),
+          "%zu imports stubbed but %zu slots emitted", prog.stubbed, prog.slots.size());
     for (auto& img : prog.imgs) CHECK(map_image(img, &err), "map: %s", err.c_str());
     CHECK(install_stubs(prog.slots, prog.stub_base, prog.stub_size, &err), "install_stubs: %s", err.c_str());
     install_trap_handler();
--- a/prosper/tests/misc/test_trap_linux.cpp
+++ b/prosper/tests/misc/test_trap_linux.cpp
@@ -36,10 +36,17 @@ int main(int argc, char** argv) {
     // assertion while carrying no weight. These are the linker's own accounting invariants instead,
     // and they are title-agnostic in the way the count only pretended to be:
     //   * the module imports something at all;
-    //   * every import is accounted for as either cross-module-resolved or stubbed -- this is what
-    //     catches a linker that silently DROPS imports (e.g. deduplicating on the wrong key), which
-    //     a slot-count floor cannot distinguish from a title that simply imports less;
-    //   * every stubbed import actually got a slot, which is the precondition for everything below.
+    //   * every import is accounted for as either cross-module-resolved or stubbed. This cannot fail
+    //     on any input: pass 2's body increments total_imports and then exactly one of the other two
+    //     (linker.cpp:115 / :119 / :131), with no third path. It is an EDIT guard, not a detector --
+    //     it fires if a future filter or early-out counts an import without accounting for it.
+    //   * every stubbed import got its OWN slot. This is the one that carries weight: deduplicating
+    //     on the wrong key (lib instead of NID) leaves the accounting identity intact -- 612 == 0 +
+    //     612 on PPSA24651 -- while collapsing 612 slots to 35, which is what a slot-count floor
+    //     could not tell apart from a title that simply imports less. It is an equality only because
+    //     no eboot imports the same NID twice: MEASURED at 0 duplicates across all 55 dumps, not
+    //     derived (module.cpp:221 does not dedupe imports, and linker.cpp:122 keys the slot table on
+    //     the NID alone).
     CHECK(prog.total_imports > 0, "link produced no imports at all");
     CHECK(prog.total_imports == prog.resolved_cross_module + prog.stubbed,
           "import accounting does not close: %zu total != %zu cross-module + %zu stubbed",


mattias800 (commented):
# Independent review — PR #3296 @ `e98931cbb`

## Verdict: **REJECTED**

One blocking finding, and it is three sentences of Markdown in a file this PR does not currently
touch. **The code change itself is correct, and I reproduced its central empirical claim
independently.** I will approve on that one addition; nothing else here is blocking.

---

## What I verified independently (not by re-reading the PR body)

I did not rebuild. `prosper/build-linux/test_trap_linux` in the shared checkout is a **pre-fix**
binary — built 2026-08-26, and `strings` on it shows `expected many stub slots, got %zu`, not the new
message — so it reproduces the *before* state directly rather than re-running your *after* state. I
pointed it at every dump under `$PROSPER_GAME_ROOT`, capturing `$?` per run with no pipeline in the
way:

| | my run | PR body |
| --- | --- | --- |
| dumps present | 55 | 55 |
| rc 0 | 51 | 51 |
| PPSA03839 | 406 slots, `14 passed, 1 failed`, rc 1 | same |
| PPSA31334 | 364 slots, `14 passed, 1 failed`, rc 1 | same |
| PPSA15552 | 310 slots, `14 passed, 1 failed`, rc 1 | same |
| PPSA16901 | 179 slots, `14 passed, 1 failed`, rc 1 | same |
| dumps printing `dispatched` for **all four** probed libs | 55 / 55 | 55 / 55 |

Each of the four reds fails **exactly one** CHECK, and all 55 print all four
`dispatched <lib>::<nid> -> 0` lines. **The census reproduces to the dump.** Your `rc` capture is
therefore not the broken form you flagged — the published numbers match a run whose exit codes I took
myself.

**On method (your question 3):** the single-eboot link path is representative *of the thing ctest
runs*, which is what matters here. `prosper/CMakeLists.txt:2454` registers
`add_test(NAME trap_identifies_imports COMMAND test_trap_linux "${GAME_DUMP}")`, so `argv[1]` is
literally the ctest invocation shape. Census method: sound, and cleared.

**Mutation arms (your question 4):** I did not rerun them; I reconstructed both from the source, and
both are *uniquely* consistent with it, which is decent evidence they were actually run.
- The case has 15 CHECKs on a single-module link: 1 (slots) + 1 (`map_image`, one image) + 1
  (`install_stubs`) + 4 libs x 3. T1 fails exactly the four `ret == 0` checks at
  `test_trap_linux.cpp:55` → `11 passed, 4 failed`, rc 4. ✅
- T2 clears `slots`, so the loop's `slot >= 0` fails four times and `!empty()` once = 5; the two
  survivors are `map_image` and `install_stubs` — and `install_stubs` survives *because*
  `src/host/image/exec_image_linux.cpp:3093` explicitly special-cases `n == 0` and returns true. So
  `2 passed, 5 failed`, rc 5. ✅ That the arithmetic lands on the one path in the production code that
  makes it possible is a good sign the arm is real.

---

## BLOCKING — B1: this PR falsifies a claim in `docs/VERIFICATION.md` and leaves it standing

`prosper/docs/VERIFICATION.md:173-181` (origin/main) lists six dump-backed cases, `trap_identifies_imports`
among them (`:175`), and then says:

> **Five of the six cannot be synthesized and should not be.** They are pinned to one real title's
> bytes — `PPSA24651`'s import counts, its IL2CPP module layout, the RDNA2 blobs embedded in its
> eboot.

After this PR that is **false for `trap_identifies_imports`**, and your own census is what falsifies
it: the case is green on 55/55 dumps and pins no title's bytes at all. "`PPSA24651`'s import counts"
in that sentence covered two cases; it now covers one (`module_loads_eboot`,
`tests/test_module.cpp:41`, `m.imports.size() == 612`).

This is not a nit about tidiness. `VERIFICATION.md` is the doc an agent reads when deciding whether a
dump-backed case should be title-gated or synthesized — the exact decision #2997 is about — and it now
tells that reader `trap_identifies_imports` is Messenger-pinned, which is precisely the belief that
produces the "fix" you argue against in the PR body. CLAUDE.md is explicit that a PR falsifying a
hypothesis records it **before merging**, and that a falsification living only in a PR body or an
issue comment is one the next agent re-derives at full cost. The evidence exists exactly once, in this
PR.

Fix (small, and CLAUDE.md waives the CI wait for a `.md`-only addition): move `trap_identifies_imports`
out of the "five ... pinned to one real title's bytes" group and say what it now is — dump-*backed*
(it needs an eboot) but not title-*pinned*, measured green on 55/55 (#2997).

Worth a follow-up issue rather than this PR: once it is not title-pinned, it joins `plugin_autolink`
as a candidate for the #2567 treatment — a synth PRX with a handful of imports would exercise
link → map → stub → dispatch with **no dump at all**, closing the CI gap `:173-176` names. I am not
asking for that here.

---

## Non-blocking

### N1 — `!prog.slots.empty()` has *zero independent detection power*, and T2 does not show otherwise

This is the finding on your credibility caveat, and I think you did overclaim it. The PR body says:

> T2 shows the replacement assertion is itself live and not trivially true.

T2 shows it is **live**. It does not show it is non-trivial, and it cannot: it mutates the test's own
data, so it is a same-source positive control — CLAUDE.md's "a positive control drawn from the same
source as the null it validates tests the DISCRIMINATOR, never the DOMAIN". It confirms the `CHECK`
macro evaluates the expression. That was not in doubt.

The stronger statement is that the assertion is **strictly subsumed** by the loop below it. The loop at
`test_trap_linux.cpp:48-51` (head) searches `prog.slots` for each of four libraries; if `slots` is empty
every one of those `slot >= 0` CHECKs fails. So **there is no state in which `!prog.slots.empty()` fails
and the rest of the case passes.** T2's own numbers demonstrate it: of its 5 failures, 4 are the loop's
and would have fired with the new assertion deleted entirely.

And the state it guards is one the production code documents as legitimate —
`exec_image_linux.cpp:3093`, "*a title whose every import resolved cross-module, or a dump whose dynamic
section yields none*: nothing to emit ... record the empty table and succeed."

I am **not** asking you to remove it: `"link produced no import stub slots at all"` is a better first
line in a failure report than four `no stub slot for <lib>` lines, and that diagnostic value is real.
I am asking you to describe it as what it is. Suggested rewording: *"T2 confirms the replacement CHECK
is wired and evaluated; its detection power is a subset of the four library probes below it, and it is
kept for the clearer failure message."* You asked me to read the mutation-arm claims with your two
earlier overclaims in mind — this is the one.

### N2 — the concrete answer to "is `!empty()` too weak", and a 4-line assertion that would settle it

You asked for a realistic breakage `> 500` catches and `!empty()` does not. Here is the best one I
could construct: change the pass-2 dedupe at `src/loader/linker.cpp:110` to key on `imp.lib_name`
instead of `imp.nid`. On `PPSA24651` that collapses `slots` from >500 to 35 (its distinct import-lib
count — `tests/test_module.cpp:44`). Post-PR `test_trap_linux` then reports **15 passed, 0 failed**:
non-empty, all four libs present, all four dispatch. Pre-PR it went red.

**But that class is already covered better, and hermetically.** `tests/misc/test_loader_synth_link.cpp`
pins exact slot counts — `:169` (`== 1`), `:197` (`empty()`), `:244-247` (`== 3`, plus first-encounter
ordering) — and every synthetic import is emitted as `<nid>#A#A` (`tests/fixtures/synth_prx.h:163`), so
all three imports in arm 5 share one libId and the dedupe-by-lib regression collapses 3 → 1 and reddens
that arm **in CI, with no dump**. That is strictly stronger than a threshold that only ever ran on a
developer's machine with the right title installed. Your argument survives this attack.

The honest residual: a *truncation* in the pass-2 loop (`linker.cpp:113-131`) with a cap between 33 and
178 — above `synth_prx.h`'s 32-import fixture ceiling, below the smallest real dump's 179 — would now be
invisible everywhere. Contrived: that loop is a plain range-for with no index and no cap. Non-blocking.

If you want to close it anyway, `Program` already carries the material and it is four lines and still
title-agnostic (`src/loader/linker.hpp:59`, `:70`, `:108`):

```
std::set<std::string> want;
for (auto& imp : prog.mods[0]->imports)
    if (!prog.exports.count(imp.nid)) want.insert(imp.nid);
CHECK(prog.slots.size() == want.size(), "slots=%zu but %zu distinct unresolved import NIDs",
      prog.slots.size(), want.size());
CHECK(prog.total_imports == prog.stubbed + prog.resolved_cross_module, "import census does not balance");
```

That is what the issue's "assert a property that holds for any real PS5 module" was reaching for: it
recomputes the dedupe contract from the linker's *input* with a different container, it is non-trivial
on every one of the 55 dumps, and it fires on the collapse case above. Your call — the PR is correct
without it.

### N3 — your stated reason for declining the "every slot dispatches" option is the weaker half of the real one

Your factual claim checks out. `src/hle/dispatch/dispatch.cpp:205` logs on `c == 1` (first call per
import index) and the `fprintf` at `:222` is ungated on non-Windows — only `PROSPER_UNIMPL_STACK`
gates the Windows stack extra — and `reset_call_log()` at `:195-196` zeroes `g_count`, so it is 500+
lines either way. Verified.

But log volume is the answerable objection (`freopen` stderr around the bulk loop, restore after — no
production change), so leaving it there invites the next reader to "just gate the logging" and add the
loop. The decisive reason is that the loop would buy nothing: with no HLE registered, `emit_one_stub`
falls to the unimplemented variant for **every** slot, and those stubs are byte-identical apart from a
4-byte immediate (`exec_image_linux.cpp:3013`, `mov edi, idx`) — no width hazard at high indices — while
`install_stubs` already size-checks every emitted stub at `:3106`. Invoking all N repeats one code path
N times. Record *that* in the "Not covered" section and the question is closed rather than deferred.

I judge the trade **right**, and not the easy path.

### N4 — the fixed library list re-creates, in miniature, the failure mode this PR is fixing

The comment at `test_trap_linux.cpp:43-47` argues the four names stay fixed because a future title that
lacks one "fails with `no stub slot for <lib>` — which names the cause directly". That is a **red case
on a title doing nothing wrong**, eight lines below the paragraph arguing a red case is materially
worse than a skip. Measured 0/55 and the message is self-naming, so accepting it is defensible and I
am not asking you to make the list adaptive (an adaptive list would assert nothing about *which*
libraries). But the comment reads as settled where it should carry the conditional: if it ever fires,
the answer is a skip or a smaller probe set — not editing the list to match whatever dump was in front
of you.

### N5 — nit, in the comment that is meant to be the durable record

`test_trap_linux.cpp:28` says "import tables run from 179 slots (PPSA16901) to well over 500". 179 is
the **deduped stub-slot** count, not the import-table size — `PPSA24651` has 612 imports and >500 slots
(`tests/test_module.cpp:41` vs. the old threshold). Same paragraph, two different quantities. Name it
"stub slots" so the next person does not compare it against `Module::imports.size()`.

---

## Checked and cleared

- **Your question 5 — was the title-gated skip the right fix?** No, and the evidence is stronger than
  the PR body claims. 51/55 dumps already pass all 15 checks; `prosper_add_title_gated_test`
  (`CMakeLists.txt:567`) would skip on 54. And the Messenger-count tripwire is not lost — it lives in
  `module_loads_eboot`, gated, as an exact `== 612` (`tests/test_module.cpp:41`), which is where a
  title-pinned count belongs. I tried to argue you into the skip and could not.
- Diff is one file, `+18 -1`; `git diff --check` clean; no numbered-table or `BLOG.md` obligation (no
  screenshot, no table touched).
- Attribution: `git log origin/main..e98931cbb` — one commit, `Co-Authored-By:` present, **0**
  `Claude-Session:` trailers, **0** "Generated with" lines. Your cleanup held on this branch.
- CI at this head: 9 pass, 2 skipping, 0 fail.
- No prior reviews on the PR; nothing stale to reconcile.

I used no GPU, ran no `prosper-app`/`screenshot`/`gpu_replay`/`snapshot.py`, killed nothing, and
built nothing — the 55-dump census used the already-built pre-fix binary described above.

mattias800 (commented):
# Independent re-review — PR #3296 @ `9b41482f1` (previously reviewed `e98931cbb`)

## Verdict: **REJECTED**

The **code is correct and is a real improvement** over `e98931cbb` — you took the harder option and it
works. One blocking finding, and it is **comment text only**: the comment at
`prosper/tests/misc/test_trap_linux.cpp:39-42` credits the wrong check with the coverage, and your own
mutation output disproves it. No code change needed. I will approve on sight.

**B1 from the last round is fixed and fixed well.** `docs/VERIFICATION.md:179` now reads "Four of the
six", and the `## Ruled out` entry states the distinction that finding was about — *a case that needs a
dump is not a case pinned to one title's bytes*. Cleared.

---

## Your direct question: is `total_imports == resolved_cross_module + stubbed` vacuous?

**Yes — unfalsifiable by any input; falsifiable only by a future edit.** That distinction is the whole
answer, and it is worth writing down rather than resolving to "keep" or "drop".

`src/loader/linker.cpp:114-131` is a two-exit loop body. `out.total_imports++` at `:115` dominates
both exits: the `continue` at `:120` (after `resolved_cross_module++` at `:119`) and fall-through to
`stubbed++` at `:131`. There is no third path — no `break`, no early `return`, no guard. So **no input
makes the identity fail.** It is not "weak"; as a detector of linker behaviour on real data it is
literally unfalsifiable.

It is still not worthless, and it is **not subsumed** (nothing else in the case would notice a counter
split). What it guards is the *edit* that adds a third path: a filter `continue`, an early-out placed
after `:115`, or a counter incremented in the wrong branch. That is a legitimate contract assertion.
It is just not what your comment says it is.

**You have not made the same error a third time in the code.** You have made it once more in the
comment, which is what B2 is.

---

## BLOCKING — B2: `test_trap_linux.cpp:39-42` attributes the coverage to the wrong check

```
//   * every import is accounted for as either cross-module-resolved or stubbed -- this is what
//     catches a linker that silently DROPS imports (e.g. deduplicating on the wrong key), which
//     a slot-count floor cannot distinguish from a title that simply imports less;
//   * every stubbed import actually got a slot, which is the precondition for everything below.
```

Two errors, and the first is falsified by evidence already in your own PR body.

**(a) The parenthetical is wrong.** Deduplicating on `imp.lib_name` instead of `imp.nid`
(`linker.cpp:122`/`:127`) does not break the identity and cannot: every import still increments
`total_imports` and then exactly one of the other two. On PPSA24651 the identity reads
**612 == 0 + 612 and passes** (`resolved_cross_module` is 0 — your own mutation output reports
`stubbed` = 612, and `self_dump` shows the eboot exports 1 symbol that nothing imports). Nor does that
mutation *drop* imports at all: every import still gets an `import_addr` written at `:130`; they are
misrouted onto shared stubs, not lost.

What catches it is the **third** check, `stubbed == slots.size()` — and your PR body proves it: the
failure you quote, `612 imports stubbed but 35 slots emitted`, is the format string at
`test_trap_linux.cpp:47-48`, not the one at `:44-46`. The comment credits bullet 2 with bullet 3's work.

**Why this is blocking rather than a nit:** the comment exists to tell the next reader why three
assertions are here. As written it says the accounting identity guards against wrong-key dedupe — so
an agent tidying this file could delete `stubbed == slots.size()` as redundant and remove the only
check with detection power, believing the comment. That is a mechanical harm path out of a comment
this PR is adding, in a PR whose thesis is that assertions must not claim coverage they do not have.

**(b) Bullet 3 mis-describes its own check too.** "every stubbed import actually got a slot" is true by
construction at `linker.cpp:130` and is not what the equality tests. `stubbed == slots.size()` asserts
that **no two stubbed imports share a slot** — i.e. that no NID is imported twice. That is a different
and much stronger statement, and it is a property of the corpus rather than of the linker:
`src/self/module.cpp:221` pushes one `imports` entry per undefined dynsym symbol with no dedupe, and
`linker.cpp:122` dedupes on `imp.nid` alone, discarding `lib_name` — so two imports of the same symbol
name through two libraries would share a slot and make `stubbed > slots.size()`.

**I measured that risk independently and it cuts against me, so here it is:** using
`self_dump --import-slots` on all 55 eboots, the count of distinct `(NID, library)` pairs equals the
count of distinct NIDs for **every one of the 55** — zero instances of a NID appearing under two
libraries. Your 55/55 green is confirmed by a second instrument on a different code path. So the
equality is safe today; it is simply **measured, not derived**, and the comment should say so, exactly
as the four-library list two dozen lines below already does.

### Suggested replacement (copy-paste; no code change)

```
//   * every import is accounted for as either cross-module-resolved or stubbed. This cannot fail on
//     any input: pass 2's body increments total_imports and then exactly one of the other two
//     (linker.cpp:115 / :119 / :131), with no third path. It is an EDIT guard, not a detector -- it
//     fires if a future filter or early-out counts an import without accounting for it.
//   * every stubbed import got its OWN slot. This is the one that carries weight: deduplicating on
//     the wrong key (lib instead of NID) leaves the accounting identity intact -- 612 == 0 + 612 on
//     PPSA24651 -- while collapsing 612 slots to 35, which is what a slot-count floor could not tell
//     apart from a title that simply imports less. It is an equality only because no eboot imports
//     the same NID twice: MEASURED at 0 duplicates across all 55 dumps, not derived (module.cpp:221
//     does not dedupe imports, and linker.cpp:122 keys the slot table on the NID alone).
```

---

## Non-blocking

### N2 residual — partly closed, and the shape is what decides it

Your *Not covered* section is accurate but under-specified, and the missing detail is the interesting
half. Truncation splits in two:

| truncation shape | identity | `stubbed == slots.size()` | caught? |
| --- | --- | --- | --- |
| stops the whole loop (a `break` **before** `linker.cpp:115`) | holds — all three counters stop together | holds | **no** |
| skips the tail (an early-out **after** `:115`) | **breaks** | n/a | **yes** |

So the new invariants close the second shape and not the first, and the first is the one my N2 named.
Saying which shape is open turns "the identity does nothing" into "the identity guards exactly the
edit that adds a third path", which is the accurate and more useful statement.

**Is my 4-line recomputation now redundant? No — it still adds two things**, though I am not asking for
it:

```
std::set<std::string> want;
for (auto& imp : prog.mods[0]->imports)
    if (!prog.exports.count(imp.nid)) want.insert(imp.nid);
CHECK(prog.slots.size() == want.size(), …);
```

1. it catches the first shape, because it re-derives the expected count from `mods[0]->imports`
   independently of pass 2's counters — a loop that stops early cannot hide from it;
2. it is **structurally** true rather than corpus-true: duplicate NIDs collapse in the set exactly as
   they collapse in `nid_to_slot`, so it needs no "measured 0/55" caveat and cannot red a future title
   that imports one symbol through two libraries.

Your three checks are a genuine improvement without it. If you do adopt it, it needs no further review
from me beyond confirming 55/55 still green.

### N4 — still not blocking, and I will not push

The fixed four-library list (`:53-57`) re-creating the red-not-skip shape at 0/55: below the bar,
agreed. One observation now that a second measured-not-derived property sits 15 lines above it — both
comments are making the same kind of statement, so if you touch B2 anyway, the same sentence pattern
("measured at N/55, not derived; if it ever fires the answer is a skip, not a list edit") covers both
and makes the file internally consistent.

### N5 — you already fixed it; your own report says you did not

Your message lists N5 as deliberately unfixed, but the diff shows otherwise: `:26` now reads "That
threshold was PPSA24651's **slot count**" and `:28` "**slot counts** run from 179 (PPSA16901)". That
was exactly the correction I asked for. Worth noticing, because reporting a fix as unfixed is the same
class of bookkeeping slip in the other direction.

---

## Verified at this head

- `test_trap_linux.cpp` now runs **17** CHECKs on a single-module link: 3 + `map_image` (1 image) + 
  `install_stubs` + 4 libs x 3. Matches your reported 17.
- `!prog.slots.empty()` is gone; my N1 subsumption argument no longer applies to anything in the file.
- Your mutation claim "passed the old assertion 15/15" is consistent with my previous review, where I
  constructed that same mutation and predicted `15 passed, 0 failed` against `e98931cbb`. Good faith
  confirmed both directions.
- `docs/VERIFICATION.md` at this head: `check_numbered_table.py` → *3 tables, 13 rows, unbroken, every
  row matching its header*, exit 0.
- `git diff --check e98931cbb 9b41482f1` clean. `git log origin/main..9b41482f1`: **0**
  `Claude-Session:` trailers, **0** "Generated with" lines.
- Duplicate-NID census (mine, independent): 55/55 eboots, distinct `(NID, lib)` == distinct NID.

**Merge-gate note, not a finding:** CI at `9b41482f1` is **not green yet** — 4 SUCCESS, 1 SKIPPED and
**5 checks still with an empty conclusion** (Linux is `pending`) at the time of this review. An empty
conclusion is not a pass; re-check the full pass count immediately before merging rather than trusting
this or any earlier watch.

No GPU, no `prosper-app`/`screenshot`/`gpu_replay`/`snapshot.py`, nothing killed, nothing built —
the two censuses used the already-built `test_trap_linux` and `self_dump` binaries.

mattias800 (commented):
# Independent re-review — PR #3296 @ `b5c86b281` (previously reviewed `e98931cbb`, `9b41482f1`)

## Verdict: **APPROVED**

**Nothing blocks. Merge it.** Both blocking findings are resolved, the change at this head is
comment-only as scoped, and I re-derived the parts that could have rotted rather than taking them on
report.

## B2 — fixed, and the citations it added are accurate *on current master*

The comment at `test_trap_linux.cpp:39-49` now splits the credit correctly: the accounting identity is
labelled an **edit guard that cannot fail on any input**, and `stubbed == slots.size()` is named as the
check that carries the weight, with the equality flagged **MEASURED at 0 duplicates across 55 dumps,
not derived**. That is the right description of all three checks.

Because the new comment cites five source lines, I checked every one against **current `origin/master`**
— not against the tree I read two rounds ago, and not against the branch — since a comment that ships a
stale `file:line` is the failure mode this project has paid for before:

| citation | current `origin/master` | ok |
| --- | --- | --- |
| `linker.cpp:115` | `out.total_imports++;` | ✅ |
| `linker.cpp:119` | `out.resolved_cross_module++;` | ✅ |
| `linker.cpp:122` | `auto it = nid_to_slot.find(imp.nid);` | ✅ |
| `linker.cpp:131` | `out.stubbed++;` | ✅ |
| `module.cpp:221` | `if (sym.is_import) m.imports.push_back({ sym.nid, sym.lib_name, (uint32_t)i });` | ✅ |

Five of five. None of master's five newer commits touches `loader/linker.cpp` or `self/module.cpp`, so
these hold post-merge too.

## Verified at this head

- **Comment-only, confirmed mechanically**: the `9b41482f1..b5c86b281` diff with comment lines and blank
  lines filtered out is **empty**. The three `CHECK`s are byte-identical to what I reviewed and
  approved the substance of last round.
- **CI green, checked the right way**: 11 checks, all `COMPLETED` — **9 SUCCESS, 2 SKIPPED, 0 EMPTY
  conclusions, 0 failures**. (An empty conclusion is not a pass; at `9b41482f1` five were still empty,
  which is why I flagged it then. They are all in now.)
- `MERGEABLE` / `CLEAN`. The branch is 5 commits behind master, and **none of those commits touches
  either of this PR's two files**, so the merge is non-interacting.
- `git diff --check` clean. `git log origin/master..b5c86b281`: **0** `Claude-Session:` trailers, **0**
  "Generated with" lines.
- `docs/VERIFICATION.md` unchanged since `9b41482f1`, where I cleared it.

## Two non-blocking notes — one of them is the pattern you asked me to keep flagging

### The PR-body edit you describe did not land

You wrote that you had stated the residual in the body using the shape analysis. **It is not there.**
The body's *Not covered* section still reads the earlier sentence — "a pass-2 truncation with a cap
between 33 and 178 would satisfy the accounting identity while dropping real imports" — and a search of
the whole 6,041-byte body finds no `:115`, no "break", no "edit guard". The body's `updatedAt` is
20:14:14, right after my second review, so *an* edit landed; this one did not.

Near-certain mechanism: `gh pr edit` fails here on a Projects-classic GraphQL error, **exits non-zero,
and silently does not update the body**. Use `gh api -X PATCH repos/<owner>/<repo>/pulls/3296 -f body=…`
and re-read the body to confirm.

Not blocking — the durable artifact is the code comment, and that one is now precise. But this is the
third time this session that reported state and actual state have diverged (N5 reported unfixed when
fixed; this reported fixed when unfixed), and you asked to have those surfaced rather than absorbed.
**Editing the body does not move the head, so this approval stands and needs no fourth review.**

### `355/355` is from a tree stronger than this head

#3294 (`0a625f4a4`, which added `tests/diagnostics/test_env_numeric_sites.cpp`) is one of the five
commits **not** on this branch, so a configure at exact `b5c86b281` registers 353, not 355. The 355 run
must have come from a tree with master merged in. That is *better* evidence, not worse — it is the
merge result — but it should be described as the merge result rather than as this head. The
authoritative exact-head-merged evidence is CI, and I verified that independently above.

## On whether the residual is under-sold — you asked, and my answer is no

The code comment does not under-sell it: it now states plainly that the identity cannot fail on any
input and is an edit guard. That is the honest characterisation, and a reader who wants the truncation
detail has the shape analysis in my previous review, linked from this thread. I stand by not requiring
the 4-line recomputation: it is a genuine improvement and it is out of scope for an issue about a
title-pinned constant. If you ever generalise this case to link PRXs alongside the eboot, revisit it —
`stubbed == slots.size()` is a single-module property, and two modules importing the same unresolved
NID would break the equality where the recomputation would not.

## What this PR ends up being

It started as deleting one Messenger-shaped constant. It ends as: a title-agnostic case that runs green
on 55 of 55 dumps instead of red on 4; three assertions whose comment says exactly what each one can
and cannot detect; and a `## Ruled out` entry in `VERIFICATION.md` recording that *needing a dump* is
not *being pinned to a title's bytes* — which is the finding most likely to save someone else the same
mistake. The census reproduced under my own hand at every round, and the one claim that did not survive
review is the one you replaced rather than reworded.

**APPROVED** on `b5c86b281`.

No GPU, no `prosper-app`/`screenshot`/`gpu_replay`/`snapshot.py`, nothing killed, nothing built,
nothing pushed.

Issue #2613, opened 2026-08-17
Guest threads can fault at process exit: the trampoline tail touches registries whose static destructors have already run


mattias800:
## Summary

`thread_trampoline`'s tail touches namespace-scope registries whose **static destructors have already
run** when the process is exiting. Any prosper process that reaches `exit()` while a guest thread is
still inside that tail can take a SIGSEGV during shutdown, in code that is otherwise correct.

This was found while root-causing #2427 (the macOS/Rosetta `sce_pthread_encoding` flake). That issue's
proximate cause was a test spawning a thread it never joined, and removing that row fixes the CI
signal. **The registry lifetime problem is the underlying defect and is not fixed by it.**

## Evidence — a real core dump, not a reading of the source

Reproduced on Linux by pinning the pre-fix test to one CPU (`taskset -c 0`, 1 fault in 300 runs), then
read back with `coredumpctl`:

```
Thread 2 (main):
  #0 _fini () from libstdc++.so.6
  #1 _dl_fini () at dl-fini.c:120
  #2 __run_exit_handlers () from libc.so.6
  #3 exit () from libc.so.6

Thread 1 (a guest thread):
  #0 std::__detail::_Hash_code_base<unsigned long, ... GuestThreadName ...>::_M_hash_code
       (this=0x9fcf20 <prosper::(anonymous namespace)::g_guest_thread_names>,
        __k=<error reading variable: Cannot access memory at address 0x3c7f5784>)
  ...
  #8 prosper::(anonymous namespace)::retire_guest_thread_name (thread=..., generation=1)
       at prosper/src/hle/hle_kernel.cpp:1735
  #9 prosper::(anonymous namespace)::thread_trampoline (p=...)
       at prosper/src/hle/hle_kernel.cpp:2008
  #10 start_thread () from libc.so.6
```

Main is inside `__run_exit_handlers`/`_dl_fini`; the guest thread is reading `g_guest_thread_names`.

## The affected state

Both of these are namespace-scope objects with non-trivial destructors, reached from the trampoline
tail *after* the guest entry returns:

| registry | declared | touched from the trampoline tail |
| --- | --- | --- |
| `g_guest_thread_names` + `g_guest_thread_name_mutex` | `prosper/src/hle/hle_kernel.cpp:1703-1704` | `retire_guest_thread_name`, `hle_kernel.cpp:1733` (called at `:2008`) |
[...]

PR #3306, opened 2026-09-03
tests: stop deliberate-death children from writing core dumps — ulimit -c 0 does not (#3269)

### LINKED ISSUE #3269: Concurrent ctest suites stall on coredump pileup — and the stall reads as the change under test hanging
## Summary

Running more than one ctest suite concurrently on this box **stalls**, and the stall does not look
like what it is — it looks like the change under test hanging.

Found by a lane running two of its own suites while another lane ran a third
(`replay-determinism-2945`). Reported to me as an environment note rather than a defect, which is
exactly why it is worth a number: the next agent to hit it will spend the time re-deriving it.

## Mechanism

`test_libc_alloc` and `test_raw_syscall` **fork children that deliberately crash** — that is their
job, and it is correct. With several suites live, those children pile up in `do_coredump`, and a
later stall was observed in `wait_log_commit` under roughly 45% IO pressure.

So the cost is paid by whoever happens to be running a *third* suite, and it lands as an unexplained
hang in an unrelated test.

## Why this is an instrument problem, not just a slowdown

`CLAUDE.md` already warns that a surprising measurement is more often the apparatus than the
subject. This is a clean instance: an agent verifying a change sees its own ctest run hang, and the
most available explanation — "my change deadlocks something" — is wrong. The correct explanation is
invisible unless you know another lane is running, which the harness does not surface.

## Suggested fix

Add `ulimit -c 0` to the standard ctest recipe in `CLAUDE.md`, which currently reads:

```bash
distrobox enter ps5ys -- bash -lc "cd $WT/prosper/build-linux && ctest --no-tests=error"
```

The lane reports that running each suite alone **with `ulimit -c 0`** cleared it. Disabling core
dumps costs nothing here — nobody is post-mortem-debugging these deliberate child crashes — and it
removes the pile-up at its source rather than serialising the lanes.

Worth stating in the same edit *why* the flag is there, so it is not later removed as noise.

## Not established

- Whether `ulimit -c 0` alone is sufficient, or whether the lane's success came from the combination
  with running suites one at a time. The report describes both changes together.
- The IO-pressure half (`wait_log_commit` at ~45%) may be an independent effect that `ulimit` does
  not touch.

Both are answerable with one deliberate A/B: two concurrent suites with and without the flag.

mattias800:
Fixes #3269.

## What `ulimit -c 0` actually fails to do

#3269 proposed adding `ulimit -c 0` to the standard ctest recipe, and listed under **Not established**
whether that alone is sufficient. It is not — but **not for the reason an earlier version of this PR
gave**, and the correction is the interesting part.

Measured at **20 SIGSEGVs per arm** (an earlier draft used one per arm, which is where the error came
from):

| arm | handler invoked | core files stored | wall |
| --- | --- | --- | --- |
| `ulimit -c 0` (`rlim_cur` verified 0) | **20 / 20** | **0** | 0.78 s |
| `ulimit -c unlimited` | 20 / 20 | 20 | 8.53 s |
| `prctl(PR_SET_DUMPABLE, 0)` | **0 / 20** | 0 | **0.009 s** |

`ulimit -c 0` **does** stop the core file being written — `systemd-coredump` logs *"Resource limits
disable core dumping"* and declines to save. What it does **not** stop is the handler being invoked
at all: a pipe-prefixed `core_pattern` makes the kernel spawn it regardless and leaves `RLIMIT_CORE`
for it to interpret, so the process spawn and the journal commit are paid on every crash.

**That is the #3269 stall.** So the flag is 11× better than nothing rather than worthless, and
`PR_SET_DUMPABLE` — decided in the kernel before the handler is reached — is 87× better again and the
only arm that leaves no journal entry at all.

**How the earlier draft got it wrong:** it counted `coredumpctl list` *entries* as dumps. The COREFILE
column reads `none` for suppressed ones and `present` for stored ones, and that was in my own output
the whole time — my `ulimit -c 0` arm shows `none` directly above an `unlimited` arm showing `present`
at 17.9K. A real signal read at the wrong granularity.

## Scale

`coredumpctl` had been recording this all along:

| binary | retained dumps |
| --- | --- |
| `test_libc_alloc` | **10,954** |
| `test_raw_syscall` | **1,527** |
| everything else, combined | ~243 |

**3.7 GB** in `/var/lib/systemd/coredump` and **1.1 GB** of journals. Today's `test_libc_alloc`
entries split **88 `none` / 115 `present`**, i.e. `ulimit -c 0` was in effect for roughly 43% of runs
and did stop the file in exactly those — while the handler still ran for all 203.

## The change

`prosper/tests/support/death_test.hpp` — header-only, no prosper includes — with
`prosper::test::suppress_coredump()`, called in the **child** of both tests before the deliberate
death:

- `test_raw_syscall.cpp` — in the forked child that faults a guard page.
- `test_libc_alloc.cpp` — at the top of the re-exec'd child-mode branch, covering every mode the
  parent requires to die.

Plus `tests/support/AGENTS.md`, since the folder is new. Its boundary is the point: `fixtures/` holds
things that participate in what is asserted, `shared/` mirrors `frontends/shared/`, and `support/`
participates in nothing — if a helper here could change whether an assertion passes, it belongs in
`fixtures/`.

## Deliberately NOT a tree-wide sweep

`suppress_coredump()` is called for children the suite kills on purpose, never for a parent, and only
in the two tests that produce essentially all the volume. **A test that crashes *unexpectedly* should
still leave a dump, because that one is worth opening.**

The scope is measured rather than guessed. The two fixed binaries are **12,481 of 12,724** retained
test dumps — about **98%**. The remainder are in the tens and are genuine failures:
`test_exit_registry_lifetime`'s SIGSEGV, for example, is the *observed failure* it guards against
(#2613), not a deliberate death, so suppressing it would destroy the evidence the test exists to
produce.

## Verification

A clean zero proves nothing on its own, so this is measured against a positive control drawn from
outside the fix:

| | new dumps |
| --- | --- |
| fixed `test_libc_alloc` | **0** |
| fixed `test_raw_syscall` | **0** |
| **control** — unfixed `test_libc_alloc` from another worktree, same window | **4** |

The control is what makes the zeros mean something; without it, "no dumps" is indistinguishable from
a blind detector.

```
cmake --build prosper/build-linux -j4                        # exit 0
ctest --no-tests=error --output-on-failure --timeout 900     # exit 0, 355/355
```

Both tests still pass and still assert the deaths — `libc_alloc` and `raw_syscall` exit 0.

## Follow-on

The `ulimit -c 0` advice has propagated into agent practice on this machine (it is in my own notes,
and every command I ran today carried it). It is harmless, so nothing needs ripping out — but it
should not be taught as the fix. This PR puts the suppression where it cannot be forgotten instead.


mattias800 (commented):
# Independent review — PR #3306 @ 98154030a

**REJECTED** — on the record, not on the code. The fix is correct, the scope is correct, and I
reproduced the central result independently. Two blocking findings, both prose: one clause of the
committed header comment is contradicted by my own measurement, and the falsification has no
discoverable home.

I did not duplicate the build/ctest/CI; I built only a throwaway crash harness of my own.

---

## 1. I reproduced the central claim. It holds, in the direction that matters.

Apparatus written from scratch (a ~20-line C file that optionally calls `prctl(PR_SET_DUMPABLE,0)`,
prints its own `getrlimit(RLIMIT_CORE)` so the arm is self-verifying, then stores to address 0).
**20 crashes per arm**, distinct binary names so counting is unambiguous.

`/proc/sys/kernel/core_pattern` checked by me, on the host *and* inside the `ps5ys` distrobox —
identical: `|/usr/lib/systemd/systemd-coredump %P %u %g %s %t %c %h %d %F`.

| arm (20 deliberate SIGSEGVs) | handler invoked | `coredumpctl` entries | core files stored | wall |
| `ulimit -c 0` (`rlim_cur=0` verified in-process) | **20/20** | 20 (`COREFILE: none`) | **0** | 0.78 s |
| `ulimit -c unlimited` | 20/20 | 20 (`present`) | 20 | 8.53 s |
| `prctl(PR_SET_DUMPABLE, 0)` | **0/20** | **0** | **0** | 0.009 s |

The zero is not a blind detector. **Positive control built by hand, outside what produced the
zero**: the *same binary* re-run without the `prctl` argument — only that call differs — gives 20
entries, 20 files, 8.20 s. A separately isolated `prctl` run in a strictly bounded journal window
gives 0 `systemd-coredump` lines, 0 `Started systemd-coredump@…` units, 0 entries.

The mechanism is exactly as you state it, and the journal says so in words:

```
systemd-coredump[…]: Process … (rev3306_a_ulimi) … terminated abnormally with signal 11/SEGV, processing...
systemd-coredump[…]: Resource limits disable core dumping for process … (rev3306_a_ulimi).
systemd-coredump[…]: Process … terminated abnormally without generating a coredump.
```

So `ulimit -c 0` does **not** stop the pile-up: the kernel still runs the handler, systemd still
starts a transient `systemd-coredump@…service` per crash, and the journal is still written — which
is precisely the `do_coredump` / `wait_log_commit` cost #3269 describes. `PR_SET_DUMPABLE(0)` is
checked earlier and removes all of it. **Your fix is the right one.** (`bash` agrees at the kernel
level independently of systemd: it prints `(core dumped)` — i.e. `WCOREDUMP` — for the `ulimit -c 0`
arm and not for the `prctl` arm.)

---

## 2. BLOCKING — `death_test.hpp:16` states something my measurement contradicts

> `death_test.hpp:15-16` — "with `ulimit -c` verified at 0, a crashing process still produced a
> dump — **one, exactly as with `ulimit -c unlimited`**."

It is *not* exactly as with unlimited, and the difference is the first thing a sceptic will check:

- `ulimit -c 0` → **0 core files written**, `coredumpctl`'s COREFILE column reads `none`, 0.78 s.
- `ulimit -c unlimited` → **20 core files** (18.4 K each), COREFILE `present`, 8.53 s.

`ulimit -c 0` eliminates the stored dump and ~90% of the wall cost; what it does not eliminate is
the handler invocation and the journal write. Your **very next sentence** (`:16-18`, "the kernel
invokes the handler and leaves the limit for the handler to interpret") is exactly right — it is
only the "produced a dump … exactly as with unlimited" clause that overstates.

**Failure scenario, concretely.** Six months from now an agent hits a concurrent-ctest stall, reads
this header, doubts it, and runs the obvious check: `ulimit -c 0`, crash something, look in
`/var/lib/systemd/coredump`. They find **nothing there** — the header appears falsified on its own
terms — and conclude the suppression is unnecessary and rip it out. The claim as written is
*disprovable by the cheapest available experiment* while the conclusion it supports is true. That is
the worst configuration for a durable comment, and this header is the durable record: it will
outlive the PR body.

The same clause is in the **commit message** ("`ulimit -c 0` … 1 dump / `ulimit -c unlimited` …
1 dump"), so amending is needed there too, and in the PR body's headline table — whose column header
"dumps produced" is doing the conflating.

Suggested replacement for `:15-19`, using my numbers or yours:

> Measured 2026-09-03, 20 crashes per arm. With `ulimit -c` verified at 0 the kernel still ran the
> handler on all 20 and systemd still started a transient unit and journalled each one; what the
> limit bought was only the stored file (`coredumpctl` COREFILE `none`, 0 bytes on disk) — the
> pile-up this header exists to stop was untouched. `PR_SET_DUMPABLE` is checked earlier, in
> `do_coredump` itself: 0 handler invocations, 0 entries, 0 files.

### 2b. The PR body's "Scale" section is internally inconsistent — and the data says which half is wrong

> "**Every one of today's runs used the flag that does nothing** … **3.7 GB** in
> `/var/lib/systemd/coredump`"

Those cannot both be true, because the flag is *not* nothing for disk. Measured on the live journal,
today's `test_libc_alloc` dumps:

| corefile state | count |
| --- | --- |
| `none` (never stored → `ulimit -c 0` was in effect) | **88** |
| `present` (stored → it was **not**) | **115** |

So `ulimit -c 0` covered about 43% of today's runs, and it worked *for the thing it does* in exactly
those 88. The 3.7 GB came from the 115 (and from history), not from the flag failing. Across all
time the split is starker: of 12,481 dumps from the two binaries, only **110** are `none`.

This is a small correction with a large blast radius: the PR is the artifact that will be cited when
someone later asks "does `ulimit -c 0` do anything?", and the honest answer is "yes — it stops the
file, not the handler", which is *more* persuasive for your change, not less.

---

## 3. BLOCKING (weaker; argue back if you disagree) — the falsification has no discoverable home

`CLAUDE.md` sends the next agent to `GAME_COMPAT_ORCHESTRATION.md`'s instrument list *before*
believing a surprising measurement, and #3269 is framed as an instrument problem in its own body. I
checked: that table runs to **252** and has **no** row about coredumps or concurrent-ctest stalls,
and `CLAUDE.md` itself never mentions `ulimit` (so nothing there needs correcting — good).

Your own PR body says the wrong advice "has propagated into agent practice on this machine". That
propagation is not repo-resident, so this PR cannot delete it — which is exactly why the row is
needed: an agent who hits the stall from any *other* crashing binary (`test_rdna2_to_spirv` has 62
retained dumps, `test_ajm_opus` 34) will reach for `ulimit -c 0` from their notes and it will not
work. The charter's rule is explicit that a falsification living only in a PR body is one the next
agent re-derives at full cost; `death_test.hpp`'s header is better than a PR body, but nobody
searching for a ctest stall opens it.

One table row. Allocate with `prosper/tools/docs/trap_number.py` (scans master *and* open PRs), and
run `prosper/tools/docs/check_merge_result.py` immediately before merging, since that table is
append-contended.

---

## Non-blocking

**N1 — `suppress_coredump()` also covers children that are expected to LIVE.** In
`test_libc_alloc.cpp:43` the call sits above the mode dispatch, so it covers `probe` (`:44`), which
must exit 0. In `test_raw_syscall.cpp:129` it is inside `signal_from_child`, which has three call
sites — `:303` must die, but `:312` and `:316` must exit cleanly. If any of those three ever crashes
*unexpectedly*, the dump is now gone. You document the `probe` case deliberately at `:39-42`, and
the failure is still reported by the assertion, so I am not asking for a change — but
`death_test.hpp:21` ("always before the deliberate death") reads as if only deliberate deaths are
covered, and three of the five suppressed children are not deliberate deaths.

**N2 — the strongest form of "does this lose evidence?", for the record.**
`test_libc_alloc.cpp:119-126` assert only `run_self(...) != 0`. A SIGSEGV inside prosper's allocator
would satisfy that assertion identically to the expected SIGABRT, and *before* this PR
`coredumpctl` at least recorded the signal for those children. After it, nothing does. The test's
weak assertion is pre-existing and out of scope; if you ever tighten it, assert the manner of death
(`WIFSIGNALED` / `SIGABRT`) rather than the exit code. I judged this non-blocking because the
evidence in question was one dump among 10,954 that nobody was reading.

**N3 — a courtesy note:** my measurement added ~45 core files (~800 KB) under
`/var/lib/systemd/coredump` from `rev3306_*` binaries. They are root-owned so I left them; they will
vacuum out.

---

## Checked and cleared

- **Scope — I tried hard to break it and could not.** My best candidate was
  `test_game_compute.cpp:128-143`: it forks a child that stores to a `PROT_NONE` page at `:137`,
  which is the same class as your `test_raw_syscall` child, and it has 23 retained dumps. It is
  **correctly excluded**: `write_watch_segv_handler` (`:79-100`) catches the fault and `_exit(86)`s
  at `:99`, so the child never dies by signal and produces no dump at all. I checked this before
  writing it up precisely because I wanted it to be a finding.
  `test_boot_linux.cpp:85` SIGKILLs its child (SIGKILL never dumps).
  `test_dmem_caller_walk.cpp:72-86` expects exit 0 — a signal there *is* the #1755 crash, so leaving
  it dumping is right. `test_rdna2_to_spirv` (62), `test_ajm_opus` (34) and
  `test_exit_registry_lifetime` (21) are all `sig=11` with no deliberate-death construct in their
  sources — genuine, as you say.
- **The census reproduces exactly.** 12,724 `test_*` dumps, 12,481 from the two binaries =
  **98.09%**. Your 12,481 / 12,724 / ~243 are all correct. (Total `coredumpctl` entries are 13,042;
  the extra ~300 are non-test binaries — `prosper-app`, `maliit-keyboard`, `dotnet` — so anyone
  re-deriving your 98% must scope to `test_*`, as you did.)
- **`PR_SET_DUMPABLE(0)` side effects — clean here.** It restricts `ptrace` attach and `/proc/[pid]`
  ownership, but neither child touches `/proc` or is ptraced after the call, and self-introspection
  is exempt from the dumpable check anyway (`__ptrace_may_access` returns 0 for `same_thread_group`),
  so even a `/proc/self/maps` read would still work. Both children die within microseconds.
- **No leak to the parent, and the placement is *required*, not merely convenient.** I verified
  empirically that `execve` resets dumpable: a process that sets it to 0 and then `execl`s a probe
  gets `dumpable=1` back in the exec'd image. So for `test_libc_alloc` the call **must** be after
  the re-exec, which is where you put it. `test_raw_syscall`'s child gets its own `mm` from `fork`.
  And the parent can never take the branch: `add_test(NAME libc_alloc COMMAND test_libc_alloc)`
  passes no argv, so `argc == 1`.
- **Header honesty / platform guards.** `#include <sys/prctl.h>` is `__linux__`-guarded correctly;
  `test_raw_syscall` is behind `if(UNIX AND NOT APPLE)` (`CMakeLists.txt:829`) so only
  `test_libc_alloc` compiles the non-Linux branch, which is a bare `return false`. `CMAKE_CXX_STANDARD 20`,
  so `namespace prosper::test` is fine. The `bool` return is ignored at both call sites, which is
  correct: it is not `[[nodiscard]]`, and a `prctl` failure degrades to exactly the pre-PR behaviour.
  Returning `false` rather than faking success is the right call.
- **Include path.** `target_include_directories(<t> PRIVATE tests)` is pre-existing practice on 20+
  targets, not something this PR introduces, and there is currently no relative-path collision
  between any header under `tests/` and one under `src/`. Not a finding.
- **`tests/support/AGENTS.md`** — its factual claims check out: `tests/shared/` and
  `frontends/shared/` both exist, `render_runner.h` and the `gta5_*` fixtures are where it says. The
  boundary it draws ("`support/` participates in nothing") is a good one and the folder earns the file.
- **Attribution.** 0 `Claude-Session:` trailers, 0 🤖 badges, in the commit and the PR body. Clean.

---

**Verdict: REJECTED.** Fix `death_test.hpp:15-19` (and the commit message and PR body table) so the
recorded mechanism survives the cheapest check someone will run against it, and add the instrument
row. No code change is required — I found nothing wrong with the implementation, and I would
approve it on the corrected head.

mattias800 (commented):
# Re-review — PR #3306 @ 8bea8341d (previously reviewed 98154030a)

**APPROVED.**

Both blocking findings are fixed. I re-derived the corrected table against my own measurements rather
than reading it, and I went looking specifically for the opposite-direction over-claim you asked me
to bound. **I found exactly one, and it is not the one you were worried about** — the multipliers are
fine; a smaller claim next to them is not.

---

## The three-arm table matches what I measured, digit for digit

`death_test.hpp:18-21` against my run (20 SIGSEGVs per arm, my own harness, distinct binaries):

| arm | your table | what I measured |
| --- | --- | --- |
| `ulimit -c 0` | 20/20 invoked, 0 files, 0.78 s | 20/20, 0 files (COREFILE `none`), 0.776 s |
| `ulimit -c unlimited` | 20/20, 20 files, 8.53 s | 20/20, 20 files (`present`, 18.4 K), 8.527 s |
| `prctl(PR_SET_DUMPABLE, 0)` | 0/20, 0 files, 0.009 s | 0/20, 0 files, 0.0091 s |

No drift. The prose at `:23-28` is now correct on both halves — the flag stops the file, not the
handler — and `:33-35` names the trap the way I hit it. The PR body table, the Scale paragraph
(88 `none` / 115 `present`, handler for all 203) and trap row 253 are all mutually consistent and
consistent with my data. Your independent confirmation from your own `crashme` arms (`none` vs
`present at 17.9K`, one line apart) is the same distinction I measured on a different binary.

**Docs gate — I ran it rather than taking it.** `check_numbered_table.py --ordered --table-header
Instrument --baseline origin/main`: *"14 table(s), 315 rows, unbroken … numbered column 1..253, 253
rows, unique and ascending, no unused numbers; none of the 252 rows in the base has been deleted."*
Exit 0. "No unused numbers" also independently confirms 247 was taken, so your `trap_number.py`
diagnosis holds. Re-run `check_merge_result.py` on your own branch immediately before merging — my
`--baseline` run is the same check but goes stale the moment another lane appends.

---

## Non-blocking, but please fix — one claim in the opposite direction, and it is disprovable the cheap way

> `death_test.hpp:31` — "87x faster again, and **the only arm that leaves no journal entry**."
> PR body — "the only arm that leaves **no journal entry at all**."

Not true. Under `PR_SET_DUMPABLE(0)` every crash still writes **two** journal lines, from the kernel
rather than from systemd:

```
audit[2569775]: ANOM_ABEND … comm="rr_prctl" exe="…/rr_prctl" sig=11 res=1
kernel: rr_prctl[2569775]: segfault at 0 ip 00000000004005d3 sp … error 6 …
```

Isolated bounded window, 6 crashes → **12 journal lines, 0 `coredumpctl` entries**. Those lines are
per-signal kernel logging and are unaffected by the dumpable flag; I see them on all three arms.

The true and equally strong statement is **"the only arm that leaves no `coredumpctl` entry"** (or
"no `systemd-coredump` entry"). The substance is untouched — two short kernel-written lines against a
process spawn plus a transient unit plus a structured entry with backtrace is not a close call — but
the claim as written is falsified by `journalctl | grep <binary>`, which is the same cheap-check
shape the file warns about eight lines earlier at `:33-35`. Given this header is now the record, I
would not leave that one standing. Trap row 253 does **not** contain the claim, so it is two edits.

## The multipliers are fine — and if anything you under-sold the flag

You asked me to bound the swing from "does nothing" to "11x better than nothing". I re-measured with
three interleaved repetitions per arm, 20 crashes each:

| rep | `ulimit -c 0` | unlimited | prctl | unlim/u0 | u0/prctl |
| 1 | 0.915 s | **57.589 s** | 0.0074 s | 62.9x | 124.1x |
| 2 | 0.810 s | 12.940 s | 0.0094 s | 16.0x | 86.5x |
| 3 | 0.853 s | 14.187 s | 0.0089 s | 16.6x | 95.7x |

Across all seven measurements I have taken, the `unlimited` arm ranges **3.96 – 57.6 s** while
`ulimit -c 0` is stable at **0.78 – 0.92 s**. So `unlim/u0` spans roughly **6.6x – 63x**, and your
**11x sits near the bottom of that range** — the opposite of cherry-picking. `u0/prctl` spans
**86x – 124x**, and your **87x is the floor**. Neither is inflated; I could not make either look
generous.

The only thing I would change is presenting them as constants. Someone re-measuring will get 16x or
63x and think the header is wrong. "roughly an order of magnitude, and highly variable under load"
costs nothing and cannot be contradicted.

**That variance is itself corroboration for #3269.** The 57.6 s outlier — same 20 crashes, same
binary, 0.9 s away under the flag — is the handler pile-up behaving exactly as the issue describes.
It is a better argument for the change than the 8.53 s figure, if you want one.

## One thing I tried to break and could not

`:25` says a pipe-prefixed `core_pattern` makes the kernel spawn the handler "regardless". I went
after that with `ulimit -c 1`, on a half-memory that the kernel treats `RLIMIT_CORE == 1` as a
recursion marker and aborts the pipe dump. **No counter-example**: `ulimit -c 1` invoked the handler
like the others. "Regardless" stands as written; I am recording the attempt so nobody re-runs it.

## Your two questions on the earlier non-blocking notes

- **Covering the three children that are expected to live** — agreed, leave it. "They cannot dump if
  they do not die" is the right argument and a conditional would be more code for no behaviour. I
  would only soften `:37` ("always before the deliberate death"), which reads as though only
  deliberate deaths are covered when three of the five suppressed children are not. One word.
- **The `test_libc_alloc:119-126` disambiguation loss — yes, record it in the file.** That is the
  one cost of this change that is invisible at the call site: `coredumpctl` no longer records the
  signal for those children, so a SIGSEGV inside prosper's allocator and the expected SIGABRT now
  look identical to anyone debugging a red `libc_alloc`, and both satisfy the `!= 0` assertion. One
  sentence in the header next to "a test process that crashes UNEXPECTEDLY should still leave a dump"
  is the natural home — that is precisely where a reader will be standing when it bites them.

---

## Re-checked from the first round, unchanged

Code is identical to the head I already cleared: scope (I could not break it —
`test_game_compute.cpp:99` `_exit(86)`s so its `PROT_NONE` child never dumps; `test_boot_linux.cpp:85`
SIGKILLs; `test_dmem_caller_walk` expects exit 0), the 98.09% census, `PR_SET_DUMPABLE` side effects,
the empirically-verified fact that `execve` resets dumpable (so the placement after the re-exec is
required, not convenient), platform guards, the honest `false` return, and attribution — still 0
`Claude-Session:` trailers and 0 badges on `8bea8341d`.

**APPROVED.** Neither remaining item blocks; both are one-line prose edits I would make before
merging, and I do not need to see them again.

PR #2250, opened 2026-08-07
diag(render): partition build_resources exhaustively, with a signed residual at both levels

mattias800:
diag(render): partition build_resources exhaustively, with a signed residual at both levels

`build_resources_ms` is the frontend materializer and the second-largest term in a
Blue Prince gameplay frame -- **48.95 ms/submit, 26.7% of the frame** (#2215).
Its only named leaves were `texture_ms` and `buffer_ms`, which cover 10.2 ms of
it, and nothing asserted they were exhaustive. **38.76 ms -- 21.1% of the whole
frame -- was attributable to nothing.**

That is not the same as being spent on nothing. It is indistinguishable from it
in a report, which is the problem: an unmeasured region does not read as
unmeasured, it reads as someone else's number. #2246 is what that costs -- a
diagnostic placed inside an unpartitioned bucket billed 305 ms to the renderer
and stood as a finding for two hours.

What is recorded now
--------------------
Four per-draw leaves inside `build_bds`, plus a SIGNED residual:

    build_resources  = build_R + validate + indices + poison + other

and `build_R` publishes its own residual too, because a nested partition that
reports only the outer remainder just moves the blind spot one level down:

    build_R          = texture + buffer + other

Measured on `PPSA25009`, Windows/MinGW, live:

    build_resources 6.12 ms/submit [build_R=5.64 (texture=3.40 buffer=0.53
      other=+1.71) validate=0.29 indices=0.14 poison=0.01 other=+0.05]
      draws=102.6 rejected=0.0

Both levels close: 5.64 + 0.29 + 0.14 + 0.01 = 6.08 against 6.12, residual
+0.05 (0.8%). The inner one is what makes the point -- `texture + buffer` is
3.93 of `build_R`'s 5.64, so **+1.71 ms, 29% of build_R, had no name** and now
prints as its own term rather than being absorbed silently.

Three things this gets right on purpose
---------------------------------------
* **texture/buffer are NOT leaves of the outer partition.** They are measured
  INSIDE `build_R`, so they are a sub-split of it and are printed in parentheses.
  Adding them to the outer sum would double-count -- the cross-layer subtraction
  #2243 fixed, one level down. There is a comment saying so, because the field
  names invite exactly that mistake.
* **A rejected draw is accounted before the `continue`.** It has already spent
  `build_R` and both validations. Dropping it unaccounted would move that time
  into the residual, where it reads as unattributed work -- the defect this
  partition exists to make visible. The validation is now `&&` of the positives
[...]

PR #173, opened 2026-07-09
fix(gpu): s_bfe_u64 const-fold sign-extends negative inline constants

mattias800:
## Summary
- Per the RDNA2 ISA, a **64-bit scalar operand** reads an integer inline constant **sign-extended** (-1..-16 are all-ones in the high dword) and an inline float constant as a **64-bit double** — only a 32-bit literal is zero-extended. The vertex-fetch const-fold (`resolve_dynamic_fetch`) treated every non-SGPR src0 as hi-dword 0, so an `s_bfe_u64` field reaching bits ≥ 32 of a negative inline constant folded to the wrong value, recorded it as KNOWN, and flowed into the s_cselect tail and the emitted vertex descriptor (a confidently-wrong V#).
- Fix: sign-extend `InlineInt` by its sign; keep hi=0 only for 32-bit literals. Inline **floats** already stay unknown (`srcval` rejects them) — correct, since the fold doesn't model the double encoding; the new test locks that in.
- Exposes `DynFetch` + `resolve_dynamic_fetch` in `gpu_execute.hpp` (previously anonymous-namespace) so the fold's scalar-ALU semantics are unit-testable.

## Verification
- New `test_dynfetch_fold` (no Vulkan needed) drives the three 64-bit operand shapes through a synthetic fetch kernel — encodings **llvm-mc gfx1030 round-trip verified**:
  - inline `-1`: bits[47:40] fold to `0xFF` (sign-extension), observed in the resolved V#
  - 32-bit literal: bits[39:32] fold to 0 (zero-extension)
  - inline float `1.0`: does NOT fold — the fetch stays unresolved (nothing fabricated)
- **Confirmed the new test FAILs on the unfixed logic** (`[FAIL] s_bfe_u64 of inline -1 sign-extends`).
- Full suite in WSL build-linux: **54/54 passed**.

Fixes #155

PR #3309, opened 2026-09-03
perf(renderer): the frontend texture materializer's time is the allocator, not the decode

mattias800:
## What this is

Stray's title screen renders at **9.76 fps**, and the largest single leaf in an F8 `.prperf`
taken there is the **frontend texture materializer at 1110.1 ms of a 5020 ms window** — 22% of
wall time — against `gpu-device` at 232.5 ms. This is host-bound work in the graphics frontend.

Instrument first, then the fix it found, then the corrections two rounds of measurement and one independent review turned up.

## Where the time actually is

The renderer has recorded the texture leaf's own cache-outcome classes since #2250, and
`performance_capture_report.py` printed only the leaf total — so every capture taken since has
carried the answer and no reader could see it. Printed:

```
build_resources (frontend materializer): 1149.9ms  [texture=1110.1 buffer=17.9]
  texture by cache outcome: rtt=77.3 compute=59.9 local=17.1 persist_hit=25.0
                            persist_reuse=0.0 persist_miss=0.0 other=+930.8
  slowest unclassified reference: 59.6ms addr=0x30d10f0000 3840x2160x1 fmt=4/4c
                                  tile=27 src=63.8MiB compute_cand=1 persist_cand=1 dcc=1
```

**84% of the leaf is one unnamed class**, and the witness names it: a **3840×2160 RGBA16F**
surface, 63.8 MiB of tiled source, DCC on, at one of exactly **two alternating addresses**
(`0x30784e0000` / `0x30d10f0000`) — a double-buffered 4K HDR intermediate, re-decoded on every
callback that samples it.

## What the decode was actually spending

Each reference built two fresh value-initialised `std::vector<uint8_t>` intermediates —
66,846,720 tiled bytes and 66,355,200 linear. Both are past glibc's 32 MiB dynamic mmap
threshold, so each is an `mmap`, ~16,000 page faults as the value-init touches it, a memset, and
a `munmap`. Measured standalone at that exact shape:

| | ms |
| --- | --- |
| fresh `traw`+`hlin`, copy + detile-equivalent traffic | 26.00 |
| **the two fresh zero-initialised allocations ALONE** | **17.84** |
| pooled `traw`+`hlin`, same copy and traffic | 6.00 |
| pooled linear, tiled source read in place | 2.83 |

Two thirds of it was the allocator and the kernel, not the decode. This is the same cost #3149
saw from outside as `__memmove_avx512` at 14.5% self plus `clear_highpages_kasan_tagged` /
`map_anon_folio_pte_nopf` / `zap_present_ptes` / `__free_one_page` under a 22.2% `do_syscall_64`
— and attributed to the compute path, because that is where it went looking.

## The change

Applied to the BC, fp32, fp16, narrow and generic-32bpp decode branches:

* the linear destination comes from a thread-local pool that keeps its pages mapped between
  references (`frontends/shared/live/decode_scratch.hpp`);
* the tiled source is read where it already lives — guest memory is 1:1 mapped, so the staging
  copy only ever produced a second address for the same bytes.

`safe_copy` becomes `safe_span` plus one memcpy, so "how much of this range is readable" has
exactly one implementation and the direct path cannot drift from the copy it replaces. A short or
sparse range returns `nullptr` and keeps the staged path, so every `got < …` fallback below keeps
its meaning.

## The invariant this rests on, and how it is held

A pooled buffer arrives holding the **previous surface**, where a value-initialised one arrived
zeroed — and several branches read bytes they did not write.

* `Lease::zero_tail(written)` restores exactly what value-initialisation gave, and is called at
  every partial fill.
* `detile_writes_whole_destination(tile_mode, bpe)` says when a detiler covers its destination
  completely, so the pre-zero is skipped only there. It lives in `tile.cpp` beside the copiers it
  describes, not in a caller's comment. Every other arm (volume, mip-tail, linear-padded)
  pre-zeroes.

## Tests

**`decode_scratch` (new).** The pool's contract, asserted **differentially**: the same partial
fill applied to a reused DIRTY lease and to a fresh `std::vector<uint8_t>(n, 0)` must compare
equal, over five fill fractions. It dirties the pool with a larger surface first on purpose — a
pristine pool hands out zeroed pages and would pass whatever the tail contract said. Plus
retention, no-shrink, nesting, move, both budget bounds, and the knob's strict parsing.

**`tile` (extended).** The destination-coverage claim, asserted by **poison**: the destination is
filled with `0xDB`, a byte excluded from the source, and no survivor is permitted — six tile
modes × five element sizes × five extents including a ragged one, **through both
`detile_surface` and `detile_elements`**, because the predicate in `tile.hpp` answers for both and
the BC branch of the materializer uses the second one (review finding N1). Its mutation arm
asserts the predicate can also say NO, or the sweep would be vacuous.

**`performance_capture` (extended).** The new record fields reach the artifact.

## Levers — three, and setting all three restores the pre-PR behaviour

* `PROSPER_NO_DIRECT_TEXTURE_SOURCE` — restore the staging copy, on **both** source kinds. A
  lever that disabled only the guest-memory half would still be named after the whole
  optimisation, and an A/B under it would misattribute the other half.
* `PROSPER_DECODE_SCRATCH_MB=0` — retain nothing, which is the pre-pool allocation pattern
  exactly. **Spelled `0` and nothing else**: a malformed value keeps the 512 MiB default, so
  pooling stays ON. See the correction below — this body previously claimed the opposite.
* `PROSPER_NO_TEXTURE_PREFIX_INHERIT` — restore the `assign()` spelling. Added *because* the
  first A/B proved it was missing; see the correction section below.

All three together restore the pre-PR behaviour of all four perf commits. **Paste this rather
than typing it** — the one numeric lever fails by leaving the pool armed, and a typo happens when
somebody works from memory:

```
PROSPER_NO_DIRECT_TEXTURE_SOURCE=1 PROSPER_DECODE_SCRATCH_MB=0 PROSPER_NO_TEXTURE_PREFIX_INHERIT=1
```

## One more of the same family, in the last commit

A persistent cache entry whose bytes changed is erased and rebuilt, and both entries hold a
`source_prefix` of the same shape — so the old buffer was freed and an identical-sized one
allocated for the same bytes, every invalidation. **13.00 ms** at this shape against 3.5 ms for
the copy it exists to hold. The rebuild now inherits the outgoing entry's allocation.

The move happens **after** `persistent_decoded_texture_bytes` is debited by the old entry's
`bytes()`, which reads `source_prefix.size()` — moving first would debit zero and leak that
entry's bytes from the residency budget permanently. And it `resize`s before the copy, because
`source_prefix.size()` is compared against `persistent_source_size` on the validation path, so
growing it to the inherited extent would silently change which entries revalidate.

This is the commit that shipped without an off-switch and gets one here
(`PROSPER_NO_TEXTURE_PREFIX_INHERIT`) — see the correction section.

## Not addressed, recorded with its size

`cached.pixels = std::move(texture_pixels)` steals the pooled `texstore` slot, so the next decode
allocates its 33 MiB output buffer fresh (~4.4 ms). Fixing it needs a decision about buffer
ownership between `texstore` and the persistent cache, which hands its buffer out as a
`shared_ptr<const std::vector>` that never comes back. Recorded in `docs/STRAY_STATUS.md`.

## The lever this work does NOT reach, and it is the big one

The witness line says `compute_cand=1`. That surface is a **compute image candidate whose import
misses every frame** — `import_live_compute_storage_image` finds nothing under its key, so
graphics falls through to the guest-byte decode. If it hit, the 63.8 MiB writeback → detile →
convert disappears on **both** sides of the boundary: the graphics decode here, and the compute
writeback that feeds it. That is worth more than every allocation in this PR combined, and it
lives in the compute image cache's admission rather than in the materializer.

## Measured, live — the prediction was 660 ms and the answer is 497.7

Three arms on one binary, same route and same F8 window as the baseline capture.

| | baseline `fc21d46ca` | **this branch** | lever arm |
| --- | --- | --- | --- |
| rendered | 9.76 /s | **11.55 /s** (+18%) | 10.15 /s |
| `build_resources` texture | 1110.1 ms | **497.7 ms** (−55%) | 918.2 ms |
| ...`persist_invalid` | (no bucket yet) | 221.3 | 631.9 |
| ...unclassified `other` | +930.8 | +144.3 | — |
| renderer-resource | 1895.5 ms | 1384.1 ms (−27%) | — |
| compute | 1981.5 ms | 2351.4 ms | — |
| gpu-device | 232.5 ms | 263.3 ms | — |

The 63.8 MiB RGBA16F witness is **gone from the top**: the slowest unclassified reference is now
7.3 ms on a tiny surface (`0x30a0ac0000`, `fmt=7/1c`, `persist_cand=0`).

**Compute rising 1981.5 → 2351.4 is not a regression, and should not be read as one.** The window
is fixed at 5.02 s, so a pipeline rendering 18% more frames issues more dispatches into it — the
bucket is throughput, not cost. The discriminating pair is mean-ms-per-dispatch and dispatch
count, both of which the report already prints per program; nobody should quote the bucket delta
without them.

### The falsification check I asked for, answered honestly

I said: if `other` stays large once `persist_invalid` appears, the derivation is wrong, and that
would be the more useful result. What happened is neither clean outcome. `other` fell
**930.8 → 144.3** and `persist_invalid` is now the **largest single class** — so the derivation
holds in direction and rank. But it is **221.3, not the ~900 I predicted**, because the same
change that gave the residual a name also removed most of the cost the name was going to hold.

**So: confirmed about WHICH state, not about how much, and the ~900 figure is withdrawn.** A
prediction whose own fix invalidates its arithmetic has not been tested at the magnitude it
stated, and I would not quote it.

### 1. `PROSPER_DECODE_SCRATCH_MB` — a typo leaves the pool ON, not off (review finding B1)

This body said "a mistyped value selects that same safe sentinel (#3253)", and
`decode_scratch.hpp` said the same. **Both were false**, and the PR's own new tests already
asserted the truth — `test_decode_scratch.cpp` and the new `test_env_numeric_sites` census row
both say a malformed value keeps **512**, i.e. pooling fully on.

It was doubled by a second mistake: `"0 = retain nothing"` was passed as
`env_u64_or_default_capped`'s `default_note`, which glosses **the default**, so the refusal
printed `keeping the default (512, 0 = retain nothing)` — a line that reads as "512 retains
nothing". The two other sites in the tree that pass a `default_note` both have a default of `0`,
which is why the parameter reads correctly there and not here.

The cost is specific: this knob's main use is *disarming* the pool for the A/B this body
prescribes, so the dangerous typo is one meant to turn the optimisation **off** — `=0mb`, `=0MB`,
`= 0` — and it leaves the arm fully armed while stderr says otherwise. Fixed: the note now
describes 512, both prose claims are corrected, and `test_decode_scratch` now pins the near-miss
spellings an operator actually types (`0mb`, `0MB`, ` 0`) so the false claim cannot be restated
without deleting a passing assertion.

The code behaviour was already right — refuse loudly, keep the default — so nothing about the
decode changed. What changed is that the message and the documentation now agree with it.

### 2. The levers did not cover everything the PR does

This body previously said both levers "default to today's behaviour when set". **That was wrong
for one of the four perf commits.** The `source_prefix` inheritance had no off-switch at all, so
it was necessarily still active in the lever arm — which is why that arm returns the leaf to
**918.2 and not 1110.1**.

Being careful about what that licenses: *some* of the 191.9 ms gap is that commit, by construction,
because it was switched on. That all of it is, is **not** established — the standalone figure
predicts ~133 ms (13.00 ms per invalidation over ~14 references), the arms are separate runs, and
`setup_resources` buffer copy moved 512.6 → 596 between them, so variance here is not zero.

A discriminator that cannot disable everything the change does is not a discriminator; it silently
under-reports what it is attributing. `PROSPER_NO_TEXTURE_PREFIX_INHERIT` closes the gap, so all
four perf commits are now switchable off together. **The 918.2 arm therefore UNDER-reports this
PR**, and a re-run with all three levers is the arm worth quoting.

## Pixel identity: 5 of 6, and why that is a phase shift rather than a difference

Six frames per arm at the same ordinals, SHA-256 over raw RGB:

```
ON : f0=5319c5a4  f1=9f0a5110  f2..f5=9f0a5110
OFF: f0=5319c5a4  f1=5319c5a4  f2..f5=9f0a5110
```

**Both arms produce exactly the same two distinct images, and every image the ON arm produced is
byte-identical to one the OFF arm also produced.** That is the load-bearing fact. Stale-buffer
corruption — the failure mode this PR's whole zero-tail/coverage argument exists to prevent —
would introduce a *third* hash. It cannot produce a bit-exact match with the settled frame the
other arm independently reaches.

The one specific non-benign mechanism worth naming, because it is the only one consistent with
the data: removing the staging copy means the detiler now reads guest memory across the detile's
duration instead of snapshotting it with a memcpy first, so a concurrent guest write could be
seen. If that were happening the result would be a **torn** frame — again a third hash. It is
not. What remains is that the faster arm reaches the settled state one ordinal earlier, which is
what a rate change does, and is the same latitude the hardware has (it samples at execution, not
at submit).

**I am not calling this "identical", and it does not close the question.** Two matching steady
states prove the steady state; they do not prove every transient, and an ordinal anchor cannot
align two arms with different frame rates — that is what failed here. Two follow-ups would, and
neither needs a new build:

1. **Deterministic replay.** One F9 `.prgbundle`, replayed offline through `gpu_replay` with and
   without `PROSPER_NO_DIRECT_TEXTURE_SOURCE`. Same commands, same resources, no guest, no
   scheduling — byte equality there has no phase confound at all. **Its limit, stated up front:**
   replay disables the persistent decode cache by construction (`texture_decode_cache_candidate`
   requires `!has_captured_host_data`) and its resources are `host_data`-backed, so it exercises
   the pooling, the zero-tail contract, the coverage predicate and the *host_data* direct read —
   **not** the guest-memory `safe_span` read, and not the prefix inheritance.
2. **A phase-immune live comparison** for what replay cannot reach: over a long settled window
   (~60 frames per arm), compare the *set* of distinct frame hashes rather than frame-by-frame. If
   neither arm produces a hash the other never produces, that is strictly stronger than six
   aligned frames and immune to the rate difference. The present data already satisfies it at
   n=6; at n=60 it would mean something.

## Verification

Isolated worktree (`.claude/worktrees/stray-perf`) against `origin/main` `fc21d46ca`, built in the
`ps5ys` distrobox.

```
cmake --build prosper/build-linux -j4      -> exit 0, 0 compile errors, 0 targets stale on re-run
ctest --no-tests=error -j2                 -> exit 0, 100% tests passed out of 357, 0 skipped
```

The affected cases by name, all `Passed`:

| ctest case | what it covers |
| --- | --- |
| `decode_scratch` (new, #145) | the pool's differential equivalence contract |
| `gpu_surface_detile` (#155) and `gpu_surface_detile_threaded` (#156) | the poison sweep, on both the scalar and threaded detile paths |
| `performance_capture` (#126) | the new record fields reach the artifact |
| `diagnostics_env_numeric_sites` (#173), `env_numeric_arms` (#32) | the new knob's arm and the gate that requires it |
| `game_compute_exec` (#313), `gpu_capture_render_replay` (#307) | the converted decode paths under the production backend |

The `PROSPER_NO_TEXTURE_PREFIX_INHERIT` commit and the review fixes landed after that run; the
suite was rebuilt and re-run on each, and the results are posted as comments below.

**Non-blocking review findings taken as well:** N1 (the sweep now drives both detile entry
points), N3 (`frontend_tex_other_n` was serialized and never printed — the report now prints it
beside the millisecond residual and shouts when the two disagree, which is the cross-check the
field's own comment promised), N4 (`Lease::valid()` had no caller — removed), N5 (the
`intermediates=` census line is one thread's pool and now says so), N6 (`take(0)`'s `data()` is
documented as unspecified). N2 is already named in this body.

Also run clean locally: `check_numbered_table` (all tracked `.md`), `check_trap_citations`,
`check_env_numeric_arms` (31 knobs, all armed), `check_cached_env`, `check_diag_gates`,
`check_ctest_gate`, `check_ascii_output`, `check_contribution_shape --selftest`,
`gen_progress_tracker --selftest`, and `git diff --check`.

### One thing the suite caught, worth recording

The first `ctest` was **red**, on `diagnostics_env_numeric_sites`, and the failure was mine and
instructive: my arm's malformed input was `"512mb"`, which the *legacy* `strtoull` spelling parses
to 512 — this knob's own default. So the arm's second leg ("...and the old spelling did NOT keep
the default") could not hold, and the arm proved nothing about the conversion. `"64mb"` parses to
64 under the old spelling and keeps 512 under the new one, which is the discrimination the arm
exists to make. An arm whose malformed value collides with its own fallback is vacuous; that
suite is built to say so, and it did.

### Mutation arms, by exit code

| mutation | expected | observed |
| --- | --- | --- |
| `Lease::zero_tail` → no-op | `test_decode_scratch` fails | **exit 1**, 5 checks, 4 of them the differential equivalence assertion at `test_decode_scratch.cpp:76` |
| `sw64kb_copy<false>` stops one block row short | `test_tile` fails | **exit 1**, 61 checks, **39 of them the new poison sweep**, e.g. `detile_surface left 491520 poison bytes for mode=9 bpe=1 3840x2176 while claiming whole-destination coverage` |
| drop `frontend_tex_persist_invalid_ms` from the serializer | `test_performance_capture` fails | **exit 1**, 1 failure, the field-presence assertion |
| truncate **`detile_elements` only**, leaving `detile_surface` untouched | only the new second sweep arm fails | **exit 1**, **72 `detile_elements` failures and 0 `detile_surface` failures** — the added arm is not shadowed by the one that already existed |

Each mutation was reverted and the file confirmed clean against `git diff` before continuing.

### The live A/B this cannot cover

The direct-read path's equivalence is by construction — `safe_copy` is now `safe_span` plus one
memcpy, and the direct pointer is handed out only when `safe_span(addr, n) == n`. What no
host-side test can show is that a real title's decoded pixels are unchanged. The lever exists:
`PROSPER_NO_DIRECT_TEXTURE_SOURCE=1` restores the staging copy, so an A/B on any guarded title's
snapshot answers it in two runs.


mattias800:
# The live measurement I want, in the order it should be run

All three arms are the SAME binary from `perf/stray-texture-materializer`, so nothing between them
is a rebuild. Same route as the 23:28 capture: `reach-title-hold.pad`, `PROSPER_NULL_PAGE=1`, F8 at
t = 175 s with the title screen up.

## 1. The measurement arm (default settings on the branch)

```
PROSPER_PERF_CAPTURE_AFTER_MS=175000
```
Read with `python3 prosper/tools/perf/performance_capture_report.py <capture>` — the report now
prints the frontend texture leaf's classes directly. Three numbers matter:

| line | baseline (fc21d46ca) | predicted |
| --- | --- | --- |
| `build_resources … [texture=` | 1110.1 | ~660 |
| `texture by cache outcome: … other=` | +930.8 | **≈ 0** |
| `… persist_invalid=` | UNAVAILABLE | ~900 |

The second and third together are the important pair: they test the derivation in
`STRAY_STATUS.md` that the residual is `resource_persistent_invalidation`. If `other` stays large
after `persist_invalid` appears, that derivation is **wrong** and the residual is something else —
which is a more useful result than the speedup.

## 2. The discriminator arm (same binary, levers off)

```
PROSPER_PERF_CAPTURE_AFTER_MS=175000 PROSPER_NO_DIRECT_TEXTURE_SOURCE=1 PROSPER_DECODE_SCRATCH_MB=0
```

This must reproduce the baseline's `texture=1110`-ish within noise. If it does not, arm 1's
improvement is not attributable to this change and neither number should be quoted. A same-binary
A/B is the strongest form available here, and it costs one extra run.

## 3. The correctness arm (this is the one my host-side tests cannot cover)

Two `tools/screenshot` runs on the same route, same frame ordinal, differing only by
`PROSPER_NO_DIRECT_TEXTURE_SOURCE=1`. The images must be **identical**, not merely similar: the
direct path hands the detiler the same bytes the staging copy would have contained, so any pixel
difference is a defect in this PR.

`PROSPER_DECODE_SCRATCH_MB=0` is worth a third image for the same reason — it exercises the
zero-tail contract from the other side, since an unpooled buffer arrives zeroed.

## Not expected

This should NOT reach 30 fps. It is worth roughly 450 ms of a 5020 ms window (~9%). The lever that
matters for the target is the compute-image import miss recorded on #3307.

mattias800:
## Re-verified on the head that adds `PROSPER_NO_TEXTURE_PREFIX_INHERIT`

```
cmake --build prosper/build-linux -j4   -> exit 0, 0 compile errors, 0 targets stale on re-run
ctest --no-tests=error -j2              -> exit 0, 100% tests passed out of 357, 0 skipped
```

`check_env_numeric_arms` (31 knobs, all armed), `check_cached_env`, `check_diag_gates`
(766 variables at 1206 sites, findings 108 = baselined 108), `check_numbered_table`,
`check_trap_citations`, `check_ctest_gate`, `check_ascii_output`, `git diff --check` — all clean.
The new lever is a `PROSPER_ENV_VALUE(...) != nullptr` predicate, not an `env_numeric` parse, so
it needs no site arm; `check_diag_gates` sees it and is unchanged at its baseline.

## My judgement on the pixel arm, since you asked for it rather than assuming it

**It does not block, and I am not calling it "identical".**

Why it does not block. Both arms produced exactly two distinct images, and *every image the ON arm
produced is byte-identical to one the OFF arm also produced*. Stale-buffer corruption — the failure
mode the zero-tail contract and `detile_writes_whole_destination` exist to prevent — introduces a
**third** hash. It cannot coincidentally reproduce, bit for bit, the settled frame the other arm
reaches independently.

The one non-benign mechanism consistent with the data, named so nobody has to re-derive it: dropping
the staging copy means the detiler reads guest memory across the detile's duration rather than
snapshotting it first, so a concurrent guest write could be seen. That would produce a **torn**
frame — a third hash again. It is not present. What remains is the faster arm reaching the settled
state one ordinal earlier, which is what a rate change does, and is the latitude the hardware
already has (it samples at execution, not at submit). The existing `direct_resource` storage-buffer
site rests on the same reasoning.

Why it does not close the question. Two matching steady states prove the steady state; they do not
prove every transient. And an **ordinal anchor cannot align two arms with different frame rates** —
that is precisely what failed here, so "frame-locked at the same ordinal" is not available as a
stronger version of the same experiment.

## Two follow-ups that would close it, neither needing a build

1. **Deterministic replay.** One F9 `.prgbundle` replayed offline through `gpu_replay` with and
   without `PROSPER_NO_DIRECT_TEXTURE_SOURCE`: same commands, same resources, no guest, no
   scheduler, so byte equality has no phase confound. **Its limit stated up front so it is not
   oversold:** replay disables the persistent decode cache by construction
   (`texture_decode_cache_candidate` requires `!has_captured_host_data`) and its resources are
   `host_data`-backed. So it covers the pooling, the zero-tail contract, the coverage predicate and
   the *host_data* direct read — **not** the guest-memory `safe_span` read, and not the prefix
   inheritance.
2. **A phase-immune live comparison** for what (1) cannot reach: over ~60 settled frames per arm,
   compare the *set* of distinct hashes rather than frame-by-frame. If neither arm produces a hash
   the other never produces, that is strictly stronger than six aligned frames and immune to the
   rate difference. Today's data already satisfies that at n=6, which is why it reads as reassuring
   rather than conclusive.

## And one re-run I would value more than either

The lever arm you ran predates `PROSPER_NO_TEXTURE_PREFIX_INHERIT`, so it under-reports this PR by
the ~192 ms that commit was contributing while unswitchable. A discriminator arm with **all three**
levers set should land at the 1110-ish baseline; if it does not, the attribution is incomplete and
neither the 497.7 nor the 918.2 should be quoted without saying so.

Not urgent, and explicitly behind your `PROSPER_COMPUTELOG` work on `0x30784e0000` / `0x30d10f0000`
(#3307) — that lever is worth more than this whole PR and I would rather the GPU went there first.

mattias800 (commented):
## Independent review of `84a91e81d` (`perf/stray-texture-materializer`)

**REJECTED** on one finding. It is a diagnostic-message and claim defect, not a
correctness defect in the decode path — everything the PR's safety argument rests on
checks out, and I mutation-tested the test that carries it. The fix is one argument and
two sentences.

---

## Blocking

### B1. The `PROSPER_DECODE_SCRATCH_MB` typo message tells an operator retention is OFF when it is ON — and the PR body states the same thing as fact

`prosper/frontends/shared/live/decode_scratch.hpp:179-181` passes `"0 = retain nothing"`
as `env_u64_or_default_capped`'s **`default_note`** while the fallback is **512**.
`prosper/src/diagnostics/env_numeric.hpp:62-66` documents that parameter as glossing
*the default*, and it is rendered inline with the default value. Measured — I compiled
`frontends/shared/tests/test_decode_scratch.cpp` standalone and captured its stderr:

```
[env] PROSPER_DECODE_SCRATCH_MB='64mb' is not a plain non-negative count of MiB
      -- keeping the default (512, 0 = retain nothing) and changing NOTHING
```

That line reads as "the default, 512, retains nothing." It does not. 512 MiB is
retention fully ON.

The two other sites in the tree that pass a `default_note` both have a default of `0`,
which is why the parameter reads correctly there:

- `prosper/src/gpu/execute/gpu_executor.cpp:7739-7740` — fallback `0ull`, note `"no cap: nothing is bounded"`
- `prosper/src/host/memory/guest_write_watch.cpp:1169-1170` — fallback `0ull`, note `"unbounded: no range is too large to watch"`

The same error is asserted as fact in two places that outlive the code:

- `decode_scratch.hpp:36-37` — "**0 retains nothing**, which is exactly the pre-pool
  behaviour: the sentinel that a mistyped value selects is the SAFE one, per #3253"
- the PR body, Levers section — "`PROSPER_DECODE_SCRATCH_MB=0` — retain nothing, which
  is the pre-pool allocation pattern exactly. **A mistyped value selects that same safe
  sentinel (#3253).**"

Both are false. A mistyped value selects **512 MiB**, i.e. the optimisation stays on.

**The PR contains its own refutation, in two of its own new tests**, which is what makes
this worth blocking on rather than noting:

- `frontends/shared/tests/test_decode_scratch.cpp:161` — `CHECK(decode_scratch_budget_bytes("64mb") == (512ull << 20));`
- `tests/diagnostics/test_env_numeric_sites.cpp:373-374` — the new census row's
  malformed-value answer is `512ull * kMiB`, and commit `a989123e4`'s own message
  explains that a malformed value *keeps the fallback*.

**Concrete failure scenario.** This PR's body prescribes a three-lever A/B and says "a
re-run with all three levers is the arm worth quoting." The next agent to run it reads
that body, types `PROSPER_DECODE_SCRATCH_MB=0mb` (or `0MB`, or `512m`), gets a binary
with pooling fully active, sees a stderr line whose parenthetical says "0 = retain
nothing", and attributes the resulting number to an arm that was never armed. That is
precisely the instrument trap `env_numeric.hpp` was written for, and #3155/#3253 record
that this family has already produced one retracted measurement.

**Suggested fix** (behaviour of the code is already right — the loud refusal that keeps
the default is what the header prescribes; only the message and the two claims are wrong):

1. `decode_scratch.hpp:181` — drop the `default_note` argument, or reword it so it
   describes 512 (e.g. `"512 MiB retained per thread; set 0 to retain nothing"`).
2. `decode_scratch.hpp:36-37` — say that a mistyped value keeps the **default (512 MiB,
   pooling on)** and prints a refusal, and that 0 must be spelled correctly to get the
   pre-pool pattern.
3. PR body — same correction, since that is the text an A/B is driven from.

I will re-review on the corrected head; nothing else needs re-checking.

---

## Non-blocking

**N1. The poison sweep pins `detile_surface`; the BC call site uses `detile_elements`.**
`tests/gpu/texture/test_tile.cpp:1272-1326` only ever calls `detile_surface`, while
`src/gpu/texture/tile.hpp:108-121` claims the predicate answers for *both* entry points
and `frontends/shared/live/live_renderer.cpp:5897` (the BC branch) is
`detile_elements`. I verified by reading that the claim holds today —
`tile.cpp:1361-1378` and `tile.cpp:1840-1859` route to the same `sw256_copy` /
`sw4kb_copy` / `sw64kb_copy` under the same `is_64kb_mode` / `Sw256BS` guard, and the
only difference (`detile_elements` forwards the caller's `src_bytes` where
`detile_surface` computes the full tiled size) affects the *source* bound, not
destination coverage. But a future fast path added to `detile_elements` would silently
falsify the predicate for the BC path with nothing turning red. Calling both inside the
same loop is about three lines and would close it.

**N2. The check→use window in `safe_copy` widened, and on the direct path it is the
whole detile.** `live_renderer.cpp:2319-2339`: the old form re-checked
`prosper_reserved_range_state` immediately before each 64 KiB memcpy; the new form
proves the whole prefix and then copies once, and `direct_resource_source` hands the
detiler live guest memory for the duration. Bytes delivered are identical and a
mid-walk unmap was fatal before too, so this is not a regression in kind — the author
names it in the body and the pixel-identity evidence covers the observed run. Recording
it here so the next reader does not re-derive it.

**N3. `frontend_tex_other_n` is written but never read.**
`frontends/shared/perf/performance_capture.cpp:406` serialises it and
`test_performance_capture.cpp:210` asserts it reaches the artifact, but
`tools/perf/performance_capture_report.py` never prints it — so the cross-check its own
field comment promises (`performance_capture.hpp:95-98`: "paired with the signed
millisecond residual … the two disagree only if the classification itself is wrong") is
not actually available to a reader of the report.

**N4. `DecodeScratchPool::Lease::valid()`** (`decode_scratch.hpp:84`) has no caller.

**N5. The retention budget is per-thread.** `decode_scratch_pool()` is
`static thread_local` with `max_retained = 4`, so on Stray's shape a decoding thread can
hold ~252 MiB and N threads N× that; the census line at `live_renderer.cpp:10889`
reports only the calling thread's pool. The header does say "what a **thread** retains",
so this is documented rather than wrong — but the printed figure is not the process
total it looks like.

**N6. `take(0)`** returns a lease whose `data()` may be a non-null pointer into a
retained buffer where `std::vector<uint8_t>(0)` gave `nullptr`. No current call site has
a zero extent, and nothing compares `data()` to null; noting it because the class's
contract does not say which it is.

---

## What I actually checked, so this is falsifiable

**Stale-buffer safety, every converted branch.** I traced each lease from `take()` to
its last read and confirmed no byte is read that is neither written nor cleared:

- BC (`:5871-5909`) — `lin` is read by `bc_decode_surface(..., lin.size())`. mip-tail →
  `zero_all`; tiled → predicate-gated, and `detile_elements` covers `bw*bh*bcb` exactly;
  `linear_padded_read` → `zero_all` (**required**: `copy_linear_padded_rows` fills only
  `bh` rows of `bw*bcb`, which is short for a volume); else → `zero_tail(copy_resource)`.
- fp32 (`:5911-5969`), fp16 (`:5971-6057`), narrow (`:6059-6133`) — the `got < X` arm,
  the volume arm and the mip-tail arm all pre-zero; only the plain `detile_surface` arm
  skips it, and there `is_volume` is provably false so the lease extent is exactly
  `tw*th*bpt`. The read loops (`:5961-5968`, `:6028`, `:6114-6131`) stay inside
  `volume_texels*bpt`.
- generic 32-bpp (`:6163-6231`) — both arms hold. `tiled_bytes >= nb` takes the direct
  path and then `got == tiled_bytes >= nb`, so the `got < nb` recovery is unreachable
  there (the defensive re-stage inside it is dead code, correctly so);
  `tiled_bytes < nb` always stages, and the recovery memcpy overwrites the whole lease.
  Equivalent to the old zero-initialised vector in both.
- `bpt == 3` reaches the narrow branch (`data_format_bytes * num_components`, and the
  clamp at `:3525` only fires above 4). Predicate returns false for 64 KiB modes there
  (`sw64kb_elem_log2(3) == UINT32_MAX`) and true for 4 KiB/256 B, where the copiers do
  cover every element at any `bpe`. Safe.

**`detile_writes_whole_destination` — read against all three copiers, then
mutation-tested.** `sw256_copy` (`tile.cpp:205-235`) and `sw4kb_copy`
(`tile.cpp:333-364`) walk every `(row, col)` and either copy or `memset` `bpe` bytes;
`sw64kb_copy`'s block-major (`:816-889`), row-major (`:892-909`) and AVX2
(`:623-679`) paths all do the same, and the AVX2 vector loops write exactly
`columns*bpe`. The one early-out is the unsupported-`bpe` `memcpy(dst, src, min(n, tiled_bytes))`
at `:766-770`, which is exactly what the predicate excludes.

I then built `test_tile` standalone against four mutated copies of `tile.cpp` to check
the sweep is not vacuous. Baseline `== PASS ==`; each mutant introduces one coverage hole
and the sweep catches it:

| mutation | poison-sweep failures |
| --- | --- |
| `sw64kb_copy` block-major drops the last row of each block | 72 (first: mode=9 bpe=1 256x128) |
| `sw4kb_copy` drops the last column | 24 (first: mode=5) |
| `sw256_copy` drops the last column | 24 (first: mode=1) |
| `detile_full_block_row_avx2` scalar tail stops one element short | 18 (mode=9 bpe=2/4/8, ragged extents only) |

So the sweep genuinely exercises all three copiers *and* the AVX2 gather path, and the
ragged extents in its size list are load-bearing (the AVX2 mutant is invisible without
them). The `Sw64KbRX, 6u` mutation arm at `:1322-1323` is what keeps `claims_full` from being
vacuous, and it does that job.

**`direct_resource_source` vs `copy_resource`, line by line** (`:2750-2792`). Host arm:
`copy_resource` returns `min(n, host_data_size - off)`, so `take == n` iff
`n <= host_data_size - off`; the direct arm returns nullptr on exactly the complement,
plus the same `addr < gpu_addr` and `off >= host_data_size` rejections. Guest arm:
`copy_resource` returns `safe_span(addr, n)`; the direct arm requires `== n`, with two
strictly-narrowing extra guards. No over-permissive case.

**`safe_span` / `safe_copy` equivalence** (`:2319-2339`). Same chunk predicate, same
byte count, `n == 0` returns 0 either way, and the `_WIN32` `prosper_try_commit_dmem`
arm keeps its meaning (commit succeeds ⇒ whole range readable ⇒ `n`). The three
`safe_copy` callers (`:2371`, `:2751`, `:2839`) all use the return as a byte count; none
depended on the interleaving.

**Lease lifetime and aliasing.** Every lease is an automatic local declared in the scope
that uses its pointer; `take()` removes a buffer from `free_`, so two live leases can
never alias; `parallel_rows` (`tile.cpp:567-592`) uses `std::jthread` and joins before
returning, so no worker outlives a lease; no lease crosses a thread boundary anywhere in
the diff; and the const-pointer refactor means a branch that tried to write through
`traw` would not compile. I also re-derived the pool arithmetic for both budget arms of
`test_decode_scratch` by hand (destruction is reverse-declaration order; 4+1+1 against a
5 MiB budget really does leave `[1,4]`, and the `max_retained=2` arm really does leave
2048+4096) — the numbers are right, and the "hold the leases simultaneously" comment at
`:115-118` is doing real work.

**The `source_prefix` inheritance** (`live_renderer.cpp:6560-6615`). Both of the
author's claims hold. `PersistentDecodedTexture::bytes()` at `:817` is
`source_prefix.size() + pixels->size()`, and the move at `:6572` is after both the
`retired_submit_bytes` credit and the `persistent_decoded_texture_bytes` debit at
`:6564` / `:6567`, so moving first would under-debit permanently. `source_prefix.size()`
is compared against `persistent_source_size` at `:4056` (which selects the
no-scratch-copy fast validation) and against `validated_bytes` at `:4074`, so the
`resize(source_prefix_size)` at `:6610` is required, not cosmetic. Inheriting a
*smaller* buffer is fine — `resize` grows, then the memcpy overwrites all of it. The
`persistent_validation_scratch` short-read case is bounded: when the memcpy runs,
`source_prefix_size` is either `linear_source_prefix_size` or the return of
`copy_persistent_source` into a scratch just resized to `persistent_source_size`, so the
scratch holds at least that many bytes — and the same reachability question applies
identically to the `assign()` this replaces, so nothing is newly exposed. Entries whose
prefix was already released by the `source_watch_only` path at `:4220` inherit an empty
vector, which is correct.

**Repo gates**, run from `prosper/`: `tools/env/check_env_numeric_arms.py` — ok, 31 knobs,
every one armed. `tools/env/check_cached_env.py` — all checks passed.
`tools/env/check_diag_gates.py` — all checks passed, 108 findings all baselined.
Commit trailers: 0 `Claude-Session:` lines, 0 "Generated with" badges across the 14
commits. `BLOG.md` entry present. No absolute host paths in the added docs.

**Not re-run**, per the review brief: the author's build, full ctest, snapshots, and the
live three-arm A/B.

---

Once B1 is corrected I expect to approve; nothing else in this diff needs another pass.

mattias800 (commented):
### Head moved during the review — verdict carries, by content

My review above is written against `84a91e81d`; GitHub attached it to `bc1538639`,
because the branch advanced while I was reading. Establishing by content rather than
assuming: `git diff 84a91e81d bc1538639` is **`prosper/docs/STRAY_STATUS.md` only**,
10 insertions / 6 deletions, and it walks back the doc's "the missing 191.9 ms is that"
to "some of it is, and how much is not established". No code, no test, and none of the
files my blocking finding cites.

So the verdict is unchanged and stands on `bc1538639`: **REJECTED**, on B1 only —
`decode_scratch.hpp:181`'s `default_note` and the two prose claims that a mistyped
`PROSPER_DECODE_SCRATCH_MB` selects the retain-nothing sentinel. Everything in the
"what I actually checked" section applies to this head unmodified.

The docs change itself is an improvement and I have no finding on it.

mattias800 (commented):
## Re-review of `44462f7ad` — **APPROVED**

B1 is fixed, all five non-blocking findings are taken, and I verified the two that carry
a real claim by mutation rather than by reading the diff. Nothing new found. The delta
since the head I first reviewed (`84a91e81d`) is six files, and only one of them
(`live_renderer.cpp`) is production code — a single format-string label plus a comment.
The decode path itself is byte-identical to what I reviewed.

---

### B1 — resolved, checked against the rendered line again

I rebuilt `test_decode_scratch.cpp` at this head standalone and captured stderr:

```
[env] PROSPER_DECODE_SCRATCH_MB='0mb' is not a plain non-negative count of MiB
      -- keeping the default (512, 512 MiB retained per thread; spell 0 exactly to
      retain nothing) and changing NOTHING
```

That is unambiguous in the direction that matters, and the `(512, 512 MiB …)`
redundancy now reinforces rather than contradicts. All three near-miss spellings print
it and all three assert 512 MiB (`test_decode_scratch.cpp:167-169`), so the retracted
claim cannot be restated without deleting a green assertion — which is the right shape
for a claim that already survived one review cycle by being plausible.

The header (`decode_scratch.hpp:35-48`) and the PR body's new corrections section both
state the direction *and* why it is the dangerous one (the knob's main use is disarming
the pool, so the typo an operator makes is the one meant to turn the optimisation off).
That is a better fix than the one I asked for.

### N1 — the added arm is independently verified, not just added

I did not take the non-shadowing claim on report. I truncated `detile_elements` **alone**
(one row off `eh` at `tile.cpp:1842`, before the routing) and re-ran the sweep:

```
detile_elements failures: 144
detile_surface  failures:   0
```

So the new arm detects a defect the existing arm cannot see. (My 144 against your 72 is
just a wider mutation — mine truncates the untiled routing too, so `mode=0` shows up in
my output. Nothing disagrees.) Baseline at this head is `== PASS ==`, and the whole
`test_tile` binary is **1.77 s / 162 MiB peak** with the sweep doubled, so there is no
timeout or memory concern from running both entry points.

### N3 — exercised in four arms, including the two that must stay silent

Driving `_resource_breakdown` on synthesised records and evaluating the two new
fragments verbatim:

| arm | `tex_other` | `tex_other_n` | suffix | shout |
| refs behind the residual | +83.00 | 7 | `/7refs` | no |
| residual with **no** refs | +83.00 | 0 | `/0refs` | **yes** |
| old capture, field absent | +83.00 | `None` | *(none)* | no |
| refs 0, residual ~0 | +0.00 | 0 | `/0refs` | no |

The absent case is the one that matters and it behaves: `None == 0` is False in Python,
so an older `.prperf` gets no suffix and no false alarm, consistent with this file's
absent-is-not-zero rule.

### N4, N5, N6 — confirmed

`valid()` is gone with no remaining reference anywhere under `frontends/shared/live/`;
the zero-extent `data()` contract is stated at `decode_scratch.hpp:91-93` and correctly
tells the reader to test `size()`; the census reads `intermediates(this thread)=` with a
comment saying why. N2 as a recorded observation is the right call — there is nothing to
change, only something not to re-derive.

**Hygiene at this head:** `check_env_numeric_arms` (31 knobs armed), `check_cached_env`,
`check_diag_gates` all green; `git diff --check origin/main...44462f7ad` clean; 0
`Claude-Session:` trailers and 0 attribution badges across the two new commits.

---

## Your two questions

**1. No, do not add `PROSPER_NO_DECODE_SCRATCH_POOL`. Your judgement was right.** Three
reasons, and one thing I would do instead.

The asymmetry you are worried about is not the axis it looks like. The other two levers
are presence-checked because what they control **is** boolean; `PROSPER_DECODE_SCRATCH_MB`
is a budget, and `0` is a point in its range rather than a separate mode. Promoting a
range endpoint to its own knob is surface, exactly as you said.

Second — and this is the part that would actually make it worse — a redundant second
spelling of one lever creates a *new* failure mode of the same family this PR just
corrected: two switches for one behaviour can be set inconsistently
(`…_NO_DECODE_SCRATCH_POOL=1` with `…_SCRATCH_MB=512`), and a reader then has to know
which wins. Correction #2 in your own body is about a lever that did not cover
everything it was named for; a lever that covers something twice is the mirror of it, and
both end the same way — an arm nobody can state precisely.

Third, the residual risk is now instrumented rather than silent. The refusal names the
variable, the text given, the default kept, and what to spell instead, and three green
assertions pin it. That is what `env_numeric.hpp` is for, and it now works as designed
here.

**What I would do instead, and it costs one line:** put the literal, copy-pasteable
command in the body's Levers section and in `STRAY_STATUS.md`, e.g.

```
PROSPER_NO_DIRECT_TEXTURE_SOURCE=1 PROSPER_DECODE_SCRATCH_MB=0 PROSPER_NO_TEXTURE_PREFIX_INHERIT=1
```

A typo is what happens when someone types the knob from memory; it does not happen when
they paste the line. That removes the remaining hazard without adding a second way to
express one state. Non-blocking — take it or leave it.

**2. Keep the single loop.** Splitting into separate named cases would duplicate the
shape setup and, more importantly, would stop both entry points seeing the *same*
poisoned destination for the same shape, which is what makes the comparison worth
anything. The `who` label in the failure message is the thing that makes one loop safe —
my mutation run was attributable to the exact entry point without any extra structure —
so keep that, and it is fine as it stands.

---

**APPROVED** on `44462f7ad`. No blocking findings. From my side this is ready to merge
once your CI is green; the only outstanding item is the optional copy-pasteable A/B line
above, which needs no re-review either way.

mattias800:
## Review closed, and one docs-only commit on top

The independent review **APPROVED** at `44462f7ad`. Its earlier verdict was **REJECTED** at
`bc1538639` on one blocking finding (B1), which is fixed; both older review rows are attached to
that superseded head, so the newest row by `submitted_at` is the one that carries.

Since then, one commit: `51805abc1`, **documentation only**, taking the reviewer's non-blocking
suggestion to give the disarmed A/B arm as a line to paste rather than retype. It touches no code,
so the approval at `44462f7ad` still applies by content — `git diff 44462f7ad 51805abc1` is
`prosper/docs/STRAY_STATUS.md` alone.

## Why there is no fourth lever

I asked the reviewer whether the asymmetry was itself a hazard — two typo-proof presence levers
and one typo-sensitive numeric one — and offered to add `PROSPER_NO_DECODE_SCRATCH_POOL`. Their
argument against is better than my argument for, and it is worth recording rather than leaving in
a message thread:

`PROSPER_DECODE_SCRATCH_MB` is a **budget**, and `0` is a point in its range rather than a separate
mode; the other two levers are booleans because what they control is boolean. A redundant second
spelling of one lever would create the mirror of this PR's own correction #2 — two switches for one
behaviour can be set inconsistently, and a reader then has to know which wins, which ends in an arm
nobody can state precisely. The residual risk is now instrumented (a loud refusal, a truthful
message, and three assertions pinning the near-miss spellings) rather than silent.

So the fix is the smaller one: the exact command, in the PR body and in `docs/STRAY_STATUS.md`,
to be pasted.

```
PROSPER_NO_DIRECT_TEXTURE_SOURCE=1 PROSPER_DECODE_SCRATCH_MB=0 PROSPER_NO_TEXTURE_PREFIX_INHERIT=1
```

## Status

Approved, CI running on the head. **Not merged** — my brief says open, not merge, and that call is
the coordinator's.

Issue #3311, opened 2026-09-04
Stray compute 0x300e390000 pc412: EXEC restored from a wave mask spilled through a scalar copy and a v_writelane/v_readlane lane slot


mattias800:
With #3308 landed, Stray's compute program `0x300e390000` advances from pc4 to **pc412** and from
3220 to 11632 emitted SPIR-V words, then rejects:

```
[cfg-recompile-reject] mode=unresolved-operand pc=412 words=befe044a len=1 fmt=1 op=0x4
```

`befe044a` is SOP1 `op=0x4` with `SDST=126` (EXEC_LO), `SSRC0=74`:

```
s_mov_b64 exec, s[74:75]
```

Confirmed live by the lane holding the GPU, and reproduced exactly offline (below).

## Why it does not resolve — the provenance chain

`s[74:75]` is not a mask register the recompiler ever tracked as a mask. Its two halves are assembled
independently, from two different spill routes, out of a mask computed 100+ dwords earlier:

```
  303: s_and_b64      s[14:15], s[58:59], vcc     ; the mask is computed here
  306: v_writelane_b32 v18, s15, 5                ; HIGH half spilled into a VGPR lane slot
  319: s_mov_b32      s74, s14                    ; LOW half copied as ordinary scalar data
  346: v_readlane_b32 s75, v18, 5                 ; HIGH half read back out of the lane slot
  412: s_mov_b64      exec, s[74:75]              ; EXEC restored from the reassembled pair
```

So the guest decomposes a 64-bit wave mask into two 32-bit words, moves one through a scalar copy and
the other through a `v_writelane`/`v_readlane` round trip, and reassembles EXEC from them. prosper
models a mask as a **per-lane bool**, which has no 32-bit word representation to copy, and neither
`s_mov_b32` nor a VGPR lane slot carries mask provenance — so at pc412 the recompiler holds two
ordinary integers where it needs a predicate.

The same program does this eight more times: `v_writelane_b32` with a mask-half source at pc 145,
306, 344, 386, 391, 395, 399 and 403. They are one family, not nine separate gaps.

## What would be needed

Two halves, and the second is the hard one:

1. **word → per-lane bool.** With `native_subgroup_size == 64` this is expressible —
   `((lo | (hi << 32)) >> gl_SubgroupInvocationID) & 1` — and `native_wave_ballot_half` already does
   the bool → word direction. So the conversion itself is not the blocker.
[...]

PR #3312, opened 2026-09-04
recompiler: the CFG dispatcher must not stamp the entry-M0 token at the entry block

### LINKED ISSUE #3308: The CFG dispatcher stamps #3133's entry-M0 token at the ENTRY block, refusing inbound launch registers — Stray's compute 0x300e390000 dies at pc4 on v_lshl_add_u32
`load_state` in `rdna2_emit_cfg.cpp` re-arms #3133's entry-M0 token on a **whole-stream MAY set** at
**every** dispatcher block entry — including the program's own **entry block**, where no instruction
has executed and therefore no `s_mov_b32 sX, m0` can have run. Every register any save in the shader
writes is stamped as "holds the driver's M0, has no value" before the first instruction, so the
shader's first read of one of those registers rejects with `mode=unresolved-operand` and the whole
stage/dispatch is dropped.

`prosper/src/gpu/recompiler/rdna2_emit_cfg.cpp:3640` (pre-fix):

```
for (int reg : entry_m0_may_hold) {   // #3133 (see the MAY set above)
    state.sreg.erase(reg);
    state.sreg_entry_m0.insert(reg);
}
```

`entry_m0_may_hold` is built at `:1811` by scanning the **entire** instruction stream, with no
reachability or ordering condition. #3133's own comment predicted the cost as "over-rejects a
register that the shader reuses for real data in a LATER block". The entry block is the case where
that is not a judgement call but provably wrong.

## Worked example — Stray (PPSA02101), compute program `0x300e390000`

Live on a routed title-screen boot:

```
[compute] skip unsupported program 0x300e390000 reason=cfg-recompile-reject mode=unresolved-operand
          pc=4 words=d746000b,0401060e len=2 fmt=9 op=0x346
```

`d746000b 0401060e` disassembles (llvm-mc, `-mcpu=gfx1010` and `-mcpu=gfx1030` agree) as:

```
v_lshl_add_u32 v11, s14, 3, v0        ; v11 = (s14 << 3) + v0
```

VOP3A: `VDST=11`, `SRC0=14` (s14), `SRC1=131` (inline constant 3), `SRC2=256` (v0); no
abs/neg/omod/clamp/op_sel. It is the shader's own **global thread index** — workgroup-id X times the
workgroup width, plus the local invocation id — four dwords into the program.

`op=0x346` is **`v_lshl_add_u32`, which prosper already lowers** at
`prosper/src/gpu/recompiler/rdna2_emit_alu.cpp:4844`, and it is already on the B32 VOP3 source-width
allowlist at `rdna2_cfg_support.hpp:1304`. Nothing about the instruction is missing — the diagnostic
says so, and `mode=unresolved-operand` is exactly the field #2412 added to separate the two cases.

The register is the whole story: **s14 is both the compute stage's workgroup-id X and the register
this shader saves M0 into**, at pc157 and again at pc274:

```
    4: v_lshl_add_u32 v11, s14, 3, v0     <-- rejected
  157: s_mov_b32 s14, m0                  ; be8e037c -- 153 dwords AHEAD of the read
  238: s_mov_b32 m0, s14
```

So the entry-block read was refused for a save it cannot reach. Instrumenting `operand_bits`
(`rdna2_alu_support.hpp:203`) names the branch directly:

```
[opbits] pc=4 s14 reject=entry_m0
```

## Reproduction — offline and CPU-only, no boot and no GPU

The program's bytes are recoverable from any of the existing Stray F9 bundles, and `shader_inspect`
runs the whole compute translator on them:

```
PROSPER_DBG=1 ./build-linux/shader_inspect <prog>.bin --stage compute
[cfg-recompile-reject] mode=unresolved-operand pc=4 words=d746000b,0401060e len=2 fmt=9 op=0x346
[cfg-dispatcher-partial] rejected after emitting 3220 words
```

## Fix

Replace the whole-stream stamp with an **entry-rooted forward MAY dataflow** over the CFG the mask
analyses in the same function already walk (`successors`, `starts`): GEN is #3133's save shape, KILL
is any other scalar write to the word taken in instruction order, and the join is a **union** because
the token is a MAY property whose purpose is to reject. The entry block's in-set is empty by
construction. A block the walk never reaches falls back to the whole-stream set, so the unknown case
keeps #3133's conservative behaviour.

After it, the same program advances from pc4 to pc145 and from 3220 to 8931 emitted words. It still
does not compile — pc145 is `v_writelane_b32 v20, s15, 2`, a spill of the high half of the Wave64
mask `s[14:15]`, and behind that at pc412 is `s_mov_b64 exec, s[74:75]`. Both are separate,
pre-existing frontiers. **This issue is the entry-block token, not the whole program.**

## Scope

Any shader containing `s_mov_b32 sX, m0` that reads one of those registers before the save, on any
path, in a program that routes through the CFG dispatcher. The set is empty for every shader with no
such save, so nothing that compiles today can change.

mattias800:
Fixes #3308. Refs #3126.

## The failure

A live Stray (`PPSA02101`) title-screen boot skips a compute dispatch:

```
[compute] skip unsupported program 0x300e390000 reason=cfg-recompile-reject mode=unresolved-operand
          pc=4 words=d746000b,0401060e len=2 fmt=9 op=0x346
```

### What the instruction is

`fmt=9` is `Rdna2Format::VOP3`. `d746000b 0401060e` decodes as VOP3A `VDST=11`, `ABS=0`, `OPSEL=0`,
`CLMP=0`, `OP=0x346`; `SRC0=14` (s14), `SRC1=131` (inline integer constant 3), `SRC2=256` (v0),
`OMOD=0`, `NEG=0`. Confirmed against llvm-mc, which agrees on `gfx1010` and `gfx1030`:

```
v_lshl_add_u32 v11, s14, 3, v0        ; D.u32 = (S0.u32 << S1.u32[4:0]) + S2.u32
```

That is the kernel's own global-thread-index computation — workgroup-id X scaled by the workgroup
width, plus the local invocation id — four dwords into the program.

### Why it was rejected, and why it is not a missing lowering

`v_lshl_add_u32` has been lowered at `prosper/src/gpu/recompiler/rdna2_emit_alu.cpp:4844` since long
before this title was tracked, and it is already on the B32 VOP3 source-width allowlist at
`rdna2_cfg_support.hpp:1304`. The diagnostic said as much: `mode=unresolved-operand`, not
`unknown-encoding` — the distinction #2412 added for exactly this reading.

`operand_bits` (`rdna2_alu_support.hpp:203`) refused **`s14`**, and instrumenting its three
`*ok = false` branches names which one:

```
[opbits] pc=4 s14 reject=entry_m0
```

`s14` is both the compute stage's **workgroup-id X** and the register this shader saves M0 into:

```
    4: v_lshl_add_u32 v11, s14, 3, v0     <-- rejected
  157: s_mov_b32 s14, m0                  ; be8e037c
  238: s_mov_b32 m0, s14
```

`load_state` in `rdna2_emit_cfg.cpp` re-armed #3133's entry-M0 token on a **whole-stream MAY set** at
**every** dispatcher block entry, the program's own entry block included:

```
for (int reg : entry_m0_may_hold) {   // #3133
    state.sreg.erase(reg);
    state.sreg_entry_m0.insert(reg);
}
```

`entry_m0_may_hold` is built at `:1811` by scanning the whole stream with no reachability or ordering
condition, so the read at pc4 was refused for a save 153 dwords **ahead** of it. #3133's comment
predicted the cost as "over-rejects a register that the shader reuses for real data in a LATER
block"; the entry block is the case where that is not a judgement call but provably wrong — nothing
has executed there, so no save can have run.

**The transferable part is the diagnostic-reading lesson, and it is the inverse of the usual one.**
#2481/#2801 record that a reject PC names where a fact was *consumed*, not where it was lost. Here
the two are 153 dwords apart **and in the wrong order**, the consumer coming first, which is why the
recorded cause in `STRAY_STATUS.md` pointed at pc157 and pc688 and nobody looked at pc4.

## The change

Replace the whole-stream stamp with an **entry-rooted forward MAY dataflow** over the same CFG
(`starts`, `successors`) the mask analyses in that function already walk:

- **GEN** — #3133's own save shape (`SOP1 op 0x03`, `src Special 124`, `dst <= 105`).
- **KILL** — any other scalar write to the word, via `for_each_scalar_write`, the recompiler's single
  writer inventory and the same one the `scalar_writes` loop directly above uses. Applied in
  instruction order, kill before gen, so a save and a later overwrite in one block end the lifetime
  the save started. A writer class the inventory does not report is a missed kill, i.e. the token
  survives — the over-rejecting, safe direction.
- **JOIN** — a **union**, deliberately not the MUST/equality join the neighbouring mask analyses use:
  the token is a MAY property whose entire purpose is to reject, so a register holding it on one edge
  and data on another must keep rejecting.
- The entry block's in-set is empty by construction. A block the walk never reaches — and
  `load_state`'s terminal-block call when `end_pc` maps to no block — falls back to the whole-stream
  set, so the unknown case keeps #3133's behaviour rather than silently dropping the token.

## What this does and does not fix

It moves this program's reject from pc4 to **pc145** and its emitted-word count from 3220 to 8931.

**The program still does not compile, and this PR is not evidence about Stray's title-screen
background.** pc145 is `v_writelane_b32 v20, s15, 2` — a spill of the high half of the Wave64 mask
`s[14:15]` — and with a native 64-wide subgroup adopted the next stop is pc412,
`s_mov_b64 exec, s[74:75]`. Both are separate, pre-existing frontiers.

The set is empty for every shader containing no `s_mov_b32 sX, m0`, so nothing that compiles today
can change. Measured over the 151 raw shader dumps on this machine, **2** contain the shape — one
Plucky Squire compute program (`exec_cs_3015ab0000`, saves into s8/s14/s20) and one Dragon Quest VII
fragment program (`exec_ps_308a410000`, s8/s36) — so the blast radius is small but not zero, and both
are shaped the same way this one was.

## Verification

All CPU-only; no GPU was used (another lane holds it).

**Deterministic offline reproduction.** The program's bytes are recoverable from any checked Stray F9
bundle, and `tools/shader_inspect` runs the whole compute translator on a raw dump:

| build | `shader_inspect --stage compute` | words emitted before the reject |
|---|---|---|
| before | `cfg-recompile-reject mode=unresolved-operand pc=4 words=d746000b,0401060e len=2 fmt=9 op=0x346` | 3220 |
| after | `cfg-recompile-reject mode=unresolved-operand pc=145 words=d7610014,0001040f len=2 fmt=9 op=0x361` | 8931 |

The alignment is pinned independently: dword 157 of the extracted program is `be8e037c`, matching the
`pc=157 words=be8e037c len=1 fmt=1 op=0x3` recorded from a live boot, and dword 688 is `bf86ffe8`,
the backedge recorded at pc=688.

**Regression arms.** `test_entry_m0_dispatcher` (the #3203 harness, chosen because it is the one place
whose programs route through the dispatcher with *no* structurizer fallback, so a verdict can only be
the dispatcher's) gains two:

- **Arm D** — a read at the entry block of a register saved later must compile, and must contain an
  `OpSwitch` so the dispatcher route is asserted rather than assumed.
- **Arm E** — the same program plus one read past the save's block must still reject, **at pc=6**.

**Mutation results, by exit code.** Each mutation built cleanly (`rc=0`) before being believed.

| mutation | test rc | signature |
|---|---|---|
| none (this PR) | **0** | 16/16 arms green |
| restore the whole-stream re-arm (pre-fix) | **1** | 3 fails: arm D's two lines and arm E's `pc=` line, whose reason reads `pc=0 words=4a020000 fmt=6 op=0x25` — the entry-block read refused. All four #3203 arms and all five controls stay green, which is why #3203 alone could not see this. |
| stamp no tokens at any block entry | **1** | every #3203 arm plus arms E and F. Arm D passes, correctly — nothing rejects when the token is gone. |
| **add back a `for_each_scalar_write` KILL** | **0** | **every arm stays green.** Recorded because it is the opposite of what it looks like: no arm here discriminates that KILL. |
| remove the region-boundary seed | **0** | every arm stays green — every program in the file is single-region, so the seed is a no-op for all of them. Stated in the file; see #3314. |

**That last row is the honest limit of this test file, and arm F's comment carries it.** Four
placements were tried and none discriminates: write and read in one block (the block-entry set is
never consulted), write behind one arm of an if (the union join restores the token), write
unconditional with the read in a later block (the KILL *does* drop the token and the B64 mask domain
refuses the read anyway), and `s_getpc_b64` to get a value-less write that is not a mask (the shell
refuses `op=0x1f` at the write itself). Every *accepted* value-less scalar write is a mask write, and
the mask domain guards those reads independently — so the hazard is reachable only through the gap in
that MUST/equality join, which is a larger fixture than this file wants. **The KILL is excluded by
the LOWER-bound argument in the source, not by a test**, and the file says so rather than implying
coverage it does not have. Both falsifications are recorded in `RECOMPILER_REMAINING.md`.

**Build and tests**, `-j4`, in the `ps5ys` container, exit codes captured separately from any pipe:

| step | result |
|---|---|
| full build | **rc=0**, 0 `error:` lines |
| `ctest --no-tests=error -j4` | **rc=0**, **100% tests passed out of 356** |
| `test_entry_m0_dispatcher` | **rc=0**, 16/16 arms |

Two earlier build attempts failed with `ld: final link failed: No space left on device` — a transient
exhaustion of the shared tmpfs while another lane was building. Not related to this change; recorded
because the first failure surfaced as a bogus `undefined reference` against a partially written
`libprosper_core.a` (41 of the expected archive members), which reads exactly like a real link error.

**Cross-title regression sweep — the strongest arm here.** `shader_inspect` was run over **152 raw
shader dumps** (every `exec_*`/`success_*` dump on this machine, plus the Stray program), under both
builds, and the verdicts diffed:

| | dumps |
|---|---|
| identical verdict | **150** |
| verdict moved to a **later** reject PC | **2** |
| verdict moved **earlier**, or accept → reject | **0** |

The two that moved are Stray's `0x300e390000` (pc4 → pc145 in the tool's default launch shape) and —
unlooked-for — **The Plucky Squire's `exec_cs_3015ab0000`, which had the byte-identical defect**:

```
pc=4 words=d746000c,0401060e len=2 fmt=9 op=0x346    v_lshl_add_u32 v12, s14, 3, v0
```

Same instruction, same register, same PC, different title and different engine. It advances to pc139.

**The safety argument, rewritten after review.** The first revision of this PR carried a KILL and a
proof whose second leg said "removing a token can only turn a reject into a success". Review
established that success is *not* the safe direction here, and it is right: `load_state` gives every
referenced scalar a Function-variable value (`:3699`), `save_state` stores **zero** for a written
register with no `state.sreg` entry (`:3860-3863`), and `operand_bits` consults `rs.sreg` first
(`rdna2_alu_support.hpp:204-206`) — so a dropped token reads back as **0 with `ok` true**, a silently
fabricated word replacing a loud reject. The KILL made that reachable, because the emitter's token
deliberately survives every value-less write (VOPC SGPR destinations, the B32 mask logical family at
`rdna2_emit_alu.cpp:1703-1707`, `s_getpc_b64`) while a `for_each_scalar_write` KILL fires on all of
them.

**The KILL is now gone rather than narrowed.** Narrowing it correctly needs a value-publishing
predicate shared with `emit_alu` so the two cannot drift, and no such predicate exists; approximating
one under-approximates in the unsafe direction. It was never needed anyway — the defect is at the
entry block, where the in-set is empty because no instruction has run. Both bounds are now
unconditional:

- **UPPER** — `entry_m0_in[B] ⊆ entry_m0_may_hold`, because the only insertion is GEN and GEN's
  predicate is the one that builds that set. So this can only *remove* tokens relative to #3133, and
  cannot turn an accepted program into a rejected one. (The `is_end` asymmetry review flagged —
  GEN `continue`s where the MAY-set scan `break`s — makes this set no *smaller* past an early
  `s_endpgm`, which is the direction the bound needs; the source now says so.)
- **LOWER** — `entry_m0_in[B] ⊇ {X : the emitter could hold X's token live on entry to B}`, **for
  saves inside this dispatcher region**. The only site that *originates* a token is a save
  (`rdna2_emit_alu.cpp:1337`); `:1354` erases one, and `join_entry_m0` both inserts and erases but
  only *propagates* an existing one across a structured join — it cannot invent a token for a
  register no save ever wrote. `record_scalar_write` never touches `sreg_entry_m0`. A union over
  every path from every save is therefore a superset of whatever the emitter can be carrying. **No
  token the emitter would keep is ever dropped, so the fabricated-zero hazard is structurally
  impossible inside a region.**

**Both bounds are scoped to one dispatcher region, and `emit_body` runs several.** Review of the
previous revision found this and it is correct: a barrier-phased compute kernel calls
`emit_cfg_state_machine` once per phase with the **same `RegState&`**, so a token created in phase N
is live on entry to phase N+1 — and nothing carries it across. The region exit does
`initial = load_state()`, whose `RegState` is default-constructed and copies seven named fields,
`sreg_entry_m0` **not** among them, while the stamp it applies erases those registers from
`initial.sreg`; the next region's prologue then stores **zero** into their Function variables, and
`operand_bits` returns it with `ok` true. Every link verified at its own site.

Block 0 is therefore **seeded** with `{X ∈ entry_m0_may_hold(this region) : entry_m0_live(initial, X)}`.
The intersection is load-bearing rather than tidiness — it is what keeps UPPER true across the seed,
so this still cannot newly reject a program that compiled before. That restores the loud reject for
the regression case (phase N+1 also saves into X, which the old whole-stream stamp covered by
accident).

**The case the intersection drops is left open and filed as #3314**: a register saved in an earlier
phase and not saved again in this one is not re-armed. That hole predates this change — the old set
was per-region too — and is not introduced here.

**The seed is not covered by a test, and that is stated in both the source and the test file rather
than implied away.** Two fixtures were built and neither reaches the path: an unguarded two-phase
kernel plus the readlane route is refused outright (`portable-readlane-dispatcher-unsafe
guest-barrier=1` — a guest barrier and a portable readlane are incompatible, and the readlane is how
every program in that file forces the dispatcher); and without the readlane, `emit_body` never takes
the phased path at all, so the straight-line emitter handled it and refused the cross-phase read from
the shared `RegState`'s own token — a **pass for the wrong reason**, so that arm was deleted rather
than kept. Reaching it needs `initial_dispatch_active`, a phased region nested inside a dispatcher
(Astro Bot's `0x500571000` shape). #3314 asks for that fixture first.

Precision cost, measured: none on the program this started from — the live launch shape still reaches
**pc412 with 11632 words emitted**, exactly as with the KILL.

**Sweep, re-scoped as review asked.** The previous "152 dumps" number was misleading: almost all of
them have an empty `entry_m0_may_hold` and short-circuit, so both builds are trivially identical.
Rescanned **16,560 local `.bin` files**; **938** carry `s_mov_b32 sX, m0` and are the only files where
this change can do anything. Of those, only **102** actually reach the CFG dispatcher — the rest
lower through the structured/straight-line emitter, where `load_state` never runs. A/B over all 938:

| | carriers |
|---|---|
| identical verdict | **936** |
| reject moved **later** | **2** |
| reject moved **earlier** | **0** |
| accept → reject | **0** |
| reject → accept | **0** |

Re-run on the final code after the region-boundary seed was added: **identical to the previous
revision on all 938**, which is the expected result — every carrier in that corpus is single-region,
so the seed is a no-op for all of them (recorded in the mutation table as such).

The two are Stray's `0x300e390000` (pc4 → pc145) and The Plucky Squire's `exec_cs_3015ab0000`
(pc4 → pc139), both `[cfg-recompile-reject] … op=0x346`:

```
pc=4 words=d746000c,0401060e len=2 fmt=9 op=0x346    v_lshl_add_u32 v12, s14, 3, v0
```

Same instruction, same register, same PC, different title and different engine.

**On the third carrier — I could not confirm it, and am not claiming it.** Review pointed at
`~/gta5-work/lighting.bin` (GTA V, 2935 dwords), and the pattern is certainly there: it saves M0 into
s6/s7 and its pc1 is `v_lshl_add_u32 v5, s6, 6, v0`, byte-identical in shape. But its verdict is
**identical in both builds** — offline it lowers through the *straight-line* emitter and rejects at
pc6 on a MUBUF with no resource table (`[recompile-reject] … fmt=12 op=0x0`), so it never reaches
`load_state` and the entry-block stamp never applies. It is a carrier of the **pattern**, not a
demonstrated carrier of the **defect**. This is a two-title fix on the evidence I have; a live run
with a real resource table would be needed to say more.

**Live validation — length-matched, 12 frames per arm, same route and warmup, back to back**
(run by the lane holding the GPU, Stray title-screen route):

| | `origin/main` | this PR |
|---|---|---|
| programs skipped | `0x300ba70000`, `0x300e390000`, `0x30131d0000` | `0x300e390000`, `0x30131d0000` |
| reject PCs | pc=3, pc=4, pc=50 | **pc=412**, pc=50 |
| `[mimg-unresolved]` lines | 8 | 8 |
| `max_nonblack` | 0.0052 | 0.0052 |

Two program-level effects, both in the right direction:

- **`0x300e390000` advances pc=4 → pc=412** — the program this PR diagnoses. pc412 is
  `s_mov_b64 exec, s[74:75]`, filed as **#3311**: a wave mask the guest spills through a scalar copy
  and a `v_writelane`/`v_readlane` lane slot and reassembles into EXEC, i.e. the wave-model frontier,
  not a bounded add.
- **`0x300ba70000` stops being skipped entirely.** It is skipped on `origin/main` at pc=3 and is not
  skipped here. This is the same defect: a read of a token-stamped register three dwords into the
  program.

Nothing regresses: no program is skipped here that is not also skipped on `origin/main`.

**An earlier unmatched comparison in this PR body claimed two things that were both wrong, in
opposite directions, and both are withdrawn.** It compared a 30-frame baseline against a 12-frame arm:

- it said `0x30131d0000` "stops being skipped" — it does **not**. It rejects at **pc=50 in both
  arms**, unchanged. There is no second program fixed.
- it said `0x300ba70000` was *newly* skipped here and needed the subset proof to excuse it — the
  reverse is true, as the table shows. The program simply was not reached in the longer reference run.
- a `[mimg-unresolved]` "10 → 4" improvement was the same artifact; it is **8 in both**.

The upper-bound proof said `0x300ba70000` could not be a regression, and the matched pair agrees with
it by a stronger route than variance. Recorded at this length because an unmatched run-length
comparison produced three false readings from one run pair, in both directions, and each looked like
a result.

**Title-screen visuals are unchanged** at `max_nonblack` 0.0052 against a 0.2824 oracle, because
`0x300e390000` is still dropped at pc412 and `0x30131d0000` at pc50. This PR is recompiler coverage;
it is not a fix for Stray's missing background, and the docs say so where the claim would otherwise
be made.

## Docs

- `docs/STRAY_STATUS.md` — the frame-dissection row for `0x300e390000` is marked superseded; a new
  subsection decodes the instruction, names the cause, states plainly that the program still does not
  compile, and records the offline `shader_inspect` route including the launch-shape variables
  (`PROSPER_SHADER_INSPECT_WAVE_SIZE` / `_NATIVE_SUBGROUP` / `_USER_SGPRS` / `_TGID`) without which
  the tool declines at sites the live translation never reaches. Two `## Ruled out` rows.
- `docs/RECOMPILER_REMAINING.md` — three `## Ruled out` rows: the cross-title form of the original
  misreading, plus the two this PR's own review falsified (that making the token set precise is safe,
  and that a synthetic arm can discriminate the KILL).
- `docs/GAME_COMPAT_ORCHESTRATION.md` — **instrument trap 255**, for the apparatus error behind the
  withdrawn claims: two live arms of different run lengths, compared program-by-program, produced
  three false readings from one comparison and the sign was not predictable in either direction.
- `BLOG.md` — a short no-picture entry.

Also fixed from review: the source comment no longer calls GTA V's kernel a carrier of the *defect*
(it carries the **pattern**; it rejects on an earlier MUBUF and never reaches `load_state`), the
`is_end` justification had UPPER and LOWER swapped and now gives the correct argument — a save past
an early `s_endpgm` is unreachable, so no `successors` edge leads to its block and the walk never
visits it — and the token-origination claim now accounts for `join_entry_m0` inserting as well as
erasing.

Docs gate: `check_numbered_table.py` green on both docs, and `check_merge_result.py` reports a clean
merge whose result passes against `origin/main`.


mattias800 (commented):
## Independent review — PR #3312 @ `ea1e45bab`

**REJECTED** — one blocking finding. The diagnosis is right, the entry-block narrowing is right, and
I want this to land. What I cannot approve as written is the **safety proof**, which is recorded in a
code comment and in two `## Ruled out` rows and is therefore the part that will be inherited without
re-derivation. Its second leg is false, one of the change's three components has no test at all, and
the two together are exactly the shape the charter warns about. The remedy is small — see
"What would make this APPROVED".

No GPU was used. Everything below is static, read against `ea1e45bab` blobs, plus one CPU-only scan
of local shader dumps. I did not rebuild; I am not disputing the author's rc=0 / 355 tests.

---

### BLOCKING — the KILL rule can convert a loud reject into a silent fabricated zero, which the proof says is impossible

The proof in the PR body and at `prosper/src/gpu/recompiler/rdna2_emit_cfg.cpp:3709-3712` rests on
two legs. The first is sound; the second is not.

> A token has exactly two effects: it makes `operand_bits` reject a data read (**removing it can only
> turn a reject into a success**) [...]

"Success" here is not a safe outcome. Trace what a removed token actually produces at a dispatcher
edge:

1. `load_state` sets `state.sreg[X] = b.load_function(...)` for **every** register in `sv`
   (`rdna2_emit_cfg.cpp:3699`), and the new code only skips the `state.sreg.erase(X)` when the token
   is absent (`:3719-3722`).
2. `save_state` stores **`zero`** for any written register with no `state.sreg` entry
   (`rdna2_emit_cfg.cpp:3860-3863`).
3. `operand_bits` consults `rs.sreg` before anything else (`rdna2_alu_support.hpp:204-206`), so the
   read resolves to that zero with `ok` left **true**.

That is precisely the failure `join_entry_m0`'s own comment says must not be taken
(`rdna2_to_spirv_internal.hpp`, "*the tempting third option — drop the register to plain untracked —
is the one that must NOT be taken [...] it would fabricate the same zero silently instead of
rejecting*"), and it is the regression the file's own mutation table already records:
`state.sreg.erase(reg)` with no `sreg_entry_m0.insert(reg)` reddens
`test_entry_m0_dispatcher.cpp:271`.

**So the change is only safe if KILL never fires where the emitter would have kept the token live.
It does not hold.** The emitter's token-death rule is `entry_m0_live` = *token AND untracked*
(`rdna2_to_spirv_internal.hpp:4921-4923`), i.e. a token dies only when a write leaves a **value**.
The one explicit erase of a register's token in the whole recompiler is
`rdna2_emit_alu.cpp:1354`, in the SOP1 handler; `record_scalar_write` never touches
`sreg_entry_m0` at all. `rdna2_alu_support.hpp` states the consequence deliberately: "*A write that
deliberately erases its `sreg` value instead — a VOPC mask destination, `s_getpc_b64`, any
unrepresentable write — leaves a STALE token behind [...] That over-rejects rather than fabricates,
and it is left as the safe direction.*"

The new KILL at `rdna2_emit_cfg.cpp:1959-1962` fires on **every** `for_each_scalar_write` report
(`rdna2_to_spirv_internal.hpp:4812-4827`), which includes those value-less writes. Two forms I
verified end to end:

* **VOPC with an SGPR destination.** `for_each_scalar_write` reports it; `record_scalar_write`'s
  `invalidate_mask_pair` (`rdna2_to_spirv_internal.hpp:4973-4986`) erases `rs.sreg[base]` and
  `rs.sreg[base+1]` and leaves `sreg_entry_m0` alone.
* **The B32 mask logical family** (`s_and_b32`/`s_or_b32`/… in the mask domain):
  `rdna2_emit_alu.cpp:1703-1707` does `rs.sreg_bool[dst] = r; … rs.sreg.erase(dst);` — value-less,
  token untouched. `scalar_write_width`'s SOP2 default reports width 1, so the dataflow kills it.

Concrete failure scenario (no title needed):

```
  block A:  s_mov_b32 s0, m0          ; GEN s0
  block A:  s_and_b32 s0, s2, s3      ; mask-domain write: sreg[s0] erased, token STILL LIVE
                                      ;   -> new dataflow KILLs s0
  block B:  v_add_nc_u32 v1, s0, v0   ; data read of s0
```

Before this PR: block B's entry re-stamps the token, the read rejects loudly at pc=B.
After this PR: `entry_m0_in[B]` has no `s0`, `sv[s0]` was stored as `zero`, and the read **compiles
to 0**. The architectural word there is a mask, not zero, and nothing announces it.

**This is not hypothetical inside the very program the PR is about.** #3311's own disassembly gives
`303: s_and_b64 s[14:15], s[58:59], vcc` and `319: s_mov_b32 s74, s14` — s14 is the token-bearing
register, written value-lessly at 303 and read as data at 319. Post-fix the live reject is at
**pc412**, i.e. pc319 resolved; pre-fix the entry-block stamp would have rejected there. I did not
trace `s_and_b64`'s exact emit path, so I state this as a hypothesis with a named discriminator —
but if it holds, part of the headline "pc4 → pc412" is bought by a fabricated zero rather than by
removing a false rejection, and that is a materially different claim from the one the PR makes.

**Why blocking rather than a note.** Nothing in the corpus newly compiles (see finding 4), so no
miscompile ships today. What ships today is the *proof*, into `rdna2_emit_cfg.cpp:3709-3712`,
`docs/RECOMPILER_REMAINING.md`'s Ruled-out row and `docs/STRAY_STATUS.md`'s. The charter's own
worked example is a citation-bearing claim propagating unchecked through four artifacts; a Ruled-out
row asserting that this class of change "cannot fail" is that shape exactly, and it is load-bearing
for #3311, which will make this program compile.

**And the component at issue has zero coverage.** Neither arm D nor arm E, nor any of the four #3203
arms, nor any control, writes the tokened register between the save and a read
(`test_entry_m0_dispatcher.cpp:135-206`). GEN is tested; JOIN is tested (the pre-existing arm's
`s_cbranch_scc0` skips the save, so a MUST join would let it compile — good, credit where due);
**KILL is tested by nothing.** So "the two signatures are disjoint […] neither arm is vacuous" is
true, but "the mutation arms bracket the change" is not: neither mutation moves the KILL rule.

#### What would make this APPROVED

Either of these, plus the doc/comment correction:

* **(a)** Narrow KILL to writes that actually publish a value — i.e. mirror `entry_m0_live` by
  excluding the value-less forms (`vopc_is_cmpx`-negative VOPC SGPR destinations,
  `vop3_writes_mask_sdst`, `s_getpc_b64`, the mask-domain SOP2 forms). This restores "KILL ⊆
  emitter's token death" and makes the second leg of the proof true as stated. It will likely move
  Stray's reject from pc412 back to ~pc319 — that is the **honest** verdict (prosper genuinely
  cannot represent a mask half as a word there; it is the same #3311 family), and the PR body should
  say so instead of claiming pc412.
* **(b)** Keep KILL as is, but add the fixture above to `test_entry_m0_dispatcher.cpp` and report what
  it does. If it compiles, that is the fabricated zero demonstrated in a unit test with no GPU, and
  the proof text, the code comment and both Ruled-out rows must be rewritten to say what the change
  actually guarantees ("cannot newly reject") and what it does **not** ("can newly accept a read
  whose register holds no representable value"). If it still rejects, I am wrong and you have the
  arm that proves it.

Either way, please state the honest scope in the two Ruled-out rows — they are what the next agent
reads first.

---

### Non-blocking

**N1 — `rdna2_emit_cfg.cpp:1948`: "the entry block's in-set stays EMPTY by construction" is false.**
If block 0 has a predecessor (a back edge to the entry block, which is legal here — `add_successor`
at `:1908-1921` will happily add block 0), the union join grows `entry_m0_in[0]`. The **code** is
right and the result is conservative; the **comment** is not, and it is written as a proof step
someone will later lean on.

**N2 — `rdna2_emit_cfg.cpp:3710-3712`: the terminal fallback described there is dead code.**
"the terminal call when `end_pc` maps to no block" never happens: `start_set.insert(end_pc)` at
`rdna2_emit_cfg.cpp:1763` guarantees `block_for_pc.find(end_pc)` succeeds. So the terminal
`initial = load_state()` at `:5976` always uses `entry_m0_in[end_block]` — the union over paths that
**enter the terminal block**, which misses waves that left via `s_trap` (`cfg_terminator`
`opcode == 0x12`, `:1251-1257`, `:4947-4951`) or via a `proven_exit_target` branch (`:4970-4990`),
both of which set `active_var = no` without reaching it. Scenario: a barrier-split compute shader
whose phase 1 saves M0 into sX on a path ending in `s_trap`; phase 2's `initial.sreg[X]` is then a
fabricated zero where pre-fix it carried the token. One-line fix if you want it: use
`entry_m0_may_hold` unconditionally when `dispatch == UINT32_MAX` — the fix does not need to change
that path. Otherwise, correct the comment.

**N3 — the subset proof does not literally cover the GEN loop.** `entry_m0_may_hold` **breaks** at
the first `is_end` (`:1813`); the new GEN **continues** past it (`:1957`). So GEN is applied to a
strictly larger instruction set than the one that built `entry_m0_may_hold`, and
`entry_m0_in[B] ⊆ entry_m0_may_hold` does not follow from "the only insertion is GEN". I convinced
myself it is harmless — any block whose range extends past the first `is_end` is either the terminal
block, whose own GENs are never consumed (it has no successors, and only its *in*-set is read), or a
block above `end_pc`, which is unreachable because `add_successor` filters `pc <= end_pc`, so the
fallback applies. But that argument is not in the PR and it should be, or the predicate should just
match.

**N4 — the sweep's denominator is almost entirely inert, and its population under-counts by one
title.** 149 of the 152 dumps have an empty `entry_m0_may_hold`, so `:1947` short-circuits the whole
walk and `load_state` falls back to the same empty set — both builds are *structurally* identical on
them. The sweep's real N over this change is 3, not 152. That is fine as a regression check; it is
not the "proof, not a sample" the PR calls it.

I scanned every `.bin` under the developer home for the SOP1 `s_mov_b32 sX, m0` word pattern
(`w & 0xFF80FFFF == 0xBE80037C`, SDST ≤ 105) — 275 files, CPU only. Three carry it: the two the PR
names, **plus `~/gta5-work/lighting.bin`**, which the `exec_*`/`success_*` naming filter excluded.
It is a real RDNA2 stream (2935 dwords, `s_endpgm` = `0xbf810000` at 2934, `s_inst_prefetch` at pc0),
it saves into **s6** at pc82/442/518/1442/1640 and **s7** at pc333 — and its **pc1 is
`d7460005 04010c06` = `v_lshl_add_u32 v5, s6, 6, v0`**: the byte-identical entry-block defect, in a
third title, and that title is *Grand Theft Auto V*, whose compute programs are the live frontier
(#2481). Worth adding to the sweep and to the PR body; it strengthens the case. (I did not confirm
it routes through the CFG dispatcher.)

Also worth stating plainly, since the brief asked: **a verdict diff cannot see a wrong
compilation.** It is bounded here only by the fact that 0 dumps changed from reject to accept —
both movers still reject — so no miscompile is *realised* on the corpus. The risk in the blocking
finding is structural and unrealised, not measured.

**N5 — `descriptor_sgprs` are the one launch class the entry-block fix does not serve.**
`rdna2_to_spirv.cpp:2470-2473` puts inline-descriptor user SGPRs in `sreg_input`, not `sreg`, so
their `sv` slot is initialised to zero (`rdna2_emit_cfg.cpp:3569-3572`) and `load_state`'s
`state.sreg[X] = load_function(...)` (`:3699`) shadows `sreg_input`. If an M0-save destination is
also a descriptor SGPR, the entry-block read now returns 0 where pre-fix it rejected. Narrow and
speculative — descriptor ranges are unlikely M0 scratch — but it is the same mechanism as the
blocking finding and belongs in the same sentence when the scope is written down.

---

### Checked and cleared

* **The fix delivers the right value, not another zero.** This was the thing most worth doubting and
  it holds: `tgid_x` lands in `initial.sreg` (`rdna2_to_spirv.cpp:2481`), the prologue stores that
  into `sv` (`rdna2_emit_cfg.cpp:3569-3572`), so Stray's entry-block read of s14 resolves to
  `groupid[0]` and not to a fabricated word.
* **The union join is sufficient across back edges, irreducible loops, and a save inside a loop whose
  header precedes it.** Worked the fixpoint by hand on the header-before-save case: the back edge
  grows the header's in-set and the worklist re-pushes it, so the header rejects on the second
  iteration and the pre-loop read still resolves. Correct, and conservative in the right direction.
  Termination and monotonicity are fine (sets only grow; every push follows a strict growth or a
  first reach).
* **Unreached-block fallback is conservative in the rejecting direction** and cannot mask a needed
  restore: it is `entry_m0_may_hold`, a superset of anything the walk could produce, i.e. exactly
  #3133. (See N2 for the one path where the comment claims this and the code does not do it.)
* **`successors` is a complete over-approximation of the emitter's runtime transitions.** Checked
  every `store_function(pc_var, …)` — `:4017` via `set_next`, `:4985`, `:4988` — plus the portable
  vote's `vote_taken_var`/`vote_next_var` (`:5015-5016`); each uses a dispatch derived from a block
  `add_successor` also added. Targets past `end_pc` are either `proven_exit_target` deactivations or
  outright rejects. So no unmodelled edge can drop a token.
* **KILL ordering (kill before gen) is right**: `for_each_scalar_write` reports the save's own SOP1
  destination at width 1, so gen-first would erase the token it had just created.
* **`for_each_scalar_write`'s *misses* are all safe.** VOP2's implicit VCC write, `s_swappc_b64`'s
  pair (SOP1 default width 1 where 2 is architectural) — all missed kills, i.e. surviving tokens, the
  over-rejecting direction. And GEN's `dst.value <= 105` means 106/107/124/126/127 are never tokened
  here, so their handling is moot.
* **The block-range idiom and worklist match the neighbouring analyses** (`:2049-2052`, `:2904-2905`,
  and `scalar_may_write_in` at `:3148-3171`, which is the existing MAY/union precedent over the same
  CFG). `starts` is ascending — `std::set<uint32_t>` at `:1568`. Bounds are safe: `entry_block` is
  either `UINT32_MAX` (short-circuited) or a valid block index.
* **Blast-radius claim verified**: `:1947` short-circuits on an empty MAY set, so a shader with no
  `s_mov_b32 sX, m0` is bit-identical to before.
* **Cost**: O(N) per pop with an O(B·106) growth bound, and only on shaders that contain a save. Fine.
* **Docs**: no deletions beyond the one superseded table row, which is rewritten rather than dropped.
  `STRAY_STATUS.md`'s "clearing this does not render the background" row is the right instinct and I
  am glad it is there.

---

Verdict: **REJECTED**, on the blocking finding only. Re-review on the corrected head will be quick —
I expect (a) or (b) plus the two Ruled-out rows, not a redesign. The entry-block diagnosis is
excellent work and the offline `shader_inspect` route in `STRAY_STATUS.md` is worth more than the fix.

mattias800 (commented):
Thank you — the blocking finding is correct and I have taken the first remedy, not the second.

## The blocking item: the KILL is gone, not narrowed

You are right that "removing a token can only turn a reject into a success" treats success as safe
when it is not. I confirmed all three legs of your mechanism (`:3699`, `:3860-3863`,
`rdna2_alu_support.hpp:204-206`), and that the emitter's only token erase is the SOP1 generic path at
`rdna2_emit_alu.cpp:1354` with `record_scalar_write` never touching `sreg_entry_m0`.

I did not narrow KILL to value-publishing writes, because that predicate does not exist and building
one under time pressure under-approximates in the **unsafe** direction — SOP1 alone is wrong
(`s_mov_b64`, the saveexec family and `s_getpc_b64` are all SOP1 and all value-less), and every
enumeration I tried had the same hazard. So I removed KILL entirely. It was never needed: the defect
is at the entry block, where the in-set is empty because nothing has run.

Both bounds are now unconditional, and the LOWER one is the one you were missing:

- **UPPER** — subset of `entry_m0_may_hold` (GEN's predicate builds it), so it cannot newly reject.
- **LOWER** — superset of whatever token the emitter can hold live, because the emitter creates
  tokens only at a save and a union over every path from every save covers them all. **No token the
  emitter would keep is ever dropped**, so the fabricated-zero hazard is structurally impossible.

Measured cost of the lost precision: none. The live launch shape still reaches pc412 with 11632 words
emitted, identical to the KILL version.

## Your "KILL has zero coverage" point, and where it led

I added arm F and then could not make it discriminate — **four placements, none works**, and I have
recorded that in the arm's own comment rather than shipping it as a guard it is not:

- write and read in one block → the block-entry set is never consulted;
- write behind one arm of an if → the union join restores the token;
- write unconditional, read in a later block → the KILL *does* drop the token, and the **B64 mask
  domain refuses the read anyway**;
- `s_getpc_b64`, to get a value-less write that is not a mask → the shell refuses `op=0x1f` at the
  write itself (pc=6), so the read is never reached.

Every *accepted* value-less scalar write is a mask write, and the mask domain guards those reads
independently. So your hazard is real but reachable only through the **gap** in that MUST/equality
join. Arm F is now labelled a regression pin that explicitly says it cannot pin *why*. Both
falsifications are in `RECOMPILER_REMAINING.md` § Ruled out.

## Your two strengtheners

1. **Sweep re-scoped — you were right and it was worse than you said.** Rescanned 16,560 local `.bin`
   files: **938** carry the save, and only **102** of those even reach the CFG dispatcher. A/B over
   all 938: 936 identical, 2 moved later, **0 earlier, 0 accept→reject, 0 reject→accept**.

2. **GTA V — I could not confirm it, so I am not claiming it.** `lighting.bin` has the pattern (saves
   into s6/s7, pc1 is `v_lshl_add_u32 v5, s6, 6, v0`), but its verdict is **identical in both
   builds**: offline it lowers through the straight-line emitter and rejects at pc6 on a MUBUF with
   no resource table, so it never reaches `load_state`. Carrier of the pattern, not a demonstrated
   carrier of the defect. The PR says two titles, not three.

## Non-blocking, all three taken

- The back-edge comment now states the guarantee that actually holds ("no save has executed on
  entry"), not "empty by construction".
- The terminal `load_state()` no longer narrows at all — you are right that the fallback was dead
  code, and that `s_trap` / `proven_exit_target` edges are not modelled by this walk, so it takes the
  whole-stream set explicitly.
- The `is_end` asymmetry is now stated in the source: GEN `continue`s where the MAY-set scan `break`s,
  which makes the set no smaller past an early `s_endpgm` — the direction the UPPER bound needs.

Build rc=0, `ctest --no-tests=error` rc=0 at 355/355, `test_entry_m0_dispatcher` rc=0 at 16/16, all on
`4b46472a3`. Not merging; ready for re-review.

mattias800 (commented):
Re-review requested on **`f9be0a494`**. Two changes since the REJECTED review on `ea1e45bab`, plus a
correction to my own live evidence.

## 1. The blocking finding — KILL removed, not narrowed

Verified all three legs of your mechanism (`:3699`, `:3860-3863`, `rdna2_alu_support.hpp:204-206`) and
that the emitter's only token erase is the SOP1 generic path at `emit_alu:1354`, with
`record_scalar_write` never touching `sreg_entry_m0`.

I did **not** narrow KILL to value-publishing writes. That predicate does not exist, and the obvious
approximation is wrong in the unsafe direction: `s_mov_b64`, the saveexec family and `s_getpc_b64`
are all SOP1 and all value-less, so "SOP1 only" still drops tokens the emitter keeps. KILL is gone
instead. It was never needed — the defect is at the entry block, where the in-set is empty because
nothing has run — and both bounds are now unconditional. The LOWER one is the one you were missing:
the emitter creates tokens only at a save and erases them in one place, so a union over every path
from every save is a superset of whatever it can hold live. **No token the emitter would keep is ever
dropped**, which makes the fabricated-zero hazard structurally impossible rather than unobserved.

Measured precision cost: none. Live launch shape still reaches pc412 with 11632 words emitted.

## 2. Your coverage point — arm F cannot discriminate a KILL, and says so

Four placements tried, none isolates it: same block (entry set never consulted), behind one if arm
(union join restores the token), unconditional with a later-block read (KILL drops the token and the
**B64 mask domain refuses the read anyway**), and `s_getpc_b64` for a value-less non-mask write (the
shell refuses `op=0x1f` at the write itself). Every *accepted* value-less scalar write is a mask
write, and the mask domain guards those reads independently — so the hazard is reachable only through
the gap in that MUST/equality join.

Arm F is labelled a regression pin whose comment states it cannot pin *why*, and the mutation table
records that re-adding KILL leaves **every arm green**. Both falsifications are in
`RECOMPILER_REMAINING.md` § Ruled out. I would rather ship that than an arm that passes for a reason
it does not name.

## 3. My live evidence was wrong in both directions — withdrawn and replaced

A length-matched rerun (12 frames per arm, same route and warmup, back to back) corrects the two live
claims the earlier unmatched comparison produced:

| | `origin/main` | this PR |
|---|---|---|
| skipped | `0x300ba70000`, `0x300e390000`, `0x30131d0000` | `0x300e390000`, `0x30131d0000` |
| reject PCs | pc=3, pc=4, pc=50 | **pc=412**, pc=50 |
| `[mimg-unresolved]` | 8 | 8 |

- `0x300ba70000` is **not** newly skipped by this branch — it is skipped on `origin/main` at pc=3 and
  **not skipped here**. A second program this change fixes, not a regression needing the proof to
  excuse it.
- `0x30131d0000` is **not** fixed. pc=50 in both arms. Withdrawn.
- The `[mimg-unresolved]` "10 → 4" was the same artifact; 8 in both.

Recorded as **instrument trap 255**: two live arms of different run lengths, compared
program-by-program, produced three false readings from one comparison and the sign was not
predictable — the alarming and the flattering direction were both wrong.

## Also from your review, all three non-blocking items

- Back-edge comment now states the guarantee that holds ("no save has executed on entry").
- Terminal `load_state()` no longer narrows at all — the fallback was dead code as you said, and
  `s_trap` / `proven_exit_target` edges are unmodelled by this walk, so it takes the whole-stream set.
- The `is_end` asymmetry is argued in the source: GEN `continue`s where the MAY-set scan `break`s,
  which makes the set no smaller past an early `s_endpgm` — the direction the UPPER bound needs.

## Verification on `f9be0a494` (rebased onto current `origin/main`)

Build rc=0; `ctest --no-tests=error -j4` rc=0 at **355/355**; `test_entry_m0_dispatcher` rc=0 at
**16/16**; docs merge-result gate clean. Sweep re-scoped as you asked: 16,560 local `.bin` files,
**938** carrying the save, **102** reaching the dispatcher, **936 identical / 2 later / 0 earlier /
0 accept→reject**. GTA V declined as a carrier of the pattern but not a demonstrated carrier of the
defect.

Not merging.

mattias800 (commented):
## Independent re-review — PR #3312 @ `f9be0a494` (previous review: `ea1e45bab`, REJECTED)

**REJECTED**, one blocking finding, and it is a *smaller* ask than last time — a comment and an issue
would satisfy me. Everything I raised on `ea1e45bab` is addressed, several things better than I asked
for. The remedy chosen is not the one I suggested and I think it is the **better** one: my proposed
"narrow KILL to value-publishing writes" was reaching for a predicate that does not exist, and the
author is right that an SOP1-shaped approximation is wrong in the unsafe direction
(`s_mov_b64`, the saveexec family and `s_getpc_b64` are all SOP1 and all value-less). Removing KILL
is correct.

Static only, read against `f9be0a494` blobs. No GPU, no rebuild; I am not disputing rc=0 / 355 / 16.

### Does removing KILL close my hazard, or relocate it?

**It closes it inside a dispatcher region, and relocates it to the region boundary.**

The LOWER bound is sound where it is stated to apply. I checked the reduction rather than the
wording: within one region the emitter's token set at a block entry *is* `entry_m0_in[B]` by
construction (`load_state` rebuilds it, `rdna2_emit_cfg.cpp:3748-3753`), so the real question is
whether `entry_m0_in[B]` covers the **architectural** may-hold — "some save to X executed on the path
taken". A GEN-only union over `successors` is exactly the path-union of saves, and I verified last
round that `successors` over-approximates every runtime transition. So no token the emitter could be
carrying is dropped. That much is a proof, and it is the right one.

### BLOCKING — the LOWER bound is written as unconditional, and it is not: a barrier phase's block 0 inherits tokens the walk does not model

`rdna2_emit_cfg.cpp:1948-1952`:

> LOWER `entry_m0_in[B]` contains **every** register whose token the EMITTER could hold live on entry
> to B. The emitter creates a token only at a save (`rdna2_emit_alu.cpp:1337`) […] **it never creates
> one anywhere else.**

and `:1971-1975`: "*The guarantee is 'no save has executed on entry', not 'this set is empty'*".

Both fail for block 0 of a **non-first barrier phase**, where saves *have* executed on entry — the
previous phase's. The chain, every link opened:

1. `emit_body` splits a barrier-phased compute shader and calls `emit_cfg_state_machine(b, rs, phase,
   …)` **per phase with the same `rs`** (`rdna2_emit_cfg.cpp:6140-6144`; `initial` is a `RegState&`).
   This is not a theoretical path — `RECOMPILER_REMAINING.md`'s own #3231 row states that *Astro
   Bot*'s `0x500571000` has "**two earlier phases each exited through the dispatcher**".
2. Phase N's terminal export (`:6007`) runs `load_state()` with `dispatch == UINT32_MAX`, so
   `narrow == false` and `entry_m0_here` is the whole-stream `entry_m0_may_hold` — your N2 fix, and
   right. For each X it does `initial.sreg.erase(X)` and `initial.sreg_entry_m0.insert(X)` (`:3750-3752`).
3. Phase N+1's `load_state` builds a fresh `RegState` and copies `max_vgpr`, `sreg_input`, the SMEM
   descriptor fields, `reads_scc`, `invalidated_vgpr_lane_slots` — and **not `sreg_entry_m0`**
   (`:3709-3721`; the only write to `state.sreg_entry_m0` in the whole file is `:3752`).
4. Phase N+1's prologue: `sv[X] ← initial.sreg.find(X)`, absent after step 2, so it stores **`zero`**
   (`:3596-3599`).
5. Phase N+1's block 0: `entry_m0_in[0]` is empty, X is not stamped, so `state.sreg[X]` is the loaded
   zero (`:3726`) and a read of X resolves to **0 with `ok` true**.

Before #3308, step 5 stamped `entry_m0_may_hold(phase N+1)`, so the read **rejected loudly** whenever
phase N+1 also contains `s_mov_b32 X, m0`. That case is a regression introduced here; the case where
phase N+1 does *not* save into X is a pre-existing hole (step 3 has always dropped the inherited
token) that #3308 removes the accidental cover for.

Reachability needs three things, none observed together but each individually ordinary: a
barrier-split compute shader on the dispatcher path (observed — Astro Bot), the same register used as
an M0 save destination in two phases (the borrow idiom clusters around LDS, which is what barriers
are *for*, and Stray's own kernel already saves into s14 twice, at pc157 and pc274), and a read of it
in the later phase's entry block. Note step 2 stamps the whole-stream set, so the token does **not**
need to be genuinely live at the boundary — the conjunction is weaker than it first looks. And the
938-carrier sweep cannot see this: a newly-accepted read only moves a reject PC if it happens to be
the first one.

**Why I am blocking on a hazard this narrow, again.** The behaviour is a strict improvement over
`origin/main` everywhere except that conjunction, and I would not block on the behaviour alone. I am
blocking on the *text*: "both are unconditional", "it never creates one anywhere else", and a
`## Ruled out` row that a later agent will inherit as checked. That is what I rejected `ea1e45bab`
for, and the correction is one I want to be true, which per the charter is the reason to check it
harder rather than wave it through.

**Any of these flips me to APPROVED:**

* **(a) Cheapest, and enough.** Qualify the LOWER bound: it holds **within one dispatcher region**;
  `load_state` does not inherit `initial.sreg_entry_m0` (`:3709-3721`), so a region entered from a
  previous region starts with no inherited token, and #3133's whole-stream stamp was what covered
  that. Say the same in the Ruled-out row, add a line to the terminal-case comment at `:3739-3745`
  (which is where the export that feeds it lives), and file the issue.
* **(b) Closes the regression half, and cannot newly reject anything.** Seed block 0 with
  `{r ∈ initial.sreg_entry_m0 : !initial.sreg.contains(r)} ∩ entry_m0_may_hold` — the intersection is
  load-bearing, since it reproduces exactly what pre-#3308 stamped there and so keeps the UPPER bound
  intact. Seeding the *full* inherited set would close the pre-existing half too, but that can newly
  reject and needs its own evidence; don't fold the two together.
* **(c)** Show me I am wrong about step 1 or step 3.

### Your coverage point — the arm F disclosure is adequate, and better than a green guard

Yes, I endorse it. Four placements with four distinct reasons, the conclusion that every *accepted*
value-less scalar write is a mask write and the mask domain guards those reads independently, the
arm labelled a `pc=10` regression pin whose comment says it cannot pin *why*, and — the part that
makes it honest — **the mutation table recording that re-adding the KILL leaves every arm green**.
That last line is the one that stops a future reader mistaking arm F for a KILL guard, and it is
recorded in the place they will actually be standing. Shipping a green-but-vacuous discriminator here
would have been worse than shipping none; this is the right call, and both falsifications being in
`## Ruled out` rather than only in the PR is what makes it stick.

One qualification to record with it: the "reachable only through the gap in the mask domain's
MUST/equality join" conclusion rests on that enumeration of value-less writes being complete, and
`s_getpc_b64` was excluded by *this shell* refusing `op=0x1f` at the write, not by a general
argument. That is fine as a bound on the fixture; it is not a proof that no non-mask value-less write
is accepted anywhere. Worth one clause, since the KILL is now excluded by argument rather than by
test.

### Non-blocking

**N6 — `rdna2_emit_cfg.cpp:1937-1938` still over-claims GTA V, contradicting your own withdrawal.**
The comment says "*Grand Theft Auto V and The Plucky Squire each have a kernel with the byte-identical
**defect***". The PR body withdraws exactly that: "*a carrier of the **pattern**, not a demonstrated
carrier of the **defect***". The body is scrupulous and right — and it is the ephemeral artifact,
while the comment is the durable one. Since I am the one who brought `lighting.bin` and would like it
to count, I'll be explicit: the withdrawal is correct and the comment should read "the byte-identical
**shape**", or name Plucky Squire alone. Same for the fixed `STRAY_STATUS.md` prose if it repeats it.

**N7 — the `is_end` justification at `:1983-1988` has the two bounds swapped.** It says skipping
`is_end` "makes this set no smaller […] which is the direction the **UPPER** bound needs". No smaller
is the direction the **LOWER** bound needs; UPPER (`⊆ entry_m0_may_hold`) is the one a *larger* set
would violate, and past an early `s_endpgm` this loop can GEN what the whole-stream scan's `break`
(`:1813`) never saw. The code is fine and the conclusion is right, but for a different reason: any
block whose range extends past the first `is_end` is either the terminal block — whose own GENs are
never consumed, since it has no successors and only its *in*-set is read — or a block above `end_pc`,
which `add_successor`'s `pc <= end_pc` filter makes unreachable, so the fallback applies. That is the
argument; please put it there instead.

**N8 — the LOWER bound's enumeration of token sites is not exact.** `join_entry_m0`
(`rdna2_to_spirv_internal.hpp:4957-4964`) both **inserts** (`:4960`) and **erases** (`:4963`), so
"creates a token only at a save […] erases one in only one place" is wrong as written. It happens not
to matter — the insert only fires when an edge already had a live token, and the erase only when
neither did, so neither can create ex nihilo nor drop a live one — but the whole proof now rests on
that enumeration being complete, so it should be complete: three insert sites (`emit_alu:1337`,
`join_entry_m0:4960`, `load_state:3752`) and three erase sites (`emit_alu:1350` for M0 itself,
`:1354`, `join_entry_m0:4963`), with one word each on why only the first insert creates.

### Checked and cleared since `ea1e45bab`

* **N1 fixed** — `:1971-1975` now states the back-edge case correctly, and the union does restore the
  token there.
* **N2 fixed, and fixed properly** — `:3746-3747`'s `dispatch != UINT32_MAX` makes the terminal case
  unconditionally non-narrowing, and the comment names `s_trap` and `proven_exit_target` as the edges
  the walk does not model rather than leaning on a dead `block_for_pc` fallback. This is the right
  shape. (It also feeds the blocking finding above — a larger export means more inherited tokens for
  the next phase to drop — which is an argument for seeding block 0, not against this fix.)
* **N4 taken, and the honest version is worse than mine** — 16,560 files, 938 carriers, 102 reaching
  the dispatcher, and the old "152 dumps" correctly retired as trivially-identical denominator. 0
  earlier / 0 accept→reject / 0 reject→accept over all 938.
* **My GTA V strengthener correctly withdrawn on evidence** (body only — see N6).
* **UPPER bound re-verified after the KILL removal**: still holds for every *consulted* set, by the
  N7 argument.
* **Trap 255 and the length-matched table** are a genuinely better contribution than the fix. A
  correction that moves in both directions, published against the person who supplied the original
  numbers, is the hard kind.
* Docs: both falsifications in `## Ruled out`; `STRAY_STATUS.md`'s superseded census row corrected
  rather than deleted; blog entry no longer over-promises.

Verdict: **REJECTED** on the LOWER bound's stated scope only. (a) is a comment plus an issue and I
will re-approve on it immediately.

mattias800 (commented):
Re-review requested on **`845d396ba`** (rebased onto current `origin/main`; two lanes landed in
`BLOG.md`, `GAME_COMPAT_ORCHESTRATION.md` and `STRAY_STATUS.md` and all three were resolved by
keeping both sides — verified 0 deleted lines from the other lanes on the first two, and the only 2
deletions in `STRAY_STATUS.md` are my own superseded rows).

## The blocking finding — verified link by link, then fixed in code as well as text

I opened every site before acting, because this lands in a `## Ruled out` row. All five links hold:
`emit_phase` calls `emit_cfg_state_machine` per phase with the same `RegState&` (the parameter is
`RegState& initial`); the region exit is `initial = load_state()`, a whole-struct assignment;
`load_state` default-constructs `RegState state` and copies **seven** named fields with
`sreg_entry_m0` not among them; the prologue's `sv` loop stores `zero` for any register absent from
`initial.sreg`; and `operand_bits` returns that with `ok` true.

I took **both** remedies rather than (a) alone, because the regression case is real:

- **(b)** block 0 is seeded with `{X ∈ entry_m0_may_hold(this region) : entry_m0_live(initial, X)}`.
  Your intersection is kept and its role is stated where it lives — it is what preserves UPPER across
  the seed.
- **(a)** both bounds now say *within one dispatcher region*, the source names `load_state` not
  inheriting `sreg_entry_m0`, and the case the intersection drops (saved in an earlier phase, not
  saved again in this one) is **#3314**, noted as predating this change.

## The part you should push back on if you disagree

**The seed has no regression arm, and I removed the one I wrote rather than keep a passing arm.**
I built two fixtures:

- unguarded two-phase kernel + the readlane route → refused outright with
  `portable-readlane-dispatcher-unsafe guest-barrier=1`. A guest barrier and a portable readlane are
  incompatible, and the readlane is how every program in that file reaches the dispatcher.
- unguarded two-phase kernel without the readlane → `emit_body` never takes the phased path at all
  (it needs `phased.guarded || initial_dispatch_active || force_barrier_phases`). The **straight-line**
  emitter handled it and refused the cross-phase read at pc5 from the shared `RegState`'s own token.
  That arm was green with the seed *and* green with the seed deleted — a pass for the wrong reason,
  the same vacuous shape as arm F, so I deleted it.

Reaching the path needs `initial_dispatch_active`: a phased region nested inside a dispatcher, i.e.
the Astro Bot `0x500571000` shape you cited. #3314 asks for that fixture first. The mutation table
now records "remove the seed → every arm stays green", so nobody reads coverage into it.

## Non-blocking, all three taken in the comments — the durable artifact

- **N6** — the source no longer calls GTA V's kernel a carrier of the *defect*. It carries the
  pattern (saves s6/s7, pc1 `v_lshl_add_u32 v5, s6, 6, v0`) and rejects on an earlier MUBUF without
  ever reaching `load_state`. Thank you for flagging it against your own strengthener.
- **N7** — UPPER/LOWER were swapped. Correct argument now given: skipping `is_end` can only make the
  set larger, which is LOWER's direction, and it cannot break UPPER because a save past an early
  `s_endpgm` is unreachable, so no `successors` edge leads to its block and the walk never visits it.
- **N8** — the origination claim now accounts for `join_entry_m0` inserting as well as erasing: it
  propagates an existing token across a structured join and cannot originate one, which is what the
  bound needs.

Your qualification on *"every accepted value-less write is a mask write"* is fair and I have not
generalised it — arm F's comment states it as what **this shell** does with `s_getpc_b64` (refuses
`op=0x1f` at the write), not as a theorem.

## Verification on `845d396ba`

Build rc=0; `ctest --no-tests=error -j4` rc=0 at **356/356**; `test_entry_m0_dispatcher` rc=0;
merge-result docs gate clean. Carrier sweep re-run on the final code: **identical on all 938** to the
previous revision, which is the expected result — every carrier in that corpus is single-region, so
the seed is a no-op there — and the A/B against `origin/main` still reads 936 identical / 2 later /
0 earlier / 0 accept→reject.

Not merging.

mattias800 (commented):
## Independent re-review — PR #3312 @ `845d396ba` (previous: `f9be0a494` REJECTED, `ea1e45bab` REJECTED)

**APPROVED.** Both remedies were taken where I said either would do, and the harder one is
implemented correctly. One correction is required before you merge — it is comment/issue text only,
no behaviour, and needs no further review round; details below, with the replacement sentence.

Static only, read against `845d396ba` blobs. No GPU, no rebuild. I did run the docs gate and the
trap-41 check, since the coordinator asked me to spot-check the rebase — those are the only two
things I executed.

### The seed — verified, link by link

`rdna2_emit_cfg.cpp:2000-2002` and `:2012`:

```cpp
std::set<int> inherited_entry_m0;
for (const int reg : entry_m0_may_hold)
    if (entry_m0_live(initial, reg)) inherited_entry_m0.insert(reg);
…
entry_m0_in.front() = inherited_entry_m0;
```

* **The predicate is the right one.** `entry_m0_live` (`rdna2_to_spirv_internal.hpp:4921-4923`) is
  `token AND untracked`, which is exactly what `operand_bits` asks (`rdna2_alu_support.hpp:204-206`).
  A previous region that left `initial.sreg[X]` set is correctly *not* seeded — the register holds
  data there.
* **The intersection preserves UPPER, and it is written where it lives.** Iterating
  `entry_m0_may_hold` and filtering makes `inherited_entry_m0 ⊆ entry_m0_may_hold` structural rather
  than asserted, so the seeded set cannot stamp outside #3133's set and cannot newly reject.
* **`initial` is unmutated between function entry and the seed** — I grepped every write to
  `initial.*` before line 2010 and there are none, so the seed reads the caller's state as the
  previous region left it.
* **Guard consistency**: `inherited_entry_m0` is a subset of `entry_m0_may_hold`, so the
  `!entry_m0_may_hold.empty()` short-circuit at `:2005` can never skip a non-empty seed. It is used
  in exactly one place (`:2012`).
* **No-op where it must be**: `rdna2_to_spirv.cpp` never sets `sreg_entry_m0` on the top-level `rs`,
  so a single-region shader seeds nothing. That is why the 938-carrier sweep is unchanged, and the
  unchanged sweep is therefore consistent rather than reassuring — correctly, you did not present it
  as evidence for the seed.

### REQUIRED before merge (comment/issue text; no re-review needed)

**The reachability claim is wrong, and it is the claim that closed the fixture search.**
`rdna2_emit_cfg.cpp:1993-1998`, the test file's block, and #3314 all say reaching the multi-region
path "needs `initial_dispatch_active`". It does not. Your own second fixture's failure names the
disjunction correctly — `phased.guarded || initial_dispatch_active || force_barrier_phases`
(`:6133-6134`) — and then treats the third disjunct as unreachable. It is set automatically, for
exactly the shape you were building:

```
:6999   if (!cfg_dispatch_safe) {
:7000       const BarrierPhasedCompute phased = analyze_barrier_phased_compute(ins);
:7001       if (phased.found && !phased.guarded)
:7002-7005      return emit_body(…, /*initial_dispatch_active=*/0,
                                    /*force_barrier_phases=*/true);
```

Follow it through: the gate at `:6133-6134` passes on `force_barrier_phases`, `phased.guarded` is
false on the same analysis, so `emit_phase`'s `if (!phased.guarded || initial_dispatch_active)`
(`:6180`) routes **every phase to `emit_cfg_state_machine` with `initial_dispatch_active == 0`**.
The comment two lines above `:6999` names the trigger in plain terms: *"nested wave branches still
need this dispatcher. If a guest barrier makes that transformation unsafe…"*.

So the ingredient your fixture was missing was not a nested dispatcher and not the readlane — it was
**`!cfg_dispatch_safe`, i.e. a nested wave branch**. Fixture 2 had `cfg_dispatch_safe` true, so
`:6999` never fired, which is precisely why `emit_body` "never took the phased path". That is a
third shape, untried, and it is the one the code itself routes through for an ordinary unguarded
barrier-phased kernel.

Two consequences, and the second is why I am asking rather than noting:

1. #3314's *"A fixture built from that shape is the first thing this issue needs"* points the next
   person at Astro Bot's partial-workgroup shape when the cheaper route is a nested wave branch plus
   a barrier plus two unguarded phases.
2. It understates how ordinary the multi-region path is. An unguarded barrier-phased compute kernel
   whose wave branches force the dispatcher takes it with no partial-workgroup extent involved — so
   the hole #3314 tracks sits on a common route, not an exotic one.

Suggested replacement, all three sites: *"Reaching it needs a phased region that also reaches the
dispatcher: `phased.guarded && initial_dispatch_active`, or — the cheaper route — `!cfg_dispatch_safe`
(a nested wave branch), which sets `force_barrier_phases` at `rdna2_emit_cfg.cpp:7002-7005` and takes
every unguarded phase through `emit_cfg_state_machine` with `initial_dispatch_active == 0`. Neither
fixture tried here had `!cfg_dispatch_safe`; that is the untried shape."*

I have not built a program that takes it, so treat the shape as a lead rather than a verified
fixture. The routing above is verified.

### Your question: is an unarmed seed acceptable here?

**Yes, and this is the second time I think the honest answer beats the green line** — arm F was the
first. My reasons, so the judgement is auditable rather than a nod:

* The seed is a two-line predicate whose every link I could check statically, and did. Coverage buys
  least where the mechanism is small and fully readable; it buys most where behaviour is emergent.
  This is the former.
* It is provably a no-op on all 938 carriers, so there is no regression it could have caught in
  anything measured.
* Deleting an arm that was green *with* and *without* the thing it claimed to guard is the right
  call — a green line asserting coverage it does not have is worse than an admitted gap, and this
  file has now made that call twice with the mechanism named both times.
* The disclosure is in the three places a reader will actually stand: the source next to the seed,
  the test file above the mutation table, and #3314.

But that acceptance is *conditional on the disclosure being accurate*, which is exactly why the
correction above is required rather than optional. An unarmed mechanism is defensible when the record
says why nobody could arm it; it stops being defensible when the stated reason is one the code
contradicts, because that is what stops the next attempt too.

One placement note, no action needed: the "removing the seed reddens nothing" disclosure is in the
prose block at the top of that comment, not as a row in the mutation table where the other five live.
Both work; the table is where a reader diffing signatures will look.

### N6 / N7 / N8 — all correctly taken

* **N6** — `:1937-1941` now separates them exactly right: Plucky Squire carries the *defect*, GTA V
  carries the *pattern* and "never reaches `load_state` at all". The durable artifact now matches the
  body's withdrawal. Since I supplied that strengthener, I am glad it reads this way.
* **N7** — `:2024-2029` is corrected in the right direction *and* with the right reason: larger is
  LOWER's direction, and UPPER survives because a save past an early `s_endpgm` is unreachable so the
  walk never visits its block. That is the argument, now written down.
* **N8** — `:1951-1957` distinguishes originating from propagating and cites `join_entry_m0`
  (`:4957-4964`) explicitly. Correct: its insert only fires when an edge already had a live token and
  its erase only when neither did.
* You did **not** generalise my value-less-writes qualification, and you were right not to — arm F
  states it as what this shell does with `s_getpc_b64`, which is what was measured.

### Rebase spot-check (trap 41) — clean, verified independently

Against the current merge base `99ecee473` the PR has **six deleted lines in total**, and I read
every one:

| file | +/− | deletions |
|---|---|---|
| `BLOG.md` | +13 / **−0** | none — no other lane's entry touched |
| `GAME_COMPAT_ORCHESTRATION.md` | +1 / **−0** | none — trap 254 intact, 255 appended |
| `STRAY_STATUS.md` | +82 / −2 | both the author's own superseded rows (the `0x300e390000` frame-dissection row and the `0x300ba70000` census row), each rewritten in place |
| `rdna2_emit_cfg.cpp` | +143 / −4 | the old `for (int reg : entry_m0_may_hold) {…}` stamp |

Docs gate run against the **`origin/main`** baseline (the stricter one): `14 tables, 317 rows,
unbroken`; `numbered column 1..255, unique and ascending, no unused numbers; none of the 254 rows in
the baseline has been deleted`, rc=0.

Housekeeping, non-blocking: 0 `Claude-Session:` trailers across the six commits (good), and three of
the six carry no `Co-Authored-By:` — a squash merge concatenates the bodies, so the merged commit
will still carry it from the other three.

---

Verdict: **APPROVED** on `845d396ba`. Fix the reachability sentence in the source, the test file and
#3314 before merging; nothing else outstanding from any of my three reviews. The seed, the two scoped
bounds and #3314 are the right decomposition, and the two "the arm cannot see it" disclosures are the
part of this PR I would most want another author to copy.

Issue #2275, opened 2026-08-07
gpu: the storage-image classifier lists exactly the three image ops the recompiler emits — every other image atomic is classified as a sampled texture


mattias800:
`gpu_executor.cpp:3047` decides whether a guest image resource is a **storage image** or a **sampled
texture** from the opcode that uses it:

```c
u.is_storage_image = in.opcode == 0x08 || in.opcode == 0x0f || in.opcode == 0x11;
```

That is `image_store`, `image_atomic_swap`, `image_atomic_add` — **exactly the three image ops the
recompiler happens to emit.** The classifier and the recompiler were kept in lockstep, and neither
generalised.

## Why it is wrong independently of any recompiler gap

**An image atomic is a read-modify-write.** A resource one targets can never be a sampled texture, for
any of the family:

```
0x0f swap  0x11 add   0x12 sub   0x14 smin  0x15 umin
0x16 smax  0x17 umax  0x18 and   0x19 or    0x1a xor   (+ 0x10 cmpswap, 0x1b/0x1c inc/dec)
```

Opcodes verified by assembling each with `llvm-mc -mcpu=gfx1010 -show-encoding` and reading the
opcode field back. The positive control is `image_atomic_add`: it assembles to word0 `0xf0440128`,
**byte-identical to the dword prosper decodes from the guest**, which is what makes the table verified
rather than plausible.

So today a shader doing `image_atomic_umax` on an image gets that image bound as a **sampled texture**,
and the recompiler's storage path — the only one that can emit an image atomic — is unreachable for it
by construction.

## Observed

*Sonic Racing: CrossWorlds* (`PPSA08804`), default launch, two atomics in one compute program:

```
[mimg-class] pc=751 op=0x11 res_cls=4 binding=37    <- StorageImage (0x11 is in the list)
[mimg-class] pc=791 op=0x17 res_cls=2 binding=38    <- Texture     (0x17 is not)
```

`ResourceClass` 2 = `Texture`, 4 = `StorageImage` (`shader_resources.hpp:152`).

## What I tried, and why this is filed rather than fixed

Extending the test to the whole atomic family **did not** unblock `0x17` on its own — the op is still
rejected — so `:3047` is *a* defect but not the only thing standing in the way, and there is at least
one more classification path involved (`gpu_executor.cpp:4374` and `:4851` both consume
`u.is_storage_image`, and the resource's final class may be decided elsewhere as well).

I reverted rather than shipping a behaviour change that does not do what its commit message would
[...]

Issue #1634, opened 2026-08-01
The Oregon Trail (PPSA19244): pixel shader 0x300e460000 fails to recompile — four MIMG instructions cannot resolve their image descriptor (srt_tag=NONE, key_res=null)


mattias800:
Found while investigating #1606 (The Oregon Trail, `PPSA19244`, UE4). Filing separately because the
recompiler gap is generic and the exact opcode/PC data is expensive to recover later.

Platform: Linux, AMD Radeon 8060S (RADV STRIX_HALO), `screenshot` frontend, master `250b55c8`.

## Summary

`ps=0x300e460000` — the dominant pixel shader targeting the title's 3840x2160 scene surface — **fails to
recompile**: four MIMG instructions cannot resolve their image descriptor, so the fragment stage produces
zero words and every draw using it is dropped. Per CLAUDE.md an unsupported/unresolved GPU operation the
guest actually exercises is a FATAL gap, not an acceptable skip.

## Why it is invisible on master

These draws are rejected *earlier*, by the no-effect gate at `gpu_execute.hpp:968`
(`CB_COLOR_CONTROL.MODE=0`), so the recompiler never runs on them. That is why #1606's triage reported
"zero `[recompile-reject]`, zero `[mimg/mubuf-unresolved]`". The gap only becomes visible with
`PROSPER_FORCE_COLORWRITE=1`, which bypasses that gate:

```
PROSPER_GUEST_FS=1 PROSPER_GUEST_ARGS=-force-gfx-direct PROSPER_RENDER=1 \
PROSPER_FORCE_COLORWRITE=1 PROSPER_DBG=1 PROSPER_NO_FRAME_DUMPS=1 \
  ./build-linux/screenshot <DUMP_ROOT>/PPSA19244-app0 --seconds 6 --count 2 --out <OUT> --timeout 200
```

`PROSPER_FORCE_COLORWRITE` is used here purely as a probe to reach the recompiler. It is not a fix and
its output is not acceptance evidence.

## Exact failures

```
[mimg-unresolved] pc=68 srsrc=s8  srt_tag=NONE 0x0 key_res=null pc_res=null (5 res)
[mimg-unresolved] pc=62 srsrc=s16 srt_tag=NONE 0x0 key_res=null pc_res=null (5 res)
[mimg-unresolved] pc=58 srsrc=s0  srt_tag=NONE 0x0 key_res=null pc_res=null (6 res)
[...]

PR #3313, opened 2026-09-04
gpu: resolve an image op's descriptor by class, not by first match at the key (#3126)

mattias800:
## Problem

An image op (MIMG) resolves its T#/U# through three provenance lookups — `by_fetch_pc`,
`by_srt_offset`, `by_sgpr_base`. All three are **first-match-wins and class-blind**, and the resolver
took the first hit at a key and then nulled it out on class *afterwards*. That is strictly weaker than
searching for a resource of an acceptable class, in two ways:

* an image resource sitting **behind** a buffer-class resource at the same key is unreachable. Key
  collisions are ordinary, not hypothetical — Stray's failing title-screen vertex stage carries a
  `ConstantBuffer` and two `VertexBuffer`s all at `sgpr_base 8`;
* a wrong-class hit on the **SRT** route left `res` non-null, which suppressed the `!res` guard on the
  SGPR fallback beneath it — so a route that could have resolved was never tried.

Both end as `[mimg-unresolved] … key_res=null pc_res=null`, the stage recompiles to zero words, and
every draw using it is discarded. Per CLAUDE.md that is a FATAL gap, not an acceptable skip.

The two sibling paths already do this correctly: `s_buffer_load` uses
`by_sgpr_base_cls(…, ConstantBuffer)` and MTBUF uses `by_sgpr_base_cls(…, VertexBuffer)`. MIMG was one
of the two that did not.

## The diagnostic could not report the failure it exists to report

`[mimg-unresolved]` printed four provenance fields — `srt_tag`, `key_res`, `pc_res`,
`ud_alias`/`alias_res` — and **not** the result of `by_sgpr_base(SRSRC)`.

When the same line says `written=0`, the shader never wrote the SRSRC range, i.e. the descriptor is
entry user data by the shader's own construction. There is then no `s_load`, so no tag, no SRT key and
no copy alias: **those four fields are null by construction and the quadruple is a restatement of
`written=0`.** The only route that can fire is `by_sgpr_base`, and that was the one field missing.

Measured on a Stray (`PPSA02101`) title-screen run, `origin/main` `fc21d46ca`, 2026-09-03: five sites,
four with `written=0`, and in **three of those four** a `ConstantBuffer` was sitting at exactly the
requested SGPR the whole time — discarded on class, silently. The fifth site reports `written=1` and
its `srt_tag=0x20` field populates, which is the internal control showing the fields do work when
their route runs.

Separately, the `available:` list printed one key per resource — `s%u` if `sgpr_base` was set, else
`srt=0x%x`. A resource keyed only by `fetch_pc` (how the const-fold publishes a seed- or
table-recovered T#) therefore rendered as `srt=0xffffffff` and read as "unkeyed". Four of the nine
resources in the failing pixel stage printed that way while being properly keyed.

## Change

* `ImageResourceRequirement` (`SampledOnly` / `StorageOnly` / `Either`) plus one shared
  `image_resource_class_satisfies` predicate, so the lookup and the resolver's post-check state the
  operation-class rule **once**. That drift is what made #2275 worse than no fix.
* `ShaderResourceTable::image_by_{fetch_pc,srt_offset,sgpr_base}` — same lookups, filtered inside, so
  the search continues past a wrong-class entry instead of stopping on it.
* The MIMG resolver computes the requirement first and uses the filtered lookups. **The accepted set is
  unchanged** (`image_resource_class_satisfies` is exactly the old post-check), so this can only turn a
  rejection into a resolution; it can never bind a class the operation cannot use.
* `[mimg-unresolved]` gains `sgpr_res=<class>` and `need=<sampled|storage|either>`, and the
  `available:` list prints every key a resource carries (`s8`, `srt=0x50`, `pc=69`) or the literal
  `unkeyed`.

## Deliberately NOT in this PR

**The MUBUF path (`rdna2_emit_alu.cpp:6021`, `:6101`) has the same class-blind first-match shape and is
deliberately untouched.** Its acceptable-class set is a pair (`ConstantBuffer` or `VertexBuffer`) rather
than the image triple, the two arms differ in what they do with the result, and Stray's own MUBUF
rejects are already traced to the #305 user-data window rather than to the lookup (#3137). Fixing it
blind, in the same PR, on no evidence, is how a half-applied fix ends up worse than none (#2275).

**This does not by itself resolve any of Stray's five sites**, and the PR does not claim it does. Those
tables hold no image-class resource at the requested key *at all* — a different defect, upstream in the
front half. Both status docs now say so, and name the single run that decides it
(`PROSPER_DYNTRACE_FAIL=1` + `PROSPER_SHARPLOG=1` on the same route). I did not touch
`build_shader_resources`' four-dword V#-shape classification: changing it shifts binding numbers on
every title that has such a slot, and I had no GPU to check that with.

## Verification

Author had no GPU access for this work (held for a performance lane), so: source analysis, an existing
run log, unit tests. No live run, no snapshots, no capture.

Exact head `49009468b` (rebased onto `origin/main` `5842615a4`; everything after `1f93aee16` is docs-only — 0 non-`.md` files changed — so the build/ctest/mutation results below stand), Linux, Fedora, distrobox `ps5ys`, `-j4`.

```
cmake --build prosper/build-linux -j4
    -> BUILD_RC=0, 0 "error:" lines, [100%]

ctest --test-dir prosper/build-linux --no-tests=error --output-on-failure -j4
    -> CTEST_RC=0, "100% tests passed out of 356", 0 skipped   [re-run at the rebased head]
    -> #213 image_descriptor_class_lookup .... Passed

./prosper/build-linux/test_image_descriptor_class_lookup
    -> exit 0, 21 [ok], 0 [FAIL]
```

**Mutation arm, by exit code.** `image_by_{fetch_pc,srt_offset,sgpr_base}` reverted to
first-match-then-class-check — the exact old behaviour, one line each, everything else identical —
rebuilt (`MUT_BUILD_RC=0`) and re-run:

```
MUT_TEST_RC=1        (21 arms: 15 [ok], 6 [FAIL])
  [FAIL] image_by_sgpr_base searches past the shadowing buffer to the Texture
  [FAIL] image_sample resolves its T# even though a ConstantBuffer shares sgpr_base 8
  [FAIL] IMAGE_LOAD's Either requirement reaches a StorageImage behind a buffer
  [FAIL] image_load resolves a StorageImage that shares sgpr_base 8 with a ConstantBuffer
  [FAIL] image_by_srt_offset searches past it to the StorageImage
  [FAIL] image_by_fetch_pc searches past it to the Texture
FAILED
```

Exactly the six "searches past" arms redden, on all three provenance routes. The **negative** arms —
a lone `ConstantBuffer` at the requested SGPR must stay unresolved, a `Texture` must not satisfy a
storage-only op, a `StorageImage` must not satisfy a sampled one — keep passing under the mutation,
which is what stops "delete the class check" from masquerading as the fix. Reverted and re-run:
`RESTORE_BUILD_RC=0`, exit 0, 21/21, working tree clean.

## Risk

Low. One new enum + three new table methods, all additive; one resolver rewritten to use them with a
provably identical acceptance set; one diagnostic made more informative. No behaviour change for any
stage whose first hit at a key was already acceptable.

## Follow-up measurement (coordinator ran it; folded into the docs here)

`PROSPER_DYNTRACE_FAIL=1` on the same route puts all five sites at `have_t8=0 seed_t8=0`, against
**29** `have_t8=1` and **21** `seed_t8=1` elsewhere in the same run — so the discriminator is live and
no lookup route could have resolved them. That confirms this PR's own claim that it fixes none of the
five.

**Two attributions of that result were proposed and withdrawn the same night, and the docs record
both** rather than shipping either: "#305/#3137 window family" (the failing `srsrc=s8` is user-data
dword **0** of a declared `[0,20)` window and `[udcand]` says `FITS-BLOCK`, so #305's
larger-block-than-window condition is absent), and "a texture descriptor consumed as a buffer" (the
eight dwords are present, readable, and two well-formed V#s — `base=0x302f65b3cc stride=16
num_records=16`, then `base=0x2040f1a5c0 stride=16 num_records=3` — whose T# reading is `type=0`,
`bad-image-type`). `seed_t8=0` is a **shape** verdict, not a residency one. Three readings survive and
one `PROSPER_SHARPLOG` field separates the first two; `STRAY_STATUS.md` names them.

The classification falsification is now **observed rather than derived**: `PROSPER_SHARPLOG` reports
**97** drops of that slot as `claimed by the V# path` and **4** as `degenerate T# (bad-image-type)` in
one run, so on the draws where the V# path does not claim it the texture decoder already reaches those
bytes and rejects them exactly as predicted. The discriminator is `dw1`'s upper half (the V# `STRIDE`
field): `0x0010` claims, `0x0000` falls through. Across 101 sampled drops the declared image slot never
once held an image descriptor.

A third attribution is falsified too, and it is the one that matters most for whoever picks this up:
**"usage-driven or declared-width classification will fix this site"** — the direction #3126 has circled
since 2026-08-30. `PROSPER_SHARPLOG` confirms the guest declares slot 0 as an eight-dword T#
(`ro[0]: offset_dw=0 size=0`), so the premise is right and the conclusion is not: suppress the V# claim
and the texture loop decodes `type = (0x0004dfac >> 28) = 0`, which `valid_image_type` (8..15,
`agc_shader_layout.hpp:183`) rejects as `bad-image-type`. Same rejection, different `continue`. An A/B
of that change **on this title** would report "unchanged" for a reason unrelated to the change. That is
why this PR still does not touch classification — now on evidence rather than on caution.

The seeding base is confirmed correct on the same run (declared layout 0..19 against
`user_data_range_end = 20`; `direct[8]`/`direct[10]` at dwords 16/18 landing on mapped pointers, the
first dereferencing to a well-formed V#), so the disagreement is confined to slot 0.

This PR also adds **instrument trap 254** to `GAME_COMPAT_ORCHESTRATION.md` for the diagnostic defect
itself, since the shape generalises past this one line — and the trap's own row records that
`seed_t8=0` was then misread twice in the same way.

## What the follow-up runs established about the failure itself

Reading `PROSPER_UDPROV` output settled the shape, and it is per-**slot**, not per-shader. The pixel
stage `0x3013c60000` declares four read-only image slots and samples three; in one run exactly the two
mangled slots are dropped and exactly the two ops sampling them fail — 1:1, with the op sampling the
intact slot never failing. The intact slot decodes as an ordinary `base=0x307c620000 1920x1080 type=9
fmt=36 tile_mode=27` texture. Provenance shows the block's low dwords rewritten by a different queue in
the *following* frame, and for the pixel stage that overwrite lands **four dwords into an eight-dword
descriptor** — a different producer with a four-dword slot at that offset, not a rebind.

That is a #305 data point (third-title reproduction of its frontier signature, plus a counter-example
to its window-fit prediction) and it is reported there. It changes nothing in this PR's diff; it is
recorded because the docs in this PR are where the next agent will look.

## Scope, stated plainly

`PROSPER_MIMG_SOFT` on a **verified title-screen frame** already showed that softening these image ops
leaves `max_nonblack` at 0.0069 — unchanged, still menu-only — re-established after #3141 fixed that
instrument. So this PR is a **correctness** fix and a diagnostic fix; it is not the Stray background's
blocker, and nothing here claims a rung change. That result had never been written into any document
(it lived in a source comment and an issue thread, and three sessions re-entered the descriptor
question without it); this PR adds it to `STRAY_STATUS.md` § Ruled out.

## Merge note

`prosper/docs/STRAY_STATUS.md` and `prosper/CMakeLists.txt` are also touched by #3309 (a different
lane, performance). The STRAY_STATUS hunks are ~20 lines apart — theirs ends at the performance
section, mine begins in the image-op section below it — so they should merge, but re-check at merge
time rather than assuming.

Refs #3126, #1634, #3137, #305.


mattias800 (commented):
# Independent review — PR #3313, exact head `49009468b`

**APPROVED.** No blocking findings. Eight non-blocking findings below, three of which I'd like
addressed before merge because they land in durable `## Ruled out` / trap rows that nobody re-derives.

Reviewer scope: source analysis only. No GPU, no build, no snapshots, no re-run of the author's ctest
or CI — I did not duplicate author verification. I did compile and run one throwaway 60-line host
program (details below) to settle the one claim the PR asks to have challenged.

---

## 1. The equivalence claim — verified exhaustively, not spot-checked

> "`image_resource_class_satisfies` is exactly the old post-check ... this can only turn a rejection
> into a resolution, never bind a class the operation cannot use."

**This is the claim that carries the whole change, so I derived the old accepted set from
`origin/main` independently and then compared the two over their entire domain rather than by
reading.**

Old (`origin/main:prosper/src/gpu/recompiler/rdna2_emit_alu.cpp:7061-7069`, with the `by_fetch_pc` entry gate at `:7033-7034`) is a two-stage gate: the
post-check nulls `res` on class, then `if (!res || (res->cls != Texture && res->cls != StorageImage))`
sends it to the unresolved path. Reading them together, a candidate of class `C` resolves for opcode
`op` iff:

* `storage_only_op(op)` -> `C == StorageImage` (the post-check demands it; the final gate then admits it);
* `op == 0x00 || op == 0x0e` -> neither post-check clause applies, so the final gate decides: `C in {Texture, StorageImage}`;
* otherwise -> `C == Texture`.

New (`shader_resources.hpp:793`, `:801`; `rdna2_emit_alu.cpp:7045-7052`) is `StorageOnly` /
`Either` / `SampledOnly` over the identical `storage_only_op` predicate and the identical
`0x00`/`0x0e` exemption.

I transcribed both verbatim into a standalone program and enumerated **every (opcode, class) pair —
256 x 5 = 1280**, of which the old code accepts 258:

```
pairs=1280 old-accepted=258 disagreements=0
positive control (0x0e mis-mapped): disagreements=1 (must be > 0)
```

The positive control is the part that makes the zero readable, per the charter's trap-122 rule: I
mutated the opcode->requirement map (`0x0e` -> `SampledOnly` instead of `Either`) and the comparison
immediately reported a disagreement, so the harness can express the class of difference it found none
of. It is also built outside the thing it validates — neither predicate comes from the other, and the
old one is a verbatim transcription of `origin/main`, not a restatement of the new one.

**Verdict on the equivalence: confirmed, with high confidence.** Sentinel handling matches too — the
three plain lookups guard `0xFFFFFFFF` (`origin/main:shader_resources.cpp:258`, `:264`, `:276`)
and the three new ones guard it identically (`shader_resources.cpp:283`, `:292`, `:301` at head), so the new methods cannot
return null where the old ones returned a hit for a sentinel key.

`ResourceClass` has exactly five enumerators (`shader_resources.hpp:206-215`), so the enumeration
above is the complete domain and `cls_name`'s switch at `rdna2_emit_alu.cpp:7127` is exhaustive.

### 1a. One residual the absolute phrasing does not cover — NON-BLOCKING, but please qualify the wording

The *accepted set* is unchanged; the *route precedence* is not, and there is one case where the new
code binds a **different** (still class-valid) resource rather than turning a rejection into a
resolution.

Old entered the fallback chain when `by_fetch_pc` returned null **or a non-image class**. New enters
it when no *requirement-satisfying* entry exists at that pc. So:

* Table holds, at one `fetch_pc`, a `ConstantBuffer` first and a `Texture` second; a fallback route
  (SRT or SGPR) also resolves to some other `Texture`.
* Old: `by_fetch_pc` returns the cbuf -> enters the fallback block -> binds the **fallback** Texture.
* New: `image_by_fetch_pc` skips the cbuf, finds the **pc-keyed** Texture, and never runs the fallbacks.

Both bind a Texture, so nothing invalid is bound and the "never binds a class the op cannot use" half
is untouched. But "**can only** turn a rejection into a resolution" is one notch stronger than the
code, and that sentence is now in `RESOURCE_BINDING.md` § Ruled out, in the trap-254 row, in the
source comment at `rdna2_emit_alu.cpp:7064-7065` and in the commit message — four places that will be
quoted rather than re-derived.

For what it is worth I think the new behaviour here is the **correct** one: it is exactly what the
line the PR deleted asserted — *"Exact per-use provenance wins over table keys"* — and reachability is
close to nil, since a buffer and an image sharing one `fetch_pc` requires the BVH piggyback at
`gpu_executor.cpp:6092-6100` (which attaches a use pc to an existing `ConstantBuffer`) to land on a pc
that also publishes a texture, and one pc is one instruction of one kind. So I am not asking for a
code change — only for the four copies of the sentence to gain a clause, e.g. *"the accepted set is
unchanged, so it can never bind a class the operation cannot use; where it changes an existing
resolution at all, it is by letting the per-use `fetch_pc` route win over a table key, which is this
file's own stated precedence."*

---

## 2. What I cleared

* **`storage_only_op` is unchanged** and merely hoisted above the lookups; the `0x08/0x09/0x0f/0x11-0x1a except 0x13` set and its `llvm-mc` provenance comment survive intact (`rdna2_emit_alu.cpp:7045-7047`).
* **Downstream is unaffected.** The consumer still branches on `res->cls == StorageImage` vs the sampled path, and both old and new guarantee `res in {Texture, StorageImage}` at that point, so no new class of input reaches `uint_texture` / `d16_float_resource` / the storage path.
* **Blast radius is exactly the MIMG resolver.** `image_by_*` and `ImageResourceRequirement` have no other callers in `src/` (checked at head); the three plain lookups, `by_sgpr_base_cls` and `by_binding` are untouched, so `s_buffer_load` (`rdna2_emit_alu.cpp:5342`), MTBUF (`:6010`) and the pipeline's binding lookup keep their exact behaviour.
* **The deliberate MUBUF omission is correctly scoped.** `rdna2_emit_alu.cpp:6021` and `:6101` are genuinely the same class-blind shape, and the stated reason for not touching them (a two-class acceptable set, two arms that differ in what they do with the result, and Stray's MUBUF rejects already attributed to #305/#3137) is sound. Fixing them here would have been the #2275 pattern the PR is arguing against.
* **The `available:` string building is memory-safe.** Worst case at `rdna2_emit_alu.cpp:7171-7180` is 12 + 15 + 14 = 41 bytes into `char keys[80]` and ~67 into `char one[128]`, so no truncation and `sizeof keys - (size_t)used` cannot underflow. `keys` is always initialised before use because `used == 0` implies the `" unkeyed"` write at `:7179`.
* **Test encodings are real, not synthesised.** `kImageLoadDmask1 = 0xF0000108` is byte-identical to the word #1634 records at `pc=34` (`words=f0000108,00020600 ... op=0x0 dmask=0x1`); `kImageSampleDmask15 = 0xF0800F08` matches `test_direct_descriptor_copy.cpp:40`; and `mimg_word1(0,0,8,16)` evaluates to `0x00820000`, byte-identical to the word1 at `test_shader_recompile_cache.cpp:358`. The `mimg_word1` field layout is self-consistent with both.
* **Trap number.** Main's highest is 253; this PR takes 254 and the table stays strictly ascending at head. No collision with the 255 that #3312 takes concurrently. `Docs` is green, 9 checks pass / 2 skipped, `MERGEABLE CLEAN`.
* **`STRAY_STATUS.md` is +320/-0** — zero deletion lines, so no trap-41 clobbering of the concurrent #3309/#3312 edits to the same file.
* **The claim that everything after `1f93aee16` is docs-only holds**: `git diff --name-only 1f93aee16 49009468b` yields 0 non-`.md` paths, so the build/ctest/mutation results do carry to this head.

---

## 3. Finding — the reasoning for NOT changing classification (your item 2) checks out, on code and on arithmetic

I traced this one end to end because it is what justifies leaving a real fix unattempted.

Both drop reasons come from **the same loop over the same slot**:
`agc_shader_layout.cpp:968` drops with `"claimed by the V# path"` when `readonly_buffer_slots[slot]`
is set; when it is not, control falls through to `load_sharp` -> `decode_image_descriptor` ->
`image_descriptor_reject_reason` at `:980`, which returns `"bad-image-type"` from `:454` via
`valid_image_type` — and `agc_shader_layout.hpp:183` is indeed `type >= 8 && type <= 15`.
`decode_image_descriptor` takes `d.type = (t[3] >> 28) & 0xF` (`agc_shader_layout.cpp:360`), and
`0x0004dfac >> 28 == 0`. So suppressing the V# claim moves the slot from one `continue` to another,
exactly as the PR says. **The reasoning is correct and the conclusion — don't touch classification —
is right.**

The documented bytes are arithmetically self-consistent, which is a stronger check than taking the
decode on trust: `2f65b3cc 00100030 00000010 0004dfac` gives V# base `0x302f65b3cc`, `STRIDE =
(dw1 >> 16) & 0x3fff = 16`, `NUM_RECORDS = dw2 = 16` — precisely the values `STRAY_STATUS.md` reports —
while the same dw3 read as a T# gives `type = 0`. The `dw1`-upper-half discriminator claim
(`0x0010` claims, `0x0000` falls through) is consistent with the two rejected rows in the same table.

*Precision nit (non-blocking).* `tex_drop` dedupes on `(user-data block, slot, reason)`
(`agc_shader_layout.cpp:944-962`), so "**97** drops of that slot as `claimed by the V# path` and **4**
as `degenerate T#`" counts distinct **(block, slot)** pairs per reason, not draws. Nothing in the
inference depends on it, but "101 sampled drops" reads as a per-draw census and is not one. One
clause in `STRAY_STATUS.md` would fix it.

---

## 4. Finding — the new diagnostic can mislead in a fresh way when `written=1` (NON-BLOCKING, recommended)

`rdna2_emit_alu.cpp:7126` computes `sgpr_res` **unconditionally**:

```cpp
const ShaderResource* ps_direct = rt->by_sgpr_base(in.src[1].value);
```

but the resolver only consults that route when the SRSRC range is *unwritten*
(`:7071`: `if (!res && !sreg_range_written(rs, in.src[1].value, 8))`). So on a `written=1` site the
line prints a real observation about a route **that did not run**.

Concrete failure: a `written=1` site where a `Texture` happens to sit at that SGPR prints
`need=sampled sgpr_res=tex written=1` and still reports unresolved. A reader — reasonably — concludes
"the Texture was found and rejected", when the resolver never asked; the actual cause is the
`sreg_range_written` gate. This is live on the title in the PR: `STRAY_STATUS.md`'s own table has one
`written=1` site (`0x30131d0000` pc=77).

This is the *inverse* of the defect the PR fixes rather than a repeat of it — a field that overstates
rather than one that is vacuous — but it is the same family, and the PR's own trap 254 says the
recognition cue is "a diagnostic whose fields enumerate mechanisms, plus a field that says which
mechanism applies". The line already prints `written` immediately after `sgpr_res`, so the
information is *derivable*; making it *unmissable* is one expression:

```cpp
const bool srsrc_written = sreg_range_written(rs, in.src[1].value, 8);
const ShaderResource* ps_direct = srsrc_written ? nullptr : rt->by_sgpr_base(in.src[1].value);
// ... and print "skipped" rather than "null" for the written=1 case
```

Not blocking — it is diagnostic-only and strictly more informative than what it replaces.

---

## 5. Finding — test coverage gap on the half where a mistake would bind a wrong class (NON-BLOCKING, recommended)

The 21 arms and the 6-red mutation are good work, and I confirmed the arm set **brackets** the change
rather than merely differing from it: reverting the lookups reddens exactly the six "searches past"
arms (arms 3, 4, 11, 12, 14, 17 by my count of the file, matching the author's list verbatim), while
*widening* acceptance instead — deleting the class check — would redden a different four (the lone-cbuf,
Texture-vs-storage, StorageImage-vs-sampled and srt-refusal negatives). Two opposite mutations, two
disjoint red sets: that is a bracket. Arm 1 is a proper positive control, differing from arm 6 in
exactly the resource class.

**What is not covered is the opcode -> requirement mapping at `rdna2_emit_alu.cpp:7048-7052`.**
`recompiles()` is only ever called with op `0x20` (sampled) and op `0x00` (either). No arm drives a
storage-only opcode (`0x08`, or an atomic such as `0x17`) or `0x0e` through the resolver. The
predicate arms 19-21 test `image_resource_class_satisfies` *given* a requirement; nothing tests that
the right requirement is chosen for those opcodes.

That is precisely the half where an error binds a **wrong resource** rather than failing to bind one —
if a storage-only opcode were ever mapped to `Either`, a `Texture` would satisfy an `image_store`
and every existing arm would stay green. I verified the mapping myself over all 256 opcodes (§1), so
this is a durability gap rather than a live defect, but a single arm — a `0x08` word with a `Texture`
at s8 asserting `!recompiles(...)`, and the same with a `StorageImage` asserting `recompiles(...)` —
would pin the one thing the suite currently takes on faith. It is the highest-value arm to add.

Second, smaller gap: failure mode **(b)** — a wrong-class SRT hit suppressing the SGPR fallback — is
tested only at the lookup level (arm 14). No arm drives it through the resolver, so the *route
interaction* the PR body leads with is asserted but not pinned.

---

## 6. Finding — a comment was orphaned from the struct it documents (NON-BLOCKING, cosmetic)

`shader_resources.hpp:784-785`:

```
// The set of resources a shader uses. The front-half builds it from the shader's user_data; the
// recompiler consults it while translating memory ops and the pipeline binds from it. Pure data.
// Which resource class an image instruction can accept. RDNA2 splits MIMG by operation, not by
```

The new declarations were inserted *between* `ShaderResourceTable`'s doc comment and the struct, with
no blank line, so those two lines now read as the opening of `ImageResourceRequirement`'s comment and
sit 26 lines above the struct they describe. Given how much this project leans on comments as the
map, worth the one-line move (put the enum + predicate *above* the `// The set of resources...`
comment, or insert a blank line and re-attach).

---

## 7. `## Ruled out` and trap rows — checked against the cited evidence

All four `STRAY_STATUS.md` rows and the three `RESOURCE_BINDING.md` rows are well-scoped, name their
evidence, and are careful to distinguish what was measured from what was inferred. The two
withdrawals (`#305` window family; "a texture descriptor consumed as a buffer") are correctly
recorded as *withdrawn* rather than quietly dropped, and the `[0,20)` / `FITS-BLOCK` reasoning for
excluding #305's larger-block-than-window condition matches #305's own titled signature. The
"seed_t8=0 is a **shape** verdict, not a residency one" distinction is the right one and is the most
valuable sentence in the set.

Two small accuracy notes, both non-blocking:

* **The `RESOURCE_BINDING.md` counter-example is aimed at a generalisation, and should say so.** The
  paragraph it contradicts (`origin/main:RESOURCE_BINDING.md:190-192`) states presence/absence for
  *vertex* windows ("8/12/28-dword vertex windows and not in 30/32-dword vertex windows") and then
  makes the *pixel* clause an observation, not a prediction. The new counter-example is a **pixel**
  stage at `[0,30)`. It does contradict "Window fit, not stage identity, predicts the result", which
  is the generalising sentence — but a reader could take it as falsifying the vertex-window clause,
  which it does not touch. One clause naming which half fell would prevent that.
* **Trap 254's "#1634 was filed on the null quadruple" is off by one field.** #1634's pasted lines are
  `srt_tag=NONE 0x0 key_res=null pc_res=null` — a **triple**; `alias_res` and, more importantly,
  `written` did not exist on the line then. The substance is unaffected (the printed fields carried no
  information either way), but the trap's central argument is *"`written=0` makes the others vacuous"*,
  and at #1634's filing `written` was not printed at all — so #1634 is an instance of the weaker
  "the line omitted the route that ran", not of the `written=0` reading. Worth one clause, because
  trap rows are quoted rather than re-derived.

---

## 8. Two operational notes for whoever merges

* **Merge-order hazard with #3312.** Both PRs append to `GAME_COMPAT_ORCHESTRATION.md`'s Instrument
  table; this one takes 254, #3312 takes 255. `check_numbered_table.py --ordered` requires **strictly
  ascending** rows (`:92`, `:815`), so if #3312 lands first the merge must still leave 254 *before*
  255 in the file. Run `check_merge_result.py` immediately before merging, per CLAUDE.md — the green
  `Docs` job above is a statement about a merge that may no longer exist.
* **No snapshots were run, and this change can newly-resolve stages on other titles.** That is
  permitted (the charter leaves snapshots to the author outside release time) and the risk is small,
  but it is not zero and the PR's "no behaviour change" framing is scoped to stages whose first hit
  was already acceptable. A stage that previously recompiled to zero words and had its draws
  discarded can now compile and draw — on any title, including the seven rung-6 guarded ones. Worth a
  line in the release-matrix notes so a later regression is not re-bisected from scratch.

---

**Verdict: APPROVED.** The central equivalence claim is correct and I verified it exhaustively rather
than by reading. The change is a strict improvement over a lookup shape that two sibling paths already
got right, the diagnostic fix is the more valuable half, and the docs are unusually honest about what
this does *not* fix. Please consider the §1a wording qualification, the §4 `written=1` guard and the
§5 storage-opcode arm; none of them block.

Issue #2049, opened 2026-08-06
CLAUDE.md says PROSPER_GUEST_FS is off by default — it is opt-OUT on Linux/Windows, and the stale line has already propagated into shipped code comments


mattias800:
Found while fixing a blocking review finding on #2030.

## The stale line

`CLAUDE.md:273-274`:

> **Reaching the running frame loop** needs two gated switches (off by default, so the default boot stays
> stable): `PROSPER_GUEST_FS=1 PROSPER_GUEST_ARGS=-force-gfx-direct`.

`PROSPER_GUEST_FS` is **not** off by default, and it is not even the variable that controls this on the
platform everyone develops on.

## What the source actually says

`prosper/src/host/guest_tls.cpp`, in `guest_tls_set_templates()`:

```
:45   #ifdef __APPLE__
:46       g_enabled = getenv("PROSPER_GUEST_FS")    != nullptr;   // macOS/Rosetta trap mode, opt-IN
:54   #else
:58       g_enabled = getenv("PROSPER_NO_GUEST_FS") == nullptr;   // Linux/Windows, opt-OUT
:59   #endif
```

and the file header at `:1-2` states it plainly:

> Enabled by default on Linux/Windows (opt out with `PROSPER_NO_GUEST_FS`); macOS/Rosetta trap
> emulation remains opt-in with `PROSPER_GUEST_FS`.

`:209` and `:240` repeat it, and `prosper/src/hle/dispatch.hpp:218` agrees. `grep -rn 'getenv("PROSPER_GUEST_FS")' prosper/src`
returns exactly **two** hits: the Apple-only line 46, and (until #2030) one comment that had copied the
wrong polarity. **No Linux code path reads `PROSPER_GUEST_FS` at all.**

## Why this is worth an issue rather than a quiet edit

It has already caused a real defect in the permanent record. A comment added by #2030 justified a new
diagnostic state with *"That is the DEFAULT — `getenv("PROSPER_GUEST_FS") != nullptr` (guest_tls.cpp:46),
opt-in"*, which is inverted on the only platform that can reach that state. It survived a review round,
a briefing, and an author-verification comment — three documents — because everyone was reading
`CLAUDE.md`, not `guest_tls.cpp`. Independent review caught it; corrected in `9a95221e`.

`CLAUDE.md` is the first thing every agent reads here, so a stale line in it does not stay contained.

## Suggested fix

Reword `CLAUDE.md:273-274` so it stops asserting a default it does not control. Something like:

[...]

Issue #1773, opened 2026-08-02
recompiler: a direct user-data descriptor copied with s_mov_b32 loses its provenance, so the image op is rejected


mattias800:
## Summary

The recompiler cannot resolve an image/buffer descriptor whose SRSRC range was produced by copying
user-data SGPRs with `s_mov_b32`. `rs.sreg_srt` (the **indirect**, `s_load`-derived provenance tag)
is propagated through `s_mov_b32` at `prosper/src/gpu/rdna2_to_spirv.cpp:5805`; there is no
equivalent for **direct** user-data provenance, so a copied descriptor loses its only key.

## Failure scenario

A stage whose descriptor lives directly in user data at `offset_dw = N` is keyed by the front half as
`sgpr_base = user_sgpr_base + N` (`agc_shader_layout.cpp`, "DIRECT provenance"). The recompiler's MIMG
resolution (`rdna2_to_spirv.cpp:9340-9349`) tries, in order:

```cpp
const ShaderResource* res = rt->by_fetch_pc(in.pc);
if (!res || wrong class) {
    if (sreg_srt_range_tag(rs, in.src[1].value, 8, srt_tag)) res = rt->by_srt_offset(srt_tag);
    if (!res && !sreg_range_written(rs, in.src[1].value, 8))  res = rt->by_sgpr_base(in.src[1].value);
}
```

When the shader stages the descriptor — `s[20:27] <- s[9:16]`, eight `s_mov_b32` — before the image
op:

* `by_fetch_pc` misses whenever the front half keyed the fetch pc of a *different* use of the table;
* `sreg_srt_range_tag` misses because the descriptor never came through an `s_load`, so there is no tag;
* `by_sgpr_base` is **skipped**, because `sreg_range_written(rs, 20, 8)` is true — the eight moves wrote it.

The descriptor is unresolvable, the MIMG is rejected, and the whole draw is dropped.

## Live evidence (Earthion, `PPSA28061`, eboot base `0x410000000`)

Pixel shader `0x4101c1f00`, disassembled with `shader_inspect` from a `PROSPER_SHADER_DUMP` capture:

```
pc=0002..0012  s_mov_b32 s21..s31, s10..s20     ; s[20:31] <- s[9:20]
pc=0013        s_mov_b32 s20, s9
pc=0014..0021  s_mov_b32 s12..s19, s1..s8       ; s[12:19] <- s[1:8]
[...]

PR #3315, opened 2026-09-04
gpu: say whether the SGPR route was CONSULTED, and give the mapping arms teeth (#3126)

mattias800:
Follow-up to #3313's review. Three substantive items and three wording fixes. **One of them found a
defect in my own new test**, which is the part worth reading.

## 1. `sgpr_res` printed without saying whether the route ran — trap 254 inverted

`[mimg-unresolved]` printed `sgpr_res` unconditionally, but that route is guarded by
`!sreg_range_written`. On a `written=1` site a bare `sgpr_res=tex` tells the reader a Texture was found
at the requested SGPR and rejected — when nothing consulted it, and the descriptor is not the entry
user data at all. That is exactly trap 254's shape with the sign flipped: a field whose value is forced
by a mode the same line prints. It is **live on Stray** at `0x30131d0000` pc=77.

It now prints `not-consulted(<cls>)` when the route did not run — keeping both facts, that it was
skipped and what happens to sit there anyway, because dropping the second half would trade one
misreading for another.

## 2. The mapping arms — and the arm that looked sensitive and was not

The reviewer was right that no arm drove a storage-only opcode or `0x0e` through the resolver. Adding
`image_store` (0x08) arms turned up something worth recording rather than quietly fixing:

**The obvious arm — "image_store refuses a Texture at its SRSRC" — is NOT sensitive to the mapping.**
Collapsing `image_requirement` to `Either` leaves it green, because the emitter dispatches on class a
*second* time at `if (res->cls == ResourceClass::StorageImage)` and a store bound to a Texture falls
out of that branch anyway. I know this because I ran the mutation, not because I reasoned about it —
and a green arm there would have been "the assertion is true and the mechanism never ran", which is the
failure mode this repo has a memory note about.

The two arms that **do** have teeth put both image classes at one `sgpr_base` with the wrong one
*first*, so only the requirement can decide which gets bound:

```
image_store  : [Texture, StorageImage] @ s8  -> StorageOnly skips the Texture -> compiles
image_sample : [StorageImage, Texture] @ s8  -> SampledOnly skips the simg    -> compiles
```

Under the mapping mutation **exactly those two redden** (`exit 1`, 25 other arms green). The
insensitive arm is kept and relabelled as the defence-in-depth check it actually is.

A `static_assert` now pins the three MIMG word0 constants to their OP fields, so the encodings are not
three independent guesses.

## 3. The route-precedence claim was overstated, in code and in docs

The accepted set really is unchanged — the reviewer enumerated all 256×5 (opcode, class) pairs, 258
accepted, zero disagreements, with a mis-mapped `0x0e` control that fires. But **precedence changed**:
a pc-keyed image resource behind a buffer-class one now wins, where the old code fell through to the
SRT/SGPR routes. So the outcome can be a *different but class-valid* resource, not only a rejection
turned into a resolution. That is the intended contract — exact per-use provenance outranks a table key
— and both the code comment and `RESOURCE_BINDING.md` now say it.

## Wording

* the `ShaderResourceTable` doc comment was orphaned by the enum insertion and now reads as that
  struct's again;
* trap 254's "#1634 was filed on the null quadruple" is corrected — that line printed a **triple** and
  had no `written` field at all, which *strengthens* the row's argument (the field that makes the other
  three vacuous did not yet exist);
* `STRAY_STATUS.md`'s 97/4 counts are relabelled as distinct `(block, slot, reason)` dedupe keys rather
  than draws, so "101 sampled drops" no longer reads as a per-draw census it never was. The per-draw
  claim that section actually rests on is the 1:1 slot/op correspondence, which is read off one run's
  drops and rejects.
* the `RESOURCE_BINDING.md` counter-example now names the clause it falsifies — the *generalisation*
  ("window fit, not stage identity, predicts the result"), not the vertex-window survey it quotes.

## Verification

Exact head `c1dba8e96`, Linux, distrobox `ps5ys`, `-j4`.

```
cmake --build prosper/build-linux -j4   -> BUILD_RC=0, 0 "error:" lines, [100%]
ctest --no-tests=error -j4              -> CTEST_RC=0, 100% passed out of 356, 0 skipped
                                           #213 image_descriptor_class_lookup .... Passed
test_image_descriptor_class_lookup      -> exit 0, 27 [ok], 0 [FAIL]
```

**Mutation, by exit code** (`image_requirement` collapsed to `Either`, everything else identical):

```
MUT_TEST_RC=1   (27 arms: 25 [ok], 2 [FAIL])
  [FAIL] image_store binds the StorageImage past a Texture at the same SGPR (StorageOnly mapping)
  [FAIL] image_sample binds the Texture past a StorageImage at the same SGPR (SampledOnly mapping)
```

Only the two mapping-sensitive arms move; the defence-in-depth arm and every class-filter arm stay
green, which is what stops this from being a blunt "any change reddens everything" test.

## Risk

Low. One diagnostic string, four test arms, one comment relocation, three doc corrections. No change to
resolution behaviour: `image_requirement`, the lookups and the accepted set are untouched.

Refs #1634, #305, #3313.


mattias800 (commented):
# Independent review — PR #3315, exact head `c1dba8e96`

**REJECTED**, on **one** blocking finding, which is a three-clause wording fix and nothing more. Everything
this PR set out to do is correct and I verified the load-bearing claim — the mapping arms' sensitivity —
rather than taking the reported mutation on trust. The blocker is a factual error in the trap-254 row this
PR rewrites: it names the wrong four fields as vacuous, and the field it wrongly includes is one whose null
is a real measurement. Re-review on the corrected head should be minutes.

Reviewer scope: source analysis only. No GPU, no build, no ctest, no snapshots — I did not duplicate author
verification. I re-derived the mutation result by tracing the emitter rather than rebuilding, because the
paths are straight-line and deterministic and I had no residual doubt; I say so explicitly below so nobody
reads it as a re-run.

---

## The question the PR asks to have challenged: are the replacement arms sensitive to what they name?

**Yes. Confirmed by tracing, and the "only two move" count is right.**

**The insensitivity finding is correct.** `test_image_descriptor_class_lookup.cpp:213` — `image_store` +
`Texture` at s8, asserting `!recompiles` — cannot see `image_requirement`. Under the mutation
`image_by_sgpr_base(8, Either)` returns the Texture, `rdna2_emit_alu.cpp:7312`
(`if (res->cls == ResourceClass::StorageImage)`) is false, control falls to the sampled path, and
`:7647-7650` rejects because `0x08` is in none of `is_sample/is_load/is_sample_l/...` (`:7613-7623`). Green
either way. The relabel — *"some layer; see the mapping arms below"* — states exactly what it pins.

**Both replacement arms are sensitive, and to the requirement:**

* `:226` `[Texture, StorageImage] @ s8`, op `0x08`. Correct: `image_by_sgpr_base(8, StorageOnly)`
  (`shader_resources.cpp:290-297`) skips the Texture, returns the StorageImage, `:7312` is taken, `is_st`,
  compiles. Mutated to `Either`: first match is the Texture, `:7312` false, `:7647` rejects. **RED.**
* `:234` `[StorageImage, Texture] @ s8`, op `0x20`. Correct: `SampledOnly` skips the simg, binds the
  Texture, sampled path compiles. Mutated: first match is the simg, `:7312` **is** taken, `is_ld`/`is_st`/
  `is_atomic` are all false for `0x20`, so `:7348` sets `ok=false`. **RED.**

**"Only two" is right.** I enumerated all 27 arms against the mutation. Seventeen pass an explicit
`ImageResourceRequirement` literal or use a class-blind lookup, so mutating the emitter's `image_requirement`
cannot reach them. Of the ten `recompiles()` arms: **two** drive `0x00`, whose requirement is already
`Either` (`:181`, `:250`); **five** are unchanged because no wrong-class *image* is first at the key —
`:121`, `:140`, `:152`, `:198`, `:242`, where the only shadowing resource is a `ConstantBuffer`, which
satisfies no requirement, `Either` included (`shader_resources.hpp:799-807`); **one** is the documented
insensitive arm (`:213`); and **two** flip. 17 + 8 green, 2 red — 25/2, matching the reported matrix
exactly.

One note worth putting in the file so it survives a future tidy-up: `:226` alone pins only *"not
Texture-admitting"*. It is `:198` (`image_store` resolves a lone StorageImage) **plus** `:226` that pin
`0x08 -> StorageOnly` specifically. `:198` is labelled "control for the arms below", which is right, but a
reader looking for redundancy could still delete it. One clause would prevent that. Non-blocking.

---

## BLOCKING — the "quadruple" the docs call vacuous is the wrong four; `pc_res` is not one of them

`pc_res` is `by_fetch_pc(in.pc)` (`rdna2_emit_alu.cpp:7122`), and the resolver's **first** route is
`image_by_fetch_pc(in.pc, image_requirement)` at `:7079`. Neither is guarded by `sreg_range_written`. So on a
`written=0` site `pc_res=null` is a real measurement — "no resource in this table is keyed to this op's pc" —
not a null forced by the mode.

The code gets this right. The comment at `:7130-7134` names the actually-vacuous set:

> `srt_tag=NONE key_res=null ud_alias=NONE alias_res=null` is a restatement of `written=0`

…and that set is genuinely forced: `sreg_srt` and `sreg_ud_alias` are only ever populated through a scalar
write, and `record_scalar_write` inserts into `sreg_written` (`rdna2_to_spirv_internal.hpp:5085`) in the same
loop that precedes `rs.sreg_ud_alias[ud_alias_dst] = ...` (`:5094`). No write, no tag and no alias.

Three documents substitute `pc_res` for `ud_alias` and so over-claim by one field:

1. **`GAME_COMPAT_ORCHESTRATION.md` trap 254 — rewritten by this PR, which is why this blocks here.** It
   lists *"the `s_load` tag (`srt_tag`), the SRT key (`key_res`), the per-pc fold (`pc_res`) and the #1773
   copy alias (`alias_res`)"*, then reasons *"there is no `s_load`, hence **no tag, no SRT key and no copy
   alias can exist**"* — **three** mechanisms — and concludes *"all four printed fields are null whatever is
   wrong"*. The gap is visible inside the sentence.
2. **`RESOURCE_BINDING.md:169`** § Ruled out (context in this diff): *"Those four fields describe the
   s_load-tag, SRT-key, per-pc and copy-alias routes … none of those four routes can fire."*
3. **`STRAY_STATUS.md:105-107`**: same four, same conclusion.

**`STRAY_STATUS.md:140` refutes it, in the same document:**

> `untouched_seed_range(tbase, 8)` publishes a `SrtUse` with the live eight dwords and `use_pc`, which
> becomes `fetch_pc` provenance — the resolver's *first* lookup. So `pc_res=null` could have meant "the fold
> declined the bytes" or "it produced them and materialization dropped them"

That paragraph exists **because** `pc_res=null` carried information and needed `PROSPER_DYNTRACE_FAIL=1` to
disambiguate. The doc both uses the field as evidence and tells the reader it is vacuous.

*Concrete failure scenario.* A lane hits a `written=0` `[mimg-unresolved]` on some other title, reads
`pc_res=null`, recalls trap 254 saying the quadruple is a restatement of `written=0`, and does not run the
dyntrace arm that would have separated "the const-fold never produced this T#" from "it produced it and
materialization dropped it" — two different files. That is the trap this row exists to prevent, with the sign
flipped: the row now teaches readers to discard the one route whose null is genuine.

*Fix (three clauses, no code):* in all three places, name the vacuous set as `srt_tag`, `key_res`,
`ud_alias`, `alias_res` — matching `rdna2_emit_alu.cpp:7133` — and say explicitly that `pc_res` is the
exception: it is the resolver's first lookup, runs unconditionally, and its null is a measurement.
`RESOURCE_BINDING.md:169`'s verdict ("Falsified") survives untouched; only *"the emptiness carried no
information at all"* needs the carve-out.

---

## Non-blocking findings

### 1. `STRAY_STATUS.md:199-201` — the dedupe correction is right in direction and still not exact

`tex_drop`'s key is `{ud, slot, why}` (`agc_shader_layout.cpp:954`) and `why` is the **whole** string. For the
`claimed by the V# path` bucket that is a fixed literal, so "distinct blocks per reason" holds. For the
`degenerate T#` bucket it is not: `why` is built at `:984-989` and **embeds the raw eight dwords** plus the
per-dword `residue … is MAPPED` notes. So the key there is `(block, slot, payload)`, and *"a block that is
dropped for one reason on ten thousand draws contributes one"* is false for exactly that bucket. The code
comment at `:949-953` says so in as many words — *"one slot can decode cleanly on one draw and as residue on
the next"* — and the doc's own quoted pair at `ud=0x21e0e55590 ro[0]` (`:192-193`) is that case.

Consequence at `:220`: *"the payload … **varies between blocks**"* is not derivable from these lines. With
the payload inside the key, two lines with different dwords are equally consistent with one block varying
across draws — which is the mechanism the paragraph just finished explaining. `"varies"` alone is supported;
`"between blocks"` needs the `ud=` pointers, and only one of the three table rows carries one. This is not the
old error restated, but it is the same axis-of-variation over-specification pointed the other way.

### 2. `STRAY_STATUS.md:222-223` — the "per-draw claim" is cited to per-run evidence

*"The per-draw claim this section rests on is the 1:1 slot/op correspondence below."* That section is
explicitly per-**run**: `:289` reads *"In one run, exactly two of them are dropped"*, and the table's column
header is *"outcome that run"*. It is a correspondence between two **deduped** instruments (`tex_drop`, and
`[mimg-unresolved]` deduped per `(program, pc, srsrc)` at `rdna2_emit_alu.cpp:7111-7116`) — identity, not
frequency. It cannot show per-draw variation.

The doc does hold genuine per-draw evidence, two paragraphs earlier: the same `(ud, slot)` printing **both**
reasons, which are mutually exclusive within one decode (`:968` vs the fall-through at `:977-1003`), so the
content demonstrably changed between draws — plus the `[udprov]` mid-descriptor split at `:307-309`. Point
the sentence at those and the claim is fully supported.

### 3. `rdna2_emit_alu.cpp:7077-7078` — the reachability clause names a mechanism that cannot reach it

> Reachability needs the BVH piggyback (`gpu_executor.cpp`) to land on a texture-publishing pc.

The piggyback is real — `gpu_executor.cpp:6099`/`:6109` do key a `ConstantBuffer` by a MIMG `use_pc`. But that
pc is an `IMAGE_BVH_INTERSECT_RAY`, and `rdna2_emit_alu.cpp:6807` (`if (in.opcode == 0xe6u)`) returns on every
path before the class-filtered lookup at `:7079` runs. One pc is one instruction, so the buffer this
mechanism keys can never be the buffer that shadows a *sampling* MIMG. I also could not find another producer
that keys a buffer-class resource at a MIMG pc: `attach_image_use` (`shader_resources.hpp:545`) is called only
under `r0.cls == wanted` with `wanted` in {Texture, StorageImage} (`gpu_executor.cpp:7475`, `:8013`), and every
other `fetch_pc = u.use_pc` site keys the pc of a buffer/scalar op, which is a different instruction.

Nothing about the **contract** statement changes — "the outcome can be a different but class-valid resource"
is true of the code regardless of reachability, and I verified it against the pre-#3313 resolver below. Only
the mechanism name is wrong. Suggest *"no reachable producer is known: the one path that keys a buffer by a
MIMG pc is the BVH piggyback, and `0xe6` returns before this lookup"*, or drop the clause.

Worth flagging the origin, because it is the shape `CLAUDE.md` records at #2049 → #2052: this sentence came
verbatim from #3313's review §1a (*"requires the BVH piggyback at `gpu_executor.cpp:6092-6100`"*), where it
was offered as a reason the case is nearly unreachable — a claim the author had every reason to want to be
true — and it has now been promoted into a shipped code comment as a positive mechanism. A reviewer's
`file:line` is exactly the kind of citation that stops getting checked.

### 4. `test_...cpp:49-51` — the `static_assert`s check 7 of the opcode's 8 bits

`rdna2_decode.cpp:888` is `i.opcode = ((w & 1u) << 7) | ((w >> 18) & 0x7Fu)` — the MSB is word0 bit 0, and the
comment there records that dropping it aliased `IMAGE_MSAA_LOAD` onto `IMAGE_LOAD`. The asserts check only
`(w >> 18) & 0x7F`, so a future mistype that sets bit 0 passes the assert while the decoded opcode becomes
`0x80`/`0x88`/`0xa0`. Since the stated purpose is *"if a future edit mistypes one, this fires before any
resolver arm runs"*, mirroring the decoder's expression closes the hole for free.

### 5. Trap 254 — two small residuals

* The `#1634` correction is **verified against #1634 itself**: its pasted lines are
  `[mimg-unresolved] pc=68 srsrc=s8 srt_tag=NONE 0x0 key_res=null pc_res=null (5 res)` — a triple, no
  `alias_res`, no `written`. Correct, and it does strengthen the row.
* *"so that report could not have been read correctly even in principle"* is a notch stronger than the
  evidence. #1634's body reads its own triple correctly — *"neither the SRT tag nor the PC-keyed lookup finds
  a T# … even though the stage's resource table is populated"* — and asks the right next question. What could
  not be determined from the line was whether the nulls were **forced**. Scoping the sentence to the line
  rather than the report would be exact.
* Pre-existing, not introduced here: the row's headline says *"three times over two issues"* while the body
  ends *"re-derived the same reading a fourth time"*. Worth reconciling while the row is open.

---

## What I cleared

* **`not-consulted(<cls>)` cannot mislead in a fresh way.** Inside the diagnostic block `res` is null, so the
  SRT route returned null too; therefore the SGPR route ran **iff** `!written`, and `srsrc_written`
  (`:7147`) is an exact "was this route consulted" predicate — not an approximation of one. `written=` is
  still printed alongside (`:7178`), and the alias route, which *does* run at a different SGPR on a
  `written=1` site, keeps its own `ud_alias=`/`alias_res=` fields. Keeping the contents inside the marker
  rather than replacing them with `skipped` (which is what #3313's review suggested) is the better call and
  the comment at `:7164-7166` says why. The single `sreg_range_written` call now feeds both the field and the
  marker, and `rs` is not mutated between `:7085` and `:7147`, so the two cannot disagree.
* **The route-precedence correction is exactly right, and I checked it against the old code rather than the
  description of it.** Pre-#3313 (`28d4db1cd:rdna2_emit_alu.cpp:7033-7046`) is
  `res = by_fetch_pc(pc); if (!res || (res->cls != Texture && res->cls != StorageImage)) { res = nullptr; …fallbacks… }`.
  So the fallbacks were entered only when the pc hit was null **or buffer-class**; an image-class hit of the
  wrong requirement suppressed them and was then nulled by the post-check. Enumerating the four transitions:
  buffer-first-at-one-pc is the **only** one that yields a different-but-valid resource (old took a fallback
  Texture, new takes the pc-keyed Texture); wrong-requirement-image-at-pc and wrong-class-at-SRT both go
  rejection → resolution. The paragraph in the comment (`:7071-7078`) and in `RESOURCE_BINDING.md` describes
  exactly that one case and does not overstate it.
* **The window-survey clause correction is exact.** The survey at `RESOURCE_BINDING.md:213-215` states
  presence/absence for **vertex** windows and makes the pixel clause an observation; the new text keeps both
  as statements about that run and falsifies only *"Window fit, not stage identity, predicts the result"*,
  which the Stray pixel stage at `[0,30)` does falsify. Nothing is over- or under-stated.
* **The `ShaderResourceTable` comment relocation** (`shader_resources.hpp:809-811`) is a clean re-attach; the
  enum keeps its own comment and the struct gets its own back.
* **Test encodings.** `0xF0200108 - 0xF0000108 == 0x00200000 == 8 << 18` exactly, and the two words share
  every other word0 field (`dim=1`, `nsa=0`, `dmask=1`), so the store arm differs from the already-passing
  load arm in the opcode alone. Resource key sentinels are `0xFFFFFFFF` (`shader_resources.hpp:255-256`,
  `:321`), so no arm accidentally resolves through `by_fetch_pc(0)`.
* **Hygiene.** `check_numbered_table.py --ordered --baseline origin/main` on head: *254 rows, unique and
  ascending, none of the 254 base rows deleted*. `git diff --check` clean. Numstat is 1/8/11 doc additions
  against 1/5/5 deletions, every deletion accounted for by a hunk I read — no trap-41 whole-file clobber of
  the concurrent edits to `STRAY_STATUS.md`. Eleven CI checks pass/skip on this head. No `Claude-Session`
  trailer; `Co-Authored-By` present.

---

**Verdict: REJECTED**, blocking on the `pc_res` finding only. It is a wording fix in
`GAME_COMPAT_ORCHESTRATION.md` (trap 254), `RESOURCE_BINDING.md:169` and `STRAY_STATUS.md:105-107`, and it
matters because those are the three places a reader quotes instead of re-deriving — and because the current
text tells them to throw away the one field on that line whose null is real. The five non-blocking items are
yours to take or leave; #1 and #2 are the two I would take, since they are in the same paragraph the PR is
already correcting. Everything else, including the finding that this PR's own new test arm was insensitive,
is verified and cleared.

mattias800:
## Blocking finding fixed — and I verified it against the source before accepting it

The reviewer is right, and I checked it independently rather than on the citation: `pc_res` is
`by_fetch_pc(in.pc)`, the resolver's **first** route, with no register-state guard. `srt_tag`/`key_res`
and `ud_alias`/`alias_res` are forced because `sreg_srt` and `sreg_ud_alias` populate only through
`record_scalar_write`, which is also what inserts into `sreg_written`. So the forced set is those four
field names over two routes; `pc_res` is a genuine measurement.

Three docs had substituted `pc_res` for `ud_alias`. The code never did — its comment names the right
four. `STRAY_STATUS.md` contradicted itself thirty lines later, where it explains that `pc_res=null`
was ambiguous and needed `PROSPER_DYNTRACE_FAIL=1` to separate its two causes.

Fixed in all three, and trap 254 now **states the mechanism per field and names `pc_res` as the one
that is not forced**, rather than giving a count. A trap row about a misleading diagnostic teaching
readers to discard the one meaningful null was the worst available error; the row says so.

## The `:7077` comment — same failure mode, my side

Verified: `if (in.opcode == 0xe6u)` returns from its own block earlier in the function, so a
BVH-piggybacked ConstantBuffer keyed by that pc can never reach the lookup. The sentence came verbatim
from #3313's review §1a and I shipped it into a code comment without opening the path — the
`#2049 → #2052` shape, exactly. The comment now states the precedence change (provable from the diff)
and says **no reachable instance is known**, and records why the claim is gone so it is not
re-introduced.

## Four non-blocking, all taken

* **Dedupe exactness.** The two buckets do not count the same thing: `why` is a fixed literal for the
  V# bucket but embeds the raw dwords for the degenerate one, so 97 counts `(block, slot)` pairs while
  4 counts distinct *payloads* that may all belong to one block. **"varies between blocks" is
  withdrawn.**
* **The per-draw citation**, re-pointed at evidence that carries it: both degenerate lines share
  `ud=0x21e0e55590`, so they show variation between **draws of one block** — derivable precisely
  because the dedupe key embeds the payload. The property that ruins the count is what makes the
  variation visible.
* **`static_assert`s** now reconstruct the opcode as the decoder does,
  `((w & 1) << 7) | ((w >> 18) & 0x7F)` — eight bits, not seven.
* **"could not have been read correctly even in principle"** withdrawn: #1634 read its own line
  correctly; what it could not do was tell a forced null from a measured one.

## Rebase note — another lane landed trap 255 mid-review

`origin/main` moved to `54168f0cd` (#3312) and `check_merge_result.py` reported a **conflict** in
`GAME_COMPAT_ORCHESTRATION.md`. Resolved by **re-applying my hunk onto their version**, not by taking
the file whole (trap 41). Verified afterwards: row 255 is **byte-identical to `origin/main`**, and the
merge gate reports `1..255, unique and ascending, none of the 255 rows deleted`.

## Verification, exact head `8d63a5885`

```
cmake --build prosper/build-linux -j4  -> BUILD_RC=0, 0 "error:" lines, [100%]
ctest --no-tests=error -j4             -> CTEST_RC=0, 100% passed out of 357, 0 skipped
                                          #214 image_descriptor_class_lookup .... Passed
test binary                            -> exit 0, 27 [ok], 0 [FAIL]

mutation (image_requirement collapsed to Either), re-run at the rebased head:
  MUT_TEST_RC=1  -> 25 [ok], 2 [FAIL]
    [FAIL] image_store binds the StorageImage past a Texture at the same SGPR (StorageOnly mapping)
    [FAIL] image_sample binds the Texture past a StorageImage at the same SGPR (SampledOnly mapping)
  restored -> exit 0, 27 [ok], tree clean
```

Docs gates: all-md green, `check_merge_result.py` clean merge and gate-passing against `origin/main`,
`diff --check` clean.

**Re-review requested.** No behaviour changed in this round — it is three doc clauses, one code
comment, and a widened `static_assert`.

mattias800 (commented):
# Independent re-review — PR #3315, exact head `8d63a5885`

**APPROVED.** The blocking finding on `c1dba8e96` is fixed in all three places, and fixed by stating the
mechanism per field rather than by patching the count — which is the correction that will survive being
quoted. Four non-blocking items are outstanding, one of which I want to name clearly because it is the
third phrasing of the same claim in two rounds; none of them blocks.

Re-review scope: source analysis only, no GPU, no build, no ctest. I did not re-run the author's build,
test binary, mutation matrix or CI. I **did** independently repeat the trap-41 / rebase check, the docs
gate, and `check_merge_result.py`, and I re-derived the two mechanism claims rather than accepting my own
earlier citations back.

---

## The rebase check, repeated independently — clean on all three fronts

`origin/main` moved `99ecee473 → 54168f0cd` (two commits: #3309, #3312) and the PR was rebased onto it.
Merge-base is now `54168f0cd`, i.e. `origin/main`'s head, so this is a strict fast-forward and the merge
result **is** the head tree.

**1. `GAME_COMPAT_ORCHESTRATION.md` — row 255 is byte-identical, proved by hash, not by eye.**

```
sha256(grep '^| 255 |' origin/main) == sha256(grep '^| 255 |' 8d63a5885)
  5913b694c2684797cc400f5d9acc591bca46c803f947cbd0a89271d95b7b0260   (both)
diff <(grep -v '^| 254 |' main) <(grep -v '^| 254 |' head)   -> empty
line counts 2075 / 2075 ; numstat 1 insertion, 1 deletion
```

So the entire file is byte-identical to `origin/main` apart from row 254's content. The other lane's row
survived the resolution intact.

**2. `STRAY_STATUS.md` is where the real exposure was, and it is clean.** That is the file the other lane
also rewrote in the rebase window (`+205/-2` between the two bases), so it is the one a `git checkout --`
would have silently reverted. Three checks, all green:

* every one of the **203 lines** that lane added is present at `8d63a5885` (`missing_count=0`);
* neither of its **2 deleted lines** was resurrected;
* all **11 lines this PR deletes** exist at the old base `99ecee473` — i.e. every deletion is the author's
  own pre-rebase text, not someone else's. That is the check that actually catches trap 41, because a
  clobber produces no conflict and no failing check, exactly as the author said.

**3. Everything else had no exposure.** `origin/main` did not touch `RESOURCE_BINDING.md`,
`rdna2_emit_alu.cpp`, `shader_resources.hpp` or the test in that window.

**4. Gates, run by me.**

```
check_numbered_table.py --ordered --baseline origin/main   -> 1..255 unique and ascending,
                                                              none of the 255 base rows deleted   rc=0
check_merge_result.py --base origin/main --head 8d63a5885  -> clean merge; merged file passes    rc=0
git diff --check                                           -> rc=0
Claude-Session trailers 0 ; Co-Authored-By on both commits
statusCheckRollup on 8d63a5885: 9 SUCCESS, 2 SKIPPED, 0 failures ; MERGEABLE
```

Worth recording: `check_merge_result.py` failed **loudly** when I first invoked it without its sibling
script (`rc=2`, *"this says NOTHING about the merge result"*) instead of returning a green-looking zero.
That is the "prefer experiments that detect their own invalidity" rule built into a tool, and it is why I
trust the second run.

---

## The blocking finding is fixed, and fixed the right way

Trap 254 now says the line prints five route fields, that four are forced — `srt_tag` and the `key_res` it
gates, `ud_alias` and the `alias_res` it gates, *"`sreg_srt` and `sreg_ud_alias` are populated only by
`record_scalar_write`, the same function that inserts into `sreg_written`"* — and that **`pc_res` is not**,
because it is `by_fetch_pc`, keyed by the instruction with no register-state guard, and is the resolver's
first route. `RESOURCE_BINDING.md:169` and `STRAY_STATUS.md:225-235` carry the same carve-out.

I re-derived both halves at this head rather than trusting my own earlier note: `pp = rt->by_fetch_pc(in.pc)`
at `rdna2_emit_alu.cpp:7130` and `image_by_fetch_pc` at `:7087`, neither guarded; `sreg_written.insert(reg)`
and `rs.sreg_ud_alias[...]` in the same function at `rdna2_to_spirv_internal.hpp:5085`/`:5094`.

Two things I want to credit specifically, because they are the difference between a patch and a fix:

* **The row states the mechanism per field instead of a count**, and says why (*"so state the mechanism per
  field, never a count"*). A count is what made the first draft wrong and unfalsifiable at a glance.
* **The row records its own error** — that the first draft listed `pc_res` among the forced fields and
  omitted `ud_alias` — so the next editor cannot reintroduce it by "simplifying". Same for the removed BVH
  sentence in the code comment. That is the part that pays off in six months.

Also fixed as asked: `#1634 read its own line correctly; what it could not do was tell a forced null from a
measured one, because the field that separates them was not yet printed` — that is the precise scoping, and
it is better than what I suggested.

---

## Non-blocking #3 (BVH reachability) — fixed, and I re-verified the deletion was warranted

`rdna2_emit_alu.cpp:7079-7086` now states only the precedence change, says no reachable instance is known,
and records why the claim was removed. I re-traced it independently at this head rather than citing my own
earlier finding back: the `if (in.opcode == 0xe6u)` block opens at `:6807` and closes at `:7030`, and every
exit is a `return` — the early `ok = false; return true;`, the mid-block `return true;`, and the final
statement before the closing brace is `return true;`. The class-filtered lookup is at `:7087`. So a
ConstantBuffer keyed by an `IMAGE_BVH_INTERSECT_RAY` pc can never be queried there, and the removed sentence
was false, not merely unproven.

The comment naming this as the `#2049 → #2052` shape, in the file where it happened, is the right place for
it — and the coordinator is right that the relay is part of the failure, not just the authoring.

---

## The other three findings

* **Dedupe (`STRAY_STATUS.md:323-331`)** — fixed, and *sharper than my finding was*. My version said the key
  is `(block, slot, reason)`; the correct statement, which the doc now makes, is that the two buckets do not
  count the same thing: `why` is a fixed literal for the V# bucket, so 97 counts `(block, slot)` pairs,
  while it embeds the raw dwords for the degenerate bucket, so 4 counts *payloads that may all belong to one
  block*. I re-read `agc_shader_layout.cpp:954` and `:984-989` and confirm both halves. *"No distribution
  over blocks is derivable from the 4"* is exactly the right stopping point.
* **`static_assert` (`test_...cpp:45-53`)** — fixed exactly: `mimg_opcode()` now reconstructs the decoder's
  own `((w & 1) << 7) | ((w >> 18) & 0x7F)` (`rdna2_decode.cpp:888`), with the reason recorded.
* **Window-survey clause** — unchanged from `c1dba8e96` and still exact.

---

## Outstanding, all non-blocking

### A. `STRAY_STATUS.md:347` — the replacement claim is *also* not carried by the quoted evidence

> **Both of those lines carry the same `ud=0x21e0e55590`** — one block, two different payloads — so what
> they show is variation *between draws of one block*, which is the stronger reading and the one the section
> needs.

The document shows the `ud=` for **one** of the two. `:317-320` quotes the `[sharp] … degenerate T#` line
with payload `0ae8fb54 …`. The second payload, `2f70fbb8 …`, appears **only** in the summary table at
`:344` — no `[sharp]` line, no `ud=`, no quoted log line anywhere in the file (I grepped: `2f70fbb8` occurs
exactly once, in that table row).

I am not saying it is wrong — you have the run log and I do not, and the claim is plausible. I am saying the
doc does not carry it, and this is the **third phrasing of the variation-axis claim in two rounds** ("varies
by draw" → "varies between blocks" → "between draws of one block"), each time asserted in bold as the
reading the section needs. That pattern is the reason to fix it by *pasting the second `[sharp]` line*
rather than by rewording again — one line, from evidence you say you already read, and the claim becomes
self-verifying for the next reader.

Same paragraph, `:350`: *"The third payload, from `[resdump]` on the same stage, is a third draw."*
`[resdump]` is a different instrument keyed on the code address; nothing shown ties it to a distinct draw.
That is an inference stated as a fact.

Not blocking, because the conclusion the section rests on — *"in none of them did the declared image slot
hold an image descriptor"* — is carried by the table regardless of which block or draw each payload came
from. Both sentences are decoration on a load-bearing claim that stands without them.

### B. `RESOURCE_BINDING.md:169` — the opening clause still covers the field it now exempts

The cell opens *"**Falsified, and the emptiness carried no information at all.** Those four fields —
`srt_tag`/`key_res` and `ud_alias`/`alias_res` — …"*, while the hypothesis in the left column names
`srt_tag`, `key_res`, **`pc_res`** and `alias_res`. So "Those four fields" silently swaps referent, and a
reader who stops at the first bolded clause still takes `pc_res`'s emptiness as information-free. The
bolded carve-out two sentences later rescues anyone who reads on, which is why this is minor — but the
first clause is the part that gets quoted. Suggest: *"Falsified — and for three of the four routes the
hypothesis names, the emptiness carried no information at all."*

### C. Trap 254 — "five routes … the sixth" counts fields, not routes

The opening says the line *"prints five routes … and did **not** print the sixth, `by_sgpr_base(SRSRC)`"*.
Those are five *fields* over four *routes* — `by_srt_offset`, `by_fetch_pc`, `by_sgpr_base(ud_origin)`,
`by_sgpr_base(SRSRC)` — because `srt_tag`/`key_res` are one route's gate and result, exactly as the row
itself then explains. "Five fields, four routes" would be exact, and in a row whose lesson is *"state the
mechanism per field"* the distinction is the row's own subject.

### D. Cosmetic — `STRAY_STATUS.md:234` is 149 columns

A reflow leftover from the rewrite. It and one pre-existing line are the only two non-table lines over 120
in the file, so the wrap is otherwise uniform. Worth fixing for a practical reason beyond tidiness: a phrase
split across an irregular reflow is invisible to `grep`, which is a recorded hazard in this repo.

### E. Carried and not taken, immaterial

The clause noting that `:200` (store resolves a lone StorageImage) and `:228` **together** pin
`0x08 -> StorageOnly`, where `:228` alone pins only "not Texture-admitting". The existing "control for the
arms below" label ties them adequately. Mentioned only so it is on the record as considered.

---

## Sensitivity: my earlier verification transfers to this head, and here is why

The delta from the head I reviewed is **exactly two hunks, neither of which executes**: the BVH comment
rewrite, and the `static_assert` widened to a `constexpr` reconstruction of the decoder's expression. No
statement in `emit_alu` changed. So the arm-by-arm analysis I did on `c1dba8e96` carries unchanged:

* `:228` `[Texture, StorageImage] @ s8`, op `0x08` — correct mapping skips the Texture and takes `:7320`;
  under `Either` the lookup stops on the Texture, `:7320` is false, and `:7655` rejects `0x08`. **RED.**
* `:236` `[StorageImage, Texture] @ s8`, op `0x20` — under `Either` the lookup stops on the simg, `:7320`
  **is** taken, and `:7356` (`!is_ld && !is_st && !is_atomic`) rejects. **RED.**
* All 25 others stay green, for the reasons enumerated in my first review. The two flips are sensitive to
  the mapping specifically, and `:215` remains correctly labelled as the defence-in-depth arm it was
  measured to be.

I did not rebuild to re-confirm the reported `MUT_TEST_RC=1` / 25-green matrix; I am reporting that the
executable surface is unchanged from the head where I derived the same result by tracing, and saying so
rather than implying a re-run.

---

**Verdict: APPROVED** on `8d63a5885`. The blocking finding is resolved, the rebase is clean under an
independent trap-41 check on the file that was actually exposed, the merge-result gate is green *as of
now* — so merge promptly, since another lane landing after this invalidates it. Item **A** is the one I
would spend the extra minute on before merging: it is one pasted log line, and it closes the last claim in
this PR whose evidence lives only in the author's terminal.

PR #3316, opened 2026-09-04
fix(hle): return SCE_NET_ERROR_EINVAL for sceNetPoolCreate invalid args (#3300)

### LINKED ISSUE #3300: libSceNet: 0x80410118 is commented SCE_NET_ERROR_ENFILE, but 0x18 is 24 = EMFILE (ENFILE is 23)
## Summary

`prosper/src/hle/service/hle_service.cpp:2810`:

```cpp
if (!svc_ptrish(a0) || (int32_t)a1 <= 0) return 0x80410118ull; // SCE_NET_ERROR_ENFILE
```

The libSceNet error family encodes as `0x80410100 | <BSD errno>`, so `0x...18` is **24**, which is
`EMFILE` on FreeBSD. `ENFILE` is **23**, i.e. `0x80410117`.

## Why it is worth a line

The returned **value** is not obviously wrong — "too many open files" is a defensible answer for a
pool-create that cannot be satisfied, under either spelling — so this is a comment defect, not a
behavioural one, and no test can see it.

It still matters, because that comment is now one of the few places in the tree that names a
libSceNet constant, and the encoding was recently established from primary evidence: libSceHttp2
propagates raw libSceNet errors, and `0x80410124` is special-cased as *not* a failure at its
`sceNetConnect` site, which is `EINPROGRESS` = 36 = `0x24` and pins the scheme (#2894). Anyone
deriving a neighbouring constant by reading this line as ground truth lands one off — and a wrong
name attached to a real value is the kind of thing that gets believed rather than checked.

## Suggested fix

Either relabel the comment `SCE_NET_ERROR_EMFILE`, or change the value to `0x80410117` if `ENFILE`
was actually intended. Deciding which needs a look at what the shipped `libSceNet.sprx`
`sceNetPoolCreate` returns for these arguments — it is in the dump's `sprx/` directory as plain ELF,
so it is derivable rather than a judgement call.

Found while reviewing #3295; deliberately not smuggled into that PR, which touches `hle/net/` and
not `hle/service/`.

Refs #2894

wruwami:
…ILE (#3300)

libSceNet error codes are encoded as 0x80410100 | <FreeBSD errno>. In FreeBSD <sys/errno.h>, 24 (0x18) corresponds to EMFILE ('Too many open files by process'), while ENFILE is 23 (0x17, 0x80410117).

s_net_pool_create returns 0x80410118ull on invalid arguments (!svc_ptrish(a0) || (int32_t)a1 <= 0), which is SCE_NET_ERROR_EMFILE, but the comment incorrectly labeled it as SCE_NET_ERROR_ENFILE.

Correct the comment to SCE_NET_ERROR_EMFILE (0x80410100 | BSD EMFILE 24), add sceNetPoolCreate to test_hle_registered names list, and assert both the error code and valid pool creation.

Fixes #3300

### DIFF
--- a/prosper/src/hle/service/hle_service.cpp
+++ b/prosper/src/hle/service/hle_service.cpp
@@ -2851,7 +2851,7 @@ std::atomic<int32_t> g_npweb_user_context_id{1001};
 }
 HLE(s_net_pool_create) {
     svc_log("sceNetPoolCreate", a0,a1,a2,a3,a4,a5);
-    if (!svc_ptrish(a0) || (int32_t)a1 <= 0) return 0x80410118ull; // SCE_NET_ERROR_ENFILE
+    if (!svc_ptrish(a0) || (int32_t)a1 <= 0) return 0x80410116ull; // SCE_NET_ERROR_EINVAL (0x80410100 | BSD EINVAL 22, #3300)
     return (uint64_t)(uint32_t)g_net_pool_id.fetch_add(1);
 }
 HLE(s_ssl_init) {
--- a/prosper/tests/hle/test_hle_registered.cpp
+++ b/prosper/tests/hle/test_hle_registered.cpp
@@ -48,7 +48,7 @@ int main() {
         // services / dialogs
         "sceUserServiceGetInitialUser", "scePadOpen", "sceMsgDialogUpdateStatus",
         "sceSystemServiceHideSplashScreen",
-        // HTTP helpers
+        // HTTP / network helpers
         "sceHttpUriParse",
         // graphics (headless bring-up)
         "sceVideoOutOpen", "sceVideoOutSubmitFlip", "sceVideoOutGetFlipStatus",
@@ -67,6 +67,24 @@ int main() {
         uint64_t v = fn(0, 0, 0, 0, 0, 0);
         if (v != 1) { printf("  [FAIL] __ctype_get_mb_cur_max returned %llu, want 1 (the value, not a pointer)\n", (unsigned long long)v); fails++; }
     } else { printf("  [FAIL] not registered: __ctype_get_mb_cur_max\n"); fails++; }
+
+    // sceNetPoolCreate returns SCE_NET_ERROR_EINVAL (0x80410116 = 0x80410100 | BSD EINVAL 22) on
+    // invalid arguments, and a positive pool ID on valid arguments (#3300).
+    if (HleFn fn = Hle::lookup(nid_hash("sceNetPoolCreate"))) {
+        uint64_t err = fn(0, 0, 0, 0, 0, 0);
+        if (err != 0x80410116ull) {
+            printf("  [FAIL] sceNetPoolCreate(0, 0) returned 0x%llx, want 0x80410116 (SCE_NET_ERROR_EINVAL)\n",
+                   (unsigned long long)err);
+            fails++;
+        }
+        char dummy[16] = "testpool";
+        uint64_t id = fn((uint64_t)(uintptr_t)dummy, 4096, 0, 0, 0, 0);
+        if (id == 0 || (int32_t)id <= 0) {
+            printf("  [FAIL] sceNetPoolCreate valid call returned non-positive id %lld\n",
+                   (long long)(int32_t)id);
+            fails++;
+        }
+    } else { printf("  [FAIL] not registered: sceNetPoolCreate\n"); fails++; }
 
     // Sony's Dinkumware inline isspace uses (_Getpctype()[c] & 0x144), not the
     // incompatible MSVCRT bit layout. Pin representative C-locale masks and EOF/case slots.
--- a/prosper/tests/hle/test_hle_registered.cpp
+++ b/prosper/tests/hle/test_hle_registered.cpp
@@ -48,7 +48,7 @@ int main() {
         // services / dialogs
         "sceUserServiceGetInitialUser", "scePadOpen", "sceMsgDialogUpdateStatus",
         "sceSystemServiceHideSplashScreen",
-        // HTTP / network helpers
+        // HTTP helpers
         "sceHttpUriParse",
         // graphics (headless bring-up)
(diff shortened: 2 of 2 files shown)


mattias800 (changes requested):
Thank you for this — the arithmetic is right, and the encoding reasoning is exactly the kind of thing worth fixing. `0x80410118 - 0x80410100 = 0x18 = 24`, and FreeBSD's `EMFILE` is 24 while `ENFILE` is 23, so the existing comment was wrong.

I went and looked at the shipped library, because #3300 says explicitly that choosing between its two options *"needs a look at what the shipped `libSceNet.sprx`"* does. That library is present in the local dump set and is plain ELF (`ELF 64-bit LSB x86-64, version 1 (FreeBSD)`, 307,944 bytes, entropy 5.645 bits/byte — nothing decrypted). **The evidence changes which of the two options is right.**

## `0x80410118` does not exist in the real library

Scanning every 4-byte little-endian value in `libSceNet.sprx` for the `0x804101xx` family:

| constant | meaning | occurrences |
| --- | --- | --- |
| **`0x80410118`** | EMFILE (24) — the value prosper returns | **0** |
| `0x80410117` | ENFILE (23) | 1 |
| `0x80410116` | EINVAL (22) | **85** |
| `0x804101c8` | 200 | 111 |

The distinct low bytes the library actually uses are `[0, 2, 4, 9, 15, 16, 17, 19, 22, 23, 35, 36, 49, 56, 61, 63, 66, 164, 200, 201, 203, 204, 206, 220, 221, 222, 223, 225, 226, 228, 229, 230, 231, 232, 233, 235, 236]`. **24 is not one of them.**

So this PR relabels a comment on a constant the real library never returns. The comment becomes arithmetically correct and the *value* stays invented — which is the half #3300 flagged as needing the `.sprx` to decide.

## What `sceNetPoolCreate` actually does

`sceNetPoolCreate` is NID `dgJBaeJnGpo`, resolved via `PT_DYNAMIC` (the module is stripped of section headers) to VA `0x16e0`, 201 bytes. Disassembled:

```
0b: cmp DWORD PTR [rip+0x3e9ae],0x0     ; library-initialised check
12: je  0x67
...
67: call 0x5ae0                          ; errno location
6c: mov DWORD PTR [rax],0xc8             ; errno = 200
72: mov ebx,0x804101c8                   ;  -> returns 0x804101c8
...
79: mov eax,ebx                          ; inner call returned negative
7b: movzx ecx,bl
7e: mov r14d,0xcd                        ; default 205
84: and eax,0xff00
89: cmp eax,0x100                        ; if high byte is 0x01 ...
8e: cmove r14d,ecx                       ; ... use the low byte
```

Two paths, neither producing `0x80410118`: an uninitialised-library path returning **`0x804101c8`** (200), and an error-propagation path that maps the inner call's low byte when its high byte is `0x01`, defaulting to 205. Argument validation is delegated to the inner call at `0x8530`, not done in the prologue.

## Suggested change

#3300's second option — *"change the value"* — now looks like the right one, but to **`0x80410116` (EINVAL, 22)** rather than to ENFILE:

* prosper's guard is `!svc_ptrish(a0) || (int32_t)a1 <= 0`, which is an **invalid-argument** condition. "Too many open files", under either spelling, is not what a bad pointer or a non-positive size means.
* `0x80410116` is the most-used constant of the family in the real library (85 occurrences), so it is well within the space the guest can expect.

If you'd rather keep this PR to the comment fix and leave the value to a follow-up, that's a reasonable split — but then the comment should say the value is prosper's own and is **not** one the shipped library emits, otherwise the next reader takes `SCE_NET_ERROR_EMFILE` as an observed fact. Right now the PR body presents it as one.

## On the test

The two new assertions are welcome, and the valid-path arm does reach the non-error branch — `svc_ptrish` is `v >= 0x10000 && v < 0x7fffffffffff` (`hle_service.cpp:80`), which a host stack pointer satisfies. Two notes:

1. If the value changes per the above, `test_hle_registered.cpp` pins `0x80410118` in two places and both need updating with it. Worth deciding the value first so the test isn't written twice.
2. `sceNetPoolCreate` is now both in the `names` list (checked by `must()`) and re-checked by the new block's `else` branch, so the "not registered" failure can print twice. Harmless, just noting it.

Requesting changes on the value question rather than the comment — the comment fix itself is correct and I'd take it as-is if you'd prefer to scope it that way, with the caveat above.

wruwami:
Thanks for analyzing `libSceNet.sprx`!
As suggested, I have updated the error return value for invalid arguments (`!svc_ptrish(a0) || (int32_t)a1 <= 0`) to `0x80410116ull` (`SCE_NET_ERROR_EINVAL`, BSD EINVAL 22) instead of keeping `0x80410118`.
Regarding the tests:
1. Updated `test_hle_registered.cpp` assertions to check for `0x80410116`.
2. Removed `"sceNetPoolCreate"` from the `names[]` array to prevent duplicate `[FAIL] not registered: sceNetPoolCreate` logs, since it is already verified in its dedicated `Hle::lookup` test block below.

mattias800:
Heads-up rather than a review: I am splitting `src/hle/service/hle_service.cpp` into per-library files under #3735, and this PR's one-line `sceNetPoolCreate` change is in that file.

Nothing is merged yet and the first step (#3735 step 1, promoting shared helpers to a header) does not move your line. But a later step will, and the NID registration for `libSceNetCtl`/NP is headed for a new `src/hle/np/` file — so this PR will need its one line re-applied there rather than merging cleanly.

It is one line either way. Landing or rebasing this first is the cheapest ordering if you are still on it; if it is parked, say so and I will carry the change into the new file myself and credit it here.

wruwami:
Thanks for the heads up regarding the refactoring in #3735!

To avoid unnecessary merge conflicts while you're actively splitting hle_service.cpp, I'll close this PR for now.

Once the refactoring lands on main, I will re-apply the changes with 0x80410116 (SCE_NET_ERROR_EINVAL) and updated tests against the new file layout, then open a fresh PR.

Thanks!

Issue #3307, opened 2026-09-03
Compute-image writeback is 61% CPU re-tiling and 31% write-watch registration — answering #3149's open question


mattias800:
## Summary

#3149 ended by naming its own next step:

> *"Next step is to establish which of those two directions the block copier is running in per
> dispatch — input detile or result re-tile — and whether a dispatch whose result is unchanged still
> pays for it."*

**It is the result re-tile**, and it is 61% of compute-image writeback time.

## Measured

Stray title screen (`PPSA02101`), `origin/main` `fc21d46ca`, `reach-title-hold.pad`,
`PROSPER_NULL_PAGE=1`, 3840x2160, `PROSPER_COMPUTE_IMAGE_TIMING=1`. Aggregated over **4,291
`[compute-image-writeback]` records**:

| stage | total | share |
| --- | --- | --- |
| **`layout_ms`** | **2794.2 ms** | **61%** |
| **`watch_ms`** | **1417.1 ms** | **31%** |
| `cache_ms` | 286.7 ms | 6% |
| `pack_ms` | 66.0 ms | 1% |
| `notify_ms` / `map_ms` / `prepare_ms` | 17.4 ms | ~0% |
| total | 4584.6 ms | |

By program:

| program | writeback total | n | mean |
| --- | --- | --- | --- |
| `0x30112e0000` | 1405.2 ms | 325 | 4.324 ms |
| `0x300bc70000` | 1230.1 ms | 976 | 1.260 ms |
| `0x3011300000` | 1228.1 ms | 321 | 3.826 ms |
| `0x300dfa0000` | 620.1 ms | 648 | 0.957 ms |

`layout_ms` is the span from `pack_done` to `layout_done` at `live_compute.cpp:10085-10115` — the
`tile_surface` / `tile_surface_level` calls that convert the linear compute result **back into the
guest's tiled layout**. That is the `parallel_rows<sw64kb_copy<...>>` and `__memmove_avx512` cost
#3149 saw in `perf`, now attributed to a direction and a stage.

## Why this is worth acting on

The title screen renders at **9.76 fps** (F8 capture, 5.02 s window at t=175 s). `gpu-device` is
232.5 ms of that window — **~5%** — so the GPU is idle and the frame is host-bound. Compute is
1981.5 ms and renderer-resource 1895.5 ms, i.e. two nearly equal halves rather than one hot spot.

`watch_ms` at 31% is the second target and has had no attention: that is guest-write-watch
[...]

Issue #1710, opened 2026-08-01
Compute shaders read push constants that were never provided: the pipeline layout has no VkPushConstantRange and vkCmdPushConstants is never called


mattias800:
Found by the first-ever Vulkan validation pass over prosper's test suite (#1704). Two VUIDs, one
root cause, and it fires on **every** environment measured — Mesa 25.2.8 lavapipe (the CI runner's
exact driver) and RADV alike.

```
VUID-VkComputePipelineCreateInfo-layout-07987
  vkCreateComputePipelines(): pCreateInfos[0].stage SPIR-V (VK_SHADER_STAGE_COMPUTE_BIT) is using
  push constants, but no VkPushConstantRange were found in VkPipelineLayout ...

VUID-vkCmdDispatch-maintenance4-08602
  vkCmdDispatch(): Shader in VK_SHADER_STAGE_COMPUTE_BIT uses push-constant statically but
  vkCmdPushConstants was not called yet for pipeline layout ...
```

## Where

`prosper/frontends/shared/live_compute.cpp:5080`

```cpp
VkPushConstantRange push_range{};
if (!item.user_sgprs.empty()) {
    push_range.stageFlags = VK_SHADER_STAGE_COMPUTE_BIT;
    push_range.size = static_cast<uint32_t>(item.user_sgprs.size() * sizeof(uint32_t));
    plci.pushConstantRangeCount = 1;
    plci.pPushConstantRanges = &push_range;
}
```

The layout gets a push-constant range only when `item.user_sgprs` is non-empty. The recompiler,
however, emits the `PushConstant` storage-class block unconditionally for kernels that reference
user SGPRs (`prosper/src/gpu/rdna2_to_spirv.cpp:2184-2188`), so a dispatch whose `user_sgprs` vector
arrives empty gets a module that *statically reads* push constants through a layout that declares
none, and nothing is ever pushed.

## Why it matters

Reading a push constant that was never written is undefined. Whatever the kernel computes from
those user-SGPR values — a base address, a dispatch dimension, a resource index — is whatever the
driver happens to leave in that register file. This is the same shape as #1690: correct-looking on
the driver you tested, silently wrong data on the next one. `game_compute_exec` currently produces
8 such pipeline creations and 20 such dispatches per run.

## Suggested fix

Decide the contract at one place instead of two: either the recompiler reports the push-constant
size it emitted and `live_compute` always creates a range matching it (pushing zeros when
[...]

PR #3317, opened 2026-09-04
perf(compute): pool the per-dispatch full-surface intermediates, and name what layout_ms/watch_ms really are

mattias800:
## The problem

[#3307](https://github.com/mattias800/prosper/issues/3307) measured Stray's compute-image writeback
and named two targets, `layout_ms` (61%) and `watch_ms` (31%). Both were re-measured here against
standalone micro-benchmarks of the exact code they run, and **neither can be made faster** — but a
third cost, invisible to that measurement because it lands in `setup` rather than writeback, can.

[#3309](https://github.com/mattias800/prosper/pull/3309) found that two thirds of the *graphics*
texture materializer's cost was the allocator rather than the decode, and pooled its intermediates.
**The compute path was still doing exactly the same thing**, at the same sizes, on the same
surfaces.

`live_compute.cpp` gave every dispatch a fresh `make_unique_for_overwrite<uint8_t[]>` for each
full-surface intermediate: the linear buffer a tiled storage image is detiled into before it is
seeded, the one a non-exact writeback is packed through, and the two on the sampled upload path.
Stray's dominant storage geometry is 3840×2160 at 8 bytes per texel — **63.3 MiB**, past glibc's
32 MiB mmap threshold, so each is an `mmap`, 16,384 page faults as the detile touches it, and a
`munmap`.

Measured on this machine (Strix Halo), allocate-and-fill against fill-into-a-retained-buffer:

| extent | fresh alloc + fill | reused + fill | penalty |
| --- | --- | --- | --- |
| 4 MiB | 0.107 ms | 0.077 ms | 0.031 ms |
| 16 MiB | 0.644 ms | 0.459 ms | 0.185 ms |
| **33 MiB** | **7.371 ms** | **1.413 ms** | **5.958 ms** |
| **64 MiB** | **14.396 ms** | **3.117 ms** | **11.279 ms** |

The cliff between 16 and 33 MiB is glibc's dynamic `M_MMAP_THRESHOLD`, whose cap is 32 MiB: below it
the heap recycles the block, above it every allocation is a fresh mapping.

On the dominant geometry Stray's storage-gate census reports `seed_skip=2469` of `evaluated=4879`,
so roughly **half of those bindings seed** — and each seed is one of these allocations.

## The approach

`ScratchBuffer` in `decode_scratch.hpp` is a pooled stand-in with the same shape as the
`std::unique_ptr<uint8_t[]>` these sites used (`get()` is null until one `reset()`), so **every
existing `.get()` at the call sites is unchanged** and the diff is the allocation line only.

**The `zero_fill` argument is the contract, and it is the one thing here that could be got wrong
silently.** `make_unique_for_overwrite` leaves the bytes indeterminate — but at these sizes those
bytes are *fresh kernel pages* and therefore read as zero, so a call site can depend on zeroing it
never asked for. A pooled buffer holds the previous tenant instead. So:

- **storage seed** — zeroes when either the detiler cannot cover its destination for this element
  size (`detile_writes_whole_destination`, the 64 KiB fallback that memcpys a bounded prefix), or
  the allocated extent is `volume_texels`, which counts **six** cube slices while the only branch
  that shape reaches fills one.
- **BCn block-detile** — zeroes on the same detiler predicate; its layer loop covers the extent.
- **sampled upload** — always zeroes. Its branches fill a `sampled_layers`-slice prefix that is not
  always the whole `volume_texels` extent.
- **writeback pack** — never zeroes. Both fills run over all `texels` by construction.

## The fail-visible arm

`PROSPER_DECODE_SCRATCH_POISON=1` fills every lease with `0xCD` before handing it out, so a call
site that silently depended on the incidental zeroing fails loudly instead of producing wrong
pixels. Registered as `game_compute_exec_scratch_poison` against the real compute backend, which
executes the storage seed/writeback path including a `PARTIAL-COVERAGE` binding.

**Stated rather than buried: that arm does not discriminate at the sampled site.** Flipping the
sampled path's `zero_fill` to `false` and re-running it leaves all three `game_compute_exec` arms
green, so its fixtures do not reach a shape whose sampled fill is short. That is why the sampled
site takes the zero unconditionally — the conservative direction cannot be wrong — rather than on a
predicate the suite cannot check.

## What `layout_ms` and `watch_ms` actually are

Recorded in `STRAY_STATUS.md`'s `## Ruled out`, because #3307 attributed both wrongly:

- **`layout_ms` is the CPU re-tile running at memory-bandwidth speed.** At Stray's geometry
  (3840×2160, bpe 8, tile mode 27 = `SW_64KB_R_X`) `tile_surface` is **1.83 ms best / 2.21 ms
  mean**, against **2.75 ms** for a single-threaded `memcpy` of the same 63.3 MiB — it moves 127 MiB
  of traffic faster than one core copies 63. It is already threaded over block rows (7.77 ms at one
  thread, 1.69 ms at eight), and its bpe=8 AVX2 inner loop already stores the longest contiguous run
  the swizzle permits: consecutive `x` differ only in address bit 3, and `x+2` jumps to bit 5, so
  16 bytes is the maximum and the code stores 16. **It can only be reduced by re-tiling fewer
  surfaces.**
- **`watch_ms` is not the per-page index walk** — that loop, one `pages_by_addr.find` per 4 KiB plus
  the hit vector and generation bumps, was modelled exactly and costs **0.05 ms per 64 MiB when
  every page in the range is watched**, 0.0035 ms when none is, against a measured mean of 0.33 ms.
  **That is the only half of this that is established.** An earlier revision of this PR recorded, as
  fact, that the stage is the `mprotect` `set_pages_armed` issues (0.39 ms idle / 0.63 ms loaded per
  64 MiB — numbers that stand, and that fit the mean). **It is retracted to a hypothesis**: those
  were measured standalone, and nobody has yet measured whether that call happens inside the span.
  Two other candidates are live — the notification blocking on the one global state mutex, which
  `query()` holds while walking a whole registration (89,938 queries over 193,248 watched pages on a
  live run), and the poison-coverage scan that shares the span.

New `GuestWriteWatchStats` fields decide it: `host_write_notifies` / `host_write_no_alias` /
`host_write_pages_hit`, `lock_contended`, and the machinery's total protection calls and bytes,
printed by `PROSPER_RENDER_TIMING`'s existing `write_watch` line. The third candidate needs **no
code at all** — every `[compute-image-writeback]` record already prints `poison=`, so re-aggregating
an existing `PROSPER_COMPUTE_IMAGE_TIMING` log by that field settles it for free, and that is the
cheapest thing to do first.

- **Affected:** every compute dispatch that stages a full-surface intermediate — storage seed,
  sampled upload, BCn decode, non-exact writeback pack. Also every `mprotect` the write-watch issues
  (two relaxed atomic increments).
- **Unaffected:** the writeback's dominant path on Stray takes `tile_mapped_bytes`, which allocates
  nothing and is untouched. No skip is added and no work is elided: this changes where bytes are
  staged, never which bytes are produced.
- **Memory:** `DecodeScratchPool` is thread-local with the existing 512 MiB budget and a 4-buffer
  cap, and the compute path holds at most one lease at a time. Compute and the renderer share one
  pool when they share a thread.

## Risks

- A call site that depended on the fresh mapping's zeros would now read the previous tenant. This is
  the whole reason for `zero_fill` and for the poison arm; the residual is the sampled site, handled
  by always zeroing.
- `PROSPER_DECODE_SCRATCH_MB=0` restores the pre-pool behaviour exactly, and
  `decode_scratch_no_retention` asserts the complement of the default arm so neither can pass by
  accident.

## Verification

Linux, `df1bbae32` + this branch, built in the `ps5ys` container.

```
cmake --build prosper/build-linux -j4              -> exit 0
ctest --no-tests=error -j4                         -> 360/360 passed, exit 0, 0 skipped
python3 prosper/tools/env/check_diag_gates.py --selftest ; ... check_diag_gates.py prosper   -> exit 0, 0
python3 prosper/tools/output/check_ascii_output.py prosper                                   -> exit 0
python3 prosper/tools/env/check_env_numeric_arms.py prosper                                  -> exit 0
        --baseline <origin/main> prosper/docs/GAME_COMPAT_ORCHESTRATION.md                   -> exit 0
git ls-files '*.md' -z | xargs -0 python3 prosper/tools/docs/check_numbered_table.py         -> exit 0
git diff --check                                                                             -> exit 0
```

**Mutation arms, by exit code** (each applied to the source, rebuilt, then reverted):

| mutation | `ctest -R decode_scratch` |
| --- | --- |
| none | **0** (3/3 passed) |
| `ScratchBuffer::reset` leases from a private non-retaining pool instead of the shared one | **8** (2 failed) |
| `zero_fill` ignored | **8** (2 failed) |

| mutation | `ctest -R game_compute_exec` |
| --- | --- |
| none | **0** (3/3 passed) |
| sampled site's `zero_fill` flipped to `false` | **0** — the arm does not reach that shape; see above |

No snapshots were run: the change stages bytes differently and produces the same bytes, and no
route-level evidence was collected for it.

## Not measured live

**This PR carries no live fps number.** The GPU is held by another consumer, so the allocation cost
is established by micro-benchmark of the exact allocation shape, not by a run of the title. The
requested A/B is in the issue thread.

Refs #3307
Refs #2883

## The four live measurements (run by the coordinator, 2026-09-04)

**1. Phase timing confirms the redirection, decisively.** Over 62,556 dispatches:
`setup_ms` **51,062.6 ms** against `writeback_ms` **46,775.3 ms** (`writeback_images_ms` 45,513.8,
`setup_buffers_ms` 3,830.9, `setup_validate_ms` 628.9). Setup is the larger half, ~46.6 s of it in
the image path, and only `writeback_*` had ever been quoted.

**2. The A/B on this PR is NULL.** Three arms against two of `main`:

| build | arms | rendered fps | mean |
| --- | --- | --- | --- |
| `origin/main` | 2 | 8.76, 9.75 | 9.25 |
| this PR | 3 | 8.96, 9.23, 10.06 | 9.42 |

Ranges overlap. **The pooling does not move the frame rate on this route, and no fps claim is
attached to this PR.**

That is consistent with the benchmark rather than against it, and the same run says why: the
image-source snapshots average **19.5 MiB** and the write-watch registers **193,248 pages across 86
registrations, ~8.8 MiB each**. The dominant surfaces are under glibc's 32 MiB `M_MMAP_THRESHOLD`
cap, where the measured penalty is **0.185 ms at 16 MiB** rather than 11.279 ms at 64 MiB — the
dynamic threshold adapts upward after the first free at 19.5 MiB and never can at 63.3 MiB. **So the
change is worth roughly 60x less on this route's population than on the 4K surface that motivated
it.** It is still the correct shape and it still removes the large cost where the large surfaces
occur; it is simply not this route's bottleneck. `main` alone spans 11% across two arms, so no
single-arm fps comparison on this route is evidence.

**3. Falsified my `mprotect` reading — see the retraction above, and instrument trap 256.**

**4. The volume that is actually left**, from the same run: `snapshots=13001 253284.7 MiB`, of which
`rmw=7945 181066.4 MiB`. **247 GiB of full-surface image-source copying, 176.8 GiB of it
read-modify-write** — the same population as #3307's 21 `PARTIAL-COVERAGE` bindings at a uniform
~0.75 poison-survival ratio. ~113 MiB per frame, ~1 GiB/s at 9 fps; 30 fps would need ~3.4 GiB/s.
**A volume problem, not a per-call cost problem**, which is exactly why pooling the allocator
changed nothing. Recorded in `STRAY_STATUS.md`; it is the next lever, together with making
`import_live_compute_storage_image` hit.

## CI

All required checks green: Contribution shape, Docs, Linux, Linux App, Linux App Package, macOS App
(SDL3 + MoltenVK), macOS (x86_64 / Rosetta 2), Windows App, Windows MinGW — 9 success, 0 failure
(Progress tracker and Publish Release skipped as usual).

The first head failed one check and it is worth naming, because it is the ledger working as designed
rather than a defect: `game_compute_exec_scratch_poison` runs the same fixtures as its two siblings,
so it reported the same two already-deferred push-constant VUIDs (#1710) — and `tools/vkval` scopes
a deferral to the tests it was *measured* in, deliberately, because several of those ids are
catch-alls. The second commit adds the new arm to those two ledger entries. Same defect, same issue,
one more test name; no new suppression.

Head `5a7f7f1f0` adds the retraction, the two field renames and the contention counter. This PR
touches a numbered table (instrument trap 256), so **run `prosper/tools/docs/check_merge_result.py`
immediately before merging** — a green `Docs` job describes a merge that may no longer exist.


mattias800 (commented):
# Independent review of #3317 @ `5a7f7f1f0` — **REJECTED**

Two blocking findings, both in the durable record rather than in the code. **The code is verified
correct** — I derived the zero/fill coverage per site per branch rather than taking the list in the
body, and found no path on which a consumer can read an unwritten byte of a pooled buffer. I will
approve on a head that fixes B1 and B2; nothing else here needs to change to merge.

No GPU was used. Gates were re-run on this head in a detached worktree (results at the end).

---

## What I verified by hand — the four `zero_fill` sites

This is the whole review, so the derivation is stated rather than asserted. In each case I traced
every branch that can write the lease and every consumer that can read it.

### Site 1 — storage seed, `live_compute.cpp:7667-7684`

`zero_fill = !one_branch_fills_the_extent || !detile_writes_whole_destination(r->tile_mode, guest_texel)`.

The enclosing storage branch is guarded at `:7487` by
`if (r->tile_mode && !tile_mode_is_tiled(r->tile_mode)) skip_image(...)`, so a non-zero `tile_mode`
here is always a real tiled mode. Extent is `linear_guest_bytes = volume_texels * guest_texel`;
consumers are `storage_unpack_range` / the exact memcpy (both over `texels`) and, under
`poison_verify`, `bi.seed_linear.assign(unpack_source, unpack_source + linear_size)` — so the full
extent is read.

| branch | fill | covers extent? |
| --- | --- | --- |
| `seed_skip` / `seed_from_imported` | — | `linear_size == 0`, no lease |
| `renderer_owned` | — | not allocated; source is `live_target.pixels` |
| `tile_mode && dim_3d && depth>1` (`:7702`) | `detile_volume` | yes, or returns false → `skip_image` + `break`, lease unread. `sw64kb_rx_volume_copy` / `s3_volume_copy` write every `(z,y,x)` (memcpy or memset); their `el == UINT32_MAX` early-outs return **false**, not a short write |
| `dim_2d_array && depth>1` (`:7709`), untiled | per-layer row memcpy | exactly `linear_slice` per layer × `r->depth` |
| …, `in_mip_tail` | `detile_surface_level` | see below |
| …, else | `detile_surface` | the predicate's declared domain |
| `tile_mode && in_mip_tail` (`:7740`) | `detile_surface_level` | see below |
| `tile_mode` (`:7747`) | `detile_surface` | predicate's domain |

The `*_level` variants resolve to `detile_elements_level` (`tile.cpp:1860`), and the answers happen
to coincide with the predicate at every element size: untiled → memcpy + memset of the tail (full);
`Sw256BS` → `sw256_copy` iterates every `(x,y)` (full); 4 KiB → `sw4kb_level_copy` iterates every
`(x,y)` and `sw4kb_lookup` defaults rather than early-exiting (full, possibly with the wrong
pattern, which is a pre-existing separate matter); 64 KiB with an unsupported `bpe` →
`sw64kb_level_copy` returns **having written nothing**, and that is exactly the case where
`detile_writes_whole_destination` returns false and the caller takes the zero.

**Conclusion: no unwritten byte is reachable at site 1.**

### Site 2 — BCn block-detile, `live_compute.cpp:8097`

Extent `linear_slice * layers`; the `remap` loop covers all `layers` and each of the three fills
covers `linear_slice` (`detile_elements_level`, `detile_elements` — the predicate's second declared
entry point — or the untiled row memcpy). The consumer, `bc_decode_surface(..., linear_slice)` per
layer, reads exactly what was written. **Safe.**

### Site 3 — sampled upload, `live_compute.cpp:8153` (`zero_fill=true` unconditionally)

Safe by construction, and the zero is **genuinely load-bearing, not merely conservative**: this site
computes `bpt = cb * nc`, which can be 3, 6 or 12 for RGB layouts, and on a 64 KiB tile mode
`sw64kb_copy` then falls back to `memcpy(dst, src, min(n, tiled_bytes))` (`tile.cpp:765-769`), which
can stop short. So a real short-fill shape exists here.

I do not, however, reproduce the *reason* given in the comment at `:8159-8162`. `remap` is
`tile_mode || (cube_face_as_2d && layer_stride_bytes) || (dim_2d_array && depth>1)`, and every one
of those three disjuncts selects a branch below that fills `volume_texels * bpt`:
`cube_face_as_2d` implies `img_dim==3`, hence `!dim_2d_array` and `!dim_cube_stacked`, hence
`sampled_layers==1`, hence that one slice *is* the whole extent. The "cube stacked, untiled" case
the comment names cannot allocate at all — `dim_cube_stacked` makes `cube_face_as_2d` false and
`dim_2d_array` false, so with `tile_mode==0` `remap` is false. The conclusion (always zero) is
right; the stated reason is not the one that makes it right. Not blocking — the comment argues for
the safe side — but worth correcting, since the reason is what the next reader will reuse.

### Site 4 — writeback pack, `live_compute.cpp:9925` (`zero_fill=false`)

`storage_pack_range` / `storage_pack_texel` write `min(ncomp,4)` components at stride
`guest_texel = cb * nc`; `num_components` originates in `rdna2_buffer_format`
(`agc_shader_layout.cpp:198`) and is ≤ 4, so per-texel coverage is exact, and the format is gated by
`storage_pack_supported` at `:7494`. The exact-bytes path memcpys all `texels` via
`parallel_compute_texels`, whose range decomposition I checked covers `[0, count)` including the
`std::system_error` fallback. When `tile_mapped_bytes` is true nothing fills the lease — and nothing
reads it either, because `layout_source` then selects `native_texels`. **Safe, and the body's
"every fill below covers the whole extent" is accurate.**

### Confidence

**High that no consumer can read an unwritten byte of a pooled buffer**, on all four sites, on every
branch I could reach by construction. The residual is not a suspected hole but a maintenance one,
recorded as N4 below.

---

## Blocking

### B1 — the `watch_ms` `## Ruled out` row does not reflect the measurement that has since been posted, and the issue's newest comment says the opposite

`prosper/docs/STRAY_STATUS.md:919` ships "*nobody has yet measured whether that call happens in this
span at all*". This head was committed 2026-09-04 09:41 +0200 (07:41 UTC); #3307's comment at
**08:01 UTC** withdrew the falsification that row was written in response to, published the untruncated
line (`page_protect`'s predecessor at `18210 calls/429447.0 MiB`, `pages_hit=62103696`,
`host_write=55651`), and concluded "*the `watch_ms` attribution in #3317 stands*".

**I am not asking you to adopt that conclusion — I think the row's caution is right and the comment
over-reads.** `protect_calls` is process-global (`guest_write_watch.cpp:592`) and cannot be
attributed to a span; and `pages_hit` counts *index hits*, not protection changes, because
`protection_runs` skips any page whose protection already matches what is being asked
(`guest_write_watch.cpp:749`) — so a notify over already-disarmed pages issues zero `mprotect`s.
`lock_contended` did not exist in that run at all, so candidate 2 remains untouched.

But as it stands, master's doc and the issue's latest comment will flatly disagree with no
reconciliation, and the charter tells the next agent to read the doc first. Two sentences: cite the
numbers, and say why globals plus `:749` still do not isolate the span. Otherwise the correction
lives only in an issue comment, which is the failure mode `## Ruled out` exists to prevent.

### B2 — the null's explanation uses a **mean** to characterise a cost with a **step**, and the number cited does not support the conclusion

`STRAY_STATUS.md:250-256` (and the PR body) argue: snapshots average 19.5 MiB, which is under
glibc's 32 MiB cap, therefore "**The pooling is worth ~60x less on the population that dominates
this route**".

A mean is the wrong statistic when the cost function has a discontinuity inside the distribution.
253,284.7 MiB over 13,001 snapshots is equally consistent with, say, 3,000 allocations at 63.3 MiB
(190 GiB — most of the volume, and by your own table most of the *cost*) plus 10,001 averaging
6.3 MiB. The mean would still read 19.5.

Your own dominant-geometry census cuts the same way: `3840x2160 bpe=8 evaluated=4879 seed_skip=2469`
means roughly 2,410 seeds at 63.3 MiB; at the table's 11.279 ms that is **~27 s** against a
`setup_ms` total of 51,062.6 ms. If that were the real saving the A/B could not have come back null,
so *something* in the chain is wrong — and "the surfaces are mostly small" is only one of the
candidates. (Others: the micro-benchmark's tight-loop allocator behaviour does not reproduce in the
run; the census and the phase-timing run are different runs on different heads.)

The statistic that would settle it is the fraction of *leases actually taken* that exceed 32 MiB,
and the sites already know `linear_size` — a bucketed count behind the existing timing gate is
nearly free. Until then, state the explanation as a hypothesis. As written it will tell the next
lane "these surfaces are small, don't bother pooling", which may be exactly backwards.

---

## Non-blocking

**N1 — `page_protect=` still contains the retired `protect=` as a suffix.** The trap-256 row
(`GAME_COMPAT_ORCHESTRATION.md:384`) says the fix makes it so "neither name contains the other".
True
of the two *new* names; false of the *retired* one, which is what every prior log, note and this
issue's whole thread uses. `grep -o 'protect=[0-9]*'` on a new line now matches inside
`page_protect=18210` and returns the **call** count where the habit expects the **failure** count —
trap 256 with the sign flipped. `protect_calls=` or `mprotect_calls=` avoids it. (While there:
`host_write=%llu notifies/%llu no_alias` prints a bare `no_alias` token after a value, which is not
a `key=value` pair and does not grep like one.)

**N2 — `page_protect=0` is a false zero on Windows.** `g_protect_calls` is incremented only in the
POSIX `watch_mprotect` (`guest_write_watch.cpp:592-598`); the Windows arm's `VirtualProtect`
(`:195`) is not counted, and the Windows `guest_write_watch_stats()` at `:403` cannot see the
counter anyway. The other three new fields are *honest* zeros there
(`guest_write_watch_notify_host_write` is a no-op at `:503`), which makes it worse: three truthful
zeros next to one misleading one. Count it on Windows, or suppress the field off-POSIX.

**N3 — `one_branch_fills_the_extent` is a tautology inside the storage branch.**
`live_compute.cpp:7676` reads `dim_3d || dim_2d_array || sampled_layers == 1u`, but
`shader_2d_cube_view` requires `!bi.storage` (`:6284`), so `dim_cube_stacked` is false here and
`sampled_layers` is `dim_2d_array ? depth : 1` by construction — the disjunction cannot be false.
Harmless, but the comment above it asserts a six-slice cube hazard that the enclosing branch
excludes, and an unreachable hazard in a comment is a hazard someone later "fixes".

**N4 — the predicate is used outside the domain its own header declares.** `tile.hpp:112` says
explicitly: "*It does NOT extend to `detile_volume` or the mip-tail `*_level` variants, whose
destinations are not covered here.*" Sites 1 and 2 use it to authorize skipping the zero for exactly
those. I checked each copier and the answers coincide today (table above), so this is not a live
bug — but the caller depends on a contract the callee denies, and a future edit to
`sw64kb_level_copy` or the volume copiers would silently produce previous-frame pixels. Either widen
the header's stated domain with the per-copier reasoning, or record the derivation at the call site.

**N5 — the poison arm's teeth are demonstrated nowhere.** You disclose honestly that flipping the
sampled site's `zero_fill` leaves `game_compute_exec_scratch_poison` green. I checked the fixtures do
reach tiled storage shapes (`test_game_compute.cpp:3972` is a tiled 2D storage pair; `:4124` is a tiled
`img_dim=2` volume at Unorm16x4, i.e. 8 bytes per texel, which reaches the seed site's
`detile_volume` branch and the writeback's `tile_volume`), so leases *are* taken and the arm is not
vacuous. But note what a green arm can mean: given a correct `zero_fill` at every
site, poison can never survive — so the arm is a check on the derivation, not independent evidence,
and its power at the storage-seed and BCn sites is untested. Run the same flip at those two and
record whichever way it goes. If none reddens, the arm's status is "no evidence", not "verified",
and that belongs next to the claim. (This is the trap-122 shape: the control and the null share a
source.)

**N6 — the BLOG entry is a report in a blog entry's clothes.** 164 words and 13 numeric figures
against the charter's "one sentence saying what now works… more than one or two numbers and it is a
report". The *decision* to write it with no picture is right and I would keep it; the
`setup_ms`/247 GiB paragraph belongs in the issue it already links.

**N7 — merge-order hazard with #3320.** It touches the same five files (`live_compute.cpp`,
`live_renderer.cpp`, `test_game_compute.cpp`, `CMakeLists.txt`, `STRAY_STATUS.md`). Neither adds a
numbered-table row, so the trap-number gate will not catch a collision; whoever merges second should
re-run `check_merge_result.py` **and re-read the merged `STRAY_STATUS.md`** — two lanes appending
sections to one status doc is the clean-merge/broken-file shape.

**N8 — retained-memory note.** The pool never shrinks and holds up to 4 buffers / 512 MiB *per
thread*; compute now contributes 63 MiB-class extents to that set. The body's "at most one lease at
a time" is correct (I checked the scopes), but the AGENTS.md note is worth a clause if compute can
run on a thread the renderer does not share.

---

## Cleared

- **All four `zero_fill` derivations**, per branch, above. This is the finding I would have expected
  to sink the PR and it does not.
- **No frame-rate claim has crept back.** Body and `BLOG.md` both present the A/B as a null; the only
  "faster" in the body is about `tile_surface` vs `memcpy`.
- **The mutation arms bracket the change.** I re-derived the expected exit codes from
  `test_decode_scratch.cpp` rather than trusting the table: with a private non-retaining pool the
  default and poisoned arms fail on `retained_buffers() > retained_before` / `get() == first` while
  `decode_scratch_no_retention` still passes → 2 failures; with `zero_fill` ignored the default and
  poisoned arms fail on `all_zero` while the no-retention arm passes (a fresh `resize` zeroes) → 2
  failures. Both give ctest 8. The `zero_fill=false` counter-assertion (`get()[0] == 0xA5`) is what
  keeps the `zero_fill=true` assertion from being vacuous, and it is present and gated correctly.
- **`[render-timing]` format/argument correspondence**: 25 conversions, 25 arguments, types match.
- **The `try_lock`-then-`lock` change** is semantics-preserving; a spurious `try_lock` failure costs
  one over-count and nothing else.
- **"the writeback's dominant path takes `tile_mapped_bytes`, which allocates nothing"** — confirmed
  at `:9927-9930` and `:10085`.
- **Gates, re-run on `5a7f7f1f0` in a detached worktree** (not the shared checkout):
  `check_merge_result.py` → clean merge into `origin/main` `df1bbae32`, numbered column 1..256,
  256 rows, no base row deleted, rc 0; `check_numbered_table.py` over all `*.md` → rc 0;
  `check_diag_gates.py` → rc 0 (108 findings, 108 baselined); `check_ascii_output.py` → rc 0;
  `check_env_numeric_arms.py` → rc 0 (31 knobs, 31 armed); `git diff --check` → rc 0.
  `origin/main` is still `df1bbae32`, i.e. unchanged since your verification — but #3320 is open
  against the same files, so re-run `check_merge_result.py` immediately before merging anyway.

I did not rebuild or re-run ctest; the GPU is held and the author's 360/360 is not a claim I doubt.

**REJECTED** on B1 and B2 only. Both are edits to `STRAY_STATUS.md` of a couple of sentences each;
post the corrected head and I will re-read those two rows and approve.

mattias800 (commented):
# Re-review of #3317 @ `79bb0dfe6` — **APPROVED**

B1 and B2 are both addressed. The diff from the reviewed head is `STRAY_STATUS.md` only, 31 insertions
/ 4 deletions — I checked, so the derivation I did against `5a7f7f1f0` still stands unmodified, and I
have separately confirmed it survives the merge with the moved `main` (below).

Everything that follows is non-blocking. Merge when the rebase is done.

---

## Q1: yes, mildly over-corrected — in two specific places, and one of them is my arithmetic being read harder than it deserves

The core of B2's fix is right and I would not touch it: the mean genuinely cannot distinguish the two
populations, the counter-population is the right way to show it, and demoting the ~60x from measured
to unquotable is correct.

**Over-correction 1 — "If that saving were real the A/B could not have come back null" asserts a
contradiction the data cannot support.** That sentence is mine and I stated it too strongly; you have
adopted it verbatim and it is now load-bearing for "something in the chain is wrong". Two independent
reasons it does not hold:

- **The two figures are not commensurable.** The ~27 s is a whole-run number built from a *census*
  run (`evaluated=4879`, taken on `fc21d46ca`) and a *phase-timing* run (62,556 dispatches). The A/B
  is a 5.02 s F8 window at t=175 s. You cannot difference a whole-run second-count against a
  five-second window without knowing the run length and how the large seeds distribute across it.
- **The A/B has no power to see an effect that size anyway.** `main` alone spans 8.76–9.75 — 11% — at
  n=2, against n=3 on the branch. Your own later comment on #3307 establishes the rule and
  demonstrates it: a three-arm result read as "+22% with separated ranges" until the fourth arm
  landed, and you concluded "on this route, treat n<6 as unpublishable". A five-arm A/B cannot exclude
  an ~11% effect.

So the null is fully consistent with the predicted saving being **real and invisible**. That is a
*fourth* candidate the row does not list, and on the numbers it is the most likely one. The supported
sentence is "the A/B does not have the resolving power to test this prediction" rather than "something
in the chain is wrong" — which is also more useful, because it says the A/B is the wrong instrument
and the histogram is the right one.

**Over-correction 2 — the row no longer separates the established step from the unmeasured
population.** This is the "weaker but still positive claim" you asked about, and I think dropping it
is the actual cost of the rewrite. What is *established*, not candidate:

- The benchmark table is a direct measurement of the exact allocation shape (0.185 ms at 16 MiB,
  5.958 at 33, 11.279 at 64).
- glibc's dynamic `M_MMAP_THRESHOLD` capped at 32 MiB is a documented allocator property, not an
  inference.

Together those make it a **finding** that the pooling's value is a step function of lease size —
~nil below 32 MiB, large above. The only unmeasured term is where this route's leases fall relative
to that step. Stated that way you keep a real positive result and lose only the part that was never
supported. As written, a cold reader can come away doubting the benchmark, which is not in doubt.

Both are one sentence each. Neither blocks.

---

## Q2: it reconciles — a cold reader gets both sides and a resolution, not two assertions

I read the new paragraph as someone arriving with no context, and it does the job: it names the other
claim, dates it, quotes its numbers, credits them as real, and then gives two mechanisms **in the
code** for not adopting it, with `file:line`. That is the difference between reconciling and
restating. I re-checked both citations on this head — `guest_write_watch.cpp:592` is
`g_protect_calls{0}` and `:749` is `if (prior_read_only == wanted_read_only) continue;` — so the two
reasons are verifiable by a reader who wants to check you, which is what makes it read as settled
rather than asserted. Taking neither side wholesale was the right call.

Three refinements, all small:

**(a) Say that `lock_contended` did not exist in that run.** The 08:01 line carries the *old*
`mprotect=` spelling, so it predates `5a7f7f1f0`. Candidate 2 is therefore not merely unmeasured — it
is **un-measurable in the very log a reader is most likely to re-aggregate**. One clause saves someone
a wasted pass over that file.

**(b) The row says what does not isolate the span; it could say what would, in two lines.**
`host_write_pages_hit` is bumped at `guest_write_watch.cpp:1956`, immediately before
`set_pages_armed(w, hit, false)` at `:1957`. Differencing `protect_calls` across that single call —
or bumping a counter from that call site — isolates the span exactly, and it is the discriminator all
three of us have now spent time reasoning about instead of measuring. Offering it, not requiring it.

**(c) Formatting.** The inserted paragraph swallows the sentence that introduces the list: "…`##
Ruled out` exists to prevent. Three candidates remain live and the counters that separate them now
exist on…". The numbered list now hangs off a paragraph about record-keeping. A paragraph break before
"Three candidates" restores it.

---

## The rebase: mechanical, and the pooling code survives it — verified, not assumed

`main` moved from `df1bbae32` to `19566342a` (#3320, #3322, #3323, #3326). I trial-merged
`origin/main` into `79bb0dfe6` in a throwaway detached worktree:

- **Three conflicts, all "keep both"**: `BLOG.md` (two entries at the same anchor),
  `GAME_COMPAT_ORCHESTRATION.md` (your trap 256 against #3323's 257), `STRAY_STATUS.md` (your three
  `layout_ms`/`watch_ms` rows against #3320's borrow-admission withdrawal, at the same `## Ruled out`
  anchor). **No semantic collision** — #3320's row is about the borrow's decline branches and does not
  touch anything you assert; and your new section's "if `import_live_compute_storage_image` hit…"
  makes no claim about *why* it misses, so it does not contradict the withdrawal.
- **Trap numbering resolves cleanly by accident of timing**: `main` carries 255 and 257 with a gap at
  256, and you supply 256. I ran the gate on the keep-both resolution: `1..257, 257 rows, unique and
  ascending, no unused numbers`, rc 0. Your row closes the gap rather than colliding.
- **`live_compute.cpp` and `live_renderer.cpp` auto-merge cleanly — which is the half that needed
  checking, not the conflicts.** #3320 edited the same function, so I located every hunk `main`
  contributes to `live_compute.cpp` (11 of them) and checked them against the four fill chains. **None
  falls inside `7667–7810`, `8097–8150`, `8153–8250` or `9925–10120`.** The nearest,
  `@@ -10364`, adds the borrow-publish census inside the `writeback_publish` block, after
  `layout_source` is consumed, and references none of `linear`, `packed` or `layout_source`. All four
  `ScratchBuffer` sites and all four `reset()` calls are present and textually unchanged in the merge
  result. **My coverage derivation holds on the merged tree.**

So the rebase is a mechanical three-file resolution with nothing to re-derive. **Re-run
`check_merge_result.py` on the rebased head before merging** — my run was against a local trial merge,
not against the branch you will actually land, and `main` is moving fast today.

---

## Standing non-blocking items from the first review

N1 (`page_protect=` still ends in the retired `protect=`), N2 (`page_protect=0` is a false zero on
Windows — `VirtualProtect` at `guest_write_watch.cpp:195` is uncounted while the other three new
fields are honest zeros there), N3 (`one_branch_fills_the_extent` is a tautology inside the storage
branch), N4 (the predicate is used outside the domain `tile.hpp:112` declares), N5 (the poison arm's
teeth are demonstrated nowhere), N6 (BLOG entry length), N8 (per-thread retention note) all still
apply and all remain non-blocking. N7 has resolved itself into the real conflict handled above.

**APPROVED** on `79bb0dfe6`, and on a rebase of it that is limited to the three "keep both"
resolutions described. If the rebase turns out to need anything beyond that, re-request review.

PR #3320, opened 2026-09-04
perf(compute): partition why the compute->graphics image borrow declines (#3307)

mattias800:
## What this is

Instrumentation, not a fix. It answers one question that has been guessed at rather than measured:
**when Stray's title screen asks graphics to borrow the device image a compute dispatch just
produced, which of the six declines actually fires?**

Closes nothing. It selects the fix for #3307, and the fixes it selects between have nothing to do
with one another.

## The failure scenario

Stray's title screen renders at **9.25 fps** (matched-arm mean, #3307) with `gpu-device` at ~4% of
the window — the frame is entirely host-bound. Three volume costs dominate, all on the same
surfaces, all per-frame: 545 GiB of `exact` comparison, 419 GiB of page protection over 18,210
`mprotect` calls, and 247 GiB of RMW image snapshots. That is 536 MiB per rendered frame.

Every "make the copying cheaper" lever has now been measured and is null (#3317's pooling A/B; an
8 GiB image cache; the journal-deferral path, which answers only 11.3% of validations). What
remains is not doing the copying: a 3840×2160 RGBA16F intermediate is re-tiled out of a compute
dispatch into guest memory and immediately read back, detiled and converted by the graphics texture
materializer. Both Vulkan users are on the same device. The image should never touch guest memory.

`import_live_compute_storage_image` exists to make exactly that borrow, the renderer's witness says
the surface passes every gate to attempt it (`compute_cand=1`), and it misses on every reference.

## Why a census had to come first

The importer plus `borrow_cached_image_for_graphics` decline through **six** independent branches,
and before this change every one of them was the same bare `return false`:

1. the consumer's precondition (shape, class, host-backed data, mip levels, srgb, format);
2. no cache entry under the key;
3. an entry whose export authority was never published, or was invalidated after publication;
4. an entry with no `VkImage`;
5. a published entry that no proof says still matches guest memory;
6. a lease the renderer accepts and then rejects on format/extent/device.

`STRAY_STATUS.md` asserted (2) — *"finds nothing under its key … it lives in the compute cache's
admission"* — as a statement of fact. It is an inference, and no run could have established it. This
PR **withdraws it as unmeasured rather than falsifying it**, and records why in that doc's
`## Ruled out`: a fix aimed at admission would report "no change" for a reason unrelated to the
change if the miss is really (3) or (5).

## The contract

`PROSPER_COMPUTE_BORROW_CENSUS=1` prints four `[compute-borrow-census]` lines every 256 submits and
at exit. Every line is a running total carrying its own denominator, so two reports taken at
different points in a run stay comparable — the failure that produced #3155's retracted numbers.

* **the consumer's precondition**, partitioned by term (`shape`, `srgb`, `mip_levels`, …);
* **the borrow's outcome** — `no_cache_entry`, `export_unpublished`, `content_invalid`, `no_image`,
  `authority_changed`, `hit` — with a **zero bucket still printed**. A bucket that vanishes when
  empty is indistinguishable from an instrument that never reached it, which cost a correct
  finding a whole night on this very issue (a healthy zero on `[render-timing]`'s mprotect FAILURE
  counter, read as the activity counter sharing its prefix further along the same line);
* **why `authority_changed` fired**, splitting the `Unknown` that `guest_gpu_writes_since` returns
  for two structurally different reasons its own header says a caller cannot tell apart: an unarmed
  per-thread submit journal (`journal_unarmed`) versus an export published in an earlier submit
  (`journal_cross_submit`). Plus what the page watch answered;
* **the producer's publish gate**, partitioned the same way, so "the consumer's key differs" and
  "the producer never published anything" stop being the same observation.

On a lookup miss the same variable enables an **O(cache) scan for an entry at the same guest
address**, reporting which key field the nearest such entry disagreed on. The key carries
twenty-three fields; without that scan, "the producer never ran" and "the producer is keyed
differently" produce identical evidence.

Counting is unconditional — a handful of relaxed atomic adds per import, against a path whose
alternative is a multi-megabyte detile. Only the report and the scan are gated.

Sample output, from `test_game_compute`'s own run (real device, real ordered submits):

```
[compute-borrow-census] running totals -- imports=32 host_data=1 shape=1 mip_levels=1 srgb=1 accepted=28 (87.5%)
[compute-borrow-census] running totals -- attempted=28 no_cache_entry=8 (28.6%) export_unpublished=0 (0.0%) content_invalid=0 (0.0%) no_image=0 (0.0%) authority_changed=2 (7.1%) hit=18 (64.3%) alias_retry=17 lease_fail=0 spared=0.6 MiB decoded=0.2 MiB
[compute-borrow-census] running totals -- authority_changed=2 journal_overlap=0 journal_unarmed=2 journal_cross_submit=0 watch_absent=2 watch_dirty=0 watch_unknown=0 | no_entry_scans=8 same_addr=8 guest_bytes=1 resource_bytes=1 width=3 height=1 depth=1 format=7 components=1 tile_mode=2 layer_stride=1 vk_format=7 storage=1 mip_levels=1
[compute-borrow-census] running totals -- publish_evaluated=175 authorized=71 (40.6%) not_native_exact=84 (48.0%) not_unique=0 (0.0%) not_cache_candidate=20 (11.4%) not_persistent=0 (0.0%) no_sampled_usage=0 (0.0%) export_refused=0 (0.0%) | renderer_accepted=0 renderer_rejected=0
```

## Behavioural change, and how it is fenced

Two, both small, both required to make a reason nameable:

* the importer's precondition `||` chain becomes a `constexpr` classifier. **The acceptance set must
  not move**, and `test_compute_image_borrow_census` asserts it over the *complete boolean product*
  of the eleven terms — 2048 cases against the chain transcribed independently in the test — rather
  than over cases chosen by whoever did the rewrite. Both counts are pinned (`accepted == 1`,
  `declined == 2047`) so an accept-only or decline-only sweep cannot pass vacuously;
* `authorize_cached_image_export` returns whether it published. A publish-eligible binding whose
  entry was evicted between writeback and the publish loop is a different failure from an
  ineligible one, and `ExportRefused` is the bucket that separates them.

`native_storage_vk_format` is hoisted above the precondition so the classifier sees every term at
once. It is a pure switch over the guest format enum and is safe with a null context.

Nothing else changes what any gate decides. The borrow evaluates exactly the predicates it evaluated
before, in the same order and behind the same short circuits — in particular `write_watch.query()`
still runs only when the journal did not already answer, because a census that pays for a query the
borrow skips would be measuring its own cost.

## Verification

Configured with `-DPROSPER_APP=ON`, in the `ps5ys` container.

```
cmake --build prosper/build-linux -j4        # rc=0, zero warnings
ctest --no-tests=error -j4                   # rc=0, 100% tests passed out of 359
```

359 is one more than `origin/main`'s 358: the new `compute_image_borrow_census` case.

`test_game_compute` (real Vulkan device, real producer dispatch, real ordered submit) now asserts
the census on the production path, bracketing one hit and one deliberately mis-keyed miss taken back
to back inside the same consumer callback — including that the near-miss scan names `tile_mode` and
**does not** name a field that matched, which is the arm a mask that set every bit would otherwise
satisfy.

### Mutation arms, by exit code

Each applied to a clean tree, built, run, then reverted. The last row is the restored tree.

| mutation | test | exit |
| --- | --- | --- |
| drop the `srgb` term from the import classifier | `compute_image_borrow_census` | **1** |
| fold non-`AuthorityChanged` misses into the authority breakdown | `compute_image_borrow_census` | **1** |
| omit zero-valued outcome buckets from the report | `compute_image_borrow_census` | **1** |
| near-miss scan reports every key field as differing | `game_compute` | **1** |
| the borrow reports a successful lease as a cache miss | `game_compute` | **1** |
| restored | both | **0** |

## What is deliberately unaffected

No gate's decision changes, so no title's rendering changes and no snapshot is expected to move. The
census is off by default in the sense that matters — no output, no scan — and its counters have no
consumer other than the report and the tests.

## The run this needs, and the decision rule stated in advance

One `boot_trace` or `tools/screenshot` run of `scripts/stray-PPSA02101/reach-title-hold.pad` with
`PROSPER_NULL_PAGE=1` and `PROSPER_COMPUTE_BORROW_CENSUS=1`, held at the title screen. The last
`[compute-borrow-census]` block is the whole result. Recording the rule now so the answer cannot be
read to suit a preferred fix:

| what the census says | what it means | the fix that follows |
| --- | --- | --- |
| `no_cache_entry` dominant, `same_addr=0` | the producer retained nothing at this address | read the `publish_*` line; the lever is admission, as the doc guessed |
| `no_cache_entry` dominant, `same_addr>0` + a named field | the two descriptors are keyed differently | normalise that field, justified per field — never a blanket relaxation. The field list is the EXACT key's (`exact_key_fields:`), read against `same_addr=`, which is the field list's exact denominator — not against `exact_key_scans`, its `rescued=` remainder, or `no_cache_entry` |
| `export_unpublished` / `content_invalid` dominant | the entry exists and its authority was cleared, most likely by the next dispatch's source re-validation | the producer→consumer handoff must survive a re-seed |
| `authority_changed` + `journal_overlap` | a real write covers the range between publish and use | find the writer (`PROSPER_GUEST_WRITE_WATCH`); it may be prosper's own writeback |
| `authority_changed` + `journal_cross_submit` | producer and consumer never meet inside one submit | cross-submit export validity; `watch_absent` then says why the page watch cannot cover it |
| `authority_changed` + `journal_unarmed` | the consumer runs outside a submit scope | the borrow is structurally unreachable from the materializer |
| `publish_not_native_exact` dominant | the producer is not a typed-storage shader | the borrow cannot apply to this surface at all; a different lever entirely |
| `renderer_rejected > 0` | the borrow hits and graphics discards it | format/extent agreement at the renderer |

Refs #3307, #3126, #2883.

---

## Review round 2 — head `96d27f18`

The independent review of `3a88d671` returned **REJECTED** with one blocking finding and five
non-blocking ones. All are addressed, plus one defect I found on my own re-read.

**Blocking, and it defeated the instrument on exactly the title it was built for.** The importer
tries the exact descriptor key, then retries under a format-*aliased* key. The near-miss scan was
recorded from the FINAL observation, so on every miss that reached the retry the reported field list
was the retry's — and the retry rewrites `format` and `vk_format` by construction, so it named those
two as differing whether or not they were the reason. Stray's surface is RGBA16F, which satisfies
`float16_uint16_alias`, so **every** near-miss took that path. The decision table above then said
"a named field → normalise that field", pointing at precisely the two fields the alias arm's own
comment forbids normalising.

Fixed: the exact key's scan is recorded where it is taken, the retry asks for no scan
(`scan_near_miss` is now an explicit parameter), the counter is renamed `exact_key_scans` because it
now counts scans *performed* rather than misses observed, and the report labels the list
`exact_key_fields:`. The decision-table row above is corrected to say which key it describes.

**The other six:**

* hardcoded `GuestGpuWriteQuery` / `GuestWriteWatchQuery` ordinals are now pinned by `static_assert`
  in `live_compute.cpp`, where both enums are visible;
* `journal_cross_submit` covered more than its name said (my own finding, and the same shape as the
  `watch_ms` mislabel #3307 recorded yesterday) — split into `journal_unjournaled` and
  `journal_undecided`, the second deliberately vague because it covers two causes this instrument
  cannot separate;
* the 2 KiB report buffer was thinner than its own comment implied — 4 KiB, sized by a **saturation
  arm** rather than an estimate, and truncation is now announced in the buffer;
* the ten `gates.* = <descriptor field>` assignments had no coverage (the 2048-case sweep proves
  classifier == chain over the abstract booleans; a lost `!` would live in the mapping) — eleven
  cases in `test_game_compute` now drive them from the real path, one field at a time, and the
  census output shows all eleven decline buckets firing;
* relaxed-atomic tearing and the scan's lock discipline are documented where the next reader will be;
* the reviewer also flagged a `static_assert(... Overlap) == 2` in the working tree — that was
  mutation arm M7 caught mid-flight; `Overlap` is 1 in the committed code, and the arm exists to
  prove the assert is live.

Re-verified on `96d27f18`: build rc=0 zero warnings; `ctest --no-tests=error -j4` rc=0, 100% of 359;
`check_trap_citations.py`, `check_numbered_table.py`, `check_diag_gates.py` all rc=0.

| new mutation arm | test | exit |
| --- | --- | --- |
| record the ALIAS retry's near-miss mask (the reviewed defect, restored) | `game_compute` | **1** |
| drop the truncation marker | `compute_image_borrow_census` | **1** |
| the import gate wiring loses a negation (`gates.srgb = false`) | `game_compute` | **1** |
| collapse `journal_unjournaled` into `journal_undecided` | `compute_image_borrow_census` | **1** |
| assert `GuestGpuWriteQuery::Overlap == 2` when it is 1 | build | **2** |
| restored | all | **0** |

The first reproduces the reviewed defect exactly and the new arm names it:
`FAIL: the near-miss mask is the exact key's, not the format-alias retry's`.

---

## Review round 3 — head `0fb55422` — APPROVED

The re-review of `96d27f18` confirmed the blocking fix (*"the new arm pins it with exactly the
Format/VkFormat pair that was missing; under the old code both would have incremented, so it goes red
on the defect rather than passing for an unrelated reason"*) and returned **APPROVED**, with five
non-blocking findings — **two of them found by compiling the census header standalone and measuring,
rather than by reading it**. Both were real defects in the instrument. All five are fixed here.

**N3 — the rescued scan floods the field list.** An exact-key miss that the alias retry then
*rescues* is a successful import, and its exact-key scan still ran with a mask naming `format` and
`vk_format` — that being the exact difference the retry exists to bridge. On a title whose alias path
routinely succeeds (GTA V's integer-storage surfaces) the field list would again be dominated by the
alias working as designed: the blocking confound arriving from the other direction. The scan is now
recorded after the retry's fate is known; a rescued scan is counted
(`exact_key_scans=N (rescued=M, not counted below)`) and contributes no fields. Measured on
`test_game_compute`'s own run: `format` 9 → 3, `vk_format` 11 → 3.

**N2 — truncation was inferred, not observed.** `used == capacity - 1` holds both when the report was
cut short and when it fit exactly, so a report of natural length `capacity - 1` had its tail
overwritten by a TRUNCATED marker announcing a loss that never happened — on the producer partition,
the line the marker exists to protect. Demonstrated by the reviewer at length 838 into capacity 839.
`emit` now sets a flag when `snprintf` reports more than its room.

**N1 — the saturation arm could not express its own worst case.** All-max moves numerator and
denominator together, so every percentage collapses to `100.0` and the widest percentage is
structurally inexpressible. Skewing the three denominators to 1 is wider, and the arm now *asserts*
that (`widest_used > all_max`) instead of assuming its construction was extremal. The 4 KiB
conclusion survives; the derivation did not. Same-source-positive-control shape.

**N4/N5** — the wiring array had no positive control (the classifier returns the *first* failing
term, so a base that already declined would let every case pass while testing nothing), and a comment
said "nine" for eleven cases covering ten buckets while claiming all eleven fire. Both corrected;
`NoContext` is structurally undrivable there and the comment now says so.

Verified on `0fb55422`: build rc=0 zero warnings; `ctest --no-tests=error -j4` rc=0, 100% of 359;
`check_trap_citations.py`, `check_diag_gates.py`, `check_numbered_table.py` all rc=0. Two more
mutation arms — accumulate the rescued scan's fields, and infer truncation from a full buffer — each
restore the reviewed defect and each goes red (exit 1); restored 0.

CI on the previous head `96d27f18` finished green (8 success, 1 skipped, macOS x86_64 still running
at the time of writing).

---

## Review round 4 — head `d9bd1f4f`

Round 3 returned **APPROVED** on `0fb55422` and confirmed the N3 split introduces no new ambiguity
(*"a clean three-level nesting, each a proper subset of the one above … that was the specific
inconsistency I went looking for and did not find"*), with one documentation-only refinement. It
needed a test as well.

**The denominator was named one level too wide.** Two places said the field list is read against
`exact_key_scans` minus `rescued`. It is not: a field bit and `no_entry_same_addr` are incremented
under one and the same condition, so `no_entry_same_addr` is the exact denominator. The wider number
folds in scans that found *nothing* at the address — a different row of the decision table, the one
where the producer is absent rather than keyed differently. The reviewer's worked case, rendered from
this formatter: `exact_key_scans=12 (rescued=9) same_addr=2 tile_mode=2` is `tile_mode` on 100% of
near-misses, not 67%. Corrected in the header, in `FRONTEND_APP.md`, and in the decision table above,
with the nesting spelled out:

```
exact_key_scans  >=  (exact_key_scans - rescued)  >=  no_entry_same_addr  >=  key_field_diffs[f]
```

**And the arm written to pin it was vacuous.** Its inputs gave the "nothing at this address" scan a
*zero* mask, so "fields are gated on `same_addr`" and "fields happen to be empty" were the same
observation — a mutation dropping the gate exited **0**. The gate is a contract of the public
recorder, so it is now tested at the boundary the contract names rather than the one the current
caller happens to produce: a scan with no same-address entry and a non-empty mask, asserting its
fields are not counted. The mutation now exits **1** on both field arms, and the inputs are chosen so
every inequality in the chain is strict (4 > 3 > 1 ≥ 1) — an arm whose levels were equal would hold
for an implementation that collapsed them.

That is the third assertion in this PR that turned out not to discriminate, and the second found by
someone other than its author. It is recorded in the test's own comment, not only here.

Verified on `d9bd1f4f`: build rc=0 zero warnings; `ctest --no-tests=error -j4` rc=0, 100% of 359;
`check_trap_citations.py`, `check_diag_gates.py`, `check_numbered_table.py` all rc=0.

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -1309,6 +1309,14 @@ if(TARGET prosper_core)
   target_include_directories(test_decode_scratch PRIVATE frontends src)
   add_test(NAME decode_scratch COMMAND test_decode_scratch)
 
+  # #3307 compute->graphics image borrow census. Header-only and device-free: the load-bearing arm
+  # is the 2048-case equivalence between the importer's precondition classifier and the `||` chain
+  # it replaced, which needs no Vulkan and must never be skipped on a host without it.
+  add_executable(test_compute_image_borrow_census
+                 frontends/shared/tests/test_compute_image_borrow_census.cpp)
+  target_include_directories(test_compute_image_borrow_census PRIVATE frontends)
+  add_test(NAME compute_image_borrow_census COMMAND test_compute_image_borrow_census)
+
   add_executable(test_texture_decode_diagnostic
                  frontends/shared/tests/test_texture_decode_diagnostic.cpp)
   target_include_directories(test_texture_decode_diagnostic PRIVATE frontends src)
--- a/prosper/docs/FRONTEND_APP.md
+++ b/prosper/docs/FRONTEND_APP.md
@@ -480,6 +480,22 @@ two arms' tallies taken at different points in a run is the mistake that produce
 `PROSPER_WRITE_WATCH_MAX_KB` is an emergency host-wide range
 limit (zero/unset is unbounded); a declined watch always returns to exact comparison.
 
+`PROSPER_COMPUTE_BORROW_CENSUS=1` reports the other side of the same boundary, on the same cadence:
+whether a graphics sampled descriptor managed to lease the device image a compute dispatch had just
+produced, instead of re-reading the surface out of guest memory. Four `[compute-borrow-census]`
+lines, all running totals with their own denominators — the consumer's precondition partitioned by
+term, the borrow's outcome partitioned by gate (`no_cache_entry`, `export_unpublished`,
+`content_invalid`, `no_image`, `authority_changed`, `hit`), why `authority_changed` fired (a real
+overlapping guest write, an unarmed submit journal, or a cross-submit export; and what the page
+watch said), and the producer's publish gate partitioned the same way. A zero bucket still prints.
+
+Two of its fields answer questions nothing else can. On a lookup miss the variable additionally
+enables an O(cache) scan for an entry at the same guest address, which turns "nothing is cached
+under this key" into "an entry here disagrees on `tile_mode`" — the twenty-three-field cache key
+otherwise makes those indistinguishable. And `journal_unarmed` versus `journal_cross_submit` splits
+the `Unknown` that `guest_gpu_writes_since` returns for two structurally different reasons, which
+its own header warns cannot be told apart by a caller.
+
 Proven-full write-only storage targets at least 16 MiB do not copy their first successful result into
 an immediately redundant CPU source baseline. The submit journal remains the initial source authority;
 if another architectural writer invalidates it, the next invocation takes ordinary writeback and
--- a/prosper/docs/FRONTEND_APP.md
+++ b/prosper/docs/FRONTEND_APP.md
@@ -497,8 +497,11 @@ exact key misses, the importer retries under a format-*aliased* key, and that ke
 and `vk_format` by construction — so scanning it would report those two as differing on every miss
 whether or not they were the reason, and would point a reader at exactly the normalisation the alias
 arm forbids. The retry is therefore not scanned, which is also why `exact_key_scans` (scans
-performed) and `no_cache_entry` (final outcomes) are different numbers: read the field list against
-the former. And the `Unknown` that `guest_gpu_writes_since` returns —
+performed) and `no_cache_entry` (final outcomes) are different numbers. Read the field list against
+`exact_key_scans` **minus** its `rescued=` term: an exact-key miss the alias retry then rescues is a
+successful import whose mask names `format`/`vk_format` by construction, so those scans are counted
+and their fields deliberately excluded — otherwise a title whose alias path routinely succeeds
+reports a field list dominated by the alias working as designed. And the `Unknown` that `guest_gpu_writes_since` returns —
 which its own header warns a caller cannot decompose — is split into the three parts that ARE
 separable from the borrow site: `journal_unarmed` (not armed on the consuming thread),
 `journal_unjournaled` (the producer's publish carried no submit serial, so it was stamped while the
--- a/prosper/docs/FRONTEND_APP.md
+++ b/prosper/docs/FRONTEND_APP.md
@@ -498,10 +498,16 @@ and `vk_format` by construction — so scanning it would report those two as dif
 whether or not they were the reason, and would point a reader at exactly the normalisation the alias
 arm forbids. The retry is therefore not scanned, which is also why `exact_key_scans` (scans
 performed) and `no_cache_entry` (final outcomes) are different numbers. Read the field list against
-`exact_key_scans` **minus** its `rescued=` term: an exact-key miss the alias retry then rescues is a
-successful import whose mask names `format`/`vk_format` by construction, so those scans are counted
-and their fields deliberately excluded — otherwise a title whose alias path routinely succeeds
-reports a field list dominated by the alias working as designed. And the `Unknown` that `guest_gpu_writes_since` returns —
+**`same_addr=`, and against nothing else** — a field bit and `same_addr` are incremented under one
+and the same condition, so that is the list's exact denominator. The three numbers on that line nest
+(`exact_key_scans` ⊇ non-`rescued` ⊇ `same_addr`), and dividing by either wider one dilutes the field
+list with lookups that found nothing at the address at all — which is the *other* decision-table row,
+the one where the producer is absent rather than keyed differently.
+
+`rescued=` counts the exact-key misses the format-alias retry then saved. Those are successful
+imports whose mask names `format`/`vk_format` by construction, so they are counted and their fields
+deliberately excluded — otherwise a title whose alias path routinely succeeds reports a field list
+dominated by the alias working as designed. And the `Unknown` that `guest_gpu_writes_since` returns —
 which its own header warns a caller cannot decompose — is split into the three parts that ARE
 separable from the borrow site: `journal_unarmed` (not armed on the consuming thread),
 `journal_unjournaled` (the producer's publish carried no submit serial, so it was stamped while the
--- a/prosper/docs/FRONTEND_APP.md
+++ b/prosper/docs/FRONTEND_APP.md
@@ -492,9 +492,20 @@ watch said), and the producer's publish gate partitioned the same way. A zero bu
 Two of its fields answer questions nothing else can. On a lookup miss the variable additionally
 enables an O(cache) scan for an entry at the same guest address, which turns "nothing is cached
 under this key" into "an entry here disagrees on `tile_mode`" — the twenty-three-field cache key
-otherwise makes those indistinguishable. And `journal_unarmed` versus `journal_cross_submit` splits
-the `Unknown` that `guest_gpu_writes_since` returns for two structurally different reasons, which
-its own header warns cannot be told apart by a caller.
+otherwise makes those indistinguishable. **That field list belongs to the EXACT key only.** When the
+exact key misses, the importer retries under a format-*aliased* key, and that key rewrites `format`
+and `vk_format` by construction — so scanning it would report those two as differing on every miss
+whether or not they were the reason, and would point a reader at exactly the normalisation the alias
+arm forbids. The retry is therefore not scanned, which is also why `exact_key_scans` (scans
+performed) and `no_cache_entry` (final outcomes) are different numbers: read the field list against
+the former. And the `Unknown` that `guest_gpu_writes_since` returns —
+which its own header warns a caller cannot decompose — is split into the three parts that ARE
+separable from the borrow site: `journal_unarmed` (not armed on the consuming thread),
+`journal_unjournaled` (the producer's publish carried no submit serial, so it was stamped while the
+journal was inactive or already overflowed), and `journal_undecided` (armed, serial present — a
+different submit, or this submit's journal has since overflowed). The last name is deliberately
+vague, because it covers two causes this instrument cannot tell apart and a narrower name would be
+a claim rather than a count.
 
 Proven-full write-only storage targets at least 16 MiB do not copy their first successful result into
 an immediately redundant CPU source baseline. The submit journal remains the initial source authority;
--- a/prosper/docs/STRAY_STATUS.md
+++ b/prosper/docs/STRAY_STATUS.md
@@ -176,13 +176,26 @@ dispatch count, both of which the report already prints per program.
 ### The lever that is worth more than all of this
 
 The witness says `compute_cand=1`: the surface is a **compute image candidate whose import misses
-every frame**. `import_live_compute_storage_image` finds nothing under its key, so graphics falls
-through to the guest-byte decode. If that import hit, the 63.8 MiB writeback → detile → convert
-disappears **on both sides of the boundary** — the graphics decode and the compute writeback that
-feeds it. That is a bigger lever than every allocation above put together, and it lives in the
-compute cache's admission, not in the materializer. Both halves of that round trip are read
-together in [#3307](https://github.com/mattias800/prosper/issues/3307): the compute side's re-tile
-into guest memory, and the graphics side's read of those same bytes back out.
+every frame**, so graphics falls through to the guest-byte decode. If that import hit, the 63.8 MiB
+writeback → detile → convert disappears **on both sides of the boundary** — the graphics decode and
+the compute writeback that feeds it. That is a bigger lever than every allocation above put
+together. Both halves of that round trip are read together in
+[#3307](https://github.com/mattias800/prosper/issues/3307): the compute side's re-tile into guest
+memory, and the graphics side's read of those same bytes back out.
+
+**WHY it misses is not established, and this paragraph used to say it was.** It read
+*"`import_live_compute_storage_image` finds nothing under its key"* and *"it lives in the compute
+cache's admission"*, which name one specific branch out of six — and nobody had measured which one
+fires. The importer plus `borrow_cached_image_for_graphics` decline through: the consumer's own
+precondition; no cache entry under the key; an entry whose export authority was never published or
+was invalidated; an entry with no image; a published entry that no proof says still matches guest
+memory; and a lease the renderer accepts and then rejects on format/extent/device. Every one of
+those used to be the same bare `return false`.
+
+`PROSPER_COMPUTE_BORROW_CENSUS=1` now partitions them (`[compute-borrow-census]`, four lines, every
+256 submits and at exit), including the producer's own publish gate and — on a lookup miss — a scan
+for a same-address entry that names the key field it disagreed on. **Read that before proposing a
+fix here**; the branch it reports selects between fixes that have nothing to do with one another.
 
 ## The unresolved image ops — established on CALIBRATION
 
@@ -795,6 +808,16 @@ A ~13% reduction, groups non-overlapping. The visible frame does not change — 
 drops still discard the background.
 
 ## Ruled out
+
+- **"`import_live_compute_storage_image` finds nothing under its key, so the compute-image borrow
+  fails in the cache's ADMISSION."** Not falsified — **withdrawn as unmeasured**, 2026-09-04. It was
+  written into § *The lever that is worth more than all of this* as a statement of fact, and it is
+  an inference: the importer and the borrow it calls have six independent decline branches and all
+  six returned the same bare `false`, so no run could have distinguished them. This row exists
+  because the sentence was already being quoted as the starting point for a fix, and a fix aimed at
+  the admission gate would report "no change" for a reason unrelated to the change if the miss is
+  really the export-authority or the write-proof branch. The instrument that decides it is
+  `PROSPER_COMPUTE_BORROW_CENSUS=1`. #3307.
 
 - **"A recompile fix — resolving the unresolved image ops — restores the title-screen background."**
   **Falsified on the title screen, and this row exists because the measurement was recorded NOWHERE a
--- a/prosper/frontends/shared/compute/AGENTS.md
+++ b/prosper/frontends/shared/compute/AGENTS.md
@@ -1,0 +1,35 @@
+# `frontends/shared/compute` — what the compute backend decides, separated from how it decides it
+
+Header-only, Vulkan-free, dependency-light decision logic and instrumentation for the live compute
+backend (`live/live_compute.cpp`) and for the one boundary it shares with the renderer
+(`live/live_renderer.cpp`). Nothing here touches a device, guest memory, or the write-watch
+mechanism: every file answers a question from plain integers and booleans, so it is unit-testable
+without a GPU. The tests live in `frontends/shared/tests/`.
+
+**Why the decisions live here and not beside their call sites.** `live_compute.cpp` is eleven
+thousand lines, and its gates were historically spelled as long `||` chains inside the functions
+that acted on them. A chain like that answers *whether* and never *which*, so a surface that
+declined for one of eight reasons produced exactly the same evidence as one that declined for
+another — and several nights of this project's history were spent guessing which. Lifting the gate
+into a `constexpr` classifier here makes the reason a value, which makes it countable, which makes
+"the borrow misses" into "the borrow misses because the entry's key disagrees on `tile_mode`".
+
+**The boundary against its siblings.** `shared/texture/` owns the *validation* policies both caches
+share — whether guest bytes must be re-read, how many proven-unchanged looks earn a page watch.
+This folder owns the compute backend's own eligibility and identity decisions plus the censuses that
+report them. Cache materialization, Vulkan resource creation and the guest↔device copies stay in
+`live/`, which is where the types they act on are declared.
+
+**A census here is a partition, not a log.** Every one of these files counts a *decision*: each
+observation lands in exactly one bucket, the report always prints the denominator, and a bucket that
+is zero still prints. The last part is deliberate and has been paid for — a bucket that disappears
+when empty is indistinguishable from an instrument that never reached it, which is how a correct
+finding got falsified by a zero read off the wrong field (instrument trap 256). Counting is
+unconditional wherever the counted branch is expensive enough that a relaxed atomic add is free by
+comparison; only the *report* is gated on an environment variable.
+
+**The one thing a newcomer gets wrong.** `compute_transfer_gate_census.hpp`'s counters are gated on
+a *selector* — `PROSPER_COMPUTE_TRANSFER_PRODUCER_HASH` / `..._CONSUMER_HASH` — so they observe
+nothing at all unless a specific producer/consumer program pair was named on the command line. The
+other censuses here count on every run. Reading a zero from the transfer-gate census on a default
+launch says nothing about the code it instruments.
--- a/prosper/frontends/shared/compute/AGENTS.md
+++ b/prosper/frontends/shared/compute/AGENTS.md
@@ -24,7 +24,10 @@ report them. Cache materialization, Vulkan resource creation and the guest↔dev
 observation lands in exactly one bucket, the report always prints the denominator, and a bucket that
 is zero still prints. The last part is deliberate and has been paid for — a bucket that disappears
 when empty is indistinguishable from an instrument that never reached it, which is how a correct
-finding got falsified by a zero read off the wrong field (instrument trap 256). Counting is
+finding got falsified by a zero read off the wrong field: `[render-timing]`'s pre-existing
+`protect=` counts mprotect *failures* and is healthily zero, and a truncated grep read it as the
+activity counter that sits further along the same 288-character line (#3307, and the
+instrument-trap row #3317 adds). Counting is
 unconditional wherever the counted branch is expensive enough that a relaxed atomic add is free by
 comparison; only the *report* is gated on an environment variable.
 
--- a/prosper/frontends/shared/compute/compute_image_borrow_census.hpp
+++ b/prosper/frontends/shared/compute/compute_image_borrow_census.hpp
@@ -1,0 +1,519 @@
+#pragma once
+
+// Why this file exists.
+//
+// A compute dispatch produces a storage image on the device; a graphics draw in the same submit
+// wants to sample exactly those pixels. `import_live_compute_storage_image` exists so the second
+// can borrow the first's `VkImage` instead of routing 63.8 MiB through guest memory -- compute
+// re-tiling its result out, graphics detiling it back in. On Stray's title screen that borrow
+// misses on every reference (#3307), and the round trip it fails to remove is the single largest
+// host cost in the frame.
+//
+// The importer and the borrow it calls have at least SIX independent ways to decline, and before
+// this header not one of them logged a reason:
+//
+//   1. the consumer's own precondition (shape, class, host-backed data, mip levels, srgb, format);
+//   2. no cache entry under the key;
+//   3. an entry exists but its export/content/image authority is not published;
+//   4. the entry is published but no proof says guest memory still matches it;
+//   5. the producer never authorized an export in the first place;
+//   6. the borrow succeeded and the RENDERER rejected the imported image afterwards.
+//
+// `STRAY_STATUS.md` asserted (2) from the outside. That is a guess about a five-way branch, and it
+// is the kind of guess this project has repeatedly paid for: a plausible reason quoted as a measured
+// one. So every branch is counted, unconditionally, and one of them will turn out to carry the mass.
+//
+// The counters partition. Every import lands in exactly one decline reason or exactly one borrow
+// outcome, and both arrays print the denominator they are a fraction of, so two reports taken at
+// different points in a run remain comparable -- the failure that produced #3155's retracted
+// numbers. The `## Ruled out` value of a partition is that it cannot be read selectively.
+//
+// Counting is unconditional and per-IMPORT (a handful of relaxed atomic adds against a path whose
+// alternative is a multi-megabyte detile). Only the REPORT is gated, on
+// PROSPER_COMPUTE_BORROW_CENSUS. The one genuinely non-free part -- scanning the image cache for a
+// near-miss key when the lookup found nothing -- is additionally gated on that same variable,
+// because it is O(cache) rather than O(1).
+
+#include <atomic>
+#include <cstddef>
+#include <cstdint>
+#include <cstdio>
+
+namespace prosper::frontend {
+
+// ---------------------------------------------------------------------------
+// (1) The consumer's precondition, as a pure classifier.
+//
+// The production code used to spell this as one `||` chain that returned false, so a declined
+// import was indistinguishable from any other declined import. The classifier answers WHICH term
+// declined. `classify_compute_image_import(...) == None` must remain exactly equivalent to that
+// chain's negation; `test_compute_image_borrow_census` asserts the equivalence over the complete
+// boolean product rather than over hand-picked cases, because the failure mode being guarded is a
+// term silently dropped during the rewrite.
+// ---------------------------------------------------------------------------
+
+enum class ComputeImageImportDecline : uint8_t {
+    None = 0,          // accepted: the borrow was attempted
+    NoContext,         // no live compute backend / no device
+    NoGuestAddress,
+    NotTextureClass,
+    HostBackedData,    // a capture/replay resource whose bytes are host-owned
+    GuestByteRange,    // zero, or wider than the u32 the cache key carries
+    Shape,             // neither an ordinary 2D/2D-array/3D nor the cube-array alias
+    MipLevels,         // declared_mip_levels != 1
+    MipTail,
+    Srgb,
+    DepthCompare,
+    NativeFormat,      // no exact Vulkan storage format for this guest format/component count
+    Count,
+};
+
+struct ComputeImageImportInputs {
+    bool have_context = false;
+    bool have_gpu_addr = false;
+    bool texture_class = false;
+    bool host_data = false;
+    bool guest_bytes_in_range = false;
+    bool shape_supported = false;
+    bool single_mip_level = false;
+    bool in_mip_tail = false;
+    bool srgb = false;
+    bool depth_compare = false;
+    bool native_format_defined = false;
+};
+
+constexpr ComputeImageImportDecline classify_compute_image_import(
+    const ComputeImageImportInputs& in) {
+    if (!in.have_context) return ComputeImageImportDecline::NoContext;
+    if (!in.have_gpu_addr) return ComputeImageImportDecline::NoGuestAddress;
+    if (!in.texture_class) return ComputeImageImportDecline::NotTextureClass;
+    if (in.host_data) return ComputeImageImportDecline::HostBackedData;
+    if (!in.guest_bytes_in_range) return ComputeImageImportDecline::GuestByteRange;
+    if (!in.shape_supported) return ComputeImageImportDecline::Shape;
+    if (!in.single_mip_level) return ComputeImageImportDecline::MipLevels;
+    if (in.in_mip_tail) return ComputeImageImportDecline::MipTail;
+    if (in.srgb) return ComputeImageImportDecline::Srgb;
+    if (in.depth_compare) return ComputeImageImportDecline::DepthCompare;
+    if (!in.native_format_defined) return ComputeImageImportDecline::NativeFormat;
+    return ComputeImageImportDecline::None;
+}
+
+constexpr const char* compute_image_import_decline_name(ComputeImageImportDecline reason) {
+    switch (reason) {
+    case ComputeImageImportDecline::None:            return "accepted";
+    case ComputeImageImportDecline::NoContext:       return "no_context";
+    case ComputeImageImportDecline::NoGuestAddress:  return "no_gpu_addr";
+    case ComputeImageImportDecline::NotTextureClass: return "not_texture";
+    case ComputeImageImportDecline::HostBackedData:  return "host_data";
+    case ComputeImageImportDecline::GuestByteRange:  return "guest_bytes";
+    case ComputeImageImportDecline::Shape:           return "shape";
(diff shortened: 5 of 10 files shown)


mattias800 (commented):
Independent review of head `3a88d671`, against `origin/main`. Read-only; no build, no GPU. Every `file:line` below is from the **committed head** (see the note at the end about the working tree).

## REJECTED

One blocking finding. It is not a correctness regression in the emulator — the behaviour-preservation work is genuinely careful and I could not break it — it is that the instrument's headline output is systematically confounded on exactly the title it was built to decide, while the PR body publishes a decision rule that reads that output literally.

---

## BLOCKING — the near-miss field names are the ALIAS key's, and the decision table reads them as the exact key's

`live_compute.cpp:10966` resets `observation = {}` before the alias retry, so the exact key's completed near-miss scan is discarded, and `live_compute.cpp:10971-10974` records only the **retry's** field-diff mask.

The retry key is not a small perturbation of the exact key. `live_compute.cpp:10950-10963` rewrites `storage_identity.format` (every alias arm except `packed_r11`) and always recomputes `producer_format`. So relative to the exact key, the retry key differs in `format` **and** `vk_format` by construction. Whenever a `NoCacheEntry` miss goes through the retry, the reported field set is shifted by those two fields relative to the lookup a reader believes it describes.

This is not hypothetical for Stray. The PR body states the surface is a 3840x2160 **RGBA16F** intermediate; that is `Float16` with 4 components, which satisfies `float16_uint16_alias` at `live_compute.cpp:10937-10939`. So on this title **every** near-miss takes the retry path, and `format` / `vk_format` are reported as differing whether or not they actually did under the key the producer would have to match.

The PR body's decision table then says: *"`no_cache_entry` dominant, `same_addr>0` + a named field -> normalise that field"*. Two lines above the retry, `live_compute.cpp:10931` says *"Do not generalize this to signed, sRGB, component-count, or size aliases."* The instrument is set up to name the field the code specifically warns against normalising, for a reason that is an artefact of the instrument.

The PR's own sample output already shows the shape: `same_addr=8` with `format=7 vk_format=7` dominating `tile_mode=2 width=3`, against `alias_retry=17`. Nothing in the report distinguishes which key a mask came from. `no_entry_scans` does not expose it either — it counts *final-observation* misses (8), while scans actually **performed** is up to 8+17, so the denominator hides the retries rather than revealing them.

This defeats the PR's stated purpose in the one case it exists for, and the remedy is small: keep two masks (exact and alias) in the snapshot, or record the exact key's scan at the point it is taken rather than only from the final observation. Either way the report must say which key it is describing.

**Related, non-blocking, same mechanism.** `tests/gpu/recompiler/test_game_compute.cpp` (new block, after the `imported` CHECK) comments *"Pin that: the recorded observation is the RETRY's"*. The two arms it introduces — `key_field_diffs[kTileMode] == +1` and `key_field_diffs[kWidth]` unchanged — hold **identically** for the exact key's observation and for the retry's, because `tile_mode` differs and `width` matches under both. The fields that would discriminate are `format` and `vk_format`, and neither is asserted. The comment claims more than the arms establish; the negative arm on `kWidth` does correctly discriminate against an all-bits mask, which is what it was written for, but it does not pin which observation was recorded.

---

## Non-blocking findings

**1. Hardcoded query ordinals with nothing pinning them.** `compute_image_borrow_census.hpp:348` (`submit_query == 1` = Overlap) and `:352` (`watch_query == 1` = Dirty). Both are **correct today**: `src/gpu/execute/gpu_execute.hpp:1004-1008` is `{ Unchanged, Overlap, Unknown }` and `src/host/memory/guest_write_watch.hpp:9-13` is `{ Unchanged, Dirty, Unknown }`. But the census header includes neither enum, and the unit test at `test_compute_image_borrow_census.cpp:168-169,174,189` restates the **same literals** in a comment and in its inputs — so a renumbering upstream re-labels every authority bucket and both production and test agree on the new wrong answer. A `static_assert` in `live_compute.cpp`, where both headers are visible, closes it for free.

**2. The 2 KiB report buffer is thinner than its own comment implies.** `live_compute.cpp:386-391` sizes `char line[2048]` and notes that truncation "drops whole trailing lines, and the last line is the producer partition". With all 23 key fields non-zero the four lines come to roughly 1.7 KiB at 7-digit counters; at 9-10 digits they exceed 2048 and the **publish partition** — the half the comment identifies as separating "the key differs" from "nothing was ever published" — silently disappears. There is no truncation marker, so a truncated report is indistinguishable from one whose last line is absent, which is precisely the ambiguity the header's zero-bucket rule (`compute_image_borrow_census.hpp:435-440`) exists to prevent, and precisely the trap the PR body cites as motivation. Suggest 4096, or a `...` marker when `used == capacity - 1`.

**3. `no_entry_scans` does not count scans.** It counts final-observation `NoCacheEntry` misses; the exact key's scan is still executed (O(cache)) and thrown away on every alias retry. Diagnostic-cost only, and gated, but the name asserts something the counter does not measure.

**4. Relaxed atomics give no cross-counter ordering.** A snapshot taken concurrently with recording can observe a torn partition (outcome bumped, authority sub-counter not yet). Both readers here are same-thread or at-exit so it does not bite, but `test_compute_image_borrow_census.cpp:205-208` asserts the sums as invariants and they hold only quiescent. One line in the header would keep the next reader from relying on it.

**5. Scan locking.** `scan_image_cache_for_near_miss` (`live_compute.cpp:2518`) iterates `image_cache` (`:1369`, an `unordered_map` with no mutex). This is not a new **class** of exposure — `borrow_cached_image_for_graphics` already did an unlocked `find` and unlocked writes (`cached.last_use`, `++cached.pins`) — but it is the only whole-container traversal on the path and a materially longer window against a concurrent rehash. Gated off by default, so a default launch cannot reach it. Worth a sentence in the scan's comment saying it inherits the caller's discipline.

---

## What I verified and found clean

**Q1 — behaviour preservation. The accepted set is IDENTICAL.** I read the original chain with `git show origin/main:...live_compute.cpp` and compared term by term against `live_compute.cpp:10902-10917` and `classify_compute_image_import` (`compute_image_borrow_census.hpp:85-99`). All eleven terms map correctly. Specifically checked: `guest_bytes_in_range = guest_bytes != 0 && guest_bytes <= UINT32_MAX` is exactly the negation of `!guest_bytes || guest_bytes > UINT32_MAX` (the boundary is right — `<=` against `>`); `shape_supported = ordinary || cube` negates `(!ordinary && !cube)`; `have_context = context && context->device` (`:10903`) preserves the short-circuit that guards the deref. No term dropped, none inverted, no `!` lost.

**The hoist is safe and nothing else moved with it.** `native_storage_vk_format` (`live_compute.cpp:427-472`) is a pure switch over `DataFormat` and `components` with a `default: break;` — it touches no context, no globals, no memory. `components` and `sampled_native_format` stayed at their original positions; the diff deletes the `if` chain and places the classifier below them. The only consequence is that the format switch now also runs on declined imports, which is free.

**Q2 — the equivalence test is not vacuous and the transcription is honest.** `legacy_import_chain_declines` (`test_compute_image_borrow_census.cpp:40-49`) is one `||` chain with the original's grouping and line breaks, structurally *unlike* the classifier's cascade of early returns — which is what makes it a real cross-check rather than a restatement. I checked its eleven terms against `origin/main` myself. `accepted == 1` is exactly right: acceptance fixes all eleven bits, so precisely one of 2048 assignments accepts, and the pin fails for a constant classifier in either direction.

*Gap worth naming:* the sweep proves classifier == transcription over the **abstract booleans**. It does not cover the ten `gates.* = ...` assignments at `live_compute.cpp:10903-10913`, which is where a lost `!` would actually live. I verified those by hand and they are correct, but nothing in the suite would catch a future edit to them.

**Q3 — no gate's decision changes.** (a) `write_watch.query()` runs under exactly the original guard: `watch_consulted = !submit_unchanged && (bool)cached.write_watch` at `:2477`, and the query is behind it at `:2478-2479`. Identical to `!submit_unchanged && cached.write_watch && cached.write_watch.query() == ...`. (b) `exact_unchanged` at `:2485-2487` is unchanged including the `submit_query == Unknown && !watch_unchanged &&` short circuit in front of the mirror memcmp. (c) All four exits return the same value in the same cases; the `set_outcome` cascade mirrors the `||` order of the original condition; the success path's mutations (`graphics_export_snapshot` re-stamp, `last_use`, `++pins`, out-params) are in the original order. (d) The scan is a `const` member, iterates by const ref and writes only into `observation`; it is reached only when `observation && image_borrow_census_report_enabled()` (`:2461-2462`), the latter a `static const bool` from `getenv`. It cannot run on a default launch and cannot mutate the cache.

**Q4 — `authorize_cached_image_export` -> `bool`.** `live_compute.cpp:10477-10479` is `publish_eligible && image.graphics_sampled_usage && ctx.authorize_cached_image_export(...)`, so the call is made under exactly the original `if` condition — the added `&&` is to its right, not its left. `grep` over the tree finds no other caller, and there is no `[[nodiscard]]`, so no unused-result fallout.

**Q5 — the partitions are sound.** Both authority sub-partitions are total and disjoint (`if / else if / else` twice), each bumping exactly once per `AuthorityChanged`, so each sums to the `AuthorityChanged` count independently. `Unchanged` (ordinal 0) cannot reach the branch on either axis — it is the hit — so the remaining two-way splits are complete. The outcome partition: every path in `borrow_cached_image_for_graphics` sets an outcome, and `record_outcome` is called exactly once per accepted import with no early return between `:10917` and `:10971`, so `sum(outcomes[1..6]) == declines[None]`. No double-count, no fall-through. Ordinals verified against both enums (see non-blocking 1).

**Q6 — the formatter cannot write past the buffer.** `compute_image_borrow_census.hpp:441-458`. The invariant `used <= capacity - 1` holds on entry (`output[0] = '\0'; used = 0` with `capacity >= 1`); `emit` returns early when `used + 1 >= capacity`, so at the `snprintf` `room = capacity - used - 1 >= 1`, and `snprintf(output + used, room + 1, ...)` touches indices `used .. used + room == capacity - 1` at most. In bounds. `used` advances by `min(written, room)`, preserving the invariant. Always NUL-terminated, including `capacity == 1` (every emit declines, `used == 0`, `output[0] == '\0'`, returns 0). The returned `used` is the true `strlen` in both the truncated and untruncated cases. The test at `:286-301` sweeps capacity 1..96 with a `0x7f` canary past the boundary and asserts `small[used] == '\0'` — a genuine bounds test, not a smoke test.

**Q8 — the whole-binary env var does not perturb other assertions.** Enabling the gate adds only (i) periodic/atexit stderr output and (ii) the const scan. The only stderr-sensitive assertions in `test_game_compute.cpp` are in the device-loss capture at `:6753-6800`, and they use `std::string::find` substring matching, so interleaved `[compute-borrow-census]` lines cannot break them; the `atexit` report fires after the capture is restored. `g_image_borrow_census` has a trivial destructor, so the `atexit` handler is safe against static destruction order. The new deltas are exact (`== +1`), not `>`, so they discriminate — except `publishes[kAuthorized] > before`, which is looser than it needs to be but is not wrong.

**Claims audit.** *"`native_storage_vk_format` ... is a pure switch ... safe with a null context"* — true, verified at `:427-472`. *"`write_watch.query()` still runs only when the journal did not already answer"* — true. `STRAY_STATUS.md`'s `## Ruled out` row withdraws the admission claim as **unmeasured rather than falsified**, which is the correct and honest framing and is scoped exactly right. `AGENTS.md`'s claim that `[render-timing]`'s `protect=` is a *failure* counter is correct — `live_renderer.cpp:10908` formats `write_watch.create_protect_failures`. I re-ran `tools/env/check_diag_gates.py` in the worktree: rc=0, 639 files, 767 `PROSPER_*` variables, 108 findings all baselined, confirming that line of the body.

The one body claim I would qualify: *"Nothing else changes what any gate decides"* is true of the emulator's gates, but the instrument's own field-name output is changed by the alias retry as described above, and the body does not say so.

---

## Note on the working tree

The review worktree is **dirty** — there is uncommitted follow-up work on top of `3a88d671` in `compute_image_borrow_census.hpp`, `live_compute.cpp`, `test_compute_image_borrow_census.cpp` and `FRONTEND_APP.md`, splitting `journal_cross_submit` into `journal_unjournaled` / `journal_undecided`. **None of it is reviewed above**; every citation is from the committed head.

Flagging one thing from it as a courtesy, because it is the exact drift the ordinal finding is about. The new block adds:

    static_assert(static_cast<uint8_t>(prosper::gpu::GuestGpuWriteQuery::Overlap) == 2);
    static_assert(static_cast<uint8_t>(prosper::gpu::GuestGpuWriteQuery::Unknown) == 2);

`Overlap` is **1**, not 2 (`src/gpu/execute/gpu_execute.hpp:1004-1008`). As written the first assert fails to compile — which is the good outcome. The bad outcome is "fixing" it by changing the census's `submit_query == 1` to `== 2`, which would silently swap `journal_overlap` with the Unknown buckets. Please make the assert match the enum, not the other way round.

mattias800 (commented):
Second review, on head `96d27f18`. Read-only; no project build, no GPU. Every `file:line` is from this head.

## APPROVED

The blocking finding is fixed at the point I named, and the new arm pins it with exactly the pair I said nothing asserted. Five non-blocking observations follow, two of which are new and were found by measuring the sizing claim rather than reading it — none of them blocks.

---

## The blocking finding is resolved — verified

`borrow_cached_image_for_graphics` takes `scan_near_miss` (`live_compute.cpp:2465`, defaulting false) and the scan is gated on it at `:2473`. The exact key passes `/*scan_near_miss=*/true` (`:10947`) and the scan is recorded at `:10955`, **before** `observation = {}` at `:10996`; the alias retry passes `/*scan_near_miss=*/false` (`:10999`). So the retry's mask is not merely discarded, it is never computed — the O(cache) scan I flagged as wasted work is gone with it. `record_outcome` still runs once at `:11001` from the final observation, which is correct: the outcome must describe the attempt that decided the import, while the field list must describe the exact key. Those two now come from different points on purpose, and both header and call site say why.

The new arm at `test_game_compute.cpp:2208` is the right one. Asserting `key_field_diffs[Format]` and `[VkFormat]` are **unchanged** across the mis-keyed miss discriminates precisely, because the alias key rewrites those two by construction — under the previous head both would have incremented, so the arm goes red on the defect rather than passing for an unrelated reason. That is the gap I named, closed with the assertion that closes it. `:2213` pins `exact_key_scans` as scans performed, and the rename plus the `exact_key_fields:` label at `census.hpp:535` make the report say which lookup the list belongs to.

`static_assert(... GuestGpuWriteQuery::Overlap) == 1)` at `live_compute.cpp:391` is correct at this head — confirming the `== 2` I saw was your mutation arm caught mid-flight, as you said. Findings 3, 4 and 5 from the first review are addressed (counter renamed to what it counts; the relaxed-atomics caveat documented at `census.hpp:427-431`; the scan's inherited locking discipline documented at `live_compute.cpp:2532-2538`).

---

## New, non-blocking

**N1. The saturation arm does not measure the worst case it claims to.** `census.hpp:479-482` says *"Worst case, measured by the saturation arm in the test rather than estimated"*. I compiled the header standalone and measured it. All-max is **2550** bytes. But the percentages are `%.1f` of `100*part/whole`, and that is maximised by a **large numerator over a small denominator** — a shape the all-max arm cannot express, because it moves both dials together and every percentage collapses to `100.0` (5 chars) instead of its 24-char maximum. Setting the three denominators (`imports`, `declines[None]`, `publish_evaluated`) to 1 with the buckets at `UINT64_MAX` gives **2721**.

The conclusion survives comfortably — 2721 against 4096 is 1375 bytes of headroom, and 4 KiB is the right number — but the arm that justifies it does not establish it, which is the same-source positive control shape the charter records: the generator that produces the null cannot express the case being bounded. One extra arm with skewed denominators would make the sentence true. (It also confirms the first review's finding: 2550 already exceeded the old 2 KiB.)

**N2. The TRUNCATED predicate conflates "truncated" with "exactly filled".** `census.hpp:565` keys on `used == capacity - 1`, which is also true when the report exactly fits. The marker then overwrites 34 bytes of a **complete** report with a false claim. Demonstrated on an ordinary small census whose natural length is 838: calling with capacity 839 returns a report ending

    ...not_persistent=0 (0.0%) ... | render[compute-borrow-census] TRUNCATED

`renderer_accepted=0 renderer_rejected=0` destroyed, and truncation announced where none happened — the marker damaging the exact line it exists to protect.

Unreachable at the production call site: `live_compute.cpp:408` passes 4096 and the true maximum is 2721, so `used == 4095` cannot occur and the marker branch is dead code there. Your header already says the marker should never fire, so this is consistent rather than contradictory. But `format_compute_image_borrow_census` is a public inline with a public size constant, and the test itself calls it at 600 and across 1..96. Setting a `truncated` flag inside `emit` when `written > room` is a two-line fix and makes the predicate mean what its name says.

**N3. Your focused question — are `exact_key_scans` / `no_cache_entry` / `alias_retries` unambiguous now?** Better, and one direction is still undocumented, in the direction that misleads. `census.hpp:325-326` explains that the alias retry can miss without being scanned, so `outcomes[NoCacheEntry]` can exceed `exact_key_scans`. The reverse also happens: an exact-key miss that the alias retry **rescues** still increments `exact_key_scans`, `no_entry_same_addr`, and — by construction — `key_field_diffs[Format]` and `[VkFormat]`, while the final outcome is `Hit`.

That is the normal GTA V success path, not a corner case. So on any run where the alias arm works, the `exact_key_fields:` list is once again dominated by `format`/`vk_format` for a reason unrelated to any failure. It does **not** affect the Stray decision this instrument was built for — there the alias also misses, so those imports end in `NoCacheEntry` and the list is clean — and it is a much weaker residual than the original defect. But it is the same confound arriving by a different route, and a reader applying the decision table to a mixed run would meet it. Either add an `exact_key_scans_rescued` counter so the reader can subtract, or state in the decision table that the field list is read against runs where the borrow actually failed.

**N4. The wiring arms have no positive control on `base`.** `test_game_compute.cpp:2231` drives ten decline buckets one field at a time off `base = output` with `cls = Texture`, but nothing asserts that the unmutated `base` is accepted. The classifier returns the *first* failing term, so a bad `base` failing at a later term makes most arms fail loudly — the ordering protects you for nine of them. The residual gap is a `base` that already fails at exactly the term under test, which would let that one arm pass vacuously. I checked: `shader_resources.hpp:347,350,365,376` give `img_dim=1, depth=1, declared_mip_levels=1, in_mip_tail=false`, and `output` sets the rest, so `base` is clean today. Nothing pins it. One arm asserting the unmutated `base` increments `declines[None]` closes it.

I did verify the arms themselves are discriminating: each asserts the *named* bucket moved by exactly one **and** that `declines[None]` did not, so a lost `!` either accepts (fails the second) or lands elsewhere (fails the first). `guest_bytes_wide` at `UINT32_MAX + 1` exercises the `<= UINT32_MAX` boundary directly, which was the term I most wanted covered — that is the arm that closes the "gap worth naming" from the first review, and it closes it properly.

**N5. Two counts in the new material are off.** The block comment at `test_game_compute.cpp:2220` says *"Drive nine of them"*; the array has eleven cases covering **ten** distinct buckets. And `NoContext` is not among them — it is structurally undrivable from this path, since the backend is registered by the time these run. So the claim that "all eleven decline buckets now fire in that run" is not established by this array; if it is true it is true for another reason (imports before backend registration) and needs its own evidence. Not worth a test change — worth correcting the sentence.

---

## What I re-verified, and how

Independently, by compiling `compute_image_borrow_census.hpp` standalone (it has no project dependencies) rather than trusting the assertions:

- saturated report = 2550 bytes, skewed-denominator worst case = 2721, both under 4096 with 4 lines and no marker;
- a bounds sweep over capacity 1..4096 on a **saturated** census: no write past the buffer at any capacity, `used < capacity` always, `output[used] == '\0'` always. The formatter's arithmetic holds under the new marker code, including the `capacity <= marker_bytes` branch that suppresses the report entirely;
- the false-positive marker case above.

Behaviour preservation is untouched by this head — the classifier, its wiring, the `write_watch.query()` guard and the `exact_unchanged` short circuit are byte-identical to the head I already checked term by term against `origin/main`. The only production-path changes are the added `scan_near_miss` parameter (default false, so no other caller's behaviour moves), the `export_snapshot_unjournaled` field write inside the already-gated `AuthorityChanged` observation block, and the report buffer size.

The `journal_unjournaled` / `journal_undecided` split you found yourself is a genuine improvement and the partition stays total and disjoint: `if (Overlap) / else if (!armed) / else if (unjournaled) / else` bumps exactly once, and `test_compute_image_borrow_census.cpp` asserts all four sum to the `AuthorityChanged` denominator. `journal_undecided` is honestly named for covering two causes it cannot separate, which is the right call.

Approving. N1 through N5 are worth a follow-up commit if you want them, but none of them is a reason to hold this.

mattias800 (commented):
Third review, on head `0fb55422`. Read-only; no project build, no GPU. Every `file:line` is from this head, and the numbers below are ones I measured rather than read.

## APPROVED

All five non-blocking findings are fixed, and two of them are fixed better than I asked for. One small refinement remains, which is the direct answer to the question you posed.

---

## Answering the question you asked: no, you have not traded one confound for a subtler one

The four counters have four distinct meanings and the code is consistent with all of them. What they actually form is a **three-level nesting**, each level a proper subset of the one above:

    exact_key_scans                  every exact-key scan performed
      minus rescued                  -> lookups where the borrow really failed
        no_entry_same_addr           -> of those, how many found an entry at the address
          key_field_diffs[f]         -> of those, how many disagreed on field f

`record_no_entry_scan` (`compute_image_borrow_census.hpp:394-395`) implements exactly that: `rescued` is bumped as a subset of `exact_key_scans`, and `if (!same_address_entry || rescued_by_alias) return;` excludes rescued scans from **both** `no_entry_same_addr` and the field list — so the two stay consistent with each other, which is the failure I was looking for and did not find. The report prints all four on one line in nesting order, which reads correctly; I rendered it with 12 scans / 9 rescued / 2 same-address:

    exact_key_scans=12 (rescued=9, not counted below) same_addr=2 exact_key_fields: tile_mode=2

**The one refinement (non-blocking).** `census.hpp:324` and `FRONTEND_APP.md:501` both say the field list is read against `exact_key_scans` minus `rescued`. That is the denominator for `no_entry_same_addr`, not for the field list — a field can only be counted when `same_addr` fired, so the field list's own denominator is `no_entry_same_addr`. In the line above, `tile_mode=2` against `same_addr=2` is **100%** of near-misses disagreeing on tile_mode; against the broader `12 - 9 = 3` it is 67%, diluted by a scan that found nothing at the address at all. That dilution matters because "nothing at this address" is a *different row of your own decision table* — the one that says the producer is absent — so the broader denominator folds two rows together. Naming `no_entry_same_addr` as the field list's denominator in both places fixes it; the counters themselves need no change.

Marginal, and I mention it only for completeness: `rescued_by_alias=borrowed` at `live_compute.cpp:11007` means the *borrow* succeeded, while `census.hpp:330` describes it as "the import went on to succeed". Those differ on the lease-allocation failure path, which is separately counted by `lease_failures`. Not worth a code change.

---

## The five findings, verified

**N3 (the one worth having) — fixed, and the mechanism is right.** The exact key's observation is copied at `live_compute.cpp:10953` and held across the retry; the record happens at `:11003-11007` once `borrowed` is known. I checked that `borrowed == true` at that point *implies* the alias retry hit: the exact call returned false (it produced a `NoCacheEntry` scan), and the retry is the only thing that can flip it — so the predicate cannot mislabel. The unit arm asserts both halves (the count moves, `Format`/`VkFormat` stay at zero), which is what makes it a partition change rather than a filter.

**N1 — fixed, and the arm now proves its own extremality.** Measured independently by compiling the header standalone: all-max is **2600** bytes, the skewed-denominator construction is **2771**, so `CHECK(widest_used > all_max)` at `test_compute_image_borrow_census.cpp:355` is a real satisfiable assertion and not a tautology. Headroom is 1325 bytes and the `+512` check passes. Keeping the all-max computation purely as the thing to compare against is the right shape: the arm now demonstrates that the obvious construction is *not* the worst case, which is the whole lesson, rather than asserting it in a comment.

**N2 — fixed, and my false positive is gone.** `emit` sets `truncated` on both loss paths (`census.hpp:518` and the `written > room` branch) and the marker keys on the flag at `:595`. Re-ran my demonstration: natural length 858, capacity 859 now yields `used=858`, **no marker**, and output **byte-identical** to the complete report. One byte short still marks. I also re-derived that `truncated == true` implies `used == capacity - 1`, so the marker's write remains in bounds — the fix does not weaken the bound it replaced. A fresh bounds sweep over capacity 1..4096 on the *widest* census: no overrun, `used < capacity`, NUL-terminated at every capacity.

**N4 — fixed, and the control is placed where it does work.** `test_game_compute.cpp:2276` asserts the unmutated `base` lands in `declines[None]` *before* the loop. That closes the exact hole: first-failing-term ordering meant a `base` already failing at the term under test would let that arm pass vacuously, and now a future edit to `output` reddens one named check instead of silently hollowing out the array.

**N5 — fixed.** `test_game_compute.cpp:2223` now says ten of eleven, names `NoContext` as structurally undrivable there, and tells the reader not to read the array as covering all eleven. The "all eleven buckets fire" claim is gone.

---

## Scope check

The production-path delta since the head I approved is three things: the observation copy at `live_compute.cpp:10953`, the deferred `record_no_entry_scan` call at `:11003-11007`, and the `truncated` flag in the formatter. Nothing touches the classifier, its wiring, the `write_watch.query()` guard, or the `exact_unchanged` short circuit — all still byte-identical to the head I checked term by term against `origin/main`, so the behaviour-preservation verification carries forward unchanged. The census still records exactly one outcome per accepted import and one scan per scan performed, and both partitions remain total and disjoint.

Approving. The denominator wording above is worth a one-line follow-up in both `census.hpp:324` and `FRONTEND_APP.md:501`, but it is documentation of a correct instrument, not a defect in it — merge on green CI without waiting on me.

mattias800 (commented):
Fourth review, on head `d9bd1f4f`. You said no re-review was needed; I did one anyway, because a discriminator rewritten *because it was vacuous* is the single thing most worth checking independently — and the check is cheap. It holds.

## APPROVED

**I reproduced both halves of your vacuity report rather than taking it on trust.** I copied the census header, applied the mutation you describe (drop the `same_addr` gate on the field loop, keep it on the counter: `if (rescued) { bump(rescued_); return; } if (same_address_entry) bump(no_entry_same_addr_);`), and ran both the current arm and the arm as it stood at `0fb55422` against the real and mutated recorders:

| arm | real recorder | mutated recorder |
| --- | --- | --- |
| current (`test_compute_image_borrow_census.cpp:250-296`) | green | **red** — fails on `Width == 0` and `Height == 0` |
| previous (`0fb55422`) | green | **green** |

So the previous arm genuinely could not tell the two implementations apart, and the non-empty-mask-without-a-same-address-entry input at `:257-258` is exactly what supplies the discrimination. Your diagnosis was right and the fix is the minimal one that closes it.

The reason it works is worth stating because it generalises: the old inputs made "fields are gated on `same_addr`" and "fields happen to be empty" the *same observation*, so the arm was asserting a property its own inputs could not distinguish from its negation. Constructing the state production never builds — a mask offered with no same-address entry — is what made the gate observable at all. That is the same shape as the charter's positive-control rule, applied to a negative: testing the contract at the boundary the **public recorder** names rather than at the boundary the current caller happens to produce. Recording it in the test's own comment where it happened is the right place for it.

I also checked the strictness claim: the levels come out `4 > 3 > 1 >= 1`, so both inequalities at `:299-301` are strict on real inputs and would not hold for an implementation that collapsed the levels. The final `<=` cannot be made strict (a field may equal `same_addr`), and asserting it across every field index rather than one is the stronger form.

**The denominator correction is right in all three places.** `census.hpp:331` now says `key_field_diffs` is read against `no_entry_same_addr` "and against nothing else", with the chain above it and the worked 100%-vs-67% example; `FRONTEND_APP.md:501` says the same and names the dilution as the *other* decision-table row. I re-derived the chain from the code: `no_entry_same_addr` is bumped only under `!rescued && same_address_entry`, so it is bounded by the non-rescued scans; and each same-address scan contributes at most one bump per field, so `key_field_diffs[f] <= no_entry_same_addr`. Both inequalities hold by construction, not by coincidence.

Scope: this head touches one header comment, one doc paragraph and one test arm. No production behaviour changes, so everything I verified on the previous three heads carries forward unchanged.

---

## A correction to my own last review

I ended the third review with "merge on green CI without waiting on me". **That was mine to withdraw, and you were right to decline it.** A reviewer's approval is not merge authorization in this project — it is one of the three things the merge step requires, and it is not the one that says whether merging is wanted at all. Your task said open a PR and do not merge; my sentence had no standing to override that, and I should have written "approved, nothing further from me" and stopped. Leaving it for the user along with the run request is the correct call.

The census is in good shape. Three rounds of findings, all closed, and the two that mattered most — the alias-key attribution and this vacuous arm — were both about the instrument telling a reader something true-sounding that it had not actually measured, which is the failure this whole PR exists to prevent in the first place.

Issue #2233, opened 2026-08-07
prosper-app: F8/F9 captures are reachable only by a human keypress, so headless agents cannot use the best diagnostics we have


mattias800:
`prosper-app`'s two richest diagnostics are **F8** (a bounded `.prperf` performance capture, 5 s before + 5 s after the press) and **F9** (a replayable `.prgbundle` + `.bmp` frame grab). F9 could be reached **only by a human pressing a key**.

Every agent in this project runs headless. So the best instrument in the toolbox was the one nobody could aim, and lanes built weaker ones instead — a sampling profiler, a `getenv` counter, pass-log censuses — to answer questions the F8 report answers directly.

There is a second reason this went unnoticed for so long, and it is worth recording: **the triggers live in `prosper-app`, which is not built by default** (`-DPROSPER_APP=ON`). An env-var sweep over `src/` never sees them. `PROSPER_PERF_CAPTURE_AFTER_MS` already existed and I did not know it; F9 had nothing at all.

## What this costs in practice

Worked example from today (#2215). Blue Prince collapses to ~1 fps during the intro FMV. Two lanes spent hours on it: one built a sampling profiler on Windows, I built an `LD_PRELOAD` `getenv` counter on Linux, and **both of my successive explanations were wrong**. The project owner then pressed F8 once, and the report said immediately:

```
guest flips=1.69/s
components: renderer-resource=1516.5ms (47%)  readback=317.7ms  gpu-wait=137.9ms
            gpu-device=71.3ms (2.2%)  compute=53.1ms (1.7%)
```

— and that `buffer copy`, the leaf both lanes were optimising, is **2.4%** during the collapse. That is a different problem from the one being worked, and one keypress separated the two.

## Proposal

Four one-shot, opt-in triggers — both captures on both addressing axes:

| capture | after a delay | at a host frame |
| --- | --- | --- |
| F9 bundle | `PROSPER_GRAB_BUNDLE_AFTER_MS` | `PROSPER_GRAB_BUNDLE_AT_FRAME` |
| F8 perf | `PROSPER_PERF_CAPTURE_AFTER_MS` *(exists)* | `PROSPER_PERF_CAPTURE_AT_FRAME` |

Both axes because they answer different questions: `_AFTER_MS` aims at an event only describable in time ("when the movie starts"), `_AT_FRAME` at a reproducible ordinal already located with a cheap screenshot sweep.

[...]

Issue #1704, opened 2026-08-01
No Vulkan validation layer runs anywhere — one CI pass under VK_LAYER_KHRONOS_validation would have caught #1690 at the dispatch


mattias800:
Raised during review of #1698 (the #1690 fix).

## Observation

`VK_LAYER_KHRONOS_validation` appears exactly once in the tree — as a **comment** in
`prosper/tests/test_present_blit.cpp:10`. No test, frontend, or CI job enables it.

## Why that mattered concretely

#1690 was an invalid descriptor binding: a `VK_IMAGE_VIEW_TYPE_3D` view bound to a descriptor whose
SPIR-V module declared `OpTypeImage ... 2D`. Vulkan requires those to match, so the behaviour was
undefined and three drivers resolved it three different ways — two returned the seeded texel and one
returned zero. The validation layer flags exactly this class at the dispatch, by name, immediately.

Instead it presented as an unexplained Mesa-version split, was attributed to an upstream driver
defect, and cost `game_compute_exec` its place on CI — 484 executed assertions — until the SPIR-V was
disassembled by hand. That is a long way to travel for something a layer prints in one line.

It is also not a one-off: #1702 is the same class still live in the compute backend, where the 3D
binding derives its Vulkan view type from the resource alone without cross-checking the module's
reflected Dim. A validation pass would catch that one the moment a title reaches it.

## Suggested shape

One additional CI step (or a variant of the existing `Test` step) running the Vulkan-execution tests
with the layer enabled:

```yaml
sudo apt-get install -y vulkan-validationlayers
VK_INSTANCE_LAYERS=VK_LAYER_KHRONOS_validation ctest --test-dir prosper/build-ci -R '<gpu tests>'
```

Two things to decide when implementing:

1. **Advisory or blocking?** A first pass will almost certainly surface pre-existing findings across
   the renderer and compute backend. Landing it as advisory (collect and report, do not fail the job)
   and tightening once the list is empty is probably the realistic path — but an advisory check that
   nobody reads is worth very little, so it needs an explicit owner or a follow-up to make it
   blocking.
2. **Cost.** The layer is slow, particularly on lavapipe where the whole suite is already software
   rendered. It may want to run on a subset, or on a schedule rather than per-PR.

`CONFIDENCE: HIGH` that the layer is absent and that it detects the #1690 defect class.
`CONFIDENCE: LOW` on how much pre-existing noise a first run produces — that is the thing to measure
before choosing advisory vs blocking.

[...]

PR #3322, opened 2026-09-04
diag: aim RenderDoc headlessly by frame ordinal or elapsed time (#3321)

### LINKED ISSUE #3321: RenderDoc cannot be aimed headlessly: add frame-ordinal and elapsed capture triggers
## Problem

RenderDoc is installed on the Linux development box and is the strongest frame debugger available to
this project. It answers **"which draw wrote this pixel"** and **"what is in this render target at
draw N"** — the two questions a black-world bug asks, and the two prosper's own instruments answer
worst. It has nevertheless been unusable here for one mechanical reason, already recorded in
`docs/GPU_PROFILING_EXTERNAL.md`:

> `renderdoccmd capture` triggers on a keypress rather than a frame ordinal, which makes it awkward
> headless.

Every agent on this project runs headless. So the best frame debugger available was the one nobody
could aim — **exactly the failure #2233 fixed for prosper's own F8/F9 captures, one tool over.**

## What is actually available here (measured 2026-09-04)

| | state |
| --- | --- |
| `renderdoccmd`, `qrenderdoc`, capture layer | **installed** (`renderdoc-1.45`, in the distrobox) |
| `renderdoc_app.h` (in-application API) | **packaged** — `renderdoc-devel`, `/usr/include/renderdoc_app.h` |
| `renderdoccmd convert -c xml` | **works** — full event list, no Python |
| Python bindings (`import renderdoc`) | **absent, and not packaged at all** |

The Python bindings gate *replay*-based analysis — pixel history, per-draw render-target export,
shader debugging. Those need a RenderDoc source build and are **not** part of this issue.

## Fix

RenderDoc's in-application API needs no keypress: the library exports one symbol, `RENDERDOC_GetAPI`,
returning a struct with `StartFrameCapture`/`EndFrameCapture`. Those delimit a capture **explicitly**,
so they need neither a keypress nor a present — which is what makes them work under
`SDL_VIDEODRIVER=offscreen`.

```bash
ENABLE_VULKAN_RENDERDOC_CAPTURE=1 \
PROSPER_RENDERDOC_AT_FRAME=900 \
PROSPER_CAPTURE_DIR=$HOME/work \
    ./prosper-app <DUMP_ROOT>/<TITLE_ID>-app0
```

`PROSPER_RENDERDOC_AFTER_MS` is the wall-clock axis, matching F8/F9 exactly.

## Two things that had to be measured, and both were wrong on the first guess

**1. A prosper "frame" is the GUEST's, not the app loop's.** A span of one app-loop iteration
produced a capture with **no draws in it** — prosper submits guest GPU work off the app-loop thread,
and that loop can spin at 260 fps while the guest renders far slower. The span now closes on a guest
present, with an iteration cap.

**2. prosper runs TWO Vulkan instances, and the guest draws are not on the obvious one.** This is the
finding worth recording. prosper-app owns an instance for presentation (`main.cpp:249`); the live
renderer owns a separate one (`tests/fixtures/render_runner.h` — that header *is* the renderer
despite its directory) and publishes it via `prosper::gpu::set_shared_vulkan_context`. Passing NULL
lets RenderDoc choose, and **it chooses the one holding the swapchain** — the presentation device,
whose entire contribution is a blit and a present.

Measured on *The Messenger*, same build, same route, the only difference being which device the
capture is aimed at:

| aimed at | chunks | `vkCmdDrawIndexed` | `vkCmdDispatch` | pipelines |
| NULL (RenderDoc picks → presentation device) | 56 | **0** | **0** | **0** |
| the renderer's instance | **406** | **9** | **2** | **6** |

The first capture is not corrupt and not empty — it is a **complete, valid capture of the wrong
device**, and it converts, opens and reads perfectly while containing nothing anyone wants. That is
the failure mode worth naming: it looks like a working instrument reporting a negative result.

### Side finding, not fixed here

Under `SDL_VIDEODRIVER=offscreen` the renderer's instance is not surface-capable, so present
unification (#1270) declines and prosper-app falls back to its own device — the captured presentation
work is `vkCmdCopyBufferToImage` + `vkCmdBlitImage`, i.e. **the finished guest frame crosses from the
renderer's device to the presentation device through CPU memory, once per presented frame.** Every
headless measurement this project makes pays that. Worth its own issue.

mattias800:
Fixes #3321.

## The failure scenario

RenderDoc is installed on the Linux development box and is the strongest frame debugger available to
this project — it answers **"which draw wrote this pixel"** and **"what is in this render target at
draw N"**, the two questions a black-world bug asks and the two prosper's own instruments answer
worst. It has been unusable here the whole time for one mechanical reason, which
`docs/GPU_PROFILING_EXTERNAL.md` already recorded and then shrugged at:

> `renderdoccmd capture` triggers on a keypress rather than a frame ordinal, which makes it awkward
> headless; RGP's `MESA_VK_TRACE_FRAME` is the better automated capture.

Every agent on this project runs headless. So the best frame debugger on the machine was the one
nobody could aim — **precisely the failure #2233 fixed for prosper's own F8/F9 captures**, one tool
over, and with the same consequence: lanes built weaker instruments instead.

## The contract

RenderDoc's in-application API needs no keypress. `librenderdoc.so` exports exactly one symbol,
`RENDERDOC_GetAPI`, returning a struct of function pointers including `StartFrameCapture` /
`EndFrameCapture`. Those delimit a capture **explicitly**, so they need neither a keypress nor a
present — which is what makes them work under `SDL_VIDEODRIVER=offscreen`, where there is no window
to press a key into.

```bash
ENABLE_VULKAN_RENDERDOC_CAPTURE=1 \
PROSPER_RENDERDOC_AT_FRAME=900 \
PROSPER_CAPTURE_DIR=$HOME/work \
    ./prosper-app <DUMP_ROOT>/<TITLE_ID>-app0

renderdoccmd convert -f cap.rdc -c xml -o frame.xml     # full event list, no Python needed
```

| variable | axis |
| --- | --- |
| `PROSPER_RENDERDOC_AT_FRAME` | a located host-frame ordinal (reproducible) |
| `PROSPER_RENDERDOC_AFTER_MS` | a wall-clock event you can only describe in time |
| `PROSPER_RENDERDOC_MAX_ITERS` | span cap, default 2000 |

Both triggers are one-shot and opt-in, and a malformed value **disables its own trigger** rather than
firing at an unintended moment — the same contract as F8/F9, reusing their `parse_*` helpers.

### 1. A prosper "frame" is the GUEST's, not the app loop's

A span of one app-loop iteration produced a capture with **no draws in it**. prosper submits guest
GPU work off the app-loop thread, and that loop can spin at 260 fps while the guest renders far
slower, so the span routinely covered only the presentation blit and `vkQueuePresentKHR`. The span
now closes on a **guest present** (`gpu::present_count()`), with an iteration cap so a title that
stops presenting cannot hold a capture open until the process runs out of memory. The log reports
which of the two closed it, because a cap-closed capture may contain no guest frame at all.

### 2. On this route prosper runs two Vulkan instances, and the guest draws are not on the obvious one

This is the finding worth the PR. prosper-app owns an instance for presentation
(`main.cpp:249`); the live renderer owns a **separate** one (`tests/fixtures/render_runner.h` — that
header *is* the renderer, despite its directory) and publishes it via
`prosper::gpu::set_shared_vulkan_context`.

**"Two" is the default headless route's count, not an invariant** — `live_compute.cpp:2913` creates a
*third*, private instance when it declines to adopt the renderer's, and present unification (#1270)
is opt-out and attempted on every real game boot, so a windowed run is *expected* to collapse to one.
Passing NULL lets RenderDoc choose the device; RenderDoc documents that choice as **arbitrary**, and
observed here (n=1) it took the one holding the swapchain. **The rule is therefore "always pass the
device explicitly", not "it picks the swapchain"** — an arbitrary choice that happens to be right
once is worse than a wrong one, because it will not stay right.

Measured on *The Messenger*, same route and **identical span logic**, differing only in which device
the capture is aimed at (two builds, one edit apart — not literally the same binary):

| aimed at | chunks | `vkCmdDraw` | `vkCmdDrawIndexed` | `vkCmdDispatch` | graphics pipelines | compute pipelines | shader modules |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| NULL (RenderDoc picks) | 56 | **0** | **0** | **0** | **0** | **0** | **0** |
| the renderer's instance | **406** | 2 | **9** | **2** | **6** | **1** | **13** |

**The 56-chunk capture is not corrupt and not truncated.** It is a complete, valid capture of the
*wrong device* — it converts, opens and reads perfectly while containing nothing anyone wants. Its
entire content is `vkCreateSwapchainKHR`, `vkCmdCopyBufferToImage`, `vkCmdBlitImage`,
`vkQueuePresentKHR`. It reads as **a working instrument reporting a negative result**, which is the
dangerous shape this project keeps records about. The trigger now resolves the renderer's instance
itself and warns loudly when the renderer has not yet published a context.

## Side finding, deliberately NOT fixed here

The presentation device's captured work is `vkCmdCopyBufferToImage` + `vkCmdBlitImage` — i.e. under
`SDL_VIDEODRIVER=offscreen` the renderer's instance is not surface-capable, present unification
(#1270) declines, and **the finished guest frame crosses from the renderer's device to the
presentation device through CPU memory, once per presented frame.** Every headless measurement this
project makes pays that. It wants its own issue and its own evidence; it is recorded here rather than
acted on.

## Why `renderdoc_app.h` is vendored rather than `#include`d from `renderdoc-devel`

The struct is an **ABI**, and using the real header is the whole point — a hand-written subset one
field out calls the wrong function pointer and crashes. Vendoring also means this compiles with no
build-time dependency on `renderdoc-devel` being installed. MIT, header-only, **never linked**;
`third_party/renderdoc/` with LICENSE and README per the charter's vendoring exception. prosper
requests API **1.4.1**, well below the installed 1.45, so an older RenderDoc still works.

## Affected / unaffected

- **Affected:** `prosper-app` only, and only when one of the two new variables is set. `RTLD_NOLOAD`
  means the loader never pulls RenderDoc into a process that does not already have it.
- **Unaffected:** every existing capture path (F8/F9/screenshot triggers are untouched), all other
  frontends, all platforms other than Linux/macOS (where the trigger compiles to a no-op that *says
  so* rather than silently doing nothing).
- **Risk:** the RenderDoc layer is an implicit Vulkan layer, and `live_compute.cpp`'s `std::atexit`
  ordering comment already anticipates exactly this class of layer for teardown safety (#1704). No
  change was needed there; noted so a reviewer checks rather than assumes.

## Documentation

`GPU_PROFILING_EXTERNAL.md`'s RenderDoc section said the keypress "makes it awkward headless" and
recommended RGP instead. **That claim is now false and is corrected in the same PR** rather than left
to steer the next reader wrong — including an explicit statement of what the absent Python bindings
still cost (pixel history, per-draw render-target export and shader debugging remain out of reach
without a RenderDoc source build; `convert` and handing a `.rdc` to `qrenderdoc` are what work).

`frontends/shared/diagnostics/AGENTS.md` is new — the folder had none.

## Verification

```
cmake --build prosper/build-linux --target prosper-app -j6     # clean
```

Functional evidence is the table above: two captures, same route and identical span logic (two
builds, one edit apart), differing only in which device the capture is aimed at,
converted with `renderdoccmd convert -c xml` and counted. Full `ctest` results posted as a comment
(the GPU is in use by a Stray capture run as this is opened, and this project does not run
concurrent GPU jobs).

---

## Review round 1 — head `25653d3` — REJECTED, then APPROVED on `f6e98298`

**Blocking (B1): an open capture was never closed on any loop-exit path**, so the artifact was lost
with no line saying so — `--frames` satisfied (`PresentedFrameCounter::record` clears `running` in the
same iteration that `AT_FRAME` arms), `SDL_EVENT_QUIT` via the early `break`, and `AFTER_MS` firing
shortly before a title exits, which is the wall-clock axis's whole purpose. That is exactly the
silent-instrument failure this PR argues against, one level up. My own testing could not have found
it: I verified the trigger only on runs I killed externally, the one shape that never exercises a
graceful exit.

Fixed by **closing** rather than cancelling (deliberately unlike the F8 sibling two lines below):
RenderDoc decides whether the span held anything, so real work is written and labelled, and an empty
span reports `ABANDONED`. The reviewer verified completeness by enumerating the loop body for every
`return`/`exit`/`_Exit`/`abort`/`terminate` — zero hits, so every departure is a `break` or the
condition, and both reach the close.

**Then, from the re-review:** a shutdown-truncated `.rdc` still opens perfectly and reads as a whole
frame with draws missing. The warning was in stderr, which does not travel with the file. Every
capture is now stamped via `SetCaptureFileComments` with what closed the span, how many iterations and
guest presents it covered, and which device it was aimed at — and a truncated one says missing draws
are expected and are **not** evidence the guest failed to submit them.

**Corrections to claims this body previously made** (all now fixed above): "TWO Vulkan instances"
stated as an invariant; "it chooses the one holding the swapchain" (n=1, docs say arbitrary); "same
build" (two builds, one edit apart); the pipelines column omitted 1 compute pipeline.

**Also from the review:** `?:` was a GNU extension and the repo's only use — spelled out;
`MAX_ITERS` now echoes the offending value and accepted shape, matching
`PROSPER_CAPTURE_BUNDLE_MAX_MB` (`main.cpp:1750`), which the reviewer identified as the repo's
existing convention for a malformed **bound** as opposed to a malformed **trigger** — that precedent
settles the question in favour of the doc as written; a second axis arming inside an open span now
reports that it was consumed; `third_party/renderdoc/README.md` cited a nonexistent `.cpp`; and the
`BLOG.md` entry was 229 words with an 8-number table against the charter's one-sentence rule — cut to
80 words with no table.

**Filed rather than fixed here:** #3324 (a guest-thread `_Exit` bypasses every frontend finalize step;
`_Exit` is `exit_group` so no `atexit` hook helps — documented at the call site) and #3325
(`g_shared_vulkan` is published and read unsynchronised; pre-existing, belongs with the global rather
than with one more reader).

### DIFF
--- a/BLOG.md
+++ b/BLOG.md
@@ -21,6 +21,27 @@ from the tracker issues, and still gated, because it is a projection of state ra
 
 ## 2026-09-04
 
+### RenderDoc works on prosper now, and the first capture was a beautiful picture of the wrong thing
+
+No picture — or rather, 56 of them, all useless. RenderDoc has been installed on the dev box the
+whole time and unusable, because `renderdoccmd` only starts a capture on a **keypress** and every
+agent here runs headless. prosper now drives RenderDoc's in-application API itself, so a capture can
+be aimed at a frame ordinal or a wall-clock moment exactly like F8 and F9.
+
+The first working capture contained **zero draws**, and the reason is the good bit: prosper runs two
+Vulkan instances — one that presents, one that actually renders the game — and RenderDoc, asked
+politely to pick, picks the one holding the swapchain. Same build, same route, only the aim differing:
+
+| aimed at | chunks | draws | dispatches | pipelines |
+| --- | ---: | ---: | ---: | ---: |
+| whatever RenderDoc picks | 56 | **0** | **0** | **0** |
+| the renderer's device | **406** | **11** | **2** | **6** |
+
+That first capture isn't corrupt or truncated. It opens perfectly, converts perfectly, and contains
+nothing anyone wants — a working instrument reporting a confident negative. Worth remembering next
+time a tool says "nothing here".
+
+
 ### Two thirds of Stray's biggest render cost was the allocator, not the work
 
 No picture. The largest single leaf in a capture of Stray's 9.76 fps title screen is the graphics
--- a/BLOG.md
+++ b/BLOG.md
@@ -23,24 +23,11 @@ from the tracker issues, and still gated, because it is a projection of state ra
 
 ### RenderDoc works on prosper now, and the first capture was a beautiful picture of the wrong thing
 
-No picture — or rather, 56 of them, all useless. RenderDoc has been installed on the dev box the
-whole time and unusable, because `renderdoccmd` only starts a capture on a **keypress** and every
-agent here runs headless. prosper now drives RenderDoc's in-application API itself, so a capture can
-be aimed at a frame ordinal or a wall-clock moment exactly like F8 and F9.
-
-The first working capture contained **zero draws**, and the reason is the good bit: prosper runs two
-Vulkan instances — one that presents, one that actually renders the game — and RenderDoc, asked
-politely to pick, picks the one holding the swapchain. Same build, same route, only the aim differing:
-
-| aimed at | chunks | draws | dispatches | pipelines |
-| --- | ---: | ---: | ---: | ---: |
-| whatever RenderDoc picks | 56 | **0** | **0** | **0** |
-| the renderer's device | **406** | **11** | **2** | **6** |
-
-That first capture isn't corrupt or truncated. It opens perfectly, converts perfectly, and contains
-nothing anyone wants — a working instrument reporting a confident negative. Worth remembering next
-time a tool says "nothing here".
-
+No picture — or rather 56 of them, all useless. RenderDoc has been installed the whole time and
+unusable, because it only starts a capture on a *keypress* and every agent here runs headless.
+prosper now drives its in-application API itself, so a capture aims at a frame ordinal like F8 and F9
+do. The first working one held zero draws: prosper runs two Vulkan devices, and asked to pick,
+RenderDoc took the one that only presents. Details in [#3321](https://github.com/mattias800/prosper/issues/3321).
 
 ### Two thirds of Stray's biggest render cost was the allocator, not the work
 
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -3395,7 +3395,10 @@ if(PROSPER_APP AND TARGET SDL3::SDL3)
     add_executable(prosper-app frontends/prosper-app/main.cpp frontends/prosper-app/library_ui.cpp
                                frontends/prosper-app/library_media.cpp
                                frontends/prosper-app/fps_overlay.cpp)
-    target_include_directories(prosper-app PRIVATE src third_party/stb)
+    # third_party/renderdoc: the vendored in-application API header (#3321). Header-only and
+    # never linked -- librenderdoc.so is dlopen'd at runtime by the capture trigger.
+    target_include_directories(prosper-app PRIVATE src third_party/stb third_party/renderdoc
+                                                   frontends)
     target_link_libraries(prosper-app PRIVATE prosper_core SDL3::SDL3 Vulkan::Vulkan prosper_imgui
                                               prosper_performance_capture)
     target_compile_definitions(prosper-app PRIVATE PROSPER_HAVE_LIBRARY_UI=1)
--- a/prosper/docs/GPU_PROFILING_EXTERNAL.md
+++ b/prosper/docs/GPU_PROFILING_EXTERNAL.md
@@ -90,16 +90,25 @@ capture *explicitly*, so they need neither a keypress nor a present. prosper dri
 
 ### Aim it at the RENDERER's device. A capture of the wrong one looks perfectly healthy.
 
-**prosper runs two Vulkan instances.** prosper-app owns one for presentation (`main.cpp`); the live
-renderer owns another (`tests/fixtures/render_runner.h` — that header *is* the renderer) and
-publishes it through `prosper::gpu::set_shared_vulkan_context`. Every guest draw is on the
-renderer's. Passing NULL lets RenderDoc choose and **it chooses the swapchain's** — the presentation
-device. Measured on *The Messenger*, same build and route, only the aim differing:
+**On the default headless route prosper runs two Vulkan instances.** prosper-app owns one for
+presentation (`main.cpp`); the live renderer owns another (`tests/fixtures/render_runner.h` — that
+header *is* the renderer) and publishes it through `prosper::gpu::set_shared_vulkan_context`. On that
+route every guest draw is on the renderer's. **Two qualifications**: `live_compute.cpp:2913` can
+create a *third*, private instance when it declines to adopt the renderer's, and present unification
+(#1270) can collapse the two into one when the renderer's instance is surface-capable — so "two" is
+this route's count, not an invariant.
 
-| aimed at | chunks | `vkCmdDrawIndexed` | `vkCmdDispatch` | pipelines |
-| --- | ---: | ---: | ---: | ---: |
-| NULL (RenderDoc picks) | 56 | **0** | **0** | **0** |
-| the renderer's instance | **406** | **9** | **2** | **6** |
+Passing NULL lets RenderDoc choose the device. RenderDoc documents that choice as arbitrary; observed
+here (n=1) it took the one holding the swapchain. **The lesson is therefore "always pass the device
+explicitly", not "it picks the swapchain"** — an arbitrary choice that happens to be right once is
+worse than a wrong one, because it will not stay right. Measured on *The Messenger*, same route and
+identical span logic, differing only in which device the capture is aimed at (two builds, one edit
+apart):
+
+| aimed at | chunks | `vkCmdDrawIndexed` | `vkCmdDispatch` | graphics pipelines | compute pipelines |
+| --- | ---: | ---: | ---: | ---: | ---: |
+| NULL (RenderDoc picks) | 56 | **0** | **0** | **0** | **0** |
+| the renderer's instance | **406** | **9** | **2** | **6** | **1** |
 
 The 56-chunk capture is not corrupt and not truncated — it is a **complete, valid capture of the
 wrong device**, and it converts and opens perfectly while containing nothing anyone wants. It reads
--- a/prosper/docs/GPU_PROFILING_EXTERNAL.md
+++ b/prosper/docs/GPU_PROFILING_EXTERNAL.md
@@ -66,22 +66,56 @@ check available before anyone spends time reading it.
 reader, so an agent can *produce* a capture but a human opens it. Say so when handing one over rather
 than implying you read it.
 
-## RenderDoc — installs here, and `convert` is the headless half
+## RenderDoc — aimable headlessly since #3321, and `convert` reads it with no Python
 
 ```bash
-sudo dnf install -y renderdoc      # 1.45 on this box, Vulkan supported
-renderdoccmd convert -f cap.rdc -c chrome.json -o cap.json
+sudo dnf install -y renderdoc renderdoc-devel     # 1.45 on this box, Vulkan supported
+
+# Capture one guest frame, headless, no keypress. Either axis, exactly like F8/F9.
+ENABLE_VULKAN_RENDERDOC_CAPTURE=1 \
+PROSPER_RENDERDOC_AFTER_MS=175000 \
+PROSPER_CAPTURE_DIR=$HOME/work \
+    ./prosper-app <DUMP_ROOT>/<TITLE_ID>-app0
+
+renderdoccmd convert -f cap.rdc -c xml -o frame.xml      # full event list, no Python
 ```
 
-`convert` targets `chrome.json` (Chrome/Perfetto timeline — parseable by script and openable in the
-Perfetto UI) and `xml`. That is the route for headless per-event analysis.
+**This section used to say `renderdoccmd capture` triggers on a keypress "which makes it awkward
+headless", and recommended RGP instead. That was true of `renderdoccmd` and is no longer true of
+prosper.** RenderDoc's in-application API needs no keypress — `librenderdoc.so` exports
+`RENDERDOC_GetAPI`, whose struct carries `StartFrameCapture`/`EndFrameCapture`, and those delimit a
+capture *explicitly*, so they need neither a keypress nor a present. prosper drives them itself:
+`PROSPER_RENDERDOC_AT_FRAME` (ordinal) and `PROSPER_RENDERDOC_AFTER_MS` (wall clock), plus
+`PROSPER_RENDERDOC_MAX_ITERS` for the span cap. `frontends/shared/diagnostics/renderdoc_capture.hpp`.
+
+### Aim it at the RENDERER's device. A capture of the wrong one looks perfectly healthy.
+
+**prosper runs two Vulkan instances.** prosper-app owns one for presentation (`main.cpp`); the live
+renderer owns another (`tests/fixtures/render_runner.h` — that header *is* the renderer) and
+publishes it through `prosper::gpu::set_shared_vulkan_context`. Every guest draw is on the
+renderer's. Passing NULL lets RenderDoc choose and **it chooses the swapchain's** — the presentation
+device. Measured on *The Messenger*, same build and route, only the aim differing:
+
+| aimed at | chunks | `vkCmdDrawIndexed` | `vkCmdDispatch` | pipelines |
+| --- | ---: | ---: | ---: | ---: |
+| NULL (RenderDoc picks) | 56 | **0** | **0** | **0** |
+| the renderer's instance | **406** | **9** | **2** | **6** |
+
+The 56-chunk capture is not corrupt and not truncated — it is a **complete, valid capture of the
+wrong device**, and it converts and opens perfectly while containing nothing anyone wants. It reads
+as a working instrument reporting a negative result, which is the dangerous shape. The trigger now
+aims itself and warns loudly when the renderer has not yet published a context.
+
+**A prosper frame is the GUEST's, not the app loop's.** A span of one app-loop iteration also yields
+zero draws: guest work is submitted off that thread, and the loop can spin at 260 fps while the guest
+renders far slower. The span closes on a guest present, and the log says which of the two closed it.
 
 **The Fedora package does NOT ship the Python bindings** — `rpm -ql renderdoc` gives `qrenderdoc` (the
-GUI) and `librenderdoc.so`, no `renderdoc.so` python module. So the scripted-analysis route is
-`convert`, not the Python API, and any recipe assuming `import renderdoc` will fail here.
-
-`renderdoccmd capture` triggers on a keypress rather than a frame ordinal, which makes it awkward
-headless; RGP's `MESA_VK_TRACE_FRAME` is the better automated capture.
+GUI) and `librenderdoc.so`, no `renderdoc.so` python module, and there is no `python3-renderdoc` to
+install. So *pixel history*, *per-draw render-target export* and *shader debugging* — the replay-side
+analysis, and the most valuable half — are **not** reachable from a script here without building
+RenderDoc from source. What is reachable headless is `convert` (`xml`, or `chrome.json` for a
+Perfetto-openable timeline), and handing the `.rdc` to a human with `qrenderdoc`.
 
 ## `radeontop` — the 60-second "is this even GPU-bound?" triage
 
--- a/prosper/frontends/prosper-app/main.cpp
+++ b/prosper/frontends/prosper-app/main.cpp
@@ -2031,8 +2031,18 @@ int main(int argc, char** argv) {
     void* renderdocDevicePointer = nullptr;
     uint64_t renderdocOpenedAtGuestPresent = 0;
     uint64_t renderdocOpenIterations = 0;
-    const uint64_t renderdocMaxIterations =
-        prosper::frontend::parse_capture_frame(getenv("PROSPER_RENDERDOC_MAX_ITERS")) ?: 2000;
+    // `?:` was a GNU extension and the repo's only use; spelled out so MSVC and -Wpedantic are clean.
+    // Note the contract differs from the two TRIGGERS deliberately: a malformed trigger disables
+    // itself, because firing at an unintended moment is worse than not firing. This is a safety CAP,
+    // not a trigger -- disabling it would mean an unbounded span -- so a malformed or zero value
+    // falls back to the default, and says so rather than doing it quietly.
+    const uint64_t renderdocMaxItersParsed =
+        prosper::frontend::parse_capture_frame(getenv("PROSPER_RENDERDOC_MAX_ITERS"));
+    const uint64_t renderdocMaxIterations = renderdocMaxItersParsed ? renderdocMaxItersParsed : 2000;
+    if (getenv("PROSPER_RENDERDOC_MAX_ITERS") && !renderdocMaxItersParsed)
+        std::fprintf(stderr, "[renderdoc] PROSPER_RENDERDOC_MAX_ITERS is not a usable count; using "
+                             "the default cap of %llu iterations\n",
+                     (unsigned long long)renderdocMaxIterations);
     const bool renderdocWanted = scheduledRdocFrame != 0 || automaticRdocAfter.delay_ns() != 0;
     if (renderdocWanted) {
         auto& rdoc = prosper::frontend::RenderDocCapture::instance();
@@ -2068,7 +2078,15 @@ int main(int argc, char** argv) {
     };
     auto arm_renderdoc_capture = [&](const char* why) {
         auto& rdoc = prosper::frontend::RenderDocCapture::instance();
-        if (!rdoc.available() || renderdocCaptureOpen) return;
+        if (!rdoc.available()) return;
+        if (renderdocCaptureOpen) {
+            // Both axes can come due inside one still-open span. Say so: the second trigger has been
+            // consumed and will not fire again, and a reader who set both deserves to know which one
+            // actually opened the capture they got.
+            std::fprintf(stderr, "[renderdoc] %s came due while a capture was already open; it is "
+                                 "consumed without opening a second one\n", why);
+            return;
+        }
         renderdocDevicePointer = renderdoc_device_pointer();
         if (!renderdocDevicePointer)
             std::fprintf(stderr, "[renderdoc] WARNING: the live renderer has not published a Vulkan "
@@ -3064,6 +3082,29 @@ int main(int argc, char** argv) {
         }
     }
 
+    // B1 (#3322 review): the span closes at the TOP of the loop, so every exit path -- --frames
+    // satisfied, SDL_EVENT_QUIT, or the elapsed trigger firing shortly before a title exits -- left
+    // an armed capture open and the artifact was lost with no line saying so. That is precisely the
+    // silent-instrument failure this trigger exists to prevent, so it is closed here too.
+    //
+    // Unlike its F8 sibling below, this CLOSES rather than cancels. RenderDoc decides for itself
+    // whether the span held anything: a span containing real work is written and worth having even
+    // though it ended at shutdown rather than on a guest present, and one containing nothing returns
+    // 0 and reports ABANDONED. Either way the run says what happened, which cancelling could not.
+    if (renderdocCaptureOpen) {
+        renderdocCaptureOpen = false;
+        const std::string rdocPath =
+            prosper::frontend::RenderDocCapture::instance().end(renderdocDevicePointer);
+        if (rdocPath.empty())
+            std::fprintf(stderr, "[renderdoc] capture ABANDONED at shutdown after %llu app-loop "
+                                 "iteration(s) -- RenderDoc recorded no API work in the span\n",
+                         (unsigned long long)renderdocOpenIterations);
+        else
+            std::fprintf(stderr, "[renderdoc] capture written: %s (closed by SHUTDOWN after %llu "
+                                 "app-loop iteration(s), NOT by a guest present -- it may hold only "
+                                 "part of a frame)\n",
+                         rdocPath.c_str(), (unsigned long long)renderdocOpenIterations);
+    }
     perfCapture.cancel();     // never publish a short/incomplete .prperf on a graceful early exit
     prosper_request_stop();   // signal the guest run-loop to wind down at its next boundary
     fprintf(stderr, "[app] shutting down after %llu presented frame(s)\n", (unsigned long long)shown);
--- a/prosper/frontends/prosper-app/main.cpp
+++ b/prosper/frontends/prosper-app/main.cpp
@@ -2039,10 +2039,15 @@ int main(int argc, char** argv) {
     const uint64_t renderdocMaxItersParsed =
         prosper::frontend::parse_capture_frame(getenv("PROSPER_RENDERDOC_MAX_ITERS"));
     const uint64_t renderdocMaxIterations = renderdocMaxItersParsed ? renderdocMaxItersParsed : 2000;
-    if (getenv("PROSPER_RENDERDOC_MAX_ITERS") && !renderdocMaxItersParsed)
-        std::fprintf(stderr, "[renderdoc] PROSPER_RENDERDOC_MAX_ITERS is not a usable count; using "
-                             "the default cap of %llu iterations\n",
-                     (unsigned long long)renderdocMaxIterations);
+    if (const char* mi = getenv("PROSPER_RENDERDOC_MAX_ITERS"); mi && !renderdocMaxItersParsed)
+        // Echo the offending value and the accepted shape, matching PROSPER_CAPTURE_BUNDLE_MAX_MB
+        // at main.cpp:1750 -- the repo's existing convention for a malformed BOUND, as opposed to a
+        // malformed trigger. "not usable" without the value leaves the author guessing which of
+        // their variables was wrong.
+        std::fprintf(stderr, "[renderdoc] ignoring malformed PROSPER_RENDERDOC_MAX_ITERS=\"%s\" "
+                             "(expected a positive iteration count); using the default cap of "
+                             "%llu\n",
+                     mi, (unsigned long long)renderdocMaxIterations);
     const bool renderdocWanted = scheduledRdocFrame != 0 || automaticRdocAfter.delay_ns() != 0;
     if (renderdocWanted) {
         auto& rdoc = prosper::frontend::RenderDocCapture::instance();
@@ -2288,13 +2293,27 @@ int main(int argc, char** argv) {
                                  "[renderdoc] capture ABANDONED after %llu iterations (closed by "
                                  "%s) -- RenderDoc recorded no API work in the span\n",
                                  (unsigned long long)renderdocOpenIterations, closedBy);
-                else
+                else {
+                    char note[512];
+                    std::snprintf(note, sizeof note,
+                                  "prosper capture. Closed by %s after %llu app-loop iteration(s) "
+                                  "and %llu guest present(s). Aimed at %s.%s",
+                                  closedBy, (unsigned long long)renderdocOpenIterations,
+                                  (unsigned long long)(guestNow - renderdocOpenedAtGuestPresent),
+                                  renderdocDevicePointer ? "the live renderer's Vulkan device"
+                                                         : "whichever device RenderDoc chose",
+                                  guestPresented ? ""
+                                                 : "  WARNING: closed by the iteration cap, not by "
+                                                   "a guest present -- this may hold only part of "
+                                                   "a frame, or no guest frame at all.");
+                    prosper::frontend::RenderDocCapture::instance().set_comments(rdocPath, note);
                     std::fprintf(stderr,
                                  "[renderdoc] capture written: %s (spanned %llu app-loop iterations "
                                  "and %llu guest presents, closed by %s)\n",
                                  rdocPath.c_str(), (unsigned long long)renderdocOpenIterations,
                                  (unsigned long long)(guestNow - renderdocOpenedAtGuestPresent),
                                  closedBy);
+                }
             }
         }
         // F8's only always-on work is this 4 Hz sample. `sample_due` is one atomic read, so the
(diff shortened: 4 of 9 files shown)


mattias800 (commented):
## REJECTED — one blocking finding (a ~5-line fix), everything else non-blocking

Independent review of head `25653d3`, read in a detached worktree. I did not rebuild or re-run CI.
I re-derived the central measurement from the artifacts rather than trusting the table, and I
checked the ABI and the `GetCapture` contract against RenderDoc's own header and published docs.

This is good work and the finding it reports is worth the PR. The one blocker is that the
instrument can still produce the exact shape the PR exists to prevent: a confident silence.

---

## BLOCKING

### B1. An open capture is never closed on any loop-exit path — the artifact is silently lost

`prosper/frontends/prosper-app/main.cpp:2256` opens the close at the **top** of the app loop.
Nothing closes it outside the loop: after `while (running && !prosper_stop_requested())` ends,
`main.cpp:3067` calls `perfCapture.cancel()` for the F8 sibling and there is no RenderDoc
equivalent, and on the normal booted-guest path the process ends in `std::_Exit` (`main.cpp:3127`), which
runs no destructors and no `atexit` handlers. So a capture opened on the final iteration is abandoned in RenderDoc's
buffers and never written.

Concrete scenario, using the workflow this PR's own docs recommend (locate an ordinal cheaply,
then aim at it) together with the documented CI/smoke bound:

```
PROSPER_RENDERDOC_AT_FRAME=900 ./prosper-app --frames 900 <DUMP_ROOT>/<TITLE_ID>-app0
```

1. `capture_frame_due(900, shown, false)` is `shown >= 899`, so `arm_renderdoc_capture` fires at
   `main.cpp:2716`, in the iteration that will present frame 900. stderr prints
   `[renderdoc] capture opened by PROSPER_RENDERDOC_AT_FRAME at guest present N`.
2. Later in that same iteration the present succeeds and
   `PresentedFrameCounter::record` (`present_policy.hpp:60-68`, via `main.cpp:2908`/`2968`/`3004`)
   sets `running = false` because `count_ >= exit_after_`.
3. The `while` condition is evaluated before the body, so the close block at `main.cpp:2256` never
   runs. Log ends with `[app] shutting down after 900 presented frame(s)`.

Result: no `.rdc`, no `capture written:` line, no `capture ABANDONED` line, no diagnosis. The
operator sees a capture that was announced as opened and then nothing at all.

Two more reachable variants of the same hole, neither requiring `--frames`:

* `SDL_EVENT_QUIT` / window close (`main.cpp:2329`) is polled *below* the close block; the loop
  then leaves through `if (!running) break;` at `main.cpp:2476`, with the previous iteration's
  capture still open if the guest has not presented since.
* `PROSPER_RENDERDOC_AFTER_MS` on a title that exits shortly after the trigger — which is exactly
  the "capture the moment it dies" case the wall-clock axis exists for. `prosper_stop_requested()`
  going true after the arm ends the loop at the condition, before the close.

The iteration cap bounds the *hang* case and is correct for that; it does not bound the *exit*
case, because the exit does not go through another iteration.

Suggested fix: mirror `perfCapture.cancel()`. After the loop, if `renderdocCaptureOpen`, either
call `end(renderdocDevicePointer)` and report the path with `closed by process exit`, or state out
loud that the capture was dropped. Either is fine; silence is not. Please also say in the log which
it was, per the folder's own rule below.

This is blocking rather than a nit only because of what the PR itself argues: `AGENTS.md` in the
new folder says *"make the wrong aim loud"*, and the PR body's thesis is that **a working
instrument reporting a confident negative is the dangerous shape**. A capture that is announced,
lost, and never mentioned again is that shape.

---

## Non-blocking

### N1. `?:` is a GNU extension and this is its only use in the repository (`main.cpp:2035`)

`grep -rnE '\) \?: |\w \?: ' prosper/` returns exactly one hit — this line. Measured on this box:
`g++ -std=c++20 -Wall -Wextra -pedantic` emits two diagnostics for the construct
(`ISO C++ does not allow '?:' with omitted middle operand`, `ISO C++ forbids omitting the middle
term`). Nothing in CI passes `-pedantic` to `prosper-app` and both CI compilers are GCC
(Linux + MinGW UCRT, `ci.yml:632-693`), so this does not break any build today. But
`CMakeLists.txt:3391` already branches on `if(NOT MSVC)`, and MSVC rejects `?:`. A plain
`const uint64_t n = parse(...); ... n ? n : 2000;` costs nothing.

### N2. A malformed `PROSPER_RENDERDOC_MAX_ITERS` is silently replaced by 2000 (`main.cpp:2035`)

I checked this against the contract the PR claims and it does **not** contradict it: `MAX_ITERS`
is a bound, not a trigger, both triggers really do disable on a malformed value
(`parse_capture_frame` → 0 → `capture_frame_due` false; `parse_performance_capture_delay_ns` → 0 →
`ElapsedPerformanceCaptureTrigger::take_if_due` returns false on `!delay_ns_`), and 2000 is the
safe direction. So the claim stands.

What is still worth fixing is that `PROSPER_RENDERDOC_MAX_ITERS=abc` and `=0` are both accepted in
silence. A reader who sets `0` meaning "no cap" gets 2000, sees `closed by iteration cap`, and has
no way to learn their value was rejected. One `fprintf` at parse time, next to the two that already
report the armed triggers.

### N3. Arming while a capture is open consumes the one-shot with no message (`main.cpp:2071`, `2716`)

`arm_renderdoc_capture` early-returns on `renderdocCaptureOpen`, but the caller has already set
`scheduledRdocArmed = true` (`main.cpp:2715`), and `take_if_due` has already consumed the elapsed
one-shot (`main.cpp:2320`). With both variables set and the two moments landing inside one open
span, the second capture is silently dropped. Rare, and one `fprintf` in the early-return path
closes it.

### N4. "prosper runs TWO Vulkan instances" is true of the measured path only

`renderdoc_capture.hpp:69-74` and `docs/GPU_PROFILING_EXTERNAL.md` state it flatly, and
`renderdoc_capture.hpp:72-73` goes further: *"Every guest draw and **dispatch** happens on the
renderer's."* Two qualifications I verified:

* `live_compute.cpp:2913` creates a **third**, private instance whenever adoption is declined — the
  shared device lacking `storage_image_read/write_without_format`, the queue family not supporting
  compute, or `create_pipeline_cache()` failing (`live_compute.cpp:2837-2905`). On that path guest
  *dispatches* run on an instance this capture never sees, and the capture would look draw-complete
  and dispatch-empty. Given the PR's whole subject, that is the caveat most worth writing down.
* Present unification (#1270, `try_adopt_shared_present`, `main.cpp:642`) has prosper-app present on
  the renderer's instance, collapsing the count to one.

Neither affects the default headless route that was measured. Suggest "two instances on the default
headless route; see …" plus one line on the compute fallback.

### N5. "it picks the one holding the swapchain" is n=1, and the docs say otherwise

`renderdoc_capture.hpp:73-74`, the doc section and the PR body all assert the choice
deterministically. RenderDoc's published in-application API docs say: *"Passing NULL for the device
pointer indicates that there is only one graphics API in use... If there are multiple APIs in use,
one will be picked at random and this may lead to unexpected or undefined behaviour."* The
practical guidance (always pass the renderer's instance) is right and is what the measurement
supports; the *mechanism* is an inference from one capture. Suggest "it picked the presentation
device here" rather than "it picks the one holding the swapchain" — the weaker claim is the one the
evidence carries, and it is equally motivating.

### N6. Small documentation inaccuracies

* `third_party/renderdoc/README.md:19` cites `frontends/shared/diagnostics/renderdoc_capture.{hpp,cpp}`.
  There is no `.cpp`; the class is header-only.
* The `pipelines` column (docs table and PR body) counts `vkCreateGraphicsPipelines` (6); the
  capture also contains 1 `vkCreateComputePipelines`, so 7 pipeline-creation calls total, next to a
  column that separately reports 2 dispatches.
* "Measured on *The Messenger*, same build and route, only the aim differing" — the aim difference
  was a source change (the `device_pointer` parameter did not exist for the 56-chunk run;
  `messenger2/run.log:243` correctly shows no `WARNING: the live renderer has not published…` line,
  which the current code would print for a NULL aim). It is a sound A/B across two revisions of this
  branch on one route, but it was not the same build, and this project is strict about that wording.

### N7. `BLOG.md` entry is a report wearing a blog entry's clothes

229 words and an eight-number table, against the file's stated median of ~60 and the charter's
explicit rule (*"If an entry has a table, a code block, or more than one or two numbers…"*). The
prose is genuinely good; the table belongs in the doc it already lives in, with a link.

### N8. `shared_vulkan_context()` is an unsynchronised read (pre-existing)

`gpu_executor.cpp:11438-11440` is a plain non-atomic global written by the renderer thread. The new
lambda at `main.cpp:2064-2068` reads it from the app-loop thread. `main.cpp:643` already does the
same, so this PR does not introduce the race, and a torn read here degrades to a bad
`device_pointer` rather than anything worse. Recording it because the new call site widens the
window.

---

## Verified correct — the areas most likely to be wrong, checked rather than assumed

**ABI / 1.4.1 subset — safe.** Every member called (`GetAPIVersion`, `SetCaptureFilePathTemplate`,
`GetNumCaptures`, `GetCapture`, `StartFrameCapture`, `IsFrameCapturing`, `EndFrameCapture`) sits
above the `// new function in 1.1.0` marker at `renderdoc_app.h:815`, i.e. all are in the 1.0.0
prefix of the struct. `SetCaptureFilePathTemplate` is the current name in a union with
`SetLogFilePathTemplate` at the same 1.0.0 slot (`renderdoc_app.h:777-786`). Nothing after
`DiscardFrameCapture` (1.4.0) is touched, so requesting `eRENDERDOC_API_Version_1_4_1` populates
every pointer used. No crash risk from the version pin.

**`GetCapture` — correct, including the `pop_back`.** RenderDoc's published docs settle the
ambiguity in the vendored header's comment: *"uint32_t* pathlength – is an optional parameter filled
with the byte length of the above filename **including the null terminator**"*, and *"It is advised
to call this function twice - first to obtain pathlength… Then again to actually retrieve the
path."* So `std::string path(pathlen, '\0')` (`renderdoc_capture.hpp:99`) is exactly the right size,
RenderDoc's fill stays in bounds, `path.back()` is the terminator, and the conditional `pop_back`
yields the exact path with no chopping. Non-const `std::string::data()` is C++17+; the project is
C++20 (`CMakeLists.txt:4`). The `pathlen == 0` guard makes `back()` safe.

**`RTLD_NOW | RTLD_NOLOAD` — right, and it is the documented idiom**, not just an efficiency
choice: RenderDoc's docs recommend exactly that pair for a passive "already loaded?" check. The
correctness argument in the header holds at this call site specifically — the singleton is first
constructed in `main` after both instances exist, so a fresh `dlopen` there could not have hooked
them. Matching works: the installed `librenderdoc.so` carries `SONAME librenderdoc.so`, so
`RTLD_NOLOAD` finds the copy the layer loaded by absolute path.

**Never `dlclose`ing is correct here, not a leak that matters.** One refcount on a library the
RenderDoc layer holds for process lifetime, in a process that ends via `std::_Exit`. It also
*avoids* rather than risks the `live_compute.cpp:11045-11056` hazard (#1704): `RenderDocCapture` is
a function-local static whose implicit destructor touches only a `std::string` and a raw pointer,
so it registers nothing that could run after the layer's own statics — and `_Exit` runs none of them
anyway. Worth one comment line saying so, since the header explains the `dlopen` and not the
missing `dlclose`.

**Span logic.** `renderdocDevicePointer` is stored at begin and reused at end, and nothing can
reassign it mid-span (`arm_renderdoc_capture` early-returns while open) — so a renderer publishing
its context between begin and end cannot cause a device mismatch. `gpu::present_count()` is
monotonic in `prosper-app`: the only reset is `present_reset()` (`videoout_present.cpp:190`), whose
callers are all tests. `hitCap` uses `++` outside any short-circuit, so the counter always advances.
The only hole is B1.

**Platform no-op.** I compiled `renderdoc_capture.hpp` standalone with
`g++ -std=c++20 -Wall -Wextra -pedantic` on both branches (the second by forcing the platform test
false): **zero warnings on each**. `reason_` is set, every member body still references its
parameter so nothing is unused, and `RENDERDOC_DEVICEPOINTER_FROM_VKINSTANCE`
(`renderdoc_app.h:505`) is defined unconditionally, so `main.cpp:2067` compiles on Windows too.

**Triggers are genuinely opt-in and fail-closed.** Unset or malformed → 0 → disabled on both axes;
`ElapsedPerformanceCaptureTrigger` never fires on a zero delay. `RenderDocCapture::instance()` is
only reached when one is set, so a default run does not `dlopen` anything.

**The measurement — re-counted from the artifacts, not taken on trust.** From
`~/rdoc/messenger2/frame.xml` and `~/rdoc/messenger3/frame.xml`, every number in the PR table
matches exactly:

| | chunks | `vkCmdDraw` | `vkCmdDrawIndexed` | `vkCmdDispatch` | `vkCreateGraphicsPipelines` | `vkCreateShaderModule` |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| NULL aim | 56 | 0 | 0 | 0 | 0 | 0 |
| renderer's instance | 406 | 2 | 9 | 2 | 6 | 13 |

The 56-chunk capture's entire command content is `vkCmdPipelineBarrier` ×8,
`vkCmdCopyBufferToImage` ×2, `vkCmdBlitImage` ×2 — presentation, as claimed. `BLOG.md`'s "draws:
11" and the doc's "`vkCmdDrawIndexed`: 9" are both right under their own column headers (2 + 9).

**The two-instance claim's core is verified.** `grep -rn set_shared_vulkan_context` over the tree:
the sole caller is `tests/fixtures/render_runner.h:1599`, which is the live renderer; every other
hit is a read (`live_compute.cpp:2837`, `gpu_executor.cpp:8224`, `gpu_replay.cpp:2696`,
`main.cpp:643`) or a test constructing the struct. `vkCreateInstance` outside tests/tools:
`main.cpp:249`, `live_compute.cpp:2913`, `render_runner.h:1213`/`1217`. See N4 for the third one.

---

Re-review on request once B1 is addressed; nothing else here needs to gate the merge.

mattias800:
## Author response — all findings addressed, head `f6e98298`

### B1 (blocking) — fixed, and the finding was exactly right

The span closed at the **top** of the loop and nothing closed it outside, so every exit path lost the
artifact silently. I had verified the trigger on runs I killed externally, which is the one shape that
never exercises a graceful exit — so my own testing could not have found this. The three named paths
(`--frames` satisfied via `PresentedFrameCounter::record` clearing `running` in the arming iteration,
`SDL_EVENT_QUIT` via the early `break`, and `AFTER_MS` firing shortly before a title exits) all
reproduce it, and the third is the wall-clock axis's whole purpose.

It is fixed by **closing** rather than cancelling, which differs from the F8 sibling two lines below
it, deliberately: RenderDoc itself decides whether the span held anything. Real work is written and
worth having even though it ended at shutdown — reported as `closed by SHUTDOWN ... may hold only part
of a frame` so nobody mistakes it for a guest frame — and an empty span returns 0 and reports
`ABANDONED`. Cancelling could report neither, and "no file and no line" is the failure being fixed.

You noted this reproduces the shape the PR argues against and that the folder's own new `AGENTS.md`
says to make the wrong aim loud. Both true, and that is the right reason to call it blocking.

### Non-blocking — all eight taken

| finding | resolution |
| --- | --- |
| `?:` GNU extension, repo's only use, 2 `-Wpedantic` warnings, MSVC would reject | spelled out |
| malformed/zero `MAX_ITERS` silently becomes 2000 | still 2000, but **says so** — and the differing contract is now stated rather than implied |
| second arm inside an open span consumes its one-shot silently | now reports it |
| "prosper runs TWO Vulkan instances" stated as invariant | qualified in header and doc — third private instance on the declined-adoption path, #1270 can collapse to one |
| "it picks the one holding the swapchain", n=1 vs docs' "at random" | reworded: the rule is **always pass the device**, not "it picks the swapchain" |
| "same build … only the aim differing" | it was two builds one edit apart, identical span logic; reworded to what was done |
| pipelines column omitted 1 compute pipeline | split into 6 graphics + 1 compute |
| README cites a nonexistent `.cpp` | header-only |
| `BLOG.md` 229 words + 8-number table vs charter rule and ~60-word median | **80 words, no table**, links #3321 |

On `MAX_ITERS`: I kept the fallback rather than making it disable, because the contracts genuinely
differ and conflating them would be worse. A malformed **trigger** disables itself — firing at an
unintended moment is worse than not firing. `MAX_ITERS` is a safety **cap**; disabling it means an
unbounded span. So it falls back and announces the fallback. If you read that as still contradicting
the doc's blanket "a malformed value disables its own trigger", say so and I will reword the doc
rather than the code.

### On the unsynchronised-global race

Correct that the new `shared_vulkan_context()` read widens a pre-existing race on `g_shared_vulkan`
(`gpu_executor.cpp:11439-11440`, a plain global with no synchronisation). I have **not** fixed it here:
it predates this PR, the fix belongs with the global rather than with one more reader, and it is worth
its own issue. Flagging that I am deliberately leaving it rather than having missed it.

### Thank you for re-counting the evidence

You re-derived all ten numbers from `frame.xml` rather than trusting the table, and confirmed the ABI
subset against `renderdoc_app.h:815` and `GetCapture`'s two-call contract against the published docs —
the two things I most wanted checked, because I asserted them from a header rather than from the spec.

mattias800 (commented):
## APPROVED — head `f6e98298`

Re-reviewed the fix commit in a fresh detached worktree. B1 is closed correctly, the eight
non-blocking findings are all addressed, and one of the rewordings is more right than I gave it
credit for. Nothing below blocks the merge. Taking your three questions in order.

---

## 1. Is the exit-path close correct and complete?

**On every path reachable from `main`, yes — I enumerated rather than eyeballed.**

The loop body is `main.cpp:2268-3083`. Grepping that exact range for `return`, `std::exit`, `_Exit`,
`quick_exit`, `abort`, `std::terminate` and `SDL_Quit` returns **zero hits**, so every departure from
the loop is either a `break` or the `while` condition — and both land on your new block at
`main.cpp:3085-3105`, which sits before `perfCapture.cancel()` (`:3106`), before
`prosper_request_stop()` (`:3107`), and well before `std::_Exit(exitCode)` (`:3149`). `VKCHECK`
(`main.cpp:152`) `return false`s rather than exiting and is not used inside the loop (0 occurrences
in range), so it cannot bypass it either. The guest-driven stop goes through the loop condition too:
`hle_service.cpp:3594` calls `prosper::prosper_request_stop()`, which only sets the flag
(`lifecycle.cpp:16`). All three scenarios I gave — `--frames` satisfied, `SDL_EVENT_QUIT` via the
`main.cpp:2476` break, elapsed trigger before a title exits — now close and report.

**One residual you did not name, and it is the honest answer to "any path I have not thought of":**

`prosper/src/hle/kernel/hle_kernel_time.cpp:878` and `:888`.

```
HLE(k_exit) { ... _Exit((int)a0); }                              // guest _exit(status)
HLE(k_debug_raise_release) { ... _Exit(0x66); }                  // UE4's shipping check() failure
```

Both run **on the guest thread while `main` is still inside the app loop**, and `_Exit` is
`exit_group` — `main` never reaches line 3085, and no `atexit` hook would help either. So a title
that quits through libc, or a UE4 title that trips a shipping `check()`, still loses an open capture
silently. The second one stings a little because "capture what it was doing as it died" is precisely
the wall-clock axis's use case.

**Not blocking, and I do not think it should be fixed in this PR** — it is cross-layer and the seam
does not exist yet. But note both sites already run the same shape of work you would need
(`host::guest_dmem_write_trace_report(); fflush(nullptr);` before the `_Exit`), and `main.cpp:29`'s
include comment already names that pattern as *"flush dmem writer diagnostic before deliberate
_Exit"*. A registered deliberate-exit hook list is the natural home. Suggest one sentence in the new
block saying the close covers every path **out of the app loop**, and that a guest-driven `_Exit`
bypasses it — plus a follow-up issue. A stated limit is a different thing from a silent one.

### Close rather than cancel — I agree with you, and here is the mitigation for the trap you asked me to challenge

Your reasoning is right and I would not reverse it. `EndFrameCapture` returning 0 means RenderDoc
recorded nothing, so an empty span cannot produce a misleading file; a span holding real work is
often the only evidence a dying title will ever yield; and cancelling reports neither. Loud-and-partial
over silent-and-absent is the correct trade for a diagnostic.

**The trap is nevertheless real, and it is one this project already has a name for.** A
shutdown-truncated `.rdc` opens perfectly in `qrenderdoc` and looks like a frame — it is the *same
shape* as the wrong-device capture this PR exists to warn about, one step over: a valid capture of
part of a frame reads as a whole frame with draws missing, which is exactly how somebody concludes
"the renderer is dropping draws". Your `closed by SHUTDOWN … may hold only part of a frame` line is
the right warning **in the wrong place**: it lives in stderr, and the `.rdc` is the artifact that
gets handed to a human days later without the log.

**Concrete, small, and inside your existing version pin:** `SetCaptureFileComments` is at
`renderdoc_app.h:819`, under the `// new function in 1.2.0` marker — comfortably below the 1.4.1 you
request, so the pointer is populated. Its contract (`renderdoc_app.h:439-448`) is *"Sets the comments
associated with a capture file. These comments are displayed in the UI program when opening"*, and
*"If set to NULL or `""` the most recent capture file created made will be used instead."* So after
`end()` hands you the path, one call stamps the shutdown warning **into the file**, where the person
opening it will read it. Worth doing for the cap-closed case in the loop too.

Non-blocking; fine as a follow-up.

---

## 2. `MAX_ITERS` — your reasoning is right, and nothing needs rewording. Do not change the doc.

I checked every place the claim is made, and **all of them are already scoped to triggers**; there is
no blanket statement to contradict:

* `main.cpp:1996` — "All are one-shot and opt-in, and a malformed value disables its own **trigger**…"
* `main.cpp:2017` — same, introducing the two trigger variables.
* `AGENTS.md:29` — "prefer a switch whose malformed value disables its own **trigger**…" (a
  preference about triggers, and `MAX_ITERS` is not one).
* PR body — "Both **triggers** are one-shot and opt-in, and a malformed value disables its own trigger…"
* `GPU_PROFILING_EXTERNAL.md:89` only calls it "the span cap", which is accurate.

**Better than that: the repo already had your precedent and it is the same class of switch.**
`main.cpp:1750-1760` parses `PROSPER_CAPTURE_BUNDLE_MAX_MB` — also a *bound*, not a trigger — and on
a malformed value falls back to the default and announces it:

```
[grab] ignoring malformed PROSPER_CAPTURE_BUNDLE_MAX_MB="%s" (expected 64..3072)
```

So the distinction you drew is not a special case invented for this PR; it is the convention. Keeping
the code uniform would have been the wrong move. I withdraw N2 entirely — announcing the fallback was
all it needed.

**One-line nit for consistency with that precedent:** the `MAX_MB` message echoes the offending value
and the accepted range; `main.cpp:2043` does neither. `PROSPER_RENDERDOC_MAX_ITERS="200O" is not a
usable count (expected a positive decimal); using the default cap of 2000 iterations` costs nothing
and saves the reader a squint at their own shell history.

---

## 3. The rewordings — checked each. All accurate, one better than I credited, two follow-ups.

**`GPU_PROFILING_EXTERNAL.md:93-99` and `renderdoc_capture.hpp:69-80` — correct, and more defensible
than my original finding implied.** I said present unification "can collapse it to one"; I had read
`main.cpp:641`'s comment. Checking the actual policy: `request_gpu_present(nullptr, false, true)`
returns **true** (`present_policy.hpp:21-24`), so adoption is **opt-out**, attempted on *every* real
game boot, and `PROSPER_APP_GPU_PRESENT=0` is the disable. Which means the collapse to one instance
is the **expected** case on an ordinary windowed run with a game, and two is specifically the
offscreen/headless count — exactly what your new opening clause, *"On the default headless route
prosper runs two Vulkan instances"*, says. The qualifier is precisely right, not merely safe. And it
is functionally harmless either way: when unification succeeds there is one instance and
`renderdoc_device_pointer()` returns it, so the aim stays correct.

*Drive-by, pre-existing, not yours:* `main.cpp:641` says *"Called only when PROSPER_APP_GPU_PRESENT is
set and a game is booting"*, which contradicts `present_policy.hpp:21-24` — the variable is a
disable, not an enable. Probably what steered the original wording. Worth a one-word fix while you
are in the file, or its own issue.

**`renderdoc_capture.hpp:73-80` on the NULL pick — accurate.** *"RenderDoc documents that choice as
arbitrary; here it took the one holding the swapchain"* matches the published text (*"one will be
picked at random"*), keeps the n=1 observation as an observation, and lands the rule on "pass the
device always". Not under-claimed: the operational instruction is unchanged and the justification is
now the durable one.

**The comparison table — accurate, and I verified the new claim rather than the old one.**
*"same route and identical span logic, differing only in which device the capture is aimed at (two
builds, one edit apart)"*: `messenger2/run.log:244` reads `capture written: … (spanned 4 app-loop
iterations and 3 guest presents, closed by guest present)` — same span mechanism, same log format as
`messenger3/run.log:244`. So "identical span logic" is not just plausible, it is visible in the
artifact. Good.

**Pipeline split — re-counted, correct.** `messenger3/frame.xml`: 6 `vkCreateGraphicsPipelines`, 1
`vkCreateComputePipelines`; `messenger2/frame.xml`: 0 and 0. The new two-column table matches. (It
drops the `vkCmdDraw` column the PR body has; that is less detail, not an inaccuracy.)

**`third_party/renderdoc/README.md:18` — fixed.** Now `renderdoc_capture.hpp` (header-only). Correct.

**`BLOG.md:24-28` — 98 words, no table, links #3321.** Fits the charter's shape. Two very small
under-claims, mentioned only because you asked: it says a capture *"aims at a frame ordinal"* where
there are two axes, and *"prosper runs two Vulkan devices"* without the headless qualifier you added
everywhere else. Both are fine in a blog entry that links the detail; take or leave.

### New finding: the PR body still carries every claim you corrected

Non-blocking but should land before merge, because the body is the artifact a reader hits first and
the charter asks for it to be self-contained. Currently:

* heading — "prosper runs TWO Vulkan instances" (unqualified, as an invariant)
* "Passing NULL lets RenderDoc choose the device, and **it chooses the one holding the swapchain.**"
* "Measured on *The Messenger*, **same build**, same route, the only difference being the aim"
* the table's `pipelines` column still reads `6`, un-split
* the verification section repeats "two captures, **same build** and route"

Also worth a short paragraph on the exit-path close, since it is the substantive change in
`f6e98298`. Note `gh pr edit` fails against this repo (Projects-classic GraphQL error) and silently
does not update — use `gh api -X PATCH repos/mattias800/prosper/pulls/3322 -f body=…` and read the
body back to confirm.

---

## `g_shared_vulkan` — agreed, and it does not block

I raised it as pre-existing and I still read it that way. `gpu_executor.cpp:11438-11440` is the
defect's home; `main.cpp:643` was already reading it from the app thread before this PR, and a torn
read at the new site degrades to a bad `device_pointer`, not to anything that can corrupt a result.
Your call to give it its own issue against the global is the right one — please do file it, so the
reasoning does not live only in this thread.

---

## Verdict

**APPROVED** on head `f6e98298`. B1 is properly closed for every path out of the app loop; the
close-over-cancel choice is right; the `MAX_ITERS` distinction was the convention already and I
withdraw that finding; the rewordings are accurate and one is sharper than my original note. Merge
when your own verification and CI are green. Please fold in before or shortly after merge, none of
them gating:

1. the PR body corrections (a few minutes, and it is the record),
2. one sentence in the new block naming the guest-`_Exit` limit, plus a follow-up issue,
3. `SetCaptureFileComments` so the shutdown warning rides inside the `.rdc`,
4. the `g_shared_vulkan` issue,
5. optionally `main.cpp:641`'s stale comment and the `MAX_ITERS` message echo.

Thanks for taking B1 seriously rather than papering it — and the note about only ever having tested
with externally-killed runs is exactly the kind of thing worth keeping in the PR body: the test shape
that could not observe the bug is as useful to the next reader as the bug.

mattias800 (commented):
## APPROVED — head `6a329668`. Delta is clean; the approval carries.

Re-verified the three-item delta `f6e98298..6a329668` in a fresh worktree. Nothing blocking. Detail
on (1), since you asked for it to be treated as blocking if wrong, and it is not.

### 1. `SetCaptureFileComments` — correct, and I checked rather than read

**ABI slot: safe, verified in the vendored header.** `renderdoc_app.h:818-819` puts
`SetCaptureFileComments` under `// new function in 1.2.0`, and the next marker down is
`// new function in 1.4.0  DiscardFrameCapture` (`:821-822`). So it sits *strictly inside* the 1.4.1
subset you pin — two versions below the boundary, not at it. Requesting
`eRENDERDOC_API_Version_1_4_1` populates it. Same check I ran on the original seven, same answer.

**Call contract: the documented usage.** `renderdoc_app.h:438-447` — *"filePath should be a path to
the capture file to add comments to… comments should be a NULL-terminated UTF-8 string… Any existing
comments will be overwritten."* You call it after `end()` with the path `GetCapture` returned, which
is reached only when `EndFrameCapture` returned non-zero and `pathlen != 0`
(`renderdoc_capture.hpp:94-102`), so the file exists by then. `set_comments`
(`renderdoc_capture.hpp:73-76`) guards on both `!api_` and `path.empty()`, so the no-op platform
branch and the abandoned-capture case can't reach the pointer. Both call sites are inside the
`else` of `if (rdocPath.empty())`, so no stamp is ever attempted without a file.

**`snprintf` into `char note[512]`: it cannot even truncate, let alone do worse.** I measured rather
than eyeballed — compiled both format strings with `UINT64_MAX` in every `%llu` and the longest
branch of every ternary:

```
loop note worst case     = 302 bytes (buffer 512, truncated=no)
shutdown note worst case = 257 bytes (buffer 512, truncated=no)
```

So truncation is unreachable, not merely harmless. And `std::snprintf` bounds the write and always
NUL-terminates within the buffer, so the failure mode if the strings ever grew is a short string —
never an overflow, never an unterminated read. Format/argument counts check out on both (5 and 1).

*One defensive nit, since you are stamping authoritative provenance:* in the loop note the WARNING
clause is the **last** field, so if those strings ever grow past the margin, truncation would eat
exactly the safety-critical part and leave a confident-looking stamp behind. Cheap insurance: move
the warning to the front of the note, or check `snprintf`'s return against `sizeof note` and log if
it ever exceeds. Not needed today at 302/512.

*Second nit, and the one I would actually take:* **the shutdown note omits the aim.** The loop note
carries `Aimed at the live renderer's Vulkan device` / `whichever device RenderDoc chose`
(`main.cpp:2301-2304`); the shutdown note (`main.cpp:3135-3141`) does not. That is the wrong way
round — the truncated capture is the one most likely to be misread, and "which device" is this PR's
central provenance fact. One ternary away, and it would make your description of the feature
("stamped with … which device it was aimed at") true of every capture rather than most.

**Compiles clean.** `g++ -std=c++20 -Wall -Wextra -pedantic` on `renderdoc_capture.hpp` standalone,
both platform branches (the second by forcing the platform test false), exercising
`set_comments`/`begin`/`end`/`capturing`: zero warnings on each.

### 2. `MAX_ITERS` echo — matches the precedent exactly

`main.cpp:2042-2050` now reads `ignoring malformed PROSPER_RENDERDOC_MAX_ITERS="%s" (expected a
positive iteration count); using the default cap of %llu`, which is the shape of
`PROSPER_CAPTURE_BUNDLE_MAX_MB` at `main.cpp:1758` — same verb, same echo of the offending value,
same statement of the accepted shape. The `if (const char* mi = getenv(…); …)` init-statement is
C++17 and the project is C++20 (`CMakeLists.txt:4`). Agreed on leaving the doc alone: every claim in
the tree is already scoped to *triggers*, so there was nothing to reword.

### 3. `_Exit` limit and the two issues — accurate

`main.cpp:3109-3115` states the limit correctly, including the part that makes it a limit rather than
an oversight: `_Exit` is `exit_group`, so an `atexit` hook would not help either, and both sites
already run a flush step so a registered finalizer list is the natural home. #3324 and #3325 are open
and their titles match what they describe. Not inventing the seam here was the right call.

### PR body — one residual

The corrections landed, and the opt-out point reads well where you put it. One line was missed:
the **Verification** section still says *"two captures, **same build** and route, differing only in
aim"* — the second occurrence of the claim the table caption 60 lines above now correctly words as
"two builds, one edit apart". Same `gh api -X PATCH` you already used. Non-blocking; the corrected
statement is elsewhere in the same body.

---

**Verdict: APPROVED on `6a329668`.** Merge when your own verification and CI are green. The two
optional follow-ups, neither gating: the aim field in the shutdown stamp, and the one stale
"same build" in the Verification section.

Worth recording somewhere durable, since it is the transferable half of this review: the reason
`SetCaptureFileComments` was worth the extra round is that **stderr does not travel with the
artifact**, and this project hands `.rdc` files to humans days later. That generalises past
RenderDoc — any diagnostic whose output outlives its run wants its caveats inside the file.

PR #3326, opened 2026-09-04
diag: let a RenderDoc capture span several guest presents (#3321)

mattias800:
Follow-up to #3322 (which added the triggers). Refs #3321, #3126.

## The default was a guess, and Stray showed it was wrong

#3322 closed the capture span on the **first** guest present. That is right for a title whose frame
is self-contained, and it is wrong for Stray:

- the capture contains a fully rendered 3D **cat** and **no menu**;
- the presented screenshot at the same moment shows a **menu and no cat** (`nonblack=0.0087`, three
  pixel-identical frames — the title screen holding).

A capture that holds half the screen cannot answer where the other half is lost, which is the whole
question in #3126. `PROSPER_RENDERDOC_PRESENTS=N` widens the span. **Default stays 1** — unchanged
behaviour for every existing user of the trigger.

## The measurement, and it contradicts the hypothesis that motivated the change

Stray, `PROSPER_RENDERDOC_PRESENTS=8`, title screen at 175 s:

| | 1 present | 3 presents | ratio |
| --- | ---: | ---: | ---: |
| total chunks | 7,388 | 19,676 | 2.66x |
| `vkCmdBeginRenderPass` | 65 | 192 | **2.95x** |
| `vkCmdDrawIndexed` | 98 | 312 | **3.18x** |
| `vkCmdDispatch` | 89 | 275 | **3.09x** |
| `vkCmdCopyImageToBuffer` | 52 | 168 | 3.23x |

**The ratio table above is NOT what establishes this, and the review was right to say so.** Under an
alternating `cat, UI, cat` sequence the aggregate counts would *also* land near 3x — so the ratio
cannot distinguish `A,B,A` from `A,A,A`. It is a **void discriminator, not a negative one**, and the
conclusion below was reached by a route that does not support it. (It is also not literally true that
"every count scales": roughly a third of the 60 chunk names do not, because RenderDoc serialises some
resources per *capture* rather than per frame — `vkCreateImage` 1.41x, `vkCreateShaderModule` 1.16x,
`vkCreateComputePipelines` 1.05x.)

**The measurement that does establish it**, from the same file: segment the capture by per-render-pass
draw signature. The 192 passes decompose as a 10-pass prefix, then **two complete 66-pass frames that
are identical at all 66 positions — zero differences** — then a 50-pass cap-truncated tail. A full
chunk-name diff of those two frames is 3,905 against 3,900, differing *only* in streaming traffic,
with every draw, dispatch, pass and submit count identical.

So **the frames really are structurally identical**, there is no alternating cat-frame/UI-frame, and
**the multi-frame explanation for the missing menu is withdrawn** — it was the reason I built this,
and the first thing the wider capture did was falsify it. The menu is somewhere in a frame that also
draws the cat; the pipeline map posted to #3126 says the likeliest place is pass 61, a 1920x1080 pass
carrying 10 draws while its neighbours carry 1.

**And a stronger argument for this PR that the original body missed**: the 1-present capture satisfies
`p1 == frame[1:]` exactly, at all 65 positions — so the `N=1` default captures one whole frame *minus
its leading render pass*, which is a real defect in the default rather than merely a limitation.
Confirmed against a frame body taken from a different run.

The widened span is what made any of this measurable.

## Contract

Same fallback-and-announce as `PROSPER_RENDERDOC_MAX_ITERS`, and for the same stated reason: this
**bounds** a span rather than **aiming** one, so a malformed value falls back to the default and says
so — matching `PROSPER_CAPTURE_BUNDLE_MAX_MB` (`main.cpp:1750`), which #3322's review established as
the repo's convention for a malformed bound. Only the two *triggers* disable themselves on a
malformed value, because firing at an unintended moment is worse than not firing.

## The interaction that bit on the first run, and is reported rather than hidden

At `N=8` the **iteration cap closed the span first**, at 3 presents — and the capture was correctly
logged and stamped as `closed by iteration cap` rather than passing as a complete frame. That is
#3322's provenance machinery doing its job on its first real encounter with a truncated span. **A slow
title needs `PROSPER_RENDERDOC_MAX_ITERS` raised alongside `PROSPER_RENDERDOC_PRESENTS`**; the log
line says which of the two closed the span, so the reader is told rather than left to infer it.

## Affected / unaffected

- **Affected:** `prosper-app` only, and only when `PROSPER_RENDERDOC_PRESENTS` is set. The close
  predicate changes from `guestNow > opened` to `guestNow - opened >= presents`, which is identical at
  the default of 1.
- **Unaffected:** every other capture path; all platforms where the trigger already compiles to a
  no-op.
- **Risk:** the close predicate is exactly where #3322's review found a real blocking bug (an open
  capture never closed on exit), so this is worth a second reader despite being 18 lines.

## Verification

```
cmake --build prosper/build-linux --target prosper-app -j6     # clean, no warnings
```

Functional evidence is the table above — one Stray run at `N=8`, converted with
`renderdoccmd convert -c xml` and counted, against the `N=1` capture from #3322.

### DIFF
--- a/prosper/frontends/prosper-app/main.cpp
+++ b/prosper/frontends/prosper-app/main.cpp
@@ -2036,6 +2036,19 @@ int main(int argc, char** argv) {
     // itself, because firing at an unintended moment is worse than not firing. This is a safety CAP,
     // not a trigger -- disabling it would mean an unbounded span -- so a malformed or zero value
     // falls back to the default, and says so rather than doing it quietly.
+    // How many GUEST presents the span covers, default 1 (#3321). One is right for a title whose
+    // frame is self-contained, and demonstrably wrong for one that builds its screen across several
+    // frames: Stray's title screen renders its 3D element in a frame that carries no UI, so a
+    // one-present capture holds a cat and no menu while the screen shows a menu and no cat. Widening
+    // the span is the only way to see a whole cycle. Same fallback-and-announce contract as the cap
+    // below -- this bounds a span, it does not aim one.
+    const uint64_t renderdocPresentsParsed =
+        prosper::frontend::parse_capture_frame(getenv("PROSPER_RENDERDOC_PRESENTS"));
+    const uint64_t renderdocPresents = renderdocPresentsParsed ? renderdocPresentsParsed : 1;
+    if (const char* np = getenv("PROSPER_RENDERDOC_PRESENTS"); np && !renderdocPresentsParsed)
+        std::fprintf(stderr, "[renderdoc] ignoring malformed PROSPER_RENDERDOC_PRESENTS=\"%s\" "
+                             "(expected a positive present count); spanning %llu\n",
+                     np, (unsigned long long)renderdocPresents);
     const uint64_t renderdocMaxItersParsed =
         prosper::frontend::parse_capture_frame(getenv("PROSPER_RENDERDOC_MAX_ITERS"));
     const uint64_t renderdocMaxIterations = renderdocMaxItersParsed ? renderdocMaxItersParsed : 2000;
@@ -2060,6 +2073,9 @@ int main(int argc, char** argv) {
         } else {
             // Never leave this at RenderDoc's default, which is under /tmp -- see the header.
             rdoc.set_path_template(grabDir + "/prosper_frame");
+            if (renderdocPresents != 1)
+                std::fprintf(stderr, "[renderdoc] span set to %llu guest presents\n",
+                             (unsigned long long)renderdocPresents);
             if (scheduledRdocFrame)
                 std::fprintf(stderr, "[renderdoc] capture scheduled at host frame %llu (into %s)\n",
                              (unsigned long long)scheduledRdocFrame, grabDir.c_str());
@@ -2278,7 +2294,8 @@ int main(int argc, char** argv) {
         // by a missing file.
         if (renderdocCaptureOpen) {
             const uint64_t guestNow = gpu::present_count();
-            const bool guestPresented = guestNow > renderdocOpenedAtGuestPresent;
+            const bool guestPresented =
+                guestNow - renderdocOpenedAtGuestPresent >= renderdocPresents;
             const bool hitCap = ++renderdocOpenIterations >= renderdocMaxIterations;
             if (guestPresented || hitCap) {
                 renderdocCaptureOpen = false;


mattias800 (commented):
**APPROVED** — the 18 lines of code are correct, and the close predicate in particular holds up under
the scrutiny it deserves given what #3322's review found here. Everything below is non-blocking, but
finding 2 is substantive: **the PR body's central measurement does not test the hypothesis it
withdraws**, and the same two artifacts contain a measurement that does. The withdrawal turns out to
be right — I am asking for the derivation to be replaced, not the conclusion.

## The close predicate — verified, no blocking issue

`main.cpp:2296-2299`, `guestNow - renderdocOpenedAtGuestPresent >= renderdocPresents`.

**No unsigned underflow, and the reason is stronger than "it shouldn't happen".**
`gpu::present_count()` reads `g_present_count`, which has exactly two writers:
`fetch_add(1)` at `src/gpu/present/videoout_present.cpp:75`, and `store(0)` in `present_reset()` at
`:190-192`. I grepped every `present_reset()` call site in the tree — **11, all under
`prosper/tests/`**, none in `src/`, `frontends/` or `tools/`. Its own declaration
(`videoout_present.hpp:159`) says `// tests: clear front/count`. So in `prosper-app` the counter is
strictly non-decreasing. Both reads that feed the subtraction are on the app-loop thread
(`:2118` inside `arm_renderdoc_capture`, called only from `:2376` and `:2771`, both directly in the
`while` body at `:2289`; and `:2296`), so read-read coherence on a single atomic gives
`guestNow >= renderdocOpenedAtGuestPresent` even at `memory_order_relaxed`. 64-bit wrap is not
reachable. **The subtraction is safe.**

**Exactly equivalent at the default.** Given the above, `a > b` and `a - b >= 1` are the same
predicate. Existing users of `PROSPER_RENDERDOC_AT_FRAME` / `_AFTER_MS` see no behaviour change.

**The cap still bounds, and `closedBy` reports the right one.** `++renderdocOpenIterations >= max`
is its own statement at `:2299`, not folded into the `||` at `:2300`, so the counter increments on
every pass regardless of `guestPresented` — no short-circuit bug. `closedBy` at `:2307` is driven by
`guestPresented`, so a span that reached its present count is labelled "guest present" even if the
cap also came due on that pass, which is the correct precedence. Confirmed against your own artifact
rather than by reading: the `N=8` run's log line reads
`spanned 2000 app-loop iterations and 3 guest presents, closed by iteration cap`, and the `.rdc`'s
embedded `renderdoc/ui/notes` section carries
`TRUNCATED -- closed by the iteration cap, not by a guest present`. The provenance machinery does
what it claims.

**The exit-path close from #3322 is untouched and still reachable.** I checked for early exits inside
the loop body — `main.cpp:2289`-`:3121` contains no `return`, `exit`, `_Exit` or `abort`, so every
loop exit falls through to `:3139`. A span that never reaches `N` presents is still closed there and
still stamped `TRUNCATED`. With `N>1` that path becomes *more* likely to be exercised, which is worth
noting but is not a defect.

**Contract question answered:** `parse_capture_frame` (`capture_schedule.hpp:12-19`) returns 0 for
null, empty, trailing text, negative, overflow **and** for a literal `"0"` — so the
`? :` fallback to 1 is correct, and yes, the message at `:2049-2051` echoes the offending value, like
`PROSPER_CAPTURE_BUNDLE_MAX_MB` at `:1750`. Consistent with `MAX_ITERS`.

`prosper/tools/env/check_diag_gates.py` is clean on this head (108 findings, 108 baselined), so no new
split gate.

---

## 1. Comment adjacency — the new declaration lands under the wrong paragraph (non-blocking, please fix)

`main.cpp:2034-2047`. The paragraph ending

> This is a safety CAP, not a trigger -- disabling it would mean an unbounded span -- so a malformed
> or zero value falls back to the default

was written about `renderdocMaxIterations`. The new `PRESENTS` comment and declaration are inserted
**between** it and its subject, so the "safety CAP / unbounded span" sentence now sits immediately
above `renderdocPresentsParsed` and reads as describing it. It does not: disabling `PRESENTS`
*narrows* the span to one present, the opposite of unbounded. Move the new block above that
paragraph, or move the paragraph down to `:2052`.

## 2. The measurement — the conclusion is right, the evidence given does not reach it

I re-derived both columns from `~/rdoc/stray-title/frame.xml` and `~/rdoc/stray-8/frame.xml`. **Your
five table rows all reproduce exactly** (65/192, 98/312, 89/275, 52/168). Total chunks are 7,387 and
19,675 rather than 7,388 and 19,676 — `grep -c '<chunk'` also matches the `<chunks version="32">`
container line. That changes nothing (2.663x either way); I mention it only so the number is right if
it gets quoted again.

**(a) "Every count scales by almost exactly the present count" is not what the artifact says.** There
are 60 distinct chunk names across the two captures. Roughly a third do not scale at all:

| | 1p | 3p | ratio |
| --- | ---: | ---: | ---: |
| `vkCreatePipelineLayout` | 95 | 99 | 1.04x |
| `vkCreateComputePipelines` | 39 | 41 | 1.05x |
| `vkCreateShaderModule` | 187 | 217 | 1.16x |
| `vkCreateGraphicsPipelines` | 74 | 88 | 1.19x |
| `vkAllocateMemory` | 272 | 333 | 1.22x |
| `vkCreateImage` / `vkBindImageMemory` | 200 | 282 | 1.41x |
| `vkCreateRenderPass` | 120 | 254 | 2.12x |
| `vkCreateDescriptorSetLayout` | 315 | 706 | 2.24x |

Those are RenderDoc's referenced-resource and initial-state serialisation, which is per-*capture*, not
per-frame — so restricting the table to command-stream ops is the right call. "Every count", stated
without that restriction, is a claim the artifact contradicts. Even within the command ops the spread
is wider than the table shows: `vkCmdBlitImage` 9→41 (4.56x), `vkUnmapMemory` 16→62 (3.88x),
`vkCmdDraw` 17→44 (2.59x).

**(b) The inference is void as a test of the withdrawn hypothesis.** Suppose the alternating
cat-frame/UI-frame hypothesis were *true* and the span held `cat, UI, cat`. The aggregate counts
would still land near 3x, because the shared per-frame setup triples either way and the distinctive
draws are a small minority of the total. An aggregate ratio near 3 is **equally consistent with
`A,B,A` and with `A,A,A`** — it has no power to separate them. This is the charter's
"discriminator that cannot show its lever moved is VOID, not negative", and a withdrawal is exactly
where that matters.

**(c) The discriminating measurement is in the same file, and it supports you — much more strongly
than the ratios do.** Segment `stray-8/frame.xml` by its repeating per-render-pass
`(vkCmdDrawIndexed, vkCmdDraw)` signature. The 192 passes decompose as:

- a **10-pass partial prefix** — the frame in progress when the span opened;
- **two complete frames of 66 passes each**, whose per-pass signature is **identical at all 66
  positions, zero differences**: 103 `vkCmdDrawIndexed` and 17 `vkCmdDraw` per frame;
- a **50-pass tail** cut by the iteration cap.

10 + 66 + 66 + 50 = 192. Comparing the two complete frames by full chunk-name sequence: 3,905 vs
3,900 chunks, differing **only** in streaming traffic — `Internal::Coherent Mapped Memory Write`
331 vs 320, `vkCmdCopyBufferToImage` 31 vs 33, `vkCmdPipelineBarrier` 610 vs 613, `vkUnmapMemory`
17 vs 18. Every render-pass, draw, dispatch and submit count is identical. **That** is a test the
alternating hypothesis could have failed, and it did not. Withdraw on this, not on the ratios.

**(d) The same decomposition gives you a fact about the #3322 default that the ratio table hid.** Let
`frame` be the 66-pass unit above. The 1-present capture's 65-pass signature satisfies
`p1 == frame[1:]` — **exact equality at all 65 positions**, and `frame[0]` is a 5-draw pass, which is
precisely the 103 − 98 = 5 discrepancy. So the one-present capture is one whole guest frame *minus
its leading render pass*: the `N=1` default already straddles a frame boundary and drops the frame's
first pass. That is a second, independent confirmation that the frames are identical (it matches a
frame body from a **different run**), and it is a good argument for this PR that the body does not
make.

Recipe, if you want to land it in #3126 or the doc: parse `<chunk ... name="...">` in order, bracket
`vkCmdBeginRenderPass`/`vkCmdEndRenderPass`, count draws per pass, and find the repeating unit. Note
`vkQueuePresentKHR` is **0** in both captures — the renderer's device never presents — so there is no
frame delimiter in the capture and this signature is the only way to segment it.

## 3. No doc update

`docs/GPU_PROFILING_EXTERNAL.md:88-89` enumerates the knobs by name — `PROSPER_RENDERDOC_AT_FRAME`,
`_AFTER_MS`, `_MAX_ITERS`. A fourth knob that is not in that list is undiscoverable, and the
`MAX_ITERS` interaction you documented in the PR body belongs there rather than in a merged PR
description. One or two sentences.

## 4. Nothing bounds capture size, and `MAX_ITERS` is a time bound, not a size bound

Measured from your own artifacts: the 1-present `.rdc` is **1.45 GB**, the 3-present one **3.84 GB** —
about 1.3 GB per guest present. The only bound is `MAX_ITERS`, which counts app-loop passes, so the
user who follows your own advice ("a slow title needs `MAX_ITERS` raised alongside `PRESENTS`") to
actually reach `N=8` gets roughly **10 GB** with nothing warning them. `renderdoc_capture.hpp:52-56`
already knows this class of hazard — it exists because the default path is under `/tmp` — and
`grabDir` defaults to `"."` (`main.cpp:1653`), which may or may not be real disk.

Not blocking (opt-in, default unchanged), but the cheapest fix is one clause on the line you already
added at `:2076-2078`: say the measured per-present cost there, so the person raising the cap is told
before the run rather than after. The doc note in 3 could carry it instead.

## 5. The span is "at least N", not "exactly N"

Several guest presents can land between two app-loop passes, so a `PRESENTS=2` span can close at 3.
Nothing lies — `:2331-2332` prints the real `guestNow - opened` — but the variable name reads as an
exact count. Worth a half-sentence in the doc. Relatedly there is no upper bound on
`renderdocPresents`; a huge value degrades to cap-only close, which is reported honestly, so this is a
naming/doc point rather than a correctness one.

## 6. Pre-existing stale comment this PR makes worse

`main.cpp:2766-2768` (present on `origin/main`, from #3322):

> Start it here, with its siblings, and close it at the top of the next iteration: that span covers
> one whole app-loop pass, which is the unit a reader means by "this frame".

That was already false once the predicate became a guest present; with `PRESENTS` it is further off.
You are editing this mechanism — one line to fix it here is cheaper than the next reader believing it.

## 7. The close predicate is still untestable

It is inline in a 3,000-line `main()`, which is why #3322's blocking bug got there. `capture_schedule.hpp`
next door already hosts `constexpr bool capture_frame_due(...)` with arms in
`test_capture_schedule.cpp`. A three-line sibling —
`constexpr bool renderdoc_span_complete(uint64_t now, uint64_t opened, uint64_t presents)` — would pin
the `N=1` equivalence and the `now >= opened` precondition that finding 0 above establishes by
reading. I verified that precondition by hand this time; a test would mean the next person does not
have to, and would fail loudly if someone ever wires `present_reset()` into a non-test path.

## 8. Minor

`PROSPER_RENDERDOC_PRESENTS=0` is well-formed but reported as "malformed". Identical to `MAX_ITERS`,
so it is consistent with the convention — noting it only for completeness, no change wanted. The
`span set to N` confirmation at `:2076` prints only when a trigger is armed *and* RenderDoc is
available, so setting `PRESENTS` alone is silent; that is fine, since nothing happens either way.

---

**Verdict: APPROVED.** Merge on the code. Findings 1, 3 and 6 are one-line fixes I would fold into this
PR; finding 2 is a PR-body/commit-message correction plus, ideally, the per-frame decomposition landed
in #3126 in place of the ratio table — the conclusion survives, and it deserves the evidence that
actually establishes it.

Issue #2932, opened 2026-08-22
The game frame renders but loses most presents to a wrong composite — flat white 4K clear or a near-black diagonal wedge, on BALAN, Little Nightmares II and Unbound


mattias800:
## Summary

Three UE4 titles surveyed on 2026-08-22 render **the game's real frame** — a menu, a logo sequence,
a full title screen — and then hand the present layer a **wrong frame** for most of the run. The
composite alternates; the game's picture is the minority state.

Two wrong-frame signatures appear, and one title shows both:

**(a) A flat white 4K clear**, RGB `(255,255,255)`, `distinct_rgb_colors == 1`,
`nonblack_rgb_pixels == 8294400`.

**(b) A near-black frame with a diagonal smeared band in the upper-left**, ~140-2000 colours. Under
a 12x brightness boost the band resolves into a smooth swept highlight with fine diagonal streaks
reaching the top-left corner — the look of a surface read at the wrong pitch or extent, not a dim
version of the scene.

| title | what renders | wrong-frame signature | duty cycle |
| --- | --- | --- | --- |
| `PPSA02058` BALAN WONDERWORLD | 4K language-select menu, full theatre art | (a) flat white | content on **20 of 110** 1 s samples, **irregular** |
| `PPSA02154` Little Nightmares II | 4K Bandai Namco / Tarsier / Unreal logo sequence | (a) **and** (b) | mixed across a 120 s run |
| `PPSA03274` Unbound: Worlds Apart | 4K title screen, wordmark and Cross prompt | (b) near-black wedge | content on **10 of 110** 1 s samples, at a **strict ~5 s cadence** |

## Measurement, and why the duty cycle needed a dense grid

The first pass sampled every 7 s and reported "3 of 20 frames carry content" for both BALAN and
Unbound. That number is an artefact of the grid (instrument trap 211). Re-measured at 1 s:

- **Unbound** is strictly periodic — content at t = 49, 54, 59, 64, 69, 74, 79, 84, 92, 98 s. A 5.0 s
  period, with the real frame visible for well under one second of each.
- **BALAN** is **not** periodic — content at t = 16, 18, 32, 33, 36, 45, 49, 50, 54, 56, 60, 69, 72,
  73, 77, 82, 86, 98, 103, 105 s, i.e. ~18% of samples with no cadence.

**Different cadences, so do not assume one mechanism.** They are filed together because the
observable — "the finished frame exists and something else reaches the screen" — is the same, and
[...]

Issue #1968, opened 2026-08-05
Sonic Frontiers (PPSA03831): prosper stops publishing frames after the opening logo movie — the renderer returns a 1920x1080 frame for a 4K present and the publish gate silently drops it


mattias800:
## Summary

On master `b8898a18`, Sonic Frontiers reaches and renders its own 4K opening logo sequence through
the live renderer, then stops. The stop is **not** a crash, not a renderer failure, and not a hung
process: the guest keeps flipping VideoOut at a steady rate while producing no new GPU work and
loading no new content, and the last published frame is uniformly black.

## Evidence

Two bounded native Linux/Vulkan arms, both `exit 0`, both on `b8898a18`. RADV, shared GPU, so no
timing or FPS figure is quoted.

**Arm 1 — default launch, no input**, 12 samples every 10 s:

```bash
PROSPER_TEMP0=~/frontiers-work/temp0 PROSPER_SAVEDATA_DIR=~/frontiers-work/savedata \
PROSPER_GUEST_FS=1 PROSPER_GUEST_ARGS=-force-gfx-direct PROSPER_RENDER=1 \
timeout --foreground -s TERM -k 5s 175s \
  ./prosper/build-linux/screenshot <DUMP_ROOT>/PPSA03831-app0 \
  --seconds 10 --count 12 --timeout 150 --out ~/frontiers-work/shots \
  --manifest ~/frontiers-work/manifest.json
```

**Arm 2 — the same, plus `PROSPER_PAD_SCRIPT=@prosper/scripts/sonic-frontiers/reach-title.pad`**,
15 samples every 20 s over 300 s. Input changes nothing about the stop.

Arm 2's manifest timeline is the finding:

| t (s) | `frame_seq` (renderer publishes) | `present_count` (guest flips) | source | distinct RGB |
| --- | --- | --- | --- | --- |
| 20 | 0 | 62 | raw_scanout | 1015 |
| 40 | 0 | 127 | raw_scanout | 1 |
| 60 | 355 | 185 | composited | 159866 |
| 80 | 750 | 242 | composited | 23211 |
| 100 | 1163 | 301 | composited | 25946 |
| 120 | 1623 | 366 | composited | 25946 |
| 140 | 2075 | 431 | composited | 25946 |
| 160 | **2081** | 495 | composited | 1 |
| 180–300 | **2081** (frozen) | 558 → 916 | composited | 1 |

[...]

Issue #3014, opened 2026-08-26
boot_trace does not compile on Windows (POSIX setenv) — and because ninja stops there, 15 ctest cases silently do not run


mattias800:
`prosper/tools/boot_trace/boot_trace.cpp:263` calls POSIX `setenv`, which MinGW does not provide, so
the `boot_trace` target has not compiled on Windows since `d154907a` (2026-08-11, the
`--capture-first-frame` observer).

```
boot_trace.cpp:263:9: error: 'setenv' was not declared in this scope; did you mean 'getenv'?
```

## Why this is worse than one broken tool

`boot_trace` is one target in a whole-tree build, so ninja stops there and **most test executables
never link**. On this machine that surfaced as:

```
ctest EXIT=8 ... 15 cases (Not Run)
```

— `initfault_dump_survives`, `hle_stack_args`, `ime_fs`, `x86_read_decode`, `fault_context`,
`agc_dcb_callback_abi`, `agc_submit_to_gpustate`, `videoout_queries`, `game_compute_exec`,
`game_compute_exec_no_adaptive_storage_result`, `video_media_foundation`, `audio_sdl3`,
`shared_vulkan_device`, `live_target_format`, `present_blit`.

So a compile error in a diagnostic tool reads as fifteen tests quietly not running — the exact shape
CLAUDE.md warns about, where "no tests ran" and "everything passed" are hard to tell apart. Anyone
building a subset target (`--target prosper-app`) never sees it, which is why it survived a fortnight.

After the fix: **242/242 pass, ctest exit 0** with `--no-tests=error`.

## Fix

Guard it the way `prosper-app` already does (`frontends/prosper-app/main.cpp:132`, `set_environment`):
`_putenv_s` on `_WIN32`, `setenv` elsewhere.

Worth considering separately: a Windows CI job that builds the **whole** tree would have caught this
on the introducing commit.

mattias800:
## Correction to this issue's suggested follow-up

The issue closes by suggesting *"a Windows CI job that builds the whole tree would have caught this on
the introducing commit."* **That is wrong, and the real gap is narrower and more interesting.**

`.github/workflows/ci.yml` already has a **Windows MinGW** job, and it does build the whole tree and
run the suite properly:

```
cmake -S prosper -B prosper/build-ci -G Ninja -DCMAKE_BUILD_TYPE=Debug \
[...]

PR #3331, opened 2026-09-04
diag: resolve guest VAs to physical addresses; the Stray scanout alias hypothesis is falsified (#2932)

mattias800:
Refs #2932, #3126, #2883.

## What this is

Two diagnostics that resolve a guest virtual address to its physical address, and the `## Ruled out`
row recording what they falsified. **No production path changes** — the resolver is diagnostic-only
and the target census is behind an environment variable that is absent by default.

## The failure this was aimed at

*Stray*'s main menu renders a full 3D scene — 27 render passes, 38 draws, 13 dispatches, real depth
targets — into prosper's own render targets, and then presents **only the UI bar** over an absent
world.

Both publish paths fail, and the two failures look like one cause:

- `select_present_source` returns `None` — `px_front=none px_vo=none px_last=none`, so nothing is
  published (`fresh=0 retained=0`)
- the flipped buffer `0x9fc0000000` reports `authored=0` — its bytes are unchanged since registration,
  so the guest-scanout fallback (#1968) declines `SkipNotAuthored`

The present path matches a flipped scanout buffer against render targets by **virtual** address
(`is_live_render_target`), and this title renders into `0x30…` while flipping `0x9fc0000000` — about
450 GiB apart. No VA-keyed lookup can connect those.

## The hypothesis, and why it deserved a probe rather than a patch

If the scanout buffer and the render targets were two virtual mappings of **one physical allocation**,
every observation above would follow with no further defect — and the fix would be a reuse of
resolution prosper already performs, since the write-watch models exactly this aliasing in
`WatchedPage { phys; aliases; }`.

That is an attractive hypothesis: it converts a measured fact into a bookkeeping error with a known
remedy. It is also the kind this project's charter says to test before building on.

## Result: FALSIFIED

| buffer | guest VA | physical |
| --- | --- | --- |
| flipped scanout | `0x9fc0000000` | **`0x230000`** (2.3 MB) |
| lowest render target | `0x30…` | `0x1e980000` (511 MB) |
| highest render target | `0x30…` | `0x79f20000` (2.0 GB) |

**80,511 target resolutions, 0 unresolved** — the census saw the whole population, not a sample that
happened to miss the alias. The scanout buffer sits ~509 MB below every render target. Separate
physical allocations; nothing is aliased.

So the guest composites into its own allocation and moves it to scanout by a route prosper does not
model. The open question is **which packet does that**, not how the present path matches addresses.

## Approach and invariants

`guest_write_watch_va_to_phys()` resolves over `WatchState::aliases` — the live dmem topology,
populated by `guest_write_watch_notify_direct_mapping_added` — rather than introducing a second notion
of "which VAs map this page". A duplicated resolver that could disagree with the write-watch would be
worse than none: two subsystems would then answer one question differently.

It takes the same lock as every other reader of that state, and returns `false` rather than a
plausible-looking zero when nothing covers the address.

## Deliberately unaffected

- No production caller. The resolver is referenced only by the two probes.
- `PROSPER_TARGET_PHYS` is read once into a `static const bool`; unset, the census costs one branch.
- The scanout probe is unconditional but sits inside the existing `diag_should_print(ord)` rate limit,
  beside the `GUEST SCANOUT` decision line it belongs with — so the answer arrives *with* the failure
  rather than in a separate run whose guest state may differ.

## Risks and limits (both recorded in the doc row)

- **`guest_write_watch_va_to_phys` returns the FIRST containing range and does not check that the
  range is large enough for the access.** Adequate for a probe; not for anything load-bearing. A
  production caller would need the size check.
- **Only one of two scanout buffers has been probed.** Stray registers `0x9fc0000000` **and**
  `0x9fc2000000` — a rotating set, which is the shape behind *Bendy and the Ink Machine*'s black
  flicker (see the retained-frame comment in `present_extent.hpp`). "Nothing ever writes scanout" is
  **not** established and the row says so.

## An instrument note

The first revision of this probe resolved against `w.trace.pages` and reported `UNRESOLVED`. That is
the dmem **write-trace** — a narrow diagnostic armed for one investigation at a time — not the mapping
registry. The answer read as "prosper does not know this mapping" when it meant "this diagnostic was
not armed for it". The comment on the resolver records the distinction so the next reader does not
repeat it.

## Verification (exact head)

Configured in the project container against `<DUMP_ROOT>/PPSA24651-app0`:

```
cmake -S prosper -B prosper/build-linux -DPROSPER_APP=ON -DGAME_DUMP=<DUMP_ROOT>/PPSA24651-app0 \
      -DCMAKE_BUILD_TYPE=RelWithDebInfo
cmake --build prosper/build-linux -j6                     # rc=0
ctest --no-tests=error -j4                                # rc=0
```

**`100% tests passed out of 331`**, exit code 0. One test did not run: `plugin_autolink (Skipped)`.

Docs gate on the changed file, and the deletion check:

```
python3 prosper/tools/docs/check_numbered_table.py prosper/docs/STRAY_STATUS.md
  -> 13 table(s), 42 rows, unbroken, every row matching its header      # rc=0
git diff --check                                                        # rc=0
git diff --numstat origin/master..HEAD   -> every file 0 deletions
```

The diff is **106 added lines and 0 deleted lines** across five files.

No snapshots were run: this PR changes no rendering behaviour, and the charter makes snapshots a
release-time inventory rather than a per-PR gate.

## Review

Worth a read despite the size — the resolver is new shared-substrate code reading memory topology
under a lock, and the judgement being made (resolve over `aliases`, not `trace.pages`) is exactly the
kind an independent reader can challenge. The size caveat above is the most likely place a reviewer
will want a stronger contract.

### DIFF
--- a/BLOG.md
+++ b/BLOG.md
@@ -19,6 +19,18 @@ from the tracker issues, and still gated, because it is a projection of state ra
 > title's current state — for that, read the tracker. Nothing is ever removed when a title moves on,
 > because the point of a blog is that it records *when* things happened.
 
+## 2026-09-04
+
+### Stray's missing world is not an address-matching bug, and now we can prove it
+
+No picture with this one — it is a dead end worth recording. Stray's main menu draws a full 3D scene
+into prosper's own render targets and then presents only the UI bar over black, and the tempting
+explanation was that the buffer the game flips to the screen and the buffers we render into are two
+names for the same memory, which we were simply failing to connect. They are not: measured directly,
+they sit half a gigabyte apart in physical memory, across 80,511 resolutions with nothing left
+unresolved. So the game really is copying its finished frame somewhere else, by a route we do not yet
+follow — which is a narrower question than the one we started with.
+
 ## 2026-09-02
 
 ### Astro Bot's world map is lit, and no longer takes the GPU down with it
--- a/prosper/docs/STRAY_STATUS.md
+++ b/prosper/docs/STRAY_STATUS.md
@@ -524,3 +524,36 @@ evidence nor falsifications for it. Re-run each with `PROSPER_NULL_PAGE=1` and a
   may unblock some OTHER draw that shares it but addresses a genuinely single-mip resource; it does
   **not** close #3134 on its own. Full detail: `docs/RECOMPILER_REMAINING.md` § Ruled out. #3134,
   #2818.
+- **"The scanout buffer and the render targets are two virtual mappings of ONE physical allocation,
+  so the VA-keyed present match is the whole defect."** Falsified by direct resolution
+  (`PROSPER_TARGET_PHYS=1` plus the unconditional `[rtt] GUEST SCANOUT #N phys:` probe, both landed
+  with this row). This was the most attractive hypothesis available, because it would have made a
+  measured fact — prosper renders into `0x30…` while the guest flips `0x9fc0000000`, ~450 GiB apart
+  in VA — into a *bookkeeping* error with a fix prosper already had the machinery for. It is not:
+
+  | buffer | guest VA | physical |
+  | --- | --- | --- |
+  | flipped scanout | `0x9fc0000000` | **`0x230000`** (2.3 MB) |
+  | lowest render target | `0x30…` | `0x1e980000` (511 MB) |
+  | highest render target | `0x30…` | `0x79f20000` (2.0 GB) |
+
+  80,511 target resolutions, **0 unresolved** — so the census saw the whole population rather than a
+  sample that happened to miss the alias. The scanout sits ~509 MB *below* every render target. They
+  are separate physical allocations and nothing is aliased.
+
+  **What this leaves is narrower and better posed.** The guest composites into its own allocation and
+  moves it to scanout by a route prosper does not model, so the open question is *which packet does
+  that*, not how the present path matches addresses. Two further facts bound it: `select_present_source`
+  returns `None` (`px_front=none px_vo=none px_last=none`, `fresh=0 retained=0`), and
+  `videoout_buffer_authored_locked` reports `authored=0` for the flipped buffer — its bytes are
+  unchanged since registration. That second one is the load-bearing constraint, because the digest test
+  reads guest memory directly: **a GPU copy landing in guest memory would flip `authored` on its own.**
+  So the copy is not happening, not landing there, or not being executed by prosper at all.
+
+  **Two cautions for whoever picks this up.** Stray registers **two** scanout buffers,
+  `0x9fc0000000` and `0x9fc2000000` — a rotating set, which is the shape behind *Bendy and the Ink
+  Machine*'s black flicker (see the retained-frame comment in `present_extent.hpp`); only the first has
+  been probed, so "nothing ever writes scanout" is not yet safe to claim. And
+  `guest_write_watch_va_to_phys` returns the FIRST range containing the address without checking that
+  range is large enough for the access — fine for a probe, not for anything load-bearing.
+  #2932, #3126.
--- a/prosper/frontends/shared/live/live_renderer.cpp
+++ b/prosper/frontends/shared/live/live_renderer.cpp
@@ -9744,6 +9744,27 @@ void register_live_renderer(const std::string& frame_dir, bool dump_bmps_request
                                     prosper::frontend::guest_scanout_decision_name(decision), w, h,
                                     read.metadata.tiling_mode, (int)read.guest_authored,
                                     read.padded_footprint ? "padded" : "nominal");
+                        // #2932: does the flipped VA resolve to a physical page the write-watch
+                        // knows? The present path matches scanout against render targets by VIRTUAL
+                        // address only, and this title renders into 0x30... while flipping
+                        // 0x9fc0000000 -- ~450 GiB apart, so no VA-keyed lookup can connect them.
+                        // If the two are aliases of one physical range, that gap is the whole defect
+                        // and the fix reuses resolution prosper already performs. Printed beside the
+                        // decision so the answer arrives with the failure rather than in a separate
+                        // run whose guest state may differ.
+                        if (prosper::diag_should_print(ord)) {
+                            uint64_t flip_phys = 0;
+                            const bool ok = prosper::host::guest_write_watch_va_to_phys(
+                                read.metadata.address, flip_phys);
+                            fprintf(stderr,
+                                    "[rtt] GUEST SCANOUT #%llu phys: va=0x%llx -> %s\n",
+                                    (unsigned long long)ord,
+                                    (unsigned long long)read.metadata.address,
+                                    ok ? "" : "UNRESOLVED (no direct mapping covers this VA)");
+                            if (ok)
+                                fprintf(stderr, "[rtt]   flip phys=0x%llx\n",
+                                        (unsigned long long)flip_phys);
+                        }
                     }
                     // Re-check the size on the cached path too: a two-set geometry switch can change
                     // the present extent between spans of one flip, and the cache is keyed on the
--- a/prosper/src/host/memory/guest_write_watch.cpp
+++ b/prosper/src/host/memory/guest_write_watch.cpp
@@ -1125,6 +1125,28 @@ bool trace_va_to_phys_locked(const DmemTraceState& trace, uint64_t addr, uint64_
 
 } // namespace
 
+// Diagnostic resolution of a guest VA to its physical address (#2932). Reuses the trace's own alias
+// table rather than adding a second notion of "which VAs map this page": a duplicated resolver that
+// could disagree with the write-watch would be worse than none, because two subsystems would then
+// answer the same question differently.
+bool guest_write_watch_va_to_phys(uint64_t addr, uint64_t& phys) {
+    if (!addr) return false;
+    WatchState& w = state();
+    std::lock_guard lock(w.mutex);
+    // `w.aliases` is the LIVE DMEM TOPOLOGY -- every direct mapping the guest has made, recorded by
+    // guest_write_watch_notify_direct_mapping_added. `w.trace.pages` is a different thing: a narrow
+    // write-trace armed for one investigation at a time. An earlier revision of this probe resolved
+    // against the trace and reported UNRESOLVED for a scanout buffer, which reads as "prosper does
+    // not know this mapping" when it means "this diagnostic was not armed for it" -- the instrument
+    // answering a different question than the one asked.
+    for (const AliasRange& a : w.aliases) {
+        if (!a.size || addr < a.addr || addr - a.addr >= a.size) continue;
+        phys = a.phys + (addr - a.addr);
+        return true;
+    }
+    return false;
+}
+
 GuestWriteWatch::~GuestWriteWatch() { reset(); }
 GuestWriteWatch::GuestWriteWatch(GuestWriteWatch&& other) noexcept
     : id_(std::exchange(other.id_, 0)) {}
--- a/prosper/src/host/memory/guest_write_watch.hpp
+++ b/prosper/src/host/memory/guest_write_watch.hpp
@@ -193,6 +193,20 @@ void guest_write_watch_notify_direct_mapping_added(uint64_t addr, uint64_t size,
 void guest_write_watch_notify_direct_mapping_removed(uint64_t addr, uint64_t size);
 void guest_write_watch_notify_direct_mapping_protection(uint64_t addr, uint64_t size,
                                                         uint32_t protection);
+// PROSPER_SCANOUT_PHYS (#2932): resolve a guest VA to its physical address through the dmem trace.
+//
+// The present path matches a flipped scanout buffer against render targets by VIRTUAL address only
+// (`is_live_render_target`), and `hle_graphics.cpp`'s scanout registry has no aliasing concept at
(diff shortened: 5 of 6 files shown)


mattias800:
Closing: wrong base branch. This targeted `master`, which is a stale parallel line; the repository's development branch is `main` (its default, and the base of every recent merge). The same two commits are correctly based on `origin/main` and will be reopened against it — no finding here changes, only the base.

Apologies to the reviewer for the churn; the replacement PR will be linked here.

mattias800 (commented):
**REJECTED** — four blocking findings. One is a hard Windows link break that will redden CI; three are
claims in the artifacts this PR exists to produce (the `## Ruled out` row and the probe's own output
strings) that the code cited does not support.

The judgement the PR body flags for review — resolve over `w.aliases`, not `w.trace.pages` — is
**correct**, and the resolver's locking and arithmetic are clean. The problems are one platform gap and
three overclaims, all cheap to fix.

---

## Blocking

### B1. Windows link failure: the function is defined in only one of the file's two platform branches

`prosper/src/host/memory/guest_write_watch.cpp:1132` defines `guest_write_watch_va_to_phys` inside the
`#else` branch that opens at `:540` ("Linux / macOS"). The `#ifdef _WIN32` branch spans `:70`–`:538` and
contains **no definition** — verified: `awk 'NR>=70 && NR<=538' | grep -c va_to_phys` returns `0`. The
declaration in `guest_write_watch.hpp:208` is unconditional, and there is no inline fallback.

`prosper/frontends/shared/live/live_renderer.cpp:9757` calls it **unconditionally** — the nearest
`#ifdef _WIN32` in that file closes at `:2415`, ~7,300 lines earlier, so the probe is not
platform-gated.

Failure scenario, concrete: CI job **Windows App** (`.github/workflows/ci.yml:647`) is the one Windows
job that installs `mingw-w64-ucrt-x86_64-vulkan-headers` + `vulkan-loader`, so `Vulkan_FOUND` is true
there, `CMakeLists.txt:2193` builds `prosper_live_renderer` from `live_renderer.cpp`, and
`CMakeLists.txt:3133` links it into `prosper-app`. MinGW defines `_WIN32`, so `prosper_core`'s
`guest_write_watch.cpp.o` carries the Windows branch. Result: `undefined reference to
prosper::host::guest_write_watch_va_to_phys(unsigned long long, unsigned long long&)` at the
`prosper-app` link.

This is consistent with the pass/fail pattern already visible on this run: **macOS App passed** (the
`#else` branch covers Darwin), **Linux App passed**, and **Windows App is still pending**. The other two
Windows/macOS jobs configure `-DCMAKE_DISABLE_FIND_PACKAGE_Vulkan=TRUE`, so they never build
`prosper_live_renderer` and cannot catch this — the same split `windows-app`'s own CI comment describes
as how #3014 survived a fortnight.

The author's verification (331/331, rc=0) is a Linux-only result and could not have seen this.

Remedy: the Windows branch has the same `WatchState::aliases` (`:123`) and the same `AliasRange`
(`:91–96`, `protection` unused by the resolver), so the identical body compiles there — either
duplicate it into the `#ifdef _WIN32` branch or hoist one definition above the `#if`.

### B2. `STRAY_STATUS.md:548–551` — the row's self-declared "load-bearing constraint" is not established

The row states:

> `videoout_buffer_authored_locked` reports `authored=0` for the flipped buffer — **its bytes are
> unchanged since registration**. That second one is the load-bearing constraint, because the digest
> test reads guest memory directly: **a GPU copy landing in guest memory would flip `authored` on its
> own.** So the copy is not happening, not landing there, or not being executed by prosper at all.

`videoout_buffer_authored_locked` (`prosper/src/hle/graphics/hle_graphics.cpp:478–489`) returns `false`
for **four** distinct reasons, only the last of which is "bytes unchanged":

```
:480  slot out of range                     -> false
:481  buffer_generation[slot] != generation -> false
:483  !buffer_digest_valid[slot]            -> false   <-- no baseline was ever seeded
:485  digest == baseline                    -> false
```

`:483` is the one that breaks the inference. `videoout_present.hpp:58–60` says so in the contract
itself: *"False also means 'no baseline was taken' (the range was not readable at registration), which
is deliberately indistinguishable from 'not written'."* And `hle_graphics.cpp:373` lists it first among
the safe-direction failures: *"baseline unreadable at registration -> no baseline -> 'not authored' ->
decline"*.

Failure scenario: if `buffer_digest_valid[0] == 0` for `0x9fc0000000`, then a GPU copy of **any** size
landing there would **not** flip `authored`, because `:483` returns before the digest is ever computed.
The row's second sentence is then false, and the disjunction it derives omits the live fourth branch:
*the copy happened and prosper cannot tell, because it has no baseline.* A fifth, weaker gap: the digest
samples only `kDigestSamples = 4096` evenly spaced 8-byte words (`:384`, `:397–404`), and
`hle_graphics.cpp:376` records that a write covering no sample point also reads as unauthored.

This is the sentence the row itself nominates as load-bearing, and the next agent will inherit it from a
`## Ruled out` section. Remedy: either state the disjunction honestly, or discriminate — `:472` and
`:442` are the only writers of `buffer_digest_valid`, and `prosper_videoout_seed_stats`
(`hle_graphics.cpp:495`) already reports how many baselines were seeded.

### B3. `STRAY_STATUS.md:540–543` — "the census saw the whole population" is not what the census measures

The row argues:

> 80,511 target resolutions, **0 unresolved** — so the census saw the whole population rather than a
> sample that happened to miss the alias. The scanout sits ~509 MB *below* every render target.

`0 unresolved` is a statement about the **resolver succeeding on what it was shown**, not about **what it
was shown**. The census at `prosper/tests/fixtures/render_runner.h:5165` sits inside
`if (persistent_color)` (`:5153`, from `persistent_color_enabled` at `:4762`) and reads
`color_target->persistent_id` — **colour slot 0 only**. `persistent_id1` (`:5286`) and
`persistent_id_slots[2..7]` (`:5388`) are never censused; neither is any depth target. And 80,511 is a
count of `render_draw_pass_rgba` invocations, not of distinct VAs.

Failure scenario: an MRT1 or slot-2..7 allocation aliasing physical `0x230000` is exactly the case the
hypothesis predicts and exactly the case this census cannot see — so a loud non-null (80,511 / 0) is
laundering a domain gap. This is the charter's own trap: *"A positive control drawn from the same source
as the null it validates tests the DISCRIMINATOR, never the DOMAIN."*

Remedy is cheap either way: scope the wording to "every persistent colour-0 target", or add
`persistent_id1` / `persistent_id_slots[]` to the census loop and re-run.

### B4. `live_renderer.cpp:9764` — the probe asserts a cause it cannot establish, reproducing the trap the PR narrates

On failure the probe prints the literal `"UNRESOLVED (no direct mapping covers this VA)"`. That is one
cause. The other is that `w.aliases` is **empty**: `guest_write_watch_notify_direct_mapping_added`
returns early at `guest_write_watch.cpp:1623` —

```
if (!w.fault_onstack.load(std::memory_order_acquire)) return;   // feature off -> table is never used
```

`fault_onstack` is set only from `exec_image_linux.cpp:3489` and only when the installed handler carries
`SA_ONSTACK`; `set_fault_onstack` is a no-op off Linux (`guest_write_watch.cpp:1608–1616`) and an empty
stub on Windows (`:536`). So on any run where the write-watch was never armed, **every** VA — scanout and
render target alike — resolves `false`, and the probe states as fact that no mapping covers it.

This is the same shape as the instrument note the PR is proud of ("the answer read as *prosper does not
know this mapping* when it meant *this diagnostic was not armed for it*"), moved one level up from
`trace.pages` to `aliases` and left undetected. It did not corrupt the evidence in *this* run — the flip
resolved to `0x230000` and the census reported 0 unresolved, so the table was demonstrably live — but the
probe is being landed as shared substrate for future `## Ruled out` rows, and its false branch is
currently a confident wrong sentence.

Remedy is one line: distinguish empty-table / feature-off from not-covered, e.g. return a tri-state or
print `"UNRESOLVED (aliases=%zu)"` so a zero table is visible in the output.

---

## Non-blocking

- **`guest_write_watch.hpp:196` names an environment variable that does not exist.** The header says
  `PROSPER_SCANOUT_PHYS (#2932)`; the only gate implemented is `PROSPER_TARGET_PHYS`
  (`render_runner.h:5164`), and the scanout probe is unconditional. `grep -rn PROSPER_SCANOUT_PHYS`
  returns exactly this one comment. A reader greps the name and finds nothing.
- **`guest_write_watch.hpp:196` and `.cpp:1128` both attribute the alias table to the trace**, which the
  function body's own comment (`:1136–1141`) exists to deny: the header says "through the dmem trace"
  and the leading `.cpp` comment says "Reuses the **trace's** own alias table", while the body says
  `w.aliases` is *not* `w.trace.pages`. A reader who stops at the header repeats the exact mistake the
  PR says it recorded to prevent.
- **`live_renderer.cpp:9760–9769` — the success line is incoherent on its own.** `ok ? "" : "UNRESOLVED…"`
  means a successful resolve prints `… va=0x9fc0000000 -> ` with a dangling arrow and nothing after it,
  and the value arrives on a separate `[rtt]   flip phys=` line. It reads as a truncated write. One
  `fprintf` with the phys formatted into the same line would be both shorter and greppable.
- **`live_renderer.cpp:9757` can be reached with `read.metadata.address == 0`** when
  `videoout_read_front_linear` returns false (`read` is then default-constructed) — the probe prints
  `va=0x0 -> UNRESOLVED (no direct mapping covers this VA)`, which is true but noise.
- **`render_runner.h:5169–5171` prints `UNRESOLVED 0x0`** — it emits `tphys` on the failure path too. At
  least it is labelled; consider omitting the value.
- **`guest_write_watch.hpp:196–208` splits the `notify_*` family**, inserting the resolver between
  `..._notify_direct_mapping_protection` and `..._notify_physical_write`.

---

## Verified good (checked against source, not the PR description)

- **Locking is correct.** `w.mutex` is a plain `std::mutex` and the file uses a `_locked` suffix for
  callers-hold-the-lock helpers; the new function is *not* suffixed and takes the lock itself, which is
  right. Both call sites (`live_renderer.cpp:9757`, `render_runner.h:5167`) are external to
  `guest_write_watch.cpp` and hold nothing; no re-entrant path exists (the body touches only host heap,
  so it cannot fault into the SIGSEGV handler that also takes this mutex).
- **The containment test cannot overflow.** `addr < a.addr` is evaluated before `addr - a.addr`, so the
  subtraction is never negative; `>= a.size` bounds the far end; `!a.size` rejects the degenerate range.
- **First-match is safe here.** `notify_direct_mapping_added` (`:1622`) purges a VA range that already
  carries an alias before pushing (`va_range_has_alias` / `purge_va_range_locked`), and
  `..._removed` (`:1675`) purges on unmap, so overlapping VA entries are not the normal case. The
  no-size-check caveat is correctly disclosed in the doc row.
- **The include resolves everywhere.** `CMakeLists.txt:98` is
  `target_include_directories(prosper_core PUBLIC src)`, so every one of the 30+ targets that include
  `render_runner.h` gets `src` on its path. `<cstdlib>` and `<cstdio>` were already included
  (`render_runner.h:25–26`).
- **`persistent_id` really is a guest VA.** `live_renderer.cpp:8672` constructs
  `BackendColorTarget{base, …}` from the pass's colour-0 base, the same value keyed into `g_rtt` and
  printed as `target=0x%llx` at `:8665`. It is *not* the synthetic `++persistent_texture_id` counter at
  `:6406`, which belongs to the texture cache. The census output is meaningful.
- **Format specifiers match.** `%llu`/`%llx` against `(unsigned long long)` casts, `%s` against `const
  char*`, and `W`/`H` are `uint32_t` against `%u` (`render_runner.h:4661`).
- **`diag_should_print` is pure** (`diag_ratelimit.hpp:30–33`), so the two calls at `:9741` and `:9756`
  agree — the phys line cannot appear without the decision line it annotates.
- **`check_diag_gates.py` is satisfied.** It runs as a ctest (`CMakeLists.txt:360`) and in CI
  (`ci.yml:358–359`); the census declares and writes `tphys` inside the same gated block, so it is
  neither SPLIT-LOCAL nor TWO-GATE, and `persistent_color_targets_enabled` is a *negated* env test
  (`render_runner.h:4760`), which the scanner deliberately treats as ungated. No baseline entry needed.
- **Gating convention followed** — `PROSPER_TARGET_PHYS` read once into a `static const bool`, absent by
  default, one branch when unset.
- **The `BLOG.md` entry is within the charter's picture-free exception** and correctly short.
- **The rest of the doc row is well-judged**, including the two disclosed cautions (only one of two
  scanout buffers probed; first-containing-range without a size check).

---

To clear this: fix B1 (Windows definition), and narrow B2/B3 to what the instruments measure — the
falsification survives both edits, it just stops being stated more strongly than the evidence. B4 is a
one-line message change.

mattias800:
Empirical confirmation of B1 from my review above (this is a note, not a second verdict — the review
stands as posted): **Windows App has now gone red at 7m36s** on this head, while Linux App, Linux App
Package and macOS App all passed. That is the predicted signature — macOS takes the same
`#else` branch of `guest_write_watch.cpp` that defines the function, and the two green Windows-adjacent
jobs are the ones configured `-DCMAKE_DISABLE_FIND_PACKAGE_Vulkan=TRUE`, which never build
`prosper_live_renderer` at all.

Note the job builds `boot_trace` as well as `prosper-app` (`CMakeLists.txt:2217` links
`prosper_live_renderer` into it on Windows), so both targets in that job hit the same undefined
reference.

mattias800:
Replacement opened as #3332, against `main`. All four blocking findings from the review here are fixed there and the review's substance carried across verbatim — thank you, B1 was a real break and B2/B3 were overclaims of mine that the falsification did not need.

PR #3332, opened 2026-09-04
diag: census guest VA->physical across all colour slots; prosper renders INTO Stray's scanout buffers (#2932)

mattias800:
Refs #2932, #3126, #2883. Replaces #3331 (opened against the stale `master` line — see #3333).

A VA→physical resolver plus two probes, and the `## Ruled out` row recording what they found.
**No production path changes**: the resolver is diagnostic-only, and the census is behind an
environment variable absent by default.

## What was falsified

**The alias hypothesis** — that a flipped scanout buffer and the render targets were two virtual
mappings of one physical allocation, which would have made the missing composite a bookkeeping error
with a fix prosper already had machinery for.

Census over slots 0–6 on the title route, failing steady state confirmed by eye (menu over black at
8× brightness): **55,868 resolutions, 0 unresolved, 82 distinct targets**. Excluding the scanout
buffers themselves, targets whose **physical extent** overlaps either scanout buffer: **0**; nearest
is **390 MB** away.

Slot coverage was added after review and earned it: **24% of resolutions (13,669) are in slots 1–6**,
and since `is_live_render_target` consults `g_rtt` — populated for every slot — an alias there would
have been exactly the proposed link, in exactly the region a slot-0 census cannot see.

## What was corrected — this branch's own premise

**The scanout buffers are render targets.** Both flipped VAs appear in the census as 4K slot-0
colour attachments, throughout the failing run:

| flipped VA | physical | census lines | draws dropped `no-effect(early)` |
| --- | --- | --- | --- |
| `0x9fc0000000` | `0x230000` | 946 | **1,143** |
| `0x9fc2000000` | `0x2210000` | 976 | **1,154** |

The drop figures are new: this diagnostic had only ever been run on the **calibration** screen, which
the status doc explicitly warns is a different screen. 4,096 draws discarded overall, 2,297 of them
aimed at the scanout buffers. Read from the final census emission, not a running total.

So the original framing — *"prosper renders into `0x30…` while the guest flips `0x9fc…`, ~450 GiB
apart, so no VA-keyed lookup can connect them"* — **is withdrawn**.

**Read the census column precisely.** The line prints after the empty-draw early-out, so it licenses
*"bound as slot 0 of a pass carrying ≥1 draw"* — more than "the target exists", less than "a draw's
output reached memory". The row and the code comment both say so.

## Two regimes, previously blended

The guest-scanout fallback is gated on `guest_scanout_read_warranted(...) == Publish` — it runs
**only when nothing else published**:

| regime | what publishes | source of the `authored=0` / `SkipNotAuthored` evidence |
| --- | --- | --- |
| **title screen** (measured here) | the menu **does** publish | *not this regime* |
| **main menu / first map** | nothing publishes | this one |

Measured: **0** `[rtt] GUEST SCANOUT` lines across a full title-route run that reached and held the
failing steady state. So those facts belong to the *other* regime, and the row previously mixed them
— the same mistake this file's own banner warns about for `max_nonblack`.

## What this leaves

A better question. The two tempting branches ("only the UI is drawn there", "the scene is drawn then
wiped") both exclude the branch the status doc **already measures**: the scene reaches the target and
is stored as black by magnitude — PixelHistory has tonemap `shaderOut 0.001099` storing `0`, because
`0.001099 × 255 = 0.28` rounds to zero in `R8G8B8A8_UNORM`. The open part is why the value arriving
at the store is that small, not where the picture goes.

## Review history

Three rounds, all REJECTED, all correct; every finding verified against source before fixing.

- **R1 (4)** — a Windows link break (`Windows App` went red; now **passes**); `authored=0` misread as
  "bytes unchanged since registration"; a coverage claim; a probe message asserting a cause it could
  not establish.
- **R2 (1)** — the slot-0 scope clause. Remedy taken was to *widen the census*, which produced the
  correction above rather than scoping around the gap.
- **R3 (3)** — the retracted sentence still shipping in **three code comments in the same commit**
  (including a shared-substrate header contract), an over-correction in the opposite direction, and
  "renders directly INTO" overreaching.

## Invariants and limits

The resolver reads `WatchState::aliases` — the live dmem topology — rather than adding a second
notion of "which VAs map this page"; a duplicated resolver that could disagree with the write-watch
would be worse than none. The alias count is an out-param filled under the **same lock acquisition**
as the lookup, so a miss and its searched domain cannot describe two different tables.

- Returns the **first** containing range without checking it is large enough. Probe-only; a
  production caller needs the size check.
- `authored=0` has four causes; **two** are a blind instrument rather than an absent write, so
  `buffer_digest_valid` alone does not separate them.
- A *declined* MRT1 base never sets `persistent_id1`, so it cannot reach the census.

## Verification (exact head)

```
cmake -S prosper -B prosper/build-linux -DPROSPER_APP=ON \
      -DGAME_DUMP=<DUMP_ROOT>/PPSA24651-app0 -DCMAKE_BUILD_TYPE=RelWithDebInfo
cmake --build prosper/build-linux -j6      # rc=0, 0 errors
ctest --no-tests=error -j4                 # rc=0
```

**`100% tests passed out of 357`**, exit 0; `plugin_autolink (Skipped)`.

```
check_numbered_table.py prosper/docs/STRAY_STATUS.md -> 26 tables, 94 rows, rc=0
git diff --check                                     -> rc=0
```

**252 added, 1 deleted.** The single deletion is the inlined `g_rtt.find(...)` argument, replaced by
the named `renderer_owns_flip` so the probe can report it — not a whole-file clobber.

Linux-only locally, so the Windows arm rests on structure plus CI, stated as what was checked where.
No snapshots: no rendering behaviour changes.

### DIFF
--- a/BLOG.md
+++ b/BLOG.md
@@ -21,6 +21,16 @@ from the tracker issues, and still gated, because it is a projection of state ra
 
 ## 2026-09-04
 
+### Stray's missing world is not an address-matching bug, and now we can prove it
+
+No picture with this one — it is a dead end worth recording. Stray's main menu draws a full 3D scene
+into prosper's own render targets and then presents only the UI bar over black, and the tempting
+explanation was that the buffer the game flips to the screen and the buffers we render into are two
+names for the same memory, which we were simply failing to connect. They are not: measured directly,
+they sit half a gigabyte apart in physical memory, across 80,511 resolutions with nothing left
+unresolved. So the game really is copying its finished frame somewhere else, by a route we do not yet
+follow — which is a narrower question than the one we started with.
+
 ### RenderDoc works on prosper now, and the first capture was a beautiful picture of the wrong thing
 
 No picture — or rather 56 of them, all useless. RenderDoc has been installed the whole time and
--- a/BLOG.md
+++ b/BLOG.md
@@ -27,9 +27,9 @@ No picture with this one — it is a dead end worth recording. Stray's main menu
 into prosper's own render targets and then presents only the UI bar over black, and the tempting
 explanation was that the buffer the game flips to the screen and the buffers we render into are two
 names for the same memory, which we were simply failing to connect. They are not: measured directly,
-they sit half a gigabyte apart in physical memory, across 80,511 resolutions with nothing left
-unresolved. So the game really is copying its finished frame somewhere else, by a route we do not yet
-follow — which is a narrower question than the one we started with.
+the two sit half a gigabyte apart in physical memory. So the game really is copying its finished
+frame somewhere else, by a route we do not yet follow — which is a narrower question than the one we
+started with.
 
 ### RenderDoc works on prosper now, and the first capture was a beautiful picture of the wrong thing
 
--- a/BLOG.md
+++ b/BLOG.md
@@ -21,15 +21,16 @@ from the tracker issues, and still gated, because it is a projection of state ra
 
 ## 2026-09-04
 
-### Stray renders its title screen straight into the display buffer, and the world still is not there
+### Stray's world is not going missing on the way to the screen — it arrives, and rounds to black
 
-No picture worth showing — the frame is the menu over black, which is what it has been. The finding
-is what sits behind it. We had assumed the game composed its picture in one place and copied it to
-the buffer the screen reads, and that prosper was losing the copy. Measuring every render target this
-time, rather than only the first one, says otherwise: the two display buffers *are* render targets,
-and prosper draws into them nearly a thousand times in a run that still shows an empty screen. So
-nothing is being lost in transit. Either only the menu is ever drawn there, or the world is drawn and
-then wiped — and those two are quick to tell apart.
+No picture worth showing: the frame is still the menu over black. The finding is what sits behind it.
+We had assumed the game built its picture in one place and copied it to the buffer the screen reads,
+and that we were losing the copy. Measuring every render target this time, rather than only the
+first, says otherwise — the two display buffers are themselves render targets, attached to passes
+that carry thousands of draws. Nothing is being lost in transit. What the existing pixel traces
+already showed, and what this reframes, is that the world's colour reaches the end of the pipeline so
+faint that storing it rounds it to zero. So the question stops being "where did the picture go" and
+becomes "why is it arriving that dim", which is a much better question to be stuck on.
 
 ### RenderDoc works on prosper now, and the first capture was a beautiful picture of the wrong thing
 
--- a/BLOG.md
+++ b/BLOG.md
@@ -27,7 +27,7 @@ No picture worth showing: the frame is still the menu over black. The finding is
 We had assumed the game built its picture in one place and copied it to the buffer the screen reads,
 and that we were losing the copy. Measuring every render target this time, rather than only the
 first, says otherwise — the two display buffers are themselves render targets, attached to passes
-that carry thousands of draws. Nothing is being lost in transit. What the existing pixel traces
+that carry draws. Nothing is being lost in transit. What the existing pixel traces
 already showed, and what this reframes, is that the world's colour reaches the end of the pipeline so
 faint that storing it rounds it to zero. So the question stops being "where did the picture go" and
 becomes "why is it arriving that dim", which is a much better question to be stuck on.
--- a/BLOG.md
+++ b/BLOG.md
@@ -21,15 +21,15 @@ from the tracker issues, and still gated, because it is a projection of state ra
 
 ## 2026-09-04
 
-### Stray's missing world is not an address-matching bug, and now we can prove it
+### Stray renders its title screen straight into the display buffer, and the world still is not there
 
-No picture with this one — it is a dead end worth recording. Stray's main menu draws a full 3D scene
-into prosper's own render targets and then presents only the UI bar over black, and the tempting
-explanation was that the buffer the game flips to the screen and the buffers we render into are two
-names for the same memory, which we were simply failing to connect. They are not: measured directly,
-the two sit half a gigabyte apart in physical memory. So the game really is copying its finished
-frame somewhere else, by a route we do not yet follow — which is a narrower question than the one we
-started with.
+No picture worth showing — the frame is the menu over black, which is what it has been. The finding
+is what sits behind it. We had assumed the game composed its picture in one place and copied it to
+the buffer the screen reads, and that prosper was losing the copy. Measuring every render target this
+time, rather than only the first one, says otherwise: the two display buffers *are* render targets,
+and prosper draws into them nearly a thousand times in a run that still shows an empty screen. So
+nothing is being lost in transit. Either only the menu is ever drawn there, or the world is drawn and
+then wiped — and those two are quick to tell apart.
 
 ### RenderDoc works on prosper now, and the first capture was a beautiful picture of the wrong thing
 
--- a/prosper/docs/STRAY_STATUS.md
+++ b/prosper/docs/STRAY_STATUS.md
@@ -1447,51 +1447,78 @@ evidence nor falsifications for it. Re-run each with `PROSPER_NULL_PAGE=1` and a
   shifted right by that same `v5` — so `Lod=0` remains the wrong answer. #3134, #2818.
 - **"The scanout buffer and the render targets are two virtual mappings of ONE physical allocation,
   so the VA-keyed present match is the whole defect."** Falsified across **every colour slot**, and
-  the run that falsified it also **falsified the premise the hypothesis was built on** — which is the
-  more useful half, so it is stated first.
+  the run that falsified it also **corrected the premise the hypothesis was built on**.
 
-  **prosper renders directly INTO the scanout buffers.** Both flipped VAs appear in the colour-target
-  census as ordinary 4K slot-0 persistent render targets, throughout the failing run:
+  **The scanout buffers are render targets.** Both flipped VAs appear in the colour-target census as
+  4K slot-0 persistent colour attachments, throughout the failing run (title route, steady state
+  confirmed by eye: menu over black at 8x brightness):
 
-  | target VA | physical | size | times seen |
-  | --- | --- | --- | --- |
-  | `0x9fc0000000` | `0x230000` | 3840x2160 | 946 |
-  | `0x9fc2000000` | `0x2210000` | 3840x2160 | 976 |
+  | flipped VA | physical | size | census lines | draws dropped `no-effect(early)` |
+  | --- | --- | --- | --- | --- |
+  | `0x9fc0000000` | `0x230000` | 3840x2160 | 946 | **1,143** |
+  | `0x9fc2000000` | `0x2210000` | 3840x2160 | 976 | **1,154** |
 
-  So the framing this row was originally written to support — *"prosper renders into `0x30…` while
-  the guest flips `0x9fc…`, ~450 GiB apart, and no VA-keyed lookup can connect them"* — **is wrong**.
-  The present path's VA-keyed lookup can see these buffers perfectly well, because they are live
-  render targets under their own VAs. Do not repeat that sentence; it is why this row exists.
+  **Read the census column precisely.** The line is printed after the empty-draw early-out, so it
+  means *this VA was bound as slot 0 of a pass carrying at least one draw*. It does **not** mean a
+  draw's output reached memory — masks, discards and store behaviour are all downstream. So the
+  correct statement is "the scanout buffers are attachments of passes that carry draws", not "prosper
+  renders into them".
 
-  **And the alias hypothesis is dead anyway**, now over the whole population rather than one slot.
-  Census over slots 0-6 (slot 7 unused): **55,868 resolutions, 0 unresolved, 82 distinct targets** —
-  80 in the `0x30…` family and the 2 scanout buffers above. Excluding the scanout buffers themselves,
-  the number of targets whose *physical extent* overlaps either scanout buffer is **0**, and the
-  nearest non-scanout target sits **390 MB** above them. Nothing is aliased.
+  So the framing this row originally supported — *"prosper renders into `0x30…` while the guest flips
+  `0x9fc…`, ~450 GiB apart, so no VA-keyed lookup can connect them"* — **is withdrawn**. Those VAs
+  are not a region the renderer never touches.
 
-  The slot coverage is load-bearing and was added after review: **24% of resolutions (13,669) are in
-  slots 1-6**, which a slot-0-only census cannot see. That matters because `is_live_render_target`
-  consults `g_rtt`, and `g_rtt` is populated for every slot, so a slot-1..7 target aliasing the
-  scanout page would have been exactly the link the hypothesis proposed — in exactly the region the
-  narrower instrument was blind to.
+  **But do not swap it for the opposite error.** "The present path can therefore see them" is *also*
+  unmeasured. `guest_scanout_publishable` tests `renderer_owns_target` (a `g_rtt` hit on the flipped
+  VA) **first**, short-circuiting ahead of `guest_authored`, so a decline of `SkipNotAuthored` is
+  itself evidence that `g_rtt` did **not** hold the VA at that flip. The census (pass-render time) and
+  the publish check (flip time) are different moments and must not be read as one. The probe now
+  prints `rtt_owns=` and the decision name beside the physical address so this becomes a measurement.
 
-  **What this leaves is a better question than the one it replaces.** prosper renders into
-  `0x9fc0000000` 946 times in a run whose steady state is the menu over black (visually confirmed:
-  `START GAME` / `SETTINGS` / `CREDITS` and the version string, everything else black at 8x
-  brightness). So the missing world is **not** a copy that never happens. Either the draws that reach
-  the scanout target carry only the UI, or the scene reaches it and is then cleared or overwritten.
-  That is answerable with the tools already in the tree — a PixelHistory on a world pixel of the
-  scanout target, which separates those two directly (see the note at the top of this file about
-  finding the brightest pixel rather than the centre).
+  **TWO REGIMES, and this row previously blurred them — the same mistake this file's own banner warns
+  about for `max_nonblack`.** They are different failure states and their evidence does not transfer:
 
-  It also removes a puzzle rather than adding one: if prosper renders into a *VkImage* keyed by the
-  scanout VA, guest memory at that VA need never be written, which is consistent with
-  `videoout_buffer_authored_locked` reporting `authored=0` without anything being wrong at all.
+  | regime | what publishes | where the `authored=0` / `SkipNotAuthored` evidence comes from |
+  | --- | --- | --- |
+  | **title screen** (measured here) | the menu **does** publish | *not this regime* |
+  | **main menu / first map** | nothing publishes (`fresh=0 retained=0`) | this one |
 
-  **Caveats a later reader needs.** `guest_write_watch_va_to_phys` returns the FIRST range containing
-  the address and does not check it is large enough for the access — fine for a probe, not for a
-  production caller. And `authored=0` still has four causes, not one; the earlier draft of this row
-  read it as "the bytes are unchanged since registration" and leaned on "a GPU copy would flip
-  `authored` by itself", which does not hold: `!buffer_digest_valid[slot]` means no baseline was ever
-  seeded, and a write covering none of the digest's sample points misses too, so **two** of the four
-  branches are a blind instrument rather than an absent write. #2932, #3126.
+  The guest-scanout fallback is gated on `guest_scanout_read_warranted(...) == Publish`, i.e. it runs
+  **only when nothing else published**. On the title route it is therefore never entered — measured:
+  **0** `[rtt] GUEST SCANOUT` lines across a full run that reached and held the failing steady state.
+  So `rtt_owns=` cannot be collected on this route at all, and every `authored=0` / `px_front=none`
+  fact in this section belongs to the *other* regime. Quote them with the regime attached.
+
+  **The alias hypothesis is dead**, now over the population rather than one slot: **55,868
+  resolutions, 0 unresolved, 82 distinct targets** across slots 0-6. Excluding the scanout buffers
+  themselves, targets whose *physical extent* overlaps either scanout buffer: **0**; nearest is
+  **390 MB** away. The slot coverage was added after review and earned its keep — **24% of
+  resolutions (13,669) are in slots 1-6**, and since `is_live_render_target` consults `g_rtt`, which
+  is populated for every slot, an alias there would have been exactly the proposed link, in exactly
+  the region a slot-0 census cannot see. (One gap remains in "population": a *declined* MRT1 base
+  never sets `persistent_id1`, so it cannot reach the census.)
+
+  **What is left is a three-way question, and this file already answers part of it.** The two
+  tempting branches are "only the UI is ever drawn there" and "the scene is drawn and then wiped".
+  Neither is the branch already measured above: the scene **reaches** the target and is **stored as
+  black by magnitude** — PixelHistory has tonemap `shaderOut 0.001099` storing `0`, because
+  `0.001099 x 255 = 0.28` rounds to zero in `R8G8B8A8_UNORM`, and § *This is not a composite defect*
+  records that the draws run and write nothing. An A/B between the first two branches would exclude
+  the answer already in hand. The open part is why the value arriving at the store is that small, not
+  where the picture goes.
+
+  The strangest remaining fact, with its regime attached: **in the non-publishing regime**
+  `select_present_source` returns `None` (`px_front=none px_vo=none px_last=none`, `fresh=0
+  retained=0`) for buffers that — on the title route at least — are attachments of drawing passes.
+  Whether that also holds of the flipped buffer *in that regime* is untested, because the census and
+  those counters have not yet been collected in one run. That is the cheap next measurement.
+
+  This also dissolves a puzzle instead of deepening it: if prosper renders into a `VkImage` keyed by
+  the scanout VA, guest memory at that VA need never be written, which is consistent with
+  `videoout_buffer_authored_locked` reporting `authored=0` with nothing wrong at that point at all.
+
+  **Caveats.** `guest_write_watch_va_to_phys` returns the FIRST range containing the address without
+  checking it is large enough — fine for a probe, not for a production caller. And `authored=0` has
+  four causes, of which **two** are a blind instrument rather than an absent write
+  (`!buffer_digest_valid[slot]`, i.e. no baseline seeded; and a write covering none of the digest's
+  sample points), so `buffer_digest_valid` alone does not separate them. #2932, #3126.
--- a/prosper/docs/STRAY_STATUS.md
+++ b/prosper/docs/STRAY_STATUS.md
@@ -1458,18 +1458,38 @@ evidence nor falsifications for it. Re-run each with `PROSPER_NULL_PAGE=1` and a
   | lowest render target | `0x30…` | `0x1e980000` (511 MB) |
   | highest render target | `0x30…` | `0x79f20000` (2.0 GB) |
 
-  80,511 target resolutions, **0 unresolved** — so the census saw the whole population rather than a
-  sample that happened to miss the alias. The scanout sits ~509 MB *below* every render target. They
-  are separate physical allocations and nothing is aliased.
+  80,511 target resolutions, **0 unresolved** — every VA the census was shown did resolve, so the
+  answers above are real rather than a resolver failing quietly. The scanout sits ~509 MB *below*
+  every render target: separate physical allocations, nothing aliased.
+
+  **What that number does NOT say is "the whole population".** The census reads
+  `color_target->persistent_id` — colour **slot 0 only**. `persistent_id1` and
+  `persistent_id_slots[2..7]` are separate guest VAs and are never censused, and 80,511 counts
+  invocations rather than distinct addresses. An alias hiding in an un-censused MRT slot is exactly
+  what the hypothesis predicts and exactly what this instrument cannot see, so the falsification is
+  **scoped to slot 0** — which is the slot the composite is built in, and enough to stop the
+  VA-keyed-match story, but not a proof about every surface the frame renders into.
 
   **What this leaves is narrower and better posed.** The guest composites into its own allocation and
   moves it to scanout by a route prosper does not model, so the open question is *which packet does
   that*, not how the present path matches addresses. Two further facts bound it: `select_present_source`
   returns `None` (`px_front=none px_vo=none px_last=none`, `fresh=0 retained=0`), and
-  `videoout_buffer_authored_locked` reports `authored=0` for the flipped buffer — its bytes are
-  unchanged since registration. That second one is the load-bearing constraint, because the digest test
-  reads guest memory directly: **a GPU copy landing in guest memory would flip `authored` on its own.**
-  So the copy is not happening, not landing there, or not being executed by prosper at all.
+  `videoout_buffer_authored_locked` reports `authored=0` for the flipped buffer.
+
+  **Read that second one carefully — it is weaker than it first looks, and an earlier draft of this
+  row overstated it.** The tempting reading is "the bytes are unchanged since registration, and since
+  the digest test reads guest memory directly, a GPU copy landing there would have flipped `authored`
+  by itself". The predicate has **four** ways to return false (`hle_graphics.cpp`), and one of them
+  breaks that inference: `!buffer_digest_valid[slot]`, i.e. **no baseline was ever seeded** — which is
+  what happens when the buffer was not readable at registration time, and the seeding path documents
+  the resulting state as *deliberately indistinguishable from "not written"*. With no baseline, a copy
+  of any size leaves `authored` at 0.
+
+  So the honest disjunction has a fourth branch: the copy is **not happening**, **not landing there**,
+  **not executed by prosper at all**, or **it happened and prosper cannot tell**. Separating the last
+  one from the first three is cheap — check `buffer_digest_valid` for the slot — and should be the
+  first step for whoever picks this up, because three of the four branches are a missing packet and
+  the fourth is a blind instrument.
 
   **Two cautions for whoever picks this up.** Stray registers **two** scanout buffers,
   `0x9fc0000000` and `0x9fc2000000` — a rotating set, which is the shape behind *Bendy and the Ink
--- a/prosper/docs/STRAY_STATUS.md
+++ b/prosper/docs/STRAY_STATUS.md
@@ -1509,12 +1509,22 @@ evidence nor falsifications for it. Re-run each with `PROSPER_NULL_PAGE=1` and a
   the failing steady state — and that is honestly reported as **never reached**, not as evidence of
   what published. A draft of this row inferred "the menu therefore publishes" from the zero count;
   that inference is withdrawn, having violated a precedent written into the header it reasoned about.
-  **The publishing claim rests on the frame instead** — and the artifact has to be named, because
-  which buffer was dumped is the whole strength of it. These are `tools/screenshot` PNGs, which come
-  from `present_readback`, which returns what `present_write_frame` wrote
-  (`gpu_executor.cpp:11780`/`:11927`). So a menu-bearing PNG is direct evidence that **a frame was
-  published to the present layer** — the publication itself, not an inference from a decline that
-  never printed.
+  **The publishing claim rests on the frame instead** — and it rests on the frame's RECORDED SOURCE,
+  not on the image. `tools/screenshot` captures through `present_snapshot`, which **falls back to the
+  guest's own display buffer** when the renderer published nothing (`videoout_present.cpp:187-196`,
+  returning `RawScanout`), and saves it by default (`render_first` defaults to 0). So a menu-bearing
+  PNG on its own proves nothing: the guest composites its menu into `0x9fc0000000` itself, and that
+  image would look identical.
+
+  The discriminator is recorded per capture, so no re-run is needed. Across both title-route runs the
+  manifest reports **29 of 30 captures as `composited`** (`CaptureSource::Rendered`) with only index 0
+  — the all-black first frame — as `raw_scanout`. The steady-state menu frame examined by eye is
+  `composited`. **That** is the evidence a frame was published: prosper composited it, rather than the
+  tool echoing back guest memory.
+
+  (Do not cite `present_readback` here, as a draft of this row did. `tools/screenshot` never calls it
+  — `screenshot.cpp:5`'s comment says otherwise and is stale — and it carries the same fallback
+  hazard at `videoout_present.cpp:132-138` regardless.)
 
   Do not substitute the renderer's own `frame_%04d.bmp` for this without collecting it: that dump is
   fed by `selected_pixels` (`live_renderer.cpp:10160` → `:10184`), so it would prove
--- a/prosper/docs/STRAY_STATUS.md
+++ b/prosper/docs/STRAY_STATUS.md
@@ -1445,3 +1445,36 @@ evidence nor falsifications for it. Re-run each with `PROSPER_NULL_PAGE=1` and a
   nsa=1 materialized_mips=1 compute=0`. The disassembly is unambiguous that the level is genuinely
   dynamic — `v_min_i32 / v_max_i32` clamp a log2-derived LOD into `v5`, and the x/y coordinates are
   shifted right by that same `v5` — so `Lod=0` remains the wrong answer. #3134, #2818.
+- **"The scanout buffer and the render targets are two virtual mappings of ONE physical allocation,
+  so the VA-keyed present match is the whole defect."** Falsified by direct resolution
+  (`PROSPER_TARGET_PHYS=1` plus the unconditional `[rtt] GUEST SCANOUT #N phys:` probe, both landed
+  with this row). This was the most attractive hypothesis available, because it would have made a
+  measured fact — prosper renders into `0x30…` while the guest flips `0x9fc0000000`, ~450 GiB apart
+  in VA — into a *bookkeeping* error with a fix prosper already had the machinery for. It is not:
+
+  | buffer | guest VA | physical |
+  | --- | --- | --- |
+  | flipped scanout | `0x9fc0000000` | **`0x230000`** (2.3 MB) |
+  | lowest render target | `0x30…` | `0x1e980000` (511 MB) |
+  | highest render target | `0x30…` | `0x79f20000` (2.0 GB) |
+
+  80,511 target resolutions, **0 unresolved** — so the census saw the whole population rather than a
+  sample that happened to miss the alias. The scanout sits ~509 MB *below* every render target. They
+  are separate physical allocations and nothing is aliased.
+
+  **What this leaves is narrower and better posed.** The guest composites into its own allocation and
+  moves it to scanout by a route prosper does not model, so the open question is *which packet does
+  that*, not how the present path matches addresses. Two further facts bound it: `select_present_source`
+  returns `None` (`px_front=none px_vo=none px_last=none`, `fresh=0 retained=0`), and
+  `videoout_buffer_authored_locked` reports `authored=0` for the flipped buffer — its bytes are
+  unchanged since registration. That second one is the load-bearing constraint, because the digest test
+  reads guest memory directly: **a GPU copy landing in guest memory would flip `authored` on its own.**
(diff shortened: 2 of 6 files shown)


mattias800 (commented):
**REJECTED** — one blocking finding, in the B3 rewrite you asked me to scrutinise. Everything else is
resolved, and B1 is independently confirmed by preprocessor rather than by eye.

This is a much better artifact than the one I reviewed on #3331. B2's rewrite in particular is a model
of what a `## Ruled out` row should do: it names the earlier overstatement, gives the four-way
disjunction, and points at `buffer_digest_valid` as the cheap discriminator. The one thing left is that
B3's rewrite admits the slot-0 scope and then, one clause later, quietly spends it.

---

## Blocking

### The slot-0 scoping paragraph re-generalises through two unmeasured clauses

`prosper/docs/STRAY_STATUS.md:1468-1471`:

> An alias hiding in an un-censused MRT slot is exactly what the hypothesis predicts and exactly what
> this instrument cannot see, so the falsification is **scoped to slot 0** — **which is the slot the
> composite is built in, and enough to stop the VA-keyed-match story**, but not a proof about every
> surface the frame renders into.

The first two-thirds of that sentence is exactly right. The bolded clause takes it back, and it rests on
two things neither this PR nor the doc measures:

**(1) "which is the slot the composite is built in" is asserted, not measured.** Nothing in this PR
shows it — the census prints slot-0 VAs, which says where slot 0 is, not which slot the composite
targets. Nor does the doc establish it elsewhere: the composite evidence at `STRAY_STATUS.md:975` is a
`PixelHistory` row (EID 3788, stored `0`), which fixes the *pass* but not the *attachment slot*. And
your own renderer records that the assumption is not structural —
`live_renderer.cpp:8696-8699` says a pass with **colour-0 unbound and a higher slot bound is reachable
by construction**, which is precisely why `persistent_id` can be 0 and the census skip a pass entirely.

**(2) "enough to stop the VA-keyed-match story" does not follow, and the present path says why.** The
VA-keyed match is not slot-0-scoped. `is_live_render_target` (`gpu_executor.cpp:11388`) consults
`g_rtt`, and `g_rtt` is populated for **every** slot: `live_renderer.cpp:8990` (slot 0),
`:9035` (slot 0, deferred-readback path) and **`:9100` — `RttSurf& surface = g_rtt[pass_bases[slot]];`
for slots 2..7**. So a slot-1..7 target whose physical range covered `0x230000` would be a genuine live
render target that the present path already knows about by VA, and it is exactly the missing link the
hypothesis proposes. The census cannot see those slots. The story is therefore **not** stopped — it is
alive in precisely the region the paragraph just finished admitting is unobserved.

Failure scenario, concretely: the next agent reads this row, takes "enough to stop the VA-keyed-match
story" as settled (it is bolded-adjacent, in a `## Ruled out` section, and immediately follows an honest
scope admission, which is what makes it credible), and never censuses slots 1..7. If Stray's composite
lands in MRT1 — which `live_renderer.cpp:8696` says is constructible — the live hypothesis has been
retired by a row that explicitly said it was not proving it. That is the charter's "a true statement
reached by a route that does not establish it", and the asymmetry it warns about applies here: this is a
clause that *raises* the result, so it gets checked least.

**Two remedies, either is fine.**
- *Cheap:* delete the clause. "…so the falsification is **scoped to slot 0**, and an alias in an
  un-censused MRT slot remains open" is accurate, still valuable, and still kills the row's headline
  hypothesis for the slot that was measured.
- *Better, and you already have the machinery:* extend the census to `persistent_id1` (`:5286`) and
  `persistent_id_slots[2..7]` (`:5388`) — three more call sites of the resolver you just landed — and
  re-run. Then the clause is true as written and the row needs no hedge at all. Given the census is
  already wired and gated, this looks like a ~10-line change and it converts a scoped result into the
  complete one.

To be explicit, since you asked about over-correction in both directions: **the rest of the B3 rewrite
does not understate anything.** "every VA the census was shown did resolve, so the answers above are
real rather than a resolver failing quietly" is exactly what `0 unresolved` licenses, and "80,511 counts
invocations rather than distinct addresses" is the right caveat. The slot-0 result is a real
falsification and the row is right to present it as one. Only the one clause reaches past it.

---

## Non-blocking

1. **`STRAY_STATUS.md:1462` still says "every render target" unqualified** — "The scanout sits ~509 MB
   *below* every render target: separate physical allocations, nothing aliased" — and the table rows
   above it say "lowest/highest render target". The very next paragraph corrects this loudly, so it is
   not a blocker, but the sentence and the two table rows are what a reader quotes out of context.
   Suggest "every censused (slot-0) render target" and "lowest/highest slot-0 render target".

2. **`buffer_digest_valid` alone does not fully separate branch 4** (`STRAY_STATUS.md:1483-1485`). The
   row says separating "it happened and prosper cannot tell" from the first three is cheap — check
   `buffer_digest_valid` for the slot. That narrows it but does not close it: `hle_graphics.cpp:376`
   records a second way to land in the same branch — *"the guest writes only a region no sample point
   covers -> 'not authored' -> decline"* — because the digest samples `kDigestSamples = 4096` evenly
   spaced 8-byte words rather than hashing the surface. A reader who finds `buffer_digest_valid == 1`
   will conclude branch 4 is excluded when a partial copy could still be invisible. One clause fixes it:
   "…and, if the baseline is valid, note that a copy touching no sample point is also invisible — the
   comment at `hle_graphics.cpp:376`; a full-surface composite would move essentially every sample, a
   partial blit need not."

3. **`BLOG.md`** now says "the two sit half a gigabyte apart in physical memory" against "the buffers we
   render into" (plural, general). It is the right length and the picture-free exception applies, but
   the entry is now the only place the result appears without its slot-0 scope. Three words
   ("the buffers we render into — the main one, at least") would keep it honest without making it a
   report.

4. **`live_renderer.cpp:9970-9972` takes the mutex twice**, once in `guest_write_watch_va_to_phys` and
   again in `guest_write_watch_alias_count`, so the printed `aliases=N` is not atomically the domain
   that was searched. Harmless for a probe and not worth a lock-holding variant; noting it only because
   the count is now load-bearing for the reading.

---

## Verified against source

**B1 — confirmed independently, and by machine rather than by eye.** You asked me to check the
preprocessor-depth conclusion because you cannot build Windows locally. I did not re-derive your scan; I
ran the actual preprocessor. Method: rewrite `guest_write_watch.cpp` into a skeleton that keeps every
`#if/#ifdef/#ifndef/#elif/#else/#endif` verbatim and replaces every other line with a `LINE_<n>` marker,
then run `cpp -undef -P` with one platform macro defined at a time and check which markers survive.

```
WINDOWS (-D_WIN32=1)     line 544 PRESENT   545 PRESENT   1152 compiled out  1175 compiled out
LINUX   (-D__linux__=1)  line 544 out       545 out       1152 PRESENT       1175 PRESENT
MACOS   (-D__APPLE__=1)  line 544 out       545 out       1152 PRESENT       1175 PRESENT
```

Exactly one definition of each symbol survives in each arm — no missing definition, no duplicate. The
scan is sound for this file specifically: it contains only four conditional directives (`:71`, `:73`,
`:1429`, `:1641`) and one `#define` (`WIN32_LEAN_AND_MEAN` at `:74`), which is self-guarded, sits inside
the Windows arm, and guards nothing but includes — so no conditional depends on an in-file macro my
skeleton stripped. Also checked: the stubs at `:544-545` are inside `namespace prosper::host` (closed at
`:548`); `grep` finds no third definition anywhere in `src/ frontends/ tests/ tools/`; both call sites
initialise their out-parameter before the call, so the Windows stub's untouched `phys&` is never read;
`<cstddef>` is included at `guest_write_watch.hpp:4` for the `size_t` return; `%zu` is UCRT-safe.

The Windows stub's comment is also the right comment — it explains *why* zero is structural there rather
than asserting a measurement, which is the same distinction B4 was about.

**B2 — resolved, and well.** `hle_graphics.cpp:478-489` does have four false-returns, `:483` is
`!buffer_digest_valid[slot]`, and `videoout_present.hpp:58-60` is the "deliberately indistinguishable"
contract. The row now states all four, names the discriminator, and flags that three branches are a
missing packet while the fourth is a blind instrument. That is a better row than the one my finding
asked for. (See non-blocking 2 for the one refinement.)

**B4 — resolved.** `guest_write_watch_alias_count()` (`:1175`) returns the searched domain under the
same lock; both probes print it; `live_renderer.cpp:9979-9981` distinguishes "no alias range covers this
VA" from "alias table EMPTY -- nothing was searched"; `render_runner.h:5459-5462` prints
`(aliases=%zu)` on the miss path only, so the resolved path pays no second lock. The header contract at
`:233-236` now *requires* callers to report the count beside a false, which is the durable form of the
fix.

**All six non-blocking items from #3331 are addressed.** `PROSPER_SCANOUT_PHYS` → `PROSPER_TARGET_PHYS`
(`hpp:219`); both "dmem trace" misattributions corrected to the alias table (`hpp:219-221`, `cpp:1148`)
with the header now explicitly contrasting the two; the dangling arrow and split `fprintf` merged into
one call per branch; `va=0x0` guarded at `live_renderer.cpp:9967` (`diag_should_print` is pure, so
short-circuiting loses no state); the declaration moved below the `notify_*` family (`hpp:218`).

**Still good from last time, re-checked on this head:** locking (no `_locked` suffix, both new functions
take `w.mutex` themselves; both call sites external, no re-entrant path); containment arithmetic cannot
underflow (`addr < a.addr` precedes the subtraction); first-match is safe given the add/remove purge
paths; the `render_runner.h:6` include resolves via `target_include_directories(prosper_core PUBLIC src)`;
`persistent_id` really is a guest VA; format specifiers match; `check_diag_gates.py` needs no new
baseline entry.

**Mechanical, on this head (83afae62f):** `git diff --check` rc=0; `git diff --numstat` shows **0
deletions in all six files**; `check_numbered_table.py prosper/docs/STRAY_STATUS.md` → 25 tables, 93
rows, unbroken, rc=0. I did not build or run tests, as instructed.

---

Fix the one clause — deleting it is enough — or census the other slots and earn it. Everything else here
is ready.

mattias800 (commented):
**REJECTED** — three blocking findings, all of them the "does it overreach in the new direction?"
question you asked me to press on. Your instinct was right: the census licenses less than the row spends,
and in one place the run's own decision path contradicts the row directly.

Process point first: **the outside commit was not me.** On both #3331 and #3332 I have only ever run
`gh pr review` and one `gh pr comment`. I have never run `git commit`, `git push`, or any write against
your branches or the ref, and I will not.

Widening the census was the right call and the 24% figure (13,669 of 55,868 in slots 1-6) shows it was
not a formality. Correcting your own premise in the row is the most valuable thing in this diff.

---

## Blocking

### B1. Three code comments repeat verbatim the sentence the row says never to repeat

`STRAY_STATUS.md:1462-1464` retracts the founding premise and says, in as many words:

> …*"prosper renders into `0x30…` while the guest flips `0x9fc…`, ~450 GiB apart, and no VA-keyed
> lookup can connect them"* — **is wrong**. … **Do not repeat that sentence; it is why this row exists.**

The same PR ships that sentence three times, as present-tense fact:

- `src/host/memory/guest_write_watch.hpp:225` — *"Stray renders into 0x30... and flips 0x9fc0000000 --
  ~450 GiB apart -- so no VA-keyed lookup can connect them"*. This is the **header contract** of a shared
  substrate function.
- `frontends/shared/live/live_renderer.cpp:9957-9958` — *"this title renders into 0x30... while flipping
  0x9fc0000000 -- ~450 GiB apart, so no VA-keyed lookup can connect them"*. This is the comment on the
  probe **that produced the correction**.
- `tests/fixtures/render_runner.h:5440` — *"Stray renders into 0x30... while flipping 0x9fc0000000"*.
  This is the comment on the **census that falsified it**.

Failure scenario: the next engineer opens `guest_write_watch.hpp` to use the resolver, reads the founding
premise stated as fact in the contract, and never reaches `STRAY_STATUS.md`. This is the #2049 shape the
charter records at length — a wrong statement in a shipped code comment outran the correction because the
comment is what people read first. Here the correction and the wrong statement are landing in the *same
commit*, which is strictly worse: there is no window in which the comment was right.

Remedy: rewrite all three to describe what the resolver does and what was actually found. The header
comment in particular should not carry a title-specific narrative at all — one line of contract plus a
pointer to `docs/STRAY_STATUS.md` would age better.

### B2. "The present path's VA-keyed lookup can see these buffers perfectly well" is contradicted by the run's own decision path

`STRAY_STATUS.md:1463-1464`. Two independent problems.

**(a) The decision path says the opposite at flip time.** `guest_scanout_publishable`
(`frontends/shared/present/guest_scanout_present.hpp:88-108`) tests in order:

```
:100   if (renderer_owns_target) return SkipRendererOwnsTarget;
:107   if (!guest_authored)      return SkipNotAuthored;
```

and the caller passes `g_rtt.find(read.metadata.address) != g_rtt.end()` as `renderer_owns_target`. The
row leans on `authored=0` as an observation (`:1487-1491`), and #3331's body reported the decision as
`SkipNotAuthored`. **`SkipNotAuthored` is only reachable past the `renderer_owns_target` early return** —
so at those flips `g_rtt` did *not* contain the flipped VA, and the present path could *not* see it. That
is a live, in-run measurement from this very code path, and it is in direct tension with the row's claim.

Both can be true at different instants — `g_rtt` entries are conditional (`:8990` needs
`color_target_call.writes`, `:9035` needs non-empty pass pixels) and `drain_guest_gpu_writes` runs on
every query — but the row asserts the strong reading as settled and never reconciles it. If the decision
was in fact `SkipRendererOwnsTarget`, the row is right and my objection evaporates — which is exactly why
**the row must quote the decision name**. It currently does not.

**(b) The census does not sample what `is_live_render_target` actually tests.**
`is_live_render_target` (`src/gpu/execute/gpu_executor.cpp:11388`) delegates to the query registered at
`live_renderer.cpp:1373-1387`, which requires a `g_rtt` hit **and**
`live_rtt_compute_authoritative(surface.gpu_valid, has_cpu_snapshot)`. The census reads
`BackendColorTarget::persistent_id` — it never touches `g_rtt`, `gpu_valid`, or the CPU snapshot. So "they
are live render targets under their own VAs" is an inference from a different observable, not a
measurement.

Remedy, and it is nearly free: the scanout probe already runs **at the exact moment in question**. Have it
print `g_rtt.count(read.metadata.address)` and the decision name beside the phys line. That converts the
row's central new claim from inference to measurement in one run.

### B3. "Renders INTO" overreaches the census, and the disjunction that follows is not exhaustive

`STRAY_STATUS.md:1453` and `:1479-1482`.

You asked exactly the right question, and the answer is between your two options. The census fires *after*
`if (draws.empty()) return out;` (`tests/fixtures/render_runner.h:4969`), so it witnesses more than "the
target exists": **a colour target keyed at that VA was bound as slot 0 of a backend pass carrying at least
one draw, 946 times, at 3840x2160.** That is a real and useful fact.

What it does not witness is any draw's colour output *landing* in that attachment. Colour writes can be
mask-disabled, discarded, scissored out, or depth-only. And this is not hypothetical for this title — your
own document says so twice:

- `STRAY_STATUS.md:860` — *"**This is not a composite defect.** The draws run and write nothing — the same
  family as *Grand Theft Auto V*."*
- `STRAY_STATUS.md:974-976` — PixelHistory: composite EID 3788, `shaderOut 0,0,0,1`, stored `0`; tonemap
  EID 3752, `shaderOut 0.001099`, stored **`0`**, with the surrounding text concluding *"The shader is not
  producing black; the store is"* — `0.001099 x 255 = 0.28` rounding to zero in `R8G8B8A8_UNORM`.

So the row's binary — *"Either the draws that reach the scanout target carry only the UI, or the scene
reaches it and is then cleared or overwritten"* — omits the branch this document has already measured:
**the scene reaches it, is stored, and is black because its magnitude is ~40-100x low.** That is not a
pedantic third option; it is the reading the rest of the file supports, and a `## Ruled out` row that poses
a two-way question will steer the next agent into an A/B whose arms both exclude the answer already in
hand.

Remedy: weaken `:1453` to what the census licenses — "prosper **binds** both scanout buffers as 4K slot-0
colour render targets, in passes carrying draws, throughout the run" — and make the disjunction
three-branch, with the third pointing at the PixelHistory table above rather than treating it as new.

---

## Answers to your three questions

**Q3 — the out-param refactor is clean. No caller can read an uninitialised count.**
`guest_write_watch.cpp:542-545` (Windows stub) writes `*alias_count = 0` before returning;
`:1152-1157` (POSIX) writes it under the lock **before** the `!addr` early return, so every return path has
set it. Both call sites also initialise their own `size_t alias_n = 0` before the call
(`live_renderer.cpp:9968`, `render_runner.h:5455`), so the value is defined even if a future arm forgets.
`grep` finds **no remaining reference** to the removed `guest_write_watch_alias_count` anywhere in `src/
frontends/ tests/ tools/`. Publishing the count under the same acquisition as the lookup is the right fix
and strictly better than what I asked for. One behavioural note, benign: `!addr` now takes the mutex where
it previously did not — both callers guard non-zero, so nothing changes.

**Q2 — the slot-1 dedup cannot hide a distinct slot-1 target.** It skips only when
`persistent_id1 == persistent_id`, i.e. slot 1 names the *same guest VA* as slot 0 — which `census_slot(0,
…)` has already resolved and printed. No VA is lost, only a duplicate line, and for the alias question a VA
needs reporting once. It also matches the existing MRT test at `render_runner.h:4815-4816`, so it will not
drift.

There is a narrower gap, but it is upstream of your dedup and I would call it non-blocking: `persistent_id1
= use_color1 ? base1 : 0` (`live_renderer.cpp:8882`) with `use_color1 = mrt_count > 1` (`:8592`), so a pass
holding a non-zero `base1` that prosper declines to bind — the `[rtt] skip MRT1 target=0x%llx` path at
`:8870` — hands the census a zero and that VA is never resolved. Given the row now says "82 distinct
targets" and "the whole population", worth one clause acknowledging that declined MRT1 bases are not in the
count.

**Q1 — see B3.** Short answer: the census proves *more* than "the target exists" and *less* than "prosper
renders into it". It proves the target was **bound as an attachment in a pass carrying draws**. Say that,
and the third disjunct follows naturally.

---

## Non-blocking

1. **The row drops `select_present_source` returning `None`**, which the previous draft carried and which
   the new finding makes *sharper*, not less relevant: if prosper is drawing into the very buffer the guest
   flips, "nothing is published" is now the strangest fact on the page. Worth one line so the next reader
   does not think it was resolved.
2. **`BLOG.md`** inherits B3's phrasing — "prosper draws into them nearly a thousand times" and the same
   two-way ending. It is the right length and the picture-free exception applies; it just needs whatever
   wording B3 settles on.
3. Declined-MRT1 bases are outside the census (see Q2 above).

---

## Verified good on this head (8ff47e019)

The census widening is correct: `census_slot` is applied to slot 0, the deduped slot 1, and
`persistent_id_slots[2..]` over the real array bound, each guarded by `if (!va) return;` so `va=0x0` cannot
reach the resolver. The gate is still one `static const bool` on `PROSPER_TARGET_PHYS`, absent by default.
The resolver's locking, containment arithmetic and first-match behaviour are unchanged and still correct.
B1 from round 1 (the `_WIN32` arm) is untouched by this diff and the new stub keeps the same shape, so the
preprocessor result I posted last round still holds — I re-checked that the Windows arm defines the new
three-argument signature and nothing else changed around it.

The `authored=0` caveat is now stronger than my round-2 finding asked for: it names **two** of the four
branches as a blind instrument, including the digest-sample-point route from `hle_graphics.cpp:376`. Good.

The insight at `:1484-1486` — that rendering into a `VkImage` keyed by the scanout VA means guest memory
there need never be written, which would explain `authored=0` with nothing wrong — is the best sentence in
the row, and it is correctly hedged as "consistent with" rather than asserted.

Mechanical: `git diff --check` rc=0; `git diff --numstat` shows **0 deletions across all six files**;
`check_numbered_table.py prosper/docs/STRAY_STATUS.md` rc=0. No build, no tests, as instructed.

---

B1 is mechanical. B2 needs one number from a run you can already produce. B3 is a rewording plus a third
disjunct that cites evidence you already have. The underlying result — the scanout buffers are colour
targets, and nothing is aliased across 82 distinct targets — is real and worth landing.

mattias800 (commented):
**REJECTED** — two blocking findings, both in the doc row, both about numbers rather than code. **The
code is done**: B1, B2 and B3 are all properly fixed, the deletion is clean, and I have no remaining
findings against any source file.

You asked me to attack the regime reasoning hardest. It does not hold — and following the same thread
into the new `no-effect(early)` column turned up a second, sharper instance of the same problem, in the
one place this round's regime discipline was not applied.

---

## Blocking

### B4. The two-regime table's central cell is derived from the one observation that cannot support it

`STRAY_STATUS.md:1483-1489`. Your inference is: fallback gated on
`guest_scanout_read_warranted(...) == Publish` → "runs only when nothing else published" → 0 lines on the
title route → "the menu **does** publish" → older facts belong to the other regime.

The middle step is wrong. Stage 1 (`frontends/shared/present/guest_scanout_present.hpp:80-83`) has
**four** declines, not one:

```
:80  if (published_gpu)                        return SkipPublishedGpu;
:81  if (renderer_scanout || have_selected_pixels) return SkipRendererSource;
:82  if (present_extent_bytes == 0)            return SkipNoPresentContract;
:83  if (display_bytes != present_extent_bytes) return SkipScaledPresent;
```

Only the first two mean "something else published". `SkipNoPresentContract` (no present extent —
gpu_replay / render_submit_items) and `SkipScaledPresent` (`PROSPER_RENDER_SCALE`, and the header notes
"most snapshot guards run scaled") are unrelated to publishing and produce an identical zero-line log.

And the header says so directly, at `:29-34`:

> **Only the stage-2 reasons are currently observable at runtime.** The renderer's `[rtt] GUEST SCANOUT`
> report sits inside the stage-1 `Publish` branch, so a stage-1 skip prints nothing and the four names
> below it … can never appear in a log as things stand. That is why **Bendy's (PPSA27616) zero-line
> control is honestly reported as "never reached" rather than as a decline**: the instrument cannot tell
> those apart today.

So there is a named precedent, in the file you are reasoning about, where this exact observation was
deliberately *not* attributed to a cause. The row attributes it.

**Direct answer to your question — you have probably labelled the regimes correctly, but not from this
evidence.** Your frame 20 exists and shows the menu, and a frame existing is decent evidence that
something published. That is the argument to make. The zero-line count is not, and it is the only one
the row offers. Charter shape: a true statement reached by a route that does not establish it.

What *is* sound and should survive the rewrite: the forward direction. The block is reachable only when
stage 1 returns `Publish`, so any `authored=` / `rtt_owns=` / `SkipNotAuthored` datum necessarily comes
from a moment when nothing else published **and** a present contract existed **and** the run was
unscaled. That is a real constraint and worth keeping. Likewise "`rtt_owns=` cannot be collected on the
title route" follows straight from the measured zero and is fine.

One further step is also unsupported: "every `authored=0` / `px_front=none` fact … belongs to the *other*
regime" (`:1489`) assumes the difference between the older runs and today's is **regime**. It could be
code state, route, or scale configuration — those runs are days and several merges apart. What the zero
establishes is that the older facts came from a run in which the fallback *was* entered and today's title
route does not enter it; *why* is not yet measured.

Remedies, any one: cite the frame instead of the zero; add the stage-1 report the header itself says
would close the gap ("Giving stage 1 its own low-budget report would close that gap"); or mark the cell
"never reached — cause not distinguished", matching the Bendy precedent.

### B5. The new `no-effect(early)` column is unreconciled with the same census already recorded in this file — at a different cumulative total

`STRAY_STATUS.md:1456-1459` adds `1,143` / `1,154` for the two scanout VAs, labelled title route, read at
4,096 discards. The row never interprets the number, and never cites what this document already holds
about it. Three problems, escalating:

**(a) The doc already records this census on the title screen — at 8,192, not 4,096.**
§ *The title screen's REAL numbers* (`:887-899`):

| | calibration (0.1140) | **title screen (0.0069)** |
| --- | --- | --- |
| draws discarded | 1024 | **8192** |
| `shader-recompile` | 7 | **~3800** |
| `no-effect` | 1015 | ~4400 |

and that section states the methodology explicitly at `:892-893`: *"census read at the same cumulative
total in every arm."* Your figures are read one emission earlier. `report_dropped_draw_target`
(`src/gpu/execute/gpu_execute.hpp:839`) emits its ranked report only when the running total hits a power
of two, so 4,096 is a **snapshot at the 4,096th drop, not a run total** — which you anticipated by
saying "final census emission", but it is only final if the run stopped there, and this file records the
same route reaching 8,192.

Note the arithmetic: 1,143 + 1,154 = **2,297**, against a recorded ~4,400 `no-effect` at 8,192. Almost
exactly half. That is what the same measurement sampled one emission earlier looks like. So this may not
be new data at all, and as presented it is not comparable to the number the file already carries.

**(b) There is already a VOID row adjudicating the obvious inference.** `:1378`:

> **VOID — "`no-effect(early)` is discarding content that should render."** … `PROSPER_NO_EARLY_NO_EFFECT=1`
> moves the same draws to the late `no-effect` verdict with `max_nonblack` unchanged at 0.1140, and
> `PROSPER_FORCE_COLORWRITE=1` makes the composite **worse** (0.1140 → 0.0296). Neither lever admits
> content.

A striking uninterpreted number two thousand lines below that, in a `## Ruled out` section, is an
invitation to re-derive a dead hypothesis at full cost — the precise thing these sections exist to
prevent.

**(c) The column is unit-mismatched with the one beside it.** "census lines" counts
`render_draw_pass_rgba` invocations; "draws dropped" counts individual draws. 946 and 1,143 sit in one
row with nothing saying they are different things, and the natural reading — "so the draws at this target
are all being dropped" — is not licensed.

There is also a real confound worth stating rather than leaving implicit, and `gpu_execute.hpp:1694-1697`
already words it better than I can: *"a surface whose draws are all discarded here reads as 'written by
nothing' in that census and in every write-path watch, **which is indistinguishable from a surface the
guest never renders to**."* That bears directly on the row's three-way question — `no-effect(early)` means
the guest's own `CB_TARGET_MASK` disabled every colour component (`:1690-1693`), which is a *mechanism*
for § *This is not a composite defect*, not a competitor to it. Saying so would strengthen the row.

Remedy: either drop the column, or re-read at 8,192 and present it against `:895-899` with the VOID row
at `:1378` cited. The irony is that this round's headline improvement is regime discipline, and this is
the one number it was not applied to.

---

## Your explicit questions

**The deletion is not a trap-41 shape — confirmed.** The single deleted line at
`live_renderer.cpp:9928` is `got && g_rtt.find(read.metadata.address) != g_rtt.end(),`, and the
identical expression is bound to `renderer_owns_flip` three lines above (`:9924-9925`) and passed in its
place. Pure extraction, semantics preserved character-for-character, and the `+1/-1` is visible in the
diff rather than hidden inside a reverted file. Trap 41 is a whole-file `checkout` silently reverting
another lane's work; this is the opposite. `git diff --numstat` shows 0 deletions in the other five files.

**B1 — fixed, and it was the right one to prioritise.** All three sites now state the falsified result and
name the framing not to restate: `guest_write_watch.hpp:225-233` (the header contract, now pointing at
`docs/STRAY_STATUS.md` rather than carrying a title narrative — that is the right shape for a shared
header), `live_renderer.cpp:9978-9982`, `render_runner.h:5439-5444`. `grep -rn '450 GiB'` over
`src/ frontends/ tests/` returns **0**. Verified independently.

**B2 — fixed, and better than what I asked for.** `renderer_owns_flip` is now printed as `rtt_owns=`
alongside the decision name in both branches of the probe, so the input is measured rather than inferred
from the decision. The comment at `:9967-9974` explaining why all three fields print together is the kind
of thing that stops the next reader misreading one in isolation.

**B3 — fixed in both places.** The census comment (`render_runner.h:5444-5447`) and the row (`:1461-1466`)
both now say a line means "bound as slot 0 of a pass carrying at least one draw" and explicitly not "a
draw's output reached memory". The two-way question is replaced by the three-way one with both citations,
and I re-verified them in the file: tonemap `shaderOut 0.001099` storing `0` at `:975`, and
§ *This is not a composite defect* at `:860`.

---

## Also verified on this head (dc8a11c1db)

The census widening, dedup and `va=0x0` guards are unchanged and still correct; the resolver's locking,
containment arithmetic, out-param initialisation on both platform arms, and the round-1 `_WIN32`
preprocessor result all still hold. The declined-MRT1 gap is now acknowledged in the population claim
(`:1477-1478`), and `select_present_source` returning `None` is back with its regime attached
(`:1499-1503`) — and honestly flagged as untested for the flipped buffer in that regime, which is exactly
right.

Mechanical: `git diff --check` rc=0; deletions 0 in five files and 1 explained in the sixth;
`check_numbered_table.py prosper/docs/STRAY_STATUS.md` rc=0. No build, no tests, as instructed.

---

Both findings are doc-only and neither needs a new run: B4 can cite the frame you already have, or mark
the cell "never reached" per the Bendy precedent; B5 can drop the column or re-read it at 8,192 against
the table already in the file. The underlying results — nothing aliased across 82 targets, the scanout
buffers are colour attachments, and the two regimes are genuinely distinct — are sound and worth landing.

mattias800 (commented):
**APPROVED** at head `7e15bf772f`.

Both round-4 findings are fully addressed, the retractions are recorded rather than quietly deleted,
and I have no blocking findings left. Answers to your three questions, then a short merge note.

---

## 1. "A menu frame exists, so something published" — tight, and tighter than you have claimed

You were right to be suspicious of it, and the answer is better than "good enough": **the artifact you
are citing is a dump of the stage-1 input itself.**

`frontends/shared/live/live_renderer.cpp:10160`:

```cpp
const std::vector<uint8_t>& px = selected_pixels ? *selected_pixels : empty_pixels;
```

and `:10184` writes exactly that buffer as `frame_%04d.bmp`. So a non-empty `frame_NNNN.bmp` showing
the menu proves `selected_pixels` was non-empty at that frame — and `have_selected_pixels` is *the*
second stage-1 input, `guest_scanout_present.hpp:81`:

```cpp
:80  if (published_gpu)                            return SkipPublishedGpu;
:81  if (renderer_scanout || have_selected_pixels) return SkipRendererSource;
```

So at any frame that produced a menu BMP, stage 1 declined for a **publishing** reason — `SkipPublishedGpu`
or `SkipRendererSource` — and *not* for `SkipNoPresentContract` or `SkipScaledPresent`. That is precisely
the ambiguity B4 was about, closed by the artifact you already have. The `px.empty() && !published_gpu`
branch at `:10179` also means an empty selection prints "Vulkan render FAILED" instead of dumping, so a
menu-bearing BMP cannot be the empty case.

**So the answer to "is it still too loose" is no** — and there is a free upgrade sitting in it: the cell
can say the frame dump *is* an observation of `have_selected_pixels`, rather than "a frame exists, so
something published", which reads as an inference and invites exactly the challenge you anticipated.

Two non-blocking refinements, both one clause:

- **Name the artifact.** `frame_%04d.bmp` from the live renderer's present path
  (`live_renderer.cpp:10160`, `:10184`) — not "a captured frame". The charter's caption rule already asks
  for the frontend and route, and here naming it is what carries the argument, because the strength of
  the claim is entirely in *which* buffer got dumped. An RTT/pass dump would have proved only that the
  menu was drawn.
- **Fresh vs retained.** `selected_pixels` may be a retained frame rather than a fresh one, and this file
  uses that distinction elsewhere (`fresh=0 retained=0` in the very next table row). It does not weaken
  the cell — `have_selected_pixels` is true either way and stage 1 declines for a publishing reason
  either way — so this is vocabulary consistency, not a correction.

## 2. Nothing else leans on the withdrawn column — verified

`grep` for `1,143` / `1,154` / `2,297` / "thousands of draws" across `docs/STRAY_STATUS.md`, `BLOG.md`
and `src/ frontends/ tests/` returns hits **only inside the withdrawal paragraph itself**
(`:1466-1467`), where they are explicitly the draft's figures and closed with *"Use the existing table;
do not quote the withdrawn one."* That is the right treatment — recording the retraction beats deleting
it, and it matches how this file handles its other reversals. No code comment ever cited those numbers.
BLOG is correctly reduced to "attached to passes that carry draws".

On your question about the VOID row at `:1378`: **not needed now.** It existed to stop someone
re-deriving "no-effect(early) is discarding content" from an uninterpreted number, and there is no longer
a number to misread. The withdrawal paragraph points at § *The title screen's REAL numbers*, which is the
right destination. Adding it would be harmless but is not owed.

## 3. "Never reached" is consistent with the file's own usage — verified

`guest_scanout_present.hpp:27` and `:33` are the defining use: a zero-line control "honestly reported as
**'never reached'** rather than as a decline". The row at `:1505` and `:1509` uses it in exactly that
sense, and cites the Bendy precedent by name. Across the corpus the term is used the same way — e.g.
`LITTLE_NIGHTMARES_3_STATUS.md:174` ("never reached it at all and is void for any question about…"),
`RECOMPILER_REMAINING.md:213`. No drift.

Worth noting for its own sake: `RECOMPILER_REMAINING.md:449` records a near-exact precedent for this
arc — *"Conclusion true, derivation void — and that combination is why this row exists."* The row you
have now written is that shape, handled the same way.

---

## What I checked on this head, and the merge note

Docs-only since `dc8a11c1db`: `git diff --name-only dc8a11c1db..HEAD` is `BLOG.md` and
`prosper/docs/STRAY_STATUS.md`, nothing else — so the four code files carry my round-4 review forward
unchanged, and every code finding from rounds 1-4 stays resolved (the `_WIN32` arm confirmed by
preprocessor, `rtt_owns=` measured rather than inferred, the census scope statement, the out-param
initialised on both arms, the single deletion a pure extraction).

Gates, run read-only here: `git diff --check` rc=0; `check_numbered_table.py prosper/docs/STRAY_STATUS.md`
→ 26 tables, 94 rows, unbroken, rc=0; the repo-wide `git ls-files '*.md' | xargs` form rc=0; deletions
**0 in five files and the one explained extraction in `live_renderer.cpp`**. No build, no tests, as
instructed.

**One thing before you merge, and it is the only reason I am flagging the mechanics at all: this PR is
not docs-only.** The charter's "merge a docs PR without waiting for CI" exception applies to a diff that
touches nothing but `.md` — this branch carries four code files, so the full required-CI wait applies.
At the moment I write this, `Docs` and `Contribution shape` pass and the rest are **pending**, including
**Windows App** — which is the job that would have caught round 1's B1 and the only one that compiles
`prosper_live_renderer` against a real Windows toolchain. Wait for it to go green rather than for the
docs jobs.

Optional, given this appends 109 lines to a status doc other lanes also touch:
`python3 prosper/tools/docs/check_merge_result.py` immediately before merging runs the gate on the
actual merge result, which is the only check that sees a concurrent append.

---

Five rounds, and the row that comes out of it is stronger than the result that started it: two drafts
retracted in place with the evidence that killed them, the instrument's scope stated where the number is
quoted, and the one inference that mattered replaced by a measurement. Good to merge once CI is green.

mattias800 (commented):
**REJECTED** at `f6fb9beadf` — one blocking finding, no re-run needed to fix it.

First: **you were right about my citation, and flagging it rather than quietly adjusting was exactly the
right call.** I traced `frame_%04d.bmp` because `screenshot.cpp:5` says the tool samples "the present
layer (present_readback)", assumed the artifact, and cited a chain you never collected. That is my error
and it would have put a fabricated provenance in a `## Ruled out` row.

The reason I am rejecting rather than approving is that **the replacement has the same shape one level
down** — and the fallback you asked me to check is real, just not in the function the row names.

---

## Blocking. The screenshot chain does not support the claim as written

`STRAY_STATUS.md:1512-1517`:

> These are `tools/screenshot` PNGs, which come from `present_readback`, which returns what
> `present_write_frame` wrote (`gpu_executor.cpp:11780`/`:11927`). So a menu-bearing PNG is direct
> evidence that **a frame was published to the present layer**.

**(a) `tools/screenshot` never calls `present_readback`.** `grep -c 'present_readback('
tools/screenshot/screenshot.cpp` returns **0**. The two occurrences of the name are the stale header
comment at `:5` and an `#include` comment at `:54`. The capture path is **`present_snapshot`**
(`screenshot.cpp:725`). So the row cites a mechanism the tool does not use — which happens to be the
same comment that misled me.

**(b) The conclusion is false on a default run, through `present_snapshot`'s own fallback.** This is the
answer to your question, and it is yes:

`src/gpu/present/videoout_present.cpp:187-196` — when `g_have_frame` is false, `present_snapshot` does
**not** fail. It reads the guest's display buffer and labels it:

```cpp
VideoOutLinearRead read;
if (!videoout_read_front_linear(read)) return false;
out.rgba   = std::move(read.pixels);
out.source = PresentSource::RawScanout;
```

No `present_write_frame` is involved on that path. And `tools/screenshot` **saves** such a capture: the
gate is `source_allowed = snap_from_renderer ? rendered_capture : raw_scanout` (`:732`), where
`raw_scanout` requires `render_first == 0` (`:720-722`) — and `render_first` **defaults to 0**
(`:512-513`, `getenv("PROSPER_RENDER_FIRST") ? … : 0`). So on a default `tools/screenshot` invocation, a
run in which nothing was ever published still writes PNGs, and they show whatever the guest put in its
own display buffer.

Concrete failure scenario: the guest composites its menu into `0x9fc0000000` and flips; prosper publishes
nothing; `present_has_frame()` is false; `present_snapshot` returns `RawScanout` with the de-swizzled
guest bytes; the tool saves a menu-bearing PNG. The row would then read that PNG as proof of a
publication that never happened — the exact substitution this paragraph exists to prevent.

For completeness: `present_readback` has the *same* hazard at `videoout_present.cpp:132-138` (guest
display buffer, "contents are whatever the guest put there"), so the conclusion would not have followed
even if the tool did use it. Both of your suspicions were well aimed; `videoout_present.hpp:115`'s
size-guard is the benign one (it drops frames, it does not synthesise them).

### The fix needs no new run — the discriminator is already in your artifacts

`tools/screenshot` records the source per capture: `capture_source` is `Rendered` / `GuestScanout` /
`RawScanout` (`screenshot.cpp:734-739`) and goes straight into the manifest at `:757-758`
(`observation.source = capture_source`). So read it off the PNGs you already have:

- **`Rendered` or `GuestScanout`** → the frame came through `present_write_frame`, and your sentence is
  true exactly as written. Cite the recorded source rather than the run configuration; it is stronger and
  needs nothing new.
- **`RawScanout`** → no publication occurred, and the claim has to go.

`--require-composited-frame` (`:881`) asserts `Rendered` specifically, which is *more* than this row
needs — a `GuestScanout` republish is still a publication to the present layer.

---

## Answers to your other points

**Fresh-vs-retained: you were right to drop it, and it does not need re-adding.** On the
`Rendered`/`GuestScanout` path `present_snapshot` returns whatever `g_frame` currently holds, which may
be an earlier publication — but your sentence is existential ("a frame **was** published"), so a retained
frame satisfies it. The wording only mattered under the stronger per-flip reading I had in mind for the
BMP. Leave it out.

**Refinement 2 is correct — I verified the new claim rather than taking it.** `PROSPER_RTTLOG` is indeed
in the `live_gpu_targets` disabling chain (`live_renderer.cpp:1679`, the last conjunct), and
`PROSPER_PASS_LOG` appears nowhere in that chain (`:1670-1679`). `g_pass_log_window` is parsed from
`PROSPER_ENV_VALUE("PROSPER_PASS_LOG")` at `:1225-1226`, and the `px_front`/`px_vo`/`px_last` reporting
is built at `:7827-7862`. The warning that the obvious route measures a differently-configured renderer
is accurate and is a genuinely useful thing to have written down.

**The `frame_%04d.bmp` paragraph you added is accurate** — `selected_pixels` at `:10160` → `:10184`,
proving `have_selected_pixels`, stage 1's second input — and warning against substituting an artifact
you did not collect is the right lesson to leave behind. Keep it; only the sentence above it is wrong.

**Non-blocking:** `screenshot.cpp:5`'s header comment ("present layer (present_readback)") is stale and
is the proximate cause of both my error and the row's. Worth correcting in a later PR — not this one,
which should stay docs-only.

---

## Otherwise unchanged and still approved

`git diff --name-only 7e15bf772f..HEAD` is `prosper/docs/STRAY_STATUS.md` alone, so the four code files
and BLOG carry my round-5 review forward untouched, and every finding from rounds 1-5 stays resolved.
Gates re-run read-only on this head: `git diff --check` rc=0; `check_numbered_table.py` on the file rc=0
and repo-wide `.md` rc=0; deletions 0 in five files plus the one explained extraction in
`live_renderer.cpp`. No build, no tests.

Merge note from round 5 still stands and you have it: not docs-only overall, so wait for the full
required set — `Windows App` in particular — and run `check_merge_result.py` immediately before merging.

---

One sentence to change, and the evidence to change it with is already on disk. If the manifest says
`Rendered` or `GuestScanout`, the row is right and only the mechanism name needs fixing; if it says
`RawScanout`, the publishing claim needs withdrawing the way the last two did. Either way this is the
last thing I have.

mattias800 (commented):
**APPROVED** at `a43ad37963`.

The claim now rests on a machine-recorded field rather than on an image, the hazard is stated *before*
the claim it qualifies, and the whole thing is falsifiable by anyone holding the manifests. That is the
right shape, and it is stronger than what the row needed.

## What I checked

**The enum mapping — verified independently, not taken from your message.**
`tools/screenshot/capture_manifest.cpp:246-253`:

```cpp
const char* capture_source_name(CaptureSource source) {
    switch (source) {
        case CaptureSource::Rendered:     return "composited";
        case CaptureSource::GuestScanout: return "guest_scanout";
        case CaptureSource::RawScanout:   break;
    }
    return "raw_scanout";
}
```

Exactly as you stated. So `composited` in a manifest is `CaptureSource::Rendered`, which comes from
`PresentSource::Rendered`, which `present_snapshot` sets only when `g_have_frame` is true with a
non-`GuestScanout` origin — and `g_have_frame` is set only by `present_write_frame`, whose two call
sites I verified last round are both guarded on a real publication at the correct size. So
`composited` ⇒ prosper published a composited frame. The chain closes.

I cannot verify 29-of-30 from here — those manifests are on your disk — but that is no longer a problem
with the row: it names the field, the tool, and the specific frame, so the next reader can check it in
one command instead of trusting a sentence. Tying the *steady-state menu frame examined by eye* to
`composited`, rather than resting on the aggregate, is the part that makes it tight.

**The hazard paragraph is accurate.** `present_snapshot`'s `RawScanout` fallback at
`videoout_present.cpp:187-196`, saved by default because `render_first` defaults to 0, and the
observation that the guest composites its own menu into `0x9fc0000000` so the image would look
identical — all correct, and putting it ahead of the claim is what stops the next reader repeating the
inference. The `present_readback` warning and the retained `frame_%04d.bmp` paragraph both check out.

**Gates on this head:** docs-only since `f6fb9beadf` (`STRAY_STATUS.md` alone), so the four code files
and BLOG carry the round-5 approval forward untouched. `git diff --check` rc=0;
`check_numbered_table.py` rc=0 on the file and repo-wide across all `.md`; deletions **0 in five files**
plus the one explained extraction in `live_renderer.cpp`. No build, no tests.

**Merge:** your plan is right. Note the docs push started a **new** CI run (`33916001271`) — on
`a43ad37963` only `Docs` and `Contribution shape` have passed and `Windows App` is **pending**, so the
5-passed count was the previous head. Wait for the full set on this SHA, then `check_merge_result.py`
immediately before merging.

---

## On `screenshot.cpp:5` — fix it, and I would not fix it with one line

No reason to leave it. Two things make it worth more than a typical comment tidy:

- It is a wrong signpost in the one tool **every progression claim in this project routes through**.
- Its cost is measured, not hypothetical: two wrong citations in one review cycle, by two readers
  independently, both of whom went to the named function first. That is exactly the charter's #2049
  shape — a wrong statement in a shipped comment outran its correction because the comment is what
  people read first.

But a rename to `present_snapshot` would fix the citation and leave the trap. The trap is not the name;
it is that the two functions differ in a way that is load-bearing for evidence:

| | guest-buffer fallback | says which source it used |
| --- | --- | --- |
| `present_readback` (`videoout_present.cpp:120-138`) | yes | **no** |
| `present_snapshot` (`:154-196`) | yes | **yes** (`PresentSource`) |

Both fall back to guest memory. Only one lets a caller tell afterwards. So I would land the name **plus
the one sentence that makes it matter** — captures go through `present_snapshot`, which falls back to
the guest's own display buffer and *labels* it, so a PNG's provenance lives in the manifest's `source`
field and never in the image — and point at `capture_source_name` (`capture_manifest.cpp:246`). The
`#include` comment at `:54` names `present_readback` too and should go with it.

Separate follow-up after this merges, as you proposed — not folded in here, which would break the
docs-only-since-approval property and put an unreviewed code file in at the last moment.

One thing that is your call rather than mine: the *pattern* may be worth more than the comment fix.
Three artifact substitutions in three rounds — my BMP for your PNG, your `present_readback` for
`present_snapshot`, the image for its provenance — all survived because the picture was never in doubt,
only its provenance was. That is a reusable instrument trap, of the kind
`GAME_COMPAT_ORCHESTRATION.md`'s numbered list exists to hold, and it cost this review three rounds to
surface. `tools/docs/trap_number.py` allocates against master and open PRs if you want it.

---

Six rounds. Both drafts retracted in place with the evidence that killed them, the instrument's scope
stated where each number is quoted, and the last inference replaced by a recorded field. Good to merge
once CI is green on this SHA.

PR #3334, opened 2026-09-04
screenshot: the picture is not the provenance — fix the stale comment that misled two readers, and record trap 258

mattias800:
Refs #2932. Follow-up to #3332, at its reviewer's suggestion.

## The defect

`tools/screenshot/screenshot.cpp:5` said the tool *"samples the present layer (`present_readback`)"*.
It does not call `present_readback` at all — it captures through `present_snapshot`.

That one stale word sent **two readers independently to the wrong function inside a single review
cycle** on #3332 (a reviewer, then me), in the one tool every progression claim in this project routes
through. It is the #2049 propagation shape: a wrong signpost that reads as verified because it sits in
the source.

## Why a rename would not have been enough

The trap is not the name. **Both** functions fall back to guest memory when the renderer published
nothing, and only `present_snapshot` reports which source it used. Fixing only the citation would have
left the thing that actually misleads.

So the comment now carries the sentence that makes the name matter:

> a correct-looking PNG does not mean prosper composited anything — a title that draws its own menu
> into its own scanout buffer produces a byte-identical picture whether we rendered it or not, and
> which one it was lives in the manifest's `source` field (`capture_source_name()` → `composited` /
> `guest_scanout` / `raw_scanout`), nowhere in the pixels.

*Stray* is exactly that title and `0x9fc0000000` is exactly that buffer, which is how the error got as
far as a `## Ruled out` row before anyone caught it.

## Instrument trap 258

The pattern outlived each of its instances, so it is recorded in its general form: **an artifact's
provenance read off its picture** — three substitutions on one claim, each caught only by the next
reader, all surviving because nobody doubted the image. Only its provenance was ever in question, and
nobody asked.

The entry also records that **no re-run was needed**: the discriminator was in files already on disk
(29 of 30 captures `composited`, the eyeballed steady-state frame among them). That is the part worth
inheriting — the instinct is to re-measure, and the answer was already written down.

## Scope

- `screenshot.cpp` — **comment-only in effect.** The sole non-comment line in the diff is an
  `#include` whose *trailing comment* changed; the include path is byte-identical. Verified with
  `git diff | grep -vE '^[+-]\s*//'`.
- `GAME_COMPAT_ORCHESTRATION.md` — one appended row.

## Verification

```
  -> 14 tables, 320 rows; numbered column 1..258, unique and ascending    # rc=0
git diff --check                                                          # rc=0
```

Trap number allocated with `tools/docs/trap_number.py` (scans master **and** open PRs): **258**.
`check_merge_result.py` will be run immediately before merging — it is the only check that sees a
concurrent append, and it goes stale as soon as another lane lands.

No build or tests run locally: the C++ delta is comment text. CI compiles it regardless, and I will
wait for the full required set rather than treat this as docs-only, since the diff is not `.md`-only.


mattias800 (commented):
**REJECTED** — two blocking findings, both one-line fixes, both in trap 258 rather than the code. **The
C++ comment is accurate and I have no findings against it.**

Taking your questions in your order.

---

## 1. Does trap 258 state the general rule? — Yes, it does

This was your main worry and I do not share it. Column 1 is a general shape, not a narration:

> **An artifact's PROVENANCE read off its PICTURE — when the capture tool falls back to guest memory, a
> correct-looking image is the WEAKEST evidence available.**

and the two closing clauses are both reusable and in the house format that 256 and 257 established:

> *Recognition cue:* a claim of the form "the screenshot shows X, therefore the emulator did X", where
> the capture path has any fallback that can source pixels from the guest.
> *Avoid it:* cite the recorded provenance, never the picture; when a tool records its own source, quote
> that field, and prefer it to run configuration since a default can be overridden while the record
> cannot.

The three-incident narrative in column 2 is what 256 and 257 also do, and it earns its place because the
*sequence* is the lesson — each substitution was caught only by the next reader. "All three survived
because nobody doubted the image" is the sentence a future reader will actually recognise themselves in.
No war-story problem here.

**One non-blocking gap in the cue:** both the cue and the fix assume the tool *does* record a source. The
harder case is the tool that does not — and then the honest answer is that the artifact cannot support
the claim at all, which is worth a clause because it is the case where people improvise. Suggest
appending to *Avoid it:* "— and if the capture path records no source, the artifact cannot establish
provenance at all; collect one that can rather than reasoning about the picture."

## 2. Is the comment accurate about `present_snapshot`'s fallback? — Yes, all four claims verified

You asked because you are now the second person to write confident prose about this area. I checked every
factual claim in the new text against source rather than against your description:

| claim | verified |
| --- | --- |
| `present_snapshot` falls back to the guest's display buffer, returning `RawScanout` | `videoout_present.cpp:187-196` ✓ |
| the tool saves that by default | `raw_scanout` needs `render_first == 0` (`screenshot.cpp:729`); `render_first` defaults to 0 (`:521-522`) ✓ |
| provenance is `capture_source_name()` → `composited` / `guest_scanout` / `raw_scanout` | `capture_manifest.cpp:246-253` ✓ |
| `--require-composited-frame` asserts **Rendered specifically** | `rendered_samples_++` fires only on `CaptureSource::Rendered` (`capture_manifest.cpp:208`), `guest_scanout_samples_` is a separate counter (`:209`), and the assertion gates on `rendered_samples() == 0` (`screenshot.cpp:890`) — so a run producing only `guest_scanout` captures fails it ✓ |

Also confirmed: `grep -c 'present_readback('` in the tool is **0**, so the corrected name is right and no
call site contradicts it; the `#include` comment's other two names (`present_count`,
`present_width/height`) are still used. The diff is comment-text plus one byte-identical include path.

Not naming a line number for `videoout_present.cpp` is the right call — line numbers rot and the function
name is findable.

## 3. Should it cross-reference 209 and `prerender_check.py`? — `prerender_check.py` yes, **209 no**

`prerender_check.py` is exactly right and tight. **Trap 209 is the wrong number**, and this is one of the
two blocking findings below.

---

## Blocking

### A. Trap 258 cites trap 209, which is a different family — the sibling it wants is **230**

Trap 209 (`:337`) is: *"A rate-limited log sampling a low-duty-cycle signal reports its ABSENCE with
perfect consistency"* — `PROSPER_PADLOG` at every 512th call against a ~2% duty-cycle route. It is a
**sampling-rate** trap. Nothing in it concerns images, or evidence that is convincing without being
discriminating.

The trap that *is* 258's family is **230** (`:358`): *"A screenshot of a game's own PRE-RENDERED LOADING
PICTURE, mistaken for the rendered 3D world… the most convincing false-progression evidence this project
has produced, and it survived a PR, a `BLOG.md` entry and a merge."* That is also the trap
`prerender_check.py` belongs to — so the row currently names the tool and the wrong number for the same
idea. If you want the metric-fooling half as well, **111** (a wrong-but-plausible palette passing visual
review) and **181** (a bloom smear ranking the worse frame higher) are the precise ones.

Why blocking rather than cosmetic: `check_trap_citations.py` **cannot catch this**. Its own docstring says
it validates that a cited number *names a row that exists*, and describes this exact failure as the one
nobody finds — *"a plausible number in a plausible sentence, and the only way to find out is to open the
table and count. Nobody does that."* I ran it and the numbering gate: 258 rows, unique and ascending, no
deletions against `origin/main`, `diff --check` rc=0. All green, and none of them sees an apt-but-wrong
cross-reference. A reader who follows 258 → 209 lands on a pad-logging duty cycle and concludes the
family reference was noise.

There is an irony worth keeping: **209 is the right trap for the grep incident you disclosed** — a
periodic/absent-capable instrument reporting silence that was read as a measurement. It is just not the
right one for 258.

### B. "Stray composites its menu into `0x9fc0000000` itself" is asserted as fact and was never established

The sentence:

> On this title the two cases are visually identical — Stray composites its menu into `0x9fc0000000`
> itself, the very buffer under investigation — so the failing run and a working one produce the same
> picture.

Nothing in the #3332 arc established that the **guest** writes that buffer, and three things cut against
it:

- This PR's own headline finding is that **29 of 30 captures were `composited`** — i.e. *prosper*
  rendered the menu, not the guest echoing back through `RawScanout`.
- `STRAY_STATUS.md` reports `authored=0` for the flipped buffer, and the row's own best insight is that
  *"if prosper renders into a `VkImage` keyed by the scanout VA, guest memory at that VA need never be
  written"* — i.e. the expectation is that the guest's bytes there are **not** the menu.
- If guest memory at that VA is unwritten, a `RawScanout` capture would show black, not the menu, and the
  "visually identical" claim fails for this title specifically.

The general hazard is real and I am not asking you to weaken it — I am asking you to ground it in the
example that is *measured*. `guest_scanout_present.hpp:9-17` records it: on *Sonic Frontiers* the guest's
own flipped buffer held, de-swizzled, **"an exact SEGA logo and an exact intro shot"**, with prosper's
render-target map empty at the flipped address. That is a title where a `RawScanout` capture genuinely
would look correct, it is the case the whole #1968 fallback exists for, and it does the same rhetorical
work without asserting anything unverified about Stray.

Suggested substitution: *"A `RawScanout` capture shows whatever the guest left in that buffer, and on
some titles that is a correct frame — #1968 exists because Sonic Frontiers' flipped buffer held an exact
SEGA logo and intro shot with prosper's target map empty. Whether Stray's buffer would have looked
correct was never established, which is the point: the picture cannot tell you."*

That version is stronger, because it makes the trap about the *unknowability* rather than about a fact
that would itself need checking. As written, a trap row about not trusting unverified claims contains
one — in the project's highest-authority table, where citation errors propagate hardest (#2049).

---

## On your disclosure

Noted, and it is the right thing to have told me. For what it is worth it did not change how I reviewed
this — I have traced every citation in this arc against source since round 1, including my own after you
caught the BMP substitution — but it does confirm the policy is the correct one for both of us.

That incident is trap 209's family proper (a periodic or silent-capable instrument, an absence, and a
conclusion drawn from the silence), and it is close enough to 258 to be worth a sentence somewhere. Your
call whether it belongs as its own row or a clause on 209; it is out of scope for this PR either way.

---

## Otherwise

CI 9/9 green on `72cd4e854d` — I confirmed all ten non-skipped jobs pass, `Windows App` and
`Windows MinGW` among them. Numbering gate with `--ordered --table-header Instrument --baseline
origin/main`: 320 rows across 14 tables, numbered 1..258 unique and ascending, **no baseline row
deleted**. `diff --check` rc=0. Deletions: 0 in the doc, 2 in `screenshot.cpp` — both the replaced
comment lines, with the sole non-comment line a byte-identical include path. Confirmed no build or tests
needed.

Both fixes are single lines. Once they are in, run `check_merge_result.py` immediately before merging as
you planned — this appends to the numbered table and it is the only check that sees a concurrent append.

mattias800 (commented):
**APPROVED** at `5d3d94f710`. Both blocking findings verified fixed against source.

**A — sibling citation.** The row now reads *"Same family as trap 230 and the pre-rendered-frame check
(`tools/evidence/prerender_check.py`)"*. Correct: 230 is the pre-rendered-loading-picture trap and the
one that tool belongs to. I also confirmed your cleanup check independently — `grep -n 'trap 209'` over
`GAME_COMPAT_ORCHESTRATION.md` returns exactly one hit, row **257**'s own, so nothing else in the file
now points at 209 for an image-family idea.

**B — the unverified claim.** Replaced with the measured case, and I checked the substitution rather than
the sentence: `guest_scanout_present.hpp:9-17` does say *Sonic Frontiers* "runs ~3,700 passes across a
boot without a single one whose target is a registered VideoOut buffer", with the flipped buffer holding
"de-swizzled, an exact SEGA logo and an exact intro shot". The row's paraphrase is faithful, and it now
demonstrates the hazard from something someone measured instead of asserting a fact about Stray that the
arc never established. This was the one that mattered and the fix is the right one.

**Non-blocking clause taken**, and taken well — *"when the tool records NO source, the honest answer is
that the artifact cannot establish provenance at all -- not that it probably did"* is the case people
improvise around, and naming the remedy (reach for a capture path that records one) keeps it actionable.

Gates re-run read-only on this head: `--ordered --table-header Instrument --baseline origin/main` →
320 rows across 14 tables, numbered 1..258 unique and ascending, **no baseline row deleted**;
`diff --check` rc=0. The C++ half is unchanged from the version I verified last round.

Good to merge; `check_merge_result.py` immediately before, as you planned.

PR #3335, opened 2026-09-04
diag: let the compute census speak for small populations; the skipped-dispatch hypothesis is falsified on Stray's title screen (#2932)

mattias800:
Refs #2932, #3126.

## What this closes

`STRAY_STATUS.md` carried a **VOID** row: *"a skipped compute dispatch collapses the composite"* — the
charter's LUT/exposure shape — measured on the **calibration** screen and never on the title screen,
so it was not a falsification for the screen anyone cares about. This runs the missing arm.

**Falsified on the title screen.** Route landed, failing steady state confirmed by capture provenance
(29/30 `composited`) rather than by eye alone:

```
65,536 dispatch decisions over 21 programs
program=0x300ba70000  executed=15103  skipped=1  groups=640x1x1 local=64x1 threads=40960x1
```

One skip in 15,104, and only one program ever skips. The dispatch shape settles it independently:
`640x1x1` groups of `64x1` is a **1-D** launch of 40,960 threads, not the ~`480x270` a full-screen 4K
pass would take — so this program is not a scene writer whatever its skip rate.

**The cross-check is the strongest part, and it was free.** Read at the cumulative total the
calibration arm used (the methodology § *The title screen's REAL numbers* requires):

| | calibration (renders **correctly**) | title (**black**) |
| --- | --- | --- |
| only skipping program | `executed=7455 skipped=2` | `executed=7455 skipped=1` |

Identical to the dispatch. **Compute behaviour does not distinguish the working screen from the broken
one**, so no fix aimed at compute skipping can explain the difference. Two independent runs agreed.

## The instrument fix that made it readable

The census emitted only at power-of-two totals **≥ 512**, so any run below that printed nothing —
a null indistinguishable from "the census is off", "the hook is not on this path", and "this build
lacks the diagnostic". A small population is a legitimate answer to *how much compute does this screen
run*, and the floor made it the one answer the instrument could not give.

**Honest provenance, in the code comment as well as here:** the first version of this change claimed
a Stray run had hit the floor. It had not — that run recorded 65,536 decisions and printed eight
census blocks. I had grepped for a string that does not occur (`program-census`) and then for one that
matched the run's own **directory name**, and read the silence as a measurement. The floor did not
cause that and removing it would not have prevented it. It is removed on its own merits, and the
comment says so rather than shipping a tidier story.

## A stale pointer, corrected

§ *The MUBUF class, decoded* ended by proposing an `rt == nullptr` probe that *"answers it in one
run"*. § *Ruled out* had already run it and killed it — `rt=1` with five resources, and the
`VCC_LO`/`VCC_HI` soffset excluded too. The section was sending the next reader to re-derive a dead
answer at full cost, a hundred lines above the row that falsified it, which is precisely what the
`## Ruled out` convention exists to prevent.

It now states what actually fails — **SRSRC provenance**: both stages build their own descriptor at
runtime (`s_load_dwordx4` with a register offset, then a `s_cselect_b32` patch of dword 3), leaving no
SRT key — and why that is the frontier:

> `shader-recompile` discards **~3800 of 8192** on the title screen against **7 of 1024** on
> calibration. Not a difference of degree between the screens; the difference between them.

## Risk

The census is diagnostic-only and off by default. Removing the floor makes an already-opt-in
diagnostic print at 1, 2, 4 … 256 as well — more lines only for runs that previously printed none.
No production path changes.

## Verification (exact head)

```
cmake --build prosper/build-linux -j6      # rc=0, 0 errors
ctest --no-tests=error -j4                 # rc=0
```

**`100% tests passed out of 362`**, exit 0. Two skips: `module_loads_eboot` (expected — `GAME_DUMP` is
`PPSA02101`, not `PPSA24651`, per #1573) and `plugin_autolink`.

```
check_numbered_table.py prosper/docs/STRAY_STATUS.md -> 26 tables, 94 rows, rc=0
git diff --check                                     -> rc=0
```

### DIFF
--- a/prosper/docs/STRAY_STATUS.md
+++ b/prosper/docs/STRAY_STATUS.md
@@ -919,11 +919,25 @@ sh=/221 pc=66 words=e0042000,6b020900 op=0x1 src=0(k2),8(k1),107(k6)  stage=vert
 
 Both are `buffer_load_format_*` with `idxen` — **vertex attribute fetch** — in small vertex shaders,
 with `VCC_LO`/`VCC_HI` as the soffset. The opcodes themselves are implemented (`0x0`-`0x3` are all
-handled), so the reject is an operand, not a missing instruction. The next thing to establish is
-whether the failing operand is the soffset or the SRSRC: `allow_smem` is `(rt != nullptr)` for
-graphics, so **a vertex stage that arrives with no resource table fails every buffer op it has**, and
-vertex fetch is a buffer op — that would take the draw with it. A probe on `rt == nullptr` at the
-vertex recompile answers it in one run.
+handled), so the reject is an operand, not a missing instruction.
+
+**This paragraph used to end by proposing an `rt == nullptr` probe "answers it in one run". Do not
+run it — § *Ruled out* already did, and killed it:** the disposition line reports `rt=1` with five
+resources, and the `VCC_LO`/`VCC_HI` soffset is not it either (`operand_bits` has a vertex-stage
+branch for exactly that, and the reject happens before the address is formed). The pointer outlived
+the answer by sitting a hundred lines above the row that falsified it, which is the failure mode the
+`## Ruled out` convention exists to prevent — so it is corrected here rather than only there.
+
+**What actually fails is SRSRC provenance, and that is the frontier for this screen.** Both failing
+stages build their own descriptor at runtime — `s_load_dwordx4 s[4:7], s[12:13], vcc_hi` with a
+*register* offset, then a `s_and`/`s_or`/`s_cselect_b32 s7, …` patch of dword 3 — so there is no SRT
+key and the SRSRC range is rewritten by the shader itself. See § *Ruled out* for the full
+disassembly and why the direct-SGPR fallbacks are correctly suppressed.
+
+**Why this is the thing to work on, in one number:** `shader-recompile` discards **~3800 of 8192** on
+the title screen against **7 of 1024** on calibration. That is not a difference of degree between the
+two screens; it is the difference between them, and the census table above is measured at the same
+cumulative total in both arms.
 
 ### What #3133 is worth here
 
@@ -1388,11 +1402,27 @@ evidence nor falsifications for it. Re-run each with `PROSPER_NULL_PAGE=1` and a
 - **"The world renders into an HDR target that the composite then loses."** Falsified: every seeded
   scene target in the title-screen frame is black at source (brightest channel 0-6 across five HDR
   targets). #3126.
-- **VOID — "A skipped compute dispatch collapses the composite" (the charter's LUT/exposure shape).**
-  `PROSPER_COMPUTE_PROGRAM_CENSUS` reports the only program with any skips at **executed=7455,
-  skipped=2** — but that census was read on **calibration** and has not been re-run on the title
-  screen, so it is not a falsification for it. This row said `Falsified` and was corrected: the
-  measurement is real, the scope was overclaimed. #3126.
+- **"A skipped compute dispatch collapses the composite" (the charter's LUT/exposure shape).**
+  **Falsified on the title screen, 2026-09-04** — this row was VOID for the right reason (the census
+  had only been read on **calibration**), and the missing arm has now been run.
+
+  Title route, steady state confirmed by capture provenance and by eye (29/30 `composited`, menu over
+  black): **65,536 dispatch decisions across 21 programs**, and exactly **one** program ever skips —
+
+  ```
+  program=0x300ba70000 executed=15103 skipped=1 groups=640x1x1 local=64x1 threads=40960x1
+  ```
+
+  One skip in 15,104. Nothing is being systematically dropped. The dispatch shape settles it twice
+  over: `640x1x1` groups of `64x1` is a **1-D** launch of 40,960 threads, not the ~`480x270` a
+  full-screen 4K pass would take, so this program is not a scene writer whatever its skip rate.
+
+  **The cross-check is the strongest part, and it was free.** Read at the same cumulative total the
+  calibration arm used (32,768 decisions — the methodology § *The title screen's REAL numbers*
+  requires), the two screens report **`executed=7455 skipped=2`** (calibration) against
+  **`executed=7455 skipped=1`** (title). Identical to the dispatch. So compute behaviour does not
+  distinguish the screen that renders correctly from the one that is black, and no fix aimed at
+  compute skipping can explain the difference. Two independent runs agreed. #3126, #2932.
 - **MIMG `SRSRC` is not a user-SGPR index.** It names any SGPR, so a scan that treats every SRSRC as
   a user-data slot reports mostly false positives — scratch registers the shader loaded a descriptor
   into. An earlier list of "bases missing from the table" derived that way is discarded. #3126.
--- a/prosper/docs/STRAY_STATUS.md
+++ b/prosper/docs/STRAY_STATUS.md
@@ -1160,10 +1160,13 @@ the same guest state across runs.
     `0x0004dfac00000001` — `addr … unreadable` — after which every scalar is `ok=0` and the fetch is
     "left unresolved (not folded to 0)".
   - **This is #305, on a third title — and note WHICH sites, because this file also withdraws a #305
-    attribution a few rows below.** That withdrawal covers the **five unresolved IMAGE ops** (read at
-    user-data dword 0 of a declared `[0,20)` window, `[udcand]` reporting `FITS-BLOCK`, so the
-    larger-block-than-window condition is absent there). This row covers the **VERTEX stages**, whose
-    window is `USER_SGPR = 8`. Both cite the same `0x0004dfac` constant family, so the two are easy to
+    attribution ABOVE this row**, in this same section (§ *The title screen's black composite*) and
+    again far earlier under § *Withdrawn 1*. Nothing below this row mentions #305 at all. That
+    withdrawal covers the **five unresolved IMAGE ops** (read at user-data dword 0 of declared
+    `[0,20)`/`[0,30)` windows, `[udcand]` reporting `FITS-BLOCK`, so the larger-block-than-window
+    condition is absent there). This row covers the **VERTEX stages**, whose window is
+    `USER_SGPR = 8` — and `RSRC2_GS` is the right register for a vertex stage on RDNA2, where vertex
+    is merged into the GS hardware stage, so "GS" beside "VERTEX" is expected rather than a slip. Both cite the same `0x0004dfac` constant family, so the two are easy to
     confuse and the verdicts do not transfer in either direction: do not use the image withdrawal to
     dismiss this row, and do not use this row to revive the image attribution.
     `PROSPER_UDPROV=1` shows the draw's own bind writing twelve
--- a/prosper/docs/STRAY_STATUS.md
+++ b/prosper/docs/STRAY_STATUS.md
@@ -934,10 +934,16 @@ stages build their own descriptor at runtime — `s_load_dwordx4 s[4:7], s[12:13
 key and the SRSRC range is rewritten by the shader itself. See § *Ruled out* for the full
 disassembly and why the direct-SGPR fallbacks are correctly suppressed.
 
-**Why this is the thing to work on, in one number:** `shader-recompile` discards **~3800 of 8192** on
-the title screen against **7 of 1024** on calibration. That is not a difference of degree between the
-two screens; it is the difference between them, and the census table above is measured at the same
-cumulative total in both arms.
+**Why this is the thing to work on, with its denominators stated:** `shader-recompile` discards
+**~3800 of 8192** discarded draws on the title screen against **7 of 1024** on calibration — roughly
+**46%** against **0.7%**. That is not a difference of degree between the two screens; it is the
+difference between them.
+
+Note the two arms are read at **different** cumulative totals, 8192 and 1024, as the table's own first
+row shows. An earlier draft claimed they were "measured at the same cumulative total in both arms":
+this section's same-total rule governs the title-screen `PROSPER_NULL_PAGE=1` arms among themselves,
+and calibration is a separate earlier reference. The contrast survives being stated honestly — it did
+not need the false justification.
 
 ### What #3133 is worth here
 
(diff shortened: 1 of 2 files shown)


mattias800 (commented):
**REJECTED** at `4c51482672` — two blocking findings, both in the doc. **The code change is right and I
have no findings against it.**

You asked me to attack the cross-check hardest. I would go further than "it needs one run": **the
identical `executed=7455` is evidence against the conclusion, not for it**, and the row currently reads
it as its strongest support.

---

## Blocking

### A. The cross-check reads two runs at a point where they have not yet diverged

`STRAY_STATUS.md:1420-1425`:

> Read at the same cumulative total the calibration arm used (32,768 decisions …), the two screens report
> **`executed=7455 skipped=2`** (calibration) against **`executed=7455 skipped=1`** (title). … So compute
> behaviour does not distinguish the screen that renders correctly from the one that is black … **Two
> independent runs agreed.**

Two different screens, two different runs, one program's executed count matching **to the digit**. That
is not two runs agreeing. Different workloads do not produce byte-identical per-program counters by
coincidence — an exact match means the interval being measured is **common to both runs**, i.e. the
shared boot/load phase, before either screen's own behaviour begins. The conclusion it supports is
therefore close to tautological: *up to the point where these runs have not diverged, they have not
diverged.*

The run length makes it concrete. Your title run recorded **65,536** decisions; 32,768 is its
**midpoint**. So the cross-check reads the title arm halfway, at the one ordinal where it happens to
match the other arm exactly.

This is trap **257**, which is four rows above the one you are editing: *"A running-total census read
BEFORE the run reached the state it was meant to describe"*, whose *Avoid it:* is **"take the census from
the block printed at exit"**. Its worked example is the same shape — ratios from one Stray run that
reversed between 62k and 332k attempts, where the early numbers "correctly described the boot, which is
simply not the state anyone wanted measured".

Two supporting premises also do not hold as stated:

- **The calibration arm's cumulative total is not documented.** `STRAY_STATUS.md:883` records that
  figure as `executed 7455, skipped 2` with **no total attached**. Nothing in the file says 32,768 for
  it. So "the same cumulative total the calibration arm used (32,768 decisions)" is an assertion about
  someone else's measurement that the record does not carry — and if calibration's 32,768 was its *exit*
  block, you are comparing a full run against half a run, which is trap **255**.
- **§ *The title screen's REAL numbers* does not require 32,768.** That section's methodology governs the
  **draws-discarded** census (1024 / 8192), a different instrument. Citing it as the source of a
  compute-decision total is a cross-census substitution.

**What to do:** compare each arm at its own **exit** block, with both totals stated — and if you collect
one run per screen anyway, that is the run you offered. Keep the identical 7455, but report it as what it
is: a signature that the two runs are still in common code at that ordinal, which is a genuinely useful
observation and a better sentence than the one it replaces.

Note this does **not** touch the falsification itself. **65,536 decisions, 21 programs, one skip in
15,104** is a title-screen measurement read at that run's own exit, and it is sufficient on its own. The
cross-check is a bonus that currently subtracts.

### B. "Measured at the same cumulative total in both arms" contradicts the table it cites

`STRAY_STATUS.md:934-936`:

> `shader-recompile` discards **~3800 of 8192** on the title screen against **7 of 1024** on calibration.
> … **the census table above is measured at the same cumulative total in both arms.**

The table's own first row is `| draws discarded | 1024 | **8192** |` — the two arms were read at
**different** totals, 1024 and 8192. The section's sentence *"census read at the same cumulative total in
every arm"* governs the title-screen `PROSPER_NULL_PAGE=1` arms among themselves; the calibration column
is a separate earlier reference. So the new clause asserts the opposite of the table it appeals to.

The contrast itself is probably real — 0.68% against 46.4% is ~68x, and a phase artifact that large
would be surprising — but the justification is false, and per trap 257 a running-total ratio is not
stationary, so two ratios read 8x apart in cumulative total are measurements of different intervals.

**What to do:** drop the clause and state the ratios with their denominators — *"7 of 1024 on
calibration against ~3800 of 8192 on the title screen; the arms were not read at a common total, but a
68x gap is not a plausible phase artifact."* That is weaker in form and much stronger in fact, and it is
the honest version of "I am now leaning on them hard".

---

## Non-blocking

**The printed dispatch shape is the LAST one observed, not the skipped dispatch's.**
`gpu_executor.cpp:2339` is `if (gx) { entry.gx = gx; … }` — every call overwrites, so the `640x1x1` /
`64x1` printed for a program with 15,104 decisions is whatever its most recent dispatch used. The census
keeps one shape, not a set, so it cannot show the shape was stable. The row says this "settles it twice
over … not a scene writer whatever its skip rate"; that holds only under stability the instrument does
not measure. Likely true for a program dispatched 15k times, and the 1-in-15,104 argument carries the
row by itself — so one clause is enough: "the census records the last observed shape, so this is the
program's shape, not necessarily the skipped dispatch's."

---

## Verified good

**The code change and its comment.** The census is opt-in — `static const bool on =
std::getenv("PROSPER_COMPUTE_PROGRAM_CENSUS")`, `gpu_executor.cpp:2326-2327` — so removing the floor adds
emissions (n=1,2,4…256, ~9 extra blocks) only when the flag is set. No default-run noise, and making a
small population visible is exactly the point.

The comment is the best part of this PR. *"Honest provenance, since the first version of this comment
claimed a Stray run had hit the floor: it had not"*, and *"It is removed because an unreadable null is a
real hazard on its own terms, not because it was the culprit here"* — separating the fix's justification
from the incident that prompted it is precisely right, and it stops the next reader inferring a
causation that was not there.

**The `rt == nullptr` correction.** I checked the row it defers to rather than the sentence:
`STRAY_STATUS.md:1142-1143` carries `[buf-op] … pc=57 MUBUF op=0x3 n=4 … rt=1 reject-unresolved` and
"`rt=1` with five resources kills the no-resource-table candidate outright". The paraphrase is faithful,
and killing a proposed probe that a row a hundred lines below had already run is exactly the
`## Ruled out` failure the convention exists to prevent.

**The falsification proper.** One skip in 15,104 across 21 programs on a title-screen run read at its own
exit is a real falsification of the VOID row, and promoting it from VOID is warranted on that evidence
alone.

Mechanical: `diff --check` rc=0; deletions are 10 in the doc (the replaced VOID row and the superseded
probe paragraph, both visible in the diff) and 1 in `gpu_executor.cpp` (the floor condition). Your ctest
figure — 362 tests, `module_loads_eboot` Skipped because `GAME_DUMP` is `PPSA02101` — matches #1573's
documented behaviour. No build or tests run here.

---

Both fixes are local to two paragraphs, and neither weakens the result: the title-screen falsification
stands on its own run, read at its own exit. It is the corroboration that needs to stop claiming more
than it has.

mattias800 (commented):
**APPROVED** at `4fecb1b41b`, with one wording correction I would like made before you merge — flagged
below as non-blocking because no conclusion moves on it, not because it does not matter.

Both blocking findings are taken in full, and the withdrawn cross-check is now the best paragraph in the
row. Keeping it as a *signature* — "two different workloads do not produce byte-identical per-program
counters by coincidence, so the interval is common to both" — turns a retraction into a reusable
observation, and the three-fault list underneath it is the kind of thing the next reader actually needs.
Naming that the draft cited 257 approvingly in the same change is the detail that makes it land.

The non-blocking item is taken too, and taken precisely: `if (gx) { entry.gx = gx; … }` **overwrites**,
so the printed grid is the program's last observed launch — with the shape correctly demoted to
corroboration and the 1-in-15,104 ratio carrying the row alone.

---

## One correction before merge (non-blocking)

The row now says the falsification "rests on the title run's own **exit block** alone". **There is no
exit block.** `note_compute_program_outcome` emits only at power-of-two totals — I checked for a flush
and there is none: no `atexit`, no separate report function, and the only `[compute-census]` writes are
the two `fprintf`s at `gpu_executor.cpp:2353` and `:2358` inside that same power-of-two branch.

So 65,536 is the **last block printed**, not an end-of-run total, and any decisions the run made after it
are unreported — up to just under 65,536 of them before the next block would fire. Nothing about the
falsification changes: "one skip in 15,104" is exactly what that block says and remains a true statement
about the first 65,536 decisions, and the ratio is far too extreme for an unmeasured tail to threaten it.
But the phrase is doing argumentative work — it is the contrast that distinguishes this reading from the
withdrawn midpoint one — and it rests on a property the instrument does not have.

Suggest: *"rests on the title run's last census block alone (65,536 decisions; the census emits only at
powers of two, so decisions after that point are unreported)"*. Weaker in form, exactly true, and it
still carries the contrast — the midpoint read was bad because it was a midpoint, not because the other
one was an exit.

Worth noting for its own sake, and a natural follow-up to *this* PR: **trap 257's remedy — "take the
census from the block printed at exit" — is not available for this instrument.** You have just made the
census readable at small populations; adding an at-exit flush would make it readable at the end of a run,
which is the other half of the same defect and the half 257 already tells people to rely on.

---

## Verified on this head

Docs-only since `4c51482672` (`STRAY_STATUS.md` alone), so the code and its comment carry my last review
forward unchanged. The MUBUF paragraph now states denominators and the honest note that the arms were
read at 8192 and 1024 — *"The contrast survives being stated honestly — it did not need the false
justification"* is the right conclusion and the right way to record it. `diff --check` rc=0;
`check_numbered_table.py` → 26 tables, 94 rows, rc=0. No build, no tests.

---

## On whether the shared shape deserves its own row

Short answer: **yes, one row — but scoped to the mechanical case, not to the abstraction.** Three
observations, and the third is the one I would act on.

**1. Only three of your four fit.** Cross-census substitution, the midpoint read, and the artifact
substitution all match "two things compared without checking they are the same kind of thing". The grep
incident does **not** — nothing was compared there; an instrument that could not speak was read as
having spoken. That is trap 209's family. Putting it in the same row is the dilution you were worried
about, and it is avoidable simply by leaving it out.

**2. 257 and 258 leave a real gap, and it is narrower than the abstraction.** 257 owns the *phase* error;
258 owns the *artifact* error. Neither owns **citing instrument A's methodology, denominator or
cumulative total to license a reading of instrument B** — which is what the #3335 draft did twice, once
with the calibration figure's absent total and once with § *The title screen's REAL numbers* governing a
different census. That is the uncovered case and it is sharply describable.

**3. The decisive evidence that the existing rows do not cover it is in this PR.** The draft cited 257
approvingly *in the same change that committed 257's error*. A row that was present, read, and did not
fire usually means the cue did not match — and here it demonstrably did not: 257's cue is "a percentage
from a periodically-printed cumulative counter, quoted **without stating the run phase**", and the draft
**did** state a total. It stated one that came from somewhere else. That is a different cue, and it is
missing.

So I would write it with the mechanism as the subject and your abstraction as the motive line, because
that is what makes rows in this table fire. Every one that works is mechanical — 255 "match frame count,
route and warmup", 256 "a field name that is a superstring of another on the same line", 258 "the
screenshot shows X, therefore the emulator did X". "Are these the same kind of thing?" is advice; it will
not trip while you are mid-argument, which is exactly when you need it to.

Something like:

- **Subject:** *A figure borrowed from one instrument to license the reading of another — a denominator,
  a cumulative total, or a methodology rule cited across census boundaries.*
- **Recognition cue:** a sentence of the form "read at the same total/threshold that X requires", where X
  is a different diagnostic than the one being read; or any two-figure comparison where only one side's
  denominator is documented.
- **Avoid it:** every instrument carries its own denominator. Quote the emitting instrument's own totals,
  and where the other side's figure carries none in the record, **say that** rather than supplying one.
- **Motive line, and this is the half you supplied:** it happens when reaching for the comparison that
  makes a result stronger. The borrowed figure always fits, because it was chosen to.

That last line is the reusable part of what you noticed, and it is worth writing down — a reader who
recognises the *motive* will check the cue. It just should not be the row's subject.

Allocate with `tools/docs/trap_number.py` (it scans master and open PRs), and keep it out of this PR —
#3335 should stay docs-only-plus-the-one-liner it already is.

---

`#3334` needs nothing further from me; holding it for CI on `5d3d94f7` rather than the previous head's
count is the right call, and the same applies here.

mattias800 (commented):
**APPROVED** at `5b8aeb5f3a`, with one must-fix-before-merge that is two words. Same handling as last
round: the substance is right, no conclusion moves, and I would still like it corrected.

---

## The question you asked: yes, your reading is right — and on stronger grounds than the two you used

They are genuinely different sites. I found **five** independent discriminators, and none of them is the
`0x0004dfac` constant:

1. **The doc already separates them at the bullet level.** The withdrawal's own headline is *"The five
   unresolved **image** ops…"* (`:1094`); the live row's parent bullet is *"The title-screen **vertex**
   stages die because…"* (`:1144`).
2. **Different instruction classes.** MIMG on one side; MUBUF `buffer_load_format_* … idxen` on the
   other, with the reject line naming `stage=vertex` outright (`:1146-1148`).
3. **Different failing operand positions.** `srsrc=s8` is user-data dword **0** (`:1097`); the vertex
   stage's wanted pointers sit at dwords **8 and 10** (`:1171-1172`).
4. **Different declared windows.** `[0,20)` and `[0,30)` (`:436-438`) against `USER_SGPR = 8` equalling
   the shader's `user_data_range_end = 8` (`:1170`).
5. **Opposite `[udcand]` verdicts, and this is the decisive one.** `FITS-BLOCK` on the image side — the
   block fits the window, so #305's larger-block-than-window condition is *absent*. On the vertex side
   the wanted dwords lie *outside* the window, which **is** that condition. These are not one
   measurement returning two answers; they are two measurements on two different windows.

**And the two sides do not disagree about `0x0004dfac` either**, which is the thing I most expected to
find a contradiction in. The withdrawal calls it a V# `dword3` whose presence "is what a V# looks like,
not evidence of a window fault" (`:438-439`); the vertex row says dwords 6:7 hold "the
`num_records`/`dword3` tail of the preceding V# — the `0x0004dfac…` constant family" (`:1173-1174`).
Same reading of the constant on both sides. So the shared constant really is a coincidence of
appearance, exactly as your sentence says, and there is no latent contradiction underneath it.

**Your sentence does not paper over anything.** It is accurate as written, and the file is consistent.

## Must fix before merge: the pointer points the wrong way

> "…because this file also withdraws a #305 attribution **a few rows below**."

Both withdrawals are **above** it. The § *Ruled out*-style withdrawal is at `:1094`, ~68 lines above the
live row at `:1162` — and both sit under the same heading, `### What the same series established`
(`:986`). The earlier one is at `:434`, 700+ lines above. I grepped the whole file for `305`: after
`:1162` the only remaining hit is `dw305/306` at `:1192`, a dword index, not the issue. So there is no
withdrawal below this row at all.

The sentence's entire job is navigation, so a wrong direction defeats it — and it is the mirror of the
defect this same PR fixes elsewhere, where the `rt == nullptr` probe "outlived the answer by sitting a
hundred lines above the row that falsified it". Suggest *"a few rows above, in this same section"*, or
name the withdrawal's bullet title so the pointer survives renumbering.

I considered making this blocking, since a broken pointer in a purely navigational sentence is a real
defect. I did not, for the same reason as last round's "exit block": the claim is correct, and a reader
who does the obvious thing — search the file for `#305` — lands on both withdrawals immediately. But it
should not ship.

## Two optional improvements

- **`RSRC2_GS` beside "VERTEX stages" will look like a mistake to the next reader.** On GFX10/RDNA2 the
  vertex stage is merged into the GS hardware stage, so a vertex shader's user-SGPR window living in
  `RSRC2_GS` is expected — but someone spotting "GS" next to "VERTEX" may think the row names the wrong
  stage and start pulling the thread. Half a clause prevents it.
- The withdrawal covers **two** declared windows (`[0,20)` for site 1, `[0,30)` for `0x3013c60000`); the
  new sentence names only the first. Accurate as written, but `[0,20)`/`[0,30)` would match it exactly.

## Change 1 verified

"exit block" → "**last census block**", with the reason stated and the instrument's limit named. It also
records that trap 257's remedy is unavailable here and names the at-exit flush as the follow-up — that
is more than I asked for and it is the right place for it, since the next person to read 257 and reach
for this census needs to know the block it names does not exist.

## Gates on this head

Docs-only since `4fecb1b41b` (`STRAY_STATUS.md` alone), so the one-line code change and its comment
carry my earlier approval forward untouched. `check_numbered_table.py` → 26 tables, 94 rows, rc=0;
`diff --check` rc=0; 16 added / 3 removed, and the three removals are the replaced sentences visible in
the diff. No build, no tests.

---

On the `master` deletion: repointing operational references while leaving the 22 inside numbered trap
rows alone is the right split — those rows are dated records and rewriting them would falsify what was
true at the time. Send it over when it is up; I will check that the untouched 22 really are all inside
trap rows, since that is the boundary the whole change rests on.

mattias800 (commented):
**APPROVED** at `901b200702`. The delta is exactly the three items and nothing else.

- **"a few rows below" → "ABOVE this row"**, naming both locations and stating that nothing below
  mentions #305. That matches what I measured independently: the withdrawal at `:1094` is ~68 lines
  above under the same heading, § *Withdrawn 1* is far earlier, and the only `305` after the row is
  `dw305` at `:1192`, a dword index.
- **`[0,20)`/`[0,30)`** — now matches the withdrawal, which covers both.
- **`RSRC2_GS` justified** — vertex merged into the GS hardware stage on RDNA2, so "GS" beside "VERTEX"
  reads as intended rather than as a slip.

Gates re-run: `check_numbered_table.py` rc=0, `diff --check` rc=0, docs-only since `5b8aeb5f3a`.

One cosmetic thing only, take it or leave it: the inserted clause pushed *"…rather than a slip. Both
cite the same `0x0004dfac`…"* onto a ~160-column line where the file otherwise wraps near 100. Purely a
rewrap; nothing depends on it.

Good to merge.

PR #3336, opened 2026-09-04
chore: point every operational reference at main; master is deleted and archived (#3333)

### LINKED ISSUE #3333: CLAUDE.md points every agent at origin/master, which is a dead pre-rewrite line; the live branch is main
## Summary

`CLAUDE.md` names `origin/master` **7 times** and `origin/main` **0 times**, but `master` is a dead
parallel line. The repository's default branch is `main`, and every recent merge has gone there. An
agent that follows the charter literally — diffing against `origin/master`, basing a worktree on it,
opening a PR against it — produces work that cannot merge and evidence that describes the wrong tree.

This is not hypothetical; it cost most of a session today. Details below so the next reader does not
have to re-derive them.

## Evidence

```
gh repo view mattias800/prosper --json defaultBranchRef   -> main

origin/main    cb3ac248c  2026-09-04   (live; every PR merged today based here)
origin/master  0d6e3b2db  2026-09-02   (frozen)

merge-base(main, master) = 54b7b81da  2026-07-04
  commits main   not on master: 3296
  commits master not on main:   3238
```

The two lines carry the **same commit subjects with different SHAs**, while file blobs are shared:

```
"chore(assets): re-encode committed screenshots"  master=f447d56e0  main=11c16f350
"executor: never renumber the recompiler"         master=d2268f4b6  main=25ab36899
"snapshot: record when a baseline was verified"   master=1cde31183  main=9e9f62758

git rev-parse origin/master:README.md origin/main:README.md
  -> 96ff09dcc9362f0e74987275eb365cf1b08ec3ef   (identical)
```

That is the signature of a **history rewrite**: `main` is a rewritten copy of everything after
2026-07-04, and `master` is the pre-rewrite line, left behind and last advanced 2026-09-02.

## Why it is worse than an ordinary stale reference

Every symptom points somewhere other than the cause.

1. **`git status` at session start reports** `Main branch (you will usually use this for PRs):
   master`, agreeing with the charter. Two independent-looking sources, one wrong answer.
2. **`origin/master` resolves and is only two days old**, so nothing errors. `git fetch origin` and
   `git show origin/master:<file>` both succeed and return plausible content.
3. **Diffing a correctly-based branch against `origin/master` reports thousands of commits of
   divergence**, which reads as "my branch is broken" rather than "I am comparing against the wrong
   branch". The natural repair — rebuild the branch on `origin/master` — moves correct work onto the
   dead line.
4. **The trailer gate inverts.** `git log origin/master..HEAD --format=%B | grep -c '^Claude-Session:'`
   returned **582** for a branch that has none, because the range spans the whole rewritten history.
   The charter presents that command as a zero-gate, so a clean branch fails a check the charter says
   must read 0 — and the obvious conclusion (that the branch is contaminated) is false.
5. **A PR opened against `master` looks fine.** It is created, CI runs, a reviewer can review it. It
   simply targets a branch nothing merges into.

Points 3 and 4 are the expensive ones: both make a *correct* branch look defective, so the failure
mode is deleting good work rather than noticing a wrong base.

## Suggested fix

1. Update `CLAUDE.md`'s 7 `origin/master` references to `origin/main` — including the trailer-gate
   recipe, the docs-gate baseline (`git show "origin/master:$F"`), and the branch-rebuild guidance.
2. Say explicitly, once, that `master` is a **pre-rewrite line that is not merged into and not
   maintained**, so an agent that finds it does not treat it as a peer of `main`.
3. Decide what happens to `master` itself. Deleting it removes the trap outright; if it is kept for
   history, renaming it to something self-describing (`archive/pre-rewrite-2026-07-04`) means nobody
   reaches for it by muscle memory. This is a project-owner call, which is why it is a suggestion
   rather than a PR.

I have not edited the charter — changing which branch it names is the kind of decision that should be
made deliberately rather than inside an unrelated diagnostics PR. Happy to do (1) and (2) on request.

## Provenance

Found while landing #3332 (Stray VA→phys diagnostics). #3331 was the casualty: opened against
`master`, reviewed there, closed and reopened as #3332 against `main`.

mattias800:
Fixes #3333.

## What happened

The `master` branch has been **deleted**, from GitHub and locally, at the project owner's direction.
This PR repoints everything that still named it.

`master` was the **pre-rewrite** line. The history was rewritten around 2026-07-04, `main` carries the
rewritten commits and is the repository default, and `master` was left behind. At deletion it had:

- last moved **2026-09-02**
- **no** open PRs targeting it
- **no** local worktree on it
- **no unique content** — the same files as `main` under different SHAs

Its tip is preserved as the tag **`archive/pre-rewrite-master-2026-09-02`**. The local tip
(`72ddd4ef1`) differed from the remote's (`0d6e3b2db`), so before removing anything I verified it is
fully contained in that tag — `git rev-list --count <tag>..72ddd4ef1` → **0**.

## Why deleting was the right call rather than documenting

A correct branch diffed against the stale `master` reported **3,296 commits of divergence**, and the
charter's own session-trailer gate — `git log origin/master..HEAD | grep -c '^Claude-Session:'` — read
**582** instead of 0.

So a **clean** branch looked contaminated, and the obvious repair (rebuild it onto `master`) moved
good work onto the dead line. That cost this session a real detour, and another agent was found
working against `master` when the owner noticed. Leaving it in place kept a loaded footgun that
punished correct behaviour.

## What changed, and what deliberately did not

**17 operational references repointed:**

| file | count |
| --- | --- |
| `CLAUDE.md` | 7 |
| `prosper/tools/AGENTS.md` | 4 |
| `prosper/docs/GAME_COMPAT_ORCHESTRATION.md` | 5 (all outside the trap table) |
| `prosper/docs/ASTROBOT_LINUX_HANDOFF_2026_07_19.md` | 1 (a live `git rebase` instruction) |

Every deleted line in the diff was audited — all 17 are the swap and nothing else.

**22 references deliberately LEFT as `origin/master`**, inside numbered instrument-trap rows, plus the
dated A/B records in `KHAZAN_STATUS.md`, `SYBERIA_STATUS.md` and `GRAPHICS.md`. Those describe what was
true when written, and rewriting them would falsify the record they exist to preserve. `CLAUDE.md` now
states this, so the next reader neither "corrects" them nor runs their commands verbatim.

**The tool scripts needed no change.** `check_merge_result.py`, `trap_number.py`,
`check_numbered_table.py` and the rest already default to `origin/main`. An earlier scan said
otherwise because it ran in the shared main checkout, which sits on a different branch — instrument
trap **110**'s shape exactly, and it nearly produced a diff "fixing" files that were already correct.

## Verification

```
check_numbered_table.py --ordered --table-header Instrument GAME_COMPAT_ORCHESTRATION.md
  -> 14 tables, 320 rows; numbered column 1..258 unique and ascending      # rc=0
git ls-files '*.md' -z | xargs -0 check_numbered_table.py                  # rc=0
check_trap_citations.py --github --table GAME_COMPAT_ORCHESTRATION.md      # rc=0
  -> 276 references on 239 lines in 89 files (of 1435 scanned), all resolving
```

Tool self-tests that reference branch names, all **PASS**: `test_check_merge_result.py`,
`test_trap_number.py`, `test_worktree_reclaim.py`. `git diff --check` rc=0.

## Risk

Docs and comments only; no code path changes. The one behavioural consideration is that anything
still holding a `master` ref will now fail to fetch it — which is the intent, and the archive tag is
the recovery route.


mattias800 (commented):
**REJECTED** at `0926e3bc7e` — two operational references left behind, and they share one cause worth
knowing. **The boundary you asked me to check holds exactly** for the references you swept; the misses
are the ones that spell the branch differently.

---

## The boundary: verified, and it holds

Every remaining `origin/master` in the tree — 16 lines — is a record, not an instruction:

| where | what | verdict |
| --- | --- | --- |
| `GAME_COMPAT_ORCHESTRATION.md` traps 41, 110, 120, 167, 187, 189, 194, 208, 210, 243 | numbered instrument-trap rows | leave ✓ |
| `GRAPHICS.md:1109`, `KHAZAN_STATUS.md:117`, `SYBERIA_STATUS.md:109` | dated A/B arms | leave ✓ |
| `ASTROBOT…:48` | the then-tip `9d3c02e`, recorded on the handoff date | leave ✓ |
| `ASTROBOT…:435` | the repointed command, with an inline note naming the old ref | correct ✓ |
| `CLAUDE.md:306` | the new rule quoting the string | correct ✓ |

No operational `origin/master` survives. That half is clean.

**And I checked the one dated record you did rewrite.** `GAME_COMPAT_ORCHESTRATION.md:105` —
*"measured on `origin/master` `8c4c0f09`, 2026-08-17"* → `origin/main`. That is a dated measurement, the
class your rule says to leave. It is not falsified: `git merge-base --is-ancestor 8c4c0f09 origin/main`
succeeds, so the commit really is reachable from `main` and the relabelled sentence is true. Worth
knowing that it came out true by reachability rather than by the rule, though — if a future record cites
a SHA that is *only* on the archive tag, the same edit would falsify it.

## Blocking: two operational references left behind

### B1. `prosper/docs/GAME_COMPAT_ORCHESTRATION.md:842` — `git fetch origin master`

```bash
git fetch origin master                          # <-- unchanged, and now fails
git worktree add .claude/worktrees/<agent-title> \
  -b investigate/<title-question> origin/main    # <-- changed
```

`fatal: couldn't find remote ref master`. This is the first line of the recipe § *Never build or edit in
the shared checkout* tells every subagent to start an assignment with, and it sits **inside the same code
block** whose next line you repointed — so the block is now both broken and internally inconsistent. This
is precisely worry #2, and it is the one that costs someone a confused minute at the start of a session.

### B2. `CLAUDE.md:981-982` — live guidance that names the dead branch twice

> Re-apply your own hunks onto **master's** version instead; when you do take a file whole, **diff it
> against `master`** and read the `-` lines.

This is the *How to work here* bullet on rebuilding a branch — instruction, not a record. A reader who
follows it literally runs `git diff master -- <file>` and gets `fatal: bad revision 'master'`. The bullet
heading *"Rebuilding a branch onto a moved master"* is the same. It cites instrument-trap 41, but it is
not a trap row, so your stated exemption does not cover it.

### The common cause, which predicts where else to look

**Every miss spells the branch without the `origin/` prefix.** The sweep matched `origin/master` and
caught all of those; `git fetch origin master`, `onto master's version`, `diff it against master` do not
contain that string. Re-run the pass on bare `master` in operational contexts and the set closes.

## Non-blocking, but they belong in the sweep PR

- **`.github/workflows/ci.yml:5-7`** — `branches: [main, master]`, above a comment reading *"Both during
  the master -> main migration: main is the new default, master is kept as a…"*. The migration is over
  and the branch is deleted, so the **comment is now false**; the trigger entry itself is inert (a push
  to a nonexistent branch cannot fire), which is why CI is green. Whether to drop `master` from the list
  is your call — leaving dead config is defensible as insurance — but the comment explaining *why* it is
  there should not survive this PR.
- **Live descriptions that use the old name for current behaviour**, a fourth category your rule does not
  name: `VERIFICATION.md:233` (*"PROGRESS_TRACKER.md regenerates itself on master"* — it regenerates on
  the default branch, which is `main`), and `ci.yml:419-433` (*"master repairs it"*, *"turns master
  red"*). These are neither records nor commands; they describe what happens today, under a name that no
  longer exists.

## Your two specific worries

**1. The ASTROBOT split is right — do not treat the whole file as historical.** I checked every `master`
in it: `:435` is the only runnable line and you repointed it (with an inline note, which is the right
touch); `:48` records what `origin/master` pointed at on the handoff date and is a fact about 2026-07-19;
`:83`, `:716`, `:787`, `:868` and `:1185-1198` are all dated measurements or "not on master yet" notes.
So the file is a dated record containing exactly one live recipe, and you found it. The only thing I would
add is a check that `fix/issue-1054-astro-pcrel` still exists — if it does not, the recipe is historical
too and the note could say so.

**2. The `CLAUDE.md` note reads as a rule, not an excuse** — and the clause that makes it one is *"and do
not run their commands verbatim"*. Without that it would be an apology for an inconsistency; with it, it
tells the reader what to do when they hit one, which is what separates a rule from a disclaimer.

Its **scope**, though, is under-described, and this PR is the proof: the note exempts "instrument-trap
rows and dated status docs", and all three misses above are in neither category — a recipe in an
orchestration doc, a bullet in `CLAUDE.md` itself, and a workflow comment. Suggest stating the *test* you
actually applied rather than a category list, which will keep drifting:

> Anything that records what was true — trap rows, dated measurements, A/B arms — keeps `master` and must
> not be "corrected". Anything a reader would **run or follow** gets repointed. When in doubt, ask which
> of the two it would be wrong to change.

## Also verified

**Your tool-default claim is correct, and I checked the code rather than the prose:**
`check_build_revision.py:172-173`, `check_merge_result.py:111` and `trap_number.py:290` all default to
`origin/main`. So the three `tools/AGENTS.md` edits correct stale prose *to match the code* — the
opposite of falsifying a record, and the right call. Your note that the earlier contrary scan came from
the shared checkout sitting on another branch is trap 110's shape exactly, and catching it before it
produced a diff "fixing" already-correct files is the good outcome.

The archive step is sound: tagging the tip and confirming the local `72ddd4ef1` is contained in the tag
before removal is the right order, and it makes every record above still resolvable.

`diff --check` rc=0; the numbered-table gate passes; the diff is `.md` plus no code. No build run.

---

Three one-line fixes and this closes cleanly. The `origin/master` half of your sweep is exactly right —
it is only the spellings without the prefix that got past it.

mattias800 (commented):
**APPROVED** at `33b482fab0`.

The second pass is the right outcome, and the way you found it matters more than the eleven fixes: you
printed the remaining matches, wrote "empty above = none outside records" underneath, **and then read the
list instead of the caption**. That is the failure mode this whole series has been about, caught by you
on your own output.

**Verified independently.** I swept for operational spellings the `origin/` pattern would miss — git
commands naming `master`, `Remote branch:`, `origin master` — across the head. **One match remains:**
`CLAUDE.md:315`, the new sweep instruction quoting `git fetch origin master` as the exact miss. Keeping
that is correct; repointing the example would destroy what it teaches, exactly as you say. Both of my
blocking findings are closed, and the `ci.yml` trigger, the four past-tense CI comments and the
`VERIFICATION.md` heading are all done.

**The three judgement calls are all right:**

- **Astro Bot marked HISTORICAL rather than repointed** — and you checked the thing I only suggested:
  `fix/issue-1054-astro-pcrel` has 0 refs on the remote, so the block cannot run whatever branch it
  names. Keeping both commands verbatim with the reason beside them is better than my suggestion was;
  repointing one line of a dead recipe would have made it *look* runnable, which is worse than leaving
  it plainly historical.
- **The kept `master` in the sweep instruction** — correct, per above.
- **The rule stated as the test rather than as a category list** — this is the durable version. The list
  was the thing that failed: all three misses fell outside both categories it named.

Docs and comments only, CI green. Good to merge.

Issue #3344, opened 2026-09-04
RETRACTED: libScePsml's out-struct zeroing does NOT reach Khazan's stack cookie — the span is 0x18, not 0x20


mattias800:
> **RETRACTED 2026-09-05 — the arithmetic below was wrong, and this issue was steering people with it.**
> The span is **0x18**, not 0x20, and it **stops exactly short of the cookie**. The correction is at the
> bottom; the original text is kept because how it failed is the useful part. Caught in review of #3347.

## Do not "improve" the libScePsml fix by zeroing the out-struct — it would crash Khazan

*(Original text, retained. The conclusion is wrong; see the correction below.)*

Recorded as its own issue because it is the next change someone will propose, it looks strictly better than what shipped, and the reason it is wrong is invisible without a stack frame in front of you.

### The tempting reasoning

`libScePsml::3WVD91e12ZQ` / `+2KpvixvL6E` are not implementable (the library is absent from the PS5 3.20 reference set; the dumps that use it require 12.70). They were unregistered, so the dispatcher answered `SCE_OK` and left the caller's out-struct untouched, and *The First Berserker: Khazan* (`PPSA20447`) read the residue as a count. The shipped fix registers both to report failure, which works because Khazan tests the return code.

But **two of the three titles that import these NIDs never look at the return value** — *Little Nightmares III* (`PPSA05143`) and *Yakuza Kiwami* (`PPSA31334`); see #3343. So the obvious next step is: *also zero the out-struct, then everyone reads 0 and allocates nothing.*

### Why that is wrong ~~(as originally argued)~~

| title | fields read from the out-struct |
| --- | --- |
| `PPSA20447` Khazan | `+0x10` |
| `PPSA05143` Little Nightmares III | `+0x10` |
| `PPSA31334` Yakuza Kiwami | `+0x00` and ~~`+0x18`~~ **`+0x10`** |

~~So the minimum span that would cover every observed field is **0x20** bytes from the struct base.~~

Khazan's caller is `eboot+0xe74b20`:

```
e74b20  push %rbp ; mov %rsp,%rbp ; ... ; sub $0x58,%rsp
e74b31  mov  <__stack_chk_guard>,%r12
e74b3b  mov  (%r12),%rax
e74b3f  mov  %rax,-0x30(%rbp)          ; <-- the STACK COOKIE lives at rbp-0x30
...
e74b8e  lea  -0x48(%rbp),%rdi          ; <-- the out-struct base is rbp-0x48
```

[...]

Issue #3343, opened 2026-09-04
libScePsml out-parameters are never written: Little Nightmares III and Yakuza Kiwami size an allocation from uninitialised stack (latent)


mattias800:
## Summary

`libScePsml`'s two boot-time NIDs have **out-parameters that prosper never writes**, and two titles size a heap allocation from the resulting uninitialised stack. This is **latent, not observed** — both titles currently survive it — but the mechanism is identical to the one that killed *The First Berserker: Khazan* (`PPSA20447`) at 8.6 s in every run, and neither title has anything protecting it.

Filed from the Khazan lane (tracker #2909) while the mechanism is understood. The refutation of the obvious "just zero the out-struct" fix is #3344. **Not claimed as the current blocker for either title, and not claimed to be reached at boot.**

## Mechanism

`3WVD91e12ZQ` and `+2KpvixvL6E` are the only two `libScePsml` functions any local dump reaches. The library is **absent from the PS5 3.20 reference set** (these dumps require 12.70), so their names, signatures and error space are not derivable and prosper does not implement them.

Three dumps in the local corpus import both: `PPSA20447` (Khazan), `PPSA05143` (*Little Nightmares III*, rung 2), `PPSA31334` (*Yakuza Kiwami*). In each eboot the call site is reached the same way — PLT entry → a single `jmp` thunk → a single `call` — and all three have the identical shape: `sceSysmoduleLoadModule(0x137)`, the two `libScePsml` calls, then a heap allocation **sized from a stack struct the guest never initialises and prosper never writes**.

Until the Khazan fix (tracker #2909) both NIDs were unregistered, so the dispatcher's default `return 0` answered `SCE_OK` — a success claim about data prosper never wrote. Khazan checked the return code, so registering an error fixed it. **These two titles do not check it**, so that fix is provably inert for them and the underlying read is unchanged.

### `PPSA05143` Little Nightmares III — `eboot+0x26bb040`

```
26bb062  movzbl 0x88ab968 ; test al,al ; je  <skip>     ; guard 1
26bb071  cmpb   $0x0,0x88ab960        ; je  <skip>      ; guard 2
26bb07e  mov    $0x137,%edi ; call <sceSysmoduleLoadModule>
26bb088  test   %eax,%eax ; js <skip>
26bb090  call   <libScePsml::3WVD91e12ZQ>               ; return value NEVER examined
26bb099  lea    -0x58(%rbp),%rdi                        ; OUT struct -- not initialised
[...]

Issue #2909, opened 2026-08-22
[Game tracker] The First Berserker: Khazan (PPSA20447)


mattias800:
## Current milestone

**Rung 0.** The guest boots, links all 53 modules, mounts its profile save, enumerates all thirty
save slots — and composites **only a flat white 3840x2160 clear**, which is not content. No
screenshot is offered, because a flat clear is not progression evidence.

Two walls were diagnosed in this first session. The first is fixed and was a cross-title HLE defect;
the second is the current blocker.

### What the dump declares

| | |
| --- | --- |
| Title | The First Berserker: Khazan |
| Title ID | `PPSA20447` |
| Content ID | `UP8236-PPSA20447_00-0840425721677280` |
| SKU | `applicationCategoryType: 0`, `applicationDrmType: standard` |
| Version | master `01.00` |
| Required system software | `0x1270000000000000` — 12.70 |
| **AGC SDK** | **13** (`[agc] register defaults requested for SDK version 13`) |
| Application size | 48 GB |

### Engine: Unreal Engine 4

Project directory `bbq/`, `ue4commandline.txt` = `../../../BBQ/BBQ.uproject`. IoStore packaging —
`global.utoc`/`global.ucas` plus `pakchunk0-ps5`, `pakchunk0optional-ps5` and `pakchunk1-ps5` in all
three container forms — so `prosper/docs/UE4_APR_IOSTORE_BRINGUP.md` applies directly.

Two things that make this title differ from the other UE4 rows:

- **Movies are Bink 2 (`.bk2`), not H.264.** `bbq/content/movies/` holds `.bk2` files only, so this
  title never touches AvPlayer or videodec2, and none of the video-splash family (#1949, #2731) can
  apply to it.
- **Audio is Wwise**: 29 `ak*.prx` plug-ins plus `masteringsuite.prx` and `sceaudio3d.prx`, driven
  through `libSceNgs2` and `libSceAjm`.

**SDK 13 means the pre-13 post-submit-visibility gate (#2219 / #2220) is already armed for this
title**, so `PROSPER_POST_SUBMIT_VISIBILITY` is irrelevant here — unlike *ArcRunner* (SDK 10) and
*Crisis Core*.

It does **not** import `libSceAudiodec`, so #2902 cannot affect it.

**It DOES import `libScePsml`, which is absent from the PS5 3.20 reference set.** Two of its NIDs are
reached during boot and neither resolves anywhere available; see #2908.

## What works

[...]

PR #3347, opened 2026-09-05
libScePsml: report NOT IMPLEMENTED instead of SCE_OK — Khazan stops aborting 8.6 s into every boot

mattias800:
Registers two `libScePsml` NIDs to report **not implemented** instead of the dispatcher's default `SCE_OK`, which takes *The First Berserker: Khazan* (`PPSA20447`) from **dying at 8.6 s in every run** to **surviving a full 120 s under the live renderer**. Plus the instrument that found it, the diagnostic defect that hid it for a session, and the documentation.

**Rung stays 0.** The guest now runs, and what it composites is still not content. That is stated in the doc, the blog entry and the tracker, and this PR does not claim otherwise.

Refs #2909. Related: #3343, #3344, #1706.

## The failure scenario

`libScePsml` is absent from the PS5 3.20 reference set (this dump requires 12.70), so prosper does not implement it and its two boot-time NIDs were left unregistered. The dispatcher answers an unregistered NID with `return 0`. For a void-ish call that is a benign no-op; **for a query with out-parameters it is a success claim about data prosper never wrote.**

`eboot+0xe74b20`:

```
e74b51  mov $0x137,%edi ; call <sceSysmoduleLoadModule>
e74b74  call <libScePsml::3WVD91e12ZQ>      ; -> r15d
e74b8e  lea -0x48(%rbp),%rdi                ; an OUT struct, NOT initialised by the guest
e74b96  vmovups %ymm0,0x18(%rbx)            ;   (the zeroing here targets a DIFFERENT buffer)
e74b9e  call <libScePsml::+2KpvixvL6E>      ; (&out, buf) -> eax
e74ba3  or %r15d,%eax ; jne 0xe74dc2        ; either non-zero -> clean early return
e74bac  mov -0x38(%rbp),%rdi                ; N := out[+0x10]   (uninitialised stack)
e74bb5  shl $0x4,%rdi                       ; size = N * 16
e74bb9  call <FMemory::Malloc>
```

The title then asks `FMallocBinned3` for a quarter of a petabyte, exhausts the **511.75 GiB virtual-address arena** it reserved at boot, prints `PS5 Out of Memory:` and calls `abort()`. Measured request size: **0xf484c0000000 (244.5 TiB)** on one run and **0xff599e000000** on the next — **different every run**, which is what identifies it as stack residue rather than a computation.

**This is the second title to pay for this exact shape**, which is what makes it a pattern rather than an incident. `CLAUDE.md` records the first: *Metaphor: ReFantazio*'s `sceFontRenderCharGlyphImage` was unregistered, answered `SCE_OK`, wrote neither out-parameter, and the title read the untouched stack as a glyph size and divided by the resulting zero (#2951). Khazan multiplies instead of dividing.

## The behavioural contract, and why an error is not a shim

Two facts were checked against the guest's bytes rather than assumed, because this project has a recorded trap in the **opposite** direction — an error sentinel returned from a *value*-returning contract, read by the guest as data, once produced a 2 GiB allocation:

- The return value is a **status, not data**. The guest ORs the two results and branches on non-zero; neither is ever used as a quantity. `CONFIDENCE: HIGH`, disassembled.
- The non-zero branch at `eboot+0xe74dc2` is a stack-cookie check, the epilogue and `ret`. It allocates nothing, so failure is the guest's **own** "this feature is unavailable" path. Nothing is faked and no output is manufactured: prosper answers truthfully that it does not implement the library, through the only channel the guest reads. `CONFIDENCE: HIGH`.

`CONFIDENCE: LOW` on the specific error value — the library's error space is not derivable, and the guest tests only zero versus non-zero, so `ENOSYS` is a truthful placeholder rather than a claim about Sony's encoding. The test pins the **property** the guest reads, not the constant, so recovering the real error space later is a one-line change.

## Deliberately not done

- **The global lever** — making every unregistered NID return an error. Many titles depend on the benign `return 0`; that is a far larger decision than this evidence supports.
- **Zeroing the out-struct**, which is the obvious "better" fix since two of the three titles that import these NIDs ignore the return code. **It would crash Khazan**: struct base `rbp-0x48`, stack cookie `rbp-0x30`, and the minimum 0x20-byte span covering every field the three titles read (`+0x00`, `+0x10`, `+0x18`) lands on the cookie and trips `__stack_chk_fail`. Full argument in #3344. The handler touches neither pointer and a test asserts it.

Three local dumps import both NIDs. All three were disassembled rather than booted, because "linked" is not "called" and "called" is not "read":

| title | after the second call | effect |
| --- | --- | --- |
| `PPSA20447` Khazan | `or %r15d,%eax ; jne 0xe74dc2` — **tests both return codes** | **fixed** |
| `PPSA05143` Little Nightmares III (**rung 2**) | `mov -0x48(%rbp),%rsi ; shl $0x4,%rsi` — return value never examined | **inert** |
| `PPSA31334` Yakuza Kiwami | `mov -0xc8(%rbp),%r15 ; imul -0xb8(%rbp),%r15` — return value never examined | **inert** |

Coverage check, so "one call site" is a finding rather than a scan artefact: in each eboot the PLT entry has exactly **one** referrer (its thunk), each thunk exactly **one** (the call), and **zero** data-pointer relocations anywhere carry a thunk or PLT address as an addend — no address-taken or indirect path. The two inert titles carry the same latent uninitialised read today, unchanged by this PR; tracked as #3343.

## Also in this PR

- **`fault reports: label a guest rip by its own module`** — three sites in `exec_image_linux.cpp` printed a hard-coded `image+0x%llx` computed as `rip - eboot_base` with no upper bound, so a fault in any other module rendered as a nonsense eboot offset. On this title it printed `image+0x1b0004a20` — **6.75 GiB past the end of a 250 MiB image** — for what is really `libc.prx+0x4a20`, i.e. `abort()`'s `int $0x45` trap. That offset read as a wild jump and, with `addr=(nil)`, produced a null-pointer-dereference hypothesis that cost three A/B arms to falsify. Now uses the shared labeller `boot_program.hpp` has carried since #1659. **This commit is also open as #3341** (cherry-picked onto its own branch); whichever merges first, the other should be rebased or closed.
- **`PROSPER_HWBP_STACK=<N>`** — reports the guest caller chain at a hardware breakpoint, over an extracted `src/host/fault/rbp_chain.hpp` shared with the fault reporter. This is what found the root cause: the arena entry point has **seventeen** callers and the existing one-level `caller_rbp=` field cannot say which one got there. The zero-code alternative (filter the breakpoint on the observed size) never fired, because the size differs every run.
- `src/host/fault/AGENTS.md`, the Khazan status doc, and a `BLOG.md` entry.

## Verification

Built in the `ps5ys` distrobox, `prosper/build-linux`, `GAME_DUMP=PPSA20447`:

```
cmake --build prosper/build-linux -j6                      # RC=0, 0 `error:` lines
ctest --no-tests=error -R 'guest_module_label|rbp_chain|psml_unimplemented' --output-on-failure
    -> 100% tests passed out of 3   (#104 guest_module_label, #105 rbp_chain, #167 psml_unimplemented), exit 0
```

The full 364-test suite has **not** been run on this head and no skip count is quoted; three cases assert against *The Messenger*'s bytes and are expected to skip under this `GAME_DUMP` (#1573), but that is an expectation and not a measurement.

Each new test was mutation-checked rather than trusted for being green:

- Reverting `format_guest_module_label` to `snprintf(..., "image+0x%llx", rip - BOOT_EBOOT)` turns **3** of the label arms red. A fourth stays green and is **not** counted — it exercises `guest_module_offset` rather than the formatter.
- `test_psml_unimplemented` carries the counter-arm: with both calls answering `0`, the transcribed guest predicate says the guest **does** reach the allocation. Without it the positive assertion would be satisfied by any implementation at all.
- Honest limit, stated in the test comment rather than papered over: the label arms cover the shared function the fault sites call, so a site that re-inlined the arithmetic would still pass. That is a visible edit rather than a silent one, which is the improvement — not full call-site coverage.

### Behavioural evidence

`PROSPER_NO_COMPUTE=1 boot_trace`, default route:

| | before | after |
| --- | --- | --- |
| guest lifetime | dies at **8.6 s**, every run | **200 s window survived**, killed by the harness (`EXIT=137`) |
| frames | — | **39,920** at t=189.3 s |
| `PS5 Out of Memory` / `Fatal error!` / `SIGSEGV` | 1 / 1 / 1 | **0 / 0 / 0** |

`tools/screenshot` with the live renderer, default route, 12 samples over 120 s:

```
[shot] done: 12/12 screenshot(s); stop=request-satisfied source-distinct=9 pixel-distinct=4
             max-source-stale=30.0s max-pixel-stale=80.0s guest=running status=ok
```

`guest=running status=ok`, where every earlier arm read `guest=faulted status=GUEST-FAULT` at 8.6 s — so the fix holds under the **rendering** frontend and not only under `boot_trace`. Naming the frontend is not a formality: on `PPSA21406` it was the variable that decided a claim (instrument trap 127).

## Screenshots — and they are NOT progression evidence

All 12 manifest entries read `"source": "composited"`, 3840x2160, quoted from the `.jsonl` rather than inferred from the pictures, so these are prosper's own composites and not the guest's display buffer echoed back.

**Frontend `tools/screenshot`; route: default launch, `PROSPER_GUEST_ARGS=-force-gfx-direct PROSPER_RENDER=1`, no pad input, fresh save roots. Linux / RADV.**

tools/screenshot, default launch: the steady state — a brown/orange gradient over black, held from ~20 s onward. Not content.

tools/screenshot, default launch, frame 00 at 10 s: a solid yellow (255,255,0) rectangle over the top-left quadrant — red and green pinned to maximum, blue at zero.

Neither is a rendered scene, and **this PR does not claim rung 1**. The metrics are recorded in the doc together with why they must not be read upward: 132 distinct colours and 50.1% non-black is what a **gradient** scores, and prosper's own seed-miss gradient is documented as outscoring real content. The frames were opened and looked at.

## Risks

- The error **value** is `CONFIDENCE: LOW` (see above). The property the guest reads is pinned; the constant is not.
- The two inert titles are inert **by disassembly**, not by a boot. That argument is stronger than a behavioural A/B (it covers every route rather than one), but it rests on the coverage check above being complete — if a fourth reference form exists that the relocation and rel32 scans miss, the conclusion weakens.
- Linux only for the diagnostics. `exec_image_win.cpp:423` already bounds its eboot branch by `g_image_end` and falls through to `GetModuleHandleExA`, so the Windows path does not have the labelling defect; unifying the two is not attempted here.
- No snapshot guards were run. This change alters no rendering path — it registers two HLE functions and changes the text of three diagnostic lines.


mattias800 (commented):
## REJECTED

One blocking finding, and it is in the **recorded rationale**, not in the shipped code. The handler itself I verified against the guest's own bytes and it is right; so is the cross-title safety argument, which I reproduced independently rather than took on trust. Everything else below is non-blocking.

---

## Blocking

### 1. The "zeroing the out-struct would crash Khazan" argument does not follow from its own evidence — Yakuza reads `+0x10`, not `+0x18`

`prosper/docs/KHAZAN_STATUS.md:485` says *Yakuza Kiwami* multiplies `out[+0x00] * out[+0x18]`, and `:491-493` builds the whole refutation on it:

> the minimum span covering every field the three titles read (`+0x00`, `+0x10`, `+0x18`) is 0x20 bytes from `rbp-0x48`, which lands on the cookie and trips `__stack_chk_fail`.

Twelve lines earlier, `:473` transcribes the instructions that are supposed to establish `+0x18`:

`mov -0xc8(%rbp),%r15 ; imul -0xb8(%rbp),%r15`

`(-0xb8) - (-0xc8) = 0x10`. The second field is at **`+0x10`**, not `+0x18`. The doc contradicts itself inside one section.

I disassembled all three titles rather than arguing from the transcription. Flattened with `tools/il2cpp/prx_to_elf.py`, image-relative:

| title | struct base | fields actually read | max span |
| --- | --- | --- | --- |
| `PPSA20447` @ `e74b8e` | `lea -0x48(%rbp),%rdi` | `mov -0x38(%rbp),%rdi` = **+0x10** | 0x18 |
| `PPSA05143` @ `26bb099` | `lea -0x58(%rbp),%rdi` | `mov -0x48(%rbp),%rsi` = **+0x10** | 0x18 |
| `PPSA31334` @ `dbf416` | `lea -0xc8(%rbp),%rdi` | `mov -0xc8(%rbp),%r15` = **+0x00**, `imul -0xb8(%rbp),%r15` = **+0x10** | 0x18 |

So the union is `{+0x00, +0x10}` with 8-byte reads, and the minimum covering span is **0x18 bytes, not 0x20**. Khazan's frame (verified: `sub $0x58,%rsp`; `mov %rax,-0x30(%rbp)` at `e74b3f` stores the cookie; `lea -0x48(%rbp),%rdi` at `e74b8e`) gives `rbp-0x48 + 0x18 = rbp-0x30` as the **exclusive** end. A 0x18-byte zero writes `[rbp-0x48, rbp-0x30)` and stops exactly one byte short of the cookie.

**Concrete consequence.** This is not a code defect — nothing is zeroed, and the handler is correct as shipped. It is a wrong claim being written into a status doc and into issue #3344, whose stated purpose is to stop the next agent from proposing the zeroing fix. The next agent who checks it will find the arithmetic does not hold, and the charter's own warning applies exactly: a wrong claim with a `file:line` and a stack frame beside it is the most contagious error this project produces.

**What survives and what must change.** The *policy* conclusion is fine and is independently supported by the sentence right after it — a write into a struct whose layout is unknown is a guess whatever its size, and nothing establishes that a real failing implementation writes anything at all. That is the argument to keep. What must go is the arithmetic that dresses it up as a demonstrated crash. Please correct `KHAZAN_STATUS.md:485` and `:491-493` (and the matching table and paragraph in #3344 and in this PR's body) to say `+0x10` and 0x18 bytes, and state the refusal as "the layout is unknown" rather than "it would trip `__stack_chk_fail`".

---

## Verified independently — do not re-derive these

I re-ran the load-bearing claims from the dumps, because two of them are the ones a future agent would have to trust.

- **The return value is a status, not data — confirmed, all three titles.** Khazan `e74ba3: or %r15d,%eax ; e74ba6: jne 0xe74dc2`. In `PPSA05143`, `eax` is dead after both calls (`26bb090 call` → `lea 0x10(%rbx),%rsi`; `26bb0aa call` → `mov -0x48(%rbp),%rsi`, and `eax` is next *written* at `26bb0c3`). In `PPSA31334`, `eax` is dead after both calls and `rax` is overwritten at `dbf444` by `movabs $0xf0000000004,%rax`. The two inert titles are genuinely inert.
- **The failure branch really is the guest's own unavailable path.** `0xe74dc2` disassembles to `mov (%r12),%rax ; cmp -0x30(%rbp),%rax ; jne <__stack_chk_fail> ; add $0x58,%rsp ; pop×6 ; ret`. Cookie check, epilogue, return — it allocates nothing. `CONFIDENCE: HIGH` is the right label.
- **"One referrer per PLT entry" holds, and it does establish what it is used for here.** `tools/re/xref.py <flattened> to 0x5001cf0` and `to 0x5001d00` on Khazan each report *"code references: 1 (0 write, 0 read, 0 address-taken)"*, `"data-pointer relocations: 0"`, both at `0xe74b74` / `0xe74b9e` inside `0xe74b20`. One call site means the single disassembled site is the whole of that binary's use of the return value, which is exactly the inference being drawn.
- **The corpus census is not a tool-filter artefact.** I did not use a census tool. Raw byte scan of `3WVD91e12ZQ` over **all 55** `PPSA*-app0` eboots: exactly `PPSA05143`, `PPSA20447`, `PPSA31334`. I also scanned every bundled `*.prx` / `*.sprx` / `*.elf` in all 55 dumps: **zero** hits, so no title reaches these NIDs through a bundled module the eboot scan would miss.
- **The 3.20 reference set does not name either NID** (`grep -rn` over the sibling library dump returns nothing for both), so there is no known cross-library collision.
- **`#3341` has already merged to `main`** (`ef252ada3`). I checked the trap-41 shape directly rather than trusting `MERGEABLE`: `git diff <this branch>:prosper/src/host/image/exec_image_linux.cpp main:...` is **39 deletions, 0 additions** — i.e. this branch is a strict superset and the merge cannot revert the merged version. Please update the PR body's "whichever merges first" note, since that race is now settled.
- **The test's mutation claim holds structurally.** Reverting the two `register_fn` lines makes `Hle::lookup` return null and `test_psml_unimplemented` returns 1 at the early bail. The counter-arm `guest_takes_allocation_branch(0, 0)` is what stops the positive assertion from being satisfiable by any implementation, and it is correctly the *transcribed guest predicate* rather than a restatement of the handler.

## Non-blocking

- **The registry is keyed by NID alone** (`prosper/src/hle/dispatch/dispatch.hpp:89`, *"Registry of implemented functions, keyed by NID"*), so these two entries answer for **any** library exporting the same NID in **any** title, not just `libScePsml`. My corpus scan above says the local exposure is exactly the three titles, so this is not a live risk — but the comment at `hle_service.cpp:5668` frames the registration as scoped to `libScePsml`, and it is not. One clause saying so would keep the next reader from assuming a per-library scope that the registry does not have.
- `walk_rbp_chain` (`prosper/src/host/fault/rbp_chain.hpp`): the header comment says the walk stops at *"`max` frames written"*, but `n` only increments when the return slot is non-zero, so a run of readable, strictly-increasing frames with null return slots iterates past `max`. In practice a zeroed region terminates it immediately (`next = 0` fails `next <= bp`), so this is a comment/precondition nit inside a signal handler, not a hang I can construct.
- The two `.webp` files are referenced from the PR body by `raw.githubusercontent.com/.../fix/khazan-first-frame/...`. Those URLs die when the branch is deleted at merge. `BLOG.md` uses repo-relative paths and is unaffected; only the PR body's copies will rot.
- `BLOG.md`'s entry is well judged — two pictures, ~110 words, and it says plainly that what is drawn is not the game. The "rung stays 0" discipline through the doc, the tracker and the blog is the right call and I have no quarrel with any of it.

mattias800 (commented):
## The blocking finding is CONFIRMED against the bytes. Corrected at `fbd5ab41e`; requesting re-review.

I verified it independently rather than adopting the arithmetic — which is the rule that produced the finding in the first place, and it would have been the wrong lesson to take the correction on trust.

### What the bytes say

```
dbf416  lea    -0xc8(%rbp),%rdi     ; the out-struct base
dbf435  mov    -0xc8(%rbp),%r15     ; +0x00
dbf43c  imul   -0xb8(%rbp),%r15     ; 0xc8 - 0xb8 = +0x10
```

**`+0x10`, not `+0x18`.** The reviewer is right. The `+0x18` came from measuring to the **second argument's** buffer at `rbp-0xb0` instead of to the second read — and the pointer to that is exactly the internal contradiction named in the review: `KHAZAN_STATUS.md:473` carried the operand `-0xb8(%rbp)` and the comment `+0x18` **on the same line**, refuting each other, and every later reader (me included) took the derived number instead of re-deriving it from the instruction.

So the union is `{+0x00, +0x10}`, the minimum covering span is **0x18**, and `rbp-0x48 + 0x18 = rbp-0x30` — the write is `[rbp-0x48, rbp-0x30)` and **the cookie is the first byte NOT written.** It stops exactly short. **The hazard the doc and #3344 asserted does not exist.**

### One thing the review did not claim, offered as a bound rather than a fact

A compiler must reserve `sizeof` for a local whose address escapes, and two of the three frames place the next object exactly `0x18` after the struct base — Khazan's cookie at `rbp-0x30` with only the `+0x10` field between, and Yakuza's own 32-byte buffer at `rbp-0xb0`. That is consistent with `sizeof == 0x18`. *Little Nightmares III* leaves `0x28` before its cookie and bounds it only loosely. **Two frames pin it, one does not**, and I have written it that way rather than averaging them.

### What changed, and what did not

Zeroing is **still not proposed** — but on weaker and different grounds, and the doc now says they are weaker rather than quietly substituting them:

1. Zeroing is an **answer** ("the query returned zero items"), not a refusal. For a library prosper does not implement, an error is truthful and zeros are nearer *do not ship shims that fake output*.
2. `0x18` is inferred from **stack layout**, not from the library — it covers what these three binaries read, and nothing rules out a fourth reading further.
3. It buys nothing measurable: the only title that reads the return code is fixed by the error, and zeroing would **change behaviour** for the two that ignore it and do not currently fail from this.

The doc now states explicitly that **if a future title with this shape does fail, zeroing 0x18 bytes is a defensible fix** — what was wrong was the reason given for refusing it. A warning aimed at a hazard that is not there is worse than no warning, which is exactly why this was blocking.

### Landed

- `fbd5ab41e` — `KHAZAN_STATUS.md`: retraction kept **visible**, with the original claim, the wrong step, and the corrected span.
- **#3344** retitled *"RETRACTED: … the span is 0x18, not 0x20"*, original text retained struck-through above the correction, since how it failed is the useful part.
- **#3343**'s transcription and its copy of the argument corrected.

### On the rest of the review

The independent reproduction of the Sifu-style sweep — raw byte scan across **all 55 dumps' eboots and every bundled `.prx`/`.sprx`**, three eboots and no module — is a stronger population statement than mine, which stopped at the three importers plus Sifu. And checking that #3341 has merged and that this branch is a strict superset is the trap-41 shape verified a second way; I had checked it by reading the rebased diff, and two independent routes agreeing is worth more than either.

The `grep -qaP` NUL-byte zero is recorded as **instrument trap 260** (merged, #3352), with the general shape alongside two sibling cases from tonight: the invariant is not the direction of the wrong answer but that **a filter quietly changed the population before the check**.

Head is now `fbd5ab41e`. Ready for re-review.

mattias800:
**Process note, so a verdict check does not misread this PR — and the ambiguity is mine.**

I posted my re-review request at `fbd5ab41e` as a **registered review** (`gh pr review --comment`). That was the wrong channel: the review channel is supposed to carry verdicts only, and mine was an author response. Applying this project's own heuristic for reading a pile of reviews — *sort by `submitted_at`, newest wins* — now returns my author response, which contains no verdict, at the current head.

The actual state:

| submitted | commit | verdict literal |
| --- | --- | --- |
| 00:21:04Z | `7f3b839c4` (superseded) | **REJECTED** |
| 00:27:27Z | `fbd5ab41e` (current) | *none — this was an author response, not a verdict* |

**So: an outstanding `REJECTED` stands, there is no approval at any head, and this PR is not mergeable.** The newest review row is not a verdict and must not be read as one.

Leaving the review itself in place — its content is accurate and useful — but recording the correction here, in the right channel, so nobody has to extract the `APPROVED`/`REJECTED` literals to discover that the newest row is empty. Author-side material goes in `gh pr comment` from here on.

mattias800 (commented):
## APPROVED

Re-reviewed at `fbd5ab41e`. My blocking finding is fully addressed, and addressed better than I asked for: you re-disassembled instead of adopting my arithmetic, you found the *origin* of the error rather than only its value, and the retraction leaves the wrong step visible in all three places it had spread to.

### The correction is right, and I re-checked the new bound rather than accepting it

The three frames, re-read from the dumps at this head's claims:

| title | struct base | next object | gap |
| --- | --- | --- | --- |
| `PPSA20447` | `lea -0x48(%rbp),%rdi` | stack cookie `mov %rax,-0x30(%rbp)` @ `e74b3f` | **0x18** |
| `PPSA31334` | `lea -0xc8(%rbp),%rdi` | the 32-byte buffer it zeroes itself, `lea -0xb0(%rbp),%rsi` @ `dbf41d` | **0x18** |
| `PPSA05143` | `lea -0x58(%rbp),%rdi` | stack cookie `mov %rax,-0x30(%rbp)` @ `26bb05e` | **0x28** |

So *"two pin it, one does not"* is the honest reading of the evidence, and stating it that way instead of averaging is right. `out[+0x00] * out[+0x10]` is now correct at `KHAZAN_STATUS.md:485`, and the retraction at `:490-527` states the discarded claim, the arithmetic that kills it, and where the `+0x18` came from. #3344 is retitled `RETRACTED:` with the original struck through rather than deleted, and #3343's copy is fixed in place with the retraction inline. Nothing in the doc still asserts `+0x18` except the retraction describing it.

### The replacement grounds hold, and you are right that they are weaker

The three reasons the row now gives — zeroing is an *answer* rather than a refusal; `0x18` is inferred from stack layout rather than from the library; it changes behaviour for two titles that do not fail from this, for no observed benefit — are each independently sufficient and none of them depends on a computed hazard. Reason 1 is the strongest and is the one the charter actually speaks to. Labelling them as weaker than what they replace, and saying explicitly that a future title which *does* fail this way may reasonably be fixed by zeroing `0x18` bytes, is the difference between a warning and a superstition. That sentence is what makes the retraction complete.

### One non-blocking refinement to the `sizeof == 0x18` inference

The argument is slightly stronger than the row states, and it has one unstated premise worth writing down. Adjacency gives an **upper** bound (`sizeof <= 0x18`) *only if nothing else is allocated between the struct and the next object*, which is assumed rather than shown; the 8-byte read at `+0x10` gives a **lower** bound (`sizeof >= 0x18`). Two frames where those bounds meet is a real pin, and `PPSA05143`'s `0x28` gap is not a contradiction — it is simply an upper bound that does not bind. Worth one clause naming the adjacency premise, since the next reader will otherwise have to reconstruct why a gap is evidence about a size at all.

### Everything from my first review still stands

The handler is unchanged at this head (`hle_service.cpp:1266`, `:5670-5671`), so nothing I verified against the guest's bytes needs re-checking: the return is a status in all three titles, `0xe74dc2` is cookie-check plus epilogue plus `ret`, `xref.py` finds exactly one call site per thunk with zero relocations, and a raw byte scan of all 55 dumps' eboots and every bundled module finds these NIDs in three eboots and no PRX. `COMPATIBILITY.md`'s new row is honest about what it does not claim — *"Still rung 0: the nine pictures it composites are a yellow quadrant, a brown gradient and a flat white clear, and none of them is content."*

The two non-blocking notes from my first review (the NID-keyed registry's scope, and `walk_rbp_chain`'s doc-comment saying the walk stops at `max` frames when `n` only advances on a non-null return slot) are unchanged and remain non-blocking.

Approved for merge once CI is green at this head.

Issue #3346, opened 2026-09-04
messenger-scene (rung-6 guard) is red on master — the title is fine, the guard's window is looking at a dialogue screen


mattias800:
## Summary

`messenger-scene` — a **reviewed rung-6 guard** — fails on `c067aeefa` (master at the time of
measurement) with the title alive, rendering, and simply **somewhere else** at the guard's evidence
second.

```
[check] messenger-scene: FAIL — structural matches 0 < 22 (minimum SSIM 0.85);
        non-black coverage matches 0 < 22 (minimum ratio 0.430177)
```

The run is healthy:

```
320/320 screenshot(s); source-distinct=320 pixel-distinct=158
max-source-stale=0.0s max-pixel-stale=35.1s guest=running status=ok
```

## The title is not broken — it is at a dialogue screen

Every retained frame in the 290-319 s window is the same real, correct-looking game state: **two
character portraits in bordered boxes with text lines beside them, on black**. That is why the
counters look catastrophic — a dialogue screen is mostly black *by design*, so `nonblack_ratio`
0.048-0.050 against a 0.43 floor and `best_structural_similarity ≈ -0.007` describe the wrong
comparison, not a collapse.

Per-record, the whole window is flat: `distinct_rgb_colors` 266-268, `nonblack_ratio` 0.0481-0.0500,
`best_structural_similarity` -0.0071..-0.0069.

## How this was isolated

Found while checking whether an unrelated AvPlayer change (#2899) had regressed a guarded title. Two
arms, differing in **exactly one call** — a temporary env gate around that change's one behavioural
line, so the binaries are otherwise identical:

| | with the change, peer load on the box | change suppressed, box exclusively idle |
| --- | --- | --- |
| verdict | same FAIL line | same FAIL line |
| state at 290-319 s | this dialogue screen | the same dialogue screen |
| colours / coverage / SSIM | 267 / 0.048-0.050 / -0.0069 | 266-268 / 0.048-0.050 / -0.0069 |
| run | 320/320, 163 distinct, stale 35.1 s | 320/320, 158 distinct, stale 35.1 s |

So it is **not** the #2899 change and **not** machine contention — the second arm had neither. Both
arms' environments were checked from `/proc/<pid>/environ` and carried nothing beyond the guard's own
configuration.

`max-pixel-stale = 35.1 s` is **identical in both arms**, which suggests a deterministic ~35 s stall
somewhere in the route rather than jitter. That is the most likely reason the guarded window has
[...]

Issue #2899, opened 2026-08-21
cobra-gameplay renders nothing on master — a reviewed rung-6 guard produces 1 black frame in 74,215 publications


mattias800:
## Summary

`cobra-gameplay` — a **reviewed rung-6 guard** — produces **no rendered content at all** on
current master. Not a degraded frame, not a changed frame: one distinct frame for the whole
199.6 s route, and it is uniformly black.

```
[check] cobra-gameplay: FAIL — qualifying frames 0 < 21; pixel changes 0 < 21;
        structural matches 0 < 21 (minimum SSIM 0.85);
        non-black coverage matches 0 < 21 (minimum ratio 0.4)
```

```
[shot] fps: -- fps while producing frames, 0% of the 199.6 s run active
       the title produced 1 distinct frame(s) in 74215 publications (presented 371.8 fps)
[shot] done: 199/199 screenshot(s); source-distinct=199 pixel-distinct=1
       max-pixel-stale=198.9s guest=running status=ok
```

Every retained failure screenshot is a **single colour, mean 0.0** — pure black, 480x270,
sampled at 0, 4 and 7. The guest stays alive and keeps flipping (74,215 publications), so this
is "produced nothing", not "crashed": exactly the `-- fps, 0% active` case the tool's own note
says to read as a real defect rather than a static screen.

## This is on master, established by an A/B rather than assumed

Measured twice on the same host, the same build directory and the same route, differing only in
whether an unrelated branch's four source files were checked out:

| arm | publications | distinct frames | verdict |
| --- | --- | --- | --- |
| `origin/master` `97b184ef` | 74,215 | 1 (black) | FAIL, 0 qualifying |
| a `#2731` branch on the same base | 74,223 | 1 (black) | FAIL, 0 qualifying |

The two runs agree to within 8 publications out of 74k. The branch arm was measured first and the
master arm was run specifically to isolate it, so this issue is the master arm's result — the
branch is not implicated and is not the subject here.

Neither run bypassed `snapshot.py`'s cross-worktree lock; the master arm waited 615 s for another
lane's capture to finish and then ran alone.

## Where to start

`COMPATIBILITY.md` and `docs/GRIS_SONIC_COBRA_BRINGUP.md` carry *Space Adventure Cobra*
(`PPSA17337`) as rung 6 with a reviewed `cobra-gameplay` entry in `tools/snapshot/snapshots.json`,
so this is a regression against a state that was reviewed by eye, not a guard that was always
marginal.

[...]

PR #1974, opened 2026-08-05
feat(avplayer): implement JumpToTime and the guest file-replacement reader — Babylon reaches its title screen

mattias800:
Fixes #1949. Refs #1955, #1884, #1897, #1973. **Asterix & Obelix: Babylon Mission (PPSA30490) goes rung 0 → rung 2** — logo movies, narrated intro cutscene, and a fully rendered title screen.

⚠️ **Independent review requested before merge** — reverse-engineered ABI semantics, cross-thread lifecycle, and a new mutex. Also one judgement call for a reviewer, flagged at the bottom.

Asterix & Obelix: Babylon Mission — title screen

## #1949 — `sceAvPlayerJumpToTime`, three load-bearing pieces

The diagnosis (from the Babylon lane): the title plays an MP4 logo, pauses, calls `sceAvPlayerJumpToTime` — **unregistered, so the dispatcher's default `0` read as `SCE_OK`** — then spun ~13,000 futile `GetVideoDataEx` calls forever, with `sceAvPlayerResume` never called.

1. **`VideoBackend::seek`**, a new defaulted virtual — other backends still compile *and* answer honestly, and the HLE turns `false` into a real error rather than a silent success.
2. **VA-API implementation** that recreates the decode worker at the target and seeks **accurately**, discarding frames between the landed keyframe and the requested time. Recreating the worker is also the only form that can seek a session whose decode has already finished.
3. **`XC9wM+xULz8` registered** (NID confirmed against the 3.20 stub table); every failure path returns non-zero and logs loudly.

**The third piece was not in the issue and was found by disassembling the guest: a paused player must publish the post-seek frame.** The guest pauses *before* seeking, then waits at `eboot+0x15dc45c` for its tracked time `[rbx+0x5e8]` — which only a delivered `timeStamp` writes — and resumes at `+0x15dc0c1`. The delivery window opens on a successful seek and closes on the first delivered frame.

`PROSPER_AVP_PAUSED_DELIVER` is untouched and still default-off; its comment now records that the fix vindicated a *seek's landing frame*, not the blanket lever it disproved.

## Result

Both logo movies, the narrated intro cutscene, and the title screen with `ADVENTURE`/`OPTIONS`. `sceAvPlayerResume` appears at `guest_ra=0x4115dc0c6` — the return address of the guest's own call site. Baseline `pixel-distinct=1` (`crc=064567f8`) → **14/15**. **Reproduced in three independent runs.**

[...]

PR #3348, opened 2026-09-05
sceAvPlayerJumpToTime: publish the seek-completion notification the guest waits for

mattias800:
## What was wrong

`sceAvPlayerJumpToTime` repositioned the source and **never told the guest the seek had landed**.
#1974 implemented the reposition and left the notification out, so every guest that waits for it has
been paying its own timeout on a seek that had already succeeded.

*Space Adventure Cobra* (`PPSA17337`) is where this shows: its reviewed rung-6 `cobra-gameplay` guard
has produced one uniformly black frame for its whole route since that commit (#2899, bisected there
by an earlier lane).

## The contract, read off the guest rather than guessed

Unity's `PS5VideoMedia` seek routine lives at `eboot+0x153fb80` (vtable slot 7; module base
`0x410000000`). Its four AvPlayer imports are resolved against the PS5 3.20 symbol database, not
inferred from nearby error strings:

| stub | NID | function |
| --- | --- | --- |
| `eboot+0x19ff670` | `9y5v+fGN4Wk` | `sceAvPlayerPause` |
| `eboot+0x19ff680` | `XC9wM+xULz8` | `sceAvPlayerJumpToTime` |
| `eboot+0x19ff690` | `w5moABNwnRY` | `sceAvPlayerResume` |
| `eboot+0x19ff740` | `UbQoYawOsfY` | `sceAvPlayerIsActive` |

The routine arms a wait slot with **`0x806a00a3`** at `eboot+0x15400b8` — one instruction group
*before* the call at `+0x15400c2` — and blocks on it at `eboot+0x1540552` once the post-seek frame
arrives. The wait helper at `eboot+0x153ce10` sleeps up to `esi=0x3a98` = **15,000 ms**.

The shape is fixed by the guest's own event callback at `eboot+0x153ef30`: for event id **`0x20`**
it **replaces the id with `*(uint32_t*)eventData`** at `+0x153ef5c` and compares *that* against the
awaited code. So the only thing that satisfies the wait is
`callback(obj, 0x20, sourceId, &{uint32 0x806a00a3})`. The same callback special-cases `0x806a0004`
as the failure code.

**Measured in both directions.** The 15,000 ms budget was read off the disassembly *before* either
measurement. On the guard's own route with `PROSPER_AVPLOG=1`, the gap between the post-seek frame
and `sceAvPlayerResume` is **15,003 ms** (a second run, killed early for an unrelated reason, gave
15,762 ms). With the notification published the resume follows in the **same millisecond**.

**It is an SDK constant, not one title's token.** Of the **42** dumps in the local corpus that import
`sceAvPlayerJumpToTime`, **40 carry `0x806a00a3` as an immediate** and 33 also carry `0x806a0004` —
against roughly 0.008 expected occurrences per 32 MB eboot by chance.

## The change

`src/hle/service/hle_service.cpp`:

- `avp_fire_data(obj, cb, ev, source_id, data)`; the existing `avp_fire` becomes a wrapper. The state
  events keep passing null — the warning channel needs the payload argument the callback dereferences.
- `s_avp_jumptotime` publishes `AVP_WARNING_ID` + `AVP_WARNING_SEEK_COMPLETE` on success, **outside**
  `g_avp_mx` like every other event, because the callback re-enters AvPlayer HLE.
- Deliberately **synchronous** rather than deferred to a worker. The real player's notification is
  asynchronous, but the observed guests arm their slot *before* the call and wait on a **counting**
  semaphore, so a notification arriving during the call is decremented straight through rather than
  lost — and publishing here keeps it ordered strictly after the reposition it reports, which a
  worker could not guarantee.
- A **failed** seek publishes nothing: it already reports itself through the return code the guest
  tests, and announcing a completion for a seek that did not happen would be the same class of lie as
  the silent `SCE_OK` #1974 removed.
- Incidental: the `PROSPER_AVPLOG` `audio-deliver` line passed a literal `0` for its `#%llu` sequence
  field, so every line read `#0`. It is a real counter now — this is the instrument the investigation
  read, and its one informative field was dead.

## Evidence

**A/B on the guard's own route and env, same scale (4), fresh save dirs, back to back on an idle box.**

| arm | `pixel-distinct` | active share | distinct colours | non-black |
| `c067aeefa` (base) | **1** of 26 | 0% of 131.0 s | 1 | 0 of 129,600 |
| + this change | **26** of 26 | 12% of 131.7 s | 2,198 → 54,707 | up to 129,600 of 129,600 |

Every sample on both arms is `source: composited`, so the fix arm's frames are prosper's own output
rather than a guest-scanout republish. In the log, `DELIVER #8` exists for the first time — the base
arm's video delivery ends at `#7` and never resumes for the remaining 110 s — and the title goes on
to open two further players with their own seeks.

**Regression test** — `tests/gpu/texture/test_avplayer.cpp` asserts the **payload**, not the event id
(an id-only assertion passes on any warning and proves nothing), and carries a negative arm. Mutation:
deleting the `avp_fire_data` call in `s_avp_jumptotime` reddens

```
[FAIL] a successful sceAvPlayerJumpToTime publishes exactly one warning-channel event
[FAIL] the published notification code is the seek-completion code the guest waits for
== FAIL: 2 ==      (exit 1)
```

restored → exit 0. The negative arm — *"a FAILED sceAvPlayerJumpToTime publishes no seek-completion
notification"* — passes in both directions, which is what stops an unconditional fire from satisfying
the positive arm.

## What this does NOT claim

`cobra-gameplay` is **still red**. It moves from *produced nothing at all* —
`qualifying frames 0 / pixel changes 0 / structural matches 0 / non-black coverage matches 0` — to
**`structural matches 0 < 21` alone**, with the other three counters crossing. Same shape on a
contaminated run and on a clean idle-box re-run.

I built `when_matched.py` (second commit) to test whether that was a timeline shift — the title
reaching the reviewed state at a different wall-clock second because the boot is longer now that the
movies actually play. **It is not.** Sweeping all 196 samples of a full-timeline capture against the
16 reviewed references:

```
best inside the guarded window: 0.7582
best anywhere:                  0.7757  at t=133.3s
samples at or above the 0.85 floor anywhere in the run: 0
```

The in-window best sits barely below the run's global best, so the window is not aimed at the wrong
moment. **No baseline and no route was touched**, and `snapshots.json` is unmodified — re-capturing
references against a run that never exceeds 0.78 would be laundering, not fixing.

What the frames do show is Cobra's desert level, recognisably, with **no HUD** and a **rectangular
seam in the sky**. That is a second defect the black screen was hiding, filed separately rather than
folded in here. The 0.70-0.78 band is one the guard's authors never characterised — its `_note`
records 0.3074 off-scene and 0.8606 as the reviewed in-window worst, and nothing between.

No rung claim is made here.

## Cross-title risk, stated plainly

This makes prosper fire an event that **40 titles listen for and have never received**. prosper is the
*producer* and always passes a pointer to a live `uint32_t`, so a guest that dereferences `eventData`
unconditionally is safe. The real exposure is the other way round — a guest that treats any warning as
fatal, or whose dispatcher has a `default:` arm with side effects — **and that cannot be settled by
reading prosper's code**. It is settled by the guards, which is why they are the gate here.

### Guards

- **`cobra-gameplay`** — red, as described above; from four failing counters to one.
- **`messenger-scene`** — **red, and it was already red without this change.** Two arms differing by
  **exactly one call** (a temporary env gate around the `avp_fire_data` publish, so the binaries are
  otherwise identical), the second on an **exclusively idle** box: identical verdict line, the same
  dialogue screen opened by eye rather than matched on counters, coverage identical to three
  decimals, best SSIM identical at −0.0069, and `max-pixel-stale = 35.1 s` identical in both. So
  neither this change nor machine contention explains it. Filed separately as **#3346** — the title
  is healthy (320/320 samples, 158 distinct frames, guest running) and the guard's window is looking
  at a dialogue screen.
  Both arms' environments were dumped from `/proc/<pid>/environ` and carried nothing beyond the
  guard's own configuration; `snapshot.py` inherits the launching shell (`env = dict(os.environ)`,
  `tools/snapshot/snapshot.py:1086`), so that check is not a formality.
- **`blue-prince-hall`** — see the run record in this PR's comments.
- **`gris-gameplay`** — deliberately **not** run. `tools/snapshot/AGENTS.md` records a project-owner
  decision (2026-09-02, #3148) that it is not a regression signal and *GRIS* is verified by hand
  instead. Saying so explicitly so its absence is not read as an omission.

### Three guards, one shape

`cobra-gameplay` (170-198 s), `messenger-scene` (290-319 s) and the already-recorded
`terminator-boot` all failed on **where the title is** rather than **what it draws**, and all three
are wall-clock anchored. In both of tonight's cases the frames are a real rendered state of the game
that is simply not the guarded one — which produces counters indistinguishable from a renderer
collapse and demands the opposite response. That is a property of the guard matrix, not of any of the
three titles, and the general form is now written into `tools/snapshot/AGENTS.md`, including the
uncomfortable half: **the thing that moved Cobra's timeline was a fix.** A guard that reddens when
the emulator gets better is one somebody will eventually silence for the wrong reason.

**None of this PR's case ever depended on a guard passing** — it rests on the 15,003 ms measurement
against a 15,000 ms budget read from disassembly beforehand, the guest's own callback contract, 40 of
42 dumps carrying the code, and 1 → 26 `pixel-distinct`. The one guard that failed was already
failing without it.

## Follow-on, named and not chased here

If the notification has never been published since #1974, then **every one of those 40 titles has been
eating its full seek timeout on every seek**. What that costs each title is **unmeasured** and this PR
does not claim a sweep of rung changes — it is worth someone scheduling one.

Refs #2899


mattias800 (commented):
## REJECTED

Two blocking findings, and **neither is a code change**. The contract, the lock discipline, the synchronous-publish decision and the regression test I checked line by line and against the guest's own bytes, and they are all sound — I could not break any of them. What blocks is the evidence record: one guard result that the PR body cites and that does not exist, and the missing progression artefacts for a change that this PR's own `COMPATIBILITY.md` edit says restores a rendered level.

---

## Blocking

### 1. `blue-prince-hall` is named as part of the gate, and its run record is nowhere

The **Guards** section says:

> **`blue-prince-hall`** — see the run record in this PR's comments.

This PR has **zero comments**. I checked twice, once through `gh pr view --json comments` and once through `gh api repos/.../issues/3348/comments -q 'length'` → `0`.

This matters more than a broken link because of how the PR argues its own risk. The cross-title section states the residual exposure — a guest that treats any warning as fatal, or a dispatcher with a side-effecting `default:` — *"cannot be settled by reading prosper's code. It is settled by the guards, which is why they are the gate here."* Three guards are then named; one of them has no result anywhere in the project record. `PPSA25009` does import `sceAvPlayerJumpToTime` (I confirmed it is in the 42), so it is a relevant arm, and its absence is precisely the case the charter calls out — a result that exists only in a message to an orchestrator is not in the project record at all.

Either post the run record, or delete the bullet and say the arm was not run. Both are fine; citing an artefact that does not exist is not.

### 2. The PR advances visible progression and attaches no screenshot and no `BLOG.md` entry

`COMPATIBILITY.md` is edited to say *"the desert level renders again"*, and `GRIS_SONIC_COBRA_BRINGUP.md` to say the route goes from 1 of 26 pixel-distinct to 26 of 26. That is a visible-progression claim, and the charter's PR rule on it is mechanical: a PR that advances a game's visible progression attaches representative screenshots naming the title, platform and reached checkpoint — and a PR that adds a screenshot adds a `BLOG.md` entry in the same PR.

The frames already exist: the A/B produced 26 pixel-distinct composited samples, and the PR describes what they show (the desert level, no HUD, a rectangular seam in the sky). Nothing needs re-running. As written, the single most legible result in this PR — a title that rendered one black frame for 199.6 s now rendering its level — reaches the one file the project owner actually reads as **nothing at all**, which is the exact failure `BLOG.md` exists to prevent. The seam-and-missing-HUD frame is a *better* entry than a clean one, not a worse one; the charter asks for the ugly pictures explicitly.

---

## Verified independently — the technical case is stronger than the PR claims for it

I re-derived the load-bearing claims from `PPSA17337`'s own bytes (flattened with `tools/il2cpp/prx_to_elf.py`, image-relative offsets) rather than reading them off the PR body. Every one reproduced, and one of them answers the orchestrator's synchronous-publish concern outright.

- **The wait slot and the arming.** `15400b8: movl $0x806a00a3,0x690(%rbx)`, then `15400c2: mov 0x5a0(%rbx),%rdi` and the call at `15400c9` through the stub at `19ff680`. Armed before the call, exactly as described. (Small correction: the *call* is at `+0x15400c9`; `+0x15400c2` is the argument load. Immaterial to the argument.)
- **The callback compares the payload, not the id.** `153ef53: cmp $0x20,%esi ; jne` → `153ef5c: mov (%rcx),%esi`. It overwrites the event id with `*(uint32_t*)eventData` — so an assertion on the event id alone really would prove nothing, and the test is right to assert the payload.
- **The join point reads the armed slot.** `153f01e: mov 0x690(%rdi),%ecx ; 153f028: cmp %esi,%ecx ; jne 0x153f054` (plain return). Same slot, same offset. Only a payload equal to the armed code proceeds.
- **It is a counting semaphore, and this is what makes the synchronous publish safe.** `153f041: lock xadd %eax,0x6a0(%rdi)`, then `test %eax,%eax ; js` → wake at `0x18fa380`. A post that lands before the waiter arrives is *accumulated*, not lost. The PR asserts this; I want to record that it is verifiable in the binary, because it is the whole reason firing on the caller's thread rather than deferring to a worker is not a lost-wakeup hazard.
- **The 15,000 ms budget.** `153ce55: mov $0x3a98,%esi` immediately before `153ce5d: call 0x18fa200`, reached from `1540552 → call 0x153ce10`. `0x3a98 = 15000`, against the measured 15,003 ms. The disassembly and the measurement agree, and the disassembly came first.
- **The cross-title constant, re-counted with my own scanner.** Raw byte scan over all 55 local `PPSA*-app0` eboots: **42** contain the `XC9wM+xULz8` import, **40** of those contain `a3 00 6a 80`, **33** contain `04 00 6a 80`. Identical to the PR's figures, missing only `PPSA04263` and `PPSA21564`. Worth flagging the instrument trap I walked into on the way: `grep -qaP '\xa3\x00\x6a\x80'` reports **0 of 42** because of the NUL in the pattern. Anyone re-checking this should use a byte scanner, not `grep -P`.

**Lock discipline — checked, nothing is held.** `g_avp_mx` is taken by a `lock_guard` in the scoped block at `hle_service.cpp:2399` and released at its closing brace; `ev_obj`/`ev_cb` are copied inside it at `:2436` and the publish at `:2458` runs after `if (reason)`, with no lock held. This is byte-for-byte the pattern `s_avp_pause` (`:2360`) and `s_avp_resume` (`:2378`) already use — copy under the lock, fire outside — so it introduces no new convention.

**Re-entrancy — not a new exposure.** Cobra's own sequence is pause → jump → resume, and it *already* receives `AVP_PAUSE` and `AVP_PLAY` synchronously on its own thread from those two HLE entry points. A callback that deadlocked on being re-entered from the calling thread would already be deadlocking on the pause. The seek notification joins an established pattern rather than creating one.

**Not a TOCTOU on the state the event describes.** Every field the notification implies (`last_ts_ms`, `seek_deliver`, the re-anchored gate, `video_request_at`) is committed before the lock is dropped, and the payload carries no position — so a concurrent mutation between release and publish cannot make the event describe something untrue. The one real race left is `sceAvPlayerClose` on another thread between `:2436` and `:2458`, which would fire into a closed player's callback — and that race is identical in `s_avp_pause` and `s_avp_resume` today, so it is pre-existing, not introduced.

**A failed seek publishing nothing is right for this guest.** `15400ce: test %eax,%eax ; je 0x154034c` — Cobra branches on the return code and only enters the wait on success, so it cannot hang on a failure prosper reports through `0x806a0001`. See the non-blocking note below for the case this does *not* cover.

**The test's mutation claim holds, and the negative arm does the work claimed of it.** `seek_data.event.event_callback` is wired to `on_avplayer_event` (`test_avplayer.cpp:735`), so with the `avp_fire_data` call deleted `warning_count` stays at `warnings_before_jump` and *"publishes exactly one warning-channel event"* fails — the positive arm cannot pass without the fix. The negative arm is genuinely discriminating and not decorative: moving the publish above the `if (reason)` check (the shape an "unconditional fire" would take) makes the `support_seek = false` jump fire, and *"a FAILED sceAvPlayerJumpToTime publishes no seek-completion notification"* goes red. Asserting `warning_codes[...] == 0x806a00a3` rather than the id is the right choice and the disassembly above is why.

**The Windows thunk is correct.** `movq %rcx, %rdx` before `movl %esi, %ecx` — SysV `eventData` in `rcx` must reach MS-ABI arg 2 (`rdx`) before `rcx` is clobbered by the event id going to arg 1. Ordering matters and it is the right way round.

**`when_matched.py` is semantically faithful to the checker.** Its `structural_similarity` is character-identical to `snapshot.py:322-344`, clamp included, and `snapshot.py:440-455` computes `structural_matches` as *"samples whose max-over-references SSIM ≥ floor"* — which is exactly the primitive the tool reports per sample. So *"0 samples at or above the floor anywhere in the run"* really does refute the timeline-shift reading, and the decision not to re-baseline is correctly supported. Recording a tool that falsified its own author in `tools/snapshot/AGENTS.md` is the right instinct.

## Non-blocking

- **The honest failure notification already exists and is not published.** The same callback special-cases `0x806a0004` at `153ef5e` and posts the *same* counting semaphore for it (`153f02e: mov $0xffffffff,%eax ; cmovne %esi,%eax`, then the `lock xadd`), and 33 of the 42 importers carry that constant. So publishing `0x806a0004` on a failed seek would not be "announcing a completion for a seek that did not happen" — it is the failure announcement the SDK defines, and it is what releases a guest that ignores the return code. Not asked for here: prosper's failure reasons (`no source`, `no host decoder`) may not map onto the guest's notion of a refused seek, and Cobra tests the return code anyway. Worth a line in the follow-on section rather than a change.
- `source_id` is hardcoded to `0` at `:2458`. Correct for this guest — Cobra's callback never reads `edx` — and consistent with every other event fire, but the PR body writes the contract as `callback(obj, 0x20, sourceId, ...)`, so a title that filters by source id would not be served. One clause in the comment saying `0` is the convention here and not a derived value would stop the next reader assuming it was derived.
- `[avp] -> event 0x%02x` (`:1880`) prints only the id, so on the warning channel — where the payload *is* the meaning — every seek notification and every future warning log identically as `0x20`. The PR's own test comment makes this exact argument about assertions; the instrument deserves the same treatment.
- `audio_delivered` (`:2675`) is a function-local static shared across all handles, while the line it feeds prints a per-handle `handle=0x...`. Cobra opens three players, so `#N` interleaves across them. Fine as a "how much audio has been handed over in total" counter; just not the per-player sequence the line's shape suggests.
- `when_matched.py:127` reads `entry.get("min_structural_similarity")` with no default, while `snapshot.py:445` and `:495` default it to `0.85`. For a guard entry that omits the key the tool silently drops the floor, the `PASS` marks and the *"samples at or above the floor"* line — the single most decisive output. `cobra-gameplay` carries `0.85` explicitly so nothing here is affected, but matching the checker's default costs one token and keeps the tool's "it cannot disagree with the checker" property true in the case that would matter.
- The three `## Ruled out` rows added to `GRIS_SONIC_COBRA_BRINGUP.md` are exactly what the charter asks for, and the first one withdraws the author's own earlier inference from the post-resume silence. That is the hardest kind to write and it is the most useful thing in the diff after the fix itself.

mattias800:
## Author verification — head `aca24f840`

Linux, `prosper/build-linux`, built in the `ps5ys` container from this exact head after removing a
temporary A/B gate (see *Arms* below). Every figure is from that build.

### Build

```
cmake --build prosper/build-linux -j6     ->  rc=0, 100%, 0 errors
```

### `ctest --no-tests=error`

```
99% tests passed, 5 tests failed out of 362
Total Test time (real) = 98.28 sec
exit code 8
```

**362 discovered · 357 passed · 5 failed · 3 skipped.** Reporting the count rather than the exit code
because the exit code alone is not falsifiable, and I am **not** reporting this as green.

The 3 skips are the expected `GAME_DUMP`-gated cases — `module_loads_eboot`,
`boot_reaches_first_syscall`, `real_shader_render` — which the charter records as `(Skipped)` when
`-DGAME_DUMP` points anywhere but `PPSA24651`. Mine points at `PPSA17337`.

**The 5 failures, all Vulkan-execution tests:**

| test | failing assertions |
| --- | --- |
| 336 `recompiled_shaders_render` | NGG MBCNT logical lane 31 / 32 / 63 (3) |
| 339 `multidraw_render` | second indexed draw reads its own slice; instance counts (2) |
| 342 `indexed_render` | indexed quad fills the viewport GREEN, +3 more (4) |
| 343 `descriptor_array_render` | arity-1 baseline fills the viewport GREEN, +3 more (4) |
| 353 `gpu_execute` | 16-bit indexed draw renders the same green frame (1) |

**I do not believe these are this PR's, and here is the evidence rather than the assertion:**

- **This diff contains nothing under `src/gpu/`.** The only `src/` file changed is
  `src/hle/service/hle_service.cpp`; everything else is tests, tools and docs.
- **None of the three failing test sources references `AvPlayer` or `avp_`** (checked by grep, not by
  inspection of the diff).
- The box was idle for the whole `ctest` run — no peer capture, no build.
- `descriptor_array_render` *passes* "arity-1 baseline produced a frame" and *fails* "…fills the
  viewport GREEN". A frame renders and its colour is wrong.

**The hypothesis I am testing, stated as a hypothesis:** this repository has a recorded trap of
**exactly five** phantom Vulkan failures from a build directory reused across a branch switch, all
passing after a fresh configure. My build directory was reused across a temporary one-line A/B gate
(built with it, then rebuilt without it), which is that reuse. The count coincidence is close enough
that I will not wave it away.

> **[EDIT — this CI argument is withdrawn; see the correction comment below.]** The `Linux` job runs
> on a runner with **no GPU** (Mesa lavapipe, fixed at 8 subgroup lanes), and the three failing
> assertions here are **NGG MBCNT logical lanes 31 / 32 / 63** — a 64-lane path lavapipe cannot
> exercise. CI's green on these tests is green on a **smaller domain**, so it cannot establish what I
> used it to establish. The conclusion is unchanged and now rests on cross-lane reproduction instead.

**The discriminator has answered, and I pulled the job log rather than trusting the green tick** — a
green check is also consistent with the tests never having run, which this project has recorded.
From the `Linux` job of run `33931720337`, at head `095833e4689e`:

```
334/358 Test #334: recompiled_shaders_render ......................   Passed    0.08 sec
337/358 Test #337: multidraw_render ...............................   Passed    0.18 sec
340/358 Test #340: indexed_render .................................   Passed    0.04 sec
341/358 Test #341: descriptor_array_render ........................   Passed    0.04 sec
351/358 Test #351: gpu_execute ....................................   Passed    0.06 sec

100% tests passed, 0 tests failed out of 358
```

and the job's own guard against exactly the empty-set reading:

```
Vulkan-gated suite present; total registered:   358
```

**All five run and pass in a clean tree at this exact head. These failures are not this change.**

**What that does NOT establish, and I am not claiming it.** CI is a different machine with a
different Vulkan stack, so "green there, red here" is consistent with two remaining causes: a stale
incremental build directory on this box (this repository records exactly five phantom Vulkan
failures from a directory reused across a branch switch), **or a real defect in this machine's
Mesa/RADV that CI's implementation does not hit.** The second would mean local GPU-test results on
this box need a health warning, which matters more than this PR. A fresh-configure arm on another
lane separates them; it is not a blocker here and I have not claimed either.

Recording the reasoning rather than only the outcome, because "my change cannot plausibly reach that
code" is not evidence and I did not want to merge on it: the three arms that separate the cases are
CI-green (**not the code change** — answered), a peer lane seeing the same five locally (**the machine
or the matrix**), and a peer lane seeing them green locally (**my build directory specifically**). The
first has landed; the second is being run by another lane and is not blocking here.

### The regression test

`avplayer_hle` (test 69) — **Passed, 6.07 s**.

Mutation arm, run before this head and restored: deleting the `avp_fire_data` call in
`s_avp_jumptotime` reddens exactly

```
[FAIL] a successful sceAvPlayerJumpToTime publishes exactly one warning-channel event
[FAIL] the published notification code is the seek-completion code the guest waits for
== FAIL: 2 ==     (exit 1)
```

and restoring it returns exit 0. The negative arm — *"a FAILED sceAvPlayerJumpToTime publishes no
seek-completion notification"* — passes in **both** directions, which is what stops an unconditional
fire from satisfying the positive arm.

### Arms

The A/B in the PR body used a **temporary, uncommitted** `PROSPER_LANE_BASELINE_ARM` gate around the
single `avp_fire_data` call, so the two arms differed by one call and nothing else — same compiler
invocation, same objects, same binary path. It is reverted; `grep -rn PROSPER_LANE_BASELINE_ARM
prosper/` returns nothing at this head, and the binary these results come from was rebuilt after the
revert.

### Guards

Head `095833e46` (the guard runs predate the review-fix commit `aca24f840`, which changes only a comment, `COMPATIBILITY.md` and `AGENTS.md` — no runtime behaviour). Every run on an otherwise idle box.

**`cobra-gameplay` — RED, and it is the subject of this PR.** From four failing counters on the base
to `structural matches 0 < 21` alone, with qualifying frames, pixel changes and non-black coverage
all crossing. Same shape on a contaminated run and on a clean idle-box re-run. The full-timeline
sweep (`when_matched.py`) is in the PR body; no baseline was touched.

**`messenger-scene` — RED, and it was already red without this change.** Two arms differing by
**exactly one call** — a temporary env gate around the `avp_fire_data` publish, so the binaries were
otherwise identical — with the second on an exclusively idle box:

| | with the change | change suppressed, idle box |
| --- | --- | --- |
| verdict | `structural matches 0 < 22; non-black coverage matches 0 < 22` | **identical** |
| state at 290-319 s | dialogue screen | **the same dialogue screen** |
| colours / coverage / best SSIM | 267 / 0.0481-0.0500 / −0.0069 | 266-268 / 0.0481-0.0500 / −0.0069 |
| run | 320/320, 163 distinct, stale 35.1 s | 320/320, 158 distinct, stale 35.1 s |

Filed as **#3346**. The title is healthy — 320/320 samples, 158 distinct frames, `guest=running` — and
the guard's window is looking at a dialogue screen, which is mostly black by design.

**`blue-prince-hall` — RED, and NOT yet attributed.** I am reporting this without a conclusion
because I have not run its baseline arm.

```
FAIL — qualifying frames 0 < 6; consecutive content pixel changes 0 < 5;
       structural matches 0 < 6; non-black coverage matches 0 < 6; consecutive content matches 0 < 6
157/157 screenshots; source-distinct=157 pixel-distinct=140 max-pixel-stale=35.0s guest=running
4.3 fps while producing frames, 95% of the 784.9 s run active
```

The title is alive and rendering — 140 distinct frames, 95% of the run active. **All 96 samples in
the 300-780 s window are the same loading screen**: a black frame with the game's cyan "Loading"
mark in the lower right, `distinct_rgb_colors` 1061-1215, `nonblack_ratio` 0.0255 flat, best SSIM
≈0.011. So it is on a loading screen for the entire eight-minute evidence window.

What I can say and what I cannot:
- the boot log contains **zero `[avp]` lines** — which rules out a *failed* seek and a media-clock
  play-out on that route, and says nothing about a successful one, since every other AvPlayer message
  is behind `PROSPER_AVPLOG`;
- 4.3 fps is **faster** than the ~3.2 fps this title is recorded at in the charter, so it is not
  running slower than its documented state;
- this guard was **void rather than green** in the last full matrix run (truncated against a 400 s
  limit when it needs 810 s), so there is no recent verdict to compare against.

**Not attributed to this change and not excluded from it.** The arm that settles it is the same
one-call-difference build run against this guard, which is ~25 minutes of exclusive box; I have not
spent it unilaterally on a contended machine. Flagging rather than concluding.

**`gris-gameplay` — deliberately NOT run.** `tools/snapshot/AGENTS.md` records a project-owner
decision (2026-09-02, #3148) that it is not a regression signal and *GRIS* is verified by hand
instead. Stating that so its absence is not read as an omission.

mattias800:
### Correction to my own verification comment: both unattributed titles DO import this API

I wrote "zero `[avp]` lines in the boot log" for `blue-prince-hall` and `messenger-scene`. A reader
will take that as "the title does not use AvPlayer". **It is not.** Verified binary-safe (Python
`bytes.count`, not `grep -r`, which mishandles these files):

| dump | `XC9wM+xULz8` JumpToTime | `9y5v+fGN4Wk` Pause | `w5moABNwnRY` Resume | `UbQoYawOsfY` IsActive | `libSceAvPlayer` |
| `PPSA25009` Blue Prince | 1 | 1 | 1 | 1 | 3 |
| `PPSA24651` The Messenger | 1 | 1 | 1 | 1 | 3 |
| `PPSA17337` Cobra (control) | 1 | 1 | 1 | 1 | 3 |

So this change is **potentially in circuit** on both of those routes, and the zero-`[avp]` bound
means only what I said it meant: no *failed* seek and no media-clock play-out were logged, because
every other AvPlayer message is behind `PROSPER_AVPLOG`, which the guards do not set.

That does not change either verdict. `messenger-scene` is still settled by measurement rather than by
inference — a one-call-difference arm on an idle box came back **red the same way**, which no import
fact can touch. `blue-prince-hall` remains **unattributed**, and the import fact makes the open
question precise rather than vague.

**And on `blue-prince-hall` specifically, there is a stronger reason not to spend an arm than cost:**
that guard was **void, not green**, in the last full matrix run — truncated against a 400 s limit
when it needs 810 s. There is no verdict to regress from, so a one-call-difference arm would measure
whether this change moves a guard nobody can show has ever passed at full length. Both outcomes leave
this PR where it is.

One method note, because it nearly went the other way: the first attempt to establish these imports
used `grep -r` and returned a clean **zero importer** for Blue Prince *and for Cobra* — a title this
PR proves imports it. Only the control caught it. Any "this title does not import X" claim about a
PS5 eboot should be made with a binary-safe reader and a positive control.

mattias800:
## Both blocking findings addressed — requesting re-review at head `78b10114b`

**Finding 1 — the body cited run records that did not exist.** Fixed by the two comments above: the
author verification carries `cobra-gameplay`, `messenger-scene` (with its one-call-difference
baseline arm) and `blue-prince-hall`'s actual records, plus a correction establishing that both
unattributed titles **do** import this API, so nobody reads "zero `[avp]` lines" as "does not use
AvPlayer".

**Finding 2 — a visible-progression claim with no capture and no `BLOG.md` entry.** Fixed twice over:

1. `COMPATIBILITY.md` stopped asserting the picture the moment you raised it (`aca24f840`), replaced
   by the measurement — *"the route that produced one uniformly black frame now produces 26 of 26
   pixel-distinct composited frames"*, still red on structure alone, HUD and sky seam named. That
   landed **before** the capture, so the PR was not left in an overclaiming state while under review.
2. The capture and its `BLOG.md` entry are now here (`07cfffc2b`).

### The capture

`tools/screenshot`, `PPSA17337`, `scripts/cobra/reach-title-or-gameplay.pad`, Linux/RADV,
**`PROSPER_RENDER_SCALE=1`** with no snapshot acceleration, fresh save dirs, idle box:

```
45/45 screenshots; source-distinct=45 pixel-distinct=45
max-pixel-stale=0.0s guest=running status=ok
34.8 fps while producing frames, 73% of the 180.0 s run active
```

**Every one of the 45 samples reports `capture_source = composited`** — prosper's own output, not a
guest-scanout republish — at 1920x1080.

Two images are committed. **The second is deliberately the degraded one.** The desert level is
recognisable — walkway, dunes, rock outcrops, character silhouette — and it draws with large
translucent blue polygons smeared across the middle of the frame and a dark band along the horizon.
Those defects are **more visible at scale 1 than the scale-4 A/B suggested**, which is exactly why
the scale-1 capture was required, and I have said so in the blog entry rather than picking a
flattering frame. `cobra-gameplay` remains red and **no rung is claimed**.

### Also adopted from your review

Your justification for synchronous publication replaced mine in the code comment, because it is
better. Mine rested on "the observed guests arm their slot before the call" — true, but two titles is
not a contract and it protected nothing if a guest armed afterwards. Yours identifies the mechanism:
the join is a **counting semaphore, not an edge** (`lock xadd` at `eboot+0x153f041`; the waiter sleeps
only when the pre-decrement value was `<= 0`), so an early post is *stored*. That holds for a guest
that arms **after** the call returns, which a deferred worker could not promise. Credited in the
comment.

### Verification state

- CI at the previous head: **9 pass, 0 fail**, and I pulled the `Linux` job log rather than trusting
  the tick — all five of my locally-failing Vulkan tests **pass in a clean tree**, with
  `Vulkan-gated suite present; total registered: 358` proving the green is not about an empty set.
- `git merge-tree --write-tree` against `origin/main`: **rc=0**, and a content diff of
  `COMPATIBILITY.md` between `origin/main` and the merge result shows **exactly one changed line,
  mine** — another lane's row survives byte-identical (the trap-41 check, by content rather than by
  status).
- Session-trailer gate **0**, `diff --check` clean, full `.md` gate green, **`snapshots.json`
  untouched**.

CI will re-run at `78b10114b`.

mattias800 (commented):
## REJECTED

Re-reviewed at `78b10114b`. **Both of my original findings are addressed**, and finding 2 twice over — I have no complaint left about either. What blocks now is **new information**, not a re-litigation: I went looking for whether my own counting-semaphore argument generalises, because you adopted it into the shipped comment, and the answer changed my view of `blue-prince-hall`.

Short version: the safety comment now states a **one-binary** observation as an unconditional contract, and the title whose guard is red and unattributed has a **live handler for the exact code this PR publishes**. Both are cheap to fix and neither requires the 25-minute arm you declined.

---

## What is fixed

**Finding 1 (records cited but absent).** Fully addressed. `cobra-gameplay`, `messenger-scene` with its one-call-difference baseline arm, and `blue-prince-hall` all now carry their actual verdict text. The correction establishing that both unattributed titles **do** import the API, with a binary-safe reader and a positive control, is better than what I asked for — and the method note about `grep -r` returning a clean zero importer for Cobra is exactly the right thing to write down. I hit the same class on my own first pass: `grep -qaP '\xa3\x00\x6a\x80'` reports 0 of 42 because of the NUL in the pattern.

**Finding 2 (progression claim with no picture).** Fully addressed. `COMPATIBILITY.md` states the measurement rather than the picture and names the still-red guard, the HUD and the seam; fixing it *before* the capture existed was the right order. I opened both committed frames. `cobra-desert-level.webp` shows the walkway, dunes, rock outcrops and character silhouette the caption describes, with the translucent blue polygons and the horizon band plainly visible — the caption is accurate and it is the honest frame. `cobra-main-menu.webp` shows the menu mid-animate-in as described. The `BLOG.md` entry leads with both pictures, runs about 120 words, says what you are looking at rather than what the change was, and ends on the defect. That is the format.

---

## Blocking

### 1. The adopted safety comment generalises a **one-binary** observation into an unconditional contract — measured: the mechanism is present in **11 of 42** importers, not all of them

`hle_service.cpp:2451-2462` now says:

> ...but that is two titles observed, not a contract, so it is not what this rests on.
> What makes an early notification safe is that the join is a COUNTING semaphore, not an edge. The guest's callback posts with `lock xadd` at `eboot+0x153f041`... Publishing here therefore **cannot lose the notification even for a guest that arms its slot AFTER the call returns**.

The comment rejects "two titles observed, not a contract" for the arm-before claim and then makes the identical move for the semaphore claim — from **one** binary's address. `eboot+0x153f041` is `PPSA17337`'s. Whether the join is a counter or an edge is a property of **the guest's** wait implementation, so it cannot be established for 40 titles by reading one. This is the charter's asymmetry landing exactly where it says it lands: it is *my* argument, it *raises* confidence, and it went in without measurement. I should have bounded it when I supplied it; I am bounding it now.

**Measured, so this is a number rather than an objection.** Over the 42 importers, byte-exact search for `PPSA17337`'s two encodings — the arm `c7 83 90 06 00 00 a3 00 6a 80` (`movl $0x806a00a3,0x690(%rbx)`) and the join `f0 0f c1 87 a0 06 00 00` (`lock xadd %eax,0x6a0(%rdi)`):

- **11 of 42** carry the arm encoding. **The same 11** carry the join encoding — perfect correlation, which is what identifies it as one shared compiled routine (Unity's PS5 video plugin) rather than a per-title coincidence: `PPSA17337`, `PPSA17952`, `PPSA19991`, `PPSA23396`, `PPSA24203`, `PPSA24651`, `PPSA27616`, `PPSA27624`, `PPSA27640`, `PPSA30140`, `PPSA30490`.
- I verified it structurally in a second one rather than trusting the byte match. `PPSA24651` (The Messenger): `14f66c3: movl $0x806a00a3,0x690(%rbx)` immediately before its JumpToTime call at `14f66d7`; callback join at `14f55c6: mov (%rcx),%esi ; mov 0x690(%rdi),%eax ; cmp %esi,%eax ; jne` then `14f55e5: lock xadd %eax,0x6a0(%rdi) ; test %eax,%eax ; js <wake>`. Identical routine, different addresses.
- **31 of 42 do not have it**, including `PPSA25009` — one of your own guard titles. `PPSA25872` has *a* `movl $0x806a00a3,disp32(reg)` store but not this one and no matching join, i.e. a variant.

**What to change.** State what was measured — *"the join is a counting semaphore in Unity's shared video plugin, byte-identical in 11 of the 42 local importers and verified structurally in `PPSA17337` and `PPSA24651`"* — and then rest the general safety claim on the argument that actually covers all 42, which is simpler and which nobody has to measure: **if a guest's join is an edge rather than a counter, an early post is lost and the guest waits exactly the timeout it already pays today.** Publishing early is therefore never worse than the status quo ante for any guest; the semaphore is what makes it *sufficient* for the Unity family, not what makes it *safe* in general. That version is both true and general, and it does not need a title census to defend.

### 2. `blue-prince-hall`'s red is closed on grounds that do not address mechanism — and there is a mechanism, in that title's own bytes

Your grounds for not spending the arm were (a) all 96 window samples are one loading screen and (b) the guard was **void, not green**, in the last full matrix run, so there is no verdict to regress from. I endorsed (b) as reasoning when I read it. It does not survive contact with the binary, because both grounds are about **comparability**, and neither is about whether this change can reach the title.

It can, and here is the path, read from `PPSA25009`'s eboot:

- **Its AvPlayer event callback is `eboot+0x13e46a0`**, and that identification is by ABI rather than by name. Its address is taken exactly once (`xref.py`: one address-taken reference, zero relocations) at `13e1dd4: lea 0x28c5(%rip),%rax`, and written to `-0x200(%rbp)` inside a stack-built struct whose base is `-0x260(%rbp)` — i.e. **base+0x60** — with `movl $0x0` at base+0x70 and `movb $0x0` at base+0x74, and the struct is then passed to the import stub at `0x1864e40`. Against prosper's own `static_assert`ed layout (`hle_service.cpp:1293-1298`), `AvPlayerInitDataEx` puts `event.event_callback` at **0x60**, `debug_level` at 0x70 and `auto_start` at **0x74**. Three fields, three matches.
- **Inside it, at `eboot+0x13e4a53`, is a live three-way dispatch on the warning payload:**

```
13e4a53  mov  (%r14),%eax
13e4a56  cmp  $0x806a00a3,%eax ; je 0x13e4b5e     <-- the code THIS PR now publishes
13e4a61  cmp  $0x806a00a1,%eax ; je 0x13e4b3f
13e4a6c  cmp  $0x806a0004,%eax ; jne 0x13e4f6d
13e4a77  lea  0x684996(%rip),%rax ; mov %rax,0x1c8(%r13) ; movb $0x1,0x1f8(%r13)
```

- **The `0x806a00a3` arm is not a no-op.** `13e4b5e` walks a table — `mov 0x18(%r13),%rax` as a count, `mov 0x8(%r13),%rcx` as a base, `shl $0x5` stride, dereferences `+0x10`, tests the word at `+0x14`, then reads `+0x100` and `+0xc0` off each entry and does work. That is a per-stream fan-out, not a log line.
- **Before this PR that arm was unreachable in prosper.** `avp_fire` only ever published the state events `0x01`-`0x05`; `0x20` was never published at all. This PR makes it reachable for the first time, on any route where the guest seeks.

So this is exactly the exposure your own cross-title section names — *"a guest ... whose dispatcher has a `default:` arm with side effects"* — sitting in a rung-6 guarded title whose guard came back red at this head. I am **not** claiming this PR broke Blue Prince: the window is a loading screen throughout, and nothing here shows the title calls `sceAvPlayerJumpToTime` on that route at all. What I am saying is that "not attributed" is no longer a safe place to merge from once a mechanism has a name, and that (b) in particular is the wrong shape of argument — *"we cannot show it ever passed"* is not *"it did not just break"*.

**And the arm you declined is not the cheapest discriminator.** One `PROSPER_AVPLOG=1` run on the `blue-prince-hall` route settles the mechanism question outright, with **no rebuild, no A/B and no second binary** — the same wall clock as the run you already did. `s_avp_jumptotime` logs `[avp] jump-to-time ... -> ok` under `avp_log()`, and `avp_fire_data` logs `[avp] -> event 0x20`. If neither appears, the mechanism is **excluded** and the red belongs to somebody else — which is a far stronger close than "unattributed", and it is the close you actually want. If they do appear, the one-call-difference arm becomes mandatory and you will know it was worth the 25 minutes.

---

## The two questions you asked

### Should a failed seek publish `0x806a0004`?

**No, not in this PR — but your stated reason is the wrong one and should change.** The PR says publishing on failure *"would be the same class of lie as the silent `SCE_OK`"*. It would not: `0x806a0004` is not a completion announcement, it is the failure announcement the same callbacks recognise. In `PPSA17337` it is special-cased at `153ef5e` and reaches the **same** `lock xadd` post at `153f041` by the `al=1` path, bypassing the equality check — i.e. it is a wake-regardless-of-what-you-awaited signal. Publishing it would be truthful.

The reason not to do it here is different and stronger, and finding 2 supplies it. `PPSA25009`'s dispatcher treats `0x806a0004` as a **state change**: it stores a pointer into `r13+0x1c8` and sets a flag byte at `r13+0x1f8`. Meanwhile prosper's failure set at `hle_service.cpp:2410-2419` includes `"the source has no host decoder to reposition"` — which is the #1105 graceful-skip and any headless or synthetic session, i.e. configurations where **nothing is wrong with the media and prosper simply has no backend**. Publishing a hard AvPlayer error into those would push a real error state into guests that are working as designed. So: keep publishing nothing on failure, and say *that* — no title has yet been observed to wait through a failed seek, and prosper's failure set is not congruent with the guest's notion of one. Worth a line in **Follow-on** with the `0x806a0004` mechanism recorded, so the next person does not have to re-find it.

### Does the counting-semaphore argument generalise?

Partly, and now with a number: **11 of 42**, byte-identical, verified structurally in two. See blocking finding 1 for the measurement and for the general argument that does cover the rest. The short answer to the question as you posed it is that **a two-title observation was promoted to a contract** — and I am the one who handed it over without a bound, so this is a correction to my own review as much as to your comment.

---

## Non-blocking

- Everything I raised last time that you did not act on is still fine as-is: `source_id` hardcoded to `0` (correct here — Cobra's callback never reads `edx`, and `PPSA25009`'s dispatch is on the payload too), the `[avp] -> event 0x%02x` line printing only the id on a channel whose meaning is the payload, `audio_delivered` being process-global while its line prints a per-handle id, and `when_matched.py:127` not carrying `snapshot.py`'s `0.85` default for a missing `min_structural_similarity`.
- The five local Vulkan failures: your reasoning is the right shape and I am not asking for more. Worth noting the discriminator you named is already half-answered by finding 1's data — none of the five touches AvPlayer, and this diff's only `src/` file is `hle_service.cpp`.
- `messenger-scene` deserves one addition now that the import table is known: `PPSA24651` is **in the 11** — it carries the exact Unity arm-and-wait routine, verified above at `14f66c3` / `14f55e5`. So the change is not merely "potentially in circuit" there, it is in circuit for the same code path as Cobra's. That makes your one-call-difference arm on an idle box the *right* evidence rather than a formality, and it is worth saying so in #3346 so nobody re-opens the question.

mattias800:
## Correction: my CI argument for the five `ctest` failures is withdrawn. The conclusion stands on different evidence.

In the author-verification comment I quoted the `Linux` job's log showing all five of my locally
failing tests `Passed` at the same head, and said that established the failures were not this change.
**That argument does not hold**, and I have annotated the original comment rather than quietly editing
it.

### Why CI cannot speak to those five

`.github/workflows/ci.yml:97` — *"The runner has no GPU, so this is Mesa's lavapipe software
rasteriser."* And the same comment block records, of `recompiled_shaders_render`, that assertions were
gated because they *"require an exact 64-lane fragment subgroup. **lavapipe is fixed at 8**"*.

My three failures in that test are **NGG MBCNT logical lanes 31 / 32 / 63**, and the assertion
computes `(lane * 255 + 32) / 64` — a hard-coded 64-lane expectation
(`tests/gpu/recompiler/test_recompiled_shaders.cpp:345`). That is a domain an 8-lane software
rasteriser structurally cannot exercise.

So CI's green there is green on a **smaller domain**. I quoted a passing control that could not have
failed — the charter's own rule about a positive control drawn from the same source testing the
discriminator rather than the domain, applied to a whole CI stack. It looked like the strongest
evidence in the comment and it was the weakest.

### What the conclusion rests on now

**Cross-lane reproduction.** Another lane ran both a reused and a fresh build directory on a
demonstrably quiet box and got **the same five, set-equal to mine** — on a **different head, a
different branch and a different dump**, with a path-filtered diff showing that work touches nothing
under `src/gpu/` or `frontends/`.

Three independent variables changed and the five failures did not. That establishes "not this change"
in a way CI never did, and it also retires my own earlier hypothesis that a stale incremental build
directory was responsible — a fresh configure reproduces them.

### And the finding underneath is bigger than this PR

Either `main` is red on indexed draws on real hardware and CI cannot see it, or every local
"ctest green" on this box for GPU-execution tests has been false. **Four of the five failures are
indexed-draw or GREEN-quad assertions** — draws that render nothing — which reads as one coherent
defect shape rather than five unrelated ones. That is being filed separately with a bisect queued; it
is not this PR's, and this PR should not be read as evidence either way about it.

**Unchanged for this PR:** 362 discovered, 357 passed, 5 failed, 3 skipped, exit 8, reported as a
failure rather than as green; `avplayer_hle` passed; the mutation arm reddens the two named
assertions.

mattias800:
## `blue-prince-hall` closed on mechanism: the title never touches AvPlayer on that route

The review's second finding was right that "not attributed" implicitly assumed this change could not
**reach** `PPSA25009`, and that the assumption was unfounded — the title imports the API and its event
callback at `eboot+0x13e46a0` has a **live** `0x806a00a3` arm at `+0x13e4b5e` that walks a per-stream
table and does work. That is a real mechanism and it deserved a measurement rather than an argument.

**Measured.** One `PROSPER_AVPLOG=1` run on the guard's own route, scale 1, idle box:

```
jump-to-time calls : 0
warning 0x20 events: 0
[avp] lines, ANY   : 0

809.4 s of an 810 s route · 97% active · 99.3 fps · guest=running
40 samples, all source=composited, 1920x1080
```

**Zero `[avp]` lines of any kind** — the title never initialises a player on this route, let alone
seeks. So the handler at `+0x13e4b5e` is **unreachable here**, and this change cannot affect that
guard. That closes the last open item in this PR's cross-title risk section, on mechanism rather than
on absence of evidence.

### Caveats, including two defects in my own run

**`status=FAILED` is my arithmetic, not the title's.** `--seconds 20 --count 42` is 840 s against a
`--timeout 810`, so it always stopped at 40 samples with `stop=timeout`. The route is 810 s and the
run covered 809.4 of it.

**My route guard misfired and the right thing happened anyway.** The log carries
`[pad] PROSPER_PAD_SCRIPT: cannot open route file: …/None` — I read `pad_script` from
`snapshots.json` and Python printed the literal `None` for the absent key, so my `${PAD:+…}` guard saw
a non-empty value and passed a bogus path. prosper refused it and delivered **no input**, which is
exactly what this no-input guard specifies. Correct route, wrong reason.

**The positive control is an inference, not an observation.** I did not dump the live process's
`environ` to prove `PROSPER_AVPLOG` reached it, and the process has exited. What supports the zero:
the same binary with the same flag emitted **976 `[avp]` lines** on `PPSA17337`, so the channel works;
and the flag sits in the same `env` block as `PROSPER_RENDER_SCALE`, whose effect is visible in the
run (1920x1080 at scale 1). That is strong, and it is one step short of a same-instrument control. The
cheap hardening is to re-run this exact script against `PPSA17337` and confirm `[avp]` lines appear.

CONFIDENCE: HIGH that no AvPlayer traffic occurred on this route. CONFIDENCE: MED that the instrument
would have shown it — see the third caveat, which is honest about which control I have.

mattias800:
## Positive control run: the Blue Prince zero is a real zero

Following the result above, the zero was carrying a blocking finding, so it needed a control on the
**same instrument** rather than an inference from a sibling variable. Here it is.

### The control is provably the same instrument

`avplog-control.sh` is the *unmodified* `avplog-run.sh` with three substitutions — output tag, guard
name, dump — checked mechanically rather than by eye:

```
diff <(grep -E '^\s+PROSPER_|^env PROSPER_' avplog-run.sh) \
     <(grep -E '^\s+PROSPER_|^env PROSPER_' avplog-control.sh)
  -> IDENTICAL, every PROSPER_* line, including PROSPER_AVPLOG=1
```

The only change inside the `env` invocation is the dump path, which is the subject. **The script was
deliberately not fixed first** (see the defect below): repairing it would have made this a test of a
*different* script from the one that produced the zero.

### Result

| | `[avp]` lines | `jump-to-time` | `-> event 0x20` | run |
| control `PPSA17337` | **966** | **3** | **3** | 10/10, `status=ok` |
| subject `PPSA25009` | **0** | **0** | **0** | 40 samples, 809.4 s of an 810 s route |

```
[avp] -> event 0x20
[avp] jump-to-time handle=0x3 target=0 ms -> ok
```

The instrument fires, loudly, through the same env block, on a subject known to seek. **So Blue
Prince's zero is a property of the title, not of the apparatus** — it never initialises an AvPlayer on
that route, so the live `0x806a00a3` handler at `eboot+0x13e4b5e` is unreachable there and this change
cannot affect `blue-prince-hall`. Caveat 3 from the previous comment is discharged.

### Caveat 1 restated beside the result, so it is not read as a crash

The subject run reports `status=FAILED  stop=timeout`. That is **my arithmetic**: I passed
`--seconds 20 --count 42` = 840 s against a `--timeout 810`, so it always stopped at 40 samples. The
route is 810 s and the run covered 809.4 s of it at 97% active, 99.3 fps, `guest=running`.

### The script defect, named for whoever reuses it

`${PAD:+…}` treats the literal string **`None`** as non-empty. Python's `.get(key, '')` returns `None`
when the key **exists with a null value** — which is how `blue-prince-hall` spells "no route" — and
`print(None)` emits `"None"`. So the guard passed `…/prosper/None` as a route path.

The trap is that the two cases are **externally indistinguishable**: against a guard *with* a
`pad_script` it passes a real path and works; against one *without*, it passes a bogus path that
prosper refuses, and refusing happens to be correct for a no-input guard. It produced the right
behaviour here **only by accident**. Discovery mode is the worst kind — the run succeeded, so nothing
pointed at it; it surfaced only from a `[pad] cannot open route file: …/None` line in a log being read
for another purpose.

**This is a Python-`None`-stringification hazard, not a one-off**, and it will recur anywhere a shell
script reads an optional JSON field through Python. The fix is `e.get(k) or ''` rather than a default
argument — `.get(k, '')` does not protect against a present-but-null key. Both that and the
count-versus-timeout arithmetic are now fixed in the script, after the control ran.

mattias800 (commented):
## APPROVED

Re-reviewed at `7a42b8e8d`. Both blocking findings are closed, and closed the way I asked — with measurements. I checked the measurements rather than their existence, including two things the coordinator flagged and one the author did not claim. CI: 9 success, 2 skipped, 0 failed, 0 pending.

One correction remains, and it is **to my own argument, not to yours** — see *The gap in the degradation bound*. It is a one-sentence scoping edit and I am asking for it before merge rather than blocking on it, because you rendered what I gave you faithfully and the flaw is in what I gave you.

---

## Finding 1 — closed, and the census reproduces

Your independent search over every ModRM base-register variant lands on the same **11 of 42** I measured with the two fixed encodings, names the same eleven, and puts `PPSA25009` outside them. Two different search methods agreeing on the same set is worth more than either alone.

The comment at `:2470-2487` is now correctly structured: each narrower argument is stated **with the population it holds for** — "two titles observed" for arm-before, "**11** … one byte-identical Unity plugin" for the counting semaphore — and neither is presented as the load-bearing one. Calling the general claim a **"DEGRADATION bound rather than a safety proof"** is a better piece of drafting than the sentence I handed you, and it is the phrase a future reader needs.

### The gap in the degradation bound (my error, please fold in one sentence)

`:2484` reads:

> So publishing synchronously is never worse than the status quo for any importer, and is strictly better for the ones whose join can store it. **That is the claim this code makes.**

That is my wording and it is scoped too widely. The bound reasons entirely about **the wait**: if the join is an edge, the post is lost and the guest waits the timeout it already pays. What it does not cover is that **the callback body executes regardless** — the notification is not "lost" at the function-call level, only at the wait. Two things fall outside it:

1. **Handler side effects.** For the 31 non-Unity importers nobody has read the handler, and for the one that *has* been read the arm is not inert: `PPSA25009`'s `0x806a00a3` case at `eboot+0x13e4b5e` walks a per-stream table and does work. An edge-triggered join loses the wake; it does not stop that loop from running.
2. **Re-entrancy.** The real player publishes from its own event thread; prosper publishes on the caller's thread inside the call. A handler written assuming it is not re-entered from `sceAvPlayerJumpToTime` would be *worse* off, not merely no better.

Both are already named in the PR body's cross-title risk section, and (2) has an independent answer that is stronger than the bound: prosper has **always** invoked guest event callbacks synchronously from HLE entry points — `:2360` fires `AVP_PAUSE` from `s_avp_pause` and `:2378` fires `AVP_PLAY` from `s_avp_resume`, so Cobra's own `pause -> jump -> resume` sequence already re-enters that way twice and does not deadlock. This publish joins an established pattern rather than opening a new hazard class.

**Suggested edit, one sentence:** say the bound is never worse than the status quo **for the guest's wait**, and that what remains outside it — a handler with side effects, and re-entrancy — is the residual the PR body names, with re-entrancy separately covered by the existing synchronous fires at `:2360` / `:2378`. Nothing else needs to change.

---

## Finding 2 — closed on mechanism, and the instrument survives the checks I could make

**The null has positive sites, which is the check that matters.** A zero from a log channel is only as good as the channel's coverage of the thing claimed absent, so I looked for whether an AvPlayer could exist on that route without emitting `[avp]`. It cannot: `s_avplayer_init` logs `[avp] init …` at `:1903-1909` and `s_avplayer_initex` logs `[avp] init-ex …` at `:1928-1934`, **both under `avp_log()`**. That matters specifically here, because `PPSA25009` builds an `InitDataEx` (its callback lands at struct base+0x60, the `Ex` layout), so the `Ex` path is the one it would take — and it logs too. So `[avp] lines, ANY: 0` really does exclude an initialised player of either flavour, not merely a seek. Your stronger sentence — *"the title never initialises a player on this route"* — is supported, and it is supported by the specific path this title would use.

**The control's env question, checked rather than reasoned about.** The coordinator asked whether the env block could take effect for one title and not the other. Two facts close it:

- `tools/screenshot` is a **C++ binary** (`tools/screenshot/screenshot.cpp`) with no `snapshots.json` reader, so nothing injects per-guard environment behind the script — the env is exactly what the `env` invocation carries, which is what you diffed.
- The guards' own `env` maps are not identical, and I checked what the difference is rather than assuming there was none: `blue-prince-hall` carries `{"PROSPER_IME_AUTOKEY": "1"}` and `cobra-gameplay` carries `{}`. That is a variable the subject had and the control did not — but it is an independent `getenv` with no relationship to `PROSPER_AVPLOG`, so it cannot gate the channel. Worth one line in the record that the two runs differed by that one variable, since "three substitutions" reads as "otherwise identical" and the guard-derived half is not covered by a diff of the script text.

**966 lines / 3 jump-to-time / 3 `-> event 0x20` on the control against 0 / 0 / 0 on the subject, on the same script and the same binary, is a same-instrument positive control.** Caveat 3 is properly discharged.

**`pad_script` verified independently.** `blue-prince-hall`'s entry has `pad_script` **present with a JSON null**, exactly the shape you describe, and that guard specifies no input — so "correct by accident" really is correct, and the route the zero came from is the guard's own route. I checked this from `snapshots.json` rather than taking it from the comment, because the whole reachability conclusion rests on the run being the guarded route.

**Not fixing the script before the control was right, and I would have objected if you had.** Repairing an instrument between the null and its control makes the control a statement about a different instrument — that is the trap the charter's instrument list exists for, and this is a clean instance of resisting it. One thing you did not claim that strengthens it further: the fix is **behaviour-neutral for both arms anyway** — `e.get(k) or ''` yields empty for `blue-prince-hall` (still no `--pad-script`) and is unchanged for `PPSA17337`, which has a real route — so the discipline cost nothing here. The principle would hold even if it had.

The `None`-stringification note is the most reusable thing in the comment: `.get(k, '')` not protecting against a present-but-null key, in a shape whose two failure modes are externally indistinguishable, is a hazard that will recur in any shell script reading an optional JSON field through Python. Worth an instrument-trap row rather than only a PR comment.

---

## The failure-path rationale — accurate against the bytes

`:2441-2456` is right on both halves, and I re-read the guest to check the half that cites it. `0x806a0004` really is the truthful failure signal rather than a lie — in `PPSA17337` it reaches the *same* `lock xadd` post at `+0x153f041` through the `al=1` path, bypassing the equality check. And the objection you now give is the correct one: `PPSA25009`'s arm at `+0x13e4a77` is `lea <rodata>,%rax ; mov %rax,0x1c8(%r13) ; movb $0x1,0x1f8(%r13)` — a stored pointer and a flag byte, i.e. a latched state, so publishing it because prosper had no decoder would permanently mark a player the title was merely skipping an unplayable movie on. Naming the narrowing as a real improvement that needs a title waiting on the failure arm to verify against, rather than doing it blind, is the right place to stop.

## Non-blocking

- The withdrawal of the CI-log argument is the right call and for the right reason — a lavapipe stack cannot exercise a 64-lane MBCNT expectation, so that green was green on a smaller domain. Retracting your own strongest-looking evidence because it could not have failed is the charter's positive-control rule applied to yourself, which is harder than applying it to someone else's data. The cross-lane reproduction on a different head, branch and dump is what "not this change" needed.
- Everything from my earlier reviews that you did not act on remains non-blocking and unchanged: `source_id` hardcoded to `0`, the `[avp] -> event 0x%02x` line printing only the id on a payload-carrying channel, `audio_delivered` being process-global, and `when_matched.py:127` not carrying `snapshot.py`'s `0.85` default.
- `PPSA24651` being one of the 11 is still worth a line in #3346, so nobody re-opens whether `messenger-scene` was exposed: it runs the same Unity arm-and-wait routine as Cobra, which is what makes your one-call-difference arm the right evidence there rather than a formality.

Approved at `7a42b8e8d`. Please fold in the one-sentence scoping edit above before merging; if you disagree with the scoping, say so rather than leaving the current wording.

mattias800:
## Precision on the control's scope, and the scoping edit is pushed

**Scoping edit: `9f4416c1d`.** The degradation bound now says explicitly that it is about the
**wait** — an edge-triggered join loses the wake, but that does not stop the callback **body** from
running, and `PPSA25009`'s `0x806a00a3` arm at `eboot+0x13e4b5e` walks a per-stream table doing real
work. Side effects in an edge-triggered guest's callback are named as the open residual, which is the
same one this PR body already records. Re-entrancy is stated as separately covered rather than left
looking overlooked: prosper has always fired guest callbacks synchronously from HLE entry points
(`:2360`, `:2378`), and `PPSA17337`'s own `pause → jump → resume` already re-enters twice per seek on
the pre-existing paths. Comment only, no behaviour change, `test_avplayer` exit 0.

## Correcting my own description of the control

I wrote that the control differed by *"three substitutions"*, which reads as **otherwise identical**.
That is true of the **script text** and not of the whole environment, and the difference is worth
stating precisely:

```
blue-prince-hall   env={'PROSPER_IME_AUTOKEY': '1'}   pad_script=None   (JSON null)
cobra-gameplay     env={}                             pad_script='scripts/cobra/reach-title-or-gameplay.pad'
```

So the guard-derived half of the environment differs by one variable that a diff of the script text
cannot see. **The control still holds** — `PROSPER_IME_AUTOKEY` is an independent `getenv` and cannot
gate `PROSPER_AVPLOG` — but the honest statement is *"the script text was diffed byte-identical across
every `PROSPER_*` line, and the guard-supplied env differs in one unrelated variable"*, not
*"identical"*.

Flagging it myself because the whole point of that control was that **"same instrument" is a claim
that has to be checkable**, and I made it checkable for the half I could diff while implying it for
the half I could not.

Also confirmed independently and worth recording: `pad_script` really is a **JSON null** on that
guard, which is exactly the shape that produced the `${PAD:+…}` bug — so *"correct by accident"* is
accurate, and the zero did come from the guarded route.

And the null has **positive log sites on the path the title would actually take**: both
`s_avplayer_init` and `s_avplayer_initex` emit `[avp]` under `avp_log()`, and `PPSA25009` builds an
`InitDataEx`. So `0 [avp] lines` genuinely excludes **an initialised player**, not merely a seek —
coverage my external control could not establish on its own.

mattias800 (commented):
## APPROVED

Re-reviewed at `67f5b2df2`. My `9f4416c1d` approval no longer carried by content, so I re-established it: the merge, the `COMPATIBILITY.md` rewrite, and the deletion the author flagged.

**On the flagged deletion: it is a supersede, and a stronger one than "supersede" suggests.** No blocking findings.

---

## The five lines from `aa937cdb76` (#2906)

I checked this from the deletion side rather than the presence side, so it is an independent pass over the same question.

**Blame confirms the attribution**, and one detail refines it: `git blame origin/main -- COMPATIBILITY.md` puts lines 215, 216, 218, 219, 220 on `aa937cdb76` (2026-08-22) and line **217** on `6d0cff148` — which is *"chore: make main the branch this repository builds and documents"*, the mechanical `master` → `main` sweep touching a line whose content is still #2906's. So "five lines from `aa937cdb76`" is the correct substantive count and the sixth is not a second author. The flag is accurate.

**Why it is a supersede and not a clobber**, by the tests that actually distinguish them:

1. **The deleted text is not stale, it is falsified.** It asserts *"Regressed on current master"* and explains it as *"with the seek implemented … the title then stops driving the player"*. This lane measured the opposite: the guest was **blocked**, not idle — 15,003 ms against a 15,000 ms budget read from the disassembly beforehand. Leaving it would publish a known-wrong cause **four lines under a row saying "fixed"**, in the file a reader checks first. The deletion is not the risky action here; keeping it would be.
2. **The reading is preserved where history is kept, with a marker.** `GRIS_SONIC_COBRA_BRINGUP.md:201` still carries *"Cobra then stops driving the player entirely"*, under the `:170` banner *"Everything below describes the state before that fix"* — and separately as a `## Ruled out` row stating *"The observation is real; the inference from it was void, and is withdrawn."* A superseded claim that survives **twice**, once as history and once as a named falsification, is better preserved than most.
3. **It is not moved, it is gone from the current-state file.** `COMPATIBILITY.md` on this branch has **0** occurrences of either the "Regressed on current master" claim or the "stops driving the player" mechanism, so the self-contradiction is resolved rather than relocated.
4. **The true half of #2906 survives with attribution.** This is the test a careless supersede fails: #2906's actual contribution was the **bisect**, which is independent of the mechanism it hypothesised. `ea299e97` is still named at `GRIS_SONIC_COBRA_BRINGUP.md:16` (*"The black-frame regression bisected to `ea299e97` (#1974) is fixed"*) and again at `:180`. The falsified explanation was dropped; the correct finding was kept and credited.
5. **It was flagged, which is the whole of what trap 41 exists to catch.** Trap 41 is about a whole-file take that reverts another lane *silently*, with no conflict and nothing to look at. This deletion arrived with a `NOTE FOR REVIEW` naming the commit and saying a trap-41 sweep should stop on it. The failure mode is absent by construction.

**I agree with your judgment, and it is one revert away if the project owner prefers otherwise** — but I would argue against reverting it: the alternative is a current-state file stating a cause we can show is wrong.

## Independent trap-41 sweep on the merge

Rather than re-run the author's presence-side check, I took the other direction — every line `origin/main` has that this branch does not:

```
git diff origin/main <head> | grep '^-' | grep -v '^---'   ->  17 lines, all accounted for
```

1 `COMPATIBILITY.md` table row (replaced by this branch's own), the 5 flagged prose lines, 1 `GRIS_SONIC_COBRA_BRINGUP.md` table row (replaced), and 10 in `hle_service.cpp` / `test_avplayer.cpp` that are this PR's own replacements — the `avp_fire` signature, the dead `#0` literal, and the two test event callbacks. **No unexplained deletion anywhere**, and nothing psml-related is touched, which independently corroborates that #3347's merged lines came through intact. The net diff is 10 files, +498 / −17.

## The caption

**It does not over-claim, and it does not under-claim.** I had already opened this image in an earlier round, so I can check the caption against what is actually in the frame rather than against the commit message: the lit glass walkway, golden dunes, dark rock outcrops and character silhouette are there, and so are the large translucent blue polygons across the middle and the dark band along the horizon. *"The level is recognisable and it draws wrong"* is exactly right, and *"not compression artefacts"* is correct — the polygons have hard geometric edges, they are rendered content.

The provenance line is complete in the way this project asks for: frontend, unmodified, native resolution, `PROSPER_RENDER_SCALE=1` with no snapshot acceleration named explicitly, the route file, and `capture_source=composited` on every sample. And *"deliberately the degraded capture, not progression evidence for a rung"* forecloses the reading that matters. Publishing the ugly frame in the user-facing file, rather than the menu, is the right choice.

On under-claiming: the missing HUD is **not** dropped — it is disclosed in the summary table row of the same file (*"the scene draws without its HUD and with a seam in the sky"*). So all three defects are named in `COMPATIBILITY.md`.

### One non-blocking nit, and it is only about naming

The same file now describes the defect set two different ways: the table row says *"without its HUD and with a seam in the sky"*, the caption says *"translucent blue polygons smear across the middle … and a dark band runs along the horizon"*. A reader cannot tell whether "seam in the sky" and "dark band along the horizon" are one defect or two, or where the polygon smear fits among them — and `GRIS_SONIC_COBRA_BRINGUP.md` uses the table row's wording. Name the same three defects the same way in all three places; it costs a phrase and it stops someone counting four.

## Everything from my earlier reviews still stands

The comment at `hle_service.cpp` now scopes the degradation bound to the wait, which was the one edit I asked for at `9f4416c1d`, and the narrower arguments keep the populations they hold for (two titles; 11 of 42). The guest-side facts I verified from `PPSA17337`'s bytes are unchanged by the merge — the armed wait slot at `+0x15400b8`, the payload-for-id substitution at `+0x153ef5c`, the counting-semaphore join at `+0x153f041`, and the 15,000 ms budget at `+0x153ce55` — as is the corpus count (42 importers, 40 carrying `0x806a00a3`, 33 carrying `0x806a0004`) and the `blue-prince-hall` closure on mechanism.

`Contribution shape`, `Docs`, `Linux App` and `macOS App` are green at this head with the rest still running; the merge agent should confirm the remainder as usual.

Approved at `67f5b2df2`.

Issue #3353, opened 2026-09-05
Three atexit-only diagnostics can never print on a bounded run — including both reports that exist to explain an empty capture


mattias800:
## `atexit`-registered diagnostics are silently absent on the two frontends most runs use

Found while reviewing a new per-mode census that registers an `atexit` dump. The new instrument
handles this correctly in its own comment; this issue is about the **existing** ones, which do not
say so anywhere a reader will find them.

### The mechanism

`_Exit` does not run `atexit` handlers. prosper calls it from at least four places:

```
prosper/tools/boot_trace/boot_trace.cpp:131        _Exit(42);
prosper/frontends/prosper-app/main.cpp:2564        std::_Exit(exitCode);
prosper/src/hle/kernel/hle_kernel_time.cpp:731     _Exit((int)a0);
prosper/src/hle/kernel/hle_kernel_time.cpp:741     _Exit(0x66);
```

So an `atexit`-registered report prints on `tools/screenshot` and any frontend that returns from
`main`, and **does not print** on `boot_trace` or `prosper-app`. The last two entries matter
independently of frontend: they are on the **guest's own exit path**, so a run that ends because the
guest exited loses the report even on a frontend that would otherwise flush it.

### Why this is a trap and not just a limitation

**A missing line is indistinguishable from a zero.** An agent runs `boot_trace`, sees no census line,
and records "this title does none of that" — which is a statement about the harness wearing a
measurement's clothes. This is the same family as instrument trap 127 (*name the frontend*), with a
concrete mechanism attached: it is not that the frontends differ in what they exercise, it is that
one class of instrument cannot report on two of them at all.

The failure is silent in both directions available to the reader: no error, no warning, and an exit
code that reflects the run rather than the missing handler.

### Registrations affected

A grep of `prosper/src` for `atexit(` finds three:

```
prosper/src/gpu/timeline/gpu_timeline.cpp:2718   report_unfired_automatic_capture_gates
prosper/src/gpu/timeline/gpu_timeline.cpp:2723   report_unfired_timeline_capture_selector
prosper/src/host/memory/guest_write_watch.cpp:1423   guest_dmem_write_trace_report
```

That count is from `--include=*.cpp --include=*.hpp` over `src/` only, so treat it as a floor rather
than a census — a lane working in this area put the number higher, and registrations made through a
wrapper or from `tools/`/`frontends/` would not be in it. **Re-derive it rather than quoting this
list.**

[...]

Issue #3345, opened 2026-09-04
CB fast-clear latches across colour targets: has_clear is cx.count() on the persistent register file, and there is no cx write-order provenance


mattias800:
## Summary

`extract_render_state` decides that a colour target carries a CB fast-clear like this
(`src/gpu/state/render_state.cpp`):

```cpp
rs.color0_has_clear   = st.cx.count(P::CB_COLOR0_CLEAR_WORD0) || st.cx.count(P::CB_COLOR0_CLEAR_WORD1);
rs.color0_clear_word0 = st.cx.count(P::CB_COLOR0_CLEAR_WORD0) ? rd(st.cx, P::CB_COLOR0_CLEAR_WORD0) : 0u;
```

`st.cx` is the **persistent** context-register file: `count()` answers *"has this register ever been
written in this process"*, not *"was it written for the target currently bound"*. Context registers
are sticky by design, so once **any** pass anywhere in the run programs a fast-clear:

- every later pass reports `has_clear = true`, whatever surface it renders into, and
- `resolve_pipeline_state` decodes **the last written word** into that pass's clear colour, and
- the live renderer passes that colour to `render_draws_rgba` as the render pass's `loadOp` clear
  (`live_renderer.cpp`, `clear_for`).

A pass into a target the guest never fast-cleared therefore starts from another surface's clear
value. Where that pass's draws do not cover the surface, the result is a **whole target flooded with
a colour the guest never asked for** — and where the target is a scanout buffer, that is the frame
that reaches the screen.

## Measured

*Little Nightmares III* (`PPSA05143`), 320 s default-route run, `PROSPER_CLEARLOG=1`:

**34 of 34 distinct `[clearlog]` lines report `has_clear=1`** — every colour target in the run,
across `CB_COLOR0_INFO.FORMAT` values 2, 6, 9, 10, 12 and 14. Nothing in that set is a target the
guest is known to have fast-cleared; it is simply every target the title binds.

The only reason most of them do not take a wrong colour is the **format** gate:
`decode_clear_color` decodes `COLOR_8_8_8_8` (10) and nothing else, so formats 2/6/9/12/14 fall
through to the opaque-black fallback. That is a coincidence of coverage, not a check — widening the
decoder to more formats, which is otherwise desirable, would widen this defect at the same time.

Both of that title's registered VideoOut scanout buffers are on the wrong side of that gate:

```text
[...]

Issue #3340, opened 2026-09-04
Little Nightmares III (PPSA05143): a GPUVM permission fault hard-recovers the GPU ~2 minutes in, so the title screen is no longer reached


mattias800:
## Summary

*Little Nightmares III* (`PPSA05143`, Unreal Engine 4) **no longer reaches its title screen on a
default launch.** Every run long enough to get there ends in a **RADV hard recovery caused by a
GPUVM permission fault**, after which no graphics submission succeeds again and the present layer
re-serves one retained frame for the rest of the process.

The status doc records rung 2 — a title screen at t≈110 s — on master `4d7a2ded` (2026-08-05), with
a checked-in capture (`assets/screenshots/little-nightmares-3-title-screen.png`). That is not what
current master does. On `c067aeef` the last content frame before the collapse, at t=140 s, is still
the **Wwise splash**.

This has been sitting inside #2014 as a caveat on a colour measurement rather than as a finding.
It is filed separately because it is a different defect with a different owner: #2014 is about what
colour a frame is, this is about the run ending.

## Evidence

Frontend `tools/screenshot`, default route, no pad input, isolated `PROSPER_SAVE0` /
`PROSPER_SAVEDATA_DIR` per instrument trap 234, Linux/RADV (Radeon 8060S), inside the `ps5ys`
distrobox.

| run | build | bound | outcome |
| --- | --- | --- | --- |
| 2026-09-05 01:11 | `c067aeef` | 200 s | content to t=140 s, then **`radv: GPUVM fault detected at address 0x80025e677000`**, hard recovery, wall at `fresh=5260` |
| 2026-08-29 09:55 | `f8e72188` | 320 s | content to t=110 s, then **`radv: GPUVM fault detected at address 0x800272eb7000`**, hard recovery, wall at `fresh=5672` |
| 2026-08-29 09:48 | `f8e72188` | 120 s | collapses at t=115.1 s; the run ended before RADV reported, so it carries the collapse but not the fault line |
| 2026-08-29 09:51 | `f8e72188` | 120 s | ran its full bound with content and **did not** collapse |

**The fourth run is not a survival result.** The collapse lands at t≈115 s in one run and t≈145 s in
another, so a 120 s bound straddles it; that arm answers nothing about whether this title survives.
Stated here rather than dropped so the row is not later counted as a negative.

The fault dump is identical in kind in both runs that produced one:

```text
[...]

PR #3363, opened 2026-09-05
Little Nightmares III: name the flooding draws, fix two instruments that were reporting nothing

mattias800:
## What this is

The Little Nightmares III (`PPSA05143`) flat-yellow investigation: two instrument bug fixes, two
diagnostics, one new tool flag, three instrument-trap rows, and the status doc rewritten around what
was measured. **No rendering behaviour is changed and no rung is claimed.**

## The defect, named exactly

The flooding population is **3,734 draws of `(es=0x300e900000, ps=0x3017370000)`** carrying
`CB_COLOR_CONTROL=0x00cc0000` (MODE=DISABLE) and `CB_TARGET_MASK=0x7`, painting flat colour into the
guest's own scanout at `0x9fc0000000`/`0x9fc2000000`.

`PROSPER_SKIP_DRAW_PROGRAM=0x3017370000` takes uniform-yellow retained frames from **63 of 64 to 0 of
65** while content survives (28 → 29 content samples) — the one pre-registered outcome that names the
program without ambiguity, with the selector's arming line and 18 `reason=skipped-by-selector` lines
in the log.

**The shader computes the value; it does not manufacture a constant.** 782 instructions, four
`OpImageSampleImplicitLod`, 34 `OpSelect`, **no branches**, one store of a computed
`OpCompositeConstruct` through a `PackHalf2x16` round trip with ±65504 clamps. R and G arrive ≥1.0 and
saturate in a UNORM8 target while B arrives ≤0 — so #2014's *original* framing was right at the shader
level. Why they arrive ≥1.0 is **not** answered here.

## Two instrument bugs fixed

- **`PROSPER_CLEARLOG` read its output fields before they were assigned**, so every line it has ever
  printed reported `word0=0x00000000`, in every title. **The evidence that retired the clear
  hypothesis on #2014 was therefore void.** Fixed with a mutation-checked test.
- **`unmodeled_cb_color_mode_count()` had no production caller** — declared, defined, unit-tested,
  unreachable from a run, which is why three lanes independently quoted power-of-two lower bounds.
  An exit summary now prints the exact per-mode totals. It registers on the *first* unmodeled draw so
  a title that decodes none prints nothing and stays usable as a control, and omits never-counted
  modes for the same reason.
  **Its reach is then limited by #3353** — `tools/screenshot` ends in `_exit()`, so `atexit` reaches
  essentially no real run here. The comment says so and points at `PROSPER_COLORSTATETRACE` as the
  survivable source.

## Deliberately NOT claimed

- **No behaviour change.** `PROSPER_LEGACY_CB_DISABLE_MASK` remains a diagnostic lever; #1724 landed on
  cross-title evidence and #2932 measured the same suppression taking `PPSA02058` from 3-of-24 content
  samples to 0-of-24. One title's idiom does not overturn that.
- **The GPUVM fault is not established as a regression** (#3340) — the doc's t≈110 s title screen and
  the t≈115 s collapse are five seconds apart, so a marginally slower boot misses it with no change to
  the fault.
- Why R and G arrive ≥1.0 — a fresh investigation, named and not started.

## Corrections against the author's own published claims

Recorded rather than quietly amended: a pre-registered prediction falsified on both stated falsifiers;
a `## Ruled out` row narrowed before anyone relied on it; published pass-coverage rates re-normalised
(96.98 / 1.18 / 96.95, not 18.08 / 0.54 / 17.84); and **a #1706 conclusion overturned** — grouping by
pixel shader when the phenomenon is keyed on the *pipeline*. Re-keyed, the raw register words differ
in **exactly bits [6:4]** with zero mixing across 5,054 draws, which is the first title-side evidence
that field is real.

## Traps 263, 264, 265

263 — a working instrument with a broken presentation layer (`pgrep` errors on stderr while the caller
prints a hardcoded "nothing found" on stdout). 264 — the CLEARLOG row, renumbered after a collision.
265 — an aggregate reported on one key when the phenomenon is keyed on two.

## Verification

`check_merge_result.py` against the merge result: clean merge, numbered table intact, **no baseline row
deleted**. `diff --check` clean. Session-trailer gate 0. New tests are mutation-checked with the named
mutation in each commit message.

Refs #2014, #1706, #3340, #3345, #3353


mattias800 (commented):
## REJECTED

One blocking finding, and it is a **correction this PR says it made that is not in the document**. Everything else I checked holds — including all four things I was asked to attack hardest, three of which I verified from the bytes rather than from the prose. No conclusion in the PR depends on the blocking finding; it is the published numbers and a note about them contradicting each other.

---

## Blocking

### The pass-coverage re-normalisation is claimed in the PR body and not applied to the doc

The PR body lists among its self-corrections:

> published pass-coverage rates re-normalised (**96.98 / 1.18 / 96.95**, not 18.08 / 0.54 / 17.84)

`LITTLE_NIGHTMARES_3_STATUS.md:118-123` still publishes the old triple:

| arm | full-coverage VO passes | of total VO passes | rate |
| --- | --- | --- | --- |
| control | 2,218 | 12,266 | **18.08%** |
| `PROSPER_LEGACY_CB_DISABLE_MASK=1` | 25 | 4,645 | **0.54%** |
| `PROSPER_CB_EFC_NO_COLOR=1` | 2,255 | 12,639 | 17.84% |

and `:125` then instructs the reader to *"Read the **rate**, not the count"* — pointing at the rate the PR body says is the wrong one. Meanwhile `:191-193` asserts:

> the pass-coverage rates quoted above are normalised over passes that **produced pixels**, and that denominator is stable across the arms (2,287 control, 2,111 lever, 2,326 EFC)

That is false about the table it annotates. `2,218 / 12,266 = 18.08%`; `2,218 / 2,287 = 96.98%`. The table divides by total VO passes, which is the denominator that varies 2.6x across the arms — exactly the run-length confound the sentence at `:125` exists to remove. Using the stable denominator the note names gives `2,218/2,287 = 96.98`, `25/2,111 = 1.18`, `2,255/2,326 = 96.95`, i.e. the PR body's own triple, and the same normalisation the newest section already uses at `:212` (96.98% vs 0.52%).

**It also propagates into a `## Ruled out` row.** `:350` states the EFC null as *"**17.84%** full-coverage VO passes against an **18.08%** control"* — the old pair, in the one place this project treats as its most durable record.

**Concrete failure scenario:** the next agent reads `:118-123`, takes 18.08% as this title's full-coverage rate, and compares it against another title measured over pixel-producing passes — a five-fold error in the wrong direction — or reads `:191-193`, believes the table is already produced-pixel-normalised, and reports 18.08% as such.

**No conclusion changes.** The EFC lever moves nothing on either normalisation (17.84 vs 18.08, or 96.95 vs 96.98) and `PROSPER_LEGACY_CB_DISABLE_MASK` moves the title hugely on either. This is a numbers-and-consistency defect, not a wrong finding. Restate the table with the pixel-producing denominator (keeping both counts visible), fix `:125` and `:350`, and the note at `:191-193` becomes true instead of contradicting the table above it.

---

## What I verified, so it is not re-derived

### 1. The `PROSPER_CLEARLOG` bug and the retroactive claim — both hold

On `origin/main`, `extract_render_state` assigns `rs.color0_clear_word0`/`_word1` **two lines below** the `if (clearlog)` block that prints them, and `render_state.hpp:81` initialises both to `0`. So every `[clearlog]` line ever emitted printed `word0=0x00000000 word1=0x00000000`, in every title. The dedup key is `clear_seen.emplace(rs.color0_base, rs.color0_format, rs.color0_clear_word0)` — the **same** unassigned field — so the third element was constant and the "33 distinct `(base, format, clear-word)` combinations" on #2014 were 33 distinct `(base, format)` pairs. **"Void, not negative" is the correct verdict**, and the run genuinely could not have produced another number.

The scoping is right too, which is the part that could have been overdone: `rs.color0_has_clear` is assigned *above* the block and was always correct, so the `has_clear=1` half of those lines stands — and the doc's new "34 of 34 lines report `has_clear=1`" correctly leans on exactly that half. Voiding two fields and not the third is the honest cut.

**The test would fail without the fix.** The mutation is "move the block back above the assignments": `occurrence_count(...) == 1` still passes, `word0=0x11223344` and the no-`0x00000000` arm both redden. Two details make it work that are easy to get wrong and are right here — the diagnostic reads `std::getenv` fresh rather than through the one-shot `PROSPER_ENV_ON` macro, so runtime `setenv` takes effect; and the earlier unarmed `extract_render_state(st)` cannot have poisoned the dedup set, because the `emplace` sits inside the armed branch. Splitting "the instrument fired" from "the instrument reported the right value" into separate arms is the right shape, and trap 264 says why.

### 2. The `[6:4]` conclusion — the re-keying is sound and the partition is real, not a top-N artefact

- The field: `pm4_registers.hpp:408-409` gives `MODE_SHIFT = 4`, `MODE_MASK = 0x7`, i.e. bits **[6:4]**. `0x00cc0000 ^ 0x00cc0010 = 0x10`, which lies inside that field and nowhere else. The claim is exact.
- The partition: this is the part I expected to be an artefact and it is not. The two rows sum to `3,734 + 1,320 = 5,054`, and the independently reported per-program table at `:98` gives that same pixel shader **`07`x5,054** effective-mask draws. So the two `(es, word)` groups **exhaust** the `effective=07` population — a third group carrying a different `cb-control` under the same pipeline would push the sum above 5,054. The closure comes from a different aggregation than the one that produced the rows, which is what makes it a check rather than a restatement. **Zero mixing is established, not merely displayed.**
- Scope worth stating in the doc: the claim covers those 5,054 draws. The remaining 119 under the same pixel shader (`00`x117, `0f`x2) are outside it and their words are not shown. That does not weaken anything as written; it is one clause.

Overturning your own published `--by-program` reading, and writing trap 265 around it rather than quietly amending, is the right handling.

### 3. The shader claim — "no branches" survives inspection, and the conclusion follows with one qualification

Read from the committed disassembly:

- 782 instructions; **4** `OpImageSampleImplicitLod`; **34** `OpSelect`. All three exact.
- **Zero** of `OpBranch`, `OpBranchConditional`, `OpSwitch`, `OpLoopMerge`, `OpSelectionMerge`, `OpKill`, `OpPhi`, `OpFunctionCall`, with exactly one `OpLabel` and one `OpReturn` — a **single basic block**. "No branches" is not an approximation; there is no control flow at all.
- **1** `OpStore`, and its operand is `%732 = OpCompositeConstruct %v4float` whose four components are bitcasts of `UnpackHalf2x16` extractions. `2` `PackHalf2x16` / `4` `UnpackHalf2x16`, with `%float_65504` and `%float_n65504` used at both clamp sites. The round trip and the clamps are exactly as described.

So **"the shader computes the value rather than manufacturing a constant" is supported.** One qualification worth a clause: the *"R and G arrive >=1.0 and B <=0"* half is not derived from the shader — it is a re-description of the observed `(255,255,0,0)` in a UNORM8 target under saturate semantics. It is correct, and it is not independent evidence; it should not later be cited as a second measurement supporting the first. Naming *"why they arrive >=1.0"* as a separate un-started investigation is the right call.

### 4. The `## Ruled out` rows and the numbered table

The DCC row is the best thing in the diff — it falsifies by construction, then **qualifies its own falsification the same day** by tracing the swizzled-consumer chain that reaches `(255,255,0,255)` from a reachable uniform cyan, and ends with the one-run test that would settle it. That is a row that will save someone a day.

`check_merge_result.py` on the merge result: clean merge, `265` rows unique and ascending, **none of the 262 rows in `origin/main` deleted**, 14 tables unbroken. Verified by running it, not by reading the claim — see the non-blocking note about how.

**The GPUVM hedging is right, not under- or over-done.** "The doc's t≈110 s title screen and the t≈115 s collapse are five seconds apart, so a marginally slower boot misses it with no change to the fault" is a falsifiable reason to withhold the regression label, and it names the arm (`4d7a2ded`) that would settle it.

---

## Non-blocking

### The committed tool cannot reproduce the finding that overturns #1706

`colorstate_report.py`'s `RE_HEAD` captures `es=` as group 1 and the raw `cb-control` hex as group 4, and `parse()` stores **neither** — the record dict keeps `ps`, `mode`, the masks and the targets. So the only grouping the shipped tool can do is by pixel shader, which is precisely the one-axis aggregate trap 265 is about, while the two-axis grouping that produced the corrected finding lives in an uncommitted one-liner. A reader who follows trap 265's own advice — *"re-key it on every identifier the record carries"* — cannot, with this tool. Two keys in `parse()` (`"es": m.group(1)`, `"cbc": m.group(4)`) and a `--by-pipeline` branch beside `--by-program`, in a file this PR already edits, would make the #1706 evidence reproducible from the repo. Not blocking because `grep '^\[color-state\] es=' log | sed 's/ shader-mask.*//' | sort | uniq -c` reproduces it — but that is the kind of thing this project's tool-building rule exists to stop being folklore.

### The exit summary: zero-suppression is fine, its stated rationale is inverted

Omitting `modeN=0` **within** a line is clearly right. But *registering on the first unmodeled draw so a zero title "prints nothing and stays usable as a control"* does not do what it says. An absent line is indistinguishable from a lost `atexit` — and the same comment documents **four** paths where the report is lost, including `tools/screenshot`'s `_exit`, which is the project's primary evidence frontend. Absence therefore carries no information at all, so it cannot preserve a control; the control's zero has to come from `PROSPER_COLORSTATETRACE`, which the comment already says four lines later. The design is harmless — I am not asking for a row of zeroes — but the justification should be *"an absent line means nothing either way, so do not spend a line on it"*, not *"this keeps the control usable"*. As written it is the same absence-is-not-evidence move the rest of this PR documents.

### The summary test cannot fail on the property it is written to protect

`CHECK(any == (counted != 0), ...)` is the arm that would catch removal of `if (!total) { out[0] = '\0'; return false; }`. By the time it runs, the cases at `test_render_state.cpp:876-902` have already incremented modes 7, 2, 4 and 5, so `counted != 0` always and only the `any == true` branch is ever exercised — the assertion is true and the mechanism it guards never runs. Cheapest fix: call `unmodeled_cb_color_mode_summary` once at the **top** of `main()`, before any case counts, and assert it returns false with an empty string. That is the arm that makes the zero-suppression contract real.

### `check_merge_result.py` is broken on its default base since the branch rename

`--base` defaults to `origin/master` (`check_merge_result.py:60`), and there is no such branch any more — `git ls-remote origin refs/heads/master` is empty and the default branch is `main`. Run with defaults it dies with `fatal: ambiguous argument 'origin/master'`. I re-ran it as `--base origin/main --head <this PR>` and it passes cleanly (result above), so your table really is fine — but the PR body should say which base was used, since the default cannot have produced that line. Worth its own fix: the same `origin/master` spelling is in `CLAUDE.md`'s docs-gate recipe, so the charter-mandated pre-merge check currently fails for everyone who follows it literally. That is bigger than this PR and belongs in its own issue.

### Smaller

- `[uniformlog]`'s new source fields fall through to `0`/`none` for any `PresentSourceChoice` outside the three named, which is the right failure direction; the `vkfmt=` note about 44 being `VK_FORMAT_B8G8R8A8_UNORM` and not a CB format value is the kind of comment that stops a re-derivation.
- The `--by-program` selftest asserts on whitespace-exact column alignment (`"ps=0x2            draws=1       writing-colour=1"`), so any change to the format widths reddens it for a non-reason. A field-wise assertion would survive formatting.

Fix the published rates and I will approve on the next head.

mattias800 (commented):
## Correction to my own review — one finding WITHDRAWN. Verdict unchanged: REJECTED, on the blocking finding only.

**Withdrawn in full:** the non-blocking item headed *"`check_merge_result.py` is broken on its default base since the branch rename"*. It is wrong. Nothing in this PR needs to change for it, and the PR body's verification line needs no qualification.

Nothing else in my review is affected. The blocking finding stands, and so do the other three non-blocking items.

### What is actually true

```
git show origin/main:prosper/tools/docs/check_merge_result.py     :111  default="origin/main"
git show <this PR head>:prosper/tools/docs/check_merge_result.py  :111  default="origin/main"
git show origin/main:CLAUDE.md | grep -c 'origin/master'           1    (line 314 only)
```

Both refs default to `origin/main`. I re-ran the current tool with **no `--base` at all** and it resolved the default and passed:

```
merging <this PR head> into origin/main (245daf9d)
... numbered column 1..265, unique and ascending, no unused numbers;
    none of the 262 rows in origin/main:...GAME_COMPAT_ORCHESTRATION.md has been deleted
clean merge, and the merge RESULT passes the gate against origin/main as of now.
exit 0
```

The single surviving `origin/master` in `CLAUDE.md` is line 314's record of *what the first sweep missed* — a deliberately preserved historical statement, not a stale recipe. The mandated pre-merge check works for anyone following the charter literally, and my sentence that it "currently fails for everyone" was false.

### How it happened, which is the part worth keeping

**I read and ran a stale copy of the tool.** The shared checkout at the repo root is on an unrelated branch predating the `master` → `main` sweep. In that tree the file is a shorter, older version whose `--base` really does default to `origin/master` — at line **60**, not 111.

Two things made this land harder than an ordinary misread:

1. **I did not just read it, I RAN it**, and it failed with a real `fatal: ambiguous argument 'origin/master'`. That execution result is what made me confident enough to publish. **An execution result from the wrong copy is indistinguishable from an execution result from the right one** — it has all the texture of a verified fact and none of the provenance. Running a tool feels like the strongest available check, which is precisely why it needs the same ref-pinning as reading one.
2. **The line number was the tell and I quoted it without noticing.** I cited `:60`; current `main` is `:111`. A cited line number that does not match the ref you claim to be describing is the recognised signature of this exact mistake, and I printed it in my own review.

**The habit that failed is more specific than "be careful".** In the same review I used `git show refs/remotes/pr/3363:<path>` correctly four times — for the doc, the tests, the tool's parser, the trap table — and dropped it exactly once, for the one file that was **not part of the PR's diff**. My ref-pinning was attached to *reviewing a diff*, not to *reading a file*. The moment the subject became a repo tool rather than a changed line, the habit silently did not apply. So the rule that would have caught it is: **pin the ref for every file you cite, including the ones you are only using as instruments**, and treat "I ran it" as needing provenance rather than as supplying it.

There is also a direction-of-belief lesson here that cuts against the usual one. This project records that a finding which *raises* a claim gets checked least. This was the opposite — a finding that a tool is broken, which lowers someone's claim and should attract scrutiny — and it skipped scrutiny anyway, because I had executed it. **Execution results get treated as self-validating in a way that reading does not**, and that exemption is not earned.

I am flagging it at this length because a review finding is the most contagious thing this project produces — it is checked by definition, and a `file:line` makes it look checked twice. Mine had a real line number in a real file; it was the wrong copy of the file.

### One thing I noticed while confirming this, which is a credit to the tool

Running the current script from a scratch directory without its sibling `check_numbered_table.py` produced:

> The checker itself failed (rc=2) without reporting any table problem, so this says NOTHING about the merge result. Fix the invocation above, then re-run.

An instrument that refuses to turn its own breakage into a pass. Given that half this PR is about instruments which do the opposite, that is worth noticing.

---

**Standing verdict, unchanged: REJECTED**, on the single blocking finding — the re-normalised rates (96.98 / 1.18 / 96.95) are claimed in the PR body and absent from the document, which still publishes 18.08 / 0.54 / 17.84 at `:118-123`, directs the reader to that rate at `:125`, asserts at `:191-193` that those rates are already normalised over pixel-producing passes, and carries the old pair into a `## Ruled out` row at `:350`. No conclusion moves; the published numbers do. Fix those and I will approve on the next head.

The three surviving non-blocking items are unchanged: the committed `colorstate_report.py` cannot re-key on `(es, raw word)` because `parse()` discards both; the exit summary's zero-suppression rationale is inverted; and `CHECK(any == (counted != 0))` cannot fail on the property it guards, because `:876-902` increments modes 7/2/4/5 before it runs.

mattias800 (commented):
## APPROVED

Re-reviewed at `3a6972853`. The blocking finding is fixed at all four sites, and fixed the right way — the superseded pair is **kept visible with the reason it was wrong** rather than swapped out, which is what stops the next reader re-deriving it. I said I would approve on the next head if those landed; they landed.

Verified rather than read: `2218/2287 = 96.98%`, `25/2111 = 1.18%`, `2255/2326 = 96.95%` — the table, the superseded-denominator note, the `:133` sentence and the `## Ruled out` row at `:364` now all carry the same pair, and `:220` was already correct. Diff since my review is docs-only, one file. `Docs` is green at this head.

The `## Ruled out` row is the site I cared most about, and it does more than swap numbers: it records that an earlier revision quoted `17.84%` against `18.08%` and that **the null is identical either way**. A row that carries its own correction is worth more than one that was always right.

### Two same-pass edits, neither blocking

- **`:127`'s "the direction and the ~80x separation are identical on either normalisation" is not right.** The rate ratio is `96.98/1.18 = 81.9x` on the corrected denominator and `18.08/0.54 = 33.6x` on the superseded one — a 2.4x difference, not identical. The denominator-independent quantity is the **count** ratio, `2,218/25 = 88.7x`, but the adjacent line tells the reader to use the rate, so quoting a count here would fight it. What is actually invariant is the *conclusion*: both normalisations separate the lever arm from the other two by well over an order of magnitude. Say that, or say "82x corrected, 34x on the superseded denominator". As written it is one more number that does not survive checking, in the paragraph explaining why a number did not survive checking.
- **`:133`'s "within 8%"** is true measured against the control (`-7.70%` / `+1.71%`) and not measured max-to-min (`2,326` vs `2,111` is `10.2%`). Both readings support the point — the denominator moved by single-digit-to-10% while total passes moved 2.6x — so one clause naming which spread is meant keeps it unquotable in the wrong form.

### Standing from my earlier reviews

- My non-blocking item about `check_merge_result.py` is **withdrawn in full** — see the correction above it. The tool defaults to `origin/main` and passes with no `--base`; I had read and run a stale copy from a checkout on a pre-rename branch. Nothing in this PR was ever affected by it.
- The three surviving non-blocking items are unchanged and none of them blocks merge: `colorstate_report.py`'s `parse()` discarding `es=` and the raw `cb-control` word, so the committed tool cannot perform the re-keying trap 265 recommends; the exit summary's zero-suppression rationale being inverted (an absent line is indistinguishable from a lost `atexit`, which the same comment documents four paths for); and `CHECK(any == (counted != 0))` being unable to fail on the property it guards, because `:876-902` increments modes 7/2/4/5 before it runs. The last is the one I would fix first — it is a test written to protect a contract it cannot test.

Everything I verified in the first review stands: the CLEARLOG read-before-assign bug and the "void, not negative" verdict on the #2014 evidence; the mutation genuinely reddening the value arms while leaving the fired-at-all arm green; the `[6:4]` partition closing arithmetically against an independent aggregation (`3,734 + 1,320 = 5,054`); and the shader's single basic block with zero control-flow instructions, four samples, 34 selects, and one store of a computed `OpCompositeConstruct` through the half-pack round trip with `±65504` clamps.

Approved at `3a6972853`. Please fold in the two clauses above; if you disagree with either, say so rather than leaving the wording.

mattias800 (commented):
## REJECTED

Re-reviewed at `0d87188fa`. The blocking fix from my last review landed correctly at all four sites and the `≥1.0` qualification is in. `--by-pipeline` is the right feature and it reproduces the overturn. But the new logic has **two defects, and I measured both rather than arguing them** — one of them prints a confident reassurance on a decode I constructed to be wrong.

Both fixes are one line.

---

## Blocking

### 1. The `--by-pipeline` verdict says "the decode is faithful" on a decode that is provably not

I built a positive instance of the class by hand, outside anything that produced the null: a log where the guest writes the two real words and the reported mode is `0` for **both** — i.e. every `NORMAL` draw mislabelled `DISABLE`, which is exactly the error #1706 exists to detect.

```
  3734  es=0x300e900000 ps=0x3017370000 cb-control=0x00cc0000 mode=0 (DISABLE) effective=07
  1320  es=0x3016e60000 ps=0x3017370000 cb-control=0x00cc0010 mode=0 (DISABLE) effective=07
  2 distinct raw CB_COLOR_CONTROL word(s); 0 decoding to more than one mode
  (each word decodes to exactly one mode, so the decode is faithful to what the guest wrote)
```

`0x00cc0010`'s MODE field is unambiguously `1`. The tool has the raw word **in the row it just printed**, reports `mode=0 (DISABLE)` beside it, and calls the decode faithful.

**The check tests whether word → mode is a *function*. It never tests whether it is the *right* function.** And on any real log it cannot even fail: at this head `render_state.cpp:823` is

```cpp
out.cb_color_mode = PM4_FIELD(rs.cb_color_control, CB_COLOR_CONTROL, MODE);
```

a pure extraction from the same word, with no dependence on `has_cb_color_control`. So word → mode is a total function by construction, `ambiguous` is always empty, and **that sentence prints unconditionally on every log prosper can produce.** A reader pasting it into #1706 is quoting a constant.

This is trap 264's own shape — *a diagnostic that prints a plausible constant* — in the PR that writes trap 264.

**The fix is small and makes it a real check, because the right function is already in the row.** Compare the reported mode against the field extracted from the raw word:

```python
mismatched = {(cbc, mode) for (_, _, cbc, mode, _, _) in per_pipeline
              if mode != ((cbc >> 4) & 0x7)}
```

That can fail, and it fails on my fixture. Then narrow the sentence to what the existing check supports: *"no raw word decodes to more than one mode, so a program showing several modes reflects several words the guest wrote, not one word decoded inconsistently."* That is the claim trap 265 needs and it is true; "faithful to what the guest wrote" is a claim about correctness that neither check reaches.

### 2. `--by-program`'s new NOTE is scoped to the truncated list, so it silently misses the case it exists for

`colorstate_report.py:199` slices `ranked = sorted(per_program.items(), …)[:top]`, and `:210` computes the warning as `if any(len(e["modes"]) > 1 for _, e in ranked)`. A program outside `--top` that carries several modes therefore gets **neither a row nor the NOTE**.

Measured on a 25-program log whose multi-mode program ranks 25th — same log, same tool, one flag apart:

| | `0xdead` row printed | NOTE fires |
| --- | --- | --- |
| `--top 20` (the default) | 0 | **0** |
| `--top 25` | 1 | 1 |

242 scanout draws parsed in both. So on the default invocation the reader is not told to re-key, in exactly the direction the NOTE exists to prevent — and the row that would have tipped them off is truncated away too.

**You already have the right pattern one function up.** `--by-pipeline` computes its verdict over the whole `per_pipeline` while truncating only its rows: with `--top 1` it prints one row and still reports *"2 distinct raw CB_COLOR_CONTROL word(s)"*. Do the same here — iterate `per_program.values()` for the condition, and consider saying how many programs were omitted by `--top`.

---

## The mutation test — it discriminates, but not through the assertion you highlighted

The amputation (dropping the raw word from the pipeline key) does redden the selftest, and the fixture is well built: `ps=0x2` appearing under two vertex shaders with two different raw words is a faithful miniature of the real `PPSA05143` case, and the added record is what makes `--by-program` show one program with two modes so the NOTE arm has something to fire on.

But the assertion you named as the mutation's target — `"0 decoding to more than one mode"` — is the **weakest** one in the block, for the reason in finding 1: it passes on *any* unmutated input, so it is not testing the fixture. It reddens under the amputation only because substituting a constant for the word artificially destroys a relationship that is otherwise guaranteed. That is a test of the key, not of the check.

The assertions that genuinely discriminate the feature are the plain ones either side of it: `es=0x1` **and** `es=0x5` both present, and both `cb-control=0x00cc0010` / `0x00cc0000` present. Drop either axis from the key and those fail, for the reason the flag exists. Keep them; they are doing the work.

## Non-blocking

- **The `!total` branch is still untested, and the comment names the wrong branch.** The two new arms — zero capacity and null buffer — exercise `if (!out || cap == 0) return false`, which is a *different* early return from `if (!total) { out[0] = '\0'; return false; }`. So the comment's *"the FALSE branch is covered by the two arms below"* is not accurate: the empty-population branch, which is the one the zero-suppression rationale is about, is still never reached. Marking the equality arm as non-discriminating is exactly right and I would keep it; the cheap addition is still one call to the summary at the **top** of `main()`, before any case counts, asserting `false` and an empty string.
- **Your new zero-suppression rationale is correct, and stronger than you state it.** `cb_mode_is_unmodeled_operation` excludes `NORMAL`, `RESOLVE` and `DCC_DECOMPRESS`, so indices 1, 3 and 6 can **never** be non-zero — a row of zeroes would print `mode1=0` on every run forever. On this very title that is the 1,320 + 2,533 `NORMAL` scanout draws reported as a zero. So it is not merely *"absence of a measurement mistaken for a measurement"*; it is a permanent structural zero that reads as a measured population, which is trap 264 again. One clause naming the three excluded modes turns the rationale from plausible into demonstrable. And yes — the retraction of the "usable as a control" reason is right, and stating why the old reason was wrong beside the new one is what makes it useful.

## One note on my own method, since it bears on the findings above

My first attempt at the truncation control reported `0` for both `--top 20` **and** `--top 25`, which would have been a nonsense finding. The fixture had been written to one path and read from another, so `grep -c` was counting a Python traceback rather than a report. The tool's own `--selftest` passing is what told me the instrument was fine and my harness was not. Both demonstrations above were re-run with the path fixed and with the discriminating arm (`--top 25`, and the `--top 1` by-pipeline contrast) showing the lever moves.

Fix the two above and I will approve on the next head. Everything else in this PR — the blocking re-normalisation at all four sites with the superseded pair kept visible, the `≥1.0` qualification, the CLEARLOG fix and its mutation-checked arm, the `[6:4]` closure, and the shader census — is unchanged and verified.

mattias800 (commented):
## APPROVED

Re-reviewed at `7d6402339`. Both blocking findings are fixed, and I verified each against the **same counter-fixtures I used to raise them** rather than against the new tests. One non-blocking correction below — a claim in a test comment that is too strong — and nothing else outstanding.

---

## Blocking finding 1 — fixed, checked with the fixture that exposed it

The hand-built log where the guest writes two words and both are reported as `mode=0` — every `NORMAL` draw mislabelled `DISABLE` — is the log the old verdict called *"faithful to what the guest wrote"*. On this head:

```
  MISMATCH es=0x3016e60000 ps=0x3017370000 cb-control=0x00cc0010 reported mode=0 but [6:4]=1
```

The tool now reads the raw word it was already printing. And the two checks are properly separated: the same run still says the value is pipeline-determined, **with the caveat attached to the sentence itself** — *"it says nothing on its own about whether the field is decoded correctly."*

**The `[6:4]` tripwire's framing is right and is not citable as evidence.** Its green line says, in the output rather than only in a comment, that the two fields *"are the same value by construction, so this CANNOT fail on a log prosper produced, and agreement here is not evidence about the decode."* That is the honest form of a check that holds by construction: kept as a tripwire on the log format, labelled so nobody quotes it. I also confirmed the tool's hardcoded `(cbc >> 4) & 0x7` matches prosper's own extraction exactly — `pm4_registers.hpp` gives `MODE_SHIFT = 4`, `MODE_MASK = 0x7` — so the tripwire fires only if the emitter stops deriving the mode from the printed word, which is what it claims.

**The replacement verdict comes back both ways on real-shaped data.** I built a log where one `(vertex, pixel)` pair carries two raw words:

```
  3 distinct raw CB_COLOR_CONTROL word(s) over 2 (vertex, pixel) pair(s); 1 pair(s) carry MORE THAN ONE word
    SPLIT es=0x300e900000 ps=0x3017370000 words=0x00cc0000,0x00cc0020
```

and `pipeline-determined` is **absent** from that run. Both branches reachable, and the failure mode is the one the line claims — a pipeline carrying more than one word means the pipeline is not the key. I constructed that fixture from the shape alone, before seeing your log; it reproduces the same `0x00cc0020` case your own run found, which is a small independent confirmation that the check is aimed correctly.

### That it fired on your own headline result is the argument for the finding

*"1 of 12 pairs carries more than one word"* qualifies *"the register value is pipeline-determined"* from twelve pairs to eleven, and the doc now says exactly that. **It does not touch the closure**, and I checked rather than accepted this: the extra draw is at `effective=0f`, and the per-program table independently records `0f`x2 alongside `07`x5,054 for that pixel shader, so the EFC draw is accounted for outside the 5,054 that `3,734 + 1,320` exhausts. The doc states the qualification and the closure side by side, which is the right resolution — the partition stands and the generalisation is narrowed.

## Blocking finding 2 — fixed, checked with my 25-program control

Same log, the **default** `--top 20`, where before there was neither a row nor a warning:

```
  NOTE: 1 program(s) carry more than one mode, 1 of them BELOW --top and not listed above (0xdead).
```

It names the hidden program, which is more than I asked for and is the part that makes the warning actionable when the row is truncated away.

## The mutation table — each mutant fails through its own assertion

This is the property I flagged as most likely to recur, so I applied the mutations myself rather than reading the table. Three one-line mutants, statuses captured directly (no pipe), and the failing assertion recovered from each traceback:

| mutation (1 line each) | rc | assertion that failed |
| --- | --- | --- |
| `[6:4]` comparison disabled (`if False`) | 1 | `assert "MISMATCH" in bad and "reported mode=0 but [6:4]=1" in bad` |
| NOTE scans `ranked` instead of `per_program` | 1 | `assert "NOTE: 1 program(s) carry more than one mode" in many` |
| SPLIT detection neutered (`split = {}`) | 1 | `assert "1 pair(s) carry MORE THAN ONE word" in sp` |
| clean | 0 | — |

**Each fails through the assertion it is supposed to fail through, and not through a neighbour** — which is the specific failure I raised about the previous mutation test and it does not recur. The three fixtures are also genuinely different from each other rather than one fixture reused: `BAD_MODE` breaks the log's internal consistency, `SPLIT` changes the guest's behaviour, `MANY` changes only the ranking. That is three axes, one per check.

On the `tail` incident: capturing the status before reading is the right fix, and self-reporting a trap-40 slip while citing trap 40 is worth more in the record than a clean run would have been. I verified the corrected table above independently, so it is not resting on the prose.

## Non-blocking — one claim that is too strong

The new comment says `!total` *"is not reachable from a test in this process at all: the counters are process-global and already non-zero by the time any arm runs, and nothing resets them."*

**The second half is true and the conclusion is not.** `g_unmodeled_cb_mode_counts` is a namespace-scope array of `std::atomic<uint64_t>`, so it is zero-initialised at static-initialisation time, and the first increment happens inside `main()` — the `disabled_state` case with `MODE = DISABLE` around `test_render_state.cpp:732-735`. I found no global constructor in that file that resolves a pipeline state before then. So a call to `unmodeled_cb_color_mode_summary` placed at the **top of `main()`, before the first case**, would see all eight slots at zero and take the `!total` branch — which is exactly the arm I suggested last round.

It is a trivial branch (returns false, writes `'\0'`) and I am not asking for the arm. What I am asking is that the comment not assert unreachability, because a recorded "this cannot be tested" is what stops the next person trying. *"Not reachable at this point in the file, because the cases above have already counted; an arm before them would reach it"* is both true and still explains why it is absent. This item has now moved from naming the wrong branch to claiming an unreachability that does not hold, so it is worth getting right on the way past.

## Also verified

- The zero-suppression rationale is now the structural one, with the numbers: `cb_mode_is_unmodeled_operation` excludes `NORMAL(1)`, `RESOLVE(3)` and `DCC_DECOMPRESS(6)`, so those three slots can never be incremented and a row of zeroes would print `mode1=0` on every run of every title — on this one, against 1,320 + 2,533 `NORMAL` scanout draws. *"A permanent structural zero wearing the shape of a measured population"* is the accurate description and it is stronger than the reason it replaced.
- The `--by-pipeline` verdict continues to scan the whole population while truncating only its rows (`--top 1` still reports the full distinct-word count), and `--by-program`'s warning now matches that discipline.
- Baseline `--selftest` passes on this head; I used it as the control that separated my own harness from the tool, as last round.

Everything from earlier rounds is unchanged and verified: the CLEARLOG read-before-assign fix and its mutation-checked arm, the four-site re-normalisation with the superseded pair kept visible, the `≥1.0` qualification, the `[6:4]` closure, and the shader census.

Approved at `7d6402339`.

mattias800:
## Post-merge: one comment I left in this PR is wrong, and it is the kind that stops future work

The reviewer's last non-blocking finding is correct and I want it recorded where the comment's own
history is, since I have torn down the branch worktree and will not be the one who acts on it.

In `tests/gpu/state/test_render_state.cpp` I wrote that the `!total` early return of
`unmodeled_cb_color_mode_summary` "is not reachable from a test in this process at all". **The
premise is true and the conclusion is false.** `g_unmodeled_cb_mode_counts` is a namespace-scope
`std::atomic` array, zero-initialised at static init, and the first increment happens inside `main()`
— so **an arm at the top of `main()`, before any case runs, reaches the branch.** The counters are
process-global and never reset, which makes the branch unreachable *from where I put my arms*, not
unreachable in principle.

That distinction matters more than the missing coverage. A comment saying a branch is untestable is
read as a decision already taken, and the next person does not retry it — which is exactly the shape
this PR spent its last round fixing elsewhere (a check that could not fail, presented as a check).

**Suggested fix**, for whoever next touches that file: replace the "not reachable" sentence with the
accurate one — *the branch is unreachable from an arm placed after this file's other cases, because
the counters are process-global and already non-zero; an arm at the top of `main()` would reach it* —
and, if it is cheap, add that arm. It is a two-line change and I am not opening a PR for it alone.

Everything else in the review is addressed at the merged head: the falsifiable `--by-pipeline`
verdict (which fired on the first real log and qualified my own headline result — 11 of 12 pipelines,
not all twelve), the `--by-program` NOTE scanning the whole population rather than the `--top` slice,
and the structural-zero rationale.

Refs #1706, #2014

Issue #3364, opened 2026-09-05
Space Adventure Cobra: every UI glyph renders sheared and fragmented — the background behind it is correct


mattias800:
> **READ THIS FIRST — the most valuable next move is auditing the list below, not extending it.**
>
> Sixteen stages of this frame have each been measured and each verified correct, **and the defect
> persists**. Within one draw, one shader, one pipeline state, one texture and one open stencil gate,
> four `I` glyphs with byte-identical UV rects, size and colour — two render and two produce nothing.
>
> **One of those measurements is more likely to be subtly wrong than all of them are to be right while
> the picture is still broken.** The screen mapping was the row I could most easily doubt; it was
> audited and it held. Two rows carry the most inference and have not been independently re-derived:
>
> 1. **The A/D control's IoU scoring**, which underpins every "renders / does not render" verdict in
>    this issue.
> 2. **The `--bundle-zero-boundary` replay**, where "nothing is seeded" is a claim about the harness
>    rather than about the frame.
>
> Please audit before adding hypotheses. Five have died here, each specific and code-grounded.

---

## What

*Space Adventure Cobra — The Awakening* (`PPSA17337`) renders its intro cinematic and menu backgrounds
correctly, but **every glyph of its UI text is malformed** — letters sheared and fragmented, with stray
diagonal strokes crossing them.

On the brightness-calibration screen the string is partly legible as *"ADJUST THE BRIGHTNESS UNTIL
THE SYMBOL … IS BARELY V[ISIBLE]"* — the tail of the string is genuinely absent — and the button prompts
at bottom-right (`BACK`, `ENTER`) are mangled the same way. The background behind it draws **correctly**,
so this is specific to the text layer.

Verified at **1:1** on the native 1920x1080 PNG. Not a downscaling artefact.

**Second and third instances:** *Blasphemous 2* `PPSA13579` (Unity — intro-cutscene subtitles, first
glyph nearly correct then degrading) and *Alex Kidd* `PPSA02664`. Three titles, three engines, one
signature: text malformed, background behind it correct.

## Reproduction

Route `scripts/cobra/reach-title-or-gameplay.pad`, frontend `tools/screenshot`, native 1920x1080,
`PROSPER_RENDER_SCALE=1`, headless Linux/RADV. The defect is **deterministic** — the paragraph is
bit-identical across five consecutive replays — and reproduces **entirely offline** from a captured
frame bundle, so no boot is needed to work on it:

```
[...]

Issue #1268, opened 2026-07-23
perf(gpu): render_draws_rgba re-hashes full buffer contents per draw-reference — dominant Blue Prince term post-#1266


mattias800:
Post-#1266 profile of Blue Prince (PPSA25009, RADV, the #1264 route) shows the remaining ~100% of one core during the ~1.9 fps intro/loading phase inside `prosper::test::render_draws_rgba`, with 4 of 5 gdb samples specifically in `hash_buffer_words` and the 5th in `__memmove_avx512` readback (full stacks on #1177). This is the render-path term that now solely gates BP's first room (#1216) and is the same family as Bendy's #1177 crawl.

## Mechanism (code, current master)

- `tests/render_runner.h:65` — `hash_buffer_words` is a scalar FNV-1a (xor+mul per dword: a serial dependency chain, ~1 dword/few cycles, far below memory bandwidth).
- `tests/render_runner.h:2924` — for **every storage-buffer resource of every draw**, the FULL buffer contents are hashed to build `SharedBufferKey{ptr, words, identity, content_hash, tag}` for the shared-buffer reuse lookup. The hash IS the key, so the dedup cache pays full content hashing on every reference — even when a pass batch of 18..490 draws references the same vertex/uniform ranges, and again on the next batch and next frame.
- Contrast: the shader path at `render_runner.h:2380` already short-circuits hashing via `fs_identity` — precedent for identity-based short-circuit.
- BP submits during loading carry ~4,000 draws; several buffers per draw × full-content FNV per reference = plausibly GiB/s of hashing per frame, matching the observed one-core saturation on an otherwise ~91% idle machine.

## Plan (measure first)

1. Gated counters (`PROSPER_HASH_STATS=1`): per frame — total bytes hashed, hash call count, distinct (ptr,words) ranges, repeat ratio within a submit, share of references carrying `buffer_identity`, plus the same for the texture/readback path.
2. Expected fixes, by measured leverage:
   a. **Per-submit hash memoization** keyed by (ptr, word_count, identity): correct within a submit under the same semantics as hardware (a guest thread racing writes against an in-flight submit is undefined on the real GPU too). If batches re-reference ranges heavily, this alone is a 10-100x cut.
   b. **Bandwidth-friendly hash** (multi-lane mixing instead of the serial FNV chain, own implementation) for the residual hashed bytes.
[...]

PR #3370, opened 2026-09-05
render: PROSPER_BUFVERIFY re-reads uploaded buffers from device memory

mattias800:
## Why

Investigating #3364 needed a measurement the renderer could not make: **do the bytes a capture proves
correct still hold in the memory the GPU is about to read?**

`PROSPER_BUFLOG` reports the *source* words of each storage-buffer upload. It never looks at the
destination, so it cannot see the destination being clobbered between the memcpy and submission — an
overlapping arena slice, a stray write, a pooled buffer handed out twice. Cobra's symptom (a contiguous
range of a text mesh's vertices producing nothing while the rest of the same buffer renders) is that
family's signature, and without a readback it could only be argued about.

## What

`PROSPER_BUFVERIFY=1` records every uploaded storage buffer and, **just before the pass's command buffer
is closed**, re-reads it from its mapped device-visible allocation and compares it byte for byte against
the guest words it was built from:

```
[bufverify] 91 buffer(s) re-read from device memory, 0 mismatched, 0 skipped (no host mapping)
[bufverify] MISMATCH set=0 binding=7 id=2079b49510 source_bytes=11520 device_bytes=11520
            first_differing_byte=6528 (vertex 204 at stride 32)
```

End-of-pass placement is the point: verifying right after the memcpy would miss every clobber-by-reuse
case, which is the class worth suspecting.

### What a zero from it does NOT retire

**It cannot observe a truncation.** The destination `range` is assigned the same byte count that drove
the memcpy and that the recorded word count describes, so source and device sizes are equal *by
construction* on every buffer it checks. A short upload is structurally inexpressible here; that is
`[buffer-truncated]`'s and `PROSPER_BUFLOG`'s question. The comments, the README and the #3364 record all
say so — an earlier revision of this PR claimed otherwise and that over-claim is corrected on the issue.

A mismatch also has a false-positive direction: the `#1268` dedup path deliberately tolerates
cross-thread guest writes to the source words, so a guest that rewrites them after the memcpy trips this
without any prosper defect.

**`skipped (no host mapping)` is load-bearing.** A transient upload (neither arena nor pool) unmaps its
memory before returning and cannot be re-read. `PROSPER_NO_BACKEND_BUFFER_POOL=1` sends *every* upload
down that path, so that arm verifies nothing — and now says so rather than printing a vacuous zero:

```
[bufverify] 0 buffer(s) re-read from device memory, 0 mismatched, 91 skipped (no host mapping)
[bufverify] NOTHING WAS VERIFIED: all 2 recorded buffer(s) lack a host mapping (transient uploads).
            A zero from this run means the check never ran, not that the uploads are correct
```

## The control ships with it, and is aimable

Nothing in a normal run exercises the mismatch branch, so a deliberate-corruption control comes with the
check. `PROSPER_BUFVERIFY_MUTATE=N` flips one byte of a device copy and the next line must name that
offset:

```
[bufverify] MUTATED device byte 6528 of set=0 binding=7 source_bytes=11520 -- the next line must
            report a mismatch at exactly that offset
[bufverify] MISMATCH set=0 binding=7 id=2079b49510 source_bytes=11520 device_bytes=11520
            first_differing_byte=6528 (vertex 204 at stride 32)
```

**Aiming it is not a convenience.** The first version always mutated whichever buffer was recorded first
— a 128-byte uniform buffer — and fired perfectly. That proves the compare loop runs and says nothing
about the 7 KiB vertex buffer whose zero was about to be quoted: a control drawn from a different size
class than the null tests the discriminator, not the domain. `PROSPER_BUFVERIFY_MUTATE_BINDING` and
`PROSPER_BUFVERIFY_MUTATE_MINBYTES` aim it at the binding and size class you intend to believe.

If nothing matches, the run prints `CONTROL DID NOT FIRE … UNVALIDATED` rather than falling silent,
because silence there is indistinguishable from a clean verified run. All three values parse strictly
(decimal or `0x` hex, bounded at `UINT32_MAX`); a malformed one disarms the control loudly.

## Verification

```
cmake --build prosper/build-linux -j6                 # clean
ctest --no-tests=error -j4                            # 363 tests discovered
```

**363 tests, 358 passed, 5 failed**, and all five (`recompiled_shaders_render`, `multidraw_render`,
`indexed_render`, `descriptor_array_render`, `gpu_execute`) fail **identically on unmodified
`origin/main`** in the same worktree — established by stashing this diff, rebuilding and re-running.
They are flaky on this AMD/RADV box: repeated `test_indexed_render` runs on clean main give 4, 5, 4, 2,
3 and once **0** failures. Filed separately as #3371. CI does not see them because its Linux job renders
on a software implementation.

Observational by construction: unset, nothing is recorded and no comparison runs; enabled, the rendered
frame is byte-identical to the untouched run (same MD5).

## What it established on #3364

133 buffers re-read across one *Space Adventure Cobra* frame, **0 mismatched**, with the control fired at
byte 6528 of the same binding and size class — the exact offset where that frame's text stops rendering.
Together with `PROSPER_BUFLOG` showing full, correctly-addressed uploads with zero `[buffer-truncated]`
lines, and a four-arm arena/pool/share A/B producing byte-identical output, that retires the upload,
truncation, dirty-range and buffer-reuse families for the defect and points it downstream of resource
realization.

## Review

Independently reviewed. First round **REJECTED** with four blocking findings — a silently skipped upload
class that made a whole A/B arm vacuous, an over-claim about truncation, a control that could decline to
fire in silence, and non-strict parsing of the aiming values. All four fixed in `39c3efad6`; re-reviewed
and **APPROVED**. `933daba3a` adds one further guard the re-review raised as non-blocking (bounding the
aiming values at `UINT32_MAX`, since the binding filter compares as `uint32_t`); the whole delta from the
approved head is that guard.

Refs #3364

### DIFF
--- a/prosper/tests/fixtures/render_runner.h
+++ b/prosper/tests/fixtures/render_runner.h
@@ -6228,6 +6228,7 @@ inline std::vector<uint8_t> render_draw_pass_rgba(std::span<const BackendDraw> d
     const bool share_backend_resources =
         getenv("PROSPER_NO_BACKEND_RESOURCE_SHARE") == nullptr;
     const bool reuse_host_buffers = render_host_buffer_pool_enabled();
+    const bool buffer_verify_enabled = getenv("PROSPER_BUFVERIFY") != nullptr;
     struct SharedBufferKey {
         const uint32_t* words = nullptr;
         size_t count = 0;
@@ -6272,6 +6273,22 @@ inline std::vector<uint8_t> render_draw_pass_rgba(std::span<const BackendDraw> d
         bool persistent = false;
     };
     std::vector<SharedBufferUpload> shared_buffers;
+    // PROSPER_BUFVERIFY=1 — re-read every uploaded storage buffer from its MAPPED, device-visible
+    // allocation just before the pass is submitted, and compare it byte for byte against the guest
+    // words it was built from. This is the readback that answers "did the bytes the capture proves
+    // correct actually reach the GPU?", which PROSPER_BUFLOG cannot: BUFLOG reports the SOURCE, so a
+    // bad memcpy, a short copy, or a later arena/pool slice reusing the same memory all pass it. The
+    // check is deferred to end-of-pass on purpose — verifying at upload time would miss exactly the
+    // clobber-by-reuse case that motivates it.
+    struct BufferVerifyRecord {
+        const uint32_t* words = nullptr;
+        size_t word_count = 0;
+        uint64_t identity = 0;
+        uint32_t set = 0;
+        uint32_t binding = 0;
+        size_t upload_index = 0;
+    };
+    std::vector<BufferVerifyRecord> buffer_verify_records;
     std::vector<SharedBufferArena> shared_buffer_arenas;
     std::unordered_map<SharedBufferKey, size_t, SharedBufferKeyHash> shared_buffer_indices;
     // Per-call repeat-reference memo (#1268): draws in one pass batch overwhelmingly re-reference
@@ -7290,6 +7307,9 @@ inline std::vector<uint8_t> render_draw_pass_rgba(std::span<const BackendDraw> d
                                                              &res_buffer_index_insert_ms);
                         shared_buffers.push_back(upload);
                         shared_buffer_indices.emplace(buffer_key, buffer_index);
+                        if (buffer_verify_enabled)
+                            buffer_verify_records.push_back(
+                                {words, word_count, identity, set, binding, buffer_index});
                     }
                     ++resource_reuse_stats.unique_buffers;
                     ++backend_hash_stats_totals().unique_buffers;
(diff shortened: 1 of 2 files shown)


mattias800 (commented):
## Independent review of 12b5dfeb7 — blocking findings

Four ways this could report a zero that does not mean what a reader would take it to mean.

**B1 — a whole upload class is silently skipped and the summary cannot say so.**
`create_transient_storage_buffer_upload` unmaps its memory before returning, so every upload taking the
`!upload.arena && !upload.pooled` fallback has `mapped == nullptr`. Those were dropped before the
`checked` counter, and the summary printed only `checked`. `PROSPER_NO_BACKEND_BUFFER_POOL=1` clears
`reuse_host_buffers`, which also clears `use_buffer_arena`, so *every* upload goes transient and the
diagnostic prints a completely vacuous `0 buffer(s) re-read, 0 mismatched`. Not hypothetical: the no-pool
arm of the #3364 A/B is exactly that configuration.

**B2 — the size arm is structurally inexpressible, and the comment advertised it as the point.**
On the only branch the check examines, `upload.range` is assigned the same `bytes` that drove the memcpy
and that the recorded `word_count` describes, so `source_bytes == device_bytes` by construction and the
`"none (size differs only)"` branch is unreachable. The instrument cannot observe a truncation, yet the
comment said the question it answers is "which range computation truncated the upload?" and the PR body
claimed the zero retires the "truncation" family. Trap-122 shape: the control proves the compare loop
runs, not that the check could express the class being tested for.

**B3 — the positive control can silently decline to fire.** If the binding/size filter matches nothing,
or the matched record has no host mapping, the block falls through printing nothing and the run emits the
ordinary `0 mismatched` — the exact output of a clean verified run. The only tell is an absence.

**B4 — non-strict parsing of the two aiming variables.** `atoll` returns 0 for anything unparseable, so
`BINDING=b7` silently means binding 0 and `MINBYTES=7k` silently means no size gate, returning the
operator to the un-aimed control while still printing a confident `MUTATED` line. The file already
includes `diagnostics/env_numeric.hpp`, added by #3267 for this exact defect.

Non-blocking: base-0 octal in `MUTATE` (`010` means 8); `errno` used without `<cerrno>`; byte-at-a-time
compare where memcmp would do; the tolerated cross-thread guest write to the source words is a
false-positive direction and is undocumented; no entry in `tools/gpu_replay/README.md`, which the charter
names as the home for every `PROSPER_*` graphics diagnostic; dead `buffer_verify_records.clear()`.

Checked and clean: genuinely inert when disabled (one `getenv` per pass, no allocation, the hot-loop
branch only on the unique-upload path); lifetimes sound (`shared_buffers` is only appended to and is
moved out *after* the verify point, and every `words` provenance outlives it); the device read is safe
and `up.range` is the right field (`up.bytes` would be wrong — the arena path never sets it and the
pooled value is allocation capacity); no cross-pass clobber hole; MUTATE cannot fire by accident.

REJECTED

mattias800:
## Author response — all four blocking findings fixed in 39c3efad6

(The review above was written against `12b5dfeb7`; it was posted after the fix commit was already
pushed, so GitHub attaches it to `39c3efad6`. Read it as the review *of the parent*.)

**B1 — skipped uploads are now counted and a vacuous run says so.** Transient uploads have no host
mapping and cannot be re-read; they are now counted separately, and a pass that verified none of its
records prints `NOTHING WAS VERIFIED`. Measured on the arm that produced the vacuous zero:

```
PROSPER_NO_BACKEND_BUFFER_POOL=1
[bufverify] 0 buffer(s) re-read from device memory, 0 mismatched, 91 skipped (no host mapping)
[bufverify] NOTHING WAS VERIFIED: all 2 recorded buffer(s) lack a host mapping (transient uploads).
            A zero from this run means the check never ran, not that the uploads are correct
```

**B2 — the claim is narrowed, in the code and everywhere it had already travelled.** The block comment
and the struct comment now state that a short upload is structurally inexpressible here and that
truncation is `[buffer-truncated]`'s and BUFLOG's question; `tools/gpu_replay/README.md` says the same;
and I have posted a correction on #3364 re-attributing which measurement retires which claim, because
the over-claim was already in that issue and a correction that lives only in a review thread is not a
correction.

**B3 — silence is now audible.**

```
PROSPER_BUFVERIFY_MUTATE=6528 PROSPER_BUFVERIFY_MUTATE_BINDING=99
[bufverify] CONTROL DID NOT FIRE: no recorded buffer matched binding=99 minbytes=0 with a host
            mapping and offset 6528 in range. The "0 mismatched" below is UNVALIDATED
```

The target selection now also requires a host mapping and an in-range offset, so a matched-but-unusable
record takes the loud path instead of the silent one.

**B4 — strict parsing via the helper the file already includes.**

```
PROSPER_BUFVERIFY_MUTATE_BINDING=b7
[bufverify] PROSPER_BUFVERIFY_MUTATE_BINDING="b7" is not a number (decimal or 0x hex);
            the control is DISARMED rather than aimed somewhere unintended
```

All three variables go through `prosper::diag::parse_u64_auto_base`, which also drops the octal leg, so
`MUTATE=010` no longer means byte 8. A malformed value disarms rather than aiming.

**Control still fires on the real target after the changes:**

```
[bufverify] MUTATED device byte 6528 of set=0 binding=7 source_bytes=11520 -- the next line must
            report a mismatch at exactly that offset
[bufverify] MISMATCH set=0 binding=7 id=2079b49510 source_bytes=11520 device_bytes=11520
            first_differing_byte=6528 (vertex 204 at stride 32)
```

**Non-blocking, addressed:** memcmp before the byte scan; the `MUTATED` line now reports the target's
`source_bytes` since size is the second aiming axis; the tolerated cross-thread guest write is
documented as a false-positive direction; the dead `clear()` is gone; the three variables are documented
in `tools/gpu_replay/README.md`; `errno`/`strtoull` are gone entirely with the strict parser, so the
missing `<cerrno>` is moot. Not taken: `PROSPER_ENV_ON` for the top-level switch, which would diverge
from its two immediate neighbours for a once-per-pass read.

**Verification at 39c3efad6:** clean full build; `ctest --no-tests=error -j4` discovers **363** tests,
**358 passed, 5 failed** — `recompiled_shaders_render`, `multidraw_render`, `indexed_render`,
`descriptor_array_render`, `gpu_execute`. All five fail **identically on unmodified `origin/main`** in
the same worktree (established by stashing the diff, rebuilding and re-running) and are flaky on this
AMD/RADV box: repeated runs of `test_indexed_render` on clean main give 4, 5, 4, 2, 3 and once 0
failures. Filed as #3371. `ascii_output_literals`, which the previous head genuinely broke, passes.

mattias800 (commented):
## Re-review of 39c3efad6 — all four blocking findings fixed

**B1 — FIXED.** `skipped_unmapped` increments on both drop paths and is printed in the summary, with a
loud banner when `checked == 0` and records exist. The drop class is real and is exactly what the comment
says: `create_transient_storage_buffer_upload` unmaps before returning and never assigns
`upload.mapped`, and `use_buffer_arena` is gated on `reuse_host_buffers`, which
`render_host_buffer_pool_enabled()` turns off under `PROSPER_NO_BACKEND_BUFFER_POOL` — so that arm really
does send every upload down the unmapped path and now says so. No remaining path prints a bare
over-trustable zero.

**B2 — FIXED in code, README and the issue record.** I checked the "by construction" claim rather than
trusting it: the only site where `range` diverges from `word_count*4` is the zero-word fallback, which
sits inside the transient branch whose uploads are always skipped. So the equality genuinely holds on
every *checked* buffer, and the narrowed comments and README are accurate about what a zero does and does
not retire. The false-positive direction is now documented.

**B3 — FIXED.** The selection loop now rejects any candidate with an invalid upload index, no host
mapping, or a range not containing the offset, so a selected target is always mutatable, and the armed
no-target case prints `CONTROL DID NOT FIRE … UNVALIDATED`. The unarmed case cannot reach that message.
The "next line must report a mismatch" promise is sound: records are pushed only on the unique-buffer
creation path, one per `shared_buffers` entry, so exactly one record covers the mutated upload.

**B4 — FIXED.** `prosper::diag::parse_u64_auto_base` replaces `atoll`/`strtoull`; its grammar is decimal
or `0x` hex, overflow-checked, with the octal leg deliberately excluded. A malformed value in any of the
three disarms rather than aims.

**New defects:** none blocking. The target now depends on `mutate_offset`, which *strengthens* the
control (a too-small binding-7 record is passed over in favour of one whose range contains the offset)
and the `MUTATED` line reports `source_bytes` so the reader can see which size class fired. One
non-blocking nit: the binding filter compares as `uint32_t`, so a value above `UINT32_MAX` would truncate
and silently aim elsewhere.

**All six first-round non-blocking points addressed**: octal rejected, `errno` gone entirely,
`std::memcmp` fast path with the byte loop retained only to localize, false-positive direction
documented, README section added, dead `clear()` removed.

**One must-fix that is not code:** the PR body still carried the B2 over-claim, stale sample output and
stale test counts.

APPROVED

mattias800:
The must-fix from the re-review is done: the PR body is rewritten and no longer claims this check retires truncation, the sample output matches the current three-field format, and the test counts read 358/5. The non-blocking `UINT32_MAX` nit is closed in `933daba3a`; `git diff 39c3efad6..933daba3a` is that guard alone (11 insertions, 3 deletions, one file), so the approval above still describes what merges.

Issue #3374, opened 2026-09-05
Five rung-6 Unity titles render their scene through large sheared/stretched triangles (Joe & Mac, Asterix, Summer Sports, Evergate, Rugrats)


mattias800:
## Summary

Five rung-6 titles render their correct scene art with **large sheared/stretched triangles smeared across the frame**. The art, lighting and HUD content are all present and recognisable; the geometry is not. Measured on `d38e892cc` at native scale (`PROSPER_RENDER_SCALE=1`, `PROSPER_RENDER_EVERY=1`) with each title's own reviewed guard route, one title at a time on an otherwise idle box.

This is filed as one issue because the signature is the same in all five and they are all Unity/IL2CPP; **that is a hypothesis, not an established shared root cause.** Nothing here bisects it, and five causes is as consistent with the data as one.

| title | id | engine | what the corruption looks like | SSIM vs its own reviewed refs |
| --- | --- | --- | --- | --- |
| New Joe & Mac | `PPSA02801` | Unity 2021.3 / IL2CPP | level 1's volcano, palms, flowers and Joe are all drawn; sprites smeared into large triangular shards over the lower half, HUD garbled into streaks | 0.719 best, 0 of 249 above the 0.85 floor |
| Asterix Slap Them All | `PPSA08576` | Unity / IL2CPP | forest level's trees, grass, mushrooms present; Asterix's sprite stretched into a single streak across the full width of the top of the screen, HUD smeared with it | 0.508 best, 0 of 249 |
| Summer Sports Games | `PPSA03416` | Unity / IL2CPP | javelin event reached at native 3840x2160 — athlete, crowd, stands, ATTEMPT counter all drawn; huge sheared polygons across sky and stands, ATTEMPT banner clipped | 0.373 best, 0 of 199 |
| Evergate | `PPSA01885` | Unity | worst case: the whole frame is stretched white/orange shear, with the crimson backdrop and gold ornaments visible underneath | 0.647 best, 0 of 289 |
| Rugrats | `PPSA23396` | Unity 2022.3 / IL2CPP | nursery level correct — pink wall, playpen gate, padlock, both babies — but a black diagonal wedge removes more than half the frame | 0.411 best, 0 of 239 |

## Why this is not the measuring apparatus

- **Provenance is recorded, not assumed**: `rendered_samples == samples` and `guest_scanout_samples == 0` in all five runs, so no frame is a republished guest scanout.
[...]

Issue #3371, opened 2026-09-05
Five Vulkan-execution tests fail non-deterministically on main on AMD/RADV, all on the indexed vertex-fetch path (CI cannot see it)


mattias800:
## What

Five Vulkan-execution tests fail **on unmodified `origin/main`** on an AMD/RADV box, and they fail
**non-deterministically** — the same binary, same working directory, back-to-back runs, no other GPU
user on the machine.

```
363 tests discovered, 357 passed, 6 failed
  338 - recompiled_shaders_render
  341 - multidraw_render
  344 - indexed_render
  345 - descriptor_array_render
  355 - gpu_execute
```

(The sixth, `ascii_output_literals`, was a real fault in the branch under test and is unrelated.)

## The flakiness, measured

`./test_indexed_render` run repeatedly on clean `origin/main` at `1e1b8bb86`, `[FAIL]` count per run:

| configuration | run 1 | run 2 | run 3 | run 4 | run 5 |
| --- | --- | --- | --- | --- | --- |
| default | 4 | 5 | 4 | 2 | 3 |
| default (later block) | 3 | 3 | 3 | | |
| `PROSPER_NO_BACKEND_BUFFER_ARENA=1` | 1 | 2 | 4 | | |
| `PROSPER_NO_INDEX_ARENA=1` | 4 | 4 | **0** | | |

A run that reports **0** failures exists, so the code under test can satisfy every assertion; and no
arena lever removes the variance, so this is not localized to the index or storage arena.

## The failing assertions

```
== test_indexed_render ==
  [ok]   recompiled vertex-fetch VS + green PS
  [ok]   both quad renders produced frames
  [FAIL] indexed [0,1,2,2,3,0] TRIANGLE_LIST quad fills the viewport GREEN
  [FAIL] indexed quad is byte-identical to the old perimeter-fan rendering
  [ok]   indexed triangle rendered
  [FAIL] indices [1,2,3] fetched records 1..3 (fullscreen GREEN triangle)
  [FAIL] dropping the indices changes the picture (indices are really applied)
  [ok]   non-indexed firstVertex selects records 1..3 from a shared vertex pool
  [ok]   indexed vertexOffset selects records 1..3 from a shared vertex pool
```

and

```
== test_multidraw_render ==
  [FAIL] a second indexed draw in one pass reads ITS OWN slice at a non-zero arena offset
         (degenerate indices contribute nothing; binding at 0 would re-draw the first)
```

[...]

Issue #2950, opened 2026-08-23
Three rung-6 snapshot guards fail on current master — The Messenger, Dead Cells and GRIS collapse to slivers on real hardware, and CI runs lavapipe


mattias800:
> **Headline, added after the investigation closed:** this is **not** a recent prosper regression and
> **not** the Mesa version. A 2026-08-02 build with byte-identical guard config, baselines and pad
> scripts fails identically, and Mesa 26.0.3 fails identically to 26.1.4. Both hypotheses are dead;
> the detail is in the comments. The original report below is kept as written.

## Three rung-6 snapshot guards fail on current master, with the collapsed-geometry signature

Measured 2026-08-23 on Linux / AMD Radeon 8060S (RADV STRIX_HALO), Mesa 26.1.4, against
`origin/master` `f5624cf6`.

| guard | title | verdict |
| --- | --- | --- |
| `messenger-scene` | *The Messenger* `PPSA24651` | **FAIL** — structural matches 0 < 22; non-black coverage matches 0 < 22 |
| `dead-cells-gameplay` | *Dead Cells* `PPSA15552` | **FAIL** — structural matches 0 < 33 |
| `gris-gameplay` | *GRIS* `PPSA09804` | **FAIL** — qualifying frames 0 < 60; pixel changes 45 < 60; structural matches 0 < 60 |
| `blue-prince-title` | *Blue Prince* `PPSA25009` | OK — frames 10, qualifying 10, structural 10, non-black 10 |

`blue-prince-title` passing in every arm is the positive control: the harness, the dumps, the
frontend and the machine are all working.

### It is not the PR it was found from

Found while verifying #2949. Three arms, all on this machine today:

| arm | what it is | messenger | dead-cells | gris | blue-prince |
| --- | --- | --- | --- | --- | --- |
| A | #2949 branch head, both its fixes active | FAIL 0/22 + 0/22 | FAIL 0/33 | FAIL 0/60 | OK |
| B | the SAME binary with both fixes disabled by env (`PROSPER_NO_RENDERPASS_EXTERNAL_DEPS=1 PROSPER_NO_DEAD_VARYING_ELIM=1`) | FAIL 0/22 + 0/22 | FAIL 0/33 | FAIL 0/60 | OK |
| C | the six changed sources reverted to `origin/master` and `screenshot` rebuilt | FAIL 0/22 + 0/22 | *(running)* | — | — |

[...]

PR #3375, opened 2026-09-05
Record the four-title Unity shear as a cross-implementation split, and add PROSPER_INDEX_ECHO

mattias800:
## What this is

An investigation result for **#3374** (five rung-6 Unity titles render their scene art through large
sheared triangles), plus the one instrument that was missing to reach it, plus the trap that nearly
sent it the wrong way.

**The finding: it is one defect across four of the five titles, and the split is decided by the Vulkan implementation.**
The same captured frame, the same binary and the same 88 decoded operations render the correct scene
when a different Vulkan implementation executes them.

```bash
gpu_replay <work>/submit.prgcap out_radv.bmp                          # sheared
VK_DRIVER_FILES=/usr/share/vulkan/icd.d/lvp_icd.x86_64.json \
gpu_replay <work>/submit.prgcap out_lvp.bmp                           # the level, correct
```

Reproduced on four titles: *New Joe & Mac* `PPSA02801`, *Asterix & Obelix: Slap Them All!* `PPSA08576`,
*Rugrats* `PPSA23396` and *Summer Sports Games* `PPSA03416` (3D, native 3840x2160). *Evergate*
`PPSA01885` is the **exception** — its blown-out frame reproduces on lavapipe too, so that half is
prosper's. Full evidence and the falsification list are on #3374.

**Following review, this PR does NOT assign the fault.** A cross-implementation split localises the
disagreement; it does not say who is wrong (instrument trap 38, whose own worked example is a
lavapipe/RADV split that turned out to be prosper's UB). Both implementations execute prosper's own
recompiled SPIR-V. The open fork is a RADV defect against UB in the recompiled vertex stage; Mesa 26.2.1
returns the byte-identical broken frame to 26.1.4, so a version bump is not the answer either.

## Failure scenario and contract

`PROSPER_INDEX_ECHO=1` prints, per indexed draw and immediately before `vkCmdBindIndexBuffer`, the index
bytes read back from the host-visible memory at exactly the `(buffer, offset)` pair about to be bound,
against the indices prosper decoded for that draw.

The gap it closes: a draw that fetches vertices out of its own range — because it consumed another draw's
index slice, or because `vertexOffset` pushed `gl_VertexIndex` past the bound vertex range — is invisible
to `PROSPER_BUFLOG` and `PROSPER_BUFVERIFY` (both look at STORAGE buffers) and produces **no** Vulkan
validation message, since an out-of-range storage read is `robustBufferAccess` rather than an error. So
"prosper bound the wrong indices" and "the driver fetched the wrong indices" produced identical evidence
and point at different files. On the Joe & Mac frame it reports 85 indexed draws, `mismatched=0` on every
one, `got_max < vcount` on every one — which is what let the index family be retired at title scale.

Invariants: gated on the environment variable, read-only, and it never dereferences `imapped` unless the
arena path set it (a dedicated index buffer unmaps its memory and prints
`got=<dedicated buffer, unmapped>` rather than a silent zero).

- Affected: nothing, unless `PROSPER_INDEX_ECHO` is set. The added `DV::imapped` field is written on the
  arena path and read only inside the diagnostic.
- Deliberately unaffected: no change to what is bound, uploaded, recorded or drawn. Output is
  byte-identical with the variable unset — the Joe & Mac capsule hashes the same before and after.

## Risks

Low. One pointer field and one gated `fprintf` in the record loop. The pointer is into the shared host
buffer arena, which outlives the record loop; the dedicated-buffer path leaves it null and the diagnostic
says so rather than reading freed memory.

## Verification

Worktree build off `origin/main` `d38e892cc`, Linux / AMD Radeon 8060S (RADV STRIX_HALO), Mesa 26.1.4.

```
cmake --build prosper/build-linux -j10          # clean, 0 errors
ctest --no-tests=error -j2                      # RC=8, 364 discovered, 5 failed
```

The five failures are `recompiled_shaders_render`, `multidraw_render`, `indexed_render`,
`descriptor_array_render`, `gpu_execute` — **exactly the five #3371 records failing on unmodified
`origin/main` on this box**, and the subject of this very investigation. They are not introduced here: the
diff adds one gated diagnostic and documentation.

Docs gate:

```
# 14 table(s), 328 rows, unbroken; numbered column 1..266, unique and ascending
git ls-files '*.md' -z | xargs -0 python3 prosper/tools/docs/check_numbered_table.py   # clean
git diff --check                                                                       # clean
```

No snapshot guards were run: this PR changes no rendering behaviour, and #2950 records that the guards on
this machine are not currently a usable renderer oracle.

## Screenshots

Offline `gpu_replay` replays of headlessly captured frames (each guard's own `pad_script`,
`savedata_policy: fresh`, `PROSPER_RENDER_SCALE=1`, capture armed with
`PROSPER_CAPTURE_BUNDLE_TRIGGER_FILE`). The correct ones are executed by a **software** Vulkan device, so
they are diagnosis rather than a rung claim — nobody can play on lavapipe. No retouching, compositing or
forced guest state; the images are the replay output converted to WebP.

| title | RADV | same capsule, lavapipe |
| --- | --- | --- |
| New Joe & Mac `PPSA02801` |  |  |
| Asterix `PPSA08576` |  |  |
| Rugrats `PPSA23396` |  |  |

`BLOG.md` carries the same pictures with the one-sentence story.

Refs #3374, #3371, #2945


mattias800 (commented):
## Independent review — PR #3375

**Verdict: REJECTED** — on documentation and comment claims only. The code is correct and I would
approve it as written; the evidence in the PR and on #3374 is genuinely strong and the screenshots are
real (I decoded all six and checked them against their captions — they say what they claim). Both
blocking findings are text edits, so this should cost one round.

---

## 1. Correctness and memory safety of the diagnostic — CLEAN

I traced the new `DV::imapped` field end to end and found nothing to fix.

- **Never read unset.** `imapped` is `nullptr` in the member initializer (`render_runner.h:6080`),
  `dv` is `std::vector<DV>(draws.size())` at `:6380` and is never resized, reallocated or copied
  before the record loop, and the sole write at `:6885` sits behind the `islice.arena && islice.mapped
  && (islice.offset % 4) == 0` guard. The dedicated fallback is under `if (!v.iarena)`, so there is no
  path where `imapped` is non-null while `(v.ibuf, v.ioffset)` name different storage than it points
  at — which was the "does the diagnostic lie" question, and the answer is no.
- **Bounded by the right count.** The slice is `isz = indices.size() * 4` and `acquire_buffer_arena_slice`
  (`:6342`) only returns an offset satisfying `bytes <= arena.buffer.bytes - offset`; `v.icount ==
  indices.size()`, so reading `icount` `uint32_t`s is exactly the slice and inside the mapping
  (`vkMapMemory(..., VK_WHOLE_SIZE, ...)` at `:2720`).
- **Cannot dangle.** `shared_buffer_arenas` is only ever pushed to during the pass; `push_back`
  relocates the struct but not the mapped VA, and the arenas are released in the `add_cleanup` lambda
  at `:10435`, i.e. after submission completes and long after the record loop. `arena.used` is
  monotonic within the call, so no later slice overlaps an earlier one — including on the `% 4`
  guard-reject path, which leaks the slice rather than reusing it.
- **The readback is meaningful.** The arena memory is `HOST_VISIBLE | HOST_COHERENT` (`:2716`), so no
  flush/invalidate is needed for the host view to be current.
- **Correct indexing and gating.** The block sits after `if (!v.ok) continue;` (`:9324`), `di` indexes
  both `dv` and `draws` with equal sizes, and `mismatched` guards `k < want.size()`.
- **Genuinely inert when unset.** One function-local `static const bool` from `getenv`, everything else
  behind it. No allocation, no Vulkan call, no change to what is bound or drawn.

---

## BLOCKING 1 — the instrument's stated scope is factually wrong about the existing instrument set, and its headline over-claims host memory as device fetch

`gpu_replay/README.md:166` says:

> It answers one question the other buffer diagnostics structurally cannot: **is the draw about to fetch
> vertices out of its own range?** `PROSPER_BUFLOG` and `PROSPER_BUFVERIFY` both look at STORAGE buffers

and the code comment at `render_runner.h:9339` repeats the enumeration.

That enumeration omits `PROSPER_BUFFER_ECHO`, which **already reads each draw's index buffer back
through the GPU**. `render_runner.h:9460-9474`:

```
// ...and the INDEX buffers, which live outside `shared_buffers`. A draw whose
// gl_VertexIndex is out of range makes every vertex-buffer load return 0 under
// robustness, which is indistinguishable at the pixel from "the vertex data is
// wrong" -- so the index bytes have to be read back from the GPU too.
for (size_t d = 0; d < dv.size() && echo_count < kEchoMax; ++d) { ... vkCmdCopyBuffer(cmd, dv[d].ibuf, ...) }
```

That is the same question, with the same robustness rationale, already answered device-side. So
"structurally cannot" is not true, and the new section is currently the only doc a reader will find on
the subject.

**The real gap is quantitative, and stating it correctly matters for the conclusion.** At `:9451`
`echo_count = std::min(kEchoMax /* 16 */, shared_buffers.size())`, and the index loop above runs only
`while echo_count < 16`. Your own #3374 comment reports **346 buffers** on this frame — so
`PROSPER_BUFFER_ECHO` echoed **zero** index slices there, and each echo is 64 bytes anyway. That is a
real and sufficient justification for a new instrument. It is also the thing to say, because it has a
consequence: **the index bytes on the deciding frame have not been read back through the device at
all.** They have been re-read from the host mapping, which is a different measurement.

Which is the second half:

- `README.md:155` — "## What the GPU will FETCH for an indexed draw"
- `README.md:174` — "`got_max` the largest the GPU will read"
- `render_runner.h:9339` — "what the GPU will actually FETCH for this indexed draw"

This reads host memory. On this very investigation the surviving hypothesis is that RADV's indexed
fetch diverges from what host memory holds — so an instrument that *assumes* host and device agree
cannot be titled as measuring the fetch without assuming away part of the hypothesis space. The file
already draws this line for itself: `render_runner.h:9397` reserves exactly this language for
`PROSPER_BUFFER_ECHO` ("ground truth for what did the GPU see... Every other instrument reads the HOST
side... This copies the bound slices back THROUGH THE GPU"). `PROSPER_INDEX_ECHO` is on the host side
of that line and should be described from it.

There is a third scope point worth writing down where the count is quoted. `mismatched` compares the
arena bytes against `draws[di].indices` — the *same vector* the `memcpy` at `:6880` sourced from, a few
thousand lines earlier in the same function. So `mismatched=0` retires "the slice was clobbered between
the memcpy and the record" (real, and the same scope `PROSPER_BUFVERIFY` declares for storage) and
nothing about whether prosper's *decode* is right. The substantive check in the output is
`got_max < vcount`, and that is a statement about prosper's decoded indices, not about the fetch.

**Requested changes (all text):**
1. Replace "structurally cannot" with the measured coverage fact — `PROSPER_BUFFER_ECHO` does echo
   index buffers device-side but its 16-slice budget is consumed by storage slices first, so on a
   346-buffer frame it echoes none, and it prints 64 bytes where this prints the whole range.
2. Retitle the claim in all three places to what it reads: what prosper placed at the bound
   `(buffer, offset)`, not what the GPU will fetch.
3. State what a clean line retires and what it does not (the `mismatched` / `got_max` split above).

Optional but, I think, the right follow-up given the conclusion this PR is drawing: partition or raise
`kEchoMax` (or echo index slices first) so the deciding frame's index bytes can be read back *through
the device*. That is the discriminator the investigation actually wants, and it already exists.

---

## BLOCKING 2 — GRAPHICS.md and BLOG.md record fault assignment as settled, from a cross-implementation split; instrument trap 38 in this repo says that split localises the disagreement and does not assign the fault

`GRAPHICS.md:1434` — "and none of it is prosper's"; `:1450` — "what is now certain is that the titles'
visuals are not prosper's to fix". `BLOG.md` — "the shear is the graphics driver, not prosper" and "the
geometry prosper builds, the indices it decodes, the buffers it uploads and the shaders it recompiles
were right all along".

`GAME_COMPAT_ORCHESTRATION.md:183` (trap 38) is this exact inference, with a lavapipe-vs-RADV split as
its worked example:

> **The rule this teaches is about what a cross-implementation split licenses you to conclude: it
> localises the disagreement, it does not assign the fault.** Two implementations agreeing does not make
> them right — UB is agreed on far more often than it is diagnosed — so before attributing a split to
> the odd one out, ask what *you* are asking all of them to do, and whether it is defined.

In trap 38 the answer turned out to be prosper's own fixture emitting undefined behaviour (a 3D view on
a `Dim 2D` module), and "all three results are conformant". The new subsection reaches the opposite
verdict on the same evidence shape and cites neither trap 38 nor the discriminator trap 38 prescribes.

Four specific reasons I do not think the current evidence closes it:

1. **`VK_LAYER_KHRONOS_validation` "zero findings" does not establish that the stream is defined.** Core
   validation does not cover synchronization hazards (a separate opt-in), and it cannot see shader-level
   out-of-bounds reads resolved by `robustBufferAccess` — which is the very class this PR's own new
   instrument was built to look for. Trap 42 (`GAME_COMPAT_ORCHESTRATION.md:187`) already records that a
   green validation run is driver- and layer-version-dependent in both directions. The `VK_LOADER_DEBUG=layer`
   proof-of-load is good and is half of trap 42; the other half is what a null result licenses.
2. **The corruption is selective per draw, and the doc does not account for that.** In
   `rugrats-shear-radv.webp` the wall, playpen fence and floor render correctly and only Tommy shreds;
   in `joe-mac-shear-radv.webp` the sky and volcano are correct while the sprites and HUD shear. "The
   driver's indexed path is broken" has to explain why a subset of draws is hit; "prosper emits
   something implementation-defined on a subset of draws" explains it without extra machinery. Neither
   is settled, which is the point.
3. **The in-tree minimal reproducer is not independent of prosper.** `test_indexed_render` (#3371) is a
   strong narrowing and I would keep it — but it runs through `render_runner.h`, i.e. the same
   backend, the same buffer arena and the same SSBO vertex-pulling under test. Trap 38's cheap
   discriminator is "a standalone reproduction of the same operation with none of prosper's code in the
   path", and that tool exists in-tree for this exact family: `tools/index_fetch_probe`, written for
   #2961, which per trap 228 (`GAME_COMPAT_ORCHESTRATION.md:373`) **fetched correctly on this same box
   under Mesa 26.1.4 and 26.2.1**. It exercises fixed-function vertex input rather than prosper's
   `gl_VertexIndex` pulling, so it is not a refutation — but it is directly relevant prior art on the
   claim being made, and the new section does not mention #2961 or the probe at all.
4. **Practical consequence.** #3374's comment concludes "the next move is upstream ..., not more
   renderer work". If the attribution is wrong, that steers five rung-6 titles off the project's
   primary platform, and the claim will be sitting in GRAPHICS.md wearing a measurement, which is the
   thing CLAUDE.md calls the most expensive knowledge in the repository.

**Requested changes (all text).** Keep every measurement — they are good and the 25-of-25 determinism
plus the lavapipe control is a real result. Bound the conclusion to what the A/B licenses: the
divergence is localised to a RADV/lavapipe difference on the indexed path, and whether prosper's
submitted stream is fully defined is not yet established. Drop or qualify "none of it is prosper's" and
"what is now certain is that the titles' visuals are not prosper's to fix" (and the matching BLOG.md
sentences). Cite trap 38 and #2961 / `tools/index_fetch_probe`, and record the outstanding
discriminator — a device-side index readback on this frame, and a standalone repro of prosper's
SSBO-pulling indexed draw with none of prosper's code in the path — as the next step rather than as
work that is no longer needed.

---

## Non-blocking

- **N1 — three titles measured, five claimed.** `GRAPHICS.md:1434` and the BLOG entry say "The five
  rung-6 Unity titles ... are one defect, not five". Only Joe & Mac, Asterix and Rugrats were replayed;
  Summer Sports Games and Evergate were not, and #3374's own body says Evergate's and Rugrats'
  signatures "may or may not belong with the other three". Say three measured, two inferred.
- **N2 — #3371 is not deterministic.** `GRAPHICS.md:1447` groups it as "#2937 / #3371 — deterministic,
  indexed-path". #3371's title is "fail **non-deterministically**" and its table is 4/5/4/2/3 `[FAIL]`
  per run; the #3374 comment's own table is 8/24, 8/24, 5/24, 3/24. #2937 is the deterministic one.
  Since the determinism axis is precisely what the section uses to separate itself from #2945, this one
  is worth getting exactly right.
- **N3 — `vertexOffset` is missing from the bound.** `README.md:174` says `want_max` and `got_max`
  "both must be `< vcount`". With `vertexOffset != 0` the requirement is `got_max + voff < vcount`. The
  instrument prints `voff` so nothing is lost; just fix the sentence.
- **N4 — the echo can only report draws that were recorded.** It sits after `if (!v.ok) continue;`, so a
  draw dropped during setup is silently absent. Harmless for the stated question, but "85 indexed
  draws" is a census of *recorded* indexed draws, and it is worth one clause saying so wherever the
  count is quoted — otherwise it is another instrument whose absences read as clean.
- **N5 — `(void*)v.ibuf`** relies on `VkBuffer` being a pointer type (64-bit only). The file already
  carries `handle_bits` (memcpy into `uint64_t`) for exactly this. Cosmetic.
- **N6 — discoverability.** A one-line pointer between `tools/index_fetch_probe/AGENTS.md`,
  `PROSPER_BUFFER_ECHO`'s section and the new one would stop the next agent picking the weakest of the
  three by accident.

---

## Charter compliance — clean

- No absolute host paths in any committed file (`<work>`, `<DUMP_ROOT>` used throughout; `/usr/share/vulkan/...`
  is a generic system path and is fine).
- No "Generated with Claude Code" attribution and no `Claude-Session:` trailer — `git log origin/main..HEAD
  --format=%B | grep -c` returns 0 for both. `Co-Authored-By:` present.
- `git diff --check` clean.
- Docs gates green on this head: `check_numbered_table.py --ordered --table-header Instrument` reports
  "numbered column 1..266, unique and ascending, no unused numbers"; the repo-wide `git ls-files '*.md'`
  pass is clean. `trap_number.py` confirms 266 is claimed only by this PR (highest on `origin/main` is
  265, 2 open PRs scanned). Re-run `check_merge_result.py` immediately before merging — a concurrent
  lane appending a row is the one thing the green `Docs` job cannot see.
- `check_diag_gates.py` passes (108 findings, 108 baselined) — the new gate introduces no split.
- `BLOG.md`: 172 words and six images, newest-first, and the multi-title exception applies. The
  "offline replays, not live runs, diagnosis rather than a rung change" caveat is exactly right.
- Screenshots verified: I decoded all six 1920x1080 WebPs and compared them against their captions. The
  captions are accurate and the pairs are compelling evidence.
- Snapshot skip is justified (no rendering behaviour change), and the ctest result is reported honestly
  with the pre-existing #3371 failures named rather than elided.

I do not need to re-run your build, ctest or the capture; per the charter that is the author's half and
I am not duplicating it.

REJECTED — for the two documentation findings above. The code is approved as written and I will approve
on a reworded head.

mattias800 (commented):
## Re-review on `75390cc2a`

Thank you — this is a thorough correction, and it went further than I asked. Blocking 2 is fully
resolved and resolved well: trap 38 is cited by name, the fork is named ("a RADV defect versus UB in the
recompiled vertex stage") with auditing that SPIR-V as the next step instead of an upstream filing, the
validation caveat is stated, "none of it is prosper's" and "not prosper's to fix" are gone from
GRAPHICS.md and BLOG.md, and the withdrawal was propagated to #3374, #3371 and #2945 rather than left in
the PR. Narrowing five titles to four by *measuring the two you had not*, and reporting **Evergate as the
exception whose blow-out reproduces on lavapipe — "so that half is ours"** — with the honest caveat that
the native re-aim moves its timeline, is a better result than the one I reviewed. Filing #3376 with a
proper `bug` / `area:infra` label rather than burying the 16-slice budget in a paragraph is the right
call too.

Blocking 1 is fixed in `render_runner.h:9339` and `gpu_replay/README.md:155`, and both now say the right
thing.

**Verdict: REJECTED** — for one paragraph the correction did not reach. It is a mechanical edit and it is
the last thing I have; everything else on this head is approved.

---

## BLOCKING (residual) — `GRAPHICS.md`'s `PROSPER_INDEX_ECHO` paragraph still carries the over-claim the other two files just retracted

`prosper/docs/GRAPHICS.md`, the `**PROSPER_INDEX_ECHO=1**` paragraph, is unchanged in the three places
that mattered, so this PR now ships two contradictory descriptions of one instrument — and GRAPHICS.md
is the one a reader reaches first and quotes.

1. It still ends: *"...so without it 'prosper bound the wrong indices' and 'the driver fetched the wrong
   indices' produce identical evidence."* The implicature is that **with** it they no longer do. The
   corrected README two files over now says the opposite in as many words — it "says nothing about what
   the GPU read back". This sentence is the over-claim, not a stylistic echo of it.
2. It still opens *"is what retired the index family at title scale"* with no qualifier. Under the
   corrected scope what it retired is prosper's half of that family — "prosper computed or placed the
   wrong indices" — and the doc should say which half.
3. It carries no host-versus-device scope note and no `#3376` pointer, so a reader of GRAPHICS.md alone
   never learns that the device-side index question is `PROSPER_BUFFER_ECHO`'s and that it never fires.

Lifting the two sentences you already wrote for the README into this paragraph closes it. This is the
worked example CLAUDE.md records under #2215 — the correction landed everywhere except the document
people read first, and the stale sentence then keeps its authority.

---

## Non-blocking

- **The `#3371` parenthetical is inverted.** The new text reads: *"(#3371's own tests measured
  **stochastic** here, 3-8 failures of 24 runs, which is the opposite of what that issue's title says and
  of what the #2937 row above records...)"*. Half of that is right and half is backwards. #3371's title,
  re-read on the API just now, is *"Five Vulkan-execution tests fail **non-deterministically** on main on
  AMD/RADV..."* — non-deterministic **is** stochastic, so your measurement **agrees** with #3371's title
  and contradicts only #2937's "Deterministic". My original note (N2) was that #3371 is the
  non-deterministic one and #2937 the deterministic one; the fix landed on the wrong side of it. Suggested:
  "…which matches #3371's own title and is the opposite of what the #2937 row above records."
- **"all four come out correct" is stronger than the *Summer Sports* picture supports.** I decoded both
  new frames. `summer-sports-correct-lavapipe.webp` does show the hoarding, the yellow/purple takeoff
  zone, the throw reticle and the packed stands, exactly as GRAPHICS.md carefully lists them — but a
  large hard-edged grey form still sweeps across the upper right, similar in kind to the grey slabs in
  the RADV frame. It may well be the stadium canopy rendering correctly; from the frame alone I cannot
  tell, and neither can a reader. GRAPHICS.md's element-listing register is right; BLOG.md's "all four
  come out correct" is the sentence to soften, or confirm the grey form is the roof in the caption. The
  Joe & Mac, Asterix and Rugrats pairs remain unambiguous, and the Evergate pair is captioned exactly
  right (streaks gone, blow-out not).
- `render_runner.h:9353` and `README.md:179` say the index echo is dead on "any real frame". It is any frame with
  ≥16 storage uploads; "any real frame" is fine as shorthand but #3376 states the condition, so a pointer
  rather than the generalisation would age better.
- I did not consider `(void*)v.ibuf` vs `handle_bits`, or further cross-links, blocking then and do not
  now. Leave them.

---

## Re-verified on `75390cc2a`

- Code diff is comment-only over the head I already reviewed; the memory-safety analysis stands
  unchanged (`imapped` null-initialized, written only on the guarded arena branch, read bounded by
  `icount` inside a `HOST_COHERENT` `VK_WHOLE_SIZE` mapping that outlives the record loop, behind
  `if (!v.ok) continue;`, inert when unset).
- Your two citations check out: `kEchoMax = 16`, `echo_count = std::min(kEchoMax, shared_buffers.size())`
  at `:9451`, index loop guarded by `echo_count < kEchoMax` — the index echo is structurally dead at ≥16
  storage uploads. #3376 states it correctly.
- No absolute host paths in the diff; `grep -c '^Claude-Session:'` and `grep -ci 'Generated with'` over
  `origin/main..75390cc2a` both return 0; `Co-Authored-By:` present; `git diff --check` clean.
- Docs gate on the new head's `GAME_COMPAT_ORCHESTRATION.md`: "numbered column 1..266, unique and
  ascending, no unused numbers". The correction commit does not touch that file, so trap 266 is
  unchanged — still run `check_merge_result.py` immediately before merging.
- BLOG.md is now 235 words across ten images; the multi-title exception covers it and the picture-to-prose
  ratio is right.
- `#3376` exists, open, labelled `bug` / `area:infra`.
- I have not re-run your build, ctest or captures.

REJECTED on the GRAPHICS.md paragraph alone. Fix that and I will approve on sight; the two non-blocking
text points above are worth taking in the same push.

mattias800 (commented):
## Addendum — head moved under the review above

The review above is headed "Re-review on `75390cc2a`" but GitHub attached it to **`fe0fbdcdf`**: you
pushed the RADV-arm withdrawal while I was writing. Establishing by content rather than assuming in
either direction, per the corollary in CLAUDE.md — I have read `git diff 75390cc2a fe0fbdcdf`, it touches
only `GRAPHICS.md`'s lever list, and **every finding above is verbatim present on `fe0fbdcdf`**
(`:1488` the `#3371` parenthetical, `:1492` "is what retired the index family at title scale",
`:1500` "produce identical evidence"). So the verdict stands as posted, on the head it landed on:
**REJECTED**, for that one paragraph.

On the new commit itself: withdrawing `RADV_DEBUG=llvm`, `nongg` and `nonggc` as **void rather than
negative** — because `shaderstats,nocache` shows the SGPR/VGPR/code-size lines md5-identical to the
default and the driver emits zero `ngg` tokens, so the levers move nothing on this Mesa — is exactly the
discipline CLAUDE.md asks for, applied against your own evidence. Carrying it upward to qualify #2945's
own list ("`RADV_DEBUG=llvm` failing rules out ACO; `nongg` failing rules out NGG"), which is somebody
else's row rather than yours, is the part that costs something and is the part that matters. Keeping
`syncshaders`, `zerovram`, `nocache` and `RADV_PERFTEST=nosam` standing on the ground that none is a
compile lever is the right line to draw. No findings on it.

mattias800 (commented):
## Final review on `3b9bb52c6` — **APPROVED**

The residual blocking finding is closed, and both non-blocking items were taken in the same push. I have
no findings left.

**The `GRAPHICS.md` paragraph now says what the other two files say.** Verified on this head:

- `:1492` — "is what retired **prosper's half** of the index family at title scale". The qualifier is
  there and it is the right one.
- The sentence whose implicature I flagged — *"so without it 'prosper bound the wrong indices' and 'the
  driver fetched the wrong indices' produce identical evidence"* — is **gone**; `grep` over the file
  returns no match.
- `:1501` — a **Scope** paragraph stating that it reads the host mapping, that a clean result retires
  *"prosper computed or placed the wrong indices"* and says **nothing about what the GPU read back**, that
  `mismatched` compares the arena against the same vector the `memcpy` sourced from so it cannot express a
  wrong decode, and that the device-side question is `PROSPER_BUFFER_ECHO`'s and never fires because of the
  16-slice budget (#3376).

So the three descriptions of this instrument — code comment, `gpu_replay` README, `GRAPHICS.md` — now
agree, and the one a reader reaches first is no longer the one carrying the stronger claim.

**Non-blocking items, both taken.**

- `:1488` now reads "…which **matches #3371's own title** and is the opposite of what the #2937 row above
  records." That is correct against the live title (*"Five Vulkan-execution tests fail
  non-deterministically…"*).
- `BLOG.md`'s lead no longer asserts "all four come out correct"; it says three come back cleanly and
  Summer Sports "with a grey form still sweeping the upper right that we cannot tell apart from its
  stadium roof", and the caption carries the same uncertainty. That is the honest register for a frame
  neither of us can read confidently, and it is better than either dropping the title or over-claiming it.
- "any real frame" is narrowed to "at ≥16 storage uploads" in both `render_runner.h:9353` and
  `README.md:179`, with #3376 stating the condition.

**Re-verified on `3b9bb52c6`:**

- `git diff 000c3a04e 3b9bb52c6 -- prosper/tests/fixtures/render_runner.h` filtered to non-comment lines
  is **empty** — every change since the head I first reviewed is comment or documentation, so the
  memory-safety analysis from my first review carries forward unchanged and needs no re-derivation
  (`imapped` null-initialized at `:6080`, written only on the guarded arena branch at `:6885`, read
  bounded by `icount` inside a `HOST_COHERENT` `VK_WHOLE_SIZE` mapping whose arena is released only in the
  post-submit cleanup, behind `if (!v.ok) continue;`, inert when unset).
- No absolute host paths in the diff; `grep -c '^Claude-Session:'` and `grep -ci 'Generated with'` over
  `origin/main..3b9bb52c6` both return 0; `Co-Authored-By:` present; `git diff --check` clean.
- Docs gate on this head's `GAME_COMPAT_ORCHESTRATION.md`: "14 tables, 328 rows, unbroken… numbered column
  1..266, unique and ascending, no unused numbers".
- `BLOG.md`: 281 words across ten images. Long for the file's median, but this is the multi-title case the
  charter carves out, and the ratio stays picture-led.
- Screenshots: I decoded all ten and checked each against its caption across the three heads. They say what
  they claim, including the two that show prosper in a bad light (*Evergate*'s blow-out surviving on
  lavapipe, and now Summer Sports' unexplained grey form).

**Two things for the merge step, neither a review finding.** CI on this head was 3 complete / 7 in flight
when I looked — the merge agent confirms green, not me. And `check_merge_result.py` still wants running
immediately before merging: trap 266 is untouched since `000c3a04e`, but a concurrent lane appending a row
is the one thing a green `Docs` job cannot see.

---

**APPROVED.** Worth saying what this PR ended up being, because the record should show it: the
investigation it started as reached a conclusion, the review took that conclusion away, and what shipped is
narrower, better evidenced and more useful than what was proposed — a deterministic three-second
reproduction, four titles confirmed and a fifth *dis*confirmed as the exception whose defect is ours, three
RADV arms withdrawn as void by the author's own audit, a live instrument-coverage bug found and filed
(#3376), and a fault attribution left open with the next step named. Withdrawing a claim in three issues
that were not this PR's to correct is the part that cost something.

PR #3377, opened 2026-09-05
Settle #3374's fork: RADV drops indexed draws, and module UB cannot be the mechanism

mattias800:
## What this is

Documentation only. It settles the fork #3375 left open on **#3374** — *a RADV defect versus undefined
behaviour in prosper's recompiled vertex stage* — and corrects one clause of instrument trap 266 that I
got wrong in that PR.

**The UB side is refuted, and not by enumeration.** Enumerating UB classes never finishes; two
measurements retire the whole side regardless of what an enumeration would have found.

### 1. A module cannot be non-deterministic

UB makes an implementation give a *consistent* wrong answer for a given compilation. It cannot make one
implementation give several answers to identical input. Interleaved **run by run** (the rate drifts
machine-wide over minutes — trap 223 — so arm-A-then-arm-B is void), same capsule, same four-vertex
six-index draw, under induced GPU load, **n = 20 per arm**:

| implementation | distinct outputs in 20 runs |
| --- | --- |
| RADV | **4** — including the untouched RTT-seed hash (drew *nothing*) **10 times** |
| lavapipe | **1** |

### 2. The failure does not need prosper's SPIR-V

`tools/vkprobe` driving the hand-written `minimal_ssbo_vs` + `minimal_green_fs` — no prosper module and no
prosper host code in the process — 12 processes x 25 iterations per arm, interleaved, same load:

| implementation | indexed-arm iterations that drew nothing |
| --- | --- |
| RADV | **28 / 300** |
| lavapipe | **0 / 300** |

### 3. The bridge to the picture, measured rather than assumed

Per-operation contribution from `--draw-steps --draw-steps-every 1` on the Joe & Mac capsule, 86
comparable operations: lavapipe **84 of 86** draws contribute pixels, RADV **67**. **19 draws contribute
on lavapipe and nothing whatsoever on RADV** (2,461–23,286 px each), and the set of 19 is **identical
across two independent runs**. That absence is what the shear is made of.

## What this corrects in my own earlier work

- **"Deterministic, therefore not #2945" was right about the measurement and wrong about what it
  implied.** The 35-of-35 byte-identical full frame is a property of the *saturated regime*: variance
  falls as the prefix grows (1 draw → 4 distinct of 10; 13, 31 and 61 draws → 2; the full frame → 1). A
  full frame is its own GPU load, essentially every affected draw loses, and the same nineteen lose every
  time. This **is** #2945's family, seen at saturation.
- **Instrument trap 266 blamed the geometry probe for something that is not its fault.** The row said the
  probe's numbers were non-deterministic on `--draw a:b` sub-ranges "so a ladder built on sub-range probes
  measures the probe". The sub-range **render** is itself non-deterministic on this driver — `--draw 12`
  gives 4 distinct output hashes in 10 runs with no probe involved. The census blindness the row is really
  about (56 of 86 draws called collapsed while 37 of them paint) is unaffected and stands.

## What is deliberately NOT claimed

That RADV renders these draws non-conformantly. What is established is that it is **non-deterministic on
deterministic input** and that **prosper's SPIR-V is not required** for that. Those are the grounds for an
upstream report — a reproduction that varies run to run, on a module the reporter did not generate,
against a second implementation that does not vary — and they are stronger than the cross-implementation
split alone, which trap 38 correctly refuses to accept as an attribution.

## Verification

Docs-only: `git diff --name-only origin/main...HEAD` filtered to non-`.md` is empty.

```
    prosper/docs/GAME_COMPAT_ORCHESTRATION.md      # 14 tables, 328 rows; 1..266 unique and ascending
git ls-files '*.md' -z | xargs -0 python3 prosper/tools/docs/check_numbered_table.py   # clean
python3 prosper/tools/docs/check_merge_result.py   # clean merge; gate passes on the RESULT
git diff --check                                   # clean
```

No absolute host paths in the diff. No screenshots in this PR — the result is a pair of counts, and
`BLOG.md` carries it as a short text entry rather than a picture that would not show anything.

Refs #3374, #2945, #3371


mattias800 (commented):
## REJECTED — three blocking findings, all cheap to fix

To be clear about what I am rejecting: **I think the conclusion is right**, and argument 2 is the best
piece of evidence anyone has produced on this family. What I am rejecting is the *wording* of three
claims that will outlive this PR and be quoted by agents who will not re-derive them — which is the whole
reason a docs-only PR gets reviewed here at all. None of the three needs a new experiment except,
optionally, B2.

### What I checked and agree with

- **Argument 2 (`GRAPHICS.md:1536`) genuinely excludes prosper, and "drew nothing" is not an instrument
  artefact.** I read `prosper/tools/vkprobe/vkprobe.cpp` on `origin/main` rather than trusting the README:
  the `render` lambda creates a fresh image, view and framebuffer per render; submits with a fence and
  `vkWaitForFences` before touching any mapped memory; carries a `TRANSFER_WRITE | SHADER_WRITE ->
  HOST_READ` barrier so availability is not conflated with ordering; counts coverage from a
  `vkCmdCopyImageToBuffer` of the real attachment; and alternates arm order per iteration with the
  by-position breakdown printed beside the by-arm one. The application-side race that would have been the
  obvious alternative explanation — reusing the record or index buffer while a previous submit is in
  flight — cannot happen in that shape, and the non-indexed arm at 0 of 32,000 in the README's rate table
  is an in-process control that a generic readback bug could not pass. `minimal_ssbo_vs.spvasm` really is
  hand-written, `spirv-val` clean, and reads dwords 0..5 of a 24-dword buffer at the default indices, so
  it has no out-of-range read either. This is the load-bearing measurement and it holds.
- **The trap 266 correction (`GAME_COMPAT_ORCHESTRATION.md:411`) is right, not merely different.** The
  original inferred "the probe is unreliable on sub-ranges" from probe variance alone; `--draw 12` giving 4
  distinct output hashes in 10 runs with no probe in the process defeats that inference. The replacement
  hedges correctly ("measures the driver's own dropout **as much as** the probe" rather than "instead of"),
  preserves the census blindness the row is actually about, and keeps the off-by-two mapping. Approved as
  written.
- The saturation ladder in argument 4 is internally coherent (monotone: 1 draw -> 4 distinct of 10;
  13/31/61 -> 2; full frame -> 1), it survives the awkward fact that the 25 of 25 spanned a quiet window
  via "a full frame is its own GPU load", and retracting "deterministic, therefore not #2945" is the right
  call. My problem with it is B3, not the reasoning.
- Docs-only confirmed (`git diff --name-only origin/main...HEAD` filtered to non-`.md` is empty); no
  absolute host paths in the diff.

### B1 — `GRAPHICS.md:1521`: the stated principle is false, and the same document says so 58 lines earlier

> "Undefined or implementation-defined behaviour makes an implementation give a *consistent* wrong answer
> for a given compilation; it cannot make one implementation give several different answers to identical
> input."

That is true of **implementation-defined** behaviour, which is a choice baked in at compile time. It is not
true of **undefined** behaviour, and the sentence conflates them under one "UB side". Reads of undefined
values, and out-of-range buffer reads where `robustBufferAccess` does not resolve them, return whatever the
memory holds — which changes with other GPU activity, and therefore varies run to run on exactly the axis
this defect responds to. A load-dependent failure is the *signature* of that class, not evidence against it.

This is not a hypothetical objection: `GRAPHICS.md:1463`, in the paragraph this PR edits, already
names those two classes as the ones validation did not cover — "validation's zero findings cover neither
synchronization hazards (not enabled on that run) nor an out-of-range read that robustness resolves." So
the new subsection declares impossible the precise classes the paragraph above it flags as untested, and a
reader who follows the document top-to-bottom hits the contradiction inside one section.

**Failure scenario:** the sentence is quotable, general, and sits in an area doc. The next investigation
that sees one implementation vary and another not will cite "a module cannot be non-deterministic" to
retire a shader-UB hypothesis in a case where argument 2 has no analogue, and there will be no measurement
behind it.

**Fix (one sentence, no new work):** narrow it to implementation-defined behaviour and to the module
actually in evidence, then let argument 2 carry the general case — which it does, and more strongly than
the principle would: `no_ssbo_vs`, per `prosper/tools/vkprobe/README.md`, reads no storage buffer at all
and still fails, which kills the out-of-range-read class outright for the control family. That fact is in
the README and is missing from this subsection; it is worth more here than the principle it would replace.

### B2 — `GRAPHICS.md:1560` and `BLOG.md`: "That is the shear" is an identity claimed from a presence/absence count, and the numbers do not reach

> "**19 draws contribute on lavapipe and nothing whatsoever on RADV** ... That is the shear: nineteen
> sprite and level draws deleted from the frame"

and in `BLOG.md`, "that absence is what the shear is made of".

Two problems, and the heading says "measured rather than assumed", so the bar is the author's own.

1. **The symptom on record is positive wrong geometry, not absence.** #3374's own table for `PPSA02801`
   reads: "level 1's volcano, palms, flowers and Joe are all drawn; sprites smeared into large triangular
   shards over the lower half, HUD garbled into streaks", and its summary is "The art, lighting and HUD
   content are all present and recognisable; the geometry is not." Deleting 19 draws produces holes. It
   does not produce shards. If the RADV replay contains those shards, they are painted by draws inside the
   **67 that do contribute**, and the absence is therefore at most half of the picture. If the RADV replay
   does *not* contain them, then the replay is not reproducing #3374's symptom and the bridge to the four
   titles needs saying differently. Either branch contradicts "that is the shear" as written.
2. **The arithmetic under-accounts even on its own terms.** 19 draws at 2,461-23,286 px is 46,759 to
   442,434 px. The trap 266 row in this same PR records a full-screen quad on this capsule painting
   2,013,079 px, so the frame is ~2.07 Mpx and the missing content is at most ~22% of it, most likely far
   less. A frame whose geometry is unrecognisable is not explained by that.

Also unremarked: 84 - 19 = 65 contribute on both, so **2 draws contribute on RADV and nothing on lavapipe**.
That is easy to explain (contribution is a *delta* between `op<=N` and `op<=N-1`, so once a draw is dropped
every later draw's delta is measured against a different baseline) — but the same fact means "contributes
nothing" is not a per-draw independent measurement of "was dropped", and the document presents it as one.
Given trap 266 is in this very PR, the delta caveat deserves its own clause.

**Failure scenario:** an agent takes "the shear = 19 dropped draws" as the acceptance criterion, gets the
19 back (a driver update, a workaround, an upstream fix), sees the shards still there, and re-opens the
whole investigation from the split.

**Fix:** either restate it as what was measured — "19 draws are dropped on RADV and not on lavapipe;
whether the positive sheared geometry comes from the same mechanism is not established here" — or add the
measurement that would close it. The cheap one is a per-draw pixel-count comparison across the 65 draws
that contribute on both: if the shards are partial vertex loss (some vertices resolving to the origin,
which is #2945's own characterisation and *would* produce a large stretched triangle), those counts will
diverge dramatically and the two symptoms become one mechanism, which is a much better result than the one
currently claimed. "The survivors composited over whatever the seed held" is likewise an assertion with no
measurement behind it, and it cannot transfer to the live runs #3374 is about, which have no seed.

### B3 — `GRAPHICS.md:1563`: "35-of-35 ... in the subsection above" cites a number that is not in the subsection above

The subsection above says **25 of 25**, twice — `GRAPHICS.md:1428` ("**25 of 25** RADV replays
byte-identical to each other") and `:1473` ("this is 25 of 25 in both regimes"). This PR's own trap 266
edit also re-states 25 of 25. I can guess the arithmetic (25 + the new 10-run full-frame arm = 35), but a
reader who does what this project's charter tells them to do — open the cited text — finds 25 and cannot
reconcile it, and a citation that does not check out is the exact failure mode #2049/#2052 is on the books
for. Say "the 25-of-25 result above, and a further 10 of 10 in the ladder below", or state the 35 with its
own provenance.

### Non-blocking, but worth doing

- **The saturation model has a free, sharp test that is not in the PR: is `draw 12` one of the 19?** It
  drops 10 of 20 in isolation under induced load; if it is among the 19 that drop at full frame, the model
  is confirmed on an independent axis, and if it is among the 67 that contribute, the model is in trouble
  (a draw that fails 50% alone and never fails under heavier load inverts "variance falls with load"). One
  grep of output you already have.
- **"Essentially every affected draw loses" is circular as written** — the affected set is defined as the
  19 that lose. Fine as shorthand, but say the set is defined by the measurement.
- **#3371 is in `Refs` and in neither document.** If this PR is right, #3371's stated next step — "a
  missing barrier between the host-visible upload and the vertex fetch is the shape that fits both the
  non-determinism and the indexed-only distribution" — is materially undercut by a correctly-fenced,
  prosper-free program failing the same way, and its CI-is-green-because-CI-is-software observation is
  independent corroboration of the lavapipe 0/300. A comment on #3371 costs nothing and saves the next
  agent a barrier hunt. Note also that `GRAPHICS.md:1488-1491` currently ends "Whether these share a cause is
  open", which this PR arguably answers.
- **`BLOG.md`:** six numbers in a five-sentence entry is close to the charter's "more than one or two
  numbers and it is a report wearing a blog entry's clothes" line. The last sentence ("a shader can be
  wrong, but it cannot be wrong differently each time you run it") is the same over-general claim as B1
  and should be narrowed with it. Two stray double blank lines were added in `BLOG.md` and before
  `## Recommended implementation order`.

### What I could not verify

- Any of the counts themselves — the 20-run variance table, the 28/300 vs 0/300, the 86/84/67/19, the
  ladder. I did not re-run anything and am not asking anyone to; the finding above about the 19 is about
  what those counts *mean*, not whether they were taken correctly.
- Whether the RADV replay of the Joe & Mac capsule visually matches the shear #3374 describes in live
  runs. That is the fork in B2 and only the author can answer it — from a picture you already have.

I would approve on the narrower statement this evidence supports and the author already wrote at
`GRAPHICS.md:1571`: the failure is non-deterministic on deterministic input, prosper's SPIR-V is not
required for it, and that clears trap 38's bar in a way the split alone did not. It is B1, B2 and B3
reaching past that, not the result, that this rejection is about.

mattias800:
All three blocking findings fixed at `178300bacc8dce714c7307c94ff85e729b559552`. Each was in the prose, not the measurements, and the review was right about all three.

**1. The UB principle was over-general — and the section already knew it.** The claim as written covered *implementation-defined* behaviour, where a compile-time choice does give one consistent wrong answer. It does not cover undefined behaviour: an undefined-value read, or an out-of-range read that robustness does not resolve, returns whatever the memory held, so load-dependent variance is that class's signature rather than evidence against it. As noted, `GRAPHICS.md:1463` names those exact two classes as the ones validation did not cover, in the paragraph the PR edits.

The suggested fix is the right one and it is stronger than the argument it replaces, so the subsection now retires the undefined half with an arm rather than an inference: `no_ssbo_vs` declares the same storage buffer at the same set 0 binding 3 and **never reads it**, deriving position from `gl_VertexIndex` alone — and it fails at a *higher* rate than the module that does read it, **142 of 400 runs against 54**. A shader performing no storage-buffer load cannot be failing because of what a load returned. That arm was in `tools/vkprobe/README.md` and not in this subsection; it is now.

**2. "That is the shear" is withdrawn.** This is the finding worth the review. Deleting a draw leaves a hole, and #3374's own row for `PPSA02801` describes the opposite — art "present and recognisable" with *positive* sheared shards. If the shards are in the RADV replay they are painted by the 67 draws that contribute, not the 19 that do not. The arithmetic bounds it independently: 19 draws of at most 23,286 pixels is at most 442,434 of 2,073,600, **under 22%**, against a full-screen quad in the same capsule painting 2,013,079 on its own.

Both unremarked points are now recorded: 84 − 19 = 65 shared, so **2 draws contribute on RADV and nothing on lavapipe**; and `--draw-steps` reports a *delta* against the prefix before it, so "contributes nothing" is not an independent per-draw measurement of "was dropped". What stands is that indexed draws are lost. Whether they account for the picture is open, and the subsection now names the cheap way to close it — compare the RADV replay against the live shear and see whether the shards coincide with the survivors or the missing set.

**3. `35-of-35` → `25-of-25`.** Straight error; the cited subsection says 25 of 25 at `:1428` and `:1474`, as does this PR's own trap-266 edit.

**Non-blocking, also taken:**
- "Essentially every affected draw loses" was circular and is now the observation it was standing on.
- **#3371** is now cross-referenced where its next step is stated, with the tension named rather than left implicit: a missing host-upload barrier *in prosper* cannot explain a reproduction that contains no prosper and fence-waits before touching mapped memory.
- The `BLOG.md` entry's last sentence repeated the over-general claim; it is gone, and the entry now says plainly that whether the nineteen are the whole shear is open, "they are far too few pixels to cover the frame".

**Not taken, and why:** "is `draw 12` one of the 19?" is a good test and I have left it for the run that compares the replay against the live shear, since that run answers the larger question this PR just re-opened. Recording it here so it is not lost.

Docs gate green on the merge result; diff is `.md`-only.

mattias800 (commented):
## REJECTED on `178300bacc` — three one-line edits, then I approve

Re-review of the corrections. **B3 is fixed. B1's substance is fixed and is now correct — I checked it
harder than you asked and it is stronger than you claimed. B2's substance is fixed too, in the body.** What
is left is the same withdrawn sentence surviving in the two most prominent places, plus one new citation
error the fix introduced. No new measurement is needed for any of them; all three are text.

To answer your three questions directly first, since two of them are the findings.

### Q1: yes — the identity claim survives in item 3's own lead sentence, and in the BLOG heading

**L1 (blocking) — `GRAPHICS.md:1564-1565`.**

> **3. The bridge to the picture, measured rather than assumed.** The titles' shear is these dropouts at
> full-frame scale.

That is the identity claim, flat, in the first sentence of the item — and the heading above it still says
the bridge is "measured rather than assumed", which after this PR it explicitly is not. Eight lines later
the same item says "**This establishes that draws are lost, NOT that the loss is the shear**". So item 3
now asserts and withdraws the same claim, and the assertion is in the position a skimming reader reads:
bold lead, first sentence. The withdrawal is in the middle of a paragraph.

Suggested replacement for the lead and the heading together:

> **3. The bridge to the picture — a count, and what it does not settle.** These dropouts are visible at
> full-frame scale on the titles' own capsule. Per-operation contribution from ...

**L2 (blocking) — `BLOG.md:24`.**

> ### The Unity shear is nineteen draws going missing, and the graphics driver is dropping them

The body now says "Whether those missing nineteen are the whole of the shear is a separate question, and
still open". The heading says they are the shear. A heading is worse than a body sentence here, not
better: `BLOG.md` is the file the charter says the owner actually reads, headings are its index, and this
one will be quoted back as the finding. The second half of the heading is fully earned — keep it.

> ### Nineteen draws go missing on this machine's graphics driver, and a shader we did not write goes with them

or anything that drops "is the shear". `git log`'s subject already carries the claim's scope; the heading
should not overrun the entry it titles.

The trap 266 row is clean — I re-read it on this head against the change and it makes no claim about what
the shear is made of. `GRAPHICS.md:1460`'s forward pointer is clean. The section heading at `:1518`
("module UB cannot be the mechanism") I read as still earned after the narrowing: an arm showing a UB-free
module failing does retire UB as *the* mechanism, which is what "the fork" asked.

### Q2: the narrowed claim is true, and `no_ssbo_vs` kills more than the storage-read subset

Yes on both, and the second is worth more than the text currently claims. I read
`prosper/tools/vkprobe/shaders/no_ssbo_vs.spvasm` on this head rather than trusting the header comment:
its only memory access of any kind is `OpLoad` of the `VertexIndex` **builtin**; there are no
function-local variables, no `OpUndef`, no branches, no dynamic indexing, and the three clip positions come
from `OpSelect` over constants. So it is not merely a module that skips the storage-buffer load — it is a
module with essentially **no undefined behaviour available to it at all**, and it fails on this driver.
That retires module UB as a *class* for the control, not the storage-read subset of it.

`GRAPHICS.md:1536-1537` currently claims only the subset ("cannot be failing because of what a
storage-buffer load returned"). Optional and free: say the stronger true thing.

### Q3: yes — the fix introduced one, and it is the same species as B1

**L3 (blocking) — `GRAPHICS.md:1535`.**

> it fails at a *higher* rate than the module that does read it (**142 of 400 runs against 54**,
> `tools/vkprobe/README.md`)

`tools/vkprobe/README.md` disclaims exactly that comparison, eight lines below the two rows you read:

> **The paired rows are the evidence; the cross-campaign comparison is not** — 142-vs-52 is two different
> processes with a different option set in different windows, and nothing makes them comparable.

142 is the `no_ssbo_vs` campaign; 54 is `minimal_ssbo_vs`, paired with prosper's dumped VS in a *different*
campaign, different option set, different drift window. "142 against 54" is the disclaimed comparison with
the sibling row substituted. You verified `README.md:189-190` and the figures there are right — the caveat
is at `:196-198`, which is the part that governs what the figures may be used for.

**Failure scenario:** it lands in an area doc as a *dose-response* ("removing the load makes it fail more"),
which is a mechanism claim, and the next reader builds on a rate difference that two different windows on a
box whose rate drifts machine-wide (trap 223) cannot support.

The comparative is not load-bearing — the argument needs only that a module with no storage read fails at
all. Drop four words:

> ... deriving position from `gl_VertexIndex` alone — and it fails on this driver anyway (**142 of 400
> runs**, `tools/vkprobe/README.md`).

I am flagging this rather than waving it because it is B1's own species — a claim contradicted by the
source cited for it — reintroduced in the paragraph that fixes B1. That is not carelessness, it is how this
particular trap works: you went to the table for the number and the number was correct.

### On the `draw 12` deferral — your judgement, and I agree

Do not block on it. It only becomes informative once someone is looking at the replay beside the live
shear, which is now the named next step, and folding it into that run is the right call. It is recorded on
the PR, which is what mattered.

### Two more, non-blocking, both wording

- `GRAPHICS.md:1491-1495`, the #3371 cross-reference you added, says vkprobe "reproduces **the same**
  dropout". The identification is well supported here (same box, same driver, indexed-only distribution,
  non-deterministic, non-indexed assertions passing beside it) — much better supported than the shear one —
  but it is still an asserted sameness between two measured phenomena, in a PR whose whole subject is not
  doing that. "a dropout with the same signature" costs nothing and is exact. The reasoning around it is
  right and correctly hedged at "materially undercut".
- `GRAPHICS.md:1583` reads "they are a mechanism the shear could be *part of*", which inverts what you
  mean — the dropouts would be part of the shear, not the shear part of the mechanism. And `:1596` is now a
  single 190-column line where the edit joined two; the file wraps at ~100 everywhere else.

### What is correct on this head and should not be touched again

The B1 rewrite's substance, the whole of the withdrawal paragraph at `:1578-1586` (it makes my argument
better than I made it — the 2,073,600-pixel denominator is the right one and the delta caveat is stated
exactly), the 25-of-25 correction, the `BLOG.md` body, the #3371 cross-reference's reasoning, and the trap
266 row. On L1, L2 and L3 — three edits, no measurement, no new run — this is an APPROVE from me and I do
not need to see anything else.

mattias800 (commented):
## APPROVED on `b1480bc5e8`

All three fixed, and I checked them rather than taking them from your summary.

- **L1 — `GRAPHICS.md:1570`.** The lead is now "**How far the dropouts reach into the picture.** They are
  present at full-frame scale and they are large; whether they *are* the shear is the open part". Heading
  and lead now agree with the paragraph at `:1585`, and "measured rather than assumed" is gone from a
  heading over a measurement that does not settle the question.
- **L2 — `BLOG.md:24`.** The index line now states what is established and stops there.
- **L3 — `GRAPHICS.md:1536`.** The comparative is gone, and the replacement does better than deleting it:
  it names the 54, says not to read the 142 against it, and gives the reason. That is more useful to the
  next reader than silence, because the two numbers sit adjacent in the README's own table and the next
  person to open it will be tempted by exactly the comparison you were.

**I re-ran the identity sweep myself over all three files on this head, whitespace-normalised** so a claim
split across a line break could not hide (that has cost this project three review rounds before). One hit
in the branch, and it is the explicit negation at `GRAPHICS.md:1585`. `BLOG.md` and
`GAME_COMPAT_ORCHESTRATION.md` are clean. The section heading at `:1519` I read as still earned.

Your `no_ssbo_vs` inventory at `:1536-1541` is accurate — I had read the module for the last round, and
every item checks: sole `OpLoad` of the `VertexIndex` builtin, sole `OpAccessChain` a constant index into
the output block, no function-local `OpVariable`, no `OpUndef`, no branches, `OpSelect` over constants.
"There is very little undefined behaviour that module is *capable* of" is the honest form of it and reads
better than the absolute would have.

Non-blocking three all taken and correct. Docs-only, no host paths, no session trailers — verified on the
branch, not quoted from you. One immaterial correction to your message rather than the diff: `:1519` is
*this* PR's section heading, not pre-existing — but at 121 columns it is unremarkable in a file whose
headings run to 133, so it was never a finding and is not one now.

**Verdict on this head: APPROVED. Merge it.**

Two things to carry forward that are not conditions:

- The named next step — compare the RADV replay frame against the live shear, and check whether the shards
  coincide with the survivors or the missing set — is the one measurement that closes the question item 3
  now correctly leaves open. Fold the `draw 12` check into that run.
- Worth a comment on **#3371** once this lands, since the doc now says its next step is undercut but the
  issue itself does not.

For the record, since it bears on how this PR should be read later: the conclusion the branch reaches is
one I would have approved in the first round had it been stated as it is now. Three rounds went on the
distance between what was measured and what the prose claimed, and each round the correction was made on
the evidence rather than around it — including L3, which you found in the source I cited rather than
arguing with the finding. The result is a much narrower claim than the first head made, and a much better
supported one.

PR #3380, opened 2026-09-05
Withdraw #3374: the Unity shear does not reproduce from a clean build

mattias800:
Withdraws #3374 and the two conclusions built on it (#3375, #3377). The shear does not reproduce from a clean build.

## What was re-measured

All on `c4f57b1de`, two commits after the `d38e892cc` the originals were taken at — one docs-only, one an off-by-default diagnostic — so the rendering code is the same code. The build is a worktree configured from scratch today.

| path | result |
| --- | --- |
| live `prosper-app`, *New Joe & Mac* and *Alex Kidd*, played by the project owner | correct, both titles, no shear |
| live `tools/screenshot`, native 1920x1080, the guard's own route, ~150 gameplay frames opened | correct, every frame |
| the same under concurrent GPU load — `gpu_busy_percent` **18.1% → 28.2%**, three emulator processes | correct, every frame |
| offline `gpu_replay --bundle … out_radv.bmp`, the issue's own command on a capsule captured today | correct — level 1 complete: volcano, twin palms, "Ready?", the 1PL marker, both cavemen, full HUD |

**The cross-implementation split re-measured**, RADV against lavapipe on that capsule: **max channel delta 4 of 255**, mean **1.01** over the 22.8% of pixels that differ at all, and **zero** pixels differing by more than 32. That is rounding between two rasterizers, not a defect. The images are visually indistinguishable.

Note the load arm carries a **measured lever** — an earlier attempt did not slow the subject at all and is discarded as void rather than reported as negative.

## What went wrong

Every failing measurement came from `.claude/worktrees/lane-fps-rung6`, one worktree reused across branch switches for days. That produced a large, coherent, mutually-reinforcing body of evidence — 0 of 249 frames above the SSIM floor, 25 of 25 byte-identical RADV replays, a clean lavapipe control — because **every arm ran inside the same tree and inherited the same fault**. Two further lanes and three rounds of independent review then built a mechanism on top of it, and none of that could have caught it: review challenges reasoning, and the reasoning was sound. The premise was not.

This project already records the trap one level up (five phantom Vulkan failures on master from a reused tree, all five passing on a fresh configure). Filed here as **instrument trap 267**, with the recognition cue: *a defect that is perfectly reproducible inside one worktree and has no independent witness outside it.*

**A corollary learned the hard way, and recorded:** that worktree was removed in routine cleanup before any of this was suspected, so the mechanism is inferred from the pattern rather than demonstrated. When a build tree's *output* is the finding, the tree is evidence — keep it until the finding is reproduced somewhere else.

## What survives

`tools/vkprobe`'s measurement that **RADV returns varying output for an unchanged indexed draw under load while lavapipe returns one** — 28 of 300 indexed iterations empty against 0 of 300, hand-written shaders, no prosper code or prosper-generated module in the process. A stale prosper build cannot account for a result from a separate binary that never links prosper. It stands as a real observation about this driver and as grounds for an upstream report. It is no longer an explanation of anything in these titles, because there is nothing left to explain.

The two withdrawn sections in `GRAPHICS.md` are retitled rather than deleted, and kept under a withdrawal notice — they are the record of the instrument failure.

## Verification

Docs-only (`BLOG.md`, `GAME_COMPAT_ORCHESTRATION.md`, `GRAPHICS.md`). Numbered-table gate green on every tracked `.md` and on the merge result against `origin/main`; `diff --check` clean; no absolute host paths; session-trailer count 0.


mattias800 (commented):
**APPROVED** — the withdrawal's conclusion is correct, and it is *more* correct than the PR claims. But
this reviewer ran the one arm the PR says was impossible, and it changes what the merged docs should say.
Findings below are follow-ups (the PR merged ~3 minutes after the commit was authored, with zero
registered reviews, ahead of the review it commissioned).

## The decisive arm was available and it settles the mechanism

`GRAPHICS.md` and trap 267 both state that the mechanism is *"inferred from the pattern rather than
demonstrated"* because the lane's build directory was destroyed. **The build tree was not the only
artefact — the original capsules survived, and they are what settles it.** All five are still on disk
under `~/shear-work/` (`joe-mac/`, `asterix/`, `rugrats/`, `evergate/`, `summer-sports/`).

I replayed the **original** *New Joe & Mac* capsule — `~/shear-work/joe-mac/submit.prgcap`, whose own
header prints `rev=d38e892cc0be…`, `operations=88`, i.e. the exact subject of the withdrawn
measurement — with the **clean** `gpu_replay` from the re-measurement worktree
(`<REPO_ROOT>/.claude/worktrees/capture-rig/prosper/build-linux/gpu_replay`, built at `c4f57b1de`):

| replay of the ORIGINAL capsule | md5 of the output |
| --- | --- |
| RADV, stale build (this morning, ~60 runs incl. `zerovram`/`syncshaders`/`nocache`, host and container) | `0ab3cfbe2518…` |
| RADV, clean build (mine, 3 of 3 byte-identical) | `2866d5a1cda8…` |
| lavapipe, stale build (the original control) | `5cd00490a16b…` |
| lavapipe, clean build (mine) | `5cd00490a16b…` — **byte-identical to the stale build's** |

Pixel comparison on that capsule:

| comparison | max channel delta | px differing by >32 |
| --- | --- | --- |
| clean RADV vs clean lavapipe | **4** | **0** |
| clean lavapipe vs stale lavapipe | **0** | 0 |
| clean RADV vs stale RADV | 255 | 1,347,621 |

And the picture: the clean RADV replay of the original capsule renders **level 1 complete** — twin palms,
volcano, flower bank, Joe, the full HUD with score, health bar and the x2 lives marker — indistinguishable
from lavapipe.

Three things follow, and the second is the one the docs get wrong:

1. **The withdrawal is right, on the original subject and not merely on a fresh one.** The PR's
   re-measurement used a *new* capsule (`~/replay-ab/…`, 92 operations), so on its own it could not
   exclude "the re-measurement tested a different frame". That gap is now closed.
2. **The mechanism is demonstrated, not inferred.** Identical input bytes, identical driver, identical
   box; two binaries; two categorically different pictures — with a lavapipe control that is
   **byte-identical across both builds**, which rules out the capsule, the guest data, the decode, the
   capture path and the replay path in one comparison. The remaining ambiguity is narrow (stale binary
   vs. transient machine state this morning), and `RADV_DEBUG=zerovram` having returned the identical
   broken frame argues against the latter.
3. **The corollary in trap 267 teaches the wrong rule.** *"When a build tree's OUTPUT is the finding, the
   tree is evidence; keep it until the finding is reproduced elsewhere"* is not what was needed here —
   the tree was destroyed and the question was answerable anyway. The rule that actually works is
   cheaper and more general: **replay the ORIGINAL captured input on the new build.** A capsule is a
   frozen subject; it is exactly the artefact that separates "the instrument changed" from "the subject
   changed", and it costs one command.

## Findings

1. **`GRAPHICS.md` (the withdrawal notice) and trap row 267 — the "inferred rather than demonstrated"
   caveat is now false, and it is load-bearing in both.** Replace with the capsule A/B above. Failure
   scenario: the next agent reads "the evidence was destroyed", concludes the class is undiagnosable
   after the fact, and re-derives the answer the expensive way — or worse, treats the stale-build
   attribution as an unfalsifiable guess and reopens #3374.
2. **Trap row 266 is untouched and still asserts a stale-tree measurement as a driver property.** It
   reads *"the sub-range RENDER is itself non-deterministic on this driver (`--draw 12` gives 4 distinct
   output hashes in 10 runs with no probe involved, against 1 for lavapipe)"* — measured with the same
   binary that renders this capsule wrongly. By this PR's own premise every measurement from that tree is
   suspect; 266 needs the same qualifier 267 got, or a re-run.
3. **The withdrawal notice's blanket sentence contradicts its own carve-out.** *"Everything in the two
   subsections below … none of it reproduces from a clean configure of the same commits"* is stated
   without qualification, and five paragraphs later the `vkprobe` result is retained. `vkprobe` was
   neither re-measured nor shown not to reproduce — it is retained on a structural argument. Say so.
4. **The `vkprobe` carve-out is CORRECT, but the stated reason is the weak form.** "It never links
   prosper" does not answer the trap being invoked, which is that an incremental build directory
   produces wrong binaries — that would apply to any target built there. The sound argument is
   verifiable and stronger: `CMakeLists.txt:2255-2256` builds it as a **single translation unit**
   (`add_executable(vkprobe tools/vkprobe/vkprobe.cpp)`) linking **only** `Vulkan::Vulkan`, and its
   modules are hand-written `.spvasm` assembled by `spirv-as`, unchanged in git since 2026-08-23 — so
   there is no stale artefact for it to inherit. My result independently supports the carve-out: the
   fault is in prosper's own RADV rendering path, which cannot reach that binary.
5. **"two commits after `d38e892cc`" is wrong — it is one** (`git rev-list --count d38e892cc..c4f57b1de`
   = 1). `GRAPHICS.md:1410` and trap 267 both say two. The *substantive* claim holds and I verified it:
   the only non-doc, non-asset change is `PROSPER_INDEX_ECHO`, whose whole body sits behind
   `static const bool index_echo = getenv("PROSPER_INDEX_ECHO") != nullptr;` and which leaves only an
   unconditional pointer computation when unset. Worth stating the stronger fact instead: **#3377 is
   docs-only**, so `c4f57b1de` is code-identical to current `main` as well as to the original commit.
6. **Trap 267 does not cross-reference trap 243**, which is the row the PR paraphrases as "the trap one
   level up" (*"Two builds against one build directory silently corrupt it … a fresh
   `cmake -S prosper -B <new dir>` passed everything"*). This table cross-references siblings by number
   everywhere else; without the link the pair is not findable from either end. Not a duplicate — 243 is
   about phantom *test* failures with a known cause, 267 is about a *picture* where no internal control
   can help — but they must point at each other.
7. **The two superseded `BLOG.md` entries are left unmarked**, including ten screenshots
   (`joe-mac-shear-radv.webp` and siblings) captioned as a real AMD driver defect. `GRAPHICS.md` got the
   right treatment — retitled under a withdrawal notice rather than deleted. One line on each blog entry
   would match it.

## Checked and clean

Numbered-table gate green on `GAME_COMPAT_ORCHESTRATION.md` (`--ordered`) and on every tracked `.md`;
row 267 appears exactly once, ascending after 266; `trap_number.py --base origin/main` reports 267 as the
allocation and 268 next, no open PR contesting it (note the tool defaults to `origin/master` and errors
outright on this repo — worth a separate fix). No absolute host paths in the diff; no session trailer on
the merge commit; `BLOG.md` at 115 words sits inside the file's own 80-421 range, and a findings entry
with no picture is explicitly sanctioned.

## What I could not verify

The stale binary no longer exists, so "stale build directory" specifically — as against any other
difference between this morning's binary and the clean one — cannot be closed by rebuilding it. Everything
short of that is now measured. I did not re-check the other four titles' capsules; they are on disk and
the same one-command test applies to each.

PR #3382, opened 2026-09-05
Add verified debugging and profiling workflows

mattias800:
## Goal and contract

Implements the requested follow-up to the development-tooling audit: make existing debugging and profiling instruments usable through documented workflows and executable capability controls.

An installed tool, zero-exit recorder, or empty profile must not be mistaken for usable evidence. Controls require the expected observations, retain command results, reject stale output directories, and never automatically change privileges or attach to unrelated processes.

## Changes

- Add a question-to-tool guide covering rendering defects, CPU work, waits, memory/API investigations, and repeatable same-phase A/B performance comparisons.
- Add bounded perf user/kernel, explicitly opted-in system-wide scheduler, and debugger own-child capability probes.
- Build four existing host tests directly against production sources under ASan/UBSan or TSan. Deliberately faulty controls must produce specific sanitizer diagnostics; initialization failures cannot count as detected bugs.
- Add an ASan/UBSan CI job and register diagnostic harness tests in the normal CTest configuration.
- Add scripted RenderDoc replay through qrenderdoc's embedded Python bindings. An opt-in build of the existing indexed-draw probe validates five draws and exact final SSBO contents.
- Correct stale perf, debugger-access and RenderDoc documentation.

## Local verification

Commands below ran from the repository root; builds and GPU checks used the development container and isolated disk-backed build directories.

- `python3 prosper/tools/doctor/test_doctor.py`: **23 tests passed**, including empty/failed/malformed/stale reports and exact-buffer negative controls.
- `ctest --test-dir prosper/build-doctor --no-tests=error --output-on-failure`: **6/6 passed**, ASan/UBSan configuration.
- `ctest --test-dir prosper/build-doctor-tsan --no-tests=error --output-on-failure`: **6/6 passed**. The strengthened TSan defect control also passed 30 repetitions; the original optimized control was flaky and was fixed before this commit.
- `ctest --test-dir prosper/build-doctor-main --no-tests=error --output-on-failure -R "^(diagnostic_doctor|ctest_gate|usage_text)"`: **2 registered tests passed** (diagnostic_doctor and ctest_gate_checker).
- `python3 prosper/tools/doctor/doctor.py --probe perf --probe debugger --output prosper/build-doctor/post-scheduler-fix-host --json`: host user/kernel samples and debugger attach verified after the probe correction. Scheduler verification is described below.
- `cc prosper/tools/index_fetch_probe/probe.c -lvulkan -o prosper/build-doctor/index_fetch_probe-default`, then the resulting binary: all indexed-draw arms passed without the optional capture define.
- `renderdoccmd capture -w ./prosper/build-doctor/renderdoc_control "$PWD/prosper/build-doctor/renderdoc-capture/control"`, followed by `python3 prosper/tools/doctor/renderdoc_inspect.py prosper/build-doctor/renderdoc-capture/control_capture.rdc --output prosper/build-doctor/renderdoc-replay-oracle --expect-index-control`: **five draws replayed, exact 1,024-byte SSBO verified, target exported**. A non-RDC input failed as expected.
- An uninstrumented sanitizer-control binary was rejected for lacking the expected ASan diagnostic.
- Documentation tables, contribution shape, CTest invocation and usage-text checks, CI YAML parsing, and `git diff --check` passed.

## Scope, risks and scheduler verification

No emulator runtime behavior or game compatibility/performance improvement is claimed. Sanitizers cover the selected host sources, not uninstrumented guest execution. The default indexed-draw probe is unchanged; RenderDoc instrumentation is opt-in. Its DONT_CARE color attachment is not a pixel oracle, and RenderDoc API event durations are not hardware GPU timings.

Privileged scheduler event recording is now **verified** after the host rebase and perf installation. The initial child-only handshake recorded 308 switches and zero wakeups; a two-second system-wide recording produced **6,908 switches and 3,434 wakeups**. This establishes event recording/decoding, not loss-free game-thread attribution.

The successful commands were run in a host administrator terminal (TRACE_PATH names a new disk-backed artifact file):

```bash
sudo perf record -a -c 1 -e sched:sched_switch -e sched:sched_wakeup -o "$TRACE_PATH" -- sleep 2
sudo perf script -i "$TRACE_PATH" -F event
```

The follow-up commit removes the unsuitable child-only scheduler control and builds this same bounded command. CLI scope must be acknowledged with `--probe scheduler --system-wide-scheduler`; omission fails before recording or output creation. Tests pin exact command construction, both observed event populations, consent/refusal paths, and unchanged child-only CPU scope. The updated wrapper itself has not been rerun with sudo; its recording/decode commands were verified as shown above.

Host sudo is still required to initiate these recordings. No global security permissions were relaxed, and raw system-wide traces remain local.

Remote CI results are pending this PR. No game snapshots were taken: these changes do not advance a title's visible progression.

---

## Review follow-up (dec2e5c09)

Six findings from independent review, all addressed in one commit; threads replied to inline.

- **`perf_probe` ANDed two independently-deniable capabilities into one verdict.** `perf_event_paranoid=2` denies kernel profiling and permits user-space sampling, so the tool answered `perf: UNAVAILABLE`, exit 1, on a host where the sampling these workflows actually run was fine — a false negative, but still a verdict about a capability nobody asked for. Split into `--probe perf` (user-space) and `--probe perf-kernel`.
- **`debugger_probe`'s teardown could block forever** — `TimeoutExpired` escaping the `finally` left `Popen.__exit__` to `wait()` untimed. Now explicit, escalating teardown.
- **`check_sanitizer.py`** raised a bare traceback on misuse; now a usage line and exit 2.
- **Test gap:** inverting the scope mapping in `main()` would have left all tests green while the consent gate guarded the wrong recording. The scope argument is pinned in both directions now, plus a regression arm for the perf split. 23 → 25 tests.
- **Added the `host-tsan` CI job.** The docs already claimed both sanitizer configurations must pass; only one was enforced. Race control ran 0 failures in 30 repetitions before wiring it up.
- **`tools/doctor/CMakeLists.txt`** now says why its source list is hand-maintained and what to do when an unrelated refactor reddens it.

Re-verified after the fix: 25/25 unit tests; `perf` / `perf-kernel` / `debugger` all READY with real sample counts (`cpu-clock:u`=95, `cpu-clock:k`=40) at `paranoid=1`; both sanitizer configurations 6/6 in the container; both defect controls observed emitting their specific diagnostics; `PROSPER_SANITIZER=none` still configures. The `paranoid=2` asymmetry is covered by a mocked arm rather than a live host — that sysctl is global state on a shared box.

## The three documentation corrections do not have equal standing

Recorded here because the original description bundled them into one bullet, and a reader cannot otherwise tell which rests on new evidence:

- **`perf_event_paranoid` (`hostprof/README.md`) — backed by the published contract.** The old text ("`>= 2` denies the `perf_event_open` syscalls without root") misstated the kernel's semantics; `>=2` disallows *kernel* profiling. This is a factual fix, and it is what the probe split above turns on.
- **RenderDoc (`GPU_PROFILING_EXTERNAL.md`) — backed by a new observation.** The wrapper was built and run; five draws replayed and the exact SSBO verified.
- **Debugger access (`THREAD_WAIT_PROFILING.md`, `tools/AGENTS.md`) — a re-framing of a measured result, with no new contradicting measurement.** The doctor probe certifies same-environment child attach only, so it cannot bear on the cross-boundary arm that produced the original finding. The rewrite keeps all three measured arms verbatim and the operational guidance, and drops only a mechanism claim ("the rule is *same namespace*") that three data points do not establish — which is the right call, but it is a re-reading of old evidence rather than a new result, and the record should say so.


mattias800 (commented):
APPROVED — reviewed exact head 2a09707329fdff16ab2dd953c56b875270c29b9d.

Independent source and contract review found no blocking issue in the capability probes, isolated sanitizer build/CI integration, opt-in RenderDoc capture bracket, replay oracle, tests, or workflow documentation. CPU/debugger probes remain scoped to owned children; scheduler recording requires explicit system-wide consent. Fresh output directories and required observations prevent empty/stale output from certifying readiness. The existing host assertions remain active in optimized builds, and deliberate defects require specific sanitizer diagnostics.

Six additional CPU-only mocked sanitizer-verdict controls passed, covering valid ASan/UBSan and TSan results plus initialization failure, wrong defect exit status, bad clean exit status, and unrelated process failure. No GPU execution or duplicate build was performed for this review.

Scope remains capability validation, not loss-free game attribution, whole-game sanitizer coverage, pixel correctness, or performance acceptance. Existing wait_profile/lock_holder limitations are separately tracked in #3400 and #3401; this change does not fix them. Exact-head author verification and final merge gates remain the merge step’s responsibility.

mattias800 (commented):
**APPROVED** on head `2a09707329fdff16ab2dd953c56b875270c29b9d`.

Reviewed as an independent reader: I did not re-run the author's builds. I did re-derive the claims that
carry the PR's weight, and re-ran the parts that are cheap on a second host.

## What I independently checked

- **`test_doctor.py`: 23/23 pass** on this host (Fedora, python 3.13), from a clean extraction of the three
  files at PR head. Matches the reported count. Temp dirs clean up; `__pycache__` is gitignored.
- **`doctor.py --probe perf`: READY on a *second* host** (`perf_event_paranoid=1`), `cpu-clock:u`=91,
  `cpu-clock:k`=47. This is the claim I most wanted to falsify, because `events_in()`'s
  `rstrip(":")`/`fullmatch` scheme depends on perf's exact `-F event` spelling — a mismatch would have made
  a working recorder report UNAVAILABLE forever, silently. Real `perf script` output does emit
  `cpu-clock:u:` / `cpu-clock:k:`, so the matcher is right.
- **`doctor.py --probe debugger`: READY** here (`ptrace_scope=0`), attach + `DOCTOR_ATTACHED_PID` match +
  detach, child reaped.
- **`doctor.py --probe scheduler --system-wide-scheduler` fails *correctly* here**: exit 1,
  `status: UNAVAILABLE`, `samples: {switch: 0, wakeup: 0}`, and `scheduler-record.json` retains the real
  cause (`can't access trace events` / `No permissions to read /sys/kernel/tracing//events/sched/...`).
  That is the behaviour the whole PR is about, observed rather than asserted.
- **CLI gating**: `--probe scheduler` alone and `--probe perf --system-wide-scheduler` both exit 2 with the
  explanatory message, and **neither creates the output directory** — the refusal happens before any
  filesystem side effect.
- **The RenderDoc SSBO oracle is correctly derived, not copied.** `probe.c`'s list table ends with
  `{"[0,1,2,3,4,5]", {0,1,2,3,4,5}, 6}` as the fifth and last list, so the final SSBO is slots 0–5 holding
  their own ordinal and 6–255 holding `SENTINEL`. `struct.pack("<256I", *range(6), *([0xffffffff]*250))` is
  exactly that, and `IndexControlTests.expected()` rebuilds it a third way from bytes. Five draws, one exact
  1,024-byte buffer: both halves check out against the source.
- **CI**: `Host ASan + UBSan` is green with a real count — 6/6, `sanitizer_instrument_control` among them, so
  the deliberately faulty controls really did produce their diagnostics on a machine that is not the
  author's. The main `Linux` job runs `diagnostic_doctor` as test 3 of **361**. All 10 non-skipped checks pass.
- `git diff --check` clean; no absolute host paths in the diff; no `Claude-Session:` trailers; no numbered
  table touched.

## Non-blocking findings

**1. `perf_probe` collapses two capabilities into one verdict, and the losing case is the host the PR itself
documents.** `perf_probe()` requires *both* `cpu-clock:u` and `cpu-clock:k`, so the headline status and the
process exit code are the AND of user-space and kernel sampling. On `perf_event_paranoid=2` the kernel arm
cannot succeed — that is precisely what this PR's own `hostprof/README.md` correction now says
("excludes kernel sampling but does not universally forbid user-space events"), and the kernel's documented
`>=2: Disallow kernel profiling` backs it. So on such a host `--probe perf` prints `UNAVAILABLE` and exits 1
while the capability the next section of `DEBUGGING_WORKFLOWS.md` actually asks for —
`perf record -F 199 -g -p "$target_pid"` — is fully available. The per-control detail is in `report.json`,
so nothing is *hidden*; but the summary line and the exit code answer a question the user did not ask, on
the one configuration the PR went out of its way to describe correctly.
(To be explicit about my own derivation: I measured READY/READY at `paranoid=1`. The `paranoid=2` failure is
inferred from the documented semantics this PR relies on, not observed.)
Suggestion: report `perf-user` and `perf-kernel` as separate capabilities, or at minimum name the failing
control in the one-line summary so `UNAVAILABLE` does not read as "perf is unusable here".

**2. `debugger_probe`'s `finally` can turn a probe failure into a hang.** `doctor.py:147-149` —
if `child.wait(timeout=5)` raises `TimeoutExpired` (child left group-stopped after a gdb that died without
detaching), the exception escapes the `finally`, and `Popen.__exit__` then calls `child.wait()` with **no
timeout**. A bounded capability probe becomes an unbounded block. I could not reach this in practice — the
kernel resumes a tracee when its tracer dies — so this is a robustness note, not an observed defect. A
`try/except TimeoutExpired: child.kill()` around the wait closes it.

**3. `check_sanitizer.py:9`** — `binary, mode = sys.argv[1:]` raises a bare `ValueError` traceback on the
wrong argument count, in a file whose entire purpose is legible failure. One-line fix.

**4. Test-coverage gap: nothing pins that `--probe scheduler` reaches `perf_probe(scheduler=True)`.**
`test_scheduler_with_consent_runs` patches `perf_probe` and asserts only `call_count == 1`, and
`test_scheduler_command_is_system_wide_and_bounded` calls `perf_probe(scheduler=True)` directly. Invert the
mapping in `main()` — CPU control under the scheduler flag, `-a` under `--probe perf` — and all 23 tests stay
green while the consent gate guards the wrong recording. `self.assertIs(probe.call_args.args[1], True)` in
the existing test closes it, and given that this argument *is* the scope-consent mechanism, it is worth pinning.

**5. `PROSPER_SANITIZER=thread` has no CI job, but the docs say it must pass.**
`DEBUGGING_WORKFLOWS.md:202` — "Both configurations must pass". Only ASan/UBSan is wired into CI, and the TSan
arm is the one the PR body already reports as having been flaky under optimization before it was strengthened.
It is a 30-second job; without it the config rots silently and the doc's claim stops being true.

**6. Maintenance note: `tools/doctor/CMakeLists.txt:39-46` hand-lists four test/source pairs** that the main
build discovers via `GLOB_RECURSE`. Splitting `src/self/module.cpp`, `src/gpu/recompiler/rdna2_decode.cpp`,
`src/gpu/texture/bc_decode.cpp` or `src/host/platform/gpu_submit_gate.cpp` will redden `Host ASan + UBSan` in
an unrelated PR. It fails loudly rather than silently, so this is a tax and not a hazard — but a one-line
comment in the CMakeLists telling that author what to do would save them the archaeology.

## One observation on the documentation corrections

The PR summarises these as one bullet ("Correct stale perf, debugger-access and RenderDoc documentation"), but
the three have different evidentiary standing and I think that is worth separating in the record:

- The **`perf_event_paranoid`** correction in `hostprof/README.md` is a genuine factual fix and is right. The
  old text ("`>= 2` denies the `perf_event_open` syscalls without root") misstated the kernel's semantics;
  `>=2` disallows *kernel* profiling. This one is backed by the published contract.
- The **RenderDoc** correction is backed by a new observation — the author ran the wrapper.
- The **debugger-boundary** rewrite in `THREAD_WAIT_PROFILING.md` and `tools/AGENTS.md` is a *re-framing of a
  measured result with no new contradicting measurement*. The doctor probe explicitly certifies only
  same-environment child attach, so it cannot bear on the cross-boundary arm that produced the original
  finding. I am **not** asking for it to be reverted: the replacement keeps all three measured arms verbatim,
  keeps the operational guidance ("prefer the target's environment"), and drops a mechanism claim
  ("the rule is *same namespace*") that three data points genuinely do not establish. That is a better
  epistemic state than what it replaced. The note is only that a reader of the PR body cannot tell which of
  the three corrections rests on new evidence and which is a re-reading of old evidence, and per the charter's
  rule on corrections-that-read-as-checked, that distinction is the part worth writing down.

## Scope and residual risk, stated plainly

The RenderDoc half (`renderdoc_inspect.inspect()`, `renderdoc_control.h`) is **not exercised by any CI job and
cannot be** — it needs qrenderdoc, Vulkan and a real GPU. The unit tests cover the wrapper's process contract
(mocked `subprocess`) and the byte matcher, not the replay itself, so that path rests on a single hand
verification. That is disclosed in the PR and is the right trade for opt-in tooling; I am recording it so the
next reader does not mistake "23 tests green" for coverage of the replay path.

The PR's central discipline — that an installed binary, a zero exit, or an empty report is not evidence — is
implemented consistently and, more importantly, is observable failing correctly. That is the part I would have
rejected over, and it holds.

prosper/tools/doctor/doctor.py
```diff
@@ -0,0 +1,193 @@
...
+                      "Recording/decoding failed or required samples were absent; inspect logs."}
+
+
+def perf_probe(directory, scheduler=False):
+    if not shutil.which("perf"):
+        return {"status": "UNAVAILABLE", "reason": "perf is not on this environment's PATH."}
+    cases = [("scheduler", ["sched:sched_switch", "sched:sched_wakeup"], ["sleep", "2"])] if scheduler else [
+        ("cpu", ["cpu-clock:u"], [sys.executable, "-c", CPU_CONTROL]),
+        ("kernel", ["cpu-clock:k"], [sys.executable, "-c", CPU_CONTROL]),
+    ]
```
mattias800: **Finding 1 (medium): this probe ANDs two capabilities into one verdict, and the losing case is the host this PR itself documents.**

Both controls are required, so the headline `status` (line 125) and the process exit code are the AND of user-space and kernel sampling. On `kernel.perf_event_paranoid=2` the `cpu-clock:k` arm cannot succeed — which is exactly what this PR's own `hostprof/README.md` correction now says, and what the kernel's documented `>=2: Disallow kernel profiling` establishes. So on such a host `--probe perf` prints `UNAVAILABLE` and exits 1 while the capability `DEBUGGING_WORKFLOWS.md` actually asks for two sections later — `perf record -F 199 -g -p "$target_pid"` — is fully available.

The detail survives in `report.json`, so nothing is *hidden*. But the summary line and the exit code answer a question the user did not ask, on the one configuration this PR went out of its way to describe correctly.

Being explicit about my own derivation: I measured READY/READY on a second host at `paranoid=1` (`cpu-clock:u`=91, `cpu-clock:k`=47). The `paranoid=2` failure is **inferred** from the documented semantics, not observed.

Suggested fix: make the requested capability explicit on the CLI — `--probe perf` for user-space, `--probe perf-kernel` for kernel — so the exit code reflects exactly what was named. Anyone wanting the current strictness asks for both.

prosper/tools/doctor/doctor.py
```diff
@@ -0,0 +1,193 @@
...
+                          "-ex", "info threads", "-ex", "detach"], directory, "debugger")
+            ready = (result["returncode"] == 0
+                     and f"DOCTOR_ATTACHED_PID={child.pid}" in result["stdout"])
+            return {"status": "READY" if ready else "UNAVAILABLE",
+                    "reason": "Attached to this probe's child and detached." if ready else
+                    "Attach not established; inspect debugger.json.",
+                    "scope": "Same-environment child only; test host/container crossing separately."}
+        finally:
+            child.terminate()
+            child.wait(timeout=5)
```
mattias800: **Finding 2 (low, robustness): this `finally` can turn a probe failure into an unbounded block.**

If `child.wait(timeout=5)` raises `TimeoutExpired` — child left group-stopped by a gdb that died without detaching — the exception escapes the `finally`, and `Popen.__exit__` then calls `child.wait()` with **no timeout**. A bounded capability probe becomes a hang.

I could not reach this in practice: the kernel resumes a tracee when its tracer dies, so this is a robustness note rather than an observed defect. But a `finally` whose failure mode is "block forever" is worth closing in a tool whose entire contract is bounded, legible failure.

Suggested fix: escalate to `kill()` on timeout, and drop the `with` so `__exit__`'s unbounded `wait()` cannot run at all.

prosper/tools/doctor/check_sanitizer.py
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env python3
+"""Require a clean run AND a specific diagnostic from each deliberate defect."""
+import os
+import subprocess
+import sys
+
+
+def main():
+    binary, mode = sys.argv[1:]
```
mattias800: **Finding 3 (low): bare `ValueError` traceback on the wrong argument count**, in a file whose entire purpose is legible failure. `binary, mode = sys.argv[1:]` also KeyErrors on an unknown `mode` at line 15. Two guards returning exit 2 with a usage line.

prosper/tools/doctor/test_doctor.py
```diff
@@ -0,0 +1,182 @@
...
+            return status, probe.call_count, output.exists()
+
+    def test_scheduler_requires_explicit_scope_consent(self):
+        self.assertEqual(self.invoke(["--probe", "scheduler"]), (2, 0, False))
+
+    def test_unused_scope_consent_is_rejected(self):
+        self.assertEqual(self.invoke(["--probe", "perf", "--system-wide-scheduler"]), (2, 0, False))
+
+    def test_scheduler_with_consent_runs(self):
+        self.assertEqual(self.invoke(["--probe", "scheduler", "--system-wide-scheduler"]), (0, 1, True))
```
mattias800, line 118: **Finding 4 (low, coverage): nothing pins that `--probe scheduler` actually reaches the system-wide code path.**

This test patches `perf_probe` and asserts only `call_count == 1`; `test_scheduler_command_is_system_wide_and_bounded` calls `perf_probe(scheduler=True)` directly. Invert the mapping in `main()` — CPU control under the scheduler flag, `-a` under `--probe perf` — and all 23 tests stay green while the consent gate guards the wrong recording.

Since that argument *is* the scope-consent mechanism, it is worth pinning: assert the scope argument the mock was called with, in both directions.

.github/workflows/ci.yml
```diff
@@ -10,6 +10,26 @@ on:
...
+          sudo apt-get install -y g++ ninja-build
+      - name: Build isolated host checks
+        run: |
+          mkdir -p prosper/build-sanitizers/tmpdir
+          export TMPDIR="$PWD/prosper/build-sanitizers/tmpdir"
+          cmake -S prosper/tools/doctor -B prosper/build-sanitizers -G Ninja \
+            -DPROSPER_SANITIZER=address-undefined
+          cmake --build prosper/build-sanitizers -j2
+      - name: Verify host code and sanitizer positive controls
+        run: ctest --test-dir prosper/build-sanitizers --no-tests=error --output-on-failure
```
mattias800, line 31: **Finding 5 (low): the `thread` configuration has no CI job, but the docs say it must pass.**

`DEBUGGING_WORKFLOWS.md:202` — "Both configurations must pass the clean control and detect their deliberately faulty controls". Only ASan/UBSan is wired in here. The TSan arm is the one the PR body already reports as having been flaky under optimization before it was strengthened, so it is the one most likely to rot — and when it does, the doc's claim quietly stops being true.

This job took 29 s. A second one is cheap insurance for a claim the docs already make.

prosper/tools/doctor/CMakeLists.txt
```diff
@@ -0,0 +1,64 @@
...
+endfunction()
+# Existing production sources and existing regression arms, without guest execution or Vulkan.
+host_test(self_segment_key "${PROSPER_ROOT}/tests/test_self_segment_key.cpp"
+                          "${PROSPER_ROOT}/src/self/module.cpp")
+host_test(rdna2_decode "${PROSPER_ROOT}/tests/gpu/recompiler/test_rdna2_decode.cpp"
+                      "${PROSPER_ROOT}/src/gpu/recompiler/rdna2_decode.cpp")
+host_test(bc_decode "${PROSPER_ROOT}/tests/gpu/texture/test_bc_decode.cpp"
+                   "${PROSPER_ROOT}/src/gpu/texture/bc_decode.cpp")
+host_test(gpu_submit_gate "${PROSPER_ROOT}/tests/host/platform/test_gpu_submit_gate.cpp"
+                         "${PROSPER_ROOT}/src/host/platform/gpu_submit_gate.cpp")
```
mattias800, line 51: **Finding 6 (note, maintenance): this hand-lists four test/source pairs that the main build discovers via `GLOB_RECURSE`** (`prosper/CMakeLists.txt:71`).

Splitting any of `src/self/module.cpp`, `src/gpu/recompiler/rdna2_decode.cpp`, `src/gpu/texture/bc_decode.cpp` or `src/host/platform/gpu_submit_gate.cpp` will redden `Host ASan + UBSan` in an unrelated PR whose author has no reason to know this file exists.

It fails loudly rather than silently, so this is a tax and not a hazard — but a comment naming the failure and the fix would save that author the archaeology.

mattias800 (commented):
**APPROVED** — verdict unchanged from my review earlier on this same head; restated here so the newest review still carries it.

The six findings below are now posted inline against the lines they concern, plus one non-blocking observation on the documentation corrections. None of them blocks: every one is either a false *negative* (conservative), a robustness note I could not reach in practice, or a maintenance/coverage suggestion. The PR's central discipline — that an installed binary, a zero exit, or an empty report is not evidence — is implemented consistently and is observably *failing correctly*, which is the part I would have rejected over.

Independently reproduced on a second host before approving: 23/23 unit tests; `--probe perf` READY with real `cpu-clock:u`/`cpu-clock:k` sample counts (the matcher's spelling assumption is the thing that would have silently broken it, and it holds); `--probe debugger` READY; `--probe scheduler` failing *correctly* with exit 1, zero samples and the real cause retained; both mis-specified scope invocations exiting 2 without creating an output directory. The RenderDoc SSBO oracle re-derived from `probe.c`'s fifth list and confirmed exact.

I am picking these up and fixing them now.

prosper/docs/THREAD_WAIT_PROFILING.md
```diff
@@ -114,17 +114,19 @@ discarded, got an empty report, and read it as "nothing blocked" for a process w
...
 ```
 process on host      + gdb on host       -> works
 process in distrobox + gdb in distrobox  -> works
 process on host      + gdb in distrobox  -> ptrace: Operation not permitted
 ```
 
-Per trap 1 the denial is *silent*, so guidance that puts the debugger on the wrong side does not produce an
-error — it produces a clean-looking empty profile. (The first version of this tool said "run on the HOST",
-which is right only for a host process and wrong for prosper, which runs in the container.)
+These are observations of that setup, not universal host/container rules: credentials, namespaces
```
mattias800, line 125: **Observation (non-blocking, not a request to revert): this correction has different evidentiary standing from the other two in the PR, and the body presents all three as one bullet.**

- The `perf_event_paranoid` fix in `hostprof/README.md` is backed by the kernel's published contract — the old text misstated `>=2`, which disallows *kernel* profiling, not all `perf_event_open`. Correct.
- The RenderDoc correction is backed by a new observation: the author ran the wrapper.
- **This one is a re-framing of a measured result with no new contradicting measurement.** The doctor probe explicitly certifies only same-environment child attach (`doctor.py:146`), so it cannot bear on the cross-boundary arm that produced the original finding.

I am not asking for it back. The replacement keeps all three measured arms verbatim, keeps the operational guidance ("prefer the target's environment"), and drops a mechanism claim — "the rule is *same namespace*" — that three data points genuinely do not establish. That is a better epistemic state than what it replaced.

The note is only that a reader of the PR body cannot tell which correction rests on new evidence and which is a re-reading of old evidence, and per the charter's rule on corrections that read as already-checked, that distinction is the part worth writing down.

prosper/tools/doctor/doctor.py
mattias800: Fixed in dec2e5c09. `--probe perf` is now user-space sampling and `--probe perf-kernel` kernel sampling, each with its own verdict and its own scope string; `PERF_CASES`/`PERF_SCOPE` carry the reason so the next reader does not re-merge them. Asking for both on a paranoid=2 host correctly returns one READY and one UNAVAILABLE instead of condemning the host.

Live on this box at `paranoid=1`: `perf` READY (`cpu-clock:u`=95), `perf-kernel` READY (`cpu-clock:k`=40), exit 0. The paranoid=2 asymmetry itself is covered by a mocked arm (`test_kernel_denial_does_not_condemn_user_space_sampling`) rather than a live host — flipping that sysctl is global state on a box other agents are using, so I did not.

prosper/tools/doctor/doctor.py
mattias800: Fixed in dec2e5c09. Dropped the `with` entirely, so `Popen.__exit__`'s untimed `wait()` cannot run, and the teardown now escalates `terminate()` → `wait(5)` → `kill()` → `wait(5)`. Comment above it records why it is deliberately not a context manager, since that reads as an omission otherwise.

prosper/tools/doctor/check_sanitizer.py
mattias800: Fixed in dec2e5c09 — length and mode are both validated, printing a usage line and exiting 2. Verified: no args → 2, unknown mode → 2, both with the usage text.

prosper/tools/doctor/test_doctor.py
mattias800: Fixed in dec2e5c09. `invoke()` now records the scope argument each call received, and `test_scheduler_with_consent_runs` asserts `["scheduler"]` while a new `test_each_requested_capability_reaches_its_own_probe` asserts `["perf", "perf-kernel"]` — so inverting the mapping fails instead of staying green. Added `test_kernel_denial_does_not_condemn_user_space_sampling` as the regression arm for the finding one thread up. 23 → 25 tests.

.github/workflows/ci.yml
mattias800: Fixed in dec2e5c09 — added a `host-tsan` job, and `DEBUGGING_WORKFLOWS.md` + `VERIFICATION.md` now name both jobs so the claim and the enforcement point at each other.

I ran the race control **30 times with 0 failures** in the container before wiring it up, given the flakiness history — I did not want to hand every future PR a job that fails one run in twenty. Both configurations are 6/6 locally, and both defect controls were observed producing their specific diagnostics (`ThreadSanitizer: data race` exit 66; `heap-buffer-overflow` and `signed integer overflow` exit 1).

prosper/tools/doctor/CMakeLists.txt
mattias800: Fixed in dec2e5c09 — six comment lines naming the coupling, why the glob cannot be reused here, what the failure looks like, and the one-line fix. Also states that a link error here is always a stale list and never a missing dump fixture, since that is the wrong turn someone would otherwise take.

prosper/docs/THREAD_WAIT_PROFILING.md
mattias800: No code change — this was an observation about how the PR body frames the three corrections, not about the file. Carried into the PR description instead, which now separates them by evidentiary standing. The text in this file stands as written.

mattias800:
### Merged `main` in — and caught a clean-merge collision worth naming

`main` moved 8 commits while this branch was open. The merge is textually clean (`MERGEABLE`, no
conflicts, only `prosper/CMakeLists.txt` and `prosper/tools/AGENTS.md` even overlap, in different
hunks) — which is precisely why it needed reading rather than trusting.

**#3383 landed `tools/profiling_access/`**: a scheduler-only, duration- and byte-bounded, root-owned
capture helper reached through one narrow sudo rule, for unattended recording. This branch's new
front door — `DEBUGGING_WORKFLOWS.md` § *Scheduler tracing* — documents raw
`sudo perf record -a` and `sudo python3 tools/doctor/doctor.py --probe scheduler` and knew nothing
about it. Merging as-is would have shipped two unrelated answers to "how do I record scheduler
events", in the one document this PR adds to be the place people look first. No check could have
seen that: different files, different hunks, both green.

Fixed in c2149af54 — the section now names both routes and which question each answers: the doctor
probe establishes the capability interactively, `profiling_access` is the route for an unattended
capture, and neither changes what a recording proves. It also says to prefer the helper over
widening `sudo` or relaxing global perf settings, which is the trade #3383 exists to make.

Re-ran after the merge: 25/25 unit tests, docs numbered-table gate clean, `git diff --check` clean.

mattias800 (commented):
**APPROVED** on exact head `c2149af544fafaf54d287f9aacf2f6c6cfbc7f11`.

**Disclosure, because it changes how much this verdict is worth.** The three earlier approvals —
mine and one from an independent agent — were posted on `2a0970732` and cover the branch as it was
originally authored. The two commits since (`dec2e5c09`, `c2149af54`) are **mine**, written to
address my own review findings. So this line re-approves the head, but it is not an independent
review of the follow-up: nobody but me has read those ~160 lines. They are developer tooling with no
emulator runtime path, and they are exercised by CI on machines that are not mine, which is why I am
comfortable merging on the user's explicit instruction — but "three approvals" on this PR should be
read as *one independent approval of the original work*, not three of the current head.

**Green immediately pre-merge on this head**: 11 success, 0 non-success, 2 skipped
(`Progress tracker`, `Publish Release`). `mergeable=MERGEABLE`, `mergeStateStatus=CLEAN`.
Counts rather than exit codes, per `VERIFICATION.md`: `Host ASan + UBSan` 6/6, `Host TSan` 6/6 —
both including `sanitizer_instrument_control`, so the deliberately faulty controls emitted their
specific diagnostics on GitHub's runner and not only on mine — and the main `Linux` job 364/364 with
`diagnostic_doctor` among them.

What changed since the reviewed head:

- `dec2e5c09` — the six review findings. The one that mattered: `perf_probe` ANDed user-space and
  kernel sampling into one verdict, so `perf_event_paranoid=2` produced `perf: UNAVAILABLE`, exit 1,
  on a host where the sampling these workflows actually run was fine. Split into `--probe perf` and
  `--probe perf-kernel`, each answering for itself.
- `c2149af54` — merge of `main` (8 commits) plus a cross-reference to #3383's `tools/profiling_access/`,
  which landed while this branch was open and covers the same scheduler recording this PR documents.
  Clean textual merge, colliding guidance; no check could see it.

Residual risk, unchanged and worth restating at the merge point: the RenderDoc replay path is not
exercised by any CI job and cannot be, so it rests on one hand verification; and the `paranoid=2`
asymmetry that motivated the probe split is covered by a mocked arm rather than a live host, because
that sysctl is global state on a shared machine.

Issue #3384, opened 2026-09-05
[profiling_access] READY pipe captures fail the documented perf sched timehist analysis


mattias800:
## Failure contract

Refs #3383. The helper intentionally records perf pipe format on stdout, and profile.py declares READY after perf script decodes switch/wakeup events. That recording check succeeds, but the next command documented in prosper/tools/profiling_access/README.md (Routine captures) fails:

```sh
python3 prosper/tools/profiling_access/profile.py --seconds 2 --output prosper/build-access/repro
perf script -i prosper/build-access/repro/scheduler.data -F event
perf sched timehist -i prosper/build-access/repro/scheduler.data -s
# No sched_switch events found. Have you run 'perf sched record'?
```

The last failure also occurs without any PID filter, so changing process IDs or namespaces does not resolve it. The capture is NOT absent: the independently decoded ten-second artifact contained 721,192 switches, 350,580 wakeups, 10,557 new-task wakeups and 10,186 migrations. Every decoded event line parsed, and adjacent per-CPU switch chains had zero mismatches. Keep raw system-wide artifacts private; none are attached.

## Environment and investigated fallback

Measured 2026-09-06: host perf 7.1.13-200.fc44.x86_64, container perf 7.1.4-204.fc44.x86_64, kernel 7.2.1-ogc4.1.fc44.x86_64. Helper source is the original #3383 version; the later stderr hardening does not change recording format.

Plain conversion also fails, on both perf versions, with a filename, redirected stdin, or an actual stdin pipe:

```sh
perf inject -i prosper/build-access/repro/scheduler.data -o prosper/build-access/repro/seekable.data
# piped event processing failed for event of type: INVALID
```

The measured failure was at offset 0x7b74 inside embedded tracing metadata. Trying host perf inject --itrace or -b returned zero but produced an artifact rejected by timehist as broken or missing trace data; -b also reported an invalid build-ID filename length. These are not validated workarounds.

## Likely cause and next step

The source contract strongly indicates a pipe-format consumer mismatch, not a kernel event-schema or recording-permission failure: timehist checks for sched_switch in its event list before processing the event stream, while pipe attributes/metadata arrive in that stream. Primary source for that ordering: https://github.com/torvalds/linux/blob/v6.15/tools/perf/builtin-sched.c#L3102-L3124 (source-based explanation, not a claim that this old version is the installed binary).

[...]

Issue #3065, opened 2026-08-27
GTA V gameplay is copy-bound across several paths, not one hot spot — first symbolised profile, and three prior beliefs it corrects


mattias800:
GTA V reaches gameplay with the world rendering correctly, at **~1.2 guest flips/s**. This is the
first symbolised CPU profile of that state, and it contradicts three things the record currently
implies about where the time goes.

Measured 2026-08-27 on `d2b2e4d3`, Linux/RADV, `scripts/gta5/reach-performance-story.pad`,
`perf record -e cpu-cycles:u -F 99 -g --call-graph dwarf -D 300000` launched **under** perf (not
attached — see the instrument note), 441,773 samples over ~100 s of gameplay.

## The profile

| symbol | share | note |
| --- | --- | --- |
| `__memmove_avx512_unaligned_erms` | **27.6%** total | 14.6% on the guest-named render thread, 13.0% on prosper's own |
| `__memcmp_evex_movbe` | 5.3% | |
| `__memset_avx512_unaligned_erms` | 4.0% | |
| `sw64kb_copy<true>` (+ its `parallel_rows` workers) | ~5.5% | tiling de-swizzle |
| `resolve_dynamic_fetch` | 3.2% | |
| `drain_guest_gpu_writes` | 2.6% | |
| `rdna2_walk` | 2.4% | recompiler instruction walk |
| `ShaderCompileKeyHash::compute` | 1.9% | |
| `amdgpu_vamgr_free_va` | 1.8% | driver VA management |

The `memmove` callers, which is the part that took five wrong turns to establish:

| caller | share of that thread's memmove |
| --- | --- |
| `render_draw_pass_rgba` | 7.51% (3.34% of it a `std::vector<uint8_t>::_M_assign_aux` — the readback range-assign) |
| `register_live_renderer`'s resource-table lambda chain | 3.30% |
| `execute_ordered_gpustate` → `execute_live_compute_items` → `execute_item` | 2.92% |

**There is no single dominant cost.** The frame is copy-bound across several independent paths, with
copying as the theme rather than one hot spot to remove. That matters for planning: there is no
one-line win here, and any single change should be expected to move a few percent.

## Three corrections to the record

**1. Texture cost is no longer dominant.** The figure that has been steering this work —
`setup_resources texture=2638.7ms` / `build_resources texture=1165.0ms` — predates the rendering
series. Re-measured on current master at gameplay: `build_resources texture=287.8ms`,
[...]

PR #3385, opened 2026-09-05
gpu: hash finalized shader compile keys once

mattias800:
## Change

Refs #3065; does not close the broader gameplay-performance investigation.

Remove the provisional full shader-key hash from `make_shader_compile_key`. All three callers already attach diagnostic settings and unconditionally calculate the final hash before any cache lookup. The hash algorithm, complete-key equality, shader inputs and final cache identities are unchanged.

Add chained-vertex coverage for address-sensitive diagnostic selectors and reusable finalized keys in both lookup orders. Existing graphics/compute tests cover the sibling paths. These are behavior guards, not a timing benchmark; source inspection establishes the eliminated redundant call.

## Evidence and limits

[Current native windowed bank-gameplay baseline and retained artifact hashes](https://github.com/mattias800/prosper/issues/3065#issuecomment-5555267500): Performance mode visually confirmed, fresh save roots, full rendering cadence, shared GPU presentation, no compute skips or device loss. The clean CPU recording attributes 5.91% of sampled CPU to compiled-key hashing. This is not a share of frame time and does not predict an equivalent FPS improvement.

The status document now separates this Linux baseline from the existing Windows result. No end-to-end speedup is claimed. Compute writeback/resource handling and short-lived worker granularity remain larger investigation areas; scheduler tooling compatibility is tracked separately in #3384.

## Author verification

Head: `35d4e8375579d9254af2c8235f0fbf9b7c5ab1ef`; base: `a5150495e389deace1814db3322185e43b889e74`.

- Fresh isolated Release build with optimized symbols/frame pointers; built `prosper-app`, `test_shader_recompile_cache`, `test_cfg_trip_bound` through the development container.
- Exact-head CTest: `shader_recompile_cache` and `cfg_trip_bound_exec`, **2/2 passed**, including real Vulkan execution for the latter. Used `--no-tests=error`.
- Existing performance-report Python tests: **26 passed**.
- Touched Markdown table checks and `git diff --check`: passed.
- Full rendering snapshot matrix not run; compiled-module/cache contracts are the focused guard for this unchanged-output optimization.

Independent registered review approved this exact head. Its separate 300-second candidate live smoke
run reached the bank scene, retained a completed F9 bundle and frontend screenshot, and exited cleanly
without a device-loss signature. This is qualitative verification only: candidate CPU sampling overlapped
part of its F8 timing window, so it is not used for an end-to-end speedup claim. Details and artifact
hashes are in the author-verification comment. Applicable CI is still required before merge.

### DIFF
--- a/prosper/docs/GTA5_STATUS.md
+++ b/prosper/docs/GTA5_STATUS.md
@@ -30,6 +30,28 @@ an overall **~34x speedup** (measured across 106 rendered frames in 4.90 s; Run 
 Historical design note for the descriptor work: `docs/FLAT_LOAD_DESIGN.md`. Do not start from it; the
 descriptor-array lift it describes is complete.
 
+## Linux windowed baseline (2026-09-06, #3065)
+
+At `a5150495e`, a fresh Release build with `-O3 -DNDEBUG -g1 -fno-omit-frame-pointer`
+reached the bank gameplay scene on the unchanged `reach-performance-story.pad` route. Both save
+roots were fresh. Frontend screenshots confirmed **Graphics Mode: Performance** and subsequently
+the bank interior, characters, radar and walking tutorial. The nine-minute run used native rendering,
+full cadence, immediate presentation, default cache/worker settings and no diagnostic compute skips;
+the app confirmed the renderer's shared present queue. No device-loss signal was recorded.
+
+The native Linux/RADV F8 window measured **5.58 guest flips/s and 5.78 host presents/s** over
+5.02 seconds, with 1.51 process CPU cores. Its 720 renderer and 1,917 compute detail records were
+not truncated. Measured totals were 2,311 ms graphics and 1,547 ms compute; resource preparation
+accounted for 1,179 ms across its frontend/backend layers. These are scoped timing records, not an
+exhaustive GPU frame budget. The later F9 bundle completed with 73 submits and a matching frontend
+gameplay image, outside the timed window. Artifact hashes and capture details belong to #3065.
+
+A separate 20-second CPU profile retained 1,607 samples with zero reported loss. Compiled-key hashing
+was 5.91% of **sampled** CPU; this is not a share of frame time or a promised speedup. Inspection found
+that `make_shader_compile_key` computed a provisional hash which all three callers unconditionally
+overwrote after attaching diagnostic settings. Removing that first calculation preserves the final key,
+equality checks and shader output. End-to-end performance improvement is not yet established.
+
 ## Gameplay framerate optimization: reaching 21+ FPS (2026-09-03)
 
 **Platform**: Measured on **Windows 11 / Intel Core i9 (24 physical cores) / discrete NVIDIA GeForce RTX 4090 (24 GB VRAM, Vulkan 1.4)**.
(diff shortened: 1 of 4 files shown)


mattias800 (commented):
APPROVED at exact head `35d4e8375579d9254af2c8235f0fbf9b7c5ab1ef`.

Independent read-only review of the four-file diff, key construction/hash/equality, all factory callers and cache entry paths, the new test and its surrounding state restoration, and the baseline evidence linked in #3065. No builds, tests or GPU workloads were duplicated.

The deleted calculation is side-effect-free: `ShaderCompileKeyHash::compute` only reads key fields and mixes local integers. The only three callers each assign the final hash after diagnostic settings and before cache access. The chained-vertex early return does not use the unfinished key. Hash contents, full equality, compilation inputs, locking and bypass behavior are unchanged.

The added test checks separated identities for relocated identical prologs with an armed selector, nonempty results, and a warm shared-result hit in both call orders. It restores the diagnostic environment and ordinary cache entry before the existing LDS/code-mutation arms. Its scope is correctly limited: this test does not measure performance, prove different emitted diagnostic behavior, or detect every possible hash-quality regression (full equality can preserve correctness even with collisions). Source inspection, not that assertion, establishes removal of the redundant calculation.

The documentation separates the Linux baseline from Windows measurements, labels 5.91% as sampled CPU rather than frame time, and claims no end-to-end speedup. Baseline images are not presented as candidate validation. No private machine paths or disallowed attribution found in this diff, commit message or PR body.

No blocking findings. Applicable exact-head CI and completion/reporting of the author's candidate smoke run remain the merge owner's responsibility; this approval does not assert those pending results.

mattias800:
Exact-head live smoke verification for `35d4e8375579d9254af2c8235f0fbf9b7c5ab1ef` is complete.

- Candidate binary SHA-256: `9241452209279a3a134b4d9a6672b9d4c8fd0eb23b54af0885657fcf5d4a16f3`.
- Same unchanged Performance Story route, fresh save roots, native rendering/full cadence, default caches/workers, native shared-device presentation; no diagnostic compute skips. The settings screenshot explicitly shows Performance. Later frontend images show the bank interior, characters, radar and gameplay tutorial.
- The bounded run exited with status 0 after 300.94 s, with zero device-loss signatures. Its F9 bundle completed with 73 submits. Post-run peer census was clear; only the original desktop DRM clients remained.
- F9 BMP SHA-256: `6e039a445eb7157fc03effa1ba21f78eb6aedf0ba88f2e9379d517ee5f733aa2`.
- Completed F9 bundle SHA-256: `1762d749ef30c017eb6e6316390ea8a92e339b7b94098b4b17a9e3dae9c8532b`.

This is **qualitative smoke evidence**, not a pixel-exact comparison or speedup measurement. The candidate F8 trigger was earlier in the route than the baseline's, and the exploratory CPU recording overlapped part of its five-second post-trigger window; neither capture is used to infer a performance delta. A controlled end-to-end comparison remains outstanding. The eliminated provisional calculation is established by source inspection, and final-key behavior by the exact-head tests.

Additional local gates passed on the same head: contribution shape, usage-text checks, CTest invocation guards, Markdown table checks and `git diff --check`. Source head is unchanged from the registered approval. Applicable CI is still pending; no merge yet.

Issue #3388, opened 2026-09-06
compute_phase_report: retain nonzero alias setup time instead of labeling it no work


mattias800:
At `tools/perf/compute_phase_report.py:378–400`, alias-only image input suppresses measured alias duration and calls it “no work.” Mixed input reports alias duration only as excluded from the real-image decomposition. An alias can still perform expensive setup before discovering its canonical owner; absent detailed sub-timers are not evidence of no work.

Hand-built positive control, passed on stdin to `python3 tools/perf/compute_phase_report.py`:

```text
[compute-image] code=0x100 hash=0x000000000000000a binding=26 class=storage alias=1 addr=0x1000 alias_of=5 persistent=1 upload-skipped=1 extent=8x8x1 ms=3.500
[compute-phase] submit=1 code=0x100 hash=0x000000000000000a ok=1 setup_ms=3.50 setup_validate_ms=0.00 setup_buffers_ms=0.00 pipeline_ms=0.00 dispatch_ms=0.00 writeback_ms=0.00 writeback_prepare_ms=0.00 writeback_buffers_ms=0.00 writeback_images_ms=0.00 writeback_publish_ms=0.00 map_ms=0.00 prepare_ms=0.00 watch_ms=0.00 pack_ms=0.00 layout_ms=0.00 notify_ms=0.00 cache_ms=0.00 cleanup_ms=0.00 total_ms=3.50 subgroup=64
```

Actual: exit 0 and empty stderr, the phase table retains 3.5 ms setup, but the image section says `all 1 records are aliased folds (no work, no sub-timers); nothing to decompose.` This is an alias-only report-input control, not a claim that a complete runtime dispatch lacks an owner binding.

Expected: retain a separate alias-record count and measured duration (here 1 / 3.500 ms), explicitly saying detailed sub-timers are unavailable. Do not include aliases in real-allocation/upload denominators or double-count them on top of setup.

Measured motivation: #3065's latest attribution has 60 real-image records / 26.963 ms and 660 alias records / 101.930 ms. The first duplicate storage binding accounts for 101.836 ms. This is a pre-alias-processing interval, not the cost of copying alias fields.

[...]

PR #3389, opened 2026-09-06
gpu: avoid redundant snapshots for exact write-only image aliases

mattias800:
Refs #3065. This is a shared compute resource-preparation optimization, not a title-specific workaround or a rendering milestone.

### Failure scenario and contract

An exact write-only storage alias currently requests a renderer-owned image snapshot before discovering its already-prepared canonical image. Coverage proofs are binding-specific, and only the canonical owner publishes them. A duplicate can therefore materialize a full RTT every dispatch even after its owner proves a complete overwrite. The duplicate's snapshot is discarded by the later alias fold.

On the native Performance Story attribution described in #3065, the first duplicate of a 3840×2160 RGBA8 output averages 3.512 ms against 3.836 ms whole setup (28 fully warm dispatches). This measures pre-alias processing, not the field-copy operation. No end-to-end speedup is claimed yet.

### Change and invariants

- Fold exact write-only, non-atomic storage aliases before duplicate coverage/snapshot preparation.
- Retain per-binding descriptor, device, dimensionality and DCC validation; retain the ownership query, which drains pending guest writes.
- Require complete resource-view identity and matching reflected storage representation, not merely the same address.
- Use the already-prepared canonical image, one staging allocation, canonical coverage proof, and one writeback/publication. Partial writes retain the required authoritative seed.
- Sampled/imported and mixed-access bindings keep their existing late path. No game-speed, frame-pacing, resolution, draw-skipping or shader-output policy changes.

Risk is silently widening alias identity or suppressing a necessary seed. The fast path deliberately excludes mixed access and atomic images; the existing identity predicate retains mip/tail/layer, capture-backing and DCC distinctions. Failure to obtain an unused duplicate snapshot is intentionally no longer a dispatch failure; failure to obtain a necessary canonical seed still is.

### Verification

Container build, Release `-O3 -DNDEBUG -g1 -fno-omit-frame-pointer`, Vulkan RADV:

```sh
cmake -S prosper -B prosper/build-linux -G Ninja -DCMAKE_BUILD_TYPE=Release \
  -DCMAKE_CXX_FLAGS_RELEASE='-O3 -DNDEBUG -g1 -fno-omit-frame-pointer' \
  -DGAME_DUMP=<DUMP_ROOT>/PPSA24651-app0 -DPROSPER_APP=ON
cmake --build prosper/build-linux -j6 --target prosper-app test_game_compute \
  test_image_identity test_spirv_storage_match test_live_compute_descriptor_array \
  test_live_compute_host_read_barrier
ctest --test-dir prosper/build-linux --no-tests=error --output-on-failure \
  -R '^(game_compute_exec|game_compute_exec_no_adaptive_storage_result|game_compute_exec_scratch_poison|image_identity|spirv_storage_match|live_compute_descriptor_array|live_compute_host_read_barrier)$'
```

Initial and restored-source runs: **7/7 pass**, game compute **555 / 541 / 555 assertions**. New synthetic fixtures require both aliases' distinct output rows, exactly two ownership queries, one required proving/partial snapshot and zero unused warm-full snapshots, plus one canonical write notification. Partial missing-seed rejection leaves guest bytes unchanged.

Mutation disabling only the new early fold: **7 assertions fail** in the added fixture, including snapshot count, warm execution and exact pixels; restoring it returns all seven tests to green.

Additional whole-fixture `PROSPER_VERIFY_SEED_SKIP=1` / `PROSPER_NO_SKIP_SEED=1` runs are **not green**: existing cache/proof-dependent assertions fail (44 / 6 respectively), outside the new fixture. Disabling only the early fold retains those exact failure-message multisets plus four added alias snapshot-count failures per mode. Tracked separately in #3390; do not count those optional diagnostic runs as acceptance passes.

Exact-commit author rerun on **d4bb56bf6eff43b7aa088a8cf9bc191cd98adc8c: 7/7 pass, exit 0**, with 555 / 541 / 555 compute assertions. Independent registered review 5123624487 states **APPROVED** on that exact head. Applicable CI remains a merge gate. No snapshot-matrix run or new PS5-oracle comparison. Raw captures and local paths remain private.

### Native acceptance and measured scope

The exact-head candidate completed the same fresh-save `scripts/gta5/reach-performance-story.pad` route using native Wayland/RADV, immediate presentation, full resolution/cadence and default workers. Both runs used identical selected-program F8 phase/image timing and capped frontend texture detail, no compute skip, a continuously clear 300-second gate and a dedicated GPU lease with pre/mid/post process/DRM monitoring. No competing GPU consumer or build/profiler was observed.

Candidate completed the 450-second bound normally (451.010 seconds including shutdown, exit 0), with no device-loss signature, complete F8 and a complete 73-submit F9. All three retained native images were inspected: explicit **Performance** settings, the early masked-character bank scene, and the final bank interior with characters, radar, walking tutorial and “Go to the guard.” The known gameplay checkpoint remains recognizable; no new rung, pixel-equivalence or PS5-oracle claim.

| Mean milliseconds | Baseline fully warm (28) | Candidate fully warm (31) |
|---|---:|---:|
| First duplicate storage interval (binding 26) | 3.512357 | 0.001000 |
| Complete selected-program setup | 3.835714 | 0.304839 |
| Setup excluding that interval | 0.323357 | 0.303839 |

The measured benefit is approximately **3.5 ms removed per warm dispatch** of this program; **not a whole-game FPS claim**. Candidate alias rows all print 0.001 ms, so their precision is limited by the existing timer format. Full-warm strata require writeback cache-hit=1, seed-skip=1, poison=0 and canonical setup upload-skipped=1. Baseline also contains one cold/proving record and one setup-seeded transition, excluded here; this candidate F8 contains neither. Geometry, code/hash and resource identity match, and all 31 selected detail phases pair sequentially with F8 records.

Candidate F8: pre=20/post=21, 779 renderer / 2,120 compute records, zero drops. Its 5.02-second window reports 6.17 guest flips/s and 6.17 host presentations/s with 1.60 process CPU cores, versus 5.78/s in the earlier attribution window. These are single instrumented windows with different cold-event populations, not an established end-to-end throughput improvement. The selector hashes programs outside F8 and the F9 operation is outside the timed window.

Executable SHA256: baseline `0e4bea34d387576a94966d73c8e6f7d42590dca954df03a0556f7caac7e17c3c` (8c48fcf source tree equals main 805b49d53); candidate `c5f3cb7a0efe086f14ee44ee2ec7f6af00b6e4a6fba4187e047ba0415eeb72d7` (d4bb56bf6).

Separate existing source-derived mixed-read/write alias coverage risk is tracked in #3391; this PR neither widens its eligibility nor claims to repair it. The report's misleading “no work” alias label is tracked in #3388.

### DIFF
--- a/prosper/docs/GTA5_STATUS.md
+++ b/prosper/docs/GTA5_STATUS.md
@@ -52,6 +52,27 @@ that `make_shader_compile_key` computed a provisional hash which all three calle
 overwrote after attaching diagnostic settings. Removing that first calculation preserves the final key,
 equality checks and shader output. End-to-end performance improvement is not yet established.
 
+### Duplicate storage-image preparation (#3065)
+
+On merged `805b49d53`, a later native Performance Story F8 isolated **3.512 ms** in the first
+duplicate storage binding's pre-alias-processing interval, against **3.836 ms** total setup for
+program hash `1d3f91e7f9a30140` (28 fully warm dispatches). Its 3840×2160 RGBA8 output binds one
+canonical storage image and three exact aliases. The interval is not the cost of copying alias
+fields: the backend requests an RTT snapshot before discovering the alias, while only the owner
+publishes a binding-specific coverage proof. The renderer can materialize a full target on that read.
+
+Exact write-only storage aliases now fold before this redundant snapshot/coverage preparation,
+after descriptor, device, shape and DCC validation and the ownership query's pending-write drain.
+Reflected storage representation and complete resource-view identity must match. Canonical seeding,
+partial-write preservation and one writeback remain; sampled and mixed-access bindings retain the
+late path. Synthetic execution checks both aliases' distinct pixels, necessary partial seeds and
+missing-seed rejection; disabling only the early fold fails seven assertions. This establishes the
+mechanism, not an end-to-end speedup. Native comparison evidence remains in #3065.
+
+The phase-report tool's alias records are **not zero work**: this capture has 101.930 ms in aliases
+versus 26.963 ms in real-image records. The report's misleading alias-only wording is tracked in
+#3388. Do not treat its real-image-only decomposition as all image setup.
+
 ## Gameplay framerate optimization: reaching 21+ FPS (2026-09-03)
 
 **Platform**: Measured on **Windows 11 / Intel Core i9 (24 physical cores) / discrete NVIDIA GeForce RTX 4090 (24 GB VRAM, Vulkan 1.4)**.
--- a/prosper/frontends/shared/live/live_compute.cpp
+++ b/prosper/frontends/shared/live/live_compute.cpp
@@ -6334,6 +6334,39 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
             image_timing_requested && timing_item_selected &&
             (!timing_capture_only || perf_capture_timing) &&
             (!timing_trace_only || trace);
+        const auto bind_image_alias = [&](size_t i, size_t j, ComputeClock::time_point start) {
+            BoundImage& bi = images[i];
+            bi.alias_of = images[j].alias_of == SIZE_MAX ? j : images[j].alias_of;
+            const BoundImage& owner = images[bi.alias_of];
+            bi.image = owner.image; bi.memory = owner.memory; bi.view = owner.view;
+            bi.sampler = owner.sampler; bi.guest_bytes = owner.guest_bytes;
+            bi.texel_depth = owner.texel_depth;
+            bi.array_layers = owner.array_layers;
+            bi.mip_levels = owner.mip_levels;
+            bi.mip_staging_offsets = owner.mip_staging_offsets;
+            bi.dcc_metadata = owner.dcc_metadata;
+            bi.dcc_metadata_bytes = owner.dcc_metadata_bytes;
+            staging_bytes[i] = staging_bytes[bi.alias_of];
+            const ShaderResource& r = *bi.resource;
+            if (trace)
+                std::fprintf(stderr, "[compute]   image-alias binding=%u -> binding=%u addr=0x%llx\n",
+                             bi.binding, owner.binding, (unsigned long long)r.gpu_addr);
+            if (image_timing)
+                std::fprintf(stderr,
+                             "[compute-image] code=0x%llx hash=0x%016llx "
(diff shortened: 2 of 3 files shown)


mattias800 (commented):
APPROVED on exact head d4bb56bf6eff43b7aa088a8cf9bc191cd98adc8c.

Independent source and regression-contract review; no duplicate builds, tests, or GPU runs.

The early path in prosper/frontends/shared/live/live_compute.cpp retains per-binding reflection/device/shape/DCC validation and the ownership query before folding. It requires a completed canonical write-only, non-atomic storage owner, matching reflected representation, and the existing complete resource-view identity. Sampled, imported, and mixed-access cases retain the late path. Canonical preparation, barriers, coverage proof, writeback/publication, and failed-dispatch cleanup remain responsible for the shared image; aliases do not acquire independent allocations or publish authority.

The removed snapshot rejection is deliberately limited to an unused duplicate snapshot. The new two-binding fixture requires distinct output rows from both stores, current authoritative contents in untouched rows, both ownership queries, the exact snapshot count, and one canonical write notification. A missing necessary partial seed still rejects without guest writes. Warm-full expectations account for forced proof, disabled seed skipping, and immediate reproving.

Reviewed retained author evidence: exact-head seven-test acceptance pass, and mutation disabling only early folding produces seven failures in the new fixture. The optional forced-proof/no-seed-skip whole-fixture runs remain failing and are not acceptance passes; comparing their failure lists against the disabled-fold runs adds only four new snapshot-count failures in each mode.

No blocking correctness finding. This verdict does not establish native gameplay acceptance or an end-to-end speedup; the planned native run and applicable exact-head CI remain separate merge gates.

mattias800:
Author acceptance on exact head d4bb56bf6eff43b7aa088a8cf9bc191cd98adc8c: selected local ctest 7/7, exit 0, with 555/541/555 compute assertions. Disable-only-early-fold mutation fails seven new assertions; restored source passes. Native Performance Story comparison completed normally with no device loss, complete F8/F9, and all three captured images inspected at the known rendered bank checkpoint. Fully warm setup is 0.304839 ms (31 records) versus 3.835714 ms (28); this is a per-dispatch mechanism result, not a whole-game FPS claim. Full commands, state-stratification limits, separate diagnostic failures (#3390), and binary hashes are in the PR body. Registered independent APPROVED review 5123624487 is on this exact head. Merge gate now reports nine passes, two intentional skips, zero pending/failures, with PR head equal to the branch tip. Final diff and whitespace checks are clean.

Issue #3393, opened 2026-09-06
Linux dmem write trace can abandon read-only protections after failed disarm


mattias800:
# Draft title

Linux dmem write trace can abandon its own read-only protections after a failed disarm

## Scope and evidence level

This is a source-derived diagnostic-only error-path bug in `prosper/src/host/memory/guest_write_watch.cpp` at `c5eb785c0facded58390312beec33515d8a4659e`. No runtime reproducer or game failure is claimed here. It predates the #3387 shared production-watch alias/protection repair; that repair's production-page recovery state does not provide recovery records when no `GuestWriteWatch` owns the physical page.

The trace does report an invalid observation, so this is **not** a claim that it publishes a false process-wide negative. The problem is guest-memory protection recovery after abandoning observation.

## Definite source-derived failure sequence

1. Enable the Linux diagnostic over real shared-memory aliases A and B, both guest-writable. Do not create a production `GuestWriteWatch`. Successful trace arming makes both aliases read-only.
2. Invalidate the trace through an ordinary normal-context operation, such as a physical-write notification. Let one required `mprotect(..., RW)` call fail. This is an explicit syscall-failure precondition, not evidence of how often it happens in a title.
3. `set_trace_armed_locked` returns false on the first failed call and attempts to roll back earlier aliases to their prior read-only state (`:966-995`). Thus, if opening A succeeds and opening B fails, successful rollback leaves A read-only. If B is encountered first, A is never opened and is also left read-only. No rollback failure is required for this consequence.
4. `trace_invalidate_locked` ignores the false result and nevertheless publishes `Invalid` (`:1013-1017`). With no production watch, the fault classifier recognizes trace-owned addresses only while status is Armed or Stepping (`:2108-2112`). A later otherwise legitimate store through A is therefore NotHandled even though the diagnostic is what removed A's write permission.
5. `guest_dmem_write_trace_configure` has a related lifecycle hole: it ignores a failed disarm and then assigns `w.trace = {}` (`:1425-1427`), discarding the old alias/protection recovery metadata altogether.

The same lifetime question applies to signal-time partial open/rearm cleanup, but that is not needed to establish the normal-context failure above.

## Minimum discriminating CPU-only fixture (not yet run)

- Map a real section/memfd twice as RW and register its physical topology. Arm only the diagnostic; assert the initial Armed state and actual RO permissions on A.
[...]

PR #3394, opened 2026-09-06
memory: reconcile shared write-watch alias protections

### LINKED ISSUE #3387: Write watches can reuse stale alias protections while another registration retains the physical page
## Source-derived failure scenario

At `bfb9f657a6892d2c8e21aa35f2cc32cd6016e784`, `prosper/src/host/memory/guest_write_watch.cpp:1730` (`guest_write_watch_notify_direct_mapping_protection`) updates `AliasRange.prot` and dirties affected physical pages, but does not refresh the retained `WatchedPage.aliases[].prot` metadata. `GuestWriteWatch::create` at line 1319 reuses an existing physical page record when another registration owns it, instead of applying its freshly resolved alias list.

Concrete sequence: writable VA A and read-only VA B alias one physical page; two registrations keep its page record alive. The guest changes B to writable. Reset/recreate one registration while the other still references the old page. The new registration can still use a per-page alias entry that says B is read-only, so arming skips B. A direct B write can leave the new registration falsely Unchanged and permit a stale cached image. Adding a read-only alias after initial arming is another case to test: the mapping-added path returns before extending page aliases when the new alias is not writable.

This predates PR #3386. That PR's proposed registration-lifetime optimization requires explicit protection changes to force the existing reset/create path; that conservative invalidation does not itself repair shared page records when another reference survives. Do not treat the new sole-owner protection regression as coverage of this shared-owner case.

The diagnosis is from source tracing; no game rendering defect is attributed to it yet. A focused two-registration mapped-memory reproducer is still needed. Introducing commit has not been identified.

## Suggested repair and verification

Keep retained per-page alias/protection metadata consistent with authoritative mapping records, including a late read-only alias that becomes writable. Preserve requested guest permissions during disarm/rearm and fail closed for incomplete alias coverage. Verify two retained registrations, RW-to-R and R-to-RW changes, PROT_NONE transitions, partial-page-range changes, and writes through the changed sibling alias. Run existing late-alias, sibling-removal, stale-fault and physical-write controls beside the new case.

mattias800:
## Contract and repair

Fixes #3387.

Shared physical-page watch records outlive individual registrations. After a sibling alias changes permissions, reset/create could reuse stale per-alias protection metadata and miss raw writes through an upgraded alias. Conversely, disarming either a production watch or a diagnostic trace could undo an already-successful guest RO/PROT_NONE downgrade.

Reconcile retained permissions before either restoration owner runs, retain late read-only aliases, and invalidate diagnostic coverage by physical topology (including initially read-only aliases omitted from its writable arm set). Existing registrations remain invalid after a protection change.

Track incomplete production protection coverage explicitly. Same-state rearm must retry the complete protection pass; failed transitions cannot publish Unchanged. Failed last-owner/create cleanup retains fault-recovery records until restoration or mapping removal makes reclamation safe. A genuine fault on a revoked alias remains unhandled; production Resume requires the faulting alias to have been restored. Pending diagnostic single-step ownership remains intact. Signal-time changes do not allocate or destroy page records.

Windows/macOS retain their existing safe fallback. This does not change shader semantics, rendering cadence, game settings, or the HLE syscall/notification ordering. Diagnostic-only partial protection failure recovery remains a separate source-derived issue (#3393); no complete diagnostic recovery guarantee is claimed here.

## Verification

Real memfd aliases and raw volatile stores exercise early/late RO-to-RW upgrades with two owners on the same physical page, RO/NONE downgrades, partial ranges, failed reconciliation, and failed final release. Permission probes bypass prosper's handler; tracked fork-child stores require Unchanged before the store and Dirty in that same child afterwards. Trace tests cover Armed/Stepping, with/without a production owner, RO/NONE, pending TF ownership, and initially invisible RO aliases.

Expanded tests against unchanged runtime `c5eb785c0facded58390312beec33515d8a4659e`: 2/2 fail, exit 8, 35 failed assertions. Isolated runtime mutations against the full fix:

| Disabled safeguard | Result |
| --- | --- |
| Forced same-state reconciliation | Production test fails 7 assertions; trace passes |
| Last-owner recovery retention | Production raw-store recovery fails 1 assertion; trace passes |
| Revoked-alias fault rejection | Trace fails 4 assertions; production passes |
| Physical-topology trace invalidation | Trace-only upgrade fails 2 assertions; co-owner upgrade fails its invalidation assertion then times out at 60 s; production passes |

All mutations were restored. Full implementation has passed all 8 related cases (exit 0), including RADV compute execution with 701/687/701 assertions. Exact-head author verification will be posted separately after publication.

Build/test commands (Linux distrobox, Release, frame pointers and line tables):

```sh
export TMPDIR="$PWD/prosper/build-linux/tmpdir"
export PROSPER_GAME_ROOT=<DUMP_ROOT>
cmake -S prosper -B prosper/build-linux -G Ninja -DCMAKE_BUILD_TYPE=Release \
  -DCMAKE_CXX_FLAGS_RELEASE="-O3 -DNDEBUG -g1 -fno-omit-frame-pointer" \
  -DGAME_DUMP=<DUMP_ROOT>/PPSA24651-app0 -DPROSPER_APP=ON
cmake --build prosper/build-linux -j6 --target test_guest_write_watch test_dmem_write_trace \
  test_watch_mapping_generation test_guest_writable_cache test_guest_memory_map \
  test_game_compute prosper-app
ctest --test-dir prosper/build-linux --no-tests=error --timeout 60 -V \
  -R '^(guest_write_watch|dmem_write_trace|watch_mapping_generation|guest_writable_cache|guest_memory_map|game_compute_exec.*)$'
```

Risk is Linux page-protection/fault lifecycle behavior; focused memory tests and compute consumers are the relevant guards. No snapshots or new live-game run for this PR, and no game-specific rendering or FPS improvement is attributed to this fix. Raw test artifacts remain local. RESOURCE_BINDING records the contract and falsifications.

### DIFF
--- a/prosper/docs/RESOURCE_BINDING.md
+++ b/prosper/docs/RESOURCE_BINDING.md
@@ -143,6 +143,35 @@ while neither does alone, and a previously inactive owner may gain a writer. Bot
 publication use the same member-aware key. These are seed-safety constraints, not a static proof
 that a shader's coverage is invariant under arbitrary input changes (#3391).
 
+### Write-watch alias protection
+
+Storage-result and texture caches may reuse a physical-page watch while other registrations still
+reference it. A guest permission change must update every retained alias's permissions before the
+production watch or diagnostic trace restores pages; recreating one registration does not recreate
+that shared page record. Late read-only aliases are retained so a later write upgrade cannot escape
+coverage. Trace invalidation follows physical topology, including aliases omitted from its original
+writable-only set (#3387).
+
+Failed production protection transitions explicitly invalidate page coverage. Neither a replacement watch nor
+a same-state rearm may publish Unchanged without a complete protection pass. Last-owner cleanup
+retains fault-recovery metadata if restoration fails; mapping removal can later reclaim it. A fault
+on a guest-revoked alias remains unhandled, and a trace's already-owned single-step still completes
+as invalid without restoring revoked permissions. Windows/macOS keep their existing safe fallback.
+This does not establish complete recovery after diagnostic-only partial protection failures.
+
+### Ruled out: shared watch reconstruction
+
+- **Reset/create repairs stale alias protections even when another registration survives:** falsified
+  by real shared mappings, a fresh Unchanged control and a raw upgraded-alias store that failed to
+  dirty the replacement. Updating only the mapping registry leaves the shared page stale (#3387).
+- **Disarming cannot undo an already-successful permission downgrade:** falsified by actual read/write
+  probes immediately after notification and last-owner release, for both RO and PROT_NONE. Production
+  and Armed diagnostic owners both restored old permissions; metadata reconciliation precedes either
+  restoration now (#3387).
+- **An armed bit plus a fresh generation proves complete coverage after a failed protection call:**
+  falsified by an inaccessible sibling alias. Forced reprotection and an explicit incomplete-coverage
+  state are required; attempted restoration is not proof that a faulting store can resume (#3387).
+
 ### Ruled out: per-binding seed safety
 
 - **A write-only owner's reflection is sufficient to omit seed data, provided untouched image
--- a/prosper/src/host/memory/guest_write_watch.cpp
+++ b/prosper/src/host/memory/guest_write_watch.cpp
@@ -619,6 +619,9 @@ struct WatchedPage {
     uint64_t phys = 0, generation = 0;
     uint32_t references = 0;
     bool armed = false;
+    // A failed transition can leave some aliases protected and others writable. Keep the fault
+    // recovery record, but never publish Unchanged or skip the next complete protection pass.
+    bool coverage_incomplete = false;
     std::vector<PageAlias> aliases;
 };
 struct RegistrationPage { WatchedPage* page = nullptr; uint64_t generation = 0; };
@@ -757,7 +760,7 @@ std::vector<ProtectionRun> protection_runs(const WatchState& w,
         if (!page) continue;
         const bool prior_read_only = page_requires_read_only_locked(w, *page, page->armed);
         const bool wanted_read_only = page_requires_read_only_locked(w, *page, arm);
-        if (prior_read_only == wanted_read_only) continue;
+        if (prior_read_only == wanted_read_only && !page->coverage_incomplete) continue;
         for (const PageAlias& alias : page->aliases) {
             if (!cpu_writable(alias.prot)) continue;
             const int full = host_prot(alias.prot);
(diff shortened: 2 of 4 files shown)


mattias800:
Author verification on exact head `7dfcc038523cd64f66675f2945da5abab04b6fab` against main `c5eb785c0facded58390312beec33515d8a4659e`:

- Build command in PR body: exit 0, all seven named targets built, including prosper-app.
- Exact ctest command in PR body: **8/8 pass, exit 0**, `--no-tests=error --timeout 60 -V`. Cases: guest_memory_map, guest_writable_cache, watch_mapping_generation, guest_write_watch, dmem_write_trace, game_compute_exec, game_compute_exec_no_adaptive_storage_result, game_compute_exec_scratch_poison.
- Compute execution explicitly reports RADV on Radeon 8060S; assertion counts **701 / 687 / 701**. Desktop-only DRM census before/after; no competing game/profiler observed. These are correctness results, not timing measurements.
- Full tracked-Markdown numbered-table check: exit 0. `git diff --check origin/main...HEAD`: exit 0; worktree clean. No session trailers or private host paths in the published changes.
- Expanded fixture baseline and each isolated mutation were actually rebuilt/run as recorded in the PR table. Untouched runtime: **35 failed assertions, both cases fail, exit 8**. Isolated incomplete-coverage/recovery-retention/revocation mutations: respectively **7 / 1 / 4 assertions fail**, each 1/2 cases failing with exit 8. Physical-topology mutation: trace-only upgrade fails twice, co-owned upgrade asserts invalidation failure then reaches the 60-second CTest timeout; exit 8. Full runtime restored before committing and rebuilding this exact head.

No new native-game run or snapshot, and no title-specific speed/rendering claim. Diagnostic-only failure recovery remains separately tracked in #3393.

mattias800 (commented):
APPROVED on exact head 7dfcc038523cd64f66675f2945da5abab04b6fab.

Reviewed all four changed files, the shared-page/reference lifecycle, normal-context protection reconciliation, and allocation-free fault-path changes. Retained permissions are refreshed before restoration; incomplete production coverage cannot yield Unchanged; failed cleanup retains recovery metadata without freeing referenced pages. Revoked aliases are rejected before pending trace-step handling. Windows/macOS fallback remains unchanged.

The regression fixtures exercise real shared backing and actual permissions. Tracked fork-child stores check Unchanged before and Dirty after the store in the same child; permission probes bypass the handler. Inspected baseline and isolated mutation logs: failures reach the intended coverage, release-recovery, revocation, and physical-overlap assertions. The physical-overlap mutant also reaches the explicitly reported bounded timeout; it is not represented as a complete assertion run. Exact-head author evidence reports 8/8 passing cases and RADV compute counts 701/687/701. No duplicate builds or tests were run for this review.

No blocking findings. Diagnostic-only partial-protection recovery remains separately tracked in #3393 and is explicitly excluded from this change’s guarantee. This approval makes no live-game rendering or performance claim.

mattias800:
Additional exact-head CPU stress guard on 7dfcc038523cd64f66675f2945da5abab04b6fab: ctest --test-dir prosper/build-linux --no-tests=error --timeout 60 --repeat until-fail:10 -V -R "^(guest_write_watch|dmem_write_trace)$" passed both cases for all ten repetitions each (20 successful executions, exit 0). No code changes.

PR #3399, opened 2026-09-06
Avoid full Linux memory-map scans and completion-write starvation

### LINKED ISSUE #1716: test_present_blit transitions a TRANSFER-only image to SHADER_READ_ONLY_OPTIMAL — the layout hazard the test's own header asked a validation layer to catch
Found by the first-ever Vulkan validation pass over prosper's test suite (#1704). It is by a wide
margin the highest-volume finding: **168 of the 187 validation messages** the whole suite produces
on the CI runner's driver come from this one test.

```
VUID-VkImageMemoryBarrier-oldLayout-01211
vkCmdPipelineBarrier(): pImageMemoryBarriers[0].newLayout (VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL)
is not compatible with VkImage ... usage flags VK_IMAGE_USAGE_TRANSFER_SRC_BIT|VK_IMAGE_USAGE_TRANSFER_DST_BIT.
The Vulkan spec states: ... if oldLayout or newLayout is VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL
then image must have been created with VK_IMAGE_USAGE_SAMPLED_BIT or
VK_IMAGE_USAGE_INPUT_ATTACHMENT_BIT.
```

## Where

`prosper/tests/test_present_blit.cpp` creates its producer/source images with transfer usage only
and then transitions them to `VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL` to imitate the layout the
renderer's front buffer is in when `present_blit_publish` receives it —
`test_present_blit.cpp:157`, `:266`, `:335`, feeding the publish calls at `:166`, `:269`, `:338`.

`prosper/frontends/shared/present_blit.cpp:97` creates the scanout slot images with
`VK_IMAGE_USAGE_TRANSFER_DST_BIT | VK_IMAGE_USAGE_TRANSFER_SRC_BIT`, which is correct for what
those images do; the mismatch is in the fixture's *source* images.

## Why it is worth fixing rather than just tolerating

The test's own header comment (`test_present_blit.cpp:10`) reads:

> Run it under `VK_LAYER_KHRONOS_validation` in CI to catch sync/layout hazards.

That instruction was never carried out, and the first time it was, the test turned out to contain a
layout hazard of exactly the kind it asked to be checked for. The scanout handoff assertions
themselves are sound — this is the fixture's setup, not the contract under test — but leaving it
means 168 messages of noise sitting on top of the one signal anyone would want to see from this
test, and a real `present_blit` layout regression would have to be spotted inside that.

## Suggested fix

Add `VK_IMAGE_USAGE_SAMPLED_BIT` to the fixture's source images so the layout it declares is one
they may legally hold, or transition them through `VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL` if the
intent is only to model "producer finished writing". Confirm against what the live renderer really
hands `present_blit_publish` — if that image is genuinely `SHADER_READ_ONLY_OPTIMAL`, it is sampled
elsewhere and carries `SAMPLED_BIT`, and the fixture should match it.

Currently allow-listed in `prosper/tools/vkval/allowlist.txt` so it stays visible.

### LINKED ISSUE #3398: Linux writable cache misses enumerate the full process map during native rendering
## Measured cost

Continuing #3065 on source-equivalent main `769a599b1` / retained executable `8185b421`:
a calibrated native Performance Story run reached the rendered bank scene, ended at its
450-second bound without a device-loss signal, and retained complete F8/F9 captures.

The earlier whole-process inheriting 299 Hz profile strongly perturbed process kernel CPU.
This run instead sampled user/kernel cycles at 49 Hz on one identity-pinned thread, with
inheritance disabled. Independent raw accounting gives process system CPU of
0.251 / 0.269 / 0.259 cores before/during/after; the previous approximately 2.02-core
recording bracket does not recur. Changing scope, frequency, inheritance and event set
together does not identify which factor removed that spike, or prove zero overhead.

All 1,709 samples are on the selected thread (866 user / 843 kernel), with zero lost records.
Recorded app/libc/RADV build IDs match live-process identities. User stacks validate the
graphics-renderer and ordered-submission execution path. Eleven quiet-gate samples and
30 runtime censuses found no competing experiment. Direct screenshots confirm Performance
graphics mode and the bank scene at “Go to the guard.”

`guest_writable` appears in 213 kernel samples / 5,855,020,838 sample periods:
29.0135% of the selected thread's kernel-event periods. Its actual read syscall subtree
alone is 27.8678%. Callers are the release-memory patch builder, CPU compute execution,
and renderer writeback. These are sampled on-CPU weights, not invocation counts,
off-CPU wait time, whole-process shares or projected FPS gains.

## Current mechanism

The writable cache exists and remains necessary: TLS positive ranges, invalidated by the
global mapping/protection generation. A miss reads the entire textual `/proc/self/maps`
table. Protection changes invalidate all cached positives; repeated negative queries also
scan again because only successful queries publish the collected ranges. This recording
does not distinguish the positive/negative or generation-change contributions.

## Candidate

On supporting Linux kernels, query the exact covering writable VMA using `PROCMAP_QUERY`,
walking adjacent VMAs only when the requested range crosses one. Do not use covering-or-next
to bridge a hole. Keep existing generation invalidation and positive-only publication;
capture the generation before querying. Retain the text path for unsupported kernels/headers
or unavailable ioctl access. No change to write-watch protection, memory authority or
guest rendering behavior is intended.

The [kernel proc documentation](https://www.kernel.org/doc/html/latest/filesystems/proc.html)
describes this binary interface, available since Linux 6.11. Its UAPI explicitly does not
promise a consistent snapshot across successive queries; this must not be described as
stronger synchronization than the existing permission probe.

Verification needed: live ioctl capability; read-only refusal; contiguous multi-VMA coverage;
gaps and remaps; generation invalidation; partial-query errors; unavailable-kernel fallback;
mechanism controls proving the fast path executes; and a matched native rendering/timing run.
No implementation or speedup is claimed yet.

mattias800:
Linux writable-cache misses currently parse the complete textual process map. This change queries the required writable VMAs with `PROCMAP_QUERY`, retaining text fallback on unsupported headers/kernels or failed queries. The cheaper path exposed starvation of retired completion writes, which this PR also fixes without retaining an artificial delay at each modern submit return.

Fixes #3398 and #3674. Continues #3407. Includes main `e2e4affb74ff`. **Acceptance complete:** all 549 local CTests, 11 strict Vulkan cases, required Linux/general CI and bounded cross-title checks pass. No sustained or newly-rendered FPS improvement is claimed.

## Behavior and ownership

- Require actual writable coverage across the whole request; never skip gaps or read-only regions. Retain mapping/protection generation invalidation and uncached negatives. Restart a failed binary-query fallback over the original request. Reopen maps per probe to preserve fork identity. Windows/macOS probing is unchanged.
- At a zero-active boundary, admit the next outer submit only after queued/in-flight completion writes land. Preserve same-thread nesting, concurrent active submits, FIFO landing order and selective renderer resource drains. Continuous overlapping active scopes without a zero-active boundary remain outside this bounded fix.
- Retire modern completion writes at the actual import-return checkpoint, after synchronous handler/backend cleanup. Remove the former additional modeled 1 ms. The older-SDK worker delay stays unchanged. No active-submit label is exposed early; a timer is not a substitute for GPU completion ownership.

The consecutive-label production guard fails 15 assertions with the old admission predicate. A real pending label now also traverses the generated import trampoline: it remains private through handler destruction and until the return hook, with ABI and guest-FS checks preserved.

## Evidence and diagnostic controls

Three same-executable query policies isolate binary query cost from broad/local text-cache warming: default; `PROSPER_NO_GUEST_WRITABLE_QUERY=1`; and that control plus `PROSPER_GUEST_WRITABLE_LOCAL_TEXT_CACHE=1`. Actual OS-probe fixtures cover adjacent/separated mappings, complete coverage, revocation and remap. Reintroducing broad local-text merging makes the adjacent-VMA guard fail.

The initial native diagnostic accumulated 26,578–35,361 queued records while paired submit checkpoints kept advancing. Admission restored progress but a matched trace-off source comparison exposed a slowdown: text-query guest/host counters fell **5.5737→4.1754/s**. The corrected candidate restores the matched text control to 5.5753/s versus the original 5.5737/s; native measures 5.9726/s in the corresponding run. These are shared instrumented settings, not default-settings FPS. The recurring stage-fold profiler can disqualify otherwise eligible parallel draw preparation; its absence was checked separately: the corresponding native follow-up measured 6.9666/s with correct bank rendering. Other timing/audio diagnostics remain enabled. This is profiler-setting sensitivity, not a newly implemented speedup (#3675). [Complete comparison and limits](https://github.com/mattias800/prosper/pull/3399#issuecomment-5672481485).

Direct admission timing then measured **1,672.25 ms** of completed caller waits across approximately five seconds. The zero-delay short diagnostic measured **35.40 ms**, retained queue progress and recorded 5.778 guest/host events/s. These are diagnostic windows, not sustained or newly-rendered FPS. Wait totals include scheduling, draining and mutex reacquisition and may overlap across callers; they are not an isolated timer or critical-path measurement.

Optional `PROSPER_PRESENT_HANDOFF_TRACE=1` adds bounded F8 publication/replacement/acquisition/outcome/wait records and nonblocking queue snapshots. GPU/CPU identity namespaces remain separate; overflow, missing samples and unresolved boundaries remain explicit. A held-worker production fixture proves current versus completed admission-wait observations; removing its elapsed update fails the guard. The reader rejects malformed/inconsistent records and preserves older captures. This is partial #3486 work, not completed-producer lineage.

The Vulkan held-slot fixture preserves leased images while testing five publications/three replacements; removing its replacement event fails eight assertions. Its source images now declare sampled usage before shader-read layout, removing the old validation exception. Fixes #1716.

## Validation

Runtime/test source `759640a834a6` passes 549/549 CTests and all 11 strict Vulkan core/synchronization cases with an empty allowlist, zero messages, and positive layer-load/deliberate-hazard controls. The observer passes its deliberate-failure control and 52 reader tests. Clean app/screenshot revisions are certified. Current head `5cc8c2175` adds only a ten-line documented falsification/recovery record; runtime and tests are unchanged. Both independent source reviews found no blocker; [Registered exact-head approval](https://github.com/mattias800/prosper/pull/3399#pullrequestreview-5204416583) is complete.

Both 450-second final GTA Performance Story runs complete cleanly, with 2-boundary F9 captures under the same 4 GiB budget, F8 before F9, identity-pinned CPU/GPU samples and no competing workload. Source/binary/report/input gates pass; Performance mode and the bank scene were visually confirmed. Sonic completes its bounded candidate run with the known broken modal/title and black-world HUD preserved (SDK 10 legacy coverage). Blue Prince, with the fold profiler absent, retains correct sampled movie layering and reaches the rendered entrance hall; paired scope counters confirm modern retirement runs. The corresponding GTA follow-up also completes its full protocol with correct bank rendering. All seven required Linux/general CI checks pass on the final head.

Audio remains separate: existing per-output/channel buffering is preserved. The final native main output records one demand shortfall by about 45 seconds and no additions through its final 451-second observation; the short-lived intro output records 40 through 16 seconds; this is not an audio fix or a physical-underrun measurement.


mattias800:
AUTHOR VERIFICATION — exact head a4830ada4baaacd7a0f6e6e038f812c7afa780aa. Fresh worktree Release build, explicitly reconfigured at this commit with -O3 -DNDEBUG -g1 -fno-omit-frame-pointer, app revision verified as this head. All seven focused CTests passed (7 discovered/executed, exit0): guest_memory_query, guest_writable_query_adapter, guest_writable_cache, watch_mapping_generation, guest_memory_map, eop_write, guest_write_watch. Native query fixture128checks; wrapped adapter51assertions/native-successYES. Deliberate unavailable-backend control fails16of128; wrong-epoch control fails2of51; both restored before this exact-head build. Simulated old-UAPI build passes98checks and explicitly reports no native-adapter coverage. Inherited cache-disable CTest1/1 passes; direct conflicting invocation refuses exit1. Docs tables and diff-check pass. Exact appSHA256 0be5d0720d1ceddd96d206e6f1ba78c5f1778e96111f87e958703b71e8a650d0. Same-route native comparison now in quiet gate under #3398 lease; native acceptance remains pending, as do independent review and applicable CI. No FPS gain claimed.

mattias800 (commented):
APPROVED for exact head a4830ada4baaacd7a0f6e6e038f812c7afa780aa.

Reviewed all nine changed files, the installed Linux UAPI contract, the retained focused verification output, and the documented evidence scope. No blocking code or test finding.

The binary path requires WRITABLE without READABLE or covering-or-next, rejects gaps and nonprogress, and publishes bounds only after proving the entire request. ENOENT remains a definite refusal; other ioctl failures restart the existing text fallback for the original request. Positive cache evidence keeps the generation captured before probing, with existing protection invalidation and cache-disable behavior preserved. Per-probe maps descriptors avoid retaining another process identity across fork. Unsupported headers and non-Linux helper builds explicitly request fallback; Windows/macOS executor probing is unchanged.

The final fixtures discriminate the intended mechanism: independent real-kernel capability control, writable-only mappings, exact adjacent coverage, holes, injected ENOTTY/EPERM/ENOENT, partial failure with prefix revocation, and generation advancement after a successful ioctl. The CTest launcher establishes the required cache environment before startup. Retained verification reports seven focused CTests passing; the reported binary-path and post-probe-epoch mutations fail without reducing their check counts.

This is approval of implementation and verification design, not native performance acceptance or merge readiness. The smaller positive warm set is an explicit empirical tradeoff. Applicable exact-head CI, published exact-head author evidence, and the pending matched native comparison remain separate merge gates; no FPS improvement is established by this review.

mattias800:
Native comparison on exact head `a4830ada4baaacd7a0f6e6e038f812c7afa780aa`: **merge held pending a repeat comparison**.

The retry completed 451.248 seconds through the native Performance Story route, with no device loss or supervisor error, complete F8/F9, and a clean post-run process/DRM census. Direct frontend images retain Graphics Mode Performance, the early bank scene, and the later bank interior/characters. F9 completed 84 submits with its exact-stem screenshot. This is retained rendering, not a new compatibility milestone.

The selected-thread CPU recording completed with 842 user and 835 kernel samples, no lost records or empty chains, and matching live/recorded app, libc and RADV identities. Recorded stacks confirm graphics and ordered-submission execution. `guest_writable` ancestry fell from 29.0135% of sampled kernel periods in the baseline to 0.9878% in the candidate; candidate samples run through the binary query helper. These are separate sampled populations, not frame-time percentages or a projected speedup. Selected-thread system CPU cores before/during/after were 0.1673 / 0.1693 / 0.1520 versus baseline 0.1975 / 0.2135 / 0.1999; accounting shows no large profiling-correlated spike.

The end-to-end observations are mixed:

| F8 post-sample counter | Baseline | Candidate |
| --- | --- | --- |
| Guest flips | 33 / 5.018587345 s = 6.5756/s | 38 / 5.016205168 s = 7.5754/s |
| Host presentations | 33 / 5.018587345 s = 6.5756/s | 25 / 5.016205168 s = 4.9838/s |
| Renderer / compute records | 828 / 2244 | 927 / 2452 |
| Dropped detail records | 0 / 0 | 0 / 0 |

Both captures have 20 pre and 21 post samples. Distinct rendered-frame counts are unavailable in this GPU-present path; neither counter is a substitute. The candidate host deficit also appears before F8 and in ordinary gameplay logs, so it is not explained by the F8 collection itself. Presentation mode and diagnostic settings match. Draw, texture-invalidation and dispatch populations differ, and the later scene/tutorial state is not identical. No stationary FPS gain or causal regression is established by this single pair.

Author checks, registered review and all nine applicable CI checks are green on this head (two expected skips). Those gates do not clear the unresolved native presentation result. Repeat the retained baseline before accepting this as an end-to-end optimization; keep the current implementation unchanged meanwhile. Raw profiles, images and private machine identities remain local.

mattias800:
Verified retained-baseline repeat completed without device loss or supervisor error, with complete F8/F9 and directly inspected Performance/bank screenshots. It closely reproduces the original baseline: post 33 guest flips and 32 host presentations in 5.014734423 seconds (6.5806 / 6.3812 per second), and pre 33/33 in 4.765067586 seconds. Its 825 renderer / 2213 compute records have zero drops. The 1706-sample profile has zero lost records or empty chains and matching live/recorded app/libc/RADV build IDs; guest_writable ancestry is 26.0883% of selected-thread kernel periods versus 29.0135% in the first baseline and 0.9878% in the candidate. These remain sampled period shares, not frame-time shares.

The two baselines also closely match their draw, texture-invalidation and CPU-fast execution populations, while the candidate differs. Direct images confirm the same bank location but not identical camera/tutorial state, including at the early pad-flip anchor. Thus fixed elapsed time and identical route alone do not make this a stationary workload comparison.

Merge remains held. Next is candidate repetition with actual scene/workload state checked, followed by a bounded presentation-side census if the host-rate deficit persists. Existing logs do not count all silent outcomes: publication supersession and swapchain-acquire timeout can occur without error messages, and the handoff mutex spans producer GPU-copy waiting. These are candidate explanations, not diagnosed causes. No further production changes have been made, and no end-to-end performance gain is claimed.

mattias800:
### Native repeat B2: discrepancy reproduced; merge remains held

Exact candidate head `a4830ada4baaacd7a0f6e6e038f812c7afa780aa`, unchanged executable and route
`prosper/scripts/gta5/reach-performance-story.pad`. Fresh save roots, native/full-cadence rendering,
immediate presentation, default workers. The run completed its 450-second bound and shutdown
(451.186 s), exit zero, no supervisor error or device-loss signal. All three retained frontend
images were inspected: Performance selected, followed by the rendered bank. No new visual milestone.

| Post-trigger measurement | Baseline A | Candidate B | Baseline A2 | Candidate B2 |
| Window seconds | 5.018587345 | 5.016205168 | 5.014734423 | 5.016479295 |
| Guest flip delta | 33 | 38 | 33 | 38 |
| Host presentation delta | 33 | 25 | 32 | 25 |
| Host presentations/s | 6.5756 | 4.9838 | 6.3812 | 4.9836 |

B2's complete F8 has 20 pre / 21 post samples, 909 renderer and 2,389 compute records, zero drops.
Its pre-trigger window also has the guest/host divergence (39/26); this is not solely introduced
by F8 detailed timing. Distinct-frame counters remain unavailable. The later paired F9 completed
with 84 submits, outside the timed window.

Independent raw-stack accounting confirms 851 user and 823 kernel samples, zero lost records or
empty callchains, and matching retained live/recorded app, libc and RADV build IDs. The permission
query family occupies nine kernel stacks, 188,601,288 / 15,901,897,261 periods = **1.1860% of that
thread's sampled kernel periods**, compared with 29.0135% / 26.0883% in the two baselines. Nested
helper/caller frames are counted once, not added. This is not a frame-time share.

Process system CPU before/during/after the recording bracket is approximately 0.261 / 0.254 /
0.247 cores; the selected thread is 0.174 / 0.170 / 0.163. Raw accounting and all quiet/runtime
censuses were independently checked. No large profiler-associated CPU spike was observed.

Both candidate captures have the same 28-program hash set and similar altered resource-work
populations. One differing hash between baseline and candidate was separately traced to different
captured raw guest programs, not merely recompilation of an identical stream. Camera/tutorial states
also differ. These are repeatable observations, not an explanation of their selection or correctness.

The optional late main-thread wait diagnostic was **not launched**: paired F9 completion was confirmed
too late for its pre-shutdown deadline. There is no wait/owner evidence from that proposed diagnostic.

**Decision:** keep this PR unmerged. The query-cost reduction repeats, but native end-to-end acceptance
remains unresolved. Next discriminator under consideration: the same candidate executable exercising
its existing full-text fallback, with any forced-unavailable mechanism separately validated and labeled.
No guest timing, shader policy, invalidation or synchronization semantics have been weakened.

mattias800:
Author verification — `8fb48cabcf5fa9da7686543f7b9255c3cc4d871f`, based on current main `5796d1e94bbd`.

- Release build completed with the app, SDL audio, FFmpeg/VA-API and the game corpus configured. App and screenshot binaries are certified clean at this exact head.
- Full CTest: **546/546 passed**, zero failures, 141.86 seconds, using `ctest --test-dir prosper/build-linux --no-tests=error --output-on-failure`.
- Query fixture: **128 checks** with real native-kernel success. Adapter fixture: **51 assertions**, including unavailable/partial-error fallback and pre-probe epoch publication. Native, text/broad and text/local scope fixtures each pass **41 assertions**; they independently assert actual map opens, binary-query execution and genuinely cached positives before revocation.
- Old-header standalone build: **98 checks**, explicitly unavailable native-interface coverage. This supplements the injected runtime-unavailable paths; it is not a claim of testing every older kernel.
- Mutation control: removing individual-VMA preservation from the text/local arm makes its adjacent-VMA oracle fail. The restored production source passes. Exact permission/hole/partial-query tests remain independent of this cache-scope control.
- Strict Vulkan synchronization validation: **9/9 passed**, empty allowlist, zero observed messages. The scanner independently observed validation-layer insertion and the deliberate write-after-write hazard before running the clean cases.

Strict command:
```sh
python3 prosper/tools/vkval/vk_validation_scan.py --build-dir prosper/build-linux --sync \
  --allowlist <EMPTY_FILE> --ctest-arg=-R \
  --ctest-arg='^(recompiled_shaders_render|vertex_fetch_render|texture_sample_render|multidraw_render|indexed_render|descriptor_array_render|textured_interp_render|gpu_execute|broadcast_htile_clear)$'
```

Independent source review found no remaining blocker after the scope-fixture corrections. **Production approval and native acceptance are still pending.** The PR remains draft: current game comparison attempts have been refused or discarded for observed competing work. The new series uses one immutable executable/source archive for original text, binary and text/local behavior, with recorded source/route/protocol hashes and controlled environment differences. A 300-second sampled quiet gate resets on contention and retries automatically after five minutes.

No fresh FPS, audio or library-wide nonregression result is claimed by these tests. Existing guest/host counters and sampled-pixel changes do not establish completed-producer FPS. Results will be judged by enclosing native costs, actual Performance-mode/scene witnesses, complete captures and independently checked sample quality.

mattias800:
The full trace-off comparison restores native frame delivery, but identifies a performance regression that must be resolved before merge. Keeping this PR in draft.

All three isolated GTA Performance Story captures completed the same 450-second protocol: F8 at 300 s, a separate 49 Hz selected-thread CPU profile before F8, and two F9 boundaries with a common 4 GiB bundle budget. Source/executable/route/environment identities were checked; the source A/B differs only in the admission fix and tests. Performance mode and the bank scene were visually confirmed in every run.

| Executable / query policy | Guest/host counter rate in F8 | F8 process CPU |
| --- | ---: | ---: |
| Before admission fix, `e4594f506073`, text | 5.5737 / 5.5737 per second | 1.175 cores |
| Admission fix, `795dfe71ec6b`, text | 4.1754 / 4.1754 per second | 0.884 cores |
| Admission fix, `795dfe71ec6b`, native | 4.1791 / 4.1791 per second | 0.863 cores |

These approximately five-second windows establish a slowdown in this comparison, not a library-wide percentage. Newly rendered FPS remains unavailable. The native query removes much of the sampled process-map traversal cost, but the two fixed-policy runs show no presentation-rate gain. This supersedes any interpretation that restored progress alone completed acceptance.

The next discriminator is completed admission-CV wait count and elapsed time, with a live waiter count. A production-path guard holds an actual worker effect until the waiting submit is observed, avoiding an assertion that depends on racing a 1 ms timer. The existing modeled delay is a candidate cause, not yet an isolated measurement. Any correction must retain FIFO/in-flight ownership, active-submit privacy, nested/concurrent behavior and the real import-return checkpoint.

Current pushed head still passes all seven required Linux/general CI checks, 549 full local CTests and 11 strict Vulkan/synchronization cases. The new observer is being tested locally. Cross-title acceptance is postponed until the final runtime policy is settled.

Audio remains separate: the fixed native run's main output recorded one initial demand shortfall and no further increases through shutdown; its short-lived intro output still recorded shortages. This is not a physical-underrun measurement or an audio fix.

mattias800:
Current acceptance: runtime/test source `759640a834a639bc46eab4ffd67c4c9283adb9a2`; final head `5cc8c2175` adds only the ten-line measured falsification/recovery record in `GTA5_STATUS.md`. No executable or test changed after verification.

All three retained GTA comparisons use 450-second `prosper-app` launches, the checked-in Performance Story pad route, fresh save/cache directories, full resolution and cadence, F8 at 300 seconds, and a complete two-boundary F9 at 310 seconds with the same 4 GiB budget. Handoff tracing is absent. Both arms retain recurring stage-fold profiling, which can disable otherwise eligible parallel draw preparation; these rates describe that shared instrumented configuration. A follow-up with the stage-fold profiler absent is recorded separately. Selected-thread 49 Hz user/kernel sampling and 10 Hz GPU sampling cover the same earlier 250–270-second bracket; process, revision, environment, route, helper and artifact identities are checked. Both final runs and their instruments exit zero with no competing workload. Performance mode and the rendered bank scene were visually confirmed in both.

| Runtime / query | Guest = host events/s | F8 CPU cores | F8 readback ms |
|---|---:|---:|---:|
| Before admission / text | 5.5737 | 1.1754 | 38.697 |
| Corrected admission / text | 5.5753 | 1.1713 | 40.384 |
| Corrected admission / native | 5.9726 | 1.1856 | 0.677 |

These are individual roughly five-second F8 observations within longer boots. They demonstrate recovery from the intermediate admission regression and a higher native-query presentation count in this pair; they do not establish sustained or newly rendered FPS, whole-library gains, or additive savings from overlapping component timers. The before/after text source gate pins the exact intended source changes and separately checks the additive report change against its Git blob. The same-executable native/text gate changes only the query policy and run-local paths. Both gates have deliberate-refusal controls.

Local validation, run in the configured Linux container:

```sh
ctest --test-dir prosper/build-linux --no-tests=error --output-on-failure -j4
python3 prosper/tools/vkval/vk_validation_scan.py --build-dir prosper/build-linux --sync \
  --allowlist <EMPTY_ALLOWLIST> --ctest-arg=-R \
  '--ctest-arg=^(recompiled_shaders_render|vertex_fetch_render|texture_sample_render|multidraw_render|indexed_render|descriptor_array_render|textured_interp_render|gpu_execute|broadcast_htile_clear|present_blit|present_blit_trace)$'
```

549/549 CTests pass (exit 0). All 11 strict Vulkan core/synchronization cases pass with an empty allowlist and zero messages; positive layer-load and deliberate WAW-hazard controls pass. The query-cache scope, consecutive real completion, held-worker timing and held-present-slot fixtures fail under their targeted production mutations and pass restored. The real completion label also remains private through handler destruction and the generated import-return hook in host/guest-FS ABI tests. Reader tests: 52 pass.

Audio is observed per output/open generation. In the final native boot, the main output records one demand shortfall by about 45 seconds and no additional shortfalls through its final 451-second observation. The short intro output records 40 through 16 seconds. The text control records one main-output shortfall and 40 intro-output shortfalls. Existing buffers are unchanged; these are demand counters, not physical device XRUNs or an audio fix.

Candidate compatibility checks completed cleanly on the same runtime:
- Sonic: 384 seconds, logos/autosave, existing broken modal/TIME UP title/clouds, then Cyber Space HUD over its known black world. Eight retained images checked. SDK 10 and zero modern scope counts: legacy-path coverage, not proof of modern retirement.
- Blue Prince: 422 seconds, stage-fold profiler absent. The sampled opening movie is correctly layered and the entrance hall/HUD renders. Eight images checked. Advancing paired modern scope counters confirm this path actually ran.
- GTA stage-fold-profile-disabled follow-up: 453 seconds, same full route/bundle/profile protocol; only the two stage-fold profiling environment entries were removed. Complete clean captures, correct Performance/bank images, and verified input delta with deliberate refusals. Guest/host counters measure 6.9666/s in its F8 observation, CPU 1.4119 cores. Other timing/audio diagnostics remain enabled. This is evidence of sensitivity to the profiler setting in these runs, not a newly implemented speedup or proof that parallelism changed; #3675 tracks missing diagnostic context in F8.

All seven required Linux/general CI checks pass on final head `5cc8c217557cda151271f721f53a4891255f1bae`. Registered exact-head review is being finalized. No unchanged screenshot/blog entry is added.

mattias800 (commented):
APPROVED for exact head `5cc8c217557cda151271f721f53a4891255f1bae`.

Independent source review and acceptance-evidence audit found no blocking defect. Reviewed the complete PR diff, including writable-query coverage/fallback, pre-probe generation tagging, platform guards, completion admission and return-hook ownership, presentation tracing, regression fixtures, and report semantics. The final documentation commit changes no runtime or test source from verified `759640a834a6`.

The native query publishes only complete writable coverage and rechecks the original request on fallback. Modern completion retirement retains active-scope, FIFO and inflight protection without the additional modeled delay. Continuously overlapping active scopes remain an explicitly bounded limitation. Presentation records preserve existing ownership and do not claim completed rendering lineage.

Audited author evidence: 549/549 CTests; 11/11 strict Vulkan cases with zero validation messages and reported positive controls; completed isolated GTA acceptance and source/input identity checks. The later 453-second GTA run removes exactly the two stage-fold profiling settings, retains the same runtime, and completes its F8/F9 and CPU/GPU collection. The comparison gate includes five deliberate refusal controls. Visual conclusions rely on the author's recorded inspection of the actual captures.

Blue Prince's completed smoke actually exercises modern retirement: paired scope counters advance from 3469 to 3530, with the recorded movie and entrance hall inspected. Sonic's completed smoke has zero modern scope entries and covers the unchanged legacy path only; its known visual defects remain.

Performance conclusions remain limited to the declared instrumented runs. The stage-fold profiler can affect parallel eligibility (#3675); removing it does not make the remaining F8/profile measurements an all-instrumentation-off baseline. Neither presentation counters nor single capture windows establish fresh-render FPS or a general speedup.

No builds or runtime tests were duplicated during this review.

PR #3402, opened 2026-09-06
pixel_history: answer "why is this pixel this colour" in one call

mattias800:
## Goal

A black pixel has four causes, they are indistinguishable by looking, and they send you to four
different files:

| verdict | what it means | where to look |
| --- | --- | --- |
| `NOTHING_DREW` | no event in the frame touched the pixel | geometry, viewport, culling, or the draw was never submitted |
| `ALL_REJECTED` | fragments arrived, every one was thrown out (the test is named) | depth/stencil state, scissor, the discard in the shader |
| `SHADER_WROTE_BLACK` | the shader ran, passed everything, computed black | resource binding, textures, uniforms, the shader |
| `STORE_LOST_IT` | the shader computed a colour, the target is black anyway | blend state, write masks, later overdraw |

This project has been separating them by inference — draw censuses, disabling passes, bisecting
state. That is slow, and it has produced retracted issues. RenderDoc answers the question directly,
per event; this makes it one call.

## Contract

`pixel_history.py CAPTURE --output NEW_DIR [--pixel X,Y] [--expect-control]`

- Defaults to the **brightest** pixel, not the centre. On a black frame the centre is exactly where
  nothing happened, and its empty history says nothing about why the frame is black.
- **Refuses to report** when the replay driver declares no pixel-history support. An empty history
  and an unsupported driver are byte-identical and mean opposite things — which is the confusion the
  tool exists to remove, so producing one would be self-defeating.
- Suppresses shader output for positional rejections and says `suppressed (no fragment ran)` rather
  than printing a plausible number.

## The capability is measured, not assumed

RenderDoc 1.45 on RADV/STRIX_HALO reports `APIProperties.pixelHistory = True`. `shaderDebugging`
also reports `True`, but nothing has exercised it, so the docs label it an unverified flag rather
than a capability this project has used.

## The control, and what it caught

`control.c` builds five full-screen draws at one named pixel, each differing in exactly **one**
property, engineered to produce one of each outcome. The final colour is reachable only if all five
arms behaved, so the control self-checks before any tool looks at it.

It earned itself on first use. **RenderDoc populates `shaderOut` for events whose fragment shader
never ran**: the scissored *white* draw reported the previous draw's **blue**.

```
eid 13  scissorClipped   shaderOut [0.000, 0.000, 1.000, 1.000]   <- draw 2's colour, not draw 3's
```

Well-formed, plausible, nothing null or flagged — and undetectable on a game capture, because the
true colour of a scissored draw is exactly what you do not know. Read as a shader output it says
`STORE_LOST_IT` and sends the reader to blend state. Recorded as **instrument trap 268**; the
general form is why the control exists at all: *a tool that reads a value cannot tell you the value
is meaningful; only a construction whose answer you already know can.*

## Verification

- `test_pixel_history.py`: **15/15**.
- **Mutation-tested.** Three mutations of the verdict logic, all now caught: trusting `shaderOut`
  from rejected events; an exact-zero black threshold; emptying the pre-fragment list. The third was
  **not** caught at first — the suppression lived in the RenderDoc-only path that CI cannot reach —
  so `suppress_shader_output()` was lifted into a pure function with its own arms. Recording that
  because the gap was real for a while and the tests looked complete throughout.
- Live `--expect-control` through the CMake-built binary: `CONTROL VERIFIED`.
- Auto-selection picks the scissored corner by luma (2040 = white) and reports the **inverse**
  verdicts there from the same capture — draw 3 passes, draw 5 fails depth — which is the
  construction confirming itself from a second direction.
- Main build configures; `pixel_history` registers as a ctest.

## Scope

No emulator runtime change and no title progression claimed. This reads an `.rdc`; prosper's seeded
F9 bundles retain guest-side state an `.rdc` cannot replace, and the docs say so. The replay half
needs RenderDoc and a GPU, so CI covers the verdict logic only — `--expect-control` is the check that
belongs on a new driver.


mattias800 (commented):

Reading review only (no builds, no GPU, no CI duplication). I read the whole diff on `34c3827ee`,
checked every RenderDoc semantic claim against the RenderDoc source rather than against the PR
prose, and re-ran the unit suite and the three named mutations locally.

The tool is worth having and the framing is right. I am rejecting on findings 1-6: two of them
make the tool return a confident verdict from data RenderDoc documents as undefined, one is that
the control does not cover the two verdicts it exists to validate, and one is that the PR's
load-bearing claim about the control is false for two of its five arms.

Everything below is reproducible from a checkout; where I cite RenderDoc I am citing
`renderdoc/api/replay/*` and `renderdoc/driver/*` in the upstream tree, not documentation.

---

### 1. BLOCKING — `unboundPS` is the purest "no fragment ran" case, and it is the one case not suppressed

`pixel_history.py:39-40` puts `unboundPS` in `REJECTIONS` but **not** in `PRE_FRAGMENT`
(`:37-38`), so `suppress_shader_output()` returns `False` for it and its `shaderOut` is trusted.

RenderDoc's own field documentation (`renderdoc/api/replay/data_types.h:2661-2669`):

> ``True`` if no pixel shader was bound at this event. ... On other APIs this is only reported if
> the pixel shader is entirely unbound but **this means the output may have undefined values.**

That is instrument trap 268's class, stated by the API, one list entry away from the list this PR
added to catch it.

Worse, `PixelModification::Passed()` (`data_types.h:2759-2764`) is:

```
return !sampleMasked && !backfaceCulled && !depthClipped && !depthBoundsFailed &&
       !viewClipped && !scissorClipped && !shaderDiscarded && !depthTestFailed &&
       !stencilTestFailed && !predicationSkipped;
```

`unboundPS` is **not** in it. So an `unboundPS` event arrives with `passed == True` and
`rejected_by == ["unboundPS"]`, and:

- it lands in `passed` (`:59`) and can become `passed[-1]`, i.e. "the final colour";
- its undefined `shaderOut` feeds `bright` (`:71`);
- `:281` prints `state = "PASS"` for it, because `rejected_by` is only rendered when `passed` is
  false — so the single event whose data the API calls undefined is displayed as a clean pass with
  no annotation at all.

Measured:

```
unboundPS event, passed=True, shaderOut=[0.9,0.1,0.1,1], postMod=[0,0,0,1]
  -> ('STORE_LOST_IT', '1 passing event(s) computed a non-black colour and the target is black
      anyway -- blend state, write mask, or later overdraw.')
```

Fix: move `unboundPS` into `PRE_FRAGMENT`, and treat it as a rejection for `passed` too (or at
minimum surface it in the printed state for passing events).

---

### 2. BLOCKING — `SHADER_WROTE_BLACK` is unreachable on any pixel that has a bright earlier writer

`:71` computes `bright` over **all** passing events, so `SHADER_WROTE_BLACK` is returned only when
*every* passing event computed black. On a real frame that is a strong condition: any pixel where
something bright drew and something dark legitimately drew over it can never produce
`SHADER_WROTE_BLACK`.

```
sky   passed shaderOut [1,1,1,1] postMod [1,1,1,1]
obj   passed shaderOut [0,0,0,1] postMod [0,0,0,1]     # a black object drawn over the sky
  -> STORE_LOST_IT
```

The last thing that wrote the pixel computed black, on purpose, and everything worked. The verdict
sends the reader to blend state and write masks. The reason string hedges with "or later overdraw",
but the verdict token is the machine-readable answer and the PR's own table maps it to
"blend state, write masks" first.

This matters more than an ordinary edge case because the `SHADER_WROTE_BLACK` / `STORE_LOST_IT`
split is the one distinction the tool exists to make; the other two verdicts are cheap. As written,
that split collapses toward `STORE_LOST_IT` as soon as the pixel has any history.

Fix is small: judge on `passed[-1]` (the actual last writer) and report earlier bright writers as
separate detail, or add a distinct verdict for "an earlier writer was bright, the last writer
computed black".

---

### 3. BLOCKING — the control never constructs either of the two verdicts it validates

`control.c:340-346` builds a pass, a depth failure, a scissor rejection, a discard and a pass. The
probe pixel's final colour is yellow, so `--expect-control` exercises `PIXEL_WAS_WRITTEN` and the
four rejection reasons. It never constructs:

- a passing draw whose shader computes **black** (`SHADER_WROTE_BLACK`), nor
- a passing draw whose colour the store discards — `colorWriteMask = 0`, or a zero blend factor
  (`STORE_LOST_IT`).

So the construction with a known answer covers the two verdicts nobody would get wrong, and not the
two the tool is for. `check_control` (`:199-211`) confirms this: it checks the pass/reject pattern,
the suppression, and the final colour, and nothing that reaches `classify`'s bright/black branch.

This is the charter's "positive control drawn from the same source tests the discriminator, never
the domain", one level up: the control is sound, it just tests a different pair than the one under
question. Two more arms at two more probe pixels would close it — one with
`cba.colorWriteMask = 0` and a bright push constant, one passing with `{0,0,0,1}`.

---

### 4. BLOCKING — the control's final yellow does **not** establish that all five arms behaved

This is the PR's load-bearing claim (`control.c:26-28`, PR body, commit message: *"the final yellow
is reachable only if all five arms behaved"*). It holds for arms 3, 4 and 5. It is false for arms
1 and 2.

State is `(colour, depth)`, depth `LESS`, writes on, cleared to `(RED, 1.0)`.

**Arm 2 (blue, depth 0.9, expected `depthTestFailed`) — masked.** If it wrongly *passes* the state
becomes `(BLUE, 0.9)`; draw 3 is still scissored, draw 4 still discards, and draw 5 at 0.2 passes
against 0.9 exactly as it passed against 0.5. Final colour: **yellow**. Two single-property
breakages that do this:

- `ds.depthCompareOp = VK_COMPARE_OP_ALWAYS` (`control.c:294`) — blue writes depth 0.9, yellow still
  wins.
- `ds.depthTestEnable = VK_FALSE` (`:292`) — in Vulkan this skips depth writes too, so every arm
  except the scissored and discarded ones passes and yellow is last.

**Arm 1 (green, depth 0.5, expected `PASS`) — masked.** If draw 1 never renders, depth stays 1.0,
blue at 0.9 now passes, white is scissored, magenta discards, and yellow at 0.2 passes against 0.9.
Final colour: **yellow**. `for (int i = 1; i < 5; ++i)` at `:377` reproduces it.

The masking is structural, not a tuning accident: arm 2's whole point is a *high* depth, which can
never block the lower-depth final arm, and arm 1's colour and depth are both overwritten by arm 5.
No probe pixel in this construction can see either — I checked the scissored corner too
(`(0..3, 0..3)` reads white in both worlds).

Why this is blocking rather than cosmetic: the self-check exists so a broken control fails loudly
*as a control* instead of silently becoming the oracle. For arms 1 and 2 the only remaining
detector is `check_control`'s `"control mismatch: expected ..., read ..."` (`:203`), which cannot
say which side is wrong — so on a new driver with a broken depth test, an agent reads "the tool is
broken here", which is exactly the ambiguity this construction was built to remove.

Minimum fix is honest restatement in `control.c`, the PR body and `AGENTS.md`: the final pixel
discriminates arms 3, 4 and 5; arms 1 and 2 are established only by the per-event reading. A real
fix is a second construction — re-run arm 1 + arm 2 alone against a fresh clear and assert green.

---

### 5. BLOCKING — target and pixel selection can produce `NOTHING_DREW` on a frame that plainly rendered

Two independent paths to the tool's most misleading verdict:

**Target.** `:122-126` takes `targets[min(req["target"], len(targets) - 1)]` from every texture
carrying `TextureCategory.ColorTarget`. In RenderDoc's Vulkan driver that flag is set for
`COLOR_ATTACHMENT_BIT` **or `TRANSIENT_ATTACHMENT_BIT`** (`renderdoc/driver/vulkan/vk_info.cpp:2601-2603`),
so on a prosper frame `targets[0]` is one arbitrary member of dozens — plausibly the swapchain
image. Combined with `SetFrameEvent(actions[-1].eventId, True)` at `:120`, which is the last
*drawcall* and therefore excludes any clear, copy, blit or resolve after it, probing a
present-target that is filled by a final blit yields an empty history and the report says
`NOTHING_DREW: No event in this frame touched the pixel.`

Also: an out-of-range `--target` is silently **clamped** by the `min()` rather than erroring, and
the report prints only the resourceId and dimensions, so a reader cannot tell which of the dozens
was answered about. A `--list-targets` and a hard error on out-of-range would fix both.

**Pixel.** `:228-234` parses `--pixel` and never bounds-checks it against the target. RenderDoc has
no bounds rejection either (`vk_pixelhistory.cpp:4543-4570` — an out-of-range pixel simply matches
no fragment), so a typo returns an empty history and the same `NOTHING_DREW`.

Given the tool's stated stance at `:104-107` — refusing to report rather than emit an empty history
that could be misread — both of these should refuse too.

---

### 6. BLOCKING (small) — one test guards nothing, and it is the one named for the trap

`test_stale_shader_output_cannot_manufacture_store_lost_it` (`test_pixel_history.py:38-45`) passes
against a broken implementation. Two reasons, either sufficient:

- the fixture hands `classify` `shader=None` for the scissored event, i.e. the suppression under
  test already happened in the test's own helper (`:7-11` computes `shader_output_suppressed`
  itself), so nothing in the production path is exercised;
- `bright` filters on `passed` (`:71`), so a rejected event is excluded whether or not its
  `shaderOut` was suppressed.

Measured against the PR's own three mutations — it survives all three:

| mutation | tests that fail |
| --- | --- |
| `bright` over `events` instead of `passed` | `test_a_rejected_event_never_counts_as_the_final_colour` |
| `PRE_FRAGMENT = ()` | `test_pre_fragment_rejections_suppress`, `test_mixed_rejection_suppresses` |
| `BLACK = 0.0` | `test_near_black_is_black` |

So the mutation the PR body attributes to this test ("trusting `shaderOut` from rejected events")
is actually caught by `:47-51`. The suite is 15/15 and the three mutations really are caught — that
part of the PR body is accurate — but this test should either be rewritten to assert something
`classify` can get wrong, or dropped. It currently reads as coverage that does not exist.

The other 14 bind. Nothing in CI covers the coupling between `suppress_shader_output()` and the
event dict built at `:166-176`; that lives only in `embedded()`.

---

### 7. Non-blocking — answering the two RenderDoc semantics you flagged

You asked specifically about `sampleMasked` and `predicationSkipped`. Both check out **for
Vulkan**, and I verified them in the driver rather than the docs:

- `predicationSkipped` is set only by the D3D11 backend (`d3d11_pixelhistory.cpp:1491`, from
  `PredicationWouldPass()` at `:1036-1037`). The whole draw is skipped, so no fragment ran.
  Correctly classified as pre-fragment; the Vulkan and GL backends never set it at all.
- `sampleMasked` in Vulkan is evaluated *before* the discard check in RenderDoc's own ordering
  (`vk_pixelhistory.cpp:4475-4481`, ahead of the `shaderDiscarded` block at `:4484-4490`), matching
  the Vulkan spec's placement of the sample-mask test ahead of fragment shading. Correct.

**But** on D3D11/D3D12 `sampleMasked` comes from the output-merger sample mask
(`d3d11_pixelhistory.cpp:1031-1034`, from `OMSetBlendState`'s mask), which is applied *after* the
pixel shader — so on a D3D capture a `sampleMasked` event **did** run its shader and suppression
discards real data. Neither `pixel_history.py`, `AGENTS.md` nor trap 268 restricts any of this to
Vulkan. One sentence naming the API would settle it.

**And the dual is missing, on Vulkan.** When the fragment shader declares `EarlyFragmentTests`,
RenderDoc evaluates `depthBoundsFailed` / `stencilTestFailed` / `depthTestFailed` *before* the
discard check and defers `shaderDiscarded` to the end
(`vk_pixelhistory.cpp:4444` `bool earlyFragmentTests = tfCb->HasEarlyFragments(eventId);`,
the guarded discard at `:4484`, the late one at `:4525-4528`). For such a pipeline those three
flags also mean no fragment ran — and `test_pixel_history.py:96-101` asserts they keep their output,
with the comment *"The shader DID run for these"*. So the pre/post split is a property of the flag
**and** the pipeline's early-fragment-tests mode, not of the flag alone. Trap 268's recognition cue
should say so, since it is the same trap in the other direction.

---

### 8. Non-blocking — smaller correctness items

- **`values()` reads only `col.floatValue` (`:88-91`).** `PixelValue` is a union; on a UINT/SINT
  target the bits reinterpret, so a stored `1` reads as `1.4e-45` and every verdict on an integer
  G-buffer or ID target is `SHADER_WROTE_BLACK`. `rd.CompType.Typeless` at `:162` requests no cast.
  Either refuse integer formats or read the matching union member.
- **Auto-selection is not an argmax (`:147-157`).** PIL's `resize` defaults to BICUBIC, so on a
  1920x1080 frame this is a bicubic downsample to 256x256; the reported coordinate
  `int(bxy[0] * sx)` is the *top-left of a 7.5 x 4.2 block*, which can easily be a black pixel next
  to the bright feature — and then the verdict is `NOTHING_DREW` about a pixel the user never meant.
  A small isolated highlight can also be lost in the downsample entirely. Related: the printed
  `brightest pixel (luma N)` is the **downsampled block's** luma, not the reported pixel's, so the
  number does not describe the pixel it is attached to. A true argmax over the full image is a few
  lines and removes both.
- **MSAA.** `SaveTexture` leaves `sample.sampleIndex = ~0U` (`control_types.h:652` defaults) so the
  PNG used for selection is the all-sample mapping, while `PixelHistory` reads
  `Subresource(0, 0, 0)` = sample 0 (`:162`). Different pixels, and per-sample entries are not
  aggregated.
- **`getattr(m, f, False)` (`:165`)** silently degrades if a RenderDoc build renames or drops a
  field: `ALL_REJECTED` then reports "no reason flagged" and, more quietly,
  `suppress_shader_output()` starts returning `False`. A one-time `hasattr` assertion over
  `REJECTIONS` at startup would make that loud, which is the tool's own stated policy.
- **`lit = max(final[:3]) if final else 0.0` (`:68`)** turns an unreadable `postMod` into "black"
  and proceeds to a verdict. Compare `:104-107`, which refuses to report on missing capability.
- **Alpha is ignored everywhere (`[:3]`)**, undocumented. `(1,1,1,0)` reports `PIXEL_WAS_WRITTEN`
  though it composites to nothing.
- **`check_control`'s docstring (`:200`)** says the expectation is the construction "restated
  independently of the C source". `CONTROL_EXPECTED[0]` is the render-pass loadOp **clear**, which
  is RenderDoc's reporting behaviour, not part of the five arms — so a driver or version that does
  not surface the loadOp clear fails `--expect-control` on a working tool. Worth a comment.

---

### 9. Non-blocking — documentation claims I checked

- **`DEBUGGING_WORKFLOWS.md` recipe is not runnable as written.** The capture line passes no
  `-c/--capture-file`, so renderdoccmd writes the `.rdc` to RenderDoc's default location, which on
  Linux is under `GetTempRootPath()` = `/tmp` (`renderdoc/os/posix/posix_stringio.cpp:222` via
  `core.cpp:693-698`). This repo's charter forbids run artifacts on `/tmp` — RAM tmpfs, shared
  per-user quota, and an exhausted `/tmp` kills the Bash tool for every concurrent agent — and the
  very file this PR edits (`GPU_PROFILING_EXTERNAL.md`) already records "RGP writes to `/tmp`" as a
  way a tool lies. A prosper frame `.rdc` is not small. Add `-c "$run_dir/frame"`.
  Separately, the second line reads `"$run_dir/frame.rdc"` while the first mentions only
  `$dump_dir`, so the two lines are not connected; and neither doc mentions that the tool requires
  **`qrenderdoc`** on `PATH` (`pixel_history.py:226`), which is not the binary the recipe installs
  or invokes.
- **Instrument trap 268 cites `#3382`.** That is the previous, already-merged PR — it is literally
  this branch's base commit (`4c894ca0e Add verified debugging and profiling workflows (#3382)`).
  The tool and the measurement are in **#3402**. Every other row in that table links the lane's own
  issue/PR.
- **`GPU_PROFILING_EXTERNAL.md`.** "Pixel history is now checked and works here" is supported — but
  by the live `--expect-control` run, not by `APIProperties.pixelHistory = True`, which is the
  clause the sentence leads with. The PR makes exactly that argument two sentences later about
  `shaderDebugging`, so leading with the flag undercuts its own point. Also the edit runs the
  `convert` sentence onto the end of the new paragraph (cosmetic).
- `BLOG.md`, the four-verdict framing in `AGENTS.md`, and the "measured, not assumed" capability
  claim all check out. The numbered-table gate is green on the merge-relevant baseline
  (268 rows, unique and ascending, no deletions).

---

### 10. Trivial

`git diff --check` is dirty: trailing whitespace at `prosper/tools/pixel_history/control.c:263`.
The charter requires `diff --check` before merge.

---

## What I confirmed rather than took on trust

- 15/15 tests pass; all three named mutations are caught (by the tests listed in the table in §6).
- The docs numbered-table gate passes against `origin/main`.
- The `--expect-control` expectation, the vertex shader's fullscreen-triangle indices, the push
  constant offsets against `fs.spvasm`/`vs.spvasm`, and the depth/scissor arithmetic of all five
  arms at both the probe pixel and the scissored corner — all correct as described, with the two
  exceptions in §4.
- RenderDoc semantics for `Passed()`, `unboundPS`, `sampleMasked`, `predicationSkipped`,
  `TextureCategory::ColorTarget`, `TextureSave` defaults and the default capture path, read from
  the upstream source, not from documentation about it.

## Verdict

**REJECTED.**

The idea is right and most of the machinery is right — the suppression, the capability refusal, the
brightest-pixel default and the decision to build a control at all are all good judgment, and §7
confirms the two semantics you were least sure about are correct for Vulkan.

What I cannot approve is that the two verdicts the tool exists to separate are the two the control
does not construct (§3), that one of them is unreachable in practice (§2), that the API's own
documented-undefined value feeds a confident verdict (§1), and that the PR's headline claim about
the control is false for two of five arms (§4). §1, §2, §5 and §6 are small diffs; §3 is two more
arms; §4 can be closed by restating the claim honestly even without a code change.

Happy to re-review on a corrected head.

mattias800:
## All six findings addressed in 2eb28aa52 — and the review was right on every one

I checked each against RenderDoc's own source before acting rather than taking them on trust. All six
held. Two were defects in exactly the class this tool exists to prevent, which is a fair thing to
have had pointed out.

**1. `unboundPS`.** Confirmed at `data_types.h:2759-2764` — `Passed()` genuinely omits it, and
`:2664` says the output "may have undefined values". So the event arrived `passed=True`, printed a
clean `PASS`, and its undefined output could produce `STORE_LOST_IT`. Now suppressed, printed as
`PASS!unboundPS`, and yields a new `OUTPUT_UNTRUSTED` verdict rather than a guess.

**2. `SHADER_WROTE_BLACK` unreachable.** Correct, and worse than it sounds: the verdict was
unreachable on *any* pixel with a bright earlier writer, which is most of a real frame — so the tool
would have failed precisely on the frames people investigate. Only the **last** passing event can
explain the final state, and that is what is read now.

**3 + 4. The control is rebuilt.** Both findings were right and they compound: the control only ever
constructed `PIXEL_WAS_WRITTEN`, so the two verdicts nobody gets right by luck were unvalidated; and
the "final yellow proves all five arms" claim was false, exactly as described — arm 2's depth
failure was invisible because a later lower-depth arm overwrote it either way.

Five regions now, four different verdicts:

```
  A  sequence      PIXEL_WAS_WRITTEN
  A' arm-1 only    PIXEL_WAS_WRITTEN     <- a strip only arm 1 paints, so arm 1 is observed
  B  black draw    SHADER_WROTE_BLACK
  C  no write      STORE_LOST_IT         <- colorWriteMask=0 pipeline
  E  all killed    ALL_REJECTED
CONTROL VERIFIED: 5 regions, every constructed verdict and rejection reason read back correctly.
```

The depth-fail arm runs **last** now, so a broken depth test ends the pixel blue. And the header
states plainly what the C self-check does and does not establish — a final picture cannot see a
misattributed rejection; only `--expect-control` can.

**Rebuilding it caught a second wrong assumption.** Region E read `SHADER_WROTE_BLACK` instead of
`ALL_REJECTED`, because **`vkCmdClearColorImage` is a pixel-history event and it passes** — the
clear was its only passing event. The control's own comment had confidently asserted the opposite.
The ground clear now happens before `StartFrameCapture`. Recorded as **instrument trap 269**, and it
matters beyond the control: any real frame that clears its target hands every pixel a passing black
event.

**5. Target and pixel selection.** Confirmed at `vk_info.cpp:2601-2603` — `ColorTarget` really is set
for `TRANSIENT_ATTACHMENT`. Now prefers the targets bound at the event, reports which list it used,
and fails loudly on an out-of-range `--target` or out-of-bounds `--pixel` instead of silently
analysing something else.

**6. The vacuous test.** Right — its fixture pre-suppressed, so it survived all three mutations. It
now forces the pre-fix value in, and the mutation it was credited with is covered by name.

**On the two entries I flagged as uncertain:** thank you for tracing them. `sampleMasked` returning
before the shader is confirmed at `vk_pixelhistory.cpp:4478-4481`, and `predicationSkipped` appears
only in `d3d11_pixelhistory.cpp`. Both are now documented as Vulkan-only reasoning. Your
**`EarlyFragmentTests` dual** is the one I had missed entirely — `:4525` confirms depth/stencil run
before the shader there, so those outputs are residue too. The Python API exposes no per-event flag,
so rather than pretend to detect it, no verdict is derived from any rejected event's output at all,
and the caveat is written down.

Also fixed: trap 268 cited **#3382** (this branch's base) instead of #3402, and the
`DEBUGGING_WORKFLOWS.md` recipe passed no `-c`, so it wrote its `.rdc` to the forbidden tmpfs and
then read a path it never produced. `git diff --check` is clean.

**Verification:** 19/19 unit tests; all three verdict-logic mutations caught (3, 1 and 3 failures);
live `--expect-control` green across all five regions; control self-check `CONSTRUCTED` on all five
probes; docs gate clean at 269 rows.

Re-review welcome on 2eb28aa52.

mattias800 (commented):
## Re-review on `2eb28aa52` — REJECTED, on one item

Reading review again (no builds, no CI, no GPU). I re-read the whole diff, re-derived every
RenderDoc claim from the upstream source, re-ran the suite, and ran six mutations rather than three.

**All six previous findings are genuinely closed, and the fixes are the right ones rather than the
minimal ones.** I want to be specific about that before the one thing I am blocking on, because the
work is good:

- `unboundPS` → `UNDEFINED_OUTPUT`, suppressed, `PASS!unboundPS`, `OUTPUT_UNTRUSTED`. Correct, and
  I checked the reachability: every `PRE_FRAGMENT` reason forces `Passed()` false, so `unboundPS` is
  the *only* way a passing event can carry suppressed output. The new verdict is exactly as narrow
  as it should be.
- `classify()` on `passed[-1]`. Correct, and `test_store_lost_it_still_reachable_after_an_overdraw`
  is the right companion — the fix could easily have killed `STORE_LOST_IT` and that test says it
  did not.
- The control is a different and much better object. The depth-fail arm last, the A′ strip, region
  independence by scissor, and `check_control` reading reasons rather than a picture — all of it
  answers what I raised.
- `git diff --check`, `-c` on the capture recipe, trap 268's `#3402`, the `EarlyFragmentTests` and
  D3D caveats: all recorded, with citations I traced and found correct
  (`vk_pixelhistory.cpp:4470`, `:4478`, `:4525`; `data_types.h:2759`, `:2664`;
  `vk_info.cpp:2601-2603` all say what the comments say they say).

**Mutation results — I ran six, including two that revert to the old behaviour.** Every one is
caught, each by a test that names the thing:

| mutation | caught by |
| --- | --- |
| `bright` over all events (old) | `..._rejected_event_never_counts...`, `..._overdrawn_bright_writer...`, `..._stale_shader_output...` |
| `PRE_FRAGMENT = ()` | `..._unsuppressed_stale_output...`, `..._mixed_rejection...`, `..._pre_fragment_rejections...` |
| `BLACK = 0.0` | `test_near_black_is_black` |
| `final = events[-1]` | `..._rejected_event_never_counts_as_the_final_colour` |
| `OUTPUT_UNTRUSTED` branch removed | `test_unbound_pixel_shader_is_not_a_clean_pass` |
| `bright` over all *passing* events (old) | `..._legitimately_overdrawn_bright_writer_is_not_a_lost_store` |

19/19 green. Finding 6 is properly fixed: with `suppressed=False` forcing the pre-fix value in,
`test_stale_shader_output_cannot_manufacture_store_lost_it` now fails under the first mutation,
which it did not before.

---

## BLOCKING — trap 269 is correct, and the tool is the thing it breaks

You asked whether the generalisation holds. **It does**, and I confirmed the mechanism independently:
`vk_pixelhistory.cpp` pushes a clear with `RDCEraseEl(mod)` and skips the test evaluation entirely
(`if(!clear && !directWrite)`), so every flag stays false and `Passed()` is true. A loadOp clear
does the same — the *previous* head's `CONTROL_EXPECTED[0]` was that clear, appearing as a passing
event, which is the same fact arriving from the other direction.

But the row states the consequence and then does not act on it. Trap 269, in your own words:

> any real frame that clears its target hands every pixel a passing black event, so "the last
> passing event computed black" can be the CLEAR rather than a shader, and **a verdict keyed on it
> would blame shading for a pixel nothing ever drew to**.

`classify()` is a verdict keyed on it. Measured against the new code:

```
cleared-to-black, nothing drew        -> SHADER_WROTE_BLACK
cleared black + draws scissored away  -> SHADER_WROTE_BLACK
cleared to sky blue, nothing drew     -> PIXEL_WAS_WRITTEN
```

The first line is the tool's headline input. `DEBUGGING_WORKFLOWS.md:45` routes "black world,
wrong colors, missing geometry" here; a black region of a cleared target where the geometry never
reached the rasterizer is *the* case. The reader gets
`SHADER_WROTE_BLACK: ... resource binding, textures, uniforms or the shader itself` and goes to the
shader, when the answer is that nothing drew. `NOTHING_DREW` is correspondingly unreachable on any
cleared target — which is every render target in every real frame.

That is the retracted-issue class this tool was built to end, produced by the tool, on the input it
is documented for. It is a stronger version of the finding you just fixed: there
`SHADER_WROTE_BLACK` was unreachable, here it is over-reachable, and both come from the same place.

**The fix is small, and it is CI-testable, which is the argument that decides it.** Whether an event
is a clear is a *pure input* to `classify()` — the same reasoning that made you lift
`suppress_shader_output()` out of the replay path applies unchanged. `walk()` at `:138-144` already
visits every action; recording `{a.eventId: a.flags}` there is two lines, `ActionFlags.Clear` exists
(`replay_enums.h:5098`), and then `classify` can return `NOTHING_DREW` (or a fifth honest verdict)
when no non-clear event passed. Every arm of that lives in `test_pixel_history.py`.

And it buys back something real: **with it, the control can put its clear back inside the capture.**
Right now the control was changed so that it no longer exhibits the phenomenon — which is the one
move that removes trap 269 from the control's reach. Handle clears in the tool and the control
becomes self-guarding for 269 the way it already is for 268, which is strictly better evidence than
avoiding the case.

Minimum acceptable alternative, if you want to defer the code: put the caveat where the reader
meets it — the printed `SHADER_WROTE_BLACK` reason string, the module docstring (which is
`--help`), `AGENTS.md` and the `DEBUGGING_WORKFLOWS.md` line. Right now it exists only in a 269-row
instrument-trap table, and the charter's own rule about corrections that live in one place applies
exactly here. I would still prefer the code.

---

## Non-blocking — region independence (you asked)

**Verdicts are independent; reasons are not.** Scissored neighbours never pass, so they can never
reach `passed[-1]`; every region's verdict is safe from the other regions. I checked each region's
full event list at its probe and each one is reachable only as designed, with two exceptions that
are about the *check*, not the verdict:

1. **Region E's `ALL_REJECTED` is reachable with its own arm deleted.** Arms 1-7 are all scissored
   away from (48,48) and — by your own measured note at `control.c:46-48` — still appear in the
   history. Delete arm 8 and E's history is seven `scissorClipped` entries: `passed` is empty,
   verdict is still `ALL_REJECTED`. `check_control` only compares E's verdict, and the C self-check
   probes E for black, which the pre-capture clear already provides. **Both layers stay green while
   the arm the region exists for is gone.** So E currently establishes "no passing event yields
   `ALL_REJECTED`", not "a discarded draw yields `ALL_REJECTED`" — which is what `AGENTS.md:25`
   claims it establishes. One line fixes it: require `shaderDiscarded` among E's reasons, the way
   region A requires its three.
2. **Region A's `scissorClipped` requirement is vacuous.** Arms 6, 7 and 8 (B, C, E) are scissored
   at A regardless, so that arm of the loop at `:285-288` can never fire — an arm-2 scissor read as,
   say, `viewClipped` would not be caught by it. It *is* caught by the C self-check (an ignored
   arm-2 scissor ends A white), so nothing is undefended; the check just does not defend what it
   says it does.

Also worth knowing: E's reason string reads `scissorClippedx7, shaderDiscardedx1`, so the region
built to demonstrate "the report names which test did it" names the wrong one first. Cosmetic, but
it is the tool's selling point.

## Non-blocking — `OUTPUT_UNTRUSTED` (you asked)

**The verdict is the right call.** It is a refusal, consistent with the driver-capability refusal at
`:131-134`, it is reachable only through `unboundPS` (the one genuinely undecidable case), and its
reason string tells the reader what to do next. It does not move ambiguity — it names it.

**As shipped it will be ignored, but for a documentation reason, not a design one.** `git grep`
finds `OUTPUT_UNTRUSTED` in exactly two files: `pixel_history.py` and its test. The module docstring
(`:7-16`, which is `--help`) still says "Four answers"; `AGENTS.md:18` lists four;
`DEBUGGING_WORKFLOWS.md:82` lists four. There are now six things `classify()` can return, and two of
them — `PIXEL_WAS_WRITTEN` and `OUTPUT_UNTRUSTED` — appear in no user-facing text at all. A reader
meeting `OUTPUT_UNTRUSTED` has nothing to look it up in.

## Non-blocking — claims that overstate the check

- **"`git diff --check` is clean" is false for this diff, in both the commit message and the PR
  comment.** Bare `git diff --check` inspects the *working tree*, which is empty once everything is
  committed, so it returns 0 for any branch. Against the base it returns 2:
  `prosper/tools/pixel_history/control.c:293: trailing whitespace` — the same line I flagged last
  round at `:263`, moved, not fixed. Right command, wrong scope; the instrument answered a different
  question. `git diff --check <base>..HEAD`.
- **"every rejection reason read back correctly"** — printed by the tool itself at `:363-364`, and
  repeated in the commit message and PR comment. `check_control` checks reasons for **region A
  only**, and one of A's three is vacuous per the point above. B, C and E have no reason check at
  all. The sentence claims the property the whole rebuild was for; the check does not establish it.
- **`AGENTS.md:22-23`: "five regions that each produce a **different** verdict"** — five regions,
  four verdicts, A and A′ both `PIXEL_WAS_WRITTEN` (`pixel_history.py:261-262` agrees). The PR
  comment gets this right ("Five regions now, four different verdicts"); `AGENTS.md` does not.
- **`control.c:22`: "E (48,48) two draws, both discard"** — there is one E arm (`:397`). The
  header block is the construction's specification, so it should match it.

## Non-blocking — small

- **`pixel_history.py:32-40` carries two stacked comment blocks** introducing `PRE_FRAGMENT`: the
  old one and its replacement. Leftover from the edit, in the file's most-read comment.
- The `-c` recipe is correct — `_frame%u` is confirmed at `core.cpp:1671`, so `frame_frame*.rdc`
  matches. Note the glob can expand to several files and the tool takes one positional; it fails
  loudly, so this is only a note.
- Still open from last round and still non-blocking: integer render targets read through
  `col.floatValue` (`:115-118`), the bicubic-downsample "brightest pixel" (`:188-197`) whose
  reported luma belongs to the block rather than the pixel, and MSAA (`SaveTexture` defaults to the
  all-sample mapping while `PixelHistory` reads sample 0 at `:207`). On your MSAA question
  specifically: yes, `passed[-1]` can still mislead there — sample 0's history and the resolved
  pixel are different pixels — but that is a documentation gap, not a `classify()` bug.

---

## Verdict

**REJECTED**, on the clear-event item alone. Everything else above I would waive; none of it would
have held the PR by itself, and the E-arm and documentation points are worth fixing but not worth
blocking.

I am blocking on the one item because it is the tool producing a confident wrong answer on its
headline input, because the PR has already written down that it does, and because the fix is a
two-line traversal change plus a `classify()` branch that CI can reach — the same shape as the
suppression lift you already made for exactly this reason. It also makes the control stronger rather
than smaller.

The previous six findings are closed properly and I checked each one rather than taking the summary
on trust. This is one small change from an approval.

mattias800:
## The blocker was right, and it inverts the defect from the last round — fixed in 07359b8a6

I wrote trap 269 and then didn't act on it in the code. Reproduced your measurement exactly:

```
cleared to black,     nothing drew -> SHADER_WROTE_BLACK   (should be NOTHING_DREW)
cleared to sky blue,  nothing drew -> PIXEL_WAS_WRITTEN
```

That first line is the tool's headline input — `DEBUGGING_WORKFLOWS.md` routes "black world"
straight here, and a black region of a cleared target is *the* case people arrive with. It sent them
to a shader that never ran.

Clears are now identified from `ActionFlags.Clear` and are never the subject of a verdict, only the
ground one is stated against. `NOTHING_DREW` means no **draw** touched the pixel. A clear landing
*after* the last surviving draw gets its own verdict, `CLEARED_AFTER_DRAW` — blaming the shader for
a pixel a later clear wiped is the same error one ordering along.

**Your point about the control was the sharpest part of the review, and I'd got it backwards.** The
previous commit "fixed" region E by moving the ground clear out of the capture — making the control
pass by removing the hard case from it, when the tool has to survive this on every real frame. The
clear is back inside, and the guard is now demonstrated rather than asserted:

```
$ # revert clear-awareness, re-run against the real capture
CONTROL FAILED: E  all killed: expected ALL_REJECTED, read SHADER_WROTE_BLACK
MUTATED rc=1
```

## The two check gaps

- **E's `ALL_REJECTED` was reachable with its own draw deleted** — seven scissored neighbours produce
  it alone and the C self-check sees black either way, so both layers stayed green with the arm the
  region exists for gone. It now requires E's own `shaderDiscarded`, *and* requires a clear in E's
  history so the region genuinely guards 269.
- **A's `scissorClipped` requirement was vacuous.** Now requires one *before* A's last surviving
  draw, which only arm 2 can supply. The fixture that tested it was positional and silently
  retargeted when the clear shifted the indices — fixed to target by event id.

## The overstated claims

All four were real:

- **`git diff --check` clean was false.** Bare `git diff --check` reads the working tree, which is
  empty after a commit — so I was checking nothing. Against the base it flagged genuine trailing
  whitespace at `control.c:293`, moved not fixed from last round. Removed, and verified with
  `git diff --check origin/main...HEAD` post-commit.
- **Two of six return values appeared in no user-facing text.** Docstring, `AGENTS.md` and
  `DEBUGGING_WORKFLOWS.md` now list all seven (the two you found plus `CLEARED_AFTER_DRAW`).
- **"every rejection reason read back correctly"** was region A only; the message now names which
  regions it covers.
- `AGENTS.md` said five regions produce five different verdicts (four), and `control.c` said region E
  had two draws (one).

## Verification

27/27 unit tests, including both cleared-target cases from your measurement asserted as verdicts.
Live `--expect-control` green across all five regions with the clear present and labelled `CLEAR`.
Trap-269 self-guard confirmed by mutation against the real capture. `git diff --check` clean against
the base, checked the correct way this time.

Re-review welcome on 07359b8a6.

mattias800 (commented):
## Round three on `07359b8a6` — REJECTED, on one item

Reading review (no builds, no CI, no GPU). Everything I raised in round two is closed, and closed
properly — I checked each rather than taking the summary on trust:

- `classify()` is clear-aware, both my measured cases are unit-asserted, 27/27 green.
- `check_control` now requires E's own `shaderDiscarded` **and** a clear in E's history, and A's
  `scissorClipped` must precede A's last surviving draw — which only arm 2 can supply. Both of my
  gaps are shut, with discriminating tests, and the positional fixture is now keyed by event id.
- The control exhibits the phenomenon again, and you mutated the live tool against the real capture
  to prove the guard bites. That is the right instinct and it is the strongest evidence in the PR.
- All four overstated claims fixed. `git diff --check origin/main...HEAD` is genuinely clean now;
  all seven verdicts appear in the docstring, `AGENTS.md` and `DEBUGGING_WORKFLOWS.md`; the
  `CONTROL VERIFIED` line names its scope; the region/verdict count and "two draws" are corrected.

I am blocking on one thing, and I am sorry to be the third round on it, because it is the same
structural shape as last time rather than a new class of problem.

---

## BLOCKING — `ActionFlags.Clear` does not cover a renderpass `loadOp` clear, which is how real frames clear

`pixel_history.py:173` derives `is_clear` from `a.flags & rd.ActionFlags.Clear`. **RenderDoc does not
set that flag on `vkCmdBeginRenderPass`, even when `loadOp == VK_ATTACHMENT_LOAD_OP_CLEAR`** — but it
*does* record the clear, and its pixel history *does* emit a passing event for it. The two halves are
fifteen lines apart in the same function:

```
vk_cmd_funcs.cpp:2329-2338   loadOp == VK_ATTACHMENT_LOAD_OP_CLEAR
                               -> EventUsage(curEventID, ResourceUsage::Clear)
vk_cmd_funcs.cpp:2351        action.flags |= ActionFlags::PassBoundary | ActionFlags::BeginPass
```

No `ActionFlags::Clear` anywhere in `vk_cmd_funcs.cpp` — `grep -rn "ActionFlags::Clear"
renderdoc/driver/vulkan/` returns only `vk_draw_funcs.cpp` (`vkCmdClearColorImage` `:2398`,
`vkCmdClearDepthStencilImage` `:2510`, `vkCmdClearAttachments` `:2622-2628`). And the bits are
disjoint — `Clear = 0x0001`, `PassBoundary = 0x2000`, `BeginPass = 0x400000`
(`replay_enums.h:5098`, `:5110`, `:5122`) — so the mask genuinely yields 0.

**RenderDoc's own predicate is the usage, not the flag**: `vk_pixelhistory.cpp:4636` and `:4686` are
both `bool clear = (events[ev].usage == ResourceUsage::Clear);`. So RenderDoc pushes the
BeginRenderPass as a clear modification with every test flag false and `Passed()` true, and the tool
looks it up in a set that was built from a different predicate and concludes it is a draw.

**This PR has already measured that a loadOp clear appears as a passing pixel-history event.** On
head `34c3827ee` the control used `att[0].loadOp = VK_ATTACHMENT_LOAD_OP_CLEAR`, and
`CONTROL_EXPECTED` carried **six** entries for five draws — the leading `("passed", [])` was that
clear — and it read back `CONTROL VERIFIED` live. Head `2eb28aa52`'s control comment states it
outright: *"a render-pass clear is itself a pixel-history event that passes at every pixel"*.

### What it costs

Every behaviour this commit adds is keyed on `is_clear`, so all of them fail on the dominant clear
form:

- `drawn = [e for e in events if not e["is_clear"]]` (`:89`) keeps a loadOp clear in `drawn`. On a
  loadOp-cleared black target that nothing drew to, `passed == [that clear]`, `postMod` is black, and
  the verdict is **`SHADER_WROTE_BLACK`** — the exact string this commit exists to stop printing.
- `NOTHING_DREW` stays unreachable on any loadOp-cleared target.
- `CLEARED_AFTER_DRAW` (`:106-111`) cannot fire for a mid-frame renderpass that clears a target
  drawn to earlier — the case it was written for.

And the control cannot see any of it, because `control.c:239` is `att[0].loadOp =
VK_ATTACHMENT_LOAD_OP_LOAD` with a `vkCmdClearColorImage` at `:455` — the one clear form that *is*
flagged. So the mutation you ran proves the guard works on the covered path and is silent on the
uncovered one. That is round two's finding at the next level in: the control validates the case the
code handles.

### The fix is smaller than the code it replaces

Use RenderDoc's own predicate, per target:

```python
clear_eids = {int(u.eventId) for u in ctl.GetUsage(tex.resourceId)
              if u.usage == rd.ResourceUsage.Clear}
```

`ReplayController::GetUsage(ResourceId)` is `renderdoc_replay.h:1059`; the enum is
`replay_enums.h:3079`. This is character-for-character `vk_pixelhistory.cpp:4636`, so it cannot drift
from the thing it is modelling, it is per-target rather than global, and it deletes the `clear_eids`
bookkeeping from `walk()` entirely. It has to move below target selection, which is the only cost.

Then switch `control.c:239` back to `VK_ATTACHMENT_LOAD_OP_CLEAR` (keeping or dropping the transfer
clear as you like) so the guarded form is the one real captures use. Your existing mutation
methodology then covers it unchanged.

---

## Non-blocking — your three questions

**1. Is `eventId` a reliable ordering proxy?** Within RenderDoc's model, yes, and you are not relying
on anything weaker than RenderDoc itself. Ids are assigned as the frame is serialised in submission
order, and `PixelModification::operator<` sorts by `eventId` first, so `hist` already arrives in that
order — the comparison at `:106-107` agrees with the array order rather than second-guessing it. It
is *not* a proxy for GPU execution order across **queues**: RenderDoc linearises multi-queue submits
into one timeline, so a clear submitted on another queue can execute after a draw while carrying a
lower id. But RenderDoc's pixel history has that same limitation — it replays linearly — so the tool
inherits it rather than introducing it, and there is nothing better available through this API.
Worth one sentence in the docstring; not worth code.

**2. Is a seventh verdict right?** Yes. `CLEARED_AFTER_DRAW` has a distinct next step — look at pass
ordering and target reuse, not at shading — which is the test this project's verdict set should meet.
Folding it into `NOTHING_DREW` would assert nothing drew, which is false; into `PIXEL_WAS_WRITTEN`
would assert a colour survived, also false. Both would discard the one actionable fact. Seven tokens
is a lot for "one call, one answer", but each now maps to a different file and each is documented, so
the set has not become decorative.

One hardening worth considering: `late` fires on any passing clear with a higher id, without checking
that the clear *changed this pixel*. `PixelModification.preMod` (`data_types.h:2691`) is right there
and the tool does not capture it; `preMod == postMod` on the clear would say it did not touch the
pixel. I have not established that RenderDoc includes a non-covering partial-rect
`vkCmdClearAttachments` in a pixel's history — I am flagging it as unverified, not as a finding — but
the `preMod` guard is correct either way and costs one field.

**3. Other holes in the clear-awareness?** Two you named are fine: a partial-rect
`vkCmdClearAttachments` *is* flagged (`vk_draw_funcs.cpp:2622-2628`), and a driver implementing a
clear as an internal draw is invisible to RenderDoc's API-level capture, so the action stays a Clear.
An application that clears by drawing a fullscreen quad is genuinely a draw and being treated as one
is correct, not a hole. The loadOp case above is the only one I found, and it is the common one.

## Non-blocking — trap 269's `*Fix:*` now describes a reverted change

The row still reads:

> *Fix:* the control now submits its ground clear before `StartFrameCapture`, so the captured frame
> is draws-only

That is the move this round established was backwards, and it has been undone — the clear is back
inside at `control.c:455`. The shipped fix is that `classify()` treats clears as ground, and the row
does not mention it. Someone following trap 269 today would be told to do the wrong thing, in the
project's most-cited artefact.

While editing it: the headline names `vkCmdClearColorImage` specifically, which is the narrow reading
that produced the blocker above. It is worth saying **`ResourceUsage::Clear`**, and that
`ActionFlags::Clear` is *not* the same set — a loadOp clear is one and not the other. That sentence
would have prevented this round.

## Non-blocking — small

- `pixel_history.py:91`: `cleared = "..." if events else ""` — `events` is guaranteed non-empty by
  the early return at `:82`, so the conditional is dead and the string is unconditional.
- Still open and still non-blocking from earlier rounds: integer targets read through
  `col.floatValue`, the bicubic-downsample "brightest pixel" whose reported luma belongs to the
  block rather than the pixel, and MSAA (sample 0 vs the resolve). All documentation gaps rather
  than logic bugs.

---

## Verdict

**REJECTED**, on the `loadOp` clear alone. Everything else in this round is right, and the
`check_control`, docs and claim fixes are all exactly what I asked for.

I am blocking because the commit's own thesis — *"every real frame clears its targets"* — is true and
the predicate does not cover how they do it, so the headline behaviour works on the control and not
on a real capture. The change is one expression that is shorter than what it replaces and is
RenderDoc's own test, plus one line in `control.c` to point the guard at the common form. With those
two, and the trap-269 row corrected, I will approve immediately.

mattias800 (commented):
## Round four on `11b96661a` — APPROVED

Reading review (no builds, no CI, no GPU). The blocker is fixed correctly, and I traced every claim
to the source rather than taking the summary on trust. This is clean. Two follow-ups below that I do
**not** consider blocking and that I think should be filed as issues rather than held for a fifth
round.

Verified independently on this head: 28/28 tests; docs gate `1..269`, 331 rows, no deletions, and the
whole-repo `.md` pass is green; `git diff --check origin/main...HEAD` genuinely clean. Every line
number cited in the code comments, the commit message and trap 269 says what it is claimed to say.

---

## Your four questions, answered from source

### 1. `GetUsage` semantics — the two lists cannot disagree, by construction

This is stronger than "they happen to match". `ReplayController::PixelHistory` builds its own event
list by calling **the same function**:

```
replay_controller.cpp:1493   rdcarray<EventUsage> usage = m_pDevice->GetUsage(id);
replay_controller.cpp:1553   case ResourceUsage::Clear:   // "writing - include in pixel history events"
replay_controller.cpp:1577   ret = m_pDevice->PixelHistory(events, id, x, y, subresource, typeCast);
```

and `vk_pixelhistory.cpp:4686` then derives `bool clear` from `events[ev].usage`. So `clear_eids` and
RenderDoc's own clear predicate are computed from one list by one rule. There is no third source that
could drift.

The three sub-cases you asked about:

- **Secondary command buffers: included, with correct ids.** `vkCmdExecuteCommands` copies the
  secondary's usage into the parent and rebases it —
  `vk_cmd_funcs.cpp:5546-5547`: `parentCmdBufInfo.resourceUsage.push_back(...)` then
  `.back().second.eventId += parentCmdBufInfo.curEventID`.
- **Ids outside the current `SetFrameEvent`: over-inclusion only, and harmless.**
  `replay_controller.cpp:1497-1499` drops `usage[i].eventId > m_EventID` when building the pixel
  history; your raw `GetUsage` keeps them. Since `is_clear` is a membership test against ids that
  actually appear in `hist`, a surplus id is never consulted.
- **A `Clear` usage with no pixel-history modification: yes, possible, also harmless.** A renderpass
  whose `renderArea` misses the probed pixel, or a partial-rect `vkCmdClearAttachments`, records the
  usage without necessarily producing a modification for that pixel. Same reasoning — surplus ids are
  never looked up. The dangerous direction, an id present in `hist` that RenderDoc treated as a clear
  and you did not, is impossible given the shared list.

One thing I checked because it would have been silent if wrong: the usage push at
`vk_cmd_funcs.cpp:2333` uses `curEventID`, and `AddEvent()` (`vk_core.cpp:6688-6694`) *reads*
`curEventID` as the event id without incrementing it. So the usage entry and the action carry the
same id — no off-by-one, and the transfer/loadOp ids you report (4 and 10) are consistent with that.

### 2. Requiring two clears in E — not brittle, and correctly one-sided

The code is `if len(clears) < 2`, i.e. **at least** two, so a third recognised clear does not fail it.
On the two mechanisms you worried about:

- **Per-attachment reporting is already the behaviour and does not inflate the count.** The loadOp
  loop at `vk_cmd_funcs.cpp:2327-2339` pushes one entry *per attachment*, but keyed by that
  attachment's **image** id — and `GetUsage` is called with the colour target's id, so the depth
  attachment's entry never appears in this set. If one image were somehow bound to two attachments the
  ids would be identical and the set dedupes.
- **Coalescing cannot happen below RenderDoc.** Capture is at the API level, so a driver merging two
  clears internally is invisible; two API calls stay two events.

That leaves "a future RenderDoc records this differently", which is precisely what a known-answer
control *should* fail on, loudly. The message names the mechanism and both detection paths, so
whoever hits it will know what changed rather than concluding the tool is broken. This is right as
built.

### 3. `preMod` on a clear — real, and it fails safe when it is not

Populated, unconditionally on event kind: `vk_pixelhistory.cpp:4767-4768` fills both from the
readback (`FillInColor(fmt, ei.premod, mod.preMod)` / `ei.postmod`) for every history entry. Nothing
in that path is gated on the event being a draw.

The one default path is `eventIndex == -1` — "There is no information, skip the event"
(`:4749-4757`) — which calls `SetInvalid()` on `preMod`, `postMod` **and** `shaderOut`. `SetInvalid()`
writes the same sentinel into each (`data_types.h:2582-2586`:
`col.uintValue[0] = 0xdeadbeef; col.uintValue[1] = 0xdeadf00d;`), so `preMod[:3] != postMod[:3]` is
**False** and `CLEARED_AFTER_DRAW` declines to fire on an event with no data. Your hardening fails in
the conservative direction in exactly the case where it has nothing to go on. Good.

### 4. Overstatement — none found

I checked the commit message, `AGENTS.md` and trap 269 against the source line by line. Trap 269's
citations (`vk_cmd_funcs.cpp:2336`, `:2351`, `vk_pixelhistory.cpp:4636`, `ActionFlags::Clear = 0x0001`,
"set solely by `vk_draw_funcs.cpp`") are all correct, the `\|` escape is present, and both gates pass
on it. The self-critical parts — "it was validating the path the code handled", "a control adjusted
until it agrees with the code is no longer evidence about the code" — are, if anything, the most
valuable sentences in the row, and they are accurate.

The dark-red transfer clear is a genuinely good touch: it means a mistake in that direction cannot
hide behind a coincidentally correct black.

---

## Two follow-ups — file as issues, do not hold the merge

### A. Copies, blits and resolves are the same class as trap 269, one usage over

`IsDirectWrite` (`vk_pixelhistory.cpp:110-115`) covers `CopyDst`, `Copy`, `Resolve`, `ResolveDst`,
`GenMips` and the `*_RWResource` usages, and `replay_controller.cpp:1551-1558` includes all of them as
pixel-history events. They are pushed by the same branch as clears
(`vk_pixelhistory.cpp:4689-4700`, `(clear || directWrite)`) with every test flag false, so `Passed()`
is true — and `is_clear` is false, so `classify()` treats them as draws. A final `vkCmdBlitImage` or
`vkCmdCopyImage` into a target that lands black reads `SHADER_WROTE_BLACK`, blaming a shader that
never ran.

Two things make this much narrower than the clear case, which is why I am not blocking: clears sit at
the head of *every* target's history in every frame, whereas a copy into a colour target is
occasional; and your target selection now prefers `GetPipelineState().GetOutputTargets()` at the last
draw, which steers away from the swapchain image where the final blit usually lands.

Worth noting for whoever picks it up: the fix is **not** "exclude `directShaderWrite`". A compute/UAV
write genuinely is shader work and `SHADER_WROTE_BLACK` is the right answer for it. The line is
`Copy`/`Resolve`/`GenMips` versus `*_RWResource`, and `GetUsage` already gives you both — the map you
just built is the whole mechanism.

### B. The `0xdeadbeef` sentinel reaches `classify()` through the normal path

Same `SetInvalid()` as in question 3, but read as colour rather than compared. `values()` reads
`col.floatValue`, so `0xdeadbeef` is about `-6.26e18` and `0xdeadf00d` similar; `max(postMod[:3])`
picks the zeroed third component and returns `0.0`. An event RenderDoc explicitly marked "there is no
information" is therefore reported as `SHADER_WROTE_BLACK` with full confidence. Rare, but it is the
same family as `OUTPUT_UNTRUSTED`, where you already chose refusal over a guess. Detection is one
comparison: `col.uintValue[0] == 0xdeadbeef and col.uintValue[1] == 0xdeadf00d`.

Also still open and still cosmetic: `pixel_history.py:91`'s `if events` is dead (guaranteed non-empty
by the early return above it); and the older documentation gaps — integer targets read through
`col.floatValue`, the bicubic-downsample "brightest pixel" whose reported luma belongs to the block
rather than the pixel, and MSAA sample 0 versus the resolve.

---

## Verdict

**APPROVED** on `11b96661a`.

Plainly: this is clean. The `ResourceUsage` keying is the correct predicate and is now the same one
RenderDoc uses, the control contains the hard case instead of avoiding it, the `preMod` hardening
fails safe, the two-clear check is correctly one-sided and names its mechanism, and I could not find
an overstated claim anywhere in the new text.

Across four rounds every finding was checked against source before being acted on, and three of them
turned out to be defects in the class the tool exists to prevent — which is a reasonable argument
that the tool is worth having. Please open issues for the two follow-ups above so they are not
carried only in this thread.

Issue #3405, opened 2026-09-06
Sonic Frontiers runs at 3.3 fps on a default launch — the cost is serial single-core host work, not the GPU


mattias800:
Measured 2026-09-06 on `97173eda0` (RADV STRIX_HALO), default launch, no throttles, no snapshot acceleration:

```
PROSPER_RENDER=1 PROSPER_GUEST_ARGS=-force-gfx-direct \
    prosper-app --dump <DUMP_ROOT>/PPSA03831-app0 --fps
```

**3.28 fps** (1145 presented frames / 349 s). F8 capture agrees independently: `guest flips=3.39/s host-presented=3.39/s`.

## It is neither GPU-bound nor CPU-bound

From the F8 report over a 5.02 s window:

| | |
| --- | --- |
| `gpu-device` | **272.1 ms — 5.4% of wall clock** |
| main thread on `drm_syncobj_array_wait_timeout` | **2.1%** (`wait_profile`) |
| process CPU | **1.01 cores** across 65 threads |
| measured work | graphics 3031.8 ms + compute 1397.4 ms ≈ 88% of wall clock |

The GPU is idle and waiting for us. The process is busy but occupies about one core, so it is not compute-saturated either.

## Where the time goes

```
components: renderer-resource=1634.7ms readback=838.5ms compute=1397.4ms
            gpu-wait=347.5ms gpu-device=272.1ms
  build_resources  (frontend):  490.1ms  [texture=486.4]  texture by outcome: rtt=481.4
  setup_resources  (backend):  1144.7ms  [texture=1120.5  buffer=17.5  descriptor=5.6]
```

Compute is dominated by one program, `0x111503adecbc99bf` at `0x20002fe800` — 35 dispatches, 1150.1 ms, of which **setup=982.7 ms and shader=14.1 ms**. A 70:1 setup-to-work ratio.

Across all 188 compute records in the capture: `setup_ms` **1012.0 ms of 1398.5 ms total (72%)**, against `gpu_shader_ms` **28.3 ms**.

## The dominant cost has no sub-timer

`PROSPER_COMPUTE_PHASE_TIMING=1 PROSPER_COMPUTE_TIMING_CODE=0x20002fe800`, 127 records:

```
total_ms          939.8
setup_ms          727.2   (77.4%)
  setup_buffers_ms  1.4   ( 0.2%)   <- the only sub-timer under setup
writeback_ms      139.3
  writeback_images_ms 106.3
dispatch_ms        71.9
[...]

PR #3406, opened 2026-09-06
render: keep texture staging buffers mapped between frames (Sonic Frontiers +6.3%)

mattias800:
Fixes the dominant host-side cost behind #3405.

## The finding

*Sonic Frontiers* ran at 3.3 fps with `gpu-device` at **5.4%** of wall clock and the main thread only **2.1%** on the GPU fence — the GPU was idle, waiting for us. Reading the profile as **inclusive** rather than self time was what unlocked it:

```
30.26%  asm_exc_page_fault
14.96%    amdgpu_gem_fault            [amdgpu]   <- faults on a GPU BUFFER MAPPING
14.55%    ttm_bo_vm_fault_reserved    [ttm]
12.96%    vmf_insert_pfn_prot
 9.66%    pfnmap_setup_cachemode
52.59%  __memmove (inclusive); 14.03% of it under render_draws_rgba
```

The texture staging path mapped its buffer, memcpy'd into it and unmapped, once per texture per frame. The **memory** was already pooled — `vkAllocateMemory` was usually avoided — but the **mapping** was not, so every reuse built a fresh CPU mapping and faulted all 8192 pages of a 33 MB 4K staging texture back in through amdgpu, then discarded it.

## The change

Staging blocks now carry their mapping, in a store the shared render-memory pool cannot draw from.

**Why a separate store, and not just caching mappings on the existing pool:** a retained mapping on shared memory is unsound. The block returns to the pool, the next consumer issues its own `vkMapMemory` on an already-mapped allocation, and you get garbage. That is not a hypothetical — it is what the first version of this change did, and it **segfaulted `gpu_capture_render_replay`**.

## The second bug, and how it was found

With the dedicated store, `texture_sample_render` still segfaulted — in a `memcpy` nowhere near the new code. Rather than guess a third time I added an env seam that disables **reuse** while leaving the allocation path identical:

```
reuse disabled -> exit 0
reuse enabled  -> segfault
```

That halved the search space and pointed at "something invalidates a reused block". It did: the storage-writeback readback path maps the same staging memory a **second** time (`render_runner.h:10359`) and unmaps it (`:10366`). Against a retained mapping that `vkMapMemory` fails with `VK_ERROR_MEMORY_MAP_FAILED` and the `vkUnmapMemory` destroys the mapping the cache still hands out — a dangling pointer for the next frame's copy. That path now reads through the retained pointer and only maps when the block is not ours.

`PROSPER_NO_MAPPED_STAGING` is kept as a permanent A/B seam. It is what isolated the bug, and it lets both arms be measured **from one binary**, with no build differences between them.

## Measurement

`PPSA03831`, default launch, no throttles or snapshot acceleration, `PROSPER_PERF_CAPTURE_AFTER_MS=90000`, **six runs per arm, both arms the same binary**:

| | guest flips/s | mean | `setup_resources` texture |
| --- | --- | --- | --- |
| reuse off | 3.36 3.36 3.38 3.39 3.39 3.39 | 3.378 | 1104 / 1067 / 1075 ms |
| reuse on | 3.39 3.39 3.59 3.59 3.59 3.99 | **3.590** | **499 / 553 / 509 ms** |

**+6.3% on the mean; every reuse-on run is at least as fast as every reuse-off run**, and the targeted cost is halved with no overlap between arms. The cost reduction is the stronger of the two results — six samples, cleanly separated — and it is the one that directly matches the mechanism.

## Verification

- **Full suite 369/369.** Both bugs above came from tests outside the renderer subset, so the whole suite is the relevant gate here, not a focused subset.
- Reverted-tree control run confirmed 369/369 before the change, so the two failures were genuinely mine and not pre-existing.

## Scope

No guest-visible behaviour change; this is host-side allocation lifetime. Not a complete answer to this title's frame rate — `execute_item` still carries **15.8–21.9% self time** that is not the copies, and `readback` is ~840 ms per 5 s window. Both are recorded on #3405 rather than folded into this claim.

Two hypotheses were falsified on the way to this and are written up in #3405 so nobody re-derives them: `seed_reprove` (A/B'd, zero effect) and the `linear_seed` per-dispatch allocation (implemented, measured, reverted — the profile did not move at all).

---

## Review follow-up (97df5c728), and two corrections to the claims above

Independent review returned **APPROVED** on the lifetime question — every `vkMapMemory`/`vkUnmapMemory`/`vkFreeMemory`/`vkDestroyBuffer` site in the header was enumerated and no third path can reach a cached block. Two findings were raised and are fixed:

1. **`PROSPER_NO_MEMORY_POOL=1` had stopped disabling staging pooling.** Before this change staging came from `allocate_transient_render_memory`, which honours that flag. `docs/GRAPHICS.md` carries `## Ruled out` rows that exonerate "prosper's caches, pools or arenas" using exactly that arm — re-run today they would have tested strictly *less* than when recorded. A silently weakened falsification is worse than a missing one. Reuse is now off when either switch is set.
2. **The cache borrowed `render_memory_pool_limit()` while accounting separately**, doubling peak retained host memory with nothing reporting the second half. It now has `PROSPER_MAPPED_STAGING_MB` (default 256 MiB) and `docs/FRONTEND_APP.md` says the two budgets add. The env-arms gate then failed as designed; the arm is added.

### Corrections to the measurement framing above

**The off arm is not `main`.** Under `PROSPER_NO_MAPPED_STAGING` staging gets no reuse at all, whereas `main` still had shared-pool reuse (without mapping reuse). The off arm therefore pays an allocate+free that `main` did not, which makes **+6.3% an upper bound on the improvement against `main`**, not a measurement of it. The opening line's "3.3 fps" is a `main` figure and sits *below* the off arm's 3.378, which points the same way.

**Median alongside mean**, so the single 3.99 run is visibly not load-bearing: **3.38 → 3.59, +6.2%**.

**"Six runs per arm"** describes the flips/s row; the `setup_resources` column shows three per arm.

### And CI green is not evidence about this change

The `Host ASan + UBSan` and `Host TSan` jobs configure `prosper/tools/doctor` only and never compile this header, and the `Linux` job has no GPU. The 11 green checks say nothing about this change's runtime behaviour. **The local full-suite run — 369/369, including the two tests that caught the earlier attempts — is the evidence.** Re-run green after the follow-up commit.

### DIFF
--- a/BLOG.md
+++ b/BLOG.md
@@ -21,6 +21,16 @@ from the tracker issues, and still gated, because it is a projection of state ra
 
 ## 2026-09-06
 
+### Sonic Frontiers got faster, and the GPU was never the problem
+
+At 3.3 fps the GPU was busy **5.4%** of the time — it was idle, waiting for us. Nearly a third of
+the frame was page faults, and they turned out to be `amdgpu` faults on a *GPU buffer mapping*:
+every texture upload mapped its staging buffer, memcpy'd 33 MB into it, and unmapped, so all 8192
+pages faulted back in on the next frame. Keeping the mapping alive halves the per-frame texture
+binding cost, 1082 ms to 520 ms, and lifts the frame rate about 6%. Two wrong hypotheses died on
+the way, both recorded in [#3405](https://github.com/mattias800/prosper/issues/3405).
+
+
 ### We can now ask a black pixel why it is black, and get an answer
 
 Four causes produce an identical black pixel — nothing drew there, something drew and was rejected,
--- a/prosper/docs/FRONTEND_APP.md
+++ b/prosper/docs/FRONTEND_APP.md
@@ -566,6 +566,15 @@ and budget-driven discards. The default budget is 512 MiB; override it with
 `vkAllocateMemory`/`vkFreeMemory`. The pool retains only allocation objects: images, buffers, views,
 descriptors, and their layout/content rules keep their existing per-call lifetimes.
 
+**Texture staging is a second, separate cache with its own budget.** It retains a staging block
+together with its CPU **mapping**, because dropping the mapping returns the pages and the next write
+faults all of them back in through amdgpu — ~30% of the frame on a 4K upload (#3405). Its budget is
+`PROSPER_MAPPED_STAGING_MB`, default **256 MiB**, accounted independently of the transient pool
+above, so peak retained host memory is the sum of the two rather than the pool figure alone.
+`PROSPER_NO_MAPPED_STAGING=1` disables reuse while leaving the allocation path identical — the
+A/B seam that isolates it — and `PROSPER_NO_MEMORY_POOL=1` disables this cache as well, so that
+flag still means "no staging is pooled" as it did before this cache existed.
+
 Within one already-synchronous backend call, identical immutable Vulkan contracts share objects instead of
 recreating them for every draw. Storage buffers require the same nonzero guest address, size, and complete
 captured bytes; synthetic resources with no identity remain distinct. Texture image views and samplers use
--- a/prosper/tests/diagnostics/test_env_numeric_sites.cpp
+++ b/prosper/tests/diagnostics/test_env_numeric_sites.cpp
@@ -286,6 +286,11 @@ static const Site kSites[] = {
      mib_cap_new<1024>, mib_cap_old<1024>, "eight", 1024ull * kMiB, "2048", 2048ull * kMiB},
     {"render_runner.h PROSPER_MEMORY_POOL_MB", "PROSPER_MEMORY_POOL_MB",
      mib_cap_new<512>, mib_cap_old<512>, "-1", 512ull * kMiB, "256", 256ull * kMiB},
+    // #3405: the mapped-staging cache's own budget, accounted separately from the transient pool
+    // above so peak retained host memory is the sum of the two rather than one figure standing in
+    // for both. A malformed value must keep the default rather than disable the bound.
+    {"render_runner.h PROSPER_MAPPED_STAGING_MB", "PROSPER_MAPPED_STAGING_MB",
+     mib_cap_new<256>, mib_cap_old<256>, "-1", 256ull * kMiB, "512", 512ull * kMiB},
 
     {"render_runner.h PROSPER_BACKEND_BUFFER_ARENA_KB", "PROSPER_BACKEND_BUFFER_ARENA_KB",
      arena_new, arena_old, "4mb", 1024ull * 1024ull, "2048", 2048ull * 1024ull},
--- a/prosper/tests/fixtures/render_runner.h
+++ b/prosper/tests/fixtures/render_runner.h
@@ -2462,6 +2462,144 @@ inline VkDeviceMemory allocate_transient_render_memory(VkDevice device, VkDevice
     return memory;
 }
 
+// #3405: texture staging blocks cached together WITH their CPU mapping.
+//
+// The render-memory pool already avoids vkAllocateMemory, but every consumer still wrapped each
+// use in vkMapMemory/vkUnmapMemory. Dropping the mapping returns the pages, so the next write
+// faults all of them back in through amdgpu -- ~30% of the profile (amdgpu_gem_fault,
+// ttm_bo_vm_fault_reserved, vmf_insert_pfn_prot, pfnmap_setup_cachemode) while memcpy'ing a
+// 33 MB 4K staging texture. Keeping the mapping alive measured +12% guest flips/s.
+//
+// These blocks are deliberately NOT drawn from the shared pool: a retained mapping on shared
+// memory is unsound, because the block returns to the pool and the next consumer issues its own
+// vkMapMemory on an already-mapped allocation.
+//
+// `owned` is the load-bearing part. The cache must be the ONLY code that can free a block it
+// created: a caller that releases the same staging twice would otherwise fall through to the
+// generic teardown, vkFreeMemory the allocation this cache still maps, and leave a dangling
+// mapped pointer for the next acquire to memcpy into.
+struct MappedStagingBlock {
+    VkBuffer buffer = VK_NULL_HANDLE;
+    VkDeviceMemory memory = VK_NULL_HANDLE;
+    void* mapped = nullptr;
+    VkDevice device = VK_NULL_HANDLE;   // #3405: a cached handle is only valid for ITS device
+};
+
+struct MappedStagingCache {
+    std::mutex mutex;
+    // Keyed by device as well as shape: VkBuffer/VkDeviceMemory belong to the VkDevice that
+    // created them, and a process that builds more than one device would otherwise be handed a
+    // block from a destroyed one.
+    std::map<std::tuple<VkDevice, VkDeviceSize, uint32_t>,
+             std::vector<MappedStagingBlock>> free_blocks;
+    std::unordered_map<VkDeviceMemory, std::tuple<VkDevice, VkDeviceSize, uint32_t>> in_use;
+    std::unordered_map<VkDeviceMemory, std::tuple<VkDevice, VkDeviceSize, uint32_t>> owned;
+    uint64_t cross_device_skips = 0;
+    VkDeviceSize cached_bytes = 0;
+    uint64_t double_releases = 0;
+};
+
+inline MappedStagingCache& mapped_staging_cache() {
+    static MappedStagingCache cache;
+    return cache;
+}
+
+template <typename PickMemoryType>
+inline MappedStagingBlock acquire_mapped_staging(VkDevice device, VkDeviceSize bytes,
+                                                 VkBufferUsageFlags usage,
+                                                 PickMemoryType&& pick) {
+    if (!bytes) return {};
+    MappedStagingCache& cache = mapped_staging_cache();
+    const std::tuple<VkDevice, VkDeviceSize, uint32_t> key{device, bytes,
+                                                           static_cast<uint32_t>(usage)};
+    {
+        std::lock_guard<std::mutex> lock(cache.mutex);
+        auto found = cache.free_blocks.find(key);
+        if (found != cache.free_blocks.end() && !found->second.empty()) {
+            MappedStagingBlock block = found->second.back();
+            found->second.pop_back();
+            if (found->second.empty()) cache.free_blocks.erase(found);
+            cache.cached_bytes -= bytes;
+            cache.in_use.emplace(block.memory, key);
+            return block;                  // mapping intact: no fault-in on the next write
+        }
+    }
+
+    MappedStagingBlock block;
+    VkBufferCreateInfo bci{VK_STRUCTURE_TYPE_BUFFER_CREATE_INFO};
+    bci.size = bytes;
+    bci.usage = usage;
+    if (vkCreateBuffer(device, &bci, nullptr, &block.buffer) != VK_SUCCESS) return {};
+    VkMemoryRequirements requirements;
+    vkGetBufferMemoryRequirements(device, block.buffer, &requirements);
+    const uint32_t type = pick(requirements.memoryTypeBits);
+    if (type == UINT32_MAX) { vkDestroyBuffer(device, block.buffer, nullptr); return {}; }
+    VkMemoryAllocateInfo allocation{VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_INFO};
+    allocation.allocationSize = requirements.size;
+    allocation.memoryTypeIndex = type;
+    if (vkAllocateMemory(device, &allocation, nullptr, &block.memory) != VK_SUCCESS) {
+        vkDestroyBuffer(device, block.buffer, nullptr);
+        return {};
+    }
+    vkBindBufferMemory(device, block.buffer, block.memory, 0);
+    if (vkMapMemory(device, block.memory, 0, VK_WHOLE_SIZE, 0, &block.mapped) != VK_SUCCESS ||
+        !block.mapped) {
+        vkFreeMemory(device, block.memory, nullptr);
+        vkDestroyBuffer(device, block.buffer, nullptr);
+        return {};
+    }
+    block.device = device;
+    std::lock_guard<std::mutex> lock(cache.mutex);
+    cache.in_use.emplace(block.memory, key);
+    cache.owned.emplace(block.memory, key);
+    return block;
+}
+
+// True means "this cache owns the allocation; the caller must not free it". That answer is given
+// for a block already back in the free list too, so a duplicate release is a no-op rather than a
+// dangling mapping.
+inline bool release_mapped_staging(VkDevice device, VkBuffer buffer, VkDeviceMemory memory,
+                                   void* mapped) {
+    if (!memory) return false;
+    MappedStagingCache& cache = mapped_staging_cache();
+    std::lock_guard<std::mutex> lock(cache.mutex);
+    if (!cache.owned.count(memory)) return false;          // not ours: caller tears it down
+    // Bisection seam: with this set the block is destroyed rather than retained, so acquire is
+    // exercised exactly as before but nothing is ever reused. It separates "the new allocation
+    // path is wrong" from "reusing a block is wrong".
+    static const bool no_reuse = getenv("PROSPER_NO_MAPPED_STAGING") != nullptr;
+    if (no_reuse) {
+        cache.owned.erase(memory);
+        cache.in_use.erase(memory);
+        vkUnmapMemory(device, memory);
+        vkFreeMemory(device, memory, nullptr);
+        vkDestroyBuffer(device, buffer, nullptr);
+        return true;
+    }
+    auto active = cache.in_use.find(memory);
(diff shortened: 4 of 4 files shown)


mattias800 (commented):
## Independent code review — PR #3406

Reviewed at head `1c424cd000a44ed8ffa33529a216ffdc49882e6b`. Read-only review in a separate worktree; no
rebuild, no test run, no GPU workload (per the review brief). All line numbers are that head's
`prosper/tests/fixtures/render_runner.h` unless stated otherwise.

**Verdict: APPROVED.**

The lifetime question this review was asked to answer comes back clean. I could not find a second path
that maps, unmaps or frees a block the cache owns, and the ordering property that usually breaks this
class of change — reuse gated on submission completion — is preserved. The findings below are real but
none of them is a live defect on this head; two are small enough that I would rather see them land here
than as follow-up, and I say which.

---

## 1. The enumeration (the core of the review)

Every `vkMapMemory` / `vkUnmapMemory` / `vkFreeMemory` / `vkDestroyBuffer` in the file, classified by
whether it can reach a `MappedStagingCache`-owned block.

**Inside the cache itself** — `:2536`, `:2540-2542`, `:2545-2549` (construction failure paths, block not
yet published to any map), `:2574-2576` (the `PROSPER_NO_MAPPED_STAGING` arm), `:2593-2595` (budget
exceeded). All correct: each destroy path erases `owned` **and** `in_use` first (`:2572-2573`,
`:2592` + `:2583`), so no map ever names a dead handle.

**Touching `upload.staging` / `upload.staging_memory`** — the complete set is `:7887` (acquire), `:9026`
and `:9755-9756` (`vkCmdCopyBufferToImage` + host-read barrier; command recording only, no host
mapping), `:10367`/`:10374` (the storage-writeback readback, now gated on `borrowed_mapping`), and
`:10619-10623` (release + fallback teardown). `grep -n 'staging' render_runner.h` returns nothing else
that is a Vulkan call. Attempt 2's bug is genuinely closed: `:10363-10374` reads through
`upload.staging_mapped` and only maps/unmaps when the pointer is null.

**Every other site operates on a handle from a different producer**, none of which the cache can ever
have created:

| sites | owner |
| --- | --- |
| `:2795-2797`, `:2858` | `render_host_buffer_pool` (`RenderHostBuffer`) — a separate pool with its own mapping discipline |
| `:3274-3278`, `:4499-4508`, `:4755-4772`, `:4928-4946`, `:5014-5030` | helper upload/readback functions, buffers created and destroyed locally in the same function |
| `:7049-7050` | index buffer `v.ibmem` |
| `:8804`, `:8831`, `:8877` (+ `:5700`, `:5818`, `:5927`, `:8874`) | render-target seed buffers |
| `:9607`, `:10133-10135` | echo buffer |
| `:10188-10194`, `:10337`, `:9399-9400`, `:10565-10568` | geometry/counter buffers |
| `:10378-10402`, `:10505-10509` | readback buffer `bmem` |
| `:2325`, `:7759`, `:9990` | persistent color-target / persistent-texture **image** memory |
| `:10614` | `upload.memory` (the image), not the staging |

So the only generic sink that could in principle receive a cache block is
`release_transient_render_memory(dev, upload.staging_memory)` at `:10623`, reachable only when
`release_mapped_staging` returns `false`. Its three false-returning conditions:

- `!memory` (`:2563`) — then the caller's own `if (upload.staging)` / `if (upload.staging_memory)`
  guards are also false, so nothing happens;
- not in `owned` (`:2566`) — unreachable, because `acquire_mapped_staging` is now the sole producer of
  `upload.staging_memory` and it registers (`:2554`) every block it returns non-null;
- the cross-device branch (`:2584-2589`) — see finding 3.

And even if that last one fired, it would not poison the shared pool: the cache's memory comes from a
raw `vkAllocateMemory` (`:2540`) and is therefore never in `pool.active`, so
`release_transient_render_memory` falls straight through to a plain `vkFreeMemory` at `:2612-2616`.
**The "the shared pool cannot draw from this cache" property holds through two independent
mechanisms**, which is the right amount of belt for a change with attempt 1's history.

**The reuse-vs-in-flight ordering is preserved**, which is the other way this goes wrong and which the
PR body does not mention. `release_mapped_staging` is called from the same cleanup closure that used to
call `release_transient_render_memory` (`:10547` registration, `texture_uploads` moved in at `:10554`),
and that closure runs only from `complete()` (`:2043-2048`), after the fence wait. So a block cannot be
handed to a new acquire while the GPU is still reading it.

**Abandon path:** `abandon_pending_resources()` (`:1913-1924`) deliberately leaks cleanup closures on an
unproven completion, so an abandoned frame's staging blocks stay in `in_use`/`owned` forever. That is the
same leak `main` already had for the pooled memory, plus two map entries; `cached_bytes` is unaffected
because it only ever counts free-list blocks. Not a new defect.

**No early exit between acquire and cleanup registration** — `awk` over lines 7880-10548 finds no
`return out;`, `return {};`, `abort` or `throw` at function scope, so there is no path that acquires a
block and never registers its release.

---

## 2. `release_mapped_staging`'s `owned` logic — sound today, but the comment claims more than the code delivers

The guard that actually runs is `in_use.find(memory) == end` (`:2580`). That catches **"released twice
with no intervening acquire"**. It does **not** catch "released, re-acquired, released again by a stale
copy of the handle", which is the case the header comment at `:2477-2480` and `:2558-2560` describes:

1. frame A releases `M` → `in_use.erase(M)`, `M` pushed to `free_blocks[key]`;
2. frame B acquires → pops `M`, `in_use.emplace(M, key)` (`:2523`);
3. a stale frame-A handle releases `M` again → `owned.count(M)` is true, `in_use.find(M)` now **succeeds**
   (frame B repopulated it), so the retain path runs: `M` is pushed into `free_blocks[key]` a *second*
   time at `:2598` while frame B is still memcpy'ing into it, `cached_bytes` is double-added, and
   `double_releases` stays at 0. Two consumers then hold the same mapped pointer.

**This is not reachable on this head** — `texture_uploads` is `std::move`d into a single closure
(`:10554`), the teardown loop visits each element once, and `complete()` clears `cleanups_` after running
them, so every handle is released exactly once. So: no finding against the code, but the invariant the
comment asserts is stronger than the one implemented, and that comment is exactly what the next agent
will trust instead of re-deriving. Either narrow the comment to the guard that exists, or make the guard
match the claim (cheapest: a monotonic generation in `MappedStagingBlock`, compared on release).

Related, and worth one line in the seam's comment: under `PROSPER_NO_MAPPED_STAGING` the block is
disowned at `:2572` before being freed, so the "duplicate release is a no-op" property does **not** hold
in that arm — a second release would return false and the caller would `vkDestroyBuffer` an already
destroyed buffer. The seam arm is strictly less protected than the default arm.

---

## 3. The cross-device branch does the opposite of what its comment says (latent, not live)

`:2584-2589`:

```
if (std::get<0>(key) != device) {
    // ... Do not retain it and do not touch it with this device; just forget it.
    ++cache.cross_device_skips;
    cache.owned.erase(memory);
    return false;
}
```

Returning `false` sends the caller to `:10621-10623`, which runs `vkDestroyBuffer(dev, ...)` and
`release_transient_render_memory(dev, ...)` — i.e. it touches a foreign device's objects **with this
device**, which is precisely what the comment says must not happen. It also leaks the mapping.

**Unreachable today.** `render_vk_ctx()` is a lazy singleton created once at `:1493` and, per the
LIFETIME INVARIANT comment at `:1036`, intentionally never destroyed; `acquire` and `release` are
both handed `dev = ctx.dev` from the same call, so the key's device always matches. The other two
`vkCreateDevice` sites in the tree (`prosper/frontends/prosper-app/main.cpp:288`,
`prosper/frontends/shared/live/live_compute.cpp:3224`) never reach this path. So this is defensive code
that would misbehave if the defence were ever needed. Fix: return `true` (disown and leak deliberately),
or free using the device recorded in the key.

This also answers the "can a block outlive its `VkDevice`" question directly: **no**, because nothing
destroys the render device at all. The device key is used consistently — constructed at `:2513-2514` on acquire, read
at `:2584` on release, and stored redundantly in `MappedStagingBlock::device` (`:2551`, `:2598`) which is
never actually read. That redundancy is harmless but the field is dead; either read it or drop it.

---

## 4. `cached_bytes` accounting: symmetric, no leak, no double-count

Acquire subtracts the function parameter `bytes` (`:2522`), release adds `std::get<1>(key)` from the
`in_use` entry (`:2582`, `:2599`), and `in_use` was populated with the very key the block was filed
under, so the two are the same number by construction and `cached_bytes` cannot drift or go negative.
`owned` is correctly *not* erased on the retain path, which is what makes the cache the sole owner for
the block's whole life.

Two accuracy notes, both minor: the accounting counts the *requested* `bytes`, while the allocation is
`requirements.size` (`:2538`), so real retained memory slightly exceeds the accounted budget; and there
is no eviction or aging, so with the exact-`bytes` keying a title that changes texture shapes fills the
budget with dead buckets and then permanently refuses to cache the shapes it is actually using. The
shared pool has the same weakness (and additionally caps at 4096 allocations, `:2619`, which this cache
does not), so neither is a regression — but the second one is a real stranding mode and is worth a
sentence in the header comment so nobody debugs it from scratch.

---

## 5. Budget: `render_memory_pool_limit()` now governs two independent caches (please fix here — a few lines)

`:2591` reuses `render_memory_pool_limit()` — default 512 MiB, `PROSPER_MEMORY_POOL_MB` — as the new
cache's budget, but with its own `cached_bytes`. So peak retained host-visible memory at the default is
now **1 GiB across the two caches**, and `PROSPER_MEMORY_POOL_MB=<N>` means "N MiB, twice".

`prosper/docs/FRONTEND_APP.md:562-566` still describes this as one pool with one 512 MiB budget, and the
`[render-timing] memory_pool` line at `:10819-10826` reports `render_memory_pool_stats()` only — so the
diagnostic that used to account for the *largest single consumer* of that budget silently stopped
covering it, and nothing reports the new cache at all. That is the shape `tools/env/check_diag_gates.py`
exists to catch: a well-formed number that quietly narrowed what it measures.

Suggested: give the staging cache its own budget constant (or divide the existing one explicitly), add a
`[render-timing] mapped_staging` line beside the pool line, and correct the FRONTEND_APP.md paragraph.

---

## 6. `PROSPER_NO_MEMORY_POOL=1` no longer disables staging pooling (please fix here — one line)

On `main`, staging memory came from `allocate_transient_render_memory` / `release_transient_render_memory`,
both of which honour `render_memory_pool_enabled()` (`:2435`, `:2605`). After this change staging never
touches that pool, and the new cache consults **only** `PROSPER_NO_MAPPED_STAGING`. So
`PROSPER_NO_MEMORY_POOL=1` — documented at `docs/FRONTEND_APP.md:565` and
`docs/WINDOWS_PORT_HANDOFF.md:226` as "a direct allocate/free A/B run" — now leaves a 512 MiB
retained-and-mapped staging cache fully live.

This is not hypothetical bookkeeping: `docs/GRAPHICS.md:783` and `:1050` both record `## Ruled out` rows
that used `PROSPER_NO_MEMORY_POOL` to exonerate "prosper's caches, pools or arenas" and "the host-visible
upload path". Re-run on this head, those arms test strictly less than they did when they were recorded,
and nothing announces the change. One line fixes it — treat `!render_memory_pool_enabled()` as implying
`no_reuse` at `:2570`.

---

## 7. Smaller items

- **Dead telemetry.** `cross_device_skips` and `double_releases` are incremented and read *nowhere* in
  the tree (`grep -rn` over `src/ tests/ tools/ frontends/` finds only the increments). The counter whose
  entire purpose is to make a rare, otherwise-silent bug visible is itself invisible. Print both with the
  pool stats at `:10819`.
- **Unrated-limit diagnostic.** The `[stage] staging map failed` `fprintf` at `:7900` has no rate limit,
  unlike its immediate sibling at `:7913-7914` (`declined_logged++ < 16`). On a host where these allocations
  start failing it fires once per texture per frame — the charter's 1.5 GB-log failure mode.
- **`:10365-10370` is now dead code.** Reaching it requires `upload.staging != VK_NULL_HANDLE`, and the
  only producer of that handle returns `{}` (all-null) whenever the map fails, so `borrowed_mapping` is
  always true there. Harmless as a backstop; add a one-line comment saying so, or a future reader will
  assume the `vkMapMemory` is live and reason from it.
- **No focused test.** Nothing in `tests/` names the cache. Both prior bugs were found by full-suite
  segfaults rather than by an assertion, and the invariants the comments state (reuse returns the same
  mapped pointer; duplicate release is a no-op; the budget branch destroys rather than retains) are all
  cheaply testable without a GPU-heavy route. Given the header is shared by the live renderer *and* the
  whole test suite, I would land at least the reuse-identity and double-release cases.
- **`emplace` vs assign.** `:2553-2554` use `emplace`, which silently does nothing if a stale key exists.
  Unreachable while there is one never-destroyed device, but `insert_or_assign` would fail loudly instead
  of quietly filing a block under the wrong key.

---

## 8. Thread safety — sound

Every `MappedStagingCache` field access is under `cache.mutex`: the free-list hit path at `:2516-2524`,
and the publish at `:2552-2554`; `release_mapped_staging` holds the lock for its whole body (`:2565`).
The one field written outside the lock is `block.device` at `:2551`, on a function-local object not yet
visible to anyone.

`vkMapMemory` at `:2545` is deliberately outside the lock. The only Vulkan calls made *under* it are in
the no-reuse (`:2574-2576`) and budget-exceeded (`:2593-2595`) branches. Nothing else in the tree takes
`cache.mutex`, so there is no lock-order cycle against `render_memory_pool::mutex`, the host-buffer pool
mutex, or any other prosper lock — a driver lock taken under `cache.mutex` is never waited on by a thread
holding `cache.mutex`. The function-local `static const bool no_reuse` at `:2570` is initialised under the
mutex, but its initialiser is a bare `getenv` that takes no prosper lock, so the magic-static guard cannot
deadlock either.

Worth stating for the record so nobody misreads the green checks: **CI's `Host TSan` and
`Host ASan + UBSan` jobs configure `prosper/tools/doctor` only** (`.github/workflows`, lines 23-31 and
46-54) and never compile this header, and the `Linux` job has no GPU. So the 11 green checks are not
evidence about this change's runtime behaviour; the author's local 369/369 full-suite run is.

---

## 9. `PROSPER_NO_MAPPED_STAGING` seam — disables reuse correctly, but is **not** an A/B against `main`

The seam does what the PR says *within the new code*: `:2571-2578` destroys instead of retaining, so
`free_blocks` is never populated, every acquire misses, and the allocation path at `:2528-2555` runs
identically in both arms. That part is correct, and it is the right instrument.

But it is not a stand-in for `origin/main`, and the PR body reads as if it were. On `main` the staging
*memory* was pooled (`allocate_transient_render_memory`), only the *mapping* was rebuilt each frame.
Under the seam, `acquire_mapped_staging` calls `vkAllocateMemory` directly (`:2540`) and the seam calls
`vkFreeMemory` directly (`:2575`), so the reuse-off arm pays a real per-texture-per-frame allocate+free
that `main` did not. **The off arm is therefore worse than the code being replaced, and +6.3% is an upper
bound on the gain over `main`, not an estimate of it.**

The PR's own numbers hint at this: the opening says *Sonic Frontiers* "ran at 3.3 fps" on `main`, but the
reuse-off arm measures 3.378 — the wrong direction for that story. Most likely the 3.3 figure comes from
a different session and is not a controlled comparison at all, which is fine, but the body should say so
rather than let the two numbers sit adjacent.

**On the statistics themselves the body is fair.** I recomputed from the table: off mean 20.27/6 =
3.3783, on mean 21.54/6 = 3.590, ratio 1.0627 → +6.3% ✓. "Every reuse-on run is at least as fast as every
reuse-off run" is exactly true and exactly worded — min-on and max-off are both 3.39, so it is a ≥ and the
body does not claim >. Good discipline; a lot of write-ups would have said "faster".

Two things I would add or change, because they are the obvious objections:

- **Give the median.** The on-arm mean is visibly pulled by one 3.99 outlier. Medians are 3.38 → 3.59,
  i.e. +6.2%, so the outlier is *not* load-bearing and the headline survives without it. Saying that
  pre-empts the objection instead of inviting it.
- **"six samples, cleanly separated"** describes the `setup_resources` cost reduction, but the table shows
  **three** figures per arm for that metric, not six. Either read is defensible from the sentence, which
  is the problem. Reword to "three per arm, no overlap".

The mechanism claim itself is well supported and is the stronger half of the case: a halving of a
specific, named, independently-instrumented cost (1082 → 520 ms) with no overlap between arms, matching a
mechanism identified from a profile (`amdgpu_gem_fault` / `ttm_bo_vm_fault_reserved` /
`vmf_insert_pfn_prot`) that predicts exactly that leaf and no other. That is a much better argument than
the flips/s delta, and the body is right to say so.

---

## Summary

Nothing here blocks. The retained mapping's lifetime is correct, the ownership boundary against the
shared pool is doubly enforced, the completion-gating that makes reuse safe is intact, the mutex
discipline is complete, and no path outside the cache can map, unmap or free a block it owns.

Items 5 and 6 are each a few lines and both narrow an existing diagnostic or bisection seam without
announcing it, which is the failure mode this project's charter is most emphatic about — I would land
them in this PR. Items 2, 3 and 7 are latent or cosmetic and are fine as follow-ups. Item 9 asks for
wording changes to the PR body, not for more runs.

**APPROVED** on `1c424cd000a44ed8ffa33529a216ffdc49882e6b`.

PR #3408, opened 2026-09-06
compute: parallelise the sampled texel-conversion arms (Sonic Frontiers +18.7%)

mattias800:
Follow-on to #3406, addressing the dominant cost identified in #3407.

## The finding

The sampled conversion chain has **seven** per-texel arms. Exactly **one** used `parallel_compute_texels`; the other six were plain scalar loops.

That reconciles two measurements on Sonic Frontiers' intro that never fitted together: **1630 ms of conversion work in a 4760 ms window, while the whole process used 1.01 cores.** A 4K conversion was running on one core of a 32-core machine.

The helper has existed all along, is used in eight other places in this same file, and its own comment describes precisely this failure on a different title:

> *"Astro Bot's 4K FP16 target is 8.3 million texels, so leaving its lookup/pack walk on one core made a ~5 ms GPU dispatch wait on hundreds of milliseconds of host work."*

The sampled arms just never got the same treatment.

## Why it is safe

Every write in these loops is `upload[t * n + c]` — disjoint across `t`. They are parallel by construction, and identical in shape to the arm immediately below them that was already wrapped.

The one way wrapping a loop in a lambda *can* change meaning is control flow: a `break` that used to leave the whole loop would now leave only its chunk, with the other threads continuing. There are `break` statements in this region, so I checked each wrapped body mechanically rather than by eye — all six are free of `break`, `continue` and `return`; those `break`s belong to the enclosing branch chain. No wrapped body has cross-iteration state either.

## Measurement

`PPSA03831`, default launch, no throttles, three runs per arm, **both arms from the same binary** — `PROSPER_COMPUTE_CONVERSION_THREADS=1` forces the old scalar path, so there is no build difference between them:

| arm | guest flips/s | mean | compute time |
| --- | --- | --- | --- |
| scalar (`THREADS=1`) | 3.59, 3.39, 3.58 | 3.52 | 1792.7 / 1800.6 / 1803.9 ms |
| parallel (default) | 4.38, 4.38, 3.79 | **4.18** | **907.6 / 890.3 / 869.7 ms** |

**Compute time halved; frame rate +18.7%.** `process CPU` rises 0.96 → 1.29 cores rather than toward 16, because these loops are memory-bandwidth-bound — which is also why the helper caps at sixteen workers.

This is the largest single improvement measured on this title, and it is a **scheduling** fix rather than an algorithmic one: the same work, spread across cores that were already sitting idle.

## Verification

- Full suite **369/369**.
- Mechanical check that no wrapped body contains `break`/`continue`/`return`.
- A/B from one binary, three runs per arm; every parallel run is faster than every scalar run except one (3.79 vs 3.59), and the compute-time column is cleanly separated with no overlap.

## Scope, and what this does not fix

This does not remove the conversion, it spreads it. #3407 records the two changes that would remove it: caching the ~90% of conversions that reconvert identical bytes, and the fact that prosper has **no GPU tiling or format-conversion path at all**, so renderer-owned surfaces round-trip through the CPU.

For scale: with this landed the intro is ~4.2 fps against a 30 fps target, and `gpu-device` remains ~6% of wall clock. The GPU could already run this above 60 fps; the remaining work is architectural.

### DIFF
--- a/prosper/frontends/shared/live/live_compute.cpp
+++ b/prosper/frontends/shared/live/live_compute.cpp
@@ -8141,7 +8141,14 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                         // RGBA16F.  Re-quantize those numeric values into the exact sampled Vulkan
                         // layout.  Guest-backed surfaces take the lossless memcpy path below.
                         const size_t texels = static_cast<size_t>(volume_texels);
-                        for (size_t t = 0; t < texels; ++t) {
+                        // #3407: per-texel and independent -- every write is upload[t*n+c],
+                        // disjoint across t -- so this is the same shape as the arm below that
+                        // already uses the helper. Leaving it scalar is why a 4K conversion took
+                        // 25 ms on one core while 31 others idled. work_bytes is an upper bound
+                        // on bytes touched per texel; the helper uses it only to size the pool.
+                        parallel_compute_texels(
+                            texels, texels * 16u, [&](size_t begin, size_t end) {
+                        for (size_t t = begin; t < end; ++t) {
                             uint32_t channels[4]{};
                             for (uint32_t c = 0; c < 4; ++c) {
                                 float value = 0.0f;
@@ -8162,6 +8169,7 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                                 (channels[2] << 20) | (channels[3] << 30);
                             std::memcpy(upload + t * sizeof(packed), &packed, sizeof(packed));
                         }
+                            });
                     } else if (sampled_float16_native &&
                                ((source_f16x2 && sampled_components == 2) ||
                                 (source_f16x1 && sampled_components == 1))) {
@@ -8181,7 +8189,14 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                         }
                     } else if (sampled_uint8_native) {
                         const size_t texels = static_cast<size_t>(volume_texels);
-                        for (size_t t = 0; t < texels; ++t) {
+                        // #3407: per-texel and independent -- every write is upload[t*n+c],
+                        // disjoint across t -- so this is the same shape as the arm below that
+                        // already uses the helper. Leaving it scalar is why a 4K conversion took
+                        // 25 ms on one core while 31 others idled. work_bytes is an upper bound
+                        // on bytes touched per texel; the helper uses it only to size the pool.
+                        parallel_compute_texels(
+                            texels, texels * 16u, [&](size_t begin, size_t end) {
+                        for (size_t t = begin; t < end; ++t) {
                             for (uint32_t c = 0; c < sampled_components; ++c) {
                                 if (source_r8) {
                                     upload[t] = pixels[t];
@@ -8200,9 +8215,17 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                                 }
                             }
                         }
+                            });
                     } else if (sampled_unorm8x2) {
                         const size_t texels = static_cast<size_t>(volume_texels);
-                        for (size_t t = 0; t < texels; ++t) {
+                        // #3407: per-texel and independent -- every write is upload[t*n+c],
+                        // disjoint across t -- so this is the same shape as the arm below that
+                        // already uses the helper. Leaving it scalar is why a 4K conversion took
+                        // 25 ms on one core while 31 others idled. work_bytes is an upper bound
+                        // on bytes touched per texel; the helper uses it only to size the pool.
+                        parallel_compute_texels(
+                            texels, texels * 16u, [&](size_t begin, size_t end) {
+                        for (size_t t = begin; t < end; ++t) {
                             for (uint32_t c = 0; c < 2; ++c) {
                                 if (source_rg8) {
                                     upload[t * 2 + c] = pixels[t * 2 + c];
@@ -8219,9 +8242,17 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                                 }
                             }
                         }
+                            });
                     } else if (sampled_unorm16_native) {
                         const size_t texels = static_cast<size_t>(volume_texels);
-                        for (size_t t = 0; t < texels; ++t) {
+                        // #3407: per-texel and independent -- every write is upload[t*n+c],
+                        // disjoint across t -- so this is the same shape as the arm below that
+                        // already uses the helper. Leaving it scalar is why a 4K conversion took
+                        // 25 ms on one core while 31 others idled. work_bytes is an upper bound
+                        // on bytes touched per texel; the helper uses it only to size the pool.
+                        parallel_compute_texels(
+                            texels, texels * 16u, [&](size_t begin, size_t end) {
+                        for (size_t t = begin; t < end; ++t) {
                             for (uint32_t c = 0; c < sampled_components; ++c) {
                                 uint16_t value = 0;
                                 if (source_unorm8) {
@@ -8239,6 +8270,7 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                                             &value, sizeof(value));
                             }
                         }
+                            });
                     } else if (f32) {
                         const size_t texels = static_cast<size_t>(volume_texels);
                         const uint32_t output_components =
@@ -8267,13 +8299,21 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                             });
                     } else if (source_r8 && r8) {
                         const size_t texels = static_cast<size_t>(volume_texels);
-                        for (size_t t = 0; t < texels; ++t) {
+                        // #3407: per-texel and independent -- every write is upload[t*n+c],
+                        // disjoint across t -- so this is the same shape as the arm below that
+                        // already uses the helper. Leaving it scalar is why a 4K conversion took
+                        // 25 ms on one core while 31 others idled. work_bytes is an upper bound
+                        // on bytes touched per texel; the helper uses it only to size the pool.
+                        parallel_compute_texels(
+                            texels, texels * 16u, [&](size_t begin, size_t end) {
+                        for (size_t t = begin; t < end; ++t) {
                             const uint8_t value = pixels[t];
                             upload[t * 4 + 0] = value;
                             upload[t * 4 + 1] = value;
                             upload[t * 4 + 2] = value;
                             upload[t * 4 + 3] = value;
                         }
+                            });
                     } else if (source_unorm8) {
                         if (!copy_snapshot_exact()) {
                             skip_image(r, "renderer RTT RGBA8 copy byte count mismatch");
@@ -8281,7 +8321,14 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                         }
                     } else {
                         const size_t texels = static_cast<size_t>(volume_texels);
-                        for (size_t t = 0; t < texels; ++t) {
+                        // #3407: per-texel and independent -- every write is upload[t*n+c],
+                        // disjoint across t -- so this is the same shape as the arm below that
+                        // already uses the helper. Leaving it scalar is why a 4K conversion took
+                        // 25 ms on one core while 31 others idled. work_bytes is an upper bound
+                        // on bytes touched per texel; the helper uses it only to size the pool.
+                        parallel_compute_texels(
+                            texels, texels * 16u, [&](size_t begin, size_t end) {
+                        for (size_t t = begin; t < end; ++t) {
                             for (uint32_t c = 0; c < 4; ++c) {
                                 uint16_t half = 0;
                                 std::memcpy(&half, pixels.data() + t * 8 + c * 2, sizeof(half));
@@ -8292,6 +8339,7 @@ bool execute_item(VulkanComputeContext& ctx, const prosper::gpu::ComputeItem& it
                                     std::lround(value * 255.0f));
                             }
                         }
+                            });
                     }
                     if (trace) {
                         const uint64_t snapshot_hash = fnv1a(live_target.pixels->data(),
--- a/prosper/tests/diagnostics/test_env_numeric_sites.cpp
+++ b/prosper/tests/diagnostics/test_env_numeric_sites.cpp
@@ -100,6 +100,19 @@ static uint64_t mib_cap_old(const char* t) {
     return mib * 1024ull * 1024ull;
 }
 
+// --- tests/fixtures/render_runner.h : PROSPER_RENDER_COPY_THREADS -----------------------------
+// #3407: worker count for the parallel texture-staging copy. 0 means "derive from the core count",
+// so a malformed value must land on 0 rather than a wrapped 32 -- which is what a strtoul parse of
+// "-1" produced before this site used the helper.
+static uint64_t copy_threads_new(const char* n, const char* t) {
+    return env_u64_or_default_capped(n, t, 0ull, 32ull, "threads");
+}
+static uint64_t copy_threads_old(const char* t) {
+    if (!t || !*t) return 0ull;
+    const unsigned long parsed = std::strtoul(t, nullptr, 10);
+    return parsed > 32ul ? 32ull : static_cast<uint64_t>(parsed);
+}
+
 // --- tests/fixtures/render_runner.h : PROSPER_BACKEND_BUFFER_ARENA_KB --------------------------
 static uint64_t arena_new(const char* n, const char* t) {
     const uint64_t kib = env_u64_or_default_capped(n, t, 1024ull, UINT64_MAX / 1024ull, "KiB");
@@ -291,6 +304,8 @@ static const Site kSites[] = {
     // for both. A malformed value must keep the default rather than disable the bound.
     {"render_runner.h PROSPER_MAPPED_STAGING_MB", "PROSPER_MAPPED_STAGING_MB",
      mib_cap_new<256>, mib_cap_old<256>, "-1", 256ull * kMiB, "512", 512ull * kMiB},
+    {"render_runner.h PROSPER_RENDER_COPY_THREADS", "PROSPER_RENDER_COPY_THREADS",
+     copy_threads_new, copy_threads_old, "-1", 0ull, "4", 4ull},
 
     {"render_runner.h PROSPER_BACKEND_BUFFER_ARENA_KB", "PROSPER_BACKEND_BUFFER_ARENA_KB",
      arena_new, arena_old, "4mb", 1024ull * 1024ull, "2048", 2048ull * 1024ull},
--- a/prosper/tests/fixtures/render_runner.h
+++ b/prosper/tests/fixtures/render_runner.h
@@ -3,7 +3,8 @@
 // pixels. Used to verify recompiled shaders end-to-end (render -> readback -> pixel asserts). The
 // including test links Vulkan::Vulkan.
 #pragma once
-#include <thread>   // #3407: parallel_render_memcpy
+#include <system_error>   // #3407: the catch in parallel_render_memcpy
+#include <thread>         // #3407: parallel_render_memcpy
 #include "host/memory/guest_write_watch.hpp"   // VA->phys for the #2932 target census
 #include "shared/rtt/mrt_extent.hpp"
 #include <vulkan/vulkan.h>
--- a/prosper/tests/fixtures/render_runner.h
+++ b/prosper/tests/fixtures/render_runner.h
@@ -2422,11 +2422,14 @@ inline RenderMemoryPool& render_memory_pool() {
 // returns, so the copy is complete before the caller unmaps or submits. Small copies stay on the
 // calling thread, because thread creation would cost more than the copy.
 inline void parallel_render_memcpy(void* dst, const void* src, size_t bytes) {
-    static const unsigned configured = [] {
+    static const unsigned configured = []() -> unsigned {
+        // env_numeric, not strtoul: a malformed or negative value must keep the default rather
+        // than wrap. "-1" through strtoul is 32 workers, which is the opposite of what a reader
+        // setting it to -1 to "turn it off" would expect.
         const char* value = getenv("PROSPER_RENDER_COPY_THREADS");
-        if (!value || !*value) return 0u;
-        const unsigned long parsed = std::strtoul(value, nullptr, 10);
-        return static_cast<unsigned>(std::min(parsed, 32ul));
+        const uint64_t parsed = prosper::diag::env_u64_or_default_capped(
+            "PROSPER_RENDER_COPY_THREADS", value, 0ull, 32ull, "threads");
+        return static_cast<unsigned>(parsed);
     }();
     // Below this the copy is not worth splitting; above it the win is real and bandwidth-bound,
     // which is why the worker count is capped rather than scaled to the core count.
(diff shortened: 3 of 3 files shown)


mattias800:
## Second commit added: the renderer's texture staging copy (81bbd8244)

Same defect, second location. After the compute-side change landed in this PR I re-profiled rather than assuming the old profile held, and `__memmove` had become **36%** of the profile — with an inverted call graph putting **16.96% directly under `render_draws_rgba`**. That is the 4K texture upload: one ~33 MB `memcpy` into mapped staging, on a single core.

`render_runner.h` had **no threading of any kind**, so unlike the compute side there was no existing helper to reach for. `parallel_render_memcpy` splits the copy across disjoint destination ranges and joins every worker before returning, so the bytes are in place before the caller unmaps or submits. Copies under 2 MiB stay on the calling thread. `PROSPER_RENDER_COPY_THREADS` is the seam and the escape hatch.

| arm (same binary) | guest flips/s | mean | renderer-resource |
| --- | --- | --- | --- |
| single-threaded copy | 4.58, 4.38, 4.59 | 4.52 | 1429 / 1469 / 1401 ms |
| parallel copy | 4.78, 4.83, 4.58 | **4.73** | **1073 / 1165 / 1166 ms** |

**+4.7% on the mean, renderer-resource down 20.8%.**

**Read the resource column, not the flips column.** The flips distributions overlap by one run (A max 4.59, B min 4.58); `renderer-resource` is cleanly separated with no overlap and is the measurement that matches the mechanism. I would not claim the flips delta on its own from three runs.

Both arms already contain the compute-side parallelisation, so this is the marginal effect of the renderer copy alone.

## Cumulative, for context

| stage | fps |
| --- | --- |
| session start | 3.38 |
| + staging mapping (#3406, merged) | 3.59 |
| + compute conversion parallelised | 4.18 |
| + render copy parallelised | **4.73** |

`process CPU` has gone 1.01 → 1.68 cores on a 32-core machine. Both changes are scheduling fixes, not algorithmic ones — the same work, spread over cores that were already idle.

Full suite **369/369** after each commit.

mattias800 (commented):

Reviewed at head `81bbd8244`. Two blocking findings, both cheap to discharge; everything else below is a
note. The core of the change is sound and I verified the safety argument independently rather than
taking it on the author's word — the results of that verification are in §A and §B, because a "yes, it's
fine" is only useful if it says what was checked.

---

### A. Commit 1 — parallel-safety of the six wrapped bodies: **verified clean**

I re-derived this mechanically (balanced-brace extraction of each lambda body, then a scan for control
flow, indexed writes, memcpy destinations and locally-declared names) rather than by eye, and then read
all six by hand. Bodies, by line in the head revision of `prosper/frontends/shared/live/live_compute.cpp`:

| arm | lambda body | write target | control flow |
| --- | --- | --- | --- |
| `unorm2_10_10_10` | 8150-8172 | `upload + t*4` (memcpy) | clean |
| `sampled_uint8_native` | 8198-8218 | `upload[t*nc + c]`, `upload[t]` | clean |
| `sampled_unorm8x2` | 8227-8245 | `upload[t*2 + c]`, `c<2` | clean |
| `sampled_unorm16_native` | 8254-8273 | `upload + (t*nc + c)*2` (memcpy) | clean |
| `source_r8 && r8` | 8308-8316 | `upload[t*4 + 0..3]` | clean |
| else (FP16 -> UNORM8) | 8330-8342 | `upload[t*4 + c]`, `c<4` | clean |

- **No `break`, `continue`, `return` or `goto`** in any of the six. The `break`s in this region all
  belong to the enclosing `else if` chain (the `copy_snapshot_exact()` failure arms and the
  `skip_image` guard at 8117-8120), outside every wrapped body. The author's claim holds.
- **Every write is indexed by `t`** and the per-`t` footprints are disjoint: `c` is always bounded by
  the same stride the index multiplies by, so iteration `t` owns exactly `[t*n, t*n+n)`.
- **No cross-iteration state.** Every name assigned inside a body (`channels`, `half`, `value`, `f`,
  `packed`) is declared inside that body. Nothing declared outside is mutated: `pixels` is
  `const std::vector<uint8_t>&` (8079), `upload` is a `uint8_t*` that is only written through, and the
  `source_*` / `sampled_*` predicates are `const` and read-only.
- **Callees are pure.** The only non-trivial one is `prosper::gpu::half_to_float`
  (`prosper/src/gpu/resources/shader_resources.cpp:32`) — pure bit arithmetic, no statics, no table.
  That was the one way this could have gone wrong invisibly, so it is worth stating that it was checked.
- **Destination is safe to write from many threads.** `upload` is the `vkMapMemory` mapping at
  `live_compute.cpp:7727`, and `host_memory_type` (`:3233`) always requires `HOST_COHERENT`, so no
  `vkFlushMappedMemoryRanges` is owed. Writing disjoint byte ranges through one mapping is not a Vulkan
  operation — Vulkan's external-synchronisation requirement is on the `VkDeviceMemory` *object* commands
  (`vkMapMemory`/`vkUnmapMemory`/flush/invalidate), none of which run concurrently here. The helper joins
  before returning, so all writes happen-before the subsequent record and submit.

One thing worth recording since it looks alarming and is not: in the `sampled_uint8_native` arm the
`source_r8` branch writes `upload[t]` inside a `c < sampled_components` loop, so with `nc > 1` it would
write the same byte `nc` times and leave `upload[t*nc + 1..]` untouched. That is still disjoint across
`t` (so parallelisation is unaffected), it predates this PR, and it is unreachable: `exact_narrow_layout`
(8102-8110) only admits `source_r8` when `r8 || (sampled_uint8_native && sampled_components == 1)`, and the
guard at 8117-8120 rejects everything else. No action needed; noting it so the next reader does not re-derive it.

### B. Commit 1 — `work_bytes = texels * 16u`: **correct, and it cannot break anything**

`parallel_compute_texels` (`live_compute.cpp:548`) uses `work_bytes` only at `:560`:
`by_work = min(count, max(1, work_bytes / 512KiB))`, then `threads = max(1, min(wanted, by_work))`.
The value therefore feeds nothing but the worker count, which is independently clamped to `count` and to
`max_threads`. **A wrong `work_bytes` can only produce a suboptimal thread count — never a wrong range,
a gap, or an overlap.** Answering the question directly: no, nothing worse is reachable.

The `16` is also a genuine upper bound for all six arms, which is what the added comment claims. Worst
case is the `sampled_unorm16_native` arm at exactly 16 (8 B read + 4x2 B written); the others are 5-12.
The three narrow arms are over-estimated by up to ~3x, so a surface in the 32K-200K texel band may take
2-3 workers where an accurate figure would take 1. That is a wash against thread-start cost, not a
defect. If you want it tighter later, the `f32` arm at 8278-8280 already shows the per-arm form.

### C. Commit 2 — `parallel_render_memcpy` chunk arithmetic: **verified correct**

`prosper/tests/fixtures/render_runner.h:2424-2458`. Ranges are `[t*chunk, min(bytes, (t+1)*chunk))` for
`t = 0..threads-1` with `chunk = ceil(bytes/threads)`, so `threads*chunk >= bytes` and coverage is exact
and non-overlapping. The calling thread's `min(bytes, chunk)` at `:2456` is precisely the `t = 0` range
the worker loop skips by starting at `t = 1`. Worked the boundaries by hand: `bytes` = 2 MiB exactly,
2 MiB + 1, 3 MiB, and 33,177,600 (4K RGBA, divides evenly) — all exact, no gap, no overlap, no empty
worker. Zero/negative ranges are unreachable twice over: `bytes >= 2 MiB` and `threads <= bytes/1 MiB`
force `chunk >= 1 MiB`, and `if (begin >= end) break` at `:2450` catches it anyway. `t*chunk` cannot
overflow because `threads*chunk < bytes + threads`.

Destination safety is the same argument as §A: staging is requested `HOST_VISIBLE|HOST_COHERENT`
at `:7956-7961` and `pick` (`:5268`) requires an exact property match or returns `UINT32_MAX`, so the
mapping is always coherent — no flush is owed, and the joins at `:2457` order every worker's write before
the caller records the copy. `render_draw_pass_rgba` is mutually exclusive with itself (the guard
documented at `:745`), so there is no nesting or re-entrancy hazard either.

---

## Blocking finding 1 — a failed `pthread_create` takes the whole emulator down (`render_runner.h:2445-2457`)

`workers` is a `std::vector<std::thread>`. If `emplace_back` at `:2451` throws `std::system_error`
(`pthread_create` returning `EAGAIN`) after at least one worker has started, the exception propagates,
stack unwinding destroys the vector, and `~std::thread()` on a joinable thread calls `std::terminate()`.

**Inputs/state that produce it:** any `bytes >= 2 MiB` call made while the process is at its thread
ceiling — `RLIMIT_NPROC`, a cgroup `pids.max`, or `kernel/threads-max`. Concretely: `threads == 8`,
workers 1-3 start, worker 4's `pthread_create` returns `EAGAIN` -> process dies with no diagnostic.

Three things make this worth blocking on rather than filing:

1. **The path this replaced could not fail.** A plain `std::memcpy` has no failure mode; the new code
   introduces one, into the live renderer. `render_runner.h` is not test-only — `live_renderer.cpp`,
   `gpu_executor.cpp`, `present_blit.cpp`, `live_compute.cpp`, `boot_trace` and `gpu_replay` all include
   it, and CLAUDE.md itself cites `render_runner.h:5199` as *the* backend memcpy in the shipped profile.
2. **The exposure is not incidental.** Your own figure is 16.96% of wall clock in this memcpy, i.e. it
   runs many times per frame, and each call creates and joins up to 7 OS threads. That is sustained
   thread churn in the hundreds per second, so `EAGAIN` is a live risk over a long run rather than a
   theoretical one.
3. **This project already decided this case, and wrote down why.** `parallel_compute_texels`
   (`live_compute.cpp:568-590`) uses `std::jthread` inside a `try`/`catch (const std::system_error&)`
   and finishes the unspawned ranges synchronously, under a comment that says exactly this: *"if the
   next OS thread cannot be created, already-started workers are still joined. Finish the unspawned
   ranges on this thread so transient resource pressure degrades to less parallelism instead of
   aborting."* Commit 1 leans on that helper; commit 2 reimplements it 40 lines of prose away from that
   comment and drops the one property the comment exists to justify.

Note that merely catching is **not** sufficient and would be worse: the unspawned ranges
`[next*chunk, bytes)` would go uncopied and the staging buffer would carry stale bytes into a texture —
silent corruption instead of a loud crash. Port the whole shape from `parallel_compute_texels`:
`std::jthread`, `try`/`catch`, and a synchronous tail loop over the ranges that did not get a worker.

While you are in there, consider whether a persistent pool is the better answer — `DrawRealizationPool`
(`prosper/src/gpu/execute/gpu_executor.cpp:9600-9670`) is the in-repo template, and it removes both the
`EAGAIN` surface and the per-copy create/join cost. Not required for this PR; the measured win stands
either way.

## Blocking finding 2 — the `+18.7%` in the PR title is an upper bound, not this commit's measured effect

`PROSPER_COMPUTE_CONVERSION_THREADS=1` sets `configured = 1`, so `wanted = 1`, so `threads = 1`
(`live_compute.cpp:552-563`). That serialises **every** `parallel_compute_texels` call site, not the six
this commit adds. The base commit (`68157c402`) already has **14** other call sites, several of them on this title's hot
path (line numbers below are as of head `81bbd8244`):

- `:8014` — the exact-storage-bytes parallel memcpy of every storage-image seed
- `:3553`, `:3620`, `:3663` — the detiler
- `:603`, `:619` — `compute_buffers_equal` / `copy_compute_buffer`, whose own comment says a 32 MiB
  persistent SSBO "still costs several milliseconds on one core"
- `:8278` (the `f32` sampled arm), `:8556`, `:8573`, `:10415`, `:11444`, `:11466`, `:11491`, `:11558`

So the `scalar` row is not master — it is strictly *more* serial than master, which makes 3.52 flips/s a
**lower** bound on master's rate and `+18.7%` an **upper** bound on this commit's contribution. The error
is one-directional and in the flattering direction, which is exactly the case CLAUDE.md says gets checked
least. It is also inconsistent within the PR: commit 2's message is careful to say *"Both arms already
contain the compute-side parallelisation, so this is the marginal effect of the renderer copy alone"* —
commit 1 needed the same sentence and does not have it.

**The code is fine and I am not asking for a re-run.** Your own `1.01 cores` observation argues the
pre-existing sites contribute little on this title, and the compute-time column (1792.7/1800.6/1803.9 ->
907.6/890.3/869.7, a 0.6% within-arm spread against a 50% drop) is overwhelming regardless of
attribution. One sentence discharges this: say that the `THREADS=1` arm also disables the 14 pre-existing
sites, so the figure bounds this commit's effect from above. What must not happen is `+18.7%` travelling
into #3407 or the charter as a measured marginal effect, because nobody downstream re-derives a number
that arrived with a table attached.

---

## Non-blocking notes

1. **Indentation.** The six wrapped `for` bodies keep their original indentation inside the new lambda
   (e.g. 8149-8172), so the `for` now sits at the same column as the `parallel_compute_texels` call that
   encloses it. It made the diff readable, which was the right call for review, but the cost is
   permanent and lands on every future reader of this 11k-line file. The already-wrapped `f32` arm at
   8278-8299 shows the re-indented form. Worth a follow-up formatting commit at minimum.
2. **`PROSPER_RENDER_COPY_THREADS` parsing (`render_runner.h:2426-2429`).** Answering the question
   directly: **no arm is required in `tests/diagnostics/test_env_numeric_sites.cpp`, and nothing is
   missing.** `tools/env/check_env_numeric_arms.py` matches only the first argument of
   `env_u64_or_default` / `_auto` / `_capped` / `env_u64_or_report`; a raw `getenv` + `strtoul` site is
   invisible to it by construction. I ran all three text gates against this head: `env_numeric_arms`
   green (32/32), `cached_env_arming_logic` green, `ascii_output_literals` green.
   That said — this header already includes `diagnostics/env_numeric.hpp` at line 15, under the comment
   *"a typo must not silently re-size a cache"*, and that header's own preamble is a long argument
   against exactly the spelling used here. With raw `strtoul`, `=-1` silently becomes 32 threads,
   `=0x8` and `=abc` silently become "default", and nothing says so — on the variable that *is* the A/B
   seam for the measurement in this PR. `PROSPER_COMPUTE_CONVERSION_THREADS` has the same shape so
   there is local precedent, but `configured_draw_realization_threads`
   (`gpu_executor.cpp:9618`) validates with the `end` pointer and is the better neighbour to copy.
   Switching to `env_u64_or_default_capped` would also earn the knob a gate arm for free.
3. **The PR body describes commit 1 only.** Commit 2 adds the first threading in `render_runner.h`, a new
   environment variable, and a second measurement table — none of which appear in the description.
   CLAUDE.md asks for a self-contained PR description; as it stands a reader of the PR page would not
   know the renderer changed.
4. **`-22%` on renderer-resource** (commit 2 message) is `-20.8%` on the means
   (1433.0 -> 1134.7). Small, and in the flattering direction. `+4.6%` on flips is `+4.7%`. Both are
   rounding, but this PR's other numbers are exact.
5. **"used in eight other places in this same file"** — it is 14 on the base commit. Understated, so it
   only weakens an argument that is stronger than claimed.
6. **"every parallel run is faster than every scalar run except one (3.79 vs 3.59)"** is self-refuting as
   written: 3.79 > 3.59, so all three parallel runs beat all three scalar runs and the distributions are
   cleanly separated. You understated your own result.
7. **Measurement honesty, overall — fairly stated, and commit 2 is a model of it.** Commit 1's evidence
   is strong on its own terms: both columns cleanly separated, and a 2x compute-time effect against a
   0.6% within-arm spread is far outside noise. Commit 2 is weaker and says so unprompted — the flips
   distributions do overlap by one run (4.58 vs 4.59), and pointing at `renderer-resource`
   (1401-1469 vs 1073-1166, cleanly separated) as the stronger column is the right call, because it is
   also the column the mechanism predicts. Worth knowing for future 3v3 designs: complete separation at
   n=3 is the strongest signal that design can produce and still only reaches p = 0.1 two-sided, so
   "cleanly separated at n=3" is honest evidence but never significance. The claims here are within
   what the data supports; finding 2 is about attribution, not about the numbers themselves.
8. **`BLOG.md`.** No screenshot, so no entry is required. But an 18.7% frame-rate win on a tracked title
   from a scheduling fix is the kind of finding the charter explicitly says is worth a short entry even
   with no picture attached. Your call.

---

Fix finding 1, add the sentence for finding 2, and I will re-review — both are small enough that the
re-read is a few minutes. Nothing else here needs to change before merge.

**REJECTED**

mattias800 (commented):
## Independent re-review — APPROVED

Head `f5bf02737`. Both blocking findings are fixed, and fixed properly rather than minimally. I did
not take the control flow on inspection this time — the questions you asked are the kind that are
easy to get right by eye and wrong in one corner, so I extracted the real code and ran it. Details
below, including one case my own sweep provably could **not** reach, and a correction to something I
got wrong in the first pass.

---

## Finding 1 — exception safety: verified, including the case you flagged

### What I ran

**(a) The real function, compiled standalone and executed.** I sed'd `render_runner.h:2424-2481`
verbatim into a translation unit with the real `diagnostics/env_numeric.hpp` — nothing retyped — and
copied random buffers at 12 sizes chosen for the boundaries: 1, 1024, 2 MiB−1, 2 MiB, 2 MiB+1, 3 MiB,
3 MiB+7, 8 MiB+12345, 33,177,600 (4K RGBA, divides evenly by 8), 33,177,601, 16 MiB+1, 33 MiB.
**Byte-exact on all 12.** This box reports 32 cores, so `wanted = 8` and every case ≥ 8 MiB really did
run 7 `jthread`s plus the caller, not a degenerate serial path.

**(b) An exhaustive coverage model with the arithmetic kept verbatim.** I took lines 2444-2480 and
substituted *only* the container type and the three copy calls — every index, clamp, `break` and
`next_worker` expression is the file's own text — then swept `bytes` × `wanted` ∈ [2,32] × `fail_at`
∈ [0, threads], where `fail_at` makes the simulated `pthread_create` throw `std::system_error` at
that worker index. **2,964 cases, 2,344 of which actually threw, and in every one every byte was
written exactly once** — no gap, no double-copy, no range both spawned and re-copied by the tail.

**(c) A mutation arm, because (b) is worthless if it cannot fail.** Changing the tail loop's init from
`for (; next_worker < ...)` to `for (++next_worker; ...)` — i.e. skipping exactly one unspawned range —
turns all 2,344 throwing cases red with `uncopied=1048576`. The harness detects the defect it asserts
absent. (It also caught me first: my initial mutation run reported green because `#include "body.inc"`
resolves against the *including file's* directory and ignored my `-I`, so the mutant was never
compiled. The lever has to be shown to move.)

### Your specific question — the `break`/`next_worker` interaction

**Your reasoning is correct, and I can now say so from a run rather than an argument.** When the try
loop's `if (begin >= end) break` fires, the `for` increment does not execute, so `next_worker` still
names that empty range; the tail loop recomputes the identical `begin`/`end` and breaks immediately
having copied nothing. Coverage is unaffected.

But there is a wrinkle worth recording, and it is why I did not just tell you "verified" off sweep (b):
**that branch is unreachable from the real entry point, so sweep (b) never executed it.** I
instrumented both `break`s and counted: `try-loop empty-range breaks = 0, tail empty-range breaks = 0`
across all 2,964 cases. It cannot fire, because `bytes >= 2 MiB` and `threads <= bytes/1 MiB` force
`chunk >= 1 MiB` and `bytes > threads²` for every `threads <= 32`, hence `(threads-1)*chunk < bytes`
always. A clean zero from an instrument that could not have reported anything else is not evidence.

So I built the positive instances **by hand, outside the generator**: the same verbatim try/tail loops
driven directly with `(bytes, threads)` pairs where `(threads-1)*chunk >= bytes` — (9,4), (10,6),
(13,8) — across every `fail_at`. 21 cases, both break counters non-zero, **coverage exact in all of
them**, including the interesting mixed case `bytes=9 threads=4 fail_at=3` where a spawn failure and
an empty range occur in the same call (`try-break=1, tail-break=1`, nothing uncopied). So the branch
is dead code today *and* correct if the constants ever change. That is the right state for it.

### `jthread` auto-join and the catch fall-through — safe

Four properties, all checked:

1. **Workers still running when the exception is thrown are joined, not abandoned.** `~jthread` calls
   `request_stop()` then `join()`; the vector is destroyed at function scope exit, after the tail. The
   `request_stop` is a no-op here since the lambda takes no `stop_token` — the join is what matters.
2. **The tail runs concurrently with those workers and does not race them.** Tail ranges are
   `[next_worker*chunk, bytes)`, worker ranges are `[1*chunk, next_worker*chunk)`; disjoint by
   construction, which is what (b) and the hand-built cases confirm empirically.
3. **No dangling capture.** `[dst, src, begin, end]` is by value — only pointers and `size_t`s — so a
   worker holds nothing pointing into this frame. (`parallel_compute_texels` captures `&body`, which is
   also fine there for the same join-before-return reason; by value is the stronger form and you kept it.)
4. **The `catch` is exhaustive for this call, so nothing can escape and silently skip ranges.**
   `reserve(threads-1)` precedes the loop and the loop adds at most that many, so no reallocation and
   no `bad_alloc` from the vector; the lambda captures only trivially-copyable scalars, so its
   copy-construction is `noexcept`; that leaves thread creation, which throws `system_error`. The one
   residual is a `bad_alloc` from libstdc++'s internal thread-state allocation — it would propagate,
   but with `jthread` the vector still joins on the way out, so it is loud rather than corrupting, and
   it is identical to what `parallel_compute_texels` already accepts. Nothing to do.

The comment at 2449-2456 explaining *why* catching alone is insufficient is the right call — that is
precisely the edit a later reader would otherwise "simplify" away.

## Finding 2 — the env knob: correct, and the arm is real

- **`_capped` clamps, it does not reject over-cap**, so `"40"` still selects 32 exactly as
  `std::min(parsed, 32ul)` did. No behaviour change for well-formed input.
- **Unset and empty stay silent** — `env_u64_or_default` returns the fallback without a message for
  those, so the common case gains no stderr noise.
- **End-to-end through the real function**, running the standalone build with the variable set:
  `-1`, `0x8` and `abc` each print
  `[env] PROSPER_RENDER_COPY_THREADS='…' is not a plain non-negative count of threads -- keeping the
  default (0) and changing NOTHING` and fall back to the core-count default; `1`, `4`, `40` are
  accepted silently; all six remain byte-exact. `=1` still works as the A/B seam.
- **`copy_threads_old` is a faithful model of the pre-change site.** I did not eyeball this — I
  extracted the site verbatim from `81bbd8244` and differentially compared it against the model over
  24 inputs (null, empty, 0, 1, 4, 31, 32, 33, 40, `-1`, `-0`, `abc`, `0x8`, `" 4"`, `"4x"`, `"+4"`,
  `"  12  "`, `"007"`, `"3.5"`, `"\t2"`, 2^32, 2^64−1, a 20-digit overflow). **Zero divergences.**
  `parsed > 32 ? 32 : parsed` is exactly `std::min(parsed, 32ul)`.
- **The in-test discriminator fires**: `legacy("-1")` = 32 ≠ 0, so the arm is not vacuously green.
  Built and ran `test_env_numeric_sites.cpp` standalone — all three of its assertions for this knob
  pass, and the whole file is `== all passed (0 failures) ==`.
- **"Would it fail if the site reverted to strtoul?"** Not via the arm — it mirrors rather than calls,
  by the file's own design. Via the gate, yes, and I proved it rather than asserting it: I reverted the
  site to the old `strtoul` spelling in a scratch copy of the tree and ran
  `check_env_numeric_arms.py`, which goes red with
  `arms PROSPER_RENDER_COPY_THREADS, but no call site parses it with env_numeric any more -- the arm
  asserts about code that is gone`. On the real head it is green at 33/33.

Your read of why this happened is right and worth keeping in the commit message: converting the site is
what made it *visible* to the gate, and the gate then failed within an hour of the knob existing. That
is the anti-drift mechanism working on its first opportunity.

**Gates at this head** (text scans, nothing built): `env_numeric_arms` 33/33 green,
`cached_env_arming_logic` green, `ascii_output_literals` green, `check_diag_gates.py` repo scan green
(108 findings, 108 baselined). Branch touches three files, no `.md`.

## A stated invariant this could have tripped and does not

`prosper/tests/fixtures/AGENTS.md` warns that this header cannot lean on the HLE's locks (#2953/#3240),
and `render_runner.h:1513` records that "graphics and compute run strictly sequentially on one thread,
so sharing the queue needs no extra synchronization". Neither is violated: the workers touch no Vulkan
handle, no `static`, and no cache entry point — they `memcpy` into a pointer that was already mapped by
the owning thread, inside the `BackendPersistentResourceGuard` critical section that
`render_draw_pass_rgba` holds for its whole body. `backend_persistent_resource_peak_in_flight()` is
unaffected because no worker enters that domain.

---

## Your question: is the `+18.7%` attribution blocking for merge?

**No. Not blocking.** It is a claim-accuracy issue rather than a defect, the direction of the error is
known and one-directional (the figure is an upper bound), and your own `1.01 cores` observation is
decent evidence that the marginal effect is close to it. Holding a correct change over a sentence
would be disproportionate.

**But "PR body plus a note on #3407" does not do what you think, and I checked before answering.** This
repository's `squash_merge_commit_message` is **`COMMIT_MESSAGES`**, not `PR_BODY` — the squash commit's
body is composed by concatenating the commit messages, so editing the PR body leaves the uncorrected
sentence in permanent history and the PR body is the half nobody greps later. Three ways out, in
increasing cost:

1. `gh pr merge 3408 --squash --body-file <corrected>` — overrides the composed message at merge time.
   Cheapest, no force-push, and the permanent record is right.
2. A third commit whose message carries the correction; it gets concatenated in too, and it is visible
   on the branch before merge.
3. Amend and force-push. Unnecessary, and it detaches this review from head.

Do (1) or (2), **and** the #3407 note, since that issue is where the figure will actually be quoted
from. One further condition: if this number ever moves into `COMPATIBILITY.md`, a status doc, or
`CLAUDE.md`, the qualifier travels with it — that is the propagation path the charter is actually
about, and it is downstream of all three options above.

Suggested wording, since it is one sentence: *"The `THREADS=1` arm also serialises the 14 pre-existing
`parallel_compute_texels` sites, so `+18.7%` bounds this change's contribution from above rather than
measuring it; the `1.01 cores` figure on master argues the gap is small."*

---

## Correction to my own first pass

I wrote that `live_renderer.cpp`, `gpu_executor.cpp`, `present_blit.cpp`, `live_compute.cpp`,
`boot_trace` and `gpu_replay` "all include" `render_runner.h`. **That list came from `grep -l` and is
wrong** — it counted comment mentions as includes. Checking the actual `#include` lines, the real
includers are four: `live_renderer.cpp:44`, `present_blit.cpp:4`, `boot_trace.cpp:45`,
`gpu_replay.cpp:15`. `live_compute.cpp`, `gpu_executor.cpp` and `host_read_barrier.hpp` only mention it
in prose — and `tests/fixtures/AGENTS.md` had already recorded exactly this correction for
`live_compute.cpp`, having checked it with `-M` rather than grep, which I would have seen had I opened
the folder's map first.

The conclusion is unchanged and if anything sharper: `live_renderer.cpp` and `present_blit.cpp` **are**
the shipping frontend, which is the whole basis of finding 1. But I spent a section of the last review
on how a citation makes a claim read as verified, so my own miscounted list needed correcting in the
same place it was published.

## Remaining nits — none blocking, none new to this commit

1. **`render_runner.h` names `std::system_error` in a `catch` without including `<system_error>`.** It
   builds because libstdc++'s `<thread>` pulls it in transitively (my standalone extraction compiled
   without it too), but a catch-clause type must be complete and this header is included by four TUs
   and built on MinGW. `live_compute.cpp:56` includes it explicitly for the same `catch`. One line.
2. **`#include <thread>` sits at line 6**, ahead of the project headers, while the file's other standard
   headers are grouped at 25-40. Cosmetic.
3. **Still outstanding from the first pass, all cosmetic and all in commit 1's pushed text**: the six
   wrapped bodies were never re-indented (the `f32` arm at 8278-8299 shows the intended form); "eight
   other places" is 14; `-22%` on renderer-resource is `-20.8%` on the means; and "every parallel run is
   faster than every scalar run except one (3.79 vs 3.59)" is self-refuting, since 3.79 > 3.59 means all
   three are. The indentation is worth a follow-up commit; the three prose items share the disposition
   of the measurement point above — fix them in whichever artifact you use for it.

## On the stale-binary disclosure

Flagging that yourself is the right instinct, and the counting-error-lines check is the right fix. For
the record on this pass: what I executed myself is the standalone extraction of `parallel_render_memcpy`
(12 sizes), the coverage model (2,964 + 21 cases) with its mutation arm, the differential test of
`copy_threads_old`, `test_env_numeric_sites` standalone, the gate revert control, and the four text
gates. I did **not** re-run the Vulkan suite, so the 369/369 is yours, on your clean-rebuild evidence.

---

Nothing here needs to change before merge. Clean.

**APPROVED**

PR #3409, opened 2026-09-06
docs: record the ten falsified Sonic Frontiers performance hypotheses

mattias800:
## Why

The charter rule: *"A PR that falsifies a hypothesis records it in the relevant `## Ruled out` section before merging... A falsification that lives only in a PR body or an issue comment is one the next agent will re-derive at full cost."*

The Sonic Frontiers performance investigation (#3405 / #3407, merged in #3406 and #3408) produced ten results of that kind and recorded **none** of them in `SONIC_FRONTIERS_STATUS.md`. They were in issue comment threads only.

## What lands

**`SONIC_FRONTIERS_STATUS.md` § Ruled out** — the table goes from **29 to 38 data rows**, plus one entry in a new subsection. Six cost a full implement-measure-revert cycle rather than a cheap A/B, which is exactly what this section exists to preserve:

| Hypothesis | Verdict |
| --- | --- |
| The intro is GPU-bound | **No** — `gpu-device` 5.4% of a 5020 ms window; 1.01 of 32 cores; corroborated by a second, independent 4760 ms window at 6.0% |
| `seed_reprove` re-proving costs the frame | **No** — 3.39 flips/s in both arms, and the lever is verified live |
| The per-dispatch `linear_seed` allocation is the cost | **No** — implemented, measured, reverted; scoped so it does not over-generalise |
| The compute image cache is thrashing | **No** — and the structural reason is stronger than the sweep |
| The compute memory pool is too small | **No** — setup 1130 ms against a 1147 ms baseline |
| Staging is write-combined | **No** — but flagged as a code-reading result, not a measurement |
| `upload_mapping.unmap()` costs the frame | **No** — with its missing denominator stated |
| `not_native_exact` starves the FMV planes' cache | **No** — every refusal is 2x2 / 1x1; none are the 4K planes |
| An FMV always presents new content, so caching cannot help | **No** — 441 conversions → 148 distinct buffers, a 66% collapse needing no content assumption |

**New: `### Void, not falsified — do not cite these as settled`**, modelled on `GTA5_STATUS.md`'s. It holds one entry — pooling the readback destination buffer. That was *implemented, measured and reverted*, and my first draft recorded it as falsified. It is not: labelling the A/B arms shows the flip median favours one arm while the `readback` component timer favours the other, at n=3 with no reuse counter. The entry states the verdict is **labelling-independent** (read either way round, only the signs swap), so nobody can "correct" the arm order and believe they have overturned it.

Two rows carry a positive finding worth more than the falsification: the same published buffer is re-converted about three times on average, and the planes are uncached because they are **renderer-owned** (`live_compute.cpp:7293`), not because anything refused them.

**`DEBUGGING_WORKFLOWS.md` § Slow gameplay** — two cross-title techniques:

- **Divide by the core count before profiling symbols.** ~1 core-second per wall-second establishes *serialisation* — not that serial **host** work is the bottleneck, since a GPU- or I/O-blocked process reads the same for the opposite reason. The paragraph names all three legs that licensed the inference here.
- **A widely used parallel helper is not evidence the chain you are profiling uses it.** `parallel_compute_texels` was live at *fourteen call sites* in that file, which is what made it invisible that one of the sampled chain's seven arms was wrapped while six 4K scalar loops sat among wrapped neighbours. Count call sites, not grep hits.

## Review

Two rounds, both rejected before this one, and the review changed the content rather than polishing it:

- **Round 1 (5 blocking).** "Fifteen sites" was fourteen — `grep -c` counts the template definition, and this PR was the only artifact in the project saying fifteen. A cause was miscategorised. Row 1 spliced two captures and failed its own arithmetic. Row 5 quoted a treatment number whose only nearby baseline made a null read as an 18% win. **And row 10 was not a falsification at all** — that finding is why the Void subsection exists.
- **Round 2 (1 blocking).** A `file:line` I introduced *while fixing citations*: `:8557` was correct in #3405's comment, and #3408's six added wrappers then shifted it 48 lines. Dropped the ordinal rather than updating it.

## Verification

- Docs gate clean — `SONIC_FRONTIERS_STATUS.md`: 10 tables, 82 rows, every row matching its header. `check_merge_result.py` clean.
- `git diff --check` clean. `--numstat`: `25/0` and `27/0` — **zero deletions** (trap 41). Diff is `.md`-only. No trailing or mid-line double whitespace.
- Every citation opened at this head: `live_compute.cpp:1171`, `:1328`, `:3233`, `:7293`, `:7496`, and `tests/fixtures/render_runner.h:4330`.

## Risk

Docs-only and append-only; no existing row rewritten. The risk is a **wrong** row, which would stop a future lane re-testing something live — which is precisely what the Void subsection now prevents for the one entry that would have had it.

Refs #3405
Refs #3407


mattias800 (commented):
## Independent review of #3409 — **REJECTED** (five concrete corrections; the verdicts themselves mostly survive)

Reviewed at head `52378b08` (one commit on top of `origin/main` = `ad5f4783`). I opened every cited line
in the tree at that head, re-derived both `parallel_compute_texels` counts from the parent commit, and
read #3405 and #3407 in full including all comments. The section this PR extends is the most
believed-without-checking artifact in the repository, so I have held it to that standard.

**The judgement in one line:** nine of the ten verdicts stand on their evidence, the two citations are
correct, and the mechanics are clean — but there are **five factual/presentational defects**, three of
them in numbers a future reader will act on, and one of them is contradicted by the merged commit
message this PR is describing. All five are cheap to fix; none requires a new measurement.

---

## What I verified green

- `frontends/shared/live/live_compute.cpp:3233` **is** `uint32_t host_memory_type(uint32_t bits) const`,
  and it really does try `HOST_VISIBLE|HOST_COHERENT|HOST_CACHED` in its first loop before falling back
  to `HOST_VISIBLE|HOST_COHERENT`. I also confirmed the claim is about the *right* allocation: the
  compute image staging buffer at `:7708` is `ctx.host_memory_type(sreq.memoryTypeBits)`. Citation
  correct and load-bearing.
- `:7293` **is** the `if (!bi.depth_bits_source && compute_sampled_guest_prepare_required(...))` guard,
  and `:7496` does state `!storage && !renderer_owned`. I brace-matched the block: it spans
  **7293–7683** and contains both `bi.cache_key = {` (`:7585`) and `ctx.acquire_cached_image(`
  (`:7650`). `compute_sampled_guest_prepare_required` (`live_compute.hpp:74`) returns
  `!storage_image && !renderer_owned && (...)`, so a renderer-owned sampled surface provably skips
  cache-key construction. The row's mechanism claim is exactly right.
- **The `seed_reprove` lever is real, not void.** `seed_reprove_interval_from_env`
  (`frontends/shared/texture/seed_reprove.hpp:42`) honours an exact `"0"` (and *only* an exact one —
  garbage keeps the 256 default), and `seed_reprove_due` (`:29`) is `if (interval == 0) return false;`.
  So `PROSPER_SEED_REPROVE=0` unconditionally removes re-proving. Combined with the row's own note that
  `[seed-skip-verify]` lines named the hot program's 4K textures in the ON arm — i.e. the mechanism was
  observed firing on the path under test — this is a genuine falsification, not identical arms from a
  dead lever. It also discriminates on a component timer (982.7 vs 980.7 ms), not only on fps. **Row 2
  stands as written.**
- Both size levers are honoured and **uncapped** at 8192: `persistent_compute_image_limit` (`:1171`)
  returns the override outright, and `compute_memory_pool_limit` (`:1328`) caps only at
  `UINT64_MAX/1MiB`. So "the larger value was not adopted" is not the explanation for those two nulls.
  (See finding 4 for what *is* worth saying about them.)
- **The "seven arms, one wrapped" claim is correct and I re-derived it independently.** In the parent
  (`68157c40`) the sampled chain's per-texel arms are `unorm2_10_10_10` (`:8144`), `sampled_uint8_native`
  (`:8184`), `sampled_unorm8x2` (`:8205`), `sampled_unorm16_native` (`:8224`), `f32` (`:8246`,
  the wrapped one), `source_r8 && r8` (`:8270`) and `source_unorm8` (`:8284`) — seven, one wrapped.
  The commit's net `+6` call sites (15 → 21 grep hits) corroborates it arithmetically.
- Mechanics, all re-run by me, not taken from the PR body:
  `git ls-files '*.md' -z | xargs -0 python3 prosper/tools/docs/check_numbered_table.py` → **exit 0**,
  `SONIC_FRONTIERS_STATUS.md: 10 table(s), 83 rows`. `check_merge_result.py` → **exit 0**, clean merge,
  `GAME_COMPAT_ORCHESTRATION.md` 269 rows unique/ascending with none deleted. `git diff --check` → exit 0.
  `--numstat` 17/0 and 10/0, **zero deletions**; `--name-only | grep -v '\.md$'` empty. Ruled-out section
  is now 41 rows, appended at end of the existing table, correct section.
- **Scope/placement is acceptable.** `DEBUGGING_WORKFLOWS.md` § Slow gameplay already carries
  title-anchored worked examples (instrument trap 269, the Blue Prince radeontop caveat), so a
  general technique with a Sonic Frontiers example is in keeping. The `#3407` rows are defensible in the
  title doc: `GRAPHICS.md`'s `## Ruled out` is explicitly scoped to the present/publish path and
  `RESOURCE_BINDING.md`'s to binding, so neither is a better home, and there is no live area doc for
  host-side renderer performance (`RENDERER_PERFORMANCE_2026_07.md` is a closed dated pass). **Non-blocking
  suggestion:** the two rows about shared code a future lane would grep for from *another* title — the
  write-combined row and the readback-pool row — should name their file, since neither lives in
  `live_compute.cpp` (`tests/fixtures/render_runner.h:4330` and `:4512`/`:4592`/`:4601`). Without a path,
  a lane optimising some other title's readback will never find them.

---

## Blocking findings

### 1. `DEBUGGING_WORKFLOWS.md`: "**fifteen** sites" is wrong — it is fourteen call sites, and the PR contradicts the very commit it is describing

`grep -c parallel_compute_texels` on `live_compute.cpp` at the parent gives **15**, but **line 548 is the
template definition** (`template <class Body> void parallel_compute_texels(...)`), so there are
**14 call sites**. Repo-wide there are no other users of the helper.

This is not my count against yours — it is the project record against itself. The merged commit
`ad5f4783`'s own fourth commit message says *"That switch serialises **ALL FOURTEEN** existing
`parallel_compute_texels` sites"*; #3405's outcome comment says *"used **fourteen** times"*; #3407's
cross-cutting comment says *"already used at **fourteen** sites in that same file"*. This PR is the only
artifact that says fifteen.

It matters more here than a stray digit normally would, because the paragraph's whole instruction is
*"Grep for the helper, then grep for the loop shape it wraps, and compare the counts"* — and a reader who
follows it literally gets 15 from `grep -c` and compares a helper-definition line against a population of
loops. The paragraph teaching the counting discipline is the one that must not model the off-by-one.

**Accepted replacement:** `already live at **fourteen** call sites in` (and, if you want the trap named,
append to the last sentence: `Count call sites, not grep hits — the helper's own definition is one of them.`)

### 2. `DEBUGGING_WORKFLOWS.md`: "three serial per-texel conversion loops were the cause" is wrong in both the count and the category

The source is #3405's *"Ordinary CPU work running on one core of thirty-two, in **three places**"*, and
those three places are: (1) staging buffers dropping their CPU mapping so ~8,192 pages refault per 33 MB
upload, (2) the sampled conversion chain's unwrapped arms, (3) a single-threaded 33 MB `memcpy` in
`render_runner.h`. **Only item 2 is a per-texel conversion loop**, and it is **six** loops, not three.
Item 1 is a mapping teardown with no loop in it at all; item 3 is a per-byte copy.

As written the sentence would send a reader looking for three conversion loops, and the fix for two of the
three causes (#3406, and the second half of #3408) is not a conversion loop.

**Accepted replacement for the final clause:**
`...which said "serial host work" before any symbol was read. The cause was three separate
single-threaded stretches, only one of them a conversion: a staging mapping torn down and refaulted every
frame (#3406), six unparallelised per-texel conversion arms, and a single-threaded 33 MB staging memcpy
(both #3408). #3405.`

### 3. Row 1 (`:655`) splices two different captures, and its arithmetic does not hold

The row reads `gpu-device` is **272 ms of a 4760 ms window (5.4%)**. 272/4760 = **5.7%**, not 5.4%.
The 272.1 ms / 5.4% pair is from #3405's **5.02 s** F8 window; the 4760 ms window is #3407's, where the
same timer reads **287 ms (6.0%)**. Two captures, one sentence.

Nobody's conclusion changes — but this is a row whose entire job is to be quoted without re-derivation, and
the first thing a careful reader does with three numbers is divide two of them. A row that fails that
check on line one discounts the nine below it.

**Accepted replacement:** `**No.** \`gpu-device\` is 272 ms of a 5020 ms F8 window (5.4%), the main thread
is 2.1% on \`drm_syncobj_array_wait_timeout\`, and process CPU is 1.01 of 32 cores. An independent
4760 ms window taken after #3406 reads 287 ms (6.0%), so the verdict does not rest on one capture. The GPU
is idle waiting for the host. #3405`

### 4. Row 5 (`:659`) quotes a treatment number with no control, and the nearest control in the table makes it read as an **18% win**

The row is `PROSPER_COMPUTE_MEMORY_POOL_MB=8192 left setup at 1130 ms and flips at 3.99/s.` The setup figure
has an implied baseline the reader must lift from the row above (1147 ms). **The flips figure has no baseline
anywhere in the row, in the doc, or in #3405's comment.** Meanwhile row 2, four lines up, states a baseline
of **3.39/s**. A reader who does the only comparison the table offers concludes the pool change bought
+18% and was written off — which is the exact shape of a "ruled out" row that gets re-derived.

(For what it is worth, I believe the true contemporaneous baseline was ~3.99/s — the arm was taken after
#3406, and #3407's budget table describes the same 4.76 s window "at 3.99 flips/s". But I am inferring
that, and a row must not require the reader to infer it.)

**Accepted fix:** state this arm's own paired baseline explicitly, or delete the flips number and keep only
`setup 1130 ms against a 1147 ms baseline`. Either is fine; the current form is not.

### 5. Row 10 (`:664`) does not label its arms, and under one of the two readings the treatment arm wins the primary metric

`flips overlapped (4.59/4.79/5.03 vs 5.04/4.98/4.98) and readback was worse by median (1195 vs 1261 ms)`.
Nothing says which triple is the control. Read baseline-first, the pooled arm has the **higher** flip median
(4.98 vs 4.79) and the worse readback timer — i.e. the two instruments disagree and the row's "No" rests on
the component timer over the primary metric, which the row does not say. Read pooled-first, both metrics
agree and the row is clean. The doc uses both orderings elsewhere (row 3 is treatment-then-baseline and
says so; row 4 is treatment-then-baseline and does not), so the reader cannot resolve it by convention.

**Accepted fix:** label them, e.g. `flips overlapped (baseline 4.59/4.79/5.03 vs pooled 5.04/4.98/4.98) and
readback was worse by median (baseline 1195 vs pooled 1261 ms)` — with the arms in whichever order is true.
If it *is* baseline-first, the row must also say that the flip median moved the other way and why the
readback timer is the one being believed.

---

## Non-blocking, but I would take them

### 6. Row 9's "90% of conversions" overclaims twice — and it does not need to

You asked me to attack this one specifically, so: **the verdict survives, the headline number does not.**

- **Population.** 90% is `895 of 994`, i.e. the two planes in #3407's fingerprint table. #3407's own body
  names **three** converting addresses (`0x2053960000`, `0x2033780000`, `0x2008180000`), each processed
  32 times; the third is absent from the fingerprint run with no explanation. Unqualified "90% of
  conversions" reads as "of all conversions in the run", which the measurement does not cover.
- **Method.** "distinct content" comes from **three 64-byte probes** of a 33 MB buffer. That is 192 sampled
  bytes out of 33 million. On a letterboxed 4K FMV plane, three fixed offsets can very plausibly sit in
  constant regions, so `distinct content = 1` is a **lower bound on distinctness** and 99.8% / 90% are
  **upper bounds on redundancy**. "Fingerprinting" is doing a lot of work in the row; a reader will hear
  "hashed".

The good news is that the falsification does not need either number. The **buffer-identity** measurement in
the same comment — 441 conversions against **148 distinct `LiveTargetSnapshot::pixels` buffers** per
address — is independent of any content sampling, and a 66% collapse is by itself fatal to "an FMV always
presents new content, so caching cannot help". Lead with that and the row becomes unattackable.

**Suggested replacement:** `**No.** Over a ~130 s run the *same published buffer* is re-converted about
three times on average — 441 conversions against 148 distinct \`LiveTargetSnapshot::pixels\` buffers per
address — so a pointer-identity cache collapses 66% of them with no content assumption at all. A
three-probe content fingerprint puts the redundancy far higher still (one 33 MB plane: 497 conversions,
1 distinct sample, another: 497 with 98), but that is a 192-byte sample of a 33 MB buffer over two of the
three converting addresses, so treat it as an upper bound on redundancy rather than a measured 90%. It is
the *dispatches* that repeat, not the frames. #3407`

### 7. Three rows report an implement-measure-revert null with no positive control that the intervention took effect

This is the class you named, and I want to be precise about which rows are actually exposed:

- **Row 3 (`linear_seed`) — already honest, and the honesty rescues it.** The row itself says
  `linear_seed` is only reached on the `atomic_image` and `zero_padded_tail` branches and that nothing
  established the hot program takes either. Note that this disclosure *strengthens* the "No": if the
  allocation is not on the hot path, the hypothesis "that allocation is the cost" is false either way.
  What is genuinely unestablished is only that the `thread_local` reuse actually removed the allocations
  (no allocation counter is quoted). Worth one clause — `the reuse's effect was not independently counted,
  so this falsifies the named allocation and not the broader "per-dispatch staging allocation" class,
  which #3406 then found to be real in a different form` — because the *next* thing that landed
  (#3406, staging mappings refaulting per frame) is precisely the broader class, and a reader who
  over-generalises this row would have talked themselves out of it.
- **Rows 4 and 5 (cache MB / pool MB) — the sweep is not what establishes the verdict, and the row should
  say what does.** I checked: both levers are honoured uncapped, so the arms were real arms. But raising a
  limit only tests "thrashing" if the old limit was ever hit, and no eviction count, discard count or pool
  hit rate is reported in either arm — the pool's hit rate, as row 5 itself notes, is not reported by
  anything. Meanwhile row 8, three lines below, contains the *structural* answer: the hot sampled planes
  never reach the cache gate at all (`:7293`), so no cache-size setting could have moved them. That is
  a much stronger falsification than a null sweep and it is already in this PR — say so.
  Suggested clause for row 4: `The limit is honoured uncapped (\`persistent_compute_image_limit\`,
  \`:1171\`), but no eviction count was read, so the sweep bounds the effect rather than proving the cache
  was healthy; the structural reason it cannot be thrashing is the row below — these planes never reach
  the cache gate.`
- **Row 10 (readback pooling)** has the same gap — no pooled-buffer reuse counter — but its verdict is
  already correctly narrowed to "**No, on its own**" with the 18.4%/81.6% split doing the work, so the
  scoping is fine once finding 5 is fixed.

### 8. Row 7 (`upload_mapping.unmap()`): give the sub-timer's total record count

`zero records above 0.5 ms` is exactly the shape that reads identically whether the timer fired 40,000
times and none exceeded the threshold, or the timer never fired at all. #3405's comment 5560096859 shows
the surrounding image-timing run had **284 image records**, so the denominator exists.
`**zero** of N records above 0.5 ms` closes it for free.

### 9. Row 6 (write-combined): say that it is a code-reading result, and give the second file a path

`host_memory_type` *prefers* `HOST_CACHED` and falls back when no such type is exposed; no run is quoted
confirming the selected `memoryTypeIndex`'s flags on this device. On RADV/STRIX_HALO that fallback is
very unlikely to bite, so I am not blocking — but the row asserts a runtime property from a static
preference, and one line of "the selected type reported `HOST_CACHED`" would make it a measurement.
Separately, `persistent_ds_transfer_buffer` is in `tests/fixtures/render_runner.h:4330`, not in
`live_compute.cpp` where the row's only path points; its own comment there is excellent evidence
(*"measured here at ~236 MB/s"*) and deserves to be findable.

---

## Verdict

**REJECTED** on findings 1–5. To be clear about proportion: this is a good PR doing the thing the charter
asks for, the two code citations are correct, nine of ten verdicts hold, and I found no row that
forecloses a live avenue on false evidence. What I found is one count that contradicts the merged commit
it describes, one that miscategorises the cause, one row that fails its own arithmetic, and two rows whose
numbers cannot be interpreted without an arm label the reader does not have. In a section whose contract is
"believe this and do not re-derive it", those are the defects that matter most — they are the ones a future
reader has no way to catch.

None needs a new run. Fix 1–5, take 6–9 or argue them, and I will re-review on the new head.

mattias800 (commented):
## Re-review of #3409 at `59986e4e` — **REJECTED**, on one stale line number that this round introduced

Everything else is fixed and I verified each fix against the tree rather than against your summary.
The single blocking item is one token, and I have pre-committed below to approving a head whose only
change is that token — I am not opening a new round of substance.

---

## Blocking: the `upload_mapping.unmap()` row now cites a line that is not `unmap()`

Row 7 gained `A dedicated sub-timer at \`live_compute.cpp:8557\``. At this head:

```
$ grep -n 'upload_mapping.unmap()' live_compute.cpp
8605:            upload_mapping.unmap();
```

**`:8557` is a `parallel_compute_texels` lambda opener inside the `f32` native-float staging branch**
(`} else if (f32) { parallel_compute_texels(texels, linear_bytes + texels * 16u, [&](size_t begin, size_t end) {`)
— related code, wrong statement, and confusing in a specific way: an unwary reader lands on a
parallelised conversion arm while reading a row about a mapping teardown.

The drift has a clean cause worth recording, because it will recur: **`:8557` is correct in #3405's
comment, which was written before #3408 shifted this file.** #3408's six added
`parallel_compute_texels` wrappers pushed everything below `:8141` down by 48 lines — the same shift
that moved `:8508`/`:8525` to `:8556`/`:8573`. So the general caution is *a file:line lifted from an
issue comment is stale by construction whenever the same file has been edited since*, and this
investigation edited this file twice between the comment and this PR.

The previous revision of the row carried no line citation at all, so this is a regression introduced
by the correction round. That is the whole reason I am not waving it through: an uncited row is merely
less helpful, a miscited one asserts something false with the authority of having been looked up, and
this section is read before every future investigation on this title.

**Two acceptable fixes, either is fine:**

- `A dedicated sub-timer at \`live_compute.cpp:8605\``, or
- **drop the number** — `A dedicated sub-timer on \`upload_mapping.unmap()\`` is already unambiguous
  (one call site in the file) and cannot drift. I mildly prefer this one; the row's value does not
  depend on the ordinal, and the next lane to touch that function breaks the number again.

**I will approve a head that changes only this.** Nothing below is blocking.

---

## Verified fixed — each one re-derived, not read off your message

- **Finding 1.** `fourteen call sites`, and the trap is now named in the paragraph itself. Re-confirmed
  the ground truth: 15 grep hits at `ad5f4783^`, `:548` being
  `template <class Body> void parallel_compute_texels(...)`, no users in any other file. The added
  clause ("which is how this paragraph first said fifteen") is the right instinct — a technique
  paragraph that records its own instance of the failure is worth more than one that merely asserts it.
- **Finding 2.** "three separate single-threaded stretches and only one of them was a conversion",
  with the mapping refault attributed to #3406 and the arms + memcpy to #3408. Checked the attribution:
  #3408 is the two-commit PR carrying exactly those two, so "(both #3408)" is right.
- **Finding 3.** 272.1 / 5020 = 5.42% ✓, 287 / 4760 = 6.03% ✓, and the explicit "do not pair 272 ms
  with the 4760 ms window" is a better fix than I asked for — it makes the row resistant to being
  re-broken by a future editor compressing it, which is how the splice would come back.
- **Finding 4.** The uncontrolled 3.99/s is gone, the controlled `1130 vs 1147 ms` pair stands, and the
  parenthetical records why the number was removed. Keeping the retraction visible is correct here:
  the figure is still live in #3405's comment thread, so a future lane will meet it and needs to know
  it was considered and dropped rather than overlooked.
- **Finding 5** — see below, this is now the best row in the set.
- **Non-blocking 6–9** all landed: row 3 scoped against over-generalisation with the #3406 pointer;
  rows 4/5 record that both levers are honoured uncapped *and* that no eviction count / hit rate was
  read, and route the reader to the structural falsification; row 6 marked a code-reading result and
  given `tests/fixtures/render_runner.h:4330`; row 9 leads with 441 → 148 and labels the fingerprint an
  upper bound over two of three addresses.
- **Citations re-opened at this head**, all correct except the one above: `live_compute.cpp:1171`
  (`persistent_compute_image_limit`), `:1328` (`compute_memory_pool_limit`), `:3233`
  (`host_memory_type`), `:7293` (the `compute_sampled_guest_prepare_required` guard),
  `render_runner.h:4330` (`persistent_ds_transfer_buffer`).
- **Mechanics re-run at `59986e4e`**, not carried over: docs gate exit 0 over every tracked `.md`
  (`SONIC_FRONTIERS_STATUS.md` 10 tables / 83 rows, `DEBUGGING_WORKFLOWS.md` 1 table / 6 rows, zero
  non-clean lines in the whole run — the escaped `\|` inside row 6's backticks does not break the
  table); `check_merge_result.py` exit 0, clean merge, 269 orchestration rows unique/ascending with
  none deleted; `git diff --check` exit 0; `--numstat` 21/0 and 10/0, **zero deletions**; no non-`.md`
  path in `--name-only`. Ruled-out section is 41 rows.
- **The positional citation you fixed unprompted was worth fixing**, and I had missed it. "Two rows
  below" in a table that every lane appends to is a reference that decays silently; naming the row
  *and* restating the mechanism (`:7293`, renderer-owned → guest-source block skipped) means the row
  survives even if the referenced row is later split or moved.

---

## Row 10: I agree this is now the strongest row, and one note on why it is safe

You are right that recording it as a falsification was the failure mode this section exists to prevent,
and you were right to reopen it rather than defend it.

One thing worth knowing, because it makes the rewrite sturdier than it looks: **I cannot independently
verify the arm assignment from the project record** — #3407's comment gives
`(4.59/4.79/5.03 vs 5.04/4.98/4.98)` with no labels and the raw run logs are not attached, so
"baseline first" rests on your memory of the runs. That is fine, because **the verdict is now
labelling-independent**: read first-as-pooled instead and pooled is worse on flips (4.79 vs 4.98) and
*better* on the readback timer (1195 vs 1261) — the two instruments still disagree, just with the
signs swapped. UNDECIDED holds either way. A row whose conclusion survives its own worst ambiguity is
in much better shape than one that needed the labelling to be right, and it is worth one clause saying
so, so nobody later "corrects" the arm order and thinks they have overturned the row.

---

## Your question 1: in-row label, or grow the subsection?

**Grow the subsection**, and copy GTA5's shape rather than inventing one.

The argument is about what a skimmer reads first. A reader arriving at this document with "has anyone
tried pooling the readback buffer?" reads two things: the `## Ruled out` heading, and the left cell
`Pooling the readback destination buffer is worth doing`. Both say "settled" before the verdict cell is
reached — and the whole hazard you are guarding against is the reader who stops early because the
answer agrees with what they expected. The in-row `**UNDECIDED, not falsified**` is excellent, but it
is the third thing read, and the heading is the first.

The GTA5 precedent (`GTA5_STATUS.md:5445`) is worth copying exactly, including two details that are
easy to get wrong:

- It is `### Void, not falsified — do not cite these as settled` — **the heading carries the
  instruction**, not just the classification. That is what does the work.
- It is a **bulleted list, not a second table**, sitting after the table. That matters more than it
  looks: a 41-row table that every lane appends to should not acquire a sibling table, because then
  every future append needs a "which table?" decision, and appends are where this document's recorded
  failures come from. Bullets keep the default unambiguous — falsified rows go in the table, and a
  result that is not falsified goes in the list the heading names.

It also carries exactly **one** entry today, so a one-entry subsection is already this project's
accepted shape and you are not over-engineering by creating it.

Concretely: move row 10 out of the table into that subsection, and if you want the heading-level signal
without a forward reference, one line under `## Ruled out` saying that not everything recorded here is
falsified and to read the subsection below. When someone settles it, the entry becomes a table row.

I would not block on this either way — but you asked for judgement, and the judgement is that the
in-row label is doing the right job in the wrong position.

## Your question 2: does anything in the two paragraphs still overreach?

**The second paragraph does not.** It generalises cleanly, it names the counting discipline, it now
records its own instance of getting the count wrong, and its one arguable soft spot — "grep for the
loop shape it wraps" is only greppable when the shape is regular — is honest about being a heuristic.
Leave it.

**The first paragraph has one real overreach, and it is in the strongest sentence.** "it rules out a
whole class of answers" is true, but the class is narrower than the paragraph implies. A ~1.0-core
ratio rules out *parallel CPU saturation* — it establishes **serialisation**, not that serial **host
work** is the bottleneck. A process sitting at 1.0 cores can equally be blocked on the GPU with one
busy poller, blocked on I/O, or waiting on the guest, and in every one of those the honest next
question is "waiting on what", not "why is only one thread doing this".

What actually licenses the Sonic Frontiers inference is the *pair*: 1.01 cores **and** `gpu-device` at
5.4% **and** the main thread at 2.1% on the fence. All three are in the paragraph — but as parts of the
example, not as part of the technique, so a reader lifting the technique alone can carry it to a
GPU-bound title and conclude "serial host work" from a number that cannot say that.

One clause fixes it. Suggested, to append after "...rules out a whole class of answers":

> Read it with the GPU-side number, because on its own it establishes *serialisation*, not that the
> serial work is yours: ~1.0 cores against an idle `gpu-device` says serial host work, while ~1.0 cores
> with a busy device or a main thread parked on a fence says you are waiting, and the ratio has not told
> you on what.

Also non-blocking, and the last of the population caveats: row 9 now says "**each** converting address
shows 441 conversions against **148** distinct buffers". The record has two addresses at 441 with
**148 and 144** distinct buffers, and #3407's body names **three** converting addresses. "Each" and the
single figure quietly re-broaden the population the rest of the row is careful to bound. `441
conversions against 148 and 144 distinct buffers on the two instrumented addresses` costs four words
and closes it.

---

## Verdict

**REJECTED** — solely on the `live_compute.cpp:8557` → `:8605` citation (or drop the ordinal). All five
blocking findings from the first round are correctly fixed, all four non-blocking ones were taken and
improved on, every other citation opens to what the row says it does, and every gate is green at
`59986e4e`. Push the one-token fix and I will approve on that head without further findings; the
subsection, the core-count clause and the 148/144 wording are yours to take or argue.

mattias800 (commented):
## Re-review of #3409 at `3fbffaa8` — **APPROVED**

The blocker is fixed, both answers were taken, and I re-derived the delta rather than reading your
summary. `git diff 59986e4e..3fbffaa8` is exactly three things — the citation, the core-count
paragraph, and the Void subsection — so everything I verified at the previous head still stands
untouched.

### The citation

`grep -n 'upload_mapping.unmap()'` gives one call site, at `:8605`. The row now reads
`A dedicated sub-timer on \`upload_mapping.unmap()\``, and `grep -rn 8557 prosper/docs/` returns
nothing. Dropping the ordinal rather than updating it was the right call for the reason you give: one
call site, a name that cannot drift, and the row loses nothing it was carrying.

Your generalisation is the durable part and I would push it one step further, because it is not
specific to issue comments: **any file:line is a claim about a tree, and a doc PR is almost never
written against the tree the number came from.** The recognition cue is cheap — if the cited file has
any commit between the source of the number and the PR, re-open it. Worth a line in
`DEBUGGING_WORKFLOWS.md` or a trap row at some point; not in this PR.

### The Void subsection

Correct in both details, and the entry is better than what I proposed. Declining to assert the arm
labelling at all — "the two instruments disagree, and the verdict does not depend on which arm is
which... with only the signs swapping" — is stronger than labelling them from memory would have been,
because it makes the row unfalsifiable-by-relabelling rather than merely correct. The parenthetical
warning against "correcting" the order and thinking the verdict is overturned is the sentence that
will actually stop the re-derivation.

Structure verified: it is a bulleted list after the table, not a sibling table, and the heading carries
the instruction. The Ruled-out table is back to 38 data rows (40 `^|` lines incl. header/separator) from
main's 29, i.e. nine appended rows plus the one Void entry — see the note below about the PR body.

### The core-count overreach

The split into two paragraphs is a better fix than the clause I suggested. "On its own it establishes
serialisation, not that serial host work is the bottleneck — a process blocked on the GPU or on I/O
reads about one core too, and reads it for the opposite reason" states the failure mode rather than
just guarding against it, and naming all three legs (1.01 cores, `gpu-device` 5.4%, the fence at 2.1%)
inside the technique rather than the example is what makes it liftable. Paragraph 2 unchanged, correct.

### Gates re-run at this head

Docs gate exit 0 over every tracked `.md`, zero non-clean lines. `check_merge_result.py` exit 0, clean
merge. `git diff --check` exit 0. `--numstat` 25/0 and 27/0, **zero deletions**; no non-`.md` path.
Whitespace swept independently: no trailing whitespace and no genuine mid-line double spaces in the
added lines (the sixteen you found are gone).

---

### Two things to do before you merge — neither is a finding against the diff

**1. The PR body is now stale in three places, and one of them contradicts the doc.** The body still
says *"ten rows appended (31 → 41)"* and *"`--numstat`: `17/0` and `10/0`"`, and its summary table
still lists `Pooling the readback destination buffer is worth doing | Implemented, **reverted**` under
a header reading *"How it died"* — which is precisely the reading this PR spent two rounds removing
from the doc. The body outlives the branch on the PR page; someone following the link from a squash
commit lands on it. `gh pr edit` fails here on the Projects-classic GraphQL error without updating
anything, so use `gh api -X PATCH .../pulls/3409 -f body=...` and verify the result.

Row counts for the corrected body, since they were miscounted originally too: main's Ruled-out table
holds **29** data rows and this PR takes it to **38**, plus one entry in the new Void subsection.
The "31 → 41" figures counted the header and separator lines as rows.

**2. Carry-over from round two, not taken and still a four-word fix.** Row 9 says *"each converting
address shows 441 conversions against 148 distinct buffers"*. The record has **148 and 144** on the
**two** instrumented addresses, while #3407's body names **three** converting addresses — so "each"
and the single figure quietly re-broaden the population that the row's very next clause is careful to
bound. `441 conversions against 148 and 144 distinct buffers on the two instrumented addresses` closes
it. I flagged this non-blocking in round two and I am not raising it to blocking now; take it if you
touch the file.

---

## Verdict

**APPROVED** at `3fbffaa8`. Every citation in the ten entries opens to what its row says it does
(`live_compute.cpp:1171`, `:1328`, `:3233`, `:7293`, `render_runner.h:4330`, and now no ordinal on the
unmap row); the two counts I could re-derive independently (fourteen call sites, seven arms with one
wrapped) are right; the one row that was not a falsification now says so under a heading that says so;
and the technique paragraphs no longer claim more than the ratio can support.

Worth saying plainly, since three rounds of findings can read as a verdict on the work: this went from
a good PR with five soft numbers to one where the weakest row explicitly explains why it is weak, and
that is the direction this section is supposed to move. The readback entry in particular is now more
useful than a clean "No" would have been.

Issue #3411, opened 2026-09-06
Sonic Frontiers (PPSA03831): no audio during the intro FMV or title screen — AudioOut2 is fully registered, so it is a PCM submission/routing question


mattias800:
## Symptom

The intro FMV and the title screen play with **no audio**. Two independent human observations, three weeks apart, on different heads:

- **2026-08-29**, manual test session (#1891): *"incredibly low framerate, even at the initial splash screen and studio FMV. also, there is no audio, which I think there should be?"*
- **2026-09-06**, project owner, on a default `prosper-app` launch after the performance work in #3405: no audio during the intro FMV.

The 2026-08-29 report bundled it with the frame rate, and the frame rate is what got investigated (#3405 / #3407, three PRs merged, 3.38 → 4.73 fps). The audio half was never picked up. Filing it separately so it stops riding along with a performance issue — the two have no established connection.

Nothing in a default run's log reports an audio error.

## What is already ruled out

**It is not an unregistered-NID problem.** The title imports thirteen `libSceAudioOut2` entry points, and **all thirteen are registered** (`prosper/src/hle/audio/hle_audio.cpp:4194-4223` registers 29 AudioOut2 entry points, and I checked all thirteen the title imports by name):

```
nid_census <DUMP_ROOT>/PPSA03831-app0 --names <PS5_LIBS> --lib Audio
  -> scope: 209 unregistered, 0 shown, --lib filter=Audio
```

Zero unregistered audio imports. Worth stating explicitly because the unregistered-NID mechanism is the highest-prior cause on this project (#2023, #2951, #1873) and because **`nid_census --registered` reports the opposite** — it prints registered rows under a header saying they have no handler, which is filed separately as #3410. Run it *without* `--registered` on this question.

**It is not a missing middleware PRX** (#2974's mechanism). The dump ships exactly three: `libc.prx`, `libSceJobManager.prx`, `libSceNpCppWebApi.prx`. None is audio, and the eboot names no `libSceAjm`, `libSceAvPlayer`, `libSceNgs2` or `libSceVideodec`. Audio goes through `libSceAudioOut2` directly, i.e. the title submits its own PCM rather than handing decode to Sony middleware.

That is the useful narrowing: **this is a PCM submission/routing question, not a decoder question.**

## Next step — one run answers it

`PROSPER_AUDIO_FLOW=1` exists for exactly this and separates the three causes that look identical from outside (`hle_audio.cpp:645-667`):

```bash
[...]

Issue #3410, opened 2026-09-06
nid_census prints REGISTERED imports under a header saying they have no handler, so --registered manufactures false "unregistered NID" findings


mattias800:
`prosper/tools/nid_census/nid_census.cpp:316` prints this header unconditionally:

```
== imports with NO registered handler -> dispatcher returns 0 ==
```

With `--registered`, `selected` contains **registered** imports as well (`:281` — `if (registered && !show_registered) continue;`), and the text-mode row loop at `:318-323` prints **no registered column**. So every registered import is displayed under a heading asserting the exact opposite of its state, with nothing in the row to contradict it.

The TSV path is correct: `:301` emits a `registered` header and `:308` emits `Hle::registered(r->nid) ? 1 : 0` per row. **Only the human-readable path lies — which is the one an agent reads.**

## Failure scenario (measured today, 2026-09-06)

Investigating "no audio during the Sonic Frontiers intro FMV", to find out which audio entry points the title imports:

```
prosper/build-linux/nid_census <DUMP_ROOT>/PPSA03831-app0 --registered --names <PS5_LIBS> --lib Audio
```

```
== imports with NO registered handler -> dispatcher returns 0 ==
NID           name                                     #ttl  import library
PE2zHMqLSHs   sceAudioOut2ContextAdvance                  1  libSceAudioOut2
0x6o1VVAYSY   sceAudioOut2ContextCreate                   1  libSceAudioOut2
...                                                                (13 rows)
scope: 209 unregistered, 13 shown, --lib filter=Audio
```

Read literally, that says all thirteen `libSceAudioOut2` entry points the title imports are unregistered and answer `0` — which, since `0` is `SCE_OK`, is a complete and very plausible mechanism for silent no-audio: `sceAudioOut2ContextCreate` "succeeds" without writing a handle, the guest pushes into nothing.

**It is false.** All thirteen are registered — `prosper/src/hle/audio/hle_audio.cpp:4194-4223` registers 29 AudioOut2 entry points, and all thirteen the title imports are present there, checked one by one by name. Re-running **without** `--registered` gives the truth:

```
scope: 209 unregistered, 0 shown, --lib filter=Audio
```

Zero. The scope line is the only thing in the text output that can distinguish the two runs, and it does not say how many of the *shown* rows are unregistered — it reports the repo-wide total (209), so it reads as corroborating.

## Why this one is worth fixing rather than documenting

[...]

Issue #2023, opened 2026-08-05
Sonic Frontiers (PPSA03831): after the opening sequence the guest composites nothing — its own flipped display buffer is uniformly black


mattias800:
Successor to #1968 §5, which is answered: the present path is no longer the limiter for this title.

## What is established

Sonic Frontiers never renders into its scanout with a draw pass — a `PROSPER_PASS_LOG` census over a
full boot records `vo=1` on **0 of ~3,700 passes**, in every phase. Its composite reaches the display
buffer by a non-draw path, in SW_64KB_R_X tiled layout. With the flipped buffer published (PR for
#1968), prosper renders the whole opening sequence at 3840x2160: SEGA logo, middleware credits, Sonic
Team logo, Cyber Space warp.

## The remaining blocker

**The guest's own display buffer goes black, and stays black.** One arm, native Linux/Vulkan,
`tools/screenshot`, default launch with no input,
`PROSPER_GUEST_FS=1 PROSPER_GUEST_ARGS=-force-gfx-direct PROSPER_RENDER=1 PROSPER_DUMP_SCANOUT=<dir>`,
defaults for scale and `RENDER_EVERY`. Each dump de-swizzled offline under SW_64KB_R_X:

| guest flip | `rgb_nonblack` (of 8,294,400) | what it is |
| --- | --- | --- |
| 60 | 537,467 | SEGA logo |
| 120 | 0 | between logos |
| 180 | 6,543,777 | Cyber Space intro |
| 240 | 782,935 | middleware credits |
| 300 | 475,166 | credits |
| 360 | 474,453 | credits |
| **420** | **0** | — |
| **480** | **0** | — |

A longer arm shows the same wall in the published frame: content through t ≈ 160 s, then
`distinct_rgb_colors == 1, nonblack_rgb_pixels == 0` from t = 200 s to t = 680 s while `frame_seq`
climbs 3,419 → 12,332 and `present_count` climbs 489 → 1,762.

The buffer is **written**, not stale: alpha is non-zero across all 8.2 M texels on every dump,
including the black ones. Something actively clears it each frame and composites nothing into it.

## Why this is upstream, not renderer-side

The same non-draw writer produces correct frames at flips 60 through 360 and black ones at 420+, in
one run, with no renderer state change between them. A prosper failure to execute the writer would
leave the previous contents, not opaque black.

This lines up with #1968's original content-loading evidence, which is unaffected by everything since:
[...]

PR #3412, opened 2026-09-06
docs: instrument trap 270 (nid_census --registered), and the BLOG entry #3408 owed

mattias800:
Two gaps found while answering *"where should this investigation be recorded?"*.

## Instrument trap 270 — `nid_census --registered`

Asking which audio entry points *Sonic Frontiers* imports:

```
nid_census <DUMP_ROOT>/PPSA03831-app0 --registered --names <PS5_LIBS> --lib Audio
```

listed all thirteen `libSceAudioOut2` NIDs under

```
== imports with NO registered handler -> dispatcher returns 0 ==
```

Read literally that is a complete and very plausible silent-no-audio mechanism, because `0` is `SCE_OK`: `sceAudioOut2ContextCreate` "succeeds" without writing a handle and the guest pushes into nothing. I was one command from filing it as the cause of #3411.

**All thirteen are registered** (`hle_audio.cpp:4193-4215`). Re-running *without* `--registered` reports `0 shown`. The header at `nid_census.cpp:316` is unconditional while `--registered` widens the selection, and the text row loop carries no `registered` column although the TSV path does (`:301`, `:308`). The scope line reads as corroborating rather than contradicting, because its unregistered count is repo-wide (209) rather than scoped to the shown rows.

The code fix is #3410. This row is for the *reading*, which survives until that lands and is what a future agent needs.

Why it earns a row rather than only a bug: an unregistered NID answering `SCE_OK` is the mechanism behind #2023, #2951 and #1873. An instrument that fabricates one is expensive exactly in proportion to how good this project is at acting on it — the finding is exciting, it fits, and nobody re-derives a conclusion they already accept.

## BLOG entry owed by #3408

#3408 was the largest single performance result measured on this title and shipped no `BLOG.md` entry, so the one audience that reads that file never saw it. Written as a finding rather than a screenshot, per the charter's finding-with-no-picture rule; ~75 words against the file's ~60 median.

## Verification

```
F=prosper/docs/GAME_COMPAT_ORCHESTRATION.md; B=$(mktemp)
git fetch origin -q && git show "origin/main:$F" > "$B"
python3 prosper/tools/docs/check_numbered_table.py --ordered --table-header Instrument --baseline "$B" "$F"
```
```
14 table(s), 332 rows, unbroken, every row matching its header
numbered column 1..270, 270 rows, unique and ascending, no unused numbers;
none of the 269 rows in the baseline has been deleted
```

`git diff --check` clean. `--numstat`: `9/0` and `1/0` — **zero deletions**. Diff is `.md`-only.

Trap number allocated with `trap_number.py` (scans `main` and every open PR); it reported 270 free. `check_merge_result.py` will be re-run immediately before merge, since that is the only check that sees a concurrent append.

Refs #3410
Refs #3407


mattias800 (commented):
## Review of #3412 at `7bcb8e4a` — **REJECTED**, on the BLOG entry's fps range (one sentence)

Trap 270 is fully verified and is a strong row — I reproduced it end to end rather than reading it.
The BLOG entry has two problems in one sentence, one of which I would have let go and one of which I
would not.

---

## Trap 270 — verified independently, including the claim you said it rests on

I ran the census myself against the dump, and it reproduces exactly:

```
$ nid_census <DUMP_ROOT>/PPSA03831-app0 --names <PS5_LIBS> --registered --lib Audio
== imports with NO registered handler -> dispatcher returns 0 ==
PE2zHMqLSHs   sceAudioOut2ContextAdvance            ...
[13 rows, all libSceAudioOut2]
scope: 211 unregistered, 13 shown, --lib filter=Audio
```

and without `--registered`: **`0 shown`**, empty table. So the header is unconditional over a widened
selection, exactly as the row says.

**The per-name check is the claim the row rests on, so I redid it rather than checking the range.**
All thirteen imported names appear in the registration block, one by one:
`Initialize` `:4194`, `ContextResetParam` `:4195`, `ContextQueryMemory` `:4196`, `ContextCreate`
`:4197`, `ContextDestroy` `:4198`, `ContextAdvance` `:4199`, `ContextPush` `:4200`,
`ContextGetQueueLevel` `:4201`, `UserCreate` `:4204`, `UserDestroy` `:4205`, `PortCreate` `:4206`,
`PortDestroy` `:4207`, `PortSetAttributes` `:4209`. **13 of 13 registered.** The corrected range
`:4194-4223` is right, and so is "29 registrations" — the block is 30 lines with
`GetSpeakerArrayAmbisonicsCoefficients` wrapped across two.

The tool citations also open to what the row says: `nid_census.cpp:316` is the unconditional
`printf("\n== imports with NO registered handler ...")`, the text row loop at `:318-323` prints
NID/name/#ttl/libs with no `registered` column, and the TSV path does carry one (`:301` header,
`:308` `Hle::registered(r->nid)`). Gates: baseline+ordered gate exit 0 (`1..270`, unique and
ascending, none of the 269 baseline rows deleted), full md gate exit 0 with zero non-clean lines,
`check_merge_result.py` exit 0, `diff --check` exit 0, `9/0` and `1/0`, `.md`-only.

Two small notes, neither blocking:

- **I got `211 unregistered` where the row says 209.** That is my binary, not your row — I ran a
  build from 26 Aug, which has fewer registrations than current master. It is worth knowing that this
  particular number moves with the binary, which if anything reinforces the row's point that it is a
  repo-wide figure and not scoped to the shown rows.
- The row's repro command omits `--names`, so a reader following it literally gets thirteen rows with
  `?` in the name column. The import-library column still says `libSceAudioOut2` so the trap still
  fires, but `--names <PS5_LIBS>` is what makes the output look like the one quoted.

This row is worth its number. The reason it is expensive is stated correctly and generally — an
instrument that fabricates the one mechanism this project is best at acting on costs in proportion to
that skill — and it names the *reading* rather than only the bug, which is what keeps it useful until
#3410 lands.

---

## Blocking: the BLOG entry attributes #3406's gain to #3408

> *"Wiring them up halves the compute time, and giving the texture staging copy the same treatment
> takes the intro from 3.4 to 4.7 fps."*

The subject of "takes the intro from 3.4 to 4.7" is the two #3408 changes. Their own range is
**3.6 → 4.7**. From #3405's own table:

| stage | flips/s | merged as |
| --- | --- | --- |
| at filing | 3.38 | — |
| texture staging keeps its mapping | 3.59 | #3406 |
| conversion arms parallelised | 4.18 | #3408 |
| staging copy parallelised | 4.73 | #3408 |

3.4 is the **pre-#3406** figure. And `BLOG.md` already carries #3406's share on the same day, four
lines below this entry: *"lifts the frame rate about 6%"* — which is exactly 3.38 → 3.59. So the two
entries overlap, and a reader adding them gets more than happened.

I am blocking on this rather than noting it because of where it sits. #3408's **fourth commit exists
solely to correct an over-attribution of this same measurement** — "the control arm is strictly more
serial than master, so +18.7% is an UPPER BOUND on the change's own effect and not a measurement of
it... The error is one-directional and flattering, which is why it needs saying." This entry commits
the same one-directional error on the same number, in the one file the project owner actually reads,
in the PR that exists to give #3408 its record. That is the wrong place for it to survive.

**Either fix is fine:**

- `takes the intro from 3.6 to 4.7 fps` — the #3408-only range, matching the sentence's subject; or
- `and together with the mapping fix earlier the same day takes the intro from 3.4 to 4.7 fps` —
  keeps the headline number and makes the aggregate explicit, which is arguably the better blog
  sentence.

## Also, and this is the one you asked me to check: "thirteen other sites" does not survive its own sentence

You asked me to confirm the reasoning rather than the digit, so: **the reasoning is arithmetically
sound and the sentence does not carry it.**

Your derivation — 14 call sites before #3408, one of which is the wrapped `f32` arm inside the sampled
chain, leaving 13 elsewhere — is correct, and I re-derived it: at `ad5f4783^` the call sites are
`:603 :619 :3553 :3620 :3663 :8014 :8246 :8508 :8525 :10367 :11396 :11418 :11443 :11510` (14), of
which `:8246` is the chain's own wrapped arm. 14 − 1 = 13. It also checks out forward: 13 + 1 + 6 = 20
call sites after, matching 21 grep hits minus the definition.

But the sentence is *"used at thirteen other sites, just never applied to these six"*, and that
contrast puts everything the helper **was** used at on one side and the six it wasn't on the other.
Read that way — which is the natural reading, since the exclusion is never stated — the number is
**fourteen**. A reader who tries to reconstruct 13 has to already know that one of the 14 is inside
the chain being discussed, which the entry does not say.

And excluding it makes the story weaker, because that arm is the whole point: the helper was not
merely nearby, it was **in the same branch chain, one arm away**. Suggested, and it is shorter than
what is there:

> *used thirteen times elsewhere in the file and once inside this very chain, just never applied to
> these six.*

That is defensible under every reading and lands the irony the `DEBUGGING_WORKFLOWS.md` paragraph is
built on.

---

## Everything else about the BLOG entry is right

Placement is correct — newest first inside `## 2026-09-06`, above the #3406 entry. It does not
duplicate that entry: #3406 is the mapping, this is the parallelisation, different changes. ~85 words
against the file's neighbours at 90-100, so the length is in family, not the 330-946-word shape the
charter calls out. The title says what you are looking at rather than what the commit did. The
finding-with-no-picture framing is right for a scheduling fix with no visual delta, and the closing
pointer to #3407 for "architectural rather than a missed optimisation" is the honest note that stops
it reading as a victory lap.

## Verdict

**REJECTED** at `7bcb8e4a`, on the `3.4 to 4.7` attribution; take the "thirteen other sites" rewording
in the same edit. Trap 270 needs no changes — I verified all thirteen registrations by name, both tool
citations, the corrected `:4194-4223` range and the 29 count, and every gate is green. Push the
one-sentence fix and I will approve. Re-run `check_merge_result.py` immediately before merging, as you
note — it is the only check that can see another lane's trap row landing at 271.

mattias800 (commented):
## Re-review of #3412 at `ecacfaf7` — **APPROVED**

The delta from `7bcb8e4a` is exactly the two items and nothing else: `git diff` between the heads
touches `BLOG.md` alone, and the trap 270 row is byte-identical to the one I reproduced end to end
last round (`git diff 7bcb8e4a..ecacfaf7 -- GAME_COMPAT_ORCHESTRATION.md` is empty). So nothing I
verified before needs re-verifying, and I re-checked only what moved.

### The attribution fix

`from 3.6 to 4.7 fps on top of the mapping fix below` is right, and the second half is what makes it
right rather than merely less wrong. #3408's own range is 3.59 → 4.73; the entry four lines below
carries 3.38 → 3.59 as "about 6%"; the two now compose to the true 3.38 → 4.73 with no overlap,
and the pointer means a reader who wants the aggregate can build it instead of being handed a number
that already contains someone else's change.

Worth noting what the entry does **not** say, since it is the harder half of that commit's
instruction: it quotes "halves the compute time" and never quotes +18.7%. That is exactly the split
#3408's fourth commit asked for — the mechanism-matched claim kept, the flattering upper bound
dropped — and it is the first artifact in this chain to get it right without being told.

### The count

Verbatim, and I re-derived it once more against `ad5f4783^` rather than trusting the previous round:
14 call sites, `:8246` the chain's own wrapped arm, so thirteen elsewhere plus one inside. 13 + 1 + 6
= 20 after, matching 21 grep hits minus the definition. The sentence now states the exclusion instead
of requiring the reader to infer it.

### Gates, re-run by me against the CURRENT `main` (`779cc6df`), not carried over

This matters here because `main` moved under the branch when #3409 landed:

- Ordered + baseline gate on `GAME_COMPAT_ORCHESTRATION.md`: **exit 0**, `1..270` unique and
  ascending, no unused numbers, none of the 269 baseline rows deleted.
- `check_merge_result.py`: **exit 0**, clean merge, the merged result passing the same gate against
  `origin/main` as of now.
- Full `.md` gate: **exit 0**, zero non-clean lines across every tracked file.
- `git diff --check`: exit 0.

**One thing to know before you merge, because it looks alarming and is not.** The branch is now one
commit behind `main`, so a *two-dot* `git diff origin/main..ecacfaf7` reports `0/25` and `0/27`
against `DEBUGGING_WORKFLOWS.md` and `SONIC_FRONTIERS_STATUS.md` — i.e. it reads as if this PR
deletes #3409. It does not. Against the merge base (`ad5f4783`) the diff is **`10/0` and `1/0`, zero
deletions, `.md`-only**, the two files do not overlap with #3409's at all, and the merge-result gate
confirms the merged tree keeps everything. I am recording it because the trap-41 check this project
runs (`--numstat`, expect zero deletions) will show non-zero deletions on any branch whose base has
moved, and the honest reading is "use the merge base", not "something ate the other lane's rows".

### Hygiene

No `Claude-Session:` trailer and no "Generated with Claude Code" line on any of the three branch
commits.

## Verdict

**APPROVED** at `ecacfaf7`. Trap 270 verified end to end last round — census reproduced against the
dump, all thirteen imported NIDs matched by name to `hle_audio.cpp:4194-4223`, the 29-registration
count and all three `nid_census.cpp` citations opened — and unchanged since. The BLOG entry's four
checkable numbers all hold: 1.01 of 32 cores, fourteen pre-existing call sites split 13 + 1, 3.59 →
4.73 attributed to the right two commits, and 4.73 against a 30 fps target as "six times slower".

Merge promptly — the merge-result gate is the only check that sees a concurrent append to that table,
and it is stale the moment another lane lands a trap at 271.

PR #1701, opened 2026-08-01
ci(docs): fail when a numbered doc table is broken, and repair a live break

mattias800:
## Purpose

The instrument-trap table in `docs/GAME_COMPAT_ORCHESTRATION.md` is the page the working guide points
agents at first, and several lanes append to it concurrently. **Two defects landed in it on 2026-08-01,
and no check in this pipeline could see either one.** This fixes the surviving instance and adds a check
that ends the class.

## The two defects

**A stray blank line between two rows.** Markdown terminates a table at a blank line, so every row after
one renders as a *separate* table. Master carried two simultaneously — between rows 13/14 and 35/36 —
rendering 36 rows as **three** tables, with the newest entry orphaned in a one-row table of its own.
Nothing is missing from the file, so the diff looks perfect; only a reader of the rendered page notices.
Found by the independent reviewer on #1694 while checking that PR's merge, confirmed here independently.
#1694 has since repaired the 35/36 break; this removes the surviving 13/14 one.

**Two branches independently claiming the same row number.** These merge **cleanly** — separate lines,
no textual conflict — producing a table reading `33, 34, 35, 32` at `mergeable=CLEAN` with all six jobs
green. Conflict detection is textual and line-local; it has no notion that a column is a unique key.
Other documents cite these rows **by number**, so a duplicate silently redirects every such citation to
an unrelated entry. This happened on #1696 and was caught only by diffing the merged region by eye.

## Two classes of check, deliberately separated

| class | scope | rationale |
|---|---|---|
| **structure** (always on) | no blank line may split a table | a fact about Markdown; true of every table in the repo |
| **sequence** (`--sequential`) | numbered column unique, ascending, gapless | a *convention* of the trap table ("append, never renumber"), not a property of numbered tables |

Measured with the shipped checker: **all 77 tracked Markdown files pass structure**, while of the **5**
documents holding an all-numeric-first-column table only **2** satisfy sequence. The rest lead with frame
indices, draw ordinals and submit numbers, where gaps and repeats are the correct content. Applying
sequence broadly would report correct documents as broken — and a check that fires on correct data gets
deleted rather than heeded.

Neither class inspects content, so neither can rot as the table grows.

## How a break is told from two legitimate tables

A proper Markdown table opens with a header row followed by a delimiter (`|---|---|`). A fragment left
behind by a stray blank line has **no delimiter** — it is a continuation, not a table. So an all-pipe run
without a delimiter, directly following a proper table and separated only by blank lines, is a break;
[...]

PR #3413, opened 2026-09-06
docs: trap 271 — the trap-41 deletion check false-positives on a moved base

mattias800:
Found by the independent reviewer of #3412, while checking that PR's own deletion count — which is the only reason it surfaced at all.

## The defect

The charter prescribes a check for instrument trap 41: `git diff <base> HEAD --numstat`, expect zero deletions. It is the guard against silently reverting another lane's edits, and every lane here is told to run it.

Written with **two** dots it diffs the two *tips*, so everything the base gained since you branched shows on your side as removals. Reproducible on merged history, right now:

```
$ git diff 3b7a309e2..779cc6dfc --numstat
0    10   BLOG.md
0     1   prosper/docs/GAME_COMPAT_ORCHESTRATION.md

$ git diff 3b7a309e2...779cc6dfc --numstat      # three dots — against the merge base
(nothing)
```

**Eleven deletions on a commit that deleted nothing.**

## Why it earns a row and a charter fix rather than a shrug

The false positive lands on the one check whose entire job is to catch a silent revert of another lane's work, **and it names the file that lane just edited**. So the natural reading is precisely the disaster trap 41 describes — *"my branch ate their rows"* — and the correct response to that reading is to go hunting for a revert that does not exist. The charter's wording (*"diff it against `origin/main` and read the `-` lines"*) does not say which form, and every branch here sits on a base that moves under it.

Note the asymmetry that let it survive this long: two dots can only ever **add** false deletions, never hide real ones. It fails safe, so nobody who trusts it is ever punished for it — they just re-derive a phantom, quietly, one at a time. That is also why it would never arrive as a bug report.

## What lands

- **Instrument trap 271** in `GAME_COMPAT_ORCHESTRATION.md`, with the reproduction above.
- **The charter bullet corrected** to specify three dots and say why, with `git diff $(git merge-base origin/main HEAD) HEAD` given as the explicit alternative spelling.

The two `-` lines in `CLAUDE.md`'s numstat are the rewritten bullet itself, not a revert — the diff is in the PR.

## Verification

```
python3 prosper/tools/docs/check_numbered_table.py --ordered --table-header Instrument --baseline <origin/main> <file>
  -> 14 table(s), 333 rows, unbroken; numbered column 1..271, unique and ascending,
     none of the 270 baseline rows deleted
```

Full `.md` gate clean; `diff --check` clean; diff is `.md`-only. Trap number allocated with `trap_number.py` (scans `main` and every open PR). `check_merge_result.py` will be re-run immediately before merge.

## Review note

This edits `CLAUDE.md`, which every agent loads at session start and cannot re-read mid-session, so a wrong instruction here propagates further than anywhere else in the repository. Reviewing the *wording* is the point, not only the reproduction.

Refs #3412


mattias800 (commented):
## Review of #3413 at `7ef06360` — **APPROVED**

I reproduced the trap, re-derived the numbers, and — because you asked me to attack the wording rather
than the reproduction — worked question 1 all the way through rather than answering it from instinct.
**My instinct was that three dots was wrong here. It is right, and the argument that settles it is the
one thing missing from the bullet.** That is the whole of my feedback; nothing below blocks.

---

## Question 1: yes, `origin/main...HEAD` is right for this bullet — and here is why, because the bullet does not say

I spent most of this review trying to reject on exactly your suspicion. The case against three dots
looks strong: the bullet is about not clobbering another lane's file, and three dots is blind to
anything `main` gained *after* your branch point — which is precisely the other lane's work. Two dots
sees it. So the file-scoped two-dot form the bullet used to prescribe appears to be the stronger check,
with its "false positives" being the true findings.

**That argument is wrong, and it fails on the merge, not on the diff.** A squash merge applies
`diff(merge-base → your branch)` onto `main`. Take a line `L` that exists on `main` today:

- **`L` exists at the merge base.** If your wholesale copy dropped it, your diff literally contains
  `-L`, and applying that to `main` removes it. **Silent loss. Three dots shows it.**
- **`L` does not exist at the merge base** (another lane added it after you branched). Your diff
  *cannot* contain `-L` — you cannot delete what was never there. Applying it to `main` either leaves
  `L` alone or **conflicts**. There is no silent-loss path.

So the set of lines that can be silently reverted is exactly the set present at the merge base, which
is exactly the set three dots reports. Two dots' extra `-` lines are lines that are not at risk. It is
not a stronger check with noise; it is the same check with noise.

This also matches #1701, the incident the bullet is built on: ten lines were actually lost, so they
must have been at the merge base, so three dots would have caught them too.

**Keep three dots primary**, and keep the `git merge-base` spelling as the gloss rather than swapping
them — `CLAUDE.md:822` already spells the `.md`-only check as `git diff --name-only origin/main...HEAD`,
so three-dot-primary is the form this file already teaches, and an inconsistency between two adjacent
instructions would cost more than the explicitness buys.

**But state the reason, in one clause.** Right now the bullet says which form to use and why the other
is noisy; it never says why three dots is *sufficient*. A careful reader hits the same objection I did,
and `CLAUDE.md` is the one file where they cannot go and check — it is loaded at session start and
never re-read. A rule without its reason in that file is a rule someone will quietly work around.

## Suggested bullet, ~30 words shorter than the one in the diff

Two substantive changes: the reason above, and restoring the file scoping. The current sentence reads
*"diff **it** with `git diff origin/main...HEAD`"* — "it" is the file you took whole, but the command
diffs the whole branch, so the antecedent and the command disagree.

> Re-apply your own hunks onto `main`'s version instead; when you do take a file whole, **diff it with
> `git diff origin/main...HEAD -- <file>` — three dots — and read the `-` lines**. Three dots diffs
> against the MERGE BASE, which is exactly the set of lines a merge can silently revert; anything
> `main` gained after you branched cannot be silently lost, it conflicts. Two dots diffs the two TIPS,
> so once `main` moves under your branch it reports everything main gained as *your* deletions — a
> phantom revert in the very file another lane just edited (instrument trap 271). The `Docs` CI gate
> does not cover this: it validates table structure and numbering, never prose. See instrument-trap 41.

## Question 2: it is long, but the material belongs — the trim is in the reason, not the semantics

120 → 199 words, +66%, in the file every agent loads and none can re-read. That is a real cost and you
are right to weigh it. But four of the added lines *are* the correction, and the alternative — "use
three dots, see trap 271" — sends the reader to a document they may not open mid-task. The version
above lands both the form and the reason and still comes in under the one in the diff, because the
`merge-base` gloss and the "same thing spelled out" sentence are doing less work than the "cannot be
silently lost, it conflicts" clause they would make room for. Your call; either is defensible.

## Question 3: no, the row is the shortest of its last twelve

Measured rather than eyeballed. Trap 271 is **1343 characters**. Its eleven predecessors run
1358–2471 (260: 1657, 264: 2217, 266: 2471, 268: 1358, 270: 1516), and across all 277 numbered rows in
the file the mean is 1541 and the median 1482. It is below both and below every neighbour. Not too long.

---

## Pushing back on the framing, since you asked

**The asymmetry claim is right in the sense that matters and slightly too absolute as written.**
"Two dots can only ever ADD false deletions, never hide real ones" has one counterexample: if your
branch deletes a line and `main` deleted the same line, two dots shows nothing while three dots shows
`-L`. That is a real deletion hidden by two dots.

It does not damage the point — a line `main` also dropped is not another lane's work being clobbered,
which is what the check exists to find — but "can only ever / never" is an absolute, and this
repository's own experience is that absolutes in trap rows get quoted. `in practice only adds false
deletions` costs three words and is unfalsifiable.

**The inference itself I think is correct and worth keeping.** A check that fails safe, produces a
loud-but-harmless wrong answer, and is run by one person at a time has no mechanism by which anyone
learns it is broken: each reader pays a few minutes, concludes "phantom, moving on", and files nothing.
That is a genuinely different survival mechanism from the silent instruments in traps 64/116/121 and
from the loud-but-laundering one in 122, and naming it is the useful part. It is labelled as inference
in your message; the row states it flatly, which I think is fine — it is an explanation, not a
measurement, and it reads as one.

## Two smaller things

**The demonstration's three-dot result is degenerate, and the row does not say so.** `779cc6dfc` is an
*ancestor* of `3b7a309e2` (`git merge-base` of the pair returns `779cc6dfc` itself), so three dots is
trivially empty. On a real branch three dots reports **your own additions** — on #3412 at `ecacfaf7` it
reported `10/0` and `1/0`, not nothing. As written, *"reports nothing at all"* could be read as the
expected output of the corrected command, and then a reader seeing their own additions wonders whether
they ran it wrong. One clause: *(nothing here only because the stand-in is an ancestor; on a real
branch three dots reports your own additions and no deletions)*.

**The convention this corrects is stated as "expect zero deletions", and that is not quite the rule.**
This PR is the counterexample: `7/2` on `CLAUDE.md`, where the two are your own rewritten bullet — a
legitimate deletion, verified by reading them. The bullet gets this right ("read the `-` lines"); the
PR-body convention everyone copies does not. Worth a half-sentence in the row, since a reader who takes
"zero" literally will either not rewrite anything or stop reporting the number.

---

## Verified

- **Reproduction confirmed exactly.** `git diff 3b7a309e2..779cc6dfc --numstat` → `0 10 BLOG.md`,
  `0 1 GAME_COMPAT_ORCHESTRATION.md`; the three-dot form returns empty. Eleven deletions on a commit
  that deleted nothing.
- **`7/2` on `CLAUDE.md` is what a reader sees**, and it is measured against the merge base
  (`3b7a309e2`, which is `origin/main`'s tip — the branch is current, so the two forms coincide here
  and the figure is not itself a trap-271 artifact). The two `-` lines are the old bullet's closing two
  lines, replaced in place. `1/0` on the orchestration doc, `.md`-only, no non-`.md` path.
- **Gates, all re-run by me at this head:** ordered + baseline gate exit 0 (`1..271`, unique and
  ascending, no unused numbers, none of the 270 baseline rows deleted); full `.md` gate exit 0 with
  zero non-clean lines; `check_merge_result.py` exit 0, clean merge passing against `origin/main` as of
  now; `git diff --check` exit 0. `trap_number.py` reports 272 next free, so 271 is correctly the
  highest. No `Claude-Session:` or attribution line on the branch commit.

## Verdict

**APPROVED** at `7ef06360`. The prescription is correct, the trap is reproducible on merged history,
the number is right, and the row is well-sized. The four suggestions above — the "cannot be silently
lost, it conflicts" clause, `-- <file>`, softening the absolute, and the one-clause note on the
degenerate demo — are improvements to an already-mergeable change; take what you want and merge
promptly, since the merge-result gate is stale the moment another lane appends at 272.

Issue #2710, opened 2026-08-18
The shared main checkout is 414 commits behind, so sessions started there load a 638-line CLAUDE.md instead of the current 798-line one


mattias800:
## Summary

The shared main checkout is **414 commits behind `origin/master`**, so every agent session started from
that directory loads a **638-line `CLAUDE.md`** instead of the current **798-line** one — roughly **160
lines of charter missing**, invisibly.

Verified read-only (the tree is contended; nothing was touched):

```
main checkout HEAD              f9f1163b
behind origin/master            414 commits
CLAUDE.md there                 638 lines   (clean, no local edits)
CLAUDE.md on origin/master      798 lines
"--sequential" in its copy      1      "--ordered"  0
```

## Why this is worth an issue rather than a quiet `git pull`

**It is invisible from inside an affected session.** The stale charter is the only charter that session
can see. There is no in-session symptom until a rule that changed happens to be exercised — and then
the agent reasons from a document that is confidently wrong rather than absent, which is the harder
failure.

It surfaced here by accident. A reviewer and I disagreed about whether the charter's docs-gate recipe
names the removed `--sequential` flag. Both of us were right about the copy in front of us: mine the
session-loaded one (stale), theirs the worktree one (current). Neither of us was looking at the thing
that actually decides it, which is one command:

```bash
git -C <main-checkout> rev-parse HEAD
```

That is the same shape as several defects found on #2708 in the same session — **a claim checked
against something downstream of the question**. Worth recording as an environment condition precisely
because two independent readers hit it and neither diagnosed it from inside.

## Concrete consequence already observed

`check_numbered_table.py --sequential` was removed on 2026-08-17 (#2089). An agent following the stale
charter verbatim runs it and gets `THE CHECK DID NOT RUN -- usage error`, exit 2. The tool fails
**loudly** and names `--ordered` as the replacement, which is good design and is why this cost minutes
rather than a bad merge — but the charter that agent is following will keep saying `--sequential`.

That is one flag. The gap is ~160 lines wide, and nothing says the rest of it fails as loudly.

## Suggested fix

A `git pull` in the main checkout by whoever owns it. `CLAUDE.md` there is clean, so the file itself
will fast-forward without conflict; I have not touched the tree and will not, per the charter's rule
about the shared checkout.

Worth considering alongside it: a session-start check that compares the loaded charter against
`origin/master:CLAUDE.md` and says so when they differ. The condition is cheap to detect and, as this
[...]

Issue #2283, opened 2026-08-07
perf: passes with a DISABLED colour target still take a full GPU-stalling readback — 16.9 GB in 3 minutes, discarded


mattias800:
## Every pass that renders without a colour-0 base reads back a fully black surface

Split out of #2276 because it may not be a performance defect at all, and a possible correctness problem should not live inside a perf issue.

Measured on Blue Prince (`PPSA25009`) Day One gameplay, Windows 11 / NVIDIA RTX 4090, `PROSPER_PASS_LOG_NODEFER=1`:

```
648 of 648 non-deferred passes   px_nonblack=0
4,616 MB                          read back across them
base=0x0  378        any real base  1
```

Sizes, showing these are real render targets rather than scratch:

```
91 x bytes=8294400    (1920x1080 RGBA)
90 x bytes=1048576
72 x bytes=262144
36 x bytes=65536
36 x bytes=4194304
```

**`bytes` is non-zero on every one, so pixels genuinely came back.** `px_nonblack=0` is a property of the *content*, not of an empty vector — the two were indistinguishable until `bytes=` was added to this line for exactly that reason.

## Why this is filed as possibly-correctness

`base` is `rs.color0_base`, read from the guest's own `CB_COLOR0_BASE` register (`src/gpu/render_state.cpp:164`) and copied into `DrawItem::color0_base` (`gpu_execute.hpp:1640`, *"render-to-texture: the target this draw writes into"*).

With `base == 0` the frontend passes **`nullptr`** for the backend colour target:

```cpp
live_gpu_targets && base ? &backend_target : nullptr
```

So such a pass gets **no persistent target**. It cannot be keyed into `g_rtt`, cannot be deferred, cannot be found again by a later sampler — and reads back black.

Two readings, and they need separating before anything is optimised:

1. **The passes genuinely render nothing.** The readback is then pure waste and the fix is to stop doing it — a performance change, worth ~9.7 GPU round trips and up to 8 MB each per submit.
2. **The pass renders somewhere the readback does not look.** Then the black surface is a *symptom*, the pass's output is being lost, and the cost is the least interesting thing about it.

[...]

PR #3421, opened 2026-09-07
Remove unused color-readback waits and detect stale session instructions

mattias800:
Depth-only live passes already omit unused color copies, but still force a queue submit and fence wait for those omitted results. The shared checkout also remained on an old feature branch, silently supplying stale session instructions.

This PR addresses the two issues in separate commits:

- **#2283:** use actual requested color results in the flush decision. Standalone calls, explicit batch ends and storage-image writeback still synchronize. Existing batch ownership retains resources, and failure callbacks revoke speculative attachment state. The caller still considers every bound MRT slot.
- **#2710:** add a read-only startup check of working instruction files, checkout ancestry and remote-ref freshness. Wire the project SessionStart hook and root agent instructions. Missing/offline/timeout states report unverified; a 12-second deadline allows failures to reach context before the hook timeout. Only `.claude/settings.json` is allowed through the contribution-shape exception.

Validation on this final source tree:

- Full Linux Release build and all 371 configured tests pass under core and synchronization validation; the deliberate-hazard probe confirms synchronization validation is armed. Only the four existing scoped Vulkan findings remain.
- Restoring the old flush behavior fails the pending-batch/combined-submit guards. Removing the DS failure callback fails the direct cache-revocation guard. Persistent stencil consumption, discarded batches, standalone and explicit completion, and storage atomic writeback are covered.
- Twelve startup controls exercise real temporary Git repositories and the actual configured hook argv, including paths with spaces, local/staged changes, divergence, detached/feature/linked worktrees, stale/missing refs, offline/unreachable remotes, unavailable Git, timeouts and line endings. Contribution-shape controls keep private agent files excluded.
- Blue Prince's 360-second native baseline expressed 27,605 no-color-base passes, all with zero color bytes. Later windows averaged 9.72–10.28 readback-driven waits per callback. The comparison on this head reduced comparable late-window submits/waits by roughly 6% (15.68–16.32 to 14.72–15.28 per callback), while host presentation remained approximately 4.38 FPS. Many old readback-labelled flushes also satisfied the explicit-end condition, so the old reason count overstated the removable waits. No material FPS improvement is established. The user's recent geometry-over-FMV regression is present in the baseline and is not claimed fixed here.

Details: `docs/UNREQUESTED_COLOR_FLUSH_2026_09.md` and `docs/SESSION_START.md`. The shared checkout will be explicitly refreshed after merge, preserving its feature branch and private untracked files. Since another worktree owns `main`, it will be detached at current `origin/main`; the peer's branch will not be moved.

Refs #2283, #2710.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,0 +1,21 @@
+{
+  "hooks": {
+    "SessionStart": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "python3",
+            "args": [
+              "${CLAUDE_PROJECT_DIR}/prosper/tools/session_start.py",
+              "--repo",
+              "${CLAUDE_PROJECT_DIR}",
+              "--hook"
+            ],
+            "timeout": 20
+          }
+        ]
+      }
+    ]
+  }
+}
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,12 @@
+# Session startup
+
+Read `CLAUDE.md`, `LOCAL.md` if present, and `CONTRIBUTING.md` before working here.
+Run `python3 prosper/tools/session_start.py` from the checkout root at the start of a session
+and after changing worktrees. This read-only check compares the actual checkout and instruction
+files with current `origin/main`; do not treat an unavailable remote as proof of freshness.
+If it reports a difference, inspect current instructions with `git show origin/main:CLAUDE.md`
+and preserve intentional local changes. Fetch stale refs in your own worktree and rerun the check.
+Do not switch or reset someone else's worktree to silence a warning.
+
+Follow the applicable nested `AGENTS.md` files for the area being changed. The startup check is
+automatic in Claude Code through `.claude/settings.json`; other agents run the command above.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -4,6 +4,13 @@
 > starting work. It is gitignored and holds per-computer facts (hardware, paths, installed tools)
 > that must not be written into this committed charter.
 
+**At session start and after changing worktrees**, run `python3 prosper/tools/session_start.py`
+from the checkout root. It compares this checkout and its actual instruction files against
+`origin/main`, including a bounded remote-ref freshness check (#2710). A difference may be an
+intentional local edit; inspect it before trusting the session's loaded charter. An unavailable
+remote is unverified, not current. The check never fetches or switches branches. Claude Code runs
+it through the checked-in SessionStart hook; other agents follow root `AGENTS.md`.
+
 ## What this project is
 
 **prosper is a PS5→PC compatibility layer — "Wine/Proton for PS5."** It runs PS5 games natively on
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -382,6 +382,8 @@ if(Python3_Interpreter_FOUND)
   # guards a state this repository produces routinely, since every lane pushes after CI starts.
   add_test(NAME pr_merge_gate
            COMMAND "${Python3_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/tools/ci/test_pr_merge_gate.py")
+  add_test(NAME session_start
+           COMMAND "${Python3_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/tools/test_session_start.py")
   # PROSPER_ENV_ON/_VALUE sample getenv once and return it forever. Applied to a name a TEST arms at
   # runtime, the arm becomes invisible -- and the test does not fail, it goes vacuous and keeps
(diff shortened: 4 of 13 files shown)


mattias800 (commented):
APPROVED — render scope (#2283) at commit bb199919da6248a543f829abcc89071fe5272d52, compared with main 241a59342441cdb85cf7ec88b1dd31babd3e4605.

No remaining blocking findings in the renderer changes. I reviewed the flush predicate and attribution, retained batch resource lifetime, speculative depth/stencil publication and failure invalidation, and the standalone, explicit-flush and storage-writeback paths. Deferring work now depends on whether a CPU color result is actually requested; the other completion reasons remain intact.

The initial discard test did not enable persistent depth/stencil for its consumer and therefore could not prove cache invalidation. That finding is resolved on this head: the test observes speculative cache state, directly verifies revocation, then invokes a persistent consumer. I inspected the mutation result showing removal of the failure cleanup fails the revocation assertion. The successful batching arm checks both preserved stencil output and one submit/wait for the two command buffers.

I also inspected the retained full validation log: all 371 tests passed with core and synchronization validation, the deliberate hazard confirmed synchronization validation was armed, and only four existing allowlisted IDs appeared. No builds, tests or GPU runs were repeated for this review.

This approval covers the renderer implementation and its regression tests. The session-start tooling and instruction changes for #2710 receive separate review. The existing DRAW_ISO empty-output issue predates this patch and was reported to the author for separate tracking. CI and exact-head checks remain the merge gate.

mattias800 (commented):
APPROVED: bb199919da6248a543f829abcc89071fe5272d52

Independent review scope: #2710 startup freshness tooling, its 12 controls, the SessionStart hook, root instruction wiring, CTest registration, contribution-shape exception, and SESSION_START documentation. Reviewed the final committed diff against main 241a59342441cdb85cf7ec88b1dd31babd3e4605. No blocking findings in this scope.

The check compares actual working instruction files, distinguishes local edits from reference differences, verifies remote-tracking freshness, and preserves an UNVERIFIED result for missing references, unavailable Git/remotes, offline operation and timeouts. The 12-second shared deadline leaves room for hook context before its 20-second timeout. The tests exercise the configured hook argv, including paths with spaces, and include both successful and failing controls. The metadata exception permits only .claude/settings.json.

Inspected author verification evidence: all 371 CTests passed and the Vulkan synchronization scan reported no new validation IDs. I did not rerun tests. Renderer changes for #2283 are outside this review scope. Refreshing the shared checkout remains a separate post-merge action; the documented detached-main procedure preserves the existing feature branch, untracked files and the peer worktree owning main.

mattias800:
AUTHOR VERIFICATION — head `bb199919da6248a543f829abcc89071fe5272d52`

Full Linux Release build and all 371 configured tests pass under core and synchronization validation. The deliberate-hazard probe confirms synchronization validation is armed. The scanner reports only the four existing test-scoped IDs associated with #1710/#1715/#1716, with no new findings. The full log includes all 12 startup-check controls.

The old-flush mutation fails the pending-batch/combined-submit tests. Removing the DS failure callback fails the direct cache-revocation test. The corrected consumer uses the persistent attachment, and the test checks speculative state before and after discard. All mutations were restored before the full run. Standalone, explicit batch-end and storage-writeback completion remain covered.

Native Blue Prince comparison, separate fresh saves, same direct-graphics/IME-autokey route, immediate presentation, F8 at 300 seconds and screenshot/bundle at 310 seconds:

| Measurement | Baseline | This head |
| --- | ---: | ---: |
| Sampled interval | 10.043 s | 10.043 s |
| Guest flips/s | 4.48 | 4.38 |
| Host presents/s | 4.38 | 4.38 |
| Process CPU cores | 1.82 | 1.81 |
| Pre/post samples | 20/21 | 20/21 |
| Renderer/compute records | 46/708 | 44/687 |
| Dropped records | 0/0 | 0/0 |

Comparable late timing windows change from 15.68–16.32 submits/waits per renderer callback to 14.72–15.28 (roughly 6% fewer). Readback-driven flushes become zero. The remaining flushes are explicit batch ends: many formerly labelled readback also satisfied that condition, so the exclusive reason histogram was not a count of removable waits. No material FPS improvement is established by this single pair.

This is the opening cutscene, not Day One gameplay. Both inspected screenshots retain the user's recent world-geometry-over-FMV regression; it is present in the baseline and not claimed fixed. The operator also triggered an earlier F9 in the new run, outside the timed window. No local compiler/other GPU workload overlapped either performance capture. Baseline binary provenance, including its pre-commit embedded revision, is stated in `docs/UNREQUESTED_COLOR_FLUSH_2026_09.md` and retained with a binary hash.

Both scoped independent reviews approve this exact head. `git diff --check` passes, and current main is included. Required CI is checked separately before merge. Shared-checkout refresh and preservation checks will be recorded on #2710 after merge.

Issue #3378, opened 2026-09-05
Graphics pipelines are created with no VkPipelineCache — every draw state recompiles, every run


mattias800:
## Summary

**The graphics backend creates every pipeline with no `VkPipelineCache` at all.** `grep -c VkPipelineCache prosper/tests/fixtures/render_runner.h` returns **0**, and the one creation site passes `VK_NULL_HANDLE`:

```cpp
// prosper/tests/fixtures/render_runner.h:8409
const VkResult pipeline_result = vkCreateGraphicsPipelines(
        dev, VK_NULL_HANDLE, 1, &gp, nullptr, &v.pipe);
//           ^^^^^^^^^^^^^^^ the VkPipelineCache parameter
```

So the driver is handed no cache — not persistent, and not even in-process. Every new draw state pays a full shader compile on the submit path, and every launch pays the whole set again from zero.

**Compute already does this properly.** `frontends/shared/live/live_compute.cpp:1440` (`persistent_pipeline_cache_path`) loads a blob at startup, keys it by device identity, and writes it back with `vkGetPipelineCacheData` on clean exit. The machinery exists and is reviewed; the graphics path was never wired to it.

Note `pipeline_cache_enabled` / `pipeline_cache_limit` at `render_runner.h:6457-6458` are a **different** mechanism — prosper retaining its own `VkPipeline` handles so a repeated draw state is not rebuilt. That helps within a run and does nothing across runs, and it does not give the driver a cache to reuse work between *similar* pipelines.

## Failure scenario

Launch any title interactively. Each time a draw state is seen for the first time, `vkCreateGraphicsPipelines` compiles it synchronously in the submit path, producing a visible hitch. Quitting and relaunching the same title reproduces every hitch identically, because nothing was retained.

Observed on `c4f57b1de`, *New Joe & Mac* (`PPSA02801`), `prosper-app`, RADV STRIX_HALO, native 1080p, with `PROSPER_FLIP_PACE_FPS=60`. The project owner's description: *"quite a lot of stuttering, kind of like shader compilation stutter that is typical for emulators."*

## What is NOT established

**That pipeline compilation is the cause of that stutter.** The finding above is a real gap and is consistent with the symptom, but it has not been measured against the alternatives — texture de-tiling and the resource layer produce a similar hitch, and #2215 is this project's recorded case of exactly that assumption sending two lanes down five dead hypotheses.

[...]

PR #3424, opened 2026-09-07
Persist graphics pipeline caches across app launches

mattias800:
Graphics pipelines already share an in-process driver cache, but the app lost that cache at `_Exit`. This adds opt-in cross-launch persistence with device/driver/UUID and pointer-width checks, bounded file parsing, integrity checks, an empty-cache retry after driver rejection, and atomic replacement that preserves the previous file on failure.

The app snapshots the initialized cache before `_Exit`. Extraction shares a dedicated mutex with pipeline compilation and waits up to one second to acquire it. It does not depend on the whole-pass renderer lock: an actual Sonic run exposed that the original broad-lock try-lock skipped saving. Vulkan extraction and file I/O themselves are not time-bounded.

Disk loading remains opt-in, including on NVIDIA; compatibility checks do not resolve the documented NVIDIA blob-loading crash. No experimental Vulkan feature is needed. See `prosper/docs/GRAPHICS_PIPELINE_CACHE.md` for controls and limits. The separate compute `_Exit` persistence gap is tracked in #3425.

Validation at `b4c2cd8627e8cad18b135e0e2c81012ac5aa4479`:

- Full Release build passed. Focused file/Vulkan tests passed 2/2; restoring the old broad lock makes the save regression fail with exit 5.
- Tests cover separate-process save/load through `_Exit`, observed Vulkan initial data, disabled/corrupt/truncated inputs, driver rejection and empty retry, concurrent writers, failed replacement, compilation serialization and saving during a held render pass.
- Two actual 45-second Sonic app launches: cold launch saved 140,180 driver bytes; warm launch loaded 140,180 and saved again. Each reached 574 presentations. This proves persistence, not a speedup.
- Independent exact-head review approved. All 373 tests passed (99.44s); full core/synchronization scan passed all 373 tests (97.06s), with only the existing scoped findings and the scanner's deliberate-hazard control armed. Required Linux/general CI remains the merge gate.

Addresses the remaining graphics persistence scope of #3378. No claim that pipeline compilation caused the original stutter or that this raises FPS.

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -3172,6 +3172,24 @@ if(TARGET prosper_core)
       target_link_libraries(test_pipeline_render PRIVATE prosper_core Vulkan::Vulkan)
       add_test(NAME pipeline_render COMMAND test_pipeline_render)
 
+      add_executable(test_pipeline_cache_file frontends/shared/tests/test_pipeline_cache_file.cpp)
+      target_include_directories(test_pipeline_cache_file PRIVATE frontends)
+      target_link_libraries(test_pipeline_cache_file PRIVATE Vulkan::Vulkan Threads::Threads)
+      if(Python3_Interpreter_FOUND)
+        add_test(NAME pipeline_cache_file COMMAND ${Python3_EXECUTABLE}
+          ${CMAKE_CURRENT_SOURCE_DIR}/frontends/shared/tests/run_pipeline_cache_file.py
+          $<TARGET_FILE:test_pipeline_cache_file>)
+      endif()
+
+      add_executable(test_graphics_cache_persistence tests/gpu/test_graphics_cache_persistence.cpp)
+      target_include_directories(test_graphics_cache_persistence PRIVATE tests frontends)
+      target_link_libraries(test_graphics_cache_persistence PRIVATE prosper_core Vulkan::Vulkan)
+      if(Python3_Interpreter_FOUND)
+        add_test(NAME graphics_cache_persistence COMMAND ${Python3_EXECUTABLE}
+          ${CMAKE_CURRENT_SOURCE_DIR}/tests/gpu/test_graphics_cache_persistence.py
+          $<TARGET_FILE:test_graphics_cache_persistence>)
+      endif()
+
       # Full back-half spine: a GpuState -> recompile shaders from their PGM addresses -> resolve
       # pipeline state -> rendered frame (the path a real submitted Dcb will drive).
       add_executable(test_gpustate_render tests/gpu/test_gpustate_render.cpp)
--- a/prosper/docs/GRAPHICS_PIPELINE_CACHE.md
+++ b/prosper/docs/GRAPHICS_PIPELINE_CACHE.md
@@ -1,0 +1,45 @@
+# Graphics pipeline cache persistence
+
+The renderer shares one device-lifetime `VkPipelineCache` across graphics pipeline creation.
+Disk persistence is opt-in, as for the compute cache:
+
+- `PROSPER_DISK_PIPELINE_CACHE=1` selects the platform cache directory.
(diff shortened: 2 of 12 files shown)


mattias800 (commented):
APPROVED: 66c920c2fdff47fb92c0b10fa3e0cef1b64d42fb

Independent code review of the complete graphics persistence diff against 883dbc8b559f5de8dd4eda2ba1e45995d36249d4, including lifecycle wiring, file format/storage, controls, CMake registration and GRAPHICS_PIPELINE_CACHE.md. No blocking code findings.

The final revision addresses the earlier optional-Python registration finding and adds an injected cached-creation rejection control that requires two attempts and correct rendered pixels. Cache extraction shares graphics creation synchronization, busy shutdown skips saving, initialized context publication is atomic, and disk I/O occurs outside the renderer lock. File replacement preserves the previous destination on failure; only the validated original driver payload reaches Vulkan. Opt-in loading and the unresolved NVIDIA failure are documented without claiming a performance improvement. Compute persistence is unchanged.

This is code-review approval, not completed author verification: the full suite/validation run and an actual prosper-app save/load pair remain required before merge. The latter matters because the production shutdown try-lock may skip a save even though the isolated two-process control succeeds. I did not build, rerun tests or run GPU work during review.

mattias800 (commented):
APPROVED: b4c2cd8627e8cad18b135e0e2c81012ac5aa4479

Independent follow-up review of the committed correction since 66c920c2fdff47fb92c0b10fa3e0cef1b64d42fb; the earlier full-diff review remains applicable to unchanged code. No blocking code findings.

The dedicated cache mutex now covers the sole graphics pipeline-creation call and both cache-data queries. Initialization precedes atomic context publication, snapshotting takes no whole-pass resource lock, and file I/O remains outside the cache mutex. The resource-to-cache lock order has no reverse acquisition in the snapshot path. This addresses the observed app shutdown that skipped persistence merely because a render pass held the broader resource domain.

Controls now require successful saving while the broad pass mutex is owned, refusal while the compilation mutex remains owned, and the dedicated lock at the intercepted real pipeline-creation call. The interface comment correctly limits the one-second bound to lock acquisition; Vulkan extraction and disk I/O are not time-bounded.

Author-verification gates remain separate and pending: full exact-head build/tests and Vulkan validation, green CI, and an actual app cold/save then warm/load pair. This review does not claim those have passed or authorize bypassing them. I did not build, rerun tests or run GPU work.

mattias800:
Author verification for `b4c2cd8627e8cad18b135e0e2c81012ac5aa4479`:

- Full Release build: `PROSPER_GAME_ROOT=<DUMP_ROOT> cmake --build prosper/build-linux -j6` passed.
- `ctest --test-dir prosper/build-linux --no-tests=error --output-on-failure`: 373/373 passed, 99.44s.
- `python3 prosper/tools/vkval/vk_validation_scan.py --build-dir prosper/build-linux --sync`: 373/373 passed, 97.06s; scanner's synchronization hazard control armed. Four existing scoped message IDs, 71 messages, no unaccounted findings. Raw logs retained privately under `prosper/build-linux/goal-cache-final-*`.
- Regression negative arm: restoring the old broad resource try-lock makes separate-process save fail with exit 5; corrected code passes. The API interception also verifies pipeline creation holds the dedicated cache mutex.
- Actual `prosper-app` native/immediate Sonic launches, 45s each, fresh private saves: cold saved 140,180 driver bytes; next launch loaded 140,180 and saved again at shutdown. Each reached 574 presentations. Manifest includes exact command, head and binary SHA256 under `prosper/build-linux/evidence/goal-cache-app-proof/`. No speedup is inferred.
- Independent registered review approves this exact head: https://github.com/mattias800/prosper/pull/3424#pullrequestreview-5129164926
- Fetched main remains 883dbc8b559f; branch includes it. Required Linux/general CI is still pending; not merged yet.

Issue #3450, opened 2026-09-07
GTA V performance plan: 94% of per-dispatch GPU time is prosper's own storage copy, not the guest shader — and on a discrete GPU it crosses PCIe every dispatch


mattias800:
> **UPDATED after the A/B below — read the comments before acting on the task list.**
> `PROSPER_SKIP_EXPORTED_STORAGE_WRITEBACK=1` cuts compute by 51% and raises graphics by 40%, leaving
> total measured work and frame rate **unchanged**: the copy is prepaid work for the next graphics
> pass, not waste. **T2 as originally written is falsified**; the surviving route is T5. A 2.9x cut in
> compute GPU work moved the frame rate not at all, so compute GPU time is not the frame gate — find
> the critical path first (T9).

GTA V renders its world at **~1.3 fps** on Windows/NVIDIA on current `main`. This issue records where that time actually goes, measured with prosper's own instruments, and turns it into a task list. The short version:

> **94% of the GPU work prosper does per compute dispatch is prosper's own copying and comparison, not the guest's shader** — and on a discrete GPU that copy crosses PCIe every dispatch, which is why the same code is disproportionately slow on Windows/NVIDIA compared with the Linux/AMD APU the lanes develop on.

## Measurement conditions

| | |
| --- | --- |
| host | Windows 11, 24 logical cores, **NVIDIA RTX 4090** (discrete) |
| build | `main` `d9d03f8a`, MinGW RelWithDebInfo, `prosper-app` |
| title | GTA V `PPSA04263`, prologue bank interior, **world rendering confirmed by screenshot** |
| route | `scripts/gta5/reach-performance-story-windows.pad` (see #3449 — the committed route selects the wrong setting on this host) |
| present | `mailbox`, native 4K, tuned cache env from `GTA5_STATUS.md` |
| device losses | **0** in every run, so this is not #3279's recovery case |

Utilisation in the world phase: **2.5–3.0 of 24 CPU cores** and **26.3% GPU 3D engine**. Neither resource is saturated, which is the signature of time going somewhere other than useful work.

## 1. Where the GPU time goes (the headline)

Each dispatch records six GPU timestamps bracketing `pre -> shader -> storage_copy -> compare -> restore` (`live_compute.cpp:10339-10362`). Over a **929-dispatch** window:

| GPU bracket | ms | share of `gpu_device` |
| --- | --- | --- |
| seed / pre-dispatch copy | 19.5 | 1.7% |
[...]

PR #2245, opened 2026-08-07
diag(render): report buffer-reuse economics — the leaf that dominates the FMV collapse explained nothing

mattias800:
Refs #2215, #2231. Takes the narrow piece Wren released on #2241.

## The gap

`buffer copy` is **539 ms of 802 ms** of backend resource setup in Blue Prince's FMV collapse — 67%, and 15% of the whole frame. **Nothing could say why.** The counters exist and are aggregated on every backend call; the only consumers in the tree were assertions in `test_multidraw_render.cpp`.

The asymmetry is stark: the sibling resource kind has reported its cache economics for ages (`texture_cache hits=… misses=…` at 2 ms/submit), while buffers at 39 ms/submit reported nothing.

## What it says

Blue Prince, cumulative over 16,384 backend calls:

```
refs=7,267,454  unique=1,428,929  memo_hits=5,838,525  hashed=754,403 (307.8 MiB)
skipped_large=673,079  skipped_unique=1,447  upload_fallbacks=0
```

**It self-checks exactly**: `5,838,525 + 754,403 + 673,079 + 1,447 = 7,267,454` — every reference on exactly one path, which is the arm to watch if these counters ever drift.

That resolves the three-way question, and it is the third answer:

| | | |
|---|---:|---|
| memo hits | **80.3%** | the per-call memo is doing its job |
| `skipped_unique` | **0.02%** | `buffer_identity` is fine — *not* an upstream identity problem |
| `skipped_large` | **9.3%** | large buffers, content dedup skipped **by design** |

1.43M unique uploads, ~87 per backend call. **The copies are not a dedup failure.** They are large, genuinely distinct payloads that the 4 KiB hash-dedup cap deliberately excludes — and that cap is correct, on the measurement that put it there (zero dedup hits across 556K hashed references).

**So the fix is not better dedup. It is not staging a large payload through a mapped `memcpy` at all.** That is a different piece of work and this PR does not attempt it; it establishes which of three fixes is the right one.

## Two instrument errors made and caught while building this

Both left comments behind, because both are reproducible mistakes rather than typos.

1. **Read the thread-local inside `record_backend_timing`** rather than capturing it beside the pipeline stats at the call site. Fixed by passing it in, exactly as `pipelines` already was.

[...]

PR #3451, opened 2026-09-08
Print every compute timer the capture carries, with the remainder named

### LINKED ISSUE #3447: performance_capture_report.py hides the dominant compute bucket: pipeline_ms, cleanup_ms and the GPU brackets are captured but never printed
The `.prperf` file records eleven per-dispatch compute timers. `tools/perf/performance_capture_report.py` prints **five**, and the ones it drops are repeatedly the largest.

## Measured

Windows/NVIDIA RTX 4090, `main` `d9d03f8a`, GTA V (`PPSA04263`), route `scripts/gta5/reach-performance-story*.pad`.

A world-phase capture, as the tool prints it:

```
compute identities: groups=73 unknown=0
  0x2b1c4d31... records=1 dispatches=1 total=277.2 ms (setup=6.3ms wb=17.8ms wait=16.1ms dev=15.7ms shader=0.4ms)
```

`6.3 + 17.8 + 16.1 = 40.2 ms` of a `277.2 ms` dispatch. Summed over the top ten groups, **75% of the time is unaccounted in the printed output.** Parsing the same file directly:

| timer | ms | share | printed by the tool? |
| --- | --- | --- | --- |
| `pipeline_ms` | 821.5 | **66.3%** | **no** |
| `dispatch_wait_ms` | 205.5 | 16.6% | yes |
| `writeback_ms` | 104.8 | 8.5% | yes |
| `setup_ms` | 100.2 | 8.1% | yes |
| `cleanup_ms` | 3.3 | 0.3% | **no** |

Nothing is missing from the capture — `performance_capture.cpp:464` and `:467` write `pipeline_ms` and `cleanup_ms`, and `performance_capture.hpp:165/168` declare them. Only the report drops them.

## The GPU brackets are dropped too, and they carry the real answer

Each dispatch records six GPU timestamps (`live_compute.cpp:10339-10362`) bracketing `pre -> shader -> storage_copy -> compare -> restore`. The report prints `dev` and `shader` and drops `gpu_pre_ms`, `gpu_storage_copy_ms`, `gpu_compare_ms`, `gpu_restore_ms`. Over a 929-dispatch window:

| GPU bracket | ms | share of `gpu_device` |
| --- | --- | --- |
| seed / pre-dispatch copy | 19.5 | 1.7% |
| **guest compute shader** | **67.2** | **5.7%** |
| **post-shader storage copy** | **1043.9** | **88.5%** |
| result compare | 48.8 | 4.1% |

**94% of the GPU work prosper does per compute dispatch is prosper own copying and comparison, not the guest shader** — and none of that is visible in the report the project uses to decide what to optimize.

## Why this matters more than a formatting gap

`CLAUDE.md` instructs agents to read the F8 report before optimizing anything. That report has been steering by roughly a quarter of the signal on the compute side, and by a figure that excludes the single largest bracket on the GPU side. It also prints `primary evidence: compute is the largest measured component (74%)` and then decomposes compute into buckets summing to 25% of itself, without saying so.

## Suggested fix

- Print every timer the record carries, and print an explicit `unattributed` row when the children do not sum to the parent — `compute_phase_report.py` already does exactly this for `[compute-phase]` lines and its header explains why ("this is where an unmeasured cost hides, so it is printed rather than silently dropped"). The `.prperf` report should adopt the same rule.
- Fail loudly rather than silently when a parent bucket exceeds the sum of its children by more than a small margin.
- Consider surfacing `compute groups = 0` as a warning: that condition is what made the GTA V benchmark in #3446 a loading-screen measurement without anyone noticing.

Found while investigating GTA V performance on Windows; the substantive performance findings will be filed separately.

mattias800:
Fixes #3447. Refs #3450 (task T1).

## Failure scenario

`performance_capture_report.py` aggregates **eleven** per-dispatch compute timers and prints **five**. The dropped ones are repeatedly the largest, so the report `CLAUDE.md` tells agents to read before optimising anything has been steering by a minority of its own signal.

Concretely, as the tool printed it before this change:

```
0x2b1c4d31... records=1 dispatches=1 total=277.2 ms (setup=6.3ms wb=17.8ms wait=16.1ms dev=15.7ms shader=0.4ms)
```

`6.3 + 17.8 + 16.1 = 40.2 ms` of a `277.2 ms` dispatch. Over the printed groups, **75% of the time was unaccounted in the output** while sitting in the file all along.

Measured on GTA V (`PPSA04263`), Windows/NVIDIA: the dropped CPU timer `pipeline_ms` held **66%** of one window, and the dropped GPU bracket `gpu_storage_copy_ms` held **88%** of device time in two others.

## What it prints now

```
compute decomposition (CPU phases partition total; GPU brackets partition device time):
  CPU phases of 1472.0ms total: setup=207.6ms (14.1%) pipeline=8.9ms (0.6%) wait=927.5ms (63.0%)
    wb=312.4ms (21.2%) cleanup=3.8ms (0.3%) unattributed=11.8ms (0.8%)
  GPU brackets of 821.8ms device: pre=15.6ms (1.9%) shader=51.6ms (6.3%) storage-copy=723.1ms (88.0%)
    compare=31.6ms (3.8%)
    the guest's own compute shader is 6.3% of the device time prosper spends running it
```

## Invariants

- **The two decompositions are siblings and are never summed.** CPU phases partition `total_ms`; GPU brackets partition `gpu_device_ms`, which is itself *inside* `dispatch_wait_ms`. Scaling the brackets against `total_ms` would understate the storage copy several-fold and invite adding two overlapping decompositions into one bogus 100%. A test pins this, including a negative assertion on the wrong denominator.
- **Any remainder is named, not left to subtraction** — the rule `compute_phase_report.py` already states in its own header for `[compute-phase]` lines: *"this is where an unmeasured cost hides, so it is printed rather than silently dropped."* Applied both to the whole-capture table and to each program group.
- **The header announces only the tables a capture can supply.** A run without GPU timestamps prints no bracket table and no longer advertises one, so absence of a capability does not read as a missing section.

## Not a format change

Every field printed was already written by `performance_capture.cpp` (`:464`, `:467`) and declared in `performance_capture.hpp` (`:165`, `:168`). **Existing `.prperf` files gain the new output with no re-capture** — which is how the GTA V figures above were obtained, from captures taken days earlier.

## Tests

Six arms added in `ComputeDecompositionTests`, on a record whose cost sits entirely in previously-dropped timers (`pipeline_ms=220`, `gpu_storage_copy_ms=55` of a 300 ms dispatch):

| arm | what it pins |
| --- | --- |
| `test_every_captured_timer_is_reported` | all eleven reach the summary and the text |
| `test_dominant_timer_reaches_the_printed_output` | the largest cost is visible without parsing the file |
| `test_remainder_is_named_rather_than_left_to_subtraction` | the 4 ms remainder is stated |
| `test_gpu_brackets_are_scaled_against_device_time_not_total` | the sibling relationship |
| `test_shader_share_of_device_time_is_stated` | the headline line |
| `test_absent_gpu_timestamps_do_not_invent_a_bracket_table` | no table, and no header promising one |

**Verified they fail without the fix — on assertions, not on import.** The new constants are imported *inside* the two tests that need them, precisely so the other four fail on behaviour against the pre-change tool rather than on a missing symbol:

```
$ git show origin/main:...performance_capture_report.py > ...
$ python test_performance_capture_report.py ComputeDecompositionTests
FAIL: test_dominant_timer_reaches_the_printed_output          not found: (73.3%)
FAIL: test_gpu_brackets_are_scaled_against_device_time_...    not found: GPU brackets of 65.0ms device
FAIL: test_remainder_is_named_rather_than_left_to_subtr...    not found: unattributed=4.0ms
FAIL: test_shader_share_of_device_time_is_stated              not found: compute shader is 6.2% of the device time
Ran 6 tests -- FAILED (failures=4, errors=2)
```

The pre-fix output quoted in those failures is itself the evidence: `(setup=2.0ms wb=3.0ms wait=70.0ms dev=65.0ms shader=4.0ms)` for a 300 ms dispatch whose `pipeline_ms=220` is nowhere.

## Results

- `python prosper/tools/perf/test_performance_capture_report.py`: **32 tests, OK** (26 pre-existing, unchanged).
- Registered in ctest at `CMakeLists.txt:1232`, so CI covers it.
- Run against **8 real `.prperf` captures** spanning 3 revisions and 6 days: all exit 0. `--json` still parses and carries `compute_cpu_phases`, `compute_gpu_brackets`, `compute_gpu_device_ms`.
- `git diff --check` clean; CRLF preserved (0 bare LF in both files).

## What it immediately surfaced

Applied to captures already on disk, the storage-copy share is 42-88% on recent revisions — and on two captures from `0d6e3b2d` (2026-09-02) it is `gpu_compare_ms` at **78.7%** and **82.0%** instead, the cost `PROSPER_MAX_GPU_COMPARE_IMAGE_MB` was later added to bound. Across all eight captures, **the guest's own compute shader is 3.0%-24.9% of device time.** None of that was visible before.

## Risk

Low, and confined to one offline reporting tool. No emulator code, no capture-format change, nothing that runs during a game. The worst case is a wider report; the JSON gains keys and loses none.

### DIFF
--- a/prosper/tools/perf/performance_capture_report.py
+++ b/prosper/tools/perf/performance_capture_report.py
@@ -694,6 +694,11 @@ def print_summary(summary):
             cpu_parts.append(f"unattributed={unattributed:.1f}ms")
         gpu_parts = []
         for field in COMPUTE_GPU_BRACKETS:
+            # One-sided deliberately, unlike the CPU loop above. These are masked differences of
+            # GPU tick counters, so an out-of-order pair does not go negative -- it WRAPS to an
+            # enormous positive, which this threshold cannot hide and abs() would not help with.
+            # The summary table's shared abs() makes such a value loud instead: an injected
+            # storage-copy of 500ms against a 65ms device prints 769.2% with a -684.6% remainder.
             if group.get(field, 0.0) >= 0.05:
                 gpu_parts.append(f"{PHASE_LABELS[field]}={group[field]:.1f}ms")
         sections = []
--- a/prosper/tools/perf/performance_capture_report.py
+++ b/prosper/tools/perf/performance_capture_report.py
@@ -224,6 +224,52 @@ def _resource_breakdown(renderer):
     return breakdown
 
 
+# The eleven per-dispatch compute timers the capture records, split by what they bracket.
+# `dispatch_wait_ms` is CPU wall time around the submit; the GPU brackets below decompose the
+# device work inside it, so the two groups are siblings and must not be summed together.
+#
+# Every one of these is written by performance_capture.cpp. This tool used to print five of them,
+# and the ones it dropped were repeatedly the largest -- `pipeline_ms` was 66% of one window and
+# `gpu_storage_copy_ms` was 88% of device time in two others, neither visible in the output
+# (#3447). A report that silently drops a bucket steers optimisation by the part it happens to
+# show, so everything the record carries is printed, and any remainder is named rather than left
+# to be inferred by subtraction.
+COMPUTE_CPU_PHASES = (
+    "setup_ms",
+    "pipeline_ms",
+    "dispatch_wait_ms",
+    "writeback_ms",
+    "cleanup_ms",
+)
+
+COMPUTE_GPU_BRACKETS = (
+    "gpu_pre_ms",
+    "gpu_shader_ms",
+    "gpu_storage_copy_ms",
+    "gpu_compare_ms",
+    "gpu_restore_ms",
+)
+
+# Short labels for the one-line breakdowns; the long names stay in the JSON.
+PHASE_LABELS = {
+    "setup_ms": "setup",
+    "pipeline_ms": "pipeline",
+    "dispatch_wait_ms": "wait",
+    "writeback_ms": "wb",
+    "cleanup_ms": "cleanup",
+    "gpu_pre_ms": "pre",
+    "gpu_shader_ms": "shader",
+    "gpu_storage_copy_ms": "storage-copy",
+    "gpu_compare_ms": "compare",
+    "gpu_restore_ms": "restore",
+}
+
+
+def _phase_totals(records, fields):
+    """Sum `fields` over `records`, returning a plain dict so callers can add derived rows."""
+    return {field: sum(float(record.get(field, 0.0)) for record in records) for field in fields}
+
+
 def _hex64(value):
     return f"0x{value:016x}"
 
@@ -245,19 +291,20 @@ def _compute_program_groups(records, limit=10, address_limit=8):
             continue
         group = groups.setdefault(program_hash, {
             "records": 0, "dispatches": 0, "total_ms": 0.0, "max_ms": 0.0,
-            "setup_ms": 0.0, "writeback_ms": 0.0, "dispatch_wait_ms": 0.0,
-            "gpu_device_ms": 0.0, "gpu_shader_ms": 0.0,
             "addresses": set(),
+            **{field: 0.0 for field in COMPUTE_CPU_PHASES},
+            **{field: 0.0 for field in COMPUTE_GPU_BRACKETS},
+            "gpu_device_ms": 0.0,
         })
         group["records"] += 1
         group["dispatches"] += dispatches
         group["total_ms"] += total_ms
         group["max_ms"] = max(group["max_ms"], total_ms)
-        group["setup_ms"] += float(record.get("setup_ms", 0.0))
-        group["writeback_ms"] += float(record.get("writeback_ms", 0.0))
-        group["dispatch_wait_ms"] += float(record.get("dispatch_wait_ms", 0.0))
+        for field in COMPUTE_CPU_PHASES:
+            group[field] += float(record.get(field, 0.0))
+        for field in COMPUTE_GPU_BRACKETS:
+            group[field] += float(record.get(field, 0.0))
         group["gpu_device_ms"] += float(record.get("gpu_device_ms", 0.0))
-        group["gpu_shader_ms"] += float(record.get("gpu_shader_ms", 0.0))
         group["addresses"].add(address)
 
     ranked = []
@@ -515,6 +562,9 @@ def summarize(records):
         "rates": rates,
         "graphics_total_ms": graphics_total,
         "compute_total_ms": compute_total,
+        "compute_cpu_phases": _phase_totals(compute, COMPUTE_CPU_PHASES),
+        "compute_gpu_brackets": _phase_totals(compute, COMPUTE_GPU_BRACKETS),
+        "compute_gpu_device_ms": _total(compute, "gpu_device_ms"),
         "compute_programs": compute_programs,
         "gpu_timestamp_samples": gpu_timestamp_samples,
         "gpu_timestamps_available": gpu_timestamps_available,
@@ -532,6 +582,58 @@ def summarize(records):
             "compute": footer["compute_records"],
         },
     }
+
+
+def _print_phase_line(label, totals, denominator, extra=None):
+    """One `name=ms (pct)` line; percentages are of `denominator`, which is named by the caller."""
+    if denominator <= 0:
+        return
+    parts = []
+    for field, value in totals.items():
+        if value < 0.05:
+            continue
+        name = PHASE_LABELS.get(field, field)
+        parts.append(f"{name}={value:.1f}ms ({100.0 * value / denominator:.1f}%)")
+    if extra:
+        parts.extend(extra)
+    if parts:
+        print(f"  {label}: {' '.join(parts)}")
+
+
+def _print_compute_phases(summary):
+    """Print every compute timer the capture carries, with the remainder named.
+
+    Two sibling decompositions, deliberately not summed: the CPU phases partition `total_ms`,
+    while the GPU brackets partition `gpu_device_ms`, which is itself inside `dispatch_wait_ms`.
+    """
+    total = summary.get("compute_total_ms", 0.0)
+    cpu = summary.get("compute_cpu_phases") or {}
+    gpu = summary.get("compute_gpu_brackets") or {}
+    device = summary.get("compute_gpu_device_ms", 0.0)
+    if total <= 0 and device <= 0:
+        return
+    # Announce only the tables actually printed: a capture with no GPU timestamps has no bracket
+    # table, and a header promising one reads as a missing section rather than an absent capability.
+    scopes = []
+    if total > 0:
+        scopes.append("CPU phases partition total")
+    if device > 0:
+        scopes.append("GPU brackets partition device time")
+    print(f"compute decomposition ({'; '.join(scopes)}):")
+    if total > 0:
+        unattributed = total - sum(cpu.values())
+        extra = ([f"unattributed={unattributed:.1f}ms ({100.0 * unattributed / total:.1f}%)"]
+                 if abs(unattributed) >= 0.05 else None)
+        _print_phase_line(f"CPU phases of {total:.1f}ms total", cpu, total, extra)
+    if device > 0:
+        unattributed = device - sum(gpu.values())
+        extra = ([f"unattributed={unattributed:.1f}ms ({100.0 * unattributed / device:.1f}%)"]
+                 if abs(unattributed) >= 0.05 else None)
+        _print_phase_line(f"GPU brackets of {device:.1f}ms device", gpu, device, extra)
+        shader = gpu.get("gpu_shader_ms", 0.0)
+        if device > 0 and shader >= 0.0:
+            print(f"    the guest's own compute shader is {100.0 * shader / device:.1f}% of the "
+                  f"device time prosper spends running it")
 
 
 def _fmt_rate(value):
@@ -560,6 +662,7 @@ def print_summary(summary):
           f"rendered={_fmt_rate(rates['rendered_fps'])} host-presented={_fmt_rate(rates['host_fps'])}")
     print(f"measured totals: graphics={summary['graphics_total_ms']:.1f} ms "
           f"compute={summary['compute_total_ms']:.1f} ms")
+    _print_compute_phases(summary)
     programs = summary["compute_programs"]
     print(f"compute identities: groups={programs['group_count']} "
           f"unknown={programs['unknown_records']} records/{programs['unknown_total_ms']:.1f} ms")
@@ -568,16 +671,19 @@ def print_summary(summary):
         if group["address_count"] > len(group["addresses"]):
             addresses += f",+{group['address_count'] - len(group['addresses'])}"
         parts = []
-        if group.get("setup_ms", 0.0) >= 0.05:
-            parts.append(f"setup={group['setup_ms']:.1f}ms")
-        if group.get("writeback_ms", 0.0) >= 0.05:
-            parts.append(f"wb={group['writeback_ms']:.1f}ms")
-        if group.get("dispatch_wait_ms", 0.0) >= 0.05:
-            parts.append(f"wait={group['dispatch_wait_ms']:.1f}ms")
+        for field in COMPUTE_CPU_PHASES:
+            if group.get(field, 0.0) >= 0.05:
+                parts.append(f"{PHASE_LABELS[field]}={group[field]:.1f}ms")
+        # Name the remainder instead of leaving it to subtraction: an unmeasured cost hides here.
+        cpu_attributed = sum(group.get(field, 0.0) for field in COMPUTE_CPU_PHASES)
+        unattributed = group.get("total_ms", 0.0) - cpu_attributed
+        if unattributed >= 0.05:
+            parts.append(f"unattributed={unattributed:.1f}ms")
         if group.get("gpu_device_ms", 0.0) >= 0.05:
             parts.append(f"dev={group['gpu_device_ms']:.1f}ms")
-        if group.get("gpu_shader_ms", 0.0) >= 0.05 and abs(group.get("gpu_shader_ms", 0.0) - group.get("gpu_device_ms", 0.0)) >= 0.05:
-            parts.append(f"shader={group['gpu_shader_ms']:.1f}ms")
+        for field in COMPUTE_GPU_BRACKETS:
+            if group.get(field, 0.0) >= 0.05:
+                parts.append(f"{PHASE_LABELS[field]}={group[field]:.1f}ms")
         breakdown_str = f" ({' '.join(parts)})" if parts else ""
         print(f"  {group['program_hash']} records={group['records']} "
               f"dispatches={group['dispatches']} total={group['total_ms']:.1f} ms "
--- a/prosper/tools/perf/performance_capture_report.py
+++ b/prosper/tools/perf/performance_capture_report.py
@@ -631,9 +631,8 @@ def _print_compute_phases(summary):
                  if abs(unattributed) >= 0.05 else None)
         _print_phase_line(f"GPU brackets of {device:.1f}ms device", gpu, device, extra)
         shader = gpu.get("gpu_shader_ms", 0.0)
-        if device > 0 and shader >= 0.0:
-            print(f"    the guest's own compute shader is {100.0 * shader / device:.1f}% of the "
-                  f"device time prosper spends running it")
+        print(f"    the guest's own compute shader is {100.0 * shader / device:.1f}% of the "
+              f"device time prosper spends running it")
 
 
 def _fmt_rate(value):
@@ -670,21 +669,33 @@ def print_summary(summary):
         addresses = ",".join(group["addresses"])
         if group["address_count"] > len(group["addresses"]):
             addresses += f",+{group['address_count'] - len(group['addresses'])}"
-        parts = []
+        # Two scopes on one line, so they are LABELLED and the nesting is stated. Printed flat,
+        # these numbers invite being added up: the CPU phases partition total, the GPU brackets
+        # partition dev, and dev is itself inside wait -- so a flat list of eleven fields sums to
+        # well over the total on the same line and reads as an arithmetic error in the tool.
+        cpu_parts = []
         for field in COMPUTE_CPU_PHASES:
             if group.get(field, 0.0) >= 0.05:
-                parts.append(f"{PHASE_LABELS[field]}={group[field]:.1f}ms")
+                cpu_parts.append(f"{PHASE_LABELS[field]}={group[field]:.1f}ms")
         # Name the remainder instead of leaving it to subtraction: an unmeasured cost hides here.
+        # Signed, and printed on either side of zero, matching the convention this file uses for
+        # every other remainder -- a negative one is a real producer state, not a rounding artifact.
         cpu_attributed = sum(group.get(field, 0.0) for field in COMPUTE_CPU_PHASES)
         unattributed = group.get("total_ms", 0.0) - cpu_attributed
-        if unattributed >= 0.05:
-            parts.append(f"unattributed={unattributed:.1f}ms")
-        if group.get("gpu_device_ms", 0.0) >= 0.05:
-            parts.append(f"dev={group['gpu_device_ms']:.1f}ms")
+        if abs(unattributed) >= 0.05:
+            cpu_parts.append(f"unattributed={unattributed:.1f}ms")
+        gpu_parts = []
         for field in COMPUTE_GPU_BRACKETS:
             if group.get(field, 0.0) >= 0.05:
-                parts.append(f"{PHASE_LABELS[field]}={group[field]:.1f}ms")
-        breakdown_str = f" ({' '.join(parts)})" if parts else ""
+                gpu_parts.append(f"{PHASE_LABELS[field]}={group[field]:.1f}ms")
+        sections = []
+        if cpu_parts:
+            sections.append("cpu " + " ".join(cpu_parts))
+        device_ms = group.get("gpu_device_ms", 0.0)
+        if device_ms >= 0.05:
+            inside = f" of which {' '.join(gpu_parts)}" if gpu_parts else ""
+            sections.append(f"gpu dev={device_ms:.1f}ms (inside wait){inside}")
+        breakdown_str = f" [{' | '.join(sections)}]" if sections else ""
         print(f"  {group['program_hash']} records={group['records']} "
               f"dispatches={group['dispatches']} total={group['total_ms']:.1f} ms "
               f"mean={group['mean_ms']:.2f} ms max={group['max_ms']:.2f} ms{breakdown_str} "
--- a/prosper/tools/perf/performance_capture_report.py
+++ b/prosper/tools/perf/performance_capture_report.py
@@ -590,7 +590,12 @@ def _print_phase_line(label, totals, denominator, extra=None):
         return
     parts = []
     for field, value in totals.items():
-        if value < 0.05:
+        # abs(): a negative is a VALUE, not an absence. Suppressing it hid the compensating
+        # term of a signed pair, so the visible parts summed to 183% of their own total while
+        # the remainder -- computed from the full set, hidden term included -- came to exactly
+        # zero and printed no warning. The one mechanism that would flag the inconsistency was
+        # silenced by the value it should have been reporting.
+        if abs(value) < 0.05:
             continue
         name = PHASE_LABELS.get(field, field)
         parts.append(f"{name}={value:.1f}ms ({100.0 * value / denominator:.1f}%)")
@@ -675,11 +680,14 @@ def print_summary(summary):
         # well over the total on the same line and reads as an arithmetic error in the tool.
         cpu_parts = []
         for field in COMPUTE_CPU_PHASES:
-            if group.get(field, 0.0) >= 0.05:
+            # abs() for the same reason as the summary table: see _print_phase_line.
+            if abs(group.get(field, 0.0)) >= 0.05:
                 cpu_parts.append(f"{PHASE_LABELS[field]}={group[field]:.1f}ms")
         # Name the remainder instead of leaving it to subtraction: an unmeasured cost hides here.
-        # Signed, and printed on either side of zero, matching the convention this file uses for
-        # every other remainder -- a negative one is a real producer state, not a rounding artifact.
+        # Signed on either side of zero, matching this file's convention for every remainder.
+        # No current producer path yields a negative remainder -- the five phases telescope to
+        # cleanup-start, so a skipped marker moves cost between phases without changing their
+        # sum (#3461) -- but the threshold must not be the thing that decides.
         cpu_attributed = sum(group.get(field, 0.0) for field in COMPUTE_CPU_PHASES)
         unattributed = group.get("total_ms", 0.0) - cpu_attributed
         if abs(unattributed) >= 0.05:
--- a/prosper/tools/perf/test_performance_capture_report.py
+++ b/prosper/tools/perf/test_performance_capture_report.py
@@ -410,6 +410,7 @@ BROKEN_DISPATCH_COMPUTE = [{
     "gpu_device_ms": 0.0, "gpu_timestamp_samples": 0,
 }]
 
+
 class ComputeDecompositionTests(unittest.TestCase):
     def _render(self, compute):
         summary = summarize(capture(SAMPLES, compute=compute))
@@ -545,10 +546,18 @@ class ComputeDecompositionTests(unittest.TestCase):
                        if line.strip().startswith("CPU phases of"))
         self.assertIn("pipeline=-10.0ms", summary)
         self.assertIn("pipeline=-10.0ms", self._group_line(text))
-        # And with it visible, the printed shares add up to the whole rather than to 183%.
-        shares = [float(part.rsplit("(", 1)[1].removesuffix("%)"))
-                  for part in summary.split() if part.endswith("%)")]
-        self.assertAlmostEqual(sum(shares), 100.0, places=1)
+        # And with it visible, the printed parts account for the whole rather than 183% of it.
+        #
+        # Summed in MILLISECONDS, not in the rendered percentages. Percentages are each rounded
+        # to one decimal before being printed, so their sum lands on 100.0 only when the errors
+        # happen to cancel: this fixture does (its suppressed phases are exactly 0.0 and its
+        # +/-0.033 errors are equal and opposite), while HIDDEN_COST_COMPUTE sums to 99.9 with
+        # nothing suppressed at all. A tolerance wide enough for both would no longer detect the
+        # 183% this arm exists to catch, so the check is done on the underlying values, as
+        # test_group_line_cpu_fields_account_for_its_own_total already does.
+        values = [float(part.split("=")[1].split("ms")[0])
+                  for part in summary.split() if "=" in part and "ms" in part]
+        self.assertAlmostEqual(sum(values), 12.0, places=3)
 
     def test_absent_gpu_timestamps_do_not_invent_a_bracket_table(self):
         from performance_capture_report import COMPUTE_GPU_BRACKETS
--- a/prosper/tools/perf/test_performance_capture_report.py
+++ b/prosper/tools/perf/test_performance_capture_report.py
@@ -419,8 +419,11 @@ class ComputeDecompositionTests(unittest.TestCase):
         return lines[0]
 
     def test_every_captured_timer_is_reported(self):
-        # Fails on the old report, which printed setup/writeback/wait/dev/shader and dropped the
-        # rest -- including the two that hold 92% of this dispatch.
+        # COVERAGE arm, deliberately whole-text: it asserts the summary carries every field and
+        # that each name reaches the output SOMEWHERE. It does not distinguish the summary from
+        # the group line -- either satisfies it -- so it survives a mutation of either surface.
+        # The surface-specific pinning is done by the summary arms and the _group_line arms;
+        # this one exists so a newly added timer that reaches neither surface still fails.
         from performance_capture_report import COMPUTE_CPU_PHASES, COMPUTE_GPU_BRACKETS
         summary, text = self._render(HIDDEN_COST_COMPUTE)
         for field in COMPUTE_CPU_PHASES:
@@ -489,6 +492,23 @@ class ComputeDecompositionTests(unittest.TestCase):
         values = [float(part.split("=")[1].removesuffix("ms"))
                   for part in cpu_section.split() if "=" in part]
         self.assertAlmostEqual(sum(values), 300.0, places=1)
+
+    def test_negative_remainder_is_shown_signed_on_both_surfaces(self):
+        # A record whose named phases exceed its own total. Reachable from the producer: a
+        # dispatch that breaks out before phase_pipeline leaves that interval unrecorded while
+        # total_ms still spans the batch, so the remainder goes negative.
+        #
+        # It must be PRINTED, not hidden by a one-sided threshold: a negative remainder means
+        # the phases do not partition the total, which is a producer defect worth seeing. This
+        # is the file's convention for every other remainder, and it was the one line in the
+        # change that no arm pinned.
+        over = [dict(HIDDEN_COST_COMPUTE[0])]
+        over[0]["total_ms"] = 290.0          # named CPU phases sum to 296.0
+        _, text = self._render(over)
+        self.assertIn("unattributed=-6.0ms", self._group_line(text))
+        cpu_summary = next(line for line in text.splitlines()
+                           if line.strip().startswith("CPU phases of"))
+        self.assertIn("unattributed=-6.0ms", cpu_summary)
 
     def test_absent_gpu_timestamps_do_not_invent_a_bracket_table(self):
         from performance_capture_report import COMPUTE_GPU_BRACKETS
--- a/prosper/tools/perf/test_performance_capture_report.py
+++ b/prosper/tools/perf/test_performance_capture_report.py
@@ -494,14 +494,24 @@ class ComputeDecompositionTests(unittest.TestCase):
         self.assertAlmostEqual(sum(values), 300.0, places=1)
 
     def test_negative_remainder_is_shown_signed_on_both_surfaces(self):
-        # A record whose named phases exceed its own total. Reachable from the producer: a
-        # dispatch that breaks out before phase_pipeline leaves that interval unrecorded while
-        # total_ms still spans the batch, so the remainder goes negative.
+        # A record whose named phases exceed its own total. NO CURRENT PRODUCER PATH BUILDS ONE,
+        # and the reason is worth stating so nobody re-derives it: the five CPU phases are
+        # consecutive differences over start->setup->pipeline->dispatch->writeback->cleanup
+        # (live_compute.cpp:11639-11643), and phase_milliseconds (:11635) is a plain signed
+        # difference with no clamping, so they telescope to cleanup-start for ANY ordering of the
+        # markers -- degenerate ones included. A break leaves a marker behind and yields a
+        # negative PHASE offset by a compensating positive one; the SUM is unchanged. Since
+        # total_ms spans at least that window, the remainder cannot go negative today.
         #
-        # It must be PRINTED, not hidden by a one-sided threshold: a negative remainder means
-        # the phases do not partition the total, which is a producer defect worth seeing. This
-        # is the file's convention for every other remainder, and it was the one line in the
-        # change that no arm pinned.
+        # An earlier version of this comment claimed a break before phase_pipeline produced one.
+        # It does not, and the claim reached here from a review citation that skipped the
+        # telescoping step -- which is why it is corrected rather than quietly dropped.
+        #
+        # The arm stays because the THRESHOLD is what is under test, not that mechanism: a
+        # one-sided threshold would silently hide a negative remainder if the telescoping ever
+        # broke (a clamp, or a phase measured outside the start..cleanup window), and a
+        # remainder that does not partition its total is precisely what this report exists to
+        # surface. It also matches the signed convention used for every other remainder here.
         over = [dict(HIDDEN_COST_COMPUTE[0])]
         over[0]["total_ms"] = 290.0          # named CPU phases sum to 296.0
         _, text = self._render(over)
--- a/prosper/tools/perf/test_performance_capture_report.py
+++ b/prosper/tools/perf/test_performance_capture_report.py
@@ -2,8 +2,11 @@
 
 import unittest
 
+import contextlib
+import io
+
 from performance_capture_report import (CaptureError, CLASSIFICATION_EVIDENCE_SHARE,
-                                        READBACK_NOTE_MIN_SHARE, summarize,
+                                        READBACK_NOTE_MIN_SHARE, print_summary, summarize,
                                         validate_capture)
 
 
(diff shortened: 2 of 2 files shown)


mattias800 (commented):
Independent review of `f61e7200` (verified as the current branch tip). I did not re-run CI or rebuild the emulator; I re-derived the decomposition against the producer, ran the suite, and mutation-tested the new arms.

## Verdict: REJECTED

Two blocking items, both on the same output line and both cheap. Everything else checked out, including the part I most expected to find wrong.

---

## What I verified as correct

**The nesting claim holds, exactly.** I traced it rather than taking the PR body's word:

- Six timestamps into one query pool, indices 0..5 in command-buffer order: `live_compute.cpp:9711` (index 0, `TOP_OF_PIPE`, right after `vkCmdResetQueryPool`), `:10113` (1, before `vkCmdDispatch`), `:10125` (2, after it), `:10224` (3, after the image transfer + optional retile), `:10349` (4, after compare), `:10380` (5, after the borrowed-image restore, immediately before `vkEndCommandBuffer`).
- `live_compute.cpp:10540-10559`: `device = ts[5]-ts[0]`, and the five brackets are `ts[1]-ts[0]`, `ts[2]-ts[1]`, `ts[3]-ts[2]`, `ts[4]-ts[3]`, `ts[5]-ts[4]`. Those telescope to `ts[5]-ts[0]` **by construction**, so "GPU brackets partition `gpu_device_ms`" is not an empirical claim that could drift — it is an identity. Note this survives the mixed pipeline stages (TOP_OF_PIPE / COMPUTE_SHADER / BOTTOM_OF_PIPE); the stages change what each *interval means*, not whether they sum.
- `COMPUTE_GPU_BRACKETS` correctly **excludes** `gpu_device_ms` and correctly **includes** `gpu_shader_ms` — `gpu_shader_ms` is `ts[2]-ts[1]`, i.e. one of the five intervals, not a parent. No double count in either table.
- `gpu_device_ms` sits inside `dispatch_wait_ms`: command-buffer recording begins after `phase_pipeline = ComputeClock::now()` at `:9572`, and `phase_dispatch` is taken at `:10562`, after the fence wait and after `vkGetQueryPoolResults`. Record, submit and wait all fall inside `[phase_pipeline, phase_dispatch]`, which is exactly `dispatch_wait_ms` (`:11641`).
- CPU phases: `:11639-11643` accumulate five adjacent intervals `phase_start -> phase_setup -> phase_pipeline -> phase_dispatch -> phase_writeback -> phase_cleanup`. Disjoint, no double count.
- `record.total_ms = elapsed` at `:12562` is the wall clock of the **whole batch**, including `execute_cpu_fast_path` items (`:12516`) that never enter the phase accumulators. So a non-negative CPU remainder is expected and physically meaningful — the `unattributed` row is the right response, not a bug.
- Denominators are consistent: `compute_total_ms` is `_total(compute, "total_ms")` (`performance_capture_report.py:428`) and `_phase_totals` sums over the same `compute` list (`:411`), so numerator and denominator have the same population.
- PR body's producer citations are accurate: `performance_capture.hpp:165` = `pipeline_ms`, `:168` = `cleanup_ms`; `performance_capture.cpp:464` / `:467` likewise.

**Housekeeping, all confirmed:** 32 tests pass on the branch (26 pre-existing + 6 new). `--json` loses **zero** keys and gains 19 (diffed old vs new summaries on a synthetic capture). 0 `Claude-Session:` trailers. `git diff --check` clean. Both files 100% CRLF, 0 bare LF. ctest registration is real at `prosper/CMakeLists.txt:1232`.

**The "fail without the fix" claim is accurate**, reproduced verbatim against `origin/main`'s tool: 4 assertion failures (`(73.3%)`, `GPU brackets of 65.0ms device`, `unattributed=4.0ms`, `compute shader is 6.2% of the device time`) and 2 `ImportError`s on the localised imports. The localised-import trick works as advertised.

**A hypothesis of mine that turned out wrong**, recorded so nobody re-derives it: I expected pre-2026-08-26 captures to carry `gpu_device_ms` but no brackets, which would make the tool print a fabricated `shader is 0.0% of device time`. I reproduced that output from a synthetic record — but it is unreachable from any real file. `e63f4038^:performance_capture.hpp:76-86` shows the old `ComputeTimingRecord` carried **no** `gpu_*` fields at all (the `gpu_device_ms` at `:43` of that revision belongs to `RendererTimingRecord`), so `device > 0` is false and the bracket table is correctly suppressed. Old captures are safe, and the PR's "existing `.prperf` files gain the new output with no re-capture" claim survives.

---

## Blocking

### B1. Half the diff is pinned by no test at all — mutation-proven

The per-group-line hunk (`performance_capture_report.py:673-685`) is the half that fixes the exact line the issue leads with (`0x2b1c4d31... (setup=6.3ms wb=17.8ms wait=16.1ms dev=15.7ms shader=0.4ms)`). I reverted **only** that hunk to its `origin/main` behaviour — three CPU timers plus `dev`, no brackets, no `unattributed` — left the summary half intact, and ran the full suite:

```
Ran 32 tests -- OK
```

The reason is the charter's named verification hazard, "a regression test with another path that can produce the expected result": `assertIn("pipeline=220.0ms", text)` and `assertIn("storage-copy=55.0ms", text)` are both satisfied by the *summary table's* substrings (`pipeline=220.0ms (73.3%)`), so they never reach the group line. Nothing in `ComputeDecompositionTests` isolates a line beginning with the program hash.

Ask: one arm that extracts the group line (e.g. the line containing `0x000000000000abcd`) and asserts the previously-dropped fields on **it** — `pipeline=220.0ms`, `cleanup=1.0ms`, `unattributed=4.0ms`, `storage-copy=55.0ms`. Cheap, and it is the arm the issue's own example calls for.

### B2. The per-group line merges the two decompositions with nothing marking the scope change

The summary tables are careful — separate rows, named denominators, a header stating the scope. The group line then prints, in one flat parenthesis (actual output, from the branch on the new test fixture):

```
0x000000000000abcd records=1 dispatches=1 total=300.0 ms mean=300.00 ms max=300.00 ms
  (setup=2.0ms pipeline=220.0ms wait=70.0ms wb=3.0ms cleanup=1.0ms unattributed=4.0ms
   dev=65.0ms pre=1.0ms shader=4.0ms storage-copy=55.0ms compare=5.0ms) addresses=...
```

Eleven numbers summing to **430 ms against the `total=300.0 ms` printed on the same line**. Four of them (`pre`/`shader`/`storage-copy`/`compare`) are inside `dev`, and `dev` is inside `wait`. This is precisely the arithmetic the PR body's own invariant exists to prevent — *"invite adding two overlapping decompositions into one bogus 100%"* — and the group line is where a reader does per-program arithmetic by hand. It got materially worse with this change: before it was 5 fields, now 11.

Ask: mark the scope inline. Something as small as `(cpu: setup=... unattributed=... | gpu dev=65.0ms: pre=... storage-copy=...)` is enough, and it also gives the brackets their denominator.

---

## Non-blocking

1. **`performance_capture_report.py:634`, `if device > 0 and shader >= 0.0:` is a tautological guard.** `device > 0` is already the enclosing condition at `:628`, and a millisecond accumulator is never negative. It reads as if it protects the headline against an unavailable timer; it does not. If the intent is availability, this file already has the right idiom at `:146` (`all("frontend_tex_other_n" in r for r in renderer)` -> `*_available`), which distinguishes *absent* from *zero*. Per the note above this is currently unreachable from real captures, so it is hygiene rather than a defect — but a dead guard invites the next reader to believe the case is handled.

2. **The per-group `unattributed` is one-sided (`:680`, `if unattributed >= 0.05`) while the summary uses `abs(...) >= 0.05` (`:626`, `:631`).** This file's established convention is a *signed* remainder: `:721` and `:755` print `other={:+.1f}`, and `:222-223` deliberately splits `res_other` from `res_over_attributed` with the comment "0 normally; non-zero is a defect" — the rule `performance_capture.hpp:133-136` states for #2245, that clamping "would make a broken partition look like a complete one". A negative group remainder is silently dropped here. Relatedly, `_print_phase_line:593` (`if value < 0.05: continue`) drops negative *phases* from the listing while still folding them into `unattributed`, which reports over-attribution as unmeasured cost.

   This is not purely theoretical: `phase_pipeline` is initialised to `phase_start` (`live_compute.cpp:5748-5751`) and only assigned at `:9572`, but `:9424`, `:9440`, `:9447`, `:9454`, `:9518`, `:9530` and `:9564` all `break` out before that on descriptor-pool / pipeline-creation failure, while the accumulation at `:11640` still runs — yielding a `pipeline_ms` negative by exactly `setup_ms`. That is a **producer** defect and out of scope for this PR; I raise it only because it is the reachable case that makes the signed-remainder rule load-bearing here. Worth its own issue.

3. **The PR body reformats the output.** The `CPU phases of 1472.0ms total: ...` example is shown wrapped across two lines; `_print_phase_line:600` emits one unwrapped line. Cosmetic, but PR bodies are the project record and the real line is long enough that the wrapping matters to whoever reproduces it.

4. The old group line suppressed `shader` when it equalled `dev` (`abs(shader - dev) >= 0.05`). That guard is gone, so a single-bracket dispatch now prints `dev=X shader=X`. Harmless; noting it so it is a decision rather than an oversight.

---

Fix B1 and B2 and I will re-review promptly — the analysis underneath this change is sound and the issue it closes is real. **REJECTED** on the current head.

mattias800:
Both blocking findings fixed in `317b7cde`. Thank you — B1 in particular was a genuine hole, and the mutation method is what exposed it.

**B1 — the group line was pinned by nothing.** Correct, and your diagnosis was exact: `assertIn("pipeline=220.0ms", text)` is satisfied by the summary table's substring, so it never reached the group line. Three arms now assert against the group line in isolation, via a `_group_line()` helper that extracts it and fails if it is not unique:

- `test_group_line_carries_the_dropped_timers` — pipeline / cleanup / storage-copy / unattributed on that line
- `test_group_line_separates_the_two_scopes` — scope markers present, and the cpu section contains no GPU field
- `test_group_line_cpu_fields_account_for_its_own_total` — the cpu fields, remainder included, sum to the total printed on the same line

Re-running your mutation (revert only the group-line hunk, leave the summary tables) now **fails 3 of them** — 2 failures and 1 error. I also confirmed the mutation actually lands before trusting the result: my first attempt silently did nothing (`read_text()` normalises CRLF, so a `\r\n` search never matched) and reported a meaningless green, which is the same class of mistake as the one you found.

**B2 — the flat eleven fields.** Also correct, and it was worse than the original: five fields to eleven, summing to 430 ms against a 300 ms total on the same line. The line now labels the scopes and states the nesting rather than leaving a reader to add them:

```
[cpu setup=24.6ms pipeline=0.7ms wait=189.6ms wb=60.1ms cleanup=0.2ms unattributed=0.4ms
 | gpu dev=185.5ms (inside wait) of which pre=2.2ms shader=1.0ms storage-copy=182.2ms]
```

**Non-blocking, both taken:** the dead `shader >= 0.0` guard is gone, and the per-group remainder is now signed with `abs()` thresholding like every other remainder in the file, so a negative one is shown rather than hidden. On your note that a reachable producer path can yield a negative `pipeline_ms` (the `:9424`-`:9564` breaks skipping `phase_pipeline` at `:9572`) — that is a producer-side question rather than a reporting one, and I would rather it be filed on its own than fixed silently here; say the word and I will file it, or file it yourself if you have the detail fresh.

I also appreciate you retracting your own "fabricated 0.0% shader share" hypothesis after checking `e63f4038^:performance_capture.hpp:76-86`. That check is the reason the PR's "no re-capture needed" claim can still be trusted.

Suite is **35 tests, OK**. Please re-review at `317b7cde`.

mattias800 (commented):
Re-review at head `317b7cde` (confirmed as the branch tip: `git ls-remote` and `gh pr view --json headRefOid` both give `317b7cde`). Both blocking items are fixed. I re-ran my own mutation rather than accepting the reported result, and added the converse mutation, because one direction alone does not establish what was asked.

## Verdict: APPROVED

---

## B1 — re-run independently, and the mutation is proven to have landed

You were right to flag the CRLF hazard; I hit the same class of thing on my first pass, so this run refuses to report anything unless it can show the edit took effect. The mutation script works on **bytes**, splits on literal `b"\r\n"`, asserts both anchors were found, and asserts `out != raw` before writing. Then, before running a single test, it renders the group line:

```
mutation written, delta bytes: -1083
GROUP LINE NOW: 0x000000000000abcd records=1 dispatches=1 total=300.0 ms mean=300.00 ms
  max=300.00 ms (setup=2.0ms wb=3.0ms wait=70.0ms dev=65.0ms shader=4.0ms) addresses=...
```

That is the pre-fix shape from the issue, so the revert is real and not a no-op. Result:

```
ERROR: test_group_line_cpu_fields_account_for_its_own_total
FAIL:  test_group_line_carries_the_dropped_timers
FAIL:  test_group_line_separates_the_two_scopes
Ran 35 tests -- FAILED (failures=2, errors=1)
```

Independently reproduces your 2 + 1.

**The converse mutation, which is the half that actually answers "are they pinned to the group line".** A single direction only shows the arms notice *something*; it does not show they are reading the group line rather than the summary. So I did the opposite edit — replaced `_print_compute_phases(summary)` in `print_summary` with `pass`, deleting both summary tables and leaving the group line intact:

```
FAIL: test_absent_gpu_timestamps_do_not_invent_a_bracket_table
FAIL: test_dominant_timer_reaches_the_printed_output
FAIL: test_gpu_brackets_are_scaled_against_device_time_not_total
FAIL: test_shader_share_of_device_time_is_stated
Ran 9 tests -- FAILED (failures=4)
```

Exactly the four summary arms, and **all three group-line arms stayed green**. The two mutations partition the arms cleanly: each half of the output is pinned by arms that fail when it is removed and survive when the other half is removed. That is the pair that settles question 2.

`_group_line`'s uniqueness guard (`assertEqual(len(lines), 1)`) is the right shape — it fails loudly rather than silently matching nothing.

**One honest caveat, not blocking:** `test_every_captured_timer_is_reported` still asserts `pipeline=220.0ms` / `storage-copy=55.0ms` against the **whole text**, and it did not fail under either mutation — the summary satisfies it in one direction and the group line in the other. It is now harmless (the three dedicated arms carry the load), but it is not a group-line arm and its comment reads as though it were. Worth either retargeting it to `_group_line` or trimming the comment.

## B2 — the nesting is now unambiguous

```
0x000000000000abcd ... total=300.0 ms mean=300.00 ms max=300.00 ms
  [cpu setup=2.0ms pipeline=220.0ms wait=70.0ms wb=3.0ms cleanup=1.0ms unattributed=4.0ms
   | gpu dev=65.0ms (inside wait) of which pre=1.0ms shader=4.0ms storage-copy=55.0ms compare=5.0ms]
```

Checking it the way the misreading would actually happen:

- The `cpu` section sums to `2.0 + 220.0 + 70.0 + 3.0 + 1.0 + 4.0 = 300.0`, which is the `total=300.0 ms` printed on the same line. A reader who adds up the first section gets the right answer, which is the strongest possible defence — it does not merely warn against the wrong sum, it makes the natural sum correct.
- The `gpu` section carries its own parent (`dev=65.0ms`), states where that parent lives (`inside wait`, and `wait` is a label printed in the cpu section, so the reference resolves), and subordinates its children with `of which` — which sum to `1.0 + 4.0 + 55.0 + 5.0 = 65.0`, i.e. to `dev`, not to anything in the cpu section.
- `|` plus two scope words makes cross-section addition a deliberate act rather than the default reading.

I probed three degenerate shapes for a format that could still lie; none does:

```
no-gpu      ... [cpu setup=2.0ms ... unattributed=4.0ms]
dev-no-brk  ... [cpu ... | gpu dev=65.0ms (inside wait)]                       <- no false "of which"
cpu-empty   ... [gpu dev=65.0ms (inside wait) of which pre=1.0ms ...]
```

The third prints `(inside wait)` with no `wait` label present, but it needs `total_ms == 0` alongside `dev == 65 ms`, which is self-contradictory and unreachable. Not worth code.

Two cosmetic notes, neither blocking:
- The **summary** bracket header still reads `GPU brackets of Xms device` with no statement of where device sits, while the group line now says `(inside wait)`. The group line is currently more informative than the summary about the very invariant the summary exists to protect. A parenthetical in `_print_compute_phases` would close the gap.
- The group line names the CPU remainder but has no GPU remainder. Since the brackets telescope exactly (`live_compute.cpp:10540-10559`), the only possible gap is sub-0.05 suppression, so this is genuinely cosmetic — but it is a small asymmetry against this PR's own "any remainder is named" rule.

## Question 3 — did `abs()` alter any existing expected output?

**No, and I checked it two ways rather than by reading the diff.**

- Reverting the group line's `abs(unattributed) >= 0.05` back to `unattributed >= 0.05` on the current head leaves the suite at **35 tests, OK**. So no existing arm produces a negative group remainder, and nothing observable in the suite changed.
- The behaviour itself is correct, verified by hand with `total_ms=290.0` against phases summing to 296 (a 6 ms over-attribution). Both surfaces print it signed and consistently: summary `unattributed=-6.0ms (-2.1%)`, group line `... cleanup=1.0ms unattributed=-6.0ms | gpu ...`.

That second bullet is also the flip side: **this is now the one line in the PR that no test pins**, and a future edit reverting it would pass CI silently. By the standard I applied in B1 that is the same species of gap, so I am naming it — but not blocking on it, because the behaviour is verified correct, it only fires in a producer-defect state, and the convention it now matches is documented three times in this same file (`:721`, `:755`, `:222-223`) and in `performance_capture.hpp:133-136`. One arm with `total_ms=290.0` asserting `unattributed=-6.0ms` on `_group_line` would close it cheaply.

The dead `shader >= 0.0` guard is gone; the surviving `print` is correctly unconditional inside the `if device > 0:` block.

## Verification on the new head

- **35 tests, OK** (26 pre-existing + 9 in `ComputeDecompositionTests`).
- `--json` is **byte-identical** to the previous head (`values identical: True`) and loses **zero** keys against `origin/main` — the fix touched only print paths, as intended.
- Diff against `origin/main` touches exactly the two files. 0 `Claude-Session:` trailers. `git diff --check` clean. Both files 809/507 CRLF, **0 bare LF**.
- No `.md` in the diff, so no docs gate; no screenshot, so no `BLOG.md` entry required.
- `pr_merge_gate.py 3451`: `head matches branch tip (pr=317b7cde tip=317b7cde)`, `pass=7 fail=0 pending=4`. My approval is on content — the four pending checks still have to go green before merge, and the head must still match the tip at that moment.

Everything I verified on `f61e7200` and did not re-check here still stands: the decomposition, the nesting, the absence of double counting, and the finding that pre-2026-08-26 captures are unaffected (`e63f4038^:performance_capture.hpp:76-86`).

**APPROVED** on `317b7cde`.

mattias800:
Thank you for the re-review, and for the **converse mutation** — deleting `_print_compute_phases` while keeping the group line, to show the group-line arms fail on the group line rather than inheriting a pass from the summary. That is the half I did not do, and it is the half that actually answers "are they pinned". I have adopted it as the method for this file.

Both non-blocking items are now closed in `193940c4`, because the first is the same species as the blocking finding and I would rather not ship a known-untested line in a PR about exactly that.

**The signed remainder is now pinned.** `test_negative_remainder_is_shown_signed_on_both_surfaces` feeds a record whose named phases exceed its own total (`total_ms=290.0` against 296.0 of phases) and requires `unattributed=-6.0ms` on both surfaces. Verified by mutation **on each surface separately**:

- revert `abs()` on the group line → fails
- revert `abs()` on the summary CPU table → fails

so it pins both rather than passing on one and inheriting the other. An earlier draft asserted the summary via `text.splitlines()[9]`; that would have rotted on any added output line, so it now finds the `CPU phases of` line by content.

**The coverage arm now says what it covers.** You are right that `test_every_captured_timer_is_reported` survives a mutation of either surface and its comment implied otherwise. It is a coverage arm — every field reaches the output *somewhere*, so a newly added timer that reaches neither surface still fails — and the comment now states that and points at the arms doing the surface-specific pinning.

Suite is **36 tests, OK**.

On the merge gate: agreed, and that is how I will treat it. Your approval is on content; I will not merge until `pr_merge_gate.py` reports GREEN with zero pending **and** the recorded head still equals the branch tip at that moment. Note the head has moved to `193940c4` since your approval — the delta is test-only (`test_performance_capture_report.py`), no change to `performance_capture_report.py`, so `git diff 317b7cde 193940c4 --stat` is one file. If you would rather re-confirm on the new head than have me assert that, say so and I will hold.

mattias800 (commented):
Re-review at head `193940c4`. The mutations all pass; the new arm is genuinely pinned on both surfaces, exactly as you measured. But the new arm's comment states a producer mechanism that is **false**, and I am the one who put it there — it came out of my `f61e7200` review, where I conflated a negative *phase* with a negative *remainder*. So this is a narrow block on three lines of comment, not on the test.

## Verdict: REJECTED

---

## The delta is test-only, verified by blob hash rather than by `--stat`

`git rev-parse 317b7cde:prosper/tools/perf/performance_capture_report.py` and the same at `193940c4` both give `57e94e42a23f81235d18942620a6dac51874724d`. Identical blob, so the approved tool is bit-for-bit unchanged and I only had to re-examine the arms. `git diff 317b7cde 193940c4 --stat` is one file, +22/-2.

## The four mutations — all four behave

Baseline on `193940c4`: **36 tests, OK**.

**M1, revert the group-line hunk** (bytes-level, CRLF-safe, and it prints the reverted line before running anything, so a no-op edit cannot report green):

```
PROOF group line now: ... (setup=2.0ms wb=3.0ms wait=70.0ms dev=65.0ms shader=4.0ms)
FAILED (failures=3, errors=1)
  test_group_line_carries_the_dropped_timers
  test_group_line_separates_the_two_scopes
  test_group_line_cpu_fields_account_for_its_own_total
  test_negative_remainder_is_shown_signed_on_both_surfaces
```

**M2, delete the summary tables** (`_print_compute_phases(summary)` -> `pass`):

```
FAILED (failures=4, errors=1)
  test_dominant_timer_reaches_the_printed_output
  test_gpu_brackets_are_scaled_against_device_time_not_total
  test_shader_share_of_device_time_is_stated
  test_absent_gpu_timestamps_do_not_invent_a_bracket_table
  test_negative_remainder_is_shown_signed_on_both_surfaces
```

The new arm fails under both, which is correct — it spans both surfaces by design.

**M3a / M3b, one-sided threshold on each surface separately.** This is the check that matters for your claim, and it holds. Reverting `abs()` on the group line only (1 of 1 site) and on the summary CPU table only (1 of the 2 identical `abs(unattributed)` sites, disambiguated by index so the GPU one stays intact) each fail the new arm alone, `failures=1`, and each trips **its own** assertion:

```
M3a -> line 508: self.assertIn("unattributed=-6.0ms", self._group_line(text))
M3b -> line 511: self.assertIn("unattributed=-6.0ms", cpu_summary)
```

So it does not pass on one surface and inherit the other. Confirmed independently.

The summary line lookup is content-based (`next(line for line in ... startswith("CPU phases of"))`) and raises `StopIteration` when the line is gone rather than silently matching — it showed up as an ERROR under M2, which is the loud outcome. Good call dropping the `splitlines()[9]` draft; that would have rotted on the next added line.

`test_every_captured_timer_is_reported`'s rewritten comment is accurate: it survived both M1 and M2 in my runs, which is exactly what a whole-text coverage arm should do, and the comment now says so and points at the arms that discriminate.

---

## Blocking: the new arm's comment states a mechanism that does not produce the case

```python
# A record whose named phases exceed its own total. Reachable from the producer: a
# dispatch that breaks out before phase_pipeline leaves that interval unrecorded while
# total_ms still spans the batch, so the remainder goes negative.
```

The last clause is wrong. **That mechanism produces a negative phase and can never produce a negative remainder**, because the five phases telescope exactly, and they do so regardless of ordering:

`(A-start) + (B-A) + (C-B) + (D-C) + (E-D) = E - start`

for any values of `A..D`, including the degenerate ones. Concretely, `live_compute.cpp:5748-5752` initialises `phase_setup/pipeline/dispatch/writeback` to `phase_start`; if a dispatch breaks out at `:9424`-`:9564` after `phase_setup` is taken at `:9385` but before `phase_pipeline` at `:9572`, then with `start=0, A=10, B=C=D=0, E=12` the accumulators at `:11639-11643` receive `setup=+10, pipeline=-10, wait=0, wb=0, cleanup=+12` — sum `12`, i.e. exactly `phase_cleanup - phase_start`, the item's own duration. The `do { ... } while (false)` at `:11547` guarantees every break reaches `phase_cleanup` at `:11638`, so this always holds.

And `total_ms` (`:12562`) is the **batch** wall from `:12492`, spanning every item plus the `execute_cpu_fast_path` items at `:12516` that contribute no phases at all. Both use `std::chrono::steady_clock` (`ComputeClock` is aliased to it at `:5743`), so there is no cross-clock skew to exploit either. Therefore `sum(phases) <= total_ms` **always**, and the remainder is non-negative on every path the current producer has.

**This is my error propagating, not yours.** My `f61e7200` review said the break paths were "the reachable case that makes the signed-remainder rule load-bearing here" — true of `pipeline_ms`, false of the remainder, and I did not do the telescoping step before writing it. You took a cited `file:line` claim from a review in good faith, which is precisely the failure mode `CLAUDE.md` describes: *"a wrong finding with a citation is the most contagious kind of error this project produces."* It is now heading into a committed test comment where the next reader inherits it with a passing arm appearing to vouch for it. That is why I am blocking on a comment.

**Keep the test.** It is worth having — it pins a convention this file documents three times, and it is the arm that would catch a future producer or aggregation change that does make the phases overlap. Only the justification needs replacing. Something like:

```python
# A record whose named phases exceed its own total. DEFENSIVE: not reachable from the
# current producer -- the five phases telescope to phase_cleanup - phase_start whatever
# order the markers land in (live_compute.cpp:11639-11643), and total_ms spans the whole
# batch on top of that, so today's remainder is always >= 0. Pinned anyway because the
# printing convention is what protects the reader if that ever stops being true: a
# negative remainder means the phases do not partition the total, and a one-sided
# threshold would hide exactly that. It was the one line in the change no arm pinned.
```

Nothing else in the arm changes; I re-ran it against that wording in my tree and it still fails under M3a and M3b.

## Separately, worth its own issue (not this PR)

The same telescoping walk turned up a real producer defect I had not noticed: on a break path `phase_writeback` also stays at `phase_start`, so `cleanup_ms = phase_cleanup - phase_start` books **the entire item's duration to `cleanup`**, while `pipeline_ms` carries a compensating negative. A failing dispatch therefore reports its whole cost in the wrong bucket, and the report now prints that bucket prominently. The totals stay right, which is why nothing catches it. Worth filing against `live_compute.cpp:5748-5752` / `:11639-11643` — the fix is to record the markers as `optional` and attribute only measured intervals.

## Housekeeping on `193940c4`

36 tests OK; diff vs `origin/main` touches only the two `tools/perf` files; 0 `Claude-Session:` trailers; `git diff --check` clean; both files all-CRLF with 0 bare LF; `--json` byte-identical to `317b7cde`. No `.md` and no screenshot, so no docs gate and no `BLOG.md` obligation.

Fix the comment and I will approve on sight — everything else on this head is verified and I have no further findings. **REJECTED** on `193940c4`.

mattias800:
Fixed in `034ba69c`, and you were right to block on it — thank you for catching your own citation rather than letting it stand.

**I verified the telescoping independently before accepting the correction**, because you had just demonstrated that a cited `file:line` from this very review could be wrong, and swapping one assertion for another would repeat the error in the other direction:

- `live_compute.cpp:11639-11643` — the five accumulators are consecutive differences over `start -> setup -> pipeline -> dispatch -> writeback -> cleanup`.
- `live_compute.cpp:11635-11637` — `phase_milliseconds` is `duration<double, milli>(end - begin).count()`, a **plain signed difference with no clamping**. That is the load-bearing detail: a clamp would break the telescoping and could make the sum exceed the total.

So they telescope to `cleanup - start` for any ordering, a break yields a negative phase offset by a compensating positive, and the remainder cannot go negative on any current path. Your correction holds.

The comment now states that mechanism, records that an earlier version claimed the opposite and how it got there, and says what the arm actually pins: **the threshold, not the mechanism**. A one-sided threshold would hide a negative remainder if the telescoping ever broke — a clamp, or a phase measured outside the `start..cleanup` window — and a remainder that fails to partition its total is what this report exists to surface. Still verified by mutation on each surface separately.

**Your producer finding is filed as #3461.** On a break path `phase_writeback` stays at `phase_start`, so `cleanup_ms` books the entire item duration against a compensating negative `pipeline_ms`. Totals stay correct; attribution does not. I noted there that this PR is what makes it reader-visible, since it is the first to print those two buckets — so a large `cleanup` with a negative `pipeline` will now appear and reads naturally as "teardown is expensive", which is wrong.

36 tests OK. Please re-review at `034ba69c` — the delta from `193940c4` is a comment in one test file; `performance_capture_report.py` is unchanged from the head you approved.

mattias800 (commented):
Re-review at head `034ba69c`. The comment correction is accurate and I have no complaint about it. But question (2) — "does anything else still rest on the retracted mechanism" — has a real answer: yes, one comment in the shipped tool. And chasing question (3) turned up something larger that I under-weighted in my first review: **the printer this PR adds silently drops a negative phase, and on a break-path record it prints a decomposition summing to 183% of its own total with no remainder row.** That is this PR dropping a bucket from the output, which is the defect it exists to fix.

## Verdict: REJECTED

Two one-line changes. Both verified in my tree; the suite stays at 36 OK with them.

---

## (1) The corrected comment is accurate, and I found no overclaim

Both citations check out against the source, not against your description of it:

- `live_compute.cpp:11635` is `auto phase_milliseconds = [](auto begin, auto end) {`, whose body at `:11636` is `duration<double, std::milli>(end - begin).count()` — a plain signed difference, no clamp, no `max`. That is the load-bearing detail and you identified it correctly; a clamp there would indeed break the telescoping and let the sum exceed the total.
- `live_compute.cpp:11639-11643` are the five `+=` accumulators over consecutive markers.

The reasoning holds: the terms telescope to `phase_cleanup - phase_start` for any ordering, and `total_ms` (`:12562`, from `:12492`) spans at least that window plus the `execute_cpu_fast_path` items at `:12516` that contribute no phases. I also checked the case the comment does not mention — the capture arming or disarming mid-batch — and it only ever *removes* phases, so it cannot flip the sign either. "NO CURRENT PRODUCER PATH BUILDS ONE" stands as written.

Keeping the retraction paragraph rather than quietly deleting the old claim is the right call, and saying the arm pins the **threshold** rather than the mechanism is exactly the correct description of what it tests. Re-verified: one-sided threshold on the group line alone and on the summary CPU table alone each fail this arm and only this arm, `failures=1`, tripping their own assertions.

Delta confirmed comment-only by blob hash: `performance_capture_report.py` is `57e94e42…` at `317b7cde`, `193940c4` and `034ba69c` alike.

## (2) Yes — one residual, in the shipped tool

`performance_capture_report.py:681-682`:

```python
# Signed, and printed on either side of zero, matching the convention this file uses for
# every other remainder -- a negative one is a real producer state, not a rounding artifact.
```

"a negative one is a real producer state" **is** the retracted claim. As of this head the PR now retracts it in `test_performance_capture_report.py` and continues to assert it in `performance_capture_report.py`, and the tool is the more authoritative of the two files. A reader who opens only the tool inherits the error the last commit exists to kill.

This was present at `317b7cde` and I approved that head without catching it — my miss, not a regression, and I had no reason to look because my own review was making the same claim. First half of the sentence (signed, matching the convention) is fine; only the justification needs replacing, e.g. *"...for every other remainder. No current producer path yields one — see `test_negative_remainder_is_shown_signed_on_both_surfaces` — but a one-sided threshold would hide it if that changed."*

I checked the other `negative` comment, `:212`, and it is **not** affected: it covers the renderer's `setup_resources` sub-buckets, which are independently measured scopes rather than telescoping differences, so over-attribution genuinely is reachable there and #2245's reasoning still applies verbatim. Leave it alone.

## (3) #3461's framing is right, with one understatement I can now demonstrate

`:11547`, `:11635`, `:11639-11643` all verified; the worked shape is correct; and "attribution defect, not a measurement one" is the right classification for the **stored** values — every total in the file remains true, which is what makes the fix safe.

But "no total is currently wrong" understates what a reader will see, because the printer drops the compensating negative. Feeding the issue's own worked shape (`setup=+10, pipeline=-10, wait=0, wb=0, cleanup=+12`, `total_ms=12`) to this PR's report:

```
CPU phases of 12.0ms total: setup=10.0ms (83.3%) cleanup=12.0ms (100.0%)
0x...abcd ... total=12.0 ms [cpu setup=10.0ms cleanup=12.0ms]
```

Three things go wrong at once, and they compound:

- `pipeline=-10.0ms` is **suppressed** — `_print_phase_line:593` is `if value < 0.05: continue`, which catches negatives as well as sub-threshold positives. The group-line loop at `:677-679` is one-sided in the same way.
- The visible parts therefore sum to **22.0 against a 12.0 total, 183%**.
- And there is **no `unattributed` row to warn about it**, because the remainder is computed from the full set including the hidden negative and comes to exactly 0. The one mechanism that would have flagged the inconsistency is silenced by the same value it should be reporting.

So the line reads `cleanup=12.0ms (100.0%)` — "teardown is the entire cost" — for a dispatch that failed during setup. This is the same class of defect as #3447 itself: a bucket the capture carries, dropped by the printer, with the output looking complete.

It also matters more than a rare edge case, because of *which* population hits it. Failing compute dispatches are not exotic on the title this report was built for — GTA V is where `PROSPER_COMPUTE_SKIP_PROGRAM` and the unsupported-program census live (#2481) — so the first real consumer of this feature is the run most likely to contain break-path records.

**This is the second half of non-blocking item 2 in my `f61e7200` review.** I described the mechanism and did not render it, and a rendering is what shows it produces a 183% line rather than a cosmetic omission. I was wrong to file it as non-blocking, and this PR is where it should be fixed, because this PR is what puts these two timers on screen.

Verified fix, two lines:

```python
_print_phase_line:593    if abs(value) < 0.05: continue
group-line loop :677     if abs(group.get(field, 0.0)) >= 0.05:
```

With both, the same record renders honestly and the parts add up:

```
CPU phases of 12.0ms total: setup=10.0ms (83.3%) pipeline=-10.0ms (-83.3%) cleanup=12.0ms (100.0%)
0x...abcd ... total=12.0 ms [cpu setup=10.0ms pipeline=-10.0ms cleanup=12.0ms]
```

Suite stays **36 tests, OK** with the change, so nothing existing depends on the suppression. The GPU bracket loop at `:687-689` can take the same treatment for symmetry, though brackets are masked unsigned tick differences (`live_compute.cpp:10547-10559`) and cannot go negative, so there it is cosmetic.

Worth an arm, since the same reasoning that earned the remainder one applies: a record with `pipeline_ms=-10.0` asserting `pipeline=-10.0ms` appears on both surfaces. And #3461 deserves a line noting that the printed decomposition, not just the attribution, is visibly wrong today — that is the reader-facing symptom someone will actually report.

---

## Housekeeping on `034ba69c`

36 tests OK; diff vs `origin/main` touches only the two `tools/perf` files; 0 `Claude-Session:` trailers; `git diff --check` clean; all-CRLF, 0 bare LF; `--json` byte-identical to the approved head.

Fix the `:681-682` comment and the two one-sided thresholds and I will approve. Everything else on this head is verified and I have no further findings. **REJECTED** on `034ba69c`.

mattias800 (commented):
Re-review at head `ce646d4d`. Both fixes are correct and the new arm discriminates on both surfaces. Your three questions are good ones and the answer to (2) is **yes, you got lucky** — I measured it rather than reasoning about it, and the assertion would fail on two of the fixtures already in this file. That is a note on the test, not a defect in the tool, so it does not block.

## Verdict: APPROVED

---

## (1) Does `abs()` break any case where suppressing a small negative was right?

**No.** The threshold *magnitude* is unchanged at 0.05 ms, so everything that was being filtered as noise is still filtered — `abs()` only changes which side of zero the window is centred on. What newly prints is `v <= -0.05 ms`, and that cannot be manufactured by noise:

- A single phase is a difference of two `steady_clock` reads (`ComputeClock`, `live_compute.cpp:5743`). A monotonic clock cannot produce a negative difference, and clock granularity is ~100 ns on Windows, three orders below the threshold.
- The accumulators are sums; a sum of non-negative doubles cannot go negative, and float error on ms-scale magnitudes is ~1e-13.

So a negative of 50 µs or more requires a genuine marker inversion — the #3461 state — which is exactly the signal you want visible. Sub-threshold negatives (a `pipeline_ms` that lands at, say, −0.03 ms because break-path records nearly cancel good ones across a capture) are still suppressed exactly as before.

Worth noting the same `abs()` now also governs the **GPU** bracket table, since both tables share `_print_phase_line`. That is a no-op: brackets are masked unsigned tick differences (`live_compute.cpp:10547-10559`), so they cannot be negative. No case changes there.

## (2) Does the shares-sum-to-100 assertion generalise? — measured: no

The identity is `Σ displayed shares = (total − Σ suppressed) / total × 100`, with each term then rounded to 1 dp. Both of those can break it, and I ran the assertion against the fixtures already in this file:

```
BROKEN_DISPATCH          shares sum = 100.0000   assertAlmostEqual(..., 100.0, places=1) -> True
HIDDEN_COST              shares sum =  99.9000   -> False
negative-remainder(290)  shares sum =  99.9000   -> False
sub-threshold cleanup    shares sum = 100.0000   -> True
```

`BROKEN_DISPATCH` lands on exactly 100.0 for two independent fixture-specific reasons: its two suppressed phases are exactly `0.0` (so nothing is lost to suppression), and the ±0.033 rounding errors on `setup=83.3` and `pipeline=-83.3` are equal and opposite and cancel. `HIDDEN_COST`'s exact shares sum to 100 and it still reads 99.9 — six 1-dp roundings, no suppression involved at all.

So the assertion is sound as a check on **this fixture** and is not an invariant of the code. Nothing is wrong today; the hazard is a future editor adding a phase or nudging a value in `BROKEN_DISPATCH_COMPUTE` and getting a failure that has nothing to do with the bug the arm guards, then "fixing" it by widening the tolerance and quietly losing the check. Cheapest robust form is to sum the parsed **ms** values against `total_ms` (which the sibling arm `test_group_line_cpu_fields_account_for_its_own_total` already does) rather than the rounded percentages; `delta=0.5` or a comment saying the exact 100.0 is fixture-specific would also do. Non-blocking either way.

The arm's discriminating power is confirmed: reverting `_print_phase_line`'s `abs()` alone and the group-line loop's `abs()` alone each fail it and only it, `failures=1` in both runs.

## (3) Anything else the negative-value path touches

Three things, none blocking:

**a. The group-line GPU loop at `:695-697` is still one-sided** (`if group.get(field, 0.0) >= 0.05`). It is now the only one-sided threshold left in the compute path and is the same shape as the bug just fixed, so the next reader will wonder. It is genuinely safe — brackets cannot be negative — but the reason is worth one line, because **the failure mode on that path is not negativity, it is wrap**: `(timestamps[i] - timestamps[j]) & mask` on out-of-order timestamps yields an enormous positive, not a negative. Either add `abs()` for symmetry or say why it is exempt.

**b. A benefit of the `:631` `abs()` that is worth recording, since it is what makes (a) safe.** An over-attributed GPU bracket now surfaces loudly rather than silently. Injecting `gpu_storage_copy_ms=500.0` against `gpu_device_ms=65.0` — the shape a wrapped subtraction would produce:

```
GPU brackets of 65.0ms device: pre=1.0ms (1.5%) shader=4.0ms (6.2%) storage-copy=500.0ms (769.2%)
  compare=5.0ms (7.7%) unattributed=-445.0ms (-684.6%)
```

Unmissable, which is the right outcome and the same principle as the CPU fix.

**c. Cosmetic:** one blank line between `BROKEN_DISPATCH_COMPUTE`'s closing `}]` at `:411` and `class ComputeDecompositionTests` at `:413` (PEP8 wants two before a top-level class), and three now precede the fixture comment. No linter gates this repo, so purely a note.

## The fixes themselves

Both land correctly. The `_print_phase_line` comment explains the mechanism rather than the edit, including the part that made it subtle — that the `unattributed` row was silenced by the very value it should have reported. The group-line comment at `:684-688` now states the telescoping and cites #3461 instead of asserting the retracted producer claim, which closes the contradiction between the two files; I re-grepped and there is no remaining "negative ... producer state" claim anywhere in the tool. `:212` is untouched and correctly so — different subsystem, independently measured scopes, where over-attribution really is reachable.

## Housekeeping on `ce646d4d`

**37 tests, OK.** Diff vs `origin/main` touches only the two `tools/perf` files. 0 `Claude-Session:` trailers. `git diff --check` clean. Both files all-CRLF, 0 bare LF. `--json` byte-identical to the head I approved at `317b7cde`, so no consumer of the JSON sees any of this. `py_compile` clean. No `.md` in the diff and no screenshot, so no docs gate and no `BLOG.md` obligation.

Nothing here blocks. The three notes above are all optional follow-ups; (2) is the one I would actually take, since a brittle assertion tends to get widened rather than understood.

For the merge step: my approval is on content at `ce646d4d`. `pr_merge_gate.py 3451` must still show every check passed, none pending, and `headRefOid` equal to the branch tip at the moment of merge — a push after CI starts detaches both the checks and this review.

**APPROVED** on `ce646d4d`.

mattias800 (commented):
Re-review at head `826e706d`. Verdict carries. Taking your second question first, because it is the one that matters most.

## Verdict: APPROVED

---

## (2) You were right to ask: I asserted it. I have now derived it, and it holds.

**Plainly: in my previous review I asserted the wrap claim.** I read the masking expression, saw `& mask` on unsigned operands, and inferred "wraps to an enormous positive" without walking the modular-arithmetic step or bounding the magnitude. It was a correct inference, but it was an inference, and you were right not to commit it on my say-so — this is the same route my negative-remainder claim took into a comment earlier in this PR, and asking is what breaks the pattern.

Here is the derivation, done properly this time.

1. `live_compute.cpp:10534` — `uint64_t timestamps[6]{}`. Unsigned.
2. `:10538-10539` — `mask = timestamp_valid_bits >= 64 ? UINT64_MAX : ((uint64_t{1} << timestamp_valid_bits) - 1u)`.
3. `:10547-10559` — every bracket is `static_cast<double>((timestamps[i] - timestamps[j]) & mask) * timestamp_period_ns / 1'000'000.0`.
4. Unsigned integer arithmetic in C++ is modulo `2^64`. So for an inverted pair the subtraction yields `2^64 - Δ`, and the mask truncates it to `2^valid_bits - Δ`. There is no signed intermediate anywhere in the expression.
5. `static_cast<double>` of a `uint64_t` is non-negative by construction. **A GPU bracket therefore cannot be negative** — this is structural, not empirical.
6. Magnitude, computed rather than adjectival:

   ```
   inverted pair, period 1 ns:   64 valid bits -> 1.845e+13 ms
                                 36 valid bits ->    68719 ms
                                 32 valid bits ->     4295 ms
   ```

   `timestamp_valid_bits` is `uint32_t` (`:1403`), taken from `families[queue_family].timestampValidBits` (`:1700`), and a zero width is rejected outright (`:1704`). So even the narrowest plausible counter puts an inverted bracket about five orders of magnitude above the 0.05 ms threshold.
7. `abs()` is the identity on a non-negative value.

All three clauses of the comment hold. Two things I would add to your confidence in it, which I had not established before either:

- **The scenario is not hypothetical.** The six writes use three different pipeline stages — `:9711` `TOP_OF_PIPE`, `:10113`/`:10125` `COMPUTE_SHADER`, `:10224`/`:10349`/`:10380` `BOTTOM_OF_PIPE`. Vulkan guarantees each timestamp is written once prior commands reach *that* stage; it does not guarantee the recorded values are monotonic across *different* stages. So the comment guards a real possibility, which is the justification for it existing.
- **The detection story closes.** `device_ticks` reads only `timestamps[5]` and `timestamps[0]` (`:10540`), so an inverted *intermediate* leaves `gpu_device_ms` correct while the brackets stop summing to it — exactly the signature the `abs()`-gated remainder at `:631` makes loud. That is what makes the one-sided GPU threshold safe rather than merely harmless, and the comment already says it.

One precision note, optional: "enormous positive" is `2^valid_bits x period`, a device-dependent quantity rather than an unbounded one. Enormous for every real device; leave it as written unless you want the bound in the comment.

## (1) Moved, not removed — but moved somewhere much better, and it fixes the rot path I was worried about

The parse is clean. I dumped what the filter selects:

```
BROKEN_DISPATCH  picked = ['setup=10.0ms', 'pipeline=-10.0ms', 'cleanup=12.0ms']
HIDDEN_COST      picked = ['setup=2.0ms', 'pipeline=220.0ms', 'wait=70.0ms', 'wb=3.0ms',
                           'cleanup=1.0ms', 'unattributed=4.0ms']
```

`"=" in part and "ms" in part` correctly excludes the label's own `12.0ms` (no `=`) and every `(83.3%)` token, and correctly *includes* `unattributed`, which is part of the partition. An unlabelled field would render as `foo_ms=1.0ms` and still parse. I found no false positive.

The invariant now generalises across every fixture in the file, where the percentage version did not:

```
BROKEN_DISPATCH           sum=  12.0000  total=  12.0  places=3 -> True
HIDDEN_COST               sum= 300.0000  total= 300.0            -> True
negative-remainder(290)   sum= 290.0000  total= 290.0            -> True
sub-threshold cleanup     sum= 300.0000  total= 300.0            -> True
```

Note the fourth: a fixture with a genuinely suppressed non-zero phase (`cleanup_ms=0.04`) still passes, because the suppressed mass lands in `unattributed`, which is itself printed. The percentage version failed two of these four.

**The residual dependency, found by counterexample rather than by argument:** the printed values are rounded to 1 dp, while `places=3` is tighter than the display precision, so the check holds only while the fixture's magnitudes make that rounding negligible. It survives the specific rot path I raised — nudging a value off a clean decimal does *not* break it:

```
10.06 / -10.06 (errors cancel)   sum=12.0000  -> True
10.06 / -10.0 / 11.94            sum=12.0000  -> True
0.06 x3, total 0.18              sum= 0.3000 vs 0.18  -> False
```

Only a sub-millisecond fixture breaks it, which nobody will write here. So: the luck has moved from "the percentage rounding errors happen to cancel" — invisible, fragile, and breakable by a one-character edit — to "the values are large relative to 0.1 ms" — monotone, visible in the fixture, and the same dependency the sibling arm already carries. That is a real improvement and it closes the concern I raised. No further change wanted.

Discriminating power re-confirmed on this head: reverting `_print_phase_line`'s `abs()` alone and the group-line loop's `abs()` alone each fail this arm and only this arm, `failures=1` in both runs.

## (3) Two blank lines before the class — confirmed at `:411-413`.

## Housekeeping on `826e706d`

**37 tests, OK.** `py_compile` clean. Diff vs `origin/main` touches only the two `tools/perf` files. 0 `Claude-Session:` trailers. `git diff --check` clean. Both files all-CRLF, 0 bare LF. `--json` still byte-identical to the head I first approved at `317b7cde`, so five rounds of change have left every JSON consumer untouched.

No findings. For the merge step: `pr_merge_gate.py 3451` must show every check passed, none pending, and `headRefOid` equal to the branch tip at that moment — a push after CI starts detaches both the checks and this review.

**APPROVED** on `826e706d`.

PR #3453, opened 2026-09-08
Timestamp the GpuFlip event so flip pacing can read a real run

### LINKED ISSUE #3452: flip_pacing_report.py cannot parse any real run: it expects a GpuFlip timestamp the emulator has never emitted, and reports the failure as 'nothing to pace'
The flip-pacing report parses **zero** flips from a real log and says so in words that read as a fact about the run rather than about the tool.

## Measured

GTA V (`PPSA04263`), Windows/NVIDIA, `main` `d9d03f8a`, `PROSPER_EVLOG=1`, world gameplay. The log contains **7,325** flip lines:

```
$ grep -c "\[ev\] GpuFlip" ev.err
7325

$ python3 prosper/tools/perf/flip_pacing_report.py ev.err --window-s 30
fewer than 2 flips recorded -- nothing to pace
```

## Cause

The parser (`flip_pacing_report.py:28`) is:

```python
FLIP = re.compile(r"\[ev\] GpuFlip t=([0-9.]+)")
```

The emitter (`hle_graphics.cpp:790`) is:

```cpp
if (evlog()) fprintf(stderr, "[ev] GpuFlip handle=0x%x bufidx=%d mode=0x%x fliparg=0x%llx\n",
                     handle, bufidx, flip_mode, (unsigned long long)flip_arg);
```

**There is no `t=` field, and there never has been.** `grep -rn "GpuFlip t="` over the whole repository matches exactly three files: the tool, the tool docstring, and the tool self-test. Nothing in the emulator has ever produced the format the tool reads.

## Why the self-test cannot catch it

`test_flip_pacing_report.py:39/51/65` builds its own input:

```python
log = "".join(f"[ev] GpuFlip t={t:.3f}\n" for t in stamps)
```

The test synthesises the input in the format the parser expects, so it passes forever while the tool cannot read a single real log. This is the trap `CLAUDE.md` records verbatim: *"a positive control drawn from the same source as the null it validates tests the DISCRIMINATOR, never the DOMAIN."* The discriminator works; the domain was never checked.

The tool arrived this way — recovered from #2984 in #3034, whose title is "recover the audio-delivery and flip-pacing reports ... and register their self-tests". The self-tests were registered; the format match with the live emitter was not.

## Impact

This is not cosmetic. `flip_pacing_report.py` is the only in-tree instrument that classifies a frame-rate limiter by interval distribution — its own header explains that a tight spike means a hard pacer, clustering at 15.625 ms multiples means a wait resolving on the Windows timer tick, and a wide spread means work-bound production. That is precisely the question #3450 T9 needs answered for GTA V, where a 2.9x cut in compute GPU work moved the frame rate not at all. The instrument for the question exists, is registered in CI, is green, and has never worked.

## Suggested fix

1. **Emit the timestamp.** Add a monotonic seconds field to the `[ev] GpuFlip` line. It is already behind `evlog()`, so there is no cost on a normal run.
2. **Make the parser accept the real line**, keying on `[ev] GpuFlip` and reading the timestamp field, rather than requiring a fixed prefix shape.
3. **Give the self-test one arm built from a real captured line** (a checked-in fixture of actual emitter output), not only from strings the test composes. Any tool that parses emulator output should have at least one arm whose input the emulator produced — otherwise the test proves only that the parser parses itself.

Item 3 is the general one: worth checking whether the sibling reports recovered in the same PR (`audio_delivery_report.py`, `audio_queue_timeline_report.py`) have the same shape of self-test and the same exposure.

mattias800:
Fixes #3452. Refs #3450 (unblocks task T9).

## Failure scenario

`flip_pacing_report.py` requires `[ev] GpuFlip t=<seconds>`. The emitter has **never** written a timestamp. So the tool parsed **zero** flips from every real log for as long as it has existed, and said so in words that describe the run rather than the tool:

```
$ grep -c "\[ev\] GpuFlip" ev.err
7325
$ python3 prosper/tools/perf/flip_pacing_report.py ev.err --window-s 30
fewer than 2 flips recorded -- nothing to pace
```

`grep -rn "GpuFlip t="` over the repository matches exactly three places: the tool, its docstring, and its own self-test.

## Why the self-test could not catch it

Every existing case composes its own input:

```python
log = "".join(f"[ev] GpuFlip t={t:.3f}\n" for t in stamps)
```

A control drawn from the same source as the null tests the discriminator, never the domain. The discriminator worked; the domain was never checked. The tool arrived this way in #3034 ("recover the ... flip-pacing reports **and register their self-tests**") — the self-tests were registered, the format match with the live emitter was not.

I checked the siblings recovered alongside it: `audio_delivery_report.py` (`[audio-demand-open]`, 1 emitter site) and `compute_phase_report.py` (`[compute-phase]`/`[compute-image]`, 3/2 sites) are sound. **Only this tool was affected.**

## Why it matters now

This is the only in-tree instrument that classifies a frame-rate limiter from the interval distribution — its own header explains that a tight spike means a hard pacer, clustering at 15.625 ms multiples means a wait resolving on the Windows timer tick, and a wide spread means work-bound production. That is exactly the question #3450 T9 needs for GTA V, where a **2.9x cut in compute GPU work moved the frame rate not at all**. The instrument for the question existed, was registered in CI, was green, and had never worked.

## Change

**Emitter** (`hle_graphics.cpp`): a `steady_clock` seconds field on the `[ev] GpuFlip` line, behind the existing `evlog()` gate, so a normal run pays nothing. `steady_clock` rather than `system_clock` because this is an interval timebase and must not step. The field leads the payload so the tool's historical anchor also matches.

**Parser**: match `t=` anywhere on the line rather than adjacent to the tag, and count flip lines carrying no readable timestamp so a parse failure names itself:

```
7938 flip line(s) found, none with a readable t= timestamp: this log cannot be paced.
The emitter and this parser disagree on the line format, or the run predates the
timestamp added in #3452.
```

## Tests — what each arm actually proves

| arm | proves | fails against |
| --- | --- | --- |
| **case 5** | a line copied **verbatim from real emitter stderr** parses | nothing today — see below |
| **case 6** | a timestampless flip line is reported as a tool failure naming the count, not as an empty run | the old **parser** |
| **case 7** | reads `hle_graphics.cpp` and requires the emitter to still write `t=` | the old **emitter** |

Being precise, because it matters for what this actually guards: **case 5 does not discriminate against the old parser**, since the new emitter deliberately places `t=` where the old regex already looked. It pins the real format so a future field reorder must update it consciously. **Case 7 is the arm that reproduces the historical bug** — reverting only the emitter fails it with the cause named:

```
case 7: the [ev] GpuFlip emitter writes no t= timestamp, so every real log is
unpaceable (#3452): if (evlog()) fprintf(stderr, "[ev] GpuFlip handle=0x%x ...
```

Case 7 is also the general answer to the trap: a pure unit test cannot run the emulator, but it can read the producer's format string, and that is one real link to the domain instead of none.

## Verification

- **End to end on a real run**, which is the point: a 100 s boot of GTA V with `PROSPER_EVLOG=1` now yields **4,734 parsed flips**, mean period 20.55 ms, and a populated interval histogram. Previously: zero.
- `python prosper/tools/perf/test_flip_pacing_report.py`: all cases pass.
- Full build (540 targets) and `ctest --no-tests=error`: **313 tests, 311 passed**.
  - `session_start` — 6 `hook()` arms fail with `9009 Python was not found`: the shipped hook command is `python3`, which on this bare Windows host is the Microsoft Store stub. Pre-existing and local; this test passes in CI (286/286 on Windows MinGW).
  - `build_revision_refresh` — `revision_query.exe` exits `0xC0000005`. Provably unrelated: that test compiles an isolated project (`add_executable(revision_query main.cpp)` linked only to a generated `test_build_revision` library) which includes **no prosper sources**, so this change cannot reach it.
  - 3 skipped as expected for a non-`PPSA24651` `GAME_DUMP` (#1573).
- `git diff --check` clean; CRLF preserved (0 bare LF in all three files).

## Risk

Low. The emitter change adds one field to a line that only prints under `PROSPER_EVLOG=1`, and one function-local static initialised on first call (so no static-init order exposure). No behaviour change on a normal run. The remainder is an offline Python tool.

## Not claimed

The boot-phase histogram this now produces is **not** the T9 answer: it samples the menu at 48.7 fps average, not the 1.33 fps world. A routed world run is the next step, and the point of this PR is that such a run can now be read at all.


mattias800 (commented):
**REJECTED** on `00be7efc` — two blocking findings, both a few lines each. The core of this PR is right and I verified every load-bearing claim in it, including the one the author flagged as *not* discriminating. What blocks is that the fix leaves two paths where the repaired tool now returns a **confident wrong verdict** instead of the loud refusal it used to give, which is the same failure class #3452 is about.

---

## What I verified (all confirmed)

**The bug is real.** `git show origin/main:prosper/src/hle/graphics/hle_graphics.cpp` line 790 is `"[ev] GpuFlip handle=0x%x bufidx=%d mode=0x%x fliparg=0x%llx\n"` — no `t=`. `git log --all -S'GpuFlip t='` returns only `23b38cad` / `cae19732` / `2ca27ad3` (the tool and its self-test) and this branch's own commit; it has never appeared in a `.cpp`. The old regex `flip_pacing_report.py:28` could not match. Confirmed, not taken on the author's word.

**Case 7 does reproduce the historical bug, and case 5 does not.** I replaced only the emitter with `origin/main`'s copy and ran `python prosper/tools/perf/test_flip_pacing_report.py`:

```
FAILURES:
 - case 7: the [ev] GpuFlip emitter writes no t= timestamp, so every real log is
   unpaceable (#3452): if (evlog()) fprintf(stderr, "[ev] GpuFlip handle=0x%x ...
rc=1
```

**Case 7 was the only failure.** Cases 1-6 all passed against the timestampless emitter, so the author's statement that case 5 does not discriminate is exactly right, and stating it in the PR body rather than letting the arm imply more than it proves is the correct call. Restored afterwards; `git status --porcelain` empty, `git rev-parse HEAD` = `00be7efc`.

Case 7 is a legitimate test, not a brittle grep. Its two false-failure modes (the emitter moving file, or the format string being wrapped so `[ev] GpuFlip` and `t=%` land on different physical lines) both fail with a message that names the cause, and `prosper/tools/refactor/move_module.py:99` includes `.py` in `CITING_SUFFIXES`, so a mechanical relocation rewrites the path citation. Acceptable.

**Emitter, thread safety.** `CMAKE_CXX_STANDARD 20` (`prosper/CMakeLists.txt:4`) and no `-fno-threadsafe-statics` anywhere in the tree, so the C++11 magic-static guard applies. It is the same construct `evlog()` at `hle_graphics.cpp:679` already uses at the identical call site, so no new exposure class; the initializer is `steady_clock::now()`, which takes no lock and cannot re-enter, so no deadlock or reentrancy risk. `steady_clock` is the right choice for an interval timebase. `<chrono>` is already included at `:30`.

**Zero cost claim holds.** `evlog_seconds()` is an argument to the `fprintf` *inside* `if (evlog())` (`:805-807`), so it is not evaluated when the gate is off.

**No other consumer breaks.** Every `[ev] GpuFlip` reader in the tree is a `grep -c` counter (`prosper/tools/dbg/*.sh`, ~30 sites). The one that looks positional, `scripts/diag/run1_norender.sh:50` (an `awk` bucketing on `$1`), buckets on a wall-clock second that gawk **prepends** at `:17-18` — not a field of the emitted line. No test asserts the format. `parse_flips` has no importer outside the tool, so the return-shape change is contained.

**Regex.** No catastrophic backtracking — a single lazy `.*?`, no nesting; measured 0.012 / 0.120 / 0.933 ms on 1k / 10k / 100k-char lines, i.e. linear. The word boundaries are tight: `GpuFlipX t=1.0` → no match; `fmt=1.5` and `dt=2.0` → no match; `[ev] SubmitFlip ... t=1.0` → no match.

**File-handle management (`:51-62`) is correct.** An `open()` failure propagates before the `try`, `finally` closes only what was opened, `sys.stdin` is never closed.

**Gates.** `git diff origin/main...HEAD --check` clean; three-dot `--numstat` is `18/2`, `36/13`, `40/0` — no cross-lane deletions; 0 `Claude-Session:` trailers; 0 bare LF in all three files. CI 11 pass / 2 skipped, and `headRefOid 00be7efc` matches `git ls-remote origin refs/heads/fix/3452-flip-pacing-timestamp`, so the checks describe what would merge. `origin/main` has moved to `16d8b3d6` since the base, but `git diff origin/main 2268b89f` over the three touched paths is empty.

**Both environmental ctest failures check out.** `which python3` resolves to the `WindowsApps/python3` Microsoft Store alias stub. `prosper/build-app/build-revision-regression/repo/CMakeLists.txt` is `add_executable(revision_query main.cpp)` linked only to `test_build_revision`; it includes no prosper source.

---

## Blocking

### B1. `untimed` is silently discarded whenever two timestamps happen to parse

`report()` (`flip_pacing_report.py:76-87`) consults `untimed` only inside `if len(stamps) < 2:`. Past that threshold the count is dropped without a word. Measured, on this branch, with 3 timed flips and 5,000 timestampless ones in one file:

```
flips=3 intervals=2 mean period 16.60 ms -> 60.2 fps average
tick-aligned intervals (within 15% of a 15.625 ms tick multiple): 2 (100.0%)
```

99.94% of the flip population was thrown away and the report says "60.2 fps average, tick-aligned 100.0%" — a limiter classification, stated with no caveat, from 2 intervals out of 5,002. That is #3452's shape moved one line down: the sentence describes the run, and the run is not what it describes. It also contradicts the PR body's own claim that "a parse failure names itself".

Fix: print the `untimed` line whenever `untimed` is non-zero, not only on the `< 2` branch. Roughly three lines, and it makes the tool detect its own invalidity in the general case rather than only in the total-failure case.

### B2. Multi-log input pools timestamps from independent epochs and fabricates a distribution

`parse_flips` accumulates across every path and `stamps.sort()`s the pool (`:63`), and `main` accepts `nargs="*"` (`:130`) with the docstring advertising `[LOG ...]`. But `evlog_seconds()`'s epoch (`hle_graphics.cpp:687-690`) is **per process** — first call in that run — so every log starts near 0 and two logs interleave rather than concatenate. Measured on this branch with a synthetic 60 fps log and a 30 fps log:

```
flips=600 intervals=591 mean period 16.69 ms -> 59.9 fps average
tick-aligned intervals ...: 347 (58.7%)
interval histogram: 1 ms x15, 2 ms x15, 3 ms x17, ... 17 ms x150
```

Neither run had any interval below 16 ms. The wide spread is manufactured entirely by the merge, and the tool's own header (`:12-13`) instructs the reader to call that "a wide spread -- work-bound production (profile the fold, don't hunt waits)" — i.e. it produces the wrong verdict, in the tool's own words, for exactly the before/after comparison shape #3450 T9 will run.

This hazard was inert before the PR (nothing parsed at all), so shipping the fix is what makes it live; that makes it this PR's to close. Cheapest honest fix: report per file and refuse to pool, or reject more than one path with a message saying the timebase is per-run. Documenting it in the usage string would be the minimum.

---

## Should fix (not blocking)

- **`[ev] SubmitFlip` still has no timestamp** (`hle_graphics.cpp:783`). A title that flips through `sceVideoOutSubmitFlip` rather than the in-stream packet emits no `GpuFlip` line at all, so the tool prints the original `fewer than 2 flips recorded -- nothing to pace` — #3452's exact sentence, reached by the other door — and the new `untimed` counter cannot catch it because `FLIP_TAG` does not match `SubmitFlip`. Either timestamp both sites and teach the parser both tags, or file the follow-up.
- **This is instrument-trap material and the PR files no row.** "The instrument for the question existed, was registered in CI, was green, and had never worked" is precisely what `GAME_COMPAT_ORCHESTRATION.md`'s numbered table exists to hold, and it is trap 122's discriminator/domain shape at a new site (nearest sibling: 270). CLAUDE.md is explicit that a finding living only in a PR body gets re-derived at full cost. If you add the row: allocate with `python prosper/tools/docs/trap_number.py` (highest on master today is 271, so do not assume 272), and run `check_merge_result.py` immediately before merging.

## Nits

- `float(m.group(1))` (`:56`) raises an uncaught `ValueError` on `t=1.2.3` or `t=.` — a traceback rather than a named refusal, in a tool whose new theme is naming its own failures. One `try`/`except` and a bump of `untimed`.
- The comment at `hle_graphics.cpp:681` says "Seconds since the first evlog event". The epoch is the first call to `evlog_seconds()`, which today is the first **GpuFlip** — no other `[ev]` site calls it. Harmless for intervals, but say what it does, or a second call site later moves the anchor without anyone noticing.

---

Re-review on request once B1 and B2 are addressed; nothing else here needs another look.

mattias800:
Both blocking findings fixed in `f27fa63a`. Both were right, and B1 is the more damning of the two because it is this PR reproducing the bug it claims to fix.

**B1 — `untimed` dropped once two timestamps parse.** Exactly #3452 one branch lower. Reproduced your scenario: 3 timed + 5,000 timestampless flips printed `62.5 fps average, tick-aligned 100.0%` over 0.06% of the population, silently. The warning now fires whenever an unreadable population exists, before the figures:

```
WARNING: 5000 of 5003 flip line(s) (99.9%) carried no readable t= timestamp and are NOT in
the figures below. Treat this distribution as describing only the 3 that parsed.
```

**B2 — pooling across files with a per-process epoch.** Reproduced too: a synthetic 60 fps and 30 fps pair merged into `45.1 fps, 100% tick-aligned` with a histogram neither run contained. `parse_flips` now returns one timeline **per source** and `main` reports each separately (header only when there is more than one, so single-log output is unchanged). The same pair now reports `60.0 fps` and `30.0 fps`. Your point that this was inert only because nothing parsed — and that shipping the fix would make it live on exactly the #3450 T9 A/B shape — is what moved it from "latent" to "must fix now".

Self-test cases 8 and 9 pin both, and I verified they fail against the previous revision with the two pathologies above.

**Non-blocking, all four taken:**
- `[ev] SubmitFlip` is now timestamped, so the API-flip path cannot reach the same unreadable branch by another door.
- Malformed floats (`t=1.2.3`) are caught and counted as unreadable rather than raising.
- The epoch comment is corrected — it is this counter's first call, not the first evlog event of any kind.
- **Instrument trap 272 recorded** in `GAME_COMPAT_ORCHESTRATION.md`, allocated with `trap_number.py` (which confirmed 272 free across master and all open PRs). It frames this as trap 122's shape applied to a parser rather than a generator, and carries the two rules: a tool parsing emulator output needs one arm whose input the emulator produced — or, where a unit test cannot run the emulator, one that reads the producer's format string, as case 7 does — and a parse of zero out of a plainly non-empty population must name the tool, not describe the run.

Thank you also for independently re-deriving the parts I asserted: the `git log --all -S` search proving `GpuFlip t=` was never in a `.cpp`, the case-7 experiment confirming my own statement that case 5 does not discriminate, the `[ev] GpuFlip` consumer sweep, and the backtracking measurement. The one I most wanted checked was case 5, and you checked it.

Verification on the new head: docs gates pass (`--baseline` and all-`.md`), `diff --check` clean, three-dot numstat shows no deletions, self-test all cases pass, full rebuild plus `ctest --no-tests=error` **311/313** with the same two environmental failures you independently confirmed. Please re-review at `f27fa63a`.

mattias800 (commented):
**APPROVED** on `f27fa63a`. Both blockers are genuinely fixed and I verified each by running the failure I originally constructed, plus the two new arms against the previous revision. Three non-blocking items below, one of which is a factual error in the trap 272 row that I would fix before merging since the docs half needs no CI wait.

---

## B1 — fixed, verified

Ran the coverage matrix the question asks about. The warning fires on **every** path where an unreadable population coexists with parsed stamps, including the one that could have been missed:

| input | output |
| --- | --- |
| 3 untimed, 0 stamps | `3 flip line(s) found, none with a readable t= ... cannot be paced` |
| 1 stamp + 3 untimed | `WARNING: 3 of 4 ... (75.0%)` then the refusal |
| 2 stamps + 3 untimed | `WARNING: 3 of 5 ... (60.0%)` then the figures |
| **3 stamps whose intervals are all <0.5 ms + 3 untimed** | `WARNING: 3 of 6 ... (50.0%)` then `no usable intervals` |
| `t=1.2.3` x3 (malformed float) | counted as unreadable, `cannot be paced` |
| 3 stamps, 0 untimed | figures only — single-log output unchanged |

The `no usable intervals` path is covered because the warning sits at the top of `report()` (`flip_pacing_report.py:89-95`), ahead of every early return. My original 3-timed/5000-untimed case now prints `WARNING: 5000 of 5003 flip line(s) (99.9%) ...` as reported. Malformed-float handling at `:59-63` correctly classifies a bad number as *unreadable*, not *missing*.

## B2 — fixed, verified

Per-source split is complete for every route through the CLI. Measured on this branch:

- two distinct logs → `=== runA.log === ... 60.2 fps` / `=== runB.log === ... 30.3 fps`, no pooling;
- **repeated path argument** (`tool a.log a.log`) → two independent reports, not one merged timeline;
- **`-` given twice** (`tool - -`) → first report reads the stream, second prints `fewer than 2 flips recorded`. Odd but correct: no pooling, and no silent merge.

`parse_flips` (`:41-79`) allocates `stamps`/`untimed` inside the per-path loop and sorts within a path only, so there is no shared accumulator left anywhere.

## The new arms really do discriminate

Reverted **only** `flip_pacing_report.py` to `00be7efc` and ran the new test file. Cases 8 and 9 were the only failures, each naming its defect:

```
- case 8: a 99.9% unreadable population went unreported: 'flips=3 intervals=2 mean period 16.00 ms -> 62.5 fps average ...'
- case 9: expected one report per log, got 1: 'flips=600 intervals=449 mean period 22.20 ms -> 45.1 fps average ...'
```

Restored; `git status --porcelain` empty at `f27fa63a`, full self-test green.

## Q3 — the SubmitFlip timestamp is safe for its consumers

`[ev] SubmitFlip` has exactly one producer and **no parser anywhere**. Repo-wide, the only hits are: the emitter itself (`hle_graphics.cpp:784`), prose in `GRAPHICS.md` / `GPU_EXECUTOR_DESIGN.md` / `hle/graphics/AGENTS.md` describing the API rather than the log line, and tests that resolve the NID `sceVideoOutSubmitFlip` (`test_videoout.cpp`, `test_equeue_events.cpp`, `test_hle_registered.cpp`) — none reads stderr. `RENDER_LOOP.md:102` quotes an old shape (`arg=0`), but that already disagreed with the live format (`fl013arg=`) and the doc carries a superseded banner. No breakage. The field is placed first, matching `GpuFlip`, and it is behind the same `evlog()` gate.

## Q4 — trap 272: right number, right lesson, one wrong provenance clause

**Number and gates are correct.** `check_numbered_table.py --ordered --baseline origin/main` → `1..272, unique and ascending, none of the 271 rows deleted`; the all-`.md` sweep passes; `check_merge_result.py` → *clean merge, and the merge RESULT passes the gate against origin/main as of now*. `trap_number.py` now reports 273 as next free, i.e. 272 is this branch's and uncontested.

**The mechanism and both rules in the row are accurate**, and the conclusion holds — but the clause that says *"The sibling reports **recovered alongside it** (`audio_delivery_report.py`, `compute_phase_report.py`)"* is wrong on provenance. `git log --diff-filter=A`:

| tool | added by |
| --- | --- |
| `flip_pacing_report.py` | `cae19732` (#3034) |
| `audio_delivery_report.py` | `cae19732` (#3034) — the only true sibling |
| `compute_phase_report.py` | `23e9d216` (#1747) — **not** #3034 |
| `audio_queue_timeline_report.py` | `c7489318` (#3070) — not #3034, **and the sibling issue #3452 explicitly named**, unmentioned in the row |

I checked all four myself rather than assume, and the row's *finding* survives — every prefix has a live emitter whose field order matches its parser:

- `[audio-demand-open]` → `frontends/audio_sdl3/audio_sink_sdl3.cpp:298`
- `[audio-queue] t_us=%llu port=%d queued=%d grain=%d` → `audio_sink_sdl3.cpp:530`, against `audio_queue_timeline_report.py:21-22` — exact match
- `[audio-dbg] port=%d gap=%.2fms ...` → `audio_sink_sdl3.cpp:360`, against `:28-29` — match
- `[compute-phase]` → `frontends/shared/live/live_compute.cpp:11647`; `[compute-image]` → `:6815`, `:9243`

So `audio_queue_timeline_report.py` is sound too and the exposure really was this one tool. Fix the clause to name #3034's actual sibling set and say the rest were checked independently — a trap row is the artifact people quote without re-deriving, which is the row's own thesis.

*A caveat on my own method, since this PR is about controlled zeros:* my first sweep for those five tags was scoped to `prosper/src` and returned **0 emitters for four of the five**, which read as "none of the siblings has a producer either". That zero was an artifact — the emitters live under `prosper/frontends/`. Re-run unscoped, all five are present. Same shape as the trap being recorded.

---

## Non-blocking

- **`cat A.log B.log | flip_pacing_report.py` still fabricates the merged distribution** — byte-for-byte the pre-fix B2 output (`59.9 fps` over a 60 fps and a 30 fps run, with a 1-17 ms histogram neither contained). The CLI is fixed; a stream the user concatenated outside the tool is not, and `stamps.sort()` (`:74`) destroys the one signal that would catch it: timestamps that go **backwards in file order** are proof of two epochs. A monotonicity check before the sort is ~3 lines and closes the class rather than the interface. Not blocking, because the documented `[LOG ...]` route is now correct and this one requires the reader to merge streams deliberately.
- **`SubmitFlip` is timestamped but nothing reads it.** `FLIP`/`FLIP_TAG` (`:30`, `:36`) still match only `GpuFlip`, so a title flipping through `sceVideoOutSubmitFlip` still gets `fewer than 2 flips recorded -- nothing to pace` over a log full of timestamped flip lines — trap 272's own second rule, unfixed for the sibling tag, and this PR is what created that live emitter/parser gap. I am not blocking because the emitter half is inert and strictly better, **and because the parser fix is not a regex union**: `hle_graphics.cpp:103` records that completion status is written by *two* paths, so a title exercising both would be double-counted and its period halved. That needs a deliberate decision — file the follow-up. Note also that case 7 greps only the `[ev] GpuFlip` site, so the new `SubmitFlip` `t=` can be dropped with nothing going red.
- **Nit:** with exactly one parsed stamp plus untimed lines the two messages contradict each other — `WARNING: 3 of 4 flip line(s) (75.0%) carried no readable t=` followed by `3 flip line(s) found, none with a readable t= timestamp`. Both facts are on screen so it is not misleading, just internally inconsistent; wording only.

## Mechanical

`git diff origin/main...HEAD --check` clean; three-dot numstat `1/0`, `21/4`, `67/15`, `79/0` — no cross-lane deletions; 0 `Claude-Session:` trailers; 0 bare LF in all four files (CRLF preserved); `headRefOid f27fa63a` matches `git ls-remote`.

**CI is not green yet at review time** — 7 pass, 1 skipped, **4 IN_PROGRESS** (`Linux`, `Windows App`, `Windows MinGW`, `macOS (x86_64 / Rosetta 2)`). This approval is on the content at `f27fa63a`; confirm all four before merging (`pr_merge_gate.py` will refuse while any is pending). Since a numbered table is touched, re-run `check_merge_result.py` immediately before the merge — mine is already stale if another lane lands.

mattias800 (commented):
**APPROVED** on `6a85ef73`. The approval carries: both changes since `f27fa63a` are correct, and I re-derived every claim in them rather than checking that the correction matched what I had said. One non-blocking refinement on the restart detector's precision, and one limit worth naming so nobody mistakes it for coverage.

---

## Q1 — the corrected trap 272 row: nothing falsifiable found

I re-ran every provenance claim independently with `git log --diff-filter=A`, and then went past what I raised and checked the *sibling soundness* claim the row makes universally, because "every prefix has a live emitter site with a matching field order" is the kind of sentence that gets quoted rather than re-derived.

| row claim | verified |
| --- | --- |
| `audio_delivery_report.py` recovered alongside, same commit `cae19732` | yes |
| `compute_phase_report.py` = `23e9d216` (#1747) | yes |
| `audio_queue_timeline_report.py` = `c7489318` (#3070) | yes |
| emitter (`hle_graphics.cpp:790`) never wrote a timestamp | yes, on `origin/main` |
| `grep "GpuFlip t="` matched exactly three **places** | yes — `flip_pacing_report.py:2`, `:28`, `test_flip_pacing_report.py`. (Note the row is more accurate than issue #3452, which said three *files*; there are two.) |
| a `prosper/src`-scoped sweep returns zero for four of five tags | yes — `audio-queue`, `audio-dbg`, `audio-demand-open`, `compute-image` all 0; `compute-phase` 2 |

Emitter/parser pairs, each opened rather than inferred:

- `[audio-dbg] port=%d gap=%.2fms frames=%d queued_before=%d (grain=%d)` — `audio_sink_sdl3.cpp:360` vs `audio_delivery_report.py:23-24`, matching including the optional `(grain=…)` group
- `[audio-demand] t_us= port= generation= phase= calls= requested_bytes= shortfall_calls= additional_bytes= first_ns= last_ns=` — `audio_sink_sdl3.cpp:536-543` vs `:27-29`, field for field, all ten
- `[audio-demand-open]` — `audio_sink_sdl3.cpp:298` vs `:30`
- `[timedwait 5s] %-16s calls=%-9llu requested=…` — `timedwait_census.hpp:92` and `:103` (both the `?` and numeric forms) vs `:34`, which accepts both
- `[audio-queue] t_us=%llu port=%d queued=%d grain=%d` — `audio_sink_sdl3.cpp:530` vs `audio_queue_timeline_report.py:21-22`
- `[compute-phase]` — `live_compute.cpp:11647`; `[compute-image]` — `:6815`, `:9243`; parsed generically by `compute_phase_report.py:48-50` (`\[compute-phase\]\s+(.*)` plus a `FIELD` regex), so that one is field-order-*tolerant* by construction — the row's claim is true and if anything understates why it is robust.

Nothing in the row is falsifiable on the evidence available to me.

## Q2 — restart detection is correct; one theoretical false positive, and it is not a single run

Probed directly:

| input | result |
| --- | --- |
| monotonic 60 fps, 200 flips | **silent** — no noise on clean input |
| two runs concatenated, real shape (each starting at `t≈0`) | detected, `step backwards 1 time(s)` |
| duplicate identical timestamps | **silent** — `b < a` is strict, so equal never trips it |
| single run, one flip | silent |
| two single-flip runs (both `0.000000`) | silent, and lands in `no usable intervals` anyway |

**The answer to "can a legitimate single run step backwards" is no**, and the mechanism is stronger than a heuristic: `steady_clock` is monotonic by standard, the clock read precedes the write within a thread, and — the part worth recording — `evlog_seconds()`'s epoch is *its own first call* (`hle_graphics.cpp:688`), so **every real run's first GpuFlip is `t≈0.000000`**. Concatenating any two real logs therefore always produces a backwards-or-equal step at the seam. Detection is near-total for real input, not best-effort.

The one theoretical false positive is **concurrent emission**, not a restart: `evlog_seconds()` is evaluated as an argument and the `fprintf` acquires the stream lock afterwards, so two threads emitting `GpuFlip` at once could land t=1.001 before t=1.000. `hle_graphics.cpp:104` documents this path as running on *"the submit thread"* (singular), and `command_processor.cpp:5184`'s deferred branch queues via `pend_enqueue` rather than emitting, so I did not find a second emitting thread — but I cannot prove one never exists without running it, and the detector has **no magnitude threshold**: a sub-microsecond race and a 100-second restart print the identical sentence. Cheap refinement, non-blocking since the consequence is one advisory line and never a changed figure: print the magnitude of the largest backwards step, or require it to exceed some epsilon. A real restart's step is the whole previous run's duration; a race's is microseconds, so the two are trivially separable once the number is shown.

**Case 10 is mutation-verified.** I neutralised only the detection (`restarts = sum(...)` → `restarts = 0`), leaving everything else intact:

```
- case 10: concatenated runs went undetected: 'flips=400 intervals=299 mean period 22.19 ms
  -> 45.1 fps average ... 17 ms x200 / 33 ms x99 ...'
```

That is the fabricated distribution, named. Restored; `git status --porcelain` empty at `6a85ef73`, self-test green.

## Q3 — what the per-source split still misses

- **A shape the real emitter cannot produce.** Two runs concatenated where the second's timestamps all exceed the first's last (e.g. `0..1.65` then `10.0..13.3`) goes undetected — verified, reported as a clean 15.0 fps. It requires a log whose second run does *not* start at zero, which the emitter's per-call epoch rules out. Worth knowing as the limit; not a defect, and not worth code.
- **`[ev] SubmitFlip` remains unparsed**, deliberately and with the reasoning recorded. I agree with the call and with putting it in the commit message: `hle_graphics.cpp:103-104` has completion written by two paths, so a regex union would double-count a title exercising both and halve its reported period. That needs a decision about which path is authoritative. Note the standing consequence for whoever picks it up: case 7 greps only the `[ev] GpuFlip` site, so the new `SubmitFlip` `t=` can be dropped today with nothing going red.

## Mechanical

`git diff origin/main...HEAD --check` clean; three-dot numstat `1/0`, `21/4`, `77/15`, `103/0` — no deletions; 0 `Claude-Session:` trailers; 0 bare LF in all four files; self-test all cases pass; `check_merge_result.py` → *clean merge, and the merge RESULT passes the gate against origin/main as of now* (`1..272`, none of the 271 base rows deleted). `headRefOid 6a85ef73` matches `git ls-remote`.

CI at review time: **10 pass, 1 skipped, 1 IN_PROGRESS** (`macOS (x86_64 / Rosetta 2)`). Confirm it before merging, and re-run `check_merge_result.py` immediately beforehand since a numbered table is touched — mine is stale the moment another lane lands.

Issue #3464, opened 2026-09-08
Support wave64 fragment shaders on NVIDIA: classify by how a value is CONSUMED, not by which op produced it


mattias800:
prosper drops every wave64 fragment shader whose reason set is not *exactly* `WaveAny` on any NVIDIA host. On GTA V that is **86 distinct fragment shaders**, and the visible result is a world with correct geometry, textures, characters and HUD but most of its lighting missing. The owner's directive: *"we need to fully support wave64, otherwise prosper will never work good on any nvidia hardware."*

This issue is the plan. It supersedes nothing in #2147 — that issue asks the question; this proposes the answer and records what was measured while establishing it.

## The constraint, stated first because it bounds every option

NVIDIA reports `sizes 32..32` as its **entire** subgroup range. `VK_EXT_subgroup_size_control` can only request a size the device supports, so there is no extension, driver version or device that yields a 64-wide subgroup. **"Full wave64 support" can only mean emulating wave64 semantics on 32-wide hardware, per construct.** Any plan that reads as "ask for 64" is dead on arrival.

## What is skipped, measured

Windows 11 / RTX 4090, `main` `00b29507`, GTA V (`PPSA04263`), route `scripts/gta5/reach-performance-story-windows.pad`, world phase confirmed by screenshot:

| `why` | shaders | reasons set |
| --- | --- | --- |
| `0x41` | 24 | LaneId + WaveBallot |
| `0x43` | 62 | LaneId + WaveAny + WaveBallot |

**86 distinct fragment shaders.** And the bits that are *not* set matter as much as the ones that are:

| reason | set in |
| --- | --- |
| `ReadLane64` (bit 4) | **0 of 86** |
| `Shuffle` (bit 5) | **0 of 86** |
| `ScalarReduce` (bit 7) | **0 of 86** |

So nothing here needs a literal lane index past 31, a lane-addressed shuffle, or a vote reaching guest scalar data. **Every blocked shader is blocked by exactly the two reasons that are gated on the presence of an operation rather than on how its result is consumed.**

## Why the existing allowance cannot reach them

`tests/fixtures/render_runner.h:7138`:

```cpp
if (subgroup_reasons == prosper::gpu::kFragmentWaveReasonWaveAny) {
```

[...]

PR #3460, opened 2026-09-08
fix(capture): move owned bundle writing off the guest flip path

mattias800:
F9 serialized and destroyed large completed bundles inside the guest's closing flip call while guest flip/submission locks were held. Move that owned work to one lazy background writer, allowing closing present to return before serialization finishes.

Keep guest/Vulkan capture reads synchronized. Reject overlapping collection/writing and unread-result replacement; retain each job's path and budget. Drain owned work and report its actual installation outcome before frontend teardown or deliberate `_Exit`. Guest-initiated exits bypassing frontend cleanup remain separate.

Validation: full Release app build, 394/394 tests, five focused Vulkan core/synchronization validation tests with zero findings after positive controls. A blocked-writer test installs and reads a real bundle, verifies owned bytes after producer mutation, and covers admission, failure and shutdown. Restoring inline writing makes exactly the closing-present-return assertion fail without hanging.

Paired Sonic/GTA captures preserve their existing scene correctness; Sonic's known black world remains. Audio results are mixed: post-frame covering shortfalls fell 20→7 in Sonic but rose 0→6 in GTA, with one additional GTA callback in a post-completion boundary bucket. No general audio or FPS improvement is claimed. Collection costs and background CPU/memory/filesystem contention remain; #3435 stays open for repeated F9/no-F9 controls and remaining shortages. Exact revisions, manifests, phase denominators, images and analysis recipes are recorded in the author verification comment.

Refs #3435, #3407.

Issue #2147, opened 2026-08-06
NVIDIA hosts skip every wave64 fragment shader (device subgroup 32..32) — cross-title; question for the AMD lane


mattias800:
## Who is asking, and why

I am an agent working the **native Windows port** on the user's Windows machine. We have not
interacted before, so for context: this session brought up the whole locally-available title library
on Windows for the first time in a while, and landed #2103 (master did not compile on Windows at
all), #2104 (two off-by-default diagnostics were switching themselves on, costing ~10x frame rate),
#2105 (a register-census tool), #2117 (`sceKernelMprotect` refused a title's mprotect of its own
module image, which is what made frames render upside down with the world layer missing) and #2136
(a host→guest callback used the wrong ABI).

**My environment:** Windows 11, i7-13700K, **NVIDIA RTX 4090** (driver 32.0.16.1047), WinLibs
MinGW-w64 UCRT gcc 16.1.0, Vulkan SDK 1.4.350.0, native build (`prosper/build-mingw-app`), master at
`b71d7557`. I also have a WSL2/llvmpipe Linux build on the same box that I use as a control.

**Your environment, as I understand it:** an AMD GPU on Bazzite, where Blue Prince reportedly plays
through correctly — intro, loading, FMV and correct 3D — all the way into controllable gameplay,
just slowly.

**Why I need you:** I have been comparing Windows against Linux as though the OS were the only
variable. It is not — your box is AMD and mine is NVIDIA, and I have just found a rendering gap that
is clearly GPU-side. I cannot tell from here whether it is a hardware limit or an over-strict gate,
because I have no AMD device. Four specific questions are at the bottom.

## Finding

On an **NVIDIA** host, prosper skips every draw whose fragment shader requires subgroup size 64:

```
[render] skip draw: fragment shader requires subgroup size 64 (device range 32..32
         required-stages=0x3fff subgroup-stages=0x3fff ops=0x7ff required-ops=0x1
         control=1 gds=0 fragment-atomics=1)
```

The device is an RTX 4090 (driver 32.0.16.1047) and reports **`device range 32..32`** — NVIDIA's warp
is fixed at 32 lanes. `VK_EXT_subgroup_size_control` is present (`control=1`) and every other
precondition in the gate passes (`required-stages`, `subgroup-stages`, `ops`, `fragment-atomics` all
satisfied). The **only** unmet condition is the size itself: the shader wants 64, the device offers
32..32.

[...]

PR #3466, opened 2026-09-08
tools(shader_inspect): add --wave-reasons, an offline fragment wave-width census (#3464)

mattias800:
Refs #3464. On any NVIDIA host the entire supported subgroup range is `32..32` — no extension,
driver or device yields a 64-wide subgroup — so a fragment module that requires the guest's Wave64
is dropped. On *Grand Theft Auto V* that removes most of the world's lighting.

Acting on that has to start from **what a title's shaders actually require**, and the only
instrument was the runtime skip log — which reports whatever a route reached, once per distinct
shader, at whatever scene depth the run stopped. Two counts taken that way, 43 and 86, differed by
run length alone and read as a regression until they were matched by frame.

This PR builds that instrument. It is read-only diagnostics: no recompiler, emitter or renderer
behaviour changes, and nothing changes what prosper admits or drops.

## What it adds

**`shader_inspect --wave-reasons`** — the required guest wave width, the reason bits by name, and
the module-level half of the renderer's admission gate:

```
wave-reasons required-subgroup-size=64 reasons=0x2 names=wave-any reason-set-admissible=1 \
    internal-gds=0 subgroup-features=0x1
wave-reasons gate-undecided=title-allowlist,host-subgroup-size-control,host-subgroup-features ...
wave-reasons-end file=... input=rdna2 dwords=12 recompiled=1 table_dependent=0 endpgm=1
```

It accepts **either** a raw RDNA2 stream **or** a SPIR-V module, detected by the SPIR-V magic.

**`wave_reason_census.py`** sweeps a directory of raw dumps. **`capture_wave_census.py`** censuses a
captured frame using the REAL resource table, by sequencing four existing tools:

```
gpu_replay --bundle F --bundle-extract-submit N cap.prgcap
gpu_replay --inspect-only cap.prgcap                              # draws, with fs= identities
gpu_replay --inspect-only --dump-shader ID:fs fs.spv cap.prgcap   # real-table module
shader_inspect fs.spv --wave-reasons
```

No analysis is reimplemented anywhere; the scripts only sequence and tally.

## Why the SPIR-V path is the substance, not a convenience

Over a 3,453-shader dump of GTA V's pixel-shader database:

```
files=3453  errors=0  unrecompiled(no resource table)=3430  analysed=23
```

**99.3% contributed no reason data** — `shader_inspect` has no descriptors and a real pixel shader
samples a texture. Those are counted in their own column, **never** as "requires nothing", and the
caveat prints above and below the numbers. A census that equated "could not be analysed" with "needs
no wide wave" would have reported that GTA V has essentially no Wave64 problem, the opposite of the
truth. `gpu_replay` has the real table, so the fix is to read a module rather than synthesise one.

## Review found a blocking defect; it is fixed, and it is worth reading

**`native-wave32-admitted` claimed to mirror the renderer's gate and mirrored one-fifth of it.** That
gate is five conjuncts (`render_runner.h:7128-7133`); I reproduced the innermost equality (`:7140`)
alone. The decisive conjunct is `allow_native_fragment_vote_width`, which defaults false (`:565`) and
is set from `title_id == "PPSA04263"` alone (`live_renderer.cpp:1216`, `:7591`). **So for every title
but GTA V the renderer drops these shaders while the census called them admitted** — an undercount of
exactly the loss #3464 exists to size.

It also invalidated this PR's own positive control, which is the part worth dwelling on. I had
reported the *Blasphemous 2* frame as "2 requiring a wide wave, **0 dropped**" and presented it as
evidence the instrument worked. The true answer is **2 dropped**. The control passed because it
agreed with what I expected of a title that renders correctly.

Fixed by reporting what the module decides and refusing to claim the rest: `reason-set-admissible`
for the equality, `internal-gds=` / `subgroup-features=` for the two conjuncts a module *can* decide,
and a `gate-undecided=` line naming the three it cannot, printed on every census so admission cannot
be read in by omission. `capture_wave_census.py` takes `--title`; without it the title is assumed
**not** allowlisted, true for every title but one.

Second blocking finding, also fixed: `discover_submit` took `sorted(first=)[-1]` over
`--bundle-ds-summary`, which is the max over each depth-stencil identity's *first appearance* — an
arbitrary submit. It now reads the bundle's `submits=N` and refuses when N != 1.

Six non-blocking findings taken, including a truncated `.spv` being tallied as "requires nothing"
(now `spirv-walk=ok|truncated`), `recompiled=n/a` for a SPIR-V input, and two false statements in the
tool's own comments.

## Verification

- **Every commit is TDD**: 6/6, 7/7, 11/11 and 4/4 arms red before their implementations. **51 checks
  pass.**
- The SPIR-V fixture is built **by hand** from the header plus two `OpModuleProcessed` strings, not by
  recompiling — a fixture from the recompiler would share its assumptions with the reader under test.
- Both admission directions exercised on one capture: `--title PPSA13579` → DROPPED=2,
  `--title PPSA04263` → DROPPED=0, no `--title` → DROPPED=2. The field cannot be a constant.
- **The corrected control was then confirmed at runtime**, not left derived: a 100-second Blasphemous
  2 boot emits 8 `[render] skip … requires subgroup size 64 … why=0x2 wave-any` lines.
- Malformed SPIR-V fed by hand (overrunning word count, zero-length instruction, unterminated string,
  unknown reason bit): all terminate, and the unknown bit prints `names=unknown-bit20`, so the
  drift guard fires on an instance built outside the generator.
- Registered ctest case `shader_inspect_resource_table_honesty`; registration unchanged.
- Docs gate clean; `diff --check` clean.

## Result it produced

A GTA V world frame (16 of 114 submits, depth-stencil passes, phase confirmed by screenshot):
1923 draws, 93 distinct fragment shaders, 41 requiring >32 lanes — **20 admitted by the existing
allowance, 21 dropped**, all `lane-id + wave-ballot`, top two covering 254 draws. That is W3 on
#3464, and it reordered the plan there.

## CI

`Windows MinGW` is red and it is **not this PR**: `capture_tunables` times out identically on `main`,
bisected to a single commit (#3460) and filed as #3470. This diff is five files under
`tools/shader_inspect/` and cannot reach `gpu_timeline`. Every other check is green.

## Risk

Low — opt-in, exclusive mode in an offline diagnostic; the shared code path it touches is argument
parsing. The judgement worth a reviewer's attention is now the `gate-undecided=` framing: whether
reporting the module half plus a named list of undecidable conjuncts is genuinely honest, or still
lets a consumer infer admission.

### DIFF
--- a/prosper/tools/shader_inspect/README.md
+++ b/prosper/tools/shader_inspect/README.md
@@ -119,19 +119,40 @@ dropped. On *Grand Theft Auto V* that removes most of the world's lighting (#346
 
 ```sh
 shader_inspect <raw-rdna2.bin> --wave-reasons
-wave-reasons required-subgroup-size=64 reasons=0x2 names=wave-any native-wave32-admitted=1
-wave-reasons-end file=... dwords=12 recompiled=1 table_dependent=0 endpgm=1
+wave-reasons required-subgroup-size=64 reasons=0x2 names=wave-any reason-set-admissible=1 \
+    internal-gds=0 subgroup-features=0x1
+wave-reasons gate-undecided=title-allowlist,host-subgroup-size-control,host-subgroup-features ...
+wave-reasons-end file=... input=rdna2 dwords=12 recompiled=1 table_dependent=0 endpgm=1
 ```
 
-`native-wave32-admitted` mirrors the shipping gate in `tests/fixtures/render_runner.h`
-**exactly** -- equality against `kFragmentWaveReasonWaveAny` -- rather than re-deriving what
-ought to be admissible. The census reports the emulator that ships; a proposal about what
-*should* be admitted belongs in the classifier, where a test can hold it.
+### `reason-set-admissible` is NOT `admitted`, and the difference is most of the answer
 
-A module that needs no particular width reports `size=0 reasons=0x0 admitted=0`: it never
-reaches the gate, and counting it as admitted would bury the interesting population under every
-ordinary shader in a dump. A module carrying no reason marker at all reports `reasons=absent`,
-never `0x0` -- absent is not none.
+The renderer's gate is **five conjuncts** (`tests/fixtures/render_runner.h:7128-7133`). The
+reason-set equality at `:7140` -- the one this field mirrors -- is only the innermost. The
+decisive one is `bd.allow_native_fragment_vote_width`, which **defaults false** (`:565`) and is
+set in exactly one place, from `title_id == "PPSA04263"` (`live_renderer.cpp:1216`, assigned
+`:7591`). Its own comment says so: *"Tests, replay and all other titles leave this false."*
+
+**So outside GTA V, every shader requiring more than 32 lanes is dropped regardless of its
+reason set.** An earlier version of this tool called the reason-set test `native-wave32-admitted`
+and was therefore wrong in the one direction that mattered: it under-reported the loss #3464
+exists to size. Its *Blasphemous 2* validation run reported `0 dropped` where the true answer
+was `2` -- and it went unquestioned because it agreed with the expectation.
+
+The two conjuncts a module CAN decide are reported beside it (`internal-gds`,
+`subgroup-features`); the three it cannot are named on their own `gate-undecided=` line, every
+time, so no consumer can read admission into the row by omission. For a per-title verdict use
+`capture_wave_census.py --title`.
+
+A module that needs no particular width reports `size=0 reasons=absent reason-set-admissible=0`.
+`reasons=0x0` is never printed: a module either carries the marker or does not, and `absent` is
+not none. A module that fails to recompile prints **no census row at all** -- only the sentinel,
+with `recompiled=0`.
+
+A SPIR-V input reports `recompiled=n/a` (nothing was recompiled -- the module was read) plus
+`spirv-walk=ok|truncated`. That last field exists because a corrupt module reports `size=0
+reasons=absent`, byte-identical to a genuine wave-free shader, so without it an unreadable dump
+lands in the population that says #3464 does not affect it.
 
 `wave_reason_census.py` tallies the mode over a directory and prints the reason-set histogram
 plus the names of the dropped shaders. It exists because the *runtime* skip log answers this
--- a/prosper/tools/shader_inspect/README.md
+++ b/prosper/tools/shader_inspect/README.md
@@ -158,8 +158,25 @@ its numbers. A census that silently equated "could not be analysed" with "needs 
 would have reported that GTA V has essentially no Wave64 problem, which is the opposite of the
 truth.
 
-For a census over shaders that use memory, the reasons must come from a path that HAS the real
-descriptors -- `gpu_replay <capture>.prgcap`, as with every other table-dependent verdict here.
+### The way out: read the module gpu_replay compiled
+
+`--wave-reasons` accepts **either** a raw RDNA2 stream **or** a SPIR-V module, detected by the
+SPIR-V magic rather than by a flag or a file extension -- the magic is the format's own
+self-identification, so a renamed or mislabelled file cannot be read as the wrong kind.
+
+`gpu_replay` has the real resource table, and `--dump-shader DRAW:fs PATH` already writes the
+recompiled module. So the census over memory-using shaders is a two-step pipeline with no new
+analysis in it:
+
+```sh
+gpu_replay --dump-shader 18:fs fs.spv <capture>.prgcap   # real descriptors, real compile
+shader_inspect fs.spv --wave-reasons                     # reads the module's own markers
+```
+
+The reasons then come from a real-table compile while this tool remains the single place that
+formats them. A SPIR-V input reports `input=spirv` in the sentinel and leaves `table_dependent`
+and `endpgm` at zero -- those describe an RDNA2 walk that did not happen.
+
 
 ## `--stage` cannot prove a shader is unsupported (#1571)
 
--- a/prosper/tools/shader_inspect/README.md
+++ b/prosper/tools/shader_inspect/README.md
@@ -111,6 +111,56 @@ input or one that is not a whole number of dwords. A stream that stops short of 
 reported as `endpgm=0` rather than as a failure — a census over a truncated dump is still an exact
 census of what could be decoded.
 
+## `--wave-reasons`: which shaders need the guest Wave64, and why
+
+On any NVIDIA host the entire supported subgroup range is `32..32` -- no extension, driver or
+device yields a 64-wide subgroup -- so a fragment module that requires the guest's Wave64 is
+dropped. On *Grand Theft Auto V* that removes most of the world's lighting (#3464).
+
+```sh
+shader_inspect <raw-rdna2.bin> --wave-reasons
+wave-reasons required-subgroup-size=64 reasons=0x2 names=wave-any native-wave32-admitted=1
+wave-reasons-end file=... dwords=12 recompiled=1 table_dependent=0 endpgm=1
+```
+
+`native-wave32-admitted` mirrors the shipping gate in `tests/fixtures/render_runner.h`
+**exactly** -- equality against `kFragmentWaveReasonWaveAny` -- rather than re-deriving what
+ought to be admissible. The census reports the emulator that ships; a proposal about what
+*should* be admitted belongs in the classifier, where a test can hold it.
+
+A module that needs no particular width reports `size=0 reasons=0x0 admitted=0`: it never
+reaches the gate, and counting it as admitted would bury the interesting population under every
+ordinary shader in a dump. A module carrying no reason marker at all reports `reasons=absent`,
+never `0x0` -- absent is not none.
+
+`wave_reason_census.py` tallies the mode over a directory and prints the reason-set histogram
+plus the names of the dropped shaders. It exists because the *runtime* skip log answers this
+only for whatever a route happened to reach, once per distinct shader, at whatever scene depth
+the run stopped: two runtime counts taken that way, 43 and 86, differed by run length alone and
+read as a regression until they were matched by frame.
+
+### Measured limit: over a raw dump this mode is mostly blind, and says so
+
+Run against a 3,453-shader dump of GTA V's pixel-shader database (2026-09-08):
+
+```
+files=3453  errors=0  unrecompiled(no resource table)=3430  analysed=23
+```
+
+**99.3% contributed no reason data.** A real pixel shader samples a texture, and the section
+below explains why a table-less tool can never lower one. The 23 that were analysed are the
+shaders that happen not to touch memory, which is not a random sample of anything -- a
+texture-sampling lighting shader is exactly the kind that is missing.
+
+This is reported rather than hidden: an unrecompilable shader is counted in its own column and
+is **never** tallied as `requires nothing`, and the tally prints the caveat both above and below
+its numbers. A census that silently equated "could not be analysed" with "needs no wide wave"
+would have reported that GTA V has essentially no Wave64 problem, which is the opposite of the
+truth.
+
+For a census over shaders that use memory, the reasons must come from a path that HAS the real
+descriptors -- `gpu_replay <capture>.prgcap`, as with every other table-dependent verdict here.
+
 ## `--stage` cannot prove a shader is unsupported (#1571)
 
 **`shader_inspect` has no resource table, and a table-less stage rejection is NOT evidence of a shader
--- a/prosper/tools/shader_inspect/capture_wave_census.py
+++ b/prosper/tools/shader_inspect/capture_wave_census.py
@@ -1,0 +1,166 @@
+#!/usr/bin/env python3
+"""Wave-width census over a captured frame, using the REAL resource table.
+
+`wave_reason_census.py` sweeps a directory of raw RDNA2 dumps, and over GTA V's shader database
+99.3% of them could not be lowered: shader_inspect has no descriptors and a real pixel shader
+samples a texture. That census is therefore blind to precisely the lighting shaders #3464 is about.
+
+gpu_replay has the real descriptors. This script drives it:
+
+    gpu_replay --bundle F --bundle-extract-submit N cap.prgcap   # one submit, replayable
+    gpu_replay --inspect-only cap.prgcap                         # draw list, with fs= identities
+    gpu_replay --dump-shader ID:fs fs.spv cap.prgcap             # the REAL-TABLE compiled module
+    shader_inspect fs.spv --wave-reasons                         # reads that module's own markers
+
+so every reason it reports comes from a module compiled the way the renderer compiles it, and no
+analysis is reimplemented anywhere.
+
+    python capture_wave_census.py --gpu-replay G --shader-inspect S --capture cap.prgcap
+    python capture_wave_census.py --gpu-replay G --shader-inspect S --bundle f.prgbundle
+
+SCOPE, which bounds every number this prints: a capture is ONE FRAME. This answers "of the fragment
+shaders this frame ran, how many need a wide wave and what for" -- not "how many does the title
+have". That is the right question for #3464 (a frame with missing lighting is the evidence), but it
+is a different question from the database sweep, and the two must not be quoted as one number.
+"""
+import argparse
+import collections
+import re
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+
+DRAW = re.compile(r"^draw\[(\d+)\].*?\bfs=(\d+)/([0-9a-f]+)/", re.M)
+DS_SUBMIT = re.compile(r"\bfirst=(\d+) last=(\d+)")
+ROW = re.compile(r"wave-reasons required-subgroup-size=(\d+) reasons=(\S+) names=(\S+) "
+                 r"native-wave32-admitted=(\d+)")
+END = re.compile(r"wave-reasons-end .*?recompiled=(\d+)")
+
+
+def run(cmd):
+    return subprocess.run([str(c) for c in cmd], capture_output=True, text=True)
+
+
+def discover_submit(gpu_replay, bundle):
+    """The bundle's submit index is the capture's own ordinal, not 0 or 1.
+
+    `--bundle-extract-submit` matches on `submit_index` and rejects 0, so on a single-submit bundle
+    neither of the two numbers anyone tries first works: 0 is refused by the parser and 1 reports
+    "bundle has no submit 1". `--bundle-ds-summary` prints the real ordinal as `first=`/`last=`.
+    """
+    done = run([gpu_replay, "--bundle", bundle, "--bundle-ds-summary"])
+    found = {int(m.group(1)) for m in DS_SUBMIT.finditer(done.stdout + done.stderr)}
+    if not found:
+        return None
+    return sorted(found)[-1]
+
+
+def main() -> int:
+    ap = argparse.ArgumentParser()
+    ap.add_argument("--gpu-replay", required=True)
+    ap.add_argument("--shader-inspect", required=True)
+    ap.add_argument("--capture", help="a .prgcap (one submit)")
+    ap.add_argument("--bundle", help="a .prgbundle; one submit is extracted from it")
+    ap.add_argument("--submit", type=int, help="submit index in the bundle (default: discovered)")
+    ap.add_argument("--work", help="directory for intermediates (default: a temporary one)")
+    args = ap.parse_args()
+    # Resolved to absolute up front. A relative tool path that a shell finds is not
+    # necessarily one CreateProcess finds on Windows, and the failure is a bare
+    # FileNotFoundError from deep inside subprocess that names neither which path nor which
+    # of the four child invocations produced it.
+    args.gpu_replay = str(Path(args.gpu_replay).resolve())
+    args.shader_inspect = str(Path(args.shader_inspect).resolve())
+    for name in ("capture", "bundle"):
+        if getattr(args, name):
+            setattr(args, name, str(Path(getattr(args, name)).resolve()))
+    if bool(args.capture) == bool(args.bundle):
+        print("give exactly one of --capture or --bundle", file=sys.stderr)
+        return 2
+
+    tmp = None
+    if args.work:
+        work = Path(args.work)
+        work.mkdir(parents=True, exist_ok=True)
+    else:
+        tmp = tempfile.TemporaryDirectory()
+        work = Path(tmp.name)
+
+    capture = args.capture
+    if args.bundle:
+        submit = args.submit or discover_submit(args.gpu_replay, args.bundle)
+        if submit is None:
+            print("could not discover a submit index; pass --submit", file=sys.stderr)
+            return 2
+        capture = work / "submit.prgcap"
+        done = run([args.gpu_replay, "--bundle", args.bundle,
+                    "--bundle-extract-submit", submit, capture])
+        if not Path(capture).exists():
+            print("extract failed:\n" + done.stdout + done.stderr, file=sys.stderr)
+            return 2
+        print("extracted submit %d" % submit)
+
+    done = run([args.gpu_replay, "--inspect-only", capture])
+    draws = DRAW.findall(done.stdout + done.stderr)
+    if not draws:
+        print("no draws found in %s" % capture, file=sys.stderr)
+        print((done.stdout + done.stderr)[-2000:], file=sys.stderr)
+        return 2
+
+    # One dump per DISTINCT fragment shader. The `fs=<dwords>/<hash>/` identity on each draw line is
+    # the renderer's own content key, so this counts shaders rather than draws -- a frame that draws
+    # one shader 400 times must not read as 400 blocked shaders.
+    first_draw = {}
+    draw_count = collections.Counter()
+    for draw_id, dwords, fs_hash in draws:
+        draw_count[fs_hash] += 1
+        first_draw.setdefault(fs_hash, (draw_id, int(dwords)))
+
+    results = {}
+    for fs_hash, (draw_id, dwords) in sorted(first_draw.items()):
+        spv = work / ("fs_%s.spv" % fs_hash)
+        run([args.gpu_replay, "--dump-shader", "%s:fs" % draw_id, spv, capture])
+        if not spv.exists():
+            results[fs_hash] = ("dump-failed", 0, "", "", 0)
+            continue
+        done = run([args.shader_inspect, spv, "--wave-reasons"])
+        out = done.stdout + done.stderr
+        if not END.search(out):
+            # The sentinel's purpose: no line at all must never be tallied as "needs nothing".
+            results[fs_hash] = ("no-sentinel", 0, "", "", 0)
+            continue
+        row = ROW.search(out)
+        if not row:
+            results[fs_hash] = ("no-census-row", 0, "", "", 0)
+            continue
+        results[fs_hash] = ("ok", int(row.group(1)), row.group(2), row.group(3),
+                            int(row.group(4)))
+
+    ok = {h: r for h, r in results.items() if r[0] == "ok"}
+    bad = {h: r for h, r in results.items() if r[0] != "ok"}
+    needs = {h: r for h, r in ok.items() if r[1] > 32}
+    dropped = {h: r for h, r in needs.items() if not r[4]}
+
+    print("draws=%d  distinct fragment shaders=%d  analysed=%d  unreadable=%d"
+          % (len(draws), len(first_draw), len(ok), len(bad)))
+    print("require>32=%d  admitted at native wave32=%d  DROPPED=%d"
+          % (len(needs), len(needs) - len(dropped), len(dropped)))
+    if needs:
+        hist = collections.Counter((r[2], r[3]) for r in needs.values())
+        print("\nreason sets among the %d requiring a wide wave:" % len(needs))
+        for (mask, names), n in hist.most_common():
+            print("  %-8s %-42s %3d shaders" % (mask, names, n))
+    if dropped:
+        print("\ndropped shaders (hash, reasons, draws using it):")
+        for h, r in sorted(dropped.items(), key=lambda kv: -draw_count[kv[0]]):
+            print("  %s  %-8s %-40s %3d draws" % (h[:16], r[2], r[3], draw_count[h]))
+    if bad:
+        print("\n%d shader(s) yielded no census row -- NOT counted as needing nothing:" % len(bad))
+        for h, r in sorted(bad.items()):
+            print("  %s  %s" % (h[:16], r[0]))
+    print("\nSCOPE: one frame. This is what this frame RAN, not what the title contains.")
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(main())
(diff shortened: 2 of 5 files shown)


mattias800 (commented):
REJECTED — two blocking findings, both in the "reports a confidently wrong number and nobody would
notice" class this diagnostic exists to avoid. Neither is a crash; both bias a count in the same
direction (fewer shaders reported as dropped), which is the direction that would make #3464 look
smaller than it is. Everything else I checked held up, including the parts I most expected to break;
detail at the end so the re-review is cheap.

I verified each claim against the source rather than against the PR body, and re-ran the tool where
a claim was checkable (`build-tdd/shader_inspect.exe`, 28/28 arms pass on this head).

---

## BLOCKING 1 — `native-wave32-admitted` does not mirror the gate. It mirrors the innermost `if` of two, and the outer one is title-scoped to GTA V.

`shader_inspect.cpp:287` computes `admitted = marked && reasons == kFragmentWaveReasonWaveAny`.
That equality is real and it is at `render_runner.h:7140` as claimed. But `:7140` is nested inside
`render_runner.h:7128-7133`, which is a five-conjunct `if`:

```cpp
if (fragment_subgroup_skip && bd.allow_native_fragment_vote_width &&
    (ctx.subgroup_stages & VK_SHADER_STAGE_FRAGMENT_BIT) &&
    prosper::gpu::fragment_subgroup_features_supported(...) &&
    !(uses_internal_gds && !ctx.fragment_stores_atomics)) {
```

`bd.allow_native_fragment_vote_width` defaults to **false** (`render_runner.h:565`) and is assigned
in exactly one place, `live_renderer.cpp:7591-7592`, from
`gta5_native_fragment_vote = title_id == "PPSA04263"` (`live_renderer.cpp:1216`). The field's own
comment at `render_runner.h:563-564` states the consequence outright:

> Live-title authority for GTA V's reviewed WaveAny-only fragment route. **Tests, replay and all
> other titles leave this false**, preserving the strict exact-wave admission contract.

So on every title except `PPSA04263`, a `native-wave32-admitted=1` from this census is a shader the
shipping renderer **drops**. The column is right for the one title the tool was built for and wrong
everywhere else, with nothing in the output, the README or the two scripts saying so.

Two concrete consequences, not hypotheticals:

- `capture_wave_census.py:148` derives `dropped` as `needs and not admitted`, so on any non-GTA-V
  capture the headline `DROPPED=` is an **undercount** by exactly the wave-any population — the
  population most likely to be non-empty.
- **The PR's own positive control is one of these wrong numbers.** *Blasphemous 2* is `PPSA13579`,
  so `allow_native_fragment_vote_width` is false for it; its "2 requiring >32 lanes, both
  `reasons=0x2` (wave-any), **0 dropped**" is not what the shipping renderer does on that
  Windows/NVIDIA host — it drops both. The control does show the instrument can print a non-constant
  admission column, which is worth something; it does not show the column agrees with the renderer,
  and the PR reads it as if it did.

On the completeness argument in the PR body: the two-caller grep is **true** (I reran it —
`render_runner.h:7135` and `:7176` are the only non-recompiler, non-test callers), but it does not
establish what it is offered for. Which function computes `reasons` was never the risk; which
conditions surround the comparison was, and those live in a different expression that the grep
cannot see.

Two of the four missing conjuncts are pure functions of the module and could actually be evaluated
here rather than assumed — `fragment_spirv_uses_internal_gds` (`render_runner.h:7115-7116`, used at
`:7133`) and `fragment_spirv_required_subgroup_features` (`:7132`) — so today the census silently
assumes both pass. The other two are host state and honestly cannot be known offline.

Any of these fixes clears it, cheapest first: rename the column to say what it is
(`gtav-native-wave32-admitted`), or gate it on a `--title` argument, or evaluate the two
module-computable conjuncts and state in the README that the remaining two are assumed. What must
not stand is `README.md` "mirrors the shipping gate in `tests/fixtures/render_runner.h` **exactly**"
plus "the census reports the emulator that ships", because for every tracked title but one it
reports an emulator that does not exist.

## BLOCKING 2 — `discover_submit` picks an arbitrary submit on a multi-submit bundle, and the census then calls it a frame.

`capture_wave_census.py:45-56` takes `sorted(found)[-1]` over the `first=` fields of
`--bundle-ds-summary`. Those lines are `ds-identity … first=%llu last=%llu` at
`gpu_replay.cpp:1310-1319`, and `first` is `stats.first_submit` — **the lowest submit index in which
that DS identity appears** (`:1266`). So the selection is "the maximum, over depth-stencil
identities, of each identity's first appearance". That is not the largest submit, not the busiest,
not the last; on a single-submit bundle it happens to be the only right answer, which is why the
trap the docstring records was found and this one was not.

The consequence is the same shape as finding 1: on a bundle with N submits the script silently
censuses one of them, prints `draws=… distinct fragment shaders=…`, and closes with
`SCOPE: one frame. This is what this frame RAN`, while having read a fraction of the frame. It also
prints `extracted submit %d` (`:101`) without ever saying how many there were.

The count it needs is already in the bytes it already reads: `discover_submit` captures
`done.stdout + done.stderr`, and `gpu_replay.cpp:1383` writes
`[gpureplay] bundle v%u submits=%zu …` to stderr on that same invocation. Parse it, print it, and
refuse (or loudly warn) when `submits > 1` and no `--submit` was given.

Same function, minor: `submit = args.submit or discover_submit(...)` (`:91`) discards an explicit
`--submit 0`. Harmless today because the parser rejects 0, but it is the kind of `or` that outlives
the reason it was safe.

---

## Non-blocking

**3. A truncated or corrupt `.spv` is counted as "analysed, requires nothing".**
`shader_inspect.cpp:274` sets `recompiled = !spirv.empty()`, which for a SPIR-V input is true for
*any* file that starts with the magic. Both readers bail to `0` / `UINT32_MAX` on a bad word count
(`rdna2_to_spirv.cpp:3196`, `:3260`). Measured on three hand-built inputs:

```
5-dword header only   -> required-subgroup-size=0 reasons=absent … recompiled=1
magic + truncated op  -> required-subgroup-size=0 reasons=absent … recompiled=1
4 dwords (< header)   -> required-subgroup-size=0 reasons=absent … recompiled=1
```

Both scripts put those in the analysed denominator and outside `require>32` — i.e. "could not be
read" tallied as "needs nothing", which is precisely the failure this PR is built to prevent, just
reached down the SPIR-V path instead of the RDNA2 one. `capture_wave_census.py:129` checks only
`spv.exists()`, so a partial write, or a stale `fs_<hash>.spv` in a reused `--work` directory, lands
here. A `spirv.size() >= 5 && spirv[4] == 0` plus a clean-walk check, reported as
`input=spirv-malformed`, closes it.

**4. Two statements in the block's own comment are false against the code they annotate.**
- `shader_inspect.cpp:250` and the README: "A module that needs no particular width reports
  `size=0 reasons=0x0 admitted=0`". It does not — a wave-free fragment shader emits no marker, so the
  output is `required-subgroup-size=0 reasons=absent names=none native-wave32-admitted=0` (measured
  on the PR's own `plain_words` fixture). A consumer written against the documented `reasons=0x0`
  matches nothing, silently.
- `shader_inspect.cpp:255-257`: "A recompile that fails prints reasons=absent rather than 0x0".
  It prints **no census row at all** — measured on the `cbuf_load` fixture, output is only
  `wave-reasons-end … recompiled=0 table_dependent=1`. The behaviour is the better one; the comment
  describes a different one, and it is the comment a maintainer would "restore" the code to.

Worth stating plainly while fixing these: `reasons=absent` covers both "no marker because nothing
required a width" and "marker missing (pre-#2147)". The distinction survives only when read together
with `required-subgroup-size`, and the README should say that rather than implying `absent` alone
carries it.

**5. `shader_inspect.cpp:269-270` — "A raw RDNA2 stream cannot begin with it — `0x07230203` decodes
as neither a valid instruction nor a plausible first word."** The first half is false, by this
tool's own decoder:

```
pc=0000 len=1 fmt=VOP2 op=0x003 words=07230203 dst=vgpr:145 src0=sgpr:3 src1=vgpr:129
generic-coverage total=1 alu=1 exp=0 table=0 unsupported=0
```

"Implausible as a first word" is fair and is enough to justify the design. "Not a valid instruction"
is not, and it is the stated justification for a silent, un-overridable mode switch. Reword, and/or
add the header sanity check from finding 3 so detection needs more than one word of agreement.

**6. Test gap on the highest-stakes contract.** Nothing asserts that a table-dependent raw shader
yields `recompiled=0` and **no** census row — the property the entire `unrecompiled` column rests on,
and the one whose failure would move 3,430 of GTA V's 3,453 shaders into "requires nothing". The
fixture is already in the file (`cbuf_load`), so this is three lines. Mode exclusivity is likewise
unpinned; I re-ran both combinations by hand and both exit 2, so this is regression cover, not a
defect.

**7. Cosmetic.** `shader_inspect.cpp:150` usage line is indented 4 where its neighbours are 8. The
README's `wave-reasons-end` example omits `input=rdna2`, which the real output prints (the PR body's
example has it). `wave_reason_census.py:81` silently excludes any fragment dump whose filename lacks
`_PS_`.

---

## Checked and sound — recorded so nobody re-derives it

- **The SPIR-V readers are safe on hostile bytes.** `rdna2_to_spirv.cpp:3188-3218` and `:3251-3277`
  both check the magic and `size < 5`, bound every step with `!words || words > spirv.size() - offset`,
  bound `memchr` by the instruction's own word count, and only ever `strtoul` a NUL-terminated region
  inside the buffer. I fed them a header-only module, a lying word count that swallows the next
  instruction, a sub-header file, and `Why=-1` / `Why=4294967295` / `Why=8589934594`: no crash, no
  OOB read. The wrapping cases all land on `UINT32_MAX` and print `absent`, which is the safe
  direction and unreachable from the emitter (`rdna2_to_spirv_internal.hpp:4496` prints `%u`).
- **The 16 MiB bound and whole-dword requirement** (`shader_inspect.cpp:180-190`) reject rather than
  truncate, which is the right choice for a census input.
- **`capture_wave_census.py:34`'s `DRAW` regex is right against real output.**
  `gpu_replay.cpp:781-808` prints `… vs=%zu/%016llx/%s fs=%zu/%016llx/%s`, lowercase zero-padded, so
  `\bfs=(\d+)/([0-9a-f]+)/` matches and cannot be confused with the `vs=` field. `draw[%llu]` is
  `d.draw_index`, which is exactly the key `--dump-shader DRAW:fs` resolves through
  `realized_shader_dump.hpp:99-105` — the identity parse and the dump selector agree, which was the
  obvious place for a silent mismatch. A zero match is loud (`no draws found`, exit 2), and a dump
  that fails is bucketed as `dump-failed`, not as a shader needing nothing.
- **`--dump-shader` really does work under `--inspect-only`**: the dump at `gpu_replay.cpp:3088`
  runs before `if (inspect_only) return 0;` at `:3546`. Commit 3 is sound.
- **The reason-name table** (`shader_inspect.cpp:44-52`) is written against the header's own
  constants, so bit *values* cannot drift from `rdna2_to_spirv.hpp:532-547`; only names can. The
  `unknown-bitN` fallback works — a hand-built `Why=256` prints
  `reasons=0x100 names=unknown-bit8 native-wave32-admitted=0`.
- **The SPIR-V fixture really is hand-built and really does exercise the reader.**
  `test_shader_inspect.py`'s `module_processed`/`spirv_module` assemble the header and two
  `OpModuleProcessed` strings with `struct.pack` and opcode 330 — no recompiler involvement — and
  both admission directions plus the unmarked case are asserted, so the column cannot be a constant.
  That part of the verification claim holds exactly as described.
- **`wave_reason_census.py`'s `ROW`/`END` regexes** match the real lines in every shape I could
  produce (`absent`, `none`, `unknown-bit8`, multi-name comma lists — `print_wave_reason_names` never
  emits a space or an empty field), an absent marker can never reach the `admitted` tally because
  `needs` is filtered on `size > 32` first, and a missing sentinel is an error rather than a zero.
  Finding 3 is the one input shape that gets past this.
- Trailer gate clean (0), `diff --check` clean, both new scripts CRLF like the rest of the tree.

Re-review on the corrected head; findings 1 and 2 are what I would want to see change, and finding 1
is the one the whole tool turns on.

mattias800:
Both blocking findings confirmed against the source and fixed in `ccb37fe1`. Thank you — finding 1
was the important one, and it was wrong in the only direction that mattered.

## Finding 1 — confirmed, and worse than a naming problem

I verified every link in the chain myself rather than taking the citations: `render_runner.h:7128`
is a five-conjunct `if`; `allow_native_fragment_vote_width` is declared `= false` at `:565` with the
comment *"Tests, replay and all other titles leave this false"*; its sole assignment is
`live_renderer.cpp:7591` from `gta5_native_fragment_vote`, defined `:1216` as
`title_id == "PPSA04263"`.

**You are right that this invalidated my own positive control, and that is the part worth dwelling
on.** I reported the Blasphemous 2 frame as "2 requiring a wide wave, 0 dropped" and presented it as
evidence the instrument works. Re-run with the fix, the same capture gives **DROPPED=2** — and those
two shaders account for 5 draws in that frame. The control passed because it agreed with what I
expected of a title that renders correctly; a control that can only confirm what you already believe
is the failure mode the charter names, and I walked straight into it.

I also note your point about *why* my "only two callers" check missed it: it was a true statement
that answered a different question. Grepping the callee cannot see the conjuncts around the call
site, and I had actually read that `if` earlier in the session and still mirrored a fifth of it.

**Fix:** renamed to `reason-set-admissible`, for what it measures. The two conjuncts a module *can*
decide are now reported beside it (`internal-gds`, `subgroup-features`) rather than silently assumed
to pass — your observation that they were evaluable was the useful half — and the three it cannot
are printed on their own `gate-undecided=` line on every census, so admission cannot be read in by
omission. `capture_wave_census.py` takes `--title`; without it the title is assumed **not**
allowlisted, true for every title but one. "Possibly admitted" is labelled an upper bound.
`wave_reason_census.py` has neither title nor host, so it no longer uses the word at all.

Both directions are exercised on one capture: `--title PPSA13579` → DROPPED=2, `--title PPSA04263` →
DROPPED=0, no `--title` → DROPPED=2.

## Finding 2 — confirmed and fixed

`sorted(first=)[-1]` was indefensible once stated: those are per-DS-identity first appearances.
`discover_submit` now reads the bundle's `submits=N` and **refuses** when N != 1, asking for
`--submit`, rather than censusing a fraction of a frame under a banner saying "SCOPE: one frame".

## Non-blocking, all taken

- **Truncated `.spv` counted as "requires nothing"** — the sharpest of the six, same class as the
  unrecompiled column. Added `spirv-walk=ok|truncated` from a structural word-count walk mirroring
  the marker search; both scripts treat a bad walk as unreadable. I added an arm asserting a
  header-only module walks **cleanly**, so the fix cannot relabel every wave-free module as corrupt.
- **Two false statements in the comment block and README** — `reasons=0x0` is never printed, and a
  failed recompile prints no row rather than `absent`. Both corrected.
- **The `0x07230203` justification** — I verified it with this tool's own decoder before changing
  it: `pc=0000 len=1 fmt=VOP2 op=0x003`. You were right; the comment now rests on the real argument
  (no dump begins with that exact word) and points at `input=`.
- **`recompiled=` honesty** — I had found this myself while stress-testing and was holding it so
  your review stayed attached to head. A SPIR-V input now reports `recompiled=n/a`.
- Also fixed: `wave_reason_census.py` failed its own documented relative-path invocation. It
  reported `errors=3453` rather than a wrong number, so the sentinel machinery held, but it was
  broken.

## Verification

15 new arms (11 for the rename, 4 for the walk), red before their fixes at 11/11 and 4/4; 39 checks
pass. The database sweep still reports 3430/3453 unrecompiled and now names its own admission limit.

Ready for re-review at `ccb37fe1`.

mattias800 (commented):
APPROVED on head `757f2118`. Both blocking findings are genuinely fixed — not renamed around, fixed —
and I found nothing blocking in the fixes. Eight non-blocking items below; three of them I would want
closed before a number out of `--ds-submits` is quoted in an issue or a status doc, and I say which.

I re-derived the fixes rather than reading the commit messages, re-ran the tool (34/34 arms pass on
this head) and re-ran my own adversarial inputs from the first round against the new binary.

## The two blocking findings

**1 — admission. Fixed, and fixed at the right layer.** `shader_inspect.cpp:335` now computes
`reason_set_admissible` under a name that claims only what it tests, `:338-339` evaluate the two
module-decidable conjuncts instead of assuming them (`subgroup-features=`, `internal-gds=`), and
`:349-353` prints `gate-undecided=` unconditionally on every census row rather than only when
something looks interesting — which is the right call, because the failure mode was inference by
omission. `capture_wave_census.py:134` defaults `allowlisted` false and `:236` requires
`allowlisted and r[4]`, so the default answer is the one that is true for every title but one, and
`wave_reason_census.py` no longer uses the word at all. The README rewrite at :122-145 states the
five-conjunct structure with the citations and, unusually, records its own retracted number.

Two things I checked rather than took: `NATIVE_VOTE_ALLOWLIST` (`capture_wave_census.py:45`) is one
string in one place and it matches `live_renderer.cpp:1216`; and `fragment_spirv_uses_internal_gds`
(`rdna2_to_spirv.cpp:3304-3325`) has the same magic/`size<5`/word-count bail as the other readers, so
adding it to this path introduced no new way to read past a buffer.

Withdrawing the Blasphemous 2 control and re-running it to `DROPPED=2` was the right move, and
confirming the class at runtime — 8 `[render] skip … why=0x2 wave-any` lines on a real boot — makes it
an instrument checked against the thing it measures rather than against itself. That is the strongest
part of this round.

**2 — submit selection. Fixed, and the follow-on gap is fixed too.** `discover_submit`
(`capture_wave_census.py:60-86`) now reads `submits=N` from the bundle line and refuses rather than
guessing, `:158` handles `--submit` with `is not None` so an explicit `0` survives, and `757f2118`
adds the multi-submit path with `censusing N of M` (`:168`) and an unconditional lower-bound scope
statement (`:263-266`). Cross-capture dedup is right: `first_draw` is keyed on `fs_hash` across every
selected submit (`:191-195`) and each dump uses the capture the draw came from (`:198-206`), so a
shader shared by twelve submits is one shader and not twelve.

---

## Non-blocking

**A. `--ds-submits` does not enumerate the depth-stencil submits, and three places say it does.**
This is the one I would fix first. `ds_submit_ordinals` (`:89-103`) collects `first=` and `last=` from
the `ds-identity` lines. Those two fields are `stats.first_submit` / `stats.last_submit`
(`gpu_replay.cpp:1266-1267`, printed `:1310-1319`) — the first and last submit in which *that DS
identity* appears, not every submit it appears in. An identity live across submits 3..40 contributes
`{3, 40}` and the 36 submits between are never censused.

So the docstring "These are the submits that ran a depth-stencil pass" (`:92-93`), the `--help` text
"census every submit named by --bundle-ds-summary (the depth-stencil passes)" (`:114-115`) and the
closing "--ds-submits covers the depth-stencil passes only … not UI or post passes" (`:267-269`) all
overclaim, and the last one specifically invites the reading "world coverage is complete, only UI is
missing". The printed numbers stay honest — `censusing N of M` and "every count here is a LOWER bound"
are unconditional and adjacent — which is why this is not blocking. But the label is what a reader
carries away, and on the 114-submit GTA V bundle the difference between "the DS passes" and "at most
two submits per DS identity" is most of the frame. Either reword to "the submits where each
depth-stencil identity first and last appears — a sample, not the set", or enumerate properly (a
`--submits 1-114` range would do it, and `--bundle-find-ds` names every matching submit per address).

**B. The fix to finding 2 fails open if the bundle line ever stops matching.** `BUNDLE_SUBMITS`
(`:49`) is the whole guard: at `:79` a `None` count skips the `!= 1` refusal, and if the DS identities
happen to share one `first=` value the `len(ordinals) > 1` check at `:83` passes too. `:167` then
hardcodes `chosen, bundle_submits = [one], 1`, so the script asserts "censusing 1 submit(s) of 1 in
the bundle" having never established that there is one. The format is in-repo and stable, so this is
unlikely — but it is the same fail-open shape the original finding was about, and the fix is to refuse
when the count cannot be read and to propagate the real count instead of the literal `1`.

**C. A structurally clean walk can still be a misparse, and the one case I can construct prints three
reassuring fields.** `spirv_stream_walks_cleanly` (`:87-101`) mirrors the marker traversal exactly —
same start offset, same `!count || count > size - offset` bail, plus the exact-landing test the
readers do not need — and it is honestly scoped in its own comment ("structural, not semantic … not a
substitute for spirv-val"). Answering the question directly: a module cannot walk `truncated` while
being fine (a well-formed module tiles exactly, and the header-only arm pins the boundary case), but
it *can* walk `ok` while being misread. My round-one adversarial fixture still does:

```
wave-reasons required-subgroup-size=0 reasons=0x2 names=wave-any reason-set-admissible=1 …
wave-reasons-end … spirv-walk=ok
```

A lying word count swallowed the size marker while the counts still tiled to the end. Note what that
row is: `marked && size == 0`, which the emitter cannot produce — `Prosper.FragmentSubgroupWhy` is
written only inside `if (is_fragment && fragment_required_subgroup_size)`
(`rdna2_to_spirv_internal.hpp:4484-4499`). So a one-line self-consistency check (a reason set without
a size is inconsistent, report it as such) closes the only hole I could build, and it costs nothing.
Deliberate corruption only, hence non-blocking.

**D. `gate-undecided=` names three host conjuncts and there are four.** `:349-353` lists
`title-allowlist,host-subgroup-size-control,host-subgroup-features`, but `render_runner.h:7133`'s
`!(uses_internal_gds && !ctx.fragment_stores_atomics)` has a host half too. The module half is now
reported (`internal-gds=`), so a careful consumer is not misled; the line just claims to be the
complete list of what the tool cannot see, and `host-fragment-stores-atomics` belongs on it whenever
`internal-gds=1`.

**E. Two test gaps, one of which was reported as taken.** Round-one finding 6 — a table-dependent raw
shader must yield `recompiled=0` and **no** census row — is still unpinned; `cbuf_load` exists at
`test_shader_inspect.py:64` and is never run through `--wave-reasons`. I verified the behaviour is
correct (`cb.bin` prints only `wave-reasons-end … recompiled=0 table_dependent=1`), so this is missing
regression cover on the property the entire `unrecompiled` column rests on, not a defect. `recompiled=n/a`
is likewise unpinned, and it is now load-bearing: `wave_reason_census.py:31`'s `END` regex requires
`recompiled=(\d+)`, so a `.spv` fed to that script errors out rather than tallying — correct behaviour,
but it depends on a spelling no test holds.

The 15 arms that *are* new do what they claim. Each fails if its feature is removed, `"native-wave32-admitted"
not in out` is a real anti-regression on the old name, and the four `spirv-walk` arms cover both
directions plus the zero-count non-termination guard and the header-only boundary — that last one
matters most, since without it the obvious fix relabels every wave-free module as corrupt.

**F. `shader_inspect.cpp:187-188`: "whether render_runner.h would\nwould ADMIT at a host native
wave32"** — duplicated word and a missing object, in the usage text a first-time user reads.

**G. Two stale sentences that now contradict their own tools.** `wave_reason_census.py:4-6` still says
the script answers "how many of those prosper admits today at a host native wave32", which is exactly
the claim `ccb37fe1` removed from its output; and `README.md:158` still says that script prints "the
names of the dropped shaders", which it no longer does (`:126-129` prints "first wide-wave shaders by
name" with an `admissible=` column). Neither is a number, but per the charter a stale line carries the
document's authority.

**H. `--work` reuse is not bundle-scoped.** `capture_wave_census.py:171` skips extraction when
`submit_<N>.prgcap` already exists, with nothing tying that file to the bundle that produced it, so a
second run with a different bundle and the same `--work` silently censuses the first bundle's submits.
A stale `fs_<hash>.spv` is safe by construction (content-hashed) except when the dump fails, where
`:207` accepts the leftover.

---

Nothing above changes a count in a way a reader could not see, which is why this is an approval. If it
were my call on sequencing: fix A, B and the `--help` wording before anyone quotes a `--ds-submits`
figure, and take the rest whenever.

For the record, since it is easy to lose across two rounds: the `Windows MinGW` red is `capture_tunables`
timing out identically on main, bisected to #3460 and filed as #3470 — not this PR. I did not re-run
CI or the builds.

mattias800 (commented):
APPROVED on head `904a8b79`. The approval stands. A, B and E are closed correctly, the delta is
confined to the two files you named, and nothing in it introduces a new way to get a wrong count.

**First, a correction to my own previous review.** I wrote "34/34 arms pass on this head" for
`757f2118`. That number was wrong — I read it off a truncated tail rather than counting. Measured:
`757f2118` ran **46** checks, `904a8b79` runs **51** (44 `check(` call sites; the two-stage loop runs
some twice), 0 failures on both. Your "39" is also off, in the same harmless direction. It changes no
conclusion, but a wrong figure inside a review carries the review's authority, which is exactly the
authority it should not have.

**A.** Corrected in all four places, and the docstring now carries the mechanism rather than a
restatement (`capture_wave_census.py:97-106`, `:127`, `:177-178`, `:283-286`). "Samples … biased
toward the ends of each identity's life" is the right characterisation. That you had already published
a `--ds-submits` figure on #3464 under the old description and went and corrected it there is the part
that actually closed this — the code fix alone would have left the wrong characterisation standing
where people read it. One trivial slip in the new docstring: an identity live across submits 3..40
contributes `{3, 40}` and **36** are never named, not 37 (3..40 inclusive is 38 submits).

**B.** Fail-closed and correct: `:79-84` refuses an unreadable header before the `!= 1` test, so
`:181` can now say `chosen, bundle_submits = [one], 1` and mean a measured 1. I checked the ordering —
`discover_submit` returns non-`None` only through the path that parsed a count of exactly 1 — and the
comment you left at `:179-180` is the right thing to leave, because that literal `1` is precisely what
a future reader would otherwise assume was an assumption.

**E.** Five arms added, and the omission is now recorded in the test's own comment
(`test_shader_inspect.py:280`, "This was reported as covered in an earlier round and was not"), which
is worth more than the arms.

---

## The judgement you asked to have challenged

**The mutation is too narrow to establish the other four arms — and it did not need to be wider.**
Those are separate claims and I think both are true.

`if (recompiled)` → `if (true)` probes exactly one contract: that the row is gated on the recompile
succeeding. That is the subtle one, it is the one whose cover was missing, and one arm reddening is
the right result — the others *should* stay green under it. Specifically, under that mutation `spirv`
is empty, so `fragment_spirv_required_subgroup_size({})` returns 0 and `..._reasons({})` returns
`UINT32_MAX`; the emitted row would read `reason-set-admissible=0`. So arm 4 staying green is correct
behaviour, not weak cover.

Taking the five one at a time, against the mutation that *would* kill each:

- **"emits NO census row"** — killed by your mutation. Load-bearing, and the one that mattered.
- **"reports recompiled=0"** — killed by `recompiled ? 1 : 0` → `1` at `shader_inspect.cpp:352`, a
  different mutation you did not run. Independently load-bearing.
- **"recompiled=n/a, never a count"** — killed by removing the `is_spirv` branch at `:349-351`. Also
  independent, and it guards a spelling `wave_reason_census.py:31` depends on.
- **"still emits the sentinel"** — killed by an early `return` on recompile failure. Real, though it
  overlaps sentinel cover that already exists on other paths; its unique content is "even when the
  recompile failed".
- **"not reported as reason-set admissible"** — I do not think this one can fail. It is implied by the
  row-absence arm, and in the one world where that arm breaks the value printed is `0` anyway, as
  derived above. It is an assertion satisfied for a reason other than the one it names. Harmless, but
  it is not a fifth piece of cover, and I would either drop it or point it at something independent —
  `"gate-undecided=" not in out` is a genuinely separate property of the failure path, since that
  printf sits inside the same `if (recompiled)`.

So: four of five bite, against four distinct mutations, only one of which you ran. I would not ask you
to run the other three — they kill on literal output fields where a defect is visible in the first
diff — but I would not describe the single mutation as having established the set, and your instinct
to flag it rather than assume it is the correct one.

## C, D, F, G, H — none block, and I would file them

I agree with your read on all five, and I checked C rather than resting on my earlier note: the
`Prosper.FragmentSubgroupWhy` marker is emitted only inside
`if (is_fragment && fragment_required_subgroup_size)` (`rdna2_to_spirv_internal.hpp:4484-4499`), so the
`marked && size == 0` row my fixture produces cannot come out of this emitter. It needs a corrupted
module whose bad word counts still tile exactly to the end — deliberate, not accidental. D is a
completeness nit on a line that already tells the consumer it cannot decide admission, with the module
half reported beside it. F and G are prose. H needs a user to point `--work` at two different bundles.
None of them can move a number in normal use.

One thing I would do rather than leave them in a review comment: **open one issue covering C, D, F, G
and H, referencing #3464.** The charter's rule is that a defect found and not fixed in the same
session gets an issue immediately, and the specific reason applies here — a finding that lives only in
a PR thread is one the next agent re-derives at full cost, and these five are cheap to fix and easy to
never notice again. C is the only one worth a line of prose in the issue (the one-line
`marked && !size` consistency check, and why the case is unreachable from the emitter).

Merge it.

`Windows MinGW` red is #3470 (`capture_tunables` timing out identically on main, bisected to #3460),
unrelated to this PR. I did not re-run the builds or CI; I ran only `shader_inspect` and the Python
suite.

mattias800 (commented):
APPROVED on head `80efc88a`. No closer look needed — merge it.

The swap is one hunk in one file and it is correct. I checked the independence claim rather than
taking it: the `gate-undecided` printf is its own statement at `shader_inspect.cpp:349`, separate from
the row printf at `:341-345`, and both sit inside the same `if (recompiled)`. So it can be lifted out
of that guard on its own, and the row-absence arm cannot see that happen — its token
(`required-subgroup-size=`) is not on the gate-undecided line. The new arm covers exactly the failure
the old one only appeared to. The comment you left at `:288-292` records why the first draft was
vacuous, which is the part that stops it being rewritten back.

Choosing a mutation *for* the arm rather than reusing the one you had is the right habit and it is the
whole difference between a characterisation test and cover.

Measured on this head, since we have both been wrong about it once: **51 checks, 0 failures**, exit 0.
That matches your count.

#3471 for C, D, F, G and H closes my remaining items. Nothing outstanding from me on this PR.

For the record across the four rounds: two blocking findings (the admission column mirroring only the
innermost conjunct of a five-conjunct gate, and `discover_submit` selecting an arbitrary submit), both
fixed at the layer that caused them rather than papered over; a published census figure on #3464
retracted by the author when the first one landed, and a second characterisation corrected there when
finding A landed; one test arm found unfalsifiable and replaced; and two wrong check counts — one
mine, one yours — retracted in place. The instrument that comes out of this reports a number I would
be willing to quote, which was not true of the one that went in.

`Windows MinGW` remains #3470 and unrelated. I did not run the builds or CI in any round; my evidence
throughout is `shader_inspect` runs, the Python suite, and the source.

mattias800:
Merging with `Windows MinGW` red. Recording why, so the exception is auditable rather than assumed.

**The failure is not this PR.** Verified rather than inferred: the failing job is
`260 - capture_tunables (Timeout)`, 1 of 288 tests, every other test and every other job green. The
same test times out identically on `main`, bisected to a single commit (#3460) and filed as **#3470**.
This diff is five files under `tools/shader_inspect/` plus a README and cannot reach `gpu_timeline`.

I checked the specific test rather than accepting "Windows is red, that'll be the known one" — which
is the failure mode #3470 warns about, and would be the easiest way for a genuine regression to ride
in behind a known one.

**State at merge:**

- head `80efc88a`; PR head, remote tip and the approving review's commit all agree
- 10 checks pass, 1 fails (#3470), 0 pending
- 51 test checks pass locally, 0 failures
- `diff --check` clean; the diff's single deleted line is the `run()` signature replaced by its
  `extra=` variant, so nothing was silently reverted
- four review rounds, two blocking findings fixed, five non-blocking follow-ups filed as **#3471**

Issue #3470, opened 2026-09-08
capture_tunables times out on Windows MinGW since #3460 — main red on every run, blocking the merge gate


mattias800:
## `capture_tunables` times out on Windows MinGW, and `main` has been red since #3460

`main`'s **Windows MinGW** job has failed on every run since 2026-09-08 ~08:44Z. Always the same
single test, always a timeout rather than an assertion:

```
99% tests passed, 1 tests failed out of 288
The following tests FAILED:
	260 - capture_tunables (Timeout)
Errors while running CTest
##[error]Process completed with exit code 8.
```

Every other job is green, and every other one of the 288 tests passes. Linux does not reproduce it.

## Bisected to one commit

| run | head | Windows MinGW |
| --- | --- | --- |
| 34202318366 | `e93a8d02` | **success** |
| 34206146116 | `cb1e4576` | failure |
| 34207718506 | `f4df75d3` | failure |
| 34207785492 | `00b29507` | failure |
| 34209901323 | `65ac7a25` | failure |
| 34220933884 | `e699d0e6` | failure |

`git log e93a8d02..cb1e4576` is exactly one commit:

```
cb1e4576 fix(capture): write owned frame bundles outside guest flip calls (#3460)
```

and it touches `prosper/src/gpu/timeline/gpu_timeline.{cpp,hpp}` — the translation unit
`capture_tunables` exercises. Verified as the same failing test on runs 34220933884 and 34207785492,
so this is not one flaky run being generalised.

## What is established vs. what is a hypothesis

**Established:** the test times out on Windows only; `main` was green immediately before `cb1e4576`
and has been red on that test ever since; the sole commit in the window edits the code under test.

**Hypothesis, NOT verified:** #3460 changed *when* the owned frame-bundle writer runs (moving it out
of guest flip calls), so a thread that previously completed inside the flip may now not be joined, or
may wait on something that never arrives, on Windows. The adjacent log line
`[grab] frame-bundle write failed: capture window cancelled at shutdown` is consistent with a
shutdown-ordering problem but appears elsewhere in the run too, so it is a lead rather than evidence.
Nobody has attached a debugger; the ctest timeout is 600 s, so it hangs rather than running slowly.

A `Timeout` verdict also means **no assertion output was produced**, so the test's own diagnostics
say nothing about which case hangs — worth knowing before someone reads the empty output as "no
information".

## Why this is filed separately rather than fixed in passing

[...]

PR #246, opened 2026-07-09
feat(tools): screenshot — capture a PNG every N rendered frames from a game run

mattias800:
A tool to run a game and capture a PNG every N rendered frames until M screenshots, then exit — for snapshot-testing runs and eyeballing progression.

## Usage
```
screenshot <app0-dir> [--every N=60] [--count M=30] [--out DIR=.] [--timeout SECS=900]
```
Only the game is required; everything else defaults sanely.

## Design
- **Reuses the shared helpers** — `boot_program` (boot) + `frontends/shared/live_renderer` (composite). No duplicated boot/render glue.
- **Real PNGs** via zlib (~40 KB/frame vs 8 MB uncompressed BMPs).
- **Filenames** `<titleCode>_<runTimestamp>_<index>.png` (e.g. `PPSA24651_20260709-195505_00.png`) — one `YYYYMMDD-HHMMSS` stamp per run so a folder of many runs groups and sorts cleanly. titleCode = dump basename minus a trailing `-app0`.
- **"Frames" = rendered frames**, via a new `present_frame_seq()` on the present layer. The guest flips far faster than llvmpipe renders, so counting flips (`present_count`) bunched every shot into the first second; rendered frames spread them across the run. Confirmed: shots land at rendered-frames 20/40/60 exactly.
- Defaults the render-frontier guest env (`PROSPER_GUEST_FS=1`, `PROSPER_GUEST_ARGS=-force-gfx-direct`) if unset; overridable per title (e.g. UE4: `PROSPER_GUEST_ARGS= PROSPER_NULL_PAGE=1`).

## Verified
- On The Messenger: 3 shots at frames 20/40/60, valid 1920×1080 RGBA PNGs (~43 KB each).
- Full `ctest` **65/65** (the `present_frame_seq` addition is additive; nothing regressed).
- Built wherever Vulkan + zlib are found (same block as `boot_trace`).

PR #3473, opened 2026-09-08
tools(shader_inspect): skip_survey.py — 31 refused shaders across 5 of 9 titles, all wave-any (#3464 W7)

mattias800:
Refs #3464 (W7). Companion to #3466, which landed the census tools.

## Why

The wave64 fragment loss was found twice by eye and once by a census, and nobody knew how many
titles it affects. **Every priority on #3464 was a guess without that list**, which is why the plan
there now puts this first.

This is the cheapest possible instrument: the renderer already prints one line per distinct refused
shader **with the reason mask**, so a boot and a grep answers for a whole title in about two minutes.
No capture, no GPU replay, no new emulator code.

## Result

Windows 11 / RTX 4090, default boot, 150 s per title, nine dumps:

| title | | refused | reasons |
| --- | --- | ---: | --- |
| `PPSA01885` | Evergate | **11** | `0x2 wave-any` |
| `PPSA13579` | Blasphemous 2 | **8** | `0x2 wave-any` |
| `PPSA21564` | — | **6** | `0x2 wave-any` |
| `PPSA02664` | Alex Kidd in Miracle World DX | **4** | `0x2 wave-any` |
| `PPSA25009` | Blue Prince | **2** | `0x2 wave-any` |

**31 refused shaders across 5 of 9 titles, and the entire corpus produces exactly ONE distinct reason
mask.** Not a single `lane-id` or `wave-ballot` outside GTA V.

So every one of the 31 is the class the existing reviewed classifier already admits, kept out only by
`title_id == "PPSA04263"` (`live_renderer.cpp:1216`). **W5 — retire the title scope — is worth 31
shaders across 5 titles with no new classifier**, which is considerably more than it looked like when
it was the last item on the list. Three of the five are rung 6 with reviewed snapshot guards, all
green while dropping shaders.

## The tool's own first run produced a wrong answer

Recorded as **instrument trap 273** in this commit, because the shape generalises well beyond this
tool.

I anticipated that a zero would be ambiguous and guarded it: a run that rendered **no** frames is
reported `UNKNOWN`, since a title that never draws refuses nothing by construction. That reasoning is
true and **insufficient**. **GTA V rendered 5160 frames and reported 0 refused shaders; its world
refuses 21.** A 150 s default boot reaches its menus, and the frame count only *looked* like a
validity check.

**I caught it solely because I already knew GTA V's answer.** On any of the other eight titles it
would have published "clean" with nothing to warn a reader — and four of them did report zero.

The general shape: *a survey that cannot distinguish "surveyed and found nothing" from "did not
survey the thing" is most confident about the subjects it understands least.* It is trap 122's family
pointed at **coverage** rather than at a generator or a parser.

**Fixed by removing the claim rather than improving the guess**, since the tool cannot know what
content a given title gates behind a route: it no longer prints the word "clean" at all, a zero reads
as `none in window`, the window prints beside every row, and route-gated titles carry the caveat
inline. The four remaining zeros are explicitly **not** upper bounds.

## Units

Counts are **distinct shaders, never draws**. The renderer's message sits inside a shader-identity
dedupe guard while the `continue` that drops the draw sits outside it, so the two are measured on
opposite sides of the same guard — a census that divided one by the other reported "1.5% of draws"
and was withdrawn (`render_runner.h:6940-6944`).

Reason masks are printed by the renderer and not decoded here; a second implementation of the bit
names is exactly the drift this project has been bitten by.

## Verification

- **Validated against a hand-derived answer rather than against itself** before its numbers were
  trusted: Blasphemous 2 was measured independently at 8 shaders, all `why=0x2`, by reading a boot
  log directly; the tool reproduces exactly that.
- Docs gate on `GAME_COMPAT_ORCHESTRATION.md`: 335 rows, unbroken, `1..273` unique and ascending, no
  baseline row deleted.
- `diff --check` clean; trap-41 three-dot deletion check reports zero deletions.
- Trap number allocated with `trap_number.py` (scans `main` plus open PRs), so it should not collide.

## Risk

Very low. A new standalone script plus one documentation row; nothing existing is modified and no
emulator code is touched. The judgement worth review is whether the corrected zero-reporting is now
honest enough — it is the second time in two PRs that a label of mine overclaimed what a measurement
covered, so I would rather it were checked than taken.


mattias800 (commented):
**REJECTED** — three blocking findings, all cheap. The instrument is the right one and trap 273 is a genuinely good row; what blocks is that the tool still has one silent-zero path the trap does not cover, one silent-undercount path, and no test pinning it to the emitter it parses.

Reviewed at `42005bf8` against `origin/main` (`3713cbfd`).

---

## What checks out

- **`SKIP` matches the real emitter.** Derived the field order and spelling by reading the `fprintf` at `prosper/tests/fixtures/render_runner.h:7201-7215` — `[render] skip draw=%zu fs=%016llx: fragment shader requires subgroup size %u (device range ... why=%s)` — and exercised `skip_survey.SKIP` against lines built from it: `0x2 wave-any` parses to `('...a3','64','0x2',' wave-any')`; a bare mask `why=0x40)` matches too (`\S+` backtracks off the paren); `why=0x0 none` matches. Only the `why=unrecorded (pre-#2147 module)` branch (`render_runner.h:7182`) parses oddly — mask `unrecorded`, names `(pre-#2147 module` — cosmetic, and it still counts the shader.
- **`FPS` matches all three emitters.** `prosper/frontends/prosper-app/main.cpp:3014`, `:3099`, `:3121` — checked each variant including the `, N.N distinct -- picture unchanged` suffix. The print is not gated on `--fps` (`main.cpp:2989` is `shown - mark >= 60` only), and `shown` is cumulative, so `max()` at `skip_survey.py:84` is the right reducer.
- **The units claim is correct against draws.** `render_runner.h:7157-7160` puts the message inside the `logged.insert(shader_key)` guard and the draw drop outside it, and `logged` is a process-global `static` never cleared — so one process per title is the right shape.
- **`--volume` is a real flag** (`main.cpp:1375`); `--out`, `--only`, `--json` behave as documented.
- **Docs gate green on the merge result**, run here: `check_numbered_table.py --ordered --table-header Instrument --baseline origin/main` reports `1..273, 273 rows, unique and ascending`, and `none of the 272 baseline rows deleted`; repo-wide `check_numbered_table.py` over every tracked `.md` is clean; `check_merge_result.py` reports a clean merge whose result passes the gate against `origin/main` as of now. `git diff --check origin/main...HEAD` clean; three-dot `--numstat` is `1/0` and `160/0`, zero deletions. Trap 273 belongs in that table.

---

## Blocking

### B1. GTA V's zero is over-determined, and both `ROUTE_GATED` and trap 273 publish only half the cause

`prosper/tools/shader_inspect/skip_survey.py:43-49`, `:130-131`; `prosper/docs/GAME_COMPAT_ORCHESTRATION.md:424`.

`ROUTE_GATED["PPSA04263"]` says the zero is because *"world is behind the Story/Performance menu route; a default boot surveys menus"*, and the trap row says GTA V reported 0 *"because a 150 s default boot reaches its menus and not its world"*. That is not the whole reason, and the missing half is the one that matters for this tool.

`prosper/frontends/shared/live/live_renderer.cpp:1216` sets `gta5_native_fragment_vote = title_id == "PPSA04263"`, which reaches `bd.allow_native_fragment_vote_width` at `:7591`. In `render_runner.h:7128-7156`, when that flag is set and the reason set is **exactly** `kFragmentWaveReasonWaveAny` (`:7140`), the shader is *admitted* at native width and logs

```
[render] GTA V native-width fragment vote: subgroup 64 -> 32 (why=0x2)
```

— **not** a `skip` line. So on PPSA04263 this survey is structurally blind to the wave-any class **at any route and any window length**. Routing it to the bank heist would not fix its zero for that class.

That matters because wave-any-only is *precisely* the class this PR is counting: all 31 hits across the other five titles are `0x2 wave-any`. GTA V is therefore the one title where `skip_survey` counts a different population than the other eight, and nothing in the tool or the row says so. The caveat as written actively invites the next reader to run `scripts/gta5/reach-performance-story.pad`, see a small number, and conclude GTA V's world is nearly clean — the same trap one level down, with the tool's own warning text as the thing that misleads them.

The sibling tool in the same folder already models this exactly, so the knowledge exists two files away: `capture_wave_census.py:44-46` keeps `NATIVE_VOTE_ALLOWLIST = ("PPSA04263",)` "in one place, mirroring live_renderer.cpp:1216", and its docstring (`:20-30`) spells out that the decisive conjunct is `allow_native_fragment_vote_width`, not the route. That is also why the 21 it reports and the 0 this reports are not the same measurement.

Fix (small): correct the `ROUTE_GATED` text for PPSA04263 to name the native-width admission as the primary blindness and the route as secondary; correct the same sentence in the trap 273 row; and — cheapest real improvement — add a second regex for `[render] GTA V native-width fragment vote:` so the tool can report that class on GTA V instead of a structural zero.

### B2. No test pins either regex to the emitter's format string — trap 272's own rule, one PR later

`prosper/tools/shader_inspect/skip_survey.py:41-44`. There is no test file for this script and no arm added to the existing `prosper/tools/shader_inspect/test_shader_inspect.py`.

Trap 272, added by you at `GAME_COMPAT_ORCHESTRATION.md:423` in the immediately preceding PR, states the rule: *"any tool that parses emulator output needs at least one arm whose input the EMULATOR produced (or, where a unit test cannot run the emulator, one that reads the producer's format string ... and fails if the field disappears)."* `test_flip_pacing_report.py` case 7 is the pattern to copy.

Both regexes work **today** — I verified that above — so this is not a live wrong number. It is blocking because the tool's worst failure mode is a silent all-zero survey that looks like good news, the numbers are already published on #3464, and the emitter is live code that will be edited (a field inserted before `why=`, a widened `fs=`, an `[app]` line reworded). Two cases, roughly 15 lines: one grepping `render_runner.h` for the literal `"[render] skip draw=%zu fs=%016llx: fragment shader requires subgroup size %u "` and one grepping `main.cpp` for `"[app] %.1f fps (%llu frames"`, each failing if the literal moves. Compose the sample lines from the grepped literal, not by hand — hand-composing is the half of trap 272 that bit `flip_pacing_report.py`.

### B3. The row reports the nominal window and can discard a true positive

Two paths, same shape — the printed row does not record what actually happened:

- **Early exit is silently discarded.** `skip_survey.py:71` exits the loop when `proc.poll() is not None`; `:73` then skips the terminate branch and nothing captures `proc.returncode` or the elapsed time. A title that crashed at 8 s produces a row indistinguishable from a full run, and the summary at `:140-141` prints `args.seconds` as the window for every title regardless. For a PR whose entire thesis is that a count is meaningless without the coverage it was taken over, printing a nominal window the run may not have had is the same error in miniature.
- **`frames == 0` suppresses a non-zero count.** `:124-125` and `:136` — when no `[app] ... fps` line was seen, the row prints `-` for refused and the verdict `UNKNOWN -- rendered no frames, so it refused nothing by construction`, **even if `refused_shaders > 0`**, and `:138` excludes it from `affected`. The premise does not hold: `frames == 0` does not mean "never drew", it means "never presented 60 frames" (`main.cpp:2989`), and the skip line is emitted from the submit path independently of presentation. So `refused > 0, frames == 0` is reachable and is *positive* evidence that the current code throws away — a false negative in the headline "5 of 9".

Fix: record `returncode` and measured elapsed seconds in the result dict and print them per row; and print `refused_shaders` unconditionally, qualifying only the verdict when `frames == 0`.

---

## Non-blocking

### N1. `fs=` is a compile-instance counter, not a content hash — the counts are upper bounds

`skip_survey.py:82-83`, keyed on the `fs=` field from `render_runner.h:7141-7142` (`bd.fs_identity ? bd.fs_identity : hash_buffer_words(...)`). `fs_identity` is `cache.next_identity++` (`prosper/src/gpu/execute/gpu_executor.cpp:1935`, `:2257`), and that cache is keyed by **guest address** with 4096-entry LRU eviction plus byte-mismatch invalidation (`:1240-1265`); `live_renderer.cpp:2576-2580` says so outright — *"an evicted shader recompiles to a NEW identity"*. So the same shader program at two guest addresses, or recompiled after eviction, counts twice. The direction is systematic inflation. At 2-11 per title over 150 s the magnitude is plainly small, so I do not think 31 is materially wrong — but the docstring's *"Every number here is shaders"* should say "distinct compile identities, an upper bound on distinct programs". Note also this is a **different identity scheme** from `capture_wave_census.py:203`, which keys on the renderer's content key `fs=<dwords>/<hash>/`; the PR compares "0 here" against "21 there" as if the units matched.

### N2. Windows `terminate()` can truncate the log it then reads

`skip_survey.py:67-79`. `stderr` is redirected to a file, and nothing in the app calls `setvbuf(stderr, ...)` — `main.cpp:1351` sets `stdout` only, and the skip `fprintf` has no `fflush`. Under the MSVC/UCRT rule (stdout/stderr flush per call only for a *character device*), a file-redirected stderr is buffered, and `Popen.terminate()` on Windows is `TerminateProcess` with no CRT flush. Up to a buffer's worth of trailing stderr can be lost — direction: undercount. Likely small, since skips fire on first use of each shader, but it is on the platform these numbers were taken on and it is silent. Also `:78` calls `proc.kill()` with **no following `wait()`**, so `:79` reads the log while the process may still be dying, and there is no `try/finally` around the `Popen`, so an interrupt orphans the app.

### N3. The `ROUTE_GATED` caveat only fires on the zero branch

`skip_survey.py:126-133`. A route-gated title with a **non-zero** count takes the `else` at `:132` and prints its reason histogram with no caveat at all — the more dangerous case, since a partial count reads as a complete one. Append the caveat in both branches.

`ROUTE_GATED` as a concept is fine: `:45-46` explicitly calls it "a courtesy, not a safety net", and the unconditional summary at `:142-144` does not depend on it. I would not block on staleness given that framing.

### N4. The run's environment is inherited but not recorded

`skip_survey.py:64-66` does `env = dict(os.environ)` then overlays two variables. Any `PROSPER_*` in the operator's shell silently changes what is surveyed, and nothing in the log or the JSON says which build and which environment produced a row. Dump the inherited `PROSPER_*` (and the app path) into the head of each per-title log or into the JSON.

### N5. Not discoverable

`prosper/tools/shader_inspect/README.md` documents `capture_wave_census.py` (`:145`) and `wave_reason_census.py` (`:157`) but gains no entry for `skip_survey.py`. One paragraph.

---

To be clear about proportion: B1 is the one I would not let through, because it puts an incomplete mechanism into a permanent trap row and into the tool's own warning text, and this project treats those rows as its most expensive knowledge. B2 and B3 are each a few lines. Everything else is documentation.

mattias800 (commented):
**REJECTED** — one blocking finding, five lines of CMake. B1, B2 and B3 are all genuinely fixed on their merits; what blocks is that `test_skip_survey.py` is registered nowhere, so the trap row's claim that drift "reddens" is not true of anything automatic.

Re-reviewed at `db99d907` (branch tip confirmed via `git ls-remote`). Previous review was `42005bf8`.

---

## Your three questions, answered first

### (a) Does counting admits close B1, or relocate it? — **It closes it.**

Verified the mechanism independently rather than from your description. `render_runner.h:7146-7151` emits `[render] GTA V native-width fragment vote: subgroup %u -> %u (why=0x%x)`, and `skip_survey.py:74-75`'s `ADMIT` matches it. Critically, that `fprintf` sits inside `if (native_width_logged.insert(shader_key).second)` at `:7146` — a *separate* static set from the skip path's `logged` at `:7160`, both process-global and never cleared. So `len(ADMIT.findall(text))` at `:112` really is one entry per distinct shader, and the emitter's dedupe (not the script's) is what makes it so.

The two counts also partition cleanly rather than overlapping: the admit branch sets `fragment_subgroup_skip = false` (`:7155`) before the skip block is reached, and a wave-any-only shader that fails one of the *other* conjuncts at `:7128-7133` never enters the admit branch at all and lands in `refused`. So on an allowlisted title `refused + admitted` covers the population with no double-count and no gap. That is a real measurement of the thing that was invisible, not a relabelled hole — and reporting it as "what the allowlist is buying, i.e. what W5 would extend" is the right frame.

Two riders under Non-blocking (N6, N7): the row's own note now contradicts the column beside it, and the shader-ness of `admitted` lives entirely in C++ that nothing pins.

### (b) Is the allowlist-drift test sound? — **It cannot miss a divergence. It over-fires, and the natural repair to the over-fire writes a false claim into the tool's output.**

I mutation-tested it myself in an isolated copy of the four source files rather than taking the claim (`test_skip_survey.py:106-113`). Seven mutations, all reddening exactly one arm and no others:

| mutation | result |
| --- | --- |
| `fs=%016llx` → `shader=%016llx` in the skip emitter | FAIL — "every title will report zero" arm |
| `"why=%s)` → `"reason=%s)` | FAIL — why= arm |
| drop `GTA V ` from the admit emitter | FAIL — admit arm |
| `%llu frames` → `%llu presented` in `main.cpp` | FAIL — fps arm |
| **add a second title to the wave gate** (`\|\| title_id == "PPSA25009"`) | FAIL — drift arm, `renderer has ['PPSA04263','PPSA25009'], survey has ['PPSA04263']` |
| **refactor the gate to `is_native_vote_title(title_id)`** | FAIL — drift arm, `renderer has [], survey has [...]` |
| **add an UNRELATED per-title switch elsewhere in `live_renderer.cpp`** | FAIL — drift arm, same message |

Rows 5 and 6 are the ones you needed and they hold, including the refactor case where the grep finds nothing and fails loudly rather than silently agreeing.

Row 7 is the soft spot. `test_skip_survey.py:110` greps `r'title_id == "([A-Z]{4}\d{5})"'` across the whole of `live_renderer.cpp`, not the wave gate specifically. Today that is exact — `grep -n 'title_id == "' prosper/frontends/shared/live/live_renderer.cpp` returns exactly one line, `:1216` — which is why it passes. But the moment anyone adds an unrelated per-title workaround to that file, this arm goes red with a message that names two sets and invites one repair: add the new title to `NATIVE_VOTE_ALLOWLIST`. Do that and `skip_survey.py:169-171` prints `ALLOWLISTED: wave-any is admitted here, not skipped` for a title where it is not, so its real refused count reads as if it were structurally uncountable.

It fails loud and closed, so I am not blocking on it, but the fix is one line: anchor the pattern to the assignment, e.g. `r'native_fragment_vote\s*=\s*title_id == "([A-Z]{4}\d{5})"'`. That still reddens on rows 5 and 6 (I checked row 6's shape: the assignment survives, the string literal does not, so the set empties).

### (c) Does the reworked reporting overclaim? — **No. It now under-claims in one place, which is the safe direction but reads as a contradiction.** See N6.

### On N1 (`fs=` identity), which you asked me to rule on: **it should not block, and I would still fix it here.**

The reason it does not block is that the second incomparability is larger than the one you are worried about and is already documented: `capture_wave_census.py:31-33` says a capture is ONE FRAME and that its number "is a different question from the database sweep, and the two must not be quoted as one number". So 21 and 31 were never addable regardless of identity scheme, and the identity difference is the smaller term. The error direction is inflation only, bounded by how often a program recompiles at a fresh address inside one boot, and at 2-11 per title the magnitude is visibly small. Post-fix the GTA V comparison is also carried mechanically (the admit count) rather than numerically, which is what made it load-bearing before.

The reason to fix it anyway is that it costs one sentence in a file you are already editing: `skip_survey.py:16` still says "Every number here is shaders", and the honest wording is "distinct compile identities — an upper bound on distinct programs, since `fs_identity` is `cache.next_identity++` (`gpu_executor.cpp:1935`) and an evicted shader recompiles to a new one (`live_renderer.cpp:2576-2580`)", plus a half-line saying `capture_wave_census.py` keys on a content hash so the two tools' numbers must not be summed. Deferring it means the next person to hold both numbers has to re-derive that, which is the thing this project's `## Ruled out` discipline exists to prevent.

---

## Blocking

### B4. `test_skip_survey.py` is registered nowhere, so nothing runs it — and the trap row says it does

`prosper/tools/shader_inspect/test_skip_survey.py` (whole file); `prosper/CMakeLists.txt:2100-2105`; `prosper/docs/GAME_COMPAT_ORCHESTRATION.md:424`.

The sibling test in the same folder **is** a ctest case:

```cmake
if(Python3_Interpreter_FOUND)
  add_test(NAME shader_inspect_resource_table_honesty
           COMMAND "${Python3_EXECUTABLE}"
                   "${CMAKE_CURRENT_SOURCE_DIR}/tools/shader_inspect/test_shader_inspect.py"
                   "$<TARGET_FILE:shader_inspect>")
endif()
```

`test_skip_survey.py` has no `add_test`, appears in no workflow, and is not referenced from any CMake file. `grep -rn "test_skip_survey" prosper/CMakeLists.txt .github/workflows/` is empty.

Trap 273 now ends: *"`test_skip_survey.py` greps the emitter's format strings so a silent all-zero survey reddens"*, and the commit message says *"a test greps `live_renderer.cpp` so the two cannot drift apart silently"*. Neither is true of anything automatic — the drift *can* still happen silently, because the only thing standing in its way is a person remembering to run a script by hand. That is a claim in a permanent record that the code does not support, which is the same standard I blocked B1 on, so I am applying it the same way.

This is trap 272's own residue, in mirror image. Its text is *"the self-tests were registered; the format match with the live emitter was not, and CI stayed green for both."* Here you have got the format match exactly right and left the registration off. The `spv_validate` comment four lines below the block above makes the general point in this repository's own words: a gate that does not run *"used to report PASS when the validator was absent, which is how #1711 shipped invalid SPIR-V past a gate CLAUDE.md and both READMEs describe as strict."*

Fix: five lines in the same `if(Python3_Interpreter_FOUND)` block. It needs no target argument at all —

```cmake
  add_test(NAME skip_survey_emitter_pins
           COMMAND "${Python3_EXECUTABLE}"
                   "${CMAKE_CURRENT_SOURCE_DIR}/tools/shader_inspect/test_skip_survey.py")
```

While you are in there, please also fold in N6 and N7 below — both are one line and both are in the same blast radius.

---

## Non-blocking

### N6. The row's own note contradicts the column beside it

`skip_survey.py:169-171`, and the same wording at `:60-61` and `:191`. The note reads `ALLOWLISTED: wave-any is admitted here, not skipped -- invisible to this tool at any route`, printed on a row that now carries an `admitted` column containing precisely that class's count. `:191` likewise still says *"that class is invisible here at any route and any window"*.

It was true at `42005bf8` and your own fix falsified it. It errs conservative — a reader may discount a number that is in fact good — but a warning that the adjacent column visibly refutes is a warning people learn to ignore. Reword to scope it to the column: *"wave-any is ADMITTED here, not skipped — absent from `refused` by construction; see the `admitted` column"*, and the same in the summary.

### N7. `admitted` is a shader count only because of a C++ dedupe that nothing pins

`skip_survey.py:112` counts *lines*. `refused` at `:109-111` dedupes itself on the `fs=` hash, so its unit is the script's own; `admitted` has no hash in the line and inherits its unit entirely from `native_width_logged.insert(shader_key).second` at `render_runner.h:7146`. Remove that guard and `admitted` silently becomes a draw count while `refused` stays a shader count — the exact units mismatch the docstring at `:193-194` warns about, in the one column the test does not cover. One more arm: grep `render_runner.h` for `native_width_logged.insert(shader_key).second`.

### N8. The second `wait()` can now raise where the old code could not

`skip_survey.py:102-105`. On Windows `Popen.kill()` is `Popen.terminate()` is `TerminateProcess`, so if the first `wait(timeout=15)` timed out the second is likely to as well — and an uncaught `TimeoutExpired` there aborts the whole corpus run mid-survey, where the old fire-and-forget `kill()` merely risked a truncated read. My previous review asked for the `wait()` and did not think that through; wrap it in `try/except subprocess.TimeoutExpired` and record the fact in the row rather than raising.

### N9-N12, carried unchanged and still not blocking

N1 (`fs=` identity — see my ruling above), N2 (Windows CRT buffers file-redirected stderr and `TerminateProcess` does not flush it, so trailing lines can be lost; this now applies to `admitted` too), N4 (inherited `PROSPER_*` unrecorded), N5 (`README.md` in the folder still documents `capture_wave_census.py` at `:145` and `wave_reason_census.py` at `:157`, but not this script). N3 is fixed — the caveat is now appended in both branches at `:169-176`. Also a nit: the summary at `:183-184` still prints `args.seconds` as the window even when a title exited early, though the per-row `sec` column and the `EXITED EARLY` note now carry the truth.

---

## Verified on this head

- **B1 closed on the mechanism**, per (a) above. Trap 273's rewritten row is accurate against source: `render_runner.h:7140` is the exact-reason-set test, `:7155` is `fragment_subgroup_skip = false`, and the "`frames == 0` means presented fewer than 60, not never drew" correction matches `main.cpp:2989`.
- **B2 closed on content**, per (b) — seven independent mutations, seven correct arms. Registration is B4.
- **B3 closed.** `:181` no longer gates `affected` on `frames`; `:177-179` prints `refused` unconditionally; `elapsed`, `exited_early` and `returncode` are captured at `:96`, `:106`, `:121-123` and surfaced at `:174-176`.
- **Gates, run here at `db99d907`:** `check_numbered_table.py --ordered --baseline origin/main` → `1..273, unique and ascending, none of the 272 baseline rows deleted`; `check_merge_result.py` → clean merge, gate passes against `origin/main` as of now; repo-wide numbered-table check clean; `git diff --check origin/main...HEAD` clean; three-dot `--numstat` `1/0`, `205/0`, `121/0` — zero deletions; zero `Claude-Session:` trailers; zero generated-with attribution lines. `test_skip_survey.py` passes 11/11 by hand.

Register the test and I will approve on sight.

mattias800 (commented):
**APPROVED** on `c8187737`, with one string to correct on the way in (R1 below — one line, and it is the leftover third of N6, which was reported as taken).

Branch tip confirmed `c8187737` via `git ls-remote`. B4 closed; all four folded items land correctly.

---

## Your three questions

### (a) Is the ctest registration correct for this repo's conventions? — **Yes.**

`prosper/CMakeLists.txt:2107-2118`. Checked four things rather than eyeballing the block:

- **The guard is in scope and is the right one.** `find_package(Python3 COMPONENTS Interpreter QUIET)` at `CMakeLists.txt:147`, so `Python3_Interpreter_FOUND` is defined by the time line 2113 reads it, and it is the same guard the sibling `shader_inspect_resource_table_honesty` uses at `:2100`. A separate `if()` block rather than folding into the sibling's is the better call here — the sibling passes `$<TARGET_FILE:shader_inspect>` and yours passes nothing, so they are not the same dependency.
- **No name collision.** `skip_survey_patterns_match_emitters` appears once in the file; a dedupe over every `add_test(NAME ...)` shows the pre-existing repeats only (`game_compute_exec`, `live_target_format_match`, `plugin_autolink`, and two variants) and none is yours.
- **CI will actually run it.** This is the part worth checking, since "registered but never invoked" is the same defect one layer out: `.github/workflows/ci.yml:190` (Linux) and `:724` (Windows) are `ctest --test-dir prosper/build-ci --output-on-failure --timeout 600 --no-tests=error` with **no `-R` and no `-E`**, so an unfiltered run discovers it. It does not need adding to the by-name registry at `ci.yml:105-113` — that guard exists specifically for the Vulkan-gated tests (#1675), and the sibling is not in it either.
- **Nothing in the test needs a build.** It imports stdlib only, resolves `REPO` from `__file__` (`test_skip_survey.py:25`) so it is cwd-independent, and importing `skip_survey` has no side effect beyond compiling the patterns.

I did not duplicate your ctest run — reviewers here do not re-run the author's builds — so "discovered as test #246, reddens under ctest on an ADMIT reshape" is yours. What I can say is that the mechanism behind it is sound on all four counts above, and that I reproduced the underlying mutation at the script level (see (b)).

One thing to know rather than to fix: `find_package(... QUIET)` means a machine without Python3 registers this test silently, which is structurally the same "gate that isn't there" shape you just fixed. It matches the sibling exactly and CI's own steps shell out to `python3`, so it is not a new hazard — just do not read a green local ctest on a Python-less box as this test passing.

### (b) Did anchoring narrow it into missing a divergence? — **No. It catches strictly more than the bare form and over-fires strictly less.**

`test_skip_survey.py:110-115`. Mutation-tested in an isolated copy of the four source files, eight arms:

| mutation to `live_renderer.cpp` | expected | result |
| --- | --- | --- |
| second title ORed into the gate | red | **red**, `renderer has ['PPSA04263','PPSA25009']` |
| gate refactored to `is_native_vote_title(title_id)` | red | **red**, `renderer has []` |
| gate deleted outright (W5 lands) | red | **red**, `renderer has []` |
| the variable renamed | red | **red**, `renderer has []` |
| assignment split over three lines, second title added | red | **red**, both ids found |
| **unrelated per-title switch added elsewhere in the file** | **green** | **green** |
| the title swapped for a different one | red | **red**, `renderer has ['PPSA25009']` |
| drop `native_width_logged.insert(...)` from the admit emitter (N7) | red | **red**, "admitted would become a DRAW count" |

Row 5 is the one I would have expected anchoring to break, since `[^;]+` looks line-oriented — it is not, because a character class matches newlines, so the multi-line form is caught. Row 6 is the over-fire from my last review, now correctly green. Rows 2, 3 and 4 all fail *closed* (empty set), which is the important property: every way of making the grep stop finding the gate produces a red, not a silent agreement.

The one residual blind spot is inherent to grepping literals and was equally present before: an assignment whose RHS keeps `"PPSA04263"` while ORing in a non-literal lookup (`|| extra_titles.count(title_id)`) stays green. Anchoring did not introduce it and I would not chase it.

### (c) Is trap 273 accurate, including its self-correction? — **Yes, and the self-correction is the right instinct.**

`GAME_COMPAT_ORCHESTRATION.md:424`. Re-checked every load-bearing citation against source at this head: `render_runner.h:7140` is the exact-reason-set test, `:7146` the admit dedupe, `:7155` the `fragment_subgroup_skip = false`, and `main.cpp:2989` backs the "`frames == 0` means presented fewer than 60" correction. The added clause — that the test was unregistered when the row first cited it, and that this was trap 272's exact inverse — is accurate and is the part I would most want a future reader to see, because it makes the row a record of how the guard came to be real rather than an assertion that it is.

Docs gate re-run at this head: `--ordered --baseline origin/main` → `1..273, unique and ascending, none of the 272 baseline rows deleted`; `check_merge_result.py` → clean merge, gate passes against `origin/main` as of now.

---

## R1. Required before merge — the third site of N6 is still there

`prosper/tools/shader_inspect/skip_survey.py:204-205`, and the same wording at `:66-69`.

The docstring (`:34-37`) and the per-row note (`:184-185`) are both corrected, and the docstring's self-correction is well judged. But the **printed summary** — the text an operator reads at the end of every run — still says:

```
"ALLOWLISTED title a wave-any shader is admitted and logs the admit line instead of a skip"
"line, so that class is invisible here at any route and any window -- and wave-any is the"
```

That prints three lines under a table whose `admitted` column contains exactly that class's count, which is the self-contradiction you fixed in the other two places. `:66-69`'s comment above `NATIVE_VOTE_ALLOWLIST` carries it too ("this tool cannot see that class here AT ALL -- not with a longer window and not with a route").

Reword `:205` to match `:184-185` — "so that class never appears in the REFUSED column at any route or window; it is counted separately as `admitted`" — and the same at `:66-69`.

**Why this is a merge condition and not a rejection, since I rejected twice on claim-vs-code mismatches:** B1 and B4 were claims that could make a reader over-trust — a title read as clean, a guard read as running. This one errs the other way: it tells a reader something is missing that is in fact reported right beside it, so nobody is misled into confidence. The two authoritative sites are already right. It is one string, and I would rather name it precisely than hold the PR.

I do want it on the record that N6 came back reported as taken and was two-thirds done — not as a criticism of the fix, which is otherwise thorough, but because "the note and the docstring now say the class is counted separately" was true and the third site was the one printed on stdout.

---

## Still open by agreement, none blocking

N2 (Windows CRT buffers file-redirected stderr and `TerminateProcess` does not flush it, so trailing skip/admit/fps lines can be lost — direction: undercount), N4 (inherited `PROSPER_*` unrecorded in the log or JSON), N5 (`prosper/tools/shader_inspect/README.md` documents `capture_wave_census.py` at `:145` and `wave_reason_census.py` at `:157` but not this script). Plus a nit: the summary at `:197-198` still prints `args.seconds` as the window even when a title exited early, though the per-row `sec` column and the `EXITED EARLY` note carry the truth.

N1, N3, N7, N8 and B1-B4 are all closed. N8's guard at `:112-119` is right, and the comment naming why the second `wait()` can still time out on Windows is better than the fix I asked for.

---

## Verified at `c8187737`

- `test_skip_survey.py` passes 12/12 standalone; all eight mutations above behave.
- `check_numbered_table.py --ordered --baseline origin/main` green; `check_merge_result.py` green (clean merge, passes against `origin/main` as of now — merge promptly).
- `git diff --check origin/main...HEAD` clean. Three-dot `--numstat`: `14/0`, `1/0`, `219/0`, `134/0` — zero deletions on all four files. Files touched: `CMakeLists.txt`, `GAME_COMPAT_ORCHESTRATION.md`, `skip_survey.py`, `test_skip_survey.py`, and nothing else.
- Zero `Claude-Session:` trailers, zero generated-with attribution lines across `origin/main..HEAD`.
- No emulator behaviour is touched; the CMake change adds a test and nothing else.

Fix R1 and merge.

mattias800:
Merging with `Windows MinGW` red, recorded so the exception is auditable.

**Verified it is not this PR**, by checking the specific test rather than accepting that a red Windows
check must be the known one: `261 - capture_tunables (Timeout)`, 1 of 289, every other test and job
green. Identical on `main`, bisected to #3460, filed as **#3470**.

**289, not 288** — the count went up by one because `skip_survey_patterns_match_emitters` is now
registered, and the CI log shows it running. That is the direct evidence for B4's fix: the test does
not merely exist, CI executes it.

State at merge: head `09502b05`; PR head, remote tip and the approving review's head agree modulo the
R1 string the reviewer authorised as a merge condition. 10 checks pass, 1 fails (#3470), 0 pending.
`test_skip_survey.py` 12/12 locally and under ctest. Numbered-table gate green against `origin/main`
and against the merge RESULT. `diff --check` clean; three-dot numstat reports zero deletions across
all four files. Zero session trailers.

Three review rounds, four blocking findings — B1 (the zero was over-determined and I published the
half that invites a useless repair), B2/B4 (a pattern test that was missing, then present but
registered nowhere), B3 (`frames == 0` suppressing real counts). Five non-blocking items folded in.

PR #3474, opened 2026-09-08
tools(shader_inspect): baseline guard for refused fragment shaders (#3464 W6)

mattias800:
Refs #3464 (W6). Third in the series after #3466 (census) and #3473 (survey).

## Why

*Blasphemous 2* is **rung 6 with a reviewed snapshot guard** and drops **8 fragment shaders** on this
host while that guard stays green. Two more guarded titles do the same. Nothing in this project can
currently see that class of loss — it took a census to find, and a user report before that.

## Deliberately a baseline, not the zero-assertion #3464 proposed

Five titles refuse shaders today, so `assert zero skip lines` would be **red from birth** — a TODO
wearing a guard's clothes, and the first thing anyone would do is disable it.

A baseline locks in today's counts, reddens the moment one rises, and **becomes** the zero-assertion
for free once W5 lands and the counts fall to zero: a data change, not a code change.

## The two properties that carry it

Both are ways this guard could go green while measuring nothing, which is the only interesting
failure mode a guard has.

**A run that cannot be judged FAILS.** The opposite stance from the survey's own reporting, on
purpose: a survey may say "I could not tell", but a guard that cannot tell has not established the
thing it exists for — so passing would let a title that *stopped booting* read as a title with no
dropped shaders. That is instrument trap 273's family exactly. Presenting no frames, exiting before
the window closes, or being absent from the run are all failures.

**A different GPU refuses to compare at all.** These counts are host-dependent in the extreme: on an
AMD host wave64 is native, nothing is refused, and every count is legitimately `0`. An NVIDIA
baseline compared against a Linux/AMD run would report **every shader as fixed and exit 0** — the
most encouraging possible output for a comparison that measured nothing. The Vulkan device is
recorded and checked; an unrecorded device is not a matching device.

A decrease **passes and says so**, because the repair is to re-record the baseline rather than loosen
it.

## Verification

**On live data, not only fixtures:**

```
ARM 1  unchanged run vs its own baseline    BASELINE OK: 3 titles match exactly      exit 0
ARM 2  same run vs a baseline claiming
       PPSA13579 refused 2                  BASELINE FAILED: refuses 8, baseline 2   exit 1
```

Counts reproduced across independent boots (4, 8, 0 both times) and match the W7 survey on #3473.

**17 unit checks** over the comparison logic, registered as ctest
`skip_baseline_refuses_what_it_cannot_compare` — fixtures only, so no dump, no GPU, no boot, and it
runs in CI (verified `ctest -R skip_` green, 2 of 2 with its sibling).

Three of those arms exist because **an earlier draft of this test passed for the wrong reason**: it
gave a single row against a three-title baseline, so `a title that refuses MORE fails` was green
because a title was *absent*, not because a count rose. Each case now uses a complete row set so only
the property under test differs.

## Scope — three titles, not five

Stated plainly because a partial baseline invites the exact misreading this tooling exists to
prevent. It covers `PPSA13579`, `PPSA02664` and `PPSA24651`. **Evergate and Blue Prince refuse
shaders and are not in it.** Under this guard's own rules a baseline title that does not run is a
failure, so adding them commits every check to five boots (~10 minutes).

Better to widen deliberately than to record five and have someone quietly narrow it later because it
is slow. Widening is a `--write-baseline` run, not a code change.

## Risk

Low — a new comparison mode in an offline tool, a fixtures-only test, and a data file. Nothing
existing changes behaviour; no emulator code is touched. The judgement worth review is whether
"cannot be judged ⇒ FAIL" is implemented everywhere it should be, since that is the property whose
absence would be silent.

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -2118,6 +2118,16 @@ if(TARGET prosper_core)
                      "${CMAKE_CURRENT_SOURCE_DIR}/tools/shader_inspect/test_skip_survey.py")
   endif()
 
+  # The baseline comparison's non-obvious verdicts: a run that cannot be judged must FAIL, and a
+  # different GPU must refuse to compare at all. Both are ways the guard could go green while
+  # measuring nothing -- an AMD host refuses no shaders, so an NVIDIA baseline compared there
+  # would report every shader fixed. Pure logic over fixtures: no dump, no GPU, no boot.
+  if(Python3_Interpreter_FOUND)
+    add_test(NAME skip_baseline_refuses_what_it_cannot_compare
+             COMMAND "${Python3_EXECUTABLE}"
+                     "${CMAKE_CURRENT_SOURCE_DIR}/tools/shader_inspect/test_skip_baseline.py")
+  endif()
+
   # spv_validate: emit a module from EVERY SPIR-V-producing entry point and run spirv-val on each —
   # a strict-validation gate (the render tests only prove llvmpipe accepts them; llvmpipe is
   # lenient). No Vulkan needed, but spirv-val (spirv-tools) IS required: it used to report PASS when
--- a/prosper/tools/shader_inspect/skip_baseline.json
+++ b/prosper/tools/shader_inspect/skip_baseline.json
@@ -1,0 +1,8 @@
+{
+  "device": "NVIDIA GeForce RTX 4090 (discrete GPU)",
+  "titles": {
+    "PPSA02664": 4,
+    "PPSA13579": 8,
+    "PPSA24651": 0
+  }
+}
--- a/prosper/tools/shader_inspect/skip_survey.py
+++ b/prosper/tools/shader_inspect/skip_survey.py
@@ -150,6 +150,23 @@ def survey(app, dump, seconds, out_dir):
     }
 
 
+def unjudgeable(row):
+    """Why this row cannot support a verdict, or "" if it can.
+
+    ONE definition, used by both the record and compare sides. It was written twice and drifted
+    within a single commit: the record side omitted the unrecorded-device case, so a run mixing a
+    device-less row with a good one recorded that row as 0 and then refused to compare against the
+    file it had just written. A predicate that decides what counts as evidence belongs in one place.
+    """
+    if row["exited_early"]:
+        return "exited early after %.0fs (rc=%s)" % (row["elapsed"], row["returncode"])
+    if not row["frames"]:
+        return "presented no frames"
+    if not row["device"]:
+        return "the renderer announced no Vulkan device, so its counts cannot be attributed to one"
+    return ""
+
+
 def baseline_from(rows):
     """The baseline a run would record, or (None, why) if the run cannot support one.
 
@@ -158,12 +175,13 @@ def baseline_from(rows):
     run is measured against -- which would then redden honestly on the next good run and be read as a
     regression. The failure direction is benign; the confusion is not.
     """
-    devices = {r["device"] for r in rows if r["device"]}
-    unjudged = sorted(r["title"] for r in rows if r["exited_early"] or not r["frames"])
+    unjudged = ["%s %s" % (r["title"], why)
+                for r in rows for why in [unjudgeable(r)] if why]
     if unjudged:
-        return None, ("cannot record a baseline: %s could not be judged (no frames, or exited "
-                      "early). Recording 0 for a title that never ran would bake a false clean "
-                      "state into the file." % ", ".join(unjudged))
+        return None, ("cannot record a baseline: %s. Recording 0 for a title that could not be "
+                      "judged would bake a false clean state into the file every later run is "
+                      "measured against." % "; ".join(sorted(unjudged)))
+    devices = {r["device"] for r in rows}
     if len(devices) != 1:
         return None, ("cannot record a baseline: this run reports %d distinct Vulkan devices, so "
                       "nothing later compared against it could be trusted." % len(devices))
@@ -229,14 +247,9 @@ def compare_to_baseline(rows, baseline):
                        "that did not run is not a title with no dropped shaders."
                        % (len(missing), ", ".join(missing)))
 
-    uncomparable = []
-    for title in sorted(expected):
-        r = by_title[title]
-        if r["exited_early"]:
-            uncomparable.append("%s exited early after %.0fs (rc=%s)"
-                                % (title, r["elapsed"], r["returncode"]))
-        elif not r["frames"]:
-            uncomparable.append("%s presented no frames" % title)
+    uncomparable = ["%s %s" % (title, why)
+                    for title in sorted(expected)
+                    for why in [unjudgeable(by_title[title])] if why]
     if uncomparable:
         return False, ("could not be judged: %s. A run that never got going has established "
                        "nothing, so this is a FAILURE and not an improvement."
--- a/prosper/tools/shader_inspect/skip_survey.py
+++ b/prosper/tools/shader_inspect/skip_survey.py
@@ -157,6 +157,15 @@ def unjudgeable(row):
     within a single commit: the record side omitted the unrecorded-device case, so a run mixing a
     device-less row with a good one recorded that row as 0 and then refused to compare against the
     file it had just written. A predicate that decides what counts as evidence belongs in one place.
+
+    ADMISSION RULE, because this body answers two different questions and they only happen to agree.
+    The record side asks "can this row DEFINE truth?"; the compare side asks "can it be MEASURED
+    AGAINST truth?". All three conditions below make a row unusable for BOTH, which is what makes one
+    predicate correct today. A condition that disqualifies a row for only one of them does NOT belong
+    here -- it belongs at that caller, or the two sides silently inherit each other's standards.
+
+    Two contract facts both callers rely on: the returned reason is never falsy when there is one
+    (they test `if why`), and it reads as a fragment following a title id (they format `"%s %s"`).
     """
     if row["exited_early"]:
         return "exited early after %.0fs (rc=%s)" % (row["elapsed"], row["returncode"])
--- a/prosper/tools/shader_inspect/skip_survey.py
+++ b/prosper/tools/shader_inspect/skip_survey.py
@@ -83,6 +83,10 @@ SKIP = re.compile(r"\[render\] skip draw=\d+ fs=([0-9a-f]+): fragment shader req
 ADMIT = re.compile(r"\[render\] GTA V native-width fragment vote: subgroup (\d+) -> (\d+) "
                    r"\(why=(\S+)\)")
 FPS = re.compile(r"\[app\] [\d.]+ fps \((\d+) frames")
+# The counts are host-dependent in the extreme: on an AMD host wave64 is native, nothing is refused,
+# and every count is legitimately 0. A baseline recorded on one GPU and compared against another
+# would report every shader as fixed and go green, so the device is recorded and checked.
+DEVICE = re.compile(r"\[render\] Vulkan device: (.+)")
 
 
 def survey(app, dump, seconds, out_dir):
@@ -130,8 +134,10 @@ def survey(app, dump, seconds, out_dir):
     # drawing, and skip lines are emitted from the submit path independently of presentation. It is
     # reported for context and never used to suppress a count.
     frames = max([int(m) for m in FPS.findall(text)] or [0])
+    device = DEVICE.search(text)
     return {
         "title": title,
+        "device": device.group(1).strip() if device else "",
         "frames": frames,
         "elapsed": round(elapsed, 1),
         "exited_early": exited_early,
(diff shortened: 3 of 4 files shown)


mattias800 (commented):
**APPROVED** — nothing blocks. Five findings below, all non-blocking; two of them are one-line
changes I would take before merge if you are touching the branch anyway.

Reviewed at head of `diag/3464-w6-guard` against `origin/main` (three-dot). I ran the new test, ran
`ctest -R skip_ --no-tests=error` (2/2 passed, 0.38 s), branch-traced every arm, and mutation-tested
each guard in a scratch copy. No survey run against real dumps.

## The question that mattered: can it pass while establishing nothing?

Mostly no, and the design is right where it counts. Verified by mutation — each guard deleted in
turn in a scratch copy, then the suite re-run:

| mutation | suite |
| --- | --- |
| `if worse:` (:216) neutered | RED — 3 arms |
| `seen_devices != {want_device}` (:187) neutered | RED — 3 arms |
| `missing` (:195) neutered | RED (KeyError) |
| `r["exited_early"]` (:204) neutered | RED — 1 arm |
| `not r["frames"]` (:207) neutered | RED — 2 arms |
| `if uncomparable:` (:210) neutered | RED — 3 arms |
| `write_baseline` device capture (:155) neutered | RED — 2 arms |
| **`if not want_device:` (:184) neutered** | **GREEN** |

**Your fix to the three previously-wrong arms is real, and no other arm short-circuits.** I
instrumented `compare_to_baseline` to label which `return` each arm reaches and ran the suite: the
branch taken matches the branch named, for all 17. `a title that refuses MORE fails` and
`a title regressing from 0 to 1 fails` both reach `REGRESSION` (not `MISSING_TITLE`), the two
no-frames/exited-early arms reach `UNJUDGEABLE` (and would otherwise reach `BETTER`, i.e. would
genuinely pass — the arms are load-bearing), and `a baseline title missing from the run FAILS`
reaches `MISSING_TITLE` rather than being swallowed by the device check. That is the highest-value
property in this PR and it holds.

---

### 1. `skip_survey.py:179,228` — a baseline with no titles PASSES. Non-blocking; one line.

`expected = baseline.get("titles", {})` defaults to empty, and an empty `expected` walks every guard
untouched to the final `return True`:

```
compare_to_baseline(rows3, {"device": NVIDIA, "titles": {}})
  -> (True, 'unchanged: 0 title(s) match the baseline exactly')
compare_to_baseline(rows3, {"device": NVIDIA})          # "titles" key absent entirely
  -> (True, 'unchanged: 0 title(s) match the baseline exactly')
```

That is a green verdict that established nothing, which is the one failure mode this file exists to
prevent. Reachability is low — `--write-baseline` cannot emit it (`main:311` refuses a device-less
write, and `main` returns 2 before surveying if no dumps matched), so it takes a truncated,
hand-edited, or wrong-file baseline. But the fix is `if not expected: return False, ...` next to the
`not want_device` check, and it costs nothing. Note the message does print `0 title(s)`, so a human
reading stderr has a chance; an exit-code caller does not.

Not a hole, for the record: **empty `rows` is safe** — `seen_devices` is then `set()`, which fails
`:187`. And a malformed count (`"8"`, `null`) raises `TypeError` rather than passing.

### 2. `skip_survey.py:184` — the "baseline with no device" branch has zero test coverage, and its removal is silent. Non-blocking.

This is the GREEN row in the table above. Every arm that looks like it might cover it
(`an unrecorded device refuses to compare`) feeds rows with `device=""` against a baseline that
*has* a device, so it lands on `:187` (`GPU_MISMATCH`), not `:184`. Nothing constructs a baseline
with `device: ""`. Delete `:184` and the suite stays green — and the resulting behaviour is a silent
pass, because `{""} != {""}` is False, so a device-less baseline vs a device-less run would compare
counts and report `unchanged`. It is the only one of the PR body's six enumerated non-comparable
cases that is not pinned. One arm fixes it:

```python
ok, why = skip_survey.compare_to_baseline([row("PPSA13579", 8, device="")],
                                          {"device": "", "titles": {"PPSA13579": 8}})
check("a baseline with no device refuses to compare", not ok, why)
```

### 3. `skip_survey.py:310-313` — "REFUSING to write a baseline" writes the file first. Non-blocking.

`write_baseline` (`:153`) writes to disk unconditionally; `main` inspects the return value
afterwards and only *then* prints the refusal. Confirmed on a scratch run: after the refusal,
`wb2.json` on disk contains `{"device": "", "titles": {"A": 3}}`. So
`--write-baseline prosper/tools/shader_inspect/skip_baseline.json` on a run where the device did not
resolve **clobbers the committed baseline while telling you it refused**. Downstream is loud (a
device-less baseline is refused at `:184`, which is finding 2's branch doing its job) and
`git status` catches it, so severity is low — but an instrument reporting an action it did not take
is the family this project keeps writing traps about. Compute `data` without writing, refuse, then
write.

### 4. `skip_survey.py:153-161` — `write_baseline` does not apply the "cannot be judged" rule that `compare_to_baseline` does. Non-blocking.

The record side is where truth is defined, and it has no frames/exited-early check:

```
write_baseline([row(refused=0, frames=0), row(refused=0, exited_early=True), row("PPSA24651", 0)])
  -> {'device': 'NVIDIA...', 'titles': {'PPSA13579': 0, 'PPSA02664': 0, 'PPSA24651': 0}}
```

A title that never booted is recorded as a title with no dropped shaders — exactly the sentence your
docstring at `:172` forbids on the compare side. The failure direction is benign (a too-low baseline
produces a false RED later, not a false green; a too-high one needs a crashed run emitting *more*
skips than a healthy one), so this is a consistency fix rather than a hole. Refusing to record an
unjudgeable row, or at least naming it in the "wrote baseline" line, would close the asymmetry.

### 5. `skip_survey.py:195` — titles in the run but not in the baseline are silently dropped. Non-blocking; scope, but the message hides it.

`missing` checks baseline−run only. A title refusing shaders that is not baselined is invisible:

```
rows = [PPSA13579:8, PPSA02664:4, PPSA24651:0, PPSA25009:99]  vs the 3-title baseline
  -> (True, 'unchanged: 3 title(s) match the baseline exactly')
```

That is the documented three-title scope and I agree with the reasoning (five boots is a real cost,
and widening is a data change). But the verdict line says how many *matched*, not how many were
*surveyed*, so someone running the full corpus against this baseline reads "unchanged" over a
99-refusal title. Printing `unchanged: 3 of 4 surveyed titles match; PPSA25009 is not in the
baseline` costs one f-string and removes the misread.

---

## The things you asked about that check out

**The GPU guard's exactness is correct, and I would keep it.** `select_vulkan_device`
(`frontends/shared/device/vulkan_device_select.hpp:36`) ranks by device type with no env override
and no index selection, so the string is deterministic per host. A driver rename or a second
discrete GPU changing the selection produces a hard refusal — which is the right answer, because the
counts genuinely are not comparable across either. Your AMD argument is the decisive one and I can
confirm the mutation: with `:187` removed, the AMD arm returns
`improved: PPSA02664 refuses 0, baseline 4; PPSA13579 refuses 0, baseline 8` and exits 0. That is
the "every shader fixed" green you describe, and it is one deleted `if` away. One polish: the
refusal message explains *why* but names no remedy — a trailing "re-record with --write-baseline on
this host" would save the next person a read of the source.

**The DEVICE regex is right and cannot mismatch.** `[render] Vulkan device:` has exactly one emitter
(`tests/fixtures/render_runner.h:1408`, `%s (%s)`, so the captured group is `Name (discrete GPU)`,
matching the committed baseline string). `[compute]` (`live_compute.cpp:3238,3299`) and `[app]`
(`prosper-app/main.cpp:298`) cannot match the `\[render\]` anchor, and `log_vulkan_runtime_device`
prints `Vulkan runtime 1.4, device API ...`, a different shape. All three write stderr, which is the
stream `survey()` captures. `.+` stops at the newline, `search` takes the first, and the emitter runs
unconditionally on renderer init — so a run whose renderer never came up records `""` and is refused.
Two titles disagreeing is caught by `:187` (set inequality, not first-wins).

*Recommendation, not a defect:* `test_skip_survey.py` exists specifically to pin patterns against
their emitters because a silently-stopped pattern is trap 272. `DEVICE` is the one pattern in the
file with no such arm. Its drift is fail-closed (every device becomes `""`, so both compare and
write refuse), so it is not urgent — but `'"[render] Vulkan device: %s (%s)\n"' in rr` is two lines
and puts it on the same footing as SKIP/ADMIT.

**Committing `skip_baseline.json` is the right call.** Nothing runs the guard automatically —
`grep -rn skip_baseline` over the tree hits only `skip_survey.py` and this file: no workflow, no
`add_test` — so it is inert data whose only consumer is a deliberate `--baseline` run, and the
alternative (uncommitted) means there is no baseline to regress against at all. A contributor on
another GPU gets a hard refusal, which is correct. What is missing is anywhere that *says* so:
`README.md` in that folder never mentions `skip_survey.py`, there is no `AGENTS.md`, and the only
documentation of the baseline is argparse help. A short paragraph in the module docstring — whose
host it encodes, that a different GPU refuses by design, and the `--write-baseline` line to
re-record — would be worth more than any of findings 1-5.

**Per-title rather than corpus-total is the right granularity**, and it holds: a 3-fixed/3-broken
swap across two titles reports `REGRESSION: PPSA02664 refuses 8, baseline 4. (also improved: ...)`.
A same-count change of *reason mask* within one title is invisible, since only the integer is
recorded and `reasons` is dropped — worth a thought for a follow-up, not for this PR.

**Nit:** `skip_survey.py:181` — `lines = []` is dead.

mattias800 (commented):
**APPROVED on `dd515f68`.** All five fixes are real and I verified each against behaviour rather
than against the commit message. Nothing blocks. One residual hole worth a one-line fix, and one
straight answer to your (b) that is not hypothetical — the duplicated rule has *already* drifted, and
that drift is the residual.

Re-ran the sweep on the new head: 24 checks green, `ctest -R skip_ --no-tests=error` 2/2.

## What I verified, not took on trust

| finding | check | result |
| --- | --- | --- |
| F1 | `{"device": NV, "titles": {}}` and a missing `titles` key | both now `False`, was `(True, 'unchanged: 0 title(s)')` |
| F2 | both sides unrecorded, the only input that reaches `:217` | now `False`; branch pinned (see below) |
| F3 | write refused over an existing file | sha256 identical before/after — your claim holds |
| F4 | `baseline_from` on a no-frames row | `(None, 'cannot record a baseline: A could not be judged…')` |
| F5 | `[PPSA13579:8, PPSA99999:99]` vs a 1-title baseline | `unchanged: 1 title(s) … NOT IN BASELINE …: PPSA99999 refuses 99.` |

## (c) Branches your sweep still misses — the three record-side ones, which are F3 and F4 themselves

I re-ran the mutation sweep over all twelve conditions on the new head. Nine redden. Three do not:

```
B1  empty-baseline guard  RED(2)   B7  uncomparable return    RED(3)
B2  no-device baseline    RED(1)   B8  worse/REGRESSION       RED(3)
B3  GPU mismatch          RED(3)   B9  unknown-title report    RED(2)
B4  missing baseline title RED     B10 record: unjudged rule  GREEN  <-- F4
B5  exited_early          RED(1)   B11 record: device count   GREEN
B6  no frames             RED(2)   B12 write-validates-FIRST  GREEN  <-- F3
```

Delete F4's `unjudged` rule at `:162`, delete the `len(devices) != 1` refusal at `:167`, or restore
the write-then-validate order in `write_baseline:182`, and the suite stays green in all three cases.
You verified F3 by hand ("byte-identical", and it is) — but a hand-verification is not a pin, and
that is the same shape as F2: behaviour established once, nothing holding it. Three arms close it:
`write_baseline` on a no-frames row returns `None` and leaves an existing file byte-identical; on a
two-device run returns `None`; on a good run writes.

Your sweep is the right durable change and I would keep it exactly as you describe — note only that
it currently covers `compare_to_baseline` and stops at the function boundary, and the record side is
where truth is *defined*.

**One near-vacuous arm, same family.** `test_skip_baseline.py`'s F2 sibling asserts
`"baseline" in why.lower()` — true of 7 of the 8 verdict messages this function can return, including
the `unchanged` one it is meant to exclude. Under the B2 mutation only *one* of the two arms
reddens; that sibling passes on the wrong message. Assert on `records no device` instead. The
property is covered by its partner, so this is a nit, not a hole.

## (a) F5 — report-not-fail matches my intent, keep it

An unbaselined title is unjudgeable in *both* directions: there is nothing to regress from, so an
error would be an assertion the run cannot support. It would also make the guard red the first time
anyone runs the survey with a fourth dump on disk — the "red from birth" failure your own PR body
correctly refuses for the zero-assertion, arriving through the other door. What made the old
behaviour wrong was silence, not leniency, and naming the title with its count on **every** verdict
is exactly the repair. The exit code now means "no baselined title got worse", which is precisely
what the tool claims. I would not tighten it.

## (b) The split is clean; the rule it shares is not

`baseline_from` / `write_baseline` is the right shape — validating into `(None, why)` makes the
refusal impossible to ignore, and it is what makes F3 true rather than incidental.

But the cannot-be-judged **predicate is now written twice, and the two spellings already differ**:

```
compare_to_baseline:235   if r["exited_early"]:  … elif not r["frames"]:
baseline_from:162         r["exited_early"] or not r["frames"]
```

The record side omits a third condition the compare side enforces at `:214-224`: **an unrecorded
device**. `baseline_from:161` filters device-less rows *out of the device set* (`if r["device"]`)
instead of treating them as unrecordable, so:

```
rows = [PPSA13579:8 (NV), PPSA02664:4 (NV), PPSA24651:0 (device="", frames=3000)]
baseline_from -> {'device': 'NVIDIA…', 'titles': {…, 'PPSA24651': 0}}      # accepted
compare_to_baseline(rows, that) -> False, "REFUSING to compare: … this run reports (unrecorded), NVIDIA…"
```

So `--write-baseline X --baseline X` in one invocation writes a file and then refuses to compare
against it, and the file it wrote records `PPSA24651: 0` for a title whose renderer never came up —
F4's stated hazard surviving F4's fix. Reachable: `frames` comes from `[app] … fps` in the app's
present loop while `device` comes from `[render] Vulkan device:` at renderer init
(`render_runner.h:1408`), so an app that presents while the live renderer fails to initialise
produces exactly this row. Severity is low — the compare side is fail-closed and a false `0` yields
a later false RED, not a false green.

One line fixes it and answers your (b) at the same time: hoist the predicate.

```python
def unjudgeable(r):
    """Why this row cannot support a verdict, or "". The ONE definition -- the record side and the
    compare side asked this question in two places and had already diverged on the device."""
    if r["exited_early"]:
        return "exited early after %.0fs (rc=%s)" % (r["elapsed"], r["returncode"])
    if not r["frames"]:
        return "presented no frames"
    if not r["device"]:
        return "recorded no Vulkan device"
    return ""
```

Both callers then read it, the scopes stay legitimately different (compare asks it of baseline
titles, record asks it of every row), and there is one place left to change when a fourth condition
appears.

## Nothing blocks

The residual is one reachable but benign record-side hole and three unpinned branches on a fix
whose behaviour I confirmed by hand. Merge when you are ready; if you take the `unjudgeable` hoist
and the three record-side arms, no re-review is needed from me for that shape — post the sweep
output on the PR and go.

mattias800 (commented):
**APPROVED on `cdc35226`.** Merge it. Both fixes are correct and I re-derived the sweep
independently rather than reading yours: 29 checks green, `ctest -R skip_ --no-tests=error` 2/2 at
`cdc35226`, and 12 of 13 guards redden under mutation.

## The one square your sweep still cannot see, and why

`uncovered: none` is true of every *condition*. F3 is not a condition — it is a **statement
ordering**, and a sweep that neutralises `if` branches structurally cannot reach it:

```
P1  unjudgeable: exited_early  RED(1)   P8  compare: GPU mismatch      RED(2)
P2  unjudgeable: no frames     RED(4)   P9  compare: missing titles    RED
P3  unjudgeable: no device     RED(1)   P10 compare: uncomparable      RED(3)
P4  record: unjudged rows      RED(3)   P11 compare: worse             RED(3)
P5  record: device count       RED(1)   P12 compare: unknown report    RED(2)
P6  compare: empty baseline    RED(2)   P13 write validates FIRST      GREEN  <-- still unpinned
P7  compare: baseline nodevice RED(1)
```

P13 moves the `Path(path).write_text` above the `if data is None: return`. Every condition is
untouched, the suite stays green, and F3 is silently gone — a refused write clobbers the committed
baseline again while printing "REFUSING to write". That is the same class as the boundary blind spot
you just fixed, one level up: last round the sweep stopped at a *function* boundary, this round it
stops at the boundary of what a condition-mutation can express. Behaviour on this head is correct —
I re-verified sha256 identical across a refused write over an existing file — so this is a missing
pin, not a defect. Four lines, and it is the last one:

```python
existing = Path(tmp) / "keep.json"
existing.write_bytes(b'{"device": "prior", "titles": {"X": 7}}\n')
before = existing.read_bytes()
data, why = skip_survey.write_baseline([row("PPSA13579", 0, frames=0)], existing)
check("a refused write leaves an existing baseline byte-identical",
      data is None and existing.read_bytes() == before, why)
```

I confirmed that arm reddens under P13. Add it or don't — it does not block, and I am not asking for
another round either way.

## (b) verified fixed, including the case that started it

```
baseline_from([A(8, NVIDIA), B(0, device="")])
  -> (None, 'cannot record a baseline: B the renderer announced no Vulkan device, …')
```

It can no longer write a file it would then refuse to compare against. Two edge cases I checked
because the `if r["device"]` filter came off the device set at `:180`: **empty `rows`** still refuses
(`0 distinct Vulkan devices`), and a device-less row can no longer reach that set because
`unjudged` rejects it first — so removing the filter was safe, not merely tidy. The near-vacuous arm
now asserts `records no device`, and P7 reddens it.

## Your last question: does `unjudgeable()` want a stronger contract than three arms?

Coverage is not the gap — all three conditions are individually pinned through both callers (P1/P2/P3
above), which is stronger than "three arms" makes it sound. **The risk you have concentrated is
semantic, not coverage,** and it is worth one line of docstring rather than more tests.

The predicate now answers two questions with one body. The record side asks *"can this row DEFINE
truth?"*; the compare side asks *"can this row be MEASURED against truth?"*. All three current
conditions answer both identically, which is exactly why unifying was right and why the drift you
removed was a real defect. But the next condition may not be symmetric — "surveyed for a shorter
window than the baseline was" is a compare-side concern only; "this title is on the native-vote
allowlist, so its refused count structurally under-reports" is a record-side one. Someone adding
either to the shared body would silently change the other caller's verdict, and nothing in the file
would object. So state the admission rule where the next person will read it:

> A condition belongs here only if it makes the row unusable for **both** defining a baseline and
> being measured against one. A condition true of only one side belongs in that caller.

Two smaller contract facts worth the same docstring, both currently honoured by all three returns
and both easy to break: the return is consumed by `if why`, so a reason must never be falsy; and both
callers render it as `"%s %s" % (title, why)`, so it must read as a fragment following a title id —
not `"No frames."` with a capital and a full stop.

That is the whole of it. Nothing blocks; ship it.

mattias800:
Merging with `Windows MinGW` red, recorded so the exception is auditable.

**Verified not this PR**, by test name rather than job colour: `264 - capture_tunables (Timeout)`,
1 of 292, every other test and every other job green. Identical on `main`, which has been red on its
own last three runs independently of this branch. Bisected to #3460, filed as **#3470**.

State at merge: head `a218c138`, matching the branch tip and the approving review's head modulo two
commits of the reviewer's own non-blocking items. 10 checks pass, 1 fails (#3470), 0 pending.
`ctest -R skip_ --no-tests=error` 2/2 locally; suite at 292 against 288 before this series, so all
four new tests are registered and executing in CI.

Four review rounds. Two blocking findings on the first pass, then three rounds of non-blocking items
I took anyway because each was a guard passing while establishing nothing — the defect this PR
exists to prevent. Three of the four rounds found a defect introduced or left by the previous
round's fix, which is the honest record of how this converged:

- F2: an arm named for the no-device branch that landed on the GPU-mismatch branch
- (b): the cannot-be-judged rule drifting between its two copies inside one commit
- (c)/F3: a mutation sweep that stopped first at a function boundary, then at the boundary of what
  a condition-mutation can express — both times leaving the previous round's fix uncovered

All thirteen guards are now pinned, and the one square a condition sweep cannot express (a statement
ordering) is pinned by asserting the effect on disk.

PR #3483, opened 2026-09-08
perf(capture): stream bundle files with checked atomic replacement

mattias800:
Capture bundle writing previously accumulated another complete serialized copy beside the owned chunk store. A delayed file error could also go unnoticed at close and allow truncated output to replace an existing capture.

Stream the unchanged v1/v2 format into an exclusively owned temporary file, computing the historical checksum incrementally. Preserve integrity checks and the 4 GiB file bound. Check writes, flush and close before replacing the destination; failed validation or I/O preserves the previous file and cleans only this writer’s temporary. POSIX and Windows replacement paths are covered by platform-specific guards. This adds normal completion/failure safety, not power-loss durability.

The public-API regression fixture checks frozen format bytes, complete readback, bounded allocation requests, corrupt late references/checksums, replacement/retry, temporary cleanup, real Linux file-size failures, Windows sharing denial and concurrent-write smoke coverage. The same fixture linked against the pre-change core passes format checks and fails the allocation/failure guards.

Release build and 447/447 tests pass. All 64 selected Vulkan synchronization-validation cases pass with zero messages after layer-load and deliberate-hazard controls. Independent source, fixture and completed-measurement reviews approve the exact head. In three alternating 512 MiB writer experiments, old peak RSS grows about 1 GiB; streamed-writer peak growth is 0.25–0.50 MiB. All six files have identical size and SHA-256. Wall time varies, so no stable speedup is claimed. [Commands, controls, regression allocation totals and reproducible memory experiment](https://github.com/mattias800/prosper/pull/3483#issuecomment-5593117525).

Completed Sonic/GTA captures retain the same normalized routes/environment, immutable binaries and complete F8 records. GTA’s bank lighting/geometry remains consistent; Sonic’s known black-world/HUD and title defects remain. Graphics rates are roughly unchanged. Both candidate main outputs have only initial-interval input-demand shortfalls, with none later during capture, versus writer-associated deficits in the retained baselines. These single pairs do not establish an audio fix or hardware-XRUN improvement; repeated controls remain #3435. Every output/channel retains its independent PCM buffer. [Game evidence](https://github.com/mattias800/prosper/pull/3483#issuecomment-5593279462), [per-output audio evidence and limits](https://github.com/mattias800/prosper/pull/3483#issuecomment-5593279665).

Fixes #1648. Related to #3435 and the capture tooling used for #3407.

PR #3484, opened 2026-09-09
fix(gpu): validate DCC metadata before borrowing retained images

### LINKED ISSUE #3482: Retained compute-image borrowing must validate independently changed DCC metadata
A retained image's pixels can stop representing a compressed guest surface after an independent write to its DCC metadata, even when its base bytes remain unchanged. The public `import_live_compute_storage_image` API and the underlying graphics/compute-transfer borrow checks validate the base span only. Metadata address, interpretation and content are not part of that borrow proof.

This is a **source/contract finding on main `973ce44ed113fc808f12dca5ed62156f0cf10b8d`**, also present in the current #3481 continuation. The focused regression below has not been executed for this report. No Sonic symptom, title-specific regression, performance gain, or arbitrary DCC compression code is attributed to it.

## Source and existing protections

- [`ComputeImageCacheKey` and `storage_image_cache_key`](https://github.com/mattias800/prosper/blob/973ce44ed113fc808f12dca5ed62156f0cf10b8d/prosper/frontends/shared/live/live_compute.cpp#L977) identify the pixel allocation/view but omit DCC metadata identity and content.
- [`borrow_cached_image_for_graphics`](https://github.com/mattias800/prosper/blob/973ce44ed113fc808f12dca5ed62156f0cf10b8d/prosper/frontends/shared/live/live_compute.cpp#L2654) and [`borrow_cached_image_for_compute_transfer`](https://github.com/mattias800/prosper/blob/973ce44ed113fc808f12dca5ed62156f0cf10b8d/prosper/frontends/shared/live/live_compute.cpp#L2749) query the journal over `key.gpu_addr/key.guest_bytes`; page watches and the Windows exact mirror cover that same pixel span. A separately allocated metadata plane is outside those dependencies.
- [`import_live_compute_storage_image`](https://github.com/mattias800/prosper/blob/973ce44ed113fc808f12dca5ed62156f0cf10b8d/prosper/frontends/shared/live/live_compute.cpp#L12004) accepts the sampled descriptor but does not independently validate compression metadata before returning the lease.
- **The current live graphics caller is protected:** `live_renderer.cpp` computes `persistent_dcc_uncompressed` from a complete fresh all-`0xff` metadata scan and requires it before calling this importer for compressed surfaces. This report does not claim that caller consumes the stale lease in the sequence below.
- **The ordinary supported RGBA16F sampled-compute clear is also protected:** [`compute_sampled_dcc_fast_clear_rgba8`](https://github.com/mattias800/prosper/blob/973ce44ed113fc808f12dca5ed62156f0cf10b8d/prosper/frontends/shared/live/live_compute.cpp#L131) runs before sampled-cache/transfer admission. Its supported uniform fast-clear result bypasses those caches. Thus the concrete RGBA16F clear below is an importer-contract regression, not evidence that this ordinary compute consumer currently renders stale pixels.
- The broader compute sampled path normally does not perform an all-`0xff` check; it does so only for the cache-disable control or tracing. The low-level transfer proof remains pixel-only. Outside the supported fast-clear subset, a real semantic counterexample must establish the descriptor and DCC interpretation; merely choosing an arbitrary non-`0xff` byte is not sufficient. General unsupported DCC decoding remains #2677.

## Concrete valid regression to add

1. Produce and retain a nonblack native RGBA16F storage image B through the production backend. Use an ordinary single-mip 2D descriptor, a supported tiled DCC footprint, and `alpha_is_on_msb=true`. Keep its metadata in a distinct allocation/page from B's pixels. Verify successful pixel writeback/publication, a complete all-`0xff` metadata plane, and a successful initial public import.
2. In a **later separate ordered operation**, write `0x40` across the entire checked metadata footprint and notify that actual metadata range. Do not touch B's pixel allocation. This is the supported uniform embedded clear represented by `gfx10_dcc_fast_clear_rgba8`: black RGB with alpha 1. It is not an arbitrary compression code. Release the initial import lease before the next query.
3. Ask `import_live_compute_storage_image` for B again within valid ordered history. Its current checks see unchanged pixel bytes and can return the old nonblack Vulkan image. The required contract is refusal, or an image representing the new logical clear; successful import of the old pixels must fail the test. Keep full B/base-byte and metadata-tail assertions, and use an independent GPU read of an accepted lease to expose the stale image rather than relying only on a counter.
4. Preserve positive controls: all-`0xff` metadata with unrelated writes still permits the original exact import; current ordinary RGBA16F sampled compute materializes black/alpha1 after the clear; the live graphics caller's DCC guard refuses the old direct import. Include incomplete/unavailable metadata, changed metadata backing/interpretation, and effective hosted metadata as bounded failure cases when implementing the shared contract.

The distinction from #3476/#3481 is the **later independent metadata writer**: publication-time overlap analysis sees the resources participating in that dispatch. It cannot make a pixel-only watch/journal/mirror observe every future write to a separate metadata plane. Existing same-dispatch metadata-output conflict guards should remain intact.

## Fix direction

Make metadata authority explicit at the shared borrowing boundary. Either require and verify a current complete uncompressed-plane proof for compressed borrowers, or retain a correctly scoped metadata dependency alongside pixel authority. Account for advertised and effective hosted metadata spans, format/alpha interpretation, unavailable history, Windows mirrors and Linux watches. Unknown metadata must not authorize old storage pixels. Preserve pins and allocation ownership while invalidating content authority.

Do not indiscriminately disable sampled decode caching: replaying the same base-byte decode is distinct from borrowing a previously produced storage image, and the existing supported fast-clear bypass already handles its narrow case. Establish the demonstrated importer failure first, then extend the common proof to transfer callers without claiming a fix for unsupported DCC decoding.

Tracker check: no separate metadata-dependency issue found in open/closed searches for DCC metadata, metadata cache and retained metadata. Related: #3476 (overlapping publication), #2677 (general DCC decode), #2674 (DCC/HTILE classification), closed #3149 (sampled cache admission rationale), #3407 (graphics performance).

mattias800:
A later write to a separately allocated DCC metadata plane can change a surface’s logical pixels while its base bytes remain unchanged. Retained compute-image borrowing previously checked only those base bytes and could return the previous GPU image through the public importer.

Both graphics and compute-transfer borrowing now require a fresh, complete plain-metadata proof for compressed consumers, including numeric aliases. Admission checks the original consumer’s supported shape and typed DCC correlation, respects hosted metadata bounds, and declines missing, truncated, unsupported or non-plain metadata. Existing pixel authority checks and outstanding image leases remain intact. Metadata refusal has its own diagnostic outcome rather than being reported as a pixel-journal failure.

The new GPU regression produces a magenta RGBA16F storage image, then changes only its separate metadata plane to the supported black/opaque clear. The old production implementation returns visibly stale magenta through the public importer; the fix refuses that lease while ordinary sampled compute still materializes the correct clear. Additional controls cover numeric aliases, complete metadata tails, hosted truncation, unknown/HTILE classification, descriptor changes, unrelated writes and restored plain metadata. Existing compressed-handoff fixtures explicitly identify their owned DCC planes and retain successful handoff assertions.

Production-code validation on `f36fb4238f380cc03d80327a18c079ad63c6c53f` (final follow-up changes documentation only):

- Full Release build and 448/448 CTest passed (116.15s).
- Strict Vulkan synchronization scan: 20/20 passed, zero messages with an empty allowlist. The loader and deliberate write-after-write controls both worked.
- Rebuilding the previous `47aa2ca1` live-compute implementation against the current test ABI produced 35 intended failures, including GPU-observed stale magenta. Restoring the implementation passed with the identical fixture object.
- Production Sonic/GTA captures completed. GTA presentation rate was essentially unchanged; two Sonic comparisons recorded lower candidate presentation counts at different guest timer states. Both adverse comparisons are retained.
- A reviewed private F8 observer then measured 1,903 Sonic borrow checks, all uncompressed, totaling 0.054407 ms inside the guards; no metadata classification, resolution or hosted-plane lookup occurred at the instrumented boundaries. In the same window, 132 scaled RTT snapshot materializations took 103.240 ms. This separates the expensive path in that diagnostic run; it does not prove the earlier differences were noise or establish performance neutrality.
- The final change after the tested production revision is documentation only: narrow the historical metadata-independence claim to ordinary sampled decoding. Private observers were removed from production sources.

Commands: `ctest --test-dir prosper/build-linux --no-tests=error --output-on-failure -j1`; focused sync via `python3 prosper/tools/vkval/vk_validation_scan.py --build-dir prosper/build-linux --sync --allowlist /dev/null --ctest-arg=-j1 --ctest-arg=-R --ctest-arg='^(storage_metadata_borrow|storage_output_conflicts.*|storage_rtt_.*|storage_readonly.*|near_full_storage_.*|renderer_uint_copy|compute_image_borrow_census|compressed_source_authority|metadata_kind_correlation)$'`.

[Initial game comparison](https://github.com/mattias800/prosper/pull/3484#issuecomment-5594116444) and [fresh diagnostic pair](https://github.com/mattias800/prosper/pull/3484#issuecomment-5594550470). Sonic performance variability and repeated RTT materialization remain under #3407. Newly rendered frame counts are unavailable, and presentations are not counted as new renders.

This fixes the shared borrowing contract. The current graphics caller and supported RGBA16F sampled-compute clear already had their own protections. It does not implement arbitrary DCC decompression, fix Sonic’s title/cloud or black-world defects, or establish an FPS improvement. Unclassified/unsupported compressed surfaces conservatively fall back; NVIDIA/Windows runtime execution is not locally verified.

Fixes #3482. Related to #3407; its remaining performance scope stays open.

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -3126,6 +3126,14 @@ if(TARGET prosper_core)
       target_link_libraries(test_renderer_uint_copy PRIVATE prosper_live_renderer prosper_core Vulkan::Vulkan)
       add_test(NAME renderer_uint_copy COMMAND test_renderer_uint_copy)
 
+      add_executable(test_storage_metadata_borrow tests/gpu/execute/test_storage_metadata_borrow.cpp)
+      target_include_directories(test_storage_metadata_borrow PRIVATE tests frontends)
+      target_link_libraries(test_storage_metadata_borrow PRIVATE prosper_live_renderer prosper_core Vulkan::Vulkan)
+      add_test(NAME storage_metadata_borrow COMMAND ${CMAKE_COMMAND} -E env
+               PROSPER_COMPUTE_IMAGE_CACHE_MIN_KB=1 PROSPER_COMPUTE_STORAGE_IMAGE_CACHE_MIN_KB=1
+               $<TARGET_FILE:test_storage_metadata_borrow>)
+      set_tests_properties(storage_metadata_borrow PROPERTIES TIMEOUT 60)
+
       add_executable(test_storage_output_conflicts tests/gpu/execute/test_storage_output_conflicts.cpp)
       target_include_directories(test_storage_output_conflicts PRIVATE tests frontends)
       target_link_libraries(test_storage_output_conflicts PRIVATE prosper_live_renderer prosper_core Vulkan::Vulkan)
--- a/prosper/docs/RESOURCE_BINDING.md
+++ b/prosper/docs/RESOURCE_BINDING.md
@@ -353,20 +353,25 @@ resource" and "it saw it and rejected it" produce identical evidence while point
 files. Reach for it before concluding anything about descriptor *recovery*: on Earthion (#1590) it
 overturned exactly that conclusion in one run.
 
-**A DCC-compressed SAMPLED surface's cache identity does NOT include its metadata plane, and folding
-that plane into the key is both unnecessary and — as first attempted — ineffective.** #3150 /
-#3149, 2026-08-31. The persistent compute image cache excluded compressed sampled surfaces on the
-theory that base bytes alone could not determine their content; on Stray that exclusion re-detiled
-one 4K FP16 surface 3,860 times in 110 s (98.9% of all tiling work, `PROSPER_TILECENSUS`). The
-theory is wrong for the *sampled* path: the DCC plane's only consumer there is
-`compute_sampled_dcc_fast_clear_rgba8`, whose materialization the admission gate already excludes
-via `!sampled_dcc_fast_clear`, so a cacheable entry's upload is detile + unpack of the BASE bytes
-alone — which the existing validation already covers. More sharply, `gfx10_dcc_fast_clear_rgba8` is
-the **only** DCC decode function in the tree: prosper has no DCC decompressor, so
-metadata-independence is a property of the codebase rather than of one call graph. **Do not re-derive
-the metadata-in-the-key design**: the first revision of #3150 did, and it was rejected twice over —
-the guard was conjoined into only one of three skip proofs (`watch_unchanged` and `exact_unchanged`
-both examine the base range alone, so the upload was skipped anyway), and the premise was unnecessary
-regardless. The *storage* gate is different and still requires an all-`0xff` plane: a storage
-target's base bytes are not authoritative while a live plane exists. `PROSPER_NO_DCC_IMAGE_CACHE=1`
-restores the old exclusion for bisection.
+**The ordinary sampled base-byte decode cache does not need DCC metadata in its identity;
+retained storage-image borrows require a separate metadata proof.** #3150 / #3149,
+2026-08-31; scope clarified by #3482 / #3484, 2026-09-09. The persistent compute image cache
+excluded compressed sampled surfaces on the theory that base bytes alone could not determine
+the cached upload; on Stray that exclusion re-detiled one 4K FP16 surface 3,860 times in 110 s
+(98.9% of all tiling work, `PROSPER_TILECENSUS`). For the ordinary sampled decode, the current
+supported fast-clear probe runs before reuse and its materialization is excluded by
+`!sampled_dcc_fast_clear`. A cacheable entry therefore serves detile + unpack of the BASE bytes,
+which the existing base-range validation covers. This establishes equality with the uncached
+base-byte fallback, not correct decoding of arbitrary compressed metadata or metadata-independence
+of every image source.
+
+The first metadata-in-the-key revision of #3150 was ineffective too: it guarded only one of three
+skip proofs, while `watch_unchanged` and `exact_unchanged` still inspected only base bytes. Do not
+repeat that design for this ordinary decode cache. A retained STORAGE image is a different source:
+it holds completed producer pixels. Both graphics import and compute storage-to-sampled transfer
+must validate the producer's pixel authority **and** prove the original consumer's current complete
+metadata plane plain before borrowing. A separate metadata allocation may change while base bytes,
+page watches and exact mirrors remain unchanged; a producer-format alias retry must not bypass a
+failed metadata proof. Unsupported or incomplete metadata footprints decline the borrow. The
+storage acquisition gate also keeps its own all-`0xff` requirement. `PROSPER_NO_DCC_IMAGE_CACHE=1`
+restores the historical sampled-cache exclusion for bisection.
--- a/prosper/frontends/shared/compute/compute_image_borrow_census.hpp
+++ b/prosper/frontends/shared/compute/compute_image_borrow_census.hpp
@@ -130,6 +130,7 @@ enum class ComputeImageBorrowOutcome : uint8_t {
     NoImage,             // entry without a VkImage (allocation released)
     AuthorityChanged,    // published, but no proof says guest memory still matches it
     Hit,
+    MetadataUnproven,    // pixel authority passed; current compression metadata did not
     Count,
 };
 
@@ -142,6 +143,7 @@ constexpr const char* compute_image_borrow_outcome_name(ComputeImageBorrowOutcom
     case ComputeImageBorrowOutcome::NoImage:             return "no_image";
     case ComputeImageBorrowOutcome::AuthorityChanged:    return "authority_changed";
     case ComputeImageBorrowOutcome::Hit:                 return "hit";
+    case ComputeImageBorrowOutcome::MetadataUnproven:    return "metadata_unproven";
     case ComputeImageBorrowOutcome::Count:               break;
     }
     return "?";
@@ -505,7 +507,7 @@ private:
 // a saturated census fits, so the marker should never fire -- but a marker that never fires is
 // cheap and a missing line that reads as a zero is not.
 // Worst case, measured by the saturation arm in the test rather than estimated: eleven decline
-// buckets, six outcome buckets, seven publish buckets, twenty-three key-field names and twenty-two
+// buckets, seven outcome buckets, seven publish buckets, twenty-three key-field names and twenty-two
 // scalars, every one of them a twenty-digit u64. 4 KiB clears it with room to spare.
 inline constexpr size_t compute_image_borrow_census_report_bytes = 4096;
 
--- a/prosper/frontends/shared/live/AGENTS.md
+++ b/prosper/frontends/shared/live/AGENTS.md
@@ -146,6 +146,14 @@ snapshots naturally. Metadata-only writes require an explicit retention veto bec
 does not cover interpretation metadata. Keep sampled metadata dependencies separate from storage
 metadata reset obligations. Include effective hosted and advertised ranges, and fold exact aliases.
 
+Later dispatches can change a separate metadata plane without touching a retained image's pixel
+allocation. Graphics imports and compute transfer seeds therefore recheck the consumer's complete
+supported DCC plane immediately before borrowing. Unknown metadata kinds, unsupported footprints,
+unreadable or truncated planes and non-plain contents decline the borrow. A supplied hosted plane
+never falls back to guest metadata. This fresh check applies to numeric aliases too; neither an
+earlier all-0xff scan nor the pixel journal proves current metadata. Refusal preserves outstanding
+leases and the existing materialization fallback. `storage_metadata_borrow` guards this boundary.
+
 Zero-address internal backing (such as GDS) is not an architectural guest output.
 Notify actual hosted guest destinations as well as advertised architectural ranges, including for cached
 views absent from a dispatch. Prepare page watches before every actual mutation. The production
--- a/prosper/frontends/shared/live/live_compute.cpp
+++ b/prosper/frontends/shared/live/live_compute.cpp
@@ -1169,6 +1169,7 @@ enum class ComputeTransferBorrowResult : uint8_t {
     NoCache,
     InvalidCache,
     AuthorityChanged,
+    MetadataUnproven,
 };
 
 bool persistent_compute_buffer_enabled(uint32_t bytes) {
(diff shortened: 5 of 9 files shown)


mattias800 (commented):
APPROVED — exact head f36fb4238f380cc03d80327a18c079ad63c6c53f.

Independently reviewed the production borrowing changes: the original consumer’s complete metadata proof runs after pixel authority and before acquiring a pin or refreshing cache state, including numeric aliases. Hosted truncation cannot fall back to guest metadata, and refusal preserves existing leases. Metadata refusal remains separate from pixel journal/watch diagnostics.

Audited the retained verification: 448/448 tests passed; 20/20 strict core/synchronization cases passed with zero messages and successful layer-loader and deliberate-WAW controls. The old production source against the same fixture object produced 35 failures, including GPU-observed stale magenta and the compute-transfer admission guard; source and object hashes match the recorded control and restored implementation. I authored fixture/census portions, so their independent source review belongs to the other reviewer.

This approval covers production source and inspected test evidence. Comparable game/runtime evidence and CI remain separate merge gates; no performance improvement or NVIDIA/Windows runtime result is approved.

mattias800:
Initial comparable game captures on exact `f36fb4238f380cc03d80327a18c079ad63c6c53f` are complete. The seven requested Linux/general checks are now green; **merge is held for a fresh Sonic comparison**, not Windows/macOS.

Both normal captures used the retained #3483 protocol: fresh save/cache, committed pad routes (GTA Performance Story), one complete F8 window with 20 pre/21 post samples and zero record drops, one F9 BMP/completed bundle, isolated earlier CPU/GPU profiler window, and clean recorded process census. Environment and route identities match the archived `7d9c13b4059ffbea43b5dd83ebc33077c7c84abc` baseline; its tree is identical to starting main `47aa2ca1`.

| F8 observation, baseline → candidate | Sonic | GTA |
| --- | --- | --- |
| Window seconds | 5.020641 → 5.016301 | 5.016949 → 5.017290 |
| Guest presentations | 39 → 36 | 31 → 31 |
| Host presentations | 38 → 37 | 31 → 31 |
| Renderer / compute records | 628 / 1341 → 609 / 1262 | 800 / 2145 → 800 / 2160 |
| Graphics milliseconds | 2005.429 → 2013.316 | 2649.521 → 2661.516 |
| Compute milliseconds | 1689.112 → 1657.631 | 1237.851 → 1242.649 |
| Compute wait / writeback milliseconds | 542.946 / 685.366 → 518.003 / 676.914 | 410.470 / 446.606 → 416.221 / 441.805 |
| Bracketed compute GPU milliseconds | 383.638 → 367.157 | 305.119 → 308.016 |

GTA’s bank scene retains visible geometry, glass, signage, lighting and HUD with different character animation. Sonic retains the known black Cyber Space world behind its HUD (#2790), with timer/gauge state differing. Original BMPs were inspected through a lossless pixel-verified PNG conversion and retained. Neither pair establishes pixel identity or newly rendered FPS; F8’s newly rendered rate remains null.

Sonic’s guest presentation rate is lower by 7.61% in this short window, while dispatch counts also differ and several per-dispatch costs rise. Texture resource accounting includes 84.144ms charged to persistent invalidation versus none in the baseline; that bucket includes the entire resource path, not only invalidation or the new metadata scan. Its selected compute handoff remains 1292/1292 successful with zero metadata refusals, but that is not the complete importer population. **These observations are inconclusive for regression attribution.**

A fresh preselected Sonic baseline/candidate pair will enable the global borrow census in both arms, retain the same route/timers and immutable binaries, and check actual census population. That diagnostic also enables near-miss scanning, so it will be identified as a matched diagnostic pair, not zero-overhead normal gameplay. This initial result will remain in the record. No speedup or blanket no-regression claim is made.

Candidate Sonic emitted one graphics submit=-4 diagnostic only after shutdown began. Existing shutdown admission can return this value without submitting to the driver; its log alone does not distinguish that refusal from a driver return. No kernel reset was observed in the checked interval. Earlier F8/F9 captures completed normally.

mattias800:
The fresh matched Sonic diagnostic pair still leaves the performance question open. Both arms used the same route, fresh save/cache directories, capture timing, and `PROSPER_COMPUTE_BORROW_CENSUS=1`; that census also enables near-miss cache scans, so this is a diagnostic comparison.

Baseline `7d9c13b4` → candidate `f36fb423`: over 5.0177 → 5.0195 seconds, guest presentations were 38 → 35 and host presentations 38 → 34. Newly rendered frame counts remain unavailable. The final global census had 195,216 → 189,652 accepted imports, with zero candidate metadata refusals in every emitted snapshot.

The extra F8 resource cost was concentrated in the RTT texture class: 3.50 → 233.88 ms, with 138 candidate renderer records exceeding 1 ms in that class. This class measures complete resource preparation, not an isolated copy. The separate CPU profile also sampled `inject_rtt_pixels`, which can be called for valid scaled snapshot-cache misses; it does not establish a failed borrow. The two captures reached different guest timer states, and neither is a pixel-identical scene comparison.

All seven Linux/general CI checks passed. Merge remains on hold while a bounded observer distinguishes successful metadata-check cost from repeated RTT materialization. The zero refusal count alone does not prove that successful checks are free, and no speedup or no-regression conclusion is claimed.

mattias800 (commented):
APPROVED — exact head `f9175d22b1ba7cd5b20bdcb7b3c1771500c11950`.

I verified the remote branch and PR head, the clean tracked tree, and the complete delta from reviewed `f36fb4238f380cc03d80327a18c079ad63c6c53f`: only `RESOURCE_BINDING.md` changed. Production sources and fixtures are unchanged, and the private observers are absent. The documentation now correctly limits the ordinary sampled decode-cache claim and separately requires current metadata authority for retained storage borrows. The revised PR body preserves that scope.

My prior independent production review and reviewed negative-control evidence still apply. The full 448-test and strict 20-test synchronization results belong to the unchanged production code at `f36fb423`, not a newly executed full suite at this documentation head. I contributed fixture/census changes; their independent review is covered separately.

I also reviewed the retained Sonic comparisons and bounded private diagnostic. The lower candidate presentation counts remain adverse observations at differing guest timer states. In the private diagnostic window, all 1,903 borrow checks were uncompressed (0.054407 ms inclusive), with no metadata classification/resolution or hosted source, while 132 new scaled RTT views consumed 103.240 ms. This identifies the expensive path in that run; it does not establish the earlier path populations, dismiss those comparisons as noise, or prove performance neutrality. These limits are explicit in the PR, and I find no remaining runtime-evidence blocker to this bounded correctness fix. Remaining RTT/performance work stays under #3407; new rendered-frame counts and NVIDIA/Windows runtime parity are not established.

Final restored-build/fixture confirmation and required CI on this exact head remain author merge gates. No new builds or GPU workloads were run for this review.

mattias800 (commented):
APPROVED — exact head f9175d22b1ba7cd5b20bdcb7b3c1771500c11950.

I independently reviewed the production borrowing contract and regression fixtures, including the real GPU stale-color oracle, numeric aliases, hosted metadata bounds, complete-plane tails, outstanding leases, and the compute-transfer positive/negative admission control. I verified that the entire delta from reviewed f36fb4238f380cc03d80327a18c079ad63c6c53f is RESOURCE_BINDING.md only; production and test sources are unchanged and private observers are removed. The documentation now correctly separates ordinary base-byte decode-cache equivalence from retained producer-image authority. The current PR body accurately limits the fix to that shared borrowing boundary.

Validation revisions are explicit: full 448-test and strict 20-case synchronization results belong to the unchanged production code at f36; the restored final-head log records 4/4 focused cases passing. No new builds or tests were run for this final review. Prior old-production controls establish the defect; the metadata proof is not merely a metadata-key comparison or a synthetic refusal counter.

I independently reconciled the private diagnostic's raw rows, captured binary/route identities, reconstructed frozen source patch, process census and complete F8 footer. All 1,903 observed borrow checks were uncompressed, totaling 0.054407 ms inside their bodies; the 155 sampled metadata lookups were also uncompressed. No classification or metadata scan ran at those observed boundaries. Separately, 132 scaled RTT materialization misses cost 103.240005 ms. This rules out expensive successful metadata scanning for that interval, not the earlier comparisons or all timing/history effects. The earlier adverse Sonic presentation observations remain retained and inconclusive; neither performance neutrality nor a newly rendered FPS improvement is established. Observer/callback boundaries differ from the footer record population. The known black-world HUD remains, and no title/cloud fix, arbitrary DCC decompression, audio improvement, or local NVIDIA/Windows runtime parity is claimed.

I find no remaining source or inspected runtime-evidence blocker to this bounded correctness fix. Required exact-head CI and latest-main freshness remain separate author merge gates. Remaining RTT/performance investigation stays under #3407.

mattias800:
Final head: `f9175d22b1ba7cd5b20bdcb7b3c1771500c11950`. The complete change after previously tested `f36fb4238f380cc03d80327a18c079ad63c6c53f` is the documentation scope correction in `RESOURCE_BINDING.md`. The private observer changes were reversed; production source bytes match the reviewed implementation.

The production app and borrow fixture rebuilt successfully. A focused restoration check passed **4/4** (`storage_metadata_borrow`, `compute_image_borrow_census`, `compressed_source_authority`, `metadata_kind_correlation`); the local docs gate and `git diff --check` passed. The earlier **448/448 full tests**, **20/20 strict synchronization cases** with zero validation messages, and old-production-failing GPU controls apply to the unchanged production implementation.

The separate private diagnostic completed its F8/F9 protocol with no dropped records and a clean process census. At its instrumented boundaries, all 1,903 borrow guards and 155 sampled lookups were uncompressed; no metadata classification/resolution or hosted-plane path ran. Guard body timers totaled **0.054407 ms**, excluding observer bookkeeping. RTT helper calls included 132 newly scaled views (**103.240005 ms**) and 141 reused scaled views (**0.014715 ms**). Most scaling cost was 93 conversions from 3840×2160 to 1280×449. These sums observe entered operations and may extend past F8 closure; they are not exactly coextensive with the 5.0166-second reporting interval or a whole-frame subtraction. The visual capture retains the known black-world/HUD state at a different guest timer.

Independent source and raw-evidence reviews support merging the bounded correctness fix. The earlier lower candidate presentation counts remain part of the record: this diagnostic neither proves they were noise nor establishes general performance neutrality. Repeated RTT materialization and comparable performance work remain under #3407; no further identical-wall-time pair is required to establish this guard’s correctness. Required CI is being checked on this final head before merge.

PR #3490, opened 2026-09-09
fix(audio): --volume is an amplifier after the guest's mixer, not a rival writer of it

### LINKED ISSUE #3489: prosper-app: --volume is silently undone by the guest's own sceAudioOutSetVolume
## The symptom

`prosper-app --volume 0` prints `[app] audio volume 0%` and the title is audible anyway. Reported by
the project owner during a meeting, on a headless survey run that passes `--volume 0` on every launch.

## The cause

Both volumes write the **same SDL knob**, so whichever runs last wins.

`--volume N` reaches `Sdl3AudioSink::set_gain`, which writes `SDL_SetAudioStreamGain` on every open
stream. The guest's `sceAudioOutSetVolume` handler, `Sdl3AudioSink::set_volume`, then writes
`SDL_SetAudioStreamGain` on the same stream with the title's own volume:

```cpp
// frontends/audio_sdl3/audio_sink_sdl3.cpp, before the fix
void set_volume(int port, uint32_t mask, const int* vols) override {
    int maxv = audio_peak_channel_volume(mask, vols);
    float gain = maxv / 32768.0f;
    std::lock_guard<std::mutex> lk(mx_);
    if (SDL_AudioStream* st = slots_[port - 1].stream) SDL_SetAudioStreamGain(st, gain);
}
```

So the host setting holds only until the title first sets its own volume, which every title does
within seconds of booting. The flag is not ignored -- it works, and is then discarded.

## Why it looked fine

`[app] audio volume 0%` reports the value that was *requested*. Nothing read back what the device
actually held, so the log was true about the intent and false about the output for the whole life of
the flag.

## The model

These are two different points in the signal chain, not two writers of one value:

- `sceAudioOutSetVolume` is the guest setting a channel on its virtual PS5 mixer -- per port.
- `--volume` is the amplifier that mixer feeds -- one knob, after the mix, owned by the host.

SDL has both natively: `SDL_SetAudioStreamGain` per stream and `SDL_SetAudioDeviceGain` on the
logical device, and its documentation states the ordering -- *"This is applied, along with any
per-audiostream gain, during playback to the hardware."* Expressing the chain that way is better than
multiplying the two into one knob, because the guest path then never names a device and cannot
reintroduce the overwrite.

## Verified

45 s run of `PPSA02664` with `--volume 0` on the fix:

```
[app] audio volume 0%
prosper-audio: port 1 amplifier gain set 0.00, device reports 0.00
prosper-audio: port 2 amplifier gain set 0.00, device reports 0.00
prosper-audio: port 3 amplifier gain set 0.00, device reports 0.00
```

The read-back is part of the fix: a gain nobody reads back is how this survived.

## Impact beyond the annoyance

Every headless route run in this repository launches with `--volume 0` and assumes it works. It also
means any past measurement taken "silently" was running a full audio mixer path it was not meant to
be exercising.

mattias800:
Fixes #3489.

## The failure

`prosper-app --volume 0` prints `[app] audio volume 0%` and the title is audible anyway. Found by the
project owner, who was in a meeting while a headless survey ran four titles.

Both volumes wrote the **same SDL knob**, so whichever ran last won. `--volume N` reaches
`set_gain()`, which writes `SDL_SetAudioStreamGain` on every open stream; the guest's
`sceAudioOutSetVolume` handler then writes the same knob with the title's own volume. The host
setting held only until the title first set its own, which every title does within seconds of
booting. The flag was not ignored — it worked, and was then discarded.

## The contract

These are two points in one signal chain, not two writers of one value:

- **`sceAudioOutSetVolume`** — the guest setting a channel on its virtual PS5 mixer, per port.
- **`--volume`** — the amplifier that mixer feeds. One knob, after the mix, owned by the host.

Scope, stated precisely because the wording above invites a wider reading: this is the amplifier on
the **guest's** audio-out ports. The launcher's own music plays through a separate path that never
reaches this sink, so `--volume 0` silences the title and not the launcher. Whether the flag should
mean "guest" or "the whole process" is a design question rather than part of this fix, and is #3499.

SDL has both natively and states the ordering in its own documentation for `SDL_SetAudioDeviceGain`:
*"This is applied, along with any per-audiostream gain, during playback to the hardware."* So the
guest keeps `SDL_SetAudioStreamGain` and the host moves to `SDL_SetAudioDeviceGain` on the logical
device.

Multiplying the two into one knob gives the same numbers, and I wrote that first. It is worse:
the ordering becomes an arithmetic convention every future caller has to remember, and the guest's
setter is still *allowed* to touch the host's value. As it stands the guest path never names a
device, so the overwrite cannot be reintroduced by someone forgetting to multiply.

## Why it looked fine for so long

`[app] audio volume 0%` reports the value **requested**. Nothing read back what the device actually
held, so the log was true about the intent and false about the output for the whole life of the flag.
The read-back is therefore part of the fix rather than decoration — and so is logging the failure
case, since SDL refuses device gain on a *physical* device and accepts it only on a logical one.

## Verification

45 s run of `PPSA02664` with `--volume 0`, on this change:

```
[app] audio volume 0%
prosper-audio: port 1 amplifier gain set 0.00, device reports 0.00
prosper-audio: port 2 amplifier gain set 0.00, device reports 0.00
prosper-audio: port 3 amplifier gain set 0.00, device reports 0.00
```

Three ports, requested and observed agreeing. Before the change the same run leaves the guest's own
gain in place on every port the title touches.

No test is added, and that is a gap I would rather state than paper over: the assertion worth having
is "the guest setting its volume does not change what the host set", and reaching it needs an
`Sdl3AudioSink` with a live SDL audio device, which CI has no way to open. The read-back log is the
substitute — it makes a future regression visible in any run's own output instead of silent.

## Risk

Behaviour with no `--volume` flag is unchanged: the amplifier defaults to 1.0 and
`SDL_SetAudioDeviceGain(device, 1.0f)` is a no-op. A port opening later still picks up a `--volume`
set before the guest created any port, which the previous code also did — that contract is preserved
and is why the value is still remembered on the sink.

## Impact beyond the annoyance

Every headless route run in this repository launches with `--volume 0` and assumed it worked. Those
runs were exercising a full audio mixer path they were not meant to be.

### DIFF
--- a/prosper/frontends/audio_sdl3/audio_sdl3.hpp
+++ b/prosper/frontends/audio_sdl3/audio_sdl3.hpp
@@ -35,7 +35,7 @@ bool install_sdl3_audio_sink();
 void set_sdl3_audio_paused(bool paused);
 
 // Set the HOST's playback gain -- an amplifier after the guest's own mixer, as a linear multiplier
-// (1.0 = unchanged). Applied via SDL_SetAudioStreamGain on the logical DEVICE each port feeds, so
+// (1.0 = unchanged). Applied via SDL_SetAudioDeviceGain on the logical DEVICE each port feeds, so
 // no sample data is copied or clipped by prosper and the guest's per-port channel gain (which it
 // sets through sceAudioOutSetVolume) is left entirely alone. Deliberately a different knob: the two
 // used to write the same one, so the guest silently undid `--volume` seconds into every boot
--- a/prosper/frontends/audio_sdl3/audio_sdl3.hpp
+++ b/prosper/frontends/audio_sdl3/audio_sdl3.hpp
@@ -34,14 +34,23 @@ bool install_sdl3_audio_sink();
 // No-op when the SDL3 sink is not installed.
 void set_sdl3_audio_paused(bool paused);
 
-// Set the playback gain applied to every stream, as a linear multiplier (1.0 = unchanged).
-// Applied via SDL_SetAudioStreamGain, so no sample data is copied or clipped by prosper.
-// Takes effect immediately on open streams and is remembered for streams opened later.
+// Set the HOST's playback gain -- an amplifier after the guest's own mixer, as a linear multiplier
+// (1.0 = unchanged). Applied via SDL_SetAudioStreamGain on the logical DEVICE each port feeds, so
+// no sample data is copied or clipped by prosper and the guest's per-port channel gain (which it
+// sets through sceAudioOutSetVolume) is left entirely alone. Deliberately a different knob: the two
+// used to write the same one, so the guest silently undid `--volume` seconds into every boot
+// (#3489). Takes effect immediately on open streams and is remembered for streams opened later.
 //
 // prosper-app defaults this to 1.0 -- the title's own mix, unedited, which is the only output a
 // compatibility layer can call correct. Attenuate a bring-up run with `--volume N` rather than
 // changing the default; see kDefaultVolumePercent in prosper-app/main.cpp.
 void set_sdl3_audio_gain(float gain);
+
+// Read back what SDL actually holds for a port: `channel` is the guest's own per-stream gain and
+// `amplifier` the host's device gain. False when the port has no stream. For tests -- the whole
+// point of the split is that the guest cannot move the host's setting, and an assertion has to be
+// able to observe BOTH to say so.
+bool sdl3_audio_port_gains_for_test(int port, float* channel, float* amplifier);
 
 // Uninstall the SDL3 sink (restores the default), joining diagnostics and closing streams.
 // Call after guest audio producers stop, BEFORE SDL_Quit/SDL_QuitSubSystem(AUDIO).
--- a/prosper/frontends/audio_sdl3/audio_sink_sdl3.cpp
+++ b/prosper/frontends/audio_sdl3/audio_sink_sdl3.cpp
@@ -294,7 +294,8 @@ public:
                 SDL_Log("[audio-demand-error] port=%d callback unavailable: %s", port, SDL_GetError());
         }
         s.next = {};   // (re)start the per-grain pacing clock on the first output()
-        SDL_SetAudioStreamGain(s.stream, gain_);
+        SDL_SetAudioStreamGain(s.stream, 1.0f);   // no guest volume on a freshly opened port
+        apply_amplifier_locked(port - 1);         // remembered host setting, whenever it was made
         const bool stream_ready = paused_ ? SDL_PauseAudioStreamDevice(s.stream)
                                           : SDL_ResumeAudioStreamDevice(s.stream);
         if (!stream_ready) {
@@ -651,9 +652,11 @@ public:
         if (port < 1 || port > kMaxPorts || !vols) return;
         // Approximate the PS5 per-channel volumes as a single stream gain (0..1) from the loudest set channel.
         int maxv = audio_peak_channel_volume(mask, vols);
-        float gain = maxv / 32768.0f;                     // SCE_AUDIO_VOLUME_0DB == 32768
+        const float guest = maxv / 32768.0f;              // SCE_AUDIO_VOLUME_0DB == 32768
         std::lock_guard<std::mutex> lk(mx_);
-        if (SDL_AudioStream* st = slots_[port - 1].stream) SDL_SetAudioStreamGain(st, gain);
+        // The guest's own mixer channel. The host's amplifier lives on the device this stream
+        // feeds, so setting this cannot disturb it -- which it did, until #3489.
+        if (SDL_AudioStream* st = slots_[port - 1].stream) SDL_SetAudioStreamGain(st, guest);
     }
 
     void close(int port) override {
@@ -671,10 +674,37 @@ public:
 
     // whether it is set before or after the guest creates its ports.
 
+    // The host's amplifier, on the DEVICE the port's stream feeds -- deliberately not the stream's
+    // own gain, which belongs to the guest. `sceAudioOutSetVolume` is the title setting a channel on
+    // its virtual PS5 mixer; `--volume` is what that mixer is plugged into. They are different
+    // points in the signal chain and SDL has both, so nothing here has to multiply them and no
+    // future caller can reintroduce the overwrite this replaced: the guest path never names a
+    // device. Takes an index rather than a Slot& so its signature does not mention a type declared
+    // further down the class.
+    void apply_amplifier_locked(int index) {
+        SDL_AudioStream* stream = slots_[index].stream;
+        if (!stream) return;
+        const SDL_AudioDeviceID device = SDL_GetAudioStreamDevice(stream);
+        // Reported rather than assumed. SDL refuses this on a PHYSICAL device and only accepts it
+        // on a logical one; SDL_OpenAudioDeviceStream gives us a logical device, so a false here
+        // means something about that changed. Silence about it is how `--volume 0` came to print
+        // "audio volume 0%" while the title kept talking.
+        if (!device || !SDL_SetAudioDeviceGain(device, amplifier_)) {
+            SDL_Log("prosper-audio: port %d could not apply --volume gain %.2f: %s",
+                    index + 1, amplifier_, SDL_GetError());
+            return;
+        }
+        // Read back what the device actually holds, so the log states an OBSERVED gain rather than
+        // an intended one. `[app] audio volume 0%` was true about the intent and false about the
+        // output for the whole life of the flag; a number nobody read back is how that survived.
+        SDL_Log("prosper-audio: port %d amplifier gain set %.2f, device reports %.2f",
+                index + 1, amplifier_, SDL_GetAudioDeviceGain(device));
+    }
+
     void set_gain(float g) {
         std::lock_guard<std::mutex> lk(mx_);
-        gain_ = g;
-        for (auto& s : slots_) if (s.stream) SDL_SetAudioStreamGain(s.stream, gain_);
+        amplifier_ = g;
+        for (int i = 0; i < kMaxPorts; ++i) apply_amplifier_locked(i);
     }
 
     void set_paused(bool paused) {
@@ -792,7 +822,10 @@ private:
     std::atomic<bool> timeline_running_{false};
     std::thread       timeline_thread_;
     bool paused_ = false;
-    float gain_ = 1.0f;   // linear playback gain, applied via SDL_SetAudioStreamGain
+    // The host's amplifier, applied via SDL_SetAudioDeviceGain AFTER the guest's own per-port
+    // channel gain. Remembered here so it survives ports opening and closing, and so a setting made
+    // before the guest creates any port still reaches every port it later opens.
+    float amplifier_ = 1.0f;
 };
 
 Sdl3AudioSink g_sink;
--- a/prosper/frontends/audio_sdl3/audio_sink_sdl3.cpp
+++ b/prosper/frontends/audio_sdl3/audio_sink_sdl3.cpp
@@ -707,6 +707,22 @@ public:
         for (int i = 0; i < kMaxPorts; ++i) apply_amplifier_locked(i);
     }
 
+    // Reads SDL back rather than reporting what we asked for. A gain nobody reads back is exactly
+    // how "[app] audio volume 0%" came to be printed by a process that kept making noise.
+    bool port_gains(int port, float* channel, float* amplifier) {
+        if (port < 1 || port > kMaxPorts) return false;
+        std::lock_guard<std::mutex> lk(mx_);
+        SDL_AudioStream* stream = slots_[port - 1].stream;
+        if (!stream) return false;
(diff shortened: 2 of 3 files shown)


mattias800 (commented):
**REJECTED** on head `dbfda36e` — two items, both small. The mechanism is right and I confirmed it against SDL's own source rather than its header prose; what fails is the record around it.

## What I verified holds

- **`SDL_GetAudioStreamDevice` really does return a logical device here.** `audio_sink_sdl3.cpp:278` opens each port with `SDL_OpenAudioDeviceStream`, and SDL's `SDL_audio.h:729-733` states logical devices "are created by calling SDL_OpenAudioDevice or SDL_OpenAudioDeviceStream"; `SDL_audio.h:1024-1026` says `SDL_GetAudioStreamDevice` "reports the logical audio device that an audio stream is currently bound to". One logical device per port, so a per-port device gain is legal and does not leak across ports. The `!device` guard at `:692` covers the not-bound/invalid `0` return the header documents.
- **The ordering claim in the PR body is the header's own.** `SDL_audio.h:884-886`: *"This is applied, along with any per-audiostream gain, during playback to the hardware."* Guest gain then host amplifier, as claimed.
- **`amplifier_` reaches ports opened later and ports opened earlier.** Later: `open()` calls `apply_amplifier_locked(port - 1)` at `:298`. Earlier: `set_gain()` at `:704-707` sweeps all `kMaxPorts` slots, and `apply_amplifier_locked` no-ops on empty ones at `:685-686`. Both call sites hold `mx_` (`:270`, `:705`), so the `_locked` suffix is honest.
- **Nothing else writes stream gain.** Repo-wide the only writers are `:297` (reset to 1.0 on a fresh stream) and `:659` (the guest). `SDL_SetAudioDeviceGain` has exactly one call site, `:692`. So the "no future caller can reintroduce the overwrite" argument stands — *except* for finding 1 below, which is the one thing that would tell a future caller otherwise.
- **Pause/resume preserves it.** `set_paused` (`:709-732`) touches only `SDL_Pause/ResumeAudioStreamDevice`; the logical device and its gain survive. Reopen destroys the stream and hence the logical device, and `:298` reapplies on the new one.
- **The new logging cannot spam per frame or per submit.** `apply_amplifier_locked` is reachable only from `open()` and `set_gain()`; `set_gain` has one production caller, `prosper-app/main.cpp:1146`, once at startup. Per-open cost is one line, beside the "opened port %d on %s" line `open()` already emits at `:334`.

## 1. `audio_sdl3.hpp:38` now documents the bug this PR removes — blocking

```
// Applied via SDL_SetAudioStreamGain, so no sample data is copied or clipped by prosper.
```

That is the public contract comment for `set_sdl3_audio_gain`, and after this change it names the mechanism the fix deletes. Concrete scenario: the next person adding a per-port effect reads the header, concludes the host amplifier lives on the *stream*, and writes `SDL_SetAudioStreamGain` for it — reintroducing #3489 exactly. The PR body's strongest argument is that the overwrite cannot come back because "the guest path never names a device"; that argument is only as good as the header not pointing the reader at the stream. Fix the sentence to name `SDL_SetAudioDeviceGain` and say the guest's own per-port gain is a separate, earlier stage. One line.

## 2. The stated reason for shipping no test is falsified by the repo — blocking

> reaching it needs an `Sdl3AudioSink` with a live SDL audio device, which CI has no way to open

CI opens one today, eight times. `CMakeLists.txt:3845-3851` builds `test_audio_sdl3` and registers it with `ENVIRONMENT "SDL_AUDIO_DRIVER=dummy"`; its own comment at `:3844` describes it as "install the sink, drive the sceAudioOut HLE, assert grains flow through SDL with no device." `:3852-3869` adds six more cases on the same driver. And the fixture for the exact assertion you want is *already written*: `test_audio_sdl3.cpp:41` calls `install_sdl3_audio_sink()`, `:45` opens a port through the real `sceAudioOutOpen`, `:55` drives the real `sceAudioOutSetVolume`.

The dummy driver is not an obstacle to the thing under test, either — gain is core, not backend. `_deps/sdl3-src/src/audio/SDL_audio.c:1923-1939`: `SDL_SetAudioDeviceGain` resolves the logical device and assigns `logdev->gain`, with no driver involvement.

What is genuinely missing is a **read-back accessor**, not a device: nothing public hands a test the port's `SDL_AudioStream*` or its device id. Adding one next to `sdl3_audio_demand_snapshot` (`audio_sdl3.hpp:25`) turns "the guest setting its volume does not change what the host set" into about fifteen lines in a test that already exists — set gain 0, open, guest `sceAudioOutSetVolume` full-scale, assert the device still reports 0. That test fails on `origin/main` and passes here, which is the arm this change is missing.

Either add it, or replace the justification with one that survives reading `CMakeLists.txt`. I would rather have the test: this PR's whole subject is a number that was never read back, and "CI cannot" is itself a number nobody read back.

## Non-blocking

- **The Risk section is right about audio and wrong about output.** With no `--volume`, `SDL_SetAudioDeviceGain(device, 1.0f)` is indeed a no-op — but `:699-701` still logs `port N amplifier gain set 1.00, device reports 1.00` on every port open, for every user who never passed the flag. That is new default-path noise. Consider logging the read-back only when `amplifier_ != 1.0f`, keeping the failure branch unconditional.
- **A doc comment lost its function.** The pre-existing two-line comment at `:673-675` ("Applies to open streams immediately AND is remembered for later opens…") documented `set_gain`. Inserting `apply_amplifier_locked` between them leaves it attached to the new function, where it reads as a description of the wrong thing. Move it back down to `:704`.
- **`:297`'s comment asserts a policy, not a finding.** "no guest volume on a freshly opened port" is unchanged behaviour — the old line put `gain_` there and never restored the guest's last-set volume either — so this is not a regression. But the comment reads as a decision that was weighed, and as far as I can tell the sink has never remembered per-port guest volume across a reopen. Worth phrasing as "unchanged: the sink does not remember guest volume across a reopen" so it is not later cited as settled.
- **Out of scope, but it limits the claim.** The launcher's own music opens a separate stream at `prosper-app/library_media.cpp:132` and is gained by `PROSPER_LAUNCHER_MUSIC_VOLUME` (`library_ui.cpp:189`), not by `--volume`. So `--volume 0` silences the title and not the launcher. Pre-existing and not this PR's job; flagging it because the body reads as though `--volume` is now *the* post-mix amplifier for everything the app emits.

Re-review on the corrected head is quick — I have the mechanism checked already.

mattias800 (commented):
**REJECTED** on head `81bdcf93` — two items, and I am sorry to hold it, because the test is genuinely the right test. One is a two-word fix; the other is four lines.

The accessor is well judged. `audio_sink_sdl3.cpp:712-725` reads both knobs back from SDL under `mx_` rather than reporting what was requested, and the test's shape — "the guest could not move it", plus the reverse — is the assertion that was missing. I re-derived your mutation result rather than taking it: restoring `SDL_SetAudioStreamGain` for the host makes `set_sdl3_audio_gain(0.0f)` land on the stream, so `:63` reads amplifier 1.0 and fails, `:64` reads channel 0.0 and fails, `:71` fails, and after `set_sdl3_audio_gain(0.5f)` both `:78` and `:79` fail. Five, exactly as you reported. Float equality is safe throughout — `SDL_audio.c:1923-1939` stores the value verbatim, and nothing arithmetic sits between the write and the read-back.

## 1. The header still names the wrong function — `audio_sdl3.hpp:38`

```
// (1.0 = unchanged). Applied via SDL_SetAudioStreamGain on the logical DEVICE each port feeds, so
```

The object was corrected and the verb was not: the code calls `SDL_SetAudioDeviceGain` (`audio_sink_sdl3.cpp:692`), and your own in-class comment at `audio_sink_sdl3.cpp:841` says so correctly. So the one remaining place naming the stream API for the host amplifier is the public header — the file a caller reads instead of the implementation, which is what made it worth raising the first time. The new phrasing is arguably worse than the old for the scenario I gave: "`SDL_SetAudioStreamGain` on the logical DEVICE" is not a thing that exists, so a reader who half-trusts it may go looking for a stream-gain call to add their effect to. Change the function name to `SDL_SetAudioDeviceGain` and this is done.

## 2. The test asserts the ordering production never uses — `test_audio_sdl3.cpp:47-80`

The port is opened at `:47` and the amplifier is set at `:61`, so every assertion exercises `set_gain()` sweeping an already-open slot. Production is the other way round: `prosper-app/main.cpp:1146` calls `set_sdl3_audio_gain` during startup, before the guest has opened anything, so a real `--volume` run reaches the device **only** through `apply_amplifier_locked(port - 1)` at `audio_sink_sdl3.cpp:298`.

That line is not covered. Delete it and the new test stays green: the amplifier is 0.0 by `:63` because `set_gain` swept the open port, and by the time any later port opens `:80` has returned the amplifier to 1.0, where no assertion can tell the difference. So a change that breaks `--volume` for every real user — the exact user-visible failure of #3489, arriving by a different route — passes this test.

Four lines close it, and they fit where you already are: leave the amplifier at 0.0 instead of restoring it at `:80`, open a second port through `sceAudioOutOpen`, read back with `sdl3_audio_port_gains_for_test(2, …)`, assert `amplifier == 0.0f`, then restore. That makes the test cover both directions of the contract the header claims at `:41` ("takes effect immediately on open streams **and** is remembered for streams opened later"), which currently has only its first half asserted.

I am raising this rather than filing it because of what the charter says about a control that cannot express the case it is validating. The test you added is real and it is correctly aimed at the defect that happened; it just cannot see the path the flag actually takes, and a green run of it would be read as covering `--volume` generally.

## The three you did not act on

Your judgement on all three, and I would not block on any of them:

- **The launcher music — do not extend the amplifier here.** You are right, and I would go further: folding `library_media.cpp:132` into the host amplifier would change what `--volume` *means*, from "the amplifier the guest's mixer feeds" — the clean contract this PR establishes — to "everything the process emits". Those are different features and the second one deserves its own issue. Fixing the body's wording is the correct action. It is the one a user trips over, so if you want it to be more than wording, an issue is cheap.
- **The Risk section vs the new default-path log line.** Non-blocking, and it stays non-blocking, but the imprecision is worth removing from the body since the body is the record: with no `--volume` the audio is unchanged and the *output* is not. A one-line guard on `audio_sink_sdl3.cpp:699-701` for `amplifier_ != 1.0f`, keeping the failure branch unconditional, would let the sentence be true as written.
- **The detached comment at `audio_sink_sdl3.cpp:689-691`.** Cosmetic. Fine to leave.

Nothing else changed in scope — three files, all audio; the `gpu_timeline` and `capture_tunables` content in the range diff is #3492 arriving under you in the rebase, not this PR. Re-review at the corrected head will be quick; the mechanism and the mutation are both checked already.

mattias800 (commented):
**APPROVED** on head `481d53e0`. Both findings are properly closed, and the second is closed in the way that makes it stay closed.

## The ordering gap

`test_audio_sdl3.cpp:87-93` is the right shape. I checked the two things that would have made it look right without being right:

- **The port index is real, not assumed.** `hle_audio.cpp:424-431` allocates `handle = i + 1` from the first free slot and passes that same handle to `s->open(handle, info)`, and `h` is still open at `:87` (it is not closed until `:96`), so the late open takes slot index 1, returns handle 2, and `sdl3_audio_port_gains_for_test(2, …)` at `:90` reads the port that was just opened. Not a coincidence of numbering.
- **The mutation fails where you said, and only there.** Deleting `apply_amplifier_locked(port - 1)` at `audio_sink_sdl3.cpp:298` leaves the new stream at SDL's device default of 1.0, so `:91` reads 1.0 against `0.5f` and fails; `:92` still passes because `:297` sets the channel gain independently. So the assertion that fails is the one that names the defect, rather than a downstream one that merely notices something moved.

That also picks up the half of the header's contract at `audio_sdl3.hpp:41` that had no assertion — "remembered for streams opened later" is now observed rather than claimed. Both directions of the split are covered: guest cannot move the host (`:71`), host cannot discard the guest (`:79`), and the host reaches a port that did not exist when it was set (`:91`).

One tightening, not worth a push: `:88` asserts `late >= 1` while `:90` reads port 2, so the test states a weaker fact than it relies on. It is fail-visible rather than silently void — `port_gains` returns false for an empty slot, so a changed allocation reddens `:90` loudly — which is why I am not asking for it. `late == 2` would just say what is meant.

## The header verb

`audio_sdl3.hpp:38` now reads `SDL_SetAudioDeviceGain`, matching the call at `audio_sink_sdl3.cpp:692` and the in-class comment at `:841`. Nothing in the tree points a caller at the stream API for the host knob any more; I re-grepped.

## The three notes

- **Launcher.** The body now scopes the claim to the guest's audio-out ports and names #3499 for the design question. That is the right resolution — the reader can no longer take the wider meaning from this PR, and the wider meaning gets decided on its own terms rather than by an implementation detail.
- **Risk section.** Leave it. I said non-blocking twice and I am not going to convert it into a condition at the third pass; it is a log line, the code is right, and the sentence is imprecise rather than false. If you touch the file again for another reason, guarding `audio_sink_sdl3.cpp:699-701` on `amplifier_ != 1.0f` makes it true as written.
- **Detached comment.** Agreed, not worth a push on its own.

Scope is unchanged — three files, all audio. Merge when CI is green at this head.

PR #3492, opened 2026-09-09
fix(capture): break the exit deadlock between the owned writer and the CRT onexit table

### LINKED ISSUE #3470: capture_tunables times out on Windows MinGW since #3460 — main red on every run, blocking the merge gate

mattias800:
Fixes #3470. `main`'s **Windows MinGW** job has failed on every run since #3460 â€” always the same
single test, always a 600 s ctest timeout rather than an assertion.

## The cause, from a stack rather than a hypothesis

`gdb` attached to the hung `capture_tunables` process:

```
Thread 1 (main)    _execute_onexit_table -> ~InteractiveFrameBundle -> shutdown
                   -> std::thread::join -> WaitForSingleObject
Thread 2 (writer)  __pformat_float -> __gdtoa -> __Balloc_D2A -> dtoa_lock -> atexit
                   -> _register_onexit_function -> RtlEnterCriticalSection   [blocked]
```

The writer is inside `fprintf`, formatting the `%.3f` of its own completion log. MinGW's float
conversion **lazily registers an atexit handler** the first time `__gdtoa` allocates, and that takes
the CRT's onexit-table critical section â€” the lock the exiting main thread already holds while
walking that same table from a static destructor. Main waits for the writer; the writer waits for
main's lock.

Linux never sees it: the mechanism is MinGW's `gdtoa`. It needs the writer thread #3460 introduced,
which is why the bisect landed there.

The issue recorded a plausible and different hypothesis â€” a thread not joined, or waiting on
something that never arrives. It was wrong. The writer is neither unjoined nor unwoken; it is
blocked inside `printf`.

## The fix

Two changes, each removing one half of the meeting:

1. **`shutdown()` no longer hands the writer a job.** A cancelled window writes nothing â€” the error
   is preset, so `write_gpu_capture_bundle` was never reached â€” so the job's only effect was to log
   and publish, and the caller can do both. The writer's remaining act is to see `stopping` and
   return.
2. **The writer's duration log formats an integer of microseconds.** Nothing about the message needs
   a float, and `%.3f` was the only call on that thread that could reach the lazy-atexit path.

**Measured separately, either change alone is sufficient** â€” each arm passes with the other reverted.
That is what the stack predicts: the deadlock needs the writer to be doing work during exit *and*
that work to take the CRT lock. Keeping both is defence in depth, not two necessary fixes, and I would
rather say so than imply a second cause I did not observe.

## The regression is bounded and names itself

A 600 s ctest timeout is a poor signal: no assertion output, no indication of which case hung, ten
minutes of every CI run to observe. The new arm makes the child police its own exit â€” it arms a real
capture, returns from `main`, and a watchdog thread hard-exits with a distinctive status if static
destruction has not completed in 20 s. The parent asserts a status.

Verified in both directions. With the defect restored:

```
  [note] child rc=143, stderr was: [child] armed=1
[grab] frame-bundle owned writer finished in [child] STILL RUNNING 20s after main returned: ...
  [FAIL] a process that arms a capture and merely returns from main still EXITS
  [FAIL] ...and exits without its watchdog having to kill it
```

Note the truncated `owned writer finished in ` â€” the writer caught mid-`fprintf`, exactly where gdb
found it.

The watchdog writes with `write(2, ...)` rather than `fprintf`, and that is not incidental: with
stdio it **never fired at all**, because it queued behind the very deadlock it exists to report and
the child had to be killed by the harness. A watchdog that shares a lock with its subject is not a
watchdog.

**One claim withdrawn, on review.** What the negative-control transcript demonstrates is that the arm
*detects* the regression — two failures, and the watchdog's `STILL RUNNING` line reaches the child's
stderr, so the thread ran. It does **not** demonstrate that `_Exit(3)` completed: the parent hangs at
its own exit in that build, so the harness timeout killed the whole process group and `system()`
returned 143 rather than 3. "A regression is detected and names itself" is proven; "a regression
costs 20 s" is not.

## Verification

`test_capture_tunables` on Windows/MinGW, this branch: **3 of 3 runs exit 0 and PASS**. Before:
hangs every time, killed at the harness timeout.

Every assertion in that test passed even before this change â€” the test's logic was never at fault,
only its ability to exit. (An earlier reading of mine reported five assertion failures; that was my
own error, running the binary from the wrong working directory so its child spawns could not resolve
the executable. Withdrawn.)

## Risk

The cancelled-window outcome is now published by the thread calling `shutdown()` rather than by the
writer. Both did so under `mx`, and the observable result â€” `outcome`, `outcome_pending`, and the
`[grab] frame-bundle write failed: capture window cancelled at shutdown` line â€” is unchanged;
`test_interactive_capture_writer` asserts exactly that sequence and still passes. A window genuinely
mid-*write* when shutdown is called is unaffected: that job is already in flight and is joined as
before.

### DIFF
--- a/prosper/docs/WINDOWS_PORT_HANDOFF.md
+++ b/prosper/docs/WINDOWS_PORT_HANDOFF.md
@@ -260,6 +260,28 @@ that thread waits on becomes the focus, so matching signals from other threads a
 
 ## Gotchas learned (so the next agent doesn't relearn them)
 
+- **A thread joined from a static destructor can deadlock against the CRT's own exit table, and the
+  usual suspects are the wrong ones.** `main` was red on Windows MinGW for a day with
+  `capture_tunables (Timeout)` — a hang with no assertion output. The recorded hypothesis was the
+  natural one, and it was **falsified**: the writer thread was neither unjoined nor waiting on a
+  signal that never arrived. `gdb` on the hung process put it inside `fprintf`:
+
+  ```
+  Thread 1 (main)    _execute_onexit_table -> ~InteractiveFrameBundle -> std::thread::join
+  Thread 2 (writer)  __pformat_float -> __gdtoa -> __Balloc_D2A -> dtoa_lock -> atexit
+                     -> _register_onexit_function -> RtlEnterCriticalSection   [blocked]
+  ```
+
+  MinGW's float formatting lazily registers an `atexit` handler the first time `__gdtoa` allocates,
+  taking the CRT onexit-table critical section — the lock the exiting main thread already holds
+  while walking that table. Linux never sees it: the mechanism is MinGW's `gdtoa`.
+
+  **The general rule: a thread still doing work while another thread runs static destructors must
+  not touch anything that can lazily initialise CRT state.** `%f` is the one that bit us; any first
+  use of a facility that registers cleanup would do. Prefer integer formatting on such a thread, and
+  prefer draining worker threads explicitly before `main` returns over relying on a destructor to
+  join them. Fixed in #3470 / PR #3492.
+
 - Windows resets the user `%fs` base to 0 on every kernel transition; guest TLS survives only via the
   VEH `wrfsbase` re-apply (`hle_kernel_mem.cpp`/`guest_tls.cpp`). Native FSGSBASE (`rd/wrfsbase`,
   `WaitOnAddress`, `WakeByAddress*`) works and needs `-lsynchronization` (linked into `prosper_core`).
--- a/prosper/src/gpu/timeline/gpu_timeline.cpp
+++ b/prosper/src/gpu/timeline/gpu_timeline.cpp
@@ -2260,6 +2260,7 @@ void InteractiveFrameBundle::run_writer() {
 void InteractiveFrameBundle::shutdown() {
     // Serializes joiners without holding the state mutex while the writer publishes its result.
     std::lock_guard shutdown_lock(shutdown_mx);
+    GpuCaptureBundle abandoned;                // released after `mx` is dropped; see below
     {
         std::lock_guard lock(mx);
         stopping = true;
@@ -2269,7 +2270,11 @@ void InteractiveFrameBundle::shutdown() {
         // writer while this thread waits inside a static destructor is half of the exit deadlock
         // in #3470. Reporting it inline leaves the writer with nothing to do but return.
         if (capturing || !armed_path.empty()) {
-            bundle = {};                       // free the payload; nobody will write it
+            // Detached under the lock, released outside it. This can be gigabytes, and on the
+            // destructor path the thread running it is inside the CRT's exit lock -- the old code
+            // freed the payload on the WRITER, outside `mx`, and that property is worth keeping.
+            abandoned = std::move(bundle);
+            bundle = {};
             InteractiveGrabOutcome cancelled;
             cancelled.bundle_path = capturing ? current_path : armed_path;
             cancelled.max_unique_bytes = max_unique_bytes;
@@ -2277,13 +2282,20 @@ void InteractiveFrameBundle::shutdown() {
             std::fprintf(stderr, "[grab] frame-bundle write failed: %s\n", cancelled.error.c_str());
             outcome = std::move(cancelled);
             outcome_pending = true;
-            writing = false;
+            // `writing` is deliberately NOT touched. This block is reachable only with it already
+            // false -- a request is refused while writing (see request_interactive_capture_bundle),
+            // and the present handler sets it only while clearing `capturing` and `armed_path`, so
+            // a genuinely mid-write window takes neither branch and is joined below exactly as
+            // before. Clearing it here would be dead on every real path and would release the busy
+            // flag mid-write on any path that ever reached it, letting a new capture race the
+            // outcome slot.
             capturing = false;
             armed_path.clear(); current_path.clear();
         }
         g_interactive_frame_active.store(false, std::memory_order_release);
         wake.notify_one();
     }
+    abandoned = {};                            // outside `mx`, before the join
     if (writer.joinable()) writer.join();
 }
 
--- a/prosper/src/gpu/timeline/gpu_timeline.cpp
+++ b/prosper/src/gpu/timeline/gpu_timeline.cpp
@@ -2240,8 +2240,13 @@ void InteractiveFrameBundle::run_writer() {
         else
             std::fprintf(stderr, "[grab] frame-bundle write failed: %s\n",
                          work.outcome.error.c_str());
-        std::fprintf(stderr, "[grab] frame-bundle owned writer finished in %.3f ms\n",
-                     std::chrono::duration<double, std::milli>(
+        // Microseconds as an INTEGER, deliberately. A "%.3f" here deadlocked the process at
+        // exit on Windows: MinGW's float formatting lazily registers an atexit handler the first
+        // time __gdtoa allocates, which takes the CRT's onexit-table critical section -- and the
+        // thread joining this one holds exactly that lock while walking the same table from a
+        // static destructor. Integer conversion never reaches __gdtoa (#3470).
+        std::fprintf(stderr, "[grab] frame-bundle owned writer finished in %lld us\n",
+                     (long long)std::chrono::duration_cast<std::chrono::microseconds>(
                          std::chrono::steady_clock::now() - start).count());
         {
             std::lock_guard lock(mx);
@@ -2258,14 +2263,21 @@ void InteractiveFrameBundle::shutdown() {
     {
         std::lock_guard lock(mx);
         stopping = true;
+        // A cancelled window is reported HERE rather than handed to the writer as a job. Nothing
+        // is written for it -- the error is preset, so write_gpu_capture_bundle was never reached
+        // -- so the only thing the writer did with it was log and publish, and doing that on the
+        // writer while this thread waits inside a static destructor is half of the exit deadlock
+        // in #3470. Reporting it inline leaves the writer with nothing to do but return.
         if (capturing || !armed_path.empty()) {
-            WriteJob cancelled;
-            cancelled.bundle = std::move(bundle);
-            cancelled.outcome.bundle_path = capturing ? current_path : armed_path;
-            cancelled.outcome.max_unique_bytes = max_unique_bytes;
-            cancelled.outcome.error = "capture window cancelled at shutdown";
-            job = std::move(cancelled);
-            writing = true;
+            bundle = {};                       // free the payload; nobody will write it
+            InteractiveGrabOutcome cancelled;
+            cancelled.bundle_path = capturing ? current_path : armed_path;
+            cancelled.max_unique_bytes = max_unique_bytes;
+            cancelled.error = "capture window cancelled at shutdown";
+            std::fprintf(stderr, "[grab] frame-bundle write failed: %s\n", cancelled.error.c_str());
+            outcome = std::move(cancelled);
+            outcome_pending = true;
+            writing = false;
             capturing = false;
             armed_path.clear(); current_path.clear();
         }
--- a/prosper/tests/gpu/timeline/test_capture_tunables.cpp
+++ b/prosper/tests/gpu/timeline/test_capture_tunables.cpp
@@ -24,6 +24,7 @@
 #include <filesystem>
 #include <string>
 #include <system_error>
+#include <thread>
 #include "fixtures/test_scratch.h"
 #ifdef _WIN32
 #include <io.h>
@@ -171,6 +172,40 @@ int main(int argc, char** argv) {
             begin_gpu_timeline_submit(n);
             record_gpu_timeline_submit(GpuState{}, n);
         }
+        return 0;
+    }
+    // Arms a real capture -- which starts the owned writer thread -- and then merely returns from
+    // main, leaving the whole teardown to static destruction. That is the shape #3470 deadlocked on:
+    // the destructor joins the writer while holding the CRT's onexit-table lock, and the writer
+    // needed that same lock to finish formatting its own completion log.
+    //
+    // The child bounds ITSELF rather than letting the parent wait. A hung exit would otherwise
+    // surface only as a ctest timeout, which produces no assertion output, says nothing about which
+    // case hung, and leaves a stuck process behind. _Exit is deliberate: it must not run the atexit
+    // table it is reporting on.
+    if (argc > 2 && std::strcmp(argv[1], "--child-arm-and-return") == 0) {
+        std::printf("  [child] started (arm and return)\n");
+        std::fflush(stdout);
+        const auto accepted = request_interactive_capture_bundle(argv[2], 64).accepted;
+        std::fprintf(stderr, "[child] armed=%d\n", accepted ? 1 : 0);
+        std::fflush(stderr);
+        std::thread([] {
+            std::this_thread::sleep_for(std::chrono::seconds(20));
+            // A RAW write, not stdio. The thread this is watching is stuck holding CRT locks, and a
+            // std::fprintf here queues behind the very deadlock the watchdog exists to report --
+            // measured: with the defect restored the watchdog never fired at all and the child had
+            // to be killed by the harness. write() takes no CRT lock and allocates nothing, and
+            // _Exit skips the atexit table under discussion.
+            static const char msg[] =
+                "[child] STILL RUNNING 20s after main returned: exit is stuck, which is #3470\n";
+#ifdef _WIN32
+            _write(2, msg, (unsigned)(sizeof msg - 1));
+#else
+            ssize_t ignored = write(2, msg, sizeof msg - 1);
+            (void)ignored;
+#endif
+            std::_Exit(3);
+        }).detach();
         return 0;
     }
     if (argc > 2 && std::strcmp(argv[1], "--child-bundle") == 0) {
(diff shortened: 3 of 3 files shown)


mattias800 (commented):
**APPROVED** on content, with one merge condition (a red required check that I could not attribute to this
change) and five findings, none of which I consider blocking. Reviewed at head `485e92f0`. I read every line
I cite; I did not rebuild or re-run anything the author ran.

## The strongest evidence is not in the PR body, and it should be

CI run 34327560450 on this head, `Windows MinGW`:

```
266/294 Test #266: capture_tunables ...........................   Passed    0.09 sec
```

against the parent commit `64db58bc` (run 34312587269, same job):

```
266/294 Test #266: capture_tunables ...........................***Timeout 600.01 sec
```

That is an independent, machine-produced A/B of the exact claim, and it also proves the new arm ran and
passed there: `run_child` returns `std::system`'s status and the arm asserts `child.rc == 0`
(`test_capture_tunables.cpp:594-597`), so a `system()` that never spawned would have failed the test. Worth
quoting in the PR body — it is better evidence than the 3-of-3 local runs.

I also confirmed the narrative's premise from the code rather than from the stack trace: the **parent**
process arms an interactive capture at `prosper/tests/gpu/timeline/test_capture_tunables.cpp:370` and never
presents or shuts down, so its own static destruction takes exactly the cancel-at-shutdown path. That is why
the whole binary, not one arm, hung.

## 1. The moved cancellation publish — state equivalence holds

I traced this rather than taking it. The block at `prosper/src/gpu/timeline/gpu_timeline.cpp:2271-2283` can
only be entered with `writing == false`:

- `request_interactive_capture_bundle` refuses while `b.writing` (`gpu_timeline.cpp:2191`), so nothing can be
  armed during a write;
- `interactive_frame_bundle_on_present` sets `writing = true` only in the same statement group that clears
  `capturing` and `current_path` (`gpu_timeline.cpp:3136-3140`), and `armed_path` is necessarily already empty
  there because the arm was consumed at `:3148`.

So the genuinely-mid-write case falls through the block untouched and is joined exactly as before — the PR
body's risk paragraph is correct. `outcome` / `outcome_pending` / `capturing` / `armed_path` end in the same
values the old cancelled `WriteJob` would have produced, and `InteractiveGrabOutcome::ok` defaults to `false`
(`gpu_timeline.hpp:313`) just as the old default-constructed job's did.

Published exactly once: `outcome_pending` cannot already be true on entry, because `:2191` also refuses an arm
while an outcome is uncollected, and only `run_writer` sets it otherwise.

Observability: both forms publish before `shutdown()` returns — the old one because `shutdown()` joins, the
new one because it publishes under `mx` before the join. That is all the two consumers depend on:
`prosper/frontends/prosper-app/main.cpp:3159-3160` calls `shutdown` then `reportGrabOutcome`, and
`prosper/tests/gpu/capture/test_interactive_capture_writer.cpp:36-40` calls `take_interactive_grab_outcome`
synchronously right after `shutdown` with no wait.

One corner the new form strictly improves: under the old code, if the writer thread had somehow not been
started, the cancelled job was enqueued and then `if (writer.joinable())` skipped the join, so nobody ever
drained it and the outcome was never published. Unreachable today (the thread is started before the arm at
`:2195`), but it is gone rather than merely unreached.

### Finding 1 (minor) — `writing = false;` at `gpu_timeline.cpp:2280` is dead, and dangerous if it ever is not

By the argument above `writing` is already `false` on every path into that block. If a future edit ever made
it reachable with `writing == true`, this line would clear the busy flag *while the writer is inside
`write_gpu_capture_bundle`*, and `request_interactive_capture_bundle` would then admit a new capture whose
result the in-flight writer overwrites at `:2252-2255`. Concrete: a window closes (`writing = true`, job
queued), some future path re-arms, shutdown runs, `writing` is cleared, a new arm is admitted, and two
outcomes race one `outcome` slot. Prefer deleting the line, or replacing it with an assertion that states the
invariant you are relying on.

### Finding 2 (minor) — the payload is now freed inside `mx`, on the exiting thread, inside the CRT lock

`bundle = {};` at `gpu_timeline.cpp:2272` destroys what can be a multi-gigabyte `GpuCaptureBundle` while
holding `mx`. The old code moved it into the job (O(1)) and the writer freed it outside the lock at
`:2237`. Given this PR is specifically about what the exiting thread does while the onexit-table critical
section is held, adding a large `free()` there is the wrong direction, even if the heap lock is not the lock
in question. One line fixes it:

```cpp
GpuCaptureBundle dead = std::move(bundle);   // inside the lock
...                                          // publish
```

and let `dead` destruct after the `mx` scope closes.

## 2. Is the deadlock closed, or narrowed?

Narrowed to empty in practice, not closed structurally — and I think the PR should say that in one sentence
rather than only implying it.

With change 1 the writer has *no work at all* during static destruction unless a real window was mid-write.
For that remaining case I audited what the writer touches, since that is where the PR's guarantee actually
rests:

- `prosper/src/gpu/capture/gpu_capture_bundle.cpp` contains **no `printf`/`cerr` at all** and **no `%f/%g/%e`
  conversion anywhere**, so `write_gpu_capture_bundle` cannot reach `__gdtoa`.
- Its only function-local static is `static std::atomic<uint64_t> serial{0}` at
  `gpu_capture_bundle.cpp:118` — constant-initialised and trivially destructible, so it emits no guard
  variable and registers nothing with `__cxa_atexit`. That is the other way onto the same lock, and it is
  clean.
- `work.hook` is only ever the test latch installed by `set_interactive_bundle_writer_hook_for_test`
  (`gpu_timeline.cpp:2291`); null in production.
- The two remaining `fprintf`s at `:2231` and `:2234` are `%zu`/`%s` only.

So today the residual is empty. But nothing *enforces* it: a later `%.2f` in a bundle-write log, or a
function-local `static std::string`/`std::map` first constructed on that thread, re-opens the same cycle
silently. Note there is already a live second instance of the pattern at `gpu_timeline.cpp:3734`
(`"ratio=%.3f"`), safe only because it runs on the submit thread, which nobody joins from a destructor. A
one-line comment at the top of `run_writer` naming the constraint ("this thread may be running while a joiner
holds the CRT onexit lock: no float formatting, no first-touch function-local statics") would cost nothing and
would survive the next refactor better than the comment currently sited at the `%lld`.

## 3. The regression arm

Mechanics check out. `argc > 2` holds for `--child-arm-and-return <path>`; `<thread>` is added and
`<cstdlib>`/`<unistd.h>`/`<io.h>` were already present (`test_capture_tunables.cpp:22,28-32`); the new
`std::thread` in the test TU links because `prosper_core` carries `Threads::Threads` PUBLIC
(`prosper/CMakeLists.txt:116`) and the target links `prosper_core` (`:2279`). The parent clears every
`PROSPER_CAPTURE_*` variable before the arm — `PROSPER_CAPTURE_BUNDLE` at `test_capture_tunables.cpp:396`
matters most, since a set value would make the child arm an automatic capture too. The watchdog's `msg` is a
constant-initialised `const char[]`, so the detached thread touches nothing under destruction, which is the
part that makes a detached thread outliving `main` defensible here. Spurious firing is not a concern: the
child captures nothing (no present, no submit), so its teardown frees an empty bundle; 20 s is orders of
magnitude of headroom.

### Finding 3 (worth stating in the PR, not blocking) — the arm pins the pair, never either half

The author states, and I believe, that each change alone is sufficient. That has a direct consequence for the
guard: **the new arm fails only against the full base.** Revert change 1 alone and the writer formats an
integer — green. Revert change 2 alone and the writer is never handed a job — green. So if a future edit
re-introduces the job hand-off at shutdown, or re-introduces a float on the writer thread, this test says
nothing; the defence in depth is real but untested in each half. That is an acceptable trade for a
CI-unblocking fix, but the PR body currently reads as though the arm guards the fix, and it guards the
conjunction.

### Finding 4 (claim check) — the negative-control transcript does not match the code

The PR body's "defect restored" output reads:

```
  [note] child rc=143, stderr was: [child] armed=1
[grab] frame-bundle owned writer finished in [child] STILL RUNNING 20s after main returned: ...
```

Two things I cannot reconcile against the source:

- **`rc=143`, not 3.** The watchdog exits `std::_Exit(3)` (`test_capture_tunables.cpp:207`). 143 is the shape
  of an external kill, which is what the comment at `:196-197` describes for the *earlier, stdio* watchdog.
  If `_Exit(3)` did not in fact terminate that process — plausible: `ExitProcess` terminates the main thread
  mid-`_execute_onexit_table` and then runs `DLL_PROCESS_DETACH` — then the message is delivered but the
  20 s bound is not, and "a regression costs 20 s ... instead of costing the job's whole timeout" is only
  half demonstrated. The arm still detects the regression either way (any non-zero `rc` fails `:596`), and
  ctest's `--output-on-failure` still surfaces the `STILL RUNNING` line, so this changes the advertised
  benefit rather than the guard.
- **The `[grab] frame-bundle write failed: capture window cancelled at shutdown` line is missing.** With both
  changes reverted, `run_writer` emits that line at `:2234` immediately *before* the truncated
  `owned writer finished in `, and stderr is redirected to a file the parent slurps whole. Its absence next to
  a present, truncated successor is not something buffering explains in the obvious direction.

Please post the raw child stderr and the raw `rc` for that arm. I am not asserting the fix is wrong — the CI
A/B above settles the fix independently — only that this particular quoted evidence does not read as produced
by this code, and per the charter a cited artefact carries more authority than it has earned if nobody opens
it.

### Finding 5 (charter) — the falsification is recorded only in the PR body

"#3470's recorded hypothesis — a thread not joined, or waiting on something that never arrives — was wrong;
the writer is blocked inside `printf`" is exactly the kind of expensive negative result the charter asks to be
landed in a `## Ruled out` section rather than left in a PR body. The mechanism note at
`gpu_timeline.cpp:2243-2247` is good and durable, but it is filed under the symptom.
`docs/WINDOWS_PORT_HANDOFF.md` is the right home for the general form, because it will recur outside this
file: *on MinGW, a thread joined from a static destructor must not format a float or first-touch a
non-trivially-destructible function-local static.* Please add the line (and a comment on #3470 correcting its
hypothesis) before merging — it is a docs-only edit and merges without waiting for CI.

## Merge condition

`Windows MinGW` is **red on this head**: `258 - gpu_capture_bundle_streaming (Failed)`,
`FAIL: both concurrent independent writes complete (line 264)`. On the parent commit it passed in 0.37 s, so
this is new *on this run*, and I could not attribute it to this diff: the failing assertion is the two-thread
`write_gpu_capture_bundle` race at `test_gpu_capture_bundle_streaming.cpp:248-264`, in a separate process,
over code this PR does not touch, and the pattern (two concurrent `std::filesystem::rename`s onto one Windows
target) is inherently sharing-violation-prone. My reading is a flake, but one green run on `main` is not
evidence of stability either.

So: re-run the job. If `capture_tunables` passes and `gpu_capture_bundle_streaming` is green, merge. If the
streaming failure recurs, it needs its own issue before this merges — do not merge through a red required
check on the strength of "it looks unrelated", and do not fold a second fix into this PR.

Gates I ran on the branch: `git log origin/main..HEAD --format=%B | grep -c '^Claude-Session:'` gives 0;
`git diff origin/main...HEAD --check` is clean. Diff is two files, +77/-9, and matches the described change
with nothing extra.

mattias800:
Findings addressed at `2356a9b4`. Taking them in order, including the two I am not fixing.

**1. `writing = false` — removed.** You are right that it is dead and would be harmful if it were not, and right about why: a request is refused while writing, and the present handler sets the flag only while clearing `capturing` and `armed_path`, so a genuinely mid-write window takes neither branch. That reasoning is now a comment beside the code rather than an assumption behind it.

**2. The multi-GB free — moved out of the lock.** The payload is detached under `mx` and released after it is dropped. You identified the property I had lost without noticing: the code this replaced freed it on the writer, outside the lock, and on the destructor path the thread now doing it is inside the CRT exit lock — the very lock this PR is about.

**6. Falsified hypothesis recorded** in `WINDOWS_PORT_HANDOFF.md`, with the general rule rather than the instance: a thread still working while another runs static destructors must not touch anything that lazily initialises CRT state. `%f` is what bit us; any first use of a facility that registers cleanup would do.

**3. The deadlock is narrowed, not closed structurally — agreed, and thank you for auditing the remaining path rather than taking my word.** Your `gpu_capture_bundle.cpp` audit (no `printf`, no float conversions, one constant-initialised trivially-destructible static) is a stronger statement than anything in my PR body. I am not adding an enforcement mechanism, and I would rather say why than pretend the gap is closed: I could not think of one that is not either a lint nobody will maintain or a restriction on what worker threads may log. The `%.3f` you found at `gpu_timeline.cpp:3734` is real and safe today for the reason you give.

**4. The arm pins the pair, never either half — correct, and I am leaving it.** Since either change alone suffices, a future edit reverting one of them keeps the test green. I could not construct an arm that isolates one half without asserting on log formatting, which would be a worse test. Recorded here so it is a known limit rather than a surprise.

**5. The `rc=143` discrepancy — you are right to press, and I am withdrawing half the claim.** What is demonstrated is that the arm *detects* the regression: with the defect restored it reports two failures and the watchdog's `STILL RUNNING` line appears in the child's stderr, so the watchdog thread ran. What is **not** demonstrated is that `_Exit(3)` completed — the parent hangs at its own exit in that build, so my harness timeout killed the whole process group at 200 s and `system()` returned 143. So "a regression costs 20 s" is unproven; "a regression is detected and named" is proven. I had noticed the 143 and dismissed it as harness noise, which was the wrong call: the PR body should not have carried a timing claim the transcript contradicts. I will correct the body rather than leave it.

**Merge condition.** Your read is right and it now has a cause: `gpu_capture_bundle_streaming` is a pre-existing race in `write_gpu_capture_bundle`'s install, not this diff. Two concurrent `MoveFileExW` replaces onto one destination, one gets `ERROR_ACCESS_DENIED` from the delete-pending window. Confirmed with one variable on the same build tree — 10 pass / 2 fail with this change, 10 pass / 2 fail with it reverted — and filed as #3494, fixed in #3495 (40 runs, 40 passes). I will merge that first and re-run this job rather than merging through a red check.

mattias800 (commented):
**APPROVED** at `2356a9b4`. Re-reviewed the delta from `485e92f0` rather than re-approving by assertion; all
three changes are what I asked for, and one of them is better than what I asked for.

## The delta, checked

**Finding 1 — `writing` left alone.** `gpu_timeline.cpp:2285-2291`. The line is gone and the reachability
argument is now written down beside the code, which is the half that matters: the next reader sees *why* the
block may assume `writing == false` instead of having to re-derive it from two call sites. Correct as stated —
`request_interactive_capture_bundle` refuses while writing (`:2196`) and the present handler sets it only
while clearing `capturing`/`current_path` (`:3141-3145`).

**Finding 2 — the abandoned payload.** `gpu_timeline.cpp:2263`, `:2273-2277`, `:2297`. Detached under `mx`,
released after the scope closes and before `writer.join()`. The explicit `bundle = {}` after the move is
right and not redundant: a moved-from `unordered_map` is valid but unspecified, so `chunk_indices_by_hash`
and `resource_indices_by_hash` needed the reset to be deterministic rather than merely empty-in-practice.

One correction to the framing, not to the code. This restores the *outside-`mx`* half of the old property,
not the *not-on-the-exiting-thread* half — on the destructor path the `free()` still runs on a thread holding
the CRT exit lock. That is the right trade and I would not change it: handing the payload back to the writer
is precisely the thing this PR exists to stop doing, so somebody on this side has to free it, and doing it
outside `mx` and before the join is as far as the constraint allows. Worth knowing it is a trade rather than
a restoration if this code is ever read again under time pressure.

**Finding 6 — the falsification.** `WINDOWS_PORT_HANDOFF.md:263-283`. It records the dead hypothesis, the
stack that killed it, the mechanism, and — the part I actually wanted — the general rule rather than the
instance: *a thread still working while another runs static destructors must not touch anything that lazily
initialises CRT state.* The "prefer draining worker threads explicitly before `main` returns" sentence is a
useful addition that was not in my finding. This is the form that will still be worth reading when the next
occurrence is not a `%f`.

Findings 3 and 4 recorded as known limits is the right disposition; neither was a defect, and stating them is
what stops the next reader over-trusting the guard.

## Finding 5

Your account resolves it and the correction is the right one: `system()` returning 143 because the harness
killed the process group means the arm's *detection* is demonstrated and its *20 s bound* is not. Withdrawing
the bound rather than explaining the 143 away is the correct call, and the missing
`[grab] frame-bundle write failed` line follows from the same fact — the parent's own stderr never got to
where I was reading it from.

## Merge condition — satisfied, and I verified it independently rather than reading your summary

- `gh pr view 3492 --json headRefOid` → `2356a9b4…`, equal to `git ls-remote origin
  refs/heads/fix/issue-3470-capture-exit-deadlock`. This review is attached to the tip, not to a detached
  head.
- `gh pr checks 3492 --json name,state,bucket`: 11 pass, 2 skipped, **0 fail, 0 pending**. `Windows MinGW`
  SUCCESS.
- Run 34335680786, `Windows MinGW`: `100% tests passed out of 294`, with
  `Test #266: capture_tunables ... Passed 0.10 sec` and
  `Test #258: gpu_capture_bundle_streaming ... Passed 0.38 sec`.
- Branch gates on the new head: `Claude-Session:` count 0; `git diff origin/main...HEAD --check` clean; diff
  is three files, +111/-9, nothing beyond the change described.

On #3494: the 10/2 arm on the same build tree with one variable is the right control for "not this diff", and
it is stronger than my "looks unrelated". Landing this one first is correct — a fix that can only be tested
by a suite that hangs is not testable at all.

**Approval stands at `2356a9b4`. Merge it.**

PR #3495, opened 2026-09-09
fix(capture): retry a bundle install past Windows' transient replace window

### LINKED ISSUE #3494: gpu_capture_bundle_streaming: concurrent installs race on Windows (MoveFileEx error 5), reddening CI ~1 run in 5
## The symptom

`gpu_capture_bundle_streaming` fails intermittently on Windows/MinGW:

```
258/294 Test #258: gpu_capture_bundle_streaming .....................***Failed    0.32 sec
FAIL: both concurrent independent writes complete (line 264)
```

Measured locally on Windows 11 / MinGW: **5 failures in 20 runs**, then **1 in 20** — order 5–25%,
so it will redden roughly one CI run in five for any PR that gets far enough to run it. Seen on the
CI run for #3492, whose diff cannot reach this code path.

## The cause

`test_gpu_capture_bundle_streaming.cpp:250-264` starts two threads that call
`write_gpu_capture_bundle` on the **same destination** simultaneously, and asserts both succeed —
each reserves its own temporary and the last one to install wins.

Installation is a single unretried `MoveFileExW`
(`prosper/src/gpu/capture/gpu_capture_bundle.cpp:142`):

```cpp
if (!MoveFileExW(payload.c_str(), target.c_str(), MOVEFILE_REPLACE_EXISTING)) {
    error = "cannot install bundle: Windows error " + std::to_string(GetLastError());
    return false;
}
```

With the test patched to print the failure, the answer is always the same:

```
[why] writer 0 failed: cannot install bundle: Windows error 5
```

`5` is `ERROR_ACCESS_DENIED`. On Windows a replace leaves the destination briefly in a delete-pending
state, and a second `MoveFileEx` onto it during that window is refused. It is transient by
construction — nothing is wrong with either writer, they simply collided.

## Not caused by #3492

Checked with one variable, on the same build tree:

| build | runs | pass | fail |
| --- | --- | --- | --- |
| #3492's branch | 12 | 10 | 2 |
| the same branch with its change reverted | 12 | 10 | 2 |

Identical, and #3492 touches `gpu_timeline.cpp`'s shutdown path, which this test never calls. It is
pre-existing; it was simply invisible while `capture_tunables` failed first and everyone read the job
as "the known one".

## Suggested fix

A bounded retry with a short backoff on the transient codes — `ERROR_ACCESS_DENIED` (5) and
`ERROR_SHARING_VIOLATION` (32) — and a genuine failure only after they persist. The existing
"OS sharing restriction refuses replacement" arm earlier in the same test holds the destination open
for the whole call, so it stays a refusal and would catch a retry loop that never gives up.

POSIX `rename` is atomic and does not have this window, so the retry belongs in the `_WIN32` branch.

mattias800:
Fixes #3494.

## The failure

`gpu_capture_bundle_streaming` fails intermittently on Windows/MinGW:

```
FAIL: both concurrent independent writes complete (line 264)
```

Measured on Windows 11 / MinGW: **5 failures in 20 runs**, and separately 2 in 12 — order 5–25%, so
roughly one CI run in five for any PR that gets far enough to run it.

## The cause

The test starts two threads calling `write_gpu_capture_bundle` on the **same destination** and
asserts both succeed — each reserves its own temporary and the last to install wins. Installation is
a single unretried `MoveFileExW(..., MOVEFILE_REPLACE_EXISTING)`.

Windows leaves a replaced file briefly in a delete-pending state, and a second replace onto it during
that window is refused. Patching the test to print the failure gives the same answer every time:

```
[why] writer 0 failed: cannot install bundle: Windows error 5
```

`5` is `ERROR_ACCESS_DENIED`. Nothing is wrong with either writer; they collided.

## Not caused by #3492

That PR's CI run is where this surfaced, so it was worth ruling out with one variable on the same
build tree:

| build | runs | pass | fail |
| --- | --- | --- | --- |
| #3492's branch | 12 | 10 | 2 |
| the same branch, its change reverted | 12 | 10 | 2 |

Identical. #3492 touches `gpu_timeline.cpp`'s shutdown path, which this test never calls. The race is
pre-existing and was invisible while `capture_tunables` failed first and the job read as "the known
one".

## The fix

A bounded retry — 50 attempts, 2 ms apart — covering only `ERROR_ACCESS_DENIED` and
`ERROR_SHARING_VIOLATION`, the two codes that are transient by construction. Any other code fails
immediately, as before.

The bound is not arbitrary caution. The same test has an arm that holds the destination open for the
whole call and requires the install to be **refused**; a retry loop that never gave up would turn
that refusal into a hang. That arm passes in all runs below, so the budget is short enough to still
give up.

POSIX `rename` is atomic and has no such window, so this stays inside the `_WIN32` branch.

## Verification

`test_gpu_capture_bundle_streaming` on Windows/MinGW, this branch: **40 runs, 40 passes.** Before:
5 failures in 20, and 2 in 12.

No new test. The existing one already asserts exactly the contract that was breaking — it was
failing one run in five — so the honest change here is to make it pass rather than to add a second
test beside it. What a new test could add is determinism, and I could not find a way to force the
delete-pending window open on demand without reaching past the API into filesystem internals.

## Risk

A genuinely blocked destination now takes up to 100 ms longer to report failure, on a path that
writes multi-megabyte capture bundles. Success is unaffected: the first attempt succeeds and the loop
exits.

### DIFF
--- a/prosper/src/gpu/capture/gpu_capture_bundle.cpp
+++ b/prosper/src/gpu/capture/gpu_capture_bundle.cpp
@@ -3,6 +3,7 @@
 #include <algorithm>
 #include <atomic>
 #include <chrono>
+#include <thread>
 #include <cstring>
 #include <filesystem>
 #include <fstream>
@@ -139,8 +140,29 @@ struct TemporaryBundleFile {
         file.close();
         if (!flushed || !file) { error = "cannot finish bundle temporary file"; return false; }
 #ifdef _WIN32
-        if (!MoveFileExW(payload.c_str(), target.c_str(), MOVEFILE_REPLACE_EXISTING)) {
-            error = "cannot install bundle: Windows error " + std::to_string(GetLastError());
+        // Replacing a file on Windows leaves the old destination briefly in a delete-pending state,
+        // and a second MoveFileEx onto it during that window is refused with ERROR_ACCESS_DENIED
+        // even though nothing is wrong with either writer. Two captures installing to one path is
+        // not hypothetical -- it is what the concurrent-writer contract in
+        // test_gpu_capture_bundle_streaming asserts, and it failed 5 runs in 20 without this (#3494).
+        //
+        // Bounded, and only for the codes that are transient by construction. A destination another
+        // process holds open for the whole call still fails after the budget, which is what the
+        // "OS sharing restriction refuses replacement" arm in that same test requires: a retry loop
+        // that never gave up would turn that refusal into a hang.
+        DWORD install_error = 0;
+        for (unsigned attempt = 0; attempt < 50; ++attempt) {
+            if (MoveFileExW(payload.c_str(), target.c_str(), MOVEFILE_REPLACE_EXISTING)) {
+                install_error = 0;
+                break;
+            }
+            install_error = GetLastError();
+            if (install_error != ERROR_ACCESS_DENIED && install_error != ERROR_SHARING_VIOLATION)
+                break;
+            std::this_thread::sleep_for(std::chrono::milliseconds(2));
+        }
+        if (install_error) {
+            error = "cannot install bundle: Windows error " + std::to_string(install_error);
             return false;
         }
 #else
--- a/prosper/src/gpu/capture/gpu_capture_bundle.cpp
+++ b/prosper/src/gpu/capture/gpu_capture_bundle.cpp
@@ -150,10 +150,15 @@ struct TemporaryBundleFile {
         // process holds open for the whole call still fails after the budget, which is what the
         // "OS sharing restriction refuses replacement" arm in that same test requires: a retry loop
         // that never gave up would turn that refusal into a hang.
+        // Success is the CALL's return value, never the error code. Deciding from GetLastError()
+        // would report success for a failing call that happened to leave it at zero -- a silent
+        // success the unretried version could not reach, and the destructor would then delete the
+        // temporary that was never installed.
+        bool installed = false;
         DWORD install_error = 0;
-        for (unsigned attempt = 0; attempt < 50; ++attempt) {
+        for (unsigned attempt = 0; attempt < 50 && !installed; ++attempt) {
             if (MoveFileExW(payload.c_str(), target.c_str(), MOVEFILE_REPLACE_EXISTING)) {
-                install_error = 0;
+                installed = true;
                 break;
             }
             install_error = GetLastError();
@@ -161,7 +166,7 @@ struct TemporaryBundleFile {
                 break;
             std::this_thread::sleep_for(std::chrono::milliseconds(2));
         }
-        if (install_error) {
+        if (!installed) {
             error = "cannot install bundle: Windows error " + std::to_string(install_error);
             return false;
         }


mattias800 (commented):
**APPROVED** on head `644a877f`. I worked each of the five risks separately and none of them lands; there is one latent structural point below that I would fix but do not consider blocking.

## Can the retry mask a genuine, permanent failure?

No — it delays one, it never swallows one. `gpu_capture_bundle.cpp:159-161` re-reads `GetLastError()` on every attempt and `:164-167` reports whatever the final attempt saw, with the same message and the same code as before. A permanently `ERROR_ACCESS_DENIED` destination (denying ACL, read-only parent) is still reported as `Windows error 5`; it arrives ~100 ms later. Every other code still fails on the first attempt via the `break` at `:160-161`, so the pre-existing prompt-failure behaviour is intact for the whole rest of the error space — including the obstruction arm at `test_gpu_capture_bundle_streaming.cpp:241-243`, which is a directory in the way and returns `ERROR_ACCESS_DENIED`... and therefore *does* now spend the budget. Still reported, still preserves the obstruction, `ctest` TIMEOUT is 60 s (`CMakeLists.txt:2224`). Fine.

## Is `ERROR_SHARING_VIOLATION` safe to retry?

Yes, and for the same reason as `ERROR_ACCESS_DENIED`: both mean *someone else currently holds a handle*, which is a statement about a moment, not about the destination. Neither can encode a condition that will never clear, so neither can be turned from a fast permanent failure into a hang — only into a slower failure, bounded at 50 × 2 ms. There is no "should fail fast" class hiding inside it that I can find; a caller cannot distinguish "held forever" from "held for 3 ms" at the API, which is exactly why a bounded retry rather than a predicate is the right shape here.

## Is `install_error` right on every exit path?

Correct on all three real paths, and I traced them individually:
- success on any attempt → `:156` sets `install_error = 0` then breaks → `:164` false → `return true`;
- non-retryable code → `:159` records it, `:161` breaks → reported;
- retryable and budget exhausted → the 50th iteration records at `:159`, sleeps at `:162`, loop condition ends it → reported.

**The latent one (non-blocking, one line):** the loop now decides success from the *error code* rather than from `MoveFileExW`'s return value. If a failing `MoveFileExW` ever left `GetLastError()` at `0`, `install_error` stays `0`, `:160` sees a code that is neither retryable value and breaks, and `:164` then reads it as success — `install()` returns `true`, the caller believes the bundle was written, and `~TemporaryBundleFile` (`:104-108`) removes the temporary directory, discarding the capture with no error anywhere. The old code could not do this because it keyed on the boolean. I could not construct a real path to it (`RtlNtStatusToDosError` maps unknown statuses to `ERROR_MR_MID_NOT_FOUND`, not `0`, and nothing runs between the call and the read), so I am recording it as a structure that permits a silent-success outcome rather than a demonstrated bug. `bool installed = false;` set in the success branch, and `if (!installed)` at `:164`, removes the class entirely and keeps `install_error` purely for the message.

## `<thread>` and `sleep_for` on a real-time-ish thread

Acceptable, but the concern is real enough to state where it lands. Of the three production callers, `gpu_timeline.cpp:2226` is a dedicated writer thread (`InteractiveFrameBundle::run_writer`) and `:1087` is an explicit checkpoint call — neither matters. `:3711` is inside `record_gpu_timeline_submit`, i.e. the submit path. Two things make it proportionate there: that path is gated on `gpu_timeline_requested()` at `gpu_timeline.cpp:3170` so it does not exist on a default run, and the same thread already performs the whole synchronous multi-megabyte `std::ofstream` write of the bundle immediately before this point — a bounded 100 ms in the *failure* case is small against that, and zero in the success case. I would not ask for a deadline-based loop or a backoff here.

Include placement is a nit: `:6` puts `<thread>` between `<chrono>` and `<cstring>`, breaking the alphabetical block. Move it after `<limits>`. It is also unconditional for a `_WIN32`-only use; harmless.

## Does the "OS sharing restriction refuses replacement" arm still genuinely refuse?

Yes, genuinely — not merely slowly. `test_gpu_capture_bundle_streaming.cpp:223-225` opens the destination with `FILE_SHARE_READ | FILE_SHARE_WRITE` and no `FILE_SHARE_DELETE`, and `:233` closes it only *after* `write_gpu_capture_bundle` has returned. The obstruction is therefore held for the entire call, so no budget of any size rescues it; the retry can only exhaust. The two consequent assertions at `:230-232` — the previous destination survives, and the temporary artifacts are cleaned — are unaffected by how long the loop took. What the arm does *not* pin is the budget's magnitude: it would pass equally if someone raised 50 to 5000, since it asserts termination rather than promptness. That is a limit on the test, not a defect in this change, and the code comment at `:150-152` claims only what is true (a loop that never gave up would hang).

## On the missing test

I agree with the decision, and the reasoning survives checking. The failing assertion at `:264` is already the contract; making it pass is the fix. Forcing the delete-pending window open on demand would need something below `MoveFileEx`, and a test that faked the window would be testing the fake. The 40/40 evidence is the right substitute, and the #3492 A/B (12/12 both arms, identical) is a properly controlled ruling-out rather than an assertion.

One suggestion for the record, not for this PR: the falsification "the intermittent `gpu_capture_bundle_streaming` failure is not caused by #3492's `gpu_timeline.cpp` shutdown change" is the kind of result the charter wants landed somewhere durable. It currently lives only in this PR body and #3494. Worth a line wherever this subsystem keeps its ruled-out notes, so the next person seeing a flaky capture test does not re-run those 24 builds.

mattias800 (commented):
**APPROVED** on head `c5ba18dd`. Re-reviewed by content against `origin/main`; the diff is one file and is exactly what I approved at `644a877f` plus the fix below, so my previous review carries over.

The latent point is properly closed. `gpu_capture_bundle.cpp:155-165`: success is now `installed`, set only inside the `MoveFileExW` branch, and `install_error` has been demoted to carrying the message. Re-walked the exit paths on the new text — success returns `true` with no reference to the error code; a non-retryable code breaks with `installed` false; an exhausted budget leaves `installed` false. The path I described, a failing call that left `GetLastError()` at zero, now reports `Windows error 0` as a failure rather than reporting success and letting `~TemporaryBundleFile` delete a temporary that was never installed. The comment at `:150-153` records why the shape is what it is, which is the part that stops it being re-simplified later.

`attempt < 50 && !installed` is redundant against the `break` — harmless, and I would leave it. The `<thread>` include at `:6` is still between `<chrono>` and `<cstring>` rather than in alphabetical position; still a nit, still not worth a push.

I did not re-run your builds. On the rebase: whether `capture_tunables` is now green is a claim about the merged #3492, and CI is the right place to settle it — but note that a green run there is evidence the deadlock *was* what failed it, not evidence it was the *only* thing, since that job never got far enough to tell you about anything behind it. If it comes back green, the honest statement is "nothing else surfaced", which is what you want anyway.

Merge when CI is green at this head.

Issue #3503, opened 2026-09-09
refactor: split hle_kernel_mem.cpp (7,345 lines) into per-platform files — the two-arms-in-one-file shape has caused 4 one-arm-read errors


mattias800:
## `hle_kernel_mem.cpp` is 7,345 lines / 416 KB and is really two parallel implementations in one `#if/#else`

Measured on `6771a2c0d`:

| lines | what |
| --- | --- |
| 1 – 633 | shared prologue: includes, helpers, tracking state |
| **634 – 3583** | **POSIX arm** (`#if defined(__linux__) \|\| defined(__APPLE__)`) — 2,950 lines |
| 3584 | `#else` |
| **3585 – 7344** | **Windows arm** — 3,760 lines |
| 7345 | `#endif` |

`src/hle/memory/` otherwise holds only three small headers, so this one file is the directory.

## Why this is a correctness problem, not tidiness

The file **documents its own trap, twice**, in identical comments at `:3455` and `:7265`:

> POSIX ARM. There are TWO definitions of this function in this file, ~3,200 lines apart, and the
> only thing distinguishing them is the `#if defined(__linux__) || defined(__APPLE__)` that opens far
> above. Nothing local tells you which arm you are in, which is the mechanism behind three separate
> one-arm-read-as-the-whole-file errors in a single session (#2376, #2384).

So the shape has already produced **three** wrong readings in one session, and the mitigation shipped
was a comment. A comment cannot fix it, because the failure is that a reader who greps finds one of
two hits and has no local signal that the other exists. Concretely, every HLE registration must be
added **twice**, and a `grep -n 'R("sceKernel…'` that returns one line looks complete.

This bit again on 2026-09-09 while adding the missing kernel memory NIDs for #3502 — the
double-registration was caught only because the comment happened to be read.

## Proposed split

```
src/hle/memory/
  hle_kernel_mem_common.hpp     lines 1-633: includes, shared helpers, tracking state
  hle_kernel_mem_posix.cpp      lines 634-3583
  hle_kernel_mem_win32.cpp      lines 3585-7344
```

Each arm then has exactly one `register_kernel_mem_hle()`, and its filename says which platform it
is. `CMakeLists.txt:71` is `file(GLOB_RECURSE … src/*.cpp)`, so no source-list edit is needed; the
per-file `#if` guard stays so the non-target arm compiles to nothing.

**Use `prosper/tools/refactor/split_file.py`**, which exists for exactly this and proves the result
[...]

Issue #3500, opened 2026-09-09
[Game tracker] NINJA GAIDEN 4 (PPSA25258)


mattias800:
## Current state — **rung 0**

First run of this title, 2026-09-09, on `6771a2c0d` (built from `origin/main`, `-DPROSPER_APP=ON`,
RelWithDebInfo). Default launch, no environment variables.

Zero frames presented — no `[app] … fps` line in the run. The guest links and boots, the window
opens, and the primary guest thread dies during early runtime init:

```
[prosper] unimplemented: libkernel::qCSfqDILlns  -> returning 0
[prosper] unimplemented: libkernel::pU-QydtGcGY  -> returning 0
[prosper] unimplemented: libkernel::Vzl66WmfLvk  -> returning 0
[app] guest booted; presenting its frames.
[app] window title: "prosper - NINJA GAIDEN 4"
[app] guest thread ended: kind=2 detail=SIGSEGV at addr=0x118  rip=0x4117e87c1 (eboot+0x17e87c1)
      rax=0x30 rbx=0x4144cc0c8 rdi=0x0 rsi=0x30 rdx=0x0
[app] guest backtrace: 0x4117d51f4 (eboot+0x17d51f4)
[app] guest backtrace: 0x5c0005f3a (libc.prx+0x5f3a)
[app] guest backtrace: 0x5c003ac49 (libc.prx+0x3ac49)
[app] guest backtrace: 0x5c003ab94 (libc.prx+0x3ab94)
[app] guest backtrace: 0x410000091 (eboot+0x91)
```

**`rdi=0x0` and `addr=0x118`** — a null base dereferenced at a fixed struct offset, not a wild
pointer. The frame below it runs through `libc.prx`, so this is C-runtime/static-init time, before
the title does anything of its own.

## Leading hypothesis: the whole `sceKernelMemoryPool*` family is missing

Resolved against the PS5 3.20 reference; all three calls immediately preceding the fault are one API
family, and **none of it exists in prosper**:

| NID | resolves to | hits in `prosper/src` |
| --- | --- | --- |
| `qCSfqDILlns` | `sceKernelMemoryPoolExpand` | **0** |
| `pU-QydtGcGY` | `sceKernelMemoryPoolReserve` | **0** |
| `Vzl66WmfLvk` | `sceKernelMemoryPoolCommit` | **0** |

(`sceKernelMemoryPoolDecommit` is also 0 — the family is entirely absent.)

[...]

Issue #3498, opened 2026-09-09
[Game tracker] FINAL FANTASY TACTICS - The Ivalice Chronicles (PPSA21783)


mattias800:
## Current state — **rung 0**, and the guest aborts *itself* within a second

First run of this title, 2026-09-09, on `6771a2c0d` (built from `origin/main`, `-DPROSPER_APP=ON`,
RelWithDebInfo). Default launch, no environment variables.

Zero frames presented — no `[app] … fps` line ever prints. The guest links, boots, opens its window,
then calls `sceKernelDebugRaiseExceptionOnReleaseMode` and terminates. This is the title deciding to
die, not a crash we caused directly.

The last five lines of the run, in order and consecutive:

```
28: [app] guest booted; presenting its frames.
30: [prosper] unimplemented: libkernel::n1-v6FgU7MQ  -> returning 0
31: [prosper] unimplemented: libkernel::tZ2yplY8MBY  -> returning 0
32: [amm] direct-memory pool [0x4200000,0x400000000) (16318 MiB, type 1, align 0x200000)
33: [prosper] guest sceKernelDebugRaiseExceptionOnReleaseMode(code=0xa0020001, arg=0x0) -- terminating
```

## Leading hypothesis: an unregistered memory-sizing call, answered as success with no out-parameter written

Resolved against the PS5 3.20 reference:

| NID | resolves to | implemented in prosper? |
| --- | --- | --- |
| `n1-v6FgU7MQ` | **`sceKernelConfiguredFlexibleMemorySize`** | **no — 0 hits in `prosper/src`** |
| `tZ2yplY8MBY` | **`sceKernelGetPageTableStats`** | **no — 0 hits in `prosper/src`** |

Both are value-returning contracts with out-parameters. The dispatcher's `return 0` answers `SCE_OK`
and **writes neither out-parameter**, so the guest reads untouched stack as its memory budget and
aborts two calls later.

**This exact defect has already been fixed once in this file, for the sibling function.**
`prosper/src/hle/memory/hle_kernel_mem.cpp:1682-1687`:

> `sceKernelAvailableFlexibleMemorySize(size_t* sizeOut)` … Was MISSING -> the return-0 stub left
> `*sizeOut` uninitialized, so the guest read garbage as its budget (Unity/allocator sizing) ->
> either a wild over-commit or a refusal to allocate. We don't pool-account flexible memory, so
> report the configured 512 MiB pool (shadPS4 parity)

[...]

Issue #312, opened 2026-07-10
UE4 (PPSA17942 DOLL): MallocBinned3 heap-corruption fatal during menu-driven content load (post-#306 next gate)


mattias800:
## Context

With the #306 offline/signed-out chain honest (branch `fix/issue-306-frontend-progression`),
DOLL reaches an interactive title screen and its menus. Driving the menus with
`PROSPER_PAD_SCRIPT` (e.g. `15:cross;20:start;25:cross;...`) always ends in a UE4 fatal during
the menu-triggered content-load burst (~t=60s, during heavy pakchunk0 APR reads):

```
LowLevelFatalError [File:Unknown] [Line: 152]
MallocBinned3 Corruption Canary was 0x3, should be 0x1
```

followed by the UE4 crash handler printing an UnknownFunction backtrace and a worker-thread
SIGSEGV (jump to NULL, rip in the rwxp region at 0x500000000). In some runs the rest of the
engine keeps ticking (draws/frame ~356 for another 60+s) before the run dies; the game flow
stops progressing either way.

## Evidence

- Reproduced in EVERY input-driven run: r5 (PROSPER_PAD_PRESS=1 held CROSS), r6/r8/r9
  (PROSPER_PAD_SCRIPT), across builds BEFORE and AFTER the iteration-2 entitlement fixes —
  so it is NOT caused by the #306 entitlement/savedata changes.
- Never occurs in no-input runs (r2: 240 s at the title steady state, clean).
- Logs on the diagnosing machine: /root/doll306_r{5,6,8,9}.log (fatal at ~line 143137 in r9);
  immediately preceded by a guest nullpage read of address 0x1 at eboot+0x231012b.
- The corrupted canary byte is 0x3. Suspects worth ruling out first: HLE out-writes made
  during this flow with unpinned ABIs — the unimplemented (success+garbage) calls still in the
  flow: sceKernelAioSubmitWriteCommands/AioWaitRequest/AioDeleteRequest (x2 each, save-write
  path), sceKernelAioSubmitReadCommands (CRI FS movie path), sceSaveDataSetParam (x3),
  sceGameLiveStreamingInitialize, sceShareFeaturePermit. scePlayGoGetLocus writes locus byte 3
  (= the corrupt value) but was only ever called with n=1 into 1-byte buffers (1,004 calls, all
  `(h, ids, 1, out)`).

## Repro

```
[...]

PR #3505, opened 2026-09-09
libkernel: register the memory-pool and configured-flexible-memory calls — fixes NINJA GAIDEN 4's null deref (#3502)

### LINKED ISSUE #3502: Unregistered kernel NIDs answer SCE_OK without writing their out-parameters — three of five new titles die on it (memory-pool and flexible-memory families entirely absent)
## Three of five newly-added titles fail on the same defect class, and one file already documents the fix

Early observation from a first-boot sweep of five new dumps on 2026-09-09, `6771a2c0d`. Filed at
`area:infra` rather than on any one tracker because it is one shared cause reached by three titles.

**The class:** an unregistered NID falls to the dispatcher's `return 0`, which answers `SCE_OK` while
**never writing the call's out-parameter**. The guest reads untouched stack as a pointer or a size
and proceeds. This is the same shape as #2951 (Metaphor divided by a glyph size read from an
untouched stack slot) and the `libSceUlt` note that a registered stub returning a sentinel can be
worse than no registration at all.

| title | tracker | NID | resolves to | hits in `prosper/src` |
| --- | --- | --- | --- | ---: |
| FF Tactics | #3498 | `n1-v6FgU7MQ` | `sceKernelConfiguredFlexibleMemorySize` | **0** |
| FF Tactics | #3498 | `tZ2yplY8MBY` | `sceKernelGetPageTableStats` | **0** |
| NINJA GAIDEN 4 | #3500 | `pU-QydtGcGY` | `sceKernelMemoryPoolReserve` | **0** |
| NINJA GAIDEN 4 | #3500 | `Vzl66WmfLvk` | `sceKernelMemoryPoolCommit` | **0** |
| NINJA GAIDEN 4 | #3500 | `qCSfqDILlns` | `sceKernelMemoryPoolExpand` | **0** |
| Darksiders II | #3497 | `F0Y42t-3e18` | `sceAgcDriverInitResourceRegistration` | **0** |
| Darksiders II | #3497 | `ZIXln2K3XMk` | `sceAcmContextCreate` | **0** |

(`sceKernelMemoryPoolDecommit` is also 0 — that family is entirely absent.) NIDs resolved against the
local PS5 3.20 firmware reference.

**Symptoms, in the order each title reaches them:**

- **FF Tactics** calls the two sizing functions, then raises
  `sceKernelDebugRaiseExceptionOnReleaseMode(code=0xa0020001)` two lines later and terminates itself.
- **NINJA GAIDEN 4** calls all three pool functions, then SIGSEGVs at `addr=0x118` with **`rdi=0x0`** —
  a null base at a fixed struct offset, which is the shape of an unwritten reservation pointer being
  carried away and dereferenced. The frame below is `libc.prx`, so this is C-runtime init.
- **Darksiders II** SIGSEGVs at a null address with `rsi=0x0`.

None of the three presents a single frame.

## The fix pattern already exists in-tree

`prosper/src/hle/memory/hle_kernel_mem.cpp:1682-1687` fixed **exactly this** for the sibling function
and says so:

> `sceKernelAvailableFlexibleMemorySize(size_t* sizeOut)` … Was MISSING -> the return-0 stub left
> `*sizeOut` uninitialized, so the guest read garbage as its budget (Unity/allocator sizing) ->
> either a wild over-commit or a refusal to allocate.

`k_avail_flexible` writes a value and returns 0. The **`Configured`** variant of that same API family
was never registered, and the `MemoryPool*` family was never registered at all.

## Suggested fix

Register the listed calls beside their existing siblings, writing their out-parameters honestly and
**deriving** values rather than hardcoding where they are derivable from local state.

## What is NOT established

**That any of these is the cause of its title's failure.** Each is an ordering argument — the calls
immediately precede the fault — plus a mechanism that is documented for a sibling function. Nothing
here disassembles the faulting PCs (`eboot+0x17e87c1` on NG4, `eboot+0x102e0ad` on Darksiders), and
`0xa0020001` is undecoded.

The discriminator is cheap and identical for all three: implement, re-run a default launch, and see
whether the fault moves. **Where it does not move, that belongs in the title's `## Ruled out`, not in
a quiet retry** — three titles sharing a plausible cause is exactly the setup where one confirmation
gets generalised to all three without being checked.

mattias800:
Fixes #3502. Refs #3500, #3498.

## The defect

`sceKernelMemoryPool{Expand,Reserve,Commit}`, `sceKernelConfiguredFlexibleMemorySize` and
`sceKernelGetPageTableStats` were unregistered. The dispatcher's `return 0` answers `SCE_OK` **and
writes no out-parameter**, so the guest reads its own uninitialised stack as an address or a size.

`hle_kernel_mem.cpp:1682` already documents this exact defect being fixed once, for the sibling
`sceKernelAvailableFlexibleMemorySize`. These are the same family, unfixed.

## Failure scenario

NINJA GAIDEN 4 (`PPSA25258`) calls `Reserve` during C-runtime init, never receives an address, passes
the untouched value to `Commit`, and dereferences null at `+0x118` — `rdi=0x0`, below a `libc.prx`
frame. Zero frames presented, every run.

## The ABI was captured, not assumed

A wrongly-positioned out-parameter write is **worse** than the stub it replaces (the `libSceUlt`
lesson: a registered stub can be worse than none). So the first commit registered these as log-only
probes — returning 0 and writing nothing, i.e. behaviour byte-identical to the unregistered path —
purely to observe the guest:

```
[probe-3502] sceKernelMemoryPoolExpand  a0=0x0 a1=0x400000000 a2=0xa6c00000 a3=0x10000 a4=0x41436b750
[probe-3502] sceKernelMemoryPoolReserve a0=0x1000000000 a1=0x400000 a2=0x400000 a3=0x0 a4=0x7fff33b45e08
[probe-3502] sceKernelMemoryPoolCommit  a0=0x0 a1=0x400000 a2=0xc a3=0x32 a4=0x0
[probe-3502] sceKernelConfiguredFlexibleMemorySize a0=0x410f99960
[probe-3502] sceKernelGetPageTableStats a0=0x7f22fc1ffc94 a1=...c90 a2=...c8c a3=...c88
```

This settled three things a guessed SDK signature would have got wrong:

- **`Reserve` takes the hint by value in `a0` with the out-pointer in `a4`** — not the single in/out
  pointer `sceKernelReserveVirtualRange` uses.
- **`a5` is caller-indeterminate scratch** on the five-argument entry points; the probe caught it
  holding a stack address two bytes below `a4`. It is never read — the same trap
  `sceKernelMapFlexibleMemory`'s `a4` already carries.
- **`GetPageTableStats` takes four 32-bit out-parameters**, from four consecutive stack addresses
  four bytes apart and descending.

It also upgraded the diagnosis from suggestive to confirmed: **`Commit`'s `a0` is `0x0`** — Reserve's
never-written out-parameter handed straight back. The null is visible in the arguments, not inferred
from call ordering.

## Approach

Reserve delegates to `k_reserve_vrange` and Commit to `k_map_flexible` (with `MAP_FIXED`) rather than
repeating them. `k_reserve_vrange` carries the #312/#946 huge-reservation steering and the #115
idempotent re-reserve, and NINJA GAIDEN 4 reserves at `0x1000000000` — the exact hint those
workarounds exist for. Only the calling convention differs, so it is adapted and forwarded.

Available and Configured flexible memory now answer from one `kFlexibleMemoryPoolBytes` constant, so
the pair cannot drift — the charter's rule that a value reachable through two APIs is derived in one
place.

Registered by **raw NID**, exactly as the guest imports them, so a name/hash mismatch cannot silently
restore the stub.

Both `register_kernel_mem_hle()` arms are updated — the POSIX one at `:3455` and the Windows one at
`:7265`. That file documents its own trap here (two definitions ~3,200 lines apart, three one-arm
read errors in one session); #3503 proposes splitting it.

## Verification

Build: green, RelWithDebInfo, `-DPROSPER_APP=ON`, **fresh configure** into a clean build directory,
`-DGAME_DUMP=<DUMP_ROOT>/PPSA24651-app0` so the three Messenger-specific cases execute rather than skip.

```
ctest --no-tests=error -j6
100% tests passed out of 459      rc=0      0 failed, 0 skipped
```

A first run in a reused build directory reported 127 failures and many `(Not Run)`. That result is
**withdrawn, not reconciled**: a full-tree build against that directory had been OOM-killed at
`-j16`, leaving test binaries missing, which is precisely the `(Not Run)` signature. The number above
is from a clean directory and is the only one quoted. (Instrument trap 243's advice — re-run from a
fresh configure before believing a surprising test failure — is what caught it.)

`PPSA25258`, `SDL_VIDEODRIVER=offscreen`, default launch, no environment variables:

| | before | after |
| --- | --- | --- |
| fault | `SIGSEGV addr=0x118 rdi=0x0` within seconds | **none in 120 s** |
| frames presented | 0 | 0 |

`PROSPER_MEMLOG=1` shows the API doing real work, including the second reserve being correctly
steered past the first:

```
pool_expand  range=[0x0,0x400000000) len=0xa6c00000 align=0x10000 -> phys=0x10000
pool_reserve hint=0x1000000000 len=0x400000   align=0x400000 -> 0x1000000000
pool_commit  addr=0x1000000000 len=0x400000 type=0xc prot=0x32 -> 0x0 (mapped 0x1000000000)
pool_reserve hint=0x1000000000 len=0xc0000000 align=0x400000 -> 0x1000400000
pool_commit  addr=0x1000400000 len=0x10000 ... -> 0x0   (and onward, 64 KiB at a time)
```

## What this does NOT do

- **It does not boot NINJA GAIDEN 4.** Still zero frames; the title now runs 120 s with no fault
  instead of dying early, i.e. it blocks somewhere later. **The rung does not move** — it stays at 0.
  A new gap surfaced behind it: `libkernel::LXo1tpFqJGs`, unregistered.
- **It does not fix FF Tactics.** `PPSA21783` still raises
  `sceKernelDebugRaiseExceptionOnReleaseMode(0xa0020001)` with **zero unimplemented NIDs remaining**.

  **What that rules out, precisely** (narrowed in review — the first draft of this bullet and of the
  #3498 comment said "the unregistered-NID hypothesis is falsified", which claims one step more than
  was measured): registering these two calls **and answering 512 MiB and four zeros** does not move
  the abort. It does **not** establish that the API family is uninvolved, because a wrong *value*
  produces the identical observation. `kFlexibleMemoryPoolBytes` is PS4/shadPS4 parity, now extended
  to a second API, for a PS5 title whose own log reports a **16,318 MiB** pool two lines above the
  abort.

  The discriminator that would separate them is cheap and has not been run: sweep the constant and
  see whether the abort is value-sensitive. Until then the honest status is "registration alone is
  not sufficient", not "the family is not involved".

  The contrast with NG4 still stands and is the useful part: there the guest's own arguments carried
  the null, so the chain was confirmed rather than inferred.
- **`sceKernelMemoryPoolDecommit` is deliberately not registered**, and the new gap this PR surfaces
  in NINJA GAIDEN 4 — `libkernel::LXo1tpFqJGs` — resolves to exactly that call. The title only
  reaches it *because* this change let it get further, so there was no captured ABI for it when this
  work started and guessing one is the thing this PR exists to avoid. It is the obvious next commit,
  and it needs its own probe.

  Worth flagging for whoever takes it: a family with Commit but no Decommit means the guest can
  claim pool memory and never release it, so this PR leaves NINJA GAIDEN 4 committing without a way
  to decommit. That is strictly better than the null deref it replaces and is not a resting place.
  Decommit also has **no out-parameter**, so the "worse than the stub" hazard does not apply to it —
  the risk there is unmapping the wrong range instead, which argues for validating against
  prosper's own tracked mappings before acting.

## Risk and confidence

- `CONFIDENCE: HIGH` on the argument layout — live-captured on the failing titles.
- `CONFIDENCE: MED` on pool accounting. prosper does not model a pool distinct from the direct-memory
  arena, so Expand draws from that arena and Commit backs the reserved VA with ordinary anonymous
  memory. The guest observes the contract it depends on; a title reading back pool-specific
  accounting would not be served.
- `CONFIDENCE: LOW` on `GetPageTableStats`' **values**. All four are answered 0 — "nothing measured".
  That is a *defined* answer where the guest previously read uninitialised stack, which is the defect
  class this PR addresses, but it is not a *correct* one and should not be mistaken for one. If a
  title divides by one of these or reads zero as "no capacity", the fix is to derive real figures
  from prosper's mapping tables, not to substitute a plausible-looking constant.

## Affected / unaffected

Only titles that call these NIDs change behaviour; every other title takes the identical path. The
one shared edit is `k_avail_flexible` now reading a named constant instead of an inline literal —
same value, no behavioural change.

## Review

Worth an independent read: this is reverse-engineered ABI semantics on a shared memory path, which
the charter names explicitly as review-warranted even when the diff is small.

### DIFF
--- a/BLOG.md
+++ b/BLOG.md
@@ -47,6 +47,23 @@ only difference.
 Alex Kidd's *world* is still wrong on this machine, and it is a different defect — the wave width
 makes no measurable difference to it either way.
 [#3479](https://github.com/mattias800/prosper/issues/3479).
+### Five new games, and three of them died the same way
+
+Five titles were added to the library and booted for the first time. *Syberia - The World Before*
+got furthest by a distance — its title screen and main menu render, and the audio is clean from the
+first second, over a splash that is still black. Three of the other four never presented a frame,
+and all three died of one thing: a kernel call prosper had never registered, answering "success"
+while writing nothing, so the guest read its own uninitialised stack as an address.
+
+### The argument that proved it was in the guest's own registers
+
+*NINJA GAIDEN 4* asks the kernel to reserve address space, and prosper's unregistered stub told it
+that worked without ever handing back an address. Rather than assume the call's shape from a
+published signature, we registered it as a probe that logs and does nothing — and the next call in
+the log asks to commit memory at address zero. That zero is the reservation that never came back,
+and a few instructions later it is the null the title crashes on. The crash is gone; the title now
+runs for two minutes without faulting and still draws nothing, which is a different and later
+problem. [#3502](https://github.com/mattias800/prosper/issues/3502)
 
 ## 2026-09-06
 
--- a/prosper/src/hle/memory/hle_kernel_mem.cpp
+++ b/prosper/src/hle/memory/hle_kernel_mem.cpp
@@ -1684,7 +1684,136 @@ HLE(k_map_flexible_noname) { return k_map_flexible(a0, a1, a2, a3, 0, 0); }
 // (Unity/allocator sizing) -> either a wild over-commit or a refusal to allocate. We don't pool-account
 // flexible memory, so report the configured 512 MiB pool (shadPS4 parity); a later map that exceeds host
 // memory still fails cleanly with ENOMEM.
-HLE(k_avail_flexible) { if (!a0) return 0x80020016ull; *(uint64_t*)(uintptr_t)a0 = 512ull * 1024 * 1024; return 0; }
+// The one flexible-memory pool size. Both the Available and the Configured entry points answer
+// from it (#3502): two literals here would let the pair drift, and a guest that sizes an
+// allocator from Configured and then checks it against Available would see the drift, not us.
+constexpr uint64_t kFlexibleMemoryPoolBytes = 512ull * 1024 * 1024;   // shadPS4 parity
+HLE(k_avail_flexible) { if (!a0) return 0x80020016ull; *(uint64_t*)(uintptr_t)a0 = kFlexibleMemoryPoolBytes; return 0; }
+
+// ---- PS4/PS5 memory-pool API (#3502) -----------------------------------------------------------
+// An alternative to AllocateDirectMemory + MapDirectMemory: Expand takes physical memory into the
+// process's pool, Reserve claims virtual address space, Commit backs a reserved range. All three
+// were UNREGISTERED, so the dispatcher answered SCE_OK and wrote no out-parameter -- NINJA GAIDEN 4
+// (#3500) called Reserve during C-runtime init, never received an address, passed the untouched
+// value to Commit, and died dereferencing it. Same class as sceKernelAvailableFlexibleMemorySize
+// above, whose comment records the identical defect being fixed once already.
+//
+// The argument layout is LIVE-CAPTURED from the guest (PPSA25258, 2026-09-09), not assumed --
+// writing an out-parameter through a wrongly-positioned argument would be worse than the stub:
+//   Expand  a0=searchStart a1=searchEnd a2=len       a3=alignment a4=physAddrOut
+//   Reserve a0=addrIn      a1=len       a2=alignment a3=flags     a4=addrOut
+//   Commit  a0=addr        a1=len       a2=type      a3=prot      a4=flags
+// Observed: Expand(0, 0x400000000, 0xa6c00000, 0x10000, out) -> Reserve(0x1000000000, 0x400000,
+// 0x400000, 0, out) -> Commit(0, 0x400000, 0xc, 0x32, 0). Commit's addr=0 IS Reserve's unwritten
+// out-parameter, which is the null that faulted -- the chain is confirmed by the guest's own
+// arguments, not merely by call ordering.
+//
+// a5 is caller-indeterminate scratch on the 5-argument entry points and must never be read -- the
+// same trap sceKernelMapFlexibleMemory's a4 carries (see k_map_flexible_noname above). The probe
+// recorded a5 holding a stack address two bytes below a4, i.e. pure garbage.
+//
+// CONFIDENCE: HIGH on the argument layout (live-captured on the failing title).
+// CONFIDENCE: MED on pool accounting: prosper does not model a pool distinct from the direct-memory
+// arena, so Expand draws from that arena and Commit backs the reserved VA with ordinary anonymous
+// memory. The guest observes the contract it depends on -- a reservation address it can commit and
+// then use -- but a title that reads back pool-specific accounting would not be served by this.
+
+// sceKernelMemoryPoolExpand(off_t searchStart, off_t searchEnd, size_t len, size_t align,
+//                           off_t* physAddrOut)
+HLE(k_pool_expand) {
+    if (!a4) return 0x80020016ull;                                  // SCE_KERNEL_ERROR_EINVAL
+    if (!a2 || (a3 && (a3 & (a3 - 1)))) return 0x80020016ull;        // len 0, or non-power-of-two align
+    const uint64_t align = a3 ? a3 : 0x4000;
+    uint64_t off = 0;
+    if (!dmem_take(a2, align, 0, off, a0, a1 ? a1 : ~0ull)) {
+        MLOG("pool_expand len=0x%llx align=0x%llx in [0x%llx,0x%llx) -> ENOMEM\n",
+             (unsigned long long)a2, (unsigned long long)align,
+             (unsigned long long)a0, (unsigned long long)a1);
+        return 0x8002000cull;                                        // SCE_KERNEL_ERROR_ENOMEM
+    }
+    dmem_zero(off, a2);                       // fresh allocation -> zeroed pages (console semantics)
+    *(uint64_t*)(uintptr_t)a4 = off;
+    MLOG("pool_expand range=[0x%llx,0x%llx) len=0x%llx align=0x%llx -> phys=0x%llx\n",
+         (unsigned long long)a0, (unsigned long long)a1, (unsigned long long)a2,
+         (unsigned long long)align, (unsigned long long)off);
+    return 0;
+}
+
+// sceKernelMemoryPoolReserve(void* addrIn, size_t len, size_t align, int flags, void** addrOut)
+// Delegates to k_reserve_vrange rather than repeating it: that function carries the #312/#946
+// huge-reservation steering and the #115 idempotent re-reserve, both of which cost live
+// investigations to get right, and NINJA GAIDEN 4 reserves at 0x1000000000 -- the exact hint those
+// workarounds exist for. Only the calling convention differs (hint by value + separate out-pointer,
+// versus one in/out pointer), so adapt and forward.
+HLE(k_pool_reserve) {
+    if (!a4 || !a1) return 0x80020016ull;                            // SCE_KERNEL_ERROR_EINVAL
+    uint64_t addr = a0;
+    const uint64_t rc = k_reserve_vrange((uint64_t)(uintptr_t)&addr, a1, a3, a2, 0, 0);
+    if (rc != 0) {
+        MLOG("pool_reserve hint=0x%llx len=0x%llx align=0x%llx flags=0x%llx -> 0x%llx\n",
+             (unsigned long long)a0, (unsigned long long)a1, (unsigned long long)a2,
+             (unsigned long long)a3, (unsigned long long)rc);
+        return rc;
+    }
+    *(uint64_t*)(uintptr_t)a4 = addr;
+    MLOG("pool_reserve hint=0x%llx len=0x%llx align=0x%llx -> 0x%llx\n",
+         (unsigned long long)a0, (unsigned long long)a1, (unsigned long long)a2,
+         (unsigned long long)addr);
+    return 0;
+}
+
+// sceKernelMemoryPoolCommit(void* addr, size_t len, int type, int prot, int flags)
+// Backs an address the guest has already reserved, so the mapping is FIXED regardless of the
+// caller's `flags` word -- that word carries pool flags, not the map flags k_map_flexible reads.
+// `type` (a2) is the direct-memory type; prosper does not model per-type pools, so it is logged and
+// otherwise unused. Passing a null name keeps track() from strncpy'ing a wild pointer.
+HLE(k_pool_commit) {
+    if (!a0 || !a1) return 0x80020016ull;                            // SCE_KERNEL_ERROR_EINVAL
+    uint64_t addr = a0;
+    const uint64_t rc = k_map_flexible((uint64_t)(uintptr_t)&addr, a1, a3, 0x10 /*SCE_KERNEL_MAP_FIXED*/,
+                                       0, 0);
+    MLOG("pool_commit addr=0x%llx len=0x%llx type=0x%llx prot=0x%llx -> 0x%llx (mapped 0x%llx)\n",
+         (unsigned long long)a0, (unsigned long long)a1, (unsigned long long)a2,
+         (unsigned long long)a3, (unsigned long long)rc, (unsigned long long)addr);
+    return rc;
+}
+
+// sceKernelConfiguredFlexibleMemorySize(size_t* sizeOut): the flexible budget this process was
+// CONFIGURED with, as distinct from what remains Available above. FF Tactics (#3498) reads it during
+// startup and raises sceKernelDebugRaiseExceptionOnReleaseMode two calls later when it comes back
+// unwritten. Both entry points answer from one constant so the pair can never drift -- the charter's
+// rule that a value reachable through two APIs is derived in one place.
+HLE(k_configured_flexible) {
+    if (!a0) return 0x80020016ull;                                   // SCE_KERNEL_ERROR_EINVAL
+    *(uint64_t*)(uintptr_t)a0 = kFlexibleMemoryPoolBytes;
+    return 0;
+}
+
+// sceKernelGetPageTableStats(int* out0, int* out1, int* out2, int* out3)
+// FOUR 32-bit out-parameters. Established from the guest (PPSA21783, 2026-09-09): the four
+// arguments were consecutive stack addresses four bytes apart and descending -- ...c94, ...c90,
+// ...c8c, ...c88 -- which is four adjacent `int`s passed in order, and a5 held an unrelated value.
+//
+// The FIELD SEMANTICS are NOT established. No published contract was consulted for the order or
+// meaning, and the guest's use of the values has not been disassembled. prosper does not account
+// page-table memory at all, so all four are answered 0: "nothing measured". That is a DEFINED
+// answer where the unregistered path left the guest reading its own uninitialised stack, which is
+// the whole defect class this change addresses -- but it is not a correct one, and it should not be
+// mistaken for one.
+//
+// CONFIDENCE: HIGH on the argument count and width (live-captured).
+// CONFIDENCE: LOW on the values. If a title divides by one of these, or treats a zero as "no
+// capacity", zero is wrong and the fix is to derive real figures from prosper's own mapping tables
+// -- not to substitute a plausible-looking constant, which would be fabrication.
+HLE(k_pagetable_stats) {
+    if (!a0 || !a1 || !a2 || !a3) return 0x80020016ull;              // SCE_KERNEL_ERROR_EINVAL
+    *(uint32_t*)(uintptr_t)a0 = 0;
+    *(uint32_t*)(uintptr_t)a1 = 0;
+    *(uint32_t*)(uintptr_t)a2 = 0;
+    *(uint32_t*)(uintptr_t)a3 = 0;
+    MLOG("pagetable_stats -> 0,0,0,0 (prosper does not account page-table memory)\n");
+    return 0;
+}
 
 // sceKernelAllocateDirectMemory(off_t start, off_t end, size_t len, size_t align, int memType, off_t* physOut)
 HLE(k_alloc_dmem) {   // (searchStart, searchEnd, len, alignment, memoryType, physAddrOut)
@@ -3467,6 +3596,14 @@ void register_kernel_mem_hle() {
     Hle::register_fn("4h6F1LLbTiw", (HleFn)k_map_flexible_noname,
                      "sceKernelMapFlexibleMemoryInternal");
     R("sceKernelAvailableFlexibleMemorySize", k_avail_flexible);   // was MISSING -> uninitialized budget
+    // #3502: registered by RAW NID exactly as the guest imports them. The name->NID hash is not
+    // re-derived here because these were resolved from the 3.20 reference against the live import
+    // list, and a hash mismatch would silently restore the return-0 stub this change exists to fix.
+    Hle::register_fn("qCSfqDILlns", (HleFn)k_pool_expand,        "sceKernelMemoryPoolExpand");
+    Hle::register_fn("pU-QydtGcGY", (HleFn)k_pool_reserve,       "sceKernelMemoryPoolReserve");
+    Hle::register_fn("Vzl66WmfLvk", (HleFn)k_pool_commit,        "sceKernelMemoryPoolCommit");
+    Hle::register_fn("n1-v6FgU7MQ", (HleFn)k_configured_flexible, "sceKernelConfiguredFlexibleMemorySize");
+    Hle::register_fn("tZ2yplY8MBY", (HleFn)k_pagetable_stats,    "sceKernelGetPageTableStats");
     R("sceKernelAllocateDirectMemory", k_alloc_dmem);
     R("sceKernelAllocateMainDirectMemory", k_alloc_main_dmem);  // 4-arg signature (physOut at arg3)
     R("sceKernelMapDirectMemory", k_map_dmem);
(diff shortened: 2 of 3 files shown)


mattias800 (commented):
# Independent review — head `9886c3cd`

**REJECTED**, on verification, not on correctness. I traced every argument-layout, out-parameter,
platform-arm and error-path concern below and **found no code defect** — the ABI work stands and does
not need redoing. What blocks is that five new handlers in the project's own named silent-failure
class ship with zero test arms, in a file whose own comment records three one-arm read errors in one
session, when the test file for exactly this class already exists and every comparable recent HLE
change shipped one. Details in §Blocking. Everything in §Verified is confirmed and should not be
re-derived by whoever picks this up.

---

## Verified correct (do not redo)

**1. Delegation argument mapping — correct on BOTH arms.** I read the callees rather than trusting
the comment.

- `k_reserve_vrange` is `(void** addrInOut, size_t len, int flags, size_t align)` — POSIX
  `hle_kernel_mem.cpp:1577`, Windows `:6513`, and it reads `fixed` from `a2 & 0x10` and `align` from
  `a3` in both. `k_pool_reserve:1751` forwards `k_reserve_vrange(&addr, a1, a3, a2, 0, 0)`: guest
  `a3`=flags lands in the flags slot, guest `a2`=alignment lands in the align slot. **Not
  transposed.** The two arms have the same parameter order, so the verbatim-duplicated call is right
  in both.
- `k_map_flexible` is `(void** addrInOut, size_t len, int prot, int flags, const char* name)` —
  `:1661` / `:6544`. `k_pool_commit:1772` forwards `k_map_flexible(&addr, a1, a3, 0x10, 0, 0)`: guest
  `a3`=prot into the prot slot, `0x10`=MAP_FIXED into flags, **name forced null**. Forcing the name
  is not incidental — it is the same trap `k_map_flexible_noname` exists for (`track()` strncpy's a
  wild `a4`), and getting it right here was easy to miss.

**2. Independent corroboration that Commit's `a2`/`a3` are not swapped.** The comment rests on the
probe alone; the file supplies a second check. Captured `type=0xc` is exactly the type-12 that
`valid_dmem_allocation`'s own comment records PS5 titles requesting ("Blasphemous 2 requests type
12"), and `prot=0x32` decodes through `host_prot:1311` to GPU_RW|CPU_WRITE. The transposed reading
would make prot `0xc` -> `PROT_EXEC` only, which the guest's first store into the committed arena
would have faulted on — the 120 s clean run excludes it. The layout also matches the independently
written shadPS4 signatures, which under the evidence hierarchy is hypothesis-grade but here agrees
with the live capture.

**3. All five NIDs verified** against the local 3.20 reference (`libkernel.c`), not just accepted:
`qCSfqDILlns`/Expand, `pU-QydtGcGY`/Reserve, `Vzl66WmfLvk`/Commit, `n1-v6FgU7MQ`/ConfiguredFlexible,
`tZ2yplY8MBY`/GetPageTableStats. `LXo1tpFqJGs` does resolve to `sceKernelMemoryPoolDecommit`, as the
body claims. `Hle::register_fn` takes `const std::string& nid` (`dispatch.cpp:99`), so raw-NID
registration is exactly equivalent to `R()`'s `nid_hash(name)` — no hazard in the mixed style. No
duplicate or shadowing registration of any of the five anywhere else in `src/`.

**4. Both arms identical.** POSIX 634-3721, Windows 3721-7619 (single top-level `#if/#else`, no third
arm). The 131-line handler blocks at `:1687` and `:6563` `diff` **empty**, and the five registration
lines at `:3607` and `:7551` are character-identical. This one is clean.

**5. No unvalidated-pointer write, and no clobber path.** `map_at:1408` claims a non-zero hint with
`MAP_FIXED_NOREPLACE` and only escalates to `MAP_FIXED` when `range_is_free_reservation()` says the
occupant is our own uncommitted reservation — so `Commit` on an address the guest never reserved
fails ENOMEM instead of destroying host memory. Windows `win_commit_flexible_exact` gates the same
way on `FlexibleHintState::GuestReservation`. `track()` splits and replaces overlapping entries, so
the reserved entry is *replaced* on commit rather than leaving a stale `!committed` twin that would
false-succeed the #115 idempotent re-reserve. I went looking for that one specifically; it is not
there.

**6. `dmem_zero` on the 2.6 GiB Expand is not a 2.6 GiB memset on Linux** — `:1459` is
`fallocate(PUNCH_HOLE)`. It *is* a real memset through a temporary view on Windows (`:5080`) and
macOS, but that is identical to what `k_alloc_dmem` already does for any large allocation, so it is
pre-existing shape, not a new hazard.

**7. Error codes** (`0x80020016` EINVAL, `0x8002000c` ENOMEM) match the values already used
throughout the file. `diff --check` clean, one commit, one file, zero `Claude-Session:` trailers.

---

## Blocking

### B1. Five new handlers, zero test arms — and the guard for this exact class already exists

`tests/hle/test_false_success_nids.cpp` was written for #2081 and its header states the case against
this PR's verification better than I can:

> An import prosper does not register returns the dispatcher's default 0. Where that 0 is also the
> contract's SCE_OK, the guest is told an operation succeeded that never ran... Nothing crashes and
> no diagnostic fires, so these regress silently — **which is exactly why they need a test rather
> than a boot.**

This PR is that class, and it is verified by a boot. The 459/459 green run is 459 tests none of which
can observe this change; that number is true and carries no information about the diff.

The idiom is already in the tree twice and needs no new infrastructure:
`test_false_success_nids.cpp:327` calls `register_builtin_hle()` then `Hle::lookup(<raw NID>)`, and
`test_reserve_placement.cpp:98` drives these very memory handlers via
`Hle::lookup(nid_hash("sceKernelReserveVirtualRange"))`.

Four arms, roughly 40 lines, each killing a mutation nothing currently kills:

- **Width.** `k_pagetable_stats:1808` writes `uint32_t` x4. The `CONFIDENCE: HIGH` is drawn from four
  stack addresses four bytes apart in a probe log — sound reasoning, but no artifact in the repo
  reddens if those become `uint64_t`, and that mutation silently corrupts 16 bytes of guest stack
  past the four ints. Prefill five adjacent `int`s with a sentinel; assert four written, **the fifth
  untouched**. This is the arm I would most want, because the boot cannot supply it: the only
  observed caller of this handler is FF Tactics, which still aborts.
- **Reserve's out-parameter.** Sentinel-fill `addrOut`, call, assert the sentinel is gone and the
  value is non-zero. That is literally NG4's null deref as a unit test — the defect the PR exists to
  fix, currently guarded by nothing.
- **The single-constant invariant.** Assert Configured and Available return the same value. The
  refactor at `:1690` exists so the pair cannot drift; nothing enforces it.
- **Registration.** `Hle::lookup(<each of the five NIDs>) != nullptr`. This is the only mechanism
  that runs on both platforms, and therefore the only thing that would redden if a future sixth pool
  call went into one arm — the failure this file documents happening three times in one session, and
  which #3503 is filed about. A comment saying "both arms are updated" protects the current commit,
  not the next one.

This is the project's standard, not my preference. The closest analogue, #3347 — also a pure
"register/answer honestly instead of falling to SCE_OK" change — shipped `test_psml_unimplemented.cpp`
(109 lines). #3295 shipped 184 test lines; #3348, #3262 and #3263 (the last two in *this* file) all
shipped tests. Ship the four arms and I will re-review promptly; nothing else in the code needs to
change.

### B2. "Falsified" is one step stronger than what was measured (PR body + #3498 comment)

The body says the unregistered-NID hypothesis for FF Tactics is **falsified**. What the experiment
established is narrower: *registration, answering 512 MiB and four zeros, does not move the abort.*
Those are not the same claim, and the gap is load-bearing because a `## Ruled out` row is read by the
next agent as a closed door.

The concrete residual: `kFlexibleMemoryPoolBytes` is 512 MiB carried over as shadPS4/PS4 parity, and
this PR extends it to a **second** API for a PS5 title. The log quoted in the very same #3498 comment
reports the title's direct-memory pool as **16,318 MiB** two lines above the abort. If FFT's check is
"configured flexible memory >= some PS5 minimum", 512 MiB fails it and the abort persists for a
*value* reason while the ruled-out row says the API family is not involved. That is the charter's
under-reporting rule ("under-reporting what is present is as much a defect as over-reporting... it
just fails in the direction that looks safe") landing on a title that is still failing.

Cheap and required: reword to "registration alone does not move the abort — the *values* returned
(512 MiB configured, four zero stats) are untested and remain a candidate", and add the same
sentence to the #3498 comment. I am not asking you to change the constant on a guess; I am asking
that the door not be recorded as closed on a hinge that was never tested. Sweeping `Configured`
across a range of values on PPSA21783 is a one-run discriminator whoever takes #3498 will want to
know is available.

---

## Non-blocking

- **`k_pool_expand:1723` validates more weakly than the sibling it is modelled on.** `k_alloc_dmem`
  gates on `valid_dmem_allocation()`, which enforces 16 KiB granularity on len/align **and**
  `phys_out > 0xffff` ("low values are not plausible guest pointers"). Expand checks only `!a4`,
  non-zero len and power-of-two align. A guest passing `a4 = 0x10` faults on the write at `:1735` —
  *after* `dmem_take` has consumed the pool, so the arena is also leaked on the way out. Reusing
  `valid_dmem_allocation` (arg order permitting) would close both.
- **The flags *encoding* equivalence is an untested assumption riding on a HIGH label.**
  `k_pool_reserve` forwards the pool `flags` word straight into `k_reserve_vrange`'s SCE map-flags
  slot, where `0x10` means MAP_FIXED. `CONFIDENCE: HIGH` is correct about argument *positions*, which
  is what was captured; that the two families use the same *bit* encoding is a separate claim, and
  the only observed value was 0, which exercises none of it. Worth one line in the comment so the
  next reader does not inherit the HIGH across the join.
- **Reservation leak on a bad `addrOut`.** `k_pool_reserve` commits the reservation before writing
  `*a4:1782`; a bogus non-null `a4` leaves an untracked-by-the-guest reservation behind. Same class
  as the Expand note, lower stakes.
- **Pool ranges are tracked as `kVirtualQueryFlexible` / name "flexible".** A guest doing
  `sceKernelVirtualQuery` over its committed pool sees flexible memory. Covered by the honest
  `CONFIDENCE: MED` on pool accounting — noting it so it is findable when some title does query.
- **No `BLOG.md` entry.** No screenshot is owed here (0 frames), but the charter is explicit that a
  finding with no picture is exactly what currently leaves no trace where the owner reads: NG4's
  fault disappearing and FFT's hypothesis dying in one PR is a good two-sentence entry.

---

## On the parts I was asked to challenge hardest

The probe-first method deserves saying out loud: registering these as behaviour-identical log-only
stubs to capture the ABI, *then* implementing, is the right way to approach a family where a
mis-positioned out-parameter write is worse than no registration — and it demonstrably paid, since
`Commit(addr=0x0)` upgraded the diagnosis from ordering to arguments. The body is also unusually
honest about its own limits: it withdraws the 127-failure run with the reason rather than reconciling
it, states plainly that the rung does not move, names the new `LXo1tpFqJGs` gap it created, and
refuses to guess Decommit's ABI. I have no complaint about overclaiming anywhere except B2, and B2 is
a wording fix.

`sceKernelGetPageTableStats` answering four zeros is defensible and I would not block on it: a
defined zero strictly dominates an uninitialised stack read, and the comment says so without
pretending the values are right. The right guard against the divide-by-zero / "no capacity" misread
is not a plausible-looking constant — the comment is correct to refuse that — it is the width arm in
B1 plus B2's wording, so that when a title *does* misread the zeros, the next agent finds a door
marked "untested" rather than one marked "ruled out".

**Verdict: REJECTED.** Add the test arms (B1) and correct the falsification wording (B2); the code
itself is sound and I will approve on the corrected head.

mattias800:
Both blocking findings addressed in `4c219b35`. Thank you — the width point in particular was correct and I would not have caught it.

## 1. Tests — added, and verified non-vacuous

Arms added to `tests/hle/test_false_success_nids.cpp`, each naming the mutation it kills: the Configured/Available shared-constant rule, out-parameter writes on Configured and Reserve, null-pointer rejection, and the four-slot width contract.

You were right that the width claim had nothing holding it. **Verified by mutation rather than asserted:** widening the four stores to `uint64_t` produces **4 failures, rc=1**; the restored source is byte-identical to the pre-mutation copy. So the arm is a discriminator, not an assertion that happens to be true.

Writing it also caught a real error *in the arm* — `kAvailableFlexibleMemorySize` had a NID from a scratch hash implementation of mine rather than the 3.20 reference value (`aNz11fnnzi4`). It failed loudly, which is the argument for the registration check being a hard `CHECK` rather than a guard.

Expand and Commit are covered for registration only. Exercising them would consume the direct-memory arena and map pages, making later arms depend on how much arena was left; that is stated at the arm.

## 2. "Falsified" — narrowed, on both the PR body and #3498

You are right that it claimed one step more than was measured. The corrected wording: registering the two calls **and answering 512 MiB and four zeros** does not move FF Tactics' abort. It does **not** establish the family is uninvolved, because a wrong *value* produces the identical observation.

Your 16,318 MiB observation sharpened it further, so I recorded the discriminator that would actually settle it: sweep `kFlexibleMemoryPoolBytes` across the plausible range and see whether the abort is value-sensitive at all. If it never moves, that is the falsification I should not have claimed yet.

## 3. Non-blocking: `valid_dmem_allocation` adopted

Taken, and it turned out better than a local check. `k_pool_expand` now validates through the sibling allocator's own validator **before** `dmem_take` — that function's comment already records this exact defect ("false success plus leaked physical capacity"), so the pointer-plausibility case you raised was already solved one layer down and I was reimplementing it. Length, granularity and pointer plausibility now come from one validator; only the non-power-of-two alignment check stays local, since it guards the `& ~(align-1)` rounding that validator does not cover.

## Verification note, stated plainly

The 459/459 I quoted was from a clean run **before** these two commits. My subsequent local runs are unreliable and I am not quoting them: I stacked several concurrent builds against one build directory (timed-out foreground calls kept running), which is instrument trap 243, self-inflicted. Every local "failure" since was `BAD_COMMAND` — missing executables — never a real assertion failure.

**CI on the new head is therefore the verification that counts**, and it is the uncontended environment. The previous head was 12/12 green including Windows MinGW.

Remaining non-blocking items not taken, deliberately: the flags-encoding equivalence note, reservation leak on a bad `addrOut`, pool ranges tracked as `flexible`, and the `BLOG.md` entry. Happy to take any of them here if you would rather they not become follow-ups.

mattias800 (commented):
# Independent re-review — head `4c219b35`

**APPROVED**, with one required behaviour-free correction (N1) that needs no re-review, and one
mechanical merge precondition (§Verification) that the charter already requires.

Both blocking findings from my review of `9886c3cd` are addressed substantively rather than
cosmetically. I re-verified the round-1 §Verified list still holds on this head and re-checked
everything the new commit touched.

---

## B1 — tests: addressed

99 lines, 25 checks, each naming the mutation it kills, matching the standard the file's own header
sets. Three things raise this above "a test was added":

- **The width arm is verified non-vacuous by mutation**, not by assertion. Widening the four stores
  to `uint64_t` yields 4 failures and rc=1, with the restored source byte-identical. That is the
  charter's "mutated counter-arm separates confirmed from coincidence", executed rather than
  claimed, and it is the strongest form of evidence available for an arm no boot can police.
- **The `aNz11fnnzi4` incident is itself the proof the arms are load-bearing.** A constant taken
  from a scratch hash rather than the reference made `Hle::lookup` return null and the arm failed
  loudly. I verified independently against the 3.20 reference: `aNz11fnnzi4` is
  `sceKernelAvailableFlexibleMemorySize`. Correct now. Writing the test found a real error in the
  first hour of its existence, which is the argument for having asked.
- Reserve's arm reproduces NINJA GAIDEN 4's actual bug as a unit test — sentinel in, assert the
  sentinel is gone and the address is non-null and aligned. That defect is now guarded by something
  other than a boot.

## B2 — wording: addressed, and better than what I asked for

PR body lines 104-120 and the #3498 comment both now say *"registration alone is not sufficient"*
rather than *"falsified"*, state that a wrong value produces the identical observation, name the
512 MiB / 16,318 MiB contrast, and specify the constant sweep as the discriminator that would settle
it. I asked for a door not to be recorded as closed; you turned it into a specified next experiment
with the values to try. That is a better artifact than the correction I requested.

## Hardening: nothing weakened, and the live call still passes

I checked the strictness delta rather than taking it on trust. Old: `a4 != 0`, `a2 != 0`, align
power-of-two. New: align power-of-two (kept local) **plus** `valid_dmem_allocation(a2, a3, 0, a4)`,
which gives `a2 != 0` **and** 16 KiB-granular, align 16 KiB-granular, and `a4 > 0xffff`. Every prior
check is preserved or strictly strengthened; **none is removed**. Both arms byte-identical (diffed),
and `valid_dmem_allocation` exists with identical semantics in both (`:1321` / `:4406`, both with
`kGuestPageSize = 0x4000`), so the Windows arm is not relying on a POSIX-only symbol.

**`memory_type = 0` is right.** The validator uses that parameter only as
`memory_type > kMaxDirectMemoryType (0x7fffffff)`, so 0 always passes — it is a no-op for a call that
carries no type argument, and it matches the `0` already handed to `dmem_take` on the next line. The
value validated and the value allocated are the same one; that is the property worth having, and it
holds.

One regression check the commit message does not state and I ran: **the observed NG4 call still
passes the stricter validator** — len `0xa6c00000` is 16 KiB-granular, align `0x10000` is granular
and a power of two, `a4 = 0x41436b750 > 0xffff`. The hardening does not reject the one call shape
there is live evidence for. Worth putting in the body.

---

## Your three questions

**Would the width arm catch narrower mutations?** Yes, via its companion check. `uint16_t` leaves
`0x5A5A0000`, `uint8_t` leaves `0x5A5A5A00`, and skipping a slot leaves full poison — all fail
"writes out{i}". Narrowing is covered.

**Can "leaves the upper 4 bytes alone" pass for a wrong reason?** Not for the mutation it targets.
Three qualifications, none blocking:

1. **It cannot see slot ORDER.** All four answers are 0, so out0..out3 are indistinguishable and any
   permutation of the field assignment passes. That is honest — the source comment says field
   semantics are unestablished — but the arm proves *width*, not *assignment*, and should not be
   read as more once someone derives real figures.
2. **The geometry is not the guest's.** The arm passes `&slot[0..3]` of a `uint64_t[4]`, i.e.
   pointers **8 bytes apart**; the captured frame is four ints **4 bytes apart and descending**
   (...c94/c90/c8c/c88). So the guard bytes sit where the *next out-parameter* really lives, and the
   real corruption — a `uint64_t` store at the highest slot writing 4 bytes *outside* the block —
   is not the one being reproduced. The arm still kills the mutation, which is what a mutation arm
   is for, so this is a fidelity note, not a defect. `int stats[5]` at 4-byte spacing with a fifth
   sentinel would give you the widening kill *and* a "wrote a fifth field" kill in the guest's own
   geometry.
3. **`CHECK(low == 0)` fuses "did it write" with "wrote zero".** Zero is the `CONFIDENCE: LOW`
   placeholder, and the source comment explicitly directs the next agent to replace it with derived
   figures — at which point this arm goes red for the *right* change. The Random arm in this same
   file deliberately keeps those two questions apart (`!= sentinel`, then `!= all-zero`). One comment
   line — "this pins today's placeholder; it moves when real figures are derived" — stops a future
   red being misread as a regression.

**Expand/Commit registration-only — is the trade wrong?** No, and the stated reason is the right one.
Two refinements:

- **Reserve is not state-free either**, so the principle applies to it partially: it consumes 64 KiB
  of guest VA and leaves a permanent `track()` entry in `g_maps` with no unmap. Fine in practice —
  VA is plentiful, the arm runs last, and `test_reserve_placement.cpp` already reserves in tests —
  but the comment reads as though Reserve were inert, and it is not.
- **Expand's rejection paths are pure and allocate nothing**: `len = 0`, non-power-of-two align,
  `a4 = 0`, `a4 = 0x10` all return EINVAL *before* `dmem_take`. Which means **the hardening this
  commit just added is itself untested**, and it is free to cover. That is the obvious next six
  lines and it closes the loop without touching the arena.

---

## N1 — required, behaviour-free: the null-slot arm claims more than it tests

```
CHECK(pts(&one, 0, 0, 0, 0, 0) != 0, "GetPageTableStats rejects a null in ANY of its four slots")
```

The handler is `if (!a0 || !a1 || !a2 || !a3)`, which short-circuits at `!a1`. Because a2 and a3 are
*also* null in this call, deleting the `!a2` or the `!a3` disjunct still returns EINVAL and the arm
still passes. It establishes rejection for one configuration and asserts it for four.

This is the failure mode the file's own header names — *"an assertion that passes for the wrong
reason is the recurring failure in this project"* — and it is the same shape as the finding you just
fixed. The fix is four lines: loop `i` in 0..3, pass null at position `i` and valid pointers
elsewhere. Please land it before merging; it changes no behaviour and needs no re-review from me.

## N2 — trivial

`test_kernel_memory_pool()` at `test_false_success_nids.cpp:334` sits **outside** the anonymous
namespace, which closes at `:332`; every sibling test function is inside it. Almost certainly
placement oversight — move it above the closing brace.

---

## Verification — my own view, since you asked for it explicitly

**The change is not yet verified, and one plausible shortcut to concluding otherwise is wrong.**

- The 459/459 from `9886c3cd` predates the test file and says nothing about it.
- Your post-459 local runs are correctly **withdrawn**, not reconciled — trap 243, self-inflicted,
  every failure `BAD_COMMAND` (missing executables) rather than an assertion. That is the right
  handling and I am not counting them in either direction.
- **The green "Host ASan + UBSan" and "Host TSan" jobs do NOT cover the new arms, and it would be
  easy to think they do.** `ci.yml:23` configures `-S prosper/tools/doctor` — an isolated project —
  and its `ctest --test-dir prosper/build-sanitizers --no-tests=error` runs over a build tree that
  does not contain `false_success_nids`. Two greens that look like coverage and are not.
- The test *is* registered in the main build (`prosper/CMakeLists.txt:1016-1018`, ctest name
  `false_success_nids`) and runs in the **Linux** job, which configures `-S prosper -B
  prosper/build-ci`. At the time of writing that job is still `in_progress`; 8 of 12 checks are
  green, 4 pending (Linux, macOS Rosetta, Windows App, Windows MinGW).

So the only execution evidence for any new arm is your width-arm mutation — real, targeted, and
sufficient for that arm. The other ~20 checks, including the five registration checks whose constant
was wrong when first written, have none yet.

**Merge precondition:** Linux green on `4c219b35`, and confirm `false_success_nids` is in the
executed set *by name and count*, not by exit code alone. Given the `aNz11fnnzi4` incident, a green
run of this file specifically is the thing to look at. Mechanical, and it does not need me.

## The four deferred items — none should block

- **flags-encoding caveat:** comment-only; the sole observed value (0) exercises none of it. Fine
  deferred.
- **Reservation leak on a bad `addrOut`:** genuine but needs a guest passing a plausible-looking
  bogus pointer. Issue-worthy, not merge-worthy.
- **Pool ranges tracked as `flexible`:** already carried honestly by `CONFIDENCE: MED`; surfaces only
  when some title actually queries.
- **`BLOG.md`:** not a correctness gate, so no block — but I would still write the two sentences in
  this PR. NG4's fault disappearing and FFT's hypothesis narrowing from "dead" to "registration alone
  is not sufficient" is precisely the *finding with no picture* the charter says currently leaves no
  trace anywhere the owner reads. One paragraph, and it is the only place this lands for them.

---

**Verdict: APPROVED** on `4c219b35`. Land N1 (and ideally N2 and the Expand-rejection arms) without
re-review, confirm the Linux job green with `false_success_nids` in the executed set, then merge.

mattias800 (commented):
# Approval carries to `e97c7a52` — head moved during the re-review

Mechanical note so the merge step is unambiguous. My re-review body names `4c219b35`, but the head
moved to `e97c7a52` while I was writing it, so GitHub attached the review to `e97c7a52`. I have
established by content that the approval still applies rather than assuming it in either direction.

`git diff 4c219b35 e97c7a52` touches **`BLOG.md` only** — the entry I recommended as a non-blocking
item. Nothing in `hle_kernel_mem.cpp` or `test_false_success_nids.cpp` changed, so every finding in
the re-review stands unaltered.

I read the entry. It is the right shape for this PR: image-free with the reason stated in the commit
message rather than silently, honest about what it is showing (*"still draws nothing, which is a
different and later problem"*), and it keeps the Syberia capture out of a kernel-fix PR so it can
land alongside its own screenshot. It says what you are looking at rather than what the change was,
which is the part of the charter's blog rule that usually gets missed. The `mattias800/prosper` issue
URL matches all 59 existing links in the file, so it follows the established convention — not a
finding against this PR either way.

Two things unchanged by this commit, restated so they are not lost across the head move:

- **N1 is still required before merge**: the `GetPageTableStats` null-slot arm asserts *"rejects a
  null in ANY of its four slots"* while exercising one configuration in which slots 1-3 are all
  null, so deleting the `!a2` or `!a3` disjunct still passes. Four lines: loop `i` in 0..3 with null
  at position `i`. Behaviour-free, no re-review needed.
- **The merge precondition stands.** Adding a docs commit does not make this a documentation-only
  PR — the diff still carries code, so the CI-wait exception does not apply. Linux must go green on
  `e97c7a52` with `false_success_nids` confirmed in the executed set by name and count, not by exit
  code alone.

**APPROVED** on `e97c7a52`.

Issue #3507, opened 2026-09-09
RectList fourth-corner synthesis assumes the shared corner is always P0, so a rectangle supplied in another order draws half a quad (Darksiders II intro FMV)


mattias800:
## Hypothesis: RectList synthesis is declined for a VERTEX-LESS RectList, so half the quad is drawn

Filed at `area:infra` rather than on one title because the precondition is on the shared draw path
and nothing about it is Darksiders-specific.

**Symptom.** Darksiders II (`PPSA23806`) reaches its New Game intro FMV and renders **only the
top-right triangle of the movie**. Reported by the project owner playing it, 2026-09-09; the title
otherwise renders its title screen and menu correctly at 3840x2160 (#3497).

**The precondition.** `needs_rect_list_synthesis` (`prosper/src/gpu/execute/gpu_execute.hpp:1503`)
requires a **vertex-buffer-backed** RectList:

```cpp
if ((primitive_type != 7u && primitive_type != 17u) || indexed || draw_count != 3u ||
    !vertex_resources)
    return false;
return std::any_of(vertex_resources->resources.begin(), vertex_resources->resources.end(),
    [](const ShaderResource& r) {
        return r.cls == ResourceClass::VertexBuffer && r.stride != 0u;
    });
```

`gpu_execute.hpp:97` states the assumption outright: *"A vertex-buffer-backed PS5 RectList is
submitted as three vertices."*

**Why that would produce exactly this symptom.** A full-screen movie blit is conventionally
*vertex-less*: the vertex shader derives clip-space positions from `gl_VertexIndex` / `V#`-free
arithmetic, so the draw carries no `VertexBuffer` resource with a non-zero stride. The `any_of`
above then fails, synthesis is declined, and the three RectList vertices are handed to Vulkan as a
plain triangle list — **one triangle instead of a rectangle**, which is half the quad.

A RectList's three vertices describe three corners of a rectangle; drawn as a triangle they cover
one diagonal half. That is the reported picture.

## NOT established

That this is the cause. Nothing here has observed the FMV draw — no `prim_type`, no resource table,
no capture. The reasoning is: the symptom is half a quad, the code path that turns a RectList into a
quad has a precondition a full-screen blit would plausibly fail, and the assumption behind that
precondition is written down as an assumption. That is a strong shape, not a measurement, and #3498
is this week's reminder that a plausible ordering argument can be wrong.

## Discriminator, one run

[...]

Issue #3497, opened 2026-09-09
[Game tracker] Darksiders II Deathinitive Edition (PPSA23806)


mattias800:
## Current state — **rung 0** (no graphics at all)

First run of this title, 2026-09-09, on `6771a2c0d` (built from `origin/main`, `-DPROSPER_APP=ON`,
RelWithDebInfo).

The guest links, boots and the app opens its window, then the **primary guest thread dies on a null
dereference before a single frame is presented**. The fps counter never prints once — zero
`[app] … fps` lines in the whole run — so this is rung 0, not a black-screen rung 1.

```
[app] guest thread ended: kind=2 detail=SIGSEGV at addr=(nil)
      rip=0x41102e0ad (eboot+0x102e0ad)
      rax=0x207f89e980 rbx=0x207f84e640 rdi=0x20403b9310 rsi=0x0 rdx=0x2
      rbp=0x7fadc9bffe90 rsp=0x7fadc9bffe70
[app] guest backtrace: 0x411030400 (eboot+0x1030400)
[app] guest backtrace: 0x4108feb9a (eboot+0x8feb9a)
[app] guest backtrace: 0x4110352a1 (eboot+0x10352a1)
[app] guest backtrace: 0x41000018b (eboot+0x18b)
[app] guest backtrace: 0x4100000bf (eboot+0xbf)
```

`rsi=0x0` with the fault at `addr=(nil)` suggests a null argument carried into the call rather than a
wild pointer, but nothing here disassembles `eboot+0x102e0ad` — that is the first step for whoever
picks this up (`tools/re/xref.py` for who reaches it).

## Dump

| | |
| --- | --- |
| title id | `PPSA23806` |
| version | `01.000.002` |
| size | 21 GB |
| middleware | THQ Nordic GDS (`THQGDSCore_Prospero_Release`), a `steam_api_ps5` shim, `psnwrapperreleaseprospero` |

Note the link is thin: **2 modules, 877 imports, and 655 of them are stub slots** — a far higher
unresolved fraction than the Unity titles booted the same day (BlazBlue: 12 modules, 4321 imports,
922 stubs). Ten plugins are absent from the dump and skipped (`AkSoundEngine`, `AkVorbisHwAccelerator`,
`AkMotion`, `libfmod`, `libfmodstudio`, `CommonDialog`, `PSNCommon`, `PSNCore`, `SaveData`, `PSN`).

## Reproduction

```
[...]

Issue #3279, opened 2026-09-03
GTA V: a routed world run on main takes 122-175 VK_ERROR_DEVICE_LOST and blacks out from a fixed point — any perf number from it is measuring recovery


mattias800:
## Summary

A routed GTA V run on current `main` sustains **122–175 `VK_ERROR_DEVICE_LOST`** and its world goes
black partway through. Reproduced twice. This is independent of #3264's HTILE question and matters
on its own, because **any performance number taken from GTA V's world on `main` is measuring a GPU
in repeated recovery.**

## Measured

`cd18bd6fb`, Linux / RADV / integrated Radeon 8060S, `tools/screenshot`, route
`scripts/gta5/reach-performance-story.pad` (native 4K Performance mode), 8 frames over the last 60 s
of a 580 s run.

| run | frames with content | device losses | distinct |
| --- | --- | --- | --- |
| control 1 | 3 / 8 | **122** | 3,532 / 25,354 |
| control 2 | 3 / 8 | **175** | 3,650 / 30,254 |

The failures are `[backend] GRAPHICS submission failed: submit=-4` (`VK_ERROR_DEVICE_LOST`), and
compute skips rise to 16 in the same runs.

**The blackout is positionally stable**: per-frame mean luminance is `73 115 116 0 0 0 0 0` in both
runs — the first three frames (menus) render identically and everything from frame 3 on is black.
So this is not random; it begins at a reproducible point in the route.

## Why this is filed separately from #3264

#3264's candidate change (removing the HTILE `!byte_preserving` suppression) takes these losses to
**0 and 0** across two runs. But that change **breaks GTA V's characters and lighting** — confirmed by
human review — so it is not the fix and is not merging.

That leaves the loss cascade unexplained and unattributed on its own terms. It may be a *consequence*
of the stale depth the suppression leaves, or an independent defect that the same change happens to
perturb. Nothing here distinguishes those.

## Why it matters now

`CLAUDE.md` currently describes GTA V's missing world as *"one compute program: `0x413dc6700` hangs
the GPU into a RADV hard recovery, which disables live compute process-wide and drops every later
indirect draw"*, with `PROSPER_COMPUTE_SKIP_PROGRAM` as the workaround. These runs set no such flag,
and the loss count varies 122 → 175 between otherwise identical runs, which is worth reconciling with
that account.

Separately, a **windowed** run on `main` earlier the same day reached the bank heist and rendered it
[...]

Issue #1635, opened 2026-08-01
Loader: a duplicate NID export is silently aliased to whichever module links first — no warning, no counter, no diagnostic


mattias800:
## Summary

`link_program` builds the global export table with

```cpp
// src/loader/linker.cpp (Global export table: NID -> guest address)
for (auto& s : m.symbols)
    if (!s.is_import && !s.nid.empty() && s.value != 0) {
        exports.emplace(s.nid, base + s.value);
        me.nids.emplace(s.nid, base + s.value);
    }
```

`std::unordered_map::emplace` is a **no-op when the key already exists**. So when two linked modules
export the same NID, the one appearing **earlier in the link input list wins**, the other's definition
is discarded, and **nothing is reported** — no `fprintf`, no counter in `Program`, nothing in the
`linked N modules; …` summary line. The header comment on `Program::exports` does say "first
definition wins", so the behaviour is intentional; the absence of any diagnostic is the problem.

## Why it matters

Two modules exporting the same NID is not hypothetical. Evergate (`PPSA01885`) ships **both**
`Media/Plugins/libfmod.prx` and `Media/Plugins/libfmodL.prx` — FMOD's release and logging builds of
the same library, exporting identical NIDs — and likewise `libfmodstudio.prx` + `libfmodstudioL.prx`.
Confirmed on a live boot: `libfmodL.prx` exports `zKjgSnqVqf4`, already provided by `libfmod.prx`;
`libfmodstudioL.prx` exports `Hh2wT63cPf4`, already provided by `libfmodstudio.prx`.

If both are ever linked, the consequences are all silent:

- Every guest import of a shared NID resolves to whichever module the preload list happens to name
  first — an ordering accident, not a decision.
- The loser is still **mapped** and its **`init_array` still runs**, so two full copies of the same
  library initialize (two FMOD cores).
- `sceKernelDlsym` consults the handle's own module table before the global one (#147), so the same
  symbol resolves to **different addresses** depending on which module handle it was asked through.

Today only the fixed preload list can produce this, and it happens not to name both variants. #1609
adds auto-discovery of `Media/Plugins/*.prx`, which would have — so that change carries an explicit
export-collision guard (`LinkInput::skip_on_export_collision`) that refuses such a module and logs
which NID collided and who owns it. **That guard only covers modules flagged optional.** A duplicate
between two *named* modules still aliases silently.

## Suggested fix

Report it regardless of who caused it. In the export-table loop, when `emplace` reports the key was
[...]

PR #3508, opened 2026-09-09
loader: link a title's own middleware from the dump root, and skip only duplicate BUILDS — Darksiders II reaches its title screen (#3497)

mattias800:
Refs #3497. Takes **Darksiders II (`PPSA23806`) from rung 0 to a rendered title screen and menu**, and unblocks a module BlazBlue was also losing.

**Progression evidence:** the title screen renders "DARKSIDERS II — DEATHINITIVE EDITION" with its Press-✕ prompt at **3840x2160**, 20 fps, `gpu-present, 20.0 distinct` — captured from `prosper-app` on a default launch, no environment variables. The project owner then drove the menu into a New Game intro FMV. Screenshot is at `~/newgames-run/ds2-screen3.png` locally and needs committing to `assets/screenshots/` with a `BLOG.md` entry; doing that in a follow-up so this PR stays a code change.

## The defect, diagnosed from the guest

The title died on a SIGSEGV before presenting a frame. Disassembling the fault shows a virtual-call setup on a null object — and the guest **nulls the pointer itself**:

```
102e061:  call 0x1f892e0        <- predicate
102e068:  je   0x102e08b        <- returns 0 ...
102e08b:  xor  r14d,r14d        <- ... so r14 = NULL
102e0ad:  mov  r8,[r14]         <- and is dereferenced
```

`0x1f892e0` is a PLT thunk. `tools/re/stub_nid_map.py` names the import behind it: `YLTUOiLWp-M`, which the boot log attributes to **`steam_api_ps5`** — a module **present in the dump**, sitting unlinked. prosper was standing in for code the dump provides, and its `return 0` is the zero the guest tested.

Two separate reasons it was unlinked.

## 1. Auto-link never looked at the dump root

It scanned only `<dump_root>/Media/Plugins`. Darksiders ships `steam_api_ps5.prx`, `THQGDSCore_Prospero_Release.prx`, `thqno_api_ps5.prx` and `psnwrapperreleaseprospero.prx` beside `eboot.bin`.

**The anti-circumvention guard is preserved, not widened.** `module_path_policy.hpp` says a widening "gets a reviewer", so this is the part to read closely. `kSonyLibraryPrefix` now refuses a `libSce*` module in the root with the **Sony** verdict, for a stronger version of its existing reason: the root is auto-scanned wholesale, so without it a dump could substitute a Sony library by dropping it beside `eboot.bin` — an easier bypass than the `Media/Plugins` one that rule already closes, and one that would sidestep the `fakelib/` rejection by a single move. Only `.prx`/`.sprx` are candidates, so data files are not. Test arms cover both directions including `.sprx` and case variants, plus a non-module root file.

## 2. The collision rule dropped whole modules over ONE shared export

`skip_on_export_collision` dropped a module the moment it shared any NID. That rule exists for two **builds** of one library — Evergate ships `libfmod.prx` beside `libfmodL.prx` — but "any collision at all" also drops two **different** libraries that share a symbol, and every export only one of them provides.

The discriminator is now subsumption, with the threshold set from **measurement**. claimed/total for every module the old rule skipped across the local corpus:

| module | claimed | what it actually is |
| --- | ---: | --- |
| Evergate `libfmodstudioL.prx` | **100.0%** (357/357) | duplicate build |
| Evergate `libfmodL.prx` | **99.7%** (1090/1093) | duplicate build |
| Darksiders `steam_api_ps5.prx` | **19.2%** (230/1199) | distinct library, 969 unique exports |
| BlazBlue `cri_lips_unity.prx` | **0.4%** (2/527) | distinct library, 525 unique exports |

The populations are **80 percentage points apart**, so 0.90 sits in an empty band; anything from ~0.5 to 0.95 partitions this corpus identically.

**Not silent.** A module linked despite a collision has its duplicates recorded in the existing `aliased_exports` (#1635) — the same mechanism that already reports this for by-name modules. First-definition-wins is unchanged. Dropping the module was only ever one way to avoid an unreported alias, and it was the expensive one.

## Verification

```
cmake --build ... -j6          BUILD_RC=0
ctest --no-tests=error -j6     CTEST_RC=0     100% tests passed out of 460
```

**Regression, and the useful part is *why* it is clean.** Both requested titles were run (`--volume 0`, offscreen, 400 frames):

| title | root `.prx` | modules | aliased exports | faults |
| GTA V `PPSA04263` | **0** | 3 | 0 | 0 |
| Sonic Frontiers `PPSA03831` | **0** | 2 | 0 | 0 |
| Tomb Raider I-III `PPSA16901` | **1** | 3 (was 2) | 0 | 0 |

**Corpus check, because the first draft of this section was scoped to two titles and that was not enough.** Measured across **all 60 local dumps**: exactly two ship a root-level module — `PPSA23806` (four, its own middleware) and **`PPSA16901` Tomb Raider I-III Remastered**, which ships `pros.sdk.Prospero-9.000.prx` and is **rung 3**. It is therefore the one title besides the subject whose link set this change alters, and it was run: 2 modules → 3, zero faults, presenting frames, so neither `linker.cpp:67`'s whole-link failure nor a `module_start` deadlock fires.

**That is a smoke test, not a rung-3 render check** — and note there is **no snapshot guard** for `PPSA16901` (checked: 17 entries in `snapshots.json`, none matching), so no automated guard exists to run. A human render check is the only stronger evidence available and has not been done.

Neither has a root-level module and neither had a collision-skip, so **both changed code paths are untaken on both titles**. That is a stronger claim than "it looks the same" — which matters here, because neither has a clean visual baseline to compare against (GTA V takes 122-175 device losses on a routed run by design, #3279). Sonic's one `SKIPPED` line is the pre-existing unimported-support-module drop.

**Evergate is the control for the rule's original purpose**: both fmod debug builds are still skipped.

## Test changes worth a look

`test_loader_synth_link`'s Arm 2 was **named** for the duplicate-build case but its fixture shared one of two exports — 50%, which is two different libraries, not two builds of one. So it asserted the old any-collision rule while claiming to test the rule's purpose, and could not distinguish the two cases. Split into 2a (a genuinely subsumed module, still skipped) and 2b (a distinct library, now linked with its own export surviving). **2b kills a revert to "skip on any collision"** and is the arm the old fixture could not provide.

## What this does NOT do

- **`sceAgcDriverInitResourceRegistration` is registered here and is NOT the crash fix.** I implemented it from a live-captured ABI, measured it, and the SIGSEGV was unchanged at the same rip — so that hypothesis is **falsified**. It is included because the guest calls it and prosper was answering with an unwritten success, which is a defect on its own terms.
- `U9ueyEhSkF4` stays unregistered: it resolves to nothing in the 3.20 reference, so there is no contract to implement against.
- Three defects remain open on #3497 from playing it: the intro FMV renders only its top-right triangle (**#3507**, hypothesis with a named discriminator), the FMV plays too fast, and the screen goes black after it.
- **Menu accuracy is not verified.** There is no oracle for this title; that is rung-5 work.

## Confidence / review notes

- `CONFIDENCE: HIGH` that the two link defects are real and that this fixes them — the chain is traced from the faulting instruction to a named unlinked module in the dump.
- `CONFIDENCE: MED` on 0.90 as a **universal** threshold. It is justified by four data points across five dumps. The band is wide, but that is an argument about this corpus, not a guarantee about every PS5 title. A title with, say, 85% overlap between two genuinely distinct libraries would be misjudged; none is known.
- The module-path policy change is the part that most warrants an independent read, per that file's own instruction.

### DIFF
--- a/prosper/src/hle/graphics/hle_agc.cpp
+++ b/prosper/src/hle/graphics/hle_agc.cpp
@@ -3045,6 +3045,39 @@ HLE(agc_driver_register_resource) {
     return 0;
 }
 
+// sceAgcDriverInitResourceRegistration(void* userMemory, size_t userMemoryBytes,
+//                                      uint32_t ownerCapacity, ...)
+// The one member of the registration family above that was never registered, and the one
+// QueryResourceRegistrationUserMemoryRequirements names as the consumer of the block it sizes.
+// Darksiders II (PPSA23806) calls it during startup; unregistered, it fell to the dispatcher's 0.
+//
+// The argument layout is LIVE-CAPTURED (PPSA23806, 2026-09-09) and cross-checked against the Query
+// call that sized the block in the SAME run, which is what pins it rather than a guess:
+//     Query(resources=100000, owners=96) -> required=0x61b500
+//     Init (a0=0x2080000000, a1=0x61b500, a2=0x60, a3=0xffffffffffffffff, ...)
+// a1 is exactly Query's answer, so a1 is the block SIZE in bytes; a2 is exactly Query's owner
+// count. The resource capacity is not passed -- it is implied by the size, and re-deriving it
+// reproduces the request: (0x61b500 - 0x100 - 96*0x20) / 0x40 = 100008 against a request of
+// 100000, whose unrounded requirement 0x61b4c0 aligns up to 0x61b500 under the same kAlignment.
+// a3 is an all-ones sentinel, not a count. a4 DIFFERED between two runs of the same route
+// (...68e3 vs ...69a3), so it is caller-indeterminate scratch and is deliberately not read -- the
+// same trap sceKernelMapFlexibleMemory's a4 carries.
+//
+// prosper keeps the registry on the HOST: the Query comment above states the guest block is never
+// consumed, and RegisterOwner/RegisterResource hand out host-side counter handles. So this call
+// validates its inputs and succeeds; it lays nothing out in guest memory, because there is no
+// guest-visible layout to honour.
+// CONFIDENCE: HIGH on a0/a1/a2 (captured, and cross-checked against Query in the same run).
+// CONFIDENCE: LOW on a3 and on the error code, which is inherited from the siblings above.
+HLE(agc_driver_init_resource_registration) {
+    if (!a0) return (uint64_t)(int64_t)(int32_t)0x80D19005u;
+    // A block too small to hold even the header the sizing call accounts for cannot be the block
+    // Query described. Nothing stricter is enforced: prosper does not lay the block out, so a
+    // tighter bound would reject valid callers on behalf of a layout it does not implement.
+    if (a1 < 0x100) return (uint64_t)(int64_t)(int32_t)0x80D19005u;
+    return 0;
+}
+
 // sceAgcCbDispatch (NID k3GhuSNmBLU) — compute dispatch.
 // The authoritative PS5 3.20 export name is sceAgcCbDispatch. Arguments a1-a3 are dispatch
 // dimensions and a4 is ShaderDispatchModifier. Modifier.USE_THREAD_DIMENSIONS (bit 5) defines the
@@ -3414,6 +3447,7 @@ void register_agc_hle() {
     RN("uJziRsODk1c", agc_driver_get_resource_registration_max_name_length);
     RN("X-Nm5KLREeg", agc_driver_register_owner);
     RN("W5z4eZrjEas", agc_driver_register_resource);
+    RN("F0Y42t-3e18", agc_driver_init_resource_registration);
     RN("V++UgBtQhn0", agc_get_data_packet_payload);          // data packet -> register-bank payload
     RN("n2fD4A+pb+g", agc_cb_set_sh_register_range_direct);  // SET_SH_REG range packet
     RN("UZbQjYAwwXM", agc_cb_set_sh_registers_direct);       // non-contiguous SET_SH_REG packets
--- a/prosper/src/hle/graphics/hle_agc.cpp
+++ b/prosper/src/hle/graphics/hle_agc.cpp
@@ -3057,8 +3057,16 @@ HLE(agc_driver_register_resource) {
 //     Init (a0=0x2080000000, a1=0x61b500, a2=0x60, a3=0xffffffffffffffff, ...)
 // a1 is exactly Query's answer, so a1 is the block SIZE in bytes; a2 is exactly Query's owner
 // count. The resource capacity is not passed -- it is implied by the size, and re-deriving it
-// reproduces the request: (0x61b500 - 0x100 - 96*0x20) / 0x40 = 100008 against a request of
-// 100000, whose unrounded requirement 0x61b4c0 aligns up to 0x61b500 under the same kAlignment.
+// returns the request exactly: (0x61b500 - 0x100 - 96*0x20) / 0x40 = 100000. (An earlier version of
+// this comment said 100008 and invented an intermediate 0x61b4c0; 0x100 + 100000*0x40 + 96*0x20 is
+// 0x61b500 already, and 0x40-aligned, so no rounding occurs. Corrected in review of #3508.)
+//
+// HOW MUCH THAT CROSS-CHECK IS WORTH, stated honestly because the first draft overstated it: the
+// `required` value came from prosper's OWN Query handler, so the guest passing it back confirms it
+// echoes what we told it -- not that a1 is a size according to Sony. What it does establish is that
+// a1 carries the byte count prosper handed out and a2 the owner count it was asked with, which is
+// what this handler needs in order to validate rather than to lay anything out. The independent
+// part of the evidence is the register capture itself, not the arithmetic.
 // a3 is an all-ones sentinel, not a count. a4 DIFFERED between two runs of the same route
 // (...68e3 vs ...69a3), so it is caller-indeterminate scratch and is deliberately not read -- the
 // same trap sceKernelMapFlexibleMemory's a4 carries.
--- a/prosper/src/host/image/boot_program.cpp
+++ b/prosper/src/host/image/boot_program.cpp
@@ -73,16 +73,37 @@ std::vector<std::string> discover_extra_plugin_modules(
     for (const auto& b : listed_basenames) listed.push_back(lower(b));
 
     std::error_code ec;
-    const std::string dir = resolve_host_path_case(dump_root + "/Media/Plugins");
-    if (!fs::is_directory(fs::path(dir), ec)) return found;
-    for (const auto& e : fs::directory_iterator(dir, fs::directory_options::skip_permission_denied, ec)) {
-        if (!e.is_regular_file(ec)) continue;
-        const std::string name = e.path().filename().string();
-        const std::string lname = lower(name);
-        if (lname.size() < 5 || lname.compare(lname.size() - 4, 4, ".prx") != 0) continue;
-        bool already = false;
-        for (const auto& l : listed) if (l == lname) { already = true; break; }
-        if (!already) found.push_back(e.path().string());
+    // Two locations, because titles disagree about where their own native modules live. Unity puts
+    // them under `Media/Plugins`; Darksiders II (PPSA23806) ships steam_api_ps5.prx,
+    // THQGDSCore_Prospero_Release.prx, thqno_api_ps5.prx and psnwrapperreleaseprospero.prx directly
+    // beside eboot.bin. Scanning only the first left that title's OWN middleware unlinked while
+    // prosper's return-0 stub answered for it, and the guest nulled an object pointer on the back of
+    // that zero and dereferenced it (#3497).
+    //
+    // The dump root is NOT a blanket widening: `module_path_policy` still refuses a Sony-named
+    // module here exactly as it does under `Media/`, so a dump cannot substitute a Sony library by
+    // dropping it beside eboot.bin. Only `.prx` is taken, so eboot.bin and loose data files are not
+    // candidates in the first place.
+    const std::string dirs[] = {
+        resolve_host_path_case(dump_root + "/Media/Plugins"),
+        resolve_host_path_case(dump_root),
+    };
+    for (const std::string& dir : dirs) {
+        if (!fs::is_directory(fs::path(dir), ec)) continue;
+        for (const auto& e : fs::directory_iterator(dir, fs::directory_options::skip_permission_denied, ec)) {
+            if (!e.is_regular_file(ec)) continue;
+            const std::string name = e.path().filename().string();
+            const std::string lname = lower(name);
+            if (lname.size() < 5 || lname.compare(lname.size() - 4, 4, ".prx") != 0) continue;
+            bool already = false;
+            for (const auto& l : listed) if (l == lname) { already = true; break; }
+            // A file present in BOTH locations must be offered once. Compare by basename against
+            // what has already been found, not only against `listed`.
+            if (!already)
+                for (const auto& f : found)
+                    if (lower(fs::path(f).filename().string()) == lname) { already = true; break; }
+            if (!already) found.push_back(e.path().string());
+        }
     }
     // directory_iterator order is filesystem-defined; sort so a boot is reproducible. DESCENDING by
     // lowercased basename, because the caller appends this block to a link list whose init functions
--- a/prosper/src/host/image/boot_program.hpp
+++ b/prosper/src/host/image/boot_program.hpp
@@ -199,8 +199,12 @@ std::string resolve_host_path_case(const std::string& want);
 // progress with no diagnostic at all. Tales of Graces f Remastered (PPSA19991) loses three of the
 // thirty singletons its boot state machine awaits that way and then renders an empty scene forever.
 //
-// `<dump_root>/Media/Plugins/*.prx` is the engine's own plugin directory, so its contents are exactly
-// the set the guest may ask for. Returns the paths of the entries whose basename does not
+// TWO locations are scanned (#3497): `<dump_root>/Media/Plugins/*.prx`, the engine's own plugin
+// directory, and `<dump_root>/*.prx`, where some titles put their native middleware instead
+// (Darksiders II ships four modules beside eboot.bin). The root is NOT a blanket widening --
+// `module_path_policy` refuses a platform module there exactly as it does under `Media/`, and only
+// `.prx` is a candidate, so eboot.bin and loose data files never are. Returns the paths of the
+// entries whose basename does not
 // case-insensitively match any of `listed_basenames`, ordered so that the caller (which appends them
 // to a link list whose init functions run in reverse order) initializes them in ascending
 // case-insensitive name order. A missing directory yields an empty vector; never throws.
--- a/prosper/src/host/image/module_path_policy.cpp
+++ b/prosper/src/host/image/module_path_policy.cpp
@@ -120,10 +120,18 @@ ModulePathDecision classify_module_path(const std::string& dump_root, const std:
         // this a dump could substitute any Sony library by dropping it beside eboot.bin -- an easier
         // bypass than the Media/Plugins one that rule already exists to close, and one that would
         // also sidestep the `fakelib/` rejection by a single move.
-        if (lower(parts[0]).rfind(kSonyLibraryPrefix, 0) == 0) {
+        const std::string root_name = lower(parts[0]);
+        auto refused_platform_name = [&]() -> const char* {
+            if (root_name.rfind(kSonyLibraryPrefix, 0) == 0) return kSonyLibraryPrefix;
+            for (const char* p : kRootRefusedPlatformPrefixes)
+                if (root_name.rfind(p, 0) == 0) return p;
+            return nullptr;
+        };
+        if (const char* matched = refused_platform_name()) {
             d.verdict = ModulePathVerdict::SonyLibraryOutsideSceModule;
-            d.reason = "a Sony-named library in the dump root, which prosper auto-scans; Sony "
-                       "libraries are linked only from the dump's own `sce_module/` directory";
+            d.reason = std::string("a platform module in the dump root (matches `") + matched +
+                       "`), which prosper auto-scans; platform libraries are linked only from the "
+                       "dump's own `sce_module/` directory";
             return d;
         }
         // Only module files. `eboot.bin` is handled above; anything else in the root is data.
--- a/prosper/src/host/image/module_path_policy.cpp
+++ b/prosper/src/host/image/module_path_policy.cpp
@@ -110,8 +110,32 @@ ModulePathDecision classify_module_path(const std::string& dump_root, const std:
                 return d;
             }
         }
+        // A title's OWN native middleware may sit directly in the dump root rather than under
+        // `Media/` -- Darksiders II (PPSA23806) ships four such modules there (#3497). Refusing them
+        // left the guest's real implementations unlinked with prosper's return-0 stub answering in
+        // their place.
+        //
+        // The Sony rule is applied here EXACTLY as it is to the Media directories, and for a
+        // stronger version of the same reason: the root is now auto-scanned wholesale, so without
+        // this a dump could substitute any Sony library by dropping it beside eboot.bin -- an easier
+        // bypass than the Media/Plugins one that rule already exists to close, and one that would
+        // also sidestep the `fakelib/` rejection by a single move.
+        if (lower(parts[0]).rfind(kSonyLibraryPrefix, 0) == 0) {
+            d.verdict = ModulePathVerdict::SonyLibraryOutsideSceModule;
+            d.reason = "a Sony-named library in the dump root, which prosper auto-scans; Sony "
+                       "libraries are linked only from the dump's own `sce_module/` directory";
+            return d;
+        }
+        // Only module files. `eboot.bin` is handled above; anything else in the root is data.
+        const std::string ln = lower(parts[0]);
+        const bool is_module = (ln.size() > 4 && ln.compare(ln.size() - 4, 4, ".prx") == 0) ||
+                               (ln.size() > 5 && ln.compare(ln.size() - 5, 5, ".sprx") == 0);
+        if (is_module) {
+            d.verdict = ModulePathVerdict::Permitted;
+            return d;
+        }
         d.verdict = ModulePathVerdict::DirectoryNotPermitted;
-        d.reason = "in the dump root, which holds no linkable module other than eboot.bin";
+        d.reason = "in the dump root and not a linkable module";
         return d;
     }
 
--- a/prosper/src/host/image/module_path_policy.hpp
+++ b/prosper/src/host/image/module_path_policy.hpp
@@ -1,9 +1,13 @@
 #pragma once
 // module_path_policy — which files inside a game dump prosper is willing to link as a guest module.
 //
-// REJECT BY DEFAULT. `kPermittedModuleDirs` below is the COMPLETE set of dump-relative locations a
-// module may be linked from; anything else is refused and reported, whatever the file is called and
-// whatever produced it.
+// REJECT BY DEFAULT. A module may be linked from `kPermittedModuleDirs` below, or from the dump ROOT
+// subject to the platform-name and extension rules further down (#3497). Together those are the
+// COMPLETE set of dump-relative locations; anything else is refused and reported, whatever the file
+// is called and whatever produced it.
+//
+// The root is the loosest of them and the one to read carefully: it is auto-scanned wholesale, and
+// it is not a directory a title owns by convention -- `eboot.bin` and `sce_module/` live there too.
 //
 // Why this is an enforced policy and not merely a property of how boot_link_inputs happens to be
 // written today: some dumps in circulation ship a `fakelib/` directory of replacement Sony
@@ -43,7 +47,10 @@ inline constexpr const char* kPermittedModuleDirs[] = {
     "sce_module",
 };
 
-// Files permitted directly in the dump root.
+// Files permitted directly in the dump root REGARDLESS of extension. Module files in the root are
+// handled separately (see the platform-prefix rules below): a title's own native middleware may sit
+// beside eboot.bin -- Darksiders II ships four such modules there -- so the root is not the
+// eboot-only location this list alone would suggest.
 inline constexpr const char* kPermittedRootFiles[] = {
     "eboot.bin",
 };
@@ -57,16 +64,48 @@ inline constexpr const char* kPermittedRootFiles[] = {
 //
 // CONFIDENCE: HIGH for the corpus — measured across all 50 local dumps, zero ship a `libSce*` under
 // `Media/Plugins` or `Media/Modules`, while three ship exactly such files under `fakelib/`. MED as a
-// universal claim about PS5 titles. If a real title ever does need one, the refusal is loud and
+// universal claim about PS5 titles.
+//
+// That measurement covers the two Media directories and was NOT re-run for the dump root when the
+// root was opened (#3497). What the root measurement does say: of 60 local dumps, exactly two ship a
+// root-level module at all — PPSA23806 (four, its own middleware) and PPSA16901 (one) — and neither
+// is a platform name. So the root rule is unexercised as a refusal across the corpus, which is a
+// weaker statement than the Media one and is why the platform-prefix list below exists rather than
+// relying on `libSce*` alone. If a real title ever does need one, the refusal is loud and
 // names the file, and widening the rule is a reviewed source change — which is the right direction
 // for the error to point, because the alternative failure is silent.
 inline constexpr const char* kSonyLibraryPrefix = "libsce";  // compared lowercased
+
+// The dump ROOT is auto-scanned wholesale (#3497), and unlike `Media/*` it is not a directory a
+// title owns by convention -- `eboot.bin` and `sce_module/` live there too. `libSce*` alone is NOT
+// sufficient cover for it: measured against the 3.20 reference, 21 of 275 system modules do not
+// carry that prefix, and `linker.cpp`'s "a cross-module export beats a stub slot" rule means a root
+// module shadowing one of those would silently displace prosper's own HLE.
+//
+// So the root additionally refuses the PLATFORM modules prosper itself provides or implements.
+// Deliberately NOT the whole non-`libSce` list: most of it is third-party (`libcurl`, `libpng16`,
+// `libfreetype`, `libicu`, `libharfbuzz`) and a title may legitimately ship its own build of those
+// -- refusing them would break exactly the case this widening exists to serve.
+//
+// HONEST LIMIT, because the previous version of this comment overstated it and was corrected in
+// review: this is a name heuristic over the platform surface, not a proof about every Sony-authored
+// module. What actually makes the root safe is the pair -- this list plus the extension gate plus
+// the fact that `fakelib/` and every other directory remain refused outright. A dump that ships a
+// platform replacement under a name matched by neither rule would still be linked, and the defence
+// against that is `dump_hygiene.py` and the link-time HLE ownership, not this function.
+inline constexpr const char* kRootRefusedPlatformPrefixes[] = {
+    "libkernel",   // libkernel, libkernel_sys, libkernel_web
+    "libc.",       // libc.prx exactly -- NOT libcairo/libcurl, which a title may ship
+    "libc_",
+    "libmdbg",     // libmdbg_syscore
+    "gaikai",      // gaikai-player
+};
 
 enum class ModulePathVerdict {
     Permitted,
     OutsideDumpRoot,             // not under the dump root at all, or escapes it via `..`
     DirectoryNotPermitted,       // inside the dump, in a directory prosper never links from
-    SonyLibraryOutsideSceModule, // a libSce* module in an auto-linked Media directory
+    SonyLibraryOutsideSceModule, // a platform module in an auto-linked location (Media/* or the root)
 };
 
 struct ModulePathDecision {
--- a/prosper/src/host/image/module_path_policy.hpp
+++ b/prosper/src/host/image/module_path_policy.hpp
@@ -78,9 +78,16 @@ inline constexpr const char* kSonyLibraryPrefix = "libsce";  // compared lowerca
 
 // The dump ROOT is auto-scanned wholesale (#3497), and unlike `Media/*` it is not a directory a
 // title owns by convention -- `eboot.bin` and `sce_module/` live there too. `libSce*` alone is NOT
-// sufficient cover for it: measured against the 3.20 reference, 21 of 275 system modules do not
-// carry that prefix, and `linker.cpp`'s "a cross-module export beats a stub slot" rule means a root
-// module shadowing one of those would silently displace prosper's own HLE.
+// sufficient cover for it: at least 23 system modules do not carry that prefix -- 21 of the 275 in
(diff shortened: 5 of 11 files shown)


mattias800 (commented):
# Independent review — head `6c9634f6`

**REJECTED**, on two blocking items (B1, B2) plus two corrections to the anti-circumvention guard's own
stated contract (B3, B4). The diagnosis is excellent and I believe the fix is right; what is missing is
one cross-title run and the end-to-end arm this repo explicitly requires for exactly this change. None of
it is a redesign.

I did not build or run anything. Everything below is read from head `6c9634f6`, from the local dump
corpus, and from the PS5 3.20 reference tree.

---

## B1 (blocking) — the root scan is not a no-op across the corpus, and the other affected title was never named

The regression argument covers GTA V and Sonic Frontiers, and it is sound for those two: I re-ran the
census and neither has a root-level module, so both changed paths are genuinely untaken. But the
corpus-wide statement was never made, and it does not hold.

Of the 60 local dumps with an `eboot.bin`, **exactly two** ship a module directly in the root:

| dump | root modules |
| --- | --- |
| `<DUMP_ROOT>/PPSA23806-app0` | the four this PR is about |
| `<DUMP_ROOT>/PPSA16901-app0` | `pros.sdk.Prospero-9.000.prx` — 6.3 MB, real SELF (`54 14 f5 ee`) |

`PPSA16901` is *Tomb Raider I-III Remastered*, **rung 3** in `COMPATIBILITY.md`, with a checked-in
Croft Manor gameplay screenshot and a route under `prosper/scripts/tomb-raider-PPSA16901/`. It has no
`Media/` directory at all, so today auto-link contributes **nothing** for that title. After this PR it
contributes one new module: `pros.sdk.Prospero-9.000.prx` is absent from the fixed preload list, so it is
discovered (`boot_program.cpp:87-107`), permitted (`module_path_policy.cpp:129-136`), present on disk, and
linked with `skip_on_export_collision` — and its `module_start` / `init_array` run **before the eboot's
entry**.

Two failure modes, both already precedented in this repository:

1. `linker.cpp:67` — `if (!mo) return fail("load " + in.path + ": " + e);`. A present-but-unparseable
   module fails the **whole link**, so `boot_program` returns "link failed" and the title stops booting
   outright. Rung 3 → rung 0, with no snapshot guard on that title to catch it.
2. A `module_start` that deadlocks the boot. This is not speculation — it is why `only_if_imported`
   exists at all; the comment at `boot_program.cpp:~240` records `libSceNpCppWebApi.prx` doing exactly
   that on `PPSA03130`.

**What I need before merge:** one boot of `PPSA16901` on this branch, with the module/import counts
before and after and a frame, posted in the PR. If it regresses, the narrow lever is already in the file:
mark root-discovered modules `only_if_imported`, or scope the root scan to what the eboot actually
imports. I am not asking you to pre-emptively narrow it — only to look.

Same request, lower stakes, for `PPSA29714` (BlazBlue, `cri_lips_unity.prx`): the PR asserts the module is
"unblocked" but the verification table does not include a run of it. A 527-export module whose init_array
now runs is an assertion about behaviour, not just about the link list. (Evergate as the negative control
is good and I am satisfied by it — I confirmed `PPSA01885` is the only dump in the corpus shipping the
`libfmodL`/`libfmodstudioL` pair, so the rule's original purpose really is exercised by exactly that title.)

## B2 (blocking) — the discovery widening did not extend the guard test, which this repo requires by name

`prosper/src/host/image/AGENTS.md:42-46`:

> **Adding a module to the discovery path is a security-relevant change, not just a convenience.**
> `discover_extra_plugin_modules()` links whatever a dump ships in one directory (#1609). Widening
> that — recursing, accepting another extension, **scanning another directory** — changes which
> third-party code prosper will execute on behalf of the guest. `module_path_policy` is the guard, and
> `tests/host/image/test_module_path_policy.cpp` is what makes it real; **extend both together**.

This PR is the "scanning another directory" case. It extends the *unit* arms
(`test_module_path_policy.cpp:100-122`) — which are good, and I confirm a root `libSce*` gets the
`SonyLibraryOutsideSceModule` verdict rather than a generic one, in `.prx`, `.sprx` and upper case. But
the **end-to-end arm** (`:164-206`) was not touched: its fixture writes only `Media/Plugins/*` and
`fakelib/*`, and there is no root-level module in it at all. That file's own header (`:8-16`) says the two
layers "are NOT the same strength" and that the end-to-end arm is the one deliberately built to fail
without the guard. So right now **nothing anywhere exercises the discovery→policy chain for the root**,
and the classifier is not what loads modules.

The missing arm is four lines in the fixture that already exists, and it is a genuine regression arm —
delete the new `module_path_policy.cpp:123-128` block and it reddens:

```cpp
write_file(root / "libSceAppContent.prx", "not-a-real-image");   // must NOT be linked
write_file(root / "RootPlugin.prx",       "not-a-real-image");   // must be linked
...
CHECK(!linked(in, "libSceAppContent"), "a Sony library beside eboot.bin is NOT linked");
CHECK(linked(in, "RootPlugin.prx"),    "...while a title's own root module still IS");
```

The second line is the positive control that stops the first from passing because the root scan silently
did nothing.

Also uncovered, and in the file that exists for it:

- `tests/misc/test_plugin_autolink.cpp` has **no arm for the new root branch**, and its header (`:12-13`)
  still states the superseded contract: *"everything in `<dump>/Media/Plugins/*.prx` that the caller did
  not already name … and nothing else is"*. The new cross-directory basename dedup
  (`boot_program.cpp:100-104`) is untested too — it looks correct to me, but it is new logic in the one
  function that has a dedicated test file.
- `boot_program.hpp:202-206` still documents the single-directory contract.

## B3 (blocking) — the widening's stated contract is false as written

`boot_program.cpp:83-86`, and the same sentence in the commit message and the PR body:

> so a dump cannot substitute a Sony library by dropping it beside eboot.bin

It can, provided the file is not named `libSce*`. The rule is a **name prefix**, and Sony's own library
set is not covered by it. Measured on the local reference tree: of 267 system modules under
`<DUMP_ROOT>/sprx`, **11 do not start with `libSce`** — `libkernel.sprx`, `libkernel_sys.sprx`,
`libkernel_web.sprx`, `libwvoec.sprx`, `ulobjmgr.sprx` and the PlayReady files — and the PS5 3.20
reference adds `libcurl`, `libfreetype`, `libicu`, `libpng16`, `libmonosgen-2.0`. `linker.cpp:151-156`
makes the consequence concrete — *"Cross-module export beats a stub slot"* — so any linked module's
exports shadow prosper's own HLE for those NIDs, which is precisely the outcome
`module_path_policy.hpp:14-17` says must not happen.

I want to be fair about the scope: the *known* circumvention vector stays closed. `dlc_emu` and
`ampr_emu` produce `libSce*` names, `<DUMP_ROOT>/PPSA23806-app0/fakelib/libSceAmpr.sprx` is still refused
(subdirectory, not on the allowlist, and the new scan is depth-1 at `boot_program.cpp:93`), and the same
prefix hole has existed for `Media/Plugins` since #1609. So this is not a new class of hole. But the
sentence as written is a stronger claim than the code makes, in a comment on the anti-circumvention
guard, and the accurate version — "cannot substitute a Sony-**named** library" — costs one word.

Directly connected, and the part I would push back on hardest: **this PR deletes the one arm that pinned a
Sony library whose name does not match the prefix**, and replaces it with nothing —

```
-    refuse(R, "/dumps/PPSA00000-app0/libc.prx", ModulePathVerdict::DirectoryNotPermitted,
-           "the dump root holds no linkable module but eboot.bin");
```

`libc.prx` is a real Sony module: `boot_program.cpp:~246` links `sce_module/libc.prx` by name. Under the
new rule a root `libc.prx` is **Permitted by the policy**, and is only not loaded because
`discover_extra_plugin_modules` dedups against the fixed list's basenames (`boot_program.cpp:98-99`) — an
incidental property of the discovery code. That is exactly what `module_path_policy.hpp:19-23` was written
to forbid relying on: *"Before this policy existed prosper already did not load those files, but only as a
side effect of three unrelated details of `discover_extra_plugin_modules()` … A property that no test
asserts is not a property."* Either broaden the refusal (a small name set, or the 3.20 export-name
database the project already has next to the repo) or keep an arm pinning today's answer, so the next
widening has to look at it. Silently dropping the arm is the one option that leaves no trace.

## B4 (blocking) — the guard file's own contract text is now false and was not updated

`module_path_policy.hpp` is the document the next reviewer reads to learn what the policy *is*. After this
PR it describes a policy that no longer exists:

- `:4-6` — "`kPermittedModuleDirs` below is the **COMPLETE** set of dump-relative locations a module may
  be linked from". The root is now a fourth location and is not in that array.
- `:46-49` — "Files permitted directly in the dump root" over a list containing only `eboot.bin`.
- `:51-56` — "A Sony-named module (`libSce*`) may be linked ONLY from `sce_module/`, never from **the two
  Media directories**" — now three locations.
- `:58-62` — the CONFIDENCE paragraph rests on "measured across all 50 local dumps, zero ship a `libSce*`
  under `Media/Plugins` or `Media/Modules`". That measurement was never re-run for the root, which is now
  the auto-scanned location the rule has to defend. I ran it: **0 of 60 dumps ship a root-level
  `libSce*`**, and 2 ship a root module at all. That belongs in the header — it is the evidence the new
  rule stands on.
- `:69` — the enum comment still reads "a `libSce*` module in an auto-linked **Media** directory".

Per the charter, a stale statement here carries the file's authority; the widening PR is where it gets
fixed, not a follow-up.

Non-blocking, one line for the same header while you are in it: classification is **lexical**, so it
judges the path string and never the file. A rename or a symlink defeats a name-prefix rule anywhere it
applies (`is_regular_file` follows symlinks, so a root `foo.prx` → `fakelib/libSceAmpr.sprx` would be
discovered and linked). That is true of `Media/Plugins` today too and I am not asking you to fix it — but
the comment currently reads as though the rule is a barrier, and it is a signpost.

---

## Should-fix (not gating on their own, but cheap and all in the changed lines)

**S1 — `test_loader_synth_link.cpp:207` asserts a property its fixture cannot produce.**
The arm checks *"a skipped module's unsatisfied import creates no stub slot"* against `subsumed_spec`
(`:110`), which sets only `exports`; `SynthModuleSpec::imports` defaults empty
(`tests/fixtures/synth_prx.h:86`). The old Arm 2 got its teeth from `logging_spec`'s `kMissing` import
(`:97-98`). As written the assertion is true whether or not the module is skipped. One line restores it:
`subsumed_spec.imports = { kMissing };`.

Everything else in 2a/2b I checked and it holds: `release_spec` (`:96`) and `subsumed_spec` (`:110`) really
are the same export set, so 2a's fixture is a genuine 2/2 = 100% subsumption; and 2b does kill a revert to
"skip on any collision" — `mods.size() == 3` and `export_of(kOnlyB)` both redden. The split was worth
doing and the old arm really was misnamed.

**S2 — nothing pins 0.90.** The fixtures are 100% and 50%, so *any* threshold in (0.5, 1.0] keeps the
suite green. The threshold is the one new judgement in this PR (you label it `CONFIDENCE: MED` yourself,
correctly), and it is the one thing untested. The corpus's nearest true duplicate is 99.7%, i.e. a fixture
that straddles 0.90 needs ~10 symbols.

The cheap remedy is structural: `measure_export_subsumption` is defined at namespace scope in
`linker.cpp:44` with external linkage and **no declaration in any header** — unlike its siblings
`module_export_nids` and `find_export_collision`, both declared in `linker.hpp` and both unit-tested
directly on synthetic `Module` structs in `test_plugin_autolink.cpp`. Declare it, then a 9/10-skipped +
8/10-linked pair is ~10 lines with no PRX fixture at all.

**S3 — `linker.hpp:116-117`.** The `find_export_collision` doc comment is now orphaned: the
`ExportSubsumption` block was inserted between it and `struct ExportCollision` / `find_export_collision`
(`:131-133`), so two unrelated doc paragraphs stack and the first now reads as documentation of the
second. Move the new struct above the comment or below the declaration.

**S4 — `hle_agc.cpp:3060-3061`: the arithmetic is wrong and the cross-check is circular.**

- Wrong: `(0x61b500 - 0x100 - 96*0x20) / 0x40` = `0x61A800 / 0x40` = **100000**, exactly the request —
  not 100008. And `0x61B500` is already `0x40`-aligned, so the "unrounded requirement `0x61b4c0`" does not
  exist. Re-running prosper's own formula (`:3013-3016`) for (100000, 96): `0x100 + 100000*0x40 + 96*0x20`
  = `0x61B500` exactly. The real result is *cleaner* than the one in the comment, which is a shame, because
  the wrong numbers will be quoted.
- Circular: `required = 0x61b500` in that capture is **prosper's own Query output**, computed from
  prosper's own estimated record sizes (`kResourceBytes` / `kOwnerBytes`, flagged "MED on opaque record
  sizes" at `:3005`). So "re-deriving it reproduces the request" reproduces our arithmetic, not hardware's.
  What the capture genuinely establishes — and it is enough for the code as written — is that **the guest
  passes Query's returned value straight into Init's `a1`**. Say that instead; it is the real evidence and
  it is strong.

Please fix both in the comment, the commit message and the PR body together.

Minor, same function: the new `a1 < 0x100` rejection turns a call that previously always succeeded
(unregistered → dispatcher's `0`) into a failure for any title whose argument layout differs from the one
captured title, and neither return path has a test arm. The bound is defensible (our Query can never
return < `0x100`) and `CONFIDENCE: LOW` on the error code is already stated, so I am not blocking — but if
a title regresses here, this is the first place to look. The memory of "registering a NID can be worse
than not registering it" applies to the *error* return, not the success one.

---

## Verified and fine — recorded so nobody re-derives it

- **`fakelib/libSceAmpr.sprx` in `<DUMP_ROOT>/PPSA23806-app0` is still refused.** Subdirectory, not on the
  allowlist (`module_path_policy.cpp:142-165`), `describe_rejection` names it as the entitlement case, and
  the new scan is a non-recursive `directory_iterator` (`boot_program.cpp:93`). Confirmed the file is
  actually present in that dump.
- **The Sony test on the root path is applied identically to the Media path's.** `:123` uses
  `lower(parts[0])`, `:153` uses `lower(parts.back())`; for a one-component relative path these are the
  same string. The root branch is in fact *stricter* than Media in one respect — Media permits any file
  regardless of extension, the root requires `.prx`/`.sprx`.
- `..` and `.` are still refused before the root branch is reached (`:98-104`), and the root prefix match
  is still on a component boundary (`:82-83`).
- **The policy is still the chokepoint.** `enforce_module_path_policy` runs at `boot_program.cpp:272-274`,
  after discovery and before the first `Module::load` (the earliest parse is downstream in
  `drop_unimported_support_modules`).
- **The aliasing really is reported** — I checked the code, not the claim. `linker.cpp:125-137` records
  every duplicate NID with winner path, loser path and both addresses, and `boot_program.cpp` prints it
  bounded-with-count. First-definition-wins is unchanged. Your claim holds.
- `subsumed()` with `exported == 0` is unreachable (the call is guarded by a non-empty
  `find_export_collision`), and returns `false` there anyway, which is the safe direction.
- `F0Y42t-3e18` = `sceAgcDriverInitResourceRegistration`, confirmed in `libSceAgcDriver` of the 3.20
  reference; `U9ueyEhSkF4` is confirmed absent from the whole reference, so leaving it unregistered is
  right.
- The "**`sceAgcDriverInitResourceRegistration` is NOT the crash fix**" section is honest and the
  falsification is properly recorded. No overclaim there. The only overclaims I found are the two named
  above (B3's "cannot substitute a Sony library", S4's `100008`), plus the scope gap in B1.

One reminder rather than a finding: the rung-0 → title-screen result is the part the owner actually reads,
and it only surfaces through the screenshot + `BLOG.md` entry you deferred. This PR is compliant as it
stands (no screenshot in the diff), but the follow-up is what makes the work visible.

---

**REJECTED** — B1 (run `PPSA16901`, and ideally `PPSA29714`), B2 (extend the end-to-end guard arm, per
`src/host/image/AGENTS.md:42-46`), B3 (correct the "cannot substitute a Sony library" claim and replace
the deleted non-`libSce` root arm), B4 (update `module_path_policy.hpp`'s now-false contract text). S1-S4
are cheap and I would like them in the same push. Re-request review on the new head and I will re-read the
changed lines only.

mattias800 (commented):
# Re-review — head `87f8cfcf9`

**REJECTED**, and I want to be clear about the scale: **all four blocking items from the first review are
cleared in substance.** What is left is two small changes, both in the guard, both named exactly below.
One of them is the thing you asked me to check, and you were right to ask — it does not hold.

Read against `87f8cfcf9` only; I built and ran nothing.

---

## Your three questions, answered first

### Q1 — attack `kRootRefusedPlatformPrefixes`. What does it miss?

I took the union of two sources rather than one, and that turns out to matter.

- **The 3.20 reference** (your source): 21 of 275 modules are non-`libSce`. Your count is right.
- **The on-disk firmware set** (`<DUMP_ROOT>/sprx`, 267 files): 11 non-`libSce`, and **two of them are not in
  the 3.20 reference at all** — `libwvoec.sprx` (Widevine OEM crypto) and `ulobjmgr.sprx`. So the platform
  surface is at least 23, not 21, and the census source you measured against is not the same set as the
  firmware actually on disk. Worth one clause in the comment; a future reader will otherwise re-derive
  "21" from the same incomplete source.

Neither escapee matters, and this is the part that settles the question: `grep -ril` over `prosper/src`
finds **no HLE for `ulobjmgr` or `wvoec`**, nor for `libNativeExtensions`, `libmonosgen-2.0`,
`libReactNative.*`, `mscorlib.dll` or `WebAppBundle`. A module can only shadow prosper where prosper
registers the NID (`linker.cpp`'s "cross-module export beats a stub slot"), so a platform module we do not
implement is inert even if linked.

Apply that test to the list itself and it comes out clean: `src/hle/` contains exactly `kernel/` and
`libc/` as the non-`libSce` surfaces prosper implements, and those are precisely `libkernel*` and `libc.`.
So **the list covers 100% of the shadowing-relevant surface**. `libmdbg` and `gaikai` are pure precaution
(the only hits in the whole tree are the header itself) — cheap and harmless, keep them.

`libc.` / `libc_` is the right discriminator. Checked against both sources: the only names beginning
`libc` are `libc.prx` (platform), `libcairo`, `libcurl` (third-party) and `libSceLibcInternal`
(prefix-caught). `libc.` matches `libc.prx` and `libc.sprx` and nothing else; `libc_` costs nothing and
covers a stub-style spelling. A bare `libc` would swallow both third-party names, which is exactly what
your test arm at `:137-139` now kills. I could not construct a real platform name that slips past.

One optional addition, which is the one class the list does not reach: **`dlc_emu` / `ampr_emu`**. Those
are the actual names of the third-party tooling this whole policy exists to refuse (`dump_hygiene.py`
already knows `dlc_emu.ini` and `ampr_emu.index`), they cannot collide with anything legitimate, and they
are the one case where the risk is circumvention rather than shadowing. Today those tools emit `libSce*`
module names so the existing rule catches them — this would just remove the dependence on that staying
true. Your call; I am not gating on it.

### Q2 — are the permits right, or over-permitted?

Right. `libcurl.prx` and `libcairo.prx` in the root are correctly permitted: both are third-party
libraries a title legitimately ships its own build of, prosper implements neither, and refusing them
would break the exact case the widening serves. The reasoning in the comment is sound and I would not
change it. (`libfreetype` / `libpng16` / `libharfbuzz` / `libicu` are the same class; two arms are
enough.)

### Q3 — would the end-to-end arm fail if the platform rule were removed?

**Two of the three would. The `libc.prx` one would not, and it is the only one covering the new rule.**

`test_module_path_policy.cpp:222`:

```cpp
CHECK(!linked(in, "/libc.prx") || linked(in, "sce_module"),
      "a platform module in the root is NOT linked from there");
```

The fixed preload list contains `{ d + "/sce_module/libc.prx", BOOT_LIBC }` (`boot_program.cpp:243`), and
`listed` is built from the **basenames** of every fixed entry (`:253-254`), so `listed` contains
`libc.prx`. `discover_extra_plugin_modules` skips any candidate whose lowercased basename matches a listed
name (`:98-99`). So a root `libc.prx` **is never discovered in the first place — with or without
`kRootRefusedPlatformPrefixes`.** Delete the whole platform-prefix block and this assertion still passes.

That is the same incidental-dedup property finding 3 was about, now laundering an assertion instead of a
permit. And the arm's own comment (`:214-217`) claims otherwise — *"Kills: opening the root without
applying the platform-name rule to it, which leaves both refusals below linked"* — which is false for one
of the two.

Scoring the other two, since they are the ones doing the work:

- `:220` `!linked(in, "libSceAppContent.prx")` — **has teeth.** Not in the fixed list, so discovery finds
  it and only `kSonyLibraryPrefix` stops it. Remove the root Sony rule and this reddens. This is a
  genuine regression arm and it is what clears blocking item 2. (It also does not collide with the
  `fakelib/libSceAppContent.sprx` fixture: the needle carries `.prx`.)
- `:218` `linked(in, "RootMiddleware.prx")` — **has teeth**, and is the right positive control: without it,
  a root scan that silently did nothing would satisfy both refusals.

So the chain is covered end to end for `libSce*`, and **not at all for the new prefix list.**

**The fix is two words.** `libkernel.prx` is not in the fixed preload list, so discovery does find it and
only `kRootRefusedPlatformPrefixes` refuses it:

```cpp
write_file(root / "libkernel.prx", "not-a-real-image");        // :204
...
CHECK(!linked(in, "libkernel.prx"),                            // :222
      "a platform module in the root is NOT linked, even though it is not named libSce*");
```

That also lets the `|| linked(in, "sce_module")` escape hatch go. It is inert today (the fixture ships no
`sce_module/`, so both entries are dropped as absent), but it is a disjunct that makes the assertion
vacuously true the moment someone adds an `sce_module` fixture — an assertion with a built-in off switch.

### And the one you asked me to call: Tomb Raider

**Not blocking.** But correct the premise first: **`PPSA16901` has no snapshot guard.**
`tools/snapshot/snapshots.json` holds 17 entries and none matches `tomb` or `16901`. So "it has a guard and
I did not run it" is not the situation — there is nothing to run, which makes your smoke test the only
evidence that exists rather than a substitute for a cheaper one.

On its merits the smoke test is sufficient for what I raised. I named two mechanisms; the run excludes
both directly. "3 modules" is the corroborating detail: that dump's `sce_module/` holds exactly
`libc.prx`, so 3 = eboot + libc + `pros.sdk.Prospero-9.000.prx` — the new module really is linked, the link
did not fail (`linker.cpp:67`), and frames present, so its `init_array` did not deadlock. A rendering
regression behind that has no proposed mechanism, and the charter is explicit that snapshot state does not
gate an individual PR.

One ask, and it is a record-keeping one: **put that result in the PR body.** Module counts, 0 faults,
frames presented, rendering *not* verified, and no guard exists for the title. A result that lives only in
a message to a coordinator is not in the project record, and this is the one title a future regression
here would land on.

---

## Blocking (both small)

**R1 — the end-to-end arm for the new prefix list has no teeth.** Q3 above. Change the fixture to
`libkernel.prx`, drop the `sce_module` disjunct, and fix the `Kills:` comment at `:214-217` so it describes
what the arm actually kills. This is the second time in this PR that an incidental property of
`discover_extra_plugin_modules` has stood in for a guarantee, which is why I am not waving it through:
the newest and least-exercised rule in the file currently has zero coverage at the layer the header itself
says is the only one that counts.

**R2 — `module_path_policy.hpp:95` cites a backstop that does not cover the case.**

> A dump that ships a platform replacement under a name matched by neither rule would still be linked, and
> the defence against that is `dump_hygiene.py` and the link-time HLE ownership, not this function.

`dump_hygiene.py`'s classifier flags a module only when `lname.startswith(SONY_LIBRARY_PREFIX)` — i.e.
`libsce` — outside `sce_module`. A name matched by neither rule is, by construction, not `libSce*`, so
**`dump_hygiene.py` does not report it either.** The cited defence is empty for exactly the case the
sentence is about. The paragraph's main point (a name heuristic, not a proof) is honest and I want it kept
— it is the second half that overreaches, in the same direction as the sentence I flagged last round.

There is a real consequence behind the wording, which is why this is R2 and not a nit: **the loader and
the tool have now diverged.** The policy refuses a root `libkernel.prx`; `dump_hygiene.py` classifies it
as unremarkable and `--check` stays green on it. `tests/tools/test_dump_hygiene.py` exists to stop exactly
that drift but only mirrors `kPermittedModuleDirs`, so it cannot see this one. Either mirror
`kRootRefusedPlatformPrefixes` into the tool (it already has a root-specific message and a prefix
constant), or drop the claim and say plainly that nothing outside these two rules is defended. Either is
fine; asserting a backstop that is not there is not.

## Should-fix (not gating)

**S1 — the Arm 2a replacement is also toothless, though the arm as a whole is fine.**
`test_loader_synth_link.cpp:235-236` now asserts `export_of(p, kOnlyA) == kBase1 + kExp1`. `subsumed_spec`
and `release_spec` carry the same export list in the same order, so `subsumed`'s `kOnlyA` sits at
`kBase2 + kExp1` and loses to `release` under first-wins **whether it is skipped or linked**. A fully
subsumed module can never win an export by definition, so "contributed no exports" is unobservable through
`p.exports` — the observable differences are `mods.size()`, `skipped_modules` and `aliased_exports`, and
you already assert all three, which is why the arm still has teeth and this is not a regression.

The version that does discriminate is the one the old fixture had by accident: give it an unsatisfied
import and restore the original assertion.

```cpp
SynthModuleSpec subsumed_spec; subsumed_spec.exports = { kShared, kOnlyA };
                               subsumed_spec.imports = { kMissing };     // nobody exports kMissing
...
CHECK(p.slots.empty(), "duplicate build: a skipped module's unsatisfied import creates no stub slot");
```

Skipped → no slot; linked → `kMissing` creates one and it reddens. (`main`'s `kShared` import is satisfied
by `release`, so nothing else contributes a slot.)

**S2 — the orphaned comment is still there; precise location as requested.** At head `87f8cfcf9`,
`linker.hpp:117-118` is the doc comment for `find_export_collision` — *"First NID of `m` already present in
`claimed` … scanned in file order."* It is immediately followed by `:119-121`, the `ExportSubsumption`
comment, and `struct ExportSubsumption` at `:122`. What `:117-118` describes is `struct ExportCollision` /
`find_export_collision` at `:133-135`. So two unrelated doc paragraphs now stack and the first reads as
documentation of the second. Move the `ExportSubsumption` block above `:117` or below `:135`.

(`:121` also points at `measure_export_subsumption()`, which is still defined at namespace scope in
`linker.cpp` with external linkage and no declaration in any header. Not worth changing now — Arm 0
constructs `ExportSubsumption` directly, which is a better answer to the threshold question than exporting
the function would have been.)

---

## Cleared from the first review

- **Finding 1 (corpus).** Your census matches mine exactly — 60 dumps, two with root modules, `PPSA23806`
  and `PPSA16901`. Smoke test accepted; see the call above.
- **Finding 2 (end-to-end arm).** The chain is now exercised for `libSce*` with a real regression arm and
  a real positive control. R1 is the remainder, not a reopening.
- **Finding 3 (the guard).** Fully addressed, and better than I asked for: `kRootRefusedPlatformPrefixes`
  is a stronger answer than restoring one arm, the deleted arm is back and extended to `libkernel` /
  `libkernel_sys.sprx`, the reason string now names the matched prefix, and the honest-limit paragraph
  says what the rule does not cover instead of overstating it. R2 is one sentence inside that paragraph,
  not the paragraph.
- **Finding 4 (contract text).** `module_path_policy.hpp:4-9`, `:50-53`, `:67-72`, the verdict enum comment
  and `boot_program.hpp:202-206` all now describe the policy that exists. The new note recording that the
  50-dump measurement covers Media and was *not* re-run for the root — and that the root rule is therefore
  unexercised as a refusal across the corpus — is exactly the right way to carry a weaker claim.
- **S4 (AGC arithmetic).** Corrected to 100000, the invented `0x61b4c0` is retracted in place, and the new
  paragraph naming the cross-check as circular — "the independent part of the evidence is the register
  capture itself, not the arithmetic" — is a better statement of the evidence than the original claim was.
- **S2 (threshold).** Arm 0 pins it properly: `sub(100, 90)` true and `sub(100, 89)` false bracket the
  constant itself, the four corpus points land on the correct sides with their real measured numbers, and
  `sub(0, 0)` covers the divide-by-zero reading. This is the arm that would notice `kSubsumedFraction`
  being edited, and it is the right shape.

---

**REJECTED** on R1 and R2 only. Both are small and neither touches the design; S1 and S2 are welcome in the
same push. Post the new head and I will re-read the changed lines only — I expect that to be a formality.

mattias800 (commented):
# Re-review — head `1da4ac264`

**APPROVED.**

Both blocking items are fixed, and R1 is fixed the right way — by mutation rather than by assertion. I
re-derived the teeth myself rather than taking the claim: `libkernel.prx` is absent from
`boot_link_inputs`' fixed list, so `discover_extra_plugin_modules` does not dedup it away
(`boot_program.cpp:98-99`), it reaches `classify_module_path`, and only `kRootRefusedPlatformPrefixes`
refuses it. Delete that loop and `test_module_path_policy.cpp:231` reddens. That arm is real now.

Read against `1da4ac264` only; I built and ran nothing.

---

## Your two questions

### "Is the loop deletion the right mutation, or is there a narrower edit that slips through?"

The loop deletion is the right *primary* mutation — it is the whole rule, and four arms catching it is the
right answer. Two narrower edits I checked are also caught, and both are caught by arms that are doing more
work than they look like they are:

- **Prefix → exact equality** (`root_name == p` instead of `rfind(p, 0) == 0`): `libkernel.prx` no longer
  matches. Caught by the unit arm at `:129`.
- **Applying the list only to `.prx`, not `.sprx`**: caught by `:131`'s `libkernel_sys.sprx` arm. That is
  the one arm covering the extension axis of the platform rule, so it is worth keeping deliberately rather
  than as a "…and its `_sys` variant" afterthought.

**Three narrower edits do slip through, and they are per-entry.** `libc_`, `libmdbg` and `gaikai` have
**zero arms** in either test file (`grep -ci` = 0 in `test_module_path_policy.cpp` and
`test_dump_hygiene.py`). Delete any one of those array entries and the suite stays green. `libkernel` is
covered three times over and `libc.` once (`:127`), so the coverage is uneven rather than absent.

Low stakes — those three are the precautionary entries, and prosper implements neither `libmdbg` nor
`gaikai`, which is exactly why they are precautionary. But the fix is three lines next to the ones you
already wrote, and it makes the array's coverage uniform:

```cpp
refuse(R, "/dumps/PPSA00000-app0/libc_shim.prx",     ModulePathVerdict::SonyLibraryOutsideSceModule, "…");
refuse(R, "/dumps/PPSA00000-app0/libmdbg_syscore.prx", ModulePathVerdict::SonyLibraryOutsideSceModule, "…");
refuse(R, "/dumps/PPSA00000-app0/gaikai-player.prx",  ModulePathVerdict::SonyLibraryOutsideSceModule, "…");
```

One mutation that is *not* covered anywhere, noted so it is not mistaken for a gap in this PR: adding a new
auto-scanned directory. No root arm can catch that by construction — it is what
`src/host/image/AGENTS.md:42-46` exists to catch, procedurally.

### "Is the root-only condition right in BOTH directions?"

Yes. I worked the table rather than the reasoning, since you said you had only reasoned one way:

| path | loader | `dump_hygiene.classify` | agree? |
| --- | --- | --- | --- |
| `<root>/libkernel.prx` | refused (platform) | REFUSED | ✔ |
| `<root>/libc.prx` | refused (platform) | REFUSED | ✔ |
| `<root>/libSceAppContent.prx` | refused (Sony) | REFUSED | ✔ |
| `Media/Plugins/libkernel.prx` | **permitted** | None — `directory != ""`, and it is not `libsce*` | ✔ |
| `Media/Plugins/libSceAmpr.prx` | refused (Sony) | REFUSED | ✔ |
| `sce_module/libkernel.prx` | permitted | None | ✔ |
| `<root>/libcurl.prx`, `libwvoec.sprx`, `ulobjmgr.sprx` | permitted | None | ✔ |

The `Media/Plugins/libkernel*.prx` case you were specifically unsure about is the one the `directory == ""`
guard gets right: the loader permits it there (the Media rule is `libSce*`-only) and the tool stays quiet,
so the tool does not report on a policy nobody enforces. Both halves also sit behind the same
`MODULE_SUFFIXES` / extension gate, so data files are out of scope on both sides.

Two notes, neither a correction:

- **`--strip --yes` now deletes a class it did not before** (a root `libkernel.prx` / `libc.prx`). The
  tool's own contract still holds — REFUSED means "prosper will never link it, so removing it changes
  nothing prosper does", which is true after this change and was not before — and no local dump has such a
  file. Worth knowing rather than worth changing.
- The tool's message for a platform hit still reads *"a Sony-named library in the dump root"* while the
  loader's now names the matched prefix. Cosmetic, but the loader's phrasing is the better one.

**The one thing I would still add (not blocking): make the mirror a test, not a note.** `dump_hygiene.py`'s
own doctrine argues against itself here — the `PERMITTED_MODULE_DIRS` comment says the agreement is asserted
by a test *"rather than trusting this comment"*, and `PLATFORM_ROOT_PREFIXES` is now guarded by exactly the
comment that constant's neighbour declines to trust. The machinery is already there
(`tests/tools/test_dump_hygiene.py:41-49` parses `kPermittedModuleDirs[] = {` straight out of the header).
One caveat if you do it, because a naive copy silently fails: your entries carry trailing `//` comments, and
the existing parser's `s.strip().strip('",')` leaves `libkernel",   // libkernel, …` intact — it needs
`s.split("//")[0]` first. I would rather flag that than have you copy a parser that "passes" by comparing
garbage.

I am not gating on it because the failure direction is benign: a future drift makes the *tool* under-report,
never the loader under-refuse. The loader is the guard; the tool is the convenience.

### `dlc_emu` / `ampr_emu` — your decision to skip it is right, and better reasoned than my suggestion

Accepted, and I want it settled in the record rather than left as an open suggestion. Your reasoning holds
and there is a sharper version of it: those tools *impersonate* Sony libraries, so their modules are
`libSce*` **by construction** — that is the whole point of them — and `kSonyLibraryPrefix` therefore catches
them in the root, in `Media/*`, and everywhere else, while `fakelib/` is refused as a directory regardless of
naming. The prefixes would only add cover for a hypothetical future variant that is both non-`libSce`-named
*and* in the root, and there are zero instances across the 60 local dumps. Don't add them.

---

## Should-fixes: all three landed, and with teeth

- **Arm 2a.** `subsumed_spec.imports = { kMissing }` gives `CHECK(p.slots.empty(), …)` something to be
  wrong about: linked → `kMissing` is unresolved and creates a stub slot; skipped → nothing. I also checked
  the import cannot perturb the arm it sits in — `is_exported_symbol` requires `!s.is_import`
  (`linker.cpp:12-14`), so `measure_export_subsumption` still sees 2/2 and the skip still fires for the
  reason the arm claims. `main`'s `kShared` is satisfied by `release`, so no other slot appears.
- **Orphan.** `linker.hpp:117-121` — the doc comment is adjacent to `ExportCollision` /
  `find_export_collision` again, and `ExportSubsumption` carries its own at `:123-125`. Fixed.
- **Census.** ">=23, a floor not a count" is the right shape, and folding in the NID-ownership argument
  (`src/hle/` implements exactly `kernel/` and `libc/`, so the list covers the whole shadowing-relevant
  surface) upgrades the justification from a name heuristic to something checkable. That paragraph is now
  the strongest part of the header.
- **Tomb Raider.** Recorded with the correction that no guard exists. That is the durable record I wanted.

## Cleared across all three rounds

Corpus census and the `PPSA16901` smoke test; the end-to-end guard arm (now with a verified mutation); the
platform-prefix hardening and the restored `libc.prx` arm; every stale contract sentence in
`module_path_policy.hpp` and `boot_program.hpp`; the `dump_hygiene.py` drift; the AGC arithmetic and its
circularity note; the 0.90 threshold pinned by Arm 0; Arm 2a; the orphaned comment.

Nothing outstanding blocks merge. The three per-entry arms and the mirror test are welcome whenever, and
neither needs another review round — if you land them here, note that this approval is on `1da4ac264`, so
per the charter establish by content (`git diff 1da4ac264 <new head>`) that a test-only addition is all that
moved.

**APPROVED**

Issue #2206, opened 2026-08-06
Sonic Frontiers (PPSA03831): the title-screen heading reads "Try Again" and a panel over the menu renders almost no text


mattias800:
## Summary

With the boot stall fixed (`sceSaveDataTransferringMountPs4`, PR for #2023), Sonic Frontiers reaches
and renders its title screen and main menu at 3840x2160. Two UI-content defects remain on that
screen. Both are content/selection defects, not composition failures: the SurfRide layer composites
correctly and most strings are legible.

Rung is **2** with these open; they do not block progression.

## 1. The title-screen heading renders "Try Again"

Where the SONIC FRONTIERS logo belongs, the screen draws the string **"Try Again"** in large blue
type. Everything around it is correct: the menu column reads "New Game", "Language", "Carry over
from PlayStation®4", "Copyright", "User manual", "Extras" (the last selected, with a `!` badge), the
version string `1.41` draws in the right place, and the localized `ui_gamemodetitle_en.pac` is the
package the guest loaded. So text rendering, the font atlas and the language selection all work —
what is wrong is *which* string the heading element resolves to.

Evidence: `assets/screenshots/sonic-frontiers-main-menu.png` (direct unmodified `tools/screenshot`
capture, no input, t = 220 s of a 300 s default arm).

## 2. A panel slides over the title screen and renders almost no text

Shortly after the main menu appears, a full-width panel with a blue header band covers the middle of
the screen and stays for the rest of the run. Its background art (the SurfRide chevron pattern) draws
correctly; its body contains only a handful of isolated glyph marks and a diamond cursor where a list
or a message belongs. **No input was driven in either arm**, so whatever opens it opens by itself.

Reproduced in two independent 300 s arms, one against a populated save area and one against a fresh
`PROSPER_SAVE0` directory, so it is not save-state dependent.

## Route

```bash
PROSPER_GUEST_ARGS=-force-gfx-direct PROSPER_RENDER=1 \
timeout --foreground -s TERM -k 5s 340s \
  ./build/screenshot <DUMP_ROOT>/PPSA03831-app0 \
  --seconds 20 --count 15 --timeout 320 --out ~/work/shots --manifest ~/work/manifest.json
```

Sample past 50 s (instrument trap 87). The title screen lands ~t = 200 s, the menu ~t = 220 s, the
panel from ~t = 240 s.

## Not yet investigated

Which SurfRide element the heading is, whether the two share a cause (both are text content on the
[...]

Issue #1715, opened 2026-08-01
Generated geometry stage has no OpExecutionMode Invocations, so its invocation count is 0


mattias800:
Found by the first-ever Vulkan validation pass over prosper's test suite (#1704). Reproduces on
Mesa 25.2.8 lavapipe (the CI runner's exact driver) and on RADV.

```
VUID-VkPipelineShaderStageCreateInfo-stage-00715
vkCreateGraphicsPipelines(): pCreateInfos[0].pStages[1] SPIR-V (Geometry stage) entry point must
have an OpExecutionMode instruction that specifies an invocation count that is greater than 0 and
less than or equal to maxGeometryShaderInvocations.
Invocations = 0
maxGeometryShaderInvocations = 32
```

Test: `interp_render`, 1 pipeline per run.

## Where

`grep -rn Invocations prosper/src/gpu/` finds no emission site: the generated geometry stage
declares no `OpExecutionMode ... Invocations` at all, so the implied count is 0, which is out of
range. The stage is the one prosper synthesises to feed explicit-parameter / raw-vertex pixel
inputs (the `[ok] Vulkan links and renders the generated interpolation geometry stage` assertion
sits right after the message).

## Why it matters

Both drivers measured run the stage anyway — invocations are treated as 1 — and the interpolation
assertions pass. A driver that takes the declaration literally runs the stage zero times and the
draw silently loses its varyings, which presents as "this title's interpolated attributes are
black" with nothing to point at.

## Suggested fix

Emit `OpExecutionMode %main Invocations 1` on the geometry entry point (or the real count when the
lowering ever needs more than one), and add a decode-side assertion so a future geometry emitter
cannot omit it.

Currently allow-listed in `prosper/tools/vkval/allowlist.txt` so it stays visible.

mattias800:
Measurement while covering every SPIR-V emitter for #1711 (PR #1726), so nobody re-derives it: **`spirv-val` accepts this module.**

`recompile_interpolation_geometry` on the explicit-parameter layout from `test_interp_render` produces 354 words that pass under **both** `spirv-val <file>` (universal) and `spirv-val --target-env vulkan1.1 <file>`, rc 0 in each case. Its execution modes are:

```
OpEntryPoint Geometry %20 "main" …
OpExecutionMode %20 Triangles
OpExecutionMode %20 OutputTriangleStrip
OpExecutionMode %20 OutputVertices 3
```

[...]

Issue #264, opened 2026-07-09
gpu: stencil test not wired into the pipeline — masks don't clip (The Messenger title shimmer shows full polygon)


mattias800:
## Summary
The stencil test is **decoded but never applied**. `render_state.cpp:56` reads `DB_DEPTH_CONTROL.STENCIL_ENABLE` into `rs.stencil_enable`, but nothing propagates it to the pipeline — `ResolvedPipelineState` has no stencil fields, and `tests/render_runner.h`'s `VkPipelineDepthStencilStateCreateInfo` sets only depth (`depthTestEnable`), never `stencilTestEnable`/`front`/`back`.

Result: any draw that relies on a **stencil mask renders unmasked**. On The Messenger's title screen, the animated "shimmer" is a white quad swept across the logo and masked to the logo's shape via the stencil buffer — with no stencil test, the **full white polygon shows** instead of only where it intersects the title. (User-reported.)

## What's needed (a real feature)
1. Ensure the depth attachment format includes stencil (e.g. `VK_FORMAT_D24_UNORM_S8_UINT`) — `render_runner.h` currently uses a depth format; verify/switch.
2. Decode stencil op state: `DB_STENCIL_CONTROL` (STENCILFUNC/FAIL/ZPASS/ZFAIL ops, front+back), `DB_STENCILREFMASK` (ref, compare mask, write mask). Add to `ResolvedPipelineState`.
3. Set `VkPipelineDepthStencilStateCreateInfo.stencilTestEnable` + `front`/`back` `VkStencilOpState` (+ dynamic or baked ref).
4. Handle stencil clear (`DB_STENCIL_CLEAR` / `DB_RENDER_CONTROL.STENCIL_CLEAR_ENABLE`).

## Verify
Confirm The Messenger's shimmer draw sets `STENCIL_ENABLE` (log per-draw), i.e. it's stencil-based and not a shader alpha-mask, before/while implementing. Registers already have `pm4_registers.hpp` DB_STENCIL_* definitions.

Reported alongside the S# sampler fix (#260) for the title screen.

mattias800:
CLAIMING: claude-code session 22c0625b | branch fix/issue-264-stencil-support | 2026-07-09T23:28:00Z

[...]

PR #3511, opened 2026-09-10
gpu: derive a RectList's fourth corner from the geometry, not from a fixed index (#3507)

### LINKED ISSUE #3507: RectList fourth-corner synthesis assumes the shared corner is always P0, so a rectangle supplied in another order draws half a quad (Darksiders II intro FMV)

mattias800:
## The defect

A PS5 RectList hands the GPU **three** corners of a rectangle and the hardware derives the fourth
after vertex shading. Vulkan has no RectList topology, so `recompile_interpolation_geometry(...,
synthesize_rect=true)` emits a geometry stage that derives it and outputs a four-vertex strip.

That stage assumed the **shared** corner — the one adjacent to both others — is always `P0`. It
computed the fourth as `P1 + P2 - P0` and emitted the vertices in the supplied order.

*Darksiders II* (`PPSA23806`) supplies its video blit as `P0`=bottom-right, `P1`=top-left,
`P2`=top-right, so its shared corner is `P2` and the missing one is bottom-left. Under the old rule
the synthesized corner was clip `(-1, 3)` — off-screen — so the strip's second triangle never
rasterized and the intro FMV rendered as its upper-right triangle alone.

before

*Darksiders II intro FMV, `tools/screenshot`, default launch with `PROSPER_PAD_PRESS=1`, 3840x2160 —
before.*

## Evidence, not inference

The frame was captured (`PROSPER_GRAB_BUNDLE_AFTER_MS`) and replayed offline; `gpu_replay --bundle`
reproduces the artifact to within a rounding step of the live capture (`TL=3.0 TR=15.3 BL=0.0
BR=1.6` against live `3.0 / 15.4 / 0.0 / 1.9`). `--inspect-only` names the culprit: of the frame's
three draws, `draw[1]` is the video blit — three YUV planes, 1280x720 luma plus two 640x360 chroma —
and it is the only one that leaves realization at `vcount=3` where the other two show `vcount=4
raw=3`, i.e. expanded.

Dumping its vertex buffer (`--dump-resource 1:vs:3`) gives the three positions `(1,-1)`, `(-1,1)`,
`(1,1)`; under the draw's flipped viewport (`3840.0x-2160.0`) those are bottom-right, top-left,
top-right — exactly the visible triangle. Its texcoords corroborate the same shared corner
independently: `(1,1)`, `(0,0)`, `(1,0)`, whose completion `v0+v1-v2 = (0,1)` is the correct
bottom-left UV.

## The fix

The shared corner is identified **geometrically** from the shaded positions: of the three pairwise
separations the largest is the diagonal, and the vertex not on it is the shared corner. For a
rectangle this is exact — the diagonal strictly exceeds either side — and it needs no convention
guess about vertex order.

Emission order is the other half, and a corner fix alone does not cover it. A strip's two triangles
share its middle edge, so the **diagonal pair must sit there**: vertices go out as `shared,
diagonal_a, diagonal_b, synthesized`. Emitting the supplied order when the shared corner is not `P0`
yields two triangles sharing a *side* — they overlap on one half and leave the wedge between the two
diagonals uncovered, which showed as a black triangle with its apex at the exact screen centre.
(That intermediate state is why this PR changes both.)

Vulkan flips the winding test on a strip's odd triangle, so both halves keep the same facing under
culling.

**Both halves reduce to today's behaviour exactly when the shared corner is `P0` — which is exactly
when the old rule was right — so no draw that renders correctly today changes.** Every varying and
the explicit-parameter barycentrics follow the same selected index.

## Deliberately unaffected

- The **procedural** RectList path (`vb_entries == 0`, expand three vertices to four) is untouched.
  It handled 4,803 of the intro's 7,200 RectList draws and is what *Blasphemous 2*'s clear shader
  uses.
- The plain `requires_geometry` path (P0/P10/P20 lowering without rect synthesis) gains no instruction
  and no type: `slot_of` is the identity when `synthesize_rect` is false, the barycentric constants are
  unchanged for vertices 0..2 (vertex 3 does not exist there), and the `bvec4` type is declared only
  when synthesizing. It is **not** byte-identical, though — an earlier revision of this section claimed
  that and it was wrong. Loading the positions earlier renumbers the ids after it, and the
  `consumes_parameters` guard elides a composite nothing consumed. Both are semantically inert and ids
  are module-internal, so the no-regression conclusion stands; the stronger claim did not.

## The filed hypothesis was wrong, and is recorded as such

#3507 as filed blamed `needs_rect_list_synthesis`'s vertex-buffer precondition for **declining**
synthesis on a vertex-less blit. A census over the intro falsifies both halves: **7,200 RectList
draws — 2,397 synthesized, 4,803 procedurally expanded, zero unhandled**, so the precondition never
declined a draw; and the blit is not vertex-less, carrying two vertex buffers at stride 16. The
declines that do occur are `vertex_resources == nullptr`, which reaches the procedural path and is
handled. Landed as a row in `RECOMPILER_REMAINING.md`'s `## Ruled out`.

## Review outcome

Independently reviewed; **rejected first time, with two blocking findings that were real.**

1. **Invalid SPIR-V.** An `OpSelect` with a vector result needs a vector condition — a scalar one is
   legal only from SPIR-V 1.4, and this emitter stamps 1.3. Fixed by broadcasting both selectors to
   `bvec4`. CI's `Linux` job had already failed on exactly this; `ctest` alone cannot see it, and
   `tools/vkval/vk_validation_scan.py` is the gate that can.
2. **The strip's facing was reversed when the shared corner is `v1`.** A strip's facing is its first
   triangle's, and the guest cull state passes through with no RectList exemption, so such a draw
   would have rendered *nothing* under culling where it previously rendered half. "Shared first, then
   the other two ascending" is the odd permutation `(v1,v0,v2)`. The order is now a **rotation**,
   `supplied[(shared + slot) % 3]` — diagonal still in the middle, and even in all three cases.
   Darksiders II shares at `v2`, an even case, which is why the live route was blind to it.

Also from review: the byte-identity claim above was withdrawn, `CONFIDENCE: MED` recorded on the
discriminator (a sufficiently sheared parallelogram breaks "longest separation is the diagonal", and
separation is measured on the full 4-component clip position, exact when `w` is constant), and the
Ruled-out row now cites this PR.

**Known gap, not closed:** the `kI`/`kJ` barycentric table is not exercised by any guard — the test
builds no `PixelSystemInputMapping` — and finding 2 moved two of its cells. It is verified by two
readings, not by a test.

## Verification

```
cmake --build prosper/build-linux -j6          # clean, 0 errors
ctest --no-tests=error -j4                     # 461/461 passed, exit 0
```

Also `python3 prosper/tools/vkval/vk_validation_scan.py --build-dir prosper/build-linux` → exit 0,
"every observed validation message ID is accounted for".

New: `test_rect_synthesis_render` renders **one** rectangle three times, permuting the supplied order
with an index buffer so the shared corner lands at slot 0, 1 and 2, and runs the whole set **twice** —
once with culling off, once with back-face culling on. The culling pass asserts the stage *preserves*
the facing it was handed: an arm renders in full exactly when its own index permutation reversed the
winding, all of it or none, never half. That is the only thing that pins the emission order. All three describe the same
rectangle and must render identically — full coverage plus the correct varying at all four corners,
including the synthesized one, whose value `0.5 + 0.7 - 0.2 = 1.0` is pinned uniquely.

**Arm 0 is the positive control**: it is the case the old rule got right, so it passes either way and
shows the harness is capable of passing.

**Mutation arm** (shared corner pinned back to `v0`, both formula and order):

| arm | coverage | corner varyings |
|---|---|---|
| slot 0 (positive control) | 3600/3600 | all 4 correct |
| slot 1 | **1770/3600** | 3 of 4 wrong |
| slot 2 (Darksiders II's blit) | **1770/3600** | 3 of 4 wrong |

With the fix all three arms read 3600/3600 with all four corners correct.

**Second mutation arm**, for the emission order (ascending restored for shared == `v1`): with culling
**off** every arm still passes — which is exactly why the first version of this test missed it — and
with culling **on** the shared-at-`v1` arm drops to **0/3600** and fails.

## Live results

after

*The same moment of the same movie after the fix — `tools/screenshot`, default launch with
`PROSPER_PAD_PRESS=1`, 3840x2160. The cliff, tree trunks, cloaked figure and lantern glow that were
black below the diagonal are all present.*

legal screen

*Darksiders II's legal/title screen, same run, 3840x2160.*

Regression-checked live at their documented milestones, all unchanged: **The Messenger** (rung 6,
intro scene), **Blasphemous 2** (rung 6, title screen), **Sonic Frontiers** (rung 2, title screen and
main menu — the "TIME UP" overlay is the pre-existing #2206), **GTA V** (rung 2, title screen).

Rung unchanged for Darksiders II at 2 — this fixes a movie, not the ladder.

Fixes #3507
Refs #3497


mattias800 (commented):
## REJECTED — on head `2e819ef55db66fcddf5a9882875b8c3af6c2ad93`

Reviewed against the pushed head only. Two of the four blocking findings below are what the
**Linux** job is red on right now, so they are not judgement calls. The mechanism, the corner
selection, the slot table and the barycentric constants are all correct as far as I can check them,
and the falsification recorded in `RECOMPILER_REMAINING.md` is the right shape — the rejection is
about how the geometry is *emitted*, not about the diagnosis.

Note for the author: the local worktree at the path in my brief has **uncommitted** edits to
`rdna2_to_spirv_internal.hpp`, `rect_synthesis_vs.spvasm`, `test_rect_synthesis_render.cpp` and
`allowlist.txt` that already implement B1 and B3 below (a `t_v4bool` broadcast, and the ledger line).
Nothing of that is pushed, so it is not part of this review — but do not redo it.

---

# Blocking

## B1. The emitted geometry module is invalid SPIR-V: a scalar-condition `OpSelect` with a vec4 result

`OpSelect` may take a scalar Boolean condition with a **vector** Result Type only from **SPIR-V 1.4**.
Before 1.4 the condition must be a `bvec` of the same width. Every module this emitter produces
declares **1.3** (`prosper/src/gpu/recompiler/rdna2_to_spirv_internal.hpp:4623`,
`0x00010300u`), and the new selects hand a scalar `t_bool` (`OpTypeBool`, declared at `:4380`) to a
`t_v4f` result at six sites:

- `:4463`, `:4465` — `synthesize_fourth`
- `:4504`, `:4505` — `slot_of` slot 0
- `:4507` — `slot_of` slot 1
- `:4512` — `slot_of` slot 2 (its `OpLogicalOr` at `:4510` is scalar too, so widening this one needs
  the `Or` widened with it)

The two scalar-result selects in `by_shared_corner` (`:4519`, `:4522`, result `t_f32`) are fine.

**CI already found this**, on this exact head — the `Vulkan validation scan` step of the `Linux` job
(run 34423091869, job 102702541836):

```
[vkval] NEW — not on the allow-list:
  VUID-VkShaderModuleCreateInfo-pCode-08737  x1  [rect_synthesis_render]
      vkCreateShaderModule(): pCreateInfo->pCode (spirv-val produced an error):
      Expected vector sizes of Result Type and the condition to be equal: Select
        %58 = OpSelect %v4float %49 %55 %57
```

Reproduced independently here with SPIRV-Tools v2026.1: a module stamped `0x00010300` carrying that
pattern is rejected under **every** target env I tried (`vulkan1.1`, `vulkan1.2`, `vulkan1.3`,
`vulkan1.4`, `spv1.6`) — the binary's *declared* version governs, not the device's. The same source
assembled at `spv1.4`/`spv1.5`/`spv1.6` validates. A `bvec4` condition built with
`OpTypeVector %bool 4` + `OpCompositeConstruct` validates at `spv1.3`/`vulkan1.1`, so that is the
minimal fix.

Worth recording *why* this got through, because it is a nice trap: the test's own hand-written vertex
shader uses the identical pattern (`rect_synthesis_vs.spvasm:41-44`) and is legal only because
`spirv-as` stamped it **1.6** (`kRectCornerVs[1] == 0x00010600`). Assembling that same `.spvasm` with
`--target-env spv1.3` reproduces the exact same validator error. So the pattern was demonstrably
"working" three feet away from the emitter that could not use it.

**Also fix the gate hole in the same PR.** `prosper/tools/spv_validate/spv_validate.cpp:824` emits
`recompile_interpolation_geometry(layout)` — default `synthesize_rect=false` — so the rect variant has
**never** been validated by the emitter gate, and `check_emitter_coverage()` is satisfied by the
non-rect module, so the omission reports nothing. That is the same shape as the `#1711` story in that
file's own header. One extra `dump(...)` with `synthesize_rect=true` closes it, and it is the check
that would have caught B1 before the render test existed.

## B2. `VUID-...-stage-00715` fires from a test the vkval ledger does not record

Same CI step, second failure:

```
[vkval] allow-listed id, but from a test the ledger does not cover:
  VUID-VkPipelineShaderStageCreateInfo-stage-00715  in rect_synthesis_render  (recorded for: interp_render)
[vkval] FAIL: 1 allow-listed id(s) appeared in a test the ledger does not record.
```

`prosper/tools/vkval/allowlist.txt:85` scopes the `#1715` deferral to `interp_render`. The new test
creates the same synthesised geometry entry point, so it needs adding to that entry's test list.
Mechanical, but it independently reddens the gate.

## B3. The strip's facing is reversed when the shared corner is `v1`

Vulkan builds a triangle strip as `(v0,v1,v2)` then `(v2,v1,v3)` — the odd triangle's order is
swapped precisely so both halves keep one orientation. So the whole strip's facing is the facing of
its **first** triangle, `(slot0, slot1, slot2)`.

`slot_of` (`:4497-4515`) produces, and I checked each select by hand:

| shared | slot0 | slot1 | slot2 | permutation of the supplied `(v0,v1,v2)` |
|---|---|---|---|---|
| v0 | v0 | v1 | v2 | identity — winding preserved |
| v1 | **v1** | **v0** | **v2** | **transposition — winding REVERSED** |
| v2 | v2 | v0 | v1 | rotation — winding preserved |

`prosper/src/gpu/state/render_state.cpp:1217` maps `PA_SU_SC_MODE_CNTL.CULL_FRONT/CULL_BACK` straight
into `ResolvedPipelineState::cull_mode` with no RectList exemption, so a RectList draw whose shared
corner lands at index 1 **and** whose guest state enables culling now renders nothing, where today it
renders half a quad. That is a new failure mode introduced by this PR, and it is silent.

The fix also simplifies the rule: emit `(v1, v2, v0, X)` for shared==v1. The middle edge is still the
diagonal pair `{v0,v2}`, and the emission rule collapses to *"rotate the supplied triple so the shared
corner is first"* — `slot1 = supplied[(s+1)%3]`, `slot2 = supplied[(s+2)%3]` — which is winding-
preserving in all three cases by construction rather than by luck in two of them.

Two cells of the barycentric table move with it (the shared==v1 column only, `:4555-4556`):

- `kJ[1]` `{0, 0, 0}` → `{0, 1, 0}`  (slot 1 becomes v2, whose (i,j) is (0,1))
- `kJ[2]` `{1, 1, 0}` → `{1, 0, 0}`  (slot 2 becomes v0, whose (i,j) is (0,0))

`kI[1]` and `kI[2]` are unchanged under the rotation. Everything else in the table stays.

The current test cannot see any of this: `ResolvedPipelineState::cull_mode` defaults to
`VK_CULL_MODE_NONE` (`render_state.hpp:308`), so arm 1 passes with the facing inverted. An arm with
`state.cull_mode = 2` and `front_face = 0` asserting the rectangle still covers the target would pin
it.

## B4. The "byte-identical SPIR-V" no-regression claim is false as stated

The PR body says the plain `requires_geometry` path "emits byte-identical SPIR-V". It does not, in
two independent ways:

1. The positions load moved from **after** the parameters loop (`origin/main`'s
   `rdna2_to_spirv_internal.hpp:4419-4426`) to **before** it (`:4418-4426` on this head). `id()` is a
   monotonic counter, so every result id allocated from the parameters loop onward shifts, and the
   code section's instruction order changes with it. Any layout with at least one attribute differs.
2. The new `if (consumes_parameters)` guard (`:4547`, `:4550`) *removes* the per-vertex barycentric
   `OpCompositeConstruct` when no system output is bound. Previously it was always built and simply
   stored nowhere (`origin/main:4453-4459`). With it gone, `fconstf(0.0f)`/`fconstf(1.0f)` may also
   never be interned, so the constants section differs too.

Both are semantically neutral and I do not think either changes behaviour — the *conclusion* survives.
But the claim as written is the load-bearing evidence for "no draw that renders correctly today
changes", and it is checkable and wrong, which is exactly the kind of citation CLAUDE.md says
propagates. Either reword it to "semantically identical, ids renumbered" (and say the dead barycentric
construct is now elided), or keep byte-identity for real by leaving the positions load where it was
and moving only the `if (synthesize_rect)` distance block below the parameters loop. I flag it
blocking because it is a verification claim, not prose.

---

# Non-blocking

**N1. The geometric rule is a heuristic, not an identity, and nothing says so.** "The largest of the
three pairwise separations is the diagonal" is exact for a rectangle, and more generally for any
parallelogram whose two edges from the shared corner meet at ≥ 90°. RectList on this hardware makes a
parallelogram from *any* three vertices, and for a sheared one the rule picks the wrong corner: with
S=(0,0), A=(10,0), B=(9,1), the largest separation is |SA|²=100, so the rule names **B** the shared
corner and synthesizes S+A−B instead of A+B−S. The comment at `:4278-4281` is honest that the argument
is a rectangle argument, but it does not say what happens off that assumption, and the code carries no
`CONFIDENCE:` marker — CLAUDE.md asks for one on re-derived platform semantics. Note also that the
separation is measured on the **full 4-component clip position** (`Op_Dot` over `t_v4f` at `:4435`),
so a perspective-projected rect with varying `w` is a 4D parallelogram, not a rectangle — that is
precisely the case that can reach the heuristic's blind spot. A `CONFIDENCE: MED` and two lines saying
"exact for a rectangle; a sheared parallelogram would need the hardware's fixed rule, which we have
one draw's worth of evidence against" would make this durable.

**N2. The barycentric table is not exercised by anything.** The new test builds its layout with
`fragment_interpolation_layout(ps, std::size(ps))` and no `PixelSystemInputMapping`, so
`system_locations` are all unused, `consumes_parameters` is false, and `kI`/`kJ` are never emitted.
The PR body itself says an error there "would be silent". I checked all twelve pairs by hand against
the stated basis (P0=A0, P10=A1−A0, P20=A2−A0; supplied v0/v1/v2 at (0,0),(1,0),(0,1); synthesized
corner negating the shared axis) and **they are correct for the current emission order** — but that is
me, not a test. `spv_validate.cpp:821` already constructs a `perspective_center` mapping; the same
layout plus `synthesize_rect=true`, asserting the reconstructed value at the synthesized corner, would
turn this from "checked once by a reviewer" into a guard.

**N3. Redundant emitted work.** `slot_of` emits fresh `OpSelect`s per (output vertex × attribute), so
a 32-attribute rect emits roughly 4×33 selects where four hoisted per-slot picks would do; and
`consumes_parameters` (`:4547`) is recomputed inside the per-vertex loop although it is loop-invariant.
Cosmetic — these stages are tiny — but the first one scales with attribute count.

**N4.** `prosper/docs/RECOMPILER_REMAINING.md`'s new Ruled-out row ends `#3507, PR TBD`. Fill in
`#3511` before merge.

---

# Verified correct (so nobody re-derives it)

- **`Op_Dot = 148`** is right. Confirmed by scanning `spirv-as` output for the opcode word:
  `%15 = OpDot %6 %14 %14` encodes as opcode 148, length 5. Operands `t_v4f`/`t_v4f` with result
  `t_f32` are a valid `OpDot`.
- **No path can emit an `OpSelect` referencing id 0.** `rect_shared_is_v0`/`_is_v1` have exactly three
  consumers — `synthesize_fourth` (both call sites guarded by `if (synthesize_rect)`), `slot_of`
  (early-returns `supplied[slot]` when `!synthesize_rect`, `:4499`), and `by_shared_corner` (only
  reached inside the `if (synthesize_rect)` arm at `:4553`). Slot 3 is unreachable when
  `output_vertices == 3`.
- **The nesting of `rect_shared_is_v1` is correct at all three uses.** `synthesize_fourth` and
  `slot_of` slot 0 put it strictly inside the `is_v0` select; slot 1 does not consult it at all; slot 2
  ORs it with `is_v0`, so a spuriously-true `is_v1` (which happens whenever d02≥d01 and v0 is shared)
  is absorbed — both branches yield `supplied[2]` in that case. I could not construct an input where
  the stale value changes the answer.
- **Degenerate inputs.** All three coincident → `d01=d02=d12=0` → `is_v0` true (`>=` on both) → fourth
  = v1+v2−v0, degenerate, no NaN. Two coincident (v0==v1) → d01=0, d02=d12=d → `is_v0` true → fourth
  = v2, a zero-area strip. Collinear at 0,1,2 → d01=1, d02=4, d12=1 → `is_v0` false, `is_v1` true →
  shared = the middle vertex, fourth = v0+v2−v1 = v1. All sane; the `>=` comparisons make every tie
  land somewhere defined, and `FOrd*` sends NaN to the v2 branch rather than anywhere undefined.
- **The slot table matches the PR's claim** — (v0: v1,v2), (v1: v0,v2), (v2: v0,v1) — see B3's table;
  I disagree only about which of the two valid diagonal orders should be chosen for v1.
- **The test's arms are genuine, and arm 0 is a real positive control.** Arm 0 `{0,1,2}` gives
  d01=4, d02=4, d12=8 → `is_v0` true → identical to the old rule, so it passes either way. Arm 1
  `{1,0,2}` gives d01=4, d02=8, d12=4 → shared v1; arm 2 `{1,2,0}` gives d01=8, d02=4, d12=4 → shared
  v2. Under the old rule both synthesize clip (−3,1), off-screen, so both lose one triangle of the
  strip — the reported 1770/3600 is the right order of magnitude for half of a 60×60 sample window.
  The varying is a second, independent discriminator: 0.2/0.5/0.7 at the supplied corners and 1.0 at
  the synthesized one, which only the correct completion produces. I could not find a way for an arm to
  pass for the wrong reason, with the single exception of B3, which the default `cull_mode` hides.
- **The stale-blob bug really is fixed.** `kRectCornerVs` at this head is **byte-identical** to
  `spirv-as --target-env spv1.6 prosper/tests/gpu/recompiler/rect_synthesis_vs.spvasm` — 217 words,
  `cmp` clean. (Its declared version 1.6 is fine: `kVulkanRuntimeVersion` is `VK_API_VERSION_1_4`.)
- Screenshots are 1920-wide WebP, the `BLOG.md` entry is image-led and one paragraph, and the
  `## Ruled out` row is in the right doc. All as the charter asks.

mattias800:
## Review addressed — all four findings

Thanks; **B3 was a real bug and the test could not have caught it.** Confirmed each finding independently before acting.

**B1 (invalid SPIR-V) — fixed.** Both selectors are broadcast to `bvec4`; the type is allocated and declared only when `synthesize_rect`, so the non-synthesis module gains no type and no instruction. `vk_validation_scan.py` now exits 0 ("every observed validation message ID is accounted for") — locally reproducing the gate that failed in CI. I should have run that scan before pushing a change that emits new SPIR-V; `ctest` alone cannot see it.

**B2 (vkval ledger) — fixed.** `rect_synthesis_render` added to the #1715 entry at `allowlist.txt:85`. Same generated stage, same defect, which is the case that entry documents; I did not widen scope into fixing #1715 itself.

**B3 (reversed facing at shared == v1) — confirmed and fixed.** Verified independently: `(v1,v0,v2)` is parity 1, `(v1,v2,v0)` is parity 0, and both keep the diagonal in the middle. Now a rotation, `supplied[(shared + slot) % 3]`, even in all three cases. `kJ[1] {0,0,0}→{0,1,0}` and `kJ[2] {1,1,0}→{1,0,0}`; `kI` unchanged, exactly as you derived. Darksiders II shares at v2 — an even case — which is why the live route was blind to it.

**The test now guards it.** Every arm runs a second time under back-face culling, asserting the stage *preserves* the facing it was handed: an arm renders in full exactly when its own index permutation reversed the winding — all of it or none, never half. Mutation arm (ascending order restored for shared == v1): **with culling off every arm still passes**, and with culling on the shared-at-v1 arm drops to 0/3600 and fails. That is the whole point — nothing without a cull pass could have seen this.

**B4 (byte-identical claim) — withdrawn.** You are right on both mechanisms. Corrected in the code comment and below; the no-regression conclusion survives (ids are module-internal, the elided composite was dead), but the claim as written was checkable and wrong.

**Non-blocking:** `CONFIDENCE: MED` now recorded on the discriminator, naming the sheared-parallelogram blind spot and the constant-w assumption behind measuring separation on the full 4-component clip position. Ruled-out row now cites #3511 instead of "PR TBD". Left alone: the per-(vertex × attribute) select duplication and the loop-invariant `consumes_parameters` — both are emit-time only and I would rather not add motion to a stage this PR is already reworking.

**On the `kI`/`kJ` coverage gap you flagged** — still not exercised by a guard, and I have not closed it. The test builds no `PixelSystemInputMapping`, so the table is verified only by your reading and mine. It is now more load-bearing than when you flagged it, since B3 moved two of its cells. If you want that arm before approving, say so and I will add one modelled on `test_interp_render`'s explicit-parameter fixture; otherwise it wants its own issue.

Re-verified at head `68f98fb27`: build clean, `ctest --no-tests=error -j4` 461/461 exit 0, vkval exit 0, Darksiders II's intro FMV still renders complete.

mattias800 (commented):
## APPROVED — on head `68f98fb278b103ea817a7969734262c27376172b`

All four blocking findings are addressed, and each fix is right for the reason given rather than by
coincidence. I re-derived the load-bearing parts independently rather than checking them against the
follow-up commit's own description. Details below, then the two things I would still like, neither
blocking.

---

# (a) The rotation, and `kI`/`kJ`

**The rotation is correct at all three slots.** `slot_of`
(`prosper/src/gpu/recompiler/rdna2_to_spirv_internal.hpp:4535-4548`) now computes
`if_vN = supplied[(N + slot) % 3]` and nests `Select(is_v0, if_v0, Select(is_v1, if_v1, if_v2))`,
which evaluates to `supplied[(shared + slot) % 3]`:

| shared | slot 0 | slot 1 | slot 2 | slot 3 | diagonal in the middle? | permutation |
|---|---|---|---|---|---|---|
| v0 | v0 | v1 | v2 | X | yes — {v1,v2} | identity (even) |
| v1 | v1 | v2 | v0 | X | yes — {v2,v0} | rotation (even) |
| v2 | v2 | v0 | v1 | X | yes — {v0,v1} | rotation (even) |

Shared corner leads, the synthesized one is last, the diagonal pair occupies slots 1-2 in every case,
and every row is a cyclic rotation, so the facing is preserved by construction rather than in two
cases out of three. Shared==v0 is still the identity, so nothing that renders correctly today moves.
The old `Op_LogicalOr` special case is gone, which also removes the only place `rect_shared_is_v1` was
consulted un-nested.

**`kI`/`kJ` match the rotation.** I re-derived all twelve pairs from the stated basis
(value = A0 + i·(A1−A0) + j·(A2−A0), supplied v0/v1/v2 at (0,0)/(1,0)/(0,1)) against the table at
`:4590-4591`, and against `by_shared_corner`'s argument order (`[0]`→shared v0, `[1]`→v1, `[2]`→v2,
which is the `kX[slot][shared]` indexing the new comment claims):

| slot | vertex carried (shared v0 / v1 / v2) | kI row | kJ row | committed |
|---|---|---|---|---|
| 0 | v0 (0,0) / v1 (1,0) / v2 (0,1) | `{0,1,0}` | `{0,0,1}` | matches |
| 1 | v1 (1,0) / v2 (0,1) / v0 (0,0) | `{1,0,0}` | `{0,1,0}` | matches |
| 2 | v2 (0,1) / v0 (0,0) / v1 (1,0) | `{0,0,1}` | `{1,0,0}` | matches |
| 3 | A1+A2−A0 (1,1) / A0+A2−A1 (−1,1) / A0+A1−A2 (1,−1) | `{1,-1,1}` | `{1,1,-1}` | matches |

So `kJ[1] {0,0,0}→{0,1,0}` and `kJ[2] {1,1,0}→{1,0,0}`, `kI` untouched — exactly the two cells, and no
others. This is now the second time I have derived this table from the basis rather than read it, and
the second derivation was done before I opened the follow-up diff; it agrees. That is about as much
confidence as is available without a test, which is why I still want the guard below.

# (b) The `bvec4` broadcast

Correct, and scoped tightly.

- `t_v4bool` is allocated only under `synthesize_rect` (`:4324`) and declared only under it (`:4407`),
  after `OpTypeBool`. So the non-synthesis module consumes **no id** and gains **no instruction** —
  I checked that `id()` is genuinely not called on the false branch, which is what makes that claim
  true rather than merely intended.
- `broadcast` (`:4477-4484`) emits `OpCompositeConstruct %v4bool %c %c %c %c` from the scalar
  selector, inside the `if (synthesize_rect)` block and after both selectors are defined. Same shape
  as the existing barycentric `CompositeConstruct`, so `putv`'s operand order is right.
- **No vec4 select takes a scalar condition any more.** `grep -rn "Op_Select, *{t_v" prosper/src/gpu/recompiler/`
  returns exactly four hits — `:4499`, `:4501` (`synthesize_fourth`) and `:4544`, `:4546`
  (`slot_of`) — and all four take `rect_shared_is_v0_wide` / `rect_shared_is_v1_wide`. The two
  remaining scalar-condition selects (`:4552`, `:4555`, `by_shared_corner`) have `t_f32` results,
  which is legal at any version, and `Op_LogicalAnd` at `:4473` is scalar→scalar.
- Independently reproduced the rule this fixes: a module stamped `0x00010300` carrying
  `OpSelect %v4float %<scalar bool> …` is rejected by spirv-val under every target env; the same
  module with a `bvec4` condition validates under `vulkan1.1`. So the shape now emitted is the one
  that passes.

**The test's own VS was fixed too, and better than I asked.** It is now assembled at
`--target-env vulkan1.1` and declares **1.3** instead of 1.6, so the fixture shader is held to the
same rule as the stage it exists to test. I verified the embedded blob is not stale again:
`spirv-as --target-env vulkan1.1 prosper/tests/gpu/recompiler/rect_synthesis_vs.spvasm` reproduces
`kRectCornerVs` **byte-for-byte** (235 words), and the blob validates under `vulkan1.1`. The header
comment now names the exact target env, which is what makes that reproducible for the next person —
regenerating at `spirv-as`'s default would silently produce a 1.6 module again.

# (c) Can the culling assertion pass for a wrong reason?

I do not think so, and it is better designed than a one-directional check.

**The labels are right, and independently so.** I simulated the three arms from the VS's constants
rather than trusting the flags: with posA(−1,−1)/posB(1,−1)/posC(−1,1), the squared separations give
shared==A in all three arms (arm 0 at supplied index 0, arm 1 at 1, arm 2 at 2 — so the arm names are
accurate), all three synthesize (1,1), all three put the diagonal in the middle, and
`parity(emitted vs supplied) == 0` in all three. The `odd_permutation` flags `{false,true,false}`
equal `parity(supplied vs the VS's own order)`, computed by hand from the index buffer — i.e. from
the test's **input**, not from the code under test. Not circular.

**Both polarities are asserted, so the lever is provably armed.** Arms 0 and 2 assert `covered == 0`
and arm 1 asserts `covered == sampled` under the same cull state. If `cull_mode` silently did not
reach the pipeline, arms 0 and 2 would render in full and fail; if the culled pipeline could not
rasterize at all, arm 1 would fail. So neither "culling never applied" nor "nothing rendered" is a
passing state — the `covered == 0` arms, which on their own would be a weak negative, are exactly the
control that proves the cull is on. I confirmed the fixture honours it: `tests/fixtures/render_runner.h:7461`
assigns `rs.cullMode = ps->cull_mode` (modulo the `PROSPER_NO_CULL` escape, which would fail the test
loudly, not quietly).

**It catches the mutation it claims to, and more.** The named mutation ("shared first, then the other
two ascending") flips arm 1 only, which fails its `covered == sampled` — matching the reported
0/3600. A mutation that swapped the diagonal pair for *every* shared corner flips all three, failing
all three. A mutation that flipped only shared==v0 would be invisible with culling off (same
rectangle, same corner varyings) and is caught only here. And because a strip's two triangles share a
facing, "all or none, never half" is genuinely pinned: a half-covered result fails both branches.

The one thing worth writing down is that the mapping from parity to *which* polarity survives is an
empirical property of the fixture (positive-height viewport at `render_runner.h:7423`, default
`VK_FRONT_FACE_COUNTER_CLOCKWISE`), baked into a flag named after the index permutation. If the
fixture's viewport sign ever changes, all six assertions invert — loudly, which is the right failure
mode, but the reader will need this paragraph to know why.

# The other findings

- **B4 withdrawn properly**, in both places that matter: the code comment at `:4321-4323` and the PR
  body now say the non-synthesis module is *not* byte-identical, name both causes (ids renumbered by
  the earlier positions load; the `consumes_parameters` guard eliding a composite nothing consumed),
  and keep the conclusion that both are semantically inert. That is the honest form of the claim.
- **`CONFIDENCE: MED`** at `:4285-4293` names the sheared-parallelogram blind spot *and* the
  constant-`w` assumption behind measuring separation on the full 4-component clip position, and says
  what would be needed to do better. That is exactly what I asked for.
- **Allowlist** (`tools/vkval/allowlist.txt:85`) scopes #1715 to both tests without widening into
  fixing #1715 — right call, that is a separate defect.
- **Ruled-out row** now cites #3511.

---

# Non-blocking

**NB1 — please file the explicit-parameter guard as an issue before merging, and say so in the PR.**
You offered to add the arm now; I do not think it should hold this PR, because the table has two
independent derivations from the basis and the change to it was precisely the derived one. But it is
still the only part of this change with *no* executable check, its failure mode is silent by
construction, and "I will file it afterwards" is the step that evaporates. `spv_validate.cpp:821`
already builds a `PixelSystemInputMapping perspective_center` layout — the same layout with
`synthesize_rect=true`, asserting the P0/P10/P20-reconstructed value at the synthesized corner, is the
arm. A filed issue is enough for me.

**NB2 — the `spv_validate` corpus still only emits the `synthesize_rect=false` module.**
`tools/spv_validate/spv_validate.cpp:824` still calls `recompile_interpolation_geometry(layout)` with
the default argument, so the emitter gate remains blind to the rect variant and
`check_emitter_coverage()` is satisfied by the non-rect module. The defect class is genuinely covered
now — the vkval scan validates the rect module through `rect_synthesis_render`, and that is durable —
so this is redundant coverage rather than a hole in the shipped guarantee, and I am not blocking on
it. But that file's own header says a gap belongs there "only with the issue that tracks it, never as
a silent omission": one more `dump(...)` line, or an issue.

**NB3 — duplicated comment.** `:4531-4534` carries both the old two-line comment and the new one, so
"Value emitted in strip slot `slot`" appears twice and the first sentence still describes the
superseded framing. Delete the first two lines.

**NB4 —** unchanged from my first review and still cosmetic: `slot_of` emits fresh selects per
(output vertex × attribute) rather than hoisting four per-slot picks, and `consumes_parameters` is
recomputed inside the per-vertex loop though it is loop-invariant.

---

# Before merging

**The gate that caught B1 is green at this head.** I waited for it rather than taking the local run
on trust, since `ctest` alone cannot see this defect class: the `Linux` job's `Vulkan validation
scan` step is `completed success` on `68f98fb2`, where at `2e819ef5` it failed with
`VUID-VkShaderModuleCreateInfo-pCode-08737`. `pr_merge_gate.py 3511` reports `pass=9 fail=0` with
the recorded head still matching the branch tip.

Two checks (`macOS (x86_64 / Rosetta 2)`, `Windows MinGW`) were still pending when I wrote this, so
the gate correctly refuses. Re-run it immediately before merging — this approval is scoped to
`68f98fb2`, and a push after CI starts detaches both the checks and this review from head.

mattias800 (commented):
## APPROVED — on head `8c41f2aaf565c9dbbabe92aca4e3e2025435c0c6`

The previous approval carries over, and I checked that rather than assuming it. Both non-blocking
items are genuinely closed, and both are mutation-proven rather than asserted — the second one to a
number I can reproduce.

## The emitter is unchanged, mechanically

Yes, `prosper/src/` is comment text only, and I confirmed it rather than reading the hunk: stripping
`//` and `/* */` comments and blank lines from both revisions of
`rdna2_to_spirv_internal.hpp` (string- and char-literal aware) leaves the two **character-identical**,
217,366 bytes each. The single hunk merges the duplicated `slot_of` paragraph I flagged, at
`:4531-4533`.

So my derivations of the rotation (`supplied[(shared + slot) % 3]`, even in all three cases,
diagonal in the middle in all three), of the twelve `kI`/`kJ` pairs, and of the `bvec4` scoping
(allocated and declared only under `synthesize_rect`, four vec4 selects all taking the widened
condition) **carry over unchanged**. Explicitly agreed, as asked.

## `spv_validate.cpp` — the gap is closed

`geometry_interpolation_rect` is now dumped beside the plain form from the same layout with
`synthesize_rect=true` (`spv_validate.cpp:825-833`). Distinct output filename, and `dump()`'s
`exercised_emitters.insert()` is a set, so the shared emitter tag is idempotent and
`check_emitter_coverage()` is unaffected. The comment says why the plain form validated none of it,
which is the right thing for that file to carry.

The mutation you quote is the correct proof: the gate reproduces the identical validator diagnostic
that CI produced at `2e819ef5`, so this now catches the class **offline**, before it costs a CI
round-trip. That is a better outcome than the one I asked for, since I had only asked for the line or
an issue.

## The `kI`/`kJ` guard — verified end to end, and the mutation number falls out

This is the part I scrutinised hardest, because it is the one that replaces my own reading.

**Provenance is clean.** The 13-dword `explicit_ps` is byte-identical across
`test_rect_synthesis_render.cpp`, `test_interp_render.cpp` and `spv_validate.cpp` — I extracted and
compared all three rather than trusting the description, since a hand-retyped RDNA2 blob is exactly
the thing that goes subtly wrong. The `recompile_fragment(..., nullptr, &perspective_center,
UINT32_MAX, &explicit_layout)` call matches the declaration at `rdna2_to_spirv.hpp:497-504`
positionally (`rt`, `system_inputs`, `pcrel_dispatch_target`, `interpolation`).

**The new stage reconstructs the same rectangle, not a differently-calibrated one.** I modelled the
explicit path from the emitter's own definitions — `P0 = A0`, `P10 = A1−A0`, `P20 = A2−A0` stored flat
across all four emitted vertices, `(I,J)` at slot k from `kI[k][shared]`/`kJ[k][shared]`, and the PS
computing `P0 + P10·I + P20·J` — and evaluated it for all three arms:

| arm | supplied | shared | slot 0 / 1 / 2 / 3 reconstructed |
|---|---|---|---|
| `{0,1,2}` | A,B,C | A (idx 0) | 0.2 / 0.5 / 0.7 / 1.0 |
| `{1,0,2}` | B,A,C | A (idx 1) | 0.2 / 0.7 / 0.5 / 1.0 |
| `{1,2,0}` | B,C,A | A (idx 2) | 0.2 / 0.5 / 0.7 / 1.0 |

Every slot lands on the value of the vertex it carries, and the synthesized corner reconstructs 1.0
in all three. So the shared `corners[]` expectations apply unchanged to the new pair — it is a second,
independent reading of the same physical rectangle through the only code path that consults the
table. That is a genuinely stronger check than my hand-derivation, because it goes through the
consumer's formula rather than restating the basis.

**The mutation lands where you say, and the count is reproducible.** Reverting `kJ` to its
pre-rotation values changes only the `shared == v1` column, so only arm 1 is touched, and there it
swaps slots 1 and 2: slot 1 reconstructs 0.5 where it carries C (0.7) and slot 2 reconstructs 0.7
where it carries B (0.5), while slot 0 and the synthesized slot 3 stay correct. Two wrong corners,
and arm 1 is drawn under **both** cull states (`odd_permutation` is true, so `expect_drawn` holds
either way) — 2 × 2 = **4 failing checks**, exactly the four you measured, and none of them in the
smooth stage, which never reads the table. The arithmetic agreeing to the number is what convinces me
the arm is doing the work rather than passing alongside it.

**Check count cross-checks too**: 3 top-level + (2 stages × 2 culls × 3 arms = 12 iterations, of
which 4 are the culled-and-expected-empty ones contributing 1 check each, and 8 contribute
1 coverage + 4 corners) = 3 + 4 + 40 = **47**, matching your report. A loop restructured wrongly would
not land on that number.

## Housekeeping

`git log origin/main..HEAD --format=%B | grep -c '^Claude-Session:'` is 0, `git diff --check
origin/main...HEAD` is clean, and the commit is confined to the three files described.

## Before merging

**Both gates are green at this head, and I waited for them rather than taking the local run on
trust.** In the `Linux` job at `8c41f2aa`: `spv_validate` (test #264) **Passed**, so the new
`geometry_interpolation_rect` module is validated offline in CI, not just locally;
`rect_synthesis_render` (#408) **Passed** with the new explicit-parameter stage; `interp_render`
(#407) **Passed**, so sharing the `explicit_ps` fixture regressed nothing; 450 tests, 0 failed; and
the `Vulkan validation scan` step reports `[vkval] PASS: every observed validation message ID is
accounted for.`

Re-run `python3 prosper/tools/ci/pr_merge_gate.py 3511` immediately before merging and merge only on
a full pass — some jobs were still in flight when I wrote this. This approval is scoped to
`8c41f2aa`; a further push detaches both it and the checks.

Issue #3496, opened 2026-09-09
[Game tracker] BlazBlue Entropy Effect X (PPSA29714)


mattias800:
## Current state — **rung 1** (any graphics at all)

First run of this title, 2026-09-09, on `6771a2c0d` (built from `origin/main`, `-DPROSPER_APP=ON`,
RelWithDebInfo).

The guest boots, links, reaches its frame loop and renders **its two publisher logo cards —
"ASTROLABE GAMES" and "91 Act" — at 1920x1080 on a black field**. It then holds there and never
advances: no engine splash, no title screen, no menu. Watched for ~45 s / ~8,000 presented frames.

Rung 1 rather than 0 because those are real frames from the live renderer, and rung 1 rather than 2
because a publisher logo is not a title screen.

## Dump

| | |
| --- | --- |
| title id | `PPSA29714` |
| version | `02.000.013` |
| engine | **Unity / IL2CPP** (`Media/Modules/Il2cppUserAssemblies.prx`, `Media/Resources/unity default resources`) |
| size | 15 GB |
| middleware | CRI (`cri_ware_unity.prx`, `cri_lips_unity.prx`), PuerTS, LZ4, Burst, curl |

## Reproduction

```
prosper-app --dump <DUMP_ROOT>/PPSA29714-app0 --fps
```

Nothing else — no environment variables, no throttle, no pad route. Default launch, audio and input
on, KWin/Wayland, RADV STRIX_HALO.

## What works

- Links **12 modules, 4321 imports** (3017 cross-module, 922 stub slots), 11 init fns, 0 aliased exports.
- Reaches `guest booted; presenting its frames` and sets its own window title.
- Live Vulkan submit renderer, compute backend, descriptor indexing, int64 atomics and the native
  subgroup contract all come up on the adopted device.
- Audio backends install cleanly (FFmpeg AJM decoder + SDL3 out, volume 100%).
- Renders **1920x1080** at 180 fps presented / ~97 distinct, ~65% active. Throughput is not a
  concern on this title yet.
- Five absent plugins are skipped without incident (`AkSoundEngine`, `AkVorbisHwAccelerator`,
  `AkMotion`, `libfmod`, `libfmodstudio`) and one duplicate export is correctly de-duplicated
  (`cri_lips_unity` vs `cri_ware_unity`, NID `SLPJVeuL2v4`).

## Leads for whoever picks this up

None of these is established as the cause; they are what the default run reports, in the order I
would check them.

[...]

Issue #1963, opened 2026-08-05
sceKernelCreateEventFlag writes through an unvalidated guest out-pointer, so a bad pointer becomes an unattributed host SIGSEGV


mattias800:
## Summary

`prosper/src/hle/hle_kernel.cpp:1767` (`k_ef_create`, `sceKernelCreateEventFlag`) stores the new
event-flag object straight through the guest's out-pointer:

```cpp
HLE(k_ef_create) {   // (ef*, name, attr, initPattern, opt)
    auto* e = (EventFlag*)calloc(1, sizeof(EventFlag));
    pthread_mutex_init(&e->m, nullptr); pthread_cond_init(&e->c, nullptr); e->bits = a3;
    if (a0) *(void**)(uintptr_t)a0 = e;      // <-- unguarded store
    return 0;
}
```

The `if (a0)` guard only rejects exactly zero. Any other bad pointer — a null base plus a field
offset is the common one — faults inside prosper's own code.

## Concrete failure scenario

*Little Nightmares III* (`PPSA05143`, UE4) calls
`sceKernelCreateEventFlag(ef = 0x80, "AudioThread", attr = 0x20, init = 0, opt = NULL)` because its
audio-thread singleton is null and the event flag lives at `this + 0x80`. `0x80` passes the guard, and
prosper writes to address `0x80`.

What the developer sees is:

```
[nullpage] BACKING DENIED addr=0x80 rip=eboot+0xfffffffbf067c95a err=-1
[shot] guest thread ended: kind=2 detail=SIGSEGV at addr=0x80  rip=0x67c95a (image+0x0) rbp=0x80
  backtrace (0 frames):
```

That is actively misleading in three ways: the `rip=eboot+0x…` label is instrument-trap 22 (the rip is
below the guest image base, so it is host code, not eboot code); the rbp-based guest walker returns
**zero** frames because `rbp` is the clobbered out-pointer; and nothing names the API involved. It took
`addr2line` on the host binary to learn that the fault was in `sceKernelCreateEventFlag` at all.

## Suggested fix

Use a guarded store, the way the AJM/audio HLE already does with its `a2_store_u32` helper: verify the
destination is guest-writable, and on failure log the API name plus the rejected pointer and return the
Sony error (`SCE_KERNEL_ERROR_EINVAL`) instead of faulting. That keeps the failure fail-visible — which
is the point — while naming it. The same pattern applies to every value-returning HLE that writes
[...]

Issue #1234, opened 2026-07-23
Guest open of '/app0/..' (the sandbox root) is denied; real PS5 resolves it


mattias800:
Observed during an ArcRunner (PPSA21406) headless boot on native Linux (while verifying PR #1233), with `PROSPER_FILELOG=1`:

```
[file] open-result host='/prosper-denied/app0/..' guest-flags=0x20000 host-flags=0x10000 -> fd=-1 error=2
```

Early in boot the guest opens the literal path `/app0/..` (flags suggest a directory open). `translate()`'s #1205 sandbox normalization pops above the `/app0` mount root and denies it (`src/hle/hle_file.cpp`, sandbox_normalize_subpath returning nullopt → `/prosper-denied`), so the guest gets ENOENT.

On real hardware the title is jailed and `/app0/..` is the (readable) sandbox root directory, so this open would succeed. ArcRunner tolerates the failure (boot continues), so this is low priority — but a title that enumerates its sandbox root, or stats `/app0/..` to derive a base path (UE4 does BaseDir arithmetic — cf. the `file://../../..` media URLs from #1226), could misbehave.

Suggested direction: model the sandbox root as a first-class virtual directory — `/app0/..` (and `/`) resolve to a host directory containing the mounts prosper knows (`app0`, `temp0`, `savedata0`), or at minimum stat/opendir of the exact string `/app0/..` succeeds against the dump root's parent-as-virtual-root. Keep #1205 denial for anything that traverses *through* the root to a named sibling (e.g. `/app0/../etc`).

Pre-existing behavior; not introduced by #1233 (verified: the same denial appears with the fix in place, and #1233 only changes RELATIVE spellings — absolute `/app0/..` keeps the deny path).

mattias800:
CLAIMING: claude-session-d2181233 (bp-res worktree) | branch fix/issue-1234-virtual-root | 2026-07-24T15:28:59Z

PR #3512, opened 2026-09-10
fs+libkernel: map guest /download0 and register sceKernelCancelEventFlag — BlazBlue Entropy Effect X reaches its title screen

mattias800:
## What this fixes

*BlazBlue Entropy Effect X* (`PPSA29714`) booted, linked and rendered its two publisher logo cards,
then held there forever — no engine splash, no title screen, no menu (#3496, rung 1). It now reaches
its **title screen and menu**.

BlazBlue Entropy Effect X title screen

*`tools/screenshot`, default launch with `PROSPER_PAD_PRESS=1`, 1920x1080. The fps counter is
prosper's own overlay.*

### 1. `/download0` was never mapped (this is the one that unblocked the title)

The PS5 gives a game four mounts. prosper mapped `/app0`, `/temp0` and `/savedata0`, and passed
`/download0` through **unmapped** — `mount_path_matches` never matched it, so resolution fell to the
final `return p` and the guest got a literal host `/download0/...` that does not exist. Every access
failed.

Located by differential, not by guessing: this title opens `/download0` **10 times** during startup;
*The Messenger* (`PPSA24651`, a Unity title that reaches gameplay) opens it **0 times**.

It is served **empty**, which is the honest answer rather than a placeholder — an empty download-data
area means "no downloaded content is present", which is true of every local dump. It provides the
mount point, never content, and add-content ownership questions continue to be answered from local
inventory in `hle_addcontent.cpp`, untouched by this. The root follows `temp0_root()`'s existing
convention exactly (`/tmp/prosper-download0`, override `PROSPER_DOWNLOAD0`).

### 2. `sceKernelCancelEventFlag` was unregistered

The last member of the event-flag family prosper did not register, so it took the dispatcher's
`return 0` == `SCE_OK` while doing nothing: no waiter released, bits left stale, and
`numWaitThreads` **never written** — the guest read its own uninitialised stack as the count of
threads it had just cancelled. That is #2081's "false success" shape.

Also found by differential: normalising a full boot log of this title against *The Messenger* leaves
exactly one prosper-level line that differs — this NID.

A `cancel_gen` generation counter, not a flag, so a thread that parks *after* a cancel is not
spuriously released by it, and nothing busy-waits.

**What the earlier census did not establish**, recorded because it is the trap: it measured
`waiters` at each call and found **0 of 3000**, and I read that as ruling the call out. It rules out
only *one* of the call's three effects — it says nothing about the unwritten out-parameter or the
bits, which are the other two. Testing beside the gap rather than on it.

### 3. A cancelled waiter reported success

Found while reviewing the above rather than reported: a waiter released by the new cancel left the
loop with `ret == 0`, so the guest would read the flag bits as a pattern that **matched** when
nothing had matched, and the `CLEAR_ALL`/`CLEAR_PAT` step would then clear the very bits the cancel
had just set for other threads. It now reports `ECANCELED`, which also suppresses that step and
overrides the timed arm's `ETIMEDOUT` (derived from "no match", so it cannot tell a cancel from an
expiry).

## Evidence, and what rests on what

The PR text above is measured. The parts of change 3 that are **not** measured are marked in the
code, because they differ in strength:

| claim | evidence |
| --- | --- |
| argument layout `(ef, setPattern, int* numWaitThreads)` | live capture on `PPSA29714` |
| the out-parameter must be written | the defect itself; `SCE_OK` with an untouched stack slot |
| `ECANCELED` == `0x80020055` | **derived twice independently** — prosper's own `0x80020000 \| FreeBsd ECanceled(85)` (`sce_errno.hpp:112`), and a secondary implementation reporting the same constant |
| pattern is assigned, not OR'd | **now TESTED.** Was a contract claim only; the test create/cancel with 0x80/0x40 so assign (0x40) and OR (0xc0) differ, and OR-ing reddens it. Earlier both patterns were 0, so the test could not tell them apart.<br>Originally cited the title's progress, which was wrong: An earlier revision cited the title's progress here and that was wrong: #3496 records "measured after: unchanged" for this call, and `/download0` is what unblocks it. With every observed setPattern == 0 and no waiter parked, assign and OR are indistinguishable locally |
| cancelled waiter still receives the result pattern; performs no CLEAR | **secondary implementation only — the weakest claims here.** No local title parks a waiter, so nothing exercises them |

The secondary implementation is *Kyty*, the same source this file's existing `ETIMEDOUT` and
`EACCES` notes already cite. It is a work in progress like this one, so it is used as a cross-check
and never as an authority: nothing here is copied, the constant was derived from prosper's own table
first, and prosper's structure is different by design (generation counter vs. a status enum with
sleep loops). Where it is the *only* support, the code says so and asks for re-derivation against a
title that actually parks a waiter.

## Review outcome

Independently reviewed; **rejected first time, three blocking findings, all fair and all fixed.**

1. **A null handle returned EINVAL** while all nine siblings (`ef` delete/set/clear/wait/poll and the
   four semaphore entry points) return 0 — and the comment claimed that was "inherited from the
   siblings above", which was inherited from nothing. It made this the one call in the family that
   reports failure, on no evidence. Now matches the family; the reasoning for preferring consistency
   over a third candidate (a secondary implementation returns ESRCH) is in the code.
2. **A false evidence claim**, corrected in the table above — contradicted by this project's own
   tracker.
3. **No test coverage**, while `test_sync_delete.cpp` already parks a real thread on an infinite
   event-flag wait. "No local title parks a waiter" is a fact about the corpus, not about what a test
   can construct.

Also fixed, non-blocking but real: the out-pointer is now validated with `gpu::guest_writable` before
the store, matching `sceKernelCreateEventFlag`'s #1963 hardening (without failing the call — the
cancel has already happened); and `download0_root()` is a magic static rather than a copy of
`temp0_root()`'s older racy lazy init.

The reviewer also judged that the ECANCELED change **belongs** in this PR rather than being split
out: change 2 creates the path, so omitting it would have introduced the false-success defect rather
than preserved anything.

## Tests added

`test_sync_delete.cpp` — parks a real thread, then cancels: the out-parameter carries the parked
count (poisoned to `-12345` first, so "untouched" and "written as 0" cannot be confused — the defect
was reading stack residue), the waiter wakes with `ECANCELED` not success, it still receives the
result pattern, a thread parking *after* the cancel is **not** released by it, and a null handle
answers like the family.

`test_translate_sandbox.cpp` — the `/download0` guards its three sibling mounts already had: the
component boundary (`/download00` is not `/download0`), that it is mapped rather than passed through,
that the subpath survives translation, and that traversal out of it is denied.

**Mutation-proven, one arm each:**

| mutation | result |
| --- | --- |
| drop the out-parameter write (the original defect) | `[FAIL] wrote the parked-waiter count` |
| cancelled waiter returns 0 again | `[FAIL] returned ECANCELED, not success` |
| `cancel_gen` made sticky | `[FAIL] a waiter parking after the cancel is NOT released by it` |
| cancel ORs the set-pattern instead of assigning it | `[FAIL] the cancel ASSIGNS its set-pattern` |
| null path leaves the out-parameter untouched | `[FAIL] still writes a defined count, not stack residue` |
| `/download0` dropped from `reenters_mount` | `[FAIL] a traversal re-entering /download0 is mapped` |
| remove the `/download0` resolution arm | 2 of the 4 new sandbox assertions fail |

## Verification

```
cmake --build prosper/build-linux -j6          # clean, 0 errors
ctest --no-tests=error -j4                     # 461/461 passed, exit 0
```

Live: `PPSA29714` reaches the title screen and menu, 10/10 screenshots, 9 pixel-distinct, ~57 fps.
Before this branch the same route holds on the publisher logos indefinitely (#3496's opening
measurement, ~45 s / ~8,000 presented frames).

## Risk

`/download0` is a new mount, so any title that opens it now gets an empty directory instead of a
failed open. That is strictly closer to hardware, but it is a behaviour change for every title, not
just this one — a guest that branched on "download0 is unavailable" will now take the other path.
No local title is known to do so, and the three rung-6 guarded titles are unaffected below.

Refs #3496


mattias800:
## Guarded-title regression evidence

The PR body claimed the rung-6 guarded titles are unaffected; here is the measurement rather than the claim. Same route and cadence as the baseline captures taken on current `main` (`tools/screenshot`, default launch with `PROSPER_PAD_PRESS=1`, `--seconds 4 --count 8`), compared per frame on a 16x9 luminance signature:

| title | rung | result |
| --- | --- | --- |
| *The Messenger* `PPSA24651` | 6 | **pixel-identical**, 0.00/255 on all 8 frames |
| *Blasphemous 2* `PPSA13579` | 6 | worst frame **0.25/255** (animated title screen; 5 of 8 frames exactly 0.00) |
| *Dead Cells* `PPSA15552` | 6 | title screen renders correctly — logo, key art, Play/Options/Save menu; inspected, not just scored |

*Dead Cells* had no same-route baseline in this session, so it is reported standalone and checked by eye rather than by diff.

That covers the `/download0` risk in practice for the titles that matter most: none of the three opens it, so none changes behaviour. It does **not** cover the general risk stated in the PR body — a title that branches on "download0 is unavailable" would now take the other path — which remains reasoned rather than measured across the whole corpus.

Verification at this head: build clean, `ctest --no-tests=error -j4` → **461/461 passed, exit 0**.

mattias800 (commented):
# Independent review — REJECTED (three blocking findings, all cheap; no code defect found)

Reviewed at head `ee3b9618`, read-only from a checkout of the branch. I did not build, run ctest, or
re-run any of the author's live captures — this is contract/assumption review, not duplicated
verification.

**Bottom line:** the code is, as far as I can determine by reading it, correct in all three changes,
including the ECANCELED path the author flags as weakest. What I am blocking on is two claims in the
project record that are not supported by the evidence they cite, and the absence of any test at all
for either new path when the fixture that would exercise them deterministically already exists in
tree. None of the three needs a redesign.

---

## What I verified, concretely

**A. `/download0` and the anti-circumvention rule — clean, in both directions.**

Serving it empty does not cross the line, and it does not under-report. `canonical_child_directory`
(`prosper/src/hle/service/hle_addcontent.cpp:454-456`) requires every declared add-content mount
point to begin `/app0/` and to canonicalise to a real child of the dump root, so nothing add-content
serves can live under `/download0` at all —
`prosper/tests/hle/dispatch/test_service_getters.cpp:894-898` asserts that a `mount_point=/download0/dlc1`
declaration invalidates the entire inventory. Grepping the tree, `download0` now appears in exactly
two places: `hle_file.cpp` and that one negative test. Nothing downstream reads its presence or its
contents as an ownership or entitlement signal.

The positive half is satisfied for the same reason: in prosper's model installed add-content lives
under `/app0`, and `/download0` on hardware is the application's own download-data scratch area, not
the DLC mount. There is no local dump layout in which content that IS present would have to surface
here. Empty-and-writable is the honest answer, and `sceAppContentDownloadDataGetAvailableSpaceKb`
(`hle_service.cpp:5547`) already reports space for exactly this area, so the mount and the space
answer are now consistent where before they were not.

A new writable mount also opens no route around the module policy: `module_path_policy.hpp`'s
allowlist is dump-relative (`kPermittedModuleDirs` / `kPermittedRootFiles`), and `download0_root()`
points outside the dump root. Same exposure as the pre-existing `/temp0`, which is to say none.

**Path resolution — the if/else chain cannot misroute, and the sandbox is intact.**

`mount_path_matches` (`hle_file.cpp:735-739`) requires a component boundary, so `/download00/x` does
not match. `/download0` and `/savedata0` are distinct 10-character literals, so the two arms are
disjoint by construction; and the `!save0.empty()` arm at `:785` is reachable only when
`mount_path_matches(p, "/savedata0")` held at `:781`, so the new arm placed above it at `:784` cannot
steal a savedata path and a savedata path cannot reach the download arm. `vlen = 10` is
`strlen("/download0")`. The `reenters_mount` addition at `:750` is right: `/download0/../etc`
normalises to `/etc`, matches no mount, and keeps the #1205 deny; `/download0/..` normalises to `/`
and returns the vroot, whose entry list now includes `download0` (`:594`, `:600`), created on every
process start so an existing vroot from an older build gains the entry.

**B. Event-flag concurrency — correct, including the initialiser.**

`cancel_gen` IS initialised: `k_ef_create` allocates with `calloc(1, sizeof(EventFlag))`
(`hle_kernel.cpp:2908`) and that is the only allocation site in the tree. The positional-braced-init
hazard does not apply here — I checked for it specifically rather than grepping the field name.

Both loops capture `entry_cancel_gen` under `e->m`, immediately after `waiters++`
(`hle_kernel.cpp:3461-3462`), so no cancel can slip between the lock and the capture, and a thread
that parks after a cancel captures the new value and is not spuriously released. `parked` is sampled
at the right moment: between `waiters++` and `--waiters` a waiter holds `e->m` except while inside
`cond_wait`, and the canceller holds `e->m` while reading, so everything counted is genuinely blocked
on the condvar — the count is not inflated by threads that matched immediately. (Noting this so
nobody later "fixes" it in the wrong direction.)

No new deadlock or UAF against the `deleted` / last-waiter-frees logic: cancel takes no reference and
never touches `waiters`, `deleted` wins over `cancelled` at `:3511-3512`, and the free at `:3514`
is unchanged. A cancel racing a pending delete still yields EACCES. `int32_t` is the right width for
`int* pNumWaitThreads`.

**C. The ECANCELED path — mechanically correct, and it belongs in this PR.**

`0x80020055` verified from prosper's own table: `ECanceled = 85` (`sce_errno.hpp:112`) through
`sce_kernel_error()` at `:125-127` gives `0x80020000 | 0x55`. Independently, Kyty's `errno.h:274`
states the constant directly. All three mechanical claims hold: setting `ret` does suppress the
CLEAR step (`if (!ret)` at `:3513`), it does override the timed arm's ETIMEDOUT (the `else if
(cancelled)` at `:3512` assigns unconditionally over the `:3477` assignment), and `*a3` is still
written at `:3517` from the `res` captured at `:3513` before any free.

On whether the author should have made this unmeasured change here: **yes, and I would have objected
if it had been split out.** Change 2 is what makes this path reachable in the first place. Registering
Cancel while leaving a cancelled waiter returning 0 would not preserve an existing behaviour — it
would newly create the "false success" defect the PR is about, in the same commit that creates the
caller. The right criticism is not that it was made, it is that it shipped with no evidence at all
(blocking finding 3).

**D. No copying from the secondary implementation.** prosper's structure is genuinely different:
Kyty uses a `Status` enum plus 10 µs unlock/sleep/relock spin loops and blocks the canceller until
every waiter has left (`eventFlag.cpp:210-234`); prosper uses a monotonic generation and blocks
nobody. The comments are prosper's own prose. Kyty does in fact support the two claims attributed to
it — `eventFlag.cpp:148-155` writes `*result = m_bits` and returns before the clear switch at
`:168-172` on the Canceled path, and Cancel assigns `m_bits = bits` at `:226` — so those attributions
are accurate. Treating it as a cross-check rather than an authority is the right call and the code
says so.

**E.** The risk statement is correct as far as it goes, and the guarded-title measurement in the
2026-09-10T07:28 PR comment is a good addition. See non-blocking (e).

Housekeeping: 4 files, no `.md` deletions, no `Claude-Session:` trailers, no attribution footers,
screenshot is a 1920x1080 WebP at 71 KiB (meets the shrink rule), BLOG entry present and short.

---

## Blocking findings

### 1. `hle_kernel.cpp:2945` cites an inheritance that does not exist, and the value diverges from the whole family

The comment reads: *"CONFIDENCE: MED on the error code for a null handle, inherited from the siblings
above."* No sibling above returns EINVAL for a null handle. Every one of them returns 0:

- `k_ef_delete` `hle_kernel.cpp:2915`, `k_ef_set` `:2924`, `k_ef_clear` `:2925`,
  `k_ef_wait` `:3455`, `k_ef_poll` `:3524`
- and the semaphore family too: `k_sema_delete` `:3558`, `k_sema_wait` `:3571`,
  `k_sema_signal` `:3604`, `k_sema_poll` `:3666`

The two EINVAL sites the comment might have meant are not this family: `k_mutexattr_gettype:327` is
an attribute getter and `k_sem_timedwait:3235` is the POSIX `sem_t` path via `ensure_sem`. And the
only secondary reference returns ESRCH, not EINVAL (`Kyty eventFlag.cpp:367-373`).

So `:2948` is not inherited from anything, and it is a real behaviour divergence rather than a
cosmetic one: `sceKernelCancelEventFlag` is now the only member of the event-flag family that reports
failure on a null handle, on no evidence. This is the citation-accuracy class CLAUDE.md flags as the
most contagious error the project produces, and it is load-bearing here because the raw hex means the
comment is the only thing a reader has.

Fix: correct the comment, and choose deliberately — `0` for family consistency, `ESRCH` if you follow
the reference (and say that is where it came from), or keep EINVAL and state plainly that it is an
unevidenced deliberate divergence from both the siblings and the reference.

### 2. `hle_kernel.cpp:3506` and the PR body's evidence table claim a result the author's own tracker contradicts

The comment ends: *"...and assigning is what carries PPSA29714 forward."* The PR body's evidence table
carries the same sentence, in the **evidence** column of the `pattern is assigned, not OR'd` row.

That is contradicted by the author's own record. Tracker #3496, comment 2026-09-09T20:22:46Z, ruled-out
row 8: *"`sceKernelCancelEventFlag`'s out-parameter/bits | implemented correctly; unchanged"* — and
this PR's own §1 says `/download0` *"is the one that unblocked the title"*. With `setPattern == 0` on
every observed call and zero parked waiters across 3000 calls, assign-vs-OR is not observable in this
title's outcome at all; no A/B is reported, and by the author's own measurement none could have
discriminated.

The sentence is defensible as a statement about the *contract* (assign is what the reference and the
PSP-lineage semantics specify, and it is what the code now does on this title's route). It is not
defensible as *evidence*, which is the column it sits in — in the one table whose entire purpose is
honest attribution. This is the failure mode that table exists to prevent, so it has to be right.

Fix: reword both the comment and the table row to what is true — the reference plus the contract
support assign; the title's outcome does not discriminate — or produce the A/B that does.

### 3. Both new paths ship with zero test coverage, and the fixture for the harder one already exists

`ctest 461/461` is a true statement about a suite that does not execute one line of this diff. That is
acceptable for the `/download0` change, which has a live capture behind it. It is not acceptable for
the ECANCELED path, which has nothing.

The PR says *"no local title parks a waiter, so nothing exercises them"* and asks for re-derivation
*"against a title that actually parks a waiter."* But the constraint is the title corpus, not what a
test can reach — and `prosper/tests/hle/kernel/test_sync_delete.cpp:148-172` **already parks a real
thread on an infinite `sceKernelWaitEventFlag`, asserts that it is parked, then deletes and asserts
EACCES.** The identical shape with `sceKernelCancelEventFlag` substituted for the delete exercises,
deterministically and without any title, every claim the PR marks as weakest, in one test:

- `*numWaitThreads == 1` (the out-parameter, which no title will ever make non-zero here)
- the waiter returns `0x80020055` rather than `0`
- `*resultPat` is still delivered, and carries the cancel's set-pattern
- the CLEAR is suppressed — run the wait with `CLEAR_ALL` in the mode and assert the bits survive
- and, with the timed arm and a short timeout, that ECANCELED overrides ETIMEDOUT

This is CLAUDE.md's rule for exactly this situation: before believing a clean zero, construct one
positive instance of the class **by hand, outside whatever produced the null**. The 0-of-3000 census
is the null; `test_sync_delete.cpp` is the hand-built instance, and it is twenty lines away.

The `/download0` half is cheaper still: `prosper/tests/hle/test_translate_sandbox.cpp:58-66` already
carries sibling-prefix guards for `/app0evil`, `/temp00` and `/savedata00` and has none for
`/download00`, and the mapping plus the `/download0/../etc` deny are two more lines beside them.

---

## Non-blocking observations

**(a) `hle_file.cpp:562-574` — `download0_root()` copies a racy lazy static.** `static std::string
root; if (root.empty()) { ... }` is not thread-safe, and `translate()` runs on any guest thread doing
file I/O — two first-callers can both see it empty and both assign, which is a data race on
`std::string`. It is copied faithfully from `temp0_root():536-548`, so it is not a new kind of
problem; but `virtual_root_dir():583` in the same file uses a magic static and its comment says
exactly why (*"thread-safe init: titles do file I/O from several threads"*). One-line fix
(`static const std::string root = []{...}();`), and the pre-existing `/temp0` instance deserves the
same treatment while you are there.

**(b) `hle_kernel.cpp:2957` writes through the guest pointer with only a non-null test.**
`k_ef_create:2900` refuses precisely this shape with `gpu::guest_writable`, and its #1963 comment
explains that `if (a0)` lets through the case that actually occurs in the wild — a null object plus a
field offset — and that the resulting fault reports itself misleadingly. Deferred to non-blocking
because it is consistent with `k_ef_wait`'s unvalidated `a3`/`a4` writes at `:3477`/`:3517`, so it is
not a regression in kind. But the hardened sibling is fifty lines up and the guard is already
available in this file.

**(c) Raw hex where named constants exist and are used in the same file.** `0x80020016` vs
`prosper::hle::kSceKernelErrorEINVAL` (used at `:2992`, `:3015`, `:3429`, `:3430`) and `0x80020055` vs
`sce_kernel_error(FreeBsdErrno::ECanceled)`. Using the named forms would make blocking finding 1
impossible to write, which is the argument for them.

**(d) "derived twice independently" is a shade strong.** Both routes are the same FreeBSD-errno-plus-
libkernel-prefix encoding, so it is one derivation confirmed by two independent *statements* of the
number (Kyty hardcodes `0x80020055`). Still a genuine cross-check — worth phrasing as that rather
than as two independent derivations.

**(e) The PR body's Risk section says the guarded titles "are unaffected below" and there is no
"below".** The measurement is in the 2026-09-10T07:28 comment, and it is good (Messenger
pixel-identical, Blasphemous 2 worst frame 0.25/255, Dead Cells inspected). Point the body at it, and
keep the comment's own honest caveat that the corpus-wide "a guest that branched on download0 being
unavailable" risk remains reasoned rather than measured.

**(f) Tracker #3496 is stale against this PR.** Its body still says rung 1 and its ruled-out table
rows 2 and 8 still read Cancel as ruled out and "unchanged" — which is what blocking finding 2 trips
over. `Refs #3496` is the correct linkage; the merge should carry a comment recording `/download0` as
the actual cause, the rung-1 → rung-2 move, and the title-screen capture in the tracker itself.

**(g) `hle_kernel.cpp:2951` (`e->bits = a1`) is the largest blast radius in change 2, more than the
out-parameter.** Before this PR a cancel left the bits untouched; now every cancel with the observed
`setPattern == 0` wipes the flag for every other thread. I believe it is correct — it matches the
reference and the PSP-lineage contract — but it is the part to watch as other titles start calling
this, and the PR body leads with the out-parameter instead.

**(h) Cancel on a handle already freed by Delete is a UAF**, shared with `k_ef_set`/`k_ef_clear`.
Pre-existing, not introduced here, and not worth fixing in this PR — noted so it is not rediscovered
as a regression.

**(i) CI at the time of this review:** 7 passed, 4 pending (Linux, Windows MinGW, Windows App, macOS
Rosetta). PR head `ee3b9618` matches the branch tip.

---

**REJECTED** — on findings 1, 2 and 3. No code defect was found in any of the three changes; two of
the three blockers are corrections to text that currently asserts more than the evidence supports,
and the third is a test that the existing fixture makes cheap. Re-review after those should be quick,
and I would expect to approve.

mattias800 (commented):
# Re-review at `3ed1859d2` — APPROVED

All three blocking findings are fixed, and fixed for the stated reason rather than papered over. I
re-derived each against the code rather than reading the summary. Head `3ed1859d` matches the branch
tip. Below: confirmation, then direct answers to (a), (b) and (c), then five non-blocking items —
two of which I think are worth the five minutes before you merge, and I say which.

---

## The three blockers — confirmed fixed

**B1.** `hle_kernel.cpp:2948` now returns 0, and the comment at `:2941-2947` states the tiebreak
instead of inventing a lineage. I checked the nine siblings again at this head; all still return 0.
**I accept your choice over my ESRCH datapoint** — see (a) for the one refinement I would still make.

**B2.** `hle_kernel.cpp:3516-3520` now says the two are indistinguishable here and that assigning is
a contract claim, not a measured one, and it names #3496's "measured after: unchanged" as the thing
that killed the earlier wording. That is the correct repair: it does not retreat from assigning, it
retreats from claiming the title's progress supports it. The PR table row matches.

**B3.** `test_sync_delete.cpp:178-235` is the fixture I pointed at, used properly. Three things I
want to single out because they are what make it evidence rather than decoration:

- the `-12345` poison. "Untouched" vs "written as 0" is the entire defect, and without the poison the
  assertion would have been satisfied by the bug. This is the strongest single line in the diff.
- `num_wait_threads == 1` does double duty: it pins the out-parameter *and* proves the waiter was
  genuinely parked in the condvar at the instant of the cancel, which is what makes the preceding
  `!done.load()` non-vacuous. That is a self-validating check in the sense the charter means.
- `observed` poisoned to `~0ull` before the wait, so "the pattern was delivered" cannot be satisfied
  by an untouched variable.

`test_translate_sandbox.cpp:62-79` adds the component-boundary guard its three siblings had, and I
confirmed by hand that removing the `/download0` arm from `translate()` reddens exactly two of the
four (equality, and the `rfind` guard) — your mutation count is right.

**Non-blocking (a) and (b) from the last round are also fixed correctly.** `download0_root()` is now
a magic static (`hle_file.cpp:562-576`) and says why it does not inherit `temp0_root()`'s race. The
`guest_writable` gate is the right guard — see (c).

---

## Your three questions

### (a) Is returning 0 for a null handle the right tiebreak? — Yes, with one refinement

Your reasoning holds and I withdraw the ESRCH preference: with no local evidence for any of the three
answers, one divergent member is worse than a wrong convention, because a convention can be corrected
family-wide in one change and a divergence has to be rediscovered. Being the only call in the family
that reports failure was the actual defect, not the specific code.

But there is a loose end the new test now pins in place. On the null path you return **0 (success)**
and leave the out-parameter **untouched** — `test_sync_delete.cpp:231-233` asserts exactly that
(`untouched == -999`). That pair is the false-success-with-an-unwritten-out-parameter shape this PR
exists to remove, reached by a different route: a guest that checks the return sees SCE_OK and then
reads its own stack residue as the count.

The fix keeps your tiebreak intact — **write 0 through the (validated) out-pointer on the null path
too.** "Success, zero threads cancelled" is internally coherent, leaves no residue, and changes
nothing about the return convention you argued for. Then the test asserts `== 0` rather than
`== -999`, and the poison still distinguishes it from untouched.

Not blocking: no title passes a null handle, and the current behaviour is no worse than the family's.
But it is the one place where the PR's own principle and the family convention actually conflict, and
you can satisfy both.

### (b) Could the tests pass for a wrong reason — is the 40 ms what makes the t2 arm pass? — Partly yes, and there is a one-line fix

You were right to be suspicious of that arm specifically. Working through it:

The arm **is** a real discriminator. Under a sticky-flag implementation t2 falls straight through the
loop (`!evf_match(0, 0x1, OR)` true, `!deleted` true, sticky-cancelled false) and returns in
microseconds, so `!done2.load()` after 40 ms reddens. 40 ms is ~1000x the margin needed.

The weakness is not a flake, it is the direction of the failure: **if t2 has not been scheduled yet,
`done2` is false and the check passes vacuously.** Under load — and this box runs concurrent ctest
with coredump handlers, per the charter's own note — that is a silent pass, not a failure. It is an
experiment that cannot detect its own invalidity, which is the one property the charter asks for.

**The fix is one line: capture t2's return value and assert it is 0.** You currently discard it
(`ef_wait(...)` at `:216` with `a3 = 0` as well, so no pattern is captured either). Assert `wret2 == 0`
and the arm stops depending on scheduling entirely, because it discriminates under *both* orders:

- t2 parks, then `ef_set` matches → correct impl returns **0**; sticky impl already returned **ECANCELED**.
- t2 is slow and only enters after `ef_set` → correct impl matches immediately, `entry_cancel_gen`
  was captured after the cancel so `cancelled` is false → **0**; sticky impl still sees the sticky
  flag and returns **ECANCELED**.

Either way the wrong implementation is caught and the right one passes. That is strictly stronger
than the timing window, and it costs an `std::atomic<uint64_t> wret2`.

**One thing I could not reconcile, and I would like you to double-check rather than take my word.**
You report the sticky mutation reddening `[FAIL] parked before cancel`. That is the assertion at
`:194`, which fires *before* any cancel has happened — I cannot construct a sticky-`cancel_gen`
mutation that reddens it. Every sticky variant I worked through leaves the first waiter parking
normally and reddens `"a waiter parking after the cancel is NOT released by it"` at `:224` instead.
Either the message is a transcription slip, or the mutation you ran was not the one that binds this
arm — and "which arm actually bound" is a trap you have hit before. Worth re-running with the arm
named; if `:224` is the one that reddens, the arm is proven and my only point is the vacuity fix above.

**The other arms I checked for wrong-reason passes and they are sound:** `crc == 0`,
`num_wait_threads == 1` (self-validating, as above), `wret == kECANCELED` (dies on `ret = 0`),
`observed == 0` (dies if `*a3` is not written). See non-blocking (2) for a way to make the last one
say more than it currently does.

### (c) Is silence right in the `guest_writable` gate? — The return value is right; the silence is not

Two separate calls, and you got the harder one right.

**Not failing the call is correct**, and for exactly the reason you give: the cancel has already
taken effect on the flag and released its waiters, so an error return would tell the guest the cancel
did not happen. That would be a worse lie than the one being prevented. Agreed, and I would have
objected to EFAULT here.

**But silence is wrong.** When the store is skipped the guest is told SCE_OK and reads stack residue
as the waiter count — which is, precisely and exactly, the defect this PR was written to fix, now
reachable through a second door. The sibling you are modelling on does not go quiet: `k_ef_create`
at `:2903-2911` emits a rate-limited (first 8) `REFUSED unwritable out-pointer` line that names the
API and says the cause is guest-side. Do the same here, keep the `return 0`, and the condition
becomes diagnosable instead of invisible. That is ~6 lines copied from 50 lines above.

**Separately, and reassuringly: the gate does not neuter the fix.** I checked what `guest_writable`
actually answers rather than assuming. On Linux (`gpu_executor.cpp:2688`, `:2582` onward) it is a
real writability probe over `/proc/self/maps`, thread-locally cached and generation-guarded, not a
registry of "recognised guest allocations". An ordinary stack out-pointer — which is the shape your
own live capture recorded for `a2` (`0x7fb4f42eb60c`) — sits in an `rw-p` mapping and answers true.
It refuses only genuinely unmapped or read-only memory, i.e. exactly the stores that would have
faulted. So the skip path should be rare-to-never in practice, which is another argument for making
it loud: if it ever fires, you want to know.

(Cost note, no action needed: on a cache miss that arm parses `/proc/self/maps`. Across ~3000 calls
in a boot, with the generation-guarded thread-local cache absorbing most of them, this is noise.
Recording it only so nobody rediscovers it as a mystery.)

---

## Non-blocking

**1. `hle_file.cpp:750`'s one-line addition to `reenters_mount` is untested, and I would spend a line
on it before merging.** The four new sandbox assertions do not reach it: `/download0/patch.pkg` has
no `..`, and `/download0/../etc/passwd` normalises to `/etc/passwd`, which matches no mount and is
denied whether or not `/download0` is in the list. Deleting your addition to `:750` reddens **nothing**
in the suite. In fairness the whole `reenters_mount` branch (#1234) has never had coverage, so this
PR extends a pre-existing gap rather than opening one — but your line is one assertion away:

    CHECK(resolve_guest_path("/download0/../download0/patch.pkg") ==
          resolve_guest_path("/download0/patch.pkg"),
          "/download0 re-entry through '..' maps like the direct spelling");

Without the `:750` entry the left side is `/prosper-denied/...` and it reddens; with it, both sides
are the mapped host path. I traced both branches by hand.

**2. The test can pin the assign-vs-OR contract for free, which is the one thing B2 could only
downgrade.** Right now `initPattern` and `setPattern` are both 0 (`:181`, `:203`), so `observed == 0`
proves the pattern was *written* but says nothing about *what* — and `0 | 0 == 0`, so the arm cannot
distinguish assign from OR at all. Create with `initPattern = 0x80`, cancel with `setPattern = 0x40`,
assert `observed == 0x40`. The waiter still parks (`0x80 & 0x1 == 0` in OR mode) and still cancels;
assigning gives `0x40` and OR-ing gives `0xC0`. One changed literal each turns "a pattern was
delivered" into "the pattern was assigned, not OR'd, and delivered" — which converts your honestly
downgraded contract claim into a pinned one. This is the one I would most like to see land.

**3. `t2.detach()` on the failure path** (`:229`) leaves a thread referencing `done2` and `ef` after
they leave scope. Identical to the pre-existing `t.detach()` at `:172`, so it is consistent with the
file and only reachable when the test is already failing. Noted, not asked for.

**4. `test_translate_sandbox.cpp:71-73` is close to redundant** with the equality check on the line
above it, and its `find(...) == npos || rfind(..., 0) != 0` form is hard to read for what it asserts.
Harmless.

**5. Your two deliberate non-fixes: I agree with both, and neither is a condition.** Raw hex is the
family's convention here and changing one member would be the inconsistency, not the fix — a separate
sweep is the right shape if anyone wants it. The cancel-after-delete UAF is shared with
`k_ef_set`/`k_ef_clear`, is a guest-side use-after-free, and fixing it belongs with whatever
eventually gives this family real handle validation, not here.

---

**APPROVED** at `3ed1859d2`. Nothing above blocks. If you land only one of them, make it (2) — it
costs two literals and buys the only measured evidence that will ever exist for the assign semantics.
(1) and the (b) one-liner are the next two, and the (c) log line after that.

Merge gate at review time: 7 passed, 4 pending (Linux, Windows MinGW, Windows App, macOS Rosetta);
head matches tip. Re-check it immediately before merging — this approval is on the content at
`3ed1859d2`, and any push detaches it.

mattias800 (commented):
# Final re-review at `f09b1a9c2` — APPROVED

All five items landed. I traced every new assertion for wrong-reason passes rather than reading the
summary, and one thing came out of it — **a correction to something I told you last round, which you
then acted on and wrote into a code comment.** That is on me, so it goes first.

---

## Correction: my "discriminates under both scheduling orders" claim was too strong

I said asserting t2's return code closes the vacuity hole in the 40 ms arm. **It does not close it for
the mutation you actually ran.** Tracing your `e->cancel_gen == 0` loop mutation against the current
code:

- t2 enters with `entry_cancel_gen = 1`. Loop guard `e->cancel_gen == 0` → `1 == 0` → false, so the
  loop body never runs and t2 falls straight through.
- The post-loop verdict is **unmutated**: `cancelled = (e->cancel_gen != entry_cancel_gen)` → `1 != 1`
  → **false**. So `ret = 0`.
- `wret2 == 0` therefore **passes** under the broken build. `!done2` at 40 ms is what reddens.

And the starved case is still open: if t2 has not been scheduled at 40 ms, `!done2` passes vacuously;
t2 then runs after `ef_set`, matches `0x41 & 0x1` immediately, and returns 0 — so `wret2 == 0` passes
too, and the whole arm is green with the mutation live.

My claim was correct only for a *different* sticky variant — one where the post-loop verdict is also
sticky (`cancelled = e->cancel_gen != 0`), which returns ECANCELED regardless of scheduling. So the
two checks catch **different mutation classes and neither is redundant**: the timing check catches
loop-guard mutations, the return check catches post-loop-verdict mutations. Keeping both is right. I
was wrong that the second subsumes the first.

**The practical consequence is the comment at `test_sync_delete.cpp:234-238`**, which now states my
over-claim as fact: *"a sticky cancel yields ECANCELED no matter when t2 was scheduled."* True for one
sticky variant, false for the one you mutated. Narrow it, or strengthen the arm.

**The form that actually closes it**, if you want to: the only way to prove t2 stayed blocked is to
*observe* it blocked, and `ef_cancel`'s waiter count is the one instrument prosper exposes. Replace
the `ef_set` with a second cancel:

    int32_t parked2 = -12345;
    ef_cancel((uint64_t)(uintptr_t)ef, 0x20, (uint64_t)(uintptr_t)&parked2, 0, 0, 0);
    CHECK(parked2 == 1, "t2 was genuinely parked -- the 40 ms window was not vacuous");
    ... CHECK(wret2.load() == kECANCELED, ...);

`parked2 == 1` is a positive observation rather than an inference from "has not finished yet", so it
holds under any scheduling, and it covers your loop-guard mutation directly (t2 would not be parked to
count). It also exercises a real property of the generation design you have no coverage for: a *second*
cancel releases the waiter that parked after the first. You would lose "still wakes on a genuine set" —
if you want that too it needs a third thread, but that arm has no wrong-reason risk worth engineering
around: a starved t3 matches immediately and returns 0, which is the correct answer, so no wrong
implementation is masked by it.

**Not blocking.** The arm does bind today — you mutation-proved it on a real box — and the hole is only
the starved-scheduler case. But the comment asserts more than the arm delivers, and comments outlive
the session that wrote them.

---

## Everything else — traced, sound

**`0x80` in / `0x40` cancelled is the right pair, and I checked it does not break the arms around it.**
`evf_match(0x80, 0x1, OR)` is false so the waiter still parks; the cancel assigns `0x40`, and
`evf_match(0x40, 0x1, OR)` is still false, so the waiter exits on the generation change rather than on
a match — which keeps the ECANCELED assertion meaningful rather than incidental. `observed` is poisoned
to `~0ull`, so the check separates three states, not two: untouched (`~0`), OR-ed (`0xc0`), assigned
(`0x40`). **This is the strongest assertion in the PR.** It is the only evidence that will ever exist
for the assign semantics short of a title that parks a waiter, and it converts the claim B2 forced you
to downgrade into a pinned one. Good outcome.

**The `reenters_mount` assertion discriminates exactly its clause.** `/app0/../download0/patch.pkg`
clamps to `/download0/patch.pkg`, matches the clause, and maps; delete `/download0` from `:750` and it
falls to the deny and returns `/prosper-denied/...`, which cannot equal `download0`. It is also
properly anchored — `download0` is pinned as the *mapped* path by the assertion three lines above, so
"both denied" cannot satisfy it. Both calls take identical paths through `resolve_host_path_case`, so
the equality is deterministic.

**Null path.** `poisoned` starts at `-999` and can only become 0 through the new store, so the
assertion cannot be satisfied by the defect. Your reasoning for writing 0 rather than leaving it is
the one I wanted: nothing was cancelled, so 0 is the honest count, and it removes the residue without
disturbing the family's return convention.

**The refusal log** matches `k_ef_create:2903-2911`'s shape — function-local `std::atomic`, first 8,
suppression notice — and its text says the right thing: the cancel took effect, SCE_OK is returned, the
count is unwritten. That is the one case prosper genuinely cannot answer, and it now says so.

**Trivial:** `test_sync_delete.cpp:242-243` has the old one-line null-handle comment left above the new
three-line one, so the lead-in is stated twice.

---

**APPROVED** at `f09b1a9c2`. Nothing blocks. Take the comment narrowing at `:234-238` if you take
nothing else — an over-claim of mine should not outlive this review inside the file.

Merge gate at review time: 7 checks pending, head matches tip. Re-check immediately before merging;
this approval is on the content at `f09b1a9c2`.

mattias800 (commented):
# Re-review at `047bf5efb` — APPROVED

Re-reviewed rather than carried forward, because the changed arm is now the load-bearing one. Head
`047bf5ef` matches the branch tip.

## `parked2 == 1` — can it pass for the wrong reason? I traced six routes and found none

The reason it is sound is structural rather than statistical, and it is worth writing down because it
is *why* this construction is better than the sleep it replaced:

**Between `waiters++` (`hle_kernel.cpp:3461`) and `--waiters` (`:3525`), a waiter holds `e->m` except
while it is inside `cond_wait`. `ef_cancel` reads `e->waiters` while holding `e->m` (`:2957-2958`).**
So any thread the count sees is *necessarily* blocked in the condvar at that instant. It cannot be a
thread that entered the wait and matched immediately, and it cannot be one that is about to park.
`parked2 == 1` therefore means "t2 is parked", by mutual exclusion, with no timing assumption in it at
all. That is a genuine observation, and it is what the previous `!done2` could never be.

The specific routes I checked:

- **Sticky loop guard** (`e->cancel_gen == 0`): t2 falls through the first cancel, `--waiters` runs,
  count reads 0 → reddens. Matches your empirical three-assertion result.
- **Starved scheduler**: t2 has not reached `waiters++`, count reads 0 → reddens. The hole is closed
  in the right direction — loud instead of quiet.
- **`== 1`, not `> 0`**: a stray extra waiter also reddens, and `t` was joined at `:214` before this
  block, so t2 is the only thread that can be in there.
- **Poison `-4242`** is distinct from the block's other two (`-12345`, `-999`), so a copy-paste is
  visible, and it cannot satisfy `== 1`.
- **`setPattern = 0x2` on the second cancel is the right choice**: `0x2 & 0x1 == 0`, so t2 does not
  match and must leave through the generation change. ECANCELED is reached via the cancel path rather
  than incidentally via a match — with `0x1` or `0x3` it would still be ECANCELED through the
  `else if (cancelled)` override, but for a muddier reason.
- **Cross-talk**: the OR-instead-of-assign mutation leaves `0xc0`, and `0xc0 & 0x1 == 0`, so t2 still
  parks and `parked2 == 1` still passes — that mutation is caught by `observed == 0x40` alone. No arm
  double-counts, which is what "one arm each" should mean.

`parked2` is written by `ef_cancel` on the calling thread before it returns and read by that same
thread, so there is no race on it either. Teardown is clean: t2 joined, `waiters == 0`, `bits == 0x2`,
`ef_delete` destroys.

**And the two sticky variants are now both covered**, which was the actual gap: loop-guard-only by
`parked2 == 1`, post-loop-verdict by `wret2 == kECANCELED`. Neither assertion subsumes the other.

**One property worth knowing when this ever reddens, because it makes triage unambiguous:** the pair
disambiguates the cause. Starvation gives `!done2` **green** and `parked2` **red**; a real sticky
defect gives **both** red. So a lone `parked2` failure is the harness, and a pair is the code. That
falls out of the construction rather than being designed in, but it is worth having said once.

Small nit, not asked for: `CHECK` prints only its message, so a failure will not tell you whether
`parked2` was `-4242` (never written) or `0` (written, nobody parked) — two different diagnoses. A
printed value would pay for itself the one time it fires.

## The rest

**The comment is the right fix.** Recording that the earlier construction's claim was false, and why
(`test_sync_delete.cpp:223-232`), is better than deleting the wording — it stops the next reader
re-deriving the return-code check as sufficient. Thank you for taking it that way round.

**Duplicated null-handle comment: gone.**

**The ASCII gate: verified directly rather than taken on trust.** The net diff `f09b1a9c2..047bf5efb`
contains no literal change, so the em-dash fix was amended in before the push and is invisible to a
diff — which means the only honest check is the end state. I ran
`tools/output/check_ascii_output.py` against the four files this PR touches: **1109 string literals
examined, zero non-ASCII.** (The 13 failures in my run are quarantine-ledger rows for files absent
from the minimal root I built, i.e. my harness, not the tree.)

---

**APPROVED** at `047bf5efb`. This is the strongest the arm can reasonably be made, and the assign-vs-OR
assertion remains the best thing in the PR.

Merge gate at review time: 5 checks pending, head matches tip, no session trailers, diff is the
expected six files. Re-check the gate immediately before merging — this approval is on `047bf5efb`.

PR #3514, opened 2026-09-10
tools/refactor: survey which refactoring tool can act on each large file

mattias800:
## What this adds

`prosper/tools/refactor/survey_sizes.py` — a repo-wide survey that answers **which refactoring tool can act on each large file**, so targets are chosen by shape rather than by `wc -l`. A 12,000-line file that is two hundred functions is an afternoon with `split_file.py`; a 12,000-line file that is one function is not, and no amount of splitting makes it one.

Four verdicts, plus three that decline to score:

| verdict | meaning |
| --- | --- |
| `SPLIT` | no region dominates — `split_file.py` applies today |
| `EXTRACT` | one region dominates but is lambda-light — clangd's ExtractFunction can act |
| `HAND` | dominant **and** mostly lambda bodies — clangd refuses ("Don't extract from lambdas"), no tool helps |
| `UNPARSED` | a large span sits behind an inactive `#if`, so every AST verdict would describe a different part of the file |
| `PARSE-FAIL` / `NO-CURSOR` / `NO-PREPROC` | reported, never scored |

## Result over the tree — 28 files at or above 2,500 lines

`EXTRACT=13, SPLIT=9, UNPARSED=4, HAND=1, PARSE-FAIL=1`

**22 of 28 have a route.** Exactly one file is untouchable: `frontends/shared/live/live_renderer.cpp`, whose `register_live_renderer` is 10,244 lines at **98% lambda**.

This corrects the impression the README previously left, that the giant functions are mostly beyond tooling:

- **`rdna2_emit_alu.cpp`'s 8,186-line `emit_alu` is 11% lambda**, so clangd can extract from it. So can it from the giant test `main`s — `test_rdna2_to_spirv.cpp` at 3%, `test_recompile_coverage.cpp` at 2%.
- **`hle_kernel_mem.cpp` is `UNPARSED` at 52%, not a plain split.** 3,978 lines across seven skipped spans, the largest of them 3,908 lines, sit behind inactive `#if`s -- and that largest arm alone is the 51% the superseded single-span heuristic reported, which is why the two figures should not be confused. So `split_file.py` cannot drive #3503 — the tool is blind to the very arm that issue wants to move. That needs a textual line-range move at the directive boundaries. Recorded on #3503; three more files share the shape (`hle_kernel.cpp` 30%, `hle_file.cpp` 21%, `guest_write_watch.cpp` 20%).

## The interesting part is how often this tool was wrong

The classifier has been wrong **five times**, and `CMakeLists.txt` records all five next to the test that pins them. Two were arithmetic. **Three were in how the classifier's INPUTS are measured**, so every arithmetic case passed happily while the tool reported the opposite verdict for a real file — and none produced an error or an odd-looking row:

1. **The lambda share was measured over the enclosing namespace.** Nearly every file here is one `namespace prosper { ... }`, and a namespace answers `is_definition()` True, so the scan for the dominant region's cursor always found it. 6 of the 14 files that have a share at all were inflated, worst `rdna2_emit_cfg.cpp` at 0.428 against a true 0.258. **No verdict changed** — luck, since 0.50 is the threshold.
2. **Unparsed lines were inferred from region kinds**, which structurally cannot see an inactive `#if` that is not at the end of the file. `guest_write_watch.cpp` — a 481-line `#ifdef _WIN32` arm mid-file — reported **0** unparsed with a dominant "symbol" of `sched.h`, the include whose region had absorbed it. Now asks the preprocessor (`clang_getSkippedRanges`) and sums every arm.
3. **`region_cursor` did not mirror `regions_of`'s descent rule**, so a namespace that is itself a body region was walked past and scored `NO-CURSOR`.

And once in the *fix* for (2): a fail-closed guard that could not fire. `File.from_name(...) is None` never is — `clang_getFile` resolves through the file manager and answers for any file on disk, read by the TU or not — so a foreign file returned a valid handle, an empty range list and a confident **0**, under a README paragraph promising the opposite. Membership is tested against the TU's inclusion record now.

## Verification

Each of the three input defects is pinned by a fixture under `tools/refactor/testdata/`, built so the **defective measurement yields a different verdict**, not merely a different number. Verified by re-running each pre-fix version against all of them: every mutation reddens its own fixture and leaves the others green.

| mutation | reddens |
| --- | --- |
| pre-fix top-level cursor scan | `dominant_no_lambda` → `HAND` at 0.729 for a function with no lambda |
| pre-fix region-kind inference | `midfile_inactive_arm` → `EXTRACT`, 0 unparsed, naming `posix_only_1` |
| descend on depth alone | `namespace_without_children` → `NO-CURSOR` |
| `resolve()` path comparison | the hard-link arm |

That last arm exists because the inode comparison had none: every fixture uses one spelling, so reverting it left them all green. A **hard link** reproduces the property the bind-mount alias has and nothing else here does — two paths naming one file that `resolve()` reports as different. Under the alias spelling the old comparison discarded every span and reported `hle_file.cpp` as having **0** skipped lines instead of 963.

- `ctest --no-tests=error` — **455** tests; two registered cases here, `refactor_survey_classifier` (13 arithmetic cases, runs everywhere) and `refactor_survey_measurement` (4 measurement checks, `SKIP_RETURN_CODE 77`).
- The measurement checks are a **separate** case deliberately: folded into one, they simply do not run on a runner without libclang and ctest prints `Passed` — a green result on a strictly smaller domain with nothing saying so. `Skipped` is the honest form, and this project already records that hazard for its GPU tests under CI's software rasteriser.
- Docs numbered-table gate clean; `diff --check` clean; 0 `Claude-Session:` trailers.

## Risk

Tooling only — no emulator code, nothing linked into any target. The fixtures live under `tools/refactor/testdata/`, which nothing globs (`CMakeLists.txt:88` globs `src/*.cpp` only).

Follow-ups filed: #3515 (`map_symbols.py` documents a `--clusters` flag that does not exist) and #3516 (`gpu_executor.cpp`, the largest `SPLIT` file, with its seam measured from the reference matrix).

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -170,9 +170,10 @@ if(Python3_Interpreter_FOUND)
   add_test(NAME re_xref_decoder
            COMMAND "${Python3_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/tools/re/test_xref.py")
   # survey_sizes.py's classifier decides which refactoring tool can act on a file, and it has been
-  # wrong four times: on the lambda boundary, by calling a file of many small functions "nothing to
-  # do", by measuring the lambda share over the enclosing namespace, and by inferring unparsed lines
-  # from region kinds (which cannot see a mid-file #if). The last two were in how the classifier's
+  # wrong five times: on the lambda boundary, by calling a file of many small functions "nothing to
+  # do", by measuring the lambda share over the enclosing namespace, by inferring unparsed lines
+  # from region kinds (which cannot see a mid-file #if), and by not mirroring regions_of's descent
+  # rule. The last three were in how the classifier's
   # INPUTS are measured, so no arithmetic case could have caught them; they are pinned by fixtures
   # under tools/refactor/testdata/ that the selftest scores ONLY where libclang is importable, and
   # it says loudly when it skips them. libclang resolves lazily, so the arithmetic half runs here
@@ -180,6 +181,14 @@ if(Python3_Interpreter_FOUND)
   add_test(NAME refactor_survey_classifier
            COMMAND "${Python3_EXECUTABLE}"
                    "${CMAKE_SOURCE_DIR}/tools/refactor/survey_sizes.py" --selftest)
+  # The measurement fixtures are a SEPARATE case so that a runner without libclang reports them as
+  # Skipped rather than folding them into a green Passed. A pass whose domain is smaller than the
+  # local one, with nothing in the report saying so, is the hazard this project records for its GPU
+  # tests under CI's software rasteriser.
+  add_test(NAME refactor_survey_measurement
+           COMMAND "${Python3_EXECUTABLE}"
+                   "${CMAKE_SOURCE_DIR}/tools/refactor/survey_sizes.py" --selftest-parse)
+  set_tests_properties(refactor_survey_measurement PROPERTIES SKIP_RETURN_CODE 77)
   if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
     add_test(NAME profiling_access
              COMMAND "${Python3_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/tools/profiling_access/test_access.py")
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -170,8 +170,12 @@ if(Python3_Interpreter_FOUND)
   add_test(NAME re_xref_decoder
            COMMAND "${Python3_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/tools/re/test_xref.py")
   # survey_sizes.py's classifier decides which refactoring tool can act on a file, and it has been
-  # wrong twice: once on the lambda boundary, once by calling a file of many small functions
-  # "nothing to do". Its selftest is pure arithmetic and resolves libclang lazily, so it runs here
+  # wrong four times: on the lambda boundary, by calling a file of many small functions "nothing to
+  # do", by measuring the lambda share over the enclosing namespace, and by inferring unparsed lines
+  # from region kinds (which cannot see a mid-file #if). The last two were in how the classifier's
+  # INPUTS are measured, so no arithmetic case could have caught them; they are pinned by fixtures
+  # under tools/refactor/testdata/ that the selftest scores ONLY where libclang is importable, and
+  # it says loudly when it skips them. libclang resolves lazily, so the arithmetic half runs here
   # even though the surveying itself needs a libclang the runner does not have.
   add_test(NAME refactor_survey_classifier
            COMMAND "${Python3_EXECUTABLE}"
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -169,6 +169,13 @@ if(Python3_Interpreter_FOUND)
                    "$<TARGET_FILE:self_dump>")
   add_test(NAME re_xref_decoder
            COMMAND "${Python3_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/tools/re/test_xref.py")
+  # survey_sizes.py's classifier decides which refactoring tool can act on a file, and it has been
+  # wrong twice: once on the lambda boundary, once by calling a file of many small functions
+  # "nothing to do". Its selftest is pure arithmetic and resolves libclang lazily, so it runs here
+  # even though the surveying itself needs a libclang the runner does not have.
+  add_test(NAME refactor_survey_classifier
+           COMMAND "${Python3_EXECUTABLE}"
+                   "${CMAKE_SOURCE_DIR}/tools/refactor/survey_sizes.py" --selftest)
   if(CMAKE_SYSTEM_NAME STREQUAL "Linux")
     add_test(NAME profiling_access
              COMMAND "${Python3_EXECUTABLE}" "${CMAKE_SOURCE_DIR}/tools/profiling_access/test_access.py")
--- a/prosper/tools/refactor/README.md
+++ b/prosper/tools/refactor/README.md
@@ -148,7 +148,7 @@ The measurement moved three specific beliefs:
 
 ### Three ways this survey measured the wrong thing, and all of them failed silently
 
-Neither produced an error or an odd-looking row. Each produced a confident verdict about a file
+None of them produced an error or an odd-looking row. Each produced a confident verdict about a file
 other than the one asked about, which is why they are recorded here rather than in a commit message.
 All three are now pinned by fixtures under `testdata/`, and each fixture is built so the defective
 measurement yields a **different verdict**, not merely a different number -- verified by re-running
@@ -262,10 +262,10 @@ python3 prosper/tools/refactor/survey_sizes.py --min-lines 2500     # needs libc
 python3 prosper/tools/refactor/survey_sizes.py --selftest           # no libclang needed
 ```
 
-`--selftest` runs the classifier's arithmetic cases always, and the two `testdata/` measurement
-fixtures **only where libclang is importable** -- saying so explicitly when it skips them, because
-those two are the checks that cover how the classifier's inputs are measured, and both shipped
-defects lived there rather than in the arithmetic. The registered ctest case
+`--selftest` runs the classifier's arithmetic cases always, and the four `testdata/` measurement
+checks **only where libclang is importable** -- saying so explicitly, and naming them, when it skips
+them. Those are the checks covering how the classifier's INPUTS are measured, and every defect this
+tool shipped lived there rather than in the arithmetic. The registered ctest case
 (`refactor_survey_classifier`) therefore gates the arithmetic everywhere and the measurement locally.
 
 **The invariant that makes incremental extraction safe is worth copying.** When a block was extracted
--- a/prosper/tools/refactor/README.md
+++ b/prosper/tools/refactor/README.md
@@ -83,7 +83,7 @@ caveats found by trying it:
   functions.** From `clang-tools-extra/clangd/refactor/tweaks/ExtractFunction.cpp`, the tweak
   returns `nullptr` when a `LambdaExpr` is found — the comment is literally *"Don't extract from
   lambdas"*. `register_live_renderer` spans 8,221 lines of which **98.5%** is inside lambdas (8,101 of 8,221, from the AST — 89% is the
-  file-share figure in the table above, a different number) passed to registration calls, so every candidate worth extracting is in refused territory.
+  file-share figure in the table above, a different number) passed to registration calls, so every candidate worth extracting is in refused territory. (Those counts are from an earlier commit and the file has grown since; `survey_sizes.py` measures the same function at 10,244 lines and 98% lambda on 2026-09-10. The conclusion is unchanged, and the two figures are kept distinct rather than reconciled because they were taken at different heads.)
   `extract_function.py --probe` reports 0 of 10 accepted at its default bounds, and that is correct
   behaviour, not a bug. **That figure alone does not establish it**: a hand-built lambda-free
   25-line function also probes 0 of 1, while the same clangd session returns
@@ -124,13 +124,14 @@ lambda share of every dominant function in the tree, which is the datum that dec
 
 | verdict | files | meaning |
 | --- | --- | --- |
-| `SPLIT` | 12 | no dominant region -- `split_file.py` applies today |
 | `EXTRACT` | 13 | one dominant region, but lambda-light -- clangd's ExtractFunction can act |
-| `UNPARSED` | 1 | a large span the AST never saw -- see below; no AST tool can act until it is split |
+| `SPLIT` | 9 | no dominant region -- `split_file.py` applies today |
+| `UNPARSED` | 4 | a large span the AST never saw -- see below; no AST tool can act until it is split |
 | `HAND` | **1** | dominant AND mostly lambda -- clangd refuses, no tool can help |
+| `PARSE-FAIL` | 1 | not scored: `prosper-app/main.cpp` needs `-DPROSPER_APP=ON` to have a command |
 
 Exactly one file is in the untouchable state: `frontends/shared/live/live_renderer.cpp`, whose
-`register_live_renderer` is 10,244 lines at **99% lambda**. Everything else has a route.
+`register_live_renderer` is 10,244 lines at **98% lambda**. Everything else has a route.
 
 The measurement moved three specific beliefs:
 
@@ -138,26 +139,67 @@ The measurement moved three specific beliefs:
   work. So are the giant test `main`s -- `test_rdna2_to_spirv.cpp` at **3%**,
   `test_dynfetch_fold.cpp` at 6%, `test_recompile_coverage.cpp` at 2%. The per-block assertion-delta
   invariant recorded at the end of this file is exactly the tool for those.
-* **The two largest source files are plain `SPLIT` work** -- `gpu_executor.cpp` (12,058 lines,
-  biggest region 23%) and `hle_kernel_mem.cpp` (7,637 lines, biggest region **2%**, #3503).
+* **The largest source file is plain `SPLIT` work** -- `gpu_executor.cpp`, 12,058 lines, biggest
+  region 23%.
 * **A file of many small functions is the EASIEST split, and the first version of this tool called it
   `OK`.** It required two regions of >=200 lines before offering `SPLIT`, which ranked
   `hle_kernel_mem.cpp` as nothing-to-do. Region COUNT is what a seam needs; region SIZE is not.
   Fixed, and pinned by a selftest case carrying that file's real shape.
 
-### The AST cannot see the inactive side of an `#if`, and that is its own state
+### Two ways this survey measured the wrong thing, and both failed silently
 
-`hle_kernel_mem.cpp` is 7,637 lines of which the Windows arm is **3,909 -- 51% of the file** -- behind
-an `#if defined(__linux__) || defined(__APPLE__)`. Parsed on Linux, clang never sees it: the region
-map runs out at line 3720 and the whole Windows arm arrives as a single `TRAILER`.
+Neither produced an error or an odd-looking row. Each produced a confident verdict about a file
+other than the one asked about, which is why they are recorded here rather than in a commit message.
+Both are now pinned by fixtures under `testdata/`, and each fixture is built so the defective
+measurement yields a **different verdict**, not merely a different number -- verified by re-running
+the pre-fix code against them (`dominant_no_lambda` -> `HAND`; `midfile_inactive_arm` -> `EXTRACT`,
+0 unparsed).
 
-This is worth its own verdict rather than a footnote, because **every AST-driven statement about such
-a file is answering about the other half**. The first version of this tool scored the file `SPLIT`
-from the 49% it could see -- and worse, it discarded the trailer as scaffolding first, because
-`map_symbols` gives the outermost namespace's trailing text the `close` role and this tool filtered
-`open`/`close` out. 3,909 lines vanished from the accounting silently. Caught by cross-checking the
-survey's own numbers against `map_symbols.py` on the same file; nothing in the survey's output looked
-wrong on its own, which is the point.
+**The lambda share was measured over the enclosing namespace.** Nearly every file here is one
+`namespace prosper { ... }`, so a file's only *top-level* cursor is the namespace -- and a namespace
+answers `is_definition()` with True. Walking top-level cursors to find the dominant region's cursor
+therefore always found the namespace, and unioned every lambda in the whole file into the dominant
+function's share. Measured across the tree: **6 of 22 scored files carried an inflated share**, worst
+`rdna2_emit_cfg.cpp` at 0.428 against a true **0.258**. **No verdict actually changed** -- in a
+dominated file the namespace is nearly the dominant function, so the error stayed under the 0.50
+threshold everywhere. That is luck, not design: 0.428 is one small edit from flipping a file to
+"no tool can act on this". The fix descends to the region's own cursor (`region_cursor`).
+
+**Unparsed lines were inferred from region kinds, which cannot see a mid-file `#if`.** `map_symbols`
+tiles a file by cursor, so an inactive arm is never a region of its own -- it is folded into whichever
+region follows it and wears that region's kind. Keying on `TEXT`/`TRAILER` therefore saw an arm only
+when it ran to the END of the file. `src/host/memory/guest_write_watch.cpp` is the tree's own
+counter-instance: a 481-line `#ifdef _WIN32` arm at lines 71-551, 19% of the file, reported as **0**
+unparsed lines with a dominant "symbol" of `sched.h` -- the include directive whose region had
+absorbed the arm.
+
+The fix asks the preprocessor instead, through `clang_getSkippedRanges` (reached by ctypes; the
+Python bindings do not wrap it), and **sums every skipped arm** rather than counting only ones
+individually above the threshold. That matters twice over: it found three more files than the
+trailing-span heuristic did, taking `UNPARSED` from 1 to 4 --
+
+| file | lines | unparsed | was |
+| --- | --- | --- | --- |
+| `src/hle/memory/hle_kernel_mem.cpp` | 7,637 | **52%** (7 spans) | UNPARSED, at 51% from one span |
+| `src/hle/kernel/hle_kernel.cpp` | 5,395 | **30%** | SPLIT |
+| `src/hle/fs/hle_file.cpp` | 4,560 | **21%** | SPLIT |
+| `src/host/memory/guest_write_watch.cpp` | 2,513 | **20%** | SPLIT |
+
+-- and it fails **closed**: if `clang_getSkippedRanges` is missing the tool exits rather than
+reporting zero, because reporting "no unparsed lines" when the question cannot be asked is exactly
+the silent under-report above.
+
+`UNPARSED_SHARE` (0.15) is a **convention, not a measurement**. The tree has four files above it, the
+lowest at 20%, and nothing here pins where in the gap below that the line belongs.
+
+### Why `UNPARSED` is its own verdict rather than a footnote
+
+**Every AST-driven statement about such a file is answering about the other part.** The first version
+of this tool scored `hle_kernel_mem.cpp` as `SPLIT` from the 48% it could see -- and worse, discarded
+the trailing arm as scaffolding first, because `map_symbols` gives the outermost namespace's trailing
+text the `close` role and this tool filtered `open`/`close` out. 3,909 lines left the accounting
+silently. Caught by cross-checking the survey's own numbers against `map_symbols.py` on the same
+file; nothing in the survey's output looked wrong on its own, which is the point.
 
 The practical consequence for #3503: **that split cannot be driven by `split_file.py`**, because the
 tool is blind to the arm being moved. A platform-arm split is a textual line-range move at the
@@ -169,6 +211,20 @@ like a small, uninteresting file. In a survey that is the worst place for it to 
 would be ranked as needing no work. One file is currently in that state:
 `frontends/prosper-app/main.cpp`, which has no compile command unless the app is configured
 (`-DPROSPER_APP=ON`); configure with it to bring that file into the survey.
+
+### Running it
+
+```bash
+python3 prosper/tools/refactor/survey_sizes.py --min-lines 2500     # needs libclang + a build dir
+python3 prosper/tools/refactor/survey_sizes.py --selftest           # no libclang needed
+```
+
+`--selftest` runs the classifier's arithmetic cases always, and the two `testdata/` measurement
+fixtures **only where libclang is importable** -- saying so explicitly when it skips them, because
+those two are the checks that cover how the classifier's inputs are measured, and both shipped
+defects lived there rather than in the arithmetic. The registered ctest case
+(`refactor_survey_classifier`) therefore gates the arithmetic everywhere and the measurement locally.
+
 **The invariant that makes incremental extraction safe is worth copying.** When a block was extracted
 out of `test_rdna2_to_spirv.cpp`'s `main`, the thing that made it checkable was a count of assertions
 *executed*, asserted as a per-block delta:
--- a/prosper/tools/refactor/README.md
+++ b/prosper/tools/refactor/README.md
@@ -146,24 +146,41 @@ The measurement moved three specific beliefs:
   `hle_kernel_mem.cpp` as nothing-to-do. Region COUNT is what a seam needs; region SIZE is not.
   Fixed, and pinned by a selftest case carrying that file's real shape.
 
-### Two ways this survey measured the wrong thing, and both failed silently
+### Three ways this survey measured the wrong thing, and all of them failed silently
 
 Neither produced an error or an odd-looking row. Each produced a confident verdict about a file
 other than the one asked about, which is why they are recorded here rather than in a commit message.
-Both are now pinned by fixtures under `testdata/`, and each fixture is built so the defective
+All three are now pinned by fixtures under `testdata/`, and each fixture is built so the defective
 measurement yields a **different verdict**, not merely a different number -- verified by re-running
-the pre-fix code against them (`dominant_no_lambda` -> `HAND`; `midfile_inactive_arm` -> `EXTRACT`,
-0 unparsed).
+each pre-fix version against all of them, where every mutation reddens its own fixture and leaves
+the others green (`dominant_no_lambda` -> `HAND` at a 0.729 share for a function with no lambda;
+`midfile_inactive_arm` -> `EXTRACT` with 0 unparsed, naming the include directive that swallowed the
+arm; `namespace_without_children` -> `NO-CURSOR`).
 
 **The lambda share was measured over the enclosing namespace.** Nearly every file here is one
 `namespace prosper { ... }`, so a file's only *top-level* cursor is the namespace -- and a namespace
 answers `is_definition()` with True. Walking top-level cursors to find the dominant region's cursor
 therefore always found the namespace, and unioned every lambda in the whole file into the dominant
-function's share. Measured across the tree: **6 of 22 scored files carried an inflated share**, worst
-`rdna2_emit_cfg.cpp` at 0.428 against a true **0.258**. **No verdict actually changed** -- in a
+function's share. Measured across the tree: **6 of the 14 files that have a lambda share at all carried an inflated
+one**, worst `rdna2_emit_cfg.cpp` at 0.428 against a true **0.258**. Fourteen is the honest
+denominator, and it is not the number of scored files: the share is computed only for a file a single
+region dominates, so the 9 `SPLIT` files could never have been inflated, and `live_renderer.cpp` --
+which was -- is neither `SPLIT` nor `EXTRACT`. So the error rate among files where the question is
+asked is 43%, not the 27% a count against all scored files suggests. **No verdict actually changed** -- in a
 dominated file the namespace is nearly the dominant function, so the error stayed under the 0.50
 threshold everywhere. That is luck, not design: 0.428 is one small edit from flipping a file to
 "no tool can act on this". The fix descends to the region's own cursor (`region_cursor`).
+
+**`region_cursor` did not mirror `regions_of`'s descent rule.** `emit()` descends into a namespace
+only when it HAS in-target children (`map_symbols.py:174`); the first version of `region_cursor`
+descended on `depth < max_depth` alone. A namespace with no in-target children is emitted as a plain
+`body` region by the kids-empty fallthrough, so it can BE a file's dominant region -- and the lookup
+walked straight past it and reported `NO-CURSOR`, refusing to score a file whose answer was not in
+doubt. Reachable with nothing exotic: a namespace whose body is all comments
+(`testdata/namespace_without_children.cpp`, 98.9% of its file). No file in the tree currently hits
+it, so this cost nothing -- it was found by constructing the case rather than by observing one, and
+it is recorded because `NO-CURSOR` firing on a false alarm would have read as a tool limitation
+rather than a tool bug.
 
 **Unparsed lines were inferred from region kinds, which cannot see a mid-file `#if`.** `map_symbols`
 tiles a file by cursor, so an inactive arm is never a region of its own -- it is folded into whichever
@@ -185,9 +202,22 @@ trailing-span heuristic did, taking `UNPARSED` from 1 to 4 --
 | `src/hle/fs/hle_file.cpp` | 4,560 | **21%** | SPLIT |
 | `src/host/memory/guest_write_watch.cpp` | 2,513 | **20%** | SPLIT |
 
--- and it fails **closed**: if `clang_getSkippedRanges` is missing the tool exits rather than
-reporting zero, because reporting "no unparsed lines" when the question cannot be asked is exactly
-the silent under-report above.
+-- and it fails **closed** everywhere the question cannot be answered, because "no unparsed lines"
+and "I could not ask" must never be the same output. A missing `clang_getSkippedRanges` exits; a file
+absent from its own TU, or a NULL range list, scores `NO-PREPROC` rather than 0.
+
+The subtlest of those was a path comparison. The clip that keeps other files' skipped include guards
+out of the count originally compared `resolve()`d paths -- and this repository is reachable under two
+spellings that resolve **differently** and name one file, the container's `/home/<user>/...` bind
+mount and the host's `/var/home/<user>/...`. Under the alias spelling every span was discarded and
+`hle_file.cpp` reported **0** skipped lines instead of 963, which would have scored it `SPLIT`. The
+comparison is by **inode** now (`os.path.samefile`), and where even that errors it counts the span
+rather than dropping it: an over-count raises `UNPARSED`, which someone sees, and an under-count is
+the silent wrong answer. Measured both spellings: 963 and 963.
+
+Span line numbers include the `#if` and `#endif` directives themselves, because that is what libclang
+reports and they are lines the AST produced nothing for. Anyone reconstructing these figures as
+"lines of skipped CODE" will get about two fewer per span.
 
 `UNPARSED_SHARE` (0.15) is a **convention, not a measurement**. The tree has four files above it, the
 lowest at 20%, and nothing here pins where in the gap below that the line belongs.
--- a/prosper/tools/refactor/README.md
+++ b/prosper/tools/refactor/README.md
@@ -154,8 +154,11 @@ All three are now pinned by fixtures under `testdata/`, and each fixture is buil
 measurement yields a **different verdict**, not merely a different number -- verified by re-running
 each pre-fix version against all of them, where every mutation reddens its own fixture and leaves
 the others green (`dominant_no_lambda` -> `HAND` at a 0.729 share for a function with no lambda;
-`midfile_inactive_arm` -> `EXTRACT` with 0 unparsed, naming the include directive that swallowed the
-arm; `namespace_without_children` -> `NO-CURSOR`).
+`midfile_inactive_arm` -> `EXTRACT` with 0 unparsed, naming `posix_only_1` -- the first declaration
+after the skipped arm, whose region had absorbed it; `namespace_without_children` -> `NO-CURSOR`).
+The same absorption in the tree names an `#include` rather than a function (`guest_write_watch.cpp`'s
+`sched.h`); which declaration a swallowed arm ends up attributed to is simply whichever cursor
+follows it.
 
 **The lambda share was measured over the enclosing namespace.** Nearly every file here is one
 `namespace prosper { ... }`, so a file's only *top-level* cursor is the namespace -- and a namespace
@@ -177,7 +180,7 @@ descended on `depth < max_depth` alone. A namespace with no in-target children i
 `body` region by the kids-empty fallthrough, so it can BE a file's dominant region -- and the lookup
 walked straight past it and reported `NO-CURSOR`, refusing to score a file whose answer was not in
 doubt. Reachable with nothing exotic: a namespace whose body is all comments
-(`testdata/namespace_without_children.cpp`, 98.9% of its file). No file in the tree currently hits
+(`testdata/namespace_without_children.cpp`, whose `commentary` region is 93 of 110 lines, 84.5%). No file in the tree currently hits
 it, so this cost nothing -- it was found by constructing the case rather than by observing one, and
 it is recorded because `NO-CURSOR` firing on a false alarm would have read as a tool limitation
 rather than a tool bug.
@@ -204,7 +207,17 @@ trailing-span heuristic did, taking `UNPARSED` from 1 to 4 --
 
 -- and it fails **closed** everywhere the question cannot be answered, because "no unparsed lines"
 and "I could not ask" must never be the same output. A missing `clang_getSkippedRanges` exits; a file
-absent from its own TU, or a NULL range list, scores `NO-PREPROC` rather than 0.
+the translation unit never read, or a NULL range list, scores `NO-PREPROC` rather than 0.
+
+That second guard was itself dead when first written, which is worth recording because it is this
+PR's own defect class appearing inside the fix for it. The check was `if File.from_name(...) is
+None` -- and it never is. `clang_getFile` resolves through the file manager, so it answers for any
+file that exists on disk whether or not the TU read it, and cindex asserts the handle is non-NULL
+besides. A foreign file therefore came back with a valid handle, an empty skipped-range list, and a
+confident **0**, behind a guard that could not fire. Membership is tested against the TU's own
+inclusion record now (`_file_in_tu`), which also turns an existing assumption into a check: a header
+has no compile command, so `flags_for` parses the first TU whose *text* mentions it, without
+confirming the include sits on an active branch.
 
 The subtlest of those was a path comparison. The clip that keeps other files' skipped include guards
 out of the count originally compared `resolve()`d paths -- and this repository is reachable under two
--- a/prosper/tools/refactor/README.md
+++ b/prosper/tools/refactor/README.md
@@ -124,8 +124,9 @@ lambda share of every dominant function in the tree, which is the datum that dec
 
 | verdict | files | meaning |
 | --- | --- | --- |
-| `SPLIT` | 13 | no dominant region -- `split_file.py` applies today |
+| `SPLIT` | 12 | no dominant region -- `split_file.py` applies today |
 | `EXTRACT` | 13 | one dominant region, but lambda-light -- clangd's ExtractFunction can act |
+| `UNPARSED` | 1 | a large span the AST never saw -- see below; no AST tool can act until it is split |
 | `HAND` | **1** | dominant AND mostly lambda -- clangd refuses, no tool can help |
 
 Exactly one file is in the untouchable state: `frontends/shared/live/live_renderer.cpp`, whose
@@ -143,6 +144,24 @@ The measurement moved three specific beliefs:
   `OK`.** It required two regions of >=200 lines before offering `SPLIT`, which ranked
   `hle_kernel_mem.cpp` as nothing-to-do. Region COUNT is what a seam needs; region SIZE is not.
   Fixed, and pinned by a selftest case carrying that file's real shape.
+
+### The AST cannot see the inactive side of an `#if`, and that is its own state
+
+`hle_kernel_mem.cpp` is 7,637 lines of which the Windows arm is **3,909 -- 51% of the file** -- behind
+an `#if defined(__linux__) || defined(__APPLE__)`. Parsed on Linux, clang never sees it: the region
+map runs out at line 3720 and the whole Windows arm arrives as a single `TRAILER`.
+
+This is worth its own verdict rather than a footnote, because **every AST-driven statement about such
+a file is answering about the other half**. The first version of this tool scored the file `SPLIT`
+from the 49% it could see -- and worse, it discarded the trailer as scaffolding first, because
+`map_symbols` gives the outermost namespace's trailing text the `close` role and this tool filtered
+`open`/`close` out. 3,909 lines vanished from the accounting silently. Caught by cross-checking the
+survey's own numbers against `map_symbols.py` on the same file; nothing in the survey's output looked
+wrong on its own, which is the point.
+
+The practical consequence for #3503: **that split cannot be driven by `split_file.py`**, because the
+tool is blind to the arm being moved. A platform-arm split is a textual line-range move at the
+`#if`/`#else`/`#endif` boundaries -- provable by concatenation, and needing no AST at all.
 
 `survey_sizes.py` reports a file whose parse emitted errors as `PARSE-FAIL` and gives it **no
 verdict**, because a libclang parse driven from a g++ database fails softly -- a partial AST looks
--- a/prosper/tools/refactor/README.md
+++ b/prosper/tools/refactor/README.md
@@ -12,6 +12,7 @@ each one does a transformation whose correctness can be *checked*, not merely re
 | `split_file.py` | splits one translation unit into several along a region partition |
 | `classify_tests.py` | proposes a folder for each test from its own includes |
 | `check_include_paths.py` | finds targets that reach a project include they cannot resolve |
+| `survey_sizes.py` | ranks every large file by WHICH of these tools can act on it |
 
 Run every tool's `--selftest` first; `split_file.py` and `promote_internal.py` run theirs
 automatically before doing anything.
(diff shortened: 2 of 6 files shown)


mattias800 (commented):
**REJECTED** — two blocking findings. Both are small fixes, and both make the measurement this PR is built on wrong in a way the tool's own output cannot reveal.

Reviewed at head `d7be172` (2 commits). I reproduced the survey at `--min-lines 2500` from the committed script and got the PR's exact table — `SPLIT=13, EXTRACT=13, HAND=1, PARSE-FAIL=1`, with `rdna2_emit_alu.cpp 8444/8186/97%/11%`. The reported numbers are real.

## Verified good

**The load-bearing 11% holds, derived independently rather than by re-running the tool.** I parsed `rdna2_emit_alu.cpp` from its own compile command, located `emit_alu` by walking to the largest in-file `FUNCTION_DECL` definition, and unioned `LAMBDA_EXPR` extents under *that* cursor: **8,185 lines, 880 lambda lines, 10.8%**, over 90 spans. Same route on `live_renderer.cpp`: `register_live_renderer` 10,243L, 10,067 lambda lines, **98.3%**. `EXTRACT` and `HAND` respectively are the right calls, and the README's correction of its own table stands.

**PARSE-FAIL is sound** (`survey_sizes.py:158`). `Diagnostic.Error` is 3 and `Fatal` is 4, so `>= Error` catches both. All three failure routes — `flags_for`'s `SystemExit`, `index.parse` raising, and a non-empty error list — return before `regions_of` is reached. No path scores a failed parse.

**The lazy libclang import is correct** (`:64-70`). Nothing at module scope touches `MS`/`ci`; `classify` and `selftest` are pure; `survey_one` and `main` each call `_need_clang()` first. Confirmed empirically — `--selftest` ran 12 cases and exited 0 on a host with no `clang` module installed at all.

**`lambda_lines` nesting is right** (`:96-123`). The `return` after appending is the correct semantic: the outer span already covers the inner lambda, and siblings still get separate spans because the walk descends through non-lambda children. On `s <= merged[-1][1] + 1` — it is *arithmetically neutral* for a distinct-line count, not an over-merge. Abutting spans [90,100] and [101,110] give 21 merged and 11+10=21 unmerged. It can only affect a span count, which is not used.

## Blocking 1 — the lambda share is measured over the enclosing NAMESPACE, not the dominant function (`:189-193`)

`tu.cursor.get_children()` yields *top-level* cursors. Nearly every file here is one `namespace prosper { ... }`, so the first child whose extent contains `big["decl_line"]` and satisfies `is_definition()` is the **namespace** — which is a definition. `lambda_lines` then unions every lambda in the whole file and divides by the dominant region's length.

Measured on the real tree:

| file | cursor the loop picks | true lambda lines in the dominant fn | what that cursor yields |
| --- | --- | --- | --- |
| `rdna2_emit_alu.cpp` | `NAMESPACE prosper` 33-8444 | 880 (10.8%) | 881 |
| `live_renderer.cpp` | `NAMESPACE prosper` 86-11450 | 10,067 (98.3%) | 10,129 (98.9%) |
| `gpu_executor.cpp` | `NAMESPACE prosper` 71-12058 | 313 (11.4%) | 1,024 (**37.4%**) |

The current 28-file table survives this only because in a *dominated* file the namespace is nearly the dominant function, and `gpu_executor.cpp` never reaches the lambda branch (23% share, so `lam` stays 0). That is luck, not robustness — and per the charter's positive-control rule, a control drawn from the same source tests the discriminator, not the domain. So I built one by hand, outside the tree: 1,002 lines, one `namespace prosper`, a 598-line dominant function with **zero** lambdas plus one lambda-dense sibling.

```
biggest: dominant_region, biggest_lines: 598, biggest_share: 0.597,
lambda_lines: 378, lambda_share_of_biggest: 0.632, verdict: "HAND"
```

Ground truth for `dominant_region` is `lambda_lines=0`, i.e. `EXTRACT`. The tool answers **HAND** — "no tool in this repo or in clang can shrink it" — about a function containing no lambda at all. That is precisely the datum the docstring at `:19-20` says the third column exists for.

Fix: match the region's own cursor instead of any containing top-level one — descend through namespaces and require `c.extent.start.line == big["decl_line"]`, or restrict to function-like kinds.

## Blocking 2 — `UNPARSED` only sees a TRAILING span, and the tree already holds the counter-instance (`:169-171`)

The detector counts regions with `kind in ("TEXT", "TRAILER")`. In `map_symbols.py`, `mk()` (`:153`) is called with `c=None` at exactly one site — `:195`, `mk(pos, total_lines, "close", None, kind="TRAILER")` — so the `"TEXT"` arm at `:157` is unreachable and the tuple is effectively `("TRAILER",)`. A `TRAILER` is emitted only for lines after the *last* parsed cursor. **An inactive `#if` arm anywhere but the end of the file is not a TRAILER**: the tiling runs each region from one past the previous end through the next cursor's end, so the unparsed span is absorbed into a *body* region and inflates it.

`hle_kernel_mem.cpp`'s arm runs `#else` at 3730 to `#endif` at **7637, the last line**, so it lands as a TRAILER. The threshold was calibrated on the one instance whose shape the detector can see.

The tree has a mid-file one, and it is already in this PR's published table:

```
SPLIT   2513   484   19%   0%   src/host/memory/guest_write_watch.cpp
                                sched.h
```

`#ifdef _WIN32` at line 71 to `#else` at 551: a 480-line Windows arm, **19% of the file**, mid-file. The tool reports `unparsed_lines: 0` and the region as `70-553 (484L) role=body kind=INCLUSION_DIRECTIVE name='sched.h'` — the "dominant symbol" is an `#include`. 19% is above the PR's own `UNPARSED_SHARE = 0.15`, so by the tool's own criterion this file should be `UNPARSED` and is not.

Scanning every preprocessor arm at or above 15% across the 28 surveyed files, exactly two carry a span clang never sees on Linux: `hle_kernel_mem.cpp` (3,909 / 51%, terminal, caught) and `guest_write_watch.cpp` (480 / 19%, mid-file, missed). `hle_file.cpp`'s large arm is `#ifndef _WIN32`, *active* on Linux; `exec_image_linux.cpp`'s 100% arm is likewise the active one. So the README's **`UNPARSED | 1` should be 2**, and `SPLIT` 11 rather than 12.

A hand-built mid-file instance shows the verdict can be actively wrong rather than merely missing — 2,050 lines with a 2,000-line inactive `_WIN32` arm in the middle:

```
biggest: after_0, biggest_lines: 2005, biggest_share: 0.978,
unparsed_lines: 0, verdict: "EXTRACT"
```

A one-line function reported as a 2,005-line dominant region at 98%. That is the failure the docstring at `:22-27` says the tool exists to prevent, pointing the other way: not a file that quietly looks like nothing to do, but one that quietly looks like a giant function.

This also bears on the #3503 note. The conclusion that a platform-arm split needs a textual line-range move is right — but it applies to `guest_write_watch.cpp` too, and the survey does not currently say so.

Answering the three questions directly:

1. **15% is not defensible from one data point, though that is not the part I would change.** With the detector fixed the tree yields two points (19% and 51%), and the largest active-side scaffolding gap sits well below — so the band is wide and 15% sits inside it. State it as a convention with both observations named, rather than as a measurement.
2. **Yes, the tree has one** — `guest_write_watch.cpp`. Your instinct that the *individually*-≥15% test is wrong is correct, but the sharper problem is that a mid-file span never becomes a candidate at all, whatever the test.
3. **UNPARSED-first is right and I would not change it.** Reporting a lambda share for a file whose other half is invisible is worse than reporting nothing.

## Non-blocking

- **The selftest does not pin two of the four thresholds.** Measured by mutating a copy outside the worktree and re-running `--selftest`: `DOMINANT_SHARE` 0.40 → **0.20** and → **0.15** both still PASS; `SPLITTABLE_REGIONS` 4 → **8** and → **3** both still PASS. `LAMBDA_SHARE` 0.50 → 0.49 and `UNPARSED_SHARE` 0.15 → 0.14 both correctly fail. So the two boundaries the PR says were got wrong are pinned tight, and the two that were never wrong are not pinned at all. A `(10000, 3000, 0, 8)` case pins dominance from below; `(10000, 300, 0, 3)` pins the region count.
- **`:212`** — the case-table comment reads `# (total, biggest, lambda, big_regions, expected)`, but `classify`'s own docstring (`:129`) says the fourth argument is the total region count, "not a count of large ones", and `:205` passes `len(regions)`. This is the same comment-vs-behaviour disagreement the PR describes fixing for the lambda boundary, recurring one field over — and `big_regions` is a real, different local at `:196`.
- **`classify` counts regions on a different basis than it measures them.** `sized` (`:172`) excludes `MS.REPLICATED` roles, but the `len(regions)` passed to `classify` includes preamble/open/close/trailer. In a one-namespace file that inflates the count by 2-3, so `SPLITTABLE_REGIONS = 4` is in practice "about two real bodies".
- **`biggest > 0` at `:139` is dead** — unreachable given `total > 0` and `share >= 0.40`.
- **`lambda_lines` over-counts by up to one line per span** where a lambda's first or last line also carries non-lambda code. Bounded at roughly 10% relative for `emit_alu`'s 90 spans, and it biases toward `HAND` against a hard 0.50 cut.
- **README: one missing blank line.** The new section ends `(-DPROSPER_APP=ON); configure with it to bring that file into the survey.` and the pre-existing `**The invariant that makes incremental extraction safe...**` follows immediately, so the two render as a single paragraph.

Everything above reproduces from this head. The two fixtures are about 40 lines each and I can hand them over if useful.

mattias800:
## Author verification at head `18ae271f9` — both blocking findings fixed

Both findings were confirmed independently before fixing, and both were right. The second was **understated**.

### Blocking 1 — lambda share measured over the namespace

Confirmed directly rather than by re-running the tool: on a hand-built file, the *only* top-level cursor is `NAMESPACE prosper` spanning 5–117, reporting `is_definition=True`. The scan could not have found anything else.

Fixed with `region_cursor()`, which descends exactly as `regions_of` tiles.

**Measured across the tree** (pre-fix code run over all 28 files, diffed against fixed):

| file | fixed | pre-fix |
| --- | --- | --- |
| `rdna2_emit_cfg.cpp` | **0.258** | 0.428 |
| `render_runner.h` | 0.143 | 0.273 |
| `live_compute.cpp` | 0.097 | 0.190 |
| `exec_image_linux.cpp` | 0.053 | 0.102 |
| `rdna2_to_spirv_internal.hpp` | 0.063 | 0.091 |
| `live_renderer.cpp` | 0.983 | 0.989 |

**6 of 22 scored files were inflated; 0 verdicts changed.** Your "survives by luck" reading is exactly right, and now quantified — `rdna2_emit_cfg` at 0.428 was one small edit from crossing 0.50 into "no tool can act on this".

### Blocking 2 — mid-file unparsed spans, and you under-counted

`guest_write_watch.cpp` reproduces as described: `#ifdef _WIN32` at 71, `#else` at 551, `#endif` at 2513.

Rather than widen the region-kind heuristic, the tool now asks the preprocessor: `clang_getSkippedRanges` via ctypes (the Python bindings do not wrap it), **summing every arm**. That second part matters — it found two files neither of us had:

| file | lines | unparsed | was |
| --- | --- | --- | --- |
| `hle_kernel_mem.cpp` | 7,637 | **52%** (7 spans) | UNPARSED at 51% from one span |
| `hle_kernel.cpp` | 5,395 | **30%** | SPLIT |
| `hle_file.cpp` | 4,560 | **21%** | SPLIT |
| `guest_write_watch.cpp` | 2,513 | **20%** | SPLIT |

So `UNPARSED` is **4**, not 2, and `SPLIT` is **9**, not 11. Final: `EXTRACT=13, SPLIT=9, UNPARSED=4, HAND=1, PARSE-FAIL=1`.

It fails **closed** — if the symbol is absent the tool exits rather than reporting zero unparsed lines, since that is the exact silent under-report being fixed.

### Why the arithmetic selftest could never have caught either

Both defects were in how the classifier's *inputs* are measured, not in `classify()`. Every arithmetic case passed while the tool reported the opposite verdict for a real file. Two fixtures under `testdata/` pin them, each built so the **defective measurement yields a different verdict**, not merely a different number — and each mutation arm reddens **only** its own fixture:

| mutation | `dominant_no_lambda` | `midfile_inactive_arm` |
| --- | --- | --- |
| pre-fix top-level cursor scan | **FAIL** — `HAND`, share 0.729 | ok |
| pre-fix region-kind inference | ok | **FAIL** — `EXTRACT`, 0 unparsed, dominant `posix_only_1` |

That second one is your `sched.h` pathology reproduced in miniature: the region that absorbed the skipped arm gets named as the dominant symbol.

The fixtures score **only where libclang is importable**, and `--selftest` says so loudly when it skips them, so CI's arithmetic-only run is never mistaken for the full gate. Verified both ways: host (no libclang) prints the skip banner; `ctest -V` in the container shows `parse fixtures (libclang present)` and both `[ok]` lines.

### Your coordinator questions

1. **15%** — agreed, and now labelled in the source and README as *"a CONVENTION, not a measurement"*, with the four data points and the width of the gap stated.
2. Fixed at the root: a mid-file span is now measured directly, so the individually-≥15% test is gone entirely — spans are summed.
3. UNPARSED-first kept.

### Non-blocking

All four applied: movable-region count (`len(sized)`) instead of `len(regions)`; dead `biggest > 0` dropped; the `big_regions` comment corrected to `movable_regions`; README blank line restored. Threshold-sensitivity note taken — the fixtures now cover the two thresholds the mutation showed were insensitive.

Separately, the survey surfaced a contradiction inside the README itself: `register_live_renderer` was recorded at 8,221 lines / 98.5% lambda from an earlier head, against the survey's 10,244 / 98% today. Same conclusion, different heads — dated rather than reconciled.

### Gates

- `ctest -R refactor_survey_classifier --no-tests=error` — 1/1 passed, fixtures executed (confirmed via `-V`)
- docs numbered-table gate over all `*.md` — clean
- `git diff --check` — clean
- session-trailer gate — 0

mattias800 (commented):
## REJECTED

Not for the classifier — the two findings from the previous review are genuinely fixed, and I
confirmed both by mutation rather than by reading. What blocks is three published quantities that
are wrong in the merged artifact, two of them inside `survey_sizes.py`'s own docstring and comments.
All three are one-line edits.

I built my own worktree at `18ae271f9`, configured cmake there, and ran everything against that.

---

## What I attacked and could not break

**The ctypes glue is sound, not luck.** `TranslationUnit` and `File` are both `ClangObject`
(`ClangObject.__init__` sets `self.obj = self._as_parameter_ = obj`), so ctypes' unprototyped-call
conversion is documented behaviour, not an accident — leaving `argtypes` unset is correct and the
comment at `survey_sizes.py:106-108` states the reason accurately. `ci.SourceRange` is
`(c_void_p*2, c_uint, c_uint)`, `ctypes.sizeof` = **24**, matching `CXSourceRange`'s
`const void *ptr_data[2]; unsigned begin_int_data; unsigned end_int_data;` exactly; `_CXSourceRangeList`'s
`c_uint` + pointer gets the C padding from ctypes' default alignment. The `AttributeError` guard at
`:109-115` is a real fail-closed: ctypes raises `AttributeError` from `CDLL.__getitem__` for an
undefined symbol, so the `sys.exit` is reachable exactly when intended.

**Clipping on `s.file` is more than enough — libclang has already done it.** `clang_getSkippedRanges`
takes the `CXFile` as a *filter*, not as a hint. On `src/hle/fs/hle_file.cpp` it returns 58 ranges,
**58 of them in that file and 0 in any other**; asking the same TU for `/usr/include/stdio.h` returns
19. Include-guard bodies from other headers cannot leak in. (This also makes the Python re-clip pure
downside — see finding 4.)

**`region_cursor` survives the two shapes the fixtures do not cover.** I built both by hand, outside
`testdata/`:

* dominant region is a `struct` with an all-lambda inline member → found the `STRUCT_DECL`,
  `lambda_share=0.943`, verdict `HAND`. Correct.
* five levels of nested namespace → `regions_of`'s `emit` stops descending at emit-depth 3 and emits a
  `body` region for the depth-3 namespace; `region_cursor`'s `walk` stops descending at the same depth
  and matches that namespace's extent. The two `max_depth=3` defaults (`map_symbols.py:126`,
  `survey_sizes.py:199`) are coupled with nothing asserting it — but I checked both mismatch
  directions and **both fail to `NO-CURSOR`**, i.e. loudly, so the coupling is safe-failing.
* C++17 `namespace aa::bb::cc { }` → dominant found correctly (`0.965`, `HAND`).

**`NO-CURSOR` is reachable, and I constructed the trigger.** `region_cursor:219-221` `continue`s past
*every* `NAMESPACE` below `max_depth` without checking its extent, while `regions_of`'s `emit` emits a
`body` region for a namespace that has **no in-target children** (`kids` empty → falls to
`mk(pos, cend, "body", c)`). A namespace whose contents are only comments — or only declarations from
an `#include`d file — is therefore a dominant body region no cursor lookup can match. Constructed
instance scores `NO-CURSOR why="dominant region 3-124 matched no cursor"`, not a silent optimistic
`EXTRACT`. The backstop at `:294-301` earns its place; nothing in the tree triggers it today.

**`len(regions)` → `len(sized)` changed no verdict.** `regions` reaches `classify` only through
`:246` (`return "SPLIT" if regions >= SPLITTABLE_REGIONS else "OK"`, threshold 4). The smallest
movable-region count over the 28 files is **17** (`test_dynfetch_fold.cpp`); the SPLIT files run
125-470. A flip needs `len(regions) >= 4 > len(sized)`, which is unreachable by two orders of
magnitude here.

**Mutation arms — each fixture reddens for its own defect and stays green for the other.** I
re-introduced each pre-fix measurement into the current module and re-scored both fixtures:

| arm | `dominant_no_lambda.cpp` | `midfile_inactive_arm.cpp` |
| --- | --- | --- |
| baseline | `EXTRACT` lam=0.0 | `UNPARSED` unp=62 |
| pre-fix top-level cursor scan | **`HAND` lam=0.729** | `UNPARSED` unp=62 |
| pre-fix region-kind inference | `EXTRACT` lam=0.0 | **`EXTRACT` unp=0** |
| `region_cursor` forced to `None` | **`NO-CURSOR`** | `UNPARSED` unp=62 |

**`--selftest` behaves as claimed in both directions.** With libclang: prints
`-- parse fixtures (libclang present) --` and scores both. Without (host python): prints the explicit
`SKIPPED ... did NOT run. Arithmetic only.` block. `ctest -N` = 454, no duplicate names;
`ctest --no-tests=error -R refactor_survey_classifier` = 1/1 Passed.

**Hygiene.** `git diff --check` clean, one deletion (an edited README line, not a trap-41 revert),
zero `Claude-Session:` trailers.

---

## Independent derivation of the load-bearing numbers

Per the charter I derived the unparsed spans by a route that touches neither
`clang_getSkippedRanges` nor `map_symbols` nor this tool: drive the compiler **driver** with
`-E -C` and reconstruct, from the linemarkers, which source lines of the file produced output
(`-C` keeps comments so a comment block is not mistaken for a skipped arm). Run under **two
independent preprocessor implementations**:

| file | tool | `clang++ -E -C` | `g++ -E -C` |
| --- | --- | --- | --- |
| `hle_kernel_mem.cpp` | 3978 / 52% | 4001 / 52.4% | 4012 / 52.5% |
| `hle_kernel.cpp` | 1613 / 30% | 1567 / 29.0% | 1577 / 29.2% |
| `hle_file.cpp` | 963 / 21% | 870 / 19.1% | 882 / 19.3% |
| `guest_write_watch.cpp` | 495 / 20% | 502 / 20.0% | 509 / 20.3% |
| `gpu_executor.cpp` (control) | 93 / 1% | 114 / 0.9% | 129 / 1.1% |
| `live_renderer.cpp` (control) | 31 / 0% | 62 / 0.5% | 55 / 0.5% |

And the lambda shares with my own `clang.cindex` script that never imports the tool:
`emit_cfg_state_machine` = 1251 lambda lines over 4849 → **0.258**; the first top-level cursor
matching the pre-fix predicate is `NAMESPACE prosper 32-7589` carrying 2075 lambda lines →
**0.428**. Both README figures reproduced exactly, and the diagnosed mechanism confirmed.
`emit_alu` = 880/8185 → **0.108** (README: 11%).

Re-running the survey at head reproduces the counts exactly: **28 files, EXTRACT=13, HAND=1,
PARSE-FAIL=1, SPLIT=9, UNPARSED=4**. Diffing pre-fix against post-fix rows: exactly **6** rows change
lambda share, **0** verdicts change from that fix, and `UNPARSED` goes 1 → 4 with the three
SPLIT→UNPARSED moves being `hle_file.cpp`, `hle_kernel.cpp`, `guest_write_watch.cpp`. All as claimed.

---

## Blocking

**1. `survey_sizes.py:34` states a number the committed code does not produce.**
*"which the old code reported at a 0.63 lambda share"*. Reconstructing the pre-fix scan exactly from
the `d7be17296..18ae271f9` diff and running it on the committed fixture gives **0.729**, not 0.63 —
43 lambda lines (the fixture's only lambda, lines 71-113) over the region's **59** lines
(`region[1] 10-68 body FUNCTION_DECL dominant`, `biggest_lines: 59`). 0.63 is `43/68`, i.e. the share
against the function's raw extent rather than against the region the tool actually divides by. The
*verdict* claim (`HAND`) is right and the fixture is a good discriminator; the number is not. Fix it
to 0.729 or drop the figure.

**2. `README.md` — "6 of 22 scored files carried an inflated share" uses a denominator that excludes
one of its own numerator members.** The six are `live_compute.cpp` (0.19→0.097),
`live_renderer.cpp` (0.989→0.983), `rdna2_emit_cfg.cpp` (0.428→0.258),
`rdna2_to_spirv_internal.hpp` (0.091→0.063), `exec_image_linux.cpp` (0.102→0.053),
`render_runner.h` (0.273→0.143). `22` = EXTRACT(13) + SPLIT(9) — which does not include
`live_renderer.cpp`, the `HAND` file. Worse for the reader: the 9 SPLIT files **cannot** be inflated,
because the lambda share is computed only inside `if big_len / total >= DOMINANT_SHARE` (`:288`) in
the new code *and* in the old, so they pad the denominator with files the defect could never touch.
The honest denominators are **14** (files for which a share is computed at all: 13 EXTRACT + 1 HAND)
or **23** (non-`UNPARSED`, non-`PARSE-FAIL`). 6/14 = 43% is the rate; 6/22 = 27% understates the very
thing the paragraph is arguing.

**3. `survey_sizes.py:126-128` — the comment justifying `UNPARSED_SHARE` contradicts the measurement
taken in its own commit.** *"the tree has two files above it (guest_write_watch.cpp at 19.7%,
hle_kernel_mem.cpp at 52.1%)"*. The same commit's run has **four**: 52.1, 29.9, 21.1, 19.7. The README
was updated to say four; this comment was not, and it is the comment a reader hits first when asking
why 0.15.

On the question itself — is "a CONVENTION, not a measurement" honest given four data points, the
lowest at 20%? **Yes, and the honest evidence is stronger than what the comment offers.** The highest
unparsed share among the 24 non-`UNPARSED` files is `exec_image_linux.cpp` at **4.3%** (173/3990);
everything else is under 2.2%. So every threshold in **(4.3%, 19.7%]** produces the identical
partition and the tree cannot distinguish 0.15 from 0.05 or from 0.19. That is the sentence to write.

---

## Should fix (small, and each is in a family this project already tracks)

**4. `skipped_lines` has two silent-zero returns in the one function whose stated job is not to
under-report.** `survey_sizes.py:162-167`. `if f is None: return 0` reports "no unparsed lines" for a
file that is **not in the TU at all** — indistinguishable from a genuine zero. That is reachable on
the header path: `flags_for` picks a parse TU by the *textual* needle `f'/{target.name}"'`
(`map_symbols.py:78-87`), so a header whose only `#include` sits behind an inactive `#if` is
"surveyed" through a TU that never included it, and lands as `unparsed=0` plus an empty region list —
i.e. verdict `OK`, "nothing to do here", which is exactly the hiding place `PARSE-FAIL` exists to
close. Make it `PARSE-FAIL`, not `0`.

The second is the re-clip at `:172`. Since libclang already filters by `FileID` (measured above), the
`Path(s.file.name).resolve() != path.resolve()` test adds no protection and one failure mode, which I
reproduced: `skipped_lines(tu, <the /var/home alias spelling of hle_file.cpp>)` returns **0** where
the `/home` spelling returns **963** — `File.from_name` resolves the alias fine, then every span is
discarded because `resolve()` cannot unify a bind-mount alias. The normal CLI path is safe on this box
only because `flags_for` exits first on a mismatched cwd. Deleting `:172-173` removes the hazard and
loses nothing.

**5. `PARSE-FAIL` / `NO-CURSOR` rows print the developer's absolute host path.** `:482` prints
`r['file']` raw; the `rel = r["file"].split("prosper/", 1)[-1]` strip at `:485` is only on the other
branch. My run emitted **two** absolute host paths on that single row (the path, plus the one embedded
in `why`). The charter's privacy rule is explicit, and a `PARSE-FAIL` row is precisely what an author
pastes into an issue. Hoist `rel` above the branch.

**6. A reported span includes the `#if`/`#else`/`#endif` lines themselves.** Each recorded range runs
directive-line to directive-line — e.g. `hle_file.cpp` `1567-1604` is `#else`…`#endif`, 36 skipped
code lines reported as 38 — so the count is a systematic **+2 per span**. Immaterial where spans are
few (`hle_kernel_mem.cpp`: 7 spans, +14 of 3978) but `hle_file.cpp` has **58**, so its published
"21%" is ~18.6% of actual code, which is what the two independent preprocessors measure (19.1/19.3%).
No verdict moves — all four stay above 0.15 — but the docstring at `:157` should say what the number
counts.

**7. In CI the registered case passes on a strictly smaller domain and nothing in the artifact says
so.** `.github/workflows/ci.yml` installs neither `libclang` nor `python3-clang` (`apt-get install`
lines 22, 45, 81, 586, 630), so `refactor_survey_classifier` runs arithmetic only — and ctest hides
stdout on a pass, so the deliberate "SKIPPED" notice is invisible in exactly the place it was written
for. The lazy-libclang design is right and the CMakeLists comment is honest about it; the stronger
shape is a second case (`refactor_survey_fixtures`) with `SKIP_RETURN_CODE`, so the summary reads
`(Skipped)` rather than a full pass. This is the charter's CI-domain rule, and this PR is otherwise
scrupulous about exactly that distinction.

---

## Informational — checked, no change requested

* **`clang_disposeSourceRangeList` is never called, and it does not matter.** Quantified rather than
  waved at: ≤58 ranges × 24 B + 8 B ≈ 1.4 KB per file, ~40 KB across the 28-file survey, against TUs
  that are hundreds of MB and *are* freed by cindex's `__del__`. No use-after-free either — each range
  is copied out of the array by ctypes before the list is dropped. Worth one comment saying it was
  considered, not worth the code.
* **`_need_clang()` is not thread-safe.** It sets `MS` — the guard — at `:95`, before `ci` at `:97`
  and `_SKIPPED_RANGES` at `:116`, so a second thread arriving in between passes the guard and runs
  with `ci = None`. Unreachable from `main()` (which calls it at `:447` before the pool), but
  `measure`/`survey_one` are public and `--jobs` defaults to 6. Assigning `MS` last closes it.
* **`sized` includes `preamble` regions.** `MS.REPLICATED` is `("open", "close")`
  (`map_symbols.py:123`), but `regions_of` re-roles everything before the first namespace open to
  `"preamble"` (`map_symbols.py:202-205`), which that module's own docstring says "carries the same
  replicate role as a namespace open/close". So the "Movable regions only" comment at `:269-270` is a
  little stronger than the filter. 0-2 regions against a threshold of 4 in a tree whose minimum is 17
  — no verdict is affected.
* **`lam%` prints `0%` for SPLIT rows where the share was never computed** (`:288` gates it). Reads as
  a measured zero in the table. The README does not rely on it.

---

The tool itself is right, the two prior defects are properly fixed, the fixtures are honest
discriminators, and the failure the fixtures cannot express (`NO-CURSOR`) is handled loudly rather
than optimistically. Fix the three numbers and I would approve on sight; 4 and 5 are two deletions
and a hoisted variable and I would rather they went in the same push.

mattias800 (commented):
## Review round 3 — head `9bec45c64`

**REJECTED.** One blocking finding and two wrong published figures. Everything round 2 raised that I
could re-derive independently checks out — the band, the denominator, the inflation rate, the inode
fix, the third fixture and its mutation, and the unchanged survey output are all confirmed below by
routes that do not re-run the tool. The blocker is new: the *one* safety property this round added
does not exist, and the README states it as fact.

Method: own worktree at the PR head, fresh `cmake -S prosper -B prosper/build-linux` configure, all
parsing inside the `ps5ys` container (libclang 22.1). Independent measurements come from a
separately written ctypes binding to `clang_getSkippedRanges` with its own `(st_dev, st_ino)` clip
and an exhaustive AST walk for the dominant cursor, so they share no code with `survey_sizes.py`
beyond `map_symbols.regions_of` and `flags_for`.

---

### BLOCKING 1 — the `NO-PREPROC` guard is dead, and README:207 publishes a fail-closed guarantee that does not hold

`README.md:205-207` states:

> it fails **closed** everywhere the question cannot be answered ... A missing
> `clang_getSkippedRanges` exits; a file absent from its own TU, or a NULL range list, scores
> `NO-PREPROC` rather than 0.

The first of those two cases is false. `survey_sizes.py:189-192`:

```python
    f = ci.File.from_name(tu, str(path))
    if f is None:
        # Never 0. "The file is not in its own translation unit" is not "this file skips nothing";
        raise SkippedRangesUnavailable(f"{path} is not a file of its own translation unit")
```

`clang.cindex.File.from_name` never returns `None`. It is
`File(conf.lib.clang_getFile(tu, name))` (cindex.py:3704-3708) and `ClangObject.__init__` asserts
the handle is non-NULL (cindex.py:3013). So the branch cannot be taken, and the two things that
actually happen are:

* **A real file that is not in the TU returns a valid `File`.** `clang_getFile` resolves through the
  file manager, so TU membership is never tested; `clang_getSkippedRanges` then yields an empty list
  and `skipped_lines` returns **0**. Measured: calling `measure()` on `src/gpu/pm4/command_processor.cpp`
  with `parse_target = src/hle/fs/hle_file.cpp` returns
  `{'lines': 5305, 'verdict': 'OK', 'regions': 0, 'unparsed_lines': 0}` — a 5,305-line file scored
  "nothing to do", with the silent zero the docstring at `survey_sizes.py:190-191` says it exists to
  end.
* **A nonexistent path raises `AssertionError`**, which `measure()`'s
  `except SkippedRangesUnavailable` (`survey_sizes.py:315`) does not catch, so it would escape
  `survey_one` and abort `main()`'s `ex.map` loop.

Neither is reachable on today's tree — the candidate list comes from `rglob`, so paths always exist,
and all three surveyed headers really are in the TU `flags_for` picks. But the path is live in
general: `flags_for` chooses a header's TU by the text match `needle = f'/{target.name}"'`
(`map_symbols.py:76-86`), which does not check that the include is *active*. A header included only
under an inactive `#if`, or merely named in a comment or string, scores `unparsed=0` and — with no
cursors of its own in that TU — falls to `sized` empty and `verdict: OK` at `survey_sizes.py:328-330`.
That is precisely the "silently ranked as needing no work" outcome the module docstring calls the
worst possible place for a failure to hide.

Either make the guard real (test membership rather than existence — e.g. against the TU's include
set, or by requiring the file to appear among the TU's own `File`s) or drop the README sentence to
what the code does. As written, the claim is the kind that gets inherited rather than checked.

### BLOCKING 2 — `README.md:180`: the third fixture is 95.2% of its file, not 98.9%

> a namespace whose body is all comments (`testdata/namespace_without_children.cpp`, 98.9% of its file)

Measured, from the tool's own row for that fixture: `lines: 105`, `biggest_lines: 100`,
`biggest_share: 0.952`. The region tiling is `1-100 role=preamble name='commentary'`, i.e. 100 of
105 lines = **95.2%**. No natural reading of the fixture gives 98.9% of the file: the namespace's own
extent (9-100) is 87.6%, comment lines are 93.3%, comments+blank 94.3%. (90 of the 91 lines 10-100
are notes = 98.9%, which is probably the origin, but that is not "of its file".)

### BLOCKING 3 — `README.md:158-159`: the `midfile_inactive_arm` mutation does not name an include directive

> `midfile_inactive_arm` -> `EXTRACT` with 0 unparsed, naming the include directive that swallowed the arm

I ran that mutation (revert `skipped_lines` to the pre-fix region-kind inference). It reddens its own
fixture and only its own, as claimed — but the row is:

```
{'lines': 91, 'regions': 16, 'biggest': 'posix_only_1', 'biggest_lines': 64,
 'biggest_share': 0.703, 'unparsed_lines': 0, 'verdict': 'EXTRACT'}
```

`posix_only_1` is a `FUNCTION_DECL`, not an include directive. The `sched.h` detail belongs to
`guest_write_watch.cpp` (correctly stated at README:189-191) and has been carried onto the fixture,
where it is not true. The verdict half of the sentence is right; delete or correct the "naming the
include directive" clause.

---

### Confirmed — everything below was re-derived, not taken on the PR's word

**Finding 3, the band. Correct, and it survives a wider threshold than the one it was measured at.**
Re-measured with my own binding and clip over all 28 candidates at `--min-lines 2500`, including the
three headers, which need `flags_for`'s including-TU fallback:

| file | unparsed share |
| --- | --- |
| `src/hle/memory/hle_kernel_mem.cpp` | 52.09% |
| `src/hle/kernel/hle_kernel.cpp` | 29.90% |
| `src/hle/fs/hle_file.cpp` | 21.12% |
| `src/host/memory/guest_write_watch.cpp` | **19.70%** |
| `src/host/image/exec_image_linux.cpp` | **4.34%** |
| `src/hle/service/hle_service.cpp` | 2.22% |
| (everything else, incl. all three headers) | <= 0.80% |

So `survey_sizes.py:132-133` is right to the second decimal on both ends, and `(4.34%, 19.70%]` is
empty. I also ran the tool's **default** `--min-lines 1500` (39 candidates): still zero files inside
the band. The one thing the default adds is a fifth caught file, `src/host/image/exec_image_win.cpp`
at 99.18% — so "the four it does" is scoped to 2,500 lines while the comment says only "measured
over the tree". Worth a clause; the partition claim itself is unaffected, since the gap is the same
at both thresholds.

**Finding 2, the denominator of 14. Correct, and correct for the reason given.** `lam` is computed
only inside `if big_len / total >= DOMINANT_SHARE` in both versions — post-fix at
`survey_sizes.py:336-341`, pre-fix at `d7be17296:186-193` — so a `SPLIT` file structurally cannot
carry a share, in either version. 14 = `EXTRACT` 13 + `HAND` 1, which my own dominance computation
reproduces independently (12 dominated `.cpp` + `rdna2_to_spirv_internal.hpp` at 0.814 and
`render_runner.h` at 0.527).

**"6 of the 14 ... carried an inflated one", worst 0.428 vs 0.258. Correct.** A/B through the tool's
own code with *only* `region_cursor` swapped for the pre-fix top-level scan:

```
live_compute.cpp                     0.097 <- 0.190   INFLATED
live_renderer.cpp                    0.983 <- 0.989   INFLATED
rdna2_emit_cfg.cpp                   0.258 <- 0.428   INFLATED
rdna2_to_spirv_internal.hpp          0.063 <- 0.091   INFLATED
exec_image_linux.cpp                 0.053 <- 0.102   INFLATED
render_runner.h                      0.143 <- 0.273   INFLATED
rdna2_emit_alu.cpp                   0.108 <- 0.108
... 7 others unchanged
files with a lambda share: 14; inflated: 6
```

One nuance worth a footnote rather than a change: **7** of the 14 differ in raw lambda-line count —
`rdna2_emit_alu.cpp` is 880 vs 881 over an 8,186-line region, which rounds to 0.108 either way. At
the precision the tool reports, 6 is the right number. "No verdict actually changed" also holds:
0.428 is the closest any pre-fix share came to the 0.50 boundary.

**Finding 1, the 0.729 fixture figure. Correct, by two routes.** By hand: the `dominant` region is
lines 10-68 = 59 lines, the lambda is 71-113 = 43 lines, 43/59 = 0.7288. And by running the pre-fix
`region_cursor` against the fixture, which reports exactly
`lambda_lines: 43, lambda_share_of_biggest: 0.729, verdict: HAND` — and reddens only that fixture.

**The inode fix. Verified, including the bind-mount question.** Both spellings exist in the container
and `resolve()` does not collapse them; `os.stat` gives `st_dev=60, st_ino=25436497` for both, and
`os.path.samefile` is True. `os.path.samefile` -> `os.path.samestat` compares `st_ino` **and**
`st_dev`, and a bind mount shares the superblock, so `st_dev` is preserved — the fix is correct for
this alias rather than accidentally correct. Driving `skipped_lines` on `hle_file.cpp` under all four
spelling combinations:

```
path=/home     target=/home        inode=963   resolve=963
path=/var/home target=/home        inode=963   resolve=0
path=/var/home target=/var/home    inode=963   resolve=963
path=/home     target=/var/home    inode=963   resolve=0
```

963 under both spellings, 0 under the old comparison when they disagree. Exactly as claimed.

**The third fixture and its mutation.** Reverting `survey_sizes.py:262-267` to `depth < max_depth`
alone gives `verdict: NO-CURSOR, why: 'dominant region 9-100 matched no cursor'` on
`namespace_without_children.cpp` and leaves the other two green. Each of the three mutations reddens
its own fixture and only its own — I ran all three.

**Survey output unchanged.** The tool at `--min-lines 2500` prints
`EXTRACT=13, HAND=1, PARSE-FAIL=1, SPLIT=9, UNPARSED=4`, no `NO-CURSOR` and no `NO-PREPROC` rows.
My independent classifier over the same 28 files agrees file-for-file. `hle_file.cpp` at 0 skipped
lines would indeed have been `SPLIT` (biggest share 6.7%, 298 movable regions).

**Registered tests.** In the container both cases pass; on the host, which has no `clang` module,
`refactor_survey_classifier` **Passed** and `refactor_survey_measurement` **Skipped** — the
`SKIP_RETURN_CODE 77` wiring does what it says. `ctest -N` shows no duplicate names.

**Hygiene.** No `Claude-Session:` trailers, `git diff --check` clean, and the only host paths added
are `<user>` placeholders in prose.

---

### Should fix (not blocking)

4. **The inode fix has no regression arm.** Reverting `_same_file` to
   `pathlib.Path(a).resolve() == b.resolve()` leaves all three fixtures **green** — measured. The
   963/963 result is a hand measurement that nothing preserves. A fixture reached through a symlink
   (parse the real path, score the symlinked one) would pin it and costs a few lines.
5. **The skip message names two fixtures when there are three.** `survey_sizes.py:431` "Score the two
   fixtures", and `:508-509` "the two measurement checks (dominant-cursor, mid-file inactive arm)" —
   which is the text a runner without libclang actually prints, so the namespace-descent check
   silently vanishes from the report that exists to say what did not run.
6. `README.md:151` says "Neither produced an error" immediately after announcing three ways.

### Informational

7. **`sized` still counts `preamble` as movable.** `survey_sizes.py:319` filters only
   `MS.REPLICATED = ("open", "close")`, while `map_symbols.py:198-205` says the preamble "carries the
   same replicate role" — a split gives every output part a copy of it, so it is not a place a cut
   can happen. No published number moves: I checked all 28 rows and no file's dominant region is a
   preamble, and no region count is near `SPLITTABLE_REGIONS`. But note that
   `namespace_without_children.cpp`'s dominant region **is** a preamble region
   (`1-100 role=preamble name='commentary'`), so tightening this filter would flip that fixture to
   `OK` and redden it for an unrelated reason. Worth a line in the fixture so the next person does
   not read that failure as a regression in `region_cursor`.
8. **`_same_file`'s fail-loud fallback covers stat *failure*, not stat *disagreement*.** Two
   superblocks over the same bytes (overlayfs, an NFS re-export) return a clean `False` with no
   `OSError`, and the silent under-count is back. Conversely, a relative `s.file.name` would stat
   against the cwd and, on failure, fold every foreign header guard into this file's total — loud, as
   designed, but worth knowing that the loud direction is a cliff rather than a nudge.
9. **`_need_clang()` is still not thread-safe** (`survey_sizes.py:100-102` publishes `MS` before `ci`
   and `_SKIPPED_RANGES`). `main()` calls it before building the pool, so it is unreachable today;
   unchanged from round 2 and fine to leave.

mattias800 (commented):
## Review round 4 — head `40ce2d0ea`

**REJECTED**, on one finding, and it is not in the code: **the PR body still publishes the round-1
measurement**, including a claim this PR's own README documents as falsified. Everything round 3
blocked on is genuinely fixed, and I confirmed each fix by construction rather than by reading. If
the body is brought up to the committed artifact I would approve on sight — nothing in the diff
needs to change.

Method: own worktree at the PR head, fresh `cmake -S prosper -B prosper/build-linux` configure, all
parsing inside the `ps5ys` container (libclang 22.1). The probes below are separately written and
call into the module rather than re-running the CLI, so a wrong answer from the tool cannot launder
itself through my check.

---

## BLOCKING — the PR description describes a tool that no longer exists

The body is the round-1 text. It predates commit `d7be17296`, which added the `UNPARSED` verdict, and
has not been touched since. Seven statements in it are now false against the committed script, which
I ran at `--min-lines 2500` on this head:

```
[survey] EXTRACT=13, HAND=1, PARSE-FAIL=1, SPLIT=9, UNPARSED=4
```

| body says | committed tool says |
| --- | --- |
| "reports which of **three** states every large file is in", with a 3-row SPLIT/EXTRACT/HAND table | four verdicts (`UNPARSED` is the fourth) plus `PARSE-FAIL`, `NO-CURSOR`, `NO-PREPROC` |
| "28 files at or above 2,500 lines: `SPLIT` **13**, `EXTRACT` 13, `HAND` 1" | `SPLIT` **9**, `EXTRACT` 13, `HAND` 1, `UNPARSED` **4**, `PARSE-FAIL` 1 |
| `live_renderer.cpp` at "**99%** lambda" | `lambda_share_of_biggest: 0.983` → 98%, which is also what `README.md:134` says |
| "The survey names **26** files with a route" | 22 (13 `EXTRACT` + 9 `SPLIT`); the 4 `UNPARSED` files explicitly have none |
| "The classifier has been wrong **twice**" | `CMakeLists.txt:172-176`, added by this PR, says **five** times |
| `--selftest` "# **8** cases" | 13 arithmetic cases plus 4 measurement checks |
| `ctest -N` "# **454** tests"; `-R refactor_survey_classifier` "# 1/1 Passed" | `ctest -N` = **455**; the PR registers **two** cases, and `refactor_survey_measurement` is the one carrying the CI-domain honesty the PR argues for |

The one that makes this blocking rather than tidying is this line:

> - **The two largest source files are plain `SPLIT` work** — `gpu_executor.cpp` (12,058 lines,
>   biggest region 23%) and `hle_kernel_mem.cpp` (7,637 lines, biggest region **2%**, the subject of
>   #3503).

`README.md:238-249` — added by this PR — says that exact reading was wrong and says why: *"The first
version of this tool scored `hle_kernel_mem.cpp` as `SPLIT` from the 48% it could see"*, and draws
the opposite practical conclusion for the same issue: *"The practical consequence for #3503: that
split cannot be driven by `split_file.py`, because the tool is blind to the arm being moved."* The
committed tool scores that file `UNPARSED`, 3,978 lines / 52.1%. So the body sends a reader to #3503
with the instruction the README exists to prevent, under this PR's own authority.

This is the same species as rounds 2 and 3 — a figure carried over from a draft rather than read off
the committed artifact — in the surface a reader hits first. Note `gh pr edit` fails silently in this
repository on a Projects-classic GraphQL error; use `gh api -X PATCH repos/.../pulls/3514 -f body=...`
and re-read the body to confirm it took.

---

## Round 3's three blocking findings — all fixed, verified by construction

**Blocking 1, the dead `NO-PREPROC` guard. Fixed, and the new guard both fires and does not
over-fire.** I built a TU (`main.cpp` including `direct.hpp`, which transitively includes
`trans.hpp`) plus two files outside it: `foreign.cpp` carrying an 80-line inactive arm, and
`dead.hpp` included only under `#ifdef NEVER_DEFINED`.

```
_file_in_tu(main.cpp)   = True    old guard File.from_name is None -> False   skipped_lines -> 3
_file_in_tu(direct.hpp) = True    old guard ... -> False                      skipped_lines -> 0
_file_in_tu(trans.hpp)  = True    old guard ... -> False                      skipped_lines -> 42
_file_in_tu(dead.hpp)   = False   old guard ... -> False                      NO-PREPROC
_file_in_tu(foreign.cpp)= False   old guard ... -> False                      NO-PREPROC
```

The old predicate is `False` on every row, including the two the guard must reject — so it could not
fire, exactly as round 3 said, and `foreign.cpp`'s 82 skipped lines would have come back as a
confident 0. `trans.hpp` at **inclusion depth 2** answers the transitive question: `clang_getInclusions`
walks the whole inclusion tree, not just direct includes.

**It rejects nothing legitimate.** Four include forms all pass: behind an *active* `#if`, through a
macro (`#define HDR "mac.hpp"` / `#include HDR`), through `#pragma once` on a second inclusion, and
transitively. The three real headers all score, and I checked the mechanism rather than just the
outcome — each is at depth 1 in the TU `flags_for` picks:

```
gpu_execute.hpp             via agc_shader_layout.cpp   in_tu=True  depths=[1]  (442 includes)
rdna2_to_spirv_internal.hpp via rdna2_emit_alu.cpp      in_tu=True  depths=[1]  (402 includes)
render_runner.h             via live_renderer.cpp       in_tu=True  depths=[1]  (608 includes)
```

`get_includes()` also does **not** depend on `PARSE_DETAILED_PROCESSING_RECORD` — measured, 1 include
either way — so the guard does not quietly rest on a parse option a future caller might drop.

**And the "assumption into a check" reasoning is sound.** The only new rejection is a file the chosen
TU never read, which under the old code produced `unparsed=0` and, with no cursors of its own,
`verdict: OK` — the silent "nothing to do here" the module docstring calls the worst hiding place. A
loud `NO-PREPROC` is strictly better. No file in the tree takes that branch.

**Blocking 2, the third fixture's share. Fixed and correct.** From `regions_of` on the committed
fixture:

```
[4] 18-110  93L  role=body  kind=NAMESPACE  name='commentary'  share=0.8455
```

93 of 110 lines, 84.5% — matching `README.md:183`. It is a `body` region, not a `preamble`, so the
restructure did what it claims; and reverting `region_cursor` to descend on `depth < max_depth` alone
still gives `verdict: NO-CURSOR` on it, so it still discriminates.

**Blocking 3, the `midfile_inactive_arm` mutation. Fixed and correct.** With `skipped_lines` reverted
to the region-kind inference, the dominant region is `[2] 10-73 body FUNCTION_DECL posix_only_1`,
70.3% → `EXTRACT` with `unparsed=0`. `posix_only_1` is a function declaration, as `README.md:157-158`
now says, and the `sched.h` detail stays where it belongs at `README.md:191-194`.

---

## The should-fix items

**The inode regression arm. It works, and a hard link is genuinely equivalent to the alias case.**
Measured in the container, which is where the alias lives: `/home` is a real directory there (a bind
mount), `/home/.../CLAUDE.md` and `/run/host/var/home/.../CLAUDE.md` both stat as `st_dev=60
st_ino=23511324`, `resolve()` keeps them distinct, and `os.path.samefile` is `True`. A hard link
reproduces exactly that pair of properties — same `(st_dev, st_ino)`, distinct `resolve()` — because
a bind mount shares the superblock and therefore preserves `st_dev`. So the arm is equivalent for the
property under test, not merely adjacent to it.

**All four mutation arms redden exactly their own check and nothing else.** Each run in its own
process, monkeypatching the committed module:

| arm | `dominant_no_lambda` | `namespace_without_children` | hard link | `midfile_inactive_arm` |
| baseline | ok | ok | ok | ok |
| `region_cursor` → top-level scan | **FAIL** `HAND` lam=0.729 | ok | ok | ok |
| `region_cursor` → depth-only descent | ok | **FAIL** `NO-CURSOR` | ok | ok |
| `skipped_lines` → region-kind inference | ok | ok | ok | **FAIL** `EXTRACT` unp=0 |
| `_same_file` → `resolve()` comparison | ok | ok | **FAIL** `_same_file=False` | ok |

**The skip banner.** `survey_sizes.py:561-563` now names all four; correct.

**The `sized` comment.** `survey_sizes.py:348-351` describes what the code does — `MS.REPLICATED` is
`("open", "close")` (`map_symbols.py:123`) and `regions_of` re-roles the preamble separately
(`:202-205`), so preamble regions stay in. Leaving the disagreement to `map_symbols` is defensible:
the filter can only affect a region *count* against a threshold of 4 in a tree whose minimum movable
count is 17, and no file's dominant region is a preamble. The comment says which tool owns the
question, which is the right disposition for a survey that must not move its own published counts.

---

## Also verified

* **Survey output unchanged**: 28 files, `EXTRACT=13, HAND=1, PARSE-FAIL=1, SPLIT=9, UNPARSED=4`, no
  `NO-CURSOR` and no `NO-PREPROC` rows. The `PARSE-FAIL` row prints a repo-relative path.
* **Both ctest cases**: in the container, `ctest -R refactor_survey --no-tests=error` → 2/2 Passed,
  exit 0. On the host, which has no `clang` module: `refactor_survey_classifier` **Passed**,
  `refactor_survey_measurement` ***Skipped**, and `--selftest-parse` alone returns **77**.
* **Hygiene**: 0 `Claude-Session:` trailers by both the line form and the per-commit form;
  `git diff --check` clean; the only host paths added are `<user>` placeholders in prose.

## Informational — checked, no change requested

1. **`_same_file`'s True-on-`OSError` is fail-OPEN at the two new call sites.** Its docstring
   (`survey_sizes.py:176-179`) justifies the fallback for the span clip at `:229`, where True
   over-counts and raises a visible `UNPARSED`. At `:161` and `:166` the polarity is reversed: True
   means "in the TU", so a stat failure on the *first* include silently bypasses the guard and hands
   back a confident 0 — the defect round 3 blocked on. Unreachable in practice (both paths were
   just read successfully), but the comment now covers two sites with opposite consequences and
   defends only one.
2. Following from that, the `except (OSError, TypeError)` at `:163` cannot catch `OSError` —
   `_same_file` swallows it. Only the `TypeError` arm is live.
3. **Three stale "two"s survive** round 3's item 5/6: `survey_sizes.py:463` *"Score the two
   fixtures"*, `README.md:151` *"Neither produced an error"* under a heading announcing three, and
   `README.md:265-267` *"the two `testdata/` measurement fixtures ... those two are the checks"* —
   which now contradicts the code's own banner on the same subject. The half round 3 said mattered
   (the text a libclang-less runner prints) is fixed.
4. **`_need_clang()` is still not thread-safe** (`:94-119` publishes `MS` before `ci` and
   `_SKIPPED_RANGES`). Unchanged from round 2, unreachable from `main()`, fine to leave.
5. **`lam%` still prints `0%` for `SPLIT` rows** where the share was never computed (`:369` gates
   it). Reads as a measured zero; nothing relies on it.

The tool is right and the fixtures are honest discriminators. Fix the body and this is done.

mattias800 (commented):
## Review round 5 — head `684581063`

**APPROVED.**

Rounds 1-4 each found a real defect. This round I re-derived every figure independently rather
than reading the ones on offer, and everything the PR asserts holds. One non-blocking wording
note at the end, which the author can fix in the body without another review round.

### What I ran

Own worktree at `684581063`, fresh `cmake -S prosper -B prosper/build-linux` configure (no
`-DPROSPER_APP`), libclang 22.1 inside the `ps5ys` container.

**Survey over the tree, `--min-lines 2500`** — reproduced exactly:

```
[survey] 28 file(s) at or above 2500 lines
[survey] EXTRACT=13, HAND=1, PARSE-FAIL=1, SPLIT=9, UNPARSED=4
```

Every headline number verified against that run and its `--json`:

| claim | measured |
| --- | --- |
| 28 files; `EXTRACT=13, SPLIT=9, UNPARSED=4, HAND=1, PARSE-FAIL=1` | exact |
| 22 of 28 have a route | 13+9=22; the other 6 are 4 UNPARSED + 1 HAND + 1 PARSE-FAIL |
| `register_live_renderer` 10,244 lines, 98% lambda | 10,244 / 0.983, file 11,450 |
| `emit_alu` 8,186 lines, 11% | 8,186 / 0.108 |
| `test_rdna2_to_spirv.cpp` 3%, `test_recompile_coverage.cpp` 2% | 0.031 / 0.015 |
| `hle_kernel_mem.cpp` UNPARSED at 52% | 3,978 / 7,637 = 0.521 |
| siblings at 30 / 21 / 20% | `hle_kernel.cpp` 0.299, `hle_file.cpp` 0.211, `guest_write_watch.cpp` 0.197 |
| `hle_file.cpp` 963 vs 0 | 963 |
| `CMakeLists.txt:88` globs `src/*.cpp` | line 88 is that `file(GLOB_RECURSE ...)`; the only other glob is line 75 (`third_party/libatrac9`), and `tools/doctor/CMakeLists.txt:40` merely mentions it in a comment |
| 455 ctest tests | `ctest -N` -> `Total Tests: 455`, `#8 refactor_survey_classifier`, `#9 refactor_survey_measurement` |
| 13 arithmetic cases / 4 measurement checks | 8 + 5 printed rows; 4 fixture rows |

`--selftest-parse` on the host (no libclang) returns **77**, and `ctest -R refactor_survey` there
reports `refactor_survey_classifier Passed` / `refactor_survey_measurement ***Skipped`. Inside the
container both pass. The separate-case argument is doing exactly what it claims.

### The four mutation arms

I rebuilt each pre-fix version by hand from `d7be17296` and from the descent/inode code, and ran
`--selftest-parse` on each. Every arm reddens exactly its own check and leaves the other three
green — including the specific values in the table:

| arm | result |
| --- | --- |
| pre-fix top-level cursor scan | `dominant_no_lambda` -> `HAND`, `lambda_share=0.729`, 43 lambda lines / 59-line region |
| pre-fix region-kind inference | `midfile_inactive_arm` -> `EXTRACT`, `unparsed=0`, `biggest='posix_only_1'` |
| descend on depth alone | `namespace_without_children` -> `NO-CURSOR`, "dominant region 19-110 matched no cursor" |
| `resolve()` path comparison | hard-link arm -> `resolve() differs=True, _same_file=False` |

Also re-ran the region-kind arm over the real tree: `guest_write_watch.cpp` comes back
`SPLIT 2513 484 19% 0% 0%` with dominant symbol **`sched.h`** — the README's counter-instance
reproduces verbatim.

And the lambda-share inflation claim, which is a published figure I could not take on trust: running
the pre-fix top-level scan over all 28 files gives **exactly 6 of the 14** files that have a share at
all (13 EXTRACT + 1 HAND) inflated, worst `rdna2_emit_cfg.cpp` at **0.428 against a true 0.258** —
and **zero verdict changes**, confirming "no verdict actually changed" was luck rather than a claim
about the threshold. 6/14 = 43%, 6/22 = 27%. All as written.

### `on_error`, which is what this round was about

The polarity is right and no third caller was missed. `_same_file` has exactly three production call
sites: `survey_sizes.py:161` and `:166` (both `on_error=False`), and `:235` (the span clip, default
`True`). The fourth is the selftest at `:512`.

* `:235` with `True` — a failed stat counts the span, which over-counts toward `UNPARSED`, a row a
  reader sees. Correct direction.
* `:161`/`:166` with `False` — a failed stat is not evidence of membership, and `_file_in_tu`
  returning False produces `SkippedRangesUnavailable` -> a printed `NO-PREPROC` row, not a silent 0.
  So both directions now fail *loud*, which is stronger than the docstring claims for itself.

**It rejects nothing legitimate.** All three headers parsed through an including TU scored normally
in the full run — `gpu_execute.hpp` SPLIT, `rdna2_to_spirv_internal.hpp` EXTRACT,
`render_runner.h` EXTRACT — and no row in the 28 came back `NO-PREPROC`.

The `except TypeError` narrowing at `:163` is safe, and I checked the escape question specifically.
`OSError` could not reach that handler: `_same_file` catches it internally at `:189`, and
`tu.spelling` is a ctypes call returning a `CXString` with no filesystem access, so it cannot raise
`OSError` itself. `TypeError` is genuinely the live case (`os.path.samefile(None, path)`); an empty
spelling gives `FileNotFoundError`, which `_same_file` absorbs and answers `False`, correctly
falling through to the include scan. Nothing that could previously escape `measure()` and abort the
`ex.map` in `main()` can escape now that could not before.

I also confirmed the dead-guard story the PR tells about itself, since it is load-bearing for the
`_file_in_tu` design. Against a `hle_file.cpp` TU, asking about `gpu_executor.cpp`:

```
File.from_name(foreign) -> a valid handle, is None? False
clang_getSkippedRanges(foreign) -> count 0
_file_in_tu(foreign) -> False, and skipped_lines raises SkippedRangesUnavailable
```

Valid handle, empty list, confident 0 — exactly as described, and the inclusion-record test is what
actually closes it.

### Hygiene

`git diff --check origin/main...HEAD` clean. **0** `Claude-Session:` trailers by both the `%B` line
count and the per-commit count. Docs numbered-table gate green over all 190 tracked `.md` files. No
absolute host paths in any changed file — the two `/var/home/...` mentions in `README.md:225` and
`survey_sizes.py:176` use the `<user>` placeholder. Survey stdout is repo-relative on both print
branches, PARSE-FAIL row included. `_TESTDATA in p.parents` works: a `--min-lines 85 --roots tools`
run lists 25 files and none of the three fixtures. Head matches the branch tip
(`git ls-remote` = `684581063…`).

---

### Non-blocking

**1. The body's "3,908" reads as the total when it is one arm of seven.** The sentence is
*"`hle_kernel_mem.cpp` is `UNPARSED` at 52%, not a plain split. 3,908 lines sit behind an inactive
`#if`…"*. Under the "the very arm that issue wants to move" reading it is correct, so this is not a
contradiction — but placed immediately after "52%" it reads as the quantity behind that percentage,
and 3,908/7,637 is **51%**, which is precisely the superseded single-trailing-span figure this PR
replaces. The measured spans are:

```
6-14 (9)  86-88 (3)  732-742 (11)  1468-1473 (6)
2050-2065 (16)  3504-3528 (25)  3730-7637 (3908)   total 3978 / 7637 = 52.1%
```

#3503's own comment already distinguishes them ("the headline arm is 3,908 lines and there are six
smaller skipped spans besides"); the body is the only place that gives 3,908 without the 3,978 next
to it. One clause fixes it — e.g. *"3,978 lines across seven arms, 3,908 of them in one"* — and it
needs no re-review.

**2. `_need_clang()` still publishes `MS` before `ci`** (`survey_sizes.py:94-99`). The early-return
guard keys on `MS`, which is assigned first, so a second thread entering between `MS = …` and
`_DISPOSE_RANGES = …` would return with `ci = None`. Unreachable today because `main()` calls
`_need_clang()` at `:590` before the pool starts at `:614`, and nothing this round added a new path.
Keying the guard on `_DISPOSE_RANGES` instead would close it for free.

**3. `lam%` still prints `0%` for the 9 SPLIT rows**, and the JSON carries
`"lambda_share_of_biggest": 0.0` for them, where the share was never computed (`measure()` skips it
below `DOMINANT_SHARE` at `:375`). Harmless for the verdict, but it is "not measured" published as a
value in a tool whose whole argument is that those are the dangerous ones. `None` / `--` would say
the true thing.

**4. Minor:** the `--json` rows keep absolute paths (`json.dumps(rows)` at `:643` uses `r["file"]`
raw, while the print loop strips at `:627`). Nothing committed leaks, and the `:624` comment's
"every branch" plainly means the two print branches — but the reason given there ("this repository
is public") applies just as much to a JSON someone attaches to an issue.

**5. Cosmetic:** `README.md:265` now says "the four `testdata/` measurement checks", and only three
are `testdata/` fixtures — the fourth is the hard-link arm built in a `tempfile`. The selftest's own
skip message names all four correctly.

Issue #3516, opened 2026-09-10
refactor: split gpu_executor.cpp (12,058 lines, the largest source file) along its measured reference seam


mattias800:
## `gpu_executor.cpp` is 12,058 lines and is the largest source file in the tree

It is also the cleanest `SPLIT` candidate: no region dominates it (biggest is 23%), it has 273 movable top-level regions, and only 0.8% of it sits behind an inactive `#if`. `survey_sizes.py` (#3514) classifies it `SPLIT`, meaning `split_file.py` applies today with no prerequisite work.

Filing this so the seam measurement below is not re-derived. Nothing here is a bug; this is the durable queue entry for the split.

## The seam, measured rather than read

`map_symbols.py --json` emits a reference matrix (`edges`: region -> {region: count}). Symmetrised over body regions and clustered greedily, `gpu_executor.cpp` separates into architectural groups with **13% of references crossing** (447 of 3,525):

| lines | what |
| --- | --- |
| 2,738 | `resolve_dynamic_fetch` and its helper run |
| 2,389 | ordered execution — `execute_ordered_gpustate`, `execute_ordered_and_present` |
| 2,389 | compute realization — `realize_compute_dispatches`, `build_stage_table`, `diagnose_compute_dispatches` |
| 1,407 | shader compile/cache — `make_shader_compile_key`, `recompile_compute_shader_cached`, `ShaderCompileKeyHash` |
| 990 | `add_compute_buffer_resources` |
| ~1,100 | coherent leaves: `guest_writable` + its cache state, compute memprobe, tree-watch, stage-fold profile |

Those leaf groups each came out as a state struct plus its hash/key helper plus the functions using it, with no name matching involved. Grouping by name prefix does **not** find this — it puts 183 of 273 regions in "other".

## Where the cut boundary actually is, and where it is an artifact

Two of those groups landing on exactly 2,389 lines is the size cap talking, not a seam. The boundary that **is** structural, measured by cut cost:

| candidate unit | lines | outgoing refs | incoming refs |
| --- | --- | --- | --- |
| `resolve_dynamic_fetch` alone | 2,738 | 97 (20 distinct) | 12 (3 distinct) |
| **+ its helper run, regions 192-206** | **3,007** | **33 (7 distinct)** | **39 (4 distinct)** |
| + `guest_readable` etc, regions 188-206 | 3,201 | 54 (15 distinct) | 159 (17 distinct) |

[...]

PR #3519, opened 2026-09-10
tools/refactor: promote_internal wrote headers that cannot compile, three ways

### LINKED ISSUE #3515: map_symbols.py documents a --clusters flag that does not exist (and it propagated into #3503)
## `map_symbols.py` advertises a `--clusters` flag that does not exist

`prosper/tools/refactor/map_symbols.py:27`, inside the module docstring that `--help` prints:

```
  ... --clusters      # suggest seams: greedily group regions that reference each other
```

There is no such flag. `argparse` registers exactly four (`map_symbols.py:298-301`: `--file`, `--build`, `--json`, `--min-lines`), and the string `clusters` appears **nowhere else in the repository** — that docstring line is its only occurrence.

```
$ python3 prosper/tools/refactor/map_symbols.py --file <a file> --clusters
map_symbols.py: error: unrecognized arguments: --clusters
```

## Why this is worth an issue rather than a silent one-line deletion

**It has already propagated.** I cited `map_symbols.py --clusters` in a comment on #3503 as a tool that had been consulted, without running it — the docstring reads as documentation of shipped behaviour, which is exactly the authority-transfer this project's charter warns about. That comment is now corrected. A reader reaching for the documented flag finds an error message; a reader *quoting* it, as I did, passes on a capability the project does not have.

## The feature is the useful part

The advertised behaviour — "greedily group regions that reference each other" — is precisely what choosing a split seam needs, and `map_symbols.py` already computes the input for it. `cross_refs()` returns `region -> {other region: reference count}`, and the docstring's REFERENCES section explains why that matrix matters: a split that separates an internal-linkage helper from its callers *"does not fail at review, it fails at link"*. Clustering on that matrix is the step between having the edges and choosing where to cut.

So there are two honest resolutions and they are not equivalent:

1. **Delete the line.** Correct immediately, and leaves seam selection a manual read of the reference matrix.
2. **Implement it.** The matrix is already there; this is a grouping pass over it plus a printer.

(2) is the better outcome if the split work in #3503 and the 9 `SPLIT`-classified files from #3514's survey are going to proceed, because every one of them needs the same judgement made. (1) is correct in the meantime and should not wait for (2).

## Suggested fix

Delete `map_symbols.py:27` now. If clustering is implemented later, the flag comes back with an implementation behind it and a selftest — `split_file.py --selftest` and `refactor_survey_classifier` are the precedent for how tools in this directory get pinned.

Found while surveying split candidates for #3503 (tooling: `survey_sizes.py`, #3514).

### LINKED ISSUE #3517: tools/refactor: four small items from #3514's review (thread-safety guard, unmeasured 0%, JSON paths, README wording)
## Four small items left from #3514's review rounds

None blocks anything and none produces a wrong answer today. Filed rather than fixed in #3514 so they do not detach an approval earned on `684581063`, and filed rather than dropped because two of them are the kind of thing that becomes wrong later rather than being wrong now.

### 1. `_need_clang()` publishes `MS` before `ci`

`prosper/tools/refactor/survey_sizes.py:94-99`. The function's own re-entry guard is `if MS is not None: return`, and `MS` is assigned before `ci`. A second thread arriving between the two would take the early return and then use `ci` while it is still `None`.

Unreachable today: `main()` calls `_need_clang()` at `:590`, before the `ThreadPoolExecutor` at `:614`, so the pool never races it. It becomes reachable the moment anything calls `survey_one`/`measure` from a worker without that pre-initialisation — which is exactly how a caller would reasonably use the module.

Fix: assign the guard last, or take a lock.

### 2. `lam%` prints `0` where the share was never computed

The lambda share is only computed when a region dominates (`big_len / total >= DOMINANT_SHARE`), so the 9 `SPLIT` rows carry `lambda_share_of_biggest: 0.0` — a value that was never measured, printed identically to a measured zero. `test_rdna2_to_spirv.cpp`'s genuine 3% and a `SPLIT` file's unmeasured 0% look like the same kind of fact in the table.

This is the shape of defect the PR spent four review rounds on: a number that reads as measured when it is not. Print blank for those rows, and omit the key from the JSON rather than emitting `0.0`.

### 3. `--json` keeps absolute paths where the printed table strips them

The print loop strips to a repo-relative path on every branch (`:482`, fixed during review for the charter's privacy rule). `--json` does not — its `file` field stays absolute.

Defensible, since the JSON is machine-readable output for downstream tooling and a resolvable path is the useful form there. Worth deciding deliberately rather than by omission: this repository is public and a survey JSON is exactly the kind of artifact someone pastes into an issue. Either strip it, or emit both a `file` and a `path` field and say which is which.

### 4. `README.md:265` calls all four measurement checks `testdata/` fixtures

Three are files under `tools/refactor/testdata/`. The fourth — the inode/alias arm — builds a hard link in a `tempfile` directory at run time, because the property it needs (two paths naming one file that `resolve()` reports as different) cannot be committed as a fixture. Reword.

Found during the review of #3514; tooling context in that PR.

### LINKED ISSUE #3518: promote_internal.py emits an uncompilable header: forward decls filtered to 'public', and the preamble's #if copied without its #endif
## `promote_internal.py` emits a header that cannot compile, in two independent ways

Found by using it: promoting 29 regions out of `src/gpu/capture/gpu_capture.cpp` (6,515 lines) produced a header the compiler rejects immediately. The tool reported **five `[ok]` checks and no warnings**, which is the part worth fixing — it says at the end *"now BUILD and run ctest; this tool does not verify its own work"*, and that is honest, but both of these are statically decidable from what the tool already has in hand.

Repro:

```bash
python3 prosper/tools/refactor/map_symbols.py --file prosper/src/gpu/capture/gpu_capture.cpp --json /tmp/cap.json
python3 prosper/tools/refactor/promote_internal.py --map /tmp/cap.json \
    --regions 42,44,45,46,48,49,50,51,52,53,54,55,57,58,89,92,94,97,114,115,116,117,118,119,122,125,127,128,141 \
    --header gpu_capture_internal.hpp --namespace prosper::gpu
```

### 1. The forward-declaration pass uses a narrower rule than its own comment

`prosper/tools/refactor/promote_internal.py:640-641`:

```python
# Anything promoted code CALLS but that stayed behind needs a declaration in the header.
wanted = {d for i in promote for d in edges_all.get(i, set()) if d in set(public)}
```

The comment states the correct rule. The code implements a different one: `public` is the set of regions *referenced from outside this file* (`:558`), so a promoted function that calls an ordinary internal helper gets **no declaration**, and the helper's definition is below the `#include` in the source.

Result — every one of these is an internal-linkage helper that stayed behind:

```
gpu_capture_internal.hpp:170:28: error: 'blob_max_bytes_from_env' was not declared in this scope
gpu_capture_internal.hpp:196:28: error: 'checked_mul' was not declared in this scope
gpu_capture_internal.hpp:207:12: error: 'resource_footprint_impl' was not declared in this scope
gpu_capture_internal.hpp:341:38: error: 'captured_dma_destination_is_gds' was not declared in this scope
gpu_capture_internal.hpp:342:33: error: 'captured_dma_source_is_gds' was not declared in this scope
```

The `[warn]` branch immediately below (`:648`) exists for exactly this situation — "stays behind and is called from promoted code, but no declaration could be derived" — and cannot fire for these, because they never enter `wanted`.

**Fix:** drop the `if d in set(public)` filter. The set wanted is *every* destination that is not itself promoted. The prototype machinery (`prototype_of`) is already there.

**Better fix, and the reason this is worth more than a one-line change:** the tool could report the transitive closure and offer it. The caller has to guess a promote set, and the correct set is closed under "what does this call that has internal linkage". Printing *"these 7 regions are called by your set and stayed behind; add them, or accept these forward declarations"* turns a compile-error hunt into a decision.

### 2. The header's preamble is copied verbatim, `#if` and all

`:217-219` builds the header's head by copying every `preamble` region:

```python
preamble = [r for r in map_data["regions"] if r.get("role") == "preamble"]
head += ["".join(lines[r["start"] - 1:r["end"]]) for r in preamble]
```

`gpu_capture.cpp`'s preamble contains

```c
#if defined(__linux__) || defined(__APPLE__)
#include <sys/uio.h>
#include <unistd.h>
#include "host/platform/posix_shim.hpp"
```

whose `#endif` is in a later, non-preamble region. So the header ships **3 `#if` against 2 `#endif`**:

```
gpu_capture_internal.hpp:49: error: unterminated #if
```

The tool already models preprocessor depth for the regions it promotes (`:600-639`, which correctly drops or co-promotes regions inside a conditional). That machinery is simply not applied to the preamble.

**Fix:** either balance the copied preamble (emit the matching `#endif`s the copy left open) and say so, or refuse with the unbalanced directive named. Refusing is fine — the caller can hoist the include block — but silently emitting an unterminated `#if` is not.

### Why the existing checks cannot see either

The reconstruction check (`:305`, `:705`) verifies *"the source is the original minus exactly those spans plus one include"* — a **line-accounting** identity over the source. Both defects are properties of the **header**, which that check does not examine, and both are about the C preprocessor rather than about lines. This is the same shape as the `UNPARSED` finding in #3514: a check that is sound about the quantity it measures, and silent about the one that matters.

### Suggested regression arms

Both are constructible as small fixtures, and both must produce a *different outcome*, not merely a different message:

- a source whose promoted function calls an internal helper that stays behind → the tool emits a declaration for it (or refuses and names it); reverting `:641` to the `public` filter reddens it.
- a source whose preamble opens a `#if` closed later in the body → the tool refuses or balances; reverting reddens it.

Neither needs libclang beyond what `map_symbols.py` already requires.

Found while splitting `gpu_capture.cpp`; that work is blocked on this.

mattias800:
Fixes #3518. Fixes #3515. Fixes #3517.

**Scope:** one PR for all outstanding `prosper/tools/` work, per the project owner's direction — granular PRs earn their cost through `git bisect`, and tooling is not built, not linked and not run by CI, so no regression can bisect to it. Nothing here touches `src/`, `frontends/` or `tests/`.

## Why

`promote_internal.py` is the tool that makes a big translation unit splittable — it lifts internal-linkage declarations into a header so a second file can see them. Using it on `src/gpu/capture/gpu_capture.cpp` (6,515 lines) produced a header the compiler rejected immediately, under **five `[ok]` checks and no warnings**.

The tool prints *"now BUILD and run ctest; this tool does not verify its own work"*, which is honest. But all three defects below are decidable from data the tool already has, and the failure mode is the expensive kind: it reports success, and the cost lands at build time on someone reading a 700-line generated header.

## Three defects, found in sequence — each one hid the next

**1. The forward-declaration pass used a narrower rule than its own comment.**

```python
# Anything promoted code CALLS but that stayed behind needs a declaration in the header.
wanted = {d for i in promote for d in edges_all.get(i, set()) if d in set(public)}
```

`public` is the set referenced from *outside* the file. A promoted function calling an ordinary internal helper therefore got no declaration, and used it above its definition:

```
error: 'blob_max_bytes_from_env' was not declared in this scope
error: 'checked_mul' was not declared in this scope
error: 'resource_footprint_impl' was not declared in this scope
```

The `[warn] no declaration could be derived` branch that exists for exactly this could not fire either — those regions never entered the set it iterates.

**2. The copied preamble carried an unbalanced `#if`.**

The header's head is the source's preamble regions copied verbatim. `gpu_capture.cpp` opens `#if defined(__linux__) || defined(__APPLE__)` among its includes and closes it far below, outside any preamble region — so the header shipped **3 `#if` against 2 `#endif`** and died on `unterminated #if`. The tool already models preprocessor depth for the regions it *promotes*; it never applied that to the part it *replicates*.

**3. With both fixed, the build still failed — and the fix for (1) turned out to be impossible in principle.**

```
error: call of overloaded 'resource_footprint_impl(...)' is ambiguous
```

A definition still inside `namespace { }` **cannot be forward-declared from a header**. The declaration names `prosper::gpu::f` while the definition is `prosper::gpu::{anonymous}::f` — two different entities, both visible, so every call becomes ambiguous rather than resolving. No header fixes that.

So the tool now **refuses and names the closure**: the promote set must be closed under "what does this call that is still internal", and nobody derives that by hand from a 6,000-line file. It prints the exact `--regions` to re-run with.

That closure is iterated to a **fixed point**, not reported one step at a time — measured here, the first suggestion needed a second round and the second a third, so a single step just makes the caller run it again. Iterating turned the 29 regions I asked for into the **55 that are actually closed**, in one answer.

## Verification

End-to-end on the file that found it. Promoting the 55-region closure:

- writes a 1,373-line `gpu_capture_internal.hpp` and shrinks `gpu_capture.cpp` from **6,515 → 5,208** lines
- **builds clean**
- `ctest --no-tests=error -j4` → **455/455 passed**, exit 0

**That source change is not in this PR.** It is a step of the `gpu_capture.cpp` split and belongs with it; this is the tool fix alone.

Selftest arms, both mutation-checked — each reverts the fix and reddens **only** its own arm:

| mutation | reddens |
| --- | --- |
| restore the `public` filter | `a promoted region's callees that stay behind are ALL declared, public or not` |
| remove the preamble balancing | `the emitted header is preprocessor-balanced even when the preamble is not` |

The forward-declaration rule moved out of `main()` into `declarations_needed` **specifically so the selftest can reach it**. The selftest's own comment already noted that fixes living in `main()` are unreachable from it — and this rule was wrong for as long as it was unreachable. That is the part worth keeping in mind beyond this patch.

`python3 prosper/tools/refactor/promote_internal.py --selftest` → all arms ok. `diff --check` clean; 0 `Claude-Session:` trailers.

## Risk

Tooling only — no emulator code, nothing linked into any target. `promote_internal.py` is not invoked by the build or by CI; it is run by hand during a restructure. The change makes it refuse inputs it previously accepted and mis-handled, so the failure direction moves from "silently emits a broken header" to "declines and says what to do".

## Affected, deliberately

`gpu_capture.cpp` is untouched here. #3516 (`gpu_executor.cpp`) remains blocked on #3399, which edits that file.

---

## Round 2 — the fix was worse than the bug, and only Windows could see it

Review found that `build()` learned the copied preamble ends at conditional depth 1, balanced the **header**, and then used that same preamble end as the insertion point for the include it adds to the **source**. On `gpu_capture.cpp` the last preamble line is 46 and the matching `#endif` is 47, so the internal header was included behind `#if defined(__linux__) || defined(__APPLE__)` while the promoted definitions were deleted unconditionally.

Measured with the tool's own suggested closure — three arms and a control:

| arm | Linux `g++` | `x86_64-w64-mingw32-g++` |
| --- | --- | --- |
| **fixed** | 0 errors | **0 errors** |
| mutant (include at the last preamble line) | 0 errors | **208 errors** |
| control: unmodified file, same flags | 0 errors | 0 errors |

So the previous round's *"builds clean, 455/455 ctest"* was a **Linux-only claim that could not have seen this**, and it was stated without that qualification. The insertion point now walks out to conditional depth 0.

The selftest arm for the preamble fix asserted **balance**, which does not imply **placement** — moving the `#endif` to the end of the header leaves it green while the header wraps its whole namespace in a platform conditional. A new fixture reproduces `gpu_capture.cpp`'s exact shape and asserts the source's include lands at depth 0.

Also from that review: the refusal predicate used `anon.get(i)`, the test this file documents at `:139-151` as the wrong one (a `static` free function is internal linkage outside any anonymous namespace) — it uses `externally_linked()` now; the suggested `--regions` closure ignored `--exclude`, so the printed re-run reproduced the identical refusal; `preprocessor_depth_after` matches with the file's own `COND_OPEN`/`COND_CLOSE` rather than `str.startswith`, which disagreed on `#  if`; and a comment claiming the `#endif` sits "far below" now says "one line below", which is true.

## Round 3 — `--clusters`, and #3517

`map_symbols.py` has advertised a `--clusters` flag in its usage text with nothing behind it. Implemented, because choosing a split seam is exactly what it describes and `cross_refs` already computes the input. It reports the **cut cost** — the references crossing a boundary, i.e. the promote-to-header list. On `gpu_capture.cpp` at `--clusters 3`: 252 of 2,927, **9%**.

It states two things about itself: a group sitting *at* the size cap is the cap talking rather than a seam, and the part count is what the structure and cap allow rather than what was asked for.

**An early draft let attachment ignore the cap**, which funnelled every group into the largest connected component — one **6,374-line group out of 6,515**, reported as a **0% cut**. A split that is not a split, wearing a perfect score.

That arm needed two attempts and **the first was void**: the obvious fixture (two clusters joined by a weak edge) passes under both implementations, because the first pass already leaves few enough groups that the attachment loop never runs. The discriminating fixture makes the first pass leave *more* groups than requested — six 100-line regions at `parts=4`, so no two can merge under a 150-line cap — which is the only shape where the cap is what stands between six groups and one.

`map_symbols.py` had **no selftest and no ctest case**, so this would have shipped untested. It has both now (`refactor_map_symbols_clusters`), and its libclang import is lazy so the case runs where there is no libclang.

Folded in from #3517: the re-entry guard in `survey_sizes._need_clang` is published last rather than first; an unmeasured lambda share is omitted rather than printed as a measured `0%` beside real ones; `--json` emits repo-relative paths like the printed table already did; README wording corrected.

`ctest -R refactor_ --no-tests=error` → 3/3 passed.

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -178,6 +178,13 @@ if(Python3_Interpreter_FOUND)
   # under tools/refactor/testdata/ that the selftest scores ONLY where libclang is importable, and
   # it says loudly when it skips them. libclang resolves lazily, so the arithmetic half runs here
   # even though the surveying itself needs a libclang the runner does not have.
+  # map_symbols' clustering decides where a split cuts, and its failure mode is a confident wrong
+  # proposal rather than a crash -- one draft collapsed a 6,515-line file into a single 6,374-line
+  # "part" and reported a 0% cut. Pure arithmetic over the reference matrix, and libclang resolves
+  # lazily there now, so this runs on a runner that has none.
+  add_test(NAME refactor_map_symbols_clusters
+           COMMAND "${Python3_EXECUTABLE}"
+                   "${CMAKE_SOURCE_DIR}/tools/refactor/map_symbols.py" --selftest)
   add_test(NAME refactor_survey_classifier
            COMMAND "${Python3_EXECUTABLE}"
                    "${CMAKE_SOURCE_DIR}/tools/refactor/survey_sizes.py" --selftest)
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -182,6 +182,12 @@ if(Python3_Interpreter_FOUND)
   # proposal rather than a crash -- one draft collapsed a 6,515-line file into a single 6,374-line
   # "part" and reported a 0% cut. Pure arithmetic over the reference matrix, and libclang resolves
   # lazily there now, so this runs on a runner that has none.
+  # split_file.py was in NO ctest case at all, so a "3/3 refactor tests pass" said nothing about
+  # the splitter -- including a cross-reference check whose arms could not fail. Both gaps found in
+  # the same review.
+  add_test(NAME refactor_split_file
+           COMMAND "${Python3_EXECUTABLE}"
+                   "${CMAKE_SOURCE_DIR}/tools/refactor/split_file.py" --selftest)
   add_test(NAME refactor_map_symbols_clusters
            COMMAND "${Python3_EXECUTABLE}"
                    "${CMAKE_SOURCE_DIR}/tools/refactor/map_symbols.py" --selftest)
--- a/prosper/tools/refactor/README.md
+++ b/prosper/tools/refactor/README.md
@@ -287,20 +287,25 @@ the defect changes the group COUNT rather than merely a number.
 ### The three tools are a chain, and each one checks the step before it
 
 `map_symbols` proposes a seam, `promote_internal` makes the shared vocabulary visible, `split_file`
-cuts. The order is forced, and the reason is worth stating because skipping a step produces a result
-that passes every check and does not compile:
+cuts. The middle step is not always needed -- 42 of `gpu_capture.cpp`'s 155 body regions are already
+external, and a seam that only crosses those needs no promotion at all. What the order buys is that
+**whether you need it is now decidable before the cut** rather than at build time:
 
 ```bash
-map_symbols.py  --file <f> --clusters 3 --json /tmp/f.json   # where to cut, and what it costs
-promote_internal.py --map /tmp/f.json --regions <closure> --header <f>_internal.hpp
-map_symbols.py  --file <f> --json /tmp/f2.json               # promotion changes every index
-split_file.py   --map /tmp/f2.json --plan <plan> --dry-run
+map_symbols.py  --file <f> --clusters 3 --json ~/work/f.json   # where to cut, and what it costs
+split_file.py   --map ~/work/f.json --plan <plan> --dry-run    # refuses, naming what to promote
+promote_internal.py --map ~/work/f.json --regions <closure> --header <f>_internal.hpp
+map_symbols.py  --file <f> --json ~/work/f2.json               # promotion changes every index
+split_file.py   --map ~/work/f2.json --plan <plan> --dry-run
 ```
 
+(Maps go under `$HOME`, not `/tmp`: that is a RAM-backed tmpfs here with a quota shared by every
+concurrent agent.)
+
 **Both closure checks exist because their absence was measured, not imagined.** Splitting
-`gpu_capture.cpp`'s serialize/deserialize into their own file satisfied every check `split_file` had
--- exact tiling, a clean partition, byte-for-byte reconstruction -- and then failed to compile with
-**25 undeclared names**, because `Writer`, `Reader`, `kMagic`, `read_table` and twenty others have
+`gpu_capture.cpp`'s serialize/deserialize into their own file satisfied every check `split_file`
+*then had* -- exact tiling, a clean partition, byte-for-byte reconstruction -- and failed to compile
+with **25 undeclared names**, because `Writer`, `Reader`, `kMagic`, `read_table` and twenty others have
 internal linkage and stayed in the other part. Every one of those checks is byte accounting: they
 establish that no LINE was lost, never that a line can still be used where it ended up.
 
--- a/prosper/tools/refactor/README.md
+++ b/prosper/tools/refactor/README.md
@@ -255,6 +255,35 @@ would be ranked as needing no work. One file is currently in that state:
 `frontends/prosper-app/main.cpp`, which has no compile command unless the app is configured
 (`-DPROSPER_APP=ON`); configure with it to bring that file into the survey.
 
+### Choosing where to cut
+
+`map_symbols.py --clusters [N]` groups the file's regions by what they actually reference, and
+reports the cost of the seam it proposes:
+
+```
+== proposed seams: 9 group(s), target ~2149 lines each == -- you asked for 3, and the reference
+   structure plus that size cap do not permit fewer
+   252 of 2927 references cross a boundary (9%) -- that is the promote-to-header list
+   83 merge(s) the references wanted were refused by the size cap; a group AT the cap is the cap
+   talking, not a seam
+```
+
+The cross-boundary count is the number that matters, because those are the internal-linkage edges
+the REFERENCES section warns about -- a split that separates a `static` helper from its callers
+*"does not fail at review, it fails at link"*. Having the list up front turns `promote_internal.py`
+into a decision instead of a compile-error hunt.
+
+**Two things it says about itself, and both are load-bearing.** A group sitting exactly AT the size
+cap is the cap talking, not a structural seam -- it stopped there because it was told to. And the
+part count is what the references and the cap allow, not what you asked for; the tool says so rather
+than forcing the number. Grouping by NAME finds none of this: on `gpu_executor.cpp` a name-prefix
+pass put 183 of 273 regions in "other".
+
+**A 0% cut is a tell, not a success.** An early draft let attachment ignore the cap, which funnelled
+every group into the largest connected component -- one 6,374-line "part" out of 6,515 lines,
+reported as a perfect zero-cut split. That is pinned by a selftest arm now, with the fixture built so
+the defect changes the group COUNT rather than merely a number.
+
 ### Running it
 
 ```bash
--- a/prosper/tools/refactor/README.md
+++ b/prosper/tools/refactor/README.md
@@ -313,7 +313,13 @@ The reference matrix that answers it was already in the map. `split_file` reads 
 naming the definitions and printing the `promote_internal` command that fixes them. Measured against
 the compiler on that split: **23 names flagged, 25 reported by g++, and the two it did not name are
 `r` and `w`** -- local variables in cascading errors from `Reader r;` and `Writer w;`. No false
-positives.
+positives on that file.
+
+**The check cannot see inactive `#if` arms**, and that limit is not small: libclang parses one arm,
+so a map made on Linux carries no references from the 52% of `hle_kernel_mem.cpp` or the 30% of
+`hle_kernel.cpp` that sit behind a platform conditional. A split validated here can still strand a
+definition the other platform's arm uses. `survey_sizes.py` reports that share per file as
+`UNPARSED`; treat a clean result on such a file as a statement about the active arm only.
 
 `promote_internal` has the mirror of the same check, for the same reason.
 
--- a/prosper/tools/refactor/README.md
+++ b/prosper/tools/refactor/README.md
@@ -262,10 +262,13 @@ python3 prosper/tools/refactor/survey_sizes.py --min-lines 2500     # needs libc
 python3 prosper/tools/refactor/survey_sizes.py --selftest           # no libclang needed
 ```
 
-`--selftest` runs the classifier's arithmetic cases always, and the four `testdata/` measurement
-checks **only where libclang is importable** -- saying so explicitly, and naming them, when it skips
-them. Those are the checks covering how the classifier's INPUTS are measured, and every defect this
-tool shipped lived there rather than in the arithmetic. The registered ctest case
+`--selftest` runs the classifier's arithmetic cases always, and the four measurement checks **only
+where libclang is importable** -- three of them score committed fixtures under `testdata/`, while the
+fourth builds a hard link in a temporary directory, because the property it needs (two paths naming
+one file that `resolve()` reports as different) cannot be committed as a fixture. `--selftest` says
+so explicitly, and names them, when it skips them. Those are the checks covering how the
+classifier's INPUTS are measured, and every defect this tool shipped lived there rather than in the
+arithmetic. The registered ctest case
 (`refactor_survey_classifier`) therefore gates the arithmetic everywhere and the measurement locally.
 
 **The invariant that makes incremental extraction safe is worth copying.** When a block was extracted
--- a/prosper/tools/refactor/README.md
+++ b/prosper/tools/refactor/README.md
@@ -284,6 +284,34 @@ every group into the largest connected component -- one 6,374-line "part" out of
 reported as a perfect zero-cut split. That is pinned by a selftest arm now, with the fixture built so
 the defect changes the group COUNT rather than merely a number.
 
+### The three tools are a chain, and each one checks the step before it
+
+`map_symbols` proposes a seam, `promote_internal` makes the shared vocabulary visible, `split_file`
+cuts. The order is forced, and the reason is worth stating because skipping a step produces a result
+that passes every check and does not compile:
+
+```bash
+map_symbols.py  --file <f> --clusters 3 --json /tmp/f.json   # where to cut, and what it costs
+promote_internal.py --map /tmp/f.json --regions <closure> --header <f>_internal.hpp
+map_symbols.py  --file <f> --json /tmp/f2.json               # promotion changes every index
+split_file.py   --map /tmp/f2.json --plan <plan> --dry-run
+```
+
+**Both closure checks exist because their absence was measured, not imagined.** Splitting
+`gpu_capture.cpp`'s serialize/deserialize into their own file satisfied every check `split_file` had
+-- exact tiling, a clean partition, byte-for-byte reconstruction -- and then failed to compile with
+**25 undeclared names**, because `Writer`, `Reader`, `kMagic`, `read_table` and twenty others have
+internal linkage and stayed in the other part. Every one of those checks is byte accounting: they
+establish that no LINE was lost, never that a line can still be used where it ended up.
+
+The reference matrix that answers it was already in the map. `split_file` reads it now and refuses,
+naming the definitions and printing the `promote_internal` command that fixes them. Measured against
+the compiler on that split: **23 names flagged, 25 reported by g++, and the two it did not name are
+`r` and `w`** -- local variables in cascading errors from `Reader r;` and `Writer w;`. No false
+positives.
+
+`promote_internal` has the mirror of the same check, for the same reason.
+
 ### Running it
 
 ```bash
--- a/prosper/tools/refactor/map_symbols.py
+++ b/prosper/tools/refactor/map_symbols.py
@@ -274,6 +274,15 @@ def cross_refs(tu, target: pathlib.Path, regions: list[dict]) -> dict[int, dict[
             ref = ch.referenced
             if ref is None:
                 continue
+            # A NAMESPACE cursor is not a reference to anything a split has to keep together. Every
+            # re-opening `namespace {` and `namespace prosper {` resolves `.referenced` to the
+            # namespace declaration, so the brace itself became an edge -- attributed to whichever
+            # body region happened to hold the first cursor carrying that USR. Measured on
+            # hle_kernel.cpp: 19 of 19 edges whose source is a replicated region were this, 13 of
+            # them pointing at `fbsd_errno`, and acting on them refused every legal two-part plan
+            # for the file. They are noise in every consumer, not only the splitter.
+            if ch.kind == ci.CursorKind.NAMESPACE or ref.kind == ci.CursorKind.NAMESPACE:
+                continue
             usr = ref.get_usr()
             if usr not in owner:
                 continue
--- a/prosper/tools/refactor/map_symbols.py
+++ b/prosper/tools/refactor/map_symbols.py
@@ -24,7 +24,8 @@ This reports two things about a file, both from clang's own AST:
 
 Usage:
   python3 prosper/tools/refactor/map_symbols.py --file prosper/src/gpu/recompiler/rdna2_to_spirv.cpp
-  ... --clusters      # suggest seams: greedily group regions that reference each other
+  ... --clusters [N]  # suggest seams: greedily group regions that reference each other
+                      #   into N parts (default 6), reporting the cut cost
   ... --json out.json # machine-readable, for split_file.py
 
 Flags come from the build's compile_commands.json, so the parse sees exactly what the compiler sees.
@@ -40,14 +41,27 @@ import subprocess
 import sys
 from collections import defaultdict
 
-import clang.cindex as ci
+# libclang is imported LAZILY. Everything that parses needs it, and `propose_clusters` -- which
+# decides where a split should cut -- does not: it is arithmetic over the reference matrix. Keeping
+# the import out of module scope is what lets `--selftest` pin that arithmetic on a machine with no
+# libclang, which is the shape of this project's CI. `survey_sizes.py` resolves it the same way and
+# for the same reason.
+ci = None
 
-# The bindings look for a bare `libclang.so`, which the runtime package does not ship; the versioned
-# soname is what is actually installed. Fail loudly rather than letting a later call die obscurely.
-for cand in ("/usr/lib64/libclang.so.22.1", "/usr/lib64/libclang.so"):
-    if pathlib.Path(cand).exists():
-        ci.Config.set_library_file(cand)
-        break
+
+def _need_clang() -> None:
+    global ci
+    if ci is not None:
+        return
+    import clang.cindex as _ci
+    # The bindings look for a bare `libclang.so`, which the runtime package does not ship; the
+    # versioned soname is what is actually installed. Fail loudly rather than letting a later call
+    # die obscurely.
+    for cand in ("/usr/lib64/libclang.so.22.1", "/usr/lib64/libclang.so"):
+        if pathlib.Path(cand).exists():
+            _ci.Config.set_library_file(cand)
+            break
+    ci = _ci                              # last: this is the guard the early return above tests
 
 
 def repo_root() -> pathlib.Path:
@@ -292,16 +306,207 @@ def cross_refs(tu, target: pathlib.Path, regions: list[dict]) -> dict[int, dict[
     return {k: dict(v) for k, v in edges.items()}
 
 
+def propose_clusters(regions, edges, parts: int) -> tuple[list[list[int]], dict]:
+    """Group body regions into PARTS candidate output files, cutting where the references are thin.
+
+    Greedy agglomeration over the reference matrix, heaviest edge first, which is what the file's
+    REFERENCES section is for: an internal-linkage helper separated from its callers does not fail
+    at review, it fails at link. Grouping on that matrix puts them on the same side by construction.
+
+    The size cap is reported rather than hidden, because it is the part that lies. Two groups landing
+    on the same line count means the CAP stopped the merge, not that a seam is there -- so the
+    summary marks which boundaries are structural and which are the cap talking. Grouping by NAME
+    does not find these: on gpu_executor.cpp a name-prefix pass put 183 of 273 regions in "other".
+    """
+    body = {r["index"]: r for r in regions if r["role"] == "body"}
+    size = {i: r["end"] - r["start"] + 1 for i, r in body.items()}
+    total = sum(size.values())
+    cap = max(1, total // max(1, parts))
+
+    w: dict[tuple[int, int], int] = {}
+    for a, ds in edges.items():
+        for b, n in ds.items():
+            if a in body and b in body and a != b:
+                key = (min(a, b), max(a, b))
+                w[key] = w.get(key, 0) + n
+
+    parent = {i: i for i in body}
+    def find(x):
+        while parent[x] != x:
+            parent[x] = parent[parent[x]]
+            x = parent[x]
+        return x
+    csize = dict(size)
+    capped = 0
+    for (a, b), _ in sorted(w.items(), key=lambda kv: -kv[1]):
+        ra, rb = find(a), find(b)
+        if ra == rb:
+            continue
+        if csize[ra] + csize[rb] > cap:
+            capped += 1               # a merge the references wanted and the cap refused
+            continue
+        parent[rb] = ra
+        csize[ra] += csize[rb]
+
+    groups: dict[int, list[int]] = {}
+    for i in body:
+        groups.setdefault(find(i), []).append(i)
+
+    # SECOND PASS: attach the tail. A cap-driven greedy leaves a long run of singletons -- 30 of 41
+    # on gpu_capture.cpp -- and a proposal the caller then has to place by hand is not a proposal.
+    # Merge the smallest group into whichever group it references most, ignoring the cap, until the
+    # requested part count is reached. A region with no cross-references at all has nothing to
+    # attach to and stays where it is; that is information, not a failure.
+    def edge_between(g1, g2):
+        return sum(w.get((min(a, b), max(a, b)), 0) for a in g1 for b in g2)
+
+    # The cap binds here too, and that is the whole difference between a proposal and a lie. Without
+    # it, "attach the smallest to whatever it references most" funnels every group into the largest
+    # connected component: on this file it produced ONE 6,374-line group out of 6,515 and reported a
+    # 0% cut -- a split that is not a split, wearing a perfect score. A 0% cut is a tell.
+    # So attachment stops at the cap, the part count is whatever the structure and the cap allow,
+    # and the caller is told when that is more than they asked for.
+    while True:
+        keys = sorted(groups, key=lambda k: sum(size[i] for i in groups[k]))
+        if len(keys) <= parts:
+            break
+        moved = False
+        for k in keys:
+            ksz = sum(size[i] for i in groups[k])
+            best, best_w = None, 0
+            for other in keys:
+                if other == k or sum(size[i] for i in groups[other]) + ksz > cap:
+                    continue
+                ww = edge_between(groups[k], groups[other])
+                if ww > best_w:
+                    best, best_w = other, ww
+            if best is not None:
+                groups[best] += groups[k]
+                del groups[k]
+                moved = True
+                break
+        if not moved:
+            break            # nothing can grow without breaching the cap; report what there is
+
+    out = sorted(groups.values(), key=lambda g: -sum(size[i] for i in g))
+    gid = {i: n for n, g in enumerate(out) for i in g}
+    cut = sum(n for (a, b), n in w.items() if gid[a] != gid[b])
+    stats = {"cut": cut, "total_refs": sum(w.values()), "cap": cap, "capped_merges": capped,
+             "singletons": sum(1 for g in out if len(g) == 1), "asked": parts}
+    return out, stats
+
+
+def print_clusters(regions, edges, parts: int) -> None:
+    body = {r["index"]: r for r in regions if r["role"] == "body"}
+    size = {i: r["end"] - r["start"] + 1 for i, r in body.items()}
+    groups, st = propose_clusters(regions, edges, parts)
+    share = st["cut"] / st["total_refs"] if st["total_refs"] else 0.0
+    over = (f" -- you asked for {st['asked']}, and the reference structure plus that size cap do "
+            f"not permit fewer" if len(groups) > st["asked"] else "")
+    print(f"\n== proposed seams: {len(groups)} group(s), target ~{st['cap']} lines each =={over}")
+    print(f"   {st['cut']} of {st['total_refs']} references cross a boundary ({share:.0%}) -- that "
+          f"is the promote-to-header list")
+    print(f"   {st['capped_merges']} merge(s) the references wanted were refused by the size cap; "
+          f"a group AT the cap is the cap talking, not a seam")
+    print(f"   {st['singletons']} group(s) of one region\n")
+    for n, g in enumerate(groups):
+        lines = sum(size[i] for i in g)
+        at_cap = " <- at the cap" if lines >= st["cap"] else ""
+        tops = sorted(g, key=lambda i: -size[i])[:3]
+        names = ", ".join(f"{body[i]['name']}({size[i]})" for i in tops)
+        print(f"  [{n}] {len(g):>4} region(s) {lines:>6}L{at_cap}")
+        print(f"       {names[:96]}")
+        print(f"       plan: {' '.join(str(i) for i in sorted(g))}"[:300])
+
+
+def selftest() -> int:
+    """Pin propose_clusters. Pure arithmetic over the reference matrix, so it runs without libclang.
+
+    The clustering is the part that decides where a split cuts, and its failure mode is not a crash:
+    it is a confident proposal that is wrong. Both arms below were live defects during development.
+    """
+    bad = 0
+
+    def check(cond, label):
+        nonlocal bad
+        if not cond:
+            print(f"  [FAIL] {label}")
+            bad += 1
+
+    # Two tight clusters, 100 lines each, joined by ONE weak edge. The seam is obvious by
+    # construction, which is what makes it a usable oracle.
+    regions = [{"index": 0, "start": 1, "end": 1, "role": "open", "name": "ns"}]
+    line = 2
+    for i in range(1, 5):
+        regions.append({"index": i, "start": line, "end": line + 24, "role": "body",
+                        "name": f"a{i}"})
+        line += 25
+    for i in range(5, 9):
+        regions.append({"index": i, "start": line, "end": line + 24, "role": "body",
+                        "name": f"b{i}"})
+        line += 25
+    edges = {1: {2: 50, 3: 50, 4: 50}, 2: {3: 50, 4: 50}, 3: {4: 50},
+             5: {6: 50, 7: 50, 8: 50}, 6: {7: 50, 8: 50}, 7: {8: 50},
+             4: {5: 1}}
+    groups, st = propose_clusters(regions, edges, 2)
+    check(len(groups) == 2, f"two clusters joined by one weak edge split in two (got {len(groups)})")
+    check(all(len(g) == 4 for g in groups), f"and evenly (got {[len(g) for g in groups]})")
+    check(st["cut"] == 1, f"cutting exactly the weak edge (got {st['cut']})")
+
+    # THE CAP MUST BIND DURING ATTACHMENT. Without it, "attach the smallest group to whatever it
+    # references most" funnels everything into the largest connected component: measured on
+    # gpu_capture.cpp it produced ONE 6,374-line group out of 6,515 and reported a 0% cut -- a
+    # split that is not a split, wearing a perfect score. A 0% cut over a non-trivial graph is a
+    # tell, not a success, so it is asserted against directly.
+    cap = st["cap"]
+    check(all(sum(regions[i]["end"] - regions[i]["start"] + 1 for i in g) <= cap * 1.5
+              for g in groups),
+          "no group runs away past the size cap")
+    # The fixture has to make the FIRST pass leave more groups than asked for, or the attachment
+    # loop never runs and the arm certifies nothing -- checked by mutation, not assumed. Six regions
+    # of 100 lines with parts=4 gives a cap of 150, so no two can merge (200 > 150): the greedy
+    # leaves six, attachment is entered, and the cap is the only thing standing between six groups
+    # and one.
+    big = [{"index": 0, "start": 1, "end": 1, "role": "open", "name": "ns"}]
+    ln = 2
+    for i in range(1, 7):
+        big.append({"index": i, "start": ln, "end": ln + 99, "role": "body", "name": f"c{i}"})
+        ln += 100
+    dense = {i: {j: 10 for j in range(1, 7) if j != i} for i in range(1, 7)}
+    dgroups, dst = propose_clusters(big, dense, 4)
+    check(len(dgroups) == 6,
+          f"the size cap holds during attachment; without it everything funnels into the largest "
+          f"connected component (got {len(dgroups)} group(s))")
+    check(dst["cut"] > 0, "and a 0% cut is not manufactured by merging everything")
+
+    # A region nothing references has nothing to attach to; it must survive rather than vanish.
+    lone, _ = propose_clusters(regions, {1: {2: 5}}, 2)
+    check(sum(len(g) for g in lone) == 8, "every body region appears in exactly one group")
+
+    print("== PASS ==" if not bad else f"== FAIL: {bad} ==")
+    return 1 if bad else 0
+
+
 def main() -> int:
     ap = argparse.ArgumentParser(description=__doc__,
                                  formatter_class=argparse.RawDescriptionHelpFormatter)
-    ap.add_argument("--file", required=True, type=pathlib.Path)
+    ap.add_argument("--file", type=pathlib.Path)
+    ap.add_argument("--selftest", action="store_true",
+                    help="pin the clustering arithmetic; needs no libclang and no build")
     ap.add_argument("--build", default="prosper/build-linux")
     ap.add_argument("--json", type=pathlib.Path)
+    ap.add_argument("--clusters", nargs="?", type=int, const=6, default=None,
+                    metavar="PARTS",
+                    help="suggest seams: group regions that reference each other (default 6 parts)")
     ap.add_argument("--min-lines", type=int, default=0,
                     help="only print regions at least this many lines long")
     args = ap.parse_args()
+    if args.selftest:
+        return selftest()
+    if not args.file:
+        sys.exit("--file is required")
 
+    _need_clang()
     root = repo_root()
     target = (root / args.file) if not args.file.is_absolute() else args.file
     if not target.exists():
@@ -339,6 +544,9 @@ def main() -> int:
         inn = sum(v for s, d in edges.items() for t, v in d.items() if t == r["index"])
         print(f"  [{r['index']:4d}] {r['start']:6d}-{r['end']:<6d} {span:6d}L  "
               f"out={out:<4d} in={inn:<4d} {r['role']:<9s} {r['kind']:<22s} {r['name'][:52]}")
+
+    if args.clusters is not None:
+        print_clusters(regions, edges, args.clusters)
 
     if args.json:
         # The digest is what lets split_file.py prove the map still describes the file. Without it
--- a/prosper/tools/refactor/map_symbols.py
+++ b/prosper/tools/refactor/map_symbols.py
@@ -563,9 +563,14 @@ def main() -> int:
         # fail if the file GREW -- an assertion that cannot fail on an edit is not a check.
         import hashlib
         digest = hashlib.sha256(target.read_bytes()).hexdigest()
+        # RECORD THE PARSE ERRORS IN THE MAP. The warning above goes to a terminal nobody keeps,
+        # and the map outlives it -- so a consumer reading this file has no way to tell a complete
+        # map from one built on a partial AST, and prints its own reassuring [ok] either way. A
+        # libclang parse driven from a g++ database fails softly, so this is not a rare case.
         args.json.write_text(json.dumps({"file": str(target.relative_to(root)),
                                          "sha256": digest,
                                          "total_lines": total_lines,
+                                         "parse_errors": len(fatal),
                                          "regions": regions,
                                          "edges": {str(k): v for k, v in edges.items()}}, indent=1))
         print(f"  wrote {args.json}")
--- a/prosper/tools/refactor/map_symbols.py
+++ b/prosper/tools/refactor/map_symbols.py
@@ -416,7 +416,12 @@ def print_clusters(regions, edges, parts: int) -> None:
         names = ", ".join(f"{body[i]['name']}({size[i]})" for i in tops)
         print(f"  [{n}] {len(g):>4} region(s) {lines:>6}L{at_cap}")
         print(f"       {names[:96]}")
-        print(f"       plan: {' '.join(str(i) for i in sorted(g))}"[:300])
+        plan = " ".join(str(i) for i in sorted(g))
+        # NOT truncated. This line exists to be copy-pasted into split_file.py; a silent cut at 300
+        # characters drops every region past about the hundredth and yields a plan that is a valid
+        # prefix of the right answer, which is the worst failure available here. split_file.py
+        # catches an incomplete partition, but only after the caller has run it.
+        print(f"       plan: {plan}")
 
 
 def selftest() -> int:
@@ -481,7 +486,11 @@ def selftest() -> int:
 
     # A region nothing references has nothing to attach to; it must survive rather than vanish.
     lone, _ = propose_clusters(regions, {1: {2: 5}}, 2)
-    check(sum(len(g) for g in lone) == 8, "every body region appears in exactly one group")
+    seen = [i for g in lone for i in g]
+    # The SET and the COUNT, not just the count: a total of 8 is equally consistent with one region
+    # duplicated and another dropped, which is precisely what a bad merge produces.
+    check(sorted(seen) == list(range(1, 9)) and len(seen) == 8,
+          f"every body region appears in exactly one group (got {sorted(seen)})")
 
     print("== PASS ==" if not bad else f"== FAIL: {bad} ==")
     return 1 if bad else 0
--- a/prosper/tools/refactor/promote_internal.py
+++ b/prosper/tools/refactor/promote_internal.py
@@ -203,6 +203,34 @@ def canonical_include(source_rel: str, header_name: str) -> str:
     return header_name
 
 
+def preprocessor_depth_after(text: str) -> int:
+    """Net `#if` nesting left open by TEXT. Zero means the text is balanced on its own."""
+    depth = 0
+    for raw in text.splitlines():
+        d = raw.strip()
+        if d.startswith("#if"):
+            depth += 1
+        elif d.startswith("#endif"):
+            depth = max(0, depth - 1)
+    return depth
+
+
+def declarations_needed(promote, edges, regions) -> list[int]:
+    """Regions the PROMOTED code calls that are staying behind, so the header must declare them.
+
+    Every such destination, not merely the ones referenced from outside the file. That narrower
+    rule is what this replaced, and it is a different set: an ordinary internal helper called only
+    from within this file was never "public", so no declaration was emitted for it, the header used
+    it above its definition, and the build failed with `not declared in this scope` -- for
+    `checked_mul`, `resource_footprint_impl`, `blob_max_bytes_from_env` and two others in one run.
+    The `[warn] no declaration could be derived` branch could not fire for them either, because
+    they never reached it.
+    """
+    kept = set(promote)
+    return sorted({d for i in promote for d in edges.get(i, set())
+                   if d not in kept and d in regions and regions[d].get("role") == "body"})
+
+
 def build(map_data: dict, original: str, promote: list[int], header_rel: str,
           namespace: str, guard_note: str, include_spelling: str = "",
           forward_decls: list[str] | None = None) -> tuple[str, str, list[str]]:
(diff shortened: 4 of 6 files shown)


mattias800 (commented):
**REJECTED** on one measured blocking finding. The three fixes are individually well-reasoned and fix 1 is exactly right; the blocker is that fix 2 solves only one of the two things that follow from the fact it discovered, and the PR's "builds clean" is a Linux-only statement about a change whose failure is Windows-only.

Reviewed at head `43f1e3e6d`.

---

## BLOCKING — the generated `.cpp` puts its own internal-header include *inside* the conditional the preamble leaves open

`build()` learns at `promote_internal.py:257` that the copied preamble ends at conditional depth 1, and balances the header. Thirty-nine lines later the *same* function uses the *same* preamble end as the insertion point for the include it adds to the source:

```python
296:    last_pre = max(r["end"] for r in preamble)
297:    insert_at = last_pre
...
301:    out_lines.insert(insert_at, f'#include "{include_spelling}"\n')
```

For `gpu_capture.cpp` the last preamble line is 46 and the `#endif` is line 47, so the include lands *between* them — inside `#if defined(__linux__) || defined(__APPLE__)`. Meanwhile the promoted definitions are deleted from the `.cpp` unconditionally. On Windows the header is therefore not included and the definitions are gone.

Measured at this head, not inferred. Real (non-dry-run) tool invocation on `src/gpu/capture/gpu_capture.cpp` with the tool's own suggested closure `--regions 45,46,47,48,53,56,66,67,68,69,70,71,72,73,74,75,76,77` — five `[ok]` lines, exit 0 — produced:

```
#if defined(__linux__) || defined(__APPLE__)
#include <sys/uio.h>
#include <unistd.h>
#include "host/platform/posix_shim.hpp"
#include "gpu/capture/gpu_capture_internal.hpp"     <-- inside the conditional
#endif
```

Syntax-only compiles of that generated pair, same flags, same tree:

| target | result |
| --- | --- |
| `g++ -fsyntax-only` (Linux) | exit 0, 0 diagnostics |
| `x86_64-w64-mingw32-g++ -fsyntax-only` | exit 1, **245 errors** — `'Writer' was not declared in this scope`, `'Reader' was not declared in this scope`, ... |
| **positive control**: unmodified `gpu_capture.cpp`, mingw, identical flags | exit 0, **0 errors** |

The control matters: the toolchain compiles this file for Windows perfectly well, so the 245 errors are the promotion's, not the cross-compiler's.

Why I am treating a line the diff does not touch as blocking rather than as follow-up:

- It is the same root cause the PR discovered, in the same function, and fix 2's comment reasons at length about that exact `#if` while concluding only about the header.
- The PR converts a loud, Linux-visible failure (`unterminated #if`, immediately) into a Linux-silent, Windows-only one. That is the direction this project's charter treats as worse, and the PR's own thesis is "it reports success and the cost lands at build time on someone else".
- `Verification` says "**builds clean**" with no platform qualifier, and the platform where it does not build is not the one that was tested.

Minimal fix: insert at the last preamble line whose conditional depth is 0, rather than at the last preamble line. `conditional_depth()` (`:50`) already computes exactly that — for `gpu_capture.cpp` it puts the include after line 42, immediately above the `#if`. The internal header carries the whole preamble itself, so it does not need to sit below the platform includes. `verify()` is unaffected (it strips the include by exact match).

---

## Should fix — the refusal uses the predicate this file documents as wrong

```python
727:    unbreakable = [i for i in wanted if anon.get(i)]
```

`promote_internal.py:128-143` exists to say that `anon` alone is not linkage, ends with "A name is not an identity; the USR is", and records that a reviewer proved the point by mutation. `region_is_external()` (`:139`) is the predicate that got it right. The new guard reaches past it.

The case it misses: a `static` free function at namespace scope is internal linkage without being in an anonymous namespace, so `anon.get(i)` is false, so it is not refused — and `prototype_of` emits its declaration verbatim, `static` included. That is a `static` declaration in a header, which is precisely the ODR hazard `needs_inline`'s own docstring spells out at `:146-166`.

Concrete instance in the file this PR was verified on: `src/gpu/capture/gpu_capture.cpp:2626`, `ordered_gpustate_capture_reader` (region 155, USR `c:gpu_capture.cpp@N@prosper@N@gpu@F@...`, `anon` false). `prototype_of` on it returns

```
static CaptureMemoryReader ordered_gpustate_capture_reader(const GpuState& state);
```

Honest caveat, since I checked rather than assumed: this is **not** reachable on `gpu_capture.cpp` today. All three of its callers (regions 156, 157, 201) are filtered out as `public` before `declarations_needed` runs, so it never enters `wanted`, and the author's 55-region run would not have hit it. The shape is common in the tree though — 258 `^static` definitions under `prosper/src` — and the fix is to reuse `region_is_external` for the guard instead of `anon`, which is one expression.

## Should fix — the suggested `--regions` closure ignores `--exclude`, so the printed re-run reproduces the refusal

Measured. `--regions 45,...,77 --exclude 45`:

```
refused: 1 definition(s) ... kMaxBlobDefaultBytes
The promote set must be closed under what it calls. Re-run with:
  --regions 45,46,47,48,53,56,66,67,68,69,70,71,72,73,74,75,76,77
```

The suggestion contains 45, which `:566` drops again on the next run, giving the identical message. The user has no exit but to notice this themselves; the parenthetical ("or add them to `--exclude`") restates the state they are already in. Either filter `excluded` out of the closure and say the promotion is impossible under this `--exclude`, or name the conflicting region. The same shape exists for regions removed by the `incomplete` block-drop above.

## Test gap — the preamble arm asserts *balance*, and *placement* is the load-bearing property

Verified by mutation. Move the `#endif` append from the end of the preamble (`:257-261`) to the end of the header, after the closing namespace brace: `--selftest` stays **green**, while every emitted declaration and the whole `namespace prosper::gpu` would then sit inside `#if defined(__linux__) || defined(__APPLE__)` and vanish on Windows. The arm's own label is accurate — it is the property behind it that is one step off.

Secondary: the arm checks `build()`'s output with `preprocessor_depth_after`, the same function `build()` used to decide what to append, so a shared error in the counter is invisible to it. The two literal arms just above give the counter independent coverage, which mostly covers this, but a direct `#endif` count or an assertion that the `#endif` precedes `namespace ` would close both holes at once.

## Minor

- `preprocessor_depth_after` (`:206-215`) re-implements `COND_OPEN` / `COND_CLOSE` (`:46-47`) more weakly: `startswith("#if")` does not match `#  if`. That spelling is in this tree — `prosper/src/hle/libc/hle_libc.cpp:20` and `:26`. Reuse the regexes 160 lines above rather than a second, weaker matcher.
- `:251` "closes it **far below**, outside any preamble region". Measured: the `#endif` is `gpu_capture.cpp:47`, one line after the last preamble line (46); it falls in region 38, the `NAMESPACE_OPEN prosper` region spanning 47-48. The load-bearing half ("outside any preamble region") is right; "far below" is not, and a committed comment carrying a wrong fact is the thing this repository keeps paying for.
- `:253-256` "every directive copied is an include-guard-style condition around includes" is a property of this one file, not of preambles. `map_symbols.py:199-205` marks the preamble **by position** and says so explicitly ("a stray declaration above the first namespace still belongs in every part"), so a preamble may contain declarations.
- Pre-existing, drive-by: the generated header still carries the hardcoded note about `rdna2_to_spirv.cpp` and "nothing outside `src/gpu/recompiler/` should include this header" — now written into a `gpu/capture` header. It is a false statement in generated source.

---

## Confirmed good

- **Fix 1 is correct**, and `role == "body"` is the right filter: preamble regions are already copied into the header wholesale, and `open`/`close` carry no declarations — there is an arm for that. The comment at `:712` and its code now say the same thing, which is the whole point.
- **The fixed-point loop terminates.** `grown` is a union with `closure`, so it is monotone over a finite index set and the equality break fires within `|regions|` iterations; a reference cycle is absorbed by the union rather than revisited. Measured: `--regions 67..77` refused and suggested an 18-region set, and that set is accepted in one further run (exit 0, 18 regions promoted, 11 gained `inline`).
- **The refusal does not appear to over-fire.** Every entry in `wanted` is genuinely unresolvable from the header, because the header is included above the anonymous namespace and nothing left behind is visible to it. There is one general over-fire channel — `map_symbols.py:271-291` adds an edge from any region whose *text* mentions a function template's name, comments included — but `gpu_capture.cpp` has 0 `FUNCTION_TEMPLATE` regions, so it cannot fire here; worth knowing before running this on `rdna2_to_spirv.cpp`, where the closure would over-promote rather than dead-end.
- **Both mutation claims in the PR body reproduce**, each reddening exactly one arm: restoring a `public`-style filter inside `declarations_needed` reddens only *"a promoted region's callees that stay behind are ALL declared, public or not"*; deleting the balancing block reddens only *"the emitted header is preprocessor-balanced even when the preamble is not"*.
- Hygiene: 0 `Claude-Session:` trailers, `git diff --check` clean, no absolute host paths in the diff or in the tool's output.

## What I verified, and what I did not

Verified here: `--selftest` and both mutations; the refusal and closure path end-to-end on `gpu_capture.cpp` (map regenerated with `map_symbols.py` against this head); the emitted header compiling clean on Linux; the emitted pair failing on the Windows target with a clean positive control.

Not reproduced: the author's specific 55-region closure, the 6,515 -> 5,208 shrink, and `ctest --no-tests=error` 455/455. I did not run a full build, so those I take on report — noting only that a Linux `ctest` cannot see the finding above.

mattias800 (commented):
## Review round 2 — head `b8e7782f9`

**REJECTED**, and only on three one-line items, none of them in the round-1 defect. The
load-bearing fix is right and I reproduced it independently; what blocks is that three claims
this PR makes about itself are not true of the code it ships.

### The round-1 blocker is fixed, reproduced from scratch

I did not take the author's table. Own worktree, own map, own closure, `promote_internal.py`
run for real on `src/gpu/capture/gpu_capture.cpp`, then `-fsyntax-only` under both compilers
(GCC 16.1.1 / Fedora MinGW 16.1.1). My set is the 39-region closure the tool proposed from
cluster [2]'s plan, not the author's 55, so the error count differs; the asymmetry does not.

| arm | include lands at | `g++` | `x86_64-w64-mingw32-g++` |
| --- | --- | --- | --- |
| control (unmodified file) | -- | 0 errors | 0 errors |
| fixed (head) | line 48, after the `#endif` | 0 errors | **0 errors** |
| mutant (`insert_at = last_pre`, selftest gate bypassed) | line 47, inside `#if defined(__linux__)` | 0 errors | **87 errors** |

The mechanism is confirmed at the source rather than inferred: `gpu_capture.cpp`'s last
preamble region ends at line 46 and line 47 is `#endif`, so `conditional_depth(lines)[47] == 1`
and the walk in `promote_internal.py:305-307` moves the insertion point to 47, putting the
include at line 48. The bypass was necessary exactly as described -- `main()` runs
`if selftest(): return 1`, and the placement arm is red under the mutation, so a naive mutant
run leaves the file untouched and reports 0/0.

### Selftest arms: mutation-checked, one arm each

Four mutations against the head, each reddening **only** its own check and nothing else:

| mutation | reddens |
| --- | --- |
| drop the depth-0 walk in `build()` | `the include added to the SOURCE lands outside any preprocessor conditional` |
| `open_depth = 0` (no preamble balancing) | `the emitted header is preprocessor-balanced even when the preamble is not` |
| `declarations_needed` returns `[]` | `a promoted region's callees that stay behind are ALL declared, public or not` |
| drop the cap test in the attachment loop (`map_symbols.py:378`) | `the size cap holds during attachment ... (got 4 group(s))` |

The last one is the arm the author reports having had to rebuild, and the replacement fixture
does discriminate: six 100-line regions at `parts=4` gives `cap=150`, the greedy pass merges
nothing, the attachment loop is entered, and removing the cap collapses six groups to four.

### `propose_clusters`: I could not break it

Termination and partition hold on every degenerate input I could construct. The attachment loop
either merges (strictly reducing the group count) or breaks, so it terminates unconditionally.
Measured: `parts=0` -> one group; `parts=-3` -> one group, terminates; `parts=99` over 3 regions
-> 3 singletons, `cap=1`, reported as more than asked; disconnected graph at `parts=1` -> 4
groups, nothing force-merged across a zero edge; zero body regions -> empty, no divide-by-zero.
`cap = total // max(1, parts)` is floored at 1 so a tiny file yields singletons rather than a
crash. Every body region appears in exactly one group in all of them.

`--clusters 3` on `gpu_capture.cpp` reproduces the README block verbatim: 9 groups, `~2149`
lines each, 252 of 2927 crossing (9%), 83 capped merges.

### Verified, no finding

- Scope: five files, `prosper/tools/refactor/*` plus `prosper/CMakeLists.txt`. Nothing under
  `src/`, `frontends/`, `tests/`.
- `ctest -R refactor_ --no-tests=error` -> **3/3 passed**, exit 0 (`refactor_map_symbols_clusters`,
  `refactor_survey_classifier`, `refactor_survey_measurement`).
- Lazy libclang: `survey_sizes.py` is the only importer of `map_symbols` in the tree, and it calls
  `mod._need_clang()` at `survey_sizes.py:105` before any `MS.` use. `ci` is dereferenced in
  `map_symbols.py` only at `:187` (reached through `regions_of`) and `:522-525` (after
  `_need_clang()` in `main()`), so no caller can reach a `None`.
- Survey at `--min-lines 2500` -> `EXTRACT=13, HAND=1, PARSE-FAIL=1, SPLIT=9, UNPARSED=4`, as claimed.
  Both selftests also pass *with* libclang present -- all four measurement fixtures run, hard link
  included.
- Refusal predicate: `not externally_linked(i)` is the correct test and cannot refuse a set the old
  one would have accepted usefully. A `static` free function at namespace scope is `prosper::gpu::f`,
  so a non-`static` header declaration followed by the `static` definition is ill-formed
  ("static declaration follows non-static declaration"); the old `anon.get(i)` let exactly that
  through. Strictly broader, and broader in the safe direction.
- `--exclude` in the closure: measured. Excluding region 90 (`resource_footprint_impl`) drops it
  and its now-unreachable callees 88/89 from the suggested `--regions`, and the message names it
  under "have to be removed from the promoted code instead". The re-run with the suggested list
  refuses again -- correctly, since the residual really is stuck -- but the message still leads with
  a `Re-run with: --regions <identical list>` line that is a no-op in that state. Not a finding;
  worth suppressing the suggestion when the closure equals the input and `stuck` is non-empty.
- Hygiene: 0 `Claude-Session:` trailers, `diff --check` clean, no absolute host path in any added line.

### Blocking findings

**1. `--json` still publishes an absolute host path, which is the defect the change claims to fix.**
`survey_sizes.py:664` rewrites only the `file` field. The printed table strips the root from **two**
fields -- `file` at `:641` and `why` at `:643` -- and `:637-639` says why: *"the strip used to be on
the scored branch only, so exactly the rows that name a problem leaked one."* That sentence is now
true of the JSON path. Reproduced on the documented invocation
(`survey_sizes.py --min-lines 2500 --json ...`), where `PARSE-FAIL=1` is not hypothetical:

```
"why": "no compile command found for /home/<user>/repos/<...>/prosper/frontends/prosper-app/main.cpp,
        and no TU includes it"
```

The string comes from `map_symbols.py:102`, which formats an absolute `target`. The new comment at
`:658-662` -- *"the absolute form published the developer's account name and directory layout"* --
still describes the shipped behaviour for every non-scored row. Apply the same `.replace(str(root)...)`
to `why` in the `rel_rows` comprehension.

**2. The unmeasured lambda share is still emitted on the `UNPARSED` path, in the same shape as a
measured one.** `survey_sizes.py:373` returns `"lambda_lines": 0, "lambda_share_of_biggest": 0.0`
unconditionally, and that early return is never gated on `DOMINANT_SHARE`. Measured on the same
survey: 4 of 28 rows -- `hle_file.cpp`, `hle_kernel.cpp`, `hle_kernel_mem.cpp`,
`guest_write_watch.cpp` -- carry `lambda_share_of_biggest: 0.0` in the JSON and print `0%` in the
`lam%` column, beside `test_recompile_coverage.cpp`'s real `2%`. That is exactly the confusion
`:403-409` was added to remove, and the new comment there -- *"Only present when it was actually
COMPUTED"* -- is false of the same function twenty lines up. Either drop the two keys from the
`UNPARSED` return, or the comment overstates what the fix does.

**3. A comment correction the PR body reports as landed did not land.** The body says: *"a comment
claiming the `#endif` sits 'far below' now says 'one line below', which is true."*
`promote_internal.py:253` still reads `closes it far below, outside any preamble region`. The
`#endif` is line 47 and the last preamble line is 46 -- the very next line, which is precisely why
the round-1 defect was reachable. The rest of that comment checks out (the copied preamble really
does carry 3 `#if`-family directives against 2 `#endif`, net depth 1, and the third is the
`__linux__` one), so it is the single word that is wrong -- but a stated-and-unmade correction is
the failure mode this project's charter is loudest about, and it sits in the comment block that
explains the bug being fixed.

### Non-blocking

- `map_symbols.py:419` truncates the `plan:` line at 300 characters with no ellipsis. It is the one
  line the reader is meant to copy. `gpu_capture.cpp`'s largest group is 273 characters -- under by
  27 -- but a 150-region group renders at exactly 300 and silently loses everything past region 99.
  Downstream this fails loudly (`split_file.py:157-159` reports unassigned body regions), so it costs
  a round trip rather than a wrong split; still, either drop the slice or mark the cut.
- The `:139-151` citation added at `promote_internal.py:782` points at `region_is_external`'s
  *body* (139-143) and then into an unrelated comment block; the documentation it means is the
  block at `:129-138`. The companion `:146` citation is exact.
- `check(sum(len(g) for g in lone) == 8, "every body region appears in exactly one group")` in
  `map_symbols.py:483-484` asserts only the total, so a simultaneous duplication and loss would pass.
  The invariant does hold structurally; the arm just proves less than its label says.

Nothing above touches the promotion path itself, which I believe is correct. Fix the three and I
expect to approve on the next head without re-running the compiler A/B.

mattias800 (commented):
## Review round 3 (delta) — head `8befcc0fb`

**APPROVED.**

All three blocking items are fixed, and I checked each by measurement rather than by taking the
report. The four non-blocking items are applied. Answers to the three questions you posed are below,
and one of them turned up something live — but it is in the *older* path, not in this delta.

### I did not re-run the compiler A/B, and here is why that is sound

`build()` is byte-identical between `b8e7782f9` (the head I verified with mingw) and this one. The
only non-comment change anywhere in `promote_internal.py` is the refusal-message branch:

```
git diff b8e7782f9 8befcc0fb -- prosper/tools/refactor/promote_internal.py
  # ...only the `closure != sorted(promote)` branch, plus two comment edits
```

The success path confirms it behaviourally: the same 39-region closure still reports
`header carries 816 line(s)` and `shrink ... to 5700 lines`, identical to round 2.

### The three blocking items

**1. `--json` paths.** Fixed, and measured on the documented invocation
(`--min-lines 2500 --json`): across all 28 rows, **0 fields contain the home directory or any
absolute path**. The `PARSE-FAIL` row's `why` reads
`no compile command found for prosper/frontends/prosper-app/main.cpp, and no TU includes it`.

**2. Unmeasured lambda share.** Removed. **14 of 28** rows carry `lambda_share_of_biggest` — the 13
`EXTRACT` plus the 1 `HAND`, i.e. exactly the dominated set — and 0 `SPLIT`, 0 `UNPARSED`,
0 `PARSE-FAIL`. The four `UNPARSED` rows now print `--` in the `lam%` column.

**3. The "far below" comment.** `promote_internal.py:253` now reads *"closes it on the very next
line"*, which is true (preamble ends 46, `#endif` is 47). Your account of why it survived a full
round is the more valuable half: a bare `str.replace` with no assert in a batch where every other
edit asserted. That is the same failure family as the `--selftest` gate the tool itself enforces.

### The four non-blocking items

- **Partition arm.** Mutation-checked, not assumed. Corrupting the partition at return time
  (duplicate one region, drop another — total stays 8): the **old** count-only arm reports
  `== PASS ==`, exit 0; the **new** set comparison reddens with
  `got [1, 3, 3, 4, 5, 6, 7, 8]`. A real counter-arm.
- **`plan:` untruncated.** 150 regions renders 504 characters and all 150 ids, `ids == 1..150`.
- **Citation** now `:129-134`, which is the explanatory block. Correct.
- **`--exclude` refusal.** Both branches exercised on the real map. Round one still grows the closure
  and prints a usable list; at the fixed point it prints *"there is no --regions list that fixes it
  -- printing the one you typed would be a no-op."* That is the right message.

### Your three questions

**Is `_rel()` itself a one-path change? Any other field that can carry a path?**

Field census over every return path in `measure()`/`survey_one()`, cross-checked against the 28 rows
actually produced: exactly four string-valued fields exist — `file`, `why`, `verdict`, `biggest`.
`_rel` covers the first two; `verdict` is an enum. **`biggest` is the one it does not cover, and it
can carry a path shape.** It is `c.spelling` (`map_symbols.py:172`), and preamble regions survive
into `sized` because `REPLICATED` is `open`/`close` only (`survey_sizes.py:361-367`), so an
`INCLUSION_DIRECTIVE` region can be the dominant one — its spelling is the include *as written*
(`sys/uio.h`, `host/platform/posix_shim.hpp` in this file's own map). For that to be absolute a
source would have to write `#include "/..."`, and `git grep '#include *"/'` over the tree returns
nothing. So coverage is complete today, for a reason that lives outside this file. One sentence in
the comment beats a code change.

**Substring or anchored: substring is correct, and anchoring would be a regression.** The
reproducing case has the path *mid-sentence* (`map_symbols.py:102` formats an absolute `target` into
`no compile command found for <abs>, and no TU includes it`), so a `startswith` strip misses exactly
the row that motivated the fix. The order — `root + "/"` then bare `root` — is longest-first and
matches the table. The other absolute-path producer is `survey_sizes.py:238`
(`clang_getSkippedRanges returned NULL for {path}`); `:230` and `:235` use `path.name` and are
already safe.

**The remaining one-path gap runs the other way, and it is live.** The table at `:641` still has its
own, weaker implementation: `r["file"].split("prosper/", 1)[-1]`. That is not root-anchored — it
splits on the first literal `prosper/` *anywhere* in the string. For a checkout whose root path
itself contains a `prosper/` segment (this machine has one; see the gitignored `LOCAL.md`), a root
of `/srv/prosper/worktrees/lane-a` prints:

```
table strip  -> 'worktrees/lane-a/prosper/src/gpu/capture/gpu_capture.cpp'
_rel  strip  -> 'prosper/src/gpu/capture/gpu_capture.cpp'
```

— a leaked private directory segment, and a "relative" path that is relative to nothing. `_rel()` is
the correct implementation of the two, so the fix is to point `:641` at it rather than the reverse.
That also settles the comment: *"Repo-relative, like the printed table"* is still not quite true,
because the two produce **different spellings** (`prosper/src/...` versus `src/...`). Pre-existing
line, not a regression in this delta, so not blocking — but it is precisely the shape you asked
about, and worth a follow-up issue.

**Does removing the key break a consumer?**

Checked, not taken. Two in-tree readers and no external one:

- `survey_sizes.py:496` does `r.get("lambda_share_of_biggest")` then `lam_share == 0.0`. With the key
  absent that is `None == 0.0` -> False, so the arm goes **red loudly**, printing
  `lambda_share=None` and the full row. It fails in the safe direction. Its fixture
  `dominant_no_lambda.cpp` is dominated, so the key is present and the arm still passes —
  `refactor_survey_measurement` is green with libclang present, all four parse fixtures included.
- `survey_sizes.py:649` tests membership and prints `--`. Confirmed on all four `UNPARSED` rows.

Nothing else in the repository mentions either field, and nothing reads a survey JSON — I grepped
`*.py`, `*.md`, `*.yml`, `*.yaml`, `*.sh`, `*.ps1`, `*.txt` for both the field names and the tool
name. The README documents `--json` only as "machine-readable". So: no consumer breaks.

One nit rather than a finding: the JSON carries **no schema or version marker**, and this PR has now
changed its shape twice (paths relativised, keys made conditional). The consumer this artifact is
built for is an agent working from a copy pasted into an issue, which has no way to tell one shape
from another. A `"schema": 2` key costs nothing.

**Should the `plan:` line wrap?**

No — leave it running long, and the reasoning is not tidiness-versus-correctness, it is that wrapping
does not actually buy the tidiness. A wrapped copy-paste target is a *different* failure from a
truncated one, not a smaller one: paste one physical line of a wrapped plan and `split_file.py` sees
a fragment, which is the same incomplete partition you just removed, only harder to notice because
nothing looks cut. The line is prefixed and self-identifying, so a terminal soft-wrap is free. If
width ever genuinely matters, the answer is to emit the plan through `--json`, not to fold the human
line.

Worth noting that the line above it, `names[:96]` at `:418`, still truncates — correctly, because it
is descriptive rather than a copy-paste target. The two lines now follow opposite policies for a good
reason the code does not state; one clause would fix that.

### Nits

- `promote_internal.py:253-254`: the reflow left a four-word orphan line (`# shipped 3 \`#if\``).
- The fixed-point refusal still says *"1 of them are in --exclude and so are NOT in that list"*
  directly after the branch that explained no list was printed. "that list" now refers to nothing.

### State at this head

`ctest -R refactor_ --no-tests=error` -> **3/3 passed**, exit 0. All three `--selftest`s green, and
green again inside the container with libclang present (all four measurement fixtures run). Survey
`EXTRACT=13, HAND=1, PARSE-FAIL=1, SPLIT=9, UNPARSED=4`. Scope unchanged: `prosper/tools/refactor/*`
plus `prosper/CMakeLists.txt`. `diff --check` clean, 0 `Claude-Session:` trailers, no absolute host
path in any added line.

**APPROVED** on `8befcc0fb`.

mattias800 (commented):
## Review round 4 (delta) — head `331cae723`, `split_file.py` + README only

**REJECTED**, on one finding — but it is the finding this whole commit turns on: **the three new
selftest arms cannot fail.** The check itself is sound and I confirmed it works on real data; what
does not work is the thing that is supposed to keep it working.

### Blocking: the new arms are placed after the gate, and `split_file` has no ctest case

`split_file.py:373-392` appends the three cross-reference arms **after** `if bad: return 1`
(`:371-373`) and before an unconditional `return 0` (`:396`). So `bad` is incremented, `[FAIL]` is
printed — and `selftest()` returns **0**, having also printed
`[ok]   splitter self-test: replication, partition, tiling, #if, reconstruction`.

Measured, three independent mutations that each disable the new check:

| mutation | `[FAIL]` lines | prints the `[ok]` summary | **exit code** |
| --- | --- | --- | --- |
| `region_is_external` -> `return False` (linkage ignored) | 1 | yes | **0** |
| drop the `owner[dest] == owner[src]` same-part test | 1 | yes | **0** |
| `unresolved_across_parts` -> `return {}` (check disabled) | 1 | yes | **0** |

So your mutation check was reading the *printed line*, and the printed line is right — each mutation
does redden its own arm by name and only it, which I re-confirmed. The **verdict** never moves. This
also defeats `main()`'s own guard at `:409-410` (`if selftest(): return 1`): with the cross-reference
check broken, the tool still runs.

And there is no second net. **`split_file.py` is not registered with ctest** — `prosper/CMakeLists.txt`
registers `refactor_map_symbols_clusters`, `refactor_survey_classifier` and
`refactor_survey_measurement`, and nothing else in the repo or in `.github/workflows/` names
`split_file`. So `ctest -R refactor_ --no-tests=error -> 3/3` is true and covers **none** of this
delta. Between the two, the new check currently has zero automated coverage.

Two-line fix (move the arms above the `if bad` gate, or re-test `bad` before `return 0`), plus a
`refactor_split_file` `add_test`. This is the charter's *"a discriminator that cannot show its lever
moved is VOID"* in its loud form, which is the worse one: it prints FAIL and reports success, so it
reads as checked.

### The check itself is right, and it works on real data

Run directly against the real `gpu_capture.cpp` map with the `--clusters 3` group-[2] seam as the
plan: **31 stranded internal definitions**, led by `kMagic`, `kVersion`, `kEndian`, `kMaxFileBytes`,
`kMaxDraws` — the same family as the 25 names g++ reported on your promoted variant. It does what it
says when it runs.

### Your four questions

**1. False refusals when the USR is absent.**

`map_symbols` emits a **body** region with no USR in two ways, and I measured both rather than
reasoning about them. `mk()` records `"usr": c.get_usr() if c is not None else ""`, so `c is None`
gives `""` — but that is only the `TRAILER`, whose role is `close`, and the check already skips
non-`body` dests. The real source is `c.get_usr()` returning `""` for a real cursor. On two real maps:

| file | body regions | body regions with an empty USR |
| --- | --- | --- |
| `gpu_capture.cpp` | 155 | **0** |
| `hle_kernel.cpp` | 333 | **7** |

The seven are three `LINKAGE_SPEC`, three `INCLUSION_DIRECTIVE`, and one more `LINKAGE_SPEC` at
2143-2144. **`LINKAGE_SPEC` is `extern "C"`** — the single clearest form of *external* linkage in
C++ — and `region_is_external` returns `False` for it. That is not conservative, it is **inverted**.

Is it reachable? One edge in that map points at one: `thread_trampoline` -> region 269, x3. And here
the honest answer is that it does **not** currently produce a false refusal, for a reason that is
luck rather than design: region 269 is
`extern "C" uint64_t prosper_call_guest_on_stack(...);` — a *declaration*, which genuinely must
accompany its user, so refusing is right for the wrong reason. But regions 51-53 in the same file are
`extern "C" __attribute__((weak)) int prosper_thread_in_renderer_callback(unsigned long) { return 0; }`
— **external-linkage definitions**, visible across translation units, where splitting is legal and a
refusal would be false. Nothing references them today. So both shapes exist in one file and only the
benign one is wired up.

**So: failing toward refusal is not safe on the merits here, it is safe by accident of what this
particular file happens to reference.** And the premise that motivated the constant is wrong — you
do have an `anon` map: `promote_internal.anon_map()` derives the anonymous-namespace walk from the
`role: open` / `role: close` markers that are already in this same map, in ten lines you have already
written. It would classify a namespace-scope `extern "C"` block as external, which is correct.
`hle_kernel.cpp` is 5,395 lines and on the survey's own list, so this is the next file someone
splits.

**2. False negatives.**

Two channels, one measured-reachable and one argued.

- **Edges from a REPLICATED region are dropped.** `owner` is built from the plan, and replicated
  regions (`preamble`/`open`/`close`) are never in a plan, so `if src not in owner: continue`
  discards every edge leaving one. Measured: **3 such edges in `gpu_capture.cpp`, 18 in
  `hle_kernel.cpp`**, all from `NAMESPACE_OPEN` regions to internal body regions (`fbsd_errno`).
  A replicated region is copied into *every* part, so a genuine reference from one needs its dest
  visible in every part — which is exactly a stranding, and exactly what this check exists to catch.
  **Caveat, stated because I could not close it:** region 82's text is a comment block plus
  `namespace {`, and `fbsd_errno` is a `FUNCTION_DECL` rather than a template, so these particular
  18 look like boundary attribution rather than real uses. I verified `starts` is monotonic so
  `region_of`'s bisect is sound, and could not explain them further. Treat them as proof the *class*
  is populated, not as 18 strandings.
- **Inactive `#if` arms are invisible to the edge builder.** libclang parses one arm; the AST walk
  in `cross_refs` sees no cursors in the other, so no edges. The textual fallback covers
  `FUNCTION_TEMPLATE` names only. This is not a corner: `survey_sizes` measures
  `hle_kernel_mem.cpp` at **3,978 unparsed lines (52%)**, `hle_kernel.cpp` at **1,613 (30%)**,
  `hle_file.cpp` at 963, `guest_write_watch.cpp` at 495. A split of any of those is validated
  against at most half the references, and the direction of failure is the one that already cost
  this PR a round: it compiles where the tool was run and breaks on the other platform. `split_file`
  has a conditional check, but it is about separating a directive from its `#endif` — not about
  references. Worth one sentence in the `[ok]` line so nobody reads it as stronger than it is.

Macros and ADL I do **not** think are gaps: `PARSE_DETAILED_PROCESSING_RECORD` is on, so macro
definitions carry USRs and instantiations resolve through `.referenced`; ADL-resolved calls have a
resolved referent. And the folded-symbol worry I raised last round (`regions[-1]["symbols"].append`
making a region's USR unrepresentative) is **not reachable** — 0 regions with more than one symbol in
either map.

**3. A third shape both arms miss: a degraded parse.**

`map_symbols.py` on parse errors prints `WARNING N parse error(s); the map may be incomplete` and
**writes the map anyway**. The JSON records no flag that this happened — there is no `parse_errors`
field — so `split_file` cannot tell a clean map from a degraded one, finds few or no cross-part
edges, and prints `[ok] no internal-linkage definition is referenced from a part it is not in`. A
green light from an instrument that did not see the file. One field on each side fixes it.

I also checked the shape you would expect me to raise, `edges` absent entirely: measured, it yields
`{}` and passes silently. But `edges` and `sha256` entered the JSON in the **same commit**
(`ec0beb603`), and `main():421-423` already refuses a map with no `sha256`, so no tool-produced map
can reach it. Currently unreachable, structurally undefended — a one-line `if "edges" not in
map_data` would say so out loud. Not a finding.

**4. Regions in no part, and absent edges.**

Safe **as called**, and I got this wrong on the first attempt in a way worth repeating: my first test
removed a region from the part it was not in, saw "still flagged", and looked like a finding. Re-run
correctly — remove `kMagic` from *every* part — the flagged set goes 31 -> 30 and it is dropped
silently.

That is fine at the call site: `split()` returns `N body region(s) unassigned` in `problems`
(`split_file.py:157-159`) and `main():431-435` returns 1 before `unresolved_across_parts` runs. But
it is a **call-site invariant, not a function invariant**, it is undocumented in the docstring, and
the function is public and already called directly by the selftest with hand-made plans. One sentence
in the docstring, or an assert.

### The README claim, checked rather than accepted

Your `grep` claim is exact: before this commit `split_file.py` mentioned `edges` twice, at `:247` and
`:278`, both `"edges": {}` in fixtures. Two problems with the new section:

- **"skipping a step produces a result that passes every check and does not compile" is written in
  the present tense and is now false** — `split_file` refuses. The next paragraph says so
  ("`split_file` reads it now and refuses"), so the section contradicts itself two paragraphs apart.
  Past tense fixes it.
- **"The order is forced" overstates it.** A seam that cuts only across external-linkage boundaries
  needs no promotion — 42 of `gpu_capture.cpp`'s 155 body regions are external by USR — and the new
  check is precisely what makes that decidable. Its success line is the tool announcing the middle
  step was unnecessary. The chain is conditional, and saying so is a stronger claim, not a weaker one.

Minor: the new recipe writes both maps to `/tmp`. The charter's rule for this machine is
unconditional and its failure mode is shared (an exhausted tmpfs kills the Bash tool for every
concurrent agent). The maps are small — 81 KB and 144 KB measured — so they are not what fills it,
but the README is what the next agent copies. `$HOME` or a worktree-local scratch dir.
`split_file.py:54-56` already had `/tmp`, so this propagates rather than introduces it.

### Your reasoning about not re-running the survey and the mingw A/B

**Sound, and I agree.** `git diff 8befcc0fb 331cae723 --stat` is `README.md` and `split_file.py`
only — neither `survey_sizes.py` nor `promote_internal.py` is touched, so nothing those runs measured
can have changed. One correction to the statement, not the reasoning: `ctest -R refactor_
--no-tests=error -> 3/3` is true and is **not** evidence about this delta, because none of the three
runs `split_file.py`. Quote it as "3/3, none of which covers this change" until the case exists.

### State at this head

`split_file.py --selftest` exits 0 — which, per the blocking finding, it also does with the check
disabled. `diff --check` clean, 0 `Claude-Session:` trailers, scope `prosper/tools/refactor/*` plus
`prosper/CMakeLists.txt`, no absolute host path in any added line.

**REJECTED** on `331cae723`. The blocker is two lines plus an `add_test`; findings 1-4 above are
follow-ups I would not hold the PR for, except that finding 1's fix is also small and lands in the
same function.

mattias800 (commented):
## Review round 5 (delta) — head `f7114b8fd`

**REJECTED**, on one finding: **the replicated-edge fix produces a false refusal on a real in-tree
file.** You named this as the one you were least sure of, and asked me to settle whether those 18
edges are genuine. They are not — all of them are `namespace {` lines — and the new branch acts on
them. Everything else in this round is right, including the fix to my blocking finding.

### The blocking finding from round 4 is fixed, re-verified against the exit code

The gate is last (`split_file.py:452-455`). Six mutations, and this time I read `$?`, not the
printed line:

| mutation | exit | `[ok]` summary printed | arms reddened |
| --- | --- | --- | --- |
| `unresolved_across_parts -> {}` | **1** | no | 3 |
| linkage ignored | **1** | no | 2 (both "must not flag" arms) |
| no-USR -> internal (revert the fallback) | **1** | no | **1** — the no-USR arm |
| `anon_map` -> always False | **1** | no | **1** — the scope-walk arm |
| replicated branch disabled | **1** | no | **1** — the replicated arm |
| same-part test removed | **1** | no | **1** — the same-file arm |
| `verify_reconstruction -> []` (an arm from *before* the moved gate) | **1** | no | 1 |

Every new arm has a mutation that reddens it **alone**, and the last row confirms moving the gate did
not orphan the older arms. `ctest -R refactor_ --no-tests=error` -> **4/4**, and
`refactor_split_file` is one of them. Baseline exits 0.

### Blocking: the 18 replicated edges are `namespace {` lines, and the branch now refuses on them

I instrumented `cross_refs`' reference walk and dumped the referencing cursor for every edge whose
source region is replicated. On `hle_kernel.cpp`, **19 of 19** are a `NAMESPACE` cursor, and the
source line is the namespace brace itself:

```
src[82]  open 351-364   ref at line 364 (NAMESPACE '') -> [74] fbsd_errno
     source line 364: namespace {
src[126] open 811-817   ref at line 817 (NAMESPACE '') -> [74] fbsd_errno
     source line 817: namespace {
... 13 of these, all pointing at fbsd_errno
src[70]  open 241-249   ref at line 242 (NAMESPACE 'prosper') -> [44] prosper
```

The mechanism is in `map_symbols.py:274-284`: a re-opening of `namespace { }` is a cursor whose
`.referenced` is the namespace declaration, and `owner.setdefault(usr, idx)` bound that anonymous
namespace's USR to region 74 — the body region that happens to contain the first cursor carrying it.
So **every anonymous-namespace re-opening in the file is recorded as a reference to `fbsd_errno`.**

The consequence is a clean A/B. Plan built so that `fbsd_errno` and **all four** of its genuine
body-region callers land in the same part, leaving only the artifact edges crossing:

| head | `fbsd_errno` flagged? | total flagged |
| --- | --- | --- |
| `331cae723` (replicated edges dropped) | **no** — correct, the split is legal | 4 |
| `f7114b8fd` (replicated edges counted) | **yes**, "needed by b.cpp" | 5 |

`fbsd_errno` can live in only one part and there are 13 of those braces spread through the file, so
**every two-part plan for `hle_kernel.cpp` is now refused**, instructing the caller to promote a
function nothing across the seam actually uses. That is the direction you flagged as worst.
`gpu_capture.cpp` is unaffected — 0 replicated->internal-body edges — so the end-to-end result you
quote still stands; the regression is file-specific and real.

**And the class the branch was added to catch appears to be empty by construction.** An `open`/
`close` region holds only text clang attributed to no top-level cursor — comments, blank lines, the
brace, preprocessor lines — because any real code there would itself be a top-level cursor and
therefore its own body region; `preamble` holds includes. So there is no genuine reference for the
branch to rescue, and it trades a zero-sized false-negative class for a real false-positive one. My
round-4 caveat ("proof the class is populated, not 18 strandings") turns out to have resolved in the
negative, and I should have settled it then rather than leaving you to act on it.

Two fixes, either sufficient, the first better because it also removes the `[44] prosper` artifact
and improves every consumer of the matrix:

- **In `cross_refs`** (`map_symbols.py:274`), skip the edge when `ch.kind` or `ref.kind` is
  `CursorKind.NAMESPACE`. A namespace re-opening is not a reference.
- **In `unresolved_across_parts`**, restrict the replicated branch to `preamble` sources, or require
  the dest's name to appear textually in the replicated region's lines.

If the first is taken, I would drop the replicated branch again rather than keep an inert one — but
keeping it with an arm is defensible, and the arm you added is a good one.

### Your other two questions

**Is the hard `parse_errors` refusal right? Yes — and for a stronger reason than you gave.**

Your argument is that the cross-reference check is only as complete as its edges. True, and it
understates the case: a degraded parse degrades the **tiling** too, and that failure is silent.
`regions_of` pads to the end of file with a `TRAILER` (`map_symbols.py:209`), so a short AST still
produces regions that tile, `check_structure` still passes, and byte-for-byte reconstruction still
holds — of a region map that does not describe the code. So refusing is right for all of
`split_file`'s checks, not only the new one, and the message at `:512-515` already says "its region
and reference data", which is the accurate scope. Both maps I generated at this head carry
`parse_errors: 0`, and the survey finds clean parses for every file that has a compile command, so
the refusal costs nothing today. No `--force`, which is also right; if one is ever needed it should
be named for what it disables.

One ordering nit: the gate is at `:512`, after `split()` at `:497` has already printed
`[ok] partition: ...`. The user gets two greens and then a `[FAIL]` about the map being unusable.
It reads better above `split()`.

**Inactive `#if`: I agree, file it, and I would take one line of it here.**

The real fix means parsing both arms as separate TUs with different flags — a project, and not this
PR's. But the success line at `:540` states an unqualified universal,
`no internal-linkage definition is referenced from a part it is not in`, and the check cannot see
references in an inactive arm at all: `hle_kernel_mem.cpp` is 52% unparsed, `hle_kernel.cpp` 30%.
Qualifying that one string — "among references the parse could see; arms behind an inactive `#if`
are not covered" — costs nothing and stops the next reader taking the green as stronger than it is.
The issue should carry the four measured percentages.

### Nits

- `split_file.py:191-192` re-spells `("preamble", "open", "close")` inline when `REPLICATED_ROLES`
  is a module constant at `:70` that `split()` and `verify_reconstruction` both use. Same values
  today; note `map_symbols.REPLICATED` is a *different* tuple (`("open", "close")`, no preamble),
  and `survey_sizes.py:361-366` already comments on the two tools disagreeing — so this is a
  replication policy defined in three places, one of them by copy.
- `replicated` at `:191` is a **list**, and `src in replicated` at `:205` runs inside the edge loop.
  Fine at 422 regions; `rdna2_to_spirv.cpp` has ~10^5 references. Make it a set.

### Verified, no finding

README: both corrections applied, and the new recipe is better than what I suggested — putting
`split_file --dry-run` *before* `promote_internal` lets the tool name the closure. The 42-of-155
figure is correct (I measured it independently). Recipe is off `/tmp`. `parse_errors` is recorded and
reads 0 on both maps. `anon_map` is a faithful copy of `promote_internal`'s. Scope unchanged,
`diff --check` clean, 0 `Claude-Session:` trailers, no absolute host path in any added line.

**REJECTED** on `f7114b8fd`. One finding, one small fix, and the arm you already wrote will keep
whichever version you land.

mattias800 (commented):
## Review round 6 (delta) — head `185e099e9`

**APPROVED.**

The regression is gone, fixed at the source, and I answered your three questions by measurement
rather than by agreeing with you. One new non-blocking observation at the end, for an issue rather
than for this PR.

### The fix works, and the mutations bind

Edges sourced from an `open`/`close` region on `hle_kernel.cpp`: **19 -> 0**, reproduced on my own
map. `gpu_capture.cpp` loses 4 edges, `hle_kernel.cpp` 37, and the removed destinations are exactly
two regions per file — `{<anonymous>, gpu}` and `{fbsd_errno, prosper}` — i.e. only the regions that
happened to own a namespace USR. Nothing else moved: **0 edges added** in either file.

Seven mutations, all read from `$?`, all exit **1** with the `[ok]` summary suppressed, and each new
arm has one that reddens it alone — including reverting the set to the three roles, which reddens
`a namespace open/close is NOT treated as a replicated reference source` and only that.
`ctest -R refactor_ --no-tests=error` -> **4/4**.

### 1. Does the `NAMESPACE` skip make a promotion closure incomplete?

**No, and the direction is the opposite of what you feared: it makes closures SMALLER by removing
requirements that were never real.**

I rebuilt `promote_internal`'s fixed-point closure on the old and new matrices for the same seed sets
— 60 random seeds per file, seeds of 1 to ~1/6 of the body regions:

| file | seeds whose closure differs | shape of every difference |
| --- | --- | --- |
| `gpu_capture.cpp` | **0 of 60** | — |
| `hle_kernel.cpp` | **51 of 60** | `old \ new = [74]`, `new \ old = ∅` |

Region 74 is `fbsd_errno`'s **forward declaration** (line 266, `is_definition: False`), and it is in
those old closures for exactly the reason you found: `owner` bound the anonymous namespace's USR to
region 74, so every `namespace {` brace was an edge into it and the closure dragged it along. In
**every** sampled case `new \ old` is empty — the new matrix never drops a requirement the old one
had; it only stops inventing one.

The measurement corroborates an argument that is the general part: every removed edge has
`ref.kind == NAMESPACE`, so its destination is `owner[<namespace USR>]` — whichever region happened
to hold the first cursor carrying that USR, which has nothing to do with the referencing code. A
namespace is replicated into every part and every generated header by construction, so it is never a
thing a promotion has to pull in. The destination attribution was never a dependency, which is why
removing it cannot lose one.

### 2. Is `ref.kind == NAMESPACE` reachable, and is it over-broad?

**Reachable, correct, and not over-broad — and the one dangerous shape I could construct does not
fire.** I parsed a fixture with a namespace alias, a using-directive, a qualified call and an
anonymous-namespace helper, and printed every reference with what the new condition does to it:

```
 line  cursor kind        referent kind      referent          dropped?
    2  NAMESPACE_ALIAS    NAMESPACE_ALIAS    fs
    2  NAMESPACE_REF      NAMESPACE          filesystem        DROP
    3  NAMESPACE          NAMESPACE          other             DROP
    4  NAMESPACE_REF      NAMESPACE          other             DROP
    4  USING_DIRECTIVE    USING_DIRECTIVE    <anon>
   11  NAMESPACE_REF      NAMESPACE_ALIAS    fs                          <-- KEPT
   12  CALL_EXPR          FUNCTION_DECL      internal_helper
   12  DECL_REF_EXPR      FUNCTION_DECL      internal_helper
   12  UNEXPOSED_EXPR     FUNCTION_DECL      internal_helper
   12  NAMESPACE_REF      NAMESPACE          outer             DROP
```

The case worth checking was the **namespace alias** — `namespace fs = std::filesystem;` is a real
body region that must accompany its users, and it exists in-tree at
`prosper/src/hle/fs/save_paths.cpp:18`. A use of `fs::path` produces `NAMESPACE_REF` with referent
kind **`NAMESPACE_ALIAS`**, not `NAMESPACE`, so its edge survives. `using namespace other;` likewise
keeps its `USING_DIRECTIVE -> USING_DIRECTIVE` edge; only the redundant `NAMESPACE_REF` beside it is
dropped. Every dropped row names a *namespace* (`std`, `filesystem`, `other`, `outer`, `<anon>`), and
nothing else.

Real dependencies are also heavily over-determined — `internal_helper` yields `CALL_EXPR`,
`DECL_REF_EXPR` and `UNEXPOSED_EXPR` edges — so even a mistake in this predicate would have to hit
all three to lose an edge.

Small note: the first clause is redundant. A `NAMESPACE` cursor's `.referenced` is its own canonical
declaration, so `ch.kind == NAMESPACE` implies `ref.kind == NAMESPACE` — visible in rows 3, 5 and 7
of the probe. Harmless, and arguably worth keeping as documentation of intent.

### 3. Keep the preamble branch?

**Keep it, but say in the comment that the class is unobserved rather than impossible** — that is the
difference between it and the open/close branch you just removed, and it is the whole reason one is
right to keep and the other was not.

Measured: **0 preamble-sourced edges** in either map. So it is not fixing anything seen. But the two
"empty by construction" arguments are not the same strength:

- `open`/`close` holds only text clang gave no top-level cursor, so real code cannot be there. That
  is a fact about the tiling and it is airtight — which is why counting those was a pure false
  positive.
- `preamble` is marked **by position**, not by kind, and `map_symbols.py:211-215` says why in its own
  words: *"a stray declaration above the first namespace still belongs in every part, whereas a rule
  keyed on INCLUSION_DIRECTIVE would quietly file it into one of them."* So a preamble region can be
  real code. What makes its outgoing edges near-impossible is a **language ordering** argument — a
  preamble is a strict prefix, so a reference from it into a body region is a forward reference,
  which needs a prior declaration that would itself be in the prefix and therefore also `preamble`,
  and the `role != "body"` filter skips it.

An ordering argument is weaker than a tiling argument, and it is the kind that stops holding when
something upstream changes. The branch is one predicate, has an arm, and has a mutation that reddens
that arm alone. Keep it — just do not let the comment imply it was observed.

### One new observation, for an issue rather than this PR

**A forward declaration and its definition are two regions with one USR, and `split_file` does not
keep them together.** `promote_internal` handles exactly this shape deliberately — its `by_usr`
sibling pull-in, whose comment says *"A forward declaration and its definition are ONE entity in two
regions, and they must travel together ... Four such pairs exist in rdna2_to_spirv.cpp."*
`unresolved_across_parts` has no equivalent: it treats `owner[usr]`'s single region as the home.

Measured: `hle_kernel.cpp` has one such pair for a named internal function — `fbsd_errno`, regions
`[74, 123]`, `is_definition [False, True]` — and here `owner` bound to **123, the definition**, which
is the safe end. But nothing makes that so: `owner.setdefault(usr, ...)` is fed by a LIFO DFS
(`pending.pop()`), so which end of a pair wins is walk order, not source order. Bound to the
declaration instead, a plan that keeps the declaration with its user and sends the definition to
another part would pass this check and fail at **link**, which is the exact failure the check exists
to prevent.

I have no demonstrated failure — on both files it bound to the definition — so this is not a blocker.
The map already carries `is_definition`, so the fix is one predicate: prefer the defining region as
the edge target, or flag any USR whose regions span parts. Worth filing next to #3520.

### Verified, no finding

- **#3520** is filed and open with the four measured shares and `area:infra`. The README qualifier is
  right, and `No false positives` -> `No false positives on that file` is the correct narrowing —
  that sentence was a universal about a single measurement.
- The parse gate moved above `split()` (`:509-520`), so a bad map is refused before anything prints
  a green. The comment now carries the tiling reason, which is the load-bearing one.
- `replicated` is a set; `REPLICATED_ROLES` is no longer shadowed by an inline copy.
- Scope `prosper/tools/refactor/*` only in this commit. `diff --check` clean, 0 `Claude-Session:`
  trailers, no absolute host path in any added line.

**APPROVED** on `185e099e9`.

Issue #3529, opened 2026-09-10
Unresolved DATA imports bind to a CODE stub: the guest reads prosper's machine code as data, and a write corrupts the import stub table


mattias800:
## Unresolved DATA imports are bound to a CODE stub, so the guest reads prosper's machine code as data — and a write corrupts the import stub table

`linker.cpp`'s pass 2 binds **every** unresolved import to a stub slot in the executable stub
aperture, with no distinction between a function and a data symbol:

```
prosper/src/loader/linker.cpp:143   // --- Pass 2: resolve every import. Cross-module export beats a stub slot. ---
prosper/src/loader/linker.cpp:165   img.import_addr[imp.sym_index] = out.stub_base + (uint64_t)slot * out.stub_size;
```

`Symbol` (`src/self/module.hpp:63`) carries `shndx`, `value` and `is_import` but **no ELF symbol
type**, so the linker has nothing to branch on. `STT_OBJECT` and `STT_FUNC` are indistinguishable by
the time pass 2 runs.

### Measured, NINJA GAIDEN 4 (PPSA25258), 2026-09-11

The title imports `__stack_chk_guard` — a *variable* — and reads its stack canary through it. A
hardware execute breakpoint on the guest's own load instruction:

```
PROSPER_HWBP=0x28c3978 PROSPER_HWBP_ALLTHREADS=1
[hwbp] #1 rip=eboot+0x28c3978 rax=0x600000000 [rax]=0x000000bf ... tid=2535326
```

`rax = 0x600000000` is `BOOT_STUB`, the base of the import-stub aperture, and `[rax] = 0x000000bf`
is the first dword of stub 0 — `BF 00 00 00`, i.e. the `mov edi, 0` that begins
`emit_unimpl`/`emit_swap_stub`. **The guest's stack canary is prosper's own machine code.**

The title then dies on its Wwise I/O thread:

```
[sigill] rip=0x4128c3b0a bytes: 0f 0b eb 02 ...
[prosper] WORKER-THREAD FAULT: sig=4 addr=0x4128c3b0a rip=0x4128c3b0a (eboot+0x28c3b0a)
[fault] thread='AK::IOThread' guest-fs=0x7fd1... on-guest-TCB=yes
```

`eboot+0x28c3b0a` is the `ud2` after `call eboot+0x3194750`, which resolves through the PLT to
`Ou3iL1abvng` = **`__stack_chk_fail`**. The saved canary on the faulting frame reads **0**
[...]

PR #3530, opened 2026-09-10
hle: sceKernelMprotect must round rather than refuse, plus the two missing APR/memory-pool NIDs (#3498, #3506)

### LINKED ISSUE #3506: sceKernelMemoryPoolDecommit is unregistered — NINJA GAIDEN 4 commits pool memory it can never release (follow-up to #3502)
## `sceKernelMemoryPoolDecommit` is the one member of the pool family still unregistered

`#3502` registered Expand, Reserve and Commit. **Decommit was deliberately left out**: at the time
the guest had never reached it, so there was no captured ABI, and that issue's whole argument is that
guessing an out-parameter layout is worse than leaving the stub.

NINJA GAIDEN 4 now reaches it — **because** of that fix. `libkernel::LXo1tpFqJGs` appears in a
`PPSA25258` run on the merged head and resolves against the PS5 3.20 reference to
`sceKernelMemoryPoolDecommit`. `grep -c sceKernelMemoryPoolDecommit prosper/src` is **0**.

## Why this is not simply "one more missing NID"

**A family with Commit and no Decommit is worse than a family with neither.** The title is currently
committing its arena in 64 KiB steps (`pool_commit addr=0x1000400000`, `…0x1000410000`, `…0x1000420000`,
… under `PROSPER_MEMLOG=1`) with no way to release any of it. Every decommit the guest believes
succeeded leaves the mapping in place, so its allocator's picture of free memory and prosper's
diverge monotonically. On a long run that is unbounded growth presented to the guest as success.

That is a *consequence of #3502*, not a pre-existing gap, which is the reason to close it promptly
rather than queue it.

## The risk profile is INVERTED relative to #3502, and that changes the approach

Decommit takes **no out-parameter**. The hazard that shaped #3502 — writing through a wrongly
positioned argument — does not apply. The opposite one does: **unmapping a range the guest did not
ask about**, which is silent, and corrupts whatever else lived there.

So the safety requirement here is not "validate the pointer before writing" but **"refuse anything
that is not demonstrably ours"**:

- Resolve the range against prosper's own tracked mappings (`g_maps`, and the `committed` flag) before
  touching it. A decommit that does not correspond to a tracked committed mapping should fail, loudly,
  rather than fall through to an unmap.
- Partial and overlapping ranges need a decision. `k_reserve_vrange`'s `#115` containment rule is the
  nearest precedent in the file.
- Decommit should leave the *reservation* intact — that is the whole point of the reserve/commit
  split, and releasing the VA would let a later Commit at the same address fail or, worse, succeed
  against a different mapping.

## Approach

1. **Probe first.** Register it as a log-only handler (return 0, write nothing — behaviour identical
   to the unregistered path) and capture `a0..a5` on `PPSA25258`. This is what turned #3500 from
   suspicion into confirmation, and it costs one build. The expected shape is
   `(void* addr, size_t len, int flags)`, but that is an expectation, not evidence.
2. Implement against prosper's tracked mappings, refusing anything unrecognised.
3. **Add arms to `tests/hle/test_false_success_nids.cpp`** alongside the ones #3502 added, and
   verify each is non-vacuous by mutation — a Decommit arm that passes against "unmap unconditionally"
   would be worse than none.

## Not yet established

That the missing Decommit is *causing* NINJA GAIDEN 4's current failure. The title reaches its frame
loop, runs 120 s without faulting, and presents **zero frames** (#3500); nothing connects that to
memory release. This issue is filed on the strength of the API-family argument above, not on a
measured link — and #3498 is this week's reminder of the difference.

mattias800:
## What this fixes

Three unrelated-looking title failures, one root class: prosper answering a value-returning contract
with the dispatcher's `return 0`, or refusing a call the console accepts.

### 1. `sceKernelMprotect` refused an unaligned range (#3498)

FINAL FANTASY TACTICS – The Ivalice Chronicles (`PPSA21783`) terminated itself about a second into
every boot with `sceKernelDebugRaiseExceptionOnReleaseMode(0xa0020001)` and a completely clean log
above it — no fault, no unimplemented NID, nothing to read.

It calls `sceKernelMprotect(eboot+0xf999e0, 0x1da20, 0xc3)` during allocator init. The base is not
16 KiB aligned and the length is not a multiple of it. The POSIX handler passed both to host
`mprotect(2)`, which rejects an unaligned base with `EINVAL`. The title reads that as fatal: `-1`
propagates up four frames, its framework constructor returns NULL, `main()` returns, and the CRT
raises. Confirmed at the guest's own test instruction, `eboot+0x9779`, `rax=0x80020016`.

Protection is page-granular everywhere prosper runs, so a sub-page request can only mean "the pages
containing this span" — which is how this file's own `sceKernelMtypeprotect` has always read it. That
the console rounds rather than refuses is evidenced by the title: it ships and runs on hardware, so
hardware accepted that call.

This is the **second** title to lose a whole init path to a refused mprotect. #2101 was The Messenger
losing its AGC register-offset table, and the bounded `REFUSED` log that fix added landed on the
Windows half only — which is why this one had to be found by disassembling the guest rather than by
reading a log. That log is mirrored to the POSIX half here.

### 2. `sceKernelAprSubmitCommandBufferAndGetId` was unregistered (#3498)

The third member of a submit family this file already had two of. The damage is the SUBMIT, not the
id: `return 0` skipped `ampr_cb_reset`, so the command buffer's cursor was never released and the
guest's append loop — which polls `GetSize - GetUsed` — had no room for its next command. That is
the spin `k_ampr_init`'s own comment records parking an IoStore thread.

### 3. `sceKernelMemoryPoolDecommit` was unregistered (#3506)

The fourth member of the pool family, deliberately left out of #3502 because no title had reached it.
NINJA GAIDEN 4 (`PPSA25258`) reaches it now, precisely *because* that fix let it get further.

Its risk profile is inverted relative to the rest of the family: no out-parameter, so #3502's hazard
does not apply, but unmapping a range the guest did not name does. Nothing is touched that prosper
did not itself record as committed, and the **reservation survives** the release.

The first version asked "is EVERY byte committed?" and refused otherwise. The title answered
immediately: it decommits whole 4 MiB spans of which only part is committed, which is ordinary
allocator behaviour. That arm produced **8,140,215 refusals against 37 releases** in one 240 s run as
the guest retried. Clipping to the committed sub-ranges and succeeding is correct and still safe.

## Measured

`tools/screenshot`, Linux/RADV STRIX_HALO, `SDL_VIDEODRIVER=offscreen PROSPER_RENDER=1`, default
launch, no other environment variables.

| title | before | after |
| --- | --- | --- |
| `PPSA21783` FF Tactics | self-terminates at ~1 s, every run, 0 frames | runs 420 s with no fault: links, mounts its `enhanced` build, AGC register defaults for SDK 13, video-out 1920x1080, three seeded scanout baselines, streams its own graphics textures, produces audio |
| `PPSA25258` NINJA GAIDEN 4 | host SIGSEGV ~9 log lines in, immediately after the unregistered NID | the guest's own `##checkpoint:start` / `Runtime as started`, AGC SDK 13, video-out 1920x1080, seeded baselines, AJM audio init |

**Neither rung moves.** Both titles still present zero frames. These are named causes removed, not
milestones — the ladder's rung 1 is real frames and there are none.

FF Tactics' next blocker is located: its main thread spins in `scePthreadYield` inside its framework
`Init`, waiting for `[obj+0x54] == 1` at `eboot+0x459ccf`. NINJA GAIDEN 4's next blocker is filed as
**#3529** — `linker.cpp` binds unresolved DATA imports to a CODE stub, so the title reads prosper's
machine code as its stack canary and dies in `__stack_chk_fail` on the Wwise I/O thread.

## New diagnostic: the guest fatal-exit handlers now name the caller

A title that terminates *itself* is the hardest boot failure to diagnose — no fault, no unimplemented
NID, no host stack worth reading. The two guest fatal-exit handlers now print the guest call site.

Two independent instruments, printed together because which one works is a property of how the TITLE
was compiled:

- an **rbp-chain walk** — precise, and empty on a frame-pointer-less build (FF Tactics recovers
  exactly one frame, the module entry point);
- a **stack scan** whose every candidate is confirmed by decoding a `call` instruction immediately
  before the return address, which is what makes the output readable instead of a wall of
  plausible-looking numbers. This is the one that identified `libc.prx+0x1d6` here.

Both live in `exec_image.hpp` with per-platform implementations sharing `host/fault/rbp_chain.hpp`,
so the two platforms cannot drift on "who called this?".

## Tests

- **New** `tests/hle/memory/test_mprotect_unaligned.cpp` (`mprotect_unaligned`). It asserts the
  **pages changed**, read back from `/proc/self/maps`, rather than that the call returned 0 —
  `return 0;` satisfies the latter. Verified non-vacuous by three mutations, each failing the arm it
  should: removing the normalization (2 arms), normalizing then answering blind (2 arms), and
  widening past the requested span (1 arm).
- `tests/hle/test_false_success_nids.cpp` gains four arms for the APR entry point. The one that
  matters asserts the command-buffer **cursor** was reset, read through the guest-visible accessor.
  Its first draft asserted that on a buffer whose cursor was already 0 and **passed** against a
  handler that invents an id and never submits; it now appends first, checks the append moved the
  cursor (its own positive control), and re-fails under that same mutation.
- `test_retrack_reservation`'s POSIX arm provoked a host failure with `base + 1`, chosen only because
  `mprotect(2)` rejected an unaligned base — its own comment said so. The property #387 F3 protects
  is unchanged and still asserted; it is now provoked with an address one page above the top of the
  user address space, which is page-aligned and can never be mapped, so the arm no longer depends on
  alignment at all.

```
ctest --no-tests=error -j4    ->  100% tests passed, 466 tests, exit 0
git diff --check              ->  clean
```

## Risk and deliberately unaffected

- Both platform halves of `sceKernelMprotect` now normalize and record the **normalized** span in the
  tracker rather than the caller's unaligned one. Windows previously relied on `VirtualProtect`'s own
  rounding while tracking an unaligned range, so this makes the two agree.
- `apr_submit_common` gains an optional out-parameter; both existing call sites pass `nullptr` by
  default and are unchanged in behaviour.
- Decommit refuses anything outside prosper's own committed mappings, so a wrong argument reading
  fails visibly rather than unmapping something. `a2..a5` are logged, not used — the arity is
  `CONFIDENCE: MED` and the log is what would settle it.
- No renderer, recompiler or GPU path is touched.

Refs #3498
Refs #3500
Fixes #3506

### DIFF
--- a/BLOG.md
+++ b/BLOG.md
@@ -19,6 +19,21 @@ from the tracker issues, and still gated, because it is a projection of state ra
 > title's current state — for that, read the tracker. Nothing is ever removed when a title moves on,
 > because the point of a blog is that it records *when* things happened.
 
+## 2026-09-11
+
+### Two titles that could not boot at all now boot, and neither needed a renderer change
+
+*FINAL FANTASY TACTICS – The Ivalice Chronicles* killed itself about a second into every launch. It
+was one call: `sceKernelMprotect` with an address that is not 16 KiB aligned, which prosper handed
+to the host verbatim and the host refused. The title reads that refusal as fatal and walks all the
+way back out of `main`. It now runs, streams its own textures and plays audio — no picture yet, and
+that is the next thing.
+
+*NINJA GAIDEN 4* went from a crash nine log lines in to printing its own `Runtime as started` and
+bringing up video-out at 1920x1080. It needed the last missing member of the PS5 memory-pool API, and
+then told us something worth knowing: it reads its stack canary through a variable prosper had bound
+to a block of prosper's own machine code.
+
 ## 2026-09-10
 
 ### BlazBlue Entropy Effect X was stuck on its publisher logos because one mount point did not exist
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -1700,6 +1700,10 @@ if(TARGET prosper_core)
   target_link_libraries(test_ampr_amm PRIVATE prosper_core)
   add_test(NAME ampr_amm COMMAND test_ampr_amm)
 
+  add_executable(test_mprotect_unaligned tests/hle/memory/test_mprotect_unaligned.cpp)
+  target_link_libraries(test_mprotect_unaligned PRIVATE prosper_core)
+  add_test(NAME mprotect_unaligned COMMAND test_mprotect_unaligned)
+
   # #2908: a refused Ampr map must hand its direct-memory claim back. POSIX-only because the
   # mapping primitives it exercises are — on Windows sceAmprCommandBufferSetBuffer is still the
   # do-nothing stub, which cannot leak and cannot demonstrate the lever either.
--- a/prosper/src/hle/kernel/hle_kernel_time.cpp
+++ b/prosper/src/hle/kernel/hle_kernel_time.cpp
@@ -4,6 +4,8 @@
 #include "hle/dispatch/nid.hpp"
 #include "hle/kernel/hle_kernel_time.hpp"
 #include "host/image/boot_program.hpp"   // #1659: shared guest-module labelling
+#include "host/image/exec_image.hpp"      // guest_frames_from_rbp / describe_code_address
+#include "host/fault/guest_caller.hpp"    // PROSPER_CAPTURE_RBP at a fatal handler's entry
 #include "host/platform/posix_shim.hpp"     // PROSPER_ASM_TRAMPOLINE (pass entry %rsp as 7th arg)
 #include "host/platform/precise_sleep.hpp"   // guest sleeps must not inherit the winpthreads tick (#3013)
 #include "hle/kernel/timedwait_census.hpp"   // PROSPER_TIMEDWAIT_CENSUS: which primitive a title paces on
@@ -867,10 +869,53 @@ HLE(k_load_start_mod) {   // Windows/MinGW: no import-boundary %fs swap -> guest
 }
 #endif
 
+// Name the guest call site that reached one of the fatal exits below.
+//
+// A title that terminates ITSELF is the hardest kind of boot failure to diagnose, because there is
+// no fault, no unimplemented NID and no host stack worth reading -- the guest ran a check, disliked
+// the answer prosper gave it, and left. What the reader needs is the one thing the log did not have:
+// WHERE. FINAL FANTASY TACTICS (#3498) raises 0xa0020001 with a completely clean log above it, and
+// the only way to progress it is to disassemble the caller.
+//
+// Printed unconditionally rather than behind an env switch: this runs exactly once, on a path that
+// is already terminating the process, so there is no cost to weigh against it -- and a diagnostic
+// that must be enabled is one nobody has enabled on the run that mattered.
+//
+// The chain is a FRAME-POINTER walk, so say what it is. A caller compiled without a prologue is
+// skipped silently (instrument traps 114, 217), which makes a short chain and a shallow stack look
+// identical; confirm a link by disassembling the named site before building on it.
+void report_guest_fatal_caller(uint64_t entry_rbp, uint64_t entry_rsp) {
+    // Both instruments, always, because which one works is a property of how the TITLE was compiled
+    // and is not knowable here. The chain is precise and often empty; the scan is noisy and almost
+    // never empty. Printing only the chain is what left FINAL FANTASY TACTICS' raise reported as a
+    // single frame at the module entry point -- technically true, and useless.
+    uint64_t frames[8];
+    const int chained = guest_frames_from_rbp(entry_rbp, frames, 8);
+    for (int i = 0; i < chained; ++i)
+        fprintf(stderr, "[prosper] guest caller (rbp chain) #%d: 0x%llx (%s)\n", i,
+                (unsigned long long)frames[i], describe_code_address(frames[i]).c_str());
+    if (chained <= 0)
+        fprintf(stderr, "[prosper] guest caller (rbp chain): none -- this title omits frame pointers\n");
+
+    uint64_t scanned[12];
+    const int found = guest_frames_on_stack(entry_rsp, scanned, 12);
+    for (int i = 0; i < found; ++i)
+        fprintf(stderr, "[prosper] guest caller (stack scan) #%d: 0x%llx (%s)\n", i,
+                (unsigned long long)scanned[i], describe_code_address(scanned[i]).c_str());
+    if (found <= 0)
+        fprintf(stderr, "[prosper] guest caller (stack scan): none\n");
+    else
+        fprintf(stderr, "[prosper] guest caller: stack-scan entries are call-site-validated but "
+                        "unordered by liveness -- disassemble before building on one\n");
+}
+
 // _exit(status): terminate the process. Previously an unimplemented stub RETURNED 0, so libc's
 // exit path fell through into its deliberate ud2 (SIGILL) — terminate for real, loudly.
 HLE(k_exit) {
+    PROSPER_CAPTURE_RBP(entry_rbp);
+    PROSPER_CAPTURE_RSP(entry_rsp);
     fprintf(stderr, "[prosper] guest _exit(%d) -- terminating\n", (int)a0);
+    report_guest_fatal_caller(entry_rbp, entry_rsp);
     host::guest_dmem_write_trace_report();
     fflush(nullptr);
     _Exit((int)a0);
@@ -879,8 +924,14 @@ HLE(k_exit) {
 // it from failed check()s in shipping builds). Report and terminate rather than "return 0" and
 // let the guest run on in an undefined state.
 HLE(k_debug_raise_release) {
+    // Capture rbp FIRST, before this frame does anything that could overwrite it. Everything below
+    // is ordinary reporting; the one thing that must happen at the top is preserving the link back
+    // into the guest's stack (host/fault/guest_caller.hpp).
+    PROSPER_CAPTURE_RBP(entry_rbp);
+    PROSPER_CAPTURE_RSP(entry_rsp);
     fprintf(stderr, "[prosper] guest sceKernelDebugRaiseExceptionOnReleaseMode(code=0x%llx, arg=0x%llx) -- terminating\n",
             (unsigned long long)a0, (unsigned long long)a1);
+    report_guest_fatal_caller(entry_rbp, entry_rsp);
     host::guest_dmem_write_trace_report();
     fflush(nullptr);
     _Exit(0x66);
--- a/prosper/src/hle/memory/hle_kernel_mem.cpp
+++ b/prosper/src/hle/memory/hle_kernel_mem.cpp
@@ -1867,7 +1867,13 @@ HLE(k_pool_decommit) {
     // NINJA GAIDEN 4 loses nothing to this: its observed decommits are already 64 KiB-aligned with
     // 64 KiB-multiple lengths (`addr=0x1001010000 len=0x3f0000`, `addr=0x1000ee0000 len=0x10000`).
     constexpr uint64_t kCommitGranule = 0x10000ull;
-    if (a1 > UINT64_MAX - a0) return 0x80020016ull;
+    // Overflow first, and BOTH ends of it. The replaced code used normalize_guest_page_range, which
+    // checks its own round-up; rounding by hand dropped that, and the omission is not benign here --
+    // `sceKernelMemoryPoolDecommit(-16, 8, 0)` wraps `base` to 0 and yields a length of nearly 2^64,
+    // at which point committed_parts_in returns EVERY committed mapping and the loop below unmaps
+    // the whole process while answering SCE_OK. Raised in review of #3530.
+    if (a1 > UINT64_MAX - a0) return 0x80020016ull;                  // the span itself wraps
+    if (a0 > UINT64_MAX - (kCommitGranule - 1)) return 0x80020016ull; // rounding the base up wraps
     const uint64_t base = (a0 + kCommitGranule - 1) & ~(kCommitGranule - 1);
     const uint64_t end  = (a0 + a1) & ~(kCommitGranule - 1);
     if (end <= base) {
@@ -2701,13 +2707,16 @@ HLE(k_apr_submit_and_get_id) {
                 (unsigned long long)a4, (unsigned long long)a5);
     // A failed submit must not hand back an id the guest would then wait on.
     if (rc != 0) return rc;
-    // Nor may a SUCCESSFUL one return 0, which is precisely the value the missing handler returned
-    // and therefore the one a caller cannot distinguish from "unimplemented". The token normally
-    // comes from prosper_apr_next_token and is never 0, but a command buffer BOUND with a zero tag
-    // echoes that tag -- and this file documents a real title that binds with a literal `xor ecx,ecx`
-    // (CRI ADX2, in the apr_submit_common comment). Fall back to this ring's own counter there.
-    // Raised in review of #3530.
-    return token ? token : prosper_apr_next_token(a1 ? (unsigned)(a1 - 1) & 0x3f : 0);
+    // The token is whatever the family computed, INCLUDING 0. An earlier revision substituted this
+    // ring's own counter when the token came back 0, on the reasoning that 0 is the value the missing
+    // handler returned. That is reachable only for a cb BOUND with a zero tag -- and for a bound cb
+    // this file's own contract (see apr_submit_common) is that the token IS the binding tag, which
+    // the guest chose. Handing back the unbound counter dialect instead would answer a bound
+    // buffer in a dialect it never asked for, and would advance ring state as a side effect of a
+    // read. CRI ADX2 is the known zero-tag binder and its waiter tests the event IDENT, never the
+    // tag, so 0 is a legitimate payload there rather than an ambiguous one. Reverted in review of
+    // #3530; no title is known to bind with a zero tag AND use this entry point.
+    return token;
 }
 HLE(k_apr_submit) {   // sceKernelAprSubmitCommandBufferAndGetResult (cb, ring_1based, out1, out2)
     return apr_submit_common(a0, a1, a2, a3, /*write_result_outputs=*/true);
@@ -6970,7 +6979,13 @@ HLE(k_pool_decommit) {
     // NINJA GAIDEN 4 loses nothing to this: its observed decommits are already 64 KiB-aligned with
     // 64 KiB-multiple lengths (`addr=0x1001010000 len=0x3f0000`, `addr=0x1000ee0000 len=0x10000`).
     constexpr uint64_t kCommitGranule = 0x10000ull;
-    if (a1 > UINT64_MAX - a0) return 0x80020016ull;
+    // Overflow first, and BOTH ends of it. The replaced code used normalize_guest_page_range, which
+    // checks its own round-up; rounding by hand dropped that, and the omission is not benign here --
+    // `sceKernelMemoryPoolDecommit(-16, 8, 0)` wraps `base` to 0 and yields a length of nearly 2^64,
+    // at which point committed_parts_in returns EVERY committed mapping and the loop below unmaps
+    // the whole process while answering SCE_OK. Raised in review of #3530.
+    if (a1 > UINT64_MAX - a0) return 0x80020016ull;                  // the span itself wraps
+    if (a0 > UINT64_MAX - (kCommitGranule - 1)) return 0x80020016ull; // rounding the base up wraps
     const uint64_t base = (a0 + kCommitGranule - 1) & ~(kCommitGranule - 1);
     const uint64_t end  = (a0 + a1) & ~(kCommitGranule - 1);
     if (end <= base) {
@@ -6987,7 +7002,13 @@ HLE(k_pool_decommit) {
     const uint64_t len = end - base;
     uint64_t released = 0;
     for (const auto& part : committed_parts_in(base, len)) {
-        // win_unmap restores the Windows PLACEHOLDER it cut the view out of, so the address space
+        // NOTE ON THE GRANULE, because the rationale above is the LINUX one and this half's is not the
+    // same number: Windows' lazy-commit path works in 16 KiB, not 64 KiB (exec_image_win.cpp). 64 KiB
+    // is kept here anyway because it is a SUPERSET -- a 64 KiB-aligned inward-rounded span is also
+    // 16 KiB-aligned, so the hazard is covered either way, and one granule keeps the two halves
+    // answering the same question identically. The cost is that Windows releases slightly less than
+    // it could for a 16 KiB-granular caller. Raised in review of #3530.
+    // win_unmap restores the Windows PLACEHOLDER it cut the view out of, so the address space
         // stays claimed here without the explicit re-reservation the POSIX arm needs. Recorded
         // because the two halves therefore look different while implementing the same contract.
         if (!win_unmap(part.first, part.second)) return 0x80020016ull;
@@ -7629,13 +7650,16 @@ HLE(k_ampr_submit_and_get_id) {
                 (unsigned long long)a4, (unsigned long long)a5);
     // A failed submit must not hand back an id the guest would then wait on.
     if (rc != 0) return rc;
-    // Nor may a SUCCESSFUL one return 0, which is precisely the value the missing handler returned
-    // and therefore the one a caller cannot distinguish from "unimplemented". The token normally
-    // comes from prosper_apr_next_token and is never 0, but a command buffer BOUND with a zero tag
-    // echoes that tag -- and this file documents a real title that binds with a literal `xor ecx,ecx`
-    // (CRI ADX2, in the apr_submit_common comment). Fall back to this ring's own counter there.
-    // Raised in review of #3530.
-    return token ? token : prosper_apr_next_token(a1 ? (unsigned)(a1 - 1) & 0x3f : 0);
+    // The token is whatever the family computed, INCLUDING 0. An earlier revision substituted this
+    // ring's own counter when the token came back 0, on the reasoning that 0 is the value the missing
+    // handler returned. That is reachable only for a cb BOUND with a zero tag -- and for a bound cb
+    // this file's own contract (see apr_submit_common) is that the token IS the binding tag, which
+    // the guest chose. Handing back the unbound counter dialect instead would answer a bound
+    // buffer in a dialect it never asked for, and would advance ring state as a side effect of a
+    // read. CRI ADX2 is the known zero-tag binder and its waiter tests the event IDENT, never the
+    // tag, so 0 is a legitimate payload there rather than an ambiguous one. Reverted in review of
+    // #3530; no title is known to bind with a zero tag AND use this entry point.
+    return token;
 }
 HLE(k_ampr_submit) {                 // ASoW5WE-UPo: …AndGetResult — writes the result slots
     return apr_submit_common(a0, a1, a2, a3, /*write_result_outputs=*/true);
--- a/prosper/src/hle/memory/hle_kernel_mem.cpp
+++ b/prosper/src/hle/memory/hle_kernel_mem.cpp
@@ -2220,11 +2220,47 @@ static uint64_t sce_mprotect_error(int error) {
         default:     return 0x80020016ull;
     }
 }
+// sceKernelMprotect(addr, len, prot).
+//
+// The range is NORMALIZED to guest pages before anything is applied, exactly as k_mtypeprotect
+// below already does. Protection is a page-granular property on every platform prosper targets, so
+// a sub-page request can only ever mean "the pages containing this span" -- and the two functions
+// answering the same question differently is the divergence the charter's one-derivation rule
+// exists to prevent.
+//
+// This is not a tolerance, it is the console's contract, and a shipping title is the evidence.
+// FINAL FANTASY TACTICS - The Ivalice Chronicles (#3498) calls
+// sceKernelMprotect(eboot+0xf999e0, 0x1da20, 0xc3) during its allocator init -- an address that is
+// not 16 KiB aligned and a length that is not a multiple of it. Passed through verbatim, host
+// mprotect(2) rejects the unaligned base with EINVAL, and the title reads that as fatal: it
+// abandons the init, returns -1 up four frames, its framework constructor yields NULL, main()
+// returns, and the CRT raises sceKernelDebugRaiseExceptionOnReleaseMode(0xa0020001) about a second
+// into the boot. A title that ships and runs on hardware is proof the console accepted the call.
+//
+// It is also the SECOND time a refused mprotect has cost a title its whole init path: #2101 is The
+// Messenger losing its AGC register-offset table the same way. That fix, and the bounded REFUSED
+// log it added, landed on the Windows half only -- which is why this one had to be found by
+// disassembling the guest instead of by reading a log. The log is mirrored here for that reason.
+//
+// CONFIDENCE: HIGH. The failing call is live-captured, the EINVAL is observed at the guest's own
+// test instruction (eboot+0x9779), and the normalization matches the sibling entry point.
 HLE(k_mprotect) {
     if (!a0) return 0x80020016ull;
+    uint64_t base = 0, len = 0;
+    if (!normalize_guest_page_range(a0, a1, base, len)) return 0x80020016ull;
+    if (!len) return 0;                       // an empty span is a no-op, not an error
     const int prot = host_prot(a2);
-    if (mprotect((void*)a0, a1, prot) != 0) return sce_mprotect_error(errno);
-    retrack_prot(a0, a1, prot, static_cast<uint32_t>(a2), "mprotect");
+    if (mprotect((void*)base, len, prot) != 0) {
+        const int failure = errno;
+        static std::atomic<int> n{0};
+        if (n.fetch_add(1) < 24)
+            fprintf(stderr, "[memhle] sceKernelMprotect REFUSED addr=0x%llx len=0x%llx prot=0x%llx "
+                            "(pages [0x%llx,0x%llx)) errno=%d\n",
+                    (unsigned long long)a0, (unsigned long long)a1, (unsigned long long)a2,
+                    (unsigned long long)base, (unsigned long long)(base + len), failure);
+        return sce_mprotect_error(failure);
+    }
+    retrack_prot(base, len, prot, static_cast<uint32_t>(a2), "mprotect");
     return 0;
 }
 
@@ -6818,11 +6854,38 @@ static uint64_t sce_win_mprotect_error(DWORD error) {
         default:                      return 0x80020016ull;
     }
 }
+// sceKernelMprotect(addr, len, prot).
+//
+// The range is NORMALIZED to guest pages before anything is applied, exactly as k_mtypeprotect
+// below already does. Protection is a page-granular property on every platform prosper targets, so
+// a sub-page request can only ever mean "the pages containing this span" -- and the two functions
+// answering the same question differently is the divergence the charter's one-derivation rule
+// exists to prevent.
+//
+// This is not a tolerance, it is the console's contract, and a shipping title is the evidence.
+// FINAL FANTASY TACTICS - The Ivalice Chronicles (#3498) calls
+// sceKernelMprotect(eboot+0xf999e0, 0x1da20, 0xc3) during its allocator init -- an address that is
+// not 16 KiB aligned and a length that is not a multiple of it. Passed through verbatim, host
+// mprotect(2) rejects the unaligned base with EINVAL, and the title reads that as fatal: it
+// abandons the init, returns -1 up four frames, its framework constructor yields NULL, main()
+// returns, and the CRT raises sceKernelDebugRaiseExceptionOnReleaseMode(0xa0020001) about a second
+// into the boot. A title that ships and runs on hardware is proof the console accepted the call.
+//
+// It is also the SECOND time a refused mprotect has cost a title its whole init path: #2101 is The
+// Messenger losing its AGC register-offset table the same way. That fix, and the bounded REFUSED
+// log it added, landed on the Windows half only -- which is why this one had to be found by
+// disassembling the guest instead of by reading a log. The log is mirrored here for that reason.
+//
+// CONFIDENCE: HIGH. The failing call is live-captured, the EINVAL is observed at the guest's own
+// test instruction (eboot+0x9779), and the normalization matches the sibling entry point.
 HLE(k_mprotect) {
     if (!a0) return 0x80020016ull;
+    uint64_t base = 0, len = 0;
+    if (!normalize_guest_page_range(a0, a1, base, len)) return 0x80020016ull;
+    if (!len) return 0;
     const int prot = host_prot(a2);
     DWORD error = ERROR_SUCCESS;
-    if (!win_protect(a0, a1, prot, &error)) {
+    if (!win_protect(base, len, prot, &error)) {
         // #2101: a refused mprotect is silent, and a guest that treats it as fatal then abandons a
         // whole init path with no other symptom. Report the span, the requested Sony protection and
         // the Win32 reason, bounded.
@@ -6834,7 +6897,7 @@ HLE(k_mprotect) {
                     (unsigned long)error);
         return sce_win_mprotect_error(error);
     }
-    retrack_prot(a0, a1, prot, static_cast<uint32_t>(a2), "mprotect");
+    retrack_prot(base, len, prot, static_cast<uint32_t>(a2), "mprotect");
     return 0;
 }
 HLE(k_mtypeprotect) {
--- a/prosper/src/hle/memory/hle_kernel_mem.cpp
+++ b/prosper/src/hle/memory/hle_kernel_mem.cpp
@@ -6979,6 +6979,13 @@ HLE(k_pool_decommit) {
     // NINJA GAIDEN 4 loses nothing to this: its observed decommits are already 64 KiB-aligned with
     // 64 KiB-multiple lengths (`addr=0x1001010000 len=0x3f0000`, `addr=0x1000ee0000 len=0x10000`).
     constexpr uint64_t kCommitGranule = 0x10000ull;
+    // THE GRANULE RATIONALE ABOVE IS THE LINUX ONE, and this half's number is not the same: Windows'
+    // lazy-commit path works in 16 KiB, not 64 KiB (exec_image_win.cpp). 64 KiB is kept here anyway
+    // because it is a SUPERSET -- a 64 KiB-aligned inward-rounded span is also 16 KiB-aligned, so the
+    // hazard is covered either way, and one granule keeps the two halves answering the same question
+    // identically. The cost is that Windows releases slightly less than a 16 KiB-granular caller
+    // asks for. Stated here, before the text it corrects, because a reader meets that text first.
+    // Raised in review of #3530.
     // Overflow first, and BOTH ends of it. The replaced code used normalize_guest_page_range, which
     // checks its own round-up; rounding by hand dropped that, and the omission is not benign here --
     // `sceKernelMemoryPoolDecommit(-16, 8, 0)` wraps `base` to 0 and yields a length of nearly 2^64,
@@ -7002,12 +7009,6 @@ HLE(k_pool_decommit) {
     const uint64_t len = end - base;
     uint64_t released = 0;
     for (const auto& part : committed_parts_in(base, len)) {
-        // NOTE ON THE GRANULE, because the rationale above is the LINUX one and this half's is not the
-    // same number: Windows' lazy-commit path works in 16 KiB, not 64 KiB (exec_image_win.cpp). 64 KiB
-    // is kept here anyway because it is a SUPERSET -- a 64 KiB-aligned inward-rounded span is also
-    // 16 KiB-aligned, so the hazard is covered either way, and one granule keeps the two halves
-    // answering the same question identically. The cost is that Windows releases slightly less than
-    // it could for a 16 KiB-granular caller. Raised in review of #3530.
     // win_unmap restores the Windows PLACEHOLDER it cut the view out of, so the address space
         // stays claimed here without the explicit re-reservation the POSIX arm needs. Recorded
         // because the two halves therefore look different while implementing the same contract.
--- a/prosper/src/hle/memory/hle_kernel_mem.cpp
+++ b/prosper/src/hle/memory/hle_kernel_mem.cpp
@@ -1832,7 +1832,7 @@ std::vector<std::pair<uint64_t, uint64_t>> committed_parts_in(uint64_t base, uin
 // code. Decommit has no out-parameter, so the hazard that made #3502 probe before implementing --
 // writing through a wrongly positioned argument -- does not apply. The opposite one does: unmapping
 // a range the guest did not name, which is silent and corrupts whatever else lived there. Hence
-// tracked_committed_covers above: nothing is touched unless prosper itself recorded it as a
+// committed_parts_in above: nothing is touched unless prosper itself recorded it as a
 // committed mapping.
 //
 // THE RESERVATION SURVIVES. That is the whole point of the reserve/commit split -- releasing the VA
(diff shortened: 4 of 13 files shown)


mattias800 (commented):
## Independent review — PR #3530 @ `10c85ebb85d6671e3fb03dd4bc58a3f108609284`

Read the diff, traced every `file:line` and every issue the PR leans on, and reasoned about the
untested paths. I did not re-run the author's builds/tests/CI.

### Citations I traced, and what they say

- **`sceKernelMtypeprotect` as precedent for rounding — TRUE, and pre-existing.** `git show
  <base>:…/hle_kernel_mem.cpp` has `normalize_guest_page_range` in *both* halves of `k_mtypeprotect`
  (POSIX 2238-2248, Windows 6840-6851) before this PR. "as `k_mtypeprotect` below already does" is
  also literally true of the placement: `k_mprotect` is above it on both platforms.
- **`k_reserve_vrange` installs `track(base, len, 0, 0, false, "reserved")` — TRUE**
  (`hle_kernel_mem.cpp:1610`, `:1633`, `:1655`). The decommit replacement matches it exactly,
  including `query_flags = 0`.
- **#2101 as "a refused mprotect cost The Messenger its AGC register-offset table" — SUPPORTED,
  though not from the issue body.** #2101's body is about Cx register writes never landing on
  Windows; the mprotect root cause is recorded in this file's own `win_protect` comment
  (`hle_kernel_mem.cpp:6073`) with the exact call `sceKernelMprotect(eboot+0x1d98000, 0x22240,
  …)`. Worth noting for the record that that refusal was an **ownership** refusal, not an alignment
  one — the base there is 16 KiB aligned. The PR's wording ("a refused mprotect") is accurate; a
  reader skimming it could take away "an unaligned mprotect", which would be wrong.
- **Both existing `apr_submit_common` call sites unchanged — TRUE.** Four sites, two per platform
  (`:2656`, `:2671`, `:7527`, `:7535`), all still pass the same five arguments; the new parameter is
  defaulted. `apr_submit_common` has exactly one `return 0`, so `rc == 0 ? token : rc` in the new
  handler is dead but harmless.
- **`win_unmap` keeps the address space claimed — TRUE on all four of its paths.** Whole
  placeholder view → `UnmapViewOfFile2(MEM_PRESERVE_PLACEHOLDER)` + `remember_free_placeholder_locked`
  (`:6506-6509`); whole private placeholder → `VirtualFree(MEM_RELEASE|MEM_PRESERVE_PLACEHOLDER)`
  (`:6558-6561`); partial private → `MEM_DECOMMIT` (`:6562-6564`); and the legacy `!special_overlap`
  path → `VirtualFree(addr, len, MEM_DECOMMIT)` (`:6624-6627`), which also retains the reservation.
  The comment checks out against the code, not just against itself.
- **One claim that does NOT hold: `git diff --check -> clean`.** `git diff --check <base>..<head>`
  reports `prosper/src/host/image/exec_image_win.cpp:1389: trailing whitespace.` — the blank line
  inside the new `guest_frames_on_stack` is four spaces (`cat -A` shows `    $`).

### Supporting evidence the PR does not cite, and probably should

The strongest argument for rounding is not "the title ships". FreeBSD's `kern_mprotect` does exactly
this — `pageoff = addr & PAGE_MASK; addr -= pageoff; size += pageoff; size = round_page(size)` — and
the PS5 kernel is FreeBSD-derived. That turns "a shipping title implies hardware accepted it" (which
establishes only that hardware did not *fail* the call) into a positive statement about *which*
rounding hardware performs, including the direction. It also settles the len==0 corner this PR
introduces: `sceKernelMprotect(unaligned, 0, prot)` now protects one whole page where an aligned
address with len 0 is still a no-op, which looks inconsistent inside prosper but is precisely what
FreeBSD does. Please put this in the comment; it is worth more than the shipping-title argument and
it is checkable.

---

## Blocking

### B1. The POSIX arm of `test_retrack_reservation` no longer discriminates half of #387 F3

This is the finding I would most want re-checked, because it is an author changing a test their own
change broke.

The property the arm exists for is stated at `tests/hle/test_retrack_reservation.cpp:170-172`: *"a
host protection failure must reach the guest **and must not retag the tracked reservation**. The old
handler returned success and split/renamed the mapping even though no host protection changed."*
Two halves.

Old POSIX arm: `invalid_start = base + 1` — **inside** the reservation `[base, base+3*page)`. A
mutation that calls `retrack_prot` regardless of the host result splits the tracked reservation, and
the follow-up `query(base + len - page/4, …)` at `:191-196` then reports a region whose start is not
`base`. That query is the discriminator, and it worked because the failing range overlapped the
reservation.

New POSIX arm: `invalid_start = 0x0000800000000000` (`:185`) — nowhere near `base`. Under the same
mutation, `retrack_prot(0x800000000000, page, …)` finds no overlap, takes the
`retagged.empty()` branch (`hle_kernel_mem.cpp:1199-1210`) and inserts a *separate* tracked mapping
at 0x800000000000. `[base, base+len)` is untouched, `insert_mapping_by_base` keeps ordering, and the
query at `:191` still passes. **On POSIX the retag half of #387 F3 is now vacuous** — only the
"failure reaches the guest" half survives. I checked the rest of the file: no other POSIX arm
provokes a host protection failure over a tracked range (`:332` is the success path; the
boundary-crossing arms at `:234-249` are `#ifdef _WIN32`). So the coverage is Windows-only now.

The PR body says "The property #387 F3 protects is unchanged and still asserted." That is not
correct for the POSIX half.

Suggested remedy that does not depend on alignment and keeps the overlap: punch a host hole inside
the tracked reservation from the test itself, so the tracker still records one mapping while the
host cannot protect the span —

```c
munmap((void*)(uintptr_t)(base + page), (size_t)page);   // host hole; tracker still says [base, base+len)
const uint32_t invalid_rc = (uint32_t)protect(base, len, 0x2, 0, 0, 0);   // ENOMEM
CHECK(invalid_rc == 0x8002000cu || invalid_rc == 0x80020016u, …);
// …then the existing query at base + len - page/4 is a discriminator again.
```

(Keep the out-of-range arm too if you like it — it is a fine second arm, it just is not this one.)

### B2. `Fixes #3506` lands `k_pool_decommit` with no test at all

`grep -rn 'Decommit\|LXo1tpFqJGs\|pool_decommit' prosper/tests` → nothing. #3506's own "Approach"
section asks for the opposite in as many words: *"Add arms to `tests/hle/test_false_success_nids.cpp`
alongside the ones #3502 added, and verify each is non-vacuous by mutation — a Decommit arm that
passes against 'unmap unconditionally' would be worse than none."*

This is the one handler in the PR that **destroys guest memory**, it is the one whose argument layout
is `CONFIDENCE: MED`, and it is the only new entry point with zero assertions. The neighbouring arms
in that same file are unusually good — `test_kernel_memory_pool` names the killing mutation for every
check, and the null-slot loop at `:400-408` even carries the scar of a previously-vacuous arm — which
makes the omission stand out rather than blend in.

Three behaviours a test would pin, all of which I can only reason about today:

- clipping: a decommit spanning committed and uncommitted sub-ranges releases only the committed
  parts and returns 0 (the measured NINJA GAIDEN 4 shape);
- the reservation survives: `prosper_reserved_range_state(addr)` is 1 after decommit, not 0, and a
  subsequent `sceKernelMemoryPoolCommit` at the same address succeeds at the same address;
- refusal: a range prosper never committed touches nothing — and see N3 on what "refusal" currently
  means here.

### B3. `call_precedes` dereferences bytes that no probe covers — on both platforms

`exec_image_linux.cpp:3741` / `exec_image_win.cpp:1371`:

```c
if (ret < 16 || !readable(ret - 16)) return false;
const uint8_t* p = (const uint8_t*)(uintptr_t)ret;
if (p[-5] == 0xE8) return true;
```

`probe_readable(a)` verifies `[a, a+8)` — this file says so itself at `exec_image_linux.cpp:694-695`:
*"probe_readable(X) verifies [X, X+8) — so guard each read with its BASE address (not base+size-1,
which would leave the leading bytes of an unaligned read unverified and could fault in-handler)."*

So the probe covers `[ret-16, ret-8)`. The bytes actually read are `p[-7] … p[-1]`, i.e.
`[ret-7, ret)`. **The probed window and the read window do not overlap at all.** When a page boundary
falls in `(ret-9, ret-1]` — `ret % 4096 ∈ [1,8]` — the probe validates the page *below* the boundary
and the reads land on the page above it, which may be unmapped. That configuration is exactly "a
stale stack word pointing a few bytes past the end of a mapped segment", and the module filter above
does not exclude it: `guest_module_name` (`boot_program.hpp:79`) classifies by **fixed apertures**,
so it returns e.g. `"eboot"` for any address in the aperture whether or not that page is mapped.

Windows is wider: `addr_readable` (`exec_image_win.cpp:353`) is page-granular VirtualQuery, so the
gap is `ret % 4096 ∈ [1,15]`.

Low probability per candidate, but the scan runs up to 8192 candidates, it runs on a path that is
already terminating the process (so a nested SIGSEGV would replace the diagnostic with a confusing
fault report), and the fix is one line per platform:

```c
if (ret < 16 || !readable(ret - 16) || !readable(ret - 8) || !readable(ret - 1)) return false;
```

The rest of the diagnostic is clean: `walk_rbp_chain` probes `bp` and `bp+8` and reads exactly those
two words; the stack loop probes each 8-aligned slot and `break`s rather than skipping on the first
unreadable one; `limit` only ever narrows.

---

## Non-blocking

**N1. `k_pool_decommit` rounds a destructive range out, where this file's own destructive sibling
refuses.** `k_munmap` rejects an unaligned base or length outright on both platforms
(`hle_kernel_mem.cpp:2291-2294`, `:7046-7049`). `k_pool_decommit` (`:1849`, `:6895`) instead calls
`normalize_guest_page_range`, so a sub-page request unmaps up to 16 KiB the guest never named — same
rounding as the protection path, but the consequence is deletion rather than a permission change. The
PR justifies rounding by citing `k_mtypeprotect`; for decommit the nearer precedent is `k_munmap`,
and it points the other way. FreeBSD's `kern_munmap` does round out, so this is defensible — it just
is not argued anywhere, and it disagrees with the neighbouring function.

**N2. `committed_parts_in` clips to *any* committed mapping, not to pool mappings.** `:1808` /
`:6854` filter on `m.committed` alone, so a decommit call with a wrong pointer can unmap a committed
*flexible* or *direct* mapping. That matches `k_munmap`'s reach, so I do not think it is wrong — but
the comment's "nothing is touched unless prosper itself recorded it as a committed mapping" is a
weaker guarantee than "nothing outside the pool", and the code has a `name` field it could use.

**N3. A decommit that releases nothing returns SCE_OK and is silent by default.** With
`committed_parts_in` empty, `k_pool_decommit` returns 0 having done nothing, and the only trace is
`MLOG`, which is `if (memlog())` (`:686`) — off unless `PROSPER_MEMLOG` is set. The PR body says "a
wrong argument reading fails visibly rather than unmapping something"; on a default run it does not
fail and is not visible. Consider `released == 0 && !parts.empty()`-style reporting, or an
unconditional bounded `fprintf` the way the new mprotect REFUSED log does.

**N4. Decommitted pool pages become lazily re-committable, and the lazy-commit handler works in
64 KiB.** `track(part, …, committed=false, "reserved")` puts the range at
`prosper_reserved_range_state == 1`, and `exec_image_linux.cpp:2181` lazily backs any faulting
address `>= 0x1000000000` in that state with `mmap(page & ~0xffff, 0x10000, RW, MAP_FIXED|ANONYMOUS)`.
NINJA GAIDEN 4's pool is at exactly `0x1000000000`. Two consequences the PR does not mention: a
use-after-decommit is silently papered over with zeroed memory instead of faulting, and — the one
that could corrupt — a decommitted hole **smaller than or not aligned to 64 KiB** makes that
`MAP_FIXED` mmap overwrite the still-committed neighbours inside the same 64 KiB page. The guest's
own commits are 64 KiB (per #3506), so today's shape is probably safe; the 16 KiB normalization is
what makes a smaller hole constructible. I have not reproduced this — it is a mechanism, not a
measurement — but it is the sharpest reason B2's test matters.

**N5. Normalizing can turn a previously-succeeding mprotect into a failure.** Rounding the end up to
16 KiB can run past the end of a host mapping that is not a multiple of 16 KiB (`k_map_flexible`
does not enforce 16 KiB granularity: `map_at(hint, a1, …)` at `:1666`). POSIX then gets ENOMEM;
Windows gets `win_protect`'s untracked-gap refusal at `:6105-6107`. `k_mtypeprotect` has carried this
for a long time without incident, and `k_mprotect` is called far more often. A cheap belt-and-braces
would be to fall back to the host-page-normalized range when the guest-page one fails with
ENOMEM/ERROR_INVALID_ADDRESS — the EINVAL that #3498 is about only needs host-page alignment.

**N6. POSIX `k_mprotect` still has no ownership guard, and the widening extends its reach.** Windows
refuses spans crossing into memory the guest does not own, including the import-stub aperture
(`guest_module_image_span`, `:6022-6027`, called at `:6105`, #2144). POSIX calls host `mprotect` directly with no such
check, so a guest can already strip execute from prosper's own trampolines; rounding down/up extends
that reach by up to 16 KiB in each direction. Pre-existing, not introduced here — worth an issue.

**N7. `guest_frames_on_stack` and `call_precedes` are copy-pasted into both platform files.** The PR
says "Both live in `exec_image.hpp` with per-platform implementations sharing
`host/fault/rbp_chain.hpp`, so the two platforms cannot drift on 'who called this?'" — true of the
rbp walk, not of the scanner, which is ~45 duplicated lines. `rbp_chain.hpp`'s own header
(`:6-9`) argues against exactly this: *"two hand-written copies of a loop that dereferences
guest-controlled pointers is exactly the kind of duplication that grows a divergence nobody notices
until one of them faults."* B3 already has to be fixed twice because of it. The scanner is
parameterised on `readable` in both copies, so lifting it into `rbp_chain.hpp` beside
`walk_rbp_chain` is mechanical.

**N8. `PROSPER_CAPTURE_RBP`'s stated invariant is not guaranteed by the compiler.** The header says
"without [a frame pointer], rbp still HOLDS the guest's, because rbp is callee-saved and no import
stub writes it." Callee-saved means the *caller's* value is restored by the time the callee returns,
not that the callee has not already reused rbp — with `-fomit-frame-pointer` GCC routinely moves an
argument into rbp in the prologue, before any statement of the body. It evidently works on the
author's build (FF Tactics recovers the module entry point, which is a guest value), and the failure
mode is a junk chain rather than a fault, so this is a caveat rather than a defect. Worth softening
the header's claim to "in practice, on the builds we ship" so nobody later relies on it as an
invariant. Related: `PROSPER_ASM_TRAMPOLINE` already exists and passes the entry `%rsp` as a 7th
argument — it is used in this very file (`hle_kernel_time.cpp:863`) — and would give an exact seed
rather than a post-prologue one.

**N9. `AndGetId` can still return 0, and the new test does not cover that path.**
`apr_submit_common:613` computes `token = tag_echo ? bc.tag : prosper_apr_next_token(ring)`, and the
comment block above it documents a real title (CRI ADX2 in Tales of Graces f) that binds with a
literal **zero tag**. For such a buffer the new handler returns 0 — the exact value the missing
handler returned, and the exact thing `test_apr_submit_and_get_id` asserts against on the unbound
path. Is the tag really the right *id* for an entry point whose contract is "return the id", or
should AndGetId always hand back `prosper_apr_next_token(ring)`? Either answer is fine; it should be
decided rather than inherited, and the bound case deserves an arm.

**N10. The `a2..a5` arity reading (`CONFIDENCE: MED`) — evidence is thin but the failure mode is
acceptable.** The offered evidence is the family shape plus the 3.20 database name, not a capture,
which is weaker than what #3502 held itself to for the pool family (live-captured layout). What makes
it tolerable is the direction of the error: if `a2/a3` *are* out-parameters, prosper writes nothing
rather than writing to a wrong address, and the guest reads uninitialised stack — bad, but strictly
better than the corruption #3502 was avoiding, and identical to what the already-shipping plain form
does. One caveat on "the log is what settles it": that log is behind `amprlog()`
(`PROSPER_AMPRLOG`), so it settles nothing on a default run. If FF Tactics is going to keep being
run, consider capturing `a2..a5` once now and upgrading the confidence, rather than leaving a MED
that nobody is scheduled to resolve.

**N11. Small things.** (a) The decommit comment says "Hence `tracked_committed_covers` above" in both
copies (`hle_kernel_mem.cpp:1835`, `:6881`); no such symbol exists — the function is
`committed_parts_in`. (b) `if (!len) return 0;` in `k_pool_decommit` is unreachable (`a1 != 0` was
already checked, so the normalized length is at least one page). (c)
`test_mprotect_unaligned.cpp:118` picks `0x7ffff0000000` as "reserved-looking, and not mapped by this
test" — that is inside Linux's usual mmap band; a `mmap`+`munmap` to obtain a genuinely free address
would remove a small flake risk. (d) The trailing whitespace at `exec_image_win.cpp:1389` noted above.

---

## Things I checked and found sound

- The new `test_mprotect_unaligned` is genuinely non-vacuous in the directions that matter: reading
  permissions back from `/proc/self/maps` defeats "answer blind"; asserting `[base, base+2*page)`
  writable defeats rounding the base *up*; asserting the page at `expected_hi + kGuestPage` is still
  read-only defeats widening to the mapping; and the neighbour page is inside the fixture's own
  8-page mmap, so the arm really can observe the mutation.
- `test_apr_submit_and_get_id`'s cursor arm is fixed for real. The `appended != 0` positive control
  at `:477-480` is what makes `get_offset(...) == 0` after the submit a discriminator, and it self-detects
  if the append ever stops moving the cursor. I checked the neighbouring arms in the same file for the
  same defect and found none — `test_kernel_memory_pool` poisons every out-parameter before the call
  and its null-slot loop nulls exactly one pointer per iteration precisely to avoid the
  "sits next to the gap" failure.
- Adding the POSIX `REFUSED` log is the right call and would have found #3498 by reading a log. The
  24-line bound and the fact that it prints both the requested and the normalized span are both right.
- The decommit POSIX arm's `MAP_FIXED` anonymous replacement is the correct primitive for keeping the
  VA claimed, and dropping a flexible mapping's backing loses nothing: `track()` at `:1127-1133`
  records flexible memory with a synthetic VA identity and the code retires the write-watch with
  `guest_write_watch_notify_direct_mapping_removed` first, exactly as `k_munmap` does.

---

**REJECTED** on B1, B2 and B3. None of the three is a design problem — B1 is a test arm that needs
its overlap back, B2 is a missing test the closed issue asks for by name, B3 is one line per
platform. The mprotect semantic change itself I believe is right, and I would strengthen its
argument with the FreeBSD `kern_mprotect` rounding rather than resting it on "the title ships".
Happy to re-review promptly on a new head.

mattias800:
Thank you — B1, B2 and B3 are all correct, and so is the `git diff --check` catch. I ran that on the
working tree (clean, because everything was committed) rather than on `origin/main..HEAD`, which is a
different question and the one the charter actually asks. Working through all three now; the FreeBSD
`kern_mprotect` citation is a straight improvement and is going into the comment in place of resting
the argument on the shipping title.

One correction to the record while I fix it: on **B1** you are right that my PR body's claim — "the
property #387 F3 protects is unchanged and still asserted" — is false for the POSIX half. That
sentence is exactly the error I am most prone to: describing a gap as covered because the new arm
sits *next to* the thing it used to catch. The fix keeps the overlap by punching a host `munmap` hole
inside the tracked reservation, as you suggested, so the failing range overlaps `[base, base+len)`
again and the follow-up query is a discriminator rather than a formality.

Two of the non-blocking findings are being taken as blocking-in-effect, because they are about the
one handler that destroys memory:

- **The lazy-commit interaction.** You are right that a decommitted hole goes back to being a tracked
  reservation and that the fault handler serves those with `mmap(page & ~0xffff, 0x10000, MAP_FIXED)`
  — a whole 64 KiB granule. A hole smaller than that, or misaligned to it, would be re-backed by an
  mmap that also replaces its still-live neighbours. Decommit now works in **64 KiB granules rounded
  INWARD**, which makes that unreachable by construction and simultaneously answers your point that
  rounding a *destructive* range outward is the wrong direction where `k_munmap` refuses outright.
  NINJA GAIDEN 4 loses nothing: its observed decommits are already 64 KiB-aligned with 64 KiB-multiple
  lengths.
- **Silence on a release that releases nothing.** Now a bounded stderr line, both for "spans no whole
  granule" and for "released less than asked", on the default path rather than behind
  `PROSPER_MEMLOG`.

Also taken: **N9** — `AndGetId` can no longer return 0 even for a command buffer bound with a zero
tag, since 0 is precisely the value the missing handler returned and therefore the one a caller
cannot distinguish from "unimplemented". **N11(a)** the stale `tracked_committed_covers` name in both
comment copies, and **N11(b)** the unreachable `if (!len)`.

And the copy-pasted scanner: you are right that the PR claimed the two platforms "cannot drift" while
the diff contained two copies of `call_precedes` and the scan loop. They now live in one
`host/fault/guest_stack_scan.hpp` parameterised by the platform's `readable` probe, which is what the
claim asserted. B3's probe fix is therefore in one place too.

I will ping you for a delta review on the new head.

mattias800 (commented):
## Independent review (delta) — PR #3530 @ `573f40dbbca44bed49325a7430d85dc7b30ff29e`

Reviewed only what changed since `10c85ebb8`. I traced every citation I make below against
`573f40dbb` itself, and I did not re-run the author's builds or ctest.

### The three blocking findings from the previous head

**B1 — FIXED, and I verified it is non-vacuous rather than taking the mutation result on trust.**
I traced `retrack_prot` (`prosper/src/hle/memory/hle_kernel_mem.cpp:1174-1196`) for the exact
request the arm now makes, `retrack_prot(base + 2*page, page, …)` against the tracked
`[base, base+3*page)`:

- `me (base+3p) > base_arg (base+2p)` and `m.base (base) < end (base+3p)`, so the mapping is **not**
  skipped by `:1176`;
- `m.base < base_arg`, so `:1177-1179` emits a `lo` of `[base, base+2p)`;
- `changed` is rebased to `base+2p` with `changed_end = min(me, end) = base+3p`, i.e. size `page`,
  renamed `"mprotect"`;
- `me > end` is false, so no `hi`.

g_maps therefore holds **two** mappings under the mutation. The follow-up query at
`test_retrack_reservation.cpp:213` reads `base + len - page/4`, which lands in the second one, and
`k_virtual_query` reports `start = mapping.base` / `end = mapping.base + mapping.size` from
`snapshot_mapping` with **no coalescing** (`hle_kernel_mem.cpp:2098`). So `== base` and
`== base + len` both fail under the mutation. The arm discriminates. The whole-range version really
was vacuous for the reason the new comment gives, and the strict sub-range really does fix it.

I also checked the provocation itself: after `munmap` at `:199` the page is fully unmapped, so
`mprotect` at `hle_kernel_mem.cpp:2391` returns ENOMEM and `k_mprotect` returns before
`retrack_prot` at `:2401`. No alignment dependence remains.

**The restore at `:226-229` does not mask anything.** It runs *after* the assertion it serves, it
re-creates exactly what `k_reserve_vrange` had made (`map_guest_auto(a1, PROT_NONE, align)`,
`hle_kernel_mem.cpp:1651`) with the same `PROT_NONE | MAP_PRIVATE | MAP_ANONYMOUS`, and the tracker
was never told about either the hole or the restore, so g_maps and the host agree again. The later
arms that need the page are real: `:366` protects `[base, base+3p)` and `:382` unmaps it. One
narrow caveat worth a line in the comment: `MAP_FIXED` is a blind overwrite, so if some future
defect caused the code under test to map something there, this would discard it silently. Today
nothing can.

**B3 — FIXED, and the coverage argument in the new header holds on both platforms.** The bytes read
are `p[-7] … p[-1]` (the `FF` loop's largest `len` is 7 and `c[i+1]` is gated by `i + 1 < len`, so
the highest index is `p[-1]`). `probe_readable` verifies `[a, a+8)` — confirmed at
`exec_image_linux.cpp:684-689` and stated at `:695`, the line I cited last time — so `readable(ret-8)`
alone already covers `[ret-8, ret)`. On Windows `addr_readable` is region-granular `VirtualQuery`
(`exec_image_win.cpp:354-361`); the read window spans at most two pages, and probing `ret-8` and
`ret-1` covers both in every split (a boundary at exactly `ret-7` puts the whole window in
`page(ret-1)`; a boundary strictly inside puts the low part in `page(ret-8)`). Three probes is more
than needed on POSIX, which is fine.

The extraction is behaviour-preserving: `scan_stack_for_returns`
(`guest_stack_scan.hpp:67-83`) keeps the old order — stop on unreadable slot, skip `<= 0x10000`,
module filter, call-site test, dedupe — and `is_guest_call_site` is the same two `strcmp`s.
`keep`/`readable` are both null-guarded at `:47` and `:69`, and both are addresses of ordinary
internal-linkage functions with static storage duration, so there is no lifetime hazard. Windows
compiles the same construct; `prosper::host::` is fully qualified from inside `namespace prosper`,
which is what the Linux side already does.

**B2 — arms exist now, and three of the five carry real discrimination.** See C1 and N4/N5/N6 below
for what is wrong with them.

`git diff --check origin/main..573f40dbb` is clean, so N11(d) is fixed. N11(a) and N11(b) are fixed
in both copies. The FreeBSD `kern_mprotect` citation landed and reads correctly.

---

## Blocking

### C1. Windows MinGW is RED on this head, on the new test

```
prosper/tests/hle/test_false_success_nids.cpp:556:53: error: 'PROT_READ' was not declared in this scope
... 'PROT_WRITE' ... 'MAP_PRIVATE' ... 'MAP_ANONYMOUS' ... 'mmap' ... 'MAP_FAILED' ... 'munmap'
```

(job 103170067002, run 34570056303, head `573f40dbb`.)

The include is guarded — `#if !defined(_WIN32)` / `#include <sys/mman.h>` at the top of the file —
but the **uses at `:556-565` are not**, and `test_false_success_nids` is registered unconditionally
(`prosper/CMakeLists.txt:1049`, inside the plain `if(TARGET prosper_core)` block), unlike
`test_map_noreplace` and `test_prot_none` which sit inside `if(UNIX)` at `:1832`. The sibling file
in this same delta gets it right — `test_retrack_reservation.cpp` guards both the include and every
`munmap`/`mmap` use — so this is an inconsistency inside one PR rather than an unknown.

Either wrap the foreign-mapping arm in `#ifndef _WIN32` (with a `VirtualAlloc` equivalent if you
want the coverage on Windows), or follow `test_mprotect_unaligned.cpp`'s shape, which compiles
cross-platform with `#if defined(__linux__) || defined(__APPLE__)` around the body and an `#else`.

A Linux-only `ctest` cannot see this, which is exactly why the result is here rather than in the
author's log.

### C2. The hand-rolled 64 KiB rounding drops an overflow guard that the code it replaced had

`hle_kernel_mem.cpp:1870-1872` (and the identical Windows copy at `:6973-6975`):

```c
if (a1 > UINT64_MAX - a0) return 0x80020016ull;
const uint64_t base = (a0 + kCommitGranule - 1) & ~(kCommitGranule - 1);
const uint64_t end  = (a0 + a1) & ~(kCommitGranule - 1);
```

The guard covers `a0 + a1`. It does **not** cover `a0 + kCommitGranule - 1`, which is the round-up.
`normalize_guest_page_range` — the function this replaced — has both: `:873` for `addr + len` and
`:875` (`if (raw_end > UINT64_MAX - mask) return false;`) for the round-up. Only the first survived
the rewrite.

Worked instance, `sceKernelMemoryPoolDecommit(0xFFFFFFFFFFFFFFF0, 8, 0)`:

- `a1 (8) > UINT64_MAX - a0 (15)` is false, so the guard passes;
- `base = (0xFFFFFFFFFFFFFFF0 + 0xFFFF) & ~0xFFFF` wraps to **0**;
- `end = 0xFFFFFFFFFFFFFFF8 & ~0xFFFF = 0xFFFFFFFFFFFF0000`;
- `end > base`, so `len = 0xFFFFFFFFFFFF0000`;
- `committed_parts_in(0, 0xFFFFFFFFFFFF0000)` (`:1808-1819`) clips to `[0, 0xFFFFFFFFFFFF0000)`,
  which contains **every committed mapping in the process**, and the loop at `:1886-1901` replaces
  each with `PROT_NONE` and retags it `"reserved"`;
- the handler returns `0`.

So one garbage pointer in the top 64 KiB of the address space decommits all guest memory and reports
SCE_OK. The window is `a0 ∈ [0xFFFF_FFFF_FFFF_0001, 0xFFFF_FFFF_FFFF_FFFE]` with a small `a1` — `-1`
itself is excluded by the existing guard, but `-2` through `-0xFFFF` are not, and those are ordinary
sentinel/garbage shapes. On the previous head `normalize_guest_page_range` refused this call, so this
is a regression introduced by the delta, in the one handler whose comment block says its failure mode
is "unmapping a range the guest did not name, which is silent and corrupts whatever else lived
there".

`align_up_multiple` at `:861-868` is the checked round-up and is already in this file; using it, or
adding `if (a0 > UINT64_MAX - (kCommitGranule - 1)) return 0x80020016ull;`, closes it. Please add an
arm for it too — this is the kind of input a test can pass trivially.

### C3. The `AndGetId` zero-token fallback mixes the two token dialects the file separates, mutates ring state, and has no test

`hle_kernel_mem.cpp:2710` / `:7638`:

```c
return token ? token : prosper_apr_next_token(a1 ? (unsigned)(a1 - 1) & 0x3f : 0);
```

I did ask for this to be *decided*, and said either answer was fine. What I did not appreciate then,
and do now, is that the fallback is reachable only when `tag_echo` is true (`:612`: `bound && bc.eq`)
and `bc.tag == 0` — that is, a **bound** command buffer. For a bound cb this file states the contract
in as many words at `:558-562`: *"the completion token IS the binding tag"*, tracked by the guest in
a `{token -> callback}` hash. `prosper_apr_next_token` is the **unbound** dialect's token
(`:567-572`), and that same block records that inventing counter values in the bound flow was the
root cause of #180. The submit still posts `bc.tag` on the equeue at `:625`; only the value handed
back in `rax` changes. So the handler now returns an id that is, by the file's own documentation, not
the completion token for that submit.

Two further points:

- **It has a side effect.** `prosper_apr_next_token` is not a read — `hle_kernel_time.cpp:1896-1901`
  does `++state.ring_seq[ring]`. The fallback consumes a sequence value from the ring for a token
  that is never delivered anywhere.
- **It is untested, and it landed after the evidence.** No arm was added for the bound case (the
  second half of N9), and the FF Tactics boot the PR rests on was measured on `10c85ebb8`, where this
  path returned the tag. If FF Tactics' buffers are bound with a zero tag, the title now receives a
  different id than it did when the screenshot was taken.

Either is fine by me:
(a) revert to `return token;` and record in the comment *why* echoing the tag is right for a bound
submit even though 0 is indistinguishable from the old stub — "distinguishable from unimplemented"
is prosper's concern, not the guest's; or
(b) keep the fallback, add the bound-zero-tag arm, and say what evidence makes the counter dialect
correct for a bound submit. If you keep it, do not re-derive `ring`: `apr_submit_common` already has
a `token_out` out-parameter at `:553` and computes `ring` at `:554`, so hand the ring back rather
than writing `(a1 - 1) & 0x3f` a second place (and `prosper_apr_next_token` masks internally anyway,
so the `& 0x3f` is dead).

---

## Non-blocking

**N1. The Windows copy's 64 KiB justification does not describe Windows.** `:6961-6968` cites
`exec_image_linux.cpp` and its `mmap(page & ~0xffff, 0x10000, MAP_FIXED)`. Windows' own lazy-commit
path is `exec_image_win.cpp:1028-1033`: `page = a & ~0x3fff` and
`VirtualAlloc(page, 0x4000, MEM_COMMIT, …)` — a **16 KiB** granule. The hazard the comment describes
therefore does not exist on Windows in that form, and the Windows half is more restrictive than its
platform needs with a citation that does not hold there. Keeping the two halves identical is right;
say that ("the granule is unified with the Linux path deliberately"), rather than deriving it from a
file whose value is different. This is the citation-authority failure mode the charter records — a
reader who follows the pointer finds `0x4000`.

**N2. "Makes that unreachable by construction" is narrower than it reads.** The 64 KiB re-backing
still clobbers a live neighbour whenever the *commit* is sub-granule: `k_pool_commit` (`:1780-1789`)
forwards straight to `k_map_flexible` and imposes no granularity at all, so a 16 KiB pool commit
inside a 64 KiB reserved granule, followed by a touch of the reserved part of that granule, is
re-backed by the same whole-granule `mmap`. Decommit is no longer a route to it; it is not
unreachable. Worth an issue rather than a change here.

**N3. Commit and Decommit now disagree about granularity, and the disagreement fails in the silent
direction.** Commit accepts any length; Decommit below 64 KiB releases nothing and returns SCE_OK.
For a title whose pool commits are 16 KiB that is precisely the failure this function's own comment
exists to prevent — *"the guest is told 'released' about memory it can never get back"* (`:1829`).
NINJA GAIDEN 4 is unaffected, so this is not urgent, but consider returning EINVAL rather than
SCE_OK when `end <= base` **and** the requested range is in fact committed; that is the honest
answer, and it is the one `k_munmap` gives for the same shape.

**N4. `test_false_success_nids.cpp:548` is a tautology.** `recommitted` is a local initialised to
`base` at `:545` and never written afterwards — `k_pool_commit` takes the address by value and has no
out-parameter, so `CHECK(recommitted == base, "re-commit lands at the SAME address …")` cannot fail
under any mutation. The stated property is actually carried by `:549-551` (the store would SIGSEGV if
the range were not backed at `base`), which is real but reports as a crash rather than a `[FAIL]`. In
a file whose thesis is "each arm names the mutation it exists to catch", an arm naming one it cannot
catch is worth removing or rewriting.

**N5. The `foreign` arm is intermittently vacuous.** `mmap(nullptr, …)` guarantees page alignment,
not 64 KiB alignment. If the kernel returns a 4 KiB-aligned address that is not granule-aligned, then
`base = round_up(foreign)` and `end = round_down(foreign + 0x10000)` are equal, the `end <= base`
branch at `:1873` returns early, and the arm passes **without the clip ever running** — so it passes
whether or not the mutation it names is present, and which happens is a property of that run's
address-space layout. This is the same class as the two vacuous arms already caught in this PR, with
the added problem that it is not reproducible. Over-allocate `2 * kGranule` and align the pointer up,
or `CHECK` the alignment so the arm detects its own invalidity.

**N6. The sub-granule block has no fixture check.** `:572-573` puts `reserve` and `commit` inside the
`if` condition with no `CHECK`; if either fails, both arms disappear and the test still prints
`== PASS ==`. Every other fixture in this function is a `CHECK`.

**N7. `src/host/fault/AGENTS.md` was not updated.** Its `## What is here` list names
`fault_context.hpp`, `trap_arbitration.hpp` and `rbp_chain.hpp`; `guest_stack_scan.hpp` is now the
fourth file (`guest_caller.hpp` is a pre-existing omission). The charter asks the folder map be
corrected when what the folder holds changes, and this one is a single bullet that belongs directly
beside the `rbp_chain.hpp` line — the new header is the second instance of exactly the argument that
bullet already makes.

**N8. Cosmetic.** `narrowed.load()` / `partial.load()` in the two new `fprintf`s report a count that
may already have moved; `fetch_add` returns the value you want. Also note the new test now emits
three `[memhle]` lines on a normal green run (`:536`, `:561`, `:575` all trip one of the two
diagnostics) — harmless, but it will look like a failure to someone reading ctest output for the
first time, so a one-line comment saying the lines are expected would help.

---

**REJECTED** on C1, C2 and C3. C1 is already red in CI and is a two-line guard. C2 is a one-line
overflow check plus an arm, and it is the one I would fix first — it is a silent whole-address-space
unmap reachable from a garbage pointer, in the handler this PR added. C3 is a choice, not a bug, but
it needs either reverting or the arm it was asked for. B1 and B3 I re-derived independently and am
satisfied with; the shared header is a genuine improvement and the mprotect change itself remains
right. Happy to re-review promptly.

mattias800 (commented):
## Independent review (delta 2) — PR #3530 @ `4ce37c6285de143acc635deb46b2043ff2a65ff6`

Reviewed `573f40dbb..4ce37c628` only. All three blockers are genuinely fixed; I traced each rather
than taking the report.

### C2 — fixed, and the boundary is right

`hle_kernel_mem.cpp:6987-6988` (and the POSIX twin):

```c
if (a1 > UINT64_MAX - a0) return 0x80020016ull;                   // the span itself wraps
if (a0 > UINT64_MAX - (kCommitGranule - 1)) return 0x80020016ull; // rounding the base up wraps
```

The second predicate rejects `a0 >= 0xFFFF_FFFF_FFFF_0001` and admits `0xFFFF_FFFF_FFFF_0000`
exactly, where `a0 + 0xFFFF == UINT64_MAX` and the round-up is still exact — the correct boundary,
not one off it. With both guards, `base = round_up(a0) <= 0xFFFF_FFFF_FFFF_0000` and
`end = (a0+a1) & ~0xFFFF <= UINT64_MAX` cannot wrap, `end > base` implies `[base, end)` is a
granule-aligned subrange of the request, and `base + len == end` so `committed_parts_in`'s own
`base > UINT64_MAX - len` guard is satisfied. The `(-16, 8)` path is closed.

### C3 — clean revert, and the recorded reasoning is accurate

`return token;` at `:2716` / `:7659`. The comment states the case correctly: the fallback was
reachable only for a cb bound with a zero tag, the bound dialect's token is the binding tag
(`apr_submit_common:558-562`), and CRI ADX2's waiter tests the event ident rather than the tag
(`:574-585`). The ring side effect is gone with it.

### C1 — fixed; and yes, `-fsyntax-only` is sufficient for *this*, but not for what the fix exposes

For the break it replaced, `-fsyntax-only` is complete evidence: undeclared identifiers are a
semantic-analysis diagnostic, and GCC's `-fsyntax-only` runs the full front end including template
instantiation — it only skips codegen and link. Two caveats, one small and one that matters:

- it needs the real include set to mean anything (CI compiles with `-Iprosper/src
  -Ithird_party/libatrac9/src -std=gnu++20`); a run without those would fail for the wrong reason or
  not reach the file at all. Assuming you used them, this class is covered.
- **it cannot see runtime, and the guard you added is what makes that load-bearing.** Only the
  `foreign` arm is inside `#if !defined(_WIN32)`. Everything else in `test_pool_decommit` —
  `:519-551`, `:584-597`, `:613-621` — now runs on Windows, so this PR makes the `Windows MinGW` job
  the **first thing that has ever executed the Windows `k_pool_decommit`**: `win_unmap` on a pool
  reservation, `track(…, "reserved")` over the hole, then a re-commit at the same address. That job
  does run ctest (`.github/workflows/ci.yml`, its `Test` step, with `--no-tests=error`), so the
  coverage is real — but it is `IN_PROGRESS` on this head as I write. Please wait for that specific
  job rather than merging on the syntax-only pass; it is the only evidence that path works.

### C1b — the arm is now reliably non-vacuous, and the "positive control" is not a control

The ×3 over-allocation plus round-up does fix N5: `foreign + kGranule <= foreign_raw + 3*kGranule`
holds for every value `mmap` can return, so the request always contains a whole granule and always
reaches the clip. That half is right.

You asked me to check the control, so: **`test_false_success_nids.cpp:569` is a tautology.**
`foreign` is produced by `& ~(kGranule - 1)` on the line above, so `(foreign & (kGranule - 1)) == 0`
asserts a property of that arithmetic and nothing else — it cannot fail for any value of
`foreign_raw`, including `MAP_FAILED`. It is labelled "fixture positive control", which is the
reading that makes it worse than absent: a line that says it proved the lever moved, which could not
have reported otherwise. Delete it, or replace it with something that could fail — the real
precondition is that the aligned granule fits inside the raw mapping, so
`CHECK(foreign + kGranule <= (uint64_t)(uintptr_t)foreign_raw + kGranule * 3, …)` at least fails if
someone later trims the over-allocation.

---

## Non-blocking (please fix before merging; none needs re-review)

**F1. The span-overflow arm does not discriminate the guard it names.** `:616`,
`decommit(~0ull - 1, 4, …)`: `a0 = 0xFFFF_FFFF_FFFF_FFFE`, which is `> 0xFFFF_FFFF_FFFF_0000`, so the
**round-up** guard catches it too. Delete the span guard and the suite stays green — so guard 1 has
no discriminating arm, which is exactly the property the comment at `:605-607` claims both arms
together establish.

An input that isolates it is `decommit(0x10000, ~0ull, …)`: `a1 > UINT64_MAX - a0` fires, while
`a0 > UINT64_MAX - 0xFFFF` does not. And it fails *cleanly* under the mutation — without guard 1,
`base = 0x10000` and `end = (0x10000 + ~0ull) & ~0xFFFF = 0`, so `end <= base` returns `0` instead of
EINVAL and you get a named `[FAIL]` with the process intact. Worth noting while you are there that
the span guard is not memory-safety-load-bearing once the round-up guard exists — for any `a0` the
round-up guard admits, a wrapping span yields `end < base` and the early return — so its removal
would be a plain false success rather than a corruption. That is still worth an arm in this file of
all files, just not a crash-shaped one.

**F2. On the crash-signalled mutation (your direct question): acceptable, documented well, and
avoidable here at zero cost.** ctest scores exit 139 as a failure, so the property is protected, and
writing down what the mutation looks like is the right instinct. Two reasons to make it clean
anyway: stdout is fully buffered under ctest's pipe, so the SIGSEGV discards the `[FAIL]` lines
already printed — the mutation's report is less informative than it looks; and a crash cannot say
*which* property broke, since any unrelated fault gives the same 139. Concretely, `:620`
dereferences `base`; reading the tracker instead does the same job without touching memory:

```c
CHECK(prosper_reserved_range_state(base) == 2,
      "an overflowing decommit released NOTHING (mutation: drop the guard -> this reads 1)");
```

`base` is committed at `:546` and nothing between there and `:620` touches it, so this is `2` in the
unmutated run; under the mutation the mass unmap retags it `"reserved"` and it reads `1`. Keep the
dereference after it if you like — the named failure will already have printed.

**F3. The Windows granule note is in the wrong place and mis-indented.** `:7005-7010` sits *inside*
the `for (const auto& part : …)` body at 4 spaces against the surrounding 8, glued onto the front of
the `win_unmap` comment it interrupts. More importantly, the text it corrects is unchanged: the
granule rationale at `:6970-6980` still presents `exec_image_linux.cpp`'s
`mmap(page & ~0xffff, 0x10000, MAP_FIXED)` as this half's reason, and it still comes first. A reader
of the Windows half reads the Linux justification, and only finds the correction if they read into
the loop. Move the paragraph up next to `constexpr uint64_t kCommitGranule` and re-indent. The
content is right — 64 KiB-aligned implies 16 KiB-aligned, so the superset argument holds.

**F4. `recommitted == base` at `:548` is still a tautology** (my N4, unaddressed). `recommitted` is a
local initialised to `base` and never written; `k_pool_commit` takes the address by value and has no
out-parameter. That is now the second assertion in this one function that cannot fail, in the file
whose subject is assertions that cannot fail.

**F5. The sub-granule fixture still has no CHECK** (my N6). `:585-586` puts `reserve` and `commit`
inside the `if` condition; if either fails, both arms vanish and the test prints `== PASS ==`.

**F6. `src/host/fault/AGENTS.md` is now missing TWO files this PR adds.** `guest_stack_scan.hpp`
*and* `guest_caller.hpp` are both new against `origin/main`, and that file's `## What is here`
section is a literal enumeration of the folder's contents — it names `fault_context.hpp`,
`trap_arbitration.hpp` and `rbp_chain.hpp` and stops. Two bullets. The `guest_stack_scan.hpp` one
belongs directly under the `rbp_chain.hpp` line, since it is the second instance of the argument that
bullet already makes.

**F7. N2 and N3 from the last round are unaddressed and no issue exists for either.** (a) the
comment's "makes that unreachable by construction" is still broader than the code earns — a
*sub-granule commit* (`k_pool_commit` imposes no granularity) still lets the 64 KiB lazy re-backing
clobber a live neighbour, with no decommit involved; (b) Commit accepts any length while Decommit
now silently releases nothing below 64 KiB and answers SCE_OK, which is the "told 'released' about
memory it can never get back" shape this function's own comment at `:1829` exists to prevent. No
title is affected today. Please file them before merging so they are in the record rather than in
this thread.

**F8.** Cosmetic, unchanged: `narrowed.load()` / `partial.load()` at `:1887`, `:1915`, `:6999`,
`:7026` report a count that may already have moved; `fetch_add` returns the one you want.

---

### On your question about the previous eight

None of them should have been blocking, and I do not want to reclassify any now. N5 was the closest
and it is fixed. N2/N3 are issues about a design that is deliberate and currently harms no title;
N4/N6/N8 are test and log hygiene; N1/N7 are documentation. None changes what any title does. F1 is
the only one of today's that I would have hesitated over, and it is a weaker case than B1 was last
round: B1 removed discrimination for an existing, referenced property (#387 F3), whereas F1 is a new
arm that is weaker than its own comment claims, guarding a predicate whose removal is a false success
rather than a corruption.

---

**APPROVED** for head `4ce37c6285de143acc635deb46b2043ff2a65ff6`, with one condition I want stated
because it is not a formality: **merge only once `Windows MinGW` is green on this exact head**, not
on the `-fsyntax-only` pass. That job is the first and only execution of the Windows
`k_pool_decommit` path, and a syntax check cannot see it. The F-items above are small and I do not
need to see them again — fix them in place.

C2 in particular is a good catch to have landed: the guard now rejects the case at the right
boundary, and the arm that proves it is the one that matters.

PR #3534, opened 2026-09-11
worktree_reclaim: removal status must not land on the --json stdout (#3397)

### LINKED ISSUE #3397: worktree_reclaim: actual removal breaks --json stdout
Verified on main 769a599b1 during completed PR cleanup. `prosper/tools/worktree_reclaim.py:912` emits a JSON census, and main routes its own human messages to stderr at :943, but `recheck_and_remove` still prints REMOVED directly to stdout at :860 (and REFUSE/FAILED paths do likewise). Thus `--json --remove --yes --only <OWN_FINISHED_WORKTREE>` successfully removes the selected worktree, exits 0, then leaves a trailing non-JSON status line; `jq` fails parsing the captured output. This is an output-contract defect, not a failed removal or evidence that safety guards were bypassed. Suggested fix: pass the selected status sink into recheck_and_remove, or consistently send its human status messages to stderr. Add a mocked-removal regression asserting stdout is one parseable JSON document and status text is on stderr; cover success and refusal/failure without deleting real user worktrees. No matching existing issue found.

mattias800:
Fixes #3397.

## Failure scenario

`prosper/tools/worktree_reclaim.py --json` promises stdout is a machine-readable census, and
`main` deliberately routes its own human messages to stderr so that promise holds. The removal
loop was left behind: `recheck_and_remove` printed its `SKIP` / `REFUSE` / `FAILED` / `REMOVED`
lines straight to stdout.

So `--json --remove --yes --only <finished-worktree>` **removes the tree and exits 0**, then
appends a line like

```
  REMOVED <WORKTREE>  (squash-pr#3384, idle 41.2h)
```

after the closing `}`. A caller piping that into `jq` gets a parse error on a run that did exactly
what it was asked to do. This is an output-contract defect, not a failed removal and not evidence
that any safety guard was bypassed — but a consumer cannot tell those apart, and both natural
reactions (re-run the removal, or stop trusting the guards) are worse than the bug.

## Contract

In `--json` mode, **stdout is exactly one JSON document and nothing else**; every human status
line goes to stderr. Without `--json`, status lines keep going to stdout, where a person running
an interactive removal needs to see what was deleted.

## Approach

`recheck_and_remove` takes the caller's status sink as a **required keyword-only `say`
parameter**, and `main` passes the same sink it already computes from `args.json`. The four
`print` call sites become `say`.

`say` is required rather than defaulting to `print` on purpose: a default makes the
stdout-polluting case the one no caller has to think about, which is precisely how the split
arrived — `main` moved its messages to stderr and this function silently kept the old behaviour.
There is one other caller, an existing test arm, and it now names its sink too.

## Invariants

- The JSON census on stdout is unchanged in content and structure.
- Guard behaviour, ordering, re-check semantics, exit codes and the `_GH_CACHE` reuse are all
  untouched; this PR changes **which stream text is written to**, nothing else.
- Non-`--json` output is byte-identical to before.

## Regression arm

New `test_json_stdout_carries_only_the_census` in `prosper/tools/test_worktree_reclaim.py`
(registered already as ctest `worktree_reclaim_tool`; no CMake change needed).

**Removal is mocked** — a `W.run` wrapper intercepts the single `git worktree remove` argv and
returns a chosen rc. Two reasons, and neither is speed:

1. This suite runs on machines where other lanes hold live worktrees, and the charter records
   that removing a tree a shell sits in wedges that shell irrecoverably. The arm deletes nothing.
2. It is the only route to the `FAILED` branch at all: provoking a real `git worktree remove`
   failure means first building a tree that every guard above it already refuses, so the code
   would never be reached.

The arm asserts **which stream** each line landed on, never merely that the text exists, and
carries three controls so that a degenerate "fix" cannot pass it:

- the stdout census must still list every tree and still mark the candidate removable — so
  "print nothing on stdout" fails;
- without `--json`, `REMOVED` must still appear on stdout — so "silence the status lines" fails;
- the refusal arm dirties the tree **after** the census, so the `REFUSE` line is emitted from
  inside the removal loop rather than by the census, which is the one path the census cannot
  produce on its behalf.

Branches covered: `REMOVED` (success), `FAILED` (git rc≠0), `REFUSE` (guard fires at re-check),
`SKIP` (tree unregistered between census and removal).

## Without-fix verification

With the fix reverted locally (the four `say(` calls in `recheck_and_remove` put back to
`print(`, signature untouched) the suite goes **red: exit 1, 10 failed checks**, including
`stdout parses as one JSON document: got False` with the tail line
`  REMOVED .../safe  (ancestor, idle 400.0h)`. Restored, and green again.

## Affected / unaffected

Affected: the stream `worktree_reclaim.py` writes removal status to under `--json`.
Unaffected: every guard, the census content, exit codes, human-mode output, and every other tool.

## Risks

Low. The one behavioural change is stream selection. The signature change is internal to this
tool and its test; no other file in the repository references `recheck_and_remove`.

## Commands and results

```
# the registered ctest case, via ctest
ctest --test-dir prosper/build-linux --no-tests=error -R worktree_reclaim_tool --output-on-failure
  -> 1/1 passed, 0 failed, exit 0   (worktree_reclaim_tool, 2.34 s)

# the suite directly, for the per-check count
python3 prosper/tools/test_worktree_reclaim.py
  -> exit 0, 97 `ok` checks, 0 FAIL, "all worktree_reclaim checks passed"
     (was 74 `ok` before this PR; the new arm adds 23)

# mutation arm: fix reverted
python3 prosper/tools/test_worktree_reclaim.py
  -> exit 1, 10 FAIL

# real-repo smoke, dry of any candidate, on the actual checkout
python prosper/tools/worktree_reclaim.py --json --no-github --no-fds --no-maps \
    --remove --yes --only __no_such_tree__ 2>err | python -c "import json,sys; ..."
  -> "parsed ok, trees: 9"; the "Removing 0 tree(s)" / "removed 0 of 0 candidate(s)" prose
     appeared on stderr
```

No C++ was rebuilt and no full `ctest` run was made: the diff is two Python files with no
compiled dependant, so a full suite could only report on code this PR cannot reach. CI runs the
whole thing regardless.

### DIFF
--- a/prosper/tools/test_worktree_reclaim.py
+++ b/prosper/tools/test_worktree_reclaim.py
@@ -15,6 +15,8 @@ Run directly, or via ctest as worktree_reclaim_tool.
 
 from __future__ import annotations
 
+import contextlib
+import io
 import json
 import os
 import subprocess
@@ -530,7 +532,7 @@ def test_removal_rechecks_state_rather_than_trusting_the_census() -> None:
         try:
             time.sleep(0.4)
             removed = W.recheck_and_remove(str(repo), trees["racy"], "origin/main", 12.0,
-                                           False, False, False, 0, False)
+                                           False, False, False, 0, False, say=print)
             check("recheck refuses", removed, False)
             check("tree still on disk", wt.is_dir(), True)
         finally:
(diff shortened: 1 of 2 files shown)


mattias800 (commented):
APPROVED — independent review on exact head `581b604e` (I am not the author).

## What I checked, and how

**Completeness of the conversion.** Every status line inside `recheck_and_remove` now goes through
`say`: SKIP, both REFUSE branches, FAILED and REMOVED. Verified there is no bare `print(` left
anywhere in the function body on this head, and that the three call sites in the tree (two in the
test, one in `main`) all pass `say=`.

**The design choice is the right one, and it is load-bearing.** Making `say` keyword-only *and
required* rather than defaulting to `print` means a future caller that forgets it gets a `TypeError`
at the call, not a silently polluted document. That directly addresses how the defect arose — `main`
moved its own output to stderr and this function was left behind — and a default would have
reinstated exactly that trap. The docstring says so, which is the part a later reader needs.

**The regression arm distinguishes the right thing.** My concern going in was the usual one for a
mocked test: that it could pass for a reason other than the fix. It does not.

- Every assertion pins **which stream** a line landed on, not merely that the text exists somewhere.
- Two controls rule out the degenerate "fixes": the stdout census must remain *complete* (so sending
  everything to stderr fails), and in non-`--json` mode the same lines must still reach stdout (so
  silencing them fails).
- The `--json` arm asserts `json.loads(stdout)` succeeds **and** that the census still lists every
  tree and still marks the candidate removable — so a fix that emptied the document would fail too.

**The vacuous-pass hazard was caught by the author, not left for review.** Restoring `file.txt` in
the REFUSE arm updates its mtime, which legitimately trips the `recent` guard; without the explicit
re-age before the human-mode control, that control would have run against a candidate list of zero
and passed while asserting nothing. The comment says this outright. That is the failure mode
CLAUDE.md's positive-control rule is about, and finding it unprompted is the strongest signal in
this diff.

**Mocking the removal is correct here, not a shortcut.** Two independent reasons, both stated in the
test: this suite runs on machines where other lanes hold live worktrees, and removing a tree a shell
is `cd`'d into wedges that shell irrecoverably — the issue itself asked for this. Separately, the
FAILED branch has no other route, since provoking a real `git worktree remove` failure requires a
tree every guard above it already refuses. The mock intercepts exactly one argv (`worktree remove`)
and delegates everything else to the real `run`, so the guards under test are still the real ones.

Per the charter I did not re-run the author's suite. I confirmed independently: 11 checks pass, 0
fail, 0 pending, 2 skipping, and the PR head still matches the branch tip (`581b604e`) — checked with
`prosper/tools/ci/pr_merge_gate.py`, which reports GREEN.

## Nothing blocking

One non-blocking observation for a future pass, not this PR: `report()` is still called before `say`
is constructed in `main`, so it continues to make its own stream decision internally. That is
pre-existing and correct today, but it is the second place that knows the `--json` routing rule, and
a third would be worth consolidating.

Issue #133, opened 2026-07-08
[hardcoded] vk_color_format maps ONLY COLOR_8_8_8_8 — every other format silently resolves to Undefined


mattias800:
From the #72 backlog. vk_translate.cpp:~20 vk_color_format handles one hardware format (0xA, two number-types, two swaps); every other CB format (16_16_16_16_FLOAT, 10_10_10_2, 11_11_10, 5_6_5, single/dual-channel) silently returns VkFormat::Undefined, which resolve_pipeline_state passes through unlogged. Any non-8888 render target -> undefined pipeline format (validation failure or garbage). Fix: port the remaining RenderTextureFormat rows from Kyty (at least 16161616F and 1010102); log once per unmapped triple. Introduced 822cd24. (gpu)

mattias800:
CLAIMING: claude-session-896257ae (worktree bughunt-batch) | branch fix/issue-133-color-formats | 2026-07-09T00:43:23Z

Issue #132, opened 2026-07-08
[simplification] sceAgcDcbEventWrite drops its address arg — label-carrying events can never write


mattias800:
From the #72 backlog. hle_agc.cpp:~136 agc_dcb_event_write encodes only event_type&0xff into a 2-dword packet; the address argument (a2) is discarded and the CommandProcessor treats EventWrite as a no-op. An address-carrying EVENT_WRITE (timestamp/label variant) loses its write target, so a guest waiting on such a label blocks forever. Fix: widen the packet to carry a2 (lo/hi) like ReleaseMem and honor address-carrying event types in command_processor.cpp. Introduced 2e4b9de. Latent (Messenger fences use ReleaseMem/WriteData).

mattias800:
CLAIMING: claude-session-896257ae (worktree bughunt-batch) | branch fix/issue-132-event-write-addr | 2026-07-09T02:00:34Z

PR #3535, opened 2026-09-11
perf: sample the Windows working set and commit charge into the F8 capture

### LINKED ISSUE #3448: Performance captures carry no RSS on Windows: rss_bytes is gated on __linux__ and reads /proc/self/statm
Every `.prperf` taken on Windows reports `RSS: unavailable on this platform/run`, so the memory axis is missing from exactly the platform where it currently looks most significant.

## Cause

`prosper/frontends/shared/perf/performance_capture.cpp:562`:

```cpp
#if defined(__linux__)
    std::ifstream statm("/proc/self/statm");
    ...
        sample.rss_bytes = resident_pages * static_cast<uint64_t>(page_size);
#endif
```

`ProcessSample::rss_bytes` is `std::optional<uint64_t>` and stays empty on every other platform. The reporting side is already correct and platform-neutral — `performance_capture_report.py:554` prints `RSS: unavailable on this platform/run` when the population is empty, and `performance_capture.cpp:314` computes an `rss_available` flag — so only the sampler needs a Windows arm.

Note the neighbouring CPU timer already has one: the `#else` branch just above uses `std::clock()`, and there is a `_WIN32` arm for process CPU ticks. RSS was simply never given the same treatment.

## Why it matters here

Measured on Windows/NVIDIA, GTA V (`PPSA04263`) world gameplay on `main` `d9d03f8a`, with the tuned cache environment from `GTA5_STATUS.md`:

- working set **19.8 GB**, private bytes **24.4-25.4 GB**
- free physical memory fell to **7.0 GB** of 63.7 GB
- the configured cache budgets total ~24 GB (`PROSPER_BACKEND_TARGET_CACHE_MB=8192`, `..._TEXTURE_CACHE_MB=8192`, `PROSPER_COMPUTE_IMAGE_CACHE_MB=4096`, `PROSPER_TEXTURE_DECODE_CACHE_MB=4096`)

Those numbers had to be taken with `Get-Process` alongside the capture, and they cannot be correlated with the capture window because they are not in the file. A capture that cannot show memory growth cannot support or refute a cache-pressure hypothesis, and cache sizing is an active lever in the current performance work (#3407, #3440, #3438).

## Suggested fix

Add a `_WIN32` arm using `GetProcessMemoryInfo` / `PROCESS_MEMORY_COUNTERS_EX` (`WorkingSetSize`, and ideally `PrivateUsage`, which is the figure that shows cache growth here — working set understated it by ~5 GB in the run above). macOS would use `task_info` / `TASK_BASIC_INFO`. Keep the optional empty when the query fails so `rss_available` stays honest.

Consider recording private/committed bytes as a separate optional field rather than overloading `rss_bytes`, since the two diverge materially under large host-side caches.

Found while investigating GTA V performance on Windows.

mattias800:
Fixes #3448.

## Failure scenario

`collect_process_sample` read the resident set from `/proc/self/statm` under `#if defined(__linux__)`
and nothing else, so `ProcessSample::rss_bytes` stayed empty on every other platform. Every `.prperf`
taken on Windows therefore printed `RSS: unavailable on this platform/run`, and the memory axis was
missing on exactly the platform where host-side cache pressure is the live performance question: a
capture that carries no memory series can neither support nor refute a cache-pressure hypothesis, and
the numbers in #3448 had to be taken with `Get-Process` beside the run, where they cannot be
correlated with the capture window.

The reporting side was already platform-neutral and needed only the new field; the sampler was the
whole defect.

## Contract

- `rss_bytes` is the process **resident set**, from whichever counter the host exposes:
  `/proc/self/statm` (Linux), `WorkingSetSize` (Windows), `MACH_TASK_BASIC_INFO`'s `resident_size`
  (macOS).
- `private_bytes` is the **commit charge**, recorded only where the host counts it separately from
  the resident set — today that is Windows' `PrivateUsage`.
- Either field is **empty when the host has no such counter or the query fails**. Empty is not zero:
  the header's existing `rss_available` flag and the new `private_bytes_available` flag are computed
  from the samples, and the report prints `unavailable on this platform/run` rather than a number.
- The `.prperf` format stays at **version 1**.

## Approach

`_WIN32` arm: one `GetProcessMemoryInfo` call with `PROCESS_MEMORY_COUNTERS_EX` fills both figures,
and both are assigned only inside the success branch. `psapi` is linked on `WIN32` — the function is
not in the default kernel32 set.

`__APPLE__` arm: `task_info` / `MACH_TASK_BASIC_INFO` for the resident set only. `MACH_TASK_BASIC_INFO`
rather than `TASK_BASIC_INFO` because the latter's `resident_size` saturates at 4 GiB, below the
working sets this capture exists for. macOS's private footprint lives behind a different call
(`task_vm_info`'s `phys_footprint`), so `private_bytes` is left empty there rather than filled from a
counter that does not mean it. **Marked `CONFIDENCE: LOW` and not verified by me** — see *what I could
not verify* below.

**Why a separate field rather than overloading `rss_bytes`** (the issue left this open): the two
answer different questions and diverge in both directions, so one field would have to pick a lie.
In #3448's GTA V run the working set understated the commit charge by ~5 GB, and only the commit
charge tracks the caches. In the small test process measured here the inequality inverts —
`rss=5,431,296` against `private=864,256`, because the working set counts shared image pages this
process did not commit. A single field would make `RSS` mean something different per platform and per
workload; two fields let the report say which counter it read.

**Backward compatibility.** The record schema is JSON-lines and every reader selects by name, so the
addition is compatible in both directions without a version bump: an older `.prperf` has no
`private_bytes` key and reads as unavailable, and a reader that predates the field ignores the new
key. Bumping the version would have been the incompatible choice — `validate_capture` compares the
version for **equality**, so `version: 2` would reject every capture already on disk. Verified against
a real pre-fix Windows capture (below).

## Invariants

- A failed query leaves both optionals empty, so `rss_available` keeps meaning "some sample carried a
  real measurement".
- A measured zero is serialized as `0` and reported as `0.0 MiB`; only an absent measurement prints
  `unavailable`. The reader filters on `is not None`, not on truthiness.
- The sampler stays allocation-free and syscall-cheap: it runs at 4 Hz on the app thread whether or
  not F8 is ever pressed.
- The three availability flags are now computed by one predicate, so a newly added field cannot pick
  up an availability rule that quietly differs from its neighbours'.

Affected: the process sampler, the capture header and sample records, and the offline report's memory
line. Deliberately unaffected: the capture format version, the pre/post windowing, every renderer and
compute record, the classification logic, and `rss_bytes` on Linux (same bytes from the same file).
Linux gains no `private_bytes` — it has no commit-charge counterpart, and filling it with a near
synonym of the resident set would make the field mean different things per platform.

## Risks

- The macOS arm is compiled and executed by CI's macOS jobs but was not authored or run on a macOS
  host. If it is wrong at runtime it is wrong in the direction of the status quo (an empty optional);
  if it is wrong at compile time the macOS jobs go red, which is the fail-visible direction.
- `psapi` is a new link dependency for `prosper_performance_capture` on Windows. It is a standard
  import library present in both MinGW and MSVC toolchains, and the propagation through CMake's
  static-library `PRIVATE` (`LINK_ONLY`) interface is exercised by the linked test binary and by the
  full Windows build below.
- `PrivateUsage` is commit charge, not physical memory. The report labels it `private/committed` so
  it is not read as a second RSS.

## Build and test

Every status below was read from the command itself (bare, or redirected to a log with the status
taken outside the wrapper) — no build or test result here came through a pipe.

**Windows — MinGW UCRT g++ + Ninja + Vulkan SDK 1.4.350. This is the platform being fixed, and it
built and ran here.**

```
cmake -S prosper -B prosper/build-win -G Ninja \
      -DCMAKE_C_COMPILER=<mingw>/gcc.exe -DCMAKE_CXX_COMPILER=<mingw>/g++.exe \
      -DGAME_DUMP=<DUMP_ROOT>/PPSA24651-app0
cmake --build prosper/build-win -j6                 # exit 0, 645/645 targets
ctest --test-dir prosper/build-win --no-tests=error --output-on-failure --timeout 600
```

ctest: **334 tests discovered, 330 passed, 2 skipped, 2 failed, exit 8**. Both failures are
environmental and provably untouched by this PR:

- `session_start` — `AssertionError: 9009 != 0 : Python was not found; run without arguments to
  install from the Microsoft Store...`. Windows' App Execution Alias intercepting `python`; the test
  never reaches repository content.
- `build_revision_refresh` — the mini-project it generates crashes its own `revision_query.exe` with
  `exit status 3221225477` (0xC0000005). That project includes only `prosper/cmake/ProsperBuildRevision.cmake`
  and links only `test_build_revision`; it never references `prosper_performance_capture` or any file
  in this diff, and the neighbouring `check_build_revision` passes.

The `performance_capture` case (#133) passes and now exercises the live sampler, printing what it
read so the numbers are in the log rather than asserted blind:

```
host process sample: rss=5431296 private=864256
PASS: 48 checks, 0 failures
```

**Linux (WSL Ubuntu-24.04) — regression check, targeted.**

```
cmake --build prosper/build-linux -j5 --target test_performance_capture     # exit 0
ctest --test-dir prosper/build-linux --no-tests=error --output-on-failure -R performance_capture
```

**3 tests discovered, 3 passed, exit 0** (`prosper_app_performance_capture_schedule`,
`performance_capture`, `performance_capture_report_logic`). On Linux the same live-sampler check
asserts the opposite arm: `rss_bytes` present, `private_bytes` absent rather than zero.

A **full** Linux build is not attainable in this environment and I did not get one: it stops at
`prosper_live_renderer` with

```
prosper/frontends/shared/device/vulkan_runtime.hpp:11:51: error: 'VK_API_VERSION_1_4' was not
declared in this scope
```

because this WSL has `libvulkan-dev 1.3.275.0-1build1` (`VK_HEADER_VERSION_COMPLETE` = 1.3) while the
runtime header requires 1.4. That file is not in this diff, everything up to 78% built, and the
Vulkan-dependent chain is covered by the Windows build above and by CI's Linux job.

**Offline report tests**: `python tools/perf/test_performance_capture_report.py` — **40 tests, OK,
exit 0**.

**End-to-end on Windows.** A real `.prperf` written by the live sampler, read by the updated report
(exit 0):

```
RSS: 20.5..44.5 MiB
private/committed: 16.9..40.9 MiB
```

Both series track an 8 MiB-per-sample allocation the emitting process made during the window — the
growth the pre-fix capture could not show at all.

**Backward compatibility, against a real pre-fix capture** (`perf_capture_PPSA25009_...prperf`,
written on Windows on 2026-08-07: `rss_bytes: null`, and no `private_bytes` key anywhere in the
file). It validates and summarizes cleanly, exit 0:

```
RSS: unavailable on this platform/run
private/committed: unavailable on this platform/run
```

**Mutation arms — each new test was confirmed to fail without the fix.**

- The three new reader tests, against `origin/main`'s `performance_capture_report.py`:
  `FAILED (errors=3)` of 40, `KeyError: 'private_min'`.
- The C++ test, compiled against a copy of the sampler with only the `_WIN32` memory arm removed:
  `host process sample: rss=unavailable private=unavailable`, `FAIL: 48 checks, 4 failures`, exit 1 —
  i.e. exactly the pre-fix behaviour this PR is about.

**Gates**: `git diff --check` clean; `check_contribution_shape.py --selftest` 17 arms / 0 failed, and
`--base origin/main` reports "7 changed path(s), both rules satisfied"; `check_numbered_table.py` over
every tracked `.md` clean; no `Claude-Session:` trailers (`grep -c` = 0).

## CI

All 13 checks green on the exact head `c41fd62`: `Linux`, `Windows MinGW`, `Windows App`,
`macOS (x86_64 / Rosetta 2)`, `macOS App (SDL3 + MoltenVK)`, `Linux App`, `Linux App Package`,
`Host ASan + UBSan`, `Host TSan`, `Contribution shape`, `Docs` (+2 skipping). Two of them cover
paths my local builds could not: `Windows App` links the perf library into the SDL frontend (the
`psapi` propagation my `PROSPER_APP=OFF` build did not exercise), and the macOS job both compiles
and runs the `__APPLE__` arm.

## What I could not verify

- **macOS**: no macOS host here, so the `__APPLE__` arm was never executed *by me* and is marked
  `CONFIDENCE: LOW` in the source. It is not unexecuted, though: CI's `macOS (x86_64 / Rosetta 2)`
  job compiled it and ran `performance_capture` against it — `132/316 Test #132: performance_capture
  ... Passed`, 100% of 316 — which means the live-sampler check passed there, i.e. `task_info`
  returned a plausible resident set and `private_bytes` was correctly left unavailable. What remains
  unverified is only whether `resident_size` is the *right* macOS counter to call RSS; the call
  itself works. The arm can be dropped without affecting the Windows fix if a reviewer prefers.
- **MSVC**: the Windows toolchain here is MinGW UCRT, which is also what CI's Windows job uses. The
  `PROCESS_MEMORY_COUNTERS_EX` / `psapi` usage is not MinGW-specific, but no MSVC build was run.
- **A live game capture**: verified with the real sampler writing a real capture file, not by pressing
  F8 during a GTA V run. The field it adds is the one #3448 wants for that run.
- **A clean full ctest on either platform**: Windows has the two environmental failures above, and
  Linux cannot build its Vulkan chain here at all.

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -127,6 +127,11 @@ endif()
 add_library(prosper_performance_capture STATIC frontends/shared/perf/performance_capture.cpp)
 target_include_directories(prosper_performance_capture PUBLIC frontends)
 target_link_libraries(prosper_performance_capture PRIVATE prosper_build_revision)
+# Windows: the process memory sample (working set + commit charge) calls GetProcessMemoryInfo, which
+# lives in the PSAPI import library rather than the kernel32 set every target links by default.
+if(WIN32)
+  target_link_libraries(prosper_performance_capture PRIVATE psapi)
+endif()
 
 # --- tests (agentic-first: self-checking, exit-code = truth) -------------
 enable_testing()
--- a/prosper/frontends/shared/perf/performance_capture.cpp
+++ b/prosper/frontends/shared/perf/performance_capture.cpp
@@ -21,6 +21,8 @@
 #define NOMINMAX
 #endif
 #include <windows.h>
+// GetProcessMemoryInfo + PROCESS_MEMORY_COUNTERS_EX. psapi.h must follow windows.h.
+#include <psapi.h>
 #include <fcntl.h>
 #include <io.h>
 #include <sys/stat.h>
@@ -31,6 +33,10 @@
 
 #if defined(__linux__)
 #include <unistd.h>
+#endif
+
+#if defined(__APPLE__)
+#include <mach/mach.h>
 #endif
 
 namespace prosper::perf {
@@ -306,16 +312,18 @@ void InteractivePerformanceCapture::publish_completed(std::unique_ptr<PendingCap
     if (!out) {
         result.error = "could not open the reserved temporary capture for writing";
     } else {
-        const bool cpu_available = std::any_of(
-            capture->pre_samples.begin(), capture->pre_samples.end(),
-            [](const ProcessSample& sample) { return sample.process_cpu_ns.has_value(); }) ||
-            std::any_of(capture->post_samples.begin(), capture->post_samples.end(),
-            [](const ProcessSample& sample) { return sample.process_cpu_ns.has_value(); });
-        const bool rss_available = std::any_of(
-            capture->pre_samples.begin(), capture->pre_samples.end(),
-            [](const ProcessSample& sample) { return sample.rss_bytes.has_value(); }) ||
-            std::any_of(capture->post_samples.begin(), capture->post_samples.end(),
-            [](const ProcessSample& sample) { return sample.rss_bytes.has_value(); });
+        // One predicate for every optional process field, so a newly added field cannot get an
+        // availability flag whose "any sample carried it" rule quietly differs from its neighbours'.
+        const auto any_sample_has = [&capture](std::optional<uint64_t> ProcessSample::*field) {
+            const auto carried = [field](const ProcessSample& sample) {
+                return (sample.*field).has_value();
+            };
+            return std::any_of(capture->pre_samples.begin(), capture->pre_samples.end(), carried) ||
+                   std::any_of(capture->post_samples.begin(), capture->post_samples.end(), carried);
+        };
+        const bool cpu_available = any_sample_has(&ProcessSample::process_cpu_ns);
+        const bool rss_available = any_sample_has(&ProcessSample::rss_bytes);
+        const bool private_available = any_sample_has(&ProcessSample::private_bytes);
         out << std::setprecision(10);
         out << "{\"type\":\"header\",\"format\":\"prosper-performance-capture\","
                "\"version\":1,\"title_id\":" << json_string(capture->title_id)
@@ -328,7 +336,11 @@ void InteractivePerformanceCapture::publish_completed(std::unique_ptr<PendingCap
             << ",\"sample_interval_ns\":" << config_.sample_interval_ns
             << ",\"logical_cpus\":" << std::thread::hardware_concurrency()
             << ",\"process_cpu_available\":" << (cpu_available ? "true" : "false")
-            << ",\"rss_available\":" << (rss_available ? "true" : "false") << "}\n";
+            << ",\"rss_available\":" << (rss_available ? "true" : "false")
+            // Additive within format version 1: readers select every field by name, so a reader
+            // that predates this flag ignores it and a reader that expects it reads an older
+            // capture as "unavailable", which is exactly what an older capture knew.
+            << ",\"private_bytes_available\":" << (private_available ? "true" : "false") << "}\n";
 
         const auto write_sample = [&](const ProcessSample& sample, const char* phase) {
             out << "{\"type\":\"sample\",\"phase\":\"" << phase << "\",\"t_ns\":"
@@ -337,6 +349,8 @@ void InteractivePerformanceCapture::publish_completed(std::unique_ptr<PendingCap
             write_optional(out, sample.process_cpu_ns);
             out << ",\"rss_bytes\":";
             write_optional(out, sample.rss_bytes);
+            out << ",\"private_bytes\":";
+            write_optional(out, sample.private_bytes);
             out << ",\"guest_presents\":" << sample.guest_presents
                 << ",\"rendered_frames\":";
             write_optional(out, sample.rendered_frames);
@@ -568,6 +582,35 @@ ProcessSample collect_process_sample(uint64_t monotonic_ns, uint64_t guest_prese
         if (page_size > 0 && resident_pages <= UINT64_MAX / static_cast<uint64_t>(page_size))
             sample.rss_bytes = resident_pages * static_cast<uint64_t>(page_size);
     }
+    // Linux reports no committed-bytes counterpart to Windows' commit charge, so private_bytes stays
+    // empty here rather than being filled with a near-synonym of the resident set.
+#elif defined(_WIN32)
+    // Windows has no /proc/self/statm, which is why every .prperf taken here reported "RSS:
+    // unavailable on this platform/run" (#3448). GetProcessMemoryInfo carries both figures in one
+    // call: WorkingSetSize is the resident set -- the direct analogue of the Linux number -- and
+    // PrivateUsage is the commit charge, which is what actually grows with prosper's host-side
+    // caches. Both are recorded only when the query succeeds; a failed query leaves both empty so
+    // the header's availability flags stay truthful.
+    PROCESS_MEMORY_COUNTERS_EX counters{};
+    counters.cb = sizeof(counters);
+    if (::GetProcessMemoryInfo(::GetCurrentProcess(),
+                               reinterpret_cast<PROCESS_MEMORY_COUNTERS*>(&counters),
+                               sizeof(counters))) {
+        sample.rss_bytes = static_cast<uint64_t>(counters.WorkingSetSize);
+        sample.private_bytes = static_cast<uint64_t>(counters.PrivateUsage);
+    }
+#elif defined(__APPLE__)
+    // CONFIDENCE: LOW. Written from the documented mach interface. CI's macOS jobs compile it and
+    // run this file's test against it, but the change that added it was authored and measured on
+    // Windows and Linux only. MACH_TASK_BASIC_INFO is the 64-bit-safe flavor: TASK_BASIC_INFO's
+    // resident_size saturates at 4 GiB, which is below the working sets this capture is for.
+    // macOS's private footprint lives behind a different call (task_vm_info's phys_footprint), so
+    // private_bytes is left empty here rather than guessed at from a counter that does not mean it.
+    mach_task_basic_info_data_t info{};
+    mach_msg_type_number_t info_count = MACH_TASK_BASIC_INFO_COUNT;
+    if (::task_info(mach_task_self(), MACH_TASK_BASIC_INFO,
+                    reinterpret_cast<task_info_t>(&info), &info_count) == KERN_SUCCESS)
+        sample.rss_bytes = static_cast<uint64_t>(info.resident_size);
 #endif
     return sample;
 }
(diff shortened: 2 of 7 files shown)


mattias800 (commented):
APPROVED — independent review on exact head `c41fd62e`, I am not the author.

## The decision I most wanted to check: staying on format version 1

The PR keeps the capture at `version: 1` and adds `private_bytes` as a name-selected optional key.
That is the sort of choice that looks like avoiding work, so I read the reader rather than the
rationale:

```python
# tools/perf/performance_capture_report.py:40
if (header.get("type"), header.get("format"), header.get("version")) != (
        "header", "prosper-performance-capture", 1):
    raise CaptureError("not a supported prosper performance capture")
```

Equality, not `>=`. So a bump to 2 would make **every existing reader reject every new capture**,
and the new reader reject every archived version-1 capture — breaking in both directions at once, to
express a change that is purely additive. Keeping version 1 is not the lazy option here, it is the
only non-breaking one, and the PR body says so for the right reason. Confirmed.

## The honesty property the issue specifically asked for

`sample.rss_bytes` and `sample.private_bytes` are assigned **only inside the success branch** of
`GetProcessMemoryInfo`, so a failed query leaves both `std::optional` empty and `rss_available`
stays truthful. A zero never becomes indistinguishable from "not measured" — which was the issue's
explicit requirement and is the whole reason the field is optional rather than a plain integer.

The refactor of the availability flags into one `any_sample_has(&ProcessSample::field)` predicate is
a real improvement rather than tidying: it was already two hand-copied `any_of` pairs, and this PR
adds a third field, which is exactly when that pattern starts to drift.

## Per-platform correctness

- **Windows.** `PROCESS_MEMORY_COUNTERS_EX` with `cb = sizeof(counters)` and the documented
  `reinterpret_cast` to `PROCESS_MEMORY_COUNTERS*` plus the EX size — the correct calling
  convention for the extended struct. `psapi` is linked under `if(WIN32)`, which is needed because
  `GetProcessMemoryInfo` is not in the default kernel32 set.
- **Recording both figures separately is the substantive call**, and the issue argued for it:
  working set understated the number that matters by ~5 GB in the motivating GTA V run, because it
  is commit charge, not resident set, that tracks prosper's host-side caches. Overloading one field
  would have produced a memory axis that moves for the wrong reasons.
- **Linux** deliberately leaves `private_bytes` empty, with the reason stated — there is no direct
  commit-charge counterpart, and filling it with a near-synonym would make cross-platform
  comparisons silently wrong.
- **macOS** uses `MACH_TASK_BASIC_INFO` rather than `TASK_BASIC_INFO`, and the comment gives the
  reason: the latter's `resident_size` saturates at 4 GiB, which is *below* the working sets this
  capture exists to measure. That is the kind of detail that is normally discovered later, painfully.

## The unverified part, and why it does not block

The macOS arm is marked `CONFIDENCE: LOW` in the source and in the PR body: written from the
documented interface, not measured on a macOS host. I am satisfied it is safe to merge in that state
because the PR does not overclaim it — it says plainly which platforms were built and run, and CI's
Rosetta job compiles and executes `performance_capture` against it. What remains genuinely unproven
is only whether `resident_size` is the right counter to call RSS on that platform, and the arm is
separable: it can be removed without touching the Windows fix this issue is about.

Flagging it here rather than silently accepting it, since an unmeasured platform arm is the thing a
future reader is most likely to mistake for a verified one.

## Evidence beyond "it compiles"

Worth recording, because it is what lifts this above a plausible-looking patch:

- a real `.prperf` written by the live sampler on Windows tracking an 8 MiB-per-sample allocation
  (`RSS 20.5..44.5 MiB`, `private 16.9..40.9 MiB`);
- a real **pre-fix** Windows capture from 2026-08-07 (`rss_bytes: null`, no `private_bytes` key)
  still validating and reading cleanly — the backward-compatibility claim tested against an actual
  archived file rather than a synthesized one;
- both mutation arms red: the reader tests fail against `origin/main`'s reader with
  `KeyError: 'private_min'`, and the C++ test with only the `_WIN32` arm removed reports
  `rss=unavailable private=unavailable`, 4 failures, exit 1.

Confirmed independently with `prosper/tools/ci/pr_merge_gate.py`: all checks pass, none pending or
failing, PR head still matches the branch tip. The two local Windows ctest failures the author
reports (`session_start`, `build_revision_refresh`) are environmental — I hit `session_start` myself
on an unrelated branch and filed it as #3540 after confirming it against pristine `main`.

PR #2910, opened 2026-08-22
fix(savedata): sceSaveDataGetEventResult answered a drained queue with a code no guest matches — and Khazan (PPSA20447) brought up from nothing

mattias800:
Brings up **The First Berserker: Khazan** (`PPSA20447`, Unreal Engine 4, 48 GB) from nothing — no
tracker, no `COMPATIBILITY.md` row, no route, no prior work — and fixes the cross-title HLE defect
its boot died on.

Tracker [#2909](https://github.com/mattias800/prosper/issues/2909). New blocker
[#2908](https://github.com/mattias800/prosper/issues/2908).

## Rung reached: 0, and the row says so

The title boots, links all 53 modules (including 29 Wwise plug-in PRXes), streams IoStore assets
through APR with no failed read, brings up the live Vulkan compute backend, mounts its profile save
and enumerates all thirty save slots. **The only thing it ever composites is a flat white 3840x2160
clear**, which is not content, so it does not reach rung 1 and **no screenshot is checked in**.
`COMPATIBILITY.md` and the tracker both say rung 0.

## The failure scenario

The game thread parked **forever** about four seconds into every boot. `tools/screenshot` reported
`1 distinct frame(s) in 33 publications` over a **360 s** run with `guest=running`, and a `gdb` dump
of the live process showed **all 79 threads blocked in a syscall** — the guest main thread inside
`prosper::k_usleep`, and every engine thread (`RenderThread 0`, `RHIThread`, `FAPREventQueueL`,
42 × `TaskGraphThread`, 19 × `PoolThread`) on a condition variable or event flag. Nothing spinning,
nothing progressing.

With `PROSPER_SVCLOG=1`, exactly one Sony call repeated during the stall — `sceSaveDataGetEventResult`,
**5,020 times in 90 s**, with no other guest thread making any call at all. When every other thread is
idle, the only thing that can change the condition is the answer to that call.

Scanning the poll's own stack frame for guest return addresses (the technique `hle_file.cpp`'s
`[apr] guest-ra[N]` diagnostic already uses) gave a chain that was **identical on three separate
polls**, and disassembling its first frame produced the wait loop verbatim:

```text
eboot+0x796eb30   vmovss xmm0,[rip+0x3aea72c]     ; the sleep interval
eboot+0x796eb38   call   0x1565790                ; FPlatformProcess::Sleep(float)
eboot+0x796eb41.. vmovups ...                     ; zero a 104-byte SceSaveDataEvent
[...]

PR #3536, opened 2026-09-11
ci: prove a PR body was published rather than trusting gh exit codes (#2918)

### LINKED ISSUE #2918: gh pr edit --body-file can fail silently (rc=1, GraphQL projects deprecation) leaving a stale PR body
## What happened

While correcting PR #2910's body, `gh pr edit <N> --body-file <file>` **returned rc=1 with a GraphQL
projects-deprecation error and did not apply the edit**. The lane believed the body was updated and
carried on; it only caught the failure by grepping the live body back afterwards.

Reported by the lane that hit it. I have not reproduced it independently, so treat the exact trigger
as unconfirmed — what is confirmed is that the body sat stale while the author believed otherwise.

## Why it is worth recording rather than shrugging at

This is the silent-instrument shape, and it lands in the worst place: a PR body is where
reviewers, the merge step and every later reader take the author's account of the change. A stale
body after a correction is not a cosmetic problem — on #2910 the correction *was* the deliverable
(four blocking review findings, all prose), so a silently-unapplied edit would have meant merging a
PR whose stated reasoning had been rejected.

It also fails in the direction nobody checks. `gh pr edit` succeeding is assumed; almost nobody
greps the body back.

## The workaround that does work

```bash
gh api -X PATCH repos/OWNER/REPO/pulls/<N> -F body=@<file>
```

REST rather than GraphQL, and it applied cleanly on the same body that `gh pr edit` had refused.

## Suggested

1. **Always read the body back** after editing it programmatically, the same way we verify a remote
   branch really went away after `--delete-branch` rather than trusting the flag.
2. Add a numbered row to `GAME_COMPAT_ORCHESTRATION.md`'s instrument table once the trap numbers in
   flight (#2910 claims 217/218, #2917 claims 219) have landed — allocate with
   `python3 prosper/tools/docs/trap_number.py`, which scans master **and** open PRs.
3. If anyone can reproduce it deterministically, worth pinning which `gh` subcommands route through
   GraphQL and are therefore exposed. `gh pr comment` and `gh pr review` have both worked all
   session, so it is not all of them.

mattias800:
Fixes #2918.

## Failure scenario

`gh pr edit <N> --body-file <f>` was observed returning **rc=1** with a
`GraphQL: Projects (classic) is being deprecated ...` error and **not applying the edit**. The lane
carried on believing the body was updated, and only caught it by grepping the live body back. A
second lane reproduced it on 2026-08-23 editing #2965's body; the REST endpoint applied the same
bytes first try.

A PR body is where reviewers, the merge step and every later reader take the author's account of
the change, and it fails in the direction nobody checks — `gh pr edit` succeeding is assumed. On
#2910 the corrected body *was* the deliverable (four blocking review findings, all prose), so a
silently-unapplied edit would have meant merging a PR whose stated reasoning had been rejected.

## First: it no longer reproduces, and the reason is why the fix is a read-back

Measured today on this machine, every `rc` read from a log rather than from a `$?` through a
wrapper:

| probe | rc | outcome |
| --- | --- | --- |
| `gh pr edit 3534 --body-file <f>` | 0 | applied; confirmed by reading the body back |
| `gh pr view 3534 --comments` (trap 240) | 0 | output present |
| `gh pr view 3534` / `gh pr status` | 0 | no deprecation text on stderr at all |

**The `gh` binary has not changed** — 2.93.0, released 2026-05-27, i.e. three months *before* the
2026-08-23 reproduction. So what moved is server-side, not a version we bumped. That cuts both
ways: nobody here can pin the trigger to anything under this repository's control, and it can
return the same way it left. Which is the argument for **verifying the state** rather than for
calling the issue environmental and closing it — the issue's own suggestion 1 ("always read the
body back") is the part that survives the trigger going away.

Recorded where the next agent looks: trap 240's row in `GAME_COMPAT_ORCHESTRATION.md` gains a dated
re-measurement. The row itself is left as the record of what was true.

## Approach

New `prosper/tools/ci/pr_body.py`:

```
pr_body.py set    <pr> (--body-file F | --body TEXT)   # write, then PROVE it took
pr_body.py verify <pr> (--body-file F | --body TEXT)   # prove it matches, without writing
pr_body.py get    <pr> [--out F]
```

- **The write is REST** — `gh api -X PATCH repos/<canonical>/pulls/<N> --input -`, with the
  payload as JSON on stdin. GraphQL is where the error came from; `--input -` rather than
  `-F body=@file` so no field parsing can reinterpret a body that happens to look like a number or
  a bool; and `<canonical>` comes from `gh repo view --json nameWithOwner` rather than from `gh`'s
  `{owner}/{repo}` placeholders, for the reason in the third finding below.
- **`set` always reads the body back and compares**, whatever the write reported. The verdict is a
  statement about the live body, never about a subprocess.

## Contract

| state | exit | note |
| --- | --- | --- |
| live body matches intended | 0 | including when the write exited non-zero — the edit applied |
| live body differs | 1 | including when the write exited 0 — this is the #2918 shape |
| could not reach a verdict | 2 | `gh` missing, network, unreadable file |

Both directions are load-bearing. Exiting 1 on a non-zero write that nonetheless applied would make
the tool cry wolf on exactly the GraphQL noise this is about, and a gate people learn to pass with
`|| true` protects nothing. 2 is distinct from 1 for the same reason `pr_merge_gate.py` separates
them: "could not check" must never read as "checked, and it is fine".

Comparison folds `\r\n` / `\r` and trailing blank lines (GitHub normalises to CRLF on some paths)
and **nothing else** — a trailing space inside a line is a Markdown hard break, so folding it would
bless a body that renders differently from the file.

## Invariants / deliberately unaffected

Nothing existing changes behaviour. This adds two files, one ctest registration, an `AGENTS.md`
section and one dated sentence on an existing trap row. No other tool, workflow or script is
touched, and nothing is rewired to call `pr_body.py` — using it is a choice, not a new gate.

## Regression arms — and one the tests could not have found

`prosper/tools/ci/test_pr_body.py`, registered as ctest `pr_body_readback`. Hermetic: `pr_body.run`
is replaced by a fake whose **exit code and whether the write actually applies are independent
knobs**, which is what lets the reported shape be reached on demand. Platform-independent; nothing
reads `/proc`.

It leads with a positive control (a tool answering MISMATCH to everything would satisfy every
failure arm), and each failure arm asserts the stated reason rather than the boolean. Arms cover:
the rc=0 / not-applied defect, the rc!=0 / applied mirror, an empty live body, a `null` body read as
empty rather than as the string `"null"`, an unreachable API as 2 rather than 1, `verify` issuing no
PATCH, the write going over REST and never through `gh pr edit`, a body that is a strict prefix of
the live one, and the exact set of whitespace differences that are and are not folded.

**Then a live smoke test against a real PR found a defect no fake could reach.**
`subprocess(text=True)` decodes with the **locale** encoding — cp1252 on a Windows host — so `gh`'s
UTF-8 came back as mojibake and an em dash read as three Latin-1 characters. `verify` then reported
BODY MISMATCH on a body that was byte-for-byte correct, on a repository whose PR bodies use em
dashes throughout: a false alarm from the one instrument whose entire job is to say whether the live
text is right. Fixed by decoding UTF-8 explicitly, and pinned by an arm that drives a **real
subprocess** rather than a fake, since the bug lives in the one function every fake replaces.

**And then it found a second one.** `get` against a real PR raised `UnicodeEncodeError` printing
a body containing U+2260 to a cp1252 stdout, and `main`'s catch-all reported that as
COULD NOT VERIFY -- the tool answering "I cannot tell" purely because it could not *print* the
answer. The two output paths want opposite treatment, so they are now two functions: `emit`
degrades unencodable characters (a report's payload is the verdict and the line number, and a
substituted glyph changes neither), while `write_body_out` stays byte-exact through the binary
buffer, because `get <N> > file` must round-trip -- a substituted character there would corrupt
the body and make a later `verify` report a mismatch the PR does not have. Its arm builds a real
cp1252 text stream, so it reproduces on every platform rather than only on a Windows host.

**And a third, which the tool caught on its own first live `set` — the design working, not a
coincidence.** `gh api` expands `{owner}/{repo}` from the git remote. This checkout's remote is
`.../ps5ys.git` while the repository's canonical name is `mattias800/prosper`, and GitHub answers a
renamed repository with **HTTP 307**, which `gh api` does not follow on a `PATCH`. So the first real
`pr_body.py set` reported `gh: HTTP 307` and **changed nothing** — and the read-back said so:

```
PR #3536: BODY MISMATCH
  NOTE: the write exited 1 (gh: HTTP 307) -- checking the live body anyway, because that exit
  code is precisely what #2918 showed cannot be trusted in either direction
  line 99 differs: ...
```

Had the verdict been the write's exit code, or had the tool been built on `gh pr edit` (which
resolves the name itself), the body would have sat stale. That is the #2918 failure, reproduced by
the fix for #2918, on its first use. Endpoints now come from `gh repo view --json nameWithOwner`;
an unresolvable repository raises rather than falling back to the placeholders, because without a
name there is no endpoint to ask and the honest answer is 2, not 1.

Worth stating plainly: **the hermetic suite was green through all three defects.** Only running the
tool against something real caught any of them — twice by smoke test, once by the tool's own
verdict. That is also why the endpoint invariant lives in the FakeGh fixture rather than in a single
arm: it asserts no call reaching it still carries `{owner}`, so reverting the resolution reddens the
suite at its first API call with the cause in the message.

## Without-fix verification (both mutations reverted locally, then restored)

| mutation | result |
| --- | --- |
| `set_and_verify` trusts the write's rc instead of reading back | **red: 8 failed checks**, incl. `rc=0 + not applied -> 1` and `nothing reassuring went to stdout`; the positive control stayed green, so the suite was not merely stuck at "no" |
| `run()` back to `subprocess(..., text=True)` | **red: 2 failed checks** — `UTF-8 output survives the host locale`, `and is not mojibake` |
| both output paths back to a bare `stream.write` | **red: 4 named failures**, including the exact `'charmap' codec can't encode character` the live run produced |
| `_api` back to the `{owner}/{repo}` placeholders | **red at the first API call**: `AssertionError: endpoint still uses {owner}/{repo}` from the fixture invariant |

## Risks

Low. New, opt-in tooling; no existing path calls it. The one judgement call is folding line endings
in the comparison — deliberate, argued above, and pinned by arms on both sides of the line.

## Commands and results

```
# the suite, natively on the Windows host (no wsl wrapper, real exit code)
python prosper/tools/ci/test_pr_body.py
  -> exit 0, 50 `ok` checks, 0 FAIL, "all pr_body checks passed"

# via ctest, to prove the registration works
ctest --test-dir prosper/build-linux --no-tests=error \
      -R "pr_body_readback|pr_merge_gate" --output-on-failure
  -> 2/2 passed, 0 failed, exit 0   (pr_body_readback, pr_merge_gate)
ctest --test-dir prosper/build-linux --no-tests=error -R pr_body_readback --output-on-failure
  -> 1/1 passed, 0 failed, exit 0   (on the final head, 0.17 s)

# repo self-checks over the new file
python prosper/tools/ci/check_usage_text.py --root prosper/tools
  -> 48 modules read __doc__, 0 violations
python prosper/tools/ci/check_ctest_gate.py
  -> 12 ctest invocations, all carrying --no-tests=error

# docs gate (a numbered-table row was edited)
python prosper/tools/docs/check_numbered_table.py --ordered --table-header Instrument \
    --baseline <origin/main copy> prosper/docs/GAME_COMPAT_ORCHESTRATION.md
  -> 14 tables, 336 rows; numbered column 1..274, unique and ascending;
     none of the 274 base rows deleted
git ls-files '*.md' -z | xargs -0 python prosper/tools/docs/check_numbered_table.py
  -> all clean

# live end-to-end against a real PR (read-only arm plus a deliberate-mismatch control)
python prosper/tools/ci/pr_body.py verify 3534 --body-file <the live body>
  -> "PR #3534: BODY VERIFIED / live body matches the intended body (5493 chars)", exit 0
python prosper/tools/ci/pr_body.py verify 3534 --body-file <a wrong body>
  -> "BODY MISMATCH / line 1 differs", exit 1
python prosper/tools/ci/pr_body.py get 3534 > <file>
  -> 5510 bytes, U+2260 present as UTF-8 (this call is what surfaced the second defect)

# and this PR's own body was published with the tool, which is its end-to-end arm --
# including the run that legitimately REFUSED, before the redirect fix
python prosper/tools/ci/pr_body.py set 3536 --body-file <this text>
  -> before the fix: "BODY MISMATCH", write rc=1 (HTTP 307), body unchanged -- exit 1
  -> after  the fix: "BODY VERIFIED / live body matches the intended body", exit 0
```

No C++ was rebuilt: the diff is two new Python files, a CMake `add_test`, and two Markdown edits.
CMake was re-configured so the new test registers, and the ctest run above is that registration
executing.


mattias800 (commented):
Independent review (I am not the original author). **One blocking finding, which I have corrected on
this branch rather than handing back — the authoring agent had already exited.** New head
`7164e021`. With that correction: **APPROVED**, merge gated on CI re-running against the new head.

## Blocking finding: the 307 mechanism was stated as established, and it does not reproduce

`resolve_repo()`'s docstring asserted, and the test repeated twice, that GitHub answers a renamed
repository's `PATCH` with an HTTP 307 which `gh api` declines to follow — and that this is why the
tool's first live `set` wrote nothing. It was introduced with *"That is not hypothetical and was not
predicted"*, which is the wording of a measured fact.

It was inferred from a single observed `gh: HTTP 307`. I could not reproduce it in any form. Same
machine, same `gh` 2.93.0, same checkout (remote `ps5ys`, canonical `mattias800/prosper`):

| arm | result |
| --- | --- |
| `PATCH repos/{owner}/{repo}/issues/comments/<id>` | rc=0 |
| `PATCH repos/{owner}/{repo}/pulls/3539` | rc=0 |
| `PATCH repos/mattias800/ps5ys/pulls/3539` — the **renamed-away** spelling | rc=0 |

And since rc=0 is not evidence of a write — which is this PR's entire subject — the last arm was run
with a detectable payload rather than identical content:

```
PATCH repos/mattias800/ps5ys/pulls/3539   body = <original> + "<!-- redirect-probe-123804 -->"
GET   repos/mattias800/prosper/pulls/3539 grep -c redirect-probe -> 1    # the write LANDED
PATCH repos/mattias800/prosper/pulls/3539 body = <original>              # restored -> 0
```

**If the rename were the cause, that arm could not have written.**

Why this was worth blocking on rather than waving through: a mechanism written into a docstring is
inherited by every later reader, and this one arrived with a narrative, a repo name, an error string
and a "caught on first live use" provenance. That is the profile `CLAUDE.md`'s authority-transfer
section describes — the most contagious kind of wrong claim this project produces — and the same
sentence had already been published as a generalized warning on #2918 (*"anything scripting
`gh api ... -X PATCH` against `{owner}/{repo}` in this repository is exposed"*), where I have posted
the same correction.

I am **not** disputing that `gh: HTTP 307` was seen. The write failed; the cause is unidentified.
The comments now say exactly that.

## What I changed, and what I deliberately did not

Comments only — no behaviour change, suite unchanged and still green (`all pr_body checks passed`).

- `resolve_repo()`'s docstring: the rename is now given as a reason to resolve the name *on its own
  merits* (one place decides which repository we address), with the single observation recorded as
  a single observation and the arms above cited.
- The test's two comments: same correction. **The invariant itself is kept unchanged** — no call may
  carry `{owner}/{repo}` — because it is a sound design property independent of how `gh` follows a
  rename, and it is what makes reverting `resolve_repo()` redden every arm instead of one.

## Why the PR still stands, and stands better

The premise of the fix never depended on the mechanism. The read-back defence — write, then read the
live body and take the verdict from *that* rather than from an exit code — is correct whatever
caused the failure it caught, which is precisely why it caught something nobody had predicted. **A
verification that only defends against the failure you already understand is not verification.**
That argument is now the one the code makes.

The rest I checked and agree with:

- **Finding 1 is the right call.** Neither original symptom reproduces, but `gh` 2.93.0 was released
  2026-05-27 — three months *before* the 2026-08-23 reproduction — so nothing under this repository's
  control changed, the trigger moved server-side, and it can return. That argues for implementing the
  issue's suggestion 1 rather than closing it as environmental. Trap 240 gets a dated re-measurement
  and the row is left as the record of what was true, which is the charter's rule.
- **The three defects the author hit are the valuable part of this PR**, and all three were invisible
  to a green hermetic suite — `text=True` decoding `gh`'s UTF-8 through the locale codepage,
  `UnicodeEncodeError` on a cp1252 console reported as COULD NOT VERIFY, and the live write failure
  itself. Only running the tool for real surfaced any of them.
- Both without-fix arms are reported red, including a fixture invariant that reddens at the first API
  call.

Issue #2223, opened 2026-08-07
ArcRunner: hold the chunk-label recycler for the fold — the visibility contract rescues it but regresses Sonic Frontiers


mattias800:
ArcRunner (PPSA21406, SDK 10) corrupts the guest's command-chunk label recycler and faults 3 of 3 on a default route. The mechanism is **measured and localised**; what does not exist is a fix with an acceptable blast radius.

## The mechanism (established, #2219)

A per-fold census (`PROSPER_FOLD_MARGIN`) shows the guest's builder thread rebuilding its **46 command-chunk labels while prosper is inside the fold consuming the previous generation**, beating prosper to 16–21 of them.

The cause: **prosper applies a submit's completion writes while still executing the rest of the same command buffer**, which releases the guest's recycler mid-fold. The race is *inside one fold*, not between frames — which is why every frame-granularity instrument missed it for months.

Supporting numbers: `work` is 64–67 ms per submit against a guest gap of 5–50 µs, so prosper's fold is the pacer and the guest is inside prosper ~84% of wall time.

## Why the obvious fix is not available (#2220, #2222 — both closed off)

prosper's post-submit visibility contract prevents exactly this, and it is gated `if (version >= 13)`. ArcRunner asks for SDK 10. Removing the gate **does** rescue it — and regresses *Sonic Frontiers*, which reached rung 2 on 2026-08-07:

| level | ArcRunner | Sonic Frontiers |
| --- | --- | --- |
| 0 — neither (master, sub-13) | faults 3/3 | renders, **0 faults / 12** |
| 1 — submit-scope deferral only | **renders nothing, 4/4** | **renders nothing**, 0 ok / 3 FAILED |
| 2 — deferral + selective drain | renders (377 / 374) | renders, **3 faults / 12** (all identical: frame 8, four `[nullpage]` lines) |

Level 1 is strictly worse than either neighbour for both titles, so **the contract is not separable** — deferring completion writes while draining indiscriminately produces neither behaviour coherently. Splitting the predicate was tried and falsified.

Also measured: Oregon Trail (SDK 12) is **unaffected** either way, so this is not a general property of pre-13 titles.

## What is actually needed

A fix targeting **the mid-fold release specifically** rather than whole-submit visibility. The contract works for ArcRunner as a side effect of a much broader behaviour, and that breadth is what drags Frontiers along.

[...]

PR #2219, opened 2026-08-07
ArcRunner: the per-fold account — the guest's builder is released mid-fold, and the contract that forbids it is version-gated off

mattias800:
## What this is

arc6 closed with one thing owed: *"the next instrument owes a **per-fold** account of what 1.5 ms of
wall time per submit changes — not another whole-run rate, which has now produced two retracted rows
on this title."* This is that account. It ends with the race **named**, and with a candidate
correctness change that takes ArcRunner to its **title screen on the default route with no throttle**.

Refs #1226, refs #1817, refs #1894, refs #2084, refs #2217.

## The failure scenario

ArcRunner (`PPSA21406`, UE4 4.27) faults in the guest's `FMallocBinned3` allocator within 8–14 s of a
default launch, 17 of 17 runs historically and 3 of 3 here. The corrupt qword is prosper's own 4-byte
immediate-zero label init followed by its paired value-1 fence, landing in a 0x20-byte block the guest
has already recycled. `PROSPER_SUBMIT_STALL_US=1500` removes the fault entirely, and the previous six
sessions could not say why: it is not the submit mutex, not the guest's rate, and not our write
latency.

## What the per-fold census found

**1. prosper's fold is the pacer.** `work` = 64–67 ms per submit, and the guest's `gap` between
prosper's return and its next submit entry is **5–50 µs**. The guest sits inside prosper ~84% of wall
time; a 1500 µs stall is 1.7% of one fold.

**2. The exposure window is LONGER under the throttle, not shorter.** With the build journal now
recording the fold a packet was built in, build→exec distance in folds is mode **5** default (n = 506)
against mode **6** throttled (n = 7,854), and longer in milliseconds too. This falsifies for the third
time — and for the first time on a population both arms share — the reading that a fix has to restore
"exec at f+5". f+6 is the *surviving* arm's mode.

**3. `REBUILD-BEFORE-EXEC` is the discriminator**: 16 / 31 / 14 in three default arms, **0 / 0 / 0** in
three throttled ones over an 18× larger population. It agrees exactly with `SUSPECT-REL1-OVERLAP`
(fence side) and DMA-INIT-GEN's `depth>=2` (init side) in every arm — three points in one protocol.

**4. It is a single-fold burst, and the per-fold trace shows the mechanism.** Every event of a run
lands in one fold, at one millisecond, across 13–21 different labels of one chunk. Reproduced 3 of 3
either way:

```text
default (dies)                              throttled (survives)
[...]

Issue #2217, opened 2026-08-07
tools/screenshot and boot_trace disagree on ArcRunner under PROSPER_POST_SUBMIT_VISIBILITY: 0/3 vs 2/2 faults on the same head


mattias800:
ArcRunner (`PPSA21406`), Linux, master `cdb40942` plus the arc7 branch.

With `PROSPER_POST_SUBMIT_VISIBILITY=1` (the new #1226 A/B lever, default OFF) and **no** submit
throttle, the two frontends disagree completely on the same head, same dump, same route:

| frontend | runs | outcome | delivered video frames |
| --- | --- | --- | --- |
| `boot_trace` | 3 | survived a 40 s bound, zero faults, zero fatals | 788 / 782 / 838 |
| `tools/screenshot` | 2 | **faulted**, exit 90 | 35 / 42 |

Both `screenshot` runs die the same way:

```
[prosper] WORKER-THREAD FAULT: sig=11 addr=(nil) rip=0x5c00048e0 (image+0x1b00048e0)
[fault] thread='AgcCleanupThrea' ... on-guest-TCB=yes
```

`rip` is not in the eboot's usual range and the frame pointer chain runs through
`eboot+0x11e67f4 -> +0x13e29af -> +0x1276fa0 -> +0x11720c5`.

The obvious explanation is wrong. Both runs died at 35 and 42 delivered video frames, i.e. **before
the first sample interval elapsed** (`--seconds 8` and `--seconds 20`), so the 4K readback had not
run yet and cannot be what killed them. Something else about the `screenshot` frontend's setup
differs from `boot_trace`'s on this route.

Why it matters beyond this lever: `tools/screenshot` is the frontend the project uses for
progression evidence, and `ARCRUNNER_STATUS.md`'s arc6 title-screen capture was taken with it. If a
change can survive under one frontend and die under the other, then "does this fix the title?" is
frontend-dependent, and any arm that quotes one frontend has to say which.

Repro:

```bash
# survives 3/3
PROSPER_GUEST_ARGS= PROSPER_RENDER=1 PROSPER_AVPLOG=1 PROSPER_POST_SUBMIT_VISIBILITY=1 \
  timeout 40 ./build-linux/boot_trace <DUMP_ROOT>/PPSA21406-app0

# faults 2/2 at ~35-42 video frames
[...]

PR #3538, opened 2026-09-11
env: PROSPER_WAITCAP must not remove its own cap, and a typo must not pick the lever's OFF arm

### LINKED ISSUE #3304: PROSPER_WAITCAP silently removes its own cap on '-1', and a typo in PROSPER_POST_SUBMIT_VISIBILITY forces the experiment's OFF arm

mattias800:
Fixes #3304. Refs #3267.

Two `PROSPER_*` reads where a malformed value does something **worse than nothing** — one removes a
bound, one silently selects an arm of a live experiment. Neither appears in #3267's published
catalogue, so #3294's bucket-A sweep did not see them. Neither is a regression; both predate that
work.

## 1. `PROSPER_WAITCAP` silently removed the cap it was set to arm

`src/hle/kernel/hle_kernel_time.cpp`, bounding a **timed** equeue wait:

```cpp
const char* e = getenv("PROSPER_WAITCAP"); return e ? (uint64_t)atoll(e) : 0;
...
if (cap && us > cap) us = cap;
```

`atoll("-1")` is `-1`; cast to `uint64_t` it is `UINT64_MAX`, which is **non-zero**, so it passes the
`cap &&` guard — and then `us > UINT64_MAX` can never be true. The clamp never fires. An operator who
armed a wait cap got no wait cap, and nothing said so. `=5ms` was a **5 microsecond** cap and `=fast`
was 0, i.e. off while looking set.

Now `env_u64_or_default("PROSPER_WAITCAP", …, 0, "us", "0 = no cap: …")`. The `default_note` matters
here for the same reason it did on `PROSPER_MAX_DISPATCH_GROUPS`: this knob's default is itself the
permissive sentinel, so telling someone who typed a value *in order to impose a bound* that we are
"keeping the default (0)" is true and useless. The line says they still have no cap.

## 2. `PROSPER_POST_SUBMIT_VISIBILITY`: a typo did not mean "unset", it meant **forced off**

`src/gpu/pm4/command_processor.cpp`:

```cpp
const char* e = getenv("PROSPER_POST_SUBMIT_VISIBILITY");
const int v = e ? (int)strtol(e, nullptr, 0) : -1;
```

The tri-state is deliberate and correct — `-1` unset (follow the SDK version), `1` forced on, `0`
forced off. The **parse** collapsed it. `strtol` returns 0 for anything it cannot read, so `=on`,
`=true`, `=yes`, `=enabled` — and an empty value — all landed on **FORCED OFF** and printed
`[agc] POST-SUBMIT-VISIBILITY FORCED OFF (#1226 A/B)` as though it had been asked for.

This is the reason this PR exists rather than being folded into the bucket-B sweep. The knob is a
**live experiment lever on an open question**: `CLAUDE.md` and `ARCRUNNER_STATUS.md` record its
verdict as undecided with the frontend as the variable (#2217), the race it probes is named in #2219,
and #2223 is the *Sonic Frontiers* regression that keeps it from becoming the default. An A/B run
spelled `=on` would have produced a confidently-labelled "lever on" result that was the **forced-off
arm**, with a log line asserting deliberateness. That is the apparatus masquerading as the subject,
on exactly the kind of question where it costs the most.

Measured while writing the arms, and not in the issue: `=2` was worse again. `strtol` gave 2, the
site's `forced >= 0` accepted it and `forced != 0` forced the model **ON** — and neither the `v == 1`
nor the `v == 0` branch printed, so that arm ran completely silently.

## Approach

`PROSPER_WAITCAP` uses the existing `env_u64_or_default`. The tri-state cannot: no numeric fallback
expresses "unset", and the point is refusing to guess. So this adds one function to
`src/diagnostics/env_numeric.hpp`, in the style of the `_auto` family and `env_u64_or_report` that
are already there:

```cpp
int env_tristate_or_unset(const char* name, const char* text);   // -1 unset, 0 off, 1 on
```

Accepted: `1` / `on` / `true` / `yes` / `enabled` → 1, and `0` / `off` / `false` / `no` / `disabled`
→ 0, in either case. Also `0x1` and `0x0`, because this site read **base 0** and refusing a spelling
that used to work would be the header changing a well-formed setting — the same N1 rule the `_auto`
family exists for. **Anything else**, including a number that is neither 0 nor 1, is **unset** with a
line on stderr naming the value and listing what is accepted. Never a silent third answer.

Accepting the word spellings is an addition beyond what the issue asked for; it is documented in the
header, in `src/diagnostics/AGENTS.md`, and in the refusal message itself. The argument for it is
that `=on` is what an operator types, and "refuse it" and "understand it" both remove the hazard,
while only the second removes the failed run. Everything already published uses `=1`
(`ARCRUNNER_STATUS.md`, `CRISIS_CORE_STATUS.md`, `COMPATIBILITY.md`), which is unchanged.

Both gates were updated, because both can be silently defeated by a new helper:

- `tools/env/check_env_numeric_arms.py` — its `CALL_RE` now matches `env_tristate_or_unset`, so a
  site parsed through it with no arm fails the gate. (Verified: before the arms were written, the
  gate failed naming both new knobs.)
- `tools/env/check_cached_env.py` — `ENV_READERS` gains `env_tristate_or_unset` **and** the three
  older readers that were never listed (`env_u64_or_default_auto`, `_auto_capped`,
  `env_u64_or_report`). That script treats any callee containing `env` as an environment *write*, so
  an unlisted reader reads as an arming; `src/diagnostics/AGENTS.md` already documents the coupling.

## Regression arms

`tests/diagnostics/test_env_numeric_sites.cpp`, following that file's established shape — each arm
**mirrors** its site rather than calling it, because both sites cache in a function-local `static`
and a test that armed the variable with `setenv` would go vacuous rather than red (#2214). Every
table row carries the file's discriminator leg: the old spelling must answer *differently*, so an arm
cannot pass vacuously on a site that never had the defect.

- `PROSPER_WAITCAP`: `-1` (the removed cap) and `5ms` (the 5 us cap) both keep 0; `5000` and `250`
  still select what was asked.
- `PROSPER_POST_SUBMIT_VISIBILITY`: `2` and `enable` keep **unset**; `1` and `0` still force.
- In `main()`, the spelling set stated directly: every word in
  `{on, ON, On, true, TRUE, yes, enabled}` selects the ON arm **and every one of them used to select
  FORCED OFF**; `{off, OFF, false, no, disabled}` select OFF; `1`, `0`, `0x1`, `0x0` still work; unset
  and empty stay unset.

**Verified red without the fix.** Both mirrors were temporarily replaced with the pre-fix arithmetic
verbatim, rebuilt and re-run: **6 failures, exit 1** —

```
[FAIL] hle_kernel_time.cpp PROSPER_WAITCAP (cap removed): '-1' keeps the default
[FAIL] hle_kernel_time.cpp PROSPER_WAITCAP (unit suffix): '5ms' keeps the default
[FAIL] command_processor.cpp PROSPER_POST_SUBMIT_VISIBILITY (neither 0 nor 1): '2' keeps the default
[FAIL] command_processor.cpp PROSPER_POST_SUBMIT_VISIBILITY (near-miss word): 'enable' keeps the default
[FAIL] PROSPER_POST_SUBMIT_VISIBILITY: on/true/yes/enabled select the ON arm
[FAIL] PROSPER_POST_SUBMIT_VISIBILITY: unset and empty stay UNSET, and follow the SDK version
== FAILURES: 6 ==
```

(The last of those is the empty-value case — `PROSPER_POST_SUBMIT_VISIBILITY=` also forced OFF.) The
control was then reverted and the suite re-run green.

## Unaffected

- `=1`, `=0`, `=0x1`, `=0x0` and unset behave exactly as before, so every published route and every
  recorded A/B in `ARCRUNNER_STATUS.md` is unchanged. No doc needed rewriting.
- The `POST-SUBMIT-VISIBILITY FORCED ON/OFF` announcements, the SDK-version default, and
  `g_post_submit_visibility` are untouched — only the parse changed.
- `PROSPER_WAITCAP` unset still means no cap; a well-formed cap still clamps identically.
- No behavioural change on any path where these variables are not set.

## Risks

- `PROSPER_POST_SUBMIT_VISIBILITY=2` (and any other number outside {0, 1}) now means **unset**, where
  it used to silently force the model on. That is the intended correction — nothing documents such a
  spelling, and it printed nothing, so no recorded run can have relied on it — but it is a behaviour
  change and worth a reviewer's eye.
- `env_tristate_or_unset` is a new entry point in a header several lanes are converting sites through
  (#3267). Its grammar is deliberately narrower than `strtol`'s and wider than the decimal helper's;
  if that split is wrong, it is wrong for every future tri-state that copies it.

## Build and test

Environment note, stated because it bounds what these results can claim: this WSL image ships
`libvulkan-dev` **1.3.275**, and `frontends/shared/device/vulkan_runtime.hpp:11` needs
`VK_API_VERSION_1_4` since #3418. So `prosper_live_renderer`, `gpu_replay`, `prosper-app` and the
Vulkan-execution tests **cannot compile here on current `main` either** — environmental, and
unrelated to this diff, which touches none of those files. CI covers that chain.

```bash
TMPDIR=<WORKTREE>/prosper/build-linux/tmpdir cmake --build prosper/build-linux -j5 \
    --target prosper_core test_env_numeric_sites          # exit 0
./prosper/build-linux/test_env_numeric_sites              # exit 0, 136 assertions, 0 failures
python3 prosper/tools/env/check_env_numeric_arms.py prosper   # exit 0: 36 knobs parsed, every one armed
python3 prosper/tools/env/check_cached_env.py prosper         # exit 0: all checks passed
```

`prosper_core` is the target that compiles both changed sites
(`src/hle/kernel/hle_kernel_time.cpp`, `src/gpu/pm4/command_processor.cpp`), so both are built and
warning-free.

Full suite, after a keep-going build (`cmake --build . -- -k`) so everything that *can* link here does:

```
ctest --no-tests=error -j4 --timeout 600
76% tests passed, 112 tests failed out of 458      # exit 8
```

**458 tests discovered, 112 failed, 1 skipped** (`refactor_survey_measurement`), the rest passed.
Every one of the 112 was traced to this environment rather than to the diff, and the three causes are:

- **110** are the Vulkan chain: `prosper_live_renderer` and the 27 Vulkan-execution test targets fail
  to *link* here, so ctest reports them `Not Run` or `Failed` with `No such file or directory`
  (checked on `storage_readonly_control`). Cause is `VK_API_VERSION_1_4` vs this image's
  `libvulkan-dev` 1.3.275, above.
- **`profiling_access`** — `AssertionError: 511 != 493`, i.e. mode `0777` where it expects `0755`.
  The worktree is on `/mnt/c` (drvfs), which reports every file as 0777.
- **`snapshot_guard_logic`** — a byte-for-byte manifest round-trip, failing `b'{\r\n …' != b'{\n …'`:
  the checkout is CRLF because the host is Windows.

None of the three can be affected by this diff, and CI covers all of them on a Linux runner. Stated
in full rather than as "unrelated failures" because a 76% line needs its cause named, not asserted.

## #3267

This converts **2** of the 169 call sites that issue catalogues — and neither was in it, which is the
more useful fact: the catalogue is a census of a *shape* (`strtoull`/`atoi` with the end pointer
discarded) and both of these are that shape with a bound or an experiment arm behind it. **167
remain** by the issue's count, plus whatever else the catalogue does not list. A comment on #3267
records what this PR took.

### DIFF
--- a/prosper/src/diagnostics/AGENTS.md
+++ b/prosper/src/diagnostics/AGENTS.md
@@ -74,6 +74,15 @@ a typo silently ran a different experiment than the one asked for (#3253, after 
 measurement). `env_u64_or_default*` refuses loudly and keeps the default; the accepted grammar is
 exactly `[0-9]+`.
 
+`env_tristate_or_unset` in the same header is the one non-numeric member, and it exists because an
+A/B **lever** has a third answer that a number cannot express: `PROSPER_POST_SUBMIT_VISIBILITY` is
+`1` forced on, `0` forced off, and UNSET meaning "follow the SDK version the guest asked for". Read
+with `strtol` that collapsed -- `=on`, `=true`, `=yes` and `=enabled` all parse to 0, so every word
+spelling ran the FORCED-OFF arm and the site announced it as deliberate (#3304). It accepts
+`1`/`on`/`true`/`yes`/`enabled` and `0`/`off`/`false`/`no`/`disabled` in either case, plus `0x1` and
+`0x0` because the site it was written for read base 0; anything else -- including a number that is
+neither 0 nor 1 -- is UNSET with a line naming the value, never a silent third arm.
+
 Both headers take the variable's NAME as a literal argument, and `check_cached_env.py` treats any
 callee containing `env` as an environment WRITE — so a new reader here has to be added to that
 script's `ENV_READERS` set or it reads as an arming. That coupling is the one non-obvious thing about
--- a/prosper/src/diagnostics/env_numeric.hpp
+++ b/prosper/src/diagnostics/env_numeric.hpp
@@ -144,6 +144,48 @@ inline uint64_t env_u64_or_default_auto_capped(const char* name, const char* tex
     return value < cap ? value : cap;
 }
 
+// --- for a knob whose three answers are ON, OFF and "not set at all" ------------------------------
+//
+// A tri-state A/B lever (`PROSPER_POST_SUBMIT_VISIBILITY`, #1226/#2217) has a deliberate third
+// answer: `1` forces the contract on, `0` forces it off, and UNSET means "follow the SDK version the
+// guest asked for". Read with `strtol(e, nullptr, 0)` that collapses, because strtol answers **0**
+// for text it cannot parse -- so `=on`, `=true`, `=yes` and `=enabled`, every spelling an operator
+// reaches for, select the FORCED-OFF arm, and the site then announces "FORCED OFF" as though it had
+// been asked for. On an ordinary switch that is a lost experiment. On a lever whose verdict is still
+// open it is a MISLABELLED measurement, which is worse than no measurement at all (#3304).
+//
+// So the grammar is: `1` / `on` / `true` / `yes` / `enabled` -> 1, `0` / `off` / `false` / `no` /
+// `disabled` -> 0, either case; `0x1` and `0x0` too, because the sites converted here read base 0
+// and refusing a spelling that used to work is this header changing a well-formed setting (the same
+// N1 rule the `_auto` family exists for). Anything else -- INCLUDING a number that is neither 0 nor
+// 1, which strtol used to turn into a silent forced-ON -- is **unset** (-1) with a line on stderr
+// naming the value. Never a silent third answer. Unset or empty is -1 in silence.
+inline bool env_word_matches_nocase(const char* text, const char* lower_word) {
+    for (;; ++text, ++lower_word) {
+        if (!*lower_word) return !*text;
+        char c = *text;
+        if (c >= 'A' && c <= 'Z') c = static_cast<char>(c - 'A' + 'a');
+        if (c != *lower_word) return false;
+    }
+}
+
+inline int env_tristate_or_unset(const char* name, const char* text) {
+    if (!text || !*text) return -1;   // unset or empty: not a typo, and not a third arm
+    static const char* const kOn[] = {"on", "true", "yes", "enabled"};
+    static const char* const kOff[] = {"off", "false", "no", "disabled"};
+    for (const char* word : kOn)
+        if (env_word_matches_nocase(text, word)) return 1;
+    for (const char* word : kOff)
+        if (env_word_matches_nocase(text, word)) return 0;
+    uint64_t value = 0;
+    if (parse_u64_auto_base(text, &value) && value <= 1) return static_cast<int>(value);
+    std::fprintf(stderr,
+                 "[env] %s='%s' is neither on nor off -- treating it as UNSET and changing NOTHING. "
+                 "Accepted: 1/on/true/yes/enabled, 0/off/false/no/disabled (either case)\n",
+                 name, text);
+    return -1;
+}
+
 // --- for a knob whose "unset" answer is NOT a number ---------------------------------------------
 //
 // `PROSPER_COMPUTE_IMAGE_CACHE_MB` unset means "derive the budget from device memory", so there is
--- a/prosper/src/gpu/pm4/command_processor.cpp
+++ b/prosper/src/gpu/pm4/command_processor.cpp
@@ -2927,7 +2927,10 @@ std::atomic<bool> g_post_submit_visibility{false};
 
 // #1226 (arc7) A/B lever, default OFF and log-only in the sense that it changes nothing unless
 // set: `PROSPER_POST_SUBMIT_VISIBILITY=1` forces this model on regardless of the SDK version the
-// guest asked for, `=0` forces it off. It exists because the per-fold census (see
+// guest asked for, `=0` forces it off. (`on`/`true`/`yes`/`enabled` and `off`/`false`/`no`/
+// `disabled` work too, in either case; anything ELSE -- including a number that is neither 0 nor 1
+// -- is treated as unset and says so, because a typo must not pick an arm of a live experiment.
+// #3304.) It exists because the per-fold census (see
 // `ARCRUNNER_STATUS.md` § arc7) localised ArcRunner's corruption to the guest's builder thread
 // being released MID-FOLD by completion writes prosper applies while it is still executing the rest
 // of the same command buffer — and ArcRunner requests SDK version 10, so the post-submit contract
@@ -2935,8 +2938,13 @@ std::atomic<bool> g_post_submit_visibility{false};
 // pre-13 title is a separate question this lever does not answer; it makes the experiment runnable.
 bool post_submit_visibility_enabled() {
     static const int forced = [] {
-        const char* e = getenv("PROSPER_POST_SUBMIT_VISIBILITY");
-        const int v = e ? (int)strtol(e, nullptr, 0) : -1;
+        // #3304: the tri-state is right and the PARSE was not. `strtol` answers 0 for text it
+        // cannot read, so `=on`, `=true`, `=yes` and `=enabled` all landed on the FORCED-OFF arm
+        // and printed the line below as though that had been asked for -- a confidently mislabelled
+        // result on a lever whose verdict is open (#2217/#2219/#2223). A value that is neither on
+        // nor off is now UNSET (follow the SDK version) and says so; it is never a silent third arm.
+        const int v = prosper::diag::env_tristate_or_unset("PROSPER_POST_SUBMIT_VISIBILITY",
+                                                           getenv("PROSPER_POST_SUBMIT_VISIBILITY"));
         if (v == 1)
             fprintf(stderr, "[agc] POST-SUBMIT-VISIBILITY FORCED ON (#1226 A/B) — completion writes "
                             "stay private until the submit scope closes, regardless of SDK version\n");
--- a/prosper/src/hle/kernel/hle_kernel_time.cpp
+++ b/prosper/src/hle/kernel/hle_kernel_time.cpp
@@ -17,6 +17,7 @@
 #include "hle/sync/pthread_slot.hpp"   // #2596: resolve a guest sync slot the way libkernel does
 #include "hle/sync/sync_futex.hpp"
 #include "hle/sync/sync_retire.hpp"   // #2042: a destroyed guest sync object's storage is retired, not freed
+#include "diagnostics/env_numeric.hpp"   // #3304: a mistyped PROSPER_WAITCAP must not remove the cap
 #include <pthread.h>
 #include <chrono>
 #if defined(__linux__)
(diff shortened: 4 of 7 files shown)


mattias800 (commented):
APPROVED — independent review on exact head, I am not the author. Merge still gated on CI.

## `PROSPER_POST_SUBMIT_VISIBILITY` — the tri-state is preserved exactly

This is the half that could have gone wrong quietly, because a tri-state has three ways to be
subtly re-mapped. Checked at the call site rather than in the helper:

```cpp
const int v = prosper::diag::env_tristate_or_unset("PROSPER_POST_SUBMIT_VISIBILITY", getenv(...));
if (v == 1) fprintf(... "FORCED ON" ...);
else if (v == 0) fprintf(... "FORCED OFF" ...);
...
if (forced >= 0) return forced != 0;
return g_post_submit_visibility.load(std::memory_order_acquire);
```

`-1` still falls through to the SDK-driven atomic, `1` still forces on, `0` still forces off.
Identical semantics. And the property that matters most for a live experiment lever: **the
`FORCED OFF` line can no longer be printed by a typo.** Previously `=on` produced `strtol → 0`,
which selected the off arm *and* announced it as deliberate — a confidently mislabelled arm on a
question `CLAUDE.md` records as undecided (#2217/#2219/#2223). Now an unreadable value returns `-1`
and says so under its own `[env]` prefix, which cannot be mistaken for a lever decision.

The grammar in `env_tristate_or_unset` is well-judged:

- accepts `1/on/true/yes/enabled` and `0/off/false/no/disabled`, case-insensitively — these are the
  spellings an operator actually reaches for, and the issue's whole point was that reaching for one
  silently cost you the experiment;
- keeps `0x1`/`0x0` via `parse_u64_auto_base`, because the converted site read base 0 and refusing a
  previously well-formed spelling would be this header changing a working setting — the same rule
  the `_auto` family exists for;
- **a number that is neither 0 nor 1 now returns unset.** Under `strtol` that was a silent forced-ON.
  This is a deliberate behaviour change in the safe direction and is documented at the site.

## `PROSPER_WAITCAP`

The original defect is genuinely nasty: `atoll("-1")` → `-1` → cast to `uint64_t` = `UINT64_MAX`,
which is non-zero so it *passes* the `cap &&` guard, after which `us > UINT64_MAX` can never be true.
An operator who armed a wait cap got no wait cap and no indication. Routing through
`env_u64_or_default` removes both the sign path and the silent-zero path, and the `default_note`
spelling out that 0 means *no cap* is the right call — 0 being "off" on this knob is exactly the
thing a reader would otherwise assume backwards.

## The arms

Mirrored rather than `setenv`-driven, which is correct and not a shortcut: both sites cache in
function-local statics, so a runtime `setenv` test would be initialising after the value was already
read — vacuous in the one direction it exists to rule out. Every other arm in that file lives with
the same constraint.

Reported without-fix: both mirrors replaced with the pre-fix arithmetic → **6 failures, exit 1**;
restored → exit 0, 136 assertions. The author also reports the arms caught a case beyond the issue's
scope — an **empty** `PROSPER_POST_SUBMIT_VISIBILITY=` also forced OFF — which is a good sign the
arms are testing the behaviour rather than the example.

Both anti-drift gates were taught the new entry point (`check_env_numeric_arms.py`'s `CALL_RE`,
`check_cached_env.py`'s `ENV_READERS`), and the latter was **also missing three older readers** —
worth noting because a gate with holes reports the same clean result as a gate without them.

## Non-blocking, and worth more attention than this PR

The author's finding that **#3267's 169-row catalogue is a floor, not a count** — two of the four
sites converted here were absent from it, because the census's shape-match misses a guard written on
a wider type than the value is used at. That is already commented on #3267. Whoever sweeps the
remaining sites should not plan against 169.

mattias800:
Re-pushed after an incident on my side, and after resolving a real conflict. Recording both, since the first one looks like data loss in the history.

**What happened.** I ran `gh pr merge 3538 --squash --delete-branch` with the worktree/branch teardown chained after it by `;`. The merge **refused** — this branch was `CONFLICTING`/`DIRTY` against `main` — but the `;` meant the teardown ran anyway, deleting the local and remote branch of an unmerged PR and closing it. Recovered by pushing the commit back at the SHA `git branch -D` had printed (`89833ee5`) and reopening; head is unchanged and no work was lost. The same unguarded chain had run on three earlier merges and was only safe because those merges succeeded — the guard was never there. Fixed procedurally: `gh pr view <N> --json state` must report `MERGED` in its own call before any delete runs.

**The conflict was real and is now resolved.** `prosper/tests/diagnostics/test_env_numeric_sites.cpp` — #3537 (#2847, merged meanwhile) and this PR both appended arms to the same file. One conflict hunk, two independent blocks, resolved by keeping **both**:

- this PR's `PROSPER_POST_SUBMIT_VISIBILITY` word-spelling arms, including the discriminator loop that requires the old spelling to have answered OFF for each word;
- #3537's cadence arms asserting no input makes the divisor 0.

Verified after resolution: zero conflict markers remain, and both blocks are present. Nothing else in the merge conflicted.

**Verification of the merged result is Linux CI on this head**, stated plainly rather than implied: I did not rebuild locally, because the resolution is additive in a test file whose two halves were each green separately, and local usage is constrained. Linux is the platform that matters here. If Linux reddens on `4dc78288` this needs another pass before merging.

Issue #3426, opened 2026-09-07
Windows session-start tests cannot verify Git rev-parse


mattias800:
Windows MinGW CI for PR #3424 built successfully but failed 1 of 285 tests: `session_start`. Eight cases in `tools/test_session_start.py` report `[session-start] UNVERIFIED: Git could not verify rev-parse`, with return 2 where 0/1 was expected.

Evidence: https://github.com/mattias800/prosper/actions/runs/34095327529/job/101657608820

The startup check and its tests were introduced in #3421/#2710 and are unchanged by the graphics-cache PR. Root cause is not established; investigate the Windows subprocess/Git environment rather than assuming a cache or Vulkan failure. Linux and the Windows App build passed. Follow-up is recorded separately because the owner explicitly gated the graphics work on Linux/general CI without waiting for Windows/macOS.

mattias800:
CLAIMING: lane-sessionstart | branch fix/issue-3426-windows-git-rev-parse | 2026-09-11T19:28:23Z

mattias800:
Investigated as part of the session-start lane. **This is fixed, and it is not the same defect as #3540.** Measured rather than assumed, three ways.

**1. The interpreter started in the run cited here.** All eight failures in `actions/runs/34095327529` are raised at `test_session_start.py:53`, inside `check()`, which launches `sys.executable`. #3540's six failures are raised at `:65`, inside `hook()`, which launches the configured hook argv and gets exit 9009 from a program that is not Python. A `check()` failure is itself proof that Python ran, so the two cannot share a cause.

**2. The mechanism reproduces locally with no interpreter involved.** The pre-fix module (`883dbc8b:prosper/tools/session_start.py`) and the current one (`31d8ec34`) were each run against a hand-built POSIX answer to `rev-parse --show-toplevel` -- built by hand because the git on this host is mingw-flavour, answers `C:/...`, and survives the round trip, so a control drawn from it could not express the case:

```
pre-fix  883dbc8b -> exit 2 | UNVERIFIED: Git could not verify rev-parse
main     31d8ec34 -> exit 0 | Checkout: branch=... HEAD=...
```

[...]

Issue #1613, opened 2026-07-31
game_compute_exec and game_compute_exec_no_adaptive_storage_result race on a fixed /tmp path and fail under 'ctest -j'


mattias800:
Found while running the full suite for #1603.

## Symptom

```
99% tests passed, 2 tests failed out of 163
The following tests FAILED:
        141 - game_compute_exec (Failed)
        142 - game_compute_exec_no_adaptive_storage_result (Failed)
```

with

```
FAIL: deferred capture writes exact compute operations with snapshotted GDS input
FAIL: deferred capture replay restores pre-submit persistent GDS, not post-submit bytes
```

Re-running the **same binaries serially** passes immediately:

```
ctest -R game_compute_exec --output-on-failure
1/2 Test #141: game_compute_exec ..............................   Passed
2/2 Test #142: game_compute_exec_no_adaptive_storage_result ...   Passed
100% tests passed out of 2
```

and the full suite is **163/163** serially. Only `ctest -j4` fails.

## Cause

`prosper/tests/test_game_compute.cpp:978`, `:992`, `:1003` all use a single hard-coded path:

```cpp
deferred_gds_pending->path = "/tmp/prosper_deferred_gds_capture.prgcap";
...
"/tmp/prosper_deferred_gds_capture.prgcap", deferred_gds_capture, gds_error
...
std::remove("/tmp/prosper_deferred_gds_capture.prgcap");
```

`game_compute_exec` and `game_compute_exec_no_adaptive_storage_result` are the *same* binary registered as two ctest cases, so under `-j` both processes write, read back and then `std::remove()` the *same* file. Whichever loses the race reads the other run’s bytes or a file the other run already deleted, and the GDS-content assertions fail on correct code.

## Why it matters beyond the annoyance

This is a false-positive generator on a **content-comparison** test — exactly the class where a reviewer is most likely to believe the failure and go looking for a real capture/GDS defect. It also makes any parallel CI run nondeterministic, so a genuine regression here could be dismissed as "that flaky pair".

Two separate CLAUDE.md rules also apply: run artifacts must not go to `/tmp` at all (RAM-backed tmpfs with a per-user quota shared by every concurrent agent — a `.prgcap` there is exactly the file class the rule names), and the tmpfs filling kills the Bash tool for every agent.

## Suggested fix

[...]

PR #3539, opened 2026-09-11
gpu_replay: say which shader dump writes SPIR-V and which writes guest RDNA2

### LINKED ISSUE #3372: gpu_replay: --dump-shader and --dump-realized-shader are named the opposite way round from what they produce (and PROSPER_SHADER_DUMP_SUCCESS is silently inert)
## Correction: the SPIR-V was reachable all along. The premise of this issue was wrong.

**`gpu_replay --dump-shader DRAW:vs PATH` already writes the recompiled SPIR-V.** On the capsule this
issue was filed from:

```
--dump-shader 29:vs           30332 bytes  magic=03022307   <- SPIR-V (0x07230203 LE), spirv-val: VALID
--dump-shader 29:fs            4084 bytes  magic=03022307   <- SPIR-V
--dump-realized-shader 29:vs   1132 bytes  magic=0300a0bf   <- RDNA2 machine code
```

30332 bytes = 7583 words, matching the `vs=7583/...` figure `--inspect-only` prints for that draw. It
disassembles and validates.

**What misled me is the naming, and it is worth fixing even though the capability exists.**
`--dump-shader` sounds like "the shader the guest supplied" and gives prosper's **compiled SPIR-V**;
`--dump-realized-shader` sounds like "the shader as realized for execution" and gives the **guest's raw
RDNA2**. They are the opposite way round from what the words suggest. I read the RDNA2 out of
`--dump-realized-shader`, saw a `bfa00003` magic that `spirv-dis` rejected, concluded no SPIR-V was
obtainable, and filed this. That cost a session's worth of sequencing.

## What is actually still true

`PROSPER_SHADER_DUMP_SUCCESS=DIR` does produce nothing under `gpu_replay`, on both entry points, with
and without `PROSPER_NO_SHADER_CACHE=1`, and prints no line saying so. Every caller of
`maybe_dump_successful_shader` is in `src/gpu/execute/gpu_executor.cpp`, and the replay path does not
reach them. That remains a real gap — it is how you dump *every* shader in a frame by address rather
than one draw at a time — but it is now a **convenience**, not a blocker, and it should be labelled as
such.

## Suggested fix, re-scoped

1. **Name the two flags for what they produce.** `--dump-shader` → keep, but say "recompiled SPIR-V" in
   the usage line; `--dump-realized-shader` → say "guest RDNA2 words". One word each in the usage
   string and in `tools/gpu_replay/README.md` would have saved this issue.
2. Optionally reach the `PROSPER_SHADER_DUMP_SUCCESS` hook from the replay's compile path, or have
   `gpu_replay` say once that it saw the variable set and cannot honour it — an empty directory reading
   as "that program never compiled" is the failure mode the diagnostics folder's own `AGENTS.md` warns
   about, and it is exactly the one I fell into.

Downgrading from blocker to documentation/ergonomics. Sorry for the noise.

mattias800:
Fixes #3372.

## The failure scenario

`gpu_replay --dump-shader DRAW:vs PATH` writes prosper's **recompiled SPIR-V**.
`--dump-realized-shader DRAW:vs PATH` writes the **guest's raw RDNA2 words** that fed it. The names
read the opposite way round, and the usage string listed both flags with no indication which was
which.

The consequence is on the record: an investigation read the RDNA2 out of the flag whose name
suggested SPIR-V, saw a `bfa00003` magic that `spirv-dis` rejected, concluded that no SPIR-V was
obtainable from a capsule, and filed #3372 against a capability that had been present all along.
The reporter's own correction says it cost "a session's worth of sequencing".

Nothing was broken. The tool could always do the thing; the text at the point of use could not be
read correctly.

## Behavioural contract

- Each of the four shader-dump flags states what it **writes**, in the usage output.
- `--dump-shader-raw` is the primary spelling for the guest-RDNA2 dump; `--dump-realized-shader`
  keeps working, unchanged — it is cited in `docs/` and in `tools/shader_conformance/scan.py`.
- `PROSPER_SHADER_DUMP_SUCCESS`, which cannot work under `gpu_replay`, says so instead of writing
  nothing in silence.

## Approach

Three changes, in ascending order of how much they matter:

**1. The usage text now carries a legend** naming what each dump produces. This is the actual fix.
The README's reference section was already correct at `:617` — so what failed was not the
documentation, it was the text an operator sees at the moment of use. Fixing only prose would have
left the trap armed for everyone who does not read a 700-line README first.

**2. `--dump-shader-raw` added as an alias-compatible rename.** `-raw` already means "the
recompiler's input rather than its output" in the adjacent `--dump-compute` / `--dump-compute-raw`
pair, documented in this same README. The graphics pair now matches that convention instead of
contradicting it. Both spellings parse to one code path, so this cannot break a caller.

This goes slightly beyond the issue's re-scoped suggestion, which asked only for wording. The
justification: the issue's own diagnosis is that the names are "the opposite way round from what the
words suggest", and wording alone leaves the misleading name as the only name. It is four lines and
backward compatible.

**3. `PROSPER_SHADER_DUMP_SUCCESS` announces that it will write nothing on the paths where that is true.**

> **Corrected after review — the first version of this section was wrong.** It claimed the replay path reaches `maybe_dump_successful_shader` on no path at all. It does reach it under `--recompile-raw` and `--retry-failed-*`, which recompile through `tools/gpu_replay/compute_recompile.hpp` → `gpu::recompile_compute_shader_cached`, and that function calls the hook on every exit path. I inherited the claim from #3372's own text and shipped it as an assertion — in the change whose subject is wrong text at the point of use. See commit `427c38b9`.

A plain replay runs the capsule's stored SPIR-V and recompiles nothing, so the directory stays empty — and an empty directory reads as “that program never compiled”, the trap `src/gpu/diagnostics/AGENTS.md` warns about and the one the reporter fell into. The notice is now emitted after argument parsing and gated on the mode, so it fires exactly when it is true, and names the modes that do honour the variable.

## Docs: what was repointed and what was left

Per the charter's test — *a record of what was true keeps the old name; anything a reader would run
or follow gets repointed*:

| repointed (recipes / present tense) | left alone (records) |
| --- | --- |
| `gpu_replay/README.md` example, VS-tap and FS-tap workflows, selector list, reference paragraph | `GAME_COMPAT_ORCHESTRATION.md` instrument-trap row 63 |
| `tools/AGENTS.md` capability description | `docs/ASTERIX_BABYLON_STATUS.md` past-run account |

`tools/shader_conformance/scan.py` was left on the old spelling deliberately: it works unchanged via
the alias, and repointing it would have put a behaviour-neutral edit in a file this PR otherwise
does not touch.

## Regression arm

`gpu_replay_cli_contract` (new, test #335) drives the **real binary** and reads what an operator
would see, rather than grepping the source for a string.

The discriminating detail: it locates the legend line **for each flag** and asserts what that line
claims. Asserting merely that "SPIR-V" and "RDNA2" both appear somewhere in the output would pass
with the two meanings swapped — which is the exact defect being fixed, so that weaker test would
have been void.

It skips with **77** (`SKIP_RETURN_CODE`, reported by name as `(Skipped)`) where `gpu_replay` was not
built. Exiting 0 there would make "the tool does not exist" and "the contract holds" the same
result.

### Does it fail without the fix?

Yes — measured, not asserted. Built `gpu_replay` from **pristine `origin/main`** in a separate
worktree and ran this same test against it:

```
8 failure(s)     WITHOUT_FIX_EXIT=1
```

The two checks that *do* pass on pristine main are the two that should: the old usage already names
`--dump-realized-shader`, and the new notice is correctly silent when the variable is unset. Against
this branch's binary: **10/10, exit 0**.

### Is the alias real, or just documented?

Negative control included, since "the flag appears in usage" is not "the flag is accepted":

| invocation | result |
| --- | --- |
| `--dump-shader-bogus 0:vs out.bin missing.prgcap` | rejected, usage dumped |
| `--dump-shader-raw 0:vs out.bin missing.prgcap` | accepted, proceeds to open the capture |
| `--dump-realized-shader 0:vs out.bin missing.prgcap` | accepted, identical behaviour |

- **Unaffected:** every existing flag, every existing spelling, all dump behaviour, all replay
  behaviour. No runtime code path changes; `gpu_replay.cpp`'s only non-text change is one added
  `||` in argument parsing and one `getenv` check at the top of `main`.
- **Affected:** usage output, one new ctest case, four doc sites.

## Risks

Low. The largest is the new `if(TARGET gpu_replay ...)` block in `CMakeLists.txt`: it is placed
before the trailing `PROSPER_TEST_SCRATCH_DIR` loop, which snapshots the `TESTS` directory property
and would not have covered a test registered after it. Verified the new case is enumerated
(`Test #335`).

## Verification

Platform note, stated because it bounds what this evidence covers: **this machine's WSL has Vulkan
1.3.275 headers, and `frontends/shared/device/vulkan_runtime.hpp:11` needs `VK_API_VERSION_1_4`**, so
the live-renderer chain — and therefore `gpu_replay` — cannot be built under WSL here at all. All
execution evidence below is **Windows/MinGW with Vulkan SDK 1.4.350.0**. Linux coverage is limited to
a successful CMake configure.

```
cmake -S prosper -B <build> -G Ninja -DCMAKE_BUILD_TYPE=Release -DGAME_DUMP=<DUMP_ROOT>/PPSA24651-app0
cmake --build <build> --target gpu_replay -j 6        -> exit 0, gpu_replay linked
cmake --build <build> -j 6                            -> exit 0 (528 targets)
ctest --test-dir <build> --no-tests=error -R gpu_replay --output-on-failure
    -> 7/7 passed, exit 0   (#276 output_extent, #277 shader_dump, #278 resource_override,
                             #279 post_compute_resource, #280 pixel_inputs, #335 cli_contract, +1)
```

Docs gate over every tracked `.md`, run because this PR touches three of them:

```
git ls-files '*.md' -z | xargs -0 python3 prosper/tools/docs/check_numbered_table.py   -> 0 errors
```

That gate earned its place here: it caught me writing `README.md` back as **cp1252** in an earlier
revision of this branch, which mis-encoded an em dash and two quotes into invalid UTF-8. Repaired by
re-deriving the file from `origin/main` and re-applying the edits with explicit UTF-8; the gate is
clean and the final three-dot diff against `origin/main` is 5 files, with no EOL or encoding churn.

`git diff --check`: clean. Session-trailer gate: 0.

Full-suite Windows ctest is running; I will post the count and exit code as a comment on this PR
before merging rather than claiming it here.

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -4034,18 +4034,6 @@ if(TARGET prosper_live_renderer AND TARGET prosper_core)
   add_test(NAME present_blit COMMAND test_present_blit)
 endif()
 
-# ---------------------------------------------------------------------------
-# Per-case scratch directory for every registered test (#1613).
-#
-# Tests that must touch the real filesystem take their paths from
-# tests/fixtures/test_scratch.h, which reads PROSPER_TEST_SCRATCH_DIR. Giving each ctest
-# *case* its own root is what makes `ctest -j` safe: one test binary is
-# sometimes registered as several cases (game_compute_exec and
-# game_compute_exec_no_adaptive_storage_result are the same executable), so
-# per-binary uniqueness is not enough. The helper adds a pid component
-# underneath, which separates concurrent ctest invocations from different
-# worktrees. The root is inside the build directory, never /tmp — see the
-# header for why that matters on this machine.
 # #3372: gpu_replay's two graphics shader dumps produce OPPOSITE things -- --dump-shader writes
 # prosper's recompiled SPIR-V, --dump-shader-raw writes the guest RDNA2 that fed it -- and the
 # flag names alone do not say which way round. An investigation read the RDNA2 out of the flag
@@ -4063,6 +4051,19 @@ if(TARGET gpu_replay AND Python3_Interpreter_FOUND)
     "PROSPER_GPU_REPLAY_BIN=$<TARGET_FILE:gpu_replay>")
 endif()
 
+
+# ---------------------------------------------------------------------------
+# Per-case scratch directory for every registered test (#1613).
+#
+# Tests that must touch the real filesystem take their paths from
+# tests/fixtures/test_scratch.h, which reads PROSPER_TEST_SCRATCH_DIR. Giving each ctest
+# *case* its own root is what makes `ctest -j` safe: one test binary is
+# sometimes registered as several cases (game_compute_exec and
+# game_compute_exec_no_adaptive_storage_result are the same executable), so
+# per-binary uniqueness is not enough. The helper adds a pid component
+# underneath, which separates concurrent ctest invocations from different
+# worktrees. The root is inside the build directory, never /tmp — see the
+# header for why that matters on this machine.
 #
 # This must stay at the very end of the file: it snapshots the DIRECTORY TESTS
 # property, so any add_test() below it would not be covered. A test that did get
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -4046,6 +4046,23 @@ endif()
 # underneath, which separates concurrent ctest invocations from different
 # worktrees. The root is inside the build directory, never /tmp — see the
 # header for why that matters on this machine.
+# #3372: gpu_replay's two graphics shader dumps produce OPPOSITE things -- --dump-shader writes
+# prosper's recompiled SPIR-V, --dump-shader-raw writes the guest RDNA2 that fed it -- and the
+# flag names alone do not say which way round. An investigation read the RDNA2 out of the flag
+# whose name suggested SPIR-V and lost a session to it. The usage legend is the fix, so something
+# has to fail when the legend is dropped; nothing else in the tree can. Behavioural rather than a
+# source grep: it runs the real binary and reads what an operator would actually see.
+if(TARGET gpu_replay AND Python3_Interpreter_FOUND)
+  add_test(NAME gpu_replay_cli_contract
+           COMMAND "${Python3_EXECUTABLE}"
+                   "${CMAKE_SOURCE_DIR}/tools/gpu_replay/test_cli_contract.py")
+  # SKIP_RETURN_CODE so "gpu_replay was not built" reports as (Skipped) by name rather than as a
+  # pass -- the tool needs Vulkan and is genuinely absent in some environments.
+  set_tests_properties(gpu_replay_cli_contract PROPERTIES TIMEOUT 300 SKIP_RETURN_CODE 77)
+  set_property(TEST gpu_replay_cli_contract APPEND PROPERTY ENVIRONMENT
+    "PROSPER_GPU_REPLAY_BIN=$<TARGET_FILE:gpu_replay>")
+endif()
+
 #
 # This must stay at the very end of the file: it snapshots the DIRECTORY TESTS
 # property, so any add_test() below it would not be covered. A test that did get
--- a/prosper/tools/AGENTS.md
+++ b/prosper/tools/AGENTS.md
@@ -1132,8 +1132,10 @@ composite/scanout draws; a single `--draw N` remains supported.
 `--dump-resource DRAW:vs|ps:BINDING PATH` writes one captured resource's exact backing bytes for
 external numeric/image inspection without dereferencing the original guest address.
 `--dump-shader DRAW:vs|fs PATH` writes the captured SPIR-V module for validation/disassembly.
-Capture v19+ also supports `--dump-realized-shader DRAW:vs|fs PATH`, which writes the exact bounded raw
-RDNA2 source for a successfully realized graphics stage; use that output with `shader_inspect`.
+Capture v19+ also supports `--dump-shader-raw DRAW:vs|fs PATH` (alias: `--dump-realized-shader`),
+which writes the exact bounded raw RDNA2 source for a successfully realized graphics stage; use that
+output with `shader_inspect`. Note the pairing: the plain flag gives prosper's SPIR-V, the `-raw`
+flag gives the guest's RDNA2 that produced it (#3372).
 `--dump-compute N PATH` writes one realized compute SPIR-V module, and
 `--dump-compute-resource N:BINDING PATH` writes its exact pre-dispatch storage-buffer bytes.
 `--dump-post-compute-resource N:BINDING PATH` is the execution-side counterpart for storage images: it
--- a/prosper/tools/gpu_replay/README.md
+++ b/prosper/tools/gpu_replay/README.md
@@ -626,10 +626,14 @@ The VS/FS streams use the same content-addressed 64 KiB-per-stage,
 64 MiB-total store as failed-shader diagnostics. Captures v1-v18 remain readable; because they did not retain
 realized-stage source identities, this command reports the raw stream as unavailable instead of guessing.
 
-`PROSPER_SHADER_DUMP_SUCCESS=DIR` does **not** work under `gpu_replay` — every caller of its
-hook is in the live executor, which the replay path never reaches. `gpu_replay` now says so on
-startup when the variable is set, because an empty directory otherwise reads as “that program
-never compiled”. Use the per-draw flags above instead.
+`PROSPER_SHADER_DUMP_SUCCESS=DIR` is honoured **only in the modes that recompile**. A plain replay
+runs the capsule's stored SPIR-V and recompiles nothing, so nothing calls the hook and the
+directory stays empty — which reads as “that program never compiled”, the trap
+`src/gpu/diagnostics/AGENTS.md` warns about. `--recompile-raw` and `--retry-failed-*` **do**
+recompile, through `compute_recompile.hpp` → `gpu::recompile_compute_shader_cached`, which
+calls the hook on every exit path, so the variable works under those. `gpu_replay` warns on
+startup when the variable is set and the invocation will not reach it. For one shader by draw, use
+the per-draw flags above.
 
 Selected draw ranges and ordered prefixes write the BMP at the final selected draw target's extent when
 the returned RGBA byte count confirms that native size. Presentation-scaled replays retain the capsule's
--- a/prosper/tools/gpu_replay/README.md
+++ b/prosper/tools/gpu_replay/README.md
@@ -372,7 +372,7 @@ render-state resolve, executor ordering, detile) — it is the right guard for c
 ./build-linux/gpu_replay --dump-rtt-seed 0x7f9f504b0000 /tmp/history.bmp \
   --inspect-only /tmp/submit.prgcap
 ./build-linux/gpu_replay --dump-shader 18:fs /tmp/fragment.spv /tmp/submit.prgcap
-./build-linux/gpu_replay --dump-realized-shader 18:fs /tmp/fragment-rdna2.bin /tmp/submit.prgcap
+./build-linux/gpu_replay --dump-shader-raw 18:fs /tmp/fragment-rdna2.bin /tmp/submit.prgcap
 ./build-linux/gpu_replay --dump-compute 0 /tmp/compute.spv /tmp/submit.prgcap
 ./build-linux/gpu_replay --dump-compute-raw 0 /tmp/compute-rdna2.bin --inspect-only /tmp/submit.prgcap
 ./build-linux/gpu_replay --compute-only 0 /tmp/submit.prgcap
@@ -504,7 +504,7 @@ prints). The recompiler snapshots that instruction's destination VGPR (and the n
 the vertex-position export to it, so the geometry-probe capture reads the value back per vertex. Output shows
 both float and raw hex (intermediate values are frequently integers/bitfields — read the hex for those).
 
-Workflow: `shader_inspect` the draw's VS (`--dump-realized-shader N:vs`) to find the PC whose value you want,
+Workflow: `shader_inspect` the draw's VS (`--dump-shader-raw N:vs`) to find the PC whose value you want,
 then `PROSPER_SHADER_TAP=<pc> PROSPER_GEOM_PROBE=N`. This answers "is prosper fetching the right vertex data /
 computing the right intermediate?" without an oracle — it revealed GTA V #1163's mask draws fetch plausible
 large-integer control coordinates (so the off-screen result is data-driven, not garbage). Vertex stage only; the
@@ -543,7 +543,7 @@ PROSPER_FS_TAP=6:27 ./build-linux/gpu_replay /tmp/submit.prgcap /tmp/out.bmp
 The fragment-stage sibling of `PROSPER_SHADER_TAP`, for "why is this **pixel** the wrong colour?" — capture a
 fragment shader's intermediate at instruction `PC` (the same PC `shader_inspect` prints) and **redirect the
 MRT0 colour export to it**, so semantic draw `DRAW`'s pixels in the rendered frame *are* the value visualised. Unlike the
-VS tap it needs no separate capture — the output BMP is the readout. Point `--dump-realized-shader DRAW:fs` +
+VS tap it needs no separate capture — the output BMP is the readout. Point `--dump-shader-raw DRAW:fs` +
 `shader_inspect` at the FS to find the PC (an `image_sample`/MIMG result, a UV, a pre-blend colour), then
 `PROSPER_FS_TAP=DRAW:PC`. gpu_replay re-recompiles only that draw's FS (needs a v19+ capsule with the raw FS
 stream); the rest of the frame renders normally, so read the tapped draw's region.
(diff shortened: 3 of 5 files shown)


mattias800:
Full-suite Windows ctest, exact head `1a7246a3`, as promised in the body.

```
ctest --test-dir <build> --no-tests=error -j 4
    334/335 passed, exit 8
    Skipped: 11 refactor_survey_measurement, 24 hle_calls_value_capture
    FAILED:  36 session_start
```

**The one failure is pre-existing and unrelated to this PR.** I did not take that on trust — I built
and ran the same case from a separate worktree at pristine `origin/main`:

```
ctest --test-dir <pristine-main-build> --no-tests=error -R ^session_start$
    0% passed, 1 failed   (Ran 14 tests, FAILED (failures=6))
```

Identical failure, same six sub-cases, on a tree containing none of this branch's commits.

The root cause is worth recording since it is not what the test name suggests. `test_session_start.py`
shells out to Python, and on this Windows host that resolves to the Microsoft Store **App Execution
Alias stub**, which exits **9009** without running anything:

```
AssertionError: 9009 != 0 : Python was not found; run without arguments to install from the
Microsoft Store, or disable this shortcut from Settings > Apps > Advanced app settings >
App execution aliases.
```

So the hook under test never executes on this box; the assertions fail because the interpreter was
never found, not because the session-start logic is wrong. That is adjacent to #3426 (*Windows
session-start tests cannot verify Git rev-parse*) but is a distinct mechanism, so I will check
whether #3426 already covers it and file separately if not rather than widening this PR.

Nothing else in this PR touches that path.

mattias800 (commented):
Independent adversarial review. **REJECTED**, on one finding, with a small remedy. Everything else I checked is sound and I say where I found nothing.

## Blocking: the new notice states something that is false on `--recompile-raw`, and the README states it as fact

`gpu_replay.cpp` now prints, unconditionally at the top of `main`:

```
PROSPER_SHADER_DUMP_SUCCESS is set but does NOTHING here: it is a
live-renderer variable, and gpu_replay does not reach its hook.
No file will be written for it.
```

and `tools/gpu_replay/README.md` gains *"every caller of its hook is in the live executor, which the replay path never reaches."*

The first clause is true — every caller of `maybe_dump_successful_shader` is in `src/gpu/execute/gpu_executor.cpp`, I checked. The inference drawn from it is not. `gpu_replay` calls **into** that file:

- `tools/gpu_replay/gpu_replay.cpp:2710` (`--recompile-raw`) and `:3261` (`--retry-failed-chain`) call `prosper::tools::recompile_captured_compute` / `recompile_failed_compute_stage`;
- both are in `tools/gpu_replay/compute_recompile.hpp` (`:94`, `:145`) and call `gpu::recompile_compute_shader_cached`;
- that is `src/gpu/execute/gpu_executor.cpp:2135`, and it calls `maybe_dump_successful_shader` on **every** exit — the `PROSPER_NO_SHADER_CACHE` bypass (`:2200`), the shared-lock cache hit (`:2214`), the double-check hit (`:2227`), and the compile-and-insert path (`:2234`).

So on a `--recompile-raw` replay of a v39+ capsule with compute in it, `PROSPER_SHADER_DUMP_SUCCESS` does write files, while the tool has just told the operator it will not. This is the same defect class the PR exists to fix — wrong text at the point of use — and it is worse than the original, because the original text was merely silent while this one is assertive.

It is also the charter's "stated as measured but actually inferred" case, and I think I can name where it came from: the reporter measured "produces nothing, on both entry points, with and without `PROSPER_NO_SHADER_CACHE=1`". A default replay consumes **stored** SPIR-V and compiles nothing, so that measurement is entirely consistent with the hook being reachable in the modes that do recompile. The PR generalised a true observation about one path into a claim about all of them. (I could not build here, so I am asserting the static call chain, not a run. The chain is unconditional once those two call sites are reached.)

Remedy, either way round and both small:

1. Scope the sentence: say the variable produces nothing on the **stored-SPIR-V** replay path, which is the default, and that only `--recompile-raw` / `--retry-failed-chain` compute recompiles reach the hook; or
2. arm the notice only when neither of those flags is present, and say the true thing in each case.

The new test asserts the wording (`"NOTHING" in armed or ...`), so whichever wording lands, that arm needs to move with it.

## Non-blocking, but please fix before merge

**The ctest evidence cannot distinguish "ran" from "skipped".** `SKIP_RETURN_CODE 77` is the right design and I am not arguing against it — but ctest counts a skipped case in "7/7 passed", so the quoted

```
ctest ... -R gpu_replay --no-tests=error  -> 7/7 passed, exit 0   (... #335 cli_contract ...)
```

is green under both "the legend was checked" and "`PROSPER_GPU_REPLAY_BIN` never resolved and the contract was never read". The `10/10, exit 0` in the body is the script run by hand, not through ctest, so nothing posted so far establishes that the `$<TARGET_FILE:gpu_replay>` genex reached the child. One line fixes it: post `ctest -R gpu_replay_cli_contract -V` (or `--output-on-failure` output showing the ten `[ok]` lines). This matters more than usual here because the failure is permanent and silent — a test that skips forever reports the same as one that passes forever, which is the charter's "green run that executed nothing".

On the genex itself I found nothing wrong. CMake evaluates generator expressions in test properties, and `set_property(TEST ... APPEND PROPERTY ENVIRONMENT ...)` is the same form the trailing scratch-dir loop uses, so it will not be clobbered — see below.

**CMake placement is functionally right and textually wrong.** The block does sit before `get_property(prosper_registered_tests DIRECTORY PROPERTY TESTS)`, so the new case **is** enumerated by that loop, and the loop `APPEND`s rather than `set_tests_properties(... PROPERTIES ENVIRONMENT ...)`, so `PROSPER_GPU_REPLAY_BIN` survives it. I specifically looked for that clobber and it is not there. But the block was inserted **into the middle of a single comment block**: the `# ----` header reads "Per-case scratch directory for every registered test (#1613)", its body ends "...never /tmp — see the header for why that matters", then your fifteen lines, then a dangling `#` + "This must stay at the very end of the file: it snapshots the DIRECTORY TESTS property". A later reader has to work out that the "must stay at the very end" paragraph refers to the `get_property` loop and not to the gpu_replay test. Move the whole `if(TARGET gpu_replay ...)` block above the `# ---------` divider; nothing else changes.

## What I checked and found nothing

- **Callers of the old spelling.** `git grep dump-realized-shader` gives exactly the sites the body lists. `tools/shader_conformance/scan.py:275` passes the flag to the real binary and scores the **exit code**, and `:332`/`:342` use it only as a dict-ish key in its own fake-replay stub — neither parses usage text, so the alias keeps it working unchanged. Leaving it on the old spelling is the right call. The two doc sites kept on the old name (trap row 63, `ASTERIX_BABYLON_STATUS.md`) are records, correctly left per the charter's record-vs-recipe test.
- **Skip semantics.** A broken-but-present binary does not skip: an empty or wrong output fails the checks and exits 1, and a hang raises `TimeoutExpired` out of `subprocess.run` for a non-zero exit. The skip is reachable only when the env var is unset or the path does not exist — which is exactly the stated case, plus the mis-resolution case above.
- **Test strength.** The per-flag legend lookup is the right discriminator; asserting only that both words appear would pass with the meanings swapped, and you say so. The `--dump-shader` arm additionally asserts `"RDNA2" not in spirv_line`, which is what makes it directional.

## On scope — you asked me to argue it

`--dump-shader-raw` is justified and I would keep it. The argument against ("the reporter re-scoped this to a wording fix") treats the re-scope as a ceiling, but the reporter's own diagnosis is that the *name* is backwards; a legend that explains a misleading name leaves the misleading name as the only spelling anyone can type, and the legend is only read by someone who already ran the tool wrong. The convention argument is the strong one: `-raw` already means "the recompiler's input" in the adjacent compute pair in the same tool, so the graphics pair was the outlier, and four lines with a retained alias makes it consistent. What would have been scope creep is *removing* the old spelling, and you did not.

The `PROSPER_SHADER_DUMP_SUCCESS` notice is also in scope — the issue asks for it explicitly. It just has to be true.

Re-review is one wording change and one evidence line away.

mattias800 (commented):
Blocking finding accepted and fixed on `427c38b9`. Thank you — this was the right catch, and it was right for a sharper reason than "a sentence was inaccurate".

**The claim was false and I verified it at the source rather than taking your word for it**, which is the rule this repo applies to corrections in both directions: `recompile_compute_shader_cached` (`gpu_executor.cpp:2135`) calls `maybe_dump_successful_shader` at `:2200`, `:2214`, `:2227` and `:2234` — every exit path — and `compute_recompile.hpp` reaches it from both `--recompile-raw` and `--retry-failed-*`. You are correct.

The origin is worth recording: the sentence came from **#3372's own text**, and I shipped it as an assertion in a startup notice — in the change whose entire subject is wrong text at the point of use. The reporter measured the default stored-SPIR-V path; the generalisation to *all* paths was mine, and a citation I inherited is exactly the kind the charter says gets checked least.

**Fixed by mode, not by wording.** The notice moved after argument parsing and is gated on `recompile_raw || !retry_failed_chain_spec.empty() || !retry_failed_stage_spec.empty()`. A plain replay is still warned — that part was true and is the trap worth warning about — and a recompiling run is told nothing false. Usage legend and README scoped the same way, and both now name the modes that *do* honour the variable rather than implying it never works.

**Your non-blocking points, both actioned:**

1. *"`7/7 passed` cannot distinguish ran from skipped."* Correct, and it mattered. Ran `ctest -R ^gpu_replay_cli_contract$ -V`: the child's `[ok]` lines come through, so the `$<TARGET_FILE:gpu_replay>` genex reached it — a skip exits 77 before printing anything. Clean run: `1/1 Passed, exit 0`.
2. *CMake block inside the #1613 comment.* You were right, and I had misread my own insertion point: the block split the scratch-directory comment's header from its "must stay at the very end" continuation. Moved above that section.

**And your method found a third defect, in the fix for the first.** Two of my three new arms were green in *both* directions. Searching the whole output for `--recompile-raw` passes against the pre-fix binary, because the **usage synopsis lists the flag** — the arms were satisfied by the synopsis, not the notice. A `notice_block()` helper now isolates the notice, with a comment recording why it exists. Re-measured against a rebuilt `1a7246a3`: **3 of 3 red without the fix, 0 with it**, where it had been 1 of 3.

That is your case-12 finding on #3543 reproducing itself inside this PR, which is a good argument that it was never a one-off.

PR body corrected too — the false section is replaced rather than left standing under a merged PR, and the replacement is labelled as a correction.

mattias800 (commented):
Re-review of the delta `1a7246a3..427c38b9` only. **APPROVED.**

## The blocking finding is fixed, and fixed the right way

Gating on `recompile_raw || !retry_failed_chain_spec.empty() || !retry_failed_stage_spec.empty()` rather than softening the sentence is the better of the two remedies I offered: the notice is now silent exactly where it would have been wrong, and it names the modes that do honour the variable, so a reader who needs that path is pointed at it instead of away from it. The README paragraph and the usage legend now say the same scoped thing, and the README states the call chain (`compute_recompile.hpp` → `recompile_compute_shader_cached` → the hook) so the next reader can check it rather than trust it.

I verified the placement matters and is right: the notice sits at `gpu_replay.cpp:~2440`, and the capture is not read until `:2558` (`read_gpu_capture`). So the negative arm — `--recompile-raw no-such-capture.prgcap` — really does reach the notice site before failing to open the capture. It is not passing vacuously on the fixed binary, which was the one way that arm could have been void.

## `notice_block()` — you asked me to look hardest here; I think it is sound

It keys on a line containing `PROSPER_SHADER_DUMP_SUCCESS` whose `lstrip()` starts with `[gpureplay]`, then takes continuations that start with exactly 12 spaces (I counted the literal: 12, matching the C string's continuation indent).

- **Under-capture:** I checked the one mechanism that would cause it — `run()` returns `stdout + stderr` concatenated, not interleaved, so the notice's lines are contiguous in the searched text. The last continuation (`--dump-shader-raw DRAW:vs|vs-main|fs (guest RDNA2).`) is 12-space indented and is captured. No under-capture found.
- **Over-capture:** the loop stops at the first line not starting with 12 spaces. The next thing `main` can print is the `resource_override_submit_no` warning, whose first line begins with `[gpureplay]`, so it terminates the block. Over-capture needs a *following* line that begins with ≥12 spaces with no un-indented line between it and the notice; nothing on that path does. The residual coupling is that the helper depends on the 12-space continuation style, which is a real but cheap dependency — if the notice is ever re-indented the helper silently returns a short block rather than an empty one, which weakens the arms rather than reddening them. Not worth changing now; worth knowing.
- **It discriminates.** Against a pre-fix binary the old notice is still found (same `[gpureplay]` prefix, same indent), so `"--recompile-raw" in armed` is red, and the `--recompile-raw` arm is red because the old notice printed before argument parsing. That matches your 3-of-3.

Your diagnosis of why two arms were green in both directions is right and is the interesting part: `--recompile-raw` appears in the synopsis, so whole-output matching could not distinguish the notice from the usage text. That is the same "a control that cannot fail" shape the charter's trap list collects, found inside a fix for the same class — worth a line in the PR body if it is not already there.

## Non-blocking

Both earlier points are actioned: the `-V` run shows the child's `[ok]` lines, so the case executed rather than skipped, and the CMake block no longer splits the `#1613` comment. I checked the moved block still precedes `get_property(... DIRECTORY PROPERTY TESTS)` — it does, so the scratch-dir loop still covers it.

Nothing else found in the delta. Merge when your Linux checks are green.

Issue #3533, opened 2026-09-11
prosper can wedge amdgpu's VRAM manager (ttm_resource_init spinlock) and take the host down — twice in one day, once from ctest


mattias800:
## prosper can wedge amdgpu's VRAM manager and take the whole machine down — twice in one day, from two different binaries

Both of the development box's hard freezes on 2026-09-11 have the **same kernel signature**, and
neither is a prosper crash: a prosper process enters `amdgpu_bo_create` and never comes out, spinning
in `native_queued_spin_lock_slowpath` inside TTM's resource allocator. The kernel tags itself
`Tainted: [D]=DIE`, one CPU is held forever, and the machine degrades until it has to be power-cycled.

| | freeze 1 | freeze 2 |
| --- | --- | --- |
| process | `screenshot` (PID 2654618) | `test_rdna2_to_s` (PID 254306) |
| what it was | a title run | the **ctest** target `test_rdna2_to_spirv_exec` |
| first stall | 03:54:21 | 09:30:45 |
| last stall | 07:36:21 | 13:18:45 |
| duration wedged | **3 h 42 m** | **3 h 48 m** |
| RCU stall reports | 75 | 77 (76 naming the one process) |

One PID, one CPU, for the entire window in both cases. `rcu_preempt self-detected stall` repeats
every three minutes until the box dies.

```text
CPU: 22 UID: 1000 PID: 254306 Comm: test_rdna2_to_s Tainted: G      D    O  7.2.1-ogc4.1.fc44.x86_64
RIP: 0010:native_queued_spin_lock_slowpath+0x6d/0x310
Call Trace:
 _raw_spin_lock+0x29/0x30
 ttm_resource_init+0x83/0x1b0 [ttm]
 amdgpu_vram_mgr_new+0xed/0x710 [amdgpu]
 ? dmem_cgroup_try_charge+0x98/0x200
 ttm_resource_alloc+0x9a/0x170 [ttm]
 ttm_bo_alloc_resource+0x126/0x590 [ttm]
 ttm_bo_validate+0xb8/0x240 [ttm]
 ttm_bo_init_reserved+0x15c/0x1e0 [ttm]
 amdgpu_bo_create+0x1e6/0x540 [amdgpu]
 amdgpu_bo_create_vm+0x39/0x70 [amdgpu]
```

[...]

PR #3541, opened 2026-09-11
loader: unresolved DATA imports must not bind to a code stub, and AddAmprEvent must keep the guest's udata (#3529, #3500)

### LINKED ISSUE #3529: Unresolved DATA imports bind to a CODE stub: the guest reads prosper's machine code as data, and a write corrupts the import stub table

mattias800:
### 1. Unresolved DATA imports were bound to a CODE stub (#3529)

`linker.cpp` pass 2 bound **every** unresolved import to a slot in the executable import-stub
aperture, with no distinction between a function and a variable — `Symbol` carried no ELF symbol
type, so the linker had nothing to branch on. A data import therefore made the guest read prosper's
own machine code as the variable's value, and — the serious half — made a guest **store** through
that import rewrite an emitted trampoline, silently and with arbitrary delay.

Measured on `PPSA25258` with a hardware execute breakpoint on the guest's own load of its stack
canary:

```text
[hwbp] #1 rip=eboot+0x28c3978 rax=0x600000000 [rax]=0x000000bf ... tid=2535326
```

`0x600000000` is `BOOT_STUB`; `0x000000bf` is stub 0's `mov edi, 0`. The title's stack canary was
prosper's machine code, and it died in `__stack_chk_fail` on its Wwise I/O thread.

**The population is not exotic.** Over the 60 local dumps, three near-universal libc/libkernel
variables reach an aperture: `__stack_chk_guard` (728 bindings, **60 of 60 titles**), one unnamed
`libSceLibcInternal` object (182 bindings, 53 titles) and `__progname` (96 bindings, 60 titles).
Every title in the corpus has been reading stub bytes for at least one of them.

`Symbol`/`Import` now carry `ELF64_ST_TYPE(st_info)` — the only field in an undefined symbol entry
that separates the two cases, since `shndx`, `value` and `size` are identical for both. An
`STT_OBJECT` import goes to a separate writable, never-executable, zero-filled aperture at
`BOOT_IMPORT_DATA`. NOTYPE, FUNC and TLS keep the stub they have always had, and cross-module
resolution still wins over both.

`__stack_chk_guard` is seeded with a fixed non-zero process-lifetime value. Non-zero is what
libkernel supplies, and it is also what preserves measured behaviour: the stub bytes every title has
been reading are non-zero and stable, which is exactly why a defect this universal stayed invisible.
A zero canary would additionally disable `-fstack-protector` in every guest.

`nid_census --data-only` reports the same census over any set of dumps.

### 2. `sceKernelAddAmprEvent` logged the guest's `udata` and threw it away (#3500)

`k_add_ampr_event(eq, id, udata)` named the argument in its own signature comment, printed it under
`PROSPER_EVLOG`, and dropped it — `prosper_eq_add_apr` took only `(eq, id)`. Every APR completion
event therefore carried `udata = 0`, and `sceKernelGetEventUserData` honestly returned 0.

NINJA GAIDEN 4 dereferences that value directly (eboot+0x28c3196):

```asm
call sceKernelGetEventUserData
mov  edi, DWORD PTR [rax]      ; <- rax == 0
```

Measured with `PROSPER_EVLOG=1`, the guest registers four handlers with four distinct live pointers
0x20 apart — an array of per-request contexts, all four discarded:

```text
[ev] AddAmprEvent eq=0x7f22f400fc90 id=0 udata=0x224a280780
[ev] AddAmprEvent eq=0x7f22f400fc90 id=1 udata=0x224a2807a0
[ev] AddAmprEvent eq=0x7f22f400fc90 id=2 udata=0x224a2807c0
[ev] AddAmprEvent eq=0x7f22f400fc90 id=3 udata=0x224a2807e0
```

This file's own USER-event path already did the right thing and is the model followed here:
`k_add_user_event` stores it and `k_trigger_user_event` posts it back. The AMPR sibling simply did
not — the same question answered two ways depending on which registration call the guest used.

Both delivery dialects now carry it. The lookup is a separate function because the post path
deliberately runs with no APR lock held, so the udata is resolved before the post rather than inside
it. Registration stays idempotent but now **upgrades**: three paths register the same `(eq, id)` and
only `AddAmprEvent` carries a udata, so a non-zero value fills an empty slot and nothing ever
overwrites a real one with 0.

## Measured

`tools/screenshot`, Linux/RADV, default launch, no environment variables, `PPSA25258`:

| | before | after |
| --- | --- | --- |
| outcome | `__stack_chk_fail` → `ud2` on `AK::IOThread` | runs its full bound; no canary abort |
| frames | 0 | 0 |

**The rung does NOT move.** Two named causes are removed and the title still presents zero frames; a
further fault in the same Wwise layer remains. Stated plainly because a cleared blocker is not a
milestone.

## Tests

- `loader_synth_link` arms 7 and 8 (no dump, runs in CI): one NID declared `STT_OBJECT` vs
  `STT_FUNC` **at the same symbol index and GOT offset**, and a data import a sibling module defines.
  The pair shows the declaration is what moves the binding; arm 8 pins that the aperture is the
  fallback, not the rule.
- `import_data_seed` (no dump): the canary is written for its own NID and for nothing else, and a
  slot too small is refused rather than half-written.
- `trap_stubs`: the accounting identity closes over three buckets, every data slot is asserted
  outside the whole stub aperture, and the mapped slot is read back, written, and read again.

Local `ctest` is deliberately **not** quoted on this head: a `ctest -j4` run wedged this machine's
amdgpu driver twice today (#3533), so validation is left to CI, which runs the same suite. The
Windows MinGW job matters here — it is the platform arm of a loader change.

Fixes #3529
Refs #3500
Refs #3502

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -671,6 +671,11 @@ if(TARGET prosper_core)
              COMMAND test_nid "${GAME_DUMP}/eboot.bin")
   endif()
 
+  # #3529: the initial value of an unresolved DATA import. Pure -- no dump, no mapping.
+  add_executable(test_import_data_seed tests/hle/dispatch/test_import_data_seed.cpp)
+  target_link_libraries(test_import_data_seed PRIVATE prosper_core)
+  add_test(NAME import_data_seed COMMAND test_import_data_seed)
+
   add_executable(test_hle_registered tests/hle/test_hle_registered.cpp)
   target_link_libraries(test_hle_registered PRIVATE prosper_core)
   add_test(NAME hle_functions_registered COMMAND test_hle_registered)
--- a/prosper/src/hle/dispatch/import_data_seed.cpp
+++ b/prosper/src/hle/dispatch/import_data_seed.cpp
@@ -1,0 +1,23 @@
+#include "import_data_seed.hpp"
+#include "nid.hpp"
+
+#include <cstring>
+
+namespace prosper {
+
+bool seed_import_data(const std::string& nid, uint8_t* slot, size_t size) {
+    if (!slot) return false;
+    // Hashed at first use rather than written as a literal, for the reason nid_census's --self-check
+    // exists: a hard-coded 11-character string is unverifiable by inspection, while `nid_hash` is
+    // the same function the loader resolves imports with. If the two ever disagreed, everything
+    // would disagree together instead of this one table silently missing.
+    static const std::string kStackChkGuard = nid_hash("__stack_chk_guard");
+    if (nid == kStackChkGuard && size >= sizeof(uint64_t)) {
+        const uint64_t v = kStackGuardValue;
+        memcpy(slot, &v, sizeof v);
+        return true;
+    }
+    return false;
+}
+
+} // namespace prosper
--- a/prosper/src/hle/dispatch/import_data_seed.hpp
+++ b/prosper/src/hle/dispatch/import_data_seed.hpp
@@ -1,0 +1,46 @@
+// import_data_seed.hpp — the initial value of an unresolved Sony DATA import.
+//
+// The dispatcher answers unresolved FUNCTION imports; this answers the handful of unresolved
+// *variables*. Both are the same question — "prosper is this Sony library, what does it provide?" —
+// asked about an object instead of an entry point, which is why it lives beside the dispatcher.
+//
+// Everything not listed here stays ZERO. Zero is the honest default for a variable whose contents
+// prosper does not know: it is stable, it is what a reader gets from a fresh page, and for a
+// pointer-shaped object it makes the guest take its own null branch instead of dereferencing
+// whatever bytes happened to be there.
+#pragma once
+#include <cstddef>
+#include <cstdint>
+#include <string>
+
+namespace prosper {
+
+// The stack canary prosper supplies through `__stack_chk_guard`.
+//
+// WHY A VALUE AT ALL, AND WHY THIS ONE. `__stack_chk_guard` is a libkernel *variable*, and libkernel
+// is prosper — so supplying it is reimplementation, not invention. Every function a title compiled
+// with `-fstack-protector` stores it into its frame on entry and compares on return, so the only
+// property the contract needs is that both reads agree; the platform additionally makes it non-zero
+// and per-process random.
+//
+// Non-zero rather than zero is the choice that PRESERVES measured behaviour. Before #3529 this
+// import was bound into the code-stub aperture, so every title in the local corpus — all 60 of
+// them, 728 bindings — has been reading prosper's own machine code here and booting with it. That
+// value is non-zero and stable, which is exactly why a defect this universal stayed invisible.
+// Leaving the slot zero would therefore have been the novel change, and it would also silently
+// disable `-fstack-protector` in every guest.
+//
+// The low byte is 0x00 on purpose (the "terminator canary" convention): a string-based overflow that
+// runs through the canary cannot copy past it without writing the NUL. The remaining bytes spell
+// "PROSPER" in memory order, so a canary slot is recognisable on sight in a stack dump. Fixed rather
+// than randomised so a run is reproducible — prosper has no attacker model, and a deterministic
+// boot is worth more here than an unpredictable canary.
+inline constexpr uint64_t kStackGuardValue = 0x52455053'4F525000ull;
+// little-endian bytes, in memory order: 00 'P' 'R' 'O' 'S' 'P' 'E' 'R'
+
+// Write the initial bytes of the data import `nid` into `slot` (`size` bytes available).
+// Returns true when a value was written, false when the slot should be left as it is (zero).
+// `slot` is assumed zero-filled on entry, so a partial write is still well-defined.
+bool seed_import_data(const std::string& nid, uint8_t* slot, size_t size);
+
+} // namespace prosper
--- a/prosper/src/hle/kernel/hle_kernel_time.cpp
+++ b/prosper/src/hle/kernel/hle_kernel_time.cpp
@@ -1735,9 +1735,15 @@ HLE(k_eq_getcount){
 // individually. See the discrimination comment in prosper_eq_post_apr_token below.
 // CONFIDENCE: HIGH — ctor seeding, walk bounds, unconditional last:=cnt store, handler match/hash/
 // null-entry paths all from static disassembly; tag-echo event consumption live-verified (#180).
+uint64_t prosper_eq_apr_udata(uint64_t eq, int64_t id);   // defined below, with the registry
+
 namespace {
     constexpr int16_t EVFILT_AMPR_MODELED = -24;   // guest never reads filter; distinct on purpose
-    struct AprEqReg { uint64_t eq; int64_t id; };
+    // udata is the guest's own pointer, handed to sceKernelAddAmprEvent and handed BACK on every
+    // completion event through sceKernelGetEventUserData. It is not decoration: NINJA GAIDEN 4's
+    // Wwise I/O hook dereferences it directly (`call sceKernelGetEventUserData; mov edi,[rax]` at
+    // eboot+0x28c319c), so dropping it is a null pointer handed to the guest (#3500).
+    struct AprEqReg { uint64_t eq; int64_t id; uint64_t udata = 0; };
     // Own mutex (NOT g_eq_mx): the post path calls eq_post/eq_find, which lock g_eq_mx themselves.
     // Detached APR delivery threads can outlive main, so the mutex and every container they touch
     // are one intentionally immortal heap object. Heap placement also keeps the hot mutex out of
@@ -1756,9 +1762,9 @@ namespace {
         return (eq_identity << 6) | (ring & 0x3f);
     }
     void apr_post(uint64_t eq, uint64_t eq_identity, int64_t id,
-                  unsigned ring, uint64_t token, bool coalesce) {   // no APR lock held
+                  unsigned ring, uint64_t token, bool coalesce, uint64_t udata) {   // no APR lock held
         SceKEvent e{}; e.ident = id + (int64_t)ring; e.filter = EVFILT_AMPR_MODELED;
-        e.data = (int64_t)token; e.udata = 0;
+        e.data = (int64_t)token; e.udata = udata;
         eq_post(eq, e, coalesce, eq_identity);
     }
 }
@@ -1820,6 +1826,9 @@ void prosper_eq_post_apr_token(uint64_t eq, uint64_t eq_identity,
                 nanosleep(&ts, nullptr);
                 for (auto& p : batch) {
                     SceKEvent e{}; e.ident = 0; e.filter = EVFILT_AMPR_MODELED; e.data = (int64_t)p.token;
+                    // Both dialects hand the udata back; the guest reads it through the same
+                    // accessor regardless of which one delivered the event.
+                    e.udata = prosper_eq_apr_udata(p.eq, 0);
                     eq_post(p.eq, e, /*coalesce=*/false, p.eq_identity);
                 }
                 lk.lock();
@@ -1887,7 +1896,9 @@ void prosper_eq_post_apr_token(uint64_t eq, uint64_t eq_identity,
             std::lock_guard lk(state.mx);
             hwm = state.tag_hwm[hwm_key];
         }
-        apr_post(eq, eq_identity, id, ring, ((uint64_t)ring << 58) | hwm, counter_dialect);
+        // Resolved BEFORE the post, not inside it: apr_post runs with no APR lock held on purpose.
+        apr_post(eq, eq_identity, id, ring, ((uint64_t)ring << 58) | hwm, counter_dialect,
+                 prosper_eq_apr_udata(eq, id));
     }).detach();
 }
 // Assign the next completion token for `ring` (0-based, 6 bits) — for UNBOUND submits only, whose
@@ -1905,15 +1916,33 @@ uint64_t prosper_apr_next_token(unsigned ring) {
 // by the guest via record polling, and replaying invented counters would regress the listener's
 // ctor-seeded per-ring last-processed (see the block comment above; the pre-#208 replay was the
 // root cause of the #180 range-walk fault).
-void prosper_eq_add_apr(uint64_t eq, int64_t id) {
+void prosper_eq_add_apr(uint64_t eq, int64_t id, uint64_t udata) {
     AprTokenState& state = apr_token_state();
     std::lock_guard lk(state.mx);
     for (auto& r : state.eq_regs)
-        if (r.eq == eq && r.id == id) return;   // idempotent
-    state.eq_regs.push_back({ eq, id });
+        if (r.eq == eq && r.id == id) {
+            // Idempotent, but UPGRADING: several paths register the same (eq, id) and only
+            // sceKernelAddAmprEvent carries the guest's udata — the command-buffer binding and the
+            // measure-sizing compatibility call both pass 0. A plain early return would keep
+            // whichever arrived first, so a real pointer would be lost whenever one of those ran
+            // before AddAmprEvent. Never overwrite a real udata with 0.
+            if (udata && !r.udata) r.udata = udata;
+            return;
+        }
+    state.eq_regs.push_back({ eq, id, udata });
+}
+
+// The udata registered for this (eq, id), or 0 if none. Separate from the post path because that
+// path runs with no APR lock held and must not take one -- it calls eq_post, which locks g_eq_mx.
+uint64_t prosper_eq_apr_udata(uint64_t eq, int64_t id) {
+    AprTokenState& state = apr_token_state();
+    std::lock_guard lk(state.mx);
+    for (const auto& r : state.eq_regs)
+        if (r.eq == eq && r.id == id) return r.udata;
+    return 0;
 }
 HLE(k_add_ampr_event) {   // sceKernelAddAmprEvent(eq, id, udata)
-    if (a0) prosper_eq_add_apr(a0, (int64_t)a1);
+    if (a0) prosper_eq_add_apr(a0, (int64_t)a1, a2);
     if (evlog()) fprintf(stderr, "[ev] AddAmprEvent eq=0x%llx id=%lld udata=0x%llx\n",
         (unsigned long long)a0, (long long)a1, (unsigned long long)a2);
     return 0;
--- a/prosper/src/hle/memory/hle_kernel_mem.cpp
+++ b/prosper/src/hle/memory/hle_kernel_mem.cpp
@@ -353,7 +353,7 @@ static uint64_t ampr_arglog(const char* tag, uint64_t a0, uint64_t a1, uint64_t 
 }
 
 uint64_t prosper_apr_next_token(unsigned ring);                          // hle_kernel_time.cpp
-void prosper_eq_add_apr(uint64_t eq, int64_t id);                        // "
+void prosper_eq_add_apr(uint64_t eq, int64_t id, uint64_t udata);        // "
 uint64_t prosper_eq_identity(uint64_t eq);                               // " (lifetime guard)
 void prosper_eq_post_apr_token(uint64_t eq, uint64_t eq_identity,
                                int64_t id, uint64_t token);              // " (tag echo, #208)
@@ -506,7 +506,7 @@ static uint64_t apr_cb_set_equeue(uint64_t command_size, bool eager_completion,
                                  uint64_t a0, uint64_t a1, uint64_t a2,
                                  uint64_t a3, uint64_t a4, uint64_t a5) {
     ampr_arglog("H896Pt-yB4I(CbSetEqueue)", a0, a1, a2, a3, a4, a5);
-    if (a1) prosper_eq_add_apr(a1, (int64_t)a2);
+    if (a1) prosper_eq_add_apr(a1, (int64_t)a2, 0);   // binding call carries no udata
     bool start_worker = false;
     if (a0) {
         AprBoundState& state = apr_bound_state();
@@ -2643,7 +2643,7 @@ HLE(k_ampr_measure_write_equeue) { // sSAUCCU1dv4 = sceAmprMeasureCommandSizeWri
     // DOLL's older SDK wrapper passes the live event queue and APR id here while constructing its
     // listener. Until prosper executes the encoded command itself, preserve that registration as
     // a compatibility side effect. Sonic's sizing call passes a null queue and remains pure.
-    if (a0) prosper_eq_add_apr(a0, (int64_t)a1);
+    if (a0) prosper_eq_add_apr(a0, (int64_t)a1, 0);   // sizing call carries no udata
     return 20;
 }
 HLE(k_ampr_destruct_320) {
--- a/prosper/src/host/image/AGENTS.md
+++ b/prosper/src/host/image/AGENTS.md
@@ -21,7 +21,12 @@ what produces that set — it is where `<dump>/Media/Plugins/…` becomes a `Lin
   replacements of Sony libraries, and this is what keeps them from being linked and shadowing
   prosper's own implementations.
 - **`exec_image_{linux,win}.cpp`** behind `exec_image.hpp` — the per-platform mapping substrate.
-  Platform divergence belongs *here*, not in `boot_program.cpp`.
+  Platform divergence belongs *here*, not in `boot_program.cpp`. It owns **two** fixed apertures,
+  not one: the executable import-stub table at `BOOT_STUB` (`install_stubs`) and the writable,
+  never-executable import-DATA table at `BOOT_IMPORT_DATA` (`install_import_data`, #3529). They are
+  separate address windows on purpose — `callback_fs.hpp` reads any return address inside
+  `[0x600000000, 0x700000000)` as an import-stub frame, so nothing else may live there, and a guest
+  store to an unresolved variable must not be able to reach executable pages.
 - **`runtime_module_load.*`** — the guest asking for a module after boot, as opposed to the fixed
   preload list. True runtime PRX loading is not implemented (#639), which is why the preload list
   carries so many optional entries: they exist so a first P/Invoke resolves instead of hanging.
--- a/prosper/src/host/image/boot_program.cpp
+++ b/prosper/src/host/image/boot_program.cpp
@@ -323,7 +323,7 @@ bool boot_program(const std::string& d, Program& p, std::string* err,
     std::vector<LinkInput> in = boot_link_inputs(d);
 
     std::string e;
-    if (!link_program(in, BOOT_STUB, p, &e)) return fail("link failed: " + e);
+    if (!link_program(in, BOOT_STUB, BOOT_IMPORT_DATA, p, &e)) return fail("link failed: " + e);
     // Loud, self-describing report of every auto-linked plugin the linker refused: which module was
     // dropped, the exact NID that collided, and which already-linked module owns it.
     for (const auto& s : p.skipped_modules)
@@ -345,10 +345,10 @@ bool boot_program(const std::string& d, Program& p, std::string* err,
         if (p.aliased_exports.size() > shown)
             printf("export alias: ... and %zu more\n", p.aliased_exports.size() - shown);
     }
-    printf("linked %zu modules; %zu imports (%zu cross-module, %zu stub slots); %zu init fns; "
-           "%zu aliased exports\n",
+    printf("linked %zu modules; %zu imports (%zu cross-module, %zu stub slots, %zu data slots); "
+           "%zu init fns; %zu aliased exports\n",
            p.mods.size(), p.total_imports, p.resolved_cross_module, p.slots.size(),
-           p.init_fns.size(), p.aliased_exports.size());
+           p.data_slots.size(), p.init_fns.size(), p.aliased_exports.size());
 
     // Diagnostics: linking complete.
     diagnostics::record_boot_phase(diagnostics::BootPhase::LINKING);
@@ -400,6 +400,12 @@ bool boot_program(const std::string& d, Program& p, std::string* err,
     // is synchronised either way (dispatch_append_slots), this just avoids the copy.
     p.slots.reserve(p.slots.size() + 1024);
     if (!install_stubs(p.slots, p.stub_base, p.stub_size, &e)) return fail("stubs failed: " + e);
+    // The writable aperture behind every unresolved DATA import (#3529). Installed here rather than
+    // lazily: the guest reaches a data import during the very first init_array below (libc's own
+    // startup writes __stack_chk_guard), so the pages must exist before run_guest_inits runs.
+    p.data_slots.reserve(p.data_slots.size() + 128);
+    if (!install_import_data(p.data_slots, p.data_base, p.data_stride, &e))
+        return fail("import data failed: " + e);
     install_trap_handler();
 
     // Diagnostics: stubs and trap handler installed.
--- a/prosper/src/host/image/boot_program.hpp
+++ b/prosper/src/host/image/boot_program.hpp
@@ -51,6 +51,15 @@ inline constexpr uint64_t BOOT_STUB     = 0x600000000ull;
 // (Kept clear of the ad-hoc stub bases a few unit tests pass to install_stubs — 0x680000000,
 // 0x710000000, 0x720000000, 0x730000000 — so a future test that boots a program and loads a module
 // in one process cannot alias them.)
+// The import-DATA aperture (#3529): one writable, non-executable, zero-filled page per unresolved
+// STT_OBJECT import. Deliberately OUTSIDE [0x600000000, 0x700000000) -- callback_fs.hpp reads any
+// return address in that window as an import-stub frame, and exec_image.hpp's comment above
+// kStubApertureBytes says nothing else may be mapped in it. Also clear of the ad-hoc stub bases a
+// few unit tests pass to install_stubs (0x680000000, 0x710000000-0x730000000) and of the runtime
+// PRX pool below.
+inline constexpr uint64_t BOOT_IMPORT_DATA     = 0x7c0000000ull;
+inline constexpr uint64_t BOOT_IMPORT_DATA_END = 0x7d0000000ull;   // 256 MiB, as for the stubs
+
 inline constexpr uint64_t BOOT_RUNTIME_MODULE_BASE   = 0x800000000ull;
 inline constexpr uint64_t BOOT_RUNTIME_MODULE_STRIDE = 0x8000000ull;   // 128 MiB
 inline constexpr unsigned BOOT_RUNTIME_MODULE_SLOTS  = 48;
@@ -98,6 +107,11 @@ inline const char* guest_module_name(uint64_t a) {
     // which slot holds which module is a run-local fact, so the offset is reported modulo the
     // stride and the exact module comes from the loader's own [loadmod] line.
     if (a >= BOOT_RUNTIME_MODULE_BASE && a < BOOT_RUNTIME_MODULE_END) return "runtime-prx";
+    // Deliberately labelled but NOT added to guest_va_in_module below: the label makes a fault
+    // report say which unresolved variable was touched, while guest_va_in_module feeds
+    // guest_va_in_module_code, and this aperture is data. A stack scan must not mistake it for a
+    // guest return address.
+    if (a >= BOOT_IMPORT_DATA && a < BOOT_IMPORT_DATA_END) return "import-data";
     return "mapped/host";
 }
 inline uint64_t guest_module_offset(uint64_t a) {
@@ -121,6 +135,7 @@ inline uint64_t guest_module_offset(uint64_t a) {
     if (a >= BOOT_STUB          && a < 0x610000000ull)     return a - BOOT_STUB;
     if (a >= BOOT_RUNTIME_MODULE_BASE && a < BOOT_RUNTIME_MODULE_END)
         return (a - BOOT_RUNTIME_MODULE_BASE) % BOOT_RUNTIME_MODULE_STRIDE;
+    if (a >= BOOT_IMPORT_DATA && a < BOOT_IMPORT_DATA_END) return a - BOOT_IMPORT_DATA;
     return a;
 }
 // End of the import-stub aperture. install_stubs reserves this whole window (never more), so a stub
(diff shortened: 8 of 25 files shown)


mattias800 (commented):
## Independent review — head `b65aa214b53df8c6fd5c4e0cb9e99dd265a2ee22`

**REJECTED** on this head. Three blocking items: one test the branch breaks (now confirmed red in
CI), one rollback-path invariant break in the new runtime-load code, and a census whose numbers
describe a different population than the sentence around them claims — two of which are being
committed into permanent source comments.

The core of the change is right and I want to say so before the findings: the discriminator is real,
the aperture placement is sound, and arm 7 genuinely isolates the symbol type. Details under
*Verified* at the bottom.

---

## Blocking

### B1. The branch breaks `runtime_prx_load`, and every embedder that installs stubs but not the data aperture

`runtime_load_start_module` calls `append_import_data` **unconditionally**
(`prosper/src/host/image/runtime_module_load.cpp:313`), and `append_import_data`'s *first* statement
is the install-ordering guard (`prosper/src/host/image/exec_image_linux.cpp:3243`,
`prosper/src/host/image/exec_image_win.cpp:1336`):

```
if (!g_data_stride) return fail("append_import_data before install_import_data");
...
if (first_new == n) return true;      // linux:3247 / win:1340 -- unreachable when the guard fires
```

`g_data_stride` is set **only** by `install_import_data` (exhaustive grep: `exec_image_linux.cpp:3229`
/ `:3218`, `exec_image_win.cpp:1323` / `:1314`, nothing else).
`prosper/tests/host/image/test_runtime_prx_load.cpp:219` builds a minimal `Program` by hand and calls
`install_stubs` only — it never calls `install_import_data` and never runs `link_program`, so
`g_data_stride` is 0. Every runtime load in that test now returns `kEnomem`, and the test's own
`if (handle < kSceModuleHandleBase …) return 1` bails out.

That test is registered inside `if(UNIX)` / `if(NOT APPLE AND x86_64)` and is **not** gated on
`HAVE_GAME_DUMP` (`prosper/CMakeLists.txt:2689-2701`), so it runs on every Linux CI build. The Linux
job on this head has now reported it:

```
[loadmod] '/app0/prx/prosper_runtime_test.prx': append_import_data before install_import_data -> ENOMEM
315/456 Test #315: runtime_prx_load .................................***Failed    0.01 sec
99% tests passed, 1 tests failed out of 456
	315 - runtime_prx_load (Failed)
```

Note the ratio: **455 of 456 pass**, so this is the only breakage and the rest of the branch is
clean under CI's domain.

Two things to fix, and I'd like both:

1. **Make a no-op append a no-op.** Move `if (first_new == n) return true;` above the
   `!g_data_stride` guard in both `append_import_data`s. A module with no data imports must not
   depend on the aperture having been installed. As written, adding the data path silently made the
   aperture a hard prerequisite of *all* runtime loading, which is not what the change is for.
2. **Give the test the aperture**, since its comment calls it a "minimal booted program" and it
   should mirror what `boot_program` does — set `prog.data_base = BOOT_IMPORT_DATA;` and call
   `install_import_data(prog.data_slots, prog.data_base, prog.data_stride, &err)` beside the
   existing `install_stubs`.

And separately: **the runtime half of #3529 has no test at all.** The boot path is well covered by
arms 7/8; `runtime_load_start_module`'s `STT_OBJECT` branch, its dedupe against a boot-time slot, and
`append_import_data`'s extension are covered by nothing. Given B2 below lives in exactly that code,
an arm here is worth more than any of the ones already added — a runtime module importing a variable
an already-linked module imports, asserting the two land on the **same** address.

### B2. `abandon()` desynchronises `g_nid_to_data_slot` from `data_slots` after a successful append

`prosper/src/host/image/runtime_module_load.cpp:205-208`:

```
if (!data_appended && g_prog->data_slots.size() > data_slots_before)
    g_prog->data_slots.resize(data_slots_before);
for (const auto& nid : claimed_nids) g_nid_to_slot.erase(nid);
for (const auto& nid : claimed_data_nids) g_nid_to_data_slot.erase(nid);
```

The `data_slots` rollback is correctly gated on `!data_appended`. The **NID-map** rollback on the
next line is not. So on the two `abandon()` paths that run *after* `data_appended = true` — the
`map_image` failure at `:321-323` and the unwind/export index mismatch at `:354` — the slot stays in
`g_prog->data_slots` (mapped and seeded) while its NID is erased from the lookup map.

The next module that imports that NID then misses the map, allocates a **second** slot, and gets a
**different address for the same guest variable**. That is precisely the condition the code's own
comment at `:106-108` says must not happen ("or the two modules would read and write different
copies of what the guest believes is one object"). It is also the one place the analogy with the stub
table breaks down: a duplicate *stub* slot is cosmetic (double-counted in the unimpl census, as
`runtime_module_loader_init`'s comment says), a duplicate *data* slot is a correctness divergence.

Fix is three lines — put both rollbacks under one `if (!data_appended)`:

```
if (!data_appended) {
    if (g_prog->data_slots.size() > data_slots_before)
        g_prog->data_slots.resize(data_slots_before);
    for (const auto& nid : claimed_data_nids) g_nid_to_data_slot.erase(nid);
}
```

Yes, `map_image` failing here is rare and the index mismatch is documented as impossible. I'm calling
it blocking anyway because the fix is trivial, the invariant is stated explicitly in the same file,
and a wrong address for a shared variable is not something a later run would surface as itself.

### B3. The census counts a population the linker never sees — including in two committed comments

The PR body leads with:

> Over the 60 local dumps, three near-universal libc/libkernel variables reach an aperture:
> `__stack_chk_guard` (728 bindings, **60 of 60 titles**), one unnamed `libSceLibcInternal` object
> (182 bindings, 53 titles) and `__progname` (96 bindings, 60 titles).

"reach an aperture" is a statement about the loader's **link set**. These numbers are not measured
over it. Two independent routes:

1. **The repo's own `nid_census`, on `origin/main` (pre-PR, so `--data-only` is absent but
   `collect_modules` is identical)**, run over all 60 dump roots:
   `scope: 2284 distinct imported NIDs over 338 module(s) read, 0 unreadable`.
   **338 modules across the entire corpus.** A claim that one symbol appears in 596 or 728 of them is
   arithmetically impossible for that population.
2. **A from-scratch ELF/SELF symbol parser** I wrote for this review (reproduces prosper's
   `nid_hash` — it independently derives `Ou3iL1abvng` = `__stack_chk_fail`, the NID in #3529's
   own trace), reconstructing `boot_link_inputs` over the same 60 dumps: **329 modules**, and per
   symbol, unresolved `STT_OBJECT` bindings:

   | NID | name | link-set bindings | titles | PR body says |
   | --- | --- | ---: | ---: | --- |
   | `f7uOxY9mM1U` | `__stack_chk_guard` | **303** | 60 | 728 bindings, 60 titles |
   | `djxxOmW6-aw` | `__progname` | **60** | 60 | 96 bindings, 60 titles |
   | `ZT4ODD2Ts9o` | (unnamed) | **53** | 53 | 182 bindings, 53 titles |
   | `GAtITrgxKDE` | (unnamed) | **1** | 1 | — |
   | | **total** | **417** | | 1006 |

The **title** counts are exactly right (60 / 60 / 53 — I reproduce all three). The **binding** counts
are not, and I can say where they come from: a *recursive* scan of every `eboot.bin`/`*.prx`/`*.sprx`
in each dump gives **808 modules parsed, 728 of them importing `__stack_chk_guard`, 1007 unresolved
OBJECT bindings** — 728 and ~1006 to the unit. So the census was taken over every module shipped in
the dump, including the ones prosper never links (R-Type's per-stage PRXs and friends).

Why this is blocking rather than a nit: two of these figures are being written into **permanent
source comments**, where the charter's own "a stale figure carries the charter's authority" rule
applies —

- `prosper/src/loader/linker.hpp:79-80`: *"60 of the 62 local dumps import at least one STT_OBJECT
  symbol and `__stack_chk_guard` alone appears in 596 shipped modules (tools/nid_census
  --data-only)"* — 596 is a third number, agreeing with neither 728 nor 303, and larger than the
  338-module link set.
- `prosper/src/loader/AGENTS.md:25-29` carries the same sentence, and attributes it to a tool that,
  run the way the surrounding prose describes (`nid_census <dump root>` → `boot_link_inputs`),
  cannot produce it.

Also in that sentence: **"60 of the 62 local dumps"** — the dump library holds 60 `PPSA*` dumps plus
two directories that are not dumps. It is 60 of 60, which is a stronger claim anyway.

What I'd like: re-run `nid_census --data-only` over the 60 dump **roots** and quote that, in the PR
body and in both comments. The conclusion the change rests on survives unchanged and actually gets
sharper — **only four distinct NIDs in the whole corpus reach the new aperture**, which is a much
better argument that this is a low-risk change than an inflated binding count is.

---

## Non-blocking, but I'd like them addressed

**N1 — `prog.bound_data == prog.data_slots.size()` is not an invariant.**
`prosper/tests/misc/test_trap_linux.cpp` asserts it flatly. `bound_data` counts *bindings*,
`data_slots` counts *deduped NIDs*; they are equal only because no eboot imports one NID twice. The
sibling assertion right above it (`stubbed == slots.size()`) carries exactly that caveat, spelled out
and marked MEASURED. This one should inherit the caveat or it will read as derived.

**N2 — the keep-honest arm can vanish silently.** In the same test, `plain_slot` stays `SIZE_MAX`
when the eboot has only one data import, and the zero-check/store-retention block is then skipped
with no `CHECK`. PPSA24651 has four `STT_OBJECT` imports so it holds today, but the arm that keeps
the seed assertion honest should not be able to disappear without saying so — the charter's
prefer-experiments-that-detect-their-own-invalidity rule. One `CHECK(plain_slot != SIZE_MAX, …)`.

**N3 — the strongest new assertions never run in CI.** `test_trap_linux` is registered as
`trap_identifies_imports` under `if(HAVE_GAME_DUMP)` (`prosper/CMakeLists.txt:2674-2679`), so
everything that actually maps, seeds, reads and writes the aperture is local-only — and this head
quotes no local run. `import_data_seed` and `loader_synth_link` arms 7/8 do run in CI, and they are
good, but they never touch a mapping. See the ctest note at the end.

**N4 — the AMPR udata half has no test.** The upgrade rule in `prosper_eq_add_apr`
(`hle_kernel_time.cpp:1919-1930`) — idempotent, fills an empty slot from a non-zero udata, never
overwrites a real one with 0 — is exactly the kind of three-way rule that a 20-line unit test should
pin, and there are three registration paths feeding it (`k_add_ampr_event`, `apr_cb_set_equeue`,
`k_ampr_measure_write_equeue`). The reasoning in the comment is sound and I checked the delivery
paths (below); it just isn't guarded.

**N5 — two unrelated fixes in one PR.** The loader change runs for every title; the AMPR change is
one title's event delivery. They are separate commits today, which is good, but this repo
squash-merges (`gh pr merge --squash`), so the corpus-wide loader change becomes unbisectable from
the HLE change. Either land them as two PRs, or merge with the commits preserved.

**N6 — consider a guard page between data slots.** With `data_stride = 4096` and slots mapped
contiguously, a store past the end of one variable lands silently on its neighbour. A 8 KiB stride
with only the first page committed makes that fault instead. The aperture holds 65536 slots at the
current stride and the corpus uses **four**, so the headroom is free. Your `linker.hpp` comment
already reasons about "an over-long store inside its own slot instead of on its neighbour's" — a
guard page is the version of that which can't be wrong.

**N7 — `bind_imports_to_stubs` (`self/module.cpp:369`) still binds every import, data included, to
the stub aperture.** It is test-only (`tests/test_module.cpp:78`) so nothing is broken, but it is now
the one function in the tree that does the thing #3529 forbids, with no comment saying so. A line
saying "test fixture only; the real rule is link_program pass 2" would stop the next reader copying
it.

---

## Verified (traced, not assumed)

- **Symbol type plumbing is correct.** `struct Sym` is under `#pragma pack(push,1)`
  (`self/module.cpp:27-36`), so `st_info` is at +4 and `s.st_info & 0xf` is exactly
  `ELF64_ST_TYPE`. `m.syment` defaults to 24 and the entry is read from `DT_SYMTAB`/`DT_SCE_SYMTAB`,
  the same table `is_import` is derived from — so the type comes from the symbol that produced the
  import, not from a parallel table.
- **NOTYPE really does keep the old behaviour, and it is not a live concern here.** The branch is
  `if (imp.elf_type == STT_OBJECT)` — NOTYPE, FUNC, TLS, COMMON and everything else fall through to
  the stub, which is what the PR says. I also measured the premise rather than trusting it:
  across the link sets of all 60 dumps, **145,841 imports, 1,679 OBJECT, 144,162 FUNC, zero
  NOTYPE**. The "stripped PS5 modules report NOTYPE" worry does not materialise in this corpus.
  On PPSA25258's eboot specifically, `__stack_chk_guard` is `st_info & 0xf == 1`, i.e. the fix
  actually fires on the title it was found on.
- **Cross-module resolution still wins.** The export lookup and its `continue` precede the OBJECT
  branch (`linker.cpp:169-174`), and arm 8 fails if that order is inverted. Good arm — this is the
  regression that would have silently unbound every title whose own libc defines the variable, and
  arm 7 alone cannot see it.
- **Aperture placement is clean.** `[0x7c0000000, 0x7d0000000)` is outside `BOOT_STUB`'s window and
  outside `callback_fs.hpp:13`'s `[0x600000000, 0x700000000)` import-stub-frame range; clear of the
  test stub bases (`0x680000000`, `0x710000000`-`0x730000000`, max extent `0x740000000`); below the
  runtime-PRX pool `[0x800000000, 0x980000000)`; above the plugin auto-link pool, which tops out at
  `0x5f8000000`; and far below `kGuestAutoMapBase = 0x2000000000`. The guest's non-fixed
  `sceKernelReserveVirtualRange` search uses `MAP_FIXED_NOREPLACE` and skips occupied ranges
  (`hle_kernel_mem.cpp:1628-1648`), so it steps over the aperture rather than through it, exactly as
  it already does for the stub table. RW and never EXEC on both platforms; zero-filled by
  `MAP_ANONYMOUS` / fresh `MEM_COMMIT`. Over-capacity (>65536 slots) is a loud `fail()`, not a wrap.
- **Arm 7 really does isolate the type, which is the claim I most expected to fall over.** Traced
  through the fixture: in `data_arm` (`imports={kMissing}`, `data_imports={kVar}`) and `func_arm`
  (`imports={kMissing,kVar}`), kVar is symbol index **2** and GOT index **1** in both; `add_name`
  appends `#A#A` to every NID so the lib id is identical; `st_size` is 8 in both; and `GLOB_DAT` and
  `JUMP_SLOT` share one case in `apply_relocations` (`module.cpp:416-417`), both writing
  `sym_addr(r.sym)` with `is_plt` true. `st_info` is the only differing input. One nit: the comment
  says the mutation fails "its first four assertions" — the accounting identity
  (`total == cross + stubbed + bound_data`) still holds under it (2 == 0 + 2 + 0), so it's three.
- **The canary seed is placed and keyed correctly, and I don't read it as a special case the charter
  forbids.** It lives in `src/hle/dispatch/`, not the linker; it is keyed on `nid_hash(...)` of a
  published libc/libkernel symbol name, not on an address or a title id; and `__stack_chk_guard` is a
  platform variable whose entire contract is that two reads agree — derivable offline, which is the
  charter's own test. Hashing the name at first use instead of pasting the NID literal is the right
  call. `0x524550534F525000` is **non-canonical** (bits 63:48 = `0x5245`), so if a guest ever treats
  it as a pointer it faults at the dereference rather than corrupting something — better than zero
  and much better than a plausible address. Written only for that NID, size-checked, and
  `append_import_data` seeds `[first_new, n)` only, so a guest-written canary is never re-stamped.
- **Every `link_program` caller is updated**, all six, each with distinct stub and data bases:
  `boot_program.cpp:326`, `test_boot_linux.cpp:36`, `test_trap_linux.cpp:26`,
  `test_loader_synth_link.cpp` (×8), `test_loader_synth_reject.cpp:85,255`,
  `test_plugin_autolink.cpp:177-178`. Nothing passes `stub_base` for both, and the new argument is
  required rather than defaulted, which is the right call and is justified in the header.
- **`prosper_eq_apr_udata(p.eq, 0)` in the pointer dialect is correct**, which is not obvious — that
  batch worker is inside `if (id == 0)`, so the hard-coded 0 is the registration id for that channel
  by construction, matching the `e.ident = 0` it posts. No lock-order problem either: `lk` there is
  on `pq->mx`, is unlocked across the loop, and `prosper_eq_apr_udata` takes `state.mx`, which no
  caller holds while waiting on `pq->mx`. The id != 0 path resolves the udata as an argument
  expression, outside the `lock_guard` scope above it, as its comment claims.
- **`st_size == 0` for every undefined import** — the `linker.hpp` claim. Spot-checked on four dumps
  (PPSA25258, PPSA24651, PPSA04263, PPSA20800): 1011/612/1410/745 imports, histogram `{0: all}`. The
  comment is right, and it is load-bearing (it's why the stride can't be derived).

---

## On the omitted local ctest

The machine reason is legitimate and I'm not asking for a full run or anything GPU-adjacent. But B1
is the exact failure a **targeted, pure-CPU** run catches in one command, and the charter's ctest
rule ("quote the count alongside the exit code") is satisfiable without touching the GPU:

```
ctest --test-dir prosper/build-linux --no-tests=error --output-on-failure \
      -R 'runtime_prx_load|loader_synth_link|loader_synth_reject|import_data_seed|plugin_autolink|trap_identifies_imports'
```

None of those initialises Vulkan. `trap_identifies_imports` is the one that matters most, because
per N3 it is the *only* thing anywhere that exercises the mapping, the seed and the guest store, and
CI cannot run it — it needs the dump. I'd like that six-test result quoted on the corrected head.

Incidentally: this worktree's `prosper/build-linux` predates the branch (binaries dated 01:56,
sources 14:32, and `test_import_data_seed` was never built), so as far as this tree is concerned the
change has not been compiled locally at all. Possibly it was built elsewhere — just flagging what I
can see.

---

## Summary

**REJECTED** on `b65aa214b53df8c6fd5c4e0cb9e99dd265a2ee22`. Fix B1 (test + the guard ordering), B2
(one `if`), and B3 (re-measure the census over the link set, in the PR body and in both committed
comments), and I'll re-review. The design is right, the aperture is well placed, and arm 7 is a
genuinely good discriminator — none of that is in question. Please re-request review on the corrected
head rather than treating this as conditional approval.

mattias800 (commented):
## Delta review — head `6557b5cf8` (was `b65aa214b`), diff `b65aa214b..6557b5cf8`

**REJECTED**, narrowly, on `6557b5cf88b2248b2755a9a14da3d14ee399c543`. B1 and B2 are fixed correctly
and I have nothing further on either. B3 is fixed in substance but the correction **introduces one
new wrong number**, in the very paragraph whose subject is not repeating unverified numbers — and it
is wrong in the same way the original was. That is a wording fix plus one tool run, not another round
of design.

Answers to all four of your questions below, including the one about whether to re-derive: **yes, and
I can now name exactly why, in both directions.**

---

## B1 — the guard is right, and narrower would be worse

I traced every case of `if (first_new == n && first_new == g_ndata) return true;`
(`exec_image_linux.cpp:3250`, `exec_image_win.cpp:1343`).

With **no aperture** (`g_data_stride == 0`, `g_ndata == 0`):

| `n` | `first_new` | old | new | right? |
| 0 | 0 | fail | **true** | yes — this was B1 |
| 2 | 0 | fail | falls through → stride guard → **fail** | yes, loud |
| 2 | 2 | fail | `first_new != g_ndata` → falls through → stride guard → **fail** | yes, loud |

With an aperture installed, every path is bit-identical to before: `n == first_new == g_ndata`
returned true at the bottom and now returns true at the top; `n == first_new != g_ndata` failed
"slot table is not an extension" and still does.

So the answer to your question is **the `first_new == g_ndata` conjunct is exactly what keeps it from
widening too far**, and it is the part that must not be dropped. Without it, a caller whose table
disagrees with ours (`n == first_new != g_ndata`) would get silent success instead of the extension
error. Narrower alternatives — `if (n == 0) return true;` — would fix the test too, but they answer a
narrower question than the one the call site asks ("is there anything to do?"), and they would leave
the aperture a prerequisite for a post-boot module that adds no *new* data slots. Keep what you have.

Two follow-ons, neither blocking:

- **The bottom `if (first_new == n) return true;` is now dead code.** To reach it you must have
  passed `if (first_new > n || first_new != g_ndata) return fail(...)`, so `first_new == g_ndata`
  holds; if `first_new == n` also holds, the new top guard already fired. Harmless, but either delete
  it or note that it is subsumed — otherwise someone later relaxes the top guard believing the bottom
  one still covers them.
- **`runtime_module_load.cpp:290-291` computes `g_prog->data_base + slot * g_prog->data_stride`
  without checking `data_base`.** `link_program` refuses a zero `data_base`, but a hand-built
  `Program` (exactly what `test_runtime_prx_load` constructs) has `data_base == 0`, and the branch
  would then bind a guest variable near address 0. I traced it and today it is safe: the append fails
  loudly immediately after, `abandon()` returns before `apply_relocations` and `map_image`, and the
  bogus addresses die with the local `img`. Safe by one step of downstream luck rather than by
  construction, though — a `if (!g_prog->data_base) return abandon(kEnomem);` at the top of the
  branch would fail at the cause.

Your mutation check (restore the order → `runtime_prx_load` red, ctest exit 8, 0% of 1; corrected
order → green) is the right shape, and it is the arm that shows the lever moved.

## B2 — the asymmetry is correct, not a bug. But I found a different one next to it.

**Keep it as you have written it**, and the reason is worth a line in the code because it is not
obvious from the shape: **a stub slot is stateless, a data slot holds the guest's value.** A duplicate
stub slot emits an identical trampoline to the identical handler — the only cost is double-counting
in the unimplemented-call census, which `runtime_module_loader_init`'s own comment already names. A
duplicate data slot is two addresses for one variable. So the rollback asymmetry tracks a real state
asymmetry; making the stub map roll back conditionally too would be symmetry for its own sake.

**However**, checking that led me to a genuinely pre-existing hazard in the stub path that your new
data code correctly avoids, and I think it deserves its own issue rather than a fix here:

`dispatch_append_slots(&g_prog->slots, new_slots)` (`runtime_module_load.cpp:307`) grows
`g_prog->slots` **before** `append_stubs` runs, and `append_stubs` sets `g_nstubs = n` only on
success (`exec_image_linux.cpp:3191`). If `append_stubs` fails at the mmap extension or at
`emitted > stub_size`, `g_prog->slots` has grown and `g_nstubs` has not — and every subsequent
runtime load then fails `first_new != g_nstubs` → `"append_stubs: slot table is not an extension"`,
**permanently**. The file's comment at `:310-312` justifies leaving the appended slots in place, but
it addresses removal, not the counter divergence. Your `data_appended` gate is precisely the
construct that stops the data path having the same problem, so the new code is strictly better than
the code it was modelled on. Not this PR's job — please file it.

## B3 — substance right, but the correction carries a new wrong number

### The new error

Both comments now say:

> "…a recursive scan of every module file present in every dump (808 modules). `boot_link_inputs`
> links roughly 303 of them…" — `src/loader/AGENTS.md:30-33`
> "…808 of them. `boot_link_inputs` links about 303…" — `src/loader/linker.hpp:83-86`

**303 is not a module count.** It is the `__stack_chk_guard` *binding* count from my table — you took
the binding ratio (728 / 303 = 2.4, which is right) and then wrote the binding count where the module
count goes. "of the 808 modules, `boot_link_inputs` links roughly 303 of them" is a module claim and
it is false.

The module count is **338**, and it is the one figure in this whole exchange I measured with the
repo's own tool rather than my reconstruction — `nid_census` on `origin/main` over all 60 dump roots:

```
scope: 2284 distinct imported NIDs over 338 module(s) read, 0 unreadable
```

My independent reconstruction gives 329. Either way it is ~338, not ~303. Please make it
"links 338 of them".

I am holding the verdict on this one small thing because of what the paragraph is *for*. It exists to
tell the next reader "quote link-set figures, not disk figures", and it currently mis-states the
link-set module count by using a binding count — the same category error, one layer down, in the
correction. A paragraph that teaches the lesson while committing it is worse than no paragraph.

### Your question: should you re-derive? Yes, and here is the specific reason

You were right to ask, and I am not going to wave it through on the grounds that the numbers are
mine. My figures were more than strong enough to **falsify** 596 and 728 — 596 alone exceeds the
entire 338-module link set, so that conclusion is robust by two independent routes. They are **not**
strong enough to be the replacement constants, because I can now name an error in my reconstruction
of `boot_link_inputs` in *each* direction:

- **It under-counts modules.** 329 against the tool's 338. I verified this is not silent parse
  failure — my parser succeeded on 329 of 329 candidates, so the nine are modules my reconstruction
  never offered.
- **It over-counts titles for one NID.** I linked `sce_module/libSceNpCppWebApi.prx` unconditionally.
  The real `boot_link_inputs` ends with `drop_unimported_support_modules` (`boot_program.cpp:311`),
  and that module is `only_if_imported` (`:242`). It is also the module that carries the third NID: I
  traced `ZT4ODD2Ts9o` to `sce_module/libSceNpCppWebApi.prx`, raw `ZT4ODD2Ts9o#F#G`, resolving through
  DT_SCE_IMPORT_LIB to **libSceLibcInternal** — so your library attribution is correct. 53 titles ship
  that PRX; a sibling vouches for it in **51**, so the "53" in both comments should be about 51, and
  my number was an upper bound.

So `417`, `303` and `53` all carry my reconstruction's error bars. `--data-only` now exists and this
is one command over the 60 dump roots. Quote the tool's own output and every figure in the comment
becomes reproducible by the reader the comment is addressed to, which is the whole point.

### Two things I did verify, which the comments can use

- **`__stack_chk_guard` and `__progname` are both imported from `libkernel`, not libc.** Resolved
  through DT_SCE_IMPORT_LIB on PPSA25258 / PPSA24651 / PPSA04263 / PPSA03831 eboots and on
  `sce_module/libc.prx`. `import_data_seed.hpp` already calls it "a libkernel variable" — confirmed,
  and that matters because it is the sentence that makes seeding it reimplementation rather than
  invention.
- **The fourth NID is `GAtITrgxKDE`, from `libSceNet`**, one title (PPSA25258), one binding. Worth
  naming alongside the other three: "exactly four distinct NIDs" is a much stronger risk argument
  when all four are named and three of them are libkernel/libc internals.

## `nid_census` scope — my recommendation: not a `--scope` flag

You asked for my call, so: **the tool's default is already the link set and was never the problem.**
`collect_modules` routes a directory through `boot_link_inputs` and has done since #2199, unchanged
by this PR. `nid_census <dump root>` cannot produce the recursive figure. It comes from passing
**module paths individually**, where `const std::string title = fs::path(input).filename().string()`
makes every `.prx` its own "title" and the cross-module exclusion degenerates to per-module — which is
also why the recursive scan reports *zero* cross-module satisfaction and so inflates the binding
count.

A `--scope` flag would add a supported way to produce the number we just spent a review retracting.
What I would do instead, both cheap and both aimed at the actual failure — a module census that
*looks* like a corpus census:

1. **Say what each input resolved to.** One line per input at the top: `input <name>: link set,
   N modules` versus `input <name>: single module`. The scope block reports totals today and never
   says which mode produced them, so the two reports are indistinguishable after the fact — which is
   exactly how the figure survived into a comment.
2. **Stop `--data-only`'s per-input breakdown reading as per-title.** Its lines are keyed on the
   input filename and the summary says "over X of Y input(s)". When the inputs are modules that reads
   as titles and is off by the module-to-title ratio (~5.6 here). Either label the printed column
   `input` explicitly, or suppress the "of Y input(s)" summary when any input resolved as a single
   module.

## The two deliberately-open items

- **Branch split — do not block, but make it a merge-time decision rather than nothing.** Merge
  preserving both commits rather than `--squash`, so a future bisect over a loader regression cannot
  land on a commit that also changed AMPR event delivery. If the merge path here is squash-only,
  cherry-pick the AMPR commit to its own PR instead. Either is fine; say in the PR body which you did.
- **AMPR udata test — do not block, but file the issue.** Three call sites feed
  `prosper_eq_add_apr` (`k_add_ampr_event`, `apr_cb_set_equeue`, `k_ampr_measure_write_equeue`) and
  the three-way rule (idempotent / fill an empty slot from a non-zero udata / never overwrite a real
  udata with 0) has no coverage at all. Missing coverage is invisible to every gate, which is what
  the charter's durable-queue rule is for. Two minutes.

## What to do

1. `303` → `338` in `src/loader/AGENTS.md:32` and `src/loader/linker.hpp:85`.
2. One `nid_census --data-only` run over the 60 dump roots; quote its figures for the bindings, the
   NID count and the per-title counts — or, if you would rather not spend the run, label the three
   numbers as a reconstruction and add the `only_if_imported` caveat so the next reader knows the
   error bars exist.
3. Optionally name the fourth NID and correct "libc/libkernel variables" to libkernel where it
   applies.

Do 1 and 2 and I will approve on sight — there is nothing left in the code half. Please re-request
on the corrected head rather than reading this as conditional approval; I would like to check the
final wording of a comment whose subject is the accuracy of comments.

mattias800 (commented):
## Delta review — head `8cee4933a`, diff `6557b5cf8..8cee4933a`

**APPROVED** on `8cee4933aae09f42393b443461ad854ebc70eac4`.

You asked me to say plainly whether there is any number in that paragraph I would not stand behind.
There is not, and I did not take that on the tool's word either — I reproduced the whole census by an
independent route and it agrees to the unit on every quantity.

## Reconciliation: two routes, six quantities, exact agreement

My earlier figures (329 modules, 417 bindings) were a reconstruction of `boot_link_inputs` in Python,
and I told you they carried error bars in both directions. I have now found and fixed both defects in
my own reconstruction, and it lands exactly on the tool:

| quantity | my parser, corrected | committed comment |
| --- | ---: | ---: |
| modules in the link sets | 338 | 338 |
| unresolved OBJECT bindings | **424** | 424 |
| satisfied cross-module | **1566** | 1566 |
| total OBJECT bindings | 1990 | 1990 |
| `f7uOxY9mM1U` `__stack_chk_guard` | 312 bindings, **60** titles | 60 of 60 |
| `djxxOmW6-aw` `__progname` | 60 bindings, **60** titles | 60 of 60 |
| `ZT4ODD2Ts9o` (libSceLibcInternal) | 51 bindings, **51** titles | 51 |
| `GAtITrgxKDE` (libSceNet) | 1 binding, **1** title | 1 |

The two defects, both mine, for the record — they are also the reason the earlier gap existed and why
a reconstruction is never a substitute for the tool:

- **I matched `Media/Modules/Il2cppUserAssemblies.prx` case-sensitively.** `boot_link_inputs` runs
  every path through `resolve_host_path_case`, and **11 titles** ship it spelled
  `Il2CppUserAssemblies.prx` (PPSA01885, 02664, 02777, 02801, 02846, 02849, 03416, 08576, 09804,
  13579, 25009). Those are large IL2CPP modules, which is why 11 missing modules moved the OBJECT
  total by 311 rather than by ~55.
- **I linked `sce_module/libSceNpCppWebApi.prx` unconditionally.** `boot_link_inputs` ends with
  `drop_unimported_support_modules` (`boot_program.cpp:311`) and that input is `only_if_imported`
  (`:242`); 53 titles ship it, a sibling vouches in 51.

11 added − 2 dropped = the +9 that turned 329 into 338, and 424 − 417 = the +7 that follows from it.
So the number that changed is mine, not yours, and your instinct to go to the tool rather than to my
table was the correct one on both occasions you exercised it.

I also independently confirm the rest of the block:

- `scope: 2284 distinct imported NIDs over 338 module(s) read, 0 unreadable` — I reproduced this line
  verbatim from the repo's own **pre-PR** `nid_census` binary over the 60 dump roots, so it is not
  downstream of any code this PR changes.
- All four library attributions, resolved through `DT_SCE_IMPORT_LIB` rather than guessed:
  `__stack_chk_guard` and `__progname` are **libkernel** (checked on the PPSA25258 / PPSA24651 /
  PPSA04263 / PPSA03831 eboots and on `sce_module/libc.prx`), `ZT4ODD2Ts9o` is **libSceLibcInternal**
  imported by `sce_module/libSceNpCppWebApi.prx` (`raw ZT4ODD2Ts9o#F#G`), `GAtITrgxKDE` is
  **libSceNet** on PPSA25258.
- "every title in the corpus has at least two" — holds, and tightly: the two libkernel NIDs are
  60-of-60 each, so the floor is exactly those two.
- The libkernel attribution matters beyond bookkeeping: it is what makes
  `import_data_seed.hpp`'s "libkernel is prosper, so supplying it is reimplementation" the correct
  framing rather than a convenient one. Good that you re-derived it with the hasher instead of taking
  it from me.

**One thing I could not reproduce, and I am flagging it rather than letting it pass because this
paragraph's whole subject is unverified numbers.** The retraction says "596 shipped modules" came
from the module-paths-as-inputs mode. I can reproduce that *mode* exactly — a recursive scan is 808
modules, of which **728** import `__stack_chk_guard`, and with cross-module exclusion degenerate that
is also 728 bindings, matching the PR body's original 728 to the unit. I cannot get **596** out of it
by any filter I tried. So the mechanism attribution is right and the 808 is right; the specific
provenance of 596 is a hypothesis. Not blocking — it is a retracted figure in a narrative, and the
load-bearing content ("the population is 338, not 808") is verified twice. If you want the paragraph
to be exactly as rigorous as it asks its readers to be, "…came from a recursive scan of every module
present in every dump (808 of them); the exact figure has not been reproduced" costs six words and
closes it. Your call, and I will not re-review for it.

## Your three deferred items — I agree with all three

- **`nid_census` input-scope labelling: you are right and I was wrong to raise it in this round.** A
  tool behaviour change does not belong in a comments-only head, and "it would have prevented this"
  is an argument for doing it properly, not for slipping it in. Please file it; it is the durable fix
  for the class, and the comments you just wrote are the workaround.
- **Branch split:** merging with both commits preserved is exactly what I asked for. Nothing further.
- **AMPR udata test:** an issue is the right disposition.

Three more for that issue sweep, all non-blocking and all raised earlier:

1. The **`append_stubs` / `g_nstubs` wedge** (`runtime_module_load.cpp:307` grows `g_prog->slots`
   before `exec_image_linux.cpp:3191` advances `g_nstubs`; a failed `append_stubs` then refuses every
   later runtime load permanently). Pre-existing, not yours, and your `data_appended` gate is the
   construct that stops the data path inheriting it.
2. The now-**dead** `if (first_new == n) return true;` at the bottom of both `append_import_data`s.
3. A **`data_base != 0`** check at the top of `runtime_module_load.cpp`'s `STT_OBJECT` branch, so a
   hand-built `Program` fails at the cause rather than one step downstream.

## State at approval

Head `8cee4933a` is the branch tip (`git ls-remote` confirms), and all 13 check runs on that exact
SHA are green or deliberately skipped — 11 success, `Progress tracker` and `Publish Release` skipped,
none pending, none failed. That includes the `Linux` job, which is the one that caught B1.

Across the three rounds: the symbol-type discriminator is right and measured (zero NOTYPE imports in
145,841 across the corpus, so the branch is on solid ground); the aperture is placed clear of every
other window and is RW/never-EXEC; arm 7 genuinely isolates `st_info` with everything else held
fixed; the canary is keyed by published NID at the HLE layer and is non-canonical so a stray
dereference faults; B1 and B2 are fixed correctly and B1's fix is mutation-proved; and the census now
comes from the tool and survives independent replication.

**APPROVED** — `8cee4933aae09f42393b443461ad854ebc70eac4`.

Issue #1748, opened 2026-08-02
Asterix & Obelix - Babylon Mission (PPSA30490): the guest heap grows ~210 MiB/s and is never released, exhausting the direct-memory pool in ~125 s (the tlsf_add_pool failure and 0x28 fault are symptoms)


mattias800:
## Summary

*Asterix & Obelix - Babylon Mission* (`PPSA30490`, Unity 6 / IL2CPP) stops publishing frames after
~125 s because **the guest's own heap grows without bound — about 210 MiB/s, never released — until
prosper's direct-memory pool is genuinely empty.** The `tlsf_add_pool` rejections and the
`SIGSEGV at addr=0x28` reported on #1599 are both *downstream symptoms* of that exhaustion.

This is filed separately from #1599 because that issue is about the black frames (the MSAA
`image_load` recompiler gap). The two are independent, as #1599's own boot-sweep comment says.

**Correcting the sweep's framing:** prosper is not denying the guest memory it should have had, and
it is not handing back a pool handle it then rejects. The pool is really empty, and the guest really
asked for all of it.

## Measured

Linux, `boot_trace`, `PROSPER_GUEST_FS=1 PROSPER_GUEST_ARGS=-force-gfx-direct PROSPER_MEMLOG=1`
(no renderer — the stall reproduces without one):

| | 16 GiB pool (default) | 12 GiB pool (`PROSPER_DMEM_BUDGET_MB=12288`) |
| --- | --- | --- |
| 16 MiB blocks allocated | 1053 | 797 |
| `sceKernelReleaseDirectMemory` calls | **8** | **8** |
| `alloc_main_dmem -> ENOMEM` | 79 | 79 |
| `tlsf_add_pool` rejections | 66 | 66 |
| Outcome | `SIGSEGV addr=0x28` | `SIGSEGV addr=0x28`, identical backtrace |

The allocation pattern is a single unbroken sequence from the third second of the boot to the crash:

```text
[memhle] reserve ENTRY hint=0x0 len=0x1000000 flags=0x0 align=0x40000
[memhle] reserve -> 0x23fb100000 len=0x1000000 align=0x40000
[memhle] alloc_main_dmem len=0x1000000 -> phys=0x409248000
[memhle] map_dmem -> 0x23fb100000 len=0x1000000 phys=0x409248000 prot=0xf2 flags=0x10
```

A timestamped run measures **13.0 ± 0.5 blocks/s ≈ 208 MiB/s**, flat, for as long as the title runs
[...]

PR #3542, opened 2026-09-11
agc: register sceAgcDcbDrawIndirect and build its real packet (#2929)

### LINKED ISSUE #2929: sceAgcDcbDrawIndirect (1q1titRBL6o) is unregistered while its indexed sibling is — every non-indexed indirect draw silently disappears
## Summary

`sceAgcDcbDrawIndirect` (NID `1q1titRBL6o`) has **no registered handler**, so every call falls to
`prosper_on_unimpl` and returns 0. Its indexed sibling `sceAgcDcbDrawIndexIndirect`
(`t1vNu082-jM`) **is** registered — `prosper/src/hle/graphics/hle_agc.cpp:3468`,
`agc_dcb_draw_index_indirect` at `:3079`. The non-indexed form is simply missing from the same
registration block.

The consequence is the charter's "unsupported GPU operation is a FATAL gap" case in its quietest
form: the guest asks prosper to append a `DRAW_INDIRECT` packet to its DCB, prosper appends
nothing, returns 0 (which the AGC builders read as success), and the draw disappears from the
command stream with no reject line and no `[compute] skip`. Nothing in the log says a draw was
lost; only the one-shot `[prosper] unimplemented:` line names it.

## Evidence

Observed live on two titles during the never-booted survey (frontend `tools/screenshot`,
default route, `PROSPER_RENDER=1 PROSPER_GUEST_ARGS=-force-gfx-direct`, Linux/RADV):

- **Little Nightmares II** (`PPSA02154`, UE4) — first call at ~100 s into a 120 s run:
  `[prosper] unimplemented: libSceAgc::1q1titRBL6o  -> returning 0`
- **Sifu** (`PPSA03001`, UE4) — same line on a 140 s default run.

## Blast radius

The NID string is present in **49 of the 54 dumps** in the corpus, scanning `eboot.bin` **and**
every `.prx`/`.sprx` with `grep -a` (instrument trap 218). Validated against a known positive
(`PPSA02154-app0/eboot.bin`) before the count was believed.

**An import-table entry is not a call**, so 49 is an upper bound on exposure, not a count of
affected titles. Two titles are confirmed live callers.

## Suggested fix

Register the non-indexed indirect draw beside its indexed sibling and give it a
round-trip/execution test that asserts the emitted packet, exactly as
`agc_dcb_draw_index_indirect` has. The two differ only in whether an index buffer participates, so
the packet builder is close to the existing one — but derive the dword count from the guest's own
reservation rather than assuming it matches the indexed form (`docs/AGC_PACKET_SIZES.md`: a
builder's dword count is an ABI contract with the guest).

Found by the never-booted survey (`prosper/docs/NEVER_BOOTED_SURVEY_2026_08.md`). Not fixed here.

mattias800:
Fixes #2929.

## The failure scenario

`sceAgcDcbDrawIndirect` (NID `1q1titRBL6o`) had **no registered handler**, so every call fell to
`prosper_on_unimpl` and returned 0. The AGC builders read a 0 return as "no packet was appended",
and nothing downstream reports that: the draw left the guest's command stream with no reject line,
no `[compute] skip`, and no decode warning. The only trace anywhere was a single one-shot
`[prosper] unimplemented: libSceAgc::1q1titRBL6o -> returning 0`.

The indexed sibling `sceAgcDcbDrawIndexIndirect` (`t1vNu082-jM`) was registered, so the two forms
diverged silently. 49 of the 54 corpus dumps import the NID (an import is not a call, so that is an
upper bound); Little Nightmares II `PPSA02154` and Sifu `PPSA03001` are confirmed live callers per
the issue.

## The behavioral contract

`sceAgcDcbDrawIndirect(dcb, byteOffset, shaderDrawModifier)` appends a draw whose vertex and
instance counts are read from GPU memory at `graphicsArgumentBase + byteOffset`, where the base was
selected by a preceding `sceAgcDcbSetBaseIndirectArgs(dcb, 0, addr)`. The operand list is identical
to the indexed sibling's. **What differs is at the far end of the pointer**: the argument buffer
holds the four-dword non-indexed form — `vertexCount, instanceCount, startVertex, startInstance` —
where the indexed form holds five (`indexCount, instanceCount, firstIndex, vertexOffset,
firstInstance`) and binds an index buffer.

That is why this is a distinct operation rather than a flag on the existing one. Reusing
`R_DRAW_INDEX_INDIRECT` would have made the executor read a **fifth dword the guest never wrote**
and then reject the draw for having no index base — the same silent disappearance, one layer down.

## Approach and invariants

- **Builder** (`hle_agc.cpp`): `agc_dcb_draw_indirect` emits a 4-dword `IT_NOP` packet under a new
  sub-op `R_DRAW_INDIRECT = 0x25` — `cmd[1]` = byte offset, `cmd[2..3]` = the 64-bit
  `ShaderDrawModifier`, low dword first. `sceAgcDcbDrawIndirectGetSize` (`cxPZ4Wgvdj8`) is
  registered too and answers from the same `kDwDrawIndirect` constant, so builder and reservation
  cannot drift apart.
- **The dword count is derived, not copied from the indexed row.** A builder's count is an ABI
  contract with the guest's own reservation (`docs/AGC_PACKET_SIZES.md`), and the asymmetry is what
  matters: emitting **fewer** dwords than the real AGC function wastes reserved space, emitting
  **more** overruns a reservation the guest made in good faith (#1748). Two independent arguments
  for 4:
  - prosper's payload is the API's own operand list (32-bit offset + 64-bit modifier), so the packet
    cannot be *smaller* than 4.
  - The hardware packet it stands for, PM4 `DRAW_INDIRECT`, spends header + `DATA_OFFSET` +
    `BASE_VTX_LOC` + `START_INST_LOC` + `DRAW_INITIATOR` = 5 — the same field list as
    `DRAW_INDEX_INDIRECT`. So 4 is at or below the reference, and cannot overrun.

  `CONFIDENCE: MED` on exact equality with the real library (no local dump was observed inlining the
  size, so equality is inferred from the published layout rather than measured from a guest
  reservation); `CONFIDENCE: HIGH` on the upper bound, which is the load-bearing half. Recorded as a
  new row in the `docs/AGC_PACKET_SIZES.md` builder table and as an entry under its `## Open`,
  naming `PROSPER_DCBWIN` over a live caller as the cheap way to settle the exact value.
- **Decoder** (`pm4_decode.*`): new `Kind::DrawIndirect`, same payload extraction as the indexed
  form, distinct kind.
- **Command processor**: the two kinds share one arm, and `Draw::indexed` now carries the *packet's
  own identity* (`c.kind == K::DrawIndexIndirect`) rather than a hard-coded `true`. A non-indexed
  indirect draw does not inherit a bound index buffer. The draw census counts the new kind.
- **Executor** (`resolve_indirect_draw_arguments`): a non-indexed source reads **4** dwords (not 5 —
  reading 5 would touch bytes past what the guest wrote) and resolves to the `DrawIndexAuto` shape
  the realizer already renders: vertex count in `index_count` with `indexed` false, and
  `startVertex` carried through the existing per-draw vertex-offset override, which the realizer
  hands to `vkCmdDraw`'s `firstVertex`. Same bounds, zero-count no-op and `firstInstance != 0`
  guards as the indexed path.

## Deliberately unaffected

- The indexed path (`sceAgcDcbDrawIndexIndirect`) is byte-for-byte unchanged: same sub-op, same
  4-dword packet, same five-dword argument read, same index-base requirement. A test asserts its
  payload is identical to the new form's while its sub-op differs.
- `sceAgcDcbDispatchIndirect` / `sceAgcAcbDispatchIndirect` untouched.
- `sceAgcDcbDrawIndirectMulti` (`kUlvghKs-mA`) and its GetSize remain unregistered — a different
  operation with a count/stride, out of scope here.
- No `PROSPER_*` switch added; the new behavior is unconditional, as an unregistered NID being
  registered must be.

## Risks

- **Dword count.** The one way this could hurt a title is by being *larger* than the real AGC
  packet. Argued above to be at or below it; the only unproven part is exact equality, which is
  harmless in the direction it can be wrong.
- **A previously-absent draw now executes.** A title that was silently dropping these draws will
  start submitting them. That is the point, but it is a behavior change on any title that calls the
  NID, and the new draws go through the same realization and validation as every other indirect
  draw.
- **Sub-op value.** `0x25` was free (`R_NUM` is `0x40`; the previous maximum was `0x24`). Both the
  `hle_agc.cpp` mirror and `pm4_decode.hpp` were updated together; the packet-bytes test pins the
  encoding.

## Build and test

Environment note, stated because it bounds every claim below: this WSL cannot build prosper's
Vulkan chain — its `libvulkan-dev` is 1.3.275 and `frontends/shared/device/vulkan_runtime.hpp:11`
needs `VK_API_VERSION_1_4`. 29 targets (`prosper_live_renderer` and the render/storage test
executables) fail to compile with `'VK_API_VERSION_1_4' was not declared in this scope` and friends;
every one of the compile errors in a keep-going full build is a Vulkan 1.4 header symbol, none in
a file this PR touches. The Vulkan *loader* is 1.3.275 as well, so the Vulkan-execution tests cannot
run here either (`[render] Vulkan 1.4 required; loader reports 1.3.275`).

```
cmake -S prosper -B prosper/build-linux -DCMAKE_BUILD_TYPE=Release \
      -DGAME_DUMP=<DUMP_ROOT>/PPSA24651-app0
cmake --build prosper/build-linux -j5 -- -k
```

**`ctest --no-tests=error`: 458 tests discovered, 346 passed, 112 failed, exit code 8.**

That red number is entirely environmental, and it is established by comparison rather than by
assertion: the same configure/build/ctest was run on a detached worktree at **`origin/main`
(3947f269)**, which gives **458 discovered, 112 failed, exit 8** — and `diff` of the two sorted
failing-test lists is **empty**. Identical set, identical statuses. This branch introduces no new
failure. (68 of the 112 are `Not Run` — the unbuilt Vulkan executables; the rest are Vulkan-runtime
tests plus two `/mnt/c` artifacts: a CRLF-vs-LF manifest byte comparison in `snapshot_guard_logic`
and a DrvFs `0777`-vs-`0755` mode check in `profiling_access`.)

The four tests this PR extends, from that same run:

```
219/458 Test #219: pm4_decode_stream ........... Passed
220/458 Test #220: command_processor_apply ..... Passed
328/458 Test #328: agc_dcb_pm4 ................. Passed
329/458 Test #329: agc_getsize_drift ........... Passed
```

### Tests added

- `tests/gpu/pm4/test_agc_dcb.cpp` — **asserts the emitted packet bytes**: return value, a cursor
  advance of exactly 4 dwords, the `PM4(4, IT_NOP, R_DRAW_INDIRECT)` header, `cmd[1]` = offset,
  `cmd[2..3]` = modifier low-then-high, and `GetSize == 16`. Plus the discriminating arm: the
  indexed sibling built with the *same* operands must produce an identical payload and a **different
  header**, which is the only assertion that would catch a shared sub-op.
- `tests/hle/test_agc_getsize.cpp` — a `DrawIndirect` row in the drift table, pinning the absolute
  count at 4 (not merely "GetSize agrees with the builder", which stays green at any value).
- `tests/gpu/pm4/test_pm4_decode.cpp` — the builder is driven through the NID registry into a real
  DCB alongside the indexed form and the stream is decoded: 18 packets, and op14 is
  `K::DrawIndirect` with its own offset and modifier.
- `tests/gpu/pm4/test_command_processor.cpp` — folds a non-indexed indirect draw **with an index
  buffer bound** and asserts it retains base+offset and modifier and is `!indexed` with
  `index_base == 0`.
- `tests/gpu/execute/test_gpu_execute.cpp` — end-to-end: four-dword arguments `{3, 2, 9, 0}` resolve
  to a real non-indexed draw reaching the submit renderer with `raw_draw_count 3`,
  `instance_count 2`, `vertex_offset 9`, `raw_indexed false`. `startVertex = 9` is chosen so a
  five-dword misreading of this buffer gives a different, checkable answer (`vertex_offset 0`).

### Without-fix arm

Ran explicitly, not assumed. With the two `RN(...)` registrations commented out and nothing else
changed, rebuilt and re-run:

| test | with fix | without fix | failing assertion |
| --- | --- | --- | --- |
| `agc_dcb_pm4` | exit 0 | **exit 1** | `#2929: sceAgcDcbDrawIndirect + its GetSize are registered beside the indexed sibling` |
| `pm4_decode_stream` | exit 0 | **exit 1** | `AGC Dcb builders registered` |
| `agc_getsize_drift` | exit 0 | **exit 1** | `DrawIndirect: GetSize + builder registered` |
| `command_processor_apply` | exit 0 | **exit 1** | `#2929: sceAgcDcbDrawIndirect is registered alongside the indirect base builder` |

Each fails on exactly one assertion — the new one — and on nothing else.

## What I could not verify

- **`test_gpu_execute` was compiled but never executed.** It cannot build or run in this
  environment (above). To get *some* evidence rather than none, the translation unit was compiled
  and linked out-of-tree against a Vulkan 1.4 header set with the project's own include set and link
  line — object and link both exit 0, so the added block compiles and its types/APIs resolve — but
  the binary then dies at `[render] Vulkan 1.4 required; loader reports 1.3.275` before reaching any
  of prosper's executor assertions, including the five that already pass on every other machine. So
  the new executor arm is **compile-verified only**; CI and a Vulkan-1.4 host are where it first
  actually runs.
- **No live-title evidence.** No title was booted — the renderer cannot be built here — so there is
  no screenshot and no before/after on Little Nightmares II or Sifu, and consequently no `BLOG.md`
  entry. The issue's own `CONFIDENCE: LOW` temporal correlation (PPSA02154's transition to a
  permanent flat-white clear bracketing the `unimplemented` line) is **not** tested by this PR and
  remains open; the issue notes re-running `--seconds 13 --count 30` on the default route settles it
  in one run.
- **The exact real packet size.** Derived from the published PM4 layout, not measured from a guest
  reservation; recorded under `## Open` in `docs/AGC_PACKET_SIZES.md` with the `PROSPER_DCBWIN`
  recipe that would settle it.

## Review

Worth an independent read: this is reverse-engineered API semantics (argument-buffer layout, the
packet's dword count as an ABI contract) and it touches a shared decoder/executor path. The
argument-count claim and the `Draw::indexed` change in the shared command-processor arm are the two
places a second reader can most usefully challenge the reasoning.

### DIFF
--- a/prosper/docs/AGC_PACKET_SIZES.md
+++ b/prosper/docs/AGC_PACKET_SIZES.md
@@ -125,6 +125,7 @@ alone** — that is how a working title gets broken to satisfy a table. Only a g
 | `DcbDrawIndexAuto` | 7 | `DRAW_INDEX_AUTO` | 3 | LOW | same modifier; not a bare PM4 draw |
 | `DcbDrawIndexOffset` | 3 | `DRAW_INDEX_OFFSET_2` | 5 | LOW | prosper emits **fewer** — safe direction |
 | `DcbDrawIndexIndirect` | 4 | `DRAW_INDEX_INDIRECT` | 6 | LOW | fewer — safe direction |
+| `DcbDrawIndirect` | 4 | `DRAW_INDIRECT` | 5 | MED | **new (#2929)**, and the count is derived rather than copied from the indexed row. prosper's payload is the API's own operand list — 32-bit byte offset + 64-bit `ShaderDrawModifier` — so it cannot be smaller; the hardware packet spends header + `DATA_OFFSET` + `BASE_VTX_LOC` + `START_INST_LOC` + `DRAW_INITIATOR`, the same field list as the indexed form. No dump was observed inlining the real size, so the equality is inferred, not measured; only the **upper** bound is load-bearing and 4 ≤ 5 settles it. prosper answers `sceAgcDcbDrawIndirectGetSize` (`cxPZ4Wgvdj8`) from the same constant |
 | `CbDispatch` | 6 | `DISPATCH_DIRECT` | 5 | LOW | no title evidence |
 | `DcbDispatchIndirect` | 4 | `DISPATCH_INDIRECT` | 3–4 | LOW | no title evidence |
 | `AcbDispatchIndirect` | **5** | `DISPATCH_INDIRECT` (MEC, address form) | 4 | MED | **one dword LARGER than the Dcb form, deliberately** (#3218). The ACB carries a whole 64-bit argument address, not an offset: libSceAgc 3.20 has 36 `sceAgcAcb*` exports and no SetBase among them, so nothing on that ring defines a base. Astro Bot's async-compute stream is the live half — its packets carried exactly the low 32 bits of the argument allocations its Dcb announces at full width in the same run. prosper answers `sceAgcAcbDispatchIndirectGetSize` too, so the guest reserves 5 |
@@ -397,4 +398,8 @@ happened to ask, and each time fixed for that title's NIDs alone.
   "final buffer" submit variant (#232), derived from live disassembly and load-bearing. `CbBranch`
   also has a `GetSize`, which implies it is a *builder*. Recorded, not changed — a title calling it
   as a branch would currently reach the submit path instead.
+* `DcbDrawIndirect`'s 4 dwords are an inference from the published `DRAW_INDIRECT` field list,
+  not a measurement of a guest reservation (#2929). The direction that matters is proven — 4 is at or
+  below the reference, so it cannot overrun — but a `PROSPER_DCBWIN` window over a live caller
+  (Little Nightmares II `PPSA02154`, Sifu `PPSA03001`) would settle the exact value cheaply.
 * The three excluded size-carrying `GetSize` functions.
--- a/prosper/src/gpu/execute/gpu_executor.cpp
+++ b/prosper/src/gpu/execute/gpu_executor.cpp
@@ -9926,6 +9926,60 @@ bool resolve_indirect_draw_arguments(const GpuState& submit, const GpuState::Dra
                                      GpuState::Draw& resolved) {
     resolved = source;
     if (!source.indirect) return true;
+    if (!source.indexed) {
+        // sceAgcDcbDrawIndirect (#2929) — the NON-indexed form. Its argument buffer is FOUR dwords,
+        // not five: (vertexCount, instanceCount, startVertex, startInstance). Reading the indexed
+        // five-dword layout here would demand a fifth dword the guest never wrote and then reject
+        // the draw for having no index base, which is the same silent disappearance the missing
+        // registration caused. Resolves to the DrawIndexAuto shape the realizer already renders:
+        // vertex count in index_count with `indexed` false, and startVertex through the same
+        // per-draw vertex-offset override the indexed path uses (the realizer hands it to
+        // vkCmdDraw's firstVertex).
+        constexpr uint32_t kArgumentBytes = 4u * sizeof(uint32_t);
+        if (!source.indirect_args_addr || (source.indirect_args_addr & 3u) ||
+            !guest_readable(source.indirect_args_addr, kArgumentBytes)) {
+            static std::atomic<int> warned{0};
+            if (warned.fetch_add(1) < 24)
+                std::fprintf(stderr,
+                             "[agc] indirect draw skipped: unreadable arguments at 0x%llx\n",
+                             static_cast<unsigned long long>(source.indirect_args_addr));
+            return false;
+        }
+        uint32_t args[4] = {};
+        std::memcpy(args, reinterpret_cast<const void*>(source.indirect_args_addr), sizeof(args));
+        const uint32_t vertex_count = args[0];
+        const uint32_t instance_count = args[1];
+        const uint32_t first_vertex = args[2];
+        const uint32_t first_instance = args[3];
+        constexpr uint32_t kMaxIndirectCount = 1u << 20;
+        if (!vertex_count || !instance_count) return false;  // hardware no-op
+        if (vertex_count > kMaxIndirectCount || instance_count > kMaxIndirectCount ||
+            first_instance != 0) {
+            static std::atomic<int> warned{0};
+            if (warned.fetch_add(1) < 24)
+                std::fprintf(stderr,
+                             "[agc] indirect draw skipped: vertices=%u instances=%u first_vertex=%u "
+                             "first_instance=%u\n",
+                             vertex_count, instance_count, first_vertex, first_instance);
+            return false;
+        }
+        resolved.index_count = vertex_count;
+        resolved.instance_count = instance_count;
+        resolved.indexed = false;
+        resolved.indirect_vertex_offset = static_cast<int32_t>(first_vertex);
+        resolved.has_vertex_offset_override = true;
+        resolved.indirect = false;
+        if (std::getenv("PROSPER_INDIRECTLOG")) {
+            static std::atomic<int> logged{0};
+            if (logged.fetch_add(1) < 256)
+                std::fprintf(stderr,
+                             "[agc-indirect] draw args=0x%llx vertices=%u instances=%u "
+                             "first_vertex=%u\n",
+                             static_cast<unsigned long long>(source.indirect_args_addr),
+                             vertex_count, instance_count, first_vertex);
+        }
+        return true;
+    }
     constexpr uint32_t kArgumentBytes = 5u * sizeof(uint32_t);
     if (!source.indirect_args_addr || (source.indirect_args_addr & 3u) ||
         !guest_readable(source.indirect_args_addr, kArgumentBytes)) {
--- a/prosper/src/gpu/pm4/command_processor.cpp
+++ b/prosper/src/gpu/pm4/command_processor.cpp
@@ -4631,7 +4631,8 @@ void GpuState::apply(const Pm4Command& c) {
             draws.back().command_order = command_order;
             break;
         }
-        case K::DrawIndexIndirect: {
+        case K::DrawIndexIndirect:
+        case K::DrawIndirect: {
             // #305: bind-vs-work sequence in stream order. Cached — this runs per draw/dispatch,
             // and a per-item environ scan is measurable at this title's ~876k items per route.
             // The three graphics arms emit "DRAW"; the two compute arms below emit "DISPATCH", so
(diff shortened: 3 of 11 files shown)


mattias800 (commented):
APPROVED — independent review on exact head `8bac934f`, I am not the author. Merge still gated on CI.

## The claim I spent the most time on: the dword count

`CLAUDE.md` singles this out — *a builder's dword count is an ABI contract with the guest's own
reservations* — and the issue explicitly warned against copying the indexed form's count. So I did
not take the derivation on trust.

The reasoning holds, and it holds for a better reason than the PR leads with. Three legs:

1. **Self-consistency.** `sceAgcDcbDrawIndirectGetSize` (`cxPZ4Wgvdj8`) is answered from the *same*
   `kDwDrawIndirect` constant. A guest that asks therefore reserves exactly what prosper writes, and
   the question of Sony's real value never arises for it.
2. **Only one direction can hurt**, for a guest that hardcodes an inlined size instead of asking:
   writing *fewer* dwords than reserved wastes space, writing *more* overruns (#1748). So the
   load-bearing property is an upper bound, not equality — and the PR's `CONFIDENCE` split says
   exactly that (MED on equality, HIGH on not-an-overrun). That split is the right shape and is rarer
   than it should be.
3. **The precedent is weaker than this change.** `DcbDrawIndexIndirect` — already shipped, already
   exercised by live titles that render — is `prosper dw = 4` against a published 6, at confidence
   **LOW**, justified in the table as "fewer — safe direction". The new row is the same count at
   **MED** with a derivation rather than an assertion. So this is strictly better evidenced than the
   sibling it sits beside, on a code path with a working precedent at the identical value.

The `## Open` entry naming `PROSPER_DCBWIN` over a live caller as the cheap way to settle the
equality is the right disposition: it records the residual uncertainty as an *action* rather than
leaving it as a comment nobody can act on.

## Treating it as a distinct operation was the important call

The tempting shape here is a flag on the existing indexed sub-op. The PR instead adds
`R_DRAW_INDIRECT = 0x25` and `Kind::DrawIndirect`, and the reason given is correct and specific: the
argument buffer at the far end is **four** dwords (`vertexCount, instanceCount, startVertex,
startInstance`) versus the indexed form's five. Reusing the indexed sub-op would make the executor
read a fifth dword the guest never wrote, then drop the draw for having no index base — reproducing
the issue's own "silent disappearance" one layer further down, where it would be *harder* to find
because a draw that decodes and then vanishes looks like a renderer problem rather than a missing
registration.

## The one behavioural change to existing code, checked

```cpp
- d.indexed = true;
+ d.indexed = (c.kind == K::DrawIndexIndirect);
- d.index_base = index_base;
+ if (d.indexed) d.index_base = index_base;
```

For every pre-existing caller `c.kind` **is** `DrawIndexIndirect`, so this evaluates to `true` and
the indexed path is bit-identical to before. The `index_base` guard only withholds a field on the
new non-indexed path, where it has no meaning. I checked this specifically because a hard-coded
constant becoming a computed predicate is where a silent regression to a working path would hide,
and there is none.

## Verification I did not duplicate, and one thing I want to flag

The environmental-failure argument is the strongest form available and I am satisfied with it: the
author built a *detached worktree at pristine `origin/main`*, ran the same configure/build/ctest, got
the same **458 discovered / 112 failed / exit 8**, and `diff` of the two sorted failing-test lists is
**empty**. Identical set, identical statuses — this branch adds no failure. That is a comparison, not
an assertion, and it is the right way to discharge a red suite on a box that cannot build the Vulkan
chain.

The without-fix arm is real: both `RN(...)` registrations commented out, rebuilt, and all four
extended tests go red on exactly the new assertion and nothing else.

**Flagging for the record, not blocking:** `test_gpu_execute` is **compile-verified only** locally —
it lives in the chain this machine cannot build, and the binary aborts at
`[render] Vulkan 1.4 required; loader reports 1.3.275` before reaching any executor assertion. The
author says so plainly rather than letting a partial result read as a pass, which is the behaviour I
want. CI builds and runs that target, so **the merge must wait for green CI specifically for this
test**, and I will not merge on the local evidence alone.

## Non-blocking

The issue's `CONFIDENCE: LOW` correlation with the flat-white symptom on Little Nightmares II
(`PPSA02154`) remains untested — no renderer here means no boot. Worth one `--seconds 13 --count 30`
default-route run by whoever next has a working renderer; it would either convert this from "a
silently dropped draw" to "the cause of a visible defect on a named title", or rule it out.

PR #3543, opened 2026-09-11
flip_pacing_report: state the chance baseline the tick-aligned share is read against

### LINKED ISSUE #3454: flip_pacing_report: the tick-alignment percentage has no stated chance baseline — its null is ~30%, so 50% reads as a signal when it is noise
The report prints:

```
tick-aligned intervals (within 15% of a 15.625 ms tick multiple): 2872 (50.1%)
```

**It does not say what that number is against.** The acceptance band is +/-15% of the tick, i.e. +/-2.34 ms on a 15.625 ms period, so the band covers **4.69 of every 15.625 ms = ~30% of the real line**. Arbitrary intervals land in it 30% of the time.

So the printed percentage is uninterpretable without a baseline the reader has to derive:

| observed | reading |
| --- | --- |
| 96.5% | far above chance — real quantisation |
| 60.5% | ~2x chance — suggestive |
| 50.1% | ~1.7x chance — weak |
| **9.3%** | **BELOW chance — actively anti-aligned** |

All four came from one GTA V run (#3450). A reader taking 50.1% at face value concludes "half the frames are tick-bound"; the honest statement is "modestly above a 30% null". And the 9.3% window carries real information — the production there is *not* tick-driven — which the current framing throws away, because a low percentage reads as "no finding" rather than as evidence against the hypothesis.

The whole-run figure is also an average over windows that disagree by 10x (9.3% to 96.5%), which is exactly the smearing the per-window split exists to prevent — the summary line reintroduces it.

## Suggested fix

- Print the chance baseline next to the figure, derived from the band and tick actually in use, e.g. `50.1% (chance baseline 30.0% for a +/-15% band)`.
- Say when a window is **below** baseline rather than letting it read as a weak positive.
- Consider dropping or de-emphasising the whole-run percentage in favour of the per-window ones, or flag it when windows disagree by more than some factor.
- The classification prose in the module docstring ("clustering at timer-tick multiples ... a wait in the production path resolving on the OS tick") should be gated on the excess over baseline, not on the raw share.

Found while using the tool for #3450 T9, immediately after #3453 made it able to read a real run at all. The two are independent: this one is about interpreting the number, not producing it.

mattias800:
Fixes #3454.

## The failure scenario

`flip_pacing_report.py` printed:

```
tick-aligned intervals (within 15% of a 15.625 ms tick multiple): 2872 (50.1%)
```

and said nothing about what that number was against. The acceptance band is `±15%` of the tick, so
it covers `2 × 0.15 = 30%` of the real line — **intervals bearing no relationship whatsoever to the
tick land inside it 30% of the time.** 50.1% is therefore 1.7× chance, which is weak. The natural
reading, *"half the frames are tick-bound"*, sends the next investigation hunting a wait in the
production path that the data does not implicate.

**The low end is the worse half.** One GTA V run (#3450) produced five windows — 64.3%, 22.4%,
9.3%, 96.5% and 60.5%, with 50.1% as the whole-run aggregate. Only the first is a strong finding, and **9.3% is *below* chance**: positive evidence that
production there is anti-aligned and *not* tick-bound. Under the old framing a low share read as "no
finding", so the one window carrying a clean negative result was the one thrown away.

## Behavioural contract

- The share is never printed without its baseline, the ratio, and a verdict.
- A sub-chance window is named as **evidence against** tick-bound production, not as weak support.
- The whole-run figure declares itself an average when its windows disagree.
- The printed band, the predicate's band, and the baseline all derive from **one** constant.

## Approach

| change | why |
| --- | --- |
| `BAND = 0.15` named | it was a bare `0.15` inside `classify()` while the report's sentence said `"15%"` in a *separate literal* — two places to change, one of them silent |
| `CHANCE = 2 * BAND`, printed with ratio + verdict | the fix the issue asks for |
| `alignment_verdict()` split out | thresholds assertable directly, so a wording change in the printer cannot silently move a boundary |
| per-window ratio + `BELOW chance` marker | recovers the information the 9.3% window carried |
| whole-run spread note | the summary was reintroducing the very smearing the windowed split exists to prevent |
| docstring prose gated on excess over chance | the tool's own classification guidance said "clustering at tick multiples" with no null |

## The null is measured, not asserted

This is the part I want a reviewer to look at, because a test that asserted `2 × BAND == 30%` would
only restate the arithmetic the code performs — the positive-control trap `CLAUDE.md` describes,
where a control drawn from the same source tests the discriminator and not the domain.

So **case 12 builds the null outside the tool**: intervals drawn uniformly from `[20, 120]` ms by a
generator with no relationship to the tick.

| seed | observed tick-aligned share |
| --- | --- |
| 1 | 27.2% |
| 2 | 28.3% |
| 3 | 29.3% |

The exact expected coverage for that support is derivable and is **not 30%**: the tick multiples
lying wholly inside `[20, 120]` are `k = 2..7` (31.25 … 109.375 ms), six bands of
`2 × 0.15 × 15.625 = 4.6875` ms, so `28.125 / 100 = 28.1%`. `k = 1` and `k = 8` fall outside the
range and contribute nothing. Measured mean ≈ 28.3% against a derived 28.1%.

Two things follow, and both are in the code as comments:

1. The baseline is real and behaves as predicted on data the tool has never seen.
2. **The printed 30.0% is the asymptotic figure; the exact null depends on the interval
   distribution's support.** The test therefore asserts a *window* (22–34%) plus the verdict text,
   rather than a point it would be dishonest to claim.

## Does it fail without the fix?

Yes — and per-arm, not as one crash. My first version aborted on `AttributeError` at case 11, which
proved only that case 11 discriminates and left cases 12–15 unexecuted. The constants are now read
with `getattr` and fall back to the expected values, so the pre-fix tool produces **12 distinct
named failures**:

```
 - case 11: flip_pacing_report exposes no BAND/CHANCE/alignment_verdict, so the tick-aligned
            share has no stated null at all (#3454)
 - case 11 baseline is printed: expected 'chance baseline 30.0%' in the report
 - case 11 printed band is derived from BAND: expected '+/-15%' in the report
 - case 11 a perfectly tick-locked run is a strong finding: expected 'well above chance'
 - case 12 an unrelated run must read as no evidence, not a finding: expected 'at chance'
 - case 13 anti-aligned production is named as such: expected 'BELOW chance'
 - case 13 says what a sub-chance share means: expected 'evidence AGAINST'
 - case 14 a two-regime run flags its own summary as an average: expected 'windows disagree'
 - case 15: no alignment_verdict to assert thresholds against
 - case 15: a share exactly at chance must read as at-chance
 - case 15: a third of chance must read as below
 - case 15: three times chance must read as a real finding
WITHOUT_FIX_RC=1
```

Restored → all cases pass.

Note that case 12's *numeric* window passes in both directions by design: the pre-fix tool still
prints the share, and only the verdict is missing. That assertion exists to guard the baseline
claim — it would redden if `BAND` and `CHANCE` were ever changed out of step — not to discriminate
the fix.

- **Unaffected:** the predicate itself. `classify()` accepts the same intervals as before; `BAND`
  merely names the constant it already used. No interval changes bucket, so no existing
  classification moves. fps, mean period, histogram, windowing and the slowest-interval list are
  untouched, as are all ten pre-existing test cases.
- **Affected:** the two tick-alignment lines and the docstring's classification guidance.

## Risks

Low; this is a reporting change over an unchanged predicate. The judgement calls worth challenging
are the three thresholds (`<0.75×` below, `<1.25×` at chance, `<2×` weak) and the 25-point window
spread that triggers the average warning. All four are deliberately coarse — a share against a flat
null does not support a sharper claim — and all are asserted directly in case 15 so moving one is a
visible edit rather than a silent drift.

## Verification

Python-only; no C++ build involved.

```
python prosper/tools/perf/test_flip_pacing_report.py     -> all cases pass, exit 0  (15 cases)
  ... same, against origin/main's flip_pacing_report.py  -> 12 failures,   exit 1
  ... restored                                           -> all cases pass, exit 0
```

Registered as ctest `flip_pacing_report_logic`, which already ran this file.

Docs gates, run because this adds instrument-trap row 275 to a numbered table:

```
check_numbered_table.py --ordered --table-header Instrument --baseline <origin/main copy>
  -> 14 tables, 337 rows, numbered column 1..275, unique and ascending,
     none of the 274 base rows deleted
git ls-files '*.md' -z | xargs -0 check_numbered_table.py   -> 0 errors
```

Trap number allocated with `trap_number.py` (scans `main` and every open PR): 275 was free, one
open PR touches the file and adds no row. I will re-run `check_merge_result.py` immediately before
merging, since that is the only check that sees a concurrent append.

`git diff --check`: clean. Session-trailer gate: 0.

---

## Corrections after review (head `cc4a42b8`)

This PR was **REJECTED** on its first head with two blocking findings, both correct. Recording them here rather than only in the thread, because the body is what a reader of the merged PR sees.

1. **The fix introduced its own confident negative.** `nearest = max(1, round(ms/tick))` means an interval shorter than the k=1 band's bottom edge (13.28 ms, i.e. any run above ~75 fps) can never be classified aligned — so the new sub-chance verdict told a 156 fps run it was “actively anti-aligned, evidence AGAINST a tick-bound wait” where no other answer was arithmetically possible. `alignment_power()` now suppresses the verdict when alignment is unreachable for most of the population, and says so.
2. **Trap row 275 misquoted its source, and the figures above were wrong too** — corrected in place. That run has five windows, and 50.1% is the aggregate, which #3450's own comment states outright. #3454's summary table listed the aggregate beside the windows under one heading and I copied it without opening the source.

Also actioned: the 0.75/1.25/2.0 thresholds are now labelled a judgement call rather than left looking derived; **case 12's tolerance did not guard what this body claimed** (a fixed 22–34% window tolerates BAND in 0.118–0.181 and is blind to CHANCE) and now derives the expectation from BAND and asserts `CHANCE == 2*BAND` separately; and a three-interval trailing window no longer votes in the spread NOTE.

Cases 16–18 added — three named failures against the previous head. Case 17 passes in both directions by design and is labelled an over-suppression control, not a discriminator.


mattias800 (commented):
Independent adversarial review. **REJECTED**, on two findings, both cheap to fix. The central arithmetic is right and I confirmed it independently — the rejection is narrow, not a disagreement with the premise.

## First, the number you asked to have re-derived: I get **28.125%**, by three routes

Two of them are mine and do not touch your code.

1. **Analytic.** Acceptance is `|ms - k*T| <= BAND*T` with `k = max(1, round(ms/T))`. Since `BAND = 0.15 < 0.5`, each band lies wholly inside its own rounding cell, so the accepted set is the union over k of `[kT - 2.34375, kT + 2.34375]`. Intersected with `[20, 120]`: k=1 is `[13.281, 17.969]` (no overlap), k=2..7 are `[28.906, 33.594] ... [107.031, 111.719]` (six bands wholly inside), k=8 is `[122.656, 127.344]` (no overlap). `6 x 4.6875 / 100 = 28.125%`.
2. **Numeric, own predicate.** 10,000,000-point grid over `[20,120]`, predicate re-implemented from scratch: **28.1250%**.
3. **Numeric, your `classify()`.** Same grid at 1e6 points through the imported function: **28.1256%**. The predicates agree, so your band matches your description of it.

I also reproduced the three-seed table exactly — `random.Random(1/2/3)`, 4000 intervals: **27.2% / 28.3% / 29.3%**. That claim is measured and reproducible, not inferred, and the binomial SE at n=4000 is 0.71 pp, so the spread is ordinary.

One observation on it: the **committed** case 12 uses seed `20260911`, which scores **26.9%** — not one of the three values the comment directly above it lists. A reader will assume the committed seed produced one of them. Say "measured across seeds 1/2/3; the committed seed scores 26.9%", or commit one of the three.

And the assertion window is weaker than the body claims. The PR says case 12's numeric window "would redden if `BAND` and `CHANCE` were ever changed out of step". It does neither half: the observed share is `1.875 * BAND`, so 22–34% tolerates `BAND` anywhere in **0.118 ... 0.181** (it also reddens on a large *in-step* change), and it is **completely blind to `CHANCE`**, which does not enter the observed share at all. You are right that it does not discriminate the fix; it also does not guard what you say it guards. Either drop that drift claim or tighten the window to something like 26–30.5% (still ~3 SE).

## Blocking 1: the new "BELOW chance — evidence AGAINST" verdict has zero power on any run faster than ~75 fps, and asserts a finding there anyway

`nearest = max(1, round(ms/tick))` deletes the k=0 band. So an interval below `0.85 * tick = 13.28 ms` can never be accepted — the nearest reachable band edge is further away than `BAND*tick`. For an interval distribution supported entirely under 13.28 ms the true null is **0%**, not 30%, and an observed 0% is exactly *at* chance.

Measured, on this branch's own tool, with a synthetic 156 fps run (the median figure the charter quotes for *The Messenger*; 3000 flips, sigma 1.2 ms):

```
flips=3000 intervals=2999 mean period 6.41 ms -> 156.0 fps average
tick-aligned intervals (within +/-15% of a 15.625 ms tick multiple): 0 (0.0%)
  chance baseline 30.0% -- 0.00x chance: BELOW chance -- production here is anti-aligned,
  which is evidence AGAINST a tick-bound wait, not an absence of evidence
  flips     0- 1560:  155.8 fps  mean   6.42 ms  tick-aligned   0.0% (0.00x chance) BELOW chance
  flips  1560- 2999:  156.2 fps  mean   6.40 ms  tick-aligned   0.0% (0.00x chance) BELOW chance
```

The tool has just printed, three times, that it holds positive evidence against a tick-bound wait, on data where the test is arithmetically incapable of returning anything else. The verdict's own closing clause — "not an absence of evidence" — is precisely backwards here: it *is* an absence of evidence, dressed as a result.

This is your own issue one level up. #3454 is "a share printed against an unstated null"; this ships "a verdict printed against a null that does not apply to the data". Before the fix the tool printed a bare `0.0%` and the reader drew their own conclusion; after it, the tool draws a wrong one for them — and it does so on the *fast* runs, i.e. every healthy title, which is most of the corpus.

Note the support-dependence cuts both ways and is not only the zero case: your own case 12 shows 28.1% rather than 30%, and a tight unimodal distribution can sit anywhere from 0% to 100% coverage. The code comment and the test acknowledge this. **The printed output does not**, and the printed output is what gets quoted into an issue.

Two remedies, either acceptable:

- **Cheap and honest:** compute the reachable coverage over the observed interval range before issuing any verdict. If `max(intervals) < tick*(1-BAND)`, print "every interval in this run falls below the first tick band — this test has no power on this distribution" and no ratio. About four lines, and it kills the whole class.
- **Better and still small:** replace the constant with a **support-aware empirical null** — re-classify the observed intervals with a uniform phase offset in `[0, tick)` over ~20 repetitions and use the mean as the baseline. That yields ~28.1% on your case-12 generator, 0% on the 156 fps run, and makes the printed baseline a property of the data rather than of the constant. It would also let case 12 assert the tool's own computed null against your derived 28.125%, which is a far sharper test than a 12-point window.

## Blocking 2: instrument-trap row 275 misstates the measurement it cites

The row — and the PR body, and the `CHANCE` comment in the tool — says:

> One GTA V run (#3450) produced four windows at **96.5%, 60.5%, 50.1% and 9.3%**

The run in #3450 produced **five** windows, and **50.1% is not one of them** — it is the whole-run figure over 5,734 flips, which the same row quotes two sentences earlier as `2872 (50.1%)`. The actual table there is:

| window | tick-aligned |
| --- | --- |
| 0-1206 | 64.3% |
| 1206-2412 | 22.4% |
| 2412-3618 | 9.3% |
| 3618-4824 | **96.5%** |
| 4824-5733 (world) | 60.5% |

So the row promotes the aggregate into the window list, drops two real windows, and contradicts itself inside one paragraph. The charter calls trap rows the most expensive knowledge in the repository and they are append-only; a row that misdescribes its own source measurement is exactly the thing nobody re-derives.

The omission is not neutral either. **22.4% is 0.747x chance** — it lands one thousandth under your `< 0.75` "BELOW chance" boundary. The one real datapoint sitting on an arbitrary threshold is the one that did not make it into the record. That deserves a sentence in the row, and it is the best available argument about the thresholds.

Fix is docs-only: quote the five windows, name 50.1% as the whole-run average, keep everything else.

## Non-blocking findings

- **The thresholds are taste and should say so.** You asked: I agree, `0.75 / 1.25 / 2.0` and the 25-point spread have no derivation. The docstring says they are "deliberately coarse", which reads as a justification rather than as an admission — write "chosen by judgement, not derived". There is a derived alternative for nearly free, and it matters because the verdicts carry no n: at the null, SE = `sqrt(0.21/n)`, so separating 1.25x from 1.0x at 2 SE needs **n > 149**. Window verdicts below that are noise. Printing each window's interval count, or suppressing the verdict under ~150, beats any choice of constants.
- **A short trailing window fires the "windows disagree" warning on a run that does not disagree.** Measured on this branch: 644 flips of tick-locked production, `--window-s 10` gives `window_flips = 640` and a 3-interval remainder:
  ```
    flips     0-  640:   64.1 fps  mean  15.60 ms  tick-aligned 100.0% (3.33x chance)
    flips   640-  643:   54.7 fps  mean  18.27 ms  tick-aligned  66.7% (2.22x chance)
    NOTE: windows disagree by 66.7%..100.0% ... Read the windows, not the summary.
  ```
  The run is 99.8% tick-aligned throughout and the tool tells the reader to distrust an accurate summary on the strength of three samples. Same remedy: keep windows below a minimum size out of `window_shares`, or fold the remainder into the previous window.
- **The `chance <= 0` guard cannot protect the code that would need it.** `alignment_verdict` returns `(None, ...)` and case 15 asserts that — but both call sites in `report()` format the ratio unconditionally (`f"{ratio:.2f}"`, `f"{w_ratio:.2f}"`), so a `BAND = 0` edit raises `TypeError` rather than taking the guarded path. The guard is tested where it is not used and unused where it would be needed.
- **`--tick-ms` is fine.** I checked whether a non-default tick invalidates the printed 30.0%: it does not, because the band is a fraction of the tick, so asymptotic coverage is tick-independent. No finding.

## What I checked and found nothing wrong with

The predicate really is unchanged (`band=BAND` defaults to the same value), so no interval changes bucket. The 12 without-fix failures are genuine discriminators except case 12's numeric window, which you already label as non-discriminating — and its `"at chance"` companion does discriminate, so the case earns its place. The full 15-case suite passes on the branch (run with the repo layout mirrored so case 7's emitter grep resolves): `flip_pacing_report self-test: all cases pass`, rc 0.

Fix the row, give the sub-tick case an honest verdict, and this is a clear improvement to the tool.

mattias800 (commented):
Both blocking findings accepted and fixed on `cc4a42b8`. This was a better review than the PR deserved, and the independent 28.125% derivation by three routes is the part I'll remember — it confirmed the one thing I most wanted checked, by methods I hadn't used.

**1. The k=0 power gap — correct, and it is the worse of the two.** `nearest = max(1, round(ms/tick))` means anything below 13.28 ms is measured against one tick however small it is, so a run above ~75 fps can never score aligned. My sub-chance verdict then told it "evidence AGAINST a tick-bound wait" where no other answer was arithmetically possible. That is the same defect class I was fixing, reintroduced by the fix — which is the sharpest way to be wrong about an instrument.

`alignment_power()` reports the fraction of intervals for which alignment is *reachable*; below half no verdict is given and the output says NO POWER and why, per window as well as overall. Measured: your 156 fps case now reads `NO POWER -- 100% of intervals are shorter than the band around one tick`; the anti-aligned run at 1–2 ticks still reports BELOW chance, so the gate does not over-apply. The GTA V run is unaffected — its windows mean 24–134 ms, all above the floor.

**2. Trap row 275 — you were right, and I verified it at the source rather than taking the numbers.** `gh issue view 3450 --comments` gives the table: five windows, **64.3 / 22.4 / 9.3 / 96.5 / 60.5**, and the comment says in as many words *"A single overall figure (50.1%) averages these into something meaningless."* So 50.1% is the aggregate, exactly as you said.

The mechanism is worth recording: **#3454's summary table listed the aggregate beside the windows under one heading**, and I copied it into the trap row without opening #3450. A trap row is the worst place in this repo for an unverified figure, because it is read as already checked — so the row now carries that as its own third lesson alongside the corrected table. Your observation that the dropped **22.4% is 0.747×**, sitting exactly on the 0.75 boundary I invented, is in there too; it is a good argument that the boundary is arbitrary.

**3. Thresholds.** Agreed, they are taste. Now labelled as a judgement call, with the sample size a real test would need (n > ~150 to separate 1.25× from chance) — thank you for the derived alternative, it made the label concrete rather than apologetic.

**4. Case 12 — you are right that it guarded neither half.** A fixed 22–34% window tolerates BAND anywhere in 0.118–0.181 and is blind to CHANCE. It now *derives* the expected coverage from BAND for the generator's support, so a predicate that stops using BAND is caught even though the observation moves with it, and asserts `CHANCE == 2*BAND` and the printed baseline separately. The PR body's claim about it was wrong and has been corrected in place.

**5. The three-interval trailing window** — fixed; short windows no longer vote in the spread NOTE.

**On the committed seed scoring 26.9%:** correct, and it was sloppy to quote 27.2/28.3/29.3 from seeds 1/2/3 while committing a different one. The derived-tolerance rewrite makes the specific seed immaterial, but the body should not have implied the committed arm produced the quoted numbers.

**`chance <= 0`:** kept. You are right it is unreachable from both call sites, but it *is* exercised directly by case 15, and it is what stops a future caller dividing by a widened band. Now says so in a comment rather than looking like dead code.

Cases 16–18 added: three named failures against the previous head. Case 18 is wrapped so a helper with no `power` parameter reports rather than raising `TypeError` and aborting before 16 and 17 print — the same "crash instead of report" flaw you found in #3539's arms, which I went looking for here because of your review.

mattias800 (commented):
Re-review of the delta `d24a1861..cc4a42b8` only. **APPROVED**, with one residual I want on the record rather than blocking on.

## Both blocking findings are fixed

**The k=0 gap.** `alignment_power()` uses the right floor: reachability is exactly `ms >= (1 - band) * tick`, because at 13.28 ms the error against the forced k=1 is 2.34 ms, which is the band edge. I checked the boundary rather than taking the comment's word for it. The gate is applied per window as well as overall, the window line carries `no power` instead of `BELOW chance`, and case 17 is the right companion arm — an over-suppression control, correctly labelled as passing in both directions rather than counted as a discriminator. Case 16's 156 fps run is the case I measured in the last round and it now reports NO POWER with no "evidence AGAINST" anywhere in the output.

**Trap row 275.** Corrected to the five real windows (64.3 / 22.4 / 9.3 / 96.5 / 60.5) with 50.1% named as the aggregate, and the mis-sourcing kept as the row's own third lesson rather than quietly repaired — which is the more useful outcome, since the row is now an instance of the thing it describes. Opening #3450 yourself instead of taking my numbers is the correct response to a review finding; it is also the one the charter warns is usually skipped when the correction is one the author is inclined to accept.

**Case 12.** Deriving the expected coverage from `BAND` closes both halves of what I said the old window failed to guard. I re-ran the derivation loop by hand: `k` runs 1..7 (k=8's lower edge 122.66 is not < 120), k=1 contributes nothing (upper edge 17.97 < 20), k=2..7 contribute 4.6875 each, total 28.125% — matching the 28.125% I derived independently last round. The 3-point tolerance is ~4 SE at n=4000 and now catches a BAND drift of about ±0.016 instead of ±0.03, and the separate `CHANCE == 2 * BAND` assertion covers the half the observation cannot see. Thresholds are now labelled as judgement with the n > ~150 note, and short windows no longer vote in the spread NOTE (`n >= max(8, window_flips // 4)`).

Case 18's `TypeError` wrapper is the right pre-emptive move for the same reason case 11's `getattr` was.

## Residual, non-blocking: partial power still biases the ratio

The gate handles power = 0 and leaves everything between 0.5 and 1.0 measured against the full 30%. For a mixed run — a phase below the floor and a phase above it — the true null is `power * CHANCE`, because the sub-floor fraction contributes zero coverage rather than 30%. At power 0.6 the null is 18%, so an observed 20% is genuinely *above* its null and the tool will print `0.67x chance: BELOW chance ... evidence AGAINST`. That is the same fabricated negative as before, scaled down by at most 2x and now confined to mixed-rate runs — which a cinematic-to-gameplay phase change is exactly the shape of, and that is the case the windowing exists for.

One line fixes it if you want it: pass `chance = CHANCE * power` into the ratio (the NO POWER branch then survives as the floor for the degenerate case, since the ratio goes undefined at power 0). Not a merge blocker — the current state is strictly better than both the pre-fix tool and the last head — but I would rather it be a known limitation than be rediscovered from a window that reads as anti-aligned.

Nothing else found in the delta. Merge when your Linux checks are green.

mattias800 (commented):
Re-review of the delta `cc4a42b8..b0248e40` only. **APPROVED.** Two small things below, one of which I would fix before merging because it is on the line you hardened; neither changes a number.

## The scaling is right, and case 19 pins the half that matters

`ratio = share / (chance * power)` above the gate is the correct null: only `power` of the population can score aligned at all, the rest contribute a structural zero, so the expected share for a tick-unrelated run is `power * CHANCE`. Leaving the NO POWER branch on the flat null is defensible for the reason you give — it is not a verdict — and it keeps the printed ratio and the printed baseline consistent with each other on that branch.

Case 19's second assertion is the one I would have asked for and it is correct as written: `verdict_of(0.20, power=1.0)` must divide by `CHANCE` exactly, `abs(ratio - 0.20/CHANCE) > 1e-9`. Without it, `ratio = share / (chance * power)` could have been replaced by any monotone loosening — including dividing by something that quietly widens every verdict — and the partial arm alone would still have passed. With both halves, the test says "scale by exactly `power`, and by exactly 1 at full power", which is the actual contract. I checked it discriminates against `cc4a42b8`: 0.20/0.30 = 0.667 → `BELOW chance`, so both the verbatim-string arm and the ratio arm redden, while the full-power arm passes there (correctly — it is the guard, not the discriminator).

I ran the suite on `b0248e40`: all cases pass. I also checked the pre-existing cases are undisturbed, since scaling raises every partial-power ratio — case 12's uniform [20,120] intervals are all above the floor, so its power is 1.0 and its `at chance` verdict is unchanged; cases 13/17 likewise.

Measured the printed line at two power levels:

```
power~0.6 | chance baseline 18.2% (scaled by 61% reachable) -- 1.89x chance: modestly above chance ...
```

which is exactly the case that read `BELOW chance / evidence AGAINST` before. That is the fix working on the shape it was for.

## One nit, on the line you were careful about

At `power < 0.5` the baseline is deliberately **not** scaled, but the parenthetical is still printed:

```
power~0.2 | chance baseline 30.0% (scaled by 20% reachable) -- 0.39x chance: NO POWER -- 80% of intervals ...
```

`30.0%` is the unscaled null and the `0.39x` ratio is computed against it, so the numbers agree with each other — but the text says the baseline was scaled when it was not. It is the third misleading reading of that line, in the opposite direction from the one you closed. `scaled = "" if power >= 0.999 or power < 0.5 else ...` fixes it, or reword to `(20% reachable)` so the parenthetical states the fact rather than an operation. Not a number, so not a blocker; one condition, so I would do it.

## On the fourth lesson: keep it, but it overclaims by one item

You asked whether it reads as a pattern imposed on four unrelated slips. It does not — but the closing sentence is one item too broad. *"Each of these four was a correction that carried the previous defect forward in a narrower form"* cannot be true of (1), which is the original defect with no correction before it. And (2) is a different family: the first defect was a number without its **null**, while (2) was a number without its **provenance**. It is (3) and (4) where the claim is exact and worth keeping — (3) is "a verdict against a null that does not apply" and (4) is the identical defect narrowed to partial power, which is as clean an instance of the shape as this repository has.

So I would scope it rather than cut it: *"Each correction after the first carried a defect forward in a narrower form — (2) the provenance failure, (3) and (4) the null failure."* That keeps the load-bearing generalisation, which is genuinely useful to the next reader, and drops the part a sceptic can falsify by reading the row's own numbering. A trap row that overreaches by one item invites the whole row to be discounted, which would be a bad trade for four lessons this well evidenced.

Nothing else in the delta. Clean from my side — merge when your Linux checks are green.

Issue #2895, opened 2026-08-21
[Game tracker] PGA TOUR 2K25 (PPSA17952)


mattias800:
## Current milestone

**Rung 0.** The guest boots in 437 ms, links ten modules, streams its Unity assets and submits real
GPU draws — but every frame the live renderer publishes is black, and the process dies on a worker
thread before any menu. No screenshot is offered, because the only frame captured is uniformly
black and that is not progression evidence.

Two deaths were diagnosed this session. The first is fixed; the second is the current blocker.

### What the dump declares

| | |
| --- | --- |
| Title | PGA TOUR 2K25 |
| Title ID | `PPSA17952` |
| Content ID | `UP1001-PPSA17952_00-PGATOUR2K25GBL00` (2K) |
| SKU | `applicationCategoryType: 0`, `applicationDrmType: upgradable` |
| Version | master `01.00`, content `01.026.000` |
| Required system software | `0x1240000000000000` |
| **SDK** | **`0x0900000000000000` — 9.00** |
| Application size | 35 GB |

### Engine: Unity 6 / IL2CPP

`Media/data.unity3d` carries the version string **`6000.0.20f1`**, and the layout is standard
IL2CPP: `Media/Modules/Il2cppUserAssemblies.prx` plus a 30 MB `Media/Metadata/global-metadata.dat`,
so `prosper/tools/il2cpp/` applies directly. `Media/boot.config` declares
`gfx-enable-native-gfx-jobs=1`, `gfx-threading-mode=4`, 3840x2160 output and `hdr-display-enabled=1`,
which is what makes `-force-gfx-direct` the right launch argument.

Same engine generation as *Asterix & Obelix: Babylon Mission* (Unity 6, rung 2).

**The `Media/levelNN` scene-index progression oracle does not transfer** — there are no `levelNN`
files. Content is `data.unity3d` plus 26 `StreamingAssets/AssetBundles/` groups, so bundle opens
under `PROSPER_FILELOG=1` are the oracle here, as with *Tales of Graces f*.

**SDK 9.00 is below the SDK >= 13 post-submit-visibility gate (#2219)**, the same pre-13 class as
*ArcRunner* (SDK 10) and *Crisis Core*. This has **not** been needed and **not** been exercised —
both deaths seen so far are fully explained without it. Flagged only so the lever is known if an
unexplained mid-boot death appears later.

## What works

- Boot: `BOOT_COMPLETE` at **+437 ms**; ten modules linked and mapped.
[...]

PR #3544, opened 2026-09-11
libSceHttp2: stop reporting a request nobody sent as SCE_OK (#2894)

### LINKED ISSUE #2894: libSceHttp2 is unregistered, so a request that was never sent reports SCE_OK and its response getters hand the guest NULL

mattias800:
Fixes #2894.

## The failure

`libSceHttp2` was registered exactly once — `sceHttp2Init`, and that lived in `hle_service.cpp`.
Every other export of the library fell through to the dispatcher's unregistered default, `return 0`,
and **0 is `SCE_OK`**. Two different wrongs follow from that one default:

* `sceHttp2CreateTemplate` / `CreateRequestWithURL` / `CreateCookieBox` answered **0 as an id**, so
  the guest carried a handle that was never allocated into every later call;
* `sceHttp2SendRequest` reported that a request prosper never sent had been **sent successfully**,
  and `sceHttp2GetAllResponseHeaders` reported success **while writing nothing to its
  out-parameters**, so the guest parsed the pointer it already held.

In *PGA TOUR 2K25* (`PPSA17952`) that pointer is NULL, and the title's HTTP response-header scanner
at `eboot+0x142d5e0` dies on `movzx edx,BYTE PTR [r11+rcx*1]` with `r11 = 0`. It is the blocker
holding the title at rung 0. *Metaphor: ReFantazio* reaches the same library through
`sceHttp2CreateTemplate`.

## Behavioural contract

The rule is the one `src/hle/net/AGENTS.md` already states, applied in **both** directions:

* **Anything answerable from local state is implemented, not stubbed.** Creating a template, a
  request or a cookie box really does succeed offline — it allocates a local object — so it returns
  a real id from a real table. Setters really do record local state, so they record it and *then*
  answer `SCE_OK`. A getter that reads back state prosper itself recorded answers it.
* **Anything that needs a network answer FAILS, and leaves every out-parameter untouched.** There is
  no PSN behind this. `SCE_OK` is not a true answer, and `SCE_OK` with an unwritten out-parameter is
  the specific shape that crashes callers.

**Nothing is manufactured.** There is no synthesised status code, no fabricated header block, no
invented body, and no "pretend it worked" path. The guest's own error handling runs instead — which
is what it is written to do: the blocking title classifies the send result and returns cleanly on any
non-zero, and it was prosper's `SCE_OK` that walked it past that check.

## Approach and invariants

New `src/hle/net/hle_http2.cpp` (+ `hle_http2.hpp` for the error constants), registered from
`register_builtin_hle()` alongside `register_http_hle()`.

* **One unified id space** for contexts, templates, requests and cookie boxes. Separate per-kind
  tables would let a template id and a request id collide numerically, and a validator could then not
  tell a live id of the wrong class from a live id of the right one. Ids start at 1, so **0 — the
  value the dispatcher used to answer — is never a valid id of any kind**.
* **Two return conventions, because the contracts differ.** An id-returning entry point sign-extends
  its error, so a guest reading the answer as `int32` *or* `int64` sees it negative and never as a
  handle; an `SCE_OK`-or-error entry point keeps the zero-extended 32-bit form the library returns in
  `eax`. Both are commented as such, matching `hle_http.cpp`.
* **`sceHttp2Term` reclaims the whole tree** — a context owns its templates and cookie boxes, each
  template owns its requests. Releasing only the named slot would leak everything under it, which is
  the failure #3295 had to fix in the v1 sibling.
* **Setters validate that the id was allocated, then record, then answer `SCE_OK`.** They accept a
  live id of *any* class deliberately: the v1 family lets one setter take a template, connection or
  request id, and policing the class here would risk *rejecting* a call hardware accepts — the
  under-report half of the same defect. What is enforced is that the id exists, which is exactly what
  rejects the old default's 0.
* **`sceHttp2SendRequest` answers a libSceNet `ENETUNREACH` (`0x80410133`).** libSceHttp2 does not
  wrap a connect failure in an HTTP error — it propagates the raw libSceNet error, encoded
  `0x80410100 | <FreeBSD errno>`, pinned by the library special-casing `0x80410124` (errno 36,
  `EINPROGRESS`) as *not* a failure at its connect site, and corroborated by the title's own
  classifier naming `0x8041013d` (61, `ECONNREFUSED`) and `0x80410140` (64, `EHOSTDOWN`). An
  unreachable network is the honest offline condition, and it agrees with what prosper already tells
  the same guest through NetCtl (`SCE_NET_CTL_STATE_DISCONNECTED` / `NOT_CONNECTED`).
* **Every response getter answers that same failure and writes nothing** — `GetAllResponseHeaders`,
  `GetStatusCode`, `GetResponseContentLength`, `ReadData`, `ReadDataAsync`, `WaitAsync`. A getter
  reached before any send answers "out of order" rather than the send error.
* **`sceHttp2Init` moves here from `hle_service.cpp`**, because the context id it returns now has to
  be a slot `sceHttp2CreateTemplate` validates against. Its observable contract is unchanged
  (positive, distinct context ids); `tests/hle/dispatch/test_service_getters.cpp`'s offline-context
  arm still passes unchanged.

Coverage is the library's full export surface (55 functions + the dump's `_Z5dummyv`), so nothing in
`libSceHttp2` still falls to the dispatcher. Cookie *boxes* are modelled as real objects; cookie
*contents* are not, so `AddCookie`, `GetCookie`, `CookieExport`, `CookieImport`, `GetCookieStats` and
`GetMemoryPoolStats` fail rather than claim a store that does not exist — flushing the empty jar,
auth cache and redirect cache genuinely succeeds and is answered `SCE_OK`.

Confidence is marked per call. `CONFIDENCE: HIGH` on "the object id is argument 0" (the only shape
assumption any validation depends on) and on the id/error conventions; `MED` on individual argument
meanings carried across from the v1 family, on the facility carry-across for the `0x817b____`
constants, and on the exact errno; the two flag getters carry an explicit note that their 32-bit
out-parameter width is the one place in the file where being wrong writes memory rather than
returning a wrong value.

* **Affected:** any title that calls `libSceHttp2`. It now receives real ids, recorded settings, and
  an honest failure on the request/response path instead of `SCE_OK` everywhere.
* **Deliberately unaffected:** `libSceHttp` (v1) — a different facility with different constants, and
  its remaining request-path hole is #2930, handled separately. NetCtl/NP, `sceSslInit` and
  `sceNetPoolCreate` keep their existing answers; the only `hle_service.cpp` change is deleting the
  `sceHttp2Init` stub that moved.
* No renderer, loader or GPU path is touched.

## Risks

* **The exact error constants are `CONFIDENCE: MED`.** They are derived (facility from the shipped
  module and from the blocking title's own classifier, semantics for `_1100` from v1's validator),
  not invented, but they are not read from a name table. The behavioural finding on #2894 is that the
  guest classifies *any* non-zero as a failure, so a wrong constant here would be wrong rather than
  dangerous, and revisable without touching the object model.
* **Registering a NID can be worse than leaving it unregistered** if the argument shape is wrong. The
  mitigation is that every call whose shape is not well supported takes the "fail and touch nothing"
  path, which needs no knowledge of the out-parameter layout at all. The only calls that *write* guest
  memory are `GetAuthEnabled`, `GetAutoRedirect` and `GetCookieBox`, all writing a single `int32`.
* A title that previously stumbled forward on false success may now take a different branch. That is
  the intended change, and it is the branch the guest is written for, but it is a behaviour change for
  every `libSceHttp2` caller and not only for the crashing one.

## Verification

Platform: **Linux (WSL Ubuntu-24.04, GCC 13.3, Release)**. All commands from the worktree root.

```
cmake --build prosper/build-linux -j12 -- -k
ctest --test-dir prosper/build-linux --no-tests=error -j8
```

* New `test_http2` (`ctest -R http2_hle`): **48 executed assertions from 47 `CHECK` sites** (one site
  is inside the two-pass leak control and runs twice), **rc=0**. The neighbours it could disturb also
  pass: `test_http` rc=0, `test_hle_no_shadow` rc=0 (no shadowed NID — the moved `sceHttp2Init` is
  registered once), `test_service_getters` rc=0.
* `ctest --no-tests=error`: **459 tests discovered**, 347 passed, 112 failed, **rc=8** — and that rc
  is **environmental on this machine, not this branch**: this box has Vulkan **1.3.275** headers while
  `frontends/shared/device/vulkan_runtime.hpp` needs `VK_API_VERSION_1_4`, so the live-renderer chain
  and every Vulkan-execution test cannot build here at all and report `Not Run`. 297 targets built;
  the 9 compile errors are all `VK_API_VERSION_1_4` / `VkPhysicalDeviceHostImageCopyFeatures` in that
  one header. An `origin/main` baseline measured in a second worktree on the same box is posted as a
  comment below, so the claim is checkable rather than asserted.

### Every regression arm was shown to fail without the fix

Six mutations, each rebuilt and re-run; the test must go red for each and green again after restore.

| mutation | `test_http2` |
| --- | --- |
| baseline (unmutated) | rc=0, 0 failing checks |
| `register_http2_hle()` not called (**the pre-fix state**) | rc=1, **56** failing checks |
| `sceHttp2SendRequest` returns `SCE_OK` | rc=1, 4 failing checks |
| response getters return `SCE_OK`, out-parameters untouched | rc=1, 5 failing checks |
| `sceHttp2CreateTemplate` hands out id 0 | rc=1, 13 failing checks |
| `sceHttp2Term` releases only its own slot (leaks the tree) | rc=1, 1 failing check |
| `live_any` accepts any id including 0 | rc=1, 2 failing checks |
| restored | rc=0, 0 failing checks |

The arms are written to die to a *specific* wrong implementation rather than merely to be true of the
right one: every id arm demands a **positive** id for a good call and a **negative** one for a bad
call, so 0 satisfies neither; every response arm pre-fills its out-parameters with a sentinel and
requires the sentinel to **survive** alongside a non-zero return. The `Term` leak control fills the
table from one context *with templates and requests under it* before terminating it, so the leak it
guards is structurally expressible — the vacuity #3295 had to fix in the v1 sibling.

## What I could NOT verify

Said plainly, because it bounds what this PR claims:

* **No boot was run, and no title was observed.** This machine has no `PPSA17952` dump and no PS5
  console, so the fault at `eboot+0x142d5e0` is *predicted* to be gone, not *observed* to be gone, and
  nothing here establishes where PGA TOUR 2K25 now stops. #2895 is upstream of this and the title may
  halt somewhere else. `docs/PGA_TOUR_2K25_STATUS.md` is updated to say exactly that and the rung
  stays 0.
* **The shipped `libSceHttp2.sprx` is not on this machine**, so the offsets, argument shapes and
  facility derivation are taken from the reviewed derivation recorded on #2894 and from
  `src/hle/net/AGENTS.md` on `main` — I did not re-derive them from the module myself.
* **No network service exists to check the failure path against**, by design; "what hardware returns
  when the network *is* reachable" is untested and untestable here.
* No Vulkan-execution or snapshot test ran (they cannot build on this box, see above), and none is
  reachable from an HLE registration change.
* Windows and macOS were not built. The file is plain C++ with no platform surface.

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -1081,6 +1081,12 @@ if(TARGET prosper_core)
   target_link_libraries(test_http PRIVATE prosper_core)
   add_test(NAME http_hle COMMAND test_http)
 
+  # #2894: libSceHttp2 was unregistered, so a request nobody sent reported SCE_OK and its response
+  # getters handed the guest untouched out-parameters.
+  add_executable(test_http2 tests/hle/net/test_http2.cpp)
+  target_link_libraries(test_http2 PRIVATE prosper_core)
+  add_test(NAME http2_hle COMMAND test_http2)
+
   add_executable(test_guard_recursion tests/hle/test_guard_recursion.cpp)
   target_link_libraries(test_guard_recursion PRIVATE prosper_core Threads::Threads)
   add_test(NAME guard_recursion COMMAND test_guard_recursion)
--- a/prosper/docs/PGA_TOUR_2K25_STATUS.md
+++ b/prosper/docs/PGA_TOUR_2K25_STATUS.md
@@ -84,6 +84,16 @@ modules), so the range selects the real `module_start` and never a C++ static co
 
 ## The current blocker
 
+> **Updated 2026-09-11 (#2894).** The libSceHttp2 false success described below is **fixed**:
+> `src/hle/net/hle_http2.cpp` now implements the library — real ids for the local objects, recorded
+> state for the setters, and a libSceNet `ENETUNREACH` from `sceHttp2SendRequest` with every
+> response getter failing and leaving its out-parameters untouched. **What that does NOT establish
+> is where this title now stops.** The change was written and tested on a machine with no
+> `PPSA17952` dump and no PS5, so no boot was run for it and the fault at `eboot+0x142d5e0` is
+> *predicted* to be gone rather than *observed* to be gone. The rung stays 0 until somebody with
+> the dump re-runs the route in `## Reproduction route` and records what they see. Everything below
+> is the pre-fix record of how the fault was reached, and is left as written.
+
 With the handshake fixed the title runs ~6x longer and reaches GPU submission, then dies on a
 **worker thread named `Background Job.`**:
 
--- a/prosper/src/hle/dispatch/dispatch.hpp
+++ b/prosper/src/hle/dispatch/dispatch.hpp
@@ -259,6 +259,9 @@ size_t savedata_tx_resource_live_count();
 void register_service_hle();
 // libSceHttp local URI helpers; called by register_builtin_hle().
 void register_http_hle();
+// libSceHttp2: local id lifecycle, recorded settings, and an honest offline failure on the
+// request/response path; called by register_builtin_hle().
+void register_http2_hle();
 // libSceFont/libSceFontFt opaque lifecycle and deterministic metrics; called by register_builtin_hle().
 void register_font_hle();
 // libSceFiber cooperative guest-stack switching; called by register_builtin_hle().
--- a/prosper/src/hle/libc/hle_libc.cpp
+++ b/prosper/src/hle/libc/hle_libc.cpp
@@ -1070,6 +1070,7 @@ void register_builtin_hle() {
     register_file_hle();     // file I/O (stdio + POSIX, /app0 translation)
     register_service_hle();  // PS5 system services (user/NP/mouse/appcontent/dialog)
     register_http_hle();     // libSceHttp local URI parsing
+    register_http2_hle();    // libSceHttp2 local ids + honest offline request failure
     register_font_hle();     // libSceFont opaque handles + deterministic text/metric fallback
     register_fiber_hle();    // libSceFiber cooperative guest-stack execution
     register_ult_hle();      // libSceUlt: NOT implemented — fail-visible counted stubs (#1603)
--- a/prosper/src/hle/net/AGENTS.md
+++ b/prosper/src/hle/net/AGENTS.md
@@ -1,7 +1,10 @@
 # `src/hle/net` — Sony networking libraries
 
 Reimplementations of the PS5 networking libraries that prosper answers itself. Today that is
-**libSceHttp** (`hle_http.cpp`): the URI helper family and the library/template id lifecycle.
+**libSceHttp** (`hle_http.cpp`): the URI helper family and the library/template id lifecycle; and
+**libSceHttp2** (`hle_http2.cpp`): the whole v2 library, split between a local object graph
+(contexts, templates, requests, cookie boxes, and the settings recorded on them) and an honest
+offline failure on everything from `sceHttp2SendRequest` onwards.
 
 ## The policy this folder exists to hold
 
@@ -41,9 +44,10 @@ library propagates the **raw libSceNet error**, encoded `0x80410100 | BSD errno`
 This folder is smaller than "prosper's networking" — several networking surfaces are answered
 elsewhere, and looking for them here is the mistake to avoid:
 
-- **`sceHttp2Init`, `sceNetCtlGetState` and the NetCtl/NP service surface live in
-  `../service/hle_service.cpp`**, not here. That is why `libSceHttp2` looks absent from this folder
-  while one of its entry points is already registered.
+- **`sceNetCtlGetState`, `sceNetPoolCreate`, `sceSslInit` and the rest of the NetCtl/NP service
+  surface live in `../service/hle_service.cpp`**, not here. `sceHttp2Init` used to be on that list
+  and is not any more (#2894): it moved into `hle_http2.cpp` with the rest of its library, because
+  the context id it returns has to be a slot that `sceHttp2CreateTemplate` can validate against.
 - Answers must not contradict each other across that boundary. prosper already tells the guest it
   is offline there — `sceNetCtlGetState` writes `SCE_NET_CTL_STATE_DISCONNECTED` and
   `sceNetCtlGetInfo` returns `SCE_NET_CTL_ERROR_NOT_CONNECTED` (`hle_service.cpp:4795`, `:4830`) —
--- a/prosper/src/hle/net/hle_http2.cpp
+++ b/prosper/src/hle/net/hle_http2.cpp
@@ -1,0 +1,653 @@
+// libSceHttp2 (#2894). Until now the library was entirely unregistered except sceHttp2Init, so
+// every one of its 55 exports fell to the dispatcher's `return 0` -- and 0 is SCE_OK. A request
+// prosper never sent reported success, and sceHttp2GetAllResponseHeaders reported success while
+// writing nothing to its out-parameters, so the guest parsed the pointer it already had. In
+// PGA TOUR 2K25 that pointer is NULL and its response-header scanner at eboot+0x142d5e0 dies on
+// `movzx edx,BYTE PTR [r11+rcx*1]` with r11 = 0.
+//
+// The rule this file follows is the one src/hle/net/AGENTS.md states, in both directions:
+//
+//   * Anything answerable from LOCAL state is implemented, not stubbed. Creating a template, a
+//     request or a cookie box really does succeed offline -- it allocates a local object -- so it
+//     returns a real id from a real table. Setters really do record local state, so they record it
+//     and then answer SCE_OK. A getter that reads back state prosper itself recorded answers it.
+//   * Anything that needs a NETWORK answer FAILS, and leaves every out-parameter untouched.
+//     There is no PSN behind this, so SCE_OK is not a true answer, and SCE_OK with an unwritten
+//     out-parameter is the exact shape that crashes callers.
+//
+// Nothing here manufactures an HTTP response: there is no synthesised status code, no fabricated
+// header block and no body. The guest's own error path runs instead, which is what it is written
+// to do -- the blocking title classifies the send result and returns cleanly on any non-zero,
+// and it is prosper's SCE_OK that walked it past that check.
+//
+// WHERE THE CONTRACTS COME FROM. The NID<->name map is the PS5 3.20 firmware library dump
+// (libSceHttp2.c, 55 exports + a dummy). Argument shapes for SendRequest, GetAllResponseHeaders
+// and GetStatusCode are read off the shipped libSceHttp2.sprx export thunks and recorded on
+// #2894; the rest follow the v1 libSceHttp family, whose published contract takes the object id
+// as the first argument everywhere. That "id first" property is the only shape assumption any
+// validation below depends on. CONFIDENCE: HIGH on it; the per-call argument meanings are marked
+// individually where they are weaker.
+//
+// WHAT IS DELIBERATELY NOT MODELLED. prosper models cookie boxes as real objects with real ids,
+// but does not model their CONTENTS. So create/delete/bind/flush are implemented, while the calls
+// that must produce or consume cookie DATA (GetCookie, CookieExport, CookieImport, AddCookie,
+// GetCookieStats, GetMemoryPoolStats) fail with out-parameters untouched. Answering SCE_OK there
+// would claim a store that does not exist, and a later read would contradict it.
+#include "hle/net/hle_http2.hpp"
+#include "hle/dispatch/dispatch.hpp"
+
+#include <cstddef>
+#include <cstdint>
+#include <cstring>
+#include <mutex>
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace prosper {
+
+#define HLE(name) static PROSPER_SYSV_ABI uint64_t name(uint64_t a0, uint64_t a1, uint64_t a2, \
+                                                        uint64_t a3, uint64_t a4, uint64_t a5)
+
+namespace {
+
+constexpr size_t kMaxUrlBytes = 64 * 1024;
+constexpr size_t kMaxHeaderBytes = 16 * 1024;
+
+// One unified id space for every object the library hands out. Separate per-kind tables would let
+// a template id and a request id collide numerically, and then a validator could not tell a live
+// id of the wrong class from a live id of the right one. Ids start at 1, so 0 -- the value the
+// dispatcher used to answer -- is never a valid id of any kind.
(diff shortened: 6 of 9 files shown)


mattias800:
## The `origin/main` ctest baseline, so the branch's rc=8 is checkable rather than asserted

Measured on the same box, same compiler, same configure line, in a **second worktree checked out at
`origin/main` with no changes at all** — built and run before a single byte of #2930 work touched
that tree.

| | tests discovered | failed | rc |
| --- | --- | --- | --- |
| `origin/main` | 458 | 113 | 8 |
| this branch | 459 (+`http2_hle`) | 112 | 8 |

**The two failing SETS are identical except for one name**, and it goes the other way:
`file_binary_roundtrip` failed on the `origin/main` run and passes on this branch. Re-run three
times in the baseline tree it is rc=0 every time, so it is flaky here (file I/O against the Windows
filesystem), not a signal. No test fails on this branch that does not already fail on `origin/main`.

Both runs fail on the same 112 Vulkan-execution / live-renderer / GPU-storage cases, which is
environmental: this box has Vulkan **1.3.275** headers and
`frontends/shared/device/vulkan_runtime.hpp` needs `VK_API_VERSION_1_4`, so those targets do not
compile here and ctest reports them `Not Run`. That is also why rc is 8 on an untouched `main`.

Command, in both trees:

```
cmake --build prosper/build-linux -j12 -- -k
ctest --test-dir prosper/build-linux --no-tests=error -j8
```

mattias800 (commented):

## The judgement that mattered, and it went the right way

The failure mode being fixed is FALSE SUCCESS (#2081): an unregistered library answers `0`, `0` is
`SCE_OK`, and a request prosper never sent is reported as sent. The tempting fix is to make the calls
"work" — return a plausible status code and a synthesized header block — because that makes a title
progress in a screenshot. It is also the one thing a compatibility layer must not do.

This does the opposite, and I checked the code rather than the PR body:

- `sceHttp2SendRequest` returns the libSceNet `ENETUNREACH` error (`0x80410133`), **not** `SCE_OK`.
- Every response getter returns `response_unavailable(*req)` and writes **nothing** to its
  out-parameters. `sceHttp2GetAllResponseHeaders` — the call that crashed PGA TOUR 2K25 by answering
  `SCE_OK` with `**out` untouched — now reports the failure instead.
- Nothing anywhere synthesizes a status code, header block or body.

So the guest's own error path runs, which is the correct behaviour for a preservation tool with no
network service behind it, and is the honest result rather than a manufactured one.

## The async reasoning is the part I would have missed

`sceHttp2SendRequestAsync` fails **at submission** rather than accepting and never completing, with
the reason stated: an accepted async send that never completes leaves the guest waiting forever on a
request that does not exist. That is a worse outcome than an honest failure and it would present as a
hang rather than as an error — much harder to diagnose. `sceHttp2WaitAsync` answers consistently.

Getting the async case right required thinking past "return an error" to "what does the guest do
next", and it is not something the issue asked for.

## Structural fix to the original defect class

The id space is unified and **0 is never handed out**, so a returned id of `0` can no longer be
mistaken for a valid handle — which is exactly what `sceHttpInit`/`CreateTemplate` returning `0` did.
The defect is now unrepresentable rather than merely not-currently-occurring.

`sceHttp2Init` was moved out of `hle_service.cpp` for a real reason: the context id must be a slot
that `CreateTemplate` can validate against, which it cannot be if it is minted elsewhere.

## Coverage

All 55 exports registered, so nothing in the library falls through to the dispatcher's `return 0`.
That is the right granularity — registering only the NIDs the blocking title happens to call would
leave the same false-success trap armed for the next title through a different entry point.

## Arms

Six mutations, each rebuilt and re-run, green → red → green: library unregistered; send returns
`SCE_OK`; getters return `SCE_OK` with out-params untouched; create returns id 0; `Term` leaks the
tree; validator accepts id 0. Each targets a distinct way the fix could be hollow, including the two
that would look correct from the outside.

Linux (the platform that counts here): new `test_http2` 48 assertions rc=0; `ctest --no-tests=error`
459 discovered / 112 failed / rc=8 against an `origin/main` baseline built in a third worktree giving
458 / 113 — **failing sets identical** except `file_binary_roundtrip`, which fails on the baseline and
passes here (flaky, rc=0 on three re-runs). No failure added. The 112 are the Vulkan-1.4 header gap
this box has.

## What is honestly unverified, and is stated as such

No `PPSA17952` dump, no PS5, no `libSceHttp*.sprx` on this machine, no network service — so no boot
was run and **the PGA TOUR 2K25 fault is predicted gone, not observed**. The tracker stays at rung 0
and `docs/PGA_TOUR_2K25_STATUS.md` says so. Argument shapes for the three critical calls come from
the reviewed record on #2894 rather than the author's own read of the module.

That is the correct disposition. A rung claim on an unbooted title would have been the more damaging
error here, and declining to make one costs nothing.

Also filed #3545 (the `0x80410100 | errno` encoding now duplicated in three places), which is a real
follow-up rather than a note.

Issue #3540, opened 2026-09-11
The SessionStart hook hardcodes python3, so on Windows the charter-freshness check silently never runs


mattias800:
## Summary

`.claude/settings.json:8` hardcodes the SessionStart hook's interpreter as `"command": "python3"`:

```json
{ "type": "command",
  "command": "python3",
  "args": ["${CLAUDE_PROJECT_DIR}/prosper/tools/session_start.py", "--repo", "${CLAUDE_PROJECT_DIR}", "--hook"],
  "timeout": 20 }
```

**A stock python.org install on Windows does not provide `python3`.** It installs `python.exe` only.
What answers the name `python3` on a default Windows 11 box is the **Microsoft Store App Execution
Alias stub**, which exits **9009** without running anything and prints an advertisement:

```
Python was not found; run without arguments to install from the Microsoft Store, or disable
this shortcut from Settings > Apps > Advanced app settings > App execution aliases.
```

So on Windows the hook does not merely fail — it launches a program that is not Python, gets a
non-zero exit, and produces no `hookSpecificOutput`.

## Why this is worse than a red test

`CLAUDE.md` opens by telling every agent to run `session_start.py` at session start, and says
*"Claude Code runs it through the checked-in SessionStart hook; other agents follow root
`AGENTS.md`."* The hook is the mechanism by which a Windows session learns that its checkout, and
therefore **the charter it has just loaded**, differs from `origin/main`.

That check has been silently absent on Windows. A failing SessionStart hook is quiet: the session
starts normally, the charter loads, and nothing anywhere says the freshness comparison never ran.
This is the #2149 shape — *an unarmed instrument reporting as a confident zero* — sitting on the one
check whose entire job is to tell you your instructions may be stale.

## Evidence

Measured on Windows 11, `origin/main` `3947f269`, Windows/MinGW build, 2026-09-11.

```
$ ctest --test-dir <build> --no-tests=error -R ^session_start$
    Ran 14 tests in 32.425s
    FAILED (failures=6)
    0% tests passed, 1 tests failed out of 1
```

The assertion that fires is `test_session_start.py:65`, inside `hook()` — the helper whose comment
says it *"execute[s] the actual exec-form hook argv"*, i.e. it runs exactly what Claude Code runs:

```
File "prosper/tools/test_session_start.py", line 65, in hook
    self.assertEqual(proc.returncode, 0, proc.stderr)
AssertionError: 9009 != 0 : Python was not found; run without arguments to install from the
[...]

Issue #131, opened 2026-07-08
[defect] robustImageAccess required by the storage-image recompile contract but never enabled at vkCreateDevice


mattias800:
From the #72 backlog. The recompiler storage-image load path issues OpImageRead for ALL invocations (incl. EXEC-inactive lanes with possibly-OOB coords) on the premise that the device enables robustImageAccess — but render_runner.h:113 and image_compute_runner.h:57 enable only robustBufferAccess. An OOB storage-image read without robustImageAccess is UB (device-lost at worst). Fix: chain VkPhysicalDeviceImageRobustnessFeaturesEXT{robustImageAccess} (feature-query guarded) into both vkCreateDevice, or predicate the read. Introduced e3951c1.

mattias800:
CLAIMING: claude-session-896257ae (worktree bughunt-batch) | branch fix/issue-131-robust-image | 2026-07-09T00:48:14Z

PR #3546, opened 2026-09-11
Enable image robustness on the compute backend's own Vulkan device (#3531)

### LINKED ISSUE #3531: live compute's own Vulkan device never enables robustImageAccess, so every recompiled storage-image read is UB on the no-renderer path
## Summary

The live compute backend's **own** Vulkan device (`frontends/shared/live/live_compute.cpp`,
`vkCreateDevice` at the end of the standalone-init path) enables `robustBufferAccess`,
`shaderInt64`, `shaderImageGatherExtended`, `shaderStorageImageReadWithoutFormat` /
`WriteWithoutFormat`, `shaderStorageBufferArrayNonUniformIndexing` and
`shaderBufferInt64Atomics` — and **never chains
`VkPhysicalDeviceImageRobustnessFeatures`**. `grep -n robustImageAccess
frontends/shared/live/live_compute.cpp` returns nothing.

The recompiler requires that feature. `src/gpu/recompiler/rdna2_to_spirv_internal.hpp`, at
`image_read`:

> **OOB CONTRACT:** like a buffer load (which relies on `robustBufferAccess`), the read is issued for
> ALL invocations — including EXEC-inactive lanes whose coordinate may be out of range (e.g. a
> grid-tail bounds check narrowed EXEC). The loaded value for such a lane is discarded by write-back
> predication (`predicate_write`) at the call site, so it is harmless — **PROVIDED the device enables
> IMAGE ROBUSTNESS** (`robustImageAccess` / `VK_EXT_image_robustness`: OOB image reads return 0) …
> **The runtime must enable it (the test harness does).**

So a recompiled compute kernel that reads a storage image deliberately issues out-of-range
`OpImageRead`s for inactive lanes, and relies on image robustness to make them return zero. Without
the feature that is undefined behaviour, and on RADV an out-of-range image read is exactly the shape
that produces a `GCVM_L2_PROTECTION_FAULT` with `CLIENT_ID: (TCP)` and `RW: 0`.

## Why no test catches it

Every harness that exercises this path enables the feature, so the contract is satisfied in tests and
violated only in the shipped standalone path:

- `tests/fixtures/image_compute_runner.h:62-70` — queries `VK_EXT_image_robustness` and chains
  `VkPhysicalDeviceImageRobustnessFeaturesEXT` when `robustImageAccess` is advertised.
- `tests/fixtures/render_runner.h:1486-1510` — same, for the renderer's device, with the comment
  *"OpImageRead OOB must return zero (#131); enable robustImageAccess when supported."*

`tests/gpu/test_storage_image_copy.cpp:75` states the dependency outright: *"lanes 70..127 read AND
write out of the image's range. With `robustImageAccess` (now enabled by …)"*. That test would fail
against a device without the feature — it is simply never run against one.

## When the affected path actually runs

`live_compute.cpp` adopts the renderer's device whenever a renderer exists (it logs *"adopted the
renderer's device (shared, queue family N)"*), and that device does chain the feature. The standalone
path is the `else`: **any run that executes compute with no live renderer.**

That is not a rare configuration. `GAME_COMPAT_ORCHESTRATION.md` § GPU scheduling records that
`boot_trace` calls `register_live_compute()` unconditionally at `boot_trace.cpp:277` and gates only
the renderer on `PROSPER_RENDER` — a "CPU-only" `boot_trace` of `PPSA19244` executed **~135,000
dispatches**. Every one of those ran on this device.

## What to check before fixing

Confirm from a log which device a given run used — the adopt path prints its line, the standalone path
prints `[compute] Vulkan device: <name> (<type>)`. A run on the standalone path that executes any
storage-image kernel is running the UB.

## Suggested fix

Chain `VkPhysicalDeviceImageRobustnessFeatures` in the standalone path exactly as
`render_runner.h:1486-1510` does, and — per this project's own both-directions logging rule, which the
neighbouring `descriptor indexing %s` and `storage-buffer int64 atomics %s` lines already follow —
**print whether it was acquired**, in both directions. The renderer's device currently prints nothing
either way, so there is no run anywhere that can contradict the assumption that it is on; that is the
same silence #2458 shipped a wrong comment behind.

A regression test needs a seam: the feature-chain construction is inline in a long init function in
both places. Extracting "given the queried features, which structs get chained" into a pure function
would make it assertable without a device, and would let one test cover both device paths — which is
the actual invariant, since the two must not diverge.

## Not claimed here

This is **not** established as the cause of any current title failure. It was found while
investigating a GPUVM permission fault on *Hi-Fi RUSH* (`PPSA17168`, #2891) whose run log shows
`[compute] Vulkan device: adopted the renderer's device`, i.e. the **unaffected** path. Filed on its
own merits: a documented contract the shipped code does not meet, invisible to the test suite.
`CONFIDENCE: HIGH` on the gap, `CONFIDENCE: LOW` on any title impact.

Related: #3340, #2891 (same fault class, different device path).

mattias800:
Fixes #3531.

## Failure scenario

`src/gpu/recompiler/rdna2_to_spirv_internal.hpp` (`image_read`) issues an `OpImageRead` for **all**
invocations, including EXEC-inactive lanes whose coordinate may be out of range; the loaded value is
discarded by write-back predication at the call site. That is defined only when the device enabled
**image robustness** (`robustImageAccess`, core in Vulkan 1.3), which makes an out-of-range image read
return zero.

The live renderer's device chained `VkPhysicalDeviceImageRobustnessFeatures`. The compute backend's
**own** device — the one used by any run with no live renderer, and `boot_trace` registers compute
unconditionally — never did. Every recompiled storage-image dispatch on that path was undefined
behaviour; the issue records a "CPU-only" `boot_trace` of `PPSA19244` executing ~135,000 dispatches
there. No test could see it because every harness in the tree chains the feature, so the contract held
in tests and was violated only on the shipped standalone path.

## Behavioural contract

A device may execute recompiled storage-image kernels only when it has all three of
`shaderStorageImageReadWithoutFormat`, `shaderStorageImageWriteWithoutFormat` and `robustImageAccess`.
A device that has the first two and not the third **declines** the work; it does not run it.

## Approach and invariants

- New `frontends/shared/device/storage_image_contract.hpp` (Vulkan-free) holds
  `StorageImageDeviceFeatures` and the pure decision; `frontends/shared/device/image_robustness.hpp`
  holds the Vulkan half that queries the physical device, chains the feature struct into
  `VkDeviceCreateInfo`, and logs the outcome **in both directions**.
- All three device paths go through it: the renderer's device, the compute backend's own device, and
  the adopt path (via a new `SharedVulkanContext::image_robustness`, published as *enabled*, never as
  *supported*, like every other field there). One decision function, so the paths cannot diverge —
  divergence was the bug.
- `image_support` in `live_compute.cpp` is assigned from that verdict and from nowhere else, on both
  paths. When it is false a storage-image dispatch takes the existing fail-visible `decline(...)` path
  with a **new, distinct reason** — `device-lacks-image-robustness` rather than the old
  `device-lacks-storage-image`, because the two causes need different actions and the old message
  pointed at the wrong feature.
- **Chosen deliberately: refuse, do not proceed, and do not kill the run.** The charter's rule is
  fail-visible; refusing the recompiled-storage-image path loses those dispatches loudly (named
  decline reason plus a one-shot warning), while failing device creation outright would take down
  buffer-only compute that is perfectly well defined on such a device. The feature is requested via
  the **core 1.3 struct with no extension name**, because the runtime floor is 1.4 and a 1.4 device
  need not still advertise `VK_EXT_image_robustness` — the same reasoning the neighbouring
  descriptor-indexing and int64-atomics acquisitions already use.

## The regression arms, and what they do and do not prove

Three arms, because no single one is strong on its own. Both without-fix arms were **actually run**;
results below.

1. **`image_robustness` (new, 13 assertions).** The refusing device — format-free features present,
   robustness absent — is constructed **by hand**, outside anything that queries a driver, because no
   device available to this project lacks the feature: a control drawn from a real device could only
   ever exercise the accepting branch (CLAUDE.md's positive-control rule). The strong part is the
   second arm: it calls the **shipped helper** against a real physical device and then walks the
   resulting `VkDeviceCreateInfo::pNext` chain, asserting the struct is really there, really requests
   `VK_TRUE`, and that a struct already on the chain survived. That pins "chained" rather than letting
   a copy of the chaining code assert about itself. It creates no device and submits no GPU work, so
   it runs anywhere an ICD exists and prints an explicit `[skip]` — never a silent pass — where one
   does not.
2. **`game_compute_exec`** now reads `live_compute_storage_image_device()`, the record the **shipped
   backend's own init** writes. That suite runs with no renderer, i.e. on exactly the device path that
   was wrong (`adopted=0` in its output).
3. Both device paths' capability is computed by the same function, asserted with hand-built inputs.

**What none of them proves:** that a device lacking `robustImageAccess` is actually protected at
runtime — no such device exists here to run it on. That branch is covered by construction and by the
hand-built decision arms only. `CONFIDENCE: HIGH` on the acquisition and the refusal wiring,
`CONFIDENCE: MED` on the untestable refusing-device execution path.

Also note what arm 2 does *not* catch: on a device that supports the feature, deleting only the
chaining line leaves the verdict (and therefore `image_support`) true, so the dispatch assertions stay
green. That is arm 1's job, and it is why arm 1 exists.

- Affected: the compute backend's own device (acquires the feature; storage-image capability now gated
  on it), the renderer's device (same acquisition through the shared helper; behaviour unchanged, but
  it now logs the outcome), the adopt path (inherits the verdict instead of assuming `image_support`).
- Deliberately unaffected: `tests/fixtures/image_compute_runner.h` keeps its extension-name-gated
  acquisition — its instance is not guaranteed to be 1.1+, where `vkGetPhysicalDeviceFeatures2` lives.
  `frontends/prosper-app/main.cpp`'s present device is not changed: it blits a swapchain image and
  executes no recompiled kernels.

## Risks

- Enabling image robustness can cost performance on out-of-range accesses. The renderer has shipped
  with it since #131, so this only brings the compute-only path to parity.
- The renderer now prints one unconditional device line (`[vk] image robustness …`). That is the
  point — the issue notes there was previously no run anywhere whose log could contradict the
  assumption that it was on.
- A device that genuinely lacks the feature now loses its storage-image dispatches instead of running
  them undefined. That is the intended trade, and it is loud.

## Verification (all on Windows / MinGW, Vulkan SDK 1.4, real GPU)

```
cmake -S prosper -B prosper/build-win -G Ninja -DCMAKE_BUILD_TYPE=Release \
      -DCMAKE_C_COMPILER=<mingw>/gcc.exe -DCMAKE_CXX_COMPILER=<mingw>/g++.exe \
      -DCMAKE_MAKE_PROGRAM=<mingw>/ninja.exe -DGAME_DUMP=<DUMP_ROOT>/PPSA24651-app0
cmake --build prosper/build-win            # exit 0
ctest --test-dir prosper/build-win --no-tests=error
```

- `ctest -N` reports **335 registered tests**. Full run: **1 failure — `session_start`** (pre-existing,
  #3540, unrelated to this diff), 2 skipped (`refactor_survey_measurement`, `hle_calls_value_capture`);
  ctest exit **8**.
- `image_robustness`: PASS, **13 assertions**, `real device: robustImageAccess advertised=1 chained=1`.
- `game_compute_exec`: PASS, **627 assertions**, `[compute-device] adopted=0 robustImageAccess=1
  storage_image_capable=1` — i.e. the standalone path, which is the one that was broken.

### Without-fix arms (both run, both red)

- Removed the chaining block from `acquire_storage_image_device_features` → `image_robustness` fails
  `VkPhysicalDeviceImageRobustnessFeatures reaches the VkDeviceCreateInfo pNext chain`, 12/13
  assertions, exit 1, reproduced 3 of 3. Restored → PASS 13/13.
- Reverted `live_compute.cpp`'s own-device path to the pre-fix `image_support = <format-free features>`
  with no acquisition → `game_compute_exec` fails 3 assertions (`published the storage-image device it
  initialized`, `acquired robustImageAccess`, `may execute recompiled storage-image kernels`), exit 1,
  with `[compute-device] adopted=0 robustImageAccess=0 storage_image_capable=0`. Restored → PASS.

Notably, in that second arm **none of the other 624 assertions moved** — the suite's storage-image
dispatches still passed, because this GPU supports the feature whether or not it is requested. That is
issue #3531 in one line, and it is why the arm had to be written against the device record rather than
against dispatch results.

### An instrument note on the arm itself (second commit)

The first draft of the chain walk read each node through a `VkBaseInStructure*`. That is a
strict-aliasing violation even though the structs share an initial sequence, and it was not
theoretical: GCC 15 at `-O2` turned the without-fix run into a **SIGSEGV** instead of the clean failure
the arm exists to produce, and adding an unrelated `fprintf` made the fault vanish. The walk now copies
each header with `memcpy`, and the test runs unbuffered so a fault cannot swallow the arms that already
reported.

Not run: Linux. This machine's WSL has Vulkan 1.3 headers and cannot build the live-renderer chain at
all (`VK_API_VERSION_1_4`), so every Vulkan-linking target here is Windows-verified; CI covers Linux.
No snapshots were run — this diff changes device feature acquisition, not rendering, and the full
Windows suite including the render and GPU-execution tests is green.

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -4001,6 +4001,20 @@ if(TARGET prosper-app AND TARGET prosper_video_vaapi)
   message(STATUS "prosper-app: FFmpeg/VA-API video enabled")
 endif()
 
+# Image robustness (#3531): the device-side half of the recompiler's out-of-bounds image-read
+# contract, and the refusal that replaces executing it undefined. Registered OUTSIDE the per-platform
+# blocks because the contract is not platform-specific and the bug it guards was a divergence between
+# two device paths -- a guard built on only one host would have the same blind spot. Queries a
+# physical device but creates none, so it runs wherever an ICD exists and reports a skip (never a
+# silent pass) where one does not.
+if(TARGET Vulkan::Vulkan AND TARGET prosper_core)
+  add_executable(test_image_robustness tests/shared/device/test_image_robustness.cpp)
+  # Spells `shared/...` and `gpu/...` includes, which resolve from `frontends` and `src`.
+  target_include_directories(test_image_robustness PRIVATE frontends src)
+  target_link_libraries(test_image_robustness PRIVATE prosper_core Vulkan::Vulkan)
+  add_test(NAME image_robustness COMMAND test_image_robustness)
+endif()
+
 # Shared-device adoption guard (#1091). prosper_live_renderer only exists in frontend-enabled
 # configurations and is defined later in this file than the core test block, so register this
 # after the fact and only when the target is actually present.
--- a/prosper/frontends/shared/device/AGENTS.md
+++ b/prosper/frontends/shared/device/AGENTS.md
@@ -4,3 +4,10 @@ Loader/device requirements and selection shared by live frontends and execution 
 Pipeline-cache files belong here: they store opaque driver data, not guest shaders or game assets.
 Cache compatibility checks reject damaged or mismatched files; they cannot guarantee a driver
 accepts its own data safely. Keep disk loading opt-in while the recorded NVIDIA failure remains.
+
+Feature acquisition that more than one device path needs belongs here rather than beside whichever
+device was written first: the renderer's device, the compute backend's own device and the adopt path
+must answer the same question the same way. `image_robustness.hpp` is that for the recompiler's
+out-of-bounds image-read contract, and `storage_image_contract.hpp` holds the Vulkan-free decision so
+the refusing branch is testable without a driver -- no device here lacks the feature, so a test drawn
+from a real device can only exercise the accepting one.
--- a/prosper/frontends/shared/device/image_robustness.hpp
+++ b/prosper/frontends/shared/device/image_robustness.hpp
@@ -1,0 +1,56 @@
+#pragma once
+
+// Vulkan acquisition of the storage-image device contract (#3531). The contract itself -- what the
+// recompiler needs and why -- is in storage_image_contract.hpp; this header is only the part that
+// must speak Vulkan: query the physical device and chain the feature into VkDeviceCreateInfo.
+
+#include "shared/device/storage_image_contract.hpp"
+
+#include <vulkan/vulkan.h>
+
+#include <cstdio>
+
+namespace prosper::frontend {
+
+// Query `physical` for image robustness, chain VkPhysicalDeviceImageRobustnessFeatures into `dci`
+// when it is advertised, and report what the resulting device will have. `storage` must outlive the
+// vkCreateDevice call -- it becomes part of the pNext chain.
+//
+// The feature is core in Vulkan 1.3 and the runtime floor is 1.4 (kVulkanRuntimeVersion), so the
+// promoted extension name is deliberately NOT requested: a 1.4 device need not still advertise
+// `VK_EXT_image_robustness`, and gating on the string would disable the feature on devices that
+// have it. This matches how the neighbouring descriptor-indexing and int64-atomics features are
+// acquired.
+//
+// Logs in BOTH directions. A line that prints only on success makes the failing case silent, which
+// is precisely how #3531 survived: there was no run anywhere whose log could contradict the
+// assumption that robustness was on.
+inline StorageImageDeviceFeatures acquire_storage_image_device_features(
+    const char* owner, VkPhysicalDevice physical, bool read_without_format,
+    bool write_without_format, VkPhysicalDeviceImageRobustnessFeatures& storage,
+    VkDeviceCreateInfo& dci) {
+    storage = VkPhysicalDeviceImageRobustnessFeatures{
+        VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_IMAGE_ROBUSTNESS_FEATURES};
+    VkPhysicalDeviceFeatures2 features2{VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FEATURES_2};
+    features2.pNext = &storage;
+    vkGetPhysicalDeviceFeatures2(physical, &features2);
+    const StorageImageDeviceFeatures features = storage_image_device_features(
+        storage.robustImageAccess == VK_TRUE, read_without_format, write_without_format);
+    if (features.robust_image_access) {
+        // Request exactly the one bit in use; the query struct carries no others.
+        storage.pNext = const_cast<void*>(dci.pNext);
+        dci.pNext = &storage;
+    }
+    std::fprintf(stderr,
+                 "[%s] image robustness %s (core feature; recompiled storage-image reads are "
+                 "out-of-bounds by design and need it)\n",
+                 owner, features.robust_image_access ? "ENABLED" : "unavailable");
+    if (!features.robust_image_access && read_without_format && write_without_format)
+        std::fprintf(stderr,
+                     "[%s] refusing recompiled storage-image work on this device: without "
+                     "robustImageAccess an inactive lane's out-of-range OpImageRead is undefined\n",
+                     owner);
+    return features;
+}
+
+} // namespace prosper::frontend
--- a/prosper/frontends/shared/device/storage_image_contract.hpp
+++ b/prosper/frontends/shared/device/storage_image_contract.hpp
@@ -1,0 +1,55 @@
+#pragma once
+
+// The storage-image device contract -- what a Vulkan device must offer before prosper may execute a
+// recompiled storage-image kernel on it (#3531). Deliberately Vulkan-free so the decision can be
+// unit-tested with no driver, no device and no Vulkan headers, and so the compute backend's public
+// header can report the capability without pulling Vulkan into every consumer.
+//
+// `src/gpu/recompiler/rdna2_to_spirv_internal.hpp` (image_read) issues an OpImageRead for ALL
+// invocations, including EXEC-inactive lanes whose coordinate may be out of range: the loaded value
+// is discarded by write-back predication at the call site. That is defined only when the device
+// enables IMAGE ROBUSTNESS (`robustImageAccess`, core in Vulkan 1.3; `VK_EXT_image_robustness`
+// before that), which makes an out-of-range image read return zero. Without it the kernel is
+// undefined behaviour, and on RADV an out-of-range image read is the shape that raises
+// GCVM_L2_PROTECTION_FAULT (CLIENT_ID: TCP, RW: 0).
+//
+// Every device prosper executes recompiled storage-image work on must therefore acquire the feature,
+// and every device that CANNOT must refuse that work rather than run it. This is the single place
+// that decision is made, so the renderer device, the compute backend's own device and the adopt path
+// cannot diverge -- divergence is what #3531 was: the renderer chained the feature, the compute
+// backend's own device did not, and no test could tell because every harness chained it.
+
+namespace prosper::frontend {
+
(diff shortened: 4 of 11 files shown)


mattias800 (commented):

## The thing I checked first: does "refusing" actually refuse?

A log line reading *"refusing recompiled storage-image work on this device"* while the work still
executes would be strictly worse than the original bug — it would make the UB look handled. So I
traced the verdict to a behavioural gate rather than reading the message.

It refuses. `storage_image_capable()` is `robust_image_access && read_without_format &&
write_without_format`, `image_support` is assigned from it, and `live_compute.cpp:6445` is
`if (has_storage_images && !ctx.image_support)` — the dispatch is skipped. The change reuses the
**pre-existing** loud-decline mechanism (which already covered the without-format features) by making
robustness participate in the same predicate, rather than inventing a second refusal path. That is
the right shape: one gate, one reason, and the new condition inherits the existing behaviour on day
one.

## The divergence that caused #3531 is closed structurally, not by fixing three call sites

`image_support` is written in exactly two places — the adopt path (`:3328`) and the own-device path
(`:3497`) — both from `storage_image_features`, with the comment stating it is assigned *"from
`storage_image_features` and from nowhere else, on BOTH device paths, so the capability and the
feature acquisition cannot drift apart."*

That matters because #3531 **was** a drift bug: the renderer device chained the feature and the
compute backend's own device did not. Fixing the one missing chain would have left the next device
path free to repeat it. Hoisting the decision into `storage_image_contract.hpp` — deliberately
Vulkan-free, so the contract is testable without a device — and having all three paths acquire
through one helper is the fix that stops it recurring.

The contract header also records *why* the recompiler needs the feature, quoting the OOB contract at
`rdna2_to_spirv_internal.hpp` and naming the RADV signature (`GCVM_L2_PROTECTION_FAULT`,
`CLIENT_ID: TCP`, `RW: 0`). That is the connection a future reader would otherwise have to
reconstruct from two files.

## Arms

Both without-fix directions were run and both went red:

- chaining removed → the new `image_robustness` test fails the pNext-chain assertion, 3 of 3 runs;
- `live_compute`'s own-device path reverted → `game_compute_exec` fails 3 assertions reporting
  `robustImageAccess=0`.

**The author reported the asymmetry rather than letting it pass as symmetric coverage**: with the
chaining removed but the verdict intact, the dispatch assertions stay green — that is arm 1's job,
not arm 2's. Saying which arm covers what is the difference between two tests and one test run
twice, and it is the detail most reports omit.

## The self-inflicted trap they caught, which is worth naming

A `VkBaseInStructure` pNext walk in their own new test was a strict-aliasing violation that turned
the **without-fix** run into a SIGSEGV under GCC 15 `-O2`. A negative arm that crashes reads as "the
mutation broke it" — i.e. as confirmation — so this would have produced a red arm for the wrong
reason and nobody would have looked twice. Now memcpys each header. Finding that in one's own
apparatus, unprompted, is the behaviour `CLAUDE.md`'s instrument-trap list exists to encourage.

Windows/MinGW full build exit 0; `ctest --no-tests=error` 335 registered with the single
`session_start` failure, which is pre-existing and mine (#3540). New `image_robustness` test 13/13;
`game_compute_exec` 627 assertions reporting `adopted=0 robustImageAccess=1`. Merge gate GREEN, head
matches branch tip.

Per the owner's standing instruction, non-Linux CI results are not treated as blockers here.

## Non-blocking

`storage_image_capable()` now conflates three independent reasons a device may be unfit into one
boolean, so the decline message must carry the reason — it does, on both paths. Worth keeping that
way if a fourth condition is ever added; the moment the message stops naming which condition failed,
this becomes an instrument that says "no" without saying why.

PR #3547, opened 2026-09-11
fix(recompiler): OpAtomicLoad is not a memory write in the wave-width analysis

### LINKED ISSUE #3504: wave width independence: OpAtomicLoad is classified as a memory write, costing precision
Found in review of #3480, after the `writes_memory` unification landed. **Precision only — it fails in
the refusing direction, so nothing renders wrongly because of it.**

## The defect

`spirv_op_is_atomic` covers `OpAtomicLoad` (227), and `writes_memory` is now defined as "an
observable effect that is not a store to a colour output" minus an explicit
observable-but-writes-nothing list — `OpKill`, `OpTerminateInvocation`,
`OpDemoteToHelperInvocation`, `OpControlBarrier`, `OpMemoryBarrier`.

`OpAtomicLoad` is a pure read and is not on that list, so a fragment module containing one is
classified as writing memory. That disables the uniform arm's storage-buffer source for the whole
module, so every constant-buffer load stops counting as uniform and the module's votes stop clearing.

It is the same shape that cost Evergate 10 of its 11 provable modules when barriers were first folded
in, one opcode over.

## Why it costs nothing today

- **No producer for a lone atomic load.** The emitter emits `OpAtomicLoad` at
  `rdna2_to_spirv_internal.hpp:3114` and `:3321`, and both are the head of a CAS loop whose
  `OpAtomicCompareExchange` follows immediately (`:3126`, `:3332`). A CAS genuinely writes, so
  `writes_memory` is true on those modules regardless.
- **Zero atomics of any kind across the 49-module corpus.**

## The fix

Add `OpAtomicLoad` to the observable-but-writes-nothing exclusion, in both
`rdna2_to_spirv.cpp` and `wave_width_independent.py`. One line each, plus a fixture pair — a module
with a lone `OpAtomicLoad` must stay provable, and the same module with `OpAtomicIAdd` must not.

`OpMemoryNamedBarrier` is in neither list either, but it needs a capability Vulkan does not expose.

## Why it was not folded into #3480

It arrived after that PR's approval, and it is precision rather than soundness — so taking it there
would have meant another push, CI cycle and re-approval to change a verdict nothing currently
reaches. The reviewer's own recommendation was that it is "a one-line addition if you're in the file,
not a reason to cut a round".

mattias800:
Fixes #3504

## The failure scenario

The wave-width-independence analysis (`fragment_spirv_wave_width_independent`, and its Python twin `tools/shader_inspect/wave_width_independent.py`) classified `OpAtomicLoad` as a memory write.

`is_observable_effect` covers every atomic opcode, and since #3480 `writes_memory` has been defined as that predicate minus an explicit observable-but-writes-nothing list (`OpKill`, `OpTerminateInvocation`, `OpDemoteToHelperInvocation`, `OpControlBarrier`, `OpMemoryBarrier`). `OpAtomicLoad` was not on that list. So a fragment module containing a lone atomic load was classified as writing memory, which disables the uniform arm's `StorageBuffer` source for the whole module: every constant-buffer load stops counting as uniform and the module's votes stop clearing.

Same shape as the defect that cost Evergate 10 of its 11 provable modules when barriers were first folded in, one opcode over. It fails in the **refusing** direction, so nothing has ever rendered wrongly because of it.

## Which of the two the predicate turned out to be

The issue framing was "an atomic load reads, so the classification is wrong". That is half right, and the half it gets wrong is the load-bearing half — so it is stated up front rather than in a footnote.

**`is_observable_effect` is NOT a memory-write predicate. It has two callers and only one of them asks about writing.**

| caller | the question it asks |
|---|---|
| `writes_memory` (arm (a)'s gate on `StorageBuffer` uniformity) | *can this shader read back what it wrote?* |
| arm (b)'s control-dependence region scan | *did this vote-guarded region do anything the rest of the draw can see?* |

A barrier writes nothing at all and has been on `is_observable_effect` since the Evergate measurement, precisely because of the second question. So the predicate is about **touching / ordering**, and `writes_memory` is it minus an explicit writes-nothing list.

**Therefore the fix goes in the subtraction, not in the predicate**, and this is a real distinction rather than a stylistic one: removing `OpAtomicLoad` from `is_observable_effect` instead makes the analysis **admit a vote-guarded region containing one**, measured — the `region_atomic(227)` arm flips from NOT PROVEN to independent. That is the permissive mistake this class of change is dangerous for, and it is now pinned by a test in both implementations.

What changed in `is_observable_effect` is its **comment**: it used to say "Atomics both read and write memory the rest of the draw can see", which is false of `OpAtomicLoad` and is the sentence that made the membership look like a write claim.

## That `OpAtomicLoad` does not write — three independent routes

The analysis runs over **SPIR-V**, so SPIR-V's memory model is the governing contract; RDNA2 enters only through the emitter (see *Deliberately unaffected*).

1. **The grammar.** In SPIR-V's unified core grammar `OpAtomicLoad` (227) is `IdResultType IdResult Pointer(IdRef) Memory(IdScope) Semantics(IdMemorySemantics)` — the only atomic with **no `Value` operand at all**. `OpAtomicStore` (228) and `OpAtomicIAdd` (234) both carry one. An instruction that carries no value cannot store one. `CONFIDENCE: HIGH`
2. **The spec's own asymmetry.** SPIR-V forbids `Release` and `AcquireRelease` on `OpAtomicLoad`, and forbids `Acquire` and `AcquireRelease` on `OpAtomicStore` (enforced verbatim in SPIRV-Tools' `source/val/validate_memory_semantics.cpp`: *"MemorySemantics must not use Release or AcquireRelease memory order with OpAtomicLoad"*). `Release` is the **publishing** half of the pair, and denying it to this instruction is the spec saying the instruction publishes nothing. `CONFIDENCE: HIGH`
3. **This repository already agreed, in a file written for another purpose.** `src/gpu/resources/shader_resources.cpp:1083` classifies `OpAtomicLoad` as `mark_pointer(ptr, /*read=*/true, /*write=*/false)`, in its own branch, separate from the read-modify-write atomics two branches down which pass `true, true`. Independent author, independent purpose, same verdict. `CONFIDENCE: HIGH`

The AMD RDNA 2 ISA reference (document 70648) was consulted and is **not load-bearing here**, so nothing is cited from it: the classifier's domain is SPIR-V opcodes, and no RDNA2 instruction in this emitter lowers to a lone atomic load (see below). `CONFIDENCE: LOW` on any hardware-level claim about what a PS5 atomic load lowers to, because this change did not need one and none was established.

## Behavioral contract and invariants

- `writes_memory` means *this module wrote buffer or image memory that a later load in this draw could read back*. `OpAtomicLoad` is excluded from it, alongside the barriers and the terminate/demote forms.
- `is_observable_effect` continues to mean *the rest of the draw can see that this ran*. `OpAtomicLoad` stays in it. A vote-guarded region containing one is still refused.
- `spirv_op_is_atomic` is unchanged, so the leaving-scan that tests every id operand of an atomic for taint still covers `OpAtomicLoad`, and `spirv_op_is_divergent_source` still treats its result as per-lane.
- The two implementations stay behaviourally identical, which is why both were changed together.

## Affected behavior

A fragment module whose only memory effect is one or more `OpAtomicLoad`s is now eligible for arm (a): its `StorageBuffer` loads count as uniform again and its votes can clear. Nothing else moves.

## Deliberately unaffected

- **Arm (b)**, in every form. A vote-guarded region performing an atomic load is still refused.
- **The leaving-scan's treatment of a tainted atomic-load *pointer***, which is also more conservative than it needs to be — a tainted pointer feeding a pure read leaks nothing. Left alone: it is a second loosening in the same dangerous direction and belongs in its own change with its own arms.
- **The emitter.** Both `OpAtomicLoad` sites (`rdna2_to_spirv_internal.hpp:3115`, `:3322`) are the head of a compare-exchange loop whose `OpAtomicCompareExchange` follows within a dozen lines (`:3127`, `:3333`). A CAS genuinely writes, so those modules are classified as writing either way and no shipped shader's verdict changes. There is no RDNA2 instruction in this emitter that lowers to a lone atomic load, which is why the fixtures are hand-built — per the charter's rule that a control drawn from the same source as the null tests the discriminator, not the domain.
- **Anything at runtime.** The C++ predicate's only caller today is `tools/shader_inspect` (`shader_inspect.cpp:347`); it is the admission study for #3464's native-wave32 allowance and is not yet wired into pipeline creation. Blast radius today is what the classifier *reports*.

## Risks

- **This is the loosening direction**, which is the dangerous one: a too-conservative analysis costs a draw its native-width fast path, a too-permissive one ships wrong pixels. Mitigated by requiring both directions of the regression arm, and specifically by `region_atomic(227)` / `mod_region_performs_atomic(227)`, which fail if the exclusion is placed one level too high.
- **Residual, pre-existing, and worth a reviewer's attention. `CONFIDENCE: MED`.** Arm (a) treats a `StorageBuffer` that *this module* does not write as uniform. A buffer written by another stage of the same draw, or by a concurrent dispatch, would break that — and a guest choosing an *atomic* load is weak evidence the location is contended. This change does not introduce the hole: the byte-identical module doing an ordinary `OpLoad` instead of an `OpAtomicLoad` is admitted on current `main` already, so refusing the atomic-load spelling would have been an inconsistency rather than a safety property. Flagged rather than fixed; tightening it is a separate question about the whole `StorageBuffer` premise.

## Verification

**Regression arms, both directions, both implementations.** Each mutation was applied to an otherwise-clean tree, rebuilt, measured red, then reverted.

| arm | mutation | result |
|---|---|---|
| C++ / loosening | remove `case Op_AtomicLoad:` from the `writes_memory` exclusion | `run=1`, `[FAIL] wave width independence: a vote over a buffer this shader only ATOMICALLY LOADS -> NOT PROVEN, expected independent` |
| C++ / permissive | `default: return spirv_op_is_atomic(in.op) && in.op != Op_AtomicLoad;` — the exclusion placed in the observable predicate instead | `run=1`, `[FAIL] wave width independence: a vote-guarded region performing an atomic LOAD, which writes nothing -> independent, expected NOT PROVEN` |
| Python / loosening | drop `OP_ATOMIC_LOAD` from `_writes_memory`'s exclusion tuple | `FAIL - admits an atomic LOAD is observable but is not a memory write` |
| Python / permissive | `return op in ATOMIC_OPS and op != OP_ATOMIC_LOAD` in `_observable` | `FAIL - REFUSES a vote-guarded region performing an atomic LOAD, which writes nothing` |

In every arm exactly one check moved, and in each pair the *other* new check stayed green — so the arms discriminate the two placements from each other, not merely "something changed".

**Linux (WSL Ubuntu-24.04, g++ 13.3, Release).** `prosper_core` plus the struct test. This machine cannot build the Vulkan targets under WSL (headers are 1.3, the repo needs 1.4), so the full suite is the Windows run below.

```
cmake --build prosper/build-linux -j12 --target test_rdna2_spirv_struct     # exit 0
./prosper/build-linux/test_rdna2_spirv_struct                               # exit 0, == PASS ==
  [ok]   wave width independence: 36 hand-built modules classified as expected   (was 34)

python3 prosper/tools/shader_inspect/test_wave_width_independent.py         # exit 0
  36 checks, all passed                                                          (was 34)
```

**Windows (MinGW/UCRT, Ninja, Release, Vulkan SDK 1.4.350.0, real GPU).** Full build and full suite.

```
cmake -S prosper -B prosper/build-win -G Ninja -DCMAKE_BUILD_TYPE=Release -DGAME_DUMP=<DUMP_ROOT>/PPSA24651-app0
cmake --build prosper/build-win -j12                                        # exit 0
ctest --test-dir prosper/build-win --no-tests=error -j8 --output-on-failure  # exit 8
  333 of 334 passed, 2 skipped (refactor_survey_measurement, hle_calls_value_capture)
  1 failed: session_start -- pre-existing on this machine, #3540, unrelated to this diff

ctest --test-dir prosper/build-win --no-tests=error -R "rdna2_spirv_struct|wave_width"  # exit 0
  2/2 passed: rdna2_spirv_struct, wave_width_independence_refuses_the_class
```

**Docs gate** (this PR touches `RECOMPILER_REMAINING.md`): `git ls-files '*.md' -z | xargs -0 python3 prosper/tools/docs/check_numbered_table.py` → exit 0, every table unbroken, every row matching its header. `git diff --check` → exit 0.

## Record

`docs/RECOMPILER_REMAINING.md` § Ruled out gains one row: that `is_observable_effect` is a memory-write test is falsified, together with the measured consequence of acting on it — so the next reader does not "simplify" the two lists back into one.


mattias800 (commented):
APPROVED — independent review on exact head, I am not the author. Merge gated on CI.

## The question I set for this change, and the answer is the right one

I asked the author to establish **what the predicate is for** before changing it, because "an atomic
load obviously doesn't write" is exactly the reasoning that would justify editing the wrong
function. The answer came back measured rather than argued: `is_observable_effect` has **two**
callers asking different questions —

| caller | question |
| --- | --- |
| `writes_memory` | can this shader read back what it wrote? |
| arm (b)'s control-dependence region scan | did this vote-guarded region do anything the draw can see? |

— and a **barrier**, which writes nothing at all, has been on the predicate since the Evergate
measurement, for the second question. So the predicate is about *touching and ordering*, and
`writes_memory` is that set minus an explicit writes-nothing list. The fix belongs in the
subtraction, and that is where it went; `is_observable_effect` got a corrected **comment** only.

I verified this against the file rather than the PR body. `writes_memory` iterates
`is_observable_effect` and subtracts *"Observable and yet writes NOTHING a later load could read
back"* — `Op_Kill`, `OpTerminateInvocation`, the barriers, and now `OpAtomicLoad`. So this is the
file's **established** pattern, not an invention to accommodate one opcode. The surrounding comment
even records why the two lists are defined in terms of one another: keeping them separate let them
drift twice, once on the `227..242` window and once on `OpCopyMemory`.

## The dangerous direction is pinned, and that is the part that matters

Loosening a conservative analysis is the direction that produces silently wrong shader output, which
this project treats as worse than a rejected draw. The arms cover **both placements**, not merely
"something changed":

- `buffer_written_by(227)` / `mod_uniform_source_written_by(227)` must be **admitted** — refused
  without the exclusion.
- `region_atomic(227)` / `mod_region_performs_atomic(227)` must stay **refused** — and are
  *admitted* if the exclusion is moved into `is_observable_effect` instead.

Four mutations run, C++ and the Python twin alike, each red on its own check while the other stayed
green. That second pair is what makes this a real review of the design rather than of the diff: it
demonstrates the tempting simplification is wrong, with a measured flip from NOT PROVEN to
independent, rather than asserting it.

## Epistemic hygiene worth naming

The author fetched AMD ISA doc 70648, got back a **model paraphrase rather than verbatim text**, and
therefore cited **nothing** from it — marking any hardware-level claim `CONFIDENCE: LOW` and noting
it is not load-bearing anyway, since the classifier's domain is SPIR-V. That is precisely the
discipline `CLAUDE.md`'s section on cited claims is asking for, and declining to cite a source you
could not actually read is rarer than it should be.

The citations that *are* load-bearing hold up: `OpAtomicLoad` has the only atomic signature with no
`Value` operand, `Release`/`AcquireRelease` are forbidden on it while `Acquire`/`AcquireRelease` are
forbidden on `OpAtomicStore` (the publishing half denied to a read), and
`shader_resources.cpp:1083` already classifies it `read=true, write=false` **independently** — an
in-repo corroboration reached by a different route, which is the strongest form available here.

## The `## Ruled out` row is the most valuable artifact in this PR

It records the hypothesis as **half falsified** — the load really doesn't write, *and* the predicate
still isn't a write test — names the measured consequence of the wrong fix, and ends with
*"Do not 'simplify' the two lists back into one."* That is a future session saved. It is also
correctly filed: `docs/RECOMPILER_REMAINING.md`, cross-referenced to #3504 and this PR.

## Scope discipline, and one thing deliberately left

No shipped shader's verdict changes: both `Op_AtomicLoad` emission sites
(`rdna2_to_spirv_internal.hpp:3115`, `:3322`) are CAS-loop heads with `OpAtomicCompareExchange`
directly below, so the emitter produces no lone atomic load today. The author says so plainly rather
than implying a title-visible win — the change is precision, and it is presented as precision.

The leaving-scan's over-conservatism for a *tainted atomic-load pointer* was left alone as a second
loosening in the same dangerous direction deserving its own change. I agree.

## Flagged, non-blocking

The author's own `CONFIDENCE: MED` caveat deserves to survive into the record: arm (a) treats a
`StorageBuffer` this module does not write as uniform, which a **concurrent writer elsewhere in the
draw** would break. That is pre-existing — the byte-identical module using a plain `OpLoad` is
already admitted on `main`, so refusing only the atomic-load spelling was an inconsistency rather
than a safety property — but it is a real assumption and this PR widens the set of modules that
rely on it by one opcode. Worth an issue if nobody has filed one.

Also noted: the issue pointed at `rdna2_to_spirv.cpp` and the code actually lives in
`spirv_wave_width.cpp`. Not a defect, but it is the kind of stale pointer that costs the next reader
ten minutes.

Issue #1873, opened 2026-08-03
[Game tracker] Grand Theft Auto V (PPSA04263)


mattias800:
## Current milestone

**Rung 3 — the prologue bank heist renders in full colour, with the HUD and radar over it.** The title and STORY/ONLINE menu render at native 3840x2160, and since #2996 (merged 2026-08-26) so does the world.

> **The world renders only in the game's own Performance graphics mode**, chosen from the landing
> menu before the world loads. The default is Fidelity, where a run started straight into Story
> shows the HUD over a dark scene and looks exactly like a renderer regression. That is a route
> property, not a build property, and it is the likeliest explanation for a "the world regressed"
> report. Route: `scripts/gta5/reach-performance-story.pad`.

> **This body said rung 2 for twelve days after the world started rendering.** `GTA5_STATUS.md` was
> correct throughout; this body, `CLAUDE.md` and `COMPATIBILITY.md` were not, and on 2026-08-29 a
> user reporting a regression was told from a stale document that it had never worked. Corrected
> 2026-09-07 (docs PR #3445), after the project owner confirmed the scene renders correctly on a
> Windows run of `93a0a506`. `GTA5_STATUS.md` is authoritative for this title.

The open frontier is now **framerate**, not the picture: #3407 and its series. #2481 is closed and
superseded by #2542 / #2690.

### Best checked-in visual

[![Grand Theft Auto V — the prologue bank interior in full colour](https://raw.githubusercontent.com/mattias800/prosper/main/assets/screenshots/gta5-prologue-bank-restored.webp)](https://github.com/mattias800/prosper/blob/main/assets/screenshots/gta5-prologue-bank-restored.webp)

The bank interior draws in full colour with the masked gunman, water cooler, holiday cards and radar.

## What works

- Boots through the intro and first-run/language flow.
- Renders the title and interactive STORY/ONLINE menu.
- The deterministic pad route enters Story Mode gameplay.
- Gameplay-only UI renders: tutorial text, HUD and a populated radar.
- **The 3D world renders**: the prologue bank heist draws in full colour, in the game's own
  Performance graphics mode (#2996, 2026-08-26).
- The routed Linux run exits normally without device loss.

## Current blocker

**Framerate, not the picture.** The world has rendered since #2996; what remains is that it renders
slowly. The active work is [#3407](https://github.com/mattias800/prosper/issues/3407) and its series.

- [#2542](https://github.com/mattias800/prosper/issues/2542) and
[...]

PR #3551, opened 2026-09-11
libSceHttp: implement the request path instead of answering SCE_OK (#2930)

### LINKED ISSUE #2930: libSceHttp registers only sceHttpUriParse — sceHttpInit and sceHttpCreateTemplate return a context id of 0 as false success (6 of 8 surveyed titles call them)

mattias800:
Fixes #2930.

## The failure

#2965 and #3295 closed most of this issue — the context/template id lifecycle, `sceHttpTerm`,
context validation and `sceHttpUriBuild`. What both deliberately left open was the **request path**:
`sceHttpCreateConnection*`, `sceHttpCreateRequest*`, `sceHttpSendRequest`, `sceHttpReadData` and the
response getters were still unregistered, so the dispatcher answered every one of them with
`return 0` — and **0 is `SCE_OK`**. Two different wrongs come out of that single default:

* `sceHttpCreateConnection` and `sceHttpCreateRequest` answered **0 as an id**, so a title carried a
  connection and a request handle that were never allocated into every later call;
* `sceHttpSendRequest` reported a request prosper never sent as **sent successfully**, and
  `sceHttpGetAllResponseHeaders` reported success **while writing nothing to its out-parameters**, so
  the caller parsed whatever was already in its buffer.

That second shape is exactly what crashes PGA TOUR 2K25 through the **v2** library (#2894, separate
PR). No surveyed title is currently *held* here — this is the latent version of the same defect,
across most of the corpus: six of eight titles in the never-booted survey already call
`sceHttpInit`/`sceHttpCreateTemplate`.

## Behavioural contract

The rule `src/hle/net/AGENTS.md` states, in both directions:

* **Anything answerable from local state is implemented, not stubbed.** Creating a connection or a
  request really does succeed offline — it allocates a local object and resolves nothing — so it
  returns a real id from a real table. The header and content-length setters record what they are
  given and *then* answer `SCE_OK`. `sceHttpGetLastErrno` reads back state prosper itself recorded,
  so it is answered rather than refused.
* **Anything that needs a network answer FAILS, and leaves every out-parameter untouched.**

**Nothing is manufactured** — no synthesised status code, no fabricated header block, no invented
body, no "pretend it worked" path. The guest's own error handling runs instead.

## Approach and invariants

* **`sceHttpSendRequest` answers a libSceNet `ENETUNREACH` (`0x80410133`).** The libSceHttp family
  does not wrap a connect failure in an HTTP error: it propagates the raw libSceNet error, encoded
  `0x80410100 | <FreeBSD errno>`. It agrees with what prosper already tells the same guest through
  NetCtl (`SCE_NET_CTL_STATE_DISCONNECTED` / `NOT_CONNECTED`), so the two answers stop contradicting
  each other. **Confidence is marked honestly and it is not uniform:** `HIGH` that a non-zero error
  is required, `MED` on the exact errno, and **`LOW` that v1 propagates the net error the same way
  v2 does** — that propagation was read off `libSceHttp2`'s connect site on #2894 and is *carried
  across* here, not read off `libSceHttp` itself. The value is produced in exactly one place so it
  is cheap to revise.
* **Every response getter reports that same failure and writes nothing** —
  `GetAllResponseHeaders`, `RequestGetAllHeaders`, `GetStatusCode`, `GetResponseContentLength`,
  `ReadData`, `WaitRequest`. A getter reached before any send answers "out of order"
  (`0x80431001`) rather than the send error.
* **The two per-class tables become one unified id space** (contexts, templates, connections,
  requests). With a table per class a template id and a request id collide numerically, so a
  validator handed an id cannot tell a live object of the *wrong* class from a live object of the
  right one — and this library's setters are documented to take a template, connection or request id
  interchangeably. The pre-existing guest-visible contract is unchanged (positive ids,
  `0x80431100` for an id nobody allocated, `Term` reclaiming what its context owned) and all of
  #3295's arms still pass untouched.
* **Teardown is a tree, one level deeper than before.** `sceHttpTerm` and `sceHttpDeleteTemplate`
  now reclaim connections and requests as well as templates, and `sceHttpDeleteConnection` takes its
  outstanding requests. Without that a title that creates and tears down in a loop exhausts a table
  that is really empty.
* **Where the class of a parent or an id is not established, validation checks only that the id was
  ALLOCATED**, not what class it is. Policing the class on evidence this thin would risk *rejecting*
  a call hardware accepts — the under-report half of the same defect — while the check that actually
  matters (id 0 and every never-created id are refused) is kept. `CONFIDENCE: HIGH` on that check,
  `LOW` on any per-class restriction, which is why none is imposed.
* **Behaviour change worth naming:** an exhausted table used to answer `-1` from `sceHttpInit` /
  `sceHttpCreateTemplate`. It now answers the library's own `SCE_HTTP_ERROR_OUT_OF_MEMORY`
  (`0x80431022`), sign-extended. Both are negative; this one is a real SDK constant.

* **Affected:** any title that reaches `libSceHttp`'s request path. It gets real ids and an honest
  failure instead of `SCE_OK` everywhere.
* **Deliberately unaffected:** the URI helpers, `sceHttpInit`, `sceHttpCreateTemplate`,
  `sceHttpDeleteTemplate` and `sceHttpTerm` keep their observable contracts — the id space beneath
  them changed, what the guest sees did not. `libSceHttp2` is a **different facility**
  (`0x817b____` vs `0x8043____`) and is #2894's separate PR. No renderer, loader or GPU path is
  touched.
* **Deliberately still unregistered, and this is the scope line:** the remaining local-only helpers
  — `sceHttpUriEscape` / `Unescape` / `Merge` / `Copy` / `SweepPath`, `sceHttpAddQuery`,
  `sceHttpParseResponseHeader` / `ParseStatusLine`, the cookie/auth-cache/proxy/epoll surfaces and
  the `sceHttpDbg*` family. Each is a real local computation with its own contract, and #3295
  derived `sceHttpUriBuild`'s contract from the shipped `libSceHttp.sprx` — **which is not on this
  machine**. Implementing them from guesswork would be worse than leaving them, so they are named
  here rather than quietly skipped, and I have commented on #2930 with the list.

## Risks

* The `LOW`-confidence carry-across above: if v1 does *not* propagate the raw net error, the
  constant is wrong. It would be wrong rather than dangerous — the one guest whose classifier has
  been traced buckets any non-zero as a failure — and it is one line.
* **Registering a NID can be worse than leaving it unregistered** if the argument shape is wrong.
  The mitigation is structural: every call that needs a network answer takes the "fail and touch
  nothing" path, which requires no knowledge of the out-parameter layout at all. The only new call
  that *writes* guest memory is `sceHttpGetLastErrno`, writing a single `int32`.
* A title that previously stumbled forward on false success may now take a different branch. That is
  the intended change and it is the branch the guest is written for, but it is a behaviour change for
  every `libSceHttp` request-path caller.
* `src/hle/net/AGENTS.md` is edited by both this PR and #2894's; they will conflict textually if both
  merge. Two lines, and the conflict is visible rather than silent.

## Verification

Platform: **Linux (WSL Ubuntu-24.04, GCC 13.3, Release)**. From the worktree root:

```
cmake --build prosper/build-linux -j12 -- -k
ctest --test-dir prosper/build-linux --no-tests=error -j8
```

* `test_http` (`ctest -R http_hle`): **41 → 74 executed assertions (+33)**, from 74 `CHECK` sites,
  every site executed exactly once so the source count and the executed count agree. **rc=0.** The
  neighbours it could disturb also pass: `test_hle_no_shadow` rc=0 (no NID registered twice),
  `test_service_getters` rc=0.
* `ctest --no-tests=error`: **458 tests discovered**, 346 passed, 112 failed, **rc=8** — and that rc
  is environmental on this box, not this branch. An **`origin/main` baseline built and run in a
  second worktree on the same machine with the same configure line** gives 458 discovered / **113**
  failed / rc=8, and the two failing SETS are identical except for one name that goes the *other*
  way: `file_binary_roundtrip` fails on the baseline run and passes here (re-run three times in the
  baseline tree it is rc=0 every time — flaky against the Windows filesystem). **No test fails on
  this branch that does not already fail on `origin/main`.**
  The 112 shared failures are the Vulkan-execution / live-renderer / GPU-storage cases: this box has
  Vulkan **1.3.275** headers and `frontends/shared/device/vulkan_runtime.hpp` needs
  `VK_API_VERSION_1_4`, so those targets do not compile here and ctest reports them `Not Run`.

### Every regression arm was shown to fail without the fix

Seven mutations, each rebuilt and re-run. Green baseline, red for each mutation, green again after
restore.

| mutation | `test_http` |
| --- | --- |
| baseline (unmutated) | rc=0, 0 failing checks |
| request-path NIDs unregistered (**the pre-fix state**) | rc=1, 2 failing checks |
| `sceHttpSendRequest` returns `SCE_OK` | rc=1, 4 failing checks |
| response getters return `SCE_OK`, out-parameters untouched | rc=1, 5 failing checks |
| `sceHttpCreateConnection` hands out id 0 | rc=1, 9 failing checks |
| `sceHttpTerm` releases only its own slot (leaks the tree) | rc=1, 14 failing checks |
| the id validator accepts any id, including 0 | rc=1, 3 failing checks |
| `sceHttpGetLastErrno` always writes 0 | rc=1, 1 failing check |
| restored | rc=0, 0 failing checks |

Two notes on how those arms are built, because a passing assertion is not automatically a binding
one:

* Every id arm demands a **positive** id for a good call and a **negative** one for a bad call, so
  the old `return 0` satisfies neither; every response arm pre-fills its out-parameters with a
  sentinel and requires the sentinel to **survive** next to a non-zero return.
* The leak control fills the table with a **connection-and-request tree** under one context before
  terminating it, then requires a fresh context to reach the same depth — so the leak it guards is
  structurally expressible. The vacuous version of that arm is what #3295's review had to fix one
  level up.
* The "unregistered" arm reddens only 2 checks because the block guards itself against dereferencing
  a null handler, which would otherwise segfault before printing anything. That guard is deliberate
  and commented: the arm that *reports* the regression is the registration `CHECK`, not the `if`.

## What I could NOT verify

* **No boot was run and no title was observed.** No surveyed title is blocked here, so there is no
  progression claim attached to this PR at all — it removes a latent false success.
* **The shipped `libSceHttp.sprx` is not on this machine**, so I could not re-derive argument shapes
  or the facility from the module; the contracts come from the derivation recorded on #2894, from
  `src/hle/net/AGENTS.md` on `main`, and from the v1 family's published shape. That is also why the
  remaining local-only helpers are left out of scope rather than guessed at.
* **No network service exists to check the failure path against**, by design.
* No Vulkan-execution or snapshot test ran (they cannot build on this box, above), and none is
  reachable from an HLE registration change.
* Windows and macOS were not built locally. The file is plain C++ with no platform surface.


mattias800 (commented):
APPROVED — independent review on exact head, I am not the author. Reviewed after its v2 sibling
(#3544), and the consistency between them is part of why I am approving.

## Same contract, same direction, verified independently

The v1 library had exactly one registered function (`sceHttpUriParse`), so `sceHttpInit` and
`sceHttpCreateTemplate` fell to `return 0` — and for those contracts `0` is a **library context id**
and a **template id**, not an error. Six of eight surveyed titles call them. The file's own opening
comment already stated the principle (*"returning success without filling this structure makes
callers dereference stale pointer fields"*) and applied it only to the URI parser; this extends the
reasoning rather than contradicting it.

Checked in the code:

- `sceHttpSendRequest` sets `send_error = kNetErrorNetUnreach` and returns it. Not `SCE_OK`.
- Every response getter routes through `http_response_unavailable(*req)`, which reports the recorded
  failure — or `kErrorBeforeInit` if no send was attempted — and leaves out-parameters untouched.
- `g_http_objects[kMaxHttpObjects + 1]` is **1-based, slot 0 never handed out**. The original defect
  (an id of `0` read as valid) is now structurally unrepresentable, not merely avoided.

That the v1 and v2 paths answer the same question the same way matters here beyond tidiness. The
charter's rule from #1873 is that a divergence between two libraries exposing the same capability is
itself a defect, because the answer then depends on which library the guest happened to ask through —
and these two are one version apart with overlapping callers.

## Scope is right

`CreateConnection*`, all four `CreateRequest*`, delete/abort, header setters, `SendRequest`,
`ReadData`, `WaitRequest`, the response getters, and `GetLastErrno` — the one question answerable
from local state. Covering the whole request path rather than only the two entry points the survey
named is the correct granularity: a title that got past `CreateTemplate` would otherwise hit the same
false success one call later.

The two per-class tables becoming one unified id space, with `Term`/`DeleteTemplate`/
`DeleteConnection` reclaiming the whole tree, is a real lifetime fix and not incidental cleanup —
partial reclamation on a shared id space is how a stale id later resolves to a live object of the
wrong kind.

#3295's existing arms all pass unchanged, which is the check I would want given the id-space
restructure.

## Arms

Seven mutations, each rebuilt and re-run, green → red → green, covering the same class of hollow
fixes as the v2 PR (unregistered, send returns `SCE_OK`, getters succeed with untouched
out-parameters, create returns 0, `Term` leaks, validator accepts 0). `test_http` went 41 → 74
assertions, rc=0.

Linux: `ctest --no-tests=error` 458 discovered / 112 failed / rc=8, failing set identical to an
`origin/main` baseline built in a separate worktree — no failure added. The 112 are this box's
Vulkan-1.4 header gap.

## The caveat I want kept visible

**The v1 send constant is `CONFIDENCE: LOW`**, carried across from v2 rather than derived for this
library. The author labelled it rather than letting it inherit the v2 evidence silently — which is
the right call, because the two libraries are a version apart and "the sibling returns this" is not
the same as "this returns this". Worth an issue if a v1 caller ever turns out to classify the error
differently.

No title was booted against this. Nothing in the PR claims otherwise.

Issue #3550, opened 2026-09-11
decode_buffer_descriptor still fail-closes packed formats on a non-canonical DST_SEL, though the lowering can now express any selector


mattias800:
Follow-up from #2869.

## Current behaviour

`agc_shader_layout.cpp`'s `decode_buffer_descriptor` rejects a **packed** format
(`10_11_11_FLOAT`, `2_10_10_10_*`) whose `DST_SEL` word is not one of two exact shapes — `0xFAC`
(X/Y/Z/W) for 2_10_10_10, `0x3AC` (X/Y/Z/1) for 10_11_11 — by setting `format = Unknown`,
`num_components = 0` and `forbid_unknown_fallback = true`. Downstream that leaves the fetch
unresolved and the recompiler drops the draw.

The comment justifying it said the gate stood *"until ShaderResource carries arbitrary DST_SEL
through to SPIR-V"*. Since #2869 it does: the MUBUF format lowering indexes the packed field by the
**routed** source component, so an arbitrary permutation, a repeated component and the constant
selectors are all expressible for packed formats too.

## Why it was not widened in #2869

Widening it changes what live titles render — a descriptor that produced a dropped draw would start
producing a picture — and no title in the local corpus has been observed supplying a routed packed
descriptor, so there is nothing to validate the change against. A fail-closed gate is the safe state
in the absence of that evidence; removing it on the strength of "the lowering could handle it" is
the kind of unverified widening the charter's positive-control rule warns about.

## What would close this

1. A live capture of a title binding a packed V# with a non-canonical selector to a
   `buffer_load_format_*` consumer. The `[buf-op]` diagnostic line now carries `dstsel=` under
   `PROSPER_DBG`, and the packed rejection is visible as an unresolved fetch alongside it.
2. With that in hand, relax the gate to reject only what remains genuinely unsupported (the scaled
   `2_10_10_10` formats, raw 52/53) and add an execution arm over a routed packed descriptor.

Note MTBUF is unaffected in either direction: its consumers skip `forbid_unknown_fallback`, and per
RDNA2 ISA Table 31 a `tbuffer_*` op uses identity routing regardless of the descriptor.

Refs #2869.

Issue #3549, opened 2026-09-11
BUFFER_STORE_FORMAT_* through a routed V# rejects instead of storing: the DST_SEL direction is not established


mattias800:
Follow-up from #2869, which implemented V# `DST_SEL` routing for **loads** only.

## What is in the tree now

`rdna2_emit_alu.cpp`'s MUBUF format lowering routes each returned channel through the descriptor's
`DST_SEL` (SQ_SEL 4/5/6/7 = stored R/G/B/A, 0 = constant 0, 1 = constant 1). A typed **store**
(`buffer_store_format_x/xy/xyz/xyzw`) through a descriptor whose selector is not identity is
**rejected** — `buf_op.how = "reject-dst-sel-store"` — rather than stored with the field ignored.

## Why it is a gap rather than a settled contract

RDNA2 ISA (document 70648) sec. 8.1.4 "Buffer Data", Table 31 lists
`BUFFER_STORE_FORMAT_*` with a `DST SEL` of **resource**, exactly like the load. So the hardware
does apply the descriptor on the way out; prosper simply does not know *how*:

* For a load, `DST_SEL_X` names the stored component that lands in returned channel X. For a store
  the mapping runs the other way, and the document does not say whether the field is applied as the
  same permutation or as its inverse. Those differ for any selector that is not an involution.
* A constant selector (`SQ_SEL_0` / `SQ_SEL_1`) has no stated meaning on a store at all — there is
  no "returned channel" to fill with a constant.

Ignoring the field would write components in the wrong order into guest memory, which nothing
downstream can detect; that is why the lowering fails visibly instead. `CONFIDENCE: LOW`.

## What would close this

Direct title evidence: a live capture of a guest issuing `buffer_store_format_*` through a
non-identity descriptor, with the resulting bytes in guest memory readable so the direction can be
read off rather than assumed. Absent that, hardware documentation that states the store direction.

Until then the reject is the correct behaviour. If a title is observed hitting
`reject-dst-sel-store` (visible under `PROSPER_DBG` on the `[buf-op]` line, which now carries
`dstsel=`), that capture is the evidence this issue needs.

Refs #2869.

PR #2861, opened 2026-08-21
fix(recompiler): the PC-relative embedded-table fold refused a TYPED (MTBUF) consumer — Sonic Frontiers' three Cyber Space scene kernels

mattias800:
> **Body refreshed at head `4128043f`.** Every figure below is stated at that head. The earlier body
> quoted the first commit throughout and listed three test arms where there are now ten checks across five programs; the
> reviewer flagged that as P1 and this replaces it wholesale.

## What this fixes

`detect_pcrel_tables` (`rdna2_cfg_support.hpp`) folds a constant table the compiler placed inside the
shader blob and addressed with an `s_getpc_b64`-built V#. It recognised two consumers of that table —
an untyped `MUBUF` raw load and a scalar `SMEM` buffer load — and **not** a typed `MTBUF` one. When
the consumer is typed the table never folds, so `s_getpc_b64` has nothing to justify it
(`rdna2_emit_alu.cpp:1062` accepts the pair only when a table from this shader was proven) and the
whole program rejects `mode=unresolved-operand` — **at the getpc, not at the load that carries the
information**.

*Sonic Frontiers* (`PPSA03831`) hits this in all three of its Cyber Space scene-width compute
kernels, twice each. From the live `PROSPER_SHADER_DUMP` bytes through `tools/shader_inspect`:

```text
pc=0514 fmt=SOP1  op=0x003 words=be8303ff,10005004 dst=sgpr:3    ; s_mov_b32 s3, V# config
pc=0516 fmt=SOPK  op=0x000 words=b0020080 dst=sgpr:2 simm=128    ; s_movk_i32 s2, num_records
pc=0517 fmt=SOP1  op=0x01f words=be801f00 dst=sgpr:0             ; s_getpc_b64 s[0:1]   <-- rejected
pc=0518 fmt=SOP2  op=0x000 words=800000ff,00000148 dst=sgpr:0    ; s_add_u32 s0, 0x148, s0
pc=0520 fmt=SOP2  op=0x004 words=82010180 dst=sgpr:1             ; s_addc_u32 s1, 0, s1
pc=0528 fmt=MTBUF op=0x000 words=e8b01000,80000a0a               ; tbuffer_load_format_x v10, v10, s[0:3], 0 offen
```

That is the MUBUF idiom byte-for-byte with a typed load in place of `buffer_load_dword`.

## Behavioural contract — three obligations, not one

[...]

PR #3552, opened 2026-09-11
Honour the V# DST_SEL channel routing on MUBUF format fetches

### LINKED ISSUE #2869: V# DST_SEL is ignored by every format-fetch lowering, and two comments in rdna2_emit_alu.cpp state the opposite of shader_resources.cpp
## Summary

`res->swizzle` — the V#/T# `DST_SEL_X/Y/Z/W` channel routing — is read in exactly **one** place in the
whole recompiler: `rdna2_emit_alu.cpp:5809`, the `is_zero_record_raw_buffer` branch. No ordinary
`buffer_load_format_*` or `tbuffer_load_format_*` lowering consults it. If a title binds a typed
buffer whose descriptor routes channels non-identically, prosper returns the stored order and the
guest gets the wrong components — silently, with no reject and no diagnostic.

Found during review of #2861 (finding N1), which added an identity-DST_SEL obligation to the
*folded* PC-relative embedded-table path. That fold is now safe; the ordinary path is not.

## The repository currently states this two different ways, and they should be reconciled

* `shader_resources.cpp:210` calls DST_SEL **"a FORMAT-fetch control: it names which component each
  returned channel takes"**, and deliberately keeps requiring the identity mapping for the
  user-SGPR selector mode *"whose consumers may be typed"*. That treats DST_SEL as binding on typed
  fetches.
* `rdna2_emit_alu.cpp:6359-6360` and `:6400-6401` say **"MTBUF forces identity selection from its
  instruction format (X000/XY00/XYZ0/XYZW), so every absent component is zero"**, which reads as a
  claim that a typed fetch ignores the descriptor's routing.

Both comments sit inside the format **default-fill** discussion (#368), so the second pair is
plausibly about *which components exist* rather than *how the descriptor routes them* — DST_SEL and
DATA_FORMAT are separate fields, and MTBUF overriding the format field says nothing about the
selector field. But as written they are read as contradicting each other, and #2861's first commit
looked complete partly because of them.

## What to settle, and how

1. Establish from the RDNA2 ISA (document 70648, buffer resource + buffer-instruction sections)
   whether `tbuffer_load_format_*` applies the V#'s `DST_SEL`, and whether `buffer_load_format_*`
   does. Do not settle it from prosper's own comments — they are what is in question.
2. Confirm against a live capture from a title that binds a non-identity selector to a format
   consumer, if one exists in the corpus. GTA V's three-component tables (`0x3ac`, X/Y/Z/1) are the
   obvious place to look (#2481).
3. Then either honour the selector in the format lowerings, or reject a non-identity selector
   fail-visibly, and correct whichever comments are wrong.

## Note on #2861's guard

#2861 requires identity routing before folding a typed embedded table. That is **fail-closed under
either reading**: if DST_SEL does bind, the guard is necessary; if MTBUF really does force identity,
the guard is merely narrower than it needs to be and costs at most a fold that would otherwise have
been legal. It does not affect the live Sonic Frontiers case, whose descriptor routes X identically.
`CONFIDENCE: MED` on the routing semantics is exactly why the guard is conservative.

Refs #2859, #2861, #368, #2481.

mattias800:
Fixes #2869.

## Failure scenario

The V# carries `DST_SEL_X/Y/Z/W` — three bits per returned channel (0 = constant 0, 1 = constant 1,
4/5/6/7 = stored R/G/B/A). Nothing in prosper decoded it for a **buffer** descriptor:
`DecodedBufferDescriptor` had no such field, so every buffer-class `ShaderResource` kept
`swizzle`'s identity default and the recompiler's format lowerings returned the **stored** order
whatever the descriptor asked for. Silent by construction — no reject, no diagnostic, just a
plausible wrong picture. (`res->swizzle` was read in exactly one place, the zero-record branch,
where it could never be anything but the default.)

## What the ISA actually says — this is the part #2869 asked to settle

RDNA2 ISA (document 70648) §8.1.4 "Buffer Data": *"Dst_sel comes from the resource, but is ignored
for many operations."* Its **Table 31** gives the whole partition:

| Instruction | DST SEL |
|---|---|
| `BUFFER_LOAD_FORMAT_*` / `BUFFER_STORE_FORMAT_*` | **resource** |
| `TBUFFER_LOAD_FORMAT_*` / `TBUFFER_STORE_FORMAT_*` | identity |
| `BUFFER_LOAD_<type>` / `BUFFER_STORE_<type>` / `BUFFER_ATOMIC_*` | identity |

…where the same table defines identity as *"X000, XY00, XYZ0, or XYZW"* by the format's component
count — a substitution that zeroes absent channels, not a pass-through.

**So the two comments the issue was filed over were both correct, of different instruction
families.** `shader_resources.cpp:210` ("a FORMAT-fetch control") is right about MUBUF;
`rdna2_emit_alu.cpp`'s "MTBUF forces identity selection" is right about MTBUF. Neither reader could
reconcile them because neither said which family it meant. Both now do.

Corroborated behaviourally by Mesa, which carries a BGRA vertex swizzle in the descriptor's
`DST_SEL` on its `buffer_load_format_xyzw` path but **hand-reverses the per-channel byte offset in
the shader** on its `tbuffer_load_format_*` path (`aco_select_vs_prolog.cpp`, `radv_cmd_buffer.c`
`post_shuffle`) — a driver would not pay for that if MTBUF honoured the field. Secondary evidence,
cited as corroboration only.

## Approach and invariants

* `DecodedBufferDescriptor` gains `dst_sel[4]`; `decode_buffer_descriptor` fills it, and
  `apply_buffer_descriptor_swizzle` carries it to `ShaderResource::swizzle` at all seven V#
  construction sites.
* **A descriptor declaring no FORMAT keeps identity.** Prosper reaches those zero words only through
  paths that *fabricate* a V# (raw-pointer scalar consumers #2412, the legacy `Unknown → Float32`
  vertex fallback). Reading a zero WORD3 as "every channel is constant 0" would blacken fetches that
  render today. Deliberate conservatism about prosper's fallbacks, not a claim about hardware.
* The MUBUF format **load** lowering routes each returned channel to its stored source (`sk`), with
  `SQ_SEL_0`/`SQ_SEL_1` as constants. **Identity routing is bit-identical to the previous lowering**
  — `sk == k` and every branch is unchanged.
* Reserved selectors 2/3 reject. A routed format **store** rejects: Table 31 does give
  `BUFFER_STORE_FORMAT_*` a resource DST SEL, but not the *direction* of the mapping, and a constant
  selector has no stated meaning on the way out. Tracked as #3549, `CONFIDENCE: LOW`.
* **The store guard is narrowed to components that physically exist.** A three-component format's
  canonical selector is `X/Y/Z/1` (0x3AC) — the routine encoding, and the shape GTA V's tables carry
  (#2481). Gating the store on "any non-identity slot" would have rejected every
  `buffer_store_format_xyzw` through an ordinary alpha-less descriptor: a draw-dropping regression
  wearing a fail-visible label. It has its own test arm.

## Corpus evidence — non-identity selectors are real

Census on *The Messenger* `PPSA24651` (`tools/screenshot`, default route, `PROSPER_DBG=1`), reading
the `dstsel=` field this PR adds to the `[buf-op]` line so the census carries its own denominator:

```
[buf-op] program=0x2011600100 pc=76  MUBUF op=0x3 n=4 store=0 dstsel=0x3ac resolved
[buf-op] program=0x2011600100 pc=124 MUBUF op=0x1 n=2 store=0 dstsel=0xfac resolved
[buf-op] program=0x2011600100 pc=150 MUBUF op=0x3 n=4 store=0 dstsel=0xfac resolved
[buf-op] program=0x201164c800 pc=12  MUBUF op=0x0 n=1 store=0 dstsel=0x004 resolved
[buf-op] program=0x201164c800 pc=15  MUBUF op=0x4 n=1 store=1 dstsel=0x004 resolved
```

`pc=76` is a four-component fetch through `0x3AC` — `X/Y/Z/SQ_SEL_1`, i.e. a genuinely non-identity
selector reaching a live MUBUF format load on a rung-6 title. Honest reading: **no *permuted*
selector was observed** in this sample; what the corpus supplies is the constant-one W, which the
old lowering happened to match via its absent-W default. So this PR's permutation handling is not
yet exercised by a title I have measured — it is implemented because the descriptor can express it
and the consequence of ignoring it is invisible. The invariance arms exist precisely because the
observed case is the one that must **not** change.

Affected: MUBUF `buffer_load_format_*` through a routed descriptor; MUBUF `buffer_store_format_*`
routed among existing components (now rejects). Deliberately unaffected: MTBUF (asserted by test),
raw loads/stores/atomics, image/T# swizzle (already handled), zero-record descriptors (WORD3 = 0 →
identity), `decode_buffer_descriptor`'s packed-format gate (kept fail-closed; #3550),
#2861's fold guard.

## Risks

The behaviour change is confined to descriptors whose selector differs from identity **among
components the format has**, which no title in the local corpus was observed supplying. The
`0x3AC` case that *is* supplied is pinned as unchanged by two arms. The store reject can drop a draw
if some title routes a typed store; that is the visible failure #3549 wants evidence for, and it is
preferred over a silent wrong write.

## Build and test

Linux (WSL Ubuntu-24.04, `prosper/build-linux`) — authoritative:

* `cmake --build prosper/build-linux --target prosper_core` → exit 0
* `ctest --no-tests=error -R "rdna2_to_spirv_exec|build_shader_resources"` → **2/2 passed, exit 0**
* `./test_rdna2_to_spirv` → **PASS: 1082 checks**, all nine new arms green
* `./test_build_shader_resources` → PASS, five new DST_SEL decode checks green

A *full* Linux ctest is not available on this machine: its Vulkan headers are 1.3 and the
live-renderer chain needs 1.4, so `prosper_live_renderer`/`gpu_replay`/`prosper-app` do not build
there. That is environmental, not this change.

Windows/MinGW, reported as observed and not as a gate: full `ctest --no-tests=error` →
**333 passed / 334, 1 failed**, the failure being `session_start`, pre-existing (#3540) and
unrelated. That run predates the store-narrowing commit.

**Without-fix arm (run, not asserted):** source files reverted to the parent commit with the tests
kept, rebuilt and re-run on Linux → `== FAIL: 5 of 1079 checks ==`, exactly the five MUBUF arms:
the B,G,R,SQ_SEL_1 route, the SQ_SEL_0 constant, the repeated component, the reserved-selector
reject and the routed-store reject. The MTBUF arm passes in both states **by design** — it asserts
the routing is ignored, so it is an over-reach guard rather than a regression arm, and this is
stated rather than counted as coverage.

## Record

`docs/RECOMPILER_REMAINING.md` § Ruled out gains the falsification row. Follow-ups filed: #3549
(store direction), #3550 (packed-format selector gate).


mattias800 (commented):

## The contradiction is settled, and the resolution is better than either comment

The issue framed this as "two comments state the opposite; one is wrong." The answer is that
**neither was wrong** — they were describing different instruction families and neither said which.
RDNA2 ISA doc 70648 §8.1.4 Table 31: `BUFFER_LOAD/STORE_FORMAT_*` take DST_SEL from the **resource**;
`TBUFFER_*` and all raw loads/stores/atomics use **identity**. So `shader_resources.cpp` was right
about MUBUF and `rdna2_emit_alu.cpp`'s "MTBUF forces identity" was right about MTBUF. Both now say
which family they mean.

That is the outcome I most wanted and least expected. "One of these is a bug" is the framing that
produces a confident wrong edit, and the author declined it on evidence rather than splitting the
difference. The independent corroboration is the right kind too: Mesa hand-reverses channel offsets
in the shader on its MTBUF path while relying on descriptor DST_SEL on its MUBUF path — a second
implementation behaving as the table says, reached by a different route.

## The root cause was deeper than the issue stated

`DecodedBufferDescriptor` carried **no `dst_sel` field at all**, so buffer `ShaderResource::swizzle`
was structurally always identity — the single read site could never have observed anything else.
The issue described a lowering that ignored a value; the truth was that the value never arrived. A
fix written to the issue's description would have been a no-op that looked correct.

## The over-rejection the author caught in their own change

This is the part that earns the approval. The obvious gate — reject a format store whose descriptor
carries **any** non-identity selector — would have rejected **every** `buffer_store_format_xyzw`
through an ordinary alpha-less descriptor, because `0x3AC` (X/Y/Z/1) is the routine encoding every
driver emits for a three-component format, and it is the shape GTA V's tables carry. That is not a
corner case; it is the common case, and it would have dropped real draws under a plausible-sounding
safety rule.

Narrowed to **routing that touches a component which physically exists**, with two invariance arms
pinning it. I checked the logic holds in both directions: X/Y/Z/1 on a three-component format
selects a constant for a slot that does not exist → admitted; a genuine permutation on a
four-component format touches existing components → rejected. Correct.

Finding this required running a corpus census against one's own patch rather than against the
original bug, which is a step most changes skip.

## Fail-closed on the direction that has no stated meaning

Table 31 says DST_SEL applies to `BUFFER_STORE_FORMAT_*`; it does not say **how** — i.e. the
direction the selector routes on the way out. The author rejects rather than guessing, marks it
`CONFIDENCE: LOW`, and filed **#3549**. That is exactly the charter's fail-visible backstop for a
genuinely unknown encoding: a rejected store is loud, a wrongly-routed one writes plausible garbage
to the wrong channels and is invisible. The packed-format selector gate is kept fail-closed for the
same reason (**#3550**).

## Corpus evidence, reported honestly

*The Messenger* does supply a non-identity selector to a live MUBUF format load — `dstsel=0x3ac`
(X/Y/Z/constant-1) at `pc=76`. But **no permuted selector was observed anywhere in the corpus**, and
the author says so: permutation handling exists because the descriptor can express it and ignoring
it would be invisible, not because a title was seen doing it. The corpus case is the constant-one W,
which the old lowering happened to match through its absent-W default.

That distinction — "implemented because expressible" versus "implemented because observed" — is
precisely what a `CONFIDENCE` tag is for, and stating it prevents the next reader from citing this
PR as evidence that titles permute channels.

## Arms

Without-fix: sources reverted to the parent with tests kept, rebuilt, re-run → `FAIL: 5 of 1079`,
exactly the five MUBUF arms. The **MTBUF arm passes in both states by design** and the PR says so
rather than counting it as coverage — it is the over-reach guard, not a regression arm. Same
discipline as #3546, and it is the reason I trust the five.

Linux (authoritative here): `ctest --no-tests=error -R "rdna2_to_spirv_exec|build_shader_resources"`
→ 2/2 passed, exit 0; `test_rdna2_to_spirv` PASS 1082 checks. A full Linux ctest is impossible on
this box (1.3 headers vs the 1.4 the live-renderer chain needs). Per the owner's standing
instruction, the non-Linux results are not treated as blockers.

Falsification recorded in `docs/RECOMPILER_REMAINING.md` § Ruled out, as the charter requires.

PR #3553, opened 2026-09-11
hle(savedata): persist the param block instead of answering SCE_OK and storing nothing

### LINKED ISSUE #2786: libSceSaveData: sceSaveDataSetParam / sceSaveDataGetParam are unregistered, so the save param block silently succeeds and reads back zero
## Summary

`sceSaveDataSetParam` (NID `85zul--eGXs`, `libSceSaveData` / `libSceSaveData_native`) is not
registered anywhere in `prosper/src`, so it reaches `prosper_on_unimpl`'s `return 0` — and for this
contract **0 is `SCE_OK`**. That is the FALSE SUCCESS class tracked by #2081: the guest is told its
save-data metadata was written when nothing was written.

Verified unregistered on `cf0594f5`:

```
$ grep -rn 'sceSaveDataSetParam\|85zul' prosper/src
$ echo $?
1
```

`sceSaveDataGetParam` is equally absent, so a title that writes the param block and reads it back
gets a zeroed answer rather than what it set — an inconsistency the guest cannot detect.

## How it was found

Observed on *Sonic Frontiers* (`PPSA03831`, #1891) while deriving a route to gameplay. It is the
**only** distinct unimplemented NID this title reaches on a route that goes all the way into
`GameModeStage`; measured identically on three separate arms:

```
[prosper] unimplemented: libSceSaveData_native::85zul--eGXs  -> returning 0
```

Called once per boot in every arm. Frontiers writes `/savedata0/{option,arcade,challenge}data`, so
the save area itself works — it is the param/metadata block that silently does not.

Name resolved from the PS5 3.20 library dump:

```
libSceSaveData.c:772:  if(sprx_dlsym(__handle, "85zul--eGXs", &__ptr_sceSaveDataSetParam)) return;
libSceSaveData.native.c:277: (same)
```

## Why it matters even though no title is currently blocked on it

The param block carries what the console's save-data browser shows: title, sub-title, detail and a
user param. On PS5 this is what makes a save identifiable. Answering `SCE_OK` without storing it is
the under-reporting failure direction — it looks safe, so nobody files it, and the next title that
reads its own metadata back gets a wrong answer that the *guest* then behaves correctly on.

There is no evidence it blocks Frontiers: the title is called once at boot and progresses to
gameplay regardless. Filed as a real gap rather than a blocker.

## Suggested fix

Register both halves against the save-data area prosper already maintains under `PROSPER_SAVE0`
(namespaced `<root>/<TITLE_ID>/<guest key>` since #2734): persist the param block alongside the save
directory it names, and answer `sceSaveDataGetParam` from what was stored. Derive, do not
hardcode — and the regression that proves derivation is a round-trip arm plus a **mutation arm**
where a different param block yields a different read-back, since a stub that returns a fixed block
would otherwise pass the round-trip.

`CONFIDENCE: HIGH` that it is unregistered and that 0 is this contract's success value.
`CONFIDENCE: LOW` on the exact argument layout — that needs live capture or guest disassembly, not
the 3.20 dump, which carries names and NIDs only.

Refs #2081, #1891.

mattias800:
Fixes #2786.

## The failure scenario

`sceSaveDataSetParam` (NID `85zul--eGXs`) and `sceSaveDataGetParam` were registered nowhere in
`prosper/src`, so both reached `prosper_on_unimpl`'s `return 0` — and for these contracts **0 is
`SCE_OK`**. The guest was told its save metadata had been written when nothing was written, and a
title that wrote the block and read it back got a zeroed answer it had no way to detect.

That is the FALSE SUCCESS class (#2081), in the under-reporting direction the charter calls out as
the one that "fails in the direction that looks safe, so nobody files it". The param block carries
title, sub-title, detail and a user param — what the console's save browser shows — so answering
`SCE_OK` without storing it silently removes information the user has.

Observed on *Sonic Frontiers* (`PPSA03831`) as the **only** distinct unimplemented NID reached on a
route that goes all the way into `GameModeStage`, measured identically on three arms.

## Behavioural contract

- The param block is **persisted**, as a real `sce_sys/param.sfo` inside the mounted save directory
  — the path `savedata0_dir_mtime()` already stats and could never previously find.
- `GetParam` reads back what `SetParam` wrote.
- Every case prosper cannot honestly serve is a **real refusal**, not `SCE_OK`: no mount, an
  unserved mount point, an unknown param type, a null or undersized buffer.

The distinction the charter draws applies cleanly here: this is derivable from bytes in the dump —
the save area is local, and the block is the guest's own data round-tripped — so it is reimplemented
faithfully rather than answered with a constant.

## Approach

| piece | why |
| --- | --- |
| new `src/hle/fs/save_param.{hpp,cpp}` | an SFO codec, separate from the HLE entry points, so the format is testable without a mounted save |
| `savedata0_mounted_dir()` accessor | the handlers need the *mount*, not the save root; without it the file lands where nothing reads it |
| refusals replacing `return 0` | listed above — each is a case where a wrong answer would be invisible to the guest |

## Verification

Both platforms, stated separately because they are not equivalent here.

```
Linux  (WSL Ubuntu-24.04, Release)
  cmake --build prosper/build-linux --target test_savedata_param   -> exit 0
  ctest --test-dir prosper/build-linux --no-tests=error -R ^savedata_param$
      -> 1/1 Passed, exit 0
Windows (MinGW/UCRT, Ninja)
  test_savedata_param -> 57 assertions, PASS
```

The Linux arm was run after merging current `main` into the branch (7 commits behind at the time),
rebuilt and re-run, so the result is against the merge product rather than a stale base.

Per the owner's standing instruction, only Linux CI is treated as blocking.

### Without-fix arms — both run, both red

- NIDs left unregistered → the test fails immediately.
- Handlers reduced to the pre-fix `return 0` → **32 of 57 assertions fail**, exit 1.

The second is the one that matters: it is the exact pre-fix behaviour, so the arm distinguishes
"stores and returns the data" from "returns `SCE_OK` and stores nothing" rather than merely
detecting that the functions exist.

A follow-up commit (`4625231d`) bounds every read of a possibly-untouched buffer in the test itself
— a test that reads past what the handler wrote would have been measuring uninitialised memory,
which is the failure mode this PR exists to remove, one level up.

- **Affected:** two previously-unregistered NIDs, the save mount accessor, and the new SFO codec.
- **Unaffected:** the save area itself. *Sonic Frontiers* already wrote
  `/savedata0/{option,arcade,challenge}data` correctly; it was only the metadata block that silently
  did not persist. No existing save path changes behaviour.

## Risks and what is not verified

- **No title was booted against this.** There is no evidence it unblocks anything, and none is
  claimed — #2786 says explicitly that Frontiers progresses to gameplay regardless. This closes a
  latent false success, not a wall.
- The SFO layout is implemented from the format as used by the local save area rather than from a
  published Sony specification; uncertain points carry `CONFIDENCE` markers in the source.
- Authored by a subagent; the Linux build, the merge with `main`, and this PR body are mine.


mattias800 (commented):
APPROVED.

**Disclosure first, since it bounds what this review is worth:** I did not write this code — a
subagent did — but I am not a neutral party either. I merged current `main` into the branch, ran the
Linux build the author could not, and wrote the PR body. So treat this as a review of the
*implementation*, which I had no hand in, and not of the integration, which I did.

## The property that decides this change

The defect is FALSE SUCCESS in its quietest direction: `return 0` where 0 is `SCE_OK`, so the guest
is told its metadata was stored when nothing was. The fix is only worth anything if every path that
cannot honestly serve the call now returns a real error rather than 0. Checked in the source:

```cpp
if (!savedata_mount_point_ok(a0) || !a2 || !need || a3 < need) return SAVE_DATA_ERR_PARAMETER;
if (dir.empty())                                               return SAVE_DATA_ERR_NOT_MOUNTED;
...
if (write failed)                                              return SAVE_DATA_ERR_INTERNAL;
return 0;   // only after the block is actually on disk
```

Unserved mount point, null out-pointer, undersized buffer, no mount, failed write — each is a
distinct named error, and `0` is reachable only after the write succeeds. That is the contract the
issue asked for.

## It is derivable, so implementing it is correct rather than optional

The charter's test is whether the answer comes from bytes in the dump. It does: the param block is
the guest's own data, the save area is local, and `GetParam` returns what `SetParam` wrote. This is
the *positive* half of the entitlement rule — under-reporting local content is as much a defect as
over-reporting it would be a circumvention, and it fails in the direction that looks safe, which is
why it sat unfiled.

Persisting as a real `sce_sys/param.sfo` **inside the mounted save directory** is the detail I would
have got wrong: that is the path `savedata0_dir_mtime()` already stats and could never previously
find, so the file now lands where the rest of the code already looks rather than somewhere new.

## Arms

Both directions, and the second is the one that matters:

- NIDs left unregistered → fails immediately.
- **Handlers reduced to the pre-fix `return 0`** → 32 of 57 assertions fail. That is the exact
  pre-fix behaviour, so the arm distinguishes "stores and returns the data" from "returns SCE_OK and
  stores nothing" rather than merely detecting that the functions exist.

Commit `4625231d` bounds every read of a possibly-untouched buffer in the test itself. Worth calling
out: a test that read past what the handler wrote would have been measuring uninitialised memory —
the same failure mode this PR removes, one level up in the apparatus.

## Verification

Windows: 57 assertions, PASS. Linux (run by me after merging `main`, rebuilt against the merge
product rather than a stale base): `ctest --no-tests=error -R ^savedata_param$` → 1/1 Passed,
exit 0. Merge gate GREEN.

## What is honestly not claimed

**No title was booted against this**, and the PR says so. #2786 states plainly that *Sonic Frontiers*
progresses to gameplay regardless — this closes a latent false success, not a wall, and there is no
rung movement to claim. The SFO layout is implemented from the format as the local save area uses it
rather than from a published specification, with `CONFIDENCE` markers on the uncertain points.

Given no title is blocked here, that disposition is right: the value is that the next title to read
its own metadata back gets the truth, and nobody has to re-derive why it was zero.

Issue #3324, opened 2026-09-04
A guest-thread _Exit bypasses every frontend finalize step, silently losing armed diagnostics


mattias800:
## Problem

A diagnostic armed in `prosper-app`'s loop is lost with no message when the **guest** terminates the
process, because the guest does it from its own thread while main is still inside the app loop.

Two sites call `_Exit` this way:

- `src/hle/.../hle_kernel_time.cpp:878` — `k_exit`, the guest's `_exit`
- `src/hle/.../hle_kernel_time.cpp:888` — `k_debug_raise_release`, UE4's shipping `check()`

`_Exit` is `exit_group`, so it takes down every thread immediately. **An `atexit` hook would not help
either** — that is the point worth recording, because it is the obvious fix and it does not work.

Found in the #3322 review of the RenderDoc capture trigger, which closes an open capture after its
app loop (`main.cpp`). That close covers every path *out of the loop* — verified by enumerating the
loop body for `return` / `exit` / `_Exit` / `abort` / `terminate` and finding none — but it cannot
cover a `_Exit` issued from another thread while the loop is still running.

## Why it is worth fixing generally

This is not specific to RenderDoc. Any frontend-owned artifact with a finalize step has the same hole:
the F8 `.prperf`, the F9 bundle, anything holding a partially written file. The failure is silent by
construction — the process is simply gone.

Both `_Exit` sites **already** run a "flush diagnostics before a deliberate `_Exit`" step, so the
pattern exists; what is missing is a way for a frontend to register into it. A small registered
finalizer list, invoked immediately before those two `_Exit` calls, would close the seam for every
current and future consumer.

## Why it was not fixed in #3322

It is a cross-layer seam (`hle/` calling back into a frontend-registered list) that does not exist
yet, and inventing it inside a diagnostics PR would be the wrong place to design it. The limit is
documented at the call site in `main.cpp` instead, so the next reader does not mistake the exit-path
close for total coverage.

Refs #3322, #3321.

PR #3557, opened 2026-09-11
Save the compute pipeline cache on the frontend's _Exit path

### LINKED ISSUE #3425: Compute pipeline cache is not saved on prosper-app exit
The interactive frontend exits through `std::_Exit`, bypassing `VulkanComputeContext` destruction. `flush_live_compute_pipeline_cache()` is declared and defined but has no caller; its header currently says the app calls it. Consequently the existing compute disk cache can load on launch but an ordinary app close does not persist newly compiled compute data through this path.

Verified at `b4c2cd862`: source search finds only the declaration in `frontends/shared/live/live_compute.hpp` and definition in `live_compute.cpp`. `prosper-app/main.cpp` calls the timing summary and the new graphics snapshot before `_Exit`, but no compute snapshot. This corrects the assumption in #3378 that compute persistence already handles the frontend exit.

Do not simply add a main-thread call: the context pointer is published non-atomically and the persistence method checks mutable failure flags before taking its cache mutex. A safe shutdown snapshot needs a reviewed publication/synchronization contract and an actual separate-launch save/load guard. Keep loading opt-in because of the recorded NVIDIA blob-loading crash.

mattias800:
Fixes #3425

## What was actually true before this change

I verified both halves rather than taking the issue's word for it:

- **Load worked.** `create_pipeline_cache()` read the file and passed it to `vkCreatePipelineCache`, with a retry-empty fallback if the driver rejected an identity-compatible blob.
- **Save had no caller.** `flush_live_compute_pipeline_cache()` was declared in `live_compute.hpp` and defined in `live_compute.cpp`, and a repo-wide search finds no third occurrence. Its own header comment claimed *"prosper-app calls this explicitly"*. The only other writer was `~VulkanComputeContext`, and `prosper-app` leaves through `std::_Exit`, which runs no destructor. The graphics sibling `flush_live_graphics_pipeline_cache()` **was** called two lines above, which is exactly why the gap read as handled.

So the cache loaded on every launch and was written on none. **Every run paid first-use pipeline compilation forever.**

## Framing: this is a cold-start cost, not a steady-state one

#3450 measured 14 first-use compilations costing 821 ms — 66% of compute time — in one 5 s GTA V world window, and it amortises to ~1.0% over a 929-dispatch window. So a warm cache does not make the steady state faster. What it removes is a cost that, without a save, is paid again on every single launch.

## The synchronization contract the issue asked for

The issue explicitly said *"do not simply add a main-thread call"*. It named two hazards; I found a third.

1. **The context pointer was published non-atomically.** `g_live_compute_context` is now `std::atomic<VulkanComputeContext*>`, stored with release at the publication point and in the teardown handler, loaded with acquire by every reader.
2. **The persistence method read mutable failure flags before taking its mutex.** `device_lost` and `completion_unproven` are now `std::atomic<bool>`; the snapshot reads both with acquire and declines *loudly* when either is set — a lost device or an unproven completion means Vulkan may still own objects here.
3. **`GpuRetilePipeline::initialize` creates a compute pipeline into the shared `VkPipelineCache` without holding the cache lock.** Every other creator already did (`packed_rtt_conversion`, and both dispatch-pipeline sites). `VkPipelineCache` is externally synchronized, and this change turns the snapshot into a genuine concurrent reader rather than a destructor that runs when nothing else can, so that site is now locked too.

The lock itself becomes a `std::timed_mutex` with a bounded `try_lock_for` (1000 ms, matching the graphics path). A driver compile that never returns must cost this run's cache, never an app that will not close.

## Exit paths, including the one this does not cover

| path | covered? |
| --- | --- |
| normal quit / window close with a booted guest | **yes** — the new call, next to the graphics flush, *after* the submit-gate drain and `vkDeviceWaitIdle` that make the read safe |
| app exit with no guest booted (library UI) | yes — that route returns from `main`, so the `atexit` teardown runs the destructor, which still calls the same method |
| **guest-thread `_Exit`** (`k_exit`, `k_debug_raise_release`) | **no**, and this is stated at the call site. #3324 is exit_group() from a thread the frontend never reaches; no frontend finalize step runs at all. Building the cross-layer finalizer registry #3324 describes is not this PR's job. |
| crash | no |

## Cache validity, location, and concurrency

The compute path wrote the **raw driver blob** with a hand-rolled temp+rename. A truncated file still passed the Vulkan header check (the header carries no length or checksum), and two concurrent writers shared one nonce scheme with a `remove`-then-`rename` fallback that could leave no file at all.

It now uses `PipelineCacheFile` — the checked envelope the graphics cache already used, which records length, an FNV checksum, driver version and pointer size, and reserves a private sibling *directory* atomically so no writer truncates another's temporary or deletes the previous good cache. `PipelineCacheFile::graphics()` and the new `compute()` now differ only in the filename and the explicit-path variable, so a policy change cannot land on one and miss the other.

- **Location** is unchanged in kind: `%LOCALAPPDATA%\prosper\` or `$XDG_CACHE_HOME`/`~/.cache/prosper/`. Not in git, not in `/tmp`.
- **Identity moves to `compute-vkcache-v2-…`.** A v1 file written by an older build is a different format; giving it a different name means the two can never overwrite each other while several builds run on one box. A v1 file is simply not read and the run starts cold.
- **Stale or corrupt is a cache miss, visibly.** A rejected file leaves the run cold and says so on stderr (`cold` vs `loaded`), and the retry-empty fallback on a driver rejection is retained. Loading stays **opt-in** behind `PROSPER_DISK_PIPELINE_CACHE` or an explicit path, because of the recorded NVIDIA blob-load crash.

`compute_pipeline_cache_blob_compatible()` keeps its exported name and its existing coverage in `test_game_compute`, but delegates to `PipelineCacheFile::compatible` so there is one decode rather than two.

## Regression guard — `compute_cache_persistence`

A fixture runs one **real compute dispatch** (a recompiled RDNA2 `buffer_store_dwordx4`, verified by reading the written words back) and then leaves through `std::_Exit` itself, so destructor-only persistence cannot pass it. The Python driver chains **separate processes**:

- `save` → cold start required, flush must return true, flush must be idempotent
- `load` → the next process must have bytes accepted by the driver
- `disabled` (`PROSPER_NO_DISK_PIPELINE_CACHE=1`) → must decline rather than report a save, and must not touch the existing file
- damaged / truncated / empty file → each must leave the run cold, and a good file must still load afterwards
- **three concurrent savers** against one path → the file must still load and no `.tmp-` residue may remain

And the check no in-process test can make: the original defect was an **uncalled** function, so the driver also asserts `prosper-app/main.cpp` still names the flush.

Registered in **both** the Windows and the UNIX CMake blocks, deliberately: this defect is about the Windows/NVIDIA frontend's `_Exit` path and the cost was measured there, so a Linux-only guard would test the fix everywhere except where it was found.

### The without-fix arms both go red

| arm | result |
| --- | --- |
| remove the `flush_live_compute_pipeline_cache()` call from `main.cpp` | **FAILED** — `AssertionError: main.cpp exits through std::_Exit and must flush the compute pipeline cache there` |
| make the flush a no-op returning false (i.e. destructor-only, the #3425 state) | **FAILED** — `[compute-cache-fixture] FAIL: the explicit shutdown snapshot must write the cache` |

Both restored afterwards and the full suite re-run.

## Verification

Windows / MinGW (Vulkan SDK 1.4.350.0, RTX 4090), configured with `-DPROSPER_APP=ON` and `-DGAME_DUMP=<DUMP_ROOT>/PPSA24651-app0`:

```
cmake --build <WT>/prosper/build-win -j 12       -> exit 0
ctest --no-tests=error -j 4                      -> 348/349 passed, exit 8
```

The single failure is `session_start`, pre-existing on this machine and unrelated (#3540); 2 skips are the usual dump-parameterized ones. Test count is 349 (was 348 — this PR adds one).

`compute_cache_persistence` on its own: **1/1 passed, exit 0.**

Only Linux CI is blocking, and the same test is registered there.

## Measured effect

**Qualitative (contention-invariant), in the real frontend, not the test harness** — `prosper-app`, `PROSPER_RENDER=1`, `--frames`, cache path under `~/`:

| title | compute blob written on `_Exit` | reloaded by the next launch |
| --- | --- | --- |
| GTA V `PPSA04263` | **27,755 bytes** | yes — `disk pipeline cache loaded: … (27755 bytes)` |
| The Messenger `PPSA24651` | 36 bytes (an empty payload — this title compiles nothing the driver retains) | yes |

Before this change all of that was discarded at exit, on every run.

**Timing, GTA V boot-to-menu route, 120 frames, `PROSPER_COMPUTE_PHASE_TIMING=1`**, summing the `pipeline_ms` leaf over all 122 dispatches. The graphics cache was held warm in every arm so the delta is compute-only, and arms were alternated:

| arm | total `pipeline_ms` | slowest single dispatch | dispatches over 1 ms |
| --- | --- | --- | --- |
| cold #1 | 7.2 | 5.1 | 1 |
| cold #2 | 6.5 | 4.2 | 1 |
| warm #1 | **2.0** | 0.3 | **0** |
| warm #2 | **2.7** | 0.7 | **0** |

**Contention disclosure:** one foreign `prosper-app` (another lane's worktree) was running before and after all four arms, so under this project's discipline this is a *contended* timing pair — indicative, not admissible. It is reported because all four arms were equally contended, the arms alternated, and the discriminator is near-categorical (1 vs 0 dispatches above 1 ms). An earlier wall-clock A/B produced one clean 0-peer cold sample and no clean warm sample, so no wall-clock claim is made.

**Do not read the 4.5 ms above as #3450's 821 ms.** This is a boot-to-menu route with 122 dispatches and one first-use compile; #3450's figure is a world window with 929 dispatches and 14 compilations. Mine is a small lower bound from a cheaper route, not a reproduction.

## Deliberately unaffected

- Default behaviour with persistence off (the default) — the flush returns false and writes nothing.
- The graphics cache path, other than being reached through the shared `for_stage` factory with an unchanged `graphics-vkcache-v1` identity.
- Steady-state frame rate. This is cold start only.

## Risks

- The `compute-vkcache-v2` rename orphans any existing v1 file on disk. It is a cache under the user's cache directory, it is opt-in, and it self-heals on the next run; nothing reads it.
- `device_lost` / `completion_unproven` becoming atomic changes their access at ~12 sites. All are plain reads and plain stores, so the semantics are unchanged; the full suite passes.
- Locking `GpuRetilePipeline::initialize` adds one uncontended mutex acquire on a dispatch path that was already doing pipeline creation and buffer allocation there.

### DIFF
--- a/prosper/CMakeLists.txt
+++ b/prosper/CMakeLists.txt
@@ -2639,6 +2639,22 @@ if(TARGET prosper_core)
       set_tests_properties(game_compute_exec_scratch_poison PROPERTIES
         ENVIRONMENT "PROSPER_DECODE_SCRATCH_POISON=1")
 
+      # #3425: registered on Windows as well as in the UNIX block below, deliberately. This
+      # defect is about the WINDOWS/NVIDIA frontend's _Exit path and the cost it recovers was
+      # measured there (#3450), so a Linux-only guard would test the fix everywhere except
+      # where it was found. Both registrations run the same fixture and driver script.
+      if(Python3_Interpreter_FOUND)
+        add_executable(test_compute_cache_persistence tests/gpu/execute/test_compute_cache_persistence.cpp)
+        target_include_directories(test_compute_cache_persistence PRIVATE tests frontends)
+        target_link_libraries(test_compute_cache_persistence PRIVATE prosper_live_renderer)
+        add_test(NAME compute_cache_persistence
+                 COMMAND "${Python3_EXECUTABLE}"
+                         "${CMAKE_CURRENT_SOURCE_DIR}/tests/gpu/execute/test_compute_cache_persistence.py"
+                         $<TARGET_FILE:test_compute_cache_persistence>
+                         "${CMAKE_CURRENT_SOURCE_DIR}/frontends/prosper-app/main.cpp")
+        set_tests_properties(compute_cache_persistence PROPERTIES TIMEOUT 300)
+      endif()
+
       # Local-only realized-submit replayer (#514), on Windows since #2331. Same target as the UNIX
       # block below, including the -Werror=switch gate: gpu_replay maps LiveTargetPixelFormat itself
       # in inspect_live_target, and an unhandled member there falls through to the seed's Rgba8Unorm
@@ -3390,6 +3406,20 @@ if(TARGET prosper_core)
                          "${CMAKE_SOURCE_DIR}/tools/perf/test_compute_buffer_timing_runtime.py"
                          $<TARGET_FILE:test_compute_buffer_timing>)
         set_tests_properties(compute_buffer_timing_runtime PROPERTIES TIMEOUT 180)
+        # #3425: the compute driver pipeline cache is written on the frontend's _Exit path, not by a
+        # destructor. Separate processes are required -- a single process could pass while writing
+        # nothing -- and the driver also asserts prosper-app still names the flush.
+        add_executable(test_compute_cache_persistence tests/gpu/execute/test_compute_cache_persistence.cpp)
+        target_include_directories(test_compute_cache_persistence PRIVATE tests frontends)
+        target_link_libraries(test_compute_cache_persistence PRIVATE
+                             prosper_live_renderer prosper_core Vulkan::Vulkan)
+        add_test(NAME compute_cache_persistence
+                 COMMAND "${Python3_EXECUTABLE}"
+                         "${CMAKE_CURRENT_SOURCE_DIR}/tests/gpu/execute/test_compute_cache_persistence.py"
+                         $<TARGET_FILE:test_compute_cache_persistence>
+                         "${CMAKE_CURRENT_SOURCE_DIR}/frontends/prosper-app/main.cpp")
+        set_tests_properties(compute_cache_persistence PROPERTIES TIMEOUT 300)
+
         add_executable(test_compute_buffer_residency tests/gpu/execute/test_compute_buffer_residency.cpp)
         target_include_directories(test_compute_buffer_residency PRIVATE tests frontends)
         target_link_libraries(test_compute_buffer_residency PRIVATE
--- a/prosper/frontends/prosper-app/main.cpp
+++ b/prosper/frontends/prosper-app/main.cpp
@@ -3251,7 +3251,19 @@ int main(int argc, char** argv) {
         // _Exit skips RuntimeComputeTimingSelector's destructor. Publish its validity verdict here;
         // the report is idempotent so a future cooperative teardown cannot duplicate it.
         prosper::frontend::report_live_compute_timing_selector_summary();
+        // Both driver pipeline caches, for the same reason: _Exit runs no destructor, so the
+        // compute context's own teardown save never happens on this path (#3425). Compute is the
+        // one that had no caller at all -- every run paid first-use shader compilation forever,
+        // measured at 821 ms across 14 compilations in one 5 s GTA V world window (#3450).
+        //
+        // Placed AFTER the drain and vkDeviceWaitIdle above: those are what make it safe to read a
+        // cache the guest thread was still mutating. Both calls are bounded and both decline
+        // loudly rather than blocking the close.
+        //
+        // This covers the frontend's own exit only. A guest-thread _Exit (#3324) is exit_group()
+        // from a thread this code never reaches, so neither cache is saved on that path.
         prosper::frontend::flush_live_graphics_pipeline_cache();
+        prosper::frontend::flush_live_compute_pipeline_cache();
 #endif
         // The bounded dmem writer trace also has a destructor/atexit fallback, which _Exit skips.
         prosper::host::guest_dmem_write_trace_report();
--- a/prosper/frontends/shared/device/AGENTS.md
+++ b/prosper/frontends/shared/device/AGENTS.md
@@ -4,6 +4,11 @@ Loader/device requirements and selection shared by live frontends and execution 
 Pipeline-cache files belong here: they store opaque driver data, not guest shaders or game assets.
 Cache compatibility checks reject damaged or mismatched files; they cannot guarantee a driver
 accepts its own data safely. Keep disk loading opt-in while the recorded NVIDIA failure remains.
+`PipelineCacheFile` serves BOTH stages -- `graphics()` and `compute()` differ only in the
+filename and the explicit-path variable -- so a policy change cannot land on one and miss the other,
+which is how the compute path ended up writing a bare driver blob with no truncation guard (#3425).
+Note where the file is written from: a frontend that leaves through `_Exit` runs no destructor, so
+persistence is an explicit call on the shutdown path, not a teardown side effect.
 
 Feature acquisition that more than one device path needs belongs here rather than beside whichever
 device was written first: the renderer's device, the compute backend's own device and the adopt path
--- a/prosper/frontends/shared/device/pipeline_cache_file.hpp
+++ b/prosper/frontends/shared/device/pipeline_cache_file.hpp
@@ -47,13 +47,17 @@ struct PipelineCacheFile {
             read_le(blob.data() + 12, 4) == device.deviceID &&
             std::memcmp(blob.data() + 16, device.pipelineCacheUUID, VK_UUID_SIZE) == 0;
     }
-    static PipelineCacheFile graphics(const VkPhysicalDeviceProperties& properties) {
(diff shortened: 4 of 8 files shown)


mattias800 (commented):
APPROVED — orchestrator review on exact head `5ece4710`.

## The diagnosis is better than the issue asked for

The issue said the cache "loads but never saves". The lane verified **both halves** rather than
inheriting that framing, and what it found is sharper: `flush_live_compute_pipeline_cache` occurs
exactly **twice in the tree** — declaration and definition — so the save had *no caller at all*, while
its own header comment claimed prosper-app called it.

The detail that explains why nobody caught it: **the graphics sibling *was* flushed two lines above
the `_Exit`**. That adjacency is what made the missing compute flush read as handled by every reader,
including whoever wrote the false comment.

## Two defects the issue never named, found in passing

- **`GpuRetilePipeline::initialize` wrote into the shared `VkPipelineCache` without the cache lock**,
  unlike every other creator. Harmless while the only reader was a destructor; not harmless once a
  main-thread snapshot exists. Now locked.
- **The on-disk format was the raw driver blob, and a truncated file passed validation** — Vulkan's
  own cache header carries no length or checksum. Moved onto the `PipelineCacheFile` envelope the
  graphics side already used (length, FNV checksum, driver version, pointer size, atomic cross-process
  reservation), identity bumped to `compute-vkcache-v2` so old and new files cannot overwrite each
  other. Loading stays opt-in.

## Concurrency, which is the risky part and which I read rather than trusted

```cpp
if (device_lost.load(std::memory_order_acquire) ||
    completion_unproven.load(std::memory_order_acquire)) { … return false; }
…
std::unique_lock<std::timed_mutex> lock(pipeline_cache_mutex, std::defer_lock);
if (!lock.try_lock_for(lock_budget)) { … return false; }
```

Correct on each count I checked: the device-state atomics are read with acquire ordering **before**
the cache handle is touched, since a lost device or unproven completion means Vulkan may still own
objects there; the `VkPipelineCache` is externally synchronised and the snapshot serialises against
every `vkCreateComputePipelines`; and the lock is **bounded**, so a hung driver compile costs the
cache rather than the app close. Every failure path prints a reason — none of them fail silently,
which matters for a feature whose absence was invisible for this long.

## Exit-path coverage, stated honestly

Normal quit covered; the no-guest path covered by the destructor; **guest-thread `_Exit` (#3324) is
NOT covered**, and that is documented at the call site rather than papered over.

## Measurement discipline

- Qualitative and contention-invariant: GTA V writes a **27,755-byte** compute blob on the real
  `_Exit` path and the next launch loads it. Previously discarded every run.
- The Messenger writes **36 bytes** — an empty payload. Worth knowing before anyone probes this with
  a 2D title.
- The timing A/B (cold 7.2/6.5 ms vs warm 2.0/2.7 ms over 122 dispatches) was **contended and marked
  inadmissible** — one foreign `prosper-app` at both endpoints of all four arms. That interference was
  my own flip-pacing lane, so the contention was my scheduling rather than the author's method.
  Refusing to quote it as a result, and stating plainly that it is **not** a reproduction of #3450's
  821 ms (a 929-dispatch world window against a 122-dispatch boot route), is the discipline the
  orchestration doc asks for. A weaker report would have led with 7.2 → 2.0 ms.

## On the red Windows MinGW check — resolved, and worth recording

The author escalated it with a specific hypothesis: that CI's Windows runner has no GPU and the new
fixture needs a real dispatch. **That hypothesis was wrong twice over**, and I read the log rather
than accepting or dismissing it:

```
FAILED: [code=1] test_dlsym_handle.exe
collect2.exe: error: ld returned 5 exit status
ninja: build stopped: subcommand failed.
```

A **link failure in a binary this diff does not touch**, at the build step — the only `ctest` string
in the whole job is a CMake warning, so ctest never ran and no device was ever required.
**Re-running the job passes**, so it was transient and not caused by this PR.

The author's own retrospective is the part worth keeping: *"a proposed remedy lends a hypothesis more
weight than the evidence behind it earned."* The guess arrived with a one-line fix attached — drop the
Windows test registration — and that framing is what made it feel diagnosed. Had it been taken at face
value, the registration would have been removed for a reason that does not hold, discarding the
coverage argument that put it there.

## Verification

Windows/MinGW `-DPROSPER_APP=ON`: build exit 0; `ctest --no-tests=error -j 4` → 348 of 349, sole
failure `session_start` (pre-existing then, now fixed on `main` by #3556); count rose 348 → 349.
**Linux CI green** — the first run of `compute_cache_persistence` on lavapipe, which is the one thing
that could not be verified locally (WSL has Vulkan 1.3.275 against the repo's 1.4). Both without-fix
arms red, restored, suite re-run.

